PR #7300, opened 2020-12-04
std: Add a few helpers for matching ascii strings

LemonBoy:
* startsWithIgnoreCase
* endsWithIgnoreCase

### DIFF
--- a/lib/std/ascii.zig
+++ b/lib/std/ascii.zig
@@ -339,7 +339,7 @@ pub fn startsWithIgnoreCase(haystack: []const u8, needle: []const u8) bool {
     return if (needle.len > haystack.len) false else eqlIgnoreCase(haystack[0..needle.len], needle);
 }
 
-test "mem.startsWith" {
+test "ascii.startsWithIgnoreCase" {
     std.testing.expect(startsWithIgnoreCase("boB", "Bo"));
     std.testing.expect(!startsWithIgnoreCase("Needle in hAyStAcK", "haystack"));
 }
@@ -348,7 +348,7 @@ pub fn endsWithIgnoreCase(haystack: []const u8, needle: []const u8) bool {
     return if (needle.len > haystack.len) false else eqlIgnoreCase(haystack[haystack.len - needle.len ..], needle);
 }
 
-test "mem.endsWithIgnoreCase" {
+test "ascii.endsWithIgnoreCase" {
     std.testing.expect(endsWithIgnoreCase("Needle in HaYsTaCk", "haystack"));
     std.testing.expect(!endsWithIgnoreCase("BoB", "Bo"));
 }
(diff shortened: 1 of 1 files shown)


daurnimator (approved):

FireFox317:
Wouldn't it be better to have a single `startsWith` function and use an enum parameter to specify whether you want to ignore case or not? Or maybe even a boolean, but I think an enum is more clear.

LemonBoy:
> Wouldn't it be better to have a single `startsWith` function and use an enum parameter to specify whether you want to ignore case or not? Or maybe even a boolean, but I think an enum is more clear.

Not really, `startsWith` is a generic function and works with every kind of slice, even non-u8 ones.
Ignoring the string case makes only sense for ascii strings (unicode case-folding is a completely different beast), hence the placement in `std.ascii` and the prototype with no type parameter.

lib/std/ascii.zig
```diff
@@ -335,6 +335,24 @@ test "eqlIgnoreCase" {
     std.testing.expect(!eqlIgnoreCase("hElLo!", "helro!"));
 }
 
+pub fn startsWithIgnoreCase(haystack: []const u8, needle: []const u8) bool {
+    return if (needle.len > haystack.len) false else eqlIgnoreCase(haystack[0..needle.len], needle);
+}
+
+test "mem.startsWith" {
```
kubkon, line 342: I think this would be a bit more meaningful for a test case name, wouldn't you agree?

```suggestion
test "ascii.startsWithIgnoreCase" {
```

lib/std/ascii.zig
```diff
@@ -335,6 +335,24 @@ test "eqlIgnoreCase" {
...
+    std.testing.expect(startsWithIgnoreCase("boB", "Bo"));
+    std.testing.expect(!startsWithIgnoreCase("Needle in hAyStAcK", "haystack"));
+}
+
+pub fn endsWithIgnoreCase(haystack: []const u8, needle: []const u8) bool {
+    return if (needle.len > haystack.len) false else eqlIgnoreCase(haystack[haystack.len - needle.len ..], needle);
+}
+
+test "mem.endsWithIgnoreCase" {
```
kubkon, line 351: Same suggestion here:

```suggestion
test "ascii.endsWithIgnoreCase" {
```

kubkon (approved):
Looks good to me, just a couple of nitpicks.

lib/std/ascii.zig
LemonBoy: Indeed, I did copy&paste the test cases from `mem.zig` and forgot to change the names.

andrewrk: the prefix is actually redundant, the full test name ends up getting prefixed with ascii anyway since it's in ascii.zig. But it's harmless, no need to go through another round of changes IMO

PR #7301, opened 2020-12-04
Make the CI script more robust

LemonBoy:


andrewrk:
Thanks. BTW I'm planning to take the same strategy with the Linux x86_64 CI as we do with macOS, using a zig+llvm+lld+clang tarball with zig as the C/C++ compiler, thus eliminating the dependency on apt.llvm.org altogether, as well as on docker. (However the way the aarch64 linux CI works is via docker images)

Issue #7053, opened 2020-11-10
Proposal: make switch(exhaustive-enum) safer


cajw1:
[Accepted Proposal](https://github.com/ziglang/zig/issues/7053#issuecomment-977216391)

-----

```zig
const std = @import("std");

const SparseU32 = enum(u32){one=1,};

pub fn main() void {

    var rtu32 : u32 = 99;
    var sparse : SparseU32 = .one;
    sparse = @ptrCast(*SparseU32,&rtu32).*;

    // 1.
    // Note: sparse is not undefined here, 
    // and branching on it using an if statement 
    // is fine including in release-fast

    std.debug.warn("Observing sparse using if...\n", .{}); 
    if(sparse==.one){
        std.debug.warn("observed .one ({}) \n", .{@bitCast(u32,sparse)}); 
    } else {
        std.debug.warn("observed NOT .one but {} \n", .{@bitCast(u32,sparse)}); 
    }
        // ==> This prints:
        // Observing sparse using if...
        // observed NOT .one but 99

    // 2.
    // However, branching on sparse using a switch is
    // either a bug, or UB (which I would argue is entirely
    // avoidable):

    std.debug.warn("Observing sparse using switch...\n", .{}); 
    switch(sparse){
        .one => {std.debug.warn("observed .one ({}) \n", .{@bitCast(u32,sparse)});}
        // This proposal suggest that this hidden prong 
        // be executed for every invalid value of sparse:
        // _ => {@panic("helpful invalid value message");}
    }
        // ==> This prints:
        // Observing sparse using switch...
        // observed .one (99) 

    std.debug.warn("we get here even in release-fast \n", .{});
}
```
Switching on an exhaustive enum with a tagtype capable of holding invalid values cannot guarantee that exactly one of its branches executes (or if it does, it will be the wrong branch) because Zig cannot exclude the possibility of an invalid value. Since we know this can happen, I suggest that a switch on an exhaustive enum without an else prong is always (including in --release-fast) defined as if it had an implicit _ prong which panics. I don't see why the switch should behave differently to an equivalent if ... inspection.

andrewrk:
Here is the relevant accepted portion of the language spec:

[...]

PR #7313, opened 2020-12-05
Fix a few unsound optimizations on single-element union/enum

LemonBoy:
Found this while investigating #7053, check the commit messages for more info.
Long story short, we can't elide checks if we're not 100% sure the enum value is valid (both at comptime and runtime).

The check on `tag_int_type` could be simplified with a single `type_has_one_possible_value(g, tag_int_type) == OnePossibleValueYes` if you think that's clearer.

I've also noted that `@enumToInt` does no validation on the value the user passes in, is that a bug or a feature?


andrewrk:
> I've also noted that `@enumToInt` does no validation on the value the user passes in, is that a bug or a feature?

Here's a proposal to track this question: #7320
My gut instinct is that status quo is better and we should keep status quo.

PR #7316, opened 2020-12-06
add AddressFamilyNotSupported to SendError

lun-4:
[POSIX reference](https://pubs.opengroup.org/onlinepubs/009695399/functions/sendto.html).

I'm not sure what `fs.File.writeFileAll` should do in the presence of EAFNOSUPPORT. (should it become part of `WriteFileError` or something else?)


marler8997:
Maybe send and sendto should have different error types?

PR #7321, opened 2020-12-06
Stage 2: Move wasm backend codegen to codegen.zig 

Luukdegram:
This PR moves the wasm codegen to codegen.zig. Creating this PR as a draft version to receive feedback while I finish this. A few things are remaining such as return values and constants.
cc: @ifreund 

ps: I'm still trying to find my way around stage2's architecture, so any tips are welcome!

Currently the following works:
```zig
export fn _start() u32 {
    pop(1, 2);
    return 5;
}

fn pop(a: u32, b: u32) void {}
```

Validated with [wabt's wasm-validate](https://github.com/WebAssembly/wabt).
Using the above `wasm-decompile` generates the following result:
```
export function start():int {
  f_b(1, 2)
  return 5
}

function f_b(a:int, b:int) {
}
```


src/codegen.zig
```diff
@@ -80,8 +80,8 @@ pub fn generateSymbol(
     switch (typed_value.ty.zigTypeTag()) {
         .Fn => {
             switch (bin_file.options.target.cpu.arch) {
-                .wasm32 => unreachable, // has its own code path
-                .wasm64 => unreachable, // has its own code path
+                .wasm32 => return Function(.wasm32).generateSymbol(bin_file, src, typed_value, code, debug_output),
+                .wasm64 => return Function(.wasm64).generateSymbol(bin_file, src, typed_value, code, debug_output),
```
ifreund, line 84: wasm64 is currently vaporware and this should remain unreachable until there exists a runtime with support for it.

src/codegen.zig
```diff
@@ -637,6 +637,60 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
                         try self.dbgSetEpilogueBegin();
                     }
                 },
+                .wasm32, .wasm64 => {
```
ifreund, line 640:
```suggestion
                .wasm32 => {
```

src/link/Wasm.zig
```diff
@@ -10,14 +10,14 @@ const log = std.log.scoped(.link);
...
-const codegen = @import("../codegen/wasm.zig");
+const codegen = @import("../codegen.zig");
 const link = @import("../link.zig");
 const trace = @import("../tracy.zig").trace;
 const build_options = @import("build_options");
 const Cache = @import("../Cache.zig");
 
 /// Various magic numbers defined by the wasm spec
-const spec = struct {
+const Spec = struct {
```
ifreund, line 20: this is a namespace, namespaces are snake_case by convention.

src/link/Wasm.zig
```diff
@@ -224,27 +239,15 @@ pub fn flushModule(self: *Wasm, comp: *Compilation) !void {
...
-                // in codegen.wasm.genCode() simpler.
-                var buf: [5]u8 = undefined;
-                leb.writeUnsignedFixed(5, &buf, self.getFuncidx(idx_ref.decl).?);
-                try writer.writeAll(&buf);
-            }
+            const fn_data = &decl.fn_link.wasm;
 
-            try writer.writeAll(fn_data.code.items[current..]);
+            // Write the already generated code to the file
+            try writer.writeAll(fn_data.code.items[decl.link.wasm.code_offset..]);
```
ifreund, line 245: How are incremental updates which shift function indexes handled without this logic?

ifreund (commented):
Just a drive-by review for now, will look in more depth later.

src/link/Wasm.zig
Luukdegram: When allocateDeclIndexes is called, a check is done to see if the decl already exists in the `funcs` list. If not, they're appended and assigned an id. Meaning that if function indexes shift, they keep their old index. This same index id is then used for the function call. However, I just realized that this won't work for removing functions and that will mess up the order and therefore the id's. Thanks for your feedback @ifreund, I'll take another look at this tonight. Appreciate your time.

Luukdegram: Reverted to the old logic.

Luukdegram:
Rebased to match the `Type` refactoring from 133da8692e80532797dd91b32539cf2175280a95

@ifreund Any chance you'd be able to do a review in the near future?

ifreund (commented):
As I'm sure you're aware, the design direction for codegen was discussed in the first stage2 meeting. You can see notes here: https://github.com/ziglang/zig/wiki/stage2:-meeting-notes--2020--12--10.

Specifically this bullet point is relevant:

> codegen.zig should not contain any architecture-dependent code, abstract only over stuff that is shared in all backends

I'm not experienced enough with codegen in general to know how well the abstractions required for generating stack machine/wasm code will align with those needed for register machines. It seems that integrating the current (very limited) wasm codegen with our current `codegen.zig` abstractions requires workarounds such as this fake registry with 100 registers. 

As it's very unclear to me what the global maximum we want to find will look like, I find myself wondering if it would not make more sense to develop the wasm backend further separately from the current `codegen.zig` register abstractions and see what wasm-specific abstractions naturally arise.

Luukdegram:
After a discussion with ifreund on Discord, we came to the conclusion to close the PR for now until we implement a broader functionality for the wasm backend and get a better idea of what is needed/missing to move to `codegen.zig` or if we even should at all. As can be found in the PR there are a few workarounds as well as a sacrifice of readability to make it work, which made us come to the decision to first implement more functionality in the wasm codegen before revisiting this.

Issue #7253, opened 2020-11-29
Add stack size override option in std.build.LibExeObjStep


chr-1x:
https://github.com/ziglang/zig/commit/0ff5d7b24e2017199566d1bab75254fa5ef68611 added a `--stack` flag which allows manual override of the stack size until the stack overflow obsolescence is implemented. However, this option is not exposed in the build system, which means it's much harder to use with projects that rely on e.g. packages. It probably only makes sense for executables but I'll leave that to the judgement of the implementer.

faraazahmad:
@antlilja is this issue fixed?

antlilja:
> @antlilja is this issue fixed?

Yes. This issue should probably be closed.

PR #7327, opened 2020-12-07
Add option to override default stack size in build system.

antlilja:
Implements #7253.

This PR also adds the option --stack to the build system itself to override all stack sizes set by setStackSize. But maybe this isn't that useful and should be removed?

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -1291,6 +1291,9 @@ pub const LibExeObjStep = struct {
 
     subsystem: ?builtin.SubSystem = null,
 
+    /// Overrides the default stack size
+    stack_size: ?u64 = null,
+
     const LinkObject = union(enum) {
         StaticPath: []const u8,
         OtherStep: *LibExeObjStep,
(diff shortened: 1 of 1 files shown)


andrewrk:
> This PR also adds the option --stack to the build system itself to override all stack sizes set by setStackSize. But maybe this isn't that useful and should be removed?

Yes I think this should be removed. This is a knob that makes sense for the project maintainer to control, not the end user. The project maintainer always has the ability to expose the stack size to the end user if they decide to.

antlilja:
I have two seperate stack_size fields, one in Builder and one in LibExeObjStep, so the PR already has the option as per build artifact. But I'll remove the global option.

lib/std/build.zig
```diff
@@ -1714,6 +1717,10 @@ pub const LibExeObjStep = struct {
         self.libc_file = libc_file;
     }
 
+    pub fn setStackSize(self: *LibExeObjStep, stack_size: u64) void {
+        self.stack_size = stack_size;
+    }
+
```
andrewrk, line 1723: No trivial setters please. The field can be the public interface.

andrewrk:
Thanks!

PR #7331, opened 2020-12-07
Massage CityHash to work at comptime

SpexGuy:
@Sinitax ran into issues trying to use CityHash at compile time.  This PR modifies the implementation and tests to work at compile time.

### DIFF
--- a/lib/std/hash/cityhash.zig
+++ b/lib/std/hash/cityhash.zig
@@ -6,20 +6,25 @@
 const std = @import("std");
 const builtin = @import("builtin");
 
+inline fn offsetPtr(ptr: [*]const u8, offset: usize) [*]const u8 {
+    // ptr + offset doesn't work at comptime so we need this instead.
+    return @ptrCast([*]const u8, &ptr[offset]);
+}
+
+fn fetch32(ptr: [*]const u8, offset: usize) u32 {
+    return std.mem.readIntLittle(u32, offsetPtr(ptr, offset)[0..4]);
+}
+
+fn fetch64(ptr: [*]const u8, offset: usize) u64 {
+    return std.mem.readIntLittle(u64, offsetPtr(ptr, offset)[0..8]);
+}
+
 pub const CityHash32 = struct {
     const Self = @This();
 
     // Magic numbers for 32-bit hashing.  Copied from Murmur3.
     const c1: u32 = 0xcc9e2d51;
     const c2: u32 = 0x1b873593;
-
-    fn fetch32(ptr: [*]const u8) u32 {
-        var v: u32 = undefined;
-        @memcpy(@ptrCast([*]u8, &v), ptr, 4);
-        if (builtin.endian == .Big)
-            return @byteSwap(u32, v);
-        return v;
-    }
 
     // A 32-bit to 32-bit integer hash copied from Murmur3.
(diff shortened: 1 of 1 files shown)


jedisct1:
Looking good. Does it have any performance implications?

SpexGuy:
Performance in debug mode may be impacted by the new `offsetPtr` function.  We could mark `offsetPtr` and `fetch32`/`fetch64` as `inline` to mitigate this.  But I don't think I've introduced any new bounds checks.  Runtime performance in optimized mode should be the same.  The comptime execution for the tests is pretty slow (~15 seconds on my computer), so that will affect the total std test time.  The change to the tests to [write individual bytes](https://github.com/ziglang/zig/pull/7331/files#diff-f9b4ea1e7a52edb7117030a535e2ba308cdbcb77213750646102d33305b67f5eL376-R391) instead of type punning may also cause a small increase in test time.

lib/std/hash/cityhash.zig
```diff
@@ -6,16 +6,22 @@
...
     // Magic numbers for 32-bit hashing.  Copied from Murmur3.
     const c1: u32 = 0xcc9e2d51;
     const c2: u32 = 0x1b873593;
 
-    fn fetch32(ptr: [*]const u8) u32 {
-        var v: u32 = undefined;
-        @memcpy(@ptrCast([*]u8, &v), ptr, 4);
+    fn fetch32(ptr: [*]const u8, offset: usize) u32 {
+        // ptr + offset doesn't work at comptime so we need this instead.
+        const offset_ptr = offsetPtr(ptr, offset)[0..4];
```
LemonBoy, line 23: What about 
```zig
return mem.readIntNative(u32, offsetPtr(ptr, offset)[0..4]);
```

Ditto for `fetch64` (and the other two identical fetch helpers? Can you just pull them at top level and delete a copy?)

SpexGuy: Huh, I thought these wouldn't work at comptime but apparently they do.  Since the next line is byte swap, I can replace the whole function with `readIntLittle`.

`[offset .. offset + 4]` works at comptime but at runtime this would produce a slice, so I'm sticking with the other one.

SpexGuy:
Looks like the additional comptime execution for these tests might be putting the test build over the memory limit.  That might be a reason to hold off on this until stage 2.

lib/std/hash/cityhash.zig
dbandstra: This should work at runtime:

```zig
return mem.readIntLittle(u32, ptr[offset .. offset + 4][0..4]);
```

(With slices, you can do `[offset..][0..4]` which is a nice little idiom, but that doesn't work for length-less pointers)

SpexGuy: That works but introduces a new but useless runtime check in debug modes (useless because it's only checking that `offset + 4 - offset == 4`).
`ptr[offset .. offset + 4].ptr[0..4]` would work, but for consistency I think it's worth sticking with `offsetPtr` everywhere in this file.

andrewrk:
Landed in 5cc2e500e68a60cb99b48755af39856ff60198b6 but with the comptime execution commented out in the tests. Idea being that we will not have test coverage of this for now, and re-enable them once we ship stage2.

PR #7332, opened 2020-12-07
Fix general purpose allocator incorrectly modifying total_requested_bytes in some cases.

Ekdohibs:
When recovering from a double free, the general purpose allocator will not run its `errdefer` and thus incorrectly modify `self.total_requested_bytes`.


andrewrk:
Thanks!

PR #7333, opened 2020-12-07
Implement msp430 mapping for c_longdouble.

nmichaels:


src/stage1/codegen.cpp
```diff
@@ -8658,6 +8658,9 @@ static void define_builtin_types(CodeGen *g) {
             // It's either a float or a double, depending on a toolchain switch
             add_fp_entry(g, "c_longdouble", 64, LLVMDoubleType(), &g->builtin_types.entry_c_longdouble);
             break;
+        case ZigLLVM_msp430:
+            add_fp_entry(g, "c_longdouble", 32, LLVMFloatType(), &g->builtin_types.entry_c_longdouble);
```
LemonBoy, line 8662: The ABI [documents](https://www.ti.com/lit/an/slaa534a/slaa534a.pdf?ts=1607421649531&ref_url=https%253A%252F%252Fwww.google.com%252F) `long double` as an alias for float64.

nmichaels: So it does. That's what I get for using random web sites as sources instead of the actual documentation.

PR #7334, opened 2020-12-07
Support casting enums to all int types in translate-c.

ehaas:
In C, enums are represented as signed integers, so casting from an enum to an integer
should use the "cast integer to integer" translation code path. Previously it used the
"cast enum to generic non-enum" code path, because enums were not being treated as integers.
Ultimately this can produce zig code that fails to compile if the destination type does not
support the full range of enum values (e.g. translated C code that casts an enum value to an
unsigned integer would fail to compile since enums are signed integers, and unsigned integers
cannot represent the full range of values that signed ones can).

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -1993,6 +1993,21 @@ fn transStringLiteral(
     }
 }
 
+fn cIsEnum(qt: clang.QualType) bool {
+    return qt.getCanonicalType().getTypeClass() == .Enum;
+}
+
+/// Get the underlying int type of an enum. The C compiler chooses a signed int
+/// type that is large enough to hold all of the enum's values. It is not required
+/// to be the smallest possible type that can hold all the values.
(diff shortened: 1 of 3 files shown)


ehaas:
Seems to have failed on Windows due to platform or compiler specific differences about whether an enum in a boolean context in C is implicitly cast to a signed or unsigned int? Since it's comparing to zero it doesn't actually matter, but it makes a difference regarding what Zig source is emitted. Any thoughts on how to address this?

Vexu (approved):

ehaas:
Is there a reason the Azure pipelines didn't run? That was how I caught the original problem.

Vexu:
> Is there a reason the Azure pipelines didn't run?

No, that's just how Azure is.

Apparently closing and reopening the pr will rerun the CI but I've had more success with just force pushing.

andrewrk:
Thanks @ehaas, nice work!

PR #7335, opened 2020-12-08
glibc: additionally provide -lcrypt

andrewrk:
Looks like we were missing support for this other library that glibc provides.

### DIFF
--- a/tools/update_glibc.zig
+++ b/tools/update_glibc.zig
@@ -22,6 +22,7 @@ const lib_names = [_][]const u8{
     "rt",
     "ld",
     "util",
+    "crypt",
 };
 
 // fpu/nofpu are hardcoded elsewhere, based on .gnueabi/.gnueabihf with an exception for .arm
--- a/lib/libc/glibc/abi.txt
+++ b/lib/libc/glibc/abi.txt
@@ -1655,6 +1655,8 @@ aarch64-linux-gnu aarch64_be-linux-gnu
 29
 29
 29
+29
+29
 37
 37
 37
@@ -1729,6 +1731,8 @@ aarch64-linux-gnu aarch64_be-linux-gnu
 29
 29
 38
+29
+29
 29
 29
 29
@@ -1910,6 +1914,7 @@ aarch64-linux-gnu aarch64_be-linux-gnu
 29
 29
 38
+29
 29
 29
 29
@@ -3251,6 +3256,8 @@ aarch64-linux-gnu aarch64_be-linux-gnu
 29
 29
 40
+29
+29
 29
 29
 29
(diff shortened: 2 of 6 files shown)


tools/update_glibc.zig
```diff
@@ -22,6 +22,7 @@ const lib_names = [_][]const u8{
     "rt",
     "ld",
     "util",
+    "crypt",
```
daurnimator, line 25: alphabetical order?

daurnimator (commented):
There might be other missing glibc libraries too?
```
$ pacman -Qql glibc | grep so | grep 2.32
/usr/lib/ld-2.32.so
/usr/lib/libBrokenLocale-2.32.so
/usr/lib/libanl-2.32.so
/usr/lib/libc-2.32.so
/usr/lib/libcrypt-2.32.so
/usr/lib/libdl-2.32.so
/usr/lib/libm-2.32.so
/usr/lib/libmvec-2.32.so
/usr/lib/libnsl-2.32.so
/usr/lib/libnss_compat-2.32.so
/usr/lib/libnss_db-2.32.so
/usr/lib/libnss_dns-2.32.so
/usr/lib/libnss_files-2.32.so
/usr/lib/libnss_hesiod-2.32.so
/usr/lib/libpthread-2.32.so
/usr/lib/libresolv-2.32.so
/usr/lib/librt-2.32.so
/usr/lib/libutil-2.32.so
```

libnss and ld you can probably skip; but the others....

daurnimator:
FYI: 
> new glibc drops libcrypt.h header and corresponding static library. The upstream plans to eventually remove libcrypt from glibc tree in favor of a separate project libxcrypt
> There's no breakage (famous last words) as glibc still ships libcrypt.so.1 for backwards compatibility, while libxcrypt package provides development headers and library with bumped soname version due to removal of the functions providing insecure hashes. You may need to add '#include <crypt.h>' if it was missing before as uninstd.h no longer provides a declaration of the crypt function.

andrewrk:
Ahh that's good to know. Thanks. In this case I think it should be reverted for simplicity's sake.

daurnimator:
> In this case I think it should be reverted for simplicity's sake.

maybe not?
As long as glibc keeps providing libcrypt.so.1 for backwards compatibility, we should too, right?

andrewrk:
Hmm, fair point. :thinking:

Jan200101:
on Fedora libcrypt.so.1 has already been dropped from the glibc package and is instead provided by libxcrypt

jedisct1:
Same on Arch Linux.

andrewrk:
6b7ddfbafe7ff8f9bc9952f838737e8514c9b26b

PR #7355, opened 2020-12-09
invoke LLD as a child process rather than a library

### LINKED ISSUE #3825: fatal lld errors don't get printed
If you run into a fatal error in LLD (i.e. if a section's sh_size is not a multiple of its entry size), then lld will print the error and then call `_exit`.  The fact that `lld` exits is a problem on its own, but this issue is for the missing error message.

With strace, I noticed that around the time that the fatal error message should be printed, LLD calls lseek on `stdout` and gets a `-ESPIPE` error.  This may or may not have something to do with why the error message does not get printed.

andrewrk:
Closes #3825

### DIFF
--- a/src/link/Coff.zig
+++ b/src/link/Coff.zig
@@ -907,11 +907,10 @@ fn linkWithLLD(self: *Coff, comp: *Compilation) !void {
         // Create an LLD command line and invoke it.
         var argv = std.ArrayList([]const u8).init(self.base.allocator);
         defer argv.deinit();
-        // The first argument is ignored as LLD is called as a library, set it
-        // anyway to the correct LLD driver name for this target so that it's
-        // correctly printed when `verbose_link` is true. This is needed for some
-        // tools such as CMake when Zig is used as C compiler.
-        try argv.append("lld-link");
+        // We will invoke ourselves as a child process to gain access to LLD.
+        // This is necessary because LLD does not behave properly as a library -
+        // it calls exit() and does not reset all global data between invocations.
+        try argv.appendSlice(&[_][]const u8{ comp.self_exe_path.?, "lld-link" });
 
         try argv.append("-ERRORLIMIT:0");
         try argv.append("-NOLOGO");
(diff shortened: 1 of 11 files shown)


src/link/Coff.zig
```diff
@@ -1149,45 +1148,51 @@ fn linkWithLLD(self: *Coff, comp: *Compilation) !void {
...
-            // directly outputting to stderr here.
-            std.debug.print("{}", .{stderr_context.data.items});
-            return error.LLDReportedFailure;
+
+        switch (term) {
+            .Exited => |code| {
+                if (code != 0) {
+                    // TODO parse this output and surface with the Compilation API rather than
+                    // directly outputting to stderr here.
+                    std.debug.print("{s}", .{stderr});
```
daurnimator, line 1184: why is this `std.debug.print` rather than `log.err`?

src/link/Wasm.zig
```diff
@@ -345,11 +345,10 @@ fn linkWithLLD(self: *Wasm, comp: *Compilation) !void {
...
     defer argv.deinit();
-    // The first argument is ignored as LLD is called as a library, set it
-    // anyway to the correct LLD driver name for this target so that it's
-    // correctly printed when `verbose_link` is true. This is needed for some
-    // tools such as CMake when Zig is used as C compiler.
-    try argv.append("ld-wasm");
+    // We will invoke ourselves as a child process to gain access to LLD.
+    // This is necessary because LLD does not behave properly as a library -
+    // it calls exit() and does not reset all global data between invocations.
+    try argv.appendSlice(&[_][]const u8{ comp.self_exe_path.?, "wasm-ld" });
```
daurnimator, line 351: did you mean to swap "ld-wasm" for "wasm-ld"?

andrewrk: The symlink that the LLD project installs is `wasm-ld`:

```
andy@ark ~> ls ~/local/llvm11-release/bin/*ld*
/home/andy/local/llvm11-release/bin/ld64.lld
/home/andy/local/llvm11-release/bin/ld.lld
/home/andy/local/llvm11-release/bin/lld
/home/andy/local/llvm11-release/bin/lld-link
/home/andy/local/llvm11-release/bin/llvm-rtdyld
/home/andy/local/llvm11-release/bin/scan-build
/home/andy/local/llvm11-release/bin/wasm-ld
```

src/link/Coff.zig
andrewrk: Neither is correct. The code needs to be improved to do what the todo comment says.

andrewrk:
Tests passed locally.

Issue #114, opened 2016-02-09
precedence table


PavelVozenilek:
Look at the cryptic table:

    x() x[] x.y
    !x -x ~x *x &x ?x %x %%x
    x{}
    * / %
    + - ++
    << >>
    &
    ^
    |
    == != < > <= >=
    &&
    ||
    ?? %%
    = *= /= %= += -= <<= >>= &= ^= |= &&= ||=

Why it is all needed? Who will find the time to learn it? I had not internalized C precedence after two decades. 

Language Pony does it [differently](http://tutorial.ponylang.org/expressions/infix-ops/#precedence):

> Pony takes a different approach and outlaws infix precedence. Any expression where more than one infix operator is used must use parentheses to remove the ambiguity. If you fail to do this the compiler will complain.

andrewrk:
Interesting idea. This is on the table since it adheres to the principle that being readable is more important than being writable.

emanuelpalm:
:+1:

andrewrk:
Some precedence is necessary, such as assignment, function call invocation, array index. Too much parens and the language turns into lisp.

PavelVozenilek:
Assignment, function call, array index are not infix operators. Could there be any ambiguity with them?

I can imagine having default precedence for basic arithmetic (+-/*), this is what little kids get drilled in schools, but IMO it is not worth the effort.

PavelVozenilek:
Also, this feature would make adding user defined operators into the language easier.

thejoshwolfe:
I think this would be a cool idea for some categories of operators, specifically the ones i can never remember the precedence for. e.g. `a && b || c` would be a compile error, or `a & b | c`. i don't think we need to go so far as to outlaw `a + b * c` though.

thejoshwolfe:
see also #29

phase:
My favorite quote from the Pony page is

> Most people will remember to do multiplication before addition, but what about left bit shifting versus bitwise and? Sometimes people misremember (or guess wrong) and that leads to bugs. Worse, those bugs are often very hard to spot.

I love the idea for specific parens. It makes the execution crystal clear.

thejoshwolfe:
The current operator precedence table is inspired by C. But I think Zig can do better than that. For example, `&` > `^` > `|` seems pretty silly. Even if you think `&` > `|` makes sense (because of the analogy to `*` and `+`), it's still pretty confusing that `^` is in the middle of them.

[...]

PR #7358, opened 2020-12-09
Fix operator precedence documentation

SpexGuy:
Thanks to @EleanorNB for pointing this out.

The docs had operator precedence backwards between prefix unary operators and postfix unary operators.
Things like `&x.?`, `-x.*`, `&T{}` all compile, but the previous order in the documentation would say that these require parentheses.

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -1743,8 +1743,8 @@ const B = error{Two};
       {#header_open|Precedence#}
       <pre>{#syntax#}x() x[] x.y
 a!b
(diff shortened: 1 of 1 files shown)


EleanorNB:
Every community needs a stickler 😉

andrewrk:
Thanks @EleanorNB and @SpexGuy!

andrewrk:
Related, note this old but still accepted proposal: #114

SpexGuy:
Yay my first PR!

PR #7368, opened 2020-12-09
stage2: MachO export trie cleanup

kubkon:
This PR cleans up the error message when there is not enough padding between the end of the load commands and the beginning of the `__text` section. Similarly to what @andrewrk has done for stage1, this PR also arch-gates code signature writing to only `aarch64` (it's obsolete on `x86_64` at the time of this writing) in stage2. I've also did a major refactor of the export trie generation routine, and added ability to parse a trie from some input byte stream. This will become useful when we'll either be patching an existing binary, or linking with object/dynamic library files.


andrewrk (approved):
Looks good!

Issue #7342, opened 2020-12-08
"zig cc" doesn't work with cgo (golang)


lu4p:
I would like to use "zig cc" for easily cross compiling cgo dependent go code, however the "drop-in" c compiler replacement doesn't work.

### Example cgo code

main.go
```go
package main

//int Add(int a, int b){
//    return a+b;
//}
import "C"
import "fmt"

func main() {
	a := C.int(10)
	b := C.int(20)
	c := C.Add(a, b)
	fmt.Println(c) // 30
}
```

### other compilers
```
go build main.go //build works fine (uses gcc)
CC="clang" go build main.go //build works fine (uses clang)
```

### zig cc
```
$ CC="zig cc" go build main.go
# runtime/cgo
info: Usage: zig [command] [options]

Commands:

  build            Build project from build.zig
  build-exe        Create executable from source or object files
  build-lib        Create library from source or object files
  build-obj        Create object from source or assembly
  cc               Use Zig as a drop-in C compiler
  c++              Use Zig as a drop-in C++ compiler
  env              Print lib path, std path, compiler id and version
  fmt              Parse file and render in canonical zig format
  init-exe         Initialize a `zig build` application in the cwd
  init-lib         Initialize a `zig build` library in the cwd
  libc             Display native libc paths file or validate one
  run              Create executable and run immediately
  translate-c      Convert C code to Zig code
  targets          List available compilation targets
  test             Create and run a test build
  version          Print version number and exit
  zen              Print zen of zig and exit

General Options:

  --help           Print command-specific usage

error: unknown command: -E
```

```
go version go1.15.5 linux/amd64
```

andrewrk:
So this looks like go is inserting flags *before* it passes `cc`. This will either need to be fixed in go, or you will have to work around this by creating a shell script like this:

```
#!/bin/sh
zig cc $@
```

And then putting *that* as your CC. This way zig gets invoked with `zig cc -E ...` instead of `zig -E ...`

andrewrk:
Downstream issue filed: https://github.com/golang/go/issues/43078

lu4p:
> ```
> #!/bin/sh
> zig cc $@
> ```

This indeed works, but another weird thing is, that I get these errors:
```
$ CC="zig-wrap" go build  main.go
# runtime/cgo
error: unable to create compilation: AccessDenied
# runtime/cgo
gcc_context.c:1:1: error: unable to build C object: AccessDenied
# runtime/cgo
[...]

PR #7377, opened 2020-12-10
CLI: explicit integration with cgo

andrewrk:
This is just a proof-of-concept. Maybe it could be nice for Go
developers, but also maybe we don't put additional complexity into the
zig codebase when CLI flags work just fine.

See the discussion in #7342

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -118,11 +118,6 @@ pub fn main() anyerror!void {
 const os_can_execve = std.builtin.os.tag != .windows;
 
 pub fn mainArgs(gpa: *Allocator, arena: *Allocator, args: []const []const u8) !void {
-    if (args.len <= 1) {
-        std.log.info("{}", .{usage});
-        fatal("expected command argument", .{});
-    }
-
     if (os_can_execve and std.os.getenvZ("ZIG_IS_DETECTING_LIBC_PATHS") != null) {
         // In this case we have accidentally invoked ourselves as "the system C compiler"
         // to figure out where libc is installed. This is essentially infinite recursion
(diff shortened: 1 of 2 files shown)


src/target.zig
```diff
@@ -343,3 +343,79 @@ pub fn is_libcpp_lib_name(target: std.Target, name: []const u8) bool {
...
+    } else if (std.mem.eql(u8, go_os, "windows")) {
+        os_tag = .windows;
+        abi = .gnu; // so we can provide libc
+    } else {
+        return error.UnrecognizedOperatingSystem;
+    }
+
+    var cpu_arch: ?std.Target.Cpu.Arch = undefined;
+    if (std.mem.eql(u8, go_arch, "386")) {
+        cpu_arch = null;
```
Vexu, line 387:
```suggestion
    if (std.mem.eql(u8, go_arch, "native")) {
        cpu_arch = null;
```

lu4p:
I think this would be great, to onboard members of the go community to the zig project.

src/target.zig
```diff
@@ -343,3 +343,79 @@ pub fn is_libcpp_lib_name(target: std.Target, name: []const u8) bool {
...
+        os_tag = .dragonfly;
+    } else if (std.mem.eql(u8, go_os, "freebsd")) {
+        os_tag = .freebsd;
+    } else if (std.mem.eql(u8, go_os, "illumos")) {
+        return error.UnsupportedOperatingSystem;
+    } else if (std.mem.eql(u8, go_os, "js")) {
+        return error.UnsupportedOperatingSystem;
+    } else if (std.mem.eql(u8, go_os, "linux")) {
+        os_tag = .linux;
+        abi = .musl; // go wants static linking on linux
```
lu4p, line 369: I think this is an ok default behaviour, but this should be configurable via a flag or env var.

LemonBoy:
I don't think that writing a shell script to perform the argument translation is _that_ hard.
On the other hand maintaining this set of extra switches and making sure they don't regress is much harder.

src/main.zig
```diff
@@ -154,6 +149,51 @@ pub fn mainArgs(gpa: *Allocator, arena: *Allocator, args: []const []const u8) !v
         }
     }
 
+    const is_cgo = std.os.getenvZ("ZIG_CGO") != null;
+    if (is_cgo) cgo: {
+        // Workaround for https://github.com/golang/go/issues/43078, here we fix the order of
+        // command line arguments. Dear Go developers please let the Zig project know when
+        // we can remove this!
```
komuw, line 156: https://go-review.googlesource.com/c/go/+/276412/ which is under review may fix this.

jonjohnsonjr:
> I don't think that writing a shell script to perform the argument translation is that hard.
> On the other hand maintaining this set of extra switches and making sure they don't regress is much harder.

I tend to agree, though I don't personally want to be responsible for maintaining this mapping (and it seems silly to have everyone individually reimplement it), but it's not clear where something like that _should_ live.

I'm not sure what the minimum bar is for creating a repository is under the ziglang org, but something like a github.com/ziglang/zig-cgo (or cgo-zig) that hosts the canonical translation layer would be a very nice thing to be able to reference. There seems to be precedent for this with docker-zig, zig.vim, vscode-zig, zig-mode, etc...

andrewrk:
The more I think about it, the more I think there should not be a translation layer at all - typically when you are creating a cross build, you have in mind the target that you want to specify. You will then have the task of choosing a `-target` / `-mcpu` flags for Zig and GOOS/GOARCH env vars for go. It's possible that the target you want to give to Zig can't even be described by GOOS/GOARCH. For example OS version ranges, CPU features, dynamic linker path, C ABI, glibc version. In this case you *must* manually choose the Zig target configuration. So I think we may actually be doing a disservice by leading users to configure the much more detailed Zig target configuration via the coarse-grained go env vars.

Based on this, and @LemonBoy's point, I'm going to close this. I do, however, want to solve the problem of having zig choose a more effective local cache directory. (See https://github.com/golang/go/issues/43078#issuecomment-741198473 for more details on that.)

Issue #7242, opened 2020-11-28
Minimum Supported Windows Version


marler8997:
Today I realized that `zig build` no longer works on Windows 7.  The standard library is now using features that were not introduced until Windows 8.   Microsoft has officially dropped Windows 7 support earlier this year (2020), however, according to some "googling", Windows 7 appears to hold a desktop share of 18%.  Because of issues with Windows 10, many users continue to use this older version of the OS.  I propose that because of it's apparent high market share, Zig continues to officially support Windows 7.

Windows 7 market share has been steadily declining over the years.  Windows 10 finally overtook it around the beginning of 2018.  Based on these numbers, Windows 7 will probably drop it's usage into the single digits sometime in 2022.  Based on these estimates, extending support until 2022 could be a reasonable cut-off date.

squeek502:
Just for information purposes, here's where Windows 7 stopped working: https://github.com/ziglang/zig/pull/5993#issuecomment-693126335

Mouvedia:
> […it shouldn't be too difficult to add a switch/if on Windows version and fallback to a different query value than FileNormalizedNameInformation. In fact, AFAIK ReactOS and Wine both use a generic NtQueryObject to query the path.](https://github.com/ziglang/zig/pull/5993#issuecomment-693209192)

If that's all it takes, it should be done. Since Windows 7 is not supported _yet_, it's not urgent though.
It's worth it.

andrewrk:
This is the kind of thing I mean by "non invasive". Even without explicit support, non invasive patches to widen support are welcome. They won't be tested by the CI though and will be maintained on best effort, but without mandatory testing.

Prince213:
Is Windows 7 support going to be added back?

11dim:
@andrewrk

Our customers tend to keep using their base 24x7 systems longer than Microsoft EOLs. They expect us to keep their systems running with our latest enhancements.

Supporting an older windows platform gives some *sort* of trust the end solution will work on new platforms under *certain* conditions. It also keeps Multiplatform developer's VM's small/fast. 

In our case we try to ship as much as possible 32-bit versions of our solutions to keep compatibility with working 32-bit third-party components as well.

Already created a small utility using zig 0.9 dev under windows 7 64-bit and Apple M1.

Thanks again for the zig language, I really think you will succeed in a better C!  Looking forward to keep using it !

AKDev21:
Hello,
i'm also getting this error please solve this error i want to use zig please

[...]

PR #5993, opened 2020-08-06
Implement std.os.windows.GetPathNameByHandle using NT routines only

kubkon:
This PR is part of #1840.

This PR proposes an in-house implementation of `std.os.windows.GetPathNameByHandle` routine which you'd normally link to via `kernelbase.dll`. In particular, the routine is implemented using `std.os.windows.ntdll.NtQueryInformationFile` with class flags `FileNormalizedNameInformation` and `FileVolumeNameInformation`. The former allows us to get the normalized (canonical) path to the resource, however, the volume name is *not* prepended. This is what we use the latter for. Hence, in effect we call `NtQueryInformationFile` twice. However, there is still one complication in the fact that, by default, NTOS returns NT volume names, e.g., `\Device\HarddiskVolume0` which needs to be converted to the more familiar DOS volume name such as `C:` or whatnot (an interesting fact, `C:` and `\DosDevices\C:` are simply symbolic link objects to `\Device\HarddiskVolume0`). To convert the NT volume name into DOS volume name, we issue `std.os.windows.DeviceIoControl` syscall with `IOCTL_MOUNTMGR_QUERY_POINTS` opcode, with the result of `FileVolumeNameInformation` as the input. Then, the mount manager returns all symbolic link objects that point at that volume, and we select the first available from the result pool as our DOS drive letter.

This PR took me quite a while to figure out as most of the API is not very well documented. The things got even more complicated since ReactOS doesn't provide `GetFinalPathNameByHandleW` yet, and Wine uses a slightly different (may I say dated approach?) plus Wine it seems somehow manages to obtain DOS drive letter upfront: [link for the interested](https://source.winehq.com/git/wine.git/blob/HEAD:/dlls/kernelbase/file.c#l1708). Finally, using `NtQueryInformationFile` with `FileNormalizedNameInformation` seems to agree with what I've reversed engineered using ProcMon on the actual `kernelbase.GetFinalPathNameByHandleW` on Windows.

Anyhow, lemme know what y'all reckon! Comments and suggestions most welcome!

PR #7379, opened 2020-12-10
Fix `std.os.windows.GetFinalPathNameByHandle` for os versions before `win10_rs4`

rohlem:
This PR fixes `std.os.windows.GetFinalPathNameByHandle` for older Windows versions, unblocking #7242 (although that might be an ongoing effort, so shouldn't be closed yet I think).

#5993 removed the codepath that called the `kernel32.dll` implementation. According to [some document I found](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/efbfe127-73ad-4140-9967-ec6500e66d5e), the new `.FileNormalizedNameInformation` query value was only added as recently as 2018 (not even 3 years ago), meaning we dropped support for more than just Windows 7.

Originally I just wanted to query the runtime OS version and reintroduce the old codepath, but then I tried the [prophesized `NtQueryObject` function](https://github.com/ziglang/zig/pull/5993#issuecomment-693209192), and that ended up working too. Note though that it's using an undocumented value (`ObjectNameInformation = 1` appears in `ntifs.h`, but not the documentation) and a `struct` that is named appropriately but also documented somewhere else. And yet, it still works.

_If_ we trust this new implementation, I would be inclined to maybe delete the other codepath (that needs two `ntdll` calls instead of just one) in favor of this one. (And maybe move the logic for finding DOS volume names into its own function.)
But now that I've worked on this for a day, I thought I would ask for some feedback, to make sure I'm not overlooking anything crucial.

Tested (see new tests in std/os/windows.zig) on my Windows 7 and an up-to-date Windows 10 machine. Things I tried with this implementation:
- NTFS junctions are resolved
- Symbolic links are resolved
- Hard links remain separate paths
- Returned paths > 500 characters work without issue

Also let me know if I should squish/reorganize commits in any way, I thought for documentation/discussion (and maybe unexpected troubleshooting) the commits with the kernel32 path were worth keeping for now.

### DIFF
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -955,6 +955,51 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
     return @bitCast(u64, result);
 }
 
+pub fn QueryObjectName(
+    handle: HANDLE,
+    out_buffer: []u16,
+) ![]u16 {
+    var full_buffer: [@sizeOf(OBJECT_NAME_INFORMATION) + PATH_MAX_WIDE * 2]u8 align(@alignOf(OBJECT_NAME_INFORMATION)) = undefined;
+    var info = @ptrCast(*OBJECT_NAME_INFORMATION, &full_buffer);
+    //buffer size is specified in bytes
+    const full_buffer_length = @intCast(ULONG, @sizeOf(OBJECT_NAME_INFORMATION) + std.math.min(PATH_MAX_WIDE, (out_buffer.len + 1) * 2));
+    //last argument would return the length required for full_buffer, not exposed here
+    const rc = ntdll.NtQueryObject(handle, .ObjectNameInformation, full_buffer[0..], full_buffer_length, null);
+    return switch (rc) {
+        .SUCCESS => if (@ptrCast(?[*]WCHAR, info.Name.Buffer)) |buffer| blk: {
+            //resulting string length is specified in bytes
+            const path_length_unterminated = @divExact(info.Name.Length, 2);
+            if (out_buffer.len < path_length_unterminated) {
+                return error.NameTooLong;
+            }
+            std.mem.copy(WCHAR, out_buffer[0..path_length_unterminated], buffer[0..path_length_unterminated :0]);
+            break :blk out_buffer[0..path_length_unterminated];
+        } else error.Unexpected,
+        .ACCESS_DENIED => error.AccessDenied,
+        .INVALID_HANDLE => error.InvalidHandle,
+        .BUFFER_OVERFLOW, .BUFFER_TOO_SMALL => error.NameTooLong,
+        //name_buffer.len >= @sizeOf(OBJECT_NAME_INFORMATION) holds statically
+        .INFO_LENGTH_MISMATCH => unreachable,
+        else => |e| unexpectedStatus(e),
+    };
+}
+test "QueryObjectName" {
+    if (comptime builtin.os.tag != .windows)
+        return;
+
+    //any file will do; canonicalization works on NTFS junctions and symlinks, hardlinks remain separate paths.
+    const file = try std.fs.openSelfExe(.{});
+    defer file.close();
+    //make this large enough for the test runner exe path
+    var out_buffer align(16) = std.mem.zeroes([1 << 10]u16);
+
+    var result_path = try QueryObjectName(file.handle, out_buffer[0..]);
+    //insufficient size
+    std.testing.expectError(error.NameTooLong, QueryObjectName(file.handle, out_buffer[0 .. result_path.len - 1]));
+    //exactly-sufficient size
+    _ = try QueryObjectName(file.handle, out_buffer[0..result_path.len]);
+}
+
 pub const GetFinalPathNameByHandleError = error {
         BadPathName,
         FileNotFound,
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -1145,6 +1145,30 @@ pub fn GetFinalPathNameByHandle(
     }
 }
 
+test "GetFinalPathNameByHandle" {
+    if (comptime builtin.os.tag != .windows)
+        return;
+
+    //any file will do
+    const file = try std.fs.openSelfExe(.{});
+    defer file.close();
+    const handle = file.handle;
+    //make this large enough for the test runner exe path
+    var buffer = std.mem.zeroes([1 << 10]u16);
+
+    //check with sufficient size
+    const nt_length = (try GetFinalPathNameByHandle(handle, .{ .volume_name = .Nt }, buffer[0..])).len;
+    const dos_length = (try GetFinalPathNameByHandle(handle, .{ .volume_name = .Dos }, buffer[0..])).len;
+
+    //check with insufficient size
+    std.testing.expectError(error.NameTooLong, GetFinalPathNameByHandle(handle, .{ .volume_name = .Nt }, buffer[0 .. nt_length - 1]));
+    std.testing.expectError(error.NameTooLong, GetFinalPathNameByHandle(handle, .{ .volume_name = .Dos }, buffer[0 .. dos_length - 1]));
+
+    //check with exactly-sufficient size
+    _ = try GetFinalPathNameByHandle(handle, .{ .volume_name = .Nt }, buffer[0..nt_length]);
+    _ = try GetFinalPathNameByHandle(handle, .{ .volume_name = .Dos }, buffer[0..dos_length]);
+}
+
 pub const QueryInformationFileError = error{Unexpected};
 
 pub fn QueryInformationFile(
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -28,10 +28,6 @@ pub const gdi32 = @import("windows/gdi32.zig");
 
 pub usingnamespace @import("windows/bits.zig");
 
-//version detection
-const version = std.zig.system.windows;
-const WindowsVersion = version.WindowsVersion;
-
 pub const self_process_handle = @intToPtr(HANDLE, maxInt(usize));
 
 pub const OpenError = error{
@@ -1033,65 +1029,35 @@ pub fn GetFinalPathNameByHandle(
     out_buffer: []u16,
 ) GetFinalPathNameByHandleError![]u16 {
     var path_buffer: [math.max(@sizeOf(FILE_NAME_INFORMATION), @sizeOf(OBJECT_NAME_INFORMATION)) + PATH_MAX_WIDE * 2]u8 align(@alignOf(FILE_NAME_INFORMATION)) = undefined;
-    var volume_buffer: [@sizeOf(FILE_NAME_INFORMATION) + MAX_PATH]u8 align(@alignOf(FILE_NAME_INFORMATION)) = undefined; // MAX_PATH bytes should be enough since it's Windows-defined name
+    const final_path = QueryObjectName(hFile, mem.bytesAsSlice(u16, &path_buffer)) catch |err| switch (err) {
+        // we assume InvalidHandle is close enough to FileNotFound in semantics
+        // to not further complicate the error set
+        error.InvalidHandle => return error.FileNotFound,
+        else => |e| return e,
+    };
 
-    var file_name_u16: []const u16 = undefined;
-    var volume_name_u16: []const u16 = undefined;
-    if ((comptime (std.builtin.os.version_range.windows.isAtLeast(WindowsVersion.win10_rs4) != true)) and !version.detectRuntimeVersion().isAtLeast(WindowsVersion.win10_rs4)) {
-        const final_path = QueryObjectName(hFile, mem.bytesAsSlice(u16, &path_buffer)) catch |err| return switch (err) {
-            // we assume InvalidHandle is close enough to FileNotFound in semantics
-            // to not further complicate the error set
-            error.InvalidHandle => error.FileNotFound,
-            else => |e| e,
-        };
-
-        if (fmt.volume_name == .Nt) {
+    switch (fmt.volume_name) {
+        .Nt => {
+            // the returned path is already in .Nt format
             if (out_buffer.len < final_path.len) {
                 return error.NameTooLong;
             }
             mem.copy(u16, out_buffer, final_path);
             return out_buffer[0..final_path.len];
-        }
-
-        //otherwise we need to parse the string for volume path for the .Dos logic below to work
-        const expected_prefix = std.unicode.utf8ToUtf16LeStringLiteral("\\Device\\");
-
-        // TODO find out if a path can start with something besides `\Device\<volume name>`,
-        // and if we need to handle it differently
-        // (i.e. how to determine the start and end of the volume name in that case)
-        if (!mem.eql(u16, expected_prefix, final_path[0..expected_prefix.len])) return error.Unexpected;
-
-        const index = mem.indexOfPos(u16, final_path, expected_prefix.len, &[_]u16{'\\'}) orelse unreachable;
-        volume_name_u16 = final_path[0..index];
-        file_name_u16 = final_path[index..];
-
-        //fallthrough for fmt.volume_name != .Nt
-    } else {
-        // Get normalized path; doesn't include volume name though.
-        try QueryInformationFile(hFile, .FileNormalizedNameInformation, &path_buffer);
-        const file_name = @ptrCast(*const FILE_NAME_INFORMATION, &path_buffer);
-        file_name_u16 = @ptrCast([*]const u16, &file_name.FileName)[0..@divExact(file_name.FileNameLength, 2)];
-
-        // Get NT volume name.
-        try QueryInformationFile(hFile, .FileVolumeNameInformation, &volume_buffer);
-        const volume_name_info = @ptrCast(*const FILE_NAME_INFORMATION, &volume_buffer);
-        volume_name_u16 = @ptrCast([*]const u16, &volume_name_info.FileName)[0..@divExact(volume_name_info.FileNameLength, 2)];
-
-        if (fmt.volume_name == .Nt) {
-            // Nothing to do, we simply copy the bytes to the user-provided buffer.
-            if (out_buffer.len < volume_name_u16.len + file_name_u16.len) return error.NameTooLong;
-
-            mem.copy(u16, out_buffer, volume_name_u16);
-            mem.copy(u16, out_buffer[volume_name_u16.len..], file_name_u16);
-
-            return out_buffer[0 .. volume_name_u16.len + file_name_u16.len];
-        }
-        //fallthrough for fmt.volume_name != .Nt
-    }
-
-    switch (fmt.volume_name) {
-        .Nt => unreachable, //handled above
+        },
         .Dos => {
+            // parse the string to separate volume path from file path
+            const expected_prefix = std.unicode.utf8ToUtf16LeStringLiteral("\\Device\\");
+
+            // TODO find out if a path can start with something besides `\Device\<volume name>`,
+            // and if we need to handle it differently
+            // (i.e. how to determine the start and end of the volume name in that case)
+            if (!mem.eql(u16, expected_prefix, final_path[0..expected_prefix.len])) return error.Unexpected;
+
+            const file_path_begin_index = mem.indexOfPos(u16, final_path, expected_prefix.len, &[_]u16{'\\'}) orelse unreachable;
+            const volume_name_u16 = final_path[0..file_path_begin_index];
+            const file_name_u16 = final_path[file_path_begin_index..];
+
             // Get DOS volume name. DOS volume names are actually symbolic link objects to the
             // actual NT volume. For example:
             // (NT) \Device\HarddiskVolume4 => (DOS) \DosDevices\C: == (DOS) C:
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -29,7 +29,8 @@ pub const gdi32 = @import("windows/gdi32.zig");
 pub usingnamespace @import("windows/bits.zig");
 
 //version detection
-usingnamespace std.zig.system.windows;
+const version = std.zig.system.windows;
+const WindowsVersion = version.WindowsVersion;
 
 pub const self_process_handle = @intToPtr(HANDLE, maxInt(usize));
 
@@ -962,42 +963,44 @@ pub fn QueryObjectName(
     var full_buffer: [@sizeOf(OBJECT_NAME_INFORMATION) + PATH_MAX_WIDE * 2]u8 align(@alignOf(OBJECT_NAME_INFORMATION)) = undefined;
     var info = @ptrCast(*OBJECT_NAME_INFORMATION, &full_buffer);
     //buffer size is specified in bytes
-    const full_buffer_length = @intCast(ULONG, @sizeOf(OBJECT_NAME_INFORMATION) + std.math.min(PATH_MAX_WIDE, (out_buffer.len + 1) * 2));
     //last argument would return the length required for full_buffer, not exposed here
-    const rc = ntdll.NtQueryObject(handle, .ObjectNameInformation, full_buffer[0..], full_buffer_length, null);
-    return switch (rc) {
-        .SUCCESS => if (@ptrCast(?[*]WCHAR, info.Name.Buffer)) |buffer| blk: {
+    const rc = ntdll.NtQueryObject(handle, .ObjectNameInformation, &full_buffer, full_buffer.len, null);
+    switch (rc) {
+        .SUCCESS => {
+            // info.Name.Buffer from ObQueryNameString is documented to be null (and MaximumLength == 0)
+            // if the object was "unnamed", not sure if this can happen for file handles
+            if (info.Name.MaximumLength == 0) return error.Unexpected;
             //resulting string length is specified in bytes
             const path_length_unterminated = @divExact(info.Name.Length, 2);
             if (out_buffer.len < path_length_unterminated) {
                 return error.NameTooLong;
             }
-            std.mem.copy(WCHAR, out_buffer[0..path_length_unterminated], buffer[0..path_length_unterminated :0]);
-            break :blk out_buffer[0..path_length_unterminated];
-        } else error.Unexpected,
-        .ACCESS_DENIED => error.AccessDenied,
-        .INVALID_HANDLE => error.InvalidHandle,
-        .BUFFER_OVERFLOW, .BUFFER_TOO_SMALL => error.NameTooLong,
+            mem.copy(WCHAR, out_buffer[0..path_length_unterminated], info.Name.Buffer[0..path_length_unterminated]);
+            return out_buffer[0..path_length_unterminated];
+        },
+        .ACCESS_DENIED => return error.AccessDenied,
+        .INVALID_HANDLE => return error.InvalidHandle,
+        .BUFFER_OVERFLOW, .BUFFER_TOO_SMALL => return error.NameTooLong,
         //name_buffer.len >= @sizeOf(OBJECT_NAME_INFORMATION) holds statically
         .INFO_LENGTH_MISMATCH => unreachable,
-        else => |e| unexpectedStatus(e),
-    };
+        else => |e| return unexpectedStatus(e),
+    }
 }
 test "QueryObjectName" {
     if (comptime builtin.os.tag != .windows)
         return;
 
     //any file will do; canonicalization works on NTFS junctions and symlinks, hardlinks remain separate paths.
-    const file = try std.fs.openSelfExe(.{});
-    defer file.close();
-    //make this large enough for the test runner exe path
-    var out_buffer align(16) = std.mem.zeroes([1 << 10]u16);
+    var tmp = std.testing.tmpDir(.{});
+    defer tmp.cleanup();
+    const handle = tmp.dir.fd;
+    var out_buffer: [PATH_MAX_WIDE]u16 = undefined;
 
-    var result_path = try QueryObjectName(file.handle, out_buffer[0..]);
+    var result_path = try QueryObjectName(handle, &out_buffer);
     //insufficient size
-    std.testing.expectError(error.NameTooLong, QueryObjectName(file.handle, out_buffer[0 .. result_path.len - 1]));
+    std.testing.expectError(error.NameTooLong, QueryObjectName(handle, out_buffer[0 .. result_path.len - 1]));
     //exactly-sufficient size
-    _ = try QueryObjectName(file.handle, out_buffer[0..result_path.len]);
+    _ = try QueryObjectName(handle, out_buffer[0..result_path.len]);
 }
 
 pub const GetFinalPathNameByHandleError = error{
@@ -1029,16 +1032,16 @@ pub fn GetFinalPathNameByHandle(
     fmt: GetFinalPathNameByHandleFormat,
     out_buffer: []u16,
 ) GetFinalPathNameByHandleError![]u16 {
-    var path_buffer: [std.math.max(@sizeOf(FILE_NAME_INFORMATION), @sizeOf(OBJECT_NAME_INFORMATION)) + PATH_MAX_WIDE * 2]u8 align(@alignOf(FILE_NAME_INFORMATION)) = undefined;
+    var path_buffer: [math.max(@sizeOf(FILE_NAME_INFORMATION), @sizeOf(OBJECT_NAME_INFORMATION)) + PATH_MAX_WIDE * 2]u8 align(@alignOf(FILE_NAME_INFORMATION)) = undefined;
     var volume_buffer: [@sizeOf(FILE_NAME_INFORMATION) + MAX_PATH]u8 align(@alignOf(FILE_NAME_INFORMATION)) = undefined; // MAX_PATH bytes should be enough since it's Windows-defined name
 
     var file_name_u16: []const u16 = undefined;
     var volume_name_u16: []const u16 = undefined;
-    if ((comptime (targetVersionIsAtLeast(WindowsVersion.win10_rs4) != true)) //need explicit comptime, because error returns affect return type
-        and !runtimeVersionIsAtLeast(WindowsVersion.win10_rs4))
-    {
-        const final_path = QueryObjectName(hFile, std.mem.bytesAsSlice(u16, path_buffer[0..])) catch |err| return switch (err) {
-            error.InvalidHandle => error.FileNotFound, //close enough?
+    if ((comptime (std.builtin.os.version_range.windows.isAtLeast(WindowsVersion.win10_rs4) != true)) and !version.detectRuntimeVersion().isAtLeast(WindowsVersion.win10_rs4)) {
+        const final_path = QueryObjectName(hFile, mem.bytesAsSlice(u16, &path_buffer)) catch |err| return switch (err) {
+            // we assume InvalidHandle is close enough to FileNotFound in semantics
+            // to not further complicate the error set
+            error.InvalidHandle => error.FileNotFound,
             else => |e| e,
         };
 
@@ -1046,39 +1049,40 @@ pub fn GetFinalPathNameByHandle(
             if (out_buffer.len < final_path.len) {
                 return error.NameTooLong;
             }
-            std.mem.copy(u16, out_buffer[0..], final_path[0..]);
-            return final_path; //we can directly return the slice we received
+            mem.copy(u16, out_buffer, final_path);
+            return out_buffer[0..final_path.len];
         }
 
         //otherwise we need to parse the string for volume path for the .Dos logic below to work
         const expected_prefix = std.unicode.utf8ToUtf16LeStringLiteral("\\Device\\");
-        if (!std.mem.eql(u16, expected_prefix, final_path[0..expected_prefix.len])) {
-            //TODO find out if this can occur, and if we need to handle it differently
-            //(i.e. how to determine the end of a volume name)
-            return error.BadPathName;
-        }
-        const index = std.mem.indexOfPos(u16, final_path, expected_prefix.len, &[_]u16{'\\'}) orelse unreachable;
+
+        // TODO find out if a path can start with something besides `\Device\<volume name>`,
+        // and if we need to handle it differently
+        // (i.e. how to determine the start and end of the volume name in that case)
+        if (!mem.eql(u16, expected_prefix, final_path[0..expected_prefix.len])) return error.Unexpected;
+
+        const index = mem.indexOfPos(u16, final_path, expected_prefix.len, &[_]u16{'\\'}) orelse unreachable;
         volume_name_u16 = final_path[0..index];
         file_name_u16 = final_path[index..];
 
         //fallthrough for fmt.volume_name != .Nt
     } else {
         // Get normalized path; doesn't include volume name though.
-        try QueryInformationFile(hFile, .FileNormalizedNameInformation, path_buffer[0..]);
-        const file_name = @ptrCast(*const FILE_NAME_INFORMATION, &path_buffer[0]);
-        file_name_u16 = @ptrCast([*]const u16, &file_name.FileName[0])[0..@divExact(file_name.FileNameLength, 2)];
+        try QueryInformationFile(hFile, .FileNormalizedNameInformation, &path_buffer);
+        const file_name = @ptrCast(*const FILE_NAME_INFORMATION, &path_buffer);
+        file_name_u16 = @ptrCast([*]const u16, &file_name.FileName)[0..@divExact(file_name.FileNameLength, 2)];
 
         // Get NT volume name.
-        try QueryInformationFile(hFile, .FileVolumeNameInformation, volume_buffer[0..]);
-        const volume_name_info = @ptrCast(*const FILE_NAME_INFORMATION, &volume_buffer[0]);
-        volume_name_u16 = @ptrCast([*]const u16, &volume_name_info.FileName[0])[0..@divExact(volume_name_info.FileNameLength, 2)];
+        try QueryInformationFile(hFile, .FileVolumeNameInformation, &volume_buffer);
+        const volume_name_info = @ptrCast(*const FILE_NAME_INFORMATION, &volume_buffer);
+        volume_name_u16 = @ptrCast([*]const u16, &volume_name_info.FileName)[0..@divExact(volume_name_info.FileNameLength, 2)];
 
         if (fmt.volume_name == .Nt) {
             // Nothing to do, we simply copy the bytes to the user-provided buffer.
             if (out_buffer.len < volume_name_u16.len + file_name_u16.len) return error.NameTooLong;
 
-            std.mem.copy(u16, out_buffer[0..], volume_name_u16);
-            std.mem.copy(u16, out_buffer[volume_name_u16.len..], file_name_u16);
+            mem.copy(u16, out_buffer, volume_name_u16);
+            mem.copy(u16, out_buffer[volume_name_u16.len..], file_name_u16);
 
             return out_buffer[0 .. volume_name_u16.len + file_name_u16.len];
         }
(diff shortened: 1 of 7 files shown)


lib/std/os/windows.zig
```diff
@@ -952,12 +952,64 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
 }
 
-pub const GetFinalPathNameByHandleError = error{
-    BadPathName,
-    FileNotFound,
-    NameTooLong,
-    Unexpected,
-};
+pub fn QueryObjectName(
+    Handle: HANDLE,
```
LemonBoy, line 956: Lowercase parameter name.

lib/std/os/windows.zig
```diff
@@ -952,12 +952,64 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
+    out_buffer: []u16,
+    ReturnLength: ?PULONG, //returned in bytes
```
LemonBoy, line 958: Lowercase parameter name.

lib/std/os/windows.zig
```diff
@@ -952,12 +952,64 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
+    const begin_addr = @ptrToInt(out_buffer.ptr);
+    const aligned_begin_addr = std.mem.alignForward(begin_addr, alignment);
+    const alignment_step = aligned_begin_addr - begin_addr;
+    const aligned_raw_buffer = @alignCast(alignment, std.mem.sliceAsBytes(out_buffer)[alignment_step..out_buffer.len*@sizeOf(u16)]);
+    if(aligned_raw_buffer.len < @sizeOf(OBJECT_NAME_INFORMATION)) return error.BufferTooSmall; //this would result in .INFO_LENGTH_MISMATCH
+    
+    var info = @ptrCast(*OBJECT_NAME_INFORMATION, aligned_raw_buffer);
+    info.* = undefined; //all of the fields we read will be overwritten, and the result string is placed in the same buffer after the struct
+    const out_buffer_length = std.math.cast(ULONG, aligned_raw_buffer.len) catch |e| std.math.maxInt(u32); //we'll just use as much of it as we can
+    const rc = ntdll.NtQueryObject(Handle, .ObjectNameInformation, out_buffer.ptr, out_buffer_length, ReturnLength); //buffer size is specified in u16...
```
LemonBoy, line 970: The code is extremely hairy, it takes an output buffer `out_buffer` where the pathname is stored _and_ uses it to allocate a `OBJECT_NAME_INFORMATION` too.

I'd simply allocate a buffer that's hopefully large enough (path_max plus the `OBJECT_NAME_INFORMATION` overhead) and copy the result over.

lib/std/os/windows.zig
```diff
@@ -952,12 +952,64 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
+    switch(rc){
```
LemonBoy, line 971: The missing spaces hint you're not using `zig fmt`, please run it on all the files this PR is touching.

lib/std/os/windows.zig
```diff
@@ -952,12 +952,64 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
+        .BUFFER_OVERFLOW, .BUFFER_TOO_SMALL => return error.NameTooLong,
+        .INFO_LENGTH_MISMATCH => unreachable, //aligned_raw_buffer.len < @sizeOf(OBJECT_NAME_INFORMATION) already checked previously
+        else => |e| return unexpectedStatus(e),
+    }
+}
+test "QueryObjectName" {
+    if (comptime builtin.os.tag != .windows)
+        return;
+    
+    const file: std.fs.File = try std.fs.openSelfExe(.{}); //any file will do; canonicalization works on NTFS junctions and symlinks, hardlinks remain separate paths.
```
LemonBoy, line 985: Redundant types are redundant.
And please keep the comments in a line above the one they refer to, not everyone has a huge monitor.

```suggestion
    //any file will do; canonicalization works on NTFS junctions and symlinks, hardlinks remain separate paths.
    const file = try std.fs.openSelfExe(.{}); 
```

lib/std/os/windows.zig
```diff
@@ -952,12 +952,64 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
+        FileNotFound,
+        NameTooLong,
+        Unexpected,
+    }
+    || if((comptime builtin.os.tag != .windows) or (targetVersionIsAtLeast(WindowsVersion.win10_rs4) == true))
+        error {}
+    else
+        error {
+            AccessDenied,
+            BufferTooSmall,
```
LemonBoy, line 1009: `NameTooLong` and `BufferTooSmall` ?

lib/std/os/windows.zig
```diff
@@ -952,12 +952,64 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
+            InvalidHandle,
+            Unnamed,
```
LemonBoy, line 1011: Unnamed ?

lib/std/os/windows.zig
```diff
@@ -980,31 +1032,54 @@ pub fn GetFinalPathNameByHandle(
...
-    const volume_name = @ptrCast(*const FILE_NAME_INFORMATION, &volume_buffer[0]);
+        if(fmt.volume_name == .Nt){
+            return final_path; //we can directly return the slice we received
+        }
 
-    switch (fmt.volume_name) {
-        .Nt => {
+        //otherwise we need to parse the string for volume path for the .Dos logic below to work
+        const expected_prefix = std.unicode.utf8ToUtf16LeStringLiteral("\\Device\\");
+        std.debug.assert(std.mem.eql(u16, expected_prefix, final_path[0..expected_prefix.len]));
```
LemonBoy, line 1051: This is elided for non-debug builds, is this intentional?

lib/std/os/windows.zig
```diff
@@ -980,31 +1032,54 @@ pub fn GetFinalPathNameByHandle(
...
+        const index = std.mem.indexOfPos(u16, final_path, expected_prefix.len, &[_]u16{'\\'}) orelse unreachable;
+        volume_name_u16 = final_path[0..index];
+        file_name_u16 = final_path[index..];
```
LemonBoy, line 1054: This gets the `\`, is it intentional?

lib/std/os/windows.zig
```diff
@@ -1091,6 +1166,43 @@ pub fn GetFinalPathNameByHandle(
...
 }
 
+test "GetFinalPathNameByHandle" {
+    if (comptime builtin.os.tag != .windows)
+        return;
+    
+    const S = struct {fn check(file: std.fs.File, buffer: anytype, fmt: std.os.windows.GetFinalPathNameByHandleFormat) !usize {
+        const real = try std.os.windows.GetFinalPathNameByHandle(file.handle, fmt, buffer[0..]);
+        const offset = @divExact((@ptrToInt(real.ptr) - @ptrToInt(@ptrCast([*]u16, buffer))), 2);
+        return real.len + offset;
```
LemonBoy, line 1176: What's this `offset` for? Why do you care about what's before `real.ptr`?

lib/std/os/windows.zig
```diff
@@ -1091,6 +1166,43 @@ pub fn GetFinalPathNameByHandle(
...
+    const file: std.fs.File = try std.fs.openSelfExe(.{}); //any file would do
+    defer file.close();
+    var buffer = std.mem.zeroes([1<<10]u16); //make this big enough for the test runner exe path
+
+    //check with sufficient size
+    const nt_length = try S.check(file, buffer[0..], .{.volume_name = .Nt});
+    const dos_length = try S.check(file, buffer[0..], .{.volume_name = .Dos});
+
+    //check with size insufficient by more than 1 character: always errors
+    std.testing.expect(null == S.check(file, buffer[0..nt_length-1], .{.volume_name = .Nt})
```
LemonBoy, line 1188:
```suggestion
    std.testing.expectEqual(null, S.check(file, buffer[0..nt_length-1], .{.volume_name = .Nt})
```

Ditto everywhere else, this way if the check fails it'll show you the faulty value.

lib/std/os/windows.zig
```diff
@@ -1091,6 +1166,43 @@ pub fn GetFinalPathNameByHandle(
...
+        catch |e| switch(e) {error.NameTooLong => null, else => unreachable});
+    std.testing.expect(null == S.check(file, buffer[0..dos_length-1], .{.volume_name = .Dos})
+        catch |e| switch(e) {error.NameTooLong => null, else => unreachable});
+
+    //check with size insufficient for null terminator: errors only on NtQueryObject path
+    const queryObjectPath = !std.os.windows.runtimeVersionIsAtLeast(std.os.windows.WindowsVersion.win10_rs4);
+    std.testing.expect(queryObjectPath == (null == S.check(file, buffer[0..nt_length], .{.volume_name = .Nt})
+        catch |e| switch(e) {error.NameTooLong => null, else => unreachable}));
```
LemonBoy, line 1196: There's `testing.expectError` for this.

lib/std/os/windows.zig
```diff
@@ -1601,6 +1713,67 @@ inline fn MAKELANGID(p: c_ushort, s: c_ushort) LANGID {
...
+
+/// Returns the highest known WindowsVersion deduced from reported runtime information.
+/// Discards information about in-between versions we don't differentiate.
+pub fn detectRuntimeVersion() WindowsVersion {
+    var version_info: std.os.windows.RTL_OSVERSIONINFOW = undefined;
+    version_info.dwOSVersionInfoSize = @sizeOf(@TypeOf(version_info));
+
+    switch (std.os.windows.ntdll.RtlGetVersion(&version_info)) {
+        .SUCCESS => {},
+        else => |e| { //only return value STATUS_SUCCESS is documented
```
LemonBoy, line 1726:
```suggestion
        else => unreachable
```

This API cannot fail according to the docs.

lib/std/os/windows.zig
```diff
@@ -1601,6 +1713,67 @@ inline fn MAKELANGID(p: c_ushort, s: c_ushort) LANGID {
...
+    const target_min = @enumToInt(version_range.min);
+    const target_max = @enumToInt(version_range.max);
+    return if(target_max < requested) false
+        else if(target_min >= requested) true
+        else null;
+}
+
+/// Returns whether the runtime os version is at least as new as the argument.
+pub fn runtimeVersionIsAtLeast(requested_version: WindowsVersion) bool {
+    return if(targetVersionIsAtLeast(requested_version)) |fully| fully
```
LemonBoy, line 1773:
```suggestion
    return if(targetVersionIsAtLeast(requested_version)) orelse
        @enumToInt(detectRuntimeVersion()) >= @enumToInt(requested_version);
```

lib/std/os/windows.zig
rohlem: That possibility is mentioned [here](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-obquerynamestring) (`if the given object is unnamed`), which I found used in the implementation of ReactOS's `NtQueryObject`. I don't know when it would be triggered though, might be for other "objects" than files? I'll change that to unreachable.

rohlem:
Thanks for the quick review, I might only get to it tomorrow though.

lib/std/os/windows.zig
rohlem: When I tested it seemed to give the same results as the other codepath, so I would say yes. Also the docs say absolute paths start with `\`, while relative paths don't. (`C:test\abc.txt` is a relative path.)

lib/std/os/windows.zig
kubkon: Hmm, could we not coerce `Unnamed` as simply `Unexpected`? Or put it another way, what value will conveying that the object is unnamed give to the user?

lib/std/os/windows.zig
```diff
@@ -952,12 +952,64 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
+    _ = try QueryObjectName(file.handle, out_buffer[0..required_length+1], null);
+}
+
+pub const GetFinalPathNameByHandleError = error {
+        BadPathName,
+        FileNotFound,
+        NameTooLong,
+        Unexpected,
+    }
+    || if((comptime builtin.os.tag != .windows) or (targetVersionIsAtLeast(WindowsVersion.win10_rs4) == true))
```
kubkon, line 1004: Given @LemonBoy's comments below, if we collapse a lot of these error values into the ones already existing, is it necessary to comptime check for Windows version, or would it be more readable to slightly enlarge the return error set instead? We already do that in a few places in `std.os` mind where some error values are WASI or Windows-only but they're part of the same return set.

lib/std/os/windows.zig
```diff
@@ -980,31 +1032,54 @@ pub fn GetFinalPathNameByHandle(
...
+        try QueryInformationFile(hFile, .FileNormalizedNameInformation, path_buffer[0..]);
+        const file_name = @ptrCast(*const FILE_NAME_INFORMATION, &path_buffer[0]);
+        file_name_u16 = @ptrCast([*]const u16, &file_name.FileName[0])[0 .. @divExact(file_name.FileNameLength, 2)];
+
+        // Get NT volume name.
+        try QueryInformationFile(hFile, .FileVolumeNameInformation, volume_buffer[0..]);
+        const volume_name_info = @ptrCast(*const FILE_NAME_INFORMATION, &volume_buffer[0]);
+        volume_name_u16 = @ptrCast([*]const u16, &volume_name_info.FileName[0])[0 .. @divExact(volume_name_info.FileNameLength, 2)];
+
+        if(fmt.volume_name == .Nt){
```
kubkon, line 1068: Much like the codepath below, I think a `switch` statement here with an `unreachable` for `.Dos` would convey the meaning a little better than an `if` statement without an `else`.

lib/std/os/windows.zig
```diff
@@ -1601,6 +1713,67 @@ inline fn MAKELANGID(p: c_ushort, s: c_ushort) LANGID {
...
+    } else 0;
+
+    const version: u32 = @as(u32, os_ver) << 16 | @as(u16, sp_ver) << 8 | sub_ver;
+
+    return @intToEnum(WindowsVersion, version);
+}
+
+/// Returns whether the target os versions are uniformly at least as new as the argument:
+/// true/false if this holds for the entire target range, null if it only holds for some.
+pub fn targetVersionIsAtLeast(requested_version: WindowsVersion) ?bool {
```
kubkon, line 1761: Should these not belong in `system.zig`?

lib/std/os/windows.zig
```diff
@@ -1601,6 +1713,67 @@ inline fn MAKELANGID(p: c_ushort, s: c_ushort) LANGID {
     return (s << 10) | p;
 }
 
+pub const WindowsVersion = std.Target.Os.WindowsVersion;
+
+/// Returns the highest known WindowsVersion deduced from reported runtime information.
+/// Discards information about in-between versions we don't differentiate.
+pub fn detectRuntimeVersion() WindowsVersion {
```
kubkon, line 1720: Is this call really necessary? We already call `std.os.windows.ntdll.RtlGetVersion` in `std.zig.system` and in this line here in particular: [std/zig/system.zig#L229](https://github.com/ziglang/zig/blob/master/lib/std/zig/system.zig#L229). Also, for future reference, any bit of code that works out the target/crosstarget version should go in `system.zig`.

lib/std/zig/system.zig
```diff
@@ -223,44 +223,9 @@ pub const NativeTargetInfo = struct {
...
-                                last_idx = i;
-                        }
-                        break :subver @truncate(u8, last_idx);
-                    } else 0;
-
-                    const version: u32 = @as(u32, os_ver) << 16 | @as(u32, sp_ver) << 8 | sub_ver;
-
-                    os.version_range.windows.max = @intToEnum(Target.Os.WindowsVersion, version);
-                    os.version_range.windows.min = @intToEnum(Target.Os.WindowsVersion, version);
+                    const detected_version = std.os.windows.detectRuntimeVersion();
```
kubkon, line 226: Oh I see now. This function should still be part of the `system.zig` though. I suggest putting it in `lib/std/zig/system/windows.zig` much like we have macOS specific target detection in `lib/std/zig/system/macos.zig`. @andrewrk's idea which I agree with is to have all version detection functionality in solely in the `system` module and nowhere else.

marler8997:
Just wanted to say thank you to @rohlem for showing windows some love, it desperately needs it.

lib/std/os/windows.zig
rohlem: Well, it generates `unreachable`. I don't know enough about NT paths to judge whether/where this can happen, and what to do if it does.
I suppose returning an error is safer? And I think `error.Unexpected`, since it's caused by the system and `error.BadPathName` sounds like a caller error?

lib/std/os/windows.zig
rohlem: `.Dos` is not `unreachable`, it falls through to the rest of the function. Or are you suggesting this?
```zig
switch(fmt){
  .Nt => {...},
  .Dos => {}//fallthrough
}
```
?

rohlem:
Rebased onto master and integrated all review comments, sorry it took so long.

lib/std/os/windows.zig
```diff
@@ -980,31 +1029,64 @@ pub fn GetFinalPathNameByHandle(
...
+            return final_path; //we can directly return the slice we received
+        }
 
-    switch (fmt.volume_name) {
-        .Nt => {
-            // Nothing to do, we simply copy the bytes to the user-provided buffer.
-            const volume_name_u16 = @ptrCast([*]const u16, &volume_name.FileName[0])[0 .. volume_name.FileNameLength / 2];
+        //otherwise we need to parse the string for volume path for the .Dos logic below to work
+        const expected_prefix = std.unicode.utf8ToUtf16LeStringLiteral("\\Device\\");
+        if (!std.mem.eql(u16, expected_prefix, final_path[0..expected_prefix.len])) {
```
LemonBoy, line 1055: Given this is a precondition we expect the kernel to respect, an assert is fine.

Please use `mem` instead of `std.mem`!

lib/std/os/windows.zig
```diff
@@ -980,31 +1029,64 @@ pub fn GetFinalPathNameByHandle(
...
+            //TODO find out if this can occur, and if we need to handle it differently
+            //(i.e. how to determine the end of a volume name)
+            return error.BadPathName;
+        }
+        const index = std.mem.indexOfPos(u16, final_path, expected_prefix.len, &[_]u16{'\\'}) orelse unreachable;
+        volume_name_u16 = final_path[0..index];
+        file_name_u16 = final_path[index..];
+
+        //fallthrough for fmt.volume_name != .Nt
```
LemonBoy, line 1064: fallthrough? The other case is handled in the other branch.

lib/std/os/windows.zig
```diff
@@ -980,31 +1029,64 @@ pub fn GetFinalPathNameByHandle(
...
+        try QueryInformationFile(hFile, .FileNormalizedNameInformation, path_buffer[0..]);
+        const file_name = @ptrCast(*const FILE_NAME_INFORMATION, &path_buffer[0]);
+        file_name_u16 = @ptrCast([*]const u16, &file_name.FileName[0])[0..@divExact(file_name.FileNameLength, 2)];
+
+        // Get NT volume name.
+        try QueryInformationFile(hFile, .FileVolumeNameInformation, volume_buffer[0..]);
+        const volume_name_info = @ptrCast(*const FILE_NAME_INFORMATION, &volume_buffer[0]);
+        volume_name_u16 = @ptrCast([*]const u16, &volume_name_info.FileName[0])[0..@divExact(volume_name_info.FileNameLength, 2)];
 
+        if (fmt.volume_name == .Nt) {
```
LemonBoy, line 1076: This is in the `(fmt.volume_name != .Nt)` branch, it'll never be executed.

lib/std/os/windows.zig
```diff
@@ -952,7 +955,53 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
+    out_buffer: []u16,
+) ![]u16 {
+    var full_buffer: [@sizeOf(OBJECT_NAME_INFORMATION) + PATH_MAX_WIDE * 2]u8 align(@alignOf(OBJECT_NAME_INFORMATION)) = undefined;
+    var info = @ptrCast(*OBJECT_NAME_INFORMATION, &full_buffer);
+    //buffer size is specified in bytes
+    const full_buffer_length = @intCast(ULONG, @sizeOf(OBJECT_NAME_INFORMATION) + std.math.min(PATH_MAX_WIDE, (out_buffer.len + 1) * 2));
+    //last argument would return the length required for full_buffer, not exposed here
+    const rc = ntdll.NtQueryObject(handle, .ObjectNameInformation, full_buffer[0..], full_buffer_length, null);
+    return switch (rc) {
+        .SUCCESS => if (@ptrCast(?[*]WCHAR, info.Name.Buffer)) |buffer| blk: {
```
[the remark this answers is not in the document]
rohlem: [The documentation for ObQueryNameString](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-obquerynamestring) , which I based this on since it seems to be used under the hood, mentions a `NULL` value under "Remarks" (`If the given object is unnamed [...]`).
`info.MaxLength` is also set to `0` in that case, so I'm happy to change the check to that, but returning an empty string would require checking for that in the consumer code (`GetFinalPathNameByHandle`).
The object's name isn't the empty string in this case, it has no name - I think either returning an optional or error still makes more sense.

lib/std/os/windows.zig
```diff
@@ -952,7 +955,53 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
 }
 
+pub fn QueryObjectName(
+    handle: HANDLE,
+    out_buffer: []u16,
+) ![]u16 {
+    var full_buffer: [@sizeOf(OBJECT_NAME_INFORMATION) + PATH_MAX_WIDE * 2]u8 align(@alignOf(OBJECT_NAME_INFORMATION)) = undefined;
+    var info = @ptrCast(*OBJECT_NAME_INFORMATION, &full_buffer);
+    //buffer size is specified in bytes
+    const full_buffer_length = @intCast(ULONG, @sizeOf(OBJECT_NAME_INFORMATION) + std.math.min(PATH_MAX_WIDE, (out_buffer.len + 1) * 2));
```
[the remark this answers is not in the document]
rohlem: Hmm, you're right, though now that I look at it again, with this exact size specification we could conversely remove the size check below.
But I guess the explicit check below is better in the sense of readability?

lib/std/os/windows.zig
```diff
@@ -952,7 +955,53 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
+    //last argument would return the length required for full_buffer, not exposed here
+    const rc = ntdll.NtQueryObject(handle, .ObjectNameInformation, full_buffer[0..], full_buffer_length, null);
+    return switch (rc) {
+        .SUCCESS => if (@ptrCast(?[*]WCHAR, info.Name.Buffer)) |buffer| blk: {
+            //resulting string length is specified in bytes
+            const path_length_unterminated = @divExact(info.Name.Length, 2);
+            if (out_buffer.len < path_length_unterminated) {
+                return error.NameTooLong;
+            }
+            std.mem.copy(WCHAR, out_buffer[0..path_length_unterminated], buffer[0..path_length_unterminated :0]);
```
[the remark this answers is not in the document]
rohlem: Huh, good eye. Strange to document that with the data structure, but I guess that means it could apply anywhere it's used...

lib/std/target.zig
```diff
@@ -109,6 +109,20 @@ pub const Target = struct {
...
+            pub const known_win10_build_numbers = [_]u32{
+                10240, //win10
+                10586, //win10_th2
+                14393, //win_rs1
+                15063, //win_rs2
+                16299, //win_rs3
+                17134, //win_rs4
+                17763, //win_rs5
+                18362, //win_19h1
+                19041, //win_20h1
```
Mouvedia, line 123: You forgot 2 builds: `18363, 19042`

cf https://github.com/ziglang/zig/commit/140c599559b6e048db2ec8476cb2cc08366f6aa8#r44615067

lib/std/os/windows.zig
```diff
@@ -952,7 +955,53 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
+}
+test "QueryObjectName" {
+    if (comptime builtin.os.tag != .windows)
+        return;
+
+    //any file will do; canonicalization works on NTFS junctions and symlinks, hardlinks remain separate paths.
+    const file = try std.fs.openSelfExe(.{});
+    defer file.close();
+    //make this large enough for the test runner exe path
+    var out_buffer align(16) = std.mem.zeroes([1 << 10]u16);
```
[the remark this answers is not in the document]
rohlem: Right, that is a leftover from earlier.
And sure, if we have the stack space for another PATH_MAX_WIDE that's fine by me.

lib/std/os/windows.zig
```diff
@@ -980,31 +1029,64 @@ pub fn GetFinalPathNameByHandle(
...
-    // Get NT volume name.
+    var path_buffer: [std.math.max(@sizeOf(FILE_NAME_INFORMATION), @sizeOf(OBJECT_NAME_INFORMATION)) + PATH_MAX_WIDE * 2]u8 align(@alignOf(FILE_NAME_INFORMATION)) = undefined;
     var volume_buffer: [@sizeOf(FILE_NAME_INFORMATION) + MAX_PATH]u8 align(@alignOf(FILE_NAME_INFORMATION)) = undefined; // MAX_PATH bytes should be enough since it's Windows-defined name
-    try QueryInformationFile(hFile, .FileVolumeNameInformation, volume_buffer[0..]);
 
-    const file_name = @ptrCast(*const FILE_NAME_INFORMATION, &path_buffer[0]);
-    const file_name_u16 = @ptrCast([*]const u16, &file_name.FileName[0])[0 .. file_name.FileNameLength / 2];
+    var file_name_u16: []const u16 = undefined;
+    var volume_name_u16: []const u16 = undefined;
+    if ((comptime (targetVersionIsAtLeast(WindowsVersion.win10_rs4) != true)) //need explicit comptime, because error returns affect return type
```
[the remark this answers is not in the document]
rohlem: It looks strange, but the value is of optional type `?bool`, not `bool` directly.
So this is supposed to catch both `null` and `false`. I don't think either of `!` and `== false` work for this.

lib/std/os/windows.zig
```diff
@@ -980,31 +1029,64 @@ pub fn GetFinalPathNameByHandle(
...
+        and !runtimeVersionIsAtLeast(WindowsVersion.win10_rs4))
+    {
+        const final_path = QueryObjectName(hFile, std.mem.bytesAsSlice(u16, path_buffer[0..])) catch |err| return switch (err) {
+            error.InvalidHandle => error.FileNotFound, //close enough?
```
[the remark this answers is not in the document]
rohlem: Close to the semantic meaning of "invalid file handle supplied" - assuming that a not found file be linked to an invalid handle.

lib/std/os/windows.zig
```diff
@@ -980,31 +1029,64 @@ pub fn GetFinalPathNameByHandle(
...
+    var path_buffer: [std.math.max(@sizeOf(FILE_NAME_INFORMATION), @sizeOf(OBJECT_NAME_INFORMATION)) + PATH_MAX_WIDE * 2]u8 align(@alignOf(FILE_NAME_INFORMATION)) = undefined;
     var volume_buffer: [@sizeOf(FILE_NAME_INFORMATION) + MAX_PATH]u8 align(@alignOf(FILE_NAME_INFORMATION)) = undefined; // MAX_PATH bytes should be enough since it's Windows-defined name
-    try QueryInformationFile(hFile, .FileVolumeNameInformation, volume_buffer[0..]);
 
-    const file_name = @ptrCast(*const FILE_NAME_INFORMATION, &path_buffer[0]);
-    const file_name_u16 = @ptrCast([*]const u16, &file_name.FileName[0])[0 .. file_name.FileNameLength / 2];
+    var file_name_u16: []const u16 = undefined;
+    var volume_name_u16: []const u16 = undefined;
+    if ((comptime (targetVersionIsAtLeast(WindowsVersion.win10_rs4) != true)) //need explicit comptime, because error returns affect return type
+        and !runtimeVersionIsAtLeast(WindowsVersion.win10_rs4))
```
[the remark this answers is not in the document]
rohlem: I was actually wondering whether we should cache that version in a global (that is, module-scope) variable - not just for this code path, but for all of `std.os.windows`, but it seemed like too big of a change to just throw in unsupervised.
So would an atomic version number `u32` in `std.zig.system.windows` be fitting, initialized once and then used instead of the `RtlGetVersion` call?
Or is there some benefit to having this particular flag be in its own variable directly?

lib/std/os/windows.zig
```diff
@@ -980,31 +1029,64 @@ pub fn GetFinalPathNameByHandle(
...
+            else => |e| e,
+        };
 
-    const volume_name = @ptrCast(*const FILE_NAME_INFORMATION, &volume_buffer[0]);
+        if (fmt.volume_name == .Nt) {
+            if (out_buffer.len < final_path.len) {
+                return error.NameTooLong;
+            }
+            std.mem.copy(u16, out_buffer[0..], final_path[0..]);
+            return final_path; //we can directly return the slice we received
```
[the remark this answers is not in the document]
rohlem: Ah, right, meant out_buffer of course, thanks.

lib/std/os/windows.zig
rohlem: No, the "other branch" (the else at line `1065` of the if at line `1037`) is for the newer windows version... unless I'm missing something?

lib/std/os/windows.zig
rohlem: No, this is the `win10_rs4 or newer` branch, the previous code I kept intact.

lib/std/target.zig
rohlem: I thought so at first too, but those don't impact [the `NTDDI` header constant we're trying to mirror here](https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt). (Also the values are unchanged from the previous implementation.)

Mouvedia: This page is dated 2020-06-19, `19042` was released the 2020-10-20.
i.e. it's outdated

rohlem: Okay, it seems that you're right, [this repository's own `sdkddkver.h`](https://github.com/ziglang/zig/blob/master/lib/libc/include/any-windows-any/sdkddkver.h#L144-L146) already includes more constants than I found when I was looking for them earlier. Thank you for pointing this out!

Mouvedia: FYI the next one will probably be 20231 (21H1)

lib/std/os/windows.zig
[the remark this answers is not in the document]
kubkon: I'm still on the fence with the idea of having `error.Unexpected`. If something like this happens, shouldn't we simply trigger a nonrecoverable runtime trap instead? Or do we intend for user-space code to be recoverable in those circumstances? If that's the case, anyone's got any ideas for a scenario when that would be the case?

lib/std/os/windows.zig
```diff
@@ -952,7 +955,53 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
+    if (comptime builtin.os.tag != .windows)
+        return;
+
+    //any file will do; canonicalization works on NTFS junctions and symlinks, hardlinks remain separate paths.
+    const file = try std.fs.openSelfExe(.{});
+    defer file.close();
+    //make this large enough for the test runner exe path
+    var out_buffer align(16) = std.mem.zeroes([1 << 10]u16);
+
+    var result_path = try QueryObjectName(file.handle, out_buffer[0..]);
```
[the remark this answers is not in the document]
kubkon: Huh, that's brilliant! Never thought of that 😅

lib/std/os/windows.zig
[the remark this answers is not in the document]
kubkon: Personally, I'm not sure I'm a big fan of having a global/thread-local state lying around, but on the other hand, calling `RtlGetVersion` every single time we execute `GetFinalPathNameByHandle` is a performance hog indeed. @andrewrk what are your thoughts on this?

lib/std/os/windows.zig
kubkon: Ah gotcha. It's fine as-is then.

lib/std/os/windows.zig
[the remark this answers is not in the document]
rohlem: "nonrecoverable runtime trap" (so basically `@panic`?) sounds very cut-throat to me.
If the file was supplied by a user, say interactively via `stdin`, maybe you just want to notify the user "this file didn't work, try another / moving it to a different path" (or something more specific if you figure out the underlying problem).
To me this seems like a fitting scenario for an error code: We failed to provide the requested service, but if/how to recover is up to the caller. That's why I made it return `error.Unnamed` before.
If we don't expect this API to be used with unnamed objects, then this scenario is "unexpected". `error.Unexpected` is not inherently reserved for kernel failures, so I think it fits.

But these are just my thoughts, whatever you think is best I'll implement.

lib/std/os/windows.zig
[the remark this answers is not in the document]
rohlem: @kubkon Right, that's why I originally refrained.
Maybe this is also a good place to mention that (A) we can always incrementally improve this in subsequent commits, and (B) `NtQueryObject` works for ALL windows versions (for now).
So if just removed the other path (or are satisfied with only the exhaustive `comptime` check), the runtime check instantly becomes superfluous.

lib/std/os/windows/bits.zig
```diff
@@ -65,6 +65,7 @@ pub const SIZE_T = usize;
 pub const TCHAR = if (UNICODE) WCHAR else u8;
 pub const UINT = c_uint;
 pub const ULONG_PTR = usize;
+pub const PULONG = *ULONG;
```
andrewrk, line 68: Prefer to use `*ULONG` directly rather than `PULONG`

lib/std/os/windows.zig
```diff
@@ -28,6 +28,9 @@ pub const gdi32 = @import("windows/gdi32.zig");
 
 pub usingnamespace @import("windows/bits.zig");
 
+//version detection
+usingnamespace std.zig.system.windows;
```
andrewrk, line 32: Prefer to avoid `usingnamespace` unless it's to set up the public declarations (as above). Here we should give it a name and refer to the functions therein via field access syntax.

lib/std/os/windows.zig
```diff
@@ -952,7 +955,53 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
+    handle: HANDLE,
+    out_buffer: []u16,
+) ![]u16 {
+    var full_buffer: [@sizeOf(OBJECT_NAME_INFORMATION) + PATH_MAX_WIDE * 2]u8 align(@alignOf(OBJECT_NAME_INFORMATION)) = undefined;
+    var info = @ptrCast(*OBJECT_NAME_INFORMATION, &full_buffer);
+    //buffer size is specified in bytes
+    const full_buffer_length = @intCast(ULONG, @sizeOf(OBJECT_NAME_INFORMATION) + std.math.min(PATH_MAX_WIDE, (out_buffer.len + 1) * 2));
+    //last argument would return the length required for full_buffer, not exposed here
+    const rc = ntdll.NtQueryObject(handle, .ObjectNameInformation, full_buffer[0..], full_buffer_length, null);
+    return switch (rc) {
```
andrewrk, line 968: When returning errors, prefer to put `return` on multiple lines. This will make the error return trace point nicely to the error code that triggered the crash, rather than pointing to `return switch (rc) {`.

lib/std/os/windows.zig
```diff
@@ -952,7 +955,53 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
     return @bitCast(u64, result);
 }
 
+pub fn QueryObjectName(
+    handle: HANDLE,
+    out_buffer: []u16,
+) ![]u16 {
+    var full_buffer: [@sizeOf(OBJECT_NAME_INFORMATION) + PATH_MAX_WIDE * 2]u8 align(@alignOf(OBJECT_NAME_INFORMATION)) = undefined;
```
andrewrk, line 962: Why 2 buffers with a memcpy? We could pass out_buffer directly to NtQueryObject right?

lib/std/os/windows.zig
```diff
@@ -952,7 +955,53 @@ pub fn SetFilePointerEx_CURRENT_get(handle: HANDLE) SetFilePointerError!u64 {
...
+        .INFO_LENGTH_MISMATCH => unreachable,
+        else => |e| unexpectedStatus(e),
+    };
+}
+test "QueryObjectName" {
+    if (comptime builtin.os.tag != .windows)
+        return;
+
+    //any file will do; canonicalization works on NTFS junctions and symlinks, hardlinks remain separate paths.
+    const file = try std.fs.openSelfExe(.{});
```
andrewrk, line 991: Better not to entangle this test with the functionality of openSelfExe. I suggest to use `std.testing.tmpDir`. You can see example usage in `std/fs/test.zig`.

lib/std/os/windows.zig
andrewrk, line 994: what is the purpose of calling mem.zeroes here?

lib/std/os/windows.zig
```diff
@@ -980,31 +1029,64 @@ pub fn GetFinalPathNameByHandle(
...
+    if ((comptime (targetVersionIsAtLeast(WindowsVersion.win10_rs4) != true)) //need explicit comptime, because error returns affect return type
+        and !runtimeVersionIsAtLeast(WindowsVersion.win10_rs4))
+    {
+        const final_path = QueryObjectName(hFile, std.mem.bytesAsSlice(u16, path_buffer[0..])) catch |err| return switch (err) {
+            error.InvalidHandle => error.FileNotFound, //close enough?
+            else => |e| e,
+        };
 
-    const volume_name = @ptrCast(*const FILE_NAME_INFORMATION, &volume_buffer[0]);
+        if (fmt.volume_name == .Nt) {
```
andrewrk, line 1045: Prefer to `switch` on enums rather than `if`

Side note @kubkon - why is `GetFinalPathNameByHandleFormat ` a struct containing an enum, and not just an enum?

lib/std/os/windows.zig
```diff
@@ -980,31 +1029,64 @@ pub fn GetFinalPathNameByHandle(
...
             std.mem.copy(u16, out_buffer[volume_name_u16.len..], file_name_u16);
 
             return out_buffer[0 .. volume_name_u16.len + file_name_u16.len];
-        },
+        }
+        //fallthrough for fmt.volume_name != .Nt
+    }
+
+    switch (fmt.volume_name) {
+        .Nt => unreachable, //handled above
```
andrewrk, line 1089: This construct can be replaced with `assert(fmt.volume_name == .Nt);`

lib/std/zig/system/windows.zig
```diff
@@ -0,0 +1,61 @@
...
+
+/// Returns whether the target os versions are uniformly at least as new as the argument:
+/// true/false if this holds for the entire target range, null if it only holds for some.
+pub fn targetVersionIsAtLeast(requested_version: WindowsVersion) ?bool {
+    const requested = @enumToInt(requested_version);
+    const version_range = std.builtin.os.version_range.windows;
+    const target_min = @enumToInt(version_range.min);
+    const target_max = @enumToInt(version_range.max);
+    return if (target_max < requested) false else if (target_min >= requested) true else null;
+}
```
andrewrk, line 55: This looks like duplicated code from `std.Target.Os.isAtLeast`

andrewrk (changes requested):
Thank you for taking on this project, @rohlem. This is good stuff. Needs a bit of cleanup, but the important parts are in place.

> _If_ we trust this new implementation, I would be inclined to maybe delete the other codepath (that needs two `ntdll` calls instead of just one) in favor of this one. (And maybe move the logic for finding DOS volume names into its own function.)
> But now that I've worked on this for a day, I thought I would ask for some feedback, to make sure I'm not overlooking anything crucial.

If it works empirically (which it sounds like it does) then I would be in favor of deleting the other codepath and deleting the version checking.

lib/std/os/windows.zig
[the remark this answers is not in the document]
daurnimator: `RtlGetVersion` should be pretty cheap: its a read from the PEB.
We could read from the PEB ourselves too, but there shouldn't be much of a difference.

kubkon:
I should probably explain what the motivation was for using `FileNormalizedNameInformation` instead of simply resorting to `NtQueryObject`. When I originally had a stab at this, I did a series of tests and comparisons to see how Windows user-level APIs actually do it using the Process Monitor to track the syscalls. It turned out that on my version of Win10, `GetFinalPathNameByHandleW` was calling `NtQueryInformationFile` with `FileNormalizedNameInformation` file information class. My guess was that this way we wouldn't have to worry about the maximum path name limitation, i.e., that the newer API would handle this for us. Having said that, it is true that ReactOS uses exclusively `NtQueryObject` to implement their version of `GetFinalPathNameByHandleW`. Also, I don't like the fact we have to check Windows version every time we call `GetFinalPathNameByHandle`. On the other hand, @daurnimator reassures us that it should be a pretty cheap call, so maybe this won't be a huge performance hog, but I'm not an expert here by any means. Anyhow, I'm happy if we remove the original alternate path and use `NtQueryObject` exclusively. If this poses some problems to do with path length limits down the line, we can always bring it back or come up with a better solution.

lib/std/os/windows.zig
rohlem: Yes, that's what the code did originally, which was commented on [here](https://github.com/ziglang/zig/pull/7379#discussion_r540323038), so I changed it.
I can change it back if you want - I don't think there was a particular reason besides readability - but I don't like stepping on anyone's toes, so I'll put that in as the last commit and then we can pick whichever turns out looking nicer.

lib/std/os/windows.zig
rohlem: I think that was a "sane default" during early testing that I forgot to remove. The code shouldn't rely on the value, I'll change it to undefined.

lib/std/os/windows/bits.zig
marler8997: Yeah, Zig has more "specific" ways of identifying pointers `*ULONG` vs `[*]ULONG`.  `PULONG` is more ambiguous.

lib/std/os/windows.zig
rohlem: Well, I'm not knowledgeable enough about these NT paths to say whether the returned string always starts with `\Device\`, it's just the only scenario I've encountered, but I don't have any network sharing set up f.e. .

`BadPathName` is already returned further down if the DOS volume name we've resolved unexpectedly contains null characters, which also seems like a "not-yet-disproven working thesis", so I think I'll stick to it here for consistency.
But let me know if you'd prefer the assert here, or in both places.

rohlem: Rethinking it yet again, the way we just blindly assume the prefix could lead to nonsensical results in unsafe build mode. An error return makes much more sense to me. But speak up if you feel differently.

rohlem:
Addressed all review comments, besides the one suggested `error -> assert` I commented on - that looks safer as an error to me (not blindly continuing in non-debug build modes).
The last commit removes all intermediate buffers, as @andrewrk suggested (... iiuc), to make the code more readable I added a helper function `std.mem.alignInSlice` to do the necessary in-place-alignment.

LemonBoy:
After 83 comments it looks good now!

> The last commit removes all intermediate buffers, as @andrewrk suggested (... iiuc), to make the code more readable I added a helper function std.mem.alignInSlice to do the necessary in-place-alignment.

The helper function makes the code much easier to read and review, thanks.

andrewrk:
Landed in 56c03881ebd8617e6506bfad01bf9cfdd4d3df7e

PR #7381, opened 2020-12-10
std: Improvement for utf8CountCodepoints

alamlintang:


LemonBoy:
All the extra code you see there is to make that function fast...

alamlintang:
Oh, I see. My bad then.

PR #7387, opened 2020-12-10
zig fmt: Fix alignment of initializer elements

### LINKED ISSUE #7289: zig fmt assertion failure
```zig
const moves = .{
    "U",            "L",            "F",
    "U'",
    "L'",
    "F'",
};
```

Triggers this assertion: https://github.com/ziglang/zig/blob/5b6cbd2e7cd00b19bdcfdeefc9b5d378ce8867a6/lib/std/zig/render.zig#L855

LemonBoy:
Resetting `column_counter` is not needed as the effective column number
is calculated by taking that value modulo `row_size`.

Closes #7289

cc @LakeByTheWoods


LakeByTheWoods:
Makes sense to me.

PR #7390, opened 2020-12-11
Stage 2: Fix c_longdouble abi.

antlilja:
The alignment of ```long double``` is not always 16 bytes, this PR doesn't fix the alignment but replaces it with a panic to prevent future bugs. Same for the size but the panic comes from within ```CType.longdouble.sizeInBits```.

I'm not sure what alignment long double should have even on x86 because it is highly dependent on the compiler.

### DIFF
--- a/src/type.zig
+++ b/src/type.zig
@@ -1005,7 +1005,7 @@ pub const Type = extern union {
             .f32 => return 4,
             .f64 => return 8,
             .f128 => return 16,
-            .c_longdouble => return 16,
+            .c_longdouble => return @divExact(CType.longdouble.sizeInBits(target), 8),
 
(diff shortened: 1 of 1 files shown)


andrewrk:
> it is highly dependent on the compiler.

In this case we can use the C ABI part of the target triple to find out which C compiler we are matching.

LemonBoy:
> I'm not sure what alignment long double should have even on x86 because it is highly dependent on the compiler.

Have a look at stage1, I've recently fixed/added a few `c_longdouble` definitions.

andrewrk:
I'm cleaning up old drafts, feel free to re-open if you are still working on it.

Issue #7391, opened 2020-12-11
ArrayHashMap: support all operations that ArrayList support


andrewrk:
This proposal is to treat ArrayHashMap as a data structure that can act both as an ArrayList and as a HashMap.

Here are some improvements I want to see:

 * `getOrPut` related functions should additionally return the entry index.
 * `remove` should be changed to `swapRemove`, and `orderedRemove` should be introduced. Both of these operations are present in ArrayList and match what will be happening to the Hash Map's inner ArrayList.
 * You should be able to directly modify the `entries` field which is an `ArrayList`. This causes all the methods which rely on the index data to become illegal to call, until you call `reIndex()` which then makes everything work again.
 * Introduce `fromOwnedArrayList` which takes ownership of an ArrayList as the entries field and calls `reIndex()`
 * Introduce `shrinkRetainingCapacity` which does the same thing as in ArrayList but maintains the hash map index.
 * Introduce `shrinkAndFree` which maps to ArrayList.shrink (and we should rename ArrayList.shrink to shrinkAndFree as well)
 * Introduce `pop`.

This use case comes from updating the `Cache` API `files` field to be an ArrayHashMap instead of an ArrayList:

https://github.com/ziglang/zig/blob/f7d600675c548b4e45b24f8e10ebdd892a6b8dba/src/Cache.zig#L178

daurnimator:
This sounds a lot like the http_headers data structure that we had. https://github.com/ziglang/std-lib-orphanage/blob/2c36a7894c689ecbaf63d5f489bb0c68773410c4/std/http_headers.zig#L9-L17

tetsuo-cpp:
I'm going to give this a try.

PR #7393, opened 2020-12-11
stage2: store Cache files in a map data structure

### LINKED ISSUE #7308: adding the same C source file redundantly to the same compilation results in a deadlock
To reproduce:

```
zig build-lib foo.c foo.c
```

Expected behavior: Even though it would typically cause symbol collisions, it should be allowed, because it is possible to supply each source file with a different set of preprocessor definitions, which would cause the source files to have different sets of symbols.

Actual behavior: Deadlock in the cache system

```
#0  0x00007f61a230b047 in flock ()
   from /nix/store/9df65igwjmf2wbw0gbrrgair6piqjgmi-glibc-2.31/lib/libc.so.6
#1  0x0000000000cd0d2f in std.os.flock (fd=132, operation=2) at /home/andy/dev/zig/lib/std/os.zig:4180
#2  0x0000000000b5c698 in std.fs.Dir.createFileZ (self=..., 
    sub_path_c=0x7ffea47e6ac0 "d6e0e2162c95843932835e08e7f71648.txt", flags=...)
    at /home/andy/dev/zig/lib/std/fs.zig:904
#3  0x0000000000b570e0 in std.fs.Dir.createFile (self=..., sub_path=..., flags=...)
    at /home/andy/dev/zig/lib/std/fs.zig:835
#4  0x0000000000b52245 in Cache.Manifest.hit (self=0x7ffea47e9b10)
    at /home/andy/dev/zig/src/Cache.zig:256
#5  0x0000000000c3276d in Compilation.updateCObject (comp=0x9b08278, c_object=0x9b0d540, 
    c_comp_progress_node=0x7ffea47ea048) at /home/andy/dev/zig/src/Compilation.zig:1755
#6  0x0000000000c033d5 in Compilation.performAllTheWork (self=0x9b08278)
    at /home/andy/dev/zig/src/Compilation.zig:1435
#7  0x0000000000bfdb13 in Compilation.update (self=0x9b08278)
    at /home/andy/dev/zig/src/Compilation.zig:1227
#8  0x0000000000bab707 in main.updateModule (gpa=0x8d2ec70 <c_allocator_state>, comp=0x9b08278, 
    zir_out_path=..., hook=...) at /home/andy/dev/zig/src/main.zig:1910
#9  0x0000000000aff68b in main.buildOutputType (gpa=0x8d2ec70 <c_allocator_state>, 
    arena=0x7ffea47ef898, all_args=..., arg_mode=...) at /home/andy/dev/zig/src/main.zig:1772
#10 0x0000000000ad749b in main.mainArgs (gpa=0x8d2ec70 <c_allocator_state>, arena=0x7ffea47ef898, 
    args=...) at /home/andy/dev/zig/src/main.zig:162
#11 0x0000000000ad6e0f in main (argc=649, argv=0x7ffea47efad8) at /home/andy/dev/zig/src/stage1.zig:45
```

andrewrk:
Cache now stores files as an StringArrayListHashMap instead of an
ArrayList. The key is the resolved file path, to avoid adding the same
file multiple times into the cache system when it is depended on
multiple times. This avoids re-hashing file contents that have already
been inserted into the cache.

This was an attempt to fix #7308, but it is solving a different problem.
I do think it is worth considering this improvement regardless. On the
other hand, it might be wasted CPU cycles / code bloat since the layer of code above
this one probably wants to do its own de-duplication as well.

This implementation would be much nicer with #7391.

### DIFF
--- a/src/Cache.zig
+++ b/src/Cache.zig
@@ -175,7 +175,8 @@ pub const Manifest = struct {
     hash: HashHelper,
     manifest_file: ?fs.File,
     manifest_dirty: bool,
-    files: std.ArrayListUnmanaged(File) = .{},
+    /// The key is the resolved_path.
+    files: std.StringArrayHashMapUnmanaged(File) = .{},
     hex_digest: [hex_digest_len]u8,
 
     /// Add a file as a dependency of process being cached. When `hit` is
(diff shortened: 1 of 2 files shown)


andrewrk:
Well tests are failing and I wasn't even sure if this was an improvement, so I'm going to throw this patch in the garbo

PR #7405, opened 2020-12-12
Implement refAllDeclsInTree, fix some docs

### LINKED ISSUE #7400: lazy decl analysis prevents doc comments from making it into generated html documentation
examples:
```
const NUL: u8 = 0;
const SOH: u8 = 1;
const STX: u8 = 2;
```

Rocknest:
Convert doc comments on nested namespaces to top-level doc comments so they get display in generated docs. Add a function to recursive reference such namespaces. 

Closes #7400

### DIFF
--- a/lib/std/crypto.zig
+++ b/lib/std/crypto.zig
@@ -4,8 +4,9 @@
 // The MIT license requires this copyright notice to be included in all copies
 // and substantial portions of the software.
 
-/// Authenticated Encryption with Associated Data
 pub const aead = struct {
+    //! Authenticated Encryption with Associated Data
+
     pub const aegis = struct {
         pub const Aegis128L = @import("crypto/aegis.zig").Aegis128L;
         pub const Aegis256 = @import("crypto/aegis.zig").Aegis256;
@@ -30,14 +31,15 @@ pub const aead = struct {
     };
 };
 
-/// Authentication (MAC) functions.
 pub const auth = struct {
+    //! Authentication (MAC) functions.
     pub const hmac = @import("crypto/hmac.zig");
     pub const siphash = @import("crypto/siphash.zig");
 };
 
-/// Core functions, that should rarely be used directly by applications.
 pub const core = struct {
+    //! Core functions, that should rarely be used directly by applications.
+
     pub const aes = @import("crypto/aes.zig");
     pub const Gimli = @import("crypto/gimli.zig").State;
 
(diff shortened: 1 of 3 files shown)


lib/std/crypto.zig
```diff
@@ -4,8 +4,9 @@
 // The MIT license requires this copyright notice to be included in all copies
 // and substantial portions of the software.
 
-/// Authenticated Encryption with Associated Data
 pub const aead = struct {
+    //! Authenticated Encryption with Associated Data
```
daurnimator, line 8: We generally prefer `///` comments. The `//!` form was added so that files could document themselves.

Rocknest: I know. But auto-doc tool will not show `///` comments for namespaces, and it makes sense, there is only one top level doc comment for a namespace, but it can have multiple aliases each with its own docs, so what should be displayed it that case? I think the top level doc comment. So i converted these to get them displayed in auto-docs, also its more consistent to document namespaces always with `//!`

Vexu: Generated docs not showing doc comments for namespaces is a bug in the generated docs, not with how things are documented. Any aliasing issues would also happen with regular structs since they are the same thing to the compiler.  `//!` comments are mostly intended for top-level use.

lib/std/testing.zig
```diff
@@ -435,3 +435,23 @@ pub fn refAllDecls(comptime T: type) void {
         _ = decl;
     }
 }
+
+/// Given a type, recursively reference all the declarations inside,
+/// so that the semantic analyzer sees them. Up to `levels` deep. 
+pub fn refAllDeclsInTree(comptime T: type, comptime levels: u8) void {
```
andrewrk, line 441: I don't want to do this. This is a kludge working around the language shortcomings. I don't want people fighting with the compiler about recursive levels and trying to debug why stuff does or doesn't show up in the docs.

This function is fighting against the language but we need to take a step back and figure out how to make the language better with respect to auto generated docs. I'm not satisfied with this as a solution. I'd rather keep things painful and annoying until we have a proper solution.

andrewrk (changes requested):

lib/std/testing.zig
Rocknest: @andrewrk Is there an open issue for that? Im thinking of some new visibility keyword, for example `api` to be used instead of `pub` so its compiled even when not referenced (at least when generating docs). 

By the way, should existing `refAllDecs` be removed since its a hack?

lib/std/crypto.zig
Rocknest: @Vexu is there any reason to allow `//!` not being actually top level?

Vexu: I don't think so, just more convenient to parse. Further discussion here https://github.com/ziglang/zig/pull/5801#discussion_r450041448.

PR #7411, opened 2020-12-12
std: Improve sigaction interface

LemonBoy:
Add a smoke test to prevent regressions.

### DIFF
--- a/lib/std/os/bits/darwin.zig
+++ b/lib/std/os/bits/darwin.zig
@@ -126,6 +126,11 @@ pub const timespec = extern struct {
 pub const sigset_t = u32;
 pub const empty_sigset: sigset_t = 0;
 
+pub const SIG_ERR = @intToPtr(?Sigaction.sigaction_fn, maxInt(usize));
+pub const SIG_DFL = @intToPtr(?Sigaction.sigaction_fn, 0);
+pub const SIG_IGN = @intToPtr(?Sigaction.sigaction_fn, 1);
+pub const SIG_HOLD = @intToPtr(?Sigaction.sigaction_fn, 5);
+
 pub const siginfo_t = extern struct {
     signo: c_int,
     errno: c_int,
--- a/lib/std/os/bits/darwin.zig
+++ b/lib/std/os/bits/darwin.zig
@@ -124,13 +124,35 @@ pub const timespec = extern struct {
 };
 
 pub const sigset_t = u32;
-pub const empty_sigset = sigset_t(0);
+pub const empty_sigset: sigset_t = 0;
+
+pub const siginfo_t = extern struct {
+    signo: c_int,
+    errno: c_int,
+    code: c_int,
+    pid: pid_t,
+    uid: uid_t,
+    status: c_int,
+    addr: *c_void,
+    value: extern union {
+        int: c_int,
+        ptr: *c_void,
+    },
+    si_band: c_long,
+    _pad: [7]c_ulong,
+};
 
 /// Renamed from `sigaction` to `Sigaction` to avoid conflict with function name.
 pub const Sigaction = extern struct {
-    handler: fn (c_int) callconv(.C) void,
-    sa_mask: sigset_t,
-    sa_flags: c_int,
+    pub const handler_fn = fn (c_int) callconv(.C) void;
+    pub const sigaction_fn = fn (c_int, *const siginfo_t, ?*const c_void) callconv(.C) void;
+
+    handler: extern union {
+        handler: ?handler_fn,
+        sigaction: ?sigaction_fn,
+    },
+    mask: sigset_t,
+    flags: c_uint,
 };
 
 pub const dirent = extern struct {
--- a/lib/std/os/bits/openbsd.zig
+++ b/lib/std/os/bits/openbsd.zig
@@ -753,13 +753,18 @@ pub const SIG_IGN = @intToPtr(?Sigaction.sigaction_fn, 1);
 
 /// Renamed from `sigaction` to `Sigaction` to avoid conflict with the syscall.
 pub const Sigaction = extern struct {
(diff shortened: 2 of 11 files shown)


zigazeljko:
By the way, why is `sigset_t` so large on linux? According to C headers, it should be `[NSIG / 32]u32` (i.e. `[4]u32` for MIPS, `[2]u32` for everything else), but here we have `[32]u32`.

LemonBoy:
> By the way, why is sigset_t so large on linux? According to C headers, it should be [NSIG / 32]u32 (i.e. [4]u32 for MIPS, [2]u32 for everything else), but here we have [32]u32.

This is musl's definition:
```c
typedef struct __sigset_t { unsigned long __bits[128/sizeof(long)]; } sigset_t;
```

LemonBoy:
@kubkon can you have a look at the macos crash when you have a moment?

lib/std/os/bits/darwin.zig
```diff
@@ -124,13 +124,35 @@ pub const timespec = extern struct {
 };
 
 pub const sigset_t = u32;
-pub const empty_sigset = sigset_t(0);
+pub const empty_sigset: sigset_t = 0;
```
semarie, line 127: could you adjust in the same way `empty_sigset` for openbsd ?

lib/std/os/test.zig
```diff
@@ -646,3 +646,34 @@ test "shutdown socket" {
...
+var signal_test_failed = true;
+
+test "sigaction" {
+    if (builtin.os.tag == .wasi or builtin.os.tag == .windows)
+        return error.SkipZigTest;
+
+    const S = struct {
+        fn handler(sig: i32, info: *const os.siginfo_t, ctx_ptr: ?*const c_void) callconv(.C) void {
+            // Check that we received the correct signal.
+            if (sig == os.SIGUSR1 and sig == info.signo)
```
semarie, line 659: after testing on OpenBSD, `info.signo` doesn't work here: `error: no member named 'signo' in union 'std.os.bits.openbsd.siginfo_t'`

but it seems the `siginfo_t` definition for openbsd could be improved to match it. it could be addressed separately

kubkon:
@LemonBoy I'm not sure why but it seems `SA_RESETHAND` is not preserved when checking what sigaction has been set. However, specifying it upon installing the sighandler is correctly registered and the handler is reset to `SIG_DFL` after the signal is raised. Anyhow, the following diff illustrates what I mean and passes the test on macOS:

```diff
diff --git a/lib/std/os/bits/darwin.zig b/lib/std/os/bits/darwin.zig
index a6dbff845..e0c3d12fc 100644
--- a/lib/std/os/bits/darwin.zig
+++ b/lib/std/os/bits/darwin.zig
@@ -142,6 +142,11 @@ pub const siginfo_t = extern struct {
     _pad: [7]c_ulong,
 };

+pub const SIG_ERR = @intToPtr(?Sigaction.sigaction_fn, maxInt(usize));
+pub const SIG_DFL = @intToPtr(?Sigaction.sigaction_fn, 0);
+pub const SIG_IGN = @intToPtr(?Sigaction.sigaction_fn, 1);
+pub const SIG_HOLD = @intToPtr(?Sigaction.sigaction_fn, 5);
+
 /// Renamed from `sigaction` to `Sigaction` to avoid conflict with function name.
 pub const Sigaction = extern struct {
     pub const handler_fn = fn (c_int) callconv(.C) void;
diff --git a/lib/std/os/test.zig b/lib/std/os/test.zig
index bbc6c968f..648537dbf 100644
--- a/lib/std/os/test.zig
+++ b/lib/std/os/test.zig
@@ -672,8 +672,19 @@ test "sigaction" {
     // Check that we can read it back correctly.
     os.sigaction(os.SIGUSR1, null, &old_sa);
     testing.expectEqual(S.handler, old_sa.handler.sigaction.?);
-    testing.expect((old_sa.flags & os.SA_RESETHAND) != 0);
+    testing.expect((old_sa.flags & os.SA_SIGINFO) != 0);
+
+    if (!std.Target.current.isDarwin()) {
+        testing.expect((old_sa.flags & os.SA_RESETHAND) != 0);
+    }
+
     // Invoke the handler.
     try os.raise(os.SIGUSR1);
     testing.expect(signal_test_failed == false);
+
+    if (std.Target.current.isDarwin()) {
+        // Check if the handler has been correctly reset to SIG_DFL
+        os.sigaction(os.SIGUSR1, null, &old_sa);
+        testing.expectEqual(os.SIG_DFL, old_sa.handler.sigaction);
+    }
 }
```

This might also be the case for BSD in general but I don't have any handy so can't verify.

LemonBoy:
@kubkon Thank you! For some reason [Xnu](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/kern/kern_sig.c#L493) drops some flags.

@semarie Can you verify the test compiles now?

lib/std/os/bits/openbsd.zig
```diff
@@ -789,11 +789,22 @@ pub const _ksiginfo = extern struct {
             addr: ?*c_void,
             trapno: c_int,
         },
+        __pad: [128 - 3 * @sizeOf(c_int)]u8,
```
semarie, line 792: I found it is error prone. the size is right (`116 u8`), but the definition in OpenBSD is `29 c_int`. it will be complex to track it if it changes at some point.

semarie:
on OpenBSD, the assertion `std.debug.assert(@sizeOf(siginfo_t) == 128)` failed while compiling the test

kubkon (approved):

kubkon:
@LemonBoy I assume you want to give it another go to figure out the OpenBSD failures? If not, lemme know, and I'll merge as-is.

LemonBoy:
> @LemonBoy I assume you want to give it another go to figure out the OpenBSD failures? If not, lemme know, and I'll merge as-is.

The last commit should've fixed that problem, you can merge this.

PR #7414, opened 2020-12-12
Add baudrate constants for Linux.

tauoverpi:


lib/std/os/bits/linux.zig
```diff
@@ -1826,6 +1826,39 @@ pub const tcflag_t = u32;
 
 pub const NCCS = 32;
 
+pub const B0 = 0o0000000;
```
daurnimator, line 1829: These look like they should be an `enum` in zig

tauoverpi: Issue with making them an enum is that they're ored together when used
```
linux.CS8 | linux.B2400 | linux.CLOCAL | linux.CREAD | linux.PARENB
```
Unless the intention is to `@enumToInt(.B2400) | linux.CREAD | ...` for each use.

PR #7417, opened 2020-12-13
create SendToError

lun-4:
See https://github.com/ziglang/zig/pull/7316#issuecomment-739698574 for reference.

### DIFF
--- a/lib/std/fs/file.zig
+++ b/lib/std/fs/file.zig
@@ -698,7 +698,6 @@ pub const File = struct {
             error.FastOpenAlreadyInProgress,
             error.MessageTooBig,
             error.FileDescriptorNotASocket,
-            error.AddressFamilyNotSupported,
             => return self.writeFileAllUnseekable(in_file, args),
 
             else => |e| return e,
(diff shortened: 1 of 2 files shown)


marler8997:
The point of having a separate error type for send is to remove the AddressFamilyNotSupported error. Getting that error when calling send can be an unreachable.

lun-4:
I compared the errors between send and sento (POSIX reference [here](https://pubs.opengroup.org/onlinepubs/9699919799/functions/send.html) and [here](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sendto.html)), and sendto has a larger error set than just EAFNOSUPPORT/AddressFamilyNotSupported:
 - For AF_UNIX sockets: ELOOP, ENAMETOOLONG, ENOENT, ENOTDIR
 - EDESTADDRREQ, EHOSTUNREACH, EISCONN, ENOMEM

Should those be mapped as well?

marler8997:
Yeah, map all the differences you find but keep in mind that you need to take all operating systems into account (not just posix).

PR #7420, opened 2020-12-13
std.meta.FieldEnum

daurnimator:
I've been thinking of writing this function for a while..... little did I know @tadeokondrak beat me too it; with the same name and implementation


tadeokondrak:
I know I was the one who added it, but I was thinking it might be better to go back to plain string literals because of the inconsistency with `@fieldParentPtr`.

marler8997:
Well that's clever

marler8997:
@tadeokondra, you can fix it by adding `std.meta.fieldParentPtr` correct?

PR #7422, opened 2020-12-13
stage1: float_*_ieee597: only swap bytes when targeting different endianness than native

koachan:
The forced-LE buffer ordering of those functions interferes with the ordering requirements of other functions called by buf_*_value_bytes, resulting in miscompilation of comptime `@bitCast`s on BE machines.

### DIFF
--- a/src/stage1/zigendian.h
+++ b/src/stage1/zigendian.h
@@ -1,0 +1,34 @@
+#ifndef ZIG_ENDIAN_H
+#define ZIG_ENDIAN_H
+
+// Every OSes seem to define endianness macros in different files.
+#if defined(__APPLE__)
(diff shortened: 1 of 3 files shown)


src/stage1/zigendian.h
```diff
@@ -0,0 +1,34 @@
+#ifndef ZIG_ENDIAN_H
+#define ZIG_ENDIAN_H
+
+// Every OSes seem to define endianness macros in different files.
```
LemonBoy, line 4: Maybe move this in `utils_base.hpp` instead of a new file?

koachan: It's also used in parse_f128.c, would it be fine to move it to a C++ header file?

PR #7431, opened 2020-12-13
stage1: Fix crash in can_mutate_comptime_var_state

### LINKED ISSUE #7426: compiler bug triggered in comptime function
I was told on the Zig IRC channel that this concerns a compiler bug I should file a report for.

The goal of the program listed below, is to create a form of type constraint. Similar to the way Go interfaces work. It is located in the comptime function assertInterface() (At least enough of the function to trigger the bug is included in the listing).

The program below should yield a compiler error as the second assertInterface() call should fail.
Using `zig build` gives the following:

```
$ zig build
thing...The following command exited with error code 5:
C:\zig\zig.exe build-exe C:\code\zig\thing\src\main.zig --cache-dir C:\code\zig\thing\zig-cache --name thing --enable-cache
error: the following build command failed with exit code 5:
C:\code\zig\thing\zig-cache\o\9202bbbc3ff09f161ca6c7ed75e0ce68\build.exe C:\zig\zig.exe C:\code\zig\thing C:\code\zig\thing\zig-cache
```

Calling the following manually, yields no errors at all:

```
$ C:\zig\zig.exe build-exe C:\code\zig\thing\src\main.zig --cache-dir C:\code\zig\thing\zig-cache --name thing --enable-cache
```
The same happens when manually deleting the zig-cache directory before running the build.

The code that triggers the issue:
```
const std = @import("std");
const TypeInfo = std.builtin.TypeInfo;

pub fn main() void {
    const x = FullImpl{
        .a = 1,
        .b = 2,
    };
    const y = NotFullImpl{
        .b = 1,
    };

    // This should compile without error.
    assertInterface(x, Interface);

    // This should yield a compiler error - NotFullImpl is missing field 'a' and function 'foo'.
    assertInterface(y, Interface);
}

// Interface defines the interface our types should qualify as.
const Interface = struct {
    a: u8,
    pub fn foo() void {}
    pub fn bar() void {}
};

// FullImpl is a test struct that entirely implements Interface.
const FullImpl = struct {
    a: u8,
    b: u16,
    pub fn foo() void {}
    pub fn bar() void {}
    pub fn baz() void {}
};

// NotFullImpl is a test struct that does not entirely implement Interface.
// It is missing field 'a' and function 'foo'.
const NotFullImpl = struct {
    b: u16,
    pub fn bar() void {}
    pub fn baz() void {}
};

/// assertInterface asserts that x qualifies as interface I.
/// Meaning that x is a struct which implements the fields and functions
/// specified by I.
pub fn assertInterface(x: anytype, comptime I: type) void {
    comptime {
        const haveType = @typeInfo(@TypeOf(x));
        const wantType = @typeInfo(I);
        const haveStruct = haveType.Struct;
        const wantStruct = wantType.Struct;

        // Make sure x defines all fields in I.
        inline for (wantStruct.fields) |field| {
            if (!hasField(haveStruct, field)) {
                @compileError(std.fmt.comptimePrint("type {} does not implement interface {}: missing field '{}'",
                    .{haveType, wantType, field.name}));
            }
        }

        // Make sure x defines all decls in I.
        inline for (wantStruct.decls) |decl| {
            if (!hasDecl(haveStruct, decl)) {
                @compileError(std.fmt.comptimePrint("type {} does not implement interface {}: missing decl '{}'", 
                    .{haveType, wantType, decl.name}));
            }
        }
    }
}

/// hasDecl returns true if x contains a declaration with the properties of d.
fn hasDecl(comptime x: TypeInfo.Struct, comptime d: TypeInfo.Declaration) bool {
    for (x.decls) |decl| {
        if (isDecl(decl, d)) {
            return true;
        }
    }
    return false;
}

/// hasField returns true if x contains a field with the properties of f.
fn hasField(comptime x: TypeInfo.Struct, comptime f: TypeInfo.StructField) bool {
    for (x.fields) |field| {
        if (isField(field, f)) {
            return true;
        }
    }
    return false;
}

/// isDecl returns true if a and b are considered equal.
fn isDecl(comptime a: TypeInfo.Declaration, comptime b: TypeInfo.Declaration) bool {
    if (a.is_pub != b.is_pub or !strEquals(a.name, b.name))
        return false;
 
    comptime af = a.data.Fn;
    comptime bf = b.data.Fn;

    return (af.fn_type == bf.fn_type and
            af.is_export == bf.is_export and
            af.is_extern == bf.is_extern and
            af.is_var_args == bf.is_var_args and
            af.return_type == bf.return_type and
            strListEquals(af.arg_names, bf.arg_names));
}

/// isField returns true if a and b are considered equal.
fn isField(comptime a: TypeInfo.StructField, comptime b: TypeInfo.StructField) bool {
    return (a.field_type == b.field_type and strEquals(a.name, b.name));
}

/// strListEquals returns true if the given string lists are equal.
fn strListEquals(comptime a: []const []const u8, comptime b: []const []const u8) bool {
    if (a.len != b.len)
        return false;

    for (a) |str, i| {
        if (!strEquals(str, b[i]))
            return false;
    }

    return true;
}

/// strEquals returns true iff a == b.
pub fn strEquals(comptime a: []const u8, comptime b: []const u8) bool {
    if (a.len != b.len)
        return false;

    for (a) |c, i| {
        if (c != b[i])
            return false;
    }

    return true;
}
```

LemonBoy:
No lazy value can mutate global state, no need to resolve them.

If you're reading this then please provide a minimized test case.

Closes #7426

### DIFF
--- a/src/stage1/analyze.cpp
+++ b/src/stage1/analyze.cpp
@@ -5749,6 +5749,28 @@ static bool can_mutate_comptime_var_state(ZigValue *value) {
     assert(value != nullptr);
     if (value->special == ConstValSpecialUndef)
         return false;
+
+    if (value->special == ConstValSpecialLazy) {
+        // No lazy value has side effects.
+        // Use a switch here to get a compile error whenever a new kind of lazy
+        // value is added.
+        switch (value->data.x_lazy->id) {
+            case LazyValueIdInvalid:
+                zig_unreachable();
+
+            case LazyValueIdAlignOf:
+            case LazyValueIdSizeOf:
+            case LazyValueIdPtrType:
+            case LazyValueIdOptType:
+            case LazyValueIdSliceType:
+            case LazyValueIdFnType:
+            case LazyValueIdErrUnionType:
+            case LazyValueIdArrayType:
+            case LazyValueIdTypeInfoDecls:
+                return false;
+        }
+    }
+
     switch (value->type->id) {
         case ZigTypeIdInvalid:
             zig_unreachable();
--- a/test/stage1/behavior/misc.zig
+++ b/test/stage1/behavior/misc.zig
@@ -752,3 +752,10 @@ test "extern variable with non-pointer opaque type" {
     @export(var_to_export, .{ .name = "opaque_extern_var" });
     expect(@ptrCast(*align(1) u32, &opaque_extern_var).* == 42);
 }
+
+test "lazy typeInfo value as generic parameter" {
+    const S = struct {
+        fn foo(args: anytype) void {}
+    };
+    S.foo(@typeInfo(@TypeOf(.{})));
+}


Vexu:
> If you're reading this then please provide a minimized test case.

```zig
pub fn main() void {
    comptime foo(@typeInfo(@TypeOf(.{})));
}
fn foo(args: anytype) void {}
```

PR #7438, opened 2020-12-13
Added test file in tests/stage1/behavior/bugs for issue 6905

Jarek1542:


LemonBoy:
The ticket is still open meaning this test won't pass.
You also forgot to reference the new file in `tests/stage1/behavior.zig`, that's why the test suite didn't error out.

PR #7444, opened 2020-12-14
openbsd: crt0.o file is different when build static or dynamic mode

semarie:
it is a first step for support static *and* dynamic mode for openbsd

### DIFF
--- a/src/link/Elf.zig
+++ b/src/link/Elf.zig
@@ -1441,8 +1441,14 @@ fn linkWithLLD(self: *Elf, comp: *Compilation) !void {
 
     if (link_in_crt) {
         const crt1o: []const u8 = o: {
-            if (target.os.tag == .netbsd or target.os.tag == .openbsd) {
+            if (target.os.tag == .netbsd) {
                 break :o "crt0.o";
+            } else if (target.os.tag == .openbsd) {
+                if (self.base.options.link_mode == .Static) {
+                    break :o "rcrt0.o";
+                } else {
+                    break :o "crt0.o";
+                }
             } else if (target.isAndroid()) {
                 if (self.base.options.link_mode == .Dynamic) {
                     break :o "crtbegin_dynamic.o";


src/link/Elf.zig
```diff
@@ -1441,8 +1441,14 @@ fn linkWithLLD(self: *Elf, comp: *Compilation) !void {
 
     if (link_in_crt) {
         const crt1o: []const u8 = o: {
-            if (target.os.tag == .netbsd or target.os.tag == .openbsd) {
+            if (target.os.tag == .netbsd) {
                 break :o "crt0.o";
+            } else if (target.os.tag == .openbsd) {
+                if (self.base.options.link_mode == .Static) {
+                    break :o "rcrt0.o";
```
daurnimator, line 1448: usually `rcrt0.o` is for pie builds. are there any openbsd docs you could link to that confirm its for all static builds?

semarie:
PIE is the default on OpenBSD. It is documented (as default option) on [clang-local](https://man.openbsd.org/clang-local) man page, which reference local modification to `clang`:

> clang will generate PIE code by default, allowing the system to load the resulting binary at a random location. This behavior can be turned off by passing -fno-pie to the compiler and -nopie to the linker. It is also turned off when the -pg flag is used.

Also, the exact linker invocation used when compiling a static binary is the following:
```
$ cc -v -static test.c  
OpenBSD clang version 10.0.1 
Target: amd64-unknown-openbsd6.8
Thread model: posix
InstalledDir: /usr/bin
 "/usr/bin/cc" -cc1 -triple amd64-unknown-openbsd6.8 -emit-obj -mrelax-all -disable-free -disable-llvm-verifier -discard-value-names -main-file-name test.c -static-define -mrelocation-model pic -pic-level 1 -pic-is-pie -mthread-model posix -mframe-pointer=all -relaxed-aliasing -fno-rounding-math -masm-verbose -mconstructor-aliases -munwind-tables -target-cpu x86-64 -target-feature +retpoline-indirect-calls -target-feature +retpoline-indirect-branches -dwarf-column-info -fno-split-dwarf-inlining -debugger-tuning=gdb -v -resource-dir /usr/lib/clang/10.0.1 -fdebug-compilation-dir /tmp/b -ferror-limit 19 -fmessage-length 239 -fwrapv -D_RET_PROTECTOR -ret-protector -fgnuc-version=4.2.1 -fobjc-runtime=gnustep -fdiagnostics-show-option -fno-builtin-malloc -fno-builtin-calloc -fno-builtin-realloc -fno-builtin-valloc -fno-builtin-free -fno-builtin-strdup -fno-builtin-strndup -faddrsig -o /tmp/test-dbd0b9.o -x c test.c
clang -cc1 version 10.0.1 based upon LLVM 10.0.1 default target amd64-unknown-openbsd6.8
#include "..." search starts here:
#include <...> search starts here:
 /usr/lib/clang/10.0.1/include
 /usr/include
End of search list.
 "/usr/bin/ld" -e __start --eh-frame-hdr -Bstatic -o a.out /usr/bin/../lib/rcrt0.o /usr/bin/../lib/crtbegin.o -L/usr/bin/../lib -L/usr/lib /tmp/test-dbd0b9.o -lcompiler_rt -lc -lcompiler_rt /usr/bin/../lib/crtend.o
```

daurnimator:
> PIE is the default on OpenBSD

`crt.0` would still correct for non-pie builds though, right?

I think you just need to fall through to the code path of
```
            } else if (self.base.options.link_mode == .Static) {
                if (self.base.options.pie) {
                    break :o "rcrt1.o";
                } else {
                    break :o "crt1.o";
                }
```

Possibly the default for pie on/off needs to be changed on openbsd.

semarie:
> crt.0 would still correct for non-pie builds though, right?

yes, `.Static` + `-nopie` would need `crt1.o`

> Possibly the default for pie on/off needs to be changed on openbsd.

zig is already configured to [require PIE](https://github.com/ziglang/zig/blob/97c0e1cc41c24c6cbb60117751d5b82dcd9d0e43/src/target.zig#L150) for openbsd

PR #7449, opened 2020-12-15
add termios bits for darwin

kristoff-it:
Added the necessary bits to make https://github.com/jessrud/zbox run on macOS. Funny how there are decoy headerfiles with incompatible definitions sprinkled around in `/Libary`.

@jessrud pinging you just as an FYI. I ran both examples and they work, but there are some minor problems (e.g. neither example fully resets the terminal state correctly on exit).

### DIFF
--- a/lib/std/os/bits/darwin.zig
+++ b/lib/std/os/bits/darwin.zig
@@ -1576,37 +1576,37 @@ pub const VTIME = 17;
 pub const VSTATUS = 18;
 pub const NCCS = 20; // 2 spares (7, 19)
 
-pub const IGNBRK = 0x00000001;
-pub const BRKINT = 0x00000002;
-pub const IGNPAR = 0x00000004;
-pub const PARMRK = 0x00000008;
-pub const INPCK = 0x00000010;
-pub const ISTRIP = 0x00000020;
-pub const INLCR = 0x00000040;
-pub const IGNCR = 0x00000080;
-pub const ICRNL = 0x00000100;
-pub const IXON = 0x00000200;
-pub const IXOFF = 0x00000400;
-pub const IXANY = 0x00000800;
-pub const IMAXBEL = 0x00002000;
-pub const IUTF8 = 0x00004000;
+pub const IGNBRK = 0x00000001; // ignore BREAK condition
+pub const BRKINT = 0x00000002; // map BREAK to SIGINTR
+pub const IGNPAR = 0x00000004; // ignore (discard) parity errors
+pub const PARMRK = 0x00000008; // mark parity and framing errors
+pub const INPCK = 0x00000010; // enable checking of parity errors
+pub const ISTRIP = 0x00000020; // strip 8th bit off chars
+pub const INLCR = 0x00000040; // map NL into CR
+pub const IGNCR = 0x00000080; // ignore CR
+pub const ICRNL = 0x00000100; // map CR to NL (ala CRMOD)
+pub const IXON = 0x00000200; // enable output flow control
+pub const IXOFF = 0x00000400; // enable input flow control
+pub const IXANY = 0x00000800; // any char will restart after stop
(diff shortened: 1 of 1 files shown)


lib/std/os/bits/darwin.zig
```diff
@@ -1554,3 +1554,186 @@ pub const rlimit = extern struct {
 pub const SHUT_RD = 0;
 pub const SHUT_WR = 1;
 pub const SHUT_RDWR = 2;
+
+// Term
+pub const VEOF = 0;
```
kubkon, line 1559: Would a request to add comments from the matching libc header file be an overkill? See [here](https://github.com/ziglang/zig/blob/master/lib/libc/include/x86_64-macos-gnu/sys/termios.h#L76) for comments.

lib/std/os/bits/darwin.zig
```diff
@@ -1554,3 +1554,186 @@ pub const rlimit = extern struct {
...
+pub const VEOL = 1;
```
kubkon, line 1560: Could we constrain the types rather than leave them hanging at `usize`?

lib/std/os/bits/darwin.zig
```diff
@@ -1554,3 +1554,186 @@ pub const rlimit = extern struct {
...
+    ws_row: u16,
+    ws_col: u16,
+    ws_xpixel: u16,
+    ws_ypixel: u16,
+};
+
+pub const TIOCGWINSZ = comptime ior(0x40000000, 't', 104, @sizeOf(winsize));
+pub const IOCPARM_MASK = 0x1fff;
+
+fn ior(comptime inout: u32, comptime group: usize, comptime num: usize, comptime len: usize) usize {
```
andrewrk, line 1737: no need for comptime args here

lib/std/os/bits/darwin.zig
```diff
@@ -1554,3 +1554,186 @@ pub const rlimit = extern struct {
...
+};
+
+pub const winsize = extern struct {
+    ws_row: u16,
+    ws_col: u16,
+    ws_xpixel: u16,
+    ws_ypixel: u16,
+};
+
+pub const TIOCGWINSZ = comptime ior(0x40000000, 't', 104, @sizeOf(winsize));
```
andrewrk, line 1734: this `comptime` is redundant because the expression is in the global scope.

andrewrk (changes requested):
the beauty of `comptime` is that 99% of the time you don't need to actually use the keyword :)

lib/std/os/bits/darwin.zig
kristoff-it: Truth is I don't know what these variables are used for (nor which syscall is accepting them). I could make an extern enum, but then which numeric type should the enum be? In the original code they are macro definitions.

lib/std/os/bits/darwin.zig
kristoff-it: Done!

lib/std/os/bits/darwin.zig
kubkon: Oh sure. I still think in Zig we are in the privileged position to be *always* able to specify the precise bit-width of a primitive type, and I believe we should always strive to do this as this can uncover potential bugs when something changes in the future between versions. Unlike C/C++, we don't have to resort to completely unhygienic preproc macros!

Having said that, I'm fine leaving it as-is for now, but would really welcome a subsequent PR that tries attaching a type to all these. :-)

kristoff-it: Yes, I think somebody that actually knows what this stuff does should do a cleanup pass in the future, I'm only trying to make my chat application to build on macos :)

kubkon: Chances are you are the expert on these now :D

PR #7222, opened 2020-11-25
stage1: Force union member types to be resolved

LemonBoy:
No test case because I couldn't reduce the huuuge test case.
Fixes the problem discovered by @ifreund.

PR #7453, opened 2020-12-15
stage1: Don't skip steps when analyzing union types

### LINKED ISSUE #7451: @unionInit on a union with a separately-defined tag type and zero-sized payload causes a compiler crash.
One interesting thing to note is that removing the type of the enum tag (replace `const Type = enum(u8)` with `const Type = enum`) fixes the crash.

Error:

```
Assertion failed at /home/kenta/Desktop/zig/src/stage1/codegen.cpp:4670 in ir_render_union_field_ptr. This is a bug in the Zig compiler.
Aborted (core dumped)
```

Reduced test case:

```zig
pub const Type = enum(u8) { no_op };

pub const Data = union(Type) {
    no_op: void,

    pub fn decode(buf: []const u8) !Data {
        return @unionInit(Data, "no_op", {});
    }
};

pub fn main() !void {
    var data: Data = .{ .no_op = .{} };
    _ = try Data.decode(&[_]u8{});
}
```

LemonBoy:
Don't cut any corner and properly run the type trough every single step
even though it has no fields (or, better, the sum of the size of all its
fields is zero).

Fix the logic to consider an explicit non-zero-sized tag enough to treat
the type as sized.

Closes #7451

@ifreund please try rebuilding river with this patch as it replaces #7222

### DIFF
--- a/src/stage1/analyze.cpp
+++ b/src/stage1/analyze.cpp
@@ -2550,8 +2550,8 @@ static Error resolve_union_alignment(CodeGen *g, ZigType *union_type) {
             union_type->data.unionation.gen_tag_index = 1;
         }
     } else {
-        assert(most_aligned_union_member != nullptr);
-        union_type->abi_align = most_aligned_union_member->align;
+        union_type->abi_align = most_aligned_union_member?
+                most_aligned_union_member->align : 0;
         union_type->data.unionation.gen_union_index = SIZE_MAX;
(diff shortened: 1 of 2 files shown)


ifreund:
> @ifreund please try rebuilding river with this patch as it replaces #7222

Confirmed that river still builds, thanks for the ping.

PR #7455, opened 2020-12-15
Fix system library path detection on linux

xackus:
Detect native abi properly when statically linked.

### DIFF
--- a/lib/std/zig/system.zig
+++ b/lib/std/zig/system.zig
@@ -15,8 +15,6 @@ const Target = std.Target;
 const CrossTarget = std.zig.CrossTarget;
 const macos = @import("system/macos.zig");
 
-const is_windows = Target.current.os.tag == .windows;
-
 pub const getSDKPath = macos.getSDKPath;
 
 pub const NativePaths = struct {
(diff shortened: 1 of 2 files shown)


lib/std/zig/system.zig
```diff
@@ -89,7 +89,13 @@ pub const NativePaths = struct {
         }
 
         if (!is_windows) {
-            const triple = try Target.current.linuxTriple(allocator);
+            // on linux the native abi might be different if we are statically linked
+            const target = if (Target.current.os.tag == .linux)
+                (try NativeTargetInfo.detectAbiAndDynamicLinker(allocator, Target.current.cpu, Target.current.os, .{})).target
```
andrewrk, line 94: This function should have the target passed to it, because in practice, the calling code is also going to call `detectAbiAndDynamicLinker` to obtain a target, so this will effectively cause the detection to happen twice.

LemonBoy:
I think a better solution is to add an extra parameter to `NativePaths.detect` called `native_target: NativeTargetInfo`.

lib/std/zig/system.zig
```diff
@@ -89,7 +89,13 @@ pub const NativePaths = struct {
...
+            else
+                Target.current;
+
+            const triple = try target.linuxTriple(allocator);
```
LemonBoy, line 98: `linuxTriple` is a pretty bad name for a function that has nothing to do with Linux heh
This code path is executed for every non-Windows target...

Also, the uses of `Target.current` below can be replaced with the newly introduced parameter, for consistency sake.

xackus:
> 
> 
> I think a better solution is to add an extra parameter to `NativePaths.detect` called `native_target: NativeTargetInfo`.

Good idea, this will prevent passing `Target.current` by mistake.

andrewrk:
Nice work! Sorry for the delay in merging.

andrewrk:
Landed in e4b8148e9c0dc8e1e523965456a59d469e93a26c.

Issue #7398, opened 2020-12-11
lld-link: error: could not open 'xinput1_4.lib': no such file or directory


xtian:
I am getting a similar error to #7395 when I try to link to `xinput1_4.dll` on native Windows.

```zig
// repro.zig
const win = @import("std").os.windows;

pub fn main() !u8 {
    var controller_state: XINPUT_STATE = undefined;

    _ = XInputGetState(0, &controller_state);

    return 0;
}

extern "xinput1_4" fn XInputGetState(
    dwUserIndex: win.DWORD,
    pState: *XINPUT_STATE,
) callconv(win.WINAPI) win.DWORD;

const XINPUT_STATE = extern struct {
    dwPacketNumber: win.DWORD,
    Gamepad: extern struct {
        wButtons: win.WORD,
        bLeftTrigger: win.BYTE,
        bRightTrigger: win.BYTE,
        sThumbLX: win.SHORT,
        sThumbLY: win.SHORT,
        sThumbRX: win.SHORT,
        sThumbRY: win.SHORT,
    },
};
```

```
> zig version
0.7.0+a1d3886ff

> zig run .\repro.zig --verbose-link
lld-link -ERRORLIMIT:0 -NOLOGO -DEBUG -STACK:16777216 -MACHINE:X64 -OUT:C:\Users\xtian\AppData\Local\zig\o\f375d882eaf1d453229d31861f2f2db6\repro.exe C:\Users\xtian\AppData\Local\zig\o\f375d882eaf1d453229d31861f2f2db6\repro.obj -SUBSYSTEM:console -NODEFAULTLIB -ENTRY:wWinMainCRTStartup C:\Users\xtian\AppData\Local\zig\o\deb4ff7197a5e1d3c9665624e8d64245\c.lib C:\Users\xtian\AppData\Local\zig\o\8a2b6260e3ff669d975589ee6d2d393f\compiler_rt.lib C:\Users\xtian\AppData\Local\zig\o\afd5139c5473062e83f5549830b6ae6c\kernel32.lib C:\Users\xtian\AppData\Local\zig\o\2b0ba140a5eec46d69f5a131a9665d50\ntdll.lib xinput1_4.lib
lld-link: error: could not open 'xinput1_4.lib': no such file or directory
error: LLDReportedFailure
```

LemonBoy:
[...]

PR #7458, opened 2020-12-15
mingw-w64: Add .def files for xaudio2_8

xtian:
Similar to #7398, I need `.def` files for [XAudio2](https://docs.microsoft.com/en-us/windows/win32/xaudio2/xaudio2-versions) added. However, the files from the mingw-w64 project (attached to this PR) don't have all the symbols I need. XAudio2 [makes heavy use of opaque types via interfaces](https://github.com/mirror/mingw-w64/blob/3419b2d4b2b7e8b378696dc79546e3593f00ade6/mingw-w64-headers/include/xaudio2.idl#L937) which are not present in the current `.def` files.

Is there a procedure for updating/modifying these files within the Zig project?

Here's an example of what I'm trying to do:
```zig
const win = @import("std").os.windows;

pub fn main() !u8 {
    var xaudio: IXAudio2 = undefined;
    var master_voice: IXAudio2MasteringVoice = undefined;

    _ = XAudio2Create(&xaudio, 0, 0x00000001);
    _ = IXAudio2_CreateMasteringVoice(xaudio, &master_voice, 2, 48_000, 0, null, null, 0);

    return 0;
}

const IUnknown = *opaque {};
const IXAudio2 = *opaque {};
const IXAudio2MasteringVoice = *opaque {};

extern "xaudio2_8" fn XAudio2Create(
    ppXAudio2: *IXAudio2,
    Flags: u32,
    XAudio2Processor: u32,
) callconv(win.WINAPI) win.HRESULT;

extern "xaudio2_8" fn IXAudio2_CreateMasteringVoice(
    pXAudio2: IXAudio2,
    ppMasteringVoice: *IXAudio2MasteringVoice,
    InputChannels: u32,
    InputSampleRate: u32,
    Flags: u32,
    szDeviceId: ?win.LPCWSTR,
    pEffectChain: ?[*]XAUDIO2_EFFECT_CHAIN,
    StreamCategory: i32,
) callconv(win.WINAPI) win.HRESULT;

const XAUDIO2_EFFECT_CHAIN = extern struct {
    EffectCount: u32,
    pEffectDescriptors: [*]XAUDIO2_EFFECT_DESCRIPTOR,
};

const XAUDIO2_EFFECT_DESCRIPTOR = extern struct {
    pEffect: IUnknown,
    InitialState: bool,
    OutputChannels: u32,
};
```


andrewrk:
> However, the files from the mingw-w64 project (attached to this PR) don't have all the symbols I need

Are you sure those are actually exposed directly by the DLL? It sounds to me like this is some kind of COM thing where you have to mess around with memory in order to get the function pointers that you need.

xtian:
@andrewrk That could be. I am pretty ignorant of how this stuff works. I just assumed the symbol was missing because I saw some `IFoo_Whatever` definitions in other `.def` files.

Issue #419, opened 2017-08-17
Could NOT find LLD (missing:  LLD_LIBRARIES) 


igalic:
I'm trying to build zig on Ubuntu 17.04 (zesty). I have llvm and clang installed from apt.llvm.org:

```
ii  clang-4.0                   1:4.0~svn305264-1~exp1 amd64        C, C++ and Objective-C compiler (LLVM based)
ii  clang-4.0-doc               1:4.0~svn305264-1~exp1 all          C, C++ and Objective-C compiler (LLVM based) - Documentation
ii  clang-format-4.0            1:4.0~svn305264-1~exp1 amd64        Tool to format C/C++/Obj-C code
ii  libclang-4.0-dev            1:4.0~svn305264-1~exp1 amd64        clang library - Development package
ii  libclang-common-4.0-dev     1:4.0~svn305264-1~exp1 amd64        clang library - Common development package
ii  libclang1-4.0:amd64         1:4.0~svn305264-1~exp1 amd64        C interface to the clang library
ii  libclang1-4.0-dbg:amd64     1:4.0~svn305264-1~exp1 amd64        clang library
ii  liblld-4.0-dev              1:4.0~svn305264-1~exp1 amd64        LLVM-based linker, header files
ii  liblldb-4.0                 1:4.0~svn305264-1~exp1 amd64        Next generation, high-performance debugger, library
ii  liblldb-4.0-dev             1:4.0~svn305264-1~exp1 amd64        Next generation, high-performance debugger, header files
ii  libllvm-4.0-ocaml-dev       1:4.0~svn305264-1~exp1 amd64        Modular compiler and toolchain technologies, OCaml bindings
ii  libllvm3.8:amd64            1:3.8.1-18ubuntu1      amd64        Modular compiler and toolchain technologies, runtime library
ii  libllvm4.0:amd64            1:4.0~svn305264-1~exp1 amd64        Modular compiler and toolchain technologies, runtime library
ii  libllvm4.0:i386             1:4.0~svn305264-1~exp1 i386         Modular compiler and toolchain technologies, runtime library
ii  libllvm4.0-dbg:amd64        1:4.0~svn305264-1~exp1 amd64        Modular compiler and toolchain technologies, debugging libraries
ii  lld-4.0                     1:4.0~svn305264-1~exp1 amd64        LLVM-based linker
ii  lldb-4.0                    1:4.0~svn305264-1~exp1 amd64        Next generation, high-performance debugger
[...]

PR #7466, opened 2020-12-16
Truncate user and group ids for 64 bit Linux systems

klingtnet:
Calls to `getuid`, `getgid` and their `eid` variants fail to compile on
64bit Linux systems because the return value of the syscall is of
`usize` and needs to be truncated to fit the size of `uid_t` that is 32
bit.

Thanks to @FireFox317 for figuring this out in Zig's Discord channel!

**TODO**

- [ ] squash commits
- [x] fix CI test
- [x] run regression test

### DIFF
--- a/lib/std/os/linux.zig
+++ b/lib/std/os/linux.zig
@@ -745,33 +745,33 @@ pub fn setregid(rgid: gid_t, egid: gid_t) usize {
 
 pub fn getuid() uid_t {
     if (@hasField(SYS, "getuid32")) {
-        return @truncate(uid_t, syscall0(.getuid32));
+        return @intCast(uid_t, syscall0(.getuid32));
     } else {
-        return @truncate(uid_t, syscall0(.getuid));
+        return @intCast(uid_t, syscall0(.getuid));
     }
 }
 
 pub fn getgid() gid_t {
     if (@hasField(SYS, "getgid32")) {
-        return @truncate(gid_t, syscall0(.getgid32));
+        return @intCast(gid_t, syscall0(.getgid32));
     } else {
-        return @truncate(gid_t, syscall0(.getgid));
+        return @intCast(gid_t, syscall0(.getgid));
     }
 }
 
 pub fn geteuid() uid_t {
     if (@hasField(SYS, "geteuid32")) {
-        return @truncate(uid_t, syscall0(.geteuid32));
+        return @intCast(uid_t, syscall0(.geteuid32));
     } else {
-        return @truncate(uid_t, syscall0(.geteuid));
+        return @intCast(uid_t, syscall0(.geteuid));
     }
 }
 
 pub fn getegid() gid_t {
     if (@hasField(SYS, "getegid32")) {
-        return @truncate(gid_t, syscall0(.getegid32));
+        return @intCast(gid_t, syscall0(.getegid32));
     } else {
-        return @truncate(gid_t, syscall0(.getegid));
+        return @intCast(gid_t, syscall0(.getegid));
     }
 }
 
--- a/lib/std/os/linux.zig
+++ b/lib/std/os/linux.zig
@@ -745,33 +745,33 @@ pub fn setregid(rgid: gid_t, egid: gid_t) usize {
 
 pub fn getuid() uid_t {
     if (@hasField(SYS, "getuid32")) {
-        return @as(uid_t, syscall0(.getuid32));
+        return @truncate(uid_t, syscall0(.getuid32));
     } else {
-        return @as(uid_t, syscall0(.getuid));
+        return @truncate(uid_t, syscall0(.getuid));
     }
 }
 
 pub fn getgid() gid_t {
     if (@hasField(SYS, "getgid32")) {
-        return @as(gid_t, syscall0(.getgid32));
+        return @truncate(gid_t, syscall0(.getgid32));
     } else {
-        return @as(gid_t, syscall0(.getgid));
+        return @truncate(gid_t, syscall0(.getgid));
     }
 }
 
 pub fn geteuid() uid_t {
     if (@hasField(SYS, "geteuid32")) {
-        return @as(uid_t, syscall0(.geteuid32));
+        return @truncate(uid_t, syscall0(.geteuid32));
     } else {
-        return @as(uid_t, syscall0(.geteuid));
+        return @truncate(uid_t, syscall0(.geteuid));
     }
 }
 
 pub fn getegid() gid_t {
     if (@hasField(SYS, "getegid32")) {
-        return @as(gid_t, syscall0(.getegid32));
+        return @truncate(gid_t, syscall0(.getegid32));
     } else {
-        return @as(gid_t, syscall0(.getegid));
+        return @truncate(gid_t, syscall0(.getegid));
     }
 }
 
--- a/lib/std/os/linux/test.zig
+++ b/lib/std/os/linux/test.zig
@@ -10,6 +10,7 @@ const mem = std.mem;
 const elf = std.elf;
 const expect = std.testing.expect;
 const expectEqual = std.testing.expectEqual;
+const getauxval = linux.getauxval;
 const fs = std.fs;
 
 test "fallocate" {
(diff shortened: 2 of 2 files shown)


lib/std/os/linux.zig
```diff
@@ -747,31 +747,31 @@ pub fn getuid() uid_t {
     if (@hasField(SYS, "getuid32")) {
         return @as(uid_t, syscall0(.getuid32));
     } else {
-        return @as(uid_t, syscall0(.getuid));
+        return @as(uid_t, @truncate(u32, syscall0(.getuid)));
```
FireFox317, line 750:
```suggestion
        return @truncate(uid_t, syscall0(.getuid));
```

lib/std/os/linux.zig
```diff
@@ -747,31 +747,31 @@ pub fn getuid() uid_t {
     if (@hasField(SYS, "getuid32")) {
         return @as(uid_t, syscall0(.getuid32));
```
FireFox317, line 748:
```suggestion
        return @truncate(uid_t, syscall0(.getuid32));
```
Could you fix these too?

FireFox317 (changes requested):
Thanks for the fast PR @klingtnet ! I have a couple suggestions.

lib/std/os/linux.zig
klingtnet: Thanks, wasn't sure where to place the truncate.  I adjusted all changes accordingly.

lib/std/os/linux.zig
klingtnet: As mentioned above, addressed all occurrences.

lib/std/os/linux.zig
klingtnet: Need, to remove the redundant `@as`, forgot about that.

klingtnet: Looks fine from my side now.

LemonBoy:
What about a nice test in `os/linux/test.zig` so that it won't accidentally regress?
Something like:

```
test "" {
    expectEqual(getauxval(AT_UID), getuid());
    // same for AT_GID, AT_EUID, AT_EGID
}
```

lib/std/os/linux.zig
```diff
@@ -745,33 +745,33 @@ pub fn setregid(rgid: gid_t, egid: gid_t) usize {
 
 pub fn getuid() uid_t {
     if (@hasField(SYS, "getuid32")) {
-        return @as(uid_t, syscall0(.getuid32));
+        return @truncate(uid_t, syscall0(.getuid32));
```
daurnimator, line 748: Shouldn't this be an `@intCast`? We don't actually expect values larger than `uid_t`.

daurnimator (changes requested):

klingtnet:
> What about a nice test in `os/linux/test.zig` so that it won't accidentally regress?
> Something like:
> 
> ```
> test "" {
>     expectEqual(getauxval(AT_UID), getuid());
>     // same for AT_GID, AT_EUID, AT_EGID
> }
> ```

Sounds fine to me, will address this after work.

klingtnet:
I added a regression test as suggested, unfortunately I could not find any documentation on how to run them. 

If it does not exist already then there should be some documentation for new contributors.

LemonBoy:
> I added a regression test as suggested, unfortunately I could not find any documentation on how to run them.
> 
> If it does not exist already then there should be some documentation for new contributors.

You can simply run `zig test <your .zig file>`, that's much faster than running the whole test suite (that you can run with `zig build --build-file ../build.zig test-std`, YMMV).

klingtnet:
> > I added a regression test as suggested, unfortunately I could not find any documentation on how to run them.
> > If it does not exist already then there should be some documentation for new contributors.
> 
> You can simply run `zig test <your .zig file>`, that's much faster than running the whole test suite (that you can run with `zig build --build-file ../build.zig test-std`, YMMV).

Thank you very much!  Test's will probably fail with `0.7.1` so I need to wait until my fresh build is ready (required to install `llvm` and `lld`, luckily the latter one was mentioned in https://github.com/ziglang/zig/issues/419 ).

LemonBoy:
> The test does not compile with 0,7.1:

Yes, some tests are not meant to be run stand-alone. You want to run `lib/std/os.zig` here.

lib/std/os/linux.zig
klingtnet: Sounds reasonable, replaced `@truncate` with `@intCast`.

klingtnet:
So, to run the tests locally I needed to compile `zig` locally under Arch Linux.  I will list the important steps here for documentation purposes:

```sh
$ pacman -S llvm clang lld
# follow the wiki instructions
$ cmake .. -DZIG_PREFER_CLANG_CPP_DYLIB=True
# wait ... done
```

The test still does not compile:

```sh
$ build/zig test lib/std/os/linux/test.zig
./lib/std/os/linux/test.zig:105:17: error: use of undeclared identifier 'getauxval'
    expectEqual(getauxval(AT_UID), getuid());
                ^
```

I imported `getauxval` from `std.os.linux` and then it fails with:

```sh
$ build/zig test lib/std/os/linux/test.zig | clipit
./lib/std/os/linux/test.zig:105:17: error: use of undeclared identifier 'getauxval'
    expectEqual(getauxval(AT_UID), getuid());
                ^
```

Unfortunately, I could not find a reference to `AT_UID` to import it.

LemonBoy:
> I imported getauxval from std.os.linux and then it fails with:

You can use `linux.getauxval` instead of defining an extra const for something that's only used once.

> I imported getauxval from std.os.linux and then it fails with:

Those are defined in `std.elf`... It makes no sense, I know.

klingtnet:
Seems to work fine now:

```sh
$ build/zig test lib/std/os/linux/test.zig
All 5 tests passed.
s1 alinz@kn-portable
```

klingtnet:
I need to fix the [failing CI test](https://dev.azure.com/ziglang/0184afc2-b8d3-409b-8e4f-4032aaca4c96/_build/results?buildId=11195):

```
1081/1637 os.linux.test.test "std-native-Debug-c-multi user and group ids"... expected 0, found 1001
/home/vsts/work/1/s/lib/std/testing.zig:74:32: 0x44178f in testing.expectEqual (test)
                std.debug.panic("expected {}, found {}", .{ expected, actual });
                               ^
/home/vsts/work/1/s/lib/std/os/linux/test.zig:105:16: 0x38807e in os.linux.test.test "std-native-Debug-c-multi user and group ids" (test)
    expectEqual(linux.getauxval(elf.AT_UID), linux.getuid());
               ^
/home/vsts/work/1/s/lib/std/special/test_runner.zig:61:28: 0x3c0a58 in std.special.main (test)
        } else test_fn.func();
                           ^
/home/vsts/work/1/s/lib/std/start.zig:344:37: 0x3c27a1 in start.main (test)
                                    ^
/home/vsts/work/1/s/zig-cache/o/590cc85149a19cdf50140d5408f6d1bb/test
/home/vsts/work/1/s/build/zig test /home/vsts/work/1/s/lib/std/std.zig -lc --test-name-prefix std-native-Debug-c-multi  --cache-dir /home/vsts/work/1/s/zig-cache --global-cache-dir /home/vsts/.cache/zig --name test -I /home/vsts/work/1/s/test --override-lib-dir /home/vsts/work/1/s/lib 
/home/vsts/work/1/s/zig-cache/o/b44c2f8e472997c666788c7413e997b6/build /home/vsts/work/1/s/build/zig /home/vsts/work/1/s /home/vsts/work/1/s/zig-cache /home/vsts/.cache/zig test -Denable-qemu -Denable-wasmtime
```

FireFox317:
The problem is in the `getauxval` function:
https://github.com/ziglang/zig/blob/506af7e52e0985b410ea089bf5fa3247ab2377cb/lib/std/os/linux.zig#L37-L49

As you can see the `elf_aux_maybe` is populated by the `start.zig` file and I'm pretty sure this startup code is not run when we do `zig test`. So it looks like we cannot test this (or somebody else knows a solution to this).

LemonBoy:
Oh this is some other `std.os` bullshit, slap a `if (builtin.link_libc) return error.SkipZigTest;` and let's call it a day, this is a test for our direct syscall wrappers after all.

PR #7447, opened 2020-12-14
std: non-byte-multiple sized integers have no definite representation

LemonBoy:
Closes #7445

PR #7472, opened 2020-12-17
std: hash_map: Better handling of no available slots case

### LINKED ISSUE #7468: Prevent interleaving calls to HashMap.removeAssertDiscard() / HashMap.putAssumeCapacity() from causing the hashmap to run out of memory
The test passes for `ArrayHashMap` and otherwise causes an integer overflow on the available number of `Entry`'s that may be reclaimed.

```
Test [1/1] test "HashMapUnmanaged.removeAssertDiscard() / HashMapUnmanaged.putAssumeCapac... integer overflow
/home/kenta/Desktop/zig/lib/std/hash_map.zig:621:32: 0x20b69b in std.hash_map.HashMapUnmanaged(usize,usize,std.hash_map.getAutoHashFn(usize).hash,std.hash_map.getAutoEqlFn(usize).eql,80).getOrPutAssumeCapacity (test)
                self.available -= 1;
                               ^
/home/kenta/Desktop/zig/lib/std/hash_map.zig:475:52: 0x207064 in std.hash_map.HashMapUnmanaged(usize,usize,std.hash_map.getAutoHashFn(usize).hash,std.hash_map.getAutoEqlFn(usize).eql,80).putAssumeCapacity (test)
            const gop = self.getOrPutAssumeCapacity(key);
                                                   ^
/home/kenta/Desktop/sleepy/test.zig:17:30: 0x2068c5 in test "HashMapUnmanaged.removeAssertDiscard() / HashMapUnmanaged.putAssumeCapacity()" (test)
        map.putAssumeCapacity(i, 0);
```

```zig
const std = @import("std");
const testing = std.testing;

test "HashMapUnmanaged.removeAssertDiscard() / HashMapUnmanaged.putAssumeCapacity()" {
    const allocator = testing.allocator;

    var map: std.AutoHashMapUnmanaged(usize, usize) = .{};
    defer map.deinit(allocator);

    try map.ensureCapacity(allocator, 10);

    var i: usize = 0;
    while (i < 20) : (i += 1) {
        if (i >= 10) {
            map.removeAssertDiscard(i - 10);
        }
        map.putAssumeCapacity(i, 0);
    }
}
```

LemonBoy:
Whenever self.available reaches zero we're not allowed to use free
slots, enforce this constraint while looking for a usable one.

Closes #7468

We need a test case that's independent of the hash function being used, I'll see if I can come up with something nice later.

### DIFF
--- a/lib/std/hash_map.zig
+++ b/lib/std/hash_map.zig
@@ -539,9 +539,10 @@ pub fn HashMapUnmanaged(
             const mask = self.capacity() - 1;
             const fingerprint = Metadata.takeFingerprint(hash);
             var idx = @truncate(usize, hash & mask);
+            var probe_length = self.capacity();
 
             var metadata = self.metadata.? + idx;
-            while (metadata[0].isUsed() or metadata[0].isTombstone()) {
+            while (probe_length > 0) : (probe_length -= 1) {
                 if (metadata[0].isUsed() and metadata[0].fingerprint == fingerprint) {
                     const entry = &self.entries()[idx];
                     if (eqlFn(entry.key, key)) {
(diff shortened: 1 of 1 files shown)


LemonBoy:
So... what the fuck is wrong with the Windows CI? Both this and #7447 are failing due to some OOM error that I don't see on master...

Sahnvour:
@LemonBoy Please see my answer in #7468 and let me know what you think, I currently think this it not the good way to fix the problem. I don't have time to look at it in depth at the moment however 😦

LemonBoy:
Yes, in #7494 I was wondering if this was just be a documentation problem.
TBH the `assumeCapacity` naming is a bit misleading as there's an hidden dependence on the load factor.

Issue #6710, opened 2020-10-17
simplify translate-c implementation by introducing a new pseudo-ast data structures


andrewrk:
Right now working on translate-c code is awkward because we have to simultaneously create AST nodes, which is a tree structure, while outputting a list of tokens and source in a linear order. If we had a more specialized AST data structure for translate-c, we could generate the token list and source from it instead of having to duplicate the logic.

It ends up causing bugs like this:  #6707

And this hack:

https://github.com/ziglang/zig/blob/245d98d32dd29e80de9732f415a4731748008acf/lib/std/zig/ast.zig#L26-L28

To explain another way, the current process is:
 * clang analysis => zig AST + source + token list => std.zig.render

Proposed process is:
  * clang analysis => TranslateCNode Tree => zig AST + source + token list => std.zig.render

The translate-c node tree data structure does not need to map to zig AST, it just has to be whatever is convenient for the translate-c implementation.

This probably would have a small performance penalty; the main objective with this proposal is to increase maintainability.

cc @Vexu

tadeokondrak:
Is it worth having any sort of tree over string output?

andrewrk:
In my experience the tree structure feels helpful for the implementation, and the linear string output is the tricky part to manage.

LemonBoy:
A **typed** C AST is also useful to simplify the `boolToInt`/`intToBool` insertion logic and perform several other simplifications over the C code.

Vexu:
This would definitely make translate-c easier to maintain by removing all the strange `appendToken`s needed to please Zig fmt. However generating two ASTs and adding some more type analysis seems like it could be more than a small performance penalty with something like `zig translate-c /usr/include/epoxy/gl.h -I/usr/include` already taking ~22s.

LemonBoy:
> it could be more than a small performance penalty with something like zig translate-c /usr/include/epoxy/gl.h -I/usr/include already taking ~22s.

Did you profile it? What are the bottlenecks?
Keep in mind that if the caching system works as expected the overall impact is near to zero.

pixelherodev:
It'd also let us write a C-AST -> ZIR path for stage2 and allow `zig cc` to work without LLVM ;)

pixelherodev:
> However generating two ASTs and adding some more type analysis seems like it could be more than a small performance penalty with something like zig translate-c /usr/include/epoxy/gl.h -I/usr/include already taking ~22s.

If it gets slower, but _more correct_, then that is performance well spent.

Issue #3453, opened 2019-10-14
Successfully @cImport("Windows.h")


JesseRMeyer:
A substantial milestone for translate-c is the successful c-importing of Windows.h.

JesseRMeyer:
I've given this another go with 0.5.0+c4770e7aa.

```zig
const windows = @cImport({
    @cDefine("WIN32_LEAN_AND_MEAN", "1");
    @cInclude("Windows.h");
});

usingnamespace windows;

extern "user32" fn MessageBoxA(hWnd: ?HANDLE, lpText: ?LPCTSTR, lpCaption: ?LPCSTSTR, uType: UINT) callconv(.Stdcall) c_int;

export fn WinMain(hInstance: HINSTANCE, hPrevInstace: HINSTANCE, lpCmdLine: PWSTR, nCmdShow: INT) callconv(.Stdcall) INT {
    MessageBoxA(null, "Hello", "title", 0);
}
```

```
PS C:\dev\src> zig build-exe test.zig
.\test.zig:1:17: error: C import failed
const windows = @cImport({
                ^
.\test.zig:1:17: note: libc headers not available; compilation does not link against libc
const windows = @cImport({
                ^
.\zig-cache\o\X31yopM03uxmIKZiqh1rP0rnHeAJdDfiWqbdFMfmHxY6oIsokRfiAiUhWLfJ4T-0\cimport.h:2:10: note: 'Windows.h' file not found
#include <Windows.h>
         ^
.\test.zig:6:16: note: referenced here
usingnamespace windows;
               ^
C:\dev\Zig\lib\zig\std\builtin.zig:443:73: note: referenced here
pub const panic: PanicFn = if (@hasDecl(root, "panic")) root.panic else default_panic;
                                                                        ^
```

I'm not sure what libc headers have to do with finding and importing Windows.h.

travisstaloch:
did you try `zig build-exe -lc test.zig` ?

JesseRMeyer:
I have not, but why should that be required to include Windows.h?

travisstaloch:
I'm guessing windows.h uses some things from libc.  I tried cIncluding a minimal .h file with no #includes and was able to build-exe without -lc

JesseRMeyer:
I'm sure Windows.h does use CRT stuff, but could you explain why that matters for _importing_?

[...]

PR #7479, opened 2020-12-17
Make translate-c use intermediate AST

Vexu:
Will implement  #6710


Sirius902:
Will this by chance address #3453?

andrewrk:
Still working on this? Is there perhaps a way you can start the work in such a way that it doesn't have to be 100% completed before we can start utilizing it? That way we don't have a PR sitting around bit rotting.

andrewrk:
Closing this old draft - the code is still available and linked from the respective issue description for anyone to reference it. When we tackle this again let's try to do it incrementally so that the whole rework/refactor does not have to get done before we can start merging into master branch.

Vexu:
This currently doesn't affect `translate-c` and could be merged as is for aomeone to continue, ci failure was probably unrelated.

andrewrk:
Ah I see, I will resurrect it and merge then!

andrewrk:
This is going to make #7920 much easier to update translate-c because the only thing that will have change is the render() function.

andrewrk:
Looks like the rebase got messed up

src/translate_c/ast.zig
```diff
@@ -0,0 +1,267 @@
+const std = @import("std");
+const Type = @import("../type.zig").Type;
+
+pub const Node = struct {
```
andrewrk, line 4: Looks like this is meant to be a union, yeah?

Vexu: Yup, haven't yet gotten this to compile.

andrewrk:
@Vexu whenever you feel this is ready to merge into `ast-memory-layout`, go for it

ehaas:
There is one error happening on this branch that isn't happening with the old translate-c - duplicate non-translatable macros are getting emitted twice as `@compileError`.

```c
#define FOO 5=2
#define FOO 5=2
```
produces
```zig
pub const FOO = @compileError("unable to translate C expr: unexpected token .Equal"); // ddef.c:4:9
pub const FOO = @compileError("unable to translate C expr: unexpected token .Equal"); // ddef.c:5:9
```

Previously only one `FOO` was emitted. This happens to me on MacOS 11.2 when I `#include <stdlib.h>` due to duplicate definitions of `__API_AVAILABLE_PLATFORM_macCatalyst`.

andrewrk:
One thing that will be interesting to test out on this branch, once we get it compiling again, is this benchmark. Here it is on master branch, on my machine:

#### test.c

```c
#include <windows.h>
```

(wipe global & local zig caches)

```
$ perf stat -e cache-misses -e instructions -e cycles ./zig translate-c test.c -target x86_64-windows-gnu -lc >windows_h.zig

 Performance counter stats for './zig translate-c test.c -target x86_64-windows-gnu -lc':

         6,632,603      cache-misses:u                                              
    40,151,754,610      instructions:u            #    1.39  insn per cycle         
    28,980,858,981      cycles:u                                                    

       6.669550308 seconds time elapsed

       6.539106000 seconds user
       0.132778000 seconds sys
```

Issue #513, opened 2017-10-01
Proposal: testing @compileError


PavelVozenilek:
Minor feature to test edge cases.

Zig allows custom made compiler errors.  Example is here:
http://andrewkelley.me/post/zig-programming-language-blurs-line-compile-time-run-time.html

There should be some easy way to check if they are implemented correctly. 

----------

My proposal:

```
test "expected to fail"
{
  something_generating_error();  /// DOES NOT COMPILE
}
```

If compiler reaches `@compileError` inside a test it should then look for comment `DOES NOT COMPILE` at the offending line. If it finds it there, all is OK, testing continues. If it doesn't, original error is reported, it is a bug to be fixed.

If the test compiles OK and but contains `DOES NOT COMPILE` comment, then it should fail.

---------

Alternative is special keyword for failing test, but this feels as overkill. 

Feature should be limited to user written `@compileError` only: syntactic errors (like unbalanced parenthesis) should be always error, wrong use of the language (`x = 1 + "abc"`) as well.

`DOES NOT COMPILE` comment outside a test means nothing.

Expected failed test would generate no code.

Failing tests could be certainly implemented as separately compiled files, but this is nuisance most people would avoid.

thejoshwolfe:
`@compileError()` takes a msg parameter, so testing for such an error should also take a msg that must match. How about:

```zig
test "fork" {
    switch(builtin.os) {
        Os.linux, Os.darwin, Os.macosx, Os.ios => testFork(),
        else => @expextCompileError("Unsupported OS", testFork()),
    };
}
```

PavelVozenilek:
One small improvement:  compiler should also remove the line annotated as failing, and _then_ try to compile (but not run) the remaining code. This would eliminate mistakes that would otherwise go unnoticed.


So this:
```
test "expected to fail"
{
  something_generating_error();  /// DOES NOT COMPILE
}

```

would be internally transformed into two tests:

```
test "expected to fail"
{
  xyz
  something_generating_error();  /// DOES NOT COMPILE
}

test "expected to compile, but not executed"
{
  xyz
}
```

marler8997:
I had made a duplicate of this proposal here: https://github.com/ziglang/zig/issues/3144

Along with testing compile errors, its important to note that this feature would also enable much more powerful generic specialization.  With this builtin function, generics can now test types for any feature that does or does not compile. I've included some contrived examples to demonstrate some of the capabilities this enables:

[...]

PR #7483, opened 2020-12-18
Make hasUniqueRepresentation false for slices

### LINKED ISSUE #7330: std.AutoHashMap allows string keys
I think this is a bug, feel free to close if this is intentional.  The `autoHash` function contains this check, which seems like it is meant to disallow string keys in AutoHashMap:
```zig
if (comptime meta.trait.isSlice(Key)) {
    comptime assert(@hasDecl(std, "StringHashMap")); // detect when the following message needs updated
    const extra_help = if (Key == []const u8)
        " Consider std.StringHashMap for hashing the contents of []const u8."
    else
        "";

    @compileError("std.auto_hash.autoHash does not allow slices (here " ++ @typeName(Key) ++
        ") because the intent is unclear. Consider using std.auto_hash.hash or providing your own hash function instead." ++
        extra_help);
    }
}
```
However, before running this code, it does this check:
```zig
if (comptime trait.hasUniqueRepresentation(K)) {
    return Wyhash.hash(0, std.mem.asBytes(&key));
} else {
     var hasher = Wyhash.init(0);
     autoHash(&hasher, key);
     return hasher.final();
}
```
`trait.hasUniqueRepresentation(..)` hits this case for slices:
```zig
.Pointer => return true,
```
So `autoHash` never gets compiled for the case where the key is a string, and the compile error does not trigger.

indocomsoft:
Fixes #7330 

## Questions
- Should I remove the corresponding slice check from `autoHash`?
- Is there any way to test that `@compileError` is emitted?

### DIFF
--- a/lib/std/hash/auto_hash.zig
+++ b/lib/std/hash/auto_hash.zig
@@ -169,21 +169,31 @@ pub fn hash(hasher: anytype, key: anytype, comptime strat: HashStrategy) void {
     }
 }
 
+fn typeContainsSlice(comptime K: type) bool {
+    comptime {
+        if (meta.trait.isSlice(K)) {
+            return true;
+        }
+        if (meta.trait.is(.Struct)(K)) {
+            inline for (@typeInfo(K).Struct.fields) |field| {
+                if (typeContainsSlice(field.field_type)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+}
+
 /// Provides generic hashing for any eligible type.
 /// Only hashes `key` itself, pointers are not followed.
-/// Slices are rejected to avoid ambiguity on the user's intention.
+/// Slices and structs containing slices are rejected to avoid ambiguity on the
+/// user's intention.
 pub fn autoHash(hasher: anytype, key: anytype) void {
     const Key = @TypeOf(key);
-    if (comptime meta.trait.isSlice(Key)) {
-        comptime assert(@hasDecl(std, "StringHashMap")); // detect when the following message needs updated
-        const extra_help = if (Key == []const u8)
-            " Consider std.StringHashMap for hashing the contents of []const u8."
-        else
-            "";
-
-        @compileError("std.auto_hash.autoHash does not allow slices (here " ++ @typeName(Key) ++
-            ") because the intent is unclear. Consider using std.auto_hash.hash or providing your own hash function instead." ++
-            extra_help);
+    if (comptime typeContainsSlice(Key)) {
+        @compileError("std.auto_hash.autoHash does not allow slices or structs containing slices here (" ++ @typeName(Key) ++
+            ") because the intent is unclear. Consider using std.auto_hash.hash or providing your own hash function instead.");
     }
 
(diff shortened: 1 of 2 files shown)


Vexu (changes requested):
The proper fix is to change `meta.trait.hasUniqueRepresentation` to return `.Pointer => |info| return info.size != .Slice,` for pointers.

> * Is there any way to test that `@compileError` is emitted?

Not yet: #513.

indocomsoft:
@Vexu fixed according to your comment.

I've also moved the check for `[]const u8` to AutoHashMap as @Sahnvour said in #7330. Let me know if I should move this change to another PR instead.

lib/std/hash/auto_hash.zig
```diff
@@ -169,21 +169,31 @@ pub fn hash(hasher: anytype, key: anytype, comptime strat: HashStrategy) void {
     }
 }
 
+fn typeContainsSlice(comptime K: type) bool {
+    comptime {
+        if (meta.trait.isSlice(K)) {
+            return true;
+        }
+        if (meta.trait.is(.Struct)(K)) {
```
Vexu, line 177: This check should also be done on unions with an enum tag since we also hash their fields.

indocomsoft:
@Vexu I've just added code to handle unions. I wonder if for `hasUniqueRepresentation` I should iterate through each union field and ensure that `@sizeOf(FieldType) == @sizeOf(T)` instead

lib/std/meta/trait.zig
```diff
@@ -490,6 +490,22 @@ pub fn hasUniqueRepresentation(comptime T: type) bool {
...
+            // Only packed and extern unions have guaranteed in-memory representation
+            switch (info.layout) {
+                else => return false,
+
+                .Extern, .Packed => {
+                    inline for (info.fields) |field| {
+                        const FieldType = field.field_type;
+                        if (comptime !hasUniqueRepresentation(FieldType)) return false;
+                    }
+                    return true;
```
Vexu, line 503: Unless all fields have the same bit size the union will contain padding bits for some fields, I think it is ok to just leave this as a false.

indocomsoft: Gotcha, I just did this.

lib/std/meta/trait.zig
```diff
@@ -529,10 +530,53 @@ test "std.meta.trait.hasUniqueRepresentation" {
...
+    };
+
+    testing.expect(hasUniqueRepresentation(TestUnion1));
+
+    const TestUnion2 = extern union {
+        a: u32,
+        b: u16,
+    };
+
+    testing.expect(hasUniqueRepresentation(TestUnion2));
```
Vexu, line 557: `!hasUniqueRepresentation` on both of these.

Vexu (approved):
Thanks for making the extra effort to do this properly.

lib/std/meta/trait.zig
indocomsoft: Ah, I missed these. Fixed.

Thank you for helping to review my PR!

PR #7486, opened 2020-12-18
std: Properly fix the TLS alignment problem

LemonBoy:
ad05509 introduced a fix for the wrong problem, the logic to align the
start of main_thread_tls_buffer was already there but was flawed.

Fix it for good and avoid wasting too many bytes for alignment purposes.


andrewrk:
Wouldn't it be better to simply rely on the `align(16)`? The generated machine code is probably more than 16 bytes, plus it is runtime code instead of free work done for us by the linker.

LemonBoy:
The fix is needed anyway as the alignment logic was broken, you can leave the `align(16)` if you want.

PR #7487, opened 2020-12-18
openbsd: implement segfault handling on openbsd x86_64

semarie:
to avoid making unbuildable zig on openbsd architecture without rewritten `ucontext_t` definition (which is architecture specific), I have made `have_segfault_handling_support` value dependant of `os.ucontext_t` existence on OpenBSD.

the `pub usingnamespace switch` in `os/bits.zig` permits to define constants (`ucontext_t`) based on targeted architecture or to omit the definition if not defined. Without it, `have_segfault_handling_support` will be false. It simplifies `handleSegfaultLinux()` function as it is used only on fully supported architectures for OpenBSD.

please note that `ucontext_t` on OpenBSD is hightly different from currently supported one (Linux and NetBSD).

cc @LemonBoy

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -1696,7 +1696,7 @@ fn getDebugInfoAllocator() *mem.Allocator {
 pub const have_segfault_handling_support = switch (builtin.os.tag) {
     .linux, .netbsd => true,
     .windows => true,
-    .freebsd => @hasDecl(os, "ucontext_t"),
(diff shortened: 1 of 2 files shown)


semarie:
build failed for Windows due to [LLVM ERROR: out of memory](https://dev.azure.com/ziglang/zig/_build/results?buildId=11239&view=logs&j=3e8797c7-5b0a-5f2c-07a4-1bc5e60a122e&t=c222f6c4-0121-56a3-9785-67eccb26a308&l=19892)

PR #7493, opened 2020-12-18
std/os/linux: Fix compilation of getuid/getgid/geteuid/getegid on x86_64

heidezomp:
Both uid_t and gid_t are 32 bits wide, so the syscall result needs to be truncated.


FireFox317:
Duplicate #7466

heidezomp:
> Duplicate #7466

Whoops, seems like I didn't search well enough when looking through existing PRs! #7466 looks good, so I'll close this one.

PR #7497, opened 2020-12-19
Enable segfault handling on FreeBSD

### LINKED ISSUE #5892: sa_sigaction doesn't build on FreeBSD

tetsuo-cpp:
Closes #5892.

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -1696,6 +1696,7 @@ fn getDebugInfoAllocator() *mem.Allocator {
 pub const have_segfault_handling_support = switch (builtin.os.tag) {
     .linux, .netbsd => true,
     .windows => true,
+    .freebsd => @hasDecl(os, "ucontext_t"),
     else => false,
 };
 pub const enable_segfault_handler: bool = if (@hasDecl(root, "enable_segfault_handler"))
@@ -1757,6 +1758,7 @@ fn handleSegfaultLinux(sig: i32, info: *const os.siginfo_t, ctx_ptr: ?*const c_v
 
     const addr = switch (builtin.os.tag) {
         .linux => @ptrToInt(info.fields.sigfault.addr),
+        .freebsd => @ptrToInt(info.addr),
         .netbsd => @ptrToInt(info.info.reason.fault.addr),
         else => unreachable,
     };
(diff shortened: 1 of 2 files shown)


tetsuo-cpp:
@LemonBoy This is the change we were discussing in #7409 to make `have_segfault_handling_support == true` for FreeBSD. The smoke test you added still looks healthy with this patch.

lib/std/debug.zig
```diff
@@ -1781,8 +1782,8 @@ fn handleSegfaultLinux(sig: i32, info: *const os.siginfo_t, ctx_ptr: ?*const c_v
         },
         .x86_64 => {
             const ctx = @ptrCast(*const os.ucontext_t, @alignCast(@alignOf(os.ucontext_t), ctx_ptr));
-            const ip = @intCast(usize, ctx.mcontext.gregs[os.REG_RIP]);
-            const bp = @intCast(usize, ctx.mcontext.gregs[os.REG_RBP]);
+            const ip = @intCast(usize, if (builtin.os.tag == .freebsd) ctx.mcontext.rip else ctx.mcontext.gregs[os.REG_RIP]);
```
semarie, line 1785: could you use a `switch (builtin.os.tag)` for `ip` and for `bp` ? it would simplify the merge with #7487 (where OpenBSD will need special case too)

tetsuo-cpp: Sure! Thanks for the heads up, I didn't see that PR.

tetsuo-cpp: Ok, done.

LemonBoy:
Looks great! We can solve the `mcontext_t` problem later, I think I like @semarie's approach where no compile error is triggered in case a non-x64 target is specified.

tetsuo-cpp:
> Looks great! We can solve the `mcontext_t` problem later, I think I like @semarie's approach where no compile error is triggered in case a non-x64 target is specified.

Ok done. I'm now using the same approach used in #7487.

PR #7498, opened 2020-12-19
stage2: add initial impl of LLVM backend in self-hosted compiler

FireFox317:
The following example compiles successfully in stage2 using `zig build-exe main.zig -fLLVM` (`--watch` works too!):
```zig
export fn _start() noreturn {
    foo();
    exit();
}

fn foo() void {}

fn exit() noreturn {
    unreachable;
}
```

I would appreciate some review especially on the structure of the code in `llvm_backend.zig`. I'm not sure if the current structure is good enough.


kubkon:
Just a drive-by comment for now, but is it intentional that this only touches `Elf` linker? Another thing, I think I'd prefer to keep the original file structure in place. For instance, I think having the functionality split between `src/codegen/llvm.zig` and `src/link/llvm.zig` is cleaner and carries less mental overhead when we want update/change things in the future. Maybe this is because I've been stuck in the linker for so long, but I'd expected any bit of functionality that updates decls in `src/link/xxx.zig`.

FireFox317:
@kubkon So initially I implemented it by adding a new `Tag` to the `File` struct in `link.zig`. However in the current architecture this does not make any sense, because LLVM is responsible for creating an object file and not for linking (Thus it shouldn't be a concept in `link.zig`). To hook it up to the current LLD linking architecture (which has separate paths per object format), the code has to be added to all the different linking formats. But it is just a matter of calling the same function in the `llvm_backend.zig` file, because the LLVM code is the same regardless.

Indeed, currently I only hooked it up to `Elf.zig` path, but it shouldn't be hard to add it to the `Coff` path too. (Which I will do now)

kubkon:
> @kubkon So initially I implemented it by adding a new `Tag` to the `File` struct in `link.zig`. However in the current architecture this does not make any sense, because LLVM is responsible for creating an object file and not for linking (Thus it shouldn't be a concept in `link.zig`). To hook it up to the current LLD linking architecture (which has separate paths per object format), the code has to be added to all the different linking formats. But it is just a matter of calling the same function in the `llvm_backend.zig` file, because the LLVM code is the same regardless.

Hmm, OK, in that case, would it make sense to put all relevant LLVM functionality in `llvm.zig` and `llvm/xxx.zig`? It'd be cool if there was a way to always refer to any relevant LLVM bits via `llvm.xxx` or something like that.

> Indeed, currently I only hooked it up to `Elf.zig` path, but it shouldn't be hard to add it to the `Coff` path too. (Which I will do now)

Don't forget about Mach-O please. Also, since this is stage2 after all, I'm toying with the idea of plugging in the new Mach-O backend instead of the old one. The new one is not feature complete but neither is stage2 and I think it would make sense to use the new one to weed out as many bugs as possible.

FireFox317:
@kubkon I added windows/pe/coff support in f0bcf4f2a2a533b80ec32955fc51b412bfcf5fd8. As you can see this was a simple patch.

FireFox317:
> Hmm, OK, in that case, would it make sense to put all relevant LLVM functionality in `llvm.zig` and `llvm/xxx.zig`? It'd be cool if there was a way to always refer to any relevant LLVM bits via llvm.xxx or something like that.

That is basically what I did. However the file is called `llvm_backend.zig` instead of `llvm.zig`. See 3d1f6b61ee1ef970694f224d9fc96fb00669762b for the reason why the file cannot be named `llvm.zig`. (Basically putting functions in that file will generate internal function names called `llvm.foo` and LLVM thinks these are intrinsics.) 

Also for now there was no need to split the `llvm_backend.zig` into multiple files yet, but in the future I think it makes sense to put the relevant files into a new dir `llvm_backend/xxx.zig`.

kubkon:
> > Hmm, OK, in that case, would it make sense to put all relevant LLVM functionality in `llvm.zig` and `llvm/xxx.zig`? It'd be cool if there was a way to always refer to any relevant LLVM bits via llvm.xxx or something like that.
> 
> That is basically what I did. However the file is called `llvm_backend.zig` instead of `llvm.zig`. See [3d1f6b6](https://github.com/ziglang/zig/commit/3d1f6b61ee1ef970694f224d9fc96fb00669762b) for the reason why the file cannot be named `llvm.zig`. (Basically putting functions in that file will generate internal function names called `llvm.foo` and LLVM thinks these are intrinsics.)

Ah, I see, thanks for the explanation!

src/link/Elf.zig
```diff
@@ -423,6 +437,8 @@ fn updateString(self: *Elf, old_str_off: u32, new_name: []const u8) !u32 {
 }
 
 pub fn populateMissingMetadata(self: *Elf) !void {
+    if (self.llvm_ir_module) |_| return;
```
Vexu, line 440: This seems like it should be an assertion since `populateMissingMetadata` is only called from `openPath` where we return early if we're using llvm.

Vexu (approved):

src/link/Elf.zig
FireFox317: Ah yes, thanks for noticing

src/llvm_bindings.zig
```diff
@@ -0,0 +1,507 @@
+//! We do this instead of @cImport because the self-hosted compiler is easier
+//! to bootstrap if it does not depend on translate-c.
+
+const std = @import("std");
+const assert = std.debug.assert;
+
+const LLVMBool = bool;
```
tadeokondrak, line 7: Not sure if this is intentionally different, but the LLVM C API's bool is a typedef to `int` as far as I can tell.

Issue #764, opened 2018-02-09
In/OutStream design flaw introduced with new error set changes.


Hejsil:
So, before error sets, we had In/OutStreams that could be implemented on a memory buffer, files or whatever would fit this interface, and could function to take In/OutStreams, and it would work on all of these.

Now, with error sets, I can see that In/Outstream was redesigned to be generic, taking an error set as the parameter. This ruins the useability of the In/OutStream:
```
error: expected type '&OutStream(error)', found '&OutStream(error{SystemResources,OperationAborted,IoPending,BrokenPipe,Unexpected,WouldBlock,FileClosed,DestinationAddressRequired,DiskQuota,FileTooBig,InputOutput,NoSpaceLeft,AccessDenied,})'
```

Now, you would have to pass the error type to all functions taking an In/OutStream, for it to work again:
```
fn foo(comptime Error: type, stream: &io.OutStream(Error)) !void { ... }
```

If this is the way they should now be used, then can I be sure that `foo` doesn't explode into a million instances of `foo`, as different OutStreams are used? Users of my function now also have to call it like this:

```
const file_stream = FileOutStream.init(&file);
foo(@typeOf(file_stream).Error, &file_stream.stream);
```

Which, idk, seems not so ergonomic for the users of my library. It relies on the fact, that the implementer of the OutStream adaptor was kind enough to actually provide this `Error` constant field. If the implementer did not do this, the user is more screwed, and would have to write this code:
```
const file_stream = FileOutStream.init(&file);
foo(@typeOf(file_stream.stream.writeFn).ReturnType.ErrorSet, &file_stream.stream);
```

Ooh god, plz no.

andrewrk:
Thanks for the early adoption. Let's get this sorted out.

> doesn't explode into a million instances of foo, as different OutStreams are used?

This aspect should be fine. This isn't c++ where we duplicate all the work in every file and then rely on the linker to toss out redundant definitions. LLVM actually has [an optimization pass](https://llvm.org/docs/Passes.html#mergefunc-merge-functions) that merges compatible function definitions together. We could potentially even notice when the difference between instantiations differs only in error sets and save time and space even in debug mode.

The API ergonomics and semantics on the other hand, let's talk about that.

Here are our options:
 * OutStream can become non-generic again and use `error` as the error set, the `void *` equivalent of error sets. It then "poisons" other error sets, because the union of any error set and `error` is `error`. (This would be the equivalent to how it was before error sets merge)
[...]

PR #7502, opened 2020-12-20
update depreciated code

g-w1:
* fix depreciated interface, update `outStream` -> `writer`
* make code more readable by updating `anytype` -> `std.fs.File.Writer`

Was there a reason for anytype? A compiler bug?

### DIFF
--- a/doc/docgen.zig
+++ b/doc/docgen.zig
@@ -42,7 +42,7 @@ pub fn main() !void {
 
     const input_file_bytes = try in_file.inStream().readAllAlloc(allocator, max_doc_file_size);
 
-    var buffered_out_stream = io.bufferedOutStream(out_file.outStream());
+    var buffered_out_stream = io.bufferedOutStream(out_file.writer());
 
     var tokenizer = Tokenizer.init(in_file_name, input_file_bytes);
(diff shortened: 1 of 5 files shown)


LemonBoy:
> Was there a reason for anytype? A compiler bug?

Fuck typing. You want to accept any kind of reader, be it from a `fs.File` or some other `io.reader`.

g-w1:
Ah, I had always heard to avoid anytype when possible. Should I switch it back?

g-w1:
I am also going to fix outStream -> writer throughout the whole codebase. When doing this in merge_anal_dumps.zig, it was heavily bitrotted (didn't compile), so I am fixing that.

andrewrk:
You're being told simultaneously to avoid anytype when possible and that it's needed for I/I streams. This dissonance is a recognized design flaw: #764

I don't have a solution for it yet; this wart in the language has survived for quite some time now.

g-w1:
Well for this, I think the only use is to print to stdout, so it is OK to have the type. Also when updating the depreciated interface, merge_anal_dumps.zig failed to compile. Is it even used anymore? It seems very bitrotted. I tried to fix it, but then realized that I shouldn't waste my time if its not used anymore.

PR #7504, opened 2020-12-20
Add c_longdouble mapping for bpf

mattnite:
I've been getting a c_longdouble mapping todo error in my CI and decided to do some digging since BPF does not have any floating point numbers.

### DIFF
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -8660,6 +8660,10 @@ static void define_builtin_types(CodeGen *g) {
         case ZigLLVM_msp430:
             add_fp_entry(g, "c_longdouble", 64, LLVMDoubleType(), &g->builtin_types.entry_c_longdouble);
             break;
+        case ZigLLVM_bpfel:
+        case ZigLLVM_bpfeb:
+            // eBPF does not have floating point numbers
+            break;
         default:
             zig_panic("TODO implement mapping for c_longdouble");
     }
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -8662,7 +8662,7 @@ static void define_builtin_types(CodeGen *g) {
             break;
         case ZigLLVM_bpfel:
         case ZigLLVM_bpfeb:
-            // eBPF does not have floating point numbers
+            add_fp_entry(g, "c_longdouble", 64, LLVMDoubleType(), &g->builtin_types.entry_c_longdouble);
             break;
         default:
             zig_panic("TODO implement mapping for c_longdouble");


LemonBoy:
Leaving that undefined is ... dangerous?
What is the size of othrr floating point types for bpf?

mattnite:
BPF does not have floating point numbers. From the BPF manpage:

> eBPF programs can be written in a restricted C that is compiled (using the clang compiler) into eBPF bytecode.  **Various features are omitted** from this restricted C, such as loops, global variables, variadic functions, **floating-point numbers**, and passing structures as function arguments.

Is there an `add_fp_entry` that can return a decent compile error?

LemonBoy:
I'll rephrase the question, what does `@sizeOf(T)` for `T=f32,f64`?

andrewrk:
The zig lang spec will guarantee that f16, f32, f64 work even if there are no hardware instructions for it. In such case the implementations go inside compiler-rt. Byte size of these types are 2, 4, 8, respectively, on all targets.

As for `c_longdouble`, this type is defined to match the main C compiler described by the target triple. For example, for `any-linux-gnu`, `c_longdouble` is defined to match `long double` from gcc on this target. For `any-windows-msvc`, `c_longdouble` is defined to match `long double` from MSVC on this target. (cc @SpexGuy)

I tested the following source file:

```c
long double foo(long double arg) {
    return arg;
}
```

```
./zig cc -c test.c -target bpfel-freestanding-gnu
```

It compiles without complaint. So it seems at least clang makes this type work, and clang matches gcc. So for `bpfel-freestanding-gnu` at least, `c_longdouble` should match clang. I did a follow-up test:

```c
unsigned foo(void) {
    return sizeof(long double);
}
```

```
./zig cc -c test.c -target bpfel-freestanding-gnu
$ ~/local/llvm11-release/bin/llvm-objdump -d test.o

test.o:	file format elf64-bpf


Disassembly of section .text:

0000000000000000 <foo>:
       0:	b7 00 00 00 08 00 00 00	r0 = 8
       1:	95 00 00 00 00 00 00 00	exit

```

So we have size of long double 8 bytes on this target.

However I do notice that if we try to do math, clang fails:

```c
long double add(long double a, long double b) {
    return a + b;
}
```

```
$ ./zig cc -c test.c -target bpfel-freestanding-gnu
test.c:1:13: error: A call to built-in function '__adddf3' is not supported.
long double add(long double a, long double b) {
            ^
1 error generated.
```

For the zig language spec, I think we should define all operations to always work on all the primitive types, even if it has to resort to compiler-rt.

mattnite:
Ok so instead of leaving this undefined, all I need to do is map it to a regular 64 bit floating point number, and then let some other component of the compiler handle it, brb

PR #7507, opened 2020-12-20
stage2 ARM: implement basic binary bitwise operations

joachimschmidt557:
A small intermezzo while I'm working on conditional branching in the stage2 ARM backend.

TODO in this PR:

- [ ] ~~implement comptime calculation~~ needs more work in `std.math.big.int` afaik
- [x] add test

### DIFF
--- a/src/codegen.zig
+++ b/src/codegen.zig
@@ -764,6 +764,8 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
                 .arg => return self.genArg(inst.castTag(.arg).?),
                 .assembly => return self.genAsm(inst.castTag(.assembly).?),
                 .bitcast => return self.genBitCast(inst.castTag(.bitcast).?),
+                .bitand => return self.genBitAnd(inst.castTag(.bitand).?),
+                .bitor => return self.genBitOr(inst.castTag(.bitor).?),
                 .block => return self.genBlock(inst.castTag(.block).?),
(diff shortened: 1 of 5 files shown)


joachimschmidt557:
As far as I have understood the semantics of comptime bitwise operations, the comptime calculation bit might be non-trivial here. `comptime_int` relies on BigInt functionality. In the stage1 compiler, bitwise operations transform the BigInts into twos-complement and transform them back into BigInts. On the other hand, the BigInt implementation of the standard library does not operate on twos-complement. Should the implementation of the standard library be changed or the twos-complement functionality be added as another option?

Rocknest:
Do you mean something like this `-5 | 6 == -1`? I think it should not work, only unambiguous operation should be allowed.

joachimschmidt557:
@Rocknest The current behaviour of the stage1 compiler is `-5 | 6 == 1`. Are there any issues/proposals to change this behaviour? I think a discussion on whether to keep, change or disallow this behaviour is out of scope of this PR. Unless a proposal was accepted/will be accepted that the current behaviour should change, we'll have to replicate the behaviour of stage1 in stage2.

Rocknest:
It feels wrong to automatically select some type for these kind of operations, just like negation of comptime_int. Its an opportunity to reconsider what is allowed. Unfortunately i dont have time to research this further.

Issue #7436, opened 2020-12-13
Does Zig use a stack red zone?


BinaryWarlock:
On some architectures, such as x86_64, ABIs can allow (or even mandate) a "red zone" which is a fixed size area below the stack pointer that is always valid to use, and must not be clobbered by the compiler or operating system. ([Here's an article](https://eli.thegreenplace.net/2011/09/06/stack-frame-layout-on-x86-64) explaining it in more detail.)

The SysV ABI for x86_64 mandates a 128-byte red zone:
> The 128-byte area beyond the location pointed to by %rsp is considered to be reserved and shall not be modified by signal or interrupt handlers. Therefore, functions may use this area for temporary data that is not needed across function calls. In particular, leaf functions may use this area for their entire stack frame, rather than adjusting the stack pointer in the prologue and epilogue. This area is known as the red zone.

It's used in GCC and Clang by default to optimize leaf functions by omitting bumping the stack pointer.

However, [it is customary for OS kernels to disable the red zone](https://wiki.osdev.org/Libgcc_without_red_zone) (in GCC/Clang by the flag `-mno-red-zone`), because if the compiler generates code clobbering the red zone, it can corrupt the stack in interrupt handlers causing many headaches, so this is an important use case. ([Further reading](https://stackoverflow.com/questions/25787408/why-cant-kernel-code-use-a-red-zone).)

I could not find any Zig documentation, or references in the source code, to this.

So two parts to this are:

- Does Zig currently make use of the red zone, specifically on the x86_64 arch targeting the SysV ABI (the Win64 ABI does not have a red zone)?

- If so, is there a way to disable using the red zone, as with `-mno-red-zone`, to prevent the compiler from clobbering below the stack pointer?

andrewrk:
Thanks for bringing this up! It led to the new CLI flags `-mred-zone` and `-mno-red-zone`. I believe the two questions are now answered: Zig is allowed to take advantage of red zones that are mandated by platform ABIs, unless you pass `-mno-red-zone` (new flag as of the merge of #7511 - thanks @leecannon).

PR #7511, opened 2020-12-20
Added support for no red zone

leecannon:
I have been writing a kernel in zig and never thought to check about the red zone until I saw #7436.

This pull requests adds the option to not use the red zone, the functionality is accessible in two ways:

- In `build.zig` set `no_red_zone` on `*LibExeObjStep` to true
- When calling `zig build-exe/lib/obj` pass `-fno-red-zone`

I have verified using `--verbose-llvm-ir` that `noredzone` is included in the function attributes.

This is the first time I have looked at the compiler internals so I'm sure I missed something, any feedback is appreciated.


LemonBoy:
You also want `-fred-zone` to complement `-fno-red-zone`.

leecannon:
Added `-fred-zone`

BinaryWarlock:
Awesome, this mirrors what GCC/Clang do and provides guarantees about it. :+1:, haven't tested it though.

leecannon:
@BinaryWarlock I made it private as there are parts of the design I'm not happy with. But I've split parts out into other repos.

andrewrk:
Thanks! Landed in 0edde40c25c033aec67f235365593ee26ff52740 with some minor adjustments.

PR #7515, opened 2020-12-21
Address deadlock in src/ThreadPool.zig

kprotty:
The current `ThreadPool` in src/ deadlocks when running on ARM so it was disabled and made single threaded (AFAICK). This is an attempt to fix this so that the ThreadPool can actually be used to speed up the C compilation.

**Current state:** After some testing, it looks like the deadlock might be a bug in `std.AutoResetEvent` as it affected other zig projects as well. For now, however, will try to make do with a ThreadPool that doesn't use it. There's still [effort in the works](https://gist.github.com/kprotty/0d2dc3da4840341d6ff361b27bdac7dc) for tackling the sync bug though.


FireFox317:
There have been some mentioned in Discord that the ThreadPool does not shutdown when a c object fails to build. @andrewrk

andrewrk:
You're going to want to throw this patch in the PR too:

```diff
diff --git a/ci/drone/linux_script b/ci/drone/linux_script
index 8c5dc1be2..fdc1704fb 100755
--- a/ci/drone/linux_script
+++ b/ci/drone/linux_script
@@ -17,8 +17,7 @@ git config core.abbrev 9
 
 mkdir build
 cd build
-# TODO figure out why Drone CI is deadlocking and stop passing -DZIG_SINGLE_THREADED=ON
-cmake .. -DCMAKE_BUILD_TYPE=Release "-DCMAKE_INSTALL_PREFIX=$DISTDIR" -DZIG_STATIC=ON -DCMAKE_PREFIX_PATH=/deps/local -GNinja -DZIG_SINGLE_THREADED=ON
+cmake .. -DCMAKE_BUILD_TYPE=Release "-DCMAKE_INSTALL_PREFIX=$DISTDIR" -DZIG_STATIC=ON -DCMAKE_PREFIX_PATH=/deps/local -GNinja
 
 samu install
 ./zig build test -Dskip-release -Dskip-non-native
```

src/ThreadPool.zig
```diff
@@ -113,14 +103,18 @@ fn runWorker(self: *ThreadPool) void {
             continue;
         }
 
-        if (!self.is_running) {
+        if (self.is_running) {
```
marler8997, line 106: Is it possible to shutdown the thread pool before all tasks are completed?  If so, wouldn't this check need to be done before executing the next task above?

andrewrk: The usage pattern in the self-hosted compiler will never call shutdown with running jobs. Even if it did, it would still be valid to have the code structured this way. I'm not concerned about this. I just want to figure out that damn deadlock on the drone CI.

marler8997: Ok, in that case we may want to document this since I think it would be unexpected.  Maybe something like this?
```
/// shutdown will cause the thread pool to stop, but only after all tasks have been completed
```

marler8997: Looking at the latest version, it looks like there isn't even a `shutdown` function anymore, so this comment may no longer apply.

src/ThreadPool.zig
```diff
@@ -30,49 +30,37 @@ pub fn init(self: *ThreadPool, allocator: *std.mem.Allocator) !void {
...
+    var num_threads = std.math.max(1, std.Thread.cpuCount() catch 1);
+    self.threads = try allocator.alloc(*std.Thread, num_threads);
 
     while (num_threads > 0) : (num_threads -= 1) {
         const thread = try std.Thread.spawn(self, runWorker);
-        self.threads[self.running] = thread;
-        self.running += 1;
+        self.threads[self.spawned] = thread;
+        self.spawned += 1;
     }
```
marler8997, line 40: In the loop both `num_threads` and `spawned` are being used as counters, but only one is needed.  `num_threads` can be const and we could just use `spawned` as the counter:
```zig
    const num_threads = std.math.max(1, std.Thread.cpuCount() catch 1);
    self.threads = try allocator.alloc(*std.Thread, num_threads);

    while (self.spawned < num_threads) : (self.spawned += 1) {
        const thread = try std.Thread.spawn(self, runWorker);
        self.threads[self.spawned] = thread;
    }
```

src/ThreadPool.zig
```diff
@@ -9,12 +9,12 @@ const ThreadPool = @This();
 lock: std.Mutex = .{},
```
marler8997, line 9: Good job on this.  The `ThreadPool` implementation and `WaitGroup` implementation are dead simple now.

andrewrk: Still getting a deadlock on Drone CI tho :(

LemonBoy:
I think the deadlock is caused by the ARM's weak memory model. I'd start by auditing all the atomic ops using the relaxed memory ordering.

andrewrk:
@LemonBoy over in the other PR I did I converted to using all pthreads abstractions (no atomic ops with relaxed memory) and still got the deadlock on Drone CI :-/

src/ThreadPool.zig
kprotty: Changed it, thanks.

andrewrk:
I strongly suspect the issue is not to do with multithreading but with the caching system. I expected https://github.com/ziglang/zig/pull/7520/commits/8219d92987c7d9641d6b24c4d5be29e3a80fd6b9 to fix it (by adding the `skip_linker_dependencies = true` to more `Compilation` objects), but alas it did not. My next idea is to do  this code, but globally for all cache hits in the entire process, not just c objects:

https://github.com/ziglang/zig/blob/0fd68f49e2eabb866ea1d21c4657c2a1d3c8ce53/src/Compilation.zig#L1840-L1857

The idea is that it would be debug code only, but it would identify a deadlock and print a helpful message rather than causing an actual deadlock on the flock() syscall.

andrewrk:
The deadlock was 2 different things:

 * An issue with the caching system, fixed in 8219d92987c7d9641d6b24c4d5be29e3a80fd6b9
 * std.ChildProcess calling malloc() between fork() and execv(), fixed in 3f9588ca29c721f11eb9dae8f6de8ebd1155c7bf

PR #7519, opened 2020-12-22
More pthreads integration

andrewrk:
* Introduce `sem_t` libc bits for Darwin, Linux, FreeBSD
 * `std.Mutex` implementation defaults to using pthreads mutex when `std.Thread.use_pthreads` is true
 * `std.ResetEvent` implementation defaults to using pthreads sem_t when `std.Thread.use_pthreads` is true
 * `std.valgrind` gains functions for annotating for the helgrind tool
 * Remove the `std.ResetEvent.isSet` abstraction and improve the tests to conform to expected API.

This incorporates changes from  #7515 so that we can see how this affects the situation.

The main motivation for these changes was using the [helgrind](http://valgrind.org/docs/manual/hg-manual.html) tool to detect race conditions in the multi-threaded code but it seems to be a dead-end with many false positives.


lib/std/valgrind.zig
```diff
@@ -80,6 +80,70 @@ pub const ClientRequest = extern enum {
     ChangeErrDisablement = 6145,
     VexInitForIri = 6401,
     InnerThreads = 6402,
+
+    CLEAN_MEMORY = 1212612608,
```
daurnimator, line 84: This don't go here: they should be in valgrind/helgrind.zig.
Also use the function `valgrind.ToolBase`.

See `CallgrindClientRequest` for an example.

daurnimator (changes requested):

LillyProgrammingIncorprated:
Very cool! Pthread is pretty important so utilizing it is essential in my opinion.

PR #7481, opened 2020-12-18
cast sendto to SendError inside send

lun-4:
Is this the right approach...?

Fix #7474.

PR #7522, opened 2020-12-22
Fixes wrong error set in std.net

MasterQ32:
Changes `std.os.send` error set to std.os.SendToError, as it uses `sendto` internally.


LemonBoy:
#7481 ?

MasterQ32:
Oh yeah, this seems a duplicate then.

PR #7523, opened 2020-12-22
std.c: add fmemopen

xackus:


daurnimator (approved):
Should `buf` be a `?[*]u8`?


Why not `open_memstream` and/or `fopencookie`?

xackus:
> 
> 
> Should `buf` be a `?[*]u8`?

I followed other declarations in `std.c`, like `malloc`.

> Why not `open_memstream` and/or `fopencookie`?

Because I didn't need them for what I was writing.

PR #7524, opened 2020-12-22
Added std.io.counting_reader

tgschultz:
Added because 1) it compliments io.counting_writer and 2) I had to implement it to avoid an extra allocation in a WASM module deserializer.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -362,6 +362,7 @@ set(ZIG_STAGE2_SOURCES
     "${CMAKE_SOURCE_DIR}/lib/std/io/buffered_writer.zig"
     "${CMAKE_SOURCE_DIR}/lib/std/io/change_detection_stream.zig"
     "${CMAKE_SOURCE_DIR}/lib/std/io/counting_writer.zig"
+    "${CMAKE_SOURCE_DIR}/lib/std/io/counting_reader.zig"
     "${CMAKE_SOURCE_DIR}/lib/std/io/find_byte_out_stream.zig"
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -362,6 +362,7 @@ set(ZIG_STAGE2_SOURCES
     "${CMAKE_SOURCE_DIR}/lib/std/io/buffered_writer.zig"
     "${CMAKE_SOURCE_DIR}/lib/std/io/change_detection_stream.zig"
     "${CMAKE_SOURCE_DIR}/lib/std/io/counting_writer.zig"
+    "${CMAKE_SOURCE_DIR}/lib/std/io/counting_reader.zig"
```
Vexu, line 365: I don't think you need to update this unless you actually use it in the self hosted compiler.

daurnimator: For now its a list of every zig stdlib file.
Otherwise someone needs to remember to update it when they use certain functions in stage2

lib/std/io/counting_reader.zig
```diff
@@ -0,0 +1,48 @@
...
+    var counting_stream = countingReader(fbs.reader());
+    const stream = counting_stream.reader();
+    
+    //read and discard all bytes
+    while(stream.readByte()) |_| {} else |err| {
+        testing.expect(err == error.EndOfStream);
+    }
+    
+    testing.expect(counting_stream.bytes_read == bytes.len);
+}
```
daurnimator, line 48: missing newline at EOF

Issue #7527, opened 2020-12-22
threadlocal generates a segfault on Apple Silicon


kubkon:
The following snippet:

```zig
const std = @import("std");
const assert = std.debug.assert;

threadlocal var x: i32 = 1234;

pub fn main() anyerror!void {
    const thread1 = try std.Thread.spawn({}, testTls);
    const thread2 = try std.Thread.spawn({}, testTls);
    testTls({});
    thread1.wait();
    thread2.wait();
}

fn testTls(context: void) void {
    assert(x == 1234);
    x += 1;
    assert(x == 1235);
}
```

compiles and cross-compiles fine on `aarch64-macos-gnu`; however, generates a `SEGFAULT` when run due to a deficiency in LLVM's LLD linker. The same snippet works fine when linked using Apple's system linker `ld` or on `x86_64` using LLD.

PR #7528, opened 2020-12-22
Pass full test suite on Apple Silicon

kubkon:
In order to pass the test suite on Apple Silicon, I've had to disable any test that makes use of `threadlocal` variables due to problems with LLD. See issue ziglang/zig#7527 for more details.

With this PR landing, it is now possible to successfully execute `zig build test` on Apple Silicon. One word of caution though. I've noticed that cross-compiling behavior tests to 64-bit RISC architectures sporadically fails with integer overflow runtime trap. If that happens, re-running the test suite generally fixes the problem for some mysterious reason. I haven't investigated it yet, but leaving it here for posterity.

The next thing would be to get our hands on some Apple Silicon-based CI to have the tests triggered on every push and pull request.

### DIFF
--- a/lib/std/os/test.zig
+++ b/lib/std/os/test.zig
@@ -298,6 +298,8 @@ test "cpu count" {
 
 test "thread local storage" {
     if (builtin.single_threaded) return error.SkipZigTest;
+    // TODO https://github.com/ziglang/zig/issues/7527
+    if (std.Target.current.isDarwin() and builtin.arch == .aarch64) return error.SkipZigTest;
     const thread1 = try Thread.spawn({}, testTls);
     const thread2 = try Thread.spawn({}, testTls);
     testTls({});
--- a/lib/std/os/test.zig
+++ b/lib/std/os/test.zig
@@ -299,7 +299,8 @@ test "cpu count" {
 test "thread local storage" {
     if (builtin.single_threaded) return error.SkipZigTest;
     // TODO https://github.com/ziglang/zig/issues/7527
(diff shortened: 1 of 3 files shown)


LemonBoy:
I'd run the whole suite with the `--single-threaded` flag instead, we use `threadlocal` variables in the panic handler too so this is a ticking time bomb.

Unrelated question:
Are you linking against a release build of LLVM or a debug one?
I just tried compiling Zig on a AArch64/linux box and found that a LLVM assertion is triggered when building in Release mode.

kubkon:
> I'd run the whole suite with the `--single-threaded` flag instead, we use `threadlocal` variables in the panic handler too so this is a ticking time bomb.

That might work for the CI, but what if you add something that's `aarch64` Darwin specific to libstd and want to test it? Are you suggesting that on Apple Silicon we have to force everyone to compile Zig in single threaded mode just so that `threadlocal` is never triggered in a multi-threaded context?

> Unrelated question:
> Are you linking against a release build of LLVM or a debug one?
> I just tried compiling Zig on a AArch64/linux box and found that a LLVM assertion is triggered when building in Release mode.

I'm linking against LLVM 11 compiled in `RelWithDebInfo` mode. However that might explain why I'm getting `integer overflow` traps sporadically when cross-compiling to `aarch64-linux` and `riscv-linux`.

kubkon:
On second thought, there is a doc example for `threadlocal` which will inevitably trip the test suite on AS, and I don't think it's prudent to make any tweaks to it especially since AS is tier 2 only. So yeah, I think I'll close this PR and if we do decide to go with some CI at some point, then build with `--single-threaded` enabled.

PR #4665, opened 2020-03-07
Use std.Buffer even less

daurnimator:
Followup to #4405

PR #4405, opened 2020-02-07
Stop using std.Buffer in places that don't make sense

daurnimator:
I imagine due to its name, `std.Buffer` is used in many places that it shouldn't be. (Related: #4385)
This PR moves many use-sites over to more-correct APIs

PR #7531, opened 2020-12-23
Move ArrayListSentineled to std lib orphanage

### LINKED ISSUE #6967: std.ArrayListSentineled integer overflow error.
The following code errors with an `integer overflow` error coming from the `std.ArrayListSentineled` function `len()`.
The mentioned function doesn't seem to consider the case of a "null" initialized array, and therefore tries to subtract 1 from the
length of the backing array (which is zero because it was null-initialized)

```zig
const std = @import("std");
const List = std.ArrayListSentineled(?u8, null);

pub fn main() anyerror!void {
    const alloc = std.heap.page_allocator;
    var my_list = List.initNull(alloc);
    defer my_list.deinit();
    try my_list.append(12);
    std.debug.print("len: {}\n", .{my_list.len()}); // Also fails.
}
```

Vexu:
Keeping the list sentinel terminated makes use of the type more difficult and expensive compared to a regular ArrayList. Combined with the fact that most usages of it that I could find either completely ignored the sentinel or could easily be replaced by the `toOwndedSliceSentinel` function I added and that a lot of its functionality has bitrotten makes it hard to justify keeping it in the standard library.

Closes #6968
Closes #6967


lib/std/child_process.zig
```diff
@@ -749,38 +748,38 @@ fn windowsCreateProcess(app_name: [*:0]u16, cmd_line: [*:0]u16, envp_ptr: ?[*]u1
 
 /// Caller must dealloc.
 fn windowsCreateCommandLine(allocator: *mem.Allocator, argv: []const []const u8) ![:0]u8 {
-    var buf = try ArrayListSentineled(u8, 0).initSize(allocator, 0);
+    var buf = try ArrayList(u8).init(allocator);
     defer buf.deinit();
-    const buf_stream = buf.outStream();
+    const buf_wi = buf.outStream();
```
kubkon, line 753: If we're doing cleanup, I guess this should use `writer` since `outStream` is getting deprecated?

```suggestion
    const buf_wi = buf.writer();
```

kubkon: On second thought, is `buf_wi` even needed here at all?

lib/std/net.zig
```diff
@@ -1177,10 +1179,10 @@ fn linuxLookupNameFromDnsSearch(
         if (byte == '.') dots += 1;
     }
 
-    const search = if (rc.search.isNull() or dots >= rc.ndots or mem.endsWith(u8, name, "."))
```
kubkon, line 1180: How come this change made it in this PR?

lib/std/child_process.zig
Vexu: I removed all uses of `buf_stream` as they were unceseccary, but then forgot to remove this. Will update.

kubkon (approved):
In general, I find the new transformer `ArrayList.toOwnedSliceSentinel` much cleaner than an entirely separate container type `ArrayListSentineled`. Additionally, now it should be easier to maintain the said container as well. So :+1: from me.

lib/std/net.zig
Vexu: `rc.search.isNull()` just does `rc.search.items.len == 0` in which case `rc.search.items` would be an empty string anyways so both branches would result in the same value.

kubkon: Gotcha, thanks!

andrewrk (approved):

daurnimator (approved):
Oh how far `std.Buffer` has fallen :+1: 

#4665, #4405

Issue #6044, opened 2020-08-13
'.' and '..' components in relative paths in std.os.windows.OpenFile


kubkon:
While working on #5701 and #5993 I've noticed that doing something like this on Windows:

```zig
const fd: fd_t = some_fd_thats_non_null;
const pathname = [_]u16{ '.' }; // or '..', or in fact, having '.' or '..' anywhere in the path
const new_fd = try std.os.windows.OpenFile(pathname, .{ .dir = fd });
defer std.os.windows.CloseHandle(new_fd);
```

will yield a runtime panic (actually, `NtCreateFile` will trip on `OBJECT_NAME_INVALID`). As far as I can work out, this is due to the presence of '.' and '..' in the pathname argument. My question is here, is there any simple way around this, or are we forced to re-implement `CreateFileW` where we either query the OS for the handle to cwd, or we use the `.dir` handle, get the canonical path to it using `GetFinalPathNameByHandle`, concatenate the resultant path with the specified argument path, and feed this path into `NtCreateFile` again (not sure if that would even work since I'd imagine we'd prepend `\??\` to the resultant path which could potentially not accept '.' and '..' as valid path components)?

I'm not yet convinced whether we'd want to enable this kind of behaviour for all calls using `std.os.windows.OpenFile`, however, it does seem useful for `std.os.realpath` and `std.fs.Dir.realpath`, when canonicalizing the pathnames.

squeek502:
Related: https://github.com/ziglang/zig/issues/4659

kubkon:
> Related: #4659

Nice, thanks @squeek502!

marler8997:
Here's an easy way to reproduce this:

The compiler fails with `error: Unexpected` when a source file has `..` on Windows:
```batch
> mkdir src
> echo "" > src\bug.zig
> zig build-obj src\..\src\foo.zig
error: Unexpected
```

PR #7537, opened 2020-12-24
Fix paths on Windows

marler8997:
fixes: https://github.com/ziglang/zig/issues/6044

This PR replaces the custom logic Zig has implemented to convert legacy DOS paths to NT paths with calls into `ntdll` to do the conversion for us.  These should be the same conversion functions that Win32 functions like `CreateFile` use to convert DOS paths to NT paths.

Now that `ntdll` is handling this conversion, missing features like support for `"."` and `".."` should be resolved.

### Follow Up: look for more NT calls

This PR has only replaced the implementation for the `openDir*` calls because those functions boil down to `NtCreateFile` which requires an NT formatted path.  We should see if there are any other filesystem functions that end up calling NT-specific functions and fix those as well (if there are any).  Look for places where `sliceToPrefixedFileW` is called and where those paths are passed to `NT` functions such as `NtCreateFile`.

### Follow Up: consider RtlDosPathNameToRelativeNtPathName_U_WithStatus

Another thing to follow up on is using `RtlDosPathNameToRelativeNtPathName_U_WithStatus` for relative paths.  The current mechanism uses `RtlDosPathNameToNtPathName_U_WithStatus` for the path and then removes the CWD prefix.  I don't know whether one implementation has benefits over the other.  There's also diverging implementations to consider, I noticed that Wine and ReactOS seemed to have differences in opinion on these functions, so we may want to keep our usage of them to a minimal subset.

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -1168,7 +1168,7 @@ pub const Dir = struct {
     /// Asserts that the path parameter has no null bytes.
     pub fn openDir(self: Dir, sub_path: []const u8, args: OpenDirOptions) OpenError!Dir {
         if (builtin.os.tag == .windows) {
-            const nt_path = try os.windows.NtPath.initA(sub_path);
+            const nt_path = try os.windows.NtPath.init((try os.windows.utf8ToWPathSpace(sub_path)).span());
             defer nt_path.deinit();
             return self.openDirWindows(std.fs.path.isAbsoluteWindows(sub_path), nt_path.str, args);
         } else if (builtin.os.tag == .wasi) {
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -1168,8 +1168,9 @@ pub const Dir = struct {
     /// Asserts that the path parameter has no null bytes.
     pub fn openDir(self: Dir, sub_path: []const u8, args: OpenDirOptions) OpenError!Dir {
         if (builtin.os.tag == .windows) {
-            const sub_path_w = try os.windows.sliceToPrefixedFileW(sub_path);
-            return self.openDirW(sub_path_w.span().ptr, args);
+            const nt_path = try os.windows.NtPath.initA(sub_path);
+            defer nt_path.deinit();
+            return self.openDirWindows(std.fs.path.isAbsoluteWindows(sub_path), nt_path.str, args);
         } else if (builtin.os.tag == .wasi) {
             return self.openDirWasi(sub_path, args);
         } else {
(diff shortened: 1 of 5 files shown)


lib/std/os/windows.zig
```diff
@@ -1569,6 +1569,54 @@ pub fn sliceToPrefixedFileW(s: []const u8) !PathSpace {
     return path_space;
 }
 
+/// Convert `dos_path` to an NT path that can be passed to Nt* functions like NtCreateFile.
+pub const NtPath = struct {
+    str: UNICODE_STRING,
+    pub fn initA(dos_path: []const u8) !NtPath {
```
xackus, line 1575: The "A" suffix is misleading when the function accepts utf8.

marler8997: It's the standard windows mechanism `A` vs `W`.  The implementation may utilize the `utf8` function to convert it to a u16 array, but the OS only accepts Ascii characters AFAIK.

marler8997: I've removed this method altogether and replaced it by providing a function that any code can use to convert a UTF8 encoded path to a Wide Character windows path on the stack `utf8ToWPathSpace`.

lib/std/os/windows.zig
```diff
@@ -1569,6 +1579,48 @@ pub fn sliceToPrefixedFileW(s: []const u8) !PathSpace {
     return path_space;
 }
 
+/// Convert `dos_path` to an NT path that can be passed to Nt* functions like NtCreateFile.
+pub const NtPath = struct {
+    str: UNICODE_STRING,
+    pub fn init(dos_path: [*:0]const u16) !NtPath {
+        var str : UNICODE_STRING = undefined;
+        const status = ntdll.RtlDosPathNameToNtPathName_U_WithStatus(dos_path, &str, null, null);
+        if (status != .SUCCESS) return unexpectedStatus(status);
```
andrewrk, line 1588: This introduces unnecessary heap allocation, causing `OutOfMemory` to be added to the `OpenError` set. We can do better. Check the wine implementation of `RtlDosPathNameToNtPathName_U_WithStatus` for inspiration. Opening a file should not do heap allocation for the path name.

andrewrk (changes requested):

lib/std/fs.zig
```diff
@@ -1235,15 +1235,13 @@ pub const Dir = struct {
         }
     }
 
-    /// Same as `openDir` except the path parameter is WTF-16 encoded, NT-prefixed.
+    /// Same as `openDir` except the path parameter is WTF-16 encoded.
     /// This function asserts the target OS is Windows.
     pub fn openDirW(self: Dir, sub_path_w: [*:0]const u16, args: OpenDirOptions) OpenError!Dir {
```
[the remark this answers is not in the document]
marler8997: If we delete it, then applications that have a `u16` string will now need to do 2 conversions, one to `u8` and then the second internal conversion.

marler8997:
The idea behind this PR was to remove the Windows logic from Zig's standard library and use the implementation that Windows provides instead for Nt path conversion.  However, the interface that Windows provides requires heap allocation. @andrewrk has indicated that in order to avoid this heap allocation, we should duplicate the code inside `ntdll` in the Zig standard library so we can provide an interface that does not require heap allocation.

With this new direction, I'll have to start over so I'm going to close this PR and create a new one for the new direction.

PR #7542, opened 2020-12-24
remove `git diff` feature from build.zig

g-w1:


FireFox317:
Nice one, thanks! I have run into this as well, and it was really annoying. However I do think it is weird that it returns StreamTooLong, shouldn't it just work even if the git output is really long?

g-w1:
See std/build.zig:1027 it makes a relatavly small max size (1024 * 400).

andrewrk:
Yes thanks for this. However I think it would be best to remove this code entirely from the build script. Compiler contributors will have to be aware that the cache compiler ID is not different based on working tree changes. In exchange for this we get a simpler build script, faster build, and ability to debug cache issues.

LemonBoy:
What about using something more compact like `git diff` [raw format](https://git-scm.com/docs/git-diff#_raw_output_format)?

andrewrk:
I think that would be a good bet if we wanted to keep this functionality. I say we move forward with removing it, try out development without it, and then we can reconsider adding it using the git diff raw format. Thanks for pointing that out; I was unaware of its existence.

PR #7546, opened 2020-12-25
initial support for haiku 

hoanga:
seasons greetings zig community!

the following PR adds initial support for getting zig to build stage1 on [Haiku](https://www.haiku-os.org/).  i have confirmed it compiles zig and zig0 and also still builds on other OSes (tested on mac).

the following is a short excerpt of building (requires llvm11)
```
~/src/git/zig/build> make                      
[  4%] Built target zigcpp                                      
[ 81%] Built target embedded_softfloat                             
[ 83%] Built target opt_c_util                              
[ 97%] Built target zigstage1
[ 99%] Built target zig0
[ 99%] Building self-hosted component /boot/home/src/git/zig/build/zig1.o
[ 99%] Linking CXX executable zig
[100%] Built target zig

~/src/git/zig/build> uname -a
Haiku shredder 1 hrev54798 Dec 19 2020 08:17:11 x86_64 x86_64 Haiku

~/src/git/zig/build> /boot/home/src/git/zig/build/zig env
{
 "zig_exe": "/boot/home/src/git/zig/build/zig",
 "lib_dir": "/boot/home/src/git/zig/lib",
 "std_dir": "/boot/home/src/git/zig/lib/std",
 "global_cache_dir": "/boot/home/config/cache/zig",
 "version": "0.7.1+9f3398411"
}
```

there are some pieces that could use a little more polish but wanted to submit a rough working version for review and suggestions on if the PR can be made more modular or cleaner to merge.

some things i know that could be a little better and will see about getting them fixed when i have some more time if possible (suggestions / help would be welcome)

* stub code areas could use more functional versions
* make install still won't work as-is (see below)
* could not compile using Debug mode (used Release)
* need to compile llvm for Haiku

```

~/src/git/zig/build> make install
[  4%] Built target zigcpp
[ 81%] Built target embedded_softfloat
[ 83%] Built target opt_c_util
[ 97%] Built target zigstage1
[ 99%] Built target zig0
[100%] Built target zig
Install the project...
-- Install configuration: "Release"
-- Installing: /boot/home/src/git/zig/build/bin/zig
-- Set runtime path of "/boot/home/src/git/zig/build/bin/zig" to ""
-- Installing: /boot/home/src/git/zig/build/lib
error: Unexpected
::
:: ERROR: 1
:: (execute_process)
::
:: argv: /boot/home/src/git/zig/build/zig build --override-lib-dir /boot/home/src/git/zig/lib -Dlib-files-only --prefix /boot/home/src/git/zig/build -
Dconfig_h=/boot/home/src/git/zig/build/config.h install
::
:: argv[0]: /boot/home/src/git/zig/build/zig
:: argv[1]: build
:: argv[2]: --override-lib-dir
:: argv[3]: /boot/home/src/git/zig/lib
:: argv[4]: -Dlib-files-only
:: argv[5]: --prefix
:: argv[6]: /boot/home/src/git/zig/build
:: argv[7]: -Dconfig_h=/boot/home/src/git/zig/build/config.h
:: argv[8]: install
::
CMake Error at /boot/home/src/git/zig/cmake/install.cmake:36 (message):
Call Stack (most recent call first):
  cmake_install.cmake:73 (include)


Makefile:114: recipe for target 'install' failed
make: *** [install] Error 1

```

### DIFF
--- a/lib/std/c/haiku.zig
+++ b/lib/std/c/haiku.zig
@@ -14,9 +14,20 @@ extern "c" fn _errnop() *c_int;
 
 pub const _errno = _errnop;
 
+pub extern "c" fn find_directory(which: c_int, volume: i32, createIt: bool, path_ptr: [*]u8, length: i32) u64;
+
+pub extern "c" fn find_thread(thread_name: ?*c_void) i32;
+
+pub extern "c" fn get_system_info(system_info: *system_info) usize;
+
+// TODO revisit if abi changes or better option becomes apparent
+pub extern "c" fn _get_next_image_info(team: c_int, cookie: *i32, image_info: *image_info) usize;
+
 pub extern "c" fn _kern_read_dir(fd: c_int, buf_ptr: [*]u8, nbytes: usize, maxcount: u32) usize;
 
-pub extern "c" fn _get_next_image_info(team: c_int, cookie: *i32, image_info: *image_info) usize;
+pub extern "c" fn _kern_read_stat(fd: c_int, path_ptr: [*]u8, traverse_link: bool, libc_stat: *libc_stat, stat_size: i32) usize;
+
+pub extern "c" fn _kern_get_current_team() i32;
 
 pub const sem_t = extern struct {
     _magic: u32,
@@ -27,6 +38,14 @@ pub const sem_t = extern struct {
     _padding: u32,
 };
 
+pub const pthread_attr_t = extern struct {
+    __detach_state: i32,
+    __sched_priority: i32,
+    __stack_size: i32,
+    __guard_size: i32,
+    __stack_address: ?*c_void,
+};
+
 pub const pthread_mutex_t = extern struct {
     flags: u32 = 0,
     lock: i32 = 0,
@@ -34,7 +53,6 @@ pub const pthread_mutex_t = extern struct {
     owner: i32 = -1,
     owner_count: i32 = 0,
 };
-
 pub const pthread_cond_t = extern struct {
     flags: u32 = 0,
     unused: i32 = -42,
--- a/lib/std/c/haiku.zig
+++ b/lib/std/c/haiku.zig
@@ -11,15 +11,13 @@ const builtin = std.builtin;
 usingnamespace std.c;
 
 extern "c" fn _errnop() *c_int;
+
 pub const _errno = _errnop;
 
-// not supported in haiku
-pub extern "c" fn getdents(fd: c_int, buf_ptr: [*]u8, nbytes: usize) usize;
+pub extern "c" fn _kern_read_dir(fd: c_int, buf_ptr: [*]u8, nbytes: usize, maxcount: u32) usize;
 
-pub const dl_iterate_phdr_callback = fn (info: *dl_phdr_info, size: usize, data: ?*c_void) callconv(.C) c_int;
-//pub extern "c" fn dl_iterate_phdr(callback: dl_iterate_phdr_callback, data: ?*c_void) c_int;
+pub extern "c" fn _get_next_image_info(team: c_int, cookie: *i32, image_info: *image_info) usize;
 
-//
 pub const sem_t = extern struct {
     _magic: u32,
     _kern: extern struct {
--- a/lib/std/c/haiku.zig
+++ b/lib/std/c/haiku.zig
@@ -3,6 +3,32 @@
 // This file is part of [zig](https://ziglang.org/), which is MIT licensed.
 // The MIT license requires this copyright notice to be included in all copies
 // and substantial portions of the software.
+
+//
+const std = @import("../std.zig");
+const builtin = std.builtin;
+
+usingnamespace std.c;
+
+extern "c" fn _errnop() *c_int;
+pub const _errno = _errnop;
+
+// not supported in haiku
+pub extern "c" fn getdents(fd: c_int, buf_ptr: [*]u8, nbytes: usize) usize;
+
+pub const dl_iterate_phdr_callback = fn (info: *dl_phdr_info, size: usize, data: ?*c_void) callconv(.C) c_int;
+//pub extern "c" fn dl_iterate_phdr(callback: dl_iterate_phdr_callback, data: ?*c_void) c_int;
+
+//
+pub const sem_t = extern struct {
+    _magic: u32,
+    _kern: extern struct {
+        _count: u32,
+        _flags: u32,
+    },
+    _padding: u32,
+};
+
 pub const pthread_mutex_t = extern struct {
     flags: u32 = 0,
     lock: i32 = 0,
@@ -10,6 +36,7 @@ pub const pthread_mutex_t = extern struct {
     owner: i32 = -1,
     owner_count: i32 = 0,
 };
+
 pub const pthread_cond_t = extern struct {
     flags: u32 = 0,
     unused: i32 = -42,
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -1350,7 +1350,6 @@ pub const DebugInfo = struct {
 
         return &di;
     }
-
 };
 
 const SymbolInfo = struct {
@@ -1717,17 +1716,6 @@ pub const ModuleDebugInfo = switch (builtin.os.tag) {
             unreachable;
         }
     },
-    .haiku => struct {
-        // Haiku should implement dl_iterat_phdr (https://dev.haiku-os.org/ticket/15743)
-        base_address: usize,
-        dwarf: DW.DwarfInfo,
-        mapped_memory: []const u8,
-
-        pub fn getSymbolAtAddress(self: *@This(), address: usize) !SymbolInfo {
-            // TODO: implement me
-            return SymbolInfo{};
-        }
-    },
     else => DW.DwarfInfo,
 };
 
@@ -1746,7 +1734,7 @@ fn getDebugInfoAllocator() *mem.Allocator {
 pub const have_segfault_handling_support = switch (builtin.os.tag) {
     .linux, .netbsd => true,
     .windows => true,
-    .freebsd, .openbsd, .haiku => @hasDecl(os, "ucontext_t"),
+    .freebsd, .openbsd => @hasDecl(os, "ucontext_t"),
     else => false,
 };
 pub const enable_segfault_handler: bool = if (@hasDecl(root, "enable_segfault_handler"))
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -1125,12 +1125,15 @@ pub const DebugInfo = struct {
     }
 
     pub fn getModuleForAddress(self: *DebugInfo, address: usize) !*ModuleDebugInfo {
-        if (comptime std.Target.current.isDarwin())
-            return self.lookupModuleDyld(address)
-        else if (builtin.os.tag == .windows)
-            return self.lookupModuleWin32(address)
-        else
+        if (comptime std.Target.current.isDarwin()) {
+            return self.lookupModuleDyld(address);
+        } else if (builtin.os.tag == .windows) {
+            return self.lookupModuleWin32(address);
+        } else if (builtin.os.tag == .haiku) {
+            return self.lookupModuleHaiku(address);
+        } else {
             return self.lookupModuleDl(address);
+        }
     }
 
     fn lookupModuleDyld(self: *DebugInfo, address: usize) !*ModuleDebugInfo {
@@ -1336,6 +1339,18 @@ pub const DebugInfo = struct {
 
         return obj_di;
     }
+
+    fn lookupModuleHaiku(self: *DebugInfo, address: usize) !*ModuleDebugInfo {
+        // TODO: implement me
+        var di = ModuleDebugInfo{
+            .base_address = undefined,
+            .dwarf = undefined,
+            .mapped_memory = undefined,
+        };
+
+        return &di;
+    }
+
 };
 
 const SymbolInfo = struct {
@@ -1702,6 +1717,17 @@ pub const ModuleDebugInfo = switch (builtin.os.tag) {
             unreachable;
         }
     },
+    .haiku => struct {
+        // Haiku should implement dl_iterat_phdr (https://dev.haiku-os.org/ticket/15743)
+        base_address: usize,
+        dwarf: DW.DwarfInfo,
+        mapped_memory: []const u8,
+
+        pub fn getSymbolAtAddress(self: *@This(), address: usize) !SymbolInfo {
+            // TODO: implement me
+            return SymbolInfo{};
+        }
+    },
     else => DW.DwarfInfo,
 };
 
@@ -1720,7 +1746,7 @@ fn getDebugInfoAllocator() *mem.Allocator {
 pub const have_segfault_handling_support = switch (builtin.os.tag) {
     .linux, .netbsd => true,
     .windows => true,
-    .freebsd, .openbsd => @hasDecl(os, "ucontext_t"),
+    .freebsd, .openbsd, .haiku => @hasDecl(os, "ucontext_t"),
     else => false,
 };
 pub const enable_segfault_handler: bool = if (@hasDecl(root, "enable_segfault_handler"))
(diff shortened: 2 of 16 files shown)


hoanga:
archive of patches that can be applied to llvm to get it to build for haiku (can apply llvm in zig-bootstrap)

[llvm-patches.zip](https://github.com/ziglang/zig/files/5742104/llvm-patches.zip)

lib/std/c/haiku.zig
```diff
@@ -3,17 +3,49 @@
...
+//
+const std = @import("../std.zig");
+const builtin = std.builtin;
+
+usingnamespace std.c;
+
+extern "c" fn _errnop() *c_int;
+pub const _errno = _errnop;
+
+// not supported in haiku
```
LemonBoy, line 16: Then don't declare it.

lib/std/c/haiku.zig
```diff
@@ -3,17 +3,49 @@
...
+pub extern "c" fn getdents(fd: c_int, buf_ptr: [*]u8, nbytes: usize) usize;
+
+pub const dl_iterate_phdr_callback = fn (info: *dl_phdr_info, size: usize, data: ?*c_void) callconv(.C) c_int;
+//pub extern "c" fn dl_iterate_phdr(callback: dl_iterate_phdr_callback, data: ?*c_void) c_int;
```
LemonBoy, line 20: Remove this.

lib/std/c/haiku.zig
```diff
@@ -3,17 +3,49 @@
...
+
+usingnamespace std.c;
+
+extern "c" fn _errnop() *c_int;
+pub const _errno = _errnop;
+
+// not supported in haiku
+pub extern "c" fn getdents(fd: c_int, buf_ptr: [*]u8, nbytes: usize) usize;
+
+pub const dl_iterate_phdr_callback = fn (info: *dl_phdr_info, size: usize, data: ?*c_void) callconv(.C) c_int;
```
LemonBoy, line 19: This is useless without `dl_iterate_phdr`.

lib/std/debug.zig
```diff
@@ -1311,6 +1314,18 @@ pub const DebugInfo = struct {
 
         return obj_di;
     }
+
+    fn lookupModuleHaiku(self: *DebugInfo, address: usize) !*ModuleDebugInfo {
+        // TODO: implement me
+        var di = ModuleDebugInfo{
```
LemonBoy, line 1320: This stub is useless and will crash right away if used.
You should be able to implement this in terms of `get_next_image_info` instead of waiting for someone to implement `dl_iterate_phdr`.

Does Haiku use ELF images and Dwarf debug info?

lib/std/debug.zig
```diff
@@ -1677,6 +1692,17 @@ pub const ModuleDebugInfo = switch (builtin.os.tag) {
             unreachable;
         }
     },
+    .haiku => struct {
+        // Haiku should implement dl_iterat_phdr (https://dev.haiku-os.org/ticket/15743)
```
LemonBoy, line 1696: Not really, see above.

lib/std/debug.zig
```diff
@@ -1695,7 +1721,7 @@ fn getDebugInfoAllocator() *mem.Allocator {
 pub const have_segfault_handling_support = switch (builtin.os.tag) {
     .linux, .netbsd => true,
     .windows => true,
-    .freebsd, .openbsd => @hasDecl(os, "ucontext_t"),
+    .freebsd, .openbsd, .haiku => @hasDecl(os, "ucontext_t"),
```
LemonBoy, line 1724: Does Haiku support siginfo and has a definition for `ucontext_t`?

lib/std/fs.zig
```diff
@@ -426,6 +427,16 @@ pub const Dir = struct {
                     };
                 }
             }
+
+            fn nextHaiku(self: *Self) !?Entry {
+                //const name = @ptrCast([*]u8, "placeholder-please-implement-me");
+                //const name = "placeholder-please-implement-me";
+                return Entry{
+                    .name = base64_alphabet,
```
LemonBoy, line 435: Ehh.

lib/std/os.zig
```diff
@@ -3845,6 +3847,21 @@ pub fn pipe2(flags: u32) PipeError![2]fd_t {
         };
         return fds;
     }
+    if (comptime std.Target.current.isHaiku()) {
```
LemonBoy, line 3850: Isn't this identical to the Darwin code path?

lib/std/process.zig
```diff
@@ -775,6 +775,19 @@ pub fn getSelfExeSharedLibPaths(allocator: *Allocator) error{OutOfMemory}![][:0]
             }
             return paths.toOwnedSlice();
         },
+        // Haiku should implement dl_iterat_phdr (https://dev.haiku-os.org/ticket/15743)
```
LemonBoy, line 778: Ditto.

andrewrk:
Season's greetings from the Zig project to the Haiku project! Thanks for the patches!

marler8997:
Seeing this PR begs the question, what criteria does Zig use to support platforms in the standard library?

lib/std/debug.zig
hoanga: i agree that the stub doesn't do much as is. Haiku does use the ELF format for binaries and does support dwarf debugging information (at least that is what gcc with debugging reported back)

thanks for the suggestion on not waiting around for `dl_iterate_phdr` support to roll around. implementing  with `get_next_image_info` is on my list of things i'd like to get to

lib/std/debug.zig
hoanga: for the time being, i have backed most of the changes except for more minimal stub logic for debug and will try to tackle this as a separate item (per above)

lib/std/debug.zig
hoanga: Haiku has definitions for `ucontext_t` and from what i can tell has support for siginfo although i imagine a more seasoned haiku developer will have a more definitive answer

(see `/system/develop/headers/posix/signal.h` on a Haiku system)

lib/std/fs.zig
hoanga: i've updated the logic (hopefully with something slightly less worse)

lib/std/os.zig
hoanga: yes, updated conditional and re-used Darwin code path instead

hoanga:
thank you for the detailed review of the PR.  i have pushed updates that hopefully addresses a good portion of the suggestions posted. i have tried to focus on ensuring it can at least build on Haiku and leave the fun parts for later adventures.

after some more tinkering with the build as is on Haiku things i was able to confirm:

* it still builds and hopefully doesn't break other platforms
* make install still doesn't work
* was able to play around with some of the commands ` zig targets`, `zig run` and `zig fmt ` and confirm they are somewhat functional 
* `zig build` still needs some more work
* `zig zen` runs beautifully :)

lib/std/fs/get_app_data_dir.zig
```diff
@@ -56,6 +56,13 @@ pub fn getAppDataDir(allocator: *mem.Allocator, appname: []const u8) GetAppDataD
             };
             return fs.path.join(allocator, &[_][]const u8{ home_dir, ".local", "share", appname });
         },
+        .haiku => {
+            const home_dir = os.getenv("HOME") orelse {
```
waddlesplash, line 60: Haiku applications should store their data in `~/config/settings`, which you can determine by using the [`find_directory` API](https://www.haiku-os.org/docs/api/FindDirectory_8h.html).

lib/std/os/bits/haiku.zig
```diff
@@ -0,0 +1,1410 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2015-2020 Zig Contributors
+// This file is part of [zig](https://ziglang.org/), which is MIT licensed.
+// The MIT license requires this copyright notice to be included in all copies
+// and substantial portions of the software.
+const std = @import("../../std.zig");
+const maxInt = std.math.maxInt;
+
+// See https://svnweb.freebsd.org/base/head/sys/sys/_types.h?view=co
```
waddlesplash, line 9: This looks like a copy/paste error. (Haiku's equivalent types are defined in `sys/types.h`.)

lib/std/process.zig
```diff
@@ -775,6 +775,24 @@ pub fn getSelfExeSharedLibPaths(allocator: *Allocator) error{OutOfMemory}![][:0]
             }
             return paths.toOwnedSlice();
         },
+        // revisit if Haiku implements dl_iterat_phdr (https://dev.haiku-os.org/ticket/15743)
```
waddlesplash, line 778: You could also submit a patch to Haiku implementing it ;)

Otherwise, yes, `get_next_image_info` is the way to go.

lib/std/target.zig
```diff
@@ -1501,6 +1505,10 @@ pub const Target = struct {
             .other,
             => return result,
 
+            // Operating systems in this list have been verified as not having a standard
+            // dynamic linker path.
```
waddlesplash, line 1509: Looks like a copy/paste error?

lib/std/target.zig
```diff
@@ -1501,6 +1505,10 @@ pub const Target = struct {
...
+            .haiku => return copy(&result, "/system/runtime_loader"),
```
waddlesplash, line 1510: FWIW, while this is correct at the present time, in the future when/if Haiku introduces proper multiarch (i.e. 32-bit on 64-bit) support, there will be multiple dynamic linkers for each arch, and invoking the wrong one will probably cause problems.

lib/std/thread.zig
```diff
@@ -484,6 +484,9 @@ pub const Thread = struct {
             };
             return @intCast(usize, count);
         }
+        if (std.Target.current.os.tag == .haiku) {
```
waddlesplash, line 487: You can get this either via `get_system_info` (a Haiku-specific API), or via `sysconf(_SC_NPROCESSORS_CONF)` (which probably works on other OSes, too.)

lib/std/c/haiku.zig
```diff
@@ -3,17 +3,47 @@
...
+const std = @import("../std.zig");
+const builtin = std.builtin;
+
+usingnamespace std.c;
+
+extern "c" fn _errnop() *c_int;
+
+pub const _errno = _errnop;
+
+pub extern "c" fn _kern_read_dir(fd: c_int, buf_ptr: [*]u8, nbytes: usize, maxcount: u32) usize;
```
waddlesplash, line 17: `_kern...` functions are syscall hooks, and we make no guarantees whatsoever about their ABIs, which are subject to change without notice, etc. (Calling syscalls via hooks is at least better than by number, though, as the numbers are generated at build time and so change whenever syscalls are touched in pretty much any way.)

waddlesplash (commented):
Haiku developer here. Always exciting to see new languages add support for Haiku ... and especially when the code was not written by one of the usual HaikuPorts crew, to boot!

lib/std/fs.zig
```diff
@@ -427,6 +427,61 @@ pub const Dir = struct {
...
+                    const haiku_entry = @ptrCast(*align(1) os.dirent, &self.buf[self.index]);
+                    const next_index = self.index + haiku_entry.reclen();
+                    self.index = next_index;
+                    const name = mem.spanZ(@ptrCast([*:0]u8, &haiku_entry.d_name));
+
+                    if (mem.eql(u8, name, ".") or mem.eql(u8, name, "..") or (haiku_entry.d_ino == 0)) {
+                        continue :start_over;
+                    }
+
+                    // TODO: determine entry kind
```
waddlesplash, line 475: You can do this by masking the mode using `S_IFMT` and then checking the type (S_IFDIR etc.)

waddlesplash:
> make install still doesn't work

If this is because of "read-only file system" errors, you will need to adjust your install paths to put things in `/system/non-packaged` instead, as the `/system` hierarchy itself is a union-mount of the installed/activated packages. Most software on Haiku is distributed as HPKGs built via HaikuPorter; you can of course make them "manually" via the `package` tool, as well.

lib/std/fs/get_app_data_dir.zig
daurnimator: see 3 lines below where they are creating the path `~/config/settings`

But yes, it looks like `find_directory` would be the more official API

lib/std/target.zig
daurnimator: So should this follow the pattern above (in e.g. linux) of `.haiku =>  switch (self.cpu.arch) {`?

daurnimator:
@waddlesplash what is the status of "libbe.so" on haiku (which is where I see `find_directory` is documented as belonging to)? is it a base part of the system as much as libc is? Should it be considered an extension of libc? Or is libc on top of it?

waddlesplash:
`libbe.so` is part of the official Haiku API, yes. It contains most of the "Kits" (Application Kit, Interface Kit, etc.) used in developing GUI applications on Haiku. `libroot.so` is what contains the libc, and `libbe.so` is implemented on top of that, not the other way around, of course. But yes, `libbe.so` is definitely part of the base system and will always be around.

There has been discussion of breaking up `libbe.so` into separate libraries (having most of the GUI-related kits be in `libbe.so` is a holdover from BeOS), but that is a long way off to say the least.

waddlesplash:
Actually, the documentation page is just wrong, `find_directory` is actually in `libroot.so`. So that is all a moot point.

lib/std/thread.zig
```diff
@@ -484,6 +484,9 @@ pub const Thread = struct {
...
+            return 1;        // TODO:  implement me
```
andrewrk, line 488: Please remove all bugs such as this from this pull request. You can replace this with 

```suggestion
            @panic("TODO implement std.Thread.cpuCount for Haiku");
```

Or better yet:

```suggestion
            @compileError("TODO implement std.Thread.cpuCount for Haiku");
```

No wrong implementations please!

andrewrk (changes requested):

hoanga:
Lots of feedback! Appreciate all of the helpful comments. I will be going over the suggestions and work on getting the PR in a more ready state. Hopefully will have something ready soon!

lib/std/fs/get_app_data_dir.zig
hoanga: i have added in a very bare-bones implementation using the find_directory() as suggested.

lib/std/process.zig
hoanga: thanks for the invitation :) and might consider it in the future if someone else doesn't beat me to it. for the nearer term, i prefer to keep my focus on having something that can get this in a better compiling state on Haiku

lib/std/target.zig
hoanga: yes, updated with hopefully a better note.

lib/std/target.zig
hoanga: for the time being i have left a note regarding multiarch support.

lib/std/thread.zig
hoanga: added

lib/std/c/haiku.zig
hoanga: thanks! i will keep that in mind, i think some of these exposed syscall hooks might not be as necessary when _dl_iterat_phdr_ support is available

lib/std/fs.zig
hoanga: that helps, thanks :)

lib/std/thread.zig
hoanga: makes sense, and have placed in the suggestions in (hopefully) the appropriate places. might have implemented some parts in other areas which hopefully also addresses the problematic code spots

hoanga:
i have pushed more updates that should handle feedback up to this point and still am trying to stick with the goal of compiling under Haiku.

i have separated out the changes for debug and plan to defer that for later efforts

update on what i was able to confirm:

* still building and hopefully isn't breaking on other platforms
* zig zen still is okay but previous noted behaviors still appear to be there
* make install still isn't available but the zig binary does build 


thanks again for the feedback!

andrewrk:
> Seeing this PR begs the question, what criteria does Zig use to support platforms in the standard library?

Part of it is whether an active person is around to do the labor, and the other part of it is whether supporting given platform results in any compromises for more mainstream use cases.

In the future, I'd like to investigate OS support via packages; for now we do it by upstreaming into the std lib.

In this specific case, I'm happy to add experimental Haiku support and see where it goes.

PR #7548, opened 2020-12-25
openbsd: add commonly used libc wrappers for pledge(2) and unveil(2)

semarie:
the following commit adds [pledge(2)](https://man.openbsd.org/pledge.2) and [unveil(2)](https://man.openbsd.org/unveil.2) syscalls definition in `std.c` for OpenBSD.


semarie:
regarding optional usage:

- for `pledge(2)` : `promises` and `execpromises` are optional. you could invoke `pledge("stdio", NULL)` to only set `promises` of the current process (and no exec-ed processes)
- for `unveil(2)` : calling `unveil(NULL, NULL)` is valid as it forbid futher call to `unveil(2)` and lock the policy defined by previous calls

PR #7553, opened 2020-12-26
Fix the damn deadlock

andrewrk:
This branch adds debug code for detecting any deadlock in the Cache system. The fix is not in this branch yet but hopefully this new debug code will lead us to it.

I confirmed that when removing the code protecting detection of the same C source files:

```diff
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1865,25 +1865,6 @@ fn updateCObject(comp: *Compilation, c_object: *CObject, c_comp_progress_node: *
         }
     }
 
-    {
-        const is_collision = blk: {
-            const bin_digest = man.hash.peekBin();
-
-            const lock = comp.mutex.acquire();
-            defer lock.release();
-
-            const gop = try comp.c_object_cache_digest_set.getOrPut(comp.gpa, bin_digest);
-            break :blk gop.found_existing;
-        };
-        if (is_collision) {
-            return comp.failCObj(
-                c_object,
-                "the same source file was already added to the same compilation with the same flags",
-                .{},
-            );
-        }
-    }
-
     var arena_allocator = std.heap.ArenaAllocator.init(comp.gpa);
     defer arena_allocator.deinit();
     const arena = &arena_allocator.allocator;

```

The detection code is indeed triggered:

```
[nix-shell:~/Downloads/zig/build]$ ./zig cc simple_race.c simple_race.c 
Cache deadlock detected in Cache.hit. Manifest has 1 files:
  file: /home/andy/Downloads/zig/build/simple_race.c
Cache deadlock detected
/home/andy/Downloads/zig/src/Cache.zig:277:17: 0xb54a1a in Cache.Manifest.hit (zig1)
                @panic("Cache deadlock detected");
                ^
/home/andy/Downloads/zig/src/Compilation.zig:1890:63: 0xd95574 in Compilation.updateCObject (zig1)
    const digest = if (!comp.disable_c_depfile and try man.hit()) man.final() else blk: {
                                                              ^
/home/andy/Downloads/zig/src/Compilation.zig:1805:23: 0xd94c48 in Compilation.workerUpdateCObject (zig1)
    comp.updateCObject(c_object, progress_node) catch |err| switch (err) {
                      ^
/home/andy/Downloads/zig/src/ThreadPool.zig:117:28: 0xd97cc8 in ThreadPool.Closure.runFn (zig1)
            const result = @call(.{}, func, closure.arguments);
                           ^
/home/andy/Downloads/zig/src/ThreadPool.zig:35:38: 0xbd2452 in ThreadPool.Worker.run (zig1)
                (run_node.data.runFn)(&run_node.data);
                                     ^
/home/andy/Downloads/zig/lib/std/thread.zig:268:32: 0xbd2847 in std.thread.MainFuncs.posixThreadMain (zig1)
                        startFn(arg);
                               ^
???:?:?: 0x7f66c692deac in ??? (???)
Aborted (core dumped)
```

### DIFF
--- a/lib/std/child_process.zig
+++ b/lib/std/child_process.zig
@@ -19,6 +19,7 @@ const builtin = @import("builtin");
 const Os = builtin.Os;
 const TailQueue = std.TailQueue;
 const maxInt = std.math.maxInt;
+const assert = std.debug.assert;
 
 pub const ChildProcess = struct {
     pid: if (builtin.os.tag == .windows) void else i32,
@@ -376,19 +377,44 @@ pub const ChildProcess = struct {
             if (any_ignore) os.close(dev_null_fd);
         }
 
-        var env_map_owned: BufMap = undefined;
-        var we_own_env_map: bool = undefined;
-        const env_map = if (self.env_map) |env_map| x: {
-            we_own_env_map = false;
-            break :x env_map;
-        } else x: {
-            we_own_env_map = true;
-            env_map_owned = try process.getEnvMap(self.allocator);
-            break :x &env_map_owned;
-        };
-        defer {
-            if (we_own_env_map) env_map_owned.deinit();
+        var arena_allocator = std.heap.ArenaAllocator.init(self.allocator);
+        defer arena_allocator.deinit();
+        const arena = &arena_allocator.allocator;
+
+        // The POSIX standard does not allow malloc() between fork() and execve(),
+        // and `self.allocator` may be a libc allocator.
+        // I have personally observed the child process deadlocking when it tries
+        // to call malloc() due to a heap allocation between fork() and execve(),
+        // in musl v1.1.24.
+        // Additionally, we want to reduce the number of possible ways things
+        // can fail between fork() and execve().
+        // Therefore, we do all the allocation for the execve() before the fork().
+        // This means we must do the null-termination of argv and env vars here.
+        const argv_buf = try arena.alloc(?[*:0]u8, self.argv.len + 1);
+        for (self.argv) |arg, i| {
+            const arg_buf = try arena.alloc(u8, arg.len + 1);
+            @memcpy(arg_buf.ptr, arg.ptr, arg.len);
+            arg_buf[arg.len] = 0;
+            argv_buf[i] = arg_buf[0..arg.len :0].ptr;
         }
+        argv_buf[self.argv.len] = null;
+        const argv_ptr = argv_buf[0..self.argv.len :null].ptr;
+
+        const envp = m: {
+            if (self.env_map) |env_map| {
+                const envp_buf = try createNullDelimitedEnvMap(arena, env_map);
+                break :m envp_buf.ptr;
+            } else if (std.builtin.link_libc) {
+                break :m std.c.environ;
+            } else if (std.builtin.output_mode == .Exe) {
+                // Then we have Zig start code and this works.
+                // TODO type-safety for null-termination of `os.environ`.
+                break :m @ptrCast([*:null]?[*:0]u8, os.environ.ptr);
+            } else {
+                // TODO come up with a solution for this.
+                @compileError("missing std lib enhancement: ChildProcess implementation has no way to collect the environment variables to forward to the child process");
+            }
+        };
 
         // This pipe is used to communicate errors between the time of fork
         // and execve from the child process to the parent process.
@@ -438,7 +464,10 @@ pub const ChildProcess = struct {
                 os.setreuid(uid, uid) catch |err| forkChildErrReport(err_pipe[1], err);
             }
 
-            const err = os.execvpe_expandArg0(self.allocator, self.expand_arg0, self.argv, env_map);
+            const err = switch (self.expand_arg0) {
+                .expand => os.execvpeZ_expandArg0(.expand, argv_buf.ptr[0].?, argv_ptr, envp),
+                .no_expand => os.execvpeZ_expandArg0(.no_expand, argv_buf.ptr[0].?, argv_ptr, envp),
+            };
             forkChildErrReport(err_pipe[1], err);
         }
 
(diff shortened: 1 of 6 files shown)


andrewrk:
Managed to repro this. Here is a backtrace of all threads:

https://gist.github.com/andrewrk/ee0fc81ebd4d05912a633f0108ca68cb

2 of the threads are waiting on `std.child_process.ChildProcess.spawnAndWait`.

lib/std/child_process.zig
```diff
@@ -376,19 +377,44 @@ pub const ChildProcess = struct {
...
+        // I have personally observed the child process deadlocking when it tries
+        // to call malloc() due to a heap allocation between fork() and execve(),
+        // in musl v1.1.24.
+        // Additionally, we want to reduce the number of possible ways things
+        // can fail between fork() and execve().
+        // Therefore, we do all the allocation for the execve() before the fork().
+        // This means we must do the null-termination of argv and env vars here.
+        const argv_buf = try arena.alloc(?[*:0]u8, self.argv.len + 1);
+        for (self.argv) |arg, i| {
+            const arg_buf = try arena.alloc(u8, arg.len + 1);
```
ifreund, line 395: could use `allocSentinel` here
```suggestion
            const arg_buf = try arena.allocSentinel(u8, arg.len, 0);
```

lib/std/child_process.zig
```diff
@@ -376,19 +377,44 @@ pub const ChildProcess = struct {
...
+        // The POSIX standard does not allow malloc() between fork() and execve(),
+        // and `self.allocator` may be a libc allocator.
+        // I have personally observed the child process deadlocking when it tries
+        // to call malloc() due to a heap allocation between fork() and execve(),
+        // in musl v1.1.24.
+        // Additionally, we want to reduce the number of possible ways things
+        // can fail between fork() and execve().
+        // Therefore, we do all the allocation for the execve() before the fork().
+        // This means we must do the null-termination of argv and env vars here.
+        const argv_buf = try arena.alloc(?[*:0]u8, self.argv.len + 1);
```
ifreund, line 393:
```suggestion
        const argv_buf = try arena.allocSentinel(?[*:0]u8, self.argv.len, null);
```

lib/std/child_process.zig
```diff
@@ -881,3 +910,24 @@ pub fn createWindowsEnvBlock(allocator: *mem.Allocator, env_map: *const BufMap)
     i += 1;
     return allocator.shrink(result, i);
 }
+
+pub fn createNullDelimitedEnvMap(arena: *mem.Allocator, env_map: *const std.BufMap) ![:null]?[*:0]u8 {
+    const envp_count = env_map.count();
+    const envp_buf = try arena.alloc(?[*:0]u8, envp_count + 1);
```
ifreund, line 916:
```suggestion
    const envp_buf = try arena.allocSentinel(?[*:0]u8, envp_count, null);
```

src/Cache.zig
```diff
@@ -12,6 +12,15 @@ const mem = std.mem;
 const fmt = std.fmt;
 const Allocator = std.mem.Allocator;
 
+/// Process-scoped map keeping track of all locked Cache hashes, to detect deadlocks.
+/// The plan is to enable this for debug builds only, but for now we enable
+/// it always to catch a deadlock.
```
ifreund, line 17: This is now enabled only for debug builds as of 2b97b75bb073a691400297a56daa6f42480020c7

ifreund (approved):
This looks good to me. There's a good bit of cleanup that can be achieved using `std.mem.Allocator.alloSentinel()` but the code was already working before it got moved around so whether or not to do that is up to you.

ifreund:
Did the `allocSentinel()` cleanup and added a test here: #7563

lib/std/child_process.zig
```diff
@@ -376,19 +377,44 @@ pub const ChildProcess = struct {
...
+                break :m envp_buf.ptr;
+            } else if (std.builtin.link_libc) {
+                break :m std.c.environ;
+            } else if (std.builtin.output_mode == .Exe) {
+                // Then we have Zig start code and this works.
+                // TODO type-safety for null-termination of `os.environ`.
+                break :m @ptrCast([*:null]?[*:0]u8, os.environ.ptr);
+            } else {
+                // TODO come up with a solution for this.
+                @compileError("missing std lib enhancement: ChildProcess implementation has no way to collect the environment variables to forward to the child process");
```
jsirois, line 415: This has the upshot of not allowing Linux libraries not linking libc to use `std.process.Child` even when setting an `env_map` on the Child (this has since moved, but the logic is the same). At first I thought this was a compiler bug since not all arms of the condition are comptime known:

```zig
const std = @import("std")

fn IterThingOnce(Thing: anytype) type {
    return struct {
        yielded: bool = false,
        optional_thing: ?Thing,

        pub fn next(self: *@This()) ?Thing {
            if (self.optional_thing) |thing| {
                if (self.yielded) {
                    return null;
                } else {
                    defer self.yielded = true;
                    return thing;
                }
            } else if (false) {
                @compileError("This can't happen since `false` is a comptime known literal.");
            } else {
                // Compiles and test passes by exchanging commented out line below.
                @compileError("This seems like it should not happen.");
                // return null;
            }
        }
    };
}

test "Compiler Bug?" {
    var example: IterThingOnce([]const u8) = .{.optional_thing = "Slatibartfast"};
    try std.testing.expectEqual("Slatibartfast", example.next().?);
    try std.testing.expect(example.next() == null);
}
```

```console
... error: This seems like it should not happen.
                @compileError("This seems like it should not happen.");
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

But, on reflection, that's the point of the else arm triggering a compile error - that branch can be hit at runtime.

I guess what I'm wondering is if that was the intention; i.e.: Not allow the Linux library with no libc case to get a runtime error instead iff they do not set up an `env_map`. I'd prefer the runtime error, since I know what I'm doing and set the env_map up (I actually go further and setup std.os.environ with my own ~start code in the library entry point).

alexrp: Please file an issue about this; it's very likely to get lost on a 4 years old PR. It seems like a case we should fix.

PR #7555, opened 2020-12-26
stage2 error notes

Vexu:


andrewrk (changes requested):
Sorry for the delayed review.

I'd like to suggest a different data layout for error messages & notes (diff against master branch):

```diff
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -251,6 +251,8 @@ pub const AllErrors = struct {
             column: usize,
             byte_offset: usize,
             msg: []const u8,
+            /// Children are always notes.
+            children: []Message,
         },
         plain: struct {
             msg: []const u8,
@@ -2364,6 +2366,8 @@ fn failCObjWithOwnedErrorMsg(comp: *Compilation, c_object: *CObject, err_msg: *E
 pub const ErrorMsg = struct {
     byte_offset: usize,
     msg: []const u8,
+    /// Children are always notes.
+    children: []ErrorMsg,
 
     pub fn create(gpa: *Allocator, byte_offset: usize, comptime format: []const u8, args: anytype) !*ErrorMsg {
         const self = try gpa.create(ErrorMsg);

```

Consider the following use case: An IDE wants to display a list of errors only; not notes, but then you can expand an individual error message like a tree and see all the notes.

Another example is having sibling notes such as "type declared here" or "the pointer element types have constness mismatch" alongside a stack trace such as many notes saying "called from here". An IDE would understand to display the first level children, and only expand the "called from here" when requested, due to its more deeply nested status.

Of course, with terminal output, it's less obvious, but I think it's still important to get the data structure right, and deal with representation as a separate concern. For example, even with terminal output, there could be a flag `-fdiagnostic-depth=100` which would control how many layers of "called from here" are displayed in the terminal, with some reasonable default.

PR #7570, opened 2020-12-27
stage2: make Alloc(Inferred) have mutabality info

g-w1:
This makes `Alloc` and `AllocInferred` their own ir instructions with a `bool` for mutable. This allows `analyzeInstAlloc` to check for mutability, fixing a TODO and paving the way for comptime constants that cannot be codegened.


Vexu:
Instead of adding a `mutable` field you should add an `.alloc_mutable`  instructions in the same way that `.block` and `.block_comptime` do it.

g-w1:
Ah, I assume this is to conserve memory, how there is .isnull and
.isnonnull.

Thanks

PR #7571, opened 2020-12-28
Add compileError message for StringHashMap in AutoHashMap

indocomsoft:
Follow up to #7483, since it seems that somehow I forgot to check in this change.


andrewrk:
Thanks for the follow-up!

PR #7572, opened 2020-12-28
fix race condition when starting (the event loop's) time wheel

kristoff-it:
This condition was correctly reported by TSAN.

See how `run()` checks the `is_running` flag right away.

### DIFF
--- a/lib/std/event/loop.zig
+++ b/lib/std/event/loop.zig
@@ -796,10 +796,13 @@ pub const Loop = struct {
                 .waiters = DelayQueue.Waiters{
                     .entries = std.atomic.Queue(anyframe).init(),
                 },
-                .thread = try std.Thread.spawn(self, DelayQueue.run),
+                .thread = undefined,
                 .event = std.AutoResetEvent{},
-                .is_running = true,
+                .is_running = undefined,
             };
+
+            @atomicStore(bool, &self.is_running, true, .SeqCst);
+            self.thread = try std.Thread.spawn(self, DelayQueue.run);
         }
 
         /// Entry point for the timer thread


daurnimator (commented):
If thread spawning fails does this mean that `is_running` is left as `true` when its not actually running?

lib/std/event/loop.zig
```diff
@@ -796,10 +796,13 @@ pub const Loop = struct {
...
                     .entries = std.atomic.Queue(anyframe).init(),
                 },
-                .thread = try std.Thread.spawn(self, DelayQueue.run),
+                .thread = undefined,
                 .event = std.AutoResetEvent{},
-                .is_running = true,
+                .is_running = undefined,
             };
+
+            @atomicStore(bool, &self.is_running, true, .SeqCst);
```
andrewrk, line 804: Does this need to be atomic? The fix is the fact that you start the thread after setting the field right?

kristoff-it: Isn't the compiler allowed to reorder operations unless explicitly stated otherwise through an atomic operation? I thought you always needed to "match" atomic reads and writes when there is concurrency.

Happy to be corrected though.

andrewrk: Hmm, I think we will define `std.Thread.spawn` to have a memory fence, so you would not need to worry about this particular boundary. This is already implicitly the case based on the fact that it invokes a syscall.

You do need to match atomic reads and writes, but the `is_running = true` "happens before" the thread spawn, and the atomic read "happens after" the thread spawning. So with a memory fence at thread spawn, this is fine.

daurnimator: What are you trying to order in relation to?

kristoff-it: @andrewrk done, thanks!

@daurnimator the `run` function checks that flag right after being spawned in a new thread.

kristoff-it:
> If thread spawning fails does this mean that `is_running` is left as `true` when its not actually running?

Yes, but since the init function has failed then you're not supposed to consider the struct instance valid memory. Same happens with every fallible init function that I've been in the standard library.

kristoff-it:
@andrewrk FYI I amended the commit to remove the `@atomicStore`. The PR mentions my force push but it seems it hasn't updated the web interface in any way.

andrewrk:
The `@atomicStore` is still there

Issue #7562, opened 2020-12-27
Unable to use std.heap.raw_c_allocator as backing allocator to std.heap.GeneralPurposeAllocator.


lithdew:
```zig
/// Asserts allocations are within `@alignOf(std.c.max_align_t)` and directly calls
/// `malloc`/`free`. Does not attempt to utilize `malloc_usable_size`.
/// This allocator is safe to use as the backing allocator with
/// `ArenaAllocator` and `GeneralPurposeAllocator`, and is more optimal in these cases
/// than to using `c_allocator`.
pub const raw_c_allocator = &raw_c_allocator_state;
var raw_c_allocator_state = Allocator{
    .allocFn = rawCAlloc,
    .resizeFn = rawCResize,
};
```

The following program crashes in spite of the usage guidelines documented for `std.heap.raw_c_allocator`. This is due to `std.heap.GeneralPurposeAllocator` requesting to allocate a 4096 byte-sized page on my Linux machine with an alignment of 4096, in spite of the `std.c.max_align_t` having an alignment of 16.

```zig
const std = @import("std");

pub fn main() !void {
    var gpa: std.heap.GeneralPurposeAllocator(.{}) = .{
        .backing_allocator = std.heap.raw_c_allocator,
    };
    defer if (gpa.deinit()) unreachable;

    const array = try gpa.allocator.alloc(u8, 1024);
    defer gpa.allocator.free(array);
}
```

```
$ zig run test.zig -lc
reached unreachable code
/home/kenta/Desktop/zig/lib/std/debug.zig:223:14: 0x203788 in std.debug.assert (test)
    if (!ok) unreachable; // assertion failure
             ^
/home/kenta/Desktop/zig/lib/std/heap.zig:178:11: 0x2037f8 in std.heap.rawCAlloc (test)
    assert(ptr_align <= @alignOf(std.c.max_align_t));
          ^
/home/kenta/Desktop/zig/lib/std/mem/Allocator.zig:294:40: 0x23760e in std.mem.Allocator.allocAdvancedWithRetAddr (test)
    const byte_slice = try self.allocFn(self, byte_count, a, len_align, return_address);
                                       ^
/home/kenta/Desktop/zig/lib/std/mem/Allocator.zig:262:41: 0x22ee45 in std.mem.Allocator.allocAdvanced (test)
    return self.allocAdvancedWithRetAddr(T, alignment, n, exact, @returnAddress());
                                        ^
[...]

PR #7576, opened 2020-12-28
std/heap: fix documentation of raw_c_allocator

ifreund:
This is not in fact safe to use with GeneralPurposeAllocator as GPA
requires align(page_size) but raw_c_allocator provides only
@alignOf(std.c.max_align_t).

See #7562


Vexu:
Whoops, didn't notice it claimed to be safe to use with `GeneralPurposeAllocator`.

PR #7577, opened 2020-12-28
implement emutls inside compiler_rt.zig

semarie:
This implementation follows LLVM compiler_rt emutls implementation (release 8.0 - MIT).

It is required for at least OpenBSD and Android (see #5921), but as I can't test for Android, so I don't have setup it for it.

There are still some `XXX` points which would need discussion:
- emutls needs to allocate data on heap: its owns data (a per thread array of pointers), and threadvar data (a block)
- threadvar data needs allocation at runtime know alignment. current `Allocator` interface doesn't permit that (only compile-know alignement), so the current code directly use `posix_memalign()` (like LLVM emutls.c)
- `emutls_control` struct is using data with special size: `gcc_word` which is defined in C as `typedef unsigned int gcc_word __attribute__((mode(word)))`. for now I just checked the value on OpenBSD x68_64 and use it

Regarding the code itself, the entrypoint of emutls is `__emutls_get_address()`. The structure `emutls_control` is part of the ABI (what GCC or LLVM will pass to `__emutls_get_address()` for asking a memory pointer).

Memory is automatically reclaimed when a thread is terminated (emutls memory *and* threadvar memory).

### DIFF
--- a/lib/std/special/compiler_rt/emutls.zig
+++ b/lib/std/special/compiler_rt/emutls.zig
@@ -1,0 +1,395 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2015-2018 LLVM Compiler Infrastructure
+// Copyright (c) 2020 Sebastien Marie <semarie@online.fr>
+// This file is part of [zig](https://ziglang.org/), which is MIT licensed.
+// The MIT license requires this copyright notice to be included in all copies
+// and substantial portions of the software.
+// __emutls_get_address specific builtin
+//
+// derived work from LLVM Compiler Infrastructure - release 8.0 (MIT)
+// https://github.com/llvm-mirror/compiler-rt/blob/release_80/lib/builtins/emutls.c
+//
+
+const std = @import("std");
+
+const abort = std.os.abort;
+const assert = std.debug.assert;
+const expect = std.testing.expect;
+
+// defined in C as:
+// typedef unsigned int gcc_word __attribute__((mode(word)));
+const gcc_word = usize;
+
+comptime {
+    assert(std.builtin.link_libc);
+}
+
+/// public entrypoint for generated code using EmulatedTLS
+pub fn __emutls_get_address(control: *emutls_control) callconv(.C) *c_void {
+    return control.getPointer();
+}
+
+/// Simple allocator interface, to avoid pulling in the while
+/// std allocator implementation.
+const simple_allocator = struct {
+    /// Allocate a memory chunk for requested type. Return a pointer on the data.
+    pub fn alloc(comptime T: type) *T {
+        return @ptrCast(*T, @alignCast(
+            @alignOf(T),
+            advancedAlloc(@alignOf(T), @sizeOf(T)),
+        ));
+    }
+
+    /// Allocate a slice of T, with len elements.
+    pub fn allocSlice(comptime T: type, len: usize) []T {
+        return @ptrCast([*]T, @alignCast(
+            @alignOf(T),
+            advancedAlloc(@alignOf(T), @sizeOf(T) * len),
+        ))[0 .. len - 1];
+    }
+
+    /// Allocate a memory chunk.
+    pub fn advancedAlloc(alignment: u29, size: usize) [*]u8 {
+        const minimal_alignment = std.math.max(@alignOf(usize), alignment);
+
+        var aligned_ptr: ?*c_void = undefined;
+        if (std.c.posix_memalign(&aligned_ptr, minimal_alignment, size) != 0) {
+            abort();
+        }
+
+        return @ptrCast([*]u8, aligned_ptr);
+    }
+
+    /// Resize a slice.
+    pub fn reallocSlice(comptime T: type, slice: []T, len: usize) []T {
+        var c_ptr: *c_void = @ptrCast(*c_void, slice.ptr);
+        var new_array: [*]T = @ptrCast([*]T, @alignCast(
+            @alignOf(T),
+            std.c.realloc(c_ptr, @sizeOf(T) * len) orelse abort(),
+        ));
+        return new_array[0..len];
+    }
+
+    /// Free a memory chunk allocated with simple_allocator.
+    pub fn free(ptr: anytype) void {
+        std.c.free(@ptrCast(*c_void, ptr));
+    }
+};
+
+/// Simple array of ?ObjectPointer with automatic resizing and
+/// automatic storage allocation.
+const ObjectArray = struct {
+    const ObjectPointer = *c_void;
+
+    // content of the array
+    slots: []?ObjectPointer,
+
+    /// create a new ObjectArray with n slots. must call deinit() to deallocate.
+    pub fn init(n: usize) *ObjectArray {
+        var array = simple_allocator.alloc(ObjectArray);
+        errdefer simple_allocator.free(array);
+
+        array.* = ObjectArray{
+            .slots = simple_allocator.allocSlice(?ObjectPointer, n),
+        };
+        errdefer simple_allocator.free(array.slots);
+
+        for (array.slots) |*object| {
+            object.* = null;
+        }
+
+        return array;
+    }
+
+    /// deallocate the ObjectArray.
+    pub fn deinit(self: *ObjectArray) void {
+        // deallocated used objects in the array
+        for (self.slots) |*object| {
+            simple_allocator.free(object.*);
+        }
+        simple_allocator.free(self.slots);
+        simple_allocator.free(self);
+    }
+
+    /// resize the ObjectArray if needed.
+    pub fn ensureLength(self: *ObjectArray, new_len: usize) *ObjectArray {
+        const old_len = self.slots.len;
+
+        if (old_len > new_len) {
+            return self;
+        }
+
+        // reallocate
+        self.slots = simple_allocator.reallocSlice(?ObjectPointer, self.slots, new_len);
+
+        // init newly added slots
+        for (self.slots[old_len..]) |*object| {
+            object.* = null;
+        }
+
+        return self;
+    }
+
+    /// Retrieve the pointer at request index, using control to initialize it if needed.
+    pub fn getPointer(self: *ObjectArray, index: usize, control: *emutls_control) ObjectPointer {
+        if (self.slots[index] == null) {
+            // initialize the slot
+            const size = control.size;
+            const alignment = @truncate(u29, control.alignment);
+
+            var data = simple_allocator.advancedAlloc(alignment, size);
+            errdefer simple_allocator.free(data);
+
+            if (control.default_value) |value| {
+                // default value: copy the content to newly allocated object.
+                @memcpy(data, @ptrCast([*]u8, value), size);
+            } else {
+                // no default: return zeroed memory.
+                @memset(data, 0, size);
+            }
+
+            self.slots[index] = @ptrCast(*c_void, data);
+        }
+
+        return self.slots[index].?;
+    }
+};
+
+// Global stucture for Thread Storage.
+// It provides thread-safety for on-demand storage of Thread Objects.
+const current_thread_storage = struct {
+    var key: std.c.pthread_key_t = undefined;
+    var init_once = std.once(current_thread_storage.init);
+
+    /// Return a per thread ObjectArray with at least the expected index.
+    pub fn getArray(index: usize) *ObjectArray {
+        if (current_thread_storage.getspecific()) |array| {
+            // we already have a specific. just ensure the array is
+            // big enough for the wanted index.
+            return array.ensureLength(index);
+        }
+
+        // no specific. we need to create a new array.
+
+        // make it to contains at least 16 objects (to avoid too much
+        // reallocation at startup).
+        const size = std.math.max(16, index);
+
+        // create a new array and store it.
+        var array: *ObjectArray = ObjectArray.init(size);
+        current_thread_storage.setspecific(array);
+        return array;
+    }
+
+    /// Return casted thread specific value.
+    fn getspecific() ?*ObjectArray {
+        return @ptrCast(
+            ?*ObjectArray,
+            @alignCast(
+                @alignOf(ObjectArray),
+                std.c.pthread_getspecific(current_thread_storage.key),
+            ),
+        );
+    }
+
+    /// Set casted thread specific value.
+    fn setspecific(new: ?*ObjectArray) void {
+        if (std.c.pthread_setspecific(current_thread_storage.key, @ptrCast(*c_void, new)) != 0) {
+            abort();
+        }
+    }
+
+    /// Initialize pthread_key_t.
+    fn init() void {
+        if (std.c.pthread_key_create(&current_thread_storage.key, current_thread_storage.deinit) != 0) {
+            abort();
+        }
+    }
+
+    /// Invoked by pthread specific destructor. the passed argument is the ObjectArray pointer.
+    fn deinit(arrayPtr: *c_void) callconv(.C) void {
+        var array = @ptrCast(
+            *ObjectArray,
+            @alignCast(@alignOf(ObjectArray), arrayPtr),
+        );
+        array.deinit();
+    }
+};
+
+const emutls_control = extern struct {
+    // A emutls_control value is a global value across all
+    // threads. The threads shares the index of TLS variable. The data
+    // array (containing address of allocated variables) is thread
+    // specific and stored using pthread_setspecific().
+
+    // size of the object in bytes
+    size: gcc_word,
+
+    // alignment of the object in bytes
+    alignment: gcc_word,
+
+    object: extern union {
+        // data[index-1] is the object address / 0 = uninit
+        index: usize,
+
+        // object address, when in single thread env (not used)
+        address: *c_void,
+    },
+
+    // null or non-zero initial value for the object
+    default_value: ?*c_void,
+
+    // global Mutex used to serialize control.index initialization.
+    var mutex: std.c.pthread_mutex_t = std.c.PTHREAD_MUTEX_INITIALIZER;
+
+    // global counter for keeping track of requested indexes.
+    // access should be done with mutex held.
+    var next_index: usize = 1;
+
+    /// Simple wrapper for global lock.
+    fn lock() void {
+        if (std.c.pthread_mutex_lock(&emutls_control.mutex) != 0) {
+            abort();
+        }
+    }
+
+    /// Simple wrapper for global unlock.
+    fn unlock() void {
+        if (std.c.pthread_mutex_unlock(&emutls_control.mutex) != 0) {
+            abort();
+        }
+    }
+
+    /// Helper to retrieve nad initialize global unique index per emutls variable.
+    pub fn getIndex(self: *emutls_control) usize {
+        // Two threads could race against the same emutls_control.
+
+        // Use atomic for reading coherent value lockless.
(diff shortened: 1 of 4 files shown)


lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+// derived work from LLVM Compiler Infrastructure - release 8.0 (MIT)
+// https://github.com/llvm-mirror/compiler-rt/blob/release_80/lib/builtins/emutls.c
+//
+
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+
+comptime {
+    assert(std.builtin.link_libc);
+    assert(std.Thread.use_pthreads);
```
LemonBoy, line 18: This is not needed, `use_pthreads` depends on `link_libc`.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+}
+
+/// public entrypoint for generated code using EmulatedTLS
+pub fn __emutls_get_address(control: *emutls_control) callconv(.C) *c_void {
+    const allocator = std.heap.c_allocator; // XXX
```
LemonBoy, line 23: I wouldn't use allocators here, use `c.malloc` (and realloc) directly to avoid pulling in the whole allocator implementation.
Keep in mind that you want to avoid as many abstractions as possible in `compiler_rt` to avoid as many failure modes as possible.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+
+    return control.get_pointer(allocator) catch {
+        @panic("emutls_control.get_pointer");
```
LemonBoy, line 26: My personal choice here is to use `os.abort()`, trying to trigger a panic here is hopeless as you may end up calling into `__emutls_get_address` again and get a nice infinite loop.

Ditto for all the uses of `@panic`.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+    pub inline fn get_pointer(self: *ObjectArray, index: usize, control: *emutls_control) !ObjectPointer {
+        if (self.slots[index] == null) {
+            // initialize the slot
+            const size = control.size;
+            const alignment = std.math.max(
+                @truncate(u29, control.alignment),
+                @alignOf(usize),
+            );
+
+            // XXX slot allocated using posix_memalign
```
LemonBoy, line 100: This is fine, either this or `c.malloc` is ok.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+
+        // init newly added slots
+        for (self.slots[old_size..]) |_object, i| {
+            self.slots[old_size + i] = null;
+        }
+    }
+
+    /// Retrieve the pointer at request index, using control to initialize it if needed.
+    pub inline fn get_pointer(self: *ObjectArray, index: usize, control: *emutls_control) !ObjectPointer {
+        if (self.slots[index] == null) {
```
LemonBoy, line 92: I like using and abusing `orelse` like this

```zig
return self.slots[index] orelse block: {
    // init code here...
    break :block xxx;
};
```

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+
+        if (old_size > new_size)
+            return;
+
+        // reallocate
+        self.slots = try self.allocator.reallocAtLeast(self.slots, new_size);
+
+        // init newly added slots
+        for (self.slots[old_size..]) |_object, i| {
+            self.slots[old_size + i] = null;
```
LemonBoy, line 86:
```suggestion
for (self.slots[old_size..]) |*object| {
            object.* = null;
}
```

Ditto for the other loops.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+        }
+    }
+
+    /// Retrieve the pointer at request index, using control to initialize it if needed.
+    pub inline fn get_pointer(self: *ObjectArray, index: usize, control: *emutls_control) !ObjectPointer {
```
LemonBoy, line 91: All the fn names use `camelCase`, not `snake_case`.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+    var key: std.c.pthread_key_t = undefined;
+    var init_once = std.once(ThreadStorage.init);
+
+    /// Return a per thread ObjectArray with at least the expected index.
+    pub inline fn get_array(allocator: *Allocator, index: usize) !*ObjectArray {
+        if (ThreadStorage.getspecific()) |array| {
+            // ensure the array is big enough for the wanted index
+            try array.ensureSize(index);
+            return array;
+        } else {
```
LemonBoy, line 134: No need for this `else` when the `if` body returns.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+    pub inline fn get_index(self: *emutls_control) usize {
+        // Two threads could race against the same emutls_control.
+
+        // Use atomic for reading coherent value lockless.
+        const index_lockless = @atomicLoad(usize, &self.object.index, .Acquire);
+
+        if (index_lockless != 0) {
+            // index is already initialized, return it.
+            return index_lockless;
+        } else {
```
LemonBoy, line 235: You can remove the `else` if the `if` body returns.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    pub fn unlock() void {
+        if (std.c.pthread_mutex_unlock(&Index.mutex) != 0) {
+            @panic("pthread_mutex_unlock");
+        }
+    }
+};
+
+const gcc_word = usize; // XXX typedef unsigned int gcc_word __attribute__((mode(word)));
```
LemonBoy, line 205: Please keep all the type definitions at the top of the file.

This is meant to be largest unsigned int that can fit in a register so for all we care (right now) `usize` is ok.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+comptime {
+    // XXX checked value for openbsd x86_64
+    assert(@sizeOf(gcc_word) == 8);
```
LemonBoy, line 208: This won't fly for 32bit targets.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+            self.slots[index] = @ptrCast(*c_void, aligned_ptr);
+        }
+
+        return self.slots[index].?;
+    }
+};
+
+// global stucture for ThreadStorage.
+// it provides thread-safety for on-demand storage of Thread Objects.
+const ThreadStorage = struct {
```
LemonBoy, line 124: Uppercase identifiers are reserved for types, not global singletons.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+    var key: std.c.pthread_key_t = undefined;
+    var init_once = std.once(ThreadStorage.init);
+
+    /// Return a per thread ObjectArray with at least the expected index.
+    pub inline fn get_array(allocator: *Allocator, index: usize) !*ObjectArray {
```
LemonBoy, line 129: No `inline` please, let the optimizer do its work.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+        return array;
+    }
+
+    /// deallocate the ObjectArray.
+    pub fn deinit(self: *ObjectArray) void {
+        const allocator = self.allocator;
+
+        // deallocated used objects in the array
+        for (self.slots) |_object, i| {
+            if (self.slots[i] != null) {
```
LemonBoy, line 63: Minor nit, you can skip the null check as `free` accepts `NULL` parameters.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2015-2018 LLVM Compiler Infrastructure
+// Copyright (c) 2020 Sebastien Marie <semarie@online.fr>
+// This file is part of [zig](https://ziglang.org/), which is MIT licensed.
+// The MIT license requires this copyright notice to be included in all copies
+// and substantial portions of the software.
+// __emutls_get_address specific builtin
+//
+// derived work from LLVM Compiler Infrastructure - release 8.0 (MIT)
```
LemonBoy, line 9: Are we copying old software now?

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+        return @ptrCast(
+            *T,
+            @alignCast(@alignOf(*T), try self.get_pointer(allocator)),
+        );
+    }
+};
+
+// ---- tests ----
+
+const assert = std.debug.assert;
```
LemonBoy, line 290: Declarations to the top.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+    }
+};
+
+// global counter for keeping track of requested indexes.
+const Index = struct {
+    var next: usize = 1;
+    var mutex: std.c.pthread_mutex_t = std.c.PTHREAD_MUTEX_INITIALIZER;
+
+    /// Return a next free index, incrementing global counter.
+    pub fn new() usize {
```
LemonBoy, line 184: If you just need to pull a fresh index I'd turn this into a single LOCK+GET+INCREMENT+UNLOCK sequence.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+
+    /// Initialize pthread_key_t.
+    fn init() void {
+        if (std.c.pthread_key_create(&ThreadStorage.key, ThreadStorage.deinit) != 0)
+            @panic("pthread_key_create");
+    }
+
+    /// Invoked by pthread specific destructor. the passed argument is the ObjectArray pointer.
+    fn deinit(arrayPtr: *c_void) callconv(.C) void {
+        var array = @ptrCast(
```
LemonBoy, line 170: I think it's about time this `@ptrCast(*T, @alignCast(@alignOf(T), ptr)` pattern is turned into a nice helper in `std.mem`.

And... are you sure `@alignOf(*ObjectArray)` is not `@alignOf(ObjectArray)` ?

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+
+    /// Return a per thread ObjectArray with at least the expected index.
+    pub inline fn get_array(allocator: *Allocator, index: usize) !*ObjectArray {
+        if (ThreadStorage.getspecific()) |array| {
+            // ensure the array is big enough for the wanted index
+            try array.ensureSize(index);
+            return array;
+        } else {
+            // a new array should contains at least 16 objects.
+            const size = std.math.max(16, index);
```
LemonBoy, line 136: So, if `index>16` you allocate only 16 slots and then leave `get_pointer` to read past `slots` end? `get_pointer` has no checks beside the ones inserted by the compiler in debug mode.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+};
+
+const gcc_word = usize; // XXX typedef unsigned int gcc_word __attribute__((mode(word)));
+comptime {
+    // XXX checked value for openbsd x86_64
+    assert(@sizeOf(gcc_word) == 8);
+}
+
+const emutls_control = extern struct {
+    // A emutls_control value is a global value across all
```
LemonBoy, line 212: General comment goes above `const emutls_control = extern struct`, the explanation for each field goes *above* the declaration.

LemonBoy (commented):
Nice work, I started implementing this but then life got in the way.

We need some way to force the emulated TLS on for some other platforms and/or get a OpenBSD CI machine to make sure this works (and keeps working) as intended.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,356 @@
...
+
+// ---- tests ----
+
+const assert = std.debug.assert;
+
+test "Index" {
+    Index.lock();
+    defer Index.unlock();
+
+    assert(Index.next == 1);
```
Vexu, line 296: Use `std.testing.expect`, `assert` becomes a no-op in `ReleaseFast` and `ReleaseSmall`.

lib/std/special/compiler_rt/emutls.zig
semarie: yes, and in fact, it is on purpose. More recent releases couldn't be used for derived work: else the file couldn't be under MIT licence. So old software is still good :smile:

lib/std/special/compiler_rt/emutls.zig
semarie: I am not sure to understand the exact context of the remark.

if it is just about the `new()` method, it is already done under locking (i will rename it to `new_locked()` to be more explicit).

if it is about the global index managment (`getIndex()`), grabing a full mutex would be too inefficient: the function is called at every threadvar access.

lib/std/special/compiler_rt/emutls.zig
semarie: > And... are you sure @alignOf(*ObjectArray) is not @alignOf(ObjectArray) ?

the passed argument is the same object set with `pthread_setspecific()`. I passed a `*ObjectArray`, so I should get a `*ObjectArray`. Regarding the alignment, I am a bit unexperimented with them. I am casting to `*ObjectArray` so I keep the same for the alignment. But I could be wrong.

lib/std/special/compiler_rt/emutls.zig
LemonBoy: > if it is just about the new() method, it is already done under locking (i will rename it to new_locked() to be more explicit).

Instead of requiring the caller to lock, call `new` then unlock the mutex my suggestion was to replace it with a single method like this:

```zig
fn next_index() usize {
    mutex.lock();
    defer mutex.unlock();
    const old = index;
    index += 1;
    return index;
}
```

After all you only need the mutex to make sure each index is unique.

lib/std/special/compiler_rt/emutls.zig
semarie: if `index>16`, `std.math.max()` will return index value, and not 16.

the initial size to 16 is used to avoid too much reallocation at startup.

lib/std/special/compiler_rt/emutls.zig
LemonBoy: This small snippet should help understand the subtle difference between the two:

```zig
    const A = struct {
        x: u32,
    };
    var x: A = undefined;
    @compileLog(@typeInfo(@TypeOf(&x)).Pointer.alignment); // 4
    @compileLog(@typeInfo(@TypeOf(@alignCast(@alignOf(A), &x))).Pointer.alignment); // 4
    @compileLog(@typeInfo(@TypeOf(@alignCast(@alignOf(*A), &x))).Pointer.alignment); // 8
```

lib/std/special/compiler_rt/emutls.zig
LemonBoy: Gah, my eyes kept reading `max` as `min`.
Yeah, 16 is enough for the common Zig program.

lib/std/special/compiler_rt/emutls.zig
semarie: the general usage of the index is: "get the index but if 0 initialize it with the next global value". with the fact that several threads could race to read/set the index.

the simple method could be:
- lock globally all threads for the global counter (it will also serialize access on control index)
- check the control index
- if 0, get the next index (and incr the global value)
- return the index (and unlock)

it is would involve taking a lock for every access of the index. lock couldn't be taken per variable, as `emutls_control` doesn't have structure to hold it.

the approch taken by LLVM emutls.c (and so by me too), is to use atomic reading lockless (so with possible races). in the general case (index already initialized) there is no problem and access is quick. for first access, it needs locking (and race checking as two threads could have read `0` index value and want to initialize it). so it can't be a all-in-one function for getting the next index.

eventually I could merge `Index.new()` inside `control.get_index()` function.

LemonBoy: Nevermind then, if multiple threads can race to initialize the same `emutls_control` this makes sense.
Packing `mutex` with the index is misleading, one is tempted to assume its only task is to protect the global counter while here it serves as a synchronization point for every thread/TLS object.

lib/std/special/compiler_rt/emutls.zig
semarie: Right: `@panic()` relies on a `threadvar`.

lib/std/special/compiler_rt/emutls.zig
semarie: I reorganized the code. the global variable for the index has been put toplevel. and the mutex is a global variable inside `emutls_control` and `lcok()` and `unlock()` helpers.

and now I wonder if I should put the global variable `global_index` inside `emutls_control` too.

semarie:
> We need some way to force the emulated TLS on for some other platforms and/or get a OpenBSD CI machine to make sure this works (and keeps working) as intended.

to force emulated TLS, it would need a new `-femulated-tls` and `-fno-emulated-tls` on zig command-line, and set option inside `ZigLLVMCreateTargetMachine()` too.

lib/std/special/compiler_rt/emutls.zig
semarie: well, it wasn't intented to be kept as it.

lib/std/special/compiler_rt/emutls.zig
semarie: I removed them. but I will look if it optimizer do its work well. The code path for the general case (index already initialized) could be called very often.

LemonBoy:
I can confirm this implementation works just fine on Linux w/ forced emulated TLS, thanks to this I discovered #7609.

semarie:
@andrewrk could you considere the PR to be merged ? (the freebsd CI failure isn't related)

andrewrk:
Yes I will have a look at this tomorrow!

semarie:
I folded few commits and rebased it on top of master. and maybe having a green CI and some activity might help to make progress on this.

andrewrk:
Nice work @semarie, thanks for your patience on the merge.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -0,0 +1,395 @@
...
+comptime {
+    assert(std.builtin.link_libc);
+}
+
+/// public entrypoint for generated code using EmulatedTLS
+pub fn __emutls_get_address(control: *emutls_control) callconv(.C) *c_void {
+    return control.getPointer();
+}
+
+/// Simple allocator interface, to avoid pulling in the while
```
daurnimator, line 32: typo: s/while/whole

Issue #7560, opened 2020-12-27
std.fs.Walker segfault


Cobertos:
Platform: Window 10.0.19041 x64
Zig Version: `0.7.1+8fc765387`

EDIT: Updated to latest version of zig

Trying to use `std.fs.Walker`. I'm getting a segmentation fault when traversing

```zig
const std = @import("std");

pub fn main () !void {
    const alloc = std.heap.page_allocator;
    var w = try std.fs.walkPath(alloc, "");
    while (try w.next()) |entry| {
        std.debug.print("Trying {}\n", .{ entry.path });
    }
}
```

...will crash with...

```shell
cobertos@COBERTOS-DELL-L C:\Users\cobertos\Downloads\renamer
$ "..\zig-windows-x86_64-0.7.1+8fc765387\zig.exe" run rename.zig
Trying \.git
Trying \.git\config
Trying \.git\description
Trying \.git\HEAD
Trying \.git\hooks
Trying \.git\hooks\applypatch-msg.sample
Trying \.git\hooks\commit-msg.sample
Trying \.git\hooks\fsmonitor-watchman.sample
Trying \.git\hooks\post-update.sample
Trying \.git\hooks\pre-applypatch.sample
Trying \.git\hooks\pre-commit.sample
Trying \.git\hooks\pre-merge-commit.sample
Trying \.git\hooks\pre-push.sample
Trying \.git\hooks\pre-rebase.sample
Trying \.git\hooks\pre-receive.sample
Trying \.git\hooks\prepare-commit-msg.sample
Trying \.git\hooks\update.sample
Trying \.git\index
Trying \.git\info
Trying \.git\info\exclude
Trying \.git\modules
Trying \.git\modules\vendor
Trying \.git\modules\vendor\libpcre.zig
Trying \.git\modules\vendor\libpcre.zig\config
Trying \.git\modules\vendor\libpcre.zig\description
Trying \.git\modules\vendor\libpcre.zig\HEAD
Trying \.git\modules\vendor\libpcre.zig\hooks
Trying \.git\modules\vendor\libpcre.zig\hooks\applypatch-msg.sample
Trying \.git\modules\vendor\libpcre.zig\hooks\commit-msg.sample
Trying \.git\modules\vendor\libpcre.zig\hooks\fsmonitor-watchman.sample
Trying \.git\modules\vendor\libpcre.zig\hooks\post-update.sample
Trying \.git\modules\vendor\libpcre.zig\hooks\pre-applypatch.sample
[...]

PR #7578, opened 2020-12-28
fix segfault in std.fs.Walker

vrischmann:
I'm not sure exactly what makes the test fail at 8 sub directories for me, it might not fail for other people.
The test also uses `testing.allocator` directly when calling `walkPath`, I wasn't able to reproduce it using `&arena.allocator`.

I've verified the fix on a couple of crashes using the code in #7560


andrewrk:
Thanks!

Issue #7218, opened 2020-11-25
stage2: a decl needs to have multiple compile errors attached to it


g-w1:
Right now in `Module.zig` the failed_decls field of `Module` is defined like this

```
failed_decls: std.AutoArrayHashMapUnmanaged(*Decl, *Compilation.ErrorMsg) = .{},
```

1 decl points to 1 error message. I can think of at least one use case where more than one error is needed. In @compileLog it gives a compile error, but doesn't stop the analysis. This means that we need somewhere to put multiple compile log errors, errors that don't fail a Decl, so more can be generated.

I tried to impliment this with a Decl pointing to and arraylist of Compilation.ErrorMsg, but I had to change a lot of things and it is using unnecessary memory for an arraylist because most errors will only be single.

My question is if there is another use/possibility for compile errors that don't stop analysis. If not, I think the best thing to do would be to make a separate field on Module for failed_nonstopping_decls or compile_log_decls with Decl -> ArrayList(ErrorMsg) so that we save memory.

This seems necessary for me to continue my compileLog pull request.

Edit: another use case is a switch on an error that doesn't cover all possibilities: 
```zig
./test.zig:4:5: error: error.A not handled in switch
    switch (a) {
./test.zig:4:5: error: error.C not handled in switch
    switch (a) {
```

andrewrk:
Sorry I meant to reply to this earlier. I disagree with decls having more than 1 compile errors attached to them. I think the compiler should stop trying to analyze a Decl once it hits a single compile error for it.

Furthermore, when a decl is referenced that has a compile error attached to it, that should return a "poison" value that indicates to stop analysis for the decl that is referencing the erroneous one.

1 compile error per Decl is plenty for the programmer to handle, and keeps the compiler implementation simpler, with less memory usage.

Vexu:
What about errors that have no effect on behavior like those for unused labels/variables?

I don't think we should be spending too much time optimizing error note memory usage since they are supposed to be rare.

andrewrk:
This is primarily about avoiding redundant compile errors for users, and compiler implementation simplicity. In C++ for example, the workflow is to fix the first compile error and then rebuild. In Zig the programmer should feel like fixing all the compile errors before doing another rebuild would be worth their time.

Good point about errors that have no effect on behavior; because of this I think it makes sense to have the poison state either stored separately or inferred from other state rather than based on existence of a compile error in the table.

g-w1:
[...]

PR #7587, opened 2020-12-29
stage2: @compileLog

g-w1:
This is an improvement from the previous pr because it gives Module a `compile_log_decls` field that stores a map from a Decl -> ArrayList(usize) for the locations of the compileLog in the Decl. This allows multiple `@compileLogs` to coexist with regular compile errors.

See #7218


andrewrk:
Landed in ab5f7b51566ad2c8e635e81b8515a97206daf8ab.

PR #7590, opened 2020-12-29
Note that stage1 needs to be updated in test runner

pixelherodev:
Useful if you happen to update standard library but not the compiler, or if you use a static compiler build :P


andrewrk:
This use case is not supported, always rebuild if you switch zig commits.

pixelherodev:
Its more than that though: without this PR, there's an implicit 
dependency on the official compiler. Attempting to compile with any other 
compiler would give a cryptic message.

However, the message used in this PR doesn't address that case either.

"The compiler needs to pass its path to the test runner" is probably more 
useful.

pixelherodev:
Admittedly that's not a big deal right now since no other compiler exists 
that would hit that error, but it'll be important long term.

Issue #2812, opened 2019-07-04
when multiple union fields share a body in a switch prong, use a peer result type cast, rather than requiring identical types


andrewrk:
Followup from #1107.

This test passes now:

```zig
test "switch prongs with cases with identical payload types" {
    const Union = union(enum) {
        A: usize,
        B: isize,
        C: usize,
    };
    const S = struct {
        fn doTheTest() void {
            doTheSwitch1(Union{ .A = 8 });
            doTheSwitch2(Union{ .B = -8 });
        }
        fn doTheSwitch1(u: Union) void {
            switch (u) {
                .A, .C => |e| {
                    expect(@typeOf(e) == usize);
                    expect(e == 8);
                },
                .B => |e| @panic("fail"),
            }
        }
        fn doTheSwitch2(u: Union) void {
            switch (u) {
                .A, .C => |e| @panic("fail"),
                .B => |e| {
                    expect(@typeOf(e) == isize);
                    expect(e == -8);
                },
            }
        }
    };
    S.doTheTest();
    comptime S.doTheTest();
}
```

But this diff:

```diff
-        A: usize,
+        A: u8,
        B: isize,
-        C: usize,
+        C: u16,
```

Gives this error:

```
/home/andy/downloads/zig/build/test.zig:18:28: error: capture group with incompatible types
                .A, .C => |e| {
                           ^
/home/andy/downloads/zig/build/test.zig:18:17: note: type 'u8' here
                .A, .C => |e| {
                ^
/home/andy/downloads/zig/build/test.zig:18:21: note: type 'u16' here
                .A, .C => |e| {
                    ^
```

However, I propose this should work. [peer type resolution](https://ziglang.org/documentation/master/#Peer-Type-Resolution) would make the type of `e` be `u16` and everything is fine. Likewise a `T` could be grouped with a `?T`.

fengb:
How about explicitly typing the coercion to prevent accidental rogue casts?
```zig
.A, .C => |e: u16|
```

andrewrk:
Can you give an example of an accidental rogue cast?

fengb:
I'm afraid of situations where changing the field type should trigger an error. Example:

```zig
const Union = union(enum) {
  .A: u8,
  .B: u8,
//  .C: u8,
  .C: u16,
  .D: u16,
};
```

[...]

PR #7592, opened 2020-12-29
Allow variable captures on multi-prong switch arms

LemonBoy:
ergonomic++


andrewrk:
Related proposal:  #2812

PR #7597, opened 2020-12-29
Static function declarations with no prototype should not be variadic

### LINKED ISSUE #7594: translate-c erroneously translates macro into a function with variadic parameters.
While attempting to import [libmdbx v0.9.2 (amalgmated source-code)](https://github.com/erthink/libmdbx/releases/tag/v0.9.2) into Zig, I hit the compile error:

```
./zig-cache/o/f8c145dcff9dc1971436ccf5076fab7f/cimport.zig:954:5: error: non-extern function is variadic
pub fn MDBX_MAP_RESIZED_is_deprecated(...) callconv(.C) c_int {
    ^
```

The following C code in `mdbx.h`:

```c
/** MDBX_MAP_RESIZED
 * \ingroup c_err
 * \deprecated Please review your code to use MDBX_UNABLE_EXTEND_MAPSIZE
 * instead. */
MDBX_DEPRECATED static __inline int MDBX_MAP_RESIZED_is_deprecated() {
  return MDBX_UNABLE_EXTEND_MAPSIZE;
}
#define MDBX_MAP_RESIZED MDBX_MAP_RESIZED_is_deprecated()
```

... got translated to:

```zig
pub fn MDBX_MAP_RESIZED_is_deprecated(...) callconv(.C) c_int {
    return MDBX_UNABLE_EXTEND_MAPSIZE;
}
```

ehaas:
If a static function is defined with no argument list and no prototype
is given, it should be treated as a function that takes no arguments
rather than as a variadic function.

Fixes #7594

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -5027,7 +5027,7 @@ fn transFnNoProto(
     is_pub: bool,
 ) !*ast.Node.FnProto {
     const cc = try transCC(rp, fn_ty, source_loc);
-    const is_var_args = if (fn_decl_context) |ctx| !ctx.is_export else true;
+    const is_var_args = if (fn_decl_context) |ctx| (!ctx.is_export and ctx.storage_class != .Static) else true;
     return finishTransFnProto(rp, null, null, fn_ty, source_loc, fn_decl_context, is_var_args, cc, is_pub);
 }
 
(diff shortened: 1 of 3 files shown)


andrewrk:
Thanks!

PR #7598, opened 2020-12-29
stage2: More improvements to self-hosted LLVM backend

FireFox317:
Basically we can now compile this program:
```zig
export fn _start() noreturn {
    var x: bool = true;
    var other: bool = foo(x);
    exit();
}

fn foo(cond: bool) bool {
    return !cond;
}

fn exit() noreturn {
    unreachable;
}
```
See commit messages for more explanation.

Edit: Now also includes the ability to test the LLVM backend!

### DIFF
--- a/src/Cache.zig
+++ b/src/Cache.zig
@@ -16,7 +16,8 @@ const Allocator = std.mem.Allocator;
 /// This protection is conditionally compiled depending on `want_debug_deadlock`.
 var all_cache_digest_set: std.AutoHashMapUnmanaged(BinDigest, void) = .{};
 var all_cache_digest_lock: std.Mutex = .{};
-const want_debug_deadlock = std.debug.runtime_safety;
+// TODO: Figure out how to make sure that `all_cache_digest_set` does not leak memory!
+pub const want_debug_deadlock = false;
 const DebugBinDigest = if (want_debug_deadlock) BinDigest else void;
 const null_debug_bin_digest = if (want_debug_deadlock) ([1]u8{0} ** bin_digest_len) else {};
 
--- a/src/llvm_backend.zig
+++ b/src/llvm_backend.zig
@@ -312,6 +312,7 @@ pub const LLVMIRModule = struct {
                         .arg => try self.genArg(inst.castTag(.arg).?),
                         .alloc => try self.genAlloc(inst.castTag(.alloc).?),
                         .store => try self.genStore(inst.castTag(.store).?),
+                        .load => try self.genLoad(inst.castTag(.load).?),
                         .dbg_stmt => blk: {
                             // TODO: implement debug info
                             break :blk null;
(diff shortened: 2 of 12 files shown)


src/llvm_backend.zig
```diff
@@ -338,42 +368,91 @@ pub const LLVMIRModule = struct {
...
+
+    fn genLoad(self: *LLVMIRModule, inst: *Inst.UnOp) !?*const llvm.ValueRef {
+        const ptr_val = try self.resolveInst(inst.operand);
+        return self.builder.buildLoad(ptr_val, "");
+    }
+
+    fn genBreakpoint(self: *LLVMIRModule, inst: *Inst.NoOp) !?*const llvm.ValueRef {
         // TODO: Store this function somewhere such that we dont have to add it again
         const fn_type = llvm.TypeRef.functionType(llvm.voidType(), null, 0, false);
         const func = self.llvm_module.addFunction("llvm.debugtrap", fn_type);
```
LemonBoy, line 438: Tip: LLVM has a much nicer API to fetch the intrinsic declarations, as you can see [here](https://llvm.org/doxygen/namespacellvm_1_1Intrinsic.html#a842176a2c8c60a94bc788454b0828ea2).

It already handles the memoization for you and lets you handle the other overloaded intrinsics without having to deal with strings and whatnot, check out how messy this code gets in `codegen.cpp`.

FireFox317: Ah thanks @LemonBoy, I did not know that! I indeed looked at how `codegen.cpp` currently implements it, but it would be better to use this new API. Will add in a follow up commit (No blocker for this PR)

FireFox317: Has been addressed in f49feefe692265b5c55c3378dd451b4068f17764

FireFox317:
I have pushed more commits that include functionality to test the LLVM backend using `zig build test-stage2 -Denable-llvm`!

src/Cache.zig
```diff
@@ -16,7 +16,8 @@ const Allocator = std.mem.Allocator;
 /// This protection is conditionally compiled depending on `want_debug_deadlock`.
 var all_cache_digest_set: std.AutoHashMapUnmanaged(BinDigest, void) = .{};
 var all_cache_digest_lock: std.Mutex = .{};
-const want_debug_deadlock = std.debug.runtime_safety;
+// TODO: Figure out how to make sure that `all_cache_digest_set` does not leak memory!
+pub const want_debug_deadlock = false;
```
FireFox317, line 20: @andrewrk To make sure that the compiler doesnt leak memory we have to call `deinit` on `all_cache_digest_set`, however when I tried to do that I ran into segfaults. I'm not sure how to solve that.

andrewrk: The call should go into cleanExit in main.zig. If you show how to repro the segfault I'll take a look

andrewrk:
Really nice work!!

Issue #7541, opened 2020-12-24
using `try` on a function with an empty inferred errorset, within a function that does not return an error union, doesn't cause a compile error


tgschultz:
```zig
pub fn isPtr(comptime T: type) bool {
    return std.meta.trait.is(.Pointer)(T);
}

pub fn foo(thing: anytype) !void {
    if(comptime isPtr(@TypeOf(thing))) {
        return error.IsPtr;
    }
}

pub fn main() void {
    const x = @as(usize, 10);
    
    //this causes a compile error, as it should
    //try foo(&x); 

    try foo(x); //this doesn't!
}
```

I believe this should cause a compile error, as an inferred empty errorset is not the same thing as no errorset.

Related: https://github.com/ziglang/zig/issues/5226

Vexu:
Also related #1386.

LemonBoy:
This is a side-effect of the aggressive constant-folding and the lazy analysis: `ir_analyze_instruction_test_err` folds the `testErr` instruction into a comptime-known `false` if the error set is empty, making the `try` essentially a no-op.

Removing the following lines makes the compiler emit an error for this test case (and uncovers a bunch of dead switch arms in the stdlib).

```c
             if (!type_is_global_error_set(err_set_type))
                 err_set_type->data.error_set.err_count == 0)
```

Is this a good solution? IMO it's not, optimizations should not mask problems in the compiler (unless this is intended, but I highly doubt so).

nektro:
correct behavior imo

Issue #3461, opened 2019-10-16
Compiler will out-of-memory with noreturn error union


MasterQ32:
When compiling the following program, the zig compiler will allocate a lot (> 24 GB) memory and will be killed by the OS. This is probably to some infinite loop.

```zig
extern fn someData() bool;

fn loop() !noreturn {
    while(true) {
        if(someData())
            return error.GenericFailure;
    }
}

export fn square() bool {
    return loop() catch false;
}

```

daurnimator:
@LemonBoy 
> Duplicate of #3461, the `!noreturn` combo should be disallowed.

Why?
Seems like a sensible return signature for something like `execv`?

LemonBoy:
> Why?

The `!` implies you may return an error, `noreturn` implies the function doesn't ever return.
If the `noreturn` is passed down to LLVM shit may hit the fan:

> This function attribute indicates that the function never returns normally, hence through a return instruction. This produces undefined behavior at runtime if the function ever does dynamically return. Annotated functions may still raise an exception, i.a., nounwind is not implied.

MasterQ32:
I don't think we should try to fit the LLVM detail here, because the message conveyed by `!noreturn` is exactly what you said:

> The ! implies you may return an error, noreturn implies the function doesn't ever return.

The function will either never return or throw an error, if so. So the whole non-catching part could be marked as *unreachable* by the compiler and emit the "code unreachable" error message.

Use case is pretty much any program that is not designed to ever quit working (like firmware or such), but may throw an error that is not a panic and need a soft-recover the system (as opposed to panics which are by-definition not recoverable)

LemonBoy:
#3263 is also a duplicate of this issue.

The `f() catch val` generates a pointer to `noreturn` that sends the analysis phase in a endless loop. Fixing this is quite tricky as we go directly from AST to a set of IR instructions that "explode" the error union when we don't know yet the type of `f()` nor of `val`.

This is another example where a multi-pass analysis design would have helped: the `catch` is desugared into a `IrCatchInst` and once its operands are analyzed we can selectively unwrap the pieces we need.

marler8997:
I tried using `!noreturn` today as I had a use case for it and it completely locked up my PC forcing me to restart it :(

I looked up to see if this was an existing issue and found this thread.

The usecase I had for it is I have an event loop function with sockets that's supposed to run forever unless it fails.  If it returns an error, the calling function will re-initialize and then call the function again, i.e.

```zig
[...]

PR #7606, opened 2020-12-30
stage2: astgen for `try`

g-w1:
I used `orelseCatchExpr` as inspiration for most of the `tryExpr` function. This is my first astgen expr, so please tell me if something is not idomatic.

I also moved the -femit-zir code higher up in the `updateModule` function so that even if sema fails, you still get the ir in $file.zir.

I couldn't find a way to test this.

### DIFF
--- a/src/astgen.zig
+++ b/src/astgen.zig
@@ -278,6 +278,7 @@ pub fn expr(mod: *Module, scope: *Scope, rl: ResultLoc, node: *ast.Node) InnerEr
         .For => return forExpr(mod, scope, rl, node.castTag(.For).?),
         .ArrayAccess => return arrayAccess(mod, scope, rl, node.castTag(.ArrayAccess).?),
         .Slice => return rlWrap(mod, scope, rl, try sliceExpr(mod, scope, node.castTag(.Slice).?)),
+        .Try => return tryExpr(mod, scope, rl, node.castTag(.Try).?),
         .Catch => return catchExpr(mod, scope, rl, node.castTag(.Catch).?),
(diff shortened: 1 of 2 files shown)


LemonBoy:
See #3461 for a stage1 bug related to how `try` is desugared.
See #7541 for another stage1 bug that's related to the very same thing, this time caused by the explosion of `try` in many simple (and unlinked) IR pieces that may get constant-folded separately,

g-w1:
Wow, everything is coming together. I was just thinking about how to implement inferred error sets yesterday. Both of these issues stem from that. The call that checks if the errorset matches the return type is this: https://github.com/ziglang/zig/pull/7606/files#diff-2600885f5a9e88bc263d2b47daab6d1c33418966af8f42b26c188597c108201bR930-R940 I think since I will probably also be implementing inferred error sets soon, I will keep this in the back of my head. If these bugs were to happen, they would not happen in astgen for `try` but in the coercion code and semantic analysis code for inferred/regular error sets.

LemonBoy:
> If these bugs were to happen, they would not happen in astgen for try but in the coercion code and semantic analysis code for inferred/regular error sets.

We'll see.

andrewrk (changes requested):
I'd like to get orelse/catch hooked up all the way through the compiler with test coverage before adding more astgen that uses similar logic, to make sure we are going in the right direction here.

g-w1:
I would too, #7628 will do that, then we can look at this.

andrewrk:
This PR will have to be re-done now that #7847 is landed.

Issue #5943, opened 2020-07-28
remove the "scope" parameter from std.log functions; introduce "std.log.scoped" functions


andrewrk:
[Accepted Proposal](https://github.com/ziglang/zig/issues/5943#issuecomment-669043489)

---------

In effort to make logging more mainstream and ergonomic, this is a proposal to simplify the logging API, yet still provide the advanced functionality.

It's difficult to justify the "scope" parameter, for example, in Hello World. Here I propose to do this:

```diff
diff --git a/lib/std/log.zig b/lib/std/log.zig
index d8bcba38c..3bd62e3c0 100644
--- a/lib/std/log.zig
+++ b/lib/std/log.zig
@@ -90,6 +90,8 @@ pub const default_level: Level = switch (builtin.mode) {
     .ReleaseSmall => .emerg,
 };
 
+const default_scope = .main;
+
 /// The current log level. This is set to root.log_level if present, otherwise
 /// log.default_level.
 pub const level: Level = if (@hasDecl(root, "log_level"))
@@ -117,7 +119,7 @@ fn log(
 
 /// Log an emergency message to stderr. This log level is intended to be used
 /// for conditions that cannot be handled and is usually followed by a panic.
-pub fn emerg(
+pub fn emergScope(
     comptime scope: @Type(.EnumLiteral),
     comptime format: []const u8,
     args: anytype,
@@ -126,9 +128,16 @@ pub fn emerg(
     log(.emerg, scope, format, args);
 }
 
+/// Log an emergency message to stderr. This log level is intended to be used
+/// for conditions that cannot be handled and is usually followed by a panic.
+pub fn emerg(comptime format: []const u8, args: anytype) void {
+    @setCold(true);
+    log(.emerg, default_scope, format, args);
+}
+
 /// Log an alert message to stderr. This log level is intended to be used for
 /// conditions that should be corrected immediately (e.g. database corruption).
-pub fn alert(
+pub fn alertScope(
     comptime scope: @Type(.EnumLiteral),
     comptime format: []const u8,
     args: anytype,
@@ -137,10 +146,17 @@ pub fn alert(
     log(.alert, scope, format, args);
 }
 
+/// Log an alert message to stderr. This log level is intended to be used for
+/// conditions that should be corrected immediately (e.g. database corruption).
+pub fn alert(comptime format: []const u8, args: anytype) void {
+    @setCold(true);
+    log(.alert, default_scope, format, args);
+}
+
[...]

PR #5941, opened 2020-07-27
Change std.debug.warn to std.debug.print in init-exe template.

nmichaels:

PR #7612, opened 2020-12-31
Implement #7296

### LINKED ISSUE #7296: change the default installation directory from ./zig-cache to ./
[Accepted Proposal](https://github.com/ziglang/zig/issues/7296#issuecomment-756432370)

----

If you run `zig init-exe` and then `zig build --help` you see this:

```
Steps:
  install (default)           Copy build artifacts to prefix path
  uninstall                   Remove build artifacts from prefix path
  run                         Run the app

General Options:
  -h, --help                  Print this help and exit
  --verbose                   Print commands before executing them
  --prefix [path]             Override default install prefix
  --search-prefix [path]      Add a path to look for binaries, libraries, headers
  --color [auto|off|on]       Enable or disable colored error messages

Project-Specific Options:
  -Dtarget=[string]           The CPU architecture, OS, and ABI to build for
  -Drelease-safe=[bool]       Optimizations on and safety on
  -Drelease-fast=[bool]       Optimizations on and safety off
  -Drelease-small=[bool]      Size optimizations on and safety off

Advanced Options:
...
```

As you can see, the default step is `install`. If you pass explicit steps on the command line, those steps are executed instead of the default. Currently, the install step has a default prefix path of `zig-cache`:

https://github.com/ziglang/zig/blob/fd18252a7165f04c07289c422eb0b808cc245b05/lib/std/build.zig#L194-L210

This proposal is to make the default installation directory named after the output mode:

https://github.com/ziglang/zig/blob/fd18252a7165f04c07289c422eb0b808cc245b05/lib/std/builtin.zig#L144-L149

So, if you were to run `zig build` with the defaults for init-exe, you would get:

```
Debug/bin/foo.exe
```

With `zig build -Drelease-safe` you would get:

```
ReleaseSafe/bin/foo.exe
```

However, I further propose that this changes when one takes advantage of `setPreferredReleaseMode`, in which case

```
Project-Specific Options:
  -Dtarget=[string]           The CPU architecture, OS, and ABI to build for
  -Drelease-safe=[bool]       Optimizations on and safety on
  -Drelease-fast=[bool]       Optimizations on and safety off
  -Drelease-small=[bool]      Size optimizations on and safety off
```

turns into

```
Project-Specific Options:
  -Dtarget=[string]           The CPU architecture, OS, and ABI to build for
  -Drelease=[bool]            Create a release build (ReleaseSafe)
```

where "ReleaseSafe" is whatever the preferred release mode is. In this case, `zig build -Drelease` would give you a default installation directory of "Release":

```
Release/bin/foo.exe
```

As a reminder, it is always possible to override the install prefix with the `--prefix` CLI parameter.

-----

Comments / suggestions / counter proposals welcome! Nobody will be accused of doing the B-word in this thread!

g-w1:
Closes #7296.
~~I also added a commit to change the default template from init-exe to use `std.debug.print` instead of `std.log.info`. This is so that it will  work in release modes other than `debug`. Found this when testing this. See https://freenode.irclog.whitequark.org/zig/2020-12-31#28714698; I noticed in 0.6.0 it used `std.debug.warn`, why did it change?~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,8 @@
 # -andrewrk
 
 zig-cache/
+/release/
+/debug/
 /build/
 /build-*/
 /docgen_tmp/
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -200,12 +200,16 @@ pub const Builder = struct {
             const install_prefix = self.install_prefix orelse "/usr";
             self.install_path = fs.path.join(self.allocator, &[_][]const u8{ dest_dir, install_prefix }) catch unreachable;
         } else {
-            const install_prefix = self.install_prefix orelse blk: {
-                const p = self.cache_root;
+            self.install_path = self.install_prefix orelse blk: {
+                const p = if (self.release_mode) |mode| switch (mode) {
+                    .Debug => "debug",
+                    .ReleaseSafe => "release",
+                    .ReleaseFast => "release",
+                    .ReleaseSmall => "release",
+                } else "debug";
                 self.install_prefix = p;
-                break :blk p;
+                break :blk self.pathFromRoot(p);
             };
-            self.install_path = install_prefix;
         }
         self.lib_dir = fs.path.join(self.allocator, &[_][]const u8{ self.install_path, "lib" }) catch unreachable;
         self.exe_dir = fs.path.join(self.allocator, &[_][]const u8{ self.install_path, "bin" }) catch unreachable;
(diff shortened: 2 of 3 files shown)


zigazeljko:
> I noticed in 0.6.0 it used `std.debug.warn`, why did it change?

`std.debug.warn` got deprecated in c3e0224.

g-w1:
> `std.debug.warn` got deprecated in c3e0224.
I understand this, what I really meant to ask was why didn't it change to std.debug.print?

squeek502:
Context for `std.log` being used in `init-exe`:
- https://github.com/ziglang/zig/issues/5943#issuecomment-664958579
- https://github.com/ziglang/zig/pull/5941#discussion_r467639132

PR #7615, opened 2020-12-31
openbsd: add pollfd interface

semarie:
unbreak openbsd build after e54fd2578195ce7857d921ac22b800d376fb870b : child_process.zig is using `poll()` and so needs `pollfd` struct to be defined now.


semarie:
freebsd.yml CI failed with:
```
+ release/bin/zig build test -Dskip-compile-errors -Dskip-non-native
1/30 test "self-hosted"... 
error: unknown target CPU 'athlon-xp'
```

which seems unrelated with this PR.

andrewrk:
Thanks!

PR #7616, opened 2020-12-31
stage2: inferred local variables

andrewrk:
This patch introduces the following new things:

Types:
 - inferred_alloc
   - This is a special value that tracks a set of types that have been stored
     to an inferred allocation. It does not support most of the normal type queries.
     However it does respond to `isConstPtr`, `ptrSize`, `zigTypeTag`, etc.
   - The payload for this type simply points to the corresponding Value
     payload.

Values:
 - inferred_alloc
   - This is a special value that tracks a set of types that have been stored
     to an inferred allocation. It does not support any of the normal value queries.

ZIR instructions:
 - store_to_inferred_ptr,
   - Same as `store` but the type of the value being stored will be used to infer
     the pointer type.
 - resolve_inferred_alloc
   - Each `store_to_inferred_ptr` puts the type of the stored value into a set,
     and then `resolve_inferred_alloc` triggers peer type resolution on the set.
     The operand is a `alloc_inferred` or `alloc_inferred_mut` instruction, which
     is the allocation that needs to have its type inferred.

Changes to the C backend:
 * Implements the bitcast instruction. If the source and dest types
   are both pointers, uses a cast, otherwise uses memcpy.
 * Tests are run with -Wno-declaration-after-statement. Someday we can
   conform to this but not today.

Here is an example case:

```zig
fn add(a: i32, b: i32) i32 {
    return a + b;
}

export fn main() c_int {
    const x = add(1, 2);
    return x - 3;
}
```

In ZIR form it looks like this:

```zir
fn_body main { // unanalyzed
  %0 = dbg_stmt()
=>%1 = alloc_inferred()
  %2 = declval_in_module(Decl(add))
  %3 = deref(%2)
  %4 = param_type(%3, 0)
  %5 = const(TypedValue{ .ty = comptime_int, .val = 1})
  %6 = as(%4, %5)
  %7 = param_type(%3, 1)
  %8 = const(TypedValue{ .ty = comptime_int, .val = 2})
  %9 = as(%7, %8)
  %10 = call(%3, [%6, %9], modifier=auto)
=>%11 = store_to_inferred_ptr(%1, %10)
=>%12 = resolve_inferred_alloc(%1)
  %13 = dbg_stmt()
  %14 = ret_type()
  %15 = const(TypedValue{ .ty = comptime_int, .val = 3})
  %16 = sub(%10, %15)
  %17 = as(%14, %16)
  %18 = return(%17)
} // fn_body main
```

After analysis:

```zir
@unnamed$0 = primitive(i32)
@unnamed$1 = fntype([@unnamed$0, @unnamed$0], @unnamed$0, cc=Unspecified)
@add = fn(@unnamed$1, {
  %0 = arg("a") ; deaths=0b0 0x7f9e951b8818
  %1 = arg("b") ; deaths=0b0 0x7f9e951b8840
  %2 = dbg_stmt() ; deaths=0b1000000000000000 0x7f9e951b8860
  %3 = add(%0, %1) ; deaths=0b11 0x7f9e951b8898
  %4 = return(%3) ; deaths=0b1000000000000001 0x7f9e951b88c0
})
@unnamed$3 = str("main")
@unnamed$4 = export(@unnamed$3, "main")
@unnamed$5 = single_mut_ptr_type(@unnamed$0)
@unnamed$6 = int(1)
@unnamed$7 = as(@unnamed$0, @unnamed$6)
@unnamed$8 = int(2)
@unnamed$9 = as(@unnamed$0, @unnamed$8)
@unnamed$10 = single_mut_ptr_type(@unnamed$0)
@unnamed$11 = int(3)
@unnamed$12 = as(@unnamed$0, @unnamed$11)
@unnamed$13 = primitive(c_int)
@unnamed$14 = fntype([], @unnamed$13, cc=Unspecified)
@main = fn(@unnamed$14, {
  %0 = dbg_stmt() ; deaths=0b1000000000000000 0x7f9e951757a0
  %1 = alloc(@unnamed$5) ; deaths=0b0 0x7f9e95183e18
  %2 = call(@add, [@unnamed$7, @unnamed$9], modifier=auto) ; deaths=0b111 0x7f9e95183f88
  %3 = bitcast(@unnamed$10, %1) ; deaths=0b1 0x7f9e951bc028
  %4 = store(%3, %2) ; deaths=0b1000000000000001 0x7f9e951bc048
  %5 = dbg_stmt() ; deaths=0b1000000000000000 0x7f9e951bc0a0
  %6 = sub(%2, @unnamed$12) ; deaths=0b11 0x7f9e951bc128
  %7 = intcast(@unnamed$13, %6) ; deaths=0b1 0x7f9e951bc150
  %8 = return(%7) ; deaths=0b1000000000000001 0x7f9e951bc170
})
```

You can see there is some dead code that might be nice to eliminate in a follow-up pass.

I have not played around with very many test cases yet. Some interesting
ones that I want to look at before merging:

```zig
var x = blk: {
  var y = foo();
  y.a = 1;
  break :blk y;
};
```

In the above test case, x and y are supposed to alias.

```zig
var x = if (bar()) blk: {
  var y = foo();
  y.a = 1;
  break :blk y;
} else blk: {
  var z = baz();
  z.b = 1;
  break :blk z;
};
```

In the above test case, x, y, and z are supposed to alias.

I also haven't tested with `var` instead of `const` yet.

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -275,6 +275,7 @@ pub fn generate(file: *C, module: *Module, decl: *Decl) !void {
         try writer.writeAll(" {");
 
         const func: *Module.Fn = func_payload.data;
+        //func.dump(module.*);
         const instructions = func.analysis.success.instructions;
         if (instructions.len > 0) {
             try writer.writeAll("\n");
@@ -285,6 +286,7 @@ pub fn generate(file: *C, module: *Module, decl: *Decl) !void {
                     .arg => try genArg(&ctx),
                     .assembly => try genAsm(&ctx, file, inst.castTag(.assembly).?),
                     .block => try genBlock(&ctx, file, inst.castTag(.block).?),
+                    .bitcast => try genBitcast(&ctx, file, inst.castTag(.bitcast).?),
                     .breakpoint => try genBreakpoint(file, inst.castTag(.breakpoint).?),
                     .call => try genCall(&ctx, file, inst.castTag(.call).?),
                     .cmp_eq => try genBinOp(&ctx, file, inst.castTag(.cmp_eq).?, "=="),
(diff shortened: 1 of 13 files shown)


src/codegen/c.zig
```diff
@@ -275,6 +275,7 @@ pub fn generate(file: *C, module: *Module, decl: *Decl) !void {
         try writer.writeAll(" {");
 
         const func: *Module.Fn = func_payload.data;
+        //func.dump(module.*);
```
FireFox317, line 278: Forgot to remove this?

PR #7622, opened 2020-12-31
std: Support equivalent ArrayList operations in ArrayHashMap

### LINKED ISSUE #7391: ArrayHashMap: support all operations that ArrayList support

tetsuo-cpp:
Closes #7391.

Part of the suggested changes was to rename `ArrayList.shrink` to `ArrayList.shrinkAndFree` so I've done that in a separate commit.

### DIFF
--- a/lib/std/array_hash_map.zig
+++ b/lib/std/array_hash_map.zig
@@ -323,7 +323,7 @@ pub fn ArrayHashMapUnmanaged(
         }
 
         pub fn clearAndFree(self: *Self, allocator: *Allocator) void {
-            self.entries.shrink(allocator, 0);
+            self.entries.shrinkAndFree(allocator, 0);
             if (self.index_header) |header| {
                 header.free(allocator);
                 self.index_header = null;
--- a/lib/std/array_hash_map.zig
+++ b/lib/std/array_hash_map.zig
@@ -99,6 +99,16 @@ pub fn ArrayHashMap(
             };
         }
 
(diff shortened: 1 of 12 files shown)


andrewrk (approved):
Really nice work!

Issue #2402, opened 2019-05-01
terminology update: use the phrase "detectable illegal behavior" rather than "safety-checked undefined behavior"


andrewrk:
The language reference makes use of the phrase "safety-checked undefined behavior" to mean "that which is undefined behavior in ReleaseFast and ReleaseSmall modes but will panic in Debug and ReleaseSafe modes".

This is a bit problematic because in the safe build modes, it's actually completely well-defined behavior. It will call the panic handler. Further, it will cause confusion because many people see "undefined behavior" and think it's a weakness of the language, when really it's the opposite - e.g. catching integer overflow bugs wouldn't be possible if it were always defined to be two's complement wraparound arithmetic.

I think a better phrase would be "illegal behavior". Illegal Behavior is always Undefined Behavior in the unsafe build modes. Illegal behavior at compile time is always a compile error. In safe build modes, runtime safety checks attempt to detect illegal behavior, but not all kinds of illegal behavior can be detected. Detected illegal behavior is well-defined and calls the panic handler. Undetected illegal behavior is undefined behavior, even in the safe build modes.

Instead of "safety-checked undefined behavior" the docs would say "detectable illegal behavior".

Related: #1966 #2301

Paul-Andre:
I'd use some other word instead of behavior to distinguish what you're asking the program to do and what it actually does. Perhaps "illegal operation" (an _illegal operation_ causes _undefined behavior_).

matklad:
Another language lawering argument for this: even if a hypothetical future alternative implementation of Zig defines some of these behaviors, we still want to declare them to be incorrect, and any program making use of them buggy.

matklad:
> the docs would say "detectable illegal behavior".

This is splitting hairs and I am not at all feeling strongly, but maybe just "checked illegal behavior"?

* It is simple
* It is more precise. detect**able** leaves the possibility open that the behavior isn't actually detected, while `checked` implies that it is always defined to trap (in release safe).

top-master:
How about "safety-checkable illegal behavior"?

Meaning, for "safety" reasons it is "checkable", however, it is "checked" only if the developer enables said checks at compile-time.

Don't use the word "operation" instead of "behavior" please, because for example,
if `a + b` operation overflows, then yes it's an undefined behavior, however,
if the numbers are always in a certain range, then the operation never overflows, hence,
the operation itself is never illegal nor undefined.

Further more:
* I suggest adding the `unchecked` keyword for functions.
[...]

PR #7624, opened 2020-12-31
Improve documentation for ArrayList, ArrayListUnmanaged, etc.

camconn:
I made some improvements to `ArrayList` and related structs' documentation. My goal here is to help users avoid footguns from referencing invalid pointers after an ArrayList's size changes. Here's [an example](https://github.com/ziglang/zig/issues/7560) of that in `std`.

If the changes I made don't fit the spirit of the documentation (I couldn't find any existing guidelines) then please let me know and I will happily make changes to fit these rules.

I could not make stronger statements on pointer safety since `resizeFn` varies between allocators (e.g. PageAllocator, FixedBufferAllocator). I also avoided being more explicit in changes to `ArrayListUnmanaged` since users of that struct will generally have a better idea of their use case.

Feedback, criticism, comments, etc. are appreciated.

### DIFF
--- a/lib/std/array_list.zig
+++ b/lib/std/array_list.zig
@@ -16,6 +16,10 @@ pub fn ArrayList(comptime T: type) type {
     return ArrayListAligned(T, null);
 }
 
+/// A contiguous, growable list of byte-aligned items in memory.
+/// This is a wrapper around an array of T values aligned to `alignment`-byte
+/// addresses. If the specified alignment is `null`, then `@alignOf(T)` is used.
+/// Initialize with `init`.
 pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
     if (alignment) |a| {
         if (a == @alignOf(T)) {
@@ -24,9 +28,13 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
     }
     return struct {
         const Self = @This();
-
-        /// Content of the ArrayList
+        /// Contents of the list. Pointers to elements in this slice are
+        /// **invalid after resizing operations** on the ArrayList, unless the
+        /// operation explicitly either: (1) states otherwise or (2) lists the
+        /// invalidated pointers.
         items: Slice,
+        /// How many T values this list can hold without allocating
+        /// additional memory.
         capacity: usize,
         allocator: *Allocator,
 
@@ -42,7 +50,7 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
             };
         }
 
-        /// Initialize with capacity to hold at least num elements.
+        /// Initialize with capacity to hold at least `num` elements.
         /// Deinitialize with `deinit` or use `toOwnedSlice`.
         pub fn initCapacity(allocator: *Allocator, num: usize) !Self {
             var self = Self.init(allocator);
@@ -79,11 +87,13 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
             };
         }
 
+        /// Creates a copy of this list, but with no memory managment.
+        /// Retains ownership of underlying memory.
         pub fn toUnmanaged(self: Self) ArrayListAlignedUnmanaged(T, alignment) {
             return .{ .items = self.items, .capacity = self.capacity };
         }
 
-        /// The caller owns the returned memory. ArrayList becomes empty.
+        /// The caller owns the returned memory. Empties this ArrayList.
         pub fn toOwnedSlice(self: *Self) Slice {
             const allocator = self.allocator;
             const result = allocator.shrink(self.allocatedSlice(), self.items.len);
@@ -91,7 +101,7 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
             return result;
         }
 
-        /// The caller owns the returned memory. ArrayList becomes empty.
+        /// The caller owns the returned memory. Empties this ArrayList.
         pub fn toOwnedSliceSentinel(self: *Self, comptime sentinel: T) ![:sentinel]T {
             try self.append(sentinel);
             const result = self.toOwnedSlice();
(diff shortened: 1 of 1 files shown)


lib/std/array_list.zig
```diff
@@ -16,6 +16,10 @@ pub fn ArrayList(comptime T: type) type {
     return ArrayListAligned(T, null);
 }
 
+/// A contiguous, growable list of byte-aligned items in memory.
```
andrewrk, line 19: "byte-aligned" means align(1) but here we are passing alignment as a comptime parameter

lib/std/array_list.zig
```diff
@@ -16,6 +16,10 @@ pub fn ArrayList(comptime T: type) type {
...
+/// This is a wrapper around an array of T values aligned to `alignment`-byte
+/// addresses. If the specified alignment is `null`, then `@alignOf(T)` is used.
+/// Initialize with `init`.
```
andrewrk, line 22: Mention that it stores an allocator, and that the unmanaged version can be used for one that accepts an allocator parameter for relevant functions.

lib/std/array_list.zig
```diff
@@ -24,9 +28,18 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
...
-
-        /// Content of the ArrayList
+        /// Contents of the list. Pointers to elements in this slice are
+        /// **invalid after resizing operations** on the ArrayList, unless the
+        /// operation explicitly either: (1) states otherwise or (2) lists the
+        /// invalidated pointers.
+        ///
+        /// The allocator used determines how element pointers are
+        /// invalidated, so the behavior may vary between lists. To avoid
+        /// undefined behavior, take into account the above paragraph plus the
```
andrewrk, line 38:
```suggestion
        /// illegal behavior, take into account the above paragraph plus the
```

See #2402 for context

lib/std/array_list.zig
```diff
@@ -79,19 +92,21 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
             };
         }
 
+        /// Creates a copy of this list, but with no memory managment.
```
andrewrk, line 95:
```suggestion
        /// Initializes an ArrayListUnmanaged with the `items` and `capacity` fields from this ArrayList.
```

"copy of this list" could be misleading

lib/std/array_list.zig
```diff
@@ -280,14 +298,13 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
...
-        /// The new elements have `undefined` values. This operation does not invalidate any
-        /// element pointers.
+        /// The new elements have `undefined` values. **Does not** invalidate pointers.
         pub fn expandToCapacity(self: *Self) void {
             self.items.len = self.capacity;
         }
 
         /// Increase length by 1, returning pointer to the new item.
-        /// The returned pointer becomes invalid when the list is resized.
+        /// See struct documentation for returned element pointer validity.
```
andrewrk, line 307: In general I prefer to put redundant docs in multiple places than to refer the programmer to go read docs somewhere else. For ArrayList, the struct docs you added are nice, but let's also repeat the invalidation warnings on every single function.

lib/std/array_list.zig
```diff
@@ -355,12 +376,22 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
     };
 }
 
-/// Bring-your-own allocator with every function call.
-/// Initialize directly and deinitialize with `deinit` or use `toOwnedSlice`.
+/// An `ArrayList`, but with bring-your-own allocator on every function call.
```
andrewrk, line 379:
```suggestion
/// An `ArrayList`, but the allocator is passed as a parameter to relevant functions rather than
/// stored in the struct itself. The same allocator must be used throughout the entire lifetime of
/// an ArrayList.
```

lib/std/array_list.zig
```diff
@@ -355,12 +376,22 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
...
+/// Initialize directly or with `initCapacity`, and deinitialize with `deinit`
+/// or use `toOwnedSlice`.
+///
+/// Generally you should prefer to use the analogous `ArrayList` unless
+/// you have a reason not to.
```
andrewrk, line 384: I don't agree with this advice. The API usage of ArrayListUnmanaged can sometimes actually be nicer than ArrayList.

andrewrk (changes requested):
Thanks for this! Some nice improvements here. I have left you some suggestions and then this should be good to go

camconn:
@andrewrk thanks for the feedback. I'll be fixing these things tomorrow.

lib/std/array_list.zig
kubkon: I agree with @andrewrk. For instance, in stage2 we make excessive use of `ArrayListUnmanaged` and friends which is simply easier to manage given a deep nesting of structs which all require an allocator, but at the same time, you'd like to be able to have them default-constructed without the necessity to pass in an explicit `Allocator` to the constructor (i.e., the `init` function).

lib/std/array_list.zig
camconn: Changing to "arbitrarily aligned" in lieu of better phrasing. Thoughts?

lib/std/array_list.zig
camconn: I had no clue about the preferred terminology. Thanks

lib/std/array_list.zig
camconn: I'll remove this advice.

lib/std/array_list.zig
camconn: done

camconn (commented):
I've implemented the changes you requested.

andrewrk (approved):
Thanks for the follow-up!

PR #7092, opened 2020-11-12
stage2: boolean type equality checking, compile time error comparison, validate switch errors, error set merging

g-w1:
The theme of this pr is on errors and types.

- [x] boolean type equality checking (`const b: bool = error {Thing1, Thing2} == error {Thing2, Thing1}`):
    - add error union and error set case to Type.eql()
- [x] compile time error comparison (`error.B == error.A`)
- [x] validate switch for errors
- [x] merge error sets (`var b: bool = error { Oof } || error { Other } == error { Oof, Other }`)
- [x] tests

PR #7628, opened 2021-01-01
stage2: catch at comptime!

g-w1:
improve stage2 to allow catch at comptime:

* add error_union value tag.
* add analyzeIsErr
* add Value.isError
* add TZIR wrap_errunion_payload and wrap_errunion_err for
  wrapping from T -> E!T and E -> E!T
* add anlyzeInstUnwrapErrCode and analyzeInstUnwrapErr
* add analyzeInstEnsureErrPayloadVoid:
* add wrapErrorUnion
* comparing error values for tests
* tests!!!

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -2536,7 +2536,15 @@ pub fn analyzeIsNull(
 }
 
 pub fn analyzeIsErr(self: *Module, scope: *Scope, src: usize, operand: *Inst) InnerError!*Inst {
-    return self.fail(scope, src, "TODO implement analysis of iserr", .{});
+    const ot = operand.ty.zigTypeTag();
+    if (ot != .ErrorSet and ot != .ErrorUnion) return self.constBool(scope, src, false);
+    if (ot == .ErrorSet) return self.constBool(scope, src, true);
+    assert(ot == .ErrorUnion);
+    if (operand.value()) |err_union| {
+        return self.constBool(scope, src, err_union.getError() != null);
+    }
+    const b = try self.requireRuntimeBlock(scope, src);
+    return self.addUnOp(b, src, Type.initTag(.bool), .is_err, operand);
 }
 
 pub fn analyzeSlice(self: *Module, scope: *Scope, src: usize, array_ptr: *Inst, start: *Inst, end_opt: ?*Inst, sentinel_opt: ?*Inst) InnerError!*Inst {
@@ -2839,6 +2847,52 @@ fn wrapOptional(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) !*In
     return self.addUnOp(b, inst.src, dest_type, .wrap_optional, inst);
 }
 
+fn wrapErrorUnion(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) !*Inst {
+    // TODO deal with inferred error sets
+    const err_union = dest_type.castTag(.error_union).?;
+    if (inst.value()) |val| {
+        const to_wrap = if (inst.ty.zigTypeTag() != .ErrorSet) blk: {
+            _ = try self.coerce(scope, err_union.data.payload, inst);
+            break :blk val;
+        } else switch (err_union.data.error_set.tag()) {
+            .anyerror => val,
+            .error_set_single => blk: {
+                const n = err_union.data.error_set.castTag(.error_set_single).?.data;
+                if (!mem.eql(u8, val.castTag(.@"error").?.data.name, n))
+                    return self.fail(scope, inst.src, "expected type '{}', found type '{}'", .{ err_union.data.error_set, inst.ty });
+                break :blk val;
+            },
+            .error_set => blk: {
+                const f = err_union.data.error_set.castTag(.error_set).?.data.typed_value.most_recent.typed_value.val.castTag(.error_set).?.data.fields;
+                if (f.get(val.castTag(.@"error").?.data.name) == null)
+                    return self.fail(scope, inst.src, "expected type '{}', found type '{}'", .{ err_union.data.error_set, inst.ty });
+                break :blk val;
+            },
+            else => unreachable,
+        };
+
+        return self.constInst(scope, inst.src, .{
+            .ty = dest_type,
+            // creating a SubValue for the error_union payload
+            .val = try Value.Tag.error_union.create(
+                scope.arena(),
+                to_wrap,
+            ),
+        });
+    }
+
+    const b = try self.requireRuntimeBlock(scope, inst.src);
+
+    // we are coercing from E to E!T
+    if (inst.ty.zigTypeTag() == .ErrorSet) {
+        var coerced = try self.coerce(scope, err_union.data.error_set, inst);
+        return self.addUnOp(b, inst.src, dest_type, .wrap_errunion_err, coerced);
+    } else {
+        var coerced = try self.coerce(scope, err_union.data.payload, inst);
+        return self.addUnOp(b, inst.src, dest_type, .wrap_errunion_payload, coerced);
+    }
+}
+
 fn makeIntType(self: *Module, scope: *Scope, signed: bool, bits: u16) !Type {
     const int_payload = try scope.arena().create(Type.Payload.Bits);
     int_payload.* = .{
@@ -2905,7 +2959,7 @@ pub fn resolvePeerTypes(self: *Module, scope: *Scope, instructions: []*Inst) !Ty
     return chosen.ty;
 }
 
-pub fn coerce(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) !*Inst {
+pub fn coerce(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) InnerError!*Inst {
     // If the types are the same, we can return the operand.
     if (dest_type.eql(inst.ty))
         return inst;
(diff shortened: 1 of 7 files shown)


src/zir_sema.zig
```diff
@@ -1207,7 +1225,22 @@ fn errorUnionPayloadPtr(mod: *Module, scope: *Scope, unwrap: *zir.Inst.UnOp, saf
...
     const tracy = trace(@src());
     defer tracy.end();
-    return mod.fail(scope, unwrap.base.src, "TODO implement zir_sema.errorUnionCode", .{});
+
+    const operand = try resolveInst(mod, scope, unwrap.positionals.operand);
+    if (operand.ty.zigTypeTag() != .ErrorUnion)
+        return mod.fail(scope, unwrap.base.src, "expected error union type, found '{}'", .{operand.ty});
+
+    if (operand.value()) |val| {
+        assert(val.getError() != null);
```
Vexu, line 1234: Why have an assert here but a compiler error in `errorUnionPayload`? Also both of these need to be updated to handle pointers now that `.err_union_payload_unsafe_ptr` and friends have been added.

g-w1: There is an assert here because errorUnionCode is only generated by the compiler when it knows that there *is* an error. It cannot be called by the user, unlike errorUnionPayload. Does this make sense?

g-w1: > Also both of these need to be updated to handle pointers now that `.err_union_payload_unsafe_ptr` and friends have been added.

I just rebased without any issues. What do I need to update?

Vexu: It seems that `errorUnionPayloadPtr` and `errorUnionCodePtr` were added instead of calling the same functions. You should probably implement those too.

g-w1: I did it, but I don't really know how to test it.

Vexu: For `errorUnionPayloadPtr` you just need to use address of on a catch expression like:
```zig
const a: anyerror!u32 = 0;
const b: *u32 = &(a catch unreachable);
```

`errorUnionCodePtr` probably shouldn't even exist since `catch` doesn't allow pointer payloads, not sure why it was ever added.

Vexu: `errorUnionCodePtr` definetly shouldn't exist:
```zig
const a: anyerror!u32 = error.
const b = &(a catch |err| @compileLog(@TypeOf(err)));
```
```
stage2:
Compile Log Output:
@as(type, *const anyerror)

stage1:
| anyerror
```

g-w1: Oh yeah, I just can't read :). https://github.com/ziglang/zig/blob/master/src/zir_sema.zig#L1224

src/zir_sema.zig
```diff
@@ -1204,34 +1204,116 @@ fn optionalPayload(
...
+        const data = val.castTag(.error_union).?.data;
```
Vexu, line 1272: This fails with `attempt to use null value` because you never actually wrap a value in this `error_union` value, either you need to make `wrapErrorUnion` actually wrap the value or remove it entirely.
```zig
const a: anyerror!u32 = error.Bar;
@compileLog(a catch |err| @compileLog(@TypeOf(err)));
```

g-w1: Ah I see, I didn't that extra cast.

g-w1: Ok, I think I fixed it by making wrapErrorUnion actually wrap it into an error_union SubValue.

src/Module.zig
```diff
@@ -2899,7 +2932,7 @@ pub fn resolvePeerTypes(self: *Module, scope: *Scope, instructions: []*Inst) !Ty
     return chosen.ty;
 }
 
-pub fn coerce(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) !*Inst {
+pub fn coerce(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) error{ OutOfMemory, AnalysisFail }!*Inst {
```
Vexu, line 2935:
```suggestion
pub fn coerce(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) InnerError!*Inst {
```

src/Module.zig
```diff
@@ -3011,7 +3049,7 @@ pub fn coerce(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) !*Inst
     return self.fail(scope, inst.src, "expected {}, found {}", .{ dest_type, inst.ty });
 }
 
-pub fn coerceNum(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) !?*Inst {
+pub fn coerceNum(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) error{ OutOfMemory, AnalysisFail }!?*Inst {
```
Vexu, line 3052:
```suggestion
pub fn coerceNum(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) InnerError!?*Inst {
```

src/Module.zig
```diff
@@ -2833,6 +2841,31 @@ fn wrapOptional(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) !*In
     return self.addUnOp(b, inst.src, dest_type, .wrap_optional, inst);
 }
 
+fn wrapErrorUnion(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) !*Inst {
+    const err_union = dest_type.castTag(.error_union).?;
```
Vexu, line 2845: This function needs to check that if `inst.ty == .ErrorSet` that its type is a subset of the error unions error set. Would also be helpful to add a "TODO resolve inferred error set" comment.

src/astgen.zig
```diff
@@ -1762,8 +1762,9 @@ const CondKind = union(enum) {
 
     fn thenSubScope(self: CondKind, mod: *Module, then_scope: *Scope.GenZIR, src: usize, payload_node: ?*ast.Node) !*Scope {
         if (self == .bool) return &then_scope.base;
-
-        const payload = payload_node.?.castTag(.PointerPayload) orelse {
+        const payload = if (payload_node) |node|
```
Vexu, line 1765: We return early if `self == .bool` so `payload_node` must be non-null.

src/value.zig
```diff
@@ -1820,6 +1846,87 @@ pub const Value = extern union {
...
+            .float_32,
+            .float_64,
+            .float_128,
+            .void_value,
+            .enum_literal,
+            .error_set,
+            .empty_struct_value,
+            => null,
+
+            .error_union => self.castTag(.error_union).?.data.getError(),
```
Vexu, line 1923: This should only check if the payload is an `.@"error"`, calling this recursively will cause code like
```zig
const a: anyerror!anyerror!u32 = @as(anyerror!u32, error.Foo);
const ignore: anyerror!u32 = a catch unreachable;
```
to fail.

Vexu (changes requested):
This is nearly ready to go.

src/value.zig
Vexu: Also note that if #7913 gets rejected we will have to make a distinction on whether the value is the error code or the payload.

src/Module.zig
g-w1: Does it also have to check that if `inst.ty != .ErrorSet` the payload has to be the same?

Vexu: It has to check that the value can be coerced to the payload type.

src/astgen.zig
g-w1: Oh huh, it triggered the .? before but now it doesn't. Thanks.

src/Module.zig
g-w1: > This function needs to check that if `inst.ty == .ErrorSet` that its type is a subset of the error unions error set. Would also be helpful to add a "TODO resolve inferred error set" comment.

The hacky way to implement this is to check in `wrapErrorUnion`, but the better way is to implement coercion for error unions like I did in #7092. Ill do it hackily for now, but it should be revisited.

g-w1:
> This is nearly ready to go.

Is there anything else I can do?

src/Module.zig
```diff
@@ -2836,6 +2844,56 @@ fn wrapOptional(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) !*In
...
+fn wrapErrorUnion(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) !*Inst {
+    // TODO deal with inferred error sets
+    const err_union = dest_type.castTag(.error_union).?;
+    if (inst.value()) |val| {
+        return self.constInst(scope, inst.src, .{
+            .ty = dest_type,
+            // creating a SubValue for the error_union payload
+            .val = try Value.Tag.error_union.create(
+                scope.arena(),
+                blk: {
```
Vexu, line 2856: Please use temporary variables to reduce nesting.

src/Module.zig
```diff
@@ -2836,6 +2844,56 @@ fn wrapOptional(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) !*In
...
+                    if (inst.ty.zigTypeTag() != .ErrorSet) {
+                        // .? because we know it has a value
```
Vexu, line 2858: What is this comment referring to?

src/Module.zig
```diff
@@ -2836,6 +2844,56 @@ fn wrapOptional(self: *Module, scope: *Scope, dest_type: Type, inst: *Inst) !*In
...
+                        _ = try self.coerce(scope, err_union.data.payload, inst);
+                        break :blk val;
+                    } else {
+                        switch (err_union.data.error_set.tag()) {
+                            // TODO these should be moved to coerce with error union widening
```
Vexu, line 2863: Why? This seems correct to me.

g-w1: Hmm, I guess `wrapErrorUnion` is a subset of coerce so its fine. If we see there is repeated code, we can always factor it out.

src/Module.zig
g-w1: Nothing, I must have left it in from something else.

Vexu (approved):

Issue #7630, opened 2021-01-01
Can't build Zig stage 2 with GCC x86


Prince213:
Using GCC from https://github.com/brechtsanders/winlibs_mingw/releases/download/10.2.0-11.0.0-8.0.0-r5/winlibs-i686-posix-dwarf-gcc-10.2.0-mingw-w64-8.0.0-r5.7z.
Build command:
```
cmake ../../../../src/zig -GNinja -DCMAKE_INSTALL_PREFIX=E:\zig\i686 -DCMAKE_PREFIX_PATH=E:\deps\i686 -DCMAKE_BUILD_TYPE=Release "-DCMAKE_CXX_FLAGS=-D_WIN32_WINNT=0x601"
cmake --build . -j
```
Log:
```
[161/163] Building self-hosted component E:/build/i686/zig/zig1.o
FAILED: zig1.o
cmd.exe /C "cd /D E:\src\zig && E:\build\i686\zig\zig0.exe src/stage1.zig -target native -mcpu=baseline --name zig1 --override-lib-dir E:/src/zig/lib -femit-bin=E:/build/i686/zig/zig1.o -OReleaseFast --strip -lc --pkg-begin build_options E:/build/i686/zig/config.zig --pkg-end --pkg-begin compiler_rt E:/src/zig/lib/std/special/compiler_rt.zig --pkg-end"
.\src\link\MachO\imports.zig:60129542314:21452957916594176: error: expected type 'usize', found 'u64'
        size += 1 + stream.bytes_written;
             ^
.\src\link\MachO\imports.zig:60129542314:179036880063822936: note: unsigned 32-bit int cannot represent all possible unsigned 64-bit values
        size += 1 + stream.bytes_written;
             ^
.\src\link\MachO.zig:77309413605:179036880063822936: note: referenced here
    const size = try self.binding_info_table.calcSize();
                 ^
.\src\link\MachO.zig:55834575230:179036880063822936: note: referenced here
            try self.writeBindingInfoTable();
            ^
.\src\link.zig:55834575329:179036880063822936: note: referenced here
            try base.flushModule(comp);
            ^
.\src\Compilation.zig:21474837792:179036880063822936: note: referenced here
    try self.bin_file.flush(self);
    ^
.\src\main.zig:21474838428:179036880063822936: note: referenced here
    try comp.update();
    ^
.\src\main.zig:210453399300:179036880063822936: note: referenced here
[...]

PR #7631, opened 2021-01-01
libstd: store bytes written as usize in counting writer

kubkon:
`std.io.counting_writer` currently stores bytes written as `u64` while
`std.io.counting_writer.write` returns bytes written as `usize`. This
commit suggests unifying both to always use `usize`. I think this makes
sense since on 32bit platforms will seldom (if ever) write more than
32bit worth of payload.

This fixes `MachO.zig` struct compilation errors on 32bit platforms,
and addresses partially issue #7630.

cc @LemonBoy

### DIFF
--- a/lib/std/io/counting_writer.zig
+++ b/lib/std/io/counting_writer.zig
@@ -10,7 +10,7 @@ const testing = std.testing;
 /// A Writer that counts how many bytes has been written to it.
 pub fn CountingWriter(comptime WriterType: type) type {
     return struct {
-        bytes_written: u64,
+        bytes_written: usize,
         child_stream: WriterType,
 
         pub const Error = WriterType.Error;


LemonBoy:
> std.io.counting_writer currently stores bytes written as u64 while
std.io.counting_writer.write returns bytes written as usize. This
commit suggests unifying both to always use usize

The `u64` is correct, the user may write more than 4GB even on a 32bit system. `write` returns a `usize` instead of a `u64` because it writes a slice or part of it to a stream, the slice being a memory object means its length is capped by `usize`.

kubkon:
> > std.io.counting_writer currently stores bytes written as u64 while
> > std.io.counting_writer.write returns bytes written as usize. This
> > commit suggests unifying both to always use usize
> 
> The `u64` is correct, the user may write more than 4GB even on a 32bit system. `write` returns a `usize` instead of a `u64` because it writes a slice or part of it to a stream, the slice being a memory object means its length is capped by `usize`.

Fair point. Is it practical though? I'm happy to leave it as-is and fix the Mach-O itself, just curious whether we expect this to happen often or not.

kubkon:
Or put it another way, what would such writes look like? To a series of file descriptors, or a stream for instance?

LemonBoy:
> Fair point. Is it practical though? I'm happy to leave it as-is and fix the Mach-O itself, just curious whether we expect this to happen often or not.

It can definitely happen, limiting the file sizes/offsets to 32bit quantities was a big error that was slowly amended in various libcs/kernels, that's why we have many `64`-suffixed I/O functions.

> Or put it another way, what would such writes look like? To a series of file descriptors, or a stream for instance?

One possible example is a loop copying a stream into another stream, if the source stream is 0xffffffff+1 bytes long you wouldn't want to get an integer overflow error nor a silent wraparound to zero.

kubkon:
> > Fair point. Is it practical though? I'm happy to leave it as-is and fix the Mach-O itself, just curious whether we expect this to happen often or not.
> 
> It can definitely happen, limiting the file sizes/offsets to 32bit quantities was a big error that was slowly amended in various libcs/kernels, that's why we have many `64`-suffixed I/O functions.
> 
> > Or put it another way, what would such writes look like? To a series of file descriptors, or a stream for instance?
> 
> One possible example is a loop copying a stream into another stream, if the source stream is 0xffffffff+1 bytes long you wouldn't want to get an integer overflow error nor a silent wraparound to zero.

Makes sense! Thanks for the explanation! I'm dropping this PR in favour of #7632 then! Thanks @LemonBoy!

PR #7632, opened 2021-01-01
macho: unblock stage2 on 32bit platforms

### LINKED ISSUE #7630: Can't build Zig stage 2 with GCC x86

kubkon:
Unblocks compilation of stage2 on 32bit platforms, and fixes #7630.

cc @LemonBoy: an alternative to #7631 that doesn't touch libstd.

### DIFF
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -907,7 +907,7 @@ fn linkWithLLD(self: *MachO, comp: *Compilation) !void {
                     const size = try self.binding_info_table.calcSize();
                     assert(dyld_info.bind_size >= size);
 
-                    var buffer = try self.base.allocator.alloc(u8, size);
+                    var buffer = try self.base.allocator.alloc(u8, @intCast(usize, size));
                     defer self.base.allocator.free(buffer);
 
                     var stream = std.io.fixedBufferStream(buffer);
@@ -919,7 +919,7 @@ fn linkWithLLD(self: *MachO, comp: *Compilation) !void {
                     const size = try self.lazy_binding_info_table.calcSize();
                     assert(dyld_info.lazy_bind_size >= size);
 
-                    var buffer = try self.base.allocator.alloc(u8, size);
+                    var buffer = try self.base.allocator.alloc(u8, @intCast(usize, size));
                     defer self.base.allocator.free(buffer);
 
                     var stream = std.io.fixedBufferStream(buffer);
@@ -1875,13 +1875,13 @@ pub fn makeStaticString(comptime bytes: []const u8) [16]u8 {
 
 fn makeString(self: *MachO, bytes: []const u8) !u32 {
     try self.string_table.ensureCapacity(self.base.allocator, self.string_table.items.len + bytes.len + 1);
-    const result = self.string_table.items.len;
+    const result = @intCast(u32, self.string_table.items.len);
     self.string_table.appendSliceAssumeCapacity(bytes);
     self.string_table.appendAssumeCapacity(0);
     self.string_table_dirty = true;
     if (self.d_sym) |*ds|
         ds.string_table_dirty = true;
-    return @intCast(u32, result);
+    return result;
 }
 
 fn getString(self: *MachO, str_off: u32) []const u8 {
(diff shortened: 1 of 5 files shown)


daurnimator (commented):
Can MachO files be larger than 4GB? if so perhaps much of the macho code needs to have `usize` replaced with `u64`?

kubkon:
> Can MachO files be larger than 4GB? if so perhaps much of the macho code needs to have `usize` replaced with `u64`?

Excellent question. So it depends. If you look at `macho.segment_command_64`, file offset and size are `u64` so that's affirmative. If you look at `macho.dylib_info_command`, file offset and size are `u32` so that's a negative. In case of export trie and binding info, the structures expect a `u32` values; however, I'd like to be consistent with `std.io.counting_writer` so I'll do as you suggest and have `calcSize` return `u64` and `@intCast` when inserting into an appropriate structure.

kubkon:
On second thought, I guess there's no reason to simply not return `u32` in relevant places which deal with file offsets and sizes. @daurnimator could you have a look and check I make sense and didn't make any silly mistakes please?

daurnimator:
> I guess there's no reason to simply not return `u32` in relevant places which deal with file offsets and sizes.

There would be if there was *any* way for those file offsets to be >4GB.

kubkon:
> > I guess there's no reason to simply not return `u32` in relevant places which deal with file offsets and sizes.
> 
> There would be if there was _any_ way for those file offsets to be >4GB.

Given that every file offset is with respect to the beginning of the file and *not* the beginning of the segment, I don't think that's possible. It's weird and makes wonder if this is perhaps a mistake on Apple's part? VM addressing on the other hand is relative to the start address of the respective segment.

kubkon:
OK, since I'm not 100% certain what a >4GB Mach-O would look like, I'm going to follow the libstd convention; i.e., reads return bytes written as `usize` and writes `u64`, with appropriate casts to `usize` in places where this is required such as `alloc`s. I think this is the safest approach for now which unblocks building stage2 on 32bit platforms.

PR #7634, opened 2021-01-01
Add PackedIntArray .initAllTo function

daurnimator:
Useful helper function.

Putting up this PR now as this gets close to tickling a stage1 bug:
Attempting `comptime PackedArray.initAllTo(5)` results in
```
Semantic Analysis [947/1269] Assertion failed at /home/daurnimator/src/zig/src/stage1/codegen.cpp:7611 in gen_const_val. This is a bug in the Zig compiler.
```

### DIFF
--- a/lib/std/packed_int_array.zig
+++ b/lib/std/packed_int_array.zig
@@ -205,7 +205,8 @@ pub fn PackedIntArrayEndian(comptime Int: type, comptime endian: builtin.Endian,
 
         ///Initialize all entries of a packed array to the same value
         pub fn initAllTo(int: Int) Self {
-            var self = @as(Self, undefined);
+            // TODO: use `var self = @as(Self, undefined);` https://github.com/ziglang/zig/issues/7635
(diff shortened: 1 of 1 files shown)


lib/std/packed_int_array.zig
```diff
@@ -365,7 +380,15 @@ test "PackedIntArray init" {
     const PackedArray = PackedIntArray(u3, 8);
     var packed_array = PackedArray.init([_]u3{ 0, 1, 2, 3, 4, 5, 6, 7 });
     var i = @as(usize, 0);
-    while (i < packed_array.len()) : (i += 1) testing.expect(packed_array.get(i) == i);
+    while (i < packed_array.len()) : (i += 1) testing.expectEqual(@intCast(u3, i), packed_array.get(i));
+}
+
+test "PackedIntArray initAllTo" {
+    if (we_are_testing_this_with_stage1_which_leaks_comptime_memory) return error.SkipZigTest;
```
daurnimator, line 387: I'm not 100% sure why we need this on these smaller test cases; I just copied convention from the test case preceding it.

PR #4215, opened 2020-01-18
cmake: support `make` and `make install`

mikdusan:
(2nd attempt to get this right)

Reproduced the issue of linux/static builds in docker. The cause was two-fold:

1. my local environment has `.` in `$PATH` and removing that I was able to get the same failure
2. by invoking cmake `execute_process` via `install(CODE...)` there is really no decent chance to get an error code of the child process and we had no report when `zig0` was not found in path

The solution is:

1. force the script to use an absolute pathname to `zig0`
2. switch from `install(CODE...)` to `install(SCRIPT...)` and have the sub-script check for child process exit status explicitly. And some nice error messages should help avoid a repeat.

(the child process in question here was `zig0 build libuserland install ....`)

---

tested with stripped down `PATH` environment variable:

- macOS 10.15.2 w/ Xcode 11.3.1
- archlinux w/ gcc 9.2.0
- Windows 10 w/ msvc v16.5.0-pre.1.0

<br/>

- [x] `make` skips install { `bin`, `lib` }
- [x] `make install` installs { `bin`, `lib` }
- [x] `make` followed by `make install` installs { `bin`, `lib` }
- [x]  delete random files from install dir, `make install` re-installs missing files

#### not defined: `ZIG_SKIP_INSTALL_LIB_FILES`
- [x] `msbuild ... INSTALL.vcxproj` installs { `bin`, `lib` }
- [x]  delete random files from install dir, `msbuild ... INSTALL.vcxproj` re-installs missing files

#### defined: `ZIG_SKIP_INSTALL_LIB_FILES=ON`
- [x] `msbuild ... INSTALL.vcxproj` installs `bin`, skips `lib`

PR #7637, opened 2021-01-01
Bring back ZIG_SKIP_INSTALL_LIB_FILES

daurnimator:
I think this was accidentally removed for non-MSVC by @mikdusan in https://github.com/ziglang/zig/pull/4215#issuecomment-753330074 ?

(It was also touched by 7eb0a3edcef2ab752ff1e45e6f1316b633b29606 but that just brought the help description up to date with its accidental removal)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -780,31 +780,32 @@ endif()
 
 install(TARGETS zig DESTINATION bin)
 
-set(ZIG_INSTALL_ARGS "build"
-    --override-lib-dir "${CMAKE_SOURCE_DIR}/lib"
-    "-Dlib-files-only"
-    --prefix "${CMAKE_INSTALL_PREFIX}"
-    "-Dconfig_h=${ZIG_CONFIG_H_OUT}"
-    install
-)
+set(ZIG_SKIP_INSTALL_LIB_FILES off CACHE BOOL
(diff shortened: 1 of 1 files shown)


mikdusan:
Why? I was under the impression `ZIG_SKIP_INSTALL_LIB_FILES` mechanism caused friction to maintainers and the two-step `make` + `make install` is desired.

daurnimator:
> I was under the impression `ZIG_SKIP_INSTALL_LIB_FILES` mechanism caused friction to maintainers and the two-step `make` + `make install` is desired.

I cause no friction to maintainers: they don't use it.

---

I use it for local development: I found that without this PR, working on parts of the standard library, I had to re-run `make` to use them. That took my development feedback cycle up to the scale of minutes whereas with this patch it is milliseconds.

andrewrk:
ZIG_SKIP_INSTALL_LIB_FILES is already present, but it is windows only. On non-windows, it already works as desired; simply omit the `make install` step and the installation of lib files will be skipped.

>  I had to re-run `make` to use them

This is not the case if you avoid `make install`.

andrewrk:
I reconsidered, and I think your way is nice that it's consistent between platforms even though there is still a trick to make it work on non-Windows platforms. Sorry for the noise.

Issue #7201, opened 2020-11-23
builder.standardBuildOptions(.{}) 


daurnimator:
We currently have ` .standardTargetOptions()` which adds a build option to select the architecture, os and ABI. However it doesn't allow the user to e.g. enable/disable cpu features.

With the new pie support, to expose it in build.zig, we're going to want a common 'standard' option for it. e.g. `-Dpie=true`

As we continue, there's going to be more build options that projects should expose by default.

I propose that we wrap up all these commonly exposed build options in a function `.standardBuildOptions(.{})`, which can gain options over time.

andrewrk:
I want to clarify the purpose of this function: it's for communicating to the zig build system a set of preferences or available options. For example, a set of supported targets, or the preferred release mode. It should communicate to zig build system what options are supported, what should be considered to be default, and then the build system takes that and presents the CLI options in a standard way.

Let's call it `standardOptions` since we are already in a "build" context. The preferred release mode should be deprecated and rolled into this as well.

Taking pie as an example, the person writing the build script should have to pass `.supports_pie = true` to `standardOptions` because they will have to remember to then set `.pie = true` on the build artifacts that should be affected by this option.

Or, I would also be OK with having `supports_pie` default to `true` and we could have something like this:

```zig
const opts = b.standardOptions(.{});

// ...

const exe = b.addExecutable("abc", "src/main.zig");
exe.applyOptions(opts);
```

Since `build.zig` is declarative, we can add some handy flags to the build runner for package maintainers to detect if options are supported. For example, we could add a `--list-options` flag that dumps the list of supported options to stdout in a parseable text format.

daurnimator:
> ```zig
> const exe = b.addExecutable("abc", "src/main.zig");
> exe.applyOptions(opts);
> ```

Why not have `.addExecutable` inherit defaults from the builder context?
I might throw up a quick PR doing this for `standardReleaseOptions` now to see what you think

andrewrk:
This was accepted, but after thinking it over some more, I'm changing status to rejected. In general, I want applications to control their options, and explicitly offer configurations to users and package maintainers via build options. Instead of an opaque `standardOptions()` function, the application should explicitly ask for a target, an optimization mode, and so on. If it wants to offer PIE, then it should explicitly offer PIE. Package maintainers should negotiate with upstream application authors on the advantages or disadvantages of PIE, and application authors can oblige or not depending on whether they expose the option.

daurnimator:
[...]

PR #7638, opened 2021-01-01
Have zig build steps automatically inherit mode of builder

daurnimator:
As mentioned in https://github.com/ziglang/zig/issues/7201#issuecomment-753333023


ifreund (commented):
I would make `Builder.standardReleaseOptions()` return void. It seems to me that there will be very few cases where the return value would be useful and for such cases `Builder.release_mode` could be used instead.

daurnimator:
> I would make `Builder.standardReleaseOptions()` return void. It seems to me that there will be very few cases where the return value would be useful and for such cases `Builder.release_mode` could be used instead.

That would be a follow up change (as accepted in #7201): the end goal is to have a single `b.standardOptions(.{})` that doesn't return anything.

ifreund:
> That would be a follow up change (as accepted in #7201): the end goal is to have a single `b.standardOptions(.{})` that doesn't return anything.

Makes sense, that way we only break the API once.

daurnimator:
> Makes sense, that way we only break the API once.

There should be no API break at all: the old functions will be marked deprecated and removed after the next release.

andrewrk:
No good, I want the build script writer to explicitly set the stuff on all build artifacts.

PR #7643, opened 2021-01-01
mark TypeInfo strings as null-terminated

marler8997:
Related Issue: https://github.com/ziglang/zig/pull/7644

I have some code that uses reflection to iterate over a set of function declarations.  It then passes these function names to a function that loads them (like `dlsym` https://linux.die.net/man/3/dlsym).  However, the loader function requires the name string be null-terminated, but it looks like the strings within TypeInfo aren't declared as being null-terminated.

Since string literals are guaranteed to be null-terminated, it is my guess that these strings within `TypeInfo` are also.  I've updated the `TypeInfo` definition to indicate they are, however, I haven't checked yet that the implementation guarantees this.

### DIFF
--- a/lib/std/builtin.zig
+++ b/lib/std/builtin.zig
@@ -269,7 +269,7 @@ pub const TypeInfo = union(enum) {
     /// This data structure is used by the Zig language code generation and
     /// therefore must be kept in sync with the compiler implementation.
     pub const StructField = struct {
-        name: []const u8,
+        name: [:0]const u8,
         field_type: type,
         default_value: anytype,
         is_comptime: bool,
@@ -301,7 +301,7 @@ pub const TypeInfo = union(enum) {
     /// This data structure is used by the Zig language code generation and
     /// therefore must be kept in sync with the compiler implementation.
     pub const Error = struct {
-        name: []const u8,
+        name: [:0]const u8,
     };
 
     /// This data structure is used by the Zig language code generation and
@@ -311,7 +311,7 @@ pub const TypeInfo = union(enum) {
     /// This data structure is used by the Zig language code generation and
     /// therefore must be kept in sync with the compiler implementation.
     pub const EnumField = struct {
-        name: []const u8,
+        name: [:0]const u8,
         value: comptime_int,
     };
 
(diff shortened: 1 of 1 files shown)


LemonBoy:
Apply this patch to make the compilation failure disappear:

```patch
diff --git a/src/stage1/analyze.cpp b/src/stage1/analyze.cpp
index 5f86321e9..8fb87d143 100644
--- a/src/stage1/analyze.cpp
+++ b/src/stage1/analyze.cpp
@@ -10061,8 +10061,12 @@ static void dump_value_indent(ZigValue *val, int indent) {
             } else if (val->type->data.maybe.child_type->id == ZigTypeIdErrorSet) {
                 return dump_value_indent_error_set(val, indent);
             } else {
+                assert(type_has_optional_repr(val->type));
                 fprintf(stderr, "<\n");
-                dump_value_indent(val->data.x_optional, indent + 1);
+                if (val->data.x_optional != nullptr)
+                    dump_value_indent(val->data.x_optional, indent + 1);
+                else
+                    fprintf(stderr, "<null>\n");
 
                 for (int i = 0; i < indent; i += 1) {
                     fprintf(stderr, " ");
diff --git a/src/stage1/ir.cpp b/src/stage1/ir.cpp
index d3ed3cfaa..faf39a214 100644
--- a/src/stage1/ir.cpp
+++ b/src/stage1/ir.cpp
@@ -25264,10 +25264,10 @@ static Error ir_make_type_info_decls(IrAnalyze *ira, IrInst* source_instr, ZigVa
                     // lib_name: ?[]const u8
                     ensure_field_index(fn_decl_val->type, "lib_name", 5);
                     fn_decl_fields[5]->special = ConstValSpecialStatic;
-                    ZigType *u8_ptr = get_pointer_to_type_extra(
-                        ira->codegen, ira->codegen->builtin_types.entry_u8,
-                        true, false, PtrLenUnknown,
-                        0, 0, 0, false);
+                    ZigType *u8_ptr = get_pointer_to_type_extra2(
+                            ira->codegen, ira->codegen->builtin_types.entry_u8,
+                            true, false, PtrLenUnknown,
+                            0, 0, 0, false, VECTOR_INDEX_NONE, nullptr, ira->codegen->intern.for_zero_byte());
                     fn_decl_fields[5]->type = get_optional_type(ira->codegen, get_slice_type(ira->codegen, u8_ptr));
                     if (fn_node->is_extern && fn_node->lib_name != nullptr && buf_len(fn_node->lib_name) > 0) {
                         ZigValue *slice_val = ira->codegen->pass1_arena->create<ZigValue>();
```

marler8997:
Closing since @andrewrk clarified the langauge does not guarantee reflection strings are null-terminated.  For my use-case, I believe I can use a comptime function to make them null-terminated when needed.

Issue #1616, opened 2018-10-01
stack traces don't work when there is a stack overflow


andrewrk:
I just debugged a stack overflow on windows that was solved by increasing the stack size for all executables. Once #157 is solved this can be replaced with a value based on call graph analysis.

But we still have chkstck calling `@panic`. When this happens we try to dump a stack trace, but this uses the stack, which just overflowed. So this can't work. This could potentially be a use case for `@newStackCall`.

mlugg:
https://codeberg.org/ziglang/zig/pulls/30801

PR #7647, opened 2021-01-02
stage2: comptime function calls and inline function calls

andrewrk:
See commit messages for more details.

Follow-up issues to tackle:
 * [x] Adding compile errors to the callsite instead of the callee Decl.
   - [x] add test for this
 * [ ] Proper error notes for "called from here".
   - [ ] Related: #7555
 * [x] Branch quotas for both comptime fn calls and inline function calls.
 * [ ] Complete out the debug TZIR dumping code.
 * [x] Don't redundantly generate ZIR for each inline/comptime function call. Instead we should add a new state enum tag to Fn.
 * [x] add more test cases
 * [ ] caching of comptime function calls (required for generic types to work)


src/zir_sema.zig
```diff
@@ -826,7 +828,112 @@ fn analyzeInstCall(mod: *Module, scope: *Scope, inst: *zir.Inst.Call) InnerError
 
     const ret_type = func.ty.fnReturnType();
 
-    const b = try mod.requireRuntimeBlock(scope, inst.base.src);
+    const b = try mod.requireFunctionBlock(scope, inst.base.src);
+    if (b.is_comptime) {
```
g-w1, line 832: Should you change this to make it `if (b.is_comptime or inst.kw_args.modifier == .compile_time)` ?
Or will this come later with the hashing and determining if something can be called at comptime?

andrewrk: Yep, good call. I'm actually looking into right now what it would look like to support `inline` function calls too, because I think the problem is related.

g-w1:
Also, to note, this segfaults the compiler with *no* stacktrace (pretty sure stack overflow):
```zig
export fn _start() noreturn {
    const v = x;
    unreachable;
}

const x = rec(3);

fn rec(x: u32) u32 {
    return rec(x);
}
```
Recursion support will probably come with the branch eval quota, but it was surprising to me that there was no trace.

Another compelling reason to cache the astgen, besides having something run at comptime and runtime is comptime recursion. If im not mistaken, every time this function is called at comptime, astgen is redundantly run on it.

andrewrk:
stack trace on stack overflow is #1616

yep, I have a bunch of unpushed code that addresses the ZIR caching, almost ready to push!

PR #7660, opened 2021-01-02
std/crypto: properly support arbitrary output sizes

### LINKED ISSUE #7657: blake2b: final 4 bytes not written when expecting 160-bit output
The final 4 bytes appear to be missing when using `Blake2b(160)`.

```zig
const std = @import("std");
const Hash = std.crypto.hash.blake2.Blake2b(160);

pub fn main() void {
    var digest: [Hash.digest_length]u8 = undefined;
    Hash.hash("hello", &digest, .{});
    std.debug.print("{x}", .{digest});
}
```

```
Expected:    b5531c7037f06c9f2947132a6a77202c308e8939
Actual:      b5531c7037f06c9f2947132a6a77202c00000000
```

jedisct1:
Fixes #7657

### DIFF
--- a/lib/std/crypto/blake2.zig
+++ b/lib/std/crypto/blake2.zig
@@ -137,12 +137,8 @@ pub fn Blake2s(comptime out_bits: usize) type {
             mem.set(u8, d.buf[d.buf_len..], 0);
             d.t += d.buf_len;
             d.round(d.buf[0..], true);
-
-            const rr = d.h[0 .. digest_length / 4];
-
-            for (rr) |s, j| {
-                mem.writeIntSliceLittle(u32, out[4 * j ..], s);
-            }
+            for (d.h) |*x| x.* = mem.nativeToLittle(u32, x.*);
+            mem.copy(u8, out[0..], @ptrCast(*[digest_length]u8, &d.h));
         }
 
         fn round(d: *Self, b: *const [64]u8, last: bool) void {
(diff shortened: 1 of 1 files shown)


LemonBoy:
According to RFC7693 the following output lengths are defined:
```

            Algorithm     | Target | Collision | Hash | Hash ASN.1 |
               Identifier |  Arch  |  Security |  nn  | OID Suffix |
           ---------------+--------+-----------+------+------------+
            id-blake2b160 | 64-bit |   2**80   |  20  |   x.1.5    |
            id-blake2b256 | 64-bit |   2**128  |  32  |   x.1.8    |
            id-blake2b384 | 64-bit |   2**192  |  48  |   x.1.12   |
            id-blake2b512 | 64-bit |   2**256  |  64  |   x.1.16   |
           ---------------+--------+-----------+------+------------+
            id-blake2s128 | 32-bit |   2**64   |  16  |   x.2.4    |
            id-blake2s160 | 32-bit |   2**80   |  20  |   x.2.5    |
            id-blake2s224 | 32-bit |   2**112  |  28  |   x.2.7    |
            id-blake2s256 | 32-bit |   2**128  |  32  |   x.2.8    |
           ---------------+--------+-----------+------+------------+
```

And we're not exporting/testing the `nn=160` variant of both the `blake2` hashes.

Shall we add the two missing definitions and a test?

jedisct1:
160 and 224 bits are uncommon settings, but sure, having it doesn't cost much and can be useful, if only to spot bug like these.

I'm going to add tests and types for `nn=160`.

PR #7664, opened 2021-01-03
implement nt path conversion for windows

### LINKED ISSUE #6044: '.' and '..' components in relative paths in std.os.windows.OpenFile

marler8997:
Second attempt to fix #6044  (previous attempt https://github.com/ziglang/zig/pull/7537)

The first attempt leveraged the functionality in `ntdll` to perform the Dos to NT path conversion, however, that API requires heap allocation which @andrewrk has indicated we don't want (see https://github.com/ziglang/zig/pull/7537#discussion_r550807530).

So this PR implements the logic to convert a DOS file path into an NT-compatible path in Zig's std lib.

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -1365,15 +1365,6 @@ pub const Dir = struct {
             .SecurityDescriptor = null,
             .SecurityQualityOfService = null,
         };
-        if (sub_path_w[0] == '.' and sub_path_w[1] == 0) {
-            // Windows does not recognize this, but it does work with empty string.
-            nt_name.Length = 0;
-        }
-        if (sub_path_w[0] == '.' and sub_path_w[1] == '.' and sub_path_w[2] == 0) {
-            // If you're looking to contribute to zig and fix this, see here for an example of how to
-            // implement this: https://git.midipix.org/ntapi/tree/src/fs/ntapi_tt_open_physical_parent_directory.c
-            @panic("TODO opening '..' with a relative directory handle is not yet implemented on Windows");
-        }
         const open_reparse_point: w.DWORD = if (no_follow) w.FILE_OPEN_REPARSE_POINT else 0x0;
         var io: w.IO_STATUS_BLOCK = undefined;
         const rc = w.ntdll.NtCreateFile(
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -2130,7 +2130,7 @@ pub fn replaceScalar(comptime T: type, slice: []T, needle: T, replacement: T) vo
 }
 
 /// Collapse consecutive duplicate elements into one entry.
-pub fn collapseRepeats(comptime T: type, slice: []T, elem: T) usize {
+pub fn collapseRepeatsLen(comptime T: type, slice: []T, elem: T) usize {
     if (slice.len == 0) return 0;
     var write_idx: usize = 1;
     var read_idx: usize = 1;
@@ -2143,11 +2143,15 @@ pub fn collapseRepeats(comptime T: type, slice: []T, elem: T) usize {
     return write_idx;
 }
 
+/// Collapse consecutive duplicate elements into one entry.
+pub fn collapseRepeats(comptime T: type, slice: []T, elem: T) []T {
+    return slice[0 .. collapseRepeatsLen(T, slice, elem)];
+}
+
 fn testCollapseRepeats(str: []const u8, elem: u8, expected: []const u8) !void {
     const mutable = try std.testing.allocator.dupe(u8, str);
     defer std.testing.allocator.free(mutable);
-    const actual = mutable[0..collapseRepeats(u8, mutable, elem)];
-    testing.expect(std.mem.eql(u8, actual, expected));
+    testing.expect(std.mem.eql(u8, collapseRepeats(u8, mutable, elem), expected));
 }
 test "collapseRepeats" {
     try testCollapseRepeats("", '/', "");
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -2151,7 +2151,7 @@ pub fn collapseRepeats(comptime T: type, slice: []T, elem: T) []T {
 fn testCollapseRepeats(str: []const u8, elem: u8, expected: []const u8) !void {
     const mutable = try std.testing.allocator.dupe(u8, str);
     defer std.testing.allocator.free(mutable);
-    testing.expect(std.mem.eql(u8, collapseRepeats(u8, mutable, elem), expected));
+    try testing.expect(std.mem.eql(u8, collapseRepeats(u8, mutable, elem), expected));
 }
 test "collapseRepeats" {
     try testCollapseRepeats("", '/', "");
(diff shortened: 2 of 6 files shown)


lib/std/os/windows.zig
```diff
@@ -1528,6 +1528,81 @@ pub const PathSpace = struct {
     }
 };
 
+/// The error type for `removeDotDirsSanitized`
+pub const RemoveDotDirsError = error { TooManyParentDirs };
+
+/// Removes '.' and '..' path components from a "sanitized relative path".
```
daurnimator, line 1534: this function looks like it belongs in std.fs.path. it might be a duple of `resolveWindows`?

marler8997: It's specific to windows...only works with backslashes.  We could also make it work with forward slashes but I think posix has other better way to perform this operation.  This function doesn't touch the filesystem or use `CWD` to normalize the path like `resolveWindows` does.

daurnimator: > It's specific to windows...only works with backslashes

Os-specific code is all over the standard library: it does all just get dumped into os/windows.zig.

lib/std/mem.zig
```diff
@@ -1829,6 +1829,49 @@ test "replace" {
     testing.expect(eql(u8, output[0..], "Favor reading  over writing ."));
 }
 
+/// Replace all occurences of `needle` with `replacement`.
+pub fn replaceElem(comptime T: type, slice: []T, needle: T, replacement: T) void {
```
LemonBoy, line 1833: If we follow the `std.mem` naming this should be `replaceScalar` and take an `output: []T` parameter (slice becomes `const`).
Specifying the same in/out buffers gets you the in-place replacement you want.

lib/std/mem.zig
```diff
@@ -1829,6 +1829,49 @@ test "replace" {
...
+    for (slice) |e, i| {
+        if (e == needle) {
+            slice[i] = replacement;
+        }
+    }
+}
+
+/// Collapse consecutive duplicate elements into one entry.
+pub fn collapseRepeats(comptime T: type, slice: []T, needle: T) usize {
```
LemonBoy, line 1842: Not much of a needle, it's more a `elem`ent no?
And given it squishes `slice` contents it makes sense for this to return a slice of it.

lib/std/mem.zig
```diff
@@ -1829,6 +1829,49 @@ test "replace" {
...
+    if (slice.len == 0) return 0;
+    var write_idx : usize = 1;
```
LemonBoy, line 1844: Smells like unformatted code...

marler8997: yeah, should be formatted now

lib/std/mem.zig
marler8997: fixed

lib/std/mem.zig
marler8997: I've changed `needle` to `elem`.

As for returning a slice, such an interface could be supported with a wrapper function at no runtime cost. In the current uses, only the resulting size is used and this same interface is propagated in a handful of places which makes the code simpler.

LemonBoy: You can get the old behaviour by simply adding a `.len` at the end.
Returning a slice makes the API much easier to use as `f(collapseRepeats(..))`, throwing indices around is error-prone.

marler8997: > Returning a slice makes the API much easier to use

In some cases yes.  If such a use case comes up we can add a wrapper function that does this.

LemonBoy: It should be the other way around, returning a slice returns a pointer _and_ the adjusted length.

marler8997: > It should be the other way around, returning a slice returns a pointer and the adjusted length.

My philosphy is to do less, and then only do more if you need to.  Going the other way around means always doing more even if you don't need to, and then undoing what you don't need.

marler8997: Another example of an API that uses this methodology is the Allocator `resizeFn` function.

LemonBoy: If you're adding new public-facing APIs you can't just think in terms of "do less" or think of your single use-case.
Returning the length makes using this function with sliced inputs hard to use, consider the following example:
```zig
const x = collapseRepeats(foo[bar..Baz], ...)
```
If you return a sub-slice the user is able to quickly use the modified chunk of input, while if the length is returned instead another slicing operation is required.

But given that opinions are like assholes I'll rest my case, I'm getting too old for this shit.

marler8997: Ok the fact that this is a public facing function is a good point.  How about creating a `collapseRepeatsLen` and `collapseRepeats` that wraps it?

lib/std/fs.zig
```diff
@@ -1285,15 +1285,6 @@ pub const Dir = struct {
             .SecurityDescriptor = null,
             .SecurityQualityOfService = null,
         };
-        if (sub_path_w[0] == '.' and sub_path_w[1] == 0) {
-            // Windows does not recognize this, but it does work with empty string.
-            nt_name.Length = 0;
-        }
-        if (sub_path_w[0] == '.' and sub_path_w[1] == '.' and sub_path_w[2] == 0) {
-            // If you're looking to contribute to zig and fix this, see here for an example of how to
-            // implement this: https://git.midipix.org/ntapi/tree/src/fs/ntapi_tt_open_physical_parent_directory.c
```
andrewrk, line 1294: I don't see this addressed in the PR description? Did you look into using this technique to resolve `..` in a way that respects following symlinks?

andrewrk (changes requested):
Thanks for the patch. This solution involves manipulating path strings, however Windows has symlinks, and on other platforms, when opening files with `std.fs`, ".." path components respect symlinks, so we need to do it here, too.

lib/std/fs.zig
marler8997: On windows symlinks are resolved after resolving ".." (https://github.com/ziglang/zig/issues/7751)

squeek502:
Nice work @marler8997! This hugely improves `std.fs` on Windows.

andrewrk:
Thanks for the clean-up PRs @squeek502!

PR #7672, opened 2021-01-03
std/crypto: add BLAKE2-160 types and tests

jedisct1:


lib/std/crypto/blake2.zig
```diff
@@ -476,7 +561,7 @@ pub fn Blake2b(comptime out_bits: usize) type {
             mem.set(u8, d.buf[d.buf_len..], 0);
             d.t += d.buf_len;
             d.round(d.buf[0..], true);
-            for (d.h) |*x| x.* = mem.nativeToLittle(u64, x.*);
+            for (d.h) |*x| x.* = mem.littleToNative(u64, x.*);
```
LemonBoy, line 564: The RFC states the output is `RETURN first "nn" bytes from little-endian word array h[].`.
Is `h` supposed to be in LE order or the native one?

jedisct1: It doesn't make a difference as long as we swap. But you're right, `h` appears to be in native order, so `nativeToLittle` makes more sense. Reverted, thanks!

Issue #7675, opened 2021-01-03
Assert compile error for unspecified array/slice format


marler8997:
I'm opening an issue based on my comment here: https://github.com/ziglang/zig/pull/6870#issuecomment-735450839

That PR was just merged which means that now the default format for printing slices/arrays will be integer values rather than raw values, i.e.:

```zig
const std = @import("std");
pub fn main() void {
    std.debug.warn("Hello, {}\n", .{"World"});
}
```

```bash
$ zig run hello.zig
Hello, [5:0]u8@7ff7cb1be3a0
```

The fix for this example is to specify `{s}` whenever you're printing a string.  The problem with this is now whenever a format string does not specify the correct format, the bug will not be caught until runtime and the string is inspected.  What makes this worse is that this kind of format string is commonly used for error messages, which often don't get tested or inspected until the error occurs, which many times is only seen rarely if at all.  There is now a disconnect between what code looks like, and the default way to format strings.

To remedy this, I propose that we require a format specifier for slices/arrays. On top of that, in order to support generic code I propose that we enable today's current behavior of `{}` with an explicit specifier.  Something like `{any}`.  The `{any}` specifier would mean "print this thing, I'm not picky about the exact format, just print it somehow".  The difference here is that instead having this "I'm not picky about the format" attitude by default with `{}`, the user has to opt-in to it.

Also note that with the PR that was just merged, alot of Zig code will now be broken due to missing format specifiers that were meant to be `{s}`.  Implementing this proposal means these bugs will be found at compile-time so they will be easy to find and fix.

Here's the list of array specifiers:
```
{s} - format as a list of Unicode characters characters and interpret array as a utf8 encoded string
{d} - format as a list of decimal values
{x} - format as a list of lowercase hex values
{X} - format as a list of uppercase hex values
```

daurnimator:
As an alternative solution, what if there was a `String` type in zig that made it distinct from an array of u8?

marler8997:
Yeah having distinct string type would be another way to solve this.

rohlem:
Unless I'm missing something, even if we had a distinct `String` type, we still want `[]u8` (any kind of "normal" C string that people forgot to cast) to trigger an `assert` with the default `{}` format, right?
Because that seems the essence of this footgun, the fact that the "simplest"/"most obvious" solution gives undesirable output in rarely-tested code paths. Adding "a right way" via an optional `String` type alone doesn't quite solve this I think.

marler8997:
[...]

PR #7676, opened 2021-01-03
require specifier for arrayish types

marler8997:
Implement https://github.com/ziglang/zig/issues/7675

Justification for this change requires going no further than this repo as you can see it has already caught a smattering of bugs that were missed when we changed the default format behavior of `u8` arrays.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -739,7 +739,7 @@ pub const Builder = struct {
                 return args.default_target;
             },
             else => |e| {
-                warn("Unable to parse target '{}': {s}\n\n", .{ triple, @errorName(e) });
+                warn("Unable to parse target '{s}': {s}\n\n", .{ triple, @errorName(e) });
                 self.markInvalidUserInput();
                 return args.default_target;
             },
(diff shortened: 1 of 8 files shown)


dbandstra:
Can you update the doc comment at the top of the file to mention `{p}` and the new role for `{d}`

My other comment is maybe change "i.e." to "e.g." in your compile error messages since they are not listing all the valid choices (there's also z, x, etc)

LemonBoy:
No need to introduce `{p}`, that's what `{*}` is for.

LemonBoy:
Requiring a type makes generic code hard to get right, `{}` right now serves as a catch-all to get a textual representation.

marler8997:
> Requiring a type makes generic code hard to get right, {} right now serves as a catch-all to get a textual representation.

Yeah I'm seeing that now.  We could leave as is, or, we could introduce a new specifier to enable the current behavior?

```
{G}   // generic specifier, meaning, the specifier will pick some generic default and it will always work
{D}   // same thing, stands for "Default"
```

I'm looking at this from the perspective of "what's the sanest default"?  As I see it, by default most code will not want the behavior of today's `{}`.   By default people will want to know if they forgot to include `{s}` at compile-time, and by forcing them to put something like `{G}` or `{D}`, they are explicitly opting out of this safety check, rather than using `{}` and opting out of it by default.  What do you think?

andrewrk:
Thanks, really appreciate this.

PR #6969, opened 2020-11-04
stage2: variable shadowing detection

g-w1:
This allows detection of variable shadowing in stage2, fixing a TODO in `varDecl` function. I added a `compileError` test to test/stage2/test.zig, but it didn't seem to run.

Edit:
The test actually did run and pass, but it said that it ran the wrong number of tests?
```
[jacob@archlinux ~co/zig/zig]$ zig build test-stage2
Test [1/17] test "self-hosted"... tests [1/28] referencing decls which appear later in the file [1/1] updAll 17 tests passed.
[jacob@archlinux ~co/zig/zig]$
```
Why does it say 1/17 and 1/28? and then only 17 tests pass? I think this is a bug in the testing code.

PR #7682, opened 2021-01-04
stage2: implimentation of @setEvalBranchQuota

g-w1:
@setEvalBranchQuota can be run out of comptime function contexts so it must be in any block that that has a sub-branch that can be run at comptime.
For example: 
```zig
@setEvalBranchQuota(100);
comptime {
while (true) {}
}
```
requires it, but it can be in a non-comptime function call.

A block will inherit it's parent's branch count/quota.


g-w1:
Also, the documentation specifies the size of a branch eval quota as u32, but it was u64. Is there a reason for that?

andrewrk:
The fields were correctly in the Shared struct so that sub-blocks counted toward the branch total. With this change, infinite recursion is no longer detected.

I think you're right about u32 though

g-w1:
Ah, so we just need a shared for all block insts.

src/Module.zig
```diff
@@ -782,18 +789,16 @@ pub const Scope = struct {
...
+            shared: *IShared,
             /// We use this to count from 0 so that arg instructions know
             /// which parameter index they are, without having to store
             /// a parameter index with each arg instruction.
             param_index: usize,
             casted_args: []*Inst,
             merges: Merges,
 
-            pub const Shared = struct {
+            pub const IShared = struct {
```
andrewrk, line 800: What's the `I` stand for? Presumably "Inlining"? That would be redundant. The full name is changing from

`Module.Scope.Block.Inlining.Shared`

to

`Module.Scope.Block.Inlining.IShared`

src/Module.zig
```diff
@@ -766,6 +766,13 @@ pub const Scope = struct {
         inlining: ?*Inlining,
         is_comptime: bool,
 
+        shared: *Shared,
+
+        const Shared = struct {
+            branch_count: u32 = 0,
```
andrewrk, line 772: This one should be in the Inlining.Shared struct however branch_quota indeed should be shared among the analysis for an entire Decl.

src/Module.zig
```diff
@@ -1104,6 +1109,10 @@ fn astGenAndAnalyzeDecl(self: *Module, decl: *Decl) !bool {
             var inst_table = Scope.Block.InstTable.init(self.gpa);
             defer inst_table.deinit();
 
+            var shared = try self.gpa.create(Scope.Block.Shared);
+            defer self.gpa.destroy(shared);
```
andrewrk, line 1113: if this were heap allocated it should be `const` instead of `var` but this should be stack allocated instead. See directly above we stack allocate inst_table and then pass a pointer to it into Scope.Block

src/astgen.zig
```diff
@@ -2317,6 +2317,15 @@ fn compileError(mod: *Module, scope: *Scope, call: *ast.Node.BuiltinCall) InnerE
     return addZIRUnOp(mod, scope, src, .compileerror, target);
 }
 
+fn setEvalBranchQuota(mod: *Module, scope: *Scope, call: *ast.Node.BuiltinCall) InnerError!*zir.Inst {
+    try ensureBuiltinParamCount(mod, scope, call, 1);
+    const tree = scope.tree();
+    const src = tree.token_locs[call.builtin_token].start;
+    const params = call.params();
+    const target = try expr(mod, scope, .none, params[0]);
```
andrewrk, line 2325: Missing a type coercion to u32

src/zir_sema.zig
```diff
@@ -486,6 +487,17 @@ fn analyzeInstStoreToInferredPtr(
...
 }
 
+fn analyzeInstSetEvalBranchQuota(
+    mod: *Module,
+    scope: *Scope,
+    inst: *zir.Inst.UnOp,
+) InnerError!*Inst {
+    const b = try mod.requireFunctionBlock(scope, inst.base.src);
+    const quota = @truncate(u32, try resolveInt(mod, scope, inst.positionals.operand, Type.initTag(.u32)));
+    b.shared.branch_quota = quota;
```
andrewrk, line 497: https://ziglang.org/documentation/master/#setEvalBranchQuota

>  If the new_quota is smaller than the default quota (1000) or a previously explicitly set quota, it is ignored.

src/zir_sema.zig
```diff
@@ -865,9 +880,7 @@ fn analyzeInstCall(mod: *Module, scope: *Scope, inst: *zir.Inst.Call) InnerError
         };
         // If this is the top of the inline/comptime call stack, we use this data.
         // Otherwise we pass on the shared data from the parent scope.
-        var shared_inlining = Scope.Block.Inlining.Shared{
-            .branch_count = 0,
-            .branch_quota = 1000,
+        var shared_inlining = Scope.Block.Inlining.IShared{
```
andrewrk, line 883: This is indeed where branch_count should be initialized to 0.

test/stage2/test.zig
```diff
@@ -1477,4 +1477,58 @@ pub fn addCases(ctx: *TestContext) !void {
             \\}
         , &[_][]const u8{":8:10: error: evaluation exceeded 1000 backwards branches"});
     }
+    {
+        var case = ctx.exe("@setEvalBranchQuota", linux_x64);
+        case.addCompareOutput(
+            \\export fn _start() noreturn {
+            \\    @setEvalBranchQuota(100000);
```
andrewrk, line 1484: Let's avoid wasting time since this test is going to be run countless times. We should have exactly 1 test setting the quota to exactly 1001 branches, and the rest should be using much lower values.

andrewrk (changes requested):

src/astgen.zig
g-w1: This is coerced when you do 
```zig
const quota = @truncate(u32, try resolveInt(mod, scope, inst.positionals.operand, Type.initTag(.u32)));
```

src/Module.zig
g-w1: I thought the scoping would not allow it. And if it does I was being cautious for a future when it will not. See #6969

src/astgen.zig
andrewrk: It's better when the result location does the coercion because then something like this will work:

```zig
@setEvalBranchQuota(if (cond) 1 else 2);
```

Without the result location doing the coercion, this will give you an error.

See 

https://github.com/ziglang/zig/blob/f6644255f595fb0bbfb2fa4276804ef7031330b1/src/astgen.zig#L24-L25

and its uses

src/Module.zig
g-w1: Ah gotcha, It looked to me that the above one was heap allocated because it was using gpa.

src/Module.zig
g-w1: This makes it not able to be used outside an inline function.

PR #7695, opened 2021-01-05
std.c add syslog

xackus:
I guess in theory the priority constants could be different on some weird systems.
These are standardized in RFC 5424, but that's the network protocol.
Let me know what you think about placing them in `std/os/bits.zig`

### DIFF
--- a/lib/std/os/bits.zig
+++ b/lib/std/os/bits.zig
@@ -34,3 +34,22 @@ pub const iovec_const = extern struct {
     iov_base: [*]const u8,
     iov_len: usize,
 };
+
+// syslog
+
+/// system is unusable
+pub const LOG_EMERG = 0;
+/// action must be taken immediately
+pub const LOG_ALERT = 1;
+/// critical conditions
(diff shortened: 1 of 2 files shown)


lib/std/os/bits.zig
```diff
@@ -34,3 +34,22 @@ pub const iovec_const = extern struct {
     iov_base: [*]const u8,
     iov_len: usize,
 };
+
+// syslog
+
+/// system is unusable
+pub const LOG_EMERG = 0;
```
daurnimator, line 41: These should be an enum.
`std.log.Level` might be reusable.

andrewrk: I don't want to couple `std.log.Level` with this.

daurnimator: The `std.log.Level` options intentionally match the syslog protocol.

xackus:
```c
syslog(LOG_ERR|LOG_USER, "error");
```

PR #7705, opened 2021-01-06
spinlock: Default SpinLock.state to .Unlocked to allow default struct initialization

BinaryWarlock:
`std.Mutex`, which is default struct initialized in some places (e.g. `GeneralPurposeAllocator`), is possibly defined to be `std.SpinLock` as a default (e.g. on the freestanding target), which would then raise a compile error when trying to use `std.Mutex` or one of its users (e.g. `GeneralPurposeAllocator`).

This is consistent with the other types of mutexes.


andrewrk:
Thanks!

PR #7710, opened 2021-01-06
SendError should include SocketNetConnected error

marler8997:


MasterQ32:
No it should not. See #7683  and the doc comments. Using `send` on a unconnected socket is a programming mistake and not a runtime error

PR #7712, opened 2021-01-06
remove SocketNotConnected

marler8997:
I'm not 100% sure this is correct yet. The question is whether or not ENOTCONN (or WSAENOTCONN) is always the result of a codebug, or if it can also be valid runtime error in some cases.

I suppose one way to find out is to remove the error code like I've done here, and if anyone ever sees this error in a valid use case then we can add it back in?

cc @MasterQ32


andrewrk:
Let's be sure it's correct before inserting `=> unreachable`. Until then, I think status quo is reasonable.

PR #6159, opened 2020-08-24
std.mem.count

Sobeston:
Surprised that I couldn't find a function for this

PR #7713, opened 2021-01-06
added mem.containsAtLeast

Sobeston:


daurnimator (approved):
Seems reasonable; what is your use-case?

Sobeston:
> Seems reasonable; what is your use-case?

https://github.com/ziglang/zig/pull/6159#issuecomment-679898158 This is something I've caught myself re-implementing a few times, though I can't think of anything specific.

marler8997:
Seems like the first `if (expected_count == 0) return true;` is unnecessary, what about this?
```zig
pub fn containsAtLeast(comptime T: type, haystack: []const T, expected_count: usize, needle: []const T) bool {
    assert(needle.len > 0);
    var i: usize = 0;
    var found: usize = 0;
    while (found < expected_count) {
        if (indexOfPos(T, haystack, i, needle)) |idx| {
            i = idx + needle.len;
            found += 1;
        } else return false;
    }
    return true;
}
```

Sobeston:
I mean if that also works the exact same (which it looks like it does at a glance), then that's fine too. I don't mind the style either way.

PR #7714, opened 2021-01-07
macos: reimplement OS version detection

### LINKED ISSUE #7569: macOS Big Sur real versioning hidden from pre-Big Sur binaries
macOS (formerly branded Mac OS X) has been versioned fairly consistently since the initial `10.0` release about 20 years ago. Since then the major release number `10` has been used up until the release of Big Sur which bumped major to `11`.

Apple no doubt recognized that many existing applications need to detect macOS version at configure/build/runtime and decided to put in place a backwards compatible versioning scheme that is activated by two different means:

1. executables built against 10.15 SDK and earlier
2. executables launched with environ `SYSTEM_VERSION_COMPAT=1`

If either of these is true then the OS/kernel make it very difficult to obtain real versioning for Big Sur. In other words, Apple made it super easy for backwards version compatibility for ignorant executables, but made it nigh impossible for knowledgeable executables to detect Big Sur versioning.

But how and why is this important to Zig? Here's the problem statement:

**Zig posts release and CI binaries that are built on pre-Big Sur hosts and out of the box these binaries will not work on Big Sur as expected. Some of the issues have been worked around, but ultimately, host and (default) target information will be incorrect because the host will be detected as `10.16`.**

#### use sysctl MIB `kern.osproductversion`
Returns `10.16` inside a backwards compatibility context.

#### use sysctl MIB `kern.osversion` instead
We have a "temporary" fallback measure in place to convert macOS build strings to product versions and up to and including `10.15` it was formulaic and robust. By formulaic I mean that it would work with future versions. Unfortunately Big Sur completely breaks any formulaic pattern and build strings essentially are *monotonically increasing*. For example, we cannot predict when the first "letter" in the string will indicate a new minor or patch or both minor+patch change. As of this writing here are Big Sur known builds:

|Version|Build|
|-|-|
|11.0|20A2411|
|11.0.1|20B29|
|11.0.1|20B50|
|11.1|20C69|
|11.2 beta|20D5029f|

#### detect sysctl MIB `kern.osproductversioncompat`
It is possible to detect a **new** MIB that indicates current process is in backwards compatibility context. While this might be useful to emit a warning to find a better binary or some such, that's about as far as it goes. We still cannot determine Big Sur real version.

#### setenv `SYSTEM_VERSION_COMPAT=0` in zig compiler main
Does not work in the current process; it must be set in parent process of the executable.

#### have the user setenv `SYSTEM_VERSION_COMPAT=0` in their shell
This reeks. We'd like Zig binaries to just work out-of-the-box as much as possible.

#### read/parse file `/System/Library/CoreServices/SystemVersion.plist`
This was intended to be the go-to long term solution (see #5125) and is no longer viable. The file contents indicate `10.16` when in a backwards compatibility context.

#### parse output from `sw_vers`
Returns `10.16` inside a backwards compatibility context.

## Solution: a dotfile symlink
I did a deep search on a `11.1` system and found the following `.plist` files:

```
▸ ls -al /System/Library/CoreServices | grep ' SystemVersion.*plist'
lrwxr-xr-x   1 root wheel      19 Jan  1  2020 .SystemVersionPlatform.plist -> SystemVersion.plist
-r--r--r--   1 root wheel     524 Jan  1  2020 SystemVersion.plist
-r--r--r--   1 root wheel     529 Jan  1  2020 SystemVersionCompat.plist
```

It turns out that even though the symlink points to `SystemVersion.plist` reading/parsing file by that symlink path bypasses backwards compatibility context and always returns Big Sur real version info.

Thus our logic can have a single branch with something as straightforward as:

```
// `10.16` is only returned when in backwards compatible context
if sysctl MIB `kern.osproductversion` == `10.16` {
    parse_via_dotfile_simlink_path
}
```

mikdusan:
The macOS version is now obtained by parsing `SystemVersion.plist`.

Test cases added for plist files that date back to '2003 Panther and up
to the recent '2020 Big Sur 11.1 release of macOS.

Thus we are now able to reliably identify `10.3`...`11.1` and higher.

- drop use of kern.osproductversion sysctl
- drop use of kern.osversion sysctl (fallback)
- drop kern.osversion tests
- add `lib.std.zig.system.detect()`
- add minimalistic parser for `SystemVersion.plist`
- add test cases for { 10.3.0, 10.3.9, 10.15.6, 11.0, 11.1 }

closes #7569

### DIFF
--- a/lib/std/target.zig
+++ b/lib/std/target.zig
@@ -260,7 +260,7 @@ pub const Target = struct {
                     .macos => return .{
                         .semver = .{
                             .min = .{ .major = 10, .minor = 13 },
-                            .max = .{ .major = 10, .minor = 15, .patch = 7 },
+                            .max = .{ .major = 11, .minor = 1 },
                         },
                     },
                     .ios => return .{
--- a/lib/std/zig/system.zig
+++ b/lib/std/zig/system.zig
@@ -254,38 +254,7 @@ pub const NativeTargetInfo = struct {
                     os.version_range.windows.min = detected_version;
                     os.version_range.windows.max = detected_version;
                 },
-                .macos => {
-                    var scbuf: [32]u8 = undefined;
-                    var size: usize = undefined;
-
-                    // The osproductversion sysctl was introduced first with 10.13.4 High Sierra.
-                    const key_osproductversion = "kern.osproductversion"; // eg. "10.15.4"
-                    size = scbuf.len;
-                    if (std.os.sysctlbynameZ(key_osproductversion, &scbuf, &size, null, 0)) |_| {
-                        const string_version = scbuf[0 .. size - 1];
-                        if (std.builtin.Version.parse(string_version)) |ver| {
-                            os.version_range.semver.min = ver;
-                            os.version_range.semver.max = ver;
-                        } else |err| switch (err) {
-                            error.Overflow => {},
-                            error.InvalidCharacter => {},
-                            error.InvalidVersion => {},
-                        }
-                    } else |err| switch (err) {
-                        error.UnknownName => {
-                            const key_osversion = "kern.osversion"; // eg. "19E287"
-                            size = scbuf.len;
-                            std.os.sysctlbynameZ(key_osversion, &scbuf, &size, null, 0) catch {
-                                @panic("unable to detect macOS version: " ++ key_osversion);
-                            };
-                            if (macos.version_from_build(scbuf[0 .. size - 1])) |ver| {
-                                os.version_range.semver.min = ver;
-                                os.version_range.semver.max = ver;
-                            } else |_| {}
-                        },
-                        else => @panic("unable to detect macOS version: " ++ key_osproductversion),
-                    }
-                },
+                .macos => macos.detect(&os) catch {}, // valid to ignore any error and keep os defaults
                 .freebsd => {
                     var osreldate: u32 = undefined;
                     var len: usize = undefined;
--- a/lib/std/zig/system.zig
+++ b/lib/std/zig/system.zig
@@ -222,6 +222,7 @@ pub const NativeTargetInfo = struct {
         ProcessFdQuotaExceeded,
         SystemFdQuotaExceeded,
         DeviceBusy,
+        OSVersionDetectionFail,
     };
 
     /// Given a `CrossTarget`, which specifies in detail which parts of the target should be detected
@@ -254,7 +255,7 @@ pub const NativeTargetInfo = struct {
                     os.version_range.windows.min = detected_version;
                     os.version_range.windows.max = detected_version;
                 },
-                .macos => macos.detect(&os) catch {}, // valid to ignore any error and keep os defaults
+                .macos => try macos.detect(&os),
                 .freebsd => {
                     var osreldate: u32 = undefined;
                     var len: usize = undefined;
--- a/lib/std/zig/system/macos.zig
+++ b/lib/std/zig/system/macos.zig
@@ -6,453 +6,408 @@
 const std = @import("std");
 const assert = std.debug.assert;
 const mem = std.mem;
+const testing = std.testing;
 
-pub fn version_from_build(build: []const u8) !std.builtin.Version {
-    // build format:
-    //   19E287 (example)
-    //   xxyzzz
+/// Detect macOS version.
+/// On error `os` will not be modified.
+pub fn detect(os: *std.Target.Os) !void {
+    // Drop use of osproductversion sysctl because:
+    //   1. only available 10.13.4 High Sierra and later
+    //   2. when used from a binary built against < SDK 11.0 it returns 10.16 and masks Big Sur 11.x version
     //
-    // major = 10
-    // minor = x - 4 = 19 - 4 = 15
-    // patch = ascii(y) - 'A' = 'E' - 'A' = 69 - 65 = 4
-    // answer: 10.15.4
+    // NEW APPROACH, STEP 1, parse file:
     //
-    // note: patch is typical but with some older releases (before 10.8) zzz is considered
+    //   /System/Library/CoreServices/SystemVersion.plist
+    //
+    // NOTE: Historically `SystemVersion.plist` first appeared circa '2003
+    // with the release of Mac OS X 10.3.0 Panther.
+    //
+    // and if it contains a `10.16` value where the `16` is `>= 16` then it is a red herring
+    // and is discarded, and move on to next step. Otherwise we accept this is the
+    // canonical file for versioning.
+    //
+    // BACKGROUND: `10.(16+)` is not a proper version and does not have enough fidelity to
+    // indicate minor/point version of Big Sur and later. It is a context-sensitive result
+    // issued by the kernel for backwards compatibility purposes. Likely the kernel checks
+    // if the executable was linked against an SDK older than Big Sur.
+    //
+    // STEP 2, parse next file:
+    //
+    //   /System/Library/CoreServices/.SystemVersionPlatform.plist
+    //
+    // NOTE: Historically `SystemVersionPlatform.plist` first appeared circa '2020
+    // with the release of macOS 11.0 Big Sur.
+    //
+    // Accessing the content via this path circumvents a context-sensitive result and
+    // yields a canonical Big Sur version.
+    //
+    // At this time there is no other known way for a < SDK 11.0 executable to obtain a
+    // canonical Big Sur version.
+    //
+    // This implementation uses a reasonably simplified approach to parse .plist file
+    // that while it is an xml document, we have good history on the file and its format
+    // such that I am comfortable with implementing a minimalistic parser.
+    // Things like string and general escapes are not supported.
+    const prefixSlash = "/System/Library/CoreServices/";
+    const format_failure = "macOS detect: failed to {s} '{s}': {}";
 
-    // never return anything below 10.0.0
-    var result = std.builtin.Version{ .major = 10, .minor = 0, .patch = 0 };
+    const paths = [_][]const u8{
+        prefixSlash ++ "SystemVersion.plist",
+        prefixSlash ++ ".SystemVersionPlatform.plist",
+    };
+    for (paths) |path| {
+        // approx. 4 times historical file size
+        var buf: [2048]u8 = undefined;
 
-    // parse format-x
-    var yindex: usize = 0;
-    {
-        while (yindex < build.len) : (yindex += 1) {
-            if (build[yindex] < '0' or build[yindex] > '9') break;
+        if (std.fs.cwd().readFile(path, &buf)) |bytes| {
+            if (parseSystemVersion(bytes)) |ver| {
+                // never return red herring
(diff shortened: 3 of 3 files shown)


lib/std/zig/system/macos.zig
```diff
@@ -7,452 +7,382 @@ const std = @import("std");
...
+    //   1. only available 10.13.4 High Sierra and later
+    //   2. when used from a binary built against < SDK 11.0 it returns 10.16 and masks Big Sur 11.x version
     //
-    // major = 10
-    // minor = x - 4 = 19 - 4 = 15
-    // patch = ascii(y) - 'A' = 'E' - 'A' = 69 - 65 = 4
-    // answer: 10.15.4
+    // New method is to parse and use the first valid:
+    //   /System/Library/CoreServices/.SystemVersionPlatform.plist
+    //   /System/Library/CoreServices/SystemVersion.plist
```
kubkon, line 19: These are available regardless of the fact you have Xcode or CLT installed, right?

kubkon:
Quickly skimming over this, looks great! I'll reserve some time tomorrow to go over this properly, and come back with feedback.

lib/std/zig/system/macos.zig
mikdusan: Correct, comes with vanilla macOS, no xcode or command line tools and that is absolute file name, not offset from SDK or anything. An interesting tidbit is that the xnu-kernel itself, at boot/load learns the value for what it thinks os version is from that primary file.

`. SystemVersionPlatform.plist` is **new** and may be a Big Sur and higher thing. I don't see it on 10.13.6 High Sierra system for example. I'm confident they put that in as the only way to skirt context-sensitive version presentation.

lib/std/target.zig
```diff
@@ -237,7 +237,7 @@ pub const Target = struct {
                     .macos => return .{
                         .semver = .{
                             .min = .{ .major = 10, .minor = 13 },
-                            .max = .{ .major = 10, .minor = 15, .patch = 7 },
+                            .max = .{ .major = 11, .minor = 1 },
```
Mouvedia, line 240: FYI there's a [11.2 β](https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_2-beta-release-notes).

mikdusan: So far we've just been updating the `.max` as releases come out and a dev has access to a system to very the version. I think that approach is fine for now.

Mouvedia (changes requested):

lib/std/zig/system/macos.zig
```diff
@@ -7,452 +7,382 @@ const std = @import("std");
...
+    // Quick background. `SystemVersion.plist` is context-sensitive to the reading process
+    // SDK marker. If < SDK 11.0 it behaves like sysctl and returns 10.16, masking a bonafide
+    // Big Sur version. Work around by opening `.SystemVersionPlatform.plist` first and its
+    // contents are not context-sensitive, yielding a bonafide Big Sur version.
+    //
+    // At this time there is no other known way for a < SDK 11.0 executable to obtain a
+    // bonifide Big Sur version.
+    //
+    // Historically the pathname `/System/Library/CoreServices/SystemVersion.plist` appeared
+    // around '2005 with Mac OS X 10.3.9 .
```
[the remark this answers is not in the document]
mikdusan: I have been checking the SDK redundant copy of `CoreServices/ SystemVersion.plist` and according to https://github.com/phracker/MacOSX-SDKs/tree/master/MacOSX10.3.0.sdk/System/Library
there isn't a `CoreServices/` in SDK. But if you can find a legit copy of 10.3.0's system installed `.plist` I'll add the test case. But no problems if you can't find it, 15 years back is plenty :)

lib/std/target.zig
```diff
@@ -237,7 +237,7 @@ pub const Target = struct {
                     .macos => return .{
                         .semver = .{
                             .min = .{ .major = 10, .minor = 13 },
```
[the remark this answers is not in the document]
mikdusan: I believe that is intentional to indicate officially supported macos version low-watermark.

lib/std/zig/system/macos.zig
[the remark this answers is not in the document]
mikdusan: found an install image and added test case for 10.3 .plist

lib/std/zig/system.zig
```diff
@@ -287,38 +287,7 @@ pub const NativeTargetInfo = struct {
...
-                            };
-                            if (macos.version_from_build(scbuf[0 .. size - 1])) |ver| {
-                                os.version_range.semver.min = ver;
-                                os.version_range.semver.max = ver;
-                            } else |_| {}
-                        },
-                        else => @panic("unable to detect macOS version: " ++ key_osproductversion),
-                    }
-                },
+                .macos => macos.detect(&os) catch {}, // valid to ignore any error and keep os defaults
```
daurnimator, line 290: Doesn't this mean that if the user is running with less permissions, then they silently get different behaviour?

Shouldn't we at least `log.warn` that we had an error and are falling back?

lib/std/zig/system/macos.zig
```diff
@@ -7,452 +7,382 @@ const std = @import("std");
...
-        .{ "18F132", "10.14.5" },
-        .{ "18G84", "10.14.6" },
-        .{ "18G87", "10.14.6" },
-        .{ "18G95", "10.14.6" },
-        .{ "18G103", "10.14.6" },
-        .{ "18G1012", "10.14.6" },
-        .{ "18G2022", "10.14.6" },
-        .{ "18G3020", "10.14.6" },
-        .{ "18G4032", "10.14.6" },
+    if (std.mem.eql(u8, s_expected, s_got)) return;
```
daurnimator, line 398: Call `testing.expectEqualStrings` here

lib/std/zig/system.zig
mikdusan: on error the .linux prong goes with defaults. Prior to this PR .macos prong would `@panic` if what we think is invariant behaviour on macOS doesn't hold true. The analogous thing to do in this PR would be to `@panic` on say the system .plist file being unreadable which means someone has a corrupt macOS. I don't think there's too much value in that.

daurnimator: > the system .plist file being unreadable which means someone has a corrupt macOS

or they're running in a build chroot without it
or they hit their `ulimit`
or they use one of the security products that can block access to files
etc....

mikdusan: .linux error handling:
- `std.builtin.Version.parse` error → default

.macos error handling:
1. file error → **choose one** { default, default+log, panic }
2. .plist parser error → **choose one** { default, default+log, panic }
3. `std.builtin.Version.parse` error → default

are we ok with macos { 1, 2 } → default+log ?

lib/std/zig/system/macos.zig
```diff
@@ -6,453 +6,408 @@
...
+            if (parseSystemVersion(bytes)) |ver| {
+                // never return red herring
+                if (!(ver.major == 10 and ver.minor >= 16)) {
+                    os.version_range.semver.min = ver;
+                    os.version_range.semver.max = ver;
+                    return;
+                }
+                // continue
+            } else |err| {
+                std.log.warn(format_failure, .{ "parse", path, err });
```
andrewrk, line 72: Why `warn` rather than `debug`?

lib/std/zig/system.zig
andrewrk: In general I'm not a fan of using `warn`. If it's a problem, we should deal with it appropriately, and probably fail. If it's not a problem, why do we need to warn about it?

It looks like with the .linux prong, I made the call that a failure to parse the linux version name would be solved by using the default version range. Logic here is that linux is a diversely configured kernel that perhaps could have strange distros with non-standard uname formats. Or perhaps there is some seccomp stuff going on that is making uname not available. In such cases the default linux version range would be OK.

For macOS, the system is more uniform. I think we should expect version detection to work and report failure if it does not.

andrewrk (changes requested):
I have some requested changes, but the core approach to version detection looks great. Thanks for tackling this problem.

lib/std/zig/system/macos.zig
```diff
@@ -6,453 +6,408 @@
...
+        var buf: [2048]u8 = undefined;
 
-    // parse format-x
-    var yindex: usize = 0;
-    {
-        while (yindex < build.len) : (yindex += 1) {
-            if (build[yindex] < '0' or build[yindex] > '9') break;
+        if (std.fs.cwd().readFile(path, &buf)) |bytes| {
+            if (parseSystemVersion(bytes)) |ver| {
+                // never return red herring
```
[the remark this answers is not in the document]
mikdusan: `10.16` is the backwards compatible version and it's a red herring. Apple is purposely hiding any differentiation between 11.0, 11.1, 11.2, ... . It's low fidelity for our purposes. There is no macOS that returns `10.16` via these mechanisms:

- `sw_vers` command line
- "About This Mac" in desktop Apple menu
-  `system_profiler` GUI
- `system_profiler` command line
- when running Xcode on Big Sur it defaults to `11.1` because it detects it proper

lib/std/zig/system/macos.zig
mikdusan: not sure I really understand the semantic differences and just went with it because it was suggested earlier

mikdusan: ah I misinterpreted your question as referring to `std.debug.warn` ; after scanning `std.log.Level`  I'd recommend `log.err` be used:

```
    /// Error: A bug has been detected or something has gone wrong but it is
    /// recoverable.
```

andrewrk:
> In general I'm not a fan of using warn. If it's a problem, we should deal with it appropriately, and probably fail. If it's not a problem, why do we need to warn about it?

To be clear what I meant was I'm not a fan of using *logging* for error reporting. In this code we end up doing `catch {}` which eats the error, meaning that there was in fact not a problem at all. But just prior to that we logged an error! Something's gone wrong. I don't think this code, which is standard library code, should do any logging at all. It should report the error up to the caller, and let the caller decide what is the situation.

And for this situation:

> For macOS, the system is more uniform. I think we should expect version detection to work and report failure if it does not.

I don't think `catch {}` and sticking with the default version range is the right call. `NativeTargetInfo.detect` should fail if macOS version detection fails.

andrewrk (changes requested):
> In general I'm not a fan of using warn. If it's a problem, we should deal with it appropriately, and probably fail. If it's not a problem, why do we need to warn about it?

To be clear what I meant was I'm not a fan of using *logging* for error reporting in functions such as this. In this code we end up doing `catch {}` which eats the error, meaning that there was in fact not a problem at all. But just prior to that we logged an error! Something's gone wrong. I don't think this code, which is standard library code, should do any logging at all. It should report the error up to the caller, and let the caller decide what is the situation.

And for this situation:

> For macOS, the system is more uniform. I think we should expect version detection to work and report failure if it does not.

I don't think `catch {}` and sticking with the default version range is the right call. `NativeTargetInfo.detect` should fail if macOS version detection fails.

mikdusan:
@andrewrk changes incorporated.

Issue #7706, opened 2021-01-06
linux/sparc64: stage1 seems to miscompile a bitcast in compiler_rt.addXf3


koachan:
On linux/sparc64, Zig seems to miscompile [the last bitcast in compiler_rt.addXf3](https://github.com/ziglang/zig/blob/6c2e0c2/lib/std/special/compiler_rt/addXf3.zig#L235).

The `return @bitCast(T, result)` line gets compiled to the following assembly (annotated for easier reading):
```
stx  %i0, [ %fp + 0x7cf ] # Store the upper half
ldd  [ %fp + 0x7cf ], %f0 # Load it back to the FP register

ldd  [ %i4 ], %f2         # It loads the lower half...
srl  %i1, 0, %i0          # ... before computing it...
sllx  %i3, 0x20, %i1
or  %i1, %i0, %i0
stx  %i0, [ %i4 ]         # ... and stores the result into memory!
```

So, the load of the lower half happens before the result is computed and stored, resulting in garbage values upon return.

koachan:
The correct sequence of instructions should be something like this:
```
stx  %i0, [ %fp + 0x7cf ] # Store the upper half
ldd  [ %fp + 0x7cf ], %f0 # Load it back to the FP register

srl  %i1, 0, %i0          # Compute the lower half
sllx  %i3, 0x20, %i1
or  %i1, %i0, %i0
stx  %i0, [ %i4 ]         # Store the lower half
ldd  [ %i4 ], %f2         # Load it back to the FP register
```

LemonBoy:
LLVM bug triggered at `O1` or more:

```llvm
; ModuleID = 'foo'
source_filename = "foo"
target datalayout = "E-m:e-i64:64-n32:64-S128"
target triple = "sparcv9-unknown-unknown-eabi"

define fp128 @shit(fp128 %0) {
Entry:
  %1 = bitcast fp128 %0 to i128
  %2 = add i128 %1, -1
  %3 = bitcast i128 %2 to fp128
  ret fp128 %3
}
```

I'll have a look at this whenever I have some time.

koachan:
https://bugs.llvm.org/show_bug.cgi?id=47729
Just adding this, apparently it has been known since October?

LemonBoy:
Give this [patch](https://reviews.llvm.org/D94345) a spin.

koachan:
It seem to work for me, thanks!

koachan:
[...]

PR #7716, opened 2021-01-07
stage1: SPARCv9 f128 enablement

koachan:
This is to track the progress of making f128 work on SPARCv9 (sparc64) targets.

At the moment, this still suffers from issue #7706, so some f128 values get miscomputed, but otherwise I believe that this is complete enough for a review.


andrewrk:
nice work.

PR #7721, opened 2021-01-08
self-hosted compiler browser-based wasm playground

andrewrk:
This is a proof-of-concept for several reasons, the most significant one being that it breaks the main compiler because I didn't finish reworking the fields that got moved around.

But in general, the concept works, and with some adjustments I think the maintenance burden on the codebase wouldn't be too severe.

There's a new root source file, `src/playground.zig` which defines a function `zigEval` for js code to call into.

Some problems to solve before this would be reasonable to merge:

 * [ ] cleanup, reduce maintenance burden on the main compiler use case
 * [x] expose errors to the web page
 * [x] implement the file system shim functions
 * [ ] implement more of the actual webassembly backend, at least to the point where we can do hello, world
 * [ ] create a start file that calls into pub fn main
 * [ ] provide the std lib

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -58,6 +58,7 @@ pub fn build(b: *Builder) !void {
     const static_llvm = b.option(bool, "static-llvm", "Disable integration with system-installed LLVM, Clang, LLD, and libc++") orelse false;
     const enable_llvm = b.option(bool, "enable-llvm", "Build self-hosted compiler with LLVM backend enabled") orelse (is_stage1 or static_llvm);
     const config_h_path_option = b.option([]const u8, "config_h", "Path to the generated config.h");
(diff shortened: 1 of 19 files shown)


FireFox317:
Nice stuff! Maybe it is a good idea to apply this concept to all freestanding targets in general? Basically we can have this same "Browser" compiler run on a raspberry pi bare metal, but instead of emitting wasm we emit arm instructions. In that case the compiler can run in any freestanding environment and we won't have specific browser code (`browser.active`) in the source code of the compiler. (I think this can clean up the code structure)

jedisct1:
This is rad!

andrewrk:
image

image

andrewrk:
So this proof of concept works but I'm going to put off finishing & merging this until later when more of the language works and the wasm backend is more feature complete. The branch is there for anyone to play around with.

PR #7723, opened 2021-01-08
stage2: Implicitly enable --eh_frame_hdr when compiling c/c++ files

### LINKED ISSUE #7711: zig c++: compiled kakoune binary gives "libc++abi: terminating with uncaught exception of type Kakoune::failure"
zig c++: compiled kakoune binary gives "libc++abi: terminating with uncaught exception of type Kakoune::failure"

```bash
git clone https://github.com/mawww/kakoune/

cd kakoune
```

```bash
CC='../../zig/zig-linux-x86_64-0.7.1/zig cc' CXX='../../zig/zig-linux-x86_64-0.7.1/zig c++' make install
```

Trying to run the compiled Kakoune binary:
```
$ ./scr/kak
libc++abi: terminating with uncaught exception of type Kakoune::failure
```

Am I missing an option in the zig compiler to compile it correctly? Or is it not supported at the moment? (The machine where I want to compile it eventually has an old compiler which does not support `-std=c++17`. I would like to avoid installing GCC or LLVM from source would.)

I had to fix the makefile to let kakoune pick up ncurses.
```diff
git diff
diff --git a/src/Makefile b/src/Makefile
index 590978ef..a3c09e14 100644
--- a/src/Makefile
+++ b/src/Makefile
@@ -90,8 +90,10 @@ else
        $(error "pkg-config not found in PATH")
     endif
 
-    LIBS += $(shell $(PKG_CONFIG) $(PKG_CONFIG_FLAGS) --libs ncursesw)
-    NCURSES_CFLAGS += $(shell $(PKG_CONFIG) $(PKG_CONFIG_FLAGS) --cflags ncursesw)
+    #LIBS += $(shell $(PKG_CONFIG) $(PKG_CONFIG_FLAGS) --libs ncursesw)
+    #NCURSES_CFLAGS += $(shell $(PKG_CONFIG) $(PKG_CONFIG_FLAGS) --cflags ncursesw)
+    LIBS += $(shell ncursesw6-config --libs)
+    NCURSES_CFLAGS += $(shell ncursesw6-config --cflags)
     LDFLAGS += -rdynamic
 endif
```

LemonBoy:
Matches what Clang/GCC driver do.

Closes #7711

Kakoune builds and runs just fine now.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1007,6 +1007,11 @@ fn buildOutputType(
             ensure_libc_on_non_freestanding = true;
             ensure_libcpp_on_non_freestanding = arg_mode == .cpp;
             want_native_include_dirs = true;
+            // Clang's driver enables this switch unconditionally.
+            // Disabling the emission of .eh_frame_hdr can unexpectedly break
+            // some functionality that depend on it, such as C++ exceptions and
+            // DWARF-based stack traces.
+            link_eh_frame_hdr = true;
 
             const COutMode = enum {
                 link,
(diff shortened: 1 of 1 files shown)


zigazeljko:
Is this also needed for plain C, or just C++?

LemonBoy:
> Is this also needed for plain C, or just C++?

Both of them, C has no exceptions but the unwinding infos are used to obtain stack traces (using `libgcc` or `libunwind`).

PR #7726, opened 2021-01-08
Stage2: Wasm codegen - Implement storing and loading of locals

Luukdegram:
This PR implements the loading and storing of function locals. It does this by first calculating the number of locals found inside a function body and emitting this as part of the 'code' section. 
This also adds the load and store instructions to retrieve and store those values.

For example, the following now works:
```zig
export fn _start() u32 {
    var y: u32 = 5;
    var x: u32 = 20;
    
    y = x;
    return y;
}
```
Which returns the value `20`.

This generates the following pseudo code using wasm-decompile:
```
export function start():int {
  var a:int = 5;
  var b:int = 20;
  a = b;
  return a;
}

### DIFF
--- a/src/codegen/wasm.zig
+++ b/src/codegen/wasm.zig
@@ -78,9 +78,6 @@ pub fn genCode(buf: *ArrayList(u8), decl: *Decl) !void {
             entry.entry.value += 1
         else
             entry.entry.value = 1;
-
-        // append the local to our list of locals so they can be set/retrieved later
-        try decl.fn_link.wasm.?.locals.append(buf.allocator, inst);
     }
 
     // write the amount of different types were found
@@ -117,7 +114,7 @@ fn genInst(buf: *ArrayList(u8), decl: *Decl, inst: *Inst) !void {
         .dbg_stmt => {},
         .ret => genRet(buf, decl, inst.castTag(.ret).?),
         .retvoid => {},
-        .alloc => {}, // already generated in 'genCode'
+        .alloc => genAlloc(buf, decl, inst.castTag(.alloc).?),
         .store => genStore(buf, decl, inst.castTag(.store).?),
         .load => genLoad(buf, decl, inst.castTag(.load).?),
         else => return error.TODOImplementMoreWasmCodegen,
@@ -181,8 +178,7 @@ fn genCall(buf: *ArrayList(u8), decl: *Decl, inst: *Inst.Call) !void {
 }
 
 fn genStore(buf: *ArrayList(u8), decl: *Decl, inst: *Inst.BinOp) !void {
-    const idx = decl.fn_link.wasm.?.getLocalidx(inst.lhs) orelse
-        return error.LocalDoesNotExist; // this is a developer error in the Zig compiler
+    const idx = decl.fn_link.wasm.?.getLocalidx(inst.lhs).?;
 
     const writer = buf.writer();
 
@@ -195,12 +191,14 @@ fn genStore(buf: *ArrayList(u8), decl: *Decl, inst: *Inst.BinOp) !void {
 }
 
 fn genLoad(buf: *ArrayList(u8), decl: *Decl, inst: *Inst.UnOp) !void {
-    const idx = decl.fn_link.wasm.?.getLocalidx(inst.operand) orelse
-        return error.LocalDoesNotExist; // this is a developer error in the Zig compiler
-
+    const idx = decl.fn_link.wasm.?.getLocalidx(inst.operand).?;
     const writer = buf.writer();
 
     // load the local at index `idx` onto the stack
     try writer.writeByte(0x20);
     try leb.writeULEB128(writer, idx);
+}
+
+fn genAlloc(buf: *ArrayList(u8), decl: *Decl, inst: *Inst.NoOp) !void {
+    try decl.fn_link.wasm.?.locals.append(buf.allocator, &inst.base);
 }
--- a/src/codegen/wasm.zig
+++ b/src/codegen/wasm.zig
@@ -58,50 +58,43 @@ pub fn genCode(buf: *ArrayList(u8), decl: *Decl) !void {
     // Calculate the locals of the body
     // We must calculate the occurance of each value type and emit the count of it
     const tv = decl.typed_value.most_recent.typed_value;
-    const mod_fn = tv.val.cast(Value.Payload.Function).?.func;
+    const mod_fn = tv.val.castTag(.function).?.data;
 
-    var locals: [4]u8 = [_]u8{0} ** 4;
-    for (mod_fn.analysis.success.instructions) |inst| {
+    var locals = std.AutoArrayHashMap(u8, u32).init(buf.allocator);
+    defer locals.deinit();
+
+    for (mod_fn.body.instructions) |inst| {
         if (inst.tag != .alloc) continue;
-        const ptr = inst.ty.castPointer().?;
-        const ptr_type = ptr.pointee_type.zigTypeTag();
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/codegen/wasm.zig [hunk 1]
```diff
@@ -140,3 +179,28 @@ fn genCall(buf: *ArrayList(u8), decl: *Decl, inst: *Inst.Call) !void {
         .decl = target,
     });
 }
+
+fn genStore(buf: *ArrayList(u8), decl: *Decl, inst: *Inst.BinOp) !void {
+    const idx = decl.fn_link.wasm.?.getLocalidx(inst.lhs) orelse
+        return error.LocalDoesNotExist; // this is a developer error in the Zig compiler
```
joachimschmidt557, line 185:
```suggestion
    const idx = decl.fn_link.wasm.?.getLocalidx(inst.lhs) orelse
        unreachable; // this is a developer error in the Zig compiler
```

or using the shorthand `.?` here would also make sense.

Luukdegram: Thanks @joachimschmidt557 Fixed this in all codepaths as well as the .alloc in the switch above.

src/codegen/wasm.zig
```diff
@@ -55,15 +55,48 @@ pub fn genCode(buf: *ArrayList(u8), decl: *Decl) !void {
...
 
-    // Write the size of the locals vec
-    // TODO: implement locals
-    try leb.writeULEB128(writer, @as(u32, 0));
+    // Calculate the locals of the body
+    // We must calculate the occurance of each value type and emit the count of it
+    const tv = decl.typed_value.most_recent.typed_value;
+    const mod_fn = tv.val.castTag(.function).?.data;
+
+    var locals = std.AutoArrayHashMap(u8, u32).init(buf.allocator);
```
ifreund, line 63: This will never have more than 4 keys so it could really be a `[4]u32` and avoid any allocation. Not sure if the semantics of the hashmap are worth it here for readability or not. If you do stick with the hashmap I don't think you rely on the ordering so the normal non-array hashmap would be preferable.

src/link/Wasm.zig
```diff
@@ -45,6 +46,15 @@ pub const FnData = struct {
...
     idx_refs: std.ArrayListUnmanaged(struct { offset: u32, decl: *Module.Decl }) = .{},
+    /// A list of `Inst` used to save and retrieve the indices of local variables
+    locals: std.ArrayListUnmanaged(*Inst) = .{},
+
+    /// Returns the index of a local given a pointer to an `Inst`
+    pub fn getLocalidx(self: *FnData, inst: *Inst) ?u32 {
+        return for (self.locals.items) |local, idx| {
+            if (local == inst) break @intCast(u32, idx);
+        } else null;
+    }
```
ifreund, line 57: Note that this will break when function arguments are implemented as they share the index space with locals. Might want to add a TODO.

src/codegen/wasm.zig
```diff
@@ -81,7 +114,10 @@ fn genInst(buf: *ArrayList(u8), decl: *Decl, inst: *Inst) !void {
         .dbg_stmt => {},
         .ret => genRet(buf, decl, inst.castTag(.ret).?),
         .retvoid => {},
-        else => error.TODOImplementMoreWasmCodegen,
+        .alloc => genAlloc(buf, decl, inst.castTag(.alloc).?),
+        .store => genStore(buf, decl, inst.castTag(.store).?),
+        .load => genLoad(buf, decl, inst.castTag(.load).?),
+        else => return error.TODOImplementMoreWasmCodegen,
```
ifreund, line 120: This return isn't necessary.
```suggestion
        else => error.TODOImplementMoreWasmCodegen,
```

src/codegen/wasm.zig
```diff
@@ -140,3 +176,29 @@ fn genCall(buf: *ArrayList(u8), decl: *Decl, inst: *Inst.Call) !void {
...
+    try writer.writeByte(0x21);
+    try leb.writeULEB128(writer, idx);
+}
+
+fn genLoad(buf: *ArrayList(u8), decl: *Decl, inst: *Inst.UnOp) !void {
+    const idx = decl.fn_link.wasm.?.getLocalidx(inst.operand).?;
+    const writer = buf.writer();
+
+    // load the local at index `idx` onto the stack
+    try writer.writeByte(0x20);
```
ifreund, line 198: I've been trying to follow the convention of adding a comment on the same line with the name of the opcode from the spec so far which for me at least reduces the number of times I need to look at the docs. 
```suggestion
    try writer.writeByte(0x20); // local.get
```
We could consider creating constants for all the instructions eventually, though I don't think that's necessary yet.

ifreund (commented):
Looking good! just some minor suggestions

src/codegen/wasm.zig
Luukdegram: > We could consider creating constants for all the instructions eventually, though I don't think that's necessary yet.

Agreed, as we will have quite a few more constants such as block types as well.

src/codegen/wasm.zig
Luukdegram: I actually had this initially (9a34d6c) but in hindsight, the hashmap didn't really offer more readability than I had hoped for.

Luukdegram:
Thanks for the review @ifreund. I resolved your feedback. I will bring a proposal to the next stage2 meeting that will solve quite a few problems that I was having when implementing while loops (and all other codegen required to make a while loop work (https://github.com/Luukdegram/zig/commit/2e2707e5c1a06254cc796ddca4d926b18cd1445b)). In case you're bored or curious, I already implemented the proposal in a different branch: https://github.com/Luukdegram/zig/commit/5bb91068c7d6fb19b6613f11679d74ce7b76fb15

src/codegen/wasm.zig
```diff
@@ -55,18 +55,60 @@ pub fn genCode(buf: *ArrayList(u8), decl: *Decl) !void {
...
+        const alloc = inst.castTag(.alloc).?;
+        const elem_type = alloc.base.ty.elemType();
+
+        switch (genValtype(elem_type)) {
+            0x7F => locals[0] += 1,
+            0x7E => locals[1] += 1,
+            0x7D => locals[2] += 1,
+            0x7C => locals[3] += 1,
+            else => unreachable,
+        }
```
ifreund, line 76: This could just be
```suggestion
        // valtypes are 0x7C, 0x7D, 0x7E, or 0x7F
        locals[genValtype(elem_type) - 0x7C] += 1;
```
If the code below is updated in a similar fashion.

src/codegen/wasm.zig
```diff
@@ -140,3 +185,27 @@ fn genCall(buf: *ArrayList(u8), decl: *Decl, inst: *Inst.Call) !void {
...
+fn genLoad(buf: *ArrayList(u8), decl: *Decl, inst: *Inst.UnOp) !void {
+    const idx = decl.fn_link.wasm.?.getLocalidx(inst.operand).?;
+    const writer = buf.writer();
+
+    try writer.writeByte(0x20); // local.get
+    try leb.writeULEB128(writer, idx);
+}
+
+fn genAlloc(buf: *ArrayList(u8), decl: *Decl, inst: *Inst.NoOp) !void {
+    try decl.fn_link.wasm.?.locals.append(buf.allocator, &inst.base);
```
ifreund, line 210: I don't see any code that would guarantee that this next index has the type required for the given alloc instruction

Luukdegram: @ifreund Unfortunately the latest commit (6958a56) was the best I could come up with to make this work.

ifreund (approved):
LGTM, thanks for working on this!

andrewrk:
Closing since #7797 has been merged.

PR #7727, opened 2021-01-08
stage2: fix bug in genArg

joachimschmidt557:
When an argument is unused in the function body, still increment `arg_index` so we still select the correct arguments in the `args` slice.

This example now works. Previously, the assert failed.

```zig
export fn _start() noreturn {
    assert(select4th(1, 2, 3, 4) == 4);
    exit(0);
}

fn select4th(a: u32, b: u32, c: u32, d: u32) u32 {
    return d;
}

fn assert(ok: bool) void {
}

fn exit(code: usize) noreturn {
    asm volatile ("svc #0"
        :
        : [number] "{r7}" (1),
          [arg1] "{r0}" (code)
        : "memory"
    );
    unreachable;
}
```

### DIFF
--- a/src/codegen.zig
+++ b/src/codegen.zig
@@ -1480,6 +1480,9 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
         }
 
         fn genArg(self: *Self, inst: *ir.Inst.Arg) !MCValue {
+            const arg_index = self.arg_index;
+            self.arg_index += 1;
+
             if (FreeRegInt == u0) {
                 return self.fail(inst.base.src, "TODO implement Register enum for {}", .{self.target.cpu.arch});
             }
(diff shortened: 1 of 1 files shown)


src/codegen.zig
```diff
@@ -1480,6 +1480,8 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
         }
 
         fn genArg(self: *Self, inst: *ir.Inst.Arg) !MCValue {
+            defer self.arg_index += 1;
```
andrewrk, line 1483: Can I suggest a fix that generates less code?

```suggestion
            const arg_index = self.arg_index;
            self.arg_index += 1;
```

(and then use `arg_index` instead of `self.arg_index` below)

Otherwise we needlessly have code that runs at every return point of the function (including every `try`)

With stage2 codegen it's less important, since we send all control flow to the same function return block, but stage1 duplicates the defer expressions at every return point.

andrewrk (changes requested):
Thanks for the fix!

src/codegen.zig
joachimschmidt557: I agree, this makes more sense.

PR #7730, opened 2021-01-08
stage2: fix orelse at comptime + tests

### LINKED ISSUE #7662: stage2: orelse at comptime is broken
This should work, but it does not:
```zig
export fn _start() noreturn {
    const b: ?comptime_int = 5;
    const v: comptime_int = b orelse 4;
    unreachable;
}
```
```
test.zig:3:31: error: expected pointer, found 'comptime_int'
```
Here is the untyped zir: 
```
fn_type _start { // unanalyzed
  %0 = const(TypedValue{ .ty = type, .val = type})
  %1 = const(TypedValue{ .ty = type, .val = noreturn})
  %2 = as(%0, %1)
  %3 = fntype([], %2, cc=Unspecified)
} // fn_type _start

fn_body _start { // unanalyzed
  %0 = dbg_stmt()
  %1 = const(TypedValue{ .ty = type, .val = type})
  %2 = const(TypedValue{ .ty = type, .val = type})
  %3 = const(TypedValue{ .ty = type, .val = comptime_int})
  %4 = as(%2, %3)
  %5 = optional_type(%4)
  %6 = as(%1, %5)
  %7 = const(TypedValue{ .ty = comptime_int, .val = 5})
  %8 = as(%6, %7) ; I think this is the problem. An optional comptime int is not a pointer, but if it is null, it never gets referenced.
  %9 = dbg_stmt()
  %10 = const(TypedValue{ .ty = type, .val = type})
  %11 = const(TypedValue{ .ty = type, .val = comptime_int})
  %12 = as(%10, %11)
  %13 = ref(%8)
  %14 = deref(%13)
  %15 = isnull(%14)
  %16 = block({
    %17 = condbr(%15, {
      %18 = const(TypedValue{ .ty = comptime_int, .val = 4})
      %19 = as(%12, %18)
      %20 = break("label_16", %19)
    }, {
      %21 = unwrap_optional_unsafe(%13)
      %22 = break("label_16", %21)
    })
  })
  %23 = deref(%16)
  %24 = as(%12, %23)
  %25 = dbg_stmt()
  %26 = unreachable()
} // fn_body _start
```
I think the problem is the `%23 = deref(%16)` line, which is coming from `rlWrapPtr` at the end of `orelseCatchExpr`
Adding this does not fix it: 
```diff
32c32,33
<       %22 = break("label_16", %21)
---
>       %22 = ref(%21)
>       %23 = break("label_16", %22)
```
Note, this does work:
```zig
export fn _start() noreturn {
    const b: ?comptime_int = null;
    const v: comptime_int = b orelse 4;
    unreachable;
}
```
This seems like a very easy fix, I just cannot think of a fix.

Edit:
Im pretty sure it is because of the resultLoc mechanism:
```
  %16 = block({
    %17 = condbr(%15, {
      %18 = const(TypedValue{ .ty = comptime_int, .val = 4})
      %19 = as(%12, %18)
      %20 = break("label_16", %19)
    }, {
      %21 = unwrap_optional_unsafe(%13)
      %22 = break("label_16", %21)
    })
  })
```
Here, `%19 = as(%12, %18)` is because of the .ty resultloc. We also need to use the resultloc in the else branch

Edit Edit:
it is actually because of the `rlWrapPtr` at the end of orelseCatchExpr. Changing it to `rlWrap` fixes it for the case where `b != null`, but breaks for `b == null`. It shows this:
```
fn_type _start { // unanalyzed
  %0 = const(TypedValue{ .ty = type, .val = type})
  %1 = const(TypedValue{ .ty = type, .val = noreturn})
  %2 = as(%0, %1)
  %3 = fntype([], %2, cc=Unspecified)
} // fn_type _start

fn_body _start { // unanalyzed
  %0 = dbg_stmt()
  %1 = const(TypedValue{ .ty = type, .val = type})
  %2 = const(TypedValue{ .ty = type, .val = type})
  %3 = const(TypedValue{ .ty = type, .val = comptime_int})
  %4 = as(%2, %3)
  %5 = optional_type(%4)
  %6 = as(%1, %5)
  %7 = const(TypedValue{ .ty = @Type(.Null), .val = null})
  %8 = as(%6, %7)
  %9 = dbg_stmt()
  %10 = const(TypedValue{ .ty = type, .val = type})
  %11 = const(TypedValue{ .ty = type, .val = comptime_int})
  %12 = as(%10, %11)
  %13 = ref(%8)
  %14 = deref(%13)
  %15 = isnull(%14)
  %16 = block({
    %17 = condbr(%15, {
      %18 = const(TypedValue{ .ty = comptime_int, .val = 4})
      %19 = as(%12, %18)
      %20 = break("label_16", %19)
    }, {
      %21 = unwrap_optional_unsafe(%13)
      %22 = break("label_16", %21)
    })
  })
  %23 = as(%12, %16)
  %24 = dbg_stmt()
  %25 = unreachable()
} // fn_body _start

test.zig:3:31: error: expected comptime_int, found *comptime_int
```

The reason I am submitting this is because it is a blocker for writing tests for #7628 since this also affects catch, which #7628 allows to be run at comptime.

g-w1:
There was just some untested code that did not work. .isnull -> .isnonnull, and I had to add a .ref resultloc to make types match up.

I also added tests so this doesn't happen in the future.

Fixes #7662
unblocks #7628

### DIFF
--- a/src/astgen.zig
+++ b/src/astgen.zig
@@ -885,7 +885,7 @@ fn catchExpr(mod: *Module, scope: *Scope, rl: ResultLoc, node: *ast.Node.Catch) 
 }
 
 fn orelseExpr(mod: *Module, scope: *Scope, rl: ResultLoc, node: *ast.Node.SimpleInfixOp) InnerError!*zir.Inst {
-    return orelseCatchExpr(mod, scope, rl, node.lhs, node.op_token, .isnull, .unwrap_optional_unsafe, node.rhs, null);
+    return orelseCatchExpr(mod, scope, rl, node.lhs, node.op_token, .isnonnull, .unwrap_optional_unsafe, node.rhs, null);
 }
 
 fn orelseCatchExpr(
@@ -963,7 +963,7 @@ fn orelseCatchExpr(
 
     _ = try addZIRInst(mod, &then_scope.base, src, zir.Inst.Break, .{
         .block = block,
-        .operand = try expr(mod, then_sub_scope, branch_rl, rhs),
+        .operand = try rlWrap(mod, then_sub_scope, .{ .ref = {} }, try expr(mod, then_sub_scope, branch_rl, rhs)),
     }, .{});
 
     var else_scope: Scope.GenZIR = .{
--- a/test/stage2/test.zig
+++ b/test/stage2/test.zig
@@ -1495,4 +1495,51 @@ pub fn addCases(ctx: *TestContext) !void {
             \\}
         , &[_][]const u8{":8:10: error: evaluation exceeded 1000 backwards branches"});
     }
+    {
+        var case = ctx.exe("orelse at comptime", linux_x64);
+        case.addCompareOutput(
+            \\export fn _start() noreturn {
+            \\    const i: ?u64 = 0;
+            \\    const orelsed = i orelse 5;
+            \\    assert(orelsed == 0);
+            \\    exit();
+            \\}
+            \\fn assert(b: bool) void {
+            \\    if (!b) unreachable;
+            \\}
+            \\fn exit() noreturn {
+            \\    asm volatile ("syscall"
+            \\        :
+            \\        : [number] "{rax}" (231),
+            \\          [arg1] "{rdi}" (0)
+            \\        : "rcx", "r11", "memory"
+            \\    );
+            \\    unreachable;
+            \\}
+        ,
+            "",
+        );
+        case.addCompareOutput(
+            \\export fn _start() noreturn {
+            \\    const i: ?u64 = null;
+            \\    const orelsed = i orelse 5;
+            \\    assert(orelsed == 5);
+            \\    exit();
+            \\}
+            \\fn assert(b: bool) void {
+            \\    if (!b) unreachable;
+            \\}
+            \\fn exit() noreturn {
+            \\    asm volatile ("syscall"
+            \\        :
+            \\        : [number] "{rax}" (231),
+            \\          [arg1] "{rdi}" (0)
+            \\        : "rcx", "r11", "memory"
+            \\    );
+            \\    unreachable;
+            \\}
+        ,
+            "",
+        );
+    }
 }


FireFox317:
Nice work!

g-w1:
Catch is probably also broken, but that will be fixed in #7628, with tests.

src/astgen.zig
```diff
@@ -963,7 +963,7 @@ fn orelseCatchExpr(
 
     _ = try addZIRInst(mod, &then_scope.base, src, zir.Inst.Break, .{
         .block = block,
-        .operand = try expr(mod, then_sub_scope, branch_rl, rhs),
+        .operand = try rlWrap(mod, then_sub_scope, .{ .ref = {} }, try expr(mod, then_sub_scope, branch_rl, rhs)),
```
andrewrk, line 966: This part doesn't look right - what are you trying to do here?

g-w1: It expected a pointer, but found a value, so making a pointer. Is `addZIRUnOp(mod,then_sub_scope,.ref, ...)` more idomatic? I thought a rlWrap was the best.

andrewrk: What do you mean "it expected a pointer"?
The docs say:

```
        /// Return a value from a `Block`.
        @"break",
```

and the implementation has nothing there to suggest the value should be a pointer.

andrewrk (changes requested):
The new test cases are nice, thanks for those. But this fix is not right. Before, the ZIR is:

```
  %12 = isnull(%11)
  %13 = block({
    %14 = condbr(%12, {
      %15 = const(TypedValue{ .ty = comptime_int, .val = 5})
      %16 = break("label_13", %15)
    }, {
      %17 = unwrap_optional_unsafe(%10)
      %18 = break("label_13", %17)
    })
  })
```

After this PR it is:

```
  %12 = isnonnull(%11)
  %13 = block({
    %14 = condbr(%12, {
      %15 = const(TypedValue{ .ty = comptime_int, .val = 5})
      %16 = ref(%15)
      %17 = break("label_13", %16)
    }, {
      %18 = unwrap_optional_unsafe(%10)
      %19 = break("label_13", %18)
    })
  })
```

The inverted logic appears to be in zir_sema.zig:

```zig
        .isnull => return analyzeInstIsNonNull(mod, scope, old_inst.castTag(.isnull).?, true),
        .isnonnull => return analyzeInstIsNonNull(mod, scope, old_inst.castTag(.isnonnull).?, false),
```

```zig
fn analyzeInstIsNonNull(mod: *Module, scope: *Scope, inst: *zir.Inst.UnOp, invert_logic: bool) InnerError!*Inst {
    const operand = try resolveInst(mod, scope, inst.positionals.operand);
    return mod.analyzeIsNull(scope, inst.base.src, operand, invert_logic);
}
```

As for the ref, I think the fix is going to be a *deref* in the other branch instead, or better yet, use values the whole time, no pointers.

Issue #6590, opened 2020-10-06
random crashes in event loop using async/await


vrischmann:
Hi,

I wrote a basic program using async/await which starts a `std.net.StreamServer`, starts a bunch connections to it, each connection sends some requests to it and reads the responses.
The code can be found [here](https://git.sr.ht/~vrischmann/zig-eventedio-lock/tree/master/src/main.zig).

When I run the program it appears to work most of the time but sometimes it fails with this error:
```
info(server): waiting for connection
info(server): got connection: Connection{ .file = File{ .handle = 14, .capable_io_mode = Mode.evented, .intended_io_mode = Mode.evented }, .address = 127.0.0.1:41632 }
info(session): got request: request
reached unreachable code

warning(server): listen loop stopped
/home/vincent/dev/devtools/zig/lib/std/event/loop.zig:445:60: 0x236f66 in std.event.loop.Loop.linuxWaitFd (zig-eventedio-lock)
                error.FileDescriptorAlreadyPresentInSet => unreachable, // evented writes to the same fd is not thread-safe
                                                           ^
/home/vincent/tmp/zig-eventedio-lock/src/main.zig:121:53: 0x24dc35 in Tester.run (zig-eventedio-lock)
                var frame: @Frame(sendAndReceive) = async self.sendAndReceive(conn, &lock);
                                                    ^
/home/vincent/dev/devtools/zig/lib/std/event/loop.zig:1213:25: 0x262b7b in std.event.loop.Loop.workerRun (zig-eventedio-lock)
                        resume handle;
                        ^
/home/vincent/dev/devtools/zig/lib/std/thread.zig:235:32: 0x26b563 in std.thread.MainFuncs.linuxThreadMain (zig-eventedio-lock)
                        startFn(arg);
                               ^
/home/vincent/dev/devtools/zig/lib/std/special/c.zig:247:13: 0x2a46ae in clone (c)
            asm volatile (
```

The error can be triggered easily however:
* if the program is run with strace like this: `strace -c ./zig-cache/bin/zig-eventedio-lock`. In that case it almost always fails.
* if the program is run in a loop like this: `for i in (seq 1 100) ; ./zig-cache/bin/zig-eventedio-lock ; end` it will eventually crash.

I'm using Zig version `0.6.0+58502b8bf`.

frmdstryr:
Also see this (different code base)... added a few prints in the linuxWaitFd fn in loop.zig
```
Listing on 127.0.0.1:9000
[...]

PR #7755, opened 2021-01-12
fix race condition in linuxWaitFd

billzez:
I frequently hit this race condition with evented networking on linux.

Thread 1: suspends coroutine, adds fd to epoll
Thread 2: resumes coroutine, need_to_delete is still false, linuxRemoveFd does not get called
Thread 1: sets need_to_delete to true, but fd stays in epoll, and the following call to linuxWaitFd will fail

This simple fix is to invert the logic for need_to_delete.

### DIFF
--- a/lib/std/event/loop.zig
+++ b/lib/std/event/loop.zig
@@ -440,13 +440,11 @@ pub const Loop = struct {
                 .overlapped = ResumeNode.overlapped_init,
             },
         };
-        var need_to_delete = false;
+        var need_to_delete = true;
         defer if (need_to_delete) self.linuxRemoveFd(fd);
 
         suspend {
-            if (self.linuxAddFd(fd, &resume_node.base, flags)) |_| {
-                need_to_delete = true;
-            } else |err| switch (err) {
+            self.linuxAddFd(fd, &resume_node.base, flags) catch |err| switch (err) {
                 error.FileDescriptorNotRegistered => unreachable,
                 error.OperationCausesCircularLoop => unreachable,
                 error.FileDescriptorIncompatibleWithEpoll => unreachable,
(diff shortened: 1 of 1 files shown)


billzez:
Adding a bit more information.  Without this patch, evented networking on linux is too unstable to use.  When the fd does not get removed from epoll, this process crashes:

```
thread 3476 panic: reached unreachable code
/opt/zig/lib/std/event/loop.zig:453:60: 0x246b46 in std.event.loop.Loop.linuxWaitFd (axzez-server)
                error.FileDescriptorAlreadyPresentInSet => unreachable, // evented writes to the same fd is not thread-safe
                                                           ^
/opt/zig/lib/std/event/loop.zig:1354:25: 0x27d03b in std.event.loop.Loop.workerRun (axzez-server)
                        resume handle;
                        ^
/opt/zig/lib/std/event/loop.zig:641:23: 0x27778d in std.event.loop.Loop.run (axzez-server)
        self.workerRun();
                      ^
/home/bill/server/src/process-manager.zig:202:26: 0x2651ee in process-manager.ProcessManager.start (axzez-server)
            self.loop.run();
                         ^
/home/bill/server/src/main.zig:105:19: 0x258d1a in main (axzez-server)
    axzez.pm.start();
                  ^
/opt/zig/lib/std/start.zig:335:22: 0x218dee in std.start.posixCallMainAndExit (axzez-server)
            root.main();
                     ^
/opt/zig/lib/std/start.zig:163:5: 0x218cc2 in std.start._start (axzez-server)
    ^
1/26 03:07:31 P         pm process exited with signal 6
```

frmdstryr:
Unfortunately this doesn't work for me... 

I get a crash that looks worse than the original:

```
???:?:?: 0x2458f1 in ??? (???)???:?:????:?:?
: : 0x0x2673fb???:?:?2458f1 in  in : ?????? ( (0x???)???2673fb
 in )??? (
???)
Bus error at address 0x???:?:?0
: 0x267403 in ??? (???)
???:?:?: 0x267403 in ??? (???)
???:?:?: 0x2350b0 in ??? (???)


```

Can you see if this patch/workaround here solves your case? https://github.com/ziglang/zig/commit/5279e4920b27ab5def1145e91be4071bd0dd655b

This issue was only reported in https://github.com/ziglang/zig/issues/6590 has been around for over a year now https://freenode.irclog.whitequark.org/zig/search?q=FileDescriptorAlreadyPresentInSet

PR #7757, opened 2021-01-12
define nfds_t for windows

billzez:
This fixes a regression with poll() on windows.

### DIFF
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -1311,8 +1311,8 @@ pub fn recvfrom(s: ws2_32.SOCKET, buf: [*]u8, len: usize, flags: u32, from: ?*ws
     }
 }
 
-pub fn poll(fds: [*]ws2_32.pollfd, n: usize, timeout: i32) i32 {
-    return ws2_32.WSAPoll(fds, @intCast(u32, n), timeout);
+pub fn poll(fds: [*]ws2_32.pollfd, n: c_ulong, timeout: i32) i32 {
+    return ws2_32.WSAPoll(fds, n, timeout);
 }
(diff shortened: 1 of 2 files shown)


lib/std/os/windows.zig
```diff
@@ -1311,8 +1311,8 @@ pub fn recvfrom(s: ws2_32.SOCKET, buf: [*]u8, len: usize, flags: u32, from: ?*ws
     }
 }
 
-pub fn poll(fds: [*]ws2_32.pollfd, n: usize, timeout: i32) i32 {
-    return ws2_32.WSAPoll(fds, @intCast(u32, n), timeout);
+pub fn poll(fds: [*]ws2_32.pollfd, n: c_ulong, timeout: i32) i32 {
```
marler8997, line 1314: shouldn't it be `n: nfds_t`?

billzez: I was trying to follow convention, which seems to be to use win32 types (eg ws2_32.SOCKET) in os/windows.zig and posix types (eg nfds_t) in os.zig.

marler8997: doesn't Windows use `nfds_t`?  Or does it using "unsigned long" ?  I can't even find the documentation for Windows "poll" function, so I just assumed it uses the conventional definition which is `nfds_t` right?

billzez: afaik, windows just has WSAPoll.  mingw does not even define nfds_t.  Zig using nfds_t seems consistent with how sock_t is defined to abstract SOCKET/int.

andrewrk:
Thanks!

PR #7761, opened 2021-01-12
add missing ECONNRESET from getsockoptError

billzez:
I've observed this errno occurring on Linux.


marler8997:
Can you share your kernel version/distribution details?

billzez:
kernel 5.4.83
alpine 3.12

marler8997:
Could you provide an example program that can reproduce this error code on connect?

PR #7762, opened 2021-01-13
std.os.uefi.protocols.FileProtocol: fix and expose get_position, set_position

kivikakk:
`EFI_FILE_PROTOCOL.SetPosition` expects an immediate argument, not a pointer. Also expose these for use on struct instances, and add a const for when we want to seek to the end.

Reference: UEFI Spec v2.8, pp 520–521. Tested as working in QEMU.


kivikakk:
It occurs to me most of our UEFI wrappers use camelcase (like [`SimpleFileSystemProtocol.openVolume`](lib/std/os/uefi/protocols/simple_file_system_protocol.zig)) -- I was copying `get_info` and `set_info` below, but perhaps these should all be changed. Will do now.

daurnimator (approved):
I assume the

andrewrk:
Thanks!

PR #7763, opened 2021-01-13
std.elf: expose parsing decoupled from std.fs.File

kivikakk:
I don't know what the maintenance status of `std.elf` is, or how much we want to expose. This PR separates the parsing in `std.elf.readHeader` into `std.elf.parseHeader` for non-`File`-based use.

### DIFF
--- a/lib/std/elf.zig
+++ b/lib/std/elf.zig
@@ -334,6 +334,40 @@ pub const ET = extern enum(u16) {
     pub const HIPROC = 0xffff;
 };
 
+pub const FileParseSource = struct {
+    file: std.fs.File,
+
+    fn readNoEof(self: FileParseSource, buf: []u8, offset: u64) !void {
+        var i: usize = 0;
+        while (i < buf.len) {
+            const len = self.file.pread(buf[i .. buf.len - i], offset + i) catch |err| switch (err) {
+                error.SystemResources => return error.SystemResources,
+                error.IsDir => return error.UnableToReadElfFile,
+                error.OperationAborted => return error.UnableToReadElfFile,
+                error.BrokenPipe => return error.UnableToReadElfFile,
+                error.Unseekable => return error.UnableToReadElfFile,
+                error.ConnectionResetByPeer => return error.UnableToReadElfFile,
+                error.ConnectionTimedOut => return error.UnableToReadElfFile,
+                error.InputOutput => return error.FileSystem,
+                error.Unexpected => return error.Unexpected,
+                error.WouldBlock => return error.Unexpected,
+                error.NotOpenForReading => return error.Unexpected,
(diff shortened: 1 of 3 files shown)


lib/std/elf.zig
```diff
@@ -365,8 +365,12 @@ const Header = struct {
 pub fn readHeader(file: File) !Header {
     var hdr_buf: [@sizeOf(Elf64_Ehdr)]u8 align(@alignOf(Elf64_Ehdr)) = undefined;
     try preadNoEof(file, &hdr_buf, 0);
-    const hdr32 = @ptrCast(*Elf32_Ehdr, &hdr_buf);
-    const hdr64 = @ptrCast(*Elf64_Ehdr, &hdr_buf);
+    return parseHeader(hdr_buf);
+}
+
+pub fn parseHeader(hdr_buf: *align(@alignOf(Elf64_Ehdr)) const [@sizeOf(Elf64_Ehdr)]u8) !Header {
```
daurnimator, line 371: Lets put this inside of the `Header` struct as just `parse`?

kivikakk: Done.

kivikakk:
Some build failures in CI, will mark as ready for review when addressed.

kivikakk:
@daurnimator All good now.

kivikakk:
Turns out I wanted the phdrs and shdrs too. A slightly more heavyweight change has resulted.

`anytype` nonsense is because a `union(enum)` that contained a `File` would be uninstantiable on platforms without `std.os.fd_t`.

LemonBoy:
The `*ParseSource` types are not needed if you require the stream object to implement the `io.Reader` and `io.SeekableStream` interface. This way you can pass a `fs.File` or a `io.FixedBufferStream` (or some other custom type) and have it work as wanted.

kivikakk:
@LemonBoy That's a much better idea, thanks! I'll do that.

kivikakk:
This work very nicely. Sample usage:

```zig
var elf_buffer = std.io.fixedBufferStream(dainkrnl[0..dainkrnl_size]);
dainkrnl_elf = elf.Header.read(&elf_buffer) catch |err| {
    printf("failed to parse ELF: {}\r\n", .{err});
    halt();
};
// ...
var it = dainkrnl_elf.?.program_header_iterator(&elf_buffer);
while (it.next() catch haltMsg("iterating phdr")) |phdr| {
    printf(" * type={x:0>8} off={x:0>16} vad={x:0>16} pad={x:0>16} fsz={x:0>16} msz={x:0>16}\r\n", .{ phdr.p_type, phdr.p_offset, phdr.p_vaddr, phdr.p_paddr, phdr.p_filesz, phdr.p_memsz });
}
```

PR #7767, opened 2021-01-13
std.os: add atexit() implementation

paulsmith:
This change adds a Zig wrapper for the C stdlib `atexit(3)` function, which registers a function to be called at program exit, either from normal termination or from main() returning.

### DIFF
--- a/lib/std/c.zig
+++ b/lib/std/c.zig
@@ -82,6 +82,7 @@ pub extern "c" fn printf(format: [*:0]const u8, ...) c_int;
 pub extern "c" fn abort() noreturn;
 pub extern "c" fn exit(code: c_int) noreturn;
 pub extern "c" fn _exit(code: c_int) noreturn;
+pub extern "c" fn atexit(function: ?fn () callconv(.C) void) c_int;
 pub extern "c" fn isatty(fd: fd_t) c_int;
 pub extern "c" fn close(fd: fd_t) c_int;
 pub extern "c" fn lseek(fd: fd_t, offset: off_t, whence: c_int) off_t;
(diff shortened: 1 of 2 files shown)


andrewrk:
I want to discourage programmers from calling this function because it's always the wrong thing to do. There is no valid use case for ever calling this function.

paulsmith:
Haha, ok, fair enough.

Just so you know where I'm coming from: I'm working on a simple text editor that enables raw mode on the terminal when it starts up by setting the appropriate termios flags and calling tcsetattr(). The way I'm handling restoring terminal settings is by registering an atexit() function that reverts to the original settings. I want to make sure this is done when main() exits normally or in any exit condition. Isn't that a good use-case for atexit()? What are the downsides? Is your suggestion that I set up signal handlers to do the same?

andrewrk:
You control main() right? So you can put the first line of your main function `defer resetTheTerminal();`

paulsmith:
What if the process is `kill -9`'d? Does `defer` run in that case?

andrewrk:
with `kill -9`, neither atexit handlers nor defers will run.

Problems with atexit:
 * if a library calls it, you can't de-initialize the library without exiting the entire process
 * It depends on libc
 * it's a problematic way to organize code; there is already a main function which is the simplest place to put this logic
 * it prompts more questions, such as what happens if you exit during an atexit handler. There's a really simple way to answer these questions: just put the logic in main()
 * it tricks programmers into thinking they have handled "save on exit" or similar features, but in reality there are many ways for a process to be terminated without atexit handlers running, such as power failure.

paulsmith:
Alright, I'm convinced!

PR #7768, opened 2021-01-13
crypto: add legacy keccak hash functions

gballet:
Major cryptocurrency projects still use the original Keccak function parameters (1, instead of 6). This PR makes them available from `std.crypto`.


jedisct1:
Sure!

jedisct1 (approved):

PR #7778, opened 2021-01-14
stage2: error set merging with tests

g-w1:
I had to come up with creative tests because we don't have error set type equality yet. Since this is in zir_sema.zig it should be pretty compatible with inferred error sets throwing compile errors.

### DIFF
--- a/src/zir_sema.zig
+++ b/src/zir_sema.zig
@@ -1257,13 +1257,13 @@ fn zirMergeErrorSets(mod: *Module, scope: *Scope, inst: *zir.Inst.BinOp) InnerEr
         .error_set => {
             var multiple = rhs_ty.castTag(.error_set).?.data.typed_value.most_recent.typed_value.val.castTag(.error_set).?.data.fields;
             var it = multiple.iterator();
-            while (it.next()) |name| {
-                const entry = try mod.getErrorValue(name.key);
+            while (it.next()) |entry| {
                 payload.data.fields.putAssumeCapacity(entry.key, entry.value);
             }
         },
         else => unreachable,
     }
+    // TODO create name in format "error:line:column"
     const new_decl = try mod.createAnonymousDecl(scope, &new_decl_arena, .{
         .ty = Type.initTag(.type),
         .val = Value.initPayload(&payload.base),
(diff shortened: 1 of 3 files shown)


src/zir_sema.zig
```diff
@@ -1168,7 +1168,84 @@ fn analyzeInstErrorSet(mod: *Module, scope: *Scope, inst: *zir.Inst.ErrorSet) In
...
+    if (lhs_ty.tag() == .anyerror or rhs_ty.tag() == .anyerror)
+        return mod.constInst(scope, inst.base.src, .{
+            .ty = Type.initTag(.type),
+            .val = Value.initTag(.anyerror_type),
+        });
+    // The declarations arena will store the hashmap.
+    var new_decl_arena = std.heap.ArenaAllocator.init(mod.gpa);
+    errdefer new_decl_arena.deinit();
+
+    const payload = try scope.arena().create(Value.Payload.ErrorSet);
```
Vexu, line 1189: This payload has to be allocated with `new_decl_arena` to avoid use after free bugs.

src/zir_sema.zig
```diff
@@ -1168,7 +1168,84 @@ fn analyzeInstErrorSet(mod: *Module, scope: *Scope, inst: *zir.Inst.ErrorSet) In
...
+    const new_decl = try mod.createAnonymousDecl(scope, &new_decl_arena, .{
+        .ty = Type.initTag(.type),
+        .val = Value.initPayload(&payload.base),
+    });
+    payload.data.decl = new_decl;
+
+    return mod.constInst(scope, inst.base.src, .{
+        .ty = Type.initTag(.type),
+        .val = Value.initPayload(&payload.base),
+    });
```
Vexu, line 1248:
```suggestion
    return mod.analyzeDeclVal(scope, inst.base.src, new_decl);
```

Vexu (changes requested):

src/zir_sema.zig
Vexu: Looking at this closer it seems that `analyzeDeclVal` doesn't actually copy the decl val as I expected so it is just extra work for `mod.constType`. I don't think we ever modify values in place so the lack of copy likely doesn't even matter idk. If you do change this you should probably also change it in `analyzeInstErrorSet`

g-w1: I think it is fine. In the future andrewrk said that an optimization opportunity would be to modify things in place, but I dont think that is a priority now.

Vexu (approved):

andrewrk:
Nice work. Before we do any more with error sets, we need to come up with some more advanced test cases that will exercise the lazy value system (which does not yet exist in stage2) because inferred error sets are heavily reliant on it. I expect the code from this pull request will need to get reworked when the lazy value system is introduced.

PR #7784, opened 2021-01-15
complete {Z} deprecation in std.fmt.formatIntValue; formatZigEscapes doesn't exist any more

kivikakk:
This code is broken; give the proper compile error.


marler8997:
ah missed one, thanks for the fix

PR #7785, opened 2021-01-15
organize std lib concurrency primitives and add RwLock

andrewrk:
* move concurrency primitives that always operate on kernel threads to
   the std.Thread namespace
 * remove std.SpinLock. Nobody should use this in a non-freestanding
   environment; the other primitives are always preferable. In
   freestanding, it will be necessary to put custom spin logic in there,
   so there are no use cases for a std lib version.
 * move some std lib files to the top level fields convention
 * add std.Thread.spinLoopHint
 * add std.Thread.Condition
 * add std.Thread.Semaphore
 * new implementation of std.Thread.Mutex for Windows and non-pthreads Linux
 * add std.Thread.RwLock

Implementations provided by @kprotty


BinaryWarlock:
> remove std.SpinLock. Nobody should use this in a non-freestanding
environment; the other primitives are always preferable. In
freestanding, it will be necessary to put custom spin logic in there,
so there are no use cases for a std lib version.

Doesn't AtomicMutex fall back to a spinlock there? And is it possible to change the mutex impl without changing the stdlib code? I don't see any way to do that (e.g. some `root.os` shenanigans). Same for Condition, etc.

kprotty:
@BinaryWarlock It should fall back to effectively a spin lock for platforms that don't support thread blocking primitives (no OS locks & no pthread), which is relatively rare for non-freestanding builds.

For freestanding or globally changing the implementation, Id like to propose a [ParkingLot](https://webkit.org/blog/6161/locking-in-webkit/) based solution, but one which [generalizes over the blocking primitives](https://github.com/kprotty/zap/blob/runtime4/src/sync/generic/ParkingLot.zig#L358). This would allow you to plug in how the synchronization primitive actually does the "blocking part" meaning the same algorithm can be used for Async mutexes, OS mutexes, and transitively Freestanding mutexes (instead of a global configuration option):
```zig
pub const Thread = struct {
    const OsBlockingImpl = ...;
    pub const Mutex = CoreMutex(OsBlockingImpl);
};

pub const event = struct {
    const AsyncBlockingImpl = ...;
    pub const Mutex = CoreMutex(AsyncBlockingImpl);
};

pub const CustomMutex = CoreMutex(struct {
    fn wait(self: *@This()) void {
        freestanding_thing.block();
    }
    
    fn notify(self: *@This()) void {
        freestanding_thing.unblock();
    }
});
```

That could be in this or another PR though, as it would keep the same API and semantics for the primitives made here.

andrewrk:
> That could be in this or another PR though, as it would keep the same API and semantics for the primitives made here.

Another PR please!

andrewrk:
> And is it possible to change the mutex impl without changing the stdlib code?

That's an idea worth discussing I think but doesn't need to block this PR.

kprotty:
@andrewrk Im on it

PR #7788, opened 2021-01-16
translate-c: add <assert.h> support

ehaas:
Implement __builtin_expect so C code that uses assert() can be translated.

### DIFF
--- a/lib/std/c/builtins.zig
+++ b/lib/std/c/builtins.zig
@@ -182,3 +182,9 @@ pub fn __builtin_memcpy(
     @memcpy(dst_cast, src_cast, len);
     return dst;
 }
+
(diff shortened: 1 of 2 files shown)


andrewrk (approved):

ehaas:
Looks like assert on windows is doing something else...I can investigate it over the weekend.

ehaas:
`assert()` on windows uses wide characters so I'll need to add support for that before this will work. Will do as separate PRs.

ehaas:
Currently blocked on #7989 7989

ehaas:
The remaining problem with this branch (Linux assert.h) should be fixed once the `translate-c-ast` branch is merged, due to improved if/else handling.

PR #7792, opened 2021-01-16
std: Add Priority Dequeue

zanderxyz:
Adds a double-ended priority queue to the standard library. This offers the same API as the existing `std.PriorityQueue`, except instead of offering O(log n) access to the smallest item only, it offers this for both the smallest and largest items. It is implemented as a [min-max heap](https://en.wikipedia.org/wiki/Min-max_heap).

I did [some very basic benchmarking](https://github.com/zanderxyz/min-max-heap.zig), and despite having strictly more functionality, it seems to outperform the standard queue in many situations, particularly where the item priorities are bounded to a smaller range. I expect there may be some optimisations available for the existing queue, although the min-max heap also benefits from the fact that there are half as many layers that need to be visited when sifting up or down.

### DIFF
--- a/lib/std/priority_dequeue.zig
+++ b/lib/std/priority_dequeue.zig
@@ -7,6 +7,7 @@ const std = @import("std.zig");
 const Allocator = std.mem.Allocator;
 const assert = std.debug.assert;
 const warn = std.debug.warn;
+const Order = std.math.Order;
 const testing = std.testing;
 const expect = testing.expect;
 const expectEqual = testing.expectEqual;
@@ -20,29 +21,24 @@ pub fn PriorityDequeue(comptime T: type) type {
         items: []T,
         len: usize,
         allocator: *Allocator,
-        lessThanFn: fn (a: T, b: T) bool,
+        compareFn: fn (a: T, b: T) Order,
 
-        /// Initialize and return a new dequeue. Provide `lessThanFn`
-        /// that returns `true` when its first argument should
-        /// get min-popped before its second argument. For example,
-        /// to make `popMin` return the minimum value, provide
+        /// Initialize and return a new priority dequeue. Provide `compareFn`
+        /// that returns `Order.lt` when its first argument should
+        /// get min-popped before its second argument, `Order.eq` if the
+        /// arguments are of equal priority, or `Order.gt` if the second
+        /// argument should be min-popped first. Popping the max element works
+        /// in reverse. For example, to make `popMin` return the smallest
+        /// number, provide
         ///
-        /// `fn lessThanFn(a: T, b: T) bool { return a < b; }`
-        pub fn init(allocator: *Allocator, lessThanFn: fn (T, T) bool) Self {
+        /// `fn lessThan(a: T, b: T) Order { return std.math.order(a, b); }`
+        pub fn init(allocator: *Allocator, compareFn: fn (T, T) Order) Self {
             return Self{
                 .items = &[_]T{},
                 .len = 0,
                 .allocator = allocator,
-                .lessThanFn = lessThanFn,
+                .compareFn = compareFn,
             };
-        }
-
-        fn lessThan(self: Self, a: T, b: T) bool {
-            return self.lessThanFn(a, b);
-        }
-
-        fn greaterThan(self: Self, a: T, b: T) bool {
-            return self.lessThanFn(b, a);
         }
 
         /// Free memory used by the dequeue.
@@ -100,7 +96,8 @@ pub fn PriorityDequeue(comptime T: type) type {
             const parent = self.items[parent_index];
 
             const min_layer = self.nextIsMinLayer();
-            if ((min_layer and self.greaterThan(child, parent)) or (!min_layer and self.lessThan(child, parent))) {
+            const order = self.compareFn(child, parent);
+            if ((min_layer and order == .gt) or (!min_layer and order == .lt)) {
                 // We must swap the item with it's parent if it is on the "wrong" layer
                 self.items[parent_index] = child;
                 self.items[child_index] = parent;
@@ -118,21 +115,21 @@ pub fn PriorityDequeue(comptime T: type) type {
 
         fn siftUp(self: *Self, start: StartIndexAndLayer) void {
             if (start.min_layer) {
-                doSiftUp(self, start.index, lessThan);
+                doSiftUp(self, start.index, .lt);
             } else {
-                doSiftUp(self, start.index, greaterThan);
+                doSiftUp(self, start.index, .gt);
             }
         }
 
-        fn doSiftUp(self: *Self, start_index: usize, compare: fn (Self, T, T) bool) void {
+        fn doSiftUp(self: *Self, start_index: usize, target_order: Order) void {
             var child_index = start_index;
             while (child_index > 2) {
                 var grandparent_index = grandparentIndex(child_index);
                 const child = self.items[child_index];
                 const grandparent = self.items[grandparent_index];
 
-                // If the grandparent is already better, we have gone as far as we need to
-                if (!compare(self.*, child, grandparent)) break;
+                // If the grandparent is already better or equal, we have gone as far as we need to
+                if (self.compareFn(child, grandparent) != target_order) break;
 
                 // Otherwise swap the item with it's grandparent
                 self.items[grandparent_index] = child;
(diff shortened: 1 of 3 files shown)


lib/std/priority_dequeue.zig
```diff
@@ -0,0 +1,879 @@
+const std = @import("std");
```
joachimschmidt557, line 1: Small nitpick: All files in the standard library have a license identifier at the top. It would make sense to add this here as well.

zanderxyz: Ah yes, I meant to add this. Thanks for pointing it out, will fix.

lib/std/priority_dequeue.zig
```diff
@@ -0,0 +1,879 @@
...
+        allocator: *Allocator,
+        lessThanFn: fn (a: T, b: T) bool,
+
+        /// Initialize and return a new dequeue. Provide `lessThanFn`
+        /// that returns `true` when its first argument should
+        /// get min-popped before its second argument. For example,
+        /// to make `popMin` return the minimum value, provide
+        ///
+        /// `fn lessThanFn(a: T, b: T) bool { return a < b; }`
+        pub fn init(allocator: *Allocator, lessThanFn: fn (T, T) bool) Self {
```
LemonBoy, line 27: Design question, what about making the comparator a `fn (T, T) math.Order` ?

lib/std/priority_dequeue.zig
```diff
@@ -0,0 +1,879 @@
...
+            self.len += 1;
+        }
+
+        fn isMinLayer(index: usize) bool {
+            // In the min-max heap structure:
+            // The first element is on a min layer;
+            // next two are on a max layer;
+            // next four are on a min layer, and so on.
+            const leading_zeros = @clz(usize, index + 1);
+            const highest_set_bit = 63 - leading_zeros;
```
LemonBoy, line 80: `63` should be `@bitSizeOf(usize) - 1` ?

lib/std/priority_dequeue.zig
```diff
@@ -0,0 +1,879 @@
...
+        pub fn removeMax(self: *Self) T {
+            return self.removeIndex(self.maxIndex().?);
+        }
+
+        /// Remove and return element at index. Indices are in the
+        /// same order as iterator, which is not necessarily priority
+        /// order.
+        pub fn removeIndex(self: *Self, index: usize) T {
+            const item = self.items[index];
+            const last = self.items[self.len - 1];
```
LemonBoy, line 191: Missing guard against `self.len==0`? There's also `mem.swap` to clean up this code a bit.

lib/std/priority_dequeue.zig
```diff
@@ -0,0 +1,879 @@
...
+        /// allocated with `allocator`.
+        /// De-initialize with `deinit`.
+        pub fn fromOwnedSlice(allocator: *Allocator, lessThanFn: fn (T, T) bool, items: []T) Self {
+            var dequeue = Self{
+                .items = items,
+                .len = items.len,
+                .allocator = allocator,
+                .lessThanFn = lessThanFn,
+            };
+            const half = (dequeue.len >> 1) - 1;
```
LemonBoy, line 353: Underflow if `dequeue` has 1 or 0 elements.

lib/std/priority_dequeue.zig
```diff
@@ -0,0 +1,879 @@
...
+        }
+
+        pub fn shrink(self: *Self, new_len: usize) void {
+            // TODO take advantage of the new realloc semantics
+            assert(new_len <= self.len);
+            self.len = new_len;
+        }
+
+        pub fn update(self: *Self, elem: T, new_elem: T) !void {
+            var old_index: usize = std.mem.indexOfScalar(T, self.items, elem) orelse return error.ElementNotFound;
```
LemonBoy, line 384: `self.items` or `self.items[0..self.len]` ?

lib/std/priority_dequeue.zig
zanderxyz: Thanks for the great review!

I agree that this would be a better API; I was not aware of `math.Order`. If I change this here I think I should change `std.PriorityQueue` as well. This would be a breaking change, but presumably this is common for Zig right now. I will make this change unless other reviewers disagree.

There might be an argument for making similar changes to `std.sort` for consistency, as some functions in there use `bool`, and some use `math.Order`. For example:

```
pub fn binarySearch(
    comptime T: type,
    key: T,
    items: []const T,
    context: anytype,
    comptime compareFn: fn (context: @TypeOf(context), lhs: T, rhs: T) math.Order,
) 
```

```
pub fn sort(
    comptime T: type,
    items: []T,
    context: anytype,
    comptime lessThan: fn (context: @TypeOf(context), lhs: T, rhs: T) bool,
)
```

However that seems like a bigger change and is probably out of scope here.

lib/std/priority_dequeue.zig
zanderxyz: Yes, thank you. I meant to find the right function for this.

lib/std/priority_dequeue.zig
zanderxyz: Thanks, fixed here and in `std.PriorityQueue` (and tests added).

lib/std/priority_dequeue.zig
zanderxyz: Good spot! Updated to `self.items[0..self.len - 1]` (here and in `std.PriorityQueue`).

lib/std/priority_dequeue.zig
zanderxyz: Added the guard. Are you suggesting `mem.swap` in this function, or lower down to replace segments that look like this?

```
self.items[grandparent_index] = child;
self.items[child_index] = grandparent;
```

In this particular function I'm not sure it adds anything, as we only have to move one item. I can see the argument for replacing these other blocks though.

lib/std/priority_dequeue.zig
```diff
@@ -0,0 +1,907 @@
...
+            self.items = try self.allocator.realloc(self.items, better_capacity);
+        }
+
+        pub fn resize(self: *Self, new_len: usize) !void {
+            try self.ensureCapacity(new_len);
+            self.len = new_len;
+        }
+
+        pub fn shrink(self: *Self, new_len: usize) void {
+            // TODO take advantage of the new realloc semantics
```
zanderxyz, line 386: I copied this straight from `std.PriorityQueue`. It's not a great TODO, I am not sure exactly what it's referring to. It would be nice to fix this in both places while I am touching this code, but I will need someone to point me in the right direction.

lib/std/priority_dequeue.zig
LemonBoy: That was more of a general reminder, whenever you need to swap two elements and cannot be arsed to write the usual boilerplate `mem.swap` comes in handy

lib/std/priority_dequeue.zig
LemonBoy: Check out `Allocator.realloc`, the reallocation can fail whenever the backing allocator deems it unnecessary (while `shrink` is guaranteed to get you a smaller piece of memory).
This scheme is already in use in `ArrayList.shrinkAndFree`.

zanderxyz: Thanks. I have replaced `shrink` with `shrinkRetainingCapacity` and `shrinkAndFree`. In both cases I have restricted the new length to be between the length and the capacity, as shrinking to a smaller size would invalidate the heap property. An alternative would be to remove the largest `n` items instead, thus retaining the property - I think what I have done is probably the safest default behaviour but I can see arguments both ways. This forces the user to decide which elements they want to remove if they want to shrink the queue below its current size.

lib/std/priority_queue.zig
```diff
@@ -183,19 +187,34 @@ pub fn PriorityQueue(comptime T: type) type {
...
+            assert(new_len >= self.len);
+
+            self.items = self.allocator.realloc(self.items[0..], new_len) catch |e| switch (e) {
+                error.OutOfMemory => { // no problem, capacity is still correct then.
+                    self.items.len = new_len;
+                    return;
+                },
+            };
             self.len = new_len;
         }
```
zanderxyz, line 204: I removed `resize` because I don't think this is the right behaviour for a heap. This is the same logic as in `std.ArrayList`, but although extending with uninitialised elements makes sense for a list, adding these to the end of a heap invalidates the heap property and it is no longer guaranteed to work correctly.

andrewrk:
This looks really nice, thank you! Great test coverage, responses to the reviews, and the benchmark is also appreciated.

Happy to merge this. Just a warning though, since nothing depends on this, it may sadly have to move to https://github.com/ziglang/std-lib-orphanage/ (temporarily?) until we can deliver the self-hosted compiler with better memory management.

zanderxyz:
> This looks really nice, thank you! Great test coverage, responses to the reviews, and the benchmark is also appreciated.
> 
> 
> 
> Happy to merge this. Just a warning though, since nothing depends on this, it may sadly have to move to https://github.com/ziglang/std-lib-orphanage/ (temporarily?) until we can deliver the self-hosted compiler with better memory management.

That's completely understood, thank you. I really like Zig, great work with the language so far!

I would like to do some more standard library work once the self-hosted compiler is further along and things are slightly more stable. I'll try to keep an eye on the issues tracker.

PR #7800, opened 2021-01-17
Rework synchronization primitives

kprotty:
The current ones feel a bit makeshift (not destroying pthread objects, aren't extensible to async, etc.). This tries to fix these by providing a compile time configurable implementation of [WTF::ParkingLot](https://webkit.org/blog/6161/locking-in-webkit/) which is an extensible, futex-like, blocking primitive which can be used to implement almost any synchronization primitive. 

The idea is for the synchronization type to be abstract of how it blocks the caller. That aspect can be provided at comptime to make the primitive thread blocking via OS calls, async frame blocking via suspend/resume, or custom to be used in a freestanding setting.

Other side effects:
- centralizes the ResetEvents (StaticResetEvent, ResetEvent) into one
- provides wrappers over atomic operations to emulate C11 atomics (with bit-{set|reset|toggle})
- adds eventual fairness to serialization primitives like Mutex and hopefully RwLock
- makes the primitives smaller (`u8` - `usize` possible)

Eventual goals:
- migrate `std.event` stuff to use the abstract primitives with an async backend instead of a thread backend
- add `std.sync.Channel` which gives a MPMC api similar to `std.fifo.LinearFifo`
- add more docs to the implementation since the backends may use some undocumented functions
- copy golang and use thread primitives directly (`lwp` on netbsd, `umtx_op` on bsd, etc.) instead of pthread.


kprotty:
This is turning out to be a lot bigger than I thought. Notably testing all of these synchronization primitives & their backends + other misc. additions.

An idea even farther into the future is to model the synchronization primitives as non-blocking, cancellable requests then implement blocking on top of those. It would ideally allow one to implement multiplexing (e.g. efficient join/select - for threading and async) with them. Although that will complicate the internals more so than they already are.

kprotty:
Started on a rewrite in [futex](https://github.com/kprotty/zig/tree/futex) branch use a futex-based backend instead of a parking_lot one. Will be merging it into this branch eventually. 

The pro is that it uses futex directly on platforms that support it (currently windows/linux/darwin/freebsd - zigs primary targets) instead of going through a userspace impl of pakring_lot. The con is that on platforms without native futex, it simulates it by using a userspace implementation.

kprotty:
In regards to the status of this, the [futex](https://github.com/kprotty/zig/tree/futex) is very close to being ready for review. The remaining things it needs are tests for RwLock and for the Futex themselves. I haven't been able to get around to it recently, but if anyone wants to try writing them I'm happy to accept PRs

kprotty:
Since there haven't been any updated in a while. I decided to create a new PR using the futex branch mentioned before: #8084

PR #7801, opened 2021-01-17
stage2: use %type not @type for libc stubs

jrmaingo:
Looks like some comments were missed last time these comment symbols were changed https://github.com/ziglang/zig/commit/343249efd845c30fbdcb18a37ebb356ab19da742

This change resolves an error I was having with zig-bootstrap (similar to https://github.com/ziglang/zig-bootstrap/issues/47)

I was able to build the compiler successfully after these changes.

I also added in a comment change for FileAlreadyExists -> PathAlreadyExists

### DIFF
--- a/lib/libc/musl/libc.s
+++ b/lib/libc/musl/libc.s
@@ -1,105 +1,105 @@
 .bss
 .weak ___environ
-.type ___environ, @object;
+.type ___environ, %object;
 ___environ:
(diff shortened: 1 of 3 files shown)


ifreund (approved):
My mistake, thanks for the fix!

andrewrk:
Thanks!

PR #7809, opened 2021-01-18
translate-c: Add Wide, UTF-16, and UTF-32 character literals

ehaas:
Add support for L'<wchar_t>', u'<char16_t>', and U'<char32_t>'. Currently
this just translates wide char literals to \u{NNNNNN} escape codes
(e.g. U'💯' -> '\u{1f4af}')

Another approach would be to emit UTF-8 encoded character literals
directly, but in my opinion this approaches Unicode-complete because it
would require knowledge of which Unicode codepoints have graphical
representations for the emitted source to be readable.

We could also just emit integer literals, but the current method makes
it clear that we have translated a wide character literal and not just
an integer constant.

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -2942,9 +2942,9 @@ fn transCharLiteral(
     suppress_as: SuppressCast,
 ) TransError!*ast.Node {
     const kind = stmt.getKind();
+    const val = stmt.getValue();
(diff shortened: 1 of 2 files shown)


andrewrk:
Your reasoning makes sense to me. Nice work!

PR #7814, opened 2021-01-18
std: Fixed pipe2 fallback

### LINKED ISSUE #7760: child_process: pipe2 does not set O_CLOEXEC for mac
O_CLOEXEC should be set using F_SETFD.  Perhaps pipe2 should call setSockFlags() for proper emulation of O_CLOEXEC and O_NONBLOCK?

LemonBoy:
Use both F_SETFD and F_SETFL depending on what flag we're setting.

Closes #7760

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -3780,7 +3780,7 @@ pub fn pipe2(flags: u32) PipeError![2]fd_t {
     // F_SETFD.
     if (flags & O_CLOEXEC != 0) {
         for (fds) |fd| {
-            switch (errno(system.fcntl(fd, F_SETFD, FD_CLOEXEC))) {
+            switch (errno(system.fcntl(fd, F_SETFD, @as(u32, FD_CLOEXEC)))) {
                 0 => {},
                 EINVAL => unreachable, // Invalid flags
                 EBADF => unreachable, // Always a race condition
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -3755,31 +3755,54 @@ pub fn pipe() PipeError![2]fd_t {
 }
 
 pub fn pipe2(flags: u32) PipeError![2]fd_t {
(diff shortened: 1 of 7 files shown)


lib/std/os.zig
```diff
@@ -3755,31 +3755,54 @@ pub fn pipe() PipeError![2]fd_t {
...
-        var fds: [2]fd_t = try pipe();
-        if (flags == 0) return fds;
-        errdefer {
-            close(fds[0]);
-            close(fds[1]);
-        }
-        for (fds) |fd| switch (errno(system.fcntl(fd, F_SETFL, flags))) {
-            0 => {},
+    if (flags == 0)
+        return pipe();
```
daurnimator, line 3759: This seems a little weird: could you put this branch below the `pipe2` existence check so that `pipe2` always uses the `pipe2` syscall when available?

LemonBoy: No. `pipe2` with no flags is identical to `pipe` and that's by specification. Pretty much every pipe2 libc wrapper call pipe when their argument is zero (eh. Musl implementation) so we peel an indirection layer by doing so here.

daurnimator: > Pretty much every pipe2 libc wrapper call pipe when their argument is zero

Do they? It appears to be the opposite: https://github.com/bminor/glibc/blob/21c3f4b5368686ade28d90d8c7d79c4c95c72c1b/sysdeps/unix/sysv/linux/generic/pipe.c#L29

---

I was also thinking of if I e.g. wanted to avoid the weird mess of `syscall_pipe` on MIPS

LemonBoy: > Do they?

Not _every_, just _some_ https://git.musl-libc.org/cgit/musl/tree/src/unistd/pipe2.c

> wanted to avoid the weird mess of syscall_pipe on MIPS

lol weird mess

andrewrk (approved):

PR #7818, opened 2021-01-18
Macho more cleanup

kubkon:
This PR cleans up how we write indirect symbol table into the final artifact. It also forces rewriting of code signature padding at every update so that we take into account possible section relocations and expansion of the last preceding section, e.g., the string table.

This PR also tweaks the logic responsible for managing debug lines in `DebugSymbols`. In particular, in case we update the same function, we'd previously incorrectly create a cycle adding pointer to the same `SrcFn` to itself.  This would manifest itself when running self-hosted compiler in `watch` mode with a segfault in this line: [src/link/MachO/DebugSymbols.zig#L1142](https://github.com/ziglang/zig/blob/c3dadfa95b01d140460eb3d3d47d13859302f298/src/link/MachO/DebugSymbols.zig#L1142). Since we've just created a cycle pointing `SrcFn` inside `Decl` to itself, the offsets will cancel out, leaving out us trying to save negative length into a `u32` value. A repro on macOS would be the following:

* First incremental linking update

```zig
extern "c" exit(usize) noreturn;

export fn _start() noreturn {
    exit(0);
}
```

* Second incremental linking update

```zig
extern "c" fn exit(usize) noreturn;

export fn _start() noreturn {
    print();
    exit(0);
}

fn print() void {
    _ = write(1, @ptrToInt("Hey\n"), 4);
}
```

On the second incremental update, we'd get the segfault. Since I'm not sure whether my fix is correct, I'm pinging you @andrewrk to have a look and lemme know if this makes sense. If so, I guess I should also update the matching lines in the `Elf` linker.

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -1617,8 +1617,10 @@ pub fn analyzeContainer(self: *Module, container_scope: *Scope.Container) !void 
                             // in `Decl` to notice that the line number did not change.
                             self.comp.work_queue.writeItemAssumeCapacity(.{ .update_line_number = decl });
                         },
-                        .macho => {
-                            // TODO Implement for MachO
+                        .macho => if (decl.fn_link.macho.len != 0) {
+                            // TODO Look into detecting when this would be unnecessary by storing enough state
+                            // in `Decl` to notice that the line number did not change.
+                            self.comp.work_queue.writeItemAssumeCapacity(.{ .update_line_number = decl });
                         },
                         .c, .wasm => {},
                     }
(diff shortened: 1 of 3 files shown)


andrewrk (approved):

andrewrk:
> On the second incremental update, we'd get the segfault. Since I'm not sure whether my fix is correct, I'm pinging you @andrewrk to have a look and lemme know if this makes sense. If so, I guess I should also update the matching lines in the `Elf` linker.

Oops I missed this before. I'll try to repro the issue with ELF and see what I can find.

kubkon:
> > On the second incremental update, we'd get the segfault. Since I'm not sure whether my fix is correct, I'm pinging you @andrewrk to have a look and lemme know if this makes sense. If so, I guess I should also update the matching lines in the `Elf` linker.
> 
> Oops I missed this before. I'll try to repro the issue with ELF and see what I can find.

Please do, and lemme know what you find. I think the trick is to expand a `TextBlock` by inserting a new decl just after the prologue of the expanded function, so from:

```zig
export fn _start() noreturn {
    exit();
}
```

to

```zig
export fn _start() noreturn {
    print();
    exit();
}
```

So prepending rather than appending as we have it in the tests.

kubkon:
In the meantime, I'll go ahead and merge this PR, but if you discover that you can't replicate the issue on Elf, I'll revert the few lines in MachO's `DebugSymbols` and dig deeper.

PR #7830, opened 2021-01-19
stage2 elf: fix potential overflow when reusing TextBlock

kubkon:
If `TextBlock` does not have a successor, its capacity in the worst
case will equal to almost the entire `u64` resolution, meaning
scaling it by `alloc_num / alloc_den` will cause an overflow and a
panic in debug/release-safe mode.

This fix is backported from the `MachO` change in c1768d9.

### DIFF
--- a/src/link/Elf.zig
+++ b/src/link/Elf.zig
@@ -9,6 +9,7 @@ const elf = std.elf;
 const log = std.log.scoped(.link);
 const DW = std.dwarf;
 const leb128 = std.leb;
+const math = std.math;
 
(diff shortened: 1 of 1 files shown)


andrewrk:
Can you add a test case that this fixes? The main purpose of supporting updates in test cases is so that we can add coverage for this kind of thing!

kubkon:
> Can you add a test case that this fixes? The main purpose of supporting updates in test cases is so that we can add coverage for this kind of thing!

Good call, will do!

PR #7833, opened 2021-01-19
zig cc: Support reading input from stdin

### LINKED ISSUE #6271: zig cc can't read input from stdin
Meson uses `CXX -E -dM -` to read all defines produced by a compiler.

This fails with `zig cc` and `zig c++` because the `-` flag is not recognized:
```
$ zig cc -target x86_64-windows-gnu -E -dM -
Unknown Clang option: '-'
```

How can we handle it?

Ristovski:
This fixes #6271, which allows using `zig cc` with meson.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -3043,7 +3043,7 @@ pub const ClangArgIterator = struct {
             arg = mem.span(self.argv[self.next_index]);
             self.incrementArgIndex();
         }
-        if (!mem.startsWith(u8, arg, "-")) {
+        if (mem.eql(u8, arg, "-") or !mem.startsWith(u8, arg, "-")) {
             self.zig_equivalent = .positional;
             self.only_arg = arg;
             return;


andrewrk (approved):
Thanks!

harryhaaren:
Hey @Ristovski, can I ask how you made Meson use `zig cc` as the compiler? Any time I try this, it results in either Meson complaining.. Using Zig master build `0.8.0-dev.1039+bea791b63`.

I've tried the standard `CC=zig meson build` approach, Meson fails on "Unknown Compiler Zig"
Using `CC="zig cc" meson build` results in Meson failing on "Compiler zig cc cannot compile programs".

Am I missing something?

harryhaaren:
OK - with help from folks in `#zig`, identifed root cause (Zig's lib/ dir was not available to the Zig library, resulting in compilation failures, and Meson reporting "cannot compile programs").

Workarounds include
- Using absolute path to Zig binary (leaving Zig's lib/* stuff in place)
- Moving Zig to e.g. `/usr/local/bin` and moving zig's `lib/` contents to /usr/local/lib`.

There's no "install" instructions with the Zig master download, perhaps worth including? Anyway, working here now!

PR #7839, opened 2021-01-20
std.pdb: Don't read past end of MSF Stream

camconn:
This PR ensures that `MsfStream.read` doesn't go beyond its bounds and return incorrect data. I'm unsure how to fake a `std.fs.File` for writing a unit test, but I'll happily add one to this PR if I can get some guidance. What would be the best way to do that?

And on an unrelated note:  
I've been working on a project that modifies `std.pdb` and this is the only trivial change I've found so far. I can open another PR for additions such as a file hash table implementation and de-serializer, but I'm unsure if those would be appropriate for the standard library. Would Zig accept these additions?


andrewrk:
Thanks for the patch! Regarding a further PR, I think the std lib PDB code is going to be only exactly what we need in order to implement stack traces, and nothing more. However if you think there would be some benefit, feel free to open it, and we can take a look and discuss.

daurnimator:
> what we need in order to implement stack traces, and nothing more.

Wouldn't it be useful to add structures relevant to generating PDB files?

Issue #7194, opened 2020-11-22
A plan for std.net


LemonBoy:
In #7124 a few people chimed in to ask for something more in `std.net`.
This ticket is meant to organize all the requests/ideas and hopefully turn them into code by the time `0.8` rolls around.

Now that network streams (`net.Stream`) are decoupled from files we can finally start adding more socket-specific APIs and reduce the amount of footguns linked to the use of sockets (did I really call `stat` on a socket? sure I did).

The `Stream` type is a thi(iiiiiiiii)n layer over the OS socket type, there's **no** guarantee socket handles are valid file handles and vice-versa (eg. that's the case for Winsocket handles) and there's no knowledge of the transport stream type (TCP, UDP, Unix...).
The latter is still up for discussion but the two alternatives I thought of, having separate `*Stream` type or having a tagged union type, are IMO not so nice from an API ergonomic point of view.
The `Stream` type is also stateless, it has no knowledge of the connection state, to keep the bookkeeping overhead as small as possible and avoid getting that piece of info out of sync with the kernel.

Right now the idea is to add:
* More constructors (the `Host` variant takes a string, the suffix-less one a address-port pair):
  - `connectTcpHost` (replaces `tcpConnectToHost`)
  - `connectTcp` (replaces `tcpConnectToAddress`)
  - `connectUdp` (complements `connectTcp`)
     - Do we need a `connectUdpHost` ?
  - `connectUnixHost` (complements `connectTcpHost`)
  - Support connection timeouts in blocking mode! Each 
* More property getters/setters:
  - `nodelay`
  - `keepalive`
  - `linger`
  - `nonblocking` (?)
  - Read/write timeouts (`SO_RCVTIMEO` and `SO_SNDTIMEO`)
* A way to query the local address (`getsockname`)
  - `fn getLocalAddress(self) Address`
* A way to query the remote address (`socket_getpeername`)
  - `fn getRemoteAddress(self) Address`
* A nice wrapper around `shutdown()`
  - `fn shutdown(enum { read, write, both })`

Open questions:
- UDP streams want `recvmsg`/`sendmsg` and/or `recvfrom`/`sendto`)
- Some operations make sense only for some socket types (see point above), can we really get away with a single stream type?
- Peek operation. Is that useful?

cc @Aransentin @frmdstryr @MasterQ32

MasterQ32:
> UDP streams want recvmsg/sendmsg and/or recvfrom/sendto)

[...]

PR #7844, opened 2021-01-20
std: Revamp the net API

LemonBoy:
Pushing for some early feedback, this PR is meant to implement the plan outlined in #7194 

cc @MasterQ32


MasterQ32 (changes requested):
If we're doing breaking changes anyway, i would also rename `Address` to `EndPoint` as what people usually associate with a network address is `127.0.0.1` or `::1` without a port number.

Apart from that it looks pretty nice!

aconbere:
After spending a couple of days digging into network programming in Zig I have to say that this change would have had a huge impact in the speed I was able to get up and running. Big +1 from me.

YerinAlexey (approved):
Looks good, I like the new style with `TcpListener` separate from `TcpClient` and same for UDP.

lib/std/net.zig
```diff
@@ -12,12 +12,55 @@ const os = std.os;
...
 // Windows 10 added support for unix sockets in build 17063, redstone 4 is the
 // first release to support them.
 pub const has_unix_sockets = @hasDecl(os, "sockaddr_un") and
     (builtin.os.tag != .windows or
     std.Target.current.os.version_range.windows.isAtLeast(.win10_rs4) orelse false);
 
+// Platform-independent socket descriptor.
+// Note that on some platforms this may not be interchangeable with a regular
+// files descriptor.
+pub const Socket = extern struct {
```
marler8997, line 31: This wrapper object around `socket_t` brings up some questions.

* When would someone use `Socket` vs `os.socket_t`?
* How would someone coerce an array of `Socket` into an array of handles or file descriptors on windows/posix?  Can the user make the assumption that `Socket` is always the same alignment/size as a handle/file descriptor on those platforms?  In which case they should use a cast?
* When representing an extern struct that contains a socket handle, should `Socket` or `socket_t` be used?
* Should Zig libraries that provide functions that accept sockets be accepting `socket_t` or `Socket`?

The answers to some of these questions will affect the answers to others.  Having multiple type for a socket seems problematic to me unless there are some good answers to these questions.

In my experience in working with networking libraries in various languages, the easiest ones to work with seem to be those that don't try to hide the fact that a socket is just a handle and/or file descriptor.  Creating a wrapper around socket can make simple examples look "nice" but seems to create a "split" in APIs where each version of a socket is used either out of necessity or happenstance.  Sometimes the fact that a socket is a handle/file descriptor is required knowledge to store/pass them around correctly and reflecting this equivalence in the type system seems to make things simpler.

YerinAlexey: Agree, this `Socket` wrapper is a bit useless since `os.socket_t` already exists

lib/std/net.zig
```diff
@@ -808,905 +793,600 @@ pub fn getAddressList(allocator: *mem.Allocator, name: []const u8, port: u16) !*
...
+        );
+        return value != 0;
+    }
 
-    try resMSendRc(qp[0..nq], ap[0..nq], apbuf[0..nq], rc);
+    pub fn setReadTimeout(self: TcpClient, timeout_ms: u32) !void {
+        try os.setsockopt(
+            self.socket.handle,
+            os.SOL_SOCKET,
+            os.SO_RCVTIMEO,
```
daurnimator, line 961: I don't think works as expected, IIRC `SO_RCVTIMEO` doesn't work with polling.

andrewrk:
This looks like a lot of great work, unfortunately still a draft. I hope you or someone else can get this to a mergeable point so we don't lose all the effort.

aconbere:
> This looks like a lot of great work, unfortunately still a draft. I hope you or someone else can get this to a mergeable point so we don't lose all the effort.

Hey Andrew,

For future efforts, do you have a set of criteria that you're thinking of when considering the mergability? I assume there's likely been a good deal of out of band conversations about this effort, but for those of us following along here I think it would be really useful to have a sense of what would need to be done :)

andrewrk:
This PR was marked as a draft by the author, which means "Not ready for review". It's not my criteria, it's the author themselves saying it is not ready.

PR #7846, opened 2021-01-21
stage1: don't filter test blocks with empty label

LemonBoy:
The common pattern of including a file containing all the tests in a
empty-label test block breaks down when using --test-filter.

### DIFF
--- a/lib/std/Thread.zig
+++ b/lib/std/Thread.zig
@@ -553,7 +553,7 @@ pub fn getCurrentThreadId() u64 {
     }
 }
 
-test "" {
+test {
     if (!builtin.single_threaded) {
         std.testing.refAllDecls(@This());
     }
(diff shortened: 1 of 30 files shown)


andrewrk:
Sounds good, although perhaps the language should recognize this usage pattern and allow omitting the string literal in the syntax?

mikdusan:
```zig
usingtestspace {
}
usingtestspace "alt" {
}
```

LemonBoy:
> Sounds good, although perhaps the language should recognize this usage pattern and allow omitting the string literal in the syntax?

Wouldn't that go against the "there's only one obvious way to do X" mantra?
`test { foo }` and `test "" { foo }` would be identical, unless we start limiting what's inside the test block... but that's not cool and a huge can of worms we can leave on the shelf IMO.

andrewrk:
I'm suggesting that if you could omit the string literal, it would not be eligible for test filtering, and there would be no reason to ever do `test "" { foo }` (only omitting the string literal would make it exempt from test name filtering)

PR #7847, opened 2021-01-21
stage2: rework astgen result locations

andrewrk:
Motivating test case:

```zig
export fn _start() noreturn {
    var x: u64 = 1;
    var y: u32 = 2;
    var thing: bool = true;
    const result = if (thing) x else y;
    exit(result);
}
```

The main idea here is for astgen to output ideal ZIR depending on
whether or not the sub-expressions of a block consume the result
location. Here, neither `x` nor `y` consume the result location of the
conditional expression block, and so the ZIR should communicate the
result of the condbr using break instructions, not with the result
location pointer.

Additionally, the local const should not require an allocation, since the expression never ends up needing a result location.

With this commit, this is accomplished:

```
fn_body _start { // unanalyzed
  %0 = dbg_stmt()
  %1 = const(TypedValue{ .ty = type, .val = type})
  %2 = const(TypedValue{ .ty = type, .val = u64})
  %3 = as(%1, %2)
  %4 = alloc_mut(%3)
  %5 = const(TypedValue{ .ty = comptime_int, .val = 1})
  %6 = store(%4, %5)
  %7 = dbg_stmt()
  %8 = const(TypedValue{ .ty = type, .val = type})
  %9 = const(TypedValue{ .ty = type, .val = u32})
  %10 = as(%8, %9)
  %11 = alloc_mut(%10)
  %12 = const(TypedValue{ .ty = comptime_int, .val = 2})
  %13 = store(%11, %12)
  %14 = dbg_stmt()
  %15 = const(TypedValue{ .ty = type, .val = type})
  %16 = const(TypedValue{ .ty = type, .val = bool})
  %17 = as(%15, %16)
  %18 = alloc_mut(%17)
  %19 = const(TypedValue{ .ty = bool, .val = true})
  %20 = store(%18, %19)
  %21 = dbg_stmt()
<no more alloc for result here>
  %22 = block({
    %23 = const(TypedValue{ .ty = type, .val = bool})
    %24 = deref(%18)
    %25 = as(%23, %24)
    %26 = condbr(%25, {
      %27 = deref(%4)
<no more dead store here to the result alloc>
      %28 = break("label_22", %27)
    }, {
      %29 = deref(%11)
<no more dead store here to the result alloc>
      %30 = break("label_22", %29)
    })
  })
  %31 = dbg_stmt()
  %32 = decl_val(Decl(exit))
  %33 = param_type(%32, 0)
  %34 = as(%33, %22)
  %35 = call(%32, [%34], modifier=auto)
  %36 = ensure_result_used(%35)
  %37 = return_void()
} // fn_body _start
```

However if the result location gets consumed, the break instructions
change to break_void, and the result value is communicated only by the
stores, not by the break instructions. This would happen, for example, if one or both of the branches called a function which wanted a result location pointer for the return type.

Implementation:

 * The GenZIR scope that conditional branches uses now has an optional
   result location pointer field and a count of how many times the
   result location ended up being an rvalue (not consumed).
 * When rvalue() is called on a result location for a block, it
   increments this counter. After generating the branches of a block,
   astgen for the conditional branch checks this count and if it is 2
   then the store_to_block_ptr instructions are elided and it calls
   rvalue() using the block result (which will account for peer type
   resolution on the break operands).

astgen has many functions disabled until they can be reworked with these
new semantics. That will be done before merging the branch.

There are some new rules for astgen to follow regarding result locations
and what you are allowed/required to do depending on which one is passed
to expr(). See the updated doc comments of ResultLoc for details.

I also changed naming conventions of stuff in this commit, sorry about
that.

Follow-up items to do before merging:

 * sema: after a block gets its peer type resolved, insert type coercions on
   the break instruction operands.
 * sema: when if condition is comptime-known, store_to_inferred_ptr and
   store_to_block_ptr should do the store at compile-time.
 * astgen: finish reworking all the expressions


andrewrk:
Regarding the latest commit (https://github.com/ziglang/zig/pull/7847/commits/bc4d1f9034ff24f049495f564c0146f2341f2dd9) here is some example code (note that the u32 and u64 must get peer-type-resolved to u64):

```zig
export fn _start() noreturn {
    var x: u64 = 1;
    var y: u32 = 2;
    var thing: bool = true;
    const result = if (thing) x else foo();
    exit(result);
}
```

ZIR (note no dead allocations or stores):

```
fn_body _start { // unanalyzed
  %0 = dbg_stmt()
  %1 = const(TypedValue{ .ty = type, .val = type})
  %2 = const(TypedValue{ .ty = type, .val = u64})
  %3 = as(%1, %2)
  %4 = alloc_mut(%3)
  %5 = const(TypedValue{ .ty = comptime_int, .val = 1})
  %6 = store(%4, %5)
  %7 = dbg_stmt()
  %8 = const(TypedValue{ .ty = type, .val = type})
  %9 = const(TypedValue{ .ty = type, .val = u32})
  %10 = as(%8, %9)
  %11 = alloc_mut(%10)
  %12 = const(TypedValue{ .ty = comptime_int, .val = 2})
  %13 = store(%11, %12)
  %14 = dbg_stmt()
  %15 = const(TypedValue{ .ty = type, .val = type})
  %16 = const(TypedValue{ .ty = type, .val = bool})
  %17 = as(%15, %16)
  %18 = alloc_mut(%17)
  %19 = const(TypedValue{ .ty = bool, .val = true})
  %20 = store(%18, %19)
  %21 = dbg_stmt()
  %22 = block({
    %23 = const(TypedValue{ .ty = type, .val = bool})
    %24 = deref(%18)
    %25 = as(%23, %24)
    %26 = condbr(%25, {
      %27 = deref(%4)
      %28 = break("label_22", %27)
    }, {
      %29 = decl_val(Decl(foo))
      %30 = call(%29, [], modifier=auto)
      %31 = break("label_22", %30)
    })
  })
  %32 = dbg_stmt()
  %33 = decl_val(Decl(exit))
  %34 = param_type(%33, 0)
  %35 = as(%34, %22)
  %36 = call(%33, [%35], modifier=auto)
  %37 = ensure_result_used(%36)
  %38 = return_void()
} // fn_body _start
```

TZIR (note the br_block_flat used for type coercion):

```
Module.Function(name=_start):
  @0: u64 = 1;
  @1: u32 = 2;
  @2: bool = true;
  @3: fn() u32 = (function);
  @4: fn(usize) noreturn = (function);
  %0: void = dbg_stmt()
  %1: *u64 = alloc()
  %2: void = store(%1, @0)
  %3: void = dbg_stmt()
  %4: *u32 = alloc()
  %5: void = store(%4, @1)
  %6: void = dbg_stmt()
  %7: *bool = alloc()
  %8: void = store(%7, @2)
  %9: void = dbg_stmt()
  %10: u64 = block({
    %11: bool = load(%7)
    %12: noreturn = condbr(%11, {
      %13: u64 = load(%1)
      %14: noreturn = br(%10, %13)
    }, {
      %15: u32 = call(@3)
      %16: noreturn = br_block_flat(%10, {
        %17: u64 = intcast(%15)
      })
    })
  })
  %18: void = dbg_stmt()
  %19: usize = intcast(%10)
  %20: noreturn = call(@4, %19)
```

For your convenience here are the doc comments for br_block_flat:

> Same as `br` except the operand is a list of instructions to be treated as
> a flat block; that is there is only 1 break instruction from the block, and
> it is implied to be after the last instruction, and the last instruction is
> the break operand.
> This instruction exists for late-stage semantic analysis patch ups, to
> replace one br operand with multiple instructions, without moving anything else around.

PR #7848, opened 2021-01-21
[Proposal] std/mem: add a more generic eql function

ifreund:
I had to write some very ugly code recently to compare two optional, sentinel-terminated pointers in river: https://github.com/ifreund/river/blob/cd005e15f8bcb6852b35fd3d89248e72611a5d5f/river/Option.zig#L87-L90
```zig
    } else if (self.value == .string and
        // TODO: std.mem needs a good way to compare optional sentinel pointers
        ((self.value.string == null and value.string == null) or
        (self.value.string != null and value.string != null and
        std.cstr.cmp(self.value.string.?, value.string.?) != 0)))
    {
```
As I wrote in the comment, this made me feel like we needed a function to do a better job of this in `std.mem`. I've Implemented such a function in this patch as a proposal. I'm not sure if this function should replace the current `std.mem.eql()` as that would be a widely breaking change and the `anytype` based API is not as simple when working with slices. However, this would match the API of `std.meta.eql()` so perhaps simply replacing `std.mem.eql()` is worth considering.

The new function is temporarily named `eqlGeneric()`, but if we decide not to replace current `mem.eql()` the new function should have a better name. Naming suggestions welcome.

Edit: `std.meta.deepEql()` or similar would also be a reasonable place to put this function.

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -484,6 +484,96 @@ pub fn eql(comptime T: type, a: []const T, b: []const T) bool {
     return true;
 }
 
+/// Takes a pointer-to-array, sentinel-terminated pointer, slice, or C pointer
+/// and returns whether the pointed-to items are equal.
+/// Assumes C pointers are non-null and zero-terminated
+pub fn eqlGeneric(a: anytype, b: @TypeOf(a)) bool {
+    switch (@typeInfo(@TypeOf(a))) {
+        .Optional => {
+            if (a == null and b == null) return true;
+            if (a) |non_null_a| {
+                if (b) |non_null_b| {
+                    return eqlGeneric(non_null_a, non_null_b);
+                }
+            }
+            return false;
+        },
+        .Pointer => |ptr_info| switch (ptr_info.size) {
+            .One => switch (@typeInfo(ptr_info.child)) {
+                .Array => |arr_info| return eql(arr_info.child, a, b),
+                else => @compileError("invalid type given to std.mem.eqlGeneric"),
+            },
+            .Many => if (ptr_info.sentinel) |sentinel| {
+                var i: usize = 0;
+                while (a[i] == b[i] and a[i] != sentinel) i += 1;
+                return a[i] == b[i];
+            } else {
+                @compileError("invalid type given to std.mem.eqlGeneric");
+            },
+            .Slice => return eql(ptr_info.child, a, b),
+            .C => {
+                const T = [*:0]const ptr_info.child;
+                return eqlGeneric(@as(T, a), @as(T, b));
+            },
+        },
+        else => @compileError("invalid type given to std.mem.eqlGeneric"),
+    }
+}
+
+test "eqlGeneric" {
+    { // (optional) pointer-to-array
+        const a = &[_]u8{ 1, 2, 3 };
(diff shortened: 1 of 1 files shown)


ifreund:
One other feature we should consider here is easy and efficient comparison of sentinel terminated pointers and slices. The current pattern is to use for example:
```zig
mem.eql(u8, my_slice, mem.span(my_sentinel_terminated_ptr));
```
However this is inefficient as the full length of the sentinel terminated pointer is always iterated even when the slice is shorter. Ideally the comparison would look something like this:
```zig
fn better(a: []const u8, b: [*:0]const u8) bool {
    for (a) |a_elem, idx| {
        if (b[idx] == 0 or b[idx] != a_elem) return false;
    }
    return b[a.len] == 0;
}
```
[Comparison in godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(fontScale:14,j:1,lang:zig,selection:(endColumn:31,endLineNumber:10,positionColumn:31,positionLineNumber:10,selectionStartColumn:31,selectionStartLineNumber:10,startColumn:31,startLineNumber:10),source:'const+std+%3D+@import(%22std%22)%3B%0Aconst+mem+%3D+std.mem%3B%0A%0Aexport+fn+current(a_ptr:+%5B*%5Dconst+u8,+a_len:+usize,+b:+%5B*:0%5Dconst+u8)+bool+%7B%0A++++const+a+%3D+a_ptr%5B0..a_len%5D%3B%0A++++return+mem.eql(u8,+a,+mem.span(b))%3B%0A%7D%0A%0Aexport+fn+better(a_ptr:+%5B*%5Dconst+u8,+a_len:+usize,+b:+%5B*:0%5Dconst+u8)+bool+%7B%0A++++const+a+%3D+a_ptr%5B0..a_len%5D%3B%0A++++for+(a)+%7Ca_elem,+idx%7C+%7B%0A++++++++if+(b%5Bidx%5D+%3D%3D+0+or+b%5Bidx%5D+!!%3D+a_elem)+return+false%3B%0A++++%7D%0A++++return+b%5Ba.len%5D+%3D%3D+0%3B%0A%7D'),l:'5',n:'0',o:'Zig+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:ztrunk,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),fontScale:14,j:1,lang:zig,libs:!(),options:'-OReleaseFast',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'zig+trunk+(Editor+%231,+Compiler+%231)+Zig',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)
I'm no expert in asm, but I've been led to believe that fewer instructions are better

andrewrk:
Closing old drafts; please feel free to re-open if you decide to pick this back up.

andrewrk:
I see this was intended to be a proposal; filed #8414 on your behalf

Issue #7842, opened 2021-01-20
translate-c: Incomplete array declaration with empty initializer triggers panic


ehaas:
```
➜  zig version
0.8.0-dev.1030+072d1e088
```

Put following in `bug.h` and run translate-c on it:
```c
const int i[] = {};
```

Output:
```
➜ zig translate-c bug.h
thread 14609677 panic: cast causes pointer to be null
/Users/ehaas/source/zig/lib/std/math/big/int.zig:1299:17: 0x104fb21ad in std.math.big.int.Const.toString (zig1)
        return s.len;
                ^
/Users/ehaas/source/zig/lib/std/fmt.zig:1444:26: 0x10529d7c1 in std.fmt.bufPrint (zig1)
    return fbs.getWritten();
                         ^
/Users/ehaas/source/zig/src/translate_c.zig:4359:22: 0x1050d64c8 in translate_c.transCreateNodePtrType (zig1)
    const op_token = switch (op_tok_id) {
                     ^
/Users/ehaas/source/zig/src/translate_c.zig:0:52: 0x104fa104f in translate_c.transType (zig1)
/Users/ehaas/source/zig/lib/std/math/big/int.zig:0:9: 0x104fb21d8 in std.math.big.int.Const.toString (zig1)
/Users/ehaas/source/zig/src/translate_c.zig:913:22: 0x104e820d3 in translate_c.transCreateNodeTypedef (zig1)
    const name_tok = try appendIdentifier(rp.c, checked_name);
                     ^
/Users/ehaas/source/zig/src/translate_c.zig:940:16: 0x104d5760d in translate_c.transRecordDecl (zig1)
        return try transCreateNodeIdentifier(c, name); // Avoid processing this decl twice
               ^
/Users/ehaas/source/zig/src/translate_c.zig:514:35: 0x104be9d43 in translate_c.declVisitor (zig1)
            return visitFnDecl(c, @ptrCast(*const clang.FunctionDecl, decl));
                                  ^
/Users/ehaas/source/zig/src/translate_c.zig:470:62: 0x104be0683 in translate_c.prepopulateGlobalNameTable (zig1)
    const it_end = ast_unit.getLocalPreprocessingEntities_end();
                                                             ^
???:?:?: 0x1059ef192 in __ZNKSt3__122__compressed_pair_elemIPPKcLi0ELb0EE5__getEv (???)
???:?:?: 0x1056c182a in _ZigClangASTUnit_visitLocalTopLevelDecls (???)
[...]

Issue #7832, opened 2021-01-19
Support C/C++ unsized array initialization


MoAlyousef:
Hello
Thanks for all your efforts!
First, I'm sorry if this is a duplicate, I couldn't find a similar issue.

Many C/C++ libs use unsized array initialization:
```c
const char *const names[] = {"first", "second", "third"};
```
Using zig cc builds source files containing them just fine. However, using such headers via cInclude generates the following error (assuming it's a translate-c error):
```
.\zig-cache\o\a542ccfdd97e89bd29e207e8d9116014\cimport.zig:1:65: error: expected type '[*c]const [*c]const u8', found '[3][*c]const u8'
pub export const names: [*c]const [*c]const u8 = [3][*c]const u8{
                                                                ^
.\src\main.zig:4:47: note: referenced here
    @import("std").debug.warn("hello {s}", .{c.names[1]});
                                              ^
te...The following command exited with error code 1:
C:\ProgramData\chocolatey\lib\zig\tools\zig-windows-x86_64-0.7.1\zig.exe build-exe D:\dev\zigprojs\te\src\main.zig --library c --cache-dir D:\dev\zigprojs\te\zig-cache --global-cache-dir C:\Users\may64\AppData\Local\zig --name te -I D:\dev\zigprojs\te --enable-cache
error: the following build command failed with exit code 1:
D:\dev\zigprojs\te\zig-cache\o\9492f3e39ce97d8dca9460a0b7231af7\build.exe C:\ProgramData\chocolatey\lib\zig\tools\zig-windows-x86_64-0.7.1\zig.exe D:\dev\zigprojs\te D:\dev\zigprojs\te\zig-cache C:\Users\may64\AppData\Local\zig
```

My main.zig:
```zig
const c = @cImport(@cInclude("header.h"));

pub fn main() void {
    @import("std").debug.warn("hello {s}", .{c.names[1]});
}
```
Header.h:
```c
#pragma once

const char *const names[] = {
    "first",
    "second",
    "third",
};
```
build.zig:
```zig
const Builder = @import("std").build.Builder;

pub fn build(b: *Builder) void {
    const target = b.standardTargetOptions(.{});
    const mode = b.standardReleaseOptions();
    const exe = b.addExecutable("te", "src/main.zig");
    exe.addIncludeDir(".");
    exe.linkSystemLibrary("c");
    exe.setTarget(target);
    exe.setBuildMode(mode);
    exe.install();
    const run_cmd = exe.run();
[...]

PR #7849, opened 2021-01-21
translate-c: Improve array support

### LINKED ISSUE #4831: translate-c doesn't understand automatic sized arrays.
Zig version: 0.5.0+aec4967f3

Trying to translate this C code
```
const uint32_t shader_data[] = {1, 2, 3, 4};
```
results in
```
pub export const data: [*c]const c_uint = [4]c_uint{
    @bitCast(c_uint, @as(c_int, 1)),
    @bitCast(c_uint, @as(c_int, 2)),
    @bitCast(c_uint, @as(c_int, 3)),
    @bitCast(c_uint, @as(c_int, 4)),
};
```
which fails to compile as zig, because it thinks data is a pointer while it's actually an array.

ehaas:
1. For incomplete arrays with initializer list (`int x[] = {1};`) use the
initializer size as the array size.

2. For arrays initialized with a string literal translate it as an array
of character literals instead of `[*c]const u8`

3. Don't crash if an empty initializer is used for an incomplete array.

4. Add a test for multi-character character constants

Additionally lay some groundwork for supporting wide string literals.

fixes #4831 #7832 #7842

### DIFF
--- a/src/clang.zig
+++ b/src/clang.zig
@@ -735,6 +735,15 @@ pub const StringLiteral = opaque {
     pub const getKind = ZigClangStringLiteral_getKind;
     extern fn ZigClangStringLiteral_getKind(*const StringLiteral) StringLiteral_StringKind;
 
+    pub const getCodeUnit = ZigClangStringLiteral_getCodeUnit;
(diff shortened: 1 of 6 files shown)


andrewrk:
Nice work!!

PR #7850, opened 2021-01-21
Add the ability to disable dynamic image base (Windows)

chivay:
Hi!
This is a simple addition to make linking process more configurable.

By default, PE executables are linked with dynamic base enabled,
however in some cases it may be useful to override this (eg. for easier
debugging).

Add `--no-dynamic-base` flag to change the default linker behavior.


zigazeljko:
Do we need a new flag for this or could we reuse `-fno-PIE` (which currently does nothing when linking PE)?

PR #7854, opened 2021-01-23
std: add MultiArrayListUnmanaged

ifreund:
This was requested on IRC by @andrewrk for use in the self hosted compiler. This is a draft PR as I'm not 100% settled on the API yet, which could also use expansion to support more operations similar to ArrayList.

### DIFF
--- a/lib/std/multi_array_list.zig
+++ b/lib/std/multi_array_list.zig
@@ -1,0 +1,249 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2015-2021 Zig Contributors
+// This file is part of [zig](https://ziglang.org/), which is MIT licensed.
+// The MIT license requires this copyright notice to be included in all copies
+// and substantial portions of the software.
+const std = @import("std");
+const assert = std.debug.assert;
+const meta = std.meta;
+const Allocator = std.mem.Allocator;
+
+/// Given a struct type, return a type managing growable, parallel slices
+/// of the field type of all the struct's fields.
+pub fn MultiArrayList(comptime S: type) type {
+    return struct {
+        const Self = @This();
+
+        pub const Field = meta.FieldEnum(S);
+
+        const fields = meta.fields(S);
+
+        data: [fields.len][*]u8 = undefined,
+        capacity: usize = 0,
+        len: usize = 0,
+
+        /// Init all lists to the given capacity
+        pub fn initCapacity(allocator: *Allocator, capacity: usize) !Self {
+            var self = Self{ .capacity = capacity };
+            inline for (fields) |field, index| {
+                errdefer {
+                    comptime var i = 0;
+                    inline while (i < index) : (i += 1) {
+                        allocator.free(self.allocatedSlice(@intToEnum(Field, i)));
+                    }
+                }
+                const slice = try allocator.allocAdvanced(field.field_type, field.alignment, capacity, .exact);
+                self.data[index] = @ptrCast([*]u8, slice.ptr);
+            }
+            return self;
+        }
+
+        /// Free all memory and set the MultiArrayList to undefined
+        pub fn deinit(self: *Self, allocator: *Allocator) void {
+            inline for (fields) |_, i| {
+                allocator.free(self.allocatedSlice(@intToEnum(Field, i)));
+            }
+            self.* = undefined;
+        }
+
+        pub fn items(self: *Self, comptime field: Field) []align(fieldAlign(field)) FieldType(field) {
+            return self.allocatedSlice(field)[0..self.len];
+        }
+
+        pub fn allocatedSlice(self: *Self, comptime field: Field) []align(fieldAlign(field)) FieldType(field) {
+            if (self.capacity == 0) return &[0]FieldType(field){};
+            return @ptrCast(
+                [*]FieldType(field),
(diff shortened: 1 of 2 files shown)


andrewrk:
What do you think about this (untested) sketch for an implementation that puts everything in the same allocation?

```zig
pub fn MultiArrayList(comptime S: type) type {
    return struct {
        const Self = @This();

        pub const Field = meta.FieldEnum(S);

        bytes: [*]align(@alignOf(S)) u8 = undefined,
        len: usize = 0,
        capacity: usize = 0,

        /// Array of @sizeOf each S field.
        /// Sorted by alignment, descending.
        /// (TODO)
        const field_sizes = {};
        /// Array mapping from field to its index in field_sizes array.
        /// (TODO)
        const field_size_indexes = {};

        /// Release all allocated memory.
        pub fn deinit(self: *Self, gpa: *Allocator) void {
            gpa.free(self.allocatedBytes());
            self.* = undefined;
        }

        pub fn items(self: *Self, comptime field: Field) []FieldType(field) {
            const T = FieldType(field);
            var offset: usize = 0;
            inline for (field_sizes[0..field_size_indexes(@enumToInt(field))]) |field_size| {
                offset += field_size * self.capacity;
            }
            const ptr = @alignCast(@alignOf(T), @ptrCast([*]T, self.bytes + offset));
            return ptr[0..self.len];
        }

        /// Extend the list by 1 element. Allocates more memory as necessary.
        pub fn append(self: *Self, gpa: *Allocator, elem: S) !void {
            try self.ensureCapacity(gpa, self.len + 1);
            self.appendAssumeCapacity(elem);
        }

        /// Extend the list by 1 element, but asserting `self.capacity`
        /// is sufficient to hold an additional item.
        pub fn appendAssumeCapacity(self: *Self, elem: S) void {
            assert(self.len < self.capacity);
            self.len += 1;
            inline for (meta.fields(S)) |field_info, i| {
                self.items(@intToEnum(Field, i))[self.len - 1] = @field(values, field_info.name);
            }
        }

        /// Adjust the list's length to `new_len`.
        /// Does not initialize added items, if any.
        pub fn resize(self: *Self, gpa: *Allocator, new_len: usize) !void {
            try self.ensureCapacity(gpa, new_len);
            self.items.len = new_len;
        }

        /// Attempt to reduce allocated capacity to `new_len`.
        /// If `new_len` is greater than zero, this may fail to reduce the capacity,
        /// but the data remains intact and the length is updated to new_len.
        pub fn shrinkAndFree(self: *Self, gpa: *Allocator, new_len: usize) void {
            if (new_len == 0) {
                gpa.free(self.allocatedBytes());
                self.* = .{};
                return;
            }
            assert(new_len <= self.capacity);
            assert(new_len <= self.len);

            var other = Self{};
            other.setCapacity(gpa, new_len) catch {
                self.len = new_len;
                // TODO memset the invalidated items to undefined
                return;
            };
            other.len = new_len;
            inline for (meta.fields(S)) |field_info, i| {
                const field = @intToEnum(Field, i);
                mem.copy(field_info.field_type, other.items(field), self.items(field));
            }
            gpa.free(self.allocatedBytes());
            self.* = other;
        }

        /// Reduce length to `new_len`.
        /// Invalidates pointers to elements `items[new_len..]`.
        /// Keeps capacity the same.
        pub fn shrinkRetainingCapacity(self: *Self, new_len: usize) void {
            self.len = new_len;
        }

        /// Modify the array so that it can hold at least `new_capacity` items.
        /// Implements super-linear growth to achieve amortized O(1) append operations.
        /// Invalidates pointers if additional memory is needed.
        pub fn ensureCapacity(self: *Self, gpa: *Allocator, new_capacity: usize) !void {
            var better_capacity = self.capacity;
            if (better_capacity >= new_capacity) return;

            while (true) {
                better_capacity += better_capacity / 2 + 8;
            }

            return self.setCapacity(gpa, better_capacity);
        }

        /// Modify the array so that it can hold exactly `new_capacity` items.
        /// Invalidates pointers if additional memory is needed.
        pub fn setCapacity(self: *Self, gpa: *Allocator, new_capacity: usize) !void {
            const bytes = try gpa.reallocAdvanced(
                self.allocatedBytes(),
                @alignOf(S),
                capacityInBytes(new_capacity),
                .Exact,
            );
            self.bytes = bytes.ptr;
            self.capacity = new_capacity;
        }

        fn capacityInBytes(capacity: usize) usize {
            // TODO this looks like a job for SIMD (need that @reduce builtin tho)
            var bytes_count: usize = 0;
            for (field_sizes) |field_size| {
                bytes_count += better_capacity * field_size;
            }
            return bytes_count;
        }

        fn allocatedBytes(self: Self) []align(@alignOf(S)) u8 {
            return self.bytes[0..capacityInBytes(self.capacity)];
        }

        fn FieldType(field: Field) type {
            return meta.fieldInfo(S, field).field_type;
        }
    };
}
```

andrewrk:
@ifreund you might want to ping me on IRC before you do any more work on this because I started adding more functionality to my fork of this and using it locally and I think you might like it

andrewrk:
I think the next step here is to swap out the master branch implementation with this one, and run some benchmarks to see which one has better perf / peak memory usage. The main difference is one allocation vs one per field. Depending on the results of those benchmarks, we will know which implementation to keep.

andrewrk:
Extracted into #8415.

PR #7856, opened 2021-01-23
add LTO support

### LINKED ISSUE #2845: LTO across zig/C boundary
Zig is not doing lto of C objects.

As an example, https://github.com/via/zig-miniblink
If I build the libopencm3 submodule with clang and use no lto anywhere, I get what I expect -- zig calls the C code in blink.h, which calls functions in libopencm3:
```
git submodule update
cd libopencm3
make CC='clang -target arm--eabi -mcpu=cortex-m4f -mfloat-abi=hard -mfpu=fpv4-sp-d16' CFLAGS='--sysroot=/usr/lib/arm-none-eabi'
```

Produces:
```
0801006c <fff>:
 801006c:       b580            push    {r7, lr}
 801006e:       466f            mov     r7, sp
 8010070:       b082            sub     sp, #8
 8010072:       f7ff ffc5       bl      8010000 <anotherfunc>
 8010076:       f7ff ffcb       bl      8010010 <toggle_blink>
...
08010010 <toggle_blink>:
 8010010:       b5d0            push    {r4, r6, r7, lr}
 8010012:       af02            add     r7, sp, #8
 8010014:       f240 0400       movw    r4, #0
 8010018:       f2c2 0400       movt    r4, #8192       ; 0x2000
 801001c:       6820            ldr     r0, [r4, #0]
 801001e:       2801            cmp     r0, #1
 8010020:       d00b            beq.n   801003a <toggle_blink+0x2a>
 8010022:       b990            cbnz    r0, 801004a <toggle_blink+0x3a>
 8010024:       f640 4000       movw    r0, #3072       ; 0xc00
 8010028:       f44f 5180       mov.w   r1, #4096       ; 0x1000
 801002c:       f2c4 0002       movt    r0, #16386      ; 0x4002
 8010030:       f000 f826       bl      8010080 <gpio_set>
```

If I use lto with the libopencm3, and pass lto to build blink.c:
```
cd libopencm3; make clean
make CC='clang -target arm--eabihf -mcpu=cortex-m4f -mfloat-abi=hard -mfpu=fpv4-sp-d16' CFLAGS='--sysroot=/usr/lib/arm-none-eabi -O3 -flto'
```

blink.o shows the gpio_* functions being inlined:
```
l08010040 <toggle_blink>:
 8010040:       f240 0100       movw    r1, #0
 8010044:       f2c2 0100       movt    r1, #8192       ; 0x2000
 8010048:       6808            ldr     r0, [r1, #0]
 801004a:       b128            cbz     r0, 8010058 <toggle_blink+0x18>
 801004c:       2801            cmp     r0, #1
 801004e:       d10d            bne.n   801006c <toggle_blink+0x2c>
 8010050:       2000            movs    r0, #0
 8010052:       f04f 5280       mov.w   r2, #268435456  ; 0x10000000
 8010056:       e002            b.n     801005e <toggle_blink+0x1e>
 8010058:       2001            movs    r0, #1
 801005a:       f44f 5280       mov.w   r2, #4096       ; 0x1000
 801005e:       f640 4318       movw    r3, #3096       ; 0xc18
 8010062:       f2c4 0302       movt    r3, #16386      ; 0x4002
 8010066:       601a            str     r2, [r3, #0]
 8010068:       6008            str     r0, [r1, #0]
 801006a:       4770            bx      lr
 801006c:       2000            movs    r0, #0
 801006e:       6008            str     r0, [r1, #0]
 8010070:       4770            bx      lr
 8010072:       d4d4            bmi.n   801001e <fff+0x2>
```

The zig calls in main.zig aren't inlined, but if I use -Drelease-fast they are. Still no inlining of the C calls though
```
08010000 <main>:
 8010000:       b580            push    {r7, lr}
 8010002:       f000 f804       bl      801000e <anotherfunc>
 8010006:       e8bd 4080       ldmia.w sp!, {r7, lr}
 801000a:       f000 b808       b.w     801001e <toggle_blink>

0801000e <anotherfunc>:
 801000e:       f640 4018       movw    r0, #3096       ; 0xc18
 8010012:       f44f 5100       mov.w   r1, #8192       ; 0x2000
 8010016:       f2c4 0002       movt    r0, #16386      ; 0x4002
 801001a:       6001            str     r1, [r0, #0]
 801001c:       4770            bx      lr

0801001e <toggle_blink>:
 801001e:       f240 0100       movw    r1, #0
 8010022:       f2c2 0100       movt    r1, #8192       ; 0x2000
 8010026:       6808            ldr     r0, [r1, #0]
 8010028:       b128            cbz     r0, 8010036 <toggle_blink+0x18>
 801002a:       2801            cmp     r0, #1
 801002c:       d10d            bne.n   801004a <toggle_blink+0x2c>
 801002e:       2000            movs    r0, #0
 8010030:       f04f 5280       mov.w   r2, #268435456  ; 0x10000000
 8010034:       e002            b.n     801003c <toggle_blink+0x1e>
 8010036:       2001            movs    r0, #1
 8010038:       f44f 5280       mov.w   r2, #4096       ; 0x1000
 801003c:       f640 4318       movw    r3, #3096       ; 0xc18
 8010040:       f2c4 0302       movt    r3, #16386      ; 0x4002
 8010044:       601a            str     r2, [r3, #0]
 8010046:       6008            str     r0, [r1, #0]
 8010048:       4770            bx      lr
 801004a:       2000            movs    r0, #0
 801004c:       6008            str     r0, [r1, #0]
 801004e:       4770            bx      lr
```

andrewrk:
The CLI gains -flto and -fno-lto options to override the default.
However, the cool thing about this is that the defaults are great! In
general when you use build-exe in release mode, Zig will enable LTO if
it would work and it would help.

zig cc supports detecting and honoring the -flto and -fno-lto flags as
well. The linkWithLld functions are improved to all be the same with
regards to copying the artifact instead of trying to pass single objects
through LLD with -r. There is possibly a future improvement here as
well; see the respective TODOs.

stage1 is updated to support outputting LLVM bitcode instead of machine
code when lto is enabled. This allows LLVM to optimize across the Zig and
C/C++ code boundary.

closes #2845

----

Here's an example. It's the example from the Clang LTO documentation, except I've replaced main.c with main.zig:

#### main.zig

```zig
const std = @import("std");

export fn foo4() void {
    _ = std.c.printf("Hi\n");
}

extern fn foo1() c_int;

pub fn main() u8 {
    return @intCast(u8, foo1());
}
```

#### a.c

```c
int foo1(void);
void foo2(void);
void foo4(void);

static signed int i = 0;

void foo2(void) {
  i = -1;
}

static int foo3() {
  foo4();
  return 10;
}

int foo1(void) {
  int data = 0;

  if (i < 0)
    data = foo3();

  data = data + 42;
  return data;
}
```

```
$ ./zig build-exe main.zig a.c -OReleaseFast -lc 
$ ./main
$ echo $?
42
$ objdump -d main -Mintel | grep -A7 '<main'
0000000000201530 <main>:
  201530:	48 c7 c0 ff ff ff ff 	mov    rax,0xffffffffffffffff
  201537:	66 0f 1f 84 00 00 00 	nop    WORD PTR [rax+rax*1+0x0]
  20153e:	00 00 
  201540:	48 83 7c c2 08 00    	cmp    QWORD PTR [rdx+rax*8+0x8],0x0
  201546:	48 8d 40 01          	lea    rax,[rax+0x1]
  20154a:	75 f4                	jne    201540 <main+0x10>
  20154c:	48 63 cf             	movsxd rcx,edi
  20154f:	48 89 35 7a 22 00 00 	mov    QWORD PTR [rip+0x227a],rsi        # 2037d0 <argv>
  201556:	48 89 0d 7b 22 00 00 	mov    QWORD PTR [rip+0x227b],rcx        # 2037d8 <argv+0x8>
  20155d:	48 89 15 7c 22 00 00 	mov    QWORD PTR [rip+0x227c],rdx        # 2037e0 <environ.0>
  201564:	48 89 05 7d 22 00 00 	mov    QWORD PTR [rip+0x227d],rax        # 2037e8 <environ.0+0x8>
  20156b:	b8 2a 00 00 00       	mov    eax,0x2a
  201570:	c3                   	ret    
```

The interesting thing to note here is that there was no LTO explicitly opted into. It happened automatically. And you can see here that in the main function, there is no call to `foo1` and there is no exported `foo4`. If we didn't have LTO, the call to `foo1` could not have been inlined. For example, here's what happens if we force-disable LTO:

```
$ ./zig build-exe main.zig a.c -OReleaseFast -lc -fno-lto
$ objdump -d main -Mintel | grep -A7 '<main'
00000000002015d0 <main>:
  2015d0:	50                   	push   rax
  2015d1:	48 c7 c0 ff ff ff ff 	mov    rax,0xffffffffffffffff
  2015d8:	0f 1f 84 00 00 00 00 	nop    DWORD PTR [rax+rax*1+0x0]
  2015df:	00 
  2015e0:	48 83 7c c2 08 00    	cmp    QWORD PTR [rdx+rax*8+0x8],0x0
  2015e6:	48 8d 40 01          	lea    rax,[rax+0x1]
  2015ea:	75 f4                	jne    2015e0 <main+0x10>
  2015ec:	48 63 cf             	movsxd rcx,edi
  2015ef:	48 89 35 fa 22 00 00 	mov    QWORD PTR [rip+0x22fa],rsi        # 2038f0 <argv>
  2015f6:	48 89 0d fb 22 00 00 	mov    QWORD PTR [rip+0x22fb],rcx        # 2038f8 <argv+0x8>
  2015fd:	48 89 15 fc 22 00 00 	mov    QWORD PTR [rip+0x22fc],rdx        # 203900 <environ.0>
  201604:	48 89 05 fd 22 00 00 	mov    QWORD PTR [rip+0x22fd],rax        # 203908 <environ.0+0x8>
  20160b:	e8 90 ff ff ff       	call   2015a0 <foo1>
  201610:	0f b6 c0             	movzx  eax,al
```

Now you can see we are forced to call `foo` and return its result.


LemonBoy:
Nice, but making it opt-out at high optimization levels is not so nice. LTO is slower and more resource-intensive than regular linking, a new user would have to dive into the docs to know LTO is on my default and how to turn it off (and, last but not least, know what LTO is and it's tradeoffs).

kubkon:
> Nice, but making it opt-out at high optimization levels is not so nice. LTO is slower and more resource-intensive than regular linking, a new user would have to dive into the docs to know LTO is on my default and how to turn it off (and, last but not least, know what LTO is and it's tradeoffs).

This may be true but LTO is only enabled by default in Release mode, where you'd expect longer compile/link times, would you not? If this is the case, I think having LTO by default in Release mode is nice to have. Having said that however, as a counterexample, Rust disables LTO by default even in Release mode ([link](https://doc.rust-lang.org/cargo/reference/profiles.html#release)) so clearly the defaults for LTO are debatable.

LemonBoy:
> would you not?

No.

Mouvedia:
Is [`PrepareForLTO`](https://reviews.llvm.org/rG83daa49758a12d585fe2d9a64448e54d91bcfaff) relevant for this PR?

rohlem:
> > [I]n Release mode [you would] expect longer compile/link times, would you not?
>
> No.

I too would have been under that impression. Could you elaborate where `ReleaseSafe` and `ReleaseFast` should draw the line of "too expensive"? Anything that takes longer than `Debug` (including incremental compilation), or maybe twice that time?
Would we still want a "hard boiled" build mode that defaults to going the extra step regardless of build time?

I'm not sure "universally sensible" defaults exist, or at least I'd be happy to hear opinions on this. Though I guess if it's too complex, that discussion should be its own issue somewhere.

LemonBoy:
The line is between codegen and linking, LTO doesn't impact the former.

andrewrk:
I think I can make a reasonable defense of keeping it opt out for release builds. First of all it has always been the case that the trade-off for release builds is to wait longer for compilation in exchange for better runtime performance, even by a few orders of magnitude. Whether this is to blame on codegen optimization or linking is beside the point.

The point about new users is important to consider, however consider further that for new users who start a small, new project, LTO in release modes will in fact be the correct choice for them. Only when their project grows quite large will it start to become worthwhile to consider disabling it, and that would happen after the user spend a long time working on the project, by that time graduating from new user to experienced user, having had plenty of time to learn about the option to disable LTO.

New users who join an existing project will use the already existing build script so they won't have to think about LTO at all.

In all cases, new users don't have to think about this feature.

I also want to point out that we already have this equivalent problem in zig code currently because we don't split large zig projects up into multiple compilation units. So probably we will eventually need some kind of feature that is resource-aware and will split up a release build into an arbitrary number of compilation units in order to satisfy a memory resource budget or provide better caching. This same feature clearly could be used to limit LTO in order to satisfy the same constraints.

Thus I argue the defaults set forth in this PR are the best way forward.

saethlin:
I'm not a Zig user, but I have a fair bit of bad experiences with LLVM's LTO from my experience in Rust, so I'm here advising caution.

I've seen perf regressions from turning on LTO between 10% on large codebases and 5x on microbenchmarks. The biggest failure mode seems to be around not emitting a memcpy when LTO is on. When LTO causes a regression, it the inlining decisions seem the same as far as I can tell but other things fall over.

I've also seen huge problems with old versions of perf on Rust LTO binaries. Specifically, the version CentOS 7 ships and when using `perf record --call-graph dwarf`. I've rarely seen a call graph come out of that that makes sense when it does run. Often that version of perf will segfault or crash with odd error messages. The situation is significantly better in recent builds of perf, but the call stacks are still sometimes obviously wrong.

Is there large-scale testing that can be done to see if Zig programs will suffer from these problems?

komuw:
> > Nice, but making it opt-out at high optimization levels is not so nice. LTO is slower and more resource-intensive than regular linking, a new user would have to dive into the docs to know LTO is on my default and how to turn it off (and, last but not least, know what LTO is and it's tradeoffs).
> 
> This may be true but LTO is only enabled by default in Release mode, where you'd expect longer compile/link times, would you not? If this is the case, I think having LTO by default in Release mode is nice to have. Having said that however, as a counterexample, Rust disables LTO by default even in Release mode ([link](https://doc.rust-lang.org/cargo/reference/profiles.html#release)) so clearly the defaults for LTO are debatable.

Looking at the Rust source code, it seems like `lto = false` does not mean disable LTO all-together; instead it means enable `thin LTO`. Whereas `lto = true` means enable `fat LTO`:  https://github.com/rust-lang/cargo/blob/b52fc0a8270d671e33dafe8dec355624727e8534/src/cargo/core/profiles.rs#L766-L773      
So Rust has `thin LTO` enabled by default in release mode.

saethlin:
> So Rust has `thin LTO` enabled by default in release mode.

See the actual usage of that type: https://github.com/rust-lang/cargo/blob/b52fc0a8270d671e33dafe8dec355624727e8534/src/cargo/core/compiler/lto.rs#L48

Rust has had a lot of debate about the default release settings, and it's entirely unclear how much of that is relevant to Zig: https://github.com/rust-lang/rust/issues/57968#issuecomment-458577925

andrewrk:
Thanks @saethlin - those are two things I'll be on the look out for. I'm going to make the call here to proceed with this at least for now and see how it goes. Of course if we run into trouble, this decision can be re-evaluated. But I think this is a worthwhile idea to try out at this point.

PR #7857, opened 2021-01-23
linux: add fallocate() to io_uring

jorangreef:


lib/std/os/linux/io_uring.zig
```diff
@@ -1395,3 +1435,47 @@ test "timeout_remove" {
...
+    testing.expectEqual(@as(u64, 0), (try file.stat()).size);
+
+    const len: u64 = 65536;
+    const sqe = try ring.fallocate(0xaaaaaaaa, file.handle, 0, 0, len);
+    testing.expectEqual(linux.IORING_OP.FALLOCATE, sqe.opcode);
+    testing.expectEqual(file.handle, sqe.fd);
+    testing.expectEqual(@as(u32, 1), try ring.submit());
+
+    const cqe = try ring.copy_cqe();
+    switch (-cqe.res) {
```
daurnimator, line 1463: This code is convincing me that we shouldn't try to copy the liburing API.

PR #7860, opened 2021-01-23
stage1: Fix f128 NaN check on sparc64

koachan:
On big-endian hosts, zig_f128_isNaN() takes the high and low halves
from the wrong element, resulting in buggy NaN detection behavior.
This fixes it.


andrewrk:
Thanks!

PR #7861, opened 2021-01-23
fix windows bug in Progress.zig (deadlocked compiler on windows)

FireFox317:
This bug caused the compiler to deadlock when multiple c objects were build in parallel.

Thanks @kprotty for finding this bug!

Now we can use the `win-dev-kit` again to build the compiler :)

It is probably a good idea to check all the functions in `kernel32.zig`, because this is a really annoying bug to find and fix.

### DIFF
--- a/lib/std/Progress.zig
+++ b/lib/std/Progress.zig
@@ -103,7 +103,11 @@ pub const Node = struct {
             }
             parent.completeOne();
         } else {
-            self.context.done = true;
+            {
+                const held = self.context.update_lock.acquire();
+                defer held.release();
+                self.context.done = true;
+            }
             self.context.refresh();
         }
     }
--- a/lib/std/os/windows/kernel32.zig
+++ b/lib/std/os/windows/kernel32.zig
@@ -299,6 +299,6 @@ pub extern "kernel32" fn SleepConditionVariableSRW(
     f: ULONG,
 ) callconv(WINAPI) BOOL;
 
-pub extern "kernel32" fn TryAcquireSRWLockExclusive(s: *SRWLOCK) callconv(WINAPI) BOOL;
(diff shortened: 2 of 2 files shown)


andrewrk:
Thanks, landed in b56e916fa1d3508a01a6fc0b62f3f64dd3843b85. Not sure why this would cause a deadlock though, can you explain it to me?

FireFox317:
With the windows APIs there is a difference between a `BOOL` and a `BOOLEAN`:
https://github.com/ziglang/zig/blob/b56e916fa1d3508a01a6fc0b62f3f64dd3843b85/lib/std/os/windows/bits.zig#L32-L33

The wrong return value was specified for the following function, thus the return value was probably always 0 (which is why we probably want to have a look at all the functions in `kernel32.zig`):
https://github.com/ziglang/zig/blob/b56e916fa1d3508a01a6fc0b62f3f64dd3843b85/lib/std/os/windows/kernel32.zig#L302

This function is used in the `tryAcquire` function of the Mutex:
https://github.com/ziglang/zig/blob/b56e916fa1d3508a01a6fc0b62f3f64dd3843b85/lib/std/Thread/Mutex.zig#L266-L272

And the `tryAcquire` function of the Mutex, is used in the Progress implementation (and on line 174):
https://github.com/ziglang/zig/blob/b56e916fa1d3508a01a6fc0b62f3f64dd3843b85/lib/std/Progress.zig#L163

So it was a pretty critical bug, which made the `tryAcquire` of the Mutex on Windows not work correctly.

andrewrk:
Ahh I see thank you, so the return type was the real fix and the other thing was a (harmless?) data race on a boolean

FireFox317:
Exactly!

Yeah i wasn't sure if that data race was harmless, because writing to the same value from different threads simultaneously is UB. That's why I added the lock to that, but it probably works correctly without that lock too.

PR #7863, opened 2021-01-23
make std.Thread.Impl public to allow referring to Held

kristoff-it:


kristoff-it:
King seems to have a better idea about this.

PR #7864, opened 2021-01-23
std: Define all wasm constants/instructions in std.wasm

Luukdegram:
This is the first effort to declare all constants the wasm backend will be using for its code generation similar to `elf.zig` and `coff.zig`. It allows us to make the code a lot more readable instead of hardcoding all its values and reduce the possibility of typos at the same time. e.g. this allows for `code.append(wasm.opcode(.i32_load));` instead of `code.append(0x28);`
I tried to keep the name of the opcodes equal to what is defined by the wasm spec (but replacing `.` with `_`) so parsing Wat to opcodes should be simpler.

ps: codegen/wasm.zig will be updated accordingly in a follow-up PR

cc @ifreund Hope this is what you had in mind as well.
cc @fengb

### DIFF
--- a/lib/std/wasm.zig
+++ b/lib/std/wasm.zig
@@ -1,0 +1,266 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2015-2021 Zig Contributors
+// This file is part of [zig](https://ziglang.org/), which is MIT licensed.
+// The MIT license requires this copyright notice to be included in all copies
+// and substantial portions of the software.
+const testing = @import("std.zig").testing;
+
+/// Wasm instruction opcodes
+///
+/// All instructions are defined as per spec:
+/// https://webassembly.github.io/spec/core/appendix/index-instructions.html
+pub const Opcode = enum(u8) {
+    @"unreachable" = 0x00,
+    nop = 0x01,
+    block = 0x02,
+    loop = 0x03,
+    @"if" = 0x04,
+    @"else" = 0x05,
+    end = 0x0B,
+    br = 0x0C,
+    br_if = 0x0D,
+    br_table = 0x0E,
+    @"return" = 0x0F,
+    call = 0x10,
+    call_indirect = 0x11,
+    drop = 0x1A,
+    select = 0x1B,
+    local_get = 0x20,
+    local_set = 0x21,
+    local_tee = 0x22,
+    global_get = 0x23,
+    global_set = 0x24,
+    i32_load = 0x28,
+    i64_load = 0x29,
+    f32_load = 0x2A,
+    f64_load = 0x2B,
+    i32_load8_s = 0x2C,
+    i32_load8_u = 0x2D,
+    i32_load16_s = 0x2E,
+    i32_load16_u = 0x2F,
+    i64_load8_s = 0x30,
+    i64_load8_u = 0x31,
+    i64_load16_s = 0x32,
+    i64_load16_u = 0x33,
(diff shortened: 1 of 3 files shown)


lib/std/wasm.zig
```diff
@@ -0,0 +1,274 @@
...
+
+    //extensions
+    i32_trunc_sat_f32_s = 0xFC0,
+    i32_trunc_sat_f32_u = 0xFC1,
+    i32_trunc_sat_f64_s = 0xFC2,
+    i32_trunc_sat_f64_u = 0xFC3,
+    i64_trunc_sat_f32_s = 0xFC4,
+    i64_trunc_sat_f32_u = 0xFC5,
+    i64_trunc_sat_f64_s = 0xFC6,
+    i64_trunc_sat_f64_u = 0xFC7,
```
ifreund, line 193: These need to be handled differently, the opcode is always the one byte `0xFC` which is then followed by a ULEB encoded u32. See the bottom of this page: https://webassembly.github.io/spec/core/binary/instructions.html

lib/std/wasm.zig
```diff
@@ -0,0 +1,274 @@
...
+// This file is part of [zig](https://ziglang.org/), which is MIT licensed.
+// The MIT license requires this copyright notice to be included in all copies
+// and substantial portions of the software.
+const testing = @import("std.zig").testing;
+
+/// Wasm instruction opcodes
+///
+/// All instructions are defined as per spec:
+/// https://webassembly.github.io/spec/core/appendix/index-instructions.html
+pub const Opcode = enum(u16) {
```
ifreund, line 12: Opcodes are one byte. This enum should use a byte as its tag type.
This enum should also be made non-exhaustive so that the spec adding new opcodes is not a breaking change.
```suggestion
pub const Opcode = enum(u8) {
```

lib/std/wasm.zig
```diff
@@ -0,0 +1,274 @@
...
+    i32_trunc_sat_f64_u = 0xFC3,
+    i64_trunc_sat_f32_s = 0xFC4,
+    i64_trunc_sat_f32_u = 0xFC5,
+    i64_trunc_sat_f64_s = 0xFC6,
+    i64_trunc_sat_f64_u = 0xFC7,
+};
+
+/// Returns the integer value of an `Opcode`. Used by the Zig compiler
+/// to write instructions to the wasm binary file
+pub fn opcode(op: Opcode) u16 {
```
ifreund, line 198:
```suggestion
pub fn opcode(op: Opcode) u8 {
```

ifreund (changes requested):
Thaks for working on this, should make self-hosted a lot cleaner!

lib/std/wasm.zig
Luukdegram: @ifreund Thanks for this. At least now I understand how they work. As we discussed on IRC, I'm leaving them out for now.

src/link/Wasm.zig
```diff
@@ -7,6 +7,7 @@ const assert = std.debug.assert;
 const fs = std.fs;
 const leb = std.leb;
 const log = std.log.scoped(.link);
+const spec = std.wasm;
```
ifreund, line 10: hmm, I'd just call this wasm. Renaming it seems confusing.

ifreund (approved):
LGTM! You should probably squash this into a single commit.

Luukdegram:
> LGTM! You should probably squash this into a single commit.

You're right! Done.

PR #7870, opened 2021-01-24
Add the linker flags to the `--libc` file

hvenev:
When using the system libc, this lets the user choose which libraries
are linked in.


hvenev:
I amended the commit with no changes because of an `apt-get update` error in CI.

andrewrk:
I don't think this is the right approach. This needs to go through the proposal process first.

PR #7871, opened 2021-01-24
Add --[no-]link-libc[pp]

hvenev:
Those flags override the check for `-lc -lm -lrt -ldl -lcrypt -lSystem`
in the library flags. If either `--[no-]link-libc` is specified, all
libraries are passed through without any change.

Note that currently `--no-link-libc` will be ignored if the target
requires the use of libc (e.g. non-linux).

One other change is that `-nostdlib` must now be given to `zig cc` when
building freestanding binaries, just like with the traditional `cc`.


andrewrk:
I don't agree. This needs to go through the proposal process.

PR #7884, opened 2021-01-25
Fill out more cases for std.meta.sizeof

SpexGuy:
@Srekel was running into some missing cases in `std.meta.sizeof`, specifically for string literals.  I filled in some more cases for it.  Since we don't actually have enough information at this level to correctly figure out how C's sizeof would behave in all cases, it has to guess a bit.  But until translate-c is smart enough to do this correctly without deferring to a helper, we will always have this problem.  Zig's type system just can't encode these things in the way that is needed to exactly match C.

### DIFF
--- a/lib/std/meta.zig
+++ b/lib/std/meta.zig
@@ -979,9 +979,59 @@ test "std.meta.cast" {
 /// Given a value returns its size as C's sizeof operator would.
 /// This is for translate-c and is not intended for general use.
 pub fn sizeof(target: anytype) usize {
-    switch (@typeInfo(@TypeOf(target))) {
-        .Type => return @sizeOf(target),
-        .Float, .Int, .Struct, .Union, .Enum => return @sizeOf(@TypeOf(target)),
+    const T: type = if (@TypeOf(target) == type) target else @TypeOf(target);
+    switch (@typeInfo(T)) {
+        .Float, .Int, .Struct, .Union, .Enum, .Array, .Bool, .Vector => return @sizeOf(T),
+        .Fn => {
+            // sizeof(main) returns 1, sizeof(&main) returns pointer size.
+            // We cannot distinguish those types in Zig, so use pointer size.
+            return @sizeOf(T);
+        },
+        .Null => return @sizeOf(*c_void),
+        .Void => {
+            // Note: sizeof(void) is 1 on clang/gcc and 0 on MSVC.
+            return 1;
+        },
+        .Opaque => {
+            if (T == c_void) {
+                // Note: sizeof(void) is 1 on clang/gcc and 0 on MSVC.
+                return 1;
+            } else {
+                @compileError("Cannot use C sizeof on opaque type "++@typeName(T));
+            }
+        },
+        .Optional => |opt| {
+            if (@typeInfo(opt.child) == .Pointer) {
+                return sizeof(opt.child);
+            } else {
+                @compileError("Cannot use C sizeof on non-pointer optional "++@typeName(T));
+            }
+        },
+        .Pointer => |ptr| {
+            if (ptr.size == .Slice) {
+                @compileError("Cannot use C sizeof on slice type "++@typeName(T));
+            }
+            // for strings, sizeof("a") returns 2.
+            // normal pointer decay scenarios from C are handled
+            // in the .Array case above, but strings remain literals
+            // and are therefore always pointers, so they need to be
+            // specially handled here.
+            if (ptr.size == .One and ptr.is_const and @typeInfo(ptr.child) == .Array) {
+                const array_info = @typeInfo(ptr.child).Array;
+                if ((array_info.child == u8 or array_info.child == u16) and
+                     array_info.sentinel != null and
+                     array_info.sentinel.? == 0) {
+                    // length of the string plus one for the null terminator.
+                    return (array_info.len + 1) * @sizeOf(array_info.child);
+                }
+            }
+            // When zero sized pointers are removed, this case will no
+            // longer be reachable and can be deleted.
+            if (@sizeOf(T) == 0) {
+                return @sizeOf(*c_void);
+            }
+            return @sizeOf(T);
+        },
         .ComptimeFloat => return @sizeOf(f64), // TODO c_double #3999
         .ComptimeInt => {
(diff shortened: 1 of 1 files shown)


lib/std/meta.zig
```diff
@@ -979,9 +979,59 @@ test "std.meta.cast" {
...
+    switch (@typeInfo(T)) {
+        .Float, .Int, .Struct, .Union, .Enum, .Array, .Bool, .Vector => return @sizeOf(T),
+        .Fn => {
+            // sizeof(main) returns 1, sizeof(&main) returns pointer size.
+            // We cannot distinguish those types in Zig, so use pointer size.
+            return @sizeOf(T);
+        },
+        .Null => return @sizeOf(*c_void),
+        .Void => {
+            // Note: sizeof(void) is 1 on clang/gcc and 0 on MSVC.
```
daurnimator, line 992: should we look at `builtin.abi`?

lib/std/meta.zig
```diff
@@ -979,9 +979,59 @@ test "std.meta.cast" {
 /// Given a value returns its size as C's sizeof operator would.
 /// This is for translate-c and is not intended for general use.
 pub fn sizeof(target: anytype) usize {
-    switch (@typeInfo(@TypeOf(target))) {
-        .Type => return @sizeOf(target),
-        .Float, .Int, .Struct, .Union, .Enum => return @sizeOf(@TypeOf(target)),
+    const T: type = if (@TypeOf(target) == type) target else @TypeOf(target);
+    switch (@typeInfo(T)) {
+        .Float, .Int, .Struct, .Union, .Enum, .Array, .Bool, .Vector => return @sizeOf(T),
+        .Fn => {
```
data-man, line 985:
```suggestion
        .Fn, .BoundFn => {
```

lib/std/meta.zig
```diff
@@ -979,9 +979,59 @@ test "std.meta.cast" {
...
+        .Optional => |opt| {
+            if (@typeInfo(opt.child) == .Pointer) {
+                return sizeof(opt.child);
+            } else {
+                @compileError("Cannot take C size of non-pointer optional "++@typeName(T));
+            }
+        },
+        .Pointer => |ptr| {
+            if (ptr.size == .Slice) {
+                @compileError("Cannot take C sizeof a slice");
```
daurnimator, line 1012: inconsistent whitespace. here you have "C sizeof" but on line 1007 above you have "C size of" with a space between size and of.

lib/std/meta.zig
```diff
@@ -979,9 +979,59 @@ test "std.meta.cast" {
 /// Given a value returns its size as C's sizeof operator would.
 /// This is for translate-c and is not intended for general use.
 pub fn sizeof(target: anytype) usize {
-    switch (@typeInfo(@TypeOf(target))) {
-        .Type => return @sizeOf(target),
-        .Float, .Int, .Struct, .Union, .Enum => return @sizeOf(@TypeOf(target)),
+    const T: type = if (@TypeOf(target) == type) target else @TypeOf(target);
```
daurnimator, line 982: This is odd, why do we need it?

lib/std/meta.zig
daurnimator: Why shouldn't bound functions be an error?

lib/std/meta.zig
SpexGuy:
```
/// Given a value returns its size as C's sizeof operator would.
/// This is for translate-c and is not intended for general use.
```

lib/std/meta.zig
SpexGuy: Translate C will not generate bound functions, this is not needed.

lib/std/meta.zig
SpexGuy: Not necessarily, that's a different check.  Code compiled with `zig cc` will behave like clang (using 1 for this value) even when compiling for the MSVC abi.  I think it makes sense to match the behavior of `zig cc` here, at least until translate-c is smart enough that we can delete this function.

lib/std/meta.zig
data-man: Why do you limit the use of this function only for translate C?

SpexGuy: Because that's what it was before I made these changes.  I am simply trying to unblock someone who was running into the TODO compile error at the bottom.

lib/std/meta.zig
SpexGuy: Fixed, they have all been changed to "Cannot use C sizeof on ..."

lib/std/meta.zig
```diff
@@ -991,20 +1041,58 @@ pub fn sizeof(target: anytype) usize {
             // TODO test if target fits in int, long or long long
             return @sizeOf(c_int);
         },
-        else => @compileError("TODO implement std.meta.sizeof for type " ++ @typeName(@TypeOf(target))),
+        .EnumLiteral => {
```
Vexu, line 1044: `translate-c` should never generate enum literals so I don't think we should allow them here, same goes for `.Vector`.

lib/std/meta.zig
```diff
@@ -991,20 +1041,58 @@ pub fn sizeof(target: anytype) usize {
...
+            // C enums are the size of an integer
+            return @sizeOf(c_int);
+        },
+        else => @compileError("TODO implement std.meta.sizeof for type " ++ @typeName(T)),
```
Vexu, line 1048:
```suggestion
        else => @compileError("std.meta.sizeof does not support type " ++ @typeName(T)),
```

Vexu (changes requested):

lib/std/meta.zig
SpexGuy: Makes sense for enum literal but it seems like we might eventually want to translate __m128 and friends to vector.

Vexu: I guess that could happen, I was mostly going with what already exists.

SpexGuy: I removed the .EnumLiteral case, but it seems plausible that .Vector will be used at some point, and that whoever does that update to translate-c will not remember to modify this function.  I don't really see a problem with leaving it in, but if you feel strongly about it I'll take it out.

Vexu (approved):

PR #7889, opened 2021-01-26
Fix integer overflow when calling joinZ with empty slices

fancl20:
Hi, I triggered integer overflow when calling joinZ with an empty slices.

```zig
const str = try joinZ(testing.allocator, ",", &[_][]const u8{});
```

This should fix the problem.

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -1507,7 +1507,7 @@ pub fn joinZ(allocator: *Allocator, separator: []const u8, slices: []const []con
 }
 
 fn joinMaybeZ(allocator: *Allocator, separator: []const u8, slices: []const []const u8, zero: bool) ![]u8 {
-    if (slices.len == 0) return &[0]u8{};
+    if (slices.len == 0) return if (zero) try allocator.dupe(u8, &[1]u8{0}) else &[0]u8{};
 
     const total_len = blk: {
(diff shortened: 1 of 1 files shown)


daurnimator (commented):
Should we be allocating/duping in the non-zero case too? ==> otherwise the returned pointer (despite being 0 length) may not be valid for passing to the allocator.

fancl20:
> Should we be allocating/duping in the non-zero case too? ==> otherwise the returned pointer (despite being 0 length) may not be valid for passing to the allocator.

[We will skip 0 length pointer](https://github.com/ziglang/zig/blob/79730e6f5cdf13b2514781881e69c557fd52eaea/lib/std/mem/Allocator.zig#L453) when calling allocator.free. So that should be fine?

daurnimator:
> [We will skip 0 length pointer](https://github.com/ziglang/zig/blob/79730e6f5cdf13b2514781881e69c557fd52eaea/lib/std/mem/Allocator.zig#L453) when calling allocator.free. So that should be fine?

ah yes; that appears to be true.
TIL we also special case 0-length in the allocator: https://github.com/ziglang/zig/blob/79730e6f5cdf13b2514781881e69c557fd52eaea/lib/std/mem/Allocator.zig#L281-L283

PR #7896, opened 2021-01-27
Change tuple formatting

data-man:
```zig
const std = @import("std");
const print = std.debug.print;

pub fn main () void {
    const t = .{ 1, 2.718281828, .{ 3.1415926 } };
    print("{}\n", .{t});
}
```
Before:
> struct:5:16{ .0 = 1, .1 = 2.718281828e+00, .2 = struct:5:35{ .0 = 3.1415926e+00 } }

After:
> { .0 = 1, .1 = 2.718281828e+00, .2 = { .0 = 3.1415926e+00 } }

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -484,7 +484,7 @@ pub fn formatType(
             }
         },
         .Struct => |info| {
-            try writer.writeAll(@typeName(T));
+            if (!info.is_tuple) try writer.writeAll(@typeName(T));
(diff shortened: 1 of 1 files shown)


Vexu:
If we're going to format tuples differently from structs then I think we should also omit the field names.

data-man:
@Vexu 

I was thinking of improving API formatting.

- if a type has `display` (or `print`) use it
- otherwise use `format`

The same for `debug`.

Issue #7794, opened 2021-01-16
Proposal: @offsetOf instead of @byteOffsetOf


hvenev:
I believe that the builtin used for getting the byte offset of a struct field should be called `@offsetOf` rather than `@byteOffsetOf`. Here are a few reasons why:

- It is consistent with `@sizeOf` / `@bitSizeOf`. `@alignOf` is also not called `@byteAlignOf`.
- `offsetOf` is common in other programming languages. The C macro is called `offsetof`. There are also a few Rust libraries that provide macros named `offset_of`.
- Bit-packed structs are rare, especially dealing with pointers to individual fields.

I also believe `@offsetOf` should fail when the struct field is not aligned to a byte boundary. The most obvious reason is to prevent bugs where `@byteOffsetOf` is used without checking if the offset is actually a whole number of bytes. Having this check as a part of `@offsetOf` will make subtly incorrect code invalid and will make correct code cleaner.

Using `@byteOffsetOf` on non-byte-aligned bitfields is rarely useful, and I'd imagine that all such uses will be made more clear by doing `@bitOffetOf(T, field) / 8` instead. Even if we keep the current `@byteOffsetOf` for backwards compatibility, I believe it should be phased out.

A quick `grep` through the standard library revealed that the only uses of `@byteOffsetOf` are assertions about the layout of `extern struct`s. This can also be done with the proposed `@offsetOf` with the added advantage that if we were to make the structs `packed`, we'd automatically make sure that the fields are byte-aligned.

faraazahmad:
I'm not too familiar with Zig yet so forgive my ignorance, but what was the reason `@byteOffsetOf` and `@offsetOf` were named so?

hvenev:
`@offsetOf` does not exist.

hvenev:
I've implemented the non-breaking version of this proposal at https://github.com/ziglang/zig/pull/7903.

hvenev:
I've updated the proposal.

andrewrk:
See also  #8642

PR #7903, opened 2021-01-28
Implement @offsetOf

hvenev:
@offsetOf behaves like @byteOffsetOf, with the exception that it fails
if the field does not begin at a byte boundary, i.e. the bit offset is
a multiple of 8.

### DIFF
--- a/src/astgen.zig
+++ b/src/astgen.zig
@@ -3043,6 +3043,7 @@ fn nodeMayNeedMemoryLocation(start_node: *ast.Node, scope: *Scope) bool {
                     .{ "@wasmMemoryGrow", false },
                     .{ "@mod", false },
                     .{ "@mulWithOverflow", false },
+                    .{ "@offsetOf", false },
                     .{ "@panic", false },
                     .{ "@popCount", false },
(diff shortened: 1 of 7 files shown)


andrewrk:
For language changes, let's go through the proposal process for this first, please, then PR comes after it is accepted.

hvenev:
What's the proposal process?

mikdusan:
just create an **issue** for it

hvenev:
> just create an **issue** for your proposal

Can I re-use https://github.com/ziglang/zig/issues/7794?

SpexGuy:
Yes, but for a change like this (making offsetOf fail for bit packed fields), please add motivating examples of existing code where the change would be helpful to avoid bugs or footguns, or to make the code simpler.

SpexGuy:
The process for language changes is:
1. A proposal issue is created proposing the idea.
2. Andrew will make a decision on it, either marking it as accepted or closing it.  This sometimes takes a while, but if the issue makes a very good argument it can be fast.
3. PRs for accepted proposals may be merged.

Issue #7901, opened 2021-01-28
std.json can't handle large integers


marler8997:
Currently `std.json` can't handle integers larger than `i64`.

## My Use Case

I'm working on autogenerated Win32 zig bindings from a JSON source.  In this file (https://github.com/marlersoft/win32json/blob/main/api/Js.json), the constant value for `JS_SOURCE_CONTEXT_NONE` is a number that cannot fit into an `i64`, namely, `18446744073709551615`.

mrakh:
Per RFC 7159, JSON parsers are allowed to set limits on the range and precision of accepted numbers, and setting the integer limit to the max word value of the architecture (2^64 - 1) is a common convention of JSON implementations in low-level languages.

marler8997:
Well I'm not generating the data, it's generated by microsoft so I can't really "change it".  It's representable with a `u64`, so we could make it a union like this:
```
const JsonInteger = union(enum) {
    I64: i64,
    U64: u64,
};
```
Maybe I'll create a PR for that.

SpexGuy:
Javascript doesn't have integers, only double floats, so integer values above +/-2^53 - 1 may not be preserved correctly in many json implementations.  If your goal is to be compatible with json parsers that aren't the Zig stdlib, consider using strings for these values instead.

marler8997:
@SpexGuy that's a good suggestion, maybe I'll do that

daurnimator:
Are you referring to the non-streaming parser under `json.Parser`? https://github.com/ziglang/zig/blob/236db6232fa2beb97c47e3f1edcdb2a29e59d160/lib/std/json.zig#L2095

If you instead use `json.parse` it should support all zig integer types. https://github.com/ziglang/zig/blob/236db6232fa2beb97c47e3f1edcdb2a29e59d160/lib/std/json.zig#L1410

See also #5326

marler8997:
@daurnimator it looks like that function is for deserializing JSON into a type?  How would that work for parsing a file like this? https://github.com/marlersoft/win32json/blob/main/api/Js.json

daurnimator:
> @daurnimator it looks like that function is for deserializing JSON into a type?

Yes

> How would that work for parsing a file like this? https://github.com/marlersoft/win32json/blob/main/api/Js.json

Create a zig type with that structure.

```zig
const T = struct {
  Constants: []struct {
    Name: []const u8,
    Type: struct {
      Kind: []const u8,
      Name: []const u8,
    },
[...]

PR #7904, opened 2021-01-29
std.json u64 support

marler8997:
fix issue https://github.com/ziglang/zig/issues/7901

A minimal set of changes to std.json to support integers up to the maximum of a `u64`.

NOTE: A future change may want to add support for BigInt.  Another idea I had was to encapsulate all possible number types into a `Number` sub-union, so that when user code doesn't need to know the number representation, they can work with the Number type to avoid duplicating code for each representation.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1254,6 +1254,7 @@ pub const Value = union(enum) {
     Null,
     Bool: bool,
     Integer: i64,
+    UInteger: u64,
     Float: f64,
     String: []const u8,
     Array: Array,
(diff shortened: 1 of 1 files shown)


daurnimator (commented):
I don't think this is the right direction to be going in: it's climbing a local maximum.
e.g. you would never get an unsigned integer with the high byte unset.
and it brings up questions of e.g. "why 64 bits? why not just have `i65` instead?" or `i128`, or as you said: BigNum.

marler8997:
@daurnimator I'm all ears to alternative suggestions :)

As to why `u64`, the answer is I have a use case for it.  Though I admit it's not a particularly good reason.  However, it seems to be low cost to go from `i64` to also supporting `u64`.  It climbing slightly higher on the local maximum for a low cost.  If you can pointer me towards a better maximum I'll gladly explore it :)

pfgithub:
Could integer instead be changed to `[]const u8` and then users asked to parse it themselves with `std.fmt.parseInt`? Either that or BigNum would solve this.

marler8997:
@pfgithub I mentioned in the description that could be future work.  I wanted to do the minimal change to get JSON working for my use case, is there a downside to that approach?

marler8997:
Thinking on it, I like @pfgithub 's suggestion.  It allows the parser to support the JSON spec for any number in its entirety and lets the user decide how to deal with its type.  I'm going to close this and push a new PR with that idea.

PR #7905, opened 2021-01-29
std: Add "_FromWriteFileStep" for all the remaining LibExeObjStep types

rgreenblatt:
Seems like a good idea for consistency.


Vexu:
Thanks!

PR #7909, opened 2021-01-29
Added support for passing write file args as build options

rgreenblatt:
Works the same way as addBuildOptionArtifact. This PR also adds "addPathBuildOption" for abstraction - I think this should be public, but I'm not certain. 

Note that this PR changes the behavior of addBuildOptionArtifact in the edge case where setOutputPath was called and a relative path was used. I think this behavior change is basically a bug fix because the path in build_options should really always be absolute.

I've tested this locally and everything I tried worked. I would add tests, but I don't think there are any integration tests set up for Builder.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -2016,8 +2016,11 @@ pub const LibExeObjStep = struct {
     ) void {
         // Note that pathFromRoot uses resolve path, so this will have
         // correct behavior even if getOutputPath is already absolute.
-        const abs_path = std.zig.fmtEscapes(self.builder.pathFromRoot(path));
-        self.addBuildOption([]const u8, name, abs_path);
+        const escaped_abs_path = self.builder.fmt(
+            "{s}",
+            std.zig.fmtEscapes(self.builder.pathFromRoot(path)),
+        );
+        self.addBuildOption([]const u8, name, escaped_abs_path);
     }
 
     pub fn addSystemIncludeDir(self: *LibExeObjStep, path: []const u8) void {
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -2006,23 +2006,6 @@ pub const LibExeObjStep = struct {
         self.step.dependOn(&write_file.step);
     }
 
-    /// Adds build option corresponding to the absolute (resolved) path.
-    /// Path is either an absolute path or a path relative to the build
-    /// working directory.
-    pub fn addPathBuildOption(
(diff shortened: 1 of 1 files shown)


lib/std/build.zig
```diff
@@ -2155,11 +2194,23 @@ pub const LibExeObjStep = struct {
             }
         }
 
-        if (self.build_options_contents.items.len > 0 or self.build_options_artifact_args.items.len > 0) {
-            // Render build artifact options at the last minute, now that the path is known.
+        if (self.build_options_contents.items.len > 0 or
+            self.build_options_artifact_args.items.len > 0 or
+            self.build_options_write_file_args.items.len > 0)
+        {
+            const path_option_fmt = "pub const {s}: []const u8 = \"{}\";\n";
```
Vexu, line 2201: You probably meant to use this somewhere, also remember to use `std.zig.fmtId` and  `std.zig.fmtEscapes.

rgreenblatt: nice catch

lib/std/build.zig
```diff
@@ -1939,6 +1947,40 @@ pub const LibExeObjStep = struct {
...
+        self: *LibExeObjStep,
+        name: []const u8,
+        path: []const u8,
+    ) void {
+        // Note that pathFromRoot uses resolve path, so this will have
+        // correct behavior even if getOutputPath is already absolute.
+        const escaped_abs_path = self.builder.fmt(
+            "{s}",
+            std.zig.fmtEscapes(self.builder.pathFromRoot(path)),
+        );
```
Vexu, line 1980: This will cause the path to be double escaped since `addBuildOption` already handles string escaping.

This whole function seems a bit unnecessary since all it does is `.addBuildOption([]const u8, name, b.pathFromRoot(path))` which I think is more explicit about what value you are getting.

rgreenblatt: Oh yeah, that's why I didn't orginally use fmtEscapes, forgot about that. I agree that the function isn't needed. When I was originally writing this would need to handle a bit more and then I never reconsidered.

rgreenblatt: I removed it in the last commit.

rgreenblatt:
I don't think that build fail is due to this change.

g-w1:
Try rebasing with master to re-run the ci.

rgreenblatt:
Anything blocking this from being merged?

PR #7911, opened 2021-01-29
Build and install documentation as part as the default build process

### LINKED ISSUE #7897: Generate documentation as part of the build process
There's a doc directory with these two files `docgen.zig`  `langref.html.in` but no rule in `CMakeLists.txt` to usethem to generate a proper documentation.
Any hint on how to do so, something I could use to submit a patch would be great!

Thanks

serge-sans-paille:
This should fix #7897

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -798,6 +798,28 @@ elseif(MINGW)
   target_link_libraries(zig ntdll)
 endif()
 
+file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/doc)
+add_custom_command(
+  OUTPUT "${CMAKE_BINARY_DIR}/doc/docgen"
+  COMMAND "${ZIG_EXECUTABLE}" "build-exe" "${CMAKE_SOURCE_DIR}/doc/docgen.zig"
+  DEPENDS "${ZIG_EXECUTABLE}"
+  COMMENT STATUS "Building docgen utility"
+  WORKING_DIRECTORY "${CMAKE_BINARY_DIR}/doc"
+)
+
+add_custom_command(
+  OUTPUT "${CMAKE_BINARY_DIR}/doc/langref.html"
+  COMMAND "${CMAKE_BINARY_DIR}/doc/docgen" "${ZIG_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/doc/langref.html.in" "${CMAKE_BINARY_DIR}/doc/langref.html"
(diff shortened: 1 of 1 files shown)


g-w1:
This takes a looong time. I had to kill it after 5 mins because it was eating my cpu:
image
In comparison, master:
image

serge-sans-paille:
@g-w1 I made doc an optional target, so that it only triggers if one runs `ninja doc`

g-w1:
You should also put documentation for this somewhere (pun intended).

serge-sans-paille:
@g-w1 I made doc an optional target, so that it only triggers if one runs `ninja doc`

> You should also put documentation for this somewhere (pun intended).

That was a good one :-) I propose to update https://github.com/ziglang/zig/wiki/Building-Zig-From-Source once this PR is merged.

Jan200101:
There are various things missing in the CMakeLists.txt that are present in the build.zig so I wonder if that was on purpose

serge-sans-paille:
@Jan200101 I don't know either. Is that blocking for this PR?

Jan200101:
I wouldn't call it blocking but its up in the air if this is desired behavior

Vexu:
I don't see the point of doing this instead of `make && ./zig build docs`.

serge-sans-paille:
1. `make doc` sounds more explicit than `./zig build docs` to me
2. `make doc && make install` would install documentation in a standard location

Jan200101:
It could be argued that the CMakeLists.txt is only there to get to a fully functional Zig and with that you can do anything else using zig build

serge-sans-paille:
Indeed :-)

g-w1:
> Indeed :-)

If you agree with this, then it would make sense that building documentation is not part of getting a fully functional zig and it should be done with `zig build docs`.

serge-sans-paille:
I'm currently packaging zig for Fedora. As a packager, I see value in the doc installation step, but in the end it's up to you and no big deal to me to do it explicitly if upstream is not interested ;-)

Jan200101:
> I'm currently packaging zig for Fedora. As a packager, I see value in the doc installation step, but in the end it's up to you and no big deal to me to do it explicitly if upstream is not interested ;-)

you can build the docs just fine for a fedora package
[I have done so myself](https://pagure.io/zig-spec/blob/fedora/f/zig.spec#_63)

andrewrk:
We will have this eventually, but it will not be via cmake. A `zig build` will build the self-hosted compiler as well as the docs. For now, the process is a bit more cumbersome, while we work on bootstrapping.

Issue #7906, opened 2021-01-29
Add "allow unknown fields" option to std.json.parse


clarityflowers:
I'm currently working with a huge json blob that has a lot of unnecessary information for my purposes. In the interest of speeding up both compile and run time I'd love to be able to define my result type to only have the object fields I'm interested in, and have the parser skip over the rest.

```zig
test "parse can ignore unknown fields" {
    const options = ParseOptions{ 
        .allocator = testing.allocator, 
        .ignore_unknown_fields = true, 
    };

    const T = struct {
        int: i64,
        language: []const u8,
    };
    const r = try parse(T, &TokenStream.init(
        \\{
        \\  "int": 420,
        \\  "float": 3.14,
        \\  "with\\escape": true,
        \\  "with\u0105unicode\ud83d\ude02": false,
        \\  "optional": null,
        \\  "static_array": [66.6, 420.420, 69.69],
        \\  "dynamic_array": [66.6, 420.420, 69.69],
        \\  "complex": {
        \\    "nested": "zig"
        \\  },
        \\  "veryComplex": [
        \\    {
        \\      "foo": "zig"
        \\    }, {
        \\      "foo": "rocks"
        \\    }
        \\  ],
        \\  "a_union": 100000,
        \\  "language": "zig"
        \\}
    ), options);
    defer parseFree(T, r, options);
    testing.expectEqual(@as(i64, 420), r.int);
    testing.expectEqualSlices(u8, "zig", r.language);
}
```

PR #5852, opened 2020-07-12
std: add json parsing option to skip unknown fields

daurnimator:
Requested by @Sobeston. 

I had to have the user pass in a buffer via `ParseOptions` to use for tracking the ignored state

PR #7918, opened 2021-01-30
Add option to ignore unknown fields when parsing json

EthanGruffudd:
As requested in https://github.com/ziglang/zig/issues/7906, I've reworked the previous attempt (https://github.com/ziglang/zig/pull/5852) into something that doesn't need a user-provided buffer.

I'll be honest, I'm not confident my code is perfect, but it seems to do what I want based on the tests.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1493,7 +1493,7 @@ test "skipValue" {
     { // An absurd number of nestings
         const nestings = 256;
 
-        testing.expectError(
+        try testing.expectError(
             error.TooManyNestedItems,
             skipValue(&TokenStream.init("[" ** nestings ++ "]" ** nestings)),
         );
@@ -1505,14 +1505,14 @@ test "skipValue" {
 
         try skipValue(&TokenStream.init(deeply_nested_array));
 
-        testing.expectError(
+        try testing.expectError(
             error.TooManyNestedItems,
             skipValue(&TokenStream.init("[" ++ deeply_nested_array ++ "]")),
         );
     }
 
     // Mismatched brace/square bracket
-    testing.expectError(
+    try testing.expectError(
         error.UnexpectedClosingBrace,
         skipValue(&TokenStream.init("[102, 111, 111}")),
     );
@@ -1520,11 +1520,11 @@ test "skipValue" {
     { // should fail if no value found (e.g. immediate close of object)
         var empty_object = TokenStream.init("{}");
         assert(.ObjectBegin == (try empty_object.next()).?);
-        testing.expectError(error.UnexpectedJsonDepth, skipValue(&empty_object));
+        try testing.expectError(error.UnexpectedJsonDepth, skipValue(&empty_object));
 
         var empty_array = TokenStream.init("[]");
         assert(.ArrayBegin == (try empty_array.next()).?);
-        testing.expectError(error.UnexpectedJsonDepth, skipValue(&empty_array));
+        try testing.expectError(error.UnexpectedJsonDepth, skipValue(&empty_array));
     }
 }
 
@@ -2152,8 +2152,8 @@ test "parse into struct ignoring unknown fields" {
     ), ops);
     defer parseFree(T, r, ops);
 
-    testing.expectEqual(@as(i64, 420), r.int);
-    testing.expectEqualSlices(u8, "zig", r.language);
+    try testing.expectEqual(@as(i64, 420), r.int);
+    try testing.expectEqualSlices(u8, "zig", r.language);
 }
 
 /// A non-stream JSON parser which constructs a tree of Value's.
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1105,6 +1105,10 @@ pub const TokenStream = struct {
         };
     }
 
+    fn stackUsed(self: *TokenStream) u8 {
+        return self.parser.stack_used + if (self.token != null) @as(u8, 1) else 0;
+    }
+
     pub fn next(self: *TokenStream) Error!?Token {
         if (self.token) |token| {
             self.token = null;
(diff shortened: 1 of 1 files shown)


zigazeljko:
Is the `waiting_for_closing_type` necessary? Since `StreamingParser` already checks that start and end tokens match, just checking stack depth should be enough.

daurnimator:
> Is the `waiting_for_closing_type` necessary? Since `StreamingParser` already checks that start and end tokens match, just checking stack depth should be enough.

Yeah. that was the discovery back in #5852: `skipValue` just needs to keep iterating until you reach the original stack depth.

lib/std/json.zig
```diff
@@ -1383,8 +1383,69 @@ pub const ParseOptions = struct {
...
+test "skipValue" {
+    try skipValue(&TokenStream.init("false"));
+    try skipValue(&TokenStream.init("true"));
+    try skipValue(&TokenStream.init("null"));
+    try skipValue(&TokenStream.init("42"));
+    try skipValue(&TokenStream.init("42.0"));
+    try skipValue(&TokenStream.init("\"foo\""));
+    try skipValue(&TokenStream.init("[101, 111, 121]"));
+    try skipValue(&TokenStream.init("{}"));
+    try skipValue(&TokenStream.init("{\"foo\": \"bar\"}"));
```
daurnimator, line 1446: Could you add the other (failing) test cases from #5852?

daurnimator (changes requested):

EthanGruffudd:
> Is the `waiting_for_closing_type` necessary? Since `StreamingParser` already checks that start and end tokens match, just checking stack depth should be enough.

I think I see what you mean, will be submitting the changes now.

lib/std/json.zig
EthanGruffudd: I had to increase the number of nestings for [one of them](https://github.com/ziglang/zig/pull/5852/commits/404bbd385592e79c939f851b9d17c94d209bdde0#diff-defc5ed2d34b24434f31f2d92e270f4f98a1a32ec07147799eb12c1242d5b0a4R1464) to fail, but I couldn't reproduce [the last one](https://github.com/ziglang/zig/pull/5852/commits/404bbd385592e79c939f851b9d17c94d209bdde0#diff-defc5ed2d34b24434f31f2d92e270f4f98a1a32ec07147799eb12c1242d5b0a4R1469). Could you give me some feedback on it?

daurnimator: What happended when you tried that last case?

zigazeljko: I think you meant `tokens.parser.stack_used < original_depth`, since `tokens.parser.stack` is the stack itself, not its depth.

EthanGruffudd: Applied the corrections. This will also return an error on empty arrays, so I added a test for that. Is this behavior desirable?

lib/std/json.zig
```diff
@@ -1383,8 +1383,60 @@ pub const ParseOptions = struct {
...
+
+    /// If false, finding an unknown field returns an error.
+    ignore_unknown_fields: bool = false,
 };
 
+fn skipValue(tokens: *TokenStream) !void {
+    const original_depth = tokens.parser.stack_used;
+
+    // Return an error if no value is found
+    _ = try tokens.next();
```
daurnimator, line 1395: you also need the `== original_depth` check here

daurnimator (changes requested):

lib/std/json.zig
EthanGruffudd: That is indeed an error I failed to see, but that's not what's causing the tests to fail.

[This is the problematic test](https://github.com/ziglang/zig/pull/7918/files#diff-defc5ed2d34b24434f31f2d92e270f4f98a1a32ec07147799eb12c1242d5b0a4R1947). According to my probing, `stack_used` is reduced on the third number's token, and not on the `ArrayEnd`; so, when `skipValue` returns, the parser finds an unexpected token. `complex` and `veryComplex` don't seem to be affected by this, even after adding fields with numbers in them.

EthanGruffudd: I couldn't figure out why `stack_used` was being reduced in such a way, so I made a workaround for it, would this be acceptable?

daurnimator: Ah! That'll be because of the 1 token lookahead for numbers in the streaming parser.

Does something like this work?:
```diff
diff --git a/lib/std/json.zig b/lib/std/json.zig
index ff258c7f3..e1f7e5df6 100644
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1105,6 +1105,10 @@ pub const TokenStream = struct {
         };
     }
 
+    fn stack_used(self: TokenStream) u8 {
+        return self.parser.stack_used + if (self.token != null) 1 else 0;
+    }
+
     pub fn next(self: *TokenStream) Error!?Token {
         if (self.token) |token| {
             self.token = null;

```

zigazeljko: Shouldn't that be `stackUsed` (instead of `stack_used`)?

lib/std/json.zig
```diff
@@ -1105,6 +1105,10 @@ pub const TokenStream = struct {
         };
     }
 
+    fn stackUsed(self: *TokenStream) u8 {
+        return self.parser.stack_used + if (self.token != null) @as(u8, 1) else 0;
```
daurnimator, line 1109: I wonder if this overflows when the innermost token in a 256-deep-stack is a number?
If so, will need to upgrade the return value to a `u9` (and do some casts)

EthanGruffudd: Tried with `skipValue`, but it didn't panic on an overflow. Should I add a test specific to TokenStream too?

PR #7920, opened 2021-01-31
Rework AST memory layout for better memory usage and performance

andrewrk:
This is a proof-of-concept of switching to a new memory layout for
tokens and AST nodes. The goal is threefold:

  * smaller memory footprint
  * faster performance for tokenization and parsing
  * most importantly, a proof-of-concept that can be also applied to ZIR
    and TZIR to improve the entire compiler pipeline in this way.

I had a few key insights here:

  * Underlying premise: using less memory will make things faster, because
    of fewer allocations and better cache utilization. Also using less
    memory is valuable in and of itself.
  * Using a Struct-Of-Arrays for tokens and AST nodes, saves the bytes of
    padding between the enum tag (which kind of token is it; which kind
    of AST node is it) and the next fields in the struct. It also improves
    cache coherence, since one can peek ahead in the tokens array without
    having to load the source locations of tokens.
  * Token memory can be conserved by only having the tag (1 byte) and byte
    offset (4 bytes) for a total of 5 bytes per token. It is not necessary
    to store the token ending byte offset because one can always re-tokenize
    later, but also most tokens the length can be trivially determined from
    the tag alone, and for ones where it doesn't, string literals for
    example, one must parse the string literal again later anyway in
    astgen, making it free to re-tokenize.
  * AST nodes do not actually need to store more than 1 token index because
    one can poke left and right in the tokens array very cheaply.

So far we are left with one big problem though: how can we put AST nodes
into an array, since different AST nodes are different sizes?

This is where my key observation comes in: one can have a hash table for
the extra data for the less common AST nodes! But it gets even better than
that:

I defined this data that is always present for every AST Node:

  * tag (1 byte)
    - which AST node is it
  * main_token (4 bytes, index into tokens array)
    - the tag determines which token this points to
  * struct{lhs: u32, rhs: u32}
    - enough to store 2 indexes to other AST nodes, the tag determines
      how to interpret this data

You can see how a binary operation, such as `a * b` would fit into this
structure perfectly. A unary operation, such as `*a` would also fit,
and leave `rhs` unused. So this is a total of 13 bytes per AST node.
And again, we don't have to pay for the padding to round up to 16 because
we store in struct-of-arrays format.

I made a further observation: the only kind of data AST nodes need to
store other than the main_token is indexes to sub-expressions. *That's it.*
**The only purpose of an AST is to bring a tree structure to a list of tokens.**
This observation means all the data that nodes store are only sets of u32
indexes to other nodes. The other tokens can be found later by the compiler,
by poking around in the tokens array, which again is super fast because it
is struct-of-arrays, so you often only need to look at the token tags array,
which is an array of bytes, very cache friendly.

So for nearly every kind of AST node, you can store it in 13 bytes. For the
rarer AST nodes that have 3 or more indexes to other nodes to store, either
the lhs or the rhs will be repurposed to be an index into an extra_data array
which contains the extra AST node indexes. In other words, no hash table needed,
it's just 1 big ArrayList with the extra data for AST Nodes.

Final observation, no need to have a canonical tag for a given AST. For example:
The expression `foo(bar)` is a function call. Function calls can have any
number of parameters. However in this example, we can encode the function
call into the AST with a tag called `FunctionCallOnlyOneParam`, and use lhs
for the function expr and rhs for the only parameter expr. Meanwhile if the
code was `foo(bar, baz)` then the AST node would have to be `FunctionCall`
with lhs still being the function expr, but rhs being the index into
`extra_data`. Then because the tag is `FunctionCall` it means
`extra_data[rhs]` is the "start" and `extra_data[rhs+1]` is the "end".
Now the range `extra_data[start..end]` describes the list of parameters
to the function.

Point being, you only have to pay for the extra bytes if the AST actually
requires it. There's no limit to the number of different AST tag encodings.

Results of parsing only:

  * :heavy_check_mark: 15% fewer cache-misses
  * :heavy_check_mark: 28% fewer total instructions executed
  * :heavy_check_mark: 26% fewer total CPU cycles
  * :heavy_check_mark: 22% faster wall clock time

Checklist before this can actually be merged:

  * [x] parser
  * [x] render (zig fmt)
  * [x] astgen
  * [x] translate-c

Also there are a couple common forms of AST that I did not implement yet:
 * [x] FnCall with 1 or 0 args
 * [x] Builtin call with 0, 1, or 2 args
 * [x] block with 0, 1, or 2 statements
 * [x] container decl with 0, 1, or 2 fields/decls
 * [x] `extern fn foo() void;` should be a FnProto, not FnDecl


Vexu:
Could you make the Token and Node Tags snake_case now while you're breaking the API anyways?

andrewrk:
Yep

lib/std/multi_array_list.zig
```diff
@@ -0,0 +1,353 @@
...
+        bytes: [*]align(@alignOf(S)) u8 = undefined,
+        len: usize = 0,
+        capacity: usize = 0,
+
+        pub const Elem = S;
+
+        pub const Field = meta.FieldEnum(S);
+
+        pub const Slice = struct {
+            /// The index corresponds to sizes.bytes, not in field order.
```
ifreund, line 23: This comment seems to be incorrect, the `ptrs` array is indeed in field order.

lib/std/multi_array_list.zig
```diff
@@ -0,0 +1,353 @@
...
+                other.deinit(gpa);
+                self.* = undefined;
+            }
+        };
+
+        const Self = @This();
+
+        const fields = meta.fields(S);
+        /// `sizes.bytes` is an array of @sizeOf each S field. Sorted by alignment, descending.
+        /// `sizes.indexes` is an array mapping from field to its index in the `sizes.bytes` array.
```
ifreund, line 60: `sizes.indexes` is never used, we should probably get rid of it.

ifreund (commented):
I like the MultiArrayList implementation with a single allocation. It seems the main disadvantage compared to a version with multiple allocations is that shrinking requires copying memory around and the toOwnedSlice() API is a little less useful because of that.

andrewrk:
Heads up to @alexnask I think this is going to be a major breaking change to ZLS that may require quite some effort to update. Sorry about that. The good news, however, is that it should in theory improve perf & memory usage of ZLS.

trishume:
One thing you might want to consider is writing a quick benchmark of a simple traversal over the whole AST. I would expect this change to make producing the AST faster but it might make traversing slower due to potentially taking misses on 3 cache lines when visiting a node instead of one. Similarly if you implement the two types of function calls you might add more branch mispredictions.

It would also maybe help double-check the ergonomics of traversal. Although maybe you can just wait to port `zig fmt` to do all of this since presumably it's a tree traversal.

Avokadoen:
Shouldn't the Allocator arguments in multi_array_list.zig simply be called "allocator" and not "gpa"?

andrewrk:
> Shouldn't the Allocator arguments in multi_array_list.zig simply be called "allocator" and not "gpa"?

I've been experimenting with a new convention: if the parameter is named `gpa` it indicates the allocator will be used in a way that would be better suited for a general purpose allocator - e.g. one that reclaims memory on free(). It's not a requirement, but it's a hint to callsites that, if they have the choice, a gpa would be preferable to an arena.

pixelherodev:
I have no performance numbers for this, but here's a proposal that should reduce memory usage to a static 8 + 4N per node, where N is the number of children.

**Motivation**

Currently, at n=0 8 bytes are wasted: lhs, and rhs. At n=1, 4 bytes are wasted. At n=2, none are. At n=3 and above, we now waste at least 8 bytes per node due to indirection. For n in [0,2] we use 13 bytes. For n=[3,RAM limit], we use 21 + 4n. With this proposal, all of these cases drop down to 8 + 4n. For n=0, that's five bytes of savings. For n=1, it's one. For n=2, we lose three bytes. For n=three and above, we save a whopping *thirteen bytes per node*. While the n=2 loss is not ideal, you would have to have a minimum of thirteen binary nodes for every n>2-node plus five binary nodes for each terminal node and 1 binary nodes for each unary node in order for it to not be a net gain. You would need even more binary nodes for it to be a net loss. As such, this would almost certainly save memory in every real use case.

**Compatibility with existing PR**

This should support all three stated goals. It does not conflict with any of the insights listed.

**Changes**

Currently, we have tag, main_token, lhs, and rhs, for a "static" 13 bytes. For n>2, we then replace rhs with an index into extra_data, which itself contains two *more* indices, plus the remaining children - that's *8 more* bytes per node, or 21+4n.

We can't just remove the extra data though, we need a way to have more than 2 children! So, we *also* remove lhs and rhs. In their place, we add a u32 for the number of children. This next bit may be familiar to anyone who's familiar with SPIR-V, since it's directly inspired by it after a rather lengthy discussion or six with DarkUranium spanning a few months. We currently use a struct of arrays with Node[]s - in its place, we use a []u32. Each index into this array is either a node *or* an index. Each node is followed by kid_count indices, each of which is the index of a node within the nodes array.

This has a couple of impacts. First, we no longer need to store indices to bound the children list. Second, we no longer need an extra array list here at all! Third, assuming an even distribution of node sizes with an average of 2 children, we save an average of 4 or 5 bytes per node - more or less depending on the concentration of binary nodes. This should also improve data locality, which alongside the smaller sizes should be a major win for raw performance. I have not performed any benchmarks, however, so I cannot say how this would compare to the current state of the PR.

pixelherodev:
@ifreund reports the following node usage numbers for parsing the full standard library: 766932, 89277, 264456, 123193 for n = 0, n= 1, n = 2, and n > 2 respectively.

Given the flat size of 13 bytes vs 8+4N for the first three numbers, my proposal would save >3.8MB for terminal node, 89KB for single-child nodes, and it loses roughly 800KB on binary nodes. However, since we save 13 bytes on n>2 nodes, we get back another >1.6MB from them. For the standard library, we thus save 3.8+1.6+0.09-0.8 = ~4.7MB total.

I'm not sure what the numbers would be for CPU usage. It's possible that data locality would be worsened, that it will require more CPU time to traverse, and so on. I plan on using it regardless, for my compiler, as the reduced RAM usage is worthwhile for my compiler even if it makes it a little bit slower, and I think that the 8+4N scheme is simpler and better fits my goals, but I think it's definitely worth measuring for stage2, which has low enough RAM usage that it might be worth using those extra 5MiB if it makes it a bit faster, and doesn't have the same philosophical emphasis on simplicity I am employing in mine.

pixelherodev:
More numbers:

Stage2: { 90868, 55985, 70933, 15572 }. 90868 * 5 + 55985 - 70933 * 3 + 15572 * 13 = ~500KB of savings.
River: { 8592, 8046, 8249, 1440 } 8592 * 5 + 8046  - 8249 * 3 + 1440 * 13 = ~45KB of savings.
ZLS:  { 10859, 7329, 8338, 2321 } 10859 * 5 + 7329 - 8338 * 3 + 2321 * 12 = ~65KB of savings.

In each case, the n>2 nodes *alone* more than make up for the binary nodes. In each case, the terminal nodes *alone* more than make up for the binary nodes. The n=1 nodes are a nice bonus on top, but between terminal nodes and nodes with more than two children, every single test case has significant savings.

While there are more binary nodes than there are, say, n=3 nodes, we save 13 bytes on all nodes with n *> 2*, not just n = 3. Put them all together and it adds up. There's also more terminal nodes than binary nodes in every case, which makes a lot of sense if you think about it. With `a * b`, you have a binary node for the operation, *and a terminal node for each argument*.

andrewrk:
Here's a `zig fmt` benchmark for this branch. Note that `zig fmt` is **not** the primary goal of the perf improvements of this branch. (Primary goal is the `parse -> astgen -> sema -> codegen` pipeline)

### master branch

Peak memory used: 37 MiB

```
[nix-shell:~/dev/zig/build-release]$ perf stat -e cache-misses -e instructions -e cycles ./zig fmt ~/tmp/std/fs.zig ~/tmp/std/os.zig ~/tmp/std/build.zig ~/tmp/std/json.zig

 Performance counter stats for './zig fmt /home/andy/tmp/std/fs.zig /home/andy/tmp/std/os.zig /home/andy/tmp/std/build.zig /home/andy/tmp/std/json.zig':

           112,180      cache-misses:u                                              
       109,741,573      instructions:u            #    2.08  insn per cycle         
        52,729,086      cycles:u                                                    

       0.015755514 seconds time elapsed

       0.012879000 seconds user
       0.002972000 seconds sys
```

### ast-memory-layout branch

Peak memory used: 38 MiB

```
[nix-shell:~/Downloads/zig/build-release]$ perf stat -e cache-misses -e instructions -e cycles ./zig fmt ~/tmp/std/fs.zig ~/tmp/std/os.zig ~/tmp/std/build.zig ~/tmp/std/json.zig 

 Performance counter stats for './zig fmt /home/andy/tmp/std/fs.zig /home/andy/tmp/std/os.zig /home/andy/tmp/std/build.zig /home/andy/tmp/std/json.zig':

           112,807      cache-misses:u                                              
        95,678,941      instructions:u            #    2.02  insn per cycle         
        47,307,891      cycles:u                                                    

       0.013779576 seconds time elapsed

       0.010836000 seconds user
       0.002955000 seconds sys

```

### Deltas:
 * :x:  0.5% worse cache-misses
 * :heavy_check_mark: 12.8% better instruction count
 * :heavy_check_mark: 10.3% better CPU cycle count
 * :heavy_check_mark: 12.5% better wall clock time

marler8997:
WTF why have I never heard of `perf-stat`!

ehaas:
translate-c regression: there seems to be some sort of interaction going on between function forward declarations + structs + multiple functions using the struct. This is the minimal example I found; it works properly with the previous translate-c:

```c
void bar(void);
void baz(void);
struct foo { int x; };
void bar() {
	struct foo tmp;
}

void baz() {
	struct foo tmp;
}

int main(void) {
	bar();
	baz();
	return 0;
}
```

That C code will result in `struct_foo` being defined within `fn bar()`, and then trying to reference it in `baz` and not finding it. The bug does not happen if any of the following are true:
1. `foo` and `bar` function forward declarations are not present
2. `struct foo` is defined before the function forward declarations
3. `foo` or `bar` not defined & used

andrewrk:
Thanks for finding & reporting these regressions @ehaas!

andrewrk:
Here is a translate-c benchmark. Again, keep in mind this is **not** the primary goal of the perf improvements of this branch. (Primary goal is the parse -> astgen -> sema -> codegen pipeline). C header files tend to change infrequently and the caching mechanism works well, so the cost of waiting for translate-c is paid infrequently compared to the main pipeline cost.

#### test.c

```c
#include <windows.h>
```

### master branch

 * peak memory used: 170 MiB
 * less maintainable; no intermediate AST

```
[nix-shell:~/dev/zig/build-release]$ perf stat -e cache-misses -e instructions -e cycles ./zig translate-c test.c -target x86_64-windows-gnu -lc >windows_h.zig

 Performance counter stats for './zig translate-c test.c -target x86_64-windows-gnu -lc':

         5,242,850      cache-misses:u                                              
    14,819,952,615      instructions:u            #    2.58  insn per cycle         
     5,734,684,894      cycles:u                                                    

       1.398644064 seconds time elapsed

       1.288349000 seconds user
       0.112204000 seconds sys
```

### ast-memory-layout branch

 * peak memory used: 138 MiB
 * more maintainable with intermediate AST

```
[nix-shell:~/Downloads/zig/build-release]$ perf stat -e cache-misses -e instructions -e cycles ./zig translate-c test.c -target x86_64-windows-gnu -lc >windows_h.zig

 Performance counter stats for './zig translate-c test.c -target x86_64-windows-gnu -lc':

         4,013,404      cache-misses:u                                              
    14,896,484,379      instructions:u            #    2.20  insn per cycle         
     6,763,628,217      cycles:u                                                    

       1.571640049 seconds time elapsed

       1.534413000 seconds user
       0.039061000 seconds sys
```

### deltas

 * :heavy_check_mark: 18.4% less peak memory usage
 * :heavy_check_mark: significantly more maintainable implementation with intermediate AST
 * :heavy_check_mark: 23.4% fewer cache misses
 * :x: 0.5% more CPU instructions
 * :x: 17.9% more CPU cycles
 * :x: 12.4% slower wall clock time

ifreund:
> Could you make the Token and Node Tags snake_case now while you're breaking the API anyways?

Quick note so we don't forget: the fields of `ast.Error` are still TitleCase while they should technically be snake_case as well now.

Vexu:
> Quick note so we don't forget: the fields of `ast.Error` are still TitleCase while they should technically be snake_case as well now.

I didn't make them snake_case since I saw some `TODO redo ast errors` comments which suggested that it might be wasted effort.

andrewrk:
I'll take a pass at cleaning up the parse errors. Also I noticed this, another todo item on the merge checklist:

```zig
                    if (mods.bit_range_start != 0) {
                        @panic("TODO implement this error");
                        //try p.warn(.{
                        //    .BitRangeInvalid = .{ .node = mods.bit_range_start },
                        //});
                    }
```

andrewrk:
> I'll take a pass at cleaning up the parse errors.

Done

dimenus:
I'm seeing some significantly different results from @andrewrk. Unfortunately, it seems to hinge a ton on whether or not I use `-march-native`. With it, I notice large differences between the two branches, but without it's much smaller. This could be unrelated to the PR though, as my CPU is very new and the native code may not be very mature yet.

lcspu: https://clbin.com/mpjRw

master (native):
```

 Performance counter stats for 'zig translate-c test.c -target x86_64-windows-gnu -lc':

        43,663,031      cache-misses:u                                              
    14,941,241,196      instructions:u            #    2.27  insn per cycle         
     6,593,848,742      cycles:u                                                    

       1.409600902 seconds time elapsed

       1.299786000 seconds user
       0.109939000 seconds sys

```

ast-memory-layout (native):
```
 Performance counter stats for 'zig translate-c test.c -target x86_64-windows-gnu -lc':

        41,162,962      cache-misses:u                                              
    15,019,263,658      instructions:u            #    3.78  insn per cycle         
     3,970,088,190      cycles:u                                                    

       0.860989735 seconds time elapsed

       0.838967000 seconds user
       0.023293000 seconds sys
```

master (default):
```

 Performance counter stats for 'zig translate-c test.c -target x86_64-windows-gnu -lc':

        42,981,948      cache-misses:u                                              
    14,942,615,406      instructions:u            #    3.63  insn per cycle         
     4,111,750,403      cycles:u                                                    

       0.941302617 seconds time elapsed

       0.834254000 seconds user
       0.106729000 seconds sys
```

ast-memory-layout(default):
```
 Performance counter stats for 'zig translate-c test.c -target x86_64-windows-gnu -lc':

        40,968,031      cache-misses:u                                              
    15,018,285,826      instructions:u            #    3.78  insn per cycle         
     3,976,343,934      cycles:u                                                    

       0.847089230 seconds time elapsed

       0.811185000 seconds user
       0.036692000 seconds sys
 ```
 
 - [x] 4.6% reduction in cache misses
 - [ ] 0.5% increase in CPU instructions
 - [x] 3.3% reduction in CPU cycles
 - [x] 10% quicker wall clock time.

**Comment by andrewrk** [comment]
All stage2 tests are passing again in this branch.

Remaining checklist for this branch:
 * get the rest of the zig fmt test cases passing
   - re-enable the translate-c test case that is blocking on this
 * implement the 2 `@panic(TODO)`'s in parse.zig
 * use fn_proto not fn_decl for extern function declarations

----

Another benchmark incoming. This one is for the main pipeline! Keep in mind, however, that this branch only has only affected tokenization, parsing, and astgen, and so far ZIR and TZIR are untouched.

Also, 100% of the code in this benchmark is actually used. There are some easy wins in doing less work parsing for all the top level declarations that don't end up getting used in a given compilation, and those wins do not apply here.

#### stage2-hello.zig

This is a silly smorgasbord of all the supported, working features of stage2. 

```zig
export fn _start() noreturn {
    assert(frob(3, 4) == 791);
    check_those_globals(aa, bb);
    while_loop();
    print();
    print();
    assert(local_mutable_vars(3, 4) == 7);
    assert(local_mutable_vars(20, 10) == 30);
    print();
    print();
    for_loop();
    assert(frob2(3, 4) == 791);
    assert(frob2(4, 3) == 79);
    optionals();
    exit();
}

const aa = 'ぁ';
const bb = '\x03';

fn check_those_globals(a: u32, b: u32) void {
    assert(a + b == 12356);
}

fn for_loop() void {
    for ("a much longer string") |_| print();
}

fn while_loop() void {
    var i: u32 = 0;
    while (i < 4) : (i += 1) print();
    assert(i == 4);
}

fn optionals() void {
    const a: u32 = 2;
    const b: ?u32 = a;
    const c = b.?;
    if (c != 2) unreachable;
}

fn frob(a: u32, b: u32) u32 {
    const x: u32 = blk: {
        const c = a + b; // 7
        const d = a + c; // 10
        const e = d + b; // 14
        const f = d + e; // 24
        const g = e + f; // 38
        const h = f + g; // 62
        const i = g + h; // 100
        const j = i + d; // 110
        const k = i + j; // 210
        const l = k + c; // 217
        const m = l + d; // 227
        const n = m + e; // 241
        const o = n + f; // 265
        const p = o + g; // 303
        const q = p + h; // 365
        const r = q + i; // 465
        const s = r + j; // 575
        const t = s + k; // 785
        break :blk t;
    };
    const y = x + a; // 788
    const z = y + a; // 791
    return z;
}

fn assert(ok: bool) void {
}

fn print() void {
    asm volatile ("syscall"
        :
        : [number] "{rax}" (1),
          [arg1] "{rdi}" (1),
          [arg2] "{rsi}" (@ptrToInt("Hello, World!\n")),
          [arg3] "{rdx}" (14)
        : "rcx", "r11", "memory"
    );
    return;
}

fn exit() noreturn {
    asm volatile ("syscall"
        :
        : [number] "{rax}" (231),
          [arg1] "{rdi}" (0)
        : "rcx", "r11", "memory"
    );
    unreachable;
}

fn frob2(a: u32, b: u32) u32 {
    const x: u32 = if (a < b) blk: {
        const c = a + b; // 7
        const d = a + c; // 10
        const e = d + b; // 14
        const f = d + e; // 24
        const g = e + f; // 38
        const h = f + g; // 62
        const i = g + h; // 100
        const j = i + d; // 110
        const k = i + j; // 210
        const l = k + c; // 217
        const m = l + d; // 227
        const n = m + e; // 241
        const o = n + f; // 265
        const p = o + g; // 303
        const q = p + h; // 365
        const r = q + i; // 465
        const s = r + j; // 575
        const t = s + k; // 785
        break :blk t;
    } else blk: {
        const t = b + b + a; // 10
        const c = a + t; // 14
        const d = c + t; // 24
        const e = d + t; // 34
        const f = e + t; // 44
        const g = f + t; // 54
        const h = c + g; // 68
        break :blk h + b; // 71
    };
    const y = x + a; // 788, 75
    const z = y + a; // 791, 79
    return z;
}

fn local_mutable_vars(a: u32, b: u32) u32 {
    var x: u32 = undefined;
    x = 0;
    x += a;
    x += b;
    return x;
}
```

### master branch

 * peak memory usage: 3.7 MiB

```
[nix-shell:~/dev/zig/build-release]$ perf stat -e cache-misses -e instructions -e cycles ./zig-cache/bin/zig build-exe ~/tmp/stage2-hello.zig

 Performance counter stats for './zig-cache/bin/zig build-exe /home/andy/tmp/stage2-hello.zig':

            81,640      cache-misses                                                
         7,025,847      instructions              #    0.81  insn per cycle         
         8,726,948      cycles                                                      

       0.002105141 seconds time elapsed

       0.000000000 seconds user
       0.003028000 seconds sys
```

### ast-memory-layout branch

 * peak memory usage: 3.6 MiB

```
[nix-shell:~/Downloads/zig/build-release]$ perf stat -e cache-misses -e instructions -e cycles ./zig-cache/bin/zig build-exe ~/tmp/stage2-hello.zig

 Performance counter stats for './zig-cache/bin/zig build-exe /home/andy/tmp/stage2-hello.zig':

            78,119      cache-misses                                                
         6,346,154      instructions              #    0.86  insn per cycle         
         7,370,489      cycles                                                      

       0.001997548 seconds time elapsed

       0.001888000 seconds user
       0.001888000 seconds sys
```

* :heavy_check_mark: 1.8% less peak memory usage
* :heavy_check_mark: 4.3% fewer cache misses
* :heavy_check_mark: 9.7% fewer instructions
* :heavy_check_mark: 15.5% fewer cpu cycles
* :heavy_check_mark: 5.1% faster wall clock time

andrewrk:
And now for everybody's favorite silly benchmark, printing hello world with 1 million print() function calls! 

#### print1mil.zig

```zig
export fn _start() noreturn {
    print(); // .... repeated 1 million times
    exit();
}

fn print() void {
    asm volatile ("syscall"
        :
        : [number] "{rax}" (1),
          [arg1] "{rdi}" (1),
          [arg2] "{rsi}" (@ptrToInt("Hello, world!\n")),
          [arg3] "{rdx}" (14)
        : "rcx", "r11", "memory"
    );
    return;
}

fn exit() noreturn {
    asm volatile ("syscall"
        :
        : [number] "{rax}" (231),
          [arg1] "{rdi}" (0)
        : "rcx", "r11", "memory"
    );
    unreachable;
}
```

master branch requires the following patch:

```diff
diff --git a/src/link/Elf.zig b/src/link/Elf.zig
index 18f3f5771..98f874a4b 100644
--- a/src/link/Elf.zig
+++ b/src/link/Elf.zig
@@ -3025,7 +3025,7 @@ const min_nop_size = 2;
 
 /// Writes to the file a buffer, prefixed and suffixed by the specified number of
 /// bytes of NOPs. Asserts each padding size is at least `min_nop_size` and total padding bytes
-/// are less than 126,976 bytes (if this limit is ever reached, this function can be
+/// are less than 1044480 bytes (if this limit is ever reached, this function can be
 /// improved to make more than one pwritev call, or the limit can be raised by a fixed
 /// amount by increasing the length of `vecs`).
 fn pwriteDbgLineNops(
@@ -3040,7 +3040,7 @@ fn pwriteDbgLineNops(
 
     const page_of_nops = [1]u8{DW.LNS_negate_stmt} ** 4096;
     const three_byte_nop = [3]u8{ DW.LNS_advance_pc, 0b1000_0000, 0 };
-    var vecs: [32]std.os.iovec_const = undefined;
+    var vecs: [256]std.os.iovec_const = undefined;
     var vec_index: usize = 0;
     {
         var padding_left = prev_padding_size;
```

### master branch

* peak memory usage: 663 MiB

```
[nix-shell:~/dev/zig/build-release]$ perf stat -e cache-misses -e instructions -e cycles ./zig-cache/bin/zig build-exe ~/tmp/print1mil.zig 

 Performance counter stats for './zig-cache/bin/zig build-exe /home/andy/tmp/print1mil.zig':

        76,946,085      cache-misses                                                
     9,047,131,386      instructions              #    1.69  insn per cycle         
     5,369,051,109      cycles                                                      

       1.233985278 seconds time elapsed

       0.994124000 seconds user
       0.241272000 seconds sys
```

### ast-memory-layout branch

* peak memory usage: 577 MiB

```
[nix-shell:~/Downloads/zig/build-release]$ perf stat -e cache-misses -e instructions -e cycles ./zig-cache/bin/zig build-exe ~/tmp/print1mil.zig 

 Performance counter stats for './zig-cache/bin/zig build-exe /home/andy/tmp/print1mil.zig':

        69,254,254      cache-misses                                                
     7,381,716,419      instructions              #    1.81  insn per cycle         
     4,073,661,224      cycles                                                      

       0.937389499 seconds time elapsed

       0.722121000 seconds user
       0.216335000 seconds sys
```

* :heavy_check_mark: 12.9% less peak memory usage
* :heavy_check_mark: 10.0% fewer cache misses
* :heavy_check_mark: 18.4% fewer instructions
* :heavy_check_mark: 24.1% fewer cpu cycles
* :heavy_check_mark: 24.0% faster wall clock time

This benchmark is really contrived, but it's a fun one that has been circling in the programming languages blogosphere, so I am amused that we are now achieving (on my laptop) 1.07 million lines per second on it. 

To any other compiler devs who are looking to compare numbers, note that this time measures the entire execution of the compiler, until the output x86_64 ELF file is complete and flushed to disk and can be executed by the user. Also it has proper DWARF debug info and line numbers. Did I mention we support incremental compilation & linking?

I expect these numbers to drop even more once we apply these same principles to ZIR and TZIR in a future branch.

Vexu:
Re: 2da2123128dfb78893109ad195fa536ff758f2ec
This was a regression test against [an infinite loop that commonly appeared when using zls](https://github.com/zigtools/zls/pull/22#issuecomment-628736294), should probably have mentioned that.

Re: 1247b7b9ef10d1b16117615bbd64d7bbd27507b2
This was about the `error.Foo`syntax, master branch was able to continue to the invalid and.

Also seems like the error for missing for loop payload is gone:
```zig
fn foo() void {
    for (a) {}
}
```

andrewrk:
Thanks for the insight - I'll have a second look into these test cases.

ehaas:
translate-c regression: switch statement with no `default` produces `error: break expression outside loop`:

```c
switch (1) {
    case 1: break;
}
```

```zig
switch (@as(c_int, 1)) {
    @as(c_int, 1) => {},
    else => break,
}
```

Previous version of translate-c produced:
```zig
    @"switch": {
        case: {
            switch (@as(c_int, 1)) {
                @as(c_int, 1) => break :case,
                else => break :@"switch",
            }
        }
        break :@"switch";
    }
```

ehaas:
translate-c regression - `switch` with a `break` inside of a `for` loop can produce `error: unreachable code` (the `else` statement in the translated switch). The break is being applied to the loop instead of the switch:

<details>
<summary>click for C code</summary>

```c
int main(void) {
   int x = 0;
   for (;;) {
      switch (x) {
         case 0: {
            x = 1;
            break;
         }
         default:
            return 0;
     }
  }
}
```
</details>

<details>
<summary>click for this branch translate-c output</summary>

```zig
pub export fn main() c_int {
    var x: c_int = 0;
    while (true) {
        switch (x) {
            @as(c_int, 0) => {
                {
                    x = 1;
                    break;
                }
                return 0;
            },
            else => return 0,
        }
    }
    return 0;
}
```
</details>

<details>
<summary>click for master branch translate-c output</summary>

```zig
pub export fn main() c_int {
    var x: c_int = 0;
    while (true) {
        @"switch": {
            default: {
                case: {
                    switch (x) {
                        @as(c_int, 0) => break :case,
                        else => break :default,
                    }
                }
                {
                    x = 1;
                    break :@"switch";
                }
            }
            return 0;
        }
    }
    return 0;
}
```
</details>

andrewrk:
@Vexu for this one, can you double check things on the translate-c end?

```zig
    cases.add("nested assignment",
        \\int foo(int *p, int x) {
        \\    return *p++ = x;
        \\}
    , &[_][]const u8{
        \\pub export fn foo(arg_p: [*c]c_int, arg_x: c_int) c_int {
        \\    var p = arg_p;
        \\    var x = arg_x;
        \\    return blk: {
        \\        const tmp = x;
        \\        (blk_1: {
        \\            const ref = &p;
        \\            const tmp_2 = ref.*;
        \\            ref.* += 1;
        \\            break :blk_1 tmp_2;
        \\        }).?.* = tmp;
        \\        break :blk tmp;
        \\    };
        \\}
    });
```

When I supply that expected code to zig fmt in this branch, it does not get reformatted.

andrewrk:
You know what I forgot to do way earlier in this thread was ping @LakeByTheWoods. Lachlan, there are lots of zig fmt changes happening here; now's a good time to jump in if you want to influence any adjustments to the test cases.

andrewrk:
Repeating the commit message of 1e3a200ba604c8d19fcc1119d01c2e80f2a8beca into a comment here:



This regresses the test case of `zig fmt` deleting empty line comments.
Two open questions here:

 * What should the rules be about deleting empty line comments?
   It makes sense usually, but for array initization, empty line
   comments cause a line break, affecting the row/column alignment.
   Perhaps we should therefore respect all empty line comments?
   Or should we special case array initializations?

 * If we decide to special case some kinds of line comments to respect
   them (which is status quo of master branch!), how should that be implemented in this branch?

MasterQ32:
>   Perhaps we should therefore respect all empty line comments?

This is my preferred option here, as an empty comment might be filled in later and it's a bit annoying that `zig fmt` deletes comments on *any* place.

Also, is this formatting a intended behaviour or a bug:
```zig
/// hello!
comptime { }
```
↓
```zig
comptime { }
```

Deleting documentation comments without asking is imho a serious problem, but not sure if intended or not

Vexu:
> Vexu for this one, can you double check things on the translate-c end?
> ...
> When I supply that expected code to zig fmt in this branch, it does not get reformatted.

Not sure if I understand the issue here, the formatting seems fine to me?

andrewrk:
> Also, is this formatting a intended behaviour or a bug:

I think this should be a parse error, no doc comments allowed on top level comptime. Not sure what the purpose of those would be, or how they would be represented in generated documentation.

> Not sure if I understand the issue here, the formatting seems fine to me?

Yes the formatting looks fine to me, but when enabling the test case, it fails:

```
[nix-shell:~/Downloads/zig/build-2]$ ./zig build test-translate-c test-run-translated-c

========= Expected to find: ===================
pub export fn foo(arg_p: [*c]c_int, arg_x: c_int) c_int {
    var p = arg_p;
    var x = arg_x;
    return blk: {
        const tmp = x;
        (blk_1: {
            const ref = &p;
            const tmp_2 = ref.*;
            ref.* += 1;
            break :blk_1 tmp_2;
        }).?.* = tmp;
        break :blk tmp;
    };
}
========= But file does not contain it: =======
pub usingnamespace @import("std").c.builtins;
pub export fn foo(arg_p: [*c]c_int, arg_x: c_int) c_int {
    var p = arg_p;
    var x = arg_x;
    return blk: {
        const tmp = x;
        (blk_1: {
                const ref = &p;
                const tmp_2 = ref.*;
                ref.* += 1;
                break :blk_1 tmp_2;
            }).?.* = tmp;
        break :blk tmp;
    };
}
```

ifreund:
>     * What should the rules be about deleting empty line comments?
>       It makes sense usually, but for array initization, empty line
>       comments cause a line break, affecting the row/column alignment.
>       Perhaps we should therefore respect all empty line comments?
>       Or should we special case array initializations?

I'd be against any special-casing to allow empty line comments in some situations but not others, consistency is one of the most important aspects of coding style IMO. I'd be totally fine with allowing empty line comments everywhere, the status quo of allowing them in some places but not others is not ideal and I've never fully understood where they are/are not allowed.

Note that we could continue to remove empty comments that are at the start/end of a larger comment block that is not entirely empty while leaving standalone empty comments alone.

Another option would be to force non-empty comments in order to affect array formatting, but I don't see this as ideal as it would likely cause a lot of `// foo` or similar where no comment body is desired.

>     * If we decide to special case some kinds of line comments to respect
>       them (which is status quo of master branch!), how should that be implemented in this branch?

Yeah, there's really not any good way to do that, one way would be by adding more variants to the Space enum which allow/disallow empty comments, but I'd much rather just choose a consistent strategy and not deal with it.

andrewrk:
OK, no more special handling of empty comments! :+1:

Vexu:
> Yes the formatting looks fine to me, but when enabling the test case, it fails:

That's strange, it passes just fine for me.

ifreund:
@MasterQ32 that behavior of deleting doc comments on comptime blocks no longer happens on this branch.

andrewrk:
Hmm. I'm getting this now, with the latest changes, clean build:

```
[nix-shell:~/Downloads/zig/build-2]$ ./zig build test-translate-c test-run-translated-c

========= Expected to find: ===================
pub export fn foo() void {
    if (true) {
        var a: c_int = 2;
    }
    if ((blk: {
        _ = @as(c_int, 2);
        break :blk @as(c_int, 5);
    }) != 0) {
        var a: c_int = 2;
    }
}
========= But file does not contain it: =======
pub usingnamespace @import("std").c.builtins;
pub export fn foo() void {
    if (true) {
        var a: c_int = 2;
    }
    if ((blk: {
            _ = @as(c_int, 2);
            break :blk @as(c_int, 5);
        }) != 0) {
        var a: c_int = 2;
    }
}
```

However when I run `./zig translate-c test.c` manually, with the failing test case isolated:

```c
void foo() {
    if (2) {
        int a = 2;
    }
    if (2, 5) {
        int a = 2;
    }
}
```

It produces the expected output. Also it runs clean in valgrind.

andrewrk:
Strange, with a clean build I'm getting different results:

```
[nix-shell:~/Downloads/zig/build-3]$ ./zig build test-translate-c test-run-translated-c
thread 4867 panic: reached unreachable code
/home/andy/Downloads/zig/lib/std/zig/ast.zig:871:21: 0xc540e1 in std.zig.ast.Tree.lastToken (zig1)
                    unreachable;
                    ^
/home/andy/Downloads/zig/lib/std/zig/render.zig:1863:34: 0x11fb647 in std.zig.render.renderContainerDecl (zig1)
    const rbrace = tree.lastToken(container_decl_node);
                                 ^
/home/andy/Downloads/zig/lib/std/zig/render.zig:522:39: 0x1040890 in std.zig.render.renderExpression (zig1)
            return renderContainerDecl(gpa, ais, tree, node, tree.containerDeclTwo(&buffer, node), space);
                                      ^
/home/andy/Downloads/zig/lib/std/zig/render.zig:964:25: 0x1045f67 in std.zig.render.renderVarDecl (zig1)
    try renderExpression(gpa, ais, tree, var_decl.ast.init_node, .semicolon);
                        ^
/home/andy/Downloads/zig/lib/std/zig/render.zig:144:49: 0xf4cb54 in std.zig.render.renderMember (zig1)
        .simple_var_decl => return renderVarDecl(gpa, ais, tree, tree.simpleVarDecl(decl)),
                                                ^
/home/andy/Downloads/zig/lib/std/zig/render.zig:51:25: 0xe23c2d in std.zig.render.renderMembers (zig1)
        try renderMember(gpa, ais, tree, member, .newline);
                        ^
/home/andy/Downloads/zig/lib/std/zig/render.zig:37:22: 0xd0cf3d in std.zig.render.renderTree (zig1)
    try renderMembers(buffer.allocator, ais, tree, tree.rootDecls());
                     ^
/home/andy/Downloads/zig/lib/std/zig/ast.zig:67:50: 0xbac023 in std.zig.ast.Tree.renderToArrayList (zig1)
        return @import("./render.zig").renderTree(buffer, tree);
                                                 ^
/home/andy/Downloads/zig/lib/std/zig/ast.zig:62:35: 0xb5e362 in std.zig.ast.Tree.render (zig1)
        try tree.renderToArrayList(&buffer);
                                  ^
/home/andy/Downloads/zig/src/main.zig:2203:42: 0xbc99cf in main.cmdTranslateC (zig1)
        const formatted = try tree.render(comp.gpa);
                                         ^
/home/andy/Downloads/zig/src/main.zig:1906:29: 0xb01d39 in main.buildOutputType (zig1)
        return cmdTranslateC(comp, arena, have_enable_cache);
                            ^
/home/andy/Downloads/zig/src/main.zig:184:31: 0xad8fef in main.mainArgs (zig1)
        return buildOutputType(gpa, arena, args, .translate_c);
                              ^
/home/andy/Downloads/zig/src/stage1.zig:45:24: 0xad843e in main (zig1)
                       ^
translate-c unnamed fields have predictabile names...The following command terminated unexpectedly:
/home/andy/Downloads/zig/build-3/zig translate-c -lc --enable-cache zig-cache/o/JQNr689Tho-HQrR7lf4zk_2rs_OULBZMEhO26lRm1B9mYvPOLaDGR9M9cwGWqr4d/source.h 
error: the following build command failed with exit code 6:
/home/andy/Downloads/zig/zig-cache/o/94ebfda16efd8542d4d42f805201bb53/build /home/andy/Downloads/zig/build-3/zig /home/andy/Downloads/zig /home/andy/Downloads/zig/zig-cache /home/andy/.cache/zig test-translate-c test-run-translated-c

```

Not sure what a clean build has to do with it. I can't think of anything.

andrewrk:
I can only reproduce the drone CI failure on aarch64, with `-OReleaseFast`. Here are some clues:

```diff
diff --git a/lib/std/multi_array_list.zig b/lib/std/multi_array_list.zig
index 24269c231..71dadaf54 100644
--- a/lib/std/multi_array_list.zig
+++ b/lib/std/multi_array_list.zig
@@ -228,6 +228,7 @@ pub fn MultiArrayList(comptime S: type) type {
                 if (better_capacity >= new_capacity) break;
             }
 
+            std.debug.print("MultiArrayList realloc from {d} to {d}\n", .{self.capacity, better_capacity});
             return self.setCapacity(gpa, better_capacity);
         }
 
diff --git a/lib/std/zig/parse.zig b/lib/std/zig/parse.zig
index 36da59ed8..61edc7237 100644
--- a/lib/std/zig/parse.zig
+++ b/lib/std/zig/parse.zig
@@ -28,6 +28,7 @@ pub fn parse(gpa: *Allocator, source: []const u8) Allocator.Error!Tree {
     var tokenizer = std.zig.Tokenizer.init(source);
     while (true) {
         const token = tokenizer.next();
+        std.debug.print("token[{d}] = {s} at {d}\n", .{tokens.len, @tagName(token.tag), token.loc.start});
         try tokens.append(gpa, .{
             .tag = token.tag,
             .start = @intCast(u32, token.loc.start),
@@ -49,6 +50,10 @@ pub fn parse(gpa: *Allocator, source: []const u8) Allocator.Error!Tree {
     defer parser.nodes.deinit(gpa);
     defer parser.extra_data.deinit(gpa);
 
+    for (parser.token_tags) |tag, i| {
+        std.debug.print("token[{d}] = {s} at {d}\n", .{i, @tagName(tag), parser.token_starts[i]});
+    }
+
     // Empirically, Zig source code has a 2:1 ratio of tokens to AST nodes.
     // Make sure at least 1 so we can use appendAssumeCapacity on the root node below.
     const estimated_node_count = (tokens.len + 2) / 2;
```

#### test.zig

```zig
const std = @import("std");
const ast = std.zig.ast;

var general_purpose_allocator = std.heap.GeneralPurposeAllocator(.{}){};

pub fn main() !void {
    const gpa = if (std.builtin.link_libc)
        std.heap.raw_c_allocator
    else
        &general_purpose_allocator.allocator;
    defer if (!std.builtin.link_libc) {
        _ = general_purpose_allocator.deinit();
    };
    var arena_instance = std.heap.ArenaAllocator.init(gpa);
    defer arena_instance.deinit();
    const arena = &arena_instance.allocator;

    const args = try std.process.argsAlloc(arena);

    const source_code = try std.fs.cwd().readFileAlloc(arena, args[1], std.math.maxInt(usize));

    var tree = try std.zig.parse(gpa, source_code);
    defer tree.deinit(gpa);

    for (tree.errors) |parse_error| {
        try printErrMsgToFile(gpa, parse_error, tree, args[1], std.io.getStdErr());
    }
    if (tree.errors.len != 0) {
        std.process.exit(1);
    }
    const formatted = try tree.render(gpa);
    defer gpa.free(formatted);
    if (std.mem.eql(u8, formatted, source_code)) {
        return;
    } else {
        try std.io.getStdOut().writeAll(formatted);
    }
}

fn printErrMsgToFile(
    gpa: *std.mem.Allocator,
    parse_error: ast.Error,
    tree: ast.Tree,
    path: []const u8,
    file: std.fs.File,
) !void {
    const color_on = true;
    const lok_token = parse_error.token;
    const span_first = lok_token;
    const span_last = lok_token;

    const first_token = token_starts[span_first];
    const start_loc = tree.tokenLocation(0, span_first);

    var text_buf = std.ArrayList(u8).init(gpa);
    defer text_buf.deinit();
    const writer = text_buf.writer();
    try tree.renderError(parse_error, writer);
    const text = text_buf.items;

    const stream = file.writer();
    try stream.print("{s}:{d}:{d}: error: {s}\n", .{ path, start_loc.line + 1, start_loc.column + 1, text });

    if (!color_on) return;

    // Print \r and \t as one space each so that column counts line up
    for (tree.source[start_loc.line_start..start_loc.line_end]) |byte| {
        try stream.writeByte(switch (byte) {
            '\r', '\t' => ' ',
            else => byte,
        });
    }
    try stream.writeByte('\n');
    try stream.writeByteNTimes(' ', start_loc.column);
    try stream.writeByteNTimes('^', 1);
    try stream.writeByte('\n');
}
```

#### simple.zig

```zig
const std = @import("std");

pub fn main() anyerror!void {
}
```

### output

```
[andy@arm64-test:~/zig/build]$ ./zig build-exe test.zig -OReleaseFast

[andy@arm64-test:~/zig/build]$ ./test simple.zig
MultiArrayList realloc from 0 to 20
token[0] = keyword_const at 0
token[1] = identifier at 6
token[2] = equal at 10
token[3] = builtin at 12
token[4] = l_paren at 19
token[5] = string_literal at 20
token[6] = r_paren at 25
token[7] = semicolon at 26
token[8] = keyword_pub at 29
token[9] = keyword_fn at 33
token[10] = identifier at 36
token[11] = l_paren at 40
token[12] = r_paren at 41
token[13] = identifier at 43
token[14] = bang at 51
token[15] = identifier at 52
token[16] = l_brace at 57
token[17] = identifier at 63
token[18] = period at 66
token[19] = identifier at 67
token[20] = period at 70
MultiArrayList realloc from 20 to 38
token[21] = identifier at 71
token[22] = l_paren at 75
token[23] = string_literal at 76
token[24] = comma at 113
token[25] = period at 115
token[26] = l_brace at 116
token[27] = r_brace at 117
token[28] = r_paren at 118
token[29] = semicolon at 119
token[30] = r_brace at 121
token[31] = eof at 124
token[0] = keyword_const at 0
token[1] = identifier at 6
token[2] = equal at 10
token[3] = builtin at 12
token[4] = l_paren at 19
token[5] = string_literal at 20
token[6] = r_paren at 25
token[7] = identifier at 26
token[8] = keyword_pub at 29
token[9] = keyword_fn at 33
token[10] = identifier at 36
token[11] = invalid_ampersands at 40
token[12] = r_paren at 41
token[13] = identifier at 43
token[14] = bang at 51
token[15] = identifier at 52
token[16] = l_brace at 57
token[17] = identifier at 63
token[18] = period at 66
token[19] = identifier at 67
token[20] = period at 70
token[21] = identifier at 71
token[22] = l_paren at 75
token[23] = string_literal at 76
token[24] = comma at 113
token[25] = period at 115
token[26] = l_brace at 116
token[27] = r_brace at 117
token[28] = r_paren at 118
token[29] = semicolon at 119
token[30] = r_brace at 121
token[31] = eof at 124
MultiArrayList realloc from 0 to 20
simple.zig:1:27: error: expected ';', found 'identifier'
const std = @import("std");
                          ^
simple.zig:3:12: error: expected '(', found '&&'
pub fn main() anyerror!void {
           ^

```

You can see the tokenizer produces the correct results, however once we try to iterate over the token list, after MultiArrayList reallocs, some of the token tags become incorrect. Again this is only reproducible on aarch64 with a non-Debug optimization mode, and it runs valgrind-clean.

andrewrk:
I can repro this with MultiArrayList. Put the following inside multi_array_list.zig:

```zig
test "realloc does not corrupt data" {
    const ally = std.testing.allocator;
    var list = MultiArrayList(struct {
        tag: std.zig.Token.Tag,
        start: u32,
    }){};
    defer list.deinit(ally);

    try list.ensureCapacity(ally, 20);

    try list.append(ally, .{ .tag = .keyword_const, .start = 0 });
    try list.append(ally, .{ .tag = .identifier, .start = 6 });
    try list.append(ally, .{ .tag = .equal, .start = 10 });
    try list.append(ally, .{ .tag = .builtin, .start = 12 });
    try list.append(ally, .{ .tag = .l_paren, .start = 19 });
    try list.append(ally, .{ .tag = .string_literal, .start = 20 });
    try list.append(ally, .{ .tag = .r_paren, .start = 25 });
    try list.append(ally, .{ .tag = .semicolon, .start = 26 });
    try list.append(ally, .{ .tag = .keyword_pub, .start = 29 });
    try list.append(ally, .{ .tag = .keyword_fn, .start = 33 });
    try list.append(ally, .{ .tag = .identifier, .start = 36 });
    try list.append(ally, .{ .tag = .l_paren, .start = 40 });
    try list.append(ally, .{ .tag = .r_paren, .start = 41 });
    try list.append(ally, .{ .tag = .identifier, .start = 43 });
    try list.append(ally, .{ .tag = .bang, .start = 51 });
    try list.append(ally, .{ .tag = .identifier, .start = 52 });
    try list.append(ally, .{ .tag = .l_brace, .start = 57 });
    try list.append(ally, .{ .tag = .identifier, .start = 63 });
    try list.append(ally, .{ .tag = .period, .start = 66 });
    try list.append(ally, .{ .tag = .identifier, .start = 67 });
    try list.append(ally, .{ .tag = .period, .start = 70 });
    try list.append(ally, .{ .tag = .identifier, .start = 71 });
    try list.append(ally, .{ .tag = .l_paren, .start = 75 });
    try list.append(ally, .{ .tag = .string_literal, .start = 76 });
    try list.append(ally, .{ .tag = .comma, .start = 113 });
    try list.append(ally, .{ .tag = .period, .start = 115 });
    try list.append(ally, .{ .tag = .l_brace, .start = 116 });
    try list.append(ally, .{ .tag = .r_brace, .start = 117 });
    try list.append(ally, .{ .tag = .r_paren, .start = 118 });
    try list.append(ally, .{ .tag = .semicolon, .start = 119 });
    try list.append(ally, .{ .tag = .r_brace, .start = 121 });
    try list.append(ally, .{ .tag = .eof, .start = 123 });

    const tags = list.items(.tag);
    std.testing.expectEqual(tags[1], .identifier);
    std.testing.expectEqual(tags[2], .equal);
    std.testing.expectEqual(tags[3], .builtin);
    std.testing.expectEqual(tags[4], .l_paren);
    std.testing.expectEqual(tags[5], .string_literal);
    std.testing.expectEqual(tags[6], .r_paren);
    std.testing.expectEqual(tags[7], .semicolon);
    std.testing.expectEqual(tags[8], .keyword_pub);
    std.testing.expectEqual(tags[9], .keyword_fn);
    std.testing.expectEqual(tags[10], .identifier);
    std.testing.expectEqual(tags[11], .l_paren);
    std.testing.expectEqual(tags[12], .r_paren);
    std.testing.expectEqual(tags[13], .identifier);
    std.testing.expectEqual(tags[14], .bang);
    std.testing.expectEqual(tags[15], .identifier);
    std.testing.expectEqual(tags[16], .l_brace);
    std.testing.expectEqual(tags[17], .identifier);
    std.testing.expectEqual(tags[18], .period);
    std.testing.expectEqual(tags[19], .identifier);
    std.testing.expectEqual(tags[20], .period);
    std.testing.expectEqual(tags[21], .identifier);
    std.testing.expectEqual(tags[22], .l_paren);
    std.testing.expectEqual(tags[23], .string_literal);
    std.testing.expectEqual(tags[24], .comma);
    std.testing.expectEqual(tags[25], .period);
    std.testing.expectEqual(tags[26], .l_brace);
    std.testing.expectEqual(tags[27], .r_brace);
    std.testing.expectEqual(tags[28], .r_paren);
    std.testing.expectEqual(tags[29], .semicolon);
    std.testing.expectEqual(tags[30], .r_brace);
    std.testing.expectEqual(tags[31], .eof);
}
```

then to repro it:

```
[nix-shell:~/Downloads/zig/build-2]$ ./zig test ../lib/std/multi_array_list.zig --test-filter realloc -OReleaseFast -target aarch64-linux --test-cmd qemu-aarch64 --test-cmd-bin
Test [1/1] test "realloc does not corrupt data"... thread 17112 panic: expected Tag.identifier, found Tag.semicolon
qemu: uncaught target signal 6 (Aborted) - core dumped
qemu-aarch64 ../lib/std/zig-cache/o/a77995938a4ba7e53fac293ce25d76c2/test /home/andy/Downloads/zig/build-2/zig
```

andrewrk:
*sigh* the bug is LLVM silently outputting incorrect code for the `@reduce`. This patch fixes it:

```diff
--- a/lib/std/multi_array_list.zig
+++ b/lib/std/multi_array_list.zig
@@ -263,9 +263,14 @@ pub fn MultiArrayList(comptime S: type) type {
         }
 
         fn capacityInBytes(capacity: usize) usize {
-            const sizes_vector: std.meta.Vector(sizes.bytes.len, usize) = sizes.bytes;
-            const capacity_vector = @splat(sizes.bytes.len, capacity);
-            return @reduce(.Add, capacity_vector * sizes_vector);
+            var sum: usize = 0;
+            for (sizes.bytes) |size| {
+                sum += capacity * size;
+            }
+            return sum;
+            //const sizes_vector: std.meta.Vector(sizes.bytes.len, usize) = sizes.bytes;
+            //const capacity_vector = @splat(sizes.bytes.len, capacity);
+            //return @reduce(.Add, capacity_vector * sizes_vector);
         }
 
         fn allocatedBytes(self: Self) []align(@alignOf(S)) u8 {
```

LemonBoy:
> sigh the bug is LLVM silently outputting incorrect code for the @reduce. 

Nope.
Wanna see a magic trick? Apply this patch:

```diff
diff --git a/lib/std/multi_array_list.zig b/lib/std/multi_array_list.zig
index 689105755..16bd00d4d 100644
--- a/lib/std/multi_array_list.zig
+++ b/lib/std/multi_array_list.zig
@@ -263,18 +263,9 @@ pub fn MultiArrayList(comptime S: type) type {
         }
 
         fn capacityInBytes(capacity: usize) usize {
-            // TODO move this workaround of LLVM SIMD bugs into the Zig frontend.
-            if (std.Target.current.cpu.arch == .aarch64) {
-                var sum: usize = 0;
-                for (sizes.bytes) |size| {
-                    sum += capacity * size;
-                }
-                return sum;
-            } else {
-                const sizes_vector: std.meta.Vector(sizes.bytes.len, usize) = sizes.bytes;
-                const capacity_vector = @splat(sizes.bytes.len, capacity);
-                return @reduce(.Add, capacity_vector * sizes_vector);
-            }
+            const sizes_vector: std.meta.Vector(sizes.bytes.len, usize) = sizes.bytes;
+            const capacity_vector = @splat(sizes.bytes.len, capacity);
+            return @reduce(.Add, capacity_vector * sizes_vector);
         }
 
         fn allocatedBytes(self: Self) []align(@alignOf(S)) u8 {
diff --git a/lib/std/zig/tokenizer.zig b/lib/std/zig/tokenizer.zig
index 71b8fad4a..0f0aa0046 100644
--- a/lib/std/zig/tokenizer.zig
+++ b/lib/std/zig/tokenizer.zig
@@ -74,7 +74,7 @@ pub const Token = struct {
         return keywords.get(bytes);
     }
 
-    pub const Tag = enum {
+    pub const Tag = enum(u8) {
         invalid,
         invalid_ampersands,
         invalid_periodasterisks,
```

The problem is the `Tag` type being 7bit wide, when you call `mem.copy` [here](https://github.com/ziglang/zig/blob/db4c15be507a55f63b7f19dd4cdc10717f44eb06/lib/std/multi_array_list.zig#L259) some enum values end up with a few extraneous bit set (check it out by dumping the raw content with `sliceToBytes`).

The workaround is to either
* Set `Tag` to `u8`
* Turn the `mem.copy` with `field_info.field_type` into a `mem.copy` on `u8`, copying the whole block of raw data on reallocation

andrewrk:
> Nope.
> Wanna see a magic trick? Apply this patch:

Wow, thanks for pointing this out. That's a footgun if I've ever seen one. Wait so this has nothing to do with SIMD vectors, instead it has to do with using `std.mem.copy` on a u7? I do not yet understand why this patch would change anything. The ABI size of u7 is 1 byte, so writing a u7 should clear the extra bit.

LemonBoy:
Long story short, the optimizations are kicking in and are vectorizing (exceptionally well, I must add) the `mem.copy` loop by turning the single `i7` loads into `<4 x i7>` ones (and coalescing more than a single read/write when possible, the aarch64 asm accelerates the copy by reading/writing 8 bytes at once).

Now, the problem with LLVM's vector is that they are tight, there's no padding in between the elements, unlike in a array...
You probably understand the problem now, the generated aarch64 code is slicing and dicing the bits in the wrong way.

andrewrk:
> the generated aarch64 code is slicing and dicing the bits in the wrong way.

Ah, I see. So it is an LLVM bug in vectorization, but not with `@reduce`. Is that your conclusion? Or have we made some horrific language design flaw?

LemonBoy:
> Ah, I see. So it is an LLVM bug in vectorization, but not with @reduce. Is that your conclusion?

Yes, the `@reduce` thing was a red herring, all it did was shift the code around to hide the problem.
It remains to be seen if the problem is limited to the loop vectorizer or is extended to other regular loads/stores.

andrewrk:
Benchmark for running `zig fmt` on the whole std lib:

### master branch

 * peak memory usage: 84.7 MiB
 * zig executable binary size (without `-Denable-llvm`): 3.37 MiB

```
[nix-shell:~/dev/zig/build-release]$ perf stat -e cache-misses -e instructions -e cycles ./zig fmt ~/tmp/std

 Performance counter stats for './zig fmt /home/andy/tmp/std':

         2,803,440      cache-misses                                                
     4,592,565,483      instructions              #    2.70  insn per cycle         
     1,702,969,462      cycles                                                      

       0.389872088 seconds time elapsed

       0.370892000 seconds user
       0.018994000 seconds sys
```

### ast-memory-layout branch

 * peak memory usage: 68.4 MiB
 * zig executable binary size (without `-Denable-llvm`): 3.17 MiB

```
[nix-shell:~/Downloads/zig/build-release]$ perf stat -e cache-misses -e instructions -e cycles ./zig fmt ~/tmp/std2

 Performance counter stats for './zig fmt /home/andy/tmp/std2':

         3,771,657      cache-misses                                                
     3,941,606,705      instructions              #    2.62  insn per cycle         
     1,502,290,493      cycles                                                      

       0.345329438 seconds time elapsed

       0.332314000 seconds user
       0.013012000 seconds sys
```

### Deltas:

 * :heavy_check_mark: 19.3% less peak memory usage
 * :heavy_check_mark: 10.8% reduction in binary size
 * :x: 34.5% more cache-misses
 * :heavy_check_mark: 14.2% fewer instructions
 * :heavy_check_mark: 11.8% fewer CPU cycles
 * :heavy_check_mark: 11.4% faster wall clock time

Keep in mind, `zig fmt` does a lot of extra work that `astgen` from the main pipeline does not have to do. So I think these are very promising results!

LemonBoy:
Link to [the LLVM patch](https://reviews.llvm.org/D97465), keep all your fingers crossed and hopefully it'll make into LLVM 12

Issue #5497, opened 2020-05-31
Struct with comptime field cannot be initiated


Vexu:
```zig
var things = struct {
    comptime foo: i8 = 1,
} {}; // cannot assign to constant
const T = struct {
    comptime foo: i8 = 1,
};
var things = T{}; // cannot assign to constant

var things: struct {
    comptime foo: i8 = 1,
} = .{}; // TODO: type coercion of anon struct literal to struct

var things: struct {
    comptime foo: i8 = 1,
} = undefined; // Segmentation fault at address 0x0

test "" {
    @compileLog(@TypeOf(&things.foo));
}
```

frmdstryr:
What is creating a `var` of a struct with 0  size supposed to do? Any modification is a buffer overflow.

Vexu:
> What is creating a `var` of a struct with 0 size supposed to do? Any modification is a buffer overflow.

It does the same as creating a `var` of `u0` or `void`. Modifications are allowed as long as you set them to the one possible value.
```zig
test "" {
    const S = struct {};
    var s = S{};
    s = S{};
    var a: u0 = 0;
    a += 0;
}
```
See https://ziglang.org/documentation/master/#Zero-Bit-Types for more information.

frmdstryr:
Interesting... thanks for pointing me to the relevant docs (and sorry for the snark comment). 

I didn't know comptime fields a thing in zig...  so would this allow defining types without a function?

```zig
const T = struct {
    comptime foo: i8 = 1,
};
```

Would be the same as this?

```zig

fn Foo(value: i8) type {
   return struct {
       const foo = value;
   };
}
const T = Foo(1);
```

Or what is the use case for a comptime field?

tgschultz:
I believe I have run into this same issue in structs that have comptime fields in addition to normal fields. They *require* that you initialize the comptime field to exactly its default value, but under certain circumstances doing so will throw `cannot assign to constant`, leaving you in a "damned if you do, damned if you don't" situation.

PR #7922, opened 2021-01-31
std.json support for comptime fields

### LINKED ISSUE #6231: JSON parse into union with comptime struct fields
I'm looking to use the std lib JSON parser to parse into a union (of structs). The JSON in question has a field called "type" with different values (e.g. "number", "string") that I want to use to switch which particular union value gets created.

For example, let's say I have a JSON list of:

```json
[{"type": "number", "default": 10}, {"type": "string", "default": "hello world"}]
```

and I define my union as

```zig
pub const ThingType = enum {
    String,
    Number,
};

pub const Thing = union(ThingType) {
    Number: ThingNumber,
    String: ThingString,
};

pub const ThingNumber = struct {
    comptime @"type": []const u8 = "int",
    default: ?i32,
};

pub const ThingString = struct {
    comptime @"type": []const u8 = "string",
    default: ?[]u8,
};

```

I would like to end up with a list of

```
[<ThingNumber>, <ThingString>]
```

It makes sense to me here to use a comptime struct field, e.g. `comptime @"type": []const u8 = "number",` in each struct that the union can contain, and for the parser to be able to use that information.

However, the parser gets upset here https://github.com/ziglang/zig/blob/master/lib/std/json.zig#L1486 and https://github.com/ziglang/zig/blob/master/lib/std/json.zig#L1500 with `error: cannot assign to constant` which makes sense as I don't see it taking into consideration that the field would be const.

For this to work I'm guessing, the parser would have to skip the `comptime` / `const` fields. However, I also can't see anything in, e.g., the `TypeInfo` union that would allow some additional logic to skip said fields.

Moreover, if those fields were skipped, even then I'm not entirely sure if the parser would parse into the appropriate struct.

daurnimator:
Closes #6231

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1374,6 +1374,65 @@ test "Value.jsonStringify" {
     }
 }
 
+/// parse tokens from a stream, returning `false` if they do not decode to `value`
+fn parsesTo(comptime T: type, value: T, tokens: *TokenStream, options: ParseOptions) !bool {
+    // TODO: should be able to write this function to not require an allocator
+    const tmp = try parse(T, tokens, options);
+    defer parseFree(T, tmp, options);
+
+    return parsedEqual(tmp, value);
+}
+
+/// Returns if a value returned by `parse` is deep-equal to another value
+fn parsedEqual(a: anytype, b: @TypeOf(a)) bool {
+    switch (@typeInfo(@TypeOf(a))) {
+        .Optional => {
+            if (a == null and b == null) return true;
+            if (a == null or b == null) return false;
+            return parsedEqual(a.?, b.?);
+        },
+        .Union => |unionInfo| {
+            if (info.tag_type) |UnionTag| {
+                const tag_a = std.meta.activeTag(a);
+                const tag_b = std.meta.activeTag(b);
+                if (tag_a != tag_b) return false;
+
+                inline for (info.fields) |field_info| {
+                    if (@field(UnionTag, field_info.name) == tag_a) {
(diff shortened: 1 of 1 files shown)


lib/std/json.zig
```diff
@@ -1789,6 +1799,19 @@ test "parseFree descends into tagged union" {
     testing.expectEqual(@as(usize, 1), fail_alloc.deallocations);
 }
 
+test "parse with comptime field" {
+    const T = struct {
+        comptime a: i32 = 0,
+        b: bool,
+    };
+    testing.expectEqual(T{ .a = 0, .b = true }, try parse(T, &TokenStream.init(
```
daurnimator, line 1807: Note for github issue cross linking: I had to tiptoe around #5497 here

lib/std/json.zig
```diff
@@ -1504,7 +1504,15 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options:
                                         parseFree(field.field_type, @field(r, field.name), options);
                                     }
                                 }
-                                @field(r, field.name) = try parse(field.field_type, tokens, options);
+                                if (field.is_comptime) {
+                                    const value = try parse(field.field_type, tokens, options);
+                                    defer parseFree(field.field_type, value, options);
+                                    if (value != @field(r, field.name)) {
```
daurnimator, line 1510: Possible future work: use other equality comparison here, e.g. might need to investigate `std.meta.eql`

daurnimator: I added a low-effort version of this now to enable @marler8997

Issue #2931, opened 2019-07-22
terminfo module in the standard library


daurnimator:
In order to make pretty terminal UIs (e.g. for #2865), you need to be able to read terminfo files.

I wrote a module that does this in lua: https://github.com/daurnimator/lua-tui/blob/master/tui/terminfo.lua which wouldn't be a bad thing to port to zig.

nektro:
disagree this is something the stdlib needs, nice target for community package

daurnimator:
> learly we didn't need it for #2865 and I didn't need it in #12079 (introducing more escape sequences) either.

https://github.com/ziglang/zig/pull/12079#discussion_r920710709

ratfactor:
I'd like to make an argument for including this. Properly doing anything with a terminal beyond printing characters (and maybe ANSI escape sequences for 16 colors) requires querying the Terminfo database. I don't like it either.

From what I can tell, all Linux distros and all BSDs (except the courageous devs at NetBSD) have ceded all terminal handling, including Terminfo parsing (and I believe the database itself!), to one library: `ncurses`. For this reason, `ncurses` is the _de facto_ standard for all text/terminal UIs (TUIs) above any other documentation or man pages that might exist. I find this situation somewhat galling and I'm glad to see that people are starting to create Terminfo parsing libraries in other languages:

* [arp242/termfo](https://github.com/arp242/termfo) in Go
* [mauke/unibilium](https://github.com/mauke/unibilium) alternative to ncurses in C
* [agauniyal/termdb](https://github.com/agauniyal/termdb) in C++
* [daurnimator/lua-tui](https://github.com/daurnimator/lua-tui/blob/master/tui/terminfo.lua) in Lua - **daurnimator's own!**
* [rust-terminfo/terminfo](https://github.com/rust-terminfo/terminfo) in Rust
* [npatrick04/terminfo](https://github.com/npatrick04/terminfo) in Common Lisp
* [spectreconsole/terminfo](https://github.com/spectreconsole/terminfo) in C#
* [docelic/terminfo](https://github.com/docelic/terminfo) in Crystal
* [zachariahreed/pyterminfo](https://github.com/zachariahreed/pyterminfo) in Python
* [urbic/scherm](https://github.com/urbic/scherm) in Perl

[...]

PR #7932, opened 2021-02-01
stage1/2: don't show color in compiler when in a "dumb" terminal

g-w1:
Difference: 
image
`zig` is from tarball, `zag` is one I built. A terminal is dumb when it has `env["TERM"] == "dumb"`. This includes acme and 9term.

Compliments #7931 
cc @pixelherodev

### DIFF
--- a/lib/std/fs/file.zig
+++ b/lib/std/fs/file.zig
@@ -184,11 +184,7 @@ pub const File = struct {
         }
         if (self.isTty()) {
             if (self.handle == os.STDOUT_FILENO or self.handle == os.STDERR_FILENO) {
-                // Use getenvC to workaround https://github.com/ziglang/zig/issues/3511
-                if (os.getenvZ("TERM")) |term| {
-                    if (std.mem.eql(u8, term, "dumb"))
-                        return false;
-                }
+                return !os.isDumbTerm();
             }
             return true;
         }
--- a/src/main.zig
+++ b/src/main.zig
@@ -2887,7 +2887,7 @@ fn printErrMsgToFile(
     color: Color,
 ) !void {
     const color_on = switch (color) {
-        .auto => file.isTty(),
+        .auto => file.isTty() and !(std.os.isDumbTerm()),
         .on => true,
         .off => false,
     };
--- a/src/stage1/errmsg.cpp
+++ b/src/stage1/errmsg.cpp
@@ -17,7 +17,8 @@ enum ErrType {
 
 static void print_err_msg_type(ErrorMsg *err, ErrColor color, ErrType err_type) {
     bool is_tty = os_stderr_tty();
-    bool use_colors = color == ErrColorOn || (color == ErrColorAuto && is_tty);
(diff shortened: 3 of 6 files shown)


lib/std/fs/file.zig
```diff
@@ -170,7 +170,13 @@ pub const File = struct {
     pub fn isTty(self: File) bool {
         return os.isatty(self.handle);
     }
-
+    pub fn isDumbTerm(self: File) bool {
```
LemonBoy, line 173: TFW `File.isDumbTerm` doesn't depend on the file parameter at all.
Isn't this `supportsAnsiEscapeCodes` (but wronger)?

Shitty API anyway, this has nothing to do with files and everything to do with the execution environment.

lib/std/fs/file.zig
```diff
@@ -170,7 +170,13 @@ pub const File = struct {
...
+        if (os.getenvZ("TERM")) |term| {
+            if (std.mem.eql(u8, term, "dumb"))
```
LemonBoy, line 175:
```suggestion
            return std.mem.eql(u8, term, "dumb");
```

src/main.zig
```diff
@@ -2887,7 +2887,7 @@ fn printErrMsgToFile(
     color: Color,
 ) !void {
     const color_on = switch (color) {
-        .auto => file.isTty(),
+        .auto => file.isTty() and !(file.isDumbTerm()),
```
LemonBoy, line 2890: Just a [heads up](https://no-color.org/).

```suggestion
        .auto => file.supportsAnsiEscapeCodes(),
```

src/stage1/os.cpp
```diff
@@ -844,6 +844,10 @@ bool os_stderr_tty(void) {
 #endif
 }
 
+bool os_is_dumb_term(void) {
+    return strcmp(getenv("TERM"), "dumb") == 0;
```
LemonBoy, line 848: If `TERM` is undefined this is UB :point_right: :sunglasses: :point_right:

src/stage1/errmsg.cpp
```diff
@@ -16,7 +16,7 @@ enum ErrType {
 };
 
 static void print_err_msg_type(ErrorMsg *err, ErrColor color, ErrType err_type) {
-    bool is_tty = os_stderr_tty();
+    bool is_tty = os_stderr_tty() && !os_is_dumb_term();
```
LemonBoy, line 19: I'm nitpicking here but `os_is_dumb_term` should affect `use_colors` and not `is_tty`.

LemonBoy (commented):
Here's a nickel kid, get yourself an editor (and a terminal) that doesn't suck :dollar:

This is a stopgap solution, the proper one would be to implement #2931.

src/main.zig
g-w1: This wont work because you will want colors on windows, but it supports no ansi escape codes.

g-w1:
Closing since this is a hack.

PR #7933, opened 2021-02-01
docs: Clarify that @field can work on declarations

squeek502:
This was very non-obvious to me, so I figured it'd be worth including in the docs.


Guigui220D:
Ah well I didnt know, thanks! I found it weird too that there was hasField and hasDecl but @decl didnt exist

PR #7741, opened 2021-01-10
stage2: add support for optionals in the LLVM backend

FireFox317:
We can now codegen optionals! This includes the following instructions:
- isnull
- isnonnull
- unwrap_optional
- brvoid

Also includes a test.

The test currently does not pass because of #7730.

EDIT: Also sneak in an extra test for loops.

Issue #6378, opened 2020-09-18
Allow the feature of O(1) bootstrapping to temporarily regress


andrewrk:
Related: #853

Right now the [bootstrap repository](https://github.com/ziglang/zig-bootstrap) can start with cmake, a C++ compiler, bash, and cross compile zig for all supported targets. It does this in a fixed number of steps, because the Zig compiler is currently in c++ code and depends on an LLVM backend.

This proposal is to allow this feature to regress for a period of time between now and 1.0, restoring the feature before 1.0.

What that regression would look like is:

 * Finish implementing self-hosted (stage2) and pass all the behavior tests, std lib tests, and able to build itself.
 * Finish the C backend.
 * Use the C backend to generate a .c implementation of the self-hosted compiler
 * Delete the stage1 c++ code and check that generated .c code into the zig source repository to make it easy to build zig. This does not count as bootstrapping because the generated .c code is not source code. It's more like committing a multi-target binary into source control. 
 * Exploit the fact that we only have 1 codebase for the zig compiler to iterate faster, solve bugs, finish the language, and approach 1.0. This is the main motivation for this proposal.
 * Do https://github.com/ziglang/zig/issues/5246#issuecomment-685172360
 * Release 1.0, with the O(1) bootstrapping feature restored.

joachimschmidt557:
@andrewrk Is it planned to have a split between stage0 and stage1 again? As in that stage1 of 1.0 is composed of

- stage0 code which was originally translated from zig to C, now maintained on its own (parsing, semantic analysis, codegen, etc.)
- stage2 code which is written in zig (zig fmt, etc.)

This would reduce the amount of C code needed to maintain.

andrewrk:
Yes and I have that working over in #6250 (with a fat checklist needed to be done before merging)

Rocknest:
So we would have stage1 autotranslated from stage2 after all, for some time at least?

marler8997:
If instead you took the C output and just copied/pushed that to the `zig-bootstrap` repo every so often, then wouldn't that mean you wouldn't have to maintain the stage1 C compiler by hand ever?

matu3ba:
Assume for this comment that c codegen "just works" and "cleanup can be done":
[...]

PR #768, opened 2018-02-09
document when to use error!type in function decl

bnoordhuis:

PR #7934, opened 2021-02-01
Stage2 cbe: optionals and errors

Vexu:
~~Testing error stuff depends on #7628.~~
No tests for orelse since the astgen is a bit broken which might be fixed by  #7741.

I upped the C version used in tests to C99 since we were already using `-Wno-declaration-after-statement` and C89 doesn't have compound literals which would have required a bunch of annoying workarounds.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1653,6 +1653,8 @@ pub fn performAllTheWork(self: *Compilation) error{ TimerUnsupported, OutOfMemor
                     .error_msg = null,
                     .decl = decl,
                     .fwd_decl = fwd_decl.toManaged(module.gpa),
+                    // we don't want to emit optionals and error unions to headers since they have no ABI
+                    .typedefs = undefined,
                 };
                 defer dg.fwd_decl.deinit();
 
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -207,10 +207,12 @@ pub const DeclGen = struct {
                 if (t.isPtrLikeOptional()) {
                     return dg.renderValue(writer, child_type, val);
                 }
+                try writer.writeByte('(');
+                try dg.renderType(writer, t);
                 if (val.tag() == .null_value) {
-                    try writer.writeAll("{ .is_null = true }");
+                    try writer.writeAll("){ .is_null = true }");
                 } else {
-                    try writer.writeAll("{ .is_null = false, .payload = ");
+                    try writer.writeAll("){ .is_null = false, .payload = ");
                     try dg.renderValue(writer, child_type, val);
                     try writer.writeAll(" }");
                 }
@@ -319,7 +321,11 @@ pub const DeclGen = struct {
                 if (t.isPtrLikeOptional()) {
                     return dg.renderType(w, child_type);
                 }
-                return dg.fail(dg.decl.src(), "TODO: C backend: more optional types", .{});
+
+                // TODO this needs to be typedeffed since different structs are different types.
+                try w.writeAll("struct { ");
+                try dg.renderType(w, child_type);
+                try w.writeAll(" payload; bool is_null; }");
             },
             .Null, .Undefined => unreachable, // must be const or comptime
             else => |e| return dg.fail(dg.decl.src(), "TODO: C backend: implement type {s}", .{
@@ -834,6 +840,7 @@ fn genAsm(o: *Object, as: *Inst.Assembly) !CValue {
 fn genIsNull(o: *Object, inst: *Inst.UnOp) !CValue {
     const writer = o.writer();
     const invert_logic = inst.base.tag == .is_non_null or inst.base.tag == .is_non_null_ptr;
+    const operator = if (invert_logic) "!=" else "==";
     const maybe_deref = if (inst.base.tag == .is_null_ptr or inst.base.tag == .is_non_null_ptr) "[0]" else "";
     const operand = try o.resolveInst(inst.operand);
 
@@ -843,10 +850,8 @@ fn genIsNull(o: *Object, inst: *Inst.UnOp) !CValue {
 
     if (inst.operand.ty.isPtrLikeOptional()) {
         // operand is a regular pointer, test `operand !=/== NULL`
-        const operator = if (invert_logic) "!=" else "==";
         try writer.print("){s} {s} NULL;\n", .{ maybe_deref, operator });
     } else {
-        const operator = if (invert_logic) "!=" else "==";
         try writer.print("){s}.is_null {s} true;\n", .{ maybe_deref, operator });
     }
     return local;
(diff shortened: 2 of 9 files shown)


src/codegen/c.zig
```diff
@@ -200,6 +229,27 @@ pub const DeclGen = struct {
...
+                    try dg.renderValue(writer, child_type, val);
+                    try writer.writeAll(" }");
+                }
+            },
+            .ErrorSet => {
+                const payload = val.castTag(.@"error").?;
+                // error values will be #defined at the top of the file
+                return writer.print("zig_error_{s}", .{payload.data.name});
+            },
             else => |e| return dg.fail(dg.decl.src(), "TODO: C backend: implement value {s}", .{
```
g-w1, line 253: I think you are missing error union here. I will do it as I am also making tests for this right now.

g-w1: Actually this requires some more infrastructure around runtime error comparison/wrapping/unwrapping error unions. I will try to do this.

g-w1: Could you actually rebase this? It needs to be done before I can contribute to it. I do not think I can rebase it myself.

Vexu: > Could you actually rebase this?

Sure

>I think you are missing error union here. 

We didn't have error union values anywhere in the compiler when I wrote this and I didn't want to implement them in this pr.

src/codegen/c.zig
```diff
@@ -424,6 +529,15 @@ pub fn genBody(o: *Object, body: ir.Body) error{ AnalysisFail, OutOfMemory }!voi
             .bit_or => try genBinOp(o, inst.castTag(.bit_or).?, " | "),
             .xor => try genBinOp(o, inst.castTag(.xor).?, " ^ "),
             .not => try genUnOp(o, inst.castTag(.not).?, "!"),
+            .is_null => try genIsNull(o, inst.castTag(.is_null).?),
+            .is_non_null => try genIsNull(o, inst.castTag(.is_non_null).?),
+            .is_null_ptr => try genIsNull(o, inst.castTag(.is_null_ptr).?),
+            .is_non_null_ptr => try genIsNull(o, inst.castTag(.is_non_null_ptr).?),
+            .wrap_optional => try genWrapOptional(o, inst.castTag(.wrap_optional).?),
+            .optional_payload => try genOptionalPayload(o, inst.castTag(.optional_payload).?),
+            .optional_payload_ptr => try genOptionalPayload(o, inst.castTag(.optional_payload).?),
```
g-w1, line 538: Should this be casting to `.optional_payload_ptr`?

src/codegen/c.zig
g-w1: https://github.com/Vexu/zig/pull/2

src/link/C.zig
```diff
@@ -142,17 +168,34 @@ pub fn flushModule(self: *C, comp: *Compilation) !void {
...
+    var err_typedef_buf = std.ArrayList(u8).init(comp.gpa);
+    defer err_typedef_buf.deinit();
+    const err_typedef_writer = err_typedef_buf.writer();
+    const err_typedef_item = all_buffers.addOneAssumeCapacity();
+
+    render_errors: {
+        if (module.global_error_set.size == 0) break :render_errors;
+        var it = module.global_error_set.iterator();
+        while (it.next()) |entry| {
+            // + 1 because 0 represents no error
```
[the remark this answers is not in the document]
Vexu: It does not deviate form the current semantics as 0 is not greater than 0.
 
>An error set is like an enum. However, each error name across the entire compilation gets assigned an unsigned integer greater than 0. You are allowed to declare the same error name more than once, and if you do, it gets assigned the same integer value.

>The number of unique error values across the entire compilation should determine the size of the error set type. However right now it is hard coded to be a `u16`. See #768. 

From langref [Error Set Type](https://ziglang.org/documentation/master/#toc-Error-Set-Type).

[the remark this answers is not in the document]
g-w1: So the only thing is, when implementing @errorToInt and intToError we need to add/subtract 1.

andrewrk:
> I upped the C version used in tests to C99 since we were already using `-Wno-declaration-after-statement` and C89 doesn't have compound literals which would have required a bunch of annoying workarounds.

I'm fine with this for now, but I think it would be worthwhile to target C89 eventually, because one of the use cases of the C backend is to support less common architectures that only have old crusty C compilers for them.

src/Compilation.zig
```diff
@@ -1653,6 +1653,8 @@ pub fn performAllTheWork(self: *Compilation) error{ TimerUnsupported, OutOfMemor
                     .error_msg = null,
                     .decl = decl,
                     .fwd_decl = fwd_decl.toManaged(module.gpa),
+                    // we don't want to emit optionals and error unions to headers since they have no ABI
+                    .typedefs = c_codegen.TypedefMap.init(std.testing.failing_allocator),
```
andrewrk, line 1657: I think use of failing_allocator is not the right approach here. This might drag in all manner of testing stuff. Really, what's expected here is for the code to not add anything to this array right? Should it perhaps be `undefined` or `null`?

src/test.zig
```diff
@@ -868,10 +868,9 @@ pub const TestContext = struct {
                                 std.testing.zig_exe_path,
                                 "run",
                                 "-cflags",
-                                "-std=c89",
+                                "-std=c99",
```
andrewrk, line 871: Choice of C89 was deliberate - can we talk about what it might look like to support C89 here before giving up?

Another option we have is to make some exceptions (as you see below with `-Wno-declaration-after-statement`) which act as "TODO"s but still will allow us to notice in other places where we violate C89.

src/link/C.zig
```diff
@@ -142,17 +168,34 @@ pub fn flushModule(self: *C, comp: *Compilation) !void {
...
+            try err_typedef_writer.print("#define zig_error_{s} {d}\n", .{ entry.key, entry.value + 1 });
```
andrewrk, line 189: Keep in mind that it is not a design goal of generated C code to be human-readable. For the purposes of #6378 I expect to do some one-off hacking about for a few days or maybe even weeks, in order to hack in human-readableness to the C backend temporarily, and for this one use case. Other than that, there is no expectation at all for the outputted C code to be human readable. So I think the null hypothesis would be to use simple integers for errors in generated C code.

On the other hand, if this indirection simplifies or improves the design - for example by allowing the mapping of names to integers to change independently of the usage sites - then it makes sense to have this here.

andrewrk (changes requested):

src/link/C.zig
g-w1: I think it is simpler now because only Module has the error values to save memory so we only have the strings which allows us to do 
```zig
                const payload = val.castTag(.@"error").?;
                // error values will be #defined at the top of the file
                return writer.print("zig_error_{s}", .{payload.data.name});
```
When we do not have the value.

src/Compilation.zig
Vexu: I meant to use `mem.failAllocator` but `undefined` works too.

src/test.zig
Vexu: Currently the only other c99 feature we are using is compound literals for the optionals and error unions, and I doubt any compiler stuck in c89 will support them. Getting rid of them would require either every call to `renderValue` to pass a temporary value result location or it would require `renderValue` to be able to create declarations. The first option would probably not even work with stuff like function calls and arrays but the second one should be fine.

andrewrk: OK, I'm fine with C99, at least for now. I'd much rather prioritize getting the backend done with C99 than imposing possibly unnecessary restrictions on ourselves and impeding progress.

PR #7943, opened 2021-02-04
make std.c.getErrno() return same type as _errno() aka c_int

semarie:
while here, adjust std.os.unexpectedErrno()

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -144,25 +144,27 @@ pub fn getrandom(buffer: []u8) GetRandomError!void {
             std.c.versionCheck(builtin.Version{ .major = 2, .minor = 25, .patch = 0 }).ok;
 
         while (buf.len != 0) {
-            var err: u16 = undefined;
-
-            const num_read = if (use_c) blk: {
+            const res = if (use_c) blk: {
                 const rc = std.c.getrandom(buf.ptr, buf.len, 0);
-                err = std.c.getErrno(rc);
-                break :blk @bitCast(usize, rc);
+                break :blk .{
+                    .num_read = @bitCast(usize, rc),
+                    .err = std.c.getErrno(rc),
+                };
             } else blk: {
                 const rc = linux.getrandom(buf.ptr, buf.len, 0);
-                err = linux.getErrno(rc);
-                break :blk rc;
+                break :blk .{
+                    .num_read = rc,
+                    .err = linux.getErrno(rc),
+                };
             };
 
-            switch (err) {
-                0 => buf = buf[num_read..],
+            switch (res.err) {
+                0 => buf = buf[res.num_read..],
                 EINVAL => unreachable,
                 EFAULT => unreachable,
                 EINTR => continue,
                 ENOSYS => return getRandomBytesDevURandom(buf),
-                else => return unexpectedErrno(err),
+                else => return unexpectedErrno(res.err),
             }
         }
         return;
@@ -1500,7 +1502,7 @@ pub fn getcwd(out_buffer: []u8) GetCwdError![]u8 {
         EINVAL => unreachable,
         ENOENT => return error.CurrentWorkingDirectoryUnlinked,
         ERANGE => return error.NameTooLong,
-        else => return unexpectedErrno(@intCast(usize, err)),
+        else => return unexpectedErrno(err),
     }
 }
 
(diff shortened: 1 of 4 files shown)


semarie:
With the current `u16` type for `std.c.getErrno()`, I couldn't properly use [ZeroMQ](https://github.com/zeromq/libzmq) library as it is using `errno` for application specific error using value outside `u16` range (see [`ETERM` definition](https://github.com/zeromq/libzmq/blob/53104ec1b94657908d9c23d65a6f646847622c09/include/zmq.h#L194)).

Additionnally, [Posix](https://man.bsd.lv/POSIX-2013/errno) define the `errno` value to have type `int`.

I also manually checked `_errno()`  definitions in zig std, and all platforms are using `c_int`.

lib/std/os.zig
```diff
@@ -4526,7 +4526,7 @@ pub const UnexpectedError = error{
 
 /// Call this when you made a syscall or something that sets errno
 /// and you get an unexpected error.
-pub fn unexpectedErrno(err: usize) UnexpectedError {
+pub fn unexpectedErrno(err: c_int) UnexpectedError {
```
ifreund, line 4529: I'm not sure this is the way to go. If not linking libc, `os.getErrno()` returns a `u12` not a `c_int` on linux at least. Do we really want to tie this API to the C integer type?

semarie: I am not very familiar with linux kernel internals. I assume the errno size used by kernel is part of its ABI.
But at least `u12` fully fits inside a `c_int` (assuming at least 16 bits)

daurnimator: Yes; however we don't want to use larger types just because C did...

semarie:
@daurnimator what would be the alternative for be able to use ZeroMQ with zig ? There are assuming that `errno` is at least 32 bits (see [comment](https://github.com/ziglang/zig/pull/7943#issuecomment-773090637))

should I just use `u32` ?

daurnimator:
> There are assuming that `errno` is at least 32 bits (see [comment](https://github.com/ziglang/zig/pull/7943#issuecomment-773090637))

That assumption is not correct on all platforms; however it *might* be correct on platforms that zeromq supports (or even perhaps true in all libcs).

And perhaps I should have been more clear with my comment: I think the `std.c.getErrno()` change is okay, but the `unexpectedErrno` one is not.

semarie:
some thoughts... having different types for the same thing force to using casting, which is bad

- `std.c._errno()` (imported from `std/c/xxx.zig`) :  `c_int` for everybody (due to posix...)
- `std.c.getErrno(rc: anytype)` : if rc == -1, return `std.c._errno()` : currently `u16`, but I am expecting it to be the same type as `std.c._errno()` (this PR corrects it)
- `std.os.errno` = `std.os.system.getErrno` : somehow complex rule (could be `std.c.getErrno(rc)` or something else). linux and wasi has specific `getErrno(rc)`. linux with `u12` and wasi with `usize`.
- `std.os.unexpectedErrno(err: usize)` : the type is wrong for all except wasi (linux is fine only because `u12` fits)
- when using `unexpectedErrno()` `@intCast()` is used regulary

the last commit still change `std.os.unexpectedErrno()` parameter, but to make it `anytype`. I think it is more correct instead of forcing one specific type. maybe it could be more precise and accept only `.Int` type

lib/std/os.zig
```diff
@@ -144,25 +144,15 @@ pub fn getrandom(buffer: []u8) GetRandomError!void {
...
-                err = std.c.getErrno(rc);
-                break :blk @bitCast(usize, rc);
-            } else blk: {
-                const rc = linux.getrandom(buf.ptr, buf.len, 0);
-                err = linux.getErrno(rc);
-                break :blk rc;
-            };
-
-            switch (err) {
+            const getrandom_ = if (use_c) std.c.getrandom else linux.getrandom;
```
daurnimator, line 147: This results in an indirect call; please keep the previous form.

lib/std/os.zig
```diff
@@ -4622,7 +4612,7 @@ pub const UnexpectedError = error{
 
 /// Call this when you made a syscall or something that sets errno
 /// and you get an unexpected error.
-pub fn unexpectedErrno(err: usize) UnexpectedError {
+pub fn unexpectedErrno(err: anytype) UnexpectedError {
```
daurnimator, line 4615: hmm, why do we need this?

Also note that if `unexpected_error_tracing` is false then people might start accidentally passing other types to this (enums/structs) and have no idea until someone turns on unexpected error tracing

semarie: to cite myself "maybe it could be more precise and accept only .Int type". I added a check on the type.

Vexu (approved):

ifreund (approved):
LGTM, Thanks!

PR #7946, opened 2021-02-04
SPARCv9: Handle various stack frame related quirks.

koachan:
These are some fixes to account for SPARCv9 quirks in stack frame handling. I'm not very sure with the `gen_frame_size` workaround, but it seems to fix the problematic test cases for me, at least.

I've put a detailed description of each patch in the commit messages.

### DIFF
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -4161,6 +4161,12 @@ static LLVMValueRef gen_frame_size(CodeGen *g, LLVMValueRef fn_val) {
     LLVMValueRef negative_one = LLVMConstInt(LLVMInt32Type(), -1, true);
     LLVMValueRef prefix_ptr = LLVMBuildInBoundsGEP(g->builder, casted_fn_val, &negative_one, 1, "");
     LLVMValueRef load_inst = LLVMBuildLoad(g->builder, prefix_ptr, "");
+
+    // Some architectures (e.g SPARCv9) has different alignment requirements between a
+    // function/usize pointer and also require all loads to be aligned.
+    // On those architectures, not explicitly setting the alignment will lead into @frameSize
+    // generating usize-aligned load instruction that could crash if the function pointer
+    // happens to be not usize-aligned.
     LLVMSetAlignment(load_inst, 1);
     return load_inst;
 }
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -4160,7 +4160,9 @@ static LLVMValueRef gen_frame_size(CodeGen *g, LLVMValueRef fn_val) {
     LLVMValueRef casted_fn_val = LLVMBuildBitCast(g->builder, fn_val, ptr_usize_llvm_type, "");
     LLVMValueRef negative_one = LLVMConstInt(LLVMInt32Type(), -1, true);
     LLVMValueRef prefix_ptr = LLVMBuildInBoundsGEP(g->builder, casted_fn_val, &negative_one, 1, "");
-    return LLVMBuildLoad(g->builder, prefix_ptr, "");
+    LLVMValueRef load_inst = LLVMBuildLoad(g->builder, prefix_ptr, "");
+    LLVMSetAlignment(load_inst, 1);
+    return load_inst;
 }
 
 static void gen_init_stack_trace(CodeGen *g, LLVMValueRef trace_field_ptr, LLVMValueRef addrs_field_ptr) {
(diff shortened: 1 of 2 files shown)


src/stage1/codegen.cpp
```diff
@@ -4160,7 +4160,9 @@ static LLVMValueRef gen_frame_size(CodeGen *g, LLVMValueRef fn_val) {
     LLVMValueRef casted_fn_val = LLVMBuildBitCast(g->builder, fn_val, ptr_usize_llvm_type, "");
     LLVMValueRef negative_one = LLVMConstInt(LLVMInt32Type(), -1, true);
     LLVMValueRef prefix_ptr = LLVMBuildInBoundsGEP(g->builder, casted_fn_val, &negative_one, 1, "");
-    return LLVMBuildLoad(g->builder, prefix_ptr, "");
+    LLVMValueRef load_inst = LLVMBuildLoad(g->builder, prefix_ptr, "");
+    LLVMSetAlignment(load_inst, 1);
```
LemonBoy, line 4164: Weird this problem didn't show up on AArch64, the pointers are aligned to 8 bytes but the opcodes are 4 bytes wide. Perhaps the default alignment is set to 16 or higher and we never noticed...

What is the reported ABI alignment on `fn_val` ?

Needs a comment explaining why `LLVMSetAlignment` is here.

koachan: AArch64 supports unaligned load AFAIK, that's why it's seems to be okay there.
How do I check the reported ABI alignment? gdb only says <incomplete type> when inspecting it.

LemonBoy: > How do I check the reported ABI alignment? 

LLVM's ABI alignment

koachan: `LLVMGetAlignment(fn_val)` returns some big garbage value, and the datalayout string only says `E-m:e-i64:64-n32:64-S128`, no pointer alignment value specified.
Is this the correct info?

LemonBoy: Given "big garbage value" is not a number no, I think that's not the correct info.
When in doubt you can always check the generated asm and see what `p2align` (or similar) are emitted.

Zig is incorrectly assuming every function pointer is aligned to `usize` while that's generally not true. As you can see [here](https://reviews.llvm.org/D57335), LLVM has a way of precisely describing what's the function pointer alignment or, if no option is specified in the datalayout, defaults to a safe (?) value of 4.

koachan: > Given "big garbage value" is not a number no, I think that's not the correct info.

Ah, sorry. I made a mistake in the testcase that results in it returning different values on each compilation, that's why I thought it was garbage.
I've revised [the code](https://gist.github.com/koachan/e068077e7ff2d6fe8a10127c179dc416), and now `LLVMGetAlignment(fn_val)` always returns 8. Does this help?

> When in doubt you can always check the generated asm and see what p2align (or similar) are emitted.

It emits `.p2align 2` before each function, so the functions are aligned to four bytes.
So, should I change the LLVMSetAlignment parameter to 4, or?

> As you can see here, LLVM has a way of precisely describing what's the function pointer alignment or, if no option is specified in the datalayout, defaults to a safe (?) value of 4.

Interesting, didn't know that there's an `F` directive, it doesn't seem to be documented at all.

LemonBoy: > LLVMGetAlignment(fn_val) always returns 8. Does this help?

That's because it's a pointer type. The generated LLVM IR is:

```llvm
  store void (%"(anyframe->void header)"*, i64)* bitcast (void (%"@Frame(S.first)"*, i64)* @S.first to void (%"(anyframe->void header)"*, i64)*), void (%"(anyframe->void header)"*, i64)** %ptr1, align 8
  %6 = load void (%"(anyframe->void header)"*, i64)*, void (%"(anyframe->void header)"*, i64)** %ptr1, align 8
  %7 = bitcast void (%"(anyframe->void header)"*, i64)* %6 to i64*
  %8 = getelementptr inbounds i64, i64* %7, i32 -1
  %9 = load i64, i64* %8, align 8
```

What we're interested in is the effective alignment of the function address. Let's see what the langref has to say about this:

> An explicit alignment may be specified for a function. If not present, or if the alignment is set to zero, the alignment of the function is set by the target to whatever it feels convenient.

How do we discover what's a convenient alignment value for this target then? The `F` entry in the datalayout!
(That's [well documented in the langref](https://llvm.org/docs/LangRef.html#id1308)...)

There are two API functions, `getFunctionPtrAlignType` and `getFunctionPtrAlign`, that allow the user to retrieve the value of the `F` entry. Unfortunately nobody stepped up to update all the targets with this annotation (the SPARC backend is almost on life support so I doubt it'll magically pop-up anytime soon)!

How does Zig handle this problem? `get_ptr_align` uses a safe value of 1 and calls it a day.
You can safely leave the `LLVMSetAlignment` to 1, the backend will thank you for letting it know this load is unaligned.

andrewrk:
Thank you and nice work!

PR #7649, opened 2021-01-02
std: add EnumSet type

daurnimator:
Written for the upcoming `.standardOptions`, but independently useful.

I don't love the method names, but I made them match `BufSet` for now.

PR #7954, opened 2021-02-05
Add some bit set variants

SpexGuy:
In this week's design meeting, we talked about adding some enum utility types to the standard library.  As a prerequisite to a dense enum map, we need a bit set data structure.  So I wrote four 😛.

This is my first major PR for the standard library, so please let me know if I've broken any conventions or design rules.  I'm also open to discussing api changes or bikeshedding on better names.


daurnimator:
Possible duplicate of #7649

SpexGuy:
Good call, I hadn't seen that one.  I was planning to do an EnumMap and EnumArray in addition to the EnumSet, but those can be added on later after 7649 is pulled.  This PR is just the raw bit set data structure though, so I don't think it conflicts.  It provides bulk operations that are unavailable in PackedIntArray, and uses larger words to improve performance.  It also is able to make the assumption that a single entry will never span multiple backing words, so it is much simpler.

SpexGuy:
^ just deleting some unused fields in `DynamicBitSet`.

data-man:
Some thoughts:
- `countSet` -> `count`
- add `clz` and `ctz`.
- add `VectorBitSet` or `SIMDArrayBitSet`, etc.

data-man:
Also, sets can be implemented as tries.
E.g. https://github.com/skeeto/int32trie, https://github.com/amethyst/hibitset, etc.

data-man:
> So I wrote four

```zig
pub const BitSetType = union(enum) {
  Array: usize,
  Simd,: usize,
  Dynamic,
};

pub fn BitSet(comptime bitset_type: BitSetType) type {
...
}
```

SpexGuy:
> countSet => count

I initially avoided this because C# uses Count as the function to get the number of items in a collection.  But I guess in a sense that's what this does.  I'll change this and also change `len` to `capacity`.

> clz and ctz

I suppose that makes sense. It's easy enough to add so I'll do that.

> Vector/simd/trie set

These are good ideas but are outside of the scope of this PR.

> BitSetType

I could do this, but why?  The APIs of dynamic vs static are mostly similar, but they are different enough that you can't create a bit set without knowing which one is being used.  They also target different use cases.  I think this is a niche enough case that the std lib doesn't need to support it, a user can write this function if they need it.

SpexGuy:
Ok, I made those naming changes.  After thinking more about ctz and clz, I think they are strange and unnecessary operations for a bit set.  However, I have adjusted the iterator interface so you can now iterate forwards or backwards, so this could be used to implement clz and ctz if you really need that for some reason.

data-man:
> After thinking more about ctz and clz, I think they are strange and unnecessary operations for a bit set.

No, these are very frequently used operations.

> I have adjusted the iterator interface so you can now iterate forwards or backwards, so this could be used to implement clz and ctz if you really need that for some reason.

Thanks! But it's slow.

SpexGuy:
> these are very frequently used operations

I suppose so, for example using a bit set to find the first free allocation in a block.  What do you think about this API instead?
```zig
findFirstSet() ?usize
findFirstUnset() ?usize
findLastSet() ?usize
findLastUnset() ?usize
toggleFirstSet() ?usize
toggleFirstUnset() ?usize
toggleLastSet() ?usize
toggleLastUnset() ?usize
```

> But it's slow.

The optimizer is pretty good at destroying this sort of iterator.  The generated assembly looks pretty close to what I would want if I implemented these functions by hand:
https://zig.godbolt.org/z/qETKcr

SpexGuy:
Ok one more reversal, sorry.  Supporting all 8 operations on every bit set is quite a bit of code, overall almost 800 extra lines.  Some of these operations are quite rare.  What about only supporting the most efficient ones directly:
```zig
findFirstSet() ?usize
toggleFirstSet() ?usize
```
The others can be implemented in more complex ways using other ways, but these two functions specifically are the fastest.  If you really need speed for this operation, you should plan your use of the bit set so that these are the operations you need.

If more are needed, please provide a use case that needs the other operations.

data-man:
E.g. very good https://github.com/lemire/cbitset has this API:
```c
/* Create a new bitset. Return NULL in case of failure. */
bitset_t *bitset_create( void );

/* Create a new bitset able to contain size bits. Return NULL in case of failure. */
bitset_t *bitset_create_with_capacity( size_t size );

/* Free memory. */
void bitset_free(bitset_t *bitset);

/* Set all bits to zero. */
void bitset_clear(bitset_t *bitset);

/* Create a copy */
bitset_t * bitset_copy(const bitset_t *bitset);

/* Resize the bitset. Return true in case of success, false for failure. Pad with zeroes new buffer areas if requested. */
bool bitset_resize( bitset_t *bitset,  size_t newarraysize, bool padwithzeroes );

/* returns how many bytes of memory the backend buffer uses */
static inline size_t bitset_size_in_bytes(const bitset_t *bitset);

/* returns how many bits can be accessed */
static inline size_t bitset_size_in_bits(const bitset_t *bitset);

/* returns how many words (64-bit) of memory the backend buffer uses */
static inline size_t bitset_size_in_words(const bitset_t *bitset);

/* Grow the bitset so that it can support newarraysize * 64 bits with padding. Return true in case of success, false for failure. */
static inline bool bitset_grow( bitset_t *bitset,  size_t newarraysize );

/* attempts to recover unused memory, return false in case of reallocation failure */
bool bitset_trim(bitset_t *bitset);

/* shifts all bits by 's' positions so that the bitset representing values 1,2,10 would represent values 1+s, 2+s, 10+s */
void bitset_shift_left(bitset_t *bitset, size_t s);

/* shifts all bits by 's' positions so that the bitset representing values 1,2,10 would represent values 1-s, 2-s, 10-s, negative values are deleted */
void bitset_shift_right(bitset_t *bitset, size_t s);

/* Set the ith bit. Attempts to resize the bitset if needed (may silently fail) */
static inline void bitset_set(bitset_t *bitset,  size_t i );

/* Set the ith bit to the specified value. Attempts to resize the bitset if needed (may silently fail) */
static inline void bitset_set_to_value(bitset_t *bitset,  size_t i, bool flag);

/* Get the value of the ith bit.  */
static inline bool bitset_get(const bitset_t *bitset,  size_t i );

/* Count number of bits set.  */
size_t bitset_count(const bitset_t *bitset);

/* Find the index of the first bit set.  */
size_t bitset_minimum(const bitset_t *bitset);

/* Find the index of the last bit set.  */
size_t bitset_maximum(const bitset_t *bitset);

/* compute the union in-place (to b1), returns true if successful, to generate a new bitset first call bitset_copy */
bool bitset_inplace_union(bitset_t * CBITSET_RESTRICT b1, const bitset_t * CBITSET_RESTRICT b2);

/* report the size of the union (without materializing it) */
size_t bitset_union_count(const bitset_t * CBITSET_RESTRICT b1, const bitset_t * CBITSET_RESTRICT b2);

/* compute the intersection in-place (to b1), to generate a new bitset first call bitset_copy */
void bitset_inplace_intersection(bitset_t * CBITSET_RESTRICT b1, const bitset_t * CBITSET_RESTRICT b2);

/* report the size of the intersection (without materializing it) */
size_t bitset_intersection_count(const bitset_t * CBITSET_RESTRICT b1, const bitset_t * CBITSET_RESTRICT b2);

/* returns true if the bitsets contain no common elements */
bool bitsets_disjoint(const bitset_t * b1, const bitset_t * b2);

/* returns true if the bitsets contain any common elements */
bool bitsets_intersect(const bitset_t * b1, const bitset_t * b2);

/* returns true if b1 contains all of the set bits of b2 */
bool bitset_contains_all(const bitset_t * b1, const bitset_t * b2);

/* compute the difference in-place (to b1), to generate a new bitset first call bitset_copy */
void bitset_inplace_difference(bitset_t * CBITSET_RESTRICT b1, const bitset_t * CBITSET_RESTRICT b2);

/* compute the size of the difference */
size_t  bitset_difference_count(const bitset_t * CBITSET_RESTRICT b1, const bitset_t * CBITSET_RESTRICT b2);

/* compute the symmetric difference in-place (to b1), return true if successful, to generate a new bitset first call bitset_copy */
bool bitset_inplace_symmetric_difference(bitset_t * CBITSET_RESTRICT b1, const bitset_t * CBITSET_RESTRICT b2);

/* compute the size of the symmetric difference  */
size_t  bitset_symmetric_difference_count(const bitset_t * CBITSET_RESTRICT b1, const bitset_t * CBITSET_RESTRICT b2);

/* iterate over the set bits
 like so :
  for(size_t i = 0; nextSetBit(b,&i) ; i++) {
    //.....
  }
  */
static inline bool nextSetBit(const bitset_t *bitset, size_t *i);

/* iterate over the set bits
 like so :
   size_t buffer[256];
   size_t howmany = 0;
  for(size_t startfrom = 0; (howmany = nextSetBits(b,buffer,256, &startfrom)) > 0 ; startfrom++) {
    //.....
  }
  */
static inline size_t nextSetBits(const bitset_t *bitset, size_t *buffer, size_t capacity, size_t * startfrom);

typedef bool (*bitset_iterator)(size_t value, void *param);

// return true if uninterrupted
static inline bool bitset_for_each(const bitset_t *b, bitset_iterator iterator, void *ptr);

static inline void bitset_print(const bitset_t *b);

```

data-man:
> overall almost 800 extra lines

That's why I suggested one type with options.

nurpax:
@SpexGuy Re: https://github.com/ziglang/zig/pull/7954#issuecomment-774513754

I had a need for `findFirstUnset`, it feels like at least this variant is common and would pair well with `findFirstSet`.  My use-case was in code like this:

```
        pub fn newEntity(self: *Self) u32 {
            // implement findFirstUnset using an iterator
            var unset_it = self.valid.iterator(.{ .kind = .unset });
            const idx = while (unset_it.next()) |idx| {
                break idx;
            } else {
                @panic("out of entities -- shouldn't get here");
            };
            self.valid.set(idx);
            self.tags[idx] = 0;
            return @intCast(u32, idx);
        }
```

I wouldn't want to flip the meaning of `self.valid` bitset just because `findFirstUnset` is not there.  IMO a call to `findFirstUnset` would be more readable here.

The iterator path most likely compiles into pretty good code in release builds, so would there be willingness to accept a patch that implements `findFirstUnset` and `toggleFirstUnset` using iterators?  This wouldn't add much more code into the implementation and I'd check the generate assembly with godbolt.

PR #7959, opened 2021-02-05
Make build.zig ready for generated files

MasterQ32:
*First of all*: This is a **breaking** PR which will break **all** projects using `std.build.Pkg`. This PR is a draft as well.

*Goal*: This PR tries to improve the experience with generated files by various build steps.

Right now, `std.build.FileSource` is used in some places to allow either a `std.build.WriteFileStep` or `std.build.TranslateC` to generate files and use them to create executables and libraries. This PR replaces the special cases with a generic `std.build.GeneratedFile` which is a `@fieldParentPtr` style interface that allows to retrieve the path of a generated file after its `make()` invocation in a general purpose manner.

Usage example:

```zig
const std = @import("std");

const TemplateStep = @import("src/TemplateStep.zig");

    // An arbitrary step that generates a output file
    const template_step = TemplateStep.create(b, std.build.FileSource{
        .path = "example/layout.ztt",
    });

    const exe = b.addExecutable("demo", "example/main.zig");
    exe.addPackage(std.build.Pkg{
        .name = "template",
        .path = template_step.getFileSource(), // we can now add the generated file also as a package source
    });
    exe.setTarget(target);
    exe.setBuildMode(mode);
    exe.install();
}
```

Right now, this PR is unfinished as i first want some feedback on the changes i did and if we want to pursue this change.

Waiting for a response! 
– xq


MasterQ32:
I get the following error in the CI:
```
Install the project...
-- Install configuration: "Release"
-- Installing: /usr/home/build/zig/build/release/bin/zig
-- Installing: /usr/home/build/zig/build/release/lib
./build.zig:301:8: error: expected type 'std.build.FileSource', found '[]u8'
    }) catch unreachable);
       ^
./lib/std/build.zig:1222:24: note: std.build.FileSource declared here
pub const FileSource = union(enum) {
                       ^
./build.zig:124:13: note: referenced here
            try addCmakeCfgOptionsToExe(b, cfg, tracy, exe);
            ^
./lib/std/special/build_runner.zig:157:24: note: referenced here
        .ErrorUnion => try root.build(builder),
                       ^
```

How can i reproduce this locally?

MasterQ32:
Okay, so the command failing was:
```
./build/zig build --override-lib-dir ./lib -Dlib-files-only --prefix ./release -Dconfig_h=./build/config.h install
```
I fixed the bugs happening with this

lib/std/build.zig
```diff
@@ -1265,40 +1203,68 @@ fn isLibCLibrary(name: []const u8) bool {
...
 }
 
+/// A file that is generated by a build step.
+/// This struct is an interface that is meant to be used with `@fieldParentPtr` to implement the actual path logic.
+pub const GeneratedFile = struct {
+    /// The step that generates the file
+    step: *Step,
+
+    /// A function that returns the absolute path to the generated file.
+    getPathFn: fn (self: *const GeneratedFile) []const u8,
```
SpexGuy, line 1213: Does this need to be an interface?  A boxed string (`struct { path: []const u8 = "" }`) which gets updated in the make step would be simpler.  You could check if the string is still empty when read and give a nice panic message also.

lib/std/build.zig
```diff
@@ -22,12 +22,12 @@ const fmt_lib = std.fmt;
...
-pub const WriteFileStep = @import("build/write_file.zig").WriteFileStep;
-pub const RunStep = @import("build/run.zig").RunStep;
-pub const CheckFileStep = @import("build/check_file.zig").CheckFileStep;
-pub const InstallRawStep = @import("build/emit_raw.zig").InstallRawStep;
+pub const FmtStep = @import("build/FmtStep.zig");
+pub const TranslateCStep = @import("build/TranslateCStep.zig");
+pub const WriteFileStep = @import("build/WriteFileStep.zig");
+pub const RunStep = @import("build/RunStep.zig");
+pub const CheckFileStep = @import("build/CheckFileStep.zig");
+pub const InstallRawStep = @import("build/InstallRawStep.zig");
```
SpexGuy, line 30: These changes are fine but it would be nice in the future if style changes like this and the casing changes below were separated from functional changes.  It's harder to read the diff when it's all mixed together.

SpexGuy:
Overall these are good changes, but I think they are missing a major use case:
Some file generation processes are expensive, or invalidate downstream results that would otherwise be cached.  It would be nice if generated files could declare their dependencies and check if the generation process needs to be re-done.  This is more of a wishlist thing for me, I don't have specific ideas on how this interface should look.  But I think build system support for generated files would be incomplete without this feature.

One thing to note, that might help with this feature, is that for most processes you don't need to check if any of the current dependencies have changed, it is sufficient to only check if any of the *previous* dependencies have changed.  So you could have the list of dependencies be an *output* of the make step, and save that list of files in the cache to be checked on the next run.  This eliminates needing to write separate pieces of code for doing the build vs examining dependencies.

andrewrk:
I agree with @SpexGuy but I think that the build system supporting a general purpose caching system is something we need regardless, and orthogonal to this PR. It is certainly a glaring missing feature, however it is something that can be plugged into these changes later, and also will need to be plugged into the other build steps that do not rely on the internal caching of the zig compiler.

lib/std/build.zig
MasterQ32: True, but i'm usually motivated to refactor style things when i touch the code anyways and make it progressive this way. But: I'll make it two PRs the next time

lib/std/build.zig
MasterQ32: I don't think this is simpler in the end, especially as you require additional allocations, pointer references and such. Right now, we have a single forward reference to manage: `*const GeneratedFile`

Or do you mean `GeneratedFile` will look like this:
```zig
struct {
    step: *Step,
    path: ?[]const u8 = null,
}
```
And the generator requires to update `.path` to a non-null variant?

SpexGuy: Yeah, exactly, that second one.  You don't need extra allocations since steps are never deleted, just have the referencing step point into a GeneratedFile struct inside the referenced step.

If it's a function, it can have side effects and do all sorts of things.  It's not clear what operations the calling code can tolerate and what might break.  But like this it's very clear -- the referencer will only read this value, and the referenced step must set it before that happens.

Assuming `""` is an invalid value for a path, using that as a sentinel instead of an optional will save eight bytes, which is why I went for that version.

MasterQ32: I'll rework the PR to use this case. It's definitly the superior solution here. Thanks for pointing that out!

MasterQ32:
Another thing that i left out (for now) was that `LibExeObjStep` does not provide a file source. But that's kinda not right. I think i should change that as well, so we can just use the output of a compile step as well, for example to push it directly into objdump or objcopy

lib/std/build.zig
MasterQ32: Changed the code to use this pattern instead

MasterQ32:
> Another thing that i left out (for now) was that LibExeObjStep does not provide a file source. But that's kinda not right. I think i should change that as well, so we can just use the output of a compile step as well, for example to push it directly into objdump or objcopy

I tried to change the LibExeObjStep to use the `GeneratedFile` instead, but i can't figure out the right points where to actually set the `.path` property for those files. @andrewrk i talked to you on Discord about this already, you might need to take a look at it. Apart from that, i'm ready for merging i think

MasterQ32:
I am confused. On my local machine, `zig build --override-lib-dir lib test-toolchain` runs clean.

Note here: I haven't built `zig` with the changes in the build system, but to my understanding this should not change anything.

MasterQ32:
I found some hints, will investigate. Fixed a bug already :grin:

MasterQ32:
CI is green (latest commit just deletes a file that appeared during the rebase), so i'm ready to merge. @SpexGuy, can you review the latest state again? :thinking:

lib/std/build.zig
```diff
@@ -204,120 +204,58 @@ pub const Builder = struct {
...
-        basename: []const u8,
-    ) *LibExeObjStep {
-        return LibExeObjStep.createObject(self, name, @as(FileSource, .{
-            .write_file = .{
-                .step = wfs,
-                .basename = basename,
-            },
-        }));
-    }
+    pub const addObjectSource = LibExeObjStep.createObject;
```
SpexGuy, line 224: I think it would make the source more readable and improve documentation to write out this function, even though it's just a trivial call.

lib/std/build.zig
```diff
@@ -1276,40 +1214,69 @@ fn isLibCLibrary(name: []const u8) bool {
...
+pub const GeneratedFile = struct {
+    /// The step that generates the file
+    step: *Step,
+
+    /// The path to the generated file. Must be either absolute or relative to the build root.
+    /// This value must be set in the `fn make()` of the `step` and must not be `null` afterwards.
+    path: ?[]const u8 = null,
+
+    pub fn getPath(self: *const GeneratedFile) []const u8 {
+        return self.path orelse @panic("getPath() was called on a GeneratedFile that wasn't build yet. Is there a missing Step dependency?");
```
SpexGuy, line 1228: This error message should give info about the step that failed, since you have it available.

lib/std/build.zig
```diff
@@ -1276,40 +1214,69 @@ fn isLibCLibrary(name: []const u8) bool {
...
+/// This struct is an interface that is meant to be used with `@fieldParentPtr` to implement the actual path logic.
+pub const GeneratedFile = struct {
+    /// The step that generates the file
+    step: *Step,
+
+    /// The path to the generated file. Must be either absolute or relative to the build root.
+    /// This value must be set in the `fn make()` of the `step` and must not be `null` afterwards.
+    path: ?[]const u8 = null,
+
+    pub fn getPath(self: *const GeneratedFile) []const u8 {
```
SpexGuy, line 1227: Use `self: GeneratedFile` instead, the address is not important.

SpexGuy (commented):
Overall this looks good.  Does `build.zig` in the root of the repository also need to be updated?

lib/std/build.zig
MasterQ32: True, fixed.

lib/std/build.zig
MasterQ32: I added the name of the step into the panic message

MasterQ32:
> Overall this looks good. Does `build.zig` in the root of the repository also need to be updated?

No, the CI already tested this including all build scripts in the tests folder :smiley: It's less breaking than i expected, which is good

MasterQ32:
Rebased once again

Vexu (approved):
LGTM, made all the remaining enums snake_case while we're breaking everything.

PR #7972, opened 2021-02-06
docs: document the nosuspend keyword

mguaypaq:
Here's a basic description, based on the examples I gathered from the standard library.

One detail I tested and wasn't sure whether to include: a `suspend` block can still be reached inside a `nosuspend` scope, as long as it resumes itself, and this is fine, as illustrated in the following program:

```zig
pub fn main() void {
    nosuspend bla();
}

fn bla() void {
    suspend {
        resume @frame();
    }
}
```

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -6581,13 +6581,11 @@ const std = @import("std");
 const expect = std.testing.expect;
 
 test "async and await" {
-    // Here we have an exception where we do not match an async
-    // with an await. The test block is not async and so cannot
-    // have a suspend point in it.
-    // This is well-defined behavior, and everything is OK here.
-    // Note however that there would be no way to collect the
-    // return value of amain, if it were something other than void.
-    _ = async amain();
+    // The test block is not async and so cannot have a suspend
+    // point in it. By using the nosuspend keyword, we promise that
+    // the code in amain will finish executing without suspending
+    // back to the test block.
+    nosuspend amain();
 }
 
 fn amain() void {
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -10760,10 +10760,14 @@ fn readU32Be() u32 {}
           </td>
           <td>
             The {#syntax#}nosuspend{#endsyntax#} keyword can be used in front of a block, statement or expression.
-            It marks a scope where async functions are not allowed to actually suspend execution.
-            Attempting to suspend execution inside a {#syntax#}nosuspend{#endsyntax#} scope results in safety-checked {#link|Undefined Behavior#}.
-            Code inside a {#syntax#}nosuspend{#endsyntax#} scope does not cause the enclosing function to become an {#link|async function|Async Functions#}.
+            It marks a scope where async functions are not allowed to suspend execution.
             <ul>
+              <li>Attempting to suspend execution inside a {#syntax#}nosuspend{#endsyntax#} scope results in safety-checked {#link|Undefined Behavior#}.
+                <ul>
+                  <li>An async function which {#link|resumes itself from a suspend block|Resuming from Suspend Blocks#} is not considered suspended, so this construct is allowed inside {#syntax#}nosuspend{#endsyntax#}.</li>
+                </ul>
+              </li>
+              <li>Code inside a {#syntax#}nosuspend{#endsyntax#} scope does not cause the enclosing function to become an {#link|async function|Async Functions#}.</li>
               <li>See also {#link|Async Functions#}</li>
             </ul>
           </td>
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -10759,9 +10759,12 @@ fn readU32Be() u32 {}
             <pre>{#syntax#}nosuspend{#endsyntax#}</pre>
           </td>
           <td>
-            The {#syntax#}nosuspend{#endsyntax#} keyword.
+            The {#syntax#}nosuspend{#endsyntax#} keyword can be used in front of a block, statement or expression.
+            It marks a scope where async functions are not allowed to actually suspend execution.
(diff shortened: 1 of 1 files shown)


SpexGuy:
Thanks!  This is a good description, but as you noted it doesn't capture this detail.  This behavior is intentional, the actual requirement is that when control flow leaves the `bla()` function, the function has completed.  However this description might be too complicated to be the initial description.  Maybe keep what you wrote but add a note that this construct is allowed?

mguaypaq:
Done, and I reformatted the different points as a bullet list, so that this can be a sub-point attached to the point it modifies.

SpexGuy:
Thanks!  Sorry to nitpick, but the new description isn't quite correct.  Resuming from inside a suspend block does suspend, but then immediately recurses into a resume.  So when the resume returns, the function is complete, and then the suspend returns, but the function is complete so it's not a violation of nosuspend.  This behavior is pretty weird, for example:
```zig
fn foo() void {
    print("1");
    suspend {
        resume @frame();
        print("3");
    };
    print("2");
}
```
`nosuspend foo();` is valid and prints "123".

Repeated use of resume inside a suspend block can blow up the stack, so I think giving the impression that it's equivalent to not suspending is wrong enough to be slightly dangerous.

Maybe instead of saying "is not considered suspended", put something like "does not violate the requirements of nosuspend".

mguaypaq:
That's very legitimate, I wouldn't even call it a nitpick! Thanks for the explicit code example, I'll have to ponder it before I can write a better version of the doc.

Maybe we should turn your comments and that code example into a new subsection on `nosuspend` in the "async functions" section, rather than stuff all of this into the "keyword reference" section.

mguaypaq:
Ok, after doing some more experiments, I'm now more confused than ever. Also, I think I've stumbled across some unchecked undefined behavior. Can you clarify what the intended behavior is for the code snippets below?

I tried building tiny test programs for all the possible combinations of these axes:
* Different control flow primitives:
  * Calling an async function (that is, `_ = my_fun();`).
  * Just starting the function (that is, `var f = async my_fun();`).
  * Just getting the return value (that is, `_ = await f;`).
  * Reaching a `suspend;` statement.
  * Reaching a `resume f;` statement.
* How directly inside a `nosuspend` block it happens:
  * Directly within `nosuspend` from a regular function (`main`).
  * Indirectly, within an async function called from the `nosuspend` block.
* Whether the called/awaited/etc. async function has completed or not.

I noted whether the result was:
* It runs fine.
* Runtime error: `async function called in nosuspend scope suspended`.
* Compile error: `async call`/`suspend`/`resume in nosuspend scope`.
* Garbage values produced.

Most of the results were as expected, but I was surprised by:

1. Usually calling an async function like `_ = my_fun();` is the same thing as writing `_ = await async my_fun();`. But directly within a `nosuspend` block, the simple call is allowed and the verbose `await async` is disallowed (compile error).

2. Just starting an async function with `async` is a compile error directly within a `nosuspend` block, but works fine when done indirectly:
   ```zig
   // This runs fine, maybe surprisingly?

   pub fn main() void {
       nosuspend my_fun();
   }

   fn my_fun() callconv(.Async) void {
       var f = async other_fun();
   }

   fn other_fun() callconv(.Async) void {
       suspend;
   }
   ```

3. Similarly, a direct `resume` is a compile error but an indirect `resume` is silently allowed:
   ```zig
   // This runs fine, maybe surprisingly?

   pub fn main() void {
       var f = async suspender_fun();
       nosuspend resumer_fun(&f);
   }

   fn resumer_fun(f: anyframe) callconv(.Async) void {
       resume f;
   }

   fn suspender_fun() callconv(.Async) void {
       suspend;
       suspend;
   }
   ```

4. In a surprise twist, the reverse holds for `await`: awaiting a non-complete async function indirectly raises a runtime error (as it should be), but doing it directly within a `nosuspend` block silently returns a garbage value.
   ```zig
   // This runs but prints a garbage value.

   pub fn main() void {
       var f = async not_done_yet();
       var i = nosuspend await f;
       @import("std").debug.print("wat {}\n", .{i});
   }

   fn not_done_yet() callconv(.Async) u8 {
       suspend;
       return 2;
   }
   ```

I'm not sure how much of this is me not understanding the requirements for `nosuspend`, or what "control flow leaves the function" means, or the compiler not checking all undefined behavior.

Currently, the only part of my attempted documentation that I believe is that "Code inside a `nosuspend` scope does not cause the enclosing function to become an async function."

SpexGuy:
1 seems like a compiler bug, I think that's valid code.
2 is also weird, async is a synchronous operation and can never violate nosuspend. This should also be allowed directly.
3 is the same again, resume is a synchronous operation and can never violate nosuspend.
4 is an actual miscompilation, the data exists to perform that debug check so it should work.

These bugs may come from misunderstandings of what `nosuspend` does.  It was initially called `noasync`, which was confusing.

Can you please open issues for these?  They should be tracked separately from this PR.  I think 1-3 are probably all the same bug, so they could be grouped into one issue if you want.

mguaypaq:
Done, I opened two PRs for these problems. With your clarifications, I think I now understand what `nosuspend` should mean:

1. Directly inside a `nosuspend` block, `suspend` is disallowed, because that's definitely a suspension point.
2. Directly inside a `nosuspend` block, `await f` is disallowed if `f` has not yet completed, because that's also a suspension point.
3. Point 2 applies also to a call like `some_async_fn()`, because that's equivalent to `await async some_async_fn()`, which contains an `await`.
4. Everything else is allowed, including a direct `async f()`, a direct `resume`, and any indirect usage of `async`/`await`/`suspend`/`resume` inside of called async functions.

Is this accurate?

SpexGuy:
Yeah, I think that's all correct!

mguaypaq:
I think I fixed the description now. Also I noticed that there was a good spot to use `nosuspend` in one of the code samples in the main body of the documentation, and in doing so removed a possibly outdated or misleading comment.

andrewrk:
Thank you @mguaypaq!

PR #7974, opened 2021-02-07
std: Add std.fs.path.joinZ

fancl20:
Similar to mem.join and mem.joinZ.

### DIFF
--- a/lib/std/fs/path.zig
+++ b/lib/std/fs/path.zig
@@ -119,73 +119,40 @@ fn testJoinMaybeZPosix(paths: []const []const u8, expected: []const u8, zero: bo
 }
 
 test "join" {
-    testJoinMaybeZWindows(&[_][]const u8{}, "", false);
-    testJoinMaybeZWindows(&[_][]const u8{ "c:\\a\\b", "c" }, "c:\\a\\b\\c", false);
-    testJoinMaybeZWindows(&[_][]const u8{ "c:\\a\\b", "c" }, "c:\\a\\b\\c", false);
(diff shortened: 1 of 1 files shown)


lib/std/fs/path.zig
```diff
@@ -76,67 +77,115 @@ fn joinSep(allocator: *Allocator, separator: u8, sepPredicate: fn (u8) bool, pat
...
     defer testing.allocator.free(actual);
-    testing.expectEqualSlices(u8, expected, actual);
+    testing.expectEqualSlices(u8, expected, if (zero) actual[0 .. actual.len - 1 :0] else actual);
 }
 
 test "join" {
-    testJoinWindows(&[_][]const u8{ "c:\\a\\b", "c" }, "c:\\a\\b\\c");
-    testJoinWindows(&[_][]const u8{ "c:\\a\\b", "c" }, "c:\\a\\b\\c");
-    testJoinWindows(&[_][]const u8{ "c:\\a\\b\\", "c" }, "c:\\a\\b\\c");
+    testJoinMaybeZWindows(&[_][]const u8{}, "", false);
```
daurnimator, line 122: You might as well combine the two with:
```zig
for (&[_]bool{false, true}) |zero| {
    testJoinMaybeZWindows(&[_][]const u8{}, "", zero);
    ......
```

fancl20: I noticed the duplication here but not sure if we want to merge them into a single test case. Like mem.join and mem.joinZ, they're testing different behaviours. So what's the common practice for testing Z and non-Z functions?

PR #7975, opened 2021-02-07
zig fmt: implement slices, catch, and return

ifreund:


ifreund:
Note: `Slice` could be split into `Slice` and `Slice` sentinel to save 4 bytes when there is no sentinel. If you'd like that implemented just let me know, it would only take a minute or two.

andrewrk:
Hm nice catch - that is a pretty common operation, let's do it!

PR #7981, opened 2021-02-07
zig fmt: implement break and switches

ifreund:


lib/std/zig/ast.zig
```diff
@@ -1248,6 +1284,25 @@ pub const Tree = struct {
         });
     }
 
+    pub fn switchCaseOne(tree: Tree, node: Node.Index) Full.SwitchCase {
+        const data = &tree.nodes.items(.data)[node];
+        return tree.fullSwitchCase(.{
+            .values = if (data.lhs == 0) &.{} else @ptrCast([*]Node.Index, &data.lhs)[0..1],
```
ifreund, line 1290: note, it would be nice to be able to coerce single item pointers to a slice here.

andrewrk: I'll make the fixup myself but note that this is allowed:

```zig
@as(*[1]Node.Index, &data.lhs)
```

and that will in turn coerce to `[]const Node.Index`

PR #7985, opened 2021-02-08
translate-c: call @boolToInt on return value when necessary

### LINKED ISSUE #6215: translate-c: returning a boolean expression doesn't call @boolToInt
```c
int returns_bool() { return 1 && 1; }
```
currently translates to:
```zig
pub export fn returns_bool() c_int {
    return (true and true);
}
```

ehaas:
In C, if a function has return type `int` and the return expression
is a boolean expression, there is no implicit cast. Therefore the
translated Zig code needs to call @boolToInt() on the result.

Additionally, treat a Zig if/else with a boolean result in both bodies
as a boolean result. This allows translated C ternary expressions to be
recognized as boolean results if both arms are boolean expressions.

Fixes #6215

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -78,6 +78,10 @@ const Scope = struct {
         mangle_count: u32 = 0,
         lbrace: ast.TokenIndex,
 
+        /// When the block corresponds to a function, keep track of the return type
+        /// so that the return expression can be cast, if necessary
+        return_type: ?clang.QualType = null,
+
         fn init(c: *Context, parent: *Scope, labeled: bool) !Block {
             var blk = Block{
                 .base = .{
@@ -204,6 +208,21 @@ const Scope = struct {
                 .Root => unreachable,
                 .Block => return @fieldParentPtr(Block, "base", scope),
                 .Condition => return @fieldParentPtr(Condition, "base", scope).getBlockScope(c),
+                else => scope = scope.parent.?,
+            }
+        }
+    }
+
+    fn findBlockReturnType(inner: *Scope, c: *Context) ?clang.QualType {
+        var scope = inner;
+        while (true) {
+            switch (scope.id) {
+                .Root => return null,
+                .Block => {
+                    const block = @fieldParentPtr(Block, "base", scope);
+                    if (block.return_type) |qt| return qt;
+                    scope = scope.parent.?;
+                },
                 else => scope = scope.parent.?,
             }
         }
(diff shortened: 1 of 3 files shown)


src/translate_c.zig
```diff
@@ -29,6 +29,10 @@ const Scope = struct {
     id: Id,
     parent: ?*Scope,
 
+    /// When the scope corresponds to a function, keep track of the return type
+    /// so that the return expression can be cast, if necessary
+    return_type: ?clang.QualType = null,
+
```
Vexu, line 35: This should go in a `Scope.Block` and you should add a function to find the first block where this is not null.

src/translate_c.zig
```diff
@@ -1892,7 +1896,13 @@ fn isBoolRes(res: *ast.Node) bool {
         => return true,
 
         .GroupedExpression => return isBoolRes(@fieldParentPtr(ast.Node.GroupedExpression, "base", res).expr),
-
+        .If => {
+            const if_node = @fieldParentPtr(ast.Node.If, "base", res);
+            if (if_node.@"else") |else_node| {
+                return isBoolRes(if_node.body) and isBoolRes(else_node.body);
```
Vexu, line 1902: This will not work if only one branch results in a boolean:
```c
int foo(int a) {
    return a ? a == 2 : 1;
}
```
I think the correct way to handle this would be to do the casting in `transBinaryConditionalOperator` and `transConditionalOperator`.

src/translate_c.zig
```diff
@@ -2018,16 +2028,41 @@ fn transIntegerLiteral(
...
     const return_kw = try appendToken(rp.c, .Keyword_return, "return");
-    const rhs: ?*ast.Node = if (expr.getRetValue()) |val_expr|
+    var rhs: ?*ast.Node = if (expr.getRetValue()) |val_expr|
         try transExprCoercing(rp, scope, val_expr, .used, .r_value)
     else
         null;
+    if (zigShouldCastBooleanReturnToInt(rhs, scope.return_type)) {
+        const source_loc = @ptrCast(*const clang.Expr, expr).getBeginLoc();
+
+        const as_node = try rp.c.createBuiltinCall("@as", 2);
```
Vexu, line 2052: This `@as` is redundant as the compiler will automatically add an implicit cast here.

Vexu (changes requested):

src/translate_c.zig
ehaas: Ok, I reverted this one; maybe this should be addressed in a separate PR since it's really a different problem? I also found another unrelated case where boolToInt isn't called (initializing global variables)

Vexu: I already fixed that in #7479 and I can fix this there too.

src/translate_c.zig
```diff
@@ -209,6 +213,21 @@ const Scope = struct {
         }
     }
 
+    fn findBlockReturnType(inner: *Scope, c: *Context) ?clang.QualType {
+        var scope = inner;
+        while (true) {
+            switch (scope.id) {
+                .Root => return null,
```
Vexu, line 220: This should be unreachable byt I can fix that once I merge this to #7479.

Vexu (approved):

PR #7988, opened 2021-02-10
zig fmt: various fixes

ifreund:


ifreund:
`firstToken()` for pointer types with a `**` token is a little weird, since this token is shared between two separate pointer types. I don't think this will cause issues in practice though.

PR #7990, opened 2021-02-10
Add more mingw def files

marler8997:
I've added more of the ".def" files from mingw.  The list is based on all the libraries referenced by the win32metadata project. (see https://github.com/marlersoft/zigwin32).


SpexGuy:
This is a good change but it would really be nice to bundle all these files into a .zip and decompress them as needed into the global zig-cache.  (not in scope for this PR, obviously, but seems like a good move long term.)  Most of the install time on Windows is spent creating files, which takes ~30 milliseconds per file on an SSD.  So the nearly 400 new files in this PR will add ~12 seconds of install time on windows, which is not trivial.

andrewrk:
> This is a good change but it would really be nice to bundle all these files into a .zip and decompress them as needed into the global zig-cache.

Noted. I think that's a nice idea, let's plan to do it (feel free to open separate issue as a reminder). Agreed, let's not block this PR on the issue tho.

SpexGuy:
> Side note, why does creating files on windows take so long?

I'm really not sure, but in my tests I was unable to find any way to do it faster.  Using lots of threads to create files helped slightly but not much.  There must be a bottleneck around reading/writing NTFS metadata or something.

andrewrk:
30ms to create a file is just nuts. do you have antivirus software running or something?

SpexGuy:
I agree, I'm not running any antivirus besides the builtin Windows Defender though.  Actually these benchmarks were a couple months ago, I might be misremembering, they may have been my HDD, not my SSD.  The 30 mS number was time to call CreateFileW on an *existing* file, so I'm speculating that it's the same or worse for a new file.

Testing it again to get more accurate numbers, it takes me almost exactly two minutes to unzip a zig release on the SSD.  The current compiler has just under 10,000 files (mostly from libc installations), so that averages 12 mS per file.  That's a lot better than the 30 I was thinking, but still a crazy long amount of time to create a file.

SpexGuy:
For comparison with overall write speed, the time needed to copy an uncompressed archive of the installation (~220 MB) is under two seconds, so almost all of the two minutes is overhead of having many files.

marler8997:
It went from 2 minutes to 2 seconds!!

SpexGuy:
Standard SSD write speed is ~500 MB/sec, so 2 seconds is actually kind of slow 😄
That said, I forgot to take decompression time into account.  It takes about 6 seconds to compress the 220 MB archive file, and 3ish seconds to decompress it.  But that includes the copy time (because it's decompressing into a new file) so ~3 seconds seems like the theoretical bound here.

HDDs tend to have a write speed of about 100-150 MB/sec for linear access, so for 220 MB of data it should take 2-3 seconds to write it all as one contiguous file, plus 1 or 2 seconds for decompression.

PR #8007, opened 2021-02-12
std: Deprecate 'x'/'X'/'e'/'E' special cases for u8 slices

LemonBoy:
Let's follow the road paved by the removal of 'z'/'Z', the Formatter
pattern is nice enough to let us remove the remaining four special cases
and declare u8 slices free from any special casing!


marler8997:
Maybe `x` and `X` are common enough to remain?

LemonBoy:
No.

LemonBoy:
Rebased, note that this reverts #8031 as the deprecation warnings were added (and removed) within the same release.

andrewrk (approved):

lib/std/fmt.zig
```diff
@@ -709,6 +709,87 @@ fn formatFloatValue(
...
+
+/// Return a Formatter for a []const u8 where every byte is formatted as a pair
+/// of lowercase hexadecimal digits.
+pub fn fmtSliceHexLower(bytes: []const u8) std.fmt.Formatter(formatSliceHexLower) {
+    return .{ .data = bytes };
+}
+
+/// Return a Formatter for a []const u8 where every byte is formatted as a pair
+/// of uppercase hexadecimal digits.
+pub fn fmtSliceHexUpper(bytes: []const u8) std.fmt.Formatter(formatSliceHexUpper) {
```
daurnimator, line 744: I don't love that we have different functions for this for upper vs lower case.
Could we instead take the case as an argument?

`pub fn fmtSliceHex(bytes: []const u8, case: enum{Upper, Lower}) std.fmt.Formatter(formatSliceHex) {`

LemonBoy: No strong opinion on having a single formatter, I went with two separate ones to keep it close to the `x`\`X`.

(IIRC there's a stage1 bug regarding comptime anonymous enum arguments, check out debug.zig)

PR #8014, opened 2021-02-14
Fix fifo readableSliceMut bug in overflow case (and add tests)

rgreenblatt:
The overflow case wasn't properly handled because start may be non-zero. No test was hitting this case, so I added 2 which do.

### DIFF
--- a/lib/std/fifo.zig
+++ b/lib/std/fifo.zig
@@ -153,7 +153,7 @@ pub fn LinearFifo(
             var start = self.head + offset;
             if (start >= self.buf.len) {
                 start -= self.buf.len;
-                return self.buf[start .. self.count - offset];
+                return self.buf[start .. start + (self.count - offset)];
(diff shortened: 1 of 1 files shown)


lib/std/fifo.zig
```diff
@@ -153,7 +153,7 @@ pub fn LinearFifo(
             var start = self.head + offset;
             if (start >= self.buf.len) {
                 start -= self.buf.len;
-                return self.buf[start .. self.count - offset];
+                return self.buf[start .. start + (self.count - offset)];
```
daurnimator, line 156: Might be better written as `self.buf[start..][0..self.count - offset]`

rgreenblatt: Eh, I think I like the original approach better.

rgreenblatt:
Anything blocking this?

PR #8023, opened 2021-02-16
prebuild.zig

marler8997:
This is an idea to add support for a prebuild.zig file alongside build.zig.  prebuild.zig is a normal zig program with a "main" function.  Its always executed before "build.zig".  This gives projects the opportunity to hook into the Zig build system before their build.zig code is compiled.  This means it can be used to prepare anything required in order to get build.zig to compile.  The main use case for this is to prepare dependencies that build.zig requires.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -2451,19 +2451,28 @@ pub fn cmdBuild(gpa: *Allocator, arena: *Allocator, args: []const []const u8) !v
         const std_special = "std" ++ fs.path.sep_str ++ "special";
         const special_dir_path = try zig_lib_directory.join(arena, &[_][]const u8{std_special});
 
-        var root_pkg: Package = .{
-            .root_src_directory = .{
-                .path = special_dir_path,
-                .handle = try zig_lib_directory.handle.openDir(std_special, .{}),
+        const Kind = enum { prebuild, build };
+        const kind_basenames = [_][]const u8 { "prebuild.zig", "build.zig" };
+        var root_pkgs: [2]Package = [_]Package {
+            undefined,
+            .{
+                .root_src_directory = .{
+                    .path = special_dir_path,
+                    .handle = try zig_lib_directory.handle.openDir(std_special, .{}),
+                },
+                .root_src_path = "build_runner.zig",
             },
-            .root_src_path = "build_runner.zig",
         };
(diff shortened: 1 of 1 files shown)


marler8997:
@MasterQ32 has informed me that a more flexible way to do this already exists, namely, using `build.zig` to compile a custom build_runner.  I'll close this and try that method.

marler8997:
I experimented with @MasterQ32 's suggested solution and it looks like it doesn't really work either.  I split `build.zig` into 2 files.  `preppedbuild.zig` which contains the project build logic, and `build.zig` which sets up the dependencies for `preppedbuild.zig` and then invokes it.  Using this method means I can't use the `build` entry point in `build.zig` because that entry point doesn't have access to all the information, nor all the code paths that I need to forward all commands to the new `preppedbuild.zig`.  So the solution is to use `zig run build.zig` and disable `zig build` with an error message like this:

```zig
pub fn build(b: *std.build.Builder) !void {
    std.debug.print("Error: please run 'zig run build.zig'\n", .{});
    std.os.exit(1);
}
pub fn main() !u8 {

    // ... prepare dependencies ...

    // now call 'zig build --build-file preppedbuild.zig'
    var args= std.ArrayList([]const u8).init(allocator);
    build_args.append("zig"); // PROBLEM: how do we get the zig executable here?  It's gone
    build_args.append("build");
    build_args.append("--build-file");
    build_args.append("preppedbuild.zig");
}
```

As you can see from the comment above, the problem with this is once this executable is running, there's no way to get the `zig` executable we were compiled with.  Furthermore, in concept, and executable running with `zig run` should be no different than a pre-compiled binary running on it's own, therefore, I don't think we should try to pass this information through a side channel like an environment variable to the resulting executable of a `zig run` command.

I think the `zig` executable needs to be involved throughout the entire build process here.  However, the drawback of this PR in its current state is there's no way to add packages to be used with `build.zig`.  I have one idea which is to allow `prebuild.zig` to output a list of packages that can be used to compile `build.zig` with.  Maybe `prebuild.zig` can write these to `stdout`?

marler8997:
I think I've come up with a better idea. I'm going to close this again and possibly create a new PR with a prototype for the new idea.

PR #8026, opened 2021-02-16
ci: zig fmt --check

g-w1:
I also ran zig fmt on all the invalid files and deleted a file that was not valid zig (test/stage1/behavior/bugs/4769_c.zig). See https://freenode.irclog.whitequark.org/zig/2021-02-16#1613498571-1613499112; for the latter.


g-w1:
Closing because I should do it after #7920

ifreund:
@g-w1 Now that #7920 has landed and we have much more finely grained `// zig fmt: (off|on)` support I think this should be reconsidered.

ci/azure/linux_script
```diff
@@ -69,6 +69,12 @@ release/bin/zig build test -Denable-qemu -Denable-wasmtime
 
 # Look for HTML errors.
 tidy -qe ../zig-cache/langref.html
+# look for formatting errors
+release/bin/zig fmt --check ../lib/std
+release/bin/zig fmt --check ../src
+release/bin/zig fmt --check ../tools
+release/bin/zig fmt --check ../test
+release/bin/zig fmt --check ../build.zig
```
andrewrk, line 77:
```suggestion
release/bin/zig fmt --check ../lib/std ../src ../tools ../test ../build.zig
```

andrewrk (commented):
OK, let's do it, zig fmt --check in the CI. This patch needs some fixes though:

You're missing a change to behavior.zig to not import the deleted file. In fact, all of the 4769 test cases should be removed from behavior tests, since they are testing the tokenizer only. The tokenizer tests belong in `lib/std/zig/tokenizer.zig`. Deleting the invalid zig file is good but we should move the test to become a tokenizer test for it. None of this is your fault, but would you mind cleaning this up here since it's blocking progress on this PR?

Thanks for doing this. I hope it makes contributing more ergonomic.

Issue #8000, opened 2021-02-12
Allow async and resume inside nosuspend blocks


mguaypaq:
As discussed in #7972 (specifically, [this comment](https://github.com/ziglang/zig/pull/7972#issuecomment-777233893) by @SpexGuy), the zig compiler currently (version 0.7.1) forbids the `async` and `resume` keywords within a `nosuspend` block. However, starting an async function call with the `async` keyword or resuming an async function call with the `resume` keyword are both synchronous operations, so they should be permitted within a `nosuspend` block.

Sample code which exhibits the problem:
```zig
// Both of these operations should be permitted inside `nosuspend`.

pub fn main() void {
  // Compile error: async call in nosuspend scope
  var f = nosuspend async some_async_fn();

  // Compile error: resume in nosuspend scope
  nosuspend resume f;
}

fn some_async_fn() void {
  suspend;
  suspend;
}
```

The code which generates these error messages is in `src/stage1/ir.cpp` (sorry, the file is too big to give a direct link to the specific lines in github, but the error messages appear verbatim).

The [test case for this](https://github.com/ziglang/zig/blob/d3565ed6b48c9c66128f181e7b90b5348504cb3f/test/compile_errors.zig#L1020-L1033) should also be updated.

I'm happy to slowly figure out how to make this change myself and/or let someone more experienced beat me to it.

PR #8028, opened 2021-02-17
stage1: allow async and resume inside nosuspend blocks

mguaypaq:
This is to address issue #8000.

My biggest worry is that this could invalidate some assumptions I don't know about. In particular, I may be missing something fundamental because I don't understand why the `CallModifier` enum is an enum rather than a collection of flags; are the different modifiers really mutually exclusive?

Based on looking at just `stage1/ir.cpp` and `stage1/codegen.cpp`, it looks like `CallModifierAsync` should take priority over `CallModifierNoSuspend`.

When modifying the test case, I only removed the error messages, but left the `async` call and `resume` in the code fragment being compiled, to serve as a test that errors are **not** generated anymore. I don't know if that part of the test should be moved somewhere other than `test/compile_errors.zig`.

### DIFF
--- a/src/stage1/ir.cpp
+++ b/src/stage1/ir.cpp
@@ -10095,10 +10095,6 @@ static IrInstSrc *ir_gen_fn_proto(IrBuilderSrc *irb, Scope *parent_scope, AstNod
 
 static IrInstSrc *ir_gen_resume(IrBuilderSrc *irb, Scope *scope, AstNode *node) {
     assert(node->type == NodeTypeResume);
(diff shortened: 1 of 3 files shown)


Vexu:
`nosuspend resume` works as expected and so does `nosuspend async` for the most part but trying to resume a frame created by it hits this assertion in codegen.cpp:
https://github.com/ziglang/zig/blob/0447a2c041a4be843251396e668e074186aa49a2/src/stage1/codegen.cpp#L1794

andrewrk:
Nice work!

Issue #8025, opened 2021-02-16
@tryImport


marler8997:
Currently we have `@import`:
```zig
@import(comptime path: []u8) type
```

I've thought of a solution to the `build.zig` bootstrap problem if we had `@tryImport`:

```zig
@tryImport(comptime path: []u8) ?type
```

This builtin would function the same way as `@import` except if the module isn't found, rather than a compile error it returns `null`.

This can be used in a `build.zig` file that requires other modules that may or may not be available on the initial build.

`build.zig`:
```zig
const coolbuildlib = @tryImport("coolbuildlib") orlese {
    @compileError("Error: coolbuildlib is not available, please specify it by providing --pkg-begin coolbuildlib PATH --pkg-end.");
};
```

Here's an example where `build.zig` detects when `mycoolbuildlib` is missing, resolves it and recompiles itself:

`build.zig`:
```zig
pub fn build(b: *Builder) !void {
    if (@tryImport("coolbuidlib")) |coolbuildlib| { 
        //
        // insert normal build logic using coolbuildlib
        //
    } else {
        const next_build = b.addRecompileBuild();
        next_build .addPackage(
            .name = "coolbuildlib",
            .path =findCoolBuildLib(),
        );
    }
}
```

mlarouche:
It would be cool if this proposal was combined with a way to list files at comptime. It would help if you contain code that are covered in NDA (ex: game console SDK) and would be only included if the folder is actually present on disk.

BinaryWarlock:
Yeah, a `@listDir` would be great as well as a `@tryEmbedFile`

zzyxyzz:
Actually, it would be nice to have a systematic way of catching compile errors. That would take care of both `@tryImport` and `@tryEmbedFile`, and more besides. With metaprogramming in particular, it could be quite a lifesaver. E.g., in a recent issue I had the problem of how to implement a recursive deinit method for a generic container. The idea was to try and see if the child type has a recursive deinit method and call that, otherwise call an ordinary deinit method, or do nothing if the type has neither (or doesn't support methods at all, such as an integer). A rigorous solution to that problem would require some fairly involved compiletime introspection, but if compile errors could be recovered from, the method could be written in a very short and understandable way:

```zig
fn deinitAll(this: *@This()) void {
    for (this.items) |x|
        x.deinitAll() comptimecatch x.deinit() comptimecatch {};
    this.deinit();
}
```

On the other hand, this sort of thing might open an amazing can of worms...

[...]

PR #8033, opened 2021-02-18
add @tryImport

marler8997:
An implementation for the proposed `@tryImport` builtin: https://github.com/ziglang/zig/issues/8025


Rocknest:
It think you are trying to solve a problem that is going to be solved by a package manager (someday). If this is supposed to be a temporary solution then why not just `@hasPackage`?

marler8997:
> It think you are trying to solve a problem that is going to be solved by a package manager (someday). 

The `build.d bootstrap problem` which I describe here (https://github.com/ziglang/zig/issues/8025#issue-809378278) is a fundamental problem with `build.zig` regardless of whether a package manager exists.  A package manager may implement one of the solutions I described in the linked issue above, however, they all have problems of their own because because they don't allow `build.zig` to do its job.  They all rely on other tools and/or non-zig configuration to resolve dependencies needed by `build.zig`.  Using `@tryImport` allows `build.zig` to work with any dependency using the full power of Zig to resolve them.  This is not an alternative to a package manager, it's orthogonal.

> If this is supposed to be a temporary solution then why not just @hasPackage([]const u8) bool?

This is permanent solution to a fundamental problem.  Using `@hasPackage` limits the feature to only working with package imports, but the "build.d bootstrap" problem applies to both package and file imports. Another variation I considered was `@hasImport`.  I found it is functionally equivalent to `@tryImport`, both can be used to implement the other, so the question becomes which is more common.  `@tryImport` wins if it's more common to perform an import when checking for its existence.

Vexu:
I'll close this pr for now since we want to reserve the pull request queue for prs that are ready to be merged. I'll add a reminder to the issue about this implementation.

PR #8035, opened 2021-02-18
libstd: add static archive header struct and definitions

kubkon:
I'm not sure whether this is the best place to host this, so looking for opinions!

cc @LemonBoy

### DIFF
--- a/lib/std/os/bits.zig
+++ b/lib/std/os/bits.zig
@@ -53,3 +53,42 @@ pub const LOG_NOTICE = 5;
 pub const LOG_INFO = 6;
 /// debug-level messages
 pub const LOG_DEBUG = 7;
+
+// `ar` archive file format definitions
+
+pub usingnamespace if (std.Target.current.os.tag == .windows) struct {} else struct {
+    // Archive files start with the ARMAG identifying string.  Then follows a
+    // `struct ar_hdr', and as many bytes of member file data as its `ar_size'
+    // member indicates, for each member file.
+    /// String that begins an archive file.
+    pub const ARMAG: *const [SARMAG:0]u8 = "!<arch>\n";
+
+    /// Size of that string.
+    pub const SARMAG: u4 = 8;
+
+    /// String in ar_fmag at the end of each header.
+    pub const ARFMAG: *const [2:0]u8 = "`\n";
+
+    pub const ar_hdr = extern struct {
+        /// Member file name, sometimes / terminated.
+        ar_name: [16]u8,
+
+        /// File date, decimal seconds since Epoch.
+        ar_date: [12]u8,
+
+        /// User ID, in ASCII format.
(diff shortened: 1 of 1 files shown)


LemonBoy:
`os.zig` is not the right place, this stuff should get its own file a-la `elf.zig`/`coff.zig`.
Keep in mind there are a few variants of the archive format, LLVM exposes them as `object::Archive::K_{GNU,DARWIN,BSD}`.

kubkon:
> `os.zig` is not the right place, this stuff should get its own file a-la `elf.zig`/`coff.zig`.

Good idea, thanks!

> Keep in mind there are a few variants of the archive format, LLVM exposes them as `object::Archive::K_{GNU,DARWIN,BSD}`.

Huh, now I'm intrigued. I was convinced there was literally no difference between GNU and DARWIN at least. Do you have a link handy perhaps?

LemonBoy:
> Huh, now I'm intrigued. I was convinced there was literally no difference between GNU and DARWIN at least. Do you have a link handy perhaps?

Here you [go](https://www.freebsd.org/cgi/man.cgi?query=ar&sektion=5), the DARWIN format is derived from the BSD one with some extra record, while there's also a Windows' variant that's similar to GNU.

kubkon:
> > Huh, now I'm intrigued. I was convinced there was literally no difference between GNU and DARWIN at least. Do you have a link handy perhaps?
> 
> Here you [go](https://www.freebsd.org/cgi/man.cgi?query=ar&sektion=5), the DARWIN format is derived from the BSD one with some extra record, while there's also a Windows' variant that's similar to GNU.

Hmm, as far as I can tell, the `ar_hdr` for GNU and Darwin are the same: [gnu](https://github.com/ziglang/zig/blob/bb636cb3bffadf5a03e592b97a1968492f16f8cd/lib/libc/include/generic-glibc/ar.h) and [darwin](https://opensource.apple.com/source/Libc/Libc-1439.40.11/include/ar.h.auto.html), or am I missing something?

kubkon:
Oh unless you meant some additional `const`s?

kubkon:
OK, I'm gonna go ahead and close it for now. After some offline discussion with @andrewrk and others, I think it makes more sense to have AR part of the compiler sources itself rather than the standard library.

LemonBoy:
Please check out the link I've posted, it explains pretty well the main differences between the GNU and the BSD format.
Darwin uses the BSD format but adds its own way of defining the symbol table (`SYMDEF_64`), AFAIR that's the main difference.

kubkon:
> Please check out the link I've posted, it explains pretty well the main differences between the GNU and the BSD format.
> Darwin uses the BSD format but adds its own way of defining the symbol table (`SYMDEF_64`), AFAIR that's the main difference.

Will do, thanks!

PR #7248, opened 2020-11-28
std: Avoid deadlock in the signal handler

LemonBoy:
stderr_mutex may still be held when we reach the signal handler, grab
our own stderr handle to print the error messages and avoid deadlocking.

Closes #7247

PR #8058, opened 2021-02-23
Windows Debug Printing

### LINKED ISSUE #7859: Error messages should use a mutex
The compiler recently gave me this mess as part of an error message:
```
Segmentation fault at address 0x7ff651e04988
reached unreachable code
C:\[path to my project]\src\asset manager.zig:408:53: 0x7ff651e04988 in asset manager.init (dual soul game.obj)
    full_mC:\[path to my compiler]\zig-windows-x86_64-0.6.0+352976ed2\lib\zig\std\debug.zige:t225ad:14ata : = 0xt7ff651d61708r in ystd.debug.assert  (adual soul game.objl)loc
. a l l oifc ((*!Assoekt)Met audnatreacah,a btlrey ;st r/eam/.re adaLeb128(s)s)e;r
tio                                                    n^ f
ailure
             ^
```
The cause of this mess is obvious: Multiple threads are crashing at the same time, and are all trying to talk to me at once instead of taking turns. This could be easily solved with a mutex.

Normally, using a mutex without the user's consent could be a problem, but in this case, the entire program is likely to fall apart, anyway, so waiting wouldn't be too much of a problem. However, we could still consider using the same mutex for printing in non-fatal scenarios. If this were to be done, I'd suggest making this an option for the user. Obviously, no mutex would be used for any of this if the whole build is single-threaded.

gsquire:
I may have trivialized this bug but I figured I'd give it a shot.

Fixes #7859

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -1861,6 +1861,8 @@ fn handleSegfaultWindowsExtra(info: *windows.EXCEPTION_POINTERS, comptime msg: u
         const regs = info.ContextRecord.getRegs();
         // Don't use std.debug.print() as stderr_mutex may still be locked.
         nosuspend {
+            const held = stderr_mutex.acquire();
+            defer held.release();
             const stderr = io.getStdErr().writer();
             _ = switch (msg) {
                 0 => stderr.print("{s}\n", .{format.?}),


LemonBoy:
Scroll up a few lines, `handleSegfaultLinux` explains why `print` (and `stderr_mutex`) are not used.
The segfault handlers should be serialized with a different mutex.

gsquire:
Ah, okay. I guess I'm confused as to why we need another lock then and what it will synchronize.

LemonBoy:
> I guess I'm confused as to why we need another lock then and what it will synchronize.

I thought the comment I added was pretty clear:
> Don't use std.debug.print() as stderr_mutex may still be locked.

Thought experiment, what happens if a segfault happens while `stderr_mutex` is locked?

gsquire:
Deadlock!

LemonBoy:
And that brings us back to #7248 :)

gsquire:
Okay so we need to be a bit more clever. I'll keep digging, sorry for the confusion.

LemonBoy:
NP, that stuff is quite untested and breaks from time to time.
I wonder why we don't use `panicExtra` instead of duplicating a lot of code...

gsquire:
I don't think I know enough about what output needs to be synchronized here to finish this. I'll close this out so others can take a peek at it.

PR #8060, opened 2021-02-24
Add CPU feature check to standardTargetOptions

DrDeano:
If there is a mismatch of CPU features provided
compared to the whitelist, then will fail the build and
print what the expected CPU model is and the feature
set for the model. Also prints what features need to be
removed.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -640,12 +640,18 @@ pub const Builder = struct {
         const mcpu = self.option([]const u8, "cpu", "Target CPU");
 
         const triple = maybe_triple orelse return args.default_target;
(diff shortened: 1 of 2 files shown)


MasterQ32:
I don't think adding this to the build system is the right place. This will not work when you use `zig build-exe -mcpu foobar`, only when you use a build script via `zig build`

DrDeano:
Yes, but `zig build-exe` can't take a whitelist of targets and this is meant for the build script

andrewrk:
Thank you @DrDeano and apologies for the delayed merge. Please note I applied a fixup commit on top of this before merging.

PR #8063, opened 2021-02-25
update docs and grammar to allow CRLF line endings

thejoshwolfe:
According to https://github.com/ziglang/zig/wiki/FAQ#why-does-zig-force-me-to-use-spaces-instead-of-tabs , CRLF line endings and hard tabs are supposed to be supported. This PR updates the docs to state that, and also adds a warning that stage1 doesn't comply yet. The docs are clear that CRLF line endings and hard tabs are discouraged, and zig fmt will fix them for you.

Instead of linking to a (now out dated) discussion thread, I put some of the rationale into the language reference.

I made some kind of attempt to fix the grammar to allow tabs and windows line endings (and not allow CR anywhere else), but I may have made a mistake. However, I'm pretty sure the grammar is for humans and not machines, so I don't think it matters if it's perfect yet.

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -10447,13 +10447,40 @@ fn readU32Be() u32 {}
       {#header_close#}
       {#header_open|Source Encoding#}
       <p>Zig source code is encoded in UTF-8. An invalid UTF-8 byte sequence results in a compile error.</p>
-      <p>Throughout all zig source code (including in comments), some codepoints are never allowed:</p>
+      <p>Throughout all zig source code (including in comments), some code points are never allowed:</p>
       <ul>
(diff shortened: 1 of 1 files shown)


SpexGuy:
How does this interact with multiline string literals?  Do all newlines become `\n` in the string, or does it match the file?  Probably worth a note in the docs.

g-w1:
A pr should also be made here to update the grammar. https://github.com/ziglang/zig-spec

thejoshwolfe:
oh! the grammar is real :open_mouth: 

do we have some kind of sync mechanism to keep the readme and the zig-spec repo on the same page? there's a lot of HTML escaping that seems to have been done automatically.

i'll also add a link at the top of the grammar saying that its source of truth is the other repo.

SpexGuy:
> do we have some kind of sync mechanism

None that I know of.  There are some tests in the other repo that make sure the grammar matches the actual compiler, but they aren't very thorough and currently only support linux.

PR #8064, opened 2021-02-25
translate-c: Add support for pointer subtraction

### LINKED ISSUE #7216: Subtracting a c array from a pointer to that array yields error
Hi, in my C library I have the following code:

```C
	char description[2048] = { 0 };
	DEBUGINATOR_strcpy_s(description, 16, "Preset: \n");
	char* description_end = description + DEBUGINATOR_strlen(description);
	const char** path_part = paths;
	const char** value_title_part = value_titles;
	for (int i = 0; i < num_paths; i++) {
		DEBUGINATOR_strcpy_s(description_end, sizeof(description) - (description_end - &(description[0])), *path_part);
```

Error:

```
.\zig-cache\o\a4d06eab64bef73eb2c825d1dab7333b\cimport.zig:740:5: error: expected type 'usize', found '*[2048]u8'
    var description: [2048]u8 = [1]u8{
    ^
.\zig-cache\o\a4d06eab64bef73eb2c825d1dab7333b\cimport.zig:750:146: note: referenced here
            debuginator_strcpy_s(description_end, @bitCast(c_int, @truncate(c_uint, (@sizeOf([2048]u8) -% @bitCast(c_ulonglong, (description_end - &description))))), path_part.?.*);
                                                                          
```

Changing the strcpy line to the following fixes the compile error:
``` C		
DEBUGINATOR_strcpy_s(description_end, sizeof(description) - (description_end - &(description[0])), *path_part);
``` 

That is, changing `- description` to `- &(description[0])`

UPDATE: This didn't help, it just changed the error a bit.

```
.\zig-cache\o\ddea80e39b2b9cbe41c2d7e436a4b50c\cimport.zig:750:161: error: expected type 'usize', found '*u8'
            debuginator_strcpy_s(description_end, @bitCast(c_int, @truncate(c_uint, (@sizeOf([2048]u8) -% @bitCast(c_ulonglong, (description_end - &(description[@intCast(c_uint, @as(c_int, 0))])))))), path_part.?.*);
                                                                                                                                                                ^
```

ehaas:
When two pointers are subtracted, both shall point to elements of the
same array object, or one past the last element of the array object;
the result is the difference of the subscripts of the two array elements.

The size of the result is implementation-defined, and its type
(a signed integer type) is ptrdiff_t defined in the <stddef.h> header.
If the result is not representable in an object of that type,
the behavior is undefined.

See C Standard, §6.5.6 [ISO/IEC 9899:2011]

Fixes #7216

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -1083,6 +1083,7 @@ fn transBinaryOperator(
 ) TransError!Node {
     const op = stmt.getOpcode();
     const qt = stmt.getType();
+    const isPointerDiffExpr = cIsPointerDiffExpr(c, stmt);
     switch (op) {
         .Assign => return try transCreateNodeAssign(c, scope, result_used, stmt.getLHS(), stmt.getRHS()),
         .Comma => {
@@ -1143,7 +1144,7 @@ fn transBinaryOperator(
             }
         },
         .Sub => {
-            if (cIsUnsignedInteger(qt)) {
+            if (cIsUnsignedInteger(qt) or isPointerDiffExpr) {
                 op_id = .sub_wrap;
             } else {
                 op_id = .sub;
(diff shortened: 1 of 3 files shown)


ehaas:
@Vexu maybe you have some thoughts on this:

With this PR, this C:
```c
int foo[10];
int *x = &foo[1];
int *y = &foo[5];
ptrdiff_t diff = y - x;
```
turns into this Zig:
```zig
var foo: [10]c_int = undefined;
var x: [*c]c_int = &foo[@intCast(c_uint, @as(c_int, 1))];
var y: [*c]c_int = &foo[@intCast(c_uint, @as(c_int, 5))];
var diff: ptrdiff_t = @divExact(@bitCast(c_long, @ptrToInt(y) -% @ptrToInt(x)), @sizeOf(c_int));
```

One potential problem is that it assumes that the C-language pointer-subtraction result type (`ptrdiff_t`, which is platform dependent) is the same size as Zig's `usize` (result of the subtraction of two `@ptrToInt`). However I'm not sure how to determine how those two compare to each other, to know whether I need an `@as` or `@truncate` before bitcasting. It seems like it will potentially be platform dependent since I don't think the C standard guarantees that `ptrdiff_t` and `size_t` have the same bit width).

Vexu:
A quick look in `lib/libc` suggests that most platforms define `ptrdiff_t` and `size_t` to be the same so we should be good but if it ever does become an issue we can just add a `@intCast` before the `@bitCast`.

Vexu (approved):

PR #8078, opened 2021-02-26
stage1: Fix emission of sret annotation for LLVM

### LINKED ISSUE #8075: llvm12 branch: segfault reduction
edit: further reduction

On llvm12 branch function `codegen.aarch64.Instruction.pcRelativeAddress` is triggering a segfault while using `zig0` to build-obj `zig1.o` and now we have a reduction:

#### reduction.zig
```zig
//    fn pcRelativeAddress(rd: Register, imm21: i21, op: u1) Instruction {
//        assert(rd.size() == 64);
//        const imm21_u = @bitCast(u21, imm21);
//        return Instruction{
//            .PCRelativeAddress = .{
//                .rd = rd.id(),
//                .immlo = @truncate(u2, imm21_u),
//                .immhi = @truncate(u19, imm21_u >> 2),
//                .op = op,
//            },
//        };
//    }

export fn entry() void {
    _ = doit();
}

fn doit() Instruction {
    var num: u21 = 42;
    return Instruction{
        .value = @truncate(u19, num >> 2),
    };
}

// remove keyword `packed` to avoid segfault
const Instruction = packed struct {
    value: u19,
};

pub fn panic(msg: []const u8, error_return_trace: ?*@import("builtin").StackTrace) noreturn {
    while (true) {}
}
```

##### `zig0` command
```sh
_debug/zig0 reduction.zig -target native -mcpu=baseline --name reduction --override-lib-dir lib -femit-bin=reduction.o -lc --pkg-begin build_options _debug/config.zig --pkg-end --pkg-begin compiler_rt lib/std/special/compiler_rt.zig --pkg-end
```

LemonBoy:
LLVM12 deprecated `sret` and replaced it with the `sret(<Ty>)` form.

Closes #8075

cc @mikdusan

### DIFF
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -549,7 +549,7 @@ static LLVMValueRef make_fn_llvm_value(CodeGen *g, ZigFn *fn) {
         } else if (want_first_arg_sret(g, &fn_type->data.fn.fn_type_id)) {
             // Sret pointers must not be address 0
             addLLVMArgAttr(llvm_fn, 0, "nonnull");
-            addLLVMArgAttr(llvm_fn, 0, "sret");
+            ZigLLVMAddSretAttr(llvm_fn, 0, get_llvm_type(g, return_type));
             if (cc_want_sret_attr(cc)) {
                 addLLVMArgAttr(llvm_fn, 0, "noalias");
             }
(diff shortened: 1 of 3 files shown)


andrewrk:
Oops, I just did this same work before seeing this. Your patch looks better than mine; reverting and merging.

andrewrk:
Thanks!

PR #8084, opened 2021-02-27
Rework synchronization primitives - Futex edition

kprotty:
See #7800 for initiative and reason for branching into new PR. Don't intend to really add more changes ATM but am willing to accept anymore that come in.


lib/std/time.zig
```diff
@@ -55,32 +88,29 @@ pub fn sleep(nanoseconds: u64) void {
...
-/// The return value is signed because it is possible to have a date that is
-/// before the epoch.
-/// See `std.os.clock_gettime` for a POSIX timestamp.
+/// The return value is signed due to possibly being behind the UTC epoch.
 pub fn timestamp() i64 {
     return @divFloor(milliTimestamp(), ms_per_s);
 }
 
-/// Get a calendar timestamp, in milliseconds, relative to UTC 1970-01-01.
+/// Get a calendar timestamp, in nanoseconds, relative to UTC 1970-01-01.
```
daurnimator, line 102: `milliTimestamp` returns `nanoseconds`?

daurnimator (changes requested):

lib/std/sync.zig
```diff
@@ -0,0 +1,27 @@
...
+const root = @import("root");
+
+pub const atomic = @import("./sync/atomic.zig");
+pub const futex = @import("./sync/futex.zig");
+pub const primitives = @import("./sync/primitives.zig");
+
+pub usingnamespace if (@hasDecl(root, "sync"))
+    root.sync
+else if (std.builtin.single_threaded)
+    primitives.debug
```
daurnimator, line 17: Why is this `debug` rather than `core`?

lib/std/sync/futex.zig
```diff
@@ -0,0 +1,79 @@
...
+    };
+
+    inline for (.{
+        .{os},
+        .{spin},
+        .{event},
+        .{generic.forFutex(os)},
+        .{generic.forFutex(spin)},
+    }) |futex| {
+        // @compileError("TODO: test wait/wake/nanotime");
```
daurnimator, line 77:
```suggestion
        return error.SkipZigTest; // TODO: test wait/wake/nanotime
```

lib/std/sync/futex/generic.zig
```diff
@@ -0,0 +1,274 @@
...
+// This file is part of [zig](https://ziglang.org/), which is MIT licensed.
+// The MIT license requires this copyright notice to be included in all copies
+// and substantial portions of the software.
+
+const std = @import("../../std.zig");
+const atomic = @import("../atomic.zig");
+
+const builtin = std.builtin;
+const assert = std.debug.assert;
+const helgrind: ?type = if (builtin.valgrind_support) std.valgrind.helgrind else null;
```
daurnimator, line 12: You don't need this check, the valgrind module already does it.

lib/std/time.zig
kprotty: good point

lib/std/sync.zig
kprotty: core is for algorithms, the "Debug" impls are the "serial" ones. Unfortunate naming.

lib/std/time.zig
```diff
@@ -89,202 +119,297 @@ pub fn nanoTimestamp() i128 {
...
+}
+
+/// Returns true if the platform supports 64-bit atomic operations.
+fn supports64bitAtomics() bool {
+    // TODO: check if this is actually true.
+    return @sizeOf(usize) >= @sizeOf(u64);
+}
+
+/// Returns true if the system's high performance timer implementation is actually monotonic.
+/// For some platforms, this is not always the case: https://doc.rust-lang.org/src/std/time.rs.html#227
```
komuw, line 224:
```suggestion
/// For some platforms, this is not always the case: https://doc.rust-lang.org/1.50.0/src/std/time.rs.html#227
```

komuw: future versions of rust may move that file/code, so it is better(IMO) to link to a permlink

lib/std/os/bits/darwin.zig
```diff
@@ -1737,3 +1737,56 @@ pub const IOCPARM_MASK = 0x1fff;
 fn ior(inout: u32, group: usize, num: usize, len: usize) usize {
     return (inout | ((len & IOCPARM_MASK) << 16) | ((group) << 8) | (num));
 }
+
+pub const KERN_SUCCESS = 0;
```
daurnimator, line 1741: This looks like it should be an enum

lib/std/sync/futex/generic.zig
```diff
@@ -0,0 +1,274 @@
...
+            waiters: usize = 0,
+            tree: WaitTree = .{},
+
+            var array = [_]WaitBucket{WaitBucket{}} ** bucket_count;
+
+            /// Hash a address to a wait-bucket.
+            /// This uses the same method as seen in Amanieu's port of WTF::ParkingLot:
+            /// https://github.com/Amanieu/parking_lot/blob/master/core/src/parking_lot.rs
+            fn from(address: usize) *WaitBucket {
+                const seed = @truncate(usize, 0x9E3779B97F4A7C15);
```
daurnimator, line 30: huh... googling this number shows that it is `2^64 * phi` and seems to have a long history of use in wait queues

e.g. https://softwareengineering.stackexchange.com/a/402543

lib/std/sync/primitives/Condvar.zig
```diff
@@ -0,0 +1,218 @@
...
+                deadline,
+            );
+
+            if (helgrind) |hg| {
+                hg.annotateHappensAfter(@ptrToInt(self));
+            }
+        }
+
+        pub fn notifyOne(self: *Self) void {
+            if (helgrind) |hg| {
```
daurnimator, line 64: As mentioned in another comment: no need for this check.

kubkon:
@kprotty could you provide status update on this PR, i.e., do you plan on doing more work on this branch and hopefully merging this upstream in the near future? Or is there a blocker for it or something that I might have missed? I'd be very excited to see this merged.

kprotty:
@kubkon Late response, but there are a few blockers:
* This is old code (an updated version was written after) and instead of discarding it, I was convinced to make it a PR in case someone else wanted to build on top of it. I would prefer to rewrite it all personally.
* RwLock implementation and related tests are missing
* Testing for event loop backend of sync primitives is flaky as the test code racy segfaults with std.testing.allocator and reports data races with tsan but cannot be sure if its my evented backend impl or the event loop itself.

kubkon:
> @kubkon Late response, but there are a few blockers:
> 
> * This is old code (an updated version was written after) and instead of discarding it, I was convinced to make it a PR in case someone else wanted to build on top of it. I would prefer to rewrite it all personally.
> * RwLock implementation and related tests are missing
> * Testing for event loop backend of sync primitives is flaky as the test code racy segfaults with std.testing.allocator and reports data races with tsan but cannot be sure if its my evented backend impl or the event loop itself.

Ah, gotcha. Thanks for the explanation!

Vexu:
Is this going to be mergeable anytime soon? If not it woud probably be better to close it so that it isn't cluttering the pull requests.

andrewrk:
@kprotty you might need to strategize a bit so as to not get overwhelmed and figure out how to break this into smaller, independently mergeable pieces

kprotty:
> Is this going to be mergeable anytime soon?

@Vexu Most likely no. Its ok to close this given:

> figure out how to break this into smaller, independently mergeable pieces

@andrewrk I've been taking this advice. Introduced atomic stuff in #8866, adding Futex in #9070, sync primitives based on Futex after that, then `time.Duration/Instant` most likely after. Difference between this PR and the current approach is that I won't be touching `std.event` stuff until #8224 is decided.

PR #8085, opened 2021-02-27
HashMap.put returns !void, not a !bool

daurnimator:
Should assigning `void` to `_` be a compile error? obviously this is a mistake people make...


marler8997:
> Should assigning void to _ be a compile error? obviously this is a mistake people make...

Seems like the right thing to do..

LemonBoy:
> Should assigning void to _ be a compile error?

That'd be awful for generic code, you'd have to write `if (return is void) foo() else _ = foo()`.
Assigning `void` values to `_` is a crime without victims and make `_` behaviour consistent, it sinks every kind of value regardless of its type.

Issue #4762, opened 2020-03-19
ability to detect when a C macro is @fieldParentPtr


andrewrk:
[use case](https://gitlab.freedesktop.org/wayland/wayland/-/blob/6d4497371014cacbf0dc4ed9983fd6617ec25c9a/src/wayland-util.h#L408)

```c
#define wl_container_of(ptr, sample, member)				\
	(__typeof__(sample))((char *)(ptr) -				\
			     offsetof(__typeof__(*sample), member))
```

translate-c could look for this pattern and produce something like this:

```zig
inline fn wl_container_of(ptr: var, sample: var, comptime member: []const u8) *@TypeOf(sample) {
    return @fieldParentPtr(@TypeOf(sample), member, ptr);
}
```

One problem to solve here, however, would be that callsites would have to be modified to convert identifier syntax to string literals.

PR #8092, opened 2021-02-27
translate-c: add limited OffsetOfExpr support

ehaas:
Add support for OffsetOfExpr that contain exactly 1 component, when that component
is a field.

For example, given:

```c
struct S {
  float f;
  double d;
};
struct T {
  long l;
  int i;
  struct S s[10];
};
```

Then:
```c
offsetof(struct T, i)       // supported
offsetof(struct T, s[2].d)  // not supported currently
```


ehaas:
Regarding #4762 - this PR enables the use of `container_of` style macros without special casing. The macro itself won't be translated to a function, but the pieces are now in place for the C code resulting from their use to be translated to Zig (see test case).

PR #8095, opened 2021-02-27
implement fadvise for linux

vrischmann:
This adds `posix_fadvise` for linux and the `POSIX_FADV_*` constants from fcntl.h


vrischmann:
I forgot about other archs, I'll have to look into it

vrischmann:
This syscall is confusing as hell, I looked at musl, glibc and the kernel's code to try to understand what to do but I'm still unsure if this code is correct.

I've tested some sample code with x86_64, i386, arm, aarch64 which appear to work but mips is broken.
As far as I can tell mips needs a syscall with 7 arguments which I have no clue how to implement.

I'll close this for now.

PR #8097, opened 2021-02-28
std: Swap arguments in Thread.spawn

### LINKED ISSUE #8082: Reverse parameter list of `Thread.spawn`
This is pretty minor, but it bothered me enough to want to change it.

# Situation

Currently, `std.Thread.spawn` has the following signature:

```zig
pub fn spawn(context: anytype, comptime startFn: anytype) SpawnError!*Thread
```

Thus, to spawn a thread:

```zig
std.Thread.spawn(ctx, func);
```

# Problems

* A thread is a unit of CPU utilization, which implies the execution of *something* (in this case, `startFn`). The current signature makes that *something*, which is inherent to all threads, seem secondary to contexts, which are only used by cooperative threads.
* Given that threads are so closely tied to execution, seeing and having to provide a context first when spawning them is a bit awkward and unintuitive.
* Even if it wasn't unintuitive, there's no reason to see the context first, as you wouldn't even know what the thread is doing to the data in that context without first knowing what the thread actually does.

# Proposal

Reverse the parameter list:

```zig
pub fn spawn(comptime startFn: anytype, context: anytype) SpawnError!*Thread
```

Thus, to spawn a thread:

```zig
std.Thread.spawn(func, ctx);
```

LemonBoy:
Beside the new order being consistent with the ThreadPool API and making
more sense, this shuffling allows to write the context argument type in
terms of the startFn arguments, reducing the use of anytype (eg. less
explicit casts when using comptime_int parameters, yay).

Sorry for the breakage.

Closes #8082

### DIFF
--- a/lib/std/Thread.zig
+++ b/lib/std/Thread.zig
@@ -165,18 +165,32 @@ pub const SpawnError = error{
     Unexpected,
 };
 
-/// caller must call wait on the returned thread
-/// fn startFn(@TypeOf(context)) T
-/// where T is u8, noreturn, void, or !void
-/// caller must call wait on the returned thread
-pub fn spawn(context: anytype, comptime startFn: anytype) SpawnError!*Thread {
+// Given `T`, the type of the thread startFn, extract the expected type for the
+// context parameter.
+fn SpawnContextType(comptime T: type) type {
+    const TI = @typeInfo(T);
+    if (TI != .Fn)
+        @compileError("expected function type, found " ++ @typeName(T));
+
+    if (TI.Fn.args.len != 1)
+        @compileError("expected function with single argument, found " ++ @typeName(T));
+
+    return TI.Fn.args[0].arg_type orelse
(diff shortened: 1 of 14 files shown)


lib/std/Thread.zig
```diff
@@ -165,18 +165,32 @@ pub const SpawnError = error{
...
-/// caller must call wait on the returned thread
-pub fn spawn(context: anytype, comptime startFn: anytype) SpawnError!*Thread {
+// Given `T`, the type of the thread startFn, extract the expected type for the
+// context parameter.
+fn SpawnContextType(comptime T: type) type {
+    const TI = @typeInfo(T);
+    if (TI != .Fn)
+        @compileError("expected function type, found " ++ @typeName(T));
+
+    if (TI.Fn.args.len != 1)
```
Rocknest, line 175: `args.len == 0` means `void` context, maybe?

Issue #3696, opened 2019-11-15
`a = f(a, b);` return value sometimes interfering with argument?


dbandstra:
This code is giving different results based on implementation of `concatAll`.  The implementations look equivalent.

I have a feeling this has something to do with result locations. Is this how things are supposed to be?

Note: The matrix code here is not important - you can just run the different implementations and see that they output different results. Also, this is probably not the most efficient way to design a matrix library, but it's convenient (when it works).

```zig
const std = @import("std");

pub fn concat(a: [16]f32, b: [16]f32) [16]f32 {
    return .{
        a[0] * b[ 0] + a[4] * b[ 1] + a[ 8] * b[ 2] + a[12] * b[ 3],
        a[1] * b[ 0] + a[5] * b[ 1] + a[ 9] * b[ 2] + a[13] * b[ 3],
        a[2] * b[ 0] + a[6] * b[ 1] + a[10] * b[ 2] + a[14] * b[ 3],
        a[3] * b[ 0] + a[7] * b[ 1] + a[11] * b[ 2] + a[15] * b[ 3],
        a[0] * b[ 4] + a[4] * b[ 5] + a[ 8] * b[ 6] + a[12] * b[ 7],
        a[1] * b[ 4] + a[5] * b[ 5] + a[ 9] * b[ 6] + a[13] * b[ 7],
        a[2] * b[ 4] + a[6] * b[ 5] + a[10] * b[ 6] + a[14] * b[ 7],
        a[3] * b[ 4] + a[7] * b[ 5] + a[11] * b[ 6] + a[15] * b[ 7],
        a[0] * b[ 8] + a[4] * b[ 9] + a[ 8] * b[10] + a[12] * b[11],
        a[1] * b[ 8] + a[5] * b[ 9] + a[ 9] * b[10] + a[13] * b[11],
        a[2] * b[ 8] + a[6] * b[ 9] + a[10] * b[10] + a[14] * b[11],
        a[3] * b[ 8] + a[7] * b[ 9] + a[11] * b[10] + a[15] * b[11],
        a[0] * b[12] + a[4] * b[13] + a[ 8] * b[14] + a[12] * b[15],
        a[1] * b[12] + a[5] * b[13] + a[ 9] * b[14] + a[13] * b[15],
        a[2] * b[12] + a[6] * b[13] + a[10] * b[14] + a[14] * b[15],
        a[3] * b[12] + a[7] * b[13] + a[11] * b[14] + a[15] * b[15],
    };
}

[...]

PR #8098, opened 2021-02-28
std/crypto/aes: fix AES {encrypt,decrypt}Wide

jedisct1:
These functions are  not used by anything yet, but run the last round only once.


daurnimator:
I'm surprised this code doesn't hit #3696

PR #8100, opened 2021-02-28
std/crypto: add AES-OCB

jedisct1:
OCB has been around for a long time.
    
It's simpler, faster and more secure than AES-GCM.
    
RFC 7253 was published in 2014. OCB also won the CAESAR competition along with AEGIS.
    
It's been implemented in OpenSSL and other libraries for years.
    
So, why isn't everybody using it instead of GCM? And why don't we have it in Zig already?
    
The sad reason for this was patents. GCM was invented only to work around these patents, and for all this time, OCB was that nice thing that everybody knew existed but that couldn't be freely used.
    
And that just changed a couple days ago. The OCB patents are now abandoned, and OCB's author announced that [OCB was officially public domain](https://twitter.com/jedisct1/status/1365598429158780928/photo/1).

With AES acceleration, it does well in our benchmarks:

```text
       aes128-gcm:       1302 MiB/s
       aes256-gcm:       1290 MiB/s
       aes128-ocb:       6122 MiB/s
       aes256-ocb:       4428 MiB/s
```

Phil Rogaway hopes "people will use this scheme will be used to do positive things".  So, let's do positive things!

### DIFF
--- a/lib/std/crypto.zig
+++ b/lib/std/crypto.zig
@@ -16,6 +16,11 @@ pub const aead = struct {
         pub const Aes256Gcm = @import("crypto/aes_gcm.zig").Aes256Gcm;
     };
 
+    pub const aes_ocb = struct {
+        pub const Aes128Ocb = @import("crypto/aes_ocb.zig").Aes128Ocb;
+        pub const Aes256Ocb = @import("crypto/aes_ocb.zig").Aes256Ocb;
+    };
+
     pub const Gimli = @import("crypto/gimli.zig").Aead;
 
     pub const chacha_poly = struct {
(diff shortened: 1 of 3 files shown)


jedisct1:
(this PR is on top of #8098 and #8099 - it also removes unnecessary includes for crypto tests)

zigazeljko:
Why do `encrypt` / `decrypt` do their own key expansion? Wouldn't it be better if they took an already-expanded `AesEncryptCtx` as a parameter?

jedisct1:
To expose the same simple interface as all other AEADs.

A unified interface to dissociate key setup (not necessarily AES key expansion) and encryption/decryption can always be added later.

PR #8108, opened 2021-03-01
std: add io.Writer.writeStruct

daurnimator:
We have `readStruct`, add `writeStruct` for symmetry

### DIFF
--- a/lib/std/io/writer.zig
+++ b/lib/std/io/writer.zig
@@ -4,6 +4,7 @@
 // The MIT license requires this copyright notice to be included in all copies
 // and substantial portions of the software.
 const std = @import("../std.zig");
+const assert = std.debug.assert;
 const builtin = std.builtin;
 const mem = std.mem;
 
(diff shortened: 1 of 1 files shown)


lib/std/io/writer.zig
```diff
@@ -86,5 +87,11 @@ pub fn Writer(
             mem.writeInt(T, &bytes, value, endian);
             return self.writeAll(&bytes);
         }
+
+        pub fn writeStruct(self: Self, value: anytype) Error!void {
+            // Only extern and packed structs have defined in-memory layout.
+            comptime assert(@typeInfo(@TypeOf(value)).Struct.layout != builtin.TypeInfo.ContainerLayout.Auto);
+            return self.writeAll(mem.asBytes(&value));
```
MasterQ32, line 94: I think there's a hidden footgun here. `writeStruct` should also assert that no pointer fields are contained in the struct as those won't be valid after potentially reading the struct from disk/network later.

It should also specified which endianess the struct fields should have. Imho this is too easy to use wrong (which is "just use it")

MasterQ32 (changes requested):

lib/std/io/writer.zig
daurnimator: > I think there's a hidden footgun here. `writeStruct` should also assert that no pointer fields are contained in the struct as those won't be valid after potentially reading the struct from disk/network later.

There are valid use-cases where you'd want to write a pointer to a stream.
Furthermore, this is intentionally a mirror of `Reader.readStruct`, and it has no such limitation.

> It should also specified which endianess the struct fields should have. Imho this is too easy to use wrong (which is "just use it")

`readStruct`/`writeStruct` should not care about such things: they aren't `io.Serializer`/`io.Deserializer` (https://github.com/ziglang/std-lib-orphanage/blob/master/std/serialization.zig)

kubkon (approved):

PR #8110, opened 2021-03-01
stage2: remove error number from error set map

g-w1:
This saves memory since it is already stored in module as well as allowing for better threading.
Probably conflicts a little with #7778.
Part 3 of what is outlined in #8079.

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -4083,15 +4083,15 @@ pub fn namedFieldPtr(
             const child_type = try val.toType(scope.arena());
             switch (child_type.zigTypeTag()) {
                 .ErrorSet => {
+                    var name: []const u8 = undefined;
                     // TODO resolve inferred error sets
-                    const entry = if (val.castTag(.error_set)) |payload|
-                        (payload.data.fields.getEntry(field_name) orelse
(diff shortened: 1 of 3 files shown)


src/Module.zig
```diff
@@ -4084,14 +4084,13 @@ pub fn namedFieldPtr(
...
-                            return mod.fail(scope, src, "no error named '{s}' in '{}'", .{ field_name, child_type })).*
+                    if (val.castTag(.error_set)) |payload|
+                        _ = (payload.data.fields.getEntry(field_name) orelse return mod.fail(scope, src, "no error named '{s}' in '{}'", .{ field_name, child_type }))
                     else
-                        try mod.getErrorValue(field_name);
+                        _ = try mod.getErrorValue(field_name);
 
                     const result_type = if (child_type.tag() == .anyerror)
-                        try Type.Tag.error_set_single.create(scope.arena(), entry.key)
+                        try Type.Tag.error_set_single.create(scope.arena(), field_name)
```
Vexu, line 4093: The purpose of using `entry.key` here was that it is owned by the `Module` and has a longer lifespan than `field_name`.

Vexu (changes requested):

PR #8118, opened 2021-03-01
os/linux: fix IO_Uring.timeout

vrischmann:
This PR fixes IO_Uring.timeout on i386

According to the [io_uring PDF](https://kernel.dk/io_uring.pdf) the timeout struct must be 64 bits on both 32 and 64 bit architectures:

> To retain compatibility between 32 vs 64-bit applications and kernel space, the type used must be of the following format
```
struct __kernel_timespec {
  int64_t tv_sec;
  long long tv_nsec;
};
```

I only found this issue because when running `zig build test-std -Dskip-release` on my desktop, [this test](https://github.com/ziglang/zig/blob/master/lib/std/os/linux/io_uring.zig#L1330-L1357) hangs. It can be triggered easily with `zig test -target i386-linux-musl lib/std/os/linux/io_uring.zig` while in the root of the repository.

I found it weird that the CI builds are not affected but grepping through the logs (for example [this one](https://dev.azure.com/ziglang/0184afc2-b8d3-409b-8e4f-4032aaca4c96/_apis/build/builds/12625/logs/27)) all the io_uring tests seem to be skipped for some reason.

### DIFF
--- a/lib/std/os/bits/linux.zig
+++ b/lib/std/os/bits/linux.zig
@@ -2244,3 +2244,8 @@ pub const MADV_COLD = 20;
 pub const MADV_PAGEOUT = 21;
 pub const MADV_HWPOISON = 100;
 pub const MADV_SOFT_OFFLINE = 101;
+
+pub const __kernel_timespec = extern struct {
+    tv_sec: i64,
+    tv_nsec: i64,
+};
--- a/lib/std/os/linux/io_uring.zig
+++ b/lib/std/os/linux/io_uring.zig
@@ -526,7 +526,7 @@ pub const IO_Uring = struct {
     pub fn timeout(
         self: *IO_Uring,
         user_data: u64,
-        ts: *const os.timespec,
+        ts: *const os.__kernel_timespec,
         count: u32,
         flags: u32,
     ) !*io_uring_sqe {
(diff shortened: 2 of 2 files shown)


jorangreef:
Good catch @vrischmann!

> I found it weird that the CI builds are not affected but grepping through the logs (for example this one) all the io_uring tests seem to be skipped for some reason.

The CI builds run on various kernel versions that don't always support all the io_uring ops being tested. io_uring ops have various levels of support across older kernel versions.

vrischmann:
Yeah, now that you mention it it's right there in the test but somehow I missed it before. No weirdness with the CI then.

lib/std/os/bits/linux.zig
```diff
@@ -2244,3 +2244,8 @@ pub const MADV_COLD = 20;
 pub const MADV_PAGEOUT = 21;
 pub const MADV_HWPOISON = 100;
 pub const MADV_SOFT_OFFLINE = 101;
+
+pub const __kernel_timespec = extern struct {
```
daurnimator, line 2248: we usually don't include the `__` prefix when specifying these in zig.

daurnimator (commented):
We likely have this same mistake elsewhere that timespecs are used

lib/std/os/bits/linux.zig
vrischmann: Yeah I wasn't sure about the name so I used what's in the pdf, good to know.

PR #8120, opened 2021-03-01
stage2 ARM: implement basic integer multiplication

joachimschmidt557:
This adds a new tag `mul` to `ir.Instruction`. Currently, this only works with unsigned 32-bit integers (`u32`).

### DIFF
--- a/test/stage2/arm.zig
+++ b/test/stage2/arm.zig
@@ -344,4 +344,38 @@ pub fn addCases(ctx: *TestContext) !void {
             "",
         );
     }
+
+    {
+        var case = ctx.exe("integer multiplication", linux_arm);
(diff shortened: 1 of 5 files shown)


test/stage2/arm.zig
```diff
@@ -344,4 +344,38 @@ pub fn addCases(ctx: *TestContext) !void {
             "",
         );
     }
+
+    {
+        var case = ctx.exe("integer multiplication", linux_arm);
+        // Simple while loop with assert
```
FireFox317, line 350: This is not a simple while loop anymore :D

joachimschmidt557: Oops, will fix. Copy-paste errors, argh.

Issue #4479, opened 2020-02-16
translate-c generates Zig code that does not compile.


JesseRMeyer:
Consider:
```c
typedef long HRESULT;
#define _HRESULT_TYPEDEF_(_sc) ((HRESULT)_sc)
#define DXGI_ERROR_NOT_FOUND _HRESULT_TYPEDEF_(0x887A0002L)
```
translate-c:
```zig
pub inline fn _HRESULT_TYPEDEF_(_sc: var) @TypeOf(if (@typeId(@TypeOf(_sc)) == .Pointer) @ptrCast(HRESULT, _sc) else if (@typeId(@TypeOf(_sc)) == .Int) @intToPtr(HRESULT, _sc) else @as(HRESULT, _sc)) {
    return if (@typeId(@TypeOf(_sc)) == .Pointer) @ptrCast(HRESULT, _sc) else if (@typeId(@TypeOf(_sc)) == .Int) @intToPtr(HRESULT, _sc) else @as(HRESULT, _sc);
}
const DXGI_ERROR_NOT_FOUND = _HRESULT_TYPEDEF_(@as(c_long, 0x887A0002));
```
However:
```
error: integer value 2289696770 cannot be coerced to type 'c_long'
pub const DXGI_ERROR_NOT_FOUND = _HRESULT_TYPEDEF_(@as(c_long, 0x887A0002));
```

Somewhat confusingly, in VS, the hex value mysteriously gets converted to UL:
![ULorL](https://user-images.githubusercontent.com/5846309/74611419-a8de2200-50c9-11ea-84c6-a9624d2e1d53.PNG)

I don't know if that is a function of the UI, and what the compiler ends up seeing.

What should translate-c do with C code that will not compile as Zig code?

andrewrk:
Macros are translated on a heuristic basis. Not all macros can be translated.

This one looks like it can be, but it needs a more complicated Zig expression to express what the C code is doing.

JustinRyanH:
@andrewrk  is this a scenario where we need to identify when for `signed integers` and it's one bigger than `maxInt` in the pre_processor we translate it to `minInt`

JesseRMeyer:
> Macros are translated on a heuristic basis. Not all macros can be translated.

I'm not actually expecting for every macro can be translatable.  I'm expecting that every translated macro be valid Zig.

PR #8126, opened 2021-03-02
translate-c: promote int literals to bigger types

xackus:
My attempt at solving #4479 (which should probably be renamed).
@Vexu What do you think about this?

TODO:
- [x] Previously integer literals without a suffix weren't given a type using `@as`. Giving it is probably more correct, but noisy. Decide what to do.
- [x] Fix translate-c tests.
- [x] Add translate-c/run-translated-c tests.

### DIFF
--- a/lib/std/meta.zig
+++ b/lib/std/meta.zig
@@ -1096,7 +1096,7 @@ test "sizeof" {
 
 pub const CIntLiteralRadix = enum { decimal, octal, hexadecimal };
 
-fn PromoteIntLiteralReturnType(comptime SuffixType: type, comptime target: comptime_int, comptime radix: CIntLiteralRadix) type {
+fn PromoteIntLiteralReturnType(comptime SuffixType: type, comptime number: comptime_int, comptime radix: CIntLiteralRadix) type {
     const signed_decimal = [_]type{ c_int, c_long, c_longlong };
     const signed_oct_hex = [_]type{ c_int, c_uint, c_long, c_ulong, c_longlong, c_ulonglong };
     const unsigned = [_]type{ c_uint, c_ulong, c_ulonglong };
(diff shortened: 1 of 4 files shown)


lib/std/meta.zig
```diff
@@ -1094,6 +1094,47 @@ test "sizeof" {
...
+            return list[pos];
+        }
+    }
+    @compileError("Integer literal does not fit in compatible types");
+}
+
+/// Promote the type of an integer literal until it fits as C would.
+/// This is for translate-c and is not intended for general use.
+pub fn promoteIntLiteral(comptime SuffixType: type, comptime target: comptime_int, comptime radix: CIntLiteralRadix) PromoteIntLiteralReturnType(SuffixType, target, radix) {
+    return @as(PromoteIntLiteralReturnType(SuffixType, target, radix), target);
```
Vexu, line 1124:
```suggestion
    return target;
```
This cast is redundant.

lib/std/meta.zig
```diff
@@ -1094,6 +1094,47 @@ test "sizeof" {
...
+        &signed_oct_hex;
+
+    var pos = mem.indexOfScalar(type, list, SuffixType).?;
+
+    while (pos < list.len) : (pos += 1) {
+        if (target >= math.minInt(list[pos]) and target <= math.maxInt(list[pos])) {
+            return list[pos];
+        }
+    }
+    @compileError("Integer literal does not fit in compatible types");
```
Vexu, line 1118: According to the C spec we can put these in an "extended integer type" so `{iu}128` or maybe even `comptime_int`?

Vexu (approved):
The extra noise is unfortunate but correct behavior is more important.

lib/std/meta.zig
xackus: `__int128_t` is not an "extended integer type" because `intmax_t` is still 64 bit.
https://stackoverflow.com/questions/57332208/example-of-an-extended-integer-type

However, a quick test revealed that gcc issues a warning, but still upgrades to __int128_t:
```c
#include <stdint.h>
#include <stdio.h>

int main(){
    printf("%ld\n", sizeof(typeof(21342323423423234234234)));
    printf("sizeof(__intmax_t): %ld\n", sizeof(__intmax_t));
}
```
```
$ gcc a.c
a.c: In function ‘main’:
a.c:5:35: warning: integer constant is too large for its type
    5 |     printf("%ld\n", sizeof(typeof(21342323423423234234234)));
      |                                   ^~~~~~~~~~~~~~~~~~~~~~~
$ ./a.out 
16
sizeof(__intmax_t): 8

```
clang and MSVC produce an error.

xackus:
@Vexu By the way, great job with the AST changes in translate-c!

PR #8127, opened 2021-03-02
Fix for regression in zig fmt's handling of 'for-if'

### LINKED ISSUE #8088: Zig fmt introduces newline on for-around-if where it formerly didn't
Some recent zig fmt change (introduced in the published zig master after Thu, 25 Feb 2021 03:36:23 GMT before Fri, 26 Feb 2021 03:36:51 GMT) changes
```zig
test "" {
    for (undefined) |x| if (false) {
        //...
    };
}
```
to
```zig
test "" {
    for (undefined) |x|
        if (false) {
            //...
        };
}
```

Was this an intentional change?  Or is this a regression?

I very much liked the original format, which looks quite readable in several of my use cases:
```zig
           for (expr1) |cvToken1| if (cvToken1.cv == .V) {
               for (expr2) |cvToken2| if (cvToken2.cv == .V) {
                   // note: don't try to check for duplicates, that is probably not worth it
                   try self.dvPairs.append(.{ .var1 = cvToken1.token, .var2 = cvToken2.token });
               };
           };
```

Adding the extra newlines _and_ indentation really makes this code less readable:
```zig
           for (expr1) |cvToken1|
               if (cvToken1.cv == .V) {
                   for (expr2) |cvToken2|
                       if (cvToken2.cv == .V) {
                           // note: don't try to check for duplicates, that is probably not worth it
                           try self.dvPairs.append(.{ .var1 = cvToken1.token, .var2 = cvToken2.token });
                       };
               };
```

LewisGaul:
Fixes #8088 

- Added testcase to reproduce the issue
- Factored out common code into `renderWhilePayload()`
- Determine whether newline is needed before the 'if'
- Fix indentation of the 'then' expression when no newline is inserted before the 'if'

### DIFF
--- a/lib/std/zig/render.zig
+++ b/lib/std/zig/render.zig
@@ -1018,170 +1018,95 @@ fn renderWhile(gpa: *Allocator, ais: *Ais, tree: ast.Tree, while_node: ast.full.
         try renderToken(ais, tree, inline_token, .space); // inline
     }
 
-    try renderToken(ais, tree, while_node.ast.while_token, .space); // if
+    try renderToken(ais, tree, while_node.ast.while_token, .space); // if/for/while
     try renderToken(ais, tree, while_node.ast.while_token + 1, .none); // lparen
     try renderExpression(gpa, ais, tree, while_node.ast.cond_expr, .none); // condition
 
-    const then_tag = node_tags[while_node.ast.then_expr];
-    if (nodeIsBlock(then_tag) and !nodeIsIf(then_tag)) {
-        if (while_node.payload_token) |payload_token| {
-            const brace_space = if (while_node.ast.cont_expr == 0 and ais.isLineOverIndented())
-                Space.newline
-            else
-                Space.space;
-            try renderWhilePayload(gpa, ais, tree, payload_token, brace_space);
-        } else {
-            const rparen = tree.lastToken(while_node.ast.cond_expr) + 1;
-            const brace_space = if (while_node.ast.cont_expr == 0 and ais.isLineOverIndented())
-                Space.newline
-            else
-                Space.space;
-            try renderToken(ais, tree, rparen, brace_space); // rparen
-        }
-        if (while_node.ast.cont_expr != 0) {
-            const rparen = tree.lastToken(while_node.ast.cont_expr) + 1;
-            const lparen = tree.firstToken(while_node.ast.cont_expr) - 1;
-            try renderToken(ais, tree, lparen - 1, .space); // :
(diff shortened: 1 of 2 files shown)


lib/std/zig/render.zig
```diff
@@ -1081,51 +1061,37 @@ fn renderWhile(gpa: *Allocator, ais: *Ais, tree: ast.Tree, while_node: ast.full.
...
     const rparen = tree.lastToken(while_node.ast.cond_expr) + 1;
+    const first_then_token = tree.firstToken(while_node.ast.then_expr);
     const last_then_token = tree.lastToken(while_node.ast.then_expr);
     const src_has_newline = !tree.tokensOnSameLine(rparen, last_then_token);
 
     if (src_has_newline) {
+        const base_node_is_if = mem.eql(u8, tree.tokenSlice(while_node.ast.while_token), "if");
+        const newline_before_then_token = !tree.tokensOnSameLine(rparen, first_then_token);
+        const space_before_then_token: Space = if (newline_before_then_token or base_node_is_if) .newline else .space;
+        const indent_expression = base_node_is_if or !nodeIsIf(then_tag) or newline_before_then_token;
```
LewisGaul, line 1072: This all feels a bit icky to me, especially comparing against the 'if' string. I couldn't find a way to get the original type of the `while_node`. Of course we could add a 'is_if_node' param, but at that point I question using `renderWhile()` for rending 'if's at all :)

LewisGaul: ifreund has suggested on IRC that the `base_node_is_if` could just be removed, and this sounds sensible to me. Note that this would result in a change in behaviour - the following would no longer be reformatted onto 3 lines:
```zig
if (a) |x| if (x)
    f(x);
```

ifreund (commented):
This is already an improvement over the status quo and fixes the regression. There's still some inconsistency here that I think would be good to iron out if you're game for it:

```zig
// zig fmt accepts this:
if (a)
    if (b)
        f(x);

// but changes this
if (a)
    for (b) |c|
        foo(x);
// to this:
if (a) for (b) |c|
    foo(x);
```
The same thing happens with while in the place of for

LewisGaul:
> This is already an improvement over the status quo and fixes the regression. There's still some inconsistency here that I think would be good to iron out if you're game for it:

Nice catch, I can certainly have a go at addressing that. Thanks again for taking a look.

ifreund (approved):
Thanks for getting the ball rolling on this! I took your work as a starting point and iterated a bit further to end in what I think is a much better place than the status quo :)

PR #8137, opened 2021-03-03
Implement fmtDuration using Formatter

jdknezek:
Per #8007, this removes single-purpose `std.fmt.Duration` and `std.fmt.duration` and creates `std.fmt.fmtDuration` which uses the `std.fmt.Formatter` pattern.

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -1214,7 +1214,7 @@ pub fn formatIntBuf(out_buf: []u8, value: anytype, base: u8, uppercase: bool, op
 ///
 /// - #ns
 /// - [#y][#w][#d][#h][#m]#[.###][u|m]s
-pub fn formatDuration(ns: u64, writer: anytype) !void {
+pub fn formatDuration(ns: u64, comptime fmt: []const u8, options: std.fmt.FormatOptions, writer: anytype) !void {
     var ns_remaining = ns;
     inline for (.{
         .{ .ns = 365 * std.time.ns_per_day, .sep = 'y' },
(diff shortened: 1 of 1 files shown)


lib/std/fmt.zig
```diff
@@ -1261,7 +1261,12 @@ pub fn formatDuration(ns: u64, writer: anytype) !void {
     return;
 }
 
-test "formatDuration" {
+/// Return a Formatter for a number of nanoseconds using `formatDuration`.
+fn fmtDuration(ns: u64) Formatter(formatDuration) {
```
LemonBoy, line 1265:
```suggestion
pub fn fmtDuration(ns: u64) Formatter(formatDuration) {
```

lib/std/fmt.zig
```diff
@@ -1214,7 +1214,7 @@ pub fn formatIntBuf(out_buf: []u8, value: anytype, base: u8, uppercase: bool, op
 ///
 /// - #ns
 /// - [#y][#w][#d][#h][#m]#[.###][u|m]s
-pub fn formatDuration(ns: u64, writer: anytype) !void {
+pub fn formatDuration(ns: u64, comptime fmt: []const u8, options: std.fmt.FormatOptions, writer: anytype) !void {
```
LemonBoy, line 1217: This can be un-pub'd.

lib/std/fmt.zig
```diff
@@ -1288,24 +1293,13 @@ test "formatDuration" {
...
-    pub fn format(self: Duration, comptime fmt: []const u8, options: FormatOptions, writer: anytype) !void {
-        return formatDuration(self.ns, writer);
-    }
-};
-
-/// Formats a number of nanoseconds according to its magnitude. See `formatDuration`.
-pub fn duration(ns: u64) Duration {
-    return Duration{ .ns = ns };
-}
+const Duration = @compileError("Duration has been deprecated; wrap your argument in std.fmt.fmtDuration instead");
```
LemonBoy, line 1301: Since the `Duration` formatter was introduced within this release you can simply delete those.

LemonBoy (approved):
Great job! I found some minor nits regarding the method visibility.
We could also extract the `Bi` format into its own `Formatter`, that's the last weird case baked into `std.fmt`.

jdknezek:
Thank you for the feedback! Sorry for the sloppy visibility - I only checked the test, not my own code that used it.

PR #8138, opened 2021-03-03
fix zig building after conflict between 8079 and 7778

g-w1:


g-w1:
Did it at the same time lol.

PR #8139, opened 2021-03-04
translate-c: enable pointer arithmetic with signed integer operand

ehaas:
absCast, reorder, and switch operation as necessary to enable signed
integer variables to be used in pointer arithmetic. Previously only
positive integer constants or unsigned integer variables could be used.

Pros: pointer +/- signed integer variable works
Cons: common case of pointer + integer constant is more verbose

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -1127,6 +1127,44 @@ fn transOffsetOfExpr(
     return fail(c, error.UnsupportedTranslation, expr.getBeginLoc(), "TODO: implement complex OffsetOfExpr translation", .{});
 }
 
+/// Translate an arithmetic expression with a pointer operand and a signed-integer operand.
+/// Zig requires a usize argument for pointer arithmetic, so we intCast to isize and then
+/// bitcast to usize; pointer wraparound make the math work.
+/// Zig pointer addition is not commutative (unlike C); the pointer operand needs to be on the left.
+/// The + operator in C is not a sequence point so it should be safe to switch the order if necessary.
+fn transCreatePointerArithmeticSignedOp(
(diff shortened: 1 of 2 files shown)


src/translate_c.zig
```diff
@@ -1127,6 +1127,86 @@ fn transOffsetOfExpr(
...
+/// Translate an arithmetic expression with a pointer operand and a signed-integer operand.
+/// Zig requires pointer arithmetic integer operands to be unsigned so we need to check the value at runtime
+/// and add or subtract the unsigned absolute value.
+/// Zig pointer arithmetic is also not commutative (unlike C); the pointer operand needs to be on the left.
+/// The + and - operators in C are not sequence points so it should be safe to switch the order.
+/// We store the lhs and rhs expression results into temporary variables to prevent double-evaluation
+/// of side-effects.
+///
+/// Given a pointer expression `ptr` and signed int expression `idx` we translate as follows:
+/// `ptr + idx` and `idx + ptr` become:
```
LemonBoy, line 1139: You can also cast the signed int to a `usize` with `@bitCast(usize, @intCast(isize, idx))` to tidy up the generated code.

ehaas: Wouldn't that fail if `idx` is negative? Having that said that I probably should do the `absCast` in the temp variable assignment instead of both arms of the `if`

LemonBoy: Nope, the wrap-around handles that for you.

ehaas: Well that certainly makes the code cleaner. Updated.

Vexu (approved):

PR #8142, opened 2021-03-04
Reject bare +/- input when parsing floats

cryptocode:
This PR proposes that `parseFloat` should fail when inputs are bare sign characters, the same way `parseInt` does. Currently `parseFloat(f64, "+") == 0.0` Note that leading spaces are handled elsewhere in the parser, so I think the s[0] based conditionals are fine (parseInt does the same)


ifreund (approved):
LGTM, Thanks!

PR #8147, opened 2021-03-04
stage2 Elf linker: fill in bswapAllFields

g-w1:
move to std.elf and also add test


lib/std/elf.zig
```diff
@@ -562,6 +545,26 @@ pub fn int32(need_bswap: bool, int_32: anytype, comptime Int64: anytype) Int64 {
     }
 }
 
+pub fn bswapAllFields(comptime S: type, ptr: *S) void {
+    if (comptime std.meta.activeTag(@typeInfo(S)) != .Struct) @compileError("bswapAllFields expects a struct as the first argument");
```
Vexu, line 549: Using `std.meta.activeTag` here is pointless as `@typeInfo(S) != .Struct` just fine without it.

PR #8152, opened 2021-03-05
stage1: Follow the C ABI for return types

LemonBoy:
The current implementation of the target C ABI rules is hopelessly bad,
let's tack some more rules on top in order to prevent some
miscompilations.

Truth to be told the same rule should be applied also to parameters, but
I really can't stand stage1.

### DIFF
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -487,6 +487,53 @@ static LLVMValueRef make_fn_llvm_value(CodeGen *g, ZigFn *fn) {
         addLLVMFnAttr(llvm_fn, "noreturn");
     }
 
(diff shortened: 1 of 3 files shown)


LemonBoy:
Rebased on master so that CI checks can run.

andrewrk:
We do have test coverage for this. Mind adding the newly passing cases? `test/stage1/c_abi/*`

LemonBoy:
> We do have test coverage for this. Mind adding the newly passing cases? test/stage1/c_abi/*

Done, I keep forgetting the `c_abi` folder exists.
A word of warning, the tests do not fail on x86_64. Try targeting for ARM instead, you'll notice how some sign extensions are skipped because the Zig code was expected to sign-extend the results.

PR #8160, opened 2021-03-05
std/special: init-exe,lib make import(std) its own decl

nektro:
developers will often have to make this change when adding complexity to their `build.zig`. this change makes the default to put `std` available at the top level.


MasterQ32:
I would move `std.build.Builder` just into the `fn build()` signature. I never used it outside that function

nektro:
good point

andrewrk:
Thanks!

PR #8166, opened 2021-03-06
Deprecate `B`/`Bi` formatters

LemonBoy:


data-man:
Great, now B/Bi formatting for ints >64 bits is broken.

PR #8167, opened 2021-03-06
fix parsing of assignment with 'inline for' and 'inline while'

vrischmann:
I found this because with Zig `0.8.0-dev.1417+9f722f43a` `zig fmt` is broken with the following code:
```
const std = @import("std");

pub fn main() anyerror!void {
    const data = &[_][]const u8{
        "foo", "bar",
    };

    const foo = inline while (true) {
        break data[0];
    };

    const foo2 = inline for (data) |element| {
        break element;
    };
}
```
I get this error:
```
$ zig fmt src/main.zig
src/main.zig:8:30: error: expected 'while' or 'for', found '('
    const foo = inline while (true) {
                             ~
src/main.zig:12:29: error: expected 'while' or 'for', found '('
    const foo2 = inline for (data) |element| {
```

I've also added two tests for both `inline for` and `inline while`

### DIFF
--- a/lib/std/zig/parse.zig
+++ b/lib/std/zig/parse.zig
@@ -1981,7 +1981,7 @@ const Parser = struct {
                 }
             },
             .keyword_inline => {
-                p.tok_i += 2;
+                p.tok_i += 1;
                 switch (p.token_tags[p.tok_i]) {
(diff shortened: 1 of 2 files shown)


andrewrk (approved):
Thank you!

Issue #793, opened 2018-02-25
Enum Arrays


thejoshwolfe:
This was originally proposed by @raulgrell here: https://github.com/zig-lang/zig/issues/770#issuecomment-368323984

What do you think of an enum array? It has a length equal to the member count of the enum and can only be indexed with an enum value.

```
const Axis = enum { X, Y, Z};
const vec3 = [Axis]f32 {0.0, 0.0, 0.0};
vec3[Axis.X] == 0.0;
```

You can get close to this with status quo Zig by specifying the tag type and casting 

```
const Value = enum(u2) {  Zero, One, Two };
const vals = [3]i32{3, 4, 5};
vals[u2(Value.Zero)] == 3;
```

But then if you change the number of elements, the backing type of the enum or override the values, you need to change a lot of code. And you could still access it with arbitrary integers, so if at any point the index into the array was hardcoded, it would have to be found.

An enum array basically  becomes a comptime-checked  map!

It could be approximately implemented in userland with something like this if we had a memberIndex built-in or something:

```
fn EnumArray(comptime T: type, comptime U: type) -> type {
    return struct {
        data: [@memberCount(T)]U,

        const Self = this;

        fn get(&self: Self, tag: t) U {
            return data[@memberIndex(t)];
        }

        fn set(&self: Self, tag: t, value u) void {
            data[@memberIndex(t)] = u;
        }
    }
}

const value_map = EnumArray(Value, i32) {
    .data = []i32{3, 4, 5}
}

value_map.get(Value.Zero) == 3;
value_map.set(Value.Two, 10);
```

thejoshwolfe:
I use enum arrays frequently in C++. Here are some examples:

https://github.com/thejoshwolfe/legend-of-swarkland/blob/46b1916af06e9d7f6829cb6ff552049e5cd38f28/src/thing.hpp#L196

https://github.com/thejoshwolfe/legend-of-swarkland/blob/46b1916af06e9d7f6829cb6ff552049e5cd38f28/src/serial.cpp#L389

The silly hacks (like `check_indexed_array()`) I do there to make sure I'm doing it right are begging for a language feature. I think this proposal has a lot of value.

raulgrell:
Would we have a `[Enum]null T` too? What would you expect to get in a for? Would we expose the tag? the index? both?

```
const Axis = enum { X, Y, Z};
const vec3 = [Axis]f32 {0.0, 0.0, 0.0};
vec3[Axis.X] == 0.0;

[...]

PR #8171, opened 2021-03-07
Add some enum utilities

SpexGuy:
This PR adds some enum utilities, including EnumSet, EnumMap, and EnumArray data structures, and some helper functions that would have been useful in some recent enum-heavy work I have been doing.  It conflicts with #7649 and #7882, which add similar functionality.  However, this implementation is more complete and handles extern enums with multiple fields mapped to the same value.

I'm open to any suggestions for improvements or name changes.


lib/std/enums.zig
```diff
@@ -0,0 +1,1281 @@
...
+}
+
+/// Determines the length of a direct-mapped enum array, indexed by
+/// @intCast(usize, @enumToInt(enum_value)).  The enum must be exhaustive.
+/// If the enum contains any fields with values that cannot be represented
+/// by usize, a compile error is issued.  The max_unused_slots parameter limits
+/// the total number of items which have no matching enum key (holes in the enum
+/// numbering).  So for example, if an enum has values 1, 2, 5, and 6, max_unused_slots
+/// must be at least 3, to allow unused slots 0, 3, and 4.
+fn directEnumArrayLen(comptime E: type, comptime max_unused_slots: comptime_int) comptime_int {
```
daurnimator, line 125: This should also be offset by a `min`

lib/std/enums.zig
```diff
@@ -0,0 +1,1281 @@
...
+            }
+            max_value = f.value;
+        }
+    }
+
+    const unused_slots = max_value + 1 - fields.len;
+    if (unused_slots > max_unused_slots) {
+        const unused_str = std.fmt.comptimePrint("{d}", .{unused_slots});
+        const allowed_str = std.fmt.comptimePrint("{d}", .{max_unused_slots});
+        @compileError("Cannot create a direct enum array for "++@typeName(E)++". It would have "++unused_str++" unused slots, but only "++allowed_str++" are allowed.");
```
daurnimator, line 150: This should probably return an error rather than a `@compileError`, otherwise a user cannot provide a fallback as an enum grows.

daurnimator (commented):
>  It conflicts with #7649

Note that #7649 adds use-sites at places throughout the standard library. If this is merged instead, please consider adding them too

lib/std/enums.zig
SpexGuy: The intent of this api is to be directly indexed by the integer value of the enum, hence the `direct` part of the name.  The `max_unused_slots` parameter ensures you don't have more padding than you intended.  If you want a better packing, the `EnumArray` type will offset by a min.

lib/std/enums.zig
SpexGuy: I don't understand the use case for catching this error.  This is a safety check to make sure that you are not accidentally making an array with a large number of empty slots.  If this triggers, the way to handle it is to inspect it and decide whether to increase the allowed number of unused slots or switch to a denser packing like EnumArray.  If you are writing generic code that calls this function, your code should accept this parameter alongside the enum.

Can you provide a more specific example where you would want to catch this?

SpexGuy:
Is this CI failure related to my changes?  I don't see an error message so I'm not sure what the problem is.  Maybe the additional comptime execution for the tests causing OOM?

g-w1:
You can close and re-open the pr to re-run the ci.

data-man:
> It conflicts with #7882

Absolutely no.

ifreund:
@SpexGuy should this close #793?

andrewrk:
> Is this CI failure related to my changes? I don't see an error message so I'm not sure what the problem is. Maybe the additional comptime execution for the tests causing OOM?

I think that's it, unfortunately. master branch right now is taking a whopping 8.2 GiB to run std lib tests. This branch bumps it up to 8.4, which is hardly the fault of this code, but it seems to be the straw that broke the camel's back.

I want to merge this though, so it seems I will have to find some other way to recover the wasted memory.

andrewrk:
I'm hoping these are enough:

5cbb642525fa92e16dd3c9e978a9f8a34734de4c
a6f5aa71ac1e8c9a27ecfa6a50f0445a50545a5d
5e5b35f1077d5cf77a89aee5dde939de35f2e247

If not, I'll do more!

PR #8174, opened 2021-03-07
std: Better handling of line-wrapping in Progress

LemonBoy:
In order to update the printed progress string the code tried to move
the cursor N cells to the left, where N is the number of written bytes,
and then clear the remaining part of the line.
This strategy has two main issues:
 - Is only valid if the number of characters is equal to the number of
   written bytes,
 - Is only valid if the line doesn't get too long.

The second point is the main motivation for this change, when the line
becomes too long the terminal wraps it to a new physical line. This
means that moving the cursor to the left won't be enough anymore as once
the left border is reached it cannot move anymore.

The wrapped line is still stored by the terminal as a single line,
despite now taking more than a single one when displayed. If you try to
resize the terminal you'll notice how the contents are reflowed and are
essentially illegible.

Querying the cursor position on non-Windows systems (plot twist,
Microsoft suggests using VT escape sequences on newer systems) is
extremely cumbersome so let's do something different.

Before printing anything let's save the cursor position and clear the
screen below the cursor, this way we ensure there's absolutely no trace
of stale data on screen, and after the message is printed we simply
restore it.

The Windows part is untested, CC @marler8997 and @g-w1

### DIFF
--- a/lib/std/Progress.zig
+++ b/lib/std/Progress.zig
@@ -210,13 +210,15 @@ fn refreshWithHeldLock(self: *Progress) void {
 
         saved_cursor_pos = info.dwCursorPosition;
 
-        const window_height = @intCast(windows.DWORD, info.srWindow.Bottom - info.srWindow.Top) + 1;
-        const window_width = @intCast(windows.DWORD, info.srWindow.Right - info.srWindow.Left) + 1;
+        const window_height = @intCast(windows.DWORD, info.srWindow.Bottom - info.srWindow.Top + 1);
+        const window_width = @intCast(windows.DWORD, info.srWindow.Right - info.srWindow.Left + 1);
         // Number of terminal cells to clear, starting from the cursor position
         // and ending at the window bottom right corner.
-        const fill_chars = window_width * (window_height -
-            @intCast(windows.DWORD, info.dwCursorPosition.Y - info.srWindow.Top)) -
-            @intCast(windows.DWORD, info.dwCursorPosition.X - info.srWindow.Left);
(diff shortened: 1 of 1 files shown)


FireFox317:
Has been tested on `cmd.exe` and `powershell` and works fine! (Except for the fact that windows on Windows can have 0 height and then it breaks, but that is not a blocker for this PR!)

marler8997:
Yeah seems to be working now, tested on my machine.

andrewrk (approved):

andrewrk:
Is there a way to make it so the cursor shows up at the end rather than the beginning? I think that would make it so if stray prints happen, they get appended rather than prepended

image

LemonBoy:
> Is there a way to make it so the cursor shows up at the end rather than the beginning?

More or less, it gets ugly pretty quick as we'd need more logic to save the cursor position across refresh calls on Windows and to avoid restoring it when there's no saved position in the stack.

To be honest this save/restore scheme has [its own problems](https://unix.stackexchange.com/questions/278884/save-cursor-position-and-restore-it-in-terminal), (AFAICS) there's no good and portable way of implementing this kind of stackable progress indicators. I was thinking of stealing the style used by [apt](https://blog.desdelinux.net/wp-content/uploads/2014/04/apt-progress.png), that'd also solve the problem of intermixing stray prints and progress bars.

andrewrk:
Here's an example where the behavior is kinda regressed because before this would have printed the test that it was about to run before crashing. Now I get garbled output but I don't know which test has crashed.

Yeah yeah I should improve the test harness

image

PR #8185, opened 2021-03-08
std/crypto/chacha20: add round-reduced versions & cleanup internals

jedisct1:
See https://eprint.iacr.org/2019/1492.pdf for justification.

8 rounds ChaCha20 provides a 2.5x speedup, and is still believed to be safe.
Round-reduced versions are actually deployed (ex: Android filesystem encryption), and thanks to the magic of comptime, it doesn't take much to support them.

This also makes the ChaCha20 code more consistent with the Salsa20 code, removing internal functions that were not part of the public API any more.

No breaking changes; the public API remains backwards compatible.

### DIFF
--- a/lib/std/crypto.zig
+++ b/lib/std/crypto.zig
@@ -24,8 +24,12 @@ pub const aead = struct {
     pub const Gimli = @import("crypto/gimli.zig").Aead;
 
     pub const chacha_poly = struct {
-        pub const ChaCha20Poly1305 = @import("crypto/chacha20.zig").Chacha20Poly1305;
-        pub const XChaCha20Poly1305 = @import("crypto/chacha20.zig").XChacha20Poly1305;
+        pub const ChaCha20Poly1305 = @import("crypto/chacha20.zig").ChaCha20Poly1305;
+        pub const ChaCha12Poly1305 = @import("crypto/chacha20.zig").ChaCha12Poly1305;
+        pub const ChaCha8Poly1305 = @import("crypto/chacha20.zig").ChaCha8Poly1305;
+        pub const XChaCha20Poly1305 = @import("crypto/chacha20.zig").XChaCha20Poly1305;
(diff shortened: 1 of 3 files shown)


lib/std/crypto/chacha20.zig
```diff
@@ -421,7 +662,7 @@ test "crypto.chacha20 test vector 1" {
         0x6a, 0x43, 0xb8, 0xf4, 0x15, 0x18, 0xa1, 0x1c,
         0xc3, 0x87, 0xb6, 0x69, 0xb2, 0xee, 0x65, 0x86,
     };
-    const input = [_]u8{
+    const m = [_]u8{
```
Mouvedia, line 665: nitpick

that kind of minor/preference changes are not really recommended
better to keep the original blame and keep the changes to a minimum (better for the reviewers too)

lib/std/crypto.zig
```diff
@@ -119,8 +123,14 @@ pub const sign = struct {
 pub const stream = struct {
     pub const chacha = struct {
         pub const ChaCha20IETF = @import("crypto/chacha20.zig").ChaCha20IETF;
+        pub const ChaCha12IETF = @import("crypto/chacha20.zig").ChaCha12IETF;
+        pub const ChaCha8IETF = @import("crypto/chacha20.zig").ChaCha8IETF;
```
Mouvedia, line 127: remark / question

Interesting how these are not called `ChaCha20With96BitNonce`, `ChaCha12With96BitNonce` and `ChaCha8With96BitNonce` to match `ChaCha20With64BitNonce`, `ChaCha12With64BitNonce`, `ChaCha8With64BitNonce` respectively.

lib/std/crypto/chacha20.zig
jedisct1: That goes with aligning the ChaCha code with the Salsa code. These functions are very similar, so minimizing the differences between them will eventually reduce the size and improve the clarity of future PRs.

(Salsa is going to have reduced-round versions, too, if only because this is needed for the scrypt function)

PR #8187, opened 2021-03-09
translate-c: demote usage of un-implemented builtins

ehaas:


src/translate_c.zig
```diff
@@ -1538,6 +1539,22 @@ fn transImplicitCastExpr(
     }
 }
 
+fn isBuiltinDefined(name: []const u8) bool {
+    inline for (std.meta.declarations(c_builtins)) |decl| {
```
daurnimator, line 1543: Use `@hasDecl` to avoid the loop.

daurnimator (dismissed):

src/translate_c.zig
ehaas: Doesn't that require the name to be comptime known?

daurnimator: ah right; I didn't comprehend that this needed to work at runtime.

src/translate_c.zig
```diff
@@ -1538,6 +1539,22 @@ fn transImplicitCastExpr(
...
+        if (std.mem.eql(u8, name, decl.name)) return true;
+    }
+    return false;
+}
+
+fn transBuiltinFnExpr(c: *Context, scope: *Scope, expr: *const clang.Expr, used: ResultUsed) TransError!Node {
+    const node = try transExpr(c, scope, expr, used);
+    if (node.castTag(.identifier)) |ident| {
+        const name = ident.data;
+        if (!isBuiltinDefined(name)) return fail(c, error.UnsupportedTranslation, expr.getBeginLoc(), "TODO implement function '{s}' in std.c.builtins", .{name});
```
Vexu, line 1553: Could you also add this check to `parseCPrimaryExprInner` for consistency?

Vexu (approved):

src/translate_c.zig
ehaas: Updated, although I'm not sure if I should be using `mangled_name` or `slice` there.

Vexu: `mangled_name` should be fine.

PR #8192, opened 2021-03-10
update Thread.AtomicConditional for newer mutex

dgbuckley:
Exports the Mutex.Held so Thread.AtomicConditional can use it to
release.


Vexu:
As discussed in #8637, this issue should be fixed by removing `Held` instead of making changes to it.

PR #8193, opened 2021-03-10
elf: expose machine field in ELF header

kivikakk:
I'd like to inspect the `machine` value!

I wondered if, while doing the byteswap, maybe `@enumToInt` would fail if the underlying value wasn't present in the enum (because we've initially interpreted it with the wrong endianness), and maybe I'd have to bitCast, but it works just fine.

(BE path tested on https://github.com/JonathanSalwan/binary-samples/blob/master/elf-Linux-hppa-bash.)

### DIFF
--- a/lib/std/elf.zig
+++ b/lib/std/elf.zig
@@ -1525,6 +1525,8 @@ pub const EM = extern enum(u16) {
 
     /// Linux kernel bpf virtual machine
     _BPF = 247,
+
+    _,
 };
 
 /// Section data should be writable during execution.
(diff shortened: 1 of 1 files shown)


lib/std/elf.zig
```diff
@@ -387,8 +388,14 @@ pub const Header = struct {
             else => return error.InvalidElfClass,
         };
 
+        const machine = if (need_bswap) blk: {
+            const value = @enumToInt(hdr32.e_machine);
```
LemonBoy, line 392: The use of `ptrCast` is dangerous, here you have a `hdr32.e_machine` with an invalid value due to the reversed byte ordering (and `EM` being exhaustive).
The `intToEnum` involuntarily adds a layer of protection, but only in this code path.

I'd suggest to turn `ET` into non-exhaustive enums to prevent any further problem.

kivikakk: Makes sense. (This is what I was alluding to in the PR body.)

Fixed in 24e5959.

PR #8200, opened 2021-03-10
std/crypto: add scrypt kdf

x13a:
This pull adds [scrypt](https://en.wikipedia.org/wiki/Scrypt) key derivation function and basic [phc-string](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md) parser for kdf(s).

With much help of @jedisct1 .


jedisct1:
Errrr... I don't think it is ready yet :)

x13a:
```
error: expected type 'u5', found 'u6'

const n = @as(usize, 1) << param.log_n;
```

`log_n` is `u6` which max is `(1 << 6) - 1 == 63`
usize max is `0xffff_ffff_ffff_ffff`
`0xffff_ffff_ffff_ffff > (1 << 63)`

What's wrong with my math? :(

jedisct1:
`usize` is platform-dependent. Let's keep discussing this in the repo :)

lib/std/crypto/scrypt/mcf_encoding.zig
```diff
@@ -0,0 +1,127 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2015-2021 Zig Contributors
+// This file is part of [zig](https://ziglang.org/), which is MIT licensed.
+// The MIT license requires this copyright notice to be included in all copies
+// and substantial portions of the software.
+
```
daurnimator, line 6: Could you add a doc-comment about what "MCF" is and where it is specified?

lib/std/crypto/phc_encoding.zig
```diff
@@ -0,0 +1,268 @@
...
+const version_prefix = "v=";
+pub const params_delimiter = ",";
+pub const kv_delimiter = "=";
+
+const Error1 = error{
+    ParseError,
+    InvalidAlgorithm,
+};
+
+// TODO base64 doesn't have one error set
```
daurnimator, line 28: I don't understand this `TODO`

lib/std/crypto/phc_encoding.zig
```diff
@@ -0,0 +1,268 @@
...
+            errdefer self.allocator.free(params);
+            var salt: []u8 = undefined;
+            if (self.salt) |v| {
+                salt = try b64encode(self.allocator, v);
+                i += salt.len + fields_delimiter.len;
+            }
+            errdefer self.allocator.free(salt);
+            var key: []u8 = undefined;
+            if (self.key) |v| {
+                key = try b64encode(self.allocator, v);
```
daurnimator, line 119: Where is this freed (on success)?

lib/std/crypto/scrypt/mcf_encoding.zig
```diff
@@ -0,0 +1,127 @@
...
+pub const Error = error{
+    ParseError,
+    InvalidAlgorithm,
+};
+
+pub const McfEncoding = struct {
+    const Self = @This();
+    const map64 = "./0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+
+    fn encodedLen(comptime len: usize) usize {
```
daurnimator, line 24: why is this arg `comptime`?

lib/std/crypto/scrypt/mcf_encoding.zig
```diff
@@ -0,0 +1,127 @@
...
+        return v;
+    }
+
+    fn parseParams(encoded: *const [14]u8) !scrypt.Params {
+        if (!mem.eql(u8, "$7$", encoded[0..3])) {
+            return Error.InvalidAlgorithm;
+        }
+        return scrypt.Params{
+            .log_n = try intDecode(u6, encoded[3..4]),
+            .r = @as(u32, try intDecode(u30, encoded[4..9])),
```
daurnimator, line 64: These `@as(` shouldn't be needed

lib/std/crypto/scrypt/mcf_encoding.zig
```diff
@@ -0,0 +1,127 @@
...
+        }
+        const leftover = src[i * 3 ..];
+        var v: u24 = 0;
+        for (leftover) |x, j| {
+            v |= @as(u24, x) << @intCast(u5, j * 8);
+        }
+        intEncode(dst[i * 4 ..], v);
+    }
+
+    fn intDecode(comptime T: type, src: *const [(meta.bitCount(T) + 5) / 6]u8) !T {
```
daurnimator, line 49: Why does this take a pointer for `src` while `sliceEncode` takes an array as-is?

daurnimator (changes requested):

lib/std/crypto/scrypt/scrypt.zig
```diff
@@ -0,0 +1,244 @@
...
+/// params: Optional Params. Defaults may change in future.
+pub fn kdf(
+    allocator: *mem.Allocator,
+    derived_key: []u8,
+    password: []const u8,
+    salt: []const u8,
+    params: ?Params,
+) !void {
+    if (derived_key.len == 0 or derived_key.len / 32 > 0xffff_ffff) {
+        return error.InvalidDerivedKeyLen;
```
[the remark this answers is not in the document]
Rocknest: @daurnimator  Thats a bad design choice. Why avoid using good error handling system? It should not reduce performance as much as exceptions would and it allows to use these apis on non verified input. I see no reasons to prefer DoS.

lib/std/crypto/benchmark.zig
```diff
@@ -299,6 +299,31 @@ pub fn benchmarkAes8(comptime Aes: anytype, comptime count: comptime_int) !u64 {
     return throughput;
 }
 
+const pwhash_scrypt = [_]Crypto{Crypto{ .ty = crypto.pwhash.scrypt.kdf, .name = "scrypt" }};
+
+pub fn benchmarkPwhashScrypt(comptime kdf: anytype, comptime count: comptime_int) !u64 {
```
jedisct1, line 304: Even though memory/cpu/cache-hard functions are a bit different, benchmarks compare different functions of the same class.
This is also a good opportunity to ensure that they have the same API.

`scrypt` as a password hashing function should be in a set shared with `bcrypt` (as a KDF, there's probably no need to benchmark it against PBKDF2, as benchmarking PBKDF2 is equivalent to testing the underlying hash function).

lib/std/crypto/scrypt/scrypt.zig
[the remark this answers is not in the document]
jedisct1: The output length is not something controlled by an attacker. Or if it is, the DoS vector exists without an `assert()`.

lib/std/crypto/benchmark.zig
x13a: Sensei, do we need bcrypt phc-encoding support?

jedisct1: `bcrypt` has always used the `crypt` format, I'm not aware of any application using the PHC format with it. That being said, from an API perspective it doesn't really matter.

We currently have

``zig
strHash(password: []const u8, rounds_log: u6) Error![hash_length]u8
strVerify(h: [hash_length]u8, password: []const u8) Error!void
```

for bcrypt, and something that does the same thing with a different names and parameters for scrypt.

That can become:

```zig
strHash(h: []u8, password: []const u8, options: Options) Error![]u8
strVerify(h: []const u8, password: []const u8, options: Options) Error!void
```

With `Options` being a struct that sets both the output format (`.phc`, `.crypt`) and the function-specific parameters.

Even though the format can be inferred from the string, having an `Options` parameter in the verification function may not be a bad idea if we want to limit memory usage, the number of threads, etc.

lib/std/crypto/scrypt/scrypt.zig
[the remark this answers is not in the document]
Rocknest: @jedisct1 The problem with this is that it encourages external parameter validation, duplicating code, and possibly introducing bugs, let the api report the error to me and will decide what to do, use `catch unreachable` or log or warn the user. Especially since there is `error.OutputTooLong` in crypto/error.zig!

Especially i dont like `assert(msg.len == out.len)` (ive seen this is crypto/salsa20.zig), this probably a bug that it wont let me overallocate output slice. But anyway i would prefer `error.OutputOverflow` instead of `assert(out.len >= msg.len)`.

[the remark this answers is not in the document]
x13a: I agree with @Rocknest here. Using asserts is for things that are expected. If `dk.len == 0` maybe is expected, than `dk.len / 32 > 0xffff_ffff` is not expected at all. Somebody 100% will "shoot its leg".

lib/std/crypto/benchmark.zig
x13a: I have added options for `strHash`, but I don't know what options we expect to see in `strVerify`. Encoding can be got from str, params also encoded in str.

jedisct1:
This is awesome work!

Just one thing. Returning strings that have to be deallocated is not very common in Zig, especially when the length is a constant for a given set of parameters.

In the current code, the same allocator is also used for two completely different things (allocating gigabytes of aligned memory for the duration of a computation vs returning a 64 byte string).

So, maybe the functions can accept an existing array and return the corresponding slice? For the Unix crypt format, the size is constant. For the PHC format, the upper bound is known and the exact size can be computed from the parameters.

lib/std/crypto/benchmark.zig
jedisct1: `Options` can be empty for scrypt. But the verification function accepting options is going to be useful for other primitives such as Argon2 where we may want to control the number of threads.
Another option that can be useful is a callback function to monitor progress or to avoid blocking the thread during the entire computation. But that can be done later.

x13a:
Let's wait for [base64](https://github.com/ziglang/zig/pull/8305) to be merged.

jedisct1:
Yep, this is going to remove a bit of code duplication.

jedisct1:
Hi @x13a,

CI may be acting up, but otherwise, the code looks really good!

Once CI's happy, do you think this is ready to merge, or are there more changes you would like to do?

x13a:
Hi @jedisct1 ,

Suppose it's ready. Some things I have to mention..

`Scrypt` kdf, as well, as `argon2` requires `allocator`. If we want to return exact error in `pwhash`, we have to add `mem.Allocator.Error` to `crypto.Error`.

```zig
pub const Error = @import("crypto/error.zig").Error || std.mem.Allocator.Error;
```

Or we can transform it to some internal error.

Another thing, besides I have added `pwhash` to `crypto/benchmark`, I couldn't run it. Maybe I missing something, but on compile I get:

```text
error: import of file outside package path: '../std.zig'
```

If I change it to `@import("std")`, I get:

```text
Assertion failed at /Users/user/dev/zig/src/stage1/ir.cpp:22773 in ir_analyze_container_member_access_inner. This is a bug in the Zig compiler.thread 2390872 panic:

<traceback here>
```

It will be nice if you can check if `crypto/benchmark` is ok.

jedisct1:
> Scrypt kdf, as well, as argon2 requires allocator. If we want to return exact error in pwhash, we have to add mem.Allocator.Error to crypto.Error.

No need to add `mem.Allocator.Error` to `crypto.Error`. Relevant functions can return `crypto.Error || mem.Allocator.Error`.

x13a:
Lol, really didn't think that we can merge errors on return :) Will update tomorrow.

x13a:
Is there any way to understand what is the reason for error in `ziglang.zig`?

```text
2021-04-01T18:17:21.0853863Z 1777 passed; 43 skipped.
2021-04-01T18:21:47.5480829Z The following command terminated unexpectedly:
2021-04-01T18:21:47.5624979Z /home/vsts/work/1/s/build/release/bin/zig test /home/vsts/work/1/s/lib/std/std.zig --test-name-prefix std-native-ReleaseSafe-bare-multi  -OReleaseSafe --cache-dir /home/vsts/work/1/s/zig-cache --global-cache-dir /home/vsts/.cache/zig --name test -I /home/vsts/work/1/s/test --override-lib-dir /home/vsts/work/1/s/lib 
2021-04-01T18:21:47.5630825Z error: the following build command failed with exit code 1:
2021-04-01T18:21:47.5634995Z /home/vsts/work/1/s/zig-cache/o/3c4dee4be353affb186fa64b16bc2978/build /home/vsts/work/1/s/build/release/bin/zig /home/vsts/work/1/s /home/vsts/work/1/s/zig-cache /home/vsts/.cache/zig test -Denable-qemu -Denable-wasmtime
2021-04-01T18:21:47.9445360Z ##[error]Bash exited with code '1'.
2021-04-01T18:21:48.0000186Z ##[section]Finishing: Build and test
```

jedisct1:
It may be an error happening at comptime.

Or, in the case of your function, if may be that the instance is running out of memory.

The way I usually deal with these is by having setup Azure Pipelines for a copy of the repository, with the CI scripts reduced to the bare minimum (only running the Linux tasks, running the relevant tests first, not running non-native environments).

You may get more informative messages that way, but it also makes it faster to get feedback after having made a change.

x13a:
Hello @jedisct1 ,

Merging with master helped: "All checks have passed".

jedisct1:
Yay! This is great news!

Everything looks really nice!

Just one thing: with these changes, bcrypt now requires an allocator, even though the function itself has a constant-size state and always returns a constant-size string. It the allocator even used here?

x13a:
This is done for `bcrypt`, `scrypt` and `argon2` to have the same signatures in `strHash` and `strVerify`, as two last kdf's require allocator. `PhcHasher` is generic and requires allocator for `.create()` and `.verify()` methods, because of `scrypt` and `argon2`. In `bcrypt` allocator is not used. It is a mock pointer.

lib/std/crypto/bcrypt.zig
```diff
@@ -213,84 +275,207 @@ fn strHashInternal(password: []const u8, rounds_log: u6, salt: [salt_length]u8)
...
 ///
 /// IMPORTANT: by design, bcrypt silently truncates passwords to 72 bytes.
 /// If this is an issue for your application, hash the password first using a function such as SHA-512,
 /// and then use the resulting hash as the password parameter for bcrypt.
-pub fn strHash(password: []const u8, rounds_log: u6) Error![hash_length]u8 {
-    var salt: [salt_length]u8 = undefined;
-    crypto.random.bytes(&salt);
-    return strHashInternal(password, rounds_log, salt);
+///
+/// You have to free result after use.
```
jedisct1, line 390: If that true?

lib/std/crypto/bcrypt.zig
```diff
@@ -183,28 +200,73 @@ const Codec = struct {
...
+    allocator: *mem.Allocator,
+    derived_key: []u8,
+    password: []const u8,
+    salt: []const u8,
+    params: Params,
+) !void {
+    kdf(derived_key[0..derived_key_length], password, salt[0..salt_length].*, params);
+}
+
+/// Apply BCRYPT to generate a key from a password.
```
jedisct1, line 263: Wait, what? bcrypt is not a KDF.

lib/std/crypto/bcrypt.zig
x13a: fixed docs

lib/std/crypto/bcrypt.zig
x13a: Let's look.. 

[wikipedia bcrypt](https://en.wikipedia.org/wiki/Bcrypt) tells us that: `bcrypt is a password-hashing function designed by..`, if we click on [password-hashing function](https://en.wikipedia.org/wiki/Password-hashing_function) we get redirected to [Key derivation function](https://en.wikipedia.org/wiki/Key_derivation_function). Where we can read this: 

`In cryptography, a key derivation function (KDF) is a cryptographic hash function that derives one or more secret keys from a secret value such as a main key, a password, or a passphrase using a pseudorandom function."`
...
`The first[citation needed] deliberately slow (key stretching) password-based key derivation function was called "crypt" (or "crypt(3)" after its man page), and was invented by Robert Morris in 1978.`
...
`Various algorithms have been designed specifically for this purpose, including bcrypt, scrypt and, more recently, Lyra2 and Argon2 (the latter being the winner of the Password Hashing Competition).`

jedisct1: bcrypt is a little bit different: it was only made to be a password hashing function.

bcrypt as a KDF generally refers to the [bcrypt-pbkdf](https://flak.tedunangst.com/post/bcrypt-pbkdf) function.

jedisct1:
Got it.

Requiring a "mock pointer" feels a little bit odd.

How about including the allocator in the options set of functions requiring one?

x13a:
I though that allocator is always passed as first argument to functions that require allocations. Maybe do not save the same signature between `bcrypt` and (`scrypt`, `argon2`) and remove allocator from `bcrypt`?

jedisct1:
> I though that allocator is always passed as first argument to functions that require allocations

Not necessarily.

For example `std.json.parse()` also has an options parameter (`ParseOptions`), with an `allocator` field. As long as it doesn't have a default value, it will be required by the functions needing it.

lib/std/crypto.zig
```diff
@@ -109,8 +109,16 @@ pub const onetimeauth = struct {
 ///
 /// Password hashing functions must be used whenever sensitive data has to be directly derived from a password.
 pub const pwhash = struct {
+    pub const Encoding = enum {
+        phc,
+        crypt,
+    };
+
     pub const bcrypt = @import("crypto/bcrypt.zig");
     pub const pbkdf2 = @import("crypto/pbkdf2.zig").pbkdf2;
```
Mouvedia, line 118: Should this import be consistent with bcrypt and scrypt ones?

lib/std/crypto/bcrypt.zig
```diff
@@ -213,84 +275,205 @@ fn strHashInternal(password: []const u8, rounds_log: u6, salt: [salt_length]u8)
...
-pub fn strHash(password: []const u8, rounds_log: u6) Error![hash_length]u8 {
-    var salt: [salt_length]u8 = undefined;
-    crypto.random.bytes(&salt);
-    return strHashInternal(password, rounds_log, salt);
+pub fn strHash(
+    allocator: *mem.Allocator,
+    password: []const u8,
+    options: HashOptions,
+    out: []u8,
+) (Error || mem.Allocator.Error)![]u8 {
```
Mouvedia, line 394: You are using `Error || mem.Allocator.Error` 9 times.
Could you extract/name it?

lib/std/crypto.zig
x13a: It doesn't have any public methods like `strHash` or `strVerify` and I don't know if they are needed. For now If we make it the same as `bcrypt` and `scrypt` it will have only one public function named exactly the same as variable and calling it will be like this:

```zig
const std = @import("std");
const pbkdf2 = std.crypto.pwhash.pbkdf2;

const HmacSha256 = crypto.auth.hmac.sha2.HmacSha256;

fn main() !void {
    var dk: [32]u8 = undefined;
    try pbkdf2.pbkdf2(dk, "password", "salt", 1, HmacSha256);
}
```

Of cause we can do the same as with `bcrypt` and `scrypt` and wrap `rounds` and `Prf` params with struct. Then it will have `Params` public struct and `pbkdf2` function. Also we may call all `kdf` or `hash` functions the same across `pwhash`. Maybe @jedisct1 has something to say.

jedisct1: Right. PBKDF2's output gets serialized, but in many different application-specific ways. Probably not something suitable for a standard library. It's also a generic construction and not a function.
So, we should probably just keep it the way it is.

andrewrk:
30+ days old draft; closing

x13a:
@andrewrk Merged with upstream cause I wanted to complete it in few days. Could you reopen or I have to create new?

Issue #4592, opened 2020-03-01
support detecting the native CPU architecture as being different than the current target


andrewrk:
On my PineBook I can observe the following situation:

 * CPU is aarch64 as reported by `uname -m`, and it can run (static) 64-bit arm executables
 * The OS is a 32 bit installation. For example, `file /usr/bin/env` reports:

```
env: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for GNU/Linux 3.2.0, BuildID[sha1]=736db6308048687a73c5fad96fb7779a41e3b213, stripped
```

That's a a 32-bit ARM ELF file, with a 32-bit dynamic linker path. There is no 64-bit dynamic linker in `/lib`, only the 32-bit one.

`zig builtin` reports `aarch64` for the CPU architecture since that matches the zig binary downloaded from ziglang.org. It was looking for an aarch64 dynamic linker path when looking at `/usr/bin/env` and so it did not find one, and reports the default C ABI of musl, which is incorrect for this Debian installation, and the default 64-bit dynamic linker path for aarch64, which is incorrect. It also fails to detect the system glibc version, for reasons related to how the `/usr/bin/env` ELF file is parsed (runpaths are stored a different way).

This issue is to make native target detection even more powerful, in that it can choose the native CPU architecture. In this case it would be detecting 32-bit arm even though the zig compiler itself is an aarch64 binary.

I believe a similar situation is possible with x86_64 and i386. (64-bit CPU; 32-bit OS installation)

It may make sense to do this after #4591 is complete.

mikdusan:
here's a command that might make a good discriminator:
```sh
/usr/bin/getconf LONG_BIT
```

on my x86_64 mac:
- VM debian (32-bit install) reports 32
- VM debian (64-bit install) reports 64
- VM archlinux reports 64
- macos reports 64

---

- it seems we can map native detection fairly simply:

| os | current native detect | map ? LONG_BIT != native.bit |
|---|---|---|
| debian-i386 | i386-linux-gnu |
| debian-i386 | x86_64-linux-gnu | i386-linux-gnu
| debian-x86_64 | x86_64-linux-gnu
| debian-arm | arm-linux-gnu
| debian-arm | aarch64-linux-gnu | arm-linux-gnu
| debian-aarch64 | aarch64-linux-gnu

andrewrk:
[...]

PR #8206, opened 2021-03-11
std/lib/system: Implement native CPU feature detection for ARM

joachimschmidt557:
Currently, this mirrors LLVM's implementation of CPU feature detection:

https://github.com/llvm/llvm-project/blob/2aef202981211a6744273e4bd56a18e309dd2d79/llvm/lib/Support/Host.cpp#L1591

In future, more information (e.g. from `HWCAP2`) can be extracted and evaluated.

### DIFF
--- a/lib/std/os/bits/linux/arm-eabi.zig
+++ b/lib/std/os/bits/linux/arm-eabi.zig
@@ -518,6 +518,12 @@ pub const HWCAP_IDIV = HWCAP_IDIVA | HWCAP_IDIVT;
 pub const HWCAP_LPAE = 1 << 20;
 pub const HWCAP_EVTSTRM = 1 << 21;
 
+pub const HWCAP2_AES = 1 << 0;
+pub const HWCAP2_PMULL = 1 << 1;
+pub const HWCAP2_SHA1 = 1 << 2;
+pub const HWCAP2_SHA2 = 1 << 3;
+pub const HWCAP2_CRC32 = 1 << 4;
+
 pub const Flock = extern struct {
     l_type: i16,
     l_whence: i16,
--- a/lib/std/zig/system.zig
+++ b/lib/std/zig/system.zig
@@ -889,6 +889,9 @@ pub const NativeTargetInfo = struct {
             .x86_64, .i386 => {
                 return @import("system/x86.zig").detectNativeCpuAndFeatures(cpu_arch, os, cross_target);
             },
+            .arm, .armeb => {
+                return @import("system/arm.zig").detectNativeCpuAndFeatures(cpu_arch, os, cross_target);
+            },
             else => {
                 // This architecture does not have CPU model & feature detection yet.
                 // See https://github.com/ziglang/zig/issues/4591
--- a/lib/std/zig/system/arm.zig
+++ b/lib/std/zig/system/arm.zig
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2015-2021 Zig Contributors
+// This file is part of [zig](https://ziglang.org/), which is MIT licensed.
+// The MIT license requires this copyright notice to be included in all copies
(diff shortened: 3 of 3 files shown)


LemonBoy:
This is not gonna cut it. ARM is pretty weird and the hwcaps alone are not enough to detect the system capabilities, eg. In a big.LITTLE system you'd be only detecting the features of the core you're running on.
And enabling some features on top of a "generic" mcpu is not going to do much, you want to parse the cpuinfo and extract the vendor and part number and match those against a set of known values for some known cores or, as a fallback mechanism, something to at least detect the ARM revision.

I do have some code to parse the cpuinfo I meant to use for Aarch64 feature detection, if you're interested we can extend that to ARM too.

joachimschmidt557:
@LemonBoy I agree, detecting the CPU model/implementer/etc. is probably the most important part and cannot be done with HWCAP afaik (see https://github.com/ziglang/zig/pull/8206/files#diff-f4caceaa8019692e899f9c29b695b0bc5b859727e0f3b30c5759e0b6ffbd5a08R26).

In the last stage2 meeting, we decided to not use `/proc/cpuinfo` and instead have the compiler emit a small binary which will report CPU features and use that information. However, we can revert that again if this solution turns out to not be a good idea. 

[This blog post on ARM Community](https://community.arm.com/developer/tools-software/oss-platforms/b/android-blog/posts/runtime-detection-of-cpu-features-on-an-armv8-a-cpu) also discourages `/proc/cpuinfo` as it reports the CPU features detected by the kernel. On some systems, it is possible to run a AArch32 Linux kernel on an AArch64 CPU. In those situations, the `native` target will be `AArch64`. I haven't tested this out yet, but will we still be able to get the CPU features of the native target if the kernel is running in 32-bit mode?

LemonBoy:
> On some systems, it is possible to run a AArch32 Linux kernel on an AArch64 CPU. In those situations, the native target will be AArch64

The native target is not AArch64, that wouldn't make any sense as the produced binaries could not be executed by the system.
When running in 32bit mode [the kernel](https://elixir.bootlin.com/linux/v5.11.5/source/arch/arm64/kernel/cpuinfo.c#L152) reports the CPUs as ARMv8 cores, handled by the same code path that handles the 32bit case.

> This blog post on ARM Community also discourages /proc/cpuinfo as it reports the CPU features detected by the kernel.

The post suggests not to parse cpuinfo because the feature names may vary between AArch64 and ARMv8 modes, that's something we can easily handle as, at the end of the day, we have to turn whatever the kernel says into a LLVM triple and feature set.

joachimschmidt557:
@LemonBoy I'm sorry, I think I didn't describe the situation well. https://github.com/ziglang/zig/issues/4592 is what I was referring to. Quoting Andrew: "CPU is aarch64 (...) OS is a 32 bit installation". Andrew suggests that the native target in that situation is AArch64.

As I mentioned, I don't know what `/proc/cpuinfo` will give us in that situation. If it will give us the correct and native (AArch64) CPU features, I don't see any problem taking this PR back and going the route of parsing `/proc/cpuinfo` instead.

joachimschmidt557:
Ah yes, I realized my mistake. I thought the kernel would be running in 32 bit mode, but it will run in 64 bit mode while the userspace OS will be 32 bit. In that situation, `/proc/cpuinfo` will give us the correct info, right?

LemonBoy:
The case of 64bit kernel with 32bit userspace is tricky to get right, if the zig compiler is a 64bit executable it will see "AArch64" and not "ARMv8". 
Detecting the preferred userspace bitness is somewhat orthogonal to the detection of CPU model and features, that part should be already handled by `detectAbiAndDynamicLinker` looking at `/usr/bin/env`.

joachimschmidt557:
@LemonBoy during todays stage2 meeting, we decided to go the route of parsing `/proc/cpuinfo`.

PR #8217, opened 2021-03-12
langref: Use "single pointer" and "many pointer"

ratfactor:
"Single" and "many" are short, descriptive names for the two pointer types.

The names come from src/type.zig.

Ziglearn.org and Ziglings have both converted to "many pointers" in an attempt to be consistent and avoid confusion:
* https://github.com/Sobeston/ziglearn/commit/66326929a413960d339fac10c0a95cfbaa76c2d2
* https://github.com/ratfactor/ziglings/commit/0168afd25005b4a3555d2212d2a099cbad344e2b

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -684,7 +684,7 @@ pub fn main() void {
       {#header_close#}
       {#header_open|String Literals and Unicode Code Point Literals#}
       <p>
-      String literals are single-item constant {#link|Pointers#} to null-terminated byte arrays.
+      String literals are constant single {#link|Pointers#} to null-terminated byte arrays.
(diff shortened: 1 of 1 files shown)


jayschwa:
I think this will cause more confusion than clarity. A new person probably isn't going to recognize "many pointer(s)" as a technical term, and instead use the plain English interpretation.

andrewrk:
May I counter propose: "single-item" and "many-item" ?

ratfactor:
@jayschwa I see what you mean and I think you're right. 

@andrewrk That sounds good to me. Descriptive, easy to search for, still pretty short.

I've changed them to "single-item" and "many-item".

PR #8219, opened 2021-03-12
format zig files and add formatting check to ci

g-w1:
It wouldn't let me re-open the previous pr #8026, so I made a new one.
It won't pass since one of the tests is in question whether it is valid zig: see https://freenode.irclog.whitequark.org/zig/2021-03-12#1615520860-1615521429;
The deleted files are already tested in the tokenizer, so they were redundant.

cc @LemonBoy @ifreund


ci/azure/linux_script
```diff
@@ -69,6 +69,8 @@ release/bin/zig build test -Denable-qemu -Denable-wasmtime
 
 # Look for HTML errors.
 tidy -qe ../zig-cache/langref.html
+# look for formatting errorrs
+release/bin/zig fmt --check ../lib/std ../src ../tools ../test ../build.zig
```
ifreund, line 73: Wouldn't it make sense to do this before running the tests so people don't have to wait ~3 hours for the test to runs before seeing that their code isn't properly formatted?

g-w1: I thought the opposite - you want formatting errors to be the last thing to run since it is easily fixable, but the tests are not. Now that I think about it, it would make more sense to have it before.

ifreund: I see your point, I think ideally the fmt check would happen in a different job and allow the actual testing to proceed in parallel even if errors are found.

g-w1: I only enabled it in azure, so the other tests should run. I don't think a fully other job just for fmt is good because it will also have to bootstrap zig which takes cpu time.

ifreund:
The CI is failing due to the self hosted parser rejecting this code:
```zig
fn type_expr_return1() if (true) A {}
fn type_expr_return2() for (true) |_| A {}
fn type_expr_return3() while (true) A {}
```
I'll look into whether this is valid or not according to the zig grammar, it seems that the stage1 parser at least accepts it.

g-w1:
@ifreund see this: https://freenode.irclog.whitequark.org/zig/2021-03-12#1615520860-1615521429

ifreund:
Reminder to self or whoever merges this: this should be rebased on to master and have zig fmt run on the entire codebase again right before merge to make sure no zig fmt changes that have occurred since this branch was created cause the CI to start failing.

Vexu:
This is going to be part of the ci pipeline but it is blocking on fixing the grammar and and questions about possibly also running `zig ast-check`. As the actual change is quite small I'm closing this until we can resolve those issues and also reduce the amount of open prs this'll definetly cause conflicts with.

PR #8227, opened 2021-03-12
translate-c: Implement generic selection expressions

ehaas:
Enables translation of C code that uses the `_Generic` keyword

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -1058,6 +1058,10 @@ fn transStmt(
             const compound_literal = @ptrCast(*const clang.CompoundLiteralExpr, stmt);
             return transExpr(c, scope, compound_literal.getInitializer(), result_used);
         },
+        .GenericSelectionExprClass => {
+            const gen_sel = @ptrCast(*const clang.GenericSelectionExpr, stmt);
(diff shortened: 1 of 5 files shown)


src/translate_c.zig
```diff
@@ -1058,6 +1058,13 @@ fn transStmt(
             const compound_literal = @ptrCast(*const clang.CompoundLiteralExpr, stmt);
             return transExpr(c, scope, compound_literal.getInitializer(), result_used);
         },
+        .GenericSelectionExprClass => {
+            const gen_sel = @ptrCast(*const clang.GenericSelectionExpr, stmt);
+            if (gen_sel.isResultDependent()) {
+                return fail(c, error.UnsupportedTranslation, stmt.getBeginLoc(), "TODO implement translation of result-dependent GenericSelectionExpr", .{});
```
Vexu, line 1064: Based on clang [docs](https://clang.llvm.org/doxygen/classclang_1_1GenericSelectionExpr.html), result-dependent generic selections seem like a C++ feature so we can drop the `TODO` here.

Vexu (approved):

src/translate_c.zig
ehaas: Nice, I wasn't sure about that. I removed all the references to `isResultDependent`

PR #8237, opened 2021-03-13
stage2: Warn when using --debug-log without logging enabled

Ersikan:
A warning is emitted when using the debug option --debug-log when the compiler
was not compiled using the build option -Dlog.

Additionally, the scopes are not added to log_scopes as they have no effect.

I added this warning because I was working on another issue and I could not find why there was no debug logging even though I used the option --debug-log. Maybe this should be included in the [Building Zig from source](https://github.com/ziglang/zig/wiki/Building-Zig-From-Source) page.


kubkon (approved):
LGTM!

ifreund (approved):
Thanks!

PR #8241, opened 2021-03-13
Use a unified error set for std/crypto/*

jedisct1:
This ensures that errors are used consistently across all operations.

### DIFF
--- a/lib/std/crypto/25519/curve25519.zig
+++ b/lib/std/crypto/25519/curve25519.zig
@@ -4,6 +4,7 @@
 // The MIT license requires this copyright notice to be included in all copies
 // and substantial portions of the software.
 const std = @import("std");
+const Error = std.crypto.Error;
 
 /// Group operations over Curve25519.
 pub const Curve25519 = struct {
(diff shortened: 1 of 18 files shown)


lib/std/crypto/25519/curve25519.zig
```diff
@@ -28,14 +29,14 @@ pub const Curve25519 = struct {
...
+    pub fn rejectNonCanonical(s: [32]u8) Error!void {
         return Fe.rejectNonCanonical(s, false);
     }
 
     /// Reject the neutral element.
-    pub fn rejectIdentity(p: Curve25519) !void {
+    pub fn rejectIdentity(p: Curve25519) Error!void {
         if (p.x.isZero()) {
-            return error.IdentityElement;
+            return Error.IdentityElement;
```
kubkon, line 39: I don't think this change is required since `error.IdentityElement == Error.IdentityElement` by definition of the [error set type](https://ziglang.org/documentation/0.7.1/#toc-Error-Set-Type), i.e., you can declare the same error value in multiple error sets but the same values will always get assigned the same underlying `u16` values.

lib/std/crypto/25519/curve25519.zig
```diff
@@ -28,14 +29,14 @@ pub const Curve25519 = struct {
...
 
     /// Check that the encoding of a Curve25519 point is canonical.
-    pub fn rejectNonCanonical(s: [32]u8) !void {
+    pub fn rejectNonCanonical(s: [32]u8) Error!void {
         return Fe.rejectNonCanonical(s, false);
     }
 
     /// Reject the neutral element.
-    pub fn rejectIdentity(p: Curve25519) !void {
+    pub fn rejectIdentity(p: Curve25519) Error!void {
```
kubkon, line 37: This should be enough to coerce `error.IdentityElement` to `Error` subset of error values.

lib/std/crypto/25519/curve25519.zig
jedisct1: ZLS can complete `Error`, not `error` :)

But sure, lemme change all the return statements back to `error`.

lib/std/crypto/25519/curve25519.zig
jedisct1: Changed back to `error`. Thank you!

kubkon (approved):

Issue #8234, opened 2021-03-13
zig test should offer more flexible inputs


finjo-the-great:
It would really be helpful if I could run a command like below to run all tests within the src directory:

    zig test src

I'm sure I can come up with the correct grep/find command to do something equivalent to this, but it would really be a lot easier if I could run a command like above.

LemonBoy:
This problem is already solved by using a nameless `test` block

```
test {
    _ = @import("src/test_file_1.zig");
    // ...
    _ = @import("src/test_file_N.zig");
}
```

kenaryn:
This hint shall be added to the Frequently Asked Questions in a "Testing" sub-section before being closed.

finjo-the-great:
I guess #8221 is worth noting here as well if it's a documentation issue.

Though I do still feel it would be nice to have a way to do something like this without having a list of files to maintain. Maybe some kind of comptime function could work? Then I could do something like:

```
test {
  testSuite("src");
}
```

This would import every source file within that directory. Maybe some sort of glob syntax so you could do something like:

```
test {
  testSuite("src/prefix_*.zig");
}
```

g-w1:
you can do `std.meta.refAllDecls(@This());` to reference all constants (basically all imported files).

finjo-the-great:
> you can do `std.meta.refAllDecls(@This());` to reference all constants (basically all imported files).

I see that there's that method in `std.meta`, but there appears to also be one in `std.testing`. Is the latter what should be used?

Mouvedia:
> This hint shall be added to the Frequently Asked Questions in a "Testing" sub-section or even demonstrated as a example at ziglearn.org before being closed.

https://github.com/ziglang/zig/wiki/Zig-Newcomer-Programming-FAQs

PR #8244, opened 2021-03-13
Add doc comment for std.meta.refAllDecls

finjo-the-great:
This is so that it shows up in the generated documentation, as it came
up in #8234.


g-w1:
This is not what makes something show up in the generated documentation, it has to be referenced, but please don't use the generated docs, it is very experimental and incomplete and needs a rewrite. This feature depends on stage2 and will be worked on once stage2 is in a good place to do so.

finjo-the-great:
Ok. Then we can abandon this change if you feel it's unnecessary.

PR #8247, opened 2021-03-14
zig fmt: fix non-UTF-8 encoding #2820

### LINKED ISSUE #2820: zig fmt: fix non-UTF-8 encoding
Detect Unicode Byte-Order-Mark and strip it. If the file is UTF-16, convert to UTF-8.

Ersikan:
Fixes #2820 

After reading the source code, the first two bytes are inspected, and if
they correspond to a UTF-16 BOM in little-endian order, the source code
is converted to UTF-8.

I noticed that a part of the code to format from stdin and from a path is duplicated.
Would it be desirable to create a function `fmtReader` which would read the source code from a Reader, and output the formatted one to a Writer ?

Another possibility to avoid duplicating the UTF-16 to UTF-8 conversion code and avoid allocating a copy of the data, would be to create a Reader which automatically converts UTF-16 to UTF-8 when needed. Would this be useful in the standard library ?

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -2637,6 +2637,50 @@ fn argvCmd(allocator: *Allocator, argv: []const []const u8) ![]u8 {
     return cmd.toOwnedSlice();
 }
 
+fn readSourceFileToEndAlloc(allocator: *mem.Allocator, input: *const fs.File, size_hint: ?usize) ![]const u8 {
+    const source_code = input.readToEndAllocOptions(
+        allocator,
+        max_src_size,
+        size_hint,
+        @alignOf(u16),
+        null,
+    ) catch |err| switch (err) {
+        error.ConnectionResetByPeer => unreachable,
+        error.ConnectionTimedOut => unreachable,
+        error.NotOpenForReading => unreachable,
+        else => |e| return e,
+    };
+    errdefer allocator.free(source_code);
+
+    // Detect unsupported file types with their Byte Order Mark
+    const unsupported_boms = [_][]const u8{
+        "\xff\xfe\x00\x00", // UTF-32 little endian
+        "\xfe\xff\x00\x00", // UTF-32 big endian
+        "\xfe\xff", // UTF-16 big endian
+    };
+    for (unsupported_boms) |bom| {
+        if (mem.startsWith(u8, source_code, bom)) {
+            return error.UnsupportedEncoding;
+        }
+    }
+
+    // If the file starts with a UTF-16 little endian BOM, translate it to UTF-8
+    if (mem.startsWith(u8, source_code, "\xff\xfe")) {
+        const source_code_utf16_le = mem.bytesAsSlice(u16, source_code);
+        const source_code_utf8 = std.unicode.utf16leToUtf8Alloc(allocator, source_code_utf16_le) catch |err| switch (err) {
+            error.DanglingSurrogateHalf => error.UnsupportedEncoding,
+            error.ExpectedSecondSurrogateHalf => error.UnsupportedEncoding,
+            error.UnexpectedSecondSurrogateHalf => error.UnsupportedEncoding,
+            else => |e| return e,
+        };
+
+        allocator.free(source_code);
+        return source_code_utf8;
+    }
+
+    return source_code;
+}
+
 pub const usage_fmt =
     \\Usage: zig fmt [file]...
     \\
@@ -2709,20 +2753,8 @@ pub fn cmdFmt(gpa: *Allocator, args: []const []const u8) !void {
         }
 
         const stdin = io.getStdIn();
-
-        const source_code = blk: {
-            const source_code = try stdin.readToEndAllocOptions(gpa, max_src_size, null, @alignOf(u16), null);
-            errdefer gpa.free(source_code);
-
-            // If the file starts with a UTF-16 BOM, translate it to UTF-8
-            if (mem.startsWith(u8, source_code, "\xff\xfe")) {
-                const source_code_utf16_le = mem.bytesAsSlice(u16, source_code);
-                const source_code_utf8 = try std.unicode.utf16leToUtf8Alloc(gpa, source_code_utf16_le);
-                gpa.free(source_code);
-                break :blk source_code_utf8;
-            } else {
-                break :blk source_code;
-            }
+        const source_code = readSourceFileToEndAlloc(gpa, &stdin, null) catch |err| {
+            fatal("unable to read stdin: {s}", .{err});
         };
         defer gpa.free(source_code);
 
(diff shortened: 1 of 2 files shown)


src/main.zig
```diff
@@ -2704,9 +2704,22 @@ pub fn cmdFmt(gpa: *Allocator, args: []const []const u8) !void {
...
-
-        const source_code = try stdin.readAllAlloc(gpa, max_src_size);
+        const stdin = io.getStdIn();
+
+        const source_code = blk: {
+            const source_code = try stdin.readToEndAllocOptions(gpa, max_src_size, null, @alignOf(u16), null);
+            errdefer gpa.free(source_code);
+
+            // If the file starts with a UTF-16 BOM, translate it to UTF-8
+            if (mem.eql(u8, source_code[0..2], "\xff\xfe")) {
```
LemonBoy, line 2714:
```suggestion
            if (mem.startsWith(u8, source_code, "\xff\xfe")) {
```

This handles the case where `source_code.len < 2` without erroring.

src/main.zig
```diff
@@ -2704,9 +2704,22 @@ pub fn cmdFmt(gpa: *Allocator, args: []const []const u8) !void {
...
+                const source_code_utf16_le = mem.bytesAsSlice(u16, source_code);
+                const source_code_utf8 = try std.unicode.utf16leToUtf8Alloc(gpa, source_code_utf16_le);
+                gpa.free(source_code);
+                break :blk source_code_utf8;
+            } else {
```
LemonBoy, line 2719: Nit: Should we detect the other interesting BOMs (UTF16BE and UTF32LE/BE) and print an error?

src/main.zig
```diff
@@ -2846,20 +2860,38 @@ fn fmtPathFile(
...
+            error.ConnectionTimedOut => unreachable,
+            error.NotOpenForReading => unreachable,
+            else => |e| return e,
+        };
+        source_file.close();
+        file_closed = true;
+        errdefer fmt.gpa.free(source_code);
+
+        // If the file starts with a UTF-16 BOM, translate it to UTF-8
+        if (mem.eql(u8, source_code[0..2], "\xff\xfe")) {
```
LemonBoy, line 2881: The conversion code should be factored out into its own function, there's too much copy&paste. Something like:

```zig
fn convertTextToUtf8(allocator: *mem.allocator, input: *fs.File) []const u8
```

test/cli.zig
```diff
@@ -174,4 +174,13 @@ fn testZigFmt(zig_exe: []const u8, dir_path: []const u8) !void {
     const run_result3 = try exec(dir_path, true, &[_][]const u8{ zig_exe, "fmt", dir_path });
     // both files have been formatted, nothing should change now
     testing.expect(run_result3.stdout.len == 0);
+
+    // Check UTF-16 decoding
+    const fmt4_zig_path = try fs.path.join(a, &[_][]const u8{ dir_path, "fmt4.zig" });
+    var unformatted_code_utf16 = "\xff\xfe \x00 \x00 \x00 \x00/\x00/\x00 \x00n\x00o\x00 \x00r\x00e\x00a\x00s\x00o\x00n\x00";
+    try fs.cwd().writeFile(fmt4_zig_path, unformatted_code_utf16);
```
LemonBoy, line 181: Thinking out loud, it seems that no test is cleaning up the temporary files it creates.

Ersikan: The argument `dir_path` is a temporary directory which is removed between tests. However, there is no removal at the end.

src/main.zig
Ersikan: Where should this function go ? Still somewhere in `main.zig` ?

LemonBoy: Yep

test/cli.zig
LemonBoy: I'd add a `defer fs.cwd().deleteTree(dir_path) catch {}` below `dir_path` declaration and call it a day then.

Ersikan: I just added this in 0becc17e6b18d9d755729ba0f471266d4b1ceab1.

src/main.zig
Ersikan: This function was added, and I called it `readSourceFileToEndAlloc()`. It seemed more appropriate as it replaced `readToEndAllocOptions()`. Maybe the fact that it converts should be in the name ?

Can you review the implementation ?

src/main.zig
Ersikan: This was added in 72079e1553bb4a40c927aeec9ce7725e4dc68dbe.

However, as said in the commit message, the BOM of UTF-32 is identical to the BOM of UTF-16 followed by a null character. Therefore, one check is redundant, and an invalid source file in UTF-16 containing a null character at the start will be mistaken for an UTF-32 file, which also results in an error.

src/main.zig
```diff
@@ -2846,21 +2892,13 @@ fn fmtPathFile(
...
-        max_src_size,
-        std.math.cast(usize, stat.size) catch return error.FileTooBig,
-        @alignOf(u8),
-        null,
-    ) catch |err| switch (err) {
-        error.ConnectionResetByPeer => unreachable,
-        error.ConnectionTimedOut => unreachable,
-        error.NotOpenForReading => unreachable,
-        else => |e| return e,
+    const source_code = readSourceFileToEndAlloc(fmt.gpa, &source_file, std.math.cast(usize, stat.size) catch return error.FileTooBig) catch |err| {
```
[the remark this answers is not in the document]
Ersikan: This is now done

src/main.zig
```diff
@@ -2633,6 +2633,50 @@ fn argvCmd(allocator: *Allocator, argv: []const []const u8) ![]u8 {
...
+        error.NotOpenForReading => unreachable,
+        else => |e| return e,
+    };
+    errdefer allocator.free(source_code);
+
+    // Detect unsupported file types with their Byte Order Mark
+    const unsupported_boms = [_][]const u8{
+        "\xfe\xff", // UTF-16 big endian
+        "\xff\xfe\x00\x00", // UTF-32 little endian
+        "\xfe\xff\x00\x00", // UTF-32 big endian (redundant check)
```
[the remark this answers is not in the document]
Ersikan: This is now done.

src/main.zig
```diff
@@ -2846,21 +2892,17 @@ fn fmtPathFile(
...
         std.math.cast(usize, stat.size) catch return error.FileTooBig,
-        @alignOf(u8),
-        null,
-    ) catch |err| switch (err) {
-        error.ConnectionResetByPeer => unreachable,
-        error.ConnectionTimedOut => unreachable,
-        error.NotOpenForReading => unreachable,
-        else => |e| return e,
+    ) catch |err| {
+        fatal("unable to read file {s}: {s}", .{ file_path, err });
```
Vexu, line 2900: Do not use `fatal` here, we don't want to stop formatting all the files in a directory just because one of them is using an unsupported encoding.

Ersikan: Thank you for the review ! I did not see the error was logged anyway in `fmtPathDir`. The error is now just returned.

Vexu (approved):
I think `UnsupportedEncoding` might be a more descriptive name for the error but `UnsupportedTextFormat` is fine too.

Ersikan:
@Vexu I changed the error name to `UnsupportedEncoding`.

Ersikan:
@Vexu I'm sorry for the review request. I was trying to ask for a review from @LemonBoy .

PR #8250, opened 2021-03-15
Support optional getaddrinfo arguments

aconbere:
Hi!

This is my first PR to zig, I expect it to be incorrect in a number of ways, but I'm hoping ya'll can help me get it right.

The basic idea here is to more closely mirror the getaddrinfo interface in libc. Getaddrinfo allows you to specify node, service, and hints as null pointers and will be smart about how it treats that information. It will fill in hints with some default values, will assume you want the local node, etc. The original signature in zig prevents these uses, this change hopes to allow c zero pointers to be passed.


lib/std/c/test.zig
```diff
@@ -0,0 +1,21 @@
+const std = @import("std");
+const c = std.c;
+const os = std.os;
+const testing = std.testing;
+
+test "getaddrinfo" {
+    const port = 9999;
+    const port_c = try std.fmt.allocPrint(testing.allocator, "{}\x00", .{port});
```
daurnimator, line 8: You might as well hard-code `"9999"`

lib/std/c/test.zig
```diff
@@ -0,0 +1,21 @@
...
+    defer allocator.free(port_c);
+
+    var res: *os.addrinfo = undefined;
+    defer sys.freeaddrinfo(res);
+    // Question!
+    //
+    // Is this the right way to construct a null c pointer?
+    // Or is it just `null`?
```
daurnimator, line 16: just `null`

lib/std/c/test.zig
```diff
@@ -0,0 +1,21 @@
...
+    const zero = @intToPtr(?[*:0]const u8, 0x0);
+    const status = c.getaddrinfo(zero, port_c, zero, &res);
+
+    expect(rc == 0);
```
daurnimator, line 20: you probably want `testing.expectEqual(@intToEnum(EAI, 0), status);` or similar.

daurnimator (changes requested):

lib/std/c/test.zig
aconbere: Hey, would love for you to check to make sure I did this right. I understand the previous implementation to be doing two things, specifically allocating the strong from an int, and appending the null terminator. It's not entirely clear to me if I simply create a string whether or not it will imply null termination or not, I've biased towards not and included the null byte.

g-w1: You don't need the null-terminator, since strings are normally null terminated and the type argument ``noalias service: [*:0]const u8,`` makes sure of that.

lib/std/c.zig
```diff
@@ -295,12 +295,25 @@ pub extern "c" fn kevent(
...
-    noalias hints: *const addrinfo,
+    noalias node: ?[*:0]const u8,
+    noalias service: ?[*:0]const u8,
+    noalias hints: ?*const addrinfo,
     noalias res: **addrinfo,
 ) EAI;
 
+test "getaddrinfo" {
+    if (!std.builtin.link_libc) return error.SkipZigTest;
+    const allocator = std.testing.allocator;
```
daurnimator, line 306: This is unused.

lib/std/c.zig
```diff
@@ -295,12 +295,25 @@ pub extern "c" fn kevent(
...
+    const port = "9999\x00";
```
daurnimator: zig strings are null-terminated already

lib/std/c.zig
```diff
@@ -295,12 +295,25 @@ pub extern "c" fn kevent(
...
+
+    var res: *std.os.addrinfo = undefined;
+    defer freeaddrinfo(res);
```
daurnimator, line 310: Usually you would put the `defer` after the allocating call.

However, in this case, there is no need for a `defer` at all as you don't use the result. just call `freeaddrinfo` after `getaddrinfo`.

daurnimator (changes requested):

aconbere:
Interesting results!

```
/home/vsts/work/1/s/lib/std/c.zig:302:3: error: use of undeclared identifier 'EAI'
) EAI;
  ^
/home/vsts/work/1/s/lib/std/c.zig:309:20: note: referenced here
    const status = getaddrinfo(null, port, null, &res);
                   ^
```

I've been unable to reproduce locally, but it's interesting to see it failed to build. I have to admit that I'm a little bit confused about the results. At first glance I'm having a hard time understanding why anything about EAI is problematic / changed.

lib/std/c.zig
```diff
@@ -295,12 +295,23 @@ pub extern "c" fn kevent(
...
-    noalias node: [*:0]const u8,
-    noalias service: [*:0]const u8,
-    noalias hints: *const addrinfo,
+    noalias node: ?[*:0]const u8,
+    noalias service: ?[*:0]const u8,
+    noalias hints: ?*const addrinfo,
     noalias res: **addrinfo,
 ) EAI;
 
+test "getaddrinfo" {
```
LemonBoy, line 304: No tests in `c.zig` please, this change needs no test at all.

aconbere: Sounds good, in this case the test was most useful to ensure that it had the correct effect (actually compile passing in a null and return the expected results). Is there a suggested way to make these sorts of changes and validate them? I could easily enough have written a sample program and just validated it out of the code base, just curious what the expectation would be.

LemonBoy: The [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/) standard tells this change is correct, trivial changes to the libc declaration parameters require a proof of their correctness rather than a test.

aconbere:
The build failed with an unrelated error

```
The repository 'https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_18.04  Release' no longer has a Release file.
````

It's not clear to me if this is a boop and retry issue or not, in either case I don't seem to have the bits to fix it.

aconbere:
@daurnimator @g-w1 tests all pass now, thanks ya'll :)

ifreund (approved):
Thanks!

PR #8252, opened 2021-03-15
stage2 llvm bindings: use correct type for LLVMBool for ABI compat 

tadeokondrak:
From the llvm-c header:
```c
typedef int LLVMBool;
```

### DIFF
--- a/src/codegen/llvm/bindings.zig
+++ b/src/codegen/llvm/bindings.zig
@@ -1,7 +1,7 @@
 //! We do this instead of @cImport because the self-hosted compiler is easier
 //! to bootstrap if it does not depend on translate-c.
 
-const LLVMBool = bool;
+const Bool = bool;
 pub const AttributeIndex = c_uint;
 
 /// Make sure to use the *InContext functions instead of the global ones.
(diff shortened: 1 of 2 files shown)


FireFox317 (approved):
Ah, great catch!

src/codegen/llvm/bindings.zig
```diff
@@ -1,7 +1,19 @@
 //! We do this instead of @cImport because the self-hosted compiler is easier
 //! to bootstrap if it does not depend on translate-c.
 
-const LLVMBool = bool;
+pub const Bool = enum(c_int) {
+    False,
+    True,
+    _,
```
andrewrk, line 7: hmm should we perhaps go ahead and assume that LLVM will never use a value other than 0 and 1 for LLVMBool? I think it's a safe assumption. Also keep in mind we are version locked with LLVM; if they change it we can change this code at the same time.

```suggestion
```

andrewrk (approved):

src/codegen/llvm/bindings.zig
tadeokondrak: This isn't a guard against new significant values being added to LLVM, but rather a guard against code that converts a number into a fake-bool without "normalizing" it into either 0 or 1.
I think in C/C++ it's important never to write fake-bool checks as `if (b == TRUE)` for this reason. A nonstandard value being used wouldn't matter to C/C++ programs, so if that mistake is in LLVM now or in the future, it wouldn't be spotted.
I don't think invalid values would be safety-checked across an ABI boundary, so it might cause hard-to-debug undefined behaviour if it does happen.

If you're aware of that reasoning and still want to change it, I don't have a problem with that.

andrewrk: I see! If we want to code defensively around this, we should probably change how the type is represented, right? Because the zig code should not do something like `x == .True`, instead it should rely on the `toBool` function. Should we perhaps do this instead?

```zig
/// C code represents "true" as any value other than 0.
pub const Bool = enum(c_int) {
    False,
    _,
};
```

tadeokondrak: I still think the True variant is useful for arguments, like `f(a, b, c, .True)`. I've added a comment warning against the comparison, and also made the `toBool` function simpler.

PR #8253, opened 2021-03-15
Add test for multi-line while cont expr with same-line then expr

LewisGaul:
As discussed with @ifreund in IRC - this was fixed in https://github.com/ziglang/zig/pull/8127 but is not covered by a testcase.


ifreund (approved):
Thanks!

Issue #7363, opened 2020-12-09
zig fmt messes `if` indentation after line break


Vexu:
`zig fmt` will currently turn
```zig
comptime {
    a =
        if (foo)
            bar;
}
```
into
```zig
comptime {
    a =
        if (foo)
        bar;
}
```

Same happens for `for` and `while`.

PR #8254, opened 2021-03-15
zig fmt: Respect line breaks in struct default value decls

### LINKED ISSUE #7618: zig fmt does not allow line break in struct field default value declaration
I'd like 
```zig
const Foo = struct {
    request_fullscreen: wl.Listener(*wlr.XdgToplevel.event.SetFullscreen) = wl.Listener(*wlr.XdgToplevel.event.SetFullscreen).init(handleRequestFullscreen),
};
```
to allow formatting as
```zig
const Foo = struct {
    request_fullscreen: wl.Listener(*wlr.XdgToplevel.event.SetFullscreen) =
        wl.Listener(*wlr.XdgToplevel.event.SetFullscreen).init(handleRequestFullscreen),
};

```
Which would be in line with how binary operators are handled IMO.

LemonBoy:
Bring this in line with how variable declarations are handled.

Closes #7618

CC @ifreund

### DIFF
--- a/lib/std/zig/render.zig
+++ b/lib/std/zig/render.zig
@@ -1165,8 +1165,23 @@ fn renderContainerField(
         try renderToken(ais, tree, eq_token, eq_space); // =
         ais.popIndent();
     }
-    ais.pushIndentOneShot();
-    return renderExpressionComma(gpa, ais, tree, field.ast.value_expr, space); // value
+
+    if (eq_space == .space)
+        return renderExpressionComma(gpa, ais, tree, field.ast.value_expr, space); // value
+
+    const token_tags = tree.tokens.items(.tag);
+    const maybe_comma = tree.lastToken(field.ast.value_expr) + 1;
+
+    if (token_tags[maybe_comma] == .comma) {
+        ais.pushIndent();
+        try renderExpression(gpa, ais, tree, field.ast.value_expr, .none); // value
+        ais.popIndent();
+        try renderToken(ais, tree, maybe_comma, space);
+    } else {
+        ais.pushIndent();
+        try renderExpression(gpa, ais, tree, field.ast.value_expr, space); // value
+        ais.popIndent();
+    }
 }
 
 fn renderBuiltinCall(
--- a/lib/std/zig/render.zig
+++ b/lib/std/zig/render.zig
@@ -1159,7 +1159,13 @@ fn renderContainerField(
         try renderToken(ais, tree, rparen_token, .space); // )
     }
     const eq_token = tree.firstToken(field.ast.value_expr) - 1;
-    try renderToken(ais, tree, eq_token, .space); // =
+    const eq_space: Space = if (tree.tokensOnSameLine(eq_token, eq_token + 1)) .space else .newline;
+    {
+        ais.pushIndent();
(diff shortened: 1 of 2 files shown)


lib/std/zig/render.zig
```diff
@@ -1159,7 +1159,13 @@ fn renderContainerField(
...
     }
     const eq_token = tree.firstToken(field.ast.value_expr) - 1;
-    try renderToken(ais, tree, eq_token, .space); // =
+    const eq_space: Space = if (tree.tokensOnSameLine(eq_token, eq_token + 1)) .space else .newline;
+    {
+        ais.pushIndent();
+        try renderToken(ais, tree, eq_token, eq_space); // =
+        ais.popIndent();
+    }
+    ais.pushIndentOneShot();
```
ifreund, line 1168: A "oneshot" indent won't work properly in the following case:
```zig
const Foo = struct {
    bar: u32 =
        blk: {
            break :blk 42;
        },
};
```
I think the indentation we want here would be:
```zig
ais.pushIndentNextLine();
renderToken(...);
renderExpressionComma(...);
ais.popIndent();
```
Note this is broken in the same way for assignment currently  #7363. It looks pretty ugly to fix though :/

LemonBoy: Pushed a fix for that case. Yeah, it gets ugly pretty quick :\

lib/std/zig/render.zig
```diff
@@ -1159,8 +1159,29 @@ fn renderContainerField(
...
+    if (token_tags[maybe_comma] == .comma) {
+        ais.pushIndent();
+        try renderExpression(gpa, ais, tree, field.ast.value_expr, .none); // value
+        ais.popIndent();
+        try renderToken(ais, tree, maybe_comma, space);
+    } else {
+        ais.pushIndent();
+        try renderExpression(gpa, ais, tree, field.ast.value_expr, space); // value
+        ais.popIndent();
+    }
```
ifreund, line 1184: This is written this way to handle indentation for line comments directly after the comma correctly I take it? It would probably be good to have a test case for this behavior.

LemonBoy: Not quite, `renderExpressionComma` renders the comma after the multiline string one indentation level too far.

ifreund: Ah that makes sense, though it will also fix this case:
```zig
const Foo = struct {
    bar: u32 =
        42,
    // comment
```
which will currently render as the following if I'm accurately running this code in my head:
```zig
const Foo = struct {
    bar: u32 =
        42,
        // comment
```

LemonBoy: Nope, the comment is rendered at the same level as `bar`. That's unexpected :thinking:

ifreund: Oh I think I wasn't very clear, I expect your current code to render the comment on the same level as bar. Without this special handling for the comma that you added in the latest commit though I'd expect the comment to be rendered at the wrong indentation level.

LemonBoy: All clear then, I'll shortly push another commit with the extra test

ifreund:
Thanks!

PR #8255, opened 2021-03-15
Cleaned up operator parsing code

mathis2003:
`var != notValid` is a bit ugly and hard to read. 
Cleaned it up to `is_valid(var)`


ifreund (commented):
I think extracting stuff like this hurts instead of helps readability. I now need to go check the definition of `is_valid_BinOpType()` to see what it is doing instead of just reading the simple `!=` comparison.

mathis2003:
It's a rather opinion based thing so I understand what you mean. However in this very specific case I feel like the function name is more than enough to know it checks if `op` is a valid `BinOpType`

Issue #7855, opened 2021-01-23
zig fmt will remove the comma after a multi-line string inside a switch statement


throwawayaccount12345-1:
`zig fmt` will remove the comma after a multi-line string inside a switch statement. The compiler will fail to parse after you run `zig fmt`.

```zig
pub fn main() void {
    const x: u32 = 0;
    const str = switch (x) {
        1 => "one",
        2 =>
        \\ zig fmt will delete the comma after the multiline string
        \\ and before 3
        ,
        3 => "three",
        else => "else",
    };
}
```

zig version: `0.8.0-dev.1039+bea791b63` (master release from 2021-01-21)

PR #8257, opened 2021-03-15
Add testcase for switch statement with multiline string

LewisGaul:
Fixes https://github.com/ziglang/zig/issues/7855

The linked issue seems to already be fixed, but this PR adds a testcase to prevent regression.

Also fixed zig fmt adding trailing space after before the target expression when it's rendered on a newline (e.g. in this testcase).

### DIFF
--- a/lib/std/zig/parser_test.zig
+++ b/lib/std/zig/parser_test.zig
@@ -3038,6 +3038,23 @@ test "zig fmt: switch" {
         \\}
         \\
     );
+
+    try testTransform(
+        \\test {
+        \\    switch (x) {
+        \\        foo =>
+        \\            "bar",
+        \\    }
+        \\}
+        \\
+    ,
+        \\test {
+        \\    switch (x) {
+        \\        foo => "bar",
+        \\    }
+        \\}
+        \\
+    );
 }
 
 test "zig fmt: switch multiline string" {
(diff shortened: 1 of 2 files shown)


LewisGaul:
@ifreund would you be able to review? :)

ifreund (commented):
I don't think checking if the first target token is on the next line is sufficent. Does the following test case pass?
```zig
try testTransform(
    \\test {
    \\    switch (x) {
    \\        foo =>
    \\            foo(),
    \\    }
    \\}
    \\
,
    \\test {
    \\    switch (x) {
    \\        foo => foo(),
    \\    }
    \\}
    \\
);
```
From reading your code I'd expect this to fail as the switch arm rendered without the space:
```zig
foo =>foo(),
```

LewisGaul:
@ifreund
You're exactly right (added testcase for this). I actually expected the newline to be respected although now I realise I literally wrote a comment that explains why that's not the case!

Do you think the newline should be respected or removed in the example you gave?

ifreund:
> Do you think the newline should be respected or removed in the example you gave?

I think the newline should be removed as per the status quo.

Only multi line string literals should need special handling here, if you want to make a change from the current behavior I think it would be nice if they were indented.

LewisGaul:
Yep cool I'm guessing I should just be able to change the check from `tree.tokensOnSameLine()` to checking whether the target expr token is a multiline string (trying that now). Will leave the status quo behaviour.

ifreund (approved):
LGTM, Thanks!

PR #8258, opened 2021-03-15
disallow specification of both -dynamic and -static simultaneously when calling zig build-exe

elerch:
As it says on the tin. See discussion in https://github.com/ziglang/zig/pull/8248

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -957,8 +957,14 @@ fn buildOutputType(
                     } else if (mem.eql(u8, arg, "-fno-emit-analysis")) {
                         emit_analysis = .no;
                     } else if (mem.eql(u8, arg, "-dynamic")) {
+                        if (link_mode != null and link_mode.? == .Static) {
+                            fatal("Cannot specify both -dynamic and -static", .{});
+                        }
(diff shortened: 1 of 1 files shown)


ifreund (commented):
I don't think this is a desirable change. The current behavior of the flag set later in the command line overriding any previous flags setting the same thing is inline with the behavior of all other flags. For example, all of the `-ffoo`/`-fno-foo` options work this way.

elerch:
I'm not offended either way - it was just the discussion on the other PR so I thought I'd offer the code for consideration.

/cc @data-man @tauoverpi

data-man:
> For example, all of the -ffoo/-fno-foo options work this way.

And it's bad.

`clang <some_file> -dynamic -static`
> clang-11: error: invalid argument '-dynamic' not allowed with '-static'

Vexu:
Hi @elerch, we discussed this in the [self hosted compiler meeting](https://github.com/ziglang/zig/wiki/Self-Hosted-Compiler-Meetings) and decided that the current behavior of them being able to override each other is desirable. However I have opened #9071 proposal to rename the flags to make this behavior more obvious.

Issue #5176, opened 2020-04-26
zig fmt inserts trailing commas instead of formatting on to one line


daurnimator:
e.g. if I have:
```zig
    var stream = multiOutStream(.{
        fbs1.outStream(),
        fbs2.outStream()
    });
```
then I expect running `zig fmt` to move it all onto one line.

Instead, it appends a `,`.

iansimonson:
It seems to not apply to function args, e.g.:
```
    foo(
        .{
            .a = 1,
            .b = 2,
            .c = 3
        },
        2,
        3,
        4
    );
```
becomes
```
    foo(.{
        .a = 1,
        .b = 2,
        .c = 3,
    }, 2, 3, 4);
```

LewisGaul:
I'm taking a look into this - have reproduced the issue with a testcase.

LewisGaul:
So I've found a way to address this (see #8260) by checking for a comment in the array initialisation instead of a newline. However, this breaks this testcase, which I don't really see how can be sensibly distinguished (except perhaps always making it multi-line when single-line would be too long?): https://github.com/ziglang/zig/blob/83a2665772578f6262afc858613023738a7d25ef/lib/std/zig/parser_test.zig#L1804

Seems this change in behaviour was probably caused by https://github.com/ziglang/zig/commit/1e3a200ba604c8d19fcc1119d01c2e80f2a8beca.

Any thoughts @ifreund?

ifreund:
> So I've found a way to address this (see #8260) by checking for a comment in the array initialisation instead of a newline. However, this breaks this testcase, which I don't really see how can be sensibly distinguished (except perhaps always making it multi-line when single-line would be too long?):

`zig fmt` should never be aware of line length. IMO breaking (and removing) that test case is fine, it's part of solving the inconsistency this issue is about.

LewisGaul:
Cool, I'm happy with that :) I can change that testcase to have a trailing comma, as would be recommended for users that want to maintain the behaviour.

PR #8260, opened 2021-03-15
Format array literals onto a single line if no trailing comma

LewisGaul:
Fixes https://github.com/ziglang/zig/issues/5176

Includes change in behaviour (see tweaked testcases): unless an array literal contains a comment or multiline string, there must be a trailing comma for it to be formatted over multiple lines.

cc @ifreund


lib/std/zig/parser_test.zig
```diff
@@ -1801,19 +1801,19 @@ test "zig fmt: array literal with hint" {
...
+        \\    50000, 60,
         \\};
         \\const a = []u8{0,   1, 2, 3, 40,
         \\    4,5,600,7,
         \\           80,
-        \\    9, 10, 11, 0, 13, 14, 15};
+        \\    9, 10, 11, 0, 13, 14, 15,};
         \\const a = [12]u8{
-        \\    31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31 };
+        \\    31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31, };
```
ifreund, line 1816: could we not change this test case? it was added to catch this bug: https://github.com/ziglang/zig/issues/8423

ifreund (commented):
LGTM!

lib/std/zig/parser_test.zig
LewisGaul: I had to add the tailing comma for the same code path to be exercised, but I've copied the same case minus the trailing comma anyway.

ifreund (approved):
Thanks!

Issue #8256, opened 2021-03-15
zero-length sentinel-terminated array missing sentinel when followed by another variable?


ehaas:
Zig version: 0.8.0-dev.1480+0509131e2
Built from source commit 48efa3bcb678ac2f869554b4c5383f865ad73a33
Platform: macOS 11.2 Big Sur

Ran into this while trying to develop a fix for #8215 (translating e.g. `const char x[] = ""`)

```zig
const std = @import("std");
pub export fn main() void {
    var empty: [0:0]u8 = [0:0]u8{};
    std.debug.assert(empty[0] == 0);
    var abc: [3:0]u8 = [3:0]u8{'a', 'b', 'c'};
}
```

```
➜  ./sentinel_bug
thread 18012281 panic: reached unreachable code
/Users/ehaas/source/zig/build/lib/zig/std/debug.zig:223:14: 0x101ed9de8 in std.debug.assert (sentinel_bug)
    if (!ok) unreachable; // assertion failure
             ^
/Users/ehaas/bugs/sentinel_bug.zig:4:21: 0x101eda894 in main (sentinel_bug)
    std.debug.assert(empty[0] == 0);
                    ^
???:?:?: 0x7fff205aa620 in ??? (???)
???:?:?: 0x0 in ??? (???)
[1]    54770 abort      ./sentinel_bug
```

Removing `abc` or even changing it to be length 2 results in the assert not failing, e.g. this code runs fine:

```zig
const std = @import("std");
pub export fn main() void {
    var empty: [0:0]u8 = [0:0]u8{};
    std.debug.assert(empty[0] == 0);
    var abc: [2:0]u8 = [2:0]u8{'a', 'b'};
}
```

mikdusan:
<details>
<summary>ok1.zig</summary>

```zig
const std = @import("std");
pub export fn main() void {
    var empty: [0:0]u8 = [0:0]u8{};
    std.debug.assert(empty[0] == 0);
    var abc: [3:0]u8 = [3:0]u8{'a', 'b', 'c'};
    _ = abc;
}
```
</details>

```sh
$ zig version
0.11.0-dev.1026+4172c2916

$ zig run ok1.zig
# no stdout/stderr (good)
```

PR #8264, opened 2021-03-16
translate-c: Use [N:0] arrays when initializer is a string literal

### LINKED ISSUE #8215: @cImport of string literal does not provide sentinel array
This is an extension of a question I asked on the [mailing list](https://lists.sr.ht/~andrewrk/ziglang/%3CYEpRAzVCqVbmFVRT%40gpanders.com%3E).

When `@cImport`ing header files that define string literals, Zig does not import these as type `[N:0]u8` (where N is the length of the string) but rather as `[N+1]u8` (where the +1 is from the null sentinel value).

Here's an example:

**config.h**:
```c
const char OPTION_FOO[] = "some value";
```

**test.zig**:
```zig
test {
    const config = @cImport(@cInclude("config.h"));
    const foo: []const u8 = config.OPTION_FOO;
}
```

Running

```bash
zig test -lc -I. test.zig
```

results in

```
./test.zig:3:35: error: expected type '[]const u8', found '[11]u8'
    const foo: []const u8 = config.OPTION_FOO;
```

This also precludes you from using e.g. `std.mem.span`:

```zig
test {
    const std = @import("std");
    const config = @cImport(@cInclude("config.h"));
    const foo = std.mem.span(config.OPTION_FOO);
}
```

results in

```
/usr/local/lib/zig/std/mem.zig:546:17: error: invalid type given to std.mem.Span
```

`@cImport` used to import C string literals as type `[*c]const u8`, which *could* be passed to `std.mem.span` to obtain a `[]const u8` string in Zig, but this no longer works. This seems to have been changed in https://github.com/ziglang/zig/pull/7849.

My question is two-fold:

1. Is it intended behavior for C string literals to be imported with type `[N+1]u8` rather than `[*c]const u8` or `[N:0]u8`? To be frank, I think this is quite confusing and a potential footgun, as now the sentinel value has to be manually cleaned up/handled in Zig code instead of `@cImport` simply "doing the right thing".
2. Regardless of the answer to (1), what is the method for converting a C string literal defined as in the above case into a sentinel-less Zig string? As I pointed out above,`std.mem.span` does not work. Providing a pointer to the array also does not work as expected (e.g. `&config.OPTION_FOO` results in `*[11]u8` which *can* be coerced to `[]const u8` with a value of `"some value\0"`)

Thanks!

ehaas:
Translate arrays as [N:0] when initialized by a string literal, when possible.
This preserves a bit more of the type information from the original C program.

Fixes #8215

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -1832,9 +1832,9 @@ fn transStringLiteralInitializer(
     const init_node = if (num_inits > 0) blk: {
         if (is_narrow) {
             // "string literal"[0..num_inits].*
-            const lit_node = try transNarrowStringLiteral(c, scope, stmt, .used);
-            const sliced = try Tag.string_slice.create(c.arena, .{ .string = lit_node, .end = num_inits });
-            break :blk try Tag.deref.create(c.arena, sliced);
+            var str = try transNarrowStringLiteral(c, scope, stmt, .used);
+            if (str_length != array_size) str = try Tag.string_slice.create(c.arena, .{ .string = str, .end = num_inits });
+            break :blk try Tag.deref.create(c.arena, str);
         } else {
             const init_list = try c.arena.alloc(Node, num_inits);
             var i: c_uint = 0;
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -636,7 +636,7 @@ fn visitVarDecl(c: *Context, var_decl: *const clang.VarDecl, mangled_name: ?[]co
     if (has_init) trans_init: {
         if (decl_init) |expr| {
             const node_or_error = if (expr.getStmtClass() == .StringLiteralClass)
-                transStringLiteralAsArray(c, scope, @ptrCast(*const clang.StringLiteral, expr), zigArraySize(c, type_node) catch 0)
+                transStringLiteralInitializer(c, scope, @ptrCast(*const clang.StringLiteral, expr), type_node)
             else
                 transExprCoercing(c, scope, expr, .used);
             init_node = node_or_error catch |err| switch (err) {
@@ -1412,7 +1412,7 @@ fn transDeclStmtOne(
 
             var init_node = if (decl_init) |expr|
                 if (expr.getStmtClass() == .StringLiteralClass)
-                    try transStringLiteralAsArray(c, scope, @ptrCast(*const clang.StringLiteral, expr), try zigArraySize(c, type_node))
+                    try transStringLiteralInitializer(c, scope, @ptrCast(*const clang.StringLiteral, expr), type_node)
                 else
                     try transExprCoercing(c, scope, expr, .used)
             else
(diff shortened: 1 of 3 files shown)


ehaas:
Marking this as draft for 2 reasons:

1. I believe #8256 is blocking it; since there seems to be a bug with [0:0]T arrays. This could also be worked around by translating zero-length arrays as just [0]T
2. For constant-size arrays, I am translating them as [N-1:0] *if* they are initialized with a string literal and the string literal is smaller than the declared array size. e.g. `const char[5] = "foo";` becomes `var foo: [4:0]u8 = [4:0]u8{'f', 'o', 'o', 0,};`. Should we just keep these as [N]T arrays instead?

SpexGuy:
Re: 2, I think it makes more sense to generate a raw array.  If `str: [N:0]u8` is coerced to a slice with `ref: [:0]const u8 = &str`, the resulting slice will have a length of N, which is probably not what was wanted.  Better to require the user to call `spanZ` to coerce to a null terminated slice.  The default values can be generated as `"foo".* ++ [_]u8{0} ** extra`.

ehaas:
`"foo".* ++ [_]u8{0} ** extra` is really slick! That will make the generated code look much nicer. Will look into adding that tomorrow.

src/translate_c/ast.zig
```diff
@@ -176,6 +178,7 @@ pub const Node = extern union {
         c_pointer,
         single_pointer,
         array_type,
+        null_sentinel_array_type,
```
ehaas, line 181: I'm thinking maybe it would be better to add a `null_sentinel` boolean to `Payload.Array` instead of introducing a new ast node type here? It would let me remove some ugly if statements and helper functions from `translate_c.zig`

src/translate_c/ast.zig
```diff
@@ -335,13 +338,15 @@ pub const Node = extern union {
                 .block => Payload.Block,
                 .c_pointer, .single_pointer => Payload.Pointer,
                 .array_type => Payload.Array,
+                .null_sentinel_array_type => Payload.NullSentinelArray,
```
Vexu, line 341: Just use `Payload.Array` for this.

ehaas: Ok - just to be clear - you're thinking I should add a `null_sentinel` boolean to `Payload.Array`, right?

ehaas: Nevermind, I see what you're saying...I can just handle it in `renderNode`

src/translate_c.zig
```diff
@@ -1787,52 +1796,74 @@ fn transStringLiteral(
...
+
+    if (array_size == 0) return Tag.empty_array.create(c.arena, elem_type);
+
+    const num_inits = math.min(str_length, array_size);
+    const init_node = if (num_inits > 0) blk: {
+        if (is_narrow) {
+            // "string literal"[0..num_inits].*
+            const lit_node = try transNarrowStringLiteral(c, scope, stmt, .used);
+            const sliced = try Tag.string_slice.create(c.arena, .{ .string = lit_node, .end = num_inits });
+            break :blk try Tag.deref.create(c.arena, sliced);
```
Vexu, line 1837:
```suggestion
            var str = try transNarrowStringLiteral(c, scope, stmt, .used);
            if (str_length != array_size) str = try Tag.string_slice.create(c.arena, .{ .string = str, .end = num_inits });
            break :blk try Tag.deref.create(c.arena, str);
```
Looks like we should be able to avoid the slice most of the time.

src/translate_c/ast.zig
Vexu: Sorry for the confusing review, I accidentally added this as a single comment instead of starting a review.

ehaas: No worries, I made the requested changes as separate commits. Once you're happy with it I can squash everything into a single commit if you want.

Issue #8149, opened 2021-03-05
publish binary tarballs for macos arm64


mikdusan:
This issue is to track and plan out how we intend to offer Apple silicon (arm64) published binary tarballs in our ci/release process. Comments are welcome.

- today we have:
```
https://ziglang.org/builds/zig-macos-x86_64-0.8.0-dev.1409+ffb2568a9.tar.xz
```

- tomorrow we want:
```
https://ziglang.org/builds/zig-macos-x86_64-0.8.0-dev.1409+ffb2568a9.tar.xz
https://ziglang.org/builds/zig-macos-aarch64-0.8.0-dev.1409+ffb2568a9.tar.xz
```

#### POINTS TO CONSIDER
- Apple is branding their silicon arch as `arm64` on the front of things, but clang/llvm still use `aarch64` under the hood
- some command-line tools on macos and command-line args use `arm64`
- I will assume tarball pathnames will use `aarch64`
- technically straightforward  to produce fat-binaries on macos if needed
- azure does not yet offer any Apple M1-based hardware for pipelines
- azure arm64 artifacts cannot be host/native executed
- do we still make tarball available with "not tested" caveat?
- there _might_ be unexpected differences down the road when running our macos-x86_64 artifacts on BigSur+x86_64 when compared to our macos-arm64 artifacts on BigSur+arm64. We've tackled macos-version detection and made it work with **our** macos-x86_64 artifacts (which are pre-BigSur). If needed, one solution is to make a variant of macos-x86_64 for BigSur+ so naming might be `macos11.1-x86_64` to somewhere encode sdk version used

#### CHOICES MADE IN THIS PLAN
- use `aarch64` for file/tarball naming
- do not produce fat-binaries
- skip `zig build test` until host/native execution is available
- use azure job w/ `vmImage: macos-10.15`, `xcode-12.4`
- do not encode SDK in tarball pathname
- do not spin another macos-x86_64 artifact just for BigSur+
 
#### IMPLEMENTATION ~~PLAN~~
- see #8265

mikdusan:
Apple-silicon master binaries are now being produced. The tarball is not yet integrated into the download page but it available as so:

Manually transform the macOS binary download URL's architecture component from `x86_64` to `aarch64` . eg:
```
https://ziglang.org/builds/zig-macos-x86_64-0.8.0-dev.1548+0d96a284e.tar.xz
```
to
```
[...]

PR #8265, opened 2021-03-16
azure: produce macos arm64 binaries

mikdusan:
new pipeline `BuildMacOS_arm64`
  - `vmImage: 'macOS-10.15' `

new `macos_arm64_script`
  - switch from using `make` to `ninja`
  - select xcode 12.4
  - set zig-cache env variables
  - build host-zig binary with xcode,  link against llvm for x86_64 (target macos 10.15)
  - build arm64-zig binary with xcode and host-zig, link against llvm for arm64 (target macos 11.0)
  - use host-zig for docgen
  - use host-zig for experimental std lib docs
  - sync final `release/` hierarchy with `linux_script`
  - use gnu-tar for good-practices (set owner, set sort)

enhance `CMakeLists.txt`
  - do not build `zig0` when cross-compiling
  - disable `BYPRODUCTS` directive `zig1.o` to avoid `ninja` error

see also #8149

### DIFF
--- a/ci/azure/macos_arm64_script
+++ b/ci/azure/macos_arm64_script
@@ -1,0 +1,132 @@
+#!/bin/sh
+
+set -x
+set -e
+
+brew install s3cmd ninja gnu-tar
+
+ZIGDIR="$(pwd)"
+ARCH="aarch64"
+# {product}-{os}{sdk_version}-{arch}-{llvm_version}-{cmake_build_type}
+CACHE_HOST_BASENAME="llvm-macos10.15-x86_64-11.0.1-release"
+CACHE_ARM64_BASENAME="llvm-macos11.0-arm64-11.0.1-release"
+PREFIX_HOST="$HOME/$CACHE_HOST_BASENAME"
+PREFIX_ARM64="$HOME/$CACHE_ARM64_BASENAME"
+JOBS="-j2"
+
+rm -rf $PREFIX
+cd $HOME
+wget -nv "https://ziglang.org/deps/$CACHE_HOST_BASENAME.tar.xz"
+wget -nv "https://ziglang.org/deps/$CACHE_ARM64_BASENAME.tar.xz"
+
+gtar xf "$CACHE_HOST_BASENAME.tar.xz"
+gtar xf "$CACHE_ARM64_BASENAME.tar.xz"
+
+cd $ZIGDIR
+
+# Make the `zig version` number consistent.
+# This will affect the cmake command below.
+git config core.abbrev 9
+git fetch --unshallow || true
+git fetch --tags
+
+# Select xcode: latest version found on vmImage macOS-10.15 .
+DEVELOPER_DIR=/Applications/Xcode_12.4.app
+
+export ZIG_LOCAL_CACHE_DIR="$ZIGDIR/zig-cache"
+export ZIG_GLOBAL_CACHE_DIR="$ZIGDIR/zig-cache"
+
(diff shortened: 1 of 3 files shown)


ci/azure/macos_arm64_script
```diff
@@ -0,0 +1,137 @@
...
+
+ninja $JOBS install
+
+# Disable test because binary is foreign arch.
+#release/bin/zig build test
+
+# Show file size and dylibs.
+/bin/ls -lh release/bin
+/usr/bin/otool -L release/bin/zig
+/usr/bin/file release/bin/zig
```
kubkon, line 85: I gather this was to verify that `clang` successfully built an arm64 binary, am I correct? If so, is there any point in keeping these around? I don't imagine this ever regressing with each upcoming release of Apple build tools.

kubkon:
This is just me thinking out loud, but do you think there is any chance we can make use of `zld` to produce the cross-compiled stage1? FYI, `zld` is getting upstreamed in #8282.

To elaborate on the topic further, this PR is relying on Apple build tools ability to cross-compile from `x86_64` to `arm64`, right? Or am I missing something?

mikdusan:
> This is just me thinking out loud, but do you think there is any chance we can make use of `zld` to produce the cross-compiled stage1? FYI, `zld` is getting upstreamed in #8282.

It's very possible! More on this below.

> To elaborate on the topic further, this PR is relying on Apple build tools ability to cross-compile from `x86_64` to `arm64`, right? Or am I missing something?

Yup, 2 phases:

first for host binaries:
1. uses xcode-12.4 artifact llvm+clang+lld (deployment target `macos 10.15 x86_64`)
1. uses xcode-12.4 to build hosted zig.exe stage1 (deployment target `macos 10.15 x86_64`)

then cross-target binaries:
1. uses xcode-12.4 artifact llvm+clang+lld (deployment target `macos 11.0 arm64`)
1. uses xcode-12.4 to compile zig stage1 `.cpp` (deployment target `macos 11.0 arm64`)
1. uses hosted zig.exe to compile `zig1.o` (target `aarch64`)
1. uses xcode-12.4 to link zig.exe stage1 (deployment target `macos 11.0 arm64`)

We may be able to replace cross-target steps 2,4 with zig toolchain (`zig c++` and `zld`).

...and a few notes on `zig-bootstrap` failure (the reason I went with full apple toolchain):

I have been unable to get working artifacts from `zig-bootstrap` for zig master or llvm12 branches. I've tried on x86_64 macos 11.x and 10.13 hosts.  It all seems related to zig toolchain in `cross-target` mode. Also linker flags like `-undefined suppress` is not yet supported yet used during llvm build.

That said, I am maintaining xcode artifacts `llvm+clang+ld` (llvm-12.0.0) targeting `macos 10.9 x86_64` and `macos 11.0 arm64` for llvm12 branch in lieu of zig-bootstrap.

kubkon:
Ahhh, our in-house linker cannot yet handle C++, so I reckon this is a great interim solution :+1:

kubkon (approved):
LGTM! @andrewrk would you mind having a quick look as well?

ci/azure/macos_arm64_script
mikdusan: removed and force-pushed.

kubkon:
This looks good to me, so I'm gonna go ahead and merge this so that we have arm64 macOS nightlies available. Thanks for working on this @mikdusan!

PR #8266, opened 2021-03-16
rework ZIR memory layout; overhaul source locations

andrewrk:
This is a follow-up from #7920, applying the same principles and strategies to ZIR.

The memory layout for ZIR instructions is completely reworked. See
zir.zig for those changes. Some new types:

  * `zir.Code`: a "finished" set of ZIR instructions. Instead of allocating
    each instruction independently, there is now a Tag and 8 bytes of
    data available for all ZIR instructions. Small instructions fit
    within these 8 bytes; larger ones use 4 bytes for an index into
    `extra`. There is also `string_bytes` so that we can have 4 byte
    references to strings. `zir.Inst.Tag` describes how to interpret
    those 8 bytes of data.
    - This is shared by all `Block` scopes.

  * `Module.WipZirCode`: represents an in-progress `zir.Code`. In this
    structure, the arrays are mutable, and get resized as we add/delete
    things. There is extra state to keep track of things. This struct is
    stored on the stack. Once it is finished, it produces an immutable
    `zir.Code`, which will remain on the heap for the duration of a
    function's existence.
    - This is shared by all `GenZir` scopes.

  * `Sema`: represents in-progress semantic analysis of a `zir.Code`.
    This data is stored on the stack and is shared among all `Block`
    scopes. It is now the main "self" argument to everything in the file
    that was previously named `zir_sema.zig`.
    Additionally, I moved some logic that was in `Module` into here.

`Module.Fn` now stores its parameter names inside the `zir.Code`,
instead of inside ZIR instructions. When the TZIR memory layout
reworking time comes, codegen will be able to reference this data
directly instead of duplicating it.

astgen.zig is (so far) almost entirely untouched, but nearly all of it
will need to be reworked to adhere to this new memory layout structure.

I have no benchmarks to report yet, as I am still working through
compile errors and fixing various things that I broke in this branch.

### Overhaul of Source Locations:

Previously we used `usize` everywhere to mean byte offset, but sometimes
also mean other stuff. This was error prone and also made us do
unnecessary work, and store unnecessary bytes in memory.

Now there are more types involved into source locations, and more ways
to describe a source location.

  * AllErrors.Message: embrace the assumption that files always have less
    than 2 << 32 bytes.
  * SrcLoc gets more complicated, to model more complicated source
    locations.
  * Introduce LazySrcLoc, which can model interesting source locations
    with very little stored state. Useful for avoiding doing unnecessary
    work when no compile errors occur.

Also, previously, we had `src: usize` on every ZIR instruction. This is
no longer the case. Each instruction now determines whether it even cares
about source location, and if so, how that source location is stored.
This requires more careful work inside `Sema`, but it results in fewer
bytes stored on the heap, without compromising accuracy and power of
compile error messages.

### Miscellaneous:

  * std.zig: string literals have more helpful result values for
    reporting errors. There is now a lower level API and a higher level
    API.
    - side note: I noticed that the string literal logic needs some love.
      There is some unnecessarily hacky code there.
  * cut & pasted some TZIR logic that was in zir.zig to ir.zig. This
    probably broke stuff and needs to get fixed.
  * Removed type/Enum.zig, type/Union.zig, and type/Struct.zig. I don't
    think this quite how this code will be organized. Need some more
    careful planning about how to implement structs, unions, enums. They
    need to be independent Decls, just like a top level function.


andrewrk:
## Benchmark: print1mil.zig

Caveat: this is not a realistic workload. We need more language features supported to benchmark a realistic workload.

### Results Summary

 * :heavy_check_mark: Wall Clock Time: 0.93 seconds => 0.57 seconds (39% reduction)
   * this is 1.8 million lines per second on my laptop :grin: 
 * :heavy_check_mark: Peak Memory Usage: 645 MiB => 386 MiB (40% reduction)
 * :heavy_check_mark: Cache Misses (53% reduction)
 * :heavy_check_mark: Instructions (23% reduction)
 * :heavy_check_mark: CPU Cycles (41% reduction)

### Source:

```zig
pub export fn _start() noreturn {
    print(); // repeated 1,000,000 times
    exit();
}

fn print() void {
    asm volatile ("syscall"
        :
        : [number] "{rax}" (1),
          [arg1] "{rdi}" (1),
          [arg2] "{rsi}" (@ptrToInt("Hello, world!\n")),
          [arg3] "{rdx}" (14)
        : "rcx", "r11", "memory"
    );
    return;
}

fn exit() noreturn {
    asm volatile ("syscall"
        :
        : [number] "{rax}" (231),
          [arg1] "{rdi}" (0)
        : "rcx", "r11", "memory"
    );
    unreachable;
}
```

### Raw Results

Processor: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
NVMe hard drive

#### master branch

```
[nix-shell:~/Downloads/zig/build-release]$ $(which time) -v ./zig-cache/bin/zig build-exe ~/tmp/print1mil.zig 
	Command being timed: "./zig-cache/bin/zig build-exe /home/andy/tmp/print1mil.zig"
	User time (seconds): 0.74
	System time (seconds): 0.18
	Percent of CPU this job got: 100%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.93
	Maximum resident set size (kbytes): 660308
	Minor (reclaiming a frame) page faults: 211118
	Voluntary context switches: 22
	Involuntary context switches: 3
	Swaps: 0
	File system inputs: 0
	File system outputs: 8
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0

[nix-shell:~/Downloads/zig/build-release]$ perf stat -e cache-misses -e instructions -e cycles ./zig-cache/bin/zig build-exe ~/tmp/print1mil.zig

 Performance counter stats for './zig-cache/bin/zig build-exe /home/andy/tmp/print1mil.zig':

        71,742,110      cache-misses                                                
     7,582,166,758      instructions              #    1.81  insn per cycle         
     4,200,228,900      cycles                                                      

       0.936024950 seconds time elapsed

       0.737926000 seconds user
       0.200523000 seconds sys

```

#### zir-memory-layout branch


```
[nix-shell:~/dev/zig/build-release]$ $(which time) -v ./zig-cache/bin/zig build-exe ~/tmp/print1mil.zig
	Command being timed: "./zig-cache/bin/zig build-exe /home/andy/tmp/print1mil.zig"
	User time (seconds): 0.46
	System time (seconds): 0.11
	Percent of CPU this job got: 100%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.57
	Maximum resident set size (kbytes): 395624
	Minor (reclaiming a frame) page faults: 115696
	Voluntary context switches: 21
	Involuntary context switches: 4
	Swaps: 0
	File system inputs: 0
	File system outputs: 8
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0

[nix-shell:~/dev/zig/build-release]$ perf stat -e cache-misses -e instructions -e cycles ./zig-cache/bin/zig build-exe ~/tmp/print1mil.zig

 Performance counter stats for './zig-cache/bin/zig build-exe /home/andy/tmp/print1mil.zig':

        33,900,182      cache-misses                                                
     5,838,500,070      instructions              #    2.34  insn per cycle         
     2,492,983,377      cycles                                                      

       0.572696174 seconds time elapsed

       0.460613000 seconds user
       0.113397000 seconds sys
```

andrewrk:
Follow-up issues extracted:
 * #8402
 * #8403 
 * #8404

PR #8271, opened 2021-03-16
zir-memory-layout: fix @setEvalBranchQuota

g-w1:
I didn't know if I should use `block.sema` or just `sema`, and couldn't figure out a way to test this since the code currently does not compile.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -599,8 +599,8 @@ fn zirSetEvalBranchQuota(
     const src = inst_data.src();
     try sema.requireFunctionBlock(block, src);
     const quota = try sema.resolveAlreadyCoercedInt(block, src, inst_data.operand, u32);
-    if (b.branch_quota.* < quota)
-        b.branch_quota.* = quota;
+    if (sema.branch_quota < quota)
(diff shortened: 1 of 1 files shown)


andrewrk:
Thanks for the patch! To answer the question - those reference the same thing. `sema` is passed as a first arg for convenience and because I wasn't completely sure if Block will keep the sema field. For now, prefer the `sema` if you have access to both.

src/Sema.zig
```diff
@@ -2958,13 +2958,10 @@ fn safetyPanic(sema: *Sema, block: *Scope.Block, src: LazySrcLoc, panic_id: Pani
 }
 
 fn emitBackwardBranch(sema: *Sema, block: *Scope.Block, src: LazySrcLoc) !void {
-    const shared = block.inlining.?.shared;
-    shared.branch_count += 1;
-    if (shared.branch_count > block.branch_quota.*) {
+    sema.branch_count += 1;
```
andrewrk, line 2961: `sema` has `branch_quota` but the previous code here mutating the value on the `Inlining.Shared` is still correct.

PR #8273, opened 2021-03-16
crypto/pbkdf2: simplify the check for the max number of iterations

jedisct1:
The check was difficult to read and possibly incorrect.

Simplify; all we want to check here is that a 32 bit counter is not going to overflow.

### DIFF
--- a/lib/std/crypto/pbkdf2.zig
+++ b/lib/std/crypto/pbkdf2.zig
@@ -79,7 +79,7 @@ pub fn pbkdf2(dk: []u8, password: []const u8, salt: []const u8, rounds: u32, com
     //      block
     //
 
-    const blocks_count = (dk_len + h_len - 1) / h_len;
+    const blocks_count = @intCast(u32, std.math.divCeil(usize, dk_len, h_len) catch unreachable);
     var r = dk_len % h_len;
     if (r == 0) {
         r = h_len;
--- a/lib/std/crypto/pbkdf2.zig
+++ b/lib/std/crypto/pbkdf2.zig
@@ -67,8 +67,8 @@ pub fn pbkdf2(derivedKey: []u8, password: []const u8, salt: []const u8, rounds: 
     //   1. If dkLen > maxInt(u32) * hLen, output "derived key too long" and
     //      stop.
     //
-    if (comptime (maxInt(usize) > maxInt(u32) * hLen) and (dkLen > @as(usize, maxInt(u32) * hLen))) {
-        // If maxInt(usize) is less than `maxInt(u32) * hLen` then dkLen is always inbounds
+    if (dkLen / hLen >= maxInt(u32)) {
+        // Counter starts at 1 and is 32 bit, so if we have to return more blocks, we would overflow
         return error.OutputTooLong;
     }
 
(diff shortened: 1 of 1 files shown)


lib/std/crypto/pbkdf2.zig
```diff
@@ -52,43 +52,38 @@ const Error = std.crypto.Error;
...
 
     // FromSpec:
     //
-    //   1. If dkLen > maxInt(u32) * hLen, output "derived key too long" and
+    //   1. If dk_len > maxInt(u32) * h_len, output "derived key too long" and
     //      stop.
     //
-    if (comptime (maxInt(usize) > maxInt(u32) * hLen) and (dkLen > @as(usize, maxInt(u32) * hLen))) {
-        // If maxInt(usize) is less than `maxInt(u32) * hLen` then dkLen is always inbounds
+    if (dk_len / h_len >= maxInt(u32)) {
```
Rocknest, line 70: I would still keep the divisionless version (also only comptime_ints on the right side)
```suggestion
    if (dk_len >= maxInt(u32) * h_len) {
```

lib/std/crypto/pbkdf2.zig
```diff
@@ -52,43 +52,38 @@ const Error = std.crypto.Error;
...
-    // let `L(dkLen, hLen) = (dkLen + hLen - 1) / hLen`
-    // then `L^-1(l, hLen) = l*hLen - hLen + 1`
-    // 1) L^-1(maxInt(u32), hLen) <= maxInt(u32)*hLen
-    // 2) maxInt(u32)*hLen - hLen + 1 <= maxInt(u32)*hLen // subtract maxInt(u32)*hLen + 1
-    // 3) -hLen <= -1 // multiply by -1
-    // 4) hLen >= 1
-    const r_ = dkLen % hLen;
-    const l = @intCast(u32, (dkLen / hLen) + @as(u1, if (r_ == 0) 0 else 1)); // original: (dkLen + hLen - 1) / hLen
-    const r = if (r_ == 0) hLen else r_;
+    const blocks_count = (dk_len + h_len - 1) / h_len;
```
Rocknest, line 82: This may overflow on 32 bit. 
if `dk_len` is `maxInt(32) - 1` and `h_len` is 3
then `maxInt(32) - 1 + 3 - 1` = `maxInt(32) + 1` overflows

LemonBoy: FYI there's also `std.math.divCeil`

lib/std/crypto/pbkdf2.zig
jedisct1: `h_len` is a power of 2 and comptime, so there won't be any actual division here :)

lib/std/crypto/pbkdf2.zig
jedisct1: `std.math.divCeil()` looks perfect here, thank you @LemonBoy!

Changed to `std.math.divCeil()`.

lib/std/crypto/pbkdf2.zig
Rocknest: @jedisct1 is it always a power of 2?

jedisct1: Except for SHA384 but the compiler is smart enough to convert that into shift+add.

PR #8275, opened 2021-03-16
std: make all errors in os.zig public

### LINKED ISSUE #8243: os.ListenError is private
It's part of a public API, and hence should be public so it can be used in merged error sets returned by functions calling `os.listen`.

g-w1:
Fixes #8243


YerinAlexey (approved):
LGTM, you were faster than me

PR #8279, opened 2021-03-17
translate-c: preserve zero fractional part in float literals

### LINKED ISSUE #8278: translate-c: Weird interactions with #define and floats
This [example raylib code](https://github.com/raysan5/raylib/blob/7e62d973f95383a9e09a3520baad515e2e9b0705/examples/models/models_rlgl_solar_system.c) failed to compile after translating to Zig with `translate-c`.

Tested on `zig-linux-x86_64-0.8.0-dev.1479+aa1c78056`.

Sorry beforehand if this is a duplicate and/or my overall terminology is weird. I don't know much about C so I didn't really know what terms to search.

## Reproducing:

In order to highlight what's going on, I wrote a minimal example to reproduce the problem:
```c
// This code doesn't compile after translating to Zig
#include <stdio.h>

#ifndef PI
    #define PI 3.14159265358979323846f
#endif

#ifndef DEG2RAD
    #define DEG2RAD (PI/180.0f)
#endif

int main(void) {
    printf("DEG2RAD is: %f\n", DEG2RAD);

    return 0;
}
```

## The translated code:

The code above translates without any problems, but attempting to compile the code produces this error:
```
$ zig build-exe broken.zig -lc
./broken.zig:303:52: error: float value 3.141593 cannot be coerced to type 'comptime_int'
    _ = printf("DEG2RAD is: %f\n", @floatCast(f64, 3.1415927410125732 / 180));
                                                   ^
./broken.zig:303:71: note: referenced here
    _ = printf("DEG2RAD is: %f\n", @floatCast(f64, 3.1415927410125732 / 180));
                                                                      ^
```
Here are the relevant portions of the translated Zig code:
```zig
pub const PI = @as(f32, 3.14159265358979323846);
pub const DEG2RAD = PI / @as(f32, 180.0);
```
```zig
pub export fn main() c_int {
    _ = printf("DEG2RAD is: %f\n", @floatCast(f64, 3.1415927410125732 / 180));
    return 0;
}
```

## Cause and evaluation:

There seems to be two problems.
1. Even though the translator translated the contents of `#define` macro, it still is translating every instance of `DEG2RAD` as opposed to just using the translated value.
2. Rather than the expected `@floatCast(f64, 3.1415927410125732 / 180.0)`, it uses `180` as the denominator, which causes the error.

Ideally, I think it should be translated like this:
```zig
pub export fn main() c_int {
    _ = printf("DEG2RAD is: %f\n", @floatCast(f64, DEG2RAD));
    return 0;
}
```

Of course, I know little about C so my proposed solution might be naive.

---

As a side note, the following code works without any problems:
```c
// This code works fine
#include <stdio.h>

const float pi = 3.14159265358979323846f;
const float deg2rad = pi/180.0f;

int main(void) {
    printf("deg2rad is: %f\n", deg2rad);

    return 0;
}
```
It translates into this code instead:
```zig
pub export const pi: f32 = 3.1415927410125732;
pub export const deg2rad: f32 = pi / 180;
pub export fn main() c_int {
    _ = printf("deg2rad is: %f\n", @floatCast(f64, deg2rad));
    return 0;
}
```

Vexu:
Closes #8278.

### DIFF
--- a/test/run_translated_c.zig
+++ b/test/run_translated_c.zig
@@ -3,6 +3,17 @@ const tests = @import("tests.zig");
 const nl = std.cstr.line_sep;
 
 pub fn addCases(cases: *tests.RunTranslatedCContext) void {
+    cases.add("division of floating literals",
+        \\#define _NO_CRT_STDIO_INLINE 1
+        \\#include <stdio.h>
+        \\#define PI 3.14159265358979323846f
+        \\#define DEG2RAD (PI/180.0f)
+        \\int main(void) {
+        \\    printf("DEG2RAD is: %f\n", DEG2RAD);
+        \\    return 0;
+        \\}
+    , "DEG2RAD is: 0.017453" ++ nl);
+
     cases.add("use global scope for record/enum/typedef type transalation if needed",
         \\void bar(void);
         \\void baz(void);
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -3207,11 +3207,11 @@ fn transFloatingLiteral(c: *Context, scope: *Scope, stmt: *const clang.FloatingL
     var dbl = stmt.getValueAsApproximateDouble();
     const is_negative = dbl < 0;
     if (is_negative) dbl = -dbl;
-    const str = try std.fmt.allocPrint(c.arena, "{d}", .{dbl});
-    var node = if (dbl == std.math.floor(dbl))
-        try Tag.integer_literal.create(c.arena, str)
+    const str = if (dbl == std.math.floor(dbl))
+        try std.fmt.allocPrint(c.arena, "{d}.0", .{dbl})
     else
-        try Tag.float_literal.create(c.arena, str);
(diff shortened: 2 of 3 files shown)


test/run_translated_c.zig
```diff
@@ -3,6 +3,17 @@ const tests = @import("tests.zig");
...
+    cases.add("division of floating literals",
+        \\#define _NO_CRT_STDIO_INLINE 1
+        \\#include <stdio.h>
+        \\#define PI 3.14159265358979323846f
+        \\#define DEG2RAD (PI/180.0f)
+        \\int main(void) {
+        \\    printf("DEG2RAD is: %f\n", DEG2RAD);
+        \\    return 0;
+        \\}
+    , "DEG2RAD is: 0.017453\n");
```
LemonBoy, line 15:
```suggestion
    , "DEG2RAD is: 0.017453" ++ nl);
```

Issue #5909, opened 2020-07-22
remove the destination type parameter from all cast builtins


andrewrk:
`@as` is unchanged. These are affected:

 * `@floatCast`
 * `@intCast`
 * `@alignCast`
 * `@bitCast`
 * `@errSetCast`
 * `@ptrCast`
 * `@truncate`

The motivation here is to prevent bugs. Here's an example case:

```zig
const Data = struct {
    bits: u8,
};

fn updateBits(data: *Data) void {
    // ...
    const y: usize = indexOfSomethingGuaranteedToFitInBits();
    data.bits = @intCast(u8, y);
    // ...
}
```

Here, we obtain `y`, a `usize` but for whatever reason is guaranteed to fit into the `bits` field of `Data`. So we use `@intCast` here to cast it.

Now, we rework the code, and actually we need to add some more bits, making it a `u16`:

```zig
const Data = struct {
    bits: u16,
};

fn updateBits(data: *Data) void {
    // ...
    const y: usize = indexOfSomethingGuaranteedToFitInBits();
    data.bits = @intCast(u8, y);
    // ...
}
```

We've now introduced a bug into the code. The `bits` field is updated, but the `@intCast` is still casting too small. The lack of type inference here has actually made the code buggier.

The original code theoretically should have used `data.bits = @intCast(@TypeOf(data.bits), y)` however this is so cumbersome that it's rarely done in practice, especially when `data.bits` is a more complicated expression.

With this proposal, the original code would look like this:

```zig
const Data = struct {
    bits: u8,
};

fn updateBits(data: *Data) void {
    // ...
    const y: usize = indexOfSomethingGuaranteedToFitInBits();
    data.bits = @intCast(y);
    // ...
}
```

And the updated code (which only changes the type of the `bits` field to `u16`) remains correct.

With this proposal, it is still possible to get the older semantics:

```zig
const Data = struct {
    bits: u8,
};

fn updateBits(data: *Data) void {
    // ...
    const y: usize = indexOfSomethingGuaranteedToFitInBits();
    data.bits = @as(u8, @intCast(y));
    // ...
}
```

A cast operation with an inferred result location type would be a compile error:

```zig
test "example" {
    var y: i32 = 1;
    const x = @intCast(y); // error: unable to infer result type \n hint: coerce the result using `@as`
}
```

zigazeljko:
Why are `@bitCast` and `@ptrCast` also affected? The issue with casting too small does not occur there.

[...]

PR #8280, opened 2021-03-17
@intCast takes two args

dimenus:
Extremely minor fix. '@intCast' takes two args instead of one.


andrewrk:
Note the accepted proposal #5909 which will remove the first parameter from this builtin. However I think we should wait to make that change in stage2 until we're ready to do it in stage1 at the same time.

PR #8282, opened 2021-03-17
macho: upstream zld linker

### LINKED ISSUE #6971: macOS: lld lacks code signing for macOS/arm64
As reported by @jedisct1 (and confirmed by myself):

> macOS/arm64 beginning with Big Sur now requires everything (libraries and executables) to be signed. Or else, they instantly receive a SIGKILL. The system ld automatically signs everything it creates. However, lld currently doesn't.

This is tracker issue for upstream `lld` implementation of code signing.

---

EDIT: It is now possible to build Zig without falling back to system linker due to the fact Zig now patches the LLD generated binary and adds the missing code signature. The changes making it possible landed in #7273. As mentioned in the discussion below, this will work 99 out of 100 times; however we rely on the fact that there will be enough padding between the end of the load commands and the beginning of the `__text` section left by the LLD to be able to insert the `LC_CODE_SIGNATURE` load command, and this 1 out of 100 times this will not be the case, but Zig will inform of this error and gracefully exit. In such cases, you can either add more code to your binary (yeah, no kidding), or fall back to the system linker. Once LLD adds `-headerpad <size>` option, even if they don't add code signing, we should be able to always generate a valid code signature.

### LINKED ISSUE #7527: threadlocal generates a segfault on Apple Silicon

### LINKED ISSUE #8216: Bus error on Mac arm64 using threadlocal
```
threadlocal var x: i32 = 100;
pub fn main() !void {
    x += 1;
}
```

Not entirely sure if this is due to user error but this is exiting with an 138 exit code:

```
➜  zig run a.zig ; echo "Exited with: "$?
[1]    7626 bus error  zig run a.zig
Exited with: 138
```

I'm running zig installed through homebrew (brew install zig --HEAD)

uname -a: Darwin Kernel Version 20.3.0: Thu Jan 21 00:06:51 PST 2021; root:xnu-7195.81.3~1/RELEASE_ARM64_T8101 arm64

kubkon:
This PR is upstreaming [`zld`](https://github.com/kubkon/zld), the traditional linker, as a drop-in replacement for LLD when targeting MachO. `zld` is now currently used by default when cross-compiling Zig and C (C++ to join the club soon-ish), and it's the default linker for `aarch64`. This is a bold step, but @andrewrk and myself believe definitely worth a try even though `zld` is not fully functional yet and bugs are to be expected. However, we do believe this way we'll get bugs weed out much quicker and thus progress the linker development much faster. Also, a bonus for using `zld` is the fact that TLV is finally properly handled on Apple Silicon in Zig!

Fixes #8216
Fixes #7527
Fixes #6971


kubkon:
I just wanted to point out that we pass all tests on Apple Silicon with the exception of a few standalone tests producing or linking against shared libraries, and there seems to be an issue with alignment when running translate-c test suite. Apart from that, we are green with the rest of the test suite!

PR #8293, opened 2021-03-19
std: Add syscall7 stub for Linux/MIPS

LemonBoy:
Some syscalls such as fadvise require an extra argument to comply with
the register pair alignment imposed by the ABI.

Wacky, isn't it?

cc @vrischmann, this PR unblocks #8095


vrischmann:
Thank you !

PR #8295, opened 2021-03-19
zir-memory-layout: llvm backend: use new LazySrcLoc

g-w1:
This allows to compile with ninja/make now.

Not sure if this is the right way, but I tried to follow how you did the other backends.
Like this https://github.com/ziglang/zig/blob/abdbc11c7efb2f83804af64ac2fec10b972cdf2a/src/codegen/wasm.zig#L74-L78

### DIFF
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -15,6 +15,9 @@ const Inst = ir.Inst;
 const Value = @import("../value.zig").Value;
 const Type = @import("../type.zig").Type;
 
+const LazySrcLoc = Module.LazySrcLoc;
+const SrcLoc = Module.SrcLoc;
+
 pub fn targetTriple(allocator: *Allocator, target: std.Target) ![:0]u8 {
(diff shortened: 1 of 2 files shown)


src/codegen/llvm.zig
```diff
@@ -145,6 +148,7 @@ pub fn targetTriple(allocator: *Allocator, target: std.Target) ![:0]u8 {
 
 pub const LLVMIRModule = struct {
     module: *Module,
+    decl: ?*Module.Decl = null,
```
andrewrk, line 151: This field should be below the `// TODO: The fields below should ...` comment
(or better yet, let's take the opportunity here to do the suggested reworking and introduce a per-function struct for the state that mutates, instead of mutating `LLVMIRModule`.

FireFox317: I have a WIP branch where I did this, still have to upstream that

g-w1: Ah, Ill move it into todo for now, and we can do the refactor when you do it.

Issue #1595, opened 2018-09-27
request: distinct types


emekoi:
would it be possible to add distinct types? for example, to make it an error to pass a `GLuint` representing a shader as the program for `glAttachShader`.

monouser7dig:
do you mean something like strong typedefs? https://arne-mertz.de/2016/11/stronger-types/

I'm strongly for this 😃 👍

XVilka:
Certainly a good thing to have.

emekoi:
how would comptime provide this feature? what i mean is we could do something like `const ShaderProgram = distinct u32;` and it would be an compiler time error to pass a plain `u32` as a `ShaderProgram` and vice versa.

monouser7dig:
The current workaround is (like c) to use a Struct with just one member and then always pass the Struct instead of the wrapped value.
The big downside is that setting and getting the member is always boilerplate and does discourage the use of such a typesafe feature.

andrewrk:
Without yet commenting on the feature itself, if we were to do it, I would propose not changing any syntax, and instead adding a new builtin:

`const ShaderProgram = @distinct(u32);`

nodefish:
>how would comptime provide this feature? 

You're right, I conflated this with the "strong typedefs" described in the article posted above. They are distinct concepts after all, no pun intended.

Incidentally: for "strong typedefs," you could conceivably use comptime for code generation (similar to how generics work) to produce structs like what @monouser7dig described, with some abstraction for having them have different constraints.

emekoi:
yeah, i think `@distinct` is a better than `distinct`.

raulgrell:
I'm actually quite fond of this idea. Would there be any arguments to take it further and allow functions to be declared inside?

```
// Pass a block like in @cImport()
const ShaderProgram = @distinct(u32, {
    pub fn bind() void { ... }
    pub fn unbind() void { ... }
});
```

or an alternative way with minimal changes to syntax that is consistent with enum semantics of 'underlying type'.

```
const ShaderProgram = struct(u32) {
    pub fn bind() void { ... }
    pub fn unbind() void { ... }
};
```

emekoi:
@PavelVozenilek but thats because nim has operator overloading. in zig all the operators are known at compile time, so wouldn't the compiler be able to use the implementation of the distinct type's base type?

Ilariel:
I think distinct types are useful, but I don't think they fit in Zig. There should be only one obvious way to do things if possible and reasonable.

[...]

PR #8297, opened 2021-03-19
zir-memory-layout: astgen: fill in identifier

g-w1:


src/astgen.zig
```diff
@@ -1194,9 +1194,12 @@ fn varDecl(
...
             if (!nodeMayNeedMemoryLocation(scope, var_decl.ast.init_node)) {
-                const result_loc: ResultLoc = if (var_decl.ast.type_node != 0) .{
-                    .ty = try typeExpr(mod, scope, var_decl.ast.type_node),
-                } else .none;
+                const result_loc: ResultLoc = if (var_decl.ast.type_node != 0)
+                    .{
+                        .ty = try typeExpr(mod, scope, var_decl.ast.type_node),
+                    }
+                else
+                    .none;
```
andrewrk, line 1202: is this just a formatting change? I liked it the other way :cry:

src/astgen.zig
```diff
@@ -2851,7 +2861,7 @@ fn identifier(
     }
 
     if (simple_types.get(ident_name)) |zir_const_tag| {
-        return rvalue(mod, scope, rl, @enumToInt(zir_const_tag));
+        return rvalue(mod, scope, rl, @enumToInt(zir_const_tag), src);
```
andrewrk, line 2864: Oops. This is missing type safety here. #1595 would have caught this.

src/astgen.zig
```diff
@@ -2867,26 +2877,7 @@ fn identifier(
...
-                        }),
-                    }));
-                },
-            };
-            const result = try addZIRInstConst(mod, scope, src, .{
-                .ty = Type.initTag(.type),
-                .val = val,
-            });
-            return rvalue(mod, scope, rl, result);
+            return rvalue(mod, scope, rl, try gz.addBin(.int_type, @boolToInt(is_signed), bit_count), src);
```
andrewrk, line 2880: This has the same type safety issue (src is the wrong type here) 

Also in the case of common integer types, we want to avoid allocating an instruction entirely, by taking advantage of the `zir.Const` enum.

andrewrk (changes requested):

PR #8298, opened 2021-03-19
zir-memory-layout: astgen: literals and *, &

g-w1:
this was pretty low hanging fruit

### DIFF
--- a/src/astgen.zig
+++ b/src/astgen.zig
@@ -267,6 +267,8 @@ pub fn expr(mod: *Module, scope: *Scope, rl: ResultLoc, node: ast.Node.Index) In
     const node_tags = tree.nodes.items(.tag);
     const token_starts = tree.tokens.items(.start);
 
+    const gz = scope.getGenZir();
+
     switch (node_tags[node]) {
         .root => unreachable, // Top-level declaration.
         .@"usingnamespace" => unreachable, // Top-level declaration.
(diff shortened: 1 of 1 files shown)


src/astgen.zig
```diff
@@ -267,6 +267,10 @@ pub fn expr(mod: *Module, scope: *Scope, rl: ResultLoc, node: ast.Node.Index) In
     const node_tags = tree.nodes.items(.tag);
     const token_starts = tree.tokens.items(.start);
 
+    // TODO since some of the tags don't need a gen_zir (the const ones)
+    // should it be better to do this in the individual switch cases?
```
andrewrk, line 271: Not necessary. The optimizer is fully capable of moving this to inside the branches. It's also supposed to be a cheap function to call; I doubt it will be noticeable in terms of perf. If it is, what we really need to do is introduce a new type that we pass around in astgen.zig that makes the GenZir which is in scope always be directly available rather than accessible via an indirect branch from `*Scope`.

src/astgen.zig
```diff
@@ -443,56 +447,39 @@ pub fn expr(mod: *Module, scope: *Scope, rl: ResultLoc, node: ast.Node.Index) In
...
             const lhs = try expr(mod, scope, .none, node_datas[node].lhs);
             const src = token_starts[main_tokens[node]];
-            const result = try addZIRUnOp(mod, scope, src, .deref, lhs);
-            return rvalue(mod, scope, rl, result);
+            const result = try gz.addUnNode(.deref_node, lhs, src);
+            return rvalue(mod, scope, rl, result, src);
         },
         .address_of => {
-            if (true) @panic("TODO update for zir-memory-layout");
+            const src = token_starts[main_tokens[node]];
```
andrewrk, line 456: As a rule of thumb, all the `token_starts` should be deleted from this file. astgen no longer communicates source locations by absolute byte offsets.

andrewrk (changes requested):

src/zir.zig
```diff
@@ -111,6 +111,7 @@ pub const Const = enum {
     noreturn_type,
     null_type,
     undefined_type,
+    undefined_literal,
```
andrewrk, line 114: I believe you're looking for line 125 :)

PR #8301, opened 2021-03-19
implement fadvise for linux

vrischmann:
Continuation of #8095 but I forced-push and GitHub doesn't allow me to reopen the PR.

I think I took care of all special cases for the architectures that work in zig today.

### DIFF
--- a/lib/std/os/bits/linux.zig
+++ b/lib/std/os/bits/linux.zig
@@ -2409,6 +2409,27 @@ pub const MADV_PAGEOUT = 21;
 pub const MADV_HWPOISON = 100;
 pub const MADV_SOFT_OFFLINE = 101;
 
+pub const POSIX_FADV_NORMAL = 0;
+pub const POSIX_FADV_RANDOM = 1;
+pub const POSIX_FADV_SEQUENTIAL = 2;
+pub const POSIX_FADV_WILLNEED = 3;
+pub usingnamespace switch (arch) {
+    .s390x => if (@typeInfo(usize).Int.bits == 64)
+        struct {
+            pub const POSIX_FADV_DONTNEED = 6;
+            pub const POSIX_FADV_NOREUSE = 7;
+        }
+    else
+        struct {
+            pub const POSIX_FADV_DONTNEED = 4;
+            pub const POSIX_FADV_NOREUSE = 5;
+        },
+    else => struct {
+        pub const POSIX_FADV_DONTNEED = 4;
+        pub const POSIX_FADV_NOREUSE = 5;
+    },
+};
+
 pub const __kernel_timespec = extern struct {
     tv_sec: i64,
     tv_nsec: i64,
--- a/lib/std/os/linux.zig
+++ b/lib/std/os/linux.zig
@@ -1454,6 +1454,65 @@ pub fn process_vm_writev(pid: pid_t, local: [*]const iovec, local_count: usize, 
     );
 }
 
+pub fn fadvise(fd: fd_t, offset: i64, len: i64, advice: usize) usize {
+    if (comptime std.Target.current.cpu.arch.isMIPS()) {
+        // MIPS requires a 7 argument syscall
+
+        const offset_halves = splitValue64(offset);
+        const length_halves = splitValue64(len);
+
+        return syscall7(
+            .fadvise64,
+            @bitCast(usize, @as(isize, fd)),
+            0,
+            offset_halves[0],
+            offset_halves[1],
+            length_halves[0],
+            length_halves[1],
+            advice,
+        );
+    } else if (comptime std.Target.current.cpu.arch.isARM()) {
+        // ARM reorders the arguments
+
+        const offset_halves = splitValue64(offset);
+        const length_halves = splitValue64(len);
+
+        return syscall6(
+            .fadvise64_64,
+            @bitCast(usize, @as(isize, fd)),
+            advice,
+            offset_halves[0],
+            offset_halves[1],
+            length_halves[0],
+            length_halves[1],
+        );
+    } else if (@hasField(SYS, "fadvise64_64") and usize_bits != 64) {
+        // The extra usize check is needed to avoid SPARC64 because it provides both
+        // fadvise64 and fadvise64_64 but the latter behaves differently than other platforms.
+
+        const offset_halves = splitValue64(offset);
+        const length_halves = splitValue64(len);
+
+        return syscall6(
+            .fadvise64_64,
+            @bitCast(usize, @as(isize, fd)),
+            offset_halves[0],
+            offset_halves[1],
+            length_halves[0],
+            length_halves[1],
+            advice,
+        );
+    } else {
+        return syscall4(
+            .fadvise64,
+            @bitCast(usize, @as(isize, fd)),
+            @bitCast(usize, offset),
+            @bitCast(usize, len),
+            advice,
+        );
+    }
+}
+
 test {
     if (std.Target.current.os.tag == .linux) {
         _ = @import("linux/test.zig");
--- a/lib/std/os/linux/test.zig
+++ b/lib/std/os/linux/test.zig
@@ -108,3 +108,23 @@ test "user and group ids" {
     try expectEqual(linux.getauxval(elf.AT_EUID), linux.geteuid());
     try expectEqual(linux.getauxval(elf.AT_EGID), linux.getegid());
 }
+
+test "fadvise" {
+    const tmp_file_name = "temp_posix_fadvise.txt";
+    var file = try fs.cwd().createFile(tmp_file_name, .{});
+    defer {
+        file.close();
+        fs.cwd().deleteFile(tmp_file_name) catch {};
+    }
+
+    var buf: [2048]u8 = undefined;
+    try file.writeAll(&buf);
+
+    const ret = linux.fadvise(
+        file.handle,
+        0,
+        0,
+        linux.POSIX_FADV_SEQUENTIAL,
+    );
+    try expectEqual(@as(usize, 0), ret);
+}


lib/std/os/linux.zig
```diff
@@ -1386,6 +1386,58 @@ pub fn madvise(address: [*]u8, len: usize, advice: u32) usize {
     return syscall3(.madvise, @ptrToInt(address), len, advice);
 }
 
+pub fn posix_fadvise(fd: fd_t, offset: off_t, len: off_t, advice: usize) usize {
```
LemonBoy, line 1389: The naming is a bit unfortunate as the syscall is named `fadvise` while the libc wrapper is `posix_fadvise`...

lib/std/os/linux.zig
```diff
@@ -1386,6 +1386,58 @@ pub fn madvise(address: [*]u8, len: usize, advice: u32) usize {
...
+            offset_halves[1],
+            length_halves[0],
+            length_halves[1],
+            advice,
+        );
+    } else if (comptime std.Target.current.cpu.arch.isARM()) {
+        // ARM reorders the arguments
+
+        return syscall6(
+            .arm_fadvise64_64,
```
LemonBoy, line 1420: This is not needed as `arm_fadvise64_64` is aliased to `fadvise64_64`.

lib/std/os/linux.zig
```diff
@@ -1386,6 +1386,58 @@ pub fn madvise(address: [*]u8, len: usize, advice: u32) usize {
     return syscall3(.madvise, @ptrToInt(address), len, advice);
 }
 
+pub fn posix_fadvise(fd: fd_t, offset: off_t, len: off_t, advice: usize) usize {
+    if (usize_bits == 64 and @hasField(SYS, "fadvise64")) {
+        return syscall4(
+            .fadvise64,
```
LemonBoy, line 1392: The idea is to follow musl's strategy of using `fadvise64_64` if available and then fall back unconditionally to `SYS_fadvise64`. This order ensures the syscall with fewer arguments is used when available.

lib/std/os/linux.zig
LemonBoy, line 1389:
```suggestion
pub fn posix_fadvise(fd: fd_t, offset: u64, len: u64, advice: usize) usize {
```

The offset and length cannot be negative.

lib/std/os/linux/test.zig
```diff
@@ -108,3 +108,22 @@ test "user and group ids" {
...
+        file.close();
+        fs.cwd().deleteFile(tmp_file_name) catch {};
+    }
+
+    const stat = try file.stat();
+
+    const ret = linux.posix_fadvise(
+        file.handle,
+        0,
+        @bitCast(i64, stat.size),
```
LemonBoy, line 125: This part is a bit iffy as passing 0 here means "up to the end" and this file is empty...
I'd fill the file with garbage and then pass 0 here.

lib/std/os/linux.zig
vrischmann: Ok, I'll rename to `fadvise`.

lib/std/os/linux.zig
vrischmann: Addressed in the latest commit

lib/std/os/linux.zig
vrischmann: Addressed in the latest commit.

lib/std/os/linux/test.zig
vrischmann: Addressed in the last commit.

vrischmann:
`sparcv9` has both `fadvise64` and `fadvise64_64` but for some reason `fadvise64_64` is broken:

```
$ zig build -Dtarget=x86_64-linux
$ strace -e fadvise64_64,fadvise64 ./zig-cache/bin/zig-fadvise
fadvise64(3, 0, 28, POSIX_FADV_SEQUENTIAL) = 0
$ zig build -Dtarget=sparcv9-linux
$ strace -e fadvise64_64,fadvise64 ./zig-cache/bin/zig-fadvise
fadvise64(3, 0, 0, POSIX_FADV_NORMAL)   = 0
 ```

I can't test on a real sparc machine but I'm assuming it's broken given the strace output.

If I make it use `fadvise64` for the sparc architectures it's fine. Should I special case sparc too ?

**Comment by LemonBoy** [comment]
> sparcv9 has both fadvise64 and fadvise64_64 but for some reason fadvise64_64 is broken:

Hmm, it works for me. I tried running the attached test case under qemu and the parameters look ok.

File: lib/std/os/linux/test.zig (hunk 4 above)
Comment (LemonBoy) [reply lib/std/os/linux/test.zig]: There's no need to write random data, you can inflate the file by writing garbage or using fallocate.

Comment (vrischmann) [reply lib/std/os/linux/test.zig]: Ok, I've changed it to just write an undefined buffer.

**Comment by vrischmann** [comment]
> > sparcv9 has both fadvise64 and fadvise64_64 but for some reason fadvise64_64 is broken:
> 
> Hmm, it works for me. I tried running the attached test case under qemu and the parameters look ok.

Weird. I'm running the test binary with `strace -e fadvise64,fadvise64_64 qemu-sparc64` and I get this:
```
fadvise64(3, 0, 0, POSIX_FADV_NORMAL)   = 0
```
the advice should be `POSIX_FADV_SEQUENTIAL`.

**Comment by LemonBoy** [comment]
Well fuck me (and SPARC as a whole), that's the only 64bit platform where `fadvise64_64` is defined and behaves as a `fadvise64`.
Changing the check to `@hasField(SYS, "fadvise64_64") and usize_bits != 64` should solve the problem, please add a brief comment explaining why the extra check is needed.

**Comment by vrischmann** [comment]
I added the workaround and a comment in the latest commit.

For some reason one check is still shown as pending, but it finished successfully.

**Comment by vrischmann** [comment]
I noticed the offset/len types changed with https://github.com/ziglang/zig/commit/31f1cc9a0d80e87efb1f414352ff9d6abf2067ca so I changed it here too. Also rebased on latest master.

File: lib/std/os/bits/linux.zig [hunk 5]
```diff
@@ -2409,6 +2409,27 @@ pub const MADV_PAGEOUT = 21;
...
+pub const POSIX_FADV_RANDOM = 1;
+pub const POSIX_FADV_SEQUENTIAL = 2;
+pub const POSIX_FADV_WILLNEED = 3;
+pub usingnamespace switch (arch) {
+    .s390x => if (@typeInfo(usize).Int.bits == 64)
+        struct {
+            pub const POSIX_FADV_DONTNEED = 6;
+            pub const POSIX_FADV_NOREUSE = 7;
+        }
+    else
```
daurnimator, line 2422: what is going on with indentation here?

vrischmann: well that's what `zig fmt` produces.

alexrp:
@vrischmann @LemonBoy I'm unfortunately having to touch the `fadvise` code in #20389. I don't suppose you could shed some light on the sparc64 thing? I don't see any special-casing of sparc64 in glibc or musl. The kernel syscalls also look how I'd expect. So I'm genuinely puzzled at the need for the special case here. Is it possible you were observing a QEMU bug?

vrischmann:
Hi @alexrp. Sorry, I can't remember the details but I tested [this code](https://github.com/vrischmann/zig/blob/fcc72659130e5885ad68c4c1503eb0c62ceb22c1/lib/std/os/linux/test.zig#L112-L130) again right now and it looks like it works ?

```
$ zig build -Dtarget=x86_64-linux
$ strace -e fadvise64_64,fadvise64 ./zig-out/bin/zig-fadvise
fadvise64(3, 0, 0, POSIX_FADV_SEQUENTIAL) = 0
fadvise ret: 0+++ exited with 0 +++
$ zig build -Dtarget=sparc64-linux
$ strace -e fadvise64_64,fadvise64 qemu-sparc64 ./zig-out/bin/zig-fadvise
fadvise64(3, 0, 0, POSIX_FADV_SEQUENTIAL) = 0
fadvise ret: 0+++ exited with 0 +++
```

maybe you're right and it was a qemu bug ?

alexrp:
@vrischmann just to clarify, is that actually testing with a call to `fadvise64_64` or `fadvise64` in the code?

Also, did you actually observe it having visibility broken behavior? Now I'm wondering if it could just have been an earlier version of `strace` handling it wrong. :thinking:

vrischmann:
not sure I understand the question, the code calls `os.linux.fadvise` which calls `fadvise64` or `fadvise64_64`. But my test doesn't make any sense because I left the [special case for sparc64](https://github.com/vrischmann/zig/blob/master/lib/std/os/linux.zig#L2175-L2191).
To make sure I understand, you're saying that this special case should be unnecessary, right ? 

> Also, did you actually observe it having visibility broken behavior?
Sadly I can't remember if there was a broken test case or something.

alexrp:
> not sure I understand the question, the code calls `os.linux.fadvise` which calls `fadvise64` or `fadvise64_64`. But my test doesn't make any sense because I left the [special case for sparc64](https://github.com/vrischmann/zig/blob/master/lib/std/os/linux.zig#L2175-L2191).

That's exactly what I was getting at. :slightly_smiling_face: For testing purposes, you'd need to either undo the special case (or do a raw syscall I guess).

> To make sure I understand, you're saying that this special case should be unnecessary, right ?

I *think* so, yes. I can't find any evidence in Linux, glibc, or musl sources that indicate that it should be necessary. The two libcs have no special-casing for SPARC for this system call, and the two system calls end up in the same core function in the kernel.

vrischmann:
I don't know what changed but when built for `sparc64-linux` the special case isn't even hit: it's using the [else case](https://github.com/vrischmann/zig/blob/master/lib/std/os/linux.zig#L2191-L2199).
In my commit from 2021 I used `sparcv9-linux` instead but this architecture doesn't seem to exist anymore, I'm confused.

It looks like we could just disregard that special case now.

alexrp:
Hm, ok, I'll go ahead and remove that case then. Thanks for the assistance!

PR #8305, opened 2021-03-20
std/base64: cleanups & support url-safe and other non-padded variants

jedisct1:
This makes a few changes to the base64 codecs.

* The padding character is optional. The common "URL-safe" variant, in particular, is generally not used with padding. This is also the case for password hashes, so having this will avoid code duplication with bcrypt, scrypt and other functions.
* The URL-safe variant is added. Instead of having individual constants for each parameter of each variant, we are now grouping these in a struct. So, `standard_pad_char` just becomes `standard.pad_char`.
* Types are not `snake_case`'d any more. So, `standard_encoder` becomes `standard.Encoder`, as it is a type.
* Creating a decoder with ignored characters required the alphabet and padding. Now, `standard.decoderWithIgnore(<ignored chars>)` returns a decoder with the standard parameters and the set of ignored chars.
* Whatever applies to `standard.*` obviously also works with `url_safe.*`
* the `calcSize()` interface was inconsistent, taking a length in the encoder, and a slice in the encoder. Rename the variant that takes a slice to `calcSizeForSlice()`.
* In the decoder with ignored characters, add `calcSizeUpperBound()`, which is more useful than the one that takes a slice in order to size a fixed buffer before we have the data.
* Return `error.InvalidCharacter` when the input actually contains characters that are neither padding nor part of the alphabet. If we hit a padding issue (which includes extra bits at the end), consistently return `error.InvalidPadding`.
* Don't keep the `char_in_alphabet` array permanently in a decoder; it is only required for sanity checks during initialization.
* Tests are unchanged, but now cover both the standard (padded) and the url-safe (non-padded) variants.
 * Add an error set, rename `OutputTooSmallError` to `NoSpaceLeft`  to match the `hex2bin` equivalent.

### DIFF
--- a/lib/std/base64.zig
+++ b/lib/std/base64.zig
@@ -39,13 +39,33 @@ pub const standard = Codecs{
     .DecoderUnsafe = Base64DecoderUnsafe.init(standard_alphabet_chars, '='),
 };
 
+/// Standard Base64 codecs, without padding
+pub const standard_no_pad = Codecs{
+    .alphabet_chars = standard_alphabet_chars,
+    .pad_char = null,
+    .decoderWithIgnore = standardBase64DecoderWithIgnore,
+    .Encoder = Base64Encoder.init(standard_alphabet_chars, null),
+    .Decoder = Base64Decoder.init(standard_alphabet_chars, null),
+    .DecoderUnsafe = Base64DecoderUnsafe.init(standard_alphabet_chars, null),
+};
+
 pub const url_safe_alphabet_chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_".*;
 fn urlSafeBase64DecoderWithIgnore(ignore: []const u8) Base64DecoderWithIgnore {
     return Base64DecoderWithIgnore.init(url_safe_alphabet_chars, null, ignore);
 }
 
-/// URL-safe Base64 codecs, without padding
+/// URL-safe Base64 codecs, with padding
 pub const url_safe = Codecs{
+    .alphabet_chars = url_safe_alphabet_chars,
+    .pad_char = '=',
+    .decoderWithIgnore = urlSafeBase64DecoderWithIgnore,
+    .Encoder = Base64Encoder.init(url_safe_alphabet_chars, '='),
+    .Decoder = Base64Decoder.init(url_safe_alphabet_chars, '='),
+    .DecoderUnsafe = Base64DecoderUnsafe.init(url_safe_alphabet_chars, '='),
+};
+
+/// URL-safe Base64 codecs, without padding
+pub const url_safe_no_pad = Codecs{
     .alphabet_chars = url_safe_alphabet_chars,
     .pad_char = null,
     .decoderWithIgnore = urlSafeBase64DecoderWithIgnore,
@@ -378,10 +398,10 @@ test "base64" {
     comptime testAllApis(standard, "comptime", "Y29tcHRpbWU=") catch unreachable;
 }
 
-test "base64 url_safe" {
+test "base64 url_safe_no_pad" {
     @setEvalBranchQuota(8000);
-    testBase64UrlSafe() catch unreachable;
-    comptime testAllApis(url_safe, "comptime", "Y29tcHRpbWU") catch unreachable;
+    testBase64UrlSafeNoPad() catch unreachable;
+    comptime testAllApis(url_safe_no_pad, "comptime", "Y29tcHRpbWU") catch unreachable;
 }
 
 fn testBase64() !void {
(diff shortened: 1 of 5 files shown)


x13a:
What do you think about error set here?
```zig
pub const Error = struct {
    InvalidPadding,
    InvalidCharacter,
    OutputTooSmall,
};
```

jedisct1:
`hexToBytes()` uses `error.InvalidCharacter`, `error.InvalidLength` and `error.NoSpaceLeft`. Same thing, different names. So instead of each decoder having their own error set, maybe what we really need is a common one :)

jedisct1:
Lavahut's kaputt :(

x13a:
There is one assertion in [decode](https://github.com/ziglang/zig/blob/b718fee3cf926829fe65b4a556a5631153924197/lib/std/base64.zig#L334):
```zig
assert(dest.len == decoder.calcSizeForSlice(source) catch unreachable);
```
Maybe change to..?
```zig
assert(dest.len >= decoder.calcSizeForSlice(source) catch unreachable);
```

lib/std/base64.zig
```diff
@@ -8,308 +8,339 @@ const assert = std.debug.assert;
...
+
+pub const url_safe_alphabet_chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_".*;
+fn urlSafeBase64DecoderWithIgnore(ignore: []const u8) Base64DecoderWithIgnore {
+    return Base64DecoderWithIgnore.init(url_safe_alphabet_chars, null, ignore);
+}
+
+/// URL-safe Base64 codecs, without padding
+pub const url_safe = Codecs{
+    .alphabet_chars = url_safe_alphabet_chars,
+    .pad_char = null,
```
Mouvedia, line 50: Shouldn't the pad char be `=`?
cf https://tools.ietf.org/html/rfc4648#section-5

jedisct1:
> ```zig
> assert(dest.len == decoder.calcSizeForSlice(source) catch unreachable);
> ```
> 
> Maybe change to..?
> 
> ```zig
> assert(dest.len >= decoder.calcSizeForSlice(source) catch unreachable);
> ```

Is this a good idea? If the size is not properly computed, the rest will remain uninitialized. I can hardly come with a use case where a larger sized buffer would be provided, and where that wouldn't be a bug.

The unsafe variant doesn't even return the size nor a slice, so there wouldn't even be a way to detect if it was used with a correctly sized buffer.

lib/std/base64.zig
jedisct1: Unpadded variants seem to be far more common. But let's add a definition for both. Adding a new variant is pretty straightforward now :)

Mouvedia: It's known as "base64url".

jedisct1: Done!

Padded and non-padded variants are now available, both for standard and url-safe.

x13a:
>  I can hardly come with a use case where a larger sized buffer would be provided, and where that wouldn't be a bug.

I wrote about this because I got the same assertion in current std using `base64.standard_decoder`. Example is `phc_encoding` without `allocator`. As we don't know exact size, but know upper bound I wrap arrays in:
```zig
fn Value(comptime buf_len: usize) type {
    return struct {
        const Self = @This();

        buf: [buf_len]u8 = undefined,
        len: usize = 0,

        pub fn value(self: Self) []u8 {
            return self.buf[0..self.len];
        }
    };
}
```
When I used `decode` I got assertion and it was unexpected for me. Of course I have checked that buffer has enough size before this. Yes, I can pass a slice with fixed size to it, but.. Decoder has method `decoder.calcSizeForSlice(source)` that can be used to get output actual size. The `Unsafe` suffix in `Base64DecoderUnsafe` for me says that "I know what I am doing".

Or I missing something?

jedisct1:
> I wrote about this because I got the same assertion in current std using `base64.standard_decoder`. Example is `phc_encoding` without `allocator`. As we don't know exact size, but know upper bound I wrap arrays in:
>
> Or I missing something?

There are three versions of the codecs.

The first one doesn't ignore any characters. The input and output sizes are totally predictable, with, as you mentioned `calcSizeForSlice()` and `calcSize()`.

The second one can ignore input characters to decode. The maximum size can be computed with `calcSizeUpperBound()` and the decoding function returns the slice, properly truncated if necessary.

The last one is the `unsafe` variant, that doesn't check anything. Invalid characters in the input or padding cause out of bounds memory accesses, resulting in crashes (in safe modes) or, in non-safe modes, in potentially heartbleed-like information disclosure.
Changing the assertion to allow an invalid output length in addition to that wouldn't have much benefits, besides making that unsafe function even more unsafe.

The only use case for this unsafe decoder is when the input is completely trusted, possibly constant, such as in tests. But the difference in speed with "safe" decoders is likely to be negligible, and since they all work at `comptime`, I would rather see that unsafe variant go away TBH. Having 3 ways to decode a base64 string in a little bit ridiculous.

jedisct1:
The fact that `DecoderUnsafe` is the one used as an example in the documentation also doesn't seem like a good idea.

Would anyone object to it being removed?

andrewrk (approved):
Thanks! Looks great, is it ready to be merged now?

jedisct1:
It is!

PR #8306, opened 2021-03-20
stage2: restructure LLVM backend

FireFox317:
This is a PR against the `zir-memory-layout` branch.

The LLVM backend is now structured into 3 different structs, namely
Object, DeclGen and FuncGen. Object represents an object that is
generated by the LLVM backend. DeclGen is responsible for generating
a decl and FuncGen is responsible for generating llvm instructions
from tzir in a function.


andrewrk:
Nice work!

PR #8307, opened 2021-03-20
zir-memory-layout: astgen: more instructions

g-w1:


src/astgen.zig
```diff
@@ -470,30 +466,27 @@ pub fn expr(mod: *Module, scope: *Scope, rl: ResultLoc, node: ast.Node.Index) In
...
                 ),
-                else => return rvalue(mod, scope, rl, try addZIRUnOp(
-                    mod,
-                    scope,
-                    src,
+                else => return rvalue(mod, scope, rl, try gz.addUnTok(
                     .optional_payload_safe,
                     try expr(mod, scope, .none, node_datas[node].lhs),
-                )),
+                    src,
```
andrewrk, line 488: looks like passing a byte offset where a token index is expected. Be really careful about these - the compiler can't catch these errors for us unless we figure out how to make the types be distinct!

I think probably all the `token_starts` from this entire file should be deleted.

andrewrk (changes requested):

PR #8309, opened 2021-03-20
std: Add reset to TokenIterator

Pombal:


andrewrk (changes requested):
Thanks for the patch!

I'm fine with this, although I also think it would be equally fine to instead add a documentation comment on the `index` field noting that it is part of the public API to set this to zero to reset the iterator.

Either way, would you mind adding test coverage for this? Should be straightforward, I think, and the test cases are right there nearby the implementation.

andrewrk (approved):
Thanks!

Pombal:
You are right, I added it because it wasn't obvious to me at first that I could use the index directly.
Looking at how other iterators are used, I saw the reset pattern, and thought other people might try that first too.

PR #8313, opened 2021-03-21
Add core statistical math functions

### LINKED ISSUE #7212: Consider adding 8 math functions to std.math for statistics applications
Evan Miller (@evanmiller) made a great case that adding just a few more math functions to the standard `<math.h>` C library would enable a huge range of statistical applications. Or to quote him:

> Basic statistical analysis requires special mathematical functions which almost no standard math libraries implement. These functions are not particular to statistics, but they form building blocks for a number of important statistical applications. If standard math libraries were to provide just a few more special functions, including incomplete beta, incomplete gamma, inverse normal, and better Bessel functions, developers would be able to integrate more statistical functions into their applications without having to license specialized libraries, or copy-paste code of dubious origin into their projects. (https://www.evanmiller.org/statistical-shortcomings-in-standard-math-libraries.html)

Since zig provides a rather comprehensive standard library (e.g. all the great stuff in std.crypto) it might be a good idea to add those functions to the standard math library, such that anybody can build great statistical/machine learning work in zig without resorting to third party math libraries.

Just an _idea_. The linked article provides the full rational, but I will list the 8 functions here for completneness:

> I realize that implementing all of these specialized numerical functions sounds like a lot of work, but in fact the work is essentially finished: the free Cephes library contains a host of mathematical functions with names and argument orders nearly identical to their <math.h> counterparts, as well as all of the functions I have mentioned above. So my proposal boils down to “graduating” the following eight double-precision functions, along with their single-precision and extended precision counterparts, from Cephes into the standard C library:

```
/* Regularized incomplete beta function */
double incbet(double a, double b, double x);

/* Inverse of incomplete beta integral */
double incbi(double a, double b, double y);

/* Regularized incomplete gamma integral */
double igam(double a, double x);

/* Complemented incomplete gamma integral */
double igamc(double a, double x);

/* Inverse of complemented incomplete gamma integral
double igami(double a, double p);

/* Normal distribution function */
double ndtr(double x);

/* Inverse of Normal distribution function */
double ndtri(double y);

/* Bessel function of non-integer order */
double jv(double v, double x);
```

tiehuis:
The key functions added are:

    // Regularized incomplete beta function
    pub fn incbet(a: f64, b: f64, x: f64) f64

    // Inverse of incomplete beta integral
    pub fn incbi(a: f64, b: f64, y: f64) f64

    // Regularized incomplete gamma integral
    pub fn igam(a: f64, x: f64) f64

    // Complemented incomplete gamma integral
    pub fn igamc(a: f64, x: f64) f64

    // Inverse of complemented incomplete gamma integral
    pub fn igami(a: f64, p: f64) f64

    // Normal distribution function
    pub fn ndtr(x: f64) f64

    // Inverse of Normal distribution function
    pub fn ndtri(y: f64) f64

    // Bessel function of non-integer order
    pub fn jv(v: f64, x: f64) f64

Some extra functions besides these have been exposed as well, largely
functions which the above are dependent on, but which may have relevant
use outside.

Tests are largely generated from external sources and verified, since
cephes, the library which these implementations has been ported from
lacks formal testcases.

Closes #7212.


tiehuis:
Done an initial implementation in the above PR. Marked a draft since I want to do some more testing more but it is largely complete.

There are also a few questions regarding exposed names, etc, which I haven't yet changed. I've pushed now just to indicate there is some more work here.

tiehuis:
Code is ported from [cephes](https://netlib.org/cephes/). Specifically the bessel.tgz and cprob.tgz archives. I only ported functions which were required in implementing the requested functions, but there are a few more easy additions that could be added on top of these.

Regarding the license, I had a bit more of a look and it has been used fine in BSD's. See the following remarks from the original author Although someone else can make a comment otherwise since it could be specific to that.
https://lists.debian.org/debian-legal/2004/12/msg00295.html

Anyway, here's the code.

daurnimator:
The function names seem unreasonably short/badly abbreviated. e.g.
  - `inc` expands in my brain to 'increment', not 'incomplete'
  - `ndtri` why is 'distribution' shortened to 'dtr'?

lib/std/math/prob/gamma.zig
```diff
@@ -0,0 +1,368 @@
...
+        return math.nan(f64); // Domain
+    }
+    return z / ((1.0 + 0.5772156649015329 * x) * x);
+}
+
+/// Gamma function
+///
+/// Returns gamma function of the argument.  The result is
+/// correctly signed, and the sign (+1 or -1) is also
+/// returned in a global (extern) variable named sgngam.
```
daurnimator, line 86: I don't see a global extern variable here.... but also that sounds like a terrible idea

lib/std/math/prob/gamma.zig
```diff
@@ -0,0 +1,368 @@
...
+
+    x -= 2.0;
+    var p = polevl(x, P[0..]);
+    q = polevl(x, Q[0..]);
+    return z * p / q;
+}
+
+const expectApproxEqRel = std.testing.expectApproxEqRel;
+const expect = std.testing.expect;
+const epsilon = 0.000001;
```
daurnimator, line 187: We have `std.testing.expectApproxEqAbs`

daurnimator (changes requested):

lib/std/math/prob/gamma.zig
tiehuis: Yeah I'll need to update the documentation. It is taken verbatim from the C source mostly.

tiehuis:
I haven't changed the names from upstream yet, happy to have a back-and-forth to settle on better alternatives.

I was thinking about

```
betaRegInc
betaInvInc
gammaRegInc
gammaCompInc
gammaInvCompInc
distNormal
distInvNormal
besselj
```

however it might flow better if we use the english description as a base such as `regIncBeta`. I'm just a bit partial to the uniformity in namespacing. The full names are a little verbose, I feel like this would be a decent medium between needlessly opaque and too long.

daurnimator:
What if we used the names from Mathmatica/Wolfram Alpha?

e.g.
  - > The incomplete beta function is implemented in the Wolfram Language as `Beta[z, a, b]`. 
  - > The regularized beta function is sometimes also denoted I_z(a,b) and is implemented in the Wolfram Language as `BetaRegularized[z, a, b]`.
  - > The function P(a,z) is implemented in the Wolfram Language as `GammaRegularized[a, 0, z]`

LemonBoy:
Cephes [provides](https://www.netlib.org/cephes/) some implementations for `f32` and `f128`, the latter is pretty much useless but I can see the former being useful on platforms where the FPU only supports single-precision floats (eg. some ARM chips).

tiehuis:
Renamed functions to:

```
incbet  incompleteBeta
incbi   inverseIncompleteBeta

lgam    lnGamma
igam    incompleteGamma
igamc   complementedIncompleteGamma
igami   inverseComplementedIncompleteGamma

ndtr    normalDist
ndtri   inverseNormalDist

jv      besselj

tiehuis:
Been a bit out of action, but tentatively planning to finish this up this weekend and promote from a draft.

andrewrk:
30+ days old draft; closing

scheibo:
@tiehuis This PR seems desirable and seemed fairly close to being shippable - do you plan to return to this? :)

tiehuis:
I would like to, but can't commit completely so no guarantee. Would be happy if someone else was keen to take it over too.

The remaining work I wanted to do was improving the tests and coverage and any fixes that go along with that. A clean up of the documentation as well would be useful.

PR #8315, opened 2021-03-21
Specify type in autoHash error message

### LINKED ISSUE #7970: Specify type that cannot be hashed in autoHash error message.
Hi,

I was looking to use the AutoHashMap where the key is a struct. The following code example fails to compile: 

```zig
const std = @import("std");

pub fn main() anyerror!void {
}

const Vertex = struct {
    p: [3]f32,
    c: [3]f32,
    t: [2]f32,
};

test "hashmap" {
    var allocator = std.testing.allocator;
    var hm = std.hash_map.AutoHashMap(Vertex, u64).init(allocator);
    defer hm.deinit();
    
}
```
With the error:

```
zig test src/main.zig
/zig/lib/std/hash/auto_hash.zig:98:12: error: cannot hash this type
        => @compileError("cannot hash this type"),
           ^
/zig/lib/std/hash/auto_hash.zig:65:21: note: called from here
                hash(hasher, element, .Shallow);
                    ^
/zig/lib/std/hash/auto_hash.zig:113:28: note: called from here
        .Array => hashArray(hasher, key, strat),
                           ^
/zig/lib/std/hash/auto_hash.zig:133:21: note: called from here
                hash(hasher, @field(key, field.name), strat);
                    ^
/zig/lib/std/hash/auto_hash.zig:180:9: note: called from here
    hash(hasher, key, .Shallow);
        ^
/zig/lib/std/hash_map.zig:26:25: note: called from here
                autoHash(&hasher, key);
                        ^
/zig/lib/std/hash_map.zig:21:29: note: called from here
        fn hash(key: K) u64 {
                            ^
```

From what I can see, `lib/std/hash/auto_hash.zig`'s `hash` function should support structs, but maybe it's not working properly anymore? I could of course be trying to do this wrong, or I mis-understand and the structs aren't supported.

What I understand to be happening presently in the following code snippet from `auto_hash.zig`, is that the type of `key` is stored instead of the the key (already a type) itself. Therefore, when `@typeInfo(Key)` is evaluated, the type is always `.void`?

```zig
pub fn hash(hasher: anytype, key: anytype, comptime strat: HashStrategy) void {
    const Key = @TypeOf(key);
    ...
    switch (@typeInfo(Key)) {
       ...
    }

```

As an additional example the following is output when I test:

```zig
    const Key = @TypeOf(Vertex);
    std.log.warn("typeInfo(Vertex): '{}'", .{@typeInfo(Vertex)});
    std.log.warn("typeInfo(Key): '{}'", .{@typeInfo(Key)});
```

```
(warn): typeInfo(Vertex): 'TypeInfo{ .Struct = Struct{ .layout = ContainerLayout.Auto, .fields = StructField@202f98, .decls = Declaration@0, .is_tuple = false } }'
[default] (warn): typeInfo(Key): 'TypeInfo{ .Type = void }'
```

Is AutoHashMap intended to support hashing structs? If so, how is it supposed to be used?

JorisHartog:
This commit simply specifies the type which `autoHash` can't hash in the
compile error.

Closes #7970.

### DIFF
--- a/lib/std/hash/auto_hash.zig
+++ b/lib/std/hash/auto_hash.zig
@@ -95,7 +95,7 @@ pub fn hash(hasher: anytype, key: anytype, comptime strat: HashStrategy) void {
         .EnumLiteral,
         .Frame,
         .Float,
-        => @compileError("cannot hash this type"),
+        => @compileError("unable to hash type " ++ @typeName(Key)),
 
         // Help the optimizer see that hashing an int is easy by inlining!
         // TODO Check if the situation is better after #561 is resolved.


Vexu:
Thanks!

PR #8318, opened 2021-03-21
zir: add negate/negate_wrap, implement astgen

ifreund:
These were previously implemented as a sub/sub_wrap instruction with a
lhs of 0. Making this separate instructions however allows us to save
some memory as there is no need to store a lhs.

This commit was written under the assumption that every zir instruction requires source information, which will not fit in the 8 byte `Data` union alongside 2 32 bit indexes to lhs and rhs nodes.

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -1462,6 +1462,8 @@ pub const WipZirCode = struct {
                 .str,
                 .sub,
                 .subwrap,
+                .negate,
+                .negate_wrap,
                 .typeof,
                 .xor,
                 .optional_type,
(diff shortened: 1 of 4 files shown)


ifreund:
> This commit was written under the assumption that every zir instruction requires source information, which will not fit in the 8 byte `Data` union alongside 2 32 bit indexes to lhs and rhs nodes.

On further examination, it looks like source location data is not stored for some zir instructions where it is not necessary. However, I'm pretty sure it is necessary in general for binary operators such as add/sub. I indent to implement astgen for those next.

andrewrk (approved):
Looks great!

PR #8328, opened 2021-03-22
astgen: implement pointer type

FireFox317:
This now compiles:
```zig
extern fn puts(s: [*:0]const u8) c_int;

export fn main() c_int {
    _ = puts("hello world!");
    return 1;
}
```


FireFox317:
Implemented by @ifreund in f3770dcc307e4207de8e38c44bcf1bf98dc8448c

Issue #6167, opened 2020-08-26
use-after-free occurring in std.math.big.Rational on 32-bit architectures


andrewrk:
Repro with i386-linux:

```
$ ./zig test ../lib/std/math/big/rational.zig --test-filter "setFloatString"  -target i386-native
Test [1/1] test "big.rational setFloatString"...error: TargetTooSmall
/home/andy/Downloads/zig/lib/std/heap/general_purpose_allocator.zig:561:13: 0x40481e in std.heap.general_purpose_allocator.GeneralPurposeAllocator((struct std.heap.general_purpose_allocator.Config constant)).resize (test)
            return error.OutOfMemory;
            ^
/home/andy/Downloads/zig/lib/std/heap/general_purpose_allocator.zig:561:13: 0x40481e in std.heap.general_purpose_allocator.GeneralPurposeAllocator((struct std.heap.general_purpose_allocator.Config constant)).resize (test)
            return error.OutOfMemory;
            ^
/home/andy/Downloads/zig/lib/std/heap/general_purpose_allocator.zig:561:13: 0x40481e in std.heap.general_purpose_allocator.GeneralPurposeAllocator((struct std.heap.general_purpose_allocator.Config constant)).resize (test)
            return error.OutOfMemory;
            ^
/home/andy/Downloads/zig/lib/std/heap/general_purpose_allocator.zig:561:13: 0x40481e in std.heap.general_purpose_allocator.GeneralPurposeAllocator((struct std.heap.general_purpose_allocator.Config constant)).resize (test)
            return error.OutOfMemory;
            ^
/home/andy/Downloads/zig/lib/std/heap/general_purpose_allocator.zig:561:13: 0x40481e in std.heap.general_purpose_allocator.GeneralPurposeAllocator((struct std.heap.general_purpose_allocator.Config constant)).resize (test)
            return error.OutOfMemory;
            ^
/home/andy/Downloads/zig/lib/std/heap/general_purpose_allocator.zig:561:13: 0x40481e in std.heap.general_purpose_allocator.GeneralPurposeAllocator((struct std.heap.general_purpose_allocator.Config constant)).resize (test)
            return error.OutOfMemory;
            ^
/home/andy/Downloads/zig/lib/std/math/big/int.zig:1025:21: 0x40e8eb in std.math.big.int.Const.to (test)
                    return error.TargetTooSmall;
                    ^
[...]

PR #8330, opened 2021-03-22
bigint add failures with aliasing

kivikakk:
A Discord user came to us with a failing use of bigint which was just calculating the Fibonacci sequence.

It turns out it was miscalculating on the jump between one and two limbs. See attached a test that fails currently, but should succeed. It is not written very elegantly, but it's easy to verify that it should work.

It looks like the offending code is this optimisation in `Mutable.add`:

https://github.com/ziglang/zig/blob/dce612ac2be5aa8a1aa0ee8dd670d7e875624216/lib/std/math/big/int.zig#L318-L324

Note that, if `@addWithOverflow` does overflow, `add` continues on to perform Knuth's algorithm in `ldadd`, **but** it has also overwritten the contents of `r.limbs[0]` here.  If `r` is aliased with one of its inputs, this causes the subsequent non-optimised calculation to fail.

I've confirmed the test passes if this is removed. We may want to do an alias check and skip the optimisation if overwriting `r` with garbage here will be a problem.

edit: I've added a suggested fix in case it's the right one, which skips this `@addWithOverflow` attempt if `r.limbs.ptr` is equal to `a.limbs.ptr` or `b.limbs.ptr`.

/cc @alexnask fyi

### DIFF
--- a/lib/std/math/big/int.zig
+++ b/lib/std/math/big/int.zig
@@ -316,7 +316,9 @@ pub const Mutable = struct {
         }
 
         if (a.limbs.len == 1 and b.limbs.len == 1 and a.positive == b.positive) {
-            if (!@addWithOverflow(Limb, a.limbs[0], b.limbs[0], &r.limbs[0])) {
+            var o: Limb = undefined;
+            if (!@addWithOverflow(Limb, a.limbs[0], b.limbs[0], &o)) {
+                r.limbs[0] = o;
                 r.len = 1;
                 r.positive = a.positive;
                 return;
--- a/lib/std/math/big/int.zig
+++ b/lib/std/math/big/int.zig
@@ -335,10 +335,10 @@ pub const Mutable = struct {
             }
         } else {
             if (a.limbs.len >= b.limbs.len) {
-                lladd(r.limbs[0..], a.limbs[0..a.limbs.len], b.limbs[0..b.limbs.len]);
+                lladd(r.limbs[0..], a.limbs, b.limbs);
                 r.normalize(a.limbs.len + 1);
             } else {
-                lladd(r.limbs[0..], b.limbs[0..b.limbs.len], a.limbs[0..a.limbs.len]);
+                lladd(r.limbs[0..], b.limbs, a.limbs);
                 r.normalize(b.limbs.len + 1);
             }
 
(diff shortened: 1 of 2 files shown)


lib/std/math/big/int_test.zig
```diff
@@ -1562,3 +1562,16 @@ test "big.int pow" {
         testing.expectEqual(@as(i32, 1), try a.to(i32));
     }
 }
+
+test "regression test for 1 limb overflow" {
+    // Note these happen two be two consecutive Fibonacci sequence numbers, the
```
daurnimator, line 1567: s/two be two/to be two/

kivikakk: whew, thanks

lib/std/math/big/int.zig
```diff
@@ -315,7 +315,7 @@ pub const Mutable = struct {
             return;
         }
 
-        if (a.limbs.len == 1 and b.limbs.len == 1 and a.positive == b.positive) {
+        if (a.limbs.len == 1 and b.limbs.len == 1 and a.positive == b.positive and r.limbs.ptr != a.limbs.ptr and r.limbs.ptr != b.limbs.ptr) {
```
Vexu, line 318: I think a better fix would be to add a stack temporary as the result location for the `@addWithOverflow` and then store that in `r.limbs[0]` since that would allow the optimization even if `a` or `b` alias `r`.

kivikakk: Much better idea!

kivikakk:
CI green (modulo FreeBSD CI outage), should be good now.

kivikakk:
Our Discord friend has found another bug of the same class, so marking as a draft until that's fixed too. Seems to happen when we hit the 2<sup>192</sup> limit, again only with aliasing.

kivikakk:
https://github.com/ziglang/zig/blob/749c3f014ba36a6dc824419f835790f9c181aad3/lib/std/math/big/int.zig#L1699-L1707

If `r` is aliased with `a` or `b` and `r.ensureCapacity` needs to realloc:

https://github.com/ziglang/zig/blob/749c3f014ba36a6dc824419f835790f9c181aad3/lib/std/math/big/int.zig#L1472-L1477

The memory pointed to by the alias is no longer valid.

kivikakk:
Ref https://github.com/ziglang/zig/issues/6167#issuecomment-680456728.

kivikakk:
CI green again. `add` is now a lot more correct. Other functions probably still need work.

kivikakk:
Rebased and brought back up to date.

PR #8335, opened 2021-03-22
minor fix in astgenAndSemaFn

dimenus:
Should we use a scratch buffer instead of multiple reallocs?


g-w1:
Should be against zir-memory-layout branch.

PR #8336, opened 2021-03-22
fix usage calculation in astgenAndSemaFn

dimenus:
Should we use a scratch buffer instead of multiple re-allocs?


andrewrk:
Thanks!

> Should we use a scratch buffer instead of multiple re-allocs?

For what?

PR #8338, opened 2021-03-22
zir-memory-layout: astgen: async stuff

g-w1:
Things to be resolved:

- [ ] we need a source for a suspend block for notes
- [ ] search for `how do I do this`


g-w1:
This is being deleted as it has no test coverage.

PR #8339, opened 2021-03-22
stage2: Wasm control flow

Luukdegram:
This adds extra control flow support to the wasm backend.
e.g the following now compiles  successfully for wasm:
```zig
export fn _start() void {
    assert(foo(false) == @as(i32, 20));
    assert(foo(true) == @as(i32, 30));
}

fn assert(ok: bool) void {
    if (!ok) unreachable;
}

fn foo(ok: bool) i32 {
    const val: i32 = blk: {
        var x: i32 = 1;
        if (!ok) break :blk x + @as(i32, 9);
        break :blk x + @as(i32, 19);
    };
    return val + 10;
}


andrewrk:
Nice work!

PR #8341, opened 2021-03-23
translate-c: intcast compound assignment operand if different-size

ehaas:
The compound assignment logic is still not quite right but this fixes a compile
error triggered by a common case (`x += 1` when `x` is a `short int`). Ideally
the compound assignment logic would use the same integer casting logic as
`transCCast` (following still fails due to missing truncate: `x += 10000000`)


Vexu:
> Ideally the compound assignment logic would use the same integer casting logic as `transCCast`

Is there a reason why we can't just use `transCCast` here? It seems like it should work well and there would not be any rendering issues thanks to the new AST.

ehaas:
lol, I don't even know why I didn't think to try that. I'll take a look.

ehaas:
Updated...erred on the side of over-casting but I think the code is more correct now and slightly more clear

PR #8346, opened 2021-03-23
astgen: fix an issue where the alloc wasnt elided

FireFox317:
I'm not sure if this is the proper way to solve the problem, but it does solve the problem. `addUnNode` says it returns an `Ref`, but the value was really big and it looks like it is actually an `Index`?
https://github.com/ziglang/zig/blob/668148549a822c7fa680cf08999dd845bde765aa/src/astgen.zig#L1309

We really need distinct types for Index and Ref.

### DIFF
--- a/src/astgen.zig
+++ b/src/astgen.zig
@@ -1306,11 +1306,11 @@ fn varDecl(
             if (var_decl.ast.type_node != 0) {
                 const type_inst = try typeExpr(mod, &init_scope.base, var_decl.ast.type_node);
                 opt_type_inst = type_inst;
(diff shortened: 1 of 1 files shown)


andrewrk:
> `addUnNode` says it returns an `Ref`, but the value was really big and it looks like it is actually an `Index`?

Your code is right, and the function is right - the only thing confused here is the expectation about which number will be higher. Ref is a big number, Index is a small number. Convert Index to Ref by adding `zir.const_inst_list.len + param_count` and convert Ref to Index by subtracting.

> We really need distinct types for Index and Ref.

Completely agree. Sounds like @ifreund is working on it :)

PR #8348, opened 2021-03-23
astgen: implement orelse

FireFox317:
this big boi test passes (expect for one line: `var val3_var = opt_val orelse 30;`, which i think has something to do with `.alloc_inferred_mut`):
https://github.com/ziglang/zig/blob/749c3f014ba36a6dc824419f835790f9c181aad3/test/stage2/llvm.zig#L140-L181


FireFox317:
Edit: Fixed the bug and the test now passes! Ready for review + merge!

PR #8349, opened 2021-03-23
Change defineCMacro to take separate name and value arugments

asa-z:
Before this change, when one or more of name or value are not known at
comptime, build.zig files must allocate and do the concatenation, which can be
cumbersome, and also adds a redundant allocation when name and value are
slices. The new version only does a single allocation directly in the builder's
allocator to concatenate name and value

Additional options considered: 
 - not do it at all 
 - do it in a new function (`defineCMacroKeyValue`?)
 - do it, but also keep the old behavior in a different function (`defineCMacroFlag`?)


asa-z:
Looks like I did not run enough of the test suite at first. Updated to fix. I also ended up taking the third alternative listed above because of the things I had to fix.

PR #8354, opened 2021-03-24
std.os.uefi.tables.boot_services: implement locateDevicePath

sreehax:
This is necessary to open the root volume as described [here](https://github.com/rust-osdev/uefi-rs/issues/175#issuecomment-787421884), since the default SimpleFileSystem volume is often not the boot volume. This is my first pull request and I'm still a bit of a beginner to Zig so let me know if there's any changes I need to make (optionals, etc.)


kivikakk:
This looks right:

image

You could argue either way for whether the last parameter should be nullable — the spec suggests a UEFI implementation needs to handle the possibility of a null there, but it won't be very useful if you do pass one.

PR #8355, opened 2021-03-24
std.math.sqrt_int: fixed odd size integers types

Virv12:
I found a bug in the implementation of std.math.sqrt_int.
The implementation assumed the number of bits of the type was even.
I added some tests and fixed the bug.
I also added a shortcut for small number of bits to prevent edge cases.

### DIFF
--- a/lib/std/math/sqrt.zig
+++ b/lib/std/math/sqrt.zig
@@ -39,55 +39,50 @@ pub fn sqrt(x: anytype) Sqrt(@TypeOf(x)) {
 }
 
 fn sqrt_int(comptime T: type, value: T) Sqrt(T) {
-    switch (T) {
-        u0 => return 0,
-        u1 => return value,
-        else => {},
-    }
+    if (@typeInfo(T).Int.bits <= 2) {
+        return if (value == 0) 0 else 1; // shortcut for small number of bits to simplify general case
+    } else {
+        var op = value;
+        var res: T = 0;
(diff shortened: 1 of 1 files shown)


ehaas:
Perhaps this should be addressed as a separate PR but the current implementation (both in master and this PR) enters an infinite loop if the argument to `sqrt` is a runtime-known negative integer. e.g.

```zig
test "infinite loop" {
    var x: i8 = -1;
    _ = @import("std").math.sqrt(x);
}
```

At the very least this should be documented; perhaps instead the negative integer case should be checked and `@panic` called? Alternatively the signature would have to be changed to accomodate an error value.

jedisct1:
> At the very least this should be documented; perhaps instead the negative integer case should be checked and `@panic` called? Alternatively the signature would have to be changed to accommodate an error value.

Having an `assert()` for that case would be reasonable.

jedisct1:
Would you mind rebasing this and fixing the indentation?

ehaas:
> > At the very least this should be documented; perhaps instead the negative integer case should be checked and `@panic` called? Alternatively the signature would have to be changed to accommodate an error value.
> 
> Having an `assert()` for that case would be reasonable.

It looks like that's no longer necessary; #8410 changed sqrt etc to only take unsigned integers.

Virv12:
> Would you mind rebasing this and fixing the indentation?

I rebased it and the indentation looks good to me.
Is there anything else to do before merging?

PR #8356, opened 2021-03-25
Child Process Exit Code

### LINKED ISSUE #6342: make the types match of std.process.exit and std.ChildProcess exit code
In the self hosted compiler I have this code:

```zig
const child = try std.ChildProcess.init(argv.items, arena);
defer child.deinit();

if (comp.clang_passthrough_mode) {
    child.stdin_behavior = .Inherit;
    child.stdout_behavior = .Inherit;
    child.stderr_behavior = .Inherit;

    const term = child.spawnAndWait() catch |err| {
        return comp.failCObj(c_object, "unable to spawn {}: {}", .{ argv.items[0], @errorName(e
r) });
    };
    switch (term) {
        .Exited => |code| {
            if (code != 0) {
                // TODO make std.process.exit and std.ChildProcess exit code have the same type
                // and forward it here. Currently it is u32 vs u8.
                std.process.exit(1);
            }
        },
        else => std.process.exit(1),
    }
```

I wanted to do `std.process.exit(code)` which should in theory work but the types don't match.

gsquire:
Fixes #6342

### DIFF
--- a/lib/std/child_process.zig
+++ b/lib/std/child_process.zig
@@ -78,7 +78,7 @@ pub const ChildProcess = struct {
     } || os.ExecveError || os.SetIdError || os.ChangeCurDirError || windows.CreateProcessError || windows.WaitForSingleObjectError;
 
     pub const Term = union(enum) {
-        Exited: u32,
+        Exited: u8,
         Signal: u32,
         Stopped: u32,
         Unknown: u32,
@@ -347,7 +347,7 @@ pub const ChildProcess = struct {
             if (windows.kernel32.GetExitCodeProcess(self.handle, &exit_code) == 0) {
                 break :x Term{ .Unknown = 0 };
             } else {
-                break :x Term{ .Exited = exit_code };
+                break :x Term{ .Exited = @truncate(u8, exit_code) };
             }
         });
 
--- a/lib/std/os/bits/netbsd.zig
+++ b/lib/std/os/bits/netbsd.zig
@@ -686,8 +686,8 @@ pub const TIOCSWINSZ = 0x80087467;
 pub const TIOCUCNTL = 0x80047466;
 pub const TIOCXMTFRAME = 0x80087444;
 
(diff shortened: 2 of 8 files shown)


daurnimator:
I think the more correct fix would be for `WEXITSTATUS` to return a u8?

gsquire:
So for each OS in [this search](https://github.com/ziglang/zig/search?p=2&q=WEXITSTATUS), you're suggesting we change the return type to a `u8`?

daurnimator:
> So for each OS in [this search](https://github.com/ziglang/zig/search?p=2&q=WEXITSTATUS), you're suggesting we change the return type to a `u8`?

yes.
If that's not correct, then what's in this PR wouldn't be correct either.

lib/std/os/bits/netbsd.zig
```diff
@@ -836,15 +836,13 @@ pub const ucontext_t = extern struct {
...
-    __pad: [
-        switch (builtin.arch) {
-            .i386 => 4,
-            .mips, .mipsel, .mips64, .mips64el => 14,
-            .arm, .armeb, .thumb, .thumbeb => 1,
-            .sparc, .sparcel, .sparcv9 => if (@sizeOf(usize) == 4) 43 else 8,
-            else => 0,
-        }
-    ]u32,
+    __pad: [switch (builtin.arch) {
```
gsquire, line 839: This was from my editor auto-formatting via `zig fmt`.

lib/std/child_process.zig
```diff
@@ -324,7 +324,7 @@ pub const ChildProcess = struct {
             if (windows.kernel32.GetExitCodeProcess(self.handle, &exit_code) == 0) {
                 break :x Term{ .Unknown = 0 };
             } else {
-                break :x Term{ .Exited = exit_code };
+                break :x Term{ .Exited = @truncate(u8, exit_code) };
```
gsquire, line 327: Windows seems to handle this differently which isn't terribly surprising.

PR #8366, opened 2021-03-26
A logging allocator that uses std.log

leecannon:
As noted in `logging_allocator.zig` (below), now that we have `std.log` there should be a logging allocator that uses it.
https://github.com/ziglang/zig/blob/9f01598a498ecbd868cb81dc1efdb1507010c1b8/lib/std/heap/logging_allocator.zig#L11

This is an implementation of such a logging allocator, the only annoying thing about it is the name, as `LoggingAllocator` is already taken by the one that wraps a `writer`. 

I'm not sure if we want to go the breaking change route of renaming the current `LoggingAllocator` to something else like `LogToWriterAllocator` or if this one just needs a better name than `StdLogLoggingAllocator` and `ScopedStdLogLoggingAllocator`.

Another problem with this is tests, i can't think of a clean way to capture the output of `std.log` without overriding the functionality in the root file.


lib/std/heap/stdlog_logging_allocator.zig
```diff
@@ -0,0 +1,103 @@
...
+            allocator: *Allocator,
+            len: usize,
+            ptr_align: u29,
+            len_align: u29,
+            ra: usize,
+        ) error{OutOfMemory}![]u8 {
+            const self = @fieldParentPtr(Self, "allocator", allocator);
+            const result = self.parent_allocator.allocFn(self.parent_allocator, len, ptr_align, len_align, ra);
+            if (result) |buff| {
+                logHelper(success_log_level, "alloc: {} success!", .{len});
```
LemonBoy, line 64: The alignments for the length/ptr should be logged as they are as important as the length.
Ditto in resize.

lib/std/heap/stdlog_logging_allocator.zig
```diff
@@ -0,0 +1,103 @@
...
+        allocator: Allocator,
+        parent_allocator: *Allocator,
+
+        const Self = @This();
+
+        pub fn init(parent_allocator: *Allocator) Self {
+            return .{ .allocator = Allocator{
+                .allocFn = alloc,
+                .resizeFn = resize,
+            }, .parent_allocator = parent_allocator };
```
LemonBoy, line 37: Is this line formatted with `zig fmt`? If not please format the whole file.

lib/std/heap/stdlog_logging_allocator.zig
```diff
@@ -0,0 +1,103 @@
...
+            ptr_align: u29,
+            len_align: u29,
+            ra: usize,
+        ) error{OutOfMemory}![]u8 {
+            const self = @fieldParentPtr(Self, "allocator", allocator);
+            const result = self.parent_allocator.allocFn(self.parent_allocator, len, ptr_align, len_align, ra);
+            if (result) |buff| {
+                logHelper(success_log_level, "alloc: {} success!", .{len});
+            } else |err| {
+                logHelper(failure_log_level, "alloc: {} failure!", .{len});
```
LemonBoy, line 66: Let's print the error name in case of failure.
Ditto in `resize`.

LemonBoy (commented):
Cool stuff!
The testing harness (in `test_runner.zig`) is supposed to fail whenever an error message is logged, you can re-use the same basic tests in the old logging allocator (that can be probably removed after this is merged).

lib/std/heap/stdlog_logging_allocator.zig
leecannon: As the only possible error when dealing with an Allocator is OutOfMemory is it worth printing it?

leecannon:
I did try replicating the test from the current LoggingAllocator but as a critical log message is the expected output this results in the test suite counting it as a failed test.

Also unlike the LoggingAllocator test there is no way to validate the correct messages with the correct values are output.

lib/std/heap/stdlog_logging_allocator.zig
LemonBoy: I'd print that for completeness sake, at the moment we only have a single error in the interface but who knows how many more errors may be added in future.
Anyways not a dealbreaker :point_left: :sunglasses: :point_left:

LemonBoy:
> I did try replicating the test from the current LoggingAllocator but as a critical log message is the expected output this results in the test suite counting it as a failed test.

You have to modify a bit the tests, you want to ensure there are no failures in a allocate/resize/free and that the code compiles.
How to check the log message contents is still an open question, it'll eventually be resolved soon :tm:

lib/std/heap.zig
```diff
@@ -16,6 +16,9 @@ const maxInt = std.math.maxInt;
 
 pub const LoggingAllocator = @import("heap/logging_allocator.zig").LoggingAllocator;
 pub const loggingAllocator = @import("heap/logging_allocator.zig").loggingAllocator;
+pub const ScopedLoggingAllocator = @import("heap/logging_allocator.zig").ScopedLoggingAllocator;
+pub const LogToWriterAllocator = @import("heap/log_to_writer_allocator.zig").LogToWriterAllocator;
+pub const logToWriterAllocator = @import("heap/log_to_writer_allocator.zig").logToWriterAllocator;
```
komuw, line 21: duplicates

leecannon: These are not the same, line 20 is the type it self, line 21 is a function that takes a allocator and wraps it in the type.

The `LoggingAllocator` in master has the same thing.

leecannon:
It doesnt look like the failed CI builds on linux and windows have anything to do with this change.

leecannon:
Now it's just the windows ci being completed broken

leecannon:
I accidentally deleted the branch for this pull request and recreating it with the same name does not attach this pull request to the new one.
Closing this one and opening a new one.

PR #8371, opened 2021-03-27
std.ascii: add allocation free versions of lower/upper string

xackus:


lib/std/ascii.zig
```diff
@@ -288,36 +288,46 @@ test "ascii character classes" {
     testing.expect(isSpace(' '));
 }
 
+/// Writes a lower case copy of `ascii_string` to `output`.
+pub fn lowerString(ascii_string: []const u8, output: []u8) []u8 {
+    for (output) |*c, i| {
+        c.* = toLower(ascii_string[i]);
```
LemonBoy, line 294: This is dangerous when `ascii_string` is shorter than `output`.

xackus: Thanks. The dangers of copy-paste.

lib/std/ascii.zig
```diff
@@ -288,36 +288,58 @@ test "ascii character classes" {
     testing.expect(isSpace(' '));
 }
 
+/// Writes a lower case copy of `ascii_string` to `output`.
+pub fn lowerString(ascii_string: []const u8, output: []u8) []u8 {
+    for (ascii_string) |c, i| {
```
LemonBoy, line 293: And now it's dangerous if `output` is shorter than `ascii_string` :^)
If you don't want to return an error here the function should operate on `min(ascii_string.len, output.len)`.

xackus: So is `mem.copy`. Though I should have documented this and added an explicit assert to make it clearer.
I don't like the idea of silently truncating the output.
Returning something like `error.BufferTooSmall` and letting the user write `catch unreachable` if they want would be safer, but more verbose. I wonder what other people think about this.

LemonBoy: Well everything is better than overwriting chunks of memory, returning an error here is a good design choice.

ifreund: > Well everything is better than overwriting chunks of memory, returning an error here is a good design choice.

I disagree, I think this should match the api of std.mem.copy. Reading/writing outside the bounds of a slice is check in safe build modes.

LemonBoy: I don't like assertions because you can't catch them in "safe" modes (poof, your application is terminated) and they do jack shit in "release" modes (poof, buffer overflow).
For the sake of keeping the API as uniform as possible adding an assertion and a doc comment is fine.

ifreund: The point of assertions is to catch bugs. Bugs are different from errors, errors are preventable and, I agree, should never cause a panic. Disregarding the doc comment of a function like std.mem.copy would be a bug, for example by passing a slice of unknown length:
```zig
/// dest.len must be >= source.len.
/// If the slices overlap, dest.ptr must be <= src.ptr.
```
If you want to handle the case in which the dest length is shorter than the source length, just add an if statement checking that condition before calling the function. A quick grep of the std however shows that this is almost never necessary (I didn't find a case yet). I think if std.mem.copy or this function were made to return an error this would do little to improve safety and instead result in well-justified `catch unreachable` everywhere.

andrewrk:
Heads up, prior to merging, I swapped the order of dest and src parameters to match the convention set by std.mem.copy.

Issue #7865, opened 2021-01-23
Zeroes array of extern struct with (optional) c_void pointer field crashes compiler.


JesseRMeyer:
```zig
pub const dudibly = extern struct {
    tada: ?*c_void,
};

const std = @import("std");
const zeroes = std.mem.zeroes;

pub fn main() void {
    var queues_create_infos = zeroes([1]dudibly);
}
```

Produces via build-exe
```
TODO buf_write_value_bytes maybe typethread 4372 panic:
Unable to dump stack trace: debug info stripped
```

When the optional specifier is removed from the `tada` field, the resulting error changes to:
```
Unreachable at D:\a\1\s\src\stage1\ir.cpp:30397 in buf_write_value_bytes. This is a bug in the Zig compiler.thread 1564 panic:
Unable to dump stack trace: debug info stripped
```

Zig version: 0.8.0-dev.1039+bea791b63

haberman:
I ran into this too. Here is another repro that doesn't use the standard library:

```zig
const Foo = extern struct {
    field: ?*u8 = null,
};

fn zeroes(comptime T: type) T {
    var ret: Foo = undefined;      // Also fails with "var ret = Foo{};"
    @ptrCast(*usize, &ret).* = 0;  // Succeeds if you remove this.
    return ret;
}

export const foo = zeroes(Foo);
```

Here is a stack trace with a debug build. I also added a `val->dump()` right before the `zig_unreachable()`:

```
Value@0x5629cee8b338(*u8)<!invalid ptr!>
Unreachable at ../src/stage1/ir.cpp:30408 in buf_write_value_bytes. This is a bug in the Zig compiler.thread 444991 panic: 
/home/haberman/code/zig/src/stage1.zig:171:5: 0x5629ca9a595c in stage2_panic (zig1)
    @panic(ptr[0..len]);
    ^
../src/stage1/util.cpp:20:17: 0x5629cb43d2a2 in zig_panic (../src/stage1/util.cpp)
    stage2_panic("", 0);
                ^
../src/stage1/ir.cpp:30408:17: 0x5629cb500e96 in buf_write_value_bytes (../src/stage1/ir.cpp)
                zig_unreachable();
                ^
../src/stage1/ir.cpp:30426:46: 0x5629cb501019 in buf_write_value_bytes (../src/stage1/ir.cpp)
                        buf_write_value_bytes(codegen, buf + offset, field_val);
                                             ^
[...]

PR #8372, opened 2021-03-27
Emit null or uninitialized pointers as 0 values.

haberman:
This fixes a bug where IR analysis would fail when trying to
overwrite a null or uninitialized pointer. In particular this
made std.mem.zeroes() fail on any struct that contained a pointer.

Fixes one half of https://github.com/ziglang/zig/issues/7865 (the case where the pointer is not optional).

### DIFF
--- a/src/stage1/ir.cpp
+++ b/src/stage1/ir.cpp
@@ -30396,15 +30396,21 @@ static void buf_write_value_bytes(CodeGen *codegen, uint8_t *buf, ZigValue *val)
         case ZigTypeIdFloat:
             float_write_ieee597(val, buf, codegen->is_big_endian);
             return;
-        case ZigTypeIdPointer:
-            if (val->data.x_ptr.special == ConstPtrSpecialHardCodedAddr) {
-                BigInt bn;
-                bigint_init_unsigned(&bn, val->data.x_ptr.data.hard_coded_addr.addr);
-                bigint_write_twos_complement(&bn, buf, codegen->builtin_types.entry_usize->data.integral.bit_count, codegen->is_big_endian);
-                return;
-            } else {
-                zig_unreachable();
+        case ZigTypeIdPointer: {
+            BigInt bn;
+            switch (val->data.x_ptr.special) {
+                case ConstPtrSpecialHardCodedAddr:
+                    bigint_init_unsigned(&bn, val->data.x_ptr.data.hard_coded_addr.addr);
+                    break;
+                case ConstPtrSpecialNull:
+                case ConstPtrSpecialInvalid:
+                    bigint_init_unsigned(&bn, 0);
+                    break;
+                default:
+                    zig_unreachable();
             }
+            return bigint_write_twos_complement(&bn, buf, codegen->builtin_types.entry_usize->data.integral.bit_count, codegen->is_big_endian);
+        }
         case ZigTypeIdArray:
             return buf_write_value_bytes_array(codegen, buf, val, val->type->data.array.len);
         case ZigTypeIdVector:


src/stage1/ir.cpp
```diff
@@ -30396,15 +30396,21 @@ static void buf_write_value_bytes(CodeGen *codegen, uint8_t *buf, ZigValue *val)
...
-            } else {
-                zig_unreachable();
+        case ZigTypeIdPointer: {
+            BigInt bn;
+            switch (val->data.x_ptr.special) {
+                case ConstPtrSpecialHardCodedAddr:
+                    bigint_init_unsigned(&bn, val->data.x_ptr.data.hard_coded_addr.addr);
+                    break;
+                case ConstPtrSpecialNull:
+                case ConstPtrSpecialInvalid:
```
xackus, line 30406:
```suggestion
```
`ConstPtrSpecialInvalid` is there so that zero-initialized variables have an invalid value.
https://github.com/ziglang/zig/blob/9f01598a498ecbd868cb81dc1efdb1507010c1b8/src/stage1/all_types.hpp#L238-L239
Undefined is signalled by `ConstValSpecialUndef`.

haberman: Hmm, the following program fails if I remove this line (and it is not fixed by adding `case ConstValSpecialUndef`):

```zig
fn zero() [*c]u8 {
    var ret: [*c]u8 = undefined;
    @ptrCast(*[8]u8, &ret)[1] = 123;  // Succeeds if you remove this.
    return ret;
}

export const foo = zero();
```

I wonder if there is a bug somewhere else in the compiler where `ConstValSpecialUndef` is not getting set appropriately.

xackus: I should have been clearer. `ConstValSpecialUndef` belongs to a different enum (notice Val instead of Ptr). See `ZigValue.special`.

xackus: Also keep in mind that the plan is to eventually throw away the C++ code: #6378.
To fix some of the bugs you would need to rewrite quite a bit of code.

haberman: > I should have been clearer. ConstValSpecialUndef belongs to a different enum (notice Val instead of Ptr).

I see. But it looks like `ConstPtrValue.special` is the discriminator that must be valid before we can read any other member of `ConstPtrValue.data`, no? It seems like we have to read `ConstPtrValue.data` before we can test it against `ConstValSpecialUndef`, but we can't do that without a valid discriminator. So it still seems like a bug that `val->data.x_ptr.special == ConstPtrSpecialInvalid`.

> To fix some of the bugs you would need to rewrite quite a bit of code.

Hmm, that's too bad. I guess it's throwaway work to fix bugs here if everything is going to be rewritten in Zig?

haberman:
What is the status of this PR?

The PR as written does seem to fix a bug, and I'm not aware of it introducing any other bugs.

It is perhaps working around a different bug, in which an undefined value has `val->data.x_ptr.special == ConstPtrSpecialInvalid` instead of `val->data.x_ptr.special == <something else> && val->data.x_ptr.data.<something else>->special == ConstValSpecialUndef`.

What is the way forward here?

1. Accept the PR even though it's working around a different bug?
2. Accept the PR once it fixes the other bug?
3. Consider the bug unfixable until stage1 is rewritten in Zig?

It would be a bummer to have to wait for a Zig rewrite here. This bug breaks a lot of use cases where you might want to call `std.mem.zeroes(T)`.

LemonBoy:
This is not the correct way of fixing the problem.
Have a look at the very first few lines of `buf_write_value_bytes`, every single `undefined` value is expanded there into (invalid most of the times) `ZigValue` that's then serialized, tripping that assertion in your case.

Fixing it properly is going to be quite hard but I think the problem could be worked around by:
- Removing the undefined expansion logic
- Adding some code to write `abi_size` zeros (or `0xAA`) to `buf` whenever an undefined value is serialized

Vexu:
I'm going to close this to clear up some older prs as this isn't the correct fix but if you do want to try LemonBoy's suggestion feel free to open a new pr.
Alternatively if you can find a workaround, the current estimate for the self hosted compiler is before the 0.9.0 release, which shoud also fix this.

PR #8375, opened 2021-03-28
doc: namespaced global variable - move to top level

rofrol:


FireFox317:
Why this change?

Vexu:
The whole point of that test is to demonstrate that global variables can be declared inside a function.

rofrol:
I don't get it. From documenation:

>Global variables are considered to be a top level declaration

https://ziglang.org/documentation/master/#Global-Variables

So how variable declared inside function is considered to be a global variable?

Vexu:
All variables declared inside containers are globals, it says so right above the test:
> Global variables may be declared inside a struct, union, or enum:

rofrol:
Still I don't get it:

>In computer programming, a global variable is a variable with global scope, meaning that it is visible (hence accessible) throughout the program, unless shadowed.
https://en.wikipedia.org/wiki/Global_variable

```zig
const print = @import("std").debug.print;

pub fn main() void {
    print("S.x: {}\n", .{S.x});
}

fn foo() void {
    const S = struct {
        var x: i32 = 1234;
    };
}
```

```bash
zig build-exe src/global.zig 
.\src\global.zig:4:26: error: use of undeclared identifier 'S'
    print("S.x: {}\n", .{S.x});
```

g-w1:
Please use one of the communities. https://github.com/ziglang/zig/wiki/Community It is easier to explain something to someone there than on github.

Vexu:
Global here means that the variable has static lifetime, not that it can be accessed anywhere.

rofrol:
So why call it global in documentation instead static etc.?

rofrol:
>In computer programming, a static variable is a variable that has been allocated "statically", meaning that its lifetime (or "extent") is the entire run of the program. https://en.wikipedia.org/wiki/Static_variable

Maybe I should modify my pull request to show two examples:
- one with global variable in top-level struct
- static lifetime variable not in top-level struct
- ?

rofrol:
I have found this:

>Both variables are "global". The both have the static storage duration. The difference is that the first has external linkage and the second has internal linkage that is it is invisible outside the compilation unit where it is declared. If the declaration of a variable with internal linkage is included in several compilation units then each of them has its own unique variable with this name.

https://stackoverflow.com/questions/44954524/difference-between-global-and-static-variable-in-c/44954561#44954561

rofrol:
I think this clarifies it:

>All global variables are static, but all static variables are not necessarily global.
>
>What all static variables (including global) have in common is that they have a lifetime that lasts the duration of the program. They are initialized just once, when the program begins.
>
>Local variables—which are “automatic” by default—are created each time a function is called, no matter how many times that function is on the stack… therefore they are appropriate for functions called recursively or functions involved with multiple threads. A local automatic variable is initialized each and every time the function is started.
>
>Local variables may also be declared static, in which case they work like global variables but have visibility limited to the function in which declared. Remember that a local static is like a global in that it is only initialized once

https://www.quora.com/What-is-the-difference-between-a-static-variable-and-global-variable-in-C-C++/answer/Brian-Overland-1

Issue #786, opened 2018-02-24
allow configuration of error set tag type via a public const in the root source file


andrewrk:
Right now error sets are hardcoded to codegen to a `u16`.

Ideally, zig would choose an appropriate sized integer based on the total number of error values in the compilation, but the user may do `@sizeOf(SomeStruct)` which has an error set type as a field before the entire compilation finishes analysis. This would create a circular dependency.

One idea is to have the error set integer type have a default of `u32`, but in the same way that you can override the default `panic` handler in the root source file, you can override the error set integer type in the root source file by doing:

```zig
pub const ErrorIntegerType = u8;
```

Then you'd get a compile error if zig ran out of bits when choosing error values.

daurnimator:
## Further proposal:

For performance reasons it would be nice if `errno` values mapped to their same number in the error set, e.g. `EPERM` -> `1`, that way most `errno` processing code doesn't need a lookup table.

What if the ErrorSet type could be declared in the root file? Lets make the restriction that it *must* be a non-exhaustive enum. 

The most straightforward setup:
```
pub const ErrorSetType = enum(u32) { _ };
```

Mapping errno:
```
pub const ErrorSetType = enum(u32) {
  PermissionDenied = std.os.EPERM,
  FileNotFound = std.os.ENOENT,
  ProcessNotFound = std.os.ESRCH,
  // etc.
  _,
};
```

g-w1:
What should be the max size of an error specified by the user? as @MasterQ32 pointed out, even u32 requires a LOT (34 gb) of source code to use up fully. I think u32 is a good limit. https://freenode.irclog.whitequark.org/zig/2021-02-23#1614089709-1614090346;

lemaitre:
I don't know if the language can take advantage of this, but I think `u31` could be better as it would make it possible for `anyerror!u31` to be 4 byte large.

PaperPrototype:
I think we kinda create a problem by allowing multiple errors with the same name. 

We could instead force user's to set the uint size of the error, and then let them handle the casting between different error types. It would be more verbose (also more C like), and I feel it would help reduce ambiguity.

PaperPrototype:
Since errors are just enums you could force users to do this
```zig
const Value = error(u32) {
    zero,
    one,
    two,
};
```

And if they don't it can default to `u16`. It's simple and possible since you don't have to combine errors with the same name and assign them the same number (doing that was cool, but I feel unecessary)

[...]

PR #8376, opened 2021-03-28
stage2: implement @errorToInt and @intToError

g-w1:
I could not test @errorToInt because we do not have error set equality yet (I realized this out after I implemented it :().

There is another big problem:
`error`s do *not* get removed from `Module` when they are removed from source code in an incremental update. This means that this test case fails: 
```zig
    {
        var case = ctx.exeFromCompiledC("@intToError", .{});

        case.addCompareOutput(
            \\pub export fn main() c_int {
            \\    // comptime checks
            \\    const a = error.A;
            \\    const b = error.B;
            \\    const c = @intToError(2);
            \\    const d = @intToError(1);
            \\    if (!(c == b)) unreachable;
            \\    if (!(a == d)) unreachable;
            \\    // runtime checks
            \\    var x = error.A;
            \\    var y = error.B;
            \\    var z = @intToError(2);
            \\    var f = @intToError(1);
            \\    if (!(y == z)) unreachable;
            \\    if (!(x == f)) unreachable;
            \\    return 0;
            \\}
        , "");
        case.addError(
            \\pub export fn main() c_int {
            \\    const c = @intToError(0);
            \\    return 0;
            \\}
        , &.{":2:26: error: integer value 0 represents no error"});
        case.addError(
            \\pub export fn main() c_int {
            \\    const c = @intToError(2);
            \\    return 0;
            \\}
        , &.{":2:26: error: integer value 2 represents no error"});
    }
```
Because integer value 2 does represent an error, just in a past update. Not sure what we should do about this yet ...
Also not sure if there is a better way to represent no-ops in the cbe than just making a local and returning it?

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -1237,6 +1237,8 @@ fn blockExprStmts(
                         .bit_not,
                         .error_set,
                         .error_value,
+                        .error_to_int,
+                        .int_to_error,
                         .slice_start,
                         .slice_end,
                         .slice_sentinel,
@@ -3370,6 +3372,16 @@ fn builtinCall(
             const result = try gz.addUnNode(.import, target, node);
             return rvalue(gz, scope, rl, result, node);
         },
+        .error_to_int => {
+            const target = try expr(gz, scope, .none, params[0]);
+            const result = try gz.addUnNode(.error_to_int, target, node);
+            return rvalue(gz, scope, rl, result, node);
+        },
+        .int_to_error => {
+            const target = try expr(gz, scope, .{ .ty = .u16_type }, params[0]);
+            const result = try gz.addUnNode(.int_to_error, target, node);
+            return rvalue(gz, scope, rl, result, node);
+        },
         .compile_error => {
             const target = try expr(gz, scope, .none, params[0]);
             const result = try gz.addUnNode(.compile_error, target, node);
@@ -3439,7 +3451,6 @@ fn builtinCall(
         .enum_to_int,
         .error_name,
         .error_return_trace,
-        .error_to_int,
         .err_set_cast,
         .@"export",
         .fence,
@@ -3448,7 +3459,6 @@ fn builtinCall(
         .has_decl,
         .has_field,
         .int_to_enum,
-        .int_to_error,
         .int_to_float,
         .int_to_ptr,
         .memcpy,
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -80,6 +80,9 @@ deletion_set: ArrayListUnmanaged(*Decl) = .{},
 /// Error tags and their values, tag names are duped with mod.gpa.
 global_error_set: std.StringHashMapUnmanaged(u16) = .{},
 
+/// error u16 -> []const u8 for fast lookups for @intToError at comptime
+error_name_list: ArrayListUnmanaged([]const u8) = .{},
+
 /// Keys are fully qualified paths
 import_table: std.StringArrayHashMapUnmanaged(*Scope.File) = .{},
 
@@ -1570,7 +1573,22 @@ pub const SrcLoc = struct {
                 const token_starts = tree.tokens.items(.start);
                 return token_starts[tok_index];
             },
-            .node_offset_builtin_call_arg0 => @panic("TODO"),
+            .node_offset_builtin_call_arg0 => |node_off| {
+                const decl = src_loc.container.decl;
+                const tree = decl.container.file_scope.base.tree();
+                const node_datas = tree.nodes.items(.data);
+                const node_tags = tree.nodes.items(.tag);
+                const node = decl.relativeToNodeIndex(node_off);
+                const param = switch (node_tags[node]) {
+                    .builtin_call_two, .builtin_call_two_comma => node_datas[node].lhs,
+                    .builtin_call, .builtin_call_comma => tree.extra_data[node_datas[node].lhs],
+                    else => unreachable,
+                };
+                const main_tokens = tree.nodes.items(.main_token);
+                const tok_index = main_tokens[param];
+                const token_starts = tree.tokens.items(.start);
+                return token_starts[tok_index];
+            },
             .node_offset_builtin_call_arg1 => @panic("TODO"),
             .node_offset_builtin_call_argn => unreachable, // Handled specially in `Sema`.
             .node_offset_array_access_index => @panic("TODO"),
@@ -1892,6 +1910,8 @@ pub fn deinit(mod: *Module) void {
         gpa.free(entry.key);
     }
     mod.global_error_set.deinit(gpa);
+
+    mod.error_name_list.deinit(gpa);
 
     for (mod.import_table.items()) |entry| {
         entry.value.destroy(gpa);
@@ -3346,10 +3366,12 @@ pub fn getErrorValue(mod: *Module, name: []const u8) !std.StringHashMapUnmanaged
     const gop = try mod.global_error_set.getOrPut(mod.gpa, name);
     if (gop.found_existing)
         return gop.entry.*;
-    errdefer mod.global_error_set.removeAssertDiscard(name);
 
+    errdefer mod.global_error_set.removeAssertDiscard(name);
+    try mod.error_name_list.ensureCapacity(mod.gpa, mod.error_name_list.items.len + 1);
     gop.entry.key = try mod.gpa.dupe(u8, name);
-    gop.entry.value = @intCast(u16, mod.global_error_set.count() - 1);
+    gop.entry.value = @intCast(u16, mod.error_name_list.items.len);
+    mod.error_name_list.appendAssumeCapacity(gop.entry.key);
     return gop.entry.*;
 }
 
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -177,6 +177,8 @@ pub fn analyzeBody(
             .error_set => try sema.zirErrorSet(block, inst),
             .error_union_type => try sema.zirErrorUnionType(block, inst),
             .error_value => try sema.zirErrorValue(block, inst),
+            .error_to_int => try sema.zirErrorToInt(block, inst),
+            .int_to_error => try sema.zirIntToError(block, inst),
             .field_ptr => try sema.zirFieldPtr(block, inst),
             .field_ptr_named => try sema.zirFieldPtrNamed(block, inst),
             .field_val => try sema.zirFieldVal(block, inst),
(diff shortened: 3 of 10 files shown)


andrewrk (changes requested):
As far as the incremental update goes, I have a solution to that: change the language to make this a feature, not a bug :)

Seriously- I will propose to change the language in this way (cc @SpexGuy):

`@intToError`: If the integer parameter passed to this function does not match a value which came from `@errorToInt`, the implementation must do one of two things:
 * Invoke detectable illegal behavior, or
 * Succeed; the error name is implementation-defined but must be unique from all other error names. If the same error name is encountered in the same compilation, it must correspond to the same integer value.

This is what we want for hot code swapping, since code could be running that is using old error code values that no longer exist in updated code. This language change will (clearly) also make the implementation of zig compilers simpler.

SpexGuy:
> change the language to make this a feature, not a bug

AFAIK error code mappings are never guaranteed, so there's no language change needed to allow them to be a sparse mapping.  However, they do have an ultimate limit of 65k, which may not be a completely unreasonable number if they are never cleaned up.  Other pooled resources like identifier names and cached comptime constants or function results will also need to be cleaned up over time and recompilations, so it's not like this is a unique property of errors.  This seems like another problem requirement of the domain.  If you want to make the compiler faster, you could have it do the cleanup after emitting the new binary, so it's working in the downtime between compilations.  But I don't expect this will be a very expensive operation.

src/Module.zig
```diff
@@ -3408,7 +3428,8 @@ pub fn getErrorValue(mod: *Module, name: []const u8) !std.StringHashMapUnmanaged
     errdefer mod.global_error_set.removeAssertDiscard(name);
 
     gop.entry.key = try mod.gpa.dupe(u8, name);
```
andrewrk, line 3430: In order to not leak on allocation failure, this resource needs an errdefer, or better yet, `inv_error_map.ensureCapacity` before it, and then `inv_error_map.appendAssumeCapacity` below.

src/Module.zig
```diff
@@ -80,6 +80,9 @@ deletion_set: ArrayListUnmanaged(*Decl) = .{},
 /// Error tags and their values, tag names are duped with mod.gpa.
 global_error_set: std.StringHashMapUnmanaged(u16) = .{},
 
+/// error u16 -> []const u8 for fast lookups for @intToError at comptime
+inv_error_map: ArrayListUnmanaged([]const u8) = .{},
```
andrewrk, line 84: name suggestion:

```suggestion
error_name_list: ArrayListUnmanaged([]const u8) = .{},
```

they're both inverted from the other one's perspective - I think this name communicates better what it is

src/Module.zig
```diff
@@ -3408,7 +3428,8 @@ pub fn getErrorValue(mod: *Module, name: []const u8) !std.StringHashMapUnmanaged
     errdefer mod.global_error_set.removeAssertDiscard(name);
 
     gop.entry.key = try mod.gpa.dupe(u8, name);
-    gop.entry.value = @intCast(u16, mod.global_error_set.count() - 1);
+    gop.entry.value = @intCast(u16, mod.global_error_set.count());
```
andrewrk, line 3431: This is supposed to be tied to `inv_error_map` right? should be 

```suggestion
    gop.entry.value = @intCast(u16, mod.inv_error_map.items.len);
```

for clarity

src/Sema.zig
```diff
@@ -1460,6 +1462,65 @@ fn zirErrorValue(sema: *Sema, block: *Scope.Block, inst: zir.Inst.Index) InnerEr
...
+    const op = try sema.resolveInst(inst_data.operand);
+
+    if (op.value() != null) {
+        const int = try sema.resolveInt(block, operand_src, inst_data.operand, Type.initTag(.u16));
+        if (int > sema.mod.global_error_set.count() or int == 0)
+            return sema.mod.fail(&block.base, operand_src, "integer value {d} represents no error", .{int});
+        const payload = try sema.arena.create(Value.Payload.Error);
+        payload.* = .{
+            .base = .{ .tag = .@"error" },
+            .data = .{ .name = sema.mod.inv_error_map.items[int - 1] },
```
andrewrk, line 1508: Ah I see, so there's a -1 here to account for the array list being off by one. Can I suggest a slightly modified strategy here? In stage1 we just add a 0 => "(no error)" mapping on startup to this array before analyzing any code. I think that's the way to go here. Then the array list mapping will be matched up exactly, no off by one errors to trip over.

andrewrk (changes requested):
almost there

andrewrk:
> Other pooled resources like identifier names and cached comptime constants or function results will also need to be cleaned up over time and recompilations, so it's not like this is a unique property of errors.

There is an important difference here - identifier names and cached comptime constants can be owned by the Decl in which they are used. However the mappings of error names to values must be shared globally by all Decls.

> AFAIK error code mappings are never guaranteed, so there's no language change needed to allow them to be a sparse mapping.

Yeah that's right, but the lang ref [says](https://ziglang.org/documentation/master/#intToError):

>  Attempting to convert an integer that does not correspond to any error results in safety-protected Undefined Behavior. 

What we want to be able to do in the implementation is to allow an integer value that in fact does not (any longer) correspond to any error name, *succeed* in the call to `@errorToInt` and return an identifier name that does not exist (anymore) in the user's entire codebase.

> However, they do have an ultimate limit of 65k

Note this accepted issue:  #786

src/Sema.zig
```diff
@@ -1460,6 +1462,65 @@ fn zirErrorValue(sema: *Sema, block: *Scope.Block, inst: zir.Inst.Index) InnerEr
...
+    defer tracy.end();
+
+    const inst_data = sema.code.instructions.items(.data)[inst].un_node;
+    const src = inst_data.src();
+    const operand_src: LazySrcLoc = .{ .node_offset_builtin_call_arg0 = inst_data.src_node };
+
+    const op = try sema.resolveInst(inst_data.operand);
+
+    if (op.value() != null) {
+        const int = try sema.resolveInt(block, operand_src, inst_data.operand, Type.initTag(.u16));
```
Vexu, line 1502: A cast to `u16` should also happen if the value is not comptime.

andrewrk, line 1502: Sorry didn't catch this earlier - avoid resolving the same operand twice - it's already resolved on line 1499. The smell is that we're not capturing the optional value of `op.value()` here. Have a look at what resolveInt does internally, and do that inside this block.

src/AstGen.zig
```diff
@@ -3394,6 +3396,16 @@ fn builtinCall(
             const result = try gz.addUnNode(.import, target, node);
             return rvalue(mod, scope, rl, result, node);
         },
+        .error_to_int => {
+            const target = try expr(mod, scope, .none, params[0]);
+            const result = try gz.addUnNode(.error_to_int, target, node);
+            return rvalue(mod, scope, rl, result, node);
+        },
+        .int_to_error => {
+            const target = try expr(mod, scope, .none, params[0]);
```
andrewrk, line 3405: Here pass a `.{.ty = .u16_type }` result location instead of `.none` and then actually remove the `coerce` call inside Sema

cc @Vexu

andrewrk (changes requested):
found some more stuff

src/Sema.zig
```diff
@@ -1460,6 +1462,65 @@ fn zirErrorValue(sema: *Sema, block: *Scope.Block, inst: zir.Inst.Index) InnerEr
...
+    const tracy = trace(@src());
+    defer tracy.end();
+
+    const inst_data = sema.code.instructions.items(.data)[inst].un_node;
+    const src = inst_data.src();
+    const operand_src: LazySrcLoc = .{ .node_offset_builtin_call_arg0 = inst_data.src_node };
+
+    const op = try sema.resolveInst(inst_data.operand);
+
+    if (op.value()) |value| {
```
Vexu, line 1501: Use `resolveDefinedValue` here and in `zirErrorToInt` to match stage1 and not panic with undefined.

Vexu: Searching for `.value())` reveals tons of other places that should also use this.

g-w1: Do you want me to fix this in other places?

src/Sema.zig
```diff
@@ -1460,6 +1462,65 @@ fn zirErrorValue(sema: *Sema, block: *Scope.Block, inst: zir.Inst.Index) InnerEr
...
 
+fn zirErrorToInt(sema: *Sema, block: *Scope.Block, inst: zir.Inst.Index) InnerError!*Inst {
+    const tracy = trace(@src());
+    defer tracy.end();
+
+    const inst_data = sema.code.instructions.items(.data)[inst].un_node;
+    const src = inst_data.src();
+    const operand_src: LazySrcLoc = .{ .node_offset_builtin_call_arg0 = inst_data.src_node };
+    const op = try sema.resolveInst(inst_data.operand);
+    const op_coerced = try sema.coerce(block, Type.initTag(.anyerror), op, operand_src);
```
andrewrk, line 1473: Check lang ref, this should support more types: 
https://ziglang.org/documentation/master/#errorToInt

src/codegen.zig
```diff
@@ -2557,6 +2559,18 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
             return self.fail(inst.base.src, "TODO load the operand and call genIsErr", .{});
         }
 
+        fn genErrorToInt(self: *Self, inst: *ir.Inst.UnOp) !MCValue {
+            switch (arch) {
+                else => return self.fail(inst.base.src, "TODO implement errorToInt for {}", .{self.target.cpu.arch}),
+            }
```
andrewrk, line 2565:
```suggestion
            return self.resolveInst(inst.operand);
```

src/codegen.zig
```diff
@@ -2557,6 +2559,18 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
...
+        }
+
+        fn genIntToError(self: *Self, inst: *ir.Inst.UnOp) !MCValue {
+            switch (arch) {
+                else => return self.fail(inst.base.src, "TODO implement intToError for {}", .{self.target.cpu.arch}),
+            }
```
andrewrk, line 2571:
```suggestion
            return self.resolveInst(inst.operand);
```

src/Sema.zig
g-w1: https://freenode.irclog.whitequark.org/zig/2021-03-29#1616980514-1616980596;

PR #8377, opened 2021-03-28
Add a test & fix for std.fmt.fmtDuration

jdknezek:
Minor fix for `fmtDuration`. Fixes e.g. 145000ns being printed as "145.us" by not printing any of the decimal portion if it's zero.


ifreund (approved):
LGTM, Thanks!

PR #8378, opened 2021-03-28
array_hash_map: decrement entries slice len after popping from entries in pop() to prevent oob

lithdew:
The following test fails with an out-of-bounds exception. The exception is triggered when an array hash map's capacity is expanded multiple times, subsequently followed by a pop(). The subsequent pop() attempts to access the map's entries slice which is shrunk by pop() calling entries.pop(). The fix is to delay the decrement of the entries slice length to be at the end of pop().

```zig
test "pop" {
    var map = AutoArrayHashMap(i32, i32).init(std.testing.allocator);
    defer map.deinit();

    // Insert just enough entries so that the map expands. Afterwards,
    // pop all entries out of the map.

    var i: i32 = 0;
    while (i < 9) : (i += 1) {
        testing.expect((try map.fetchPut(i, i)) == null);
    }

    while (i > 0) : (i -= 1) {
        const pop = map.pop();
        testing.expect(pop.key == i - 1 and pop.value == i - 1);
    }
}
```

I updated the "pop" test in `array_hash_map.zig` to take into account this out-of-bounds exception.

@tetsuo-cpp @gruebite

### DIFF
--- a/lib/std/array_hash_map.zig
+++ b/lib/std/array_hash_map.zig
@@ -687,8 +687,9 @@ pub fn ArrayHashMapUnmanaged(
 
         /// Removes the last inserted `Entry` in the hash map and returns it.
         pub fn pop(self: *Self) Entry {
-            const top = self.entries.pop();
+            const top = self.entries.items[self.entries.items.len - 1];
             _ = self.removeWithHash(top.key, top.hash, .index_only);
+            self.entries.items.len -= 1;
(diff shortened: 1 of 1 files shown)


andrewrk (approved):

kubkon:
Awesome! I actually hit that very issue recently while reworking the linker. Thanks @lithdew!

PR #8380, opened 2021-03-28
Fix typo in Pcg.zig's fill function

mattbork:
When filling the last (len % 4) bytes of a buffer, the random number n was only being shifted right by 4 bits for each byte instead of 8. A random u16, for example, would always have its middle two nybbles be equal when generated this way. For comparison, Isaac64.zig, Sfc64.zig, and Xoroshiro128.zig all correctly shift right by 8 bits for each of the last bytes in their nearly identical fill functions.

### DIFF
--- a/lib/std/rand/Isaac64.zig
+++ b/lib/std/rand/Isaac64.zig
@@ -208,3 +208,35 @@ test "isaac64 sequence" {
         std.testing.expect(s == r.next());
     }
 }
+
+test "isaac64 fill" {
+    var r = Isaac64.init(0);
+
+    // from reference implementation
+    const seq = [_]u64{
+        0xf67dfba498e4937c,
+        0x84a5066a9204f380,
+        0xfee34bd5f5514dbb,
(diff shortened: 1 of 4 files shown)


jedisct1:
Err.... good catch... that function seems to do a byte by byte copy, so a 4 bit shift doesn't make sense.

ifreund:
It would be nice to have a test case reproducing the incorrect behavior so that this doesn't regress.

andrewrk:
> It would be nice to have a test case reproducing the incorrect behavior so that this doesn't regress.

@mattbork what do you think? I'd merge this as is, but are you interested in polishing it off with test coverage first?

mattbork:
I've added basic tests for the fill functions of the four PRNGs that already had sequence tests. They fill a buffers one byte smaller than the generator's result size and compare those to the known sequence.

andrewrk:
Thanks!

andrewrk (approved):

Issue #7702, opened 2021-01-06
Generic SIMD types and operations are not a substitute for intrinsics


lemaitre:
First, I think having generic SIMD types like `@Vector(T, N)` (#903 or any other syntax) with most arithmetic operations defined on them is really nice and is useful to many people.

However, this will never give all the power of the intrinsics because the generic interface will never be able to cover all instructions from all SIMD ISAs (even with tradeoffs). Vendors will always be creative and invent new instructions to cover specialized work cases that don't necessarily match what other vendors do.
Plus, even when an operation exists in multiple vendor ISAs and is emulatable in the others, the exact semantic might differ and will lead to tradeoffs that would penalize people who want maximal performance (main goal of SIMD).

Examples (really far to be exhaustive):

- incompatibilities of 1/sqrt(x):

| ISA | intrinsics | precision |
|------|--------------|-------|
| SSE | `_mm_rsqrt_ps` | 1.5p-12 |
| KNCNI | `_mm512_rsqrt23_ps` | 1p-23 |
| AVX512F | `_mm512_rsqrt14_ps` | 1p-14 |
| AVX512ER | `_mm512_rsqrt28_ps` | 1p-24 |
| Neon | `vrsqrteq_f32` | 1p-8 |
| Altivec | `vec_rsqrte` | 1p-12 |
| VSX | `vec_rsqrte` | 1p-14 |

- incompatibilities of fused c - a*b (all compatible with IEEE754, but can give different results with NaNs, and infinities):

| ISA | intrinsics | meaning |
|-------|-------------|-------------|
| AVX2 | `_mm256_fnmadd_ps(a, b, c)` | (-(a * b)) + c |
| Neon | `vfmsq_f32(c, a, b)` | ((-a) * b) + c |
| Altivec | `vec_nmsub(a, b, c)` | -((a * b) + (-c)) |

- Unique operations, hard/slow to emulate:

| ISA | intrinsics | meaning |
|------|--------------|-------------|
| AVX512F, KNCNI | `_mm512_add_round_ps` | perform addition with specified rounding mode (without overhead) |
| AVX512CD | `_mm512_conflict_epi32` | compute a vector of "masks" indicating lanes having the sames values |
| SVE2 | `svbdep` | expand bits of first input to posdition indicated by bits of second input, lane-wise |

- any VLA SIMD ISAs like SVE or Risc-V Vector extension.

[...]

PR #8389, opened 2021-03-29
translate-c: Add support for vector expressions

ehaas:
Includes vector types, __builtin_shufflevector, and __builtin_convertvector

### DIFF
--- a/lib/std/meta.zig
+++ b/lib/std/meta.zig
@@ -1297,3 +1297,35 @@ pub fn globalOption(comptime name: []const u8, comptime T: type) ?T {
         return null;
     return @as(T, @field(root, name));
 }
+
+/// This function is for translate-c and is not intended for general use.
+/// Convert from clang __builtin_shufflevector index to Zig @shuffle index
+/// clang requires __builtin_shufflevector index arguments to be integer constants.
+/// negative values for `this_index` indicate "don't care" so we arbitrarily choose 0
+/// clang enforces that `this_index` is less than the total number of vector elements
+/// See https://ziglang.org/documentation/master/#shuffle
+/// See https://clang.llvm.org/docs/LanguageExtensions.html#langext-builtin-shufflevector
+pub fn shuffleVectorIndex(comptime this_index: c_int, comptime source_vector_len: usize) i32 {
+    if (this_index <= 0) return 0;
+
+    const positive_index = @intCast(usize, this_index);
+    if (positive_index < source_vector_len) return @intCast(i32, this_index);
+    const b_index = positive_index - source_vector_len;
+    return ~@intCast(i32, b_index);
+}
+
+test "shuffleVectorIndex" {
+    const vector_len: usize = 4;
+
+    testing.expect(shuffleVectorIndex(-1, vector_len) == 0);
+
+    testing.expect(shuffleVectorIndex(0, vector_len) == 0);
+    testing.expect(shuffleVectorIndex(1, vector_len) == 1);
+    testing.expect(shuffleVectorIndex(2, vector_len) == 2);
+    testing.expect(shuffleVectorIndex(3, vector_len) == 3);
+
+    testing.expect(shuffleVectorIndex(4, vector_len) == -1);
+    testing.expect(shuffleVectorIndex(5, vector_len) == -2);
+    testing.expect(shuffleVectorIndex(6, vector_len) == -3);
+    testing.expect(shuffleVectorIndex(7, vector_len) == -4);
+}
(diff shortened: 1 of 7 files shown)


andrewrk:
> is there a way to require a vector to have a certain alignment?

There is not. However, all Zig vector types have alignment 16, so I think we can safely ignore that value if it is <= 16. And I suspect that we will not encounter a higher alignment in practice (please correct me if I'm wrong).

> Most real-world vector code will use intrinsics e.g. `__builtin_ia32_packsswb` - is there any way to use these from Zig?

I think we should be able to translate these function calls into zig SIMD expressions on vectors, right? I'm not sure exactly what that function does but if it was for example vector addition then we could translate to e.g. `a + b`.

I see that you are already doing that in this PR with `@shuffle` though, so I think I did not quite understand your question fully.

andrewrk (approved):
This is pretty slick. Nice work!

ehaas:
I found an issue talking about intrinsics: https://github.com/ziglang/zig/issues/7702

`__builtin_ia32_packsswb` was just a random example I picked - if you look in `mmintrin.h` you'll see a function `_mm_packs_pi16` that calls it. It corresponds to an MMX instruction `PACKSSWB` that turns 8 signed 16-bit integers into 8 signed 8-bit integers with saturation (anything greater than `127` becomes `127`, less than `-128` becomes `-128` - see https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_m_packsswb)

There are a lot of them - just for x86 we have: https://gcc.gnu.org/onlinedocs/gcc-10.2.0/gcc/x86-Built-in-Functions.html#x86-Built-in-Functions

In theory all of them could be emulated in pure Zig but that would be a lot of work and pretty error prone and possibly miss out on the ability to compile them down to a single instruction depending on how smart the backend is. The proposal I linked talked about possibly having a way to forward them to the backend but I have no idea how difficult that would be to implement.

I think in general user code would not directly use these builtins; using our example they would call `_mm_packs_pi16` which is implemented in some Zig stdlib location. That implementation would perhaps look at the backend (llvm? pure Zig?) and the target (mmx enabled?) and either forward it to the backend, emulate it in Zig, or be a `@compileError()`

andrewrk:
The plan for these things is to have builtin functions for all of them. We have an entire `@` namespace for compiler-provided functions, so there's no concern about cluttering up the global namespace. Pretty much anything that has a dedicated hardware instruction on *some* backend, will have either syntax support or support via builtin functions. Then when a given target ISA has the instruction, the backend lowers it to machine code, and when the target ISA does not have the instruction, the backend lowers it to shim code (perhaps via a call to compiler_rt).

ehaas:
Cool, so if I'm understanding correctly, for the purposes of `translate-c` we might have something like this in `std/c/builtins.zig`:

```zig
pub fn __builtin_ia32_packsswb(a: Vector(4, i16), b: Vector(4, i16)) callconv(.Inline) Vector(8, i8) {
    return @"intrinsic.mmx.packsswb"(a, b);
}
```
and then translate-c would not have to do any special work, it would just notice that the function is defined and emit a call to it. Alternatively `translate-c` could know how to map certain `__builtin_` functions to the right intrinsic and then just directly emit a call to the `@` function.

On the topic of alignment - some AVX stuff has higher than 16 byte alignment. `avx512fintrin.h` and `avx512bf16intrin.h` define some vector types with 32 or 64 byte alignment, and `avxintrin.h` has some 32-byte aligned types. I could probably check if the vector has alignment greater than 16 and demote it in that case.

daurnimator:
@ehaas I think I'd rather see it somewhere in the zig standard library as conditional comptime assembly, e.g.

```zig
fn saturatingTruncate(comptime destType: T, src: anytype) T {
  if (!@isComptime() // see https://github.com/ziglang/zig/issues/868
    and destType == Vector(i8,8)
    and @TypeOf(src) == Vector(i16,8)
    and std.builtin.cpu.arch == .x86
    and std.builtin.cpu.features.has(.mmx)
  ) {
    return asm("packsswb ...." : src); // or alternatively, this could be the LLVM intrinsic? see https://github.com/ziglang/zig/issues/4466 or https://github.com/ziglang/zig/issues/2291
  }
  
  // implementation using primitive addition/etc.
}
```

andrewrk:
> Cool, so if I'm understanding correctly...

Yep, looks like we're on the same page. Although I'd probably want to name it something less fancy, such as `@mmx_packsswb`, and as you mentioned, have the translate-c code directly emit calls to it, rather than a wrapper function. 

> On the topic of alignment

Interesting- I'm glad you brought this up. I think we might need to consider adjusting zig's type system to take into account vector alignment. Would you mind opening a new issue for this topic?

PR #8393, opened 2021-03-30
os: handle ECONNRESET for connect() syscall

lithdew:
ECONNRESET may be returned on connect() if a peer forcibly closes the connection while it is being established. Encountered while implementing auto-reconnect for a TCP client.


ifreund (commented):
I can confirm that this can be returned by `connect()` on linux but don't have a way to test on windows currently and don't see the error listed here: https://docs.microsoft.com/en-us/windows/win32/api/Winsock2/nf-winsock2-connect. Have you observed this error being returned from `connect()` on windows?

lithdew:
> I can confirm that this can be returned by `connect()` on linux but don't have a way to test on windows currently and don't see the error listed here: https://docs.microsoft.com/en-us/windows/win32/api/Winsock2/nf-winsock2-connect. Have you observed this error being returned from `connect()` on windows?

Yep, I have tested that this error is returned from connect() on Windows as well.

ifreund (approved):
Thanks!

PR #8394, opened 2021-03-30
translate-c: fix enums that require c_uint type

### LINKED ISSUE #8003: translate-c creates extern enum(c_int) with values that don't fit in c_int
```c
enum DXGI_FORMAT {
    DXGI_FORMAT_FORCE_UINT = 0xffffffff
};
```
```zig
// -target x86_64-windows-gnu
// error: integer value 4294967295 cannot be coerced to type 'c_int'
pub const enum_DXGI_FORMAT = extern enum(c_int) {
    _FORCE_UINT = 4294967295,
    _,
};

// -target x86_64-windows-msvc
pub const enum_DXGI_FORMAT = extern enum(c_int) {
    _FORCE_UINT = -1,
    _,
};
```
If I increase the value so that it doesn't fit in a `u32`:
```zig
// -target x86_64-windows-gnu
pub const enum_DXGI_FORMAT = extern enum(c_ulonglong) {
    _FORCE_UINT = 68719476735,
    _,
};

// -target x86_64-windows-msvc
pub const enum_DXGI_FORMAT = extern enum(c_int) {
    _FORCE_UINT = -1,
    _,
};
```
Looks like clang is bug-compatible with MSVC?
https://developercommunity.visualstudio.com/content/problem/490264/standard-violation-enum-underlying-type-cannot-rep.html

xackus:
Fixes #8003.

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -1117,9 +1117,7 @@ fn transEnumDecl(c: *Context, scope: *Scope, enum_decl: *const clang.EnumDecl) E
         // default to the usual integer type used for all the enums.
 
         // default to c_int since msvc and gcc default to different types
-        const init_arg_expr = if (int_type.ptr != null and
-            !isCBuiltinType(int_type, .UInt) and
-            !isCBuiltinType(int_type, .Int))
+        const init_arg_expr = if (int_type.ptr != null)
             transQualType(c, scope, int_type, enum_loc) catch |err| switch (err) {
                 error.UnsupportedType => {
                     return failDecl(c, enum_loc, name, "unable to translate enum tag type", .{});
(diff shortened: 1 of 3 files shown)


xackus:
Enum signedness is different on windows (as the helpful comment points out). Do we
* accept what clang gives us and change tests to check the target
* pick `c_int` (or `c_uint`) as the default and compare enum values against its limits

Currently we convert `c_uint` to `c_int`, but I don't know if it's the desired behaviour or just a quick hack.

@Vexu @andrewrk

ehaas:
If I'm reading this correctly, for correct cross-platform translation it seems like we need:

1. A helper function that computes the enum tag type at comptime (instead of directly translating `enum_decl.getIntegerType()`), based on `std.Target.current.abi` and the maximum value of the enum - e.g.  `if (std.Target.current.abi == .msvc and max_val <= std.math.maxInt(c_int)) c_int else c_uint`
2. Remove `cIntTypeForEnum` and in `transCCast` add explicit cases for handling int -> enum and enum -> int casts. int -> enum will need to do a c-style cast from src_val to `@typeInfo(dst_type).Enum.tag_type` and then call `@intToEnum` on that; enum -> int will need to call `@enumToInt(src)` and then do a C-style cast to the result type. Perhaps `std.meta.cast` could help us here.

Vexu:
Did this have any issues besides `test/translate_c.zig` testcases not matching? If not we could just ignore the tag type part of the enum declarations and call it a day.

ehaas:
> Did this have any issues besides test/translate_c.zig testcases not matching?

As far as I know, that is the main issue. One implication of it is, to what extent should `translate-c` output be cross-platform?  @SpexGuy would like to see it be maximally cross-platform, which would mean determining the tag type at comptime. On the other hand this change is an improvement over status quo, I think, since it means that as long as you set your `target` properly during `translate-c` you'll get code with the correct tag type.

Vexu:
I don't see any point in trying to make translate c output cross-platform files when we are passing clang an extensive description of the target to do as much preprocessor conditional compilation as it pleases.

PR #8398, opened 2021-03-31
Use @bitCast instead of @ptrCast for HINSTANCE.

marler8997:
Currently win32metadata is defining all handle types as integral types rather than pointer types.   Maintainer @scotteson1 pointed out here (https://github.com/microsoft/win32metadata/issues/392#issuecomment-810414111) that handles are "opaque types that need to be as big as a pointer", implying that whether or not a handle is an integral or a pointer is inconsequential.  I haven't found a case that disproves this so in this change I've modified the cast to HINSTANCE to be a `@bitCast` which should handle the cases where the handle HINSTANCE is either an integral or pointer type.  I'll be updating zigwin32 to define handles as integral types like the win32metadata defines them, so this will be required to make start.zig compatible with zigwin32.  Also note that start.zig should remain compatible with the standard library's current definition of HINSTANCE as well.

### DIFF
--- a/lib/std/start.zig
+++ b/lib/std/start.zig
@@ -366,7 +366,7 @@ pub fn callMain() u8 {
 
 pub fn call_wWinMain() std.os.windows.INT {
     const MAIN_HINSTANCE = @typeInfo(@TypeOf(root.wWinMain)).Fn.args[0].arg_type.?;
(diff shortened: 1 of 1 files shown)


marler8997:
I'm closing this at the moment because I've realized that defining handle types as integral types makes them incompatible with null pointer values which contradicts the win32 documentation. (see https://github.com/microsoft/win32metadata/issues/21#issuecomment-810745057)

PR #8405, opened 2021-04-01
stage2: Add LEG backend

joachimschmidt557:
This adds a new backend to the stage2 compiler: the LEG ISA. The LEG ISA (not to be confused with ARM) is a novel state-of-the-art instruction set with a focus on speed, performance, simplicity, minimal binary size and optimality. Its applications range from tiny embedded devices to large supercomputers.

[This talk on an LLVM developer conference](https://www.youtube.com/watch?v=dQw4w9WgXcQ) gives a nice introduction to the LEG ISA and how it managed to achieve these levels of efficiency.

Here are some prelimary benchmark results of the resulting binaries (full benchmark suite is a work-in-progress):

- 420% reduction in binary size compared to ARM
- 69% performance improvement compared to x86_64

Of course, the full extent of the comparison can only be seen when more stage2 backends are complete.


jedisct1:
Amazing!

We should get this in as soon as possible.

Mouvedia:
not to be confused with https://l-e-g.github.io/

joachimschmidt557:
Happy April Fools!

For great justice,
Joachim

daurnimator:
Ha, what will they think of next? Thumb instruction set? :thinking:

PR #8410, opened 2021-04-02
Make sqrt and log functions undefined for signed integer types

### LINKED ISSUE #8359: Infinite loop in math.sqrt if argument is a runtime negative integer
First reported [here](https://github.com/ziglang/zig/pull/8355#issuecomment-806194709) but I thought it might be worth its own issue

The following results in an infinite loop:
```Zig
test "infinite loop" {
    var x: i8 = -1;
    _ = @import("std").math.sqrt(x);
}
```

At the very least this should be documented; perhaps instead the negative integer case should be checked and `@panic` called? Alternatively the signature would have to be changed to accommodate an error value. I'd be happy to make the fix if there's consensus about how it should be done.

antlilja:
Closes #8359

The functions sqrt, log, log2, log10 and ln which are undefined for negative integers now output compile errors when signed integers are used.

A test which used a signed integer was also changed to use an unsigned integer instead.

### DIFF
--- a/lib/std/math/log.zig
+++ b/lib/std/math/log.zig
@@ -31,9 +31,11 @@ pub fn log(comptime T: type, base: T, x: T) T {
         .ComptimeInt => {
             return @as(comptime_int, math.floor(math.ln(@as(f64, x)) / math.ln(float_base)));
         },
-        .Int => {
-            // TODO implement integer log without using float math
-            return @floatToInt(T, math.floor(math.ln(@intToFloat(f64, x)) / math.ln(float_base)));
+
+        // TODO implement integer log without using float math
+        .Int => |IntType| switch (IntType.signedness) {
+            .signed => return @compileError("log not implemented for signed integers"),
+            .unsigned => return @floatToInt(T, math.floor(math.ln(@intToFloat(f64, x)) / math.ln(float_base))),
         },
 
         .Float => {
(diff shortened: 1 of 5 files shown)


lib/std/math/log.zig
```diff
@@ -31,9 +31,11 @@ pub fn log(comptime T: type, base: T, x: T) T {
         .ComptimeInt => {
             return @as(comptime_int, math.floor(math.ln(@as(f64, x)) / math.ln(float_base)));
         },
-        .Int => {
-            // TODO implement integer log without using float math
-            return @floatToInt(T, math.floor(math.ln(@intToFloat(f64, x)) / math.ln(float_base)));
+
+        // TODO implement integer log without using float math
+        .Int => |IntType| switch (IntType.signedness) {
+            .signed => return @compileError("log not defined for signed integers"),
```
jedisct1, line 37: Mathematically, it is defined, and since we support complex numbers already, we could even return the result.

Maybe change the message to something like "log wouldn't be an integer" or "log of signed integers not implemented"?

lib/std/math/sqrt.zig
```diff
@@ -31,7 +31,10 @@ pub fn sqrt(x: anytype) Sqrt(@TypeOf(x)) {
             }
             return @as(T, sqrt_int(u128, x));
         },
-        .Int => return sqrt_int(T, x),
+        .Int => |IntType| switch (IntType.signedness) {
+            .signed => return @compileError("sqrt not defined for signed integers"),
```
jedisct1, line 35: Same here. This is not implemented, but well defined.

antlilja:
@jedict1 I've changed the compiler errors to be "FUNCTION not implemented for signed integers".

jedisct1:
Looking good, thank you! 👍

PR #8416, opened 2021-04-02
stage2 wasm codegen: refactor Opcode, add `sub` and `mul` operators

gracefuu:


Luukdegram (commented):
Looks good to me. If it's a lot of work to refactor the switches to generate the opcodes for conditionals/arithmetic then you could create a separate PR for that.

src/codegen/wasm.zig
```diff
@@ -305,53 +314,54 @@ pub const Context = struct {
...
-            .u64, .i64 => .i64_add,
-            .f32 => .f32_add,
-            .f64 => .f64_add,
-            else => return self.fail(inst.base.src, "TODO - Implement wasm genAdd for type '{s}'", .{inst.base.ty.tag()}),
-        };
-
-        try self.code.append(wasm.opcode(opcode));
-        return .none;
+        switch (inst.base.ty.tag()) {
+            .u32, .i32, .u64, .i64, .f32, .f64 => {
```
Luukdegram, line 325: Not really a fan of this approach. This effectively means we'll have to maintain all the switch branches here, as well as in `typeToValtype`. Instead, you should let `typeToValtype` do the work and generate an error if some unhandled/invalid type is caught. That way it becomes more maintainable as well as readable *imo*.

src/codegen/wasm.zig
```diff
@@ -305,53 +314,54 @@ pub const Context = struct {
...
-            .f32 => {
-                try writer.writeByte(wasm.opcode(.f32_const));
-                // TODO: enforce LE byte order
-                try writer.writeAll(mem.asBytes(&inst.val.toFloat(f32)));
-            },
-            .f64 => {
-                try writer.writeByte(wasm.opcode(.f64_const));
-                // TODO: enforce LE byte order
-                try writer.writeAll(mem.asBytes(&inst.val.toFloat(f64)));
+            .u32, .u64, .i32, .i64, .bool, .f32, .f64 => |tag| {
```
Luukdegram, line 340: Perhaps we should pre-check for `void` and exit early. Then you can remove the outer switch and only leave the inner switch intact to write the correct bytes.

src/codegen/wasm.zig
```diff
@@ -455,64 +465,25 @@ pub const Context = struct {
...
-            }),
-        };
-
-        const opcode = opcode_maybe orelse
-            return self.fail(inst.base.src, "TODO - Wasm genCmp for type '{s}' and operator '{s}'", .{ ty, @tagName(op) });
-
-        try self.code.append(wasm.opcode(opcode));
-        return WValue{ .code_offset = offset };
+        switch (ty) {
+            .i32, .u32, .i64, .u64, .f32, .f64 => {
```
Luukdegram, line 469: Same as my first comment. Have `typeToValtype` handle the type so we can remove this switch.

src/codegen/wasm.zig
gracefuu: This was based on an incorrect understanding of the compiler. BinOp always means numbers.

Since this only does numbers, we should always emit the same wasm (only differing in the valtype: `[iuf](32|64)`), so `typeToValtype` should handle all cases correctly.

src/codegen/wasm.zig
gracefuu: The outer switch here should still exist for constants that need to be compiled to structs (or multi values), e.g. `null: ?i64`, because emitting these constants requires more than 1 wasm op, right? I can make this change for now and worry about the struct thing later as well, if you want.

src/codegen/wasm.zig
gracefuu: Done.

src/codegen/wasm.zig
Luukdegram: Good point! Perhaps we should be switching on a different type/tag instead? I.e. switching on `zigTypeTag()` would allow us to switch on `.Int`, `.Pointer` etc. and means we don't have to be as specific in regards to the tags for now.

src/codegen/wasm.zig
Luukdegram: Good point! Perhaps we should be switching on a different type/tag instead? I.e. switching on zigTypeTag() would allow us to switch on .Int, .Pointer etc. and means we don't have to be as specific in regards to the tags for now.

src/codegen/wasm.zig
gracefuu: I think in this case, we can panic if `zigTypeTag()` is neither `.Int`, `.Float`, nor `.Bool`. the only `BinOp`s are:

```rs
                .add,
                .addwrap,
                .sub,
                .subwrap,
                .mul,
                .mulwrap,
                .cmp_lt,
                .cmp_lte,
                .cmp_eq,
                .cmp_gte,
                .cmp_gt,
                .cmp_neq,
                .store, // a little surprising that this is technically binop, but we handle this separately anyway
                .bool_and,
                .bool_or,
                .bit_and,
                .bit_or,
                .xor,
```

So for this case, we can remove the outer switch as you suggested.

It's the other case below where we might need the outer switch for, because `emitConstant` needs to handle any `inst.Constant`, where we need to handle (nearly) all types.

src/codegen/wasm.zig
gracefuu: Done.

Luukdegram (approved):
Nice work! `buildOpcode` made it a lot cleaner.

gracefuu:
Rebased onto master and fixed a typo in a comment.

lib/std/wasm.zig
```diff
@@ -231,6 +234,445 @@ test "Wasm - valtypes" {
...
+    /// Width of the operation (e.g. 8 for i32_load8_s, 16 for i64_extend16_i32_s)
+    width: ?u8 = null,
+    /// Second valtype in the opcode name (usually represents the type of the input)
+    valtype2: ?Valtype = null,
+    /// Signedness of the op
+    signedness: ?Signedness = null,
+};
+
+/// Helper function that builds an Opcode given the arguments needed
+pub fn buildOpcode(args: OpcodeBuildArguments) Opcode {
```
andrewrk, line 347: These "opinionated" functions should go into src/codegen/wasm.zig rather than into the standard library. What goes in the std lib is "bits" - that is - just enums and names for magic numbers. `buildOpcode` and related types contains logic that is not inherent to wasm, but is related to a compiler backend.

src/codegen/wasm.zig
```diff
@@ -305,56 +314,52 @@ pub const Context = struct {
...
+                    },
+                    .f32 => {
+                        // TODO: enforce LE byte order
+                        try writer.writeAll(mem.asBytes(&inst.val.toFloat(f32)));
+                    },
+                    .f64 => {
+                        // TODO: enforce LE byte order
+                        try writer.writeAll(mem.asBytes(&inst.val.toFloat(f64)));
+                    },
+                    else => unreachable,
```
andrewrk, line 358: There are more ways to represent Int, Bool, and Float than these tags. Instead of switching on tag() you should use the higher level API functions such as `intInfo`, `floatBits`.

andrewrk (changes requested):
Thanks for the patch! I have a couple suggested improvements before merging.

lib/std/wasm.zig
gracefuu: Done. moved `buildOpcode` to src/codegen/wasm.zig as suggested.

src/codegen/wasm.zig
gracefuu: Done.

Note to self: can't combine the 3 cases (Int, Bool, Float) because intInfo doesn't work on Bool and Float, floatInfo doesn't work on Int and Bool.

src/codegen/wasm.zig
```diff
@@ -747,94 +754,59 @@ pub const Context = struct {
...
+                // write opcode
+                const opcode: wasm.Opcode = buildOpcode(.{
+                    .op = .@"const",
+                    .valtype1 = try self.typeToValtype(inst.base.src, inst.base.ty),
+                });
+                try writer.writeByte(wasm.opcode(opcode));
+                // write constant
+                switch (inst.base.ty.floatBits(self.target)) {
+                    0...32 => try writer.writeIntLittle(u32, @bitCast(u32, inst.val.toFloat(f32))),
+                    64 => try writer.writeIntLittle(u64, @bitCast(u64, inst.val.toFloat(f64))),
```
gracefuu, line 804: I had a shot at fixing
```TODO: enforce LE byte order```
and this is what I found, but not sure if it's correct. Could you help me check?

File: src/codegen/wasm.zig [hunk 7]
```diff
@@ -661,12 +668,12 @@ pub const Context = struct {
...
             .load => self.genLoad(inst.castTag(.load).?),
             .loop => self.genLoop(inst.castTag(.loop).?),
-            .mul => self.genMul(inst.castTag(.mul).?),
+            .mul => self.genBinOp(inst.castTag(.mul).?, .mul),
             .not => self.genNot(inst.castTag(.not).?),
             .ret => self.genRet(inst.castTag(.ret).?),
             .retvoid => WValue.none,
             .store => self.genStore(inst.castTag(.store).?),
-            .sub => self.genSub(inst.castTag(.sub).?),
+            .sub => self.genBinOp(inst.castTag(.sub).?, .sub),
```
gracefuu, line 676: There's room here for another refactor with another helper:
```
fn binOpTagToOp(tag: Inst.Tag) Op {
    return switch (tag) {
        .cmp_eq => .eq,
        .cmp_gte => .ge,
        ...
        .add => .add,
        ...
    };
}
``` but it's lower priority.

src/codegen/wasm.zig
andrewrk: Looks great!

andrewrk:
Nice work @gracefuu!

Issue #8419, opened 2021-04-02
std.ascii isPunct returns true for several symbols


jecolon:
According to https://en.wikipedia.org/wiki/ASCII#Character_set , the following characters are symbols, not punctuation:
```
'`', '~', '$', '^', '=', '+', '|', '<', '>'
```
It so happens that this is the case in Unicode too. In `std.ascii`, there's no `isSymbol` so these characters are covered by `isPunct` which differs from both ASCII and Unicode. I have a working version at https://github.com/jecolon/ziglyph/blob/main/src/ascii.zig but don't know if I should make a PR with it or let it be handled via this issue.  That version adds `isSymbol`, with corresponding `symbol` table and `tIndex.Symbol` enum value.

g-w1:
Feel free to make a pr, or I can.

jecolon:
Ok, I'll do the PR!

daurnimator:
`ispunct` is defined by C and posix as:

>  checks for any printable character which is not a space or an alphanumeric character.

zig follows this convention, which is inherited by many formats and protocols

jecolon:
Interesting. So `isPunct` is C and POSIX compliant , not ASCII compliant, but it's in std.ascii .  Maybe there should be an `asciiIsPunct` or `isPunctAscii` that's aligned with ASCII then?

Mouvedia:
If #5019 was accepted you could have a `std.posix.isPunct` and a `std.ascii.isPunct`.

jecolon:
> If #5019 was accepted you could have a `std.posix.isPunct` and a `std.ascii.isPunct`.

That would definitely be a good thing; compliant, not misleading, less confusing.

andrewrk:
Counter-proposal: delete ascii.isPunct and also don't define it in std.os or std.posix

jecolon:
> Counter-proposal: delete ascii.isPunct and also don't define it in std.os or std.posix

Yeah, that could be the best option to avoid confusion having multiple isPunct functions in different places.

r00ster91:
As of #12448 this function is deprecated so I think this issue can be closed: https://github.com/ziglang/zig/blob/99c3578f697fe0b0151049f74208b86244b4d171/lib/std/ascii.zig#L355-L358

It will be removed eventually along with the other stuff.

jecolon:
As @r00ster91 points out, this is now a non-issue.

mnemnion:
[...]

PR #8420, opened 2021-04-02
Added isSymbol to std.ascii; fixing isPunct detecting symbols as punctuation

jecolon:
As per issue #8419 , adds isSymbol to std.ascii and fixes isPunct so it doesn't treat the following symbols as punctuation:
```
'`', '~', '$', '^', '=', '+', '|', '<', '>'
```

### DIFF
--- a/lib/std/ascii.zig
+++ b/lib/std/ascii.zig
@@ -10,7 +10,7 @@
 // (and 128 bytes is not much to pay).
 // Also does not handle Unicode character classes.
 //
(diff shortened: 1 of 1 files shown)


daurnimator:
(see https://github.com/ziglang/zig/issues/8419#issuecomment-812756866)

Vexu:
I don't want to make any changes to `std.ascii` before the linked issue is resolved. Also adding an extra bit to the table kind of nullifies the optimization meaning it would be better to change something like `isDigit` or `isSpace` to a simple switch and using it's bit instead.

PR #8439, opened 2021-04-04
stage2: wasm - "Hello world"

Luukdegram:
~~Note: conflicts with https://github.com/ziglang/zig/pull/8416. Will rebase after merge.~~

This PR implements arrays and the 'data' and 'mem' sections of wasm, allowing
for the following to work (including incremental compilation):
```zig
extern fn print(s: [*:0]const u8, len: i32) void;

pub export fn helloWorld() void {
    print("Hello, world!", 13);
}
```
Using the following JS:
```js
const fs = require('fs');
const source = fs.readFileSync("./test.wasm");
const typedArray = new Uint8Array(source);
var buffer = null;

WebAssembly.instantiate(typedArray, {
  env: {
    print: (offset, len) => {
        
        let s = "";
        for (let i = offset; i < offset + len; i++)
        	s += String.fromCharCode(buffer[i]);
        console.log(`The result is ${s}`); }

  }}).then(result => {
  	buffer = new Uint8Array(result.instance.exports.memory.buffer);
  	const helloWorld = result.instance.exports.helloWorld;
  	helloWorld();
});
```
Results in the following:
```sh
> node test test.js
> The result is Hello, world!
```

### DIFF
--- a/src/link/Wasm.zig
+++ b/src/link/Wasm.zig
@@ -47,9 +47,6 @@ offset_table_free_list: std.ArrayListUnmanaged(u32) = .{},
 /// This is ment for bookkeeping so we can safely cleanup all codegen memory
 /// when calling `deinit`
 symbols: std.ArrayListUnmanaged(*Module.Decl) = .{},
-/// Contains indexes into `symbols` that are no longer used and can be populated instead,
-/// removing the need to search for a symbol and remove it when it's dereferenced.
-symbols_free_list: std.ArrayListUnmanaged(u32) = .{},
 
 pub const FnData = struct {
     /// Generated code for the type of the function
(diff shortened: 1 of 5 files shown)


src/link/Wasm.zig
```diff
@@ -81,57 +138,112 @@ pub fn createEmpty(gpa: *Allocator, options: link.Options) !*Wasm {
...
-            else => return error.TODOImplementNonFnDeclsForWasm,
-        }
+            else => unreachable,
+        },
+        else => {},
+    }
+}
+
+/// Calculates the length of the data segment that will be occupied by the given `TypedValue`
+fn calcDataLen(self: *Wasm, typed_value: TypedValue) u32 {
```
andrewrk, line 201: Let's chat about this too - I don't quite understand why this mechanism is needed.

src/link/Wasm.zig
```diff
@@ -29,6 +30,58 @@ pub const FnData = struct {
...
+        len: u32,
+    }) = .{},
+
+    /// Returns the offset into the data segment based on a given `Decl`
+    pub fn offset(self: DataSection, decl: *const Module.Decl) u32 {
+        var cur_offset: u32 = 0;
+        return for (self.segments.items) |entry| {
+            if (entry.decl == decl) break cur_offset;
+            cur_offset += entry.len;
+        } else unreachable; // offset() called on declaration that does not live inside 'data' section
```
andrewrk, line 55: Let's chat about this. I think there is a better solution than linear scan across all decls for this.

andrewrk (changes requested):
Nice work!

I left you some suggestions to improve it, and I think we should also have a chat and plan out some of these bigger picture items such as how to organize the linker code with regards to incremental compilation.

Luukdegram:
Ok, should be all ready now!

andrewrk:
Great work! I guess the next thing to do will be to work towards WASI hello world so we can add a test case :grin:

PR #8443, opened 2021-04-05
std/build: fix ?[:0]const u8 build options

holmesmr:
As per the other string types, `?[:0]const u8` needs its own case as otherwise it will raise an error about using `{}` with slices.

There's no reasonable workaround for this, as you would have to either discount the use of the empty string value or manually rework the string to be sentinel-terminated at runtime. It's useful for passing build options to code making use of C libraries that make strong use of sentinel-terminated arrays for strings.


andrewrk:
Thanks!

PR #8447, opened 2021-04-06
stage2: add initial support for start.zig

FireFox317:
This is a draft, because i would like to get some feedback on the current structure of the code.

For example, the current code does not work correctly in the CBE, because `main` is now static and the c compiler does not like this (Probably the CBE does not take into account the exported name). `Container.fullyQualifiedNameHash` has to be implemented (I think), however I'm unsure how to do that.

Positive news:
`zig build-exe main.zig` works!
```zig
pub fn main() void {}
```

After the feedback I can fixup all the testcases :)

### DIFF
--- a/lib/std/start2.zig
+++ b/lib/std/start2.zig
@@ -1,0 +1,58 @@
+const root = @import("root");
+const builtin = @import("builtin");
+
+comptime {
+    if (builtin.output_mode == 0) { // OutputMode.Exe
+        if (builtin.link_libc or builtin.object_format == 5) { // ObjectFormat.c
+            if (!@hasDecl(root, "main")) {
+                @export(otherMain, "main");
+            }
+        } else {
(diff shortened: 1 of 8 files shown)


FireFox317:
I implemented `@hasDecl` and used this in `start2.zig` to determine whether the root zig file already exports `main` or `_start`. That means that all the current testcases should still pass. I quickly checked that, however unused strings (for example the one used in `@import`) end up in the final TZIR and some backends don't like this. We really should make sure that no unused decls end up in the final TZIR.

`main.zig`:`pub fn zigMain() void {}` now compiles! However, some backends like the CBE and LLVM backend don't take the exported name into account, which means that these tests dont pass anymore. But I think this is a good point to hand it off to @andrewrk !

andrewrk:
=>  #8470

PR #8448, opened 2021-04-06
stage2: pretty print compilation errors

### LINKED ISSUE #6033: Implement compile error notes and colored compile errors
Possibly even adding the ability to highlight "token ranges" when
emitting compile errors. This may be a little bit trickier than it seems
at first, because it may (or maybe not) involve changing the way that
Zig deals with source locations (currently byte offsets). The trick is
going to be keeping the number of bytes per ZIR instruction and other
things small (currently a single pointer-sized integer), while
facilitating quick .debug_line output which must know about line/column
numbers, and facilitating compile errors and notes that point
specifically to individual tokens.

This is a sub-task of #89.

g-w1:
Leaving as a draft as the colors will have to be factored out of `std.debug` for better error handling and ~~im getting some weird errors with `std.MultiArrayList` that only manifest in parser errors (`zig build test-stage2` does *not* trigger it): https://paste.rs/icI .~~
 
image

- [x] Fix use after free
- [x] ~~Factor the colors out of debug.zig and make more robust~~ (its already robust and handles errors exactly how we want, by just catch returning, so we will have no colors)

Fixes #6033.

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -462,7 +462,7 @@ pub const TTY = struct {
         // TODO give this a payload of file handle
         windows_api,
 
-        fn setColor(conf: Config, out_stream: anytype, color: Color) void {
+        pub fn setColor(conf: Config, out_stream: anytype, color: Color) void {
             nosuspend switch (conf) {
                 .no_color => return,
                 .escape_codes => switch (color) {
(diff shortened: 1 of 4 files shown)


g-w1:
Figured out whats going wrong, we are calling unload here 
https://github.com/ziglang/zig/blob/ec212c82bef3cbf01517eece67a8599348c7ac86/src/Compilation.zig#L1351
Not sure how to re-load it though ...

andrewrk:
Thanks! Merged in cc525bc002e456cc735f66bc4d2c5267a10b6016

PR #8449, opened 2021-04-07
stage2: implement simple enums

andrewrk:
A simple enum is an enum which has an automatic integer tag type,
all tag values automatically assigned, and no top level declarations.
Such enums are created directly in AstGen and shared by all the
generic/comptime instantiations of the surrounding ZIR code. This
commit implements, but does not yet add any test cases for, simple enums.

A full enum is an enum for which any of the above conditions are not
true. Full enums are created in Sema, and therefore will create a unique
type per generic/comptime instantiation. This commit does not implement
full enums. However the `enum_decl_nonexhaustive` ZIR instruction is
added and the respective Type functions are filled out.

This commit makes an improvement to ZIR code, removing the decls array
and removing the decl_map from AstGen. Instead, decl_ref and
decl_val ZIR instructions index into the `owner_decl.dependencies`
ArrayHashMap. We already need this dependencies array for incremental
compilation purposes, and so repurposing it to also use it for ZIR decl
indexes makes for efficient memory usage.

Similarly, this commit fixes up incorrect memory management by removing
the `const` ZIR instruction. The two places it was used stored memory in
the AstGen arena, which may get freed after Sema. Now it properly sets
up a new anonymous Decl for error sets and uses a normal decl_val
instruction.

The other usage of `const` ZIR instruction was float literals. These are
now changed to use `float` ZIR instruction when the value fits inside
`zir.Inst.Data` and `float128` otherwise.

AstGen + Sema: implement int_to_enum and enum_to_int. No tests yet; I expect to
have to make some fixes before they will pass tests. Will do that in the
branch before merging.

AstGen: fix struct astgen incorrectly counting decls as fields.

Type/Value: give up on trying to exhaustively list every tag all the
time. This makes the file more manageable. Also found a bug with
i128/u128 this way, since the name of the function was more obvious when
looking at the tag values.

Type: implement abiAlignment and abiSize for structs. This will need to
get more sophisticated at some point, but for now it is progress.

Value: add new `enum_field_index` tag.
Value: add hash_u32, needed when using ArrayHashMap.


src/type.zig
```diff
@@ -1583,6 +1619,11 @@ pub const Type = extern union {
         }
     }
 
+    /// Returns whether the type is represented as a pointer in the ABI.
+    pub fn isAbiPtr(self: Type) bool {
```
Vexu, line 1623: Seems like `isPtrLikeOptional` should do the same.

src/codegen/c.zig
```diff
@@ -472,6 +500,13 @@ pub const DeclGen = struct {
...
                 dg.typedefs.putAssumeCapacityNoClobber(t, .{ .name = name, .rendered = rendered });
             },
+            .Enum => {
+                // For enums, we simply use the integer tag type.
+                var int_tag_ty_buffer: Type.Payload.Bits = undefined;
+                const int_tag_ty = t.intTagType(&int_tag_ty_buffer);
+
+                try dg.renderType(w, int_tag_ty);
+            },
             .Null, .Undefined => unreachable, // must be const or comptime
```
g-w1, line 510: EnumLiteral can be here too

Issue #8454, opened 2021-04-07
mem.replacementSize adjacent matches bug


jumpnbrownweasel:
Adjacent matches are not handled correctly by mem.replacementSize.

```zig
test "replacementSize" {
    std.testing.expectEqual(@as(usize, 2), std.mem.replacementSize(u8, "\\n\\n", "\\n", "\n"));
}
```
Result in: ```Test [1/42] test "replacementSize"... expected 2, found 3```

This was reported on discord by @cryptocode here:
https://discord.com/channels/605571803288698900/605572581046747136/829096724500840488

I'm putting together a PR and will be working with cryptocode on it.

jumpnbrownweasel:
@cryptocode Here is a draft PR to review: #8455 

I added two tests for the 'replace' method as well as for 'replacementSize'. The changes to the existing tests for the 'replace' method were done because they were incorrect and just happened to work -- the size of two expected results just happened to be the same length.

I did not include any changes to address the O(n^2) problem that was pointed out on discord. The 'replace()' fn also has the same problem and I think it should be addressed separately with quite a bit more testing added.

andrewrk:
Thanks for the fix! Landed in bf67a3fdc9efea7126058b21e687c24868bc1268.

PR #8455, opened 2021-04-07
#8454 Fix for std.mem.replacementSize adjacent matches bug.

jumpnbrownweasel:
When two 'needle' values are adjacent in the 'input' slice, the size is not
counted correctly. The 2nd 'needle' value is not matched because the index is
incremented by one after changing the index to account for the first value.

The impact is that the size returned is incorrect, which could cause UB when this
size is used as the size of the output buffer passed to std.mem.replace.

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -1899,22 +1899,38 @@ pub fn replace(comptime T: type, input: []const T, needle: []const T, replacemen
 test "replace" {
     var output: [29]u8 = undefined;
     var replacements = replace(u8, "All your base are belong to us", "base", "Zig", output[0..]);
+    var expected: []const u8 = "All your Zig are belong to us";
     testing.expect(replacements == 1);
-    testing.expect(eql(u8, output[0..], "All your Zig are belong to us"));
+    testing.expect(eql(u8, output[0..expected.len], expected));
 
     replacements = replace(u8, "Favor reading code over writing code.", "code", "", output[0..]);
+    expected = "Favor reading  over writing .";
     testing.expect(replacements == 2);
-    testing.expect(eql(u8, output[0..], "Favor reading  over writing ."));
+    testing.expect(eql(u8, output[0..expected.len], expected));
+
+    // Try adjacent replacements.
+
+    replacements = replace(u8, "\\n\\n", "\\n", "\n", output[0..]);
+    expected = "\n\n";
+    testing.expect(replacements == 2);
+    testing.expect(eql(u8, output[0..expected.len], expected));
+
+    replacements = replace(u8, "abbba", "b", "cd", output[0..]);
+    expected = "acdcdcda";
+    testing.expect(replacements == 3);
+    testing.expect(eql(u8, output[0..expected.len], expected));
 }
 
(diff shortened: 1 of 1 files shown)


lib/std/mem.zig
```diff
@@ -1899,22 +1899,38 @@ pub fn replace(comptime T: type, input: []const T, needle: []const T, replacemen
...
 /// Calculate the size needed in an output buffer to perform a replacement.
 pub fn replacementSize(comptime T: type, input: []const T, needle: []const T, replacement: []const T) usize {
     var i: usize = 0;
     var size: usize = input.len;
-    while (i < input.len) : (i += 1) {
+    while (i < input.len) {
         if (mem.indexOf(T, input[i..], needle) == @as(usize, 0)) {
             size = size - needle.len + replacement.len;
             i += needle.len;
+        } else {
```
LemonBoy, line 1932: `std.mem.replacementSize(u8, "foo", "", "a")` will now enter an endless loop.
Passing an empty string as needle is somewhat useful to intersperse the replacement between each item, we may want to handle it or filter it out with an assertion.

lib/std/mem.zig
```diff
@@ -1899,22 +1899,38 @@ pub fn replace(comptime T: type, input: []const T, needle: []const T, replacemen
...
+
+    replacements = replace(u8, "\\n\\n", "\\n", "\n", output[0..]);
+    expected = "\n\n";
+    testing.expect(replacements == 2);
+    testing.expect(eql(u8, output[0..expected.len], expected));
+
+    replacements = replace(u8, "abbba", "b", "cd", output[0..]);
+    expected = "acdcdcda";
+    testing.expect(replacements == 3);
+    testing.expect(eql(u8, output[0..expected.len], expected));
```
LemonBoy, line 1921: Nit, use expectEqualStrings in place of expect + eql.

lib/std/mem.zig
cryptocode: Rust, Go, JS, etc support this, and I think Zig should too. It currently doesn't on master (at least `replaceOwned` panics if given an empty needle), so maybe that could be a follow-up PR and just assert in this one.

jumpnbrownweasel: I prefer to add an assert for this for now, and implement "intersperse" as a separate PR. I tried this (empty needle) with mem.replace and it just loops until the output buffer overflows. So we currently don't support this in any form. I'll add the assert to all the appropriate places and retest.

lib/std/mem.zig
jumpnbrownweasel: Will do.

lib/std/mem.zig
cryptocode: I think that plan makes sense 👍

cryptocode (approved):
LGTM, tested locally and it solves the original issue I was having

jumpnbrownweasel:
@LemonBoy Is there anything else you need from me for review? @cryptocode has approved it.

andrewrk:
Thanks @jumpnbrownweasel!

PR #8463, opened 2021-04-07
netbsd: minor fixes to allow stage1 to build

mikdusan:
- netbsd 9.1 amd64
- no testing done but basic stuff works

### DIFF
--- a/lib/std/c/netbsd.zig
+++ b/lib/std/c/netbsd.zig
@@ -18,14 +18,14 @@ pub extern "c" fn _lwp_self() lwpid_t;
 
 pub extern "c" fn pipe2(fds: *[2]fd_t, flags: u32) c_int;
 pub extern "c" fn arc4random_buf(buf: [*]u8, len: usize) void;
-pub extern "c" fn __fstat50(fd: fd_t, buf: *Stat) c_int;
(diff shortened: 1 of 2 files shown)


lib/std/c/netbsd.zig
```diff
@@ -34,7 +34,15 @@ pub extern "c" fn __getrusage50(who: c_int, usage: *rusage) c_int;
 pub extern "c" fn __libc_thr_yield() c_int;
 
 pub extern "c" fn posix_memalign(memptr: *?*c_void, alignment: usize, size: usize) c_int;
-pub extern "c" fn malloc_usable_size(?*const c_void) usize;
+
+// malloc stats is no longer in libc (netbsd 9.1 places it in `libjemalloc`)
+usingnamespace if (false)
```
andrewrk, line 39: did you mean to put if (netbsd) here?

andrewrk (changes requested):

lib/std/c/netbsd.zig
mikdusan: no, it's in a netbsd-only file. I just wanted some kind of reference that we are NOT using that fn because std.heap performs hasDecl() against it. I could do away with the usingnamespace/if and comment-out `malloc_usable_size` or simply clip the whole thing. Do you have a preference?

andrewrk (approved):

PR #8464, opened 2021-04-07
stage2 wasm: Add division and bitwise/boolean ops &, |, ^, and, or

gracefuu:
While adding div to wasm, had to add div to ir.zig first.

Also fix a typo in Sema.


andrewrk:
Nice work! Glad to see that aside from the typo, my bool_and/bool_or Sema code worked; this is the first test coverage of it. Thanks!

PR #8469, opened 2021-04-08
std docs: fix layout broken by the banner

xackus:
Main content still had `height: 100%`, so its bottom got cut off.


g-w1:
Note: this probably needs to be fixed in older versions on the website too. cc @kristoff-it

kristoff-it:
Thank you, I see that you also saved me from the embarrassment of having to make a commit to fix that damn typo. 

I'll take care of older versions.

PR #8470, opened 2021-04-09
stage2: blaze the trail for std lib integration

andrewrk:
Based on work by @FireFox317 in #8447.

This branch adds "builtin" and "std" to the import table when using the
self-hosted backend.

"builtin" gains one additional item:

```
pub const zig_is_stage2 = true; // false when using stage1 backend
```

This allows the std lib to do conditional compilation based on detecting
which backend is being used. This will be removed from builtin as soon
as self-hosted catches up to feature parity with stage1.
Keep a sharp eye out - people are going to be tempted to abuse this.
The general rule of thumb is do not use `builtin.zig_is_stage2`. However
this commit breaks the rule so that we can gain limited start.zig support
as we incrementally improve the self-hosted compiler.

This commit also implements `fullyQualifiedNameHash` and related
functionality, which effectively puts all Decls in their proper
namespaces. `fullyQualifiedName` is not yet implemented.

Stop printing "todo" log messages for test decls unless we are in test
mode.

Add "previous definition here" error notes for Decl name collisions.

This commit does not bring us yet to a newly passing test case.

Here's what I'm working towards:

```zig
const std = @import("std");

export fn main() c_int {
    const a = std.fs.base64_alphabet[0];
    return a - 'A';
}
```

Current output:

```
$ ./zig-cache/bin/zig build-exe test.zig
test.zig:3:1: error: TODO implement more analyze elemptr
zig-cache/lib/zig/std/start.zig:38:46: error: TODO implement structInitExpr ty
```

So the next steps are clear:
  * Sema: improve elemptr
  * AstGen: implement structInitExpr


andrewrk:
Here's where we're at after the latest commits:

```
$ ./zig-cache/bin/zig build-exe test.zig -ofmt=c
error(module): TODO: analyze usingnamespace decl
/home/andy/dev/zig/build/zig-cache/o/3bede4785089a42c29ed3ce42a91b97f/builtin.zig:12:25: error: use of undeclared identifier 'OutputMode'
```

I'll go ahead and implement `usingnamespace` now, and then I think this should work!

Issue #6828, opened 2020-10-26
Add a way to use System V ABI on Windows and Microsoft ABI on non-Windows


tadeokondrak:
Niche, but supported by GCC and Clang. (attributes ms_abi and sysv_abi)

https://github.com/ziglang/zig/pull/6607#issuecomment-705697670

EleanorNB:
Related: #6057

PR #8472, opened 2021-04-09
callconv: add SysV

sreehax:
This is useful for calling a foreign function that uses the SysV ABI from a non SysV ABI. For example, it's necessary in my project where I try to call into an ELF file loaded to memory (SysV ABI) from a zig UEFI binary (MSVC ABI).


tadeokondrak:
Related https://github.com/ziglang/zig/issues/6828

PR #8473, opened 2021-04-09
stage2: add error for private decls accessed from other files

g-w1:


src/Sema.zig
```diff
@@ -4687,7 +4687,8 @@ fn namedFieldPtr(
                 .Struct, .Opaque, .Union => {
                     if (child_type.getContainerScope()) |container_scope| {
                         if (mod.lookupDeclName(&container_scope.base, field_name)) |decl| {
-                            // TODO if !decl.is_pub and inDifferentFiles() "{} is private"
+                            if (!decl.is_pub and !std.mem.eql(u8, decl.container.file_scope.sub_file_path, block.base.namespace().file_scope.sub_file_path))
```
andrewrk, line 4690: You can simply compare `*Scope.File` pointers here

g-w1: Ah, didn't know if I could, thanks.

Issue #7187, opened 2020-11-21
stage2 x86_64 miscompilation with bool return type


joachimschmidt557:
Test:

```zig
export fn _start() noreturn {
    var x: usize = 1;
    var y: bool = getFalse();

    exit(x);
}

fn getFalse() bool {
    return false;
}

fn exit(code: usize) noreturn {
    asm volatile ("syscall"
        :
        : [number] "{rax}" (231),
          [arg1] "{rdi}" (code)
        : "rcx", "r11", "memory"
    );
    unreachable;
}
```

Expected exit code is 1, but the actual exit code is 0.

It seems like `x` is clobbered when moving the result from `getFalse` to the stack. This is the objdump:

```
Disassembly of section .text:

0000000008000000 <_start>:
 8000000:       55                      push   rbp
 8000001:       48 89 e5                mov    rbp,rsp
 8000004:       48 81 ec 10 00 00 00    sub    rsp,0x10
 800000b:       c7 45 fc 00 00 00 00    mov    DWORD PTR [rbp-0x4],0x0
 8000012:       c7 45 f8 01 00 00 00    mov    DWORD PTR [rbp-0x8],0x1
 8000019:       ff 14 25 08 00 00 04    call   QWORD PTR ds:0x4000008
 8000020:       89 45 f7                mov    DWORD PTR [rbp-0x9],eax         ; <- problem here?
 8000023:       48 8b 45 f8             mov    rax,QWORD PTR [rbp-0x8]
 8000027:       48 8b f8                mov    rdi,rax
 800002a:       ff 14 25 10 00 00 04    call   QWORD PTR ds:0x4000010
 8000031:       48 83 c4 10             add    rsp,0x10
 8000035:       5d                      pop    rbp
 8000036:       c3                      ret
        ...
```

I don't know much about x86_64 machine code, but I think the error is in 8000020.

PR #8474, opened 2021-04-09
stage2 x86_64: encoding helpers, fix bugs

gracefuu:
As stated in https://github.com/ziglang/zig/pull/8474#issuecomment-817270681, I've changed the design of this PR once. It is now just a set of helper functions for encoding, with the intention of making code more easily checkable by reading, and only _slightly_ more ergonomic to write (it helps you do bit fiddling), but otherwise doesn't try to be smart and e.g. automatically add prefixes for you.

List of bugs fixed:
1) Spilling registers to stack now correctly handles 32/64 bit. It doesn't yet handle 8 or 16 bit (#7187), which I intend to do in a separate PR making the change to the whole codegen at once, possibly at a much later date. 2ba51f87f8c8f3d0b0cc39215206bb059d898063 416b0d693340bf5d41207683c9494eebd4ce28bb
2) Typo d7158359c096fdbc1062cb5f1817a1a590340dd6
2) As a result of (1), many operations that were unnecessarily 64 bit before now correctly become 32 bit as specified, but this revealed that the specification was wrong for loading the offset to the GOT, which needed to be a 64 bit op but was specified as 32 bit. Similarly, reading function pointers from the GOT was incorrect because function pointers are 64 bit on x86_64 and aarch64, but were doing a 32 bit genSetReg instead. 91a8e8160b3580e9566870fea32af209bda13c9e

### DIFF
--- a/src/codegen.zig
+++ b/src/codegen.zig
@@ -1687,7 +1687,7 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
                         .register => |src_reg| {
                             // register, register use mr + 1 addressing mode: r/m16/32/64, r16/32/64
                             try self.encodeX8664Instruction(src, Instruction{
-                                .operand_size_64 = dst_reg.size() == 64,
+                                .operand_size_64 = dst_ty.abiSize(self.target.*) == 64,
                                 .primary_opcode_1b = mr + 1,
                                 // TODO: Explicit optional wrap due to stage 1 miscompilation :(
                                 //       https://github.com/ziglang/zig/issues/6515
@@ -1705,7 +1705,7 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
                             const imm32 = @intCast(u31, imm); // This case must be handled before calling genX8664BinMathCode.
                             if (imm32 <= math.maxInt(u7)) {
                                 try self.encodeX8664Instruction(src, Instruction{
-                                    .operand_size_64 = dst_reg.size() == 64,
+                                    .operand_size_64 = dst_ty.abiSize(self.target.*) == 64,
                                     .primary_opcode_1b = 0x83,
                                     .opcode_extension = opx,
                                     // TODO: Explicit optional wrap due to stage 1 miscompilation :(
@@ -1719,7 +1719,7 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
                                 });
                             } else {
                                 try self.encodeX8664Instruction(src, Instruction{
-                                    .operand_size_64 = dst_reg.size() == 64,
+                                    .operand_size_64 = dst_ty.abiSize(self.target.*) == 64,
                                     .primary_opcode_1b = 0x81,
                                     .opcode_extension = opx,
                                     // TODO: Explicit optional wrap due to stage 1 miscompilation :(
@@ -1743,7 +1743,7 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
                                 return self.fail(src, "stack offset too large", .{});
                             }
                             try self.encodeX8664Instruction(src, Instruction{
-                                .operand_size_64 = dst_reg.size() == 64,
+                                .operand_size_64 = abi_size == 64,
                                 .primary_opcode_1b = mr + 0x3,
                                 .reg = dst_reg,
                                 // TODO: Explicit optional wrap due to stage 1 miscompilation :(
@@ -1802,7 +1802,7 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
                 return self.fail(src, "stack offset too large", .{});
             }
             try self.encodeX8664Instruction(src, Instruction{
-                .operand_size_64 = reg.size() == 64,
+                .operand_size_64 = abi_size == 64,
                 .primary_opcode_1b = opcode,
                 .reg = reg,
                 // TODO: Explicit optional wrap due to stage 1 miscompilation :(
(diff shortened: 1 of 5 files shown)


src/codegen.zig
```diff
@@ -1487,8 +1488,14 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
             return dst_mcv;
         }
 
+        /// Perform "binary" operators, excluding comparisons.
+        /// Currently, the following ops are supported:
         /// ADD, SUB, XOR, OR, AND
-        fn genX8664BinMath(self: *Self, inst: *ir.Inst, op_lhs: *ir.Inst, op_rhs: *ir.Inst, opx: u8, mr: u8) !MCValue {
+        fn genX8664BinMath(self: *Self, inst: *ir.Inst, op_lhs: *ir.Inst, op_rhs: *ir.Inst) !MCValue {
+            // We'll handle these ops in two steps.
+            // 1) Prepare an output register, and put one of the arguments in it
```
joachimschmidt557, line 1496: I may be wrong about this, but technically, the destination can be a stack location as well, I think.

joachimschmidt557 (commented):
Wow, nice work!

src/codegen.zig
gracefuu: You're right. This happens when say, the LHS dies and is already on the stack, (e.g. `-0x0c(%ebp)`).

TODO: update comment to be correct.

src/codegen.zig
```diff
@@ -1549,18 +1556,109 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
...
             return dst_mcv;
         }
 
+        /// Wrap over Instruction.encodeInto to translate errors
+        fn encodeX8664Instruction(
+            self: *Self,
+            src: LazySrcLoc,
+            inst: Instruction,
+        ) !void {
+            inst.encodeInto(self.code) catch |err| {
```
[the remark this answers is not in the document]
gracefuu: Note to self:

```zig
inst.encodeInto(self.code) catch |err| switch (err) {
    error.OutOfMemory => return error.OutOfMemory,
    else => return self.fail(src, "Instruction.encodeInto failed because {s}", .{@errorName(err)}),
}
```
functionally does the same thing as an `if`. But `if` either lies entirely on 1 line, or will be split into 4 lines. `switch` has 1 case per line, making it only 2 lines long.

test/stage2/test.zig
```diff
@@ -358,6 +358,81 @@ pub fn addCases(ctx: *TestContext) !void {
...
+        case.addCompareOutput(
+            \\export fn _start() noreturn {
+            \\    exit();
+            \\}
+            \\
+            \\fn mul(a: u32, b: u32) u32 {
+            \\    return a * b;
+            \\}
+            \\
+            \\const x = mul(3, 4);
```
g-w1, line 396: Since this is never referenced, it is actually never run, you probably want to do exit(x - 12); Also, I think an inline function makes execute at comptime, so this is probably redundant.

joachimschmidt557:
I don't know if it's possible to add a review comment to non-visible code, but for multiplication, some code in `Sema.zig` needs to be adapted:

https://github.com/ziglang/zig/blob/faacae331d6b11c11479d96015bceb009d0e23f8/src/Sema.zig#L3807-L3812

Right now, this assumes that an operation with zero as rhs will always be the lhs, which is true for addition and subtraction, but not for multiplication as the result will be zero, not lhs, in all cases.

gracefuu:
I've changed approach here and replaced the monolithic instruction encoder with a set of simple helper functions because the old approach wasn't useful.

The old approach didn't make writing instructions any easier because you still had to know x86_64 to use it, and on the other hand for people who did know x86_64, it actually obscures some things such as whether the REX prefix byte is being added or not.

It was also probably quite wasteful when it comes to stack memory usage, because the Instruction struct contains every possible optional field, which probably just slows things down (not sure how well LLVM will inline them).

The new approach is just as hard to use (you still need to know x86_64) but it's much more readable. Every command corresponds exactly to zero or more bytes being written to the code array, and they occur in order.

Instead, to make writing easier, I've opted to make it more ZLS friendly by having
```
const X8664Encoder = @import("codegen/x86_64.zig").Encoder;
```
at the top of `codegen.zig`. Now tab completion works, and depending on your editor configuration might even show a description of what that particular function does.

https://user-images.githubusercontent.com/81774659/114297415-1d043200-9ae3-11eb-81bf-663b0ff6a5d3.mp4

If this push passes the test, I'll rebase onto master and mark as ready to review.

gracefuu:
Thanks a lot to @joachimschmidt557 for helping me debug the latest set of bugs :P

I considered whether to continue trying to write a full instruction encoder like an assembler, but decided that would be hard if we wanted it to handle the way instructions need different opcodes to handle different sizes (e.g. 0x88 for 8 bit mov, 0x89 for 16/32/64 bit mov) and different addressing forms (e.g. 0x01 for adding 16/32/64 bit from register to stack, 0x03 for adding 16/32/64 bit from stack to register).

I think the best solution currently is to proceed with the current structure. Useful functionality (e.g. genSetStack) should get abstracted into their own functions in an adhoc manner, but these functions should be written to directly emit machine code.

What should change is that each function should clearly define their contracts:

* Does it work with 8 bit?
* Can this function write TO the stack in addition to reading FROM it?
* Can this function address an offset with a large displacement (> 4GB)?
* Will exactly 1 instruction be emitted, and if not, does it use any temporary registers?

This requires that different architectures be in their own files, because the rules for x86_64 are different from arm/aarch. While working on this PR, I also found that the way different architectures have different `usingnamespace` made ZLS useless, and since roughly half of codegen.zig is arm/aarch, it was harder to navigate the codebase.

gracefuu:
Rebasing onto master due to LLVM 12 update

pixelherodev:
Nice work!

andrewrk:
Thanks @gracefuu! Nice work!

PR #8478, opened 2021-04-09
Fix langref.html anchor navigation

charlieman:
This is a simple css fix to allow using the back and forward browser buttons to navigate anchors. Right now when you click back or forward you can see the anchor change in the url but the page doesn't scroll.


andrewrk:
Nice!

Issue #8418, opened 2021-04-02
switch to new LLVM PassManager


andrewrk:
Upstream, LLVM is transitioning to a new PassManager, and they plan to remove support for the old one. See #8390 for a nice writeup with more details.

So now is the time for us to find problems with it, and get bugs reported upstream before it's too late.

I already tried merging #8390 into the `llvm12` branch and found 2 problems:

----

Problem 1. The following test fails:

test.zig

```zig
export fn main() i32 {
    @setEvalBranchQuota(1001);
    const y = rec(1001);
    return y - 1;
}

fn rec(n: usize) callconv(.Inline) usize {
    if (n <= 1) return n;
    return rec(n - 1);
}
```

How to reproduce:

```
# build self hosted compiler
./zig build
./zig-cache/bin/zig build-exe test.zig
```

Expected: `exit(0)`
Actual: Segmentation fault in `__zig_probe_stack`

----

Problem 2. LLVM is 2x slower for debug builds. Noooooooooooooooo!

------

This issue is to file some friendly bug reports upstream, make sure they get fixed upstream, and then try re-enabling the new pass manager in Zig and see if we run into any more issues.

mshockwave:
I have some good news for this issue.

**TL;DR: I was building the -O0 Pass pipeline in a wrong way**

Here is a longer story:
In #8390 , I was measuring the compilation time using our stage2 as the benchmark. More specifically, I was measuring the time (and its timing breakdown) of this command:
```
/path/to/my/zig build --prefix stage2 -Denable-llvm
```
But later on I realized that the above command only built stage2 with -O0. And indeed, new PM ran _a lot_ slower in this configuration.
However, if we try to build stage2 with ReleaseFast (which is essentially -O3) via this command:
```
/path/to/my/zig build --prefix stage2 -Denable-llvm -Drelease=true
```
New PM actually runs **~20%** faster than the legacy PM.

Long story short: With  #8390 , New PM only performs poorly in -O0.

It turned out that I was calling the wrong function to build the -O0 Pass pipeline 😛 . More specifically, I should use `PassBuilder::buildO0DefaultPipeline` rather than `PassBuilder::buildPerModuleDefaultPipeline`. It is actually forbidden to build -O0 with `buildPerModuleDefaultPipeline` and LLVM does put an assertion to enforce this rule (but I never turned on assertion during this development so I didn't realize). After changing to `buildO0DefaultPipeline`, the compilation time of -O0 is the same as legacy PM.
[...]

PR #8482, opened 2021-04-10
Switch to new LLVM PassManager

mshockwave:
Previous attempt: #8390 
Tracking issue: #8418 

Previously we tried to switch to new LLVM PassManager in `llvm12` branch, but reverted back due to two outstanding problems:
 1. A crash related to `@setEvalBranchQuota`
 2. Slow compilation speed ( > 2x slower )

It turned out these two problems were caused by building the -O0 Pass pipeline in a wrong way. This PR -- which is nearly identical to #8390 -- addressed this problem and try to bring new LLVM PassManager back to zig again.
I also removed two redundant Passes, which were added by accident, in the LTO pipeline.

So far this PR passes all the tests except those who are already failing in the `llvm12` branch.

### DIFF
--- a/src/zig_llvm.cpp
+++ b/src/zig_llvm.cpp
@@ -277,15 +277,8 @@ bool ZigLLVMTargetMachineEmitToFile(LLVMTargetMachineRef targ_machine_ref, LLVMM
         });
     }
 
-    // Passes for either debug or release build
-    if (is_debug) {
-      // NOTE: Always inliner will go away (in debug build)
-      // when the self-hosted compiler becomes mature.
-      pass_builder.registerPipelineStartEPCallback(
-        [](ModulePassManager &module_pm, OptimizationLevel OL) {
-          module_pm.addPass(AlwaysInlinerPass());
-        });
-    } else {
+    // Passes specific for release build
+    if (!is_debug) {
       pass_builder.registerPipelineStartEPCallback(
         [](ModulePassManager &module_pm, OptimizationLevel OL) {
           module_pm.addPass(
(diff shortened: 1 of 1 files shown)


mshockwave:
I tried to run all the tests again after this patched is applied and found that one test case failed:
```
$ zig build test-stack-traces
...
Test 10/10 stack-trace dumpCurrentStackTrace (Debug)
source.zig:7:8: [address] in foo (test)
    bar();
       ^
source.zig:10:8: [address] in main (test)
    foo();
       ^
start.zig:342:29: [address] in std.start.posixCallMainAndExit (test)
            return root.main();
                            ^
start.zig:163:5: [address] in std.start._start (test)
source.zig:7:8: [address] in foo (test)
    bar();
       ^
source.zig:10:8: [address] in main (test)
    foo();
       ^
start.zig:342:29: [address] in std.start.callMain (test)
            return root.main();
                            ^
start.zig:287:12: [address] in std.start.callMainWithArgs (test)
           ^
start.zig:241:17: [address] in std.start.posixCallMainAndExit (test)
                ^
start.zig:163:5: [address] in std.start._start (test)
```
It turns out I didn't fully clear the cache folder before running the tests prior to submitting this PR, which gives me a false positive that I passed this specific test.

I sincerely apologize for this mistake. I will fix it.

If keep reverting the commit brings too many burdens, maybe we can put this as a long term project and put it as a separate branch. After all, migrating to a new PassManager has never been an easy task (Although this specific test failure is kind of minor and I don't think it will be a blocker).

Again, sorry for this mistake

mikdusan:
actually, I just began an overhaul of stack_trace tests and should have something shortly to address this.

andrewrk:
Not to worry @mshockwave , thanks for your excellent work!

mshockwave:
> actually, I just began an overhaul of stack_trace tests and should have something shortly to address test sensitivity to platform and optimizer effects.

I believe it will be helpful.
The issue here is caused by the fact that AlwaysInliner pass has a quite different behaviors in legacy and new PM: In legacy PM, AlwaysInliner tries to inline functions (declarations) AND callsites with `alwaysinline` attributes. On the contrary, AlwaysInliner in new PM only inlines functions with such attribute. More specifically, on this problematic callsite:
```
```
This callsite was marked `alwaysinline` in LLVM IR but the callee function `callMainWithArgs` was not.

If we want to inline callsite -- regardless the existence of `alwaysinline` -- in new PM, we need to use the normal inliner Pass, which is quite heavy in terms of compilation time.

@andrewrk I remembered you mentioned before that after we have a complete self-hosted compiler, we're doing always-inlining in the self-hosted part rather than LLVM right? One possible workaround will be write our own simple inliner Pass that replicate the behavior of legacy AlwaysInliner Pass, which will be removed after the self-hosted always-inliner is complete. What do you think?

PR #8483, opened 2021-04-10
Improve PowerPC support

fengb:
This adds additional support:
- `_start()`
    - ~~This implementation has a bug where `posixCallMainAndExit` clobbers the stack (argv[0]) by saving an empty `r0`. Changing the call to `always_inline` actually removes the register dump.~~ thanks for identifying my error LemonBoy! :D
- Linux bits
- Linux syscalls
- clone() — I copied this directly from musl but didn't really test it

The std test suite still won't compile due to missing compiler-rt symbols but I managed to get my own test suite working:
```
$ qemu-ppc zig-cache/o/a2be15bb31655dca2f5c07f3b1a07277/test zig
Test [1/3] test ""... 
Test [2/3] syscall.test ""... 
Test [3/3] file.test "Mode"... 

All 3 tests passed.
```


andrewrk (approved):
I'm guessing the stack trace tests are borked because the line numbers in start.zig got shifted down. Fortunately @mikdusan is working on a fix for that so we don't keep tripping over this.

LemonBoy:
> This implementation has a bug where posixCallMainAndExit clobbers the stack (argv[0]) by saving an empty r0.

The [ELF ABI](http://math-atlas.sourceforge.net/devel/assembly/elfspec_ppc.pdf) mandates an initial stack frame to be present and the link-chain to be terminated with a NULL pointer.
Check out how it's done for PPC64 and how musl does it (in `musl/arch/powerpc/crt_arch.h`).

fengb:
> I'm guessing the stack trace tests are borked because the line numbers in start.zig got shifted down. Fortunately @mikdusan is working on a fix for that so we don't keep tripping over this.

mikdusan's branch passes the stacktrace tests. I think it's better to merge that first than generating merge headaches.

PR #8493, opened 2021-04-10
macos: add aarch64 (arm64) to downloads page

mikdusan:
add `zig-macos-aarch64` tarballs that are produced by azure CI to the main download page


andrewrk:
Nice!

PR #8494, opened 2021-04-10
Wrapping add/sub for the C backend

mattnite:
This function is significantly longer than other codegen functions in this file, I'm not sure if that's the nature of this operation, or some abstractions are needed in order to make it readable, probably the latter. I'm also curious how we want to implement safety checks, it became obvious to me that we're going to need to add some of these overflow checks to the regular addition operator.

My approach for the problem has been portability first, then in the future we can add compiler specific optimizations. The general algorithm is: "check if it's going to wrap, if it will, then calculate the wrapped result, else add normally". There is an optimization that can be done for arbitrary width unsigned integers but I've left that as a TODO for now. Since it's fairly difficult to determine what the generated code looks like here are some examples that cover the different cases, the corresponding zig code is just `return lhs +% rhs`.

C integer type.
```c
int overflow_add_c_int(int a0, int a1) {
 int t0;
 if ((a0 > 0) && (a1 > 0)) {
  int const t1 = a0 < a1 ? a0 : a1;
  int const t2 = INT_MAX - ((a0 > a1) ? a0 : a1);
  if (t1 > t2) {
   t0 = INT_MIN + t1 - t2 - 1;
   goto zig_wrap_t0_ret;
  }
 } else if ((a0 < 0) && (a1 < 0)) {
  int const t3 = a0 > a1 ? a0 : a1;
  int const t4 = INT_MIN - ((a0 < a1) ? a0 : a1);
  if (t3 < t4) {
   t0 = INT_MAX + t3 - t4 + 1;
   goto zig_wrap_t0_ret;
  }
 }
 t0 = a0 + a1;
zig_wrap_t0_ret:
 return t0;
}
```
Zig integer type whose size depends on target:
```c
intptr_t overflow_add_isize(intptr_t a0, intptr_t a1) {
 intptr_t t0;
 if ((a0 > 0) && (a1 > 0)) {
  intptr_t const t1 = a0 < a1 ? a0 : a1;
  intptr_t const t2 = INTPTR_MAX - ((a0 > a1) ? a0 : a1);
  if (t1 > t2) {
   t0 = INTPTR_MIN + t1 - t2 - 1;
   goto zig_wrap_t0_ret;
  }
 } else if ((a0 < 0) && (a1 < 0)) {
  intptr_t const t3 = a0 > a1 ? a0 : a1;
  intptr_t const t4 = INTPTR_MIN - ((a0 < a1) ? a0 : a1);
  if (t3 < t4) {
   t0 = INTPTR_MAX + t3 - t4 + 1;
   goto zig_wrap_t0_ret;
  }
 }
 t0 = a0 + a1;
zig_wrap_t0_ret:
 return t0;
}
```
Zig unsigned integer type which doesn't have an arbitrary bit size, it will defer to C's unsigned wrapping to reduce size of generated code:
```c
uintptr_t overflow_add_usize(uintptr_t a0, uintptr_t a1) {
 uintptr_t const t0 = a0 + a1;
 return t0;
}
```

Zig arbitrary width unsigned integer:
```c
uint8_t overflow_add_u3(uint8_t a0, uint8_t a1) {
 uint8_t t0;
 if ((a0 > 0) && (a1 > 0)) {
  uint8_t const t1 = a0 < a1 ? a0 : a1;
  uint8_t const t2 = 7 - ((a0 > a1) ? a0 : a1);
  if (t1 > t2) {
   t0 = 0 + t1 - t2 - 1;
   goto zig_wrap_t0_ret;
  }
 }
 t0 = a0 + a1;
zig_wrap_t0_ret:
 return t0;
}
```

Zig arbitrary width signed integer:
```c
int8_t overflow_add_i3(int8_t a0, int8_t a1) {
 int8_t t0;
 if ((a0 > 0) && (a1 > 0)) {
  int8_t const t1 = a0 < a1 ? a0 : a1;
  int8_t const t2 = 3 - ((a0 > a1) ? a0 : a1);
  if (t1 > t2) {
   t0 = -4 + t1 - t2 - 1;
   goto zig_wrap_t0_ret;
  }
 } else if ((a0 < 0) && (a1 < 0)) {
  int8_t const t3 = a0 > a1 ? a0 : a1;
  int8_t const t4 = -4 - ((a0 < a1) ? a0 : a1);
  if (t3 < t4) {
   t0 = 3 + t3 - t4 + 1;
   goto zig_wrap_t0_ret;
  }
 }
 t0 = a0 + a1;
zig_wrap_t0_ret:
 return t0;
}
```

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -846,15 +846,15 @@ pub fn genBody(o: *Object, body: ir.Body) error{ AnalysisFail, OutOfMemory }!voi
             // TODO use a different strategy for add that communicates to the optimizer
             // that wrapping is UB.
             .add => try genBinOp(o, inst.castTag(.add).?, " + "),
-            .addwrap => try genWrapOp(o, .add, inst.castTag(.addwrap).?),
+            .addwrap => try genWrapOp(o, inst.castTag(.addwrap).?, " + ", "addw_"),
             // TODO use a different strategy for sub that communicates to the optimizer
             // that wrapping is UB.
             .sub => try genBinOp(o, inst.castTag(.sub).?, " - "),
-            .subwrap => try genWrapOp(o, .sub, inst.castTag(.subwrap).?),
+            .subwrap => try genWrapOp(o, inst.castTag(.subwrap).?, " - ", "subw_"),
             // TODO use a different strategy for mul that communicates to the optimizer
             // that wrapping is UB.
             .mul => try genBinOp(o, inst.castTag(.sub).?, " * "),
-            .mulwrap => try genWrapOp(o, .mul, inst.castTag(.mulwrap).?),
+            .mulwrap => try genWrapOp(o, inst.castTag(.mulwrap).?, " * ", "mulw_"),
             // TODO use a different strategy for div that communicates to the optimizer
             // that wrapping is UB.
             .div => try genBinOp(o, inst.castTag(.div).?, " / "),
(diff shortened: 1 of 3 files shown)


andrewrk:
Implementation suggestion: use the preprocessor or inline functions for this! 

Feel free to edit this file: https://github.com/ziglang/zig/blob/master/src/link/C/zig.h

And add something like this:

```c
inline int8_t zig_saddw_i8(int8_t a, int8_t b) {
    // implementation here
}
```

or a macro (just be sure not to rely on any gnu-isms):

```c
#define zig_saddw_i32(a, b) /* implementation here */
```

And then when lowering `add_wrap` for signed ints, that part becomes simpler; lower it as a call to one of these functions/macros.

There's an interesting benefit to doing it this way: faster compilation! Because we make the C compiler do the macro expansion or function inlining, the Zig compiler ends up writing fewer bytes to disk, and the C compiler ends up reading fewer bytes from disk.

andrewrk:
>  I'm also curious how we want to implement safety checks, it became obvious to me that we're going to need to add some of these overflow checks to the regular addition operator.

Safety checks are handled in semantic analysis; the C backend does not need to concern itself. It will be told to call the panic function for example when a safety check fails. The non-wrapping arithmetic instructions are supposed to be Undefined Behavior if wrapping occurs.

xackus:
Won't this work for signed types that have a corresponding C type?
```c
int overflow_add_c_int(int a0, int a1) {
    return (unsigned)a0 + (unsigned)a1;
}
```

LemonBoy:
Henry Warren's Hacker Delight is the go-to reference to understand and implement overflow checking.
LLVM's [MathExtra](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/Support/MathExtras.h) is also a good read as it uses some simpler algorithms.

> Implementation suggestion: use the preprocessor or inline functions for this!

Inline functions ftw, expanding the macro soup required here is probably going to be slowww and the generated debug info pretty much useless.

mattnite:
I've opted for a mix between inline functions and macros, in this case the amount of macro expansion is constant sized, and we get all the specific functions. The implementations of the functions could be improved, likely through casting and masking, but I was running into some crashing of the test harness that I wasn't able to debug.

For testing you'll also notice that I'm using comptime to create program source for test cases, when we get array literals we should definitely offload all the cases into fewer test programs.

src/codegen/c.zig
```diff
@@ -798,6 +795,131 @@ fn genStore(o: *Object, inst: *Inst.BinOp) !CValue {
...
+    switch (inst.base.ty.tag()) {
+        .u8 => try w.writeAll("u8"),
+        .i8 => try w.writeAll("i8"),
+        .u16 => try w.writeAll("u16"),
+        .i16 => try w.writeAll("i16"),
+        .u32 => try w.writeAll("u32"),
+        .i32 => try w.writeAll("i32"),
+        .u64 => try w.writeAll("u64"),
+        .i64 => try w.writeAll("i64"),
+        .isize => try w.writeAll("isize"),
```
andrewrk, line 883: is usize intentionally missing here?

test/stage2/cbe.zig
```diff
@@ -794,6 +794,55 @@ pub fn addCases(ctx: *TestContext) !void {
         });
     }
 
+    {
+        // TODO: move these cases into the programs themselves once stage 2 has array literals
+        // TODO: add u64 tests, ran into issues with the literal generated for std.math.maxInt(u64)
+        var case = ctx.exeFromCompiledC("Wrapping operations", .{});
+        const programs = comptime blk: {
```
andrewrk, line 801: mind iterating over these at runtime? just to avoid blowing up the compile time of the test harness.

andrewrk (changes requested):
Nice work! I'll be sure to get those array literals working pronto so we can clean up the test cases :)

src/codegen/c.zig
mattnite: yes, it gets handled on line 811, since it's essentially a uintptr we just lean on C's existing wrapping for now

test/stage2/cbe.zig
mattnite: will do, I figured that would be the response to this ;)

andrewrk:
No progress for a few months; closing. Feel free to pick this back up any time.

andrewrk:
Hmm after having a second look, I think this is close enough for a hand-off. I can take it from here.

PR #8496, opened 2021-04-11
std.meta: add isError

xackus:


LewisGaul:
Thanks for adding this, looks good to me (not sure whether I should be offering a formal review).

lib/std/meta.zig
```diff
@@ -1334,3 +1334,13 @@ test "shuffleVectorIndex" {
     testing.expect(shuffleVectorIndex(6, vector_len) == -3);
     testing.expect(shuffleVectorIndex(7, vector_len) == -4);
 }
+
+/// Returns whether `error_union` contains an error.
+pub fn isError(error_union: anytype) bool {
+    return if(error_union) |_| false else |_| true;
```
LewisGaul, line 1340: Doesn't zig fmt add a space after the 'if'? I thought the stdlib was run through zig fmt?

g-w1: That is blocking on #8219.

ehaas:
Would it make sense to expand `isError` to work for all types? Right now it's a compile error if `error_union` isn't an error union, but there could be comptime checks for `@typeInfo(@TypeOf(param)) == .ErrorSet` and `@typeInfo(@TypeOf(param)) == .ErrorUnion`

lib/std/json/test.zig
```diff
@@ -29,8 +29,7 @@ fn ok(s: []const u8) !void {
 fn err(s: []const u8) void {
     testing.expect(!json.validate(s));
 
-    testNonStreaming(s) catch return;
-    testing.expect(false);
+    testing.expect(std.meta.isError(testNonStreaming(s)));
```
andrewrk, line 32:
```suggestion
    testing.expectError(error.WhateverTheErrorIs, testNonStreaming(s));
```

andrewrk (changes requested):
Implementation looks fine but existing tests should be updated to use `testing.expectError` rather than this function.

lib/std/json/test.zig
andrewrk: on second thought this is a pretty weird looking function and your code is definitely an improvement. happy to merge as is :+1:

PR #8497, opened 2021-04-11
Improve Improve PowerPC support

LemonBoy:
Complements #8483 

cc @fengb 
Do you have a real PPC machine to run the test suite on? I find it weird that `920.zig` is accumulating so much error in between the calculations.

### DIFF
--- a/lib/std/os/linux/tls.zig
+++ b/lib/std/os/linux/tls.zig
@@ -69,7 +69,7 @@ const tls_tcb_size = switch (builtin.arch) {
 
 // Controls if the TP points to the end of the TCB instead of its beginning
 const tls_tp_points_past_tcb = switch (builtin.arch) {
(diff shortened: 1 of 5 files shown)


LemonBoy:
> I find it weird that 920.zig is accumulating so much error in between the calculations.

False alarm. It turns out that PowerPC is the only architecture with no HW `@sqrt` opcode and thus it ends up using the `sqrt` method in `c.zig`. The differences in the results can be traced down to the approximation used to compute the square root, improving that is left as an exercise for the reader.

LemonBoy:
Not quite, it was a typo in the sqrt implementation stolen from musl.
We probably need some more extensive test coverage for the `c.zig` implementations.

andrewrk:
Nice find. Completely agree.

PR #8500, opened 2021-04-12
stage2: make repl ui better

g-w1:
As stage2 is improving, I found myself using this repl more and more and it was annoying to type out `update` every time, so I updated the ui.
`""` runs the previous command
Any first letter for a command also works for the command itself. As we add more commands, this could be fleshed out more to avoid collisions.
I made it print `updating...` every time so that if you did `""` and it took a short time :partying_face: you would know it was actually updating and not just skipping. Should a different message print here?

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -161,6 +161,13 @@ pub const CSourceFile = struct {
     extra_flags: []const []const u8 = &[0][]const u8{},
 };
 
+pub const CompileStats = struct {
+    total_decls: usize,
+    decls_modified: usize,
+    decls_added: usize,
+    decls_removed: usize,
+};
+
 const Job = union(enum) {
     /// Write the machine code for a Decl to the output file.
     codegen_decl: *Module.Decl,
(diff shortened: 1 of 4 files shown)


andrewrk:
> I made it print `updating...` every time so that if you did `""` and it took a short time partying_face you would know it was actually updating and not just skipping. Should a different message print here?

It should print nothing because we already have the progress bar that [shows up after 0.5 seconds](https://github.com/ziglang/zig/blob/d2d4df40746dc65802e074ab9a532cf78e698333/lib/std/Progress.zig#L52-L53)!

However if we want to indicate that an update occurred, we could have it print something more useful such as: 

`Update completed; 3/100 files changed, 123/900 decls modified, 12 added, 34 deleted`

or something to that effect.

g-w1:
Thanks, this is just the kind of feedback I was looking for!

g-w1:
I found finding added/deleted decls hard, but I think this is good enough for now: 
image

g-w1:
Ok, finally got around to making it show how many were added or removed thanks to help from andrewrk on irc.

g-w1:
I will rebase soon as https://github.com/ziglang/zig/pull/8554 was merged and included a lot of the same kinds of stuff.

g-w1:
Closing as actually the features that I most wanted were added.

PR #8503, opened 2021-04-12
os/posix: handle ECONNRESET for write/writev

### LINKED ISSUE #8426: Unexpected errno when other end disconnects
When you obtain a TCP stream (e.g. with `std.net.tcpConnectToAddress` or `std.net.StreamServer.accept`), if the other end disconnects, the program crashes with `unexpected errno: 104`.

Minimal example:
1. start a TCP server on port 8888 (`nc -l 8888`)
2. run this:
```zig
const net = @import("std").net;

pub fn main() !void {
    const addr = net.Address.resolveIp("127.0.0.1", 8888) catch unreachable;
    var s = try net.tcpConnectToAddress(addr);
    while (true) {
        s.writer().writeAll("Hello!\n") catch {};
    }
}
```
3. you should see the "Hello!" messages arrive at the TCP server
4. kill the TCP server

Notice how the code is explicitly ignoring all errors with `catch {}`, so if it enters the `while (true)` loop, it should never exit, but it crashes.

lithdew:
Fixes #8426. This PR does not handle ECONNRESET for write/writev on the WASI target yet as I have no means to test it right now; does anyone have any manpages/documentation to reference to see if ECONNRESET is reported as well for WASI?

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -675,6 +675,9 @@ pub const WriteError = error{
     /// This error occurs when no global event loop is configured,
     /// and reading from the file descriptor would block.
     WouldBlock,
+
+    /// Connection reset by peer.
+    ConnectionResetByPeer,
 } || UnexpectedError;
 
 /// Write to a file descriptor.
@@ -752,6 +755,7 @@ pub fn write(fd: fd_t, bytes: []const u8) WriteError!usize {
             ENOSPC => return error.NoSpaceLeft,
             EPERM => return error.AccessDenied,
             EPIPE => return error.BrokenPipe,
+            ECONNRESET => return error.ConnectionResetByPeer,
             else => |err| return unexpectedErrno(err),
         }
     }
@@ -820,6 +824,7 @@ pub fn writev(fd: fd_t, iov: []const iovec_const) WriteError!usize {
             ENOSPC => return error.NoSpaceLeft,
             EPERM => return error.AccessDenied,
             EPIPE => return error.BrokenPipe,
+            ECONNRESET => return error.ConnectionResetByPeer,
             else => |err| return unexpectedErrno(err),
         }
     }
--- a/src/main.zig
+++ b/src/main.zig
@@ -2859,6 +2859,7 @@ const FmtError = error{
     Unseekable,
     NotOpenForWriting,
     UnsupportedEncoding,
+    ConnectionResetByPeer,
 } || fs.File.OpenError;
 
 fn fmtPath(fmt: *Fmt, file_path: []const u8, check_mode: bool, dir: fs.Dir, sub_path: []const u8) FmtError!void {


12Boti:
ECONNRESET is in their [errno variant](https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#-errno-variant) (as `connreset`) which [sock_send returns](https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#-sock_sendfd-fd-si_data-ciovec_array-si_flags-siflags---resultsize-errno)

12Boti:
...however, we're always using `fd_write`, not `sock_send`:
https://github.com/ziglang/zig/blob/d2d4df40746dc65802e074ab9a532cf78e698333/lib/std/os.zig#L713

lithdew:
> ECONNRESET is in their [errno variant](https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#-errno-variant) (as `connreset`) which [sock_send returns](https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#-sock_sendfd-fd-si_data-ciovec_array-si_flags-siflags---resultsize-errno)

Do you have any links to any implementations of WASI (specifically code for fd_write) that might serve as a useful reference?

12Boti:
[Here](https://github.com/bytecodealliance/wasmtime/blob/2a3256787128523e6245e652870811770ffeb911/crates/wasi-common/src/snapshots/preview_1.rs#L472) is wasmtime's implementation of `fd_write`. They don't implement the `sock_` functions.

12Boti:
After looking at some code, I think it's not even possible to use sockets yet in WASI. Wasmtime doesn't even have a way to pass a socket to a program.

andrewrk:
Hmm, there are some other additional errors that can happen when using `write` for network, correct? It looks like none of them are in this error set, which leads me to believe that the idea here is to use `write` for non-networking and `send` or `sendto` for networking.

On the other hand, the OS does let you use `write` for sending data to sockets. So we should probably handle this in one of the following strategies:

 * `ECONNRESET => unreachable, // must use std.net.Socket for writing to sockets, not fs.File`
 * `ECONNRESET => return error.SocketUsedAsFile,` (and add all the other networking errors there too)
 * add all the networking errors to the WriteError set, in addition to ECONNRESET as you have here

My inclination here would be to keep `write` and `send` separate. However... there is no other way to write vectors to a socket other than `writev`! So I think applications are stuck using `write`-like syscalls for networking. This is additionally frustrating on Linux because of the `SIGPIPE`. There is no `sendv` that lets you disable `SIGPIPE` from happening when writing vectors to a socket. That leaves only the strategy of globally ignoring the SIGPIPE signal, which is annoying for a standard library to have to design around because it is global state and thus it can't freely mess with it! Ugh, what a sad state of affairs. I don't even know what to do here.

What's your opinion on all this @lithdew?

ehaas:
> However... there is no other way to write vectors to a socket other than writev

Can't `sendmsg` be used to send a vector? It also accepts `MSG_NOSIGNAL` since Linux 2.2 to return `EPIPE` instead of causing a `SIGPIPE`

andrewrk:
Oops! Thanks for the reminder @ehaas. I forgot and then didn't check carefully the man page. Yes, so the existence of ability to write vectors in sendmsg makes me think that the separation of concerns (and error sets) in the zig std lib with write/send is worth exploring.

12Boti:
OTOH, what if I wanted to write a library that can work with sockets *and* files? Would `std.io.StreamSource` have a variant that is a socket? What error would it return then? IMO, it's important to be able to handle different data sources/sinks with the same code.

lithdew:
> Hmm, there are some other additional errors that can happen when using `write` for network, correct? It looks like none of them are in this error set, which leads me to believe that the idea here is to use `write` for non-networking and `send` or `sendto` for networking.
> 
> On the other hand, the OS does let you use `write` for sending data to sockets. So we should probably handle this in one of the following strategies:
> 
> * `ECONNRESET => unreachable, // must use std.net.Socket for writing to sockets, not fs.File`
> * `ECONNRESET => return error.SocketUsedAsFile,` (and add all the other networking errors there too)
> * add all the networking errors to the WriteError set, in addition to ECONNRESET as you have here
> 
> My inclination here would be to keep `write` and `send` separate. However... there is no other way to write vectors to a socket other than `writev`! So I think applications are stuck using `write`-like syscalls for networking. This is additionally frustrating on Linux because of the `SIGPIPE`. There is no `sendv` that lets you disable `SIGPIPE` from happening when writing vectors to a socket. That leaves only the strategy of globally ignoring the SIGPIPE signal, which is annoying for a standard library to have to design around because it is global state and thus it can't freely mess with it! Ugh, what a sad state of affairs. I don't even know what to do here.
> 
> What's your opinion on all this @lithdew?

I'm all for the final approach of adding all networking errors (alongside numerous other errors) to the WriteError set.

Across all operating systems (WriteFile() for Windows, write() for POSIX, etc.), the write() syscall is generic and should work on all sorts of file handles (pipes, sockets, files, etc.). Adhering to the same semantics with `std.os.write` would be consistent.

For the time being, I think it would be good to merge solely support for `error.ConnectionResetByPeer` as the refactoring that would need to be done to make WriteError a superset of all possible errors that could occur when writing to any sort of file handle across different platforms will be quite a bit of work.

Happy to kickstart work on a PR for it after this one though :smile:.

andrewrk:
> I'm all for the final approach of adding all networking errors (alongside numerous other errors) to the WriteError set.

OK :+1:

PR #8511, opened 2021-04-12
A logging allocator that uses std.log

leecannon:
This is a re-opening of https://github.com/ziglang/zig/pull/8366 as I accidentally deleted the branch and recreating the branch with the same name does not 're-attach' the pull request

As noted in `logging_allocator.zig` (below), now that we have `std.log` there should be a logging allocator that uses it.
https://github.com/ziglang/zig/blob/9f01598a498ecbd868cb81dc1efdb1507010c1b8/lib/std/heap/logging_allocator.zig#L11

This is an implementation of such a logging allocator, taking over the name `LoggingAllocator` from the allocator that wraps a `writer`, which is now called `LogToWriterAllocator`.

There is a slight deficiency in testing this allocator as there is no way to capture the output of `std.log` without overriding the functionality in the root file.

### DIFF
--- a/lib/std/heap.zig
+++ b/lib/std/heap.zig
@@ -16,6 +16,9 @@ const maxInt = std.math.maxInt;
 
 pub const LoggingAllocator = @import("heap/logging_allocator.zig").LoggingAllocator;
 pub const loggingAllocator = @import("heap/logging_allocator.zig").loggingAllocator;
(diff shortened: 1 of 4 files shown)


marler8997:
> recreating the branch with the same name does not 're-attach' the pull request

You have to re-push the last commit that was on the original PR, then reopen it, then you can push a new commit to the branch.

leecannon:
Oh okay, I will remeber that for the next time I fat finger delete a branch 😄

Issue #6673, opened 2020-10-14
use case: zig build on the command line without having to scroll up to find the compile errors


andrewrk:
It's not apparent yet because some of this stuff is not yet implemented but there are multiple consumers of the zig build system. The programmer using a terminal is the main use case for now, but there is also IDEs, language servers, package manager, as well as other terminal based usages such as troubleshooting the build, visualizing it, and the --help menu.

Sometimes a `zig build` invocation can end up printing long error traces with "command failed" being printed after compile errors, forcing the user to have to scroll up. This is not an ideal experience. It is important for this troubleshooting information to be printed by default, however, it should be possible to get into a workflow where only the thing you care about is printed, which is often compile errors.

It's not super simple to solve this because compile errors could come from multiple different invocations of the zig compiler, or even from compiling C or C++ code. Furthermore, compile error messages from zig contain extra metadata that can't be communicated in the terminal. For example they are actually in a tree form, with error notes being attached to parent errors. Zig also has no error limit of how many to print, because the idea is that they would be collected by some kind of IDE and displayed inline with source code, in addition to a global list.

However, the use case of not using an IDE is also valid and so this issue is open to track progress on that front. I don't think `--quiet` adequately captures the intent here, which is to specifically focus compile errors - *not* to silence all output. It's possible we will see `--watch` turn into something that addresses this use case.

I also don't think reversing the traceback is appropriate. Status quo is the natural stack order; nothing needs to be buffered, and the beginning and ends of stack traces connect correctly to the error messages on either end - including with error return traces. Buffering and reversing the traceback is not a sufficient solution as well as introduces new drawbacks.

This issue is not yet a concrete proposal because I'd like to focus on the zig build system and the package manager (#943) together, but my main focus right now is the self-hosted compiler (#89). Suffice to say, this use case will be solved in a satisfying way eventually, but it's unclear exactly how yet. This is something I expect to be addressed in the 0.9.0 release cycle.

MasterQ32:
I strongly support more JSON compiler output

pfgithub:
<code><a href="https://github.com/ziglang/zig/pull/6669">--color=on</a> 2>&1 | less</code> works well but you can't go back to it if you want to view the output later

frmdstryr:
Something like grep's `--after-context=num` would be helpful. Then a limit could be set on the number of nodes after each error.

frmdstryr:
Bump... just had an error trace that was 1795 lines long lol

[...]

PR #8513, opened 2021-04-12
Add option to hide build command on compilation error to build_runner

leecannon:
I know https://github.com/ziglang/zig/issues/6673 is still open with no consensus for the long term improvement of cli errors, but with a project with a large number of C files added you get errors like [this](https://pastebin.com/NuJ11rMn) (~36,000 characters of build command) which is just actively hostile to a nice debugging experience.

This PR is a stopgap that is not intended to be permanent which adds `hide_build_command_on_error: bool` on builder and `--hide-build-command-on-error` command line argument when using `zig build`

The scope of this change is kept small with the flag only taken into account in `execFromStep` when encountering a `ExitCodeFailure` as this is the path that a normal failed compilation will execute, rather than checking this flag everytime a failed command is to be printed e.g. on file not found.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -45,7 +45,8 @@ pub const Builder = struct {
     verbose_llvm_ir: bool,
     verbose_cimport: bool,
     verbose_llvm_cpu_features: bool,
-    hide_build_command_on_error: bool,
+    /// The purpose of executing the command is for a human to read compile errors from the terminal
+    prominent_compile_errors: bool,
     color: enum { auto, on, off } = .auto,
     invalid_user_input: bool,
     zig_exe: []const u8,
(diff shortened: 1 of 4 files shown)


andrewrk:
Rather than `--hide-build-command-on-error` I suggest for this feature to be named after the actual goal: surface compile errors. The implementation can be the same as this stop gap solution for now, but it allows room to improve the feature to do the following:
 * Detect compile errors and present only them
 * Still print the build command as usual if a non-compile-error crash happened

Counter proposal: `--prominent-compile-errors`
Description: "Indicate the purpose of executing the command is for a human to read compile errors from the terminal"

In the future there will be other options that make the output more consumable by an IDE and other tooling, so we want to be sure to make clear the niche this flag serves.

leecannon:
Does it make sense for it to be the other way round?

i.e. zig assumes without any additional flags that its output is for human consumption and requires additional flags for more detailed output for use by tooling, compiler troubleshooting, etc.

andrewrk:
>  Description: "Indicate the purpose of executing the command is for a human to read compile errors from the terminal"

There are 2 components here, not just 1:

 * human, not computer, is the agent
 * **the intended purpose is to read compile errors**

The default use case is:

 * if detected a TTY, then it is a human, otherwise it is machine
 * intended purpose is to silently succeed the build, or loudly fail with all the information needed to troubleshoot a problem anywhere on the process stack.

leecannon:
Im not 100% happy with the description of the flag on the in the usage text
```
--prominent-compile-errors   Output compile errors formatted for a human to read
```

PR #8514, opened 2021-04-13
stage2: fully rely on std.zig, start.zig, and std.builtin

andrewrk:
Despite the branch name here, the goal of this branch is to get to the point where existing tests are passing, but with the main compilation being kicked off in the same way that it is with stage1: via an implicit `_ = @import("std");` which, in turn, imports start.zig, which, in turn, imports the root source file and decides how to handle the entry point.

There are a lot of miscellaneous things standing in the way of this, and this branch aims to tackle all of them! See individual commit messages for more info.


andrewrk:
=>  #8554

Issue #1512, opened 2018-09-13
ability to set alignment on struct fields


andrewrk:
Just like global variables and local variables support alignment, struct fields should support alignment.

The alignment of a struct should be the alignment of its most aligned member. As Zig iterates over the fields of a struct, it knows the current alignment at that position. When a field requires more alignment, padding bytes are inserted until the desired alignment is achieved. Zig currently does this incorrectly - See #1248 

Related is reordering fields for better performance and memory usage (see #168)

UniqueID1:
Here we can see a use case for specifying the alignment of struct fields: https://github.com/facebook/folly/blob/master/folly/ProducerConsumerQueue.h (search for 'alignas').

They specify the alignment to be the minimum offset between two objects to avoid false sharing.

andrewrk:
Note this workaround that is now in master branch, and also described in https://github.com/ziglang/zig/pull/1845#issuecomment-449068850:

https://github.com/ziglang/zig/blob/459045aa40d64ac883d3908ddcbfcb8aa3867c0d/src/analyze.cpp#L2698-L2707

I think we have to solve the underlying issue before this issue can be resolved.

andrewrk:
One more thing that should be done with this issue, is to make packed structs divide based on ABI size, not store size. So for this example:

```zig
const std = @import("std");
const assert = std.debug.assert;

comptime {
    assert(@sizeOf(u24) == 4);
}
const P = packed struct {
    a: u24,
    b: u8,
};
```

Currently the struct is generated as

```llvm
%P = type <{ i24, i8 }>
```

This is problematic because based on `@sizeOf(u24)` one expects to be able to write the 4th byte without overwriting the next field.

Instead Zig should generate the struct like this:


```llvm
%P = type <{ i32 }>
```

and then do the normal bit shifting and masking that happens for bit fields. This will enable packed structs to work effectively when specifying alignment. Otherwise, aligned u24 fields of packed structs would incorrectly be missing a padding byte.

andrewrk:
After implementing this, audit `@bitOffsetOf`, and `@bitSizeOf` if it is implemented (See #2177)

daurnimator:
@nrdmn notes that this is required for certain UEFI structures.

[...]

PR #8518, opened 2021-04-13
Update doc for struct field alignment.

akvadrako:
Although #1512 added support for struct field alignment, the docs still point to that issue as a planned feature. This changes the pointer to a simple example.


jedisct1:
Thanks for this!

Issue #4926, opened 2020-04-03
unable to build i386 glibc: lld: error: relocation refers to a symbol in a discarded section


andrewrk:
```
$ ./zig test ../test/stage1/behavior.zig -target i386-linux-gnu -lc
lld: error: relocation refers to a symbol in a discarded section: __x86.get_pc_thunk.bx
>>> defined in /home/andy/.cache/zig/stage1/o/QwPOfrBzzIXOEPuFhvbaQ4pBmiZk1vZIjVDP_sy7CCeeXOl8s6Zh8l0nMcfprTj6/crti.o
>>> referenced by crti.S:68 (/home/andy/Downloads/zig/lib/libc/glibc/sysdeps/i386/crti.S:68)
>>>               /home/andy/.cache/zig/stage1/o/QwPOfrBzzIXOEPuFhvbaQ4pBmiZk1vZIjVDP_sy7CCeeXOl8s6Zh8l0nMcfprTj6/crti.o:(.init+0x5)
>>> referenced by crti.S:88 (/home/andy/Downloads/zig/lib/libc/glibc/sysdeps/i386/crti.S:88)
>>>               /home/andy/.cache/zig/stage1/o/QwPOfrBzzIXOEPuFhvbaQ4pBmiZk1vZIjVDP_sy7CCeeXOl8s6Zh8l0nMcfprTj6/crti.o:(.fini+0x5)
```

LemonBoy:
Oh this is a nice one, there are a few problems:

* The `get_pc_thunk` stubs are actually emitted by gcc when PIC is emitted...
* ...but clang doesn't follow this convention (afaik they're replaced by some inline code)...
* ...anyways, glibc used to define `__x86.get_pc_thunk.bx` in a `.gnu.linkonce` section...
* ...that `lld` correctly discards, hence the error...
* ...but [glibc devs](https://sourceware.org/bugzilla/show_bug.cgi?id=20543) changed that piece of code to use a `comdat` section instead...
* ...and that fix was shipped with glibc 2.32 that Zig supports...
* ...but #6212 forgot to update the `sysdeps` folder!

Many issues (#3340, #3287, #2256, #4927 and this one) have the same root cause in common, missing/stale files in `sysdeps`.

@andrewrk the instructions on the wiki page should be updated, I don't see `sysdeps` being mentioned at all.
The last update was made in 697b1233f0, a while ago. Shipping a mix of old and new header files is no good.

zhaozg:
[...]

PR #8519, opened 2021-04-13
[WIP] Make stage2 build on SPARCv9 Linux

koachan:
As the title says, this is an effort to make stage2 build cleanly on SPARCv9 (sparc64) Linux.
It's not done yet, but I'm making a PR here to track the progress.


LemonBoy:
The sysdeps part must be updated to match the latest glibc release, we're shipping an incomplete and outdated copy that's causing a lot of problems.

@andrewrk I see you've updated the glibc headers in the llvm12 branch, would you mind updating this part too? (Check out #4926 for more info on that)

lib/std/fs.zig
```diff
@@ -501,7 +501,7 @@ pub const Dir = struct {
         },
         .linux => struct {
             dir: Dir,
-            buf: [8192]u8, // TODO align(@alignOf(os.dirent64)),
+            buf: [8192]u8 align(@alignOf(os.linux.dirent64)),
```
koachan, line 504: I'm still a bit confused about this. Forcing the `buf` alignment is needed for the stage2 builder program to work on stricter arches like SPARC, but if I do this it seems that the code will fail to build on other OSes because they don't have a os.dirent64 struct.

Any workaround for this? Tried changing it to `os.linux.dirent64` but it doesn't seem to allow it to build on other OSes.

src/target.zig
```diff
@@ -132,7 +132,22 @@ pub fn osRequiresLibC(target: std.Target) bool {
 }
 
 pub fn libcNeedsLibUnwind(target: std.Target) bool {
-    return switch (target.os.tag) {
+    // libunwind is only available on x86 and Arm family architectures.
```
LemonBoy, line 135: Not true, it also supports mips/mips64/ppc/ppc64/riscv64/sparc (the 32bit version) and a few more architectures.

The sparc64 [patch](https://reviews.llvm.org/D32450) is stalled, let's see if we can get the ball rolling again.

src/target.zig
```diff
@@ -143,6 +158,8 @@ pub fn libcNeedsLibUnwind(target: std.Target) bool {
         .windows => target.abi != .msvc,
         else => true,
     };
+
+    return has_libunwind and needs_libunwind;
```
LemonBoy, line 162: `needs_libunwind and !has_libunwind` is an error condition that should be bubbled up to the caller.

lib/std/fs.zig
LemonBoy: > Any workaround for this?

`align(if (builtin.os.tag != .linux) 1 else @alignOf(os.dirent64))`

> Tried changing it to os.linux.dirent64 but it doesn't seem to allow it to build on other OSes.

If you follow all the `@import`s you'll notice how pretty much everything depends on `std.Target.current.os.tag`. When you're cross-compiling to eg. `i386-windows-gnu` the resulting `os.linux` module will contain some Linux definitions and some Windows ones coming from `bits/windows.zig` (!).

koachan: Thanks, I've changed it following your direction.

koachan:
Closing this for now, as it's clear that we're still stuck on libunwind availability for stage2 building.
As for the other fixes/enhancements that I've made on this branch, I'll submit it again under a separate PR so that those can be merged first.

PR #8529, opened 2021-04-14
stage1: fix case close to 8520

### LINKED ISSUE #8520: compiler does not catch comparison with "undefined"
Based on an example from the documentation:

```zig
const expect = @import("std").testing.expect;
const Empty = struct {
    pub const PI = 3.14;
    pub const PI2 = undefined;
};
test "struct namespaced variable" {
    expect(Empty.PI == 3.14);
    expect(Empty.PI2 == 3.14); // fails to fail
    expect(@sizeOf(Empty) == 0);

    // you can still instantiate an empty struct
    const does_nothing = Empty{};
}
```

The above appears to pass a `zig test`.

g-w1:
Fix a case close to #8520.
```zig
const expect = @import("std").testing.expect;
const Empty = struct {
    pub const PI = 3.14;
    pub const PI2: u32 = undefined;
};
test "struct namespaced variable" {
    expect(Empty.PI == 3.14);
    expect(Empty.PI2 == 3.14); // fails to fail
    expect(@sizeOf(Empty) == 0);

    // you can still instantiate an empty struct
    const does_nothing = Empty{};
}
```
image
Note: without the u32 type annotation here, it does not work.

### DIFF
--- a/src/stage1/ir.cpp
+++ b/src/stage1/ir.cpp
@@ -18896,6 +18896,14 @@ static IrInstGen *ir_analyze_instruction_merge_err_sets(IrAnalyze *ira,
 
 static IrInstGen *ir_analyze_instruction_bin_op(IrAnalyze *ira, IrInstSrcBinOp *bin_op_instruction) {
     IrBinOp op_id = bin_op_instruction->op_id;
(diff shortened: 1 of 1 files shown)


g-w1:
I think this is a bug that has a bigger root and this is a bad fix, closing.

Issue #4426, opened 2020-02-10
reduce windows API bindings in the standard library to the minimum required for the standard library's own abstractions to work


andrewrk:
[Accepted Decision](https://github.com/ziglang/zig/issues/4426#issuecomment-858820245)

------

Windows has a huge API surface. This opens the question of, what is the purpose of std lib windows API bindings?

Is it...

 * :ballot_box:  minimal surface area for the std lib's own abstractions to work, with the idea of eventually being supplemented by a third party package for users who want more comprehensive API coverage?
 * :ballot_box:  a complete, up-to-date, maintained set of all Windows API functions, for some approved set of DLLs?
 * :ballot_box: some subset across DLLs that "feels right" for a standard library to have, with opinions on which functions should or should not be called (e.g. 'A' functions, see #534)?

In this proposal I am advocating for (1). I also think that (2) would be reasonable. I agree with @LemonBoy [here](https://github.com/ziglang/zig/commit/1cdefeb10b7496126bbb7d00709235abfee56a4a#commitcomment-37214089) that (3) is not reasonable.

Sobeston:
By bindings are we talking just the function signatures, or wrappings with zig errors and such?
The latter obviously being a lot more work.

mlarouche:
My vote for 2

LemonBoy:
My vote is for (2) too, if the conversion process can be completely automated it's even better.

Sobeston:
> 
> By bindings are we talking just the function signatures, or wrappings with zig errors and such?
> The latter obviously being a lot more work.

If we are talking the former, most of the work could be automated, with the caveat being that deciding which pointers are nullable will likely have to be done by a human.

Rocknest:
(3) it should contain preferred API's which excludes non thread safe functions and deprecated apis and possibly apis are not present in lowest supported windows version (win 8.1?)

Sobeston:
> 
> 
> and possibly apis are not present in lowest supported windows version (win 8.1?)

I'm not aware of microsoft ever removing api functions (at least in the last 20 years) - they're very against this.

Personally I'm against 3 primarily because it will spawn endless avoidable discussion on what does and doesn't constitute as an excludable function.

There is also that if you're using winapi instead of the rest of the standard library you're likely doing things very wrong in the first place (...let's assume we're at 1.0 and the standard library is all good and covers all common use cases).

Another thing to add to this question: how to we deal with multiple versions of windows, which have differing sets of functions? (related: #1907)

[...]

PR #8539, opened 2021-04-15
Fix call to selectSymbol in std.os.windows.user32.messageBoxW

Manuzor:
Arguments to `selectSymbol` were passed in the wrong order. It expects the static function pointer as first argument.

https://github.com/ziglang/zig/blob/a38042e3ac446194ae1496513d9b4634b1dd94b8/lib/std/os/windows/user32.zig#L15-L22


Manuzor:
I'm not sure how my change could've caused the azure pipelines check to fail. I assume it's not on my end? Especially since the Mac OS cmake step has failed but I only touched windows specific stuff in the std lib.

andrewrk:
You just happened to open the PR when the CI was broken due to llvm12 upgrade. Not your fault! Thanks for the fix. I want to mention however #4426 in which we will likely end up removing user32.dll extern decls from std lib altogether.

PR #8546, opened 2021-04-16
Add support for collecting time trace in the LLVM part

mshockwave:
LLVM time profiler can collect time traces and present them in a hierarchical view. Which breakdowns the time spent in each Pass or even IR unit (e.g. how many time did we spent on optimizing & compiling a function). The result is also exported into a format that can be easily visualized by the Chrome browser.

Currently this features is only used for collecting time trace in the LLVM part and controlled by the following environment variables:
 - `ZIG_ENABLE_LLVM_TIME_TRACE` toggles this feature
 - `ZIG_LLVM_TIME_TRACE_FILE` specifies the output time trace file. Otherwise the result will be put in "<input file>.o.time-trace" or "<input file>.s.time-trace".
 - `ZIG_LLVM_TIME_TRACE_GRANULARITY` controls the time granularity in ms (default to 500).

Here is an example usage:
 1. `$ env ZIG_ENABLE_LLVM_TIME_TRACE=1 ZIG_LLVM_TIME_TRACE_FILE=foo.time-trace.json zig build-exe foo.zig`
 2. Open Chrome and type "chrome://tracing" in the URL bar.
 3. Click the "Load" button at the top-left corner to load `foo.time-trace.json`.

Then you'll get a nice breakdown on the time spent on each Pass and each function LLVM compiled. Alternatively you can also visualize the time trace file using this website: https://www.speedscope.app

The only thing I would like to discuss with everyone here is that should we use command line flag to toggle this feature rather than environment variable?

### DIFF
--- a/src/zig_llvm.cpp
+++ b/src/zig_llvm.cpp
@@ -46,7 +46,9 @@
 #include <llvm/Support/CommandLine.h>
 #include <llvm/Support/Host.h>
 #include <llvm/Support/FileSystem.h>
+#include <llvm/Support/Process.h>
 #include <llvm/Support/TargetParser.h>
+#include <llvm/Support/TimeProfiler.h>
 #include <llvm/Support/Timer.h>
 #include <llvm/Support/raw_ostream.h>
 #include <llvm/Support/TargetRegistry.h>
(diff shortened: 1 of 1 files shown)


jedisct1:
> Are two environment variables required, though? Why not just enable this if ZIG_LLVM_TIME_TRACE_FILE is defined?

Ping :)

mshockwave:
> > Are two environment variables required, though? Why not just enable this if ZIG_LLVM_TIME_TRACE_FILE is defined?
> 
> Ping :)

I’m so sorry, somehow I didn’t get notified of your reply. 
I’m glad people find this useful. And I agree it will be better to consolidate the first two env vars into one. I’ll push my changes later

mshockwave:
Alright I've merged `ZIG_LLVM_ENABLE_TIME_TRACE` into `ZIG_LLVM_TIME_TRACE_FILE`. Now one only needs to do something like this:
```bash
env ZIG_LLVM_TIME_TRACE_FILE=foo.time-trace.json zig build-exe foo.zig
```

PR #8548, opened 2021-04-16
linux: fix number of arguments for tgkill syscall

Biolunar:
This fixes a compiler error when using the std.os.linux.tgkill function.


ifreund (approved):
LGTM, thanks!

PR #8551, opened 2021-04-16
Rename --(no-)allow-shilb-undefined to -f(no-)allow-shilb-undefined

ifreund:
This breaks with GNU ld but is consistent with our naming convention for
all the rest of the flags.


daurnimator:
you'll also need to fix build.zig

ifreund:
> you'll also need to fix build.zig

Indeed, thanks for the catch #8557

PR #8555, opened 2021-04-16
std: Add pidfd wrappers

chivay:
Hi,
this patch adds a few wrappers for the pidfd_* syscall family on Linux


jedisct1:
Thank you!

PR #8558, opened 2021-04-17
std/crypto: use finer-grained error sets in function signatures

jedisct1:
Returning the `crypto.Error` error set for all crypto operations was very convenient to ensure that errors were used consistently, and to avoid having multiple error names for the same thing.

The flipside is that callers were forced to always handle all possible errors, even those that could never be returned by a function.

This PR makes all functions return union sets of the actual errors they can return.

The error sets themselves are all limited to a single error.

Larger sets are useful for platform-specific APIs, but we don't have any of these in `std/crypto`, and I couldn't find any meaningful way to build larger sets.


lib/std/crypto/25519/ed25519.zig
```diff
@@ -51,7 +58,7 @@ pub const Ed25519 = struct {
             var h = Sha512.init(.{});
             h.update(&ss);
             h.final(&az);
-            const p = try Curve.basePoint.clampedMul(az[0..32].*);
+            const p = Curve.basePoint.clampedMul(az[0..32].*) catch return error.IdentityElement;
```
daurnimator, line 61: Why did you change this?

jedisct1: The multiplication doesn't assume a trusted base, and can return `WeakPublicKey`.

But a new key pair is being created here. Since we are multiplying the base point, the result is always in the main group; we are never going to compute a weak key.

`IdentityElement` was already the only possible error that could be returned.

The `catch` was just added to match the function error set. There's no behavior change here.

daurnimator: To not mess up return traces, shouldn't we rethrow the original error?

jedisct1:
I still have mixed feelings about this, vs leveraging one of Zig's most convenient features: automatically inferred error sets.

This PR is essentially a manually-maintained version of what our wonderful compiler can automatically do.

An alternative would be to restore the original error set, use `return std.crypto.Error.X` everywhere (guaranteeing that we never use different errors for the same thing), and just use implicit error sets in function signatures. And use error sets only when a function can return different errors on different platforms.

The downside of implicit error sets is that looking at a function signature, the set of possible errors is not immediately visible. But this is more of a current tooling issue. Generated documentation will eventually include this information, and it may be something `zls` can eventually support as well.

lib/std/crypto/25519/curve25519.zig
```diff
@@ -4,7 +4,11 @@
 // The MIT license requires this copyright notice to be included in all copies
 // and substantial portions of the software.
 const std = @import("std");
-const Error = std.crypto.Error;
+const crypto = std.crypto;
+
+const IdentityElementError = crypto.errors.IdentityElementError;
+const NonCanonicalError = crypto.errors.NonCanonicalError;
```
Mouvedia, line 10: Is it a convention to end the names with "Error"?

lib/std/crypto/25519/edwards25519.zig
```diff
@@ -235,18 +241,18 @@ pub const Edwards25519 = struct {
     /// Multiply an Edwards25519 point by a scalar without clamping it.
     /// Return error.WeakPublicKey if the resulting point is
```
Mouvedia, line 242: It's kind of weird to document only one of the 2 errors.
Id expect neither or both.

lib/std/crypto/25519/edwards25519.zig
```diff
@@ -4,10 +4,16 @@
...
 const debug = std.debug;
 const fmt = std.fmt;
 const mem = std.mem;
-const Error = std.crypto.Error;
+
+const EncodingError = crypto.errors.EncodingError;
+const IdentityElementError = crypto.errors.IdentityElementError;
+const NonCanonicalError = crypto.errors.NonCanonicalError;
+const NotSquareError = crypto.errors.NotSquareError;
+const WeakPublicKeyError = crypto.errors.WeakPublicKeyError;
```
Mouvedia, line 17: `const foo = IdentityElementError || WeakPublicKeyError;` maybe?

lib/std/crypto/25519/ed25519.zig
jedisct1: > To not mess up return traces, shouldn't we rethrow the original error?

How would you do that?

`catch |e| return @errSetCast(IdentityElementError, e);` works, but I don't feel comfortable using `@errSetCast` in case `clampedMul()` later returns something that's not in the `IdentityElementError` set.

That being said, the error means that the argument to this function was zero. Traces of internal functions are not super useful.

lib/std/crypto/25519/edwards25519.zig
jedisct1: They have different meanings, so I'm not even sure how such a couple would be named. `InsecureSettingForAPoint`?
But being on the twist may also be an insecure setting for a point, and if we add that third case `InsecureSettingForAPoint`, we'll get a partially used error set in all other cases.

lib/std/crypto/25519/edwards25519.zig
jedisct1: Thanks, the comment has been updated.

lib/std/crypto/25519/curve25519.zig
jedisct1: For error sets, yes.

lib/std/crypto/25519/edwards25519.zig
Mouvedia: Just saw 4 such cases in this file, that's why I suggested to extract it.
If it cannot be named easily then forget about it.

jedisct1: No matter how many error sets we have, applications still have to handle individual error cases. 

So, I'd rather avoid trying really hard to cluster errors. Maybe once the whole API is stable, and no new cases will be added. But doing it now is only going to make maintenance more difficult.

Rocknest:
Instead of the two extremes we could have an error set per crypto function class.

jedisct1:
> Instead of the two extremes we could have an error set per crypto function class. 

Current status: requires an `else` statement for every `switch` if we want to handle a specific case. With an error set for each category: still requires an `else` statement for every `switch`, to handle cases that can never be returned.

That would be the worst of all options IMHO.
Different names could be then used for the same thing (`AuthenticationFailed` can be returned by MACs and AEADs for example). And a given primitive would use different error sets simultaneously (ex: Gimli is a hash function, a MAC, an AEAD... all implemented in the same file).

From a security perspective, these different error cases should generally never be handled individually, anyway, the only exception being functions using allocators (and so far, there are none). Just catch any error. Leaking the exact error is generally not useful and can only introduce vulnerabilities (see RSA padding oracle attacks).

Rocknest:
> AuthenticationFailed can be returned by MACs and AEADs for example

Why would MAC and AEAD have different error set? AEAD is just a cipher + MAC.
But if leaking errors can lead to security vulnerabilities then a generic `error.CryptoFailed` is a better choice, because with `try` its too easy to leak errors.

jedisct1:
> Why would MAC and AEAD have different error set? AEAD is just a cipher + MAC.

Absolutely. But merging these wouldn't be having one error set per function class any more.

AEADs can still return more errors than MACs, for example they can return an error instead of `assert()`ing when the internal counter wraps. So an AEAD error set could be a cipher error set + a MAC error set.
But all ciphers don't use counters, some use a rolling state. So we need a different cipher error set and a different AEAD error set for these.

Trying to define error clusters right now seems a bit premature, as the set of possible errors is not even well defined yet. The way we handle errors may also change. So, let's keep things simple and flexible.

The problems trying to be solved are: 1) making consistency, by sharing an error set or a structure. 2) having a quick way to list the errors a function can return.

I'm not entirely convinced that 2) is necessary, but the changes in this PR address both problems. At the expense of extra manual maintenance, that can lead to functions defined with more errors than they can actually return.

Implicit errors + tooling to print inferred errors would be ideal. We have the former, not the later yet. 

So, here's the plan to move forward:

- Get this merged. So that looking at a function definition in `zls`, it's possible to immediately understand why an error can be returned
- Maybe move back to inferred errors once we have better support for these in the documentation generator and zls.
- Define clusters once the API stabilizes.

And avoid changes in actual errors being returned, this is the most important thing to avoid breaking changes :)

PR #8559, opened 2021-04-17
ci: convert macos-aarch64 to use zig-bootstrap

mikdusan:


mikdusan:
abandoned. Looks like #8736 is taking care of business :)

PR #8561, opened 2021-04-17
stage2: rename ErrorInt to MaxErrorInt and add docs

g-w1:
~~Using comptimePrint for now for simplicity. Once #786 this will have to be changed.~~


andrewrk:
Can you share a bit about what problem you're trying to solve here?

andrewrk:
I think status quo is less confusing than this. I don't want to invoke "max" since it hints there is some kind of maximum calculation happening associated with that type, and it's not the case.

PR #8565, opened 2021-04-18
std.atomic.Bool/Int.load should take const pointer to Self

Hadron67:
As `@atomicLoad` takes const pointers


ifreund (approved):
LGTM, Thanks!

PR #8566, opened 2021-04-18
Improvements to fs.Walker

Ominitay:
I have made changes so that Walker is used more like Dir.Iterator, including removing the use of paths, and making it a child of Dir. This makes everything more consistent. My changes do change the order in which entries are returned, in order to make things work without paths but since we make no guarantees regarding the order, so it shouldn't affect anything.


Ominitay:
It seems that I should be working closer to the original implementation, keeping the path that would be outputted. Since we can't get the path of the directory from the function, should we output the path relative to the directory? That would seem most logical to me at least.

andrewrk:
Looks like this was never passing tests - please re-open against latest master!

Ominitay:
> Looks like this was never passing tests - please re-open against latest master!

Checks are passing in #9475 :)

PR #8570, opened 2021-04-18
add Thread.setName and Thread.getName

vrischmann:
This PR adds `Thread.setName` and `Thread.getName`.

It's not complete, I only tested this on linux. I spent a bit of time researching what other platforms do and from what I can tell setting/getting a thread's name is supported on linux, windows (since [Windows 10](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getthreaddescription)), [macos](https://opensource.apple.com/source/libpthread/libpthread-454.80.2/include/pthread/pthread.h.auto.html), [openbsd](https://man.openbsd.org/pthread_get_name_np.3), [freebsd](https://www.freebsd.org/cgi/man.cgi?query=pthread_getname_np&apropos=0&sektion=3&manpath=FreeBSD+12.2-RELEASE+and+Ports&arch=default&format=html), [netbsd](https://man.netbsd.org/pthread_getname_np.3).
Annoyingly since it's non standard there are some slight differences between implementations of `pthread_setname_np`.

Before implementing everything I'd like to get some early feedback on the API.

### DIFF
--- a/lib/std/Thread.zig
+++ b/lib/std/Thread.zig
@@ -38,6 +38,192 @@ else
 
 impl: Impl,
 
+pub const max_name_len = switch (std.Target.current.os.tag) {
+    .linux => 15,
+    .windows => 31,
+    .macos, .ios, .watchos, .tvos => 63,
+    .netbsd => 31,
+    .freebsd => 15,
+    .openbsd => 31,
+    else => 0,
+};
+
+pub const SetNameError = error{
+    NameTooLong,
+    Unsupported,
+    Unexpected,
+} || os.PrctlError || os.WriteError || std.fs.File.OpenError || std.fmt.BufPrintError;
+
+pub fn setName(self: Thread, name: []const u8) SetNameError!void {
+    if (name.len > max_name_len) return error.NameTooLong;
+
+    const name_with_terminator = blk: {
+        var name_buf: [max_name_len:0]u8 = undefined;
+        std.mem.copy(u8, &name_buf, name);
+        name_buf[name.len] = 0;
+        break :blk name_buf[0..name.len :0];
+    };
+
+    switch (std.Target.current.os.tag) {
+        .linux => if (use_pthreads) {
+            const err = std.c.pthread_setname_np(self.getHandle(), name_with_terminator.ptr);
+            return switch (err) {
+                0 => {},
+                os.ERANGE => unreachable,
+                else => return os.unexpectedErrno(err),
+            };
+        } else if (use_pthreads and self.getHandle() == std.c.pthread_self()) {
+            const err = try os.prctl(.SET_NAME, .{@ptrToInt(name_with_terminator.ptr)});
+            return switch (err) {
+                0 => {},
+                else => return os.unexpectedErrno(err),
+            };
+        } else {
+            var buf: [32]u8 = undefined;
+            const path = try std.fmt.bufPrint(&buf, "/proc/self/task/{d}/comm", .{self.getHandle()});
+
+            const file = try std.fs.cwd().openFile(path, .{ .write = true });
+            defer file.close();
+
+            try file.writer().writeAll(name);
+        },
+        .windows => if (std.Target.current.os.isAtLeast(.windows, .win10_rs1)) |res| {
+            // SetThreadDescription is only available since version 1607, which is 10.0.14393.795
+            // See https://en.wikipedia.org/wiki/Microsoft_Windows_SDK
+            if (!res) {
+                return error.Unsupported;
+            }
+
+            var name_buf_w: [max_name_len:0]u16 = undefined;
+            const length = try std.unicode.utf8ToUtf16Le(&name_buf_w, name);
+            name_buf_w[length] = 0;
+
+            try os.windows.SetThreadDescription(
+                self.getHandle(),
+                @ptrCast(os.windows.LPWSTR, &name_buf_w),
+            );
+        } else {
+            return error.Unsupported;
+        },
+        .macos, .ios, .watchos, .tvos => if (use_pthreads) {
+            // There doesn't seem to be a way to set the name for an arbitrary thread, only the current one.
+            if (self.getHandle() != std.c.pthread_self()) return error.Unsupported;
+
+            const err = std.c.pthread_setname_np(name_with_terminator.ptr);
+            return switch (err) {
+                0 => {},
+                else => return os.unexpectedErrno(err),
+            };
+        },
+        .netbsd => if (use_pthreads) {
+            const err = std.c.pthread_setname_np(self.getHandle(), name_with_terminator.ptr, null);
+            return switch (err) {
+                0 => {},
+                os.EINVAL => unreachable,
+                os.ESRCH => unreachable,
+                os.ENOMEM => unreachable,
+                else => return os.unexpectedErrno(err),
+            };
+        },
+        .freebsd, .openbsd => if (use_pthreads) {
+            // Use pthread_set_name_np for FreeBSD because pthread_setname_np is FreeBSD 12.2+ only.
+            // TODO maybe revisit this if depending on FreeBSD 12.2+ is acceptable because pthread_setname_np can return an error.
+
+            std.c.pthread_set_name_np(self.getHandle(), name_with_terminator.ptr);
+        },
+        else => return error.Unsupported,
+    }
+}
+
+pub const GetNameError = error{
+    // For Windows, the name is converted from UTF16 to UTF8
+    CodepointTooLarge,
+    Utf8CannotEncodeSurrogateHalf,
+    DanglingSurrogateHalf,
+    ExpectedSecondSurrogateHalf,
+    UnexpectedSecondSurrogateHalf,
+
+    Unsupported,
+    Unexpected,
+} || os.PrctlError || os.ReadError || std.fs.File.OpenError || std.fmt.BufPrintError;
+
+pub fn getName(self: Thread, buffer_ptr: *[max_name_len:0]u8) GetNameError!?[]const u8 {
+    buffer_ptr[max_name_len] = 0;
+    var buffer = std.mem.span(buffer_ptr);
+
+    switch (std.Target.current.os.tag) {
+        .linux => if (use_pthreads and comptime std.Target.current.abi.isGnu()) {
+            const err = std.c.pthread_getname_np(self.getHandle(), buffer.ptr, max_name_len + 1);
+            return switch (err) {
+                0 => std.mem.sliceTo(buffer, 0),
+                os.ERANGE => unreachable,
+                else => return os.unexpectedErrno(err),
+            };
+        } else if (use_pthreads and self.getHandle() == std.c.pthread_self()) {
+            const err = try os.prctl(.GET_NAME, .{@ptrToInt(buffer.ptr)});
+            return switch (err) {
+                0 => std.mem.sliceTo(buffer, 0),
+                else => return os.unexpectedErrno(err),
+            };
+        } else if (!use_pthreads) {
+            var buf: [32]u8 = undefined;
+            const path = try std.fmt.bufPrint(&buf, "/proc/self/task/{d}/comm", .{self.getHandle()});
+
+            const file = try std.fs.cwd().openFile(path, .{});
+            defer file.close();
+
+            const data_len = try file.reader().readAll(buffer_ptr[0 .. max_name_len + 1]);
+
+            return if (data_len >= 1) buffer[0 .. data_len - 1] else null;
+        } else {
+            // musl doesn't provide pthread_getname_np and there's no way to retrieve the thread id of an arbitrary thread.
+            return error.Unsupported;
+        },
+        .windows => if (std.Target.current.os.isAtLeast(.windows, .win10_rs1)) |res| {
+            // GetThreadDescription is only available since version 1607, which is 10.0.14393.795
+            // See https://en.wikipedia.org/wiki/Microsoft_Windows_SDK
+            if (!res) {
+                return error.Unsupported;
+            }
+
+            var name_w: os.windows.LPWSTR = undefined;
+            try os.windows.GetThreadDescription(self.getHandle(), &name_w);
+            defer os.windows.LocalFree(name_w);
+
+            const data_len = try std.unicode.utf16leToUtf8(buffer, std.mem.sliceTo(name_w, 0));
+
+            return if (data_len >= 1) buffer[0..data_len] else null;
+        } else {
+            return error.Unsupported;
+        },
+        .macos, .ios, .watchos, .tvos => if (use_pthreads) {
+            const err = std.c.pthread_getname_np(self.getHandle(), buffer.ptr, max_name_len + 1);
+            return switch (err) {
+                0 => std.mem.sliceTo(buffer, 0),
+                os.ESRCH => unreachable,
+                else => return os.unexpectedErrno(err),
+            };
+        },
+        .netbsd => if (use_pthreads) {
+            const err = std.c.pthread_getname_np(self.getHandle(), buffer.ptr, max_name_len + 1);
+            return switch (err) {
+                0 => std.mem.sliceTo(buffer, 0),
+                os.EINVAL => unreachable,
+                os.ESRCH => unreachable,
+                else => return os.unexpectedErrno(err),
+            };
+        },
+        .freebsd, .openbsd => if (use_pthreads) {
+            // Use pthread_get_name_np for FreeBSD because pthread_getname_np is FreeBSD 12.2+ only.
+            // TODO maybe revisit this if depending on FreeBSD 12.2+ is acceptable because pthread_getname_np can return an error.
+
+            std.c.pthread_get_name_np(self.getHandle(), buffer.ptr, max_name_len + 1);
+            return std.mem.sliceTo(buffer, 0);
+        },
+        else => return error.Unsupported,
+    }
+}
+
 /// Represents a unique ID per thread.
 pub const Id = u64;
 
@@ -778,6 +964,94 @@ const LinuxThreadImpl = struct {
         }
     }
 };
+
+fn testThreadName(thread: *Thread) !void {
+    const testCases = &[_][]const u8{
+        "mythread",
+        "b" ** max_name_len,
+    };
+
+    inline for (testCases) |tc| {
+        try thread.setName(tc);
+
+        var name_buffer: [max_name_len:0]u8 = undefined;
+
+        const name = try thread.getName(&name_buffer);
+        if (name) |value| {
+            try std.testing.expectEqual(tc.len, value.len);
+            try std.testing.expectEqualStrings(tc, value);
+        }
+    }
+}
+
+test "setName, getName" {
+    if (std.builtin.single_threaded) return error.SkipZigTest;
+
+    const Context = struct {
+        start_wait_event: ResetEvent = undefined,
+        test_done_event: ResetEvent = undefined,
+
+        done: std.atomic.Atomic(bool) = std.atomic.Atomic(bool).init(false),
+        thread: Thread = undefined,
+
+        fn init(self: *@This()) !void {
+            try self.start_wait_event.init();
+            try self.test_done_event.init();
+        }
+
+        pub fn run(ctx: *@This()) !void {
+            // Wait for the main thread to have set the thread field in the context.
+            ctx.start_wait_event.wait();
+
+            switch (std.Target.current.os.tag) {
+                .windows => testThreadName(&ctx.thread) catch |err| switch (err) {
+                    error.Unsupported => return error.SkipZigTest,
+                    else => return err,
+                },
+                else => try testThreadName(&ctx.thread),
+            }
+
+            // Signal our test is done
+            ctx.test_done_event.set();
+
+            while (!ctx.done.load(.SeqCst)) {
+                std.time.sleep(5 * std.time.ns_per_ms);
+            }
+        }
+    };
+
+    var context = Context{};
+    try context.init();
+
+    var thread = try spawn(.{}, Context.run, .{&context});
+    context.thread = thread;
+    context.start_wait_event.set();
+    context.test_done_event.wait();
+
+    switch (std.Target.current.os.tag) {
+        .macos, .ios, .watchos, .tvos => {
+            const res = thread.setName("foobar");
+            try std.testing.expectError(error.Unsupported, res);
+        },
+        .windows => testThreadName(&thread) catch |err| switch (err) {
+            error.Unsupported => return error.SkipZigTest,
+            else => return err,
+        },
+        else => |tag| if (tag == .linux and use_pthreads and comptime std.Target.current.abi.isMusl()) {
+            try thread.setName("foobar");
+
+            var name_buffer: [max_name_len:0]u8 = undefined;
+            const res = thread.getName(&name_buffer);
+
+            try std.testing.expectError(error.Unsupported, res);
+        } else {
+            try testThreadName(&thread);
+        },
+    }
+
+    context.done.store(true, .SeqCst);
+    thread.join();
+}
 
 test "std.Thread" {
     // Doesn't use testing.refAllDecls() since that would pull in the compileError spinLoopHint.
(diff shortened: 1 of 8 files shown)


LemonBoy:
API-wise I think it makes sense to add an extra struct parameter to the thread creation function:
```zig
struct CreateThreadOptions = struct {
    thread_name: ?[:0]const u8 = null,
}
```

There's little to no point to setting the name after the thread is created.

lib/std/Thread.zig
```diff
@@ -76,6 +76,34 @@ pub fn spinLoopHint() void {
     }
 }
 
+pub fn setName(self: Thread, name: [:0]const u8) !void {
+    if (use_pthreads) {
```
LemonBoy, line 80: Early exit:
```zig
if (!use_pthreads)
    return;
```

And a pure Zig implementation for Linux can be implemented using musl's `pthread_setname_np` as guide.

lib/std/Thread.zig
```diff
@@ -76,6 +76,34 @@ pub fn spinLoopHint() void {
...
+            0 => {},
+            os.ERANGE => error.NameTooLong,
+            else => unreachable,
+        };
+    }
+}
+
+pub fn getName(self: Thread, allocator: *std.mem.Allocator) !?[:0]const u8 {
+    if (use_pthreads) {
+        // The thread name cannot be longer than 16 bytes including the null terminator.
```
LemonBoy, line 92: > The thread name  is  a  meaningful  C  language
       string, whose length is restricted to 16 characters, including the ter‐
       minating null byte ('\0').

Is this limitation lifted for other platforms?

lib/std/c.zig
```diff
@@ -276,6 +276,7 @@ pub extern "c" fn pthread_key_create(key: *pthread_key_t, destructor: ?fn (value
 pub extern "c" fn pthread_key_delete(key: pthread_key_t) c_int;
 pub extern "c" fn pthread_getspecific(key: pthread_key_t) ?*c_void;
 pub extern "c" fn pthread_setspecific(key: pthread_key_t, value: ?*c_void) c_int;
+pub extern "c" fn pthread_getname_np(thread: std.c.pthread_t, name: [*:0]const u8, len: usize) c_int;
```
LemonBoy, line 279: Non portable definitions go into their own `<platform>/c.zig` file, this is not available when targeting eg. Windows.

lib/std/Thread.zig
vrischmann: ok for the early exit. and yeah I planned to follow musl's implementation.

lib/std/Thread.zig
vrischmann: I don't think so, I wrote this comment before realising `pthread_setname_np` is non standard and looking at the man pages of the BSDs they don't mention a length limitation. 

I'll change the comment.

lib/std/c.zig
vrischmann: ok, I'll move it.

daurnimator:
> There's little to no point to setting the name after the thread is created.

It can be useful to name worker threads after what they are currently doing.

This also has implications in schedulers where workloads are not locked to specific threads.

lib/std/Thread.zig
semarie: FYI, OpenBSD has 32 bytes limit ('\0' included). the string is copied and silenciously truncated to fit

vrischmann:
I've implemented linux with or without pthreads, macOS and netbsd/freebsd/openbsd. I've only been able to test linux at the moment, I'll be able to test freebsd and probably netbsd.

I'm still working on the windows implementation.

I'll make the API change last.

lib/std/c/openbsd.zig
```diff
@@ -45,3 +45,6 @@ pub extern "c" fn posix_memalign(memptr: *?*c_void, alignment: usize, size: usiz
 
 pub extern "c" fn pledge(promises: ?[*:0]const u8, execpromises: ?[*:0]const u8) c_int;
 pub extern "c" fn unveil(path: ?[*:0]const u8, permissions: ?[*:0]const u8) c_int;
+
+pub extern "c" fn pthread_set_name_np(thread: std.c.pthread_t, name: [*:0]const u8) c_int;
+pub extern "c" fn pthread_get_name_np(thread: std.c.pthread_t, name: [*:0]u8, len: usize) c_int;
```
semarie, line 50: these functions are returning `void` on OpenBSD: you could see the [man page](https://man.openbsd.org/pthread_set_name_np).

lib/std/Thread.zig
```diff
@@ -76,6 +76,121 @@ pub fn spinLoopHint() void {
...
+            return std.mem.spanZ(buffer);
+        } else {
+            return null;
+        },
+        .windows => {
+            // TODO
+            return null;
+        },
+        .macos, .ios, .watchos, .tvos, .netbsd, .freebsd, .openbsd => if (use_pthreads) {
+            const err = c.pthread_getname_np(self.data.handle, buffer.ptr, maxNameLen);
```
semarie, line 184: `void` return for OpenBSD.

lib/std/Thread.zig
```diff
@@ -76,6 +76,121 @@ pub fn spinLoopHint() void {
...
+        .netbsd => if (use_pthreads) {
+            const err = c.pthread_setname_np(self.data.handle, name.ptr, null);
+            return switch (err) {
+                0 => {},
+                else => unreachable,
+            };
+        },
+        .freebsd, .openbsd => if (use_pthreads) {
+            // Use pthread_set_name_np for FreeBSD because the alias pthread_setname_np is FreeBSD 12.2+ only.
+            const err = c.pthread_set_name_np(self.data.handle, name.ptr);
```
semarie, line 132: `void` return for OpenBSD

semarie (changes requested):
The return type for functions for OpenBSD should be `void`. It will make it different from `.freebsd` for setting the name, and different from others for getting it.

lib/std/Thread.zig
```diff
@@ -76,6 +76,121 @@ pub fn spinLoopHint() void {
...
+                else => unreachable,
+            };
+        },
+        else => {},
+    }
+}
+
+pub fn getName(self: Thread, allocator: *std.mem.Allocator) !?[:0]const u8 {
+    // musl doesn't provide pthread_getname_np and there's no way to retrieve the thread id of an arbitrary thread.
+    if (use_pthreads and self.data.handle != getCurrentId() and comptime std.Target.current.abi.isMusl()) {
```
LemonBoy, line 144: Nope, let this case use the `prctl`/`proc` code path.

It seems that [musl](https://www.openwall.com/lists/musl/2019/07/17/3) needs a bit of prodding on this front.

lib/std/Thread.zig
```diff
@@ -76,6 +76,121 @@ pub fn spinLoopHint() void {
...
+        return null;
+    }
+
+    var buffer = try allocator.allocWithOptions(u8, maxNameLen - 1, null, 0);
```
LemonBoy, line 148: What if `maxNameLen` is zero?
Wouldn't it make more sense to have `maxNameLen` exclude the tail zero?

lib/std/Thread.zig
```diff
@@ -76,6 +76,121 @@ pub fn spinLoopHint() void {
...
+    errdefer allocator.free(buffer);
+    std.mem.set(u8, buffer, 0);
+
+    switch (std.Target.current.os.tag) {
+        .linux => if (use_pthreads and comptime std.Target.current.abi.isGnu()) {
+            const err = c.pthread_getname_np(self.data.handle, buffer.ptr, maxNameLen);
+            return switch (err) {
+                0 => std.mem.spanZ(buffer),
+                os.ERANGE => error.NameTooLong,
```
LemonBoy, line 157: Given the hardcoded upper limit on the length is `maxNameLen` this case can be marked as `unreachable`, no?

lib/std/Thread.zig
```diff
@@ -76,6 +76,121 @@ pub fn spinLoopHint() void {
...
+                else => unreachable,
+            };
+        } else if (self.data.handle == getCurrentId()) {
+            _ = try os.prctl(.GET_NAME, .{@ptrToInt(buffer.ptr)});
```
LemonBoy, line 161: Error checking.

lib/std/Thread.zig
```diff
@@ -76,6 +76,121 @@ pub fn spinLoopHint() void {
...
+
+            const file = try std.fs.cwd().openFile(path, .{});
+            defer file.close();
+
+            const read = try file.reader().readAll(buffer);
+
+            // Remove the newline character
+            buffer[read - 1] = 0;
+
+            return std.mem.spanZ(buffer);
```
LemonBoy, line 175: This can be simplified (and perhaps a check for `read == 0` should be added).

```suggestion
            return buffer[0 .. read - 1];
```

lib/std/Thread.zig
vrischmann: I'm not sure I understand. Without this check, it would execute the `proc` reading code but I don't have access to the thread id of an arbitrary `pthread_t` so that won't work.

LemonBoy: Welp, I forgot the `pthread` structure type is opaque. You'd need to dig into the private definition in `pthread_impl.h` but that's pretty ugly... let's return an error instead (`Unsupported` as done for the darwin case maybe?)

vrischmann: Ok, I'll change it to return an error.

lib/std/Thread.zig
vrischmann: Thanks, these will be fixed.

lib/std/Thread.zig
vrischmann: This doesn't work:
```
./lib/std/Thread.zig:186:26: error: expected type '[:0]const u8', found '[]u8'
            return buffer[0 .. read - 1];
                         ^
./lib/std/Thread.zig:186:26: note: destination pointer requires a terminating '0' sentinel
            return buffer[0 .. read - 1];
```

I can do this however:
```
            buffer[read - 1] = 0;

            return buffer[0 .. read - 1 :0];
```
which might be better than using `spanZ`.

lib/std/Thread.zig
vrischmann: My thinking was it would be less surprising for a reader to see the maxNameLen with the same value as in the different man pages, but I'm not particularly attached to that. I'll change it.

LemonBoy: From the Zig user POV the interesting part is the number of characters one can use, the presence of the trailing zero is an implementation detail that leaks to the user.
Moreover you get an extra byte when allocating a buffer using `var buf: [maxNameLen:0]`, dropping the trailing zero is IMO justified.

One more idea, what about a pure Zig API where the user passes in a `[]const u8` slice? Zero-terminating the string on the fly here is pretty cheap and would allow for a more idiomatic API.

lib/std/Thread.zig
LemonBoy: Indeed, I missed the `:0`.
But do we really want to return a zero-terminated slice instead of a "classic" one?

lib/std/Thread.zig
vrischmann: (`maxNameLen` changed in the latest commit).

Not sure I get your last point, do you mean something like `fn getName(self: Thread, buffer: []u8) []const u8` ?

lib/std/Thread.zig
vrischmann: Yeah you're right, we could just return a `[]const u8`.

vrischmann:
I've addressed most of the comments in the latest commit (minus those still in discussion).

I successfully tested this on FreeBSD, however I can't test on NetBSD or OpenBSD because of compilation errors somewhere else in the stdlib.

lib/std/Thread.zig
LemonBoy: Not quite, I was thinking of something like
```zig
fn getName(self: Thread, allocator: *std.mem.Allocator) !?[]const u8;
fn setName(self: Thread, name: []const u8) !void;
```

where we get rid of the sentinel.

vrischmann: Ah right. Yeah I can do that.

LemonBoy: Keep in mind that Windows `SetThreadDescription` requires a UTF16-LE encoded string so we may need an allocator for `setName` or set an arbitrary upper limit and allocate a static buffer.

vrischmann:
I've added the windows implementation. 

One thing I'm not sure about is the utf8 -> utf16 conversion, right now I'm doing this:
```
           var name_buf_w: [maxNameLen:0]u16 = undefined;
           const length = try std.unicode.utf8ToUtf16Le(&name_buf_w, name);
```
The utf8 `name` can't be bigger than `maxNameLen` so I was assuming a `[maxNameLen:0]u16` is always enough to fit the utf16 data but I'm not sure if that's correct, I'm not super familiar with UTF16.

lib/std/Thread.zig
```diff
@@ -76,6 +76,188 @@ pub fn spinLoopHint() void {
...
+                return error.Unsupported;
+            }
+
+            var name_buf_w: [maxNameLen:0]u16 = undefined;
+            const length = try std.unicode.utf8ToUtf16Le(&name_buf_w, name);
+            name_buf_w[length] = 0;
+
+            try os.windows.SetThreadDescription(
+                self.data.handle,
+                std.meta.cast(os.windows.LPWSTR, name_buf_w[0..length :0]),
```
LemonBoy, line 135: Nit, `meta.cast` is not supposed to be used outside of `translate-c`'d code. A `@ptrCast` is enough here.

lib/std/Thread.zig
```diff
@@ -76,6 +76,188 @@ pub fn spinLoopHint() void {
...
+
+            const err = c.pthread_getname_np(self.data.handle, buffer.ptr, maxNameLen + 1);
+            return switch (err) {
+                0 => std.mem.spanZ(buffer),
+                os.EINVAL => unreachable,
+                os.ESRCH => unreachable,
+                else => return os.unexpectedErrno(err),
+            };
+        },
+        .freebsd => if (use_pthreads) {
```
LemonBoy, line 243: You can merge this with the `openbsd` case below.

lib/std/Thread.zig
```diff
@@ -572,6 +754,82 @@ pub fn getCurrentThreadId() u64 {
...
+    comptime const max_length_name = blk: {
+        var tmp: [maxNameLen + 1]u8 = undefined;
+        std.mem.set(u8, &tmp, 'b');
+        tmp[maxNameLen] = 0;
+        break :blk tmp[0..maxNameLen :0];
+    };
+
+    comptime const testCases = &[_][]const u8{
+        "mythread",
+        max_length_name,
```
LemonBoy, line 768: Won't `"b" ** maxNameLen` work?

lib/std/Thread.zig
```diff
@@ -572,6 +754,82 @@ pub fn getCurrentThreadId() u64 {
...
+
+fn testThreadNameUnsupported(thread: *Thread) !void {
+    const res = thread.setName("foobar");
+    std.testing.expectError(error.Unsupported, res);
+
+    const res2 = thread.getName(std.testing.allocator);
+    std.testing.expectError(error.Unsupported, res2);
+}
+
+test "Thread: setName, getName" {
```
LemonBoy, line 790: No need to prefix the test names.

lib/std/Thread.zig
```diff
@@ -572,6 +754,82 @@ pub fn getCurrentThreadId() u64 {
...
+    if (builtin.single_threaded) return error.SkipZigTest;
+
+    const Context = struct {
+        done: std.atomic.Bool = std.atomic.Bool.init(false),
+
+        pub fn run(ctx: *@This()) !void {
+            while (!ctx.done.load(.SeqCst)) {
```
LemonBoy, line 797: I'd suggest to perform the `getName` and `setName` within the thread itself, otherwise you're not testing anything on darwin systems.

lib/std/Thread.zig
```diff
@@ -572,6 +754,82 @@ pub fn getCurrentThreadId() u64 {
...
+
+    var context = Context{};
+    var thread = try spawn(Context.run, &context);
+
+    switch (std.Target.current.os.tag) {
+        .macos, .ios, .watchos, .tvos => {
+            const res = thread.setName("foobar");
+            std.testing.expectError(error.Unsupported, res);
+        },
+        .windows => if (std.Target.current.os.isAtLeast(.windows, .win10_rs1)) |res| {
```
LemonBoy, line 811: `isAtLeast` returns `null` if a runtime-check is required, I'd simply run `testThreadName` and catch `unsupported`, in that case you can return `SkipZigTest` as that's not an error condition.

lib/std/Thread.zig
semarie: it was splitted because freebsd function is returning `int` and openbsd function is returning `void` (previously the return value was checked).

but as it, freebsd case doesn't check for `pthread_get_name_np` return value.

lib/std/Thread.zig
vrischmann: Ok, I never realized that.

lib/std/Thread.zig
vrischmann: I opted to keep them split right now mainly to add the FreeBSD comment, but I can also add the comment in the `freebsd, openbsd` case.

lib/std/Thread.zig
MasterQ32: Imho, a good ol `&name_buf_w` should do the job here, no need to slice (we know it's NUL terminated anyways) and the function seems to accept a `[*:0]const u16`.

lib/std/Thread.zig
```diff
@@ -76,6 +76,188 @@ pub fn spinLoopHint() void {
     }
 }
 
+const maxNameLen = switch (std.Target.current.os.tag) {
+    .linux => 15,
+    .windows => 31,
```
MasterQ32, line 81: Where does this limit come from? I've read the documentation about SetThreadDescription and it doesn't mention a 31 (or 32) as a maximum length

vrischmann: It's arbitrary, I chose something which seems appropriate for a thread name.

PhaseMage:
> One thing I'm not sure about is the utf8 -> utf16 conversion, right now I'm doing this:
> 
> ```
>            var name_buf_w: [maxNameLen:0]u16 = undefined;
>            const length = try std.unicode.utf8ToUtf16Le(&name_buf_w, name);
> ```
> 
> The utf8 `name` can't be bigger than `maxNameLen` so I was assuming a `[maxNameLen:0]u16` is always enough to fit the utf16 data but I'm not sure if that's correct, I'm not super familiar with UTF16.

This is correct. A string of 'n' UTF16 code units can always hold 'n' UTF8 code units. 

The normal issue is only the other way around, where one UTF16 code unit might require multiple UTF8 code units of storage. For ASCII code points, this isn't relevant. But for larger code points (eg. Emoji), this can be important. I suspect that Windows supports a maximum number of UTF16 code units. Since you allocate to implement getName(), this works fine. However, limiting maxNameLen to 31 UTF8 code units artificially limits the number of code points you can give to setName() (say, a string of Emoji). But using Emoji in thread names is probably a bad idea anyway since debuggers might not support it well :-)

vrischmann:
> > One thing I'm not sure about is the utf8 -> utf16 conversion, right now I'm doing this:
> > ```
> >            var name_buf_w: [maxNameLen:0]u16 = undefined;
> >            const length = try std.unicode.utf8ToUtf16Le(&name_buf_w, name);
> > ```
> > 
> > 
> > The utf8 `name` can't be bigger than `maxNameLen` so I was assuming a `[maxNameLen:0]u16` is always enough to fit the utf16 data but I'm not sure if that's correct, I'm not super familiar with UTF16.
> 
> This is correct. A string of 'n' UTF16 code units can always hold 'n' UTF8 code units.
> 
> The normal issue is only the other way around, where one UTF16 code unit might require multiple UTF8 code units of storage. For ASCII code points, this isn't relevant. But for larger code points (eg. Emoji), this can be important. I suspect that Windows supports a maximum number of UTF16 code units. Since you allocate to implement getName(), this works fine. However, limiting maxNameLen to 31 UTF8 code units artificially limits the number of code points you can give to setName() (say, a string of Emoji). But using Emoji in thread names is probably a bad idea anyway since debuggers might not support it well :-)

Thanks. I actually tested emojis in the thread name, Visual Studio 2019 supports them at least. 
I agree that a `maxNameLen` of 31 is arbitrary but in my opinion it should be enough, and it's in line with what other platforms do.
I could try and see if the call to `SetThreadDescription` fails with a bigger maxNameLen and get a better idea where the limit is, because I couldn't find it in the documentation.

lib/std/Thread.zig
```diff
@@ -70,6 +70,181 @@ else switch (std.Target.current.os.tag) {
...
+    .macos, .ios, .watchos, .tvos => 63,
+    .netbsd => 31,
+    .freebsd => 15,
+    .openbsd => 31,
+    else => 0,
+};
+
+pub fn setName(self: Thread, name: []const u8) !void {
```
andrewrk, line 83: This has different errors across different targets and so it must have an explicit error set (e.g. so that "Unsupported" shows up even on targets where it is always supported).

lib/std/Thread.zig
```diff
@@ -70,6 +70,181 @@ else switch (std.Target.current.os.tag) {
...
+            // Use pthread_set_name_np for FreeBSD because pthread_setname_np is FreeBSD 12.2+ only.
+            // TODO maybe revisit this if depending on FreeBSD 12.2+ is acceptable because pthread_setname_np can return an error.
+
+            c.pthread_set_name_np(self.data.handle, name_with_terminator.ptr);
+        },
+        else => return error.Unsupported,
+    }
+}
+
+pub fn getName(self: Thread, allocator: *std.mem.Allocator) !?[]const u8 {
```
andrewrk, line 164: I suggest to take an array buffer of `maxNameLen` length, and then return a slice into that buffer. This will avoid the need for an allocator.

If you want you can then provide a higher level API, `getNameAlloc`, implemented using the lower level API.

andrewrk (changes requested):
Thanks for your patience; I know it took a long time to get a review from me.

I have a couple requests and then I think this looks good to merge.

Can you please rebase against master branch and fix the conflicts in addition to the requested changes?

lib/std/Thread.zig
vrischmann: Addressed in the latest commit

lib/std/Thread.zig
vrischmann: Addressed in the latest commit

lib/std/Thread.zig
```diff
@@ -38,6 +38,180 @@ else
 
 impl: Impl,
 
+pub const maxNameLen = switch (std.Target.current.os.tag) {
```
andrewrk, line 41: This is public so let's make it conform to the [style guide](https://ziglang.org/documentation/master/#Style-Guide):

```suggestion
pub const max_name_len = switch (std.Target.current.os.tag) {
```

lib/std/Thread.zig
```diff
@@ -38,6 +38,180 @@ else
...
+            // Use pthread_set_name_np for FreeBSD because pthread_setname_np is FreeBSD 12.2+ only.
+            // TODO maybe revisit this if depending on FreeBSD 12.2+ is acceptable because pthread_setname_np can return an error.
+
+            c.pthread_set_name_np(self.getHandle(), name_with_terminator.ptr);
+        },
+        else => return error.Unsupported,
+    }
+}
+
+pub fn getName(self: Thread, buffer_ptr: *[maxNameLen:0]u8) !?[]const u8 {
```
andrewrk, line 138: This also needs an explicit error set to make it the same across targets.

andrewrk (changes requested):
Thanks for the fixups! Found 2 more things - sorry for not finding them last time, but at least now I am paying close attention, and will give this a prompt merge :)

vrischmann:
Ok, addressed your last feedback.

I had to change some logic for musl since the type returned by `getCurrentId` changed but it's working now.
I've only been able to test linux/windows, I can test freebsd later.
I don't have a mac to test on macOS unfortunately.

andrewrk (approved):

vrischmann:
Pushed the last fixes for macos, freebsd/netbsd.

PR #8571, opened 2021-04-18
std.build: if using a RunStep, show the command run on verbose

g-w1:
This was useful to help me debug something.

### DIFF
--- a/lib/std/build/run.zig
+++ b/lib/std/build/run.zig
@@ -191,6 +191,13 @@ pub const RunStep = struct {
         child.stdout_behavior = stdIoActionToBehavior(self.stdout_action);
         child.stderr_behavior = stdIoActionToBehavior(self.stderr_action);
 
+        if (self.builder.verbose) {
+            for (argv) |arg| {
+                warn("{s} ", .{arg});
+            }
+            warn("\n", .{});
+        }
+
         child.spawn() catch |err| {
             warn("Unable to spawn {s}: {s}\n", .{ argv[0], @errorName(err) });
             return err;


lib/std/build/run.zig
```diff
@@ -191,6 +191,13 @@ pub const RunStep = struct {
         child.stdout_behavior = stdIoActionToBehavior(self.stdout_action);
         child.stderr_behavior = stdIoActionToBehavior(self.stderr_action);
 
+        if (self.builder.verbose) {
```
LemonBoy, line 194:
```suggestion
        if (self.builder.verbose and argv.len > 0) {
```

g-w1: I dont think this is needed. https://github.com/ziglang/zig/blob/fbda9991f48a6cd3a008890634ba4482f946a6f4/lib/std/build.zig#L329

jedisct1:
That can indeed be very useful!

Just a purely cosmetic suggestion: maybe print something before the command (such as a `$ ` or `> ` prompt). Just to make it easier to visually distinguish these commands from other messages. Also useful for log coloring.

g-w1:
Regular build --verbose does not show a $, so I am not sure why you would want one here: 
```
❯ zig build --verbose
/home/jacob/.local/zig/0.8.0-dev.2065+bc06e1982/files/zig build-exe /tmp/b/src/main.zig --cache-dir /tmp/b/zig-cache --global-cache-dir /home/jacob/.cache/zig --name b --enable-cache
cp /tmp/b/zig-cache/o/6c7dd7e30f3b233783c6a79242c73110/b /tmp/b/zig-cache/bin/b # installed
```

jedisct1:
Right, scratch that. Not the fact that this is a useful addition :)

PR #8572, opened 2021-04-18
std: Add process_vm_readv/writev wrappers

chivay:
Hi,
this patch adds wrappers for process_vm_readv/writev syscalls on Linux.

### DIFF
--- a/lib/std/os/linux.zig
+++ b/lib/std/os/linux.zig
@@ -1410,6 +1410,30 @@ pub fn pidfd_send_signal(pidfd: fd_t, sig: i32, info: ?*siginfo_t, flags: u32) u
     );
 }
 
+pub fn process_vm_readv(pid: pid_t, local: [*]const iovec, local_count: usize, remote: [*]const iovec, remote_count: usize, flags: usize) usize {
+    return syscall6(
+        .process_vm_readv,
+        @bitCast(usize, @as(isize, pid)),
+        @ptrToInt(local),
+        local_count,
+        @ptrToInt(remote),
+        remote_count,
+        flags,
+    );
+}
+
+pub fn process_vm_writev(pid: pid_t, local: [*]const iovec, local_count: usize, remote: [*]const iovec, remote_count: usize, flags: usize) usize {
+    return syscall6(
+        .process_vm_writev,
+        @bitCast(usize, @as(isize, pid)),
+        @ptrToInt(local),
+        local_count,
+        @ptrToInt(remote),
+        remote_count,
+        flags,
+    );
+}
+
 test {
     if (builtin.os.tag == .linux) {
         _ = @import("linux/test.zig");


ifreund (commented):
Why is flags a `u64`? My man page says it is an `unsigned long` and thus 32 bits on `i386` right? It seems to me that it should be a `usize` like the rest.

chivay:
>Why is flags a u64? My man page says it is an unsigned long and thus 32 bits on i386 right? It seems to me that it should be a usize like the rest.

Yes, you're right. It should be fixed now.

lib/std/os/linux.zig
```diff
@@ -1410,6 +1410,32 @@ pub fn pidfd_send_signal(pidfd: fd_t, sig: i32, info: ?*siginfo_t, flags: u32) u
...
+pub fn process_vm_readv(pid: pid_t, local: [*]const iovec, local_count: usize, remote: [*]const iovec, remote_count: usize, flags: usize) usize
+{
+    return syscall6(
+        .process_vm_readv,
+        @bitCast(usize, @as(isize, pid)),
+        @ptrToInt(local),
+        local_count,
+        @ptrToInt(remote),
+        remote_count,
+        flags
```
ifreund, line 1422: without the trailing comma, `zig fmt` will put this all on one line.
```suggestion
        flags,
```

lib/std/os/linux.zig
```diff
@@ -1410,6 +1410,32 @@ pub fn pidfd_send_signal(pidfd: fd_t, sig: i32, info: ?*siginfo_t, flags: u32) u
...
+pub fn process_vm_writev(pid: pid_t, local: [*]const iovec, local_count: usize, remote: [*]const iovec, remote_count: usize, flags: usize) usize
+{
+    return syscall6(
+        .process_vm_writev,
+        @bitCast(usize, @as(isize, pid)),
+        @ptrToInt(local),
+        local_count,
+        @ptrToInt(remote),
+        remote_count,
+        flags
```
ifreund, line 1435:
```suggestion
        flags,
```

lib/std/os/linux.zig
```diff
@@ -1410,6 +1410,32 @@ pub fn pidfd_send_signal(pidfd: fd_t, sig: i32, info: ?*siginfo_t, flags: u32) u
     );
 }
 
+pub fn process_vm_readv(pid: pid_t, local: [*]const iovec, local_count: usize, remote: [*]const iovec, remote_count: usize, flags: usize) usize
+{
```
ifreund, line 1414: zig fmt wants the curly on the same line as the signature.
```suggestion
pub fn process_vm_readv(pid: pid_t, local: [*]const iovec, local_count: usize, remote: [*]const iovec, remote_count: usize, flags: usize) usize {
```

lib/std/os/linux.zig
```diff
@@ -1410,6 +1410,32 @@ pub fn pidfd_send_signal(pidfd: fd_t, sig: i32, info: ?*siginfo_t, flags: u32) u
...
+        @ptrToInt(local),
+        local_count,
+        @ptrToInt(remote),
+        remote_count,
+        flags
+    );
+}
+
+pub fn process_vm_writev(pid: pid_t, local: [*]const iovec, local_count: usize, remote: [*]const iovec, remote_count: usize, flags: usize) usize
+{
```
ifreund, line 1427:
```suggestion
pub fn process_vm_writev(pid: pid_t, local: [*]const iovec, local_count: usize, remote: [*]const iovec, remote_count: usize, flags: usize) usize {
```

chivay:
Applied zig fmt

ifreund (approved):
LGTM, just waiting on CI now.

PR #8580, opened 2021-04-20
std: Fix thread creation with field-less context type

### LINKED ISSUE #8524: Struct with no fields is incorrectly treated as void
The following code fails to compile:
```zig
 const std = @import("std");
 const Thread = std.Thread;

 const Context = struct { };

 fn runWorker(ctx: Context) void {
     std.debug.print("in thread\n", .{});
 }

pub fn main() !void {
    var t = try Thread.spawn(runWorker, Context{});
    t.wait();
}
```
   
It gives the error:
```
./zig-linux-x86_64-0.8.0-dev.1825+d2d4df407/lib/std/Thread.zig:270:29: error: expected type 'Context', found 'void'
                    startFn(arg);
                            ^
./pool.zig:8:17: note: Context declared here
const Context = struct {};
```

[Link to godbolt](https://godbolt.org/z/b4xx3najb)

If I add a field to the struct, it [compiles](https://godbolt.org/z/6v6jcec81):

```zig
 const std = @import("std");
 const Thread = std.Thread;

 const Context = struct { foo: bool = false };

 fn runWorker(ctx: Context) void {
     std.debug.print("in thread\n", .{});
 }

pub fn main() !void {
    var t = try Thread.spawn(runWorker, Context{});
    t.wait();
}
```

This seems wrong and I asked in the #zig-help channel on discord and some people said it was likely a bug.

LemonBoy:
Closes #8524

### DIFF
--- a/lib/std/Thread.zig
+++ b/lib/std/Thread.zig
@@ -199,7 +199,8 @@ pub fn spawn(comptime startFn: anytype, context: SpawnContextType(@TypeOf(startF
                 inner: Context,
             };
             fn threadMain(raw_arg: windows.LPVOID) callconv(.C) windows.DWORD {
-                const arg = if (@sizeOf(Context) == 0) {} else @ptrCast(*Context, @alignCast(@alignOf(Context), raw_arg)).*;
+                const arg = if (@sizeOf(Context) == 0) undefined //
+                else @ptrCast(*Context, @alignCast(@alignOf(Context), raw_arg)).*;
 
                 switch (@typeInfo(@typeInfo(@TypeOf(startFn)).Fn.return_type.?)) {
                     .NoReturn => {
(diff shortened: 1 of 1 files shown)


lib/std/Thread.zig
```diff
@@ -199,7 +199,9 @@ pub fn spawn(comptime startFn: anytype, context: SpawnContextType(@TypeOf(startF
                 inner: Context,
             };
             fn threadMain(raw_arg: windows.LPVOID) callconv(.C) windows.DWORD {
-                const arg = if (@sizeOf(Context) == 0) {} else @ptrCast(*Context, @alignCast(@alignOf(Context), raw_arg)).*;
+                const arg = if (Context == void) {} //
+                else if (@sizeOf(Context) == 0) .{} //
```
andrewrk, line 203: This works for 0 size structs but not for `u0`, enums with 1 tag, etc. Seems like there should be a simpler fix that only checks if `@sizeOf` is 0.

andrewrk (changes requested):

lib/std/Thread.zig
LemonBoy: Ah, good catch. Let's use `undefined` instead? That should work for every zero-sized type.

andrewrk: Sounds good

PR #8584, opened 2021-04-20
Provide method to set logging level per scope

leecannon:
This PR adds a way to override the logging level used by `std.log` on a scope by scope basis.

In order to do so the user defines a public `scope_levels` in the root file as per below:
```zig
pub const scope_levels = [_]std.log.ScopeLevel{
    .{ .scope = .liba, .level = .crit },
    .{ .scope = .memory, .level = .info },
    .{ .scope = .libds .level = .warn },
};
```

If a scope level is provided by the user for the scope in question then the provided level is used to determine if the message is logged and the global log level is ignored. If the scope is not within `scope_levels` then the global log level is used.

### DIFF
--- a/lib/std/log.zig
+++ b/lib/std/log.zig
@@ -117,13 +117,30 @@ pub const level: Level = if (@hasDecl(root, "log_level"))
 else
     default_level;
 
+pub const ScopeLevel = struct {
+    scope: @Type(.EnumLiteral),
+    level: Level,
+};
+
+const scope_levels = if (@hasDecl(root, "scope_levels"))
+    root.scope_levels
+else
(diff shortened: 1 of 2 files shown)


daurnimator (commented):
Could you add a test case?

leecannon:
Without overriding the log function in the root how would i test it?
As far as i know there is currently no way of capturing log messages in tests.

marler8997:
@leecannon you could use the standalone tests I think

leecannon:
I've added a test case to the compare_output tests.

marler8997:
At a quick glance this seems like an obvious feature to be able to change the log level for each scope individually.

PR #8585, opened 2021-04-20
target: drop `ppc32` and prefer `ppc`

### LINKED ISSUE #7947: Cross-compiling to PowerPC with `zig cc` is broken
Currently, cross-compiling for PPC32 seems broken:
`zig cc -c foo.c -target powerpc-freestanding`

Fails with `error: unknown target CPU 'ppc32'`

As discussed on IRC, a one-liner fix was written by @mikdusan and is available at:
https://gist.github.com/mikdusan/b772d2d6ea592567722ae73870f88519

mikdusan:
- original PR #7949 (incorrectly) patched a generated-file and changes
  have subsequently been lost/overwritten
- fix #7947 in a different way: drop `ppc32` because `ppc` already exists

cc @LemonBoy


LemonBoy:
FYI the underlying problem was fixed [with this Clang patch](https://reviews.llvm.org/D100933).

PR #8586, opened 2021-04-20
A few more PPC32 touch-ups

LemonBoy:
Let's add it to the testing plan! (and bump its tier?)

**Bikeshedding moment**

At the moment we have a few `isXXX` functions defined for the `Arch` enum, some of them refer to the whole "family" (eg. `isMIPS` is true for both 32 and 64 bit targets) while other are split (`isPPC` and `isPPC64`).
If we were to harmonize this API what's your favourite pick between:
* Separate `isARCH` and `isARCH64` methods
* A single `isARCH(enum{ 32, 64, both})` method
* A single `isARCH` method that returns true for both 32 and 64 bit variants, leaving the user to discriminate on `ptrBitWidth`
* A single `isOneOf([]Arch)` and `isAnyOf([]Arch)` methods, generalizing the problem
* Something else?

### DIFF
--- a/lib/std/os/bits/linux/powerpc.zig
+++ b/lib/std/os/bits/linux/powerpc.zig
@@ -557,18 +557,10 @@ pub const kernel_stat = extern struct {
     size: off_t,
     blksize: blksize_t,
     blocks: blkcnt_t,
-    __atim32: timespec32,
-    __mtim32: timespec32,
-    __ctim32: timespec32,
-    __unused: [2]u32,
     atim: timespec,
     mtim: timespec,
     ctim: timespec,
-
-    const timespec32 = extern struct {
-        tv_sec: i32,
-        tv_nsec: i32,
-    };
+    __unused: [2]u32,
 
     pub fn atime(self: @This()) timespec {
         return self.atim;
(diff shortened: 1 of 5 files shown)


andrewrk:
My preference:
 * No function when there is only 1 enum tag. Just check if the enum tag matches.
 * No parameters to the `isFoo` helper functions. If the logic is complicated, do it at the callsite.
 * No helper functions without callsites. If they're not needed, forget it

Regarding how to follow the style guide for casing function names, here is the algorithm:
 1. write the name in `lower_snake_case`
 2. convert the `_` into upper casing the next letter

This makes it pretty clear how to name functions. `is_ppc` would be `isPpc`. `is_riscv` would be `isRiscv`. Lots of violations in the std lib, we'll get them eventually.

LemonBoy:
> is_ppc would be isPpc. is_riscv would be isRiscv. Lots of violations in the std lib, we'll get them eventually.

Acronyms, short forms (eg. PowerPC -> PPC) and uppercase names (eg. RISC-V, MIPS) should be exempt from the first-letter-is-capitalized rule, `isPpc` or `isRiscv` or `isMips` all look pretty bad to me.

PR #8587, opened 2021-04-20
stage2: make std.fmt.parseInt ignore `_`

g-w1:
This allows more astgen stuff. Also make it compile w/o skip-non-native.

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -1358,6 +1358,7 @@ pub fn Formatter(comptime format_fn: anytype) type {
 ///  * A prefix of "0x" implies radix=16,
 ///  * Otherwise radix=10 is assumed.
 ///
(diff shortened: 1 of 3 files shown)


andrewrk:
Test coverage please. Also suggested commit message for the second one: "fixed compile error in codegen" or something like that, (ninja is not really relevant to the problem)

lib/std/fmt.zig
```diff
@@ -1448,6 +1455,7 @@ fn parseWithSign(
     var x: T = 0;
 
     for (buf_start) |c| {
+        if (c == '_') continue;
```
LemonBoy, line 1458: This is not enough as it will happily parse `_123` and `123_`, you need a check (and tests) to make sure the character is not the first nor the last in `buf_start`.

Issue #7512, opened 2020-12-20
Even more integers but stricter integer operators


RogierBrussee:
Zig already has 20 different basic integer classes.  It seems Zig inherited from C that an integer is characterised by signedness, size, and it adds a a further attribute "length is determined by C-ABI, explicit, or pointer size".  On these integers the main operators are

+, -, *, /  (can overflow)
%+, %-, %* %/ (does not overflow)
&, | , ^, <<, >>
==, !=, <, <=, >, >=
! 
%
 
lots of @functions including arithmetic with explicit overflow e.g.

@addOverflow(T; type, a:T, b:T, r:*T) bool  

This is clearly inherited from C with the exception of the modular operator %+, %-, %*, which are Zig specific

I propose to separate (potentially overflowing) arithmetic from modular arithmetic and from bitwise operators of different sizes and let them be defined on their own integer type

i8/16/32/64/128   --> +, - , *, ! , @divTrunc(), @divFloor(), @rem(), @mod(),  ==, !=, <, <=, >, >=  (signed  comparison)
u8/16/32/64/128  --> +, - , *, /,  ! , %, ==, !=, <, <=, >, >=, (unsigned comparison)
m8/16/32/64/128 --> %+, %-, % *,  ==, != , !

Update: % as an operation from say m32 -> m32 makes no sense for modular ints, 
but conversions like m32 --> m17  are well defined and harmless and can be expressed with @as(m17, n)
Update: modular division is  slightly subtle see open issues.

b8/12/32/64/128 --> &, |, ~,  ^ ,  <<, >>,  @shra(),@rotl(),rotr(), (+ more bitops),  ==, !=, !
  
Mathematically this is natural: it makes a distinction between the integers **Z** which are approximated by i64 (in the same sense that real numbers are approximated by f64), the integers modulo 2^64  **Z**/2^64**Z** which can be exactly represented by the 64 bit in m64, and a 64 bit bit vectors bitvectors {0,1}^64. 

Some open issues:
naming: 
m8/16/32... , b8/16/32/... is very minimalistic, explicit  modular8/16/32/.. and bits8/16/..can also be argued for.

Notation:
Is there still a need for the notation %+, %-, %* when the the type already makes explicit that modular arithmetic is asked for.
This is less explicit but not different from the overloading of +,-, ... between the different integer types and floating point.

Intermediate types and (in)equality
[...]

Issue #1918, opened 2019-02-05
runtime safety checking for pointer arithmetic


andrewrk:
There should be a test in test/runtime_safety.zig to make sure that unknown length pointer arithmetic does not wrap past the end of the address space.

andrewrk:
Extracted from https://github.com/ziglang/zig/issues/1059#issuecomment-462101897

 * [ ] compile error test for compile time pointer arithmetic overflowing
 * [ ] runtime safety test for pointer arithmetic overflowing
 * [ ] compile error test for comptime ptr arithmetic with undefined
 * [ ] docs

These apply to unknown length pointers and C pointers.

wooster0:
Should also check that the resulting pointer if non-`allowzero` doesn't result in zero like it can currently:
```zig
const std = @import("std");

pub fn main() void {
    const x: [*]u8 = @ptrFromInt(1);
    const a: [*]u8 = x - 1;
    std.debug.print("address: {*}\n",.{a});
}
```
```
address: u8@0
```

silversquirl:
Just ran into this when calculating the index of a pointer to an array element:
```zig
var a: [32]u8 = undefined;
const x = &a[7];
const index = &a - x; // oops! this should be `x - &a`
std.debug.print("{}\n", .{index}); // outputs "18446744073709551609"
```
I got the order backwards by accident, and it took a second to figure out what was going on. Overflow didn't immediately come to mind because I'm usually so spoilt by Zig's safety checks! ^-^

PR #8589, opened 2021-04-21
translate-c: handle negative array indices

### LINKED ISSUE #8556: translate-c: negative array indices don't work
In C, array indexing is just shorthand for pointer arithmetic, but in Zig array subscripts are required to be unsigned. That means any C code which uses a negative array index (uncommon, but legal) can't be translated directly to a Zig array indexing operation.

From the C standard
> The deﬁnition of the subscript operator [] is that E1[E2] is identical to (*((E1)+(E2))).

The if statement in this valid C code 
```c
#include <stdlib.h>
int main(void) {
	int foo[] = {0,1,2,3,4};
	int *bar = foo + 2;
	if (bar[-1] != 1) abort();
	return 0;
}
```

produces this line when translated (`error: attempt to cast negative value to unsigned integer`):
```zig
if (bar[@intCast(c_uint, -@as(c_int, 1))] != @as(c_int, 1)) {
```

One solution would be to always translate C array index operations to Zig pointer arithmetic:
```zig
if ((bar + @bitCast(usize, @as(isize, -@as(c_int, 1)))).* != @as(c_int, 1)) {
```
or bitcast indices to `usize`:
```zig
if (bar[@bitCast(usize, @as(isize, -@as(c_int, 1)))] != @as(c_int, 1)) {
```
Which is pretty ugly but I think guaranteed to work. 

Happy to implement either solution (or something better if there are other ideas) if people think it's worthwhile.

ehaas:
Take advantage of wrapping pointer arithmetic to enable negative array
subscripts.

Fixes #8556

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -3273,30 +3273,113 @@ fn transMemberExpr(c: *Context, scope: *Scope, stmt: *const clang.MemberExpr, re
     return maybeSuppressResult(c, scope, result_used, node);
 }
 
+/// ptr[subscr] (`subscr` is a signed integer expression, `ptr` a pointer) becomes:
+/// (blk: {
+///     const tmp = subscr;
+///     if (tmp >= 0) break :blk ptr + @intCast(usize, tmp) else break :blk ptr - ~@bitCast(usize, @intCast(isize, tmp) +% -1);
+/// }).*
+/// Todo: rip this out once `[*]T + isize` becomes valid.
+fn transSignedArrayAccess(
+    c: *Context,
+    scope: *Scope,
+    container_expr: *const clang.Expr,
+    subscr_expr: *const clang.Expr,
+    result_used: ResultUsed,
+) TransError!Node {
+    var block_scope = try Scope.Block.init(c, scope, true);
+    defer block_scope.deinit();
+
+    const tmp = try block_scope.makeMangledName(c, "tmp");
+
+    const subscr_node = try transExpr(c, &block_scope.base, subscr_expr, .used);
+    const subscr_decl = try Tag.var_simple.create(c.arena, .{ .name = tmp, .init = subscr_node });
+    try block_scope.statements.append(subscr_decl);
+
+    const tmp_ref = try Tag.identifier.create(c.arena, tmp);
+
+    const container_node = try transExpr(c, &block_scope.base, container_expr, .used);
+
+    const cond_node = try Tag.greater_than_equal.create(c.arena, .{ .lhs = tmp_ref, .rhs = Tag.zero_literal.init() });
+
+    const then_value = try Tag.add.create(c.arena, .{
+        .lhs = container_node,
+        .rhs = try Tag.int_cast.create(c.arena, .{
+            .lhs = try Tag.identifier.create(c.arena, "usize"),
+            .rhs = tmp_ref,
+        }),
+    });
+
+    const then_body = try Tag.break_val.create(c.arena, .{
+        .label = block_scope.label,
+        .val = then_value,
+    });
+
+    const minuend = container_node;
+    const signed_size = try Tag.int_cast.create(c.arena, .{
+        .lhs = try Tag.identifier.create(c.arena, "isize"),
+        .rhs = tmp_ref,
+    });
+    const to_cast = try Tag.add_wrap.create(c.arena, .{
+        .lhs = signed_size,
+        .rhs = try Tag.negate.create(c.arena, Tag.one_literal.init()),
+    });
+    const bitcast_node = try Tag.bit_cast.create(c.arena, .{
+        .lhs = try Tag.identifier.create(c.arena, "usize"),
+        .rhs = to_cast,
+    });
+    const subtrahend = try Tag.bit_not.create(c.arena, bitcast_node);
+    const difference = try Tag.sub.create(c.arena, .{
+        .lhs = minuend,
+        .rhs = subtrahend,
+    });
+    const else_body = try Tag.break_val.create(c.arena, .{
+        .label = block_scope.label,
+        .val = difference,
+    });
+
+    const if_node = try Tag.@"if".create(c.arena, .{
+        .cond = cond_node,
+        .then = then_body,
+        .@"else" = else_body,
+    });
+
+    try block_scope.statements.append(if_node);
+    const block_node = try block_scope.complete(c);
+
+    const derefed = try Tag.deref.create(c.arena, block_node);
+
+    return maybeSuppressResult(c, &block_scope.base, result_used, derefed);
+}
+
 fn transArrayAccess(c: *Context, scope: *Scope, stmt: *const clang.ArraySubscriptExpr, result_used: ResultUsed) TransError!Node {
-    var base_stmt = stmt.getBase();
+    const base_stmt = stmt.getBase();
+    const base_qt = getExprQualType(c, base_stmt);
+    const is_vector = cIsVector(base_qt);
+
+    const subscr_expr = stmt.getIdx();
+    const subscr_qt = getExprQualType(c, subscr_expr);
+    const is_longlong = cIsLongLongInteger(subscr_qt);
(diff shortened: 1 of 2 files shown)


LemonBoy:
> wrapping pointer arithmetic

Let me CC @SpexGuy here, what does the specification say about pointer arithmetic?
#1918 suggests the wrap-around is unintended and should be fixed (does the `+` vs `%+` distinction apply here?).

From a LLVM point of view we're codegen'ing pointer arithmetic using `inbound` GEPs and that's quite dangerous as we're potentially violating more than a single requirement:

```
If the inbounds keyword is present, the result value of the getelementptr is a poison value if one of the following rules is violated:

    The base pointer has an in bounds address of an allocated object, which means that it points into an allocated object, or to its end. The only in bounds address for a null pointer in the default address-space is the null pointer itself.
    If the type of an index is larger than the pointer index type, the truncation to the pointer index type preserves the signed value.
    The multiplication of an index by the type size does not wrap the pointer index type in a signed sense (nsw).
    The successive addition of offsets (without adding the base address) does not wrap the pointer index type in a signed sense (nsw).
    The successive addition of the current address, interpreted as an unsigned number, and an offset, interpreted as a signed number, does not wrap the unsigned address space and remains in bounds of the allocated object. As a corollary, if the added offset is non-negative, the addition does not wrap in an unsigned sense (nuw).
    In cases where the base is a vector of pointers, the inbounds keyword applies to each of the computations element-wise.
```

We'd be on the safer side by codegen'ing the non-wrapping arithmetic using non-inbounds GEPs and the wrapping one using `ptrtoint` + `inttoptr`, the latter is also "safer" from a provenance point of view (but comes at a higher cost, this kind of int<->ptr conversion inhibits some optimizations).

SpexGuy:
Good question. We haven't really thought about this yet.  I expect that in the end either `+%` on pointers will be unsupported or it will do a ptr-int-ptr conversion and allow breaking provenance.  In the meantime I would recommend implementing array access the way C does: `(ptr + offset).*`.

ehaas:
`(ptr + offset).*` won't work currently if offset is signed (e.g. ` error: expected type 'usize', found 'c_int'  note: unsigned 64-bit int cannot represent all possible signed 32-bit values`). I just want to double check that it's safe to cast the offset with `@bitCast(usize, @intCast(isize, offset))` before doing the addition? (this is already what we're doing for regular pointer arithmetic - see https://github.com/ziglang/zig/pull/8139). The alternative would be to check the value at runtime, absCast it, and add/subtract as appropriate.

SpexGuy:
Oh, my bad, yeah, looks like `ptr + isize` is a compile error.  This is a flaw in the language design, it means there are only two safe ways to do what you want:
```zig
// 1. check the sign, good for alias analysis, bad for everything else
res = if (offset >= 0) (ptr + @intCast(usize, offset)) else (ptr - @intCast(usize, -offset));
// 2. use integers, bad for alias analysis, good for everything else
res = @intToPtr(*T, @ptrToInt(ptr) +% @bitCast(usize, offset) * @sizeOf(T));
```
I'll bring this up at a design meeting, it's the same problem that prevents us from adding a signed value to an unsigned value without either giving up unsigned range or giving up overflow checks.

ehaas:
Cool, I can do option 1 for now. Should I also update the pointer arithmetic code from my previously-linked PR to use the same approach?

SpexGuy:
Yeah, that's probably safer until we figure out what `+%` actually should mean on pointers.

SpexGuy:
We talked about this at the design meeting, and decided
- `[*]T + isize` should be allowed, and overflow is checked UB if the result doesn't fit in `usize`
- `[*]T +% anything` should be a compile error

So the long term path here is to use `+`, but in the short term do what works.

LemonBoy:
>  should be allowed, and overflow is checked UB if the result doesn't fit in usize

I take this is also the case for a RHS of type `usize` and also for underflow.

andrewrk:
Hmmm. If we used `isize` for the pointer arithmetic type, then adding `ptr + byte_slice.len` would be a compile error. That's really unfortunate. And I don't think we can chop a bit off our usizes.

Maybe it makes sense to not do inbounds for pointer arithmetic. We would still be using inbounds for everything else.

I need to think about this for a while. Feel free to propose ideas.

SpexGuy:
Would it make sense to lower pointer addition with a usize as bitCast to isize + add?

LemonBoy:
There's a bit more than this, the use of inbound GEPs limits the addressable memory to the lower half and so array indices, `.len` are `u<address-space-width - 1>`.

RogierBrussee:
Without modular aka wrapping pointer arithmetic,  how are you going to translate 

extern char mem[4];

int main(){
    char* p = mem + 2;

    //The hardware just does modular aka wrapping arithmetic 
    //and that is therefore what the C compiler does 
    char* q = p +  ~(uintptr_t)0;
    assert (p - 1 == q);

    //perfectly unproblematic because q  == &mem[1]

    q[0] = 'a';
    return 0;
}

(https://godbolt.org/z/f1TY9d88x)


Of course this is a contrived example but someday people will write something like   

(char*)p  + (offset(struct foo, baz) - offsetof(struct foo, bar)) 

which (since all arithmetic with unsigned integers  in C is modular aka wrapping and the result of offsetof is an unsigned  size_t) means,  oops, the result of the the subtraction is a very big unsigned number.

Maybe we should pedantically have written

((char*)p  +  offset(struct foo, baz)) -  offsetof(struct foo, bar)

or 

(char*)p  +  (diff_t) ( offset(struct foo, baz) -  offsetof(struct foo, bar)).

but the hardware only knowns about modular aka wrapping arithmetic so the difference is harmless.

To define +% for pointers and  unsigned integers, if
 
ptr: [*]T;
uoffset : uXX 

then set

ptr +% uoffset == @intToPtr([*]T, @ptrToInt(ptr) +% @as(usize, offset) *% @sizeOf(T));
ptr -% uoffset == @intToPtr([*]T, @ptrToInt(ptr) -% @as(usize, offset) *% @sizeOf(T));

while in the signed case 

ptr: [*]T ;
soffset : iXX 

ptr +% soffset = ptr +% @bitCast(usize, @as(isize, offset))
ptr -% soffset = ptr -% @bitCast(usize, @as(isize, offset)).

and a modular pointer difference

ptr1, ptr2: [*]T;

ptr1 -% ptr2 ==  @divExact(@as(isize, @ptrToInt(ptr1) -% @ptrToInt(ptr2)), @sizeOf(T))  //preserve sign

Defined in this way the expected associativity properties hold for n,m:isize or n, m:usize 

ptr +% (n +% m) == (ptr +% n) +% m 
ptr +% (n -% m) == (ptr +% n) -% m  
ptr -% (n +% m) == (ptr -% n) -% m 
ptr -% (n -% m) == (ptr -% n) +% m 

Since _provided no overflow occurs_ we have for signed integers @as(isize, n + m) == @as(isize, n) + @as(isize,m ) (and similar for smaller unsigned integers with usize) we have for all integer types and _provided no overflow occurs_ 
 
ptr +% (n + m) == (ptr +% n) +% m

(Note that in terms of modular aka wrapping arithmetic, i.e. if we reduce the offset mod 2^{log2(@bitsize)} and if there existed modular (aka wrapping) integer type msize  (#7512) then one could simply define  

[*T] : ptr
offset: msize 

ptr +% offset == @intToPtr([*]T, @as(msize, @ptrToInt(ptr)) +% offset *% @sizeof(T))

and all the integer casting in the definitions above just amount to  @as(msize,    ) (i.e. reduction mod 2^{log2(@bitsize)}). 

Arguably, if you want to stress that pointer arithmetic is an intrinsically dangerous operation that you should try to avoid if you can, but that makes sense on the low, close to the hardware level, _only_ having modular aka wrapping pointer arithmetic +% (and -%,  and  pointer difference -% with values in msize, or in its absence isize)  may be the sane way to go.

ehaas:
The generated code is pretty wild but it works now (on my machine, we'll see what CI says). Hopefully with less undefined behavior.

Issue #6600, opened 2020-10-07
Eliminate the POSIX API layer


LemonBoy:
The `os` namespace (to be renamed `posix`) is tackling the problem of providing a cross-platform abstraction over the operating system facilities from a, in my opinion, wrong angle. In this essay I will briefly explain why.

The aim here is having a single level of abstraction that works well enough both for direct consumers (users of `std.os` namespace) or indirect consumers (eg. using `std.fs` abstractions built on `std.os`) without leaking too many details about the underlying implementation. The current approach tries to clump everything under the `posix` name, a name that carries a heavy baggage of do's and dont's that other platforms (mainly Windows) may not agree with.

A few examples:
* `rename` doesn't follow the posix semantics on Windows, unless you have a very recent Windows 10 version and `FILE_RENAME_FLAG_POSIX_SEMANTICS` is used.
* `open` accepts a well-defined set of options that's not even shared across all the posix-compatible and for Windows this means we're pinky-swearing to faithfully translate all of them into something equivalent. What if there's no direct equivalent? If you promise posix compatibility I'd expect something akin to `mingw`.
  * What we're interested in is getting a handle on a given file, with a given set of flags.
* `stat` doesn't exist on Windows, `GetFileAttributesEx` or equivalents are used and the resulting infos are chopped into a posix `stat` structure. It's a clear case of square peg in a round hole as many fields are unix-specific.
  * What we're interested in is getting a basic set of infos about the file, platform-specific extra stuff may be retrieved/modified with a companion function.
* `preadv`/`pwritev`  are not available on Windows and, at the moment at least, you get a nice compile error trying to use them.
  * What we're interested in is a `{write,read}Many` function, implemented using `p{write,read}v` or `ReadFileScatter`/`WriteFileGather`.
  * `iovec` is a posix thing, I just want to write/read a `[][]u8`!
* `copy_file_range` is a Linux-only syscall with a fallback on `pwrite`/`pread` that in turn falls back on `ReadFile`/`WriteFile` on Windows.
  * IMO this doesn't even belong to `os` as it can be safely implemented somewhere in `fs`, calling `copy_file_range` or other platform-specific APIs as needed.
  * It's not portable! It's not part of the posix standard!
  * `copy_file_range` is all about efficiency as it's done at the FS level, if the fallback path is triggered you get a disappointing read/write pair (not even a loop!)

[...]

PR #8590, opened 2021-04-21
x,  x/os/Socket: initial work on new Socket abstraction

lithdew:
Kick-start initial work on new cross-platform abstraction for sockets. Adds a test for read timeouts and a test for creating a non-blocking socket pair on Linux.

The new Socket abstraction is barebones and is made to support both blocking and non-blocking abstractions, alongside different socket protocols and domains.

Support for platform-dependent socket options that handle unsupported platforms gracefully via. comptime checks are provided for the new Socket abstraction.

This also marks the first out of many commits for introducing breaking changes to the standard library in a separate `x` package folder, which was pre-approved by @andrewrk.

The intent for the new `x` package is to introduce new async, event loop, networking, and operating system abstractions that would require breaking the standard library significantly. By having the `x` package, code in the standard library and compiler may then slowly be refactored to use the `x` package. Once modules in the `x` package are stabilized, they can be moved out of the `x` package, and a global 'grep' can be done to update import paths that resolve to the stabilized module in the `x` package.

This also kickstarts addressing #6600. Happy to fix minor things that should be addressed throughout the code in this PR; otherwise a subsequent PR might be better.


LemonBoy:
You may be interested in #7844 and its related issue too.

lithdew:
> You may be interested in #7844 and its related issue too.

Yep! Really like the layout of the API; my thoughts were to branch out TCP and UDP abstractions from this generic Socket abstraction (i.e. `TcpListener` and `TcpClient`, or `TCP.Listener` and `TCP.Client`).

lib/std/x.zig
```diff
@@ -0,0 +1 @@
+pub const os = @import("x/os/os.zig");
```
ifreund, line 1: You're missing a newline at the end of this and the other files, which will call a zig fmt CI check to fail if/when it is added: #8219

lithdew: Thanks for that - my editor didn't format on save :cry:. Pushed a commit with the files formatted.

lib/std/x/os/Socket.zig
```diff
@@ -0,0 +1,276 @@
...
+const time = std.time;
+const builtin = std.builtin;
+const testing = std.testing;
+
+const Socket = @This();
+
+/// A socket-address pair.
+pub const Connection = struct {
+    socket: Socket,
+    address: net.Address,
```
Aransentin, line 15: Not sure if it matters a whole lot, but `net.Address` can get pretty big since it's an union with `sockaddr_un`. If I'm just creating a TCP server with this, every connection will needlessly lug around an extra 100 bytes or so.

LemonBoy: I had started working on this problem [here](https://github.com/LemonBoy/zig/commit/ec622ee5b449b639791129812fe428c3ba173a1f), I hope it can serve as starting point for some further work.

lithdew: > I had started working on this problem [here](https://github.com/LemonBoy/zig/commit/ec622ee5b449b639791129812fe428c3ba173a1f), I hope it can serve as starting point for some further work.

Awesome :smile:! Would you mind if I include it and other codes in a reusable state from your branch in one/more subsequent PR(s)? Also wouldn't mind you taking charge (going to also be PR'ing a few things to do with concurrency after this one).

LemonBoy: > Would you mind if I include it and other codes in a reusable state from your branch in one/more subsequent PR(s)?

Please do! I've lost steam and left that PR to rot, it'd be great to salvage some code from there.

lithdew:
Rebased with master to re-run CI checks.

PR #8593, opened 2021-04-21
std: fix potential bug in parseInt

g-w1:
Sorry for not thinking of this case. 

/cc @LemonBoy


lib/std/fmt.zig
```diff
@@ -1454,6 +1458,8 @@ fn parseWithSign(
 
     var x: T = 0;
 
+    if (buf_start[0] == '_' or buf_start[buf_start.len - 1] == '_') return error.InvalidCharacter;
```
LemonBoy, line 1461: Needs a check on `buf_start` length

g-w1: https://github.com/ziglang/zig/blob/d10ec6e70dbcf49ba14be549069271e564fec76b/lib/std/fmt.zig#L1423
^^ Does this not do it?

LemonBoy: Not if you have a prefix, but the check below should take care of this case so carry on.

PR #8596, opened 2021-04-22
Fix a performance hazard in std.HashMap

SpexGuy:
I was doing some microbenchmarks of std.HashMap and noticed something odd:
image

Despite making more assumptions, `putAssumeCapacityNoClobber` was by far the slowest `put` variant.  It turns out the reason for this is the `assert(!self.contains(key))` at the beginning of the function, which apparently cannot be optimized away.

This patch adds guards to prevent the assert code from running in release-fast builds.  With these guards, `putAssumeCapacityNoClobber` becomes the fastest `put` in these tests.

It would be nice to dig into the IR and figure out why this can't be optimized and what piece of it is left over, but I don't really want to do that right now.  If this can be fixed on the zig side or in LLVM though, we should do that instead.

### DIFF
--- a/lib/std/hash_map.zig
+++ b/lib/std/hash_map.zig
@@ -472,9 +472,7 @@ pub fn HashMapUnmanaged(
 
         /// Insert an entry in the map. Assumes it is not already present.
         pub fn putNoClobber(self: *Self, allocator: *Allocator, key: K, value: V) !void {
-            assert(!self.contains(key));
             try self.growIfNeeded(allocator, 1);
-
             self.putAssumeCapacityNoClobber(key, value);
         }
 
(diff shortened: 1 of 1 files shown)


SpexGuy:
Here's a repro case if anyone wants to dig into the generated code:
https://zig.godbolt.org/z/jMvc7nvze

andrewrk:
Thanks for finding and reporting this issue @SpexGuy. Before merging this workaround I want to investigate a little bit. I think this could be a deeper issue and I want to make sure there isn't a root cause solution before we slap a bandage on HashMap.

andrewrk (approved):
OK I don't have time to look into this right now either, so here's my suggested plan, @SpexGuy if it works for you:

 * Open new issue to track the problem
 * Link to the issue in the comments in this patch
 * Merge

SpexGuy:
That sounds like a great solution!

Sahnvour:
@SpexGuy If that's of interest to you, I just updated my hashmap benchmark [here](https://github.com/Sahnvour/zig-containers#benchmarks).

andrewrk:
I'm guessing this is either solved or will need to be re-done based on #8975.

PR #8599, opened 2021-04-23
stage1: Allow wrapping negation on unsigned ints at comptime

### LINKED ISSUE #1770: wrapping negation -% should work on unsigned ints
This is used in C in these algorithms: http://www.pcg-random.org/posts/bounded-rands.html (search for "-range").

Semantics:

```zig
const U = u32;
const I = i32;
var u: U = foo();
assert(-%u == @bitCast(U, -%@bitCast(I, u)));
assert(-%u == ~u +% 1);
```

I'm trying to port this code to Zig, and the workarounds with `@bitCast` or `~` obscure the origin and intent of the code.

### LINKED ISSUE #7951: Wrapping negation of unsigned integers is allowed at runtime but not at comptime
Wrapping negation of unsigned integers is occasionally useful, especially in branchless code.  It converts 0 to 0, and 1 to 0xFFFF... .  This is perfect for construct masks from booleans.  Unfortunately, it doesn't work at comptime.  The workarounds here are pretty ugly, here's the simplest one I can come up with that doesn't depend on whether the mask type is signed:
```zig
    if (@typeInfo(@TypeOf(.{value})).Struct.fields[0].is_comptime) {
        // Since it's comptime, we don't need this to generate nice code.
        return if (value) ~@as(MaskInt, 0) else 0;
    }

    return -%@intCast(MaskInt, @boolToInt(value));
```

Wrapping negation should be allowed for sized unsigned integers both at runtime and comptime.

### LINKED ISSUE #8574: Wrapping subtraction/negation operator oddities
The [table of operators](https://ziglang.org/documentation/master/#Operators) in the Zig documentation contains the following:

Syntax | Relevant Types | Description | Example
-- | -- | -- | --
`a -% b` `a -%= b` | Integers | Wrapping Subtraction. <ul><li>Guaranteed to have twos-complement wrapping behavior.</li><li>Invokes Peer Type Resolution for the operands. See also `@subWithOverflow`.</li></ul> | `@as(u32, 0) -% 1 == std.math.maxInt(u32)`
`-%a` | Integers | Wrapping Negation. <ul><li>Guaranteed to have twos-complement wrapping behavior.</li></ul> | `-%@as(i32, std.math.minInt(i32)) == std.math.minInt(i32)`

I would expect the unary `-%` to behave identically to `@as(some_unsigned_type, 0) -% 1`, but instead here's what I'm seeing:

Operation | Expected Result | Actual Result
-- | -- | --
`@as(u32, 0) -% 1` | `std.math.maxInt(u32)` | `std.math.maxInt(u32)`
`-%@as(u32, 1)` | `std.math.maxInt(u32)` | `invalid wrapping negation type: 'u32'`
`-%1` | (not sure... it's a `comptime_int`, maybe it just shouldn't wrap?) | `0`

The compilation error on `-%@as(u32, 1)` is particularly interesting. I can't link the source on GitHub since the file is 1.5 MiB big (!), but the failure happens in `src/stage1/ir.cpp`, inside `ir_eval_negation_scalar()`:

```
    bool ok_type = ((scalar_type->id == ZigTypeIdInt && scalar_type->data.integral.is_signed) ||
        scalar_type->id == ZigTypeIdComptimeInt || (is_float && !is_wrap_op));

    if (!ok_type) {
        const char *fmt = is_wrap_op ? "invalid wrapping negation type: '%s'" : "invalid negation type: '%s'";
        return ir_add_error(ira, source_instr, buf_sprintf(fmt, buf_ptr(&scalar_type->name)));
    }
```

I don't know why the requirement for `scalar_type->data.integral.is_signed`: the wrapping behavior is especially useful on unsigned types. A quick `git blame` indicates this check has existed since all the way back to 0ad580f001eb151d0feb7ad884e237b237220800.

LemonBoy:
Closes #7951
Closes #8574
Closes #1770

### DIFF
--- a/lib/std/math.zig
+++ b/lib/std/math.zig
@@ -1349,15 +1349,6 @@ pub fn boolMask(comptime MaskInt: type, value: bool) callconv(.Inline) MaskInt {
         return @bitCast(i1, @as(u1, @boolToInt(value)));
     }
 
-    // At comptime, -% is disallowed on unsigned values.
-    // So we need to jump through some hoops in that case.
-    // This is a workaround for #7951
-    if (@typeInfo(@TypeOf(.{value})).Struct.fields[0].is_comptime) {
-        // Since it's comptime, we don't need this to generate nice code.
-        // We can just do a branch here.
-        return if (value) ~@as(MaskInt, 0) else 0;
-    }
-
     return -%@intCast(MaskInt, @boolToInt(value));
 }
 
--- a/src/stage1/bigint.cpp
+++ b/src/stage1/bigint.cpp
@@ -1446,10 +1446,10 @@ void bigint_negate(BigInt *dest, const BigInt *op) {
     bigint_normalize(dest);
 }
 
-void bigint_negate_wrap(BigInt *dest, const BigInt *op, size_t bit_count) {
+void bigint_negate_wrap(BigInt *dest, const BigInt *op, size_t bit_count, bool is_signed) {
     BigInt zero;
     bigint_init_unsigned(&zero, 0);
-    bigint_sub_wrap(dest, &zero, op, bit_count, true);
+    bigint_sub_wrap(dest, &zero, op, bit_count, is_signed);
 }
 
 void bigint_not(BigInt *dest, const BigInt *op, size_t bit_count, bool is_signed) {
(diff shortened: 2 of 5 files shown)


marler8997:
You can't fix 3 issues with 1 PR, that's illegal.

andrewrk (approved):

andrewrk:
> You can't fix 3 issues with 1 PR, that's illegal.

I ain't gonna snitch, are you?

marler8997:
My lips are sealed....

PR #8604, opened 2021-04-24
further haiku support system definitions

hoanga:
I have made some more progress in continuing Haiku support for Zig. The following changes should get Zig to support more commands under Haiku.  This compiles with a llvm12 installation  (tested on nightly build hrev55039) .

llvm12 support should hopefully land in Haiku (there's an rc port [available here](https://github.com/haikuports/haikuports/blob/master/sys-devel/llvm/llvm12-12.0.0%7Erc4.recipe) )

Things I have been able to confirm now functional that wasn't previously (previous commands that were working under Haiku should still be):

* zig libc now properly detects paths properly instead of hanging
* zig translate-c looks like it is working
* zig init-exe and init-lib create structure

Zig compile related commands (run, build, test) now returning an error 'getSymbolAtAddress' which I think is to be expected for now. This also affects running 'make install' for the time being.

Sample log:

```
~/src/git/zig-bootstrap/out/host/bin> uname -a
Haiku shredder 1 hrev55039 Apr 14 2021 06:02:26 x86_64 x86_64 Haiku`

~/src/git/zig-bootstrap/out/host/bin> ./zig version
0.8.0-dev.1939+5a3ea9bec

~/src/git/zig-bootstrap/out/host/bin> ./zig libc
# The directory that contains `stdlib.h`.
# On POSIX-like systems, include directories be found with: `cc -E -Wp,-v -xc /dev/null`
include_dir=/boot/system/develop/headers

# The system-specific include directory. May be the same as `include_dir`.
# On Windows it's the directory that includes `vcruntime.h`.
# On POSIX it's the directory that includes `sys/errno.h`.
sys_include_dir=/boot/system/develop/headers

# The directory that contains `crt1.o` or `crt2.o`.
# On POSIX, can be found with `cc -print-file-name=crt1.o`.
# Not needed when targeting MacOS.
crt_dir=/system/develop/lib

# The directory that contains `vcruntime.lib`.
# Only needed when targeting MSVC on Windows.
msvc_lib_dir=

# The directory that contains `kernel32.lib`.
# Only needed when targeting MSVC on Windows.
kernel32_lib_dir=

```

### DIFF
--- a/lib/std/c/haiku.zig
+++ b/lib/std/c/haiku.zig
@@ -69,3 +69,51 @@ pub const pthread_rwlock_t = extern struct {
     writer_count: i32 = 0,
     waiters: [2]?*c_void = [_]?*c_void{ null, null },
 };
+
+pub const EAI = extern enum(c_int) {
+    /// address family for hostname not supported
+    ADDRFAMILY = 1,
+
+    /// name could not be resolved at this time
+    AGAIN = 2,
+
+    /// flags parameter had an invalid value
+    BADFLAGS = 3,
+
+    /// non-recoverable failure in name resolution
+    FAIL = 4,
+
+    /// address family not recognized
+    FAMILY = 5,
+
+    /// memory allocation failure
+    MEMORY = 6,
+
+    /// no address associated with hostname
+    NODATA = 7,
+
+    /// name does not resolve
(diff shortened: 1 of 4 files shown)


andrewrk:
Nice work!

PR #8606, opened 2021-04-24
stage1: Require a block after suspend

### LINKED ISSUE #8603: always require a suspend block
This proposal is to remove `suspend;` syntax in favor of `suspend {}`.

Reasoning: 
 * It simplifies the language.
 * It hints that you might want to put something inside the suspend block, which is often true.
 * It's pretty much the same effort to type `;` or `{}`.

I'm pretty sure this is a good idea so I'm filing it as already accepted.

LemonBoy:
Closes #8603

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -6528,7 +6528,7 @@ test "suspend with no resume" {
 
 fn func() void {
     x += 1;
-    suspend;
+    suspend {}
     // This line is never reached because the suspend has no matching resume.
     x += 1;
 }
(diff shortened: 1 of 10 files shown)


g-w1:
Should there be a zig fmt fix for this? Or would that come in another pr?

LemonBoy:
> Should there be a zig fmt fix for this?

Good question, I didn't include one as this change is pretty simple to perform using a text editor but also wrote it to reformat the stdlib.

andrewrk:
I think we can get by without a zig fmt fix in this case. Thanks for the implementation @LemonBoy !

PR #8609, opened 2021-04-24
Add std.crypto.utils.timingSafeCompare

jedisct1:
A little function to complement the existing `crypto.utils.timingSafeEql()` function with a way to compare large numbers serialized as arrays.

This is useful to compare nonces and to check that group elements are in canonical form.

Absence of side channels remains a best effort, reusing the common pattern we use elsewhere.

### DIFF
--- a/lib/std/crypto/utils.zig
+++ b/lib/std/crypto/utils.zig
@@ -1,4 +1,5 @@
 const std = @import("../std.zig");
+const debug = std.debug;
 const mem = std.mem;
 const testing = std.testing;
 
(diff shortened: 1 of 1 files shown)


andrewrk:
Is the length of the arrays required to be comptime-known? Seems like these could be slices with `assert(a.len == b.len);`, no?

andrewrk:
If the algorithm that provides timing safety does not depend on comptime-ness of the array length, then I suggest to use slices. This will be more widely usable by application and library authors.

jedisct1:
Changed!

andrewrk (approved):

PR #8610, opened 2021-04-24
export winmain if linking libc and using windows subsystem

travisfoo:
Hey there, this is my first PR to zig and I'm an open source newbie so let me know if I'm doing anything wrong.

Anyway, the issue at hand is this linker error: `undefined symbol: WinMain` when linking libc and using the windows subsystem. My OS is Windows x86_64. Here's an easy way to reproduce:
1. `mkdir test`
2. `cd test`
3. `zig init-exe`
4. `zig build-exe --subsystem windows -lc src/main.zig`

I get this error:
```
lld-link: error: undefined symbol: WinMain
>>> referenced by D:\agent\_work\9\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:102
>>>               libcmtd.lib(exe_winmain.obj):(int __cdecl invoke_main(void))
error: LLDReportedFailure`
```

I took a look in start.zig and it looks like there's a special path for exporting winmain symbols that isn't being reached when linking libc. I put in a change that fixed the issue for me. I'm not a Windows expert so please let me know if there's something I'm missing here or if you would like the change edited in any way.

Thank you!

### DIFF
--- a/lib/std/start.zig
+++ b/lib/std/start.zig
@@ -39,7 +39,18 @@ comptime {
             }
         } else if (builtin.output_mode == .Exe or @hasDecl(root, "main")) {
             if (builtin.link_libc and @hasDecl(root, "main")) {
-                if (@typeInfo(@TypeOf(root.main)).Fn.calling_convention != .C) {
+                if (builtin.os.tag == .windows and 
+                    builtin.subsystem == builtin.SubSystem.Windows and
+                    !@hasDecl(root, "wWinMain")) 
+                {
+                    if (@hasDecl(root, "WinMain")) {
+                        @compileError("WinMain not supported; declare wWinMain instead");
+                    }
+                    else {
+                        @export(WinStartup, .{ .name = "wWinMain" });
+                    }
+                }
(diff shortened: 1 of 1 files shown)


g-w1:
Please run zig fmt on this code.

travisfoo:
This PR is failing due to stage1.zig exporting a C-style main. When we export WinMain it will call into main from zig, causing the callMain compile error since the return type of the main function defined in stage1.zig is not what zig wants. The subsystem auto-determines that if we link libc and we have a main function, we are in the windows subsystem, so we don't really have a way to tell if we need to export WinMain when linking libc. Unless I'm missing something, it seems like the best way to handle this is for users to manually export WinMain when linking libc.

Vexu:
I'm not too familiar with windows subsystem stuff, but something you could try is using `@typeInfo` in `builtin.subsystem` to try to detect if the main is C or Zig style.

Also is this msvc specific? I wasn't able to reproduce the error by cross-compiling to `native-windows-gnu`.

Vexu:
Closing as there has been no progress in several months.

PR #8611, opened 2021-04-25
parser: Use an operator precedence table

shachaf:
Instead of having one function per precedence level, mirroring the BNF grammar,
this uses the "precedence climbing" algorithm with a table of operator
precedences (and a few special cases that don't fit into the table as it is).

This is a first pass -- it's probably possible to put more of the parser
into this form, e.g. to support prefix/suffix operators with precedence, if
necessary, or just to simplify the code more.

It may also be possible to speed this up by putting more useful information
into the tokens during tokenization, to avoid the extra branch on the token in
operInfo.

### DIFF
--- a/lib/std/zig/parse.zig
+++ b/lib/std/zig/parse.zig
@@ -1336,45 +1336,41 @@ const Parser = struct {
     // A table of binary operator information. Higher precedence numbers are
     // stickier. All operators at the same precedence level should have the same
     // associativity.
-    fn operInfo(tag: Token.Tag) OperInfo {
-        return switch (tag) {
-            .keyword_or => .{ .prec = 10, .tag = .bool_or },
+    const operTable = std.enums.directEnumArrayDefault(Token.Tag, OperInfo, .{ .prec = -1, .tag = Node.Tag.root }, 0, .{
+        .keyword_or = .{ .prec = 10, .tag = .bool_or },
 
-            .keyword_and => .{ .prec = 20, .tag = .bool_and },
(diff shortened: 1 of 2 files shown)


andrewrk:
> It may also be possible to speed this up by putting more useful information
> into the tokens during tokenization, to avoid the extra branch on the token in
> operInfo.

That's unlikely to happen. We have a really compact 5 bytes per token right now, and doing so was a measurable performance improvement. So far all the big perf wins have been by making representation of various pieces of heap data smaller.

The code simplification in this patch is nice. Can you share any performance impact stats?

ifreund:
I've run some basic benchmarking formatting the std with perf and it looks like this PR is a very minor performance improvement if anything, certainly not a regression:
Before:
```
$ perf stat -e cache-misses -e instructions -e cycles ./build/zig fmt lib/std

 Performance counter stats for './build/zig fmt lib/std':

           787,097      cache-misses:u                                              
     4,988,785,455      instructions:u            #    2.89  insn per cycle         
     1,725,526,234      cycles:u                                                    

       0.415256074 seconds time elapsed

       0.402567000 seconds user
       0.011987000 seconds sys
```
After:
```
$ perf stat -e cache-misses -e instructions -e cycles ./build/zig fmt lib/std

 Performance counter stats for './build/zig fmt lib/std':

           818,776      cache-misses:u                                              
     4,800,522,512      instructions:u            #    2.91  insn per cycle         
     1,648,634,860      cycles:u                                                    

       0.399708854 seconds time elapsed

       0.378598000 seconds user
       0.019974000 seconds sys
```

andrewrk (approved):

PR #8616, opened 2021-04-25
Function pointer alignment

### LINKED ISSUE #8537: run-translated-c tests failing for aarch64-linux after merging llvm12
After merging llvm12 branch, drone CI is hitting an error in the `run-translated-c` tests:

```
The following command terminated unexpectedly:
cd /drone/src && /drone/src/zig-cache/o/953f294ec50580c71907aa68df1a0b93/translated_c 
/drone/src/zig-cache/o/054cb3aa222f1059385d9cbac7ab05f1/build /drone/src/build/zig /drone/src /drone/src/zig-cache /root/.cache/zig test -Dskip-release -Dskip-non-native -Dskip-compile-errors
```

More investigation is needed to find out why.

LemonBoy:
Refer to the single commits for more details.
Not sure how much of this is also needed in the stage2 compiler.

Fixes #8537

### DIFF
--- a/lib/std/meta.zig
+++ b/lib/std/meta.zig
@@ -117,10 +117,21 @@ test "std.meta.bitCount" {
     testing.expect(bitCount(f32) == 32);
 }
 
+/// Returns the alignment of type T.
+/// Note that if T is a pointer or function type the result is different than
+/// the one returned by @alignOf(T).
+/// If T is a pointer type the alignment of the type it points to is returned.
+/// If T is a function type the alignment a target-dependent value is returned.
 pub fn alignment(comptime T: type) comptime_int {
-    //@alignOf works on non-pointer types
-    const P = if (comptime trait.is(.Pointer)(T)) T else *T;
-    return @typeInfo(P).Pointer.alignment;
+    return switch (@typeInfo(T)) {
+        .Optional => |info| switch (@typeInfo(info.child)) {
+            .Pointer, .Fn => alignment(info.child),
+            else => @alignOf(T),
+        },
+        .Pointer => |info| info.alignment,
+        .Fn => |info| info.alignment,
+        else => @alignOf(T),
+    };
 }
 
 test "std.meta.alignment" {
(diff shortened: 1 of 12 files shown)


lib/std/meta.zig
```diff
@@ -117,10 +117,21 @@ test "std.meta.bitCount" {
     testing.expect(bitCount(f32) == 32);
 }
 
+/// Returns the alignment of type T.
+/// Note that if T is a pointer or function type the result is different than
+/// the one returned by @alignOf(T).
+/// If T is a pointer type the alignment of the type it points to is returned.
+/// If T is a function type the alignment a target-dependent value is returned.
```
komuw, line 124: 🔥 
i appreciate the docs.

src/stage1/analyze.cpp
```diff
@@ -4769,11 +4769,11 @@ Error type_is_nonnull_ptr2(CodeGen *g, ZigType *type, bool *result) {
...
-static uint32_t get_async_frame_align_bytes(CodeGen *g) {
-    uint32_t a = g->pointer_size_bytes * 2;
-    // promises have at least alignment 8 so that we can have 3 extra bits when doing atomicrmw
-    if (a < 8) a = 8;
-    return a;
+uint32_t get_async_frame_align_bytes(CodeGen *g) {
+    // Due to how the frame structure is built the minimum alignment is the one
+    // of a usize (or pointer).
+    // label (grep this): [fn_frame_struct_layout]
+    return max(g->builtin_types.entry_usize->abi_align, target_fn_align(g->zig_target));
```
LemonBoy, line 4776: Big question for @andrewrk, why is the compiler factoring in the target function alignment when calculating the minimum alignment for the frame type?

andrewrk: Hmmm. I don't think it should be doing that. It should be determined by the largest aligned local that has to get spilled to the function frame, or usize (because of the fields the language needs to add to make async functions work).

andrewrk:
>  Not sure how much of this is also needed in the stage2 compiler.

Looks mostly right, but maybe a few more items need to go into the `=> 4` prong:

https://github.com/ziglang/zig/blob/6f61594692af846226880e5903ad26a922014d55/src/type.zig#L1107-L1117

src/stage1/analyze.cpp
LemonBoy: Indeed, the 2 * usize_align formula used previously left me quite confused at first.

> It should be determined by the largest aligned local that has to get spilled to the function frame

I'm definitely no expert in the coroutine implementation but ideally `anyframe`'s alignment is defined as the biggest among every frame in the compilation unit in order to be compatible with every single frame, no?

PR #8617, opened 2021-04-25
translate-c: make ptrs from address_of optional

### LINKED ISSUE #8322: translate-c: assignment to dereferenced address-of produces invalid Zig
Not sure if this is a translate-c bug or a type coercion bug.

C code:
```C
void foo(void) {
    int i = 0;
    *&i = 42;
}
```
Produces:
```zig
pub export fn foo() void {
    var i: c_int = 0;
    &i.?.* = 42;
}
```

Compiling that gives:
```
error: invalid left-hand side to assignment
    &i.?.* = 42;
```

Adding parentheses does not help:
```
error: expected optional type, found '*c_int'
    (&i).?.* = 42;
      ^
```

Adding an `@as()` does:
```Zig
var i: c_int = 0;
@as(?*c_int, &i).?.* = 42; // compiles fine
// or @as([*c]c_int, &i).* = 42;
```

xackus:
Fixes #8322.

### DIFF
--- a/lib/std/meta.zig
+++ b/lib/std/meta.zig
@@ -1029,6 +1029,13 @@ test "std.meta.cast" {
     testing.expectEqual(cast(C_ENUM, @as(u64, 42)), @intToEnum(C_ENUM, 42));
 }
 
+/// Make a pointer optional.
+/// This is needed because in translate-c all pointers are assumed to be optional.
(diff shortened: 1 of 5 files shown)


andrewrk:
Can we perhaps look for this pattern specifically? It's unfortunate how much degraded the common-case code becomes with this change.

PR #8618, opened 2021-04-25
Two small patches

LemonBoy:
cc @koachan 

Without the `flushw` I was getting spurious segfaults here and there when running the test suite, even a simple example like this would segfault:
```zig
const std = @import("std");

pub fn main() !void {
    @panic("foo");
}
```

Can you confirm this is fine on your machine?

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -336,6 +336,13 @@ pub const StackIterator = struct {
     fp: usize,
 
     pub fn init(first_address: ?usize, fp: ?usize) StackIterator {
+        if (builtin.arch == .sparcv9) {
+            // Flush all the register windows on stack.
+            asm volatile (
+                \\ flushw
+                ::: "memory");
+        }
+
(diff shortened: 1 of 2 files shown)


andrewrk (approved):

lib/std/debug.zig
```diff
@@ -336,25 +336,32 @@ pub const StackIterator = struct {
     fp: usize,
 
     pub fn init(first_address: ?usize, fp: ?usize) StackIterator {
+        if (builtin.arch == .sparcv9) {
+            // Flush all the register windows on stack.
+            asm volatile (
+                \\ flushw
+                ::: "memory");
+        }
+
```
andrewrk, line 345: Would this solve the problem in an architecture-agnostic way?

```suggestion
        @fence(.Acquire);
```

LemonBoy: Nope, `flushw` is not a fence.

PR #8621, opened 2021-04-26
thread: simplify and remove useless return in spawn

vrischmann:


jedisct1:
Sure, why not!

Thank you!

andrewrk:
This is measurably worse! Zig's error return traces will point to the `return` token if they happen to occur, so having multiple `return` statements makes those stack traces really helpful. This destroys debuggability. Reverting immediately!

vrischmann:
Sorry about that. I initially just wanted to remove the superfluous return and thought I might as well use `return switch`, but I didn't think about the error return traces.

andrewrk:
No worries! It's a bit of institutional knowledge to pass on.

PR #8622, opened 2021-04-26
stage2: Add framework for host CPU detection on Linux

LemonBoy:
Add a generic framework to parse /proc/cpuinfo and implement the model
detection for 64bit SPARC targets as proof of concept. 

CC @koachan, please give this a try on your Ultrasparc T2
CC @joachimschmidt557

### DIFF
--- a/lib/std/zig/system/linux.zig
+++ b/lib/std/zig/system/linux.zig
@@ -1,0 +1,118 @@
+const std = @import("std");
+const mem = std.mem;
+const io = std.io;
+const fs = std.fs;
+const fmt = std.fmt;
+const testing = std.testing;
+
+const Target = std.Target;
+const CrossTarget = std.zig.CrossTarget;
+
+const assert = std.debug.assert;
+
+const SparcCpuinfoImpl = struct {
+    model: ?*const Target.Cpu.Model = null,
+    is_64bit: bool = false,
+
+    const cpu_names = .{
+        .{ "SuperSparc", &Target.sparc.cpu.supersparc },
+        .{ "HyperSparc", &Target.sparc.cpu.hypersparc },
+        .{ "SpitFire", &Target.sparc.cpu.ultrasparc },
+        .{ "BlackBird", &Target.sparc.cpu.ultrasparc },
+        .{ "Sabre", &Target.sparc.cpu.ultrasparc },
+        .{ "Hummingbird", &Target.sparc.cpu.ultrasparc },
+        .{ "Cheetah", &Target.sparc.cpu.ultrasparc3 },
+        .{ "Jalapeno", &Target.sparc.cpu.ultrasparc3 },
+        .{ "Jaguar", &Target.sparc.cpu.ultrasparc3 },
+        .{ "Panther", &Target.sparc.cpu.ultrasparc3 },
+        .{ "Serrano", &Target.sparc.cpu.ultrasparc3 },
+        .{ "UltraSparc T1", &Target.sparc.cpu.niagara },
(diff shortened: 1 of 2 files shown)


lib/std/zig/system/linux.zig
```diff
@@ -0,0 +1,118 @@
...
+}
+
+pub fn detectNativeCpuAndFeatures() ?Target.Cpu {
+    var f = fs.openFileAbsolute("/proc/cpuinfo", .{ .intended_io_mode = .blocking }) catch |err| switch (err) {
+        else => return null,
+    };
+    defer f.close();
+
+    switch (std.Target.current.cpu.arch) {
+        .sparcv9 => return SparcCpuinfoParser.parse(f.reader()) catch null,
```
koachan, line 113: I got a compile error at this line:
```
./lib/std/zig/system/linux.zig:113:61: error: expected type 'std.target.Arch', found 'std.io.reader.Reader(std.fs.file.File,std.os.ReadError,std.fs.file.File.read)'
        .sparcv9 => return SparcCpuinfoParser.parse(f.reader()) catch null,
                                                            ^
./lib/std/target.zig:708:26: note: std.target.Arch declared here
        pub const Arch = enum {
                         ^
./lib/std/io/reader.zig:21:12: note: std.io.reader.Reader(std.fs.file.File,std.os.ReadError,std.fs.file.File.read) declared here
    return struct {
           ^
```
I think you forgot to pass the `.sparcv9` arch object into the method.

koachan:
Okay, this looks good to me. It detected both the model and featureset correctly.

andrewrk (approved):

lib/std/zig/system/linux.zig
```diff
@@ -0,0 +1,120 @@
...
+        .{ "Jaguar", &Target.sparc.cpu.ultrasparc3 },
+        .{ "Panther", &Target.sparc.cpu.ultrasparc3 },
+        .{ "Serrano", &Target.sparc.cpu.ultrasparc3 },
+        .{ "UltraSparc T1", &Target.sparc.cpu.niagara },
+        .{ "UltraSparc T2", &Target.sparc.cpu.niagara2 },
+        .{ "UltraSparc T3", &Target.sparc.cpu.niagara3 },
+        .{ "UltraSparc T4", &Target.sparc.cpu.niagara4 },
+        .{ "UltraSparc T5", &Target.sparc.cpu.niagara4 },
+        .{ "LEON", &Target.sparc.cpu.leon3 },
+    };
```
Mouvedia, line 35: Seems like the list isn't exhaustive looking at https://en.wikipedia.org/wiki/SPARC#Implementations
Being a POC I guess it doesn't matter.

Mouvedia: So we are limited by LLVM, I see.
Can you add an assert so that it fails if a new one is added in the future?

LemonBoy: Added where? If the model is not recognized the autodetection fails and we pick a safe value.

Mouvedia: I thought that it would be pretty hard to detect/know that LLVM has added a new one and hence the list needs to be updated.

> If the model is not recognized the autodetection fails and we pick a safe value.

I am talking about upcoming models.
i.e. assert(spark count == expected)

andrewrk: We are not limited by LLVM; see `tools/update_cpu_features.zig` if you want to add CPU models that LLVM does not support.

PR #8624, opened 2021-04-26
25519: remove unused const, safeguard against unreduced scalars

jedisct1:
No behavior change, but it makes the existing code better match the forthcoming code for other curves.

Rename nonAdjacentForm() to slide(), remove an unneeded and confusing constant, and do a reduction in slide() if 257 bits would be required.

Note that in all the high-level functions, the top bit is always cleared, so the reduction is never necessary. But since the low-level functions are public, the check is a safe thing to have.

### DIFF
--- a/lib/std/crypto/25519/edwards25519.zig
+++ b/lib/std/crypto/25519/edwards25519.zig
@@ -75,15 +75,6 @@ pub const Edwards25519 = struct {
         .is_base = true,
     };
 
-    /// The edwards25519 neutral element.
-    pub const neutralElement = Edwards25519{
-        .x = Fe{ .limbs = .{ 2251799813685229, 2251799813685247, 2251799813685247, 2251799813685247, 2251799813685247 } },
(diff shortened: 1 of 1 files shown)


lib/std/crypto/25519/edwards25519.zig
```diff
@@ -75,15 +75,6 @@ pub const Edwards25519 = struct {
         .is_base = true,
     };
 
-    /// The edwards25519 neutral element.
-    pub const neutralElement = Edwards25519{
```
andrewrk, line 79: I know nothing about this, but it is marked `pub`. Just double checking, are std lib users expecting to find this constant available?

jedisct1: It is equivalent to `identityElement`.

But since `identityElement` is private, let's make that one public, and add a deprecation compile error log for `neutralElement`.

andrewrk (approved):

PR #8627, opened 2021-04-27
std.crypto: add support for the NIST P-256 curve

jedisct1:
Uses verified code generated by fiat-crypto for field arithmetic, and complete formulas to avoid side channels.

There's still plenty of room for optimizations, especially with a fixed base. But this gives us a framework to easily add other similar curves.


lib/std/crypto/pcurves/p256.zig
```diff
@@ -0,0 +1,423 @@
...
+    pub fn toCompressedSec1(p: P256) [33]u8 {
+        var out: [33]u8 = undefined;
+        const xy = p.affineCoordinates();
+        out[0] = if (xy.y.isOdd()) 3 else 2;
+        mem.copy(u8, out[1..], &xy.x.toBytes(.Big));
+        return out;
+    }
+
+    /// Serialize a point using the uncompressed SEC-1 format.
+    pub fn toUncompressedSec1(p: P256) [65]u8 {
```
Mouvedia, line 119: since we have `fromSec1`
maybe this could be renamed `toSec1`

being uncompressed is the default state, it's only worth having for `toCompressedSec1` IMO

jedisct1: `fromSec1()` can parse compressed and uncompressed representations, as this information is present.

A `toSec1()` function would would have to make a choice (or get an ugly API) since the output sizes of both representations differ.

Compressed vs uncompressed is a never ending debate (former is smaller and more secure, later is more common), so maybe we should stay neutral and give no advantage to either wire format.

Mouvedia: > or get an ugly API

(p: P256, c: bool) => [65]u8 | [33]u8?
Yeah that smells.

> we should stay neutral and give no advantage to either wire format

👍

Issue #8363, opened 2021-03-26
ArrayHashMap: store values separately from key/hash


andrewrk:
Currently there is an `Entry`  type of `ArrayHashMap` that looks like this:

```zig
        /// Modifying the key is illegal behavior.
        /// Modifying the value is allowed.
        /// Entry pointers become invalid whenever this ArrayHashMap is modified,
        /// unless `ensureCapacity` was previously used.
        pub const Entry = struct {
            /// This field is `void` if `store_hash` is `false`.
            hash: Hash,
            key: K,
            value: V,
        };
```

There are two reasons for storing values separately:
 1. If the key/hash and value have different alignments, we may be wasting space in `Entry` via padding.
 2. If the values array is separate, then it can be extracted out of an `ArrayHashMap` into a slice whose memory is managed separately. In other words, ArrayHashMap can be used to construct a contiguous array of values, which then can be "downgraded" into a `std.ArrayList`. Likewise, an ArrayList could be "upgraded" to an ArrayHashMap. In both cases, it would not require moving memory around; the contiguous array of values would remain intact for both transformations. This would allow some nice patterns, such as using ArrayHashMap as a temporary "view" of an ArrayList.

There is a real world use case for this is in the self-hosted compiler: constructing a string table. During construction, one wants to check if a string already exists, and use the same index. However once the string table is constructed, the slice of bytes is all that is desired to remain.

kubkon:
Looks great, and it'd definitely be a welcomed addition to the libstd for all of our linker worries. In addition to storing the string table, the same principle applies to storing different symbol tables (e.g., the globals at the traditional linker scope, or undefs, etc.).

matu3ba:
I do agree that this would be faster on frequent access.

This also works fine for comptime as the page size can be optimised for execution. 

It does work worse on accessing the hashmap values rarely during execution or of huge key/hash sets/tables (not fitting into cache):
1. When key and value are next to another one needs likely 1 cacheline load.
2. When they are apart from another (and one doesnt access them frequently, so they do not remain in cache) one needs to load 2 cachelines.

andrewrk:
@matu3ba I agree with you, however I think this proposal plays to the relative strengths of the ArrayHashMap vs the other HashMap implementation we have. I suspect that for the use case you outlined above, HashMap would be the appropriate tool for the job.

moosichu:
[...]

Issue #7221, opened 2020-11-25
Proposal: Change ABI alignment of zero-sized types from 0 to 1


SpexGuy:
In #6706 it was decided that pointers to zero-sized types should not be zero sized.  As a consequence of this, zero-sized types now need an ABI alignment.  This was chosen to be 0, to preserve the invariant that `@alignOf(T) <= @sizeOf(T)`.  However, in #6954, @marler8997 made a strong argument for making it 1, to preserve the invariant `@alignOf(T) >= 1`.  This issue exists to continue this discussion now that #6945 has been closed.

### Arguments for align 0:
- Preserves `@alignOf(T) <= @sizeOf(T)` (is this actually useful though?)
- You could say that zero sized types are bitpacked, if you put multiple of them together they share the same address
- Alignment being 0 is always a special case that needs to be handled in pointer type info, so having this special case in `@alignOf` isn't anything new.

### Arguments for align 1:
- Preserves `@alignOf(T) >= 1`, removing special cases when using this value.
- It's kind of weird to make a bit pointer if a zero-sized field follows an align(0) u3 in a non-packed ordered struct.  Then again, Zig doesn't have a non-packed ordered struct type that can contain align(0) fields, so this may be a moot point.

EleanorNB:
One thing which could make this cleaner: changing `align()` to take a *bit* alignment rather than a byte alignment. So `@alignOf(u8)` would be 8, `@alignOf(bool)` would be 1, and `@alignOf(ZST)` would also be 1. Thus, we eliminate special cases, and allow ZSTs to have arbitrary bit offsets. (Plus: we accommodate architectures with sub-byte or super-byte addressing, and `align()` now takes a clean `u32` rather than a lopsided `u29`, which I suspect was also LLVM's motivation for choosing that number.)

kyle-github:
Hmm, I like that idea, @EleanorNB.   This would be nice on platforms that had bytes that were not 8 bits, such as some DSPs.  On those, in C, it is sometimes difficult to tell how big something is because `sizeof` returns a size in the native bytes and not in units of 8 bits and alignment is in those `byte` units.

SpexGuy:
So the ABI alignment of sub-byte types (bool, u3, etc) is 1 byte, and
that's how it should be. Nothing gets bitpacked until you use a packed
struct or `align(0)`, which overrides the ABI alignment.  As far as I know,
there are currently no sized types that have an ABI alignment less than
[...]

Issue #6983, opened 2020-11-05
ICE when comparing optional zero-sized pointers


foobles:
It happens when comparing an optional zero-sized pointer to a zero-sized pointer:

```rust
pub fn main() !void {
    var x: u0 = 0;
    var y: u0 = 0;

    var x_opt_ptr: ?*u0 = &x;
    var y_ptr: *u0 = &y;

    var r = (x_opt_ptr == y_ptr); // ICE: Unreachable at codegen.cpp:3027
}
```

as well as when comparing optional zero-sized pointer to another optional zero-sized pointer:

```rust
pub fn main() !void {
    var x: u0 = 0;
    var y: u0 = 0;

    var x_opt_ptr: ?*u0 = &x;
    var y_opt_ptr: ?*u0 = &y;

    var r = (x_opt_ptr == y_opt_ptr); // ICE: Unreachable at codegen.cpp:3027
}
```

mikdusan:
th latest compiler:
- compiler no longer segfaults
- appears to have been fixed between 0.9.1 and 0.10.0

<details>
<summary>repro.zig</summary>

```zig
pub fn main() !void {
    var x: u0 = 0;
    var y: u0 = 0;

    {
        var x_opt_ptr: ?*u0 = &x;
        var y_ptr: *u0 = &y;

        var r = (x_opt_ptr == y_ptr); // ICE: Unreachable at codegen.cpp:3027
        _ = r;
    }
    {
        var x_opt_ptr: ?*u0 = &x;
        var y_opt_ptr: ?*u0 = &y;

        var r = (x_opt_ptr == y_opt_ptr); // ICE: Unreachable at codegen.cpp:3027
        _ = r;
    }
}
```
</details>

```sh
$ uname -a
Linux archlinux 6.1.4-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 07 Jan 2023 15:10:07 +0000 x86_64 GNU/Linux

$ zig version
0.11.0-dev.1255+f7ade7e63

$ zig build-obj repro.zig
# no stdout/stderr (good)
```

Issue #6947, opened 2020-11-02
Passing &slice[0] as argument causes compiler assert when payload is zero sized


SpexGuy:
```zig
// repro.zig
pub fn destroy(ptr: *void) void {}

test "" {
    var slice: []void = undefined;
    destroy(&slice[0]);
}
```
```batch
> zig version
0.6.0+0088efc4b
> zig test repro.zig
Assertion failed at D:\a\1\s\src\stage1\codegen.cpp:3933 in ir_render_elem_ptr. This is a bug in the Zig compiler.
Unable to dump stack trace: debug info stripped
```

PR #8632, opened 2021-04-28
std/ArrayList: Allow `ArrayList(u0)` to be created

zbanks:
I encountered this issue when writing a type-generic datastructure: if my datastructure is generic over type `T` and internally has an `std.ArrayList(T)`, then it fails to compile if `T` is zero-sized. It fails to compile because the allocator cannot allocate a slice for zero-sized types (e.g. "error: no align available for type 'void'").

Example code, [godbolt link](https://godbolt.org/z/b9jPx3Ta8) with errors
```zig
test {
    const std = @import("std");

    var voidList = std.ArrayList(void).init(std.testing.allocator);
    defer voidList.deinit();

    try voidList.append(void {});
    for (voidList.items) |_, index| {
        // Do something with `index`
    }
}
```

---

Although `ArrayList(void)` sounds dubious, I think a motivating example is creating a `HashSet(Key)` from a `HashMap(Key, void)`, which is something [the docs specifically mention](https://ziglang.org/documentation/master/#void). The current `std` implementations store the `Key` & `Value` pairs next to each other, in AoS form, which sidesteps this issue. However the problem would be encountered when implementing #8363, where the pairs are stored in separate arrays in SoA form.

I also wanted to mention that issue #7221 may impact this? I'm not fully sure of the ramifications of making zero-sized types have alignment 1. (Would it just fix this problem at a lower level? Would it break my patch's tests?)

---

For my implementation, I have `ArrayList(void)` still track the length of the list, but it never allocates additional memory. The `list.items` slice `ptr` is never updated, but the `len` is, so it can still be iterated over with `for`, etc. It sets `list.capacity` to `maxInt(usize)` (and never updates it).

I also wanted feedback on my code specifically:
- Is `ArrayList` the right place to fix this, or should the fix be at the `Allocator` level? (That sounds "scarier," more potential for a breaking change?)
- Is `@sizeOf(T) == 0` the best way to check for zero-sized types? I saw in the docs that comptime-only types, like `type` and `comptime_int`, would also match this test -- but it doesn't make sense to use an `ArrayList` for them.
- I set `capacity` to `maxInt(usize)` and never update it, even when calling `shrinkAndFree(...)` -- I think this makes sense? But it does violate an implied invariant that `list.shrinkAndFree(n); list.capacity == n`
- Is it more clear to have the test use `void` instead of `u0`? (Or are there other cases I should cover?)

### DIFF
--- a/lib/std/array_list.zig
+++ b/lib/std/array_list.zig
@@ -65,16 +65,23 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
         pub fn initCapacity(allocator: *Allocator, num: usize) !Self {
             var self = Self.init(allocator);
 
-            const new_memory = try self.allocator.allocAdvanced(T, alignment, num, .at_least);
-            self.items.ptr = new_memory.ptr;
-            self.capacity = new_memory.len;
+            if (@sizeOf(T) > 0) {
+                const new_memory = try self.allocator.allocAdvanced(T, alignment, num, .at_least);
+                self.items.ptr = new_memory.ptr;
+                self.capacity = new_memory.len;
+            } else {
+                // If `T` is a zero-sized type, then we do not need to allocate memory.
+                self.capacity = std.math.maxInt(usize);
+            }
 
             return self;
         }
 
         /// Release all allocated memory.
         pub fn deinit(self: Self) void {
-            self.allocator.free(self.allocatedSlice());
+            if (@sizeOf(T) > 0) {
+                self.allocator.free(self.allocatedSlice());
+            }
         }
 
         pub const span = @compileError("deprecated: use `items` field directly");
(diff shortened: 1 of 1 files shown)


andrewrk:
I'm definitely interested in solving this use case, however before merging this PR I want to look into solving it at a deeper level (perhaps in the allocator) rather than putting special case code into ArrayList to work around the issue.

SpexGuy:
I made #6881 a while back to fix a bunch of these issues in the allocator, but ultimately I closed it because it's mostly workarounds for #6706 and related issues like #6983, #6947, #6937, #6936, and #6861.

marler8997:
Just FYI, when I implemented the allocator refactor, I recall some code assumed and depended on the allocations sizes passed to `allocFn` and `resizeFn` to be non-zero size.  I don't know if those assumptions are still baked into the interface/implementation but just know that at some point in time they were.

andrewrk:
Thank you @zbanks

Issue #2207, opened 2019-04-07
robust float parsing in the standard library


andrewrk:
Thanks to @tiehuis's work in #1958, #375 is closed and zig has basic float parsing in the standard library. But as he notes, there are the following deficiencies:

 * it doesn't handle some edge cases
 * error behavior is not zig-idiomatic
 * handling the edge cases might require an allocator
 * does not handle hex floats

I did a proof of concept of porting musl's float parsing code to implement f128 parsing in stage1. You can see this implementation in `parse_f128.c`. I believe that further porting this code to the Zig standard library would check all the above boxes.

This issue is the "deserialization" side of #1181. Note the goal stated in https://github.com/ziglang/zig/issues/1181#issuecomment-475910095: 

> I would like to propose that the default float printing and default float parsing in Zig should have roundtrip bit-for-bit value preservation for all values for all float types.

mattsta:
jumping in randomly here: for battle-tested float parsing, also check out the sqlite and luajit input parsers and compare their tradeoffs and performance characteristics (both are implemented to be more instruction cache friendly than stdlib versions too):

- sqlite float parser: https://github.com/mackyle/sqlite/blob/ae5d3aa91a794f6f1486b3f453ec44c0aa4c086b/src/util.c#L375-L571
- luajit float parser: https://github.com/LuaJIT/LuaJIT/blob/b025b01c5b9d23f6218c7d72b7aafa3f1ab1e08a/src/lj_strscan.c#L15-L327

and as for roundtrip bit-for-bit value preservation (ignoring things like f16 having 2k "unique" NaNs (and same-meaning-different-bits just gets worse with larger binary lengths)), the best I end up with when running my systems in "super paranoid mode" is: parse string to native float/double/long double, convert native float back to string, string compare original string to see if match is exact.

But, now an entire new decision opens up: which reliable float pretty print algorithm will conversions be tested against?
- https://www.cs.indiana.edu/~dyb/pubs/FP-Printing-PLDI96.pdf
- http://kurtstephens.com/files/p372-steele.pdf
- https://cseweb.ucsd.edu/~lerner/papers/fp-printing-popl16.pdf
- https://www.cs.tufts.edu/~nr/cs257/archive/florian-loitsch/printf.pdf

tiehuis:
[...]

PR #8634, opened 2021-04-28
std: Add a parser for hexadecimal floating point numbers

LemonBoy:
Should be good enough to unblock progress on the stage2 compiler.

Unifying this parser and the regular one (and perhaps rewrite it, #2207)
is left as an exercise for the reader.

Much more readable than musl's `parse_128.c`, no?


andrewrk:
Squeaky clean :star_struck: 

I can almost smell that fresh lemony scent

Issue #8364, opened 2021-03-26
Disallow the use of the comptime keyword before an expression or statement when it is superfluous


alexnask:
Currently, `comptime` expressions can be superfluously added and even chained.  

```zig
var foo: comptime comptime u8 = comptime comptime comptime 0;
```

This allows people to insert the keyword even in explicit comptime contexts, for example the length of an array type or a type expression.  
While this is not necessarily bad, in my opinion it can only reinforce the confusion about which blocks/expressions are comptime by default and which are not.  

Furthermore, with the introduction of #1717, the following will be allowed: 
```zig
const foo = comptime fn () void {
    // ...
};
```
This is especially bad since it is easily confusable with functions of .Comptime calling convention.

rohlem:
I think I disagree with the premise, that is the issue this is trying to solve.
I don't think people are confused by the fact you can write `comptime` before an expression or statement that is, in fact, executed at comptime.
But then again, I'm not a newcomer, and I'm not involved in any communities of newcomers, so I maybe can't judge that by my own experience.

To elaborate however, from my perspective, if someone wrote `var foo: comptime u8` and expected it to mean `comptime var foo: u8`,
then to me this is a __fundamental misunderstanding__ of the modularity of Zig, and that __types are expressions__.

It is about __distinguishing what these concepts mean, and not the grammatical details__ of "where do I put the keyword".
Once they understand this, I think they would __necessarily see__ how `var foo: comptime u8` __couldn't possibly__ declare a `comptime var`,
because it is __already equivalent__ to `var foo: u8`.

(That being said, I also don't know any other language to feature `comptime var` as fluently as Zig does,
 so I'm not sure where newcomers would have gotten the idea to try this in the first place.)


Besides this uninformed opinion, here are some practical counterpoints to consider:
- While disallowing `comptime` in certain contexts can __inform the programmer while writing__ code,
  it __requires them to know these language details when reading__ code to correctly understand it.
  (This goes against both the Zen and the advertised "debug the program, not your knowledge of the language".)
- Restricting the use of `comptime` in certain places results in grammatical __asymmetry__, making code __less friendly to copy-pasting__.
  - In particular, scenarios where runtime and comptime evaluation are both valid options, I posit `comptime` is always the safer choice.
    (Less pertinent side note, I hope this will still change,
     but the current concept of comptime allocations are expressed by what results in a dangling stack pointer at runtime.
     So that's a particularly inconvenient footgun currently.)
[...]

PR #8635, opened 2021-04-28
stage2: implement #8364 (error on redundant comptime)

g-w1:


src/AstGen.zig
```diff
@@ -1451,10 +1455,13 @@ pub fn comptimeExpr(
...
 ) InnerError!Zir.Inst.Ref {
-    const prev_force_comptime = gz.force_comptime;
+    const astgen = gz.astgen;
+    if (gz.force_comptime) {
+        return astgen.failNode(node, "comptime here happens inside a comptime scope; it is redundant", .{});
+    }
     gz.force_comptime = true;
     const result = try expr(gz, scope, rl, node);
-    gz.force_comptime = prev_force_comptime;
+    gz.force_comptime = false;
```
andrewrk, line 1464: we need a `comptimeExprFromAst` for the one that comes from the `comptime` keyword, and a `comptimeExpr` for other callsites. Otherwise, we'll get false positives for this for this code for example: `comptime { _ = @field(a, "foo"); }`. I think if you look at all the callsites of `comptimeExpr` you will see what I mean.

Be sure to note this in the doc comments so that future contributors don't accidentally call the wrong one :)

src/AstGen.zig
```diff
@@ -1451,10 +1455,13 @@ pub fn comptimeExpr(
     rl: ResultLoc,
     node: ast.Node.Index,
 ) InnerError!Zir.Inst.Ref {
-    const prev_force_comptime = gz.force_comptime;
+    const astgen = gz.astgen;
+    if (gz.force_comptime) {
+        return astgen.failNode(node, "comptime here happens inside a comptime scope; it is redundant", .{});
```
andrewrk, line 1460:
```suggestion
        return astgen.failNode(node, "redundant comptime keyword in implicitly comptime scope", .{});
```

Error message wording principles:
 * Efficiency of words
 * Precise words

andrewrk (changes requested):

Issue #3779, opened 2019-11-26
Strings from builtin functions should be null terminated if needed


Vexu:
`@embedFile`, `@errorName`, `@memberName`, `@tagName` and `@typeName` should be able to return null terminated slices so this would work:
```zig
const name: [*:0]u8 = @tagName(E.SomeEnum);
```

andrewrk:
They should also be updated to return pointers to arrays rather than slices when the length is comptime known. It goes along with the same ideas behind #863.

jmc-88:
For the record, `@memberName` was removed in https://github.com/ziglang/zig/commit/45da72c5b64069b7d5238465130a50f96678a148 and https://github.com/ziglang/zig/commit/3458fb891d8c7072a9bff6a32db9f3105ab72aa4, though there's still a reference to it in a commented out code block: https://github.com/ziglang/zig/blob/85755c51d529e7d9b406c6bdf69ce0a0f33f3353/lib/std/event/channel.zig#L142

jmc-88:
Now that https://github.com/ziglang/zig/pull/9183 too has been merged, I think the only outstanding builtin would be `@typeInfo()`, which has been forked to https://github.com/ziglang/zig/issues/9182. Maybe this can be closed?

PR #8636, opened 2021-04-28
Change builtins to return a string literal

jmc-88:
Fixes issue #3779 (I think?).

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -7475,7 +7475,7 @@ test "main" {
       {#header_close#}
 
       {#header_open|@errorName#}
-      <pre>{#syntax#}@errorName(err: anyerror) []const u8{#endsyntax#}</pre>
+      <pre>{#syntax#}@errorName(err: anyerror) [:0]const u8{#endsyntax#}</pre>
       <p>
       This function returns the string representation of an error. The string representation
       of {#syntax#}error.OutOfMem{#endsyntax#} is {#syntax#}"OutOfMem"{#endsyntax#}.
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -8623,7 +8623,7 @@ test "integer truncation" {
       {#header_close#}
 
       {#header_open|@typeName#}
-      <pre>{#syntax#}@typeName(T: type) [N]u8{#endsyntax#}</pre>
+      <pre>{#syntax#}@typeName(T: type) *const [N:0]u8{#endsyntax#}</pre>
       <p>
       This function returns the string representation of a type, as
       an array. It is equivalent to a string literal of the type name.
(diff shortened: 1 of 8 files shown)


Vexu (changes requested):
`@tagName` and `@errorName` should both return a null terminated slice instead since they can be used with runtime known arguments and returning different types based on whether they are being executed at `comptime` or not is bad for generic code.

marler8997:
Looks like @andrewrk was the one to suggest that we use pointers to arrays when the length is comptime known (see https://github.com/ziglang/zig/issues/3779#issuecomment-558886346).  I can see how this could cause some issues with generic code, but I think sometimes this information is necessary, so I don't see much choice other than to expose it.

Vexu:
I assumed he was referring to `@embedFile` and `@typeName` since they are always `comptime` but I could be wrong.

jmc-88:
> Looks like @andrewrk was the one to suggest that we use pointers to arrays when the length is comptime known (see [#3779 (comment)](https://github.com/ziglang/zig/issues/3779#issuecomment-558886346)).

Yes, I took that comment as reference, and probably misunderstood it, and here we are. :smile: 

> `@tagName` and `@errorName` should both return a null terminated slice instead since they can be used with runtime known arguments and returning different types based on whether they are being executed at `comptime` or not is bad for generic code.

Just for extra n00b-friendly clarity, the desire is to ensure these builtins all return `[:0]const u8`?

Vexu:
This change should also apply to all the strings in `std.builtin.TypeInfo`, and yes `[:0]const u8` is the correct type.

Vexu:
Are you still interested in finishing this pr? I'd be happy to do it for you if not.

jmc-88:
Sorry, yeah, I'd like to work on this. I got busy with real life and now I have a scary rebase in front of me :smile: 

I'll update this PR as soon as possible and ping the thread.

jmc-88:
I have updated the PR and it seems to pass `test-behavior` locally, PTAL.

Vexu (commented):
Looks good so far, do you also want to make the change to `@typeInfo` and `@src`?

jmc-88:
> Looks good so far, do you also want to make the change to `@typeInfo` and `@src`?

I have local changes ready to update string fields contained by structures returned by these two to `[:0]const u8`.

But while applying changes to `@typeInfo()` I realized its `std.builtin.TypeInfo` output is also an input to `@Type()` and that could break existing code. For instance, this assignment fails since it uses a `[]const u8` as returned by `std.mem.tokenize()` and that can't implicitly cast to `[:0]const u8`: https://github.com/ziglang/zig/blob/e006281d79a31e10bdbc07b3e33427d719ec01df/test/behavior/bugs/6456.zig#L20

Would you say this is WAI?

Vexu:
Hmm, I didn't think of that. I see three ways to solve this:

1. Have separate types for `@typeInfo` and `@Type`, which would break even more code.
2. Leave things as they are, which kind of defeats the purpose of this change.
3. Add casts from `[]T -> [:t]T` and `*[N]T -> *[N:t]T` when the value is comptime known, which would make this change purely cosmetic.

I'll make an issue for this since this pr is already useful as is.

Issue #8051, opened 2021-02-22
std.Thread.Condition: error: no member named 'unlock' in struct 'std.Thread.Mutex'


johnLate:
Compiling

```zig
const std = @import("std");
pub fn main() !void {
    var mutex = std.Thread.Mutex{};
    var condition = std.Thread.Condition{};
    condition.wait(&mutex);
}
```
with `zig build-exe bug_m.zig` results in
```
/prefix/lib/zig/std/Thread/Condition.zig:154:14: error: no member named 'unlock' in struct 'std.Thread.Mutex'
        mutex.unlock();
             ^
```

This is in `AtomicCondition`, `pub fn wait(cond: *AtomicCondition, mutex: *Mutex) void`.

Note: `PthreadCondition` is not affected. (`zig build-exe -lpthread bug_m.zig`)

System information:
```
zig version                    | 0.8.0-dev.1158+0aef1faa8
git log -n 1 --format='%h %ci' | 0aef1faa8 2021-02-22 00:22:46 +0200
uname -m -o                    | x86_64 GNU/Linux
```

PR #8637, opened 2021-04-29
Fix AtomicCondition compile error ziglang/zig#8051

cescude:
This also changes the `wait(...)` signature to take a `Mutex.Impl.Held`
object instead of a plain `Mutex`, which IIUC is clearer to the intent
of the function (ie., you should be passing in a locked mutex, not just
any mutex).

For example:

    var mutex = Mutex{};
    var cond = Condition{};

    var held = mutex.acquire();
    defer held.release();

    while (bad_state == true) {
        // cond.wait(&mutex); // <-- Original
        cond.wait(&held);     // <-- Updated
    }

Not really a breaking change for code using `AtomicCondition` (since it
didn't compile before), but this does affect the other implementations
and would presumably break anything using that code as well.


LemonBoy:
cc @kprotty Don't know if you're still plan on overhauling the whole set of synchronization primitives but I'd like to get this merged.

Can you please add some basic tests, preferably not using any `sleep` call.

kprotty:
@LemonBoy still planning to overhaul sync primitives once #9070 is merged. I can try to PR some tests to this branch today. To be clear, not *using* `sleep` (i.e. increase the chance of something) or *relying* on `sleep` (i.e. test fails if scheduling doesn't align with delays) ?

LemonBoy:
> still planning to overhaul sync primitives once #9070 is merged. 

Awesome news!

> To be clear, not using sleep (i.e. increase the chance of something) or relying on sleep (i.e. test fails if scheduling doesn't align with delays) ?

The latter, VMs used in CI are slow and many tests where we tried to order the operations on two different threads using `sleep` started failing at random times.

kprotty:
@cescude PR which [Condition and Semaphore tests](https://github.com/cescude/zig/pull/1)

kprotty:
@cescude You're right in that the extra signal() in Semaphore.wait() can be removed, although i'm not sure how `ping` would release two threads. 

`ping` posts to `request` which increases the permits by 1 and does a wake up. Any previous waiters on `request` would sleep until this happens. Any subsequent waiters on `request` would see the non-zero permits, decrement it, and not do the extra signal since only one permit was added. 

`pong` can only post to `reply` after receiving a post() from `request`. And a post() also must wait for the acknowledgement of the `reply` before posting again. Therefore there should only ever be one thread running `pong` *or* `ping` at a given moment. Do I understand this correctly?

cescude:
Ah, ok, I think you're right.

On a different note, supposing that first call to `relay.ping()` executes on the main thread before any of the spawned threads are given a chance to run...could there be a problem of losing the first signal? (ie., main thread spawns threads, signals `self.request`, blocks on `self.reply`, and _then_ the spawn thread get cpu time & block on `self.request`).

andrewrk:
I'd like to get rid of the `Held` type entirely. It seemed like a good idea at first to prevent mistakes but it ultimately hinders composeability and I think it ended up being an unfortunate instance of overengineering.

cescude:
@andrewrk Curious as to what you're `thinking` wrt how this negatively affects composability--

When I think about "which operations are valid" on a locked vs unlocked mutex...there's no overlap (so it makes sense to me to have distinct types exposed for an API).

In getting rid of `Held`, would you consider releasing an unacquired mutex to be an error or a no-op?

kprotty:
Releasing a mutex without having an acquired handle is common for code which is split by asynchrony (e.g. lock in one function, unlock in another). Also, if you wish to acquire after releasing it, you'd need a reference to the initial Mutex. Having Held and \*Mutex both be the same thing underneath feels awkward compared to \*Mutex having release() along with acquire().

cescude:
> Releasing a mutex without having an acquired handle is common for code which is split by asynchrony (e.g. lock in one function, unlock in another).

:thinking: Should it be common, though? `pthread_mutex_unlock` can return `EPERM` if you try to unlock from a different thread, as does `ReleaseSRWLockExclusive`. IIUC this would only apply to a single-threaded async implementation; using a multi-threaded event loop and you might get `unreachable` w/ pthreads and whatever-the-heck win32 does when you violate a constraint.

> Having Held and *Mutex both be the same thing underneath feels awkward compared to *Mutex having release() along with acquire().

Maybe; it's not like we have a different type for open vs closed files, so I get that `Held` is a bit of a cute implementation. I guess I just think it clarifies intent & expectations nicely for a primitive that's prone to being coded wrong.

kprotty:
> pthread_mutex_unlock can return EPERM if you try to unlock from a different thread

Zig's stdlib Mutex doesn't (and shouldn't in the near future) have to be based on the OS' mutex implementation. Thread-bound critical section calls is mostly used internally for fair scheduling or priority inheritance, things our Mutex type currently don't guarantee.

> using a multi-threaded event loop and you might get unreachable

You can bind an asynchronous task to a thread in a multi-threaded event loop. But irrespective that part, the "asynchrony" here doesn't imply Zig's `async`. Instead it implies one where control flow is split with a continuation. A single-threaded example is using a C library which scans a data structure and calls a user provided callback with `void*`. It also happens when crossing API boundaries for a library which invokes into a single-threaded scripting language API.

>  I just think it clarifies intent & expectations nicely

It does for simple, linear-lifetime, RAII-like use cases but becomes cumbersome otherwise IMO. The unlocking/locking scenario requires storing two pointers to the same data structure for example. [A popular Rust mutex impl](https://amanieu.github.io/parking_lot/parking_lot/struct.Mutex.html#method.raw_unlock) also has a `release()` which side-steps the RAII handle. I feel as though `Held` could be justified as an abstraction if it hosted more context other than the Mutex reference itself (like `File` to `Dir`).

cescude:
> A single-threaded example is using a C library which scans a data structure and calls a user provided callback with `void*`. It also happens when crossing API boundaries for a library which invokes into a single-threaded scripting language API.
:+1: 

> It does for simple, linear-lifetime, RAII-like use cases but becomes cumbersome otherwise IMO. The unlocking/locking scenario requires storing two pointers to the same data structure for example. [A popular Rust mutex impl](https://amanieu.github.io/parking_lot/parking_lot/struct.Mutex.html#method.raw_unlock) also has a `release()` which side-steps the RAII handle. I feel as though `Held` could be justified as an abstraction if it hosted more context other than the Mutex reference itself (like `File` to `Dir`).

FWIW, it's interesting that the `parking_lot::Mutex` implementation supports both a "held" style API (`MutexGuard`) in addition to the in-place `raw_unlock()`.

Anyways, thanks for taking the time to discuss.

kprotty:
Np, this feels like it should be an issue so that more people could weight in on their preferences if they have any.

andrewrk:
Closing old draft

PR #8554, opened 2021-04-16
Stage2 whole file astgen

andrewrk:
This is an implementation of  #8516. It is a continuation of  #8514. There is much to be done before this branch passes the existing stage2 tests, but it is a major step forward in progress.

The goal of this branch is to get to the point where existing tests are passing, but with the main compilation being kicked off in the same way that it is with stage1: via an implicit `_ = @import("std");` which, in turn, imports start.zig, which, in turn, imports the root source file and decides how to handle the entry point.

There are a lot of miscellaneous things standing in the way of this, and this branch aims to tackle all of them! See individual commit messages for more info.

PR #8639, opened 2021-04-29
x: comptime bool -> comptime x: bool

jedisct1:


g-w1:
Hmm, does this mean that some of the parameters in this commit should be comptime instead of regular if they were `comptime T`? Eg `n: comptime bool` -> `comptime n: bool` instead of just `n: bool`? https://github.com/ziglang/zig/pull/8554/commits/c60d8f017ea698c5fe4c7c6046f8ca09f0b8bf1d
cc @andrewrk

andrewrk:
Even if that's the intent, not necessarily. Based on the fact that the code compiled, it meas `comptime` parameters were not semantically required. The person doing it probably thought they were improving performance, without measuring to find out if it was true. Busted!

PR #8649, opened 2021-04-30
x/os, x/net: layout tcp, ipv4/ipv6, and socket abstractions

lithdew:
The `Socket` abstraction was refactored to only comprise of methods that can be generically used/applied to all socket domains and protocols.

A more comprehensive IPv4/IPv6 module derived from @LemonBoy's earlier work was implemented under `std.x.os.IPv4` and `std.x.os.IPv6`. Using this module, one can then combine them together into a union for example in order to optimize memory usage when storing/dealing with socket addresses.

A `TCP.Client` and `TCP.Listener` abstraction is introduced that is one layer over the `Socket` abstraction, which isolates methods that can only be applied to a "client socket" and a "listening socket". All prior tests from the `Socket` abstraction, which all previously operated assuming the socket is operating via. TCP/IP, were moved. All TCP socket options were also moved into the `TCP.Client` and `TCP.Listener` abstractions respectively away from the `Socket` abstraction.

Some additional socket options from @LemonBoy's prior PR for Darwin were also moved in (i.e. NOSIGPIPE).

### DIFF
--- a/lib/std/x/net/TCP.zig
+++ b/lib/std/x/net/TCP.zig
@@ -1,0 +1,399 @@
+const std = @import("../../std.zig");
+
+const os = std.os;
+const fmt = std.fmt;
+const mem = std.mem;
+const testing = std.testing;
+
+const IPv4 = std.x.os.IPv4;
+const IPv6 = std.x.os.IPv6;
+const Socket = std.x.os.Socket;
+
+/// A generic TCP socket abstraction.
+const TCP = @This();
+
+/// A TCP client-address pair.
+pub const Connection = struct {
+    client: TCP.Client,
+    address: TCP.Address,
+
+    /// Enclose a TCP client and address into a client-address pair.
+    pub fn from(socket: Socket, address: TCP.Address) Connection {
+        return .{ .client = TCP.Client.from(socket), .address = address };
+    }
+
+    /// Closes the underlying client of the connection.
+    pub fn deinit(self: TCP.Connection) void {
+        self.client.deinit();
+    }
+};
+
+/// Possible domains that a TCP client/listener may operate over.
+pub const Domain = extern enum(u16) {
+    ip = os.AF_INET,
+    ipv6 = os.AF_INET6,
+};
+
+/// A TCP client.
+pub const Client = struct {
+    socket: Socket,
+
+    /// Opens a new client.
+    pub fn init(domain: TCP.Domain, flags: u32) !Client {
+        return Client{
+            .socket = try Socket.init(
+                @enumToInt(domain),
+                os.SOCK_STREAM | flags,
+                os.IPPROTO_TCP,
+            ),
+        };
+    }
+
+    /// Enclose a TCP client over an existing socket.
+    pub fn from(socket: Socket) Client {
+        return Client{ .socket = socket };
+    }
+
+    /// Closes the client.
+    pub fn deinit(self: Client) void {
+        self.socket.deinit();
+    }
+
+    /// Shutdown either the read side, write side, or all sides of the client's underlying socket.
+    pub fn shutdown(self: Client, how: os.ShutdownHow) !void {
+        return self.socket.shutdown(how);
+    }
+
+    /// Have the client attempt to the connect to an address.
+    pub fn connect(self: Client, address: TCP.Address) !void {
+        return self.socket.connect(TCP.Address, address);
+    }
+
+    /// Read data from the socket into the buffer provided. It returns the
+    /// number of bytes read into the buffer provided.
+    pub fn read(self: Client, buf: []u8) !usize {
+        return self.socket.read(buf);
+    }
+
+    /// Read data from the socket into the buffer provided with a set of flags
+    /// specified. It returns the number of bytes read into the buffer provided.
+    pub fn recv(self: Client, buf: []u8, flags: u32) !usize {
+        return self.socket.recv(buf, flags);
+    }
+
+    /// Write a buffer of data provided to the socket. It returns the number
+    /// of bytes that are written to the socket.
+    pub fn write(self: Client, buf: []const u8) !usize {
+        return self.socket.write(buf);
+    }
+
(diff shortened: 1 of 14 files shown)


lib/std/x/net/TCP.zig
```diff
@@ -0,0 +1,399 @@
...
+const fmt = std.fmt;
+const mem = std.mem;
+const testing = std.testing;
+
+const IPv4 = std.x.os.IPv4;
+const IPv6 = std.x.os.IPv6;
+const Socket = std.x.os.Socket;
+
+/// A generic TCP socket abstraction.
+const TCP = @This();
```
ifreund, line 13: Unless I'm blind, there are no toplevel fields in this file and it should therefore be called `tcp.zig` by convention.

lib/std/x/os/Socket.zig
```diff
@@ -23,19 +15,24 @@ pub fn init(domain: u32, socket_type: u32, protocol: u32) !Socket {
...
+/// Shutdown either the read side, write side, or all side of the socket.
 pub fn shutdown(self: Socket, how: os.ShutdownHow) !void {
     return os.shutdown(self.fd, how);
 }
 
 /// Binds the socket to an address.
-pub fn bind(self: Socket, address: net.Address) !void {
-    return os.bind(self.fd, &address.any, address.getOsSockLen());
+pub fn bind(self: Socket, comptime Address: type, address: Address) !void {
+    return os.bind(self.fd, @ptrCast(*const os.sockaddr, &address), @sizeOf(Address));
```
ifreund, line 35: This seems rather unsafe as the API allows passing literally any type which will then be cast to a sockaddr and passed directly to the kernel with no checks. Perhaps it would be better to define a union over the address types and accept that here?

lib/std/x/net/TCP.zig
lithdew: Gotcha, I'll get that done after CI checks out; waiting to see if any of the unit tests may fail on certain platforms.

lib/std/x/os/Socket.zig
lithdew: Thought about that, though it may run into the issue that @Aransentin brought up in #8590 where the union alone would unnecessarily be 100+ bytes like `std.net.Address`.

My thoughts were, as an alternative, allowing for an arbitrary `Address` type specified would allow for applications that i.e. operate only over TCP/IP to have addresses that only take up 28 bytes. Whereas, applications that operate only over Unix sockets would operate with addresses that take up ~100 bytes.

That being said, I agree that it is somewhat unsafe. Might you have any other alternatives in mind?

MasterQ32: I'd create a proper `sockaddr` locally from a zig style `Address` type instead of trying to retrofit the zig type into the syscall types. this will not work for all operating systems, as those might not even expose `sockaddr` as a type.

Introduce a function `Adress.toNative_sockaddr()` and create a local temporary `sockaddr`

lithdew: @MasterQ32 Thought about that design, though say for example we have this struct:

```zig
pub const Client = struct {
  address: std.x.net.Address,
  socket: std.x.os.Socket,
  frame: @Frame(Client.readLoop),
};
```

Suppose `Address.toNative_sockaddr()` gets called in `Client.readLoop()`.

This would mean an additional 16 (sockaddr_in) to 28 (sockaddr_in6) bytes would need to be allocated inside `Client` inside of `@Frame(Client.readLoop)` alongside the 28 bytes for the `std.x.net.Address` field inside `Client`.

As an alternative, in the case of `sockaddr` not being exposed as a type in some platforms, some comptime checks can be included to have functions that use `sockaddr` return an error.

In the end, it's overall a tradeoff between the amount of memory we want users to keep around for storing addresses vs. making a safer API. I'm happy to go with any of the approaches though.

lithdew: Went over with @MasterQ32 on his suggested approach. Previous comment I made about additional bytes being redundantly allocated with his approach alongside a `Client` is wrong.

As soon as CI checks out, I'll go with his approach :smile:.

lithdew: Finished implementing the new approach.

lithdew:
Noticed some errors from Azure CI. Gonna sign off for today and fix them tomorrow.

lib/std/x/net/tcp.zig
```diff
@@ -0,0 +1,388 @@
...
+
+const IPv4 = std.x.os.IPv4;
+const IPv6 = std.x.os.IPv6;
+const Socket = std.x.os.Socket;
+
+/// A generic TCP socket abstraction.
+const tcp = @This();
+
+/// A union of all eligible types of socket addresses over TCP.
+pub const Address = union(enum) {
```
marler8997, line 23: Seems weird that the Address struct is in a module named `tcp` when it can also be used for `udp` sockets or any other type of socket that leverages the IP protocol "InternetProtocol".

marler8997: How will the UDP Address type differ from the TCP address type?

marler8997: An IP address is orthogonal to UDP/TCP.  It' in a different layer of the packet stack.  There is no difference between an IP address used for a UDP packet and a TCP packet.  Having a different type for each one would be like having a different `File` type in the standard library for `.c` files and `.zig` files.

lithdew: Oh, my bad I misread your comment. Having a `IP.Address` rather than individual `TCP.Address` and `UDP.Address` modules makes sense. I'll make the change tomorrow; thanks!

lithdew: Pushed a new commit that moves `tcp.Address` to `ip.Address` with `ip` being a new module.

andrewrk:
Rebasing on master should fix the freebsd CI issue

lithdew:
Got CI passing all tests :smile:.

PR #8650, opened 2021-04-30
Handle EPERM and ELOOP in os.fstatat()

17dec:
Both observed in the wild on Linux 5.10. ELOOP is trivial to replicate, EPERM is sometimes returned when attempting to resolve symlinks in /proc.


ifreund (approved):
LGTM, Thanks!

PR #8657, opened 2021-04-30
Azure CI: Test std.* and the rest separately

jedisct1:
On CI, we have been running into OOM issues when running the test suite on Windows for quite some time.

Unfortunately, we are very close to having the same issues on Linux as well. Some additional comptime work immediately makes these builds fail as well.

Add a new `test-toolchain` step, that tests everything except `std.*` and documentation.

On CI, call `test-toolchain`, `test-std` and `docs` separately instead of the `test` big hammer that emcompasses all of them. The original `test` step remains available.

Change the special case we made for Windows to the same code as other platforms.

This is a stopgap measure that stage2 will eventually make useless.

Until then, it gives us some headroom.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -36,7 +36,7 @@ pub fn build(b: *Builder) !void {
     const docs_step = b.step("docs", "Build documentation");
     docs_step.dependOn(&docgen_cmd.step);
 
-    const test_step = b.step("test", "Run all the tests");
+    const toolchain_step = b.step("test-toolchain", "Run the tests for the toolchain");
 
     var test_stage2 = b.addTest("src/test.zig");
     test_stage2.setBuildMode(mode);
@@ -90,7 +90,7 @@ pub fn build(b: *Builder) !void {
     exe.install();
     exe.setBuildMode(mode);
     exe.setTarget(target);
-    test_step.dependOn(&exe.step);
+    toolchain_step.dependOn(&exe.step);
     b.default_step.dependOn(&exe.step);
 
     exe.addBuildOption(bool, "skip_non_native", skip_non_native);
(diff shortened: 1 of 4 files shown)


ci/azure/linux_script
```diff
@@ -65,7 +65,17 @@ make $JOBS install
...
+LOG=$(mktemp)
+for step in test-toolchain test-std docs; do
+  echo "* Running step: [$step]"
+  if ! release/bin/zig build $step -Denable-qemu -Denable-wasmtime 2>"$LOG" >&2; then
+    cat "$LOG" >&2
+    exit 1
+  fi
+  echo "  Done."
+done
+set -x
```
andrewrk, line 78: Here's what I don't like about this:
 * now I have to maintain control flow in bash
 * i hate bash
 * please don't make me maintain bash code
 * now we don't see progress of ongoing tests in the CI
 * if the CI hangs or some other problem happens before the `cat` we miss the output
 * this is not strictly necessary. we could just not do this and it would be fine.

build.zig
```diff
@@ -36,7 +36,7 @@ pub fn build(b: *Builder) !void {
     const docs_step = b.step("docs", "Build documentation");
     docs_step.dependOn(&docgen_cmd.step);
 
-    const test_step = b.step("test", "Run all the tests");
+    const toolchain_step = b.step("test-toolchain", "Run the tests for the toolchain");
```
andrewrk, line 39: Let's please keep "test" to do the same thing it always did, but we can additionally introduce `test-toolchain` to be the subset that you have picked out.

andrewrk (changes requested):
Thanks for the stopgap solution to this problem. I'm on board with `test-toolchain` approach (with the adjustment mentioned in the review) but I'd like to push back on the changes to the CI scripts regarding logging.

ci/azure/linux_script
marler8997: The other nice thing about always enabling logging, is you can easily compare the output from success/failure cases.

build.zig
jedisct1: `test` is `toolchain_step`+`std_step`+`docs_step`, so it should still do the same thing is always did.

ci/azure/linux_script
jedisct1: Thanks for your input! I wasn't sure about that change. The logging part has been reverted. Thanks!

build.zig
andrewrk: Ahh missed that before. Thanks!

andrewrk:
Thanks!

PR #8661, opened 2021-05-01
Stringify enum via `@tagName` when `jsonStringify` is not defined

Jarred-Sumner:
This makes it so if you pass `std.json.stringify` an `enum` and the `enum` does not define a `jsonStringify` function, it writes the enum's `@tagName` as a string instead of a `@compileError`.

Whether it should default to label or the value is a question. People typically choose JSON because it's easy and they want to move on to the next task – certainly not because they're trying to save space. Since this change only applies when stringifying JSON, it seems like `@tagName` is a better convention to choose.

In this specific case, I wanted to print a very complicated object as JSON to help with debugging. 

I am creating this as a draft PR first to make sure the tests pass before marking it as ready for review. I manually tested this by editing the version of `json.zig` in the build folder and running code locally.

Note: this does does not handle parsing.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -2595,9 +2595,11 @@ pub fn stringify(
         .Enum => {
             if (comptime std.meta.trait.hasFn("jsonStringify")(T)) {
                 return value.jsonStringify(options, out_stream);
+            } else {
+                try out_stream.writeByte('"');
+                try out_stream.writeAll(@tagName(value));
+                return out_stream.writeByte('"');
             }
-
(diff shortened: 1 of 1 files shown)


daurnimator:
This was intentionally not included, as there is no general industry agreement on if enums should be serialised in json as their name or their value/integer.
If anything, because of non-exhaustive enums and extern enums (where duplicate fields are permitted) they should be serialised as their integer value.



> In this specific case, I wanted to print a very complicated object as JSON to help with debugging.

For debugging use `std.debug.print`

Vexu:
As daurnimator said there is no general agreement on how this should work, so if you want to pursue this please do it via a proposal.

Issue #8292, opened 2021-03-19
Proposal: C-style for loop


zzyxyzz:
This is a proposal to give Zig's `while` loop a little nudge in the direction of feature parity with C's `for` loop, the lack of which is a frequently bemoaned (#5070, #8019, #358, #472, #8030).

Doing this would:
1. provide encapsulation for loop variables, thereby improving readability and safety.
2. reduce unnecessary annoyance for people coming from C and C++.

Proposed syntax:
```zig
loop (var i: i32 = 0; i < limit; i += step) {
    // do something
}
```

The `loop` keyword was chosen mostly for its lack of baggage: `while` would clash with C convention and `for` with Zig. But this is ultimately just a suggestion and open to alternative proposals.

The loop would support unwrapping just like the current `while`:
```zig
loop (var it = foo.iter(); it; it = it.next()) |val| {
    // do something
} else |err| {
    // do semething else
}
```

A simplified variant could also be permitted:
```zig
loop (cond) {
    // equivalent to current while loop
}
```

...at which point the `while` keyword might be dropped entirely.

matu3ba:
@zzyxyzz 

for-loops are used for iterating and while loops for more complex control flow.
Putting more than 3 things into a line also is bad for readability:
1.loop,2.variable stuff, 3.iterator, 4. condition, 5.update, (6. variable name alias = zig special)

Further more one wants to have the initial condition/variable (as precondition of the loop) being explicit (if it is not trivial from the iterator) for simpler code reasoning.

zzyxyzz:
@matu3ba 
> for-loops are used for iterating and while loops for more complex control flow.

Ideally, yes. But Zig's highly specialized `for` loop forces the use of `while` even for very simple control flow. In Zig it is a very common pattern to 1) declare a loop variable outside the loop, leaking it into the common scope, and 2) use a while loop with the update syntax for what is essentially very straight-forward iteration. This proposal aims to at least prevent the variable leakage (as an alternative to #8019 and explicit scope wrapping).

> Putting more than 3 things into a line also is bad for readability:
1.loop,2.variable stuff, 3.iterator, 4. condition, 5.update, (6. variable name alias = zig special)

Sometimes. But note that most of the items you mention are already allowed in Zig, and the resulting long lines are not uncommon in base. This proposal only adds the ability to pull in the loop variable into the same line. This should work fine for most loops, but if the line gets too long, it can be broken up.

[...]

Issue #358, opened 2017-05-03
Proposal: for on ranges


AndreaOrru:
```
for (a...b) |x, index| {
    ...
}
```

Where `a` and `b` can be chars, integers, anything that can define a range. This is also better syntax IMHO than:

```
{var i = 0; while (i < n; i += 1) {
    ...
}}
```

thejoshwolfe:
Previously when @andrewrk and I considered this syntax very early in Zig's development, one problematic point was the upperbound inclusivity/exclusivity. With an expression like `i < n`, there's no question that `n` is an exclusive upper bound, but with `a...b`, does that include `b`?

Surely the question can have an answer, and it should probably be exclusive, but the point is that the syntax doesn't clearly say that it's exclusive. And just to confuse things, I've seen languages (coco, for example) use two different syntaxes for `a..b` inclusive and `a...b` exclusive.

I don't think users will have a very good success rate for guessing whether the upper bound is inclusive or exclusive, which makes me dislike this proposal. That being said, iterating over numbers `0` to `n` (exclusive) is a very common pattern, even outside the usecase of array indexes, and I think this deserves more discussion.

raulgrell:
I wouldn't be against the two different tokens, but intuitively I'd say `a..b` is b exclusive and `a...b` is b inclusive. I think Rust and Wren do this too...

If we only keep `...`, I'd say keep it exclusive and consistent with how slices are made, ie array[0...1] has array.len == 1. Python's range() function has this property

andrewrk:
We do have the `[a...b]` slicing syntax, which is exclusive. So it should be clear from this.

On the other hand, an unsolved problem is that in switch statements, `...` is inclusive, for example:

```zig
switch (c) {
    'a'...'z' => {}, // inclusive
}
```

I have 2 ideas to solve this problem:

 * Change `...` syntax for slicing to `..` and keep it exclusive. Keep `...` syntax for switch statements, and keep it inclusive.
 * Change `...` syntax for slicing to `:` (like Python). Keep `...` syntax for switch statements, and keep it inclusive.

raulgrell:
> Change ... syntax for slicing to .. and keep it exclusive. Keep ... syntax for switch statements, and keep it inclusive.

I'm not sure how I feel about using the `:` as a slice. It always has something to do with types. `..` and `...` are more intuitively associated with ranges

raulgrell:
[...]

PR #8663, opened 2021-05-01
std: add range(usize)

nektro:
Many have requested an easier way to make an incrementing loop. In a way that doesn't dirty the namespace currently being worked in with an `i` variable with use of a while loop. Or suggesting a syntax change to the for loop or a new loop entirely. However, this is possible in userspace by taking advantage of the semantics of zero sized types.

As written in the doc comment it is used like so:
```zig
for (std.range(10)) |_, i| {
    // 'i' will increment from 0 -> 9
}
```

Justification: #358, #8292, and more

### DIFF
--- a/lib/std/std.zig
+++ b/lib/std/std.zig
@@ -92,6 +92,18 @@ pub const x = @import("x.zig");
 pub const zig = @import("zig.zig");
 pub const start = @import("start.zig");
 
+/// Use this as a way to increment an index using a for loop. Works with both
+/// runtime and comptime integers.
+///
+/// ```zig
+/// for (range(10)) |_, i| {
+///   // 'i' will increment from 0 -> 9
+/// }
+/// ```
+pub fn range(len: usize) []const u0 {
+    return @as([*]u0, undefined)[0..len];
+}
+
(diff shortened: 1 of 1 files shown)


andrewrk:
nah just use a while loop

Jarred-Sumner:
This would be nice with `inline` when iterating through lists with `comptime`-known length but runtime-known values. That also avoids the overhead of creating a temporary array since it would do that at comptime

Before:
```zig
const unroll_count = comptime 64 / @sizeOf(PackageID);

while (remaining.len > unroll_count) {
    comptime var i: usize = 0;
    inline while (i < unroll_count) : (i += 1) {
        installer.installPackage(remaining[i], comptime log_level);
    }
    remaining = remaining[unroll_count..];
}
```

After:
```zig
const unroll_count = comptime 64 / @sizeOf(PackageID);

while (remaining.len > unroll_count) {
    inline for (std.range(unroll_count)) |i| {
        installer.installPackage(remaining[i], comptime log_level);
    }
    remaining = remaining[unroll_count..];
}
```

nektro:
been available as a package at https://github.com/nektro/zig-range for some time now for anyone who still comes across this

Issue #8662, opened 2021-05-01
Proposal: add `-fshort-enums` support to translate-c


ehaas:
GCC & Clang have a command-line option `-fshort-enums` that can be used to control how many bytes are allocated for an enum type. It's equivalent to putting `__attribute__((packed))` on all enums. From the [GCC docs](https://gcc.gnu.org/onlinedocs/gcc-11.1.0/gcc/Structures-unions-enumerations-and-bit-fields-implementation.html#Structures-unions-enumerations-and-bit-fields-implementation): 

> Normally, the type is `unsigned int` if there are no negative values in the enumeration, otherwise `int`. If `-fshort-enums` is specified, then if there are negative values it is the first of `signed char`, `short` and `int` that can represent all the values, otherwise it is the first of `unsigned char`, `unsigned short` and `unsigned int` that can represent all the values."

Super basic demonstration, compare output of the following C program when compiled with `-fshort-enums` vs not:
```c
#include <stdio.h>
typedef enum { FOO } MyEnum;
int main(void) {
    printf("%d\n", (int)sizeof(MyEnum));
    return 0;
}
```

The relevance to translate-c is that enum sizes affect ABI compatibility. Currently translate-c always translates enums with an underlying type of `c_int` or `c_uint`, but this won't work if linking with code that was built with `-fshort-enums`. Short enums are fairly common in embedded systems where space is at a premium.

One way to solve this would be to add a special preprocessor symbol like `__ZIG_SHORT_ENUMS` (which can be used with `zig translate-c` via `-D__ZIG_SHORT_ENUMS`, and can be used with `@cImport` via `@cDefine("__ZIG_SHORT_ENUMS", {});`. translate-c could look for the existence of that symbol and use it to control whether enums are generated with the short enum algorithm described above. One downside to this approach is that it reifies the short enum decision into the generated code, so it wouldn't be possible to translate code once and have it automatically work across platforms.

andrewrk:
Not only for translate-c but I think we should also add support for it in the main CLI options as well, but make it clear that it applies only to compiled C code. I think if you go to implement this it will become clear that these features are intertwined.

ehaas:
Are you suggesting that instead of doing this in translate-c; have a CLI option `-fshort-enums` that controls the size of extern enums?

andrewrk:
Wait a minute, sorry, I just realized this should already be working, like this:

```
./zig translate-c test.c  -cflags -fshort-enums -- -lc
```

[...]

PR #8664, opened 2021-05-01
translate-c: Ensure extra_cflags are passed to clang

ehaas:
Fixes the issue identified in #8662

### DIFF
--- a/src/Cache.zig
+++ b/src/Cache.zig
@@ -11,6 +11,7 @@ const testing = std.testing;
 const mem = std.mem;
 const fmt = std.fmt;
 const Allocator = std.mem.Allocator;
+const Compilation = @import("Compilation.zig");
 
 /// Be sure to call `Manifest.deinit` after successful initialization.
 pub fn obtain(cache: *const Cache) Manifest {
@@ -61,7 +62,7 @@ pub const File = struct {
 pub const HashHelper = struct {
     hasher: Hasher = hasher_init,
 
-    const EmitLoc = @import("Compilation.zig").EmitLoc;
+    const EmitLoc = Compilation.EmitLoc;
 
     /// Record a slice of bytes as an dependency of the process being cached
     pub fn addBytes(hh: *HashHelper, bytes: []const u8) void {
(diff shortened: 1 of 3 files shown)


ehaas:
Is there a good way to write the test for this? translate-c behaves correctly now so it would just be a matter of comparing the output when the `-fshort-enums` flag is passed vs not, but I'm not sure where to put such a test.

I also noticed that the caching system does not notice that the cflag affects the output; is there more I need to do for that to work? For example if you translate a file without the flag, then translate it again with the flag, you get the same output. You have to delete the cache for it to translate correctly with a different flag.

andrewrk:
Have a look at the function `updateCObject` and how it handles `extra_flags`. There is a block of code there that should probably be extracted into a function and called from `cmdTranslateC` as well:

```zig
    {
        // Hash the extra flags, with special care to call addFile for file parameters.
        // TODO this logic can likely be improved by utilizing clang_options_data.zig.
        const file_args = [_][]const u8{"-include"};
        var arg_i: usize = 0;
        while (arg_i < c_object.src.extra_flags.len) : (arg_i += 1) {
            const arg = c_object.src.extra_flags[arg_i];
            man.hash.addBytes(arg);
            for (file_args) |file_arg| {
                if (mem.eql(u8, file_arg, arg) and arg_i + 1 < c_object.src.extra_flags.len) {
                    arg_i += 1;
                    _ = try man.addFile(c_object.src.extra_flags[arg_i], null);
                }
            }
        }
    }
```

ehaas:
Ok, translate-c and the cache seem to be working.

Next question - is there a way to use `-cflags` with `@cImport`? Or would the recommended practice be to use `zig translate-c` with the correct flags and then import the translated file?

andrewrk:
There is currently no way to do that, however I could see that being a reasonable enhancement to make to the `@cImport` feature, since you could have multiple `@cImport` in one compilation, and each could have different flags for their respective C headers. If you're interested in pursuing this, I suggest to open a new proposal for enhancing `@cImport` to support C compiler flags specific to that `@cImport` instance.

> Or would the recommended practice be to use `zig translate-c` with the correct flags and then import the translated file?

This would certainly solve the problem, and often can be a generally more preferable approach than `@cImport` because it removes a dependency on the compiler having LLVM extensions enabled, and, assuming the ABI is stable, allows some manual cleanup to the translated types, especially with regards to pointer safety and coercion.

andrewrk:
Looks great, thank you!

PR #8665, opened 2021-05-01
Miscellaneous patches

LemonBoy:
cc @koachan You may be interested in the many fixes for SPARC targets.


lib/std/os/linux.zig
```diff
@@ -502,8 +504,10 @@ pub fn pwrite(fd: i32, buf: [*]const u8, count: usize, offset: u64) usize {
             );
         }
     } else {
+        // Some architectures (eg. 64bit SPARC) pwrite is called pread64.
```
xackus, line 507:
```suggestion
        // Some architectures (eg. 64bit SPARC) pwrite is called pwrite64.
```

PR #8667, opened 2021-05-01
std.zig: handle -frandom-seed in NIX_CFLAGS_COMPILE

### LINKED ISSUE #8269: Zig fails to get the include paths on NixOS Unstable
## Problem:
NixOS fails to detect system header locations on Unstable, accompanied by the following error:
```
warning: Unrecognized C flag from NIX_CFLAGS_COMPILE: -frandom-seed=<some 10 digit string>
```
Tested on:
```
$ nixos-version
21.05pre275822.916ee862e87 (Okapi)
```

## How to reproduce:
1. Prepare a Zig project that interfaces with C code, such as the [the raylib example code on the main page](https://ziglang.org/learn/samples/#c-interoperability).
2. Use `nix-shell` to install the required dependencies (`raylib` in this case).
3. Attempt to compile the file (`zig build-exe cimport.zig -lc -lraylib` as per the code comment).

## The cause:
https://github.com/NixOS/nixpkgs/pull/102251 added `-frandom-seed` as one of the flags of `NIX_CFLAGS_COMPILE`.

An example of what happens after running `nix-shell -p raylib`:
```
$ echo $NIX_CFLAGS_COMPILE
-frandom-seed=k1in0vqnyx -isystem /nix/store/ndbz36pb6rl0n2582n6598az61bsa7pl-raylib-3.5.0/include -isystem /nix/store/ndbz36pb6rl0n2582n6598az61bsa7pl-raylib-3.5.0/include
```
Since Zig doesn't expect this, it breaks out of the loop described [here](https://github.com/ziglang/zig/blob/4ca1f4ec2e3ae1a08295bc6ed03c235cb7700ab9/lib/std/zig/system.zig#L55) and doesn't properly add the header paths.

## Possible Solutions:
Zig should adds support for passing `-frandom-seed`. Should be the ideal fix? Should just be about working on that loop mentioned above.

Also, idk if Zig is supposed to have feature parity with Clang, but `frandom-seed` seems to be mentioned [here](https://github.com/ziglang/zig/blob/a502c160cd51ce3de80b3be945245b7a91967a85/src/clang_options_data.zig#L5143).

___

According to https://github.com/NixOS/nixpkgs/issues/109033, NixOS currently lacks a way to disable this flag from being added.

As such, for people who just want to "ignore" the problem, they can add something like this to the `shell.nix` file:
```
  # This is a workaround that needs to be revised
  shellHook = ''
    NIX_CFLAGS_COMPILE="$(echo "$NIX_CFLAGS_COMPILE" | sed -e "s/-frandom-seed=[^-]*//")"
  '';
```

sagehane:
Closes #8269 by ignoring the value of the `-frandom-seed` flag.

It would be nice to have it addressed before the upcoming NixOS release (21.05).

### DIFF
--- a/lib/std/zig/system.zig
+++ b/lib/std/zig/system.zig
@@ -46,6 +46,9 @@ pub const NativePaths = struct {
             var it = mem.tokenize(nix_cflags_compile, " ");
             while (true) {
                 const word = it.next() orelse break;
+                if (mem.startsWith(u8, word, "-frandom-seed=")) {
+                    continue;
+                }
                 if (mem.eql(u8, word, "-isystem")) {
                     const include_path = it.next() orelse {
                         try self.addWarning("Expected argument after -isystem in NIX_CFLAGS_COMPILE");
--- a/lib/std/zig/system.zig
+++ b/lib/std/zig/system.zig
@@ -46,9 +46,6 @@ pub const NativePaths = struct {
             var it = mem.tokenize(nix_cflags_compile, " ");
             while (true) {
                 const word = it.next() orelse break;
-                if (mem.startsWith(u8, word, "-frandom-seed=")) {
-                    continue;
-                }
                 if (mem.eql(u8, word, "-isystem")) {
                     const include_path = it.next() orelse {
                         try self.addWarning("Expected argument after -isystem in NIX_CFLAGS_COMPILE");
@@ -56,8 +53,10 @@ pub const NativePaths = struct {
                     };
                     try self.addIncludeDir(include_path);
                 } else {
+                    if (mem.startsWith(u8, word, "-frandom-seed=")) {
+                        continue;
+                    }
                     try self.addWarningFmt("Unrecognized C flag from NIX_CFLAGS_COMPILE: {s}", .{word});
-                    break;
                 }
             }
         } else |err| switch (err) {


LemonBoy:
I think a better and more future-proof solution would be to make the unrecognized flag error a non-fatal one.

sagehane:
I guess simply removing the `break` from the unrecognized flag case would work. This makes sure the compiler keeps going after coming across an unrecognized flag, as opposed to just aborting.
```diff
diff --git a/lib/std/zig/system.zig b/lib/std/zig/system.zig
index 42099c6ef..85b5da38e 100644
--- a/lib/std/zig/system.zig
+++ b/lib/std/zig/system.zig
@@ -54,7 +54,6 @@ pub const NativePaths = struct {
                     try self.addIncludeDir(include_path);
                 } else {
                     try self.addWarningFmt("Unrecognized C flag from NIX_CFLAGS_COMPILE: {s}", .{word});
-                    break;
                 }
             }
         } else |err| switch (err) {
```
However, that would cause it to always leave a message like this:
```
$ ../zig/build/bin/zig build
warning: Unrecognized C flag from NIX_CFLAGS_COMPILE: -frandom-seed=yymb0qgbfh
```
I'm fine with this as long as getting that warn message every time isn't deemed too annoying. Or, both methods can be applied so `-frandom-seed` is properly dealt with, and any new C flag NixOS introduces won't be fatal.

andrewrk:
Thanks @sagehane!

Issue #2206, opened 2019-04-07
make @import require a string literal rather than a comptime expression


andrewrk:
Zig's comptime evaluation is already powerful enough to allows conditional importing without having to have the `@import` target be a comptime expression. For example:

```zig
const target = if (builtin.mode == .Debug) "debug_mutex.zig" else "release_mutex.zig";
const mutex = @import(target);
```

This could be rewritten - and I would even say would be preferred to be rewritten - as:

```zig
const debug_mutex = @import("debug_mutex.zig");
const release_mutex = @import("release_mutex.zig");
const mutex = if (builtin.mode == .Debug) debug_mutex else release_mutex;
```

If import syntax is required to be trivially resolvable to either a file or a package name, then Zig can find all the source files for testing without this kind of thing:

https://github.com/ziglang/zig/blob/7c38651a65b48b8cc683bc31a40952d9b469ec4d/test/stage1/behavior.zig#L1-L19

There is still a problem though, because if zig ran all tests indiscriminately, this would stop working:

https://github.com/ziglang/zig/blob/7c38651a65b48b8cc683bc31a40952d9b469ec4d/std/os/linux.zig#L1537-L1541

The general problem here is that `@import` has compile-time side effects:
 * changes the set of unit tests
 * causes additional symbols to be exported, with `export` or `@export` - potentially conflicting.
 * could cause `@compileError` to trigger based on build mode or something else in top level `comptime` blocks.
 * could cause a `@cImport` which causes a compile error

It would be weird to activate these side effects for tests but not for other builds. And these side effects are important; they're not going away.

So I think the benefits of this proposal are not clear. But I think this is a sort of important topic in Zig so it deserves an issue that we can reference.

Vexu:
One thing to consider here is that the path might be a compile time variable from somewhere like ### [`@import("build_options")`]( https://github.com/SamTebbs33/pluto/blob/0ca3542fd27da59f1f1a791a0f1f13f00139c950/src/kernel/vmm.zig#L8).

andrewrk:
> One thing to consider here is that the path might be a compile time variable from somewhere like [`@import("build_options")`](https://github.com/SamTebbs33/pluto/blob/0ca3542fd27da59f1f1a791a0f1f13f00139c950/src/kernel/vmm.zig#L8).

```zig
const tag = @import("build_options").which_package_do_you_want;
const thing = switch (tag) {
    .one => @import("one.zig"),
    .two => @import("two.zig"),
}
```

[...]

PR #8670, opened 2021-05-02
Prepare std/crypto/pcurves for ecdsa and other curves

jedisct1:
Functions generated by Fiat-crypto are not prefixed by their description any more. This matches an upstream change.

We can now use a single type for different curves and implementations.

The field type is now generic, so we can properly handle the base field and scalars without code duplication.


zigazeljko:
FYI, since #2206 is accepted, that dynamic import (via `params.fiat_code`) is likely to break in the near future.

jedisct1:
Thanks for the heads up! The @import can easily be changed to be made at instantiation time instead. 

I'll do that change as soon as I get access to a keyboard. 

Thanks again !

jedisct1:
@zigazeljko Done, thank you!

Issue #8531, opened 2021-04-14
LLD with LTO enabled incorrectly drops _tls_index symbol, causing undefined reference errors


codepoint92:
I have reduced this error to this minimal example:

main.zig:
```
const std = @import("std");

const c = @cImport({
    @cInclude("test.h");
});

pub fn main() !void {
    const num = c.getNum();
    std.debug.print("Num: {}!", .{num});
}
```

test.h:
```
int getNum(void);
```

test.c:
```
int getNum(void) {
    return 3;
}
```

Build command: `zig build-exe main.zig test.c -I.`

Error:
```
lld-link: warning: ignoring debug info with an invalid version (0) in zig-cache\o\ee8a465508ce3f1d05048c0482c74ec9\main.obj
lld-link: error: undefined symbol: _tls_index
>>> referenced by main.exe.lto.obj:(std.debug.panicExtra)
>>> referenced by main.exe.lto.obj:(std.debug.panicExtra)
>>> referenced by main.exe.lto.obj:(std.debug.panicExtra)
error: LLDReportedFailure
```

This error does not occur in Debug or other release modes or when not using C source files.

Thank you for your help.

xxxbxxx:
bisected to e9e91b4ed058f4c4e3f3380ec06cb914becd04a8 
(when LTO was enabled by default)

xxxbxxx:
moreover:

removing `-OPT:lldlto=3 ` from the linker command line
`lld-link -ERRORLIMIT:0 -NOLOGO -DEBUG -OPT:lldlto=3 -STACK:16777216 -MACHINE:X64 [...]`
still fails.

But if I do:

- `zig build-exe main.zig test.c -lc -target x86_64-windows-gnu -OReleaseFast -fno-lto --verbose-link` (1)  -> succeeds
- `zig build-exe main.zig test.c -lc -target x86_64-windows-gnu -OReleaseFast -flto --verbose-link` (2) -> fails
- And then I take the lld-link command from (2) and only change the `mingw32.lib` file to use the one from (1)  -> succeeds.

michal-z:
It looks like undefined symbol (_tls_index) comes from std.debug.panicExtra. When I do not use call to panic() in my program it links correctly even with LTO. When I call panic (everything else stays the same) I get:

PS C:\Development\projects\zig-gamedev\samples\triangle1> zig build run -Drelease-fast=true
[...]

PR #8674, opened 2021-05-02
build: workaround link error with LTO and mingw

xxxbxxx:
The symbol "_tls_index" gets lost when using LTO.
Disabling LTO only on the object file that defines it allows the link to work.

I have no idea what I'm doing :)
I tried a few random things in `tlssup.c` to see if the symbol would stay, with no success.
Short of actually figuring out the issue, maybe it would be cleaner to just drop LTO on the whole mingw32.lib?

I also tried to test with x86_64-windows-msvc, but no idea how to make it work at all when linking with libC.

fixes https://github.com/ziglang/zig/issues/8531

Maybe related:
as soon as a xxx.c file is linked (even if no symbols are pulled from it), I get a warning:
```
warning(link): unexpected LLD stderr:
lld-link: warning: ignoring debug info with an invalid version (0) in N:\zigdbg\zig-cache\o\03a5f823f3c02149079e56dc64f92fb2\zigdbg.o
```

### DIFF
--- a/src/mingw.zig
+++ b/src/mingw.zig
@@ -92,6 +92,10 @@ pub fn buildCRTFile(comp: *Compilation, crt_file: CRTFile) !void {
                     "-D_WIN32_WINNT=0x0f00",
                     "-D__MSVCRT_VERSION__=0x700",
                 });
+                if (std.mem.eql(u8, dep, "tlssup.c")) {
+                    // Can't let LTO drop symbols defined in this file (eg: _tls_index)
+                    try args.append("-fno-lto");
+                }
                 c_source_files[i] = .{
                     .src_path = try comp.zig_lib_directory.join(arena, &[_][]const u8{
                         "libc", "mingw", "crt", dep,
--- a/src/mingw.zig
+++ b/src/mingw.zig
@@ -92,8 +92,10 @@ pub fn buildCRTFile(comp: *Compilation, crt_file: CRTFile) !void {
                     "-D_WIN32_WINNT=0x0f00",
                     "-D__MSVCRT_VERSION__=0x700",
                 });
-                if (std.mem.eql(u8, dep, "tlssup.c")) {
-                    // Can't let LTO drop symbols defined in this file (eg: _tls_index)
+                if (std.mem.eql(u8, dep, "tlssup.c") and comp.bin_file.options.lto) {
+                    // LLD will incorrectly drop the `_tls_index` symbol. Here we work
+                    // around it by not using LTO for this one file.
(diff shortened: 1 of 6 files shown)


LemonBoy:
That's the correct solution. `_tls_index` is not referenced by anything within the mingw32 library itself and is thus dropped during the LTO process, while we want for that symbol to end up in the final object.

Linking to a github issue in the comment is not that useful, writing a few words on why it's being disabled is much better
Eg. ```// Can't let LTO drop any symbol defined in here```

xxxbxxx:
> _tls_index is not referenced by anything within the mingw32 library itself and is thus dropped during the LTO process, while we want for that symbol to end up in the final object

mmm. I don't get it.  
I mean mingw32.lib is just aggregated not-yet-linked object files (either machinecode or IR dumps in lto)
And then when the final program is linked, then the symbol is actually referenced (like others from the lib). Actually most symbols are like this in static libraries, so why this one is different?
Is it because it is used internally by the compiler/linker for some magical TLS stuff, and not by the actual executable code?

LemonBoy:
> Is it because it is used internally by the compiler/linker for some magical TLS stuff, and not by the actual executable code?

After some cursory testing it seems that the external symbols introduced by LLVM during the expansion of the thread-local accesses are not taken into account when computing the LTO module index.

This is probably worth reporting upstream.

xxxbxxx:
I've added a test for the issue. (linking a c file and a zig file,  with threadlocal variables)

All is fine in debug builds.

However, in release builds:
 - macos fails to link, with `error: MissingMainEntrypoint`

 - windows-gnu gives a warning
```
warning(link): unexpected LLD stderr:
lld-link: warning: ignoring debug info with an invalid version (0) in D:\a\1\s\test\standalone\mix_c_files\zig-cache\o\adbf700d0f61e162e41124f3f0fce448\test.obj
```

 - freebsd also has a warning:
```
warning(link): unexpected LLD stderr:
ld.lld: warning: linking module flags 'Dwarf Version': IDs have conflicting values ('i32 2' from /usr/home/build/zig/test/standalone/mix_c_files/zig-cache/o/9589d9679386de4941703ed5f244b416/test.o with 'i32 4' from ld-temp.o)
```

I guess I can disable the test on macos, but I don't know about the debug info warnings.

xxxbxxx:
Also, the test links with libC, but it's not actually used. 

However if I remove it, of course, the workaround for windows release builds is not used, and once again the linker can't seem to find `_tls_index` ...

andrewrk:
Thank you for this work, and apologies for the delayed merge. I'm adding a comment on #8531 now.

PR #8681, opened 2021-05-04
c.zig: fix waitpid() definition

codic12:
The definition for std.c.waitpid currently looks like this:
`pub extern "c" fn waitpid(pid: c_int, stat_loc: *c_uint, options: c_uint) c_int;`
However, this is inconsistent with the proper definition in libc, which is:
`pid_t waitpid(pid_t pid, int *stat_loc, int options);`

Notice that the proper function signature returns pid_t and takes pid_t as it's first parameter. The second parameter should also be a C pointer. While it isn't standardized as far as I can tell, both glibc and the musl bundled with Zig use `int` instead of `unsigned int`, so I've also changed it to c_int.

### DIFF
--- a/lib/std/c.zig
+++ b/lib/std/c.zig
@@ -103,7 +103,7 @@ pub extern "c" fn linkat(oldfd: fd_t, oldpath: [*:0]const u8, newfd: fd_t, newpa
 pub extern "c" fn unlink(path: [*:0]const u8) c_int;
 pub extern "c" fn unlinkat(dirfd: fd_t, path: [*:0]const u8, flags: c_uint) c_int;
 pub extern "c" fn getcwd(buf: [*]u8, size: usize) ?[*]u8;
-pub extern "c" fn waitpid(pid: c_int, stat_loc: *c_uint, options: c_uint) c_int;
+pub extern "c" fn waitpid(pid: pid_t, stat_loc: ?*c_int, options: c_int) pid_t;
 pub extern "c" fn fork() c_int;
 pub extern "c" fn access(path: [*:0]const u8, mode: c_uint) c_int;
 pub extern "c" fn faccessat(dirfd: fd_t, path: [*:0]const u8, mode: c_uint, flags: c_uint) c_int;
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -3435,10 +3435,10 @@ pub const WaitPidResult = struct {
 };
 
 pub fn waitpid(pid: pid_t, flags: u32) WaitPidResult {
-    const Status = if (builtin.link_libc) c_uint else u32;
+    const Status = if (builtin.link_libc) c_int else u32;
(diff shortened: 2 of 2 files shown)


lib/std/c.zig
```diff
@@ -105,7 +105,7 @@ pub extern "c" fn linkat(oldfd: fd_t, oldpath: [*:0]const u8, newfd: fd_t, newpa
 pub extern "c" fn unlink(path: [*:0]const u8) c_int;
 pub extern "c" fn unlinkat(dirfd: fd_t, path: [*:0]const u8, flags: c_uint) c_int;
 pub extern "c" fn getcwd(buf: [*]u8, size: usize) ?[*]u8;
-pub extern "c" fn waitpid(pid: c_int, stat_loc: *c_uint, options: c_uint) c_int;
+pub extern "c" fn waitpid(pid: pid_t, stat_loc: [*c]c_int, options: c_int) pid_t;
```
ifreund, line 108: `[*c]` pointers should never be used outside of generated code. Instead, use the pointer type that accurately describes the semantics of the API.
```suggestion
pub extern "c" fn waitpid(pid: pid_t, stat_loc: ?*c_int, options: c_int) pid_t;
```

ifreund (changes requested):
CI says that you need to add a cast in `std.os.waitpid()` due to the switch from `c_uint` to `int` as well.

lib/std/c.zig
codic12: Fixed that, thanks.

ifreund:
> Does this cast just entail switching
> ` const Status = if (builtin.link_libc) c_uint else u32;`
> to
> ` const Status = if (builtin.link_libc) c_int else u32;`
> 
> in line 3363, os.zig?

Yep, that should do it.

codic12:
Let's see if CI passes now.

lib/std/os.zig
```diff
@@ -3360,7 +3360,7 @@ pub const WaitPidResult = struct {
 };
 
 pub fn waitpid(pid: pid_t, flags: u32) WaitPidResult {
-    const Status = if (builtin.link_libc) c_uint else u32;
+    const Status = if (builtin.link_libc) c_int else i32;
```
ifreund, line 3363: changing this to `i32` is not correct, see `std.os.linux.waitpid`
```suggestion
    const Status = if (builtin.link_libc) c_int else u32;
```

codic12: i see, fixed

codic12:
CI failed. I fixed with an intCast, but I think this could lead to some unwanted behavior. Maybe I should change it back to uint; thoughts?

codic12:
Oh, as I expected, intCast doesn't work

ifreund:
> Oh, as I expected, intCast doesn't work

Indeed, the `@intCast()` must only be done if linking libc.

PR #8682, opened 2021-05-04
c.zig: add sigfillset, alarm, sigwait

codic12:
I've added these three functions to all switches except NetBSD, because I don't know what the function is named there. I've even added it on the .windows switch since all the posix functions seem to be there anyways. That probably needs some testing and cleaning up.


ifreund (changes requested):
Haven't looked in detail yet, but you should use the proper zig pointer type for these arguments, not `[*c]` which is only for use in generated code (e.g. from translate-c).

codic12:
Ah, I was also using `lc.` from my own code in addition to that. I've fixed that

Vexu:
A quick search in [NetBSD man pages](https://man.netbsd.org) suggests that it has all of these functions so they should probably just be moved out of the switch.

PR #8683, opened 2021-05-04
Initial bringup for Linux/Thumb2

LemonBoy:
@andrewrk I'd love your input on the XXX comment in `system.zig`

### DIFF
--- a/lib/std/zig/system.zig
+++ b/lib/std/zig/system.zig
@@ -349,6 +349,15 @@ pub const NativeTargetInfo = struct {
                     }
                 }
             },
+            .arm, .armeb => {
+                // XXX What do we do if the target has the noarm feature?
+                //     What do we do if the user specifies +thumb_mode?
+            },
+            .thumb, .thumbeb => {
+                result.target.cpu.features.addFeature(
+                    @enumToInt(std.Target.arm.Feature.thumb_mode),
+                );
+            },
             else => {},
         }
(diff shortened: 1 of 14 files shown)


lib/std/zig/system.zig
```diff
@@ -349,6 +349,15 @@ pub const NativeTargetInfo = struct {
                     }
                 }
             },
+            .arm, .armeb => {
+                // XXX What do we do if the target has the noarm feature?
+                //     What do we do if the user specifies +thumb_mode?
```
andrewrk, line 354: The fundamental problem here is that we have multiple different ways of representing the same state.

Let's try to eliminate redundancies. Here are some (breaking) ideas:

 * Remove the thumb sub-architecture in favor of the `thumb_mode` CPU feature. Instead of `-target thumb-linux` it would be `-target arm-linux -mcpu=generic+v7m`
 * Delete the `noarm` CPU feature since it appears to be completely redundant with `thumb_mode`. All feature sets that include `thumb_mode` also include `noarm`.

After a few minutes considering this I think I would be in favor of both these ideas.

LemonBoy: The former makes more sense, `noarm` is effectively a CPU feature that signals no interworking is possible at all.
Clang takes a similar approach and requires `-mthumb` or `-mno-thumb` to be specified, that's another possible solution to take into account.

PR #8686, opened 2021-05-04
Allow tests to fail

### LINKED ISSUE #8629: allow tests to fail
I propose to modify the default test runner so that when a test returns an error, the test is marked as failure, however subsequent tests are still run. At the end will be reported the failures and successes, and the final exit code of `zig test` will be 0 if and only if all tests succeeded with no memory leaks.

Follow-up proposal: modify `std.testing.expectEqual` and similar functions to return `error.ZigTestFailed` rather than calling panic.

Motivation is to increase developer iteration speed; it might help to find out more than one thing is broken at a time.

Vexu:
Closes #8629

There sure are a lot of tests


andrewrk:
> There sure are a lot of tests

:joy:

lib/std/special/test_runner.zig
```diff
@@ -78,6 +79,12 @@ pub fn main() anyerror!void {
...
             },
+            error.ZigTestFailed => {
+                fail_count += 1;
+                test_node.end();
+                progress.log("{s}...FAIL\n", .{test_fn.name});
+                if (progress.terminal == null) std.debug.print("FAIL\n", .{});
+            },
             else => {
                 progress.log("", .{});
                 return err;
```
andrewrk, line 90: This should also count as a failure rather than aborting the testing process.

I suggest to make the `std.testing.expectFoo` functions return `error.TestOutputMismatch` or something like that. Then put the "zig test failed" logic in the else prong here, and print the error name rather than "FAIL".

andrewrk (changes requested):
Thanks for doing this!

andrewrk:
Oops, just noticed this is a draft. Apologies for the premature review.

andrewrk:
yeah!!

BarabasGitHub:
What's the recommended way of updating existing code for this? I have so many tests... D=

Vexu:
> What's the recommended way of updating existing code for this?

Basically just doing what I did in this pr, slap a bunch of `try`s in front of calls to `std.testing.expect*` and everything should just work.

BarabasGitHub:
Yeah kind of... definitely not that easy if you wrote any kind of `expectEqual` functions yourself. Especially not when they're recursive (like with a tree).
I'll have to update this for some other time.

MasterQ32:
```
find -name "*.zig" -exec sed -i 's/std.testing.expect/try std.testing.expect/g' '{}'
```
Should update most things correctly :grin:
Self-made recursive test functions shouldn't be a problem either, you just have to slap the explicit error set (which is well-known) onto you functions instead of inferring the error set

PR #8691, opened 2021-05-05
x/net: fix tcp tests for openbsd and add missing `fmt` import

lithdew:
On OpenBSD, connecting to a newly-allocated port on an unspecified IPv4 host address causes EINVAL. This was found by @mikdusan when running TCP tests on OpenBSD.

This commit fixes TCP tests on OpenBSD by having all tests that allocate a new host-port pair to have the host IPv4/IPv6 address point to the host's loopback adapter (on localhost).

### DIFF
--- a/lib/std/x/net/ip.zig
+++ b/lib/std/x/net/ip.zig
@@ -6,6 +6,8 @@
 
 const std = @import("../../std.zig");
 
+const fmt = std.fmt;
+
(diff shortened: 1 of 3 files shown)


mikdusan (approved):
- `zig test lib/std/x.zig` now passes on OpenBSD 6.9

PR #8695, opened 2021-05-05
std.os: WSAStartup is now called upon socket creation when needed

Bxil:
std.os appears to be an OS agnostic wrapper, but it still requires WSAStartup to be manually called on windows to create sockets successfully. This PR is intended to get rid of that.

### DIFF
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -1261,7 +1261,7 @@ pub fn WSAStartup(majorVersion: u8, minorVersion: u8) !ws2_32.WSADATA {
             .WSASYSNOTREADY => return error.SystemNotAvailable,
             .WSAVERNOTSUPPORTED => return error.VersionNotSupported,
             .WSAEINPROGRESS => return error.BlockingOperationInProgress,
-            .WSAEPROCLIM => return error.SystemResources,
+            .WSAEPROCLIM => return error.ProcessFdQuotaExceeded,
             else => |err| return unexpectedWSAError(err),
         },
     };
@@ -1280,6 +1280,30 @@ pub fn WSACleanup() !void {
     };
 }
 
+var wsa_startup_mutex: std.Thread.Mutex = .{};
+
+/// Microsoft requires WSAStartup to be called to initialize, or else
+/// WSASocketW will return WSANOTINITIALISED.
+/// Since this is a standard library, we do not have the luxury of
+/// putting initialization code anywhere, because we would not want
+/// to pay the cost of calling WSAStartup if there ended up being no
+/// networking. Also, if Zig code is used as a library, Zig is not in
+/// charge of the start code, and we couldn't put in any initialization
+/// code even if we wanted to.
+/// The documentation for WSAStartup mentions that there must be a
+/// matching WSACleanup call. It is not possible for the Zig Standard
+/// Library to honor this for the same reason - there is nowhere to put
+/// deinitialization code.
+/// So, API users of the zig std lib have two options:
+///  * (recommended) The simple, cross-platform way: just call `WSASocketW`
+///    and don't worry about it. Zig will call WSAStartup() in a thread-safe
+///    manner and never deinitialize networking. This is ideal for an
+///    application which has the capability to do networking.
+///  * The getting-your-hands-dirty way: call `WSAStartup()` before doing
+///    networking, so that the error handling code for WSANOTINITIALISED never
+///    gets run, which then allows the application or library to call `WSACleanup()`.
+///    This could make sense for a library, which has init and deinit
+///    functions for the whole library's lifetime.
 pub fn WSASocketW(
     af: i32,
     socket_type: i32,
@@ -1288,18 +1312,40 @@ pub fn WSASocketW(
     g: ws2_32.GROUP,
     dwFlags: DWORD,
 ) !ws2_32.SOCKET {
-    const rc = ws2_32.WSASocketW(af, socket_type, protocol, protocolInfo, g, dwFlags);
-    if (rc == ws2_32.INVALID_SOCKET) {
-        switch (ws2_32.WSAGetLastError()) {
-            .WSAEAFNOSUPPORT => return error.AddressFamilyNotSupported,
-            .WSAEMFILE => return error.ProcessFdQuotaExceeded,
-            .WSAENOBUFS => return error.SystemResources,
-            .WSAEPROTONOSUPPORT => return error.ProtocolNotSupported,
-            .WSANOTINITIALISED => return error.NotInitialised,
-            else => |err| return unexpectedWSAError(err),
+    var first = true;
+    while (true) {
+        const rc = ws2_32.WSASocketW(af, socket_type, protocol, protocolInfo, g, dwFlags);
+        if (rc == ws2_32.INVALID_SOCKET) {
+            switch (ws2_32.WSAGetLastError()) {
+                .WSAEAFNOSUPPORT => return error.AddressFamilyNotSupported,
+                .WSAEMFILE => return error.ProcessFdQuotaExceeded,
+                .WSAENOBUFS => return error.SystemResources,
+                .WSAEPROTONOSUPPORT => return error.ProtocolNotSupported,
+                .WSANOTINITIALISED => {
+                    if (!first) return error.Unexpected;
+                    first = false;
+
+                    var held = wsa_startup_mutex.acquire();
+                    defer held.release();
+
+                    // Here we could use a flag to prevent multiple threads to prevent
+                    // multiple calls to WSAStartup, but it doesn't matter. We're globally
+                    // leaking the resource intentionally, and the mutex already prevents
+                    // data races within the WSAStartup function.
+                    _ = WSAStartup(2, 2) catch |err| switch (err) {
+                        error.SystemNotAvailable => return error.SystemResources,
+                        error.VersionNotSupported => return error.Unexpected,
+                        error.BlockingOperationInProgress => return error.Unexpected,
+                        error.ProcessFdQuotaExceeded => return error.ProcessFdQuotaExceeded,
+                        error.Unexpected => return error.Unexpected,
+                    };
+                    continue;
+                },
+                else => |err| return unexpectedWSAError(err),
+            }
         }
+        return rc;
     }
-    return rc;
 }
 
 pub fn bind(s: ws2_32.SOCKET, name: *const ws2_32.sockaddr, namelen: ws2_32.socklen_t) i32 {
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -1295,6 +1295,7 @@ pub fn WSASocketW(
             .WSAEMFILE => return error.ProcessFdQuotaExceeded,
             .WSAENOBUFS => return error.SystemResources,
             .WSAEPROTONOSUPPORT => return error.ProtocolNotSupported,
+            .WSANOTINITIALISED => return error.NotInitialised,
             else => |err| return unexpectedWSAError(err),
         }
     }
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -2724,29 +2724,25 @@ pub const SocketError = error{
 
     /// The socket type is not supported by the protocol.
     SocketTypeNotSupported,
-
-    /// The environment for socket control has not been initialised.
-    NotInitialised,
 } || UnexpectedError;
 
 pub fn socket(domain: u32, socket_type: u32, protocol: u32) SocketError!socket_t {
     if (builtin.os.tag == .windows) {
-        // NOTE: windows translates the SOCK_NONBLOCK/SOCK_CLOEXEC flags into windows-analagous operations
+        // NOTE: windows translates the SOCK_NONBLOCK/SOCK_CLOEXEC flags into
+        // windows-analagous operations
         const filtered_sock_type = socket_type & ~@as(u32, SOCK_NONBLOCK | SOCK_CLOEXEC);
-        const flags: u32 = if ((socket_type & SOCK_CLOEXEC) != 0) windows.ws2_32.WSA_FLAG_NO_HANDLE_INHERIT else 0;
-        const rc = windows.WSASocketW(@bitCast(i32, domain), @bitCast(i32, filtered_sock_type), @bitCast(i32, protocol), null, 0, flags) catch |err| switch (err) {
-            error.NotInitialised => again: {
-                // Before a socket is made Windows requires WSAStartup to be called.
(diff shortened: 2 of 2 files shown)


marler8997:
I'm not sure if this solution is scalable.  There are many functions from ws2_32 that requires WSAStartup to be called, and client's aren't guaranteed to always call the socket function first.

g-w1:
Also, please format your code.

Bxil:
@marler8997 This solution works even if WSAStartup was called in some other manner. Maybe it might be worth it to have some kind of `ensureStartup` function, but I can't think of a clean way of doing that. If you have a superior suggestion I am all ears.

@g-w1 Whoops, done.

marler8997:
The current solution is the application needs to call wsastartup on windows.  This is fine because if they forget, they will get an error code indicating wsastartup was not called, then add the call to the program's entry point.

Injecting code into the standard library that tries to do "intelligent fallback initialization" adds quite a bit of complexity to using the standard library properly.  Imagine an application that depended on this behavior, but then suddenly they reorder a ws2_32 call to appear before their call to the "socket" function, now suddently their applications is broken.  Even worse, maybe there is a corner case where their application sometimes calls a different function first, so now their application only fails in this one corner case.

What's nice about the current design is the application ALWAYS fails.  ALWAYS FAILS is better than "SOMETIMES WORKS".

Bxil:
I would argue for the opposite: not having it like this raises complexity. Something that promises to be platform independent suddenly requires a platform dependent initialization which is intuitive. By extension the following doesn't hold:
> A bug that "always fails" is much better than a bug that "sometimes works".
Since this fails on Windows, and works just fine on any other platform.

I have also investigated some more. Initializing twice is apparently harmless, and both .NET and Rust implement this in a familiar way.

marler8997:
I agree that windows networking behaving differently than other platforms is a complexity.  However, you're solution to implicitly initialize WSA when the socket function fails doesn't solve it in the general case.  This solution makes std "sometimes work the same on Windows".  The detail about calling std.os.socket before any other function that can call into ws2_32 is a detail that now every application needs to know about to use std properly.  This is what I mean by adding complexity.  Yes you've made std behave the same in this one use case, but now applications have to either ensure their app is adhering to the use case you've added support for, or go back to the original interface which was to initialize it themselves.  You haven't eliminated the current use case, you've only added another one.  That's what I mean by "adding complexity".

We may be able to get Windows to behave the same if we add this fallback initialization to every component in the std library that calls a function from ws2_32, but, that's adding alot of complexity to the implementation.  This could be justified if it provides enough benefit, but let's look at the cost of the problem we are solving.  The current cost is that if an application forgets to call WSAStartup, they will get an error code indicating they forgot to call it.  Then they will add the WSAStartup call to their entry point and they will never run into that error again for their app.  I'm not sure this issue justifies the amount of complexity it would take to implement dynamic lazy initialization to every possible permutation of function calls to std.

Bxil:
I did not consider the other functions, and I agree that it's sort of a problem this way, but I am willing to bet that this would eliminate the vast majority of unexpected special behavior, as socket creation is the far most common and earliest done thing which needs WSAStartup. However, I would still rather add a kind of `ensureStartup` function to everywhere instead. I disagree that this would count as a lot of complexity. Requiring the users to keep mind of a platform specific behavior of a supposedly platform independent API though definitely raises cognitive load.

lib/std/os/windows.zig
```diff
@@ -1258,6 +1258,7 @@ pub fn WSASocketW(
             .WSAEMFILE => return error.ProcessFdQuotaExceeded,
             .WSAENOBUFS => return error.SystemResources,
             .WSAEPROTONOSUPPORT => return error.ProtocolNotSupported,
+            .WSANOTINITIALISED => return error.NotInitialised,
```
andrewrk, line 1261: Please do the retry logic here, and in any other ws2_32 function call which can return `WSANOTINITIALISED`. The documented errors for WsaStartup are:

```
WSASYSNOTREADY | The underlying network subsystem is not ready for network communication.
-- | --
WSAVERNOTSUPPORTED | The version of Windows Sockets support requested is not provided by this particular Windows Sockets implementation.
WSAEINPROGRESS | A blocking Windows Sockets 1.1 operation is in progress.
WSAEPROCLIM | A limit on the number of tasks supported by the Windows Sockets implementation has been reached.
WSAEFAULT | The lpWSAData parameter is not a valid pointer.
```

In all cases this should map to `error.SystemResources` rather than adding a new error to the `SocketError` set.

andrewrk (changes requested):
@marler8997 thanks for your thoughtful remarks here. I think you raise some valid points, however, I am making the judgement call here that Zig's std lib networking abstraction should not require a special call to WsaStartup for Windows applications to work properly.

Note that the end-game here is likely to switch to the lower level AFD API and bypass ws2_32 altogether, and we should design the abstractions, APIs, and error sets with that in mind. This WSAStartup thing is unnecessary nonsense having to do with critically flawed API design of ws2_32.dll, but it is not a property of the kernel itself. We can bypass this flaw by using the lower level APIs.

marler8997:
I mentioned this on IRC, but another problem with doing this is that `WSAStartup` is not thread safe.  So with this implementation, any applications that creates sockets on multiple threads will have a race condition that can result in a deadlock (see https://stackoverflow.com/questions/3417579/can-calling-wsastartup-from-several-threads-cause-a-deadlock)

marler8997:
Note, we can use `std.once` when calling `WSAStartup` to address the thread safty issue.

lib/std/os/windows.zig
lithdew: > @marler8997 thanks for your thoughtful remarks here. I think you raise some valid points, however, I am making the judgement call here that Zig's std lib networking abstraction should not require a special call to WsaStartup for Windows applications to work properly.
> 
> Note that the end-game here is likely to switch to the lower level AFD API and bypass ws2_32 altogether, and we should design the abstractions, APIs, and error sets with that in mind. This WSAStartup thing is unnecessary nonsense having to do with critically flawed API design of ws2_32.dll, but it is not a property of the kernel itself. We can bypass this flaw by using the lower level APIs.

Just to clarify, it is still unclear from Windows-related networking code that I've read if it is possible to bypass initializing winsock2 (WSA); even if we make use of AFD for driving socket syscalls and events.

For projects that currently use AFD (i.e. trio, mio, libuv, etc.), they only use a subset of AFD for polling read/write events for sockets, and use winsock2 for everything else. Using undocumented, _possibly_-unstable portions of AFD that other projects do not use can cause forwards compatibility problems in the case Microsoft chooses to make breaking changes.

As an example, all projects right now rely on `WSASocketW` for initializing sockets. How a socket can be initialized solely through AFD is undocumented, and I haven't seen any code online that demonstrates how it can be done without winsock2. It might even be the case that how sockets are created may rely on startup code that is entirely encapsulated in winsock2, and thus sockets can't be created with AFD alone (perhaps WSAStartup loads ws2_32.dll which then registers something to the Transport Driver Interface (TDI) to then allow sockets to be initialized? who knows...).

For the short-term for Zig's standard library, it might be a good idea to bootstrap networking for Windows with winsock2 and IOCP. In the intermediate-term, it would be good to then do research into prior work and have Zig incrementally adopt AFD for networking.

One thing to also note is that with any of the approaches, inevitably, some sort of global state/initialization is necessary (which as @marler8997 pointed out can be maintained w/ `std.once` or static initialization). This is necessary for:

1. initializing the winsock2 subsystem globally only once,
2. keeping a global pool of AFD handles for polling socket read/write events to scale socket syscalls, and
3. loading function pointers to mandatory syscall extensions on application startup (i.e. `ConnectEx`).

For a good discussion on AFD in general, here's an issue that pries through a lot of AFD's details: https://github.com/python-trio/trio/issues/52

lithdew:
> @marler8997 thanks for your thoughtful remarks here. I think you raise some valid points, however, I am making the judgement call here that Zig's std lib networking abstraction should not require a special call to WsaStartup for Windows applications to work properly.
> 
> Note that the end-game here is likely to switch to the lower level AFD API and bypass ws2_32 altogether, and we should design the abstractions, APIs, and error sets with that in mind. This WSAStartup thing is unnecessary nonsense having to do with critically flawed API design of ws2_32.dll, but it is not a property of the kernel itself. We can bypass this flaw by using the lower level APIs.

Just to clarify, it is still unclear from Windows-related networking code that I've read if it is possible to bypass initializing winsock2 (WSA); even if we make use of AFD for driving socket syscalls and events.

For projects that currently use AFD (i.e. trio, mio, libuv, etc.), they only use a subset of AFD for polling read/write events for sockets, and use winsock2 for everything else. Using undocumented, _possibly_-unstable portions of AFD that other projects do not use can cause forwards compatibility problems in the case Microsoft chooses to make breaking changes.

As an example, all projects right now rely on `WSASocketW` for initializing sockets. How a socket can be initialized solely through AFD is undocumented, and I haven't seen any code online that demonstrates how it can be done without winsock2. It might even be the case that how sockets are created may rely on startup code that is entirely encapsulated in winsock2, and thus sockets can't be created with AFD alone (perhaps WSAStartup loads ws2_32.dll which then registers something to the Transport Driver Interface (TDI) to then allow sockets to be initialized? who knows...).

For the short-term for Zig's standard library, it might be a good idea to bootstrap networking for Windows with winsock2 and IOCP. In the intermediate term, it would be good to then do research into prior work and have Zig incrementally adopt AFD for networking.

One thing to also note is that with any of the approaches, inevitably, some sort of global state/initialization is necessary (which as @marler8997 pointed out can be maintained w/ `std.once` or static initialization). This is necessary for:

1. initializing the winsock2 subsystem globally only once,
2. keeping a global pool of AFD handles for polling socket read/write events to scale socket syscalls, and
3. loading function pointers to mandatory syscall extensions on application startup (i.e. `ConnectEx`).

For a good discussion on AFD in general, here's an issue that pries through a lot of AFD's details: https://github.com/python-trio/trio/issues/52

daurnimator:
> As an example, all projects right now rely on `WSASocketW` for initializing sockets. How a socket can be initialized solely through AFD is undocumented, and I haven't seen any code online that demonstrates how it can be done without winsock2.

midipix does it; see e.g. https://git.midipix.org/ntapi/tree/src/socket/ntapi_sc_socket_v2.c

Bxil:
> WSAStartup is not thread safe

That topic doesn't seem conclusive on the actual cause of the deadlock, and .NET doesn't use any form of synchronization in calling WSAStartup. I am doubtful that a major, official Microsoft implementation would do this incorrectly. Rust's implementation uses synchronization, but Rust demands that for everything global, so I am not sure if that's relevant. Synchronizing would probably not cause any problems, so better safe than sorry: I will use `std.once`.

And while not immediately relevant to the fate of this PR, I want to add that while I am not opposed to using AFD, it seems to me that it's an awful lot of work for little to no practical gain - and has the less of losing stable, official MS documentation.

marler8997:
> .NET doesn't use any form of synchronization in calling WSAStartup

Where does Microsoft's .NET runtime call WSAStartup?  I'm not sure how you would know this without source code, unless you are a Microsoft employee?

Also keep in mind that it's fine to call WSAStartup without synchronization so long as you haven't started any threads yet, is this what .NET does?

MasterQ32:
> Where does Microsoft's .NET runtime call WSAStartup? I'm not sure how you would know this without source code, unless you are a Microsoft employee?
> 
> Also keep in mind that it's fine to call WSAStartup without synchronization so long as you haven't started any threads yet, is this what .NET does?

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketPal.Windows.cs#L41

https://github.com/dotnet/runtime/blob/453502e6b8162aa40e96e68dfd0b7ef337cf90b4/src/libraries/Common/src/Interop/Windows/WinSock/Interop.WSAStartup.cs#L15-L40

Looks like they also use initOnce() semantics

marler8997:
Interesting.  That implementation does do some synchronization, but the actual call to WSAStartup is not being synchronized so it would be susceptible to this deadlock.  Although, this open source version of Microsoft's "dotnet" is not the one commonly found on people's Windows boxes, I don't know how well it is tested.  That being said, using Zig's `std.once` would address this issue.

Bxil:
The version of .NET you are referring to has not been in development for years now, and has been deprecated a year or so ago. The implementation we are looking at is the current and official implementation, I assure you it's well tested. .NET is by all intents and purposes an open source thing now, and is a valuable resource. The old implementation you speak of is actually source available as well, here you can how WSAStartup is called in that one: https://referencesource.microsoft.com/#System/net/System/Net/Sockets/Socket.cs,5954
This one does synchronize it, albeit the code does all sorts of other stuff as well, so who knows.

Publicly exposing `wsaStartupOnce` seems like a good idea to me, I will do that.

lithdew:
> By the way, if we are going to use `std.once`, I suggest exposing 2 public functions so that an application can provide their own arguments to `WSAStartup` if they so choose.
> 
> ```zig
> // returns true if newly initialized, false if it was already initialized
> pub fn wsaStartupOnce(major: u8, minor: u8) !bool {
>     // call WSAStartup with std.once...
> }
> // returns true if newly initialized, false if it was already initialized
> pub fn wsaStartupOnceDefault() !void {
>     return try wsaStartupOnce(2, 2);
> }
> ```
> 
> So in this case, if an application needs to initialize WSA with a specific version, they can do so without `std` trying to initialize it again.
> 
> ```zig
> pub fn main() !void {
>     if (std.builtion.os.tag == .windows) {
>         std.debug.assert(try std.os.windows.wsaStartupOnce(1, 1));
>     }
> }
> ```
> 
> This way, if the application has already called WSAStartup with their own version, `std` will not try to call it again with the default version.

Is there any code that you've seen before that passes version numbers other than v2.2 to WSAStartup?

marler8997:
> @marler8997 Is there any code that you've seen before that passes version numbers other than v2.2 to WSAStartup?

I've never seen code myself that uses a version other than 2.2.

It looks like from the documentation applications can specify older versions to get older behavior.  My guess is this would be useful if you needed to call into an old DLL that needed the old behavior or something.

marler8997:
> This one does synchronize it, albeit the code does all sorts of other stuff as well, so who knows.

Yeah interesting.  They're doing good old "double-checked locking". Is this the source for the binaries that get released to Windows machines?

Bxil:
I've not seen anything ever use any other versions, but I see no harm in including such a function.
> Is this the source for the binaries that get released to Windows machines?

It's a bit complicated. This is the runtime that's installed by default on Windows 10, but new .NET applications don't use it and instead do self-contained deployment where the binary includes the runtime. So it depends on how old the program is - both of them are actively used in production code.

PR #8698, opened 2021-05-06
std/mem: add sliceTo(), deprecate spanZ(), lenZ()

ifreund:
The current spanZ() function will not scan for a 0 terminator if the
type is not 0 terminated. This encourages using 0 terminated array
types to bind C arrays which hold 0 terminated strings. However, this is
a big footgun as nothing in the C type system guarantees there to be a
0 terminator at the end of the array and if there is none this becomes
Illegal Behavior in Zig.

To solve this, deprecate spanZ() and lenZ(), adding a new sliceTo()
function that always scans for the given terminator even if the type is
not sentinel terminated.

cc @SpexGuy

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -602,6 +602,7 @@ test "span" {
     try testing.expectEqual(@as(?[:0]u16, null), span(@as(?[*:0]u16, null)));
 }
 
+/// Deprecated: use std.mem.span() or std.mem.sliceTo()
 /// Same as `span`, except when there is both a sentinel and an array
 /// length or slice length, scans the memory for the sentinel value
 /// rather than using the length.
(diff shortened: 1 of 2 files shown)


lib/std/mem.zig
```diff
@@ -630,6 +631,140 @@ test "spanZ" {
...
+    if (@typeInfo(@TypeOf(ptr)) == .Optional) {
+        if (ptr) |non_null| {
+            return spanZ(non_null);
+        } else {
+            return null;
+        }
+    }
+    const Result = Span(@TypeOf(ptr));
+    const l = scanLenZ(ptr);
+    if (@typeInfo(Result).Pointer.sentinel) |s| {
```
SpexGuy, line 648: `scanLenZ` handles `.Array` but this only handles `.Pointer`.  Since `scanLenZ` is only used here, they should be made to handle the same set of types.  It doesn't seem like handling of `.Array` is needed.

lib/std/mem.zig
```diff
@@ -630,6 +631,140 @@ test "spanZ" {
...
+            .Slice => if (info.sentinel) |sentinel| {
+                if (sentinel != 0) {
+                    @compileError("std.mem.scanZ requires the sentinel, if any, to be 0");
+                }
+                return indexOfSentinel(info.child, sentinel, ptr);
+            } else {
+                return indexOfScalar(info.child, ptr, 0) orelse ptr.len;
+            },
+        },
+        else => @compileError("invalid type given to std.mem.scanZ"),
```
SpexGuy, line 733: You should add ` ++ @typeName(@TypeOf(ptr))` here to make debugging easier.

lib/std/mem.zig
```diff
@@ -630,6 +631,140 @@ test "spanZ" {
...
+                    if (x.sentinel) |sentinel| {
+                        if (sentinel != 0) {
+                            @compileError("std.mem.scanZ requires the sentinel, if any, to be 0");
+                        }
+                        return indexOfSentinel(x.child, sentinel, ptr);
+                    } else {
+                        return indexOfScalar(x.child, ptr, 0) orelse x.len;
+                    }
+                },
+                else => @compileError("invalid type given to std.mem.scanZ"),
```
SpexGuy, line 708: Since each handled case does `return`, you could simplify this by using `else => {}` here and having this compile error only once after the outer `switch` body.

lib/std/mem.zig
```diff
@@ -630,6 +631,140 @@ test "spanZ" {
...
+            },
+            .Many => {
+                if (info.sentinel) |sentinel| {
+                    if (sentinel != 0) {
+                        @compileError("std.mem.scanZ requires the sentinel, if any, to be 0");
```
SpexGuy, line 713: This could be simplified by checking only once in `scanZ` when you get the sentinel there, and assuming the sentinel is zero here.

lib/std/mem.zig
```diff
@@ -630,6 +631,140 @@ test "spanZ" {
...
 }
 
+/// Takes a pointer to an array, an array, a sentinel-terminated pointer,
+/// or a slice and scans for the first 0, returning the scanned slice.
+/// If the type is an array or slice that is not sentinel terminated
+/// and no sentinel is found, the full length of the array/slice is returned.
+pub fn scanZ(ptr: anytype) Span(@TypeOf(ptr)) {
+    if (@typeInfo(@TypeOf(ptr)) == .Optional) {
+        if (ptr) |non_null| {
+            return spanZ(non_null);
```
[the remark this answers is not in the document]
ifreund: Thanks, this made me realize that I forgot tests for optional pointers

ifreund:
Thanks for the review @SpexGuy! I've implemented your great suggestions :)

lib/std/mem.zig
```diff
@@ -602,32 +602,120 @@ test "span" {
     testing.expectEqual(@as(?[:0]u16, null), span(@as(?[*:0]u16, null)));
 }
 
-/// Same as `span`, except when there is both a sentinel and an array
-/// length or slice length, scans the memory for the sentinel value
-/// rather than using the length.
-pub fn spanZ(ptr: anytype) Span(@TypeOf(ptr)) {
```
SpexGuy, line 608: Did you intend to delete spanZ or just deprecate it?

ifreund: Thanks, I only intended to deprecate it, I got this right in the first round but appeared to have screwed it up in the revised version due to testing in a separate zig file to avoid running the entire stdlib test suite.

lib/std/mem.zig
```diff
@@ -630,6 +631,122 @@ test "spanZ" {
...
+            .Many => if (info.sentinel) |sentinel| {
+                assert(sentinel == 0);
+                return indexOfSentinel(info.child, sentinel, ptr);
+            },
+            .C => {
+                assert(ptr != null);
+                return indexOfSentinel(info.child, 0, ptr);
+            },
+            .Slice => if (info.sentinel) |sentinel| {
+                assert(sentinel == 0);
```
daurnimator, line 712: why? there's no reason to ban other sentinels.

daurnimator (commented):
`scan` brings to mind `scanf` and otherwise reading user input.
Could you instead just change `spanZ`?

lib/std/mem.zig
```diff
@@ -688,6 +805,7 @@ test "len" {
     }
 }
 
+/// Deprecated: use std.mem.len() or std.mem.scanZ().len
```
daurnimator, line 808: If you ban sentinels other than `0` then we shouldn't deprecate this.

lib/std/mem.zig
SpexGuy: When no sentinel is present in the type, this function assumes that the sentinel is zero.  So it would be a bit weird to both assume zero and allow nonzero.

lib/std/mem.zig
ifreund: I'd argue that `spanZ` is the wrong name for a function that works with sentinels other than 0. I also don't think there is high demand for working with any other sentinel in practice. If there is in fact demand, my preferred solution would be adding a `scanSentinel()` function that takes the sentinel type and value as arguments.

ifreund:
> `scan` brings to mind `scanf` and otherwise reading user input.
> Could you instead just change `spanZ`?

I don't think "just change `spanZ`" is an option here as the whole point of this PR is to scan for a 0 even if the type is not sentinel terminated which wouldn't work with the current `spanZ`'s support for arbitrary sentinel types (well except for C pointers where 0 is the default). Furthermore, I think deprecating a function and replacing it with a new one is better than subtly changing the behavior of an existing function for some arguments without any compile error.

I also think `scanZ` is a better name for a function that always, regardless of input, iterates over the memory scanning for a sentinel.

andrewrk (approved):

ifreund:
rebased for `std.testing` changes

lib/std/mem.zig
daurnimator: > I also don't think there is high demand for working with any other sentinel in practice

to the contrary, I work with other sentinel values regularly. Most commonly `null` to indicate the end of a list of pointers passed from C; but also other fields, e.g. when working with [`luaL_Reg`](https://www.lua.org/manual/5.4/manual.html#luaL_Reg) the other day:
> Any array of luaL_Reg must end with a sentinel entry in which both name and func are NULL.

daurnimator: I think it would be worse than the existing `lenZ`/`spanZ`: there's no need to pass in the sentinel as it's already part of the type of the pointer.

ifreund: @daurnimator if the sentinel is part of the pointer, then `len()`/`span()` work fine.

ifreund:
After [discussion on IRC](https://freenode.irclog.whitequark.org/zig/2021-05-11#1620726596-1620732448;) with @daurnimator, I decided to make the new function take the sentinel value as a comptime argument. This is much more flexible than the originally proposed `scanZ` function without becoming cumbersome.

I'm not 100% confident in the name, `scanTo` or `spanTo()` are also decent options, but I think `sliceTo()` is clear.

@SpexGuy, @andrewrk you may want to take a fresh look.

SpexGuy (approved):

lib/std/mem.zig
```diff
@@ -630,6 +631,190 @@ test "spanZ" {
...
+                        // the sentinel of the type passed.
+                        if (array_info.sentinel) |sentinel| {
+                            if (end == sentinel) {
+                                new_ptr_info.sentinel = end;
+                            } else {
+                                new_ptr_info.sentinel = null;
+                            }
+                        }
+                    },
+                    else => {},
```
daurnimator, line 658: should this be an compile error/unreachable?

lib/std/mem.zig
```diff
@@ -630,6 +631,190 @@ test "spanZ" {
...
+                        if (end == sentinel) {
+                            new_ptr_info.sentinel = end;
+                        } else {
+                            new_ptr_info.sentinel = null;
+                        }
+                    }
+                },
+                .C => {
+                    new_ptr_info.sentinel = end;
+                    new_ptr_info.is_allowzero = false;
```
daurnimator, line 674: why do we need this?

lib/std/mem.zig
```diff
@@ -630,6 +631,190 @@ test "spanZ" {
...
+                },
+            }
+            return @Type(std.builtin.TypeInfo{ .Pointer = new_ptr_info });
+        },
+        else => {},
+    }
+    @compileError("invalid type given to std.mem.sliceTo");
```
daurnimator, line 681: perhaps add `@typeName` to this error message?

lib/std/mem.zig
```diff
@@ -630,6 +631,190 @@ test "spanZ" {
...
+
+/// Takes a pointer to an array, an array, a sentinel-terminated pointer, or a slice and
+/// iterates searching for the first occurrence of `end`, returning the scanned slice.
+/// If `end` is not found, the full length of the array/slice/sentinel terminated pointer is returned.
+/// If the pointer type is sentinel terminated and `end` matches that terminator, the
+/// resulting slice is also sentinel terminated.
+/// Pointer properties such as mutability and alignment are preserved.
+/// C pointers are assumed to be non-null.
+pub fn sliceTo(ptr: anytype, comptime end: meta.Elem(@TypeOf(ptr))) SliceTo(@TypeOf(ptr), end) {
+    if (@typeInfo(@TypeOf(ptr)) == .Optional) {
```
daurnimator, line 692: do we also need a test in here for null C pointers and return null before calling `lenSliceTo`?

lib/std/mem.zig
ifreund: It is, see the `@compileError()` at the bottom of the function.

lib/std/mem.zig
ifreund: C pointers are implicitly `allowzero`, but we don't want the return type to be.

lib/std/mem.zig
ifreund: Good call, done.

lib/std/mem.zig
ifreund: In the comment it is stated that C pointers are assumed to be non-null, which is consistent with how `mem.span()` and the now deprecated `mem.spanZ()` work. I think this behavior is fine as-is.

lib/std/mem.zig
ifreund: I added an assert that `is_allowzero` is true before setting to false and a comment. `git blame` says @andrewrk originally added this line in ef3d761da545a3a72928ed0e0ba3b749a4cb74d8.

lib/std/mem.zig
daurnimator: it's odd IMO that normal optional pointers work, but C ones don't.

lib/std/mem.zig
daurnimator: no it doesn't? won't this hit the `return @Type(std.builtin.TypeInfo{ .Pointer = new_ptr_info });`?

ifreund: Ah yes you're right but this is technically fine as `sliceToLen()` will give a compile error. We rely on this for non sentinel terminated "many" pointers as well currently.

lib/std/mem.zig
ifreund: If c pointers worked like normal pointers with regards to null I would agree (i.e. if a nullable c pointer was `?[*c]`). As it stands, I think making this function work like all of the libc `strfoo()` functions and not allow null makes sense.

daurnimator: Ah hmm.
In that case, I think `sliceTo` should return an optional slice when passed a C pointer.

PR #8699, opened 2021-05-06
fix test to restore cwd after chdir

mikdusan:
- `../zig-cache/tmp` is no longer created by subsequent test(s)


kubkon:
Wow, that's a great catch and some serious oversight on my part. Thanks for fixing it!

kubkon (approved):

PR #8701, opened 2021-05-07
Fix a double free and leak in json.zig

mattbork:
If  `parse()` is used with the option `duplicate_field_behavior` set to `UseLast`, a double free occurs when a field requires allocation and an error occurs when parsing a second or later occurrence. This is because any later occurrences of the field will free the old value before trying to parse a new value, but crucially will not reset `fields_seen[i]` back to `false`. So if there is an error when parsing the new value, the earlier `errdefer` block will try to free the old value again. This can be fixed by simply clearing `fields_seen[i]`.

For example, if trying to parse into a `struct` with a `*u64` field `a` and `UseLast` set, the following JSON will cause a double free currently because an error occurs when trying to parse `0.25` as a u64:

`{
    "a": 1,
    "a": 0.25
}`

There is also a memory leak in the `.Pointer` case of `parseInternal` when parsing to a pointer to a single item. An object of type `ptrInfo.child` is first allocated, then a recursive call to `parseInternal` is tried. If that returns an error, then the object is never freed. This can be fixed by adding an `errdefer` statement before the parsing. I included this change in this pull request because otherwise the test I wrote will leak memory.

I added a test for parsing into a struct with a duplicate field. Because the testing allocator (a `GeneralPurposeAllocator`) can only detect double frees if the bucket it's allocating from is kept alive, I had to add a slightly odd ballast allocation.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1571,6 +1571,7 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options: 
                                         return error.DuplicateJSONField;
                                     } else if (options.duplicate_field_behavior == .UseLast) {
                                         parseFree(field.field_type, @field(r, field.name), options);
+                                        fields_seen[i] = false;
                                     }
                                 }
                                 if (field.is_comptime) {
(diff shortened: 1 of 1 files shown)


ifreund (approved):
LGTM, nice catch!

PR #8709, opened 2021-05-08
stage2 ARM: Generate correct code when spilling registers

joachimschmidt557:
Contrary to what I previously assumed, adding functions for allocating multiple registers at once was not enough to solve this problem. A new argument `exceptions` has been added to functions in `RegisterManager`. All registers in the `exceptions` slice will be disregarded when allocating registers.

A cleanup of the `RegisterManager` API and unit tests for `exceptions` will be introduced in a subsequent PR.


kubkon (approved):
LGTM!

PR #8711, opened 2021-05-08
std/os, x/os/socket: windows support, socket helpers, getpeername()

lithdew:
Socket I/O methods such as read, readv, write, writev, send, recv, sendmsg, recvmsg have been generalized to read(buf, flags), write(buf, flags), readVectorized(vectors, flags), and writeVectorized(vectors, flags). There is still some work left to be done abstracting both readVectorized and writeVectorized properly across platforms, which is work to be done in a future PR.

Support for setting the linger timeout of a socket, querying the remote address of a socket, setting whether or not keep-alive messages are to be sent through a connection-oriented socket periodically depending on host operating system settings has been added.

`std.io.Reader` and `std.io.Writer` wrappers around `Socket` has been implemented, which wrap around Socket.read(buf, flags) and Socket.write(buf, flags). Both wrappers may be provided flags which are passed to Socket.read / Socket.write accordingly.

Cross-platform support for `getpeername()` has been implemented.

Windows support for the new `std.x.os.Socket` has been implemented. To accomplish this, a full refactor of `std.os.windows.ws2_32` has been done to supply any missing definitions and constants based on auto-generated Windows syscall bindings by @marler8997.

`std.x.net.TCP.Listener.setQuickACK` has been moved to `std.x.net.TCP.Client.setQuickACK`.

Windows support for resolving the scope ID of an interface name specified in an IPv6 address has been provided.

`sockaddr_storage` definitions have been provided for Windows, Linux, and Darwin. `sockaddr_storage` is used to allocate space before any socket addresses are queried via. calls such as accept(), getsockname(), and getpeername().

Zig-friendly wrappers for GetQueuedCompletionStatusEx(), getpeername(), SetConsoleCtrlHandler(), SetFileCompletionNotificationModes() syscalls on Windows have been provided.

Socket.setOption() was provided to set the value of a socket option in place of os.setsockopt. Socket.getOption() will be provided in a future PR.

There is still further work to be done regarding querying socket option values on Windows, which is to be done in a subsequent PR.


lithdew:
Made changes to move Socket.{Reader, Writer} to tcp.Client.{Reader, Writer} to avoid a footgun where std.io.Writer / std.io.Reader interfaces may be used with sockets that are not connected-oriented, as suggested by @MasterQ32.

lib/std/x/os/socket_posix.zig
```diff
@@ -0,0 +1,235 @@
...
+        return os.setsockopt(self.fd, level, name, value);
+    }
+
+    /// Have close() or shutdown() syscalls block until all queued messages in the socket have been successfully
+    /// sent, or if the timeout specified in seconds has been reached. It returns `error.UnsupportedSocketOption`
+    /// if the host does not support the option for a socket to linger around up until a timeout specified in
+    /// seconds.
+    pub fn setLinger(self: Socket, timeout_seconds: ?u16) !void {
+        if (comptime @hasDecl(os, "SO_LINGER")) {
+            const settings = struct {
```
LemonBoy, line 159:
```suggestion
            const settings = extern struct {
```

lib/std/os/windows/ws2_32.zig
```diff
@@ -7,10 +7,879 @@ usingnamespace @import("bits.zig");
 
 pub const SOCKET = *opaque {};
 pub const INVALID_SOCKET = @intToPtr(SOCKET, ~@as(usize, 0));
-pub const SOCKET_ERROR = -1;
```
komuw, line 10: maybe this file should have a comment pointing out that some of the constants are from https://github.com/marlersoft/zigwin32 ?

This would help anyone modifying this file in the future

lib/std/x/os/socket.zig
```diff
@@ -0,0 +1,117 @@
...
+const builtin = std.builtin;
+
+/// Import in a `Socket` abstraction depending on the platform we are compiling against.
+pub usingnamespace switch (builtin.os.tag) {
+    .windows => @import("socket_windows.zig"),
+    else => @import("socket_posix.zig"),
+};
+
+/// A common subset of shared structs across cross-platform abstractions over socket syscalls.
+pub fn Mixin(comptime Self: type) type {
```
komuw, line 23: pedantic: maybe call it `ConnAddrMixin` ? especially since it is pub.
To me, calling it mixin is similar to calling a variable `object`.

lib/std/os/windows/ws2_32.zig
```diff
@@ -7,10 +7,879 @@ usingnamespace @import("bits.zig");
...
-pub const WSADESCRIPTION_LEN = 256;
-pub const WSASYS_STATUS_LEN = 128;
+pub const GROUP = u32;
+pub const ADDRESS_FAMILY = u16;
+pub const WSAEVENT = HANDLE;
+
+// Microsoft use the signed c_int for this, but it should never be negative
+pub const socklen_t = u32;
+
+pub const LM_HB_Extension = @as(i32, 128);
```
daurnimator, line 18:
```suggestion
pub const LM_HB_Extension:i32 = 128;
```

or alternatively: leave them as `comptime_int`.

lib/std/os/windows/ws2_32.zig
```diff
@@ -7,10 +7,879 @@ usingnamespace @import("bits.zig");
...
+pub const IN_CLASSC_NSHIFT = @as(u32, 8);
+pub const IN_CLASSC_HOST = @as(u32, 255);
+pub const IN_CLASSD_NET = @as(u32, 4026531840);
+pub const IN_CLASSD_NSHIFT = @as(u32, 28);
+pub const IN_CLASSD_HOST = @as(u32, 268435455);
+pub const INADDR_LOOPBACK = @as(u32, 2130706433);
+pub const INADDR_NONE = @as(u32, 4294967295);
+pub const IOCPARM_MASK = @as(u32, 127);
+pub const IOC_VOID = @as(u32, 536870912);
+pub const IOC_OUT = @as(u32, 1073741824);
```
daurnimator, line 553: Many of these are much more readable in hex form.
```suggestion
pub const IOC_OUT:u32 = 0x40000000;
```

lib/std/os/windows/ws2_32.zig
```diff
@@ -704,180 +1607,649 @@ pub const WinsockError = extern enum(u16) {
...
 
-/// The IOCTL is a generic Windows Sockets 2 IOCTL code. New IOCTL codes defined for Windows Sockets 2 will have T == 1.
-const IOC_WS2 = 0x08000000;
+pub extern "ws2_32" fn getpeername(
+    s: SOCKET,
+    name: *sockaddr,
+    namelen: *i32,
+) callconv(WINAPI) i32;
 
-pub const SIO_BASE_HANDLE = IOC_OUT | IOC_WS2 | 34;
```
daurnimator, line 719: why remove this?

lib/std/x/net/tcp.zig
```diff
@@ -89,41 +112,46 @@ pub const Client = struct {
         return self.socket.connect(address.into());
     }
 
-    /// Read data from the socket into the buffer provided. It returns the
-    /// number of bytes read into the buffer provided.
-    pub fn read(self: Client, buf: []u8) !usize {
-        return self.socket.read(buf);
+    /// Extracts the error set of a function.
+    /// TODO: remove after Socket.{read, write} error unions are well-defined across different platforms
+    fn ErrorSetOf(comptime Function: anytype) type {
```
daurnimator, line 117: this looks like something that belongs in `std.meta`

lib/std/x/net/tcp.zig
```diff
@@ -89,41 +112,46 @@ pub const Client = struct {
...
-    pub fn sendmsg(self: Client, msg: os.msghdr_const, flags: u32) !usize {
-        return self.socket.sendmsg(msg, flags);
+    pub fn writeVectorized(self: Client, msg: os.msghdr_const, flags: u32) !usize {
+        return self.socket.writeVectorized(msg, flags);
+    }
+
+    /// Read multiple I/O vectors with a prepended message header from the socket
+    /// with a set of flags specified. It returns the number of bytes that were
+    /// read into the buffer provided.
+    pub fn readVectorized(self: cLIENT, msg: *os.msghdr, flags: u32) !usize {
```
daurnimator, line 153: `cLIENT`?

lib/std/x/os/socket_posix.zig
```diff
@@ -0,0 +1,235 @@
...
+        try os.getsockname(self.fd, @ptrCast(*os.sockaddr, &address), &address_len);
+        return Socket.Address.fromNative(@alignCast(4, @ptrCast(*os.sockaddr, &address)));
+    }
+
+    /// Query the address that the socket is connected to.
+    pub fn getRemoteAddress(self: Socket) !Socket.Address {
+        var address: os.sockaddr_storage = undefined;
+        var address_len: u32 = @sizeOf(os.sockaddr_storage);
+        try os.getpeername(self.fd, @ptrCast(*os.sockaddr, &address), &address_len);
+        return Socket.Address.fromNative(@alignCast(4, @ptrCast(*os.sockaddr, &address)));
```
daurnimator, line 106: Can you not do the align in the ptrCast?
`@ptrCast(*align(4) os.sockaddr, ...)`

lib/std/x/os/socket_posix.zig
```diff
@@ -0,0 +1,235 @@
...
+
+    /// Set the read buffer size of the socket.
+    pub fn setReadBufferSize(self: Socket, size: u32) !void {
+        return self.setOption(os.SOL_SOCKET, os.SO_RCVBUF, mem.asBytes(&size));
+    }
+
+    /// Set a timeout on the socket that is to occur if no messages are successfully written
+    /// to its bound destination after a specified number of milliseconds. A subsequent write
+    /// to the socket will thereafter return `error.WouldBlock` should the timeout be exceeded.
+    pub fn setWriteTimeout(self: Socket, milliseconds: usize) !void {
```
daurnimator, line 214: I would put a big warning on these functions: they are incompatible with non-blocking sockets.
I'd almost prefer we didn't even provide them as they're a footgun for most users.

lib/std/x/os/socket_windows.zig
```diff
@@ -0,0 +1,392 @@
...
+        }
+
+        return @intCast(usize, rc);
+    }
+
+    /// Writes multiple I/O vectors with a prepended message header to the socket
+    /// with a set of flags specified. It returns the number of bytes that are
+    /// written to the socket.
+    pub fn writeVectorized(self: Socket, msg: os.msghdr_const, flags: u32) !usize {
+        return error.NotImplemented;
```
daurnimator, line 251: Just don't include the function for now?

lib/std/x/os/socket_windows.zig
```diff
@@ -0,0 +1,392 @@
...
+        return 0;
+    }
+
+    /// Query the write buffer size of the socket.
+    pub fn getWriteBufferSize(self: Socket) !u32 {
+        return 0;
+    }
+
+    /// Set a socket option.
+    pub fn setOption(self: Socket, level: u32, name: u32, value: []const u8) !void {
```
daurnimator, line 317: `name` doesn't seem like a good name for this variable?

daurnimator (changes requested):

lib/std/x/net/tcp.zig
lithdew: Got a few typos I need to fix 😅 will do some cleanup early tomorrow and apply your suggestions (away from the office right now).

lib/std/os/windows/ws2_32.zig
lithdew: Agreed, though this is output directly from the auto-generated bindings from @marler8997 - unsure if the constants are intentionally in decimal form due to the generator, or if the constants provided by Microsoft are in hex form.

lib/std/os/windows/ws2_32.zig
lithdew: I'll get that back in 😄.

lib/std/x/net/tcp.zig
lithdew: It's just a temporary helper method for now, though would you propose that we have a method like this in `std.meta` in general?

lib/std/x/os/socket_posix.zig
lithdew: I'll double check tomorrow,

lib/std/x/os/socket_windows.zig
lithdew: How about `option: u32`?

lib/std/os/windows/ws2_32.zig
lithdew: Sure, though the comment may better belong in an another file - if it made sense, I would've preferred having the entirety of the bindings in the standard library, but it unfortunately is ~1.6MiB of code 😅.

As constant and syscall definitions will eventually be pulled from the auto-generated bindings from other modules liker kernel32, maybe a good place to put the comment might just be in `std.os.windows`?

lib/std/x/os/socket.zig
lithdew: Hmm, I'll see about addressing this in a subsequent PR - maybe it might be better to just not forego mixins and define them as separate structs.

lib/std/x/os/socket_posix.zig
lithdew: Found out that I can directly @ptrCast to *os.sockaddr now that os.sockaddr_storage is used. Fix will appear in the next commit.

lib/std/x/os/socket_windows.zig
lithdew: Changed it to `code: u32`.

lib/std/x/os/socket_windows.zig
lithdew: Looking to have it just be a stub for now, I want to address it in the next PR :smile:.

lib/std/os/windows/ws2_32.zig
lithdew: cc @marler8997: would love to know if this is something that could be corrected by the bindings generator

lib/std/x/os/socket.zig
lithdew: Had to do a rebase just now, cleaned up how methods/structs are mixed in within the new commit.

lib/std/x/os/socket_posix.zig
lithdew: Warning added.

lib/std/x/os/socket_posix.zig
lithdew: Fix landed.

lib/std/x/net/tcp.zig
lithdew: Fixed.

lib/std/x/os/socket_windows.zig
lithdew: Implemented `{read, write}Vectorized()` for both Windows and POSIX in a new commit. More work could be done to make the abstraction much better, though that will be work for subsequent PR's.

Issue #8625, opened 2021-04-26
Proposal: have testing.expectEqual use the `eql` method if defined


LemonBoy:
It is common wisdom to define an `eql` predicate whenever the equality between two objects needs some extra logic to be determined.
Let's make `testing.expectEqual` use `eql` to compare objects for equality if available.

Opening a ticket to gauge interest.

**Downsides**

The `eql` name is officially blessed as `format` already is (perhaps we want to rename it into `equal` ?).

**Implementation**

```diff
diff --git a/lib/std/testing.zig b/lib/std/testing.zig
index ac1aa3bf6..9ea02dc01 100644
--- a/lib/std/testing.zig
+++ b/lib/std/testing.zig
@@ -45,6 +45,13 @@ pub fn expectError(expected_error: anyerror, actual_error_union: anytype) void {
 /// then aborts.
 /// `actual` is casted to the type of `expected`.
 pub fn expectEqual(expected: anytype, actual: @TypeOf(expected)) void {
+    if (comptime std.meta.trait.hasFn("eql")(@TypeOf(expected))) {
+        if (!expected.eql(actual)) {
+            std.debug.panic("expected {}, found {}", .{ expected, actual });
+        }
+        return;
+    }
+
     switch (@typeInfo(@TypeOf(actual))) {
         .NoReturn,
         .BoundFn,
```

jayschwa:
> The `eql` name is officially blessed as `format` already is (perhaps we want to rename it into `equal` ?).

What does "officially blessed" mean? I think the overall proposal is a good idea, plus a rename too. "eq", "eql", and "equal" can all be found in the standard library. Personally, I think `if (foo.equals(bar))` reads the best for a method name.

LemonBoy:
> What does "officially blessed" mean?

That we reserve the name `eql` (or `equal` or whatever) for equality predicates and we expect the prototype to be `fn (T,T) bool` with several properties (symmetric and transitive at minimum).

tauoverpi:
Would it be worth to have this for `meta.eql` too? Same could apply to a `meta.lessThan` and similar.

jedisct1:
> Would it be worth to have this for `meta.eql` too? Same could 

> apply to a `meta.lessThan` and similar.

Having this for `meta.eql` could indeed be very useful.

andrewrk:
I veto this proposal. Reasoning:

The purpose of `expectEqual` is to print diagnostics about why two things are not equal. Here are the doc comments:

[...]

PR #8713, opened 2021-05-08
std: testing.expectEqual uses the `eql` method if available

LemonBoy:
Implements the proposal outlined in #8625


andrewrk:
https://github.com/ziglang/zig/issues/8625#issuecomment-888629260

PR #8717, opened 2021-05-08
Dwarf on windows

mchudleigh:
This PR does 3 things but can be split up if needed.

- Fix a compile time crash/unreachable-code when using bash based terminals on windows
- Emit Dwarf debug info if targetting gnu abi on windows (maybe this should be a flag instead)
- Add support for reading Dwarf info from COFF executables for stack and error tracing

### DIFF
--- a/lib/std/Progress.zig
+++ b/lib/std/Progress.zig
@@ -22,6 +22,10 @@ const Progress = @This();
 /// not print on update()
 terminal: ?std.fs.File = undefined,
 
+/// Is this a windows API terminal (note: this is not the same as being run on windows
+/// because other terminals exist like MSYS/git-bash)
+is_windows_terminal: bool = false,
+
 /// Whether the terminal supports ANSI escape codes.
 supports_ansi_escape_codes: bool = false,
 
@@ -143,6 +147,7 @@ pub fn start(self: *Progress, name: []const u8, estimated_total_items: usize) !*
         self.terminal = stderr;
         self.supports_ansi_escape_codes = true;
     } else if (std.builtin.os.tag == .windows and stderr.isTty()) {
+        self.is_windows_terminal = true;
         self.terminal = stderr;
     } else if (std.builtin.os.tag != .windows) {
         // we are in a "dumb" terminal like in acme or writing to a file
@@ -192,8 +197,9 @@ const DECRC = "\x1b8";
 // supported by some terminals (eg. Terminal.app).
 
 fn refreshWithHeldLock(self: *Progress) void {
-    const is_dumb = !self.supports_ansi_escape_codes and !(std.builtin.os.tag == .windows);
+    const is_dumb = !self.supports_ansi_escape_codes and !self.is_windows_terminal;
     if (is_dumb and self.dont_print_on_dumb) return;
+
     const file = self.terminal orelse return;
 
     var end: usize = 0;
@@ -206,6 +212,8 @@ fn refreshWithHeldLock(self: *Progress) void {
         std.mem.copy(u8, self.output_buffer[end..], seq_before);
         end += seq_before.len;
     } else if (std.builtin.os.tag == .windows) winapi: {
+        std.debug.assert(self.is_windows_terminal);
+
         var info: windows.CONSOLE_SCREEN_BUFFER_INFO = undefined;
         if (windows.kernel32.GetConsoleScreenBufferInfo(file.handle, &info) != windows.TRUE)
             unreachable;
@@ -282,7 +290,7 @@ fn refreshWithHeldLock(self: *Progress) void {
         const seq_after = DECRC;
         std.mem.copy(u8, self.output_buffer[end..], seq_after);
         end += seq_after.len;
-    } else if (std.builtin.os.tag != .windows) {
+    } else if (!self.is_windows_terminal) {
         self.output_buffer[end] = '\n';
         end += 1;
     }
@@ -293,8 +301,10 @@ fn refreshWithHeldLock(self: *Progress) void {
     };
 
     if (std.builtin.os.tag == .windows) {
-        if (windows.kernel32.SetConsoleCursorPosition(file.handle, saved_cursor_pos) != windows.TRUE)
-            unreachable;
+        if (self.is_windows_terminal) {
+            const res = windows.kernel32.SetConsoleCursorPosition(file.handle, saved_cursor_pos);
+            std.debug.assert(res == windows.TRUE);
+        }
     }
 
     self.prev_refresh_timestamp = self.timer.read();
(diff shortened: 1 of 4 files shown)


mchudleigh:
Also it has been compiled and tested on Windows and Linux

lib/std/Progress.zig
```diff
@@ -143,6 +147,7 @@ pub fn start(self: *Progress, name: []const u8, estimated_total_items: usize) !*
         self.terminal = stderr;
         self.supports_ansi_escape_codes = true;
     } else if (std.builtin.os.tag == .windows and stderr.isTty()) {
+        self.is_windows_terminal = true;
```
LemonBoy, line 150: How do you get here? `isCygwinPty` and `supportsAnsiEscapeCodes` should detect this case and use escape sequences instead of WinAPI functions.

lib/std/coff.zig
```diff
@@ -162,20 +163,51 @@ pub const Coff = struct {
         try self.loadOptionalHeader();
     }
 
+    fn readStringFromTable(self: *Coff, offset: u32, buff: []u8) !u32 {
```
LemonBoy, line 166:
```suggestion
    fn readStringFromTable(self: *Coff, offset: usize, buf: []u8) ![]const u8 {
```

lib/std/coff.zig
```diff
@@ -162,20 +163,51 @@ pub const Coff = struct {
...
+        if (self.coff_header.pointer_to_symbol_table == 0) {
+            // No symbol table therefore no string table
+            return error.MissingStringTable;
+        }
+        const string_table_offset = self.coff_header.pointer_to_symbol_table + (self.coff_header.number_of_symbols * 18) + offset;
```
LemonBoy, line 171: Please add a comment explaining what `18` is.

lib/std/coff.zig
```diff
@@ -162,20 +163,51 @@ pub const Coff = struct {
...
+        const in = self.in_file.reader();
+        const old_pos = try self.in_file.getPos();
+
+        try self.in_file.seekTo(string_table_offset);
+
+        var i: u8 = 0;
+        while (i < buff.len) : (i+=1) {
```
LemonBoy, line 178: This loop can be simplified down to:
```zig
const str = try in.readUntilDelimiterOrEof(buf, 0)
```

lib/std/coff.zig
```diff
@@ -162,20 +163,51 @@ pub const Coff = struct {
...
+            const byte = try in.readByte();
+            if (byte == 0) {
+                break;
+            }
+            buff[i] = byte;
+        }
+        try self.in_file.seekTo(old_pos);
```
LemonBoy, line 185: I'd place this in a `defer` block below the other `seekTo`, this way you're also rewinding the stream in case of error.

lib/std/coff.zig
```diff
@@ -162,20 +163,51 @@ pub const Coff = struct {
...
-        // 64 bits by the way.
-        var skip_size: u16 = undefined;
+        // All we care about is the image base value and PDB info
+        // The header structure is different for 32 or 64 bit
+        var num_rva_pos: u64 = undefined;
         if (self.pe_header.magic == IMAGE_NT_OPTIONAL_HDR32_MAGIC) {
-            skip_size = 2 * @sizeOf(u8) + 8 * @sizeOf(u16) + 18 * @sizeOf(u32);
+            num_rva_pos = opt_header_pos + 92;
+
+            try self.in_file.seekTo(opt_header_pos+28);
```
LemonBoy, line 200: Please format all the files you're touching with `zig fmt`.

lib/std/coff.zig
```diff
@@ -277,6 +309,19 @@ pub const Coff = struct {
                     .characteristics = try in.readIntLittle(u32),
                 },
             });
+            const sec = &self.sections.items[self.sections.items.len-1];
```
LemonBoy, line 312: Why not move the name expansion logic before the `append` call?

lib/std/coff.zig
```diff
@@ -288,6 +333,22 @@ pub const Coff = struct {
...
+            if (mem.eql(u8, sec.header.name[0..name.len], name)) {
+                break sec;
+            }
+        } else {
+            return error.MissingCoffSection;
+        };
+        const in = self.in_file.reader();
+        try self.in_file.seekTo(sec.header.pointer_to_raw_data);
+        const out_buff = try allocator.alloc(u8, sec.header.misc.virtual_size);
+        const readBytes = in.read(out_buff);
```
LemonBoy, line 349:
```suggestion
        in.readNoEof(out_buff);
```

lib/std/debug.zig
```diff
@@ -1525,6 +1548,7 @@ pub const ModuleDebugInfo = switch (builtin.os.tag) {
     .uefi, .windows => struct {
         base_address: usize,
         pdb: pdb.Pdb,
```
LemonBoy, line 1550: The `pdb`/`dwarf` can be consolidated into a `union (enum) { pdb: pdb.Pdb, dwarf: DW.Dwarfinfo }`.

LemonBoy (commented):
Great work! I've left a few trivial comments here and there (and please run `zig fmt`).

lib/std/Progress.zig
mchudleigh: Yes, that is the case. I was just trying to make it more explicit. The first case of the if/else if chain there is:

`if (stderr.supportsAnsiEscapeCodes()` which is just `isCygwinPty()` on windows.

I didn't add a comment but `supports_ansi_escapse_codes` and `is_windows_terminal` are mutually exclusive.

lib/std/coff.zig
mchudleigh: Good catch

lib/std/coff.zig
mchudleigh: Will do

lib/std/coff.zig
mchudleigh: Good to know. I'm still getting used to the stdlib

lib/std/debug.zig
mchudleigh: Good idea

mchudleigh:
Thanks for the quick review and constructive comments.

I'll fix everything you recommended and I'll include one more commit if it's alright:
Get rid of the ugly `@intCast` in Debug.zig by normalizing DwarfInfo to take u64 for all addresses

LemonBoy:
>  Get rid of the ugly @intCast in Debug.zig by normalizing DwarfInfo to take u64 for all addresses

Sure thing, go ahead. Many functions take `u64` instead of `usize` because they are target-agnostic and you may want to load DWARF infos from a 64bit file on a 32bit system.

mchudleigh:
Okay, all changes discussed are pushed. Let me know if there's anything else.

mchudleigh:
Looking back on the commits in this PR, this is kind of a mess. Let me know if you'd prefer I rework it into fewer, tighter commits.

mchudleigh:
I rebased these changes onto a post 0.8 master so they will be much easier to merge.

Vexu (approved):

PR #8718, opened 2021-05-08
p256: properly handle neutral element & add AffineCoordinates struct

jedisct1:
Instead of multiple references to an anonymous structure to represent affine coordinates, add an actual `AffineCoordinates` structure.

Also properly handle the neutral element during coordinate conversion and fix mixed addition.

And comptime the small precomputation table for basepoint multiplication.

### DIFF
--- a/lib/std/crypto/pcurves/p256.zig
+++ b/lib/std/crypto/pcurves/p256.zig
@@ -49,21 +49,26 @@ pub const P256 = struct {
     }
 
     /// Create a point from affine coordinates after checking that they match the curve equation.
-    pub fn fromAffineCoordinates(x: Fe, y: Fe) EncodingError!P256 {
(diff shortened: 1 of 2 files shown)


lib/std/crypto/pcurves/p256.zig
```diff
@@ -49,21 +49,26 @@ pub const P256 = struct {
     }
 
     /// Create a point from affine coordinates after checking that they match the curve equation.
-    pub fn fromAffineCoordinates(x: Fe, y: Fe) EncodingError!P256 {
+    pub fn fromAffineCoordinates(p: AffineCoordinates) EncodingError!P256 {
```
Mouvedia, line 52: FYI
this should not really be considered breaking because `fromAffineCoordinates` was introduced a week ago

PR #8721, opened 2021-05-09
stage2: Move BlockData out of ir.Inst.Block

joachimschmidt557:
Now the codegen components (`codegen.zig`, `codegen/c.zig` and `codegen/wasm.zig`) are responsible for storing data for each block. This negates the need for `AnyMCValue` which makes dealing with the block data more convenient.


src/codegen/wasm.zig
```diff
@@ -1006,10 +1004,9 @@ pub const Context = struct {
             try self.emitWValue(operand);
         }
 
-        // every block contains a `WValue` with its block index.
+        // We map every block to its block index.
         // We then determine how far we have to jump to it by substracting it from current block depth
-        const wvalue = @bitCast(WValue, br.block.codegen.mcv);
-        const idx: u32 = self.block_depth - wvalue.block_idx;
```
Luukdegram, line 1012: This means the `block_idx` tag can be removed from `WValue` as well as any references to it.

PR #8722, opened 2021-05-09
fstatat() test now also fails on FreeBSD in ReleaseSmall mode

jedisct1:
Disable that test for now, if only because we haven't been updating -master builds for a couple days due to this.


jedisct1:
Scratch that, I found the actual issue.

PR #8723, opened 2021-05-09
Initialize the Stat structure

### LINKED ISSUE #8538: fstatat std.os.test failure on freebsd in ReleaseFast mode
After merging the llvm12 branch into master:

```
Test [619/1826] os.test.test "std-native-ReleaseFast--multi fstatat"... 
thread 100093 panic: index 9 incorrect. expected 140737488348536, found 140737488347880
/usr/home/build/zig/zig-cache/o/02926073cce247004b6b9db86ce3ad50/test /usr/home/build/zig/build/release/bin/zig
/usr/home/build/zig/build/release/bin/zig test /usr/home/build/zig/lib/std/std.zig --test-name-prefix std-native-ReleaseFast--multi  -OReleaseFast --cache-dir /usr/home/build/zig/zig-cache --global-cache-dir /home/build/.cache/zig --name test -I /usr/home/build/zig/test --override-lib-dir /usr/home/build/zig/lib 
/usr/home/build/zig/zig-cache/o/5705de67878dd3e9a26d17b70621f9e2/build /usr/home/build/zig/build/release/bin/zig /usr/home/build/zig /usr/home/build/zig/zig-cache /home/build/.cache/zig test -Dskip-compile-errors -Dskip-non-native
```

jedisct1:
The system `stat` structure includes padding, and, on some operating systems such as all BSDs, "spare" bytes at the end.

We can't reliably compare two `Stat` values if these are uninitialized, while being later compared.

This is what was causing the `fstatat` test to fail on FreeBSD since the update to LLVM 12. It was previously only passing by accident.

Fixes #8538

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -3408,7 +3408,7 @@ pub fn fstat(fd: fd_t) FStatError!Stat {
         @compileError("fstat is not yet implemented on Windows");
     }
 
-    var stat: Stat = undefined;
+    var stat: Stat = mem.zeroes(Stat);
     switch (errno(system.fstat(fd, &stat))) {
         0 => return stat,
         EINVAL => unreachable,
@@ -3459,7 +3459,7 @@ pub fn fstatatWasi(dirfd: fd_t, pathname: []const u8, flags: u32) FStatAtError!S
 /// Same as `fstatat` but `pathname` is null-terminated.
 /// See also `fstatat`.
 pub fn fstatatZ(dirfd: fd_t, pathname: [*:0]const u8, flags: u32) FStatAtError!Stat {
-    var stat: Stat = undefined;
+    var stat: Stat = mem.zeroes(Stat);
     switch (errno(system.fstatat(dirfd, pathname, &stat, flags))) {
         0 => return stat,
         EINVAL => unreachable,
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -3408,7 +3408,7 @@ pub fn fstat(fd: fd_t) FStatError!Stat {
         @compileError("fstat is not yet implemented on Windows");
     }
 
-    var stat: Stat = mem.zeroes(Stat);
+    var stat = mem.zeroes(Stat);
     switch (errno(system.fstat(fd, &stat))) {
         0 => return stat,
         EINVAL => unreachable,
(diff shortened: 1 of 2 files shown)


jedisct1:
The issue is in actual struct members used for padding:

```zig
pub const libc_stat = extern struct {
   ...
mode: mode_t,
__pad0: u16,
uid: uid_t,
gid: gid_t,
__pad1: u32,
rdev: dev_t,
   ...
__spare: [10]u64,
```

What is failing here is the `__spare` part, which is not touched by the `*stat` functions, but was reserved to avoid ABI breakage if new information gets eventually added.

We can fix that particular test to ignore `__spare` and `__pad*`, but what the test currently does, a straightforward comparison using `=`, is something applications can legitimately do.

These structures are small, and zeroing them is virtually free compared to the time taken by the actual system call.

LemonBoy:
Oh Indeed, I misread the test and thought it was using `mem.eql`. This is A case where #8713 comes in handy.

jedisct1:
These padding members are actually present pretty much everywhere, including Linux.

Introducing an `eql()` function would take something like this: https://github.com/jedisct1/zig/commit/8cf70793be81a267175c668489d2fb76dcc392fa

Fun to see that some platforms don't even use any special prefix for these (see on macOS and DragonflyBSD: `qspare`, `lspare`) 😄

Depending solely on `eql()` seems a little bit fragile. `eql()` functions must always remain correct as `stat` structures get added or updated, and if this is not the case, tests may still pass until an unrelated change is made (LLVM update).

I'd still rather zero them prior to usage instead. This requires no efforts to maintain, and applications can still safely use a regular equality test.

This is also a minimal change to turn the FreeBSD CI green. Releases haven't been updated since May 5 because of this.

jedisct1:
These structures are currently defined for each operating system.

Wouldn't usage of explicit bit offsets require that these offsets are constant across all architectures? Just seeing some `isize` here suggests that it may not be the case.

mikdusan:
The structs would still have to be defined on a per-platform basis but the mechanism would be, I think, more succinct than implementing `eql()`.

kubkon:
I'm merging this on the merit of getting our CI green again and cutting fresh nightly releases. If you feel there is a better, more future-proof way of handling this, please open a new PR, and we'll revert this one in favour of the new one.

PR #8725, opened 2021-05-09
meta.eql(): call the `eql()` function if defined

jedisct1:
Suggested in #8625

### DIFF
--- a/lib/std/meta.zig
+++ b/lib/std/meta.zig
@@ -691,10 +691,11 @@ test "std.meta.TagPayload" {
 }
 
 /// Compares two of any type for equality. Containers are compared on a field-by-field basis,
-/// where possible. Pointers are not followed.
+/// where possible, or using the `eql()` function, if defined. Pointers are not followed.
 pub fn eql(a: anytype, b: @TypeOf(a)) bool {
     const T = @TypeOf(a);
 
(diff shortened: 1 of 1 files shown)


lib/std/meta.zig
```diff
@@ -691,12 +691,13 @@ test "std.meta.TagPayload" {
...
 
 /// Compares two of any type for equality. Containers are compared on a field-by-field basis,
-/// where possible. Pointers are not followed.
+/// where possible, or using the `eql()` function, if defined. Pointers are not followed.
 pub fn eql(a: anytype, b: @TypeOf(a)) bool {
     const T = @TypeOf(a);
 
     switch (@typeInfo(T)) {
         .Struct => |info| {
+            if (comptime std.meta.trait.hasFn("eql")(@TypeOf(a))) return a.eql(b);
```
LemonBoy, line 700: This applies to other container types like unions or opaque ptrs, no?

jedisct1: Oh, right!

 The test has been moved before the `switch`.

Thank you!

g-w1:
See: https://freenode.irclog.whitequark.org/zig/2021-04-26#1619466090-1619466280;

jedisct1:
> See: https://freenode.irclog.whitequark.org/zig/2021-04-26#1619466090-1619466280;

Oh, thanks! And yeah, that makes sense.

Closing, then 😄

PR #8730, opened 2021-05-09
overhaul elf csu (c-runtime startup) logic

mikdusan:
- more support for linux, android, freebsd, netbsd, openbsd, dragonfly
    - centralize musl utils; musl logic is no longer intertwined with csu
    - fix musl compilation to build crti/crtn for full archs list
    - fix openbsd to support `zig build-lib -dynamic`
    - initial dragonfly linking success (with a warning)

    ancillary:

    - fix emutls (openbsd) tests to use `try`

---

this improves openbsd support to where:
- `Dskip-stage2-tests -Dskip-non-native` passes

caveat, for llvm12 on openbsd some prerequisite patches are required:
- [lld/ELF/DriverUtils.cpp ](https://github.com/openbsd/ports/blob/9cbccf7aa8e04febfab2bb6683baec2737a93a6b/devel/llvm/patches/patch-tools_lld_ELF_DriverUtils_cpp)
- [llvm/cmake/modules/GetLibraryName.cmake](https://github.com/semarie/zig-ports/blob/c4f94018618f760c49b9b120066bfb56f685e959/patches/patch-llvm_cmake_modules_GetLibraryName_cmake)

and initial dragonfly linking now works with this PR. The usual missing `std.os` bits are ready as a followup-patch not yet submitted, that when applied passes a `zig build test`.

### DIFF
--- a/lib/std/special/compiler_rt/emutls.zig
+++ b/lib/std/special/compiler_rt/emutls.zig
@@ -376,7 +376,7 @@ test "__emutls_get_address with default_value" {
 
 test "test default_value with differents sizes" {
     const testType = struct {
-        fn _testType(comptime T: type, value: T) void {
+        fn _testType(comptime T: type, value: T) !void {
             var def: T = value;
             var ctl = emutls_control.init(T, &def);
             var x = ctl.get_typed_pointer(T);
@@ -384,11 +384,11 @@ test "test default_value with differents sizes" {
         }
     }._testType;
 
-    testType(usize, 1234);
-    testType(u32, 1234);
-    testType(i16, -12);
-    testType(f64, -12.0);
-    testType(
+    try testType(usize, 1234);
+    try testType(u32, 1234);
+    try testType(i16, -12);
+    try testType(f64, -12.0);
+    try testType(
         @TypeOf("012345678901234567890123456789"),
         "012345678901234567890123456789",
     );
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1290,7 +1290,7 @@ pub fn create(gpa: *Allocator, options: InitOptions) !*Compilation {
         }
         if (comp.wantBuildMuslFromSource()) {
             try comp.work_queue.ensureUnusedCapacity(6);
-            if (target_util.libc_needs_crti_crtn(comp.getTarget())) {
+            if (musl.needsCrtiCrtn(comp.getTarget())) {
                 comp.work_queue.writeAssumeCapacity(&[_]Job{
                     .{ .musl_crt_file = .crti_o },
                     .{ .musl_crt_file = .crtn_o },
@@ -2903,7 +2903,7 @@ fn detectLibCIncludeDirs(
         const generic_name = target_util.libCGenericName(target);
         // Some architectures are handled by the same set of headers.
         const arch_name = if (target.abi.isMusl())
-            target_util.archMuslName(target.cpu.arch)
+            musl.archMuslName(target.cpu.arch)
         else if (target.cpu.arch.isThumb())
             // ARM headers are valid for Thumb too.
             switch (target.cpu.arch) {
--- a/src/libc_installation.zig
+++ b/src/libc_installation.zig
@@ -195,8 +195,8 @@ pub const LibCInstallation = struct {
                 errdefer batch.wait() catch {};
                 batch.add(&async self.findNativeIncludeDirPosix(args));
                 switch (Target.current.os.tag) {
-                    .freebsd, .netbsd, .openbsd => self.crt_dir = try std.mem.dupeZ(args.allocator, u8, "/usr/lib"),
-                    .linux, .dragonfly => batch.add(&async self.findNativeCrtDirPosix(args)),
+                    .freebsd, .netbsd, .openbsd, .dragonfly => self.crt_dir = try std.mem.dupeZ(args.allocator, u8, "/usr/lib"),
+                    .linux => batch.add(&async self.findNativeCrtDirPosix(args)),
                     .haiku => self.crt_dir = try std.mem.dupeZ(args.allocator, u8, "/system/develop/lib"),
                     else => {},
                 }
@@ -524,7 +524,7 @@ pub const CCPrintFileNameOptions = struct {
 };
 
 /// caller owns returned memory
-fn ccPrintFileName(args: CCPrintFileNameOptions) ![:0]u8 {
+pub fn ccPrintFileName(args: CCPrintFileNameOptions) ![:0]u8 {
     const allocator = args.allocator;
 
     const cc_exe = std.os.getenvZ("CC") orelse default_cc_exe;
--- a/src/link/Elf.zig
+++ b/src/link/Elf.zig
@@ -23,6 +23,7 @@ const File = link.File;
 const build_options = @import("build_options");
 const target_util = @import("../target.zig");
 const glibc = @import("../glibc.zig");
+const musl = @import("../musl.zig");
 const Cache = @import("../Cache.zig");
 const llvm_backend = @import("../codegen/llvm.zig");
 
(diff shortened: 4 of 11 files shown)


src/link/Elf.zig
```diff
@@ -3261,3 +3227,145 @@ fn padToIdeal(actual_size: anytype) @TypeOf(actual_size) {
...
+
+// Compute libc runtime initialization objects for a given link command.
+const CrtObjects = struct {
+    crt0: ?[]const u8 = null,
+    crti: ?[]const u8 = null,
+    crtbegin: ?[]const u8 = null,
+    crtend: ?[]const u8 = null,
+    crtn: ?[]const u8 = null,
+
+    fn compute(self: *CrtObjects, link_options: link.Options, comp: *const Compilation) void {
```
LemonBoy, line 3239: Nit, what about `fn compute(self: *CrtObjects, link_options: link.Options, comp: *const Compilation) CrtObjects` ?
This way you're explicitly stating that `CrtObjects` by itself is uninitialized.

lib/std/special/compiler_rt/emutls.zig
```diff
@@ -376,19 +376,19 @@ test "__emutls_get_address with default_value" {
...
             try expect(x.* == value);
         }
     }._testType;
 
-    testType(usize, 1234);
-    testType(u32, 1234);
-    testType(i16, -12);
-    testType(f64, -12.0);
-    testType(
+    try testType(usize, 1234);
```
LemonBoy, line 387: Can we get a bit more coverage for this test? Right now it's only running when targeting OpenBSD and thus prone to random breakage.

src/link/Elf.zig
```diff
@@ -3261,3 +3225,190 @@ fn padToIdeal(actual_size: anytype) @TypeOf(actual_size) {
...
+// Provide a blueprint of csu (c-runtime startup) objects for supported
+// link modes.
+//
+// This is for cross-mode targets only. For host-mode targets the system
+// compiler can be probed to produce a robust blueprint.
+//
+// Targets requiring a libc for which zig does not bundle a libc are
+// host-mode targets. Unfortunately, host-mode probes are not yet
+// implemented. For now the data is hard-coded here. Such targets are
+// { freebsd, netbsd, openbsd, dragonfly }.
```
ifreund, line 3238: zig will start using the system libc by default when targeting the native os/abi soon: https://github.com/ziglang/zig/pull/8737.

src/link/Elf.zig
```diff
@@ -3261,3 +3225,190 @@ fn padToIdeal(actual_size: anytype) @TypeOf(actual_size) {
...
+                    .dynamic_pie => result.set( "crt0.o",  null, "crtbegin.o",  "crtend.o",  null ),
+                    .static_exe,
+                    .static_pie  => result.set( "rcrt0.o", null, "crtbegin.o",  "crtend.o",  null ),
+                    // zig fmt: on
+                },
+                else => {},
+            }
+
+            // Bundled { glibc, musl } do not supply begin/end objects.
+            if (comp.wantBuildLibCFromSource()) {
```
ifreund, line 3346: I feel like this is code smell, I don't think the code performing the linking should know or care where the libc installation came from.

src/musl.zig
```diff
@@ -248,6 +248,42 @@ pub fn buildCRTFile(comp: *Compilation, crt_file: CRTFile) !void {
     }
 }
 
+pub fn archMuslName(arch: std.Target.Cpu.Arch) [:0]const u8 {
+    switch (arch) {
+        .aarch64, .aarch64_be => return "aarch64",
+        .arm, .armeb => return "arm",
```
LemonBoy, line 254:
```suggestion
        .arm, .armeb, .thumb, .thumbeb => return "arm",
```

src/musl.zig
```diff
@@ -248,6 +248,42 @@ pub fn buildCRTFile(comp: *Compilation, crt_file: CRTFile) !void {
...
+    }
+}
+
+// Return true if musl has arch-specific crti/crtn sources.
+// See lib/libc/musl/crt/ARCH/crt?.s .
+pub fn libc_needs_crti_crtn(target: std.Target) bool {
+    // zig fmt: off
+    return switch (target.cpu.arch) {
+        .aarch64, .aarch64_be,
+        .arm, .armeb,
```
LemonBoy, line 274:
```suggestion
        .arm, .armeb, .thumb, .thumbeb,
```

src/musl.zig
```diff
@@ -248,6 +248,42 @@ pub fn buildCRTFile(comp: *Compilation, crt_file: CRTFile) !void {
...
+        .s390x => return "s390x",
+        .wasm32, .wasm64 => return "wasm",
+        .x86_64 => return "x86_64",
+        else => unreachable,
+    }
+}
+
+// Return true if musl has arch-specific crti/crtn sources.
+// See lib/libc/musl/crt/ARCH/crt?.s .
+pub fn libc_needs_crti_crtn(target: std.Target) bool {
```
LemonBoy, line 270:
```suggestion
pub fn needsCrtiCrtn(target: std.Target) bool {
```
Or something similar but camelCase

src/link/Elf.zig
```diff
@@ -3254,3 +3218,178 @@ fn padToIdeal(actual_size: anytype) @TypeOf(actual_size) {
...
+        } else {
+            const lci = link_options.libc_installation orelse return error.LibCInstallationNotAvailable;
+            const crt_dir_path = lci.crt_dir orelse return error.LibCInstallationMissingCRTDir;
+            switch (link_options.target.os.tag) {
+                .dragonfly => {
+                    if (result.crt0) |*obj| obj.* = try fs.path.join(arena, &[_][]const u8{ crt_dir_path, obj.* });
+                    if (result.crti) |*obj| obj.* = try fs.path.join(arena, &[_][]const u8{ crt_dir_path, obj.* });
+                    if (result.crtn) |*obj| obj.* = try fs.path.join(arena, &[_][]const u8{ crt_dir_path, obj.* });
+
+                    var gccv: []const u8 = undefined;
```
LemonBoy, line 3354: Can't we probe this with the `cc` test (`ccPrintFileName`) and store it in the `libc_installation` object? That'd be more robust than hardcoding the gcc version here.

src/Compilation.zig
```diff
@@ -3009,7 +3009,7 @@ fn addBuildingGLibCJobs(comp: *Compilation) !void {
     });
 }
 
-fn wantBuildLibCFromSource(comp: Compilation) bool {
+pub fn wantBuildLibCFromSource(comp: Compilation) bool {
```
ifreund, line 3012: This and the other `wantBuild...` function no longer need to be made `pub`

src/link/Elf.zig
```diff
@@ -3254,3 +3218,178 @@ fn padToIdeal(actual_size: anytype) @TypeOf(actual_size) {
...
+                if (@field(result, f.name)) |*obj| {
+                    if (comp.crt_files.get(obj.*)) |crtf| {
+                        obj.* = crtf.full_object_path;
+                    } else {
+                        @field(result, f.name) = null;
+                    }
+                }
+            }
+        } else {
+            const lci = link_options.libc_installation orelse return error.LibCInstallationNotAvailable;
```
ifreund, line 3346: it would be more idiomatic to turn this if/else into something like
```zig
if (link_options.libc_installation) |lci| {
} else {
}
```

src/link/Elf.zig
mikdusan: The current feature set in CsuObjects is of statically-known nature and while I think runtime probes are going to be the only way going forward with true host-libc support, this PR is not the right time to do runtime probes. imho, we have to come up with a rock-solid way to bootstrap probes and cache the result and how to invalidate that cache. This was done long ago with libc.txt but it led to the question of how to invalidate the cache and ultimately, it was yanked, and became `probe libc at every zig invoke` unless manually someone places it on the command line, or in the zig-cache (I am fuzzy on the last part so grain of salt).

src/musl.zig
mikdusan: done and camelCase for remaining functions too

src/musl.zig
mikdusan: done

src/musl.zig
mikdusan: done. inverted logic.

src/Compilation.zig
mikdusan: done

src/link/Elf.zig
mikdusan: done

src/libc_installation.zig
```diff
@@ -524,7 +524,7 @@ pub const CCPrintFileNameOptions = struct {
 };
 
 /// caller owns returned memory
-fn ccPrintFileName(args: CCPrintFileNameOptions) ![:0]u8 {
+pub fn ccPrintFileName(args: CCPrintFileNameOptions) ![:0]u8 {
```
andrewrk, line 527: As far as I can tell this did not end up getting called from outside this file, right?

andrewrk (approved):
Great work!

PR #8733, opened 2021-05-10
std: rework math.scalbn

### LINKED ISSUE #8732: funny overflow in std.math.scalbn
Credit to @Adamcake for noting that the function "looks like it has a copy-paste error" after I made a remark on the visual appearance of float math functions, leading to the discovery of this bug.

## Explanation

The `scalbnf` ([musl](https://git.musl-libc.org/cgit/musl/tree/src/math/scalbnf.c)) function shifts the `n` input into the exponent bits after clamping it to `-126 <= n <= 127` as per IEEE 754 representation of f32. During the development of Zig's `std.math.scalbn32`, it seems that the double-precision version was transcribed first (where exp is `-1022 <= n <= 1023`), and then copy-pasted without changing all the literals to make sense for single-precision floats, leading it to clamp the exponent incorrectly (line 33):

https://github.com/ziglang/zig/blob/2d119677344c4ca6a2700b8de0a24215b9a37c06/lib/std/math/scalbn.zig#L30-L50

Later in the function, 0x7F is added to put the exponent into the range for the correct bit representation of floats, and then it's shifted into the exponent bits. This assumes it's correctly clamped the exponent, and this is where the ugly stuff happens if it's not:

https://github.com/ziglang/zig/blob/2d119677344c4ca6a2700b8de0a24215b9a37c06/lib/std/math/scalbn.zig#L52-L53

Note: `+%` seems to do nothing here? I guess it's just because it's transcribed from C, but `(-126 <= n <= 127) + 0x7F` wouldn't wrap an i32.

## Demo

This doesn't particularly affect any real world cases as x * 2^128 and above will get the float into +Inf territory generally, but we can observe the exponent overflow into the sign bit at n>255:

```zig
const std = @import("std");
const print = std.debug.print;
const scalbn = std.math.scalbn;

pub fn main() anyerror!void {
    const x: f32 = 0.5;
    var i: i32 = 250;
    while (i <= 260) {
        print("{} * 2^{} is {}\n", .{ x, i, scalbn(x, i) });
        i += 1;
    }
}
```

```
% zig build run
5.0e-01 * 2^250 is inf
5.0e-01 * 2^251 is inf
5.0e-01 * 2^252 is inf
5.0e-01 * 2^253 is inf
5.0e-01 * 2^254 is inf
5.0e-01 * 2^255 is inf
5.0e-01 * 2^256 is -0.0e+00
5.0e-01 * 2^257 is -1.0e+00
5.0e-01 * 2^258 is -2.0e+00
5.0e-01 * 2^259 is -4.0e+00
5.0e-01 * 2^260 is -8.0e+00
```

The fix itself is super simple, I've attached a PR. Apologies if this issue was hard to read.

notviri:
fixes #8732


notviri:
Hey, sorry for going inactive. I've rewritten it to be more generic, and it may be too generic, actually. I haven't written much Zig before at all so I wouldn't be surprised if this implementation is rejected, but I'd like a review regardless as I'm interested in contributing more in the future.

Two changes from the current implementation:
1) It now works with any float (including possible future binary256 support) and dynamically figures out the float proportions, this includes `c_longdouble` (not sure if I should have a test for that, but I do)
2) It no longer has the weird double-rounding correction from musl libc for small exponents below the float limits, as from testing I've observed no difference to rounding errors in subnormal ranges (not entirely confident in this one, but it's just a bonus, not a feature)

lib/std/math/scalbn.zig
```diff
@@ -9,89 +9,68 @@
...
-        y *= 0x1.0p-1022 * 0x1.0p53;
-        n += 1022 - 53;
-        if (n < -1022) {
-            y *= 0x1.0p-1022 * 0x1.0p53;
-            n += 1022 - 53;
-            if (n < -1022) {
-                n = -1022;
-            }
+    } else if (n_ < exponent_min) {
+        x_ *= scale_min;
```
LemonBoy, line 47: The rounding problem is still there, check out [this](https://git.musl-libc.org/cgit/musl/commit/src/math/scalbn.c?id=8c44a060243f04283ca68dad199aab90336141db) musl commit.

lib/std/math/scalbn.zig
```diff
@@ -9,89 +9,68 @@
...
+    try expect(scalbn(@as(f16, 1.5), 4) == 24.0);
+    try expect(scalbn(@as(f32, 1.5), 4) == 24.0);
+    try expect(scalbn(@as(f64, 1.5), 4) == 24.0);
+    try expect(scalbn(@as(f128, 1.5), 4) == 24.0);
+    try expect(scalbn(@as(c_longdouble, 1.5), 4) == 24.0);
 
-test "math.scalbn32" {
-    try expect(scalbn32(1.5, 4) == 24.0);
-}
+    // subnormals
```
LemonBoy, line 67: The tests should cover `f16, f32, f64, f128` and check the value against a known good one.

lib/std/math/scalbn.zig
```diff
@@ -9,89 +9,68 @@
...
+    try expect(math.isNormal(scalbn(@as(f32, 1.0), -126)));
+    try expect(!math.isNormal(scalbn(@as(f32, 1.0), -127)));
 
-test "math.scalbn64" {
-    try expect(scalbn64(1.5, 4) == 24.0);
+    // float limits
+    try expect(scalbn(@as(f32, math.f32_max), -128 - 149) > 0.0);
```
LemonBoy, line 72: Ditto.

notviri:
Thanks so much for the great feedback! I've applied basically every suggestion (ran into `std.math.isNormal` not supporting `f128` on the way, hope it's okay that I threw that in there), and I'm unhappy to say I ran into another issue. One `f16` and one `f128` test fails, and I'm not entirely sure why.

- An `f16` test fails where I added a test for all instances of [that musl commit](https://git.musl-libc.org/cgit/musl/commit/src/math/scalbn.c?id=8c44a060243f04283ca68dad199aab90336141db) you've linked.

  ```zig
  try expect(scalbn(@as(f16, 0x1.1FFp-1), -14 - 9) == math.f16_true_min);
  ```
  Instead of this being `f16_true_min`, at least on my machine, I get `0b0000000000001001`. My only guess as to why this would be happening is that the output seems to load it into an `f32`, do the calculation, and then load it back (at least on my machine):
  ```assembly
  vmovd     xmm0, eax
  vcvtph2ps xmm0, xmm0
  vmulss    xmm0, xmm0, dword ptr [rip + .some_offset]
  vcvtps2ph xmm0, xmm0, 4
  vmovd      edi, xmm0
  ```
  I have no idea though.
- An `f128` test fails in the tests for multiplying the true lowest number to become the highest normal number, and this works for all of them except `f128` where I get `+inf` for both of these:

  ```zig
  try expect(!math.isPositiveInf(scalbn(@as(f128, math.f128_true_min), 16383 + 16494)));
  try expect(math.isPositiveInf(scalbn(@as(f128, math.f128_true_min), 16383 + 16494 + 1)));
  ```
  Again, my only guess is that it's something about the software floats (or I've messed it up). I'm hoping it's something obvious. Sorry.

LemonBoy:
> Thanks so much for the great feedback!

Thank you for tackling this task, when you start messing with fp it's very likely to bump into several problems heh.

> An f16 test fails where I added a test for all instances of that musl commit you've linked.

Because f16 is garbage, a garbage language feature that should be ripped out. F16 is mostly meant as a storage format, it's only natively supported on some ARM CPUs and everywhere else implements every arithmetic op in terms of a wider and legal type.
The exension, truncation and rounding steps are very likely to mess up your result.

I can confirm that you're indeed right, `f16_true_min` is the correct result.

> An f128 test fails in the tests for multiplying the true lowest number to become the highest normal number

You've found a bug! I've fixed the problem in #8772

notviri:
I'm really relieved these weren't on my end, didn't feel too good about submitting this one ^^;

What should I do for the f16 tests then, should I have them at all if they're that shaky? As for the f128 one, thanks for fixing it so quickly! I assume it'll be merged once CI passes, I update this branch to latest master and enable those? I haven't done much parallel work like this with git so I'm not sure what's best.

LemonBoy:
> What should I do for the f16 tests then

Leave the failing commented out and write a comment explaining why, that's enough.

You can cherry-pick that PR into your branch, git has no problem in handling this kind of merges.

notviri:
No idea what lavahut's been up to, but the rest look good

andrewrk:
Thanks @notviri and thank you @LemonBoy for the review

PR #8735, opened 2021-05-10
stage2: Add CPU feature detection for macos

LemonBoy:
This is mostly meant to detect the current and future AArch64 core types
as the x86 part is already taken care of with OS-independent machinery.

One more step closer to removing LLVM's detection as fallback!

cc @kubkon as this is completely untested.

### DIFF
--- a/lib/std/os/bits/darwin.zig
+++ b/lib/std/os/bits/darwin.zig
@@ -1751,3 +1751,38 @@ pub const IOCPARM_MASK = 0x1fff;
 fn ior(inout: u32, group: usize, num: usize, len: usize) usize {
     return (inout | ((len & IOCPARM_MASK) << 16) | ((group) << 8) | (num));
 }
+
+// CPU families mapping
+pub const CPUFAMILY = enum(u32) {
+    UNKNOWN = 0,
+    POWERPC_G3 = 0xcee41549,
+    POWERPC_G4 = 0x77c184ae,
+    POWERPC_G5 = 0xed76d8aa,
+    INTEL_6_13 = 0xaa33392b,
+    INTEL_PENRYN = 0x78ea4fbc,
+    INTEL_NEHALEM = 0x6b5a4cd2,
+    INTEL_WESTMERE = 0x573b5eec,
+    INTEL_SANDYBRIDGE = 0x5490b78c,
+    INTEL_IVYBRIDGE = 0x1f65e835,
+    INTEL_HASWELL = 0x10b282dc,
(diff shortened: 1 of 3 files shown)


lib/std/os/bits/darwin.zig
```diff
@@ -1744,3 +1744,35 @@ pub const IOCPARM_MASK = 0x1fff;
 fn ior(inout: u32, group: usize, num: usize, len: usize) usize {
     return (inout | ((len & IOCPARM_MASK) << 16) | ((group) << 8) | (num));
 }
+
+// CPU families mapping
+pub const CPUFAMILY_UNKNOWN = 0;
```
daurnimator, line 1749:
```suggestion
pub const CPUFAMILY = extern enum(u32) {
  UNKNOWN = 0,
  POWERPC_G3 = 0xcee41549,
  // ....
  _,
};
```

LemonBoy: Aha, I was not sure if we wanted to turn the system constants into enums where it made sense.
It does look much better tho, I'll change the defs.

kubkon (approved):
Confirmed that works great on my M1. Here's the adhoc debug output of the detected cpu model:

```
detected model = Model{ .name = { 97, 112, 112, 108, 101, 95, 97, 49, 52 }, .llvm_name = { 97, 112, 112, 108, 101, 45, 97, 49, 52 }, .features = Set{ .ints = { 570268474681595960, 6937795225965428736, 6446645248 } } }, name = apple_a14detected model = Model{ .name = { 97, 112, 112, 108, 101, 95, 97, 49, 52 }, .llvm_name = { 97, 112, 112, 108, 101, 45, 97, 49, 52 }, .features = Set{ .ints = { 570268474681595960, 6937795225965428736, 6446645248 } } }, name = apple_a14
```

kubkon:
This is pretty neat; as a byproduct of this great PR by you @LemonBoy, behavior tests are no longer hanging on `@atomicRmw` on M1! 🎉

LemonBoy:
As a side-effect of this PR now you're getting code optimized for the FireStorm/IceStorm architecture. LLVM's cpu detection hardcodes the CPU model to "cyclone", corresponding to the old-ish A7 architecture.

kubkon:
> As a side-effect of this PR now you're getting code optimized for the FireStorm/IceStorm architecture. LLVM's cpu detection hardcodes the CPU model to "cyclone", corresponding to the old-ish A7 architecture.

Correct!

Issue #8144, opened 2021-03-04
Linking system C libraries fails if zig does not support the system glibc version


ifreund:
See: https://github.com/ifreund/river/issues/232

When the system glibc version is newer than the most recent glibc version zig ships, zig falls back to providing the most recent glibc version it can. However, this causes linking system libraries built against the newer system glibc version to fail as seen in the following build log taken from the above river issue:
```
warning: zig cannot build new glibc version 2.33; providing instead 2.32
ld.lld: error: /lib64/libxkbcommon.so: undefined reference to stat64@GLIBC_2.33 [--no-allow-shlib-undefined]
ld.lld: error: /lib64/libxkbcommon.so: undefined reference to fstat64@GLIBC_2.33 [--no-allow-shlib-undefined]
error: LLDReportedFailure
river...The following command exited with error code 1:
[...]

PR #8737, opened 2021-05-10
stage2: use system libc when targeting the native OS/ABI

ifreund:
Currently zig will always try to build its own libc and compile against
that. This of course makes sense for cross-compilation, but can cause
problems when targeting the native OS/ABI.

For example, if the system uses a newer glibc version than zig ships
zig will fall back to using the newest version it does ship. However
this causes linking system libraries to fail as they are built against a
different glibc version than the zig code is built against.

To remedy this, simply default to linking the system libc when targeting
the native OS/ABI.

Closes: #8144

### DIFF
--- a/ci/azure/macos_arm64_script
+++ b/ci/azure/macos_arm64_script
@@ -55,10 +55,10 @@ cmake .. \
   -DZIG_TARGET_MCPU="$HOST_MCPU" \
   -DZIG_STATIC=ON
 
-make $JOBS install
-
 unset CC
 unset CXX
+
+make $JOBS install
 
 # Build zig compiler cross-compiled for arm64
 cd $ZIGDIR
@@ -79,10 +79,10 @@ cmake .. \
   -DZIG_EXECUTABLE="$ZIG" \
   -DZIG_STATIC=ON
 
-make $JOBS install
-
 unset CC
 unset CXX
+
+make $JOBS install
 
 if [ "${BUILD_REASON}" != "PullRequest" ]; then
   mv ../LICENSE release/


ifreund:
This is a draft as I haven't figured out a good way to test it yet. However, from https://github.com/ifreund/river/issues/232 I know that telling zig to use the result of `zig libc` as the libc installation works fine to compile river and link its system dependencies with zig 0.7.1 on arch linux with glibc 2.33. This doesn't work without this workaround as zig 0.7.1 only shipped with glibc 2.32 and linking fails as zig tries to link with system libraries compiled against a different glibc version.

This patch is effectively equivalent to preforming the described and tested workaround in all cases if targeting the native os.

ifreund:
If this approach seems sound, we should also consider reverting the code to fallback to building the maximum available glibc version if a version zig does not yet ship is requested:
https://github.com/ziglang/zig/blob/24dfa61236aba6d1c613f929ef06e67f7ce5ddcf/src/glibc.zig#L772

ifreund:
Corrected the patch to check if the native ABI is targeted as well, which ensures `-target native-native-musl` for example continues to work as expected despite targeting the native OS.

This means that `detectLibCIncludeDirs` will now return null instead of the native libc installation if targeting the native OS but not the native ABI. This is more correct than the status quo of returning the native libc install which perhaps doesn't match the target ABI.

andrewrk (approved):

ifreund:
Looks like the CI is failing due to a missing `-lpthread`. I'll investigate where this needs to be added.

src/link/Elf.zig
```diff
@@ -1650,9 +1650,15 @@ fn linkWithLLD(self: *Elf, comp: *Compilation) !void {
...
-                    try argv.append("-lpthread");
-                    try argv.append("-lc");
+                    // This matches the order of glibc.libs
+                    try argv.appendSlice(&[_][]const u8{
+                        "-lm",
+                        "-lpthread",
+                        "-lc",
+                        "-ldl",
+                        "-lrt",
+                        "-lutil",
```
semarie, line 1660: it is breaking openbsd stage2 build : several libraries (`dl` and `rt`) doesn't exist on OpenBSD, and `util` isn't need

ifreund: Hrm, we could really use OpenBSD coverage in the CI. I guess we need to add information about what libc components each target supports to `std.Target`.

ifreund: @semarie I've opened #8754 to fix the build for now.

I'm also considering ways we can improve on this situation and not link libc components that aren't required .

PR #8739, opened 2021-05-10
Fix AutoArrayHashMap's store_hash logic

mattbork:
`ArrayHashMap` and `ArrayHashMapUnmanaged` take a comptime bool `store_hash` that is meant to be false when the `eql` function for keys is cheap and true when it isn't. The helper `autoEqlIsCheap` returns true for primitives and a few other types, and its return value is passed directly for `store_hash` by `AutoArrayHashMap` and `AutoArrayHashMapUnmanaged`. This means `store_hash` is true when the key has a cheap `eql` function, which wastes space, and `store_hash` is false when the key has an expensive `eql` function, which wastes time. In other words, this is exactly backwards to what was intended, and `store_hash` should instead be `!autoEqlIsCheap(K)`.

Two tests that directly handled `Entry` structs had to be changed because they were storing hashes for a key type of `i32` (which has a cheap `eql`), and this change may break user code doing anything similar. I also added a test to check that an `AutoArrayHashMap`'s `Entry` type does in fact have a void or non-void `hash` field as expected.

### DIFF
--- a/lib/std/array_hash_map.zig
+++ b/lib/std/array_hash_map.zig
@@ -18,11 +18,11 @@ const builtin = @import("builtin");
 const hash_map = @This();
 
 pub fn AutoArrayHashMap(comptime K: type, comptime V: type) type {
-    return ArrayHashMap(K, V, getAutoHashFn(K), getAutoEqlFn(K), autoEqlIsCheap(K));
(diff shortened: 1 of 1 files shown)


andrewrk:
Oops! Nice catch, thank you!

PR #8741, opened 2021-05-11
bsd: detect os version

mikdusan:
- netbsd, dragonly: use sysctlbyname
- openbsd: use sysctl
- updated default semver ranges

### DIFF
--- a/lib/std/os/bits/openbsd.zig
+++ b/lib/std/os/bits/openbsd.zig
@@ -298,15 +298,6 @@ pub const AI_NUMERICSERV = 16;
 /// only if any address is assigned
 pub const AI_ADDRCONFIG = 64;
 
-pub const CTL_KERN = 1;
-pub const CTL_DEBUG = 5;
-pub const CTL_HW = 6;
-
-pub const KERN_PROC_ARGS = 55;
-pub const KERN_PROC_ARGV = 1;
-
-pub const HW_NCPUONLINE = 25;
-
 pub const PATH_MAX = 1024;
 
 pub const STDIN_FILENO = 0;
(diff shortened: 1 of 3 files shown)


semarie:
regarding OpenBSD, please do not include unused sysctl constants: it is a shame to keep them up to date, specially for `xxx_MAXID` values. OpenBSD could remove some of them between versions, and it will be complex to update them here.

andrewrk (changes requested):
Nice work- I have 2 requests

andrewrk (approved):

lib/std/os/bits/openbsd.zig
```diff
@@ -1237,3 +1228,144 @@ pub const POLLNORM = POLLRDNORM;
...
+pub const KERN_PROC_VMMAP = 80;
+pub const KERN_GLOBAL_PTRACE = 81;
+pub const KERN_CONSBUFSIZE = 82;
+pub const KERN_CONSBUF = 83;
+pub const KERN_AUDIO = 84;
+pub const KERN_CPUSTATS = 85;
+pub const KERN_PFSTATUS = 86;
+pub const KERN_TIMEOUT_STATS = 87;
+pub const KERN_UTC_OFFSET = 88;
+pub const KERN_VIDEO = 89;
```
semarie, line 1330: how to properly define `KERN_VIDEO` only for OpenBSD 6.9 ? this constant doesn't exist in 6.8

mikdusan: this is an approach that would work
note: we currently do not parse openbsd version if specified in a `-target x86_64-openbsdX.Y` but that is the mechanism that will allow sometime on openbsd69 to target openbsd68 . And with this code, it would emit a compiler error:

```zig
const __RELEASE__: u32
    = builtin.os.version_range.semver.min.major * 1_000_000
    + builtin.os.version_range.semver.min.minor * 1_000
    + builtin.os.version_range.semver.min.patch;

pub usingnamespace if (__RELEASE__ >= 6_009_000) struct {
    pub const KERN_VIDEO = 89;
} else struct {};
```

error produced if building on 6.8:

image

PR #8744, opened 2021-05-11
azure: bump timeout for macos-arm64

mikdusan:


kubkon (approved):
Did we reach the timeout value? That’s unfortunate as I was rather happy that arm64 builds managed to shave off almost 10mins compared to using system tools.

mikdusan:
Yes there were 2 builds that timed out on arm64. Both of them during docgen and this poor performance was almost certinaly due to bad apples in the mac pool (get it? :) ) and/or some peak times and the performance drops off a cliff.

PR #8745, opened 2021-05-12
std: add math.Interval

nektro:
Implements https://en.wikipedia.org/wiki/Interval_(mathematics) for Zig.

`T` may be any of `uN`, `iN`, `f16`, `f32`, `f64`, `f128`.

Properly handles all range types:
- `[a, b]`
- `[a, b)`
- `(a, b]`
- `(a, b)`
- `a` is `-inf`
- `b` is `+inf`


mattbork:
Maybe `open` should be ignored when an endpoint is at infinity. For example, if `T` is an integer type and `from` and `to` are both null then `std.math.maxInt(T)` will only be inside the interval if `to` is also closed, whereas I'd expect the finite `maxInt(T)` to be inside both `[-∞,∞]` and `[-∞,∞)`. The `minInt` isn't affected because of the order of inf cases.

On the other hand, for floats, `std.math.inf(T)` probably should only be in `[-∞,∞]` and not in `[-∞,∞)`. Or perhaps floating inf should be always be considered outside any interval.

lib/std/math/Interval.zig
```diff
@@ -0,0 +1,101 @@
...
+            const from = self.from.n orelse min;
+            const to = self.to.n orelse max;
+            const left = if (self.from.open) x > from else x >= from;
+            const right = if (self.to.open) x < to else x <= to;
+            return left and right;
+        }
+    };
+}
+
+//
```
g-w1, line 61: Not sure why this comment is hanging out here?

nektro:
@mattbork -inf and +inf are handled specially here

https://github.com/ziglang/zig/pull/8745/files#diff-129101bf3397b523d6149596c15ed27e9abadbb8a2e436923b145edda9be83c4R40-R51

lib/std/math/Interval.zig
nektro: I added this when I was working on it to separate the public API/implementation from the tests. If this is against style, I can remove it.

mattbork:
@nektro I know, but if T is an integer, and `from` and `to` are *both* null, then when the special case for `(-inf` is run:

```
const to = self.to.n orelse max;
const right = if (self.to.open) x < to else x <= to;
return right;
```

the local `to` will be set to max. So if `self.to.open` is true, then the finite `std.math.maxInt(T)` will be excluded from the interval. As I said, I would expect any finite integer to be conatined in both [-∞,∞] and [-∞,∞). However,

```
var x = Interval(u32){
    .from = .{ .n = null, .open = false  },
    .to =   .{ .n = null, .open = true },
};

try std.testing.expect(x.contains(std.math.minInt(u32))); // Ok
try std.testing.expect(x.contains(std.math.maxInt(u32))); // Fails
```

nektro:
Ah, good catch

lib/std/math/Interval.zig
```diff
@@ -0,0 +1,101 @@
...
+        from: Endpoint,
+        /// The right hand side of the Interval.
+        /// n as null here represent infinity.
+        /// open determines if the interval should be open (exclusive) or closed (inclusive). 
+        to: Endpoint,
+
+        pub const Endpoint = struct {
+            n: ?T,
+            open: bool,
+        };
```
marler8997, line 20: It seems to me that the optional modifier should be put on the fields `from` and `to` rather than `n`, since I'm not sure that the "open" field means anything when `n` is null.

Or you could use a tagged union like this:

```zig
const Endpoint = union(enum) {
    Infinity: void,
    Bounded: struct { val: T, open: bool },
};
```

This is another variation that could be used.
```zig
const Endpoint = union(enum) {
    Infinity: void,
    Open: T,
    Closed: T,
};
```

Here's what it initializing a field would look like using the various representations:
```zig
// current
const unbounded = Interval(u8) { .from = .{ .n = null, .open = undefined }, .to = .{ .n = null, .open = undefined } };
// optional field
const  unbounded = Interval(u8) { .from = null, .to = null };
// tagged union
const unbounded = Interval(u8) { .from = EndPoint.Infinity, .to = EndPoint.Infinity };
```

nektro: I like this a lot, I'll update

nektro: updated 👍

lib/std/math/Interval.zig
```diff
@@ -0,0 +1,110 @@
+const std = @import("../std.zig");
+
+/// Create an interval of type T.
+/// See https://en.wikipedia.org/wiki/Interval_(mathematics) for more info.
+pub fn Interval(comptime T: type) type {
```
daurnimator, line 5: What if open/closed-ness was a comptime property of the type?

lib/std/math/Interval.zig
```diff
@@ -0,0 +1,110 @@
...
+        /// n as null here represents infinity.
+        /// open determines if the interval should be open (exclusive) or closed (inclusive). 
+        from: Endpoint,
+        /// The right hand side of the Interval.
+        /// n as null here represent infinity.
+        /// open determines if the interval should be open (exclusive) or closed (inclusive). 
+        to: Endpoint,
+
+        pub const Endpoint = union(enum) {
+            Infinity,
```
daurnimator, line 18: this doesn't make sense for floating point types

lib/std/math/Interval.zig
```diff
@@ -0,0 +1,110 @@
...
+            Closed: T,
+            Open: T,
+
+            pub fn n(self: Endpoint) ?T {
```
daurnimator, line 22: why `n`? (leave a doc-comment?)

lib/std/math/Interval.zig
```diff
@@ -0,0 +1,110 @@
...
+            const right = if (self.to == .Open) x < to else x <= to;
+            return left and right;
+        }
+    };
+}
+
+fn runTest(comptime T: type, from: Interval(T).Endpoint, to: Interval(T).Endpoint) !void {
+    var x = Interval(T){ .from = from, .to = to };
+
+    try std.testing.expect(x.contains(0) == (x.from == .Infinity));
```
daurnimator, line 77: use `expectEqual`

MasterQ32:
Can you please add tests for certain corner cases with integer types? Using `0…255` inclusive for `u8`, `-128…127` inclusive `i8`, `0…0` for `u0`, `0…1` for `u1` and `-1…0` for `i1`. These ranges make sure you handle your corner cases correctly and do not accidently widen or overflow a interval

lib/std/math/Interval.zig
LemonBoy: Unless you're working with bigints it doesn't even make sense for non-fp types, the number of bits is finite and the upper/lower limit is well known.

LemonBoy:
Interesting idea, here's a few notes:
* The `lower <= upper` precondition is never enforced, you either need to litter every method with assert calls or provide some `init` method (and make it fallible, that's also ugly)
* Many useful methods are missing (eg. arithmetic ops, a check for interval-contains-interval, an `empty` method, equality (via `eql`) and comparison (using `math.Order`), intersection)

Check out this [nice design document](https://www-sop.inria.fr/members/Sylvain.Pion/cxx/interval/interval.pdf), it's IMO pretty complete.

On the other hand, what's the use case for interval arithmetic and why does it deserve a place in the stdlib? IOW what's the elevator pitch?

nektro:
@MasterQ32 good idea, I'll add those

@LemonBoy thanks for the reference :) the elevator pitch would be that it works beyond arithmetic. my initial driver to have this was to have a stdlib construct for Ranges and future work on this would add support for non-number types and support for (maybe in this or in another type) negative ranges

nektro:
for instance, maybe instead of asserting that `T` is a number, it could check if `T` is a number or `T` has `.compareTo`, `.posInf`, and `negInf`

lib/std/math/Interval.zig
nektro: good catch, this isn't super self explanatory unless you look at the code. it's a helper to get the `T` value out of `.Open` or `.Closed` if those are set or null if its `.Infinity`

lib/std/math/Interval.zig
nektro: okay 👍

lib/std/math/Interval.zig
nektro: because its defined in the `Endpoint` struct and its not something that inherently needs to be an invariant so I left it there in the struct over having an `.init()` function

lib/std/math/Interval.zig
nektro: why doesn't it make sense?

LemonBoy:
> so that I could represent codepoint ranges for unicode blocks in a library I'm working on

Side note, look into folded tries for that purpose, it's quite space/time efficient as representation and is currently employed by many (all?) unicode libraries.

lib/std/math/Interval.zig
xackus: You can just make a closed endpoint with infinity (fp) or the max/min possible value (int).

nektro: perhaps but including `.infinity` makes intent clearer when constructing the value and removes a branch when in `.contains(x)`

xackus: Just make a function that returns the infinity value.
I don't see how it would result in less branches. LLVM is definitely smart enough to optimize out `num <= math.maxInt(T)`.
In the general case there are more branches, because you have to test for `.infinity`.

nektro: thanks for the insight everyone, contains has now been cleaned up significantly

andrewrk:
Thanks for the submission. This looks like a good candidate for a third party package to me - if we want to continue the discussion of adding it to the std lib, let's do that in a proposal issue.

PR #8746, opened 2021-05-12
Various Linux/SPARCv9 fixes

koachan:
Some small fixes for Linux/SPARCv9 to make more of the testsuite pass.

In particular, the dirent64 change should also benefit other arches that require strict alignment like RISC-V.


lib/std/fs.zig
```diff
@@ -501,7 +501,7 @@ pub const Dir = struct {
         },
         .linux => struct {
             dir: Dir,
-            buf: [8192]u8, // TODO align(@alignOf(os.dirent64)),
+            buf: [8192]u8 align(if (builtin.os.tag != .linux) 1 else @alignOf(os.linux.dirent64)),
```
xackus, line 504: `builtin.os.tag != .linux` will always be false here.

koachan: Yes, but it's there to prevent compile error from `os.linux.dirent64` definition not being imported when compiling for other OSes.

xackus: Ok, then a comment is a good idea 👍

koachan: Done :+1:

PR #8747, opened 2021-05-12
stage2: Drop LLVM's host CPU detection method as fallback

LemonBoy:
The CPU detection code is nearly at feature parity, we do support
detecting the native CPU on Sparc systems and macos, our ARM/AArch64
model list is quite comprehensive and so is our PPC one.

The only missing pieces are:
- ARM32 detection on Darwin hosts (I don't think anybody is planning on
  running the compiler on a old-ass iPhone)
- s390x detection on Linux hosts, this can be easily added at a later
  stage.

### DIFF
--- a/lib/std/zig/system.zig
+++ b/lib/std/zig/system.zig
@@ -208,11 +208,6 @@ pub const NativeTargetInfo = struct {
 
     dynamic_linker: DynamicLinker = DynamicLinker{},
 
-    /// Only some architectures have CPU detection implemented. This field reveals whether
-    /// CPU detection actually occurred. When this is `true` it means that the reported
-    /// CPU is baseline only because of a missing implementation for that architecture.
-    cpu_detection_unimplemented: bool = false,
-
     pub const DynamicLinker = Target.DynamicLinker;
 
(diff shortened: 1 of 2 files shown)


zigazeljko:
> ARM32 detection on Darwin hosts

Does Zig even run on ARM32 Darwin hosts (or any 32-bit Darwin hosts)? AFAIK it does not even compile due to some bugs in the standard library.

kubkon:
> > ARM32 detection on Darwin hosts
> 
> Does Zig even run on ARM32 Darwin hosts (or any 32-bit Darwin hosts)? AFAIK it does not even compile due to some bugs in the standard library.

Plus we do not intend on supporting 32bit MachO by `zig ld`, so I think it's safe to ignore this target.

andrewrk:
> Does Zig even run on ARM32 Darwin hosts (or any 32-bit Darwin hosts)? AFAIK it does not even compile due to some bugs in the standard library.

https://ziglang.org/learn/overview/#support-table

32-bit x86 is Tier 4, 32-bit arm is N/A, meaning, "we do not recognize this as a target that anyone uses for anything"

PR #8748, opened 2021-05-12
fix emutls tests after #8686

semarie:
allow testType() to fail too.


LemonBoy:
Fixed in #8730

semarie:
@LemonBoy thanks for point it to me !

Issue #8653, opened 2021-04-30
fs.copyFile test fails with qemu-6.0.0 and target i386 linux none


mikdusan:
- manifests with qemu-6.0.0 and i386
- does NOT manifest with qemu-5.2.0 and i386
- does NOT manifest with qemu-6.0.0 and aarch64
- related PR #8651
- ci qemu tarball: http://ziglang.org/deps/qemu-linux-x86_64-6.0.0.tar.xz
- ci qemu tarball: http://ziglang.org/deps/qemu-linux-x86_64-5.2.0.tar.xz
  
reduction:
```bash
$ zig test -target i386-linux-none lib/std/fs.zig --test-cmd /opt/qemu-linux-x86_64-6.0.0/bin/qemu-i386 --test-cmd-bin --test-filter "copyFile"
[default] (err): Unseekable.copyFile"...
/home/mike/project/zig/work/qemu/lib/std/heap.zig:720:17: 0x409f8b in std.heap.FixedBufferAllocator.resize (test)
                return error.OutOfMemory;
                ^
/home/mike/project/zig/work/qemu/lib/std/os.zig:5465:26: 0x410bfe in std.os.copy_file_range (test)
            EOVERFLOW => return error.Unseekable,
                         ^
/home/mike/project/zig/work/qemu/lib/std/fs.zig:2536:25: 0x40db31 in copy_file (test)
            const amt = try os.copy_file_range(fd_in, offset, fd_out, offset, math.maxInt(u32), 0);
                        ^
/home/mike/project/zig/work/qemu/lib/std/fs.zig:1963:9: 0x40c9e9 in Dir.copyFile (test)
        try copy_file(in_file.handle, atomic_file.file.handle);
        ^
/home/mike/project/zig/work/qemu/lib/std/fs/test.zig:644:5: 0x40a2e0 in fs.test.test "fs.copyFile" (test)
    try tmp.dir.copyFile(src_file, tmp.dir, dest_file, .{});
    ^
/home/mike/project/zig/work/qemu/lib/std/special/test_runner.zig:83:17: 0x432c23 in std.special.main (test)
                return err;
                ^
error: the following test command failed with exit code 1:
/opt/qemu-linux-x86_64-6.0.0/bin/qemu-i386 lib/std/zig-cache/o/f389793c3c4816b8128e2418efb5b41f/test /home/mike/project/zig/work/qemu/_build/zig
```

mikdusan:
[...]

PR #8750, opened 2021-05-12
x/io, x/os: async i/o reactor, cross-platform socket syscalls and bits

lithdew:
Cross-platform versions of msghdr, sendmsg, recvmsg, linger, and iovec were provided based on findings from glibc, musl, and Microsoft's documentation.

Implemented initial Reactor interface for epoll (linux) which wraps around I/O reactor subsystems such as epoll, kqueue, select, etc. across different platforms. The Reactor interface allows for driving async I/O in Zig applications.

A test was added for the Reactor interface to drive a TCP client/listener socket pair.

A greatest-common-subset of possible socket initialization flags (close socket on exec syscalls, initialize socket to be non-blocking) were implemented.

A test was added for using sendmsg/recvmsg syscalls across different platforms for a TCP client/listener socket pair.


lib/std/c.zig
```diff
@@ -166,7 +166,7 @@ pub extern "c" fn sendto(
     dest_addr: ?*const sockaddr,
     addrlen: socklen_t,
 ) isize;
-pub extern "c" fn sendmsg(sockfd: fd_t, msg: *const std.x.os.Socket.Message, flags: c_int) isize;
```
LemonBoy: Not quite, the libc/kernel definitions are set in stone and we shall try to follow them instead of changing the parameter signedness case-by-case.
This also avoids the horrendous conditional casts that one needs when switching between the libc interface and the syscall wrappers on Linux.

lithdew: By following them, do you mean to keep `flags` as c_int?

LemonBoy: Yep, sprinkle casts here and there as needed.

lithdew: Ok! Updated the definitions and added casts where necessary; thanks!

lithdew:
On Azure CI on x86_64-linux-musl, it seems that:

```c
sendmsg(
  socket(AF_INET, SOCK_STREAM | SOCK_CLOEXEC, IPPROTO_TCP),
  msghdr {
    iovecs: [ "hello", " world" ],
    iovecs_len: 2,
    // ... everything else is zeroed
  },
  0,
)
```

for some reason returns EMSGSIZE. The test code is:

```zig
    const message = "hello world";
    _ = try conn.client.writeVectorized(Socket.Message.fromBuffers(&[_]Buffer{
        Buffer.from(message[0 .. message.len / 2]),
        Buffer.from(message[message.len / 2 ..]),
    }), 0);
```

The full stack trace is:

```
Test [1415/1884] 
x.net.tcp.test "std-x86_64-linux-musl-Debug-c-multi tcp/client: read and write multiple vectors"... FAIL (MessageTooBig)
/home/vsts/work/1/s/lib/std/x/os/socket_posix.zig:69:13: 0xe23199 in x.os.socket_posix.Mixin(x.os.socket.Socket).read (test)
            return os.recv(self.fd, buf, flags);
            ^
/home/vsts/work/1/s/lib/std/x/net/tcp.zig:136:9: 0xe22ef7 in x.net.tcp.Client.read (test)
        return self.socket.read(buf, flags);
        ^
/home/vsts/work/1/s/lib/std/x/net/tcp.zig:70:13: 0xe22ea3 in x.net.tcp.Reader.read (test)
            return self.client.read(buffer, self.flags);
            ^
/home/vsts/work/1/s/lib/std/io/reader.zig:32:13: 0xcb6daf in io.reader.Reader(x.net.tcp.Reader,@typeInfo(@typeInfo(@TypeOf(x.net.tcp.Reader.read)).Fn.return_type.?).ErrorUnion.error_set,x.net.tcp.Reader.read).read (test)
            return readFn(self.context, buffer);
            ^
/home/vsts/work/1/s/lib/std/os.zig:5031:29: 0xe237bf in os.sendmsg (test)
                EMSGSIZE => return error.MessageTooBig,
                            ^
/home/vsts/work/1/s/lib/std/x/os/socket_posix.zig:82:13: 0xe235d9 in x.os.socket_posix.Mixin(x.os.socket.Socket).writeVectorized (test)
            return os.sendmsg(self.fd, msg, flags);
            ^
/home/vsts/work/1/s/lib/std/x/net/tcp.zig:149:9: 0xcb70c7 in x.net.tcp.Client.writeVectorized (test)
        return self.socket.writeVectorized(msg, flags);
        ^
/home/vsts/work/1/s/lib/std/x/net/tcp.zig:403:9: 0x4d0174 in x.net.tcp.test "std-x86_64-linux-musl-Debug-c-multi tcp/client: read and write multiple vectors" (test)
    _ = try conn.client.writeVectorized(Socket.Message.fromBuffers(&[_]Buffer{
        ^
```

Might anyone know what could be the cause?

My hunch is that it may be an ABI issue with the layout of `msghdr`, though the definition of `msghdr` (defined in the PR as `std.x.os.Socket.Message`) (should) match the definition of `msghdr` in musl.

Running the test with the target locally on my Linux laptop passes. All other CI targets seem to pass (still waiting for Lavahut CI to finish running tests).

cc @LemonBoy @daurnimator @ifreund

LemonBoy:
I can't reproduce the problem locally, try adding some debug prints to see what the msghdr contents are before it's passed to the kernel.
Also keep in mind that (for some unknown reason) the x86_64 tests are going trough qemu's user mode emulator.

lithdew:
@LemonBoy printed out the msghdr just now alongside information about the target; it's contents seem to be fine. I'll add more debug prints in the next commit to see the contents of the `iovec`'s that are part of the `msghdr` as well.

```
sizeof(usize): 8, endian: Endian.Little, message: Message{ .name = 0, .name_len = 0, .buffers = 274894677896, .buffers_len = 2, ._pad_1 = 0, .control = 0, .control_len = 0, ._pad_2 = 0, .flags = 0 }, flags: 0
```

lithdew:
Still haven't figured out how to get the Azure CI Linux target passing for `sendmsg()`. The Azure CI Linux target is x86_64-linux-musl, has all addresses specified as 64-bits, has the endianness of integers specified as little-endian, and is essentially Ubuntu 18.04 (w/ a patched Linux kernel).

The `msghdr` struct passed to `sendmsg()` in Zig follows the same layout as musl's `msghdr` struct (which can be found in include/sys/socket.h), and the bytes representation of the `msghdr` struct passed to `sendmsg()` in Zig seems perfectly fine:

```
{ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 120, 231, 255, 0, 64, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 }
```

The size of the `msghdr` struct is 56 bytes, and its alignment is 8 bytes as expected. The contents of the pointer to the `iovec`'s contained in the `msghdr` aren't malformed either, which I checked by inspecting the contents and lengths specified in each `iovec`.

The socket write buffer size is 2626560 bytes, which is the same on my laptop; so I assume that patches made by Azure aren't interfering with any settings to do with stream sockets.

Yet, sendmsg() still returns EMSGSIZE.

Is there any way to, for example, enable strace in the CI or to emulate the environment locally? Or, might anyone have any other ideas in mind on what could be wrong?

LemonBoy:
You're ignoring a pretty huge detail here: [qrmu](https://github.com/void-linux/void-packages/issues/23557)

lithdew:
> You're ignoring a pretty huge detail here: [qrmu](https://github.com/void-linux/void-packages/issues/23557)

Thanks for the link, it finally makes a lot of sense 😌. Looks like the only way around it is to patch qemu on CI to zero out the msghdr struct though.

@kubkon @andrewrk is patching and building qemu for CI something desirable? Or might there be any other quick workarounds for this?

LemonBoy:
>  Looks like the only way around it is to patch qemu on CI to zero out the msghdr struct though.

Consider [sending a patch](https://wiki.qemu.org/Contribute/SubmitAPatch) upstream.

cc @mikdusan as he was considering an upgrade to Qemu 6.0

mikdusan:
>cc @mikdusan as he was considering an upgrade to Qemu 6.0

- we currently use a statically-linked linux binary of qemu-5.2.0 on ci
- qemu 6.0 is in a holding pattern due to #8653 and (currently) no compelling need to jump ahead of 5.2.0
- we build our own qemu for ci [this](https://github.com/ziglang/qemu-static) way

I wouldn't be against patching qemu here before it shows up in upstream releases but if you want to pursue this angle I'd need to check first with andrew. Let me know if this is the case.

Side note: also unknown is if 6.0.0 needs patching wrt this issue. Ie: would an upgrade to 6.0.0 just fix things for you.

Side note 2: #8653 has a link to 6.0.0 tarball if you can test locally .

LemonBoy:
6.0 has the same problem, the problem was never patched nor reported upstream

andrewrk:
Patching is acceptable if we first send the patch upstream

lithdew:
@LemonBoy @mikdusan @andrewrk

Sent the patch upstream to QEMU here: https://patchew.org/QEMU/20210516091536.1042693-1-kenta@lithdew.net/

Should I make a PR to add the patch to the custom QEMU w/ patches that Zig uses now?

mikdusan:
@lithdew yes please add a PR at https://github.com/ziglang/qemu-static

lithdew:
All tests are passing on CI - except Drone, looks like something's up with it for recent commits/PR's.

kubkon:
> All tests are finally on CI smile.
> 
> ... except Drone though, looks like something's up with Drone CI for recent commits/PR's.

Yep, Drone seems unable to connect to Github lately.

lithdew:
All CI tests are passing! :smile:

lib/std/os.zig
```diff
@@ -4983,11 +4983,11 @@ pub fn sendmsg(
     /// The file descriptor of the sending socket.
     sockfd: socket_t,
     /// Message header and iovecs
-    msg: msghdr_const,
+    msg: std.x.os.Socket.Message,
```
andrewrk, line 4986: I thought the plan was to make an experimental namespace, to try out some new ideas, but this is reaching into the `std.os` namespace and making breaking changes.

lib/std/os/bits/darwin.zig
```diff
@@ -23,13 +23,6 @@ pub const sockaddr = extern struct {
     family: sa_family_t,
     data: [14]u8,
 };
-pub const sockaddr_storage = extern struct {
```
andrewrk, line 26: Why are these removed?

lib/std/x.zig
```diff
@@ -8,6 +8,7 @@ const std = @import("std.zig");
 
 pub const os = struct {
     pub const Socket = @import("x/os/socket.zig").Socket;
+    pub usingnamespace @import("x/os/io.zig");
     pub usingnamespace @import("x/os/net.zig");
```
andrewrk, line 12: can you justify the use of `usingnamespace` here? is it really necessary?

lib/std/x/os/net.zig
```diff
@@ -20,7 +20,7 @@ pub fn resolveScopeID(name: []const u8) !u32 {
     if (comptime @hasDecl(os, "IFNAMESIZE")) {
         if (name.len >= os.IFNAMESIZE - 1) return error.NameTooLong;
 
-        if (comptime builtin.os.tag == .windows) {
+        if (comptime native_os.tag == .windows) {
```
andrewrk, line 23: This comptime and the one on line 20 are redundant

andrewrk (changes requested):

lib/std/os.zig
lithdew: The reason for replacing `msghdr` and `msghdr_const` in place of `std.x.os.Socket.Message` is that `std.x.os.Socket.Message` defines `msghdr` and `msghdr_const` once for the 16 targets in CI rather than 16 different times in different platform-specific files.

Though, it definitely is a breaking change in hopes of cleaning up some of the OS bits in the current Zig standard library. I'm happy to revert and add back the `msghdr` and `msghdr_const` bits, or another option would be to move `std.x.os.Socket.Message` into std.os.bits since the definition comes straight out from findings in musl/glibc.

What do you think?

lib/std/os/bits/darwin.zig
lithdew: They were removed in favor of `std.x.os.Socket.Address.Native.Storage` which is the same definition as `sockaddr_storage`, though defined just once with platform differences taken into consideration. Though, I can revert the change as mentioned in the comment about `msghdr`.

lib/std/x.zig
lithdew: Nope, no justification - I'm hoping to address this in the next PR as I'll be breaking down the current code in x/net/io.zig further into separate files and removing that line (i.e. changing it to `pub const Reactor = @import("x/os/reactor.zig")`). Just for this PR, I'm exposing the contents of x/net/io.zig using `usingnamespace`.

lib/std/x/os/net.zig
lithdew: Thanks! I'll remove the `comptime` :smile:.

lithdew:
Rebased with master to fix merge conflicts.

lithdew:
All CI tests are now passing :smile:.

PR #8752, opened 2021-05-12
std.hash_map: use 7 bits of metadata instead of 6

Sahnvour:
>we only effectively need 1 control bit to represent 2 special states for
the metadata (free and tombstone)
this should reduce the number of actual element equality tests, but since
it's very low already, the impact is negligible

I benchmarked it but did not see any evidence of performance change.
This is mostly a more satisfiying implementation. 🙂

### DIFF
--- a/lib/std/hash_map.zig
+++ b/lib/std/hash_map.zig
@@ -304,29 +304,32 @@ pub fn HashMapUnmanaged(
         /// Metadata for a slot. It can be in three states: empty, used or
         /// tombstone. Tombstones indicate that an entry was previously used,
         /// they are a simple way to handle removal.
-        /// To this state, we add 6 bits from the slot's key hash. These are
+        /// To this state, we add 7 bits from the slot's key hash. These are
         /// used as a fast way to disambiguate between entries without
         /// having to use the equality function. If two fingerprints are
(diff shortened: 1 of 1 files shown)


andrewrk (approved):

lib/std/hash_map.zig
```diff
@@ -304,29 +304,31 @@ pub fn HashMapUnmanaged(
...
+        /// when multiple elements with different hashes end up wanting to be in the same slot.
         /// Not using the equality function means we don't have to read into
-        /// the entries array, avoiding a likely cache miss.
+        /// the entries array, likely avoiding a cache miss and a potentially
+        /// costly function call.
         const Metadata = packed struct {
-            const FingerPrint = u6;
+            const FingerPrint = u7;
+            const Free = 0b000_0000;
+            const Tombstone = 0b000_0001;
```
ifreund, line 321: Technically these should be `snake_case` not `TitleCase` as they are not types.

mattbork:
Isn't there a remote chance that the high 7 bits of a hash end up being 0 or 1? Then an active entry would be mistaken for free or deleted. This may be especially a danger when used with a user supplied hashing function, where there are no guarantees about statistical quality.

PR #8753, opened 2021-05-12
stage2+macos: always link (our static) libc++

kubkon:
With this change, it is now possible to use zig-bootstrap as the
initial C/C++ compiler, and then build stage2 with the resultant
stage1 without the need to link any system libs!


andrewrk (approved):
commit message is a bit misleading - at first I thought this meant we unconditionally put libc++.a on the linker line for all macos builds (not just when building zig itself)

patch looks good!

kubkon:
> commit message is a bit misleading - at first I thought this meant we unconditionally put libc++.a on the linker line for all macos builds (not just when building zig itself)
> 
> patch looks good!

Apologies for the confusing commit message!

Issue #4298, opened 2020-01-27
set memory to undefined in allocator implementations rather than interface


andrewrk:
Follow-up from #4087, where [I wrote](https://github.com/ziglang/zig/issues/4087#issuecomment-571421800):

> Yes it makes sense to do it in the interface rather than in allocator implementations.

Having looked at the patch and seen this comment:

```zig
         // Note: can't set shrunk memory to undefined as memory shouldn't be modified on realloc failure 
```

I am now convinced that it should be done in the allocator implementations rather than the interface, where this problem can be solved. My reasoning is that a robust debug allocator will want to do this (handle realloc correctly with regards to defined), and while it wouldn't hurt semantically to have the memory set to undefined twice, this is an area where performance might matter a great deal, and it is better design to have one canonical place where the memory is set to undefined.

The allocator implementations may be able to much more efficiently set the data to undefined for allocations as well.

One concern is that allocator implementations will "forget" to do this. For this I propose that the standard library actually include a robust allocator test, which will work on any std.mem.Allocator API. People writing their own allocators should be able to call `std.testing.allocatorInterfaceStressTest(&my_allocator, options);` and be reasonably confident their allocator is working correctly, including with regards to setting memory to undefined. The `options` could include a flag to disable testing whether the memory gets set to undefined in debug/release-safe builds.

dimdin:
There is another reason:
Setting bytes to undefined in free, hides double free detection.
The second free becomes use after free, when `@memset` to `undefined` happens.

Cloudef:
I feel like this can easily lead to multiple useless memsets when one wraps multiple allocators on top of each other. There possibility should also be a way to allocate without memset for use-cases where you want to reserve big chunk of memory but not touch it

gabe-lee:
So the tl;dr as I see it is:
CASE A: `Allocator` interface handles `@memset(_, undefined)`
- Difficulties undefining in certain cases because interface has limited access to underlying implementation
- Memory undefining performed only once and guaranteed by the interface (or is expected to be)

CASE B:  Allocator internal implementation handles `@memset(_, undefined)`
- When using allocators that themselves take child/backing allocators, each level of allocator will run redundant `@memset()` on the same memory region
- Allocator implementation author is responsible for ensuring memory undefining, possible to forget

[...]

PR #8756, opened 2021-05-12
Let GeneralPurposeAllocator retain metadata to report more double frees

mattbork:
Currently `GeneralPurposeAllocator` can only properly report double frees for small allocations from active buckets (buckets that weren't emptied at any point). Double frees for other small allocations and for all large allocations fall through to the "Invalid free" panic in `resizeLarge` (assuming `config.safety` is true). This will show a stack trace for the current free, but it can't show anything for the allocation or the first free. It also can't be recovered from, unlike properly reported double frees, and this may make debugging more onerous.

I've added a new flag, `config.retain_metadata`, that makes `GeneralPurposeAllocator` retain all empty buckets in a separate list and never remove entries from the `large_allocations` hash map (only mark them as freed). This way any double free of memory actually allocated from this allocator will always be able to find a set of stack traces to report. The extra storage for the empty bucket list in `GeneralPurposeAllocator` and the second stack trace and other bookkeeping in `LargeAlloc` is spent only when `retain_metadata` is set. Further, the performance of small allocations is not affected at all. However, because entries are never removed from `large_allocations`, lookup is likely to be slower as the map gets bigger. Leaving entries in the hash map was the simplest option and requires no extra allocations be done when freeing large objects, but maintaining a second hash map for dead large allocations is another option if the effect on performance is too high.

Another benefit of retaining metadata is that double frees can eventually be caught without having to use `config.never_unmap` and leak memory (all retained metadata is easily freed when the allocator is freed). For now, until https://github.com/ziglang/zig/issues/4298 is finished, `retain_metadata` should be used with `never_unmap` or else memsets in the allocator interface will cause segfaults. But in either case, because we're retaining metadata, the memory leaked by `never_unmap` can actually now be freed when the metadata is freed because we know exactly what memory was left deliberately unfreed. This should allow checking for double frees during only portions of a long running program with ballooning memory use. To that end, I added a public function `flushRetainedMetadata` that will free all retained metadata and, if `never_unmap` is set, recover the leaked memory as well. It is present only when `retain_metadata` is true.

I'd like to add tests for actually detecting and recovering from double frees but because the stack traces are logged immediately to stderr, I don't know if there's any way to keep a test from being marked as failing. Instead, I have a rather invasive test that just checks that metadata is kept and marked correctly after a free, and that all memory leaked by `never_unmap` is ultimately freed. For general stress testing, I also ran all of test-std with the `std.testing.allocator` modified to set `retain_metadata` (with and then without `never_unmap` set) and test_runner.zig modified to use a second `GeneralPurposeAllocator` as the backing_allocator for the testing allocator in order catch any leaks of metadata.

Since this is a large addition to somewhat intricate code that is also a basic element of the standard library, I'm happy to break down my changes in more detail if helpful.

### DIFF
--- a/lib/std/heap/general_purpose_allocator.zig
+++ b/lib/std/heap/general_purpose_allocator.zig
@@ -160,8 +160,15 @@ pub const Config = struct {
 
     /// This is a temporary debugging trick you can use to turn segfaults into more helpful
     /// logged error messages with stack trace details. The downside is that every allocation
-    /// will be leaked!
+    /// will be leaked, unless used with retain_metadata!
     never_unmap: bool = false,
+
+    /// This is a temporary debugging aid that retains metadata about allocations indefinitely.
+    /// This allows a greater range of double frees to be reported. All metadata is freed when
+    /// deinit is called. When used with never_unmap, deliberately leaked memory is also freed
+    /// during deinit. Currently should be used with never_unmap to avoid segfaults.
+    /// TODO https://github.com/ziglang/zig/issues/4298 will allow use without never_unmap
+    retain_metadata: bool = false,
 
     /// Enables emitting info messages with the size and address of every allocation.
     verbose_log: bool = false,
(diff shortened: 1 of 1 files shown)


andrewrk:
Oops, thought this was an issue, didn't realize you already implemented it. Very cool, will check this out later today :)

Vexu (approved):

PR #8757, opened 2021-05-12
SPIR-V: Features + some other stuff

Snektron:
This PR adds an initial list of SPIR-V target features, as well as some miscellaneous small improvements:
- SPIR-V targets are generated by a tool from the [SPIR-V extension registry](https://github.com/KhronosGroup/SPIRV-Registry) and [SPIR-V spec definition](https://github.com/KhronosGroup/SPIRV-Headers/) repositories.
There are 3 different types of features: Versions (SPIR-V revisions), capabilities (which when enabled add more instructions/operands) and extensions (which when enabled may add more capabilities and overal semenatics).
Unfortunately, there were some problems with this:
  * SPIR-V capabilities may depend on either a multiple of extensions, which cannot really be modeled as of yet by `std.Target`
  * SPIR-V extensions also may depend on versions/extensions, but this is not available in a machine-readable format.
  * Instructions/operands/extensions may be deprecated, and that also cannot be modeled yet.
 
  **Note**: The SPIR-V grammar definitions also contain definitions for extensions which aren't properly registered yet (these are drafts, mostly by [intel](https://github.com/intel/llvm/tree/sycl/sycl/doc/extensions/SPIRV). Note that they even abandoned some draft extensions (like SPV_INTEL_float_controls2), yet these extension definitions still remain in the grammar specification). In the future, care must be taken that only items provided by registered extensions (those in SPIRV-Registry) are used.
- ~The SPIR-V spec generation tool was updated to the (relatively) new render API. It probably also could do without though.~ Removed parsing/re-rendering in this tool, now it just emits in the right format. I also did some general maintainance on that, such as adding a disclaimer that the file is generated.
- The SPIR-V spec bindings were re-generated.
- The default output format of SPIR-V was set to spv.
- A setup was made for actually generating instructions.
- As per stage2 meeting discussion a while back, we decided to not implement incremental compilation for this target. All compilation will simply happen in flushModule, and so i also updated the current link/codegen setup to account for that.

This turned out to be a mix of a few somewhat unrelated things. I hope that is fine, otherwise i will split the feature part out into a dedicated PR and do the misc stuff another time.


andrewrk:
> This turned out to be a mix of a few somewhat unrelated things. I hope that is fine, otherwise i will split the feature part out into a dedicated PR and do the misc stuff another time.

Totally fine when you're spearheading a big new project. :+1:

src/link/SpirV.zig
```diff
@@ -30,16 +35,15 @@ const trace = @import("../tracy.zig").trace;
 const build_options = @import("build_options");
 const spec = @import("../codegen/spirv/spec.zig");
 
+// TODO: Should this struct be used at all rather than just a hashmap of aux data for every decl?
```
andrewrk, line 38: Noted. I have a plan to clean up `FnData` in the near future, feel free to just leave this messy and I'll get to it.

src/link/SpirV.zig
```diff
@@ -139,55 +118,69 @@ pub fn flushModule(self: *SpirV, comp: *Compilation) !void {
     const module = self.base.options.module.?;
     const target = comp.getTarget();
 
+    var spirv_module = codegen.SPIRVModule.init(target, self.base.allocator);
+    defer spirv_module.deinit();
+
+    // Allocate an ID for every declaration before generating code,
+    // so that we can access them before processing them.
+    // TODO: We're allocating an ID unconditionally now, are there
+    // declarations which don't generate a result?
```
andrewrk, line 127: I believe this question is answered by asking `Type.hasCodeGenBits()` for the Type of the Decl. The frontend code does not call `updateDecl` when false, but since you're iterating over `module.decl_table` manually, you'll have to do this check.

Note also that in my branch #8554, the global decl_table is gone, and I modified the C and SPIR-V backends to keep track of a set of decls in response to updateDecl and freeDecl, like this:

```diff
diff --git a/src/link/SpirV.zig b/src/link/SpirV.zig
index 7a35752f6..0a7fec941 100644
--- a/src/link/SpirV.zig
+++ b/src/link/SpirV.zig
@@ -37,9 +37,14 @@ pub const FnData = struct {
 
 base: link.File,
 
-// TODO: Does this file need to support multiple independent modules?
+/// TODO: Does this file need to support multiple independent modules?
 spirv_module: codegen.SPIRVModule,
 
+/// This linker backend does not try to incrementally link output SPIR-V code.
+/// Instead, it tracks all declarations in this table, and iterates over it
+/// in the flush function.
+decl_table: std.AutoArrayHashMapUnmanaged(*Module.Decl, void) = .{},
+
 pub fn createEmpty(gpa: *Allocator, options: link.Options) !*SpirV {
     const spirv = try gpa.create(SpirV);
     spirv.* = .{
@@ -88,6 +93,7 @@ pub fn openPath(allocator: *Allocator, sub_path: []const u8, options: link.Optio
 }
 
 pub fn deinit(self: *SpirV) void {
+    self.decl_table.deinit(self.base.allocator);
     self.spirv_module.deinit();
 }
 
@@ -95,6 +101,9 @@ pub fn updateDecl(self: *SpirV, module: *Module, decl: *Module.Decl) !void {
     const tracy = trace(@src());
     defer tracy.end();
 
+    // Keep track of all decls so we can iterate over them on flush().
+    _ = try self.decl_table.getOrPut(self.base.allocator, decl);
+
     const fn_data = &decl.fn_link.spirv;
     if (fn_data.id == null) {
         fn_data.id = self.spirv_module.allocId();
@@ -118,6 +127,7 @@ pub fn updateDeclExports(
 ) !void {}
 
 pub fn freeDecl(self: *SpirV, decl: *Module.Decl) void {
+    self.decl_table.removeAssertDiscard(decl);
     var fn_data = decl.fn_link.spirv;
     fn_data.code.deinit(self.base.allocator);
     if (fn_data.id) |id| self.spirv_module.freeId(id);
@@ -164,19 +174,15 @@ pub fn flushModule(self: *SpirV, comp: *Compilation) !void {
     defer all_buffers.deinit();
 
     // Pre-allocate enough for the binary info + all functions
-    try all_buffers.ensureCapacity(module.decl_table.count() + 1);
+    try all_buffers.ensureCapacity(self.decl_table.count() + 1);
 
     all_buffers.appendAssumeCapacity(wordsToIovConst(binary.items));
 
-    for (module.decl_table.items()) |entry| {
-        const decl = entry.value;
-        switch (decl.typed_value) {
-            .most_recent => |tvm| {
-                const fn_data = &decl.fn_link.spirv;
-                all_buffers.appendAssumeCapacity(wordsToIovConst(fn_data.code.items));
-            },
-            .never_succeeded => continue,
-        }
+    for (self.decl_table.items()) |entry| {
+        const decl = entry.key;
+        if (!decl.has_tv) continue;
+        const fn_data = &decl.fn_link.spirv;
+        all_buffers.appendAssumeCapacity(wordsToIovConst(fn_data.code.items));
     }
 
     var file_size: u64 = 0;
```

so one of us is going to have to deal with this conflict :)

andrewrk (approved):
This is really high quality work.

src/link/SpirV.zig
Snektron: I previously had a look at your branch to see the difference. The change didn't look like much work, so if you think your branch is soon ready for merging i'm also with waiting a bit and then fix the conflict before merging this PR.

andrewrk: I wouldn't mind resolving the conflict either - I think this can be merged as soon as the CI is green (except lavahut which is broken at the moment)

PR #8760, opened 2021-05-13
Add experimental Darling support for cross testing macOS

kubkon:
[Darling](https://www.darlinghq.org/) is a translation layer of macOS syscalls to Linux. This basically allows you to run native macOS binaries (MachO binaries) on your Linux box.

This commit introduces an experimental binding of Darling as means for cross testing MachO binaries and macOS specific tests on your Linux box without the need to grab an actual copy of macOS. Of course, since Darling itself is experimental, so is this binding, and currently its functionality is limited to executing binaries compiled to the same architecture as the host, and in particular, mainly `x86_64`.

In more detail, in terms of code additions:
* for cross testing stage2 tests, we use `darling shell` command since the path to the tested binary is relative to cwd
* for the `zig test` command, we simply use `darling` since the path to the binary is absolute

Testing works great except for `test-std` (yep, stage2 tests of `x86_64` macOS run and pass on my Linux box) which has several failures and hangups due to missing syscalls implementations, etc. in Darling.

Darling can be enabled just like QEMU or Wasmtime, e.g., for testing stage2:

```
zig build test-stage2 -Denable-darling
```


andrewrk (approved):
Very cool!

PR #8764, opened 2021-05-13
Fix duplicate_field_behavior UseFirst for parse in json.zig

mattbork:
Currently, calling `parse` with the option `duplicate_field_behavior` set to `UseFirst` does not work. When a duplicate field is found while parsing to a struct, if `UseFirst` is set, then `parseInternal` simply breaks from the loop over the struct's fields. This leaves `found` equal to false, so an `UnknownField` error is immediately returned. Setting `found` to true in the `UseFirst` case is not enough either because only an object key has been parsed and not the value. Further parsing would eventually return an `UnexpectedToken` error. Instead, the value must also be parsed and then immediately freed. Ideally, the value could be skipped over without incurring any allocations but this would require a new function that essentially duplicates all the logic of `parseInternal`. (Such a function could however be used to satisfy the TODO in `parsesTo` about not requiring an allocator.)

I also modified the duplicate field test case to check that `UseFirst` works correctly.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1573,11 +1573,16 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options: 
                                     //     .UseLast => {},
                                     // }
                                     if (options.duplicate_field_behavior == .UseFirst) {
-                                        // do nothing, check of fields_seen[i] below will parse value without overwriting field
+                                        // unconditonally ignore value. for comptime fields, this skips check against default_value
+                                        parseFree(field.field_type, try parse(field.field_type, tokens, options), options);
+                                        found = true;
+                                        break;
                                     } else if (options.duplicate_field_behavior == .Error) {
                                         return error.DuplicateJSONField;
                                     } else if (options.duplicate_field_behavior == .UseLast) {
-                                        parseFree(field.field_type, @field(r, field.name), options);
+                                        if (!field.is_comptime) {
+                                            parseFree(field.field_type, @field(r, field.name), options);
+                                        }
                                         fields_seen[i] = false;
                                     }
                                 }
@@ -1586,11 +1591,7 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options: 
                                         return error.UnexpectedValue;
                                     }
                                 } else {
-                                    if (fields_seen[i]) {
-                                        parseFree(field.field_type, try parse(field.field_type, tokens, options), options);
-                                    } else {
-                                        @field(r, field.name) = try parse(field.field_type, tokens, options);
-                                    }
+                                    @field(r, field.name) = try parse(field.field_type, tokens, options);
                                 }
                                 fields_seen[i] = true;
                                 found = true;
@@ -1735,7 +1736,9 @@ pub fn parseFree(comptime T: type, value: T, options: ParseOptions) void {
         },
         .Struct => |structInfo| {
             inline for (structInfo.fields) |field| {
-                parseFree(field.field_type, @field(value, field.name), options);
+                if (!field.is_comptime) {
+                    parseFree(field.field_type, @field(value, field.name), options);
+                }
             }
         },
         .Array => |arrayInfo| {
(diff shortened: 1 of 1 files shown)


lib/std/json.zig
```diff
@@ -1586,7 +1586,11 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options:
                                         return error.UnexpectedValue;
                                     }
                                 } else {
-                                    @field(r, field.name) = try parse(field.field_type, tokens, options);
+                                    if (fields_seen[i]) {
```
daurnimator, line 1589: I'd add a check here for `options.duplicate_field_behavior == .UseFirst` here that would make the branch predictor happy.

I also was hoping that we could eventually make `duplicate_field_behavior` comptime-known.

lib/std/json.zig
```diff
@@ -1573,7 +1573,7 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options:
                                     //     .UseLast => {},
                                     // }
                                     if (options.duplicate_field_behavior == .UseFirst) {
-                                        break;
+                                        // do nothing, check of fields_seen[i] below will parse value without overwriting field
```
daurnimator, line 1576: don't you also need a check in the `field.is_comptime` branch?

daurnimator (changes requested):

lib/std/json.zig
mattbork: I'm assuming that for comptime fields, the parsed value must always equal the field's `default_value`, even when the key is a duplicate that will be thrown out. For example, if parsing to a `struct { comptime x: u32 = 99 }` with `UseFirst` then parsing `{ "x": 99, "x": 99 }` is fine but `{ "x": 99, "x": 100 }` should return `UnexpectedValue` even though the second value is ostensibly being ignored. Then a comptime field should have the same behavior for the first field and for duplicate fields when either `UseFirst` or `UseLast` is set.

Writing this, I have noticed a nasty problem with `parseFree`: if a struct has a comptime field that is or contains a pointer of any kind, then `parseFree` will try to deallocate that memory even though the comptime field holds only its default value and not anything that was allocated. And this problem also extends to the `UseLast` case I recently adjusted.

mattbork:
While looking at the review, I noticed a nasty problem with `parseFree`: if a struct has a comptime field that is or contains a pointer of any kind, then `parseFree` will try to deallocate that memory even though the comptime field holds only its default value and not anything that was allocated.  For example,

```
const opts = json.ParseOptions{ 
    .allocator = std.testing.allocator,
    .duplicate_field_behavior = .Error,
};

const T = struct {
    comptime message: []const u8 = "hello"
};

const str = "{ \"message\": \"hello\" }";
const res = try json.parse(T, &json.TokenStream.init(str), opts);
json.parseFree(T, res, opts);
```

This will segfault because parseFree tries to free the `message` field of `res`, even though it holds only its default value and not a dynamic allocation. I'm closing this PR until I fix this and also sort out the logic in `parseInternal` pertaining to comptime fields.

lib/std/json.zig
daurnimator: > I'm assuming that for comptime fields, the parsed value must always equal the field's `default_value`, even when the key is a duplicate that will be thrown out. For example, if parsing to a `struct { comptime x: u32 = 99 }` with `UseFirst` then parsing `{ "x": 99, "x": 99 }` is fine but `{ "x": 99, "x": 100 }` should return `UnexpectedValue` even though the second value is ostensibly being ignored. Then a comptime field should have the same behavior for the first field and for duplicate fields when either `UseFirst` or `UseLast` is set.

That behaviour has the downside that changing a field from runtime to comptime has externally visible effects; which would be nice to avoid.
I think with `.UseFirst`,  `{ "x": 99, "x": 100 }`  should ignore the `100` unconditionally.

mattbork: That's a good point, and I think a compelling case that that's the correct interpretation of `UseFirst`. I'll change it.

lib/std/json.zig
```diff
@@ -1573,11 +1573,16 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options:
                                     //     .UseLast => {},
                                     // }
                                     if (options.duplicate_field_behavior == .UseFirst) {
+                                        // unconditonally ignore value. for comptime fields, this skips check against default_value
+                                        parseFree(field.field_type, try parse(field.field_type, tokens, options), options);
+                                        found = true;
```
daurnimator, line 1578: found would already be `true` to get here

mattbork: `fields_seen[i]` is true but `found` is only set to true immediately before breaking out of the inline for loop over the struct fields.

mattbork:
`.UseFirst` now unconditionally ignores the values of duplicate keys, so the default value of a comptime field is only checked for the first instance of a key. I expanded the duplicate fields test case to cover comptime fields as well. Importantly, I've also fixed the problem with `parseFree` trying to free comptime fields and a similar oversight I had introduced in the `.UseLast` case for duplicate fields in `parseInternal`. I modified a test with comptime fields that are slices to ensure that parseFree won't try to free them.

daurnimator (approved):

PR #8775, opened 2021-05-14
stage2: Force Clang to use LLVM's assembler for SPARC targets

LemonBoy:
This change allows cross-compiling C code for SPARC, if you hit any
error with the internal assembler please open a ticket.

cc @koachan


andrewrk:
Nice, can we pass this unconditionally? The intent here is to never shell out to system dependencies.

LemonBoy:
Sure, at the time of writing the only special-cased architecture is SPARC anyway.

andrewrk:
clarified the logic for future contributors in 8f35c60b39dad1ecb1609f6105f1e41a52df87ae

PR #8776, opened 2021-05-14
fix & add some Win32 window message constants

notviri:
- Fixes this constant:
  https://github.com/ziglang/zig/blob/612ad779bb2df8e7e15ab20dd3df3a2900ff82f0/lib/std/os/windows/user32.zig#L41
  Which is entirely wrong, `WM_NCDESTROY` is sent when the non-client area is being destroyed while `WM_DESTROY` is being sent while the client area is being destroyed (comes first). This is quite dangerous too because `WM_NCDESTROY` is commonly known to be the last message sent before a window handle is invalidated, and thus it's common to invalidate state there too despite it being actually just `WM_DESTROY` where the state is still meant to be valid afterwards.
- Since some other window message constants were missing, adds the list of known window message constants [compiled by the WINE developers](https://wiki.winehq.org/List_Of_Windows_Messages) parsed into a big list of Zig constants.

*\* There's no official list of window messages on MSDN, unfortunately.*


notviri:
The first commit removed like 5 weird undocumented constants that I can only find proof of existing in online copies of header files (not in my copy of the Windows SDK it seems) but this PR wasn't intended to remove things so I put it back.

lib/std/os/windows/user32.zig
```diff
@@ -113,16 +245,802 @@ pub const WM_RBUTTONDBLCLK = 0x0206;
 pub const WM_MBUTTONDOWN = 0x0207;
 pub const WM_MBUTTONUP = 0x0208;
 pub const WM_MBUTTONDBLCLK = 0x0209;
+pub const WM_MOUSELAST = 0x0209;
```
daurnimator, line 248: We shouldn't include `*LAST` and `*FIRST` constants IMO

notviri: They can still be used fine with for example `GetMessageW` with the min and max filter arguments provided as `xxx_FIRST` to `xxx_LAST`

daurnimator: The *LAST ones move on most updates; is it really useful to have the wrong values?

e.g. this PR changes `WM_MOUSELAST` from `0x020A` to `0x0209`, however in windows its now up to `0x020E`

notviri: I would say that you can't handle what you didn't know existed at the time, but fair point. I've removed those constants.

marler8997:
It's unclear what subset of the win32 bindings Zig will include in its std library.  However, I believe it will be some sort of "subset", possibly just enough to support the platform abstraction libraries.

If you're looking for more than that, it will be in a 3rd-party library.  You can check out my `zigwin32` project here which is intended to contain bindings for the full win32 api based on the new official win32metadata project.

https://github.com/marlersoft/zigwin32

notviri:
> It's unclear what subset of the win32 bindings Zig will include in its std library. However, I believe it will be some sort of "subset", possibly just enough to support the platform abstraction libraries.

Yeah, I definitely found it strange that there's this function pointer dispatch abstraction going on in `user32.zig`... surely though the associated constants should be there, no? Neat project though.

andrewrk:
I can confirm this change is correct:

```
[nix-shell:~/Downloads/zig/build]$ ./zig translate-c test.h -target x86_64-windows-gnu -lc | grep WM_NCDESTROY
pub const WM_NCDESTROY = @as(c_int, 0x0082);
```

Thanks for the fix. Will merge this since it is a fix, but keep in mind that user32.zig will probably be entirely removed at some point.

See #4426 for more details.

notviri:
> ... keep in mind that user32.zig will probably be entirely removed at some point.

That's alright, but does this mean I should close #8801 early? I'm not entirely clear on what "minimal surface area" means.

Issue #8777, opened 2021-05-14
zig test: std.testing.expectEqual error message for the expected value is hard to read


ZapAnton:
The error message about the expected value of the `expectEqual` is printed too close to the name of the failed test.
For example the following test
```zig
const testing = @import("std").testing;

test "failing test" {
    const x = 5;
    try testing.expectEqual(x, 2);
}
```
produces the following line:
```text
expected 5, found 2test "failing test"... FAIL (TestExpectedEqual)
```
Maybe a single space could be used after the `expected 5, found 2`?

Zig version: `0.8.0-dev.2237+8eea5eddf`

andrewrk:
Thanks for pointing this out. There were some recent big changes to the testing system (#8686) and I believe previously it made sense for there to be no newline after the failure messages. But now I think all the `std.debug.print` calls in `std.testing` need newlines. Should be quite a contributor friendly issue if anyone wants to take it on :-)

PR #8778, opened 2021-05-14
std.testing: Added newline to the error messages

ZapAnton:
Added newline to every `std.debug.print` message.
Closes https://github.com/ziglang/zig/issues/8777


andrewrk:
Thank you! Apologies for the CI issues, hopefully will have that resolved within the next couple days.

Issue #3047, opened 2019-08-11
Add --build-id flag


daurnimator:
Related to https://github.com/ziglang/zig/pull/3014

zig commands such as `build-exe` should take a `--build-id` flag to specify the build id.

See https://fedoraproject.org/wiki/Releases/FeatureBuildId#Unique_build_ID

andrewrk:
The link is dead - can you expand on this?

nektro:
https://web.archive.org/web/20210621234120/https://fedoraproject.org/wiki/Releases/FeatureBuildId live link

Jan200101:
This doesn't quite work because there are multiple different build-id types that no one can agree on.

simply adding `--build-id` to the linker call will use the fast build-id, while it may be desired to use a different one (md5,sha1,uuid)

PR #8784, opened 2021-05-15
macho: fix DWARF in dSYM and sym naming more consistent

kubkon:
* Advance line and PC prior to ending sequence in debug line program for a fn_decl. This is equivalent to closing scope in the debugger and without it, the debugger will not map source-to-address info as a result will not print the source when breaking at a symbol.
* Fix debug aranges sentinels to be of the size as the actual tuple descriptor (assuming segment selector to be ommitted). In summary, the sentinels were 32bit 0s, whereas they ought to be 64bit 0s.
* Make naming of symbols in the binary more consistent by prefixing each symbol name with an underscore '_'.

---

With these changes, here's what we get in `lldb` when debugging a Zig program compiled with self-hosted:

```
$ zig-out/bin/zig build-exe hello.zig
$ ls
hello     hello.zig zig-cache
$ lldb hello
(lldb) target create "hello"
Current executable set to '/Users/kubkon/dev/zig/examples/stage2/hello' (arm64).
(lldb) b main
Breakpoint 1: where = hello`main at hello.zig:4, address = 0x0000000100001000
(lldb) r
Process 54300 launched: '/Users/kubkon/dev/zig/examples/stage2/hello' (arm64)
Process 54300 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x0000000100001000 hello`main at hello.zig:4
   1   	extern "c" fn write(usize, usize, usize) usize;
   2   	extern "c" fn exit(usize) noreturn;
   3
-> 4   	export fn main() noreturn {
   5   	    print();
   6
   7   	    exit(0);
Target 0: (hello) stopped.
(lldb)
```

Note that we successfully print the source file info *and* the binary itself is devoid of any old-fashioned stab symbols and DWARF info. Instead, we make use of the modern alternative on macOS called the dSYM bundle which is a mirror image of the binary's in-memory layout with relocated DWARF info attached. The beautiful thing about the dSYM bundle is that it doesn't have to reside side-by-side the binary since it is the matching UUID between the binary and dSYM that identifies the pairing, and `lldb` is smart enough to ask Spotlight service to locate the dSYM from the global macOS cache. The tl;dr here is though, Zig's self-hosted is supercharged as we not only provide an incremental (and traditional!) MachO linkers, but now also implement something akin to an in-built `dsymutil` for debugging symbols management.

### DIFF
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -362,8 +362,8 @@ pub fn openPath(allocator: *Allocator, sub_path: []const u8, options: link.Optio
 
     self.base.file = file;
 
-    // Create dSYM bundle.
     if (!options.strip and options.module != null) {
+        // Create dSYM bundle.
         const dir = options.module.?.zig_cache_artifact_directory;
         log.debug("creating {s}.dSYM bundle in {s}", .{ sub_path, dir.path });
 
@@ -1223,7 +1223,11 @@ pub fn updateDecl(self: *MachO, module: *Module, decl: *Module.Decl) !void {
             self.shrinkTextBlock(&decl.link.macho, code.len);
         }
         decl.link.macho.size = code.len;
-        symbol.n_strx = try self.updateString(symbol.n_strx, mem.spanZ(decl.name));
+
+        const new_name = try std.fmt.allocPrint(self.base.allocator, "_{s}", .{mem.spanZ(decl.name)});
+        defer self.base.allocator.free(new_name);
+
+        symbol.n_strx = try self.updateString(symbol.n_strx, new_name);
         symbol.n_type = macho.N_SECT;
         symbol.n_sect = @intCast(u8, self.text_section_index.?) + 1;
         symbol.n_desc = 0;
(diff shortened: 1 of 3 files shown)


daurnimator:
> The beautiful thing about the dSYM bundle is that it doesn't have to reside side-by-side the binary since it is the matching UUID between the binary and dSYM that identifies the pairing, and `lldb` is smart enough to ask Spotlight service to locate the dSYM from the global macOS cache

FYI to support that on other platforms we need build-id support (#3047)

kubkon:
> > The beautiful thing about the dSYM bundle is that it doesn't have to reside side-by-side the binary since it is the matching UUID between the binary and dSYM that identifies the pairing, and `lldb` is smart enough to ask Spotlight service to locate the dSYM from the global macOS cache
> 
> FYI to support that on other platforms we need build-id support (#3047)

I've heard somewhere that dSYM will make it into the DWARF standard at some point, which is all more reason to have build-id support on *all* platforms.

Issue #6597, opened 2020-10-07
Make C pointers more normal (non-optional and not coercible to/from integers)


tadeokondrak:
Motivation: I want to support gcc's `__attribute__((nonnull))` and clang's `_Nonnull` pointer qualifier in translate-c, but I think this is a good idea either way.

C pointers currently (taken from language reference):

> - Supports all the syntax of the other two pointer types.
> - Coerces to other pointer types, as well as Optional Pointers. When a C pointer is coerced to a non-optional pointer, safety-checked Undefined Behavior occurs if the address is 0.

You should still be able to index with `[]` and dereference with `.*`. `.?`/`try`/`orelse` come free with optional types.

> - Allows address 0. On non-freestanding targets, dereferencing address 0 is safety-checked Undefined Behavior. Optional C pointers introduce another bit to keep track of null, just like ?usize. Note that creating an optional C pointer is unnecessary as one can use normal Optional Pointers.

C pointers should be non-optional by default, and translate-c should translate most pointers as `?[*c]T`.

- This reduces the number of special cases in the compiler, syntax for optional pointers and c pointers have separate code paths
- This allows using `[*c]T` as a type in translate-c for `__attribute__((nonnull))`-annotated function parameters or return types, or 
clang's `_Nonnull` annotatation (https://github.com/ziglang/zig/issues/2457 but useful for C consumers too!)

> - Supports Type Coercion to and from integers.
> - Supports comparison with integers.

This should be removed completely.

Removing this is already [partially accepted](https://github.com/ziglang/zig/issues/2057), other than comptime-known 0.
From what I can gather, this was initially for translate-c before it became as smart as it is now.

I think C APIs that benefit from this implicit behaviour are pretty rare, and we shouldn't make the language more weird to accomodate them.
We [already don't change the language to accomodate using enums as bitflags](https://github.com/ziglang/zig/issues/2115#issuecomment-565763241), which are much more common.

> - Does not support Zig-only pointer attributes such as alignment. Use normal Pointers please!

No change here.

PR #8785, opened 2021-05-15
mem.sliceTo on C pointers should return an optional slice

daurnimator:
https://github.com/ziglang/zig/pull/8698/files#r632922210

Also makes the compile errors from `SliceTo` more consistent

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -654,6 +654,7 @@ fn SliceTo(comptime T: type, comptime end: meta.Elem(T)) type {
                                 new_ptr_info.sentinel = null;
                             }
                         }
+                        return @Type(std.builtin.TypeInfo{ .Pointer = new_ptr_info });
                     },
                     else => {},
                 },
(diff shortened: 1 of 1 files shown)


ifreund (changes requested):
I'm not 100% sure this makes sense with the current semantics of C pointers, if https://github.com/ziglang/zig/issues/6597 were accepted I would think otherwise.

If this is the way to go, I think we need to update `std.mem.span` to match this behavior, and you need to correct the doc comment for `std.mem.sliceTo()`.

daurnimator:
> If this is the way to go, I think we need to update `std.mem.span()` to match this behavior,

hmm. I think I'd only like to do that if #6597 is accepted.



>  you need to correct the doc comment for `std.mem.sliceTo()`.

done.

ifreund:
> hmm. I think I'd only like to do that if #6597 is accepted.

I think having consistent behavior between the two is pretty important, what's your reasoning here?

daurnimator:
> I think having consistent behavior between the two is pretty important, what's your reasoning here?

`span` looks like it shouldn't accept C pointers at all; and users should instead be using `sliceTo`.
Rather than change it, I'd rather remove the `.C` branch entirely.

andrewrk:
Please re-open against master branch with passing tests

PR #8789, opened 2021-05-15
stage2: wasm backend - Enums

Luukdegram:
This PR adds initial enum support for the wasm backend.

- Only enums based on tag indexes are currently tested as astgen doesn't allow for enums with explicit values yet. (tho, wasm codegen does implement it).
- Return statements now also add an explicit `return` opcode as those are required within blocks. Having them always be emitted
rather than relying on implicit returns makes the logic a lot easier for now.
- A basic form of bitcast was also implemented (it returns the `WValue` for its operand)
- `emitConstant()` signature was updated to allow for emitting constants by generating a type (i.e. using an enum's integer type rather than enum type).

### DIFF
--- a/src/codegen/wasm.zig
+++ b/src/codegen/wasm.zig
@@ -758,6 +758,7 @@ pub const Context = struct {
         // TODO: Implement tail calls
         const operand = self.resolveInst(inst.operand);
         try self.emitWValue(operand);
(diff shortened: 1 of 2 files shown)


Luukdegram:
Rebased after astgen branch

andrewrk:
Looks great!

PR #8790, opened 2021-05-15
std.build.InstallDir: make dupe() a public function

ellistg:
Super tiny change but required this for writing a custom step. Awesome project btw!

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -285,10 +285,12 @@ pub const Builder = struct {
         return run_step;
     }
 
+    /// Allocator.dupe without the need to handle out of memory.
     pub fn dupe(self: *Builder, bytes: []const u8) []u8 {
         return self.allocator.dupe(u8, bytes) catch unreachable;
     }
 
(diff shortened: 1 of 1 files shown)


jedisct1:
Since this will be public, could you take it as an opportunity to add a little `///` comment summarizing what the function does?

That would be nice!

ellistg:
@jedisct1 there are many public dupe functions without a description as to what they do, am very happy to write a description but as all dupe functions do the same thing but to different types seems to me that this is no different. May just be that comments haven't been written yet for other dupe functions though :)

jedisct1:
Indeed, many functions, especially the oldest ones, are still missing a description. This is more of an implicit "TODO: add description" task than a set of functions meant to be left undocumented. 

If only for ZLS, it would be useful to have them for all public functions, and it looks like efforts have been made towards this goal for most newly introduced APIs.

So, a little comment on top of your brand new function would be nice and help get closer to the goal of having everything (even succinctly) documented. And if you feel like doing the same for other existing `dup*` functions, that would be fantastic!

PR #8791, opened 2021-05-15
std.json: Adds jsonParse to override parsing behavior

Bxil:
`std.json` currently has the ability to override serialization behavior by implementing `jsonStringify`. This is pretty useless, as the other way is impossible. As such I added `jsonParse` to override deserialization.
I also unified all `jsonStringify`s into a single one at the beginning of the `stringify` function.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1708,6 +1708,9 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options: 
 }
 
 pub fn parse(comptime T: type, tokens: *TokenStream, options: ParseOptions) !T {
+    if (comptime std.meta.trait.hasFn("jsonParse")(T)) {
+        return T.jsonParse(tokens, options);
+    }
     const token = (try tokens.next()) orelse return error.UnexpectedEndOfJson;
     return parseInternal(T, token, tokens, options);
 }
@@ -2014,12 +2017,12 @@ test "parse into struct with duplicate field" {
     const ballast = try testing.allocator.alloc(u64, 1);
     defer testing.allocator.free(ballast);
 
-    const options_first = ParseOptions{ 
+    const options_first = ParseOptions{
         .allocator = testing.allocator,
-        .duplicate_field_behavior = .UseFirst
+        .duplicate_field_behavior = .UseFirst,
     };
 
-    const options_last = ParseOptions{ 
+    const options_last = ParseOptions{
         .allocator = testing.allocator,
         .duplicate_field_behavior = .UseLast,
     };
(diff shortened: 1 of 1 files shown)


lib/std/json.zig
```diff
@@ -2648,17 +2673,9 @@ pub fn stringify(
             }
```
Mouvedia, line 2673: Why is l2672 different from l2666?

Bxil: I don't know, but this is completely unrelated to this PR.

lib/std/json.zig
```diff
@@ -2648,17 +2673,9 @@ pub fn stringify(
...
         },
         .Enum => {
-            if (comptime std.meta.trait.hasFn("jsonStringify")(T)) {
```
daurnimator, line 2651: only enum/union/struct can have user traits; it makes more sense to keep them here?

lib/std/json.zig
```diff
@@ -1708,6 +1708,9 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options:
 }
 
 pub fn parse(comptime T: type, tokens: *TokenStream, options: ParseOptions) !T {
+    if (comptime std.meta.trait.hasFn("jsonParse")(T)) {
```
daurnimator, line 1711: This will only work at the top level: the internal recursion uses `parseInternal`.

daurnimator (changes requested):
implementing this was always planned, but its tricky because you need to handle the token lookahead. Trying to do that in a way that isn't a foot-gun for people implementing the trait was too difficult IMO.

I was planning on an eventual overhaul of the token parser, so I had delayed implementing this feature (as it would break everyone's traits).

lib/std/json.zig
Bxil: This is exactly why I found it reasonable to move it outside: it stops this check from getting repeated in the code without altering behavior. Since I applied the same reasoning to `jsonParse` and `jsonStringify` is effectively its pair I thought it was a good idea to have the same logic applied to it.

lib/std/json.zig
Bxil: Whoops, for all examples I tried I somehow never had one where the root did not have `jsonParse` implemented. What's stopping us from simply keeping a boolean on `TokenStream` to know if we need the lookahead or not?

Bxil: I have investigated, and the token parser indeed badly needs an overhaul. I couldn't find a simple enough way to get this properly done without a refactor, but I can't do that right now, so for the time being I will close this PR.

PR #8796, opened 2021-05-16
SPIR-V: Codegen basis

Snektron:
This PR adds a bunch of basic SPIR-V code- and constant generation. I figured to make this PR before it gets too large, and now seems like a decent size. List of additions:
- Codegen was restructured a bit in general from last time. There is now a DeclGen struct, which (similar to the other backends) is used to generate declarations for a struct. It contains a bunch of types which must be persistent over different decls, and to save some extra constructing of that struct i just reset the appropriate fields every time a new decl must be generated. 
- Basic generation was implemented for:
  * Types: Void, bool, int (only non composite ones), float, and function prototypes.
  * Function setup/teardown.
  * Binary operations
  * Unary operations
  * ret/retvoid
  * Unreachable
  * arg
 
 I left a few comments in the code for guarantees i wasn't sure about, it would be beneficial if someone reviewed those. In particular, what the expected types are for arguments of binary instructions (are they always equal?) and whether unreachable/return instructions always terminate a block.

### DIFF
--- a/src/codegen/spirv.zig
+++ b/src/codegen/spirv.zig
@@ -57,6 +57,7 @@ pub const DeclGen = struct {
     module: *Module,
     spv: *SPIRVModule,
 
+    args: std.ArrayList(u32),
     types: TypeMap,
 
     decl: *Decl,
@@ -213,7 +214,17 @@ pub const DeclGen = struct {
                 prototype_id,
             });
 
-            // TODO: Parameters
+            const params = tv.ty.fnParamLen();
+            var i: usize = 0;
+
+            try self.args.ensureCapacity(params);
+            while (i < params) : (i += 1) {
+                const param_type_id = self.types.get(tv.ty.fnParamType(i)).?;
+                const arg_result_id = self.spv.allocResultId();
+                try writeInstruction(&self.spv.fn_decls, .OpFunctionParameter, &[_]u32{ param_type_id, arg_result_id });
+                self.args.appendAssumeCapacity(arg_result_id);
+            }
+
             // TODO: Body
 
             try writeInstruction(&self.spv.fn_decls, .OpFunctionEnd, &[_]u32{});
--- a/src/codegen/spirv.zig
+++ b/src/codegen/spirv.zig
@@ -6,6 +6,7 @@ const spec = @import("spirv/spec.zig");
 const Module = @import("../Module.zig");
 const Decl = Module.Decl;
 const Type = @import("../type.zig").Type;
+const LazySrcLoc = Module.LazySrcLoc;
 
 pub const TypeMap = std.HashMap(Type, u32, Type.hash, Type.eql, std.hash_map.default_max_load_percentage);
 
@@ -15,30 +16,24 @@ pub fn writeInstruction(code: *std.ArrayList(u32), instr: spec.Opcode, args: []c
     try code.appendSlice(args);
 }
 
+/// This structure represents a SPIR-V binary module being compiled, and keeps track of relevant information
+/// such as code for the different logical sections, and the next result-id.
 pub const SPIRVModule = struct {
-    next_result_id: u32 = 0,
-
-    target: std.Target,
-
-    types: TypeMap,
-
+    next_result_id: u32,
     types_and_globals: std.ArrayList(u32),
     fn_decls: std.ArrayList(u32),
 
-    pub fn init(target: std.Target, allocator: *Allocator) SPIRVModule {
+    pub fn init(allocator: *Allocator) SPIRVModule {
         return .{
-            .target = target,
-            .types = TypeMap.init(allocator),
+            .next_result_id = 0,
             .types_and_globals = std.ArrayList(u32).init(allocator),
             .fn_decls = std.ArrayList(u32).init(allocator),
         };
     }
 
     pub fn deinit(self: *SPIRVModule) void {
-        self.fn_decls.deinit();
         self.types_and_globals.deinit();
-        self.types.deinit();
-        self.* = undefined;
+        self.fn_decls.deinit();
     }
 
     pub fn allocResultId(self: *SPIRVModule) u32 {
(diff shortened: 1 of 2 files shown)


src/codegen/spirv.zig
```diff
@@ -49,31 +55,326 @@ pub const SPIRVModule = struct {
...
+    /// that size. In this case, multiple elements of the largest type should be used.
+    /// The backing type will be chosen as the smallest supported integer larger or equal to it in number of bits.
+    /// The result is valid to be used with OpTypeInt.
+    /// TODO: The extension SPV_INTEL_arbitrary_precision_integers allows any integer size (at least up to 32 bits).
+    /// TODO: This probably needs an ABI-version as well (especially in combination with SPV_INTEL_arbitrary_precision_integers).
+    /// TODO: Should the result of this function be cached?
+    fn backingIntBits(self: *DeclGen, bits: u16) ?u16 {
+        const target = self.module.getTarget();
+
+        // TODO: Figure out what to do with u0/i0.
```
andrewrk, line 151: `hasCodeGenBits()`for u0/i0 will be `false` and so the SPIR-V backend will not be asked to lower a 0 bit integer. The assertion is correct.

src/codegen/spirv.zig
```diff
@@ -49,31 +55,326 @@ pub const SPIRVModule = struct {
...
+    /// SPIR-V requires enabling specific integer sizes through capabilities, and so if they are not enabled, we need
+    /// to emulate them in other instructions/types. This function returns, given an integer bit width (signed or unsigned, sign
+    /// included), the width of the underlying type which represents it, given the enabled features for the current target.
+    /// If the result is `null`, the largest type the target platform supports natively is not able to perform computations using
+    /// that size. In this case, multiple elements of the largest type should be used.
+    /// The backing type will be chosen as the smallest supported integer larger or equal to it in number of bits.
+    /// The result is valid to be used with OpTypeInt.
+    /// TODO: The extension SPV_INTEL_arbitrary_precision_integers allows any integer size (at least up to 32 bits).
+    /// TODO: This probably needs an ABI-version as well (especially in combination with SPV_INTEL_arbitrary_precision_integers).
+    /// TODO: Should the result of this function be cached?
```
andrewrk, line 147: Let's start with the answer being "no", as you have done here, and we can do performance measurements later to find out the answer to this question.

src/codegen/spirv.zig
```diff
@@ -84,23 +385,197 @@ pub const SPIRVModule = struct {
...
+            .dbg_stmt => null,
+            .ret => self.genRet(inst.castTag(.ret).?),
+            .retvoid => self.genRetVoid(),
+            .unreach => self.genUnreach(),
+            else => self.fail(.{.node_offset = 0}, "TODO: SPIR-V backend: implement inst {}", .{inst.tag}),
+        };
+    }
+
+    fn genBinOp(self: *DeclGen, inst: *Inst.BinOp) !u32 {
+        // TODO: Will lhs and rhs have the same type?
```
andrewrk, line 470: Good question which I am happy to leave as a TODO comment until AIR tags are properly documented with doc comments. (Answer is yes except possibly for shifts which I want to double check)

src/codegen/spirv.zig
```diff
@@ -84,23 +385,197 @@ pub const SPIRVModule = struct {
...
+        const lhs_id = try self.resolve(inst.lhs);
+        const rhs_id = try self.resolve(inst.rhs);
+
+        const result_id = self.spv.allocResultId();
+        const result_type_id = try self.getOrGenType(inst.base.ty);
+
+        // TODO: Is the result the same as the argument types?
```
andrewrk, line 477: you're calling this function for comparisons (result: bool) as well as arithmetic (result type same as arguments).

src/codegen/spirv.zig
```diff
@@ -84,23 +385,197 @@ pub const SPIRVModule = struct {
...
+
+        if (info.class == .composite_integer)
+            return self.fail(.{.node_offset = 0}, "TODO: SPIR-V backend: binary operations for composite integers", .{});
+
+        const is_bool = info.class == .bool;
+        const is_float = info.class == .float;
+        const is_signed = info.signedness == .signed;
+        // **Note**: All these operations must be valid for vectors of floats, integers and bools as well!
+        // For floating points, we generally want ordered operations (which return false if either operand is nan).
+        const opcode = switch (inst.base.tag) {
```
andrewrk, line 496: I suggest to do this switch only once (above in `genInst`). I don't think the attempt to avoid code duplication by making this `genBinOp` abstraction is doing you any favors here.

src/codegen/spirv.zig
```diff
@@ -84,23 +385,197 @@ pub const SPIRVModule = struct {
...
+
+    fn genUnOp(self: *DeclGen, inst: *Inst.UnOp) !u32 {
+        const operand_id = try self.resolve(inst.operand);
+
+        const result_id = self.spv.allocResultId();
+        const result_type_id = try self.getOrGenType(inst.base.ty);
+
+        const info = try self.arithmeticTypeInfo(inst.operand.ty);
+
+        const opcode = switch (inst.base.tag) {
```
andrewrk, line 547: same deal - I think a flatter branching on AIR tags will result in simpler logic, with fewer branches.

andrewrk (approved):

src/codegen/spirv.zig
Snektron: Yea, this was mostly just a reminder for future efforts. I will turn these into proper issues when there is a bit more progress.

src/codegen/spirv.zig
Snektron: Ah, shifts, i forgot about those. I'll leave it like this for now until shifts are there.

src/codegen/spirv.zig
Snektron: I think that like this it keeps all the relevant information quite neatly in the function that actually deals with binary operators. What would you otherwise suggest? I assume moving the `if` conditions up to the genInst, but in that case it would be rather a pain to get the arithmeticTypeInfo of one of the argument types.

PR #8797, opened 2021-05-16
add missing EBADF error code for openat

marler8997:
encountered this error code on my NixOS machine


kubkon (approved):
Nice, good catch!

Issue #8798, opened 2021-05-16
std.fs.File.Reader.Error (os.ReadError) does not include error names used by std.io.Reader


chr-1x:
Consider this straightforward function that tries to be specific about which errors it can return:

```zig
const std = @import("std");

pub fn main() !void {
    const file = try std.fs.cwd().openFile(
        "foo.txt",
        std.fs.File.OpenFlags{ .read = true },
    );
    const reader = file.reader();
    _ = try readFoo(reader);
}

fn readFoo(reader: std.fs.File.Reader) std.fs.File.Reader.Error!u32 {
    return reader.readIntNative(u32);
}
```

This function does not compile:

```
./repro_fs_file_error.zig:13:32: error: expected type 'std.os.ReadError!u32', found '@typeInfo(@typeInfo(@TypeOf(std.io.reader.Reader(std.fs.file.File,std.os.ReadError,std.fs.file.File.read).readIntNative)).Fn.return_type.?).ErrorUnion.error_set!u32'
    return reader.readIntNative(u32);
                               ^
./repro_fs_file_error.zig:13:32: note: error set '@typeInfo(@typeInfo(@TypeOf(std.io.reader.Reader(std.fs.file.File,std.os.ReadError,std.fs.file.File.read).readIntNative)).Fn.return_type.?).ErrorUnion.error_set' cannot cast into error set 'std.os.ReadError'
    return reader.readIntNative(u32);
                               ^
/home/khr/repos/zig/build/lib/zig/std/io/reader.zig:51:49: note: 'error.EndOfStream' not a member of destination error set
            if (amt_read < buf.len) return error.EndOfStream;
```

However, this does:

```zig
fn readFoo(reader: std.fs.File.Reader) (error{EndOfStream} || std.fs.File.Reader.Error)!u32 {
    return reader.readIntNative(u32);
}
```

This is because `os.ReadError` does not include `EndOfStream,` which the `std.io.Reader` API can return. This makes it unnecessarily difficult to be specific about errors returned by reading from `File` streams.

Because of #2971, it is impossible to write mutually recursive functions that take `std.fs.File.Reader`s without running into this (you are forced to be specific about at least one of your error unions):

```zig
var runtime_switch: bool = false;

fn readFoo(reader: std.fs.File.Reader) !u32 {
    if (runtime_switch) {
        return reader.readIntNative(u32);
    } else {
        return readBar(reader);
    }
}

fn readBar(reader: std.fs.File.Reader) !u32 {
[...]

PR #8799, opened 2021-05-16
Include EndOfStream in io.Reader.Error and annotate more errors

chr-1x:
An attempt at addressing #8798. I didn't try to annotate the allocating or enum-related Reader functions in this PR. Feel free to suggest different/better approaches, this was just the simplest thing I could think of that would improve the API slightly.

### DIFF
--- a/lib/std/io/reader.zig
+++ b/lib/std/io/reader.zig
@@ -19,7 +19,7 @@ pub fn Reader(
     comptime readFn: fn (context: Context, buffer: []u8) ReadError!usize,
 ) type {
     return struct {
-        pub const Error = ReadError;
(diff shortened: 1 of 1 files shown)


daurnimator:
`EndOfStream` is intentionally only returned from some functions in the reader interface, e.g. `readAll` cannot return `EndOfStream`.
Similarly, `readAllAlloc` can return `AllocationFailed`, but we don't want to extend the error set of all reader functions to make room for that.

Vexu:
As daurnimator said `EndOfStream` is not needed by most of the functions so I do not want to add it there without going through a proposal.

PR #8801, opened 2021-05-17
std.os.windows: implement <timeapi.h>

notviri:
This PR adds bindings to the minimal `timeapi.h` interface, which is very useful for configuring the system timer resolution to get accurate `Sleep()` and such. Since this API is part of WinMM, a new file has been created for it, following the current pattern of one file per system library with the two basic WinMM types that `timeapi.h` depends on added as well.

I've also tested all the functions just in case since there's very few of them.

### DIFF
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -25,6 +25,7 @@ pub const shell32 = @import("windows/shell32.zig");
 pub const user32 = @import("windows/user32.zig");
 pub const ws2_32 = @import("windows/ws2_32.zig");
 pub const gdi32 = @import("windows/gdi32.zig");
+pub const winmm = @import("windows/winmm.zig");
 
 pub usingnamespace @import("windows/bits.zig");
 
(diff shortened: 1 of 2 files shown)


lithdew:
For the tests you've done, is it possible to include them since the API surface is small?

notviri:
> For the tests you've done, is it possible to include them since the API surface is small?

Do you mean in the form of a Zig testcase? The actual results ar entirely hardware dependent, if that's what you're asking.

lithdew:
> > For the tests you've done, is it possible to include them since the API surface is small?
> 
> Do you mean in the form of a Zig testcase? The actual results are entirely hardware dependent, if that's what you're asking.

How about a smoke test just to check that the methods themselves can be used?

notviri:
> > > For the tests you've done, is it possible to include them since the API surface is small?
> > 
> > 
> > Do you mean in the form of a Zig testcase? The actual results are entirely hardware dependent, if that's what you're asking.
> 
> How about a smoke test just to check that the methods themselves can be used?

I looked in the other `std.os.windows.*` files and they didn't contain any "smoke tests" either, so I'll withhold from that for now, but can do if a collaborator requests such a thing.

Vexu:
Looks like this should be exposed in `lib/std/os/windows.zig`.

notviri:
@Vexu You're totally right, my bad. Actually committed referencing the file now.

PR #8802, opened 2021-05-17
std.math.Complex: Change `new` to `init`

aizawey672:


ifreund (changes requested):
When renaming functions like this we prefer to deprecate the old one and replace it with a new one, only removing the deprecated version after a full release cycle. To this end, please add a
```zig
/// Deprecated, use init()
pub const new = init;
```

ifreund (approved):
LGTM, Thanks!

PR #8809, opened 2021-05-17
translate-c: Demote initialization of opaque types

ehaas:
This fixes a segfault in translate-c that would previously occur when
initializing structs with unnamed bitfields, due to a failed assertion in
`transInitListExprRecord`. Unnamed bitfields do not have initializers,
so `transInitListExprRecord` erroneously assumes that `init_count` equals
the number of fields in the record.

Since attempting to initialize an opaque type is a syntax error in Zig,
we can just demote any attempts to initialize them.


ehaas:
Specifically the issue is that unnamed bitfields do not have initializers, so `transInitListExprRecord` erroneously assumes that `init_count` == the number of fields in the record.

PR #8813, opened 2021-05-18
zig fmt: Break container decls if comment/multiline string is found

### LINKED ISSUE #8810: Zig fmt breaks formatting for single-line enums, unions, and structs if they contain comments
Currently, `zig fmt` formats the following:

```zig
const anon = .{
    a, // a
    b
};
const S = struct {
    a: void, // a
    b: void
};
```
(note the missing comma)

to:
```zig
const anon = .{
    a, // a
    b,
};
const S = struct { a: void, // a
b: void };
```

(zig commit: 5bd9a6451685ad056628e312aadaf80e66a4c004; this may be fixed already)

LemonBoy:
Behave as if a trailing comma is present and put every declaration on
its own line.

Closes #8810

### DIFF
--- a/lib/std/zig/parser_test.zig
+++ b/lib/std/zig/parser_test.zig
@@ -4,6 +4,22 @@
 // The MIT license requires this copyright notice to be included in all copies
 // and substantial portions of the software.
 
+test "zig fmt: comment or multiline string in field decl w/ no trailing comma" {
+    try testCanonical(
+        \\const S = struct {
+        \\    a: void, // a
+        \\    b: void
+        \\};
+        \\const S = struct {
+        \\    a: []const u8 =
+        \\        \\ foo
+        \\    ,
+        \\    b: void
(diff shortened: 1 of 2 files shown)


lib/std/zig/parser_test.zig
```diff
@@ -4,6 +4,22 @@
 // The MIT license requires this copyright notice to be included in all copies
 // and substantial portions of the software.
 
+test "zig fmt: comment or multiline string in field decl w/ no trailing comma" {
+    try testCanonical(
+        \\const S = struct {
+        \\    a: void, // a
+        \\    b: void
```
ifreund, line 11: We should insert a trailing comma if there is not one present in this case and use `testTransform()` to check that this is done. This would be consistent with the handling of trailing commas in struct/array initializers and function arguments. 
```suggestion
        \\    b: void,
```

Issue #6674, opened 2020-10-14
c allocator assertion failure in i386-windows-gnu


codehz:
```
D:\Soft\zig\current\lib\zig\std\debug.zig:225:14: 0xf3332f in std.debug.assert (tjs.o)
    if (!ok) unreachable; // assertion failure
             ^
D:\Soft\zig\current\lib\zig\std\heap.zig:47:11: 0xf330f1 in std.heap.cAlloc (tjs.o)
    assert(ptr_align <= @alignOf(c_longdouble));
          ^
>>>>>>>>>> NOTE: ptr_align == 8 and @alignOf(c_longdouble) == 4 <<<<<<<
D:\Soft\zig\current\lib\zig\std\mem\Allocator.zig:294:40: 0xf89721 in std.mem.Allocator.allocAdvancedWithRetAddr (tjs.o)
    const byte_slice = try self.allocFn(self, byte_count, a, len_align, return_address);
                                       ^
D:\Soft\zig\current\lib\zig\std\mem\Allocator.zig:262:41: 0xf895bd in std.mem.Allocator.allocAdvanced (tjs.o)
    return self.allocAdvancedWithRetAddr(T, alignment, n, exact, @returnAddress());
                                        ^
D:\Soft\zig\current\lib\zig\std\array_list.zig:374:59: 0xf864e7 in std.array_list.ArrayListAlignedUnmanaged(mods.Data,null)::std.array_list.ArrayListAlignedUnmanaged(mods.Data,null).initCapacity (tjs.o)
            const new_memory = try allocator.allocAdvanced(T, alignment, num, .at_least);
                                                          ^
D:\Project\tinysh\src\mods.zig:421:73: 0xf833a2 in mods.FunctionInfo::mods.FunctionInfo.invoke (tjs.o)
                var argsdata = std.ArrayListUnmanaged(Data).initCapacity(allocator, args.len) catch return ctx.throw(.OutOfMemory);
```

And it is working if I removed the assert...
but why @alignOf(c_longdouble) == 4?

LemonBoy:
#6385 solves your problems and answers your questions

codehz:
duplicated with #3783

andrewrk:
FYI note that with 13cccdd92cd89242801218aa808dec1df14f482d which is this:

https://github.com/ziglang/zig/blob/13cccdd92cd89242801218aa808dec1df14f482d/lib/std/c.zig#L336-L344

which is based on:

[...]

PR #6385, opened 2020-09-20
std: Make C allocator respect the required alignment

LemonBoy:
Use posix_memalign where available and the _aligned_{malloc,free} API on
Windows.

Closes #3783

PR #8835, opened 2021-05-19
libstd: do not assert if allocs are within std.c.max_align_t in raw_c_allocator

kubkon:
After landing #8554, stage2 compiler is unable to build a simple "Hello, World" on my M1 Mac tripping the following assertion in `rawCAlloc`:

```zig
extern "c" fn exit(usize) noreturn;

pub export fn main() noreturn {
    print();

    exit(0);
}

fn print() void {
    const msg = @ptrToInt("Hello, World!\n");
    const len = 14;
    _ = write(1, msg, len);
}
```

```
❯ ../../zig-out/bin/zig build-exe hello.zig
thread 2795147 panic: reached unreachable code
/Users/kubkon/dev/zig/build/lib/zig/std/debug.zig:226:14: 0x104936893 in std.debug.assert (zig)
             ^
/Users/kubkon/dev/zig/build/lib/zig/std/heap.zig:178:11: 0x1049366ab in std.heap.rawCAlloc (zig)
    assert(ptr_align <= @alignOf(std.c.max_align_t));
          ^
/Users/kubkon/dev/zig/build/lib/zig/std/mem/Allocator.zig:290:40: 0x104a3be2f in std.mem.Allocator.allocAdvancedWithRetAddr (zig)
    const byte_slice = try self.allocFn(self, byte_count, a, len_align, return_address);
                                       ^
/Users/kubkon/dev/zig/build/lib/zig/std/mem/Allocator.zig:159:52: 0x104a37b27 in std.mem.Allocator.create (zig)
    const slice = try self.allocAdvancedWithRetAddr(T, null, 1, .exact, @returnAddress());
                                                   ^
/Users/kubkon/dev/zig/src/Module.zig:3165:36: 0x104a3281f in Module.importPkg (zig)
    const new_file = try gpa.create(Scope.File);
                                   ^
/Users/kubkon/dev/zig/src/Compilation.zig:1605:37: 0x104a3034f in Compilation.update (zig)
            _ = try module.importPkg(module.root_pkg, std_pkg);
                                    ^
/Users/kubkon/dev/zig/src/main.zig:2233:20: 0x1049c0157 in updateModule (zig)
    try comp.update();
                   ^
/Users/kubkon/dev/zig/src/main.zig:1965:17: 0x10495482f in buildOutputType (zig)
                ^
/Users/kubkon/dev/zig/src/main.zig:186:31: 0x10494276f in mainArgs (zig)
        return buildOutputType(gpa, arena, args, .{ .build = .Exe });
                              ^
/Users/kubkon/dev/zig/src/main.zig:141:20: 0x104941ba3 in main (zig)
                   ^
/Users/kubkon/dev/zig/build/lib/zig/std/start.zig:458:37: 0x104cbc28f in std.start.callMain (zig)
                                    ^
/Users/kubkon/dev/zig/build/lib/zig/std/start.zig:400:12: 0x104943bc7 in std.start.callMainWithArgs (zig)
           ^
/Users/kubkon/dev/zig/build/lib/zig/std/start.zig:370:12: 0x104943af7 in std.start.main (zig)
           ^
???:?:?: 0x19ac7544f in ??? (???)
Panicked during a panic. Aborting.
[1]    19416 abort      ../../zig-out/bin/zig build-exe hello.zig
```

I believe this assertion is wrong and should be removed (as I've done in this PR) since it only really covers scalar types:

> According to [cppreference](https://en.cppreference.com/w/cpp/types/max_align_t) on `max_align_t`, its alignment is at least as large as that of _every_ scalar type. On arm64 macOS, `max_align_t == c_longdouble` which then implies `@alignOf(c_longdouble) == 8`. 

In this particular case, `@alignOf(Scope.File) == 16` since `Scope.File` contains an `i128` as a member.

---
Out of curiosity, I have verified that the same behaviour is observed with C++. That is, for the following source:

```cpp
#include <iostream>
#include <cstddef>

int main()
{
  std::cout << alignof(__int128_t) << " " << alignof(std::max_align_t) << '\n';
}
```

we get

```
> clang -Wl,-syslibroot $(xcrun --show-sdk-path) -lSystem -lc++ -std=c++14 cpp.cpp
> a.out
16 8
```

which matches Zig's behaviour.

---
If anyone has got access to an M1, I'd be very grateful if you could verify running this example against stage2 built with `master` branch and check if you see the same reported behaviour.

### DIFF
--- a/lib/std/heap.zig
+++ b/lib/std/heap.zig
@@ -854,6 +854,15 @@ test "raw_c_allocator" {
     }
 }
 
+test "raw_c_allocator alignment check" {
+    if (builtin.link_libc) {
+        const ptr_to_i128 = try raw_c_allocator.create(i128);
+        defer raw_c_allocator.destroy(ptr_to_i128);
+        ptr_to_i128.* = 10;
+        try testing.expect(ptr_to_i128.* == 10);
+    }
+}
+
 test "WasmPageAllocator internals" {
     if (comptime std.Target.current.isWasm()) {
         const conventional_memsize = WasmPageAllocator.conventional.totalPages() * mem.page_size;
--- a/lib/std/heap.zig
+++ b/lib/std/heap.zig
@@ -157,8 +157,7 @@ var c_allocator_state = Allocator{
     .resizeFn = CAllocator.resize,
 };
 
-/// Asserts allocations are within `@alignOf(std.c.max_align_t)` and directly calls
-/// `malloc`/`free`. Does not attempt to utilize `malloc_usable_size`.
+/// Directly calls `malloc`/`free`. Does not attempt to utilize `malloc_usable_size`.
 /// This allocator is safe to use as the backing allocator with
 /// `ArenaAllocator` for example and is more optimal in such a case
 /// than `c_allocator`.
@@ -175,7 +174,6 @@ fn rawCAlloc(
     len_align: u29,
     ret_addr: usize,
 ) Allocator.Error![]u8 {
-    assert(ptr_align <= @alignOf(std.c.max_align_t));
     const ptr = @ptrCast([*]u8, c.malloc(len) orelse return error.OutOfMemory);
     return ptr[0..len];
 }


LemonBoy:
You have to consider `max_align_t` from the C standpoint, where f128 and i128 are non-standard extensions. The maximum alignment is given by a `c_longdouble` that, from what I can see, is defined as a f128 when targeting Linux but not when targeting macOS [see [here](https://github.com/llvm/llvm-project/blob/3d59f9d22440645ca0237dfc5d91ca09f749174b/clang/lib/Basic/Targets/AArch64.cpp#L911)].

Malloc is meant to respect that alignment or, what many allocator do, provide a higher one. But we can't detect this (beside hardcoding a libc-specific whitelist), that's why you're hitting that assertion.

As I've already stated multiple times (see #6385 and #6674) use `c_allocator`, that's the best solution.

Side note, the `c_longdouble` definition should be updated to reflect the target ABI for AArch64.

daurnimator:
I think the `assert` is correct: `malloc` only guarantees suitable alignment for fundamental types (e.g. `long long`): its just implementation details and luck for any higher alignment.

kubkon:
Hmm, ok, that makes sense - thanks for the input @LemonBoy and @daurnimator. But the issue remains that stage2 is unusable on my M1. @andrewrk, do you think we could use `c_allocator` rather than `raw_c_allocator` in `Module.zig`? This should hopefully fix the issue.

Issue #7948, opened 2021-02-04
zig generic memoization broken with comptime slices


marler8997:
The following code will cause the compiler to go into an infinite generic function instantiation loop:

```zig
const std = @import("std");

fn genericFun(comptime s: []const u8) void {
    @compileLog("genericFun with ", s);
    genericFun((s ++ "something")[0..s.len]);
}

pub fn main() void {
    genericFun("foo");
}
```

The problem here is that when Zig checks whether `genericFun("foo")` has already been instantiated, it's not seeing that slices with the same contents are equivalent.

mikdusan:
hmm narrowed to something to do with array operators?

```zig
fn comptimeFun(comptime s: []const u8) void {
    @compileLog("comptimeFun with ", s);
}

export fn foo() void {
    comptimeFun( ("foo" ++ "something")[0..3] );
    comptimeFun( ("foo" ++ "something")[0..3] );
    comptimeFun( ("foo" ** 2)[0..3] );
    comptimeFun( ("foo" ** 2)[0..3] );

    comptimeFun( "foosomething"[0..3] );
    comptimeFun( "foosomething"[0..3] );
}
```

LemonBoy:
> it's not seeing that slices with the same contents are equivalent

Why would it? You're creating brand new slices with different comptime pointers (`++` and `**` allocate a brand new object) and same length, that's why it keeps recursing.

mikdusan:
True. While the content of the slice may compare equal, and slice.len is equal, slice.ptr has changed.

ifreund:
I've hit a similar issue with the memoization before without any `++` or `**` involved. Unable to isolate a test case though:
```
./river/Root.zig:103:59: error: expected type
'.wayland.wayland_server_core.list.Head(.wlroots.types.output_layout.Output,"link").Iterator(.wayland.wayland_server_core.Direction.forward)', found
'.wayland.wayland_server_core.list.Head(.wlroots.types.output_layout.Output,"link").Iterator(.wayland.wayland_server_core.Direction.forward)'

    return .{ .inner = self.output_layout.outputs.iterator(dir) };
                                                          ^
./river/InputManager.zig:153:54: note: called from here
    var output_it = self.server.root.outputLayoutIter(.forward);
                                                     ^
./river/InputManager.zig:144:8: note: called from here
) void {
       ^
[...]

PR #8839, opened 2021-05-20
idea to solve #7948

marler8997:
NOTE: there is a variation on this idea here https://github.com/ziglang/zig/pull/8846

This addresses the comptime memoization issue #7948 in `std.fmt` by converting the `comptime []const u8` fmt argument into a `[T]u8`.  This addresses the infinite loop of instantiating `formatType` an infinite number of times, and at the same time, also fixes `formatType` to correctly memoize instantiations based on the format string contents, rather than for each instance of a format string.  This means it will work correctly even if it is called directly by the user.

For example,  the following calls will now all forward to the same instantiation of `formatTypeImpl`:

```zig
formatType(0, "x"        , opt, w, 1);
formatType(0, "x" ++ ""  , opt, w, 1);
formatType(0, &[_]{'x'}  , opt, w, 1);
formatType(0, "{x}"[1..2], opt, w, 1);
```

I've submitted this PR as a proof-of-concept.  This pattern might be considered a bit "over the top" to address this issue, in which case we could consider other options or language improvements to make this pattern easier to adhere to.  Note that there are a few dozen other locations in `std` that have the same issue, so whatever solution we go with will establish a pattern for many other locations in `std` and I'm sure user's code also.

> NOTE: you should be able to find the other case with the following grep command:

```sh
$ grep -r "comptime [a-zA-Z_]*: \[\]const u8" lib/std
```

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -2915,3 +2915,7 @@ pub fn alignInSlice(slice: anytype, comptime new_alignment: usize) ?AlignedSlice
     const aligned_slice = bytesAsSlice(Element, aligned_bytes[0..slice_length_bytes]);
     return @alignCast(new_alignment, aligned_slice);
 }
+
+pub fn sliceToArray(comptime T: type, comptime s: []const T) [s.len]T {
+    return s[0..s.len].*;
+}
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -388,19 +388,16 @@ pub fn formatAddress(value: anytype, options: FormatOptions, writer: anytype) @T
     @compileError("Cannot format non-pointer type " ++ @typeName(T) ++ " with * specifier");
 }
 
-// This ANY const is a workaround for: https://github.com/ziglang/zig/issues/7948
-const ANY = "any";
-
 fn defaultSpec(comptime T: type) [:0]const u8 {
     switch (@typeInfo(T)) {
-        .Array => |_| return ANY,
+        .Array => |_| return "any",
         .Pointer => |ptr_info| switch (ptr_info.size) {
             .One => switch (@typeInfo(ptr_info.child)) {
                 .Array => |_| return "*",
                 else => {},
             },
             .Many, .C => return "*",
-            .Slice => return ANY,
+            .Slice => return "any",
         },
         .Optional => |info| return defaultSpec(info.child),
         else => {},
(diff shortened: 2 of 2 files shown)


lib/std/mem.zig
```diff
@@ -2748,3 +2748,11 @@ pub fn alignInSlice(slice: anytype, comptime new_alignment: usize) ?AlignedSlice
     const aligned_slice = bytesAsSlice(Element, aligned_bytes[0..slice_length_bytes]);
     return @alignCast(new_alignment, aligned_slice);
 }
+
+pub fn sliceToArray(comptime T: type, comptime s: []const T) [s.len]T {
```
SpexGuy, line 2752: This should use `s[0..s.len].*`, to avoid impacting the eval branch quota.

andrewrk:
Both of these pull requests are workarounds that are as ugly or uglier than the one that they are trying to fix. I don't really want to merge either pull request.

marler8997:
Bare with me as I try to summarize the context around this PR. The issue at hand is that almost all instances of `comptime s: []const u8` parameters don't actually want the semantics that this comptime type represents.  This parameter type will cause a new instance of the function to be instantiated for every string passed in.  The problem is that in most cases, the programmer only intends for a new instance to be instantiated for each unique string based solely on its contents, and not the memory region in which it resides.  In some cases this problem will only cause extra code to be generated or more work for the compiler, but in other cases (like this one) it will result in subtle bugs, including infinite comptime evaluation when a function like this is recursive.

There is a proposal (can't find it at the moment) to mitigate the issues around this discrepancy between what the programmer intends and what the language provides, however it can't handle all cases (such as mutable strings) and I'd argue adds a fair amount of complexity to how the language handles strings.  The language also provides a way to achieve the correct semantics, but you end up with this ugly solution.  After getting up to speed on all this, I decide to create a proposal that would allow the language to represent the correct semantics in a "non-ugly" way, which is https://github.com/ziglang/zig/issues/9056. This appeals to the Scott Meyer's principle "make it easy to do the right thing and hard to do the wrong thing".  My argument is that `comptime s: []const u8` is the "wrong thing" (semantics are wrong)  but it's easy, doing what this PR does is the "right thing" (right semantics) but it's ugly and difficult.

andrewrk:
I see, thank you for taking the extra time to explain it to me again. So can we close these 2 pull requests by linking them into the #9056 proposal as "evidence"?

marler8997:
Closing this PR which is evidence that the current language solutions are pretty ugly and a solution like #9056 may be justified.

Issue #1481, opened 2018-09-06
stage1 C ABI compatibility


andrewrk:
This issue is to track C ABI compatibility support in the stage1 compiler.

Here's what support looks like currently:
 * [x] integers, floats, pointers, enums, bools work on every target, as parameters and return values
 * [x] x86_64: struct & union parameters & return values larger than 16 bytes
 * [x] x86_64: struct & union parameters that have at least 1 integer in them
 * [x] x86_64: struct & union parameters <= 16 bytes which break structs into parameters
 * [x] x86_64: struct & union parameters <= 16 bytes which are all floats
 * [ ] x86_64: struct & union return values <= 16 bytes 
 * [x] ARM: struct & union return values 
 * [ ] ARM: struct & union parameters
 * [x] C ABI for parameters that is unsupported gives a compile error linking to this issue
 * [x] C ABI for return values that is unsupported gives a compile error linking to this issue
 * [ ] other architectures structs & unions parameters & return values

For those who find this issue from the compile error, leave a comment detailing your specific needs and I'll see if I can code those up for you to unblock you, so you don't have to wait for this issue to be 100% solved.

andrewrk:
Update:

I just merged a branch into master which does the following things:

 * Unsupported C ABI gives a compile error, like this:

```zig
const ASmallUnion = extern union {
    r: u8,
};
extern fn foo(x: ASmallUnion) void;

pub fn main() void {
    const x = ASmallUnion{ .r = 0x12 };

    foo(col);
}
```

```
/home/andy/downloads/zig/build/test.zig:4:8: error: TODO: support C ABI for more targets. https://github.com/ziglang/zig/issues/1481
extern fn foo(x: ASmallUnion) void;
       ^
/home/andy/downloads/zig/build/test.zig:4:8: note: pointers, integers, floats, bools, and enums work on all targets
extern fn foo(x: ASmallUnion) void;
       ^
```

 * There is a new type of test, C ABI tests, which covers the C ABI that we do suport. See `test/stage1/c_abi/*` for the source.

```
Test 1/9 C importing Zig ABI Tests...OK
Test 2/9 C ABI integers...OK
Test 3/9 C ABI floats...OK
Test 4/9 C ABI pointer...OK
Test 5/9 C ABI bool...OK
Test 6/9 C ABI array...OK
Test 7/9 C ABI big struct...OK
Test 8/9 C ABI big union...OK
Test 9/9 C ABI small struct of ints...OK
```

[...]

Issue #5049, opened 2020-04-15
Proposal: Integer-backed packed struct


SpexGuy:
So far, Zig doesn't have a good solution for flags/bitfields.  Packed structs are the current recommended solution, and they provide a lot of improvements over the standard C approach of defining integer constants.  But they are still not quite sufficient, even with #3133.  For me, the requirements of a good flags/bitfield type are:
1) Deterministic packing
2) Can be cast to and from an integer type, to support bulk operations (`&`, `|`, etc)
3) Well-defined conversion between little-endian and big-endian
4) Guarantees the size of its loads and stores in a predictable way, to be compatible with MMIO registers
5) Behaves as an integer when used as a parameter to an extern function, so as to be compatible with a C ABI that uses flags

Requirements (1) and (2) are already satisfied by packed structs.  Requirements (3) and (4) are not satisfied by packed structs yet, but are kind of a package deal.  If you define load/store size you get deterministic endianness behavior.  There have been ways suggested to do this, such as putting a `u0` aligned to the required word size at the beginning of a packed struct. (I can't find that suggestion anymore, if anyone knows where it is please let me know and I'll link it here).  But assuming packed structs will always behave like structs from an ABI standpoint, they may not be able to satisfy requirement (5) for all calling conventions.  This can lead to some [pretty ridiculous workarounds](https://github.com/SpexGuy/Zig-Vulkan-Headers/commit/c4fbb4cbd2e6649e0d477158ef4b348a38a2a1b9#diff-4a8b4c78462d32707b76f7b8fa91e4b7), that are not always feasible.

To solve this, I propose a variant of packed structs which are backed by a specified integer type, similar to enums backed by a specified integer type.  Just like enums *are* integers at the ABI level, integer-backed packed structs are also integers at the ABI level.  They are passed in registers across function call boundaries whenever their backing integer type would be, and they are allowed on extern boundaries only when their backing integer type would be. They can be used in atomic operations just like their integer type would be (with the exception of RMW operations like add, sub, etc, similar to enums).  Their default alignment matches the alignment of their backing integer type.

My proposed syntax is this:
```zig
const ExampleFlags = packed struct(u32) {
	int_flag: u1 = 0,
	bool_flag: bool = false,
	small_value: u3 = 3,
	aligned_value: u8 align(1),

	const Self = @This();
	pub fn union(self: Self, other: Self) Self {
		return @bitCast(Self, @bitCast(u32, self) | @bitCast(u32, other));
	}
};
```

[...]

PR #8840, opened 2021-05-20
translate-c: Add bitfield support

ehaas:
Some notes:

Implements bitfields with `packed` structs. Structs that do not contain bitfields should not be affected.

Should this problem even be solved this way - is the proper approach instead for `extern` structs to understand the ABI struct packing rules?

Even given a perfect translation in terms of layout, this might actually cause more problems than it solves since the C ABI is currently not supported for some packed structs (https://github.com/ziglang/zig/issues/1481) - for things like `@sizeOf` the size could be inlined, but calling a function and passing a struct by value won't work and in fact will cause a compile issue.

I implemented casts with `std.meta.cast` just to get something working but those can be replaced with builtins; compound assignments would still need to be translated with proper casts.

The `-target` flag to `zig translate-c` is important since the target ABI affects the layout. For example given this C struct:
```c
struct foo {
   unsigned x : 2;
   unsigned z : 16;
   unsigned y: 2;
   int *p;
};
```
With a `-target` of `x86_64-macos-gnu` I get:
```zig
pub const struct_foo = packed struct {
    x: u2 align(8),
    z: u16,
    y: u2,
    @"anon.padding_3": u44 = undefined,
    p: [*c]c_int,
};
```

With a `-target` of `thumb-freestanding-gnueabihf` I get:
```zig
pub const struct_foo = packed struct {
    x: u2 align(4),
    z: u16,
    y: u2,
    @"anon.padding_3": u12 = undefined,
    p: [*c]c_int,
};
```

### DIFF
--- a/src/clang.zig
+++ b/src/clang.zig
@@ -148,6 +148,9 @@ pub const APSInt = opaque {
 pub const ASTContext = opaque {
     pub const getPointerType = ZigClangASTContext_getPointerType;
     extern fn ZigClangASTContext_getPointerType(*const ASTContext, T: QualType) QualType;
+
+    pub const getTypeSize = ZigClangASTContext_getTypeSize;
+    extern fn ZigClangASTContext_getTypeSize(*const ASTContext, T: QualType) u64;
 };
 
 pub const ASTUnit = opaque {
@@ -181,6 +184,14 @@ pub const ArraySubscriptExpr = opaque {
 pub const ArrayType = opaque {
     pub const getElementType = ZigClangArrayType_getElementType;
     extern fn ZigClangArrayType_getElementType(*const ArrayType) QualType;
+};
+
+pub const ASTRecordLayout = opaque {
+    pub const getFieldOffset = ZigClangASTRecordLayout_getFieldOffset;
+    extern fn ZigClangASTRecordLayout_getFieldOffset(*const ASTRecordLayout, c_uint) u64;
+
+    pub const getAlignment = ZigClangASTRecordLayout_getAlignment;
+    extern fn ZigClangASTRecordLayout_getAlignment(*const ASTRecordLayout) i64;
 };
 
 pub const AttributedType = opaque {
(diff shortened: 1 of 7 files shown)


ehaas:
@SpexGuy @andrewrk curious what you think of this approach and whether it's worth finishing it up (dealing with compound assignments; directly using builtins for casting instead of std.meta.cast). I don't think it should be merged since currently it would cause compile errors for any functions that try to pass these structs by value, but I could maintain the branch until the C ABI support is ready

SpexGuy:
I don't know if packed structs are well-suited to model C bitfields, even once we fix all the bugs.  It might be better to generate something like this instead:
```zig
pub const foo = extern struct {
  _bitfield_0: c_uint,
  p: [*c]c_int,
  
  pub fn xGet(self: @This()) u2 {
    return @truncate(u2, self._bitfield_0 >> 0);
  }
  pub fn xSet(self: @This(), value: u2) void {
    self._bitfield_0 = (self._bitfield_0 & 0x3) | (std.c.zeroExtend(c_uint, value) << 0);
  }
  pub fn yGet(self: @This()) u16 {
    return @truncate(u16, self._bitfield_0 >> 2);
  }
  pub fn ySet(self: @This(), value: u16) void {
    self._bitfield_0 = (self._bitfield_0 & 0x3FFFC) | (std.c.zeroExtend(c_uint, value) << 2);
  }
  pub fn zGet(self: @This()) u2 {
    return @truncate(u2, self._bitfield_0 >> 18);
  }
  pub fn zSet(self: @This(), value: u2) void {
    self._bitfield_0 = (self._bitfield_0 & 0xC0000) | (std.c.zeroExtend(c_uint, value) << 18);
  }
};
```
It might be possible to have a single function for each field which returns a bit-packed pointer, I'm not sure if those can be created directly with the current compiler.

If #5049 is accepted it might be worth it to generate fields of those types instead, like this:
```zig
pub const foo = extern struct {
  b0: packed struct (c_uint) {
    x: u2,
    y: u16,
    z: u2,
  },
  p: [*c]c_int,
}
```

But without that, the packing and abi rules are iffy.  Especially once byte swapping gets involved.  I think there's a lot of value in having translate-c generate portable code where possible, so if we can avoid generating different code for different targets I think that's a plus.

xackus:
@SpexGuy
> I think there's a lot of value in having translate-c generate portable code where possible, so if we can avoid generating different code for different targets I think that's a plus.

This isn't possible with the way we use clang. The AST we get has all the macros already expanded, which can have target specific values.

Vexu:
This seems mostly fine but I don't want to add any packed structs to translate-c with how broken they currently are. We should look at this again once we get proper packed structs in stage2 or integer backed packed structs. I think translating bitfields as functions as suggested by SpexGuy would also be something worth looking into.

PR #8844, opened 2021-05-20
Support inline keyword as well as callconv(.Inline)

### LINKED ISSUE #6429: support both inline keyword as well as callconv(.Inline)
[accepted proposal](https://github.com/ziglang/zig/issues/6429#issuecomment-843757556)

--------------

`inline` is mutually exclusive with a custom calling convention, I think this makes it simpler.
`noinline` is unaffected.

ifreund:
Closes #6429

### DIFF
--- a/lib/std/Thread.zig
+++ b/lib/std/Thread.zig
@@ -68,7 +68,7 @@ else switch (std.Target.current.os.tag) {
 };
 
 /// Signals the processor that it is inside a busy-wait spin-loop ("spin lock").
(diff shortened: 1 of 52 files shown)


marler8997:
Poor inline feeling the whiplash

jedisct1:
Welcome back, `inline`! 🎉

ifreund:
Ah, looks like I forgot about translate-c, will fix shortly

andrewrk:
Thanks!

PR #8846, opened 2021-05-20
alternative idea to solve #7948

marler8997:
This is a variation on https://github.com/ziglang/zig/pull/8839

Instead of creating a wrapper function that takes a comptime slice and converts it to an array, this mechanism sets the array to `anytype` and then recursively calls itself if it needs to be converted to an array.


bradms:
I think changing the type of the format string to `anytype` is confusing

marler8997:
@bradms, I agree.  There's #8839 but it's pretty ugly on the implementation side.  We may need a language change to make this palatable.  For now this PR serves as an example of the current solutions that the language enables.

marler8997:
Closing this PR which is evidence that the current language solutions are pretty ugly and a solution like #9056 may be justified.

Issue #653, opened 2017-12-08
more pointer metadata: address spaces


andrewrk:
GCC has the concept of address spaces, which is useful for embedded programming: https://gcc.gnu.org/onlinedocs/gcc/Named-Address-Spaces.html

In LLVM pointer types have the concept of the "address space" of the pointer: http://llvm.org/docs/LangRef.html#pointer-type

This is easy to support. Just like alignment, global variables can specify the address space that they are in. Just like alignment, pointers can specify their address space, and if it's the default address space, it can be omitted.

We will use a simple integer for the address space. 0 is the default. If an application wants to have a name for an address space, it can assign an integer to a constant. If an application wants to coordinate address spaces with a package it depends on, the package should accept a configuration option to specify the integer mapping for a given address space name, and then both the application's constants and the package's constants will refer to the same integer.

new keyword: `addrspace`

It can be used to create an address space constant:

```zig
const default = addrspace(0);
const mapper_hardware_ram = addrspace(1);
```

It can be used in the pointer syntax:

`&addrspace(mapper_hardware_ram) u32`

The type of `mapper_hardware_ram` is `addrspace`. The only thing you can do with it is use it in pointer syntax and global variable syntax.

Global variable:

```zig
var foo: u32 align(4) addrspace(mapper_hardware_ram) = 10;
```

Implicit casting and explicit casting does not allow changing address space of a pointer.

However you can use `@addrspaceCast(addr_space, ptr)` to (unsafely) override the address space of something.

thejoshwolfe:
how about:

```zig
const default = 0;
const mapper_hardware_ram = 1;
var foo: u32 align(4) addrspace(mapper_hardware_ram) = 10;
```

why do we want an addrspace type? seems like it will cause more problems than it will solve. like, in your example, if we inline `mapper_hardware_ram`, we get `addrspace(addrspace(1))`.

PavelVozenilek:
Nim has it under name "Memory regions" ( https://nim-lang.org/docs/manual.html#types-memory-regions ). AFAIK it is pretty useless, but Nim managed to drop a feature only once.

Ilariel:
I believe the Nim "Memory regions" are a language level feature used for extra safety when dealing with pointers to possibly semantically different memory areas such as heap, stack, objects from another language and so on. As on they only exists in source but not in the compiled code.

The gcc and llvm address spaces however are a backend specific information required for correct codegen when memory and code exists in different memory.

[...]

PR #8852, opened 2021-05-21
SPIR-V: More codegen

Snektron:
This PR adds again some more SPIR-V codegen stuff:
- Reworked generation of float constants a bit, now uses Value.toFloat instead of the raw tag type.
- Some general restructuring.
- Integer encoding: This turned out to be harder than i anticipated, mostly because there is no decent way yet to obtain a 2s complement value for an int. For now, i handled it through toUnsignedInt and toSignedInt and then did some bit operations on that. I'm not yet sure how to encode "strange" integer types, so for now they're just inserted as an equivalent constant of the backing type.
- Split out genCmp and genBinOp as Andrew suggested, though i have not yet removed the extra switch in genCmp, genBinOp and genUnOp.
- A preliminary version of generating alloc, store and load.
  * They only work for local variables (with storage class `Function`) as of now. As these operations require pointers, and so more advanced stuff depends on #653. This is next on the agenda.
  * Local variables need to be declared in the first SPIR-V block, and so i made an additional arraylist in DeclGen to temporarily hold code generated by the current function, which OpVariable bypasses as it is inserted directly into spv.binary.fn_decls.  
- In order to get more useful error messages during developing i pass the instruction's source location to genType and genConstant, though i'm not actually sure whether this is the right source location (it points to the instruction instead of the type). Furthermore, how do i pass the proper source of stuff like types of function parameters? 
- With #8554 merged, dbg_stmt now contains the actual source offset and so we can nicely generate that in SPIR-V debug info. Both OpLine and OpSource are emitted now.

### DIFF
--- a/src/codegen/spirv.zig
+++ b/src/codegen/spirv.zig
@@ -15,7 +15,7 @@ const ir = @import("../ir.zig");
 const Inst = ir.Inst;
 
 pub const TypeMap = std.HashMap(Type, u32, Type.hash, Type.eql, std.hash_map.default_max_load_percentage);
-pub const ValueMap = std.AutoHashMap(*Inst, u32);
+pub const InstMap = std.AutoHashMap(*Inst, u32);
 
 pub fn writeOpcode(code: *std.ArrayList(u32), opcode: Opcode, arg_count: u32) !void {
     const word_count = arg_count + 1;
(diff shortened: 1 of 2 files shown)


andrewrk:
Nice work!

PR #8857, opened 2021-05-21
wasm: produce static lib by default and rename produced object files to simply .o from .o.wasm

kubkon:
### Produce static lib by default

This matches the behaviour for other targets in that

```
> zig build-lib math.zig -target wasm32-freestanding
libmath.a
```

produces now `libmath.a` while

```
> zig build-lib math.zig -dynamic -target wasm32-freestanding
math.wasm
```

is required to create a loadable Wasm module.

### Rename produced object files to .o from .o.wasm

This should offer more compatibility with external tooling when cross-compiling to wasm with zig.

```
> zig build-obj hello.zig -target wasm32-freestanding
hello.o
```

### DIFF
--- a/doc/docgen.zig
+++ b/doc/docgen.zig
@@ -284,6 +284,7 @@ const Code = struct {
     link_objects: []const []const u8,
     target_str: ?[]const u8,
     link_libc: bool,
+    link_mode: ?std.builtin.LinkMode,
     disable_cache: bool,
 
     const Id = union(enum) {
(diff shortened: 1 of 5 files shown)


jedisct1:
Yes!

Consistency with other platforms and compilers is great.

Also, having a different extension for objects previously confused `autoconf`-generated scripts, that didn't see `zig cc` as gcc-compatible compiler any more.

jedisct1 (approved):

PR #8863, opened 2021-05-22
fix Boyer-Moore-Horspool algorithm in indexOfPos and lastIndexOf when element type is larger than a byte

mattbork:
Currently, `indexOfPos` and `lastIndexOf` are broken when the Boyer-Moore-Horspool algorithm is used for types larger than a byte. In this case, `haystack` and `needle` are first cast from slices of `T` to slices of `u8` before a BMH search is run. This is fine, but returned indices are not adjusted from byte positions back to element positions. Further, matches are not checked for alignment with the original elements, so false positives can occur. For example, one a little-endian machine:

```
// make large enough to trigger BMH algorithm
const haystack = [1]u16{0} ** 50 ++ [_]u16 { 0xbbaa, 0xccbb, 0xddcc, 0xeedd, 0xffee, 0x00ff };
const needle0 = [_]u16{ 0xbbaa, 0xccbb, 0xddcc, 0xeedd, 0xffee };
assert(mem.indexOfPos(u16, &haystack, 0, &needle0) == 100); // oops, not 50 because byte index was returned
const needle1 = [_]u16{ 0xbbbb, 0xcccc, 0xdddd, 0xeeee, 0xffff };
assert(mem.indexOfPos(u16, &haystack, 0, &needle1) == 101); // oops, not null because misaligned match was found
```

The first search returns an index in bytes rather than `u16` elements because it doesn't divide by `@sizeOf(u16)` before returning. The second search returns a match when it shouldn't because it wasn't restricting matches to only positions that are multiples of `@sizeOf(u16)`. I've fixed these cases and added a basic test for them.

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -1107,7 +1107,9 @@ pub fn lastIndexOf(comptime T: type, haystack: []const T, needle: []const T) ?us
 
     var i: usize = haystack_bytes.len - needle_bytes.len;
     while (true) {
-        if (mem.eql(u8, haystack_bytes[i .. i + needle_bytes.len], needle_bytes)) return i;
+        if (i % @sizeOf(T) == 0 and mem.eql(u8, haystack_bytes[i .. i + needle_bytes.len], needle_bytes)) {
+            return @divExact(i, @sizeOf(T));
+        }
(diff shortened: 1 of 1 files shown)


andrewrk:
Thank you for the fix, the tests, and the detailed bug description!

Issue #8860, opened 2021-05-22
homebrew-provided LLVM/Clang prevents zig from building C files


Jarred-Sumner:
Note from @andrewrk: A workaround for this is to use the tarballs from [the download page](ziglang.org/download/).

---------

zig version: `0.8.0-dev.2591+4b69bd61e`
macOS version: `11.3.1`

To repro, run this:
```
echo "#include <assert.h>" > file.c
zig cc -lc hello.c 
```

Response:
```bash
jarred@c-73-252-132-29 ~/Code [1]> zig cc -lc hello.c
hello.c:1:10: fatal error: cannot open file '/usr/local/Cellar/zig/HEAD-4b69bd6_1/lib/zig/include/assert.h': No such file or directory
#include <assert.h>
         ^
```

Passing the include path does not fix it:
```bash
zig cc -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include /Users/jarred/Code/esdev/src/deps/picohttpparser.c 
/Users/jarred/Code/esdev/src/deps/picohttpparser.c:27:10: fatal error: cannot open file '/usr/local/Cellar/zig/HEAD-4b69bd6_1/lib/zig/include/assert.h': No such file or directory
#include <assert.h>
```

This is consistent, whether it's `zig translate-c`, `zig cc`, `@cInclude`, or `exe.addCSourceFile` etc.

Initially, I thought it was something specific to my environment -- but after wiping my hard-drive & reinstalling macOS, it's hard to see how it could be. 

Note that llvm 12 is installed and this occurs whether it's a copy of zig built from homebrew or manually running `cmake`

This seems to work on zig 0.7.1, so this seems to be something that broke recently:
```
jarred@c-73-252-132-29 ~/Code> zig cc -lc hello.c
Undefined symbol: C entry: _main
symbol(s) not found
error: FileNotFound
```

After adding a main function, `zig 0.7.1` can build it, but zig on `HEAD` cannot

Jarred-Sumner:
Hm. After lots of retries, it seems to work now. But, I'm not 100% sure why.

I suspect the necessary step was adding this to my `$HOME/.config/fish/config.fish` prior to building zig from source:
```fish
set -ax LDFLAGS "-L/usr/local/opt/llvm/lib -Wl,-rpath,/usr/local/opt/llvm/lib"
[...]

PR #8864, opened 2021-05-22
zld: permit system static libs

kubkon:
This commits permits passing in static archives using the system lib flag `-la`. With this commit, `zig ld` will now look firstly for a dynamic library (which always takes precedence), and will fall back on `liba.a` if the dylib is not found. The static archive is searched for in the system lib search dirs like the dylibs.

cc @jedisct1 @Jarred-Sumner hopefully, this addesses issue #8860, so I'd be indebted if you could try this commit out!

### DIFF
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -698,8 +698,8 @@ fn linkWithLLD(self: *MachO, comp: *Compilation) !void {
                 try positionals.append(comp.libcxx_static_lib.?.full_object_path);
             }
 
-            // Shared libraries.
-            var shared_libs = std.ArrayList([]const u8).init(arena);
+            // Shared and static libraries passed via `-l` flag.
+            var libs = std.ArrayList([]const u8).init(arena);
             var search_lib_names = std.ArrayList([]const u8).init(arena);
 
             const system_libs = self.base.options.system_libs.items();
@@ -708,9 +708,8 @@ fn linkWithLLD(self: *MachO, comp: *Compilation) !void {
                 // By this time, we depend on these libs being dynamically linked libraries and not static libraries
                 // (the check for that needs to be earlier), but they could be full paths to .dylib files, in which
                 // case we want to avoid prepending "-l".
-                // TODO I think they should go as an input file instead of via shared_libs.
                 if (Compilation.classifyFileExt(link_lib) == .shared_library) {
-                    try shared_libs.append(link_lib);
+                    try positionals.append(link_lib);
                     continue;
                 }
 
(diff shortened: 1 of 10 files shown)


kubkon:
I should add here, this should make it possible to use Homebrew distributed LLVM on x86_64 macOS with the standard CMake invocation:

```
cmake .. -DCMAKE_PREFIX_PATH=$(brew --prefix llvm)
```

and allow successful compilation of stage2 with the built stage1 compiler this way:

```
build/bin/zig build -Denable-llvm
```

arm64 macOS will require a little more work since we'll need the ability to parse TAPI files (`.tbd`) which are shipped in place of `.dylib`s.

jedisct1:
Unfortunately, that doesn't seem to fix it :(

```text
rm -fr ~/.cache/zig;zig cc -target x86_64-linux-musl  a.c
a.c:1:10: fatal error: cannot open file '/opt/zig/lib/zig/include/stdio.h': No such file or directory
#include <stdio.h>
         ^
```

`-DZIG_STATIC_LLVM=on` works, though.

kubkon:
> Unfortunately, that doesn't seem to fix it :(
> 
> ```
> rm -fr ~/.cache/zig;zig cc -target x86_64-linux-musl  a.c
> a.c:1:10: fatal error: cannot open file '/opt/zig/lib/zig/include/stdio.h': No such file or directory
> #include <stdio.h>
>          ^
> ```
> 
> `-DZIG_STATIC_LLVM=on` works, though.

Thanks for the report @jedisct1. Could you try building stage2 with this patch and without `-DZIG_STATIC_LLVM=on`, and verifying the stage2 binary actually works (i.e., doesn’t report missing symbols, etc.)?

jedisct1:
`stage2` doesn't build:

```text
error(compilation): clang failed with stderr: /Users/j/src/zig/lib/libcxx/src/utility.cpp:9:10: fatal error: cannot open file '/Users/j/src/zig/lib/libcxx/src/utility': No such file or directory

error(compilation): clang failed with stderr: In file included from /Users/j/src/zig/lib/libcxx/src/new.cpp:9:
In file included from /Users/j/src/zig/lib/libcxx/include/stdlib.h:87:
/Users/j/src/zig/lib/libcxx/include/__config:1135:46: fatal error: cannot open file '/Users/j/src/zig/lib/libcxx/include/pthread.h': No such file or directory

error(compilation): clang failed with stderr: /Users/j/src/zig/lib/libcxx/src/valarray.cpp:9:10: fatal error: cannot open file '/Users/j/src/zig/lib/libcxx/src/valarray': No such file or directory
...
```

kubkon:
> `stage2` doesn't build:
> 
> ```
> error(compilation): clang failed with stderr: /Users/j/src/zig/lib/libcxx/src/utility.cpp:9:10: fatal error: cannot open file '/Users/j/src/zig/lib/libcxx/src/utility': No such file or directory
> 
> error(compilation): clang failed with stderr: In file included from /Users/j/src/zig/lib/libcxx/src/new.cpp:9:
> In file included from /Users/j/src/zig/lib/libcxx/include/stdlib.h:87:
> /Users/j/src/zig/lib/libcxx/include/__config:1135:46: fatal error: cannot open file '/Users/j/src/zig/lib/libcxx/include/pthread.h': No such file or directory
> 
> error(compilation): clang failed with stderr: /Users/j/src/zig/lib/libcxx/src/valarray.cpp:9:10: fatal error: cannot open file '/Users/j/src/zig/lib/libcxx/src/valarray': No such file or directory
> ...
> ```

Huh, very interesting. I’m more unclear what is actually going wrong. I definitely need to try and repro all the issues with paths.

PR #8865, opened 2021-05-22
make writeIntSlice functions work for signed integers

mattbork:
The `writeInt` functions accept signed integers, and the comments for the `writeIntSlice` functions say they write two's complement integers, implying the same. However, to get the bits of the value passed in, the `writeIntSlice` functions try to `@truncate` to an unsigned type with the same number of bits. This causes a compile error because `@truncate` expects an unsigned value. For example,

```
var buf: [2]u8 = undefined;
std.mem.writeIntLittle(i16, &buf, -21555); // ok
std.debug.assert(mem.eql(u8, &buf, &[2]u8 { 0xcd, 0xab }));
std.mem.writeIntSliceLittle(i16, &buf, -21555); // compile error
std.debug.assert(mem.eql(u8, &buf, &[2]u8 { 0xcd, 0xab }));
```

The obvious fix is just to use `@bitCast` instead of `@truncate` here. I've made this change and expanded the test for the `writeIntSlice` functions to cover signed integers too.

Curiously, I noticed that trying to `@truncate` a signed integer to an unsigned integer actually works at comptime. So,

```
var x: i16 = -21555;
std.debug.assert(@truncate(u16, x) == @as(u16, 0xabcd)); // compile error
```

but,

```
comptime var x: i16 = -21555;
std.debug.assert(@truncate(u16, x) == @as(u16, 0xabcd)); // ok
```

I don't know if this is intended or if I should make a separate issue for this behavior.

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -1412,7 +1412,7 @@ pub fn writeIntSliceLittle(comptime T: type, buffer: []u8, value: T) void {
 
     // TODO I want to call writeIntLittle here but comptime eval facilities aren't good enough
     const uint = std.meta.Int(.unsigned, @typeInfo(T).Int.bits);
-    var bits = @truncate(uint, value);
+    var bits = @bitCast(uint, value);
     for (buffer) |*b| {
         b.* = @truncate(u8, bits);
         bits >>= 8;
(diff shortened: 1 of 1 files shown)


andrewrk:
> Curiously, I noticed that trying to `@truncate` a signed integer to an unsigned integer actually works at comptime.
> I don't know if this is intended or if I should make a separate issue for this behavior.

`error: expected unsigned integer type, found 'i16'` is correct - nice catch with finding the problem with comptime code not catching the issue. If you wouldn't mind filing an issue that would be much appreciated.

PR #8866, opened 2021-05-22
std.sync.atomic: extended atomic helper functions

kprotty:
* generic Atomic(T) wrapper replacing atomic.Int and atomic.Bool
* Bitwise atomic intrinsics when Atomic(T) is an Integer
* No redundant specification of type as seen in builtins atomic functions
* fetchXX() ops instead of generic atomicRmw for C11 consistency 
* contains compilerFence() and spinLoopHint()


lib/std/sync/atomic.zig
```diff
@@ -0,0 +1,604 @@
...
+
+const std = @import("../std.zig");
+const target = std.Target.current;
+const testing = std.testing;
+
+const AtomicOrder = std.builtin.AtomicOrder;
+const AtomicRmwOp = std.builtin.AtomicRmwOp;
+const AtomicBitRmwOp = enum { Set, Reset, Toggle };
+
+pub fn spinLoopHint() void {
```
LemonBoy, line 15: There's already a copy of this function in the stdlib with several enhancements.

LemonBoy:
We already have atomic wrappers for ints and bools. That approach is much safer than taking raw pointers as it prevents you from mixing atomic and non-atomic ops.

lib/std/sync/atomic.zig
kprotty: Just saw the enhanced version. Will move `std.Thread.spinLoopHint()` here and have it alias this instead

kprotty:
@LemonBoy atomic int and bool dont contain cmpxchg abstractions + there aren't any such wrappers for atomic pointers/enums/floats. Not sure about improved safety as ordering misuse/ignorance can often be equally unsafe. Mixing non-atomic ops and atomic ops can sometimes be valid. Would a generic `Atomic(T)` type with `load/storeUnchecked()` and limiting fetchXX()/bitXX() for when T is an Int be a good compromise?

LemonBoy:
> atomic int and bool dont contain cmpxchg abstractions 

Should be added if needed.

> there aren't any such wrappers for atomic pointers/enums/floats

Wrappers for pointers and enums are also a good idea. Floating point types can be added later (if ever) when somebody finds an use for those.

Not sure about improved safety as ordering misuse/ignorance can often be equally unsafe.

The wrappers try to protect the user from shooting themselves in the foot by checking the ordering parameter, avoid defining meaningless operations, and preventing the mix-up of atomic & non atomic ops. The idea of wrapper types is not novel (see Rust or C11) and strikes a good balance between ergonomics and safety.

> Would a generic Atomic(T)

I think separate wrapper types are simpler to define, you need quite a bit of meta programming and `usingnamespace` to achieve the same effect and define some helpers only when they make sense.

kprotty:
Separate wrapper types feels like repetition: They would all expose load/store/xchg/cmpxchg and ordering checks. Floats would expose that plus arithmetic rmw. Ints would expose even that plus bitwise rmw. Theres few metaprogramming conditions but it covers a lot of cases. Thats my pitch for Atomic(T)

LemonBoy:
> Thats my pitch for Atomic(T)

I'm sold!
Given we're in for some sweeping changes, what do you think of using the same nomenclature as the [C++](https://en.cppreference.com/w/cpp/atomic) atomic types (`load, store, exchange, fetch{Add,Sub,...}, compareExchange{Strong,Weak},...`) ?

kprotty:
I prefer everything except the `exchange` bits - replacing them for `swap`/`[try]compareAndSwap` is on average shorter and is reminiscent of [Go's atomic naming scheme](https://golang.org/pkg/sync/atomic/).

Actually, another issue with wrapper types is being able to store them inside `extern struct/union`. Could store the generic bytes and cast between them like `atomic_bytes: [@sizeOf(Atomic(T))] align(@alignOf(Atomic(T))` but that is pretty hacky.

LemonBoy:
> Actually, another issue with wrapper types is being able to store them inside extern struct/union

There should be no problem if the wrapper is `extern`.

kprotty:
freebsd CI: `atomic.zig:25:13: error: expected 0 parameters, found 1`

atomic.zig:25: `@fence(ordering)`

Whats up with the fence builtin?

LemonBoy:
Change [this](https://github.com/ziglang/zig/blob/6435750c99e705eb40bbdf75e51a3493d683e951/src/BuiltinFn.zig#L403) line.

lib/std/atomic.zig
```diff
@@ -3,14 +3,88 @@
...
+        else => {
+            @compileLog(ordering, " only applies to a given memory location");
+        },
+    }
+}
+
+pub fn compilerFence(comptime ordering: Ordering) callconv(.Inline) void {
+    switch (ordering) {
+        .SeqCst => {
+            // TODO: Use LLVM intrinsic `fence syncscope("singlethread")` instead of inline asm.
```
LemonBoy, line 36: There are some [unresolved problems](https://github.com/rust-lang/rust/issues/62256) with the use of singlethread fences, the biggest one is that they may produce extra instructions or a call (!) to `__sync_synchronize`.
Let's use the good old empty-asm-block trick for the time being (and group the ordering parameters, having to expand 3 layers of recursion is pretty bad for stage1, beside being hard to understand).

kprotty: Weird find.. 

fixed it

lib/std/atomic/Atomic.zig
```diff
@@ -0,0 +1,522 @@
...
+        var x = Atomic(usize).init(5);
+        try testing.expectEqual(x.swap(10, ordering), 5);
+        try testing.expectEqual(x.load(.SeqCst), 10);
+
+        var y = Atomic(enum(usize) { a, b, c }).init(.c);
+        try testing.expectEqual(y.swap(.a, ordering), .c);
+        try testing.expectEqual(y.load(.SeqCst), .a);
+
+        var z = Atomic(f32).init(5.0);
+        try testing.expectEqual(z.swap(10.0, ordering), 5.0);
```
LemonBoy, line 279: Skip the `f32` tests for AAarch64 or the build will fail with `LLVM ERROR: Cannot select: intrinsic %llvm.aarch64.stxr`.

lib/std/atomic.zig
```diff
@@ -3,14 +3,82 @@
...
+
+        // `isb` appears more reliable for releasing execution resources than `yield` on common aarch64 CPUs.
+        // https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8258604
+        // https://bugs.mysql.com/bug.php?id=100664
+        .aarch64, .aarch64_be, .aarch64_32 => "isb",
+
+        // `yield` was introduced in v6k but is also available on v6m.
+        // https://www.keil.com/support/man/docs/armasm/armasm_dom1361289926796.htm
+        .arm, .armeb, .thumb, .thumbeb => blk: {
+            const can_yield = std.Target.arm.featureSetHasAny(target.cpu.features, .{ .has_v6k, .has_v6m });
```
LemonBoy, line 66:
```suggestion
            const can_yield = comptime std.Target.arm.featureSetHasAny(target.cpu.features, .{ .has_v6k, .has_v6m });
```

lib/std/atomic/Atomic.zig
LemonBoy: Now that #8922 is merged you can remove the AArch64 check, the tests should pass on both that and ARM now.

kprotty:
@LemonBoy Any further comments before it gets merged?

lib/std/atomic/Atomic.zig
```diff
@@ -0,0 +1,522 @@
...
+            compare: T,
+            exchange: T,
+            comptime success: Ordering,
+            comptime failure: Ordering,
+        ) callconv(.Inline) ?T {
+            if (success == .Unordered or failure == .Unordered) {
+                @compileError(@tagName(Ordering.Unordered) ++ " is only allowed on atomic loads and stores");
+            }
+
+            comptime var success_is_stronger = switch (failure) {
```
LemonBoy, line 87:
```suggestion
            comptime const success_is_stronger = switch (failure) {
```

LemonBoy (approved):
LGTM.
The size of type `T` is actually bounded by the combination of architecture and feature flags being used, an extra check can be added at a later stage though.

PR #8876, opened 2021-05-23
build libc++ with lto and function-sections 

xxxbxxx:
I was wondering why a simple c++ exe is so large.
I still don't  know why, but I did stumble on a few things along the way:

I noticed that unlike libc zig doesn't pass on lto and function-sections options when compiling libc++

And that all libc++ functions are exposed in the final library/exe (since all visibilty annotation is disabled), but I think they should be all hidden instead.

(maybe function-sections should always be activated, as is the case with libc?)

using (on linux-gnu):
```bash
$ zig build-exe test_cpp.cpp -lc++ --strip -OReleaseXXX  [-ffunction-sections]
```

```c++
#include <iostream>
int main()  {
	std::cout << "test cpp++!" << std::endl;
	return 0;
}
```

  x   | ReleaseSmall | ReleaseFast |
-----|--------------|------------
base    | 573472 | 694184 |
base -fno-lto | 574032 |   694728 |
lto | 372800 | 467568 | 
function-sections + no-lto | 467384 | 549208  |


andrewrk:
Thanks for digging into this!

PR #8877, opened 2021-05-23
expose header files when linking libunwind

xxxbxxx:
Make the library headers available when using `-lunwind` 

Moreover I notice that the library doesn't quite work on linux without `--eh-frame-hdr` maybe it should be enabled automatically?

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -2776,6 +2776,15 @@ pub fn addCCArgs(
         try argv.append("-D_LIBCPP_HAS_NO_VENDOR_AVAILABILITY_ANNOTATIONS");
     }
 
+    if (comp.bin_file.options.link_libunwind) {
+        const libunwind_include_path = try std.fs.path.join(arena, &[_][]const u8{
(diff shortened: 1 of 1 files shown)


andrewrk:
Thanks!

> Moreover I notice that the library doesn't quite work on linux without `--eh-frame-hdr` maybe it should be enabled automatically?

Hmm. I think that might make sense but I'm not 100% sure. Right now our logic for linking libunwind is that it is required for some targets when linking C++ code. But this logic is not privy to whether the C++ code has been compiled with `-fno-exceptions`, which I understand is fairly common.

A real world use case is the self-hosted compiler, with the LLVM backend enabled. In such case we depend on C++ code, but I don't think LLVM uses exceptions internally, so even though libunwind is a dependency, we don't want `--eh-frame-hdr`.

That said, I could see it being a reasonable default, something like this:

 * change `--eh-frame-hdr` to `-fno-eh-frame-hdr` and `-feh-frame-hdr`, so that Zig can choose a default, but the CLI flag can override it.
 * have some logic to enable it by default when linking libunwind on linux (and others?)

PR #8879, opened 2021-05-23
Add . and .. tests for std.fs functions

### LINKED ISSUE #4659: Relative paths on Windows: when to resolve . and .. ?
Related to #4605, #4655, #4658

From IRC (talking about #4655):
> `<daurnimator>` squeek502_: so realpath is inherently racy
> `<daurnimator>` you need to do it inside of the `open` call.
> `<daurnimator>` (on windows essentially everything is done by opening)
> `<daurnimator>` e.g. to delete a file you open a file; set a flag that says "delete this file when I close it" and then close the file handle
> `<squeek502_>` which open call?
> `<squeek502_>` the open within readFileAlloc?
> `<daurnimator>` yep
> `<squeek502_>` so you're saying readFileAlloc should handle the path resolution on its own?
> `<daurnimator>` yes
> `<squeek502_>` it does seem like there needs to be a decision about what's valid to pass to the various fs functions, its unclear if `.\file` is expected to always be a valid input to any fs function on Windows or if it should be resolved before certain fs functions
> `<daurnimator>` indeed

In the following test code:

```zig
const std = @import("std");

test "windows open file" {
    var first = try std.fs.cwd().openFile("test.zig", .{});
    first.close();

    var second = try std.fs.cwd().openFile(".\\test.zig", .{});
    second.close();
}
```

If `test.zig` exists, the first open succeeds, but the second fails, even though they would resolve to the same path. This is because the file path is never resolved and when `.\test.zig` is passed as the `ObjectName` to `NtCreateFile`, it errors with `STATUS_OBJECT_NAME_INVALID`. The same thing happens if a path with `..` is given as the `ObjectName`.

It's unclear where this resolution should happen, exactly, and the current `fs` docs don't specify which types of paths each function accepts.

squeek502:
These tests check that . and .. are able to be handled by the std.fs functions. Before #7664, these tests would have failed on Windows.

Closes #4659

EDIT: Note that this doesn't test the full `fs.Dir` API, but it does test all functions that have a `fooAbsolute` counterpart (except `updateFile` which I will file an issue about shortly).
EDIT#2: The updateFile thing was my mistake actually, will add it into this PR.

### DIFF
--- a/lib/std/fs/test.zig
+++ b/lib/std/fs/test.zig
@@ -932,3 +932,64 @@ test "walker" {
         try testing.expectEqualStrings(expected_dir_name, try fs.path.relative(allocator, tmp_path, entry.path));
     }
 }
+
+test ". and .. in fs.Dir functions" {
+    if (builtin.os.tag == .wasi) return error.SkipZigTest;
+
+    var tmp = tmpDir(.{});
+    defer tmp.cleanup();
+
+    try tmp.dir.makeDir("./subdir");
+    try tmp.dir.access("./subdir", .{});
+    var created_subdir = try tmp.dir.openDir("./subdir", .{});
+    created_subdir.close();
+
+    const created_file = try tmp.dir.createFile("./subdir/../file", .{});
+    created_file.close();
+    try tmp.dir.access("./subdir/../file", .{});
+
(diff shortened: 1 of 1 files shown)


marler8997:
More tests! Yay!

PR #8882, opened 2021-05-24
dragonfly: pass `zig build test`

mikdusan:
This PR adds the bits necessary for DragonFly BSD to pass zig tests. During development, dfly 5.8 and 6.0 were used, and 6.0 was used for final testing but 5.8 is expected to work fine.


andrewrk (approved):
Sweet!

PR #8890, opened 2021-05-24
Makes std.io.StreamSource usable with freestanding

MasterQ32:
This PR solves the problem that `std.io.StreamSource` is referencing `std.fs.File` which makes it unusable on freestanding environments. This PR `void`s the `.file` field on freestanding environments and makes the


MasterQ32:
CI failure seems unrelated?

LemonBoy:
> CI failure seems unrelated?

Unrelated but dangerous, #8924

Issue #7693, opened 2021-01-05
Proposal: `ubyte`/`ibyte` for smallest addressable unit of memory


EleanorNB:
Currently, it is assumed that all hardware uses `u8` as the address unit, which is not universally true. However, this assumption is built into the language or to things fairly central to the language's use in places, such as `@memcpy`/`@memset` and the allocator interface. The language has a concept of address units in `align`/`@sizeOf`, but provides no direct interface to this.

I propose the simplest solution: `ubyte` and `ibyte` types, representing the smallest addressable memory size, scaling with the platform just like `usize` does. So, anywhere in the language where "raw memory" is required would use `[*]ubyte` rather than `[*]u8`, and be portable universally.

Alternative names: `ucell`/`icell`, `unit`/`init` (but pls no)

kyle-github:
C has CHAR_BIT so perhaps something like that?   Perhaps a const `std.?cpu?.smallest_addressable_unit_bits`?   I **definitely** like the idea of being able to determine this in a well supported way, but I am not sure you need a type for it.  All the other types, except `usize` are explicitly not platform-platform specific.  This is just @floopfloopfloopfloopfloop's proposal in #5185.

SpexGuy:
`@memcpy` and `@memset` are part of the language, so the language does need to define a type for this.  Or we could use `*anyopaque` for those pointers instead.

EleanorNB:
> Or we could use `*anyopaque` for those pointers instead.

I want to say this won't work, since `anyopaque` can't be dereferenced without a cast (a cast to...), but I'm not intimately familiar.

SpexGuy:
It's fine since the language spec doesn't need to define the implementation of memcpy/memset, just the header.  The fact that you can't implement it in a portable way is unfortunate but not required from the language itself, the std lib could fill that gap.

jayschwa:
In Go, I do like having a distinction between `byte` and `uint8`. The latter signals that something mathy is happening, whereas the former is generally just unspecific "data".

One thing I don't understand about this proposal is why there should be two variants for signedness. Isn't having an opinion on the signedness of the value an indication that it should be an integer type?

EleanorNB:
It is an integer type. That's how we represent memory in Zig. We don't have a bit vector type, and I'm not proposing one.

daurnimator:
> It is an integer type. That's how we represent memory in Zig.

[...]

PR #8891, opened 2021-05-24
translate-c: Implement flexible arrays

### LINKED ISSUE #8759: Imported C Structs with VLAs
I'm currently using 0.8.0-dev.2192+65d279bc0 if that matters.

**Problem:**

I'd like to use [these C structs from ACPICA](https://github.com/acpica/acpica/blob/c1adb9a2a/source/include/actypes.h#L1378-L1406) in my Zig code. The following code is generated from it:

```zig
// (...)/actypes.h:1382:37: warning: struct demoted to opaque type - has variable length array
pub const struct_acpi_pnp_device_id_list = opaque {};
pub const ACPI_PNP_DEVICE_ID_LIST = struct_acpi_pnp_device_id_list;
pub const struct_acpi_device_info = extern struct {
    // ... fields that aren't important ..
    CompatibleIdList: ACPI_PNP_DEVICE_ID_LIST,
};
pub const ACPI_DEVICE_INFO = struct_acpi_device_info;
```

The warning seemed fair by itself when I first looked at this, but this apparently means that I can't use `ACPI_DEVICE_INFO`/`struct_acpi_device_info` at all:

```
./zig-cache/o/3016e146699c4abaeedfebe9c82141cf/cimport.zig:153:5: error: opaque types have unknown size and therefore cannot be directly embedded in structs
    CompatibleIdList: ACPI_PNP_DEVICE_ID_LIST,
    ^
```

Having a variable amount of data at the end of a data structure seems like a common pattern in data structures dealing with hardware. I've had to write code like [this here for USB EHCI](https://github.com/iguessthislldo/georgios/blob/37d86c0296131b7b5dcfeb0013015f809660ae1f/kernel/platform/usb.zig#L75-L77), among other places to access this kind of data from hardware. I'm fine writing code like that and it seems like I'm not the first person writing Zig to do so: #6008.

**Possible Solutions?**

I'm wondering if, for the sake of better compatibility with C, Zig could only demote the variable length array field itself to an opaque type and allow them at the end of the struct. This would [mirror how VLAs act in C](https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html). It might also make previously mentioned situation of using this in normal Zig more readable and reduce the amount of pointer arithmetic, but I think that's a minor advantage if it's one at all. Or maybe this should just should be limited to `extern struct`s if it is better to isolate this behavior?

```zig
const MyType = extern struct {
  length: usize,
  data: opaque {},
};
```

Alternatively it would be better from my perspective as a user if Zig fully supported VLAs in structs, but I'm assuming that's not on the table because normal VLAs have been shot down before: #3952.

**Also**

The actual `cimport.zig` lines with the warning are:

```zig
pub const ACPI_PNP_DEVICE_ID = struct_acpi_pnp_device_id; // /data/development/os/georgios/kernel/platform/acpica/acpica/source/include/actypes.h:1382:37: warning: struct demoted to opaque type - has variable length array
pub const struct_acpi_pnp_device_id_list = opaque {};
```

It's kinda weird the comment is following `ACPI_PNP_DEVICE_ID`, shouldn't there be a newline there or it should be following `struct_acpi_pnp_device_id_list`, which is what it's actually talking about?

ehaas:
Fixes #8759

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -819,6 +819,111 @@ fn transTypeDef(c: *Context, scope: *Scope, typedef_decl: *const clang.TypedefNa
     }
 }
 
+/// Build a getter function for a flexible array member at the end of a C struct
+/// e.g. `T items[]` or `T items[0]`. The generated function returns a [*c] pointer
+/// to the flexible array with the correct const and volatile qualifiers
+fn buildFlexibleArrayFn(
+    c: *Context,
+    scope: *Scope,
+    layout: *const clang.ASTRecordLayout,
+    field_name: []const u8,
+    field_decl: *const clang.FieldDecl,
+) TypeError!Node {
+    const field_qt = field_decl.getType();
+
+    const u8_type = try Tag.type.create(c.arena, "u8");
+    const self_param_name = "self";
+    const self_param = try Tag.identifier.create(c.arena, self_param_name);
+    const self_type = try Tag.typeof.create(c.arena, self_param);
+
+    const fn_params = try c.arena.alloc(ast.Payload.Param, 1);
+
+    fn_params[0] = .{
+        .name = self_param_name,
+        .type = Tag.@"anytype".init(),
+        .is_noalias = false,
+    };
+
+    const array_type = @ptrCast(*const clang.ArrayType, field_qt.getTypePtr());
+    const element_qt = array_type.getElementType();
+    const element_type = try transQualType(c, scope, element_qt, field_decl.getLocation());
+
+    var block_scope = try Scope.Block.init(c, scope, false);
+    defer block_scope.deinit();
+
+    const intermediate_type_name = try block_scope.makeMangledName(c, "Intermediate");
+    const intermediate_type = try Tag.std_meta_flexible_array_type.create(c.arena, .{ .lhs = self_type, .rhs = u8_type });
+    const intermediate_type_decl = try Tag.var_simple.create(c.arena, .{
+        .name = intermediate_type_name,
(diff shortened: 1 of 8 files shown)


iguessthislldo:
Thanks! This was blocking me, so I was just about to ask in the issue if I could take a crack at it if no one else was planning to. I will try this out later today.

Vexu (approved):

src/translate_c.zig
```diff
@@ -819,6 +819,111 @@ fn transTypeDef(c: *Context, scope: *Scope, typedef_decl: *const clang.TypedefNa
...
+    const return_type_decl = try Tag.var_simple.create(c.arena, .{
+        .name = return_type_name,
+        .init = return_type,
+    });
+    try block_scope.statements.append(return_type_decl);
+    const return_type_ident = try Tag.identifier.create(c.arena, return_type_name);
+
+    const field_index = field_decl.getFieldIndex();
+    const bit_offset = layout.getFieldOffset(field_index); // this is a target-specific constant based on the struct layout
+    const byte_offset = bit_offset / 8;
```
ehaas, line 874: This division assumes 8 bits per byte; is it worth using `toCharUnitsFromBits` to ensure it works on exotic platforms with different-sized bytes?

Vexu: I highly doubt this will ever become an issue.

tothambrus11: Yep, zig still hasn't decided to support non-octet bytes
https://github.com/ziglang/zig/issues/7693

PR #8892, opened 2021-05-24
fix position of `orelse` and `catch` in precedence table documentation

mattbork:
The documentation lists `orelse` and `catch` as having the second lowest precedence, above only assignment operators. However, in the grammar, they are grouped with `BitwiseOp`, and in parse.zig as well they have the same `.prec` of 40 that `&`, `^`, and `|` have. Further, there is a misplaced `!` included with the binary multiplication operators, which appears in neither the grammar nor parse.zig. I think this was once the error union type operator, but that has its own line below the suffix operators.

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -1758,9 +1758,9 @@ const B = error{Two};
       </div>
       {#header_close#}
       {#header_open|Precedence#}
-      <pre>{#syntax#}x() x[] x.y
+      <pre>{#syntax#}x() x[] x.y x.* x.?
 a!b
-x{} x.* x.?
+x{}
 !x -x -%x ~x &x ?x
 * / % ** *% ||
 + - ++ +% -%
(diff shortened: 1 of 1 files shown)


mattbork:
Looking at this more, I believe the `.*` and `.?` are also in the wrong place. In the grammar they are grouped with the `x[]` and `x.y` suffix operators under `SuffixOp`, and in parse.zig they are similarly parsed together in `parseSuffixOp`. Thus they should be together on the first line, above the error union type operator. This precedence actually can make a difference in odd uses like

```
comptime var T: ?type = u32;
const Error = error { Bad };
pub fn stuff() Error!T.? { return 100; }
```

The return type is parsed like `Error!(T.?)` rather than `(Error!T).?`, which is what the current precedence table implies. The initializer list braces should remain below the error union operator, as in something like  `var x = [_] XYZError!u32 { 100 };` the `!` does bind tighter than `{ 100 }`.

andrewrk:
Thanks!

One thing to keep in mind is #114

PR #8900, opened 2021-05-25
stage1: Store target info in the LLVM module

### LINKED ISSUE #8803: SIMD instructions generated with SIMD disabled
Somewhere between 0.7.1 and 0.8.0 (supposedly on LLVM 12 update), code which explicitly disables SIMD register operations all of a sudden start generating SIMD instructions:
```
0x401022ec MOVI            V0.16B, #0xAA
0x40102318 STP             Q0, Q0, [SP,#0x60+var_40]
0x4010231C STP             Q0, Q0, [SP,#0x60+var_60]
```

While reproducible, I've been unable to make a minimal reproducible example. The best I can offer is to use [Sabaton](https://github.com/FlorenceOS/Sabaton), run `zig build virt` see if the following to appear:

```
Taking exception 1 [Undefined Instruction]
...from EL1 to EL1
...with ESR 0x7/0x1fe0000a
...with ELR 0x401022ec
...to EL1 PC 0x40101800 PSTATE 0x3c5
Taking exception 4 [Data Abort]
```

I've been able to narrow it down to storing 64+ bytes of `0xAA`(undefined fill) for variables on the stack. Could honestly be an LLVM issue for all I know.

LemonBoy:
This is needed to let LLVM (or, better, libLTO) produce code using the
target options specified by the user.

Fix #8803

### DIFF
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -589,6 +589,13 @@ static LLVMValueRef make_fn_llvm_value(CodeGen *g, ZigFn *fn) {
         // use the ABI alignment, which is fine.
     }
 
+    if (g->zig_target->llvm_cpu_name != nullptr) {
+        ZigLLVMAddFunctionAttr(llvm_fn, "target-cpu", g->zig_target->llvm_cpu_name);
+    }
+    if (g->zig_target->llvm_cpu_features != nullptr) {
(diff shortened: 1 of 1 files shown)


andrewrk:
Every single function needs this data?

LemonBoy:
Yeah, that's also what clang does and is apparently needed for cross-language inlining to work.

andrewrk:
Alrighty then, thanks for the fix!

andrewrk (approved):

PR #8909, opened 2021-05-26
p256: update to the last fiat-crypto code & share PC tables

jedisct1:
fiat-crypto now generates proper types, so take advantage of that.

Add mixed subtraction and double base multiplication.

We will eventually leverage mixed addition/subtraction for fixed base multiplication. The reason we don't right now is that
precomputing the tables at comptime would take forever.

We don't use combs for the same reason. Stage2 + less function calls in the fiat-crypto generated code will eventually address that.

Also make the edwards25519 code consistent with these changes.

No functional changes.


lib/std/crypto/pcurves/p256.zig
```diff
@@ -395,33 +400,69 @@ pub const P256 = struct {
...
+            pc2_array = precompute(p2, 8);
+            break :pc &pc2_array;
+        };
+        const e1 = slide(s1);
+        const e2 = slide(s2);
+        var q = P256.identityElement;
+        var pos: usize = 2 * 32 - 1;
+        while (true) : (pos -= 1) {
+            const slot1 = e1[pos];
+            if (slot1 > 0) {
```
Mouvedia, line 450: pseudo code
```
            const slot1 = e1[pos];
            const slot2 = e2[pos];
            if (slot1 != 0) {
                q = q.sub(pc1[@intCast(usize, abs(slot1))]);
            }
            if (slot2 != 0) {
                q = q.add(pc2[@intCast(usize, abs(slot2))]);
            }
```

That's shorter but I don't know if you would consider it more readable.

jedisct1: That wouldn't be equivalent :) We need to call `.add()` or `.sub()` according to the sign, for both points.

PR #8910, opened 2021-05-27
Use shared scratch buffer to avoid allocating and freeing many small lists in parse.zig

mattbork:
The "permanent" data allocated by `Parser` during parsing is done through the two large, shared `ArrayList` instances, `nodes` and `extra_data`. This certainly reduces the number of separate allocations as compared to naively allocating nodes one at time. Unfortunately, parsing functions also need to collect variable-length lists of nodes. This cannot be done directly into `extra_data` because of the nesting of calls to other parsing functions that may need to do the same, breaking contiguity. Exhaustively, `parseContainerMembers` collects members, `parseBlock` collects statements, `parseCurlySuffixExpr` and `parsePrimaryTypeExpr` collect initializers, `parseSuffixExpr` collects suffix operator parameters, `expectAsmExpr` collects input and output items, `parseSwitchProng` collects case items, and `parseParamDeclList` collects parameters. Thus a large number of allocations and frees of temporary ArrayLists may be done even during the parsing of small files. These separate allocations can be avoided.

The greater set of small allocations are temporary lists of `Node.Index` which are never returned but rather passed to `listToSpan` to be copied into permanent storage in `extra_data`. It is only indices into `extra_data` that are ever stored in nodes. A parsing function building a temporary list may call other parsing functions that build nested lists but because the lists are always copied to `extra_data` and then freed before returning, the lifetimes of temporary lists are strictly nested within the lifetimes of frames on the callstack. Therefore, just like the stack itself, all parsing functions can build these temporary lists in one shared scratch buffer. When a parsing function wants to start a list, it needs only to remember the current top of the scratch buffer and to add a `defer` statement that will shrink the top of the scratch buffer back to that point when returning. When the parsing function is ready to turn the list into `extra_data`, it simply calls `listToSpan` with the tail of the scratch buffer that starts from that remembered position. 

The second set of small allocations is the one exception to the strict nesting above. `parseParamDeclList`, when parsing more than two parameters, will allocate a temporary `ArrayList`. However, that memory is not freed before returning. Instead, it is wrapped in a `SmallSpan` and returned to its caller, who later frees it. Hence, to use a shared scratch buffer, the caller must remember the top of scratch buffer before calling `parseParamDeclList` and set a `defer` to shrink the scratch buffer back down, just as if it were building a temporary list itself. The only use of `parseParamDeclList` is in `parseFnProto`, which does not otherwise create a temporary list. Still, a parsing function that did need to build its own list could call a function like `parseParamDeclList` as long as two top positions were remembered: one for its own work and one immediately before the call to a function that leaves data on the scratch buffer. When used this way, `SmallSpan` can be removed entirely and `parseParamDeclList` returns a slice of `Node.Index` directly. This is how I've implemented it currently. Alternatively, to be more explicit, `SmallSpan` could be retained and rewritten to shrink the scratch buffer in its `deinit`.

(**EDIT**: Thinking about this more, there's no real reason for a temporary list to escape `parseParamDeclList`. If one needs to be built, it should be copied to `extra_data` and freed before returning. Then the `SmallSpan` union is still the return value of `parseParamDeclList`, only the `.multi` field is no longer owned memory but simply the `Node.SubRange` returned by `listToSpan`. This means the lifetimes of all temporary lists are again nested within frames on the callstack.)

These changes together eliminate all small, temporary allocations. For example, parsing std/mem.zig currently does 1025 allocations (counting new allocations and successful enlargements of existing allocations) and 984 frees. With these changes, it does only 25 allocations and 12 frees. For testing, I parsed every .zig file in this repository (excepting those that already gave parse errors before any of these changes) with and without these changes and checked that exactly the same `Tree` was produced. I did one parse with a counting allocator wrapped around a fixed buffer allocator to keep track of allocations, frees, and total memory use. Then (following the example of `perf_test.zig`) I parsed each file that is at least 5KiB another 100 times and recorded timings in bytes-of-source per second (below 5KiB, there's too much noise in the result, I think). This part I did once linking against libc and using the C allocator (on Windows), and once using the `GeneralPurposeAllocator` without linking libc.

These changes generally result in small savings in memory use, in the 10s of KiB for large files. At worst, an extra 728 bytes are used. When linking against libc and using the C allocator, there's a modest performance gain, often 5% - 10%. Using the `GeneralPurposeAllocator` is a different story, and the performance improvements for the largest files are dramatic. For example, src/AstGen.zig went from 2279 allocs and 2111 frees to 28 and 8, respectively, and the approximate parsing rate went from 28.94MiB/s to 95.87MiB/s for an improvement of 231%. I'm not too comfortable with drawing conclusions from performance data, so I've also attached the raw results. Each file lists allocs and frees, memory use, and bytes per second for all files >= 5KiB, in descending order of file size:

- [results-gpa.txt](https://github.com/ziglang/zig/files/6550707/results-gpa.txt), using the `GeneralPurposeAllocator`
- [results-libc.txt](https://github.com/ziglang/zig/files/6550708/results-libc.txt), linking libc and using the C allocator

One criticism I can make is that when using the scratch buffer, parsing functions no longer have a name for the lists they're accumulating. For example, in `parseBlock` lines like `try statements.append(statement)` are replaced by `try p.scratch.append(p.gpa, statement)`, which are less self-documenting. Many other instances, however, already just used the generic name `list`. Still, pointers like `var statements = &p.scratch` could be used to communicate intent better. Or perhaps use of the scratch buffer could be wrapped in a struct that records `scratch_top`, has `append` and `items` functions, and shrinks the scratch buffer back down in its `deinit`. There may also be better names for `scratch` and `scratch_top` themselves.

This was an experiment and exercise for me while trying to ease in to learning more about the compiler itself. I think the results are good, but I'll have no hard feelings if it's decided that these changes hurt clarity too much.

### DIFF
--- a/lib/std/zig/parse.zig
+++ b/lib/std/zig/parse.zig
@@ -43,11 +43,13 @@ pub fn parse(gpa: *Allocator, source: []const u8) Allocator.Error!Tree {
         .errors = .{},
         .nodes = .{},
         .extra_data = .{},
+        .scratch = .{},
         .tok_i = 0,
     };
     defer parser.errors.deinit(gpa);
     defer parser.nodes.deinit(gpa);
     defer parser.extra_data.deinit(gpa);
+    defer parser.scratch.deinit(gpa);
 
     // Empirically, Zig source code has a 2:1 ratio of tokens to AST nodes.
     // Make sure at least 1 so we can use appendAssumeCapacity on the root node below.
@@ -93,18 +95,7 @@ const Parser = struct {
     errors: std.ArrayListUnmanaged(AstError),
     nodes: ast.NodeList,
     extra_data: std.ArrayListUnmanaged(Node.Index),
-
-    const SmallSpan = union(enum) {
-        zero_or_one: Node.Index,
-        multi: []Node.Index,
-
-        fn deinit(self: SmallSpan, gpa: *Allocator) void {
-            switch (self) {
-                .zero_or_one => {},
-                .multi => |list| gpa.free(list),
-            }
-        }
-    };
+    scratch: std.ArrayListUnmanaged(Node.Index),
 
     const Members = struct {
         len: usize,
@@ -204,8 +195,8 @@ const Parser = struct {
     ///      /
     /// TopLevelComptime <- KEYWORD_comptime BlockExpr
     fn parseContainerMembers(p: *Parser) !Members {
-        var list = std.ArrayList(Node.Index).init(p.gpa);
-        defer list.deinit();
+        const scratch_top = p.scratch.items.len;
+        defer p.scratch.shrinkRetainingCapacity(scratch_top);
 
         var field_state: union(enum) {
             /// No fields have been seen.
@@ -233,7 +224,7 @@ const Parser = struct {
                         if (field_state == .seen) {
                             field_state = .{ .end = test_decl_node };
                         }
-                        try list.append(test_decl_node);
+                        try p.scratch.append(p.gpa, test_decl_node);
                     }
                     trailing = false;
                 },
@@ -254,7 +245,7 @@ const Parser = struct {
                                     field_state = .err;
                                 },
                             }
-                            try list.append(container_field);
+                            try p.scratch.append(p.gpa, container_field);
                             switch (p.token_tags[p.tok_i]) {
                                 .comma => {
                                     p.tok_i += 1;
@@ -294,7 +285,7 @@ const Parser = struct {
                             if (field_state == .seen) {
                                 field_state = .{ .end = comptime_node };
                             }
-                            try list.append(comptime_node);
+                            try p.scratch.append(p.gpa, comptime_node);
                         }
                         trailing = false;
                     },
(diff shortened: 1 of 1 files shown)


mattbork:
I've made the change mentioned in the **EDIT** above to keep temporary lists from escaping `parseParamDeclList`. This makes all uses of the scratch buffer for temporary lists follow the same pattern, and simplifies `parseFnProto` a little.

andrewrk (approved):
Brilliant work! Really appreciate the detailed writeup, comprehensive testing, and perf results. The code changes seem very reasonably maintainable to me, and this looks like a non-trivial parsing speed boost. Definitely worth it.

It always tickles me to see something like this:

```
  557,   522 ->  21,   9 allocs, frees | 346.8KiB -> 335.5KiB =  -11.3KiB |    103.70MiB/s ->    111.72MiB/s =   +7.74% | lib\std\zig\parse.zig
```

Fun to know that this code made itself parse +8% faster :grinning:

mattbork:
Thanks for reviewing this so quickly! I'm really glad this worked out.

PR #8912, opened 2021-05-27
build: allow specifying rpaths explicitly

kubkon:
It currently looks like that if the user links in a dylib using
`lib_or_exe.linkSystemLibrary`, and the linked lib doesn't have
a hardcoded path in its description load command but rather it
allows for any runtime path via `@rpath`, then it is not possible
to specify the runtime path explicitly using the build system.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -1341,6 +1341,7 @@ pub const LibExeObjStep = struct {
     name_only_filename: []const u8,
     strip: bool,
     lib_paths: ArrayList([]const u8),
+    rpaths: ArrayList([]const u8),
     framework_dirs: ArrayList([]const u8),
     frameworks: BufSet,
     verbose_link: bool,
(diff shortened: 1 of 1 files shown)


lib/std/build.zig
```diff
@@ -2574,6 +2580,11 @@ pub const LibExeObjStep = struct {
         }
 
         if (self.target.isDarwin()) {
```
andrewrk, line 2582: The whole patch looks good except I don't think we need the `isDarwin()` here. Other OS's have the concept of rpath too.

andrewrk (changes requested):

lib/std/build.zig
kubkon: Excellent point, thanks Andrew! I actually honestly thought the concept of `rpath`s is darwin specific :-)

kubkon: Done in 79d460a

PR #8917, opened 2021-05-27
stage1, stage2: disallow 1.e9 and 0x1.p9 as float literals

ifreund:
Instead require `1e9` and `0x1p9`, disallowing the trailing dot.

This change to the grammar is consistent with forbidding `1.` and `0x1.`
as float literals and ensures there is only one way to do things here.

This is technically a proposal, but has already been informally approved by andrew on IRC:
```
19:50 <ifreund> andrewrk: should `1.e100` be vaild or not? right now the grammar allows it but we could also force `1e100`
19:50 <andrewrk> ifreund, seems like to be consistent with `1.` being an error, `1.e100` should also be an error right?
19:50 <freshcoffee> yeah if `1e100` is accepted then `1.e100` is purely superfluous 
19:51 <ifreund> I agree, just wanted to check :)
```


ifreund:
Looks like `std.fmt` may currently print floats using this superfluous dot, I'll look into fixing that.

ifreund:
Thanks to #8918, backporting this to stage1 was trivial :).

PR #8918, opened 2021-05-28
stage1: rework tokenizer to match stage2

andrewrk:
* Extracts AstGen logic from ir.cpp into astgen.cpp. Reduces the
   largest file of stage1 from 33,551 lines to 25,510.
 * tokenizer: rework it completely to match the stage2 tokenizer logic.
   They can now be maintained together; when one is changed, the other
   can be changed in the same way.
   - Each token now takes up 13 bytes instead of 64 bytes. The tokenizer
     does not parse char literals, string literals, integer literals,
     etc into meaningful data. Instead, that happens during parsing or
     astgen.
   - no longer store line offsets. Error messages scan source
     files to find the line/column as needed (same as stage2).
   - main loop: instead of checking the loop, handle a null byte
     explicitly in the switch statements. This is a nice improvement
     that we may want to backport to stage2.
   - delete some dead tokens, artifacts of past syntax that no longer
     exists.
 * Parser: fix a TODO by parsing builtin functions as tokens rather than
   `@` as a separate token. This is how stage2 does it.
 * Remove some debugging infrastructure. These will need to be redone,
   if at all, as the code migrates to match stage2.
   - remove the ast_render code.
   - remove the IR debugging stuff
   - remove the token printing code

Apologies for not splitting the ir.cpp/astgen.cpp stuff in a separate commit. That was really sloppy of me.

The motivation for this change is twofold:
 * To bring stage1 codebase closer to stage2, making it easier to maintain both at the same time. This is now true, at least as far as the tokenizer goes.
 * To bring some of the stage2 performance improvements to stage1, in the hopes of improving compilation speed of stage1 and reducing memory requirements. The reality is that even though stage2 is making great progress, development on stage2 itself is slowed down by stage1, and contributors such as @LemonBoy and @ifreund are bogged down by the unreasonable RAM requirements of stage1.

So, how does this change fare?

Here are three measurements - behavior tests, standard library test suite, and building stage2:

```
master branch:
	Command being timed: "./zig test ../lib/std/std.zig"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:36.06
	Maximum resident set size (kbytes): 8534820 (8.14 GiB)

	Command being timed: "./zig test ../test/behavior.zig -I ../test"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:04.13
	Maximum resident set size (kbytes): 1061312 (1036 MiB)

	Command being timed: "./zig build -p zig-out -Dlog -Dskip-install-lib-files -Denable-llvm"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:52.58
	Maximum resident set size (kbytes): 9558500 (9.12 GiB)


stage1-tokenize branch:
	Command being timed: "./zig test ../lib/std/std.zig"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:39.34
	Maximum resident set size (kbytes): 8668804 (8.27 GiB)

	Command being timed: "./zig test ../test/behavior.zig -I ../test"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:04.12
	Maximum resident set size (kbytes): 1019028 (995 MiB)

	Command being timed: "./zig build -p zig-out -Dlog -Dskip-install-lib-files -Denable-llvm"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:54.73
	Maximum resident set size (kbytes): 9657708 (9.21 GiB)
```

For std lib tests and building stage2 with stage1, this is a performance regression, both in terms of speed and peak memory consumption. Sad trombone. 

However for behavior tests this is a slight improvement on both counts, and I think this illuminates the next thing that can be done to address stage1 perf. stage1 has serious design flaw, which is that not only is astgen per-decl rather than per-file, it's also repeated for each generic instantiation, and each comptime function call. I believe this explains the regression for std lib tests - string literals are now allocated in astgen rather than the tokenizer, which means if a function call happens N times then each string literal inside the function's body gets allocated N times, where previously it was 1. I suspect that std lib tests rely more on use of long-executing comptime code blocks than behavior tests do.

I don't think this means that these changes are worse; I think it reveals more strongly how big a problem it already was for astgen to be repeated for every comptime function call. I also think it would be a reasonable investment into stage1 to change it so that astgen happens only once per decl, and comptime function calls re-use the same astgen output. Migrating straight to whole-file-astgen in stage1 would probably be too much to take on at once, but a good step forwards would be making the output of astgen immutable and only doing it once per decl.

Also, the parser of stage1 uses a tagged union (!!) for every AST node, wasting a ton of memory. This work on the tokenizer doesn't even have the chance to shine when the parser is wasting memory like this.

So despite the measurements coming out poorly, I'm still confident this is a step in the right direction. However probably a bit more work should be done in stage1 before merging to make this not be a perf regression.


andrewrk:
OK the next problem to solve on the goal of making `Stage1Zir` immutable is this struct:

```c++
struct Stage1ZirBasicBlock {
    ZigList<IrInstSrc *> instruction_list;
    IrBasicBlockGen *child;
    Scope *scope;
    const char *name_hint;
    IrInst *suspend_instruction_ref;

    uint32_t ref_count;
    uint32_t index; // index into the basic block list

    uint32_t debug_id;
    bool suspended;
    bool in_resume_stack;
};
```

In particular the `suspend_instruction_ref` field is mutated inside ir.cpp (AKA Stage1Sema) by the core semantic analysis system. This is buggy and poorly designed (by me) so it's a bit tricky to edit the code. Likewise the `suspended` and `in_resume_stack` flags are mutated by the same system.

So the key to reclaiming memory in stage1 is to untangle this. If the result location / semantic analysis logic is too hard to dissect, the state can be put into a hash map inside `IrAnalyze` rather than mutating `Stage1ZirBasicBlock` directly. This will give a straightforward way to port the code over, without necessitating reworking the logic, and yet keeping `Stage1ZirBasicBlock` immutable, so that we can avoid allocating it over and over again.

andrewrk:
With that last commit, there is a small workaround to address the memory usage regression, and we actually end up on top, overall:

```
master:
	Command being timed: "./zig test ../lib/std/std.zig"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:36.34
	Maximum resident set size (kbytes): 8534944 (8.14 GiB)

	Command being timed: "./zig test ../test/behavior.zig -I ../test"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:04.25
	Maximum resident set size (kbytes): 1061576 (1.01 GiB)

	Command being timed: "./zig build -p zig-out -Dlog -Dskip-install-lib-files -Denable-llvm"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:54.54
	Maximum resident set size (kbytes): 9558616 (9.14 GiB)


stage1-tokenize:
	Command being timed: "./zig test ../lib/std/std.zig"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:38.41
	Maximum resident set size (kbytes): 8396308 (8.01 GiB)

	Command being timed: "./zig test ../test/behavior.zig -I ../test"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:04.14
	Maximum resident set size (kbytes): 991696 (0.95 GiB)

	Command being timed: "./zig build -p zig-out -Dlog -Dskip-install-lib-files -Denable-llvm"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:54.26
	Maximum resident set size (kbytes): 9407716 (8.97 GiB)
```

So this branch is now a 1.6% reduction in peak memory usage, which is pretty negligible, but at least in the right direction. And the tokenization simplification has already been a helpful improvement which you can see based on how easy it was to backport a recent tokenization fix to stage1 in 9de2c1b76e2b38e35bb17cd0a3943a61c730da76.

My goals for this branch are met; I will now merge as soon as the tests are passing.

For anyone wishing to push on this further: the next step is outlined here: https://github.com/ziglang/zig/pull/8918#issuecomment-850124990

**The stage1 parser cannot be reworked to match stage2 until stage1 astgen is reworked to only run once per decl**. So the next thing to do is to untangle astgen from sema in stage1, making astgen happen based only on the AST, just like how stage2 does it. Once that is in place, we have some more easy wins in terms of memory and perf by converting the parser to match stage2.

Issue #8920, opened 2021-05-28
stage2 crash in astgen due to errors getting ignored


g-w1:
This crashes astgen: 
```zig
fn ChaChaLetsCrashAstGen() type {
    return struct {
        pub fn AHHHHHHHHHHHHHHHHHHHHHHHHHHHHH() void {
            const bruh = 1091290;
            const bruh = 1091290;
        }
    };
}
```
```
❯ ./zig ast-check test.zig                                                                                                                                                                                                                   
thread 4039 panic: reached unreachable code
/home/jacob/dev/zig/src/AstGen.zig:2137:27: 0xee4458 in AstGen.genDefers (zig1)
            .namespace => unreachable,
                          ^
/home/jacob/dev/zig/src/AstGen.zig:6013:22: 0xdba900 in AstGen.ret (zig1)
        try genDefers(gz, &astgen.fn_block.?.base, scope, .none);
                     ^
/home/jacob/dev/zig/src/AstGen.zig:628:33: 0xc98c9b in AstGen.expr (zig1)
        .@"return" => return ret(gz, scope, node),
                                ^
/home/jacob/dev/zig/src/AstGen.zig:1821:41: 0xeeece4 in AstGen.unusedResultExpr (zig1)
    const maybe_unused_result = try expr(gz, scope, .none, statement);
                                        ^
/home/jacob/dev/zig/src/AstGen.zig:1809:41: 0xee8a7b in AstGen.blockExprStmts (zig1)
            else => try unusedResultExpr(gz, scope, statement),
                                        ^
/home/jacob/dev/zig/src/AstGen.zig:1646:23: 0xdc1b7d in AstGen.blockExpr (zig1)
    try blockExprStmts(gz, scope, block_node, statements);
                      ^
/home/jacob/dev/zig/src/AstGen.zig:718:33: 0xc9d009 in AstGen.expr (zig1)
                return blockExpr(gz, scope, rl, node, statements[0..1]);
                                ^
/home/jacob/dev/zig/src/AstGen.zig:2925:25: 0xc9245d in AstGen.fnDecl (zig1)
            _ = try expr(&fn_gz, params_scope, .none, body_node);
                        ^
/home/jacob/dev/zig/src/AstGen.zig:3390:38: 0xb5541c in AstGen.structDeclInner (zig1)
[...]

PR #8921, opened 2021-05-28
stage2: detect ambiguous decl references

g-w1:
This also found some compile errors in std, so I fixed them.
Putting this as a separate pr as my work on unused vars as #8920 is blocking that, but not this.

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -2355,6 +2355,7 @@ fn varDecl(
                 .name = ident_name,
                 .ptr = init_scope.rl_ptr,
                 .token_src = name_token,
+                .is_comptime = true,
             };
             return &sub_scope.base;
         },
@@ -2410,6 +2411,7 @@ fn varDecl(
                 .name = ident_name,
                 .ptr = var_data.alloc,
                 .token_src = name_token,
+                .is_comptime = is_comptime,
             };
             return &sub_scope.base;
         },
@@ -3354,7 +3356,7 @@ fn structDeclInner(
     };
     defer block_scope.instructions.deinit(gpa);
 
-    var namespace: Scope.Namespace = .{ .parent = &gz.base };
+    var namespace: Scope.Namespace = .{ .parent = scope };
     defer namespace.decls.deinit(gpa);
 
     var wip_decls: WipDecls = .{};
@@ -5347,6 +5349,7 @@ fn forExpr(
             .name = index_name,
             .ptr = index_ptr,
             .token_src = index_token,
+            .is_comptime = parent_gz.force_comptime,
         };
         break :blk &index_scope.base;
     };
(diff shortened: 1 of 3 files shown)


g-w1:
I think this does not compile because of a stage1 scoping bug?

test/stage2/test.zig
```diff
@@ -938,12 +938,25 @@ pub fn addCases(ctx: *TestContext) !void {
...
+        \\    const T = struct {
+        \\        fn f() void {
+        \\            _ = T;
+        \\        }
+        \\    };
+        \\};
+    , &.{
+        ":4:17: error: ambigious reference",
+        ":1:1: note: first reference here",
+        ":2:5: note: second reference here",
```
andrewrk, line 957: A reference is what is happening on line 4; these are declarations:

```suggestion
        ":1:1: note: declared here",
        ":2:5: note: also declared here",
```

src/AstGen.zig
```diff
@@ -6097,7 +6098,21 @@ fn identifier(
             },
             .gen_zir => s = s.cast(GenZir).?.parent,
             .defer_normal, .defer_error => s = s.cast(Scope.Defer).?.parent,
-            .namespace, .top => break, // TODO look for ambiguous references to decls
+            // look for ambiguous references to decls
+            .namespace => {
```
andrewrk, line 6102: Can you add a test case that ensures that this logic has the correct behavior, in the case that this loop escapes out of the namespace, and then there is a local_val or local_ptr with a matching name?

Something like this:

```zig
test {
    var bar: i32 = 1234;
    const S = struct {
        fn baz() i32 {
            return bar;
        }
    };
    var quux = S.baz();
}
```

andrewrk (changes requested):

src/AstGen.zig
```diff
@@ -2355,6 +2355,7 @@ fn varDecl(
                 .name = ident_name,
                 .ptr = init_scope.rl_ptr,
                 .token_src = name_token,
+                .is_comptime = true,
```
g-w1, line 2358: Is this correct?

src/AstGen.zig
```diff
@@ -5347,6 +5349,7 @@ fn forExpr(
             .name = index_name,
             .ptr = index_ptr,
             .token_src = index_token,
+            .is_comptime = parent_gz.force_comptime,
```
g-w1, line 5352: Is this correct?

src/AstGen.zig
g-w1: I put this as true since we have no way to know in astgen if a const is passable to lower scopes or not, so we pass all of it off to sema (with decl val ofc).

Vexu: That seems like the best approach.

src/AstGen.zig
Vexu: This should use `is_inline` since the index is also comptime known for inline loops. Also sorry for not answering sooner.

Vexu: Btw @andrewrk, @SpexGuy What should the correct behavior be for
```zig
pub fn main() void {
    comptime var i = 0;
    inline while (i < 5) : (i+=1) {
        const S = struct {
            fn print() void {
                std.debug.print("loop {d}\n", .{i});
            }
        };
        S.print();
    }
}
```

Stage1 prints
```
loop 5
loop 5
loop 5
loop 5
loop 5
```
but I think most people would assume it to print
```
loop 0
loop 1
loop 2
loop 3
loop 4
```

Vexu (approved):

PR #8929, opened 2021-05-28
Add `std.io.Writer.printAndCount`

haze:
This function is a convenience method to count how many bytes were printed to a stream. This is accomplished by wrapping the current
writer with a `std.io.CountingWriter` and returning the amount written.

### DIFF
--- a/lib/std/io/writer.zig
+++ b/lib/std/io/writer.zig
@@ -34,6 +34,12 @@ pub fn Writer(
             return std.fmt.format(self, format, args);
         }
 
+        pub fn printAndCount(self: Self, comptime format: []const u8, args: anytype) Error!u64 {
+            var countingWriter = std.io.countingWriter(self);
+            try countingWriter.writer().print(format, args);
+            return countingWriter.bytes_written;
+        }
+
         pub fn writeByte(self: Self, byte: u8) Error!void {
             const array = [1]u8{byte};
             return self.writeAll(&array);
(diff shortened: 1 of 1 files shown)


andrewrk:
Test coverage please! Implementation looks good.

lib/std/io/writer.zig
```diff
@@ -34,6 +34,12 @@ pub fn Writer(
             return std.fmt.format(self, format, args);
         }
 
+        pub fn printAndCount(self: Self, comptime format: []const u8, args: anytype) Error!u64 {
```
andrewrk, line 37: Should be `usize` to match `write`:

```suggestion
        pub fn printAndCount(self: Self, comptime format: []const u8, args: anytype) Error!usize {
```

If you want to make an argument that they should all be changed to `u64` we can have that discussion but should be a different issue.

andrewrk (changes requested):

lib/std/io/writer.zig
haze: Now this begs the question... what if we print something that is larger than the platforms `usize`?

andrewrk: How would you do that?

daurnimator: in 2 (or really: many) write calls.

e.g. imagine you have a `[1024*1024*1024*2]u8` (2GB array) and you print it with `{}` ==> you'd print off 3-5 characters per entry: for a total of 6-10GB of output.

andrewrk: In Zig we have the precedent that `usize` + `usize` is still another `usize`:

```zig
test {
    var a: usize = foo();
    var b: usize = bar();
    var c = a + b; // the type will be usize
    var d = a * b; // even for this one
}
```

same idea here with `write()` multiple times.

haze: Of course. I think the issue here is that printing can lead to a number that may be more than what a platform decides a `usize` is

haze:
```
[01:27:11]  <andrewrk> haze, I have an even better idea, no printAndCount method. when it's debatable what the right way to do it is, there's an easy solution: no code at all
```

Issue #8726, opened 2021-05-09
completely eliminate dependency on LLD


andrewrk:
In celebration of @kubkon's recent accomplishments of improving the self-hosted linker on Mach-O enough to be able to link LLVM, Clang, LLD, and Zig, this issue is to completely rely on our own linker implementation for all targets, and entirely stop linking against LLD. This issue is the spiritual successor to  #1535.

This issue does not block the release of Zig 1.0, however it is a stretch goal, and a fun indicator of progress to keep track of along the way.

Here are all the places we depend on LLD, along with an indicator of how much our own linker code can do:

* :yellow_circle: ELF:
  * :yellow_circle: x86, x86_64
  * :yellow_circle: arm, aarch64
  * :yellow_circle: riscv
  * :red_circle: hexagon
  * :red_circle: mips
  * :red_circle: ppc, ppc32
  * :red_circle: sparc
  * :red_circle: amdgpu
  * :red_circle: avr
  * :red_circle: msp430
* :green_circle: Mach-O
  * :green_circle: x86_64
  * :green_circle: aarch64
* :yellow_circle: COFF/PE
  * :yellow_circle: x86, x86_64
  * :red_circle: arm, aarch64
* :yellow_circle: WASM

#### Key

:green_circle: - major progress
:yellow_circle: - some progress
:red_circle: - no progress

#### Other Considerations

* miscellaneous features such as linker scripts (related: #3206)
* LTO. For our purposes, this would look like suporting LLVM IR bitcode files and using LLVM APIs to turn them into object(s) before completing the linking process.

PR #8932, opened 2021-05-29
cmake: LLVM needs to link against zlib

andrewrk:
Without this, LLVM is crippled. This makes building Zig on Windows more
involved and is probably going to lead to more issues filed.

For more details see
https://github.com/ziglang/zig-bootstrap/issues/57


daurnimator:
Is this a static-linking-only dependency? (i.e. the dynamic LLVM library would already depend on libz?)

andrewrk:
Yes

CMakeLists.txt
```diff
@@ -713,11 +713,14 @@ set_target_properties(zigcpp PROPERTIES
     COMPILE_FLAGS ${EXE_CFLAGS}
 )
 
+find_library(ZLIB NAMES libz.a z zlib libz)
```
daurnimator, line 716: Why is `libz.a` first? shouldn't the following `z` find it anyway?

ifreund: I copied this line from another place in the `CMakeLists.txt` in order to quickly get this working to test the zig-bootstrap changes. This could probably be cleaned up (I'm no cmake expert) and should probably be put behind some `ZIG_USE_ZLIB` flag which is automatically enabled if `ZIG_STATIC` is true.

kubkon:
One thing to keep in mind here: if this forces linking of `-lz` on macOS and LLVM links against symbols from `libz`, then macOS binary will be malformed. This will be until I add ability to parse "tbd"s in zld. Until then, might it make sense to feature-gate `libz` so that it is not pulled into macOS?

kubkon:
> One thing to keep in mind here: if this forces linking of `-lz` on macOS and LLVM links against symbols from `libz`, then macOS binary will be malformed. This will be until I add ability to parse "tbd"s in zld. Until then, might it make sense to feature-gate `libz` so that it is not pulled into macOS?

I just read https://github.com/ziglang/zig-bootstrap/pull/58 and since we are building `libz` ourselves and linking it in statically, I think we should be good on macOS without any special feature-gating or whatnot. It would only be a problem if `libz` was a system lib on a current macOS and only a tbd was available. Then, we would not correctly link `libz` since `zig ld` currently doesn't understand text-based definition files.

nektro:
this is also presumably only a temporary measure until #8726 is complete

andrewrk:
I really don't understand how this is causing a segfault on freebsd while making aarch64-linux print "zig0: not found"

andrewrk:
```
(lldb) run src/stage1.zig -target x86_64-freebsd-gnu -mcpu=baseline --name zig1 --override-lib-dir /usr/home/build/zig/lib -femit-bin=/usr/home/build/zig/build/zig1.o -OReleaseFast --strip -lc --pkg-begin build_options /usr/home/build/zig/build/config.zig --pkg-end --pkg-begin compiler_rt /usr/home/build/zig/lib/std/special/compiler_rt.zig --pkg-end
Process 1585 launching
Process 1585 launched: '/usr/home/build/zig/build/zig0' (x86_64)
Process 1585 stopped
* thread #1, name = 'zig0', stop reason = signal SIGSEGV: invalid address (fault address: 0x0)
    frame #0: 0x0000000000000000
error: Bad address
(lldb) up
error: Already at the top of the stack.
```

:man_shrugging:

andrewrk:
I'm guessing that we're incorrectly finding zlib on these systems but it's invalid for whatever reason and if we provide our own build it will work fine. Just need to update the CI tarballs to test this.

andrewrk:
I have no idea why drone CI is failing. It works fine with the bootstrap repository.

mikdusan:
> I really don't understand how this is causing a segfault on freebsd while making aarch64-linux print "zig0: not found"

- examine zig0 binary with this PR on local freebsd 13.0: debug+static build of zig:
```sh
$ file zig0
zig0: ELF 64-bit LSB executable, x86-64, version 1 (FreeBSD), dynamically linked, for FreeBSD 13.0 (1300139), FreeBSD-style, with debug_info, not stripped
$ ldd zig0
zig0:
zig0: signal 11
```
- notable is that it is a dynamic exe
- here are the interesting last args to the `c++ -static ...` command to produce zig0:
```
-lrt -lexecinfo -lpthread -lm -Wl,-Bstatic -lz -Wl,-Bdynamic -lpthread -lm -Wl,-Bstatic -lz -Wl,-Bdynamic
```
- here is what those same args look like on master:
```
-lrt  -lexecinfo  -lpthread  -lm  -lz  -lpthread  -lm  -lz
```
- and after applying the same args to this PR's build (i.e. getting rid of that `-Wl,-Bstatic` and `-Bdynamic` nonsense) some sane results appear:
```sh
$ file zig0
zig0: ELF 64-bit LSB executable, x86-64, version 1 (FreeBSD), statically linked, for FreeBSD 13.0 (1300139), FreeBSD-style, with debug_info, not stripped
$ ldd zig0
ldd: zig0: not a dynamic ELF executable
```

That first `zig0` executable is not a proper dynamic executable yet it tries to be: `c++ -static` means no ability to use dynamic libraries. Subsequent `-Wl,-Bstatic` are probably harmless, but `-Wl,-Bdynamic` is likely the culprit.

kubkon:
FYI, I've updated macos tarballs to the latest zig-bootstrap, pushed them to our s3 instance and updated the CI scripts. Fingers crossed everything is going to work out fine!

andrewrk:
Thanks @kubkon!

Based on @mikdusan's comment I expect FreeBSD to be fixed based solely on an updated tarball, which I will do today. And then it's just the Drone CI mystery left to solve. The nuclear option would be to stop trying to get alpine linux to build our tarballs, but instead only use it to test, and then use the bootstrap repository to provide the aarch64-linux tarballs. This is probably a good idea; arguably all the Zig tarballs we provide on the download page should be via zig-bootstrap.

kubkon:
> Thanks @kubkon!
> 
> Based on @mikdusan's comment I expect FreeBSD to be fixed based solely on an updated tarball, which I will do today. And then it's just the Drone CI mystery left to solve. The nuclear option would be to stop trying to get alpine linux to build our tarballs, but instead only use it to test, and then use the bootstrap repository to provide the aarch64-linux tarballs. This is probably a good idea; arguably all the Zig tarballs we provide on the download page should be via zig-bootstrap.

I definitely agree with the last bit: all the Zig tarballs we provide on the download page should be via zig-bootstrap.

andrewrk:
FreeBSD tarball is updated; let's see if it passes. I also created the aarch64-musl tarball (`zig+llvm+lld+clang-aarch64-linux-musl-0.8.0-dev.2703+c12704a33.tar.xz`) but need to rework the docker image to use it which will have to wait until my mind is fresh tomorrow.

andrewrk:
whew, I think this may finally be in a passing state.

PR #8937, opened 2021-05-30
.Bool have a unique representation too.

SrMordred:


SrMordred:
padding bits are undefined. ok then ;)

PR #8946, opened 2021-05-31
stage1: Fold comptime comparison of BoundFn values

### LINKED ISSUE #8941: Compiler bug: Unreachable at src/stage1/analyze.cpp:7331 in const_values_equal
Should be enough to reproduce: https://void.partidopirata.com.ar/dump/compiler_bug.tar.gz

in file `src/rss.zig` at line 27

```zig
    switch (uri.host) {
        .ip => |addr| {
            const str = try std.fmt.allocPrint(allocator, "{}", .{addr.format});
            defer str.deinit();
            try list.appendSlice(str);
        },
        .name => |name| try list.appendSlice(name),
    }
```

LemonBoy:
Closes #8941

### DIFF
--- a/src/stage1/analyze.cpp
+++ b/src/stage1/analyze.cpp
@@ -7326,6 +7326,7 @@ bool const_values_equal(CodeGen *g, ZigValue *a, ZigValue *b) {
             }
         }
         case ZigTypeIdBoundFn:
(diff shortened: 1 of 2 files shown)


LemonBoy:
Nope.

PR #8948, opened 2021-05-31
Call WSAStartup in `getAddressList` on Windows.

### LINKED ISSUE #8943: windows: `std.net.getAddressList` requires manual `WSAStartup` call
```zig
const std = @import("std");

pub fn main() anyerror!void {
    defer arena.deinit();
    const allocator = &arena.allocator;

    const list = try std.net.getAddressList(allocator, "ziglang.org", 443);
    std.log.info("{}", .{list.addrs[0]});
}
```

This program crashes with the error `error.Unexpected: GetLastError(10093): Either the application has not called WSAStartup, or WSAStartup failed.`. To get this to work you need to add a call to `WSAStartup` as follows:

```zig
const std = @import("std");

pub fn main() anyerror!void {
    defer arena.deinit();
    const allocator = &arena.allocator;

    _ = try std.os.windows.WSAStartup(2, 2);

    const list = try std.net.getAddressList(allocator, "google.com", 443);
    std.log.info("{}", .{list.addrs[0]});

    try std.os.windows.WSACleanup();
}
```

The `os.socket` API abstracts this call away from user-code through the `WSASocketW` in `os/windows.zig`, which initializes WSA if needed. It seems right to me that the `std.net` APIs should do the same.

zacwood9:
Fixes #8943. Uses the same steps as the Windows socket functions: if the call to `getaddrinfo` fails with a WSA error, call WSAStartup, and run `getaddrinfo` again.

I also moved the mutex used in WSASocketW to wrap the WSAStartup call into WSAStartup itself to prevent any possible data races if, for example, WSASocketW and getAddressList were called concurrently.

### DIFF
--- a/lib/std/net.zig
+++ b/lib/std/net.zig
@@ -724,7 +724,6 @@ pub fn getAddressList(allocator: *mem.Allocator, name: []const u8, port: u16) !*
             .next = null,
         };
 
-
         // On Windows, we need to run WSAStartup before we can call getaddrinfo.
         // See notes in os/windows.zig on WSASocketW for more on
         // how the standard library approaches Windows networking.
@@ -735,7 +734,7 @@ pub fn getAddressList(allocator: *mem.Allocator, name: []const u8, port: u16) !*
                 error.BlockingOperationInProgress => return error.Unexpected,
                 error.ProcessFdQuotaExceeded => return error.ProcessFdQuotaExceeded,
                 error.Unexpected => return error.Unexpected,
-                else => return error.Unexpected
+                else => return error.Unexpected,
             };
         }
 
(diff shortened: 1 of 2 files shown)


zacwood9:
@LemonBoy thanks for the link to the discussion, that was very helpful! I wrapped WSAStartup in `std.once`, it's a little ugly since std.once only supports void functions with no arguments so I had to use some global variables. If there's a more idiomatic way to do this please let me know, I didn't find any other usages of `std.once` in the standard library that weren't void functions.

zacwood9:
Closing this - thought it would just be a quick fix but looks like it's not, and I don't think I'll be getting around to this anytime soon

marler8997:
Too bad, I think this work is necessary if we want to support Windows networking without requiring users to call WSAStartup themselves.

LemonBoy:
You got sucked into over-engineering what was supposed to be a simple solution.
I've opened #9028 to let `std.Once` receive arguments and return errors but you can safely:
- Forget about the version number, let's use 2.2 and we'll eventually revisit this decision if needed
- `@panic` if an error is returned, if no winsock DLL can be loaded we can't do much

PR #8961, opened 2021-06-01
test: re-enable "float widening f16 to f128"

### LINKED ISSUE #3282: When targeting AArch64, unable to widen cast f16 to f128
```zig
export fn entry() bool {
    var x: f16 = 12.34;
    var y: f128 = x;
}
```

```
LLVM ERROR: Unsupported library call operation!
```

Note that it works if you do:

```zig
export fn entry() void {
    var x: f16 = 12.34;
    var y: f32 = x;
    var z: f128 = y;
}
```

This issue is to:
 * Work with upstream LLVM to get this working
 * In the mean time, work around the issue by generating f16 -> f128 casts by taking a brief detour through f32.

mikdusan:
- test no longer trips llvm assert for aarch64 and powerpc64le
- unclear which version of llvm fixed issue but 12.0.0 works

closes #3282

### DIFF
--- a/test/behavior/widening.zig
+++ b/test/behavior/widening.zig
@@ -29,10 +29,6 @@ test "float widening" {
 }
 
 test "float widening f16 to f128" {
-    // TODO https://github.com/ziglang/zig/issues/3282
-    if (@import("builtin").target.cpu.arch == .aarch64) return error.SkipZigTest;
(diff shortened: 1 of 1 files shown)


mikdusan:
PR cancelled. Another related test in the source code is failing.
see https://github.com/ziglang/zig/issues/3282#issuecomment-852570470

PR #8962, opened 2021-06-02
translate-c: Fix performance hazard in transPreprocessorEntities

### LINKED ISSUE #8959: translate-c: O(N^2) performance hazard in `transPreprocessorEntities`
`transPreprocessorEntities` calls `mem.len` on the source buffer once for each macro definition.

The following python program can be used to generate a C file with N preprocessor definitions
```python
import sys
for i in range(int(sys.argv[1])):
    print('#define FOO{i} {i}'.format(i=i))
```
run with `python cdefs.py 10000 > test.h`

running translate-c on the result takes about 5s for 10k records on my machine and 20s for 20k records.

ehaas:
Fixes O(N^2) behavior of `transPreprocessorEntities` due to repeated calls to
`mem.len`

Closes #8959

### DIFF
--- a/src/clang.zig
+++ b/src/clang.zig
@@ -607,6 +607,13 @@ pub const IntegerLiteral = opaque {
     extern fn ZigClangIntegerLiteral_isZero(*const IntegerLiteral, *bool, *const ASTContext) bool;
 };
 
+/// This is just used as a namespace for a static method on clang's Lexer class; we don't directly
+/// deal with Lexer objects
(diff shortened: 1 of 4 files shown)


ehaas:
On my machine, translating a file with 20,000 simple preprocessor definitions (`#define FOO1 1`...) takes 0.64 seconds with this patch; 19.5 without it.

Vexu (approved):

PR #8968, opened 2021-06-02
Ensure output from child processes is fully read

leroycep:
- This commit does that by looping over the `read` until it returns 0.


mikdusan:
Hi @leroycep, thank you for this PR. However, #8988 has been merged and fixes the issue. We went with the approach of avoiding an inner-loop for each file descriptor.

And thanks for reporting #8969!

PR #8975, opened 2021-06-03
Breaking hash map changes for 0.8.0

### LINKED ISSUE #7489: std.HashMap.get() inconsistent documentation and code.
The following code can be found in `lib/std/hash_map.zig:562`
```zig
/// Get an optional pointer to the value associated with key, if present.
pub fn get(self: Self, key: K) ?V {
```
The docstring states that this function returns an optional pointer to the value, yet the signature returns an optional copy.
I'd like to use this opportunity to open a discussion about what should the behavior be.
Personally I'd prefer getting a pointer, which would allow me to modify the held value without shelling out for `getEntry`, which also exposes the key.

### LINKED ISSUE #8363: ArrayHashMap: store values separately from key/hash

### LINKED ISSUE #8528: std lib hash maps: add a context parameter for hash and eql functions
Problem: you're implementing hash map eql or hash function, but in order to accomplish this, you need more data than just the key. Here's a real world use case:

```zig
const Context = struct {
    string_table: std.HashMap(u32, void, hash, eql),
    string_bytes: std.ArrayListUnmanaged(u8),
};
```

```zig
fn eql(a: u32, b: u32) bool {
    // error: string_bytes not defined
    return mem.eql(u8, mem.spanZ(string_bytes.items.ptr + a), mem.spanZ(string_bytes.items.ptr + b));
}
```

We need to be able to pass Context to the hash map like this:

```zig
const StringTable = std.HashMap(u32, void, hash, eql, Context);
var t = StringTable.init(allocator, context);
```

So that it can be passed to hash and eql like this:

```zig
fn eql(context: Context, a: u32, b: u32) bool {
    return mem.eql(u8, mem.spanZ(context.string_bytes.items.ptr + a), mem.spanZ(context.string_bytes.items.ptr + b));
}
```

For HashMapUnmanaged, the context should be passed in with every function call that needs it, just like allocator.

### LINKED ISSUE #8619: std lib hash maps: ability to query with custom hash and eql function
Depends on #8528.

Use case: I want a string hash map, but the memory for the strings is stored in a string table, null-terminated. That means I can save memory by using `u32` indexes (4 bytes each) into the string table rather than `[]const u8` (16 bytes each on 64-bit arch).

Example: `HashMap(u32, *Decl)`

But when using `get()` to find out if an item exists in the hash map, I need to use a name that does not exist in the string table. Perhaps it is a string literal "foo" or perhaps the string is stored elsewhere. But I should still be able to find a matching entry.

I propose additional `get` methods that allow providing alternate `hash` and `eql` functions. For convenience they can take a type that acts kind of like a plugin:

```zig
fn string_table_hash(string_table: [*:0]const u8, k: u32) u32 {
    return autoStringHash(mem.spanZ(string_table + k));
}
fn string_table_eql(string_table: [*:0]const u8, a: u32, b: u32) u32 {
    const a_string = mem.spanZ(string_table + a);
    const b_string = mem.spanZ(string_table + b);
    return mem.eql(u8, a_string, b_string);
}
var hash_map: std.HashMapUnmanaged(u32, *Decl, string_table_hash, string_table_eql, string_table) = .{};

// ...put stuff in it...

// define our key adapter
const Adapter = struct {
    const Context = [*:0]const u8;
    const Key = []const u8;
    fn hash(_: Context, k: Key) u32 {
        return autoStringHash(k);
    }
    // the eql function has `a` in the hash map key type, but `b` is the adapted key type
    fn eql(string_table: [*:0]const u8, a: u32, b: Key) bool {
        const a_string = mem.spanZ(string_table + a);
        return mem.eql(u8, a_string, b);
    }
};

// now we can get with `[]const u8` keys too!
const has_foo = hash_map.getAdapted(Adapter, string_table, "foo");
```

### LINKED ISSUE #8971: std hash map API unification
Hello,

while going through the documentation & the std lib, I found that `std.HashMap` and `std.BufMap` function to set a value are respectively called `put` and `set`. While I understand the need to have multiple different hash maps, maybe it would be interesting to unify their API to make it easier for end users, especially since both hash maps have the same `get` method.

Let me know if you need any additional information.

SpexGuy:
- hash/eql functions moved into a Context object
- *Context functions pass an explicit context (closes #8528)
- *Adapted functions pass specialized keys and contexts (closes #8619)
- new getPtr() function returns a pointer to value (closes #7489)
- remove functions renamed to fetchRemove
- new remove functions return bool
- removeAssertDiscard deleted, use assert(remove(...)) instead.
- Keys and values are stored in separate arrays (closes #8363)
- Entry is now {*K, *V}, the new KV is {K, V}
- BufSet/BufMap functions renamed to match other set/map types (closes #8971)
- fixed iterating-while-modifying bug in src/link/C.zig

### DIFF
--- a/lib/std/array_hash_map.zig
+++ b/lib/std/array_hash_map.zig
@@ -17,22 +17,35 @@ const Allocator = mem.Allocator;
 const builtin = std.builtin;
 const hash_map = @This();
 
+/// An ArrayHashMap with default hash and equal functions.
+/// See AutoContext for a description of the hash and equal implementations.
 pub fn AutoArrayHashMap(comptime K: type, comptime V: type) type {
-    return ArrayHashMap(K, V, getAutoHashFn(K), getAutoEqlFn(K), !autoEqlIsCheap(K));
+    return ArrayHashMap(K, V, AutoContext(K), !autoEqlIsCheap(K));
 }
 
+/// An ArrayHashMapUnmanaged with default hash and equal functions.
+/// See AutoContext for a description of the hash and equal implementations.
 pub fn AutoArrayHashMapUnmanaged(comptime K: type, comptime V: type) type {
-    return ArrayHashMapUnmanaged(K, V, getAutoHashFn(K), getAutoEqlFn(K), !autoEqlIsCheap(K));
+    return ArrayHashMapUnmanaged(K, V, AutoContext(K), !autoEqlIsCheap(K));
 }
 
 /// Builtin hashmap for strings as keys.
 pub fn StringArrayHashMap(comptime V: type) type {
-    return ArrayHashMap([]const u8, V, hashString, eqlString, true);
+    return ArrayHashMap([]const u8, V, StringContext, true);
 }
 
 pub fn StringArrayHashMapUnmanaged(comptime V: type) type {
-    return ArrayHashMapUnmanaged([]const u8, V, hashString, eqlString, true);
+    return ArrayHashMapUnmanaged([]const u8, V, StringContext, true);
 }
+
+pub const StringContext = struct {
+    pub fn hash(self: @This(), s: []const u8) u32 {
+        return hashString(s);
+    }
+    pub fn eql(self: @This(), a: []const u8, b: []const u8) bool {
+        return eqlString(a, b);
+    }
+};
 
 pub fn eqlString(a: []const u8, b: []const u8) bool {
     return mem.eql(u8, a, b);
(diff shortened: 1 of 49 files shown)


lib/std/array_hash_map.zig
```diff
@@ -320,40 +358,54 @@ pub fn ArrayHashMap(
...
             value: V,
         };
 
+        pub const DataList = std.MultiArrayList(Data);
+
         pub const Hash = if (store_hash) u32 else void;
 
         pub const GetOrPutResult = struct {
-            entry: *Entry,
+            key_ptr: *K,
```
daurnimator, line 400: Should this be `*const K`? You don't want people modifying the key...

SpexGuy: Modifying the key is a valid use case, and sometimes (in the case of getOrPut*Adapted) required.  But even aside from that, you might want to do a `put` with a temporary key, and then replace the key with a newly allocated version if the map took ownership of it.  `BufMap.put` does this, for example.

jedisct1:
These changes are great!

And having keys and values stored separately is very useful.

If this is not too late, it would be a nice set of changes to land before 0.8.0 is tagged.

jedisct1 (approved):

lib/std/hash_map.zig
```diff
@@ -78,6 +98,222 @@ pub const DefaultMaxLoadPercentage = default_max_load_percentage;
...
+/// type, PseudoKey = Key = K.
+pub fn verifyContext(comptime RawContext: type, comptime PseudoKey: type, comptime Key: type, comptime Hash: type) void {
+    comptime {
+        var allow_const_ptr = false;
+        var allow_mutable_ptr = false;
+        // Context is the actual namespace type.  RawContext may be a pointer to Context.
+        var Context = RawContext;
+        // Make sure the context is a namespace type which may have member functions
+        switch (@typeInfo(Context)) {
+            .Struct, .Union => {},
```
mattbork, line 117: Since an enum can have methods, it could also be used as a context.

lib/std/hash_map.zig
```diff
@@ -174,67 +494,114 @@ pub fn HashMap(
...
+        pub fn fetchRemoveAdapted(self: *Self, key: anytype, ctx: anytype) ?KV {
+            return self.unmanaged.fetchRemoveAdapted(key, ctx);
+        }
+
+        /// Finds the value associated with a key in the map
         pub fn get(self: Self, key: K) ?V {
-            return self.unmanaged.get(key);
+            return self.unmanaged.getContext(key, self.ctx);
+        }
+        pub fn getAdapted(self: Self, key: anytype, ctx: anytype) ?*V {
```
mattbork, line 544: Should the return type be `?V` here?

ArborealAnole:
Wait, are keys and values now stored separately for `HashMap`? I thought that would only be for `ArrayHashMap`. Now for `HashMap` if you've aligned your KV within cache block size, you'll need 2 block fetches instead of 1 when your K matches in order to also fetch the V. https://github.com/ziglang/zig/issues/8363#issuecomment-809637426

PR #8977, opened 2021-06-03
std.crypto: timing-safe functions to add&subtract serialized integers

jedisct1:
This is useful to increment nonces and for scalar reduction.

It will avoid code duplication in crypto/25519 and crypto/pcurves/*

### DIFF
--- a/lib/std/crypto/utils.zig
+++ b/lib/std/crypto/utils.zig
@@ -1,6 +1,7 @@
 const std = @import("../std.zig");
 const debug = std.debug;
 const mem = std.mem;
+const random = std.crypto.random;
 const testing = std.testing;
 
 const Endian = std.builtin.Endian;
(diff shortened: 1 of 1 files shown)


lib/std/crypto/utils.zig
```diff
@@ -77,6 +78,66 @@ pub fn timingSafeCompare(comptime T: type, a: []const T, b: []const T, endian: E
     return Order.lt;
 }
 
+/// Add two integers serialized as arrays of the same size, in constant time.
+/// The result is stored into `result`, and `true` is returned if the an overflow occurred.
```
daurnimator, line 82: s/the an/any/

daurnimator (changes requested):
Are there existing pieces of code that needs updating?

lib/std/crypto/utils.zig
jedisct1: Fixed, thanks!

jedisct1:
The main reason I need these is a `reduce32()` function required for ECDSA. Other scalar reduction functions can then be updated to use `reduce32()` and additions.

lib/std/crypto/utils.zig
```diff
@@ -77,6 +78,66 @@ pub fn timingSafeCompare(comptime T: type, a: []const T, b: []const T, endian: E
...
+/// Add two integers serialized as arrays of the same size, in constant time.
+/// The result is stored into `result`, and `true` is returned if an overflow occurred.
+pub fn timingSafeAdd(comptime T: type, a: []const T, b: []const T, result: []T, endian: Endian) bool {
+    const len = a.len;
+    debug.assert(len == b.len and len == result.len);
+    const bits = switch (@typeInfo(T)) {
+        .Int => |cinfo| if (cinfo.signedness != .unsigned) @compileError("Elements to be added must be unsigned") else cinfo.bits,
+        else => @compileError("Elements to be added must be integers"),
+    };
+    const Cext = std.meta.Int(.unsigned, bits + 1);
```
LemonBoy, line 90: The `+1` trick is incredibly wasteful as it forces the sum to work on 33/65 bits. Use `@addWithOverflow` as it's done in the bigint implementation instead.

jedisct1: Is @addWithOverflow guaranteed to be constant-time?

The intent of using a bigger integer is to minimize the chances of having conditional branches.

LemonBoy: > Is @addWithOverflow guaranteed to be constant-time?

No, but neither is the"bigint" arithmetic that LLVM legalizes this operation to. If the target CPU has a carry flag and a way of transferring it to a GPR, a pretty safe assumption to make, `@addWithOverflow` is cheap and constant-time.

jedisct1: The code has been changed to use `@addWithOverflow` and `@subWithOverflow`.

Thanks!

Issue #6757, opened 2020-10-21
WASI reactors


DagAgren:
I can't seem to find any information about this, so sorry if I am just missing something, but, does zig support building WASI "reactors"?

The concept of a reactor in WASI is a module that does not have a regular start function that runs and exits, but instead has a simpler _initialize() entry point, that just does whatever low-level initialisation the binary and WASI implementation needs (calling global constructors, setting up the preloaded directories), and then exits, leaving everything in place. The hosting environment then calls whatever function the module exports to actually do work.

Is this currently possible in zig, and if not, would it be possible to add support for it?

zigazeljko:
You can manually export the `_initialize` function and use `build-lib` (instead of `build-exe`):

```zig
const std = @import("std");

export fn _initialize() void {
    std.debug.print("hello, world!\n", .{});
}
```

Other than that, there is no special support for this.

DagAgren:
The issue, I guess, is that wasi-libc relies on initialisation that happens in either _start() or _initialize(), and things like fopen() will refuse to work without it. But I am not sure if there is an exposed way to trigger this initialisation outside of the default crt1.o and crt1-reactor.o.

DagAgren:
Also, it seems this happens through __attribute((constructor))__ in wasi-libc. How is that handled in zig? Is it supported?

kubkon:
https://github.com/ziglang/zig/pull/8837 unified the behavior of build commands for Wasm, where `build-lib` will by default produce a static archive that can be used for further linking with `wasm-ld`. If you want to generate a module using `build-lib` you should pass additional `-dynamic` flag to the invocation.

As far as a WASI reactor is concerned, I haven't put much thought into that yet, however, given that we now ship WASI libc, it would be good to first work out how to expose this functionality in C (i.e., with `zig cc`) since #8837 left it a TODO. For pure Zig, we have to be careful since Zig doesn't invoke any ctors. In fact, it is currently assumed the user's responsibility to call any relevant startup code that would otherwise involve dynamic memory allocation such as fetching the preopens, hence the actual startup amounts to calling `_main` in `start.zig`: https://github.com/ziglang/zig/blob/a9dd8d75431357a59238da3fd949aa1f11eead84/lib/std/start.zig#L166.

mathetake:
[...]

PR #8978, opened 2021-06-03
cc,wasi: support WASI reactors via -mexec-model flag

mathetake:
Based on this patch to LLVM (https://reviews.llvm.org/D62922). This partially resolves (only for zig cc) #6757


src/wasi_libc.zig
```diff
@@ -51,16 +51,20 @@ pub fn buildWasiLibcSysroot(comp: *Compilation) !void {
...
                 .src_path = try comp.zig_lib_directory.join(arena, &[_][]const u8{
                     "libc", try sanitize(arena, file_path),
                 }),
                 .extra_flags = args.items,
             });
+            try comp.build_crt_file(
+                file_path[std.mem.lastIndexOf(u8, file_path, "/").? + 1 .. std.mem.lastIndexOf(u8, file_path, ".").?],
+                .Obj,
+                comp_sources.items,
+            );
```
kubkon, line 66: These changes are no longer necessary and are fixed in https://github.com/ziglang/zig/pull/8992

kubkon (changes requested):
Thanks for the PR! I've got a couple of suggestions that when we address, I'm sure we'll be able to merge this change no problem.

src/Compilation.zig
```diff
@@ -719,6 +719,7 @@ pub const InitOptions = struct {
     test_filter: ?[]const u8 = null,
     test_name_prefix: ?[]const u8 = null,
     subsystem: ?std.Target.SubSystem = null,
+    want_reactor_exec_model: bool = false,
```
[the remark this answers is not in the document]
mathetake: Thanks, that would be much better! Done in https://github.com/ziglang/zig/pull/8978/commits/6614a204d060411b4bbbdbcd035ccee919a5c006

src/wasi_libc.zig
mathetake: sounds good! so should I wait until that PR is merged?

mathetake:
OK so I verified this works as expected locally.

src/wasi_libc.zig
kubkon: Actually, feel free to rebase onto my branch, and we can carry on from there. Feel free to ping me on Discord if you have any questions about the changes in that PR or whatnot!

mathetake:
@kubkon ok now rebased and added a necessary commit for checking in command/reactor crt (https://github.com/ziglang/zig/pull/8978/commits/a2b20449efdb07ba28b739fa544d6364d6a5851c). PTAL!

src/Compilation.zig
```diff
@@ -1444,6 +1451,8 @@ pub fn create(gpa: *Allocator, options: InitOptions) !*Compilation {
             // TODO add logic deciding which crt1 we want here.
             comp.work_queue.writeAssumeCapacity(&[_]Job{
                 .{ .wasi_libc_crt_file = .crt1_o },
+                .{ .wasi_libc_crt_file = .crt1_command_o },
+                .{ .wasi_libc_crt_file = .crt1_reactor_o },
```
kubkon, line 1455: OK, so my idea for this bit here is similar to the logic just a few lines above we use for working out which emulated WASI libc subcomponents we should build in this run (unless they are already in global cache). So, instead of building _all_ CRTs, how about we now use the new option you just added to work out precisely which CRT we should include in our build, something like:

```zig
const crt_file: wasi_libc.CRTFile = if (comp.bin_file.options.wasi_exec_model) |exec_model| crt_file: {
    switch (exec_model) {
        .command => break :crt_file .crt1_command_o,
        .reactor => break :crt_file .crt1_reactor_o,
    }
} else .crt1_o;
comp.work_queue.writeAssumeCapacity(&[_]Job{
    .{ .wasi_libc_crt_file = crt_file },
    .{ .wasi_libc_crt_file = .libc_a },
});
```

kubkon: Also, while we're here, could you remove the TODO as it is no longer needed.

src/link.zig
```diff
@@ -117,6 +117,8 @@ pub const Options = struct {
     version: ?std.builtin.Version,
     libc_installation: ?*const LibCInstallation,
 
+    wasi_exec_model: ?Compilation.WasiExecModel = null,
```
kubkon, line 120: Could you add a sentence or two explaining what this option holds?

src/Compilation.zig
```diff
@@ -725,6 +730,7 @@ pub const InitOptions = struct {
     test_filter: ?[]const u8 = null,
     test_name_prefix: ?[]const u8 = null,
     subsystem: ?std.Target.SubSystem = null,
+    wasi_exec_model: ?WasiExecModel = null,
```
kubkon, line 733: Could you add a sentence or two explaining what this option holds?

src/main.zig
```diff
@@ -1254,6 +1255,14 @@ fn buildOutputType(
                     .framework => try frameworks.append(it.only_arg),
                     .nostdlibinc => want_native_include_dirs = false,
                     .strip => strip = true,
+                    .exec_model => {
+                        wasi_exec_model = if (std.mem.eql(u8, it.only_arg, "reactor"))
+                            Compilation.WasiExecModel.reactor
+                        else if (std.mem.eql(u8, it.only_arg, "command"))
+                            Compilation.WasiExecModel.command
+                        else
+                            null;
```
kubkon, line 1264: Minor nit: you actually don't need to initialize to `null` here since it is already initialized to this value by default.

src/link/Wasm.zig
```diff
@@ -680,6 +680,11 @@ fn linkWithLLD(self: *Wasm, comp: *Compilation) !void {
             // Put stack before globals so that stack overflow results in segfault immediately
             // before corrupting globals. See https://github.com/ziglang/zig/issues/4496
             try argv.append("--stack-first");
+
+            // Reactor execution model does not have _start so lld doesn't look for it.
+            if (self.base.options.wasi_exec_model != null and self.base.options.wasi_exec_model.? == Compilation.WasiExecModel.reactor) {
+                try argv.append("--no-entry");
```
kubkon, line 686: I think it is more readable to unpack the optional using the `if (opt) |value|` syntax which is more readable, and use labeled block to leave early if the inner value is not a `.reactor`:

```zig
if (self.base.options.wasi_exec_model) |exec_model| blk: {
    if (exec_model != .reactor) break :blk;
    try argv.append("--no-entry");
}
```

See [Optionals](https://ziglang.org/documentation/0.8.0/#toc-Optionals) for more details.

kubkon (changes requested):

src/Compilation.zig
mathetake: https://github.com/ziglang/zig/pull/8978/commits/cd610975f2c401ec413787c04c2c1c8a34e54804 fixed!

src/link.zig
mathetake: https://github.com/ziglang/zig/pull/8978/commits/cd610975f2c401ec413787c04c2c1c8a34e54804 fixed!

src/Compilation.zig
mathetake: fixed in https://github.com/ziglang/zig/pull/8978/commits/cd610975f2c401ec413787c04c2c1c8a34e54804 with a slight modification. I couldn't compile without explicitly giving the enum value inside of these switch branches (maybe due to type inference limitation)?

mathetake: Anyway thank you so much for suggestions!

src/main.zig
mathetake: fixed in  https://github.com/ziglang/zig/pull/8978/commits/230878352006e466f66600a979bdba8bddc98acd

kubkon (approved):
Looks great! Thanks for working on this @mathetake!

kubkon:
BTW once #8992 lands, I'll merge this one next.

mathetake:
@kubkon Thanks for the review! Much appreciated, and I learned a lot from your suggestions:)

kubkon:
OK Github, why the heck did you close this PR?!

kubkon:
@mathetake would you mind reopening the PR?

kubkon:
Or I can try merging your branch manually on the command line, let's see...

mathetake:
OK, let me reopen!

kubkon:
> OK, let me reopen!

Don't worry about conflicts, I'll take care of them while merging.

mathetake:
https://github.com/ziglang/zig/pull/9052 here you go!

PR #8979, opened 2021-06-03
Json Stringify option to not write out null optional fields

hazeycode:
Fixed version of https://github.com/ziglang/zig/pull/8908 with added test to parse stringify output

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1303,14 +1303,14 @@ pub const Value = union(enum) {
                         try child_whitespace.outputIndent(out_stream);
                     }
 
-                    try stringify(entry.key_ptr.*, options, out_stream);
+                    try stringify(entry.key, options, out_stream);
                     try out_stream.writeByte(':');
                     if (child_options.whitespace) |child_whitespace| {
                         if (child_whitespace.separator) {
                             try out_stream.writeByte(' ');
                         }
                     }
-                    try stringify(entry.value_ptr.*, child_options, out_stream);
+                    try stringify(entry.value, child_options, out_stream);
                 }
                 if (field_output) {
                     if (options.whitespace) |whitespace| {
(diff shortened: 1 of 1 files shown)


hazeycode:
Found another failure case. Needs more testing.

daurnimator:
I wonder if a better option might be a `skip_defaults` to avoid rendering all defaults.
Then you could opt-in per optional field by setting their default value to `null`.

lib/std/json.zig
```diff
@@ -2731,6 +2731,9 @@ pub const StringifyOptions = struct {
 
             /// Should unicode characters be escaped in strings?
             escape_unicode: bool = false,
+
+            /// Should optional fields with null value be written?
+            write_null_optional_fields: bool = true,
```
daurnimator, line 2736: Why is this under `StringOutputOptions`?

daurnimator (changes requested):

lib/std/json.zig
hazeycode: Good question. This was an error on my part

hazeycode:
> I wonder if a better option might be a `skip_defaults` to avoid rendering all defaults.
> Then you could opt-in per optional field by setting their default value to `null`.

I'm unconvinced. It does make sense but is there a use-case where you don't want to stringify nulls that go OTW but would want to stringify some local default that doesn't make sense remotely? i.e.

```
const Message = struct {
    user_id: u32 = 1337, // should stringify
    optional: []const u8 = null, // should not stringify 
    required: []const u8, // should stringify
}
```

Of course, the problem here is easily avoided, however, not convinced that it's a good user experience.

hazeycode:
This appears to be working properly now!
Commits likely want squashing but I had all kinds of bother trying to rebase

lib/std/json.zig
```diff
@@ -2822,23 +2825,36 @@ pub fn stringify(
...
 
-                if (!field_output) {
-                    field_output = true;
-                } else {
-                    try out_stream.writeByte(',');
-                }
-                if (child_options.whitespace) |child_whitespace| {
-                    try out_stream.writeByte('\n');
-                    try child_whitespace.outputIndent(out_stream);
+                var emit_field = true;
```
andrewrk, line 2828: Looks redundant with `field_output`

andrewrk (approved):

PR #8987, opened 2021-06-04

travisstaloch:
- if a struct T passed to `parse(T, ...)` contains a `pub const __tags` of the form
  shown below, these aliases will be used as the json keys instead of the
  field name.
- adds `std.mem.{indexOfSliceScalar, indexOfSliceScalarPos}` which support
  finding aliases. adding these was my workaround to a compiler bug
  encountered when using a for loop for the same purpose.

```zig
    const S = struct {
        value: u8,
        pub const __tags = .{
            .value = .{ .alias = "__value" },
        };
    };
```

The names `__tags` and `alias` were chosen by me ad-hoc and may be easily
changed.
Inspired by #1099 https://github.com/ziglang/zig/issues/1099#issuecomment-846404132

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1541,16 +1541,19 @@ test "skipValue" {
     }
 }
 
+const FieldAliasMap = std.StringArrayHashMapUnmanaged([]const u8);
+
 // build a map of aliases from field_name => alias_name
-fn buildFieldAliasMap(comptime T: type, options: ParseOptions, allocator: *mem.Allocator) !std.StringHashMapUnmanaged([]const u8) {
-    var result = std.StringHashMapUnmanaged([]const u8){};
-    if (options.field_aliases.len == 0) return result;
-    inline for (comptime std.meta.fieldNames(T)) |field_name| {
-        for (options.field_aliases) |entry| {
-            if (std.mem.eql(u8, field_name, entry.field)) {
-                try result.put(allocator, entry.field, entry.alias);
-            }
-        }
+fn buildFieldAliasMap(comptime field_names: []const []const u8, options: ParseOptions, list: []FieldAliasMap.Data) FieldAliasMap {
+    var result = FieldAliasMap{ .entries = FieldAliasMap.DataList{
+        .bytes = @ptrCast([*]u8, list.ptr),
+        .len = 0,
+        .capacity = field_names.len,
+    } };
+    inline for (field_names) |field_name| {
+        for (options.field_aliases) |field_alias|
+            if (std.mem.eql(u8, field_name, field_alias.field))
+                result.putAssumeCapacity(field_alias.field, field_alias.alias);
     }
     return result;
 }
@@ -1567,26 +1570,33 @@ test "field aliases" {
     try testing.expectEqual(@as(u8, 42), s.value);
 }
 
-test "many field_aliases" {
-    const S = struct { value: u8, zzz: u8 };
+test "many field aliases" {
+    const S = struct { value: u8, a: u8, z: u8 };
     const text =
-        \\{"__value": 42, "__zzz": 84}
+        \\{"__value": 42, "__a": 43, "__z": 44}
     ;
 
-    // make oversized list of aliases to check if stack allocated field_alias_map has enough memory
-    comptime var field_aliases = [1]FieldAlias{undefined} ** 256;
-    comptime for (field_aliases) |*fa, i| {
-        const c = [1]u8{@intCast(u8, 'A' + (i % 64))};
-        const cs = c ** (i / 64 + 1);
-        fa.field = &cs;
-        fa.alias = &("__" ++ cs);
-    };
+    // too many aliases
+    try testing.expectError(
+        error.TooManyFieldAliases,
+        parse(S, &TokenStream.init(text), .{ .field_aliases = &.{
+            .{ .field = "value", .alias = "__value" },
+            .{ .field = "a", .alias = "__a" },
+            .{ .field = "z", .alias = "__z" },
+            .{ .field = "oops", .alias = "__oops" },
+        } }),
+    );
 
     const s = try parse(S, &TokenStream.init(text), .{
-        .field_aliases = &(field_aliases ++ [1]FieldAlias{.{ .field = "value", .alias = "__value" }}),
+        .field_aliases = &.{
+            .{ .field = "value", .alias = "__value" },
+            .{ .field = "a", .alias = "__a" },
+            .{ .field = "z", .alias = "__z" },
+        },
     });
     try testing.expectEqual(@as(u8, 42), s.value);
-    try testing.expectEqual(@as(u8, 84), s.zzz);
+    try testing.expectEqual(@as(u8, 43), s.a);
+    try testing.expectEqual(@as(u8, 44), s.z);
 }
 
 test "field alias escapes" {
@@ -1710,9 +1720,12 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options: 
                     }
                 }
             }
-            var buf: [structInfo.fields.len * (1 << 9)]u8 = undefined;
-            var fba = std.heap.FixedBufferAllocator.init(&buf);
-            var field_alias_map = try buildFieldAliasMap(T, options, &fba.allocator);
+
+            // construct a FieldAliasMap on the stack
+            const field_names = comptime std.meta.fieldNames(T);
+            var list_data: [field_names.len]FieldAliasMap.Data = undefined;
+            if (options.field_aliases.len > field_names.len) return error.TooManyFieldAliases;
+            var field_alias_map = buildFieldAliasMap(field_names, options, &list_data);
 
             while (true) {
                 switch ((try tokens.next()) orelse return error.UnexpectedEndOfJson) {
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1727,6 +1727,12 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options: 
             if (options.field_aliases.len > field_names.len) return error.TooManyFieldAliases;
             var field_alias_map = buildFieldAliasMap(field_names, options, &list_data);
 
+            // re-index the alias map.  this is a no-op unless there are more than
+            // ArrayHashMap.linear_scan_max (8) entries.
+            var buf: [if (field_names.len > 8) field_names.len * 9 else 0]u8 = undefined;
+            var fba = std.heap.FixedBufferAllocator.init(&buf);
+            try field_alias_map.reIndex(&fba.allocator);
+
             while (true) {
                 switch ((try tokens.next()) orelse return error.UnexpectedEndOfJson) {
                     .ObjectEnd => break,
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1531,7 +1531,7 @@ test "skipValue" {
     }
 }
 
-// build a map of aliases from alias_name => target_field_name
+// build a map of aliases from field_name => alias_name
 fn buildFieldAliases(comptime T: type, options: ParseOptions) !std.StringHashMapUnmanaged([]const u8) {
     const field_aliases = options.field_aliases;
     var result = std.StringHashMapUnmanaged([]const u8){};
@@ -1540,7 +1540,7 @@ fn buildFieldAliases(comptime T: type, options: ParseOptions) !std.StringHashMap
         for (alias_entries) |entry| {
             if (std.mem.eql(u8, field_name, entry.field)) {
                 const allocator = options.allocator orelse return error.AllocatorRequired;
-                try result.put(allocator, entry.alias, entry.field);
+                try result.put(allocator, entry.field, entry.alias);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


travisstaloch:
@SpexGuy suspects that this is failing due to significant comptime usage and may have to wait for stage 2.

travisstaloch:
I messed up my rebase in previous pr.  Forgive me as I've closed it and reopened this after mikdusan suggested that the real issue was needing a rebase.  Maybe this will pass now...

andrewrk:
Sorry about that. With self hosted you should be able to comptime to your heart's content. I am confident we can ship it by zig 0.9.0.

lib/std/json.zig
```diff
@@ -1624,11 +1691,16 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options:
...
+            comptime const field_aliases = buildFieldAliases(T);
+
             while (true) {
                 switch ((try tokens.next()) orelse return error.UnexpectedEndOfJson) {
                     .ObjectEnd => break,
                     .String => |stringToken| {
-                        const key_source_slice = stringToken.slice(tokens.slice, tokens.i - 1);
+                        const key_source_slice = findFieldAlias(
+                            field_aliases,
+                            stringToken.slice(tokens.slice, tokens.i - 1),
```
daurnimator, line 1702: This isn't necessarily what you think it is: it might have escapes. Consult `stringToken.escapes` to know.

daurnimator (changes requested):
I'm not sure this is the best way to accomplish this feature: I was thinking it would be better to provide on a per `parse`/`stringify` invocation via an option that contains a map of overrides

lib/std/json.zig
travisstaloch: I'll look into adding escape support soon.  @daurnimator so put this into ParseOptions?  How does this look?: 
```zig
const S = struct {value: u8};
const s = try parse(
    S, 
    &TokenStream.init("{\"__value\": 42}"), 
    .{.aliases = .{.value = "__value"}},
);
```
I'm a little worried this is less re-usable and would promode copy-pasting.  But maybe thats necessary if you wanted to parse different json data into the same struct.  I guess it would be easy enough to store the alias info in the struct decl again like so:
```zig
const S = struct {
    value: u8,
    pub const aliases = .{.value = "__value"};
};
const s = try parse(
    S, 
    &TokenStream.init("{\"__value\": 42}"), 
    .{.aliases = S.aliases}
);
```
Thoughts?

travisstaloch:
@andrewrk looking forward to 0.9 when comptime will go brrr :+1:.  

Should i leave this pr open then after everything else is resolved?  Or close it and plan to re-open / make a new pr after 0.9?

andrewrk:
I'd like to reserve the PR queue for "ready to be reviewed & merged". If something can't be merge-ready due to being blocked by other issues, then it should be an issue, and then resurrected as a PR when the time comes.

travisstaloch:
Sounds good. I opened #8993 as a reminder to revisit after 0.9.  Will close this once other issues with this pr are resolved.

travisstaloch:
@daurnimator i made some changes.  i'm not sure what needs to be done to support escapes.  are you saying that the escaped version of `stringToken.slice(tokens.slice, tokens.i - 1)` should be escaped before checking for a matching alias?  any advice how i could do that?

daurnimator:
> i'm not sure what needs to be done to support escapes. are you saying that the escaped version of `stringToken.slice(tokens.slice, tokens.i - 1)` should be escaped before checking for a matching alias? any advice how i could do that?

I'm saying you need to make sure `{"foo": "bar"}` and `{"f\x6fo":"bar"}` are treated the same.
See the line just below where you edited for how we do that:
```
(stringToken.escapes == .None and mem.eql(u8, field.name, key_source_slice)) or (stringToken.escapes == .Some and (field.name.len == stringToken.decodedLength() and encodesTo(field.name, key_source_slice)))
```
i.e. if `stringToken.escapes == .None`  we can match against the field name directly; otherwise we need to use a different equality check. We have the `encodesTo` for such situations.

Note that to do this you'll want to change your implementation to do most of the work inside the `inline for (structInfo.fields) |field, i| {` loop

----


Also note typo in your commit message: `s/PaseOptions/ParseOptions`

travisstaloch:
thank you :+1: so i added this test  which i expected to fail, but its passing:

```zig
test "field alias escapes" {
    const S = struct { foo: u8 };
    const text =
        "{\"__f\x6fo\": 42}";

    const s = try parse(S, &TokenStream.init(text), .{
        .field_aliases = &.{.{ "foo", "__foo" }},
        .allocator = std.testing.allocator,
    });
    try testing.expectEqual(@as(u8, 42), s.foo);
}
```
not sure what to make of this... :thinking:

daurnimator:
@travisstaloch you have 2 layers of escaping: `\x6f` is already understood by zig, you need to use zig `\\` strings or write `"\\x6f"`

lib/std/json.zig
```diff
@@ -1466,6 +1466,7 @@ pub const ParseOptions = struct {
     ignore_unknown_fields: bool = false,
 
     allow_trailing_data: bool = false,
+    field_aliases: ?[]const [2][]const u8 = null,
```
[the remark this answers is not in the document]
travisstaloch: how about `struct{field: []const u8, alias: []const u8}`?  i think its slightly easier to read and write.

travisstaloch:
> @travisstaloch you have 2 layers of escaping: `\x6f` is already understood by zig, you need to use zig `\\` strings or write `"\\x6f"`

Ok, well I was initially using multiline escapes, but ran into this problem.  Turns out std.json can't even parse this and is giving the same error i was seeing: 

```zig
const std = @import("std");

test "field alias escapes" {
    const S = struct { foo: u8 };
    const text =
        \\{"__f\x6fo": 42}
    ;

    const s = try std.json.parse(S, &std.json.TokenStream.init(text), .{});
    try std.testing.expectEqual(@as(u8, 42), s.foo);
}
```
gives error: 
```sh
$ zig version
0.9.0-dev.21+7462b0e5b

$ zig test /tmp/test.zig 
test "field alias escapes"... FAIL (InvalidEscapeCharacter)
~/Documents/Code/zig/zig/build/lib/zig/std/json.zig:753:21: 0x211714 in std.json.StreamingParser.transition (test)
                    return error.InvalidEscapeCharacter;
                    ^
~/Documents/Code/zig/zig/build/lib/zig/std/json.zig:284:13: 0x20bc86 in std.json.StreamingParser.feed (test)
        if (try p.transition(c, token1)) {
            ^
~/Documents/Code/zig/zig/build/lib/zig/std/json.zig:1122:13: 0x20abf0 in std.json.TokenStream.next (test)
            try self.parser.feed(self.slice[self.i], &t1, &t2);
            ^
~/Documents/Code/zig/zig/build/lib/zig/std/json.zig:1628:26: 0x20b089 in std.json.parseInternal (test)
                switch ((try tokens.next()) orelse return error.UnexpectedEndOfJson) {
                         ^
~/Documents/Code/zig/zig/build/lib/zig/std/json.zig:1795:15: 0x20985e in std.json.parse (test)
    const r = try parseInternal(T, token, tokens, options);
              ^
/tmp/test.zig:9:15: 0x207d31 in test "field alias escapes" (test)
    const s = try std.json.parse(S, &std.json.TokenStream.init(text), .{});

```

Is this an issue with std.json or an invalid input?

travisstaloch:
Ok good.  So this __should__ pass then?  I want to be sure i'm testing the right thing.  
```zig
test "field alias escapes" {
    const S = struct { foo: u8 };
    const text =
        \\{"__f\u006fo": 42}
    ;
    const s = try parse(S, &TokenStream.init(text), .{
        .field_aliases = &.{.{ .field = "foo", .alias = 
        \\__f\u006fo 
        }},
        .allocator = std.testing.allocator,
    });
    try testing.expectEqual(@as(u8, 42), s.foo);
}
```

daurnimator:
no you want to set the alias to just `__foo`. This should pass:


```zig
test "field alias escapes" {
    const S = struct { foo: u8 };
    const text =
        \\{"__f\u006fo": 42}
    ;
    const s = try parse(S, &TokenStream.init(text), .{
        .field_aliases = &.{.{ .field = "foo", .alias =  "__foo" }},
        .allocator = std.testing.allocator,
    });
    try testing.expectEqual(@as(u8, 42), s.foo);
}
```

lib/std/json.zig
```diff
@@ -1466,6 +1466,7 @@ pub const ParseOptions = struct {
     ignore_unknown_fields: bool = false,
 
     allow_trailing_data: bool = false,
+    field_aliases: ?[]const struct { field: []const u8, alias: []const u8 } = null,
```
daurnimator, line 1469: Doesn't need to be null-able: just use the empty list.

daurnimator, line 1469: Add an empty line between fields here

daurnimator (changes requested):
I don't think you need an allocator: you can bound the size of the field aliases by the number of fields in the type, which means it can be on the stack.

travisstaloch:
> I don't think you need an allocator: you can bound the size of the field aliases by the number of fields in the type, which means it can be on the stack.

the allocator is only needed for the `StringHashMapUnmanaged` returned by `buildFieldAliases()`.  how could i initialize this on the stack?  just pass in a fixedBufferAllocator?

travisstaloch:
this makes the tests pass.  not sure if `structInfo.fields.len * 512`  its a good hueristic.  its the smallest power of 2 that passes.  does this seem acceptable?

```zig
            var buf: [structInfo.fields.len * (1 << 9)]u8 = undefined;
            var fba = std.heap.FixedBufferAllocator.init(&buf);
            var field_alias_map = try buildFieldAliasMap(T, options, &fba.allocator);
```

travisstaloch:
continuous-integration/drone/pr and ziglang.zig ci jobs both passed before i got tired waiting to see if 3rd would pass.  going to push my latest commit now.

lib/std/json.zig
travisstaloch: Addressed all of these in latest commit.

lib/std/json.zig
```diff
@@ -1530,6 +1541,82 @@ test "skipValue" {
     }
 }
 
+// build a map of aliases from field_name => alias_name
+fn buildFieldAliasMap(comptime T: type, options: ParseOptions, allocator: *mem.Allocator) !std.StringHashMapUnmanaged([]const u8) {
+    var result = std.StringHashMapUnmanaged([]const u8){};
+    if (options.field_aliases.len == 0) return result;
```
daurnimator, line 1547: no need for this line I think?

lib/std/json.zig
```diff
@@ -1623,6 +1710,9 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options:
                     }
                 }
             }
+            var buf: [structInfo.fields.len * (1 << 9)]u8 = undefined;
```
daurnimator, line 1713: I don't think the limit of 512 is appropriate; I was more thinking something like iterating over the fields (with an `inline for`) and calculating the actual max size of the hash table possible (if there is e.g. 5 fields, then the hash table can't have more than 5 entries).

travisstaloch: ok thats what i was thinking too initially.  but i'm not sure how to do it.  can you provide an example of how to initialize a hash table with only N entries?

travisstaloch: i think i found a solution. will push a commit soon.

travisstaloch:
no idea why the freebsd build failed. i can't tell from the logs.

lib/std/json.zig
```diff
@@ -1624,6 +1721,18 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options:
                 }
             }
 
+            // construct a FieldAliasMap on the stack
+            const field_names = comptime std.meta.fieldNames(T);
+            var list_data: [field_names.len]FieldAliasMap.Data = undefined;
+            if (options.field_aliases.len > field_names.len) return error.TooManyFieldAliases;
+            var field_alias_map = buildFieldAliasMap(field_names, options, &list_data);
```
travisstaloch, line 1728: this should be const. forgot to change it.

Vexu:
Timed out, but it seems like all the relevant tests were run.

travisstaloch:
is there anything left to do on this or is it just waiting in line?  just wanted to ping @daurnimator as there has been no action in a couple weeks.  note that the most recent freebsd ci job seems to have run all relevent tests but just timed out for some reason.  also know that it _shouldn't_ be a problem for me to rebase as happened previously.

andrewrk (changes requested):
Isn't the idea behind this to do the aliasing at compile-time instead of runtime?

> is there anything left to do on this or is it just waiting in line?
>     * if a struct T passed to `parse(T, ...)` contains a `pub const __tags` of the form
>       shown below, these aliases will be used as the json keys instead of the
>       field name.

I don't see this in the code

travisstaloch:
In response to @andrewrk questions:
> Isn't the idea behind this to do the aliasing at compile-time instead of runtime?

I'm not sure if this is currently possible.  In this PR's first commit i had created the field aliases as a list of from_field, to_field pairs at comptime.  However if memory serves, this caused an OOM error.  

The aliases are now stored as an ArrayHashMapUnmanaged which is created with stack memory.  This was done in an attempt to avoid allocations.

> I don't see this in the code

The field_aliases were moved from a `__tags` struct decl to `ParseOptions.field_aliases` as requested by @daurnimator i believe.

travisstaloch:
Closing in favor of #10193

PR #8991, opened 2021-06-04
zig cc: zig c++: add support for -I and --include-directory arguments.

redj:
When using zig cc and zig c++ allow passing -I / --include-directory to clang.

This is re-opening https://github.com/ziglang/zig/pull/6630 which was based on the wrong (llvm11) branch.

Sorry for the long delay.


src/main.zig
```diff
@@ -1250,6 +1250,7 @@ fn buildOutputType(
                     .framework => try frameworks.append(it.only_arg),
                     .nostdlibinc => want_native_include_dirs = false,
                     .strip => strip = true,
+                    .include_dir => try clang_argv.appendSlice(it.other_args),
```
LemonBoy, line 1253: This is identical to `.other`, do you really need `include_dir` ?

redj: Thanks @LemonBoy. I think this patch was so old that it became useless. Sorry for the noise. I'm going to close this.

PR #8992, opened 2021-06-04
cc,wasi: include emulated libs in WASI libc and refactor WASI libc handling

### LINKED ISSUE #8886: wasi libc is also linked in when using wasm32-freestanding
reproduction:
```
touch test.c
zig build-lib -dynamic -target wasm32-freestanding -lc test.c
```

will yield the following error:
```
error(compilation): clang failed with stderr: /home/felix/software/zig-linux-x86_64-0.8.0-dev.2635+b8f85a805/lib/libc/wasi/libc-bottom-half/crt/crt1.c:1:10: fatal error: 'wasi/api.h' file not found

error(compilation): clang failed with stderr: /home/felix/software/zig-linux-x86_64-0.8.0-dev.2635+b8f85a805/lib/libc/wasi/libc-bottom-half/crt/crt1-command.c:1:10: fatal error: 'wasi/api.h' file not found

error(compilation): /home/felix/software/zig-linux-x86_64-0.8.0-dev.2635+b8f85a805/lib/libc/wasi/libc-bottom-half/crt/crt1.c:1:1: unable to build C object: clang exited with code 1
error(compilation): /home/felix/software/zig-linux-x86_64-0.8.0-dev.2635+b8f85a805/lib/libc/wasi/libc-bottom-half/crt/crt1-command.c:1:1: unable to build C object: clang exited with code 1
error: unable to build WASI libc sysroot: BuildingLibCObjectFailed
```

cc @kubkon

kubkon:
~~This commit includes emulated libc sublibs that were not included in the compilation and caching of WASI libc that ships with Zig. The libs include (emulated): process clocks, getpid, mman, and signal.~~

With this change, it is now possible to successfully cross-compile `wasm3` engine to WASI with `zig cc`.

~~For the future though, it might be worth considering splitting WASI libc into libc-proper and modularised emulated libs as it is done in upstream, and then have them included only if the user specifically requests emulation/parts of it.~~

EDIT:

This PR now replicates the expected behavior when using `clang` with upstream `wasi-libc` sysroot: linking emulated subcomponents such as process clocks or signals requires an explicit link flag in the compiler invocation, for example:

```
zig cc -target wasm32-wasi -lwasi-emulated-process-clocks main.c -o main.wasm
```

In `build.zig`, accessing emulated subcomponents is equivalent to linking a system lib, e.g.,:

```zig
    const exe = b.addExecutable("main", null);
    exe.linkSystemLibrary("wasi-emulated-process-clocks");
}
```

EDIT2:

I've also streamlined the compilation logic a little in that we only compile those emulated components that were requested and cache them:

> Move parsing of system libs into `main.zig` next to where we decide
if we should link libC, and, if targeting WASI, if the specified
libname equals one of the emulated components, save it on the side
and remove it from the system libs. Then, build *only* those parts
of WASI libc that were preserved in the previous step.

> This also fixes building of different crt1 bits needed to support
reactors and commands.

EDIT3:

Refactoring of linker logic fixes #8886.

cc @MasterQ32

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -651,7 +651,7 @@ pub const InitOptions = struct {
     /// * getpid
     /// * mman
     /// * signal
-    wasi_emulated_libs: []const []const u8 = &[0][]const u8{},
+    wasi_emulated_libs: []const wasi_libc.CRTFile = &[0]wasi_libc.CRTFile{},
     link_libc: bool = false,
     link_libcpp: bool = false,
     link_libunwind: bool = false,
@@ -1434,11 +1434,11 @@ pub fn create(gpa: *Allocator, options: InitOptions) !*Compilation {
             });
         }
         if (comp.wantBuildWasiLibcFromSource()) {
-            try comp.work_queue.ensureUnusedCapacity(6); // worst-case we need all components
             const wasi_emulated_libs = comp.bin_file.options.wasi_emulated_libs;
-            for (wasi_emulated_libs) |lib_name| {
+            try comp.work_queue.ensureUnusedCapacity(wasi_emulated_libs.len + 2); // worst-case we need all components
+            for (wasi_emulated_libs) |crt_file| {
                 comp.work_queue.writeItemAssumeCapacity(.{
-                    .wasi_libc_crt_file = wasi_libc.getEmulatedLibCRTFile(lib_name).?,
+                    .wasi_libc_crt_file = crt_file,
                 });
             }
             // TODO add logic deciding which crt1 we want here.
@@ -4157,12 +4157,6 @@ pub fn stage1AddLinkLib(comp: *Compilation, lib_name: []const u8) !void {
     if (comp.bin_file.options.skip_linker_dependencies) return;
 
     // This happens when an `extern "foo"` function is referenced by the stage1 backend.
-    if (comp.getTarget().os.tag == .wasi and mem.eql(u8, "wasi_snapshot_preview1", lib_name)) {
-        // Any referenced symbol from this lib, will be undefined until
-        // runtime as this lib is provided directly by the runtime.
-        return;
-    }
-
     // If we haven't seen this library yet and we're targeting Windows, we need to queue up
     // a work item to produce the DLL import library for this.
     const gop = try comp.bin_file.options.system_libs.getOrPut(comp.gpa, lib_name);
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -4157,6 +4157,12 @@ pub fn stage1AddLinkLib(comp: *Compilation, lib_name: []const u8) !void {
     if (comp.bin_file.options.skip_linker_dependencies) return;
 
     // This happens when an `extern "foo"` function is referenced by the stage1 backend.
+    if (comp.getTarget().os.tag == .wasi and mem.eql(u8, "wasi_snapshot_preview1", lib_name)) {
+        // Any referenced symbol from this lib, will be undefined until
+        // runtime as this lib is provided directly by the runtime.
+        return;
+    }
+
     // If we haven't seen this library yet and we're targeting Windows, we need to queue up
     // a work item to produce the DLL import library for this.
     const gop = try comp.bin_file.options.system_libs.getOrPut(comp.gpa, lib_name);
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -203,8 +203,8 @@ const Job = union(enum) {
     /// needed when not linking libc and using LLVM for code generation because it generates
     /// calls to, for example, memcpy and memset.
     zig_libc: void,
-    /// WASI libc sysroot
-    wasi_libc_sysroot: void,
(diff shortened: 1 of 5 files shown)


kubkon:
I'm making the PR into a draft PR since I'm not happy with the fact that we bundle emulations in wasi-libc proper. As far as I can work out, that should not be the case, and instead, it should be up to the user to explicitly request emulation libs via `-l` flags. For example:

```
/home/kubkon/dev/zig/build/lib/zig/libc/include/wasm-wasi-musl/signal.h:2:2: error: "wasm lacks signal support; to enable minimal signal emulation, compile with -D_WASI_EMULATED_SIGNAL and link with -lwasi-emulated-signal"
#error "wasm lacks signal support; to enable minimal signal emulation, \
```

I think I'd like us to match the hinted behaviour in the error message within zig itself.

Summoning @jedisct1 to offer his insight and opinion.

jedisct1:
These emulation libraries are very useful to port software to WASI.

And using `-l` flags would totally make sense to enable them.

src/Compilation.zig
```diff
@@ -646,6 +646,12 @@ pub const InitOptions = struct {
     framework_dirs: []const []const u8 = &[0][]const u8{},
     frameworks: []const []const u8 = &[0][]const u8{},
     system_libs: []const []const u8 = &[0][]const u8{},
+    /// These correspond to the WASI libc emulated subcomponents including:
+    /// * process clocks
+    /// * getpid
+    /// * mman
+    /// * signal
+    wasi_emulated_libs: []const []const u8 = &[0][]const u8{},
```
andrewrk, line 654: These should be enum items and not strings since the strings are checked by the CLI, right?

src/Compilation.zig
```diff
@@ -1426,8 +1433,19 @@ pub fn create(gpa: *Allocator, options: InitOptions) !*Compilation {
                 },
             });
         }
-        if (comp.wantBuildWasiLibcSysrootFromSource()) {
-            try comp.work_queue.write(&[_]Job{.{ .wasi_libc_sysroot = {} }});
+        if (comp.wantBuildWasiLibcFromSource()) {
+            try comp.work_queue.ensureUnusedCapacity(6); // worst-case we need all components
```
andrewrk, line 1437:
```suggestion
            try comp.work_queue.ensureUnusedCapacity(wasi_emulated_libs.len + 2);
```

src/Compilation.zig
```diff
@@ -4138,6 +4157,12 @@ pub fn stage1AddLinkLib(comp: *Compilation, lib_name: []const u8) !void {
     if (comp.bin_file.options.skip_linker_dependencies) return;
 
     // This happens when an `extern "foo"` function is referenced by the stage1 backend.
+    if (comp.getTarget().os.tag == .wasi and mem.eql(u8, "wasi_snapshot_preview1", lib_name)) {
+        // Any referenced symbol from this lib, will be undefined until
+        // runtime as this lib is provided directly by the runtime.
+        return;
+    }
```
andrewrk, line 4164: I don't quite understand why this is here. The comment doesn't make sense to me because that is true about all dynamically linked libraries. What makes `wasi_snapshot_preview1` special?

andrewrk (changes requested):

src/Compilation.zig
kubkon: `wasi_snapshot_preview1` is actually not a dynamic library per se; it's a set of syscalls implicitly provided by the sandbox, therefore we cannot really link against it. There are two options here: we either remove any mention of `wasi_snapshot_preview1` from the linker line (which will be there since we import syscalls directly into our libstd), or disable linking dynamically by `wasm-ld` (which was the case until now). I'm happy either way, and if you think we should preserve the status quo, I'll delete this bit, and remove any mention of dynamic linking from `wasm-ld` wrapper in `Wasm.zig`.

andrewrk: I think this logic would make more sense to me if it was in link/Wasm.zig which sounds like one of your suggestions :+1:

src/Compilation.zig
kubkon: The strings correspond to the names of required emulated WASI libc subcomponents imported as libraries on the linker line, e.g., `-lwasi-emulated-process-clocks` would result in us saving `wasi-emulated-process-clocks` in `wasi_emulated_libs` slice. We could store a slice of enums too, but we still need to get the name of the matching pre-compiled lib using `comp.get_libc_crt_file(arena, "libwasi-emulated-process-clocks.a")`. I'm fine either way TBH, and if you feel that matching strings to enum variants in `main.zig` makes more sense, I'm happy to make the necessary tweaks.

src/Compilation.zig
kubkon: Oh, actually. I'll move this logic into `link/Wasm.zig` but I cannot get rid of the logic which unpacks `system_libs` since we are allowed to pass static libs via `-l` flags.

andrewrk (approved):

PR #9006, opened 2021-06-06
Add zig_llvm-ar.cpp in build.zig

nico-abram:
zig_llvm-ar.cpp was added to the cmake file in 0afb5b2ec6de494efe0605b2270f6a01b95237af 8 days ago, but it seems it was not added in build.zig

I got this error when following [the Building Zig on Windows instructions](https://github.com/ziglang/zig/wiki/Building-Zig-on-Windows#option-1-use-the-windows-zig-compiler-dev-kit):

```
lld-link: error: undefined symbol: ZigLlvmAr_main
```

This patch fixes it


andrewrk:
Thanks!

PR #9008, opened 2021-06-06
Add helpers to align pointers

LemonBoy:
See the individual commits for more details.

As a side note, please, please, please stop using `@intToPtr` and `@ptrToInt` everywhere.

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -2323,6 +2323,71 @@ pub fn nativeToBig(comptime T: type, x: T) T {
     };
 }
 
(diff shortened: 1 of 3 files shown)


lib/std/mem.zig
```diff
@@ -2323,6 +2323,71 @@ pub fn nativeToBig(comptime T: type, x: T) T {
     };
 }
 
+/// Returns the number of elements that, if added to the given pointer, align it
+/// to a multiple of the given quantity.
+/// Returns an aligned pointer or null if one of the following conditions is
+/// met:
```
xackus, line 2329: Bad copy-paste

PR #9009, opened 2021-06-06
std: Fix some BPF fn definitions

LemonBoy:
Cros-checked with bpf.h of kernel 5.12.9.
[Link for the lazy](https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/bpf.h)

cc @mattnite


mattnite:
thanks for that catch

PR #9010, opened 2021-06-06
openbsd: adjust dynamic linker path

semarie:
the dynamic linker on OpenBSD is at `/usr/libexec/ld.so` (and not `/libexec/ld.so`)

with this patch, I am able to crosscompile from x86_64-openbsd to i386-openbsd (using a libc file)


jedisct1:
Thank you! (and for all you've been doing to bring Zig to OpenBSD)

PR #9012, opened 2021-06-06
std: Better formatting of tuple types

LemonBoy:
Skip the useless type name and the numeric field names.

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -495,6 +495,22 @@ pub fn formatType(
             }
         },
         .Struct => |info| {
+            if (info.is_tuple) {
+                // Skip the type and field names when formatting tuples.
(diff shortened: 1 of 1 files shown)


lib/std/fmt.zig
```diff
@@ -2164,6 +2180,9 @@ test "struct" {
     };
 
     try expectFmt("S{ .a = 456, .b = error.Unused }", "{}", .{inst});
+    // Tuples
+    try expectFmt("{ }", "{}", .{.{}});
```
MasterQ32, line 2184: I would add another test case for single-item tuples as those are another edge case (no trailing comma)

LemonBoy: Good idea, added one (and force-pushed).

PR #9015, opened 2021-06-06
std.os: (p)writev should perform partial writes if iov.len > IOV_MAX

daurnimator:
@g-w1 ran into an issue where `pwritevAll` failed with EINVAL when passed more iovecs than `IOV_MAX`.

We may need to track down `IOV_MAX` values for more OS-es before this is mergable


g-w1:
Can a test be added for this?

Issue #9017, opened 2021-06-06
unconfigurable install prefix lib directory


Jan200101:
https://github.com/ziglang/zig/blob/7d15a3ac71c5d8dc8c08dfd8ea8ad43d4eae188a/lib/std/build.zig#L195-L205

zig will always install libraries into prefix/lib
this is problematic when the desired prefix uses prefix/lib for a different format than the native one

PR #9018, opened 2021-06-06
specify the output lib directory with a flag

Jan200101:
closes https://github.com/ziglang/zig/issues/9017

by default zig will output libraries into PREFIX/lib, which is a sane default because lib\<qual\> are optional, but this may create conflicts on system where PREFIX/lib is not for the native arch

the variable names, flag or help message might need changing to be more descriptive

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -190,7 +190,7 @@ pub const Builder = struct {
     }
 
     /// This function is intended to be called by std/special/build_runner.zig, not a build.zig file.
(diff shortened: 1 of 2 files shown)


ifreund:
For reference, meson's logic for determining this default: 
https://github.com/mesonbuild/meson/blob/240434b167468f3fcad1d16b39c9aaf515b97eb6/mesonbuild/mesonlib/universal.py#L870-L886

ifreund:
After thinking about this a bit more, I don't think copying meson and inspecting the host system makes sense for zig where cross-compilation is a first class use-case

PR #9019, opened 2021-06-06
Add Linux XDP bits

Aransentin:
Add the XDP bits necessary for extreme sanic speed packet processing on Linux.

CI testing it would be iffy as it requires a whole lot of code to make even a "hello word" program (including building a separate eBPF XDP program etc), and a available network interface on the server which is liable to be left in an unclean state.


andrewrk:
The new file does not appear to be exposed

Aransentin:
Oh yeah, sorry about that. Actually being able to use it would be a good feature, yes :)

PR #9020, opened 2021-06-06
specify the output lib, exe and include paths with flags

### LINKED ISSUE #9017: unconfigurable install prefix lib directory

Jan200101:
closes #9017
supersedes #9018 

allows specifying full paths of where to install binaries, libraries and headers and so on

ifreund had suggested using a similar approach as meson and detecting what lib directory is available but since the default output directory is local I left that out

### DIFF
--- a/lib/std/special/build_runner.zig
+++ b/lib/std/special/build_runner.zig
@@ -61,6 +61,8 @@ pub fn main() !void {
     const stdout_stream = io.getStdOut().writer();
 
     var install_prefix: ?[]const u8 = null;
+    var dir_list = Builder.DirList{};
+
     while (nextArg(args, &arg_idx)) |arg| {
         if (mem.startsWith(u8, arg, "-D")) {
             const option_contents = arg[2..];
@@ -84,6 +86,21 @@ pub fn main() !void {
                 return usage(builder, false, stdout_stream);
             } else if (mem.eql(u8, arg, "-p") or mem.eql(u8, arg, "--prefix")) {
                 install_prefix = nextArg(args, &arg_idx) orelse {
+                    warn("Expected argument after {s}\n\n", .{arg});
+                    return usageAndErr(builder, false, stderr_stream);
+                };
+            } else if (mem.eql(u8, arg, "--lib-dir")) {
+                dir_list.lib_dir = nextArg(args, &arg_idx) orelse {
+                    warn("Expected argument after {s}\n\n", .{arg});
+                    return usageAndErr(builder, false, stderr_stream);
+                };
+            } else if (mem.eql(u8, arg, "--exe-dir")) {
+                dir_list.exe_dir = nextArg(args, &arg_idx) orelse {
+                    warn("Expected argument after {s}\n\n", .{arg});
+                    return usageAndErr(builder, false, stderr_stream);
+                };
+            } else if (mem.eql(u8, arg, "--include-dir")) {
+                dir_list.include_dir = nextArg(args, &arg_idx) orelse {
                     warn("Expected argument after {s}\n\n", .{arg});
                     return usageAndErr(builder, false, stderr_stream);
                 };
@@ -135,7 +152,7 @@ pub fn main() !void {
         }
     }
 
-    builder.resolveInstallPrefix(install_prefix);
+    builder.resolveInstallPrefix(install_prefix, dir_list);
     try runBuild(builder);
 
     if (builder.validateUserInputDidItFail())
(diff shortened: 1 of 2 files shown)


lib/std/special/build_runner.zig
```diff
@@ -189,6 +207,9 @@ fn usage(builder: *Builder, already_ran_build: bool, out_stream: anytype) !void
         \\  -h, --help                  Print this help and exit
         \\  --verbose                   Print commands before executing them
         \\  -p, --prefix [path]         Override default install prefix
+        \\  --output-lib-dir [path]     Override default library directory path
+        \\  --output-exe-dir [path]     Override default executable directory path
+        \\  --output-include-dir [path] Override default include directory path
```
ifreund, line 212: Maybe we should mention that these are relative to the install prefix

Jan200101: https://github.com/ziglang/zig/blob/b8588f7ab4458c21d105115c2fc8a2acbb456c9a/lib/std/build.zig#L202-L204
the paths are not relative to the install prefix

the default include and exe dir values follow the FHS so I don't think changing the directory relative to the prefix is worthwhile

though the lib dir is rather ambigious because of the optional quad and that maybe should have its own flag

Jan200101: fair point

I've pushed code that takes dest_dir into consideration when specifying the directories though I am sure the code could be improved since I am not well versed in Zig

ifreund: How about we support both absolute and relative paths. If a relative path is passed it is considered relative to the install prefix.

Jan200101: That is a fair expectation.
Added it to the code

ifreund (commented):
I think the current logic makes sense, I wonder if we should drop the `output` prefix and just use `--lib-dir` `--bin-dir` and `--include-dir`. I could see come confusion arising with the current `--output-lib-dir` variants due to the fact that `zig build-exe` has a `--output-dir` option which does something entirely different.

We should also keep in mind that there are many other system directories that we probably want to eventually support including `/etc` (requires special handling with the `/usr/` install prefix to adhear to the FHS) and the `share` dir used for non-architecture specific data.

Jan200101:
I agree but I am afraid that the `--override-lib-dir` flag might confuse people

Its probably a good idea to start working out implementations for /etc and share, is there anything already in place to allow installation to such places?
Looking further ahead perhaps it is better to replace the list of arguments with map that includes all of them instead, making the process easier in the future

Jan200101:
I've gone ahead and removed the output prefix from the flags though the override-lib-dir flag is still up to debate

I've also replaced the list of arguments resolveInstallPrefix takes with a StringHashMap to simplify things

ifreund (changes requested):
A HashMap is not the right type for this as all the possible fields are known in advance. I’d use a struct with `?[]const u8` fields.

daurnimator:
> A HashMap is not the right type for this as all the possible fields are known in advance.

I don't think that rules out use of a hashmap.

However, I think here the correct structure would be:

```zig
struct {
    lib_dir: ?[]const u8,
    exe_dir: ?[]const u8,
    include_dir: ?[]const u8,
}
```

----

You should *never* be able to opt-out of DESTDIR: it should be prefixed to *all* paths where final artifacts end up.

Jan200101:
> A HashMap is not the right type for this as all the possible fields are known in advance. I’d use a struct with default null `?[]const u8` fields.

Okay

> You should never be able to opt-out of DESTDIR: it should be prefixed to all paths where final artifacts end up.

With the current logic you can't

```zig
var lib_list = [_][]const u8{ self.install_path, "lib" };
if (dir_list.lib_dir) |dir| {
    if (std.fs.path.isAbsolute(dir)) lib_list[0] = self.dest_dir orelse "";
    lib_list[1] = dir;
}
```
`self.install_path` is made up of the dest_dir, if it isn't null, and the install_prefix
if one of the paths given is absolute we replace the install_path with dest_dir so we don't have to rely on the default prefix

andrewrk:
Note 193c529b8c9529c4303aeb3a8864dfee9b1d2485 and 7d5538c32b7acfb6878be4178af2f71977d0cf9d which I pushed to master branch.

Idea here being `--prefix` is the "normal" flag to use and these overrides are more advanced. The longer names and shared first part of the name indicate the connection between the flags.

PR #9023, opened 2021-06-07
Limit Fixed Buffer Stream seekTo

Vulfox:
This PR should limit the function `seekTo` for fixed buffer streams to not exceed the maximum position of the reader's buffer. Without this, the pos could be placed outside of the maximum buffer range and causes integer overflow panics upon reading bytes.

I have a hunch this is a bug as the `seekBy` function also performs the same limitations as this PR on line 99 below.

https://github.com/ziglang/zig/blob/fc8791c133c14fe969b9056d8f0e337094ce1461/lib/std/io/fixed_buffer_stream.zig#L87-L101

### DIFF
--- a/lib/std/io/fixed_buffer_stream.zig
+++ b/lib/std/io/fixed_buffer_stream.zig
@@ -155,6 +155,9 @@ test "FixedBufferStream output 2" {
 
     try testing.expectError(error.NoSpaceLeft, fbs.writer().writeAll("Hello world!"));
     try testing.expect(mem.eql(u8, fbs.getWritten(), "Hello worl"));
+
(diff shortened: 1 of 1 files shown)


LemonBoy:
LGTM but needs a test. Add a `seekTo` call in `FixedBufferStream output 2` and `FixedBufferStream input` to ensure there's no overflow when calling `read`.

Vulfox:
@LemonBoy added `read` and I assumed since you brought up `FixedBufferStream output 2` a `write` test for buffer out of bounds `seekTo`s.

I did take the liberty to converting `read(dest[0..4])` to `read(&dest)` in the `FixedBufferStream input` test, if that is alright.

Vexu:
Thanks!

PR #9025, opened 2021-06-07
further haiku support build.zig

hoanga:
Hello again,

Congratulations Zig team on the release! Sorry I was not available then but I have been making some more progress for Haiku support on Zig. The following changes should get Zig  to use the Zig build system to compile programs under Haiku. This was tested on nightly build hrev55136.

Things I have been able to confirm now functional that wasn't previously (previous commands that were working under Haiku should still be):

    zig build in a (simple) zig project works (yay)

When running make install to attempt to install Zig after compiling, there is still an error being generated, unfortunately. This is on my future todo to try to address (along with debug info).

```
~/src/git/zig/build> mkdir hellohaiku

~/src/git/zig/build> cd hellohaiku

~/src/git/zig/build/hellohaiku> uname -a
Haiku shredder 1 hrev55136 Jun  6 2021 06:26:21 x86_64 x86_64 Haiku

~/src/git/zig/build/hellohaiku> ../zig init-exe
info: Created build.zig
info: Created src/main.zig
info: Next, try `zig build --help` or `zig build run`

~/src/git/zig/build/hellohaiku> ../zig build --verbose run
/boot/home/src/git/zig/build/zig build-exe /boot/home/src/git/zig/build/hellohaiku/src/main.zig --cache-dir /boot/home/src/git/zig/build/hellohaiku/zig-cache --global-cache-dir /boot/home/config/cache/zig --name hellohaiku --enable-cache 
cp /boot/home/src/git/zig/build/hellohaiku/zig-cache/o/e4fd5d75a03c2a1c959b8e3ce75b06b8/hellohaiku /boot/home/src/git/zig/build/hellohaiku/zig-out/bin/hellohaiku # installed
/boot/home/src/git/zig/build/hellohaiku/zig-out/bin/hellohaiku 
info: All your codebase are belong to us.

~/src/git/zig/build/hellohaiku> ../zig version
0.9.0-dev.1+b755c46d7


### try using zig build on some other zig codebase 
~/src/git/zig/build/ziglings> ../zig build 08
Compiling 08_quiz.zig...
Checking 08_quiz.zig...
PASSED: Program in Zig!
```

### DIFF
--- a/lib/std/crypto/tlcsprng.zig
+++ b/lib/std/crypto/tlcsprng.zig
@@ -72,6 +72,11 @@ fn tlsCsprngFill(_: *const std.rand.Random, buffer: []u8) void {
         return fillWithOsEntropy(buffer);
     }
 
+    // Use OS syscall for haiku
+    if (std.Target.current.os.tag == .haiku) {
+        return fillWithOsEntropy(buffer);
+    }
+
     if (wipe_mem.len == 0) {
         // Not initialized yet.
         if (want_fork_safety and maybe_have_wipe_on_fork) {
--- a/lib/std/os/bits/haiku.zig
+++ b/lib/std/os/bits/haiku.zig
@@ -179,15 +179,29 @@ pub const dirent = extern struct {
     }
 };
 
+pub const area_info = extern struct {
+    area: u32,
+    name: [32]u8,
+    size: usize,
+    lock: u32,
+    protection: u32,
+    team_id: i32,
+    ram_size: u32,
+    copy_count: u32,
+    in_count: u32,
+    out_count: u32,
+    address: *c_void,
+};
+
 pub const image_info = extern struct {
     id: u32,
-    type: u32,
+    image_type: u32,
     sequence: i32,
     init_order: i32,
     init_routine: *c_void,
     term_routine: *c_void,
     device: i32,
-    node: i32,
+    node: i64,
     name: [1024]u8,
     text: *c_void,
     data: *c_void,
@@ -223,6 +237,19 @@ pub const system_info = extern struct {
     kernel_build_time: [32]u8,
     kernel_version: i64,
     abi: u32,
+};
+
+pub const team_info = extern struct {
+    team_id: i32,
+    thread_count: i32,
+    image_count: i32,
+    area_count: i32,
+    debugger_nub_thread: i32,
+    debugger_nub_port: i32,
+    argc: i32,
+    args: [64]u8,
+    uid: uid_t,
+    gid: gid_t,
 };
 
 pub const in_port_t = u16;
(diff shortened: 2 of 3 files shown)


lib/std/crypto/tlcsprng.zig
```diff
@@ -72,6 +72,11 @@ fn tlsCsprngFill(_: *const std.rand.Random, buffer: []u8) void {
         return fillWithOsEntropy(buffer);
     }
 
+    // Use OS syscall for haiku
+    if (std.Target.current.os.tag == .haiku) {
+        return fillWithOsEntropy(buffer);
+    }
```
andrewrk, line 78: Here are some questions to help decide the best implementation here:
 * does haiku support fork() ?
   * if so, does haiku support MADV_WIPEONFORK?
 * does haiku support `arc4random_buf` or equivalent? or is /dev/urandom the only way to get entropy from the OS?
 * does haiku support threadlocal variables?

lib/std/os/bits/haiku.zig
```diff
@@ -383,7 +407,7 @@ pub const O_SYNC = 0x0080;
 pub const O_RSYNC = 0o4010000;
 pub const O_DIRECTORY = 0x20000;
 pub const O_NOFOLLOW = 0x0100;
-pub const O_CLOEXEC = 0x00100000;
+pub const O_CLOEXEC = 0x00000040;
```
andrewrk, line 410: what was the cause of these numbers being incorrect? do all the bits need to be double-checked now?

lib/std/crypto/tlcsprng.zig
hoanga: Appreciate the feedback on this one. I have some doubts myself on this current implementation as-is but do know that it is functional.

From what I can tell, Haiku does support fork() although not intricately familiar with its implementation of fork. I don't think it supports MADV_WAVEONFORK from my short look for it. I can try to dig in a little more and see if it does support it.

Haiku does have appear to have an arc4random() call available as part of its FreeBSD network compatibility layer (think an extra lib will need to be linked) but am unsure if this is something that is a general usage or more tailored to work for Haiku's network stack. My guess is that /dev/urandom is not the only mechanism available (https://dev.haiku-os.org/ticket/14937 might be useful)
 
As for threadlocal variables, I am not sure what support has for this. I am going to spend some more time to look into this.

Thanks for the feedback! I am going to spend some more time to think about if there are better approaches to handling this.

lib/std/os/bits/haiku.zig
hoanga: i believe the cause of these numbers being incorrect was a copy and paste on my part from one of the bsd ports. some numbers initially matched which probably contributed to my (false) assumption that it was basically the same. on retrospect, that was probably a hasty decision on my part .

i think it's prudent that at least the Haiku bits are double checked. i will add it to my list of things to do and perform any clean up for that.

lib/std/crypto/tlcsprng.zig
andrewrk: If it helps - the best case scenario is that haiku has arc4random_buf or equivalent. If that is the case, there is an easy solution to this implementation.

lib/std/os/bits/haiku.zig
andrewrk: Thank you!

andrewrk:

hoanga:
pardon for being absent for awhile with this pr, other things in real life have been cropping up and requiring my attention. these changesets unfortunately fell by the wayside as a consequence. thank you for pruning it as the patches at this point are most likely too stale to apply cleanly. hopefully if time opens up, i can get a little bit of time to send updated patches.

andrewrk:
Take your time, no rush at all of course. Hope you are well. Cheers!

PR #9028, opened 2021-06-07
Breaking changes for std.Once

LemonBoy:
cc @kprotty

### DIFF
--- a/lib/std/once.zig
+++ b/lib/std/once.zig
@@ -30,8 +30,9 @@ pub fn Once(comptime f: anytype) type {
         .Void => true,
         .ErrorUnion => |info| info.payload == void,
         else => false,
-    })
+    }) {
         @compileError("expected function returning void or !void, found " ++ @typeName(T));
+    }
 
     const Args = std.meta.ArgsTuple(T);
 
--- a/lib/std/once.zig
+++ b/lib/std/once.zig
@@ -7,36 +7,54 @@ const std = @import("std.zig");
 const builtin = std.builtin;
 const testing = std.testing;
 
-pub fn once(comptime f: fn () void) Once(f) {
+pub fn once(comptime f: anytype) Once(f) {
     return Once(f){};
 }
 
 /// An object that executes the function `f` just once.
-pub fn Once(comptime f: fn () void) type {
+pub fn Once(comptime f: anytype) type {
+    const T = @TypeOf(f);
+    const info = @typeInfo(T);
+    if (info != .Fn)
+        @compileError("expected function type, found " ++ @typeName(T));
+
+    const fn_info = info.Fn;
+    if (fn_info.is_generic)
+        @compileError("cannot instantiate Once with a generic function");
+    if (fn_info.is_var_args)
+        @compileError("cannot instantiate Once with a variadic function");
+
+    if (@typeInfo(fn_info.return_type.?) != .Void)
+        @compileError("expected function returning void, found " ++ @typeName(T));
+
+    const Args = std.meta.ArgsTuple(T);
+
     return struct {
         done: bool = false,
         mutex: std.Thread.Mutex = std.Thread.Mutex{},
 
-        /// Call the function `f`.
+        /// Call the function `f` with the supplied arguments, use .{} if the
+        /// function takes no arguments.
         /// If `call` is invoked multiple times `f` will be executed only the
-        /// first time.
+        /// first time. The arguments are unused for every call but the first
+        /// one.
         /// The invocations are thread-safe.
-        pub fn call(self: *@This()) void {
+        pub fn call(self: *@This(), args: Args) void {
             if (@atomicLoad(bool, &self.done, .Acquire))
                 return;
 
-            return self.callSlow();
+            return self.callSlow(args);
         }
 
-        fn callSlow(self: *@This()) void {
+        fn callSlow(self: *@This(), args: Args) void {
             @setCold(true);
 
-            const T = self.mutex.acquire();
-            defer T.release();
+            const mut = self.mutex.acquire();
+            defer mut.release();
 
             // The first thread to acquire the mutex gets to run the initializer
             if (!self.done) {
-                f();
+                @call(.{}, f, args);
                 @atomicStore(bool, &self.done, true, .Release);
             }
         }
@@ -46,24 +64,24 @@ pub fn Once(comptime f: fn () void) type {
 var global_number: i32 = 0;
 var global_once = once(incr);
 
-fn incr() void {
-    global_number += 1;
+fn incr(x: i32) void {
+    global_number += x;
 }
 
 test "Once executes its function just once" {
     if (builtin.single_threaded) {
-        global_once.call();
-        global_once.call();
+        global_once.call(.{1});
+        global_once.call(.{1});
     } else {
-        var threads: [10]*std.Thread = undefined;
+        var threads: [4]*std.Thread = undefined;
         defer for (threads) |handle| handle.wait();
 
         for (threads) |*handle| {
             handle.* = try std.Thread.spawn(struct {
                 fn thread_fn(x: u8) void {
-                    global_once.call();
+                    global_once.call(.{x});
                 }
-            }.thread_fn, 0);
+            }.thread_fn, 1);
         }
     }
 
--- a/lib/std/once.zig
+++ b/lib/std/once.zig
@@ -14,49 +14,61 @@ pub fn once(comptime f: anytype) Once(f) {
 /// An object that executes the function `f` just once.
 pub fn Once(comptime f: anytype) type {
     const T = @TypeOf(f);
-    const info = @typeInfo(T);
-    if (info != .Fn)
+    const ty_info = @typeInfo(T);
+    if (ty_info != .Fn)
         @compileError("expected function type, found " ++ @typeName(T));
 
-    const fn_info = info.Fn;
+    const fn_info = ty_info.Fn;
     if (fn_info.is_generic)
         @compileError("cannot instantiate Once with a generic function");
     if (fn_info.is_var_args)
         @compileError("cannot instantiate Once with a variadic function");
 
-    if (@typeInfo(fn_info.return_type.?) != .Void)
-        @compileError("expected function returning void, found " ++ @typeName(T));
+    // The return type must be void or !void.
+    const Return = fn_info.return_type.?;
+    if (!switch (@typeInfo(Return)) {
+        .Void => true,
+        .ErrorUnion => |info| info.payload == void,
+        else => false,
+    })
+        @compileError("expected function returning void or !void, found " ++ @typeName(T));
 
     const Args = std.meta.ArgsTuple(T);
 
+    // The field order minimizes the amount of space wasted to padding.
     return struct {
-        done: bool = false,
         mutex: std.Thread.Mutex = std.Thread.Mutex{},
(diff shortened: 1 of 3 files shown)


lib/std/once.zig
```diff
@@ -7,65 +7,119 @@ const std = @import("std.zig");
...
+    if (fn_info.is_var_args)
+        @compileError("cannot instantiate Once with a variadic function");
+
+    // The return type must be void or !void.
+    const Return = fn_info.return_type.?;
+    if (!switch (@typeInfo(Return)) {
+        .Void => true,
+        .ErrorUnion => |info| info.payload == void,
+        else => false,
+    })
```
kprotty, line 33: style nitpick: maybe add { } around the if-body to help distinguish that its apart apart of the `if` given the condition is also indented.

marler8997:
I liked this at first but after thinking on it, the current interface that requires a function with no arguments helps avoids problems where `once` could be called with different arguments.  I think that calling `once` with different arguments would usually be an error.  Emulating this error with the current interface would be harder since it would require the `once` function to read from global mutable state which immediately raises red flags to the programmer that they need to be careful.

In my opinion, the current interface makes it harder to shoot oneself in the foot.  That said, there are actually cases where calling `once` with different arguments is a valid solution.  The WSAStartup example is one where a client could call wsaStartupOnce with a different version, but then `std` could also call `once` with a different version as a sort of "default" when the client doesn't specify one.

So maybe this is fine, but I wanted to share my thoughts/concerns in case anyone has any ideas that could alleviate these concerns altogether.

lib/std/once.zig
```diff
@@ -7,65 +7,120 @@ const std = @import("std.zig");
...
-        /// first time.
+        /// first time. The arguments are unused for every call but the first
+        /// one.
+        /// If `f` returns an error the initialization is marked as done and,
+        /// for every following call, the same error value is returned.
         /// The invocations are thread-safe.
-        pub fn call(self: *@This()) void {
-            if (@atomicLoad(bool, &self.done, .Acquire))
-                return;
+        pub fn call(self: *@This(), args: Args) Return {
```
marler8997, line 53: Is there a way we could have this function return whether or not it performed the initialization?

LemonBoy:
> In my opinion, the current interface makes it harder to shoot oneself in the foot.

The very first sentence of the documentation states `f` is only called once, the arguments values do not matter for every invocation after the first one and that looks pretty clear to me. Passing in different values is not an error if you understand what this synchronization primitive does.

>  In my opinion, the current interface makes it harder to shoot oneself in the foot.

Not really, having to route parameters/errors trough globals is a much bigger footgun and the WSA example is a good case in point.

> The WSAStartup example is one where a client could call wsaStartupOnce with a different version

That specific case is pure overengineering since the latest version 2.2 is backwards compatible with the previous ones, the stdlib assumes the whole set of winsock API to be available and, last but not least, older Winsock versions are really old (decades old) and not useful.

lib/std/once.zig
LemonBoy: What's the use case for that?

kprotty: It really feels like there should be one but can't think of any ATM.

marler8997: In the example of WSAStartup, if this is used as a "fallback initialization" routine, then if the call returns that it has already been initialized, then the function can fail right away rather than try to call the function that failed again, i.e.


```zig
fn someWsaFunc() {
    callTheFunc() catch |e| switch (e) {
        error.WsaNotInitialized => {
            if (!wsaStartupOnce.call(2, 2)) {
                unreachable,
            }
            return someWsaFunc();
        }
    };
}

```

LemonBoy: > then the function can fail right away rather than try to call the function that failed again

The function is called once, no matter if an error is returned or not. There's no way to disarm the `Once` object (it wouldn't be called `once` then) and the error condition is propagated to every caller, be it from another thread or at another time.

Providing a way to check if the function was called is possible but the idea here is to call `call` and call it a day (heh), the check is lightweight and avoids any possible data race.

marler8997: @LemonBoy, sorry my message was not clear.  The caller can try to call the function that failed again, not the "once function" again.  I think the code example I shared explains the use case.

LemonBoy: I still don't understand what your point is. What is the function that fails?

marler8997: `callTheFunc` is a WSA function that fails.  We've decided that is a WSA function fails, we will call WSAStartup and then try calling it again. In this case, we use `std.once` to call wsastartup, so if we detect that it's already been called, then we forego calling the function again.

LemonBoy: That doesn't make sense, you should call `call()` anyway and forget about it. That's the whole point of `once`.

marler8997: We are calling it anyway.  Again, I'm not saying we're not calling `call`, I'm saying after we call `call`, if we know whether or not we initialized, then we know whether or not to call the function that failed again or just to fail immediately. Look at the code example above.

LemonBoy: I see. My offer here is to add an `wasCalled` predicate (or with a better name).

marler8997: Feels like a race condition.

marler8997: But now that I think about it, the whole use case I've presented here is also a race condition :)

marler8997: I'm fine committing as is and waiting for a real use case before adding this.

LemonBoy: > Feels like a race condition.

Nope. The predicate returns `true` if the function was called and that's it, whenever you see it returning `false` you can't make any assumption about the state of the `Once` instance, that's what makes this kind of predicate tricky to use.

> I'm fine committing as is and waiting for a real use case before adding this.

Thank you, kind sir.

marler8997: > Nope. The predicate returns true if the function was called and that's it, whenever you see it returning false you can't make any assumption about the state of the Once instance, that's what makes this kind of predicate tricky to use.

The race condition would be.

1. call `wasCalled` and it returns false
2. another thread calls `call`
3. you call `call` and assume incorrectly that you are the one that got to call it

So if there's some sort of behavior that needs to know whether or not their `call` is the one that "gets through", then there's no way to know for sure.  Any thread needing this behavior would be introducing a "race condition".

LemonBoy: I'll quote my previous message again

> The predicate returns true if the function was called and that's it, whenever you see it returning false you can't make any assumption about the state of the Once instance, that's what makes this kind of predicate tricky to use.

`wasCalled` is only useful for determining if the function was run before the call returned and as such cannot be used to implement the `if (!wasCalled) call()` pattern (that'd defeat the whole point of using `Once`).

Please come back armed with willingness to read people's answers and a valid use case, if you find one.

andrewrk (changes requested):
In which use case do we actually need to pass parameters? Given that this pattern addresses multiple different callsites initializing shared global state, the existence of parameters would be problematic, since they would have to match at every callsite, and therefore should not be encouraged in the API like this.

LemonBoy:
> In which use case do we actually need to pass parameters?

Whenever the function requires some input parameters. As written in the commit body you can't really trust users to protect themselves from data-races and this mechanism is simple enough to save many feet.
Allowing arguments let's one implement this nice pattern where some extra variable is initialized, without requiring us to bloat `Once` and store a return value inline:
```zig
// Some initialize-once-read-many data.
var some_global_var: T = undefined;
fn foo() void {
    some_long_init_done_once.call(.{&some_global_var});
    // It's now possible to use some_global_var
}

**Comment by andrewrk** [comment]
I'm questioning the existence of a real use case - is there a place this would actually be used?

**Comment by LemonBoy** [comment]
I would never waste my time with useless patches, adding useless functionality.
One possible use case is #8948, in some unreleased code I'm using this functionality to call an initialization function only when needed and where I need to pass some parameters due to the lack of closures.

**Comment by andrewrk** [comment]
I'm not intending to accuse you of anything here, but I do want to participate in the API design of this standard library abstraction that will probably land in 1.0. May I see the example use case? For me it's hard to think in generalizations without looking at specific examples.

**Comment by marler8997** [comment]
I'm probably missing some context, but based just on what you've presented here, it seems like there might be better solutions for these use cases.

In the first use case, I'm not seeing the need for std.once.  At first glance I would say just call `compute` in your `init` function then you have no need for `std.once`.  But you said on IRC that the caller may not need to calculate the data, so what about this variation?

```zig
const std = @import("std");

pub fn main() !void {
    var x = T.init(42, -69);
    const table = x.compute();
    table.do_cool_thing();
    table.do_another_cool_thing();
}

const T = struct {
    param_1: i32 = undefined,
    param_2: i32 = undefined,

    fn init(param_1: i32, param_2: i32) T {
        return T {
            .param_1 = param_1,
            .param_2 = param_2,
        };
    }
    fn compute(x: *T) HugeTable {
        std.debug.print("Filling huge_table and using {} and {}!\n", .{ x.param_1, x.param_2 });
        return HugeTable { };
    }
};

const HugeTable = struct {
    huge_table: [100_000]f32 = undefined,

    fn do_cool_thing(x: HugeTable) void {
        std.debug.print("Using huge_table!\n", .{});
    }

    fn do_another_cool_thing(x: HugeTable) void {
        std.debug.print("Using huge_table again!\n", .{});
    }
};
```

The second use case looks particularly nefarious to me.  You have a "global function" named `openDb`, which takes a `Config` parameter and verifies that the options in the `Config` parameter are supported by the current version of the database.  Since you are performing this verification with `std.once`, the verification will only happen once even if `cfg` changes.  If it were me I would just expose this functionality as 2 separate functions, one that verifies a config and one that opens the DB.  Creating an api that looks like it accepts any instance of `cfg` but only actually works correctly on the first call seems unnecessarily misleading.

marler8997:
There is one use case I thought of for this, that is a potential mechanism for `WSAStartup`.  We could use `std.once` in the `std` library to call `WSAStartup` with the default version of `2.2`, but then also expose this `std.once` call to the application and allow them to call it themselves with a different version.  This would cause the version in `std` to act as a "default" value which the application can override.

However, @LemonBoy points out that we don't need to support multiple versions of WSA, and I don't have any real use case to counter with, so for now this use case isn't actually real.  And, there are also other ways of implementing this ability to override the version with current mechanisms, such as defining the version in the `root` module for example.

marler8997:
I discussed a bit with @LemonBoy on IRC and he stressed the fact that 1) the methods on `T` were called from multiple threads and 2) this "huge table" is lazy and may or may not need to be calculated.

Given these 2 points, my example that pulls `HugeTable` out into its own type doesn't solve the problem because now this huge table data would need to be protected inside some sort of critical section, basically, re-implementing what `std.once` already implements for us.

Given this, I can't actually think of a good alternative solution to LemonBoy's example and seems like a valid use case.

marler8997:
I had another idea for a variation on this solution.  Instead of supporting args passed to `call`, what if instead we added support for passing args to the construction of `Once`?  This solution would ensure that all callers of `call` are using the same initialization data, but also allows the scope of the data to match the scope of `Once`.  Unless someone has an objection, I might create another PR with this variation so we can practically compare it.

LemonBoy:
Storing the arguments data inline is a bad idea, you're bloating the structure and making their lifetime hard to track.
I really regret opening this PR.

marler8997:
> you're bloating the structure and making their lifetime hard to track

Ah that's true, good point.

PR #9022, opened 2021-06-07
fix tool/update_clang_options.zig for ArrayHashMap's Separated Key/Va…

redj:
…lue Storage breaking change.

PR #6631, opened 2020-10-09
zig cc / c++: add language arguments support

redj:
When using `zig cc` and `zig c++` allow passing -x / --language to clang to specify the programming language of a source file.

split from #4928
following #6629 and #6630 

note: git should flag some conflicts - it did for me - when applying this PR on top of the previous one simply because the added lines are at the same location. feel free to have me rebase after each PR is accepted if that's convenient.

PR #9029, opened 2021-06-07
zig cc: zig c++: add support for -x and --language arguments.

redj:
When using `zig cc` and `zig c++` allow passing `-x` / `--language` to clang to specify the programming language of a source file.

This is the long overdue re-opening of #6631. Sorry for the delay.
This required #9022 in order to run the update_clang_options tool.


src/main.zig
```diff
@@ -3329,6 +3346,7 @@ pub const ClangArgIterator = struct {
         red_zone,
         no_red_zone,
         strip,
+        treat_as_language,
```
LemonBoy, line 3349:
```suggestion
        force_language,
```

src/main.zig
```diff
@@ -1250,6 +1256,17 @@ fn buildOutputType(
...
                     .nostdlibinc => want_native_include_dirs = false,
                     .strip => strip = true,
+                    .treat_as_language => {
+                        if (mem.eql(u8, it.only_arg, "c") or mem.eql(u8, it.only_arg, "c-header")) {
+                            treat_as_file_ext = .c;
+                        } else if (mem.eql(u8, it.only_arg, "c++")) {
+                            treat_as_file_ext = .cpp;
+                        } else {
+                            treat_as_file_ext = .unknown;
+                            fatal("unsupported value '{s}' for -x or --language option", .{it.only_arg});
```
LemonBoy, line 1266: Not true, there are many [other](https://github.com/llvm/llvm-project/blob/main/clang/include/clang/Driver/Types.def) valid cases that are not handled.

src/main.zig
```diff
@@ -612,6 +612,7 @@ fn buildOutputType(
     var subsystem: ?std.Target.SubSystem = null;
     var major_subsystem_version: ?u32 = null;
     var minor_subsystem_version: ?u32 = null;
+    var treat_as_file_ext: Compilation.FileExt = .unknown;
```
LemonBoy, line 615: This is an optional switch that overrides the autodetection mechanism, using an optional type is better than initializing this variable in a known-invalid state.

```suggestion
    var force_language: ?Compilation.FileExt = null;
```

src/main.zig
```diff
@@ -1125,7 +1126,12 @@ fn buildOutputType(
                         try clang_argv.appendSlice(it.other_args);
                     },
                     .positional => {
-                        const file_ext = Compilation.classifyFileExt(mem.spanZ(it.only_arg));
+                        var file_ext: Compilation.FileExt = undefined;
+                        if (treat_as_file_ext == .unknown) {
+                            file_ext = Compilation.classifyFileExt(mem.spanZ(it.only_arg));
+                        } else {
+                            file_ext = treat_as_file_ext;
```
LemonBoy, line 1133: Is this correct in the presence of both object and source inputs? I think you only want the option to influence the latter.

src/main.zig
```diff
@@ -1125,7 +1126,12 @@ fn buildOutputType(
                         try clang_argv.appendSlice(it.other_args);
                     },
                     .positional => {
-                        const file_ext = Compilation.classifyFileExt(mem.spanZ(it.only_arg));
+                        var file_ext: Compilation.FileExt = undefined;
```
LemonBoy, line 1129: Style nit, don't initialize variables in invalid state. You can easily use a block or, if possible, the `if` expression to initialize it:
```zig
const file_ext = if (treat_as_file_ext == .unknown) X else Y;
// or
const file_ext = blk: { if (treat_as_file_ext == .unknown) break :blk X else break :blk Y };
```

src/main.zig
redj: @LemonBoy, according to https://clang.llvm.org/docs/ClangCommandLineReference.html :
```
-x<language>, --language <arg>, --language=<arg>¶
Treat subsequent input files as having type <language>
```

... so I would think you're not supposed to specify any other kind of inputs after a -x / --language option?

src/main.zig
LemonBoy: > are you suggesting I change the error message or support all the valid options?

The latter, adding more valid extensions is also good for plain `zig` as, at the moment, it will outright reject any objective-c file.

src/main.zig
LemonBoy: That was a question for you. I've just checked and clang is blindly treating any input (even binary ones) as source code, this logic is correct.

Vexu:
What is the state of this pr? All of the reviews seem to be addressed and there have been no changes in two months yet it is still marked as a draft.

andrewrk:

Issue #7396, opened 2020-12-11
Design flaw: Closure over comptime var


SpexGuy:
This issue is heavily inspired by #5895 and #5578.

As it stands right now, the compiler allows you to create a closure over mutable comptime state.  This allows some unique abilities, like [building a borrow checker](https://github.com/DutchGhost/zorrow) and lazily creating a global list based on what functions or types get compiled.

But it also causes a lot of problems:
- it wreaks havoc with the comptime memoization system.
- it makes comptime execution order observable.
- it prevents incremental compilation.

The aim of this issue is to provide a solid bedrock for how comptime var should behave, without introducing new features.  There may be extensions that add new features in future proposals.

After [discussing in the design meeting](https://docs.google.com/document/d/1GFQxcoXgzmOFThjFLpjkVASopa33el2f5esvk_OZlCo/edit#heading=h.dibk56wlkx1f), this is how we believe comptime var (and comptime mutable memory in general) should work:
- closure over comptime var is a compile error.
- closure over const pointer to mutable memory (e.g. `comptime const x: *u32 = ...`) is allowed
- when mutable pointers are written at comptime, the compiler ensures that the underlying memory is still mutable
- the memory underlying comptime vars is marked immutable when the vars go out of scope.
- if the compiler attempts to generate runtime code which dereferences a mutable pointer to a comptime var, that’s also a compile error.
- for the purposes of memoization, any values which are closed over are treated like additional parameters.

---

This means that the example given in #5578 can be implemented like this:
```zig
const GiveCtState = struct {
    times: usize,
};
fn Give(comptime T: type, comptime state: *GiveCtState) type {
    return struct {
        val: T,
        pub fn init(_val: T) @This() {
            return Self{ .val = _val };
        }

        pub fn give(self: *@This()) T {
            comptime {
                if (state.times == 0)
                    @compileError("You can only 'give' 2 times");

                state.times -= 1;
            }
            return self.val;
        }
    };
}

test "Test 1" {
    // create comptime state in this scope
    comptime var track_a = GiveTracker{ .times = 2 };
    // create a type that closes over that state
    var a = Give(usize, &track_a).init(1);
    _ = a.give();
    _ = a.give(); // not memoized because it closes over &track_a
    _ = a.give(); // <-- Should Fail

    comptime var track_b = GiveTracker{ .times = 2 };
    var b = Give(usize, &track_b).init(1); // not memoized because it accepts &track_b
    _ = b.give();
    _ = b.give();
[...]

Issue #1470, opened 2018-09-04
compile error for modification of comptime data inside runtime block


andrewrk:
Inspired by #906 

```zig
const std = @import("std");
const builtin = @import("builtin");
const debug = std.debug;

fn isSliceType(comptime T: type) bool {
    if (@typeInfo(T) == .Pointer) {
        if (@typeInfo(T).Pointer.size == .Slice) {
            return true;
        }
    }
    return false;
}

fn countSliceMembers(inp: anytype) usize {
    comptime var count = @as(usize, 0);
    const type_info = @typeInfo(@TypeOf(inp));

    inline for (type_info.Struct.fields) |field_info| {
        const FieldType = field_info.field_type;
        if (isSliceType(FieldType)) count += 1;
    }
    return count;
}

test "incorrectly caching... something" {
    const String = []const u8;
    const Number = u32;

    const S = struct {
        string: String,
        number: Number,
    };

    debug.assert(isSliceType(String) == true);
    debug.assert(isSliceType(Number) == false);

    var s = S{
        .string = "",
        .number = 10,
    };
    debug.assert(countSliceMembers(s) == 1);
}
```

`count += 1`  is modification of comptime data inside of a runtime block. It can have surprising results, as demonstrated by this test case. With this proposal, Zig would detect that `count` was declared outside the runtime block scope, and therefore this is illegal modification of comptime data.

SpexGuy:
The specific decision here is that comptime vars should become immutable when inside a runtime branch.  So for example:
```zig
fn foo() {
  comptime var x = 0; // x is mutable

  if (runtime_value) {
    comptime var y = 0;
    // y is mutable, x is immutable
  } else {
    // x is immutable
  }
  
  for (some_slice) |item, i| {
    // runtime branch, x is immutable
  }
  
  while (some_condition) {
    // runtime branch, x is immutable
  }
  
  switch (runtime_value) {
    else => {}, // runtime branch, x is immutable
  }
  
  if (comptime_value) {
    // comptime branch, x is mutable
  } else {
    // comptime branch, x is mutable
  }
  
  inline for (some_array_or_tuple) |item, i| {
    // not a runtime branch, x is mutable
  }
  
  inline while (some_comptime_condition) {
    //  not a runtime branch, x is mutable
  }
  
  switch (comptime_value) {
    else => {}, // not a runtime branch, x is mutable
  }
}
```

This is similar to how comptime vars become immutable when they go out of scope.

[...]

PR #9030, opened 2021-06-07
Stage2: implement comptime variables

Vexu:
Not sure how well this will work with arrays and structs but the invalid store logic should be good.

Inline loops do not work yet due to some other bug in analysis.

Implements #1470 but not #7396.

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -2357,7 +2357,7 @@ fn varDecl(
             return &sub_scope.base;
         },
         .keyword_var => {
-            const is_comptime = var_decl.comptime_token != null;
+            const is_comptime = var_decl.comptime_token != null or gz.force_comptime;
             var resolve_inferred_alloc: Zir.Inst.Ref = .none;
             const var_data: struct {
                 result_loc: ResultLoc,
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -1139,6 +1139,13 @@ pub const Scope = struct {
         instructions: ArrayListUnmanaged(*ir.Inst),
         label: ?*Label = null,
         inlining: ?*Inlining,
+        /// If runtime_index is not 0 then one of these is guaranteed to be non null.
+        runtime_cond: ?LazySrcLoc = null,
+        runtime_loop: ?LazySrcLoc = null,
+        /// Non zero if a non-inline loop or a runtime conditional have been encountered.
+        /// Stores to to comptime variables are only allowed when var.runtime_index <= runtime_index.
+        runtime_index: u32 = 0,
+
         is_comptime: bool,
 
         /// This `Block` maps a block ZIR instruction to the corresponding
@@ -1182,6 +1189,9 @@ pub const Scope = struct {
                 .label = null,
                 .inlining = parent.inlining,
                 .is_comptime = parent.is_comptime,
+                .runtime_cond = parent.runtime_cond,
+                .runtime_loop = parent.runtime_loop,
+                .runtime_index = parent.runtime_index,
             };
         }
 
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -509,7 +509,7 @@ pub fn analyzeBody(
         };
         if (air_inst.ty.isNoReturn())
             return always_noreturn;
-        try map.putNoClobber(sema.gpa, inst, air_inst);
+        try map.put(sema.gpa, inst, air_inst);
     }
 }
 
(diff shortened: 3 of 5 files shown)


andrewrk:
I just played with this a little bit, and it looks great! It's nice to see this being massively simpler with the SSA in tree form than how it is in stage1.

Here's another set of passing test cases you can add:

```zig
comptime {
    var x: i32 = 1;
    x += 1;
    if (x != 2) unreachable;
}
```

```zig
comptime {
    var x: i32 = 1;
    x += 1;
    if (x != 1) unreachable;
}
```

The second one correctly gives:

```
test.zig:4:17: error: unable to resolve comptime value
    if (x != 1) unreachable;
                ^
```

Regarding the todo about source location, I'll have a look at that and see if I can help.

andrewrk:
Here's the solution to the source location issue:

```diff
diff --git a/src/Sema.zig b/src/Sema.zig
index 0dd9ea47e..ea5a0e377 100644
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -4607,14 +4607,14 @@ fn zirArithmetic(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) InnerEr
 
     const tag_override = block.sema.code.instructions.items(.tag)[inst];
     const inst_data = sema.code.instructions.items(.data)[inst].pl_node;
-    const src: LazySrcLoc = .{ .node_offset_bin_op = inst_data.src_node };
+    sema.src = .{ .node_offset_bin_op = inst_data.src_node };
     const lhs_src: LazySrcLoc = .{ .node_offset_bin_lhs = inst_data.src_node };
     const rhs_src: LazySrcLoc = .{ .node_offset_bin_rhs = inst_data.src_node };
     const extra = sema.code.extraData(Zir.Inst.Bin, inst_data.payload_index).data;
     const lhs = try sema.resolveInst(extra.lhs);
     const rhs = try sema.resolveInst(extra.rhs);
 
-    return sema.analyzeArithmetic(block, tag_override, lhs, rhs, src, lhs_src, rhs_src);
+    return sema.analyzeArithmetic(block, tag_override, lhs, rhs, sema.src, lhs_src, rhs_src);
 }
 
 fn zirOverflowArithmetic(
```

Generally, the idea here is to set `sema.src` for any instruction which has a source location, allowing AstGen to skip supplying the source location if it knows it would be redundant with the previous instruction.

Vexu:
> Here's the solution to the source location issue:

Thanks, added.

Also to clarify on the inline loop issue, I checked that the loads and stores are happening correctly but for some reason the condition is only evaluated once. You'll probably be able to find the issue much quicker than me since you wrote all of the inlining logic.
```zig
comptime {
    var i: u64 = 0;
    inline while (i < 5) : (i+=5) {} // :3:12: error: evaluation exceeded 1000 backwards branches
}
```

andrewrk:
The problem was `repeat_inline` tried to set `i = 0;` but then the continue expression (`i += 1`) was making it skip over the condition. Solution:

```diff
diff --git a/src/Sema.zig b/src/Sema.zig
index 0dd9ea47e..6145b8c13 100644
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -147,7 +147,7 @@ pub fn analyzeBody(
     // directly jump to the next one, rather than detouring through the loop
     // continue expression. Related: https://github.com/ziglang/zig/issues/8220
     var i: usize = 0;
-    while (true) : (i += 1) {
+    while (true) {
         const inst = body[i];
         const air_inst = switch (tags[inst]) {
             // zig fmt: off
@@ -394,78 +394,97 @@ pub fn analyzeBody(
             // putting them into the map.
             .breakpoint => {
                 try sema.zirBreakpoint(block, inst);
+                i += 1;
                 continue;
             },
             .fence => {
                 try sema.zirFence(block, inst);
+                i += 1;
                 continue;
             },
             .dbg_stmt => {
                 try sema.zirDbgStmt(block, inst);
+                i += 1;
                 continue;
             },
             .ensure_err_payload_void => {
                 try sema.zirEnsureErrPayloadVoid(block, inst);
+                i += 1;
                 continue;
             },
             .ensure_result_non_error => {
                 try sema.zirEnsureResultNonError(block, inst);
+                i += 1;
                 continue;
             },
             .ensure_result_used => {
                 try sema.zirEnsureResultUsed(block, inst);
+                i += 1;
                 continue;
             },
             .set_eval_branch_quota => {
                 try sema.zirSetEvalBranchQuota(block, inst);
+                i += 1;
                 continue;
             },
             .store => {
                 try sema.zirStore(block, inst);
+                i += 1;
                 continue;
             },
             .store_node => {
                 try sema.zirStoreNode(block, inst);
+                i += 1;
                 continue;
             },
             .store_to_block_ptr => {
                 try sema.zirStoreToBlockPtr(block, inst);
+                i += 1;
                 continue;
             },
             .store_to_inferred_ptr => {
                 try sema.zirStoreToInferredPtr(block, inst);
+                i += 1;
                 continue;
             },
             .resolve_inferred_alloc => {
                 try sema.zirResolveInferredAlloc(block, inst);
+                i += 1;
                 continue;
             },
             .validate_struct_init_ptr => {
                 try sema.zirValidateStructInitPtr(block, inst);
+                i += 1;
                 continue;
             },
             .validate_array_init_ptr => {
                 try sema.zirValidateArrayInitPtr(block, inst);
+                i += 1;
                 continue;
             },
             .@"export" => {
                 try sema.zirExport(block, inst);
+                i += 1;
                 continue;
             },
             .set_align_stack => {
                 try sema.zirSetAlignStack(block, inst);
+                i += 1;
                 continue;
             },
             .set_cold => {
                 try sema.zirSetAlignStack(block, inst);
+                i += 1;
                 continue;
             },
             .set_float_mode => {
                 try sema.zirSetFloatMode(block, inst);
+                i += 1;
                 continue;
             },
             .set_runtime_safety => {
                 try sema.zirSetRuntimeSafety(block, inst);
+                i += 1;
                 continue;
             },
 
@@ -510,6 +529,7 @@ pub fn analyzeBody(
         if (air_inst.ty.isNoReturn())
             return always_noreturn;
         try map.put(sema.gpa, inst, air_inst);
+        i += 1;
     }
 }
 
```

Confirmed this fixes the test case you posted above, as well as existing stage2 tests.

andrewrk:
The failure is:

```
Expected error not found:
================================================================================
:4:17: error: unable to resolve comptime value
================================================================================
update_index=3 test 'comptime var' failed: WrongCompileErrors
```

The test case is exhibiting correct behavior:

```
pub fn main() void {
    comptime var x: i32 = 1;
    x += 1;
    if (x != 1) unreachable;
}
```

It correctly compiles OK, and then traps when it hits unreachable at runtime.

The test case that is passing is:

```
comptime {
    var x: i32 = 1;
    x += 1;
    if (x != 1) unreachable; // error: unable to resolve comptime value
}

pub fn main() void {}
```

PR #9031, opened 2021-06-07
wasi: always grant fd_readdir right

### LINKED ISSUE #8742: wasi: syscalls relying on fd_readdir right fail with wasmtime >= v0.23
See bytecodealliance/wasmtime#2891 for context. In summary, any function such as `std.fs.Dir.deleteTree` which requires a call to `readdir` syscall will fail when run against wasmtime at least v0.23 since the way the rights are cascaded has fundamentally changed.

This issue is a blocker at bumping supported wasmtime version from v0.22.2, and its resolution will require scanning through relevant wasmtime's implementation of rights management and working out how to handle rights distribution within Zig's libstd.

kubkon:
Since v0.23 release of Wasmtime, if we want to iterate a directory Y then directory Y needed to have been granted `fd_readdir` right. However, it is now also required for directory X to carry `fd_readdir` right, and so on, up-chain all the way until we reach the preopen (which possesses all rights by default).

This caused problems for us since our libstd implementation is more fine-grained and allowed for parent dirs not to carry the right while allow for iterating on its children. My proposal here is to always grant `fd_readdir` right as part of `std.fs.Dir.OpenDirOptions.access_sub_paths`. This seems to be the approach taken by Rust also, plus we should be justified to take this approach since WASI is experimental and snapshot1 will be discontinued eventually and replaced with a new approach to access management that will require a complete rewrite of our libstd anyhow.

PRs for context:
- https://github.com/bytecodealliance/wasmtime/issues/2891
- https://github.com/WebAssembly/wasi-filesystem/issues/31

Closes #8742

### DIFF
--- a/ci/azure/linux_script
+++ b/ci/azure/linux_script
@@ -25,10 +25,8 @@ wget -nv "https://ziglang.org/deps/$QEMUBASE.tar.xz"
 tar xf "$QEMUBASE.tar.xz"
 export PATH="$(pwd)/$QEMUBASE/bin:$PATH"
 
-# Bump to v0.23 once this issue is resolved:
-# https://github.com/ziglang/zig/issues/8742
-WASMTIME="wasmtime-v0.22.1-x86_64-linux"
-wget -nv "https://github.com/bytecodealliance/wasmtime/releases/download/v0.22.1/$WASMTIME.tar.xz"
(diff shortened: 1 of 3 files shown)


jedisct1:
Looks like the right fix to apply!

The capability-based stdlib is going to be another story :(

PR #9035, opened 2021-06-07
Add std.os.dup()

vktec:


kubkon (changes requested):
Thanks for the PR! For starters, would you mind adding a testcase verifying that we handle all OSes properly? Drop the testcase in `std/os/test.zig`.

PR #9036, opened 2021-06-07
Add std.io.Reader.readUntilDelimiter()

vktec:


lib/std/io/reader.zig
```diff
@@ -143,6 +143,23 @@ pub fn Reader(
...
+        /// large enough to hold the entire contents, `error.StreamTooLong` is returned.
+        /// Returns a slice of the stream data, with ptr equal to `buf.ptr`. The
+        /// delimiter byte is not included in the returned slice.
+        pub fn readUntilDelimiter(self: Self, buf: []u8, delimiter: u8) ![]u8 {
+            var index: usize = 0;
+            while (true) {
+                const byte = try self.readByte();
+
+                if (byte == delimiter) return buf[0..index];
+                if (index >= buf.len) return error.StreamTooLong;
```
daurnimator, line 156: This can result in losing `byte`. Perhaps move this check before the `readByte` call?

mattbork: It matches the behavior of the existing `readUntilDelimiter*` functions. The problem is that you don't want a buffer of exactly the right size to fail, but you can't know whether it will or not until one byte past the end is read.

daurnimator: > It matches the behavior of the existing `readUntilDelimiter*` functions.

That was the reason I introduced `readUntilDelimiterArrayList`: it doesn't have this problem. I thought we were going to deprecate all the other ones....

LemonBoy: > That was the reason I introduced readUntilDelimiterArrayList: it doesn't have this problem.

`readUntilDelimiterArrayList` is performing the same operations in the same order, the last byte is always lost.

LemonBoy:
Add some tests please.

lib/std/io/reader.zig
daurnimator: > `readUntilDelimiterArrayList` is performing the same operations in the same order, the last byte is always lost.

ah crap, it has a bug then!

Vexu: Sounds like both of these bugs could be fixed by including the delimiter in the output?

Vexu: #9594

Issue #8396, opened 2021-03-30
Add DateTime in std.time


InterplanetaryEngineer:
This just came up on Discord: it could be very useful to have something like:
```zig
const DateTime = struct {
    sec: u8,
    min: u8,
    hour: u8,
    year_day: u16,
    year: i16,

    pub fn monthDay(date: Self) u8;
    pub fn month(date: Self) u8;
    pub fn weekDay(date: Self) u8;
    pub fn week(date: Self) u8;
    pub fn isLeapYear(date: Self) bool;

    pub fn fromEpoch(epoch: i64, utc_offset_seconds: i32, known_leap_seconds: []const i64) Self;
    pub fn toEpoch(date: Self, utc_offset_seconds: i32, known_leap_seconds: []const i64) i64;
};
```
Maybe also with a more elegant way to format it than passing the fields seperately to `fmt.format`.

MasterQ32:
I think having a proper well-made date/time library is important, but it's not a good idea to include this in `std`. A *recommended* date/time library would be better.

But yes, i agree that having a date-time api would be a good idea!

InterplanetaryEngineer:
This proposal is fairly similar to C's `tm` from `time.h`. Are there any guidelines as to what should be in zig's std library?

Mouvedia:
```
    sec: u8, // [0, 60]
    min: u8, // [0, 59]
    hour: u8, // [0, 23] 
    year_day: u16, // [0, 365]
```
should be
```
    sec: u6, // [1, 59]
    min: u6, // [1, 59]
    hour: u5, // [1, 23]
    day: u5, // [1, 31]
    month: u4, // [1, 12]
    year_day: u9, // [1, 366]
```

RogierBrussee:
Seconds are far, far too little resolution when atomic clocks are used as a reference (e.g. in GPS).

In glibc struct timespec allows to specify nanosecond, and was the result of a long painful learning experience.
Why not  just add convenience accessor functions to specify days and months, hours, minutes and seconds, to a structure that is binary compatible with struct timespec?

InterplanetaryEngineer:
@RogierBrussee I meant IAT not because of atomic precision, but strict monotonicity. DateTime would be used for human-readable displaying, is there any use-case for nanoseconds there? You'd still do precision calculations with timestamps, of course.
> Why not just add convenience accessor functions to specify days and months, hours, minutes and seconds [...]?

That is exactly what this is.
> ... binary compatible with struct timespec

[...]

PR #9040, opened 2021-06-08
add functions to decode an epoch timestamp

marler8997:
The code added here is alternative to the libc `gmtime` function.  This function takes a unix epoch timestamp and decodes it into things like the year/month/day/time/etc.  I looked at various libc implementations to see how it was implemented and this appears to be correct (the tests also match what I get on epochconverter.com).

I split the single `gmtime` function into multiple functions based on the order in which the date information is decoded so that applications can limit their calculations to what they need rather than calculating everything in a single function.  The data structures are based on the order that pieces of the timestamp are calculated.

### DIFF
--- a/lib/std/time/epoch.zig
+++ b/lib/std/time/epoch.zig
@@ -124,7 +124,7 @@ pub const YearAndDay = struct {
         const leap_kind: YearLeapKind = if (isLeapYear(self.year)) .leap else .not_leap;
         while (true) {
             const days_in_month = getDaysInMonth(leap_kind, month);
-            if (days_left <= days_in_month)
+            if (days_left < days_in_month)
                 break;
             days_left -= days_in_month;
             month = @intToEnum(Month, @enumToInt(month) + 1);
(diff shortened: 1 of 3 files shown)


lib/std/time/epoch.zig
```diff
@@ -40,3 +43,179 @@ pub const morphos = amiga;
...
+        .oct => 31,
+        .nov => 30,
+        .dec => 31,
+    };
+}
+
+pub const YearAndDay = struct {
+    year: Year,
+    /// The number of days into the year (0 to 365)
+    day: u9,
```
Vexu, line 119: ISO 8601 says that day of the year is in range `[1,365 (366)]`, maybe this should be too?

lib/std/time/epoch.zig
```diff
@@ -40,3 +43,179 @@ pub const morphos = amiga;
...
+}
+
+pub fn getDaysInYear(year: Year) u9 {
+    return if (isLeapYear(year)) 366 else 365;
+}
+
+pub const YearLeapKind = enum(u1) { not_leap, leap };
+
+pub const Month = enum(u4) {
+    jan,
```
LemonBoy, line 75:
```suggestion
    jan = 1,
```

And so forth. You can avoid the +1 dance and the `numeric` helper.

LemonBoy:
This is possibly related to #8396 as it implements random chunks of a `Date` structure.

lib/std/time/epoch.zig
marler8997: sure that should work, done

lib/std/time/epoch.zig
marler8997: Not sure.  I actually can't think of a use case for wanting the day of the year. For the calculation we need it to be 0 based, so if we make it 1 based, we'll just be adding 1 then undoing that operation in `calculateMonthDay`. An alternative is also to create a method that will return this day value with 1 added. Not sure what the best route is here.

Vexu (approved):

PR #9041, opened 2021-06-08
translate-c: better support for static local variables

### LINKED ISSUE #8208: translate-c static function variable shadowing other functions parameters
**Zig version: 0.8.0-dev.1431+c760532be**

**Problem:  static function variable is shadowing other functions arguments with same name.** 

**Error message:** 
```
$ zig build -Dtarget=native-native-gnu
.\zig-cache\o\ea6804b44f6b8f259469a108221aed6a\cimport.zig:3:5: error: redefinition of 'size'
    var size = arg_size;
    ^
.\zig-cache\o\ea6804b44f6b8f259469a108221aed6a\cimport.zig:7:1: note: previous definition is here
var size: c_int = 0;
```

**Repro example**
https://gist.github.com/lacebal/a77a4cb24e0d738e1a2c4a977da0cf7f

**Same error was observed when const static var shadows functions params when integrating nuklear C library: https://github.com/Immediate-Mode-UI/Nuklear**

staticshadow.h
```
extern int paramsize(int);
extern int staticshadowing(void);

#ifdef STATICSHADOW_IMPL
extern int paramsize(int size) {
    size++;
    return size;
}
extern int staticshadowing(void) {
    static int size = 0;
    size++;
    return size;
}
#endif
```

main.zig
```
const c = @cImport({
    // @cDefine("NK_INCLUDE_FIXED_TYPES", {});
    // @cDefine("NK_INCLUDE_STANDARD_IO", {});
    // @cDefine("NK_INCLUDE_DEFAULT_ALLOCATOR", {});
    // @cDefine("NK_INCLUDE_VERTEX_BUFFER_OUTPUT", {});
    // // @cDefine("NK_INCLUDE_FONT_BAKING", {});
    // // @cDefine("NK_INCLUDE_DEFAULT_FONT", {});
    // @cDefine("NK_IMPLEMENTATION", {});
    // @cInclude("nuklear.h");

    @cDefine("STATICSHADOW_IMPL", {});
    @cInclude("staticshadow.h");
});
const std = @import("std");

pub fn main() void {
    // var cmds: c.struct_nk_buffer = c.struct_nk_buffer {
    //     .marker = [_]c.struct_nk_buffer_marker{ 
    //         c.struct_nk_buffer_marker{ .active = undefined, .offset = undefined } 
    //     } ** 2,
    //     .pool = undefined,
    //     .type = undefined,
    //     .memory = c.struct_nk_memory{ .ptr = undefined, .size = undefined },
    //     .grow_factor = undefined,
    //     .allocated = undefined,
    //     .needed = undefined,
    //     .calls = undefined,
    //     .size = undefined,
    // };
    // c.nk_buffer_init_default(&cmds);
    std.log.info("Size: {} {}", .{ c.paramsize(10), c.staticshadowing() });
}
```

build.zig
```
const x86 = @import("std").Target.x86;

    // Standard target options allows the person running `zig build` to choose
    // what target to build for. Here we do not override the defaults, which
    // means any target is allowed, and the default is native. Other options
    // for restricting supported target set are available.
    var target = b.standardTargetOptions(.{});
    
    // Standard release options allow the person running `zig build` to select
    // between Debug, ReleaseSafe, ReleaseFast, and ReleaseSmall.

    const exe = b.addExecutable("app-zig-bug", "src/main.zig");
    exe.setBuildMode(mode);
    exe.setTarget(target);
    // nuklear
    exe.addSystemIncludeDir("vendor/nuklear/");
    // src
    exe.addSystemIncludeDir("src/");
    
    // C and win32
    exe.linkLibC();
    exe.install();

    const run_cmd = exe.run();
    if (b.args) |args| {
        run_cmd.addArgs(args);
    }

}
```

cimport.zig extract
```
pub usingnamespace @import("std").c.builtins;
pub fn paramsize(arg_size: c_int) callconv(.C) c_int {
    var size = arg_size;
    size += 1;
    return size;
}
var size: c_int = 0;
pub fn staticshadowing() callconv(.C) c_int {
    size += 1;
    return size;
}
...
```

ehaas:
Don't move static local variables into the top-level scope since this
can cause name clashes if subsequently-defined variables or parameters
in different scopes share the name.

Instead, use a variable within a struct so that the variable's lexical
scope does not change. This solution was suggested by @LemonBoy

Note that a similar name-shadowing problem exists with `extern` variables
declared within block scope, but a different solution will be needed since
they do need to be moved to the top-level scope and we can't rename them.

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -72,6 +72,11 @@ const Scope = struct {
         /// so that the return expression can be cast, if necessary
         return_type: ?clang.QualType = null,
 
+        /// C static local variables are wrapped in a block-local struct. The struct
+        /// is named after the (mangled) variable name, the Zig variable within the
+        /// struct itself is given this name.
+        const StaticInnerName = "static";
+
         fn init(c: *Context, parent: *Scope, labeled: bool) !Block {
             var blk = Block{
                 .base = .{
@@ -1738,15 +1743,13 @@ fn transDeclStmtOne(
             const name = try c.str(@ptrCast(*const clang.NamedDecl, var_decl).getName_bytes_begin());
             const mangled_name = try block_scope.makeMangledName(c, name);
 
-            switch (var_decl.getStorageClass()) {
-                .Extern, .Static => {
-                    // This is actually a global variable, put it in the global scope and reference it.
-                    // `_ = mangled_name;`
-                    return visitVarDecl(c, var_decl, mangled_name);
-                },
-                else => {},
+            if (var_decl.getStorageClass() == .Extern) {
+                // This is actually a global variable, put it in the global scope and reference it.
+                // `_ = mangled_name;`
+                return visitVarDecl(c, var_decl, mangled_name);
             }
 
+            const is_static_local = var_decl.isStaticLocal();
             const is_const = qual_type.isConstQualified();
 
             const loc = decl.getLocation();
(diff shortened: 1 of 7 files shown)


LemonBoy:
You can emulate that behaviour by introducing an inner container like this:
```zig
fn foo() void {
    const Inner = struct {
        var similar_to_static: i32 = 42;
    };
    // Every access to `similar_to_static` must be prefixed with `Inner`
}
```

src/translate_c.zig
```diff
@@ -72,6 +72,12 @@ const Scope = struct {
         /// so that the return expression can be cast, if necessary
         return_type: ?clang.QualType = null,
 
+        /// C static local variables are wrapped in a block-local struct. The struct
+        /// is named after the (mangled) variable name, the Zig variable within the
+        /// struct itself is given this name. We use a dot in the name to ensure
+        /// it does not clash with any C identifiers
+        const StaticInnerName = ".inner";
```
Vexu, line 79: Could we use `static` instead? I think `foo.static` makes it clearer why the code looks the way it is and  using a C keyword also takes care of the shadowing issue.

ehaas: ah nice, good idea

Vexu:
I was able to reproduce the issue when using `extern` in the example in the issue, shouldn't be too hard to make this also happen for externs though.

ehaas:
I think it needs a slightly different solution since the struct name will need to be mangled and the extern variable name not changed, so I *think* (without investigating it more yet) that we will also need to keep track of the mangled struct name that corresponds to each extern var within a particular scope.

Vexu:
Fair enough, getting even just half of the issue fixed is good. Probably quite unlikely to run in to a situation like with an extern shadowing.

ehaas:
Oops, I think we force-pushed at the same time - did you only push a conflict resolution or were there other changes?

Vexu:
Just conflict resolution and removing the fixed part of the message since the issue now also applies to externs, I'll just reopen it after merging.

ehaas:
I seem to have broken something with that push, I'll take a look at it

PR #9043, opened 2021-06-08
native libc detection: respect spaces in CC env var

### LINKED ISSUE #8960: when CC environment variable is set to zig and zig cannot provide libc, error: unable to create compilation: UnableToSpawnCCompiler
To reproduce the issue, pick a system where Zig cannot provide libc, such as FreeBSD (until #2876 is solved) and override the CC environment variable to be Zig.

When trying to link libc, you will get some output like this:

```
[ 12%] Building C object CMakeFiles/z.dir/compress.o
Zig attempted to find the path to native system libc headers by executing this command:
/home/andrewrk/zig-bootstrap/out/host/bin/zig cc -fno-sanitize=all -target x86_64-native -mcpu=x86_64_v2 -E -Wp,-v -xc /dev/null
error: unable to create compilation: UnableToSpawnCCompiler
Zig attempted to find the path to native system libc headers by executing this command:
/home/andrewrk/zig-bootstrap/out/host/bin/zig cc -fno-sanitize=all -target x86_64-native -mcpu=x86_64_v2 -E -Wp,-v -xc /dev/null
error: unable to create compilation: UnableToSpawnCCompiler
```

However we can make this work, by detecting when the CC environment variable is set to Zig inside libc_installation.zig, and invoking `cc` rather than the zig compiler from CC, in order to find the native libc installation paths.

https://github.com/ziglang/zig/blob/f6932472950e7dbb451d4cfef8e5f4a1cc506ac0/src/libc_installation.zig#L241

https://github.com/ziglang/zig/blob/f6932472950e7dbb451d4cfef8e5f4a1cc506ac0/src/libc_installation.zig#L556

Another bug, which needs to be fixed in order to solve this one, is that Zig does not pick part the CC parameters and separate them by spaces, as is widely supported by systems that recognize the CC environment variable.

Until this is solved, the workaround for this problem is to execute `zig libc` before overriding the environment variable and save the output somewhere. Then use it in the CLI when overriding CC, like this:

```
zig libc >~/tmp/libc.txt
export ZIG_LIBC=$HOME/tmp/libc.txt
export CC="zig cc ..."
```

andrewrk:
Zig has detection for when it is accidentally being called recursively
when trying to find the native libc installation. However it was not
working, resulting in a cryptic failure, because zig tried to execute
a command which had spaces in it rather than tokenizing it.

This improves the user experience of `zig cc` for systems that Zig
does not support cross-compiling for.

Closes #8960

### DIFF
--- a/src/libc_installation.zig
+++ b/src/libc_installation.zig
@@ -237,26 +237,38 @@ pub const LibCInstallation = struct {
 
     fn findNativeIncludeDirPosix(self: *LibCInstallation, args: FindNativeOptions) FindError!void {
         const allocator = args.allocator;
+
+        // Detect infinite loops.
+        var env_map = try std.process.getEnvMap(allocator);
+        defer env_map.deinit();
+        const skip_cc_env_var = if (env_map.get(inf_loop_env_key)) |phase| blk: {
+            if (std.mem.eql(u8, phase, "1")) {
+                try env_map.put(inf_loop_env_key, "2");
+                break :blk true;
+            } else {
+                return error.ZigIsTheCCompiler;
+            }
+        } else blk: {
+            try env_map.put(inf_loop_env_key, "1");
+            break :blk false;
+        };
+
         const dev_null = if (is_windows) "nul" else "/dev/null";
-        const cc_exe = std.os.getenvZ("CC") orelse default_cc_exe;
-        const argv = [_][]const u8{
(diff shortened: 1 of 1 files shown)


andrewrk:
cc @lun-4 this solves 1 of the issues from https://github.com/ziglang/zig-bootstrap/pull/60

PR #9048, opened 2021-06-09
Improve error message when `std.fmt.format` is missing arguments

Jarred-Sumner:
**Before** | **After**
-----------|--------
 | image

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -359,7 +359,12 @@ pub fn format(
     }
 
     if (comptime arg_state.hasUnusedArgs()) {
-        @compileError("Unused arguments");
+        const missing_count = arg_state.args_len - @popCount(ArgSetType, arg_state.used_args);
+        switch (missing_count) {
+            0 => unreachable,
+            1 => @compileError("Unused argument in \"" ++ fmt ++ "\""),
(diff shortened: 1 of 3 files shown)


lib/std/fmt.zig
```diff
@@ -359,7 +359,12 @@ pub fn format(
     }
 
     if (comptime arg_state.hasUnusedArgs()) {
-        @compileError("Unused arguments");
+        const missing_count = arg_state.args_len - arg_state.used_args;
```
LemonBoy, line 362: `used_args` is a bitfield...

lib/std/fmt.zig
```diff
@@ -359,7 +359,12 @@ pub fn format(
     }
 
     if (comptime arg_state.hasUnusedArgs()) {
-        @compileError("Unused arguments");
+        const missing_count = arg_state.args_len - @popCount(ArgSetType, arg_state.used_args);
+        switch (missing_count) {
+            0 => unreachable,
+            1 => @compileError("Unused argument in \"" ++ fmt ++ "\""),
+            else => @compileError((comptime comptimePrint("{d}", .{missing_count})) ++ " unused arguments  in \"" ++ fmt ++ "\""),
```
LemonBoy, line 366: Beware of the double space between `arguments` and `in`.

PR #9049, opened 2021-06-09
stage2 codegen: Remove hacks for discontinued SPU II backend

joachimschmidt557:
Now `codegen.zig` only hosts targets which have registers (at least until `codegen.zig` will be refactored and split into separate files for different targets).


kubkon (commented):
Provided no-one is actually working on the SPU II backend anymore, this change is fair for me.

Issue #7070, opened 2020-11-11
Better documentation for await


tecanec:
The documentation for `await` leaves a few questions unanswered, such as:
- What happens when you resume a frame that's awaiting another?
- When and how does a function continue after an await has completed? What is the exact control flow that occurs?
- Are multiple awaits for the same function allowed?

Personally, I've stayed away from async for a long time because these question had left me confused, so I'd say that they should be answered in the docs.

euclidianAce:
I've been having similar questions lately and wrote out this snippet that I think illustrates the behavior in a couple of these questions.

I've also attempted to answer the above questions incase anybody (like myself) was also curious about the behavior of `await`.

```zig
const std = @import("std");

// logger where indentation basically represents the call stack
var indent: usize = 0;
fn log(comptime msg: []const u8) void {
    var local_indent = indent;
    while (local_indent > 0) : (local_indent -= 1)
        std.debug.print("    ", .{});
    std.debug.print("{s}\n", .{msg});
}

fn suspendOnce() void {
    log("suspend..."); // 3
    suspend {}
    log("after suspend, returning; anything now awaiting this frame should execute"); // 6
}

var the_frame: @Frame(suspendOnce) = undefined;
fn foo() void {
    log("call async suspendOnce()"); // 2
    indent += 1;
    the_frame = async suspendOnce();
    indent -= 1;
    log("[foo] awaiting the frame of suspendOnce"); // 4
    indent += 1;
    await the_frame;
    indent -= 1;
    log("[foo] after awaiting the frame of suspendOnce"); // 7
}

fn bar() void {
    log("[bar] awaiting the frame of suspendOnce");
    indent += 1;
    await the_frame;
    indent -= 1;
    log("[bar] after awaiting the frame of suspendOnce");
}

pub fn asyncMain() void {
    log("call async foo()"); // 1
    indent += 1;
    var f = async foo();
    // var b = async bar();
    indent -= 1;
    log("foo suspended by awaiting");
    log("resuming the frame of suspendOnce"); // 5
    indent += 1;
    // resume f;
    resume the_frame;
    indent -= 1;
    log("after resuming the frame of suspendOnce"); // 8
}

pub fn main() !void {
    nosuspend asyncMain();
}
```

> What happens when you resume a frame that's awaiting another?

A panic, by uncommenting the `resume f` in `asyncMain()` you get:
```
panic: awaiting function resumed
```


[...]

PR #9051, opened 2021-06-09
Clarify async/await language documentation.

bnprks:
The async/await documentation was somewhat hard for me to follow when first learning. Two particular sticking points were:
1. The alphabet example constructing the string "abcdefghi" breaks the stated rule that every async has a matching await.
2. It was somewhat unclear to me what the rules for control flow were around async/await constructs.

I've tried to improve this documentation with some minimal explanatory edits, which are correct to the best of my beginner's understanding & experimentation.

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -6509,12 +6509,19 @@ test "global assembly" {
       <p>
       When a function is called, a frame is pushed to the stack,
       the function runs until it reaches a return statement, and then the frame is popped from the stack.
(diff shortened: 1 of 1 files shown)


bnprks:
One additional note: there are some other useful beginner questions in issue #7070, but I didn't include them in this draft of the documentation. I could also add in some of this "what's not allowed" information if it would be desirable (e.g. can't await same frame twice; can't resume a completed async function)

PR #9057, opened 2021-06-09
link: don't link system libs by the wasm linker

kubkon:
The only allowed system libraries that we can link are libraries that are part of the sysroot such as libc or WASI emulated subcomponents. This is required as Wasm allows to defer symbol resolution until load time.

For example, the following import in Zig

```zig
extern "wasi_snapshot_preview1" fn proc_exit() void;
```

would normally result in appending `-lwasi_snapshot_preview1` flag to the linker line. However, for Wasm/WASI, the symbol is provided at load rather than link time, therefore, the linker should not be concerned with resolving the symbol. As a result, we should not consider system libs by the Wasm linker.

cc @jedisct1 hopefully this should fix the regression with zigly you were experiencing!


jedisct1:
👍 Looks like the best thing to do.

And I can confirm that it fixes the regression with zigly!

PR #9058, opened 2021-06-09
Re-enable multiple wasm32 vector tests

### LINKED ISSUE #5339: Miscompiled Vector(N, bool) in wasm32
While working on #5303, I've noticed that formatting `std.meta.Vector(N, bool)` generates erroneous results. Take this snippet for example:

```zig
const std = @import("std");

pub fn main() !void {
    const vs: std.meta.Vector(4, bool) = [_]bool{true, false, true, false};
    std.debug.warn("{}\n", .{vs});
}
```

### Expected result:

```
{ true, false, true, false }
```

### Actual result:

```
{ true, true, true, true }
```

## Some observations
1. This is runtime agnostic (the same behaviour observed in `wasmtime` as well as `wasmer`) which makes me believe it's something in Zig's compiler.
2. Somehow, this seems to be the issue with passing `Vector` (or it wrapped inside some structure) to `std.fmt`. For example, when the elements of `Vector` are printed one-by-one, everything works as expected:

```zig
const std = @import("std");

pub fn main() !void {
    const vs: std.meta.Vector(4, bool) = [_]bool{true, false, true, false};
    var i: usize = 0;
    while (i < 4) : (i += 1) {
        std.debug.warn("{}, ", .{vs[i]});
    }
    std.debug.warn("\n", .{});
}
```
```
true, false, true, false,
```
3. This bug only appears when cross-compiling to `wasm32` **and** when formatting `Vector` of `bool`s.
4. When we swap the first element of `Vector` from `true` to `false`, then all elements are then printed as `false`:

```
{ false, false, false, false }
```

---

I tried to figure out where the source of this error might be coming from, but it seems I know too little about the structure of Zig's compiler. My hunch here is that this is something to do with the way `comptime` and `var` are handled in `wasm32` as the target, and if anyone points me in the right direction, I'll be happy to investigate further and fix it.

kubkon:
Fixes #5339

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -2412,10 +2412,6 @@ test "vector" {
         // https://github.com/ziglang/zig/issues/4486
         return error.SkipZigTest;
     }
-    if (builtin.target.cpu.arch == .wasm32) {
-        // https://github.com/ziglang/zig/issues/5339
-        return error.SkipZigTest;
-    }
 
     const vbool: std.meta.Vector(4, bool) = [_]bool{ true, false, true, false };
     const vi64: std.meta.Vector(4, i64) = [_]i64{ -2, -1, 0, 1 };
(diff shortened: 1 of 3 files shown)


kubkon:
cc @andrewrk

PR #9070, opened 2021-06-10
std.Thread.Futex addition

kprotty:
Having a futex primitive is the first step towards building unified/simpler/efficient synchronization primitives. Some notes on platform compatibility:

* Linux [already has a futex api](https://man7.org/linux/man-pages/man2/futex.2.html). If unfamiliar with futexes in general, its a good resource. 
* Windows uses the [WaitOnAddress](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitonaddress) functions which are only available on Windows 8 and higher. This kills windows 7 support, but we could add an NtKeyedEvent futex-emulation backend if xp to vista/7 is still needed. The implementation uses the ntdll functions because the kernel32 ones [aren't actually in synchronization.lib](https://github.com/Amanieu/parking_lot/blob/master/core/src/thread_parker/windows/waitaddress.rs#L42-L45).
* Darwin uses the new undocumented [`__ulock_wait/wake`](https://crates.io/crates/ulock-sys) API available since 10.12. Its also used in LLVM's [libc++/libcxx](https://github.com/llvm/llvm-project/search?q=__ulock_wait) so it should be still available. There's a possibility it may prevent apps from being uploaded to the AppStore since dynamically linked functions to odd places are checked iirc. If this becomes an issue, we could switch to the generic posix backend.
* The generic posix backend emulates the futex api using a global hash table of wait queues. Its purposely made simple for now but could be optimized by [using fibonacci hashing on the address](https://github.com/Amanieu/parking_lot/search?q=0x9E3779B97F4A7C15), using a [custom Lock](https://github.com/Amanieu/parking_lot/blob/master/core/src/word_lock.rs) for each bucket which can be zero-initialized to reduce hash table size in binary with the [.bss segment](https://en.wikipedia.org/wiki/.bss), and using a balanced tree (red-black, avl, treap, etc.) of queues for each address instead of a flat list. This generic hash table futex approach can also be used for other OS's without a native futex api besides posix.
* For other OS's, we could specialize on their futex-equivalent apis in the future. For reference, [WASM is working on one](https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md), [FreeBSD has one](https://www.freebsd.org/cgi/man.cgi?query=_umtx_op&apropos=0&sektion=2&manpath=FreeBSD+11-current&format=html), [OpenBSD has one](https://man.openbsd.org/futex), [DragonflyBSD has one](https://github.com/golang/go/search?q=sys_umtx_sleep), but [NetBSD does not](https://man.netbsd.org/_lwp_park.2) meaning it would need futex emulation like the generic posix backend.

### DIFF
--- a/lib/std/Thread/Futex.zig
+++ b/lib/std/Thread/Futex.zig
@@ -73,13 +73,12 @@ pub fn wake(ptr: *const Atomic(u32), num_waiters: u32) void {
 
 const OsFutex = if (target.os.tag == .windows)
     WindowsFutex
-    else if (std.builtin.link_libc)
-    PosixFutex
 else if (target.os.tag == .linux)
     LinuxFutex
 else if (target.isDarwin())
     DarwinFutex
-
+else if (std.builtin.link_libc)
+    PosixFutex
 else
     @compileError("Operating System unsupported");
 
@@ -181,18 +180,11 @@ const DarwinFutex = struct {
         const flags = darwin.UL_COMPARE_AND_WAIT | darwin.ULF_NO_ERRNO;
         const status = blk: {
             if (target.os.version_range.semver.max >= 11) {
-                break :blk darwin.__ulock_wait2(flags, addr, expected, timeout_us, 0);
+                break :blk darwin.__ulock_wait2(flags, addr, expect, timeout_us, 0);
             } else {
-                break :blk darwin.__ulock_wait(flags, addr, expected, timeout_us);
+                break :blk darwin.__ulock_wait(flags, addr, expect, timeout_us);
             }
         };
-
-        const status = darwin.__ulock_wait(
-            darwin.UL_COMPARE_AND_WAIT | darwin.ULF_NO_ERRNO,
-            @ptrCast(*const c_void, ptr),
-            @as(u64, expect),
-            timeout_us,
-        );
 
         if (status >= 0) return;
         switch (-status) {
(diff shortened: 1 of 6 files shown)


lib/std/Thread/Futex.zig
```diff
@@ -0,0 +1,578 @@
...
+const single_threaded = std.builtin.single_threaded;
+
+const assert = std.debug.assert;
+const testing = std.testing;
+
+const Atomic = std.atomic.Atomic;
+const spinLoopHint = std.atomic.spinLoopHint;
+
+/// Checks if `ptr` still contains the value `expect` and, if so, blocks the caller until either:
+/// - The value at `ptr` is longer equal to `expect`.
```
lithdew, line 25: The value at `ptr` is no longer equal to `expect`.*

lib/std/Thread/Futex.zig
```diff
@@ -0,0 +1,578 @@
...
+            return error.TimedOut;
+        }
+    }
+
+    return OsFutex.wait(ptr, expect, timeout);
+}
+
+/// Unblocks at most `num_waiters` callers blocked in a `wait()` call on `ptr`.
+/// `num_waiters` of 1 unblocks at most one `wait(ptr, ...)` and `maxInt(u32)` unblocks effectively all `wait(ptr, ...)`.
+pub fn wake(ptr: *const Atomic(u32), num_waiters: u32) void {
```
lithdew, line 66: It looks like quite a few platforms only support waking up either 1 or all waiters. What do you think about replacing `num_waiters: u32` to `wake_all: bool`, or replacing `wake()` with something like `notify()` (wake one waiter) and `broadcast()` (wake all waiters)?

kprotty: `num_waiters` is more of a hint and can be used to influence the exact count for platforms that support it. Other platforms would just wake up more threads than stated. The waiters already have to handle spurious wake ups, and this keeps it consistent with linux's futex(2) api.

andrewrk (approved):
Great work!

PR #9073, opened 2021-06-10
Rename @byteOffsetOf to @offsetOf

Exonorid:
This pull request renames the @byteOffsetOf built-in function in order to be more consistent with the other built-in functions, resolving issue #7794.


Vexu (approved):
We could easily make `zig fmt` update code for this automatically. There should be plenty of examples of it in the code if you want to try doing it.

Issue #4457, opened 2020-02-14
test disabled: atomicrmw with floats


andrewrk:
https://github.com/ziglang/zig/blob/40b9db7cad6f876bb3e8fa32d7b32bbd4bc983ea/test/stage1/behavior/atomics.zig#L148-L152

Related:
 * #3174 
 * #2443
 * #4456

PR #9074, opened 2021-06-10
parse.zig: simplify parsing functions that build lists by always using scratch buffer

mattbork:
These changes simplify parsing functions that duplicate parsing logic in order to special case building lists of one or two items. This worked to avoid unnecessary small allocations. However, since https://github.com/ziglang/zig/pull/8910 avoids all small allocations, these functions can be made simpler and more uniform by always using the scratch buffer. In particular, the duplication of parsing logic is removed, and returns are generally collected together in one place, separate from the parsing logic. This commit rewrites the parsing of statement lists in `parseBlock`, initializer lists in `parseCurlySuffixExpr`, lists of suffix ops and function call arguments in `parseSuffixExpr`, dotted initializer lists and error set declarations in `parsePrimaryTypeExpr`, and argument lists in `parseBuiltinCall`.

One exception is parsing a case list in `parseSwitchProng`. Always using the scratch buffer did not end up making this simpler or clearer. For the sake of completeness, I'll note that this is largely due to the perhaps odd behavior where a one item case produces a `.switch_case_one` node unless it has a trailing comma, in which case it produces a `.switch_case` node in the same way as cases with two or more items. If one item cases always produced `.switch_case_one` nodes, the logic would be more uniform and the function would simplify slightly. Unlike nodes that have explicit trailing-comma variants, like `.array_init_one` versus `.array_init_one_comma`, I don't think anything depends on this behavior. The `firstToken` and `lastToken` functions in `Tree` don't, nor do `switchCase` and `switchCaseOne`, and even `renderSwitchCase` checks for a trailing comma itself for both types of node. 

For testing, I again parsed all valid zig files in this repo with and without these changes and checked that the resulting trees were identical. I also checked for any effect on performance. When using the C allocator, most files (especially larger ones) have a smaller than 1% change in MiB/s of parsed source on my machine, though there are a few outliers as large as +/- 3%. Running the tests with only one version of `parse` checked against itself gave differences in performance as large as 2%, so these numbers are likely noise.

### DIFF
--- a/lib/std/zig/parse.zig
+++ b/lib/std/zig/parse.zig
@@ -2976,53 +2976,48 @@ const Parser = struct {
     ///     <- SwitchItem (COMMA SwitchItem)* COMMA?
     ///      / KEYWORD_else
     fn parseSwitchProng(p: *Parser) !Node.Index {
-        if (p.eatToken(.keyword_else)) |_| {
-            const arrow_token = try p.expectToken(.equal_angle_bracket_right);
-            _ = try p.parsePtrPayload();
-            return p.addNode(.{
+        const scratch_top = p.scratch.items.len;
+        defer p.scratch.shrinkRetainingCapacity(scratch_top);
+
+        if (p.eatToken(.keyword_else) == null) {
+            while (true) {
+                const item = try p.parseSwitchItem();
+                if (item == 0) break;
+                try p.scratch.append(p.gpa, item);
+                if (p.eatToken(.comma) == null) break;
+            }
+            if (scratch_top == p.scratch.items.len) return null_node;
+        }
+        const arrow_token = try p.expectToken(.equal_angle_bracket_right);
+        _ = try p.parsePtrPayload();
+
+        const items = p.scratch.items[scratch_top..];
+        switch(items.len) {
+            0 => return p.addNode(.{
                 .tag = .switch_case_one,
                 .main_token = arrow_token,
                 .data = .{
                     .lhs = 0,
                     .rhs = try p.expectAssignExpr(),
-                },
-            });
-        }
-        const first_item = try p.parseSwitchItem();
-        if (first_item == 0) return null_node;
-
-        if (p.eatToken(.equal_angle_bracket_right)) |arrow_token| {
-            _ = try p.parsePtrPayload();
-            return p.addNode(.{
+                }
+            }),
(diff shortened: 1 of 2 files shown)


andrewrk:
Nice work.

> I'll note that this is largely due to the perhaps odd behavior where a one item case produces a `.switch_case_one` node unless it has a trailing comma, in which case it produces a `.switch_case` node in the same way as cases with two or more items. If one item cases always produced `.switch_case_one` nodes, the logic would be more uniform and the function would simplify slightly.

This seems like a mistake to me. Feel free to change things to match this simplification that you mentioned, if you are so inclined.

By the way, @mattbork, since you have the parser code fresh in your brain, do you mind looking into this test case?

```zig
fn type_expr_return1() if (true) A {}
fn type_expr_return2() for (true) |_| A {}
fn type_expr_return3() while (true) A {}
```

Right now we're getting 

```
../test/behavior/syntax.zig:26:1: error: expected ';' or '{', found 'fn'
fn type_expr_return3() while (true) A {}
^
```

But I think either 
 (1) all of those should parse, or
 (2) none of those should parse

I'm fine with either way really, whatever makes the grammar simpler.

mattbork:
Seems only the function with the for-loop return type expression is parsing incorrectly and it turns out that's because `parseForTypeExpr` is calling `expectExpr` instead of `expectTypeExpr` when parsing its `then_expr`. This lets all of `A {}` get parsed as a `CurlySuffixExpr`, leaving no braces for the function body, when what should happen is that only `A` gets parsed, as a `PrimaryTypeExpr`. Since `parseForTypeExpr` is almost identical to `parseForExpr`, I imagine this was a simple copy-paste mistake.

I'll make the changes for `parseSwitchProng` tomorrow. Is it okay to include the typo-fix for `parseForTypeExpr` as a separate commit here or should I make a new PR for that?

Vexu:
It's perfectly fine to include a fix that small with some other fixes.

andrewrk (approved):

mattbork:
Forgot to leave a comment. These changes are all finished.

Vexu:
All relevant tests passed, failure has something to do with #4457 (cc @LemonBoy).

LemonBoy:
The failure is sussy, the error hasn't popped up in any other CI run and it's a runtime one while the ones in the linked ticket are all happening at compile time.

PR #9047, opened 2021-06-09
stage2 astgen: catch unused vars

g-w1:
I am almost positive the actual impl works (besides a preexisting TODO in asm which I will fix, and maybe some other bugs like the struct type not getting referenced), the main work is just fixing all the code which will take a bunch of work.

If you see your code is in some of the unused vars (either in the code or ci failure), it would help me a lot if you could tell me the correct way to do it as I may not know.

Issues to file:
```zig
                            return astgen.failNodeNotes(ident, "'{s}' not accessible from inner function", .{ident_name}, &.{
                                try astgen.errNoteTok(local_ptr.token_src, "declared here", .{}),
                                // TODO add crossed function definition here note.
                                // Maybe add a note to the error about it being because of the var,
                                // maybe recommend copying it into a const variable. -SpexGuy
```

PR #9081, opened 2021-06-11
std: Fix tanh for negative inputs

LemonBoy:
It turns out the code was not ported correctly from C and produced wrong
results for negative input values. As a bonus fix the NaN codepath by
adding yet another missing piece of code.

Spotted in #9047


g-w1:
Thanks so much for this!

PR #9088, opened 2021-06-12
ci: enable LLVM when testing stage2 on Linux and macOS

kubkon:
Enabling LLVM as part of stage2 tests, will exercise not only stage2 itself but also any drop-in linker replacement, e.g., `zig ld`, with respect to linking C++. I realise this inevitably will add a little overhead to the entire build process, but hopefully it is not anything outrageous.

### DIFF
--- a/ci/azure/linux_script
+++ b/ci/azure/linux_script
@@ -65,9 +65,9 @@ make $JOBS install
 cmake .. -DZIG_EXECUTABLE="$(pwd)/release/bin/zig"
 make $JOBS install
 
-for step in test-toolchain test-std docs; do
-  release/bin/zig build $step -Denable-qemu -Denable-wasmtime
-done
+release/bin/zig build test-toolchain -Denable-qemu -Denable-wasmtime -Denable-llvm
+release/bin/zig build test-std -Denable-qemu -Denable-wasmtime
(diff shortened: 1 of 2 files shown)


kubkon:
Great! This already seems to have caught a problem with zig-bootstrap and linux. cc @andrewrk

andrewrk:
What was the problem? The link expired before I found the time to investigate this.

kubkon:
> What was the problem? The link expired before I found the time to investigate this.

It's a multi-faceted problem. At first level, CMake records `zig` instead of `zig c++` as the CXX compiler which means we end up with `#define ZIG_CXX_COMPILER zig` in `config.h` which will yield errors in the libc++ detection code in [`addCxxKnownPath`](https://github.com/ziglang/zig/blob/1f0b77b3b88112ef458f62c23df72003c843e7cc/build.zig#L469). This can be bypassed creating a shell script wrapper for `zig c++` but then we still hit an error because bootstrapped stage1 fails at finding `libc++.a`. I cannot remember the exact error message, but the issue is with using bootstrapped stage1 with this logic on Linux and Windows. macOS is fine since we don't detect any paths at all.

kubkon:
Gonna close this one for now and we can revisit it later.

PR #9096, opened 2021-06-13
Add OVERLAPPED_ENTRY

### LINKED ISSUE #9079: 'OVERLAPPED_ENTRY' not defined
https://github.com/ziglang/zig/blob/master/lib/std/os/windows.zig#L401
But I cannot found the definition of OVERLAPPED_ENTRY

std\os\windows.zig:401:32: error: use of undeclared identifier 'OVERLAPPED_ENTRY'

codehz:
close #9079


codehz:
Oh, didn't notice it is duplicated

PR #9099, opened 2021-06-13
Fix `lakemont` CpuModel

d18g:
See that [llvm](https://reviews.llvm.org/differential/changeset/?ref=375337) says that ``lakemont`` cpu model implies soft float.
Also Running ``zig build-exe src/main.zig -target i386-freestanding -mcpu lakemont`` results in a segfault.
I used a build.zig file to force the ``soft_float`` feature to be added to the current cpu model and that complied just fine.


LemonBoy:
This feature override must be declared [in this script](https://github.com/ziglang/zig/blob/master/tools/update_cpu_features.zig) and the feature lists re-generated using that.

But beside that, LLVM should be able to fall back to the soft-fp emulation by itself. Can you share a small snippet of code that's currently making the compiler segfault?

d18g:
I am not sure where I should add this override in that script.

The source is just the minimal source generated by ``zig exe-init``.

LemonBoy:
[Here](https://github.com/ziglang/zig/blob/37f36da391570739e639f857bef3de1170eaac50/tools/update_cpu_features.zig#L756), add the following entry:
```zig
            .{
                        .llvm_name = "lakemont",
                        .extra_deps = &.{"soft_float"},
             },
```
If you don't want to run the script add `.soft_float` between `.slow_unaligned_mem_16` and `.vzeroupper`, that's where it'd be added.

andrewrk:
If it's in LLVM's data already, we should be able to simply re-run the script, without needing to add any entries.

LemonBoy:
Lakemont has no x86, no MMX, no SSE and no way of handling any fp-math. In theory LLVM is able to implicitly use the soft-float emulation library calls to legalize any such operation but, given Zig's use of many non-standard features, sometimes we hit a weak spot in the X86 codegen backend.

Consider this as a work-around for this LLVM problem, fixing the problem in LLVM is not so high in my todo list as the target is pretty niche and Intel axed it in '19.

d18g:
@LemonBoy I am aware of that. I am ok with zig not fixing this. At least this PR will document this work around. Thank you for your time and effort.

LemonBoy:
My answer was to Andrew's question about why this change is needed, I do support this patch as it's much better than having the compiler crash.

d18g:
@LemonBoy  I didn't understood you. Re-opening.

lib/std/target/x86.zig
```diff
@@ -2241,6 +2241,7 @@ pub const cpu = struct {
             .cx8,
             .slow_unaligned_mem_16,
             .vzeroupper,
+            .soft_float
```
LemonBoy, line 2244:
```suggestion
            .soft_float,
```

d18g:
Just for reference I opened the command I run with gdb:
image
But the code location is wrong:
image
Also the the evidence of memory corruption is clear:
image

d18g:
Nice and simple ``strace``:
```bash
$ strace -e open zig build-exe -target i386-freestanding -mcpu lakemont src/main.zig
open("/home/XXX/.cache/zig/o/cd33e6b188258554a729da4dd0048da7/builtin.zig", O_RDONLY|O_LARGEFILE) = 18
open("/home/XXX/.zig/zig-linux-x86_64-0.8.0/lib/std/special/c.zig", O_RDONLY|O_LARGEFILE) = 18
open("/home/XXX/.cache/zig/o/cd33e6b188258554a729da4dd0048da7/c.o", O_WRONLY|O_CREAT|O_TRUNC|O_LARGEFILE|O_CLOEXEC, 0666) = 18
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=0x30} ---
+++ killed by SIGSEGV +++
Segmentation fault
```
The ``c.zig`` file export several functions that deals with floats. I am not sure why this file is even compiled as the ``builtin.zig`` file clearly states ``pub const link_libc = false;``.

d18g:
Hmm it does work on godblot, so maybe something in my specific environment?

LemonBoy:
I can confirm the problem. You need a debug build of LLVM to correctly see this problem is triggered by a `llvm_unreachable` in `DAGTypeLegalizer::SoftenFloatOperand`, in release mode you're only going to see garbage.

Here's a MVP showing the problem, compile with `llc /tmp/foo.ll --mtriple i386-- -mcpu=lakemont -o -`:
```llvm
define i64 @foo(double %a) {
  %r = fptoui double %a to i64
  ret i64 %r
}
```

d18g:
@andrewrk Can u make the pull request to though the checks again? It timed out for some reason

LemonBoy:
You can re-trigger the CI by closing the pr, waiting a few minutes then opening it again.

andrewrk:
Even better, trigger it by rebasing against master.

d18g:
All checks ok, please merge.

dhg-intel:
Hmm, If there was a way to set skip_linker_dependencies it might side step this workaround, no?

Issue #8755, opened 2021-05-12
zig ld: a drop-in linker replacement


andrewrk:
Now that there is significant work put into Mach-O linking thanks to @kubkon, Zig is more than just a linker driver for LLD. It is now a linker that competes in the linker space, and some projects may want to use Zig as a drop-in replacement for other linkers.

This issue is to formalize that by adding an "ld" subcommand to Zig's CLI.

This issue does *not* block 1.0.

SaptakBhoumik:
Hi @andrewrk , is zig's linker better at link time optimization?

kubkon:
> Hi @andrewrk , is zig's linker better at link time optimization?

I can speak for the MachO linker which is now fully self-hosted (written in Zig): we currently do not offer support for LLVM's LTO (if that's what you mean that is). LLVM forces a client linker to link dynamically to LTO plugin module which I am strongly against given that we already ship statically linked LLVM. I already started a conversation with LLVM maintainers about it a while back, but didn't see it through as other stuff came up. I will need to follow it up although this is not our priority right now.

SaptakBhoumik:
> > Hi @andrewrk , is zig's linker better at link time optimization?
> 
> I can speak for the MachO linker which is now fully self-hosted (written in Zig): we currently do not offer support for LLVM's LTO (if that's what you mean that is). LLVM forces a client linker to link dynamically to LTO plugin module which I am strongly against given that we already ship statically linked LLVM. I already started a conversation with LLVM maintainers about it a while back, but didn't see it through as other stuff came up. I will need to follow it up although this is not our priority right now.

Hmm I see. Thank for your answer:)

initdc:
Looking forward to this feature!

And you can try this [my rust template](https://github.com/initdc/rust-project-workflow/tree/feat/zig-cc) to get some unsupported linker flags.

absolutelynothinghere:
If possible, please also add `nm`, `objdump`, `strip`, and other LLVM commands

andrewrk:
Zig does not aim to be a drop-in linker. Out of scope.

PR #9103, opened 2021-06-13
zig cc improvements

vrischmann:
This PR improves the linker args parsing for `zig cc`.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1179,6 +1179,16 @@ fn buildOutputType(
                     .wl => {
                         var split_it = mem.split(it.only_arg, ",");
                         while (split_it.next()) |linker_arg| {
+                            // Handle nested-joined args like `-Wl,-rpath=foo`.
+                            // Must be prefixed with 1 or 2 dashes.
+                            if (linker_arg.len >= 3 and linker_arg[0] == '-' and linker_arg[2] != '-') {
+                                if (mem.indexOfScalar(u8, linker_arg, '=')) |equals_pos| {
+                                    try linker_args.append(linker_arg[0..equals_pos]);
+                                    try linker_args.append(linker_arg[equals_pos + 1 ..]);
+                                    continue;
+                                }
+                            }
+
                             try linker_args.append(linker_arg);
                         }
                     },
@@ -1394,7 +1404,7 @@ fn buildOutputType(
                     image_base_override = std.fmt.parseUnsigned(u64, linker_args.items[i], 0) catch |err| {
                         fatal("unable to parse '{s}': {s}", .{ arg, @errorName(err) });
                     };
-                } else if (mem.eql(u8, arg, "-T")) {
+                } else if (mem.eql(u8, arg, "-T") or mem.eql(u8, arg, "--script")) {
                     i += 1;
                     if (i >= linker_args.items.len) {
                         fatal("expected linker arg after '{s}'", .{arg});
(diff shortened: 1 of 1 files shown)


mikdusan:
Rather than handle each `-Wl,-OPTION=ARG` explicitly, how about generically handle any `.comma_joined` by detecting the `-OPTION=ARG` pattern and add 2 arguments to linker_args for further processing?

```diff
diff --git a/src/main.zig b/src/main.zig
index 512828bd7..8da878888 100644
--- a/src/main.zig
+++ b/src/main.zig
@@ -1179,6 +1179,15 @@ fn buildOutputType(
                     .wl => {
                         var split_it = mem.split(it.only_arg, ",");
                         while (split_it.next()) |linker_arg| {
+                            // Handle nested-joined args like `-Wl,-rpath=foo` .
+                            // Must be prefixed with 1 or 2 dashes.
+                            if (linker_arg.len >= 3 and linker_arg[0] == '-' and linker_arg[2] != '-') {
+                                if (mem.indexOfScalar(u8, linker_arg, '=')) |equals_pos| {
+                                    try linker_args.append(linker_arg[0..equals_pos]);
+                                    try linker_args.append(linker_arg[equals_pos + 1..]);
+                                    continue;
+                                }
+                            }
                             try linker_args.append(linker_arg);
                         }
                     },
```

vrischmann:
I wasn't really sure if it would be accepted to alter how `linker_args` is generated here, but ok I can do that.

mikdusan (approved):
Real nice that you clipped-out previous specialized handling of linker `-option=`.

src/main.zig
```diff
@@ -1179,6 +1179,16 @@ fn buildOutputType(
...
                         while (split_it.next()) |linker_arg| {
+                            // Handle nested-joined args like `-Wl,-rpath=foo`.
+                            // Must be prefixed with 1 or 2 dashes.
+                            if (linker_arg.len >= 3 and linker_arg[0] == '-' and linker_arg[2] != '-') {
+                                if (mem.indexOfScalar(u8, linker_arg, '=')) |equals_pos| {
+                                    try linker_args.append(linker_arg[0..equals_pos]);
+                                    try linker_args.append(linker_arg[equals_pos + 1 ..]);
+                                    continue;
+                                }
+                            }
```
andrewrk, line 1190: Are you sure this is correct? Seems weird that it's asking to pass `-rpath=foo` to the linker, but instead we pass `-rpath foo`

src/main.zig
```diff
@@ -1420,30 +1430,40 @@ fn buildOutputType(
                     // We don't need to care about these because these args are
                     // for resolving circular dependencies but our linker takes
                     // care of this without explicit args.
-                } else if (mem.startsWith(u8, arg, "--major-os-version") or
-                    mem.startsWith(u8, arg, "--minor-os-version"))
+                } else if (mem.eql(u8, arg, "--major-os-version") or
+                    mem.eql(u8, arg, "--minor-os-version"))
```
andrewrk, line 1434: This breaks `--major-os-version=foo`

andrewrk (changes requested):

src/main.zig
mikdusan: This is correct. Both macos and linux system linkers accepts BOTH 2-arg and 1-arg-with-equals syntax. Ultimately, we choose the easiest way for our linker-args pass to just see it as 2-arg (which it already supported before this PR).

andrewrk:
Note that `-Wl` is not the only way linker_args gets appended to. There is also --for-linker and -z

andrewrk:
For linker args I'd encourage pushing on #8755 by moving towards a data set of known flags and mappings to zig concepts, same as how we do for clang command line arguments with the tools/update_clang_options.zig.

Minor improvements along the way to our existing linker flag parsing like this are fine as long as they don't have any regressions, but I spotted a problem here.

src/main.zig
mikdusan: I don't understand. Previous to PR these options were no-op but `startsWith` creates some problems:
- accepts `--major-os-version`, `--major-os-version=`, and `--major-os-versionANYTHING`
- if `--major-os-version` is used it is likely the user read the man page and used 2-arg syntax, and in that case following arg is not consumed

afaict this args don't do anything anyways and IMO we should accept them in both 2-arg and 1-arg-with-equals syntaxes to properly eat the command line args.

mikdusan:
On IRC we resolved the outstanding concerns.

Issue #7153, opened 2020-11-18
add support for targeting x86_64-plan9


nektro:
https://9p.io/plan9/

zigazeljko:
Targeting plan9 might be a bit trickier, since it uses a custom binary format which LLVM does not currently support (see http://man.cat-v.org/plan_9/6/a.out). From what I read, outputting that format should be doable with a custom linker script and `--oformat=binary`.

nektro:
@zigazeljko is that not the same `a.out` kinda file that gets output by default of `zig cc` when you dont specify `-o` ?

zigazeljko:
> is that not the same `a.out` kinda file

No. The `a.out` file that gets output by default is actually in ELF (or MachO) format, and has no relation to the actual `a.out` format other than its name.

From Wikipedia:

> "a.out" remains the default output file name for executables created by certain compilers and linkers when no output name is specified, even though the created files actually are not in the a.out format.

g-w1:
I am working on this right now for stage2.

Leimy:
Still being worked? Seems extremely interesting!

g-w1:
Yes, I am just waiting for the self hosted compiler to get better at codegen. Once that happens, progress will be really quick and contributor friendly for adding plan 9 stuff to the stdlib. (The linking backend is already in place)

zigazeljko:
@g-w1 By the way, is the plan9 syscall ABI (i.e. the stuff in `os/plan9/x86_64.zig`) documented anywhere?

g-w1:
The syscalls are here http://aiju.de/plan_9/plan9-syscalls, and the abi is the same as the c one.

zigazeljko:
The page you linked describes the API provided by libc. What i'm looking for is the underlying ABI (application **binary** interface) provided by the kernel -- how the values are passed to/from kernel and what instructions to use to actually perform the system call.

Taking your `x86_64` implementation as an example, how did you find out that:

* syscall number is passed in `rbp`;
* individual arguments are passed on stack, with a zero pushed on top;
* the actual system call is performed via the `syscall` instruction (as opposed to `int 0x40`, for example);
* the result is returned in `rax`;
* register `rcx` gets clobbered.

g-w1:
Reading generated code from the c compiler/playing around a LOT. I don't know if there is a manpage for the abi itself.

TheDevtop:
[...]

PR #9105, opened 2021-06-13
stage2: bringup plan9 target and linker

g-w1:
* [x] sucessful `pub fn main() void {}`
* [x] Convert to use a global offset table (or just some fancy relocations)
* [x] sucessful hello world (This may be harder as the pwrite syscall needs arguments on the stack, so I might have to implement more inline asm which may be out of scope). ~~Out of scope, MOV for inline assembly will take more work in stage2.~~
* [x] use correct code for 64 bit targets (not `@intCast`) to u32
* [x] remove some panics/TODOs
scrot

Makes strides towards #7153.


kubkon:
How much is left (is it viable) for having plan9 autotested in stage2 tests?

g-w1:
Very good question! Sourcehut has support for plan9 on ci https://sourcehut.org/blog/2020-05-11-sourcehut-plus-plan-9/, so we could probably use that, but maybe not until its self-hosting or we can generate the binaries on a different platform, then run them on the plan9 ci.
Russ Cox also made a thing, kind of like wine, that can run plan9 user programs on some unixes. https://swtch.com/9vx/ Maybe we could use this on the ci? (I haven't used it yet, I'll try it soon)

kubkon:
> Very good question! Sourcehut has support for plan9 on ci https://sourcehut.org/blog/2020-05-11-sourcehut-plus-plan-9/, so we could probably use that, but maybe not until its self-hosting or we can generate the binaries on a different platform, then run them on the plan9 ci.
> Russ Cox also made a thing, kind of like wine, that can run plan9 user programs on some unixes. https://swtch.com/9vx/ Maybe we could use this on the ci? (I haven't used it yet, I'll try it soon)
> Edit: Someone on irc also said that you could just boot plan9 in qemu and run the program in it.

Hmm, I guess srht would be our best bet since rolling out wine-like emulation or QEMU might be a tough thing to do inside another VM (i.e., the CI). Perhaps it could be done though?

g-w1:
Successful hello world!!!: 
image

g-w1:
For reviewers: here is the link to the man page used for reference http://git.9front.org/plan9front/plan9front/819bec22ccf01927c8d31e9bbcddca91b5ed41ee/sys/man/6/a.out/f.html

lib/std/zig.zig
```diff
@@ -178,8 +178,32 @@ pub fn binNameAlloc(allocator: *std.mem.Allocator, options: BinNameOptions) erro
         .spirv => return std.fmt.allocPrint(allocator, "{s}.spv", .{root_name}),
         .hex => return std.fmt.allocPrint(allocator, "{s}.ihex", .{root_name}),
         .raw => return std.fmt.allocPrint(allocator, "{s}.bin", .{root_name}),
+        .plan9 => return std.fmt.allocPrint(allocator, "{s}.{c}", .{ root_name, archToPlan9Char(target.cpu.arch) }),
     }
 }
+fn archToPlan9Char(arch: std.Target.Cpu.Arch) ?u8 {
```
andrewrk, line 184:
```suggestion
fn archToPlan9Char(arch: std.Target.Cpu.Arch) u8 {
```

src/link.zig
```diff
@@ -328,6 +335,7 @@ pub const File = struct {
             .elf => return @fieldParentPtr(Elf, "base", base).updateDeclLineNumber(module, decl),
             .macho => return @fieldParentPtr(MachO, "base", base).updateDeclLineNumber(module, decl),
             .c => return @fieldParentPtr(C, "base", base).updateDeclLineNumber(module, decl),
+            .plan9 => @panic("PLAN 9 DEBUG INFO"),
```
andrewrk, line 338:
```suggestion
            .plan9 => @panic("TODO: implement updateDeclLineNumber for plan9"),
```

src/link/Plan9.zig
```diff
@@ -0,0 +1,377 @@
...
+const codegen = @import("../codegen.zig");
+const trace = @import("../tracy.zig").trace;
+const mem = std.mem;
+const File = link.File;
+const Allocator = std.mem.Allocator;
+
+const log = std.log.scoped(.link);
+const assert = std.debug.assert;
+
+// TODO use incremental compilation
```
andrewrk, line 17:
```suggestion
//! This implementation does all the linking work in flush(). A future improvement
//! would be to add incremental linking in a similar way as ELF does.
```

src/link/plan9/a.out.zig
```diff
@@ -0,0 +1,134 @@
...
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
```
andrewrk, line 20: this zig code is not copyright-owned by plan 9 foundation; you can and should omit this copyright notice.

andrewrk (changes requested):
I have only cosmetic changes requested on this. Impressive work!

PR #9106, opened 2021-06-13
Add formatting check to CI pipeline

Vexu:
Reopening of #8219 now that the parsing issue has been fixed. I also added a `check-ast` on the self hosted compiler `main.zig` since that should also test the standard library.

### DIFF
--- a/ci/azure/linux_script
+++ b/ci/azure/linux_script
@@ -60,7 +60,7 @@ unset CXX
 make $JOBS install
 
 # look for formatting errors
-release/bin/zig fmt --check .. || (echo "Please run 'zig fmt' to fix the non-conforming files listed above." && false)
+release/bin/zig fmt --check --ast-check .. || (echo "Please run 'zig fmt' to fix the non-conforming files listed above." && false)
 
 # Here we rebuild zig but this time using the Zig binary we just now produced to
 # build zig1.o rather than relying on the one built with stage0. See
--- a/ci/azure/linux_script
+++ b/ci/azure/linux_script
@@ -59,6 +59,9 @@ unset CXX
 
 make $JOBS install
 
+# look for formatting errors
+release/bin/zig fmt --check .. || (echo "Please run 'zig fmt' to fix the non-conforming files listed above." && false)
+
 # Here we rebuild zig but this time using the Zig binary we just now produced to
 # build zig1.o rather than relying on the one built with stage0. See
 # https://github.com/ziglang/zig/issues/6830 for more details.
--- a/lib/std/fmt/parse_float.zig
+++ b/lib/std/fmt/parse_float.zig
@@ -34,7 +34,7 @@
 // - Only supports round-to-zero
 // - Does not handle denormals
 
-const std = @import("../std.zig");
+const std = @import("std");
 const ascii = std.ascii;
 
 // The mantissa field in FloatRepr is 64bit wide and holds only 19 digits
@@ -231,6 +231,8 @@ fn parseRepr(s: []const u8, n: *FloatRepr) !ParseResult {
                 } else if (c == '.') {
                     i += 1;
                     state = .LeadingFractionalZeros;
+                } else if (c == '_') {
+                    i += 1;
                 } else {
                     state = .MantissaIntegral;
                 }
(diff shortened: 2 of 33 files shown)


g-w1:
Ast-check does not follow imports, so you should do `find .. -name "*.zig" | xargs ./zig ast-check` or something similar.

ci/azure/linux_script
```diff
@@ -59,6 +59,10 @@ unset CXX
 
 make $JOBS install
 
+# look for formatting errors
+release/bin/zig fmt --check ../lib/std ../doc ../src ../tools ../test ../build.zig
+release/bin/zig ast-check ../src/main.zig
```
andrewrk, line 64: I suggest we only do the zig fmt check for now, for a few reasons:
 * This doesn't actually check any files other than src/main.zig
 * std lib AST checking is already happening with every stage2 test
 * It's an unrelated change to adding zig fmt checking 
 * we may want to consider taking advantage of the thread pool and cache, which is already happening with std lib AST checking

ci/azure/linux_script
```diff
@@ -59,6 +59,10 @@ unset CXX
 
 make $JOBS install
 
+# look for formatting errors
+release/bin/zig fmt --check ../lib/std ../doc ../src ../tools ../test ../build.zig
```
andrewrk, line 63: Suggestion, to make it more clear what to do for a contributor whose PR is not passing:

` || echo "Please run 'zig fmt' to fix the non-conforming files listed above."`

ci/azure/linux_script
Vexu: >     * This doesn't actually check any files other than src/main.zig

Solved with `find . -name  "*.zig" ! -path "*/zig-cache/*" ! -path "*/zig-out/*" | xargs -n1 release/bin/zig ast-check`.

>     * std lib AST checking is already happening with every stage2 test

Std files can be easily excluded in the command.

>     * It's an unrelated change to adding zig fmt checking

Sure, it did find a some bugs in stage2 and a lot of nonconforming code everywhere else so it might be better to do separately.  

>     * we may want to consider taking advantage of the thread pool and cache, which is already happening with std lib AST checking

It takes less than 15 seconds to check all ~900 files on my laptop on a debug build, I don't think performance is going to be that big of an issue.

ci/azure/linux_script
```diff
@@ -61,6 +61,7 @@ make $JOBS install
 
 # look for formatting errors
 release/bin/zig fmt --check .. || (echo "Please run 'zig fmt' to fix the non-conforming files listed above." && false)
+find . -name  "*.zig" ! -path "*/zig-cache/*" ! -path "*/zig-out/*" ! -path "*/build/*" ! -wholename "*/special/compiler_rt.zig" ! -wholename "*/special/compiler_rt/atomics.zig"  | xargs -n 1 build/zig ast-check
```
Vexu, line 64: I excluded some compiler_rt files since stage1 and stage2 don't agree on how `@export` works.

andrewrk: instead of this, what about adding `--ast-check` to `zig fmt` since it's already doing the work of finding .zig files and excluding zig-cache?

lib/std/fmt/parse_float.zig
```diff
@@ -34,7 +34,7 @@
 // - Only supports round-to-zero
 // - Does not handle denormals
 
-const std = @import("../std.zig");
+const std = @import("std");
```
g-w1, line 37: Why was this changed? I think ../std.zig is more correct.

Vexu: There is a pretty much even divide between `@import("std.zig")` and `@import("std")`, I changed this because I wanted to test this file specifically.

ci/azure/linux_script
Vexu: What should I do about the two compiler_rt files that do not currently pass the check due to the `@export` error?
```
error: the first @export parameter must be an identifier
        => @export(@import("compiler_rt/stack_probe.zig").zig_probe_stack, .{
                                                         ^
```

Vexu:
```
$ hyperfine "zig-out/bin/zig fmt . --ast-check --check" --warmup 1
Benchmark #1: zig-out/bin/zig fmt . --ast-check --check
  Time (mean ± σ):      6.056 s ±  0.022 s    [User: 2.499 s, System: 3.552 s]
  Range (min … max):    6.022 s …  6.094 s    10 runs
```

Not too bad for ~860 files.

lib/std/fmt/parse_float.zig
andrewrk: Generally, std lib should import std.zig by file path, setting an example for other packages. For `zig test` you can use `--main-pkg-path` to test files that import with "..".

PR #9114, opened 2021-06-14
stage2: Remove special double ampersand parsing case

### LINKED ISSUE #8046: Valid usage of `&&` incorrectly reported as a compile error
[see below for the fix](https://github.com/ziglang/zig/issues/8046#issuecomment-844589399)

--------

Currently, `&&expression` is illegal, but `& &expression` is legal, and returns the address of the address of `expression`.
However, `zig fmt` changes `& &expression` to `&&expression`. 

*`bug.zig`*
```zig
const std = @import("std");

pub fn main() void {
    var a: usize = 4;
    var b = & &a;
    b.*.* = 92;
    std.debug.print("{}\n", .{a});
}
```
Running `zig fmt bug.zig` twice yields
```
bug.zig:5:13: error: Expected expression, found '&&'
    var b = &&a;
            ~~
```

greenfork:
Fixes #8046

I'm stuck with finding how to generate an error on AstGen level. My assumption is that I should insert `astgen.failTok` somewhere [here](https://github.com/ziglang/zig/blob/master/src/AstGen.zig#L686-L689). Or in a better place where the tree is analyzed and not constructed. For that I'd want to understand how zig parses the now incorrect structure into AST, I tried to get a flattened AST with the following snippet and so far failed:
```zig
const std = @import("std");

pub fn main() !void {
    defer arena.deinit();
    const ally = &arena.allocator;

    const src =
        \\const a = (true && false);
    ;
    const tree = try std.zig.parse(ally, src);
    const rendered_tree = try tree.render(ally);
    std.debug.print("{s}\n", .{rendered_tree});
}
```

I'm new here, let me know if I'm doing something wrong.

### DIFF
--- a/lib/std/zig/parse.zig
+++ b/lib/std/zig/parse.zig
@@ -1333,7 +1333,6 @@ const Parser = struct {
         .keyword_or = .{ .prec = 10, .tag = .bool_or },
 
         .keyword_and = .{ .prec = 20, .tag = .bool_and },
-        .invalid_ampersands = .{ .prec = 20, .tag = .bool_and },
 
         .equal_equal = .{ .prec = 30, .tag = .equal_equal, .assoc = Assoc.none },
         .bang_equal = .{ .prec = 30, .tag = .bang_equal, .assoc = Assoc.none },
@@ -1384,9 +1383,6 @@ const Parser = struct {
             switch (tok_tag) {
                 .keyword_catch => {
                     _ = try p.parsePayload();
-                },
-                .invalid_ampersands => {
-                    try p.warn(.invalid_and);
                 },
                 else => {},
             }
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -554,14 +554,19 @@ fn expr(gz: *GenZir, scope: *Scope, rl: ResultLoc, node: ast.Node.Index) InnerEr
         .bit_and  => {
             const current_ampersand_token = main_tokens[node];
             if (token_tags[current_ampersand_token + 1] == .ampersand) {
-                return astgen.failTok(
-                    current_ampersand_token,
-                    "`&&` is invalid; note that `and` is boolean AND",
-                    .{},
-                );
-            } else {
-                return simpleBinOp(gz, scope, rl, node, .bit_and);
+                const token_starts = tree.tokens.items(.start);
+                const current_token_offset = token_starts[current_ampersand_token];
+                const next_token_offset = token_starts[current_ampersand_token + 1];
+                if (current_token_offset + 1 == next_token_offset) {
+                    return astgen.failTok(
+                        current_ampersand_token,
+                        "`&&` is invalid; note that `and` is boolean AND",
+                        .{},
+                    );
+                }
             }
+
+            return simpleBinOp(gz, scope, rl, node, .bit_and);
         },
         .bit_or   => return simpleBinOp(gz, scope, rl, node, .bit_or),
         .bit_xor  => return simpleBinOp(gz, scope, rl, node, .xor),
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -551,7 +551,18 @@ fn expr(gz: *GenZir, scope: *Scope, rl: ResultLoc, node: ast.Node.Index) InnerEr
         .mul_wrap => return simpleBinOp(gz, scope, rl, node, .mulwrap),
         .div      => return simpleBinOp(gz, scope, rl, node, .div),
         .mod      => return simpleBinOp(gz, scope, rl, node, .mod_rem),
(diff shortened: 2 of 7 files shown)


g-w1:
If I were doing this, I would poke around main_tokens in that place you linked to see if there was an identifier as the previous token, the next token is &, and the token after that is an identifier. Then report a nice compile error, maybe failToking on one of the &s

greenfork:
@g-w1 thank you a lot!

I did my poking with
```zig
            const idx = 2;
            std.debug.print(
                "in .address_of expr branch: {d}, {}, {}, {}\n",
                .{ idx, node_datas[idx], node_tags[idx], token_tags[main_tokens[idx]] },
            );
```
which is 1 attempt per ~40 seconds. Is there a better approach than printf debugging?

greenfork:
Current solution produces this error as well
```
error.zig:1:16: error: `&&` is invalid; note that `and` is boolean AND
const a = true & &false;
               ^
```
which is a bit counter-intuitive. I don't know if this should be improved.

Current solution also depends on the fact that, in my opinion, bitwise "and" with an address is something wrong. I don't know if that is completely true.

g-w1:
If you want to fix that, go even further and use tree.source and then find the sources of the tokens and make sure that there is no whitespace between the tokens.

g-w1:
> Is there a better approach than printf debugging?

Yes! Use a debugger like gdb or lldb.

greenfork:
> If you want to fix that, go even further and use tree.source and then find the sources of the tokens and make sure that there is no whitespace between the tokens.

I wasn't sure if that is going to breach the abstraction layers, that high up in the AST we shouldn't directly inspect the source text. But it could be okay for errors since we get an error from semantic analysis anyway if we skip this one. I will add that then

g-w1:
Yeah, its not the best, but it seems like the best way to fix this. IMO we should rather give better errors than abide by some arbitrary guideline that hinders our ability to do so.

src/AstGen.zig
```diff
@@ -551,7 +551,18 @@ fn expr(gz: *GenZir, scope: *Scope, rl: ResultLoc, node: ast.Node.Index) InnerEr
         .mul_wrap => return simpleBinOp(gz, scope, rl, node, .mulwrap),
         .div      => return simpleBinOp(gz, scope, rl, node, .div),
         .mod      => return simpleBinOp(gz, scope, rl, node, .mod_rem),
-        .bit_and  => return simpleBinOp(gz, scope, rl, node, .bit_and),
+        .bit_and  => {
+            const current_ampersand_token = main_tokens[node];
+            if (token_tags[current_ampersand_token + 1] == .ampersand) {
```
g-w1, line 556: I'm not sure this fixes the error. Does 
```zig
const y = &&ident;
```
Work correctly after this?
If so, please add a test.
The way I would do this is to see if the previous token was an identifier (and maybe the one after) in addition to what you have, because then that would be invalid.

greenfork: Added the test for the case you have mentioned.

My understanding is that `bit_and`, as a binary operator, guarantees that the left operand is always an expression, i.e. identifier, literal boolean etc. So we should be okay when checking only for the right-hand side of the expression.

greenfork:
Added the check that 2 ampersands are strictly adjacent to each other in the source code for this error to be raised.

test/stage2/test.zig
```diff
@@ -1569,4 +1565,27 @@ pub fn addCases(ctx: *TestContext) !void {
...
+            \\    const a = true;
+            \\    const b = false;
+            \\    const c = a & &b;
+            \\}
+        , &[_][]const u8{":4:17: error: incompatible types: 'bool' and '*const bool'"});
+
+        case.addCompareOutput(
+            \\pub fn main() void {
+            \\    const b: u8 = 1;
+            \\    const c = &&b;
```
g-w1, line 1587: `c` is unused here which will cause a compile error with https://github.com/ziglang/zig/pull/9047 . Please fix it by adding _ = c; or changing it to _ = &&b;

greenfork: Sure, fixed

g-w1 (approved):
This looks good! Nice work.

lib/std/zig/parse.zig
```diff
@@ -1385,9 +1384,6 @@ const Parser = struct {
                 .keyword_catch => {
                     _ = try p.parsePayload();
                 },
-                .invalid_ampersands => {
-                    try p.warn(.invalid_and);
-                },
```
shachaf, line 1390: Not very important, but the comment about "&&" above doesn't make sense after this change (I was confused for a bit trying to figure out what it meant). This switch can probably just be an if now.

PR #9117, opened 2021-06-14
stage2: Unify compiler error printing

### LINKED ISSUE #8872: unify the two ways that compile errors get printed to stderr
https://github.com/ziglang/zig/blob/63aabbbba7872eb5178c32235bba260a13d6869a/src/Compilation.zig#L328

vs

https://github.com/ziglang/zig/blob/63aabbbba7872eb5178c32235bba260a13d6869a/src/main.zig#L3133

Probably the latter should be changed to convert to the Compilation error message struct. You can see an example of this code here:

https://github.com/ziglang/zig/blob/63aabbbba7872eb5178c32235bba260a13d6869a/src/Module.zig#L2467-L2476

and then

https://github.com/ziglang/zig/blob/63aabbbba7872eb5178c32235bba260a13d6869a/src/Compilation.zig#L394-L441

greenfork:
Fixes #8872

This introduces the following limitations on printing function in `main.zig`:
- Always prints to stderr, can't print to file now
- Tilde printing `~~~~` is no more, it always points to an error with a caret `^`

I tested it with `zig ast-check` on this file:
```zig
const std = @import("std");

pub fn main() !void {
    const a = true && false;
}
```
and it correctly produces:
```
error.zig:4:23: error: `&&` is invalid; note that `and` is boolean AND
    const a = true && false;
                      ^
```
with previous being:
```
error.zig:4:23: error: `&&` is invalid; note that `and` is boolean AND
    const a = true && false;
                      ~~~~~
```

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -2940,7 +2940,6 @@ const Fmt = struct {
 };
 
 pub fn cmdFmt(gpa: *Allocator, args: []const []const u8) !void {
-    const stderr_file = io.getStdErr();
     var color: Color = .auto;
     var stdin_flag: bool = false;
     var check_flag: bool = false;
@@ -2998,7 +2997,7 @@ pub fn cmdFmt(gpa: *Allocator, args: []const []const u8) !void {
         defer tree.deinit(gpa);
 
         for (tree.errors) |parse_error| {
-            try printErrMsgToFile(gpa, parse_error, tree, "<stdin>", stderr_file, color);
+            try printErrMsgToStdErr(gpa, parse_error, tree, "<stdin>", color);
         }
         if (tree.errors.len != 0) {
             process.exit(1);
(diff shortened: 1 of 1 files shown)


LemonBoy:
> with previous being:

Having the error messages highlight the whole span makes them much clearer IMO.

greenfork:
> Having the error messages highlight the whole span makes them much clearer IMO.

I agree with that. Would you suggest to change this behavior in this PR? This looks like a separate scope and will likely affect all the errors, not just the errors from `main.zig`. I can create an issue and a follow-up pull request to change it, or I can do it in the current PR.

LemonBoy:
> Would you suggest to change this behavior in this PR?

Whatever you're more comfortable with, this PR looks ready to merge so making a follow-up one is fine.

greenfork:
Alright, thank you! I would rather make a follow-up pull request after this one is merged.

andrewrk:
Nice work :+1: 

The tilde thing is going to be fun. I suspect it will be non-trivial.

Issue #8057, opened 2021-02-23
debug.SymbolInfo, debug.SymbolInfo.deinit and debug.LineInfo.deinit are all private


vktec:
Values of type `debug.SymbolInfo` can be obtained using `ModuleDebugInfo.getSymbolAtAddress`, but a variable or parameter of that type cannot be explicitly declared.
Values of the type cannot be deinitialized, as the `deinit` function is private - same for `LineInfo.deinit`.

```diff
diff --git a/lib/std/debug.zig b/lib/std/debug.zig
index a434fe0e8..6276aa0ad 100644
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -44,7 +44,7 @@ pub const LineInfo = struct {
     file_name: []const u8,
     allocator: ?*mem.Allocator,
 
-    fn deinit(self: LineInfo) void {
+    pub fn deinit(self: LineInfo) void {
         const allocator = self.allocator orelse return;
         allocator.free(self.file_name);
     }
@@ -1338,12 +1338,12 @@ pub const DebugInfo = struct {
     }
 };
 
-const SymbolInfo = struct {
+pub const SymbolInfo = struct {
     symbol_name: []const u8 = "???",
     compile_unit_name: []const u8 = "???",
     line_info: ?LineInfo = null,
 
-    fn deinit(self: @This()) void {
+    pub fn deinit(self: @This()) void {
         if (self.line_info) |li| {
             li.deinit();
         }
```

g-w1:
Feel free to make a pr.

PR #9123, opened 2021-06-15
enables user-custom code to work with LI and SI

ityonemo:
if user code wants to play around stacktraces, making these functions public lets that code be a good citizen and clean up the resources which get created (usually by other pub functions).


jedisct1:
Sure, I don't see any good justification for keeping these private.

luehmann:
This should resolve #8057

PR #9129, opened 2021-06-15
Fix std.Thread.Condition.wait on Windows

jcmoyer:
Fixes a compile error on Windows when using std.Thread.Condition.wait():
```
D:\...\zig-install\lib\zig\std\Thread\Condition.zig:57:19: error: no member named 'srwlock' in struct 'std.Thread.Mutex'
            &mutex.srwlock,
                  ^
```


LemonBoy:
See #8637 for a fix and some related and ongoing discussion.

PR #9138, opened 2021-06-17
HashMap.getOrPutAssumeCapacityAdapted should set key to undefined

Hadron67:
As the doc comment states, this method sets the key and value to undefined if no existing entry was found and the caller should then initialize the key and value. But it looks like the new key was mistakenly set to the adapted key passed into this method, which will cause a compile error if its type is different from the real key.


LemonBoy:
Needs a test.

Hadron67:
Done :)

PR #9144, opened 2021-06-17
stage2 Sema: implement @intToPtr

g-w1:
I am not sure how to test this and the `int_u64` checking feels a little weird, can anyone think of any better way to represent a Value pointer?

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -5769,29 +5769,50 @@ fn zirIntToPtr(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) InnerErro
     const type_res = try sema.resolveType(block, src, extra.lhs);
     if (type_res.zigTypeTag() != .Pointer)
         return sema.mod.fail(&block.base, type_src, "expected pointer, found '{}'", .{type_res});
+    const ptr_align = type_res.ptrAlignment(sema.mod.getTarget());
 
     const uncasted_operand = try sema.resolveInst(extra.rhs);
     if (try sema.resolveDefinedValue(block, operand_src, operand_coerced)) |val| {
-        if (!type_res.isAllowzeroPtr() and val.isZero())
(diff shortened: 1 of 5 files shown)


src/type.zig
```diff
@@ -1530,15 +1530,15 @@ pub const Type = extern union {
             .const_slice,
             .mut_slice,
             => {
-                if (self.elemType().hasCodeGenBits()) return @divExact(target.cpu.arch.ptrBitWidth(), 8) * 2;
+                if (!self.elemType().hasCodeGenBits()) return @divExact(target.cpu.arch.ptrBitWidth(), 8) * 2;
```
LemonBoy, line 1533: This looks wrong, if the element type is non-zero sized you have both a pointer and a length field, notice the `* 2` at the end.

g-w1: Ah, for the other ones I think it was a typo, but for this one it was not. Thanks

src/Sema.zig
```diff
@@ -5785,6 +5785,14 @@ fn zirIntToPtr(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) InnerErro
         });
     }
     try sema.requireRuntimeBlock(block, src);
+    if (block.wantSafety()) {
+        const zero = try sema.mod.constInst(sema.arena, src, .{
+            .ty = Type.initTag(.u64),
+            .val = Value.initTag(.zero),
+        });
+        const is_non_zero = try block.addBinOp(src, Type.initTag(.bool), .cmp_neq, operand_coerced, zero);
```
LemonBoy, line 5793: This is wrong as some pointer types do accept zero as value. Check out `ir_analyze_int_to_ptr`, the pointer alignment must be validated too.

Vexu: Right, thanks fixed.

PR #9145, opened 2021-06-17
stage2: simplify codegen for errorToInt and intToError

g-w1:
We can just use bitcast instead of error_to_int, int_to_error since
errorToInt and intToError do not actually do anything, just change types.
This allows us to remove 2 air ops that were the exact same as bitcast


Vexu (approved):
There shouldn't be any semantic meaning to these instructions as we add safety checks in semantic analysis in stage2.

PR #9156, opened 2021-06-18
Add support for the NO_COLOR environment variable

SpexGuy:
This PR adds checks for the [widely supported](http://no-color.org/) `NO_COLOR` environment variable.  If this variable exists, the default is changed from tty detection to no color.  Command line arguments will still take precedence over this environment variable.

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -90,11 +90,11 @@ pub fn getSelfDebugInfo() !*DebugInfo {
 }
 
 pub fn detectTTYConfig() TTY.Config {
-    var bytes: [128]u8 = undefined;
-    const allocator = &std.heap.FixedBufferAllocator.init(bytes[0..]).allocator;
-    if (process.getEnvVarOwned(allocator, "ZIG_DEBUG_COLOR")) |_| {
+    if (process.hasEnvVarConstant("ZIG_DEBUG_COLOR")) {
         return .escape_codes;
-    } else |_| {
+    } else if (process.hasEnvVarConstant("NO_COLOR")) {
+        return .no_color;
+    } else {
         const stderr_file = io.getStdErr();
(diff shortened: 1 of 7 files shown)


g-w1 (changes requested):
Could you also add it here?https://github.com/ziglang/zig/blob/fc1feebdc0ffd4730c724780265d4ef6a9515dc8/lib/std/debug.zig#L88-L103 and https://github.com/ziglang/zig/blob/fc1feebdc0ffd4730c724780265d4ef6a9515dc8/src/stage1/errmsg.cpp#L20
There may be other places in the codebase where this is needed.

SpexGuy:
@g-w1 good find, I'll update those too.  I don't see an existing way to get env vars in stage 1.  Should I add that to os.cpp or am I missing something?

g-w1:
I implemented something similar but worse in this pr, https://github.com/ziglang/zig/pull/7932/files
```cpp
bool os_is_dumb_term(void) {
    char* term = getenv("TERM");
    if (term == nullptr) return false;
    return strcmp(term, "dumb") == 0;
}
```

SpexGuy:
uh, what

...did this just get merged?

Oh fuck

g-w1:
I think you just closed it.

andrewrk:
If you accidentally deleted git commits, check your `git reflog` and the data is likely still there

SpexGuy:
Nothing's been deleted, my branch still exists, but I don't know what happened.  Doesn't look like it got merged into the zig repo, which is good.  The repoen button is greyed out, so I can't click that.  I guess I'll make a new PR?

SpexGuy:
I accidentally force pushed before committing, which apparently causes github to auto-close the PR since there are no changes.  Oops xD

SpexGuy:
Rebased to fix merge conflicts

Vexu (approved):

Issue #5656, opened 2020-06-21
Enums aren't cast from ints in macros with translate-c


shtanton:
macro.h:
```
#define noargs someargs(Flag1)

typedef enum {
    Flag0,
    Flag1
} Flags;

int someargs(Flags flags);
```
macro.c:
```
typedef enum {
    Flag0,
    Flag1
} Flags;

int someargs(Flags flags) {
    if (flags == Flag1) {
        return 1;
    } else {
        return 0;
    }
}
```
test.zig:
```
const c = @cImport({
    @cInclude("macro.h");
});

pub fn main() !void {
    _ = c.noargs;
}
```
Then compile with `zig build-exe test.zig --c-source macro.c -lc -I.` and the error
```
zig-cache/o/INfBWWWMN1znJ7cpk1aHFpWrWqAEwqd-Uefh0pPlK7Sau8WYxYrtuxOF09n_V9xa/cimport.zig:225:35: error: expected type '.cimport:2:11.enum_unnamed_1', found 'c_int'
```
is produced.

As far as I can tell this is only a problem on the right hand side of c macros, in other places `@intToEnum` is inserted to make it work. I don't know why expressions in macros are handled differently but it's probably causing other bugs too.

Vexu:
Macros are translated separately from the rest of the code and have no type info available which is why the cast is not being inserted. Fixing this would require finding the types of the function and the argument (possibly through multiple macros) and casting based on that.

Issue #2115, opened 2019-03-28
enums should disallow multiple enumerations with the same value


tiehuis:
@andrewrk: [I re-opened this to propose reversing the decision](https://github.com/ziglang/zig/issues/2115#issuecomment-827968279)

-----

A C enum allows multiple values. e.g.

```c
enum Foo {
    OPTION1 = 0,
    OPTION2 = 1,
    DEFAULT = OPTION1,
};
```

However, zig `extern enum`'s do not (note: we cannot declare a member based on another member, another possibly improvement)

```zig
const Foo = extern enum {
    Option1 = 0,
    Option2 = 1,
    Default = 0,
};

test "asd" {
    const a = Foo.Default;
}
```

Fails with compilation

```
/tmp/t.zig:4:15: error: enum tag value 0 already taken
    Default = 0,
              ^
/tmp/t.zig:2:15: note: other occurrence here
    Option1 = 0,
              ^
```

We should allow extern enums to have the same tag value for multiple members. The specific use-case here would be for translate-c. Encountered this when trying to process `gmp.h` which has this enum present:

```c
/* Available random number generation algorithms.  */
typedef enum
{
  GMP_RAND_ALG_DEFAULT = 0,
  GMP_RAND_ALG_LC = GMP_RAND_ALG_DEFAULT /* Linear congruential.  */
} gmp_randalg_t;
```

Alternatively for translate-c enums could be translated as sets of constants but this feels wrong. There are possibly some questions regarding switch cases and handling of duplicate values there that I haven't considered.

Hejsil:
I'll add something related to this, that I've hit before.
C users also uses `enum`s for flags, so sometimes you'll see this in headers:
```cpp
enum flags {
    A = 1 << 0,
    B = 1 << 1,
};
enum flags all = A | B;
```

translate_c will also fail to translate something like this if `all` is ever used.

tgschultz:
It's possible to do this using definitions:

```
pub const Compression = extern enum(u32)
{
    Uncompressed    =   0,
    RLE8            =   1,  //8bpp only
    RLE4            =   2,  //4bpp only
    Bitfields       =   3,  //Windows, 16bpp/32bpp
    
    //These seem to only exist for passthrough to
    // devices like printers and don't actually
    // exist in .bmp files
    JPEG            =   4,  //Windows v3
    PNG             =   5,  //Windows v3
    
    //This is legit
    AlphaBitfields  =   6,  //Windows v4
    
[...]

PR #9164, opened 2021-06-19
translate-c: Remove usage of `extern enum`

### LINKED ISSUE #9153: translate-c: enum constants can have incorrect type
The following code translates incorrectly due to the underlying int type of an enum type being different from the int type of an enum constant (in the below C code, a variable of type `foo` would have an underlying representation of `unsigned`, but the enum constant `bar` has type `int`).  In the translated Zig code, `bar` has type `c_uint` since that is the return type of `@enumToInt(enum_foo.bar)`; it should be `c_int`

This bug was revealed (but not caused by) #8394, I believe the fix is to coerce enum redecls to the correct int type.

```C
enum foo { bar };
int main(void) {
   char c = bar;
}
```
```zig
pub const enum_foo = extern enum(c_uint) {
    bar,
    _,
};
pub const bar = @enumToInt(enum_foo.bar);
pub export fn main() c_int {
    var c: u8 = @bitCast(u8, @truncate(i8, bar));
    return 0;
}
```
```
./enums.zig:8:44: error: expected signed integer type, found 'c_uint'
    var c: u8 = @bitCast(u8, @truncate(i8, bar));
                                           ^
./enums.zig:8:30: note: referenced here
    var c: u8 = @bitCast(u8, @truncate(i8, bar));
                             ^
```

ehaas:
If the enum type cannot be translated for some reason, omit it. In that
case the previous behavior will occur - the enum constant's type will be
the enum's tag type.

Fixes #9153

### DIFF
--- a/lib/std/zig/c_translation.zig
+++ b/lib/std/zig/c_translation.zig
@@ -11,19 +11,13 @@ const mem = std.mem;
 
 /// Given a type and value, cast the value to the type as c would.
 pub fn cast(comptime DestType: type, target: anytype) DestType {
-    // this function should behave like transCCast in translate-c, except it's for macros and enums
+    // this function should behave like transCCast in translate-c, except it's for macros
     const SourceType = @TypeOf(target);
     switch (@typeInfo(DestType)) {
         .Fn, .Pointer => return castToPtr(DestType, SourceType, target),
         .Optional => |dest_opt| {
             if (@typeInfo(dest_opt.child) == .Pointer or @typeInfo(dest_opt.child) == .Fn) {
                 return castToPtr(DestType, SourceType, target);
-            }
-        },
-        .Enum => |enum_type| {
-            if (@typeInfo(SourceType) == .Int or @typeInfo(SourceType) == .ComptimeInt) {
-                const intermediate = cast(enum_type.tag_type, target);
-                return @intToEnum(DestType, intermediate);
             }
         },
         .Int => {
@@ -35,9 +29,6 @@ pub fn cast(comptime DestType: type, target: anytype) DestType {
                     if (@typeInfo(opt.child) == .Pointer) {
                         return castInt(DestType, @ptrToInt(target));
                     }
-                },
-                .Enum => {
-                    return castInt(DestType, @enumToInt(target));
                 },
                 .Int => {
                     return castInt(DestType, target);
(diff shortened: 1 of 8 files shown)


ehaas:
This fixes the immediate issue in #9153 but does not completely remove the use of `extern enum` as discussed here https://github.com/ziglang/zig/issues/2115#issuecomment-827968279

If desired, I could address that in this PR as well.

Vexu:
> If desired, I could address that in this PR as well.

I think that would be preferable, no reason to fix something that is going to be removed.

ehaas:
I believe this also fixes #5656

kmeaw:
Is it still possible to get former behavior? I am trying to convert integer values back to symbolic names.

InKryption:
@kmeaw C enums are translated as individual declarations now, so no, not really possible achieve the old behavior.

kmeaw:
It is possible to iterate over these declarations to build (either in run-time or in compile-time) a map of keys-to-values?

InKryption:
@kmeaw Well, it's possible to do so, though not very practical, as you'd have to use `std.meta.declarations()`, which would most likely result in referencing declarations replaced by `@compileError`s; that's not even to mention how slow that would make compilation as the file gets bigger.

One thing to consider is whether it might be beneficial to make a PR to make translate-c put enums into a namespace, though that could very likely result in some extra complexity to handle the case of when those enums are referenced by other translated code.

As it stands, you're probably better off just manually creating an enum based off of the declarations.

ehaas:
> Is it still possible to get former behavior? I am trying to convert integer values back to symbolic names.
> 
> `const key = @tagName(@intToEnum(c.my_enum, 123));`

The reason translate-c can't do this in general is that C enumerator values are not required to be unique within an enum declaration, e.g. the following is valid C:

```c
enum E {
    Foo = 1,
    Bar = 1,
};
```

There are other issues with trying to map C enums onto Zig enums, but I don't think they're relevant for your use case. I would also guess that 99% of real-world C code does use unique enumerator values, so perhaps there would be interest in extending translate-c to keep track of the names within an enum and create a non-exhaustive enum if they're unique.

kmeaw:
@InKryption, thank you for your suggestions. I had attempted to iterate over `@typeInfo(c).Struct.decls` but have hit some limits - `comptime { for {} }` crashes the compiler, `inline for {}` stops on unions I don't want to process. `inline for { comptime { if {} }}` fails because the declaration name is not a constant (but a loop variable), `inline for { comptime { if {} }}` exceeds 1000 backward branches. So I suppose the only option I have is taking a reversal of this commit and adding it to bring C enums back but in a separate struct namespace (`enum`?).

Issue #8728, opened 2021-05-09
remove the ZIG_SYSTEM_LINKER_HACK feature


andrewrk:
`ZIG_SYSTEM_LINKER_HACK` is an environment variable that Zig inspects on macOS when building natively, and will execute the system linker, ld64, as a child process, rather than using LLD or self-hosted linker.

https://github.com/ziglang/zig/blob/2d119677344c4ca6a2700b8de0a24215b9a37c06/src/Compilation.zig#L764

Don't get comfortable with it. This is a temporary workaround until our Mach-O linking code is sufficient to justify removing it.

Related:  #8727

This issue blocks Zig 1.0.

PR #9171, opened 2021-06-19
zld: add TAPI parser for linkers, ship libSystem.B.tbd with Zig, and parse tbd stubs on macOS

### LINKED ISSUE #8935: zig ld: ability to link tbds on macOS
Currently, `zig ld` can successfully link object files `.o`, static libraries `.a`, and (limited to no re-exports) dynamic libraries `.dylib`. However, starting with macOS 11 aka BigSur, no actual system dynamic libraries are available for browsing on the filesystem (well, they are available _somewhere_ in the `dyld`'s global cache). Instead, the linker is supposed to use the so-called text-based definition files `.tbd`s which are compact representation of the symbol tables found within the actual dynamic library and its re-exports. `zig ld` cannot link `tbd`s yet and they are required if anyone wants to build stage2 with Homebrew's Zig and/or LLVM since LLVM is built with system `libxml2` (which is only available as a `tbd` on recent macOS).

kubkon:
Closes #8935 

* add basic TAPI parser that's based on https://github.com/kubkon/zig-yaml/commit/c3eae1e40a02aedd44ad1171e5c8b259896cbda0
* ship `libSystem.B.tbd` with Zig - currently, ~~with this first draft of the PR,~~ the file resides in `lib/libc/darwin/libSystem.B.tbd`
* `zig ld` now gains ability to parse `tbd`s in addition to dylibs which means we can now link against macOS' system libs (frameworks are to follow soon-ish)
* add basic handler for the `___dso_handle` symbol - as far as I could work out, this symbol is expected to be "generated"/filled-in by the linker on macOS so as a temporary solution until I work out precisely what its purpose is, we add a dummy self-referential proxy (pointing to the executable itself) which resolves all references to `___dso_handle`; some links on the topic: [link1](http://llvm.1065342.n5.nabble.com/llvm-dev-Providing-dso-handle-in-LLVM-tp109516p109551.html), [link2](https://reviews.llvm.org/D97007#change-GmbFgFPVebk7)
* defer creating and referencing symbols from dylibs/stubs until link time - we defer creating and referencing symbols from dylibs/stubs until link time when a symbol is actually referenced and this is when we add a load command for the referenced dylib/stub. The only exception to this rule is `libSystem.B.tbd` which always gets a load command since it's implicitly always referenced.

---

~~I am purposefully creating this PR early as a draft so that interested parties can have a look and perhaps test it out in the wild while I clean it up and make it more, well, "presentable".~~

Oh, and PS @andrewrk, we're one, big step closer to resolving #8728.

This PR also makes it finally possible to link against Homebrew's LLVM and successfully build stage2, i.e., this now produces a valid `zig` stage2 binary:

```
> mkdir build && cd build
> cmake .. -DCMAKE_PREFIX_PATH=$(brew --prefix llvm) -DZIG_STATIC_LLVM=on
...
> make install
...
> cd ..
> build/bin/zig build -Denable-llvm
> zig-out/bin/zig
info: Usage: zig [command] [options]

Commands:

  build            Build project from build.zig
  init-exe         Initialize a `zig build` application in the cwd
  init-lib         Initialize a `zig build` library in the cwd

  ast-check        Look for simple compile errors in any set of files
  build-exe        Create executable from source or object files
  build-lib        Create library from source or object files
  build-obj        Create object from source or object files
  fmt              Reformat Zig source into canonical form
  run              Create executable and run immediately
  test             Create and run a test build
  translate-c      Convert C code to Zig code

  ar               Use Zig as a drop-in archiver
  cc               Use Zig as a drop-in C compiler
  c++              Use Zig as a drop-in C++ compiler
  dlltool          Use Zig as a drop-in dlltool.exe
  lib              Use Zig as a drop-in lib.exe
  ranlib           Use Zig as a drop-in ranlib

  env              Print lib path, std path, cache directory, and version
  help             Print this help and exit
  libc             Display native libc paths file or validate one
  targets          List available compilation targets
  version          Print version number and exit
  zen              Print Zen of Zig and exit

General Options:

  -h, --help       Print command-specific usage

Debug Commands:

  changelist       Compute mappings from old ZIR to new ZIR

error: expected command argument
```


src/Compilation.zig
```diff
@@ -916,6 +916,20 @@ pub fn create(gpa: *Allocator, options: InitOptions) !*Compilation {
             }
         };
 
+        const libc_stub_path: ?[]const u8 = if (options.target.isDarwin()) libc_stub: {
+            // TODO consider other platforms than Darwin which require linking against libc here.
+            const needs_libc_stub: bool = switch (options.output_mode) {
+                .Obj => false,
+                .Lib => if (options.link_mode) |mode| mode == .Dynamic else false,
+                .Exe => true,
+            };
```
andrewrk, line 925: We have `is_exe_or_dyn_lib` in scope here.

src/Compilation.zig
```diff
@@ -1288,6 +1302,7 @@ pub fn create(gpa: *Allocator, options: InitOptions) !*Compilation {
             .use_lld = use_lld,
             .use_llvm = use_llvm,
             .system_linker_hack = darwin_options.system_linker_hack,
+            .libc_stub_path = libc_stub_path,
```
andrewrk, line 1305: Why does there need to be any code here in Compilation.zig? There is no user input configurability here, so the link code could just do this logic. No need to add a field to link.Options.

andrewrk (changes requested):
I have one small change request, but overall, wow, amazing work. After that this looks ready to merge to me.

The placement of lib/libc/libSystem.B.tbd looks good to me :+1:

src/Compilation.zig
kubkon: Excellent point. I think I even had the logic in the linker proper before, but then couldn't decide which approach would be best and ended up with this. Anyhow, now the logic is moved back into the linker!

PR #9175, opened 2021-06-20
std.Thread enhancements

kprotty:
* thread spawning api similar to the `@call(.{}, function, args_tuple)` builtin
* spawn options (first arg) with customizable thread stack size
* renames `wait()` to `join()`
* renames `handle()` to `getHandle()` for consistency
* adds ability to `detach()` thread, allowing it to clean up resources when it exits instead of calling `join()`
* ~~removes `getCurrentThreadId()` in favor of `getCurrentId()` (seemed to only be used in debug once)~~
* moves `getCurrentThreadId()` into `getCurrentId()`


LemonBoy:
> removes getCurrentThreadId() in favor of getCurrentId() (seemed to only be used in debug once)

This is not ok, the result of `pthread_self` is different from the result returned by `getCurrentThreadId` while the former is a pointer while the latter is a numeric ID that's useful for logging/debug purposes.

kprotty:
@LemonBoy Couldn't you use the `pthread_self()` pointer for logging purposes as well? For the reverse, when would you want `pthread_self()` from the abstraction instead of just calling it directly?

daurnimator:
@kprotty the thread id shows up in e.g. `ps` output. vs `pthread_self` is only really known by the process itself.

kprotty:
@LemonBoy do you think the drone CI fail here due to a musl-pthread/llvm issue? Looks like the `pthread_create()` thread isn't observing changes to values after a `pthread_join()`, with `pthread_detach()` seemingly just as wonky.

LemonBoy:
You're barking at the wrong tree heh, the faulty code is Zig's thread creation routine that's missing some more synchronization:
* The process is spliced with `clone`
* `CLONE_CHILD_SETTID` is set and `child_tid` is initialized...
* But is it initialized? Not in the parent thread!

              Store the  child  thread  ID  at  the  location  pointed  to  by
              child_tid  (clone())  or  cl_args.child_tid  (clone3())  in  the
              child's memory.  The store operation completes before the  clone
              call  returns control to user space in the child process.  (Note
              that the store operation may not have completed before the clone
              call  returns  in  the parent process, which will be relevant if
              the CLONE_VM flag is also employed.)
* `join` tries to start a spinloop/futex wait on that variable but quits upon seeing the zero (that given the presence of `CLONE_CHILD_CLEAR` means the child process terminated)
* Once `join` terminates the whole thread memory is unmapped, making the child process crash and burn

So, if you're lucky enough to get all the timings right (or wrong) you end up not seeing the effects of `testIncrementNotify` and if you're unlucky you get a nasty segfault.

kprotty:
> But is it initialized? Not in the parent thread!

Dang, and I tried to use `CLONE_CHILD_SETTID` to avoid that after noticing something similar. Missed the docs again. Thx

kprotty:
Thank you, CI deities

PR #9052, opened 2021-06-09
cc,wasi: support WASI reactors via -mexec-model flag.

mathetake:
Based on this patch to LLVM (https://reviews.llvm.org/D62922). This partially resolves (only for zig cc) #6757

PR #9178, opened 2021-06-20
Add support for WASI reactor in pure Zig-exe.

### LINKED ISSUE #6757: WASI reactors

mathetake:
This PR adds support for WASI reactor in pure Zig WASI targets. Notably this enables users to produce WASI reactor binary with `build-exe` command with `-mexec-model=reactor` just like `zig cc`, `zig c++`, and clang.

With `-mexec-model=reactor` flag, Zig 1) exports `_initialize` symbol which calls the user defined `pub fn main`, 2) does not define `_start`, and 3) does not execute `proc_exit` after `main`.

Please refer to https://github.com/WebAssembly/WASI/blob/main/design/application-abi.md for the ABI in detail.

Finally resolves #6757 together with #9052.

cc @kubkon

### DIFF
--- a/lib/std/start.zig
+++ b/lib/std/start.zig
@@ -16,7 +16,7 @@ const native_os = builtin.os.tag;
 
 var argc_argv_ptr: [*]usize = undefined;
 
-const start_sym_name = if (native_arch.isMIPS()) "__start" else if (builtin.is_wasi_reactor_exec_model) "_initialize" else "_start";
+const start_sym_name = if (native_arch.isMIPS()) "__start" else if (builtin.wasi_exec_model == .reactor) "_initialize" else "_start";
 
 comptime {
     // No matter what, we import the root file, so that any export, test, comptime
@@ -142,10 +142,9 @@ fn wasm_start() callconv(.C) void {
             _ = @call(.{ .modifier = .always_inline }, callMain, .{});
         },
         .wasi => {
-            if (builtin.is_wasi_reactor_exec_model) {
-                _ = @call(.{ .modifier = .always_inline }, callMain, .{});
-            } else {
-                std.os.wasi.proc_exit(@call(.{ .modifier = .always_inline }, callMain, .{}));
+            switch (builtin.wasi_exec_model) {
+                .reactor => _ = @call(.{ .modifier = .always_inline }, callMain, .{}),
+                .command => std.os.wasi.proc_exit(@call(.{ .modifier = .always_inline }, callMain, .{})),
             }
         },
         else => @compileError("unsupported OS"),
--- a/lib/std/start.zig
+++ b/lib/std/start.zig
@@ -16,7 +16,7 @@ const native_os = builtin.os.tag;
 
 var argc_argv_ptr: [*]usize = undefined;
 
-const start_sym_name = if (native_arch.isMIPS()) "__start" else if (builtin.wasi_exec_model == .reactor) "_initialize" else "_start";
+const start_sym_name = if (native_arch.isMIPS()) "__start" else "_start";
 
 comptime {
     // No matter what, we import the root file, so that any export, test, comptime
@@ -65,7 +65,8 @@ comptime {
             } else if (native_os == .uefi) {
                 if (!@hasDecl(root, "EfiMain")) @export(EfiMain, .{ .name = "EfiMain" });
             } else if (native_arch.isWasm()) {
-                if (!@hasDecl(root, start_sym_name)) @export(wasm_start, .{ .name = start_sym_name });
+                const wasm_start_sym = if (builtin.wasi_exec_model == .reactor) "_initialize" else "_start";
+                if (!@hasDecl(root, start_sym_name)) @export(wasm_start, .{ .name = wasm_start_sym });
             } else if (native_os != .other and native_os != .freestanding) {
                 if (!@hasDecl(root, start_sym_name)) @export(_start, .{ .name = start_sym_name });
             }
--- a/lib/std/start.zig
+++ b/lib/std/start.zig
@@ -66,7 +66,7 @@ comptime {
                 if (!@hasDecl(root, "EfiMain")) @export(EfiMain, .{ .name = "EfiMain" });
             } else if (native_arch.isWasm()) {
                 const wasm_start_sym = if (builtin.wasi_exec_model == .reactor) "_initialize" else "_start";
-                if (!@hasDecl(root, start_sym_name)) @export(wasm_start, .{ .name = wasm_start_sym });
+                if (!@hasDecl(root, wasm_start_sym)) @export(wasm_start, .{ .name = wasm_start_sym });
             } else if (native_os != .other and native_os != .freestanding) {
                 if (!@hasDecl(root, start_sym_name)) @export(_start, .{ .name = start_sym_name });
             }
--- a/lib/std/start.zig
+++ b/lib/std/start.zig
@@ -16,7 +16,7 @@ const native_os = builtin.os.tag;
 
 var argc_argv_ptr: [*]usize = undefined;
 
-const start_sym_name = if (native_arch.isMIPS()) "__start" else "_start";
+const start_sym_name = if (native_arch.isMIPS()) "__start" else if (builtin.is_wasi_reactor_exec_model) "_initialize" else "_start";
 
 comptime {
     // No matter what, we import the root file, so that any export, test, comptime
@@ -64,8 +64,8 @@ comptime {
                 }
             } else if (native_os == .uefi) {
                 if (!@hasDecl(root, "EfiMain")) @export(EfiMain, .{ .name = "EfiMain" });
-            } else if (native_arch.isWasm() and native_os == .freestanding) {
-                if (!@hasDecl(root, start_sym_name)) @export(wasm_freestanding_start, .{ .name = start_sym_name });
+            } else if (native_arch.isWasm()) {
+                if (!@hasDecl(root, start_sym_name)) @export(wasm_start, .{ .name = start_sym_name });
             } else if (native_os != .other and native_os != .freestanding) {
                 if (!@hasDecl(root, start_sym_name)) @export(_start, .{ .name = start_sym_name });
             }
@@ -134,10 +134,22 @@ fn _DllMainCRTStartup(
     return std.os.windows.TRUE;
 }
 
-fn wasm_freestanding_start() callconv(.C) void {
-    // This is marked inline because for some reason LLVM in release mode fails to inline it,
+fn wasm_start() callconv(.C) void {
+    // The entrypoint is marked inline because for some reason LLVM in release mode fails to inline it,
     // and we want fewer call frames in stack traces.
-    _ = @call(.{ .modifier = .always_inline }, callMain, .{});
+    switch (native_os) {
+        .freestanding => {
+            _ = @call(.{ .modifier = .always_inline }, callMain, .{});
+        },
+        .wasi => {
+            if (builtin.is_wasi_reactor_exec_model) {
+                _ = @call(.{ .modifier = .always_inline }, callMain, .{});
+            } else {
+                std.os.wasi.proc_exit(@call(.{ .modifier = .always_inline }, callMain, .{}));
+            }
+        },
+        else => @compileError("unsupported OS"),
+    }
 }
 
 fn EfiMain(handle: uefi.Handle, system_table: *uefi.tables.SystemTable) callconv(.C) usize {
@@ -163,12 +175,6 @@ fn EfiMain(handle: uefi.Handle, system_table: *uefi.tables.SystemTable) callconv
 }
 
 fn _start() callconv(.Naked) noreturn {
-    if (native_os == .wasi) {
-        // This is marked inline because for some reason LLVM in release mode fails to inline it,
-        // and we want fewer call frames in stack traces.
-        std.os.wasi.proc_exit(@call(.{ .modifier = .always_inline }, callMain, .{}));
-    }
-
     switch (native_arch) {
         .x86_64 => {
             argc_argv_ptr = asm volatile (
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -724,7 +724,7 @@ pub const InitOptions = struct {
     test_name_prefix: ?[]const u8 = null,
     subsystem: ?std.Target.SubSystem = null,
     /// WASI-only. Type of WASI execution model ("command" or "reactor").
-    wasi_exec_model: ?[]const u8 = null,
+    wasi_exec_model: ?std.builtin.WasiExecModel = null,
 };
 
 fn addPackageTableToCacheHash(
@@ -3638,10 +3638,14 @@ pub fn generateBuiltinZigSource(comp: *Compilation, allocator: *Allocator) Alloc
         std.zig.fmtId(@tagName(comp.bin_file.options.machine_code_model)),
     });
 
+    const wasi_exec_model_fmt = if (comp.bin_file.options.wasi_exec_model) |model|
+        std.zig.fmtId(@tagName(model))
+    else
+        std.zig.fmtId(@tagName(std.builtin.WasiExecModel.command));
     try buffer.writer().print(
-        \\pub const is_wasi_reactor_exec_model = {};
+        \\pub const wasi_exec_model = std.builtin.WasiExecModel.{};
         \\
-    , .{if (comp.bin_file.options.wasi_exec_model) |model| std.mem.eql(u8, model, "reactor") else false});
+    , .{wasi_exec_model_fmt});
 
     if (comp.bin_file.options.is_test) {
         try buffer.appendSlice(
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1449,7 +1449,7 @@ pub fn create(gpa: *Allocator, options: InitOptions) !*Compilation {
                 });
             }
             comp.work_queue.writeAssumeCapacity(&[_]Job{
-                .{ .wasi_libc_crt_file = wasi_libc.getExecModelCrtFIle(wasi_exec_model) },
+                .{ .wasi_libc_crt_file = wasi_libc.execModelCrtFile(wasi_exec_model) },
                 .{ .wasi_libc_crt_file = .libc_a },
             });
         }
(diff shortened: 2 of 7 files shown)


src/main.zig
```diff
@@ -616,7 +617,7 @@ fn buildOutputType(
     var subsystem: ?std.Target.SubSystem = null;
     var major_subsystem_version: ?u32 = null;
     var minor_subsystem_version: ?u32 = null;
-    var wasi_exec_model: ?wasi_libc.CRTFile = null;
+    var wasi_exec_model: ?[]const u8 = null;
```
kubkon, line 620: Hmm, I'd rather avoid passing strings instead of enums. How about we create an intermediate enum like you did before which then we can map to `wasi_libc.CRTFile` where required, and we can switch on in the linker rather than compare strings?

src/wasi_libc.zig
```diff
@@ -45,8 +45,20 @@ pub fn emulatedLibCRFileLibName(crt_file: CRTFile) []const u8 {
     };
 }
 
-pub fn crtFileFullName(crt_file: CRTFile) []const u8 {
-    return switch (crt_file) {
+pub fn getExecModelCRTFIle(wasi_exec_model: ?[]const u8) CRTFile {
```
kubkon, line 48: See my comment above: I really would avoid strings at all cost especially since WASI exec model has well defined variants.

lib/std/start.zig
```diff
@@ -16,7 +16,7 @@ const native_os = builtin.os.tag;
 
 var argc_argv_ptr: [*]usize = undefined;
 
-const start_sym_name = if (native_arch.isMIPS()) "__start" else "_start";
+const start_sym_name = if (native_arch.isMIPS()) "__start" else if (builtin.is_wasi_reactor_exec_model) "_initialize" else "_start";
```
kubkon, line 19: Hmm, here's a thought. Maybe we should augment the builtins and put the exec model enum there much like this is the case with `GlobalLinkage`, etc.?

kubkon (changes requested):
Thanks @mathetake! I've got a few comments I'd like to see addressed before going forward with this PR.

src/main.zig
mathetake: yeah that's what actually I was thinking about first. Thanks for the suggestion! will fix accordingly.

src/Compilation.zig
```diff
@@ -3638,6 +3638,15 @@ pub fn generateBuiltinZigSource(comp: *Compilation, allocator: *Allocator) Alloc
...
     });
 
+    const wasi_exec_model_fmt = if (comp.bin_file.options.wasi_exec_model) |model|
+        std.zig.fmtId(@tagName(model))
+    else
+        std.zig.fmtId(@tagName(std.builtin.WasiExecModel.command));
+    try buffer.writer().print(
+        \\pub const wasi_exec_model = std.builtin.WasiExecModel.{};
+        \\
+    , .{wasi_exec_model_fmt});
```
mathetake, line 3648: FYI: If I put this into a single statement, 

```zig
    try buffer.writer().print(
        \\pub const wasi_exec_model = std.builtin.WasiExecModel.{};
        \\
    , .{if (comp.bin_file.options.wasi_exec_model) |model|
        std.zig.fmtId(@tagName(model))
    else
        std.zig.fmtId(@tagName(std.builtin.WasiExecModel.command))});
```

I got some weird error around LLVM: 
```
make install
[ 77%] Built target embedded_softfloat
[ 78%] Built target opt_c_util
[ 83%] Built target zigcpp
[ 97%] Built target zigstage1
[ 98%] Built target zig0
[ 99%] Building self-hosted component /home/mathetake/zig/build/zig1.o
broken LLVM module found: Instruction does not dominate all uses!
  %369 = getelementptr inbounds %"Compilation.struct:3644:8", %"Compilation.struct:3644:8"* %39, i32 0, i32 0, !dbg !75434
  call fastcc void @std.zig.fmt.fmtId(%"std.fmt.Formatter(std.zig.fmt.formatId)"* sret(%"std.fmt.Formatter(std.zig.fmt.formatId)") %369, %"[]u8"* @13528), !dbg !75435
Instruction does not dominate all uses!
  %369 = getelementptr inbounds %"Compilation.struct:3644:8", %"Compilation.struct:3644:8"* %39, i32 0, i32 0, !dbg !75434
  %372 = bitcast %"std.fmt.Formatter(std.zig.fmt.formatId)"* %369 to i8*, !dbg !75435
Instruction does not dominate all uses!
  %369 = getelementptr inbounds %"Compilation.struct:3644:8", %"Compilation.struct:3644:8"* %39, i32 0, i32 0, !dbg !75434
  %373 = bitcast %"std.fmt.Formatter(std.zig.fmt.formatId)"* %369 to i8*, !dbg !75435

This is a bug in the Zig compiler.
Aborted (core dumped)
make[2]: *** [CMakeFiles/zig.dir/build.make:328: zig1.o] Error 134
make[1]: *** [CMakeFiles/Makefile2:148: CMakeFiles/zig.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```

lib/std/start.zig
mathetake: Thanks! I believe https://github.com/ziglang/zig/pull/9178/commits/6769659dedde8b53999fda15b0125639cc451c23 addressed this comment.

src/wasi_libc.zig
mathetake: I believe this is fixed! https://github.com/ziglang/zig/pull/9178/commits/6769659dedde8b53999fda15b0125639cc451c23 addressed this comment.

src/main.zig
mathetake: PTAL at https://github.com/ziglang/zig/pull/9178/commits/6769659dedde8b53999fda15b0125639cc451c23

mathetake:
Thanks @kubkon for a quick look:) I hope https://github.com/ziglang/zig/pull/9178/commits/6769659dedde8b53999fda15b0125639cc451c23 addressed your comments! PTAL!

src/Compilation.zig
kubkon: I believe this is because it's not enough to just add an enum to `std.builtin`. You need to make more tweaks to stage1 in C++. For starters, we need to have the exact same enum shared between both stage1 and stage2 which means you need add a corresponding enum to `src/stage1/stage1.h`. Then, you'll need to tweak logic for actually marshalling the data between stage1 and stage2 - have a look how this is for `CodeModel` or some other builtin structs.

If you get stuck, I'll be happy to help you out but will probably take a look at this in more detail tomorrow. Alternatively, we can also get @andrewrk involved if he's got a minute.

mathetake: Ah I see. Thanks for clarifying!

mathetake: so yeah I found that this also applies to other existing builtin types: 

```
const std = @import("std");

pub fn main() void {
    std.debug.print("test {}", .{if (@rem(std.time.timestamp(), 2) == 0)
        std.zig.fmtId(@tagName(std.builtin.CodeModel.tiny))
    else
        std.zig.fmtId(@tagName(std.builtin.CodeModel.default))});
}
```

HEAD and 0.8.0 fails to compile this code with the same error like this:

```
zig build-exe main.zig
broken LLVM module found: Instruction does not dominate all uses!d.debug.ModuleDebug...
  %2 = getelementptr inbounds %"struct:4:33", %"struct:4:33"* %0, i32 0, i32 0, !dbg !16731
  call fastcc void @std.zig.fmt.fmtId(%"std.fmt.Formatter(std.zig.fmt.formatId)"* sret(%"std.fmt.Formatter(std.zig.fmt.formatId)") %2, %"[]u8.18"* @390), !dbg !16733
Instruction does not dominate all uses!
  %2 = getelementptr inbounds %"struct:4:33", %"struct:4:33"* %0, i32 0, i32 0, !dbg !16731
  %5 = bitcast %"std.fmt.Formatter(std.zig.fmt.formatId)"* %2 to i8*, !dbg !16733
Instruction does not dominate all uses!
  %2 = getelementptr inbounds %"struct:4:33", %"struct:4:33"* %0, i32 0, i32 0, !dbg !16731
  %6 = bitcast %"std.fmt.Formatter(std.zig.fmt.formatId)"* %2 to i8*, !dbg !16733

This is a bug in the Zig compiler.thread 3208559 panic:
zsh: abort (core dumped)  zig build-exe main.zig
```

and the same, if we split the line to two statements. the error disappears.

mathetake: fyi: I tweaked the C++ and around stage1.zig and Compilation.zig following `CodeModel`, but the error has never gone.

mathetake: Luckily the code with split statements just works and successfully produced Wasi Reactors!

kubkon (approved):
Looks great to me, thanks @mathetake! I just need to double check with @andrewrk that we don't need to add matching logic to the C++ codebase for setting `WasiExecModel` builtin.

mathetake:
@andrewrk @kubkon I would appreciate it if you have any chance to take a look and upstream this! Thanks a lot:)

andrewrk:
Ah I'm sorry for taking so long to get to this! Looking now.

lib/std/start.zig
```diff
@@ -16,7 +16,7 @@ const native_os = builtin.os.tag;
 
 var argc_argv_ptr: [*]usize = undefined;
 
-const start_sym_name = if (native_arch.isMIPS()) "__start" else "_start";
+const start_sym_name = if (native_arch.isMIPS()) "__start" else if (builtin.wasi_exec_model == .reactor) "_initialize" else "_start";
```
andrewrk, line 19: `builtin.wasi_exec_model` should only be accessed if the target OS is WASI. Perhaps this line should be left alone and logic should be added below on line 68?

src/wasi_libc.zig
```diff
@@ -45,8 +45,18 @@ pub fn emulatedLibCRFileLibName(crt_file: CRTFile) []const u8 {
...
+    return if (wasi_exec_model) |model|
+        switch (model) {
+            .reactor => CRTFile.crt1_reactor_o,
+            .command => CRTFile.crt1_command_o,
+        }
+    else
+        CRTFile.crt1_o;
+}
+
+pub fn execModelCRTFileFullName(wasi_exec_model: ?std.builtin.WasiExecModel) []const u8 {
```
andrewrk, line 58:
```suggestion
pub fn execModelCrtFileFullName(wasi_exec_model: ?std.builtin.WasiExecModel) []const u8 {
```

src/wasi_libc.zig
```diff
@@ -45,8 +45,18 @@ pub fn emulatedLibCRFileLibName(crt_file: CRTFile) []const u8 {
     };
 }
 
-pub fn crtFileFullName(crt_file: CRTFile) []const u8 {
-    return switch (crt_file) {
+pub fn getExecModelCRTFIle(wasi_exec_model: ?std.builtin.WasiExecModel) CRTFile {
```
andrewrk, line 48:
```suggestion
pub fn execModelCrtFile(wasi_exec_model: ?std.builtin.WasiExecModel) CRTFile {
```

my mistake for calling it CRTFile instead of CrtFile. I will fix that later.

src/Compilation.zig
```diff
@@ -3638,6 +3638,15 @@ pub fn generateBuiltinZigSource(comp: *Compilation, allocator: *Allocator) Alloc
         std.zig.fmtId(@tagName(comp.bin_file.options.machine_code_model)),
     });
 
+    const wasi_exec_model_fmt = if (comp.bin_file.options.wasi_exec_model) |model|
+        std.zig.fmtId(@tagName(model))
+    else
+        std.zig.fmtId(@tagName(std.builtin.WasiExecModel.command));
+    try buffer.writer().print(
+        \\pub const wasi_exec_model = std.builtin.WasiExecModel.{};
```
andrewrk, line 3646: Rather than populating the value arbitrarily with "command" when the OS is not WASI, this declaration should be entirely omitted when the OS is not WASI.

src/link.zig
```diff
@@ -119,7 +119,7 @@ pub const Options = struct {
     libc_installation: ?*const LibCInstallation,
 
     /// WASI-only. Type of WASI execution model ("command" or "reactor").
-    wasi_exec_model: ?wasi_libc.CRTFile = null,
+    wasi_exec_model: ?std.builtin.WasiExecModel,
```
andrewrk, line 122: Rather than having this be optional, I suggest that it should be non-optional, and set to `undefined` when the target OS is not WASI.

src/main.zig
```diff
@@ -1065,6 +1066,13 @@ fn buildOutputType(
                         mem.startsWith(u8, arg, "-I"))
                     {
                         try clang_argv.append(arg);
+                    } else if (mem.startsWith(u8, arg, "-mexec-model=")) {
+                        const value = arg["-mexec-model=".len..];
+                        if (std.mem.eql(u8, value, "reactor")) {
+                            wasi_exec_model = .reactor;
+                        } else if (std.mem.eql(u8, value, "command")) {
+                            wasi_exec_model = .command;
+                        }
```
andrewrk, line 1075: Missing an error when a value other than "reactor" or "command" is specified. I suggest using `std.meta.stringToEnum` (look for other usages of that function nearby).

src/main.zig
```diff
@@ -1272,9 +1280,9 @@ fn buildOutputType(
                     .strip => strip = true,
                     .exec_model => {
                         if (std.mem.eql(u8, it.only_arg, "reactor")) {
-                            wasi_exec_model = .crt1_reactor_o;
+                            wasi_exec_model = .reactor;
                         } else if (std.mem.eql(u8, it.only_arg, "command")) {
-                            wasi_exec_model = .crt1_command_o;
+                            wasi_exec_model = .command;
                         }
```
andrewrk, line 1286: Same here, missing an error for bad exec model name.

src/wasi_libc.zig
```diff
@@ -45,8 +45,18 @@ pub fn emulatedLibCRFileLibName(crt_file: CRTFile) []const u8 {
...
+    return if (wasi_exec_model) |model|
+        switch (model) {
+            .reactor => CRTFile.crt1_reactor_o,
+            .command => CRTFile.crt1_command_o,
+        }
+    else
+        CRTFile.crt1_o;
```
andrewrk, line 55: Based on my research it looks like every WASI program is either a command or a reactor. Is it meaningful for a program to be neither? Why is crt1_o even an option?

src/Compilation.zig
```diff
@@ -724,7 +724,7 @@ pub const InitOptions = struct {
     test_name_prefix: ?[]const u8 = null,
     subsystem: ?std.Target.SubSystem = null,
     /// WASI-only. Type of WASI execution model ("command" or "reactor").
-    wasi_exec_model: ?wasi_libc.CRTFile = null,
```
andrewrk, line 727: Inside the create function of Compilation, it should resolve this `?WasiExecModel` into a `WasiExecModel` - that is - it should choose a default, and then the link Options struct should store only the non-optional WasiExecModel.

andrewrk (changes requested):
Thanks for the contribution!

I have some requested improvements.

mathetake:
Thank you so much @andrewrk, I hope I addressed your suggestion in https://github.com/ziglang/zig/pull/9178/commits/85b5503e3a3f39a37eeb4df790cfe7370433f1de, and now it looks much better:)

PTAL when you have time, thanks!

lib/std/start.zig
```diff
@@ -64,8 +64,9 @@ comptime {
                 }
             } else if (native_os == .uefi) {
                 if (!@hasDecl(root, "EfiMain")) @export(EfiMain, .{ .name = "EfiMain" });
-            } else if (native_arch.isWasm() and native_os == .freestanding) {
-                if (!@hasDecl(root, start_sym_name)) @export(wasm_freestanding_start, .{ .name = start_sym_name });
+            } else if (native_arch.isWasm()) {
+                const wasm_start_sym = if (builtin.wasi_exec_model == .reactor) "_initialize" else "_start";
+                if (!@hasDecl(root, start_sym_name)) @export(wasm_start, .{ .name = wasm_start_sym });
```
andrewrk, line 69:
```suggestion
                if (!@hasDecl(root, wasm_start_sym)) @export(wasm_start, .{ .name = wasm_start_sym });
```

src/Compilation.zig
```diff
@@ -3638,6 +3641,14 @@ pub fn generateBuiltinZigSource(comp: *Compilation, allocator: *Allocator) Alloc
         std.zig.fmtId(@tagName(comp.bin_file.options.machine_code_model)),
     });
 
+    if (comp.getTarget().os.tag == .wasi) {
```
andrewrk, line 3644:
```suggestion
    if (target.os.tag == .wasi) {
```

src/wasi_libc.zig
mathetake: yeah we should remove crt1_o in favor of crt1_command_o, where the latter lets wasm-ld call ctors rather than explicitly calling in crt.c. Also I believe this reduces the complexity here about null vs default stuff.

so I removed the support for crt1_o. cc @kubkon wdyt?

src/wasi_libc.zig
```diff
@@ -45,9 +45,15 @@ pub fn emulatedLibCRFileLibName(crt_file: CRTFile) []const u8 {
     };
 }
 
-pub fn crtFileFullName(crt_file: CRTFile) []const u8 {
-    return switch (crt_file) {
-        .crt1_o => "crt1.o",
+pub fn getExecModelCrtFIle(wasi_exec_model: std.builtin.WasiExecModel) CRTFile {
```
andrewrk, line 48:
```suggestion
pub fn execModelCrtFile(wasi_exec_model: std.builtin.WasiExecModel) CRTFile {
```

andrewrk (changes requested):
Thanks for making the requested changes - just a couple small things left and then this looks ready to merge to me :)

mathetake:
thanks for the quick review! addressed https://github.com/ziglang/zig/pull/9178/commits/943f6568a8c8e56b661b1db396d8b8ac89d3e9c3 here :)

andrewrk (approved):
Alright, this is ready to be merged, did you want to wait until @kubkon wakes up to discuss the question you asked? If not I'm happy to merge now.

mathetake:
I can discuss offline with @kubkon later, so I'm happy to merge now. I believe it's easy to bring crt1.o support even if we want later:)

PR #9179, opened 2021-06-20
Add support for linking against fat/universal dynamic libraries on macOS

### LINKED ISSUE #9169: zig ld: ability to link against universal binaries
Problem: Some libraries on macOS are distributed as universal binaries that might be the only option to link against - even when building for a single architecture.

Solution: Support linking against universal binaries

Universal binaries on macOS work by embedding the data for multiple libraries (one for each arch)  into a single file - with a "fat" header to provide the offsets for each of these. 

The format of this header is quite simple (a magic number along with the number of headers, followed by a list of headers containing the appropriate offsets). The difficulty here is making sure that zld doesn't assume one file == one library, and that we can then find the correct offset for any libraries we might want to link against for a given target. 

See the discussion in #8317 for more context.

moosichu:
This resolves #9169.

I've validated that this works by successfully linking with the vulkan dylibs on macOS, which are shipped as fat aarch64+x86_64 libraries.

### DIFF
--- a/lib/std/macho.zig
+++ b/lib/std/macho.zig
@@ -24,6 +24,19 @@ pub const mach_header_64 = extern struct {
     reserved: u32,
 };
 
+pub const fat_header = extern struct {
+    magic: u32,
+    nfat_arch: u32,
+};
+
+pub const fat_arch = extern struct {
+    cputype: cpu_type_t,
+    cpusubtype: cpu_subtype_t,
+    offset: u32,
+    size: u32,
+    @"align": u32,
+};
+
 pub const load_command = extern struct {
     cmd: u32,
     cmdsize: u32,
@@ -1039,6 +1052,20 @@ pub const MH_APP_EXTENSION_SAFE = 0x02000000;
 
 /// The external symbols listed in the nlist symbol table do not include all the symbols listed in the dyld info.
 pub const MH_NLIST_OUTOFSYNC_WITH_DYLDINFO = 0x04000000;
+
+// Constants for the flags field of the fat_header
+
+/// the fat magic number
+pub const FAT_MAGIC = 0xcafebabe;
+
+/// NXSwapLong(FAT_MAGIC)
+pub const FAT_CIGAM = 0xbebafeca;
+
+/// the 64-bit fat magic number
+pub const FAT_MAGIC_64 = 0xcafebabf;
+
+/// NXSwapLong(FAT_MAGIC_64)
+pub const FAT_CIGAM_64 = 0xbfbafeca;
 
 /// The flags field of a section structure is separated into two parts a section
 /// type and section attributes.  The section types are mutually exclusive (it
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -1539,6 +1539,34 @@ test "writeIntBig and writeIntLittle" {
     try testing.expect(eql(u8, buf2[0..], &[_]u8{ 0xfc, 0xff }));
 }
 
+/// Swap the byte order of all the members of the fields of a struct
+/// (Changing their endianess)
+pub fn bswapAllFields(comptime S: type, ptr: *S) void {
+    if (@typeInfo(S) != .Struct) @compileError("bswapAllFields expects a struct as the first argument");
+    inline for (std.meta.fields(S)) |f| {
+        @field(ptr, f.name) = @byteSwap(f.field_type, @field(ptr, f.name));
+    }
+}
+
+test "bswapAllFields" {
+    const T = extern struct {
+        f0: u8,
+        f1: u16,
+        f2: u32,
+    };
+    var s = T{
+        .f0 = 0x12,
+        .f1 = 0x1234,
+        .f2 = 0x12345678,
+    };
+    bswapAllFields(T, &s);
+    try std.testing.expectEqual(T{
+        .f0 = 0x12,
+        .f1 = 0x3412,
+        .f2 = 0x78563412,
+    }, s);
+}
+
 /// Returns an iterator that iterates over the slices of `buffer` that are not
 /// any of the bytes in `delimiter_bytes`.
 /// tokenize("   abc def    ghi  ", " ")
--- a/src/link/MachO/Dylib.zig
+++ b/src/link/MachO/Dylib.zig
@@ -1,6 +1,7 @@
 const Dylib = @This();
 
 const std = @import("std");
+const builtin = std.builtin;
 const assert = std.debug.assert;
 const fs = std.fs;
 const fmt = std.fmt;
@@ -8,6 +9,7 @@ const log = std.log.scoped(.dylib);
 const macho = std.macho;
 const math = std.math;
 const mem = std.mem;
+const native_endian = builtin.target.cpu.arch.endian();
 
 const Allocator = mem.Allocator;
 const Arch = std.Target.Cpu.Arch;
@@ -25,6 +27,10 @@ name: ?[]const u8 = null,
 syslibroot: ?[]const u8 = null,
 
 ordinal: ?u16 = null,
+
+// The actual dylib contents we care about linking with will be embedded at
+// an offset within a file if we are linking against a fat lib
+library_offset: u64 = 0,
 
 load_commands: std.ArrayListUnmanaged(LoadCommand) = .{},
 
@@ -205,8 +211,44 @@ pub fn closeFile(self: Dylib) void {
     }
 }
 
+fn decodeArch(cputype: macho.cpu_type_t) !std.Target.Cpu.Arch {
+    const arch: Arch = switch (cputype) {
+        macho.CPU_TYPE_ARM64 => .aarch64,
+        macho.CPU_TYPE_X86_64 => .x86_64,
+        else => {
+            return error.UnsupportedCpuArchitecture;
+        },
+    };
+    return arch;
+}
+
 pub fn parse(self: *Dylib) !void {
     log.debug("parsing shared library '{s}'", .{self.name.?});
+
+    self.library_offset = offset: {
+        const fat_header = try readFatStruct(self.file.?.reader(), macho.fat_header);
+        if (fat_header.magic != macho.FAT_MAGIC) break :offset 0;
+
+        var fat_arch_index: u32 = 0;
+        while (fat_arch_index < fat_header.nfat_arch) : (fat_arch_index += 1) {
+            const fat_arch = try readFatStruct(self.file.?.reader(), macho.fat_arch);
+            // If we come across an architecture that we do not know how to handle, that's
+            // fine because we can keep looking for one that might match.
+            const lib_arch = decodeArch(fat_arch.cputype) catch |err| switch (err) {
+                error.UnsupportedCpuArchitecture => continue,
+                else => |e| return e,
+            };
+            if (lib_arch == self.arch.?) {
+                // We have found a matching architecture!
+                break :offset fat_arch.offset;
+            }
+        } else {
+            log.err("Could not find matching cpu architecture in fat library: expected {s}", .{self.arch.?});
+            return error.MismatchedCpuArchitecture;
+        }
+    };
+
+    try self.file.?.seekTo(self.library_offset);
 
     var reader = self.file.?.reader();
     self.header = try reader.readStruct(macho.mach_header_64);
@@ -216,14 +258,14 @@ pub fn parse(self: *Dylib) !void {
         return error.NotDylib;
     }
 
-    const this_arch: std.Target.Cpu.Arch = switch (self.header.?.cputype) {
-        macho.CPU_TYPE_ARM64 => .aarch64,
-        macho.CPU_TYPE_X86_64 => .x86_64,
-        else => |value| {
-            log.err("unsupported cpu architecture 0x{x}", .{value});
-            return error.UnsupportedCpuArchitecture;
+    const this_arch: Arch = decodeArch(self.header.?.cputype) catch |err| switch (err) {
+        error.UnsupportedCpuArchitecture => |e| {
+            log.err("unsupported cpu architecture 0x{x}", .{self.header.?.cputype});
+            return e;
         },
+        else => |e| return e,
     };
+
     if (this_arch != self.arch.?) {
         log.err("mismatched cpu architecture: expected {s}, found {s}", .{ self.arch.?, this_arch });
         return error.MismatchedCpuArchitecture;
@@ -232,6 +274,16 @@ pub fn parse(self: *Dylib) !void {
     try self.readLoadCommands(reader);
     try self.parseId();
     try self.parseSymbols();
+}
+
+fn readFatStruct(reader: anytype, comptime T: type) !T {
+    // Fat structures (fat_header & fat_arch) are always written and read to/from
+    // disk in big endian order.
+    var res: T = try reader.readStruct(T);
+    if (native_endian != builtin.Endian.Big) {
+        mem.bswapAllFields(T, &res);
+    }
+    return res;
 }
 
 fn readLoadCommands(self: *Dylib, reader: anytype) !void {
@@ -285,12 +337,12 @@ fn parseSymbols(self: *Dylib) !void {
 
     var symtab = try self.allocator.alloc(u8, @sizeOf(macho.nlist_64) * symtab_cmd.nsyms);
     defer self.allocator.free(symtab);
-    _ = try self.file.?.preadAll(symtab, symtab_cmd.symoff);
+    _ = try self.file.?.preadAll(symtab, symtab_cmd.symoff + self.library_offset);
     const slice = @alignCast(@alignOf(macho.nlist_64), mem.bytesAsSlice(macho.nlist_64, symtab));
 
     var strtab = try self.allocator.alloc(u8, symtab_cmd.strsize);
     defer self.allocator.free(strtab);
-    _ = try self.file.?.preadAll(strtab, symtab_cmd.stroff);
+    _ = try self.file.?.preadAll(strtab, symtab_cmd.stroff + self.library_offset);
 
     for (slice) |sym| {
         const sym_name = mem.spanZ(@ptrCast([*:0]const u8, strtab.ptr + sym.n_strx));
--- a/lib/std/elf.zig
+++ b/lib/std/elf.zig
@@ -429,7 +429,7 @@ pub fn ProgramHeaderIterator(ParseSource: anytype) type {
                 if (self.elf_header.endian == native_endian) return phdr;
 
                 // Convert fields to native endianness.
-                bswapAllFields(Elf64_Phdr, &phdr);
+                mem.bswapAllFields(Elf64_Phdr, &phdr);
                 return phdr;
             }
 
@@ -441,7 +441,7 @@ pub fn ProgramHeaderIterator(ParseSource: anytype) type {
             // ELF endianness does NOT match native endianness.
             if (self.elf_header.endian != native_endian) {
                 // Convert fields to native endianness.
-                bswapAllFields(Elf32_Phdr, &phdr);
+                mem.bswapAllFields(Elf32_Phdr, &phdr);
             }
 
             // Convert 32-bit header to 64-bit.
@@ -479,7 +479,7 @@ pub fn SectionHeaderIterator(ParseSource: anytype) type {
                 if (self.elf_header.endian == native_endian) return shdr;
 
                 // Convert fields to native endianness.
-                bswapAllFields(Elf64_Shdr, &shdr);
+                mem.bswapAllFields(Elf64_Shdr, &shdr);
                 return shdr;
             }
 
@@ -491,7 +491,7 @@ pub fn SectionHeaderIterator(ParseSource: anytype) type {
             // ELF endianness does NOT match native endianness.
             if (self.elf_header.endian != native_endian) {
                 // Convert fields to native endianness.
-                bswapAllFields(Elf32_Shdr, &shdr);
+                mem.bswapAllFields(Elf32_Shdr, &shdr);
             }
 
             // Convert 32-bit header to 64-bit.
@@ -529,26 +529,6 @@ pub fn int32(need_bswap: bool, int_32: anytype, comptime Int64: anytype) Int64 {
     } else {
         return int_32;
     }
-}
-
-pub fn bswapAllFields(comptime S: type, ptr: *S) void {
-    if (@typeInfo(S) != .Struct) @compileError("bswapAllFields expects a struct as the first argument");
-    inline for (std.meta.fields(S)) |f| {
-        @field(ptr, f.name) = @byteSwap(f.field_type, @field(ptr, f.name));
-    }
-}
-test "bswapAllFields" {
-    var s: Elf32_Chdr = .{
-        .ch_type = 0x12341234,
-        .ch_size = 0x56785678,
-        .ch_addralign = 0x12124242,
-    };
-    bswapAllFields(Elf32_Chdr, &s);
-    try std.testing.expectEqual(Elf32_Chdr{
-        .ch_type = 0x34123412,
-        .ch_size = 0x78567856,
-        .ch_addralign = 0x42421212,
-    }, s);
 }
 
 pub const EI_NIDENT = 16;
--- a/src/link/Elf.zig
+++ b/src/link/Elf.zig
@@ -1108,7 +1108,7 @@ pub fn flushModule(self: *Elf, comp: *Compilation) !void {
                 for (buf) |*phdr, i| {
                     phdr.* = progHeaderTo32(self.program_headers.items[i]);
                     if (foreign_endian) {
-                        std.elf.bswapAllFields(elf.Elf32_Phdr, phdr);
+                        mem.bswapAllFields(elf.Elf32_Phdr, phdr);
                     }
                 }
                 try self.base.file.?.pwriteAll(mem.sliceAsBytes(buf), self.phdr_table_offset.?);
@@ -1120,7 +1120,7 @@ pub fn flushModule(self: *Elf, comp: *Compilation) !void {
                 for (buf) |*phdr, i| {
                     phdr.* = self.program_headers.items[i];
                     if (foreign_endian) {
-                        std.elf.bswapAllFields(elf.Elf64_Phdr, phdr);
+                        mem.bswapAllFields(elf.Elf64_Phdr, phdr);
                     }
                 }
                 try self.base.file.?.pwriteAll(mem.sliceAsBytes(buf), self.phdr_table_offset.?);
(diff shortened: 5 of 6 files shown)


src/link/MachO/Dylib.zig
```diff
@@ -74,6 +75,8 @@ pub fn closeFile(self: Dylib) void {
 pub fn parse(self: *Dylib) !void {
     log.debug("parsing shared library '{s}'", .{self.name.?});
 
+    var seek_pos = try self.file.?.getPos();
```
LemonBoy, line 78: Given `zld` assumes all the input mach-o files are 64bit ones I'd place the fat-parsing code here (or, better, in a subroutine that's called here).
The base offset can be computed once, according to the selected architecture, and then stored in `Dylib` to avoid passing `seek_pos` everywhere.

LemonBoy (commented):
Small drive-by code review, I hope @kubkon doesn't mind.

src/link/MachO/Dylib.zig
```diff
@@ -179,12 +182,39 @@ pub fn parseSymbols(self: *Dylib) !void {
...
         };
+        log.warn("Here {s}", .{name});
 
         try self.symbols.putNoClobber(self.allocator, name, &proxy.base);
     }
 }
 
 pub fn isDylib(file: fs.File) !bool {
+    {
+        const fat_header = try file.reader().readStructEndian(macho.fat_header, builtin.Endian.Big);
```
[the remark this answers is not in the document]
kubkon: Adding a check for `MH_MAGIC_64` or `MH_CIGAM_64` would be nice, and it should also be propagated to `Object.isObject`. The ultimate purpose of `isDylib` and `isObject` is to work out which parser to use `Object.parse` or `Dylib.parse` in zld, nothing more.

lib/std/io/reader.zig
```diff
@@ -296,6 +297,23 @@ pub fn Reader(
...
+        // this is copied from elf.zig - this should be DRYed, but not sure where it would
+        // best live?
+        pub fn bswapAllFields(comptime S: type, ptr: *S) void {
+            if (@typeInfo(S) != .Struct) @compileError("bswapAllFields expects a struct as the first argument");
+            inline for (std.meta.fields(S)) |f| {
+                @field(ptr, f.name) = @byteSwap(f.field_type, @field(ptr, f.name));
+            }
+        }
+
+        pub fn readStructEndian(self: Self, comptime T: type, endian: builtin.Endian) !T {
```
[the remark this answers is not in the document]
daurnimator: I think this has come up before.

A quick search around just found https://github.com/ziglang/zig/pull/8108#discussion_r584686657 though

src/link/MachO/Dylib.zig
```diff
@@ -179,12 +182,39 @@ pub fn parseSymbols(self: *Dylib) !void {
...
+            var fat_arch_index: u32 = 0;
+            while (fat_arch_index < fat_header.nfat_arch) : (fat_arch_index += 1) {
+                const fat_arch = try file.reader().readStructEndian(macho.fat_arch, builtin.Endian.Big);
+                log.warn("Fat arch header details {}, {}, {}, {}", .{
+                    fat_arch.cputype,
+                    fat_arch.cpusubtype,
+                    fat_arch.offset,
+                    fat_arch.size
+                });
+                if(fat_arch.cputype == macho.CPU_TYPE_X86_64) {
```
kubkon, line 205: As implied by @LemonBoy, we should leave the decision making whether this FAT dylib contains the architecture of interest to the actual parsing logic, and then warn the user if it doesn't but don't trigger a hard error. I guess a hard error should only ever occur if the parsing cannot be performed, read, the dylib is malformed.

src/link/MachO/Dylib.zig
kubkon: It might be worth looking at how this is done in `Archive.zig` which after all is a collection of objects, so naturally we need to seek to find the right object.

kubkon: Oh and we should not have to parse *all* dylibs within a fat dylib, only the one matching the requested architecture.

src/link/MachO/Zld.zig
```diff
@@ -1833,9 +1833,13 @@ fn resolveSymbols(self: *Zld) !void {
         var found = false;
         for (self.dylibs.items) |dylib| {
             const proxy = dylib.symbols.get(undef.name) orelse continue;
-            try self.imports.putNoClobber(self.allocator, proxy.name, proxy);
-            undef.alias = proxy;
-            found = true;
+            log.warn("{s}", .{proxy.name});
+            if(self.imports.contains(proxy.name))
```
kubkon, line 1837: Proxies should never collide meaning `self.imports.putNoClobber(...)` should never trap. If that is the case, the linking will most likely result in missing links between `Undef`s and `Proxy`s.

kubkon: Either way, this logic is completely rewritten as part of #9171 so you might wanna wait for this to land before tweaking the symbol resolution logic.

kubkon:
> Small drive-by code review, I hope @kubkon doesn't mind.

Not at all!

kubkon:
Just something to think about: I thought that a universal binary is just a collection of dylibs per architecture that are well-formed (hence fat since they explode in size I guess). With this in mind, I wouldn't expect any need to tweak any symbol resolution logic; rather, I'd expect a helper function which takes a file descriptor to a fat dylib and required architecture, and based on this, creates a `*Dylib` struct. zld would then simply append this `*Dylib` to `self.dylibs` `ArrayList` and that should be it. Does that make sense?

src/link/MachO/Zld.zig
kubkon: See the comment below: I'm surprised you need to touch the symbol resolution logic at all.

moosichu:
Cheers! Thanks for the feedback! Yeah - thanks for reviewing the obviously bad stuff & giving feedback on how you would resolve that within the context of the compiler. Will take it all on-board :)

kubkon:
> Cheers! Thanks for the feedback! Yeah - thanks for reviewing the obviously bad stuff & giving feedback on how you would resolve that within the context of the compiler. Will take it all on-board :)

Thanks for working on this @moosichu! Feel free to ping me here or on Discord if you get stuck or need to discuss something - I'll be happy to unblock you!

src/link/MachO/Zld.zig
moosichu: Yeah - it's worth pointing out that this was just a short-term hack to just get an example working. I'm getting some kind of conflict with the library I've been testing with - I just don't know why & this was a very hacky way to just work around that.

src/link/MachO/Dylib.zig
moosichu: At the moment if you try to link with a dylib that doens't match the requested architecture you _do_ get a hard error - so I'm going to match that existing logic for now, but can change both to a warning if you think that would be best afterwards.

src/link/MachO/Zld.zig
moosichu: So the reason I added that log warnings was to see why that was happening - and every proxy.name appears twice (back-to-back) for some reason. This is what i get when linking with the vulkan dynamic libraries:

```
warning(zld): _vkQueuePresentKHR
warning(zld): _vkQueuePresentKHR
warning(zld): _vkGetSwapchainImagesKHR
warning(zld): _vkGetSwapchainImagesKHR
warning(zld): _vkGetPhysicalDeviceSurfaceSupportKHR
warning(zld): _vkGetPhysicalDeviceSurfaceSupportKHR
warning(zld): _vkGetPhysicalDeviceSurfaceFormatsKHR
warning(zld): _vkGetPhysicalDeviceSurfaceFormatsKHR
warning(zld): _vkGetPhysicalDeviceSurfaceCapabilitiesKHR
warning(zld): _vkGetPhysicalDeviceSurfaceCapabilitiesKHR
warning(zld): _vkGetPhysicalDeviceQueueFamilyProperties
warning(zld): _vkGetPhysicalDeviceQueueFamilyProperties
warning(zld): _vkGetPhysicalDeviceMemoryProperties
warning(zld): _vkGetPhysicalDeviceMemoryProperties
warning(zld): _vkGetDeviceQueue
warning(zld): _vkGetDeviceQueue
warning(zld): _vkFreeMemory
warning(zld): _vkFreeMemory
warning(zld): _vkEnumeratePhysicalDevices
warning(zld): _vkEnumeratePhysicalDevices
warning(zld): _vkEndCommandBuffer
warning(zld): _vkEndCommandBuffer
warning(zld): _vkDestroySwapchainKHR
warning(zld): _vkDestroySwapchainKHR
warning(zld): _vkDestroyInstance
warning(zld): _vkDestroyInstance
warning(zld): _vkDestroyDevice
warning(zld): _vkDestroyDevice
warning(zld): _vkDestroyDescriptorPool
warning(zld): _vkDestroyDescriptorPool
warning(zld): _vkDestroyCommandPool
warning(zld): _vkDestroyCommandPool
warning(zld): _vkCreateSwapchainKHR
warning(zld): _vkCreateSwapchainKHR
warning(zld): _vkCreateShaderModule
warning(zld): _vkCreateShaderModule
warning(zld): _vkCreateSemaphore
warning(zld): _vkCreateSemaphore
warning(zld): _vkCreatePipelineLayout
warning(zld): _vkCreatePipelineLayout
warning(zld): _vkCreatePipelineCache
warning(zld): _vkCreatePipelineCache
warning(zld): _vkCreateInstance
warning(zld): _vkCreateInstance
warning(zld): _vkCreateImageView
warning(zld): _vkCreateImageView
warning(zld): _vkCreateDevice
warning(zld): _vkCreateDevice
warning(zld): _vkCreateDescriptorSetLayout
warning(zld): _vkCreateDescriptorSetLayout
warning(zld): _vkCreateDescriptorPool
warning(zld): _vkCreateDescriptorPool
warning(zld): _vkCreateComputePipelines
warning(zld): _vkCreateComputePipelines
warning(zld): _vkCreateCommandPool
warning(zld): _vkCreateCommandPool
warning(zld): _vkCmdBindPipeline
warning(zld): _vkCmdBindPipeline
warning(zld): _vkCmdBindDescriptorSets
warning(zld): _vkCmdBindDescriptorSets
warning(zld): _vkBeginCommandBuffer
warning(zld): _vkBeginCommandBuffer
warning(zld): _vkAllocateMemory
warning(zld): _vkAllocateMemory
warning(zld): _vkAllocateDescriptorSets
warning(zld): _vkAllocateDescriptorSets
warning(zld): _vkAllocateCommandBuffers
warning(zld): _vkAllocateCommandBuffers
warning(zld): _vkAcquireNextImageKHR
warning(zld): _vkAcquireNextImageKHR
warning(zld): _vkResetCommandPool
warning(zld): _vkResetCommandPool
warning(zld): _vkUpdateDescriptorSets
warning(zld): _vkUpdateDescriptorSets
```

moosichu: Oh haha - I've also just realised that the if logic I intended was inverted - whoops.

moosichu: Yep - that explains why the built binary never actually ran properly (I only had the compile "successfully" complete) - no that I've actually got the if statement doing what I intended I do now have a functioning executable! :D 

However- now I need to actually solve the problem that is causing a need for this (because it should not be needed).

moosichu:
OK - I've been able to build a little on feedback & identify at least one bug which has now been fixed - so the compiled executable (which I have been cross-compiling from my Linux desktop because I don't have a zig source-build available on my Mac) actually ran this time! 

So I believe the only _functional_ problem remaining is the fact that I've had to add that hacky if statement to Zld.zig - and I need to found out what is wrong with the previous logic to cause that. Then the rest of the work is just robustifying & cleaning-up what I have written and making it PR-ready, and we should be in a good place.

lib/std/io/reader.zig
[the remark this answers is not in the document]
moosichu: Yeah - I'm not sure I'm super comfortable changing the prototype of readStruct, but I'm also not sure if adding this function is the right solution either.

src/link/MachO/Zld.zig
moosichu: OK - I'm done for the day, so will investigate this issue another time. If you have any ideas of what might be causing these duplicate symbols please do let me know. Basically - if I don't have this if statement - I get a panic in putNoClobber.

kubkon: I think you might wanna rebase onto `zld-tapi` locally, remove any tweaks to symbol resolutions, and see if it works fine then. Kinda hoping that changes in that branch made the symbol resolution logic more robust.

moosichu: Yes! That seems to resolve the issue. Just running the built executable on a Mac now to check that it works. 

If this is the thing that solves this then - what would you like me to do w.r.t this branch/PR relative to the `zld-tapi` branch?

moosichu: The built executable ran successfully! :) I've pushed the removal of the ZLD changes to the branch here - but let me know what would be good for me to do regarding this.

Also if you have any further feedback on the current state of the PR that would be really helpful - I believe I've resolved some of it, but as this is my first real chunk of zig code that I've written (outside of a build.zig file for me C++ project), I'm aware it might be a bit idiosyncratic. 

If it does reach the point where from a code-POV it's an OK set of changes to merge, would you also like me to squash the commits?

lib/std/io/reader.zig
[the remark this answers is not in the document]
LemonBoy: Given this change is orthogonal to the PR I'd take this off the patch and use `readStruct` + `bswapAllFields`.

lib/std/macho.zig
```diff
@@ -24,6 +24,20 @@ pub const mach_header_64 = extern struct {
     reserved: u32,
 };
 
+pub const fat_header = extern struct {
+    magic: u32,
+    nfat_arch: u32,
+};
+
+pub const fat_arch = extern struct
```
LemonBoy, line 32: The formatting looks wonky, please run `zig fmt` over all the files you touch (that's a pre-requisite for the CI to pass).

lib/std/mem.zig
```diff
@@ -1534,6 +1534,29 @@ test "writeIntBig and writeIntLittle" {
...
+/// (Changing their endianess)
+pub fn bswapAllFields(comptime S: type, ptr: *S) void {
+    if (@typeInfo(S) != .Struct) @compileError("bswapAllFields expects a struct as the first argument");
+    inline for (std.meta.fields(S)) |f| {
+        @field(ptr, f.name) = @byteSwap(f.field_type, @field(ptr, f.name));
+    }
+}
+
+test "bswapAllFields" {
+    var s: Elf32_Chdr = .{
```
LemonBoy, line 1547: `Elf32_Chdr` is not defined here, I'd replace this test with something like this:
```zig
const T = extern struct {
    f0: u8,
    f1: u16,
    f2: u32,
};
var s = T{
    .f0 = 0x12,
    .f1 = 0x1234,
    .f2 = 0x12345678,
};
bswapAllFields(T, &s);
try std.testing.expectEqual(T{
    .f0 = 0x12,
    .f1 = 0x3412,
    .f2 = 0x78563412,
}, s);
```

src/link/MachO/Dylib.zig
```diff
@@ -71,9 +72,53 @@ pub fn closeFile(self: Dylib) void {
     }
 }
 
+pub fn decodeArch(cputype : macho.cpu_type_t) !std.Target.Cpu.Arch {
```
LemonBoy, line 75: If the function is not used outside of `Dylib` I'd drop the `pub`.

src/link/MachO/Dylib.zig
```diff
@@ -71,9 +72,53 @@ pub fn closeFile(self: Dylib) void {
...
+                if(libArch == self.arch.?) {
+                    // We have found a matching architecture!
+                    try self.file.?.seekTo(fat_arch.offset);
+                    library_offset = fat_arch.offset;
+                    found_arch = true;
+                    break;
+                }
+            }
+
+            if(!found_arch) {
```
LemonBoy, line 113: Little tip, `while` expressions accept an `else` branch that's executed when the loop completed or, in other words, no `break` was executed. You can drop the `found_arch` flag using this nice construction.

src/link/MachO/Dylib.zig
```diff
@@ -71,9 +72,53 @@ pub fn closeFile(self: Dylib) void {
...
+    return arch;
+}
+
 pub fn parse(self: *Dylib) !void {
     log.debug("parsing shared library '{s}'", .{self.name.?});
 
+    var library_offset : u64 = 0;
+    {
+        const fat_header = try self.file.?.reader().readStructEndian(macho.fat_header, builtin.Endian.Big);
+        if(fat_header.magic == macho.FAT_MAGIC) {
```
LemonBoy, line 93: If you want to be extra-correct you could detect all the valid `FAT_*` numbers here and report an error if the endianness/format doesn't match the only format you're parsing here.

src/link/MachO/Dylib.zig
```diff
@@ -71,9 +72,53 @@ pub fn closeFile(self: Dylib) void {
...
+            var fat_arch_index: u32 = 0;
+            var found_arch = false;
+            while (fat_arch_index < fat_header.nfat_arch) : (fat_arch_index += 1) {
+                const fat_arch = try self.file.?.reader().readStructEndian(macho.fat_arch, builtin.Endian.Big);
+                // If we come across an architecture that we do not know how to handle, that's
+                // fine because we can keep looking for one that might match.
+                const libArch = decodeArch(fat_arch.cputype) catch |err| switch(err) {
```
LemonBoy, line 100: RE: naming conventions. Variables are written in `snake_case`, methods in `camelCase`, types start with a capital letter and constants are (usually) in all-caps.

moosichu: Yes cheers! This was just out of habit from - will make sure to sort these!

src/link/MachO/Dylib.zig
moosichu: The fat headers are _always_ big endian. I don't know if it's worth checking for the incorrect endianess as that would imply something has gone horribly wrong with the source file?

LemonBoy: My bad, I missed that part in the `fat.h` header:
```
* All structures defined here are always written and read to/from disk
* in big-endian order.
```
Carry on then, and sorry for the noise!

src/link/MachO/Dylib.zig
moosichu: So I've done this work now - the only potential concern is that we are now reading the fat headers multiple times from disk (once to check if it embeds dylibs & once to search for a good candidate when parsing), I think it would be preferable to do this work only once (especially for people not compiling on an SSD).

src/link/MachO/Dylib.zig
moosichu: I've added a comment to the code though now to make this clearer :)

src/link/MachO/Dylib.zig
```diff
@@ -82,22 +123,22 @@ pub fn parse(self: *Dylib) !void {
...
+
     if (this_arch != self.arch.?) {
         log.err("mismatched cpu architecture: expected {s}, found {s}", .{ self.arch.?, this_arch });
         return error.MismatchedCpuArchitecture;
     }
 
     try self.readLoadCommands(reader);
     try self.parseId();
-    try self.parseSymbols();
+    try self.parseSymbols(library_offset);
```
kubkon, line 141: Given that we may want to parse additional file contents in the future, I think it would make more sense to preserve the offset in `Dylib` and add it whenever we `pread` from the file.

src/link/MachO/Dylib.zig
```diff
@@ -71,9 +73,48 @@ pub fn closeFile(self: Dylib) void {
     }
 }
 
+fn decodeArch(cputype: macho.cpu_type_t) !std.Target.Cpu.Arch {
+    const arch: std.Target.Cpu.Arch = switch (cputype) {
+        macho.CPU_TYPE_ARM64 => .aarch64,
+        macho.CPU_TYPE_X86_64 => .x86_64,
+        else => |value| {
```
kubkon, line 80: Could you move the `log.err` informing which arch was passed to here?

src/link/MachO/Dylib.zig
```diff
@@ -82,22 +123,22 @@ pub fn parse(self: *Dylib) !void {
...
 
-    const this_arch: std.Target.Cpu.Arch = switch (self.header.?.cputype) {
-        macho.CPU_TYPE_ARM64 => .aarch64,
-        macho.CPU_TYPE_X86_64 => .x86_64,
-        else => |value| {
-            log.err("unsupported cpu architecture 0x{x}", .{value});
-            return error.UnsupportedCpuArchitecture;
+    const this_arch: std.Target.Cpu.Arch = decodeArch(self.header.?.cputype) catch |err| switch (err) {
+        error.UnsupportedCpuArchitecture => |e| {
+            log.err("unsupported cpu architecture 0x{x}", .{self.header.?.cputype});
```
kubkon, line 128: Could you move this into `decodeArch`? Then, you can simplify it to `try decodeArch` without the need to switch on the error variants.

src/link/MachO/Dylib.zig
```diff
@@ -71,9 +73,48 @@ pub fn closeFile(self: Dylib) void {
...
+                }
+            } else {
+                log.err("Could not find matching cpu architecture in fat library: expected {s}", .{self.arch.?});
+                return error.MismatchedCpuArchitecture;
+            }
+        } else {
+            try self.file.?.seekTo(0);
+        }
+    }
+
```
kubkon, line 117: I think you could simplify this greatly using labeled blocks like so:

```zig
const library_offset: i64 = offset: {
    const fat_header = try readFatStruct(self.file.?.reader(), macho.fat_header);
    if (fat_header.magic != macho.FAT_MAGIC) break :offset 0;

    var fat_arch_index: u32 = 0;
    while (fat_arch_index < fat_header.nfat_arch) : (fat_arch_index += 1) {
        const fat_arch = try readFatStruct(self.file.?.reader(), macho.fat_arch);
        // If we come across an architecture that we do not know how to handle, that's
        // fine because we can keep looking for one that might match.
        const lib_arch = decodeArch(fat_arch.cputype) catch |err| switch (err) {
            error.UnsupportedCpuArchitecture => continue,
            else => |e| return e,
         };
         if (lib_arch == self.arch.?) {
             // We have found a matching architecture!
             break :offset fat_arch.offset;
         }
     } else {
         log.err("Could not find matching cpu architecture in fat library: expected {s}", .{self.arch.?});
         return error.MismatchedCpuArchitecture;
     }
};
try self.file.?.seekTo(library_offset);
```

src/link/MachO/Dylib.zig
```diff
@@ -185,7 +226,49 @@ pub fn parseSymbols(self: *Dylib) !void {
...
+    {
+        // If the given file is a fat/universal binary - find the first entry within it and
+        // check if it is a dynamic library
+        const fat_header = readFatStruct(file.reader(), macho.fat_header) catch |err| switch (err) {
+            error.EndOfStream => return false,
+            else => |e| return e,
+        };
+
+        if (fat_header.magic == macho.FAT_MAGIC) {
+            var fat_arch_index: u32 = 0;
```
kubkon, line 238: I don't think this var is used anywhere?

kubkon (changes requested):

src/link/MachO/Dylib.zig
moosichu: That is what it was originally - but I moved it here because a fat lib containing an unsupported arch itself to me doesn't feel like it should cause an error (because we only care about it supporting the target arch at all) . It's only an error on normal dylib.

kubkon: This will always be an error in zld since we most likely wont support anything else in the near future. Also, this helper should really be reused all over the MachO linker but that's an exercise for the future.

moosichu: Sorry, maybe I should be less cautious about this. But isn't it within the bounds of possibility that a certain library might only be available/distributed as a fat lib containing older architectures (like PowerPC) & x86 - we still want ZLD to link against the x86 part of that right without complaining about the fact that it contains another arch within it (this could also hypothetically extend to the future).

Maybe you are right though in that pragmatically - Apple are (well, Apple) and working with fat libs that contain any arch other than ARM64 & x86-64 is not going to happen for the foreseeable future (and fat lib support will become less relevant anyway as Apple Silicon becomes more prevalent) - and older fat libs that were created at previous transition points that can't be rebuilt from source are probably more than old enough to not warrant needing to support.

src/link/MachO/Dylib.zig
moosichu: Based on our other conversation this should be turned into a try right (and just error if we come across unsupported arch):
```zig

// If we come across an architecture that we do not know how to handle, that's
// fine because we can keep looking for one that might match.
const lib_arch = decodeArch(fat_arch.cputype) catch |err| switch (err) {
    error.UnsupportedCpuArchitecture => continue,
    else => |e| return e,
};
```

src/link/MachO/Dylib.zig
kubkon: Ah, I think I see what you mean: this would print a log message to the stderr when iterating over the archs contained within the fat lib which you are right, we do not want to do. If someone packs powerpc, x86, in addition to arm64 or x86_64, then that's fine as long as we find a lib we can link against. Nvm then, ignore my comment.

src/link/MachO/Dylib.zig
moosichu: Are you happy if I leave things as is currently with the PR, or should we address these concerns?

src/link/MachO/Dylib.zig
kubkon: The CI is failing since this is an used capture here. You capture any other potential architecture as `value` but never use it. In this case, just omit the capture altogether like so:

```zig
else => {
```

kubkon: The CI is failing since this is an used capture here. You capture any other potential architecture as `value` but never use it. In this case, just omit the capture altogether like so:

```zig
else => {
```

kubkon (changes requested):

src/link/MachO/Zld.zig
moosichu: I've squashed the commits (because it felt like thee right thing to do) - the only thing I'm still unsure about is if this ready to merge as-is (because the changes were only actually functional for me when merged with the zld-tapi branch). What is the eta for the zld-tapi branch to be merged into master?

kubkon: `zld-tapi` landing in master should happen really soon.

moosichu:
Beyond the multiple file-seek issue I mentioned up-top, I've addressed all the feedback (I think) so unless that is an issue I think this is now ready to be merged. Let me know if there are any other desirable changes that I should make!

kubkon:
@moosichu #9171 has now been merged. Could you rebase onto master please?

moosichu:
> @moosichu #9171 has now been merged. Could you rebase onto master please?

Sure thing! Will be able to do it later this evening.

moosichu:
All done @kubkon ! :)

Issue #9182, opened 2021-06-20
`@typeInfo` strings cannot be null terminated.


Vexu:
Taken from #8636.

> > Looks good so far, do you also want to make the change to `@typeInfo` and `@src`?
> 
> I have local changes ready to update string fields contained in structures returned by these two builtins to `[:0]const u8`.
> 
> But while applying changes to `@typeInfo()` I realized its `std.builtin.TypeInfo` output is also an input to `@Type()` and that could break existing code. For instance, this assignment fails since it uses a `[]const u8` as returned by `std.mem.tokenize()` and that can't implicitly cast to `[:0]const u8`:

I see three ways to fix this.

1. Have separate types for `@typeInfo` and `@Type`, which would break even more code.
2. Leave things as they are, which leaves the awkwardness from the issue that the pr was trying to solve.
3. Add casts from `[]T -> [:t]T` and `*[N]T -> *[N:t]T` when the value is comptime known.

The last option would likely solve some other awkwardness about dealing with sentinel terminated values but includes a lot of implicitness.

jmc-88:
The code that broke with the local changes I had for amending `std.builtin.TypeInfo` were due to the fact that the [failing testcase](https://github.com/ziglang/zig/blob/e006281d79a31e10bdbc07b3e33427d719ec01df/test/behavior/bugs/6456.zig#L16-L20) is using `std.mem.tokenize()` at comptime. The problem here is that there's no way to make the slice returned by the `TokenIterator` into a null-terminated slice without allocating dedicated memory for it, which is impossible at comptime.

So, another option: introduce a comptime allocator to unbreak code that follows a similar pattern even if `@Type()` starts requiring null-terminated slices. This would be unblocked by issue #1291.

But it'd still make its use a bit more awkward, potentially requiring the user to allocate/deallocate memory whereas now no such dance is required.

ifreund:
> The problem here is that there's no way to make the slice returned by the `TokenIterator` into a null-terminated slice without allocating dedicated memory for it, which is impossible at comptime.

Actually, this can be done:
```zig
const std = @import("std");
const testing = std.testing;

const text =
    \\f1
    \\f2
    \\f3
;

test "issue 6456" {
    comptime var gen_fields: []const [:0]const u8 = &[_][:0]const u8{};
    comptime {
        var it = std.mem.tokenize(text, "\n");
        while (it.next()) |name| {
            const name_z = name ++ "\x00";
[...]

PR #9183, opened 2021-06-20
src: return a null-terminated slice

jmc-88:
As noted by @Vexu in https://github.com/ziglang/zig/pull/8636#pullrequestreview-687047598. This amends the previous pull request so to also align `@src()` to the other builtins in returning a sentinel-terminated value, and is part of https://github.com/ziglang/zig/issues/3779.

This does not address `@typeInfo()` which is a trickier case and was forked off to https://github.com/ziglang/zig/issues/9182.


jmc-88:
For the record, the linked failure is:

```
Test [136/1013] behavior.atomics.test "behavior-native-ReleaseFast-bare-single atomicrmw with floats"... FAIL (TestUnexpectedResult)
```

which is unrelated to this change.

PR #9186, opened 2021-06-21
Add context to fatal OpenErrors

RossSmyth:
Hello, I'm new to Zig and had some frustrations earlier today building Zig and passing `--zig-lib-dir` then having "FileNotFound" error returned, but not being sure what file (or directory it seems) wasn't being found. This pull replaces the `try` on `openDir()` methods with a `catch` that if `FileNotFound`, will call `fatal()` with the directory that was supposed to be found formatted in.

I checked the call hierarchy and all of these errors should have lead to an exit of the compiler regardless, except without `fatal()` they would've just bubbled up to `main()` without any context. One thing I am unsure of is whether the paths formatted in should be absolute or relative. I did absolute when convenient, but otherwise just formatted in what was attempted to be opened.

One thing not done is add any testing of these exits. Based upon the [testMissingOutputPath()](https://github.com/ziglang/zig/blob/86ebd4b975d2f1e01f468e06daab6e28c06f9719/test/cli.zig#L144) test it seems that is desired. Also these catches should be done for `openFile()` methods as well to provide more context.

Thanks for your work, and any help/critique you can provide.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1778,7 +1778,7 @@ fn buildOutputType(
                 const handle = fs.cwd().openDir(dirname, .{}) catch |err| {
                     switch (err) {
                         error.FileNotFound => {
-                            fatal("unable to open directory: '{s}': {s}", .{ dirname, @errorName(err) });
+                            fatal("unable to open output directory '{s}': {s}", .{ dirname, @errorName(err) });
                         },
                         else => return err,
                     }
--- a/src/main.zig
+++ b/src/main.zig
@@ -2496,7 +2496,9 @@ fn cmdTranslateC(comp: *Compilation, arena: *Allocator, enable_cache: bool) !voi
         return cleanExit();
     } else {
         const out_zig_path = try fs.path.join(arena, &[_][]const u8{ "o", &digest, translated_zig_basename });
-        const zig_file = try comp.local_cache_directory.handle.openFile(out_zig_path, .{});
+        const zig_file = comp.local_cache_directory.handle.openFile(out_zig_path, .{}) catch |err| {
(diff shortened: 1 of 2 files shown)


src/main.zig
```diff
@@ -3092,7 +3174,14 @@ pub fn cmdFmt(gpa: *Allocator, args: []const []const u8) !void {
     defer fmt.out_buffer.deinit();
 
     for (input_files.items) |file_path| {
-        try fmtPath(&fmt, file_path, check_flag, fs.cwd(), file_path);
+        fmtPath(&fmt, file_path, check_flag, fs.cwd(), file_path) catch |err| {
```
Vexu, line 3177: This doesn't seem correct, `fmtPath` already handles everything except `AccessDenied` which should be handled in `fmtPathDir` and not be fatal.

Vexu (changes requested):
I don't see any reason to limit this to just `FileNotFound`, unless doing so broke something else.

src/main.zig
RossSmyth: Good point, I missed that in my first read.

RossSmyth:
> 
> 
> I don't see any reason to limit this to just `FileNotFound`, unless doing so broke something else.

After analyzing the possible errors that can returned I agree. I'll change that and the above change when I get home.

Vexu (approved):
One thing that could still be improved is that a lot of these directories have specific names e.g. local cache directory, standard library directory, which could also be added to the error messages.

Vexu (commented):
Error messages are supposed to be lower case but there does seem to be other places in the compiler where this isn't the case yet.

RossSmyth:
Should be fixed now

Vexu:
Thanks!

Issue #8970, opened 2021-06-02
proposal: run stage2 astgen when using stage1


g-w1:
Stage2 has some desirable features that are missing in stage1 in astgen (and I plan to add more like unused variables). This proposal is to run astgen unconditionally on any zig file that is being compiled, c++ or zig implementation.
Pros:
catch more errors earlier
Cons:
a tiny bit slower

mikdusan:
What happens if astgen produces errors that are redundant to those produced by stage1?

g-w1:
Astgen will run first, and if there are any errors, stage1 will abort. This can be hooked into stage1's @import command trivially I think.

g-w1:
This is implemented by https://github.com/ziglang/zig/pull/9191

andrewrk:
Ah thanks, I forgot you filed this.

andrewrk:
Uh it's not merged yet tho

andrewrk:
Landed in 31c49ad64ded02b9dde57f5d3ef102a771fa5cf7.

Issue #2984, opened 2019-07-31
make the C pointer type always a compile error unless the file declares that it is generated code


andrewrk:
Unsurprisingly, the relaxed rules about C pointer types have resulted in people using them where other, more restricted types, would be preferred. Perhaps it's time for the language to get more involved in policing this.

daurnimator:
Do you have a link to examples of people using it as such?

andrewrk:
How this will work:

 * New keyword: `thisfileisautotranslatedfromc`. Nobody wants to type that!
 * The keyword may be the first token in a file, followed by a semicolon. It is illegal otherwise.
 * If the keyword is found, then a file is allowed to contain `[*c]` pointers. Otherwise, the compile error is `error: [*c] pointers are only allowed in auto-translated C code`, `note: * and [*] pointers may be used where a [*c] pointer is expected`.

marler8997:
Another use case for using C pointers is for testing, i.e.

https://github.com/ziglang/zig/pull/3111/files#diff-9bb6309b18efdcc04260fad6ad9f55edR83

marler8997:
So would we add the `thisfileisautotranslatedfromc` token to this test file...even though it isn't translated from C?

https://github.com/ziglang/zig/pull/3111/files#diff-9bb6309b18efdcc04260fad6ad9f55edR91

I could be missing something, but there seems to be more valid use cases for C pointers besides translating from C.  If you're interfacing with a C library, maybe you need to write some utility functions that help bridge the GAP between Zig and C?  It may make sense to use C pointers to add special handling for those specific pointer types?

marler8997:
One thought came to mind.  You could also always allow C pointers in `test` blocks.  That would solve the problem of using C pointers in test code.  But still not sure whether or not there are other valid use cases for C pointers...

daurnimator:
I had a thought today re: C pointers: all `[*c]T` are currently nullable. What if we instead translated C pointers as `?[*c]T`, which would be a "optional C pointer". (you could get a plain `[*c]T` if the C source had `__attribute__((nonnull))`)
That was `[*c]` only means "unknown if single pointer or pointer to many"?

andrewrk:
[Another example today](https://freenode.irclog.whitequark.org/zig/2019-10-23#25615499;):

```
<riba> i need to provide a callback to C code that takes a ** argument
<riba> i tried it like this: [*c][*c]
<riba> now i want to check both pointers to null
[...]

PR #9191, opened 2021-06-22
run AstGen even when using the stage1 backend

andrewrk:
In celebration of @g-w1's PR #9047 landing, this changeset reduces the amount of divergence in the compiler's main pipeline logic enough to run AstGen for all files in the compilation, regardless of whether the stage1 or stage2 backend is being used.
    
Practically, this means that **all Zig code is subject to new compile errors, such as unused local variables.**

Implements #8970
    
Additionally:
 * remove leftover unsound asserts from recent hash map changes
 * fix sub-Compilation errors not indenting correctly
 * CI: no longer pass --ast-check to `zig fmt` because that is now redundant.

### Perf

This is doing strictly more work, so I expect perf to be strictly worse. This is testing on my beefy dell laptop. One observation I make here is that the perf hit with a warm cache is negligible, and the cache system is granular enough that the benefits apply to practically every compilation. Another observation is that even with a cold cache, for large compilations the perf hit is negligible.

#### Hello World, cold cache

 * wall clock: 1.40s to 1.67s (20% slower)
 * peak rss: 254 MiB to 338 MiB (33% more)

#### Hello World, warm cache

 * wall clock: 0s (no change)
 * peak rss: 42 Mib to 96 Mib (127% more)

#### std lib tests, cold cache

 * wall clock: 41s (no change)
 * peak rss: 8.4 GiB to 8.5 GiB (1% more)

#### std lib tests, warm cache

 * wall clock: 0s (no change)
 * peak rss: 122 MiB (no change)

### Merge blockers:
 * [x] Figure out why this is happening:

```
error: unable to load std: FileNotFound
error: unable to load test/std: FileNotFound
error: unable to load test/src/std: FileNotFound
```

It was happening because of #9204.

 * [x] #9205
 * [x] #9206
 * [ ] get the translate-c tests passing
 * [x] update the language reference docs

### Related issues:
 * #9192
 * #9194
 * #9203
 * #9204

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,7 +91,12 @@ set(ZIG_TARGET_MCPU "baseline" CACHE STRING "-mcpu parameter to output binaries 
 set(ZIG_EXECUTABLE "" CACHE STRING "(when cross compiling) path to already-built zig binary")
 set(ZIG_SINGLE_THREADED off CACHE BOOL "limit the zig compiler to use only 1 thread")
 set(ZIG_OMIT_STAGE2 off CACHE BOOL "omit the stage2 backend from stage1")
-set(ZIG_ENABLE_LOGGING off CACHE BOOL "enable logging")
+
+if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+    set(ZIG_ENABLE_LOGGING ON CACHE BOOL "enable logging")
+else()
+    set(ZIG_ENABLE_LOGGING OFF CACHE BOOL "enable logging")
+endif()
 
 if("${ZIG_TARGET_TRIPLE}" STREQUAL "native")
     set(ZIG_USE_LLVM_CONFIG ON CACHE BOOL "use llvm-config to find LLVM libraries")
--- a/ci/azure/linux_script
+++ b/ci/azure/linux_script
@@ -59,9 +59,9 @@ unset CXX
 
 make $JOBS install
 
-# Look for formatting errors and AST errors.
+# Look for non-conforming code formatting.
 # Formatting errors can be fixed by running `zig fmt` on the files printed here.
-release/bin/zig fmt --check --ast-check ..
+release/bin/zig fmt --check ..
 
 # Here we rebuild zig but this time using the Zig binary we just now produced to
 # build zig1.o rather than relying on the one built with stage0. See
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -965,7 +965,7 @@ test "thread local storage" {
     thread2.wait();
 }
 
-fn testTls(context: void) void {
+fn testTls(_: void) void {
     assert(x == 1234);
     x += 1;
     assert(x == 1235);
@@ -2502,6 +2502,8 @@ test "struct namespaced variable" {
 
     // you can still instantiate an empty struct
     const does_nothing = Empty {};
+
+    _ = does_nothing;
 }
 
 // struct field order is determined by the compiler for optimal performance.
@@ -3026,11 +3028,12 @@ const Foo = enum { a, b, c };
 export fn entry(foo: Foo) void { }
       {#code_end#}
       <p>
-      For a C-ABI-compatible enum, use {#syntax#}extern enum{#endsyntax#}:
+      For a C-ABI-compatible enum, provide an explicit tag type to
+      the enum:
       </p>
       {#code_begin|obj#}
-const Foo = extern enum { a, b, c };
-export fn entry(foo: Foo) void { }
+const Foo = enum(c_int) { a, b, c };
+export fn entry(foo: Foo) void { _ = foo; }
       {#code_end#}
       {#header_close#}
 
@@ -3392,9 +3395,11 @@ test "inside test block" {
 test "separate scopes" {
     {
         const pi = 3.14;
+        _ = pi;
     }
     {
         var pi: bool = true;
+        _ = pi;
     }
 }
       {#code_end#}
@@ -3432,7 +3437,7 @@ test "switch simple" {
         // Switching on arbitrary expressions is allowed as long as the
         // expression is known at compile-time.
         zz => zz,
-        comptime blk: {
+        blk: {
             const d: u32 = 5;
             const e: u32 = 100;
             break :blk d + e;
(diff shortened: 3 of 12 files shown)


andrewrk:
This PR breaks godbolt again, and we should be sure to shoot them a friendly PR this time: b9d86a1f2686e7893475addc851225a9f0e57680

andrewrk:
Alright, this PR is now blocking on translate-c generating zig code with unused local variables.

ehaas:
Ok, the latest translate-c problem is this:
<details>
<summary>C code</summary>

```C
#include <stdint.h>
typedef int16_t  __v4hi __attribute__((__vector_size__(8)));
int main(void) {
    __v4hi v4_a = {0, 1, 2, 3};
    __v4hi v4_b = {100, 200, 300, 400};
    __v4hi shuffled = __builtin_shufflevector(v4_a, v4_b, 0, 0, 0, 0);
    return 0;
}
```
</details>

Cleaned up but equivalent translated Zig looks like this:
```Zig
const std = @import("std");
const __v4hi = std.meta.Vector(4, i16);
pub fn main() void {
    var v4_a = __v4hi {0, 0, 0, 0};
    var v4_b = __v4hi {0, 0, 0, 0};
    var shuffled: __v4hi = @shuffle(i16, v4_a, v4_b, comptime std.meta.Vector(4, i32){
        std.zig.c_translation.shuffleVectorIndex(0, @typeInfo(@TypeOf(v4_a)).Vector.len),
        std.zig.c_translation.shuffleVectorIndex(0, @typeInfo(@TypeOf(v4_a)).Vector.len),
        std.zig.c_translation.shuffleVectorIndex(0, @typeInfo(@TypeOf(v4_a)).Vector.len),
        std.zig.c_translation.shuffleVectorIndex(0, @typeInfo(@TypeOf(v4_a)).Vector.len),
    });
    _ = shuffled;
}
```

The problem is that `comptime` on the shuffle mask. With it, we get `error: redundant comptime keyword in already comptime scope`. Without it, we get `error: unable to evaluate constant expression`

Moving the mask into its own variable `const shufflemask = comptime std.meta.Vector(4, i32){...` prevents the error but I wanted to check if that is the proper solution, or is this actually revealing some other bug.
Is

SpexGuy:
This feels like a false positive to me for the redundant comptime keyword, but a workaround is to put parentheses around the whole expression, including the comptime.

g-w1:
The reason it is like this is because shuffle is defined in astgen like this: 
```zig
        .shuffle => {
            const result = try gz.addPlNode(.shuffle, node, Zir.Inst.Shuffle{
                .elem_type = try typeExpr(gz, scope, params[0]),
                .a = try expr(gz, scope, .none, params[1]),
                .b = try expr(gz, scope, .none, params[2]),
                .mask = try comptimeExpr(gz, scope, .none, params[3]),
            });
            return rvalue(gz, rl, result, node);
        },
```
Note how mask is comptimeExpr. I think this workaround will not be needed when stage2 is switched to fully.

ehaas:
I'm still getting the error even with parentheses around the whole expression, and with the internal comptime calls. I think extracting it into its own variable won't be terribly difficult.

andrewrk:
It's not a false positive. The mask parameter of `@shuffle` is comptime, already forcing the entire expression to be comptime. However, stage1 does not implement this correctly. A quick bug fix to stage1 will solve the problem. Give me a few min

andrewrk:
Fixed in fc185a6f71b9bd611a6d808082999a9da0f107e8

ehaas:
How do you want to manage the commit history on this one? I could cherry-pick that change into my branch (along with the just-commited `extern enum` removal) and meld it all together, or I could create a PR against master that removes `comptime` from translate-c's shuffle mask expression.

andrewrk:
@ehaas I merged your PR and @g-w1's PRs into this branch, rebased against master to grab the stage1 fix. So I think the next step would be opening a new PR against this branch. Sound OK?

ehaas:
It looks like the extern enum stuff didn't make it in from the rebase? I think that's why it's saying there's a conflict.

andrewrk:
What extern enum stuff?

ehaas:
https://github.com/ziglang/zig/pull/9164 got merged to master but it looks like it's not in `stage1-astcheck`

andrewrk:
I don't know what went wrong with the rebase before, I must have made some kind of mistake. Anyway it looks like the only remaining item left is to update the language reference (which I'm working on). Thanks @ehaas and @g-w1!

andrewrk:
Oops, spoke too soon. Looks like this test is failing:

```c
#include <stdlib.h>
int main(void) {
   enum Foo { A, B, C };
   enum Foo a = B;
   if (a != B) abort();
   if (a != 1) abort();
   {
      enum Foo { A = 5, B = 6, C = 7 };
      enum Foo a = B;
      if (a != B) abort();
      if (a != 6) abort();
   }
   if (a != B) abort();
   if (a != 1) abort();
}
```

generates

```zig
pub export fn main() c_int {
    const A: c_int = 0;
    const B: c_int = 1;
    const C: c_int = 2;
    const enum_Foo = c_uint;
    var a: enum_Foo = @bitCast(c_uint, B);
    _ = a;
    if (a != @bitCast(c_uint, B)) {
        abort();
    }
    if (a != @bitCast(c_uint, @as(c_int, 1))) {
        abort();
    }
    {
        const A_2: c_int = 5;
        const B_3: c_int = 6;
        const C_4: c_int = 7;
        const enum_Foo_1 = c_uint;
        var a_5: enum_Foo_1 = @bitCast(c_uint, B_3);
        _ = a_5;
        if (a_5 != @bitCast(c_uint, B_3)) {
            abort();
        }
        if (a_5 != @bitCast(c_uint, @as(c_int, 6))) {
            abort();
        }
    }
    if (a != @bitCast(c_uint, B)) {
        abort();
    }
    if (a != @bitCast(c_uint, @as(c_int, 1))) {
        abort();
    }
    return 0;
}
```

```
source.zig:458:15: error: unused local constant
        const C_4: c_int = 7;
              ^
```

ehaas:
Ok, looks like I need to generate discards for non-top-level enum constants. Should be straightforward. Can you push the updated rebase

Vexu:
The issue with discards could also be solved by partially implementing #2984 and disabling the unused variable check if the keyword is present.

PR #9198, opened 2021-06-22
docs: top level global assembly -> container level

g-w1:


Vexu:
I think "container level" could be better here, it could also be used instead of "Global static variables" in #8381.

PR #9201, opened 2021-06-22
stage2: Add "tilde pointing" to errors for multi-character offenses

greenfork:
Error refactoring in #9117 removed this style from main.zig file. This commit adds it back and in addition to all other possible compiler errors.

I added the `lexeme_length` to `Compilation.AllErrors.Message` in order to get more reliable data where it can be done. As an alternative we can use a similar approach as for `Compilation.AllErrors.add` for all errors since `Message` has both `source_line` and `column` present. But it doesn't look very good to me.

Sketchy implementation in `Compilation.AllErrors.add` is inspired by
https://github.com/ziglang/zig/blob/master/lib/std/zig/ast.zig#L97-L116

I tried to test all paths with the following snippets, `zig run`:

```zig
const std = @import("std");

pub fn main() void {
    hello();
}
```
```
Compilation_add.zig:4:5: error: use of undeclared identifier 'hello'
    hello();
    ~~~~~
```
***
```zig
const std = @import("std");

pub fn main() void {
    h();
}
```
```
single_caret.zig:4:5: error: use of undeclared identifier 'h'
    h();
    ^
```
***
```zig
const std = @import("std");

pub fn main() !void {
    const hello = true || false;
}
```
```
addZir_with_token_present.zig:4:11: error: unused local constant
    const hello = true || false;
          ~~~~~
```
***
```zig
const std = @import("std");

pub fn main() !void {
    suspend {
        suspend {}
    }
    // const hello = true || false;
}
```
```
addZir_with_node_present.zig:5:9: error: cannot suspend inside suspend block
        suspend {}
        ~~~~~~~
addZir_with_node_present.zig:4:5: note: other suspend block here
    suspend {
    ~~~~~~~
```

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -317,6 +317,7 @@ pub const AllErrors = struct {
             byte_offset: u32,
             /// Does not include the trailing newline.
             source_line: ?[]const u8,
+            lexeme_length: u32,
             notes: []Message = &.{},
         },
         plain: struct {
@@ -364,7 +365,12 @@ pub const AllErrors = struct {
                             try stderr.writeByte('\n');
                             try stderr.writeByteNTimes(' ', src.column);
                             ttyconf.setColor(stderr, .Green);
-                            try stderr.writeAll("^\n");
+                            if (src.lexeme_length > 1) {
+                                try stderr.writeByteNTimes('~', src.lexeme_length);
+                                try stderr.writeByte('\n');
+                            } else {
+                                try stderr.writeAll("^\n");
+                            }
                             ttyconf.setColor(stderr, .Reset);
                         }
                     }
@@ -404,6 +410,15 @@ pub const AllErrors = struct {
             const byte_offset = try module_note.src_loc.byteOffset(module.gpa);
             const loc = std.zig.findLineColumn(source, byte_offset);
             const sub_file_path = module_note.src_loc.file_scope.sub_file_path;
+            const lexeme_length = blk: {
+                var tokenizer = std.zig.Tokenizer{
+                    .buffer = loc.source_line,
+                    .index = loc.column,
+                    .pending_invalid_token = null,
+                };
+                const token = tokenizer.next();
+                break :blk @intCast(u32, token.loc.end - token.loc.start);
+            };
             note.* = .{
                 .src = .{
                     .src_path = try arena.allocator.dupe(u8, sub_file_path),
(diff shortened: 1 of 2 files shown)


andrewrk:
Here's a fun test case, how does this look?

```zig
test {
    const x: i32 = 1;
    const y: i32 = 1;
    foo(x + y);
}

fn foo(x: bool) void {
    _ = x;
}
```

greenfork:
Thank you for these counter-examples. Funnily enough they produce same error which also doesn't work as expected:
```
start.zig:85:9: error: struct 'counter_example1.counter_example1' has no member named 'main'
    root.main();
        ^
```
and if I change it to use pub fn main then these errors:
```
counter_example1.zig:4:11: error: expected bool, found i32
    foo(x + y);
          ^
```
```
counter_example2.zig:4:11: error: expected bool, found i32
    foo(x + // x is the first operand
          ^
```

So yes, this looks like a lot more work to do. I think I don't have enough understanding to implement it. I will try to come up with a list of questions and maybe attend a weekly stage 2 meeting.

andrewrk:
> Thank you for these counter-examples. Funnily enough they produce same error which also doesn't work as expected:

Ah the instructions for running the example would be: `zig test foo.zig`

greenfork:
More on formatting:
- Rust uses `^~~~` style instead of `~~~~`, examples: https://github.com/rust-lang/rust/pull/19870. Same for gcc and clang
- Ascii art from Rust for errors spanning multiple lines: https://github.com/rust-lang/rust/pull/37369

Examples from gcc and clang for multiline errors:
```
$ cat me.c
int main() {
    return "a" /* comment */
    +          /* comment */
    1;
}

$ gcc me.c
me.c: In function ‘main’:
me.c:3:5: warning: returning ‘char *’ from a function with return type ‘int’ makes integer from pointer without a cast [-Wint-conversion]
    2 |     return "a" /* comment */
      |            ~~~~~~~~~~~~~~~~~
    3 |     +          /* comment */
      |     ^~~~~~~~~~~~~~~~~~~~~~~~
    4 |     1;
      |     ~

$ clang me.c
me.c:3:5: warning: adding 'int' to a string does not append to the string [-Wstring-plus-int]
    +          /* comment */
    ^~~~~~~~~~~~~~~~~~~~~~~~
me.c:3:5: note: use array indexing to silence this warning
    +          /* comment */
    ^
    [
me.c:2:12: warning: incompatible pointer to integer conversion returning 'char *' from a function with result type 'int' [-Wint-conversion]
    return "a" /* comment */
           ^~~~~~~~~~~~~~~~~
2 warnings generated.
```

greenfork:
I will close this PR for now as it appears quite complex. I don't know whether there should be an issue created or there will be a more broad RFC-like issue such as "friendlier errors" so I will leave it as is.

PR #9221, opened 2021-06-24
Include package root dir in stage2 error messages

jcmoyer:
This addresses a problem where error messages emitted by stage2 have an ambiguous source filename, breaking go-to navigation in text editors. This behavior differs from stage1, whose error messages always have an unambiguous filename.

For example, in vscode you can normally alt+click "src\main.zig:300:1" in the integrated terminal and it will take you to that file, line 300. However, stage2 will omit the "src" directory in error messages and navigation will fail; vscode will ask which main.zig you meant.

Before:
```
> zig build
http\http_service.zig:75:17: error: unused function parameter
    pub fn hash(self: HeaderMapContext, k: []const u8) u64 {
                ^

> zig build-exe subdir\test.zig
test.zig:2:11: error: unused local constant
    const y = 123;
          ^
```

After:
```
> zig build
D:\...\src\http\http_service.zig:75:17: error: unused function parameter
    pub fn hash(self: HeaderMapContext, k: []const u8) u64 {
                ^

> zig build-exe subdir\test.zig
subdir\test.zig:2:11: error: unused local constant
    const y = 123;
          ^
```

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -1111,6 +1111,10 @@ pub const Scope = struct {
             return buf.toOwnedSliceSentinel(0);
         }
 
+        pub fn fullPath(file: File, ally: *Allocator) ![]u8 {
+            return file.pkg.root_src_directory.join(ally, &[_][]const u8{file.sub_file_path});
+        }
+
         pub fn dumpSrc(file: *File, src: LazySrcLoc) void {
             const loc = std.zig.findLineColumn(file.source.bytes, src);
             std.debug.print("{s}:{d}:{d}\n", .{ file.sub_file_path, loc.line + 1, loc.column + 1 });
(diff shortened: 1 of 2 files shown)


src/Module.zig
```diff
@@ -1111,6 +1111,10 @@ pub const Scope = struct {
             return buf.toOwnedSliceSentinel(0);
         }
 
+        pub fn fullPath(file: File, gpa: *Allocator) ![]u8 {
```
g-w1, line 1114: Nit: you name the param gpa, but in all the use sites, you pass an arena. This could be a little confusing for some people.

jcmoyer: Fixed.

kristoff-it:
Thank you, this is sorely needed for me even though I don't use vscode.
It's a feature of all terminal emulators that try to find URLs and filepaths in the shell output.

#mousegang

Issue #6977, opened 2020-11-04
Support RELRO


daurnimator:
From https://ctf101.org/binary-exploitation/relocation-read-only/

 
> Relocation Read-Only (or RELRO) is a security measure which makes some binary sections read-only.
> 
> There are two RELRO "modes": partial and full.
> ## Partial RELRO
> 
> Partial RELRO is the default setting in GCC, and nearly all binaries you will see have at least partial RELRO.
> 
> From an attackers point-of-view, partial RELRO makes almost no difference, other than it forces the GOT to come before the BSS in memory, eliminating the risk of a buffer overflows on a global variable overwriting GOT entries.
>
> ## Full RELRO
> 
> Full RELRO makes the entire GOT read-only which removes the ability to perform a "GOT overwrite" attack, where the GOT address of a function is overwritten with the location of another function or a ROP gadget an attacker wants to run.
> 
> Full RELRO is not a default compiler setting as it can greatly increase program startup time since all symbols must be resolved before the program is started. In large programs with thousands of symbols that need to be linked, this could cause a noticable delay in startup time.

kyrias:
As it stands the compiler supports full RELRO, though contrary to what the [0.9.0 release notes says](https://ziglang.org/download/0.9.0/release-notes.html#Zig-Build-System) it is not exposed in the Zig Build System.

In addition to exposing it in the build system, I would also like to suggest that we enable full RELRO by default.  In practice it doesn't make a noticeable difference, even for very large programs, and if someone has a usecase where it's problematic they can just disable it.

To give a few data points, I believe all of the larger Linux distributions have full RELRO enabled by default.  Additionally, I enabled full RELRO by default in Rust in 2017 for all platforms that supports RELRO, except for PPC64 because RHEL 6 still shipped a glibc that had a buggy `BIND_NOW` implementation, so only partial RELRO is enabled there.  But I suspect _that_ won't be a problem for Zig since the support table only lists Linux 3.16+.

anthraxx:
@andrewrk This would be great so see fixed soon. This actually affects downstream distros trying to ship programs with full RELRO enabled.
+1 for default full with option in the build system so adjust it.

PR #9222, opened 2021-06-24
link: Recognize `-z origin|noexecstack|now|relro` linker args.

alexrp:
These are some `-z` linker options that are fairly common in C/C++ projects in the wild.

As far as I can tell, this argument parsing code is just used for `zig cc`/`zig c++`. Do we want some way to surface these to `build-exe`/`build-lib`? `noexecstack`/`relro` are common for security, and `now` has real world usefulness (I once had to use it to due to a really nasty signal handler vs lazy dynamic linking issue - long story...).


LemonBoy:
> Do we want some way to surface these to build-exe/build-lib? 

Yep, see #6977.

alexrp:
Ok, I can look into adding `-z` support for the Zig commands in a later PR.

N00byEdge:
There is also `-z max-page-size=0x1000` to set the section alignment, very useful for bare metal development.

sreehax:
Just confirming that it general -z support would also be very good for my use case since I could also really use the max-page-size option. It would also be nice to support it in a LibExeObjStep in the final product, but this seems fairly trivial to me since that would just append to zig_args.

alexrp:
That might be interesting to look at in a later PR, but it would probably need some consideration since it kind of partly exposes the underlying linker's command line interface directly as part of `zig cc`'s interface, which may or may not be desirable.

E.g. `zig cc` itself is always ~equivalent to `clang`, never `clang-cl`. That provides a nice and consistent CLI for users regardless of the target platform. That would no longer be the case if `-z` options were just blindly passed through to the linker, which can be `ld.lld`, `lld-link`, etc and therefore produce a warning/error if it doesn't recognize a particular option.

PR #9223, opened 2021-06-24
zig test: Add a --test-no-exec flag.

### LINKED ISSUE #9172: Test runner needs more flexibility regarding command line construction
On Windows, I'm trying to achieve something along the lines of:

```bash
wsl `wslpath zig-cache\o\...\test` <path to zig.exe>
```

I'm doing:

```bash
zig test main.zig -target x86_64-linux-gnu --test-cmd wsl --test-cmd "\`wslpath" --test-cmd-bin --test-cmd "\`"
```

This results in:

```
wsl `wslpath zig-cache\o\6a1bb5746dc889ca68f4734e672d2d84\test C:\Users\alex\scoop\apps\ziglang\current\zig.exe `
```

Which of course does not work.

This is a bit unfortunate since WSL (+ QEMU) would open up a lot of possibilities for cross-platform testing.

I'm not exactly sure what the solution might be, but it seems to me that the testing support is a bit too inflexible in general at the moment. In an ideal world, I would like to be able to say `zig build-test ...` and just get an executable I could run in whatever way makes sense for my scenario(s). I don't know how feasible that is, though.

alexrp:
Fixes #9172.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -414,6 +414,7 @@ const usage_build_generic =
     \\  --test-cmd [arg]               Specify test execution command one arg at a time
     \\  --test-cmd-bin                 Appends test binary path to test cmd args
     \\  --test-evented-io              Runs the test in evented I/O mode
+    \\  --test-no-exec                 Compiles test binary without running it
     \\
     \\Debug Options (Zig Compiler Development):
     \\  -ftime-report                Print timing diagnostics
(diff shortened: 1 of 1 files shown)


alexrp:
(The FreeBSD timeout doesn't seem related.)

matu3ba:
Can you add a proper description to your commit log why this is needed or useful to add? Intended goal is to use WSL (+ QEMU) for testing, which belongs into the commit log and github description.

PR #9224, opened 2021-06-24
A couple std.log conveniences

marler8997:
1. Expose default `std.log` implementation

This allows root.log implementations to capture log messages, process them, and forward them to the default implementation if desired.

2. Expose mechanism to convert log levels to string literals

This makes it easier to reproduce and customize the default log message style with the level prefix.  This can be seen by the test cases in `compare_output.zig`.

### DIFF
--- a/lib/std/log.zig
+++ b/lib/std/log.zig
@@ -100,6 +100,23 @@ pub const Level = enum {
     info,
     /// Debug: messages only useful for debugging.
     debug,
+
+    /// Returns a string literal of the given level in full text form.
+    pub fn asText(comptime self: Level) switch (self) {
+        .emerg => @TypeOf("emergency"),
+        .crit => @TypeOf("critical"),
+        .err => @TypeOf("error"),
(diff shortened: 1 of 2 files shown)


lib/std/log.zig
```diff
@@ -100,6 +100,23 @@ pub const Level = enum {
     info,
     /// Debug: messages only useful for debugging.
     debug,
+
+    /// Returns a string literal of the given level in full text form.
+    pub fn asText(comptime self: Level) switch (self) {
```
LemonBoy, line 105: Why not implement a custom `format`?

marler8997: This is used to append to other comptime strings, i.e.
```zig
level.asText() ++ ": " ++ fmt ++ "\n"
```

LemonBoy: Ah, I see. The repetition in the return type and the function body is not that nice but I can't think of a way to DRY it up.

PR #9226, opened 2021-06-24
os.getrandom(): don't use libc on linux

### LINKED ISSUE #9146: Zig requires some syscalls not available on older kernels
So I am working on an embedded linux system that unfortunately I can't upgrade the kernel.  When compiling ziglang for this platform on final link I run into two syscalls that are required but are too new for my device.

getrandom requires linux >= 3.17
copy_file_range requires linux >= 4.5

```
/opt/tomatoware/arm-soft-mmc/lib/gcc/arm-tomatoware-linux-uclibcgnueabi/10.3.0/../../../../arm-tomatoware-linux-uclibcgnueabi/bin/ld: zig1.o: in function `std.crypto.tlcsprng.fillWithOsEntropy':
zig1:(.text+0x78cb0): undefined reference to `getrandom'
/opt/tomatoware/arm-soft-mmc/lib/gcc/arm-tomatoware-linux-uclibcgnueabi/10.3.0/../../../../arm-tomatoware-linux-uclibcgnueabi/bin/ld: zig1.o: in function `std.os.copy_file_range':
zig1:(.text+0xae290): undefined reference to `copy_file_range'
collect2: error: ld returned 1 exit status
```
getrandom I think is pretty easy to workaround.
copy_file_range I thought when looking at the source had a runtime check for availability, but doesn't look like a compile-time check.

jedisct1:
The function is missing from some C libraries, yet it does nothing but call the system call directly.

Fixes #9146

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -140,11 +140,9 @@ pub fn getrandom(buffer: []u8) GetRandomError!void {
     }
     if (builtin.os.tag == .linux or builtin.os.tag == .freebsd) {
         var buf = buffer;
-        const use_c = builtin.os.tag != .linux or
-            std.c.versionCheck(builtin.Version{ .major = 2, .minor = 25, .patch = 0 }).ok;
 
         while (buf.len != 0) {
-            const res = if (use_c) blk: {
+            const res = if (builtin.os.tag != .linux) blk: {
                 const rc = std.c.getrandom(buf.ptr, buf.len, 0);
                 break :blk .{
                     .num_read = @bitCast(usize, rc),


LemonBoy:
The check should be refined instead of removed, the idea is to use the libc whenever possible.

daurnimator:
What is wrong with the existing check?

If you use the right glibc version target shouldn't this work on older machines?

jedisct1:
#9146 refers to the symbol being missing when compiling Zig on uclibc.

We check the libc version, but the check may not do what we expect here.

LemonBoy:
I'm pretty sure it's the user's fault there and that no specific glibc version was specified.
Beside that, we don't support linking against uclibc.

lancethepants:
Isn't the goal of zig to become self hosting at some point? Wouldn't reducing its dependency c be the desirable action?

Well let me bug the uclibc-ng guys. From what I recall glibc will fallback to a userspace implementation if getrandom is not present in the kernel. Maybe uclibc-ng just needs to get that implemented regardless.

LemonBoy:
That's a big misunderstanding of what self-hosted means. The standard library tries to implement the high level constructs it exposes using either the system libc or the syscall interface when possible. On Linux we support both solutions, hence the version checks and the many lines of code to support glibc idiosyncrasies and quirks.

The real problem here is that uclibc is not glibc, supporting this use case requires a bit more effort than removing a version check.

jedisct1:
Having getrandom() in uclibc would solve recurring issues, not limited to Zig. 

If you think you can make it happen, that would be fantastic. 

Alright, closing, then. You may use it to get an uclibc build, but this is not a long-term solution.

Issue #8845, opened 2021-05-20
wasi-libc: enable and pass libstd tests


kubkon:
Once #8837 lands, the next step would be to add as much test coverage as possible, and tweak libstd as required. 

Here's a snippet to get started:

```diff
diff --git a/test/tests.zig b/test/tests.zig
index 243428a53..532cc3cb2 100644
--- a/test/tests.zig
+++ b/test/tests.zig
@@ -59,6 +59,15 @@ const test_targets = blk: {
             .single_threaded = true,
         },

+        TestTarget{
+            .target = .{
+                .cpu_arch = .wasm32,
+                .os_tag = .wasi,
+            },
+            .link_libc = true,
+            .single_threaded = true,
+        },
+
         TestTarget{
             .target = .{
                 .cpu_arch = .x86_64,
```

PR #9227, opened 2021-06-24
WASI,libc: fix libstd with wasi-libc linkage, and enable tests.

mathetake:
This PR modifies libstd for WASI + wasi-libc target, and resolves https://github.com/ziglang/zig/issues/8845


mathetake:
finally resolved the symbol issue but weird.. I see "the number of locals exceeds the minimum".. look into this now

```
Error: failed to run main module `/home/mathetake/zig/zig-cache/o/3f9eec2bc857dd8c7327f378ef491e47/test.wasm`

Caused by:
    0: WebAssembly failed to compile
    1: WebAssembly translation error
    2: Invalid input WebAssembly code at offset 14965884: locals exceed maximum
wasmtime --dir=. /home/mathetake/zig/zig-cache/o/3f9eec2bc857dd8c7327f378ef491e47/test.wasm /home/mathetake/zig/build/zig
/home/mathetake/zig/build/zig test /home/mathetake/zig/lib/std/std.zig -lc --test-name-prefix std-wasm32-wasi-Debug-c-single  --single-threaded --cache-dir /home/mathetake/zig/zig-cache --global-cache-dir /home/mathetake/.cache/zig --name test -target wasm32-wasi --test-cmd wasmtime --test-cmd --dir=. --test-cmd-bin -I /home/mathetake/zig/test --zig-lib-dir /home/mathetake/zig/lib 
/home/mathetake/zig/zig-cache/o/50a49ecd131d5c652f1182575d182f4d/build /home/mathetake/zig/build/zig /home/mathetake/zig /home/mathetake/zig/zig-cache /home/mathetake/.cache/zig test-std -Dtarget=wasm32-wasi -Denable-wasmtime
```

mathetake:
OK a bit of progress.. looks like linking with libc has not succeeded yet. 
```
$ wasm-objdump /home/mathetake/zig/zig-cache/o/3f9eec2bc857dd8c7327f378ef491e47/test.wasm --section=import -x
...
 - func[6] sig=6 <exit|c> <- c.exit
 - func[7] sig=0 <malloc|c> <- c.malloc
 - func[8] sig=6 <free|c> <- c.free
 - func[9] sig=8 <write|c> <- c.write
```

mathetake:
looks like we have to remove `"c"` from here https://github.com/ziglang/zig/blob/master/lib/std/c.zig#L71-L86

mathetake:
OK now `zig-bootstrap/out/host/bin/zig build-exe main.zig -lc --single-threaded -target wasm32-wasi` has started working!

```zig
const std = @import("std");
pub fn main() void {
    std.debug.print("hello with libc!", .{});
}
```

lib/std/c.zig
```diff
@@ -68,115 +69,219 @@ pub fn versionCheck(glibc_version: builtin.Version) type {
...
-pub extern "c" fn renameat(olddirfd: fd_t, old: [*:0]const u8, newdirfd: fd_t, new: [*:0]const u8) c_int;
-pub extern "c" fn chdir(path: [*:0]const u8) c_int;
-pub extern "c" fn fchdir(fd: fd_t) c_int;
-pub extern "c" fn execve(path: [*:0]const u8, argv: [*:null]const ?[*:0]const u8, envp: [*:null]const ?[*:0]const u8) c_int;
-pub extern "c" fn dup(fd: fd_t) c_int;
-pub extern "c" fn dup2(old_fd: fd_t, new_fd: fd_t) c_int;
-pub extern "c" fn readlink(noalias path: [*:0]const u8, noalias buf: [*]u8, bufsize: usize) isize;
-pub extern "c" fn readlinkat(dirfd: fd_t, noalias path: [*:0]const u8, noalias buf: [*]u8, bufsize: usize) isize;
+pub usingnamespace switch (builtin.os.tag) {
+    .wasi => struct {
```
mathetake, line 73: this is necessary remove "c" from the extern statement in order to match symbols correctly with wasm-ld, but I must admit looks a bit ugly.. is there any better way to achieve this?

src/stage1/codegen.cpp
```diff
@@ -565,7 +565,7 @@ static LLVMValueRef make_fn_llvm_value(CodeGen *g, ZigFn *fn) {
 
         bool want_ssp_attrs = g->build_mode != BuildModeFastRelease &&
                               g->build_mode != BuildModeSmallRelease &&
-                              g->link_libc;
+                              g->link_libc && !target_is_wasm(g->zig_target);
```
mathetake, line 568: note: this is necessary to avoid stack protector insertion even with link_libc =true for Wasm target.

mathetake:
@kubkon I believe this is ready for review. I would appreciate it if you could have a chance to take a look! :)

lib/std/c.zig
LemonBoy: That's a no from me dawg, the problem is either with the compiler using the wrong name for externs or the libc exporting the wrong names.

src/stage1/codegen.cpp
Mouvedia: Wouldn't a comment explaining that be necessary?

lib/std/c.zig
mathetake: > Oh, it seems that the libc is exporting everything under the default env module. The library name is a bit overloaded given wasm two-level namespaces.

yes that is  the problem here.. at link time we must "import" from "env" module rather than a specific module name (here we trying to import from "c" and which results in the import statements being left and hosts are supposed to provide there at load time, rather than linktime.).

src/stage1/codegen.cpp
mathetake: https://github.com/ziglang/zig/pull/9227/commits/fd0effffc29bae7ed7491c72365b279ec3da145f done!

lib/std/c.zig
LemonBoy: I see, that's unfortunate and quite awful.
Can we avoid the duplicate declarations by exporting the whole libc symbols under the `c` module name? If we can't do any better than this I'd suggest to move all the decls into a separate file (eg. `c_wasi.zig`), the file is already quite hairy and ugly because of all the symbol renaming logic.

mathetake: yeah https://github.com/ziglang/zig/pull/9227/commits/0ed36a50f0d36a952a338adf9a0e359e8228b835 this is the best yet still ugly solution from me by putting all symbols in `c/wasi.zig` as you suggested..

lib/std/c.zig
```diff
@@ -68,324 +69,334 @@ pub fn versionCheck(glibc_version: builtin.Version) type {
...
 pub usingnamespace switch (builtin.os.tag) {
-    .macos, .ios, .watchos, .tvos => struct {
-        pub const realpath = @"realpath$DARWIN_EXTSN";
-        pub const fstatat = _fstatat;
-    },
+    // WASI-libc should have an empty library_name on extern statements for libc functions
+    // so that we don't have "wasm-import-{name,module}" on these function symbols. Otherwise,
+    // symbols are not resolved by wasm-ld at link time and supposed to be
+    // resolved at load time by runtimes. So instead we define them in "c/wasi.zig".
+    .wasi => struct {},
```
mathetake, line 77: I put a comment about why we fail to link wasi-libc with `"c"` or even `"env"` specifier on extern statements.

lib/std/os.zig
```diff
@@ -3467,7 +3467,7 @@ pub fn waitpid(pid: pid_t, flags: u32) WaitPidResult {
     }
 }
 
-pub const Stat = if (builtin.link_libc)
+pub const Stat = if (builtin.os.tag != .wasi and builtin.link_libc)
```
LemonBoy, line 3470: What?

lib/std/c.zig
LemonBoy: Uh that's weird. I'd expect `env` to work. Have you tired inspecting the LLVM IR and or wasm code?

mathetake: Actually I checked the LLVM IR and Wasm binary (before and after link), and found that 
- `wasm-import-{name,module}` is added only when we provide specifiers on extern statements as I explained in the comment, though this is somewhat expected as I understand from [the code](https://github.com/ziglang/zig/blob/master/src/stage1/codegen.cpp#L454-L456).
- import and function name sections are exactly the same between Wasm binaries with and without "env" specifiers.

Having said that, I guess this has something to do with Link metadata section(https://github.com/WebAssembly/tool-conventions/blob/master/Linking.md#linking-metadata-section) and maybe it is affected by the existence of "wasm-import-*" in the IR.

mathetake: I image `wasm-import-*` would emit some metadata to link_metadata section, and that results in telling wasm-ld not to resolve these symbols at link time.

lib/std/os.zig
mathetake: sorry I missed this I added at first for debugging. Fixed in https://github.com/ziglang/zig/pull/9227/commits/05364481e47b81ccb23799eba022adebc19f0958

lib/std/fs.zig
```diff
@@ -647,6 +647,9 @@ pub const Dir = struct {
             /// Memory such as file names referenced in this returned entry becomes invalid
             /// with subsequent calls to `next`, as well as when this `Dir` is deinitialized.
             pub fn next(self: *Self) Error!?Entry {
+                if (builtin.link_libc) {
+                    @compileError("unimplemented");
```
mathetake, line 651: this is a left TODO to finish this PR.

mathetake:
Almost there, but the fs tests haven't passed yet. I guess something to do with preopens, but couldn't figure out yet...

mathetake:
OK finally passes all the tests locally...🤞

mathetake:
OK I believe everything works now after rebasing 😄  hope all the tests would pass again 🤞

mathetake:
OK I did a bit of research on wasm-ld behavior on wasm-import-module and wasm-import-name attributes, and other frontend's behavior.. I found that the root cause of linktime symbol mismatch and maybe I could reduce the amount of change (mostly due to the workaround for it) here. Will work on a separate PR in next few days.

mathetake:
https://github.com/ziglang/zig/pull/9279 will reduce the fair amount of change in this PR.

mathetake:
rebased on https://github.com/ziglang/zig/pull/9279 .

kubkon:
If you don't mind, I'll convert this PR into a draft since it's blocked by #9279 so that we don't merge it prematurely.

mathetake:
Right, thanks for doing that for me!

kubkon:
@mathetake could you rebase/merge against master given that #9279 is now merged, and mark the PR as ready for review (if you still feel it's ready ofc!)?

mathetake:
@kubkon I've just made this for ready again:) Thanks!

lib/std/os/bits/wasi.zig
```diff
@@ -437,3 +464,23 @@ pub const whence_t = u8;
...
+pub const S_IFIFO = 0xc000;
+pub const S_IFLNK = 0xa000;
+pub const S_IFMT = S_IFBLK | S_IFCHR | S_IFDIR | S_IFIFO | S_IFLNK | S_IFREG | S_IFSOCK;
+pub const S_IFREG = 0x8000;
+// There's no concept of UNIX domain socket but we define this value here in order to line with other OSes.
+pub const S_IFSOCK = 0x1;
+
+pub const SEEK_SET = 0x0; // = __WASI_WHENCE_SET
+pub const SEEK_CUR = 0x1; // = __WASI_WHENCE_CUR
+pub const SEEK_END = 0x2; // = __WASI_WHENCE_END
```
kubkon, line 481: Would it be possible to reuse already defined values here rather than setting them manually (as ints) and adding a comment? Something like:

```zig
pub const SEEK_SET = WHENCE_SET;
pub const SEEK_CUR = WHENCE_CUR;
pub const SEEK_END = WHENCE_END;
```

lib/std/os/bits/wasi.zig
```diff
@@ -271,11 +276,33 @@ pub const linkcount_t = u64;
...
+    pub const O_APPEND = 1 << 0; // = __WASI_FDFLAGS_APPEND
+    pub const O_CLOEXEC = (0);
+    pub const O_CREAT = ((1 << 0) << 12); // = __WASI_OFLAGS_CREAT << 12
+    pub const O_DIRECTORY = ((1 << 1) << 12); // = __WASI_OFLAGS_DIRECTORY << 12
+    pub const O_DSYNC = (1 << 1); // = __WASI_FDFLAGS_DSYNC
+    pub const O_EXCL = ((1 << 2) << 12); // = __WASI_OFLAGS_EXCL << 12
+    pub const O_EXEC = (0x02000000);
+    pub const O_NOCTTY = (0);
+    pub const O_NOFOLLOW = (0x01000000);
+    pub const O_NONBLOCK = (1 << 2); // = __WASI_FDFLAGS_NONBLOCK
```
kubkon, line 290: Is it possible to do

```zig
pub const O_NONBLOCK = FDFLAG_NONBLOCK;
```

And for any other suitable occurrence?

lib/std/os/bits/wasi.zig
mathetake: done: https://github.com/ziglang/zig/pull/9227/commits/7814a2bd4a3ec22cd9548c622f7dc837dba968f7

lib/std/os/bits/wasi.zig
mathetake: done: https://github.com/ziglang/zig/pull/9227/commits/7814a2bd4a3ec22cd9548c622f7dc837dba968f7

lib/std/fs.zig
```diff
@@ -1084,14 +1089,16 @@ pub const Dir = struct {
...
-        if (!has_flock_open_flags and flags.lock != .None) {
-            // TODO: integrate async I/O
-            const lock_nonblocking = if (flags.lock_nonblocking) os.LOCK_NB else @as(i32, 0);
-            try os.flock(fd, switch (flags.lock) {
-                .None => unreachable,
-                .Shared => os.LOCK_SH | lock_nonblocking,
-                .Exclusive => os.LOCK_EX | lock_nonblocking,
-            });
+        if (builtin.target.os.tag != .wasi) {
+            if (!has_flock_open_flags and flags.lock != .None and builtin.target.os.tag != .wasi) {
```
kubkon, line 1093:
```suggestion
        if (!has_flock_open_flags and flags.lock != .None and builtin.target.os.tag != .wasi) {
```

I don't think the outer check is needed here.

lib/std/fs.zig
```diff
@@ -934,14 +937,16 @@ pub const Dir = struct {
...
-        if (!has_flock_open_flags and flags.lock != .None) {
-            // TODO: integrate async I/O
-            const lock_nonblocking = if (flags.lock_nonblocking) os.LOCK_NB else @as(i32, 0);
-            try os.flock(fd, switch (flags.lock) {
-                .None => unreachable,
-                .Shared => os.LOCK_SH | lock_nonblocking,
-                .Exclusive => os.LOCK_EX | lock_nonblocking,
-            });
+        if (builtin.target.os.tag != .wasi) {
+            if (!has_flock_open_flags and flags.lock != .None) {
```
kubkon, line 941:
```suggestion
        if (!has_flock_open_flags and flags.lock != .None and builtin.target.os.tag != .wasi) {
```

I think we can put all checks in one `if`.

lib/std/fs/wasi.zig
```diff
@@ -169,7 +169,7 @@ pub const PreopenList = struct {
 };
 
 test "extracting WASI preopens" {
-    if (std.builtin.os.tag != .wasi) return error.SkipZigTest;
+    if (std.builtin.os.tag != .wasi or @import("builtin").link_libc) return error.SkipZigTest;
```
kubkon, line 172:
```suggestion
    if (std.builtin.os.tag != .wasi or std.builtin.link_libc) return error.SkipZigTest;
```

lib/std/testing.zig
```diff
@@ -4,6 +4,8 @@
 // The MIT license requires this copyright notice to be included in all copies
 // and substantial portions of the software.
 const std = @import("std.zig");
+const builtin = @import("builtin");
```
kubkon, line 7: I don't think that's needed. `std.builtin` re-exports all relevant build checks.
```suggestion
```

lib/std/testing.zig
```diff
@@ -326,7 +328,7 @@ pub const TmpDir = struct {
 };
 
 fn getCwdOrWasiPreopen() std.fs.Dir {
-    if (std.builtin.os.tag == .wasi) {
+    if (std.builtin.os.tag == .wasi and !builtin.link_libc) {
```
kubkon, line 331:
```suggestion
    if (std.builtin.os.tag == .wasi and !std.builtin.link_libc) {
```

kubkon (changes requested):
Thanks for addressing the previous set of suggestions, and I apologise for not spending more time on reviewing this PR earlier. Anyway, looks good! I've just got a final set of changes I'd like you to incorporate and we should be ready to merge!

lib/std/testing.zig
mathetake: https://github.com/ziglang/zig/pull/9227/commits/ca4898c0f3590832d7a2ef84e8f4906e89d7f34e done!

lib/std/testing.zig
mathetake: https://github.com/ziglang/zig/pull/9227/commits/ca4898c0f3590832d7a2ef84e8f4906e89d7f34e done!

lib/std/fs/wasi.zig
mathetake: https://github.com/ziglang/zig/pull/9227/commits/ca4898c0f3590832d7a2ef84e8f4906e89d7f34e done!

lib/std/fs.zig
mathetake: yeah this is a little bit trick.. we need to remove out this if block since since the expression is not compiltime-known, and WASI doesn't have `os.flock` that's why I put the OS check prior to the if block. Added comment about this: https://github.com/ziglang/zig/pull/9227/commits/68617c9fb05ad7e92315edb0dec7c5fad56b6265

lib/std/fs.zig
mathetake: https://github.com/ziglang/zig/pull/9227#discussion_r688492678 same here

mathetake:
Thanks @kubkon always! I hope I addressed all your comments:)

lib/std/fs.zig
kubkon: Ah gotcha, makes sense!

kubkon (approved):

PR #9229, opened 2021-06-24
zig ld: link Obj-C, link frameworks, improve linker's implementation

kubkon:
## General gist of the PR

* add support for Obj-C  - accept `.m` sources and link them correctly with `zig ld`
* (finally) link against frameworks
* clean up a lot of redundant codepaths in linker's implementation making it easier to contribute to by other (I really do hope!)

## In more detail

This now works on macOS **without** the system linker hack (cc @floooh, @prime31, @mikdusan):

```
zig cc helloworld.m -framework Foundation -o helloworld
```

and produces a functional binary with all frameworks and dylibs (here, `libobjc.dylib`, `Foundation.dylib`, `CoreFoundation.dylib` and of course `libSystem.dylib`) autoresolved - the linker when resolving dylibs from tbd stubs now automatically descends the re-exported tree and looks for dependent libraries meaning it's enough to just link with the top framework, e.g., Foundation. This effectively mimics the behaviour of `ld64` making `zig ld` a true linker replacement for the end-user.

Scanning of dependent libs is currently missing for binary dylibs (as opposed to tbd stubs), and the linker doesn't yet track the visited dylibs meaning it may visit a linked dylib more than once if it is referenced in multiple linked frameworks. Oh, and we don't detect cycles which we probably should for malformed dynamic libraries.

I have also exposed `zig cc` to Obj-C sources, i.e., it accepts `.m` files; however, given that @redj started working on a more general exposition in https://github.com/ziglang/zig/pull/9029, I'm happy to reverse the change, or whatever is needed. Exposing this option allowed me to test linking Obj-C with Zig in a systematic and simple manner hence why it's included in this PR.


andrewrk (approved):
Really nice work!

redj:
Thanks @kubkon, I don't mind re-basing.

PR #9230, opened 2021-06-25
allow json scientific notation to coerce to integers

elerch:
Json parsing currently returns UnexpectedToken when setting an int/comptime int field if a json number has a decimal or exponent attached. Thus it cannot handle numbers represented as 10.0 or 4.2e2, even though these numbers are perfectly happy to be ints.

I've left the is_integer check as this is a useful performance optimization, but in the case that a decimal or e value occurs (which is actually what is_integer is telling us), I parse the number instead as an f128 value, round the value and compare the result. If they're identical, we can then convert to the destination int/comptime int type. If not, we have an attempt to coerce an actual decimal number to float and can return an error as before.

I've selected error.InvalidNumber for the error as this was used elsewhere already and it seemed appropriate. As I'm changing the error value from UnexpectedToken -> error.InvalidNumber, this is a breaking change for clients of the parser, albeit a minor one.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1551,8 +1551,11 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options: 
                 .Number => |n| n,
                 else => return error.UnexpectedToken,
             };
-            if (!numberToken.is_integer) return error.UnexpectedToken;
-            return try std.fmt.parseInt(T, numberToken.slice(tokens.slice, tokens.i - 1), 10);
+            if (numberToken.is_integer)
+                return try std.fmt.parseInt(T, numberToken.slice(tokens.slice, tokens.i - 1), 10);
+            const float = try std.fmt.parseFloat(f128, numberToken.slice(tokens.slice, tokens.i - 1));
+            if (std.math.round(float) != float) return error.InvalidNumber;
+            return @floatToInt(T, float);
         },
         .Optional => |optionalInfo| {
             if (token == .Null) {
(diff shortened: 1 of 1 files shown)


lib/std/json.zig
```diff
@@ -1551,8 +1551,11 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options:
                 .Number => |n| n,
                 else => return error.UnexpectedToken,
             };
-            if (!numberToken.is_integer) return error.UnexpectedToken;
-            return try std.fmt.parseInt(T, numberToken.slice(tokens.slice, tokens.i - 1), 10);
+            if (numberToken.is_integer)
+                return try std.fmt.parseInt(T, numberToken.slice(tokens.slice, tokens.i - 1), 10);
+            const float = try std.fmt.parseFloat(f128, numberToken.slice(tokens.slice, tokens.i - 1));
```
LemonBoy, line 1556: Json numbers are `f64`, why is `f128` used here?
I'd expect the function to respect the user-specified type `T` and try parsing the number as a fp-type with the same number of bits, not `f128`.

lib/std/json.zig
```diff
@@ -1551,8 +1551,11 @@ fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options:
...
+            if (std.math.round(float) != float) return error.InvalidNumber;
+            return @floatToInt(T, float);
```
LemonBoy, line 1558: In case of overflow `@floatToInt` generates a poison value on the LLVM side, this function is unsafe and potentially dangerous. On the other hand if the safety checks fail the result is a hard panic that cannot be recovered from, this failure mode is not acceptable for a json parser.

lib/std/json.zig
daurnimator: > Json numbers are `f64`, why is `f128` used here?

huh? json numbers are not specified to have any precision: many people use them to arbitrary precision.

LemonBoy: JSON uses doubles for numbers, the exponential notation means the string is formatted as a floating point number hence the limited precision.
AFAIK JSON doesnt mandate the use of arbitrarily big fp types.

PR #9243, opened 2021-06-26
stdlib: add std.os.argc

0kalekale:
Populated by startup code before main().
for obtaining the process argument count.


LemonBoy:
That's `std.os.argv.len`, not that useful by itself.

Vexu:
As LemonBoy said just use `std.os.argv.len`.

PR #9245, opened 2021-06-26
std/fmt: add fmtDurationSigned

ifreund:
When working with durations it often makes sense to use signed integers
and allow negative durations, and there is currently no nice way to
format these in std.fmt. This patch adds a simple wrapper for the
existing fmtDurtion to fit this need.

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -1461,6 +1461,87 @@ test "fmtDuration" {
     }
 }
 
+fn formatDurationSigned(ns: i64, comptime fmt: []const u8, options: std.fmt.FormatOptions, writer: anytype) !void {
+    if (ns < 0) {
+        try writer.writeByte('-');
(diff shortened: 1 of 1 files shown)


lib/std/fmt.zig
```diff
@@ -1461,6 +1461,88 @@ test "fmtDuration" {
     }
 }
 
+fn formatDurationSigned(ns: i64, comptime fmt: []const u8, options: std.fmt.FormatOptions, writer: anytype) !void {
+    _ = fmt;
+    _ = options;
+    if (ns < 0) {
+        try writer.print("-{}", .{std.fmt.fmtDuration(@intCast(u64, -ns))});
```
LemonBoy, line 1468: Emit the Minus sign if needed and then call `formatDuration` to avoid the overhead of the double fmt parsing.

ifreund: Good point, I took this approach of going through `fmtDuration` in my code as `formatDuration` isnt `pub`.

PR #8228, opened 2021-03-12
Clean up debug.zig, particularly stack trace formating

rgreenblatt:
~~The idea with this PR is to allow for overriding at a more granular level than just overriding panic. This is useful for the freestanding context so you don't have to implement all of the necessary `os` functions. It also might be useful when on a supported OS because it allows for more precise overriding. This should allow for easier usage of `std.testing`, `GeneralPurposeAllocator`, and `panicExtra` in the freestanding context.~~

This PR is now 95% clean up of the debug interface and only adds the ability to overriding the mapping of addresses to symbols and how stack traces are collected. This will allow for easer usage of `std.testing` and `panicExtra`, but not `GeneralPurposeAllocator`.

~~See https://github.com/ziglang/zig/pull/8228#issuecomment-797823004 for what a full config looks like.~~

I have also heavily refactored how `std.debug` is layed out including removing `printSourceAtAddress`. ~~I believe there are no other changes (other than additions) to the public interface of std.debug, but I could be wrong.~~ This PR also changes the name of `getStderrMutex` to `getPrintMutex` ~~(because the writer can now be overriden)~~, makes it so that `captureStackTrace` takes an allocator and can return an error, and changes the inputs of `panicExtra`. I also cleaned up a bunch of duplicated code related to dumping stack traces.

~~This functions adds the ability to override the writer used by `std.debug`. As such, we must also add the capacity to override `detectTTYConfig`. ~~I went ahead and also added these capabilites to `std.log`.~~ See here: https://github.com/ziglang/zig/pull/8228#discussion_r593950582~~

Related: https://github.com/ziglang/zig/issues/8214

TODO:
- [ ] Misc design issues (naming, overrides, etc.)
- [x] Fixing existing code in the std library which this PR breaks.
- [x] Testing (I have done manual testing, but maybe some actual unit tests would be a good idea?)
- [x] Docs

PR #9247, opened 2021-06-26
Fix issue where root.os.panic could return

rgreenblatt:
This change forces root.os.panic to have the `PanicFn` type. I think this also simplifies the code somewhat.

This could break existing code somewhere.

Edit with more detail on the issue this PR solves:
The issue is that `root.os.panic` was used without checking the type of the function (specifically, without checking the type is `PanicFn`). `PanicFn` has return type `noreturn`, but because `root.os.panic` isn't checked, the function could have return type `void`. For this reason, there was `unreachable` after `root.os.panic`. So if you accidently wrote a panic handler which returned, then you would get issues at runtime (hitting the unreachable and causing another panic or undefined behavior) instead of a compiler error. 

Additionally, the code for using the override `root.os.panic` was different than the code for using the override `root.panic` for no good reason.

### DIFF
--- a/lib/std/builtin.zig
+++ b/lib/std/builtin.zig
@@ -671,7 +671,12 @@ pub const PanicFn = fn ([]const u8, ?*StackTrace) noreturn;
 
 /// This function is used by the Zig language code generation and
 /// therefore must be kept in sync with the compiler implementation.
-pub const panic: PanicFn = if (@hasDecl(root, "panic")) root.panic else default_panic;
+pub const panic: PanicFn = if (@hasDecl(root, "panic"))
+    root.panic
+else if (@hasDecl(root, "os") and @hasDecl(root.os, "panic"))
+    root.os.panic
+else
+    default_panic;
 
 /// This function is used by the Zig language code generation and
 /// therefore must be kept in sync with the compiler implementation.
@@ -683,10 +688,6 @@ pub fn default_panic(msg: []const u8, error_return_trace: ?*StackTrace) noreturn
         while (true) {
             @breakpoint();
         }
-    }
-    if (@hasDecl(root, "os") and @hasDecl(root.os, "panic")) {
-        root.os.panic(msg, error_return_trace);
-        unreachable;
     }
     switch (os.tag) {
         .freestanding => {


lib/std/builtin.zig
```diff
@@ -664,16 +664,17 @@ pub const PanicFn = fn ([]const u8, ?*StackTrace) noreturn;
...
+    root.os.panic
+else
+    default_panic;
 
 /// This function is used by the Zig language code generation and
 /// therefore must be kept in sync with the compiler implementation.
 pub fn default_panic(msg: []const u8, error_return_trace: ?*StackTrace) noreturn {
     @setCold(true);
-    if (@hasDecl(root, "os") and @hasDecl(root.os, "panic")) {
-        root.os.panic(msg, error_return_trace);
```
matu3ba, line 674: Why did you remove the error return trace here?

lib/std/builtin.zig
```diff
@@ -664,16 +664,17 @@ pub const PanicFn = fn ([]const u8, ?*StackTrace) noreturn;
 
 /// This function is used by the Zig language code generation and
 /// therefore must be kept in sync with the compiler implementation.
-pub const panic: PanicFn = if (@hasDecl(root, "panic")) root.panic else default_panic;
+pub const panic: PanicFn = if (@hasDecl(root, "panic"))
+    root.panic
+else if (@hasDecl(root, "os") and @hasDecl(root.os, "panic"))
+    root.os.panic
```
matu3ba, line 670: This case has no cold cache hint in contrast to the original code.

matu3ba (changes requested):
Please link the related issue in this PR with a proper description.

lib/std/builtin.zig
rgreenblatt: I don't understand. `@setCold(true)` is still the first line in `default_panic`.

Perhaps you aren't recognizing that this if statement is comptime? Or maybe I am totally missing your point?

lib/std/builtin.zig
rgreenblatt: This code isn't needed because default_panic is no longer used when `root.os.panic` is defined.

rgreenblatt:
> Please link the related issue in this PR with a proper description.

This is a (small) subset of https://github.com/ziglang/zig/pull/8228 which I have extracted out. There isn't a corresponding issue.

The aim is just to properly typecheck `root.os.panic`.

lib/std/builtin.zig
matu3ba: my bad, you are correct here.

lib/std/builtin.zig
matu3ba: Currently the language reference does not mention os-specific panic handler or what the expected behavior is, so I feel unsure about this. 
I would suggest changing doc.langref.html.in starting at `{#header_open|@panic#}` to force clarification on this or ping/talk to somebody who knows how it stuff was agreed to work and quote it.

matu3ba:
Sorry to be picky, but ideally you would give a case where this can happen. Maybe a combination of custom and/or default panic handler?

rgreenblatt:
> Sorry to be picky, but ideally you would give a case where this can happen. Maybe a combination of custom and/or default panic handler?

The issue is that `root.os.panic` was used without checking the type of the function (specifically, without checking the type is `PanicFn`). `PanicFn` has return type `noreturn`, but because `root.os.panic` isn't checked, the function could have return type `void`. For this reason, there was `unreachable` after `root.os.panic`. So if you accidently wrote a panic handler which returned, then you would get issues at runtime instead of a compiler error. 

Additionally, the code for using the override `root.os.panic` was different than the code for using the override `root.panic` for no good reason.

This PR addresses both of these issues. Sorry about not clarifying exactly what this PR does and why.

lib/std/builtin.zig
rgreenblatt: Yeah unfortunately all of this is undocumented. In theory I think the docs would go here: https://ziglang.org/documentation/master/#Root-Source-File

Issue #8214, opened 2021-03-11
std.debug.panic should respect root.panic


fengb:
Right now the panic dependency graph is as follows:
![image](https://user-images.githubusercontent.com/219422/110859470-159dfe80-8281-11eb-8529-c9236810d110.png)

`std.debug.panic()` _**ignores**_ the overridden root panic handler and always calls `os.abort()`

### Thoughts on a fix:
1. Point `std.debug.panic` to `builtin.panic`
    - can we update `panic()` args to match format string?
    - or is there a nice workaround to convert the arguments?
2. Make panicExtra less of a footgun
    - either change the body so it only dumps the data (e.g. reusable in a custom panic handler)
    - or update the function name to indicate it's going to os.abort

rgreenblatt:
I posted an idea related to this on discord.  Here is my (slightly updated) post:

> One annoyance (at least for me) right now is that approach to overriding panic and stack trace printing isn't very straighforward, particularly for freestanding.
> 
> You can override `panic` easily, but in order to get an implementation of `panicExtra` on freestanding you need:
>  - `openSelfDebugInfo`. 
>  - `isatty`
>  - `stderr` (which implies file descriptors, at least fake ones to just support `stderr`)
>  - probably some other stuff I am missing
> 
> There is an implementation of `format` for the builtin `StackTrace` type which can't be overriden directly, and you need all of the same implementations from panicExtra (other than `stderr`) to use it (they both end up calling `std.debug.writeStackTrace`).
> 
> `GeneralPurposeAllocator` (which is _supppppper_ nice btw) uses the format implemention of `StackTrace`, so using that allocater requires all of the previously mentioned implementations.
> The functions in `std.testing` use `std.debug.panic`, so they also require the same (when doing integeration/runtime testing).
> 
> 
> My purposed simplification would be something like the following:
>  - `panicExtra` should use a writer returned by `getDebugWriter` (name in flux) which can be overriden.
>  - you can override a mapping from addresses to symbols (name/impl in flux)
>  - you can override printLineFromFileAnyOs
> 
> With these changes, all that you would have to implement to get access to `testing`, `GeneralPurposeAllocator`, and `panicExtra` would be `getDebugWriter`, the mapping from address to symbols, `printLineFromFileAnyOs`,  and `os.abort`

I am drafting this right now. 

It seems to me like there are multiple different types of things you might want to override with `panic`:
 - the writer (handled by above idea)
 - symbols (handled by above idea)
[...]

PR #9248, opened 2021-06-26
improve panic hierarchy by always using builtin.panic

rgreenblatt:
This PR makes all of the panic functions (`@panic`, `std.debug.panicExtra`, and `std.debug.panic`) use `builitin.panic` which is either the default implementation or a user override.

- What used to be called panicExtra is now called panicImpl. This ~~really~~ maybe should only be accessable by the standard library, but that isn't currently possible from my understanding (without moving it to a separate file which would be gross).
- I have created a new function called panicExtra which basically does what panicExtra used to do, but it doesn't take a first_trace_addr instead just taking trace, format, and args. This function formats using a buffer and calls builtin.panic (the actual panic implementation). std.debug.panic calls panicExtra with a null trace. This panicExtra function is useful in cases where you want to print out an @errorReturnTrace and also format some args.

One issue with this approach is that there isn't a way to panic with a `first_trace_addr`. `panicExtra` used to take this as an argument, but this isn't one of the arguments to `builtin.panic`. 

This was originally implemented in https://github.com/ziglang/zig/pull/8228, particularly see https://github.com/ziglang/zig/pull/8228#issuecomment-803049027 and the surronding conversation.

Resolves https://github.com/ziglang/zig/issues/8214

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -233,9 +233,32 @@ pub fn assert(ok: bool) void {
 
 pub fn panic(comptime format: []const u8, args: anytype) noreturn {
     @setCold(true);
-    // TODO: remove conditional once wasi / LLVM defines __builtin_return_address
-    const first_trace_addr = if (native_os == .wasi) null else @returnAddress();
-    panicExtra(null, first_trace_addr, format, args);
+
+    panicExtra(null, format, args);
+}
+
+/// `panicExtra` is useful when you want to print out an `@errorReturnTrace`
+/// and also print out some values.
+pub fn panicExtra(
+    trace: ?*builtin.StackTrace,
+    comptime format: []const u8,
+    args: anytype,
+) noreturn {
+    @setCold(true);
+
+    const size = 0x1000;
+    const trunc_msg = "(msg truncated)";
+    var buf: [size + trunc_msg.len]u8 = undefined;
+    // a minor annoyance with this is that it will result in the NoSpaceLeft
+    // error being part of the @panic stack trace (but that error should
+    // only happen rarely)
+    const msg = std.fmt.bufPrint(buf[0..size], format, args) catch |err| switch (err) {
+        std.fmt.BufPrintError.NoSpaceLeft => blk: {
+            std.mem.copy(u8, buf[size..], trunc_msg);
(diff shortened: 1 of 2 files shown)


lib/std/debug.zig
```diff
@@ -1624,9 +1649,14 @@ fn handleSegfaultWindowsExtra(info: *windows.EXCEPTION_POINTERS, comptime msg: u
...
-            1 => panicExtra(null, exception_address, "Segmentation fault at address 0x{x}", .{info.ExceptionRecord.ExceptionInformation[1]}),
-            2 => panicExtra(null, exception_address, "Illegal Instruction", .{}),
+            0 => panicImpl(null, exception_address, format.?),
+            1 => {
+                const format_item = "Segmentation fault at address 0x{x}";
+                var buf: [format_item.len + 64]u8 = undefined; // 64 is arbitrary, but sufficiently large
+                const to_print = std.fmt.bufPrint(buf[0..buf.len], format_item, .{info.ExceptionRecord.ExceptionInformation[1]}) catch unreachable;
+                panicImpl(null, exception_address, to_print);
+            },
+            2 => panicImpl(null, exception_address, "Illegal Instruction"),
```
rgreenblatt, line 1659: This is gross.

Possible solutions:
 - Keep this
 - make `first_trace_addr` an argument to `builtin.panic` so we don't need to use `panicImpl` here (adds a decent amount of complexity and breaks existing code)
 - don't use `exception_address`

andrewrk: What's gross? I think I'm missing something. Anyway thanks for the PR, happy to continue the discussion after merging.

rgreenblatt: In order to pass the `exception_address` we have to bypass `builtin.panic` which is potentially user provided and use the default panic implementation instead (because `builtin.panic` doesn't take a `first_trace_addr`).
This is always platform supported because this function is windows only.
So, if the end user wanted special panic behavior on windows, it wouldn't also get triggered on segfaults.

andrewrk: I see, thanks for the explanation. We'll definitely want to do a design pass over the panic infrastructure before stabilizing std.

PR #9251, opened 2021-06-26
link: Recognize -Wl,-implib,foo.lib and pass it to LLD as /implib:foo.lib.

### LINKED ISSUE #9210: Creating a DLL on Windows with `zig cc` does not produce an import library (`.lib`)
An import library is, depressingly, still required in 2021 to link against a DLL on Windows. LLD is not yet capable of linking directly against a DLL like MinGW's linker.

Normally, when you link C code with `lld-link -dll foo.obj`, you'll get both `foo.dll` and `foo.lib`. This also appears to be the case with `zig build-lib -dynamic`. However, with `zig cc -shared`, this is not the case, and `-Wl,-implib:foo.lib` does not work either.

alexrp:
Fixes #9210.

Note that this is just intended as a stopgap. The syntax isn't exactly the same as if you're working with `lld-link` directly, and we should probably think about doing this by default for DLLs.

### DIFF
--- a/src/link/Coff.zig
+++ b/src/link/Coff.zig
@@ -996,6 +996,9 @@ fn linkWithLLD(self: *Coff, comp: *Compilation) !void {
         if (self.base.options.dynamicbase) {
             try argv.append("-dynamicbase");
         }
+
+        try argv.appendSlice(self.base.options.extra_lld_args);
+
         const subsystem_suffix = ss: {
             if (self.base.options.major_subsystem_version) |major| {
                 if (self.base.options.minor_subsystem_version) |minor| {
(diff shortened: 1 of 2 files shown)


alexrp:
I don't know how decisions are made wrt what changes meet the bar for a patch release of Zig, but fwiw, this PR (and to a lesser extent #9222 and #9223) is blocking my work on [zig-msbuild-sdk](https://github.com/alexrp/zig-msbuild-sdk). So it would be great if this could make it into 0.8.1.

kkartaltepe:
I think support for the `--out-implib` syntax from `ld` can be added to this as well so it can support both calling conventions? If so I can close my parallel PR.

alexrp:
I think I like the approach in your PR better than using `lld_argv` to pass the option through. I'll leave some comments over there.

alexrp:
Closing in favor of #9347.

PR #9252, opened 2021-06-27
Implement SymbolMap interface

rgreenblatt:
It's desirable to allow users to override how addresses are mapped to symbols (`SymbolInfo`) in debug.zig. To accomplish this, a new interface is introduced: `SymbolMap`. This function used to get a new object of this interface can overriden by `root.debug_config.initSymbolMap` or by `root.os.debug.initSymbolMap`.

This is the design that was eventually decided upon in https://github.com/ziglang/zig/pull/8228. This pr is a subset of https://github.com/ziglang/zig/pull/8228.

Symbol map loading is moved from `debug.zig` to `symbol_map_unix.zig`, `symbol_map_darwin.zig`, and `symbol_map_windows.zig`. All of the functions for loading debug symbols should be exactly the same: the functions were just copy pasted with a few input and import renames. Unfortunately git isn't clever enough to recognize this in the diff (because the functions were moved). I'm not sure if there is an easy way to view the (minimal) diff of these function.

Note that the override is obtained from a config struct despite there only being a single override (the `initSymbolMap` function).  This is done because more overrides will (probably) be intruduced in a future PR (another chunk of https://github.com/ziglang/zig/pull/8228). I will wait to make this next pr until the design for this is settled.

Outstanding issues:
- [ ] The interface currently allows `anyerror`. This should maybe be narrowed down to a specific error set.
- [ ] For whatever reason, this change makes the compiler crash when building for windows :/
```
Assertion failed at /home/ryan/.cache/yay/zig-git/src/zig/src/stage1/ir.cpp:14937 in ir_analyze_container_field_ptr. This is a bug in the Zig compiler.thread 1016325 panic:
```
- [ ] Maybe `symbol_map_darwin.zig` should have a different name? (see top of file for comment)

### DIFF
--- a/lib/std/meta.zig
+++ b/lib/std/meta.zig
@@ -103,6 +103,38 @@ test "std.meta.stringToEnum" {
     try testing.expect(null == stringToEnum(E1, "C"));
 }
 
+pub fn errorInSet(err: anytype, comptime Set: type) bool {
+    const set_errs = comptime blk: {
+        const err_infos = @typeInfo(Set).ErrorSet orelse
+            return true; // global error set!
+        var errs: [err_infos.len]Set = undefined;
+        inline for (err_infos) |err_info, i| {
+            errs[i] = @field(Set, err_info.name);
+        }
+
+        break :blk errs;
+    };
+
+    for (set_errs) |set_err| {
+        if (err == @errSetCast(anyerror, set_err)) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+test "std.meta.errorInSet" {
+    testing.expect(!errorInSet(error.A, error{}));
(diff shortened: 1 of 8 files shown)


lib/std/meta.zig
```diff
@@ -103,6 +103,38 @@ test "std.meta.stringToEnum" {
     try testing.expect(null == stringToEnum(E1, "C"));
 }
 
+pub fn errorInSet(err: anytype, comptime Set: type) bool {
```
rgreenblatt, line 106: Ehh, this function isn't really needed (only used in one place and could just do the check 'by hand').

But, if this would be a useful function more generally, then maybe it is a good idea to keep it in this PR.

andrewrk:

PR #9254, opened 2021-06-27
Add fat/universal archive support to zig ld

### LINKED ISSUE #9250: zig ld: ability to link against universal static libraries
Problem: Some libraries on macOS are distributed as universal binaries that might be the only option to link against - even when building for a single architecture.

Solution: Support linking against universal binaries

Universal binaries on macOS work by embedding the data for multiple libraries (one for each arch)  into a single file - with a "fat" header to provide the offsets for each of these. 

This problem has already been solved for dynamic libraries (see #9169 & #9179) - I imagine the same solution should be extendable to static libraries & shared code factored-out.

moosichu:
Resolves #9250 

This is an extension of adding fat dylib support to zig ld, pulling out
the functionality needed to support fat headers & offsets and applying
it to zig archives.

### DIFF
--- a/src/link/MachO/Dylib.zig
+++ b/src/link/MachO/Dylib.zig
@@ -1,7 +1,6 @@
 const Dylib = @This();
 
 const std = @import("std");
-const builtin = std.builtin;
 const assert = std.debug.assert;
 const fs = std.fs;
 const fmt = std.fmt;
@@ -9,7 +8,7 @@ const log = std.log.scoped(.dylib);
 const macho = std.macho;
 const math = std.math;
 const mem = std.mem;
-const native_endian = builtin.target.cpu.arch.endian();
+const fat = @import("fat.zig");
 
 const Allocator = mem.Allocator;
 const Arch = std.Target.Cpu.Arch;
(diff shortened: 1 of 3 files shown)


moosichu:
@kubkon I managed to get cross-compilation working on my Linux machine & link with frameworks.

Please see: https://github.com/ziglang/zig/pull/9262 for a couple of other issues I had to resolve to get that running as well.

I can confirm that this PR works and does what it says on the tin.

LemonBoy:
Don't forget to run `zig fmt` on your code, the CI is failing because of that.

src/link/MachO/Dylib.zig
```diff
@@ -258,7 +225,7 @@ pub fn parse(self: *Dylib) !void {
         return error.NotDylib;
     }
 
-    const this_arch: Arch = decodeArch(self.header.?.cputype) catch |err| switch (err) {
+    const this_arch: Arch = fat.decodeArch(self.header.?.cputype, true) catch |err| switch (err) {
         error.UnsupportedCpuArchitecture => |e| {
             log.err("unsupported cpu architecture 0x{x}", .{self.header.?.cputype});
```
kubkon, line 230: Isn't it that we will duplicate the error log here given that you want `decodeArch` to print the error message in case an error?

src/link/MachO/fat.zig
```diff
@@ -0,0 +1,55 @@
...
+            return error.UnsupportedCpuArchitecture;
+        },
+    };
+    return arch;
+}
+
+fn readFatStruct(reader: anytype, comptime T: type) !T {
+    // Fat structures (fat_header & fat_arch) are always written and read to/from
+    // disk in big endian order.
+    var res: T = try reader.readStruct(T);
```
kubkon, line 27:
```suggestion
    var res = try reader.readStruct(T);
```

Just a minor nit that this type should be autoinferred.

src/link/MachO/Dylib.zig
moosichu: Yeah - that is just a plain old mistake!

PR #9258, opened 2021-06-28
Shared Cache Locking

### LINKED ISSUE #7596: caching system: support non-exclusive locks for some resources
0b46c27333960553226abd4cc7aed10e07a14b06 addresses a deadlock where:

`zig test` held an exclusive lock on compiler_rt.o but then executed a child process, which attempted to obtain an exclusive lock on compiler_rt.o. Both of these processes should be able to have the lock at the same time because neither of them are going to modify this object, it's all read-only access.

For project-local assets, because we do in-place binary patching, we grab an exclusive lock on the binary we want to patch before attempting to do it. However for global shared resources it does not make sense to do in-place binary patching. I proposed that for global resources we put everything into the cache hash, including linker options, and use a shared (read-only) lock for this.

For global resources it would look something like this:

 * try to open the manifest file with shared lock. If you get it, it's a cache hit and you have a shared lock, others can grab it too.
 * otherwise create the manifest file with an exclusive lock. this makes shared lockers wait until you finish putting the resource there, and downgrade the lock to shared, waking up the waiters who may now use the resource.

This will allow child processes to run while holding shared locks, and the behavior will be exactly what we want: prevent the cache garbage collector from reaping the resources so that their stack traces, debug info, etc will still work correctly.

The commit mentioned at the top could then be reverted, cleaning up the code, and resolving a race condition (what if the binary  gets modified in between `comp.destroy()` and the child process spawning?).

### LINKED ISSUE #9139: zig c++ hanging when invoked in parallel
I am using a combination of zig c++, golang and bazel to cross-compile a cgo program to Darwin. It compiles go stdlib in parallel and sometimes hangs.  A container with a hung `ps auxf` looks as follows:

```
root@4ba08cf5c898:/x# ps auxf | cat
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root        5093  0.4  0.0   3968  3472 pts/1    Ss   04:03   0:00 bash
root        5190  0.0  0.0   6700  2952 pts/1    R+   04:03   0:00  \_ ps auxf
root        5191  0.0  0.0   2468   516 pts/1    S+   04:03   0:00  \_ cat
root           1  0.0  0.0 709728 17392 pts/0    Ssl+ Jun16   0:01 ./bazel build -s --platforms @zig_sdk//:aarch64-macos-gnu //test:gognu
root        2423  0.0  0.0 343176 22712 pts/0    Sl+  Jun16   0:12 /root/.cache/bazelisk/downloads/bazelbuild/bazel-4.1.0-linux-x86_64/bin/bazel build -s --platforms @zig_sdk//:aarch64-macos-gnu //test:gognu
root        2428  0.5  3.1 14181376 1017740 ?    Ssl  Jun16   2:11 bazel(x) -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/root/.cache/bazel/_bazel_root/cc8755609ad61864910f145119713de9 --add-opens=java.base/java.nio=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED -Xverify:none -Djava.util.logging.config.file=/root/.cache/bazel/_bazel_root/cc8755609ad61864910f145119713de9/javalog.properties -Dcom.google.devtools.build.lib.util.LogHandlerQuerier.class=com.google.devtools.build.lib.util.SimpleLogHandler$HandlerQuerier -XX:-MaxFDLimit -Djava.library.path=/root/.cache/bazel/_bazel_root/install/f95ca91ebc34d56aa0f8ad499de91220/embedded_tools/jdk/lib/jli:/root/.cache/bazel/_bazel_root/install/f95ca91ebc34d56aa0f8ad499de91220/embedded_tools/jdk/lib:/root/.cache/bazel/_bazel_root/install/f95ca91ebc34d56aa0f8ad499de91220/embedded_tools/jdk/lib/server:/root/.cache/bazel/_bazel_root/install/f95ca91ebc34d56aa0f8ad499de91220/ -Dfile.encoding=ISO-8859-1 -jar /root/.cache/bazel/_bazel_root/install/f95ca91ebc34d56aa0f8ad499de91220/A-server.jar --max_idle_secs=10800 --noshutdown_on_low_sys_mem --connect_timeout_secs=30 --output_user_root=/root/.cache/bazel/_bazel_root --install_base=/root/.cache/bazel/_bazel_root/install/f95ca91ebc34d56aa0f8ad499de91220 --install_md5=f95ca91ebc34d56aa0f8ad499de91220 --output_base=/root/.cache/bazel/_bazel_root/cc8755609ad61864910f145119713de9 --workspace_directory=/x --default_system_javabase= --failure_detail_out=/root/.cache/bazel/_bazel_root/cc8755609ad61864910f145119713de9/failure_detail.rawproto --expand_configs_in_place --idle_server_tasks --write_command_log --nowatchfs --nofatal_event_bus_exceptions --nowindows_enable_symlinks --client_debug=false --product_name=Bazel --noincompatible_enable_execution_transition --option_sources=
root        2709  0.0  0.0   5796  1632 ?        S    Jun16   0:00  \_ /root/.cache/bazel/_bazel_root/install/f95ca91ebc34d56aa0f8ad499de91220/process-wrapper --timeout=0 --kill_delay=15 bazel-out/host/bin/external/go_sdk/builder stdlib -sdk external/go_sdk -installsuffix darwin_arm64 -out bazel-out/k8-fastbuild/bin/external/io_bazel_rules_go/stdlib_
root        2710  0.0  0.0 707676 13572 ?        Ssl  Jun16   0:01      \_ bazel-out/host/bin/external/go_sdk/builder stdlib -sdk external/go_sdk -installsuffix darwin_arm64 -out bazel-out/k8-fastbuild/bin/external/io_bazel_rules_go/stdlib_
root        2715  0.0  0.0 1159684 31992 ?       Sl   Jun16   0:03          \_ external/go_sdk/bin/go install -toolexec /root/.cache/bazel/_bazel_root/cc8755609ad61864910f145119713de9/sandbox/processwrapper-sandbox/3/execroot/bazel-zig-cc/bazel-out/host/bin/external/go_sdk/builder filterbuildid -gcflags=all= -ldflags=all=-trimpath /root/.cache/bazel/_bazel_root/cc8755609ad61864910f145119713de9/sandbox/processwrapper-sandbox/3/execroot/bazel-zig-cc -asmflags=all=-trimpath /root/.cache/bazel/_bazel_root/cc8755609ad61864910f145119713de9/sandbox/processwrapper-sandbox/3/execroot/bazel-zig-cc/bazel-out/k8-fastbuild/bin/external/io_bazel_rules_go/stdlib_ std runtime/cgo
root        5087  0.0  0.0 152932 30948 ?        Sl   Jun16   0:00              \_ /root/.cache/bazel/_bazel_root/cc8755609ad61864910f145119713de9/external/zig_sdk/zig c++ -I /root/.cache/bazel/_bazel_root/cc8755609ad61864910f145119713de9/sandbox/processwrapper-sandbox/3/execroot/bazel-zig-cc/bazel-out/k8-fastbuild/bin/external/io_bazel_rules_go/stdlib_/src/net -fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build2201875783/b087=/tmp/go-build -gno-record-gcc-switches -fno-common -o /tmp/go-build2201875783/b087/_cgo_.o /tmp/go-build2201875783/b087/_cgo_main.o /tmp/go-build2201875783/b087/_x001.o /tmp/go-build2201875783/b087/_x002.o /tmp/go-build2201875783/b087/_x003.o /tmp/go-build2201875783/b087/_x004.o /tmp/go-build2201875783/b087/_x005.o -target aarch64-macos-gnu
root        5088  0.0  0.0 152932 31092 ?        Sl   Jun16   0:00              \_ /root/.cache/bazel/_bazel_root/cc8755609ad61864910f145119713de9/external/zig_sdk/zig c++ -I /root/.cache/bazel/_bazel_root/cc8755609ad61864910f145119713de9/sandbox/processwrapper-sandbox/3/execroot/bazel-zig-cc/bazel-out/k8-fastbuild/bin/external/io_bazel_rules_go/stdlib_/src/os/user -fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build2201875783/b036=/tmp/go-build -gno-record-gcc-switches -fno-common -o /tmp/go-build2201875783/b036/_cgo_.o /tmp/go-build2201875783/b036/_cgo_main.o /tmp/go-build2201875783/b036/_x001.o /tmp/go-build2201875783/b036/_x002.o /tmp/go-build2201875783/b036/_x003.o /tmp/go-build2201875783/b036/_x004.o -target aarch64-macos-gnu
```

Both are waiting on some lock (pids are different, since I am stracing outside the container):
```
motiejus ~/code/bazel-zig-cc $ sudo strace -p 812414
strace: Process 812414 attached
futex(0x7f3743028b70, FUTEX_WAIT_PRIVATE, 4294967295, NULL^Cstrace: Process 812414 detached
 <detached ...>

motiejus ~/code/bazel-zig-cc $ sudo strace -p 812415
strace: Process 812415 attached
futex(0x7f12064f8b70, FUTEX_WAIT_PRIVATE, 4294967295, NULL^Cstrace: Process 812415 detached
 <detached ...>
```

`kill -USR1` did not produce a stack trace. Is there any more information I can provide? Steps to reproduce on a x86_64-linux machine with a working docker installation:

```
$ git clone https://git.sr.ht/~motiejus/bazel-zig-cc -b hangzig
$ cd bazel-zig-cc
$ for i in $(seq 1000); do date; echo $i; time ./hangzig; done
```

It fails more often in builds.sr.ht (therefore the test script has `--cpuset-cpus=0-1`, because builds.sr.ht allocates 2 CPUs), e.g. https://builds.sr.ht/~motiejus/job/526372. On my laptop it failed on the 15'th iteration, an iteration is ~90 seconds.

zig version: 0.9.0-dev.137+86ebd4b97. I know bazel in the loop is cumbersome, but I wasn't able to find an easy way to reproduce it without it.

### LINKED ISSUE #9187: Apparent compiler deadlock when compiling multiple separate projects in parallel
This is happening in [`Zig.Sdk`](https://github.com/alexrp/zig-msbuild-sdk), so this is easiest to reproduce if you're willing to [install the .NET SDK](https://dotnet.microsoft.com/download/dotnet/6.0).

Basically, do this:

```console
$ git clone git@github.com:alexrp/zig-msbuild-sdk.git
$ dotnet build src/sdk
$ dotnet build src/samples -v:n -p:BuildInParallel=true
```

That last `dotnet build` invocation will sometimes hang. So far, it always seems to happen either when building two executables or two dynamic libraries at the same time; building an executable and a dynamic library at the same time appears to be fine. But this might be a red herring.

```
Attempting to cancel the build...
     5>C:\Users\alex\source\repos\zig-msbuild-sdk\pkg\cache\zig.sdk\1.0.1-dev.g5da2354e66\build\Zig.Sdk.Build.targets(7,9): warning MSB5021: Terminating the task executable "zig" and its child processes because the build was canceled. [C:\Users\alex\source\repos\zig-msbuild-sdk\src\samples\clib\clib.cproj]
     3>C:\Users\alex\source\repos\zig-msbuild-sdk\pkg\cache\zig.sdk\1.0.1-dev.g5da2354e66\build\Zig.Sdk.Build.targets(7,9): warning MSB5021: Terminating the task executable "zig" and its child processes because the build was canceled. [C:\Users\alex\source\repos\zig-msbuild-sdk\src\samples\cxxlib\cxxlib.cxxproj]
```

```
Attempting to cancel the build...
     5>C:\Users\alex\source\repos\zig-msbuild-sdk\pkg\cache\zig.sdk\1.0.1-dev.g5da2354e66\build\Zig.Sdk.Build.targets(7,9): warning MSB5021: Terminating the task executable "zig" and its child processes because the build was canceled. [C:\Users\alex\source\repos\zig-msbuild-sdk\src\samples\cxxlib\cxxlib.cxxproj]
     7>C:\Users\alex\source\repos\zig-msbuild-sdk\pkg\cache\zig.sdk\1.0.1-dev.g5da2354e66\build\Zig.Sdk.Build.targets(7,9): warning MSB5021: Terminating the task executable "zig" and its child processes because the build was canceled. [C:\Users\alex\source\repos\zig-msbuild-sdk\src\samples\clib\clib.cproj]
```

```
Attempting to cancel the build...
     2>C:\Users\alex\source\repos\zig-msbuild-sdk\pkg\cache\zig.sdk\1.0.1-dev.g5da2354e66\build\Zig.Sdk.Build.targets(7,9): warning MSB5021: Terminating the task executable "zig" and its child processes because the build was canceled. [C:\Users\alex\source\repos\zig-msbuild-sdk\src\samples\cexe\cexe.cproj]
     3>C:\Users\alex\source\repos\zig-msbuild-sdk\pkg\cache\zig.sdk\1.0.1-dev.g5da2354e66\build\Zig.Sdk.Build.targets(7,9): warning MSB5021: Terminating the task executable "zig" and its child processes because the build was canceled. [C:\Users\alex\source\repos\zig-msbuild-sdk\src\samples\cxxexe\cxxexe.cxxproj]
```

Note: `Zig.Sdk` normally fetches the Zig compiler (0.8.0) from NuGet but it can be overridden to use a local build by doing `dotnet build src/samples -v:n -p:BuildInParallel -p:ZigExePath="path/to/zig"`. Also, I've seen this happen on all platforms (not just Windows) in CI.

andrewrk:
The cache system now drops from exclusive locks to shared locks after the artifact is built. Additionally, when checking the cache, if an exclusive lock cannot be obtained, a shared lock will be obtained instead. Together, this solves a design flaw in the caching system that was causing deadlocks, and allows multiple processes to share the same read-only build artifacts, while still allowing a cache garbage collection utility to avoid deleting files which are being actively used.

This required new std lib functionality: `std.fs.File.setLock`. This PR cannot be merged until the Windows implementation is completed. This may require switching strategies to using `NtLockFile` and `NtUnlockFile` rather than relying on ShareAccess flags and manual polling.

I was able to reproduce the deadlock problems reported in #9139 and observed this change fixed the problem. It also allowed me to delete another workaround to a deadlock when supplying the same source file multiple times.

Closes #7596
Fixes #9139
Fixes #9187

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -883,24 +883,39 @@ pub const Dir = struct {
     /// [WTF-16](https://simonsapin.github.io/wtf-8/#potentially-ill-formed-utf-16) encoded.
     pub fn openFileW(self: Dir, sub_path_w: []const u16, flags: File.OpenFlags) File.OpenError!File {
         const w = os.windows;
-        return @as(File, .{
-            .handle = try os.windows.OpenFile(sub_path_w, .{
+        const file: File = .{
+            .handle = try w.OpenFile(sub_path_w, .{
                 .dir = self.fd,
                 .access_mask = w.SYNCHRONIZE |
                     (if (flags.read) @as(u32, w.GENERIC_READ) else 0) |
                     (if (flags.write) @as(u32, w.GENERIC_WRITE) else 0),
-                .share_access = switch (flags.lock) {
-                    .None => w.FILE_SHARE_WRITE | w.FILE_SHARE_READ | w.FILE_SHARE_DELETE,
-                    .Shared => w.FILE_SHARE_READ | w.FILE_SHARE_DELETE,
-                    .Exclusive => w.FILE_SHARE_DELETE,
-                },
-                .share_access_nonblocking = flags.lock_nonblocking,
                 .creation = w.FILE_OPEN,
                 .io_mode = flags.intended_io_mode,
             }),
             .capable_io_mode = std.io.default_mode,
             .intended_io_mode = flags.intended_io_mode,
-        });
+        };
+        var io: w.IO_STATUS_BLOCK = undefined;
+        const range_off: w.LARGE_INTEGER = 0;
+        const range_len: w.LARGE_INTEGER = 1;
+        const exclusive = switch (flags.lock) {
+            .None => return file,
+            .Shared => false,
+            .Exclusive => true,
+        };
+        try w.LockFile(
+            file.handle,
+            null,
+            null,
+            null,
+            &io,
+            &range_off,
+            &range_len,
+            null,
+            @boolToInt(flags.lock_nonblocking),
+            @boolToInt(exclusive),
+        );
+        return file;
     }
 
     /// Creates, opens, or overwrites a file with write access.
@@ -1019,16 +1034,10 @@ pub const Dir = struct {
     pub fn createFileW(self: Dir, sub_path_w: []const u16, flags: File.CreateFlags) File.OpenError!File {
         const w = os.windows;
         const read_flag = if (flags.read) @as(u32, w.GENERIC_READ) else 0;
-        return @as(File, .{
+        const file: File = .{
             .handle = try os.windows.OpenFile(sub_path_w, .{
                 .dir = self.fd,
                 .access_mask = w.SYNCHRONIZE | w.GENERIC_WRITE | read_flag,
-                .share_access = switch (flags.lock) {
-                    .None => w.FILE_SHARE_WRITE | w.FILE_SHARE_READ | w.FILE_SHARE_DELETE,
-                    .Shared => w.FILE_SHARE_READ | w.FILE_SHARE_DELETE,
-                    .Exclusive => w.FILE_SHARE_DELETE,
-                },
-                .share_access_nonblocking = flags.lock_nonblocking,
                 .creation = if (flags.exclusive)
                     @as(u32, w.FILE_CREATE)
                 else if (flags.truncate)
@@ -1039,7 +1048,28 @@ pub const Dir = struct {
             }),
             .capable_io_mode = std.io.default_mode,
             .intended_io_mode = flags.intended_io_mode,
-        });
+        };
+        var io: w.IO_STATUS_BLOCK = undefined;
+        const range_off: w.LARGE_INTEGER = 0;
+        const range_len: w.LARGE_INTEGER = 1;
+        const exclusive = switch (flags.lock) {
+            .None => return file,
+            .Shared => false,
+            .Exclusive => true,
+        };
+        try w.LockFile(
+            file.handle,
+            null,
+            null,
+            null,
+            &io,
+            &range_off,
+            &range_len,
+            null,
+            @boolToInt(flags.lock_nonblocking),
+            @boolToInt(exclusive),
+        );
+        return file;
     }
 
     pub const openRead = @compileError("deprecated in favor of openFile");
--- a/lib/std/fs/file.zig
+++ b/lib/std/fs/file.zig
@@ -74,17 +74,28 @@ pub const File = struct {
         read: bool = true,
         write: bool = false,
 
-        /// Open the file with a lock to prevent other processes from accessing it at the
-        /// same time. An exclusive lock will prevent other processes from acquiring a lock.
-        /// A shared lock will prevent other processes from acquiring a exclusive lock, but
-        /// doesn't prevent other process from getting their own shared locks.
+        /// Open the file with an advisory lock to coordinate with other processes
+        /// accessing it at the same time. An exclusive lock will prevent other
+        /// processes from acquiring a lock. A shared lock will prevent other
+        /// processes from acquiring a exclusive lock, but does not prevent
+        /// other process from getting their own shared locks.
         ///
-        /// Note that the lock is only advisory on Linux, except in very specific cirsumstances[1].
+        /// The lock is advisory, except on Linux in very specific cirsumstances[1].
         /// This means that a process that does not respect the locking API can still get access
         /// to the file, despite the lock.
         ///
-        /// Windows' file locks are mandatory, and any process attempting to access the file will
-        /// receive an error.
+        /// On these operating systems, the lock is acquired atomically with
+        /// opening the file:
+        /// * Darwin
+        /// * DragonFlyBSD
+        /// * FreeBSD
+        /// * Haiku
+        /// * NetBSD
+        /// * OpenBSD
+        /// On these operating systems, the lock is acquired via a separate syscall
+        /// after opening the file:
+        /// * Linux
+        /// * Windows
         ///
         /// [1]: https://www.kernel.org/doc/Documentation/filesystems/mandatory-locking.txt
         lock: Lock = .None,
@@ -120,17 +131,28 @@ pub const File = struct {
         /// `error.PathAlreadyExists` to be returned.
         exclusive: bool = false,
 
-        /// Open the file with a lock to prevent other processes from accessing it at the
-        /// same time. An exclusive lock will prevent other processes from acquiring a lock.
-        /// A shared lock will prevent other processes from acquiring a exclusive lock, but
-        /// doesn't prevent other process from getting their own shared locks.
+        /// Open the file with an advisory lock to coordinate with other processes
+        /// accessing it at the same time. An exclusive lock will prevent other
+        /// processes from acquiring a lock. A shared lock will prevent other
+        /// processes from acquiring a exclusive lock, but does not prevent
+        /// other process from getting their own shared locks.
         ///
-        /// Note that the lock is only advisory on Linux, except in very specific cirsumstances[1].
+        /// The lock is advisory, except on Linux in very specific cirsumstances[1].
         /// This means that a process that does not respect the locking API can still get access
         /// to the file, despite the lock.
         ///
-        /// Windows's file locks are mandatory, and any process attempting to access the file will
-        /// receive an error.
+        /// On these operating systems, the lock is acquired atomically with
+        /// opening the file:
+        /// * Darwin
+        /// * DragonFlyBSD
+        /// * FreeBSD
+        /// * Haiku
+        /// * NetBSD
+        /// * OpenBSD
+        /// On these operating systems, the lock is acquired via a separate syscall
+        /// after opening the file:
+        /// * Linux
+        /// * Windows
         ///
         /// [1]: https://www.kernel.org/doc/Documentation/filesystems/mandatory-locking.txt
         lock: Lock = .None,
--- a/lib/std/fs/file.zig
+++ b/lib/std/fs/file.zig
@@ -840,10 +840,37 @@ pub const File = struct {
     /// TODO: integrate with async I/O
     pub fn setLock(file: File, lock: Lock, non_blocking: bool) SetLockError!void {
         if (is_windows) {
-            @compileError("TODO implement fs.File.setLock for Windows");
+            const range_off: windows.LARGE_INTEGER = 0;
+            const range_len: windows.LARGE_INTEGER = 1;
+            const exclusive = switch (lock) {
+                .None => return windows.UnlockFile(
+                    file.handle,
+                    null,
+                    &range_off,
+                    &range_len,
+                    null,
+                ) catch |err| switch (err) {
+                    error.RangeNotLocked => return,
+                    else => |e| return e,
+                },
+                .Shared => false,
+                .Exclusive => true,
+            };
+            return windows.LockFile(
+                file.handle,
+                null,
+                null,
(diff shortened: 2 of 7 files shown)


andrewrk:
This is ready to go as soon as it passes tests. Code review welcome, especially from those knowledgeable about Windows File System APIs.

lib/std/fs.zig
```diff
@@ -883,24 +883,39 @@ pub const Dir = struct {
...
+        };
+        var io: w.IO_STATUS_BLOCK = undefined;
+        const range_off: w.LARGE_INTEGER = 0;
+        const range_len: w.LARGE_INTEGER = 1;
+        const exclusive = switch (flags.lock) {
+            .None => return file,
+            .Shared => false,
+            .Exclusive => true,
+        };
+        try w.LockFile(
```
LemonBoy, line 906: You're losing the operation atomicity by acquiring the lock after opening the file.

lib/std/fs.zig
```diff
@@ -883,24 +883,39 @@ pub const Dir = struct {
...
                 .creation = w.FILE_OPEN,
                 .io_mode = flags.intended_io_mode,
             }),
             .capable_io_mode = std.io.default_mode,
             .intended_io_mode = flags.intended_io_mode,
-        });
+        };
+        var io: w.IO_STATUS_BLOCK = undefined;
+        const range_off: w.LARGE_INTEGER = 0;
+        const range_len: w.LARGE_INTEGER = 1;
```
LemonBoy, line 900: I'd lock the whole `maxInt(LARGE_INTEGER)` range instead of a single byte only.

lib/std/fs/file.zig
```diff
@@ -829,4 +829,165 @@ pub const File = struct {
     pub fn seekableStream(file: File) SeekableStream {
         return .{ .context = file };
     }
+
+    const range_off: windows.LARGE_INTEGER = 0;
```
LemonBoy, line 833: If those two fields are still needed you can slim down the structure by making them void on platforms that are not windows.

lib/std/fs/file.zig
```diff
@@ -829,4 +829,165 @@ pub const File = struct {
...
+                .Shared => false,
+                .Exclusive => true,
+            };
+            return windows.LockFile(
+                file.handle,
+                null,
+                null,
+                null,
+                &io_status_block,
+                &range_off,
```
LemonBoy, line 861: Isn't this missing a `file.`? As suggested above why not lock/unlock the whole maxInt range?

lib/std/fs.zig
andrewrk: Thanks for the review.

Based on the docs of NtUnlockFile:

> The ZwUnlockFile routine takes a range of bytes as specified by the ByteOffset and Length arguments. This range must be identical to a range of bytes in the file that was previously locked with a single call to the ZwUnlockFile routine. It is not possible to unlock two previously locked adjacent ranges with a single call to ZwUnlockFile. It is also not possible to unlock part of a range that was previously locked with a single call to the ZwUnlockFile routine.

As far as I can tell, these act like keys in a table, rather than actual spans of bytes. So the only consideration here is to pick a range that is self-consistent as well as sensible for third party code to coordinate with if it so chooses.  Given that this PR does not add documentation that guarantees the implementation on Windows, I don't think it matters what range we pick.

lib/std/fs/file.zig
andrewrk: Not sure I understand this suggestion - did you mistake these global variables for fields?

lib/std/fs/file.zig
andrewrk: Ah I see the confusion now. The range is arbitrary and hard-coded; no need to store any data in fields. If you know something I don't know about why maxint would be any different than 1, please share.

lib/std/fs.zig
andrewrk: This is consistent with Linux; addressed this concern by updating doc comments in e32530b6a31432e43cc4d4d793796007423f2edb.

lib/std/fs.zig
LemonBoy: > As far as I can tell, these act like keys in a table, rather than actual spans of bytes.

The kernel lets you lock lock/unlock specific (and disjoint) sets of bytes, not unlike the [Posix](https://gavv.github.io/articles/file-locks/) locking facilities, so the suggestion was to lock the whole span `[0..maxInt(w.LARGE_INTEGER)]` to ensure the whole file is locked.

Picking a consistent offset/length pair to lock/unlock is fine anyway fine for the purpose of ensuring the locking is effective when using the Zig stdlib.

lib/std/fs/file.zig
LemonBoy: Yep, taking them by pointer is what confused me.

PR #9264, opened 2021-06-29
stage2: print valid filename in error messages

### LINKED ISSUE #9249: stage2 error message filenames no longer relative to CWD
With the new stage2 ast checker, error messages print filenames relative to their package root rather than relative to the CWD.  This breaks a common workflow where one can copy or sometimes click a filename in an error message to open it.

jcmoyer:
Take two. Apologies for the bad PR (#9221, which was reverted in 6d47b4f39e8cdd95ead71b1efdbf67a737174e97). I should have tested it a bit more outside of typical build contexts. This PR addresses the same issue as the previous one, but includes a couple more changes to make sure it works properly in every context.

- `Module.Scope.File.pkg` is now `null` instead of `undefined` when the file being processed is not part of any package, such as when using `zig fmt` or `zig ast-check`. `pkg` being `undefined` was the source of crashes. Since this is a case that needs to be handled differently, it makes sense to be able to tell if we can safely access `pkg` or not.
- There are a couple places that assume the `pkg` pointer is valid when called. Assertions were added in these locations, but their assumptions are left unchanged.

Fixes #9249.

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -1111,6 +1111,11 @@ pub const Scope = struct {
             return buf.toOwnedSliceSentinel(0);
         }
 
+        /// Returns the full path to this file relative to its package.
+        pub fn fullPath(file: File, ally: *Allocator) ![]u8 {
+            return file.pkg.root_src_directory.join(ally, &[_][]const u8{file.sub_file_path});
+        }
+
         pub fn dumpSrc(file: *File, src: LazySrcLoc) void {
             const loc = std.zig.findLineColumn(file.source.bytes, src);
             std.debug.print("{s}:{d}:{d}\n", .{ file.sub_file_path, loc.line + 1, loc.column + 1 });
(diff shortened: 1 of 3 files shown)


andrewrk:
Thanks for the follow-up! No need to apologize.

src/Module.zig
```diff
@@ -983,8 +983,10 @@ pub const Scope = struct {
         tree: ast.Tree,
         /// Whether this is populated or not depends on `zir_loaded`.
         zir: Zir,
-        /// Package that this file is a part of, managed externally.
-        pkg: *Package,
+        /// Package that this file is a part of, managed externally. This is
+        /// `null` for files not part of any package such as those processed
+        /// by `zig fmt` or `zig ast-check`.
+        pkg: ?*Package,
```
andrewrk, line 989: rather than making this optional, the places where it is being set to null/undefined need to create a valid package

andrewrk (changes requested):

andrewrk:
Thanks! Nice work.

Issue #7971, opened 2021-02-06
translate-c: stdint.h integer constant macros


mk12:
Given `repro.h`, taken from zig/build/lib/zig/libc/include/any-macos-any/stdint.h:

```c
#define UINT32_C(v)  (v ## U)
```

`zig translate-c repro.h | grep UINT32_C` gives:

```
pub const UINT32_C = @compileError("unable to translate C expr: expected ')' instead got: HashHash"); // repro.h:1:9
pub const __UINT32_C_SUFFIX__ = U;
```

This makes it difficult to use a C library that makes use of these macros. For example, I want to use `BGFX_RESET_VSYNC` from the bgfx library, but it triggers that `@compileError`, so I have to hardcode its value instead.

If general support for stringizing (`#`) and concatenation (`##`) is hard/impossible, Zig at least ought to special case common libc macros like these.

### Environment

macOS Catalina 10.15.7
Zig built from 8661a13b748d25d796c0246dad7ad2584d9b0824 using [Option A](https://github.com/ziglang/zig/wiki/Building-Zig-From-Source#option-a-use-your-system-installed-build-tools).
`zig version`: `0.8.0-dev.1118+8661a13b7`

Vexu:
Supporting stingize/tokenize is impossible #4006, however hardcoding some libc macros should be trivial.

Rocknest:
@Vexu you should have clarified that its impossible to represent `UINT32_C` macro as a zig function, not that its impossible to translate its usage in a library.

mk12:
@Rocknest Wouldn't it be possible to represent it like this?

```zig
fn UINT32_C(comptime x: comptime_int) u32 {
    return x;
}
```

Rocknest:
@mk12 No since the type is not integer literal. This kind of macros operate
on ast tokens.

LakeByTheWoods:
@Rocknest The correct usage of INTN_C macros is to with an unsuffixed integer constant. "The argument in any instance of these macros shall be an unsuffixed integer constant (as defined in6.4.4.1) with a value that does not exceed the limits for the corresponding type."

Rocknest:
@LakeByTheWoods its not a sustainable approach to hard code every macro defined in the c standard. And a generic solution requires a type that is able to hold AST (source code) as a value and allows arbitrary manipulation of it and subsequent injection of it into the program. 
```zig
pub fn UINT32_C(v: ast) ast {
    return v ++ ast(U);
}
```

[...]

PR #9265, opened 2021-06-29
translate-c: add framework for special-casing macros

ehaas:
Some macros (for example any macro that uses token pasting) cannot be
directly translated to Zig, but may nevertheless still admit a Zig
implementation. This provides a mechanism for matching macros against
templates and mapping them to functions implemented in c_translation.zig.

A macro matches a template if it contains the same sequence of tokens, except
that the name and parameters may be renamed. No attempt is made to
semantically analyze the macro. For example the following two macros are
considered equivalent:

```C
#define FOO(X) (X ## U)
#define BAR(Y) (Y ## U)
```

But the following two are not:

```C
#define FOO(X) X + 1
#define FOO(X) X + (1)
```

### DIFF
--- a/lib/std/zig/c_translation.zig
+++ b/lib/std/zig/c_translation.zig
@@ -350,3 +350,83 @@ test "Flexible Array Type" {
     try testing.expectEqual(FlexibleArrayType(*volatile Container, c_int), [*c]volatile c_int);
     try testing.expectEqual(FlexibleArrayType(*const volatile Container, c_int), [*c]const volatile c_int);
 }
+
+pub const Macros = struct {
+    pub fn U_SUFFIX(comptime n: comptime_int) @TypeOf(promoteIntLiteral(c_uint, n, .decimal)) {
+        return promoteIntLiteral(c_uint, n, .decimal);
+    }
+
+    fn L_SUFFIX_ReturnType(comptime number: anytype) type {
+        switch (@TypeOf(number)) {
+            comptime_int => return @TypeOf(promoteIntLiteral(c_long, number, .decimal)),
+            comptime_float => return c_longdouble,
+            else => @compileError("Invalid value for L suffix"),
+        }
+    }
+    pub fn L_SUFFIX(comptime number: anytype) L_SUFFIX_ReturnType(number) {
+        switch (@TypeOf(number)) {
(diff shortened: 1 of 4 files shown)


ehaas:
This mostly implements #4762 - call sites will still use the expanded macro but `wl_container_of` (or equivalent) will be available to Zig code as a useable function.

I believe it implements #7971 or at least should make the remaining macros straightforward to do

Vexu (approved):
This looks good but could you run some benchmarks before the merge?

`libepoxy` seems to have a lot of macros so it could be a good candidate.

ehaas:
> This looks good but could you run some benchmarks before the merge?
> 
> `libepoxy` seems to have a lot of macros so it could be a good candidate.

Do you want me to benchmark translation speed, or the speed of using this `CAST_OR_CALL` function vs directly inlining the code?

ehaas:
Ok I benchmarked with the following:

1. Install libepoxy
2. Create a `test.h` file with contents:
```c
#include <epoxy/gl.h>
```
3. The referenced file ends up including 1.5 megs and has 10,031 `#define` statements
4. Test system is a 2019 2.3 GHz 8-Core Intel Core i9 macbook pro w/ 64 gigs RAM running macOS 11.4, with a Release build of llvm/clang and a debug build of Zig

Benchmarked by running `zig translate-c -I/usr/local/include test.h > /dev/null` 10 times, deleting the Zig cache each time.

Master branch: `MEAN 0.5438 STDDEV 0.019`
This branch: `MEAN 0.5497 STDDEV 0.014`

ehaas:
I also created a worst-case benchmark - a file with 100,000 instances of the `wl_container_of` macro, which is 16.5 megs:

```C
#define FOO0(ptr, sample, member)                     \
(__typeof__(sample))((char *)(ptr) -                     \
     offsetof(__typeof__(*sample), member))
// ...
#define FOO99999(ptr, sample, member)                     \
(__typeof__(sample))((char *)(ptr) -                     \
     offsetof(__typeof__(*sample), member))
```

On the master branch the average runtime was 10.284 seconds, on this branch it was 10.344

Vexu:
Those seem very reasonable, thanks!

PR #9274, opened 2021-07-01
Skip over CRs at the end of multiline literals

### LINKED ISSUE #9257: parsing failure on windows
Hi,

the following code fails to compile only on Windows:
```
fn createTestTables() !void {
    const bar = &[_][]const u8{
        \\foobar
    };
    _ = bar;
}
```
I get this error:
```
>> C:\Users\vincent\dev\zig-windows-x86_64-0.9.0-dev.321+15a030ef3\zig.exe build-obj .\sqlite.zig
.\sqlite.zig:3:17: error: expected ',', found invalid bytes
        \\foobar
                ^
```

I'm testing with `zig-windows-x86_64-0.9.0-dev.321+15a030ef3.zip`.

Weird thing is it works fine on Linux and FreeBSD.

jmc-88:
Fixes #9257.
This is needed when tokenizing input containing DOS line endings, i.e.
the CRLF sequence.

### DIFF
--- a/lib/std/zig/tokenizer.zig
+++ b/lib/std/zig/tokenizer.zig
@@ -845,7 +845,7 @@ pub const Tokenizer = struct {
                         self.index += 1;
                         break;
                     },
-                    '\t' => {},
+                    '\t', '\r' => {},
                     else => self.checkLiteralCharacter(),
                 },
 
@@ -1934,6 +1934,10 @@ test "tokenizer - multi line string literal with only 1 backslash" {
 test "tokenizer - invalid builtin identifiers" {
     try testTokenize("@()", &.{ .invalid, .l_paren, .r_paren });
     try testTokenize("@0()", &.{ .invalid, .integer_literal, .l_paren, .r_paren });
+}
+
+test "tokenizer - multiline string literal with a DOS-encoded CRLF line endings (issue #9257)" {
+    try testTokenize("\\\\foobar\r\n", &.{.multiline_string_literal_line});
 }
 
 fn testTokenize(source: [:0]const u8, expected_tokens: []const Token.Tag) !void {


ifreund (commented):
This seems wrong, aren't windows line endings supposed to be CRLF? That, CR and then LF, or `"\r\n"`.

Vexu (changes requested):
Seems to me like `\r` should be handled with `\t` as is done elsewhere.

jmc-88:
Yes, you're quite right, though applying that change miraculously breaks a bunch of other code paths. I will look into it more as soon as I get the time.

jmc-88:
Mostly note to self: a complete `zig build -Dskip-release test-std` run shows that the only failure triggered by moving the `'\r'` to the same branch as `'\t'` is the ["writev/fsync/readv" test from /lib/std/os/linux/io_uring.zig](https://github.com/ziglang/zig/blob/8a863381d11adcdcb4e0509611af93c61fc89804/lib/std/os/linux/io_uring.zig#L1063-L1129) (specifically [line 1105](https://github.com/ziglang/zig/blob/8a863381d11adcdcb4e0509611af93c61fc89804/lib/std/os/linux/io_uring.zig#L1105)).

Vexu:
What does your error return trace look like? I've also been experiencing some failures with that test on both my Arch Linux desktop and laptop. Try running `zig test lib/std/os/linux/io_uring.zig --zig-lib-dir lib` a few times and see if that works. For me it fails on clean cache about half the time but the second run always succeeds.

jmc-88:
Oh, interesting, it didn't occur to me to try running it on a different system -- this laptop is also currently running Arch Linux... I tried it on a Debian host and it passed cleanly. ¯\\\_(ツ)_/¯

This is the failure log I got on Arch Linux: https://gist.github.com/jmc-88/54997867ae0e38cb6cd6f7331fce4142
Interestingly, if I don't pipe it somewhere, the error trace seems to get stuck in an infinite loop and I have to kill `zig` manually. That may or may not be a second issue.

Given the above, I pushed an updated version of this change, since the failure appears to be unrelated. PTAL.

Vexu:
Mine also used to loop until I tried it yesterday.

@jorangreef do you have any idea what could be happening here? It has happened to me on multiple 5.x kernels on Arch Linux. If you are interested in taking a look, I can provide an executable that can reproduce the failure but the debug info in it seems to always be corrupt.

PR #9279, opened 2021-07-02
Wasm,libc: fix wasm-ld failure in matching libc symbols.

mathetake:
I hope the comment in the change explains the issue well, but the tl;dr is that we should not mangle libc symbols with "c" module name because we intend to link libc with wasm-ld which does not take module names into account.

I found this problem during working on #9227, and without this change we need to drop "c" specifier on all the extern statements of libc symbols only for WASI target which is messy and introduces complexity in the codebase, and breaks the assumption of checking  libc linkage! https://github.com/ziglang/zig/blob/master/src/Sema.zig#L3133. This is because otherwise we end up having unresolved symbols like:

```
wasm-objdump /home/mathetake/zig/zig-cache/o/3f9eec2bc857dd8c7327f378ef491e47/test.wasm --section=import -x
...
 - func[6] sig=6 <exit|c> <- c.exit
 - func[7] sig=0 <malloc|c> <- c.malloc
 - func[8] sig=6 <free|c> <- c.free
 - func[9] sig=8 <write|c> <- c.write
```

### DIFF
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -428,8 +428,18 @@ static LLVMValueRef make_fn_llvm_value(CodeGen *g, ZigFn *fn) {
         // compiler as we're telling LLVM (using 'wasm-import-name' and
         // 'wasm-import-name') what the real function name is and where to find
         // it.
-        const bool use_mangled_name = target_is_wasm(g->zig_target) &&
+        bool use_mangled_name = target_is_wasm(g->zig_target) &&
                 fn_proto->is_extern && fn_proto->lib_name != nullptr;
+        // This is subtle but important to match libc symbols at static link time correctly.
+        // We treat "c" lib_name as a special library indicating that it should be defined
+        // in libc. But if we mangle a libc symbol name here with "c" module name, then wasm-ld cannot resolve
+        // the symbol. This is because at the static link time with wasm-ld, the linker does not
+        // take module names into account, and instead looking for a pure symbol name (i.e. function name)
+        // written into the ".linking" custom section (i.e. it does not use import section).
+        // This is the intended behavior of wasm-ld, because Wasm has a concept of host functions,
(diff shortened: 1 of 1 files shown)


src/stage1/codegen.cpp
```diff
@@ -428,8 +428,18 @@ static LLVMValueRef make_fn_llvm_value(CodeGen *g, ZigFn *fn) {
...
+        // This is subtle but important to match libc symbols at static link time correctly.
+        // We treat "c" lib_name as a special library indicating that it should be defined
+        // in libc. But if we mangle a libc symbol name here with "c" module name, then wasm-ld cannot resolve
+        // the symbol. This is because at the static link time with wasm-ld, the linker does not
+        // take module names into account, and instead looking for a pure symbol name (i.e. function name)
+        // written into the ".linking" custom section (i.e. it does not use import section).
+        // This is the intended behavior of wasm-ld, because Wasm has a concept of host functions,
+        // which are undefined functions supposed to be resolved by host runtimes *with module names*
+        // at load times even if it is "static linked" with the linker.
+        use_mangled_name = use_mangled_name && (strcmp(buf_ptr(fn_proto->lib_name), "c") != 0);
```
mathetake, line 442: not sure if this is the right problem to fix.. maybe in the semantic analysis phase and remove the lib_name entirely when `target_util.is_libc_lib_name` true.

mathetake:
@kubkon PTAL!

kubkon:
Sorry but I'm totally out of my depth with this as this deals with stage1 compiler. I'll summon @LemonBoy and @andrewrk to take a look though as they have a much better grasp at what is happening here.

mathetake:
np, thanks @kubkon for summoning right people!:)

andrewrk:
Sorry for the delay, looking at this within the next couple days.

mathetake:
Thanks @andrewrk for taking a look!:)

PR #9280, opened 2021-07-02
Add std.unicode.fmtUtf16le

marler8997:
A formatter for Utf16le strings.  Especially useful on windows when you want to print a wide string.  Wide bois need love too.

### DIFF
--- a/lib/std/unicode.zig
+++ b/lib/std/unicode.zig
@@ -318,9 +318,9 @@ pub const Utf16LeIterator = struct {
         assert(it.i <= it.bytes.len);
         if (it.i == it.bytes.len) return null;
         const c0: u21 = mem.readIntLittle(u16, it.bytes[it.i..][0..2]);
+        it.i += 2;
         if (c0 & ~@as(u21, 0x03ff) == 0xd800) {
             // surrogate pair
-            it.i += 2;
             if (it.i >= it.bytes.len) return error.DanglingSurrogateHalf;
             const c1: u21 = mem.readIntLittle(u16, it.bytes[it.i..][0..2]);
             if (c1 & ~@as(u21, 0x03ff) != 0xdc00) return error.ExpectedSecondSurrogateHalf;
@@ -329,7 +329,6 @@ pub const Utf16LeIterator = struct {
         } else if (c0 & ~@as(u21, 0x03ff) == 0xdc00) {
             return error.UnexpectedSecondSurrogateHalf;
         } else {
-            it.i += 2;
             return c0;
         }
     }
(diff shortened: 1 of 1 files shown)


lib/std/unicode.zig
```diff
@@ -770,6 +770,40 @@ fn calcUtf16LeLen(utf8: []const u8) usize {
...
+    utf16le: []const u16,
+    comptime fmt: []const u8,
+    options: std.fmt.FormatOptions,
+    writer: anytype,
+) !void {
+    _ = fmt;
+    _ = options;
+    var buf: [300]u8 = undefined; // just a random size I chose
+    var it = std.unicode.Utf16LeIterator.init(utf16le);
+    var u8len: usize = 0;
```
LemonBoy, line 784: Nope, the errors must be caught and handled. I'd suggest printing U+FFFD in case of error.

marler8997: What about when the `utfEncode` fails, same thing, print `U+FFFD`?

marler8997: I updated PR to print `U+FFFD` for all decoding/encoding errors.

lib/std/unicode.zig
```diff
@@ -770,6 +769,48 @@ fn calcUtf16LeLen(utf8: []const u8) usize {
...
+    writer: anytype,
+) !void {
+    const unknown_codepoint = 0xfffd;
+    _ = fmt;
+    _ = options;
+    var buf: [300]u8 = undefined; // just a random size I chose
+    var it = Utf16LeIterator.init(utf16le);
+    var u8len: usize = 0;
+    while (it.nextCodepoint() catch unknown_codepoint) |codepoint| {
+        u8len += utf8Encode(codepoint, buf[u8len..]) catch
```
marler8997, line 786: Note that this may not be *possible*.  If we get a codepoint from decoding it, it *might* be impossible for it to fail in `utf8Encode`.  Someone who knows more might be able to clarify.

lib/std/unicode.zig
```diff
@@ -770,6 +769,48 @@ fn calcUtf16LeLen(utf8: []const u8) usize {
...
+pub fn fmtUtf16le(utf16le: []const u16) std.fmt.Formatter(formatUtf16le) {
+    return .{ .data = utf16le };
+}
+
+test "fmtUtf16le" {
+    const expectFmt = std.testing.expectFmt;
+    try expectFmt("", "{}", .{fmtUtf16le(utf8ToUtf16LeStringLiteral(""))});
+    try expectFmt("foo", "{}", .{fmtUtf16le(utf8ToUtf16LeStringLiteral("foo"))});
+    try expectFmt("𐐷", "{}", .{fmtUtf16le(utf8ToUtf16LeStringLiteral("𐐷"))});
+    try expectFmt("퟿", "{}", .{fmtUtf16le(&[_]u16{std.mem.readIntNative(u16, "\xff\xd7")})});
```
Vexu, line 806: Does `'\u{ffd7}'` not work here?

marler8997: It would work here, but it wouldn't work for the lines below because they are invalid unicode byte sequences, and the point of this one and the last one are to be the previous/next byte around the invalid range, so I used the same format to make that clear.

Vexu: I don't think we validate these unicode codepoints anywhere as https://godbolt.org/z/r6nd8Ybfh compiles just fine.

marler8997: You need to reverse the byte order.  Try `"\u{d800}"` and you should get the error.

Vexu: Right, still not seeing errors https://godbolt.org/z/q56eKfaKo. Probably would be good to add errors here though, Rust at least seems to give an error on `"\u{d800}"`.

marler8997: you get an error when you pass it to `utf8ToUtf16LeStringLiteral` like you would do to use it in this test block

Vexu: But why would you do that when it's just one character?

I guess this is fine too since it's just some test but generally I'd prefer `"\xXX"` patterns to be the last resort.

marler8997: The order of the bytes must change depending on the platform endianness, and we can't use the utf8 decoder functions to decode them because they are invalid, so I'm specifying the bytes in little endian and using the native integer decoding.  Note that before I just had the bytes in the same order but it failed on a big endian platform in the CI so I added the rest to fix it.

If you have a simpler way to make this work please share :)

Vexu: Sounds like a bitcasted array of `u8` would also work, but I might just be confusing myself with all the endianness .

PR #9288, opened 2021-07-03
require nameless tuples for std.fmt

marler8997:
A user on discord showed this curious example that unexpectedly compiled but produced unexpected output:
```zig
const std = @import("std");

pub fn main() !void {
    std.debug.print("{s}\n", std.fmt.fmtSliceHexLower("foo"));
}
```

This example prints the string "foo" but it's supposed to print it in hex.  The problem is that it's missing the braces around the args.  It should have been this:

```zig
const std = @import("std");

pub fn main() !void {
    std.debug.print("{s}\n", .{std.fmt.fmtSliceHexLower("foo")});
}
```

This PR makes the first example a compile error.  It modifies `std.fmt` to only work with nameless tuples, i.e. `.{ x, y, z}`.  I do this by checking if the first field is named `0`.  This isn't perfect but it should catch 99% of cases for a relatively small cost to checking all fields.

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -93,6 +93,9 @@ pub fn format(
         @compileError("32 arguments max are supported per format call");
     }
 
+    if (comptime fields_info.len > 0 and !std.mem.eql(u8, fields_info[0].name, "0"))
+        @compileError("Expected tuple, found " ++ @typeName(ArgsType));
+
     comptime var arg_state: struct {
         next_arg: usize = 0,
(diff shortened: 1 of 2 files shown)


marler8997:
I see this breaks actual use cases such as:
```zig
    try expectFmt("hello world!", "{[greeting]s} world{[punctuation]c}", .{ .punctuation = '!', .greeting = "hello" });
```
Not sure what the correct path forward is so closing for now.

PR #9291, opened 2021-07-03
Update Xoroshiro128 parameters

leesongun:
This updates `Xoroshiro128` PRNG parameters to match [source](https://prng.di.unimi.it/xoroshiro128plus.c).

### DIFF
--- a/lib/std/rand/Xoroshiro128.zig
+++ b/lib/std/rand/Xoroshiro128.zig
@@ -33,8 +33,8 @@ fn next(self: *Xoroshiro128) u64 {
     const r = s0 +% s1;
 
     s1 ^= s0;
-    self.s[0] = math.rotl(u64, s0, @as(u8, 55)) ^ s1 ^ (s1 << 14);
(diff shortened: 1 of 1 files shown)


jedisct1:
These RNGs are expected to be deterministic, so we probably shouldn't change a given type to use a different algorithm or different parameters. If we want to support these parameters, it would be better to do it under a different name.

That being said, the currently recommended versions are Xoroshiro++ and Xoroshiro**: https://prng.di.unimi.it

So, maybe we should add Xoroshiro++, as a different type, and make it the default non-cryptographic RNG.

Looks like using Xoroshiro++ would also solve the bias mentioned in your other issue.

leesongun:
Sorry, this was closed by mistake. I didn't know that renaming branch would close PR. 

I understand that there could be, but I'm not aware of any non-cryptographic application that is dependent on specific behavior of PRNG. Can you give me some examples?

jedisct1:
Reproducible tests. Reproducible Monte Carlo simulations. Fuzzing.

PR #9298, opened 2021-07-04
implement xoshiro256++

leesongun:
This PR adds `xoshiro256++` PRNG to `std.rand`. The reason for the choice is the following.

1. Xoshiro256 series have larger state space than the xoroshiro128 series and also faster when both are vectorized. Also, when they aren't vectorized, both speeds are similar. 
2. `++` and `**` variants are free from linear complexity concerns of `+` variants, which makes them suitable as the default PRNG.
3. It seems [the author recommends `xoshiro256++` over `xoshiro256**`](https://github.com/JuliaLang/julia/issues/27614#issuecomment-547455998), and its vectorized performance is [more thoroughly tested compared to the `**` variant](https://prng.di.unimi.it/#remarks).

### DIFF
--- a/lib/std/rand/Xoshiro256.zig
+++ b/lib/std/rand/Xoshiro256.zig
@@ -48,9 +48,9 @@ fn next(self: *Xoshiro256) u64 {
 fn jump(self: *Xoshiro256) void {
     var s: u256 = 0;
 
-    comptime var table = 0x39abdc4529b1661ca9582618e03fc9aad5a61266f0c9392c180ec6d33cfd0aba;
+    var table: u256 = 0x39abdc4529b1661ca9582618e03fc9aad5a61266f0c9392c180ec6d33cfd0aba;
 
-    inline while (table != 0) : (table >>= 1) {
+    while (table != 0) : (table >>= 1) {
         if (@truncate(u1, table) != 0) {
             s ^= @bitCast(u256, self.s);
         }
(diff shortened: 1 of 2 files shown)


lib/std/rand/Xoshiro256.zig
```diff
@@ -0,0 +1,150 @@
...
+
+// Skip 2^128 places ahead in the sequence
+fn jump(self: *Xoshiro256) void {
+    var s: u256 = 0;
+
+    var table: u256 = 0x39abdc4529b1661ca9582618e03fc9aad5a61266f0c9392c180ec6d33cfd0aba;
+
+    while (table != 0) : (table >>= 1) {
+        if (@truncate(u1, table) != 0) {
+            s ^= @bitCast(u256, self.s);
```
jedisct1, line 55: This assumes that `u256` and `u64` have the same alignment.

Maybe it would be safer to define `s` as `[4]u64` and avoid the bitcasts.  This is something the compiler should be able to vectorize anyway.

leesongun: I'm not sure if I understood the problem correctly : Does `@bitCast` require both type to have same alignment?

jedisct1: It doesn't. Sorry for the confusion. The only requirements is that types have the same size.

Avoiding the bitcasts seems to produce slightly better code, though, avoiding unnecessary loads: https://godbolt.org/z/heMEao4Kv

jedisct1:
Alright, let's get this in.

Having a replacement for Xoroshiro128+ in 0.9 is definitely a good idea.

PR #9303, opened 2021-07-04
Fix #9299

### LINKED ISSUE #9299: unexpected truncation behavior with `comptime_int` larger than `u64` range
Trying to compile the [following code](https://godbolt.org/z/PY7dnPdar)
```Zig
comptime {
    var a : u8 = @truncate(u1, 1 << 64);
}
```
results in (the value `139978047751248` changes every time)
```
./main.zig:2:18: error: integer value 139978047751248 cannot be coerced to type 'u8'
    var a : u8 = @truncate(u1, 1 << 64);
                 ^
```

This seems to be reason behind the following problems : 

1. The [following code](https://godbolt.org/z/a5Y8fYs6z) reaches `unreachable`. 
```Zig
test "large comptime_int truncation" {
    if (@truncate(u1, 0x1_0000_0000_0000_0000) != 0)
        unreachable;
}
```
2. The following code prints `0`. 
```Zig
const print = @import("std").debug.print;
test "print large comptime_int truncation" {
    print("{}", .{@truncate(u1, 0x1_0000_0000_0000_0001)});
}
```

leesongun:
This PR fixes bigint truncation. When `dest->digit_count == 1`, the union `{digit; digits;}` represents `digit`, so it should simply copy value regardless of `leftover_bits`, instead of cloning `digits` vector.

### DIFF
--- a/src/stage1/bigint.cpp
+++ b/src/stage1/bigint.cpp
@@ -88,8 +88,13 @@ static void to_twos_complement(BigInt *dest, const BigInt *op, size_t bit_count)
     size_t digits_to_copy = bit_count / 64;
     size_t leftover_bits = bit_count % 64;
     dest->digit_count = digits_to_copy + ((leftover_bits == 0) ? 0 : 1);
-    if (dest->digit_count == 1 && leftover_bits == 0) {
-        dest->data.digit = op_digits[0];
+    if (dest->digit_count == 1) {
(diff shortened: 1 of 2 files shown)


Vexu (changes requested):
Needs a test, the ones from your issue will do fine just put them somewhere appropriate in the behavior tests.

PR #9305, opened 2021-07-04
Fix bigint_shl

leesongun:
This PR fixes `bigint_shl` behavior reaching unreachable in [the following code](https://godbolt.org/z/3hbhbxTeG). To check whether the `bigint_shl` result has only one digit, we should check whether shr gives the original operand.
```Zig
comptime {
    if (3<<63 == 1<<63) 
        unreachable; // reaches unrechable
}
```


leesongun:
I don't think I'll have time to find more `bigint` bugs, so I marked this ready for review. It might be a good idea to review (or rewrite) bigint implementation considering I found two bugs from reading 3 functions(`bigint_truncate`, `to_twos_complement`, `bigint_shl`).

PR #9307, opened 2021-07-05
std.json: Add support for recursive objects to std.json.parse

greenfork:
Inferred errors don't work for recursive function calls, so we add an explicit error set.

`parseInternal` recurses in several places, current error catches this:
https://github.com/ziglang/zig/blob/b7da1b2d45bc42a56eea3a143e4237a0712c4769/lib/std/json.zig#L1768

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1468,7 +1468,9 @@ pub const ParseOptions = struct {
     allow_trailing_data: bool = false,
 };
 
-fn skipValue(tokens: *TokenStream) !void {
+const SkipValueError = error{UnexpectedJsonDepth} || TokenStream.Error;
+
+fn skipValue(tokens: *TokenStream) SkipValueError!void {
     const original_depth = tokens.stackUsed();
 
     // Return an error if no value is found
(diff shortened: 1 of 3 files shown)


lib/std/json.zig
```diff
@@ -1530,7 +1532,26 @@ test "skipValue" {
...
+    InvalidEnumTag,
+    InvalidNumber,
+    MissingField,
+    NoUnionMembersMatched,
+    Overflow,
+    UnexpectedEndOfJson,
+    UnexpectedToken,
+    UnexpectedValue,
+    UnknownField,
+} || SkipValueError || TokenStream.Error || error{OutOfMemory} || std.fmt.ParseIntError;
```
daurnimator, line 1547: `ParseInternalError` intentionally doesn't have a specified error set, as different errors are only possible/reachable with certain types passed.
e.g. `OutOfMemory` cannot be hit when `T` isn't a pointer type.

greenfork: Okay, I tried to do your suggested approach. Now I don't know how to avoid infinite recursion. In the test I wrote it goes like this:
1. Check union errors, e.g. check all union fields
2. i64 - no errors, next
3. Array - check child field type
4. Child field type is union, check all union fields
5. ...

I can't check whether it is going to be a recursion because I need to retrieve a parent of my parent, the sequence is Union -> Array -> Union. And of course more complex scenarios are possible.

daurnimator: there lies the rub.

That's why the compiler couldn't figure it out either :P

The json parser accepting pointer types opens it up for infinite recursion; and currently this can lead to unbounded stack usage! Given the future goals of zig include limiting unbounded stack usage, perhaps the solution is to remove pointers to complex objects from the accepted input of `json.parse`.

greenfork: Sorry, I don't think I understand your suggestion. Do you suggest to raise a compile error whenever there's a recursive type definition we are trying to parse? This is the exact reason for this PR, I want to be able to parse into recursive definitions. Maybe I misunderstand something.

Also, can I do it like this? commit `Don't analyze already inferred types`

greenfork: By the way, maybe I wasn't clear enough. By "infinite recursion" I meant that compiler produces an error that eval quota is overreached (`@setEvalBranchQuota`), it already saves from infinite recursion scenario.

lib/std/json.zig
```diff
@@ -1530,7 +1532,81 @@ test "skipValue" {
     }
 }
 
-fn parseInternal(comptime T: type, token: Token, tokens: *TokenStream, options: ParseOptions) !T {
+fn ParseInternalError(comptime T: type, inferred_types: []const type) type {
```
daurnimator, line 1535: I think `inferred_types` should be `comptime`?

daurnimator (commented):
Well done! looks like you found a way to make it work; no API objections from me now :)

Though I'm not sure about some of the style, or if the pattern of `inferred_types` will make the stage1 compiler cry.

Will leave it for others to review.

Vexu (changes requested):
This could also be made public as `ParseError` so that users can get the error set without using `@typeInfo`.

lib/std/json.zig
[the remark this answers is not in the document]
daurnimator: See the above comment thread: something like this is required to not blow up on recursion.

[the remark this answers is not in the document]
greenfork: If I understood correctly, then yes, recursion is the reason. I can provide a function with a single argument though if this is the desired signature by moving this code to `ParseInternalErrorImpl` and doing something like this:
```zig
fn ParseInternalError(comptime T: type) type {
    const parse_internal_inferred_types = [_]type{};
    return ParseInternalErrorImpl(T, &parse_internal_inferred_types);
}
```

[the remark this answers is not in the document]
Vexu: Oh right, because the call hasn't completed yet, missed that.

That helper would work as the `ParseError` I mentioned in the other comment.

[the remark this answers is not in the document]
greenfork: I'm not sure I completely understood your comment but implementation should be fine.

PR #9310, opened 2021-07-06
Speed up stage 1 by improving hash functions

SpexGuy:
A significant portion of stage 1 compilation time was spent in const_values_equal, called as part of hash map lookups.  This was called much more often than it should have been because the hash functions used for const values were not very good hashes.  This PR improves those hash functions, reducing the time to create a debug build of stage 2 from 37 seconds to 30 seconds on my computer.

I also have audited struct field accesses throughout the codebase to ensure that comptime fields are handled correctly.

### DIFF
--- a/src/stage1/hash_map.hpp
+++ b/src/stage1/hash_map.hpp
@@ -12,7 +12,33 @@
 
 #include <stdint.h>
 
-template<typename K, typename V, uint32_t (*HashFunction)(K key), bool (*EqualFn)(K a, K b)>
+template<typename K>
+struct MakePointer {
+    typedef K const *Type;
+    static Type convert(K const &val) {
+        return &val;
+    }
+};
+
+template<typename K>
+struct MakePointer<K*> {
+    typedef K *Type;
+    static Type convert(K * const &val) {
+        return val;
+    }
+};
+
+template<typename K>
+struct MakePointer<K const *> {
+    typedef K const *Type;
+    static Type convert(K const * const &val) {
+        return val;
+    }
+};
+
+template<typename K, typename V,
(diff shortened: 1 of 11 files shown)


src/stage1/hash_map.hpp
```diff
@@ -51,7 +77,7 @@ class HashMap {
 
         if (_index_bytes == nullptr) {
             if (_entries.length < 16) {
-                _entries.append({HashFunction(key), 0, key, value});
+                _entries.append({HashFunction(MakePointer<K>::convert(key)), 0, key, value});
```
andrewrk, line 80: forgive my lack of C++ knowledge but why not just `&key` ?

SpexGuy: MakePointer selects an implementation based on the type.  When the key is `*T`, it's `key`.  When the key is `T`, it's `&key`.

andrewrk: mmm I'd feel a lot more comfortable with the callsites being updated to use `&` where appropriate. One C++ feature here, one C++ feature there, and then soon enough Bjarne Soustroup is murdering your firstborn child

SpexGuy: That doesn't work.  When the key is a pointer, we need to pass it by value.

andrewrk: ahh I see the problem. Alright, I'm still uncomfortable with this C++ creeping in, but I think it's justified enough for me to not push back on it any more.

andrewrk:
This branch takes me down from 20s to 17s for `./zig build -p zig-out -Dskip-install-lib-files -Dskip-non-native` with a release build of stage1.

andrewrk (approved):

PR #9315, opened 2021-07-06
Set paragraph line-height of language reference

### LINKED ISSUE #6313: Text in documentation is too tight
For some reason, the text for the documentation (https://ziglang.org/documentation/master/) has a very low line-height, so it is quite difficult to read. If you add this css to the page it will look much better:
```
#contents {
  line-height: 1.5em;
}
```

It will go from this:
image

to this one:
image

paulespinosa:
The line-height of paragraphs is set to 1.7. This value is chosen based on
recommendations from W3C [1] via MDN Web Docs [2] and the contents of the
language reference.

The MDN Web Docs offers two recommendations:
1. Use a unitless number for line-height.
2. Use a minimum value of 1.5 for main paragraph content.

The line-height of this commit uses 1.7 due to the mix of text and code font in
language reference's paragraphs.

[1] https://www.w3.org/TR/WCAG21/#visual-presentation
[2] https://developer.mozilla.org/en-US/docs/Web/CSS/line-height

Closes #6313

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -10,6 +10,9 @@
         font-family: system-ui, -apple-system, Roboto, "Segoe UI", sans-serif;
         margin: 0;
       }
+      p {
+        line-height: 1.7;
+      }
       a:not(:hover) {
         text-decoration: none;
       }


paulespinosa:
I have closed this unmerged pull request because I have another branch that incorporates this change.

PR #9324, opened 2021-07-07
translate-c: Handle underscore when used as an identifier

ehaas:


src/translate_c.zig
```diff
@@ -1866,7 +1866,13 @@ fn transDeclRefExpr(
     const value_decl = expr.getDecl();
     const name = try c.str(@ptrCast(*const clang.NamedDecl, value_decl).getName_bytes_begin());
     const mangled_name = scope.getAlias(name);
-    var ref_expr = try Tag.identifier.create(c.arena, mangled_name);
+    var ref_expr = if (mem.eql(u8, mangled_name, "_"))
+        try Tag.field.create(c.arena, .{
+            .lhs = Tag.this.init(),
+            .rhs = try Tag.string_literal.create(c.arena, "\"_\""),
```
ehaas, line 1872: Does this string literal (`"\"_\""`) need to be allocated in the arena? (e.g. `mem.dupe(c.arena, u8, "\"_\"")`)? Other ones are but I assume that is because they come from clang and have a different lifetime.

src/translate_c/ast.zig
```diff
@@ -799,6 +804,7 @@ const Context = struct {
     }
 
     fn addIdentifier(c: *Context, bytes: []const u8) Allocator.Error!TokenIndex {
+        if (std.mem.eql(u8, bytes, "_")) return addTokenFmt(c, .identifier, "@\"_\"", .{});
```
ehaas, line 807: `std/zig/fmt.zig` defines a `isValidId` function which would be the other place this could be implemented; but that seems like it could have far-ranging consequences. If it's safe to make the change there though I can remove this and put the check there.

src/translate_c.zig
Vexu: No it just needs to exist until the intermediate ast is rendered and static lifetime will suffice for that.

src/translate_c/ast.zig
Vexu: `isValidId` doesn't seem to be used outside `formatId` so I think it would be safe to change there.

src/translate_c/ast.zig
```diff
@@ -1253,6 +1259,13 @@ fn renderNode(c: *Context, node: Node) Allocator.Error!NodeIndex {
                 },
             });
         },
+        .this => {
+            return renderBuiltinCall(c, "@This", &.{});
+        },
+        .field => {
+            const payload = node.castTag(.field).?.data;
+            return renderBuiltinCall(c, "@field", &.{ payload.lhs, payload.rhs });
```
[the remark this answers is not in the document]
ehaas: I get an error using bare `@"_"` on the RHS:

```zig
pub var @"_": c_int = 5;
pub fn main() void {
    var x: c_int = @"_";
    _ = x;
}
```

```
./test.zig:3:20: error: `_` may only be used to assign things to
    var x: c_int = @"_";
                   ^
```

It's only relevant when trying to assign a global `_` variable to something - local `_`'s already get mangled, so it seems pretty edge-case anyway. 

Is there a way to get the current file's struct regardless of current scope? Alternatively could put something like `const @"This" = @This();` at the top of the file.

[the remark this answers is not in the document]
Vexu: > I get an error using bare @"_" on the RHS:

That is a stage1 limitation, it could probably be fixed easily by just removing the check from stage1 completely since `ast-check` implements the it properly.

> Is there a way to get the current file's struct regardless of current scope?

Nope, I had to put a `const file_struct = @This();` in `type.zig` for the same reason.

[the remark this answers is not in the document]
Vexu: Based on some quick testing this patch should fix the problem:
```diff
diff --git a/src/stage1/astgen.cpp b/src/stage1/astgen.cpp
index e3eebab0b..876dd384a 100644
--- a/src/stage1/astgen.cpp
+++ b/src/stage1/astgen.cpp
@@ -3821,9 +3821,6 @@ static IrInstSrc *astgen_identifier(Stage1AstGen *ag, Scope *scope, AstNode *nod
             const_instruction->value->special = ConstValSpecialStatic;
             const_instruction->value->data.x_ptr.special = ConstPtrSpecialDiscard;
             return &const_instruction->base;
-        } else {
-            add_node_error(ag->codegen, node, buf_sprintf("`_` may only be used to assign things to"));
-            return ag->codegen->invalid_inst_src;
         }
     }
 
```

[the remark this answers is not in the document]
ehaas: Ok, I removed `@This()` and `@field()` from the translate-c AST and also removed the stage1 check against assigning from `_`. With that removed it looks like I can just do `var foo = @"_";`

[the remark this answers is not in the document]
ehaas: Testing out this branch some more, I noticed that with the underscore change, I can initialize and assign from `@"_"` but cannot assign *to* it:

```zig
const std = @import("std");
var @"_": u32 = 5;
test "underscore" {
    try std.testing.expectEqual(@as(u32, 5), @"_");
    var x = @"_";
    try std.testing.expectEqual(@as(u32, 5), x);
    @"_" = 6;
    try std.testing.expectEqual(@as(u32, 6), @"_");
}
```

```
expected 6, found 5
test "underscore"... FAIL (TestExpectedEqual)
/Users/ehaas/source/zig/lib/std/testing.zig:82:17: 0x108266d6b in std.testing.expectEqual (test)
                ^
/Users/ehaas/test.zig:8:5: 0x108264b03 in test "underscore" (test)
    try std.testing.expectEqual(@as(u32, 6), @"_");
    ^
```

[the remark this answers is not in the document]
Vexu: I think it's fine to treat this as yet another stage1 miscompilation since it is non breaking. We'll soon enough have stage2 ready which will handle this correctly.

src/translate_c.zig
```diff
@@ -130,7 +130,7 @@ const Scope = struct {
         fn makeMangledName(scope: *Block, c: *Context, name: []const u8) ![]const u8 {
             const name_copy = try c.arena.dupe(u8, name);
             var proposed_name = name_copy;
-            while (scope.contains(proposed_name)) {
+            while (scope.contains(proposed_name) or mem.eql(u8, proposed_name, "_")) {
```
Vexu, line 133: Why is this check needed? `@"_"` should work like any other identifier and not need special casing.

ehaas: Probably a stage 1 issue, but if a local variable is named `_` I get a compile error in the translated zig:

```
./test.zig:1027:5: error: `_` is not a declarable symbol
    var @"_": c_int = 42;
    ^
```

ehaas: Simply removing that declarable symbol check from astgen.cpp is not sufficient; doing so results in this:

```zig
pub export fn bar(arg_x: c_int) void {
    var x = arg_x;
    var @"_": c_int = 42;
    if (@"_" != x) {
        abort();
    }
}
```

```
./test.zig:1028:9: error: use of undeclared identifier '_'
    if (@"_" != x) {
        ^
```

Vexu: I see, I think I'd still prefer to remove that check and leave that in as a stage1 miscompilation as it doesn't cause any `ast-check` errors.

ehaas: Ok, removed it and took out the relevant test code

PR #9335, opened 2021-07-08
Add waitid syscall on linux

malcolmstill:
Adds `waitid` syscall on linux

- wraps `syscall5(.waitid...`
- provides values for `idtype` parameter

This syscall is particularly useful in conjunction with [`pidfd_open`](https://github.com/ziglang/zig/blob/master/lib/std/os/linux.zig#L1416) (and the `P_PIDFD` `id_type`)  as it allows waiting on a process via its `pidfd` rather than its `pid`.

### DIFF
--- a/lib/std/os/bits/linux.zig
+++ b/lib/std/os/bits/linux.zig
@@ -205,10 +205,13 @@ pub const WCONTINUED = 8;
 pub const WNOWAIT = 0x1000000;
 
 // waitid id types
-pub const P_ALL = 0;
-pub const P_PID = 1;
-pub const P_PGID = 2;
-pub const P_PIDFD = 3;
+pub const P = enum(c_uint) {
+    ALL = 0,
+    PID = 1,
+    PGID = 2,
+    PIDFD = 3,
+    _,
+};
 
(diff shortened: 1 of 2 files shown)


lib/std/os/bits/linux.zig
```diff
@@ -204,6 +204,12 @@ pub const WEXITED = 4;
 pub const WCONTINUED = 8;
 pub const WNOWAIT = 0x1000000;
 
+// waitid id types
+pub const P_ALL = 0;
+pub const P_PID = 1;
+pub const P_PGID = 2;
+pub const P_PIDFD = 3;
```
daurnimator, line 211: This looks like it should be an enum, e.g.
```suggestion
pub const P = enum(c_uint) {
    ALL = 0,
    PID = 1,
    PGID = 2,
    PIDFD = 3,
    _,
};
```

daurnimator (approved):

lib/std/os/bits/linux.zig
malcolmstill: Thanks @daurnimator, I've made that change

PR #9336, opened 2021-07-08
io_uring: add sqe prep methods for epoll_ctl, poll_add, and poll_remove

lithdew:
Implement io_uring submission queue entry preparation methods for epoll_ctl, poll_add, and poll_remove.

Poll masks are designated as 32-bit little-endian integers as specified in liburing's definitions. Refer to the patch series here for more information: https://www.spinics.net/lists/io-uring/msg02853.html

Updated io_uring_prep_rw to take in an unsigned 64-bit address instead of an anytype. io_uring_sqe by default assumes that the address specified in a submission queue entry is an unsigned 64-bit integer.

cc @ifreund @jorangreef

### DIFF
--- a/lib/std/os/linux/io_uring.zig
+++ b/lib/std/os/linux/io_uring.zig
@@ -457,6 +457,22 @@ pub const IO_Uring = struct {
         return sqe;
     }
 
+    /// Queues (but does not submit) an SQE to perform a `epoll_ctl(2)`.
+    /// Returns a pointer to the SQE.
+    pub fn epoll_ctl(
+        self: *IO_Uring,
+        user_data: u64,
+        epfd: os.fd_t,
+        fd: os.fd_t,
+        op: u32,
+        ev: ?*linux.epoll_event,
+    ) !*io_uring_sqe {
+        const sqe = try self.get_sqe();
+        io_uring_prep_epoll_ctl(sqe, epfd, fd, op, ev);
+        sqe.user_data = user_data;
+        return sqe;
+    }
+
     /// Queues (but does not submit) an SQE to perform a `recv(2)`.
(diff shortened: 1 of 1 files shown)


andrewrk:
> Updated io_uring_prep_rw to take in an unsigned 64-bit address instead of an anytype. io_uring_sqe by default assumes that the address specified in a submission queue entry is an unsigned 64-bit integer.

Even on 32 bit kernels?

lithdew:
> > Updated io_uring_prep_rw to take in an unsigned 64-bit address instead of an anytype. io_uring_sqe by default assumes that the address specified in a submission queue entry is an unsigned 64-bit integer.
> 
> Even on 32 bit kernels?

From the looks of it, yep:

https://github.com/torvalds/linux/blob/1e16624d7b4376797ede36e3c955375cf0f23298/include/uapi/linux/io_uring.h#L27

lithdew:
Actually scratch the last comment, somewhat unsure now as I took a look at linux/types.h and it seems like the __u64 type translates to an unsigned long which is a minimum of 32 bits.

This would imply we need to fix the definition of io_uring_sqe in the standard library though.

```zig
pub const io_uring_sqe = extern struct {
    opcode: IORING_OP,
    flags: u8,
    ioprio: u16,
    fd: i32,
    off: u64,
    addr: u64,
    len: u32,
    rw_flags: u32,
    user_data: u64,
    buf_index: u16,
    personality: u16,
    splice_fd_in: i32,
    __pad2: [2]u64,
};
```

cc @ifreund @jorangreef - might you know if there were any reasons why addresses are typed in io_uring_sqe to be u64's rather than i.e. usize or c_ulong?

ifreund (approved):
LGTM, Thanks!

PR #9343, opened 2021-07-09
Add log testing

### LINKED ISSUE #5738: Integrate std.log with std.testing
Inspired by https://github.com/ziglang/zig/pull/5734#discussion_r446585745

The goal of the `std.log` API is to allow developers to leave log statements in finished code, without causing any performance issues or unwanted stderr/stdout for API consumers.

This harmonizes with testing! Here's an example:

```zig
const std = @import("std");
const testing = std.testing;
const log = std.log;

test "make sure an uncommon branch is hit" {
    testing.expect(myCoolCode(3, 4) == 7);
    testing.expect(myCoolCode(999, 100) == 100);
    testing.expectLog("a special case occurred: 999\n");
}

fn myCoolCode(a: i32, b: i32) i32 {
    if (a > 100) {
        log.info(.scope, "a special case occurred: {}\n", .{a});
        return b;
    }
    return a + b;
}
```

Much like the [test runner](https://github.com/ziglang/zig/blob/master/lib/std/special/test_runner.zig) clears the testing allocator and checks for leaks between tests, it can also specify a custom log handler, which buffers the log statements during tests, and provides the API to make sure log messages did or did not occur.

It could provide an API for exact matches, substring matches, or checking for the absence of a log message.

An open question would be how the "scope" parameter fits into this. It could be part of the API for checking, or maybe there would be a `testing` scope, which by default would never get printed and only would be activated for tests.

mjoerussell:
This is intended to close #5738

This PR introduces a way to test what kinds of logs are printed during the execution of some code. Specifically, it adds the following features:

1. `std.testing.LogTestSettings` - A struct that defines some optional rules for what logs are counted. It also holds the counts themselves.
2. `std.testing.setupLogTests` and `std.testing.expectLogCounts` - The two functions required for doing log tests. `setupLogTests` should be called at the beginning of a test to define the rules that are going to be used to count logs as they happen. `expectLogCounts` should be called at the end with the number of logs that were expected to be seen per rule set.
3. `std.log.testing` - A preset logging scope to use inside tests. Logs with the `.testing` scope:
    a. Don't count towards any log tests
    b. Are always printed if triggered during a test, regardless of the log level

The reason why I structured the log tests this way is to make it easier to test large numbers of logs. By defining the rules at the beginning we can just count each log (or not) as it happens. The other solution would be to allow the test writer to put arbitrary expects at the end of the test, but that would require storing all the logs in a list and then searching them for each expectation. While this method is a bit more limiting and requires more structure, I think it's a better way to measure large number of logs.

Here's an example of this being used in a test:

```zig
test "watch logs by scope" {
    std.testing.setupLogTests(.{ 
        .scope = @tagName(.a), // count logs with the scope name '.a'
        .exact_level = .debug, // count logs with log_level '.debug'
    });

    const scope_a = std.log.scoped(.a);
    const scope_b = std.log.scoped(.b);

    scope_a.debug("Count scope and level", .{});
    scope_b.debug("Count level", .{});
    scope_a.debug("Count scope and level", .{});
    scope_a.warn("Count scope", .{});

    try std.testing.expectLogCounts(.{
        .scope = 3,
        .exact_level = 3,
    });
}

```

### DIFF
--- a/lib/std/log.zig
+++ b/lib/std/log.zig
@@ -256,7 +256,7 @@ pub fn scoped(comptime scope: @Type(.EnumLiteral)) type {
 /// The default scoped logging namespace.
 pub const default = scoped(.default);
 
-/// This scope is ignored by `std.testing`'s log testing features.
+/// This scope is ignored by `std.testing`'s log testing features. If used inside of a test, it is always printed.
 pub const testing = scoped(.testing);
 
 /// Log an emergency message using the default scope. This log level is
--- a/lib/std/log.zig
+++ b/lib/std/log.zig
@@ -256,6 +256,9 @@ pub fn scoped(comptime scope: @Type(.EnumLiteral)) type {
 /// The default scoped logging namespace.
 pub const default = scoped(.default);
 
+/// This scope is ignored by `std.testing`'s log testing features.
+pub const testing = scoped(.testing);
+
 /// Log an emergency message using the default scope. This log level is
 /// intended to be used for conditions that cannot be handled and is usually
 /// followed by a panic.
--- a/lib/std/special/test_runner.zig
+++ b/lib/std/special/test_runner.zig
@@ -49,6 +49,7 @@ pub fn main() anyerror!void {
                 leaks += 1;
             }
         }
+        std.testing.log_test_settings = .{};
         std.testing.log_level = .warn;
 
         var test_node = root_node.start(test_fn.name, 0);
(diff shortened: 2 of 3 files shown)


andrewrk:
Thanks for working on this. Next step to getting it merged is to rebase against master and get those CI tests passing. Since it's been 4 months since any commits were pushed to this PR, please open a fresh pull request.

Issue #9002, opened 2021-06-05
Find native include and libraries is broken with msvc ABI


mlarouche:
I think https://github.com/ziglang/zig/pull/8878 broke the detection of MSVC Windows SDK find in Zig. I can't find `windows.h` or any Windows library when using the msvc ABI

Tried with Zig version 0.9.0-dev.5423+7babff0bc

```shell
❯ zig build -Dtarget=native-native-msvc --verbose
C:\Programmation\Zig\zig\out\install\x64-Release\bin\zig.exe build-exe C:\Programmation\Hochelag\src\main.zig -lc -ld3d12 -ldxgi -ld3dcompiler -ldxguid --cache-dir C:\Programmation\Hochelag\zig-cache --global-cache-dir C:\Users\micha\AppData\Local\zig --name Hochelag -target native-native-msvc -mcpu=skylake-sgx --enable-cache
lld-link: error: could not open 'msvcrtd.lib': no such file or directory
lld-link: error: could not open 'vcruntimed.lib': no such file or directory
lld-link: error: could not open 'ucrtd.lib': no such file or directory
lld-link: error: could not open 'legacy_stdio_definitions.lib': no such file or directory
lld-link: error: could not open 'kernel32.lib': no such file or directory
lld-link: error: could not open 'ntdll.lib': no such file or directory
lld-link: error: could not open 'd3dcompiler.lib': no such file or directory
lld-link: error: could not open 'dxguid.lib': no such file or directory
error: LLDReportedFailure
Hochelag...The following command exited with error code 1:
C:\Programmation\Zig\zig\out\install\x64-Release\bin\zig.exe build-exe C:\Programmation\Hochelag\src\main.zig -lc -ld3d12 -ldxgi -ld3dcompiler -ldxguid --cache-dir C:\Programmation\Hochelag\zig-cache --global-cache-dir C:\Users\micha\AppData\Local\zig --name Hochelag -target native-native-msvc -mcpu=skylake-sgx --enable-cache
error: the following build command failed with exit code 1:
C:\Programmation\Hochelag\zig-cache\o\1163269d18ecaacb76d29494e60535eb\build.exe C:\Programmation\Zig\zig\out\install\x64-Release\bin\zig.exe C:\Programmation\Hochelag C:\Programmation\Hochelag\zig-cache C:\Users\micha\AppData\Local\zig -Dtarget=native-native-msvc --verbose
```

Issue #9124, opened 2021-06-15
Windows native-native-gnu can't find odbc32.lib


mjoerussell:
Not sure if this is an issue with Zig or something else that Zig is depending on, but I've noticed some issues in 0.8/master regarding using and linking to `odbc32.lib` on Windows. I've also tested on macos with master and found no issues.

I've confirmed that none of the below issues are present on Zig `0.8.0-dev.2305+d228d8605`.

1. Importing C macro `__MSABI_LONG(x)` no longer works. I mentioned in the original issue #9120 and was shown a fix that seems to work, but I'm still curious why it broke in the new release.
2. The following C `#define`'s are no longer present:
    * SQL_OV_ODBC3_80
    * SQL_CP_DRIVER_AWARE
    * SQL_PARAM_DATA_AVAILABLE
3. The most pressing issue - I'm unable to link against `odbc32.lib`. 

If I try to build a project which links against `odbc32.lib` in Zig version `0.9.0-dev.140+4173a2bc2`, I get the following error:

```
error(link): DLL import library for -lodbc32 not found
error: DllImportLibraryNotFound
zdb...The following command exited with error code 1:
C:\Users\mjoer\Code\zig\build-release\bin\zig.exe build-exe C:\Users\mjoer\Code\zdb\src\main.zig -lc -lodbc32 --cache-dir C:\Users\mjoer\Code\zdb\zig-cache --global-cache-dir C:\Users\mjoer\AppData\Local\zig --name zdb --pkg-begin odbc C:\Users\mjoer\Code\zdb\zig-odbc\src\lib.zig --pkg-end --enable-cache
error: the following build command failed with exit code 1:
C:\Users\mjoer\Code\zdb\zig-cache\o\7d4fee359512fdc30a31bf8e390544fb\build.exe C:\Users\mjoer\Code\zig\build-release\bin\zig.exe C:\Users\mjoer\Code\zdb C:\Users\mjoer\Code\zdb\zig-cache C:\Users\mjoer\AppData\Local\zig 
run
```

The same code running on Zig `0.8.0-dev.2305+d228d8605` links with no issues, on the exact same Windows configuration.

xackus:
This is probably why: #6565. Try setting the target to `native-native-msvc`.

mjoerussell:
I think I'm still missing something, once I specify that then the `@cImport` starts failing.

When I import this: 

```zig
@cImport({
    @cDefine("__MSABI_LONG(x)", "(long)(x)");
[...]

PR #9344, opened 2021-07-09
Fix argument forwarding to LLVM on Windows

### LINKED ISSUE #9225: Windows: `zig.exe: error: not ranlib, ar, lib or dlltool`
Hello! 

I have downloaded the zig compiler from the website hoping I could use it to replace my C build tools on Windows. When I run `.\zig.exe ar`, however, I receive the following error: `zig.exe: error: not ranlib, ar, lib or dlltool`. 

This happens for both versions 0.9.0-dev and 0.8.0.

SpexGuy:
Fixes #9225
Might also fix other windows issues like #9002

### DIFF
--- a/src/zig_clang_driver.cpp
+++ b/src/zig_clang_driver.cpp
@@ -338,7 +338,15 @@ static int ExecuteCC1Tool(SmallVectorImpl<const char *> &ArgV) {
 extern "C" int ZigClang_main(int argc_, const char **argv_);
 int ZigClang_main(int argc_, const char **argv_) {
   noteBottomOfStack();
-  llvm::InitLLVM X(argc_, argv_);
+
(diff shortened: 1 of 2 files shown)


mjoerussell:
For what it's worth, this has not fixed #9002 or #9124 for my project.

SpexGuy:
Yeah I'm tracking 9002 down now, it's definitely a different bug.

andrewrk (commented):
As long as this doesn't break non-Windows targets, looks good. How is it getting the argc, argv with this change?

SpexGuy:
Zig is getting argc/argv the same way it always has, building a modified version for llvm, and then passing that in.  LLVM just doesn't ignore those values on windows anymore.

PR #9347, opened 2021-07-10
Coff linker: Add IMPLIB support

kkartaltepe:
Allow --out-implib as passed by cmake and meson to be correctly passed
through to the linker to generate import libraries.

This allows most of glib to be compiled for --target=x86_64-windows-gnu by just setting the appropriate compiler. Though there are still some compiler/linker issues resulting in missing symbols compared to compile with the mingw toolchain I think they are unrelated.

One note is that cmake/meson pass --subsystem as a linker flag when they detect clang as the compiler but zig appears to support this as a compiler level flag. I'd be happy to add that in another PR as well if it sounds useful.

I am not sure if this the correct way to support the cache, any input is appreciated.


kkartaltepe:
Ah it seems this crossed paths with #9251 which looks like it adds support for the ms `link` style argument while my change is adding the `ld` style, maybe they can be combined?

alexrp:
I think you need to add a check for `out_implib` here: https://github.com/ziglang/zig/blob/3ae6dbb13af3f31947cb8dae321e62f2ee267373/src/Compilation.zig#L882-L896

src/main.zig
```diff
@@ -1476,6 +1477,12 @@ fn buildOutputType(
                     ) catch |err| {
                         fatal("unable to parse '{s}': {s}", .{ arg, @errorName(err) });
                     };
+                } else if (mem.eql(u8, arg, "--out-implib")) {
```
alexrp, line 1480: `mem.eql(u8, arg, "-implib") or mem.eql(u8, arg, "--out-implib")`

alexrp:
It doesn't *technically* matter since you're adding a new field to pass the option along, but could you include [this change](https://github.com/ziglang/zig/pull/9251/files#diff-8e55d33a14159f53ad7cd7678dd28ae6e2dd2525df591243361d44b79d6ba363) as a separate commit? I was surprised to find that the COFF linker code didn't pass `lld_argv` along like ELF and Mach-O. It seems worth fixing that so somebody doesn't run into that surprise in the future.

src/link/Coff.zig
```diff
@@ -1034,6 +1035,10 @@ fn linkWithLLD(self: *Coff, comp: *Compilation) !void {
             try argv.append(p);
         }
 
+        for (self.base.options.lib_dirs) |lib_dir| {
```
alexrp, line 1038: Is moving this here actually important?

alexrp:
#9251 can be closed in favor of this if my comments above are addressed.

src/link/Coff.zig
kkartaltepe: Git didnt try hard enough to get a small hunk in this case, maybe I can move it somewhere else so git makes a better hunk.

kkartaltepe: Oh hah, i did accidentally move it. Thanks.

kkartaltepe:
So there is still an issue with the cache only tracking the output dll and not the implib with this change, when rebuilding a project linking will fail as the implib files are not restored when zig attempts to use the cache.

Thanks for the comments, i've added the changes suggested.

alexrp:
Can you elaborate on the caching issue? I don't quite follow.

kkartaltepe:
Oh sorry it may be due to me swapping between unmodified zig and this patch that causes the caching issue actually.

kkartaltepe:
Ah actually no i have replicated, in my case im build glib. And cmake doesnt track the implib correctly so its not cleaned up on `ninja clean`. But the steps to replicate for me.

1. delete ~/.cache/zig
2. build project
3. delete build dir (or clean and then delete some specific implib)
4. rebuild project
5. observe linking to fail with missing implib.

alexrp:
Ok, I see. That actually makes sense.

I think we need to handle this similarly to how `.pdb` outputs are handled. Basically, save the requested path in `out_implib`, but output the file to the cache directory. Then, in an update hook, copy the `.lib` to `out_implib`.

See: https://github.com/ziglang/zig/blob/7b5d139fd30a7225f073125b8a53e51a2454d223/src/main.zig#L2396-L2419

kkartaltepe:
Ill have to take more look later, but i cannot see where PDBs (my zig is pretty poor) are ever output/copied to cache dir. A trivial implementation fails, presumably because the file is correctly output the build dir and nothing is added to the cache.

```diff
--- a/src/main.zig
+++ b/src/main.zig
@@ -2427,6 +2427,10 @@ fn updateModule(gpa: *Allocator, comp: *Compilation, hook: AfterUpdateHook) !voi
 
                 _ = try cache_dir.updateFile(src_pdb_path, cwd, dst_pdb_path, .{});
             }
+
+            if (comp.bin_file.options.out_implib) |out_implib| {
+                _ = try cache_dir.updateFile(out_implib, cwd, out_implib, .{});
+            }
         },
     }
 }
```

alexrp:
> but i cannot see where PDBs (my zig is pretty poor) are ever output/copied to cache dir.

I don't think Zig does this explicitly. Clang internally outputs the `.pdb` next to the output binary, which happens to be in the cache directory.

```console
$ zig cc main.c -target x86_64-windows-gnu -g
$ zig cc main.c -target x86_64-windows-gnu -g
$ tree zig-cache/o
zig-cache/o
├── 6523ca4147cba05665bc073987ed7468
│   └── main.obj
└── 8098b8dc53415f7765e265aa3b4ffb3e
    ├── a.exe
    ├── a.pdb
    └── lld.id -> 24674bd217dde2e17d4a2c84a7261bd8
```

> A trivial implementation fails

Here's what I think needs to be done exactly:

* Instead of passing `out_implib` through the compilation and linking code, just pass a Boolean like `emit_implib`.
* In the COFF linker code, if `emit_implib` is `true`, pass `-IMPLIB:<path to output file, but extension changed to .lib>`.
* In the update hook, if `out_implib` is non-`null`, copy the `.lib` from the cache directory to the output directory in exactly the same way as done for the `.pdb`.

kkartaltepe:
> Here's what I think needs to be done exactly:
> 
>     * Instead of passing `out_implib` through the compilation and linking code, just pass a Boolean like `emit_implib`.
> 
>     * In the COFF linker code, if `emit_implib` is `true`, pass `-IMPLIB:<path to output binary, but extension changed to .lib>`.
> 
>     * In the update hook, if `out_implib` is non-`null`, copy the `.lib` from the cache directory to the output directory in exactly the same way as done for the `.pdb`.

Thanks, I think the issue I have is that to support being called via `zig cc` and being consistent with other compilers when being called with `-Wl,--out-implib,some/path/some.ext` zig should ensure that `some/path/some.ext` is really created, so `bool` isnt an option to be compatible.

For now I have have zig output into the cache with a fixed name `<bin>.lib` and then copy back into the specified location.

I'm not sure how correct this is on two parts. Zig's source suggests you may be able to compile without the cache in which case creating the additional `.lib` file in the cwd might be wrong? Secondly if you were using a different directory that didnt exist I assume this copy would fail, im not sure if there is a `mkdir -p` equivalent in zig or it should be used when we try to do every copy.

alexrp:
> Zig's source suggests you may be able to compile without the cache in which case creating the additional `.lib` file in the cwd might be wrong?

What code are you referring to? You might be right, but I haven't come across a way to actually do that. It seems to me that a lot of functionality in Zig's driver wouldn't work well without the cache (e.g. compiling musl on demand).

> Secondly if you were targeting a different directory than the binary that didnt exist I assume this copy would fail, im not sure if there is a `mkdir -p` equivalent in zig or if it should be used when we try to do every copy.

This is a good point, we should definitely do that. I'm not sure whether such a function exists in the stdlib though.

kkartaltepe:
> What code are you referring to? You might be right, but I haven't come across a way to actually do that. It seems to me that a lot of functionality in Zig's driver wouldn't work well without the cache (e.g. compiling musl on demand).

This line https://github.com/ziglang/zig/blob/master/src/Compilation.zig#L589 which is the source for the cache dir in the linker.

alexrp:
> im not sure if there is a `mkdir -p` equivalent in zig or if it should be used when we try to do every copy.

I asked on the Zig Discord and they suggested that `fs.makePath(fs.cwd(), implib_path)` should work here.

alexrp:
> > What code are you referring to? You might be right, but I haven't come across a way to actually do that. It seems to me that a lot of functionality in Zig's driver wouldn't work well without the cache (e.g. compiling musl on demand).
> 
> This line https://github.com/ziglang/zig/blob/master/src/Compilation.zig#L589 which is the source for the cache dir in the linker.

Someone might be able to offer more insight here, but as far as I can tell, we don't have to worry about the non-caching case: https://github.com/ziglang/zig/blob/41d12395906f0bb71732d705a8a90bf0923bdf0b/src/main.zig#L1500-L1504

* Note the `enable_cache` assignment.
* There's no way to explicitly disable the cache from the command line.
* We only care about the C case; `-Wl,-implib` (etc) can't be passed when building Zig code.

src/link/Coff.zig
```diff
@@ -1034,6 +1038,10 @@ fn linkWithLLD(self: *Coff, comp: *Compilation) !void {
             try argv.append(p);
         }
 
+        if (self.base.options.out_implib) |out_implib| {
```
alexrp, line 1041: There's an `unused capture` error on this line; change to `|_|`.

andrewrk:
Next step here is to rebase against master and get the CI tests passing.

src/link/Coff.zig
kkartaltepe: Fixed, thanks.

kkartaltepe:
I have pushed the PR with the linting error about the closure fixed and with the steps to mkdir when there is a directory as part of the implib path.

src/main.zig
```diff
@@ -2418,6 +2428,15 @@ fn updateModule(gpa: *Allocator, comp: *Compilation, hook: AfterUpdateHook) !voi
 
                 _ = try cache_dir.updateFile(src_pdb_path, cwd, dst_pdb_path, .{});
             }
+
+            if (comp.bin_file.options.out_implib) |out_implib| {
```
kubkon, line 2432: I believe this doesn't cover the case when targeting msvc ABI for which `link-lld` by default does generate an importlib and therefore setting `out_implib` is not required. In this case, might I suggest incorporating an `else`-prong or equivalent similar to the following diff:

```diff
diff --git a/src/main.zig b/src/main.zig
index cbe57eb84..cfa1374e5 100644
--- a/src/main.zig
+++ b/src/main.zig
@@ -2448,6 +2448,35 @@ fn updateModule(gpa: *Allocator, comp: *Compilation, hook: AfterUpdateHook) !voi

                 _ = try cache_dir.updateFile(src_pdb_path, cwd, dst_pdb_path, .{});
             }
+
+            // If a .lib file is part of the expected output, we must also copy
+            // it into place here.
+            // Note that when targeting GNU (mingw), importlib will only be generated
+            // by the linker if explicitly asked for it, unlike on msvc which is always
+            // generated side-by-side with the dll. Therefore, we try to update .lib
+            // from source to destination output paths, but don't trip up when the file
+            // is not found.
+            const is_dyn_lib = comp.bin_file.options.output_mode == .Lib and comp.bin_file.options.link_mode == .Dynamic;
+            const have_importlib = is_coff and is_dyn_lib;
+            if (have_importlib) {
+                const src_bin_ext = fs.path.extension(bin_sub_path);
+                const dst_bin_ext = fs.path.extension(full_path);
+
+                const src_lib_path = try std.fmt.allocPrint(gpa, "{s}.lib", .{
+                    bin_sub_path[0 .. bin_sub_path.len - src_bin_ext.len],
+                });
+                defer gpa.free(src_lib_path);
+
+                const dst_lib_path = try std.fmt.allocPrint(gpa, "{s}.lib", .{
+                    full_path[0 .. full_path.len - dst_bin_ext.len],
+                });
+                defer gpa.free(dst_lib_path);
+
+                _ = cache_dir.updateFile(src_lib_path, cwd, dst_lib_path, .{}) catch |err| switch (err) {
+                    error.FileNotFound => {}, // just fall-through
+                    else => |e| return e,
+                };
+            }
         },
     }
 }
```

Feel free to reuse the diff and no attribution is needed.

kubkon (changes requested):
Thanks for the PR! Here's my suggestion for handling this and hopefully making `zig cc -shared` behave the same way by default as `zig build-lib -dynamic`.

From my brief investigation, I believe that by default `link-lld` generates a matching importlib side-by-side with the output PE binary if we target msvc ABI. However, for mingw (or gnu ABI), we are required to explicitly request generating a matching importlib by providing an `-IMPLIB` flag as you've done in this PR. Now, since `zig build-lib -dynamic` always generates a matching importlib regardless of the targeted ABI, my suggestion would be to replicate the same behaviour with `zig cc -shared` and append `-IMPLIB` always with the default output path if the user didn't override the path, or the path if they did. Does that make sense?

src/main.zig
```diff
@@ -1477,6 +1478,14 @@ fn buildOutputType(
                     ) catch |err| {
                         fatal("unable to parse '{s}': {s}", .{ arg, @errorName(err) });
                     };
+                } else if (mem.eql(u8, arg, "--out-implib") or
```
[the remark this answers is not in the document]
kkartaltepe: This change does not add the option for zig build generally, it only supports passing along the associated linker flag so it was not added to the usage doc.

src/Compilation.zig
```diff
@@ -727,6 +727,7 @@ pub const InitOptions = struct {
     test_filter: ?[]const u8 = null,
     test_name_prefix: ?[]const u8 = null,
     subsystem: ?std.Target.SubSystem = null,
+    out_implib: ?[]const u8 = null,
```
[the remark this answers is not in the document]
kkartaltepe: >  what the intention behind this optional actually is?

Im not sure I understand what you mean, but its optional because the flag is optional for all linkers and in practice it is not always specified by build tools. Does that make sense?

I can add a comment that its only for windows/PE but its the same as subsystem in this respect which does not contain this comment.

src/main.zig
kkartaltepe: Can we tell if we are targeting msvc linkers at the time that we are parsing flags? If so it seems it would be cleaner to just always set the out_implib option on those targets and make that functionality explicit rather than rely on the implicit behavior of those linkers?

src/main.zig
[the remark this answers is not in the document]
kubkon: Right, and that's why I think a one-liner explaining that this is available when targeting Windows would be nice. But it's a minor nit so feel free to ignore.

src/Compilation.zig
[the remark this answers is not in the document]
kubkon: > > what the intention behind this optional actually is?
> 
> Im not sure I understand what you mean, but its optional because the flag is optional for all linkers and in practice it is not always specified by build tools. Does that make sense?

I didn't mean the optional but rather what is the string meant to be: a name or a path to the imporlib, etc.

> I can add a comment that its only for windows/PE but its the same as subsystem in this respect which does not contain this comment.

Sure, but that doesn't mean we are not allowed to improve the readability of the new additions to the codebase.

src/main.zig
kubkon: That's exactly what I'd like to see. So to sum up, here are the behavior characteristics:
1. the user may explicitly specify path/name of the importlib via `--out-implib` flag which will override any default set in `out_implib` optional
2. if `--out-implib` was not set explicitly, regardless of the targeted ABI, in `Compilation.zig` (most likely), we set the default value of `out_implib` to the same path as the final binary; then, in `Compilation.updateModule` we copy the `*.lib` from cache into the final destination dir.

Does that make sense?

You can check for ABI with

```zig
some_target: std.Target = ...;
some_target.abi == .msvc;
```

In particular, in `Compilation.updateModule(comp)` you could do `comp.bin_file.options.target.abi == .msvc` or similar.

kkartaltepe: I have implemented what I tried to describe, let me know if this works for you or you would prefer to do this after the linker has been invoked like you described.

src/main.zig
```diff
@@ -2005,6 +2014,16 @@ fn buildOutputType(
         else => false,
     };
 
+    // Always output import libraries (.lib) when building for msvc to replicate
+    // `link` behavior. lld does not always output import libraries so on the
```
kubkon, line 2018: Actually, we always output import lib regardless of the ABI, so for mingw too. Hence why I suggested always setting the default for `out_implib`. This way we would unify interfaces between `build-lib` and `cc`, which I believe we should aim to do.

src/main.zig
```diff
@@ -2005,6 +2014,16 @@ fn buildOutputType(
...
+    // gnu abi users must set out_implib.
+    if (output_mode == .Lib and emit_bin == .yes and target_info.target.abi == .msvc and out_implib == null) {
```
kubkon, line 2020: If `emit_bin` is set when building static archives, this check will let that through, so you should also check for dynamic linking `comp.bin_file.options.link_mode == .Dynamic`.

src/link/Coff.zig
```diff
@@ -1095,6 +1099,10 @@ fn linkWithLLD(self: *Coff, comp: *Compilation) !void {
             try argv.append(p);
         }
 
+        if (self.base.options.out_implib != null) {
+            try argv.append(try allocPrint(arena, "-IMPLIB:{s}.lib", .{full_out_path}));
```
kubkon, line 1103: Should this not set `-IMPLIB` to the value in `out_implib` rather than `full_out_path`? What if the user sets `--out-implib` to something different than location of the final binary?

kubkon (changes requested):

src/link/Coff.zig
kkartaltepe: `full_out_path` here is the path within the cache and not the final destination. The final destination is handled is the path stored in `out_implib` that is used during the `UpdateModule` step, similar to pdbs. We just choose a path here that is unlikely to collide with other linker outputs.

src/main.zig
kkartaltepe: My understand is the import library is only needed when building dynamic libraries. When building static libraries the import parts are built into the archive in all platforms, and this is why they share the extension in windows.

Since i dont have windows handy to test this is simply my assumption and some googling suggests its probably not wrong.

kubkon: Your understanding is correct and the way we signal a dynamic library in the linker is with a combination `comp.bin_file.options.output_mode.Lib` and `comp.bin_file.options.link_mode.Dynamic`, and I'm pointing at the fact that we should check for this combo and not only if it's a `.Lib`.

src/link/Coff.zig
kubkon: Gotcha, thanks.

kubkon: I guess this is fine to leave as-is for now but this will probably need to be cleaned up and put as logic in the linker proper rather than `Compilation.updateModule`. After all, we also need to update a dSYM bundle in MachO which should not be of any concern to the `Compilation` module. cc @andrewrk

src/main.zig
kkartaltepe: Sorry and that .Lib is only set when building shared libs, from line 1510

kubkon: That's not correct. We always discriminate shared objects from static archives using the combo. See [here](https://github.com/ziglang/zig/blob/eb010ce65ddddb23ceddae86d377432cf1b2b01c/src/link/MachO.zig#L628) for instance.

kkartaltepe: I see, I think the issue is I am adding this outside of the .cc/.cpp args check. Meaning it will affect other build modes. Maybe this is what you intend with your comment on aligning build-lib? I can add the dynamic check as well but should we restrict this to cc/cpp modes as well until the option can be added as a regular zig flag instead of strictly a linker flag?

kubkon: I think we should be fine with other build modes if we set the default for the flag in further down the stack in `Compilation` rather than here. Anyhow, my suggestion would be to try it out and if things break along the way, then we can figure out what to fix and where. If you don't feel like wanting to pursue this, I'm happy to take over from here. Otherwise, I'm happy to work with you to get this merged.

kkartaltepe: Sure, by all means if you have time to work on this id be happy if you could work on it as Im mostly just muddling my way through zig to get something that works. Of course feel free to use anything in this pr as you see fit.

Kha:
Fwiw, I can confirm that this PR works for building and linking against a non-trivial import library, getting me one step closer to [shipping Zig with Lean](https://github.com/leanprover/lean4/pull/659).

andrewrk:
I rebased this and added a few commits on top of it before merging it.

```
$ ls blah*
blah.zig

$ cat blah.zig 
export fn foo() void {}

$ ./zig build-lib -dynamic blah.zig -target x86_64-windows

$ ls blah*
blah.dll  blah.lib  blah.pdb  blah.zig

$ file blah.lib 
blah.lib: current ar archive

```

@Kha - mind giving it a shot for your use case and let me know if you run into any other problems?

alexrp:
@andrewrk note that the really important test here is whether you can get an implib when building with `zig cc -shared`. AFAIK, `zig build-lib -dynamic` already produced implibs correctly before this PR.

andrewrk:
```
andy@ark ~/tmp> cat foo.c
void foo(void) {}
andy@ark ~/tmp> ls foo*
foo.c
andy@ark ~/tmp> zig cc -o foo.dll -shared foo.c -target x86_64-windows
andy@ark ~/tmp> ls foo*
foo.c  foo.dll  foo.lib
```

PR #9352, opened 2021-07-10
stage2 astgen: error for return outside of function scope

### LINKED ISSUE #9346: Internal compiler error when analyzing the scope of a pointer
While learning Zig on Godbolt I came across this compiler error:
```
Assertion failed at /home/andy/dev/bootstrap-zig/zig/src/stage1/ir.cpp:14690 in ir_analyze_container_member_access_inner. This is a bug in the Zig compiler.thread 1 panic: 
Compiler returned: 139
```

I had a look at the assertion in question and it appears to be related to the scope analysis of pointer failing because a function is returning `nullptr` where it shouldn't.

Below is the minimum reproduction code to produce the ICE.

```zig
pub const Allocator = struct {
    pub fn create(self: *Allocator, size: usize) ?[*]u8 {
        return null;
    }
};

pub const String = struct {
    _allocator: Allocator,

    _length: u32,

    _size: u32,

    _buffer: ?[*]u8,

    pub const empty = return .{
        ._allocator = Allocator{},
        ._length = 0,
        ._size = 0,
        ._buffer = null,
    };

    pub fn load(allocator: *Allocator, data: []const u8) String {
        if (allocator.create(data.len)) |allocation| {
            var unicodeLength: u32 = 0;
            var size: u32 = 0;

            while (size < data.len) : (size += 1) {
                const c = data[size];
                allocation[size] = c;
                unicodeLength += @boolToInt((c & 0xC0) != 0x80);
            }

            return .{
                ._allocator = allocator.*,
                ._length = unicodeLength,
                ._size = size,
                ._buffer = allocation,
            };
        }

        return empty;
    }
};

pub fn main() void {
  var allocator = Allocator{};
  var string = String.load(allocator, "Hello, world");
}
```

Had a look to make sure another report similar to this didn't already exist and couldn't find any.

g-w1:
Closes #9346

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -4604,7 +4604,7 @@ fn tryExpr(
     const astgen = parent_gz.astgen;
 
     const fn_block = astgen.fn_block orelse {
-        return astgen.failNode(node, "invalid 'try' outside function scope", .{});
+        return astgen.failNode(node, "'try' outside function scope", .{});
     };
 
     if (parent_gz.in_defer) return astgen.failNode(node, "'try' not allowed inside defer expression", .{});
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -6167,6 +6167,10 @@ fn ret(gz: *GenZir, scope: *Scope, node: ast.Node.Index) InnerError!Zir.Inst.Ref
     const node_datas = tree.nodes.items(.data);
     const node_tags = tree.nodes.items(.tag);
 
+    if (astgen.fn_block == null) {
+        return astgen.failNode(node, "'return' outside function scope", .{});
+    }
+
     if (gz.in_defer) return astgen.failNode(node, "cannot return from defer expression", .{});
 
     const defer_outer = &astgen.fn_block.?.base;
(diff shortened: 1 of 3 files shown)


src/AstGen.zig
```diff
@@ -6167,6 +6167,10 @@ fn ret(gz: *GenZir, scope: *Scope, node: ast.Node.Index) InnerError!Zir.Inst.Ref
     const node_datas = tree.nodes.items(.data);
     const node_tags = tree.nodes.items(.tag);
 
+    if (astgen.fn_block == null) {
+        return astgen.failNode(node, "invalid 'return' outside function scope", .{});
```
andrewrk, line 6171: "invalid" is redundant since it's a compile error; the fact that something invalid happened is implied:

```suggestion
        return astgen.failNode(node, "'return' outside function scope", .{});
```

andrewrk (changes requested):
Nice thank you for this. I have a suggestion on the wording of the error message.

src/AstGen.zig
g-w1: I copied this from `invalid 'try' outside function scope`. Should that be changed to `'try' outside function scope` also? Also that is never tested, so I could add a test for that.

andrewrk: Sounds good thank you!

g-w1: It is done :)

PR #9358, opened 2021-07-12
Remove Stage1AirInst::owner_bb, use zir owner instead.

SpexGuy:


andrewrk (approved):
Nice find! Looks great assuming it passes the test suite.

PR #9360, opened 2021-07-12
ci: add scripts for netbsd

### LINKED ISSUE #1984: enable NetBSD Continuous Integration
Zig currently uses sr.ht for FreeBSD CI. Although it does not have Tier 1 Support yet (#1759) at least the behavior tests are run in various configurations. See the file `.builds/freebsd.yml`.

~sr.ht also [has support for NetBSD](https://man.sr.ht/builds.sr.ht/compatibility.md#netbsd) and so it would be pretty straightforward to get some level of CI testing for NetBSD going.~ It seems sr.ht dropped support

lun-4:
Closes #1984. :tada: 

I'm unable to test the codepath for s3 uploading (for good reasons, I shouldn't be able to), so it wasn't tested. The compiler test suite was, and [it passes](https://builds.hut.lavatech.top/~lun4/job/10611)!

Most of this script is a copypaste from the FreeBSD one save for one change: Forcing clang as the C compiler. I was unable to run the build with the default GCC shipped in NetBSD, due to getting thousands of "undefined reference to {llvm symbol}" errors that I couldn't find a root cause for.


andrewrk:
Want to go ahead and make the respective `ci/srht/update_download_page` changes too? 

And then the final step will be making a change [in the website repo](https://github.com/ziglang/www.ziglang.org/blob/7ccaa75ead81627629bb680d426bab45e437d732/config.toml#L23-L37)

andrewrk:
I think we also want an entry in the [test matrix](https://github.com/ziglang/zig/blob/7ef85468265ecbc53efa18f67a5bd5ef46b8c7fb/test/tests.zig#L38)

PR #9365, opened 2021-07-12
std/hash_map: fix ensureUnusedCapacity() over-allocating

ifreund:
Currently this function adds the desired unused capactiy to the current
total capacity instead of the current used capactiy.


jedisct1:
Oh, good catch! 👍

andrewrk (approved):

PR #9368, opened 2021-07-12
parser: require block in suspend expression

### LINKED ISSUE #9367: Runtime assertion fails with `suspend;`
Hi!

Thanks for building zig!
I am trying to learn zig and came across this problem which confused me (I wrote `suspend;` instead of `suspend {}`).

This file:
```zig
pub fn main() void {
  suspend;
}
```
hits the following assertion and fails:
https://github.com/ziglang/zig/blob/3063f0a5ed373947badd0af056db310283c76e37/src/AstGen.zig#L971
Which was added by this commit https://github.com/ziglang/zig/commit/15e891823e34ad57ccc5f1636323c5b7e718d44f

This results in a segfault with zig 0.9-dev.401+cff7ecee0 from ziglang.org and zig master whereas there is the expected error message with zig 0.8.

I am not sure if i am doing something wrong but shouldn't the commit be reverted to provide an error instead of hard failing?

g-w1:
Closes #9367

### DIFF
--- a/lib/std/zig/parse.zig
+++ b/lib/std/zig/parse.zig
@@ -885,11 +885,7 @@ const Parser = struct {
             },
             .keyword_suspend => {
                 const token = p.nextToken();
-                // TODO remove this special case when 0.9.0 is released.
-                const block_expr: Node.Index = if (p.eatToken(.semicolon) != null)
-                    0
-                else
-                    try p.expectBlockExprStatement();
+                const block_expr = try p.expectBlockExprStatement();
                 return p.addNode(.{
                     .tag = .@"suspend",
                     .main_token = token,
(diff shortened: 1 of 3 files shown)


Vexu (changes requested):
This makes the rewriting logic in `render.zig` useless, so either you should remove the rewriting logic or make `AstGen` temporarily support blockless `suspend` (preferable).

g-w1:
From irc:
```
16:30 <g-w1> andrewrk: https://github.com/ziglang/zig/issues/9367 is caused by // TODO remove this special case when 0.9.0 is released. const block_expr: Node.Index = if (p.eatToken(.semicolon) != null) 0 else try p.expectBlockExprStatement(); this in the parser
16:30 <g-w1> should the commit be reverted, or the special case in the parser be removed?
16:37 <andrewrk> `suspend;` should now be a parse error
16:37 <g-w1> thanks.
```
I will change the logic in render.zig.

g-w1:
The netbsd ci is successful, but it not relaying to github. I think all the other ci's are passing.

PR #9369, opened 2021-07-12
A few minor compile error fixes for the stage2-air work

LewisGaul:
@andrewrk 

This PR just fixes a couple of simple compile errors due to partial changes (no need to merge if unhelpful/already addressed).

I've been looking around the code trying to work out what changes need to be made, guided by the remaining compile errors. Below is my understanding of the errors I'm currently seeing, I'd welcome any guidance on which of these might be reasonable to try and tackle if you can spare the time! I don't want to get in the way though.

---

`./src/codegen/spirv.zig:635:35: error: no member named 'air' in struct 'codegen.spirv.DeclGen'`

Various references to `self.air` have been added in `spirv.DeclGen` - I'm guessing an `air` field needs to be added?

---

`./src/Module.zig:3597:44: error: use of undeclared identifier 'ir'`

This is in `analyzeFnBody()`, where there's some use of the old AIR layout:
```zig
const arg_inst = try arena.allocator.create(ir.Inst.Arg);
arg_inst.* = .{
    .base = .{
        .tag = .arg,
        .ty = param_type,
        .src = .unneeded,
    },
    .name = undefined, // Set in the semantic analysis of the arg instruction.
};
```
This is passed to `Sema` in the `param_inst_list` field, which has the following comment:
```zig
/// For now, AIR requires arg instructions to be the first N instructions in the
/// AIR code. We store references here for the purpose of `resolveInst`.
/// This can get reworked with AIR memory layout changes, into simply:
/// > Denormalized data to make `resolveInst` faster. This is 0 if not inside a function,
/// > otherwise it is the number of parameters of the function.
/// > param_count: u32
param_inst_list: []const Air.Inst.Index,
```
I'm not sure exactly what needs doing here!

---

`./src/codegen.zig:573:53: error: no member named 'body' in struct 'Module.Fn'`

`Module.Fn` previously had a `body: ir.Body` field, but this was removed. I'm guessing `codegen.Function(...)` should be getting the body some other way - would this just be using `Air.getMainBody()`, or should it be getting an AIR instruction index corresponding to the start of the function..?

There are added uses of `self.air` inside `codegen.Function`, so I'm guessing this field should be added too (passed into or fetched in `generateSymbol()`)?

---

./src/codegen.zig:726:39: error: no member named 'liveness' in struct 'codegen.Function(std.target.Arch.arm)'

There is no `liveness` field but code has been added that assumes there is - guessing this needs adding, as with the `air` field mentioned above?


andrewrk:
I think I'm just a couple hours away from having the compile errors worked out (with some commented out code and panic(TODO)s). Then I think it will be much more clear how to contribute!

andrewrk:
latest commit in the branch has compile errors resolved if you use `-Dskip-non-native`

PR #9370, opened 2021-07-13
linux: add mknod and mknodat syscalls

Biolunar:
mknod and mknodat are used among other things to create named pipes (FIFOs) on Linux.

I used the same type `u32` for the `mode` parameter as the mkdir and mkdirat syscalls, although that parameter should be a `u16` as seen here: https://elixir.bootlin.com/linux/v5.13/source/include/linux/types.h#L19 which is the type used in the syscall https://elixir.bootlin.com/linux/v5.13/source/fs/namei.c#L3767
Any reasons it’s a `u32` here?


andrewrk:
> Any reasons it’s a `u32` here?

Nope looks like it should change to match the kernel. Good find!

PR #9377, opened 2021-07-14
astgen: catch multiple unused vars at the same time

g-w1:
Gotta Catch 'Em All!

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -2286,6 +2286,7 @@ fn checkUsed(
 ) InnerError!void {
     const astgen = gz.astgen;
 
+    var dirty: InnerError!void = {};
     var scope = inner_scope;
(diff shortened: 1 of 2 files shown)


Vexu:
My understanding was that this is undesireable since you might get multiple errors for one bug:
```zig
var foo_a: u32 = 1;
var foo_b: u32 = 2;
// doSomethingWith(foo_a, foo_b); Oops forgot to uncomment this
```

g-w1:
Imo this is better, because it allows you to fix lots at once, rather than one at a time. Right now, all you get is this 
image
, so you fix `b`, save, and then have to fix `a`. With this change, you can fix both at the same time.

g-w1:
Closing, @andrewrk thinks that the status-quo is good enough for now.

PR #9387, opened 2021-07-15
macho+zld: bringing zld to stage2 MachO backend - simplicity is key!

kubkon:
This effort is not yet finished, but I thought I'll provide a running update on what is happening on the Zig linker front. I am currently in the process of marrying `zig ld` with stage2 MachO backend which is the end-game for Zig's linker evolution: a linker capable of handling both traditional (object files) and incremental linking contexts. As it turns out, handling both with one codebase requires some careful planning on which state to preserve, which can be discarded between updates, how to allocate final virtual memory space in the traditional sense, etc.

In the process, I have rewritten the `Zld`, the traditional linker, pretty much from scratch. This time around, the basic linking unit is based around the same unit used by the incremental linker, referred to as an *atom* (the name is actually also used by Apple's `ld64` linker, and the *atom* is used there as the smallest unit of linking too, so we are on the right path with MachO at least). Treating everything around this concept required a rewrite of how we handle objects and symbols, how they are parsed, etc. However, it was definitely worth the effort as now we have a linker which fundamental design is capable of fast linking treating every input section in an object file as an *atom* (which is pretty much how it was done until now), but is also able to parse sections into smaller units where each unit corresponds to an actual symbol from the input object file symbol table. The latter opens us up for the possibility of performing some fairly advanced whole-program optimisation techniques for release builds such as dead-code stripping, etc.

I am currently in the process of figuring out how to combine the two codebases with all of the above in mind, and I am really close to achieving this, however, there are still several bits left that I need to work out/reimplement to make the merging of incremental with traditional possible. I actually took a stab at tweaking incremental to match the ideas from traditional linker in another branch https://github.com/ziglang/zig/commit/59923c59279bd6060e1f5eb2fc42687da8d8bce5 but I learnt along the way, that in its simplicity, the incremental linker had worked out several bits in an optimal way (such as symbol management), and now I'm trying to port those over to `Zld` to be able to finally merge the two into one.

ToDo:
- [x] rewrite `Zld` around the concept of *atom*s (currently referred to as `TextBlock` for historical reasons, which is due to change for all linkers in Zig)
- [x] simplify symbols management in `Zld` so that we don't refer to resolved symbols by pointer, but simply an index to the global symbol table - this concept is already present in the incremental linker and I believe it is optimal from both design and memory usage perspective; in essence, this point will answer question whether an abstraction for a symbol is needed at all, or whether it is possible to rely on the simplest unit `macho.nlist_64` (I'm leaning towards the latter but the definite answer remains to be seen)
- [x] simplify relocation parsing - like above, all we would need was some clever manipulation of symbol mapping from object to final binary, and defer parsing embedded addends, etc. to resolution time
- [ ] work out how to allocate final binary for both traditional and incremental contexts (if it is even possible)
- [ ] (optional) GOT entries can actually be represented as *atom*s themselves which would simplify the overall linker design if inserting entries into GOT was done around the concept of an *atom*
- [ ] (optional) I actually believe the point above about GOT handling can also be applied to stubs in `__TEXT,__stubs` section


src/link/MachO.zig
```diff
@@ -4386,25 +4368,43 @@ fn allocateTextBlock(self: *MachO, text_block: *TextBlock, new_block_size: u64,
...
+    });
+
     const stubs_index = @intCast(u32, self.stubs.items.len);
     try self.stubs.append(self.base.allocator, import_sym_index);
     try self.stubs_map.putNoClobber(self.base.allocator, import_sym_index, stubs_index);
-    return resolv;
+
+    // TODO discuss this. The caller context expects codegen.InnerError{ OutOfMemory, CodegenFail },
+    // which obviously doesn't include file writing op errors. So instead of trying to write the stub
+    // entry right here and now, queue it up and dispose of when updating decl.
```
andrewrk, line 4401: The linker is supposed to handle file writing errors, and add a compile error to Module in the appropriate place (there are several places that track compile errors) and then return `error.CodegenFail`.

kubkon: Perfect! Exactly as I imagined it in my head!

PR #9388, opened 2021-07-15
Remove/replace std.Target.current and std.builtin deprecated fields with @import("builtin") fields.

### LINKED ISSUE #9321: Remove `std.Target.current` in favor of `@import("builtin").target`?
I see `std.Target.current` used a lot and it sticks out as strange to me, but maybe there's something I'm not getting. What is strange is that `target.zig`, which is otherwise just type definitions and pure functions, has this one little re-exported "static" value in the middle of it. It also means there are at least two places you can grab the same thing from. I feel that `@import("builtin").target` is a more intuitive place to get it from (`@import("builtin")` being where you'd get stuff that changes depending on build settings). It's also easier on the brain that "builtin" import "target", rather than both importing each other.

edit: I think I see the story, `@import("builtin").target` didn't even exist in 0.7.1.  Maybe my proposal  is  something that is already planned (deprecation and removal of `std.Target.current`).

antlilja:
Closes #9321.

Replaces ```std.Target.current``` with ```@import("builtin")``` fields. 

Removes the current field from the Target struct and deprecated std.builtin fields.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,5 +1,4 @@
 const std = @import("std");
-const builtin = std.builtin;
 const Builder = std.build.Builder;
 const tests = @import("test/tests.zig");
 const BufMap = std.BufMap;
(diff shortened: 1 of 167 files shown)


dbandstra:
`std.builtin.target` is actually deprecated. It should be `builtin.target` by way of `const builtin = @import("builtin")`.

=> https://github.com/ziglang/zig/blob/master/lib/std/builtin.zig#L8-L28
=> https://ziglang.org/download/0.8.0/release-notes.html#importbuiltin-no-longer-re-exports-stdbuiltin

antlilja:
@dbandstra I've replaced all the occurrences of the deprecated ```std.builtin``` fields with ```@import("builtin")```fields .

andrewrk:
Thanks for the PR. I had to redo all the work myself because I let this sit too long and now it has hella conflicts. Sorry about that.

Issue #265, opened 2017-02-23
proposal: type for null terminated pointer


andrewrk:
* [Progress](https://github.com/ziglang/zig/issues/265#issuecomment-394939353)
-----

Currently the type of `c"aoeu"` is `*const u8`.

Instead, the type should indicate that the pointer is null terminated. Here are two ideas to represent that:
 * `*0 const u8`
 * `*null const u8`

This type would be implicitly castable to `*const u8`. You can explicitly cast the other way, and in debug mode this inserts a safety check to make sure there actually is a null byte there.

It should probably work for any type that supports `T == 0` or `T == null`.

We want to steer users away from this type and instead use `[]const u8`, which includes a pointer and a length. However, we still have to deal with null terminated things from C land, which makes this useful, and some kernel interfaces. For example, we currently have this:

```zig
pub fn open_c(path: *const u8, flags: usize, perm: usize) -> usize {
    arch.syscall3(arch.SYS_open, usize(path), flags, perm)
}

pub fn open(path: []const u8, flags: usize, perm: usize) -> usize {
    const buf = @alloca(u8, path.len + 1);
    @memcpy(&buf[0], &path[0], path.len);
    buf[path.len] = 0;
    return open_c(buf.ptr, flags, perm);
}
```

Having the `open_c` prototype be `*0 const u8` would make it more type-safe. Further, we could provide an `open` function that supported either type for `path`, and if it happened to be null terminated then it could avoid the stack allocation.

We could also make the type of string literals be `[]0 const u8` meaning that the pointer value for the slice has a 0 after the last byte. The length would still indicate the memory before the null byte. If you slice this type then the pointer component would change from `*0 const u8` to `*const u8`.

It would be extra helpful if automatic .h import could identify when a pointer in a function is supposed to be null-terminated, and we could emit a compile error if the user passes a pointer that is not null terminated. I'm not sure how we could detect this automatically though.

thejoshwolfe:
> in debug mode this inserts a safety check to make sure there actually is a null byte there.

How would that work? Wouldn't that be an unbounded linear search?

andrewrk:
Maybe if you cast from `[]T` to `[]0 T` then `len` then we check for 0 in the last spot (since we have len) and then in the new slice, len is -1.

andrewrk:
Proposal rejected in order to discourage use of null terminated things.

AndreaOrru:
How do you plan to interface with C-style null-terminated strings?

[...]

PR #9392, opened 2021-07-16
Docs fix array/pointer/slice type coercion section

BuzzwordChief:
The type coercion removed in #265 ([X]T => [] const T) has to be removed from the docs.

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -5337,12 +5337,14 @@ test "implicit cast to comptime_int" {
 }
       {#code_end#}
       {#header_close#}
-      {#header_open|Type Coercion: Arrays and Pointers#}
-      {#code_begin|test|coerce_arrays_and_ptrs#}
+      {#header_open|Type Coercion: Slices, Arrays and Pointers#}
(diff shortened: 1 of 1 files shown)


dbandstra:
Actually the example code is correct but the test names are wrong.

Heading: `Type Coercion: Arrays and Pointers` -> `Type Coercion: Slices and Pointers to Arrays` ?

Comment mentioning issue 265 should be amended.

Test name: `[N]T to []const T` -> `*const [N]T to []const T`

Test name: `[N]T to E![]const T` -> `*const [N]T to E![]const T`

Test name: `[N]T to ?[]const T` -> `*const [N]T to ?[]const T`

This section could also be fleshed out to talk about sentinel terminated arrays explicitly. But the coercion section is already pretty gigantic...

BuzzwordChief:
You are absolutly right! I converted slices/arrays/pointers all night and in my opinion this deserves a few more examples (espcially bc arrays are so differeny to C's). I will fix that up and write one test converting string literals to include sentinel terminated * conversions.

PR #9405, opened 2021-07-17
Added FormatOptions handling for fmtIntSize

### LINKED ISSUE #9184: std.fmt.fmtIntSizeDec() and fmtIntSizeBin() do not consider FormatOptions
Ultimately `formatSizeImpl()` is used to implement `fmtIntSizeDec()` or `fmtIntSizeBin()` and it does not consider FormatOptions and probably should.

motivating example (from @espindola  via IRC):
```zig
const std = @import("std");

pub fn main() !void {

    var buf: [32]u8 = undefined;
    var slice = buf[0..];
    const partial = try std.fmt.bufPrint(slice, "{}", .{std.fmt.fmtIntSizeDec(42 * 1000)});

    try stdout.print("X {s:<14} Y\n", .{partial});
    try stdout.print("X {:<14} Y\n", .{std.fmt.fmtIntSizeDec(42 * 1000)});
}
```

```sh
$ zig run foo.zig
X 42kB           Y
X 42kB Y
```

@andrewrk you mentioned in IRC:
```
it should also be renamed to remove the redundant fmt
```

There is a pattern in `std.fmt` where all `pub fn fmt*()` functions return a `Formatter()` object; a grep follows. Do you want these  renamed like `fmtFirstSecond()` → `firstSecond()`?
```
pub fn fmtSliceHexLower(bytes: []const u8) std.fmt.Formatter(formatSliceHexLower) {
pub fn fmtSliceHexUpper(bytes: []const u8) std.fmt.Formatter(formatSliceHexUpper) {
pub fn fmtSliceEscapeLower(bytes: []const u8) std.fmt.Formatter(formatSliceEscapeLower) {
pub fn fmtSliceEscapeUpper(bytes: []const u8) std.fmt.Formatter(formatSliceEscapeUpper) {
pub fn fmtIntSizeDec(value: u64) std.fmt.Formatter(formatSizeDec) {
pub fn fmtIntSizeBin(value: u64) std.fmt.Formatter(formatSizeBin) {
```

mitchk92:
Fixed #9184 only the code, didn't touch the naming of the functions, because I wasn't sure what to do with that as this is my first PR.

If i've missed anything let me know :)

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -827,7 +827,7 @@ fn formatSizeImpl(comptime radix: comptime_int) type {
         ) !void {
             _ = fmt;
             if (value == 0) {
-                return writer.writeAll("0B");
+                return formatBuf("0B", options, writer);
             }
 
             const mags_si = " kMGTPEZY";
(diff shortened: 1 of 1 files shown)


lib/std/fmt.zig
```diff
@@ -845,19 +845,25 @@ fn formatSizeImpl(comptime radix: comptime_int) type {
                 1024 => mags_iec[magnitude],
                 else => unreachable,
             };
-
-            try formatFloatDecimal(new_value, options, writer);
-
+            var fmt_buffer: [64]u8 = undefined; // 64bit int can't be formatted greater than 64 chars
```
Vexu, line 848: This isn't correct and can lead to a crash on input like `{:.65}`, is there a reason you can't directly use the provided `writer`?

Vexu (changes requested):

lib/std/fmt.zig
mitchk92: You need to know the size of the formatted decimal before you can work out the padding for the centre and right padding.
Otherwise you'd need to run the formatter twice, once in a dry run and a second time to actually format it. And that looked like a much larger scope than I think intended.

mitchk92:
I can't find anyway without a decent refactoring effort, to make it work with arbitrary size precision inputs.

PR #9407, opened 2021-07-17
Document Zig's built-in linker options

jwinnie:
This PR adds documentation for the `ld.lld`, `lld-link`, and `wasm-ld` options, which are currently exposed but not documented.

Note that I haven't tested this locally but it should work fine regardless since it only adds a few lines to the help string.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -56,12 +56,15 @@ const normal_usage =
     \\  test             Create and run a test build
     \\  translate-c      Convert C code to Zig code
     \\
-    \\  ar               Use Zig as a drop-in archiver
     \\  cc               Use Zig as a drop-in C compiler
     \\  c++              Use Zig as a drop-in C++ compiler
+    \\  ld.lld           Use Zig as a drop-in ELF linker (Linux)
+    \\  lld-link         Use Zig as a drop-in COFF linker (Windows)
+    \\  wasm-ld          Use Zig as a drop-in WebAssembly linker
+    \\  ar               Use Zig as a drop-in archiver
+    \\  ranlib           Use Zig as a drop-in ranlib
     \\  dlltool          Use Zig as a drop-in dlltool.exe
     \\  lib              Use Zig as a drop-in lib.exe
-    \\  ranlib           Use Zig as a drop-in ranlib
     \\
     \\  env              Print lib path, std path, cache directory, and version
     \\  help             Print this help and exit


g-w1:
Why not just use `zig ld`? I think these are only supposed to be used internally.

jwinnie:
> Why not just use `zig ld`? I think these are only supposed to be used internally.

@g-w1 `error: unknown command: ld`

This is on Zig 0.8.0.

g-w1:
Ah, sorry, I meant, build-exe or build-lib, or build-obj. You can just pass the things you would like to link to args as to those programs.

jwinnie:
> Ah, sorry, I meant, build-exe or build-lib, or build-obj. You can just pass the things you would like to link to args as to those programs.

Is `build-lib` a drop-in LD though?

Like, can I set `LD=zig build-lib`?

g-w1:
I do not think so, but also see https://github.com/ziglang/zig/issues/8755.

jwinnie:
The use case here is that I'm making a Linux distribution and a lot of build scripts (e.g. Autotools, CMake) automatically use the system linker rather than Zig's linker. I can override this by setting `LD` to something... which I'm currently setting to `zig ld.lld` and it seems to work.

jwinnie:
The problem is I had to do a lot of digging to find this information (including reading your IRC messages in the IRC log). I'm just submitting this patch so maybe the next person who is crazy enough to try to make a Linux distro based on Zig will have it slightly easier 😅

g-w1:
I think this is a workaround, it might not be that hard to just implement https://github.com/ziglang/zig/issues/8755 depending on the target os (may be a bit harder to hook up the self-hosted linker backends, but for linux, it should just be forwarding the args to `argv[0] ld.lld`.)

alexrp:
I think it's going to be a long while still until `zig ld` can replace LLD on all target platforms. Wouldn't it make sense to document these until then? The command line interface isn't set in stone until 1.x anyway.

g-w1:
Zig ld != self-hosted linker. You can still shell out to ld.lld in zig ld (as I said).

alexrp:
Oh, sorry, I misunderstood your earlier comments.

Are we sure that a single `zig ld` interface covers all command line options people might want to use? `ld.lld` and `lld-link` have radically different CLIs after all.

g-w1:
From the issue:
> some projects may want to use Zig as a **drop-in replacement** for other linkers.

andrewrk:
These are internal, intentionally not part of the publicly documented CLI. Please see #8755

PR #9408, opened 2021-07-18
[stage2-air]: cbe: fix not (it is a ty_op, not un_op)

g-w1:
The diff looks a bit weird as I deleted airUnOp as it is not used after this.


LewisGaul:
I've applied this patch and it does appear to fix the CBE test that was previously failing 👍

PR #9417, opened 2021-07-19
cmake: also check Homebrew install paths when looking for LLVM

FnControlOption:
- On Intel Macs, the path is `/usr/local/opt/llvm@12`
- On Silicon Macs, the path is `/opt/homebrew/opt/llvm@12`

This makes specifying CMAKE_PREFIX_PATH optional for Homebrew LLVM.

### DIFF
--- a/cmake/Findclang.cmake
+++ b/cmake/Findclang.cmake
@@ -14,6 +14,8 @@ find_path(CLANG_INCLUDE_DIRS NAMES clang/Frontend/ASTUnit.h
     /usr/lib/llvm-12.0/include
     /usr/local/llvm120/include
     /usr/local/llvm12/include
+    /usr/local/opt/llvm@12/include
+    /opt/homebrew/opt/llvm@12/include
     /mingw64/include
 )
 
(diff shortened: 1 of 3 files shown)


sreehax:
Just letting you know that on my machine (Apple Silicon), the homebrew LLVM prefix is in /opt/homebrew/opt rather than in /usr/local/opt. The prefix for LLVM can be found by running `brew --prefix llvm`, and I just provide the `-DCMAKE_PREFIX_PATH=$(brew --prefix llvm)` argument to cmake.

FnControlOption:
Right, I forgot about that. Thank you for reminding me.

kubkon:
Thanks for the PR @FnControlOption! Hmm, I think I'm against merging this change since this would imply that Homebrew is de facto standard way of obtaining a copy of LLVM on macOS which I don't think is true. Also, the current way of adding `-DCMAKE_PREFIX_PATH=$(brew --prefix llvm)` works fine, is hardly a hindrance, and is well documented in our wiki [here](https://github.com/ziglang/zig/wiki/Building-Zig-From-Source#option-a-use-your-system-installed-build-tools). Having said that, @andrewrk your thoughts on this change?

andrewrk:
I'm completely fine with this change, because it will make homebrew users have a better experience, while not hurting any other experience - `-DCMAKE_PREFIX_PATH` is used to specify the path when building from source and that use case would not be affected by this. I trust @kubkon's judgement but given that he passed the :fire: :potato: to me, I'm going to merge it.

PR #9420, opened 2021-07-19
Fixed wrong "unable to load" error for non-existing import files

kristoff-it:
- Changed ZIR encoding of `import` metadata from having instruction
  indexes to storing token indexes.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -2315,7 +2315,7 @@ const AstGenSrc = union(enum) {
     root,
     import: struct {
         importing_file: *Module.Scope.File,
-        import_inst: Zir.Inst.Index,
+        import_tok: std.zig.ast.TokenIndex,
     },
 };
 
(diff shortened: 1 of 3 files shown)


src/Compilation.zig
```diff
@@ -3402,14 +3405,7 @@ fn detectLibCIncludeDirs(
...
     // native abi, fall back to using the system libc installation.
-    // On windows, instead of the native (mingw) abi, we want to check
-    // for the MSVC abi as a fallback.
-    const use_system_abi = if (std.Target.current.os.tag == .windows)
-        target.abi == .msvc
-    else
-        is_native_abi;
-
-    if (use_system_abi) {
+    if (is_native_abi) {
```
andrewrk, line 3408: This change and the one below in this file look like an accident, and what was causing test failures on your machine as well as on the CI.

src/Compilation.zig
```diff
@@ -3565,7 +3568,6 @@ pub fn generateBuiltinZigSource(comp: *Compilation, allocator: *Allocator) Alloc
         \\pub const stage2_arch: std.Target.Cpu.Arch = .{};
         \\/// Temporary until self-hosted supports the `os.tag` value.
         \\pub const stage2_os: std.Target.Os.Tag = .{};
-        \\
```
andrewrk, line 3568: this also looks like a stray change

kristoff-it: sorry, fixed now

andrewrk (approved):

PR #9422, opened 2021-07-20
Removed dead code in std.os.read

charlieman:
This function never reaches this point (the index variable doesn't exist).

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -385,7 +385,6 @@ pub fn read(fd: fd_t, buf: []u8) ReadError!usize {
             else => |err| return unexpectedErrno(err),
         }
     }
-    return index;
 }
 
 /// Number of bytes read is returned. Upon reading end-of-file, zero is returned.


marler8997:
Interesting the compiler doesn't detect this dead code. I don't see any breaks out of the `while (true)` loop so this seems like a bug.

ifreund (approved):
LGTM, thanks!

I don't see any bug here, every case aside from EINTR will either assert or return.h

dbandstra:
But EINTR continues, nothing is actually breaking out of the loop (to the dead code)?

marler8997:
@ifreund I mean to say that it looks like a bug that the compiler is not detecting and asserting a compiler error from this dead code.

ifreund:
> @ifreund I mean to say that it looks like a bug that the compiler is not detecting and asserting a compiler error from this dead code.

Ahah, yes that makes much more sense.

PR #9428, opened 2021-07-20
linux stdlib: fix definition of RW flags

Biolunar:
This fixes the syntax of the definition of the RWF_* constants which are used in the preadv2 and pwritev2 syscalls.


jedisct1:
Thank you!

andrewrk:
Wow, leftover from waaaay back when type coercion was done via `a(b)` syntax.

PR #9430, opened 2021-07-21
Correct hasUniqueRepresentation for vectors

### LINKED ISSUE #9333: meta.trait.hasUniqueRepresentation handles integer types incorrectly
```Zig
comptime { std.debug.assert(@sizeOf(@Vector(3, i16)) == 6); }
```
fails with Zig 0.8.0. I don't know if this is intended behavior, but it seems to me that the padding should not be counted towards the size. `mem.asBytes`, for example, results in a slice containing undefined data, and an AutoHashMap with `@Vector(3, i16)` keys is silently broken.

InterplanetaryEngineer:
Closes #9333.


andrewrk:
Thank you!

PR #9437, opened 2021-07-22
Fixed compilation with std.thread.Condition without libc

ayende:
Right now using `std.thread.Condition` will error because of compilation errors. Looks like the API for `Mutex` changed and `AtomicCondition` wasn't updated as well.

See: https://zigforum.org/t/condition-variables-are-non-workable/651

```
~/zig/lib/std/Thread/Condition.zig:160:14: error: no member named 'unlock' in struct 'std.Thread.Mutex'
        mutex.unlock();
             ^
```

This PR introduce a new api `force_release_internal` and tries to mimics the same pattern that was already in the code.
I don't like this approach, but I'm not familiar enough with Zig to provide a better alternative.

### DIFF
--- a/lib/std/Thread/Condition.zig
+++ b/lib/std/Thread/Condition.zig
@@ -157,9 +157,9 @@ pub const AtomicCondition = struct {
             @atomicStore(bool, &cond.pending, true, .SeqCst);
         }
 
-        mutex.unlock();
+        mutex.force_release_internal();
         waiter.data.wait();
-        mutex.lock();
+        _ = mutex.acquire();
     }
 
     pub fn signal(cond: *AtomicCondition) void {
--- a/lib/std/Thread/Mutex.zig
+++ b/lib/std/Thread/Mutex.zig
@@ -50,6 +50,15 @@ pub fn acquire(m: *Mutex) Impl.Held {
     return m.impl.acquire();
 }
 
+/// This is required to get AtomicCondition to work properly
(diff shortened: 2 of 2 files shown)


andrewrk:
Test case please! I think I want to solve this a different way, so in fact the test case is the most valuable part of this PR.

ayende:
How do I test the compilation process itself? The code here fails to compile unless you have `libc` linked in.

```
const std = @import("std");

pub fn main() void {
    var mutex : std.Thread.Mutex = .{};
    var cond : std.Thread.Condition =  .{};

    cond.wait(&mutex);
}
```

Will result in :

```
~/zig/lib/std/Thread/Condition.zig:160:14: error: no member named 'unlock' in struct 'std.Thread.Mutex'
        mutex.unlock();
             ^
````

I'm not sure how to add a compilation test, I think that I need to add it here: https://github.com/ziglang/zig/blob/master/test/compile_errors.zig#L5

But I don't see how to run the test cases there to verify that it indeed fails

andrewrk:
Sorry for the delay in reviewing this. Looks like the problem was fixed in 008b0ec5e58fc7e31f3b989868a7d1ea4df3f41d.


> How do I test the compilation process itself? The code here fails to compile unless you have `libc` linked in.

No need to do anything fancy, just need to add a few tests to the file. Zig's test suite is already testing builds of the std lib without libc.

I filed #10636 for the missing tests.

Issue #6912, opened 2020-11-01
Proposal: add --git flag to init-lib and init-exe


g-w1:
If the ``--git`` flag was passed to init-lib or init-exe, then a git repo would be initialized in the current directory, and a `.gitignore` would be added. I think cargo does this by default, but I feel that it should be an explicit option in zig.

Maybe we could also add different flags for different vcs.

tauoverpi:
What of just searching for `.git`, `.pijul`, and so on instead and updating the ignore files from there? Though as long as everything that's not source is kept in zig-cache it should suffice to just print a note on init which skips needing to keep track of the many version control systems.

jedisct1:
There's not much to ignore besides `zig-cache` to be honest.

And `zig-cache` can be ignored globally.

g-w1:
> Though as long as everything that's not source is kept in zig-cache it should suffice to just print a note on init which skips needing to keep track of the many version control systems.
What do you mean by print note?

andrewrk:
I'm making a judgement call here: no opinionated integration with any particular source control software in the zig CLI. Even for the package manager, downloading from a git URL will be a "plugin".

PR #9445, opened 2021-07-23
CLI: Add git initalization in init-exe/init-lib

chivay:
When creating a new project it may be useful to have a CLI flag for basic Git setup - initialized repo with simple .gitignore.


Vexu:
This feature has been rejected in proposal #6912.

chivay:
Oh, I wasn't aware of that. Thanks!

PR #9446, opened 2021-07-24
stage2: wasm - Use `br_table` when possible for switch

Luukdegram:
Contains the following:
- Rewrite switch_br to use `br_table` instead
- When prongs are sparse values, use if/else-chain
- Allow negative values (Emit the actual bits of integers, rather than the value it represents).
- Support multi-value prongs
- Re-enable switch test cases and fix regressions

Currently, this works by calculating the highest and lowest value of the prongs.
When the difference between them exceeds 50, we use an if/else-chain.
This is something that still requires benchmarking (in the future), LLVM seems to switch from br_table to if/else-chain between 40-45.

br_table is essentially a jump table where it takes the stack value as an index, into a vector of labels.
Each label represents a branch it will jump to, the last label is the default for any values that go out of bounds. 
As the index is limited to 2^32-1, we resort to an if/else-chain for switches over targets with a bitsize over 32.

One more edge case that had to be handled cases where we use an if/else-chain and also have multi-value prongs.
For those cases, we emit an extra block and jump out of that extra block if we match to any value. (to the body the case targets).
When no value is matched, it will instead jump back out 1 further. (to the next case).

### DIFF
--- a/src/codegen/wasm.zig
+++ b/src/codegen/wasm.zig
@@ -1275,12 +1275,17 @@ pub const Context = struct {
         const blocktype = wasm.block_empty;
         const pl_op = self.air.instructions.items(.data)[inst].pl_op;
         const target = self.resolveInst(pl_op.operand);
+        const target_ty = self.air.typeOf(pl_op.operand);
         const switch_br = self.air.extraData(Air.SwitchBr, pl_op.payload);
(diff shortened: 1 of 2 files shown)


andrewrk:
Great work!

Issue #903, opened 2018-04-07
Add SIMD Support


tiehuis:
SIMD is very useful for fast processing of data and given Zig's goals of going fast, I think we need to look at how exposing some way of using these instructions easily and reliably.

#### Inline Assembly

It is possible to do simd in inline-assembly as is. This is a bit cumbersome though and I think we should strive for being able to get any speed performances in the zig language itself.

#### Rely on the Optimizer

The optimizer is good and comptime unrolling and support helps a lot, but it doesn't provide guarantees that any specific code will be vectorized. You are at the mercy of LLVM and you don't want to see your code lose a huge hit in performance simply due to a compiler upgrade/change.

### LLVM Vector Intrinsics

LLVM supports [vector types](https://llvm.org/docs/LangRef.html#vector-type) as first class objects in it's ir. These correspond to simd instructions. This provides the bulk of the work and for us, we simply need to expose a way to construct these vector types. This would be analagous to the `__attribute__((vector))__` builtin found in C compilers.

---

If anyone has any thoughts on the implementation and or usage then that would be great since I'm not very familiar with how these are exposed by LLVM. It would be great to get some discussion going in this area since I'm sure people would like to be able to match the performance of C in all areas with Zig.

abique:
I think relying on the compiler vector type is a good solution.
Both LLVM and GCC have it. If they're not present, you can always have a generic "software" fallback.

Syntax: you need a way to describe a vector type, an idea could be:

`const value = <[]> f32 {0, 13, 23, 0.4};`

So `<[ N_ELTS ]> type` would be the bracket style for vectors in this examples.

Also vector are use essentially for arithetic so regular artimetic should work.

Importants things:
 - be able to extract a single element from a vector
 - needs some kind of shuffle vector: `@shuffle(v1, v2, index0, index1, index2, ...)
 - you should be able to do an addition or multiplication between a scalar and a vector

andrewrk:
How about adding operators for arrays? Example:

```zig
const std = @import("std");

test "simd" {
    var a = [4]i32{1, 2, 3, 4};
    var b = [4]i32{5, 6, 7, 8};
    var c = a + b;
    std.debug.assert(mem.eql(i32, c[0..], [4]i32{6, 8, 10, 12} ));
}
```

This would codegen to using vectors in LLVM.

abique:
I believe you'll find out that using arrays for "simd vector" introduces more problems than solutions, and that's why llvm and gcc went a different way.

[...]

PR #9448, opened 2021-07-25
minimum/maximum builtins

Snektron:
This pull request adds `@minimum` and `@maximum` builtin functions. The main use case for these is SIMD (See #903), though they work on scalars as well.

For floats, this uses the `llvm.maxnum` and `llvm.minnum` intrinsics. Note that x86_64 `vminps` and `vmaxps` have different semantics, and so these generate some extra instructions:
```
c5 f0 5d d0              vminps xmm2,xmm1,xmm0
c5 f8 c2 c0 03           vcmpunordps xmm0,xmm0,xmm0
c4 e3 69 4a c1 00        vblendvps xmm0,xmm2,xmm1,xmm0
c3  
  ```
  
  On other architectures (riscv and arm) this shouldn't be a problem.

### DIFF
--- a/src/stage1/ir.cpp
+++ b/src/stage1/ir.cpp
@@ -3289,6 +3289,108 @@ static void float_mod(ZigValue *out_val, ZigValue *op1, ZigValue *op2) {
     }
 }
 
+static void float_max(ZigValue *out_val, ZigValue *op1, ZigValue *op2) {
+    assert(op1->type == op2->type);
+    out_val->type = op1->type;
+    if (op1->type->id == ZigTypeIdComptimeFloat) {
+        bigfloat_max(&out_val->data.x_bigfloat, &op1->data.x_bigfloat, &op2->data.x_bigfloat);
+    } else if (op1->type->id == ZigTypeIdFloat) {
+        switch (op1->type->data.floating.bit_count) {
+            case 16:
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

**Comment by Snektron** [comment]
Oh, and i also made sure to implement the float semantics in comptime: if either argument is nan, the other is returned.

**Comment by Snektron** [comment]
CI errors seem unrelated

File: src/stage1/ir.cpp [hunk 1]
```diff
@@ -3289,14 +3289,115 @@ static void float_mod(ZigValue *out_val, ZigValue *op1, ZigValue *op2) {
...
+}
+
 static void float_negate(ZigValue *out_val, ZigValue *op) {
     out_val->type = op->type;
     if (op->type->id == ZigTypeIdComptimeFloat) {
         bigfloat_negate(&out_val->data.x_bigfloat, &op->data.x_bigfloat);
     } else if (op->type->id == ZigTypeIdFloat) {
         switch (op->type->data.floating.bit_count) {
             case 16:
-                out_val->data.x_f16 = f16_neg(op->data.x_f16);
```
nektro, line 3299: why was this line removed?

Snektron: That must have been an accident. Thanks for yhe find :)

PR #9449, opened 2021-07-25
Expose register_eventfd, register_eventfd_async, unregister_eventfd i…

ayende:
…n the IO URing API

### DIFF
--- a/lib/std/os/linux/io_uring.zig
+++ b/lib/std/os/linux/io_uring.zig
@@ -629,7 +629,6 @@ pub const IO_Uring = struct {
     /// An application need unregister only if it wants to register a new array of file descriptors.
     pub fn register_files(self: *IO_Uring, fds: []const os.fd_t) !void {
         assert(self.fd >= 0);
-        comptime assert(@sizeOf(os.fd_t) == @sizeOf(c_int));
         const res = linux.io_uring_register(
             self.fd,
             .REGISTER_FILES,
@@ -641,14 +640,14 @@ pub const IO_Uring = struct {
 
     /// Registers the file descriptor for an eventfd that will be notified of completion events on
     //  an io_uring instance.
+    /// Only a single a eventfd can be registered at any given point in time.
     pub fn register_eventfd(self: *IO_Uring, fd: os.fd_t) !void {
         assert(self.fd >= 0);
-        comptime assert(@sizeOf(os.fd_t) == @sizeOf(c_int));
         const res = linux.io_uring_register(
             self.fd,
             .REGISTER_EVENTFD,
             @ptrCast(*const c_void, &fd),
-            @intCast(u32, 1),
+            1,
         );
         try convert_registration_error(res);
     }
@@ -656,14 +655,14 @@ pub const IO_Uring = struct {
     /// Registers the file descriptor for an eventfd that will be notified of completion events on
     /// an io_uring instance. Notifications are only posted for events that complete in an async manner. 
     /// This means that events that complete inline while being submitted do not trigger a notification event.
+    /// Only a single a eventfd can be registered at any given point in time.
     pub fn register_eventfd_async(self: *IO_Uring, fd: os.fd_t) !void {
         assert(self.fd >= 0);
-        comptime assert(@sizeOf(os.fd_t) == @sizeOf(c_int));
         const res = linux.io_uring_register(
             self.fd,
             .REGISTER_EVENTFD_ASYNC,
             @ptrCast(*const c_void, &fd),
-            @intCast(u32, 1),
+            1,
         );
         try convert_registration_error(res);
     }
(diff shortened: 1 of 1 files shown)


lib/std/os/linux/io_uring.zig
```diff
@@ -636,6 +636,51 @@ pub const IO_Uring = struct {
...
             @intCast(u32, fds.len),
         );
+        try convert_registration_error(res);
+    }
+
+    /// Registers the file descriptor for an eventfd that will be notified of completion events on
+    //  an io_uring instance.
+    pub fn register_eventfd(self: *IO_Uring, fd: os.fd_t) !void {
+        assert(self.fd >= 0);
+        comptime assert(@sizeOf(os.fd_t) == @sizeOf(c_int));
```
kprotty, line 646: This check looks like it can be removed given `fd_t` is [defined as `i32`](https://github.com/ziglang/zig/blob/master/lib/std/os/bits/linux.zig#L41) which should also be the [same as `c_int`](https://www.ibm.com/docs/en/ibm-mq/9.2?topic=platforms-standard-data-types-aix-linux-windows) on linux platforms.

lib/std/os/linux/io_uring.zig
```diff
@@ -636,6 +636,51 @@ pub const IO_Uring = struct {
             @ptrCast(*const c_void, fds.ptr),
             @intCast(u32, fds.len),
         );
+        try convert_registration_error(res);
+    }
+
+    /// Registers the file descriptor for an eventfd that will be notified of completion events on
+    //  an io_uring instance.
```
kprotty, line 643: It could probably be useful to note that only one eventfd can be registered on the ring at any given point. Same for `register_eventfd_async`.

lib/std/os/linux/io_uring.zig
```diff
@@ -636,6 +636,51 @@ pub const IO_Uring = struct {
...
+    pub fn register_eventfd(self: *IO_Uring, fd: os.fd_t) !void {
+        assert(self.fd >= 0);
+        comptime assert(@sizeOf(os.fd_t) == @sizeOf(c_int));
+        const res = linux.io_uring_register(
+            self.fd,
+            .REGISTER_EVENTFD,
+            @ptrCast(*const c_void, &fd),
+            @intCast(u32, 1),
```
kprotty, line 651: This can be demoted to [`@as(u32, 1)`](https://ziglang.org/documentation/master/#as) which, unlike [`@intCast`](https://ziglang.org/documentation/master/#intCast),  has no bounds checking runtime overhead (in debug mode). That can then be demoted to just `1` if you prefer.

lib/std/os/linux/io_uring.zig
ayende: I agree, but it is already there on `reigster_files`, should I remove it on both?

kprotty: Sure.

lib/std/os/linux/io_uring.zig
```diff
@@ -629,13 +629,57 @@ pub const IO_Uring = struct {
...
             self.fd,
             .REGISTER_FILES,
             @ptrCast(*const c_void, fds.ptr),
             @intCast(u32, fds.len),
         );
+        try convert_registration_error(res);
+    }
+
+    /// Registers the file descriptor for an eventfd that will be notified of completion events on
+    //  an io_uring instance.
```
jorangreef, line 642:
```suggestion
    ///  an io_uring instance.
```

lib/std/os/linux/io_uring.zig
```diff
@@ -629,13 +629,57 @@ pub const IO_Uring = struct {
...
+            @ptrCast(*const c_void, &fd),
+            1,
+        );
+        try convert_registration_error(res);
+    }
+
+    /// Registers the file descriptor for an eventfd that will be notified of completion events on
+    /// an io_uring instance. Notifications are only posted for events that complete in an async manner. 
+    /// This means that events that complete inline while being submitted do not trigger a notification event.
+    /// Only a single a eventfd can be registered at any given point in time.
```
jorangreef, line 658:
```suggestion
    /// Only a single eventfd can be registered at any given point in time.
```

lib/std/os/linux/io_uring.zig
```diff
@@ -629,13 +629,57 @@ pub const IO_Uring = struct {
...
+        const res = linux.io_uring_register(
+            self.fd,
+            .UNREGISTER_EVENTFD,
+            null,
+            0,
+        );
+        try convert_registration_error(res);
+    }
+
+    fn convert_registration_error(res: usize) !void {
```
jorangreef, line 682: What do you think of calling this `handle_registration_result()` since that's what it's doing? i.e. It's not always converting an error. Nice abstraction by the way.

jorangreef (commented):
Thanks @ayende for adding these!

lib/std/os/linux/io_uring.zig
ayende: done

kprotty:
PR looks nice. Any changes or additions you/anyone wish to make?

ayende:
If you are asking me, then no

PR #9456, opened 2021-07-25
Use cl::expandResponseFiles to expand a response file

splhack:
This PR will address the TODO for more robust response file parsing.

https://github.com/ziglang/zig/blob/7b8cb881df7e034a8626caabf355055ee81a0fef/src/main.zig#L3647-L3648

For example, `clang` and `gcc` can handle quotes very well.

    $ cat argsfile
    '-o'
    test
    test.c
    $ gcc @argsfile

    $ cat argsfile2
    -o
    'te st'
    test.c
    $ gcc @argsfile

But `zig cc` can't.

    $ zig cc @argsfile
    error: FileNotFound
    $ zig cc @argsfile2
    error: FileNotFound

This PR will change the code to use LLVM `cl::expandResponseFiles` via C interop.

### DIFF
--- a/src/codegen/llvm/bindings.zig
+++ b/src/codegen/llvm/bindings.zig
@@ -908,7 +908,7 @@ pub const WriteImportLibrary = ZigLLVMWriteImportLibrary;
 extern fn ZigLLVMWriteImportLibrary(
     def_path: [*:0]const u8,
     arch: ArchType,
-    output_lib_path: [*c]const u8,
+    output_lib_path: [*:0]const u8,
     kill_at: bool,
 ) bool;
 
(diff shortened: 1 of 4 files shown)


alexrp:
Ah, so this was the root cause of those bizarre `FileNotFound` errors I saw when trying to use response files. Thanks a ton for looking into this.

PhaseMage:
A few weeks back, I hit this "quotes in an RSP file" issue when trying to compile gRPC using "zig cc". I didn't see this PR here, and as a fun exercise, I decided to see if I could fix it myself. 

I went the route of modifying "ArgIteratorWindows" in process.zig such that it could work with arbitrary strings (or the contents of an RSP file). Then I used this new "ArgIteratorGeneral" in main.zig to address the "TODO" listed in ClangArgIterator. It has the nice attribute of handling "RSP file" arguments in the same way it handles "cmd_line" arguments. I'll probably send out a PR by this evening.

PhaseMage:
@Vexu, here's the PR I said I'd send out: #9654

andrewrk (changes requested):
Thank you for making this patch. However, we need response file parsing to work without the LLVM dependency, in case we want to add [arocc](https://github.com/Vexu/arocc) as a C frontend.

So we could *port* the code from `cl::expandResponseFiles`, but we must not call it as a dependency.

I realize that this patch would nicer in the short-term but I'm making the long-term call here, that the way forward is to invest in our own response file parsing code.

Sorry for the trouble.

PR #9458, opened 2021-07-26
Vector support for `@popCount`, `@ctz`, and `@clz`

SuperAuguste:
This pull request adds vector support for `@popCount`, `@ctz`, and `@clz`!

See also #903


andrewrk:
Nice work @SuperAuguste!

Issue #5263, opened 2020-05-03
add "select" syntax to the language to await the first function in a given set that completes


andrewrk:
I thought I had already proposed this but I could not find it. This depends on the `cancel` proposal being accepted, which is related to #3164 but I will open an explicit proposal for it.

There is one thing missing with zig's `async` / `await` features which is the ability to respond to the first completed function. Here's a use case:

```zig
fn httpRequestWithTimeout(url: []const, timeout: u64) !Response {
    var request = async httpRequest(url);
    var timeout = async sleep(timeout);

    select {
        request => |result| {
            cancel timeout;
            return result;
        },
        timeout => {
            cancel request;
            return error.TimedOut;
        },
    }
}
```

Here's a more complicated example, which shows how select also supports arrays of frames:

```zig
fn httpRequestWithMirrors(mirrors: []const []const u8) !Request {
    const frames = try allocator.alloc(@Frame(httpRequest), mirrors.len);
    defer allocator.free(frames);
    const completed_flags = try allocator.alloc(bool, mirrors.len);
    defer allocator.free(completed_flags);

    for (frames) |frame, i| {
        frame.* = async httpRequest(mirrors[i]);
        completed_flags[i] = false;
    }

    defer for (frames) |frame, i| {
        if (!completed_flags[i]) {
            cancel frame.*;
        }
    };

    var in_flight_count: usize = mirrors.len;
    while (true) {
        select {
            frames => |result, i| {
                completed_flags[i] = true;
                if (result) |payload| {
                    return payload;
                } else |err| {
                    in_flight_count -= 1;
                    if (in_flight_count == 0)
                        return err;
                }
            },
        }
    }
}
```

Here's another use case: https://github.com/ziglang/zig/blob/9b788b765c1557a414710872fa9c11fce1f8c504/lib/std/event/batch.zig#L68-L86

As an alternative syntax, we could replace `select` with `await`. Since only labeled blocks can return values, `await {` is unambiguously the "select" form of `await`.

rohlem:
I can see the conciseness of assuming `cancel` in the code examples, but the two features are **not intrinsically linked**... right?
From what I understand, if we exposed a field `.running` of `@Frame` types (as mentioned in https://github.com/ziglang/zig/issues/3164#issuecomment-527504887), we could already implement it in **pure userland** code. So am I correct in thinking that this is only syntactic sugar to more cleanly associate the frames with the code blocks handling them?

That is to say
```zig
[async/noasync/] select{
[...]

PR #9459, opened 2021-07-26
Add @select

Snektron:
```zig
@select(
    comptime T: type,
    pred: std.meta.Vector(len, bool),
    a: std.meta.Vector(len, T),
    b: std.meta.Vector(len, T)
) std.meta.Vector(len, T)
```

Constructs a vector from a & b, based on the values in the predicate vector. For indices where the predicate value is true, the corresponding element from the a vector is selected, and otherwise from.

See also #903

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -7125,6 +7125,7 @@ fn func(y: *i32) void {
       an integer or an enum.
       </p>
       {#header_close#}
+
       {#header_open|@bitCast#}
       <pre>{#syntax#}@bitCast(comptime DestType: type, value: anytype) DestType{#endsyntax#}</pre>
       <p>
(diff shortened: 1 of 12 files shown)


zigazeljko:
Why the name `@blend`? Both LLVM and WebAssembly call this operation `select`, since it just *selects* elements from the two vectors. `@blend` makes it sound like it interpolates (aka *blends*) between elements, like the [`mix`](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/mix.xhtml) function in GLSL.

Snektron:
This is the name x86_64 uses. I picked blend initially because there is an existing proposal for async which would introduce a `select` keyword, and so didnt want to cause possible confusion between the two.

I dont have a strong opinion on the name though, im fine with renaming it to `@select`.

nektro:
https://gobyexample.com/select
https://tour.golang.org/concurrency/5

https://llvm.org/docs/LangRef.html#select-instruction
https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-instr-parametric

test/behavior/blend.zig
```diff
@@ -0,0 +1,25 @@
+const std = @import("std");
+const builtin = @import("builtin");
+const mem = std.mem;
+const expect = std.testing.expect;
+const Vector = std.meta.Vector;
+
+test "@shuffle" {
```
andrewrk, line 7: don't forget to rename this

andrewrk (changes requested):
I'd like to request the name to be `@select` despite the fact that we have another accepted proposal for select syntax (#5263).

Other than that, this looks ready to merge. Nice work and thank you.

andrewrk (approved):

PR #9463, opened 2021-07-26
stage1: add c_longdouble mapping for s390x

nektro:
make progress towards https://github.com/ziglang/zig-bootstrap/issues/52


nektro:
Reference: https://wiki.debian.org/ArchitectureSpecificsMemo

PR #9468, opened 2021-07-27
translate-c: handle NAN and INFINITY macros

ehaas:


ehaas:
I'm getting a `zig translate-c` error on Linux (and I assume this is what is happening on the other OS's) just from including math.h `math.c:1:10: 'math.h' file not found`. Works fine on MacOS though.

This happens even a file that is just `#include <math.h>`

Am I just missing something really obvious or is this a bug? @Vexu

ehaas:
I assume the `test-run-translated-c` script needs to be updated to include `-lc`?

Vexu:
Adding `#include <math.h>` to any other case in the file works for me using master branch, maybe rebasing would help?

ehaas:
Sorry about that, I think I misunderstood the failure - it seems like the actual issue is that non-macos is using `__builtin_isnan` and `__builtin_isinf` which are not implemented. I'll put those in and try again.

Issue #1994, opened 2019-02-22
overaligned pointer to packed struct always gives align(1) for fields


andrewrk:
This test fails:

```zig
const S = packed struct {
    a: u32,
    b: u32,
};
test "overaligned pointer to packed struct" {
    var foo: S align(4) = undefined;
    const ptr: *align(4) S = &foo;
    const ptr_to_b: *u32 = &ptr.b;
}
```

But Zig should know that if you have a pointer aligned to 4 bytes and then you add 4 bytes to it, it will still be 4 byte aligned.

Related: #1528

PR #9480, opened 2021-07-29
langref- fix packed struct error code

nektro:
The current docs say the error for this section is `docgen_tmp/test.zig:8:11: error: unused local constant`. However, they also mention that it should be failing due to https://github.com/ziglang/zig/issues/1994 instead. This change uses the variable so that the proper error can propagate.


Vexu (changes requested):
Could you add the expected error message using the `test_err|expected type...` syntax to both of your prs so that they don't regress again.

There also seem to be a few other places that are missing the error if you want to fix those too.

PR #9491, opened 2021-07-30
Make linux syscalls accessible with non-Linux target OS

N00byEdge:
It is of my opinion that explicitly doing `std.os.linux.exit(0);` in your code should do a Linux syscall, no matter what the current target may be (like freestanding). I hit the following issue when trying to make the above call in a freestanding executable for my own operating system (where I support a couple of Linux syscalls):

```
/mnt/gentoo/zig-linux-x86_64-0.9.0-dev.185+e125ead2b/lib/std/os/linux/x86_64.zig:16:25: error: use of undeclared identifier 'SYS'
pub fn syscall1(number: SYS, arg1: usize) usize {
                        ^
/mnt/gentoo/zig-linux-x86_64-0.9.0-dev.185+e125ead2b/lib/std/os/linux.zig:639:9: note: referenced here
    _ = syscall1(.exit, @bitCast(usize, @as(isize, status)));
        ^
```
For some reason, it seems to be grabbing the linux syscall numbers from `std.os.bits`, even though that makes no sense when I'm explicitly invoking a linux syscall.

This PR makes the linux syscalls always grab `bits/linux.zig`, no matter the target OS tag.

### DIFF
--- a/lib/std/os/bits/wasi.zig
+++ b/lib/std/os/bits/wasi.zig
@@ -4,6 +4,10 @@
 // The MIT license requires this copyright notice to be included in all copies
 // and substantial portions of the software.
 // Convenience types and consts used by std.os module
+const posix = @import("posix.zig");
+pub const iovec = posix.iovec;
+pub const iovec_const = posix.iovec_const;
+
 pub const STDIN_FILENO = 0;
 pub const STDOUT_FILENO = 1;
 pub const STDERR_FILENO = 2;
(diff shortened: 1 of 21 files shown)


lib/std/os/bits/wasi.zig
```diff
@@ -4,6 +4,8 @@
 // The MIT license requires this copyright notice to be included in all copies
 // and substantial portions of the software.
 // Convenience types and consts used by std.os module
+pub usingnamespace @import("posix.zig");
+
```
ifreund, line 8: I'm pretty sure wasi doesn't have `syslog`, why don't we export only `io_vec` and `iovec_const` as we do for windows?

ifreund (commented):
Looks good to me aside from one minor cleanup!

lib/std/os/bits/wasi.zig
N00byEdge: Fair enough!

ifreund (approved):
LGTM, Thanks!

PR #9496, opened 2021-07-31
stage2: wasm - Wrapping, intcast and optionals

Luukdegram:
- Implement wrapping operands
- Add support for integer widening through intcast instruction.
- Some wasm opcodes were incorrect or missing from wasm.zig, Those have now been added/updated.
- Support for optionals (See commit message for more details).


src/codegen/wasm.zig
```diff
@@ -800,8 +817,11 @@ pub const Context = struct {
         const air_tags = self.air.instructions.items(.tag);
         return switch (air_tags[inst]) {
             .add => self.airBinOp(inst, .add),
+            .addwrap => self.airBinOp(inst, .add),
```
joachimschmidt557, line 820: I don't think this will scale for all integers (e.g. other than `i32`). This program compiled with this branch will output a wrong number:

```zig
pub export fn _start() u8 {
    var i: u8 = 255;
    return i +% 1;
}
```

I think a different function for adding and then masking bits out is necessary.

joachimschmidt557 (commented):
This is nice progress for the stage2 compiler!

src/codegen/wasm.zig
Luukdegram: Good call, @joachimschmidt557. I hadn't even thought of this, will be updating this PR soon with the improvements.

src/codegen/wasm.zig
```diff
@@ -663,6 +663,23 @@ pub const Context = struct {
...
                 } };
             },
+            .Optional => {
+                var opt_buf: Type.Payload.ElemType = undefined;
+                const child_type = ty.optionalChild(&opt_buf);
+                if (ty.isPtrLikeOptional()) {
+                    return self.fail("TODO: wasm optional pointer", .{});
+                }
+
+                try self.locals.ensureCapacity(self.gpa, self.locals.items.len + 2);
```
andrewrk, line 673: This can be improved to ensureUnusedCapacity to avoid the + operation

src/codegen/wasm.zig
andrewrk: Note that for non wrapping operations, larger integers can he used. For example for non wrapping i13, you can just use i32 operations. It's the wrapping operations that need to possibly do more calculations to honor twos complement wraparound.

Luukdegram:
Thank you both for the review. Rebased and pushed the changes, would appreciate another review round.

src/codegen/wasm.zig
```diff
@@ -972,6 +1021,62 @@ pub const Context = struct {
...
+        if (bitsize != 32 and bitsize < 64) {
+            // first check if we can use a single instruction,
+            // wasm provides those if the integers are signed and 8/16-bit.
+            // For arbitrary integer sizes, we use the algorithm mentioned above.
+            if (is_signed and bitsize == 8) {
+                try self.code.append(wasm.opcode(.i32_extend8_s));
+            } else if (is_signed and bitsize == 16) {
+                try self.code.append(wasm.opcode(.i32_extend16_s));
+            } else {
+                const result = std.math.pow(u64, 2, bitsize - @boolToInt(is_signed)) - 1;
```
joachimschmidt557, line 1062: Nitpick: why was `std.math.pow` chosen instead of a bitshift?

Luukdegram: Wanted to avoid overflowing from bit shifting as we support integer sizes > 64. But I suppose it's safe to intcast to `u6` here as the above condition already verifies the bitsize is < 64. I believe bitshifting is faster, should I change it? @joachimschmidt557

joachimschmidt557: Jup, as bitshifting is faster I would suggest using bitshifts.

Luukdegram: Done, thanks again for the review :)

PR #9505, opened 2021-08-02
std: Declare FOLDERID_{Program,RoamingApp}Data for Windows

McSinyx:
The values are taken from [mingw-w64's knownfolders.h](https://github.com/ziglang/zig/blob/85755c51d529e7d9b406c6bdf69ce0a0f33f3353/lib/libc/include/any-windows-any/knownfolders.h#L73), and they are the system-wide and roaming counterpart of `FOLDERID_LocalAppData`: https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-vista/cc766489(v=ws.10)

Context: I'm writing [a tiny library](https://git.sr.ht/~cnx/zig-loca) to determine data, config, etc. directories across platforms.  Would it be a good idea to include it in Zig std?

### DIFF
--- a/lib/std/os/windows/bits.zig
+++ b/lib/std/os/windows/bits.zig
@@ -755,7 +755,9 @@ pub const GUID = extern struct {
     }
 };
 
+pub const FOLDERID_ProgramData = GUID.parse("{62AB5D82-FDC1-4DC3-A9DD-070D1D495D97}");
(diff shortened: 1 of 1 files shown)


Vexu:
I don't think that will be in the standard library as it is not needed for the compiler. Also you might wanna check out https://github.com/ziglibs/known-folders.

McSinyx:
Thanks for the std lib criteria clarification and the link!  What a relief :smile:

PR #9513, opened 2021-08-03
Linux/SPARCv9 thread and sigreturn-related fixes

koachan:
- Account for the branch delay in freeAndExit(). I forgot to do this in my first PR for it, my bad.
- Use C calling convention for the implementation of restore_rt(). This is needed to prevent an infinite loop when calling rt_sigreturn().

With those two changes, now almost all testcases in test-std passes (12 are still failing, mostly related to sockets and bigints).

### DIFF
--- a/lib/std/os/linux/sparc64.zig
+++ b/lib/std/os/linux/sparc64.zig
@@ -169,7 +169,9 @@ pub extern fn clone(func: fn (arg: usize) callconv(.C) u8, stack: usize, flags: 
 
 pub const restore = restore_rt;
 
-pub fn restore_rt() callconv(.Naked) void {
+// Need to use C ABI here instead of naked
+// to prevent an infinite loop when calling rt_sigreturn.
+pub fn restore_rt() callconv(.C) void {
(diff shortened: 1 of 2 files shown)


lib/std/os/linux/sparc64.zig
```diff
@@ -169,7 +169,9 @@ pub extern fn clone(func: fn (arg: usize) callconv(.C) u8, stack: usize, flags:
 
 pub const restore = restore_rt;
 
-pub fn restore_rt() callconv(.Naked) void {
+// Need to use C ABI here instead of naked
+// to prevent an infinite loop when calling rt_sigreturn.
+pub fn restore_rt() callconv(.C) void {
```
koachan, line 174: Honestly I'm not very sure that this is the right fix for it, though it does make the sigaction testcase pass instead of getting trapped in an infinite loop.

Probably needs a review from someone who knows better :eyes:

andrewrk: Well, if it's better than status quo, then it can be merged!

PR #9515, opened 2021-08-03
libstd: add ArrayHashMap.popOrNull function

kubkon:
which internally calls `ArrayHashMap.pop`, however, returns `?KV`
instead and performs the bounds checking automatically.

This function correponds to `ArrayList.popOrNull` and is meant
to fill the gap for situations where we want the quick lookup offered
by the hash map with elegant ability to iterate and pop of the
container with automatic bound checking that plugs in well with
a `while`-loop such as

```zig
var map = std.ArrayHashMap(K, V).init(allocator);
map.deinit();
while (map.popOrNull()) |entry| {
  // ... do something
}
assert(map.count() == 0);
```

### DIFF
--- a/lib/std/array_hash_map.zig
+++ b/lib/std/array_hash_map.zig
@@ -414,6 +414,12 @@ pub fn ArrayHashMap(
         pub fn pop(self: *Self) KV {
             return self.unmanaged.popContext(self.ctx);
         }
+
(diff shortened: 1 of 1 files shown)


andrewrk (approved):
Thanks for the test case :+1:

PR #9531, opened 2021-08-06
Make mem.split and mem.tokenize generic instead of assuming u8

squeek502:
- More in line with the rest of the `mem` functions, most of which take a type
- Allows `split`/`tokenize` to be used for things like UTF-16 strings


Vexu (approved):

haze:
The doc comment wasn't updated :(

squeek502:
I looked at the comments and decided there was nothing to update. I guess I'm blind :frowning_face:

PR #9532, opened 2021-08-06
Add minimal support to Zig toolchain to support building iOS binaries

kubkon:
With this change, it is possible to successfully target iOS and iPhone Simulator platforms. See [kubkon/zig-ios-example](https://github.com/kubkon/zig-ios-example) for a complete project example how.

Major changes include:
* ensure we correctly transfer `-iwithsysroot` and `-iframeworkwithsysroot` flags with values from `build.zig` and that they are correctly transferred forward to `zig cc`
* move `std.zig.system.macos` into `std.zig.system.darwin.macos` since we now special-case all Apple platforms when building from another Apple platform (usually macOS) - `std.zig.system.darwin.getSDKPath` now takes two args as input: allocator and `Target`, and based on the target it will invoke `xcrun --sdk <sdk_name> --show-sdk-path` with the appropriate `sdk_name` such as `iphonesimulator` or `iphoneos`, etc. With this change, we make Zig auto-invoke `getSDKPath` when targeting an Apple platform from an Apple platform, meaning that we no longer constrain ourselves to running natively but also when targeting `ios` from `macos`, etc.
* try to look for `libSystem.tbd` in the provided syslibroot
* take into account `.simulator` ABI when targeting iOS - `.simulator` ABI signals that the binary is meant to be run on Apple iPhone Simulator rather than on a real device and therefore it should be linked against a completely standalone set of libraries and frameworks known as `iphonesimulator` SDK
* swap out `VERSION_MIN` load command for `BUILD_VERSION` - turns out, the latter is required to make the app get properly launched on the iPhone Simulator, while presence of the former generates a sigkill-equivalent for some reason

I could use some input here on the validity of treating `std.build.addSystemIncludeDir` and `std.build.addFrameworkDir` differently based on the presence of the sysroot on the command line - if there is no sysroot present, we convert them into `-isystem` and `-iframework`; if the sysroot is present however, we change that to `-iwithsysroot` and `-iframeworkwithsysroot`. This can trigger a weird error if we are auto-detecting sysroots by Zig (when targeting iOS for instance) and the user adds explicit system search paths to their `build.zig` as then we get conflicting search dirs for the headers which can lead to a compile error.

---

UPDATE: turns out that when linking against simulator ABI such as `aarch64-ios-simulator`, several libc functions such as `fstat`, etc., are provided by the host (`aarch64-macos`), and thus, it is necessary to take that into account when linking against tbd stubs. This has now been added in 83f1203.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -2672,7 +2672,11 @@ pub const LibExeObjStep = struct {
                     try zig_args.append(self.builder.pathFromRoot(include_path));
                 },
                 .raw_path_system => |include_path| {
-                    try zig_args.append("-isystem");
+                    if (builder.sysroot != null) {
+                        try zig_args.append("-iwithsysroot");
+                    } else {
+                        try zig_args.append("-isystem");
+                    }
                     try zig_args.append(self.builder.pathFromRoot(include_path));
                 },
                 .other_step => |other| if (other.emit_h) {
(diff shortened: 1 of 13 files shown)


jedisct1:
\o/

PR #9533, opened 2021-08-06
include builtin & std packages in all defined packages

### LINKED ISSUE #9204: self-hosted compiler does not map 'std' into every package
#### test.zig

```zig
const bar = @import("bar");
pub fn main() void {
    _ = bar;
}
```

#### bar.zig

```zig
const std = @import("std");

comptime {
    _ = std;
}
```

#### Actual Output

```
$ ./zig-out/bin/zig build-exe test.zig --pkg-begin bar bar.zig --pkg-end
bar.zig:1:21: error: unable to open 'std': PackageNotFound
const std = @import("std");
                    ^
```

#### Expected Output

It should succeed the build. 'builtin' and 'root' should additionally be mapped into every package.

dimenus:
Closes #9204

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1227,6 +1227,12 @@ pub fn create(gpa: *Allocator, options: InitOptions) !*Compilation {
             } else main_pkg;
             errdefer if (options.is_test) root_pkg.destroy(gpa);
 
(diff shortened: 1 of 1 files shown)


andrewrk:
Thanks!

PR #9538, opened 2021-08-08
Ip4Address parser: reject 0-prefixed components

jedisct1:
Octal is hardly useful to anything besides file permissions, but aging standards still use it.

127.0.026.1 can be parsed as 127.0.22.1, not 127.0.26.1.

Some parsers support octal, some don't, and the confusion can lead to vulnerabilities.

Reject 0-prefixed IPv4 components. This align to what the Go language did to fix CVE-2021-29923.

### DIFF
--- a/lib/std/net.zig
+++ b/lib/std/net.zig
@@ -34,6 +34,7 @@ pub const Address = extern union {
             error.InvalidEnd,
             error.InvalidCharacter,
             error.Incomplete,
+            error.NonCanonical,
             => {},
         }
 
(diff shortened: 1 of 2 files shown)


jedisct1:
Good watch for reference:

https://www.youtube.com/watch?v=_o1RPJAe4kU

jedisct1:
Thinking about it, we should probably just reject octal numbers in IP addresses. No one would use that except to trigger a vulnerability.

We should probably do what the Go people did, and reject `0`-prefixed components instead.

PR #9542, opened 2021-08-09
macho: fix linking of dylibs and frameworks

kubkon:
Previously, I have incorrectly assumed that with two-level namespace
we only need to link in dylibs/frameworks that actually export symbols
which are undefined in the linked image. Turns out, regardless of
whether we link with two-level namespace (default on macOS) or a
flat namespace (more common on other platforms), we always need to
put the dylibs/frameworks as specified by the user from the linker
line into the final linked image.

cc @floooh this should fix the issue you reported on twitter.


floooh:
Here's the twitter thread for reference: https://twitter.com/FlohOfWoe/status/1424419698628833284

Also: the fix is working :)

PR #9543, opened 2021-08-09
std.time.Clock: unified time abstraction

kprotty:
There are some slight inconsistencies with the current timer APIs in use:
* `std.time.Timer` uses `QueryPerformanceCounter()` on windows which counts time when suspended while the other OS functions used don't.
* `std.time` doesn't expose a cross-platform way to get which includes/exclusives when system is suspended.
* Documentation implies the Timer source is monotonically increasing but it can sometimes [go backwards](https://doc.rust-lang.org/src/std/time.rs.html#223-247)

This PR abstracts all the timer logic out into its own struct `std.time.Clock` with clock_ids modeled after POSIX [`clock_gettime`](https://pubs.opengroup.org/onlinepubs/9699919799/) ~~with variations of the clock_ids optimized for either speed or precision inspired by [freebsd/dragonfly](https://www.freebsd.org/cgi/man.cgi?query=clock_gettime)~~ (omitted due not added complexity for windows & not really useful broadly).

The Clock abstraction removes the need to do if checks in each `std.time` function, exposes the extra clock sources which account for system suspend, and keeps a "familiar" API for all clock sources not just monotonic and realtime. I marked this PR as a draft as there's still some docs to change and it needs more feedback on whether this is even a decent abstraction.

### DIFF
--- a/lib/std/c/darwin.zig
+++ b/lib/std/c/darwin.zig
@@ -1,7 +1,7 @@
 const std = @import("../std.zig");
 const assert = std.debug.assert;
 const macho = std.macho;
-const native_arch = builtin.target.cpu.arch;
+const native_arch = std.Target.current.cpu.arch;
 const maxInt = std.math.maxInt;
 const iovec_const = std.os.iovec_const;
 
--- a/lib/std/c/darwin.zig
+++ b/lib/std/c/darwin.zig
@@ -1,6 +1,5 @@
 const std = @import("../std.zig");
 const assert = std.debug.assert;
-const builtin = @import("builtin");
 const macho = std.macho;
 const native_arch = builtin.target.cpu.arch;
 const maxInt = std.math.maxInt;
@@ -44,7 +43,8 @@ pub const fstat = if (native_arch == .aarch64) private.fstat else private.@"fsta
 pub const fstatat = if (native_arch == .aarch64) private.fstatat else private.@"fstatat$INODE64";
 
 pub extern "c" fn mach_absolute_time() u64;
-pub extern "c" fn mach_timebase_info(tinfo: ?*mach_timebase_info_data) void;
+pub extern "c" fn mach_continuous_time() u64;
+pub extern "c" fn mach_timebase_info(tinfo: ?*mach_timebase_info_data) kern_return_t;
 
 pub extern "c" fn malloc_size(?*const c_void) usize;
 pub extern "c" fn posix_memalign(memptr: *?*c_void, alignment: usize, size: usize) c_int;
@@ -206,6 +206,7 @@ pub extern "c" fn dispatch_semaphore_signal(dsema: dispatch_semaphore_t) isize;
 pub const dispatch_time_t = u64;
 pub const DISPATCH_TIME_NOW = @as(dispatch_time_t, 0);
 pub const DISPATCH_TIME_FOREVER = ~@as(dispatch_time_t, 0);
+pub const DISPATCH_MONOTONICTIME_NOW = @as(dispatch_time_t, 1 << 63); // available(macos=10.14, ios=12.0, tvos=12.0, watchos=5.00)
 pub extern "c" fn dispatch_time(when: dispatch_time_t, delta: i64) dispatch_time_t;
 
 const dispatch_once_t = usize;
--- a/lib/std/c/darwin.zig
+++ b/lib/std/c/darwin.zig
@@ -1,7 +1,7 @@
 const std = @import("../std.zig");
 const assert = std.debug.assert;
 const macho = std.macho;
-const native_arch = std.Target.current.cpu.arch;
+const native_arch = std.builtin.target.cpu.arch;
 const maxInt = std.math.maxInt;
 const iovec_const = std.os.iovec_const;
 
--- a/lib/std/time.zig
+++ b/lib/std/time.zig
@@ -4,7 +4,7 @@ const assert = std.debug.assert;
 const testing = std.testing;
 const os = std.os;
 const math = std.math;
-const target = std.Target.current;
+const target = builtin.target;
 
 pub const epoch = @import("time/epoch.zig");
 
@@ -169,8 +169,8 @@ pub const Timer = struct {
     }
 
     /// Returns the current value of the Timer's clock.
-    /// Should not error out since the existence of the clock was checked at start().
     fn clockNative() i128 {
+        // Must not fail since existence of the clock was checked at start().
         return Clock.read(clock_id) catch unreachable;
     }
 
@@ -265,7 +265,7 @@ pub const Clock = struct {
                     // At some point we may change our minds on CLOCK.MONOTONIC_RAW,
                     // but for now we're sticking with CLOCK.MONOTONIC standard.
                     // For more information, see: https://github.com/ziglang/zig/pull/933
-                    .linux => os.CLOCK.MONOTONIC, // doesn't count time suspended
+                    .linux => os.CLOCK.MONOTONIC,
                     // Platforms like wasi and netbsd don't support getting time without suspend
                     else => null,
                 },
--- a/lib/std/time.zig
+++ b/lib/std/time.zig
@@ -249,28 +249,28 @@ pub const Clock = struct {
         // https://www.freebsd.org/cgi/man.cgi?query=clock_gettime
         fn toOsClockId(id: Id) ?i32 {
             return switch (id) {
-                .realtime => os.CLOCK_REALTIME,
+                .realtime => os.CLOCK.REALTIME,
                 .monotonic => switch (target.os.tag) {
                     // Calls mach_continuous_time() internally
-                    .macos, .tvos, .ios, .watchos => os.CLOCK_MONOTONIC_RAW,
-                    // Unlike CLOCK_MONOTONIC, this actually counts time suspended (true POSIX MONOTONIC)
-                    .linux => os.CLOCK_BOOTTIME,
-                    else => os.CLOCK_MONOTONIC,
+                    .macos, .tvos, .ios, .watchos => os.CLOCK.MONOTONIC_RAW,
+                    // Unlike CLOCK.MONOTONIC, this actually counts time suspended (true POSIX MONOTONIC)
+                    .linux => os.CLOCK.BOOTTIME,
+                    else => os.CLOCK.MONOTONIC,
                 },
                 .uptime => switch (target.os.tag) {
-                    .openbsd, .freebsd, .kfreebsd, .dragonfly => os.CLOCK_UPTIME,
+                    .openbsd, .freebsd, .kfreebsd, .dragonfly => os.CLOCK.UPTIME,
                     // Calls mach_absolute_time() internally
-                    .macos, .tvos, .ios, .watchos => os.CLOCK_UPTIME_RAW,
-                    // CLOCK_MONOTONIC on linux actually doesn't count time suspended (not POSIX compliant).
-                    // At some point we may change our minds on CLOCK_MONOTONIC_RAW,
-                    // but for now we're sticking with CLOCK_MONOTONIC standard.
+                    .macos, .tvos, .ios, .watchos => os.CLOCK.UPTIME_RAW,
+                    // CLOCK.MONOTONIC on linux actually doesn't count time suspended (not POSIX compliant).
+                    // At some point we may change our minds on CLOCK.MONOTONIC_RAW,
+                    // but for now we're sticking with CLOCK.MONOTONIC standard.
                     // For more information, see: https://github.com/ziglang/zig/pull/933
-                    .linux => os.CLOCK_MONOTONIC, // doesn't count time suspended
+                    .linux => os.CLOCK.MONOTONIC, // doesn't count time suspended
                     // Platforms like wasi and netbsd don't support getting time without suspend
                     else => null,
                 },
-                .thread_cputime => os.CLOCK_THREAD_CPUTIME_ID,
-                .process_cputime => os.CLOCK_PROCESS_CPUTIME_ID,
+                .thread_cputime => os.CLOCK.THREAD_CPUTIME_ID,
+                .process_cputime => os.CLOCK.PROCESS_CPUTIME_ID,
             };
         }
     };
--- a/lib/std/time.zig
+++ b/lib/std/time.zig
@@ -260,7 +260,7 @@ pub const Clock = struct {
                 .monotonic => getInterruptTime(),
                 .uptime => getUnbiasedInterruptTime(),
                 .thread_cputime => getCpuTime("GetThreadTimes", "GetCurrentThread"),
-                .thread_cputime => getCpuTime("GetProcessTimes", "GetCurrentProcess"),
+                .process_cputime => getCpuTime("GetProcessTimes", "GetCurrentProcess"),
             };
         }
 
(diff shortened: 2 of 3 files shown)


jayschwa:
The "fast" vs "precise" distinction is interesting, but it seems like it could still require the user to understand "how fast" and "how precise" the clocks are on various platforms. Personally, I'd be interested to see an abstraction like, "give me the fastest realtime/monotonic/uptime clock where resolution <= X", where X is a parameter.

kprotty:
@jayschwa Due to OS apis and the variability in scheduling, there's no definition of how fast or precise AFAICK:

* Speed isn't something that can be relatively measured (at least accurately) if the clock reads have to do some extra work only sometimes (e.g. NTP sync on linux or adjusting with hardware counter for Precise timing on windows)
* Precision can also change in the clock sources either due to user intervention (i.e. [SetSystemTimeAdjustment](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-setsystemtimeadjustment)) or other OS scheduling shenanigans (i.e. realtime/high-priority threads having potentially less skew)

Therefor I noted the `_fast` and `_precise` clock sources as mostly just hints: `"but tries to favor query X over Y"`. This allows the fast and precise clock sources to technically coerce into the original clock when the hints aren't available on some platforms (see Impl for Netbsd, Openbsd, and Wasi) without breaking correctness. I also need to make the monotonic property a hint as well.

andrewrk (approved):
I support this change. Would you mind rebasing/fixing conflicts and doing the merge?

lib/std/time.zig
```diff
@@ -50,51 +52,45 @@ pub fn sleep(nanoseconds: u64) void {
...
-    if (builtin.os.tag == .wasi and !builtin.link_libc) {
-        var ns: os.wasi.timestamp_t = undefined;
-        const err = os.wasi.clock_time_get(os.wasi.CLOCK_REALTIME, 1, &ns);
-        assert(err == .SUCCESS);
-        return ns;
-    }
-    var ts: os.timespec = undefined;
-    os.clock_gettime(os.CLOCK_REALTIME, &ts) catch |err| switch (err) {
-        error.UnsupportedClock, error.Unexpected => return 0, // "Precision of timing depends on hardware and OS".
+pub fn nanoTimestamp() u64 {
```
andrewrk, line 77: Why 64 bits instead of 128 bits?

lib/std/time.zig
```diff
@@ -287,3 +185,250 @@ test "Timer" {
...
+    for (clock_ids) |clock_id| {
+        _ = Clock.read(clock_id) catch |err| {
+            // Only return errors for clock sources that we use personally.
+            // The rest are just tested for their code paths
+            switch (clock_id) {
+                .realtime, .monotonic => return err,
+                else => continue,
+            }
+        };
+    }
```
andrewrk, line 433: this logic seems unnecessarily complicated. can it just have the tests instead of a bunch of meta control flow?

andrewrk (changes requested):
I have some questions on the API surface-

lib/std/time.zig
```diff
@@ -50,51 +52,45 @@ pub fn sleep(nanoseconds: u64) void {
...
+    sleep(1);
+}
+
 /// Get a calendar timestamp, in seconds, relative to UTC 1970-01-01.
 /// Precision of timing depends on the hardware and operating system.
-/// The return value is signed because it is possible to have a date that is
-/// before the epoch.
 /// See `std.os.clock_gettime` for a POSIX timestamp.
-pub fn timestamp() i64 {
+pub fn timestamp() u64 {
```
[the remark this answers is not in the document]
kprotty: Originally, the code looked like it could only go signed for windows which has a farther back epoch. Doing a bit more research though, it can go before the epoch (?) and is normally kept signed even for POSIX platforms. So i'll change it back.

* https://github.com/abseil/abseil-cpp/blob/1ae9b71c474628d60eb251a3f62967fe64151bb2/absl/time/internal/get_current_time_posix.inc#L13
* https://github.com/golang/go/blob/master/src/time/time.go#L128-L140
* https://doc.rust-lang.org/src/std/sys/unix/time.rs.html#12-14

lib/std/time.zig
kprotty: Was for consistency with Timer. Thinking about it again, `Timer.read()` returns elapsed time & program wouldn't be running for 500+ years to overflow so they shouldn't be compared.

lib/std/time.zig
kprotty: Using `inline for` on the fields itself made it a compile error to do `continue` for specific errors as it treated it as "runtime-based compile-time control-flow"

andrewrk (changes requested):
I have some merge blockers, but overall it seems the direction this branch goes is good. With a couple things addressed, it can be merged.

lib/std/c/darwin.zig
```diff
@@ -1,8 +1,7 @@
 const std = @import("../std.zig");
 const assert = std.debug.assert;
-const builtin = @import("builtin");
 const macho = std.macho;
-const native_arch = builtin.target.cpu.arch;
+const native_arch = std.Target.current.cpu.arch;
```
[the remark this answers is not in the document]
kprotty: TIL. I've been using `std.Target.current` for the longest time since it looked like the "proper" way.

lib/std/time.zig
```diff
@@ -127,85 +129,31 @@ pub const s_per_hour = s_per_min * 60;
...
 
-/// A monotonic high-performance timer.
-/// Timer.start() must be called to initialize the struct, which captures
-/// the counter frequency on windows and darwin, records the resolution,
-/// and gives the user an opportunity to check for the existnece of
-/// monotonic clocks without forcing them to check for error on each read.
-/// .resolution is in nanoseconds on all platforms but .start_time's meaning
-/// depends on the OS. On Windows and Darwin it is a hardware counter
-/// value that requires calculation to convert to a meaninful unit.
+/// A high-performance timer that tries to be monotonically increasing.
```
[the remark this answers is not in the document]
kprotty: Would require a bit of synchronization overhead for single-threaded use cases. The current alternative is to [enforce it locally](https://github.com/coilhq/tigerbeetle/blob/main/src/time.zig) instead of globally. The overhead might not make much of a difference given we're already converting timestamps to nanoseconds (mul + div, possibly "expensive") for each read()?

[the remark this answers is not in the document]
andrewrk: Ah I see your point. Well we already have the concept of creating an instance with the start() function. That instance could store the value to enforce it locally. I suggest we do that, and keep the API simpler. If `nanoTimeStamp` has to be deleted, that's fine with me.

lib/std/time.zig
```diff
@@ -285,3 +201,194 @@ test "Timer" {
     timer.reset();
     try testing.expect(timer.read() < time_1);
 }
+
+pub const Clock = struct {
+    /// Id represents the type of clock source to interact with.
+    /// Each clock sources may start, tick, and change in different ways.
+    pub const Id = enum {
+        /// Returns nanoseconds since Unix Epoch (Jan 1 1970)
```
PhaseMage, line 209: I'd like to add a bit more to the comment here to help folks understand the behavior of the 'realtime' timestamp. I'm thinking:
```
/// Returns nanoseconds since local Unix Epoch (Jan 1 1970).
/// Forward and backward jumps happen non-deterministically when
/// the system time changes (manually, from NTP, leap seconds, or 
/// others). Except for displaying the current time to a user, building any 
/// logic around this timestamp will probably lead to bugs.
/// NOTE: Comparing values across systems is undefined due to 
/// unknown number of Leap Seconds within each timestamp.
/// NOTE: On Windows, the returned value will be offset by the number of
/// Leap Seconds that were declared and then occurred since this std library
/// was compiled into your program.
///
/// TODO: Implement 'Id.atomic' to produce a timestamp which
/// doesn't have the undefined behavior of Leap Seconds. (TAI or GPS Time).
/// 'Id.atomic' would still suffer from NTP or manual time discontinuities.
```
I'd like to try my hand at implementing Id.atomic if folks like the idea.

lib/std/time.zig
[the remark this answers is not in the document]
PhaseMage: @kprotty is correct, `nanoTimeStamp` can't be monotonic, since the system time can change forward and backward for a number of reasons.

With all the issues in `nanoTimeStamp`, I'd prefer it go away entirely, so that it doesn't lead people down a path of bugs.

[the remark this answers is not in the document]
andrewrk: > For Timer to enforce monotonic values, read() would have to take in `*Timer` instead of `Timer` (to update the "last timestamp"). Is this fines?

Yes


> Also, to be clear, `nanoTimeStamp` doesn't have to be monotonic. It returns the system time so it has an unfortunate name which makes people use it for measuring elapsed time. Rename it to `systemTimestamp` or nah?

Oops, my mistake. I have another name counter-proposal: `calendarTimestamp`

lib/std/time.zig
kprotty: few notes on the doc comment:

> Comparing values across systems is undefined due to 

Would it be better phrased as "unspecified"? "Undefined" implies that if the program uses the cross-system time comparison, then the entire program becomes unsound / could do anything. The former just means the implementation can decide but the program is still sound.

> Implement 'Id.atomic' to produce a timestamp which doesn't have the undefined behavior of Leap Seconds. 

Again, I don't think leap seconds are undefined behavior. For some, it seems to be undesired info. After some minor research, it looks like linux (or POSIX ?) reports the unix timestamp *without* leap seconds. Also TAI seems to be only available on linux and i'm not sure about the availability of GPS for most of the platforms supported by the stdlib currently. If you're up to try and implement it, I would be interested in the results. Just missing context on this all.

PhaseMage: I'm happy with the word "unspecified".

Regarding leap seconds, I believe that POSIX just doesn't worry about the concept of leap seconds. It pretends that UTC is a timescale that doesn't have leap seconds. This philosophy kinda works, but it's not accurate.
What happens in practice after a leap second is that NTP is used (some time after the leap second occurs) to get an updated time, and the "current time" is updated (a second is added or subtracted). Local Timestamps that occur between when the leap second occured and when the update arrives (usually less than 5min), are off by 1 second. Time deltas measured across the update have discontinuities.

Windows, on the other hand, defines an epoch that explicitly does not have leap seconds (and has no discontinuities when leap seconds occur), but then zig uses a static variable to convert it to the Unix Epoch in UTC. But this (static) conversion will be wrong by 1 second after the next leap second.

So I'll try over the weekend to see if I can put a solution to this into code with Id.atomic.

lib/std/time.zig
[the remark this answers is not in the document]
kprotty: For `calendarTimestamp`, we would lose the other millis/second resolutions unless we had like `calendarMillisTimestamp`. Also, "calendar" feels off given my recent understanding about leap seconds, unless the time returned is time-zone specific. I've been talking with others on the discord and came up with counter-counter-proposal (also API wise):

```zig
/// Get measurements of time relative to the unix epoch (jan 1 1970)
/// Precision depends on OS.
/// For capturing date time. Use Instant or Timer for capturing elapsed time.
pub const UnixTime = struct {
    pub fn nanoTimestamp() i128
        // windows: GetSystemTimePreciseAsFileTime()
        // darwin: gettimeofday (clock_gettime calls this internally)
        // posix: clock_gettime(CLOCK_REALTIME)
    pub fn timestamp() i64 // seconds version
    pub fn milliTimestamp() i64 // millis version
};

/// Measures amount of time the callers thread or process spent running on cpu
pub const CpuTime = enum {
    thread,
    process,
    
    // measure thread/process cpu time in nanoseconds 
    pub fn read(self: CpuTime) error{Unavailable}!u64
        // windows: GetThreadTimes/GetProcessTimes
        // posix: clock_gettime(CLOCK_[THREAD|PROCESS]_CPUTIME_ID)
};

/// A snapshot of the current high-precision clock value.
/// Mostly monotonic, but is allowed to jump backwards.
/// Can be used to measure elapsed (or wallclock) time. 
pub const Instant = struct {
    timestamp: if (is_darwin or is_windows) u64 else os.timespec,
    
    // Get the current Instant snapshot.
    // Time conversion to nanoseconds happens at since() to keep now() sampling fast.
    pub fn now() error{Unavailable}!Instant
        // windows: QueryPerformanceCounter
        // darwin: mach_continuous_time
        // linux: clock_gettime(CLOCK_BOOTTIME)
        // posix: clock_gettime(CLOCK_MONOTONIC)

    // For fast comparisons without computing time since() below
    pub fn order(self: Instant, other: Instant) std.math.Order 
    
    // Returns the amount time in nanoseconds since the earlier Instant to this Instant.
    // If this Instant snapshot represents a time before earlier Instant, a negative value is returned.
    // If too much time occurred between this Instant and earlier Instant, `error.Overflow` is returned.
    pub fn since(self: Instant, earlier: Instant) error{Overflow}!i64
        // windows: QueryPerformanceFrequency / specialize on 10mhz / safeMulDiv
        // darwin: mach_timebase_info / specialize on numer==denom / safeMulDiv
        // posix: timespecsub()
};

// Same thing as before, but backed by Instant and is ensured to be monotonic
pub const Timer = struct {
    start_time: Instant,
    last_time: Instant,

    pub fn start() error{TimerUnsupported}!Timer:
        const s = Instant.now() catch return error.TimerUnsupported;
        return Timer{ .start_time = s, .last_time = s };
        
    pub fn reset(self: *Timer) void:
        self.start_time = self.getTimestamp()
        
    pub fn read(self: *Timer) u64:
        return self.toDuration(self.getTimestamp())
    
    pub fn lap(self: *Timer) u64:
        const current_time = self.getTimestamp();
        defer self.start_time = current_time;
        returns self.toDuration(current_time);
    
    // reads timestamp assuming that it cannot fail
    // makes sure the timestamp is monotonic
    fn getTimestamp() Instant {
        const current = Instant.now() catch unreachable;
        if (self.last_time.order(current) == .lt) self.last_time = current;
        return self.last_time 

    fn toDuration(current: Instant) u64 {
        assert(self.start_time.order(current) != .lt);
        return @intCast(u64, current.since(self.start_time) catch unreachable)
};
```

lib/std/time.zig
PhaseMage: I didn't understand the complexity as much as I thought I did. I spent about 10 hours researching the Linux and Windows implementations of leap seconds and writing up some notes toward a plan, but it's going to take another weekend to have something concrete to share. Fortunately we have some time, as it looks like the next leap second is at least a year away. They might even be eliminated, so I'm planning to defer some complexity of implementation until a confirmed leap second is announced.

PhaseMage: I spent l ton more time this weekend. So the best way I found to describe the issue is that UTC timestamps from a Unix Epoch are inherently "Fuzzy" in that it's unclear how many leap seconds are in a given timestamp. This is not a problem for date/time strings. We're currently in Utc2018, Windows FILETIME will stay in Utc2018 forever as of 2018 (even after future leap seconds). But timestamps in the past need to be adjusted to be accurate in the current Epoch.
I figured out how to test if a POSIX system supports CLOCK_TAI and build a consistent Utc2018 timestamp from it. Otherwise, we have to assume CLOCK_REALTIME is in Utc2018 until a leap second is declared (which might not happen).

I've created a CalendarTime struct with Gregorian year, month, day, etc.
I've got Clock.id.utc_2018 implemented that exposes a coherent timestamp that ranges from year 1601 - 3800 or so. Timestamps from before the Gregorian calendar are silly. I have conversion functions between these timestamps and the CalendarTime struct.
I have a table of leap seconds and functions to convert between fuzzyUtc and utc_2018.
Next step is to build a function that can initialize a CalendarTime struct from a string to make testing concise. Then I need to write more tests.

andrewrk:
Looks like a lot of good work in this PR but never quite made it to a mergeable state. Looking forward to a second attempt, if either of you should decide to bring it home.

PR #9545, opened 2021-08-10
Fix darwin ulock usage on macOS

thomcc:
I was doing macOS ulock stuff, and was linked to Zig's DarwinFutex implementation by @kprotty. There were a few bugs I noticed:

- `__ulock_wait2` actually timeout as a `u64`, not `u32` (this likely happens to work on at least x86_64 because of how the ABI works, but I can't say about others, and it's still UB either way).
- Semantically, `__ulock_wait2`'s timeout is in units of nanoseconds (this would just lead to waking up early which callers have to handle, but still).
- EFAULT is not handled, but seems like it should be? At least [almost](https://github.com/apple/darwin-libpthread/blob/2b46cbcc56ba33791296cd9714b2c90dae185ec7/src/pthread_cancelable.c#L290) [all](https://github.com/apple/darwin-libpthread/blob/2b46cbcc56ba33791296cd9714b2c90dae185ec7/src/pthread_dependency.c#L83) [uses](https://github.com/apple/darwin-libplatform/blob/215b09856ab5765b7462a91be7076183076600df/src/os/lock.c#L541) [in](https://github.com/apple/darwin-libpthread/blob/2b46cbcc56ba33791296cd9714b2c90dae185ec7/src/pthread_cond.c#L717) [system](https://github.com/apple/darwin-libpthread/blob/2b46cbcc56ba33791296cd9714b2c90dae185ec7/src/pthread_mutex.c#L1087) [libraries](https://github.com/apple/darwin-libplatform/blob/215b09856ab5765b7462a91be7076183076600df/src/os/lock.c#1165) [/ tests](https://github.com/apple/darwin-xnu/blob/2ff845c2e033bd0ff64b5b6aa6063a1f8f65aa32/tests/ulock.c#L39) have it handled.

r? @kprotty

### DIFF
--- a/lib/std/Thread/Futex.zig
+++ b/lib/std/Thread/Futex.zig
@@ -186,17 +186,24 @@ const DarwinFutex = struct {
         // This XNU version appears to correspond to 11.0.1:
         // https://kernelshaman.blogspot.com/2021/01/building-xnu-for-macos-big-sur-1101.html
         //
-        // ulock_wait() uses 32-bit micro-second timeouts, where 0 = INIFITE or no-timeout
+        // ulock_wait() uses 32-bit micro-second timeouts where 0 = INFINITE or no-timeout
         // ulock_wait2() uses 64-bit nano-second timeouts (with the same convention)
         const timeout_ns: u64 = timeout orelse 0;
         const addr = @ptrCast(*const c_void, ptr);
         const flags = darwin.UL_COMPARE_AND_WAIT | darwin.ULF_NO_ERRNO;
+        // If we're using `__ulock_wait` and `timeout` is too big to fit inside a `u32` count of
+        // micro-seconds (around 70min), we'll request a shorter timeout. This is fine (users
+        // should handle spurious wakeups), but we need to remember that we did so, so that
+        // we don't return `TimedOut` incorrectly.
+        var timeout_overflowed = false;
         const status = blk: {
             if (target.os.version_range.semver.max.major >= 11) {
                 break :blk darwin.__ulock_wait2(flags, addr, expect, timeout_ns, 0);
             } else {
-                const timeout_us = @intCast(u32, timeout_ns / std.time.ns_per_us);
-                break :blk darwin.__ulock_wait(flags, addr, expect, timeout_us);
+                const timeout_us = timeout_ns / std.time.ns_per_us;
+                timeout_overflowed = timeout_us > std.math.maxInt(u32);
+                const timeout_us32 = if (timeout_overflowed) std.math.maxInt(u32) else @intCast(u32, timeout_us);
+                break :blk darwin.__ulock_wait(flags, addr, expect, timeout_us32);
             }
         };
 
@@ -207,7 +214,7 @@ const DarwinFutex = struct {
             // pthread/libdispatch on darwin bother to handle it. In this case we'll return
             // without waiting, but the caller should retry anyway.
             darwin.EFAULT => {},
-            darwin.ETIMEDOUT => return error.TimedOut,
+            darwin.ETIMEDOUT => if (!timeout_overflowed) return error.TimedOut,
             else => unreachable,
         }
     }
--- a/lib/std/Thread/Futex.zig
+++ b/lib/std/Thread/Futex.zig
@@ -188,22 +188,29 @@ const DarwinFutex = struct {
         //
         // ulock_wait() uses 32-bit micro-second timeouts where 0 = INFINITE or no-timeout
         // ulock_wait2() uses 64-bit nano-second timeouts (with the same convention)
-        const timeout_ns: u64 = timeout orelse 0;
+        var timeout_ns: u64 = 0;
+        if (timeout) |timeout_value| {
+            // This should be checked by the caller.
+            assert(timeout_value != 0);
+            timeout_ns = timeout_value;
+        }
         const addr = @ptrCast(*const c_void, ptr);
         const flags = darwin.UL_COMPARE_AND_WAIT | darwin.ULF_NO_ERRNO;
         // If we're using `__ulock_wait` and `timeout` is too big to fit inside a `u32` count of
         // micro-seconds (around 70min), we'll request a shorter timeout. This is fine (users
         // should handle spurious wakeups), but we need to remember that we did so, so that
-        // we don't return `TimedOut` incorrectly.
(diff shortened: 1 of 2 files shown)


lib/std/Thread/Futex.zig
```diff
@@ -180,31 +180,33 @@ const DarwinFutex = struct {
...
-            timeout_us = @intCast(u32, timeout_ns / std.time.ns_per_us);
-        }
-
         // Darwin XNU 7195.50.7.100.1 introduced __ulock_wait2 and migrated code paths (notably pthread_cond_t) towards it:
         // https://github.com/apple/darwin-xnu/commit/d4061fb0260b3ed486147341b72468f836ed6c8f#diff-08f993cc40af475663274687b7c326cc6c3031e0db3ac8de7b24624610616be6
         //
         // This XNU version appears to correspond to 11.0.1:
         // https://kernelshaman.blogspot.com/2021/01/building-xnu-for-macos-big-sur-1101.html
+        //
+        // ulock_wait() uses 32-bit micro-second timeouts, where 0 = INIFITE or no-timeout
```
daurnimator, line 189: typo: INIFITE

lib/std/Thread/Futex.zig
```diff
@@ -180,31 +180,33 @@ const DarwinFutex = struct {
...
+        // ulock_wait2() uses 64-bit nano-second timeouts (with the same convention)
+        const timeout_ns: u64 = timeout orelse 0;
```
daurnimator, line 191: do you need to check for `0` first (and increment it to e.g. 1? or return early?)

lib/std/Thread/Futex.zig
```diff
@@ -180,31 +180,33 @@ const DarwinFutex = struct {
...
         const addr = @ptrCast(*const c_void, ptr);
         const flags = darwin.UL_COMPARE_AND_WAIT | darwin.ULF_NO_ERRNO;
         const status = blk: {
             if (target.os.version_range.semver.max.major >= 11) {
-                break :blk darwin.__ulock_wait2(flags, addr, expect, timeout_us, 0);
+                break :blk darwin.__ulock_wait2(flags, addr, expect, timeout_ns, 0);
             } else {
+                const timeout_us = @intCast(u32, timeout_ns / std.time.ns_per_us);
```
daurnimator, line 198: What if it doesn't fit? (I see this was already potentially an issue)

lib/std/Thread/Futex.zig
thomcc: I think this function is called by one that checks 0 and returns in that case, so there should be no need to handle it, no?

lib/std/Thread/Futex.zig
thomcc: 😅

lib/std/Thread/Futex.zig
daurnimator: In that case we should `assert(timeout != 0)` to make sure no one does it by mistake (also potentially informing the optimiser)

lib/std/Thread/Futex.zig
```diff
@@ -180,32 +180,41 @@ const DarwinFutex = struct {
...
+        var timeout_overflowed = false;
         const status = blk: {
             if (target.os.version_range.semver.max.major >= 11) {
-                break :blk darwin.__ulock_wait2(flags, addr, expect, timeout_us, 0);
+                break :blk darwin.__ulock_wait2(flags, addr, expect, timeout_ns, 0);
             } else {
-                break :blk darwin.__ulock_wait(flags, addr, expect, timeout_us);
+                const timeout_us = timeout_ns / std.time.ns_per_us;
+                timeout_overflowed = timeout_us > std.math.maxInt(u32);
+                const timeout_us32 = if (timeout_overflowed) std.math.maxInt(u32) else @intCast(u32, timeout_us);
```
daurnimator, line 205: `math.clamp`? https://github.com/ziglang/zig/blob/0686954802d17e87114542878615fd0a9a245e49/lib/std/math.zig#L420

lib/std/Thread/Futex.zig
```diff
@@ -180,32 +180,41 @@ const DarwinFutex = struct {
...
+        // ulock_wait() uses 32-bit micro-second timeouts where 0 = INFINITE or no-timeout
+        // ulock_wait2() uses 64-bit nano-second timeouts (with the same convention)
+        const timeout_ns: u64 = timeout orelse 0;
         const addr = @ptrCast(*const c_void, ptr);
         const flags = darwin.UL_COMPARE_AND_WAIT | darwin.ULF_NO_ERRNO;
+        // If we're using `__ulock_wait` and `timeout` is too big to fit inside a `u32` count of
+        // micro-seconds (around 70min), we'll request a shorter timeout. This is fine (users
+        // should handle spurious wakeups), but we need to remember that we did so, so that
+        // we don't return `TimedOut` incorrectly.
+        var timeout_overflowed = false;
```
daurnimator, line 198: This doesn't seem to be used outside of the block below: put the variable inside of the `else`

thomcc: Hm? It's used in the switch. Or maybe I am misunderstanding.

lib/std/Thread/Futex.zig
thomcc: The variable `timeout_overflowed` needs to be used by the switch, since if we clamp the provided timeout, it would be invalid for us to turn the system `ETIMEDOUT` into `error.TimedOut` (as we timed out on a timeout value less than the one they provided).

lib/std/Thread/Futex.zig
thomcc: Okay. It's a bit gross since we need to map null to 0 still, but yeah it's probably worth checking to ensure that assumption doesn't change.

lib/std/Thread/Futex.zig
thomcc: Yeah. The var needs to be outside the block still, but otherwise I've done that change.

I also tried to add more explanation to the comment above it about what `var timeout_overflowed` is doing, since the existing comment didn't seem to get the point across.

lib/std/Thread/Futex.zig
daurnimator: ah; I didn't see it there.
The new comment is helpful!

kprotty:
Nice, any other changes you wish to make?

thomcc:
Nope, unless anybody has issues I think its good to land

PR #9554, opened 2021-08-12
translate-c: allow string literals to be used as `char *`

### LINKED ISSUE #9126: translate-c: C code that uses string literals as non-const char *
In C string literals have the type `char[N]` (not const, but modifying the contents is undefined behavior). C code that uses this isn't compatible with Zig's `[]const u8` string literals:

```C
int main(void) {
   char *foo = "foo";
   return 0;
}
```

```zig
pub export fn main() c_int {
    var foo: [*c]u8 = "foo";
    return 0;
}
```

```
./test.zig:3:23: error: expected type '[*c]u8', found '*const [3:0]u8'
    var foo: [*c]u8 = "foo";
                      ^
./test.zig:3:23: note: cast discards const qualifier
    var foo: [*c]u8 = "foo";
```

This can occur in variable declarations, function calls, and struct and array initializers - did I miss any?

I noticed in `getExprQualType` we turn `char *` into `const char *` but I'm not sure why or what the implications of removing that are. Another solution would be to detect situations where a string literal is used as a `char *` and insert intToPtr / ptrToInt to cast away the constness.

ehaas:
In C the type of string literals is `char *`, so when using them in
a non-const context we have to cast the const away.

Fixes #9126

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -721,27 +721,13 @@ fn transQualTypeMaybeInitialized(c: *Context, scope: *Scope, qt: clang.QualType,
 
 /// This is used in global scope to convert a string literal `S` to [*c]u8:
 /// &(struct {
-///     var static: @TypeOf(S.*) = S.*;
+///     var static = S.*;
 /// }).static;
 fn stringLiteralToCharStar(c: *Context, str: Node) Error!Node {
     const var_name = Scope.Block.StaticInnerName;
 
-    const derefed = try Tag.deref.create(c.arena, str);
-    const var_type = try Tag.typeof.create(c.arena, derefed);
-
     const variables = try c.arena.alloc(Node, 1);
-    variables[0] = try Tag.var_decl.create(c.arena, .{
-        .is_pub = false,
-        .is_const = false,
-        .is_extern = false,
-        .is_export = false,
-        .is_threadlocal = false,
-        .linksection_string = null,
-        .alignment = null,
-        .name = var_name,
-        .type = var_type,
-        .init = derefed,
-    });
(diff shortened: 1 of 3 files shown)


ehaas:
Question: is this a valid Zig construct (this is how I'm translating the C global variable `char *s = "my_string";`):

```zig
pub export var s: [*c]u8 = blk: {
    var string_temp: @TypeOf("my_string".*) = "my_string".*;
    break :blk &string_temp;
};
```

Vexu:
> Question: is this a valid Zig construct

It is now but won't be after #7396. If you put the var inside a struct it should be good.

ehaas:
So something like this (possibly with an intermediate variable to hold the struct depending on what is easier to do with existing translate-c constructs):

```zig
pub export var s: [*c]u8 = blk: {
    break :blk &struct {
        var string_temp: @TypeOf("my_string".*) = "my_string".*;
    }.string_temp;
};
```

Vexu:
> So something like this

Yup, placing it inside a struct gives it static lifetime.

src/translate_c.zig
```diff
@@ -719,6 +719,30 @@ fn transQualTypeMaybeInitialized(c: *Context, scope: *Scope, qt: clang.QualType,
         transQualType(c, scope, qt, loc);
 }
 
+/// This is used in global scope to convert a string literal `S` to [*c]u8:
+/// &(struct {
+///     var static: @TypeOf(S.*) = S.*;
```
ehaas, line 724: I don't want to restart the CI builds over a comment but this doc comment is out of date now

Issue #9519, opened 2021-08-04
Building via cgo for windows/arm64 (aarch64-windows-gnu)


TopperDEL:
I want to leverage zig to built a DLL of a go-library, that can be used in Windows ARM64 (basically Hololens 2). Using go1.17.1rc2 zig gets invoked, but brings up this error:

```
2021-08-04T11:54:57.4223210Z # runtime/cgo
2021-08-04T11:54:57.4224911Z /snap/zig/3678/lib/libc/mingw/secapi/vsprintf_s.c:39:10: warning: implicit declaration of function '__ms_vsnprintf' is invalid in C99 [-Wimplicit-function-declaration]
2021-08-04T11:54:57.4225714Z   return __ms_vsnprintf (_DstBuf, _Size, _Format, _ArgList);
2021-08-04T11:54:57.4226223Z          ^
2021-08-04T11:54:57.4226624Z 1 warning generated.
2021-08-04T11:54:57.4227534Z /snap/zig/3678/lib/libc/mingw/math/arm/s_trunc.c/snap/zig/3678/lib/libc/mingw/math/arm/s_truncf.c:24:10: fatal error: '../bsd_private_base.h' file not found
2021-08-04T11:54:57.4228188Z #include "../bsd_private_base.h"
2021-08-04T11:54:57.4228651Z          ^~~~~~~~~~~~~~~~~~~~~~~
2021-08-04T11:54:57.4229332Z :26:10: fatal error: '../bsd_private_base.h' file not found
2021-08-04T11:54:57.4229850Z #include "../bsd_private_base.h"
2021-08-04T11:54:57.4230310Z          ^~~~~~~~~~~~~~~~~~~~~~~
2021-08-04T11:54:57.4230966Z 1 error generated.
2021-08-04T11:54:57.4231397Z 1 error generated.
2021-08-04T11:54:57.4522549Z ##[error]Bash exited with code '1'.
```

Any hint what I could do/try? Thank you!

I'm using zig 0.8.0

kristoff-it:
Hi, I assume you read

https://dev.to/kristoff/zig-makes-go-cross-compilation-just-work-29ho

in this post I described the workaround to a bug in Go that basically consists in Go's inability to handle the case where a compiler command has spaces in it (like in this case, where we have first the `cc` subcommand, and then the target option.

This means that when you create `zcc` you need to put in there not just `zig cc` but also the `-target` option, and that you need to do the same with `CXX`.

```sh
[...]

Issue #1290, opened 2018-07-25
Missing compiler_rt functions


skyfex:
I was playing around with the latest version of Zig (https://ci.appveyor.com/project/andrewrk/zig-d3l86/build/0.2.0+95f45cfc) on ARM Cortex-M0 and I had a lot of issues with missing compiler_rt functions. 

It was missing '__aeabi_memcpy'

It was complaining about missing '__aeabi_uldivmod', which I fixed by commenting out `if (isArmArch()) {` in compiler_rt/index.zig .. later I found out that it also helped to use `--target-arch armv6` instead of `thumb`, or adding `builtin.Arch.thumb` to `isArmArch` (so that should probably be fixed)

After that it complained about `__aeabi_h2f`, `__aeabi_f2h` and `__multi3`

This is the command I used
```
zig build-exe --static --target-os freestanding --target-arch thumb --target-environ eabihf  --libc-include-dir include --linker-script system\nrf51_xxaa.ld --verbose-link -isystem include --libc-include-dir include --library-path system --assembly system\gcc_startup_nrf51.S --object zig-cache/system_nrf51.o test.zig
```

Using `fmt.bufPrint` is what is triggering these errors for me right now.

Is it possible to get a more permanent/robust solution to compiler_rt related problems? Some kind of automated testing that everything is there somehow?

(Btw, good news is that linking with Zig (rather than GCC or LLD) seems to work just fine for me now)

skyfex:
I'm also often getting these warnings, don't know what they mean:
```lld: warning: lld may use movt/movw, no object with architecture supporting feature detected. ```

**Comment by radek-senfeld** [comment]
> After that it complained about `__aeabi_h2f`, `__aeabi_f2h` and `__multi3`

I had the same problem. It turns out you need to compile for **gnueabi** target in order to link using GCC ld.

EABI names: ```__aeabi_f2h, __aeabi_h2f```
GNUEABI names: ```__gnu_f2h_ieee, __gnu_h2f_ieee```

This works for me:

```
zig build-obj -target armv7m-freestanding-gnueabi --output-dir build/ /opt/zig/lib/zig/std/special/compiler_rt.zig

zig build-obj -target armv7m-freestanding-gnueabi --output-dir build/ src/main.zig

[...]

PR #9555, opened 2021-08-13
implement building via cgo for windows/arm64 (aarch64-windows-gnu)

marler8997:
fix issue #9519 by adding some missing stuff from mingw and compiler_rt to support windows aarch64.

### DIFF
--- a/lib/libc/mingw/math/arm-common/log2.c
+++ b/lib/libc/mingw/math/arm-common/log2.c
@@ -1,0 +1,55 @@
+/**
+ * This file has no copyright assigned and is placed in the Public Domain.
+ * This file is part of the mingw-w64 runtime package.
+ * No warranty is given; refer to the file DISCLAIMER.PD within this package.
+ */
+
+#include <math.h>
+#include <stdint.h>
+
+typedef union ieee754_double_ {
+  struct __attribute__((__packed__)) {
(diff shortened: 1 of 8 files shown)


marler8997:
This appears to be now blocked by https://github.com/ziglang/zig/issues/1290

I added `__chkstk` but then got more missing symbols:

```
lld-link: error: undefined symbol: __udivti3
>>> referenced by compiler_rt.lib(compiler_rt.obj):(.weak.__muloti4.default._alloca)

lld-link: error: undefined symbol: __divti3
>>> referenced by compiler_rt.lib(compiler_rt.obj):(.weak.__muloti4.default._alloca)

```

marler8997:
Ok I've adding the missing compiler_rt functions.  I'm able to compile a basic hello world program for `aarch64-windows-gnu`, but I'm not sure what I need to modify in order to add `aarch64-windows-gnu` to the list of cross platforms that are tested.

andrewrk:
Wow very cool!

marler8997:
wow I can't believe the new target is already passing the test suite

marler8997:
With the `aarch64-windows-gnu` target now working, I confirmed this fixes the original issue and I can use `zig cc` to cross compile the original Go project.

marler8997:
Ping

PR #9558, opened 2021-08-13
mingw.zig: fix logic to add crt sources

marler8997:
The current version of code uses isARM to check if we are compiling to any arm target then checks the target bit width to either add the 32-bit sources or 64-bit source.  However, isARM only returns true for 32-bit targets, and isAARCH64 is for the 64-bit targets.

I also replaced the unreachable with a @panic when we receive an unsupported arch because this code is reachable and should turn into an error.  With the current unreachable statement, the released compiler on windows was silently hitting the unreachable and falling through to the `isARM` case.

This is one of the errors contributing to https://github.com/ziglang/zig/issues/9519

It would be great if I could also add a regression test for this, not sure how though...any guidance here would be very helpful.

### DIFF
--- a/src/mingw.zig
+++ b/src/mingw.zig
@@ -187,27 +187,25 @@ pub fn buildCRTFile(comp: *Compilation, crt_file: CRTFile) !void {
                     };
                 }
             } else if (target.cpu.arch.isARM()) {
-                if (target.cpu.arch.ptrBitWidth() == 32) {
-                    for (mingwex_arm32_src) |dep| {
-                        (try c_source_files.addOne()).* = .{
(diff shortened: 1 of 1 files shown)


Vexu (changes requested):
`std.Target.Cpu.Arch.isAARCH64()` does not exist.

marler8997:
> `std.Target.Cpu.Arch.isAARCH64()` does not exist.

Not sure what you're saying.  Your casing is incorrect, it's supposed to be `target.cpu.arch.isAARCH64()`, and if it didn't exist then zig1.o wouldn't compile but it does, even if I rebase this branch wth master.

Vexu:
Huh, I have no idea where I was looking but I couldn't manage to find it earlier.

PR #9560, opened 2021-08-14
fs.Dir.Walker: Do not close the initial dir during/after walking it

### LINKED ISSUE #9556: fs.Dir.Walker: walked directory gets closed during iteration
I expect the following code to be correct, but it fails since the directory gets closed during the `walker.next()` call and also when the dir is closed in the `defer tmp.cleanup()`.

```zig
const std = @import("std");

test "dir closed during walk" {
    var allocator = std.testing.allocator;

    var tmp = std.testing.tmpDir(.{ .iterate = true });
    defer tmp.cleanup(); // will close tmp.dir

    var walker = try tmp.dir.walk(allocator);
    defer walker.deinit();

    // closes tmp.dir during next()
    while (try walker.next()) |entry| {
        std.debug.print("{}: basename: '{s}' path: '{s}'\n", .{ entry.kind, entry.basename, entry.path });
    }
}
```

The test case in `fs/test.zig` gets around this by opening a second dir handle and then calling `walk()` with that, but that seems less than ideal:

https://github.com/ziglang/zig/blob/0961c1975a74acdc344be11ffbb4b11e033c8df0/lib/std/fs/test.zig#L929-L933

It seems like it'd be better if walking didn't close the initial directory now that it's inside `Dir` (https://github.com/ziglang/zig/pull/9475).

Here's the error trace for failing test case above for completeness:

```
$ zig test walker-close.zig
thread 500482 panic: reached unreachable code
/home/ryan/Programming/zig/zig/build/lib/zig/std/os.zig:124:18: 0x212692 in std.os.close (test)
        EBADF => unreachable, // Always a race condition.
                 ^
/home/ryan/Programming/zig/zig/build/lib/zig/std/fs.zig:847:21: 0x20fa96 in std.fs.Dir.close (test)
            os.close(self.fd);
                    ^
/home/ryan/Programming/zig/zig/build/lib/zig/std/testing.zig:321:23: 0x20c221 in std.testing.TmpDir.cleanup (test)
        self.dir.close();
                      ^
/home/ryan/Programming/zig/tmp/walker-close.zig:7:22: 0x20b45d in test "WalkerEntry.basename" (test)
    defer tmp.cleanup(); // closes tmp.dir
                     ^
/home/ryan/Programming/zig/zig/build/lib/zig/std/special/test_runner.zig:76:28: 0x23f0f2 in std.special.main (test)
        } else test_fn.func();
                           ^
/home/ryan/Programming/zig/zig/build/lib/zig/std/start.zig:498:22: 0x23722c in std.start.callMain (test)
            root.main();
                     ^
/home/ryan/Programming/zig/zig/build/lib/zig/std/start.zig:450:12: 0x212cce in std.start.callMainWithArgs (test)
           ^
/home/ryan/Programming/zig/zig/build/lib/zig/std/start.zig:369:17: 0x20cea6 in std.start.posixCallMainAndExit (test)
                ^
/home/ryan/Programming/zig/zig/build/lib/zig/std/start.zig:282:5: 0x20ccb2 in std.start._start (test)
    ^
```

squeek502:
Closing the initial directory was unexpected to me, and does not mesh very well with how the rest of the Dir API works.

Fixes #9556, more details there

---

Note: The `"walker"` test can be cleaned up further (can remove the ArenaAllocator) once #9559 is merged, so merge that one first and I'll update this one after.

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -795,22 +795,31 @@ pub const Dir = struct {
                         .kind = base.kind,
                     };
                 } else {
-                    self.stack.pop().iter.dir.close();
+                    var item = self.stack.pop();
+                    if (self.stack.items.len != 0) {
+                        item.iter.dir.close();
+                    }
                 }
             }
             return null;
         }
 
         pub fn deinit(self: *Walker) void {
-            while (self.stack.popOrNull()) |*item| item.iter.dir.close();
+            while (self.stack.popOrNull()) |*item| {
+                if (self.stack.items.len != 0) {
+                    item.iter.dir.close();
+                }
+            }
             self.stack.deinit();
             self.name_buffer.deinit();
         }
     };
 
     /// Recursively iterates over a directory.
+    /// `self` must have been opened with `OpenDirOptions{.iterate = true}`.
     /// Must call `Walker.deinit` when done.
     /// The order of returned file system entries is undefined.
+    /// `self` will not be closed after walking it.
     pub fn walk(self: Dir, allocator: *Allocator) !Walker {
         var name_buffer = std.ArrayList(u8).init(allocator);
         errdefer name_buffer.deinit();
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -3026,7 +3026,8 @@ pub const InstallDirStep = struct {
         const self = @fieldParentPtr(InstallDirStep, "step", step);
         const dest_prefix = self.builder.getInstallPath(self.options.install_dir, self.options.install_subdir);
         const full_src_dir = self.builder.pathFromRoot(self.options.source_dir);
-        const src_dir = try std.fs.cwd().openDir(full_src_dir, .{ .iterate = true });
+        var src_dir = try std.fs.cwd().openDir(full_src_dir, .{ .iterate = true });
+        defer src_dir.close();
         var it = try src_dir.walk(self.builder.allocator);
         next_entry: while (try it.next()) |entry| {
             for (self.options.exclude_extensions) |ext| {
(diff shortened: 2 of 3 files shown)


squeek502:
Updated now that #9559 has been merged. This should be good to go.

lib/std/fs.zig
```diff
@@ -795,22 +795,31 @@ pub const Dir = struct {
                         .kind = base.kind,
                     };
                 } else {
-                    self.stack.pop().iter.dir.close();
+                    var item = self.stack.pop();
+                    if (item.dirname_len != 0) {
```
Vexu, line 799:
```suggestion
                    if (self.stack.len != 0) {
```
Both have the same effect but this conveys what property we're really interested in.

lib/std/fs.zig
```diff
@@ -795,22 +795,31 @@ pub const Dir = struct {
...
+                    }
                 }
             }
             return null;
         }
 
         pub fn deinit(self: *Walker) void {
-            while (self.stack.popOrNull()) |*item| item.iter.dir.close();
+            while (self.stack.popOrNull()) |*item| {
+                if (item.dirname_len != 0) {
```
Vexu, line 809: Same here.

Vexu (changes requested):

ominitay:
Apologies for this oversight! Perhaps the `` `self` will not be closed after walking it. `` comment is unnecessary, as it *is* expected behaviour. Again, sorry I didn't see this sooner; GitHub notifications failed me lol

squeek502:
> Apologies for this oversight! Perhaps the `` `self` will not be closed after walking it. `` comment is unnecessary, as it _is_ expected behaviour.

No worries. I also wasn't sure about adding that doc comment but I think it's probably better to be explicit for now for anyone updating their code from the closing version or the old non-Dir version. That part of the doc comment can always be removed later.

PR #9562, opened 2021-08-14
std.fmt: add support for printing slices of strings

nektro:
```zig
const std = @import("std");

const T = struct {
    a: i32,
    b: i32,
    c: i32,
};

pub fn main() void {
    std.log.info("{s}", .{std.meta.fieldNames(T)});
}
```

Before: `error: Unknown format string: 's'`

After: `info: a, b, c`

I decided to leave out the square brackets from the formatter so that each user may choose whichever bracket type makes the most sense for them (or keep with none at all)

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -544,6 +544,13 @@ pub fn formatType(
                             return formatText(value, actual_fmt, options, writer);
                         }
                     }
+                    if (comptime std.meta.trait.isZigString(info.child)) {
+                        for (value) |item, i| {
+                            if (i != 0) try formatText(", ", actual_fmt, options, writer);
+                            try formatText(item, actual_fmt, options, writer);
+                        }
+                        return;
+                    }
                     @compileError("Unknown format string: '" ++ actual_fmt ++ "'");
                 },
                 .Enum, .Union, .Struct => {


Jarred-Sumner:
this is great

dbandstra:
IMO this (pointer to array) should work the same as how arrays and slices already work. They put braces around the list, and also support more than just "strings". If that's not ideal, then maybe all the three sites could be changed, but I think they should be consistent with one another.


```zig
const std = @import("std");

pub fn main() void {
    // array
    std.debug.warn("{s}\n", .{
        @as([2][]const u8, .{ "one", "two" }),
    });

    // slice
    std.debug.warn("{s}\n", .{
        @as([]const []const u8, &.{ "one", "two" }),
    });

    // pointer to array (new)
    std.debug.warn("{s}\n", .{
        @as(*const [2][]const u8, &.{ "one", "two" }),
    });
}
```

Outputs:

```
{ one, two }
{ one, two }
one, two
```

If you change the `{s}` to `{d}` in the three `std.debug.warn` calls, the first two print `{ { 111, 110, 101 }, { 116, 119, 111 } }` and the third fails to compile.

I think this could be fixed by more or less copy pasting from the code for slices: https://github.com/ziglang/zig/blob/6a5094872f10acc629543cc7f10533b438d0283a/lib/std/fmt.zig#L574-L594

PR #9563, opened 2021-08-14
Fix args when calling clang::ASTUnit::LoadFromCommandLine.

tamaskenez:
`clang::ASTUnit::LoadFromCommandLine` interprets the first argument as the name of program (just like the `main` function).

Currently, the first 2 arguments we pass to this function is `"-x"` and `"c"` which are interpreted as a program name and an unused linker input, resulting in a _"linker input unused"_ warning from clang.

This PR shifts the arguments passing `""` for the first argument making `"-x c"` interpreted as intended.

Let me know if you prefer this to be fixed at the 2 callsites of `translate_c.translate` where the args get assembled.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -2492,6 +2492,7 @@ fn cmdTranslateC(comp: *Compilation, arena: *Allocator, enable_cache: bool) !voi
 
     const digest = if (try man.hit()) man.final() else digest: {
         var argv = std.ArrayList([]const u8).init(arena);
+        try argv.append(""); // argv[0] is program name, actual args start at [1]
 
         var zig_cache_tmp_dir = try comp.local_cache_directory.handle.makeOpenPath("tmp", .{});
         defer zig_cache_tmp_dir.close();
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -2557,6 +2557,7 @@ pub fn cImport(comp: *Compilation, c_src: []const u8) !CImportResult {
         var argv = std.ArrayList([]const u8).init(comp.gpa);
         defer argv.deinit();
 
+        try argv.append(""); // argv[0] is program name, actual args start at [1]
         try comp.addTranslateCCArgs(arena, &argv, .c, out_dep_path);
 
         try argv.append(out_h_path);


Vexu:
How do you reproduce this?

> Let me know if you prefer this to be fixed at the 2 callsites of `translate_c.translate` where the args get assembled.

This sounds like a bug in our API usage so that is also where it should be fixed.

tamaskenez:
I reproduced it by dumping the clang diagnostics `zig_lang.cpp` even if there are no errors (and calling `zig` for a source with `cimport`).
Now I'm converting this PR to draft and fixing it at the callsites of `translate`.

Vexu:
I was able to reproduce the issue with:
```cpp
template<typename T>
T foo(void) {
    return 0;
}
```
which doesn't compile with `-x c`.

Fixed with:
```diff
diff --git a/src/Compilation.zig b/src/Compilation.zig
index 7987ed95d..8d629470b 100644
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -2925,7 +2925,7 @@ pub fn addTranslateCCArgs(
     ext: FileExt,
     out_dep_path: ?[]const u8,
 ) !void {
-    try argv.appendSlice(&[_][]const u8{ "-x", "c" });
+    try argv.appendSlice(&[_][]const u8{ "clang", "-x", "c" });
     try comp.addCCArgs(arena, argv, ext, out_dep_path);
     // This gives us access to preprocessing entities, presumably at the cost of performance.
     try argv.appendSlice(&[_][]const u8{ "-Xclang", "-detailed-preprocessing-record" });
```

tamaskenez:
I changed the fix by adding an empty `argv[0]` at the the callsites of `translate_c.translate`. I tried to add a test but `run_translated_c` uses `source.c` as a file name and clang assumes `*.c` files are in the C language even without `"-x C"`. I could have tested the effect of "-x C" only if I had changed `run_translated_c` to use `source.cpp` which didn't seem right.

src/main.zig
```diff
@@ -2492,6 +2492,7 @@ fn cmdTranslateC(comp: *Compilation, arena: *Allocator, enable_cache: bool) !voi
 
     const digest = if (try man.hit()) man.final() else digest: {
         var argv = std.ArrayList([]const u8).init(arena);
```
tamaskenez, line 2494: Unrelated: isn't a `defer argv.deinit()` missing here?

PR #9568, opened 2021-08-15
macho: don't embed codesig unless targeting aarch64-macos

### LINKED ISSUE #9565: Can't codesign aarch64-ios binaries on Intel mac
I am able to build a full iOS app with Zig after #9532 - yay! And https://github.com/kubkon/zig-ios-example is a great simplified example of how to do that and run it in a simulator. However, something is preventing me from code-signing the app, which is required to run it on a real iOS device. I am using `zig-macos-x86_64-0.9.0-dev.796+fcf2ce0ff`, and running the following on @kubkon's example repo:

- `zig build -Dtarget=aarch64-ios`
- `codesign -s "<my-adhoc-certificate>" zig-out/bin/MadeWithZig.app`

The code signing fails with
`zig-out/bin/MadeWithZig.app: is already signed`
So I tried using the `-f`/`--force` flag:
- `codesign -f -s "<my-adhoc-certificate>" zig-out/bin/MadeWithZig.app`
and it spilled the beans:
```
zig-out/bin/MadeWithZig.app: replacing existing signature
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/codesign_allocate: fatal error: file not in an order that can be processed (code signature data out of place): /Users/legionjr/Documents/zig-ios-example/zig-out/bin/MadeWithZig.app/app
zig-out/bin/MadeWithZig.app: the codesign_allocate helper tool cannot be found or used
```
To inspect this existing signature, I ran
- `codesign -dv zig-out/bin/MadeWithZig.app`
which prints
```
Executable=/Users/legionjr/Documents/zig-ios-example/zig-out/bin/MadeWithZig.app/app
Identifier=app
Format=app bundle with Mach-O thin (arm64)
CodeDirectory v=20400 size=252 flags=0x2(adhoc) hashes=5+0 location=embedded
Signature=adhoc
Info.plist=not bound
TeamIdentifier=not set
Sealed Resources=none
Internal requirements=none
```

Poking around the Zig source, I see there is some logic now that throws in a code signature during MachO linking. Maybe this is something that was put in for macOS, but should be done differently for iOS builds? Not familiar with the expected format of this code signature, but `codesign` definitely doesn't like it...

kubkon:
Fixes #9565

When developing an iOS app for example, the developer is required to use Apple's codesign utility to generate a valid signature as done by Xcode.

### DIFF
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -396,6 +396,13 @@ pub fn openPath(allocator: *Allocator, sub_path: []const u8, options: link.Optio
 
 pub fn createEmpty(gpa: *Allocator, options: link.Options) !*MachO {
     const self = try gpa.create(MachO);
+    const cpu_arch = options.target.cpu.arch;
+    const os_tag = options.target.os.tag;
+    const abi = options.target.abi;
+    const page_size = if (cpu_arch == .aarch64) 0x4000 else 0x1000;
+    // Adhoc code signature is required when targeting aarch64-macos either directly or indirectly via the simulator
+    // ABI such as aarch64-ios-simulator, etc.
+    const requires_adhoc_codesig = cpu_arch == .aarch64 and (os_tag == .macos or abi == .simulator);
 
     self.* = .{
         .base = .{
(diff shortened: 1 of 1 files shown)


jmrico01:
PR already, nice! Came to post a build error, seems like CI checks are already complaining about it haha
```
./src/link/MachO.zig:402:49: error: cannot store runtime value in type 'comptime_int'
    const page_size = if (cpu_arch == .aarch64) 0x4000 else 0x1000;
                                                ^

```

That error doesn't make too much sense to me, but anyways, I got around it by using an explicit int type for now. more importantly though...

**`codesign` works! AND I can run my Zig iOS app on a real iPhone!** 🎉 🥳  **THANK YOU!!!**

PR #9571, opened 2021-08-17
stage2 Air: add struct_field_ptr_index_{0..3}

g-w1:
Since these are very common, it will save memory.

Mentioned at 2:26 in this stream https://vimeo.com/572815502

"I sped up the compiler by some percent or something"

I added the extra test case to make sure that I did not break structs that did not use the optimization.

### DIFF
--- a/src/Air.zig
+++ b/src/Air.zig
@@ -264,6 +264,13 @@ pub const Inst = struct {
         /// Given a pointer to a struct and a field index, returns a pointer to the field.
         /// Uses the `ty_pl` field, payload is `StructField`.
         struct_field_ptr,
+        /// Given a pointer to a struct, returns a pointer to the field.
+        /// The field index is the number at the end of the name.
+        /// Uses `ty_op` field.
+        struct_field_ptr_index_0,
+        struct_field_ptr_index_1,
+        struct_field_ptr_index_2,
+        struct_field_ptr_index_3,
         /// Given a byval struct and a field index, returns the field byval.
         /// Uses the `ty_pl` field, payload is `StructField`.
         struct_field_val,
@@ -510,6 +517,10 @@ pub fn typeOfIndex(air: Air, inst: Air.Inst.Index) Type {
         .wrap_errunion_payload,
         .wrap_errunion_err,
         .slice_ptr,
+        .struct_field_ptr_index_0,
+        .struct_field_ptr_index_1,
+        .struct_field_ptr_index_2,
+        .struct_field_ptr_index_3,
         => return air.getRefType(datas[inst].ty_op.ty),
 
         .loop,
--- a/src/codegen.zig
+++ b/src/codegen.zig
@@ -855,6 +855,12 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
                     .store           => try self.airStore(inst),
                     .struct_field_ptr=> try self.airStructFieldPtr(inst),
                     .struct_field_val=> try self.airStructFieldVal(inst),
+
+                    .struct_field_ptr_index_0 => try self.airStructFieldPtrIndex(inst, 0),
+                    .struct_field_ptr_index_1 => try self.airStructFieldPtrIndex(inst, 1),
+                    .struct_field_ptr_index_2 => try self.airStructFieldPtrIndex(inst, 2),
+                    .struct_field_ptr_index_3 => try self.airStructFieldPtrIndex(inst, 3),
+
                     .switch_br       => try self.airSwitch(inst),
                     .slice_ptr       => try self.airSlicePtr(inst),
                     .slice_len       => try self.airSliceLen(inst),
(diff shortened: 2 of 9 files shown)


andrewrk:
Prior to merging this, I think we should:
 * Actually use the new AIR instructions
 * Measure a perf improvement with a benchmark

g-w1:
Perf:
test.zig: https://paste.rs/gHy
branch: i-wrote
```
❯ $(which time) -v ../zig-out/bin/zig build-exe test.zig -ofmt=c
	Command being timed: "../zig-out/bin/zig build-exe test.zig -ofmt=c"
	User time (seconds): 0.23
	System time (seconds): 0.07
	Percent of CPU this job got: 100%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.30
	Maximum resident set size (kbytes): 111296
	Minor (reclaiming a frame) page faults: 59427
	Voluntary context switches: 180
	Involuntary context switches: 5
	Swaps: 0
	File system inputs: 0
	File system outputs: 55944
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
```
branch: master
```
❯ $(which time) -v ../zig-out/bin/zig build-exe test.zig -ofmt=c
	Command being timed: "../zig-out/bin/zig build-exe test.zig -ofmt=c"
	User time (seconds): 0.26
	System time (seconds): 0.07
	Percent of CPU this job got: 104%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.32
	Maximum resident set size (kbytes): 114828
	Minor (reclaiming a frame) page faults: 64622
	Voluntary context switches: 100
	Involuntary context switches: 16
	Swaps: 0
	File system inputs: 0
	File system outputs: 56928
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
```
It is an improvement!

src/codegen/c.zig
```diff
@@ -1649,15 +1655,29 @@ fn airOptionalPayload(o: *Object, inst: Air.Inst.Index) !CValue {
 
 fn airStructFieldPtr(o: *Object, inst: Air.Inst.Index) !CValue {
     if (o.liveness.isUnused(inst))
-        return CValue.none;
+        return @as(CValue, CValue.none);
```
andrewrk, line 1658: Is this change needed?

src/codegen/c.zig
```diff
@@ -1649,15 +1655,29 @@ fn airOptionalPayload(o: *Object, inst: Air.Inst.Index) !CValue {
...
 
     const ty_pl = o.air.instructions.items(.data)[inst].ty_pl;
     const extra = o.air.extraData(Air.StructField, ty_pl.payload).data;
-    const writer = o.writer();
     const struct_ptr = try o.resolveInst(extra.struct_operand);
     const struct_ptr_ty = o.air.typeOf(extra.struct_operand);
+    return airStructFieldPtrFinish(o, inst, struct_ptr_ty, struct_ptr, extra.field_index);
```
andrewrk, line 1664:
```suggestion
    return structFieldPtr(o, inst, struct_ptr_ty, struct_ptr, extra.field_index);
```

Reserve "air" prefix for the functions that are called directly from that big switch

src/Air.zig
```diff
@@ -264,6 +264,13 @@ pub const Inst = struct {
         /// Given a pointer to a struct and a field index, returns a pointer to the field.
         /// Uses the `ty_pl` field, payload is `StructField`.
         struct_field_ptr,
+        /// Same as above, but for common cases to save extra data memory.
```
andrewrk, line 267: Generally doc comments should reference stuff by name rather than saying "above" or "below" in case someone is looking at documentation out of context, or e.g. looking at generated docs of only 1 field.

Likewise they should be duplicated even for similar things, such as each of these new tags.

src/codegen.zig
```diff
@@ -1597,6 +1603,12 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
             //return self.finishAir(inst, result, .{ extra.struct_ptr, .none, .none });
         }
 
+        fn airStructFieldPtrIndex(self: *Self, inst: Air.Inst.Index, index: u8) !void {
+            const ty_op = self.air.instructions.items(.data)[inst].ty_op;
+            _ = ty_op;
+            return self.fail("TODO implement codegen struct_field_ptr_index_{d}", .{index});
```
andrewrk, line 1609: Can you wire this up to the other instruction so that they share a code path?

src/codegen/wasm.zig
```diff
@@ -1443,6 +1447,12 @@ pub const Context = struct {
 
         return WValue{ .local = struct_ptr.multi_value.index + @intCast(u32, extra.data.field_index) };
     }
+    fn airStructFieldPtrIndex(self: *Context, inst: Air.Inst.Index, index: u32) InnerError!WValue {
+        const ty_op = self.air.instructions.items(.data)[inst].ty_op;
+        const struct_ptr = self.resolveInst(ty_op.operand);
+
+        return WValue{ .local = struct_ptr.multi_value.index + index };
```
andrewrk, line 1454: Let's wire these functions together since there is shared logic (even though it is pretty simple)

andrewrk (changes requested):
Thanks for this. Looks almost ready, just a few cleanups.

andrewrk:
>  20 mili-seconds 

The duration is arbitrary since the test case could be arbitrarily long. So what is interesting is the ratio:

```
0.32 / 0.30 == 1.07
```

So you could say that this improves wall clock time in this execution of this test case by a factor of 1.07.

marler8997:
@g-w1 what's the deal with using `$(which time)` instead of just `time`?

PR #9576, opened 2021-08-17
compiler_rt: fix generic __clzsi2

matu3ba:
Example for failure: 1..1 (32 1's)
After applying shifts 16,8,4,2,1 we have y=0..0 (31 0's) and 1.
n=32-16-8-4-2-1=1, which is not 0.


matu3ba:
Probably just a copy-paste error from Hacker's Delight.

andrewrk:
Can we get test coverage for the mistake?

PR #9577, opened 2021-08-17
std/crypto: add scrypt kdf

x13a:
Hello everyone!

This is the second time to try to add scrypt kdf.

```text
add scrypt kdf
add phc encoding parser
add password hashing functions to benchmark
change bcrypt to be consistent with scrypt
```

Previous pull: [8200](https://github.com/ziglang/zig/pull/8200)
/cc @jedisct1

### DIFF
--- a/lib/std/crypto/bcrypt.zig
+++ b/lib/std/crypto/bcrypt.zig
@@ -6,21 +6,28 @@
 
 const std = @import("std");
 const crypto = std.crypto;
+const debug = std.debug;
 const fmt = std.fmt;
 const math = std.math;
 const mem = std.mem;
-const debug = std.debug;
+const pwhash = crypto.pwhash;
 const testing = std.testing;
 const utils = crypto.utils;
-const EncodingError = crypto.errors.EncodingError;
-const PasswordVerificationError = crypto.errors.PasswordVerificationError;
+
+const phc_format = @import("phc_encoding.zig");
+
+const KdfError = pwhash.KdfError;
+const HasherError = pwhash.HasherError;
+const EncodingError = phc_format.Error;
+const Error = pwhash.Error;
 
 const salt_length: usize = 16;
 const salt_str_length: usize = 22;
 const ct_str_length: usize = 31;
 const ct_length: usize = 24;
+const dk_length: usize = ct_length - 1;
 
-/// Length (in bytes) of a password hash
+/// Length (in bytes) of a password hash in crypt encoding
 pub const hash_length: usize = 60;
 
 const State = struct {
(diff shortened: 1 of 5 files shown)


jedisct1:
Good job, @x13a ! And happy to see that CI is now green! 🎉

I'm going to review it as soon as possible.

Yescrypt is the successor to scrypt and is already used in Fedora and Debian, so we want to make sure that it can be added without breaking the API. The generic parameters should already take care of this. So only the fact that Yescrypt supports encryption might require a minor API change.

Thanks again for this!

lib/std/crypto/bcrypt.zig
```diff
@@ -230,18 +180,203 @@ fn strHashInternal(password: []const u8, rounds_log: u6, salt: [salt_length]u8)
...
-    var ct_str: [ct_str_length]u8 = undefined;
-    Codec.encode(ct_str[0..], ct[0 .. ct.len - 1]);
+/// Hash and verify passwords using the PHC format.
+const PhcFormatHasher = struct {
+    const alg_id = "bcrypt";
+    const BinValue = phc_format.BinValue;
+
+    const HashResult = struct {
+        alg_id: []const u8,
+        rl: u6,
```
jedisct1, line 289: Where does this `rl` parameter come from?

Using the PHC format with `bcrypt` is not common. But:

- Python's `passlib` [uses `r`](https://passlib.readthedocs.io/en/stable/lib/passlib.hash.bcrypt_sha256.html) (and pre-hashes the input)
- This JavaScript implementation [uses `r` too](https://www.npmjs.com/package/phc-bcrypt)

So, maybe we should use the same convention (for the lack of a specification)

lib/std/crypto/phc_encoding.zig
```diff
@@ -0,0 +1,391 @@
...
+const B64Decoder = std.base64.standard_no_pad.Decoder;
+const B64Encoder = std.base64.standard_no_pad.Encoder;
+
+/// A wrapped binary value whose maximum size is `max_len`.
+///
+/// This type must be used whenever a binary value is encoded in a PHC-formatted string.
+/// This includes `salt`, `hash`, and any other binary parameters such as keys.
+///
+/// Once initialized, the actual value can be read with the `unwrap()` function.
+pub fn BinValue(comptime max_len: usize) type {
```
jedisct1, line 31: This is the same as https://github.com/ziglang/zig/pull/9134 albeit with the base64 encoding functions.

Maybe use `slice()`/`constSlice()` instead of `unwrap()` for consistency?

lib/std/crypto/phc_encoding.zig
```diff
@@ -0,0 +1,391 @@
...
+    var set_fields: usize = 0;
+
+    // FIXME
+    // macOS x86_64
+    // ./build/release/bin/zig test ./lib/std/crypto/phc_encoding.zig -OReleaseFast --test-filter "encoding"
+    // .{
+    //     .str = "$scrypt$ln=15,r=8,p=1",
+    //     .HashResult = struct { alg_id: []const u8, alg_version: ?u30, ln: u6, r: u30, p: u30 },
+    // },
+    // out.alg_version != null (`00`)
```
jedisct1, line 92: Is there an issue here?

lib/std/crypto/scrypt.zig
```diff
@@ -0,0 +1,674 @@
...
+    }
+
+    i = 0;
+    while (i < n) : (i += 2) {
+        // 32bit downcast
+        var j = @intCast(usize, integerify(x, r) & (n - 1));
+        blockXor(x, v[j * (32 * r) ..], 2 * r);
+        blockMix(&tmp, x, y, r);
+
+        // 32bit downcast
```
jedisct1, line 119: I don't understand this comment. `usize` doesn't mean 32 bit.

lib/std/crypto/scrypt.zig
```diff
@@ -0,0 +1,674 @@
...
+            pub fn fromSlice(slice: []const u8) EncodingError!Self {
+                if (slice.len > capacity) return EncodingError.NoSpaceLeft;
+                var bin_value: Self = undefined;
+                mem.copy(u8, &bin_value.buf, slice);
+                bin_value.len = slice.len;
+                return bin_value;
+            }
+
+            /// Return the slice containing the actual value.
+            pub fn unwrap(self: Self) []const u8 {
```
jedisct1, line 258: Same here, maybe use `slice()`/`constSlice()`.

lib/std/crypto/bcrypt.zig
```diff
@@ -251,45 +386,106 @@ fn strHashInternal(password: []const u8, rounds_log: u6, salt: [salt_length]u8)
...
+}
 
-    try strVerify("$2b$08$WUQKyBCaKpziCwUXHiMVvu40dYVjkTxtWJlftl0PpjY2BxWSvFIEe".*, "The devil himself");
+test "bcrypt phc format" {
+    const hash_options = HashOptions{
+        .allocator = null,
+        .params = .{ .rounds_log = 5 },
+        .encoding = .phc,
+    };
+    const verify_options = VerifyOptions{ .allocator = null };
```
jedisct1, line 466: Is there a way to define `VerifyOptions` with a default value of `null` for `allocator` instead?

lib/std/crypto/phc_encoding.zig
x13a:
```zig
var out: HashResult = undefined;
```

On macOS arm `out.alg_version` will be `null`, but on macOS x86_64 it will be `00`. It was unexpected for me. I can't say for sure if it is a bug or a future.

```zig
if (@hasField(HashResult, "alg_version")) {
    if (@typeInfo(@TypeOf(out.alg_version)) == .Optional) {
        out.alg_version = null;
    }
}
```

It is a fix for `00`. But maybe it have to be `null` on all platforms and it is a bug inside zig?

lib/std/crypto/bcrypt.zig
x13a: It was for consistence)

lib/std/crypto/phc_encoding.zig
x13a: Removed comments, mark resolved

jedisct1: `var out: HashResult = undefined;`

`out` is `undefined`, so optional pointers are not assumed to be set to `null`.

Optionals in `HashResult` can have a `null` default value, but as demonstrated here, it is easy to forget. As deserialization is expected to not leave anything uninitialized, instead of assigning `undefined` to `HashResult` and making `alg_version` a special case, maybe use `mom::zeroes()` instead. This will prevent this kind of issue from happening again if more optional values are required.

jedisct1:
Let's get this in! 🍻

PR #9583, opened 2021-08-18
std.rand.Random: add enumValue()

jwhear:
This PR adds an `enumValue` method to `Random` analogous to the existing `boolean` and `int` methods:  given an enum type, it returns a random value from that enum.

A note on the test: using the `SequentialPrng` (as used by the other tests in the file) causes the method to always return the enum's first value.  I pulled on this thread for a while and couldn't figure out why.  Using the method with the default rng gives the expected nice random distribution, so I'm not sure what's going on.  I would like to make the test better and make sure I'm not missing anything, so any suggestions are very welcome.

### DIFF
--- a/lib/std/rand.zig
+++ b/lib/std/rand.zig
@@ -47,6 +47,30 @@ pub const Random = struct {
         return r.int(u1) != 0;
     }
 
+    /// Returns a random value from an enum, evenly distributed.
+    pub fn enumValue(r: *Random, comptime EnumType: type) EnumType {
+        if (comptime !std.meta.trait.is(.Enum)(EnumType)) {
+            @compileError("Random.enumValue requires an enum type, not a "++@typeName(EnumType));
+        }
+
+        // We won't use int -> enum casting because enum elements can have
+        //  arbitrary values.  Instead we'll randomly pick one of the type's
+        //  fields (values).
+        const options = comptime std.meta.fields(EnumType);
+        if (options.len == 0) {
+            @compileError("Cannot get a random value from an empty enum");
+        }
+        const index = r.uintLessThan(u64, options.len);
+        inline for (options) |field, i| {
+            if (i == index) {
+                return @field(EnumType, field.name);
+            }
+        }
+
+        // The above loop should be exhaustive
+        unreachable;
+    }
+
     /// Returns a random int `i` such that `0 <= i <= maxInt(T)`.
     /// `i` is evenly distributed.
     pub fn int(r: *Random, comptime T: type) T {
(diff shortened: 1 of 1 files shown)


lib/std/rand.zig
```diff
@@ -47,6 +47,30 @@ pub const Random = struct {
...
+            @compileError("Cannot get a random value from an empty enum");
+        }
+        const index = r.uintLessThan(u64, options.len);
+        inline for (options) |field, i| {
+            if (i == index) {
+                return @field(EnumType, field.name);
+            }
+        }
+
+        // The above loop should be exhaustive
```
g-w1, line 70: The general pattern is to put this comment after the unreachable on the same line:
```zig
unreachable; // The above loop should be exhaustive
```
so that the stack traces show it.

jwhear:
A quick justification for this addition.  I was writing a test data generator and needed to select a random value from an enum.  The most obvious way to do this was to generate a random integer, `0..max enum value`, and then `@intToEnum` it.  I believe this will be the route many others will take given a similar need. But there are some major problems with this approach:
* You have to figure out the max integer value for the enum
* enums can used signed or unsigned integers
* enums can be non-exhaustive
* and the big one: enums are not continuous but can have arbitrary gaps, e.g. `enum { a=1, b=4, c=9}`

Given these potential footguns, I believe a _correct_ standard library solution is justified.

lib/std/rand.zig
```diff
@@ -47,6 +47,30 @@ pub const Random = struct {
...
+    pub fn enumValue(r: *Random, comptime EnumType: type) EnumType {
+        if (comptime !std.meta.trait.is(.Enum)(EnumType)) {
+            @compileError("Random.enumValue requires an enum type, not a "++@typeName(EnumType));
+        }
+
+        // We won't use int -> enum casting because enum elements can have
+        //  arbitrary values.  Instead we'll randomly pick one of the type's
+        //  fields (values).
+        const options = comptime std.meta.fields(EnumType);
+        if (options.len == 0) {
```
g-w1, line 60: This check does not need to exist as empty enums are not a thing.

jwhear: Doh, for some reason I thought the spec allowed them.  Must have confused with empty structs.

pfgithub:
Can `std.enums.values` be used?

```zig
pub fn enumValue(r: *Random, comptime EnumType: type) EnumType {
    const array = std.enums.values(EnumType);
    const index = r.uintLessThan(usize, array.len);
    return array[index];
}
```

jwhear:
> Can `std.enums.values` be used?

I was unaware of this function (doesn't appear in the master version of the docs currently), but this is an excellent suggestion.  I will submit a simplified version once my local test run finishes.

PR #9584, opened 2021-08-19
Add mask before truncating dereferenced bit pointers

Snektron:
Loading the value of a non-power-of-2 int larger than 8 bits yields something as follows in stage 1:
```llvm
%1 = load i8, i8* %the_ptr, align 1
%2 = lshr i8 %1, 1
%3 = trunc i8 %2 to i1
```
Unfortunately, LLVM does not always clear the upper bits of the resulting value after the truncate, as is evident in this snippet:
```zig
const std = @import("std");

const A = packed struct {
    a: bool,
    b: bool,
    c: bool,
    d: bool,

    e: bool,
    f: bool,
    g: bool,
    h: bool,
};

const X = union {
    x: A,
    y: u64,
};

pub fn a(
    x0: i32,
    x1: i32,
    x2: i32,
    x3: i32,
    x4: i32,
    flag_a: bool,
    flag_b: bool,
) void {
    @breakpoint();
    _ = x0;
    _ = x1;
    _ = x2;
    _ = x3;
    _ = x4;
    _ = flag_a;
    std.debug.print("{}\n", .{@ptrCast(*const u8, &flag_b).*});
}

pub fn b(x: *X) void {
    a(0, 1, 2, 3, 4, x.x.a, x.x.b);
}

pub fn main() void {
    var flags = A{
        .a = false,
        .b = true,
        .c = false,
        .d = false,

        .e = false,
        .f = true,
        .g = false,
        .h = false,
    };
    var x = X{
        .x = flags,
    };
    b(&x);
}
```
Above program yields `17`, instead of the expected `1`. Note that the many arguments are required to force `flag_b` to be placed on the stack, and the union is also required. 

Of course, the upper bits of the result of the truncate are undefined, which would be fine, if llvm didn't try to use them afterwards. With this patch, LLVM is explicitly instructed to generate a masking operation before the truncate. This is also what clang itself seems to do when naively translating above snippet to C.

### DIFF
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -3831,10 +3831,14 @@ static LLVMValueRef ir_render_load_ptr(CodeGen *g, Stage1Air *executable,
     LLVMValueRef shift_amt_val = LLVMConstInt(LLVMTypeOf(containing_int), shift_amt, false);
     LLVMValueRef shifted_value = LLVMBuildLShr(g->builder, containing_int, shift_amt_val, "");
 
+    LLVMTypeRef same_size_int = LLVMIntType(size_in_bits);
+    LLVMValueRef mask = LLVMConstAllOnes(LLVMIntType(size_in_bits));
+    mask = LLVMConstZExt(mask, LLVMTypeOf(containing_int));
+    LLVMValueRef masked_value = LLVMBuildAnd(g->builder, shifted_value, mask, "");
+
     if (handle_is_ptr(g, child_type)) {
         LLVMValueRef result_loc = ir_llvm_value(g, instruction->result_loc);
-        LLVMTypeRef same_size_int = LLVMIntType(size_in_bits);
-        LLVMValueRef truncated_int = LLVMBuildTrunc(g->builder, shifted_value, same_size_int, "");
+        LLVMValueRef truncated_int = LLVMBuildTrunc(g->builder, masked_value, same_size_int, "");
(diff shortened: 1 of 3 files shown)


andrewrk:
Needs a new behavior test case please

Snektron:
Note that this bug only manifests itself in debug mode on my machine. I suspect it might also be different on other architectures. In any case, i think that the extra mask operation justifies the potential future headaches.

Snektron:
Oh and by the way, this may seem like an esoteric bug, but i hit it during stage 2 development.

This struct:
https://github.com/ziglang/zig/blob/4c9d41730e66b253d12472d616e3a519c79a41cb/src/Zir.zig#L2147-L2154

Is loaded here, causing an incorrect value for mutable for example for the type `*align(1:2:3) u8`:
https://github.com/ziglang/zig/blob/4c9d41730e66b253d12472d616e3a519c79a41cb/src/Sema.zig#L6430-L6432

Which in turn fails this check if both pointers have a different binary value for `mutable`:
https://github.com/ziglang/zig/blob/4c9d41730e66b253d12472d616e3a519c79a41cb/src/type.zig#L420

andrewrk (approved):
Looks good to merge provided the tests pass. We should also file an LLVM bug report upstream.

Snektron:
Also submitted as LLVM issue: https://bugs.llvm.org/show_bug.cgi?id=51529

andrewrk:
Brilliant, thank you for the LLVM bug!

PR #9587, opened 2021-08-19
stage2: implement shr and shl

g-w1:
This implements it in the llvm and c backends.
x86_64 will have to be a little more work.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -5303,8 +5303,25 @@ fn zirShr(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileError!A
     const tracy = trace(@src());
     defer tracy.end();
 
-    _ = inst;
-    return sema.mod.fail(&block.base, sema.src, "TODO implement zirShr", .{});
+    const inst_data = sema.code.instructions.items(.data)[inst].pl_node;
+    const src: LazySrcLoc = .{ .node_offset_bin_op = inst_data.src_node };
+    const lhs_src: LazySrcLoc = .{ .node_offset_bin_lhs = inst_data.src_node };
+    const rhs_src: LazySrcLoc = .{ .node_offset_bin_rhs = inst_data.src_node };
+    const extra = sema.code.extraData(Zir.Inst.Bin, inst_data.payload_index).data;
+    const lhs = sema.resolveInst(extra.lhs);
+    const rhs = sema.resolveInst(extra.rhs);
+
+    if (try sema.resolveMaybeUndefVal(block, lhs_src, lhs)) |lhs_val| {
+        if (try sema.resolveMaybeUndefVal(block, rhs_src, rhs)) |rhs_val| {
+            if (lhs_val.isUndef() or rhs_val.isUndef()) {
+                return sema.addConstUndef(sema.typeOf(lhs));
+            }
+            return sema.mod.fail(&block.base, src, "TODO implement comptime shr", .{});
+        }
+    }
+
+    try sema.requireRuntimeBlock(block, src);
+    return block.addBinOp(.shr, lhs, rhs);
 }
 
 fn zirBitwise(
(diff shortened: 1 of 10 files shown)


src/Sema.zig
```diff
@@ -5303,8 +5303,29 @@ fn zirShr(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileError!A
...
+    const src: LazySrcLoc = .{ .node_offset_bin_op = inst_data.src_node };
+    const lhs_src: LazySrcLoc = .{ .node_offset_bin_lhs = inst_data.src_node };
+    const rhs_src: LazySrcLoc = .{ .node_offset_bin_rhs = inst_data.src_node };
+    const extra = sema.code.extraData(Zir.Inst.Bin, inst_data.payload_index).data;
+    const lhs = sema.resolveInst(extra.lhs);
+    const rhs = sema.resolveInst(extra.rhs);
+    const instructions = &[_]Air.Inst.Ref{ lhs, rhs };
+    const resolved_type = try sema.resolvePeerTypes(block, src, instructions, .{ .override = &[_]LazySrcLoc{ lhs_src, rhs_src } });
+    const casted_lhs = try sema.coerce(block, resolved_type, lhs, lhs_src);
+    const casted_rhs = try sema.coerce(block, resolved_type, rhs, rhs_src);
```
g-w1, line 5316: I'm unsure if peer type casting this is correct, but llvm complained that the operands to a binop were different if the types were not the same, so they have to be the same somehow, not sure how though. Maybe cast the rhs to the type of the lhs?

g-w1: Thanks Robin for the answer, the casting had to be done in the LLVM backend.

src/codegen.zig
```diff
@@ -1270,6 +1272,24 @@ fn Function(comptime arch: std.Target.Cpu.Arch) type {
             return self.finishAir(inst, result, .{ bin_op.lhs, bin_op.rhs, .none });
         }
 
+        fn airShl(self: *Self, inst: Air.Inst.Index) !void {
+            const bin_op = self.air.instructions.items(.data)[inst].bin_op;
+            const result: MCValue = if (self.liveness.isUnused(inst)) .dead else switch (arch) {
+                .arm, .armeb => try self.genArmBinOp(inst, bin_op.lhs, bin_op.rhs, .shl),
```
joachimschmidt557, line 1278: The code for `shl` is not present yet in `genArmBinOpCode`, which is called by `genArmBinOp`. This means that the code would hit the `unreachble` in that switch. It shouldn't be too hard to implement this, but I can also do that in a follow-up PR.

joachimschmidt557 (commented):
This is good progress on stage2!

test/stage2/cbe.zig
```diff
@@ -808,6 +808,29 @@ pub fn addCases(ctx: *TestContext) !void {
         });
     }
 
+    {
+        var case = ctx.exeUsingLlvmBackend("shift right + left", linux_x64);
```
Luukdegram, line 812: Shouldn't this be `ctx.exeCompiledFromC`?

src/codegen/llvm/bindings.zig
```diff
@@ -381,6 +389,15 @@ pub const Builder = opaque {
     pub const buildAnd = LLVMBuildAnd;
     extern fn LLVMBuildAnd(*const Builder, LHS: *const Value, RHS: *const Value, Name: [*:0]const u8) *const Value;
 
+    pub const buildLShr = LLVMBuildLShr;
+    extern fn LLVMBuildLShr(*const Builder, LHS: *const Value, RHS: *const Value, Name: [*:0]const u8) *const Value;
+
+    pub const buildAShr = LLVMBuildLShr;
```
Vexu, line 395:
```suggestion
    pub const buildAShr = LLVMBuildAShr;
```

g-w1: Thanks, unused decl errors will be so nice :)

src/codegen.zig
joachimschmidt557: This is more complex than anticipated, new psedo-instructions will have to be added to `codegen/arm.zig`. I'll do this in a follow-up PR.

andrewrk:
Don't forget to test with `-Denable-llvm`. It's not enabled on the CI (I should look into that soon).

```
shift right + left: execution exited with code 180:
qemu-x86_64 ./test_case 
compiler [53/118] shift right + left [2/2] test 'shift right + left' failed: ChildProcessExecution
```

Fixed with:

```diff
--- a/test/stage2/llvm.zig
+++ b/test/stage2/llvm.zig
@@ -32,18 +32,20 @@ pub fn addCases(ctx: *TestContext) !void {
         var case = ctx.exeUsingLlvmBackend("shift right + left", linux_x64);
 
         case.addCompareOutput(
-            \\pub export fn main() void {
+            \\pub export fn main() c_int {
             \\    var i: u32 = 16;
             \\    assert(i >> 1, 8);
+            \\    return 0;
             \\}
             \\fn assert(a: u32, b: u32) void {
             \\    if (a != b) unreachable;
             \\}
         , "");
         case.addCompareOutput(
-            \\pub export fn main() void {
+            \\pub export fn main() c_int {
             \\    var i: u32 = 16;
             \\    assert(i << 1, 32);
+            \\    return 0;
             \\}
             \\fn assert(a: u32, b: u32) void {
             \\    if (a != b) unreachable;
```

I'll push this as part of another commit

Issue #9308, opened 2021-07-05
std.build: support -mexec-model flag for WASI target


mathetake:
Now that `-mexec-model` option is introduced in https://github.com/ziglang/zig/pull/9178, I would like to set the option in build.zig but I realized that current std.build does not allow me to pass the flag via build.zig.

So I think we have two options to achieve this:
1. add the dedicated method in std.build to set the flag just like [-mcode-model flag](https://github.com/ziglang/zig/blob/master/lib/std/build.zig#L2537-L2540),
2. add more generic method in std.build for passing arbitrary flag to zig_args

before I work on a PR, I would like to hear maintainers' opinion here. Thank you!

PR #9598, opened 2021-08-21
build: allow specifying -mexec-model flag.

mathetake:
resolves https://github.com/ziglang/zig/issues/9308, though I'm not sure if this is the right way to achieve it. I need comments from @andrewrk @kubkon. Thank you!


Vexu (approved):
This looks correct to me.

PR #9599, opened 2021-08-21
Do not generate ref before elem_ptr and field_ptr

Snektron:
Previously AstGen would emit a Zir `ref` instruction before any `elem_ptr` and `field_ptr` instruction. This instruction is now omitted, and the relevant instructions in Sema are updated to account for this change.

`analyzeRef` is modified such that it now stores intermediates to a pointer, after which the pointer is casted to a constant version. This would otherwise fail the first check of `storePtr`. I've used `Sema.bitcast` here for now, though i guess in the future `Sema.coerce` would be more correct.

This should now more correctly handle taking addresses of values. Previously statements like `&x`, `&x.y` and `&x[0]` would fail because of the above problem.

Marked as wip because there are still some source locations that need to be fixed in the functions in Sema.

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -4854,7 +4854,7 @@ fn fieldAccess(
     const str_index = try astgen.identAsString(field_ident);
     switch (rl) {
         .ref => return gz.addPlNode(.field_ptr, node, Zir.Inst.Field{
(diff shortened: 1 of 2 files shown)


Snektron:
There does not seem to be an apparent improvement gained by removing `ref` here. Superseded by #9621.

Issue #7147, opened 2020-11-18
Allow `++` and `**` when _contents_ are not compile-time known


lucascharlesmyers:
As I understand it, `++` and `**` require that both operands are compile-time known, in their entirety. I would think `++` could be 
 extended to work when the contents of an operand are not `comptime`-known, as long as the lengths of both are `comptime`-known. The same would go for `**`, which should in theory be doable as long as the length of the left-hand-side is `comptime`-known, and the right-hand-side is `comptime`-known.

The primary use-cases I see for this are when these operators are extended to tuples (along with tuple slicing, #4625). Being able to concatenate and slice tuples of runtime-known contents is very useful in combination with any functions that take tuples as emulated var-args, especially `@call`.

mrakh:
The problem with using `++` or `**` for this is that it hides what is potentially a runtime-expensive memory copy. That kind of opaqueness goes against the Zen of Zig. I would use a dedicated built-in for this operation instead, with a name communicating the potential cost explicitly - perhaps something like `@concatenate` or `@concatArrays`?

SpexGuy:
> it hides what is potentially a runtime-expensive memory copy

This isn't unprecedented, assignment with `=` does the same thing when the object being assigned is a large struct or array value.  In general, the zig ideals don't necessarily say that operators need to be cheap, they just require that operators can't run user code or alter control flow.  I think this is reasonable to implement.

gwenzek:
From  [Zig weekly meeting notes](https://docs.google.com/document/d/1GFQxcoXgzmOFThjFLpjkVASopa33el2f5esvk_OZlCo/edit#heading=h.qi1ybyf2yvo2):

> As long as the lengths are comptime known, the programmer’s intent is clear here, and this seems like it should work.  The argument is raised in the issue that these operators would be expensive with large arrays, but so is assignment with large structs.  That doesn’t seem to be a strong reason to avoid this, and the language is more complicated to use if these operators don’t work in places where they could.

andrewrk:
This is implemented in self-hosted, which is now the default compiler.

PR #9603, opened 2021-08-22
stage2: add array concatenation + multiplication ( ++ and **) at comptime

g-w1:


src/Sema.zig
```diff
@@ -5512,8 +5512,108 @@ fn zirArrayCat(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileEr
...
+    const lhs = sema.resolveInst(extra.lhs);
+    const rhs = sema.resolveInst(extra.rhs);
+    const lhs_ty = sema.typeOf(lhs);
+    const rhs_ty = sema.typeOf(rhs);
+    const lhs_src: LazySrcLoc = .{ .node_offset_bin_lhs = inst_data.src_node };
+    const rhs_src: LazySrcLoc = .{ .node_offset_bin_rhs = inst_data.src_node };
+
+    const lhs_cat_info = getArrayCatInfo(lhs_ty);
+    if (lhs_cat_info == null)
+        return sema.mod.fail(&block.base, lhs_src, "expected array, found '{}'", .{lhs_ty});
```
Vexu, line 5526: Seems like this could be cleaned up with `orelse`.

src/Sema.zig
```diff
@@ -5512,8 +5512,108 @@ fn zirArrayCat(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileEr
...
+                    ty,
+                    val,
+                ));
+            }
+            return sema.mod.fail(&block.base, lhs_src, "TODO array_cat more types of Values", .{});
+        } else {
+            return sema.mod.fail(&block.base, rhs_src, "unable to evaluate constant expression", .{});
+        }
+    } else {
+        return sema.mod.fail(&block.base, lhs_src, "unable to evaluate constant expression", .{});
```
Vexu, line 5588: This should not be an error according to #7147.

src/Sema.zig
```diff
@@ -5512,16 +5512,156 @@ fn zirArrayCat(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileEr
...
+
+    // In `**` rhs has to be comptime-known, but lhs can be runtime-known
+    const tomulby = try sema.resolveInt(block, rhs_src, extra.rhs, Type.initTag(.usize));
+    const mulinfo = getArrayCatInfo(lhs_ty) orelse
+        return sema.mod.fail(&block.base, lhs_src, "expected array, found '{}'", .{lhs_ty});
+
+    const final_len = std.math.mul(u64, mulinfo.len, tomulby) catch return sema.mod.fail(&block.base, rhs_src, "operation results in overflow", .{});
+    if (try sema.resolveDefinedValue(block, lhs_src, lhs)) |lhs_val| {
+        if (lhs_val.tag() == .empty_array) return lhs;
+        if (lhs_val.tag() == .decl_ref) {
```
andrewrk, line 5635: Put the castTag here :)

src/Sema.zig
```diff
@@ -5512,16 +5512,156 @@ fn zirArrayCat(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileEr
...
+    if (rhs_info.sent != null and lhs_info.sent != null) {
+        if (rhs_info.sent.?.eql(lhs_info.sent.?, lhs_info.array_elem_type)) {
+            res_sent = lhs_info.sent.?;
+        }
+    }
+
+    if (try sema.resolveDefinedValue(block, lhs_src, lhs)) |lhs_val| {
+        if (try sema.resolveDefinedValue(block, rhs_src, rhs)) |rhs_val| {
+            const final_len = lhs_info.len + rhs_info.len;
+            if (rhs_val.tag() == .empty_array) return lhs;
```
andrewrk, line 5544: Instead of checking the tag() here, this code should go through the Value helper methods. What this code is effectively doing is 1. Checking if the type is a pointer and 2. Dereferencing the pointer value to its element value. 1 should be done based on the Type not the Value and 2 can be done via the pointerDeref method of Value.

andrewrk (changes requested):
(meant to make that comment a review)

src/Sema.zig
```diff
@@ -5512,16 +5512,156 @@ fn zirArrayCat(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileEr
...
+            const lhs_sub_val = lhs_sub.data.val;
+            var anon_decl = try block.startAnonDecl();
+            defer anon_decl.deinit();
+
+            const final_ty = if (mulinfo.sent) |sent|
+                try Type.Tag.array_sentinel.create(anon_decl.arena(), .{ .len = final_len, .elem_type = mulinfo.array_elem_type, .sentinel = sent })
+            else
+                try Type.Tag.array.create(anon_decl.arena(), .{ .len = final_len, .elem_type = mulinfo.array_elem_type });
+
+            if (lhs_sub_val.tag() == .empty_array) return rhs;
```
miscco, line 5645: Should that be moved right after the declaration of ´lhs_sub_val´, you are neither using `anon_decl` nor `final_ty` here

src/Sema.zig
```diff
@@ -5512,16 +5512,156 @@ fn zirArrayCat(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileEr
...
+            info.array_elem_type = Type.initTag(.u8);
+            info.sent = Value.initTag(.zero);
+        } else if (t.tag() == .array) {
+            const payload = t.castTag(.array).?.data;
+            info.array_elem_type = payload.elem_type;
+        } else if (t.tag() == .array_sentinel) {
+            const payload = t.castTag(.array_sentinel).?.data;
+            info.array_elem_type = payload.elem_type;
+            info.sent = payload.sentinel;
+        }
```
miscco, line 5605: Should there be an error case here, similar to the case where it is neither an array nor a pointer?
```suggestion
        } else return null;
```

src/Sema.zig
```diff
@@ -5512,16 +5512,156 @@ fn zirArrayCat(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileEr
...
+    }
+
+    if (try sema.resolveDefinedValue(block, lhs_src, lhs)) |lhs_val| {
+        if (try sema.resolveDefinedValue(block, rhs_src, rhs)) |rhs_val| {
+            const final_len = lhs_info.len + rhs_info.len;
+            if (rhs_val.tag() == .empty_array) return lhs;
+            if (lhs_val.tag() == .empty_array) return rhs;
+            if (rhs_val.tag() == .decl_ref and lhs_val.tag() == .decl_ref) {
+                var anon_decl = try block.startAnonDecl();
```
miscco, line 5547: could that be moved right before `const buf = ...`?

src/Sema.zig
```diff
@@ -5512,16 +5512,156 @@ fn zirArrayCat(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileEr
...
+    const lhs_ty = sema.typeOf(lhs);
+    const lhs_src: LazySrcLoc = .{ .node_offset_bin_lhs = inst_data.src_node };
+    const rhs_src: LazySrcLoc = .{ .node_offset_bin_rhs = inst_data.src_node };
+
+    // In `**` rhs has to be comptime-known, but lhs can be runtime-known
+    const tomulby = try sema.resolveInt(block, rhs_src, extra.rhs, Type.initTag(.usize));
+    const mulinfo = getArrayCatInfo(lhs_ty) orelse
+        return sema.mod.fail(&block.base, lhs_src, "expected array, found '{}'", .{lhs_ty});
+
+    const final_len = std.math.mul(u64, mulinfo.len, tomulby) catch return sema.mod.fail(&block.base, rhs_src, "operation results in overflow", .{});
```
miscco, line 5632: Speaking from painful experience working on a compiler with first class multidimensional array support, this is something you should try to avoid.

This essentially just blows up your AST without any benefit. You will perform all following semantic analysis / transformations that come afterwards essentially `tomulby` times.

If there is great caching of arbitrary transformations in place great, but in general you would really want to have a special purpose ZIR / AIR node that es essentially a binop between a value and a dimension.

That considerably flattens the AST at the cost that you have a second array type that you need to consider.

andrewrk: Thanks for the advice @miscco. This code has nothing to do with the AST but I think your point still stands. Even so, this expansion code is needed, since comptime code could, for example, modify the array at an arbitrary index. So what you are describing would be a performance optimization to make the compiler more lazy. That's a good idea, but it can come later.

miscco: AST was essentially a shorthand for a tree based representation of the program (even if it is stored in flat memory)

It is essentially an optimization and fully depends on whether it becomes an actual problem. just something to keep in the back of your head when expanding a single node into many

src/Sema.zig
g-w1: The correct thing here is else unreachable; since no other type tags will say the zigTypeTag is .Array.

src/Sema.zig
g-w1: Yeah, I have in mind some optimizations for this, but I will put them in a separate pr and benchmark them first.

src/Sema.zig
g-w1: I put the anon_decl there to signify that the whole block was using it. I don't really see a reason to move things around pointlessly.

src/Sema.zig
g-w1: I don't see why, I put it there so eventually, it could be moved out of the block as more things could use it when arraycat works at runtime also.

src/Sema.zig
```diff
@@ -5512,16 +5512,156 @@ fn zirArrayCat(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileEr
...
+        return sema.mod.fail(&block.base, lhs_src, "TODO runtime array_cat", .{});
+    }
+}
+
+const ArrayCatInfo = struct { array_elem_type: Type, sent: ?Value = null, len: u64 };
+fn getArrayCatInfo(t: Type) ?ArrayCatInfo {
+    var info: ArrayCatInfo = .{ .array_elem_type = undefined, .len = undefined };
+    if (t.zigTypeTag() == .Array) {
+        info.len = t.arrayLen();
+        if (t.tag() == .array_u8) {
```
andrewrk, line 5593: don't use `Type.tag()` or `Value.tag()` in Sema

andrewrk (changes requested):

src/codegen/llvm.zig
```diff
@@ -796,11 +796,14 @@ pub const DeclGen = struct {
                     const gpa = self.gpa;
                     const elem_ty = tv.ty.elemType();
                     const elem_vals = payload.data;
-                    const llvm_elems = try gpa.alloc(*const llvm.Value, elem_vals.len);
+                    const sento: ?Value =
+                        if (tv.ty.castTag(.array_sentinel)) |sent_pl| sent_pl.data.sentinel else null;
```
andrewrk, line 800: get the sentinel value from the type system, not the tag()

andrewrk (changes requested):

andrewrk:
Nice work, and thanks for all the follow-ups :+1:

PR #9604, opened 2021-08-22
Add std.debug.println

jdmichaud:
Add a `println` shortcut to `std.debug`.

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -73,6 +73,10 @@ pub fn print(comptime fmt: []const u8, args: anytype) void {
     nosuspend stderr.print(fmt, args) catch return;
 }
 
+pub fn println(comptime fmt: []const u8, args: anytype) void {
+    print(fmt ++ "\n", args);
+}
+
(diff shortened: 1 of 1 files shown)


andrewrk:
Thanks for the PR.

Adding `\n` into the format string is the same or less effort as calling a different function. So, in favor of one canonical way of doing things, no println shortcut function in the std lib.

marler8997:
I think other languages usually do this to make line endings more "platform normal" because of Windows, i.e.

```zig
pub fn println(comptime fmt: []const u8, args: anytype) void {
    print(fmt ++ (if (builtin.os.tag == .windows) "\r\n" else "\n"), args);
}
```

I think the counter argument for this would be that the number of applications that actually need to print with `\r\n` line endings on windows is probably small, and in those cases they could easily add this function to their own project themselves.

jedisct1:
Don't we have a constant defined to the typical platform-specific line ending convention?

jdmichaud:
@jedisct1 Apparently [yes](https://github.com/ziglang/zig/blob/master/lib/std/cstr.zig#L12).

PR #9605, opened 2021-08-22
wasm: pass --export-dynamic to wasm-ld for WASI reactors.

mathetake:
[WASI reactor](https://github.com/WebAssembly/WASI/blob/main/design/application-abi.md#current-unstable-abi) should export not only "_initialize" but also other functions so that runtimes can use these functions as their own entrypoints.

cc @kubkon


jedisct1:
Thank you @mathetake !

PR #9606, opened 2021-08-22
Fix typo introduced in 50a29f7

sylt:
Now the last argument to select() will be checked as well

 - https://github.com/ziglang/zig/pull/9459

CC: @Snektron


andrewrk:
Thanks!

PR #9607, opened 2021-08-22
std.hash_map: add getKey methods

FnControlOption:
MachO uses a string table where each string index is duplicated as both the key and the value. Adding `fetchKeyAdapted` makes it easier to retrieve the string index without needing duplication.

I originally named this method as `getKeyAdapted` but I thought that it could be confused as a method to create keys, rather than getting a matching key if present.

### DIFF
--- a/lib/std/array_hash_map.zig
+++ b/lib/std/array_hash_map.zig
@@ -298,6 +298,22 @@ pub fn ArrayHashMap(
             return self.unmanaged.getPtrAdapted(key, ctx);
         }
 
+        /// Find the actual key associated with an adapted key
+        pub fn getKey(self: Self, key: K) ?K {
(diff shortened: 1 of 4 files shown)


SpexGuy (changes requested):
Thanks! The implementation looks good.

I appreciate the thought re: getKey* being confusing, but I think getKey would still be a more consistent name.  The functions for retrieving values are also called get*, and functions which may create new entries are named getOrPut*, so I don't think it's too confusing.  `fetch` is always associated with removal in the current api, so that might cause additional confusion.

Please also add the getKey(Ptr)Context and getKey(Ptr) wrapper functions.

SpexGuy (approved):

SpexGuy:
Looks great, thanks!

PR #9611, opened 2021-08-23
Update comment: s/var/anytype/

McSinyx:
I took a glance through `rg '[^\s]\bvar\b[^\s]'` and it seems that this is the only one left.


jedisct1:
Thank you!

PR #9612, opened 2021-08-23
Don't define valgrind_support on macOS

jedisct1:
Unfortunately, Valgrind for macOS has been broken for years, and the Homebrew formula is only for Linux.


andrewrk (changes requested):
Here's the place to change it in self-hosted: https://github.com/ziglang/zig/blob/d5ef5da5947d24792933847454c84daf1282b29e/src/target.zig#L166

jedisct1:
Done for self-hosted, too. Thanks!

andrewrk (approved):

PR #4233, opened 2020-01-19
Turn errno into an enum

daurnimator:
Currently compiler is dieing with a segfault when I test locally.

PR #9613, opened 2021-08-24
std: move errno to become an nonexhaustive enum

andrewrk:
The primary purpose of this change is to eliminate one usage of
`usingnamespace` in the standard library - specifically the usage for
errno values in `std.os.linux`.

This is accomplished by truncating the `E` prefix from error values, and
making errno a proper enum.

A similar strategy can be used to eliminate some other `usingnamespace`
sites in the std lib.


daurnimator:
See also #4233

PR #9616, opened 2021-08-24
error_set_single -> anyerror coercion, comptime zirErrorName

SuperAuguste:
My first stage2 PR!

Implements coercion from `error.Joe` -> `anytype`, which then allowed me to implement `@errorName`!

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -6876,9 +6876,28 @@ fn zirEmbedFile(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileE
 }
 
 fn zirErrorName(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileError!Air.Inst.Ref {
+    const tracy = trace(@src());
+    defer tracy.end();
+
     const inst_data = sema.code.instructions.items(.data)[inst].un_node;
     const src = inst_data.src();
-    return sema.mod.fail(&block.base, src, "TODO: Sema.zirErrorName", .{});
+    _ = src;
+    const operand_src: LazySrcLoc = .{ .node_offset_builtin_call_arg0 = inst_data.src_node };
+    const op = sema.resolveInst(inst_data.operand);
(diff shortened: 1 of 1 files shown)


src/Sema.zig
```diff
@@ -8571,6 +8590,14 @@ fn coerce(
             // T to E!T or E to E!T
             return sema.wrapErrorUnion(block, dest_type, inst, inst_src);
         },
+        .ErrorSet => {
+            // error.Joe to anyerror
+            if (dest_type.tag() == .anyerror and inst_ty.tag() == .error_set_single) {
```
andrewrk, line 8595: `tag()` is the wrong abstraction level for Sema to be using here. Rather than checking how the type is *represented* which is what `tag()` provides, this code needs to ask the Type whether it is the global error set or not.

For checking `inst_ty`, use `zigTypeTag()` to make sure it is an error set, rather than only checking if it is represented as `error_set_single` - because error set values could be represented with other tag()s.

andrewrk (changes requested):
Test coverage please!

src/Sema.zig
```diff
@@ -6876,9 +6876,28 @@ fn zirEmbedFile(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileE
...
+    _ = src;
+    const operand_src: LazySrcLoc = .{ .node_offset_builtin_call_arg0 = inst_data.src_node };
+    const op = sema.resolveInst(inst_data.operand);
+    const op_coerced = try sema.coerce(block, Type.initTag(.anyerror), op, operand_src);
+
+    if (try sema.resolveDefinedValue(block, src, op_coerced)) |val| {
+        var anon = try block.startAnonDecl();
+        errdefer anon.deinit();
+
+        const bytes = try anon.new_decl_arena.allocator.dupe(u8, val.castTag(.@"error").?.data.name);
```
g-w1, line 6893: `s/.new_decl_arena.allocator/arena()/`

PR #9620, opened 2021-08-25
Recursively append include dirs of library dependencies

S0urc3C0de:
This commit makes the build system append the include directories of libraries that the LibExeObjStep is being linked against recursively.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -2666,7 +2666,7 @@ pub const LibExeObjStep = struct {
             try self.makePackageCmd(pkg, &zig_args);
         }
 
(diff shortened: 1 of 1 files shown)


andrewrk (changes requested):
* Needs test
* Needs to account for a recursive loop

andrewrk:
Thanks for the patch. I am sorry I did not get around to reviewing it for many months.

I am closing this because it is not merge-ready and it is very old. However, you are welcome to open a new PR if you wish to continue this effort. I think you should first start with an issue to track the use case this is trying to address. This way if the PR is not merged (as in this one) the issue will persist.

PR #9623, opened 2021-08-25
Allow packages to access `build_options`

### LINKED ISSUE #5375: `addBuildOption` results in `error: unable to find 'build_options'`
When setting a build option in `build.zig` via `exe.addBuildOption(bool, "the_option", true)` the package isnt available from code in the `exe` it was added to.

Adding this to a file in the exe:
```const build_options = @import("build_options");```
Result in this error:
```error: unable to find 'build_options'```

The build command seems to properly contain the pkg definition:
```
zig build-exe /Users/mikedesaro/Zig-Aya/examples/thing.zig --library c++ --library c --library SDL2 --pkg-begin build_options SNIPPED/zig-cache/run_build_options.zig --pkg-end SNIPPED
```

### PR DESCRIPTION
**Pull request by leecannon**
I currently getting around this limitation by using an ugly hack
https://github.com/leecannon/zig-libgit2/blob/d82d91ad4e8aee6247d165a45e87231785301e9f/build.zig#L54-L62

this will fix #5375

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -2244,16 +2244,23 @@ pub const LibExeObjStep = struct {
         self.include_dirs.append(.{ .other_step = other }) catch unreachable;
     }
 
-    fn makePackageCmd(self: *LibExeObjStep, pkg: Pkg, zig_args: *ArrayList([]const u8)) error{OutOfMemory}!void {
+    fn makePackageCmd(self: *LibExeObjStep, pkg: Pkg, zig_args: *ArrayList([]const u8), build_options_path: ?[]const u8) error{OutOfMemory}!void {
         const builder = self.builder;
 
         try zig_args.append("--pkg-begin");
         try zig_args.append(pkg.name);
         try zig_args.append(builder.pathFromRoot(pkg.path.getPath(self.builder)));
 
+        if (build_options_path) |build_options| {
+            try zig_args.append("--pkg-begin");
+            try zig_args.append("build_options");
+            try zig_args.append(build_options);
+            try zig_args.append("--pkg-end");
+        }
+
         if (pkg.dependencies) |dependencies| {
             for (dependencies) |sub_pkg| {
-                try self.makePackageCmd(sub_pkg, zig_args);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by MasterQ32** [comment]
Imho this is a anti-pattern and `build_options` should be removed from the `LibExeObjStep` completly.

The nicer solution here is to create a `std.build.BuildOptions` which provides a `std.build.FileSource` that you can pull into any package like this:

```zig
pub fn build(b: *Builder) !void {
    const build_options = b.createBuildOptions();
    build_options.add([]const u8, string, "This is a string");
    
    const git_pkg = std.build.Pkg{
        .name = "git",
        .path = .{ .path = prefix_path ++ "src/git.zig" },
        .dependencies = &[_]std.build.Pkg{
            .{ .name = "build_options", .path = build_options.getSource() },
        },
    };
}
```

With this, you can make as many `build_options` packages as you want, share them between packages or even whole compilations, and you are not dependent on hacks or the enforcement of all packages sharing `build_options` (which would not work as soon as two build options conflict)

leecannon:
I've mostly implemented the suggested change, however ive run into something im not 100% happy with.

Status quo outputs build options to a file called `{executable_name}_build_options.zig`, decoupling build options from the LibExeObjStep means that the caller must pass a name i.e. `const build_options = b.addBuildOptions("build_options");`

I will push the changes soon for review.

MasterQ32:
> I've mostly implemented the suggested change, however ive run into something im not 100% happy with.
> 
> Status quo outputs build options to a file called `{executable_name}_build_options.zig`, decoupling build options from the LibExeObjStep means that the caller must pass a name i.e. `const build_options = b.addBuildOptions("build_options");`
> 
> I will push the changes soon for review.

Just don't care. Output the build options into `zig-cache/build_options/${shasum}/build_options.zig` where shasum is the sum over your file content (similar to how the zig-cache does it).

This allows you to have guaranteed non-conflicting build option files

leecannon:
While this is an improvement in terms of flexibility, using the newer version of the API in the use case I needed this for has made the code much more messy. So much so that I don't think this is worth a breaking change.

PR #9634, opened 2021-08-27
stage1: condense compile error notes to make traces easier to read

nektro:
image


g-w1:
I think you are targeting a valid problem here, but with the wrong solution. Instead of hiding the notes, you (imo) should make less notes.

nektro:
Fair, but I'm not sure less notes would be helpful either. The notes in some cases are critical for actually finding the place where a bug is (ie failed assertion)

andrewrk:
"referenced here" removed entirely in #10648

Issue #2826, opened 2019-07-05
build-exe: ability to output raw binary or hex format rather than ELF


andrewrk:
Sometimes it is desirable to get the raw machine code out of an ELF and use that. This example depends on objcopy to do that, but it should be doable with only zig:

https://github.com/andrewrk/clashos/blob/3f786d4294c71f909b1ecaf3c8d0e247b90c5c86/build.zig#L32-L37

I think this would probably be more options to `--emit`. Current options are:
 * `llvm-ir`
 * `bin` (ELF or PE)
 * `asm`

Unfortunately, `bin` is a bit ambiguous here. I think it would be OK to add:
 * `ihex` (equivalent to `objcopy -O ihex`)
 * `raw` (eqivalent to `objcopy -O binary`)

We could leave `bin` the same, or potentially try to come up with a less ambiguous name.

As for how to implement it - LLVM in theory should have a way to specify this, since it's actually less work to generate raw machine code than to wrap it in an ELF. But worst case scenario, zig could do the processing after the fact.

MasterQ32:
I really like the idea of zig to directly output raw oder intel hex formats!

What would also be favourable: To output PE under linux and ELF under windows to cross-compile for different platforms.

This could finally allow to have a build system that doesn't suck at "one project, multiple platforms" without a hassle.

zimmi:
>What would also be favourable: To output PE under linux and ELF under windows to cross-compile for different platforms.

That's already possible, see [here](https://ziglang.org/#Cross-compiling-is-a-first-class-use-case).
I just tried running this `build.zig` on my linux machine and it produced a `.exe` for windows:

```zig
const Builder = @import("std").build.Builder;
const builtin = @import("builtin");

pub fn build(b: *Builder) void {
    const mode = b.standardReleaseOptions();
    const exe = b.addExecutable("my_exe", "src/main.zig");
    exe.setBuildMode(mode);
    exe.setTarget(
        builtin.Arch.x86_64,
        builtin.Os.windows,
        builtin.Abi.msvc,
    );
    exe.install();
}
```

suirad:
This would also be desirable to be able to write shellcode with zig.

mlarouche:
I will look into that issue soon

daurnimator:
> this would probably be more options to `--emit`

How should this co-exist with https://github.com/ziglang/zig/issues/2279#issuecomment-571802541 ?
> --emit is deprecated;

markfirmware:
[...]

PR #9636, opened 2021-08-27
stdlib: Add Intel HEX support to InstallRawStep

ehaas:
This allows writing HEX files with `exe.installRaw`, where `exe` is a `LibExeObjStep`. A HEX file will be written if the file extension is `.hex` or `.ihex`, otherwise a binfile will be written. The output format can be explicitly chosen with `exe.installRawWithFormat("filename", .hex);` (or `.bin`)

Part of #2826


ehaas:
Drone failed on `Test [193/1032] behavior.atomics.test "behavior-native-ReleaseFast-bare-single atomicrmw with floats"... FAIL (TestUnexpectedResult)` which seems to be unrelated.

lib/std/build/InstallRawStep.zig
```diff
@@ -13,6 +13,7 @@ const fs = std.fs;
 const io = std.io;
 const sort = std.sort;
 const warn = std.debug.warn;
+const nl = std.cstr.line_sep;
```
andrewrk, line 16: shouldn't the newline depend on the target, not the host system running the compiler?

andrewrk (commented):
Thanks for this! Looks merge-ready to me except for 1 question that I had-

lib/std/build/InstallRawStep.zig
ehaas: Upon further investigation it appears to always end in CRLF. Source: https://developer.arm.com/documentation/ka003292/latest and also running `objcopy -O ihex test.o test.hex` on a native .o file for a x86_64 linux system. So I'll update it to always to `\r\n`

PR #9637, opened 2021-08-28
std: add json.stringifyAlloc

nektro:
Previously only `std.json.stringify` was available which accepts a Writer. This PR adds an alternative that accepts an Allocator for a shorthand to return the stringification results in dynamically allocated memory.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -3333,3 +3333,23 @@ test "stringify null optional fields" {
         .{ .allocator = std.testing.allocator },
     ));
 }
+
+// Same as `stringify` but accepts an Allocator and stores result in dynamically allocated memory instead of using a Writer.
+// Caller owns returned memory.
+pub fn stringifyAlloc(allocator: std.mem.Allocator, value: anytype, options: StringifyOptions) ![]const u8 {
+    var list = std.ArrayList(u8).init(allocator);
+    errdefer list.deinit();
+    try stringify(value, options, list.writer());
+    return list.toOwnedSlice();
+}
+
+test "stringify alloc" {
+    const allocator = std.testing.allocator;
+    const expected =
+        \\{"foo":"bar","answer":42,"my_friend":"sammy"}
+    ;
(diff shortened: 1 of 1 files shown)


lib/std/json.zig
```diff
@@ -3257,3 +3257,22 @@ test "stringify struct with custom stringifier" {
 test "stringify vector" {
     try teststringify("[1,1]", @splat(2, @as(u32, 1)), StringifyOptions{});
 }
+
+// Same as `stringify` but accepts an Allocator and stores result in dynamically allocated memory instead of using a Writer.
+// Caller owns returned memory.
+pub fn stringifyAlloc(allocator: *std.mem.Allocator, options: StringifyOptions, value: anytype) ![]const u8 {
+    var list = std.ArrayList(u8).init(allocator);
```
daurnimator, line 3264: Does this need an `errdefer list.deinit()`?
e.g. if the 2nd reallocation/growing fails?

nektro: good call +1

lib/std/json.zig
```diff
@@ -3257,3 +3257,23 @@ test "stringify struct with custom stringifier" {
 test "stringify vector" {
     try teststringify("[1,1]", @splat(2, @as(u32, 1)), StringifyOptions{});
 }
+
+// Same as `stringify` but accepts an Allocator and stores result in dynamically allocated memory instead of using a Writer.
+// Caller owns returned memory.
+pub fn stringifyAlloc(allocator: *std.mem.Allocator, options: StringifyOptions, value: anytype) ![]const u8 {
```
daurnimator, line 3263: to match `stringify`, should `options` come after `value`?

nektro: I switched the order because if someone is passing the values inline it is significantly more likely that the `value` parameter is going to be spread amongst multiple lines.

kubkon:
@nektro could you rebase and fix conflicts so we could merge this?

nektro:
will do

lib/std/json.zig
nektro: fixed

PR #9639, opened 2021-08-28
zig fmt: fold nested returns

### LINKED ISSUE #9630: Return statements allow an arbitrary number of `return` keywords
This happens in stage1 and stage2, and `zig fmt` also does not complain. The behavior appears to be the same as only using a single `return` keyword, so I think it's just a quirk of the compiler and not a footgun:

```zig
fn is_even(x: u8) bool {
    return return return return return x % 2 == 0;
}

pub fn main() u8 {
    var x: u8 = 2;
    return return if (is_even(x)) 0 else 1;
}
```

jdmichaud:
Fixes #9630.

### DIFF
--- a/lib/std/zig/parser_test.zig
+++ b/lib/std/zig/parser_test.zig
@@ -2711,6 +2711,28 @@ test "zig fmt: return" {
     );
 }
 
+test "zig fmt: nested return" {
+    try testTransform(
+        \\fn foo(x: u8) bool {
+        \\    return return return return return x % 2 == 0;
+        \\}
+        \\
+        \\fn bar() void {
+        \\    return return return;
+        \\}
+        \\
+    ,
+        \\fn foo(x: u8) bool {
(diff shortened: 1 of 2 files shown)


ifreund (changes requested):
This looks good, but it needs a matching test case in `lib/std/zig/parser_test.zig`

andrewrk (dismissed):

andrewrk:
@jdmichaud thank you for this pull request. It looks good and it's ready to be merged. However, this got @ifreund and I thinking, and we realized that actually there is a missing check in AstGen for dead code (in quite a few places; not just `return`). So instead of having `zig fmt` canonicalize it, it's actually going to be come an "unreachable code" error. 

I know you were following some guidance to go down this path and make this pull request, so I apologize for giving the wrong suggestion. Sometimes we don't figure these things out until we've already done it the wrong way :sweat_smile:

PR #9642, opened 2021-08-28
Fix float formatting for 0.0 when precision is 0

BanchouBoo:
With precision 0 (`{d:.0}`), floats will format 0.0 as `0.0` instead of 0, but any other number will completely cut off the decimal point, this corrects 0.0 to also cut off the decimal point.


nektro (changes requested):
Forgot to mention this on discord, but can you also add to a test block and assert using `{d:.0}` with `0` results in `0` ?

BanchouBoo:
Done

Issue #3257, opened 2019-09-19
make `noreturn` type more like the "zero" type


vi:
* `?noreturn` is not allowed instead of being `void`-equivalent (always `null`). Like `u0` is currently `void`-equivalent.
* Empty unions are disallowed instead of being `noreturn`-equivalent.
* [unreachable code error is too strict](https://github.com/ziglang/zig/issues/3256), preventing instantiating generic code with `T=noreturn`
* Using `noreturn` for union variant does still allocate a tag for such variant

---

Java gets [unit](https://en.wikipedia.org/wiki/Unit_type) and [zero](https://en.wikipedia.org/wiki/Bottom_type) types wrong. Rust gets unit and zero types (relatively) right. Zig gets unit type right, but zero type wrong. Shall it be fixed?

andrewrk:
Another use for this: `anyerror!noreturn` would be the return type of `std.os.execve`. When there is an explicit type, it could be used directly, e.g. `anyerror`, but when one wants to infer the type, `!noreturn` makes sense.

daurnimator:
> Empty unions are disallowed instead of being `noreturn`-equivalent.

I feel like this is the "right thing". @vi do you have an example use case for this one?

vi:
@daurnimator , Obvious one is codegenerated Zig code where union variants are filled in by automatic algorithm.

Special cased error because of something happens to be empty leads to special cases and workarounds in code generator (i.e. removing the type altogether or adding a dummy variant).

If empty unions definitions and (also empty) switch statements on those unions are allowed, less corner cases need to be considered in code generator.

vi:
@daurnimator 

> I feel like this is the "right thing"

Tagged union of 2 bools (`union(enum) { var1: bool, var2: bool,}`) has total 4 possible states. Remove `var2` - now 2 possible states. Also remove `var1` - now 0 possible states. Just like `noreturn`. So allowing it and making it like `noreturn` (at least with respect to code generation) seems like the right thing. Other `noreturn` features (like convertability to any other type) are probably not good although.

tadeokondrak:
related:
https://github.com/rust-lang/rfcs/blob/master/text/1216-bang-type.md
https://github.com/apple/swift-evolution/blob/master/proposals/0102-noreturn-bottom-type.md

PR #9645, opened 2021-08-29
Astgen: more unreachable checks

### LINKED ISSUE #9630: Return statements allow an arbitrary number of `return` keywords

nektro:
Also fixes #9630

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -996,7 +996,7 @@ fn suspendExpr(
     suspend_scope.suspend_node = node;
     defer suspend_scope.instructions.deinit(gpa);
 
-    const body_result = try expr(&suspend_scope, &suspend_scope.base, .none, body_node);
+    const body_result = try exprMaybeNoReturn(&suspend_scope, &suspend_scope.base, .none, body_node);
     if (!gz.refIsNoReturn(body_result)) {
         _ = try suspend_scope.addBreak(.break_inline, suspend_inst, .void_value);
     }
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -273,19 +273,20 @@ fn typeExpr(gz: *GenZir, scope: *Scope, type_node: ast.Node.Index) InnerError!Zi
     return expr(gz, scope, coerced_type_rl, type_node);
 }
 
-/// Same as `expr` but fails with a compile error if the result type is `noreturn`.
-fn reachableExpr(
+/// Turn Zig AST into untyped ZIR istructions.
+/// When `rl` is discard, ptr, inferred_ptr, or inferred_ptr, the
+/// result instruction can be used to inspect whether it is isNoReturn() but that is it,
+/// it must otherwise not be used. Fails with a compile error if the result type
+// is `noreturn`. Otherwise, use `exprMaybeNoReturn`.
+fn expr(
     gz: *GenZir,
     scope: *Scope,
     rl: ResultLoc,
     node: ast.Node.Index,
-    src_node: ast.Node.Index,
 ) InnerError!Zir.Inst.Ref {
-    const result_inst = try expr(gz, scope, rl, node);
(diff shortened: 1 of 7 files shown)


src/AstGen.zig
```diff
@@ -8398,6 +8403,22 @@ fn rvalue(
...
+        return gz.astgen.failNode(src_node, "unreachable code", .{});
+    }
+    return rvalueMaybeUnreachable(gz, rl, result, src_node);
+}
+
+/// Applies `rl` semantics to `inst`. Expressions which do not do their own handling of
+/// result locations must call this function on their result.
+/// As an example, if the `ResultLoc` is `ptr`, it will write the result to the pointer.
+/// If the `ResultLoc` is `ty`, it will coerce the result to the type.
+fn rvalueMaybeUnreachable(
```
ifreund, line 8417: Why is this split needed? It should be documented why this variant needs to exist and when to use it in the doc comment instead of duplicating the information already in the doc comment of `rvalue()`

src/AstGen.zig
```diff
@@ -8398,6 +8403,22 @@ fn rvalue(
     rl: ResultLoc,
     result: Zir.Inst.Ref,
     src_node: ast.Node.Index,
+) InnerError!Zir.Inst.Ref {
+    if (gz.refIsNoReturn(result)) {
```
ifreund, line 8407: This function is being removed, to be replaced with `GenZir.endsWithNoReturn`.

ifreund (commented):
Why have only some uses of `expr()` been replaced with `reachableExpr()`? It seems to me that *every* usage of `expr()` needs to check if the `expr()` is noreturn. I wonder if renaming `expr()` to `exprMaybeNoReturn()` and `reachableExpr()` to `expr()` would be the best way forward.

Note that the current `reachableExpr()` also should be updated to use `GenZir.endsWithNoReturn()`.

ifreund:
On further investigation, I think the only place where we wouldn't be able to use the semantics of the current `reachableExpr()` would be in `unusedResultExpr()`.

ifreund:
Though now I've remember about #3257 and wonder if we're on the totally wrong path here...

src/AstGen.zig
nektro: https://github.com/ziglang/zig/pull/9645/files/b0f73a6847e12373ec938c0570e3b9fb93c8df28#diff-3ed5c754ecaac0e7ed512aafc83f6cb0acb474fc267eff3efa775e1afdad159dR7529-R7534

ifreund:
Note: if we apply these checks wherever technically possible this would make the following pattern fail to compile:
```zig
pub const ensureCapacity = @compileError("deprecated for ensureUnusedCapacity and ...");
```
however, this could be worked around by wrapping the `@compileError()` call in a function or block:
```zig
pub const ensureCapacity = {
    @compileError("deprecated...");
};
// or
pub fn ensureCapacity(self: *Self, new_capacity: usize) !void {
    @compileError("deprecated...");
}
```
Applying these checks wherever possible would certainly be the most consistent approach IMO, and I don't find the workarounds too unpalatable. This feels quite arbitrary though...

andrewrk (changes requested):

andrewrk:
If someone wants to pick this back up:

 * use the `reachableExpr` function that I have provided, for example in c42763f8cc9223fc43ac3478a5f15d4bd9f438e4
 * add test coverage for each one. yes, each one. again, example is in c42763f8cc9223fc43ac3478a5f15d4bd9f438e4

PR #9647, opened 2021-08-29
fix CI and new decl scan

nektro:


test/behavior/misc.zig
```diff
@@ -506,21 +506,23 @@ test "lazy typeInfo value as generic parameter" {
...
+                        const Self = @This();
+                    };
+                }
             };
         }
     };
-}
-test "non-ambiguous reference of shadowed decls" {
-    try expect(A().B().Self != A().Self);
+    try expect(S.A().B().Self != S.A().Self);
```
nektro: `self reference through fn ptr field` was the test initially triggering the ambiguous reference, and the right solution seemed like changing the parameter value to `S.A` but that triggered an `unknown identifer 'S'` in both my fork and master so I went with scoping this test instead.

Ref: https://github.com/ziglang/zig/blob/4bb5d17/test/behavior/misc.zig#L334-L348

PR #9649, opened 2021-08-29
Address Spaces

Snektron:
status:
- [x] Parsing `addrspace` on pointers, functions and variables.
- [x] Render address spaces in zig fmt
- [x] AstGen
    - [x] Add `addrspace` property to pointers/functions/variables in ZIR.
    - [x] Forbid `addrspace` on local variables
- [x] Sema 
    - [x] Update ZIR format to changes made in AstGen
    - [x] `decl_ref` pointers
    - [x] `*?T` => `*T`
    - [x] `*(!E|T)` => `*T`
    - [x] Pointer access chains (`&x[y]`, `&x.y`, `&x.?`, `&x[y].?[z..w].u`, etc)
    - [x] `x[y..z]` 
    - [x] Disallow coercing pointers between address spaces
    - [x] Check whether address space is supported by target
    - [x] Check whether address space is valid for variable
    - [x] Check whether address space is valid for function
- [x] Implement address space attribute in the LLVM backend
- [ ] Implement address space attribute in other backends
- [ ] Allow pointers for address spaces to be different widths
- [ ] Allow different values for `null` and `allowzero` depending on address space (see AMDGPU LLVM target)
- [x] Implement address space property in TypeInfo and make it work with `@Type`
- [ ] Implement address space support for `@export` and `@extern`
- [x] Implement different default address space depending on context
- [ ] Implement address spaces for targets:
    - [x] x86 / x86_64
    - [ ] AVR
    - [ ] AMDGPU
    - [ ] NVPTX
    - [ ] SPIRV

Implements #653


Snektron:
Currently, the default address space is `.generic` which is supposed to be supported (in some sense) by any target. A pointer pointing into the generic address space can be constructed either by `*addrspace(.generic) T` or by `*T`. While this does allow two different syntaxes for the same pointer, i believe that this helps with some generic programming and also mirrors how `callconv(.Unspecified)` is allowed.

Note that for some targets (like SPIR-V), there is also a specific address space called `generic`, which is not supported in the case of shaders. I don't think that's much of a problem though, and in that case the generic address space can simply be disallowed, and the user would be required to set the address space manually on all global variables.

Speaking of, some targets have different address spaces intended for different purposes. For example, on SPIR-V function-local variables are placed in a different address space than (instance specific) global variables. My plan was to infer the assigned address space from context, though that might yield some confusing code:
```zig
var a: i32 = 0; // inferred address space: .private
export fn b() void {
  var c: i32 = 0; // infered address space: .function
}
```

Currently i also forbid specifying an explicit address space on function-local variables, though i do not know whether SPIR-V actually permits locals in a different address space than `.function`.

Snektron:
At the moment there is no method to construct pointers to function in a different address space, as function pointers and declarations are kinda ambiguous in type.zig.

lib/std/zig/tokenizer.zig
```diff
@@ -11,6 +11,7 @@ pub const Token = struct {
     };
 
     pub const keywords = std.ComptimeStringMap(Tag, .{
+        .{ "addrspace", .keyword_addrspace },
```
matu3ba, line 14: consistency nit: example `\\const y: []allowzero align(8) addrspace(.generic) const volatile u32 = undefined;`
shows align before addrspace.

lib/std/zig/tokenizer.zig
```diff
@@ -132,6 +133,7 @@ pub const Token = struct {
         float_literal,
         doc_comment,
         container_doc_comment,
+        keyword_addrspace,
```
matu3ba, line 136: same consistency nit: `\const y: []allowzero align(8) addrspace(.generic) const volatile u32 = undefined;`

lib/std/zig/tokenizer.zig
```diff
@@ -251,6 +253,7 @@ pub const Token = struct {
                 .angle_bracket_angle_bracket_right => ">>",
                 .angle_bracket_angle_bracket_right_equal => ">>=",
                 .tilde => "~",
+                .keyword_addrspace => "addrspace",
```
matu3ba, line 256: same consistency nit.

src/Module.zig
```diff
@@ -366,8 +368,8 @@ pub const Decl = struct {
     is_exported: bool,
     /// Whether the ZIR code provides an align instruction.
     has_align: bool,
-    /// Whether the ZIR code provides a linksection instruction.
-    has_linksection: bool,
+    /// Whether the ZIR code provides a linksection and address space instruction.
+    has_linksection_or_addrspace: bool,
```
matu3ba, line 372: `addrspace` comes before `linksection`, so it could be called `has_addrspace_or_linksection` for consistency. Doesnt make a notable difference though, so just close if you disagree.

matu3ba (changes requested):
Mostly the doc comment stuff and minor nits regarding ordering of stuff.
Would be also great to have other address spaces than `.generic` tested (with stage2).

Do you have any plan/idea how and when this should be documented? Probably adding a stub in the docs would also be fine.
From what I am reading fs,gs and ss are used for [OS-stuff](https://reverseengineering.stackexchange.com/a/2009) and security-stuff, but I am unsure where this fits best (back-end specific stuff doesnt belong into the lang reference).

lib/std/zig/tokenizer.zig
Snektron: This remainder of this list sorted alphabetically, in which case addrspace should be placed before align.

lib/std/zig/tokenizer.zig
Snektron: Same, these are also sorted alphabetically

lib/std/zig/tokenizer.zig
Snektron: Same sorting

src/Module.zig
Snektron: Hm yes, though this would have some further reaching consequences than just the name (like the zir format). I think this is probably also fine

Snektron:
> Would be also great to have other address spaces than `.generic` tested (with stage2), so putting a TODO for "do this after selfhosted" might be good.

There also is a test in stage2/llvm.zig where the functionality of `addrspace(.fs)` is tested. Unfortunately these address spaces are hard to functionaly test, and are mainly useful on freestanding architectures.

> Do you have any plan/idea how and when this should be documented? Probably adding a stub in the docs would also be fine.
> From what I am reading fs,gs and ss are used for [OS-stuff](https://reverseengineering.stackexchange.com/a/2009) and security-stuff, but I am unsure where this fits best (back-end specific stuff doesnt belong into the lang reference).

`addrspace` is meant to also be used for other architectures, not x86 specifically. For example, AVR and GPU targets. I didn't think it was a good idea to also add these in this pr, but it does mean that the `addrspace` attribute itself is not a backend-specific feature. 

As for documentation, I wasn't sure on that either, because this is of course a feature that is currently not supported by the stage 1 compiler. I can see a few options, though:
- Add a note that this is a stage-2-only feature.
- Backport to stage 1. I'm not sure whether this is worth the effort.
- Wait until stage 1 is yeeted before merging this.

andrewrk (approved):
Excellent work.

PR #9652, opened 2021-08-29
plan9: emit debug info

g-w1:
- [x] emit line debug info in codegen
- [x] add filename info
- [x] link the line info together in the linker
### Here we should get source debugging working, but no variable checking yet (pr might end here)
- [x] ~~emit automatic (local) variable and function parameter symbols~~ (for another pr)
- [x] ~~emit acid code to type the local vars and params~~ (for another pr)


g-w1:
Ready for review, but merge depends on solution to https://github.com/ziglang/zig/issues/9672.

g-w1:
Also, free and netbsd are failing on realpath, what should I use instead?

Issue #4833, opened 2020-03-27
Support response files for zig cc and zig c++


Sahnvour:
https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-response-files?view=vs-2019

In windows build world, it's not uncommon to use `.rsp` files containing command lines and have some executables read from them.
`zig cc` or `c++` should probably support them to act as a real clang replacement.

I encountered this while trying to use `CXX="zig c++ -target x86_64-windows-gnu"` and the most simple [meson](mesonbuild.com) project.
See [my example](https://gist.github.com/Sahnvour/21c82bdd8e6ededec1c28c0375ad9999).

By default meson will generate a ninja build file, configured to use response files. The command actually ran is then `zig c++ -target x86_64-windows-gnu @zigcpp@exe/zigcpp.cpp.obj.rsp`

Quoting @andrewrk 
> zig needs to gain understanding of this @ syntax to look at command lines. because it doesn't understand it, and it's a positional argument, so it thinks its a linker object
> so it sends @zigcpp@exe/zigcpp.cpp.obj.rsp to the linker, but those args are for the compiler
> since we have to act on positional arguments, we have to know about all special positional arguments

PR #9654, opened 2021-08-29
More response file support

PhaseMage:
I hit the "quotes in an RSP file" issue when trying to compile gRPC using `zig cc`.

This modifies (and renames) `ArgIteratorWindows` in process.zig such that it works with arbitrary strings (or the contents of an RSP file).

In main.zig, this new `ArgIteratorGeneral` is used to address the "TODO" listed in `ClangArgIterator`. 

This change is meant to close out issue #4833

**Benefits:**

- It has the nice attribute of handling "RSP file" arguments in the same way it handles `cmd_line` arguments.
- High Performance, minimal allocations
- Fixed bug in previous `ArgIteratorWindows`, where final trailing backslashes in a command line were entirely dropped
- Added a test case for the above bug
- Harmonized the `ArgIteratorXxxx.initWithAllocator()` and `next()` interface across Windows/Posix/Wasi (Moved Windows errors to `initWithAllocator()` rather than `next()`)
- Likely perf benefit on Windows by doing `utf16leToUtf8AllocZ()` only once for the entire cmd_line

**Breaking Change in std library on Windows: Call `ArgIterator.initWithAllocator()` instead of `ArgIterator.init()`**

**Testing Done:**

- Wrote a few new test cases in process.zig
- `zig.exe build test -Dskip-release` (no new failures seen)
- `zig cc` now builds gRPC without error

### DIFF
--- a/lib/std/process.zig
+++ b/lib/std/process.zig
@@ -356,7 +356,7 @@ pub fn ArgIteratorGeneral(comptime options: ArgIteratorGeneralOptions) type {
         /// cmd_line_utf16le MUST be encoded UTF16-LE, and is converted to UTF-8 in an internal buffer
         pub fn initUtf16le(allocator: Allocator, cmd_line_utf16le: [*:0]const u16) InitUtf16leError!Self {
             var utf16le_slice = mem.sliceTo(cmd_line_utf16le, 0);
-            var cmd_line = std.unicode.utf16leToUtf8AllocZ(allocator, utf16le_slice) catch |err| switch (err) {
+            var cmd_line = std.unicode.utf16leToUtf8Alloc(allocator, utf16le_slice) catch |err| switch (err) {
                 error.ExpectedSecondSurrogateHalf,
                 error.DanglingSurrogateHalf,
                 error.UnexpectedSecondSurrogateHalf,
--- a/lib/std/process.zig
+++ b/lib/std/process.zig
@@ -203,6 +203,8 @@ pub const ArgIteratorPosix = struct {
     index: usize,
     count: usize,
 
+    pub const InitError = error{};
+
     pub fn init() ArgIteratorPosix {
         return ArgIteratorPosix{
             .index = 0,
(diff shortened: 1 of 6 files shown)


lib/std/process.zig
```diff
@@ -299,152 +301,223 @@ pub const ArgIteratorWasi = struct {
     }
 };
 
-pub const ArgIteratorWindows = struct {
-    index: usize,
-    cmd_line: [*]const u16,
+pub fn ArgIteratorGeneral(comptime options: anytype) type {
+    return struct {
+        allocator: *Allocator,
```
marler8997, line 306: The original implementation didn't need an allocation.  My guess is you need one to implement string escapes?  My suggestion would be to leave the allocator out of the iterator, and allow the caller to perform the escape/allocation if desired.

PhaseMage: Not exactly... The original implementation _did_ require an allocator. It was in the 'next()' call. I moved the allocator to the init() function to harmonize the interface cross-platform.

lib/std/process.zig
```diff
@@ -299,152 +301,223 @@ pub const ArgIteratorWasi = struct {
     }
 };
 
-pub const ArgIteratorWindows = struct {
-    index: usize,
-    cmd_line: [*]const u16,
+pub fn ArgIteratorGeneral(comptime options: anytype) type {
```
marler8997, line 304: I'd give this an explicit type.

PhaseMage:
@andrewrk You mentioned wanting zig-based Response file support. I believe this PR should be what you are looking for, but let me know if there is more support you want added.

lib/std/process.zig
```diff
@@ -299,196 +301,279 @@ pub const ArgIteratorWasi = struct {
...
+                .cmd_line = cmd_line,
+                .buffer = buffer,
+            };
+        }
 
-    pub fn init() ArgIteratorWindows {
-        return initWithCmdLine(os.windows.kernel32.GetCommandLineW());
-    }
+        /// Initialize the arguments from a Response File. "*.rsp"
+        pub fn initFromResponseFile(allocator: Allocator, resp_file_path: []const u8) !Self {
```
andrewrk, line 340: Let's move this logic back over to src/main.zig instead of having it here in std/process.zig.

lib/std/process.zig
```diff
@@ -299,196 +301,279 @@ pub const ArgIteratorWasi = struct {
...
+        end: usize = 0,
+        buffer: []u8,
+
+        pub const Self = @This();
+
+        pub const InitError = error{OutOfMemory};
+        pub const InitUtf16leError = error{ OutOfMemory, InvalidCmdLine };
+
+        /// cmd_line_utf8 is duplicated into an internal buffer
+        pub fn init(allocator: Allocator, cmd_line_utf8: []const u8) InitError!Self {
```
andrewrk, line 326: why is it duplicated into an internal buffer? seems unnecessary. certainly it is unnecessary for the one use case that we have for it.

andrewrk (changes requested):
@PhaseMage thanks for your patience on getting a review for this. I'm definitely interested in merging this, but I do have a couple of requests first.

Also I hope you don't mind I took the liberty to rebase this for you, made a couple of trivial edits, and force pushed to your branch.

lib/std/process.zig
PhaseMage: This is duplicated because the other init*() functions needed to duplicate, and I was just trying to keep it simple.

But it's not too difficult to add an optional member that holds the cmd_line if it should be free'd by the deinit().

PhaseMage:
I can spend more time on this over the weekend.

PhaseMage:
CI appears to be passing now (ignoring the single failure in ziglang.zig / BuildWindows, which appears to be an infrastructure failure, it was passing earlier). Let me know if there's anything else you'd like me to change.

Vexu:
I rebased this to master and made `next` not duplicate the argument by default.

andrewrk:
Thanks @PhaseMage and @Vexu for landing this improvement!

PR #9655, opened 2021-08-30
stage2: implement runtime `%` and `@rem`

nektro:


g-w1:
Can tests be added please?

nektro:
would you suggest I add new tests? trying to add all of `behavior/math.zig` yields `TODO add __trunctfhf2 to compiler-rt`

g-w1:
Yes, test/stage2/cbe.zig would be a good place, maybe llvm too.
-------- Original message --------From: Meghan ***@***.***> Date: 8/30/21  2:54 PM  (GMT-05:00) To: ziglang/zig ***@***.***> Cc: g-w1 ***@***.***>, Comment ***@***.***> Subject: Re: [ziglang/zig] stage2: implement runtime `%` and ***@***.***` (#9655) 
would you suggest I add new tests? trying to add all of behavior/math.zig yields TODO add __trunctfhf2 to compiler-rt

andrewrk:
Nice work!

Note that the behavior tests are the preferred place for testing behavior. Faster feedback time, too. And it's ok to add new cases rather than trying to move entire files at once. If you look at behavior.zig you can see some files are partially moved over, with files being split between foo_stage1.zig and foo.zig.

PR #9659, opened 2021-08-30
stage2: only initialize the llvm backend for the target we are building

### LINKED ISSUE #9596: only call LLVM initialization functions for the actual target
Currently, our LLVM bindings look like this:

https://github.com/ziglang/zig/blob/2f1abd919a8465e515875cd08816778df795aabe/src/codegen/llvm/bindings.zig#L683-L780

This is called here:

https://github.com/ziglang/zig/blob/2f1abd919a8465e515875cd08816778df795aabe/src/codegen/llvm.zig#L259-L265

Presumably LLVM breaks these functions up by target because doing the initialization for unused targets is wasted time.

Therefore, this logic should be changed to observe the Target that zig is building for, and only call the appropriate LLVM initialization functions.

An implementation of this issue should come with a quick benchmark test results so we can see if this made any difference.

nektro:
Fixes #9596

### DIFF
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -167,7 +167,7 @@ pub const Object = struct {
         const context = llvm.Context.create();
         errdefer context.dispose();
 
-        initializeLLVMTargets();
+        llvm.initializeLLVMTarget(options.target.cpu.arch);
 
         const root_nameZ = try gpa.dupeZ(u8, options.root_name);
         defer gpa.free(root_nameZ);
@@ -254,14 +254,6 @@ pub const Object = struct {
     pub fn destroy(self: *Object, gpa: *Allocator) void {
         self.deinit();
         gpa.destroy(self);
-    }
-
-    fn initializeLLVMTargets() void {
-        llvm.initializeAllTargets();
-        llvm.initializeAllTargetInfos();
-        llvm.initializeAllTargetMCs();
-        llvm.initializeAllAsmPrinters();
-        llvm.initializeAllAsmParsers();
     }
 
     fn locPath(
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -1949,3 +1949,132 @@ pub const FuncGen = struct {
         return self.llvmModule().getIntrinsicDeclaration(id, null, 0);
     }
 };
+
+pub fn initializeLLVMTarget(arch: std.Target.Cpu.Arch) void {
+    switch (arch) {
+        .aarch64, .aarch64_be, .aarch64_32 => {
+            llvm.LLVMInitializeAArch64Target();
+            llvm.LLVMInitializeAArch64TargetInfo();
+            llvm.LLVMInitializeAArch64TargetMC();
+            llvm.LLVMInitializeAArch64AsmPrinter();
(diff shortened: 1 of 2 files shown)


andrewrk (changes requested):
Implementation looks good but please put `initializeLLVMTarget` in codegen/llvm.zig rather than in the bindings file. The reason `LLVMInitializeAllTargetInfos` was in the bindings file was that it was translated from an inline function in an .h file.

src/codegen/llvm/bindings.zig
```diff
@@ -680,103 +685,108 @@ extern fn LLVMInitializeSystemZAsmParser() void;
...
+            LLVMInitializeSystemZAsmParser();
+        },
+        .i386, .x86_64 => {
+            LLVMInitializeX86Target();
+            LLVMInitializeX86TargetInfo();
+            LLVMInitializeX86TargetMC();
+            LLVMInitializeX86AsmPrinter();
+            LLVMInitializeX86AsmParser();
+        },
+        else => {},
```
andrewrk, line 788: list all the targets exhaustively please

nektro:
Should I make all the `LLVMInitialize` functions public then?

src/codegen/llvm.zig
```diff
@@ -1957,3 +1949,145 @@ pub const FuncGen = struct {
...
+            llvm.LLVMInitializeX86TargetMC();
+            llvm.LLVMInitializeX86AsmPrinter();
+            llvm.LLVMInitializeX86AsmParser();
+        },
+        .xcore => {
+            llvm.LLVMInitializeXCoreTarget();
+            llvm.LLVMInitializeXCoreTargetInfo();
+            llvm.LLVMInitializeXCoreTargetMC();
+            llvm.LLVMInitializeXCoreAsmPrinter();
+            // llvm.LLVMInitializeXCoreAsmParser();
```
andrewrk, line 2072:
```suggestion
            // There is no LLVMInitializeXCoreAsmParser function.
```

src/codegen/llvm/bindings.zig
```diff
@@ -1,5 +1,6 @@
 //! We do this instead of @cImport because the self-hosted compiler is easier
 //! to bootstrap if it does not depend on translate-c.
+const std = @import("std");
```
andrewrk, line 3: unused import

src/codegen/llvm.zig
```diff
@@ -1957,3 +1949,145 @@ pub const FuncGen = struct {
         return self.llvmModule().getIntrinsicDeclaration(id, null, 0);
     }
 };
+
+pub fn initializeLLVMTarget(arch: std.Target.Cpu.Arch) void {
```
andrewrk, line 1953:
```suggestion
fn initializeLLVMTarget(arch: std.Target.Cpu.Arch) void {
```

andrewrk (changes requested):
Thanks - almost there

andrewrk (approved):

marler8997:
Out of curiosity, what was the performance difference?

nektro:
`773,734 ns` to `90,751 ns` on my `AMD Ryzen 5 4500U (6) @ 2.375GHz`

PR #9662, opened 2021-08-31
libunwind: fix unwinding through libunwind stack frames

### LINKED ISSUE #9591: zig c++ does not support exceptions anymore
```
$ cat main.cpp
#include <iostream>
#include <exception>

int main() {
  try {
    throw std::exception();
  } catch (std::exception &) {
    std::cerr << "yikes";
  }
  return 0;
}

$ ~/Downloads/zig-linux-x86_64-0.8.0/zig c++ main.cpp && ./a.out
yikes

$ ~/Downloads/zig-linux-x86_64-0.9.0-dev.804+4c9d41730/zig c++ main.cpp && ./a.out
libc++abi: terminating with uncaught exception of type std::exception: std::exception
[5]    2340536 IOT instruction (core dumped)  ./a.out
```
Plugging the 0.8.0 `libunwind.a` into the 0.9.0-dev linker command makes it work, so perhaps something similar to https://reviews.llvm.org/D71117? Though that one is about ARM, while I'm on x86_64-linux-gnu.

Kha:
This fixes #9591 on x86_64-linux-gnu, and presumably on ARM. Not sure about other platforms, are there platforms that use libunwind but not unwind tables? And should this change come with a test (in `test/standalone/c_compiler`?)?

### DIFF
--- a/src/libunwind.zig
+++ b/src/libunwind.zig
@@ -71,6 +71,8 @@ pub fn buildStaticLib(comp: *Compilation) !void {
         try cflags.append("-Wa,--noexecstack");
         try cflags.append("-fvisibility=hidden");
         try cflags.append("-fvisibility-inlines-hidden");
+        // necessary so that libunwind can unwind through its own stack frames
+        try cflags.append("-funwind-tables");
 
         // This is intentionally always defined because the macro definition means, should it only
         // build for the target specified by compiler defines. Since we pass -target the compiler


andrewrk:
Thanks for the patch! I'm happy to accept this as is, and will do so now, but if you felt inspired to follow this up with a test case, I would suggest adding a case here: https://github.com/ziglang/zig/blob/master/test/stack_traces.zig and then adding the necessary test harness improvements needed here: https://github.com/ziglang/zig/blob/75263e160e477a210a3d2a007b6d66c63b85001f/test/tests.zig#L576

PR #9670, opened 2021-09-01
Fix libc file to check against compilation target instead of native

SpexGuy:
Libc file validation currently checks for msvc paths based on the compiler's native target, not the cross-compilation target.  So using a libc file to target android fails on a windows machine unless the file specifies a msvc path.  This PR fixes it to use the cross-compilation target for validation instead.


andrewrk:
The CI failure is zig fmt

Issue #363, opened 2017-05-05
convert TODO comments from source code into issues


andrewrk:
"TODO" should occur exactly 0 times in the source code. If there's an issue it should be documented on the issue tracker. Source code can link to an issue if it seems appropriate.

andrewrk:
Also look for "NOTE" and "XXX". Some contributors have been using those.

daurnimator:
https://todos.tickgit.com/browse?repo=https://github.com/ziglang/zig

andrewrk:
Also look for `FIXME`.

nektro:
Also look for `deprecated` inside of `@compileError`s.

PR #9678, opened 2021-09-02
compiler_rt: add __ctzsi2, __ctzdi2 and __ctzti2

matu3ba:
- structure derived from count0bits.zig
    - test cases derived from clzsi2_test.zig and
      cross-checked via short helper program

    See #1290


matu3ba:
Here is the simple helper program to generate (some) expected results https://gist.github.com/matu3ba/a6c0a4fab5d9a8a45db3fc8a7ede2c5c.
Kudos @travisstaloch for making saturation arithmetic simple.

lib/std/special/compiler_rt/ctzdi2_test.zig
```diff
@@ -0,0 +1,329 @@
...
+const testing = @import("std").testing;
+
+fn test__ctzdi2(a: u64, expected: i32) !void {
+    // XXX At high optimization levels this test may be horribly miscompiled if
+    // one of the naked implementations is selected.
+    var nakedCtzdi2 = ctz.__ctzdi2;
+    var actualCtzdi2 = @ptrCast(fn (a: i64) callconv(.C) i32, nakedCtzdi2);
+    var x = @bitCast(i64, a);
+    var result = actualCtzdi2(x);
+    try testing.expectEqual(expected, result);
```
andrewrk, line 11: what's going on here? I don't understand why this has to be complicated, can't it just call the function that it's testing like normal?

matu3ba: The idea was to provide native assembly eventually, ie [the ones for x86](https://en.wikipedia.org/wiki/X86_Bit_manipulation_instruction_set#BMI1_(Bit_Manipulation_Instruction_Set_1) since we dont have a peephole optimizer without LLVM.

Would you prefer leaving this out (for now)?

andrewrk: Yes please. Let's not have this XXX naked stuff and TODO comments everywhere when nothing is actually wrong. If there should be an inline assembly version of a function, there can be a github issue for that.

Related: #363

matu3ba: I removed that stuff in YAGNI commits.

matu3ba:
This is ready for merging.

andrewrk:
Nice work, thank you :+1:

Issue #1284, opened 2018-07-24
Saturating arithmetic


skyfex:
This proposal is more for the sake of completeness rather than something that should be added to Zig any time soon.

Zig has explicit support for integer operations with fault on overflow, and wrap-around behaviour. But there is a third relatively common behaviour for integer operations, especially in DSPs: saturating arithmetic: https://en.wikipedia.org/wiki/Saturation_arithmetic

Here's some related proposals from LLVM: 
https://reviews.llvm.org/D6976
https://reviews.llvm.org/D8371

There are some options how to add support for this:
1. Add more operators that perform saturating arithmetic on standard integer types
2. Define a saturating integer type that uses saturating arithmetic when using the standard operators.
3. Just have functions/built-ins for it

I don't have any opinion on how to implement it. This is just a basis for discussion, and I think Zig needs the input from someone with good experience with DSPs to get feedback on that use-case before this proposal is considered.

monouser7dig:
where is such behaviour used/ wanted?

tgschultz:
Seems there are instructions in several architectures with saturating behavior, including ARM and SIMD. Makes sense to me that we should explicitly declare that behavior if we want it, rather than rely on the compiler to guess intent. Maybe it isn't enough of a common case to expend operators on and we should just use builtins (@saturateAdd(), etc), but it would be more consistent. To that end, I propose `<op>|`, for the simple reason that it looks like a wall.

```
a +| b;
a +|= b;

a -| b;
a -|= b;

a *| b;
a *|= b;
```

skyfex:
@monouser7dig Basically any kind of signal processing math code. I guess what most people may be familiar is image processing. Imagine you want to double the brightness, so you multiply each value by 2. But you don't want the brightness to wrap around, so when it goes above 255, you clamp it to 255.  The Wikipedia article explained it fairly well.

@tgschultz Yeah, that seems to be the most visually descriptive. But it could be confusing that | is used with bitwise or as well.

monouser7dig:
That’s a nice example, I think builtins for these operations would be a good addition.

Tetralux:
Wouldn't `@clampedAdd` make more intuitive sense, considering that you'd implement this functionality with a call to some `clamp` function?

It's not like we want people to have to read Wikipedia articles to understand language features, right?

scottjmaddox:
[...]

PR #9679, opened 2021-09-02
sat-arithmetic: add operator support

travisstaloch:
- implements accepted syntax from #1284
- adds initial support for the operators +|, -|, *|, <<|, +|=, -|=, *|=, <<|=
- uses operators in addition to builtins in behavior test
- adds binOpExt() and assignBinOpExt() to AstGen.zig. these need to be audited


travisstaloch:
TODOS
- update langref
- c backend support
- zig fmt support - currently deleting '|' from '<<|='
- zls support

I'm not sure what needs to be done to add support to zig fmt or zls.  Any advice appreciated.

andrewrk:
Here's some advice to get you going-

> * c backend support

You can add any functions you need to src/link/C/zig.h, and then in codegen/c.zig you can emit calls to them. This part is pretty fun! I recommend to play with godbolt to test out your saturating arithmetic implementations and see how optimizable they are (the goal would be that most C compilers would be able to optimize them into machine code instructions when available).

> * zig fmt support - currently deleting '|' from '<<|='

Double check your edits to tokenizer.zig. Perhaps try adding some test cases at the bottom of tokenizer.zig (there are a few examples already there you can look at). Something to keep in mind is that tokenizer.zig and tokenizer.cpp are intentionally nearly identical implementations, so your edits to each file should look almost the same.

> * zls support

Not needed for this PR to be merged :)

src/Sema.zig
```diff
@@ -5773,10 +5774,17 @@ fn zirSatArithmetic(
...
+    const rhs = sema.resolveInst(extra.rhs);
+
+    return sema.analyzeArithmetic(block, .extended, lhs, rhs, sema.src, lhs_src, rhs_src, extended);
 }
 
+// TODO: audit - not sure if its a good idea to reuse this, adding `opt_extended` param
+// FIXME: somehow, rhs of <<| is required to be Log2T. this should accept T
 fn analyzeArithmetic(
     sema: *Sema,
     block: *Scope.Block,
```
travisstaloch, line 5790: I'm reusing analyzeArithmetic() by adding an `opt_extended` param here.  Is this a bad idea?

Also, I can't figure out where the check that rhs of <<| is a Log2T is happening (and how to disable it).  It needs to accept T, but currently doesn't.

travisstaloch:
> This part is pretty fun! I recommend to play with godbolt to test out your saturating arithmetic implementations and see how optimizable they are (the goal would be that most C compilers would be able to optimize them into machine code instructions when available).

I looked at them on godbolt with [clang](https://godbolt.org/z/rx4o8xTMj) and [gcc](https://godbolt.org/z/r4cWcfvYW), both with -O1.  They were all branchless using cmovs except for gcc signed mul with 64 bits has one jump instruction.  I'm not super confident in the signed shl impl.  

> Double check your edits to tokenizer.zig. Perhaps try adding some test cases at the bottom of tokenizer.zig

This worked well. I was updating the result instead of the state when in the `<<` transition.  

Thank you for the guildance!  Please let me know if you get a chance to look at this again.  I've left some TODOs and FIXMEs.

matu3ba (commented):
Some tests and simplification with using maxInt from math.zig.

Changing the behavior to allowing types on right hand side that are not Log2(left hand side) sounds to me inconsistent to the other `<<` operations and (from my perspective) can hide logical bugs without crashing/overflowing in tests if one is not cautious.

src/codegen/c.zig
```diff
@@ -1251,6 +1255,117 @@ fn airWrapOp(
...
+        .c_longlong => "LLONG_MAX",
+        .c_ulonglong => "ULLONG_MAX",
+        .isize => "INTPTR_MAX",
+        .usize => "UINTPTR_MAX",
+        else => blk: {
+            const pow_bits = switch (int_info.signedness) {
+                .signed => bits - 1,
+                .unsigned => bits,
+            };
+            const val = std.math.pow(u65, 2, pow_bits) - 1;
```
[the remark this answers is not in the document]
travisstaloch: maxInt requires a type while this uses bitcount. a type is not available here at comptime.

src/codegen/c.zig
```diff
@@ -1251,6 +1255,117 @@ fn airWrapOp(
...
+    const min = switch (int_info.signedness) {
+        .unsigned => "0",
+        else => switch (inst_ty.tag()) {
+            .c_short => "SHRT_MIN",
+            .c_int => "INT_MIN",
+            .c_long => "LONG_MIN",
+            .c_longlong => "LLONG_MIN",
+            .isize => "INTPTR_MIN",
+            else => blk: {
+                const val = -1 * std.math.pow(i65, 2, @intCast(i65, bits - 1));
```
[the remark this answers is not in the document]
travisstaloch: minInt requires a type which is not available here. i've added a comment locally. will push soon.

lib/std/zig/tokenizer.zig
```diff
@@ -1919,6 +1998,12 @@ test "tokenizer - invalid builtin identifiers" {
     try testTokenize("@0()", &.{ .invalid, .integer_literal, .l_paren, .r_paren });
 }
 
+test "tokenizer - saturating" {
+    try testTokenize("<<", &.{.angle_bracket_angle_bracket_left});
+    try testTokenize("<<|", &.{.angle_bracket_angle_bracket_left_pipe});
+    try testTokenize("<<|=", &.{.angle_bracket_angle_bracket_left_pipe_equal});
```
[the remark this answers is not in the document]
travisstaloch: just pushed a commit with updates to langref including precedence.  also added additional tokenization tests.

travisstaloch:
> Why do you think requiring an @intcast for the left-shift is a bad idea?

@matu3ba I was just trying to match the behavior of @shlWithSaturation which doesn't require Log2T, but accepts T.

src/Sema.zig
travisstaloch: > Why do you think requiring an @intcast for the left-shift is a bad idea?

@matu3ba I was just trying to match the behavior of @shlWithSaturation which doesn't require Log2T, but accepts T.

travisstaloch:
> Changing the behavior to allowing types on right hand side that are not Log2(left hand side) sounds to me inconsistent to the other `<<` operations and (from my perspective) can hide logical bugs without crashing/overflowing in tests if one is not cautious.

This is how [@shlWithSaturation](https://ziglang.org/documentation/master/#shlWithSaturation) is defined: 

> Unlike other @shl builtins, shift_amt doesn't need to be a Log2T as saturated overshifting is well defined.

src/Sema.zig
travisstaloch: also, this is how the [llvm intrinsics](https://llvm.org/docs/LangRef.html#llvm-sshl-sat-intrinsics) are defined

travisstaloch:
looks like i messed up during rebase. working on it.

src/AstGen.zig
```diff
@@ -574,17 +594,25 @@ fn expr(gz: *GenZir, scope: *Scope, rl: ResultLoc, node: Ast.Node.Index) InnerEr
...
+        .assign_mul_sat => {
+            try assignBinOpExt(gz, scope, node, .mul_with_saturation, Zir.Inst.SaturatingArithmetic);
+            return rvalue(gz, rl, .void_value, node);
+        },
 
         // zig fmt: off
-        .bit_shift_left  => return shiftOp(gz, scope, rl, node, node_datas[node].lhs, node_datas[node].rhs, .shl),
-        .bit_shift_right => return shiftOp(gz, scope, rl, node, node_datas[node].lhs, node_datas[node].rhs, .shr),
+        .bit_shift_left     => return shiftOp(gz, scope, rl, node, node_datas[node].lhs, node_datas[node].rhs, .shl),
+        .bit_shift_left_sat => return binOpExt(gz, scope, rl, node, node_datas[node].lhs, node_datas[node].rhs, .shl_with_saturation, Zir.Inst.SaturatingArithmetic),
```
andrewrk, line 604: This does not look like it belongs here; it should go with other calls to `binOpExt`.

test/behavior/saturating_arithmetic.zig
```diff
@@ -11,13 +11,34 @@ fn testSaturatingOp(comptime op: Op, comptime T: type, test_data: [3]T) !void {
...
-        .mul => @mulWithSaturation(a, b),
-        .shl => @shlWithSaturation(a, b),
-    };
-    try expectEqual(expected, actual);
+    {
+        const actual = switch (op) {
+            .add => @addWithSaturation(a, b),
+            .sub => @subWithSaturation(a, b),
+            .mul => @mulWithSaturation(a, b),
+            .shl => @shlWithSaturation(a, b),
```
andrewrk, line 19: Don't forget to delete these now that we have actual syntax!

andrewrk (changes requested):
I have not fully looked at the review but I am about to go do something else so I wanted to give you something to work with in the meantime-

src/Sema.zig
```diff
@@ -5786,6 +5794,7 @@ fn analyzeArithmetic(
     src: LazySrcLoc,
     lhs_src: LazySrcLoc,
     rhs_src: LazySrcLoc,
+    opt_extended: ?Zir.Inst.Extended.InstData,
```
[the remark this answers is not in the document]
travisstaloch: > This does not happen for @shlWithSaturation. So I'm not sure why you are observing any such behavior.

I should have noted that I was only seeing this Log2T requirement with stage2 when testing my cbe code.  I don't think this is a requirement anymore. 

> I suggest to not try to re-use analyzeArithmetic. Case in point, the switch on zir_tag is missing the handling for the new extended ops, and if implemented will branch unnecessarily.

I have created an alternative code path: `Sema.analyzeSatArithmetic()`.  It is incomplete (lacking support for pointers and comptime arithmetic) but seems to be working.

I also feel better about the cbe shl methods.  I've tested that they return the same values as we do using `<<|` for the types i8, u8, i16, u16, i32, u32, i64, u64 in ReleaseFast (limiting shift amounts to Log2T).  Test files [here](https://zigbin.io/ddb34a)

travisstaloch:
[CI failure](https://dev.azure.com/ziglang/zig/_build/results?buildId=16815&view=logs&j=3e8797c7-5b0a-5f2c-07a4-1bc5e60a122e&t=e20ebee5-9d72-5103-f7c4-43227f669eab)

> del : Cannot remove item D:\a\1\s\sfx.exe: The process cannot access the file 'D:\a\1\s\sfx.exe' because it is being used by another process.

I'm confident this was just a glitch.  Last few commits have passed CI

PR #9681, opened 2021-09-02
error -> anyerror coercion, comptime zirErrorName

g-w1:


src/Sema.zig
```diff
@@ -7090,9 +7090,28 @@ fn zirEmbedFile(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileE
 }
 
 fn zirErrorName(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index) CompileError!Air.Inst.Ref {
+    const tracy = trace(@src());
+    defer tracy.end();
+
     const inst_data = sema.code.instructions.items(.data)[inst].un_node;
     const src = inst_data.src();
-    return sema.mod.fail(&block.base, src, "TODO: Sema.zirErrorName", .{});
+    _ = src;
```
miscco, line 7098: Out of curiosity what is the purpose of this assignment?

`src` is always used in the if below

g-w1: Thanks for the review, I am cleaning up someone else's code, will remove.

src/type.zig
```diff
@@ -1145,6 +1145,106 @@ pub const Type = extern union {
         }
     }
 
+    pub fn isGlobalErrorSet(ty: Type) bool {
```
andrewrk, line 1148: The logic for this is unfortunately not this simple. A Type could be a `error_set_inferred` in which case it *might* be the global error set. Determining whether an inferred error set is the global error set requires "resolving" the inferred error set, which could cause compile errors to be emitted. It means we cannot implement this function here.

andrewrk (changes requested):

src/type.zig
g-w1: I have stubbed out isGlobalErrorSet and put it in Sema.zig for now, as it can return an error if the resolution fails.

PR #9685, opened 2021-09-04
openbsd: event loop: use EVFILT_TIMER instead of EVFILT_USER

semarie:
OpenBSD doesn't implement EVFILT_USER filter for kqueue(2), so we couldn't use that for event loop.

instead, use a EVFILT_TIMER filter with EV_ONESHOT (trigger only once) and delay 0sec (which trigger immediatly).
it fits the usage of EVFILT_USER which is only used to "wakeup" the kevent(2) call from userland.

### DIFF
--- a/lib/std/event/loop.zig
+++ b/lib/std/event/loop.zig
@@ -266,7 +266,7 @@ pub const Loop = struct {
                     self.extra_threads[extra_thread_index] = try Thread.spawn(.{}, workerRun, .{self});
                 }
             },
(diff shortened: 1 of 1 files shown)


semarie:
I separated `openbsd` from others BSD to avoid regressions as I am not able to test that `EVFILT_TIMER` implementation on them would behave the same than on OpenBSD.

Please note that with it, I am able to build and run tests of [zig-network](https://github.com/MasterQ32/zig-network) on OpenBSD.

andrewrk:
Nice!

PR #9687, opened 2021-09-04
compiler_rt: add __popcountsi2, __popcountdi2 and __popcountti2

matu3ba:
compiler_rt: add __popcountsi2, __popcountdi2 and __popcountti2

- apply simpler approach than LLVM for __popcountdi2
  taken from The Art of Computer Programming and generalized
- rename popcountdi2.zig to popcount.zig
- test cases derived from popcountdi2_test.zig
- tests: compare naive approach 10_000 times with
  random numbers created from naive seed 42

See #1290


matu3ba:
This is ready for review/merge.

I checked locally that each popcount works for the complete u32 range on x86_64 with the following code
```zig
// The following works on x86_64 at least:
var a: u32 = 0;
while (a < 0xffffffff) {
    const x = @bitCast(i32, a);
    try test__popcountsi2(x);
    a += 1;
}
const x = @bitCast(i32, a);
try test__popcountsi2(x); // a == 0xffffffff
```
Currently this PR uses this for testing, but I could change it to use another seed and log it for the CI:
```zig
    const RndGen = @import("std").rand.DefaultPrng;
    var rnd = RndGen.init(42);
    var i: u32 = 0;
    while (i < 10_000) : (i += 1) {
        var rand_num = rnd.random().int(i32);
        try test__popcountsi2(rand_num);
    }
```

andrewrk:
Thanks!

PR #9695, opened 2021-09-06
std: update event loop for recent usingnamespace changes

### LINKED ISSUE #9693: Error when enabling evented I/O on Linux
Maybe I am doing something wrong(new to zig) but adding this line in my code:
` pub const io_mode = .evented;`

Causes the error:
zig/std/event/loop.zig:499:40: error: container 'std.os' has no member called 'EPOLLET'
                self.linuxWaitFd(fd, os.EPOLLET | os.EPOLLONESHOT | os.EPOLLOUT);

edit: Forgot to add, I'm using latest zig master ( built myself).
I see there was an old issue related to this in 2020, that is marked fixed.

g-w1:
Closes #9693

### DIFF
--- a/lib/std/c/netbsd.zig
+++ b/lib/std/c/netbsd.zig
@@ -766,6 +766,9 @@ pub const EVFILT_TIMER = 6;
 /// Filesystem events
 pub const EVFILT_FS = 7;
 
+/// User events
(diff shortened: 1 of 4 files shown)


g-w1:
Seems I will also have to fix for bsds ...

PR #9697, opened 2021-09-07
std: fix linux event.loop in wake of std.os reorganization

nektro:


Luukdegram:
There's already an open PR for this: #9695

PR #9712, opened 2021-09-09
reexport crypto/phc_encoding

x13a:
Make `phc_encoding` public available in `crypto.pwhash.phc_format`. It is useful for serialization password hash functions that not available in std.

/cc @jedisct1

### DIFF
--- a/lib/std/crypto.zig
+++ b/lib/std/crypto.zig
@@ -109,8 +109,10 @@ pub const pwhash = struct {
         crypt,
     };
     pub const KdfError = errors.Error || std.mem.Allocator.Error;
-    pub const HasherError = KdfError || @import("crypto/phc_encoding.zig").Error;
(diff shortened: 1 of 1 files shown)


jedisct1:
Other serialization formats required for cryptographic operations may be good to have (S2V, constant-time hex/b64, PEM, etc).

So, instead of having all these in the top-level `crypto` namespace, how about a `crypto.codecs` or `crypto.serialization` namespace?

jedisct1:
Well... `phc_format` is really only useful in the context of password hashing functions, so having the relevant codecs exported here is intuitive.

So let's merge this :) And maybe reexport it in a codecs namespace once there is a justification for it.

Thanks!

jedisct1:
(looking at the diff, I originally thought it was in the top-level namespace, but it is actually in `crypto.pwhash`, so... this is fine 😄)

PR #9715, opened 2021-09-09
std: crypto - Fix p256 scalar divStep compile error

Luukdegram:
Seems we had an oopsie in 6dbdac4b605fb56b4180b6c66154275c22954a8d where `@as` was used for a non-coercable type. This PR updates it to use `@truncate` (as we don't care about the other bits).
Also, added a test case that uses divStep.

cc @jedisct1


jedisct1:
Thank you!

PR #9727, opened 2021-09-10
Fix windows.WSAIoctl() OVERLAPPED argument types

PhaseMage:
The std.os.windows.WSAIoctl() function has incorrect types and doesn't compile.

This change removes the old WIN16 version of WSAOVERLAPPED. The correct version is OVERLAPPED (with DWORD_PTR fields instead of DWORD).

Also, this fixes std.os.windows.WSAIoctl to add the required "LP" for WSAOVERLAPPED_COMPLETION_ROUTINE.


andrewrk:
Thanks!

PR #9734, opened 2021-09-11
link/include: fix invalid file path concatenation when cross-compiling for Windows -> Mac

Andoryuuta:
While investigating hexops/mach#8 with @slimsag, we found that zld is forming invalid file paths (absolute paths concatenated together), which hits the unreachable `OBJECT_NAME_INVALID` case in `openDirAccessMaskW`:
https://github.com/ziglang/zig/blob/0c091feb5ae52caf1ebf885c0de55b3159207001/lib/std/fs.zig#L1522

This is caused by appending `dir` (which is guaranteed to be absolute) to `root`, which is not valid on Windows.
https://github.com/ziglang/zig/blob/0c091feb5ae52caf1ebf885c0de55b3159207001/src/link/MachO.zig#L494-L499

Also fixes the same fundamental issue with include directories.

Fixes hexops/mach#8

Co-authored-by: Stephen Gutekanst <stephen@hexops.com>
Signed-off-by: Andrew Gutekanst <andrew.gutekanst@gmail.com>

### DIFF
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -491,7 +491,7 @@ fn resolveSearchDir(
 ) !?[]const u8 {
     var candidates = std.ArrayList([]const u8).init(arena);
 
-    if (fs.path.isAbsolute(dir)) {
+    if (!fs.path.isAbsolute(dir)) {
         if (syslibroot) |root| {
             const full_path = try fs.path.join(arena, &[_][]const u8{ root, dir });
             try candidates.append(full_path);
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -491,9 +491,21 @@ fn resolveSearchDir(
 ) !?[]const u8 {
     var candidates = std.ArrayList([]const u8).init(arena);
 
-    if (!fs.path.isAbsolute(dir)) {
+    if (fs.path.isAbsolute(dir)) {
         if (syslibroot) |root| {
-            const full_path = try fs.path.join(arena, &[_][]const u8{ root, dir });
+            const common_dir = if (std.Target.current.os.tag == .windows) blk: {
+                // We need to check for disk designator and strip it out from dir path so
+                // that we can concat dir with syslibroot.
+                // TODO we should backport this mechanism to 'MachO.Dylib.parseDependentLibs()'
+                const disk_designator = fs.path.diskDesignatorWindows(dir);
+
+                if (mem.indexOf(u8, dir, disk_designator)) |where| {
+                    break :blk dir[where + disk_designator.len ..];
+                }
+
+                break :blk dir;
+            } else dir;
+            const full_path = try fs.path.join(arena, &[_][]const u8{ root, common_dir });
             try candidates.append(full_path);
         }
     }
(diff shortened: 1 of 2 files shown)


kubkon (changes requested):
Thanks for the patch. I think to make this work correctly on Windows however, we will need to come up with a separate strategy for handling syslibroot on Windows. The current patch is incorrect in the sense that it reverses the correct operation on POSIX systems: the purpose of the syslibroot option is to prepend/swap out the typical root of the path with the new root called the `syslibroot`, and this only takes place for absolute paths and never for relative paths. For instance, when linking macOS -> macOS, we auto populate some search paths for system libraries such as `/usr/lib`. But then, this is not enough since this path doesn't exist on its own; it actually exists at a different path root `/Applications/Xcode.app/Contents/System/Libraries/usr/lib` or similar. Here, the syslibroot is `/Applications/Xcode.app/Contents/System/Libraries`, and we prepend it **only** to absolute search paths, never to relative paths - I think this may be by convention that absolute paths are always used to designate search paths for system libraries, while relative paths for user libraries.

One more thing we should have in mind while trying to come up with a better fix for this is that the `syslibroot` might be required when cross-compiling to macOS when bringing your own sysroot with frameworks, etc. Then, we use the `syslibroot` value when resolving dylib auto-exports by path, and since the dylibs will always use POSiX paths, it might be necessary to do some POSIX -> NT translations under-the-hood.

src/link/MachO.zig
```diff
@@ -491,7 +491,7 @@ fn resolveSearchDir(
 ) !?[]const u8 {
     var candidates = std.ArrayList([]const u8).init(arena);
 
-    if (fs.path.isAbsolute(dir)) {
+    if (!fs.path.isAbsolute(dir)) {
```
kubkon, line 494: I was thinking of something along those lines (please feel free to do whatever you see fit with the diff):

```diff
diff --git a/src/link/MachO.zig b/src/link/MachO.zig
index 12556b53d..684f308a8 100644
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -493,7 +493,17 @@ fn resolveSearchDir(
 
     if (fs.path.isAbsolute(dir)) {
         if (syslibroot) |root| {
-            const full_path = try fs.path.join(arena, &[_][]const u8{ root, dir });
+            const common_dir = if (std.Target.current.os.tag == .windows) blk: {
+                // We need to check for disk designator and strip it out from dir path so
+                // that we can concat dir with syslibroot.
+                // TODO we should backport this mechanism to 'MachO.Dylib.parseDependentLibs()'
+                const disk_designator = fs.path.diskDesignatorWindows(dir);
+                if (mem.indexOf(u8, dir, disk_designator)) |where| {
+                    break :blk dir[where..];
+                }
+                break :blk dir;
+            } else dir;
+            const full_path = try fs.path.join(arena, &[_][]const u8{ root, common_dir });
             try candidates.append(full_path);
         }
     }

```

Andoryuuta:
Wow! Thanks so much for explaining this. We couldn't think of any situation where combining two absolute paths would be valid. 

We've updated the code as suggested, and also fixed the same fundamental issue for include directories (`-iwithsysroot`).

kubkon (approved):
Looks great, thanks for the patch @Andoryuuta!

PR #9744, opened 2021-09-12
Fix type error for u8 in writeIntSlice

### LINKED ISSUE #9742: Shift error when trying to use `mem.writeIntSliceLittle` with `u8` truncated value
## System Information

- **Zig version**: `0.9.0-dev.951+9e2472706`
- **Operating System**: Alpine Linux (Edge) running kernel `5.10.16.3-microsoft-standard-WSL2`
- **Example Usage:** https://github.com/xpyxel/cobalt/blob/dev/src/memory.zig#L15 

## Synopsis

I feel like this is a bug, as since both the type of the value, and the type given to the function match, it should properly handle writing the value to the buffer, however instead it errors with the following:

```
/home/pyxel/.local/zig/versions/0.9.0-dev.951+9e2472706/files/lib/std/mem.zig:1450:18: error: integer value 8 cannot be coerced to type 'u3'
        bits >>= 8;
                 ^
./repro.zig:8:32: note: called from here
    std.mem.writeIntSliceLittle(u8, buffer, @truncate(u8, example_variable));
                               ^
./repro.zig:3:29: note: called from here
pub fn main() anyerror!void {
                            ^
/home/pyxel/.local/zig/versions/0.9.0-dev.951+9e2472706/files/lib/std/mem.zig:1450:14: note: referenced here
        bits >>= 8;
```

To reproduce this error you can use the following code:

```zig
const std = @import("std");

pub fn main() anyerror!void {
    const allocator = std.heap.page_allocator;
    var buffer = try allocator.alloc(u8, 65535);
    defer allocator.free(buffer);
    const example_variable: u64 = 1_000_000_000;
    std.mem.writeIntSliceLittle(u8, buffer, @truncate(u8, example_variable));
}
```

Trioct:
Fixes #9742

writeIntSlice wants to shift bits in increments of 8, but the type system doesn't like shifting u8s by u4s
Just set the slice to zero and set the byte that needs to be set to its value to avoid the nonsense

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -1439,9 +1439,13 @@ pub fn writeInt(comptime T: type, buffer: *[@divExact(@typeInfo(T).Int.bits, 8)]
 pub fn writeIntSliceLittle(comptime T: type, buffer: []u8, value: T) void {
     assert(buffer.len >= @divExact(@typeInfo(T).Int.bits, 8));
 
-    if (@typeInfo(T).Int.bits == 0)
+    if (@typeInfo(T).Int.bits == 0) {
         return set(u8, buffer, 0);
-
+    } else if (@typeInfo(T).Int.bits == 8) {
+        set(u8, buffer, 0);
+        buffer[0] = @bitCast(u8, value);
+        return;
+    }
     // TODO I want to call writeIntLittle here but comptime eval facilities aren't good enough
     const uint = std.meta.Int(.unsigned, @typeInfo(T).Int.bits);
     var bits = @bitCast(uint, value);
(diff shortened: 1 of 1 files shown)


lib/std/mem.zig
```diff
@@ -1439,9 +1439,13 @@ pub fn writeInt(comptime T: type, buffer: *[@divExact(@typeInfo(T).Int.bits, 8)]
 pub fn writeIntSliceLittle(comptime T: type, buffer: []u8, value: T) void {
     assert(buffer.len >= @divExact(@typeInfo(T).Int.bits, 8));
 
-    if (@typeInfo(T).Int.bits == 0)
+    if (@typeInfo(T).Int.bits == 0) {
         return set(u8, buffer, 0);
-
+    } else if (@typeInfo(T).Int.bits == 8) {
```
daurnimator, line 1444: Should this be `<= 8`?

Trioct: I don't think so because, like said in the doc comment
```/// The bit count of T must be divisible by 8.```
and the assertion with the @divExact should fail on the bit count of the int being not divisible by 8
```assert(buffer.len >= @divExact(@typeInfo(T).Int.bits, 8));```

File: lib/std/mem.zig [hunk 2]
```diff
@@ -1459,8 +1463,13 @@ pub fn writeIntSliceLittle(comptime T: type, buffer: []u8, value: T) void {
 pub fn writeIntSliceBig(comptime T: type, buffer: []u8, value: T) void {
     assert(buffer.len >= @divExact(@typeInfo(T).Int.bits, 8));
 
-    if (@typeInfo(T).Int.bits == 0)
+    if (@typeInfo(T).Int.bits == 0) {
         return set(u8, buffer, 0);
+    } else if (@typeInfo(T).Int.bits == 8) {
+        set(u8, buffer, 0);
```
daurnimator, line 1469: Do you need to `set` with `255` if `value` is negative?

daurnimator, line 1469: Should this be:
```suggestion
        set(u8, buffer[0..buffer.len-1], 0);
```

Trioct: I hadn't even thought of that, I was just basing my results off the negative i16 test where the rest of the values are set to zero, but maybe they should be set to 0xff. I guess it really depends on what the intention of #8865 was when it was implemented

andrewrk:
Thanks!

PR #9746, opened 2021-09-12
Language reference: define the inferred error set syntax more explicitly

roryokane:
This edit to the [Inferred Error Sets] section allows the reader to understand the syntax the section is talking about more quickly. They don’t have to read the whole code block and then understand which part of it demonstrates the feature being described.

[Inferred Error Sets]: https://ziglang.org/documentation/master/#Inferred-Error-Sets

## Before

> Inferred Error Sets before – “use this syntax”

## After

> Inferred Error Sets after – “prepend the ! operator to the function’s return type, like !T”

The above screenshot was made by editing the documentation page using the browser devtools. I didn’t test this change by compiling the documentation, as I think the change’s simplicity doesn’t warrant setting up a local environment.

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -4955,7 +4955,7 @@ test "merge error sets" {
       {#header_open|Inferred Error Sets#}
       <p>
       Because many functions in Zig return a possible error, Zig supports inferring the error set.
-      To infer the error set for a function, use this syntax:
+      To infer the error set for a function, prepend the {#syntax#}!{#endsyntax#} operator to the function’s return type, like {#syntax#}!T{#endsyntax#}:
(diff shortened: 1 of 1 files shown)


Cons-Cat:
I wonder if it's slightly misleading to tell the reader they need to prepend ``!`` to the return type for inferring the error set. I think it's really the absence of an explicit error set that tells the compilers to infer it, and the ``!`` is what says that an error set exists at all.

paulespinosa:
Good catch @Cons-Cat 

"..., omit the error set from the error union's {#syntax#}!{#endsyntax#} operator, like {#syntax#}!T{#endsyntax#}:"

PR #9748, opened 2021-09-12
Add Linux ioctl creation utilities

Aransentin:
Shamelessly stolen from @mattnite 's now closed #6443 pull request, this adds some utilities for constructing modern IOCTLs, and also uses them in `linux.zig` for some constants that were previously broken on mips, powerpc, and sparc.

From the ioctl manpages: 

> Ioctl command values are 32-bit constants.  In principle these constants are completely arbitrary, but people have tried to build some structure into them.

Therefore I have kept them as u32 values, and not as packed `Request` structures. There previous PR had also messed up the flags a little: `write` is 1 and `read` is 2, not the inverse.


MasterQ32:
This is very nice, i needed this in the past quite often already.

lib/std/os/linux/ioctl.zig
```diff
@@ -0,0 +1,56 @@
+const std = @import("../../std.zig");
+
+const bits = switch (std.Target.current.cpu.arch) {
```
Vexu, line 3:
```suggestion
const bits = switch (@import("builtin").cpu.arch) {
```
`std.Target.current.cpu.arch` is deprecated #9321.

lib/std/os/linux/ioctl.zig
```diff
@@ -0,0 +1,56 @@
...
+    return io_impl(bits.write, io_type, nr, T);
+}
+
+pub fn IOWR(io_type: u8, nr: u8, comptime T: type) u32 {
+    return io_impl(bits.read | bits.write, io_type, nr, T);
+}
+
+test "ioctl request size" {
+    std.testing.expectEqual(32, @bitSizeOf(Request));
+}
```
Vexu, line 56:
```suggestion
comptime {
    std.debug.assert(@bitSizeOf(Request) == 32);
}
```

Vexu (changes requested):

lib/std/os/linux.zig
```diff
@@ -2509,18 +2510,18 @@ pub const T = struct {
     pub const IOCGSID = 0x5429;
     pub const IOCGRS485 = 0x542E;
     pub const IOCSRS485 = 0x542F;
-    pub const IOCGPTN = 0x80045430;
-    pub const IOCSPTLCK = 0x40045431;
-    pub const IOCGDEV = 0x80045432;
+    pub const IOCGPTN = IOCTL.IOR('T', 0x30, c_uint);
+    pub const IOCSPTLCK = IOCTL.IOW('T', 0x31, c_int);
+    pub const IOCGDEV = IOCTL.IOW('T', 0x32, c_uint);
```
zigazeljko, line 2515:
```suggestion
    pub const IOCGDEV = IOCTL.IOR('T', 0x32, c_uint);
```

PR #9754, opened 2021-09-13
fix crypto/benchmark missing paths

x13a:
When adding `scrypt` I have forgotten to update import paths in `crypto/benchmark`. Sorry :(
Plus increase `bcrypt` benchmark rounds to be like "interactive".

/cc @jedisct1


jedisct1:
Looks good!

Thank you!

PR #9755, opened 2021-09-13
Zig fmt: Keep callconv(.Inline) on function pointer types

### LINKED ISSUE #9658: Zig fmt - callconv(.Inline)
`zig fmt` is too aggressive with it transformation of `callconv(.Inline)` -> `inline`

It correctly works on function definitions:
```zig
// Before fmt
pub fn aFunc() callconv(.Inline) void {}

// After fmt
pub inline fn aFunc() void {}
```

However it also incorrectly removes it from function types in other contexts:
```zig
// This is fine
pub fn callCFn(func: fn () callconv(.C) void) void {
    func();
}

// Before fmt
pub fn callInlineFn(func: fn () callconv(.Inline) void) void {
    func();
}

// After fmt
pub fn callInlineFn(func: fn () void) void {
    func();
}
```

luehmann:
Should close #9658.

This is my first PR. Here are some of the things I am not so sure about:

- Is the test I've added a good example/enough and is it in the right place?
- My definition of `is_declaration` is only dependent on whether `fn_proto` has a `name_token`
- I'm unable to run `build test-std -Dskip-release` locally because of a lot of `No such file or directory` messages about files in `lib/libc/wasi/libc-bottom-half`

Thanks for any help/tips etc.!

### DIFF
--- a/lib/std/zig/parser_test.zig
+++ b/lib/std/zig/parser_test.zig
@@ -2879,6 +2879,9 @@ test "zig fmt: functions" {
         \\pub export fn puts(s: *const u8) align(2 + 2) c_int;
         \\pub inline fn puts(s: *const u8) align(2 + 2) c_int;
         \\pub noinline fn puts(s: *const u8) align(2 + 2) c_int;
+        \\pub fn callInlineFn(func: fn () callconv(.Inline) void) void {
+        \\    func();
+        \\}
         \\
     );
 }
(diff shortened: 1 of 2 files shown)


marler8997:
An inline function pointer is nonsensical, this should be a compile error.

g-w1:
https://github.com/ziglang/zig/issues/9658#issuecomment-908694774

marler8997:
@g-w1 yeah I suppose a comptime inline function pointer could make sense, however, I think one could make an argument that declaring the calling convention for a comptime-only function has no utility.  So an inline function pointer has no utility at runtime, and no reason to exist at comptime as far as I can see.

luehmann:
@marler8997 true, but zig fmt shouldn't make invalid code valid, right?

marler8997:
> @marler8997 true, but zig fmt shouldn't make invalid code valid, right?

Oh right sorry, this is just for zig fmt.

ifreund (approved):
LGTM, Thanks!

PR #9761, opened 2021-09-14
fix __chkstk on aarch64

marler8997:
This is a follow up on changes to support `aarch64-windows`.  A previous change added the missing `__chkstk` symbol, but that symbol wasn't hooked up properly in `stack_probe.zig`, which resulted in the implementation just being an `unreachable`.  Now it's hooked up properly and it's been tested through a dll that was compiled for and then executed on a Hololens 2 platform.  

Note, the `__chkstk` function is executed during `alloca` which is called during `DllMainCRTStartup` in the mingw-w64 implementation.

### DIFF
--- a/lib/std/special/compiler_rt/stack_probe.zig
+++ b/lib/std/special/compiler_rt/stack_probe.zig
@@ -53,19 +53,6 @@ pub fn zig_probe_stack() callconv(.Naked) void {
         },
         else => {},
     }
-    if (comptime native_arch.isAARCH64()) {
-        asm volatile (
-            \\        lsl    x16, x15, #4
-            \\        mov    x17, sp
-            \\1:
-            \\        sub    x17, x17, #PAGE_SIZE
-            \\        subs   x16, x16, #PAGE_SIZE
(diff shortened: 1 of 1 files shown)


lib/std/special/compiler_rt/stack_probe.zig
```diff
@@ -118,6 +105,21 @@ fn win_probe_stack_only() void {
...
         else => {},
     }
+    if (comptime native_arch.isAARCH64()) {
+        // NOTE: page size hardcoded to 4096 for now
+        asm volatile (
+            \\        lsl    x16, x15, #4
+            \\        mov    x17, sp
+            \\1:
+            \\
+            \\        sub    x17, x17, 4096
```
marler8997, line 115: Note, I'm not sure how to pass `std.mem.page_size` here for the immediate operand to this instruction and the next one.

marler8997:
Looks like azure pipeline has fallen victim to the notorious Windows issue where you sometimes can't remove executables that were recently just exected.  I've pushed a fix for that here: https://github.com/ziglang/zig/pull/9763

PR #9769, opened 2021-09-15
Fix Thread.Condition on Windows

### LINKED ISSUE #9077: std.Thread.Condition: error: no member named 'srwlock' in struct 'std.Thread.Mutex'
semaphore.zig
```
const std = @import("std");
pub fn main() !void {
    var sem = std.Thread.Semaphore{};
    sem.wait();
}
```

`zig build-exe semaphore.zig`
```
\zig\build-release\lib\zig\std\Thread\Condition.zig:57:19: error: no member named 'srwlock' in struct 'std.Thread.Mutex'
            &mutex.srwlock,
```

OS: Windows 10
Zig version: 0.9.0-dev.70+a3f7a48d9

I also noticed running the same code on Godbolt 0.8.0 gives a different error:
```
<Compilation failed>
# Compiler exited with result code 1
Standard error:
/opt/compiler-explorer/zig-0.8.0/lib/std/Thread/Condition.zig:154:14: error: no member named 'unlock' in struct 'std.Thread.Mutex'
        mutex.unlock();
             ^
```

Kirjastonhoitaja:
Fixes #9077

The use of 'impl' here seems justified, the implementation of Condition
variables fundamentally relies on the Mutex implementation. The same can
be seen in the PthreadCondition, which directly uses 'impl' as well.

### DIFF
--- a/lib/std/Thread/Condition.zig
+++ b/lib/std/Thread/Condition.zig
@@ -54,7 +54,7 @@ pub const WindowsCondition = struct {
     pub fn wait(cond: *WindowsCondition, mutex: *Mutex) void {
         const rc = windows.kernel32.SleepConditionVariableSRW(
             &cond.cond,
-            &mutex.srwlock,
+            &mutex.impl.srwlock,
             windows.INFINITE,
             @as(windows.ULONG, 0),
         );


Kirjastonhoitaja:
I see this is a duplicate of #9729. Don't mind me, then.

PR #9770, opened 2021-09-15
net.Address: Fix writing 0-bytes when formatting Unix addresses

Kirjastonhoitaja:
The entire 'path' array would get written to the formatting function,
when it should instead be treated as a regular zero-terminated string.

Note that this doesn't handle abstract paths on Linux, those paths
*start* with a \0 byte and are hence treated as empty strings instead.
But fixing that would require more adjustments than just formatting, in
particular to getOsSockLen().

### DIFF
--- a/lib/std/net.zig
+++ b/lib/std/net.zig
@@ -157,7 +157,7 @@ pub const Address = extern union {
                     unreachable;
                 }
 
-                try std.fmt.format(out_stream, "{s}", .{&self.un.path});
+                try std.fmt.format(out_stream, "{s}", .{std.mem.sliceTo(&self.un.path, 0)});
             },
             else => unreachable,
         }
--- a/lib/std/net/test.zig
+++ b/lib/std/net/test.zig
@@ -90,6 +90,19 @@ test "parse and render IPv4 addresses" {
     try testing.expectError(error.NonCanonical, net.Address.parseIp4("127.01.0.1", 0));
 }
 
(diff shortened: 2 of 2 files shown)


lib/std/net.zig
```diff
@@ -157,7 +157,7 @@ pub const Address = extern union {
                     unreachable;
                 }
 
-                try std.fmt.format(out_stream, "{s}", .{&self.un.path});
+                try std.fmt.format(out_stream, "{s}", .{std.mem.sliceTo(std.meta.assumeSentinel(&self.un.path, 0), 0)});
```
ifreund, line 160: This `assumeSentinel()` call should not be needed, if it is I made a mistake in the implementation of `sliceTo()`
```suggestion
                try std.fmt.format(out_stream, "{s}", .{std.mem.sliceTo(&self.un.path, 0)});
```

ifreund (changes requested):
It would be great to have a test case so that this doesn't regress.

lib/std/net.zig
Kirjastonhoitaja: You're right, works fine without.

Kirjastonhoitaja:
I'm clearly not familiar with Github PR review stuff, but 997778eb1d276a9ce0eecdcdcc19d6891fd65317 has the requested changes. :)

ifreund (approved):
LGTM, thanks!

PR #9771, opened 2021-09-15
link: windows-gnu: look for Unix-style static library names

Kha:
Zig already looks for MinGW-style import library names `lib${name}.dll.a`, so this feels only consistent and simplifies porting existing MinGW projects.

### DIFF
--- a/src/link/Coff.zig
+++ b/src/link/Coff.zig
@@ -1297,8 +1297,13 @@ fn linkWithLLD(self: *Coff, comp: *Compilation) !void {
                 continue;
             }
             if (target.abi.isGnu()) {
-                const fallback_name = try allocPrint(arena, "lib{s}.dll.a", .{key});
+                const fallback_name = try allocPrint(arena, "lib{s}.a", .{key});
                 if (try self.findLib(arena, fallback_name)) |full_path| {
+                    argv.appendAssumeCapacity(full_path);
+                    continue;
+                }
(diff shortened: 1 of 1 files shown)


rohlem:
IIUC (and I might be uninformed), when both static library `.a` and dynamic import library `.dll.a` are present, this change will prioritize the static library, changing current behaviour.

Is that intentional/desirable? From the little I know, some licenses are more strict when statically linking.
Also unintentional dynamic linking (-> missing DLL) seems easier to spot than unintentional static linking (-> DLL shipped but unused).

Kha:
Well, I'd be happy to swap the two. But the current behavior is that first Windows-style `.lib` static libraries are search, which could reasonably be either "true" static libraries or import libraries, so the search order isn't clear to begin with.

andrewrk (changes requested):
Needs a test.

Note also this relevant commit: ba127058d1fcf3ca042794244eee052915998b1a

These are supposed to be dynamic libraries in this for loop, not static libraries.

andrewrk:
Thanks for the patch. I am sorry I did not get around to reviewing it for many months.

I am closing this because it is not merge-ready and it is very old. However, you are welcome to open a new PR if you wish to continue this effort. I think you should first start with an issue to track the use case this is trying to address. This way if the PR is not merged (as in this one) the issue will persist.

PR #9777, opened 2021-09-16
std.ArrayList: Implement initCapacity using ensureTotalCapacity

### LINKED ISSUE #9775: std.ArrayList: difference between initCapacity and ensureTotalCapacity
Currently, `std.ArrayList` does two different things when called with `initCapacity` vs being called with `init` and then calling `ensureTotalCapacity` afterwards:

https://github.com/ziglang/zig/blob/19691c0b174f283ffe5b6c3fe8533ef458736064/lib/std/array_list.zig#L60-L73

https://github.com/ziglang/zig/blob/19691c0b174f283ffe5b6c3fe8533ef458736064/lib/std/array_list.zig#L323-L340

I'm unsure about why this would be, or if there's any reason why `initCapacity` isn't implemented as

```zig
        pub fn initCapacity(allocator: *Allocator, num: usize) !Self {
            var self = Self.init(allocator);
            try self.ensureTotalCapacity(num);
            return self;
        }
```

squeek502:
This is one way to address #9775. Just thought I'd make it a pull request if there happens to be nothing wrong with this change.

---

- Removes (divergent) duplicate code
- Makes initCapacity use the `better_capacity` algorithm found in ensureTotalCapacity (which I'm assuming exists for a reason)

Closes #9775

### DIFF
--- a/lib/std/array_list.zig
+++ b/lib/std/array_list.zig
@@ -59,16 +59,7 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
         /// Deinitialize with `deinit` or use `toOwnedSlice`.
         pub fn initCapacity(allocator: *Allocator, num: usize) !Self {
             var self = Self.init(allocator);
-
-            if (@sizeOf(T) > 0) {
-                const new_memory = try self.allocator.allocAdvanced(T, alignment, num, .at_least);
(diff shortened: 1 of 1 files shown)


squeek502:
Closing in light of https://github.com/ziglang/zig/issues/9775#issuecomment-921316033, will submit a new PR with a more appropriate change.

PR #9781, opened 2021-09-16
Initial port to Illumos

The-King-of-Toasters:
This patch adds Tier 2 support for Illumos-based operating systems. Illumos is the spiritual successor to the OpenSolaris project that was shut down by Oracle in the early 2010s. You can think of Illumos as the base OS, as it provides the kernel and userland tools. Other distributions then add their own features atop of it:

- SmartOS, which is dedicated for Virtual Machines and Containers.
- OmniOS CE which is developed as a stable server OS, ala Debian.

The system I'm using to test is an OmniOS CE QEMU host. I've been testing against LLVM 12.0.1 with some small patches to fix header/configure differences against Solaris. I intend to provide a walkthrough of how to build it yourself, and also provide a tarball for convenience.

As for functionality, most things work fine. The stdlib, compiler, and behaviour tests pass. But I must stress, this is still very much a WIP. Here's what needs to be fixed:

1. Event loops are unimplemented. It's possible to re-use the existing epoll implementation as Illumos provides an epoll shim, but it requires changing existing code to not use the Linux types and to expose epoll types/functions under `os`. It'd probably be better to use the "blessed" event system introduced in Solaris 10: [event ports](https://illumos.org/man/3c/port_create). I would appreciate some guidance in this area, as I haven't written event-driven code before.
2. Right now, the port uses the `solaris` os tag, which was inherited from LLVM. I think going forward it'd be best to differentiate between the two, as 1) Illumos is receiving more support/updates than Solaris and 2) it's actually open source, which means that if we find any issues we can inspect the kernel/libc source code. Right now, the gcc compilers on OmniOS expose the `__illumos__` CPP definition, which can be used to differentiate against the standard `__sun__`. I'm unsure of any other way do to this when the self-hosted compiler gets finished.
3. Static linking stopped being supported after Solaris 10, and there is no static libc on x86_64. We'll probably need to add some logic to address this.
4. Related to the above, multithreaded C code must provide the `_REENTRANT` CPP define to get threaded versions of libc functions/types etc. AFAIK there's no way to do this in the codebase.

Some short term tasks after merging:
- Setup sourcehut CI support for OmniOS, so that we can have a Lavahut target.
- It'd be nice to use our connections to the LLVM project to enforce point 4 above for there as well. As mentioned, building it requires some patches due to things that the Illumos devs changed (for the better). An example: libcxx provides a support shim around the nasty Solaris locale headers. The header it includes was recently removed in Illumos, as the actual locale system is completely different under the hood. It also means that we fail tests that link with unpatched libcxx.

Long term tasks:
- Solaris introduced the slab allocator, and it's possible to use the kernel memory allocators directly in userland (see [kmem_alloc](https://illumos.org/man/9f/kmem_alloc)). There's also the libumem library provides wrappers around the kernel allocators. It would be nice to expose this as a `mem.Allocator` in the future.
- Might be interesting to port the libc over. That's a big "might" though.

### DIFF
--- a/lib/std/Thread.zig
+++ b/lib/std/Thread.zig
@@ -41,6 +41,7 @@ pub const max_name_len = switch (target.os.tag) {
     .netbsd => 31,
     .freebsd => 15,
     .openbsd => 31,
+    .solaris => 31,
     else => 0,
 };
 
@@ -112,7 +113,7 @@ pub fn setName(self: Thread, name: []const u8) SetNameError!void {
                 else => |e| return os.unexpectedErrno(e),
             }
         },
-        .netbsd => if (use_pthreads) {
+        .netbsd, .solaris => if (use_pthreads) {
             const err = std.c.pthread_setname_np(self.getHandle(), name_with_terminator.ptr, null);
             switch (err) {
                 .SUCCESS => return,
@@ -202,7 +203,7 @@ pub fn getName(self: Thread, buffer_ptr: *[max_name_len:0]u8) GetNameError!?[]co
                 else => |e| return os.unexpectedErrno(e),
             }
         },
-        .netbsd => if (use_pthreads) {
+        .netbsd, .solaris => if (use_pthreads) {
             const err = std.c.pthread_getname_np(self.getHandle(), buffer.ptr, max_name_len + 1);
             switch (err) {
                 .SUCCESS => return std.mem.sliceTo(buffer, 0),
@@ -564,6 +565,16 @@ const PosixThreadImpl = struct {
                     else => |e| return e,
                 };
                 return @intCast(usize, count);
+            },
+            .solaris => {
+                // The "proper" way to get the cpu count would be to query
+                // /dev/kstat via ioctls, and traverse a linked list for each
+                // cpu.
+                const rc = c.sysconf(os._SC.NPROCESSORS_ONLN);
+                return switch (os.errno(rc)) {
+                    .SUCCESS => @intCast(usize, rc),
+                    else => |err| os.unexpectedErrno(err),
+                };
             },
             .haiku => {
                 var count: u32 = undefined;
--- a/lib/std/c.zig
+++ b/lib/std/c.zig
@@ -252,6 +252,33 @@ pub extern "c" fn kevent(
     timeout: ?*const c.timespec,
 ) c_int;
 
+pub extern "c" fn port_create() c.port_t;
+pub extern "c" fn port_associate(
+    port: c.port_t,
+    source: u32,
+    object: usize,
+    events: u32,
+    user_var: ?*c_void,
+) c_int;
+pub extern "c" fn port_dissociate(port: c.port_t, source: u32, object: usize) c_int;
+pub extern "c" fn port_send(port: c.port_t, events: u32, user_var: ?*c_void) c_int;
+pub extern "c" fn port_sendn(
+    ports: [*]c.port_t,
+    errors: []u32,
+    num_ports: u32,
+    events: u32,
+    user_var: ?*c_void,
+) c_int;
+pub extern "c" fn port_get(port: c.port_t, event: *c.port_event, timeout: ?*c.timespec) c_int;
+pub extern "c" fn port_getn(
+    port: c.port_t,
+    event_list: []c.port_event,
+    max_events: u32,
+    events_retrieved: *u32,
+    timeout: ?*c.timespec,
+) c_int;
+pub extern "c" fn port_alert(port: c.port_t, flags: u32, events: u32, user_var: ?*c_void) c_int;
+
 pub extern "c" fn getaddrinfo(
     noalias node: ?[*:0]const u8,
     noalias service: ?[*:0]const u8,
--- a/lib/std/c/solaris.zig
+++ b/lib/std/c/solaris.zig
@@ -1,15 +1,1913 @@
+const std = @import("../std.zig");
+const builtin = @import("builtin");
+const maxInt = std.math.maxInt;
+const iovec = std.os.iovec;
+const iovec_const = std.os.iovec_const;
+const timezone = std.c.timezone;
+
+extern "c" fn ___errno() *c_int;
+pub const _errno = ___errno;
+
+pub const dl_iterate_phdr_callback = fn (info: *dl_phdr_info, size: usize, data: ?*c_void) callconv(.C) c_int;
+pub extern "c" fn dl_iterate_phdr(callback: dl_iterate_phdr_callback, data: ?*c_void) c_int;
+
+pub extern "c" fn getdents(fd: c_int, buf_ptr: [*]u8, nbytes: usize) usize;
+pub extern "c" fn sigaltstack(ss: ?*stack_t, old_ss: ?*stack_t) c_int;
+pub extern "c" fn pipe2(fds: *[2]fd_t, flags: u32) c_int;
+pub extern "c" fn arc4random_buf(buf: [*]u8, len: usize) void;
+pub extern "c" fn posix_memalign(memptr: *?*c_void, alignment: usize, size: usize) c_int;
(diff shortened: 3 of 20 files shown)


daurnimator:
>     4\. Event loops are unimplemented. It's possible to re-use the existing epoll implementation as Illumos provides an epoll shim, but it requires changing existing code to not use the Linux types and to expose epoll types/functions under `os`. It'd probably be better to use the "blessed" event system introduced in Solaris 10: [event ports](https://illumos.org/man/3c/port_create). I would appreciate some guidance in this area, as I haven't written event-driven code before.

In https://github.com/wahern/cqueues/blob/master/src/lib/kpoll.c there is a layer to unify behaviour of kqueue/epoll/ports. You can likely consult that.

The-King-of-Toasters:
Also note that this patchset is contingent on #9767 being merged, too.

The-King-of-Toasters:
@jclulow on the Illumos IRC [commented the the following nuggets of info](https://log.omnios.org/illumos/2021-09-16#1631779479-152446):

```
[18:05:15] <jclulow> To be clear, libumem is actually just a library version of allocators similar to the kernel bits. It doesn't make direct use of the kernel facility
[18:05:40] <jclulow> But I think most of the code is common
[18:06:14] <jclulow> Also, "/usr/bin/uname -o" reports "illumos" even though uname still says SunOS, if you're looking for other ways to tell us apart from Solaris
[18:07:08] <jclulow> wrt. to Zig (and LLVM), we would definitely like to be identified as illumos separately from Solaris for all the reasons you highlight.
[18:07:40] <jclulow> We've managed to get there with Rust and Go at least, which has been very helpful so far.
```

The-King-of-Toasters:
It appears that the "nasty race condition" isn't that big a deal after all. I realised that the errors resulted from following symlinks to files that were already deleted. All I needed was to pass `os.AT.SYMLINK_NOFOLLOW` to fstatat, similar to `Dir.deleteFile`.

The-King-of-Toasters:
An issue that I forgot to mention relates to stack traces. For some reason, the stack trace ends up at the null pointer:
```
thread 1 panic: foobar
/home/stephen/zig/build/test.zig:4:20: 0x20432c in main (test)
    std.debug.panic("foobar", .{});
                   ^
/home/stephen/zig/lib/std/start.zig:515:37: 0x206308 in std.start.callMain (test)
                                    ^
/home/stephen/zig/lib/std/start.zig:457:12: 0x204517 in std.start.callMainWithArgs (test)
           ^
/home/stephen/zig/lib/std/start.zig:422:12: 0x204455 in std.start.main (test)
           ^
???:?:?: 0x2035e6 in ??? (???)
???:?:?: 0x203547 in ??? (???)
???:?:?: 0x0 in ??? (???)
```

The-King-of-Toasters:
I've spent some more time on this, and I think I'm ready for first review.
There are still some errors I'm encountering, but none that significantly affects the core language functions.

## Errors
### Backtraces

As mentioned in the comment above, stack-traces will end up at 0x0. It gets worse when using the segfault handler:
  ```
Segmentation fault at address 0x418
???:?:?: 0xfffffc7feeda89ca in ??? (???)
???:?:?: 0xfffffc7feed3c55d in ??? (???)
/home/stephen/zig/lib/std/os.zig:367:35: 0x2044ea in std.os.raise (test)
      switch (errno(system.raise(sig))) {
                                ^
/home/stephen/zig/build/test.zig:7:21: 0x204495 in main (test)
  try std.os.raise(std.os.SIG.SEGV);
                  ^
/home/stephen/zig/lib/std/start.zig:515:37: 0x206568 in std.start.callMain (test)
                                  ^
/home/stephen/zig/lib/std/start.zig:457:12: 0x204747 in std.start.callMainWithArgs (test)
         ^
/home/stephen/zig/lib/std/start.zig:422:12: 0x204685 in std.start.main (test)
         ^
???:?:?: 0x203746 in ??? (???)
???:?:?: 0x2036a7 in ??? (???)
thread 1 panic: integer overflow
/home/stephen/zig/lib/std/debug.zig:142:69: 0x22c478 in std.debug.dumpStackTraceFromBase (test)
                                                                  ^
/home/stephen/zig/lib/std/debug.zig:1575:35: 0x223f9a in std.debug.handleSegfaultLinux (test)
          dumpStackTraceFromBase(bp, ip);
                                ^
???:?:?: 0xfffffc7feeda1985 in ??? (???)
???:?:?: 0xfffffc7feed9478a in ??? (???)
???:?:?: 0xfffffc7feed94abd in ??? (???)
???:?:?: 0xfffffffffffffffe in ??? (???)
???:?:?: 0xfffffc7feed3c55d in ??? (???)
/home/stephen/zig/lib/std/os.zig:367:35: 0x2044ea in std.os.raise (test)
      switch (errno(system.raise(sig))) {
                                ^
/home/stephen/zig/build/test.zig:7:21: 0x204495 in main (test)
  try std.os.raise(std.os.SIG.SEGV);
                  ^
/home/stephen/zig/lib/std/start.zig:515:37: 0x206568 in std.start.callMain (test)
                                  ^
/home/stephen/zig/lib/std/start.zig:457:12: 0x204747 in std.start.callMainWithArgs (test)
         ^
/home/stephen/zig/lib/std/start.zig:422:12: 0x204685 in std.start.main (test)
         ^
???:?:?: 0x203746 in ??? (???)
???:?:?: 0x2036a7 in ??? (???)
???:?:?: 0x0 in ??? (???)
[1]    1048 IOT instruction (core dumped)  ./zig run test.zig
```

I'm not sure if this is a logic issue on our end or on the Illumos impl of `dl_iterate_phdr` ([here's the implementation](https://github.com/illumos/illumos-gate/blob/master/usr/src/cmd/sgs/rtld/common/dlfcns.c#L2199) if you'd like to go digging), but removing all the occurrences of `address - 1` stopped this from happening. I asked about this in the Illumos IRC, and was pointed to the [`backtrace` crate implementation](https://github.com/rust-lang/backtrace-rs/blob/master/src/symbolize/gimli/libs_illumos.rs), which uses `dlopen`.

### `translate-c`
The "assign enum to uint, no explicit cast" test in `test-run-translated-c` fails. The (snipped) output of translate-c is:
```zig
pub const ENUM_0: c_int = 0;
pub const ENUM_384: c_int = 384;
pub const my_enum_t = c_uint;
pub export fn main() c_int {
	  var val: my_enum_t = @bitCast(c_uint, ENUM_384);
	  var x: i8 = @bitCast(i8, @truncate(u8, val));
	  if (@bitCast(c_int, @as(c_uint, x)) != @bitCast(c_int, @as(c_uint, @bitCast(i8, @truncate(i8, @as(c_int, 384)))))) {
		  abort();
	  }
	  return 0;
}
```

## Unimplemented Functionality
### Event Loops
I've decided to postpone implementing event loops with event ports until I have a better understanding of the existing code's control flow. Again, some help in this area would be much appreciated.

## Blocking Changes
In no particular order:
- #9793 fixes an issue where resolving scope IDs wouldn't work.

andrewrk:
Nice work! I'm happy to merge this as-is.

> 1. Event loops are unimplemented.

It's OK to leave this unimplemented for now. Event loops in zig std lib are still experimental on all operating systems at this point.

> 2\. `solaris` os tag

Let's merge this as-is, but you are welcome to file a follow-up patch which adds `illumos` to the list of operating systems in `lib/std/Target.zig`, in between `plan9` and `other`.

> 3\. no static libc

This is fine - nearly every posix OS besides Linux already works this way, so there is no new problem to solve here.

You can simply add `illumos` to the `true` section of this function:

https://github.com/ziglang/zig/blob/a032fd01e88be2c6e8d0cfb0ccd3d9859c9dffdc/lib/std/target.zig#L377

> 4\. multithreaded C code must provide the `_REENTRANT` CPP define

In Compilation.zig addCCArgs, it should check if `comp.bin_file.options.single_threaded` is `false` and if so then add the define to `argv`.

PR #6325, opened 2020-09-11
stage1: implement --omit-lib-prefix

ifreund:
When building shared libraries for use as plugins, it is often necessary to follow different naming conventions. Passing the new --omit-lib-prefix flag prevents zig from prefixing the name of generated libraries with "lib" and makes following such conventions possible.

~Consider this an alternative to #6317~
Closes #2231

PR #9783, opened 2021-09-16
Add option to set exact install name for LibExeObjStep

### LINKED ISSUE #2231: Way to set shared library filename
When creating a shared library for use as a plugin in existing programs/languages, you often need to follow certain naming conventions, e.g. `myprogram-plugin-b.so`.

I can't seem to find a way to set the output filename from `zig build-lib -dynamic`. It seems to come out as `libNAME.so.x.y.z` where NAME can be set with `--name` and `x`/`y`/`z` can come from the `--ver` flags.

Possibly related to #2230

dweiller:
This should close #2231.

This change adds an optional exact_install_filename field to LibExeObjStep that is used to set the exact filename when installing. Andrew suggested using -femit-bin in #6325, but this seems to move the file outside the cache if you just append something like -femit-bin=my-exact-bin-name to the compiler args. Not setting -femit-bin means that the contents of the cache should be unaffected by this change (i.e. the name in the cache is the standard one) and only the install location (as given by the InstallArtifactStep) is changed.

If exact_install_filename is set for a versioned dynamic library, the InstallArtifactStep for it will not install the versioned symlinks. I thought this would be the preferred behaviour - you are wanting non-standard filenames so the standard versioned symlinks are probably wrong anyway. The versioned symlinks do appear in the cache.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -1418,6 +1418,7 @@ pub const LibExeObjStep = struct {
     kind: Kind,
     major_only_filename: ?[]const u8,
     name_only_filename: ?[]const u8,
+    exact_install_filename: ?[]const u8 = null,
     strip: bool,
     lib_paths: ArrayList([]const u8),
     rpaths: ArrayList([]const u8),
@@ -2754,7 +2755,9 @@ pub const LibExeObjStep = struct {
             }
         }
 
-        if (self.kind == .lib and self.linkage != null and self.linkage.? == .dynamic and self.version != null and self.target.wantSharedLibSymLinks()) {
+        if (self.exact_install_filename == null and self.kind == .lib and self.linkage != null and
+            self.linkage.? == .dynamic and self.version != null and self.target.wantSharedLibSymLinks())
+        {
             try doAtomicSymLinks(builder.allocator, self.getOutputSource().getPath(builder), self.major_only_filename.?, self.name_only_filename.?);
         }
     }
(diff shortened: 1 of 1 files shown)


ifreund:
For reference, I solved my use case for #2231 with a custom install step: https://github.com/ifreund/rundird/blob/8a97db7d6d7a07cd0ed24fe0e9604360c4c715e4/build.zig#L41-L63

dweiller:
At the moment there is some non-declarative/order dependent interactions in the build system. At the moment `install()` needs to be called on a `LibExeObjStep` after anything that affects the where the file should be installed because the installation paths are set in the `InstallArtifactStep` when `install()` is called. Perhaps `install()` could instead cause the `make()` function of `LibExeObjStep` configure the `InstallArtifactStep`, and have `InstallArtifactStep` be a union of a configured and not-configured state.

I remember reading in either the source code, or an issue here some time ago about some other order-dependent behavior, but I cannot find it anymore - are there other order-dependent parts of the build system that could be cleaned up that anyone knows of? Perhaps this should be spun out as a separate issue to clean up any unwanted order-dependence in the build system, but I could also remove this interaction of `install()` in this PR.

andrewrk (changes requested):
Thanks for the patch.

Unfortunately I think this is a tricky enough case that we need a test for it to verify the behavior. I'm also a little unsure about the solution to install() being non-declarative. I want to investigate that myself.

I do apologize for the late review. Given that this is old, and there are now merge conflicts, I'm going to close this PR. If you are interested in pursuing this change, please open a new one, and I will make sure to give it more attention. Otherwise, sit tight, the issue #2231 is still open (although it is not set to a high priority milestone).

PR #9785, opened 2021-09-16
Update `crypto` & `hash` lib functions argument name

AliChraghi:
Update `crypto` & `hash` lib functions argument name and update benchmarks run command

### DIFF
--- a/lib/std/crypto/benchmark.zig
+++ b/lib/std/crypto/benchmark.zig
@@ -1,4 +1,4 @@
-// zig run benchmark.zig --release-fast --zig-lib-dir ..
+// zig run -O ReleaseFast --zig-lib-dir ../.. benchmark.zig
 
 const std = @import("../std.zig");
 const builtin = std.builtin;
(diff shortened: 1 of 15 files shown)


jedisct1:
While `d` is not very self-descriptive (not sure where that comes from), `self` is also way too generic.

`st`, for example, is used to represent a state. `ctx` is a context (which can be replaced a state). So looking at a short code snippet is enough to understand what it does, without having to scroll up and check what `self` is.

The fact that, unlike JavaScript, `Self` doesn't have to be named a specific way, and that `x.y()` just sets the first parameter of `x()` to `y`, is a strength of Zig IMHO. It greatly improves readability.

So, instead of renaming everything to `self`, maybe it would be better to:

1) ensure that the same thing is named the same way everywhere
2) rename things such as `d`, either to `self` if the context is obvious, or to something that better represents what that first function parameter represents otherwise.

AliChraghi:
okay. i'll draft this PR and open up when it's ready

PR #9788, opened 2021-09-17
War on ensureCapacity: update all deprecated usage

squeek502:
- Deprecates `ensureCapacity` and adds `ensureTotalCapacity`/`ensureUnusedCapacity` variants to `LinearFifo`/`PriorityQueue`/`PriorityDequeue` (same sort of thing as c8ae581fef6506a8234cdba1355ba7f0f449031a and subsequent changes to `MultiArrayList`, `HashMap`, etc)
  + The only remaining `ensureCapacity` is in `std/math/big/int.zig` which I left alone since it doesn't seem to benefit from an `Unused` version (none of the `ensureCapacity` calls would be able to be updated to an `Unused` version AFAICT), so I wasn't sure what to do.
- Updates all `ensureCapacity` usages across all source files to use the relevant `Total`/`Unused` version instead

### DIFF
--- a/lib/std/child_process.zig
+++ b/lib/std/child_process.zig
@@ -195,7 +195,7 @@ pub const ChildProcess = struct {
         };
 
         var dead_fds: usize = 0;
-        // We ask for ensureCapacity with this much extra space. This has more of an
+        // We ask for ensureTotalCapacity with this much extra space. This has more of an
         // effect on small reads because once the reads start to get larger the amount
         // of space an ArrayList will allocate grows exponentially.
         const bump_amt = 512;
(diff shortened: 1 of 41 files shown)


lib/std/child_process.zig
```diff
@@ -276,7 +276,7 @@ pub const ChildProcess = struct {
 
         // Windows Async IO requires an initial call to ReadFile before waiting on the handle
         for ([_]u1{ 0, 1 }) |i| {
-            try outs[i].ensureCapacity(bump_amt);
+            try outs[i].ensureTotalCapacity(bump_amt);
```
ifreund, line 279: Is this correct? `bump_amt` is used as a delta not a total capacity everywhere else in this file it seems...

ifreund (commented):
Just found the one fishy line, the rest looks great!

lib/std/child_process.zig
squeek502: Yeah, I think you're right. Note that this was incorrect before as well (`ensureCapacity` is an alias for `ensureTotalCapacity`).

squeek502: Actually, since it operates separate ArrayLists on each iteration (`out[0]` and then `out[1]`) it doesn't matter in this case. However, I think it probably should use the same logic as `collectOutputPosix` taking `max_output_bytes` into account:

```zig
const new_capacity = std.math.min(outs[i].items.len + bump_amt, max_output_bytes);
try outs[i].ensureTotalCapacity(new_capacity);
```

squeek502:
CI failure does not look related: `Test [67/1041] behavior.atomics.test "behavior-native-ReleaseFast-bare-single cmpxchg on a global variable"... FAIL (TestUnexpectedResult)`

ifreund (approved):
LGTM, thanks!

PR #9793, opened 2021-09-18
Resolve scope IDs using AF_INET sockets

The-King-of-Toasters:
On certain systems (Solaris), resolving the scope id from an interface
name can only be done on AF_INET-domain sockets. While we're here,
simplify the test while we're here, since there's only one address.

Also note that the loopback interface name is not stable across OSs.
BSDs and Solaris use `lo0` whilst Linux uses `l0`.


andrewrk:
Thanks!

PR #9800, opened 2021-09-19
Linux/sparc64 bits: Add missing C type definitions

koachan:
This makes stage1 Zig buildable again on SPARC.


andrewrk:
Thank you

PR #9803, opened 2021-09-19
readILEB128: Check remaining bits on last byte of negative number even if we don't overflow

malcolmstill:
I'm working on [a WebAssembly interpreter in zig](https://github.com/malcolmstill/foxwren). WebAssembly uses LEB128 encoding throughout its specification.

The WebAssembly spec requires signed LEB128 to be encoded up to a maximum number of bytes (max 5 bytes for i32, max 10 bytes for i64) and that "unused" bits are all 0 if the number is positive and all 1 if the number is negative. The Zig LEB128 implementation already enforces the max number of bytes and does check the unused bytes https://github.com/ziglang/zig/blob/master/lib/std/leb128.zig#L70-L79.

However, the WebAssembly test suite has a number of tests that were failing validation (expecting the wasm module to fail validation, but when running the tests, those examples were actually passing validation):

- https://github.com/malcolmstill/foxwren/blob/master/test/testsuite/binary-leb128.wast#L893-L902
- https://github.com/malcolmstill/foxwren/blob/master/test/testsuite/binary-leb128.wast#L934-L943

Notably the failures are both cases of negative numbers and the top 4 bits of the last byte are zero. And I believe this is the issue: we're only currently checking the "unused" / `remaining_bits` if we overflow, but in the case of `0x0_` no overflow happens and so the bits go unchecked.

In other words:
- `\xff\xff\xff\xff\7f` rightly successfully decodes (because it overflows and the remaining bits are `0b1111`)
- `\xff\xff\xff\xff\6f` rightly errors with overflow (because it overflows and the remaining bits are `0b1110`)
- `\xff\xff\xff\xff\0f` incorrectly decodes when it should error (because the top 4 bits are all 0, and so no overflow occurs and no check that the unused bits are 1 happens)

This PR adds a the `remaining_bits` check in an `else` branch of the `@shlWithOverflow` when we're looking at the last byte and the number being decoded is negative.

Note: this means a couple of the test cases in `leb128.zig` that are down as decoding shouldn't actually decode so I added the appropriate 1 bits.

### DIFF
--- a/lib/std/leb128.zig
+++ b/lib/std/leb128.zig
@@ -76,6 +76,14 @@ pub fn readILEB128(comptime T: type, reader: anytype) !T {
             const remaining_shift = @intCast(u3, @typeInfo(U).Int.bits - @as(u16, shift));
             const remaining_bits = @bitCast(i8, byte | 0x80) >> remaining_shift;
             if (remaining_bits != -1) return error.Overflow;
+        } else {
+            // If we don't overflow and this is the last byte and the number being decoded
+            // is negative, check that the remaining bits are 1
+            if ((byte & 0x80 == 0) and (@bitCast(S, temp) < 0)) {
+                const remaining_shift = @intCast(u3, @typeInfo(U).Int.bits - @as(u16, shift));
+                const remaining_bits = @bitCast(i8, byte | 0x80) >> remaining_shift;
+                if (remaining_bits != -1) return error.Overflow;
+            }
         }
 
         value |= temp;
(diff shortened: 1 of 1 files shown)


lib/std/leb128.zig
```diff
@@ -233,8 +241,8 @@ test "deserialize signed LEB128" {
     try testing.expect((try test_read_ileb128(i8, "\xff\x7f")) == -1);
     try testing.expect((try test_read_ileb128(i16, "\xff\xff\x7f")) == -1);
     try testing.expect((try test_read_ileb128(i32, "\xff\xff\xff\xff\x7f")) == -1);
-    try testing.expect((try test_read_ileb128(i32, "\x80\x80\x80\x80\x08")) == -0x80000000);
-    try testing.expect((try test_read_ileb128(i64, "\x80\x80\x80\x80\x80\x80\x80\x80\x80\x01")) == @bitCast(i64, @intCast(u64, 0x8000000000000000)));
```
ifreund, line 237: perhaps we should add these 2 cases to the `expectError()` cases above to ensure this edge case is fully tested?

ifreund (commented):
Nice find and thanks for the fix!

You've done an excellent job explaining the issue and fix in the PR description, would you mind copying that text into the commit message as well so it's easier to find with `git blame`?

lib/std/leb128.zig
malcolmstill: Good shout!

andrewrk:
Thanks!

PR #9806, opened 2021-09-20
FIX resize() for non u8 element types.

colemanbroad:
Looks like a simple typo that breaks `resize()` on slices of non u8 types?

Fixes, e.g. 

```zig
  var pts = try gpa.allocator.alloc(f32,100);
  pts = try gpa.allocator.resize(pts, pts.len+1);
```

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -147,6 +147,30 @@ test "mem.Allocator basics" {
     try testing.expectError(error.OutOfMemory, failAllocator.allocSentinel(u8, 1, 0));
 }
 
+test "Allocator.resize" {
+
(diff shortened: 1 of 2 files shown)


Vexu (changes requested):
Nice catch but needs a test so it doesn't regress again, `mem.zig` would probably be a good place to add one to.

colemanbroad:
OK, cool. Thanks for the review. I'll use `testing.allocator` I guess?

colemanbroad:
So the ziglang.zig check fails on Linux only. The log appears to show a problem with `zig fmt --check`. I tried with running code through `zig fmt`, but that doesn't seem to solve it. Now I'm stuck ;)

colemanbroad:
OK, That's probably the issue. I'm using `0.8.1`. Is there an easy way to format with latest? Or should I bite the bullet and build from source?

PR #9808, opened 2021-09-21
tokenizer: Fix index-out-of-bounds on string_literal_backslash right before EOF

squeek502:
Before this fix, the added test would fail with something like:

```
thread 3287298 panic: index out of bounds
/home/ryan/Programming/zig/zig/build/lib/zig/std/zig/tokenizer.zig:408:34: 0x21511f in std.zig.tokenizer.Tokenizer.next (fuzz-debug)
            const c = self.buffer[self.index];
                                 ^
/home/ryan/Programming/zig/zig/build/lib/zig/std/zig/parse.zig:24:37: 0x20af60 in std.zig.parse.parse (fuzz-debug)
                                    ^
```

This was found by fuzzing the tokenizer with https://github.com/squeek502/zig-std-lib-fuzzing

(this is just the beginning, there are lots more tokenizer crashes that have been found)

### DIFF
--- a/lib/std/zig/tokenizer.zig
+++ b/lib/std/zig/tokenizer.zig
@@ -700,7 +700,7 @@ pub const Tokenizer = struct {
                 },
 
                 .string_literal_backslash => switch (c) {
-                    '\n' => {
+                    0, '\n' => {
                         result.tag = .invalid;
                         break;
                     },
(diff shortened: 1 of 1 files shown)


andrewrk:
Thanks!

PR #9809, opened 2021-09-21
tokenizer: Fix index-out-of-bounds on unfinished unicode escapes before EOF

squeek502:
Follow up to https://github.com/ziglang/zig/pull/9808, pretty much the same bug but with char literals (also found [via fuzzing](https://github.com/squeek502/zig-std-lib-fuzzing)).

Turns out these were the only other crashes found by the fuzzer, they were just being repeated a ton so I thought there'd be a lot more fixes necessary. The tokenizer fuzzer is now running cleanly so far with these fixes.


andrewrk:
Thanks!

Issue #9820, opened 2021-09-22
`zig init-exe` / `zig init-lib` should not overwrite existing `src/main.zig`


hdorio:
As of today `zig init-exe` will stop with the following error when an existing build file exists
> existing build.zig file would be overwritten

It would be good to do the same thing for all other generated files (`src/main.zig` currently) or at least to skip them instead of overwritting them.

I was trying to convert a library project to an executable project, the `build.zig` file was renamed and so `zig init-exe` erased the content of my `main.zig` file without warning.

I patched my `zig` copy with the following:

```
diff --git a/src/main.zig b/src/main.zig
index 8b5fbba90..f6e9e646e 100644
--- a/src/main.zig
+++ b/src/main.zig
@@ -2751,6 +2751,12 @@ pub fn cmdInit(
         error.FileNotFound => {},
         else => fatal("unable to test existence of build.zig: {s}\n", .{@errorName(err)}),
     }
+    if (fs.cwd().access("src" ++ s ++ "main.zig", .{})) |_| {
+        fatal("existing src" ++ s ++ "main.zig file would be overwritten", .{});
+    } else |err| switch (err) {
+        error.FileNotFound => {},
+        else => fatal("unable to test existence of src" ++ s ++ "main.zig: {s}\n", .{@errorName(err)}),
+    }
     var src_dir = try fs.cwd().makeOpenPath("src", .{});
     defer src_dir.close();
```

ifreund:
This seems like it would be a nice improvement, would you mind submitting that patch as a PR?

hdorio:
@ifreund I'll let you discuss about what is the best thing to do with @g-w1 https://github.com/ziglang/zig/pull/9821#discussion_r714346459

ifreund:
Implemented in  #9821

PR #9821, opened 2021-09-22
stage2: enhance `zig init-lib` and `zig init-exe`

hdorio:
Stop `src/main.zig` from being overwritten.

See #9820

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -2800,6 +2800,12 @@ pub fn cmdInit(
         error.FileNotFound => {},
         else => fatal("unable to test existence of build.zig: {s}\n", .{@errorName(err)}),
     }
+    if (fs.cwd().access("src" ++ s ++ "main.zig", .{})) |_| {
+        fatal("existing src" ++ s ++ "main.zig file would be overwritten", .{});
+    } else |err| switch (err) {
+        error.FileNotFound => {},
+        else => fatal("unable to test existence of src" ++ s ++ "main.zig: {s}\n", .{@errorName(err)}),
+    }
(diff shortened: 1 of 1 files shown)


src/main.zig
```diff
@@ -2800,6 +2800,12 @@ pub fn cmdInit(
         error.FileNotFound => {},
         else => fatal("unable to test existence of build.zig: {s}\n", .{@errorName(err)}),
     }
+    if (fs.cwd().access("src" ++ s ++ "main.zig", .{})) |_| {
+        fatal("existing src" ++ s ++ "main.zig file would be overwritten", .{});
+    } else |err| switch (err) {
+        error.FileNotFound => {},
+        else => fatal("unable to test existence of src" ++ s ++ "main.zig: {s}\n", .{@errorName(err)}),
```
g-w1, line 2807: IMO, the better thing to do here would just be to overwrite. This should not be a hard error.

ifreund: Why not? If the file system is returning an error on this operation then it is kind of unlikely that writing will succeed anyways. Better to play it safe and fail early than risk overwriting someone's existing main.zig.

ifreund (approved):

zigazeljko:
Why was this implemented with an additional `access` call? Isn't it simpler (and more idiomatic) to just use `createFile` with `.exclusive = true` option and handle the error there?

andrewrk:
@zigazeljko that is generally true but in this case the idea is to check if *any* files would be overwritten, before attempting to write *any* files. File system APIs do not provide any way to avoid races for this use case.

andrewrk:
Thanks @hdorio

PR #9822, opened 2021-09-22
Make opaque_decl an extended instruction, adjust union/enum tag astgen.

SpexGuy:
These changes are an initial step towards comptime closures.


SpexGuy:
Merged into #9823

PR #9823, opened 2021-09-23
Stage2: Implement comptime closures and the This builtin

SpexGuy:


src/Module.zig
```diff
@@ -275,6 +275,56 @@ pub const DeclPlusEmitH = struct {
...
+            .gpa = gpa,
+            .perm_arena = perm_arena,
+        };
+    }
+
+    pub fn finalize(noalias self: *@This()) !void {
+        assert(!self.finalized);
+        // use a temp to avoid unintentional aliasing due to RLS
+        const tmp = try self.scope.captures.clone(self.perm_arena);
+        self.scope.captures = tmp;
```
andrewrk, line 309: sound like we need to have a big language design discussion

src/Module.zig
```diff
@@ -3115,18 +3187,31 @@ pub fn semaFile(mod: *Module, file: *Scope.File) SemaError!void {
...
 
         if (sema.analyzeStructDecl(new_decl, main_struct_inst, struct_obj)) |_| {
+            try wip_captures.finalize();
             new_decl.analysis = .complete;
+
+            // If an error happens resolving the fields of a struct, it will be marked
+            // invalid and a proper compile error set up. But we should still look at the
+            // other types pending resolution.
+            //const src: LazySrcLoc = .{ .node_offset = 0 };
+            //sema.resolveDeclFields(block, src, ty) catch continue;
```
andrewrk, line 3213: is this code supposed to be uncommented?

This looks like it was cut+pasted from `resolvePendingTypes` and I'm a bit confused about what it's doing here.

src/Sema.zig
```diff
@@ -8,8 +8,12 @@
 mod: *Module,
 /// Alias to `mod.gpa`.
 gpa: *Allocator,
-/// Points to the arena allocator of the Decl.
+/// Points to the temporary arena allocator of the Sema.
+/// This arena will be cleared when the sema is destroyed.
 arena: *Allocator,
+/// Points to the arena allocator for the Decl.  This arena
```
andrewrk, line 14: Which one? `owner_decl`?

src/Sema.zig
```diff
@@ -129,15 +135,29 @@ pub fn analyzeBody(
...
 
+    const parent_capture_scope = block.wip_capture_scope;
+
+    var wip_captures = WipCaptureScope{
+        .finalized = true,
+        .scope = parent_capture_scope,
+        .perm_arena = sema.perm_arena,
+        .gpa = sema.gpa,
+    };
+    defer if (wip_captures.scope != parent_capture_scope) {
```
andrewrk, line 146: RIP my poor, sweet main loop :sob: 

ok for now; I'll come back to this later and make sure the machine code here ends up looking ok. there are some tricks we can do.

src/Sema.zig
```diff
@@ -3116,7 +3166,9 @@ fn analyzeCall(
             try namespace.anon_decls.ensureUnusedCapacity(gpa, 1);
 
             // Create a Decl for the new function.
-            const new_decl = try mod.allocateNewDecl(namespace, module_fn.owner_decl.src_node);
+            const src_decl = namespace.getDecl();
+            // TODO CLOSURE verify that depth+1 is correct here
```
andrewrk, line 3170: is this comment still meaningful?

src/Sema.zig
```diff
@@ -6536,8 +6618,46 @@ fn zirThis(
...
+
+fn zirClosureCapture(
+    sema: *Sema,
+    block: *Scope.Block,
+    inst: Zir.Inst.Index,
+) CompileError!void {
+    // TODO: Compile error when closed over values are modified
+    const inst_data = sema.code.instructions.items(.data)[inst].un_tok;
+    const tv = try sema.resolveInstConst(block, inst_data.src(), inst_data.operand);
+    //const arena = sema.perm_arena;
```
andrewrk, line 6634:
```suggestion
```

src/Zir.zig
```diff
@@ -62,8 +60,8 @@ pub const ExtraIndex = enum(u32) {
 };
 
 pub fn getMainStruct(zir: Zir) Inst.Index {
-    return zir.extra[@enumToInt(ExtraIndex.main_struct)] -
-        @intCast(u32, Inst.Ref.typed_value_map.len);
+    _ = zir;
+    return 0;
 }
```
andrewrk, line 65: hmm I suggest this:

`pub const main_struct_index = 0;`

this is simpler than a function call with an unnecessary argument but if things change to work a different way in the future we will still get a compile error at the usage sites.

test/behavior/this_stage1.zig
```diff
@@ -0,0 +1,34 @@
+const expect = @import("std").testing.expect;
```
andrewrk, line 1: All the tests in this file are passing in stage2; you can delete this file. The fact that you simplified the test case to not exercise struct initialization is fine; that's not what that test was testing. There's plenty of test coverage of struct initialization syntax in the struct.zig test cases.

andrewrk (changes requested):
Awesome work.

src/Module.zig
SpexGuy: Oops, good catch, forgot to delete this.

src/Module.zig
```diff
@@ -275,6 +275,56 @@ pub const DeclPlusEmitH = struct {
...
 };
 
+pub const CaptureScope = struct {
+    parent: ?*CaptureScope,
+
+    /// Values from this decl's evaluation that will be closed over in
+    /// child decls. Values stored in the value_arena of the linked decl.
+    /// During sema, this map is backed by the gpa.  Once sema completes,
+    /// it is reallocated using the value_arena.
+    captures: std.AutoArrayHashMapUnmanaged(Zir.Inst.Ref, TypedValue) = .{},
```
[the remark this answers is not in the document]
SpexGuy: I tried that originally, but it doesn't quite work.  `zirClosureGet` uses the `un_node` field, which uses Ref as its operand.  I could reinterpret it as an index, but that would be really nasty.  The indexToRef in zirClosureCapture could be moved to be a refToIndex in zirClosureGet, but it doesn't really improve the situation much.  A second option would be to introduce a `un_node_index` field, but that also feels like overkill to me.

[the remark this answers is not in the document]
andrewrk: I don't think it's overkill - plenty of ZIR instructions have their own form. `closure_get` doesn't want a un_node, it wants a node and a zir index, right? Let's give it the type-correct thing that it wants.

src/Sema.zig
SpexGuy: Yep, clarified.

src/Sema.zig
SpexGuy: Oops, nope.  Deleted.

test/behavior/this_stage1.zig
SpexGuy: Makes sense, thanks!

src/Module.zig
SpexGuy: I'm always down for that! This is #3696 and friends.

PR #9825, opened 2021-09-23
big ints: 2s complement signed and/or

Snektron:
Similar to xor, but these require a lot more written out cases. Comments are again in the source. I also made sure again that the required memory allocations remain the same for positive cases, but note that if an argument of `bitXor`, `bitAnd` and `bitOr` are negative extra caution is required for the amount of memory that the operation needs. Especially `bitAnd` with two negative numbers is a bit unexpected.

I also noticed a small edge case where `llsignedxor` would write the overflow value even if the result is positive.


andrewrk (approved):
Super helpful 👍👍👍

PR #9827, opened 2021-09-23
Merge call zir instructions to make space for field_call

SpexGuy:


src/AstGen.zig
```diff
@@ -1934,11 +1935,12 @@ fn unusedResultExpr(gz: *GenZir, scope: *Scope, statement: Ast.Node.Index) Inner
...
         const zir_tags = gz.astgen.instructions.items(.tag);
         switch (zir_tags[inst]) {
-            // For some instructions, swap in a slightly different ZIR tag
+            // For some instructions, modify the zir data
             // so we can avoid a separate ensure_result_used instruction.
-            .call_chkused => unreachable,
             .call => {
-                zir_tags[inst] = .call_chkused;
+                const extra_index = gz.astgen.instructions.items(.data)[inst].pl_node.payload_index;
+                const extra = @ptrCast(*Zir.Inst.Call, &gz.astgen.extra.items[extra_index]);
```
SpexGuy, line 1942: @andrewrk I know you aren't a huge fan of using ptrCast on zir, can you recommend a better way to do this?

andrewrk: I'm OK with this but `Zir.Inst.Call` needs to become an `extern struct` to make it sound.

SpexGuy: I ended up just putting the flags first, so they can easily be found without needing to reinterpret anything.

src/AstGen.zig
```diff
@@ -7897,20 +7908,24 @@ fn callExpr(
     call: Ast.full.Call,
 ) InnerError!Zir.Inst.Ref {
     const astgen = gz.astgen;
-    const lhs = try expr(gz, scope, .none, call.ast.fn_expr);
+    const tree = astgen.tree;
+
+    // If the lhs is a field access, we need to generate a special field_call_bind
+    // instruction first.  If this is a inst.func() call, this instruction will capture
+    // the value of the first parameter before evaluating the other parameters.
+    // TODO NO_COMMIT make sure .ref is correct here
```
andrewrk, line 7916: `ResultLoc.ref` means that you're going to get back a pointer to `call.ast.fn_expr` no matter what. For example if the fn expr was an identifier to a local, you'll get the `alloc` instruction, rather than a `load` instruction of that alloc. If the fn expr is an expression, such as `(a + b)` then the result of that expression will be stored in the stack and you'll get a pointer to the stack. The `ref` instruction does all that - stuffs an arbitrary value into an alloc then returns the alloc.

src/Sema.zig
```diff
@@ -9350,6 +9337,150 @@ fn fieldPtr(
...
+    field_name_src: LazySrcLoc,
+) CompileError!Air.Inst.Ref {
+    // When editing this function, note that there is corresponding logic to be edited
+    // in `fieldVal`. This function takes a pointer and returns a pointer.
+
+    const mod = sema.mod;
+    const raw_ptr_src = src; // TODO better source location
+    const raw_ptr_ty = sema.typeOf(raw_ptr);
+    const inner_ty = if (raw_ptr_ty.zigTypeTag() == .Pointer and raw_ptr_ty.ptrSize() == .One)
+        raw_ptr_ty.elemType()
```
andrewrk, line 9355: `elemType` is deprecated as of 1-2 commits ago; check if you want to call `elemType2` or `childType`.

src/Sema.zig
```diff
@@ -9350,6 +9337,150 @@ fn fieldPtr(
...
+    else
+        return mod.fail(&block.base, raw_ptr_src, "expected single pointer, found '{}'", .{raw_ptr_ty});
+
+    // Optionally dereference a second pointer to get the concrete type.
+    const is_double_ptr = inner_ty.zigTypeTag() == .Pointer and inner_ty.ptrSize() == .One;
+    const concrete_ty = if (is_double_ptr) inner_ty.elemType() else inner_ty;
+    const ptr_ty = if (is_double_ptr) inner_ty else raw_ptr_ty;
+    const object_ptr = if (is_double_ptr)
```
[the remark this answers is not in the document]
andrewrk: I don't see a problem here but this concern should be easy to write a test case for, no?

src/Sema.zig
```diff
@@ -9350,6 +9337,150 @@ fn fieldPtr(
...
+fn fieldCallBind(
+    sema: *Sema,
+    block: *Scope.Block,
+    src: LazySrcLoc,
+    raw_ptr: Air.Inst.Ref,
+    field_name: []const u8,
+    field_name_src: LazySrcLoc,
+) CompileError!Air.Inst.Ref {
+    // When editing this function, note that there is corresponding logic to be edited
+    // in `fieldVal`. This function takes a pointer and returns a pointer.
```
andrewrk, line 9349: Oof is this the 3rd copy of field lookup logic? I'm OK with this for now, but we'll probably want to think about how to DRY this up a bit in the future.

src/Sema.zig
```diff
@@ -9968,6 +10099,13 @@ fn coerceInMemoryAllowed(dest_type: Type, src_type: Type, dest_is_mut: bool) InM
             return .no_match;
         }
 
+        // TODO NO_COMMIT understand this assert.
+        if (dest_info.@"align" == 0 and
+            src_info.@"align" == 0)
+        {
+            return .ok;
+        }
```
andrewrk, line 10107: The asserts were there because the logic below which does a comparison was written with the (incorrect) assumption that there would never be a 0 align value. In other words I was focused on a different change, and so was sloppy here. The logic needs to be improved to handle this case.

A value of 0 means to use the ABI alignment of the pointee type. If either of the values are zero, it means we have to ask for the abi alignment of the pointee types of both of them, an action which may cause a "dependency loop detected" compile error. For now I suggest this code:

```zig
        // If both pointers have alignment 0, it means they both want ABI alignment.
        // In this case, if they share the same child type, no need to resolve
        // pointee type alignment. Otherwise both pointee types must have their alignment
        // resolved and we compare the alignment numerically.
        if (src_info.@"align" != 0 or dest_info.@"align" != 0 or
            !dest_info.pointee_type.eql(src_info.pointee_type))
        {
            const src_align = src_type.ptrAlignment(target);
            const dest_align = dest_type.ptrAlignment(target);

            if (dest_align > src_align) {
                return .no_match;
            }
        }
```

At some point we will need to go audit all the `ptrAlignment` callsites. Obtaining `target` in this function left as an exercise for the reader :stuck_out_tongue:

src/type.zig
```diff
@@ -5,6 +5,7 @@ const Allocator = std.mem.Allocator;
 const Target = std.Target;
 const Module = @import("Module.zig");
 const log = std.log.scoped(.Type);
+const Air = @import("Air.zig");
```
andrewrk, line 8: sus

andrewrk (commented):
Nice progress!

src/Sema.zig
SpexGuy: I agree, I couldn't see an easy way to compress it since I'm new to this code but it definitely should be done.

src/type.zig
SpexGuy: Oh good catch, I added this thinking mistakenly that I needed to use instToRef, and then forgot to delete it.

PR #9839, opened 2021-09-25
std/json: use bit-stack for nesting instead of large LLVM integer type

### LINKED ISSUE #5959: json.StreamingParser u256 causes issues... sometimes
Attempts at using expectEqual with any struct that contains the StreamingParser, LLVM crashes with "LLVM ERROR: Unsupported library call operation!"

```zig
// https://godbolt.org/z/ET8W9q
export fn foo() void {
    var parser: std.json.StreamingParser = undefined;
    std.testing.expectEqual(parser, parser);
}
```

The underlying problem is that `u256` is not fully supported by LLVM, but I am surprised that the currently tested path does not have any issues.

### Fixes

1. Downgrade to a 128 element stack — not ideal
2. Switch to PackedIntArray — might be (slightly) less performant due to increased bit ops
3. Fix LLVM? (https://github.com/ziglang/zig/issues/1485)
4. **Wait for stage2**

tiehuis:
The stack has been adjusted so that instead of pushing to index 0 in the
integer we push to the current end/index of the underlying integer. This
means we don't require a shift for every limb after each push/pop and
instead only require a mask/or and add/sub on a single element of the array.

Fixes #5959.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -132,6 +132,69 @@ pub const Token = union(enum) {
     Null,
 };
 
+const AggregateContainerType = enum(u1) { object, array };
+
+// A LIFO bit-stack. Tracks which container-types have been entered during parse.
+fn AggregateContainerStack(comptime n: usize) type {
+    return struct {
+        const Self = @This();
+        const TypeInfo = std.builtin.TypeInfo;
+
+        const element_bitcount = 8 * @sizeOf(usize);
+        const element_count = n / element_bitcount;
+        const ElementType = @Type(TypeInfo{ .Int = TypeInfo.Int{ .signedness = .unsigned, .bits = element_bitcount } });
+        const ElementShiftAmountType = std.math.Log2Int(ElementType);
+
+        comptime {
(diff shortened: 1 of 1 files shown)


lib/std/json.zig
```diff
@@ -132,6 +132,63 @@ pub const Token = union(enum) {
...
+
+        memory: [element_count]ElementType,
+        len: usize,
+
+        pub fn init(self: *Self) void {
+            self.memory = [_]ElementType{0} ** element_count;
+            self.len = 0;
+        }
+
+        pub fn push(self: *Self, bit: u1) !void {
```
tiehuis, line 157: Reviewing myself, it might be a bit cleaner to return an optional for push/peek/pop and just use orelse when using the stack. We don't propagate any of these errors via try.

tiehuis:
Will review the test failures. Think I have missed some of the standalone tests due to some filters I was using during test.

tiehuis:
This is ready, assuming there are no more review comments.

PR #9843, opened 2021-09-26
stage2: array stuff

g-w1:
No tests from array_stage1.zig because apparently 
```
const a = [_]u64{a, b};
```
does not actually use array_init.

image
Works!!! (this uses array_init)
Could I use this as a test for now?

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -8223,8 +8223,49 @@ fn zirArrayInit(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index, is_ref: 
     const inst_data = sema.code.instructions.items(.data)[inst].pl_node;
     const src = inst_data.src();
 
-    _ = is_ref;
-    return sema.mod.fail(&block.base, src, "TODO: Sema.zirArrayInit", .{});
+    const extra = sema.code.extraData(Zir.Inst.MultiOp, inst_data.payload_index);
+    const args = sema.code.refSlice(extra.end, extra.data.operands_len);
+
+    var resolved_args = try sema.mod.gpa.alloc(Air.Inst.Ref, args.len);
+    for (args) |arg, i| resolved_args[i] = sema.resolveInst(arg);
+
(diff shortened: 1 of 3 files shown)


src/Sema.zig
```diff
@@ -7554,8 +7555,52 @@ fn zirArrayInit(sema: *Sema, block: *Scope.Block, inst: Zir.Inst.Index, is_ref:
...
+    for (args) |arg, i| resolved_args[i] = sema.resolveInst(arg);
+
+    var all_args_comptime = for (resolved_args) |arg| {
+        if ((try sema.resolveMaybeUndefVal(block, src, arg)) == null) break false;
+    } else true;
+
+    if (all_args_comptime) {
+        var anon_decl = try block.startAnonDecl();
+        defer anon_decl.deinit();
+        // TODO what to do if resolved_args.len == 0, where is the type info communicated?
```
g-w1, line 7571: What should I do for this, and also below.

andrewrk: AstGen will emit `struct_init_empty` instead of `array_init` if there are no fields/elements. You can simply assert ags.len != 0.

src/print_zir.zig
```diff
@@ -1811,6 +1812,21 @@ const Writer = struct {
...
+        const extra = self.code.extraData(Zir.Inst.MultiOp, inst_data.payload_index);
+        const args = self.code.refSlice(extra.end, extra.data.operands_len);
+
+        try stream.writeAll(".{");
+        for (args) |arg, i| {
+            if (i != 0) try stream.writeAll(", ");
+            try self.writeInstRef(stream, arg);
+        }
+        try stream.writeAll("}");
+        try stream.writeAll(")");
```
andrewrk, line 1827:
```suggestion
        try stream.writeAll("})");
```

andrewrk (changes requested):
Feel free to add additional behavior test cases for arrays that exercise the new code paths.

The logic looks fine, but that TODO comment should be deleted or replaced with an assert before merge. 

Thanks!

PR #9846, opened 2021-09-26
system.zig: handle environment variables

voroskoi:
Hi,

This PR adds support for various environment variables, so zig will use them when they are set.
This is quite useful when You have include or lib files on non-standard places, like with guix.

I am not sure if this should be done this way or this idea is completely broken.

Thanks,

### DIFF
--- a/lib/std/zig/system.zig
+++ b/lib/std/zig/system.zig
@@ -58,6 +58,34 @@ pub const NativePaths = struct {
             error.EnvironmentVariableNotFound => {},
             error.OutOfMemory => |e| return e,
         }
+
+        if (process.getEnvVarOwned(allocator, "C_INCLUDE_PATH")) |c_include_path| {
+            defer allocator.free(c_include_path);
+
+            var it = mem.tokenize(u8, c_include_path, ":");
+            while (true) {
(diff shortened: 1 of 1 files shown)


matu3ba:
> quite useful when You have include or lib files on non-standard places, like with guix

Can you elaborate, why [pkg-config](http://guix.gnu.org/packages/pkg-config-0.29.2/) does not work? Explanation for usage is given [here](https://zig.news/xq/zig-build-explained-part-2-1850).
I would prefer having as few hacks as possible.

Another alternative for custom installation paths is to write the config into a struct/file and read it from build.zig.

mikdusan:
Please provide some examples where `C_INCLUDE_PATH` is not already being picked up?
With zig master the following things are already effected:
- `zig build-obj thing.c ...`
- `zig cc -c thing.c ...`
- things brought in via `@cImport()`

voroskoi:
Hi,

pkg-config only works when the `.pc` file is available.

As far as I know clang and gcc also support these environment variables, so using zig cc/c++ in those cases as a drop-in replacement will not work.

But I understand if You do not want to support it. Thanks,

Vexu:
The Clang driver already handles this.

PR #9850, opened 2021-09-28
Adds Linux support for POSIX file locking with fcntl

acarrico:
On Linux, locking fails with EAGAIN (vs. EACCES on other systems).
This commit also adds FcntlErrors for EDEADLK and ENOLCK.

### DIFF
--- a/lib/std/c/darwin.zig
+++ b/lib/std/c/darwin.zig
@@ -357,11 +357,11 @@ pub const off_t = i64;
 pub const ino_t = u64;
 
 pub const Flock = extern struct {
-    l_start: off_t,
-    l_len: off_t,
-    l_pid: pid_t,
-    l_type: i16,
-    l_whence: i16,
+    start: off_t,
+    len: off_t,
+    pid: pid_t,
+    type: i16,
+    whence: i16,
 };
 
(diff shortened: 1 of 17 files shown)


Vexu:
Sorry for taking so long to get to this.
<br>
> * Splitting up the `fcntl` commands into separate procedures would yield precise error sets.

Let's keep that in mind when #6600 is implemented.

>* [Here is my test](https://gist.github.com/acarrico/8a0fdbf04db62f9fff9ce094c53bec89). I'm not sure where tests are normally added for the Zig standard library.

I cleaned it up a bit and added it to `std/os/test.zig`.

> * One issue with this test is it calls `std.time.sleep` while setting up a deadlock test.

I adjusted it to only sleep 1ms, seems to still work fine.

>* Another issue with this test is it seems to duplicate the notice:

`std.Progress` has been updated many times since this pr was opened and this no longer seems to happen.

acarrico:
Thank you

andrewrk:
The test added in this patch has been failing sporadically on the CI. Issue opened here: #11074

Test has been disabled in 76d810c568648e32af0d74af2037b73b7ea38270.

Issue #9807, opened 2021-09-20
Modular arithmetic fails to compile on msp430 target.


nmichaels:
Given the following program:

```
const std = @import("std");

export fn modNum(a: u8) u8 {
    return a % 14;
}

fn reset() noreturn {
    asm volatile ("mov 0, r0");
    unreachable;
}

export fn main() noreturn {
    var n: u8 = 0;
    while (n < 152) : (n += 1) {
        if (modNum(n) >= 14) {
            reset();
        }
    }
    while (true) {}
}

pub export fn _start() linksection(".init9") noreturn {
    asm volatile (
        \\ jmp main
    );
    unreachable;
}
```
and appropriate linker scripts from TI's Energia release (or maybe MSP430-GCC, see link below), the following command ought to build something:

`zig build-exe -Lldscripts -Lldscripts/msp430g2553 --script msp430.x --single-threaded -O ReleaseSmall -target msp430-freestanding-none -mcpu msp430 modulo.zig`

And if you change the % in `modNum` to a +, it does. (Though it won't run because there's no interrupt table and _start does none of the setup stuff it's supposed to do.) However, as is, it spits out this error:

```ld.lld: error: undefined symbol: __mspabi_remu
>>> referenced by modulo.zig:4 (/home/nathan/zig/mod/modulo.zig:4)
>>>               zig-cache/o/4ed470d32b8b5d645abaefc9c97f90f7/modulo.o:(modNum)
thread 1570666 panic: attempt to unwrap error: LLDReportedFailure
/home/nathan/zig/zig/lib/std/heap.zig:147:9: 0x55949f7bf986 in std.heap.CAllocator.resize (zig1)
        return error.OutOfMemory;
        ^
/home/nathan/zig/zig/lib/std/mem/Allocator.zig:319:16: 0x55949f9f3d79 in std.mem.Allocator.resize (zig1)
    const rc = try self.resizeFn(self, old_byte_slice, Slice.alignment, new_byte_count, 0, @returnAddress());
               ^
/home/nathan/zig/zig/lib/std/os.zig:1761:19: 0x55949fabf69c in std.os.unlinkatZ (zig1)
        ENOENT => return error.FileNotFound,
                  ^
/home/nathan/zig/zig/lib/std/fs.zig:1359:25: 0x55949f997a4f in std.fs.Dir.deleteFileZ (zig1)
            else => |e| return e,
                        ^
[...]

PR #9862, opened 2021-09-29
Beginnings of msp430 ABI support in compiler_rt.

nmichaels:
Addresses bug #9807, plenty of room for further implementation and
improvements.

### DIFF
--- a/lib/std/special/compiler_rt.zig
+++ b/lib/std/special/compiler_rt.zig
@@ -540,6 +540,85 @@ comptime {
             @export(__unordtf2, .{ .name = "__unordkf2", .linkage = linkage });
         }
 
+        if ((arch == .msp430) and !is_test) {
+            const msp430 = @import("compiler_rt/msp430.zig");
+
+            @export(
+                msp430.__mspabi_fixdi,
+                .{ .name = "__mspabi_fixdi", .linkage = linkage },
+            );
+            @export(
+                msp430.__mspabi_fixdli,
+                .{ .name = "__mspabi_fixdli", .linkage = linkage },
+            );
+            @export(
+                msp430.__mspabi_fixdlli,
+                .{ .name = "__mspabi_fixdlli", .linkage = linkage },
+            );
+            @export(
+                msp430.__mspabi_fixdu,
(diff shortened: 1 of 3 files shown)


LemonBoy:
Not so fast, all the builtin functions must be declared with a specific calling convention that Zig doesn't support yet. The plan is to have the compiler understand the `builtin` CC first and then to implement a few ones, having all of them stubbed out like this is a bad idea and a PITA for the unsuspecting user that happens to trigger one.

nmichaels:
> ... having all of them stubbed out like this is a bad idea and a PITA for the unsuspecting user that happens to trigger one.

I think it's better to get a compiler error the ones in this patch than the one I pasted in #9807 .

That said, I was really just hoping to get the discussion started and point this in the right direction.

It looked from the ABI like the majority of the functions just use the C calling convention (see comments in, e.g., `__mspabi_cmpd` for exceptions). Should those just use `.Naked` and hack around it with inline asm (like in arm.zig), or do you think we should wait until Zig supports the calling convention natively?

LemonBoy:
Fair enough, I didn't see the compileerror calls.

Adding a new CC takes [only a few LoC](https://github.com/ziglang/zig/commit/9ebdbca379fb71da7236e7534e1d3465f7fb151f), I'd start from that.

nmichaels:
That does look pretty easy. I'm still not totally clear on why this needs a new calling convention though. The functions I added are called the same way other C functions are called. There are bunches of other functions in compiler_rt.zig that use `callconv(.C)`. Are you saying we should change all of those before making Zig usable on msp430?

Or maybe I'm completely misunderstanding the point.

LemonBoy:
> Or maybe I'm completely misunderstanding the point.

You are. As stated in the LLM [frontend](https://www.llvm.org/doxygen/MSP430ISelLowering_8cpp_source.html#l00455) many of the msp-specific functions expect a given special CC, the arguments and return values are passed in some specific registers and the usable register set is also different from the usual `.C` CC.

andrewrk:
I do think it would be worth following @LemonBoy's guidance on this. I'm closing this because no recent activity, and it does not appear to be in a mergeable state. Please feel free to open a fresh one if you decide to pick it up.

PR #9873, opened 2021-09-30
Stage 2: A bunch of cleaning up

SpexGuy:
This fixes `std.target.Target.Cpu.Feature.feature_set_fns`, which may unlock other tests.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -54,7 +54,7 @@ c_object_work_queue: std.fifo.LinearFifo(*CObject, .Dynamic),
 /// These jobs are to tokenize, parse, and astgen files, which may be outdated
 /// since the last compilation, as well as scan for `@import` and queue up
 /// additional jobs corresponding to those new files.
-astgen_work_queue: std.fifo.LinearFifo(*Module.Scope.File, .Dynamic),
+astgen_work_queue: std.fifo.LinearFifo(*Module.File, .Dynamic),
 
 /// The ErrorMsg memory is owned by the `CObject`, using Compilation's general purpose allocator.
 /// This data is accessed by multiple threads and is protected by `mutex`.
@@ -446,7 +446,7 @@ pub const AllErrors = struct {
     pub fn addZir(
         arena: *Allocator,
         errors: *std.ArrayList(Message),
-        file: *Module.Scope.File,
+        file: *Module.File,
     ) !void {
         assert(file.zir_loaded);
         assert(file.tree_loaded);
@@ -1444,7 +1444,7 @@ pub fn create(gpa: *Allocator, options: InitOptions) !*Compilation {
             .emit_docs = options.emit_docs,
             .work_queue = std.fifo.LinearFifo(Job, .Dynamic).init(gpa),
             .c_object_work_queue = std.fifo.LinearFifo(*CObject, .Dynamic).init(gpa),
-            .astgen_work_queue = std.fifo.LinearFifo(*Module.Scope.File, .Dynamic).init(gpa),
+            .astgen_work_queue = std.fifo.LinearFifo(*Module.File, .Dynamic).init(gpa),
             .keep_source_files_loaded = options.keep_source_files_loaded,
             .use_clang = use_clang,
             .clang_argv = options.clang_argv,
@@ -2465,14 +2465,14 @@ pub fn performAllTheWork(self: *Compilation) error{ TimerUnsupported, OutOfMemor
 const AstGenSrc = union(enum) {
     root,
     import: struct {
-        importing_file: *Module.Scope.File,
+        importing_file: *Module.File,
         import_tok: std.zig.Ast.TokenIndex,
     },
 };
 
 fn workerAstGenFile(
     comp: *Compilation,
-    file: *Module.Scope.File,
+    file: *Module.File,
     prog_node: *std.Progress.Node,
     wg: *WaitGroup,
     src: AstGenSrc,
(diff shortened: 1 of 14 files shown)


src/Sema.zig
```diff
@@ -11754,18 +11749,14 @@ pub fn resolveDeclFields(sema: *Sema, block: *Scope.Block, src: LazySrcLoc, ty:
     switch (ty.tag()) {
         .@"struct" => {
             const struct_obj = ty.castTag(.@"struct").?.data;
-            if (struct_obj.owner_decl.namespace != sema.owner_decl.namespace) return;
+            if (struct_obj.owner_decl.namespace.parent != sema.owner_decl.namespace) return;
```
SpexGuy, line 11752: I don't understand what this is trying to check.  I had to change it to keep the tests passing, but I'm not sure if it's still correct.  @andrewrk can you please explain what this is for?

SpexGuy:
Ok this technique doesn't work.  The "correct" alternative would be a pretty big change, but I think it would make a lot of things cleaner.  That change would be to make the field initializer block into its own decl, which is a child of the "namespace" decl.  From there we could start to look at removing namespace decls entirely, which I think is a good step.

There's also a hacky short-term solution I could do if anyone is immediately blocking on this.

SpexGuy:
Alright, I ended up with a compromise.  It didn't totally make sense to make fields a decl, since they have no tv.  Instead, they now use the namespace decl, but they are analyzed in a separate sema context.  This allows us to analyze them at any time, so we don't need the array of types to process at end of scope.  I then added a namespace field to blocks, so that a block can have a narrower namespace than its containing decl.  This allows fields to be analyzed in the context of the namespace decl without making namespace decls a special case, and without creating a stub decl for the fields.

src/Compilation.zig
```diff
@@ -1889,13 +1889,13 @@ pub fn totalErrorCount(self: *Compilation) usize {
         // the previous parse success, including compile errors, but we cannot
         // emit them until the file succeeds parsing.
         for (module.failed_decls.keys()) |key| {
-            if (key.namespace.file_scope.okToReportErrors()) {
+            if (key.getFileScope().okToReportErrors()) {
```
SpexGuy, line 1892: Since I was updating all of the references to decl.namespace (see the comment on that declaration for why), I also figured I'd update the ones going straight to the file scope (which is most of them) to use `getFileScope()` instead.  This will allow us to more easily change how the file scope is accessed if e.g. we move it into the decl directly or we move it to only the root namespace or something.  It also makes the codebase more searchable in general.

src/Module.zig
```diff
@@ -350,7 +350,7 @@ pub const Decl = struct {
     /// Reference to externally owned memory.
     /// In the case of the Decl corresponding to a file, this is
     /// the namespace of the struct, since there is no parent.
-    namespace: *Scope.Namespace,
+    src_namespace: *Scope.Namespace,
```
SpexGuy, line 353: This name was changed because the semantics have changed -- most uses should now use `block.namespace` instead.  I want this to cause compile errors on other PRs if they are using the old semantics.

src/Module.zig
```diff
@@ -4226,7 +4266,7 @@ fn markOutdatedDecl(mod: *Module, decl: *Decl) !void {
     decl.analysis = .outdated;
 }
 
-pub fn allocateNewDecl(mod: *Module, namespace: *Scope.Namespace, src_node: Ast.Node.Index, src_scope: ?*CaptureScope) !*Decl {
+pub fn allocateNewDecl(mod: *Module, name: [:0]const u8, namespace: *Scope.Namespace, src_node: Ast.Node.Index, src_scope: ?*CaptureScope) !*Decl {
```
SpexGuy, line 4272: The name is available at all call sites, and having it here is nice for debugging, so I adjusted it to be passed in.

src/Module.zig
```diff
@@ -608,18 +608,23 @@ pub const Decl = struct {
...
         return tree.tokens.items(.start)[decl.srcToken()];
     }
 
     pub fn renderFullyQualifiedName(decl: Decl, writer: anytype) !void {
         const unqualified_name = mem.spanZ(decl.name);
-        return decl.namespace.renderFullyQualifiedName(unqualified_name, writer);
+        return decl.src_namespace.renderFullyQualifiedName(unqualified_name, writer);
+    }
+
+    pub fn renderFullyQualifiedDebugName(decl: Decl, writer: anytype) !void {
```
SpexGuy, line 625: I was getting confused by type names like `mem.Allocator.mem.Allocator`, so this alternate method renders them as `mem/Allocator.zig:mem.Allocator`. Which now that I think about it might indicate another namespace bug... but the format conveys more information regardless. This is what the debug report uses now.

src/Sema.zig
```diff
@@ -11896,16 +11832,16 @@ fn resolveBuiltinTypeFields(
     return sema.resolveTypeFields(block, src, resolved_ty);
 }
 
-fn analyzeStructFields(
-    sema: *Sema,
-    block: *Scope.Block,
+fn semaStructFields(
+    mod: *Module,
     struct_obj: *Module.Struct,
 ) CompileError!void {
```
SpexGuy, line 11838: I considered moving `semaStructFields` and `semaUnionFields` into `Module.zig`, but they call a lot of functions that are currently private to `sema` and there isn't a clean place to split this function because it retains a lot of locals.

SpexGuy:
With the removal of `Module.Scope`, this PR probably conflicts with literally all of the other ones.  @andrewrk once this is approved, let's coordinate on the merge.  You pull in any other PRs first, and then I'll rebase this and fix the conflicts so this can be pulled quickly as well.

src/Sema.zig
SpexGuy: This was trying to check that the sema was the correct one for resolving the struct fields.  Since we now use captures for values that cross this boundary, the whole struct resolution stack is no longer needed.  It's been deleted in this PR.

andrewrk (approved):

PR #9876, opened 2021-10-01
stage2: implement ** and ++ at runtime

g-w1:


src/Sema.zig
```diff
@@ -6526,12 +6528,70 @@ fn zirArrayCat(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Ai
...
+        try Type.Tag.array.create(sema.arena, .{ .len = final_len, .elem_type = lhs_info.elem_type });
+    const ptr_ty = try Type.ptr(sema.arena, .{
+        .pointee_type = array_ty,
+        .@"addrspace" = target_util.defaultAddressSpace(sema.mod.getTarget(), .local),
+    });
+    const alloc = try block.addTy(.alloc, ptr_ty);
+
+    // TODO there is some ownership issue where if one of these is comptime-known, it generates an invalid tv
+    const ptr_to_lhs = if (!is_pointer) try sema.analyzeRef(block, lhs_src, lhs) else lhs;
+    const ptr_to_rhs = if (!is_pointer) try sema.analyzeRef(block, rhs_src, rhs) else rhs;
```
g-w1, line 6548: Do I have to copy one of these into sema.arena if it is comptime known?
I get invalid int values inside the air array.
image

g-w1: I put in some code to handle it, and its still there ...

andrewrk (changes requested):
In your crash screenshot, the Value object that has a bad memory reference is coming from a Decl's Value. Check the Decl name in a debugger and figure out why that Decl's Value is invalid.

g-w1:
I stubbed the case that causes the uaf out as a TODO for now as it is confusing me.

g-w1:
Ok, marking as draft as memset with bitcast is being weird in llvm backend.

andrewrk:
30+ days old draft; closing.

PR #9877, opened 2021-10-01
add `NotLink` error

### LINKED ISSUE #9872: 'readLink' fails on non-symlink file
### Summary

When running the std.os.readLink function on a non-link path on Linux, an unreachable place of code is reached.

### Setup
1. WSL or Linux environment
2. Zig 0.8.1

### Reproduction
1. Choose a non-link file in linux.
```
var buffer: [2 << 20]u8 = undefined;
const result = try std.fs.cwd().readLink(link, buffer[0..]);
```
result:
```
/home/foobar/zig/lib/std/os.zig:2559:19: 0x20a017 in std.os.readlinkatZ (readlink)
        EINVAL => unreachable,    
```
### Analysis
The system call likely returns EINVAL if a non-link file is link resolved. The code at readlinkatZ doesn't expect this value as the switch case in unreachable.

### Possible solution
Add a 'NotALink' error and return it when receiving EINVAL from the system.

AliChraghi:
fixes #9872


lib/std/zig/system.zig
```diff
@@ -612,6 +612,7 @@ pub const NativeTargetInfo = struct {
             error.AccessDenied => return error.GnuLibCVersionUnavailable,
             error.FileSystem => return error.FileSystem,
             error.SymLinkLoop => return error.SymLinkLoop,
+            error.NotLink => return error.NotLink,
```
squeek502, line 615: I think the quickest fix would be to just do:

```suggestion
            error.NotLink => return error.GnuLibCVersionUnavailable,
```

AliChraghi: thx, but I can not build the zig and I have to see the result in CI which leads to a waste of resources

AliChraghi: @squeek502 thank you. build is successful!

squeek502:
Looks like you might have committed line ending changes to all files accidentally with your last commit.

AliChraghi:
yeah :smile: I added lib/ directory instead lib/std.

marler8997 (approved):
Looks good to me now

vrischmann:
The linked issue is affecting Fedora 35 users, essentially making the `zig` command unusable because it tries to readlink `/lib64/libc.so.6` which is a normal file.

I can confirm that with this PR zig works again.

andrewrk:
Thank you @AliChraghi and reviewers :+1:

PR #9880, opened 2021-10-02
deflate: Better Huffman.construct errors and error handling

squeek502:
This brings Huffman.construct error handling in line with puff.c (I'm assuming we're okay with treating `puff.c` as an oracle). The error names and logic are from puff.c (and zlib's inflate.c uses the same terminology).

Note: I am planning to add relevant test cases soon. EDIT: Done

---

This is the last piece of the puzzle for getting [the deflate.zig <-> puff.c comparison fuzzer](https://github.com/squeek502/zig-std-lib-fuzzing) (which checks that all inputs either give the same inflated output or give errors in both implementations) to run cleanly (so far at least). Other required puzzle pieces:

- https://github.com/ziglang/zig/pull/9860 (not yet merged)
- https://github.com/ziglang/zig/pull/9849

cc @LemonBoy @mattbork

### DIFF
--- a/lib/std/compress/deflate.zig
+++ b/lib/std/compress/deflate.zig
@@ -702,8 +702,22 @@ test "empty distance alphabet" {
 
 test "inflateStream fuzzing" {
     // see https://github.com/ziglang/zig/issues/9842
-    try std.testing.expectError(error.EndOfStream, testInflate("\x950000"));
+    try std.testing.expectError(error.EndOfStream, testInflate("\x95\x90=o\xc20\x10\x86\xf30"));
     try std.testing.expectError(error.OutOfCodes, testInflate("\x950\x00\x0000000"));
+
+    // Huffman.construct errors
+    // lencode
+    try std.testing.expectError(error.InvalidTree, testInflate("\x950000"));
+    try std.testing.expectError(error.InvalidTree, testInflate("\x05000"));
+    // hlen
+    try std.testing.expectError(error.InvalidTree, testInflate("\x05\xea\x01\t\x00\x00\x00\x01\x00\\\xbf.\t\x00"));
+    // hdist
+    try std.testing.expectError(error.InvalidTree, testInflate("\x05\xe0\x01A\x00\x00\x00\x00\x10\\\xbf."));
+
+    // Huffman.construct -> error.IncompleteSet returns that shouldn't give error.InvalidTree
+    // (like the "empty distance alphabet" test but for ndist instead of nlen)
+    try std.testing.expectError(error.EndOfStream, testInflate("\x05\xe0\x01\t\x00\x00\x00\x00\x10\\\xbf\xce"));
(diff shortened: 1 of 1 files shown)


lib/std/compress/deflate.zig
```diff
@@ -406,8 +417,24 @@ pub fn InflateStream(comptime ReaderType: type) type {
             if (lengths[256] == 0)
                 return error.MissingEOBCode;
 
-            try self.huffman_tables[0].construct(lengths[0..nlen]);
-            try self.huffman_tables[1].construct(lengths[nlen .. nlen + ndist]);
+            self.huffman_tables[0].construct(lengths[0..nlen]) catch |err| switch (err) {
+                error.Oversubscribed => return error.InvalidTree,
+                error.IncompleteSet => {
+                    // incomplete code ok only for single length 1 code
+                    if (nlen != self.huffman_tables[0].count[0] + self.huffman_tables[0].count[1]) {
```
squeek502, line 424: When generating test cases through fuzzing, I'm seemingly not able to find an input for which this conditional is false (and same thing for the similar conditional for distances below).

If these are not actually possible, then these `construct` calls could be simplified into simple `catch return error.InvalidTree`'s.

mattbork: That's odd, I made sure the "empty distance alphabet" test had complete code length, literal/length, and distance alphabets and tested it against puff.c, which accepts it without error. I'll have look at this more.

squeek502: Your test is totally correct, I think my wording is just confusing.

mattbork: Oh, ok, that's good. If you do still need test cases for any of the incomplete alphabet checks, they're easy to construct by hand. Let me know.

squeek502:
Test cases added. Note that https://github.com/ziglang/zig/pull/9860 should almost certainly be merged before this (I tested locally with those changes included, so if this PR fails in CI then it's likely because https://github.com/ziglang/zig/pull/9860 needs to be merged first).

lib/std/compress/deflate.zig
```diff
@@ -125,6 +127,9 @@ const Huffman = struct {
 
         self.last_code = codes[PREFIX_LUT_BITS + 1];
         self.last_index = offset[PREFIX_LUT_BITS + 1] - self.count[PREFIX_LUT_BITS + 1];
+
+        if (left > 0)
```
LemonBoy, line 131: Minor nit, why not move the check below the loop that computes `left`?

squeek502: I didn't want to potentially leave `self.last_code`/`self.last_index` uninitialized since `error.IncompleteSet` is not always treated as an error by the caller, and if `decode` uses them when uninitialized it can lead to index-out-of-bounds/UB.

PR #9673, opened 2021-09-02
Astgen: fix lrbrace Zir symbols

nektro:
Fixes #9672

https://github.com/ziglang/zig/commit/73f77f30804f3a512fbb35caad3d87ff2778ecc0 is correct in observing AstGen progresses through the source monotonically, however it also assumes that all calls to `AstGen.advanceSourceCursor` are as well. When determining the line/column position for functions, the search for `lbrace` causes it to find a source position far back from the end of the function where it is now at near `rbrace`. This PR checks this condition and adjusts the local variables accordingly so that it may  always return a correct result and the marching performance improvements can still be taken advantage of.

PR #9882, opened 2021-10-02
astgen.zig: keep source cursor increasing monotonically as much as possible

### LINKED ISSUE #9672: stage2: wrong lbrace_line and rbrace_line for "exit2" function in start.zig
command: `../zig-out/bin/zig ast-check -t ../lib/std/start.zig`
```
  [4293] exit2 line(93) hash(a1241306655cf30eeaca93c6b024b065): %540 = block_inline({
    %542 = param("code", {
      %541 = break_inline(%542, @Ref.usize_type)
    ) token_offset:94:10
    %617 = func(ret_ty={
      %543 = break_inline(%617, @Ref.noreturn_type)
    }, body={
      %544 = dbg_stmt(94, 4)
      %545 = decl_ref("builtin") token_offset:95:13
      %546 = field_val(%545, "stage2_os") node_offset:95:20
      %547 = typeof(%546) node_offset:95:20
      %552 = enum_literal("linux") token_offset:96:10
      %553 = as_node(%547, %552) node_offset:96:10
      %588 = enum_literal("plan9") token_offset:124:10
      %589 = as_node(%547, %588) node_offset:124:10
      %548 = switch_block_else(%546, else => {
        %549 = str("TODO")
        %550 = as_node(@Ref.const_slice_u8_type, %549) node_offset:146:31
        %551 = compile_error(%550) node_offset:146:17
      }, %553 => {
        %554 = decl_ref("builtin") token_offset:96:27
        %555 = field_val(%554, "stage2_arch") node_offset:96:34
        %556 = typeof(%555) node_offset:96:34
        %561 = enum_literal("x86_64") token_offset:97:14
        %562 = as_node(%556, %561) node_offset:97:14
        %570 = enum_literal("arm") token_offset:105:14
        %571 = as_node(%556, %570) node_offset:105:14
        %578 = enum_literal("aarch64") token_offset:113:14
        %579 = as_node(%556, %578) node_offset:113:14
        %557 = switch_block_else(%555, else => {
          %558 = str("TODO")
          %559 = as_node(@Ref.const_slice_u8_type, %558) node_offset:121:35
          %560 = compile_error(%559) node_offset:121:21
        }, %562 => {
          %563 = dbg_stmt(97, 16)
          %564 = int(231)
          %565 = as_node(@Ref.usize_type, %564) node_offset:100:41
          %566 = as_node(@Ref.usize_type, %542) node_offset:101:39
          %567 = extended(asm(volatile, "syscall", , input(number, "{rax}", %565), input(arg1, "{rdi}", %566)rcx, r11, memory)) node_offset:98:17
          %568 = ensure_result_used(%567) node_offset:98:17
          %569 = break(%557, @Ref.void_value)
        }, %571 => {
          %572 = dbg_stmt(105, 16)
          %573 = as_node(@Ref.usize_type, @Ref.one) node_offset:108:40
          %574 = as_node(@Ref.usize_type, %542) node_offset:109:38
          %575 = extended(asm(volatile, "svc #0", , input(number, "{r7}", %573), input(arg1, "{r0}", %574)memory)) node_offset:106:17
          %576 = ensure_result_used(%575) node_offset:106:17
          %577 = break(%557, @Ref.void_value)
        }, %579 => {
          %580 = dbg_stmt(113, 16)
          %581 = int(93)
          %582 = as_node(@Ref.usize_type, %581) node_offset:116:40
          %583 = as_node(@Ref.usize_type, %542) node_offset:117:38
          %584 = extended(asm(volatile, "svc #0", , input(number, "{x8}", %582), input(arg1, "{x0}", %583)memory, cc)) node_offset:114:17
          %585 = ensure_result_used(%584) node_offset:114:17
          %586 = break(%557, @Ref.void_value)
        }) node_offset:96:19
        %587 = break(%548, %557)
      }, %589 => {
        %590 = decl_ref("builtin") token_offset:124:27
        %591 = field_val(%590, "stage2_arch") node_offset:124:34
        %592 = typeof(%591) node_offset:124:34
        %597 = enum_literal("x86_64") token_offset:125:14
        %598 = as_node(%592, %597) node_offset:125:14
        %605 = enum_literal("aarch64") token_offset:137:14
        %606 = as_node(%592, %605) node_offset:137:14
        %593 = switch_block_else(%591, else => {
          %594 = str("TODO")
          %595 = as_node(@Ref.const_slice_u8_type, %594) node_offset:144:35
          %596 = compile_error(%595) node_offset:144:21
        }, %598 => {
          %599 = dbg_stmt(125, 16)
          %600 = int(8)
          %601 = as_node(@Ref.usize_type, %600) node_offset:131:49
          %602 = extended(asm(volatile, "push $0\npush $0\nsyscall", , input(syscall_number, "{rbp}", %601)rcx, r11, memory)) node_offset:126:17
          %603 = ensure_result_used(%602) node_offset:126:17
          %604 = break(%593, @Ref.void_value)
        }, %606 => {
          %607 = dbg_stmt(137, 16)
          %608 = int(8)
          %609 = as_node(@Ref.usize_type, %608) node_offset:140:38
          %610 = extended(asm(volatile, "svc #0", , input(exit, "{x0}", %609)memory, cc)) node_offset:138:17
          %611 = ensure_result_used(%610) node_offset:138:17
          %612 = break(%593, @Ref.void_value)
        }) node_offset:124:19
        %613 = break(%548, %593)
      }) node_offset:95:5
      %614 = ensure_result_used(%548) node_offset:95:5
      %615 = dbg_stmt(147, 4)
      %616 = unreachable(safe) node_offset:148:5
    }) (lbrace=147:4,rbrace=148:0) node_offset:94:1
```

We can see that lbrace=147 and rbrace=148, but this is obviously not true since the function is much bigger:
image

This is currently blocking 
https://github.com/ziglang/zig/pull/9652

mattbork:
Fixes #9672. As noted in https://github.com/ziglang/zig/pull/9673#discussion_r703763079, AstGen's source cursor should preferably never be reset to 0. One alternative is to allow `advanceSourceCursor` to also search backward for desired offsets and rewind the cursor. This significantly reduces the number of redundantly scanned bytes but doesn't address the root cause. For that, there are three things that currently break the desired monotonicity of the source cursor, two easily fixed, one not:

1. `fnDecl` (when the body isn't empty) and `testDecl` only call `addFunc` after having already called `expr` on their body nodes. In general, this will advance the source cursor probably many times as the constituent statements are generated. But it's `addFunc` that does the work of finding the opening and closing braces to fill in the function's `SrcLocs`, and at this point it's too late for the opening braces without resetting or rewinding the source cursor. The obvious fix is to have `fnDecl` / `testDecl` find the opening brace themselves and pass the line and column of it to `addFunc`. Since `addFunc` already takes most of its arguments in a struct, I added defaults for the new arguments to reduce the noise in other places that call `addFunc`.

2. `switchExpr` doesn't process its cases in strict source order. Instead the special `else` / `_` case is always done first because it comes first in the trailing extra data in the ZIR. But unless it is also lexically the first case, this will advance the source cursor past the actual first case. The simplest fix is to collect the special case payload in its own `ArrayList`, as the scalar and multi cases do, and process all cases in order in the main for loop. Then the extra data is copied all at once at the end of the function. This also removes a bit of repetition in `switchExpr`.

3. Defer statements, by the nature of their current implementation, require some backtracking of the source cursor. The problem is that a defer statement's body won't actually be generated until a later return statement or try statement or similar calls `genDefers`. By this point the cursor is likely well past the defer's location. And worse, a given defer statement can be regenerated many times. A defer statement at the top of a function with 10 returns will have its expression turned into ZIR afresh 10 times, always with source cursor too far ahead. The simplest mitigation is to save the defer statement's source offset, line, and column in the `Scope.Defer` returned from `makeDeferScope`. Then we wrap all the calls to `unusedResultExpr` from `genDefers` with a helper that saves the current source cursor, sets the cursor the one cached in the defer scope, processes the expression, and then finally restores the previous cursor. Some bytes in the source will still be redundantly scanned by `advanceSourceCursor`, but it will much less than with cursor resetting and still significantly less than with cursor rewinding. And in the case of a defer statement with a single expression, the cursor often doesn't need to be advanced at all.

For a sense of scale, I made builds that count redundantly scanned bytes: one with of these changes, one with of a version of `advanceSourceCursor` that resets the cursor to 0 when the desired `end` is less than the current `source_offset`, and one with an `advanceSourceCursor` that searches backwards when `end` is less than the current offset. Then I had each run ast-check against most of the .zig files in this repo. Many files don't need to rescan any bytes with all the above changes. For example, `std/array_hash_map.zig` will rescan 20,751,654 bytes with cursor resetting; will rescan only 440,705 bytes with cursor rewinding; and will rescan 0 bytes with the above fixes. A significant case was `src/Sema.zig`, which rescanned 401,181,359 bytes with cursor resetting; only 5,552,284 with cursor rewinding; and finally 18,047 with the new fixes. Over all 890 files tested -- a combined 14,218,363 bytes -- these fixes do a cumulative 60,149 bytes of redundant scanning; cursor rewinding does 71,503,770 bytes; and the O(N^2) worst case of cursor resetting does 1,911,917,938 bytes.

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -1683,7 +1683,7 @@ fn breakExpr(parent_gz: *GenZir, parent_scope: *Scope, node: Ast.Node.Index) Inn
                 const defer_scope = scope.cast(Scope.Defer).?;
                 scope = defer_scope.parent;
                 const expr_node = node_datas[defer_scope.defer_node].rhs;
-                _ = try unusedResultExpr(parent_gz, defer_scope.parent, expr_node);
+                try unusedResultDeferExpr(parent_gz, defer_scope, defer_scope.parent, expr_node);
             },
             .defer_error => scope = scope.cast(Scope.Defer).?.parent,
             .top => unreachable,
@@ -1736,7 +1736,7 @@ fn continueExpr(parent_gz: *GenZir, parent_scope: *Scope, node: Ast.Node.Index) 
                 const defer_scope = scope.cast(Scope.Defer).?;
                 scope = defer_scope.parent;
                 const expr_node = node_datas[defer_scope.defer_node].rhs;
-                _ = try unusedResultExpr(parent_gz, defer_scope.parent, expr_node);
+                try unusedResultDeferExpr(parent_gz, defer_scope, defer_scope.parent, expr_node);
             },
             .defer_error => scope = scope.cast(Scope.Defer).?.parent,
             .namespace => break,
@@ -1922,8 +1922,8 @@ fn blockExprStmts(gz: *GenZir, parent_scope: *Scope, statements: []const Ast.Nod
             .simple_var_decl  => scope = try varDecl(gz, scope, statement, &block_arena.allocator, tree.simpleVarDecl(statement)),
             .aligned_var_decl => scope = try varDecl(gz, scope, statement, &block_arena.allocator, tree.alignedVarDecl(statement)),
 
-            .@"defer"    => scope = try makeDeferScope(scope, statement, &block_arena.allocator, .defer_normal),
-            .@"errdefer" => scope = try makeDeferScope(scope, statement, &block_arena.allocator, .defer_error),
+            .@"defer"    => scope = try makeDeferScope(gz.astgen, scope, statement, &block_arena.allocator, .defer_normal),
+            .@"errdefer" => scope = try makeDeferScope(gz.astgen, scope, statement, &block_arena.allocator, .defer_error),
 
             .assign => try assign(gz, scope, statement),
 
@@ -1949,6 +1949,22 @@ fn blockExprStmts(gz: *GenZir, parent_scope: *Scope, statements: []const Ast.Nod
 
     try genDefers(gz, parent_scope, scope, .normal_only);
     try checkUsed(gz, parent_scope, scope);
+}
+
+fn unusedResultDeferExpr(gz: *GenZir, defer_scope: *Scope.Defer, expr_scope: *Scope, expr_node: Ast.Node.Index) InnerError!void {
+    const astgen = gz.astgen;
+    const prev_offset = astgen.source_offset;
+    const prev_line = astgen.source_line;
+    const prev_column = astgen.source_column;
+    defer {
+        astgen.source_offset = prev_offset;
+        astgen.source_line = prev_line;
+        astgen.source_column = prev_column;
+    }
+    astgen.source_offset = defer_scope.source_offset;
+    astgen.source_line = defer_scope.source_line;
+    astgen.source_column = defer_scope.source_column;
+    _ = try unusedResultExpr(gz, expr_scope, expr_node);
 }
 
 /// Returns AST source node of the thing that is noreturn if the statement is definitely `noreturn`.
@@ -2333,7 +2349,7 @@ fn genDefers(
                 const prev_in_defer = gz.in_defer;
                 gz.in_defer = true;
                 defer gz.in_defer = prev_in_defer;
-                _ = try unusedResultExpr(gz, defer_scope.parent, expr_node);
+                try unusedResultDeferExpr(gz, defer_scope, defer_scope.parent, expr_node);
             },
             .defer_error => {
                 const defer_scope = scope.cast(Scope.Defer).?;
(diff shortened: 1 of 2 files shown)


andrewrk:
Thank you for these changes. The quality of the work, the detailed analysis, and the extra time taken to check the effectiveness of the change make merging this a complete joy.

PR #9885, opened 2021-10-03
Big int wrapping/saturating

Snektron:
Implements a whole bunch of wrapping and saturating related big integer code:
- Big integer truncate and comptime `@truncate` for self hosted.
- Wrapping addition and subtraction, including comptime `+%` and `-%` for self hosted.
- `setTwosCompIntLimit()` to set a big integer to either limit of a 2s-complement big integer.
- Saturating addition and subtraction, also adapted to comptime `+|` and `-|` for self hosted.
- Improved version of the Karatsuba multiplication algorithm. This version requires one less allocation per recursion, and the result is guaranteed to fit in `a.len + b.len`. A very quick test showed about half the allocated memory and about 10% performance improvement on my machine. Note that this version does require some more code, and so a comptime parameter is introduced to all `ll[mul]accum*` related functions to either perform addition or subtracting accumulation.
- Improved comments in the Karatsuba function.
- The `llmulaccum` and related functions (including `llmulaccumKaratsuba`) are adapted to perform the multiplication modulo `r.len` to reduce required memory for wrapping multiplication.
- Wrapping multiplication, including comptime `*%` for self hosted.
- `saturate()` function to saturate a big integer to a certain configuration. I don't think Karatsuba can be adapted for saturating like it can for wrapping multiplication, and so i think for saturating multiplication the full multiplication needs to be performed.  `saturate()` is useful to saturate the result to the desired integer configuration.
- A whole bunch of tests for the new code.

### DIFF
--- a/lib/std/math/big.zig
+++ b/lib/std/math/big.zig
@@ -5,6 +5,7 @@ pub const Rational = @import("big/rational.zig").Rational;
 pub const int = @import("big/int.zig");
 pub const Limb = usize;
 const limb_info = @typeInfo(Limb).Int;
+pub const SignedLimb = std.meta.Int(.signed, limb_info.bits);
 pub const DoubleLimb = std.meta.Int(.unsigned, 2 * limb_info.bits);
 pub const SignedDoubleLimb = std.meta.Int(.signed, 2 * limb_info.bits);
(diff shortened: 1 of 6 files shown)


andrewrk:
Brilliant work.

PR #9895, opened 2021-10-04
translate-c: improve error messages when expecting specific tokens

FnControlOption:
Old:

    unable to translate C expr: expected ')'

New:

    unable to translate C expr: expected .RParen instead got .Ellipsis

Also, is `return m.fail(...);` supposed to be this instead?

```zig
try m.fail(...);
return error.ParseError;
```

I replaced two instances of `return m.fail(...);` in this pull request, but there are four more matches that I haven't changed.

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -5274,6 +5274,18 @@ const MacroCtx = struct {
         return self.list[self.i].id;
     }
 
+    fn skip(self: *MacroCtx, c: *Context, expected_id: std.meta.Tag(CToken.Id)) ParseError!void {
+        const next_id = self.next().?;
+        if (next_id != expected_id) {
+            try self.fail(
+                c,
+                "unable to translate C expr: expected '{s}' instead got '{s}'",
+                .{ CToken.Id.symbol(expected_id), next_id.symbol() },
+            );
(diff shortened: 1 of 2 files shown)


andrewrk:
Looks fine to me- can I get an OK from @Vexu or @ehaas ?

src/translate_c.zig
```diff
@@ -5245,6 +5245,14 @@ const MacroCtx = struct {
         return self.list[self.i].id;
     }
 
+    fn skip(self: *MacroCtx, c: *Context, expected_id: CToken.Id) ParseError!void {
```
ehaas, line 5248: Since `expected_id` will always be a a valueless tag, you can declare its type as `std.meta.Tag(CToken.Id)` (so that it's just the enum) and then the if statement can be `if (next_id != expected_id) {`

Vexu:
>New:
>    unable to translate C expr: expected .RParen instead got .Ellipsis

The token id has a `symbol` function which you should use instead of `@tagName`.

> Also, is `return m.fail(...);` supposed to be this instead?

In the two instances you changed it doesn't make amy difference since the caller just continues looping on both no error and `error.ParsingFailed`.

PR #9904, opened 2021-10-06
print_zir.zig: fill in remaining stubs

mattbork:
Covers all ZIR instructions printing "TODO". The only function that isn't strictly by rote is `writePtrType`, which marks up some of its trailing `Ref` fields as they would appear in a pointer type expression. Also fixes a couple unbalanced parentheses and brackets.


andrewrk:
Wonderful, thank you!

PR #9919, opened 2021-10-08
add test step to build.zig generated by init-exe

leecannon:
I think it makes sense to generate a `zig build test` step for `init-exe` just like `init-lib` does.
For almost every zig project adding a test step is one of the first things I've done.


g-w1:
lgtm

PR #9927, opened 2021-10-10
Fix rlimit_resource for MIPS and SPARC

vrischmann:
This PR fixes the definition of `rlimit_resource` for the MIPS and SPARC architectures.

I took the values from the glibc headers:
 * [MIPS](https://github.com/ziglang/zig/blob/master/lib/libc/include/mips-linux-gnu/bits/resource.h)
 * [SPARC](https://github.com/ziglang/zig/blob/master/lib/libc/include/sparcv9-linux-gnu/bits/resource.h)

I'm not sure the second commit should be included. It prevents a test failure on MIPS because of some behaviour [in musl](http://git.musl-libc.org/cgit/musl/commit/src/misc/getrlimit.c?id=8258014fd1e34e942a549c88c7e022a00445c352) which makes it so that `RLIM_INFINITY` is returned instead of the actual value if it goes above `2147483647` (only for MIPS 32 bit).

On my desktop this is the case for `RLIMIT_MEMLOCK`, so what happens is that the test then tries to setrlimit `RLIMIT_MEMLOCK` to `RLIM_INFINITY` and it fails with a permission denied.
It's easy to workaround this with systemd-run but I'd like the opinion of other people on this.

### DIFF
--- a/lib/std/os/linux.zig
+++ b/lib/std/os/linux.zig
@@ -4197,65 +4197,68 @@ pub const ifreq = extern struct {
 };
 
 // doc comments copied from musl
-pub const rlimit_resource = enum(c_int) {
-    /// Per-process CPU limit, in seconds.
-    CPU,
+pub const rlimit_resource = if (native_arch.isMIPS() or native_arch.isSPARC())
+    arch_bits.rlimit_resource
+else
+    enum(c_int) {
+        /// Per-process CPU limit, in seconds.
+        CPU,
 
-    /// Largest file that can be created, in bytes.
-    FSIZE,
+        /// Largest file that can be created, in bytes.
+        FSIZE,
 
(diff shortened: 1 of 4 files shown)


andrewrk:
> some behaviour [in musl](http://git.musl-libc.org/cgit/musl/commit/src/misc/getrlimit.c?id=8258014fd1e34e942a549c88c7e022a00445c352)

Is this a bug to report upstream to musl?

vrischmann:
The musl commit is a fix itself for something broken in the kernel so I don't think there's anything to report upstream.

The test assumes you can `setrlimit` the same value that `getrlimit` returns, but in the case of MIPS with a sufficiently large `RLIMIT_MEMLOCK` this isn't true since `getrlimit` returns `RLIM_INFINITY`.

PR #9928, opened 2021-10-10
Add `null` and `undefined` to Zig keywords list

samrat:
So that if `translate-c` encounters identifier `null` or `undefined`, it is formatted as `@"null"` and `@"undefined"`.


andrewrk:
Hi @samrat, thanks for the pull request. `null` and `undefined` are not keywords; they are primitive values, the same as `true` and `false`. The change that needs to be made here is making them *not* keywords in the stage1 compiler, rather than incorrectly making them keywords in self-hosted.

samrat:
@andrewrk Thanks for the response. Would that mean the `null`/`undefined` values would get represented as something else?

It seems that this check https://github.com/ziglang/zig/blob/master/src/translate_c/ast.zig#L823 needs modifying as it currently does not escape `null` and `undefined`?

This is breaking the translation of https://github.com/Hejsil/zig-nuklear/blob/master/src/c/nuklear/nuklear.h#1183

andrewrk:
`isZigPrimitiveType` had a bug where it checked the integer names (e.g. `u32`) before primitives, leading it to incorrectly return `false` for `undefined` which starts with `u`.

Fixed in d1fd864da7859989828feb8be806634606cc761c  :+1:

PR #9929, opened 2021-10-11
DRAFT: Simplified Clock, CalendarTime, and leap second correctness

### LINKED ISSUE #8781: time.nanoTimestamp epoch should not have leap seconds (propose GPS time or TAI)
UTC has leap seconds, which are a pain to deal with in a few corner cases. This results in an ambiguous description of time.nanoTimestamp. Right now, it's hard for two separate services written in Zig to share timestamps without also having to share how many leap seconds were included in the timestamp (there's also no std library API to determine how many leap seconds are included in a timestamp, so even that is hard).

Proposal:
1. The Zig std library time API standardize on a consistent epoch with no ambiguities, such as GPS time or TAI. By choosing a more consistent epoch, the API will continue to work consistently even with an executable that was compiled prior to a leap second being declared, but executed AFTER the leap second occurs.
2. For formatting of timestamps to a user-readable format, there would need to be a Zig std library API for converting nanoTimestamps to year/month/day/hour/min/sec/ns. This API would need to defer to the OS in order to take leap second updates into account (see mktime() on Linux or FileTimeToSystemTime() on Windows), since only the OS would be expected to get dynamic leap second updates as future ones are declared and be able to correctly handle them.

Windows now correctly supports leap seconds (as of October 2018 update). FILETIME is always in the NTFS/Windows Epoch, which does not jump around with leap seconds.

On Linux, CLOCK_REALTIME tries to follow UTC, which jumps around with leap seconds. CLOCK_TAI is supported since Linux 3.10, and follows International Atomic Time (no leap seconds).

PhaseMage:
This is a draft PR for changes I picked up from @kprotty and then ran with.

Primary Features:
1. Use os.clock_gettime() for POSIX compatibility (seconds + nsec since Unix Epoch)
2. Fixes Windows implementation of os.clock_gettime() to be correct even after a leap second.
3. Instead of nanoTimestamp() use nanoMonotonic() and milliMonotonic() for relative timing stamps
4. timestampNow() defines a new Zig std Canonical u96 timestamp with consistent leap second behavior where possible (Fixes #8781)
5. CalendarTime API to convert to/from RFC 3339 DateTime strings (eg. 2012-09-27T20:38:40) (supports the '60' leap second).
6. CalendarTime API to convert to/from zig std timestamps
7. Implements @kprotty 's Timer and Instant API for higher performance relative timing stamps
8. Some Cache.zig fixes that I'll break out into a separate PR

**Q: Why use a different Epoch for the canonical Zig std timestamps?**
1. The Unix Epoch doesn't handle leap seconds in a consistent way
2. Representing CalendarTime before 1972 (still in the modern era) requires awkward negative numbers
3. Starting in 1601 (right at the start of a Gregorian calendar cycle) makes the math easy
4. 1601 allows nearly all points in time that are known with day precision to be accurately specified (dates prior to the start of the Gregorian calendar are very messy).

**Q: Why use fixed-point u96 (36.60) for Zig timestamps?**
A: Fixed-point math allows for optimal performance along with select-able precision. (Need ~4ns precision? Just shift by 32).

I just noticed that @marler8997 did a calendar implementation a month ago. Except for lack of leap second support, the implementation looks correct. However it feels a bit verbose compared to a single API that just converts a timestamp to a CalendarTime struct. I profiled it, and found that the single API is ~11x faster and feels more ergonomic (to me) to use.

This DRAFT PR is so I can get feedback from specific folks if I'm on the right track or not.


PhaseMage:
> 
> 
> I'm not too sure about this invention of UTC2018
> 
> Aside: I wonder if it's better to incorporate an uncertainty/precision/range field to timestamps. That could cover not only uncertainty due to leap seconds ("UtcFuzzy" as you've called it); but also truncated timestamps (due to e.g. FAT); and timestamps from `gettimeofday` which as an API is only 1e-6 vs timestamps from `clock_gettime` which max out at 1e-9 but should consider `clock_getres`

Thanks for the feedback. I'd love to help explain what Utc2018 is. I didn't really invent it, I'm just giving it a shorter name (from "UTC as of 2018"). Windows invented it when they decided that FILETIME no longer jumps with future leap seconds: Utc2018 is basically FILETIME as of 2018 and later. Prior to 2018, Windows had a 'fuzzy' epoch definition just like POSIX still does. What happens right now, in reality, is that a brand new Unix Epoch is created every time there is a leap second. The same thing happened in Windows as well before 2018. For example, if a leap second is declared for Dec 31 2022, then a new Unix Epoch with the definition of "UTC as of Jan 2023" will be created out of thin air. Standardizing on Utc2018 is an attempt to make this into a workable consistent system. If leap seconds are eliminated in the future, then Utc2018 will become enshrined as the one and true UTC for all timestamps 2018 and forevermore.

Aside: I'd love to turn timestamps into a struct with helpful conversion APIs. Having additional 'uncertainty' field would fit nicely with that. I think it would make sense to do that work only after successfully merging this code. Unless folks think it's a merge blocker?

PhaseMage:
I'm happy we're on the same page now on the inconsistent behavior of FILETIME relative to REALTIME 😀

So the point now is to build off of that knowledge. What would be the valid usecases of a timestamp with an inconsistent Epoch like that? I submit that there aren't many. Again, I go back to trying to write the doc-comment for such a timestamp.

kprotty:
SystemTime can be stored and transmitted, but what it means to recipients not on the same machine is unspecified. File.Stat could then use SystemTime instead as well.

> The only benefit seems to be to save conversion costs in case the scenario only involves comparison ops with File.Stat times

There's other uses for SystemTime besides File.Stat: Http date values that don't need leap second correction, Millisecond or more precision for recording events (i.e. local snowflakes or request tracking), etc.

>  the extra layer of indirection imposed on the developer [...] if the ITU ever decides to eliminate Leap Seconds, we're left with this SystemTime layer that doesn't add value.

I'm not sure where the level of indirection comes from: `SystemTime.now().toCalendarTime()` vs `CalendarTime.fromStdTime(timestasmpNow())`

> For comparing times within a running process, it seems prudent to use Timer instead.

Timer and Instant would be for recording elapsed time, not a date time.

> you'd want to use Instant.now() for the initial tagging, and then have Instant expose a asStdTime() method that does the translation to Utc2018

 "StdTime" decreases the precision of the sample in order to keep a cross-platform representation. The "Instant.asStdTime" impl assumes that the first system timestamp is correctly synchronized with external clocks or wont change at runtime. Having the recorded timestamps be the SystemTime keep's the OS system clock precision and tracks any updates to it.

kprotty:
(warning: pile of thoughts) @PhaseMage  So I took your suggestions and got rid of the FILETIME/timespec layer, making "SystemTime" (name up for change) effectively be your "StdTime" but with a slight change:

* `SystemTime` now represents time since Jan 1 1601 *excluding leap seconds*
* `SystemTime.toCalendarTime` converts it to a date format *which accounts for leap seconds*
  * This uses `FileTimeToSystemTime` which understands leap seconds from [Windows Update & the like](https://techcommunity.microsoft.com/t5/networking-blog/leap-seconds-for-the-appdev-what-you-should-know/ba-p/339813)
  * This uses the compile time table for leap seconds to convert to Utc2018
  * Theres a doc comment on this function that it needs to be recompiled if a leap second is added (distinction: `SystemTime.now` does not require recomp.)
  * Q: is there a use case which would prefer leap seconds being in the timestamp itself (i.e. your timespecNow) or are leap seconds mostly for the serialization (done through CalendarTime) ?
  
Some other changes w.r.t the first alternative thing:

* `Instant` is documented to specifically include time spent suspended
* `CpuTime.system` was added for "uptime of system without suspend"

Thoughts?

<details>
<summary> API revisited </summary>

```zig
// Time since Jan 1 1601, assuming no leap seconds occured.
// Saturates to said epoch if the system reports it before.
// Precision, monotonicity, and speed depend on OS.
const SystemTime:
    native: u64 // fixed point (36.28)

    fn now(): SystemTime
        if (windows) GetSystemTimePreciseAsFileTime
        clock_gettime(REALTIME)

    fn order(other: SystemTime): std.math.Order
        ...

    fn since(earlier: SystemTime): ?u64 // fixed point (36.28)
        if (order() == .lt) return null
        ...

    // Converts no-leap-second SystemTime into leap-second aware CalendarTime.
    // Leap second awareness is Utc2018 so if new one is introduced, must recompile.
    fn toCalendarTime(): CalendarTime
        if (windows) FileTimeToSystemTime() // OS is aware of leap seconds
        convertUtcFuzzyToUtc2018()

// Date time that can be serialized and human-inspected
const CalendarTime:
    years: uN,
    month: MonthEnum,
    ...

    fn order(other: CalendarTime): std.math.Order
        ...

    fn since(earlier: CalendarTime): ?u64 // fixed point (36.28)
        if (order() == .lt) return null

// High performance timestamp that starts at an arbitrary point in time.
// Useful for measuring elapsed time withing a running program.
// Ticks while the system is suspended.
// Precision, monotonicity, and speed depend on OS.
const Instant:
    native: if (windows or darwin) u64 else timespec

    fn now():
        if (darwin) mach_continuous_time()
        if (windows) QueryPerformanceCounter()
        clock_gettime(if (linux) BOOTTIME else MONOTONIC)

    fn order(other: Instant) std.math.Order
        if (windows or darwin) order(u64)
        order(timespec)

    fn since(earlier: Instant) ?u64 // nanoseconds (should used fixed point here too?)
        ...

const CpuTime(enum):
    thread, // uptime in nanos for caller thread
    process, // uptime in nanos for caller process
    system, // uptime in nanos for the system (excluding time spent suspended)

    fn read(self: CpuTime) !u64
        // netbsd & wasi return error.Unsupported for `system`
        ...
```

</details>

PhaseMage:
@kprotty Did you get a chance to chat with Andrew about this proposal? I can do more work here to get it across the finish line.

andrewrk:
closing old draft

PR #9930, opened 2021-10-11
Fix isProblematicTimestamp

### LINKED ISSUE #6082: Cache isProblematicTimestamp false negative
In a916f63940ed0cf039ab5f70fbee022f604995ea I fixed an obvious bug which uncovered a nefarious latent bug:

```
220/1502 cache_hash.test "std-native-Debug-c-multi check that changing a file makes cache fail"...expected null, found GLuqBm2tKRkCtIQkzLj14o2jcZqFlknF
/drone/src/lib/std/testing.zig:137:36: 0x4ec287 in testing.expectEqual (test)
                    std.debug.panic("expected null, found {}", .{actual_payload});
                                   ^
/drone/src/lib/std/cache_hash.zig:596:28: 0x3c963f in cache_hash.test "std-native-Debug-c-multi check that changing a file makes cache fail" (test)
        testing.expectEqual(@as(?[32]u8, null), try ch.hit());
                           ^
/drone/src/lib/std/special/test_runner.zig:56:28: 0x44189f in std.special.main (test)
        } else test_fn.func();
                           ^
/drone/src/lib/std/start.zig:252:37: 0x4432a7 in start.main (test)
                                    ^
/drone/src/lib/libc/musl/src/env/__libc_start_main.c:94:7: 0xdfd90f in libc_start_main_stage2 (/drone/src/lib/libc/musl/src/env/__libc_start_main.c)
	exit(main(argc, argv, envp));
      ^

Tests failed. Use the following command to reproduce the failure:
/drone/src/zig-cache/o/7O-hvqSImgvR7hUjZnC1Q3KhHiiX6mD6rKD9pyn_tjYSpznCIZWAaHyAEa6I31Sh/test
The following command exited with error code 255:
/drone/src/build/zig test /drone/src/lib/std/std.zig --library c --test-name-prefix std-native-Debug-c-multi  --cache-dir /drone/src/zig-cache --name test -I /drone/src/test --override-lib-dir /drone/src/lib 

```

I tried to fix it with ce8b9c0c5cdbe4161952e6f2aa875f722949d4cb but it's still happening sometimes.

I think it's time to step back and figure out how this can work properly. As it stands, `isProblematicTimestamp` is not reliable. Empirically, I tried this code:

```zig
const std = @import("std");
const fs = std.fs;

pub fn main() !void {
    const cwd = fs.cwd();
    const temp_file = "hello.txt";
    var timer = try std.time.Timer.start();
    var count: usize = 0;
    try cwd.writeFile(temp_file, temp_file);
    var mtime = try getmtime(temp_file);
    while (true) {
        _ = timer.lap();
        try cwd.writeFile(temp_file, temp_file);
        const new_mtime = try getmtime(temp_file);
        if (new_mtime != mtime) {
            mtime = new_mtime;
            break;
        }
    }
    while (true) {
        try cwd.writeFile(temp_file, temp_file);
        count += 1;
        const new_mtime = try getmtime(temp_file);
        if (new_mtime != mtime) {
            break;
        }
    }
    const ns = timer.lap();
    std.debug.print("count={} ns={}\n", .{ count, ns });
}

fn getmtime(temp_file: []const u8) !i128 {
    const file = try fs.cwd().openFile(temp_file, .{});
    defer file.close();
    const stat = try file.stat();
    return stat.mtime;
}
```

Got output like this:

```
[nix-shell:~/Downloads/zig/build]$ ./zig build-exe test.zig --release-fast

[nix-shell:~/Downloads/zig/build]$ ./test
count=12 ns=1768757

[nix-shell:~/Downloads/zig/build]$ ./test
count=12 ns=1010152

[nix-shell:~/Downloads/zig/build]$ ./test
count=13 ns=1874677

[nix-shell:~/Downloads/zig/build]$ ./test
count=2 ns=1005522

[nix-shell:~/Downloads/zig/build]$ ./test
count=13 ns=1834186

[nix-shell:~/Downloads/zig/build]$ ./test
count=13 ns=1071032

[nix-shell:~/Downloads/zig/build]$ ./test
count=2 ns=1013771
```

The meaning of this is that over the span of about 1 millisecond, I was able to write and then read the same file 13 times without the mtime changing, despite the file system purporting to have nanosecond granularity. Womp womp.

PhaseMage:
Two issues with the isProblematicTimestamp():

1. It was looking for trailing zero bits when it should be looking for trailing decimal zeros.
2. Clock timestamps had more precision than the actual file timestamps

The fix is to grab a timestamp from a 'just now changed' temp file. This timestamp is "problematic". Any file timestamp greater than or equal to this timestamp is considered problematic. File timestamps **prior** to this **can** be trusted.

Downside is that it causes a disk I/O to write to and then read the timestamp from this file ~1ms on my system. This is partially mitigated by keeping track of the most recent problematic timestamp, and only checking for a new problematic timestamp when checking a timestamp that is equal to or larger than the last problematic one.

This fixes #6082

### DIFF
--- a/src/Cache.zig
+++ b/src/Cache.zig
@@ -187,11 +187,14 @@ pub const Manifest = struct {
     /// of the files listed in the manifest.
     failed_file_index: ?usize = null,
 
+    /// most recent problematic timestamp
+    recent_problematic_timestamp: i128 = 0,
+
     /// Add a file as a dependency of process being cached. When `hit` is
     /// called, the file's contents will be checked to ensure that it matches
     /// the contents from previous times.
     ///
-    /// Max file size will be used to determine the amount of space to the file contents
+    /// Max file size will be used to determine the amount of space the file contents
     /// are allowed to take up in memory. If max_file_size is null, then the contents
     /// will not be loaded into memory.
     ///
@@ -414,7 +417,8 @@ pub const Manifest = struct {
 
                 cache_hash_file.stat = actual_stat;
 
-                if (isProblematicTimestamp(cache_hash_file.stat.mtime)) {
+                if (try self.isProblematicTimestamp(cache_hash_file.stat.mtime)) {
+                    // The actual file has an unreliable timestamp, force it to be hashed
                     cache_hash_file.stat.mtime = 0;
                     cache_hash_file.stat.inode = 0;
                 }
@@ -485,7 +489,8 @@ pub const Manifest = struct {
 
         ch_file.stat = try file.stat();
 
-        if (isProblematicTimestamp(ch_file.stat.mtime)) {
+        if (try self.isProblematicTimestamp(ch_file.stat.mtime)) {
+            // The actual file has an unreliable timestamp, force it to be hashed
             ch_file.stat.mtime = 0;
             ch_file.stat.inode = 0;
         }
@@ -520,7 +525,7 @@ pub const Manifest = struct {
     }
 
     /// Add a file as a dependency of process being cached, after the initial hash has been
-    /// calculated. This is useful for processes that don't know the all the files that
+    /// calculated. This is useful for processes that don't know all the files that
     /// are depended on ahead of time. For example, a source file that can import other files
     /// will need to be recompiled if the imported file is changed.
     pub fn addFilePostFetch(self: *Manifest, file_path: []const u8, max_file_size: usize) ![]const u8 {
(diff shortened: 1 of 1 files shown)


jibal:
Wouldn't it be faster to write to an existing temp file rather than creating one every time?

andrewrk:
@jibal check the implementation. It does that optimization and more.

PR #9932, opened 2021-10-11
translate-c: Perform `isPrimitive` check for record fields

samrat:
A check was being performed when converting C identifiers since the code uses `zig.fmt.fmtId`. However, a check for primitives(eg. `null`, `undefined`) is also required.

This PR makes the change to use the `addIdentifier` wrapping method which escapes the token if it is a Zig primitive.

Related: https://github.com/ziglang/zig/pull/9928

### DIFF
--- a/test/translate_c.zig
+++ b/test/translate_c.zig
@@ -328,7 +328,7 @@ pub fn addCases(cases: *tests.TranslateCContext) void {
         \\pub const LIGHTGRAY = @import("std").mem.zeroInit(CLITERAL(Color), .{ @as(c_int, 200), @as(c_int, 200), @as(c_int, 200), @as(c_int, 255) });
         ,
         \\pub const struct_boom_t = extern struct {
-        \\    i1: c_int,
+        \\    @"i1": c_int,
         \\};
         \\pub const boom_t = struct_boom_t;
         ,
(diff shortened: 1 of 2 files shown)


samrat:
Hello, is there anything else that needs to be added to this PR?

test/translate_c.zig
```diff
@@ -328,7 +328,7 @@ pub fn addCases(cases: *tests.TranslateCContext) void {
         \\pub const LIGHTGRAY = @import("std").mem.zeroInit(CLITERAL(Color), .{ @as(c_int, 200), @as(c_int, 200), @as(c_int, 200), @as(c_int, 255) });
         ,
         \\pub const struct_boom_t = extern struct {
-        \\    i1: c_int,
+        \\    @"i1": c_int,
```
andrewrk, line 331: This is incorrect; the previous test case was correct. Same for `null` - zig allows `i1` and `null` to be field names without using `@""` syntax.

andrewrk (changes requested):
> Hello, is there anything else that needs to be added to this PR?

Yes, this is the wrong solution. Instead of this, stage1 needs to be improved to accept fields with same names as primitives such as `null` and `i1`.

Self-hosted already has this working.

PR #9939, opened 2021-10-12
Remove resolved_path from import table on error

### LINKED ISSUE #9404: Relative import segfaults compiler
Hi,

The following causes a segfault in the compiler:
```zig
const below = @import("../above.zig");

test "" {}
```
An empty testcase is needed or it won't trigger.
above.zig does not even need to exist.
Tested with latest master branch: 0.9.0-dev.473+9979741bf

zig test below.zig:
```zig
Segmentation fault at address 0x0
/home/flandere/repos/zig/src/Module.zig:1025:13: 0x55d2ff5e9f04 in Module.File.unloadTree (zig1)
            if (file.tree_loaded) {
            ^
/home/flandere/repos/zig/src/Compilation.zig:1734:32: 0x55d2ff5e100b in Compilation.update (zig1)
                file.unloadTree(self.gpa);
                               ^
/home/flandere/repos/zig/src/main.zig:2368:20: 0x55d2ff5777ff in main.updateModule (zig1)
    try comp.update();
                   ^
/home/flandere/repos/zig/src/main.zig:2100:17: 0x55d2ff4a5de3 in main.buildOutputType (zig1)
                ^
/home/flandere/repos/zig/src/main.zig:209:31: 0x55d2ff47ae16 in main.mainArgs (zig1)
                              ^
/home/flandere/repos/zig/src/stage1.zig:49:24: 0x55d2ff47a5d9 in main (zig1)
                       ^
```

I hope that's enough information to help, thanks in advance for the time spent :)

Have a nice day!

SamTebbs33:
This PR fixes a case where stage1 segfaults due to an undefined import table entry being left behind when `Module.importFile()` returns an error. The credit for this fix should go to @FlandreScarlet but it has been almost three months since the fix was proposed so I submitted a PR instead.

Fixes #9404

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -3502,6 +3502,7 @@ pub fn importFile(
     defer if (!keep_resolved_path) gpa.free(resolved_path);
 
     const gop = try mod.import_table.getOrPut(gpa, resolved_path);
+    errdefer _ = mod.import_table.pop();
     if (gop.found_existing) return ImportFileResult{
         .file = gop.value_ptr.*,
         .is_new = false,


andrewrk (changes requested):
Thanks for the patch.

Problems:

 * Same logic needs to be done to `importPkg`.
 * Needs test coverage.
 * `keep_resolved_path` needs to get moved down lower.

I have all these issues addressed locally; will push momentarily.

PR #2331, opened 2019-04-21
readme: Add instructions for making changes to the standard library

squeek502:
Closes #2324

---

- I didn't add a `Developing Zig > Compiler` section since I'm not familiar with that aspect but that's an obvious addition that would also be helpful.
- I went with the `make install` process rather than `--override-std-dir` since that's what @andrewrk prefers: https://github.com/ziglang/zig/issues/2324#issuecomment-485282490
- ~~Wasn't sure how best to signal 'make changes to a file' in the example so I just used `nano`. Definitely open to better ideas on that.~~ (edit: example was removed)

Issue #2324, opened 2019-04-21
[question/suggestion] Workflow when working on std lib


squeek502:
Not sure if this is documented anywhere, but it would be a big help if it were, as I'm unsure of what the intended way to make changes to the std library and test them actually is.

On Linux I was making changes to the files in `zig/std` and then doing `make install` after every change before running e.g. `bin/zig test lib/zig/std/heap.zig` from my `zig/build` directory.

On Windows I'm building a release mode zig compiler, adding `zig/build-release/bin` to `PATH` and then doing e.g. `zig test std/heap.zig --override-std-dir std` from the checked out `zig` directory.

Curious what core contributors are doing.

shritesh:
I asked @andrewrk this question during one of his streams and he showed us his workflow and I have adapted mine to be similar to what he described.

I create a `build` directory in the repository root (this is in `.gitignore` by default) and work from there.

If I'm working in userland (not the compiler), I create a Release build of the compiler using `cmake .. -DCMAKE_PREFIX_PATH=/usr/local/Cellar/llvm/8.0.0 -DCMAKE_BUILD_TYPE=Release` from the `build` directory, (I am on a Mac using Homebrew, so I specify the LLVM path that way. YMMV), and do a `make -j$(nproc) install`. Then I use the `./zig` binary and specify `--override-std-dir ../std` to every zig command (`test`, `build*`, etc) that I run.

Before submitting a pull request, I do a `make -j$(nproc) install` again, and run the tests with `./zig build --build-file ../build.zig test -Dskip-release`. The `-Dskip-release` skips the test matrix of building and testing on every possible platform and only does so for the host platform. This is what is expected out of contributors.

When working on the compiler itself, I skip the `-DCMAKE_BUILD_TYPE=Release` so that I can get better stack traces for a tiny-bit slower compilation (maybe this is no longer needed with the hybrid stage 1) .

Hope this helps. I too think this should be documented somewhere.

andrewrk:
### My process

I follow the readme instructions.

#### POSIX

I definitely recommend the default of debug builds of Zig + release build of LLVM. Debug builds of LLVM are useful sometimes, but not usually.

My CWD is the build directory. After every std lib edit, `make install` and then test with `bin/zig`. (same as @squeek502)

#### Windows

Unfortunately I was unable to make it work with a release build of LLVM + debug build of Zig. So I actually test with zig in release mode.

[...]

PR #9946, opened 2021-10-14
Adds tip about testing stdlib files to CONTRIBUTING.md

andyfleming:
This hint would have been helpful to me when starting to contribute.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,7 +110,7 @@ this information and more in the `--help` menu.
 To quickly test a change to a file in the standard library, you can run zig test and specify a custom lib directory with the follow command-line argument.
 
 ```bash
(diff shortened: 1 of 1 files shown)


squeek502:
I think this is a good addition. I use a variant of it all the time and it really speeds things up when you're working on one specific thing. Something like it was originally going to be included in `CONTRIBUTING.md` but [`make install` was recommended](https://github.com/ziglang/zig/issues/2324#issuecomment-485282490) instead ([the original PR](https://github.com/ziglang/zig/pull/2331))

The variant I use is this, from within the `build` dir:

```
./zig test ../lib/std/fs/test.zig --zig-lib-dir ../lib --main-pkg-path ../lib/std
```

Note that adding `--main-pkg-path lib/std` to your version might make it slightly safer, so that you can't accidentally escape the std lib directory with imports.

PR #9962, opened 2021-10-16
Comptime compareFn for PriorityQueue and PriorityDequeue

### LINKED ISSUE #9916: compile time compare function for `std.priority_queue`
# Introduction
Creating a priority queue is done in two steps:
1. Creating a monomorphized type of the queue.
2. Initializing an instance.
The comparison function used to internally sort the elements by priority is currently provided in step 2. Providing the comparison function at runtime is more flexible, but force all queues to use a function pointer when comparing elements and disallows many optimizations.
# Proposal
Move the comparison function to be a `comptime` argument in step 1.
# Advantages
Significant performance improvements for use cases where a simple comparison function is used (direct comparison of elements, comparison of some field(s) of a struct)
# Objections
## Less flexibility
Forcing the comparison function to be known at compile time reduces the flexibility of priority queue. I do not know the impact of this change as all my use cases of priority queue know the comparison function at compile time. If your use case requires runtime comparison function selection, please do let me know!
This could be solved using a compile time stub which dispatches to a runtime comparison function. I have not tried implementing such a stub and it may not be possible.
## Breaking change
This is a breaking change for all code which uses `std.priority_queue`.  All users of `std.priority_queue` will have to be updated, at least the initialization code.
Some use cases may no longer be possible, see point above. Please report such use cases so they can be discussed.

hollmmax:
resolves #9916

### DIFF
--- a/lib/std/priority_dequeue.zig
+++ b/lib/std/priority_dequeue.zig
@@ -9,30 +9,27 @@ const expectEqual = testing.expectEqual;
 const expectError = testing.expectError;
 
 /// Priority Dequeue for storing generic data. Initialize with `init`.
-pub fn PriorityDequeue(comptime T: type) type {
+/// Provide `compareFn` that returns `Order.lt` when its first
+/// argument should get min-popped before its second argument,
+/// `Order.eq` if the arguments are of equal priority, or `Order.gt`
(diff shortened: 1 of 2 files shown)


andrewrk:
Thanks!

PR #9473, opened 2021-07-28
linux: Add landlock syscall support

The-King-of-Toasters:
This PR adds support Linux syscall bindings for the latest security module: Landlock. It's a simpler filesystem capability api compared to the existing Seccomp interface. This demo program shows how it works (requires a system running a 5.13 kernel):

```zig
const std = @import("std");
const os = std.os;
const linux = os.linux;

pub const Landlock = struct {
    fd: os.fd_t,

    pub fn init(max_rules: u64) !Landlock {
        const ruleset: linux.landlock_ruleset_attr = .{
            .handled_access_fs = max_rules,
        };
        const rc = linux.landlock_create_ruleset(&ruleset, 0);
        return switch (os.errno(rc)) {
            0 => Landlock{ .fd = @intCast(os.fd_t, rc) },
            else => |err| os.unexpectedErrno(err),
        };
    }

    pub fn append(self: *Landlock, fd: os.fd_t, rules: u64) !void {
        const rc = linux.landlock_add_rule(
            self.fd,
            .{ .PATH_BENEATH = .{
                .allowed_access = rules,
                .parent_fd = fd,
            } },
        );

        return switch (os.errno(rc)) {
            0 => {},
            else => |err| os.unexpectedErrno(err),
        };
    }

    pub fn lock(self: *Landlock) !void {
        _ = try os.prctl(.SET_NO_NEW_PRIVS, .{ 1, 0, 0, 0 });

        const rc = linux.landlock_restrict_self(self.fd);
        return switch (os.errno(rc)) {
            0 => {},
            else => |err| os.unexpectedErrno(err),
        };
    }
};

const FifoType = std.fifo.LinearFifo(u8, std.fifo.LinearFifoBufferType{ .Static = 4096 });

pub fn main() !void {
    var ll = try Landlock.init(os.LANDLOCK_ACCESS_FS_READ_FILE |
        os.LANDLOCK_ACCESS_FS_WRITE_FILE |
        os.LANDLOCK_ACCESS_FS_READ_DIR);
    var fifo = FifoType.init();

    const etc = try std.fs.openDirAbsolute("/etc", .{});
    const self = try std.fs.openSelfExe(.{});

    try ll.append(os.STDOUT_FILENO, os.LANDLOCK_ACCESS_FS_WRITE_FILE);
    try ll.append(os.STDERR_FILENO, os.LANDLOCK_ACCESS_FS_WRITE_FILE);
[...]

PR #9965, opened 2021-10-17
Linux: Update syscall numbers for 5.14

The-King-of-Toasters:
This is a reduced-scope version of #9473, as this only updates the syscall numbers. I intend to use landlock in an external package.

For posterity's sake, if anyone needs to update the enum later on, grab the latest Linux tarball and look at the files ending in `.tbl`.


andrewrk:
Thanks!

PR #9966, opened 2021-10-17
stage2: implement `@hasField`

nektro:
struct and union are temporarily kept in the stage1 file since struct/unionFieldCount is returning zero for those containers and so `@hasField` is always returning false. I will keep digging to add them too.


andrewrk:
> the missing bit is that field parsing is not yet added to `zirStructDecl` or `zirUnionDecl`. is that in scope for this pr or should it be a different one?

not sure what you're asking - I pushed  a commit doing what I suggested and the test fully passed

nektro:
oh i see now, i misread the function name and thought you were describing what you did in https://github.com/ziglang/zig/commit/127cd06ba1c3406a41aecae01ca32162690ab4e1 would fix the issue. sorry about that 😓

Issue #8070, opened 2021-02-25
Dynamic build.zig dependencies


marler8997:
Special Thanks to @MasterQ32 , @g-w1, @mikdusan and @pixelherodev for their help in finding this solution.

This is a proposed interface to allow `build.zig` to declare package requirements based on any options/data available when its `build` function is exectued.

The problem this interface addresses is a "chicken and egg problem" with `build.zig`.  If `build.zig` needs a package, then it needs to be available before compiling `build.zig`, otherwise `@import("somepkg")` will cause a compile error.  However, we can't know which packages `build.zig` needs until after we compile it and provide it all the build options.  For example, a `build.zig` file may have an option to enable building an Android APK, in which case it requires the "androidbuild" package that provides functions to generate build constructs for Android.  But this package is only needed if we specify `-Dandroid` on the command-line which isn't available until after `build.zig` has been compiled and is running.

The problem can take some time to wrap your head around.  A compiled and running `build.zig` knows what packages it needs, but we can't compile and run `build.zig` without first knowing what package it needs.  So the idea here is to break this cyclic dependency by first compiling `build.zig` once without any packages, then we can query the resulting executable with our build options to see what packages it needs, then we can resolve those packages and recompile `build.zig` (if it needs any packages).  The question then becomes, how do we compile `build.zig` without the packages it needs?

The sticking point to compiling `build.zig` without its needed packages are its `@import("somepkg")` statements.  These statements are analyzed at comptime and cannot fail.  There is another proposal that suggests `@tryImport` be added to the language to address this (https://github.com/ziglang/zig/issues/8025), however, this proposal suggests an alternative that doesn't require a language addition.  Instead, before compiling `build.zig` we generate a module named `buildpkgs` that provides the information about what packages are available at comptime:

`buildpkgs.zig`
```zig
pub const names = [];

pub fn has(comptime name: []const u8) bool {
     inline for (names) |has_name| {
         if (@import("std").mem.eql(u8, name, has_name)) return true;
     }
}
```

Going back to the android example, this is what our `build.zig` file might look like:

```zig
// buildpkgs is our dynamically generated module
const buildpkgs = @import("buildpkgs");

pub fn build(b: *Builder) !void {
    if (b.option(bool, "android", "build an android APK")) {
        if (comptime buildpkgs.has("androidbuild")) {
            const androidbuild = @import("androidbuild");
            androidbuild.createManifest(.{ ... });
            androidbuild.setTarget(...);
            ...
        } else {
[...]

Issue #3477, opened 2019-10-18
Support build.zig failures without stacktrace?


marler8997:
I'm using `build.zig` build a project that depends on another project.  To keep it simple, I just check if the other project exists on the local filesystem.  If it doesn't I print an error message like this:
```
Error: library index file '../mylib/mylib.zig' does not exist
       have you downloaded the mylib library? Run the following to clone it:
       git clone https://github.com/USER/mylib ../mylib
```

However, after I print this error message in `build.zig` I return an error, but then `build.zig` prints a big stack trace and the error message gets lost.

Two solutions for this come to mind.

## Solution Idea 1

Support a special error code to indicate that the error has already been reported.

```zig
pub fn build(b: *Builder) !void {
    if (somethingFailed) {
        std.debug.warn("this failed, fix it with this...\n");
        return error.AlreadyReported;
    }
}
```

## Solution Idea 2

Support `!u8` as a return type for the `build` function.  This is similar to what we did for the `main` function (https://github.com/ziglang/zig/issues/2662).  For a `!ui8` return type, returning 0 means success, non-zero means an error but there's no need for a stacktrace, and an error produces a stacktrace like normal.

andrewrk:
Solved with `--prominent-compile-errors`. Build runner sends exit(2) and parent also calls exit(2) without printing the stack trace.

marler8997:
> Solved with --prominent-compile-errors. Build runner sends exit(2) and parent also calls exit(2) without printing the stack trace.

Wow this one is almost 5 years old, how time flies!  Thanks for a nice solution :)

PR #9987, opened 2021-10-20
Builder: add build_args and addBuild

marler8997:
Status: I converted this back to a draft to wait for #9989. If that gets integrated then I'll update this PR to leverage it.  After playing with this I realize that it's a bit unruly without #9989 to reduce build error noise.

This change adds the "build_args" field to the zig Builder. This provide access to the command-line arguments that were passed to the current "zig build" instance.  These arguments are forwarded to the build_runner after a fixed set of arguments passed from zig itelf.  Builds no longer need to know this internal detail to access these arguments.

Along with this, I've also added a convenient "addBuild" function to Builder that will create a RunStep to execute "zig build".

I've also preserved the original `cache_root` command-line parameter in the new `cache_root_cmdline` field.

Making these new pieces available supplies a more robust interface that enables build.zig files to bootstrap themselves.  This provides a pathway to address https://github.com/ziglang/zig/issues/8070

Here is a real-world build file that can already make use of this: https://github.com/marler8997/maros/blob/master/build.zig

Another use case I already have for this is my implemenation of `zig toolchain`: https://github.com/marler8997/zig-toolchain/blob/dc2dbe137654f4d2f691c5b716975410c8ff4694/zig-toolchain.zig#L147

That example would be slightly modified to leverage `zig build` instead of `zig build-exe`.

If accepted, I may also use this in my ziget/zigup projects.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -60,7 +60,9 @@ pub const Builder = struct {
     libc_file: ?[]const u8 = null,
     installed_files: ArrayList(InstalledFile),
     build_root: []const u8,
+    cache_root_cmdline: []const u8,
     cache_root: []const u8,
+    build_args: []const [:0]const u8,
     global_cache_root: []const u8,
     release_mode: ?std.builtin.Mode,
     is_release: bool,
@@ -138,6 +140,7 @@ pub const Builder = struct {
         build_root: []const u8,
         cache_root: []const u8,
         global_cache_root: []const u8,
+        build_args: []const [:0]const u8,
     ) !*Builder {
         const env_map = try allocator.create(BufMap);
         env_map.* = try process.getEnvMap(allocator);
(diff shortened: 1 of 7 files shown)


Vexu:
#9989 was reverted, pending on #3477.

marler8997:
I think the more pressing issue this PR addresses is #8070.  I'm currently leveraging the mechanism this PR provides in my zigup project here: https://github.com/marler8997/zigup/blob/master/build.zig#L37

This allows me to wrap my main build code `build2.zig` in the user-facing `build.zig` file, where it can resolve any needed dependencies before re-executing `build2.zig`.  The main problem is that I'm using an internal detail about the build_runner that could change and break my code.  This PR makes the build arguments apart of the public interface so I don't have to rely on any internal details.

The other problem was that having multiple zig builds call each other can cause alot of output, as each instance can end up printing it's own stack trace, but Andrew has said he doesn't see this as an issue and thinks filtering output should be handled by external tools.

So given this, this PR should actually be good to integrate, although, it would be nice to get some feedback from @andrewrk on it.

PR #10002, opened 2021-10-22
ci: azure: windows probe systeminfo

mikdusan:
Run `systeminfo` as first step to see important host details which may
effect the pipeline. Of particular interest, this command will display
OS version, CPU model, physical RAM, and VRAM (swap is VRAM - RAM).


mikdusan:
closing this without merge in anticipation of a self-hosted windows runner

PR #10003, opened 2021-10-22
std.Thread.getName/setName: rework windows implementation

viriuwu:
So, the current implementation of `Thread.getName` and `Thread.setName` on Win32 uses the `GetThreadDescription` and `SetThreadDescription` APIs respectively. The problem with this is **sort of** documented in the source:

https://github.com/ziglang/zig/blob/a3c9bfef301e0a4675fcea57356653334e07df45/lib/std/Thread.zig#L88-L93

However, it hard links to `kernel32.dll` and the function was not in there until **after** W10 1607, it was previously acquired by querying the symbol from `kernelbase.dll` at runtime. Putting loader code for a shared object in the standard library is pretty ugly though, and doesn't solve properly linking to it in cases where it's available. Thankfully as with a lot of things in Win32 you can ditch this terrible null-terminator-using API entirely, and just use `Nt*` functions to get the job done correctly.

This PR changes the functions to use the `Nt(Get/Set)InformationThread` APIs to query and set the name, and since the higher level APIs are just a wrapper for these, it's fully compatible with them, sans extra allocation for query, global error code set, and the need to check for compatibility at all because on older systems it simply just returns a status saying it's not implemented. Great!

Note: These don't have anything to do with the new implementation, these were present in the previous one too, but if you do `setName("")` then `getName()` returns `null` because NT makes no distinction between an unset thread name and an empty thread name. The other problem is if that someone else uses `Nt*` functions to make a thread name bigger than the current limit in std, it'll yield `error.Unexpected`, or potentially cause a very ugly assertion failure in `utf16leToUtf8` even (unsure), the limit on the thread name is 64KiB (`c_ushort`), not 32B like it currently says. Maybe something related to these issues should be put in this PR too, but it could be out of scope, especially since the latter only happens if raw Win32 (or FFI) is mixed with std.

### DIFF
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -2029,21 +2029,6 @@ pub fn unexpectedStatus(status: NTSTATUS) std.os.UnexpectedError {
     return error.Unexpected;
 }
 
-pub fn SetThreadDescription(hThread: HANDLE, lpThreadDescription: LPCWSTR) !void {
-    if (kernel32.SetThreadDescription(hThread, lpThreadDescription) == 0) {
-        switch (kernel32.GetLastError()) {
-            else => |err| return unexpectedError(err),
-        }
-    }
-}
-pub fn GetThreadDescription(hThread: HANDLE, ppszThreadDescription: *LPWSTR) !void {
-    if (kernel32.GetThreadDescription(hThread, ppszThreadDescription) == 0) {
-        switch (kernel32.GetLastError()) {
-            else => |err| return unexpectedError(err),
-        }
-    }
-}
-
 pub const Win32Error = @import("windows/win32error.zig").Win32Error;
 pub const NTSTATUS = @import("windows/ntstatus.zig").NTSTATUS;
 pub const LANG = @import("windows/lang.zig");
--- a/lib/std/os/windows/ntdll.zig
+++ b/lib/std/os/windows/ntdll.zig
@@ -23,23 +23,23 @@ const FILE_BASIC_INFORMATION = windows.FILE_BASIC_INFORMATION;
 const SIZE_T = windows.SIZE_T;
 const CURDIR = windows.CURDIR;
 
-pub extern "NtDll" fn RtlGetVersion(
+pub extern "ntdll" fn RtlGetVersion(
     lpVersionInformation: *RTL_OSVERSIONINFOW,
 ) callconv(WINAPI) NTSTATUS;
-pub extern "NtDll" fn RtlCaptureStackBackTrace(
+pub extern "ntdll" fn RtlCaptureStackBackTrace(
     FramesToSkip: DWORD,
     FramesToCapture: DWORD,
     BackTrace: **anyopaque,
     BackTraceHash: ?*DWORD,
 ) callconv(WINAPI) WORD;
-pub extern "NtDll" fn NtQueryInformationFile(
+pub extern "ntdll" fn NtQueryInformationFile(
     FileHandle: HANDLE,
     IoStatusBlock: *IO_STATUS_BLOCK,
     FileInformation: *anyopaque,
     Length: ULONG,
     FileInformationClass: FILE_INFORMATION_CLASS,
 ) callconv(WINAPI) NTSTATUS;
-pub extern "NtDll" fn NtSetInformationFile(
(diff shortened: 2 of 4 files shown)


lib/std/os/windows.zig
```diff
@@ -2015,21 +2015,6 @@ pub fn unexpectedStatus(status: NTSTATUS) std.os.UnexpectedError {
...
-    }
-}
-pub fn GetThreadDescription(hThread: HANDLE, ppszThreadDescription: *LPWSTR) !void {
-    if (kernel32.GetThreadDescription(hThread, ppszThreadDescription) == 0) {
-        switch (kernel32.GetLastError()) {
-            else => |err| return unexpectedError(err),
-        }
-    }
-}
-
```
kprotty, line 2032: Could these still be kept in the stdlib in case anyone wants to call them? If they're not used, then the kernel32 stubs shouldn't be in the compiled binary given Zig's comptime evaluation on reachable code.

lib/std/os/windows/ntdll.zig
```diff
@@ -23,36 +23,95 @@ const FILE_BASIC_INFORMATION = windows.FILE_BASIC_INFORMATION;
 const SIZE_T = windows.SIZE_T;
 const CURDIR = windows.CURDIR;
 
-pub extern "NtDll" fn RtlGetVersion(
+pub extern "ntdll" fn RtlGetVersion(
```
kprotty, line 26: I've seen issues where not using `NtDll` in that exact casing causes LLVM/LLD to go into an infinite loop when linking a program. This might still be the case

lib/std/os/windows.zig
viriuwu: The agreed on policy (see #4426) is to have minimal surface area so std works, and the rest should go in libraries. There's a lot of unused code and (refer to andrew's comment) eventually the parts that are not used in std should be removed, such as `*A` functions.

lib/std/os/windows/ntdll.zig
viriuwu: Let me know if you have a source for this, it works on my machine (the best source, obviously) and has always worked with rustc too which also uses LLVM. Maybe it's something on older versions? We updated to LLVM 13 recently.

lib/std/os/windows.zig
kprotty: Ah, good point

lib/std/os/windows/ntdll.zig
kprotty: Ironically, the source for it was on my machine. But it was also a few months ago. Let me try to generate a quick demo and will report back if it doesn't happen anymore as it could be fixed by LLVM update like you've said.

kprotty: I remember hitting it with `NtWaitForAlertByThreadId` but it looks like its fixed now.

viriuwu: Great to hear! Can't imagine how that happened in the first place, as the file *is* named `ntdll.lib`, all lowercase...

viriuwu:
Looks like other changes broke this PR, so I've reapplied the changes on top of master.

viriuwu:
Looks like other changes broke this PR, so I've reapplied the changes on top of master, and did `c_void` -> `anyopaque`.

Vexu (approved):

viriuwu:
Quickly checked to see what the Windows CI is about (I assumed it was the usual at the time) and noticed that the rebased version did not actually work anymore, it was due to the control flow being slightly restructured inbetween those commits. The fix was just:
```diff
diff --git a/lib/std/Thread.zig b/lib/std/Thread.zig
index 6ddc6dd75..05f6fd66d 100644
--- a/lib/std/Thread.zig
+++ b/lib/std/Thread.zig
@@ -104,7 +104,7 @@ pub fn setName(self: Thread, name: []const u8) SetNameError!void {
                 &unicode_string,
                 @sizeOf(os.windows.UNICODE_STRING),
             )) {
-                .SUCCESS => {},
+                .SUCCESS => return,
                 .NOT_IMPLEMENTED => return error.Unsupported,
                 else => |err| return os.windows.unexpectedStatus(err),
             }
```

PR #10005, opened 2021-10-22
build.RunStep: detect bad exe even if we don't run

marler8997:
This change attempts to find errors earlier.  This change modifies RunStep to detect if it's requested to run an incompatible exe.  This means we would detect an invalid build configuration for a run step even if that run step isn't executed.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -1735,7 +1735,7 @@ pub const LibExeObjStep = struct {
         // Consider that this is declarative; the run step may not be run unless a user
         // option is supplied.
         const run_step = RunStep.create(exe.builder, exe.builder.fmt("run {s}", .{exe.step.name}));
-        run_step.addArtifactArg(exe);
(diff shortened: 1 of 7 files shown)


andrewrk (changes requested):
This should be rebased now that #10265 is merged and follow the lead set by the Zig CLI: always try to execute the program, but if it fails then emit a helpful warning explaining why it might have failed.

This way if someone has some kind of binfmt_misc installed, it will work fine. It also needs to integrate with the `getExternalExecutor` API.

This PR and #10143 overlap.

marler8997:
@andrewrk I'm going to close this one since I now have a better understanding of the changes that occured (https://ziglang.org/download/0.9.0/release-notes.html#toc-Stop-Guessing-Whether-the-System-Can-Execute-Binaries).  This change looks fundamentally at odds with my change here, so I'm going to close this since I believe the correct change would be completely different.

PR #10006, opened 2021-10-23
Default target-abi based on RISC-V extensions and user selectable -mabi/target-abi

### LINKED ISSUE #9760: riscv64-freestanding: C and Zig object files default to soft-float ABI
### Issue:
`zig build-obj -target riscv64-freestanding` seems to default to `soft-float ABI` for C and Zig files.
Assembly files seem to target `double-float` as expected.

### Expected behavior:
It should probably default to `double-float`, like what every other platform seems to, at least with `-mcpu baseline_rv64+d` or more.

### How to reproduce:
Sample `main.zig` file:
```zig
const std = @import("std");

pub fn main() anyerror!void {
}
```
The ABI of the file is different when targeting `freestanding` or `linux`:
```
$ zig build-obj -target riscv64-freestanding main.zig && file main.o
main.o: ELF 64-bit LSB relocatable, UCB RISC-V, RVC, soft-float ABI, version 1 (SYSV), with debug_info, not stripped

$ zig build-obj -target riscv64-linux main.zig && file main.o
main.o: ELF 64-bit LSB relocatable, UCB RISC-V, RVC, double-float ABI, version 1 (SYSV), with debug_info, not stripped
```

### Notes:
I came across this issue while reading https://matrix89.github.io/writes/writes/experiments-in-riscv/ and got the urge to make a `build.zig` file to compile the project as I was too lazy to get the GCC cross compiler.

Since the assembly file targets `double-float` and the C file targets `soft-float`, I got the following error:
```
ld.lld: error: /home/plumeus/zig_projects/riscv_test/zig-cache/o/cc21692bae9a5e68729e9557aedcbb22/start.o: cannot link object files with different floating-point ABI
error: LLDReportedFailure
```
Examining the cause led me to open this issue.

### Possibly related issues:
#4863
#5558

akovaski:
Fixes #9760 
Zig now calculates a default `-mabi`/target-abi for the target RISC-V extensions. Zig will then pass the default target-abi to clang/llvm for Assembly, C (and Cpp, m, h), and Zig files.

This change also allows the user to manually select the target-abi for compilation. Following Clang and LLVM's example, the target-abi is separate from the triplet abi. Target-abi can be passed from the command line, like:
```bash
zig build-obj -target riscv64-freestanding -mabi lp64f main.zig
```

Or provided in `build.zig` when running `zig build` via the target_abi member of LibExeObjStep, like:
```zig
exe.target_abi = .lp64f;
```


akovaski:
* Moved target_abi from CrossTarget to LibExeObjStep for `zig build`.
* Simplified the TargetAbi type a little bit and added some validation.
* rebased and squashed

lib/std/target.zig
```diff
@@ -550,6 +550,103 @@ pub const Target = struct {
...
+                } else {
+                    return .ilp32;
+                }
+            }
+            fn validate(target_abi: TargetAbi, features: Cpu.Feature.Set) ParseError!void {
+                const has_e = riscv.featureSetHas(features, .e);
+                const has_f = riscv.featureSetHas(features, .f);
+                const has_d = riscv.featureSetHas(features, .d);
+                return switch (target_abi) {
+                    .ilp32e => if (has_e) void{} else error.FeatureAbiMismatch,
```
LemonBoy, line 579: Huh, is `void{}` even valid Zig code? `void` is a type and `{}` an empty block, it's not really meaningful.

lib/std/target.zig
```diff
@@ -550,6 +550,103 @@ pub const Target = struct {
...
+        };
+        pub fn default(arch: Cpu.Arch, features: Cpu.Feature.Set) ?TargetAbi {
+            return switch (arch) {
+                .riscv32 => Riscv32ABI.default(features),
+                .riscv64 => Riscv64ABI.default(features),
+                else => null,
+            };
+        }
+
+        // re-implementing @tagName because @tagName is not null-terminated in the stage0 compiler
```
LemonBoy, line 614: That's not true? `@tagName` returns a zero-terminated slice and that's been the case for quite some time.

src/Compilation.zig
```diff
@@ -1152,6 +1153,7 @@ pub fn create(gpa: *Allocator, options: InitOptions) !*Compilation {
         cache.hash.add(options.target.os.getVersionRange());
         cache.hash.add(options.is_native_os);
         cache.hash.add(options.target.abi);
+        cache.hash.addBytes(if (options.target_abi) |t| t.string() else ".");
```
LemonBoy, line 1156:
```suggestion
        if (options.target_abi) |t| cache.hash.addBytes(t.string());
```

lib/std/target.zig
akovaski: It compiles and works. I assume it works like regular type instantiation. Though it's not used often, from grep I see it's only being used in the tests in test/ .

In this case it's just being used in place of a missing error. I guess I could probably use an ?error as a return type instead. Or maybe there's a better way?

lib/std/target.zig
akovaski: In the final output `zig` compiler from running `make`, when that zig compiler compiles `@tagName` it is null terminated. However, when I tried using `@tagName` I found that the strings were not null terminated when I compiled `zig`. So I assumed it was a difference between the intermediate compiler and the final compiler.

I'll try switching it back and trying again though. I'll post an update.

src/Compilation.zig
akovaski: Yeah "." doesn't make sense, I copied that from a cache path.
I see that src/Cache.zig's addOptionalBytes and addOptional still add to the cache if the value is null.
I don't know whether or not that is really necessary for the hash, but maybe it would be better to do this?:
```
cache.hash.addOptionalBytes(if (options.target_abi) |t| t.string() else null);
```

lib/std/target.zig
LemonBoy: Colour me surprised, that seems an unwanted construct that's accepted by the compiler, if you try replacing `void` with some other type you'll see it doesn't really work.

You can simply replace `void{}` with `{}`.

lib/std/target.zig
akovaski: I tried again, I still have the same issue with `@tagName`.

When I try to compile, eg with `zig build-obj -target riscv32-freestanding main.zig -mabi ilp32d` I get the following warning from LLVM:
```
'ilp32dilp32eilp32flp64dlp64f' is not a recognized ABI for this target (ignoring target-abi)
```

src/Compilation.zig
LemonBoy: I don't think it really matters either way beside making the new cache entries incompatible with the old ones.

lib/std/target.zig
akovaski: Ah, I had been thinking like struct initialization.
Like
```zig
const Nothing = struct {};
var a = Nothing{};
```

It's interesting to find out that `{}` has the type of `void`. I'll change these `void{}` to `{}`.

lib/std/target.zig
akovaski: Adding some detail from my memory of a few days ago, I believe I had tried putting in an
```zig
assert(tag_name_result[tag_name_result.len] == 0);
```
into TargetAbi.string, and that was OK on my personal zig code, but raised an exception when the compiler used it.

LemonBoy: Try with this patch:

```patch
diff --git a/lib/std/build.zig b/lib/std/build.zig
index 621499f8c..01d856bd2 100644
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -2434,7 +2434,7 @@ pub const LibExeObjStep = struct {
         }
 
         if (self.target_abi) |target_abi| {
-            try zig_args.append(builder.fmt("-mabi={s}", .{target_abi.string()}));
+            try zig_args.append(builder.fmt("-mabi={s}", .{@tagName(target_abi)}));
         }
         if (self.code_model != .default) {
             try zig_args.append("-mcmodel");
diff --git a/lib/std/target.zig b/lib/std/target.zig
index 61a9a42e4..9a1557d35 100644
--- a/lib/std/target.zig
+++ b/lib/std/target.zig
@@ -611,25 +611,6 @@ pub const Target = struct {
             };
         }
 
-        // re-implementing @tagName because @tagName is not null-terminated in the stage0 compiler
-        fn tagName(value: anytype) [:0]const u8 {
-            const T = @TypeOf(value);
-            const info = @typeInfo(T);
-
-            switch (info) {
-                .Enum => |enum_info| {
-                    comptime var names: [enum_info.fields.len][:0]const u8 = undefined;
-                    inline for (enum_info.fields) |field, i| {
-                        names[i] = field.name ++ [0:0]u8{};
-                    }
-                    return names[@enumToInt(value)];
-                },
-                else => @compileError("tagName does not support type " ++ @typeName(T) ++ "."),
-            }
-        }
-        pub fn string(self: TargetAbi) [:0]const u8 {
-            return tagName(self);
-        }
         pub const ParseError = error{
             InvalidAbi,
             ArchAbiMismatch,
diff --git a/src/Compilation.zig b/src/Compilation.zig
index 81d99ae78..ef38b5585 100644
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1153,7 +1153,7 @@ pub fn create(gpa: *Allocator, options: InitOptions) !*Compilation {
         cache.hash.add(options.target.os.getVersionRange());
         cache.hash.add(options.is_native_os);
         cache.hash.add(options.target.abi);
-        cache.hash.addBytes(if (options.target_abi) |t| t.string() else ".");
+        cache.hash.addBytes(if (options.target_abi) |t| @tagName(t) else ".");
         cache.hash.add(ofmt);
         cache.hash.add(pic);
         cache.hash.add(pie);
@@ -3303,7 +3303,7 @@ pub fn addCCArgs(
             }
 
             if (comp.bin_file.options.target_abi) |target_abi| {
-                try argv.append(try std.fmt.allocPrint(arena, "-mabi={s}", .{target_abi.string()}));
+                try argv.append(try std.fmt.allocPrint(arena, "-mabi={s}", .{@tagName(target_abi)}));
             }
 
             if (comp.haveFramePointer()) {
@@ -3446,7 +3446,7 @@ pub fn addCCArgs(
             }
 
             if (comp.bin_file.options.target_abi) |target_abi| {
-                try argv.append(try std.fmt.allocPrint(arena, "-mabi={s}", .{target_abi.string()}));
+                try argv.append(try std.fmt.allocPrint(arena, "-mabi={s}", .{@tagName(target_abi)}));
             }
 
             if (target_util.clangAssemblerSupportsMcpuArg(target)) {
@@ -4370,7 +4370,7 @@ fn updateStage1Module(comp: *Compilation, main_progress_node: *std.Progress.Node
         .is_native_cpu = false, // Only true when bootstrapping the compiler.
         .llvm_cpu_name = if (target.cpu.model.llvm_name) |s| s.ptr else null,
         .llvm_cpu_features = comp.bin_file.options.llvm_cpu_features.?,
-        .llvm_target_abi = if (comp.bin_file.options.target_abi) |t| t.string().ptr else null,
+        .llvm_target_abi = if (comp.bin_file.options.target_abi) |t| @tagName(t) else null,
     };
 
     comp.stage1_cache_manifest = &man;
diff --git a/src/codegen/llvm.zig b/src/codegen/llvm.zig
index f26e474a7..bb55f2e1e 100644
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -244,7 +244,7 @@ pub const Object = struct {
         // TODO handle float ABI better- it should depend on the ABI portion of std.Target
         const float_abi: llvm.ABIType = .Default;
 
-        const abi_name: ?[*:0]const u8 = if (options.target_abi) |t| t.string().ptr else null;
+        const abi_name: ?[*:0]const u8 = if (options.target_abi) |t| @tagName(t) else null;
 
         const target_machine = llvm.TargetMachine.create(
             target,
diff --git a/src/stage1/codegen.cpp b/src/stage1/codegen.cpp
index 54df7afbb..a3f8ce7e9 100644
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -5399,8 +5399,9 @@ static LLVMValueRef get_enum_tag_name_function(CodeGen *g, ZigType *enum_type) {
     if (enum_type->data.enumeration.name_function)
         return enum_type->data.enumeration.name_function;
 
-    ZigType *u8_ptr_type = get_pointer_to_type_extra(g, g->builtin_types.entry_u8, false, false,
-            PtrLenUnknown, get_abi_alignment(g, g->builtin_types.entry_u8), 0, 0, false);
+    ZigType *u8_ptr_type = get_pointer_to_type_extra2(g, g->builtin_types.entry_u8, false, false,
+            PtrLenUnknown, get_abi_alignment(g, g->builtin_types.entry_u8), 0, 0, false,
+            VECTOR_INDEX_NONE, nullptr, g->intern.for_zero_byte());
     ZigType *u8_slice_type = get_slice_type(g, u8_ptr_type);
     ZigType *tag_int_type = enum_type->data.enumeration.tag_int_type;
 
@@ -5453,7 +5454,7 @@ static LLVMValueRef get_enum_tag_name_function(CodeGen *g, ZigType *enum_type) {
             continue;
         }
 
-        LLVMValueRef str_init = LLVMConstString(buf_ptr(name), (unsigned)buf_len(name), true);
+        LLVMValueRef str_init = LLVMConstString(buf_ptr(name), (unsigned)buf_len(name), false);
         LLVMValueRef str_global = LLVMAddGlobal(g->module, LLVMTypeOf(str_init), "");
         LLVMSetInitializer(str_global, str_init);
         LLVMSetLinkage(str_global, LLVMPrivateLinkage);
```

akovaski: That seems to work. Thanks!

akovaski:
There is a similar issue when compiling with release modes that are not Debug. I describe it back in a comment in the original issue https://github.com/ziglang/zig/issues/9760#issuecomment-955712467

This is a problem with clang/LLVM and RISC-V LTO and mabi/target-abi.
I could 'fix' the issue in Zig by defaulting LTO to off for riscv architectures, but that might just be unnecessary noise in the compiler source, I dunno. The manual workaround of setting want_lto to false when building my riscv64 application works just as well.

```diff
diff --git a/src/Compilation.zig b/src/Compilation.zig
index e17493719..1089894c5 100644
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -969,22 +969,24 @@ pub fn create(gpa: *Allocator, options: InitOptions) !*Compilation {
         const lto = blk: {
             if (options.want_lto) |explicit| {
                 if (!use_lld and !options.target.isDarwin())
                     return error.LtoUnavailableWithoutLld;
                 break :blk explicit;
             } else if (!use_lld) {
                 // TODO zig ld LTO support
                 // See https://github.com/ziglang/zig/issues/8680
                 break :blk false;
             } else if (options.c_source_files.len == 0) {
                 break :blk false;
+            } else if (options.target.cpu.arch == .riscv32 or options.target.cpu.arch == .riscv64) {
+                break :blk false;
             } else switch (options.output_mode) {
                 .Lib, .Obj => break :blk false,
                 .Exe => switch (options.optimize_mode) {
                     .Debug => break :blk false,
                     .ReleaseSafe, .ReleaseFast, .ReleaseSmall => break :blk true,
                 },
             }
         };
```

andrewrk:
I made some changes to this branch in 4c1a62326b5090fad67c004a6e58a6e4265a77ad before merging.

The changes that landed are much less drastic. If we need to add additional `std.Target.Abi` enum tags to handle additional possibilities, let's discuss that.

Issue #10013, opened 2021-10-23
Build options cannot be set to enum values which are not valid identifiers


VojtechStep:
### Remember to search before filing a new report

- [X] I searched for this bug and did not find it in the issue tracker, and furthermore, the title I used above will make this new bug report turn up in the search results for my query.

### Zig Version

0.9.0-dev.1465+ee98d8700

### Steps to Reproduce

`src/shared.zig`:
```zig
pub const ZigVersion = enum { master, @"0.8.1" };
```
`src/main.zig`:
```zig
const std = @import("std");
const build_options = @import("build_options");

pub fn main() anyerror!void {
    std.log.info("Version is: {}", .{build_options.data_version});
}
```
`build.zig`:
```zig
const std = @import("std");
const shared = @import("./src/shared.zig");

pub fn build(b: *std.build.Builder) void {
    const exe = b.addExecutable("float-option", "src/main.zig");

    const exe_options = b.addOptions();
    exe_options.addOption(shared.ZigVersion, "data_version", b.option(shared.ZigVersion, "data_version", "Zig version") orelse .@"0.8.1");
    exe.addOptions("build_options", exe_options);
    exe.install();
}
```

Run `zig build`

### Expected Behavior

The program builds, and when run outputs the string `Version is: ZigVersion.@"0.8.1"`

### Actual Behavior

The build fails, reporting the error:
```
/home/adalbert/Code/VojtechStep/zig-tests/float-option/zig-cache/options/cDQEQhqaU1X7FiSoEzEjjtqqYXbZ3CtKOXhGqqFpDf6uq0WFl5AeS-EIKbtGlurx:6:49: error: expected pointer dereference, optional unwrap, or field access, found 'float_literal'
pub const data_version: ZigVersion = ZigVersion.0.8.1;
                                                ^
```

It seems that when the zig file representing the build options is generated, it doesn't wrap the enum variant with `std.zig.fmtId`, resulting in invalid zig code being produced.

PR #10016, opened 2021-10-23
Various build options (OptionsStep) fixes

courajs:
- There was code to format a `std.builtin.Version`, but it was untested, and actually fell through to the default case, declaring the constant a second time, as an invalid expression.
- Because we're doing custom code formatting here, I added a check that the generated code is still valid zig syntax
- Fix specifying enum values with names that are not valid identifiers (fixes https://github.com/ziglang/zig/issues/10013)
- Allow arbitrary slices and arrays of primitives
- Add explicit error messaging when an invalid type/value is passed, rather than just generating some invalid code


andrewrk:
Thanks!

PR #10020, opened 2021-10-24
fix(uefi MemoryDescriptor): padding after memory type

stakach:
not 100% certain, just noticed that this is implemented slightly differently in rust
https://docs.rs/uefi/0.11.0/src/uefi/table/boot.rs.html#715


liampwll:
I'm not sure this field should be here at all as the alignment of these members should be specified by the ABI specified in section 2.3 of the UEFI spec, specifically:
>  Unless otherwise specified all data types are naturally aligned. Structures are aligned on
boundaries equal to the largest internal datum of the structure and internal data are implicitly padded to
achieve natural alignment.
Which is the same as MSVC if this isn't implemented yet.

Issue #7667, opened 2021-01-03
relocation error: symbol pthread_sigmask version GLIBC_2.2.5 not defined in file libc.so.6 with link time reference


lithdew:
Encountered this error while trying to run CI tests on Github Actions for a Zig project, using the Zig nightly bundle provided here: https://ziglang.org/builds/zig-linux-x86_64-0.7.1+dfacac916.tar.xz. Linking with musl makes the problem go away.

The latest glibc version available in Github Actions is 2.27.

```
root@dew:/github/workspace# ldd ./zig-cache/o/ea49ec9c878272fe2410dfdf1fc66939/test
        linux-vdso.so.1 (0x00007ffe5b38e000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fa3c72c5000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fa3c6f27000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fa3c6d08000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fa3c6b04000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fa3c68fc000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fa3c76b6000)
        libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007fa3c66f9000)
root@dew:/github/workspace# nm -D /lib/x86_64-linux-gnu/libpthread.so.0 | grep pthread_sig
000000000000f290 T pthread_sigmask
000000000000f3e0 T pthread_sigqueue
root@dew:/github/workspace# nm -D /lib/x86_64-linux-gnu/libc.so.6 | grep pthread_sig
root@dew:/github/workspace# 
```

Reproduction requires an 18.2GB Docker image that emulates the Github Actions CI Environment, which can be found here: https://hub.docker.com/layers/nektos/act-environments-ubuntu/18.04/images/sha256-4d991875312e2ad084ff5350aa193e8a7dbdc4c144476d48df6a69e1e3404427

References to `pthread` symbols in built executable:

```
root@dew:/github/workspace# objdump -T ./zig-cache/o/ea49ec9c878272fe2410dfdf1fc66939/test | grep pthread
[...]

Issue #8896, opened 2021-05-25
glibc dynamic stubs are deeply flawed


LemonBoy:
* glibc stubs are generated by merging together functions (marked with `F` in the abilist files) and variables (marked with `D`) and treating them as if they were functions (`STT_FUNC` entries). This is a huge problem when the linker performs [copy relocations](https://maskray.me/blog/2021-01-09-copy-relocations-canonical-plt-entries-and-protected) (eg. a static library references `__stack_chk_guard` using absolute addresses, even though that symbol is provided by an external shared object). Upon seeing a `STT_FUNC` symbol lld processes it as if it were a function, creating a PLT (!!!) entry instead of emitting a copy relocation or a GOT indirection (afaics the latter is not really used).

To understand the implications of this error here's a small snippet of ARM assembly, showing a typical function prologue where we load the stack cookie from `__stack_chk_guard` and store it on stack:
```asm
   3ee78: 10 4c 2d e9  	push	{r4, r10, r11, lr}
   3ee7c: 08 b0 8d e2  	add	r11, sp, #8
   3ee80: d0 d0 4d e2  	sub	sp, sp, #208
   3ee84: b0 c0 0e e3  	movw	r12, #57520
   3ee88: 07 c0 40 e3  	movt	r12, #7
   # Load from 0x7e0b0
   3ee8c: 00 c0 9c e5  	ldr	r12, [r12]
   3ee90: 0c c0 0b e5  	str	r12, [r11, #-12]
```
And, guess what, that address points to a PLT entry:
```asm
0007e0b0 <$a>:
   7e0b0: 00 c6 8f e2  	add	r12, pc, #0, #12
   7e0b4: 20 ca 8c e2  	add	r12, r12, #32, #20
   7e0b8: 68 f3 bc e5  	ldr	pc, [r12, #872]!
```
In short the program is using garbage for stack cookie and, what's worse, it's not even random data!

The fix is simple, the abilist files contain the symbol size, it's only a matter of passing this info to the `.txt` files used to synthesize the `.s` file.

* musl stubs are a bit better, some symbols are correctly marked as objects (`STT_OBJECT`) but are zero-sized, making `lld` promptly reject those when attempting to do a copy relocation. The fix is, again, to emit `.size <sym>, <size>` for the symbols marked as `%object`.

Note that the symbol sizes may be different across platforms (eg. 32 vs 64 bit), getting the symbol size wrong is dangerous.

cc @flyx, this is the problem you were hitting.

[...]

Issue #8714, opened 2021-05-08
Bad logic in generating glibc stubs


LemonBoy:
#8705 is currently failing with this error: `symbol lseek64 version GLIBC_2.2 not defined in file libc.so.6 with link time reference`.
Digging in glibc abilist files you can see that `lseek64` is currently defined in `libc` with version `2.1` but also in `libpthread` with version `2.2`! This is an historical wart that's [being addressed](https://patchwork.ozlabs.org/project/glibc/patch/20210319202914.2475530-4-adhemerval.zanella@linaro.org/) and helped uncover a hole in `tools/update_glibc.zig` logic:
at the moment it's handling this case by picking `libc` as library (since `libc` is the first abilist file being scanned) _but_ adds both the version found in `libc` and the one in `libpthread`, leading to this error.

PR #10036, opened 2021-10-26
abi/fns/vers versioning based on glibc version

mjonaitis1:
Fixes: #7667

This PR introduces versioning for symbol mapping files (abi.txt, fns.txt, vers.txt) which are used to generate stub shared libraries when compiling for glibc targets. 

These files are now stored under lib/libc/glibc/symbols/{GLIBC_VERSION} where {GLIBC_VERSION} is the version of glibc. Currently, I have generated these files from glibc 2.23 up to glibc 2.33.

update_glibc tool is also changed in this PR. Now it accepts a third argument - glibc version that will be updated or used to create new directory under lib/libc/glibc/symbols.


Mantelijo:
@marler8997 @andrewrk so generally, what are your thoughts for this to go through? Should I try to change these files from ASCII to binary format?

marler8997:
Status quo seems fine.  An additional change to reduce the uncompressed size using a "sparse format" feel like a "nice to have".

marler8997:
Another thing to consider is that if we design a compact binary format for these files, then it would reduce the memory needed to load them.  So we would save:

1. a little space in the released archive
2. a couple megabytes in an installed instance of zig
3. a couple megabytes of RAM when we have to load these files

One question I have is, when does Zig need to load these files?  Is it only when we need to build glibc?

andrewrk:
> generally, what are your thoughts for this to go through? Should I try to change these files from ASCII to binary format?

It seems likely this change can land unmodified, at first glance it looks great. I haven't given it a full review yet - sorry for the delay. At least I can tell you that I won't tag 0.9.0 before this lands. Deal?

> One question I have is, when does Zig need to load these files? Is it only when we need to build glibc?

Correct!

andrewrk (changes requested):
Thanks @mjonaitis1 for your patience on getting this reviewed. I finally had a deeper look into this today and I have some guidance on how to get this landed.

As it stands now, this is a nice proof-of-concept that the problem can be fixed, by modeling the data more accurately - taking into account the fact that symbols can migrate between libraries over versions.

However rather than introducing N copies of the data, one for each glibc version, let's rework the data layout in glibc.zig to support encoding this data properly, and likewise rework the encoding of fns.txt, abi.txt, vers.txt to being one dataset that encodes this information rather than being N datasets, one per glibc version.

I have some suggestions for how to go about this:
 * Start a brand new git repository which collects the .abilist files from the entire history of the glibc project (or as far back as it is practical to go). In this new repository, there would be a copy of the data per glibc version checked into the repo, namespaced into directories by version.
 * Move the tools/update_glibc.zig script to this new repository as well. Whenever glibc has an update, we can copy .abilist files into the repository under version directory, and re-run the script to compile the information into our singular dataset, which can be then copied over to the main ziglang/zig repository.
 * Let's switch to a binary encoding of the data and aim for compactness.

Here is some brainstorming on the encoding:

The lists of things that we have:

 * list of glibc versions  (44)
   - we can use u64 bitsets to describe a set of glibc versions.
 * list of glibc libraries  (7)
   - we can use u8 bitsets to describe a set of glibc libraries.
 * list of targets         (20)
   - we can use u32 bitsets to describe a set of targets. E.g. the first 20 bits of the u32 correspond to the list of targets; 12 bits unused.
 * list of glibc symbols (3970)

With that in mind, here is an idea to store the data. The list of glibc libraries is implied.

- list of glibc versions
- list of targets
- for each glibc symbol:
  - null-terminated string name
  - list of inclusions that describes where this symbol can be found. each inclusion is:
    - u32 set of targets that this inclusion applies to
      - use 1 of these bits to tell whether this is the last inclusion in the list
    - u64 set of glibc versions that this inclusion applies to
    - u8 set of glibc libraries that this inclusion applies to

With an average of 1.5 inclusions per symbol (which looks like a conservative estimate), this comes out to 75 KB for the entire dataset ( `13 * 3970 * 1.5`), plus the bytes need to store the symbol names exactly once. I think this will be a worthwhile investment, as this is a great deal smaller than 2.1 MiB, improving not only installation size on disk, but also performance when this data needs to be loaded by the compiler.

mjonaitis1:
I have created this new repo here: https://github.com/mjonaitis1/zig-glibc-abilist. Please note that `update_glibc.zig` still has old `abi/fns/vers` functionality - it can be removed later once we are sure about this new encoding. 

Now my question is what is the best approach to integrate this new format reader into zig's `/src/glibc.zig` source? One approach that I am currently trying out is adding `glibc_symbols_reader.zig` from the repo provided above into `/src/glibc_symbols_reader.zig` and loading it in `/src/glibc.zig:loadMetaData`, the structure might need some rework, but this might work.

What are your thoughts?

andrewrk:
@mjonaitis1 apologies - I only just now saw this update today. Exciting work, I will dedicate my main task slot tomorrow to this.

andrewrk:
I played with it a little bit-

Looks like the symbols file that gets generated is 102 KB. That's great! Looks like my estimate was off by 25 KB but that's peanuts. Pretty nice that `xz` is able to reduce the size further down to 21 KB.

> One approach that I am currently trying out...

The approach that you outlined here sounds perfectly reasonable to me :+1: 

Thanks for doing this!

mjonaitis1:
While I was trying to do a sample implementation in zig's `src/glibc.zig` I encountered an issue: with this new binary format we are losing some important information like what are the default versions for symbol that are available for specific target in specific library. This new binary format with inclusions collects all the information from pre-selected list of glibc versions (at the moment glibc 2.23-2.33), each inclusion is a summed up information of all targets and all versions available for a specific symbol. Now the problem is that a symbol can have different default versions on different targets. 

Example: dlopen symbol. In binary inclusion format, this is the information I have:
```
Symbol: dlopen in libdl
         Available versions: GLIBC_2.0 GLIBC_2.1 GLIBC_2.2 GLIBC_2.2.5 GLIBC_2.3 GLIBC_2.4 GLIBC_2.16 GLIBC_2.17
         Available targets: aarch64-linux-gnu aarch64_be-linux-gnu s390x-linux-gnu arm-linux-gnueabi armeb-linux-gnueabi arm-linux-gnueabihf armeb-linux-gnueabihf sparc-linux-gnu sparcel-linux-gnu sparcv9-linux-gnu mips64el-linux-gnuabi64 mips64-linux-gnuabi64 mips64el-linux-gnuabin32 mips64-linux-gnuabin32 mipsel-linux-gnueabihf mips-linux-gnueabihf mipsel-linux-gnueabi mips-linux-gnueabi x86_64-linux-gnu x86_64-linux-gnux32 i386-linux-gnu powerpc64le-linux-gnu powerpc64-linux-gnu powerpc-linux-gnueabi powerpc-linux-gnueabihf
```
In glibc-2.33 abilist file for libdl `sysdeps/unix/sysv/linux/x86_64/64/libdl.abilist` we have this:
```
GLIBC_2.2.5 dlopen F
```

But in `sysdeps/unix/sysv/linux/aarch64/libdl.abilist` for the same glibc-2.33 we have this:
```
GLIBC_2.17 dlopen F
```
If I target, let's say, glibc-2.17 and x86_64-linux-gnu - zig will build stub libdl with dlopen@GLIBC_2.17 which is actually not available in libdl on x86_64-linux-gnu. But since there is no per symbol per target information - there is no way to tell that. 

I tried to brainstorm a solution ([a8842aa](https://github.com/mjonaitis1/zig-glibc-abilist/commit/a8842aa415299c46d61273d65504d360feddcfce)) for this and I thought about having a "maximum" version for each library for each target. That way I could check which maximum version is available for a specific library for specific target, but this actually does not help as there might be a lot of overlapping versions in library but default versions for symbol might be different. 

So I think the only viable solution would be to actually have as many inclusions for each symbol as there are targets, where each inclusion is a bitset of available versions and library information. But this would drastically increase the symbols file size: `25(targets) x 9(u64bitset for versions  + u8 for lib) * 4100(roughly the number of symbols+duplications where symbol changes library)` which is `~900KB`. However, I am not exactly sure if this is the best approach.

LemonBoy:
Keep #8896 and #8714 in mind too, there's some more cases that need to be addressed.

andrewrk:
Closing in favor of  #10330

Issue #4026, opened 2020-01-02
implement all the math functions for all the floating point types


andrewrk:
Boxes without check marks do not have test coverage.

|             | f16 | f32 | f64 | f128 | c_longdouble |
|-------------|-----|-----|-----|------|--------------|
| `@sqrt`     | ✓   | ✓   | ✓   |      |              |
| `@sin`      | ✓   | ✓   | ✓   |      |              |
| `@cos`      | ✓   | ✓   | ✓   |      |              |
| `@exp`      | ✓   | ✓   | ✓   |      |              |
| `@exp2`     | ✓   | ✓   | ✓   |      |              |
| `@ln`       | ✓   | ✓   | ✓   |      |              |
| `@log2`     | ✓   | ✓   | ✓   |      |              |
| `@log10`    | ✓   | ✓   | ✓   |      |              |
| `@fabs`     | ✓   | ✓   | ✓   |      |              |
| `@floor`    | ✓   | ✓   | ✓   |      |              |
| `@ceil`     | ✓   | ✓   | ✓   |      |              |
| `@trunc`    | ✓   | ✓   | ✓   |      |              |
| `@nearbyInt`|     |     |     |      |              |
| `@round`    |     |     |     |      |              |

LewisGaul:
I've just come across a need for `exp` and `ln` with `f128`, so I'm going to take a look at this :)

LewisGaul:
I haven't got anywhere with implementations yet, but I've learnt a lot about how floats work and implementations of the `exp2()` function. I'm making notes and will track progress at https://github.com/LewisGaul/zig-f128math (taking inspiration from the initial work by @tiehuis in https://github.com/tiehuis/zig-fmath).

Mainly leaving this comment in case useful to anyone else interested in working on this.

I do also have a question about `c_longdouble` (since it's included in the table above). Quoting from https://github.com/tiehuis/zig-fmath/issues/7:
> The `c_longdouble` type is not implemented at all. The rationale is if required, the user should use the `libc` math library, which makes some sense given it is a c abi type anyway. The long double functions as well generally are more complex to implement as well.

[...]

PR #10046, opened 2021-10-27
Add f128 support and tests in std.math.exp2

LewisGaul:
I started working on this when I came across https://github.com/ziglang/zig/issues/4026, but I've just noticed that's talking about the *builtin* functions (`@exp2()` rather than `std.math.exp2()`)... I'm not sure whether the builtin functions will ever fall back to Zig stdlib (perhaps with self-hosted?), but hopefully this is a desirable addition either way? Let me know as I'd be interested in doing similar for other stdlib maths functions!

EDIT: Clarified with Andrew on IRC that Zig implementations are needed for the compiler in the long term, even if only as a fallback when not supported by the backend. I'm still not sure how to propagate this support for `f128` to the builtin `@exp2()` (for stage1 or stage2) though - I can see the place in `ir.cpp` where the compile error is raised, but adding a call to `exp2q()` there gives:
```
/mnt/c/Users/legaul/Documents/personal/zig/src/stage1/ir.cpp:24094:36: error: 'exp2q' was not declared in this scope; did you mean 'exp2l'?
24094 |             out_val->data.x_f128 = exp2q(op->data.x_f128);
      |                                    ^~~~~
      |                                    exp2l
```
For stage2 it seems to require `zigUnaryMath()` to be implemented in `Sema.zig` (I'd be happy to have a go if this seems like an easy enough first compiler contribution).

---
The CI is failng on `std-i386-linux-gnu-Debug-c-multi`, hitting a segfault when trying to add two 128-bit floats. This appears to be *uncovered* by my changes, as opposed to *caused by*.

Testing performed locally:
```
$./zig test --zig-lib-dir lib/ lib/std/math/exp2.zig
All 13 tests passed.
```

Also tested using property-based testing, comparing random values to GCC's libquadmath (see https://github.com/LewisGaul/zig-f128math/blob/main/hypothesis/test.py).

---
Points for review:
- [X] There are a couple of TODOs in the boundary-case tests where I *think* the tests are correct but they're failing using the status-quo implementation
  - Have now updated the tests based on [musl libc-test](http://nsz.repo.hu/git/?p=libc-test;a=tree;f=src/math;h=84d3f245b0004edb550361a2982c7c880e697600;hb=HEAD), all TODOs resolved
- When NaN is returned, should it always be the canonical NaN? This is what I've gone for, rather than e.g. returning the input if it's NaN
  - As I've continued looking into NaN handling (e.g. in GCC) I've got a better idea of what might be desirable here, but it would be good to consider that separately to this PR
- Any suggestions on reducing repetition in the declaration of the testcases would be welcome
  - I've come up with a generic implementation in https://github.com/LewisGaul/zig-f128math/blob/main/tests/util.zig, but I can add that separately if we want it
- Would it be valuable to add any more information (in comments) about where the code comes from (and/or comments about how it works)? E.g. stuff from notes at https://github.com/LewisGaul/zig-f128math/blob/main/README.md#implementation-of-maths-functions
- May want to move to using latest musl (or other) implementation at some point, which uses a different slightly faster method...


squeek502:
Commented this in IRC but noticed you had left, so posting it here as well:

`<squeek502_>` time-to-merge/review is not indicative of interest, usually it just means that the core maintainers are busy with other stuff
`<squeek502_>` there's a lot of `else => @compileError("x not implemented for")` that apply to f128 so i'd assume more f128 support would be appreciated
`<andrewrk>` agreed with squeek502_

LewisGaul:
> Commented this in IRC but noticed you had left, so posting it here as well:

Appreciated :)

> `<squeek502_>` time-to-merge/review is not indicative of interest, usually it just means that the core maintainers are busy with other stuff

Indeed, I didn't mean to imply otherwise!

> `<squeek502_>` there's a lot of `else => @compileError("x not implemented for")` that apply to f128 so i'd assume more f128 support would be appreciated

This is more in the area I was unsure about - is the implementation in `std.math` actually expected to be used by the compiler itself? There appears to be some stuff in C++ too for stage1...

squeek502:
> This is more in the area I was unsure about - is the implementation in `std.math` actually expected to be used by the compiler itself? There appears to be some stuff in C++ too for stage1...

Take this as speculation since I haven't done any work on the compiler, but my guess is that anything f128-related that stage1 uses will need a Zig implementation for stage2 (and stage2 would use the `std.math` implementations).

LewisGaul:
Hm, looks like a segfault is being hit in one of the relevant testcases on `std-i386-linux-gnu-Debug-c-multi`, not sure why that is since it seems to pass in general... Anyone have any ideas?
(The same is seen in https://github.com/ziglang/zig/pull/10149 for `exp()`.)

andrewrk:
Looks like this was never passing the CI and now it needs rebasing. Hopefully it's clearer how to proceed based on the changes in 41dd2beaacade94c5c98400a4a655aea07b9e2f3.

It looks like there are some valuable improvements to be salvaged from this PR - here is an open invitation for @LewisGaul or anyone else to try to pull out independent mergeable pieces in fresh PRs.

PR #10051, opened 2021-10-28
fix termios not working when linking libc

### LINKED ISSUE #9707: regression after usingnamespace changes - termios broken when using pThreads
It is no longer possible to use termios

/usr/lib/zig/std/os.zig:156:27: error: container 'std.c' has no member called 'termios'
pub const termios = system.termios;

pfgithub:
fixes #9707


ifreund (approved):
LGTM, Thanks for the fix!

user673679:
Shouldn't the various constants that `termios` needs be in `std.os` too? (`IGNBRK`, `BRKINT`, etc. etc.)

PR #10053, opened 2021-10-28
fix expectStringEndsWith error output.

sbl:
before it started outputting the actual starting, not ending characters.


andrewrk:
Thanks!

Issue #10037, opened 2021-10-26
Severe Performance Issue With `std.rand.Random`


ominitay:
## Zig Version
On `0.9.0-dev.1444+e2a2e6c14`, though should apply to all versions of Zig with `std` which include this interface.

# Steps to Reproduce
To reproduce this error, an example using the current `Random` interface, and a modified copy is provided. These can be benchmarked against each other using a tool such as `hyperfine` to show the significant performance difference. (The while loop is used just to increase the number of times we run the function so that the time taken is easily measurable).

Example 1 (Current interface):
```zig
const std = @import("std");

pub fn main() !void {
    var seed: u64 = undefined;
    try std.os.getrandom(std.mem.asBytes(&seed));
    var prng = std.rand.DefaultPrng.init(seed);
    var rand = &prng.random;

    var i: usize = 0;
    var r: i32 = undefined;
    while (i < 100000000) : (i += 1) {
        r = rand.int(i32);
        std.mem.doNotOptimizeAway(&r);
    }
}
```

Example 2 (Modified interface):
```zig
const std = @import("std");

pub fn main() !void {
    var seed: u64 = undefined;
    try std.os.getrandom(std.mem.asBytes(&seed));
    var prng = std.rand.DefaultPrng.init(seed);
    var rand = Random.init(&prng);

    var i: usize = 0;
    var r: i32 = undefined;
    while (i < 100000000) : (i += 1) {
        r = rand.int(i32);
        std.mem.doNotOptimizeAway(&r);
    }
}

// Modified `Random` interface by @SpexGuy 

const Random = struct {
    self: *c_void,
    fill: fn(*c_void, []u8) void,

    pub fn init(ptr: anytype) Random {
        const Ptr = @TypeOf(ptr);
        const gen = struct {
            fn fill_erased(erased: *c_void, bytes: []u8) void {
                const alignment = @typeInfo(Ptr).Pointer.alignment;
                const self = @ptrCast(Ptr, @alignCast(alignment, erased));
                self.fill(bytes);
            }
        };
        return .{
            .self = ptr,
            .fill = gen.fill_erased,
        };
    }

    /// Returns a random int `i` such that `0 <= i <= maxInt(T)`.
    /// `i` is evenly distributed.
    pub fn int(self: Random, comptime T: type) T {
        const bits = @typeInfo(T).Int.bits;
        const UnsignedT = std.meta.Int(.unsigned, bits);
[...]

Issue #9394, opened 2021-07-16
Expose reallocBytes from std.mem.Allocator


mlugg:
I suggest that the `reallocBytes` function in `std.mem.Allocator` be made public. My use case is as follows.

I've been using https://github.com/Snektron/vulkan-zig to write some Vulkan code in Zig, and as part of this, I tried to write a simple wrapper that allows me to use an `std.mem.Allocator` as a Vulkan allocator. To do this, I need to be able to do the standard allocation operations, alloc, realloc, and free; I ended up just making alloc and free defer to the realloc implementation, so that's what we'll be concerned with here.

The realloc callback provides me with an allocation size and an alignment (both in bytes). Crucially, this alignment is not known at compile time; that makes using any function based off of `reallocAdvancedWithRetAddr` impossible, since that takes a `comptime new_alignment: u29`. Because every other exposed reallocation function in turn uses this function, I ended up having to instead use the `resizeFn` field directly. While this works fine, I ended up with a literal copy-paste of `reallocBytes` from the standard library into my own code, since it was precisely the functionality I needed. This is an unfortunate state of affairs, and one that could (hopefully) be very easily fixed by just exposing `reallocBytes`!

PR #10055, opened 2021-10-29
Allocgate

### LINKED ISSUE #10052: Allocgate
It was noticed in #10037 that LLVM is not able to properly optimize interfaces that use `@fieldParentPtr` as per @SpexGuy's [comment](https://github.com/ziglang/zig/issues/10037#issuecomment-952205513).

The `Random` interface was changed in #10045 to instead use a type erased pointer in place of `@fieldParentPointer` resulting in significant performance improvement.

This issue is to track applying the same change to `Allocator`, however due to this interface being much more widely used it is planned to be merged just before the 0.9.0 release.

Soon I will open a draft PR to begin working on this.

leecannon:
This PR is a refactor of the Allocator interface to allow inlining more often and to prevent the issue regarding `@fieldParentPointer` and dirty tracking that LLVM suffers from as explained [here](https://github.com/ziglang/zig/issues/10037#issuecomment-952205513) 

Closes #10052


Jarred-Sumner:
What if `allocator` was a `comptime` argument? Is it common to dynamically change which allocator is in use at runtime? I haven't seen code that does that. If `allocator` became a `comptime` argument, would that mean the cost of passing the `Allocator` around becomes zero? I also wonder what the impact to zig compilation time would look like

MasterQ32:
> What if `allocator` was a `comptime` argument? Is it common to dynamically change which allocator is in use at runtime? I don't think I've seen code that does that. If `allocator` became a `comptime` argument, would that mean the cost of passing the `Allocator` around becomes zero? I also wonder what the impact to zig compilation time would look like

This would not work, as `comptime` arguments must be constant at `comptime`. Something like `FixedBufferAllocator` would not work. If you only want to pass `fn allocates(comptime Allocator: type, allocator: Allocator) !void`, you'll make everything generic, and compile times will expode. You'll also be unable to use the same module with two different allocators in your project, as you will duplicate the code and all functions/variables in that module. 

Example:
Everything that takes an `ArrayList(u8)` argument will suddenly be generic over the array list, as it would `ArrayList(Allocator, u8)`. I assume this won't scale at all, will bloat the executables similar to `std.fmt` or `std.Writer`

Jarred-Sumner:
Ah, yeah that makes sense. It becomes a kind of function coloring almost.

ifreund:
A while ago on IRC @andrewrk mentioned this as one of several issues with the allocator interface design:
```
00:55 <andrewrk> I want to do another design pass on the Allocator interface 
00:55 <andrewrk> a handful of design issues has piled up
00:55 <andrewrk> * the @fieldParentPtr way of abstracting is fundamentally more difficult to optimize than a vtable strategy
00:56 <andrewrk> * the mechanism to get more bytes than you asked for is dubious and might be at worst doing more harm than good and at best a bunch of complication for no benefit
00:57 <andrewrk> * I want allocator implementations to have access to a comptime alignment. maybe even the type??
00:57 <andrewrk> * @memset to undefined should happen in the Allocator implementations rather than the interface to avoid redundant memsets in safe modes
00:59 <andrewrk> * the API is a bit of a mess
```
In particular, I wonder how this changes interacts with the desire to make a comptime alignement or type available to the allocator implementations. It would be best to do one massively breaking change instead of two if at all possible.

marler8997:
> * the mechanism to get more bytes than you asked for is dubious and might be at worst doing more harm than good and at best a bunch of complication for no benefit

I'm curious and @andrewrk reasoning on this.  If it really has no benefit then we can remove it, but I believe there's use cases for it (i.e. ArrayList), but maybe Andrew thinks those use cases can use another mechanism and be just as good?

For most cases, realloc is fine.  The only case I think think of where realloc wouldn't necessarily work is if you just want to know how much space you have available, because you're not sure how much you'll need yet but you would benefit from knowing the full amount you have available to you.  Without an interface to get the available length, you might realloc more than is available but you end up not actually needing it, so you would end up moving memory when you didn't need to.

marler8997:
Here's an idea that should remove the need for `len_align`, remove the `@fieldParentPtr` mechanism and provide access to comptime alignment

```zig
///
/// A new allocator API.  Instead of an allocator providing 2 function pointers, they
/// provide a single function pointer that takes a tagged union represnting an "AllocatorRequest".
/// This allows us to enhance/scale the allocator interface with more operations without having
/// to add additional function pointers.
///
/// Using this new ability to add more operations, we can remove the `len_align` parameter to the
/// allocation functions and add a `query_full_capacity` request intead.
///
/// We can also use this new ability to easily add operations to support comptime aligned allocations.
///
/// Note: I've added context pointer fo the AllocFn so we don't need to depend on `@fieldParentPtr`
pub const AllocFn = fn (self: *c_void, request: *AllocatorRequest);
pub const AllocatorRequest = union(enum) {

    alloc: struct { len: usize, align: u29, ret_addr: usize },

    // means we can allocation with comptime alignments up to 16
    alloc_align2: struct { len: usize, ret_addr: usize },
    alloc_align4: struct { len: usize, ret_addr: usize },
    alloc_align8: struct { len: usize, ret_addr: usize },
    alloc_align16: struct { len: usize, ret_addr: usize },

    resize: Resize,

    /// returns the full capacity available to an allocation
    query_full_capacity: struct { buf: []u8, align: u29 },

    pub const Resize = struct {
        buf: []u8,
        align: u29,
        new_len: usize,
        ret_addr: usize,
    };
}

```

SpexGuy:
`union (enum)` with runtime known fields has a runtime known tag, so I don't see how this makes anything comptime-known.  Except that you could make specialized code paths, but you can do that anyway with the current pattern, and it's not something we should force allocators to do.  Running everything through one function seems potentially dubious, since in the case where the function is not inlined it's a lot of unnecessary branching.  This pattern means that in order to query the size of an allocation, you need to construct a 40 byte stack value in memory and then pass the address of that into a virtual function.  While this does solve the api problems, so does a vtable, and I think the vtable approach is probably better/faster/easier to optimize.

marler8997:
> union (enum) with runtime known fields has a runtime known tag, so I don't see how this makes anything comptime-known. Except that you could make specialized code paths, but you can do that anyway with the current pattern, and it's not something we should force allocators to do.

Oh yeah now that I think about it you might be right here. I think I need some clarification from Andrew about what he means by comptime-known alignment.  My suggestion here was based on an old suggestion from him where he added new allocation function pointers for a limited set of alignments.  That being said, using function pointers don't have the same downside as `union (enum)` (like you say) since function pointers don't need to branch like a tagged union would.

> Running everything through one function seems potentially dubious, since in the case where the function is not inlined it's a lot of unnecessary branching. 

I assume by "dubious" you mean "slower than an alternative".  I can't say here, but I think we should measure if we decide to add more "operations", either via union(enum) or vtable.

> This pattern means that in order to query the size of an allocation, you need to construct a 48 byte stack value in memory and then pass the address of that into a virtual function.

We need to pass these 48 bytes of memory somehow, we can't get around that.  I can't say which mechanism is faster though.  I will give you that with a tagged union, there will be some cases that hog more stack memory than they need to, but in others I'm guessing they wouldn't?  Does a tagged union with a "small activated field" only take up as much stack memory as that field needs?  Would be interested to know.

> While this does solve the api problems, so does a vtable, and I think the vtable approach is probably better/faster/easier to optimize.

You could be right here and I think you have more insight on this than myself.  I think the approach of breaking up our 2 operations into smaller ones has merit, whether it's better to use a vtable or a tagged union, I'm not sure.

lib/std/mem/Allocator.zig
```diff
@@ -5,168 +5,181 @@ const assert = std.debug.assert;
...
-/// reallocation failure, even when `new_n` <= `old_mem.len`. A `FixedBufferAllocator`
-/// would always return `error.OutOfMemory` for `reallocFn` when the size/alignment
-/// is less than or equal to the old allocation, because it cannot reclaim the memory,
-/// and thus the `std.ArrayList` would be better off retaining its capacity.
-/// When `reallocFn` returns,
-/// `return_value[0..min(old_mem.len, new_byte_count)]` must be the same
-/// as `old_mem` was when `reallocFn` is called. The bytes of
-/// `return_value[old_mem.len..]` have undefined values.
-/// The returned slice must have its pointer aligned at least to `new_alignment` bytes.
-fn reallocBytes(
```
ominitay, line 83: We should keep this in some form to accommodate for #9394.

leecannon: Added back

lib/std/mem/Allocator.zig
```diff
@@ -5,58 +5,163 @@ const assert = std.debug.assert;
...
+    /// `ret_addr` is optionally provided as the first return address of the allocation call stack.
+    /// If the value is `0` it means no return address has been provided.
+    resize: fn (ptr: *c_void, buf: []u8, buf_align: u29, new_len: usize, len_align: u29, ret_addr: usize) ?usize,
+
+    /// Free and invalidate a buffer. `buf.len` must equal the most recent length returned by `alloc` or `resize`. 
+    /// `buf_align` must equal the same value that was passed as the `ptr_align` parameter to the original `alloc` call.
+    ///
+    /// `ret_addr` is optionally provided as the first return address of the allocation call stack.
+    /// If the value is `0` it means no return address has been provided.
+    free: fn (ptr: *c_void, buf: []u8, buf_align: u29, ret_addr: usize) void,
```
marler8997, line 52: Is the purpose of adding a `free` function instead of passing size `0` to `resize` to avoid a common branch in `resize`?  Also, would passing `0` to resize be an error now, or would it redirect to `free` or something else?

leecannon: Yes, every nontrivial allocator implementation I've seen needs to branch to treat free and shrink differently.

andrewrk:
@leecannon I'd like to merge this without a6246c88af48472e68ecf4f6a19a74da9889ce12. I think we should change one thing at a time and now we have the luxury of looking at how the graphs on https://ziglang.org/perf/ will change with each thing.

Can we do the interface change alone?

leecannon:
Okay I have pulled the shrink change out and will make a new PR for it once this is merged.

And just to note the shrink change is a correctness issue not a performance improvement, which is even more reason to split it out.

andrewrk (approved):
Great work. Thank you for taking on such a monumental task.

Ready to merge as soon as we have the green light.

andrewrk:
Looking at https://ziglang.org/perf/ - I see major regression with maxrss on a few benchmarks after merging this. For example, the arena-allocator benchmark went from using 4 MiB to 11 GiB.

SpexGuy:
Holy crap! There must be a broken free somewhere, the increased size of Allocator can only have a maximum of 2x increase in memory use if every piece of memory in the program is an Allocator.

PR #10056, opened 2021-10-29
zig: expose linker options and include '-z notext'

lithdew:
Add an option to allow the '-z notext' option to be passed to the linker via. the compiler frontend, which is a flag that tells the linker that relocations in read-only sections are permitted. Certain targets such as Solana BPF rely on this flag.

Expose all linker options i.e. '-z nodelete', '-z now', '-z relro' in the compiler frontend. Usage documentation has been updated accordingly.

Expose the '-z notext' flag in the standard library build runner.


src/main.zig
```diff
@@ -401,6 +401,14 @@ const usage_build_generic =
     \\  -fno-allow-shlib-undefined     Disallows undefined symbols in shared libraries
     \\  --eh-frame-hdr                 Enable C++ exception handling by passing --eh-frame-hdr to linker
     \\  --emit-relocs                  Enable output of relocation sections for post build tools
+    \\  -z [arg]                       Append linker arguments
```
kubkon, line 404: Hmm, `-z <arg>` seems specific to ELF linkers; would it make more sense to expose each flag of interest separately with `-fnotext` or something similar? Also, we should definitely point out these are ELF-specific. @andrewrk I'd appreciate your look at this too!

Luukdegram: `-z` specifies an extension flag. `wasm-ld` supports them as well and we actually already make use of it to set the stack size for executables. So if we go with this, perhaps those should be supported too.

andrewrk:
```suggestion
    \\  -z [arg]                       Set linker extension flags
```

andrewrk:
```suggestion
    \\  -z [arg]                       Set linker extension flags
```

src/main.zig
```diff
@@ -1086,6 +1095,29 @@ fn buildOutputType(
                         linker_allow_shlib_undefined = true;
                     } else if (mem.eql(u8, arg, "-fno-allow-shlib-undefined")) {
                         linker_allow_shlib_undefined = false;
+                    } else if (mem.eql(u8, arg, "-z")) {
+                        i += 1;
+                        if (i >= args.len) {
+                            fatal("expected linker arg after '{s}'", .{arg});
```
andrewrk, line 1101:
```suggestion
                            fatal("expected linker extension flag after '{s}'", .{arg});
```

andrewrk (approved):

andrewrk:
Damn I was hoping github would let me make these changes before merging

PR #10058, opened 2021-10-29
add pathJoin to builder

marler8997:
Added a public convenience function to `Builder`.   Now instead of doing this:
```zig
    std.fs.path.join(builder.allocator, &[_][]const u8 { "foo", "bar", "baz"}) catch unreacahble
    // or with this shorthand
    std.fs.path.join(builder.allocator, &.{ "foo", "bar", "baz"}) catch unreacahble
```
`build.zig` files can now do this:
```zig
    builder.pathJoin(&[_][]const u8 { "foo", "bar", "baz"})
    // or with this shorthand
    builder.pathJoin(&.{ "foo", "bar", "baz"})
```

I've found this is a common operation, and this convenience function sits alongside existing ones like `builder.pathFromRoot` and `builder.fmt`.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -204,10 +204,10 @@ pub const Builder = struct {
     pub fn resolveInstallPrefix(self: *Builder, install_prefix: ?[]const u8, dir_list: DirList) void {
         if (self.dest_dir) |dest_dir| {
             self.install_prefix = install_prefix orelse "/usr";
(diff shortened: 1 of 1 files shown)


dbandstra:
Aside: Doesn't the compiler accept this nowadays?

```zig
builder.pathJoin(&.{"foo", "bar", "baz"})
```

marler8997:
@dbandstra I didn't realize that.  I've added an additional commit to use that shorthand instead.

andrewrk (approved):

PR #10059, opened 2021-10-29
stage2: Add support for floats in the C backend

zenith391:


src/link/C/zig.h
```diff
@@ -356,6 +357,18 @@ static inline long long zig_subw_longlong(long long lhs, long long rhs, long lon
     return (long long)(((unsigned long long)lhs) - ((unsigned long long)rhs));
 }
 
+static inline float zig_bitcast_uint32_t_to_float(uint32_t arg) {
+    float dest;
+    memcpy(&dest, &arg, sizeof dest);
```
doppioandante, line 362: Question: isn't something like dest = *((float *)&arg) supposed to work?

SpexGuy: Many compilers will generate correct code because this is a common idiom, but it is technically UB.  Because of strict aliasing rules, `float *` may not alias `uint32_t *`.  The compiler has a thorough understanding of memcpy though, and will not actually call a function here.  See also https://en.cppreference.com/w/cpp/language/reinterpret_cast

SpexGuy: (FWIW this is well-defined in Zig because our aliasing rules are less strict.  This will make it extra difficult to generate correct C code.)

andrewrk:
Try this test case:

```zig
const std = @import("std");
const x = std.math.nan_f32;
export fn foo() f32 {
    return x;
}
```

src/link/C/zig.h
tiehuis: How about doing this?

```
union { float f; uint32_t u } cast;
cast.u = arg;
return cast.f;
```

Type-punning is legal in C99 to reintepret bits and I presume we are targeting C99 at a minimum, as we rely on stdint at this time.

andrewrk: See my test case below. We need a way to lower it in a way that will work inside a global variable initializer.

SpexGuy: The union approach is also UB in C++, for the same reason, but both are well defined in C99.  So if we are targeting C99 to be compiled only by a C compiler (not a C++ compiler), this is fine.

Re: lowering, one approach would be to generate the global for this value as a uint32_t, but then generate references to this value as stub code that converts it to a float at runtime.  (and similarly, taking the address of this value gets an untyped char pointer that will be memcpy'd into any float values when read).

zenith391:
Your test case simply generates
```c
ZIG_EXTERN_C float foo(void);

float foo(void) {
 #line 4
 return zig_bitcast_uint32_t_to_float(0x7f800001);
}
```

Testing it from another Zig file with `std.log.info("foo() = {d}", .{ foo() });`, I get `info: foo() = nan` so this confirms that it is recognised as NaN.
`@bitCast(u32, foo()) == std.math.nan_u32` returns true, which also means that the specific NaN payload is conserved.

zenith391:
Using
```zig
pub export const x: u32 = 0;
export fn foo() f32 {
    return x;
}
```
Which I think is what you meant, we see the C backend doesn't generate a global variable. It's also the case with integers. So it seems to be a separate issue but we currently don't generate global variables.

zenith391:
So is there anything blocking merging this pull request now?

src/codegen/c.zig
```diff
@@ -219,11 +219,36 @@ pub const DeclGen = struct {
...
+                // Using '{}' for integer and floats seemed to error C compilers (both GCC and Clang)
+                // with 'error: expected expression' (including when built with 'zig cc')
+                .Int => {
+                    const c_bits = toCIntBits(ty.intInfo(dg.module.getTarget()).bits) orelse
+                            return dg.fail("TODO: C backend: implement integer types larger than 128 bits", .{});
+                    switch (c_bits) {
+                        8    => return writer.writeAll("0xaa"),
+                        16   => return writer.writeAll("0xaaaa"),
+                        32   => return writer.writeAll("0xaaaaaaaa"),
+                        64   => return writer.writeAll("0xaaaaaaaaaaaaaaaa"),
```
andrewrk, line 232: This needs a suffix on the integer literal, right?

src/codegen/c.zig
```diff
@@ -219,11 +219,36 @@ pub const DeclGen = struct {
...
+                        128  => return writer.writeAll("0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"),
+                        else => unreachable
+                    }
+                },
+                .Float => {
+                    switch (ty.floatBits(dg.module.getTarget())) {
+                        32 => return writer.writeAll("zig_bitcast_uint32_t_to_float(0xaaaaaaaa)"),
+                        64 => return writer.writeAll("zig_bitcast_uint64_t_to_double(0xaaaaaaaaaaaaaaaa)"),
+                        else => unreachable // other float types aren't support in renderType() as of now
```
andrewrk, line 241:
```suggestion
                        else => return dg.fail("TODO float types > 64 bits are not support in renderType() as of now", .{}),
```

src/link/C/zig.h
```diff
@@ -356,6 +357,18 @@ static inline long long zig_subw_longlong(long long lhs, long long rhs, long lon
...
+    return dest;
+}
+
+static inline float zig_bitcast_uint64_t_to_double(uint64_t arg) {
```
andrewrk, line 366:
```suggestion
static inline float zig_bitcast_f64_u64(uint64_t arg) {
```

src/link/C/zig.h
```diff
@@ -356,6 +357,18 @@ static inline long long zig_subw_longlong(long long lhs, long long rhs, long lon
     return (long long)(((unsigned long long)lhs) - ((unsigned long long)rhs));
 }
 
+static inline float zig_bitcast_uint32_t_to_float(uint32_t arg) {
```
andrewrk, line 360:
```suggestion
static inline float zig_bitcast_f32_u32(uint32_t arg) {
```

andrewrk (changes requested):
Thanks for the patch. I have some minor suggestions and then this is ready to land.

zenith391:
Ok so I think I implemented most of your suggestions, but I'm just not sure on the integer literal suffix

zenith391:
Any update?

PR #10060, opened 2021-10-29
add print field to RunStep

marler8997:
A new print field is added to RunStep that will control whether it prints the cmd before running it.  By default it will be set to `builder.verbose`.  This makes the default match what one would expect the `zig build --verbose` flag to do (see below) while still allowing each build to customize it:

```
  --verbose                    Print commands before executing them
```

### DIFF
--- a/lib/std/build/RunStep.zig
+++ b/lib/std/build/RunStep.zig
@@ -37,6 +37,9 @@ stdin_behavior: std.ChildProcess.StdIo = .Inherit,
 
 expected_exit_code: u8 = 0,
 
+/// Print the command before running it
+print: bool,
+
 pub const StdIoAction = union(enum) {
     inherit,
     ignore,
@@ -58,6 +61,7 @@ pub fn create(builder: *Builder, name: []const u8) *RunStep {
         .argv = ArrayList(Arg).init(builder.allocator),
         .cwd = null,
         .env_map = null,
+        .print = builder.verbose,
     };
     return self;
 }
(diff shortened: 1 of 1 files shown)


lib/std/build/RunStep.zig
```diff
@@ -181,6 +185,9 @@ fn make(step: *Step) !void {
     child.stdout_behavior = stdIoActionToBehavior(self.stdout_action);
     child.stderr_behavior = stdIoActionToBehavior(self.stderr_action);
 
+    if (self.print)
+        printCmd(cwd, argv);
```
matu3ba, line 189: To me it doesnt make sense to print to stderr, if the process does not return an error code. [Some argument for this behavior](https://unix.stackexchange.com/a/331831).
The user asked explicitly for the output, which should then go to stdout.

lib/std/build/RunStep.zig
```diff
@@ -37,6 +37,9 @@ stdin_behavior: std.ChildProcess.StdIo = .Inherit,
 
 expected_exit_code: u8 = 0,
 
+/// Print the command before running it
+print: bool,
```
matu3ba, line 41: print implies to me `std.debug.print` and we are logging to stderr. Is there at any documented convention why/when to do this?
Argumentation below.

matu3ba (changes requested):
Some feedback why logging to stderr in the case of no error occurring would be nice.

lib/std/build/RunStep.zig
marler8997: That link you sent seems to have arguments for both sides, I don't really see consensus.  Because there doesn't seem to be a clear standard on this, I'm hesitant to go against what the rest of zig build is doing without a maintainer (someone who will merge it) blessing such a change.

matu3ba: Ok, sounds good to me then. You can mark this as resolved then.

PR #10062, opened 2021-10-29
Add convenient create function to Builder

marler8997:
An idea to save some LOC in `build.zig` files.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -132,7 +132,7 @@ pub const Builder = struct {
         include_dir: ?[]const u8 = null,
     };
 
-    pub fn create(
+    pub fn createBuilder(
         allocator: *Allocator,
         zig_exe: []const u8,
         build_root: []const u8,
@@ -240,6 +240,13 @@ pub const Builder = struct {
             FileSource{ .path = p }
         else
             null;
+    }
+
+    /// Create the given object with the builder allocator
+    pub fn create(self: Builder, obj: anytype) *@TypeOf(obj) {
+        var ptr = self.allocator.create(@TypeOf(obj)) catch @panic("Out of memory");
+        ptr.* = obj;
+        return ptr;
     }
 
     pub fn addExecutable(self: *Builder, name: []const u8, root_src: ?[]const u8) *LibExeObjStep {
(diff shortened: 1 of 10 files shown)


lib/std/build.zig
```diff
@@ -242,6 +242,13 @@ pub const Builder = struct {
             null;
     }
 
+    /// Create the given object with the builder allocator
+    pub fn create(self: Builder, obj: anytype) *@TypeOf(obj) {
+        var ptr = self.allocator.create(@TypeOf(obj)) catch unreachable;
```
ifreund, line 247: Could we make this `@panic()` on failure instead? I know that the build system is always run in a safe build mode and that `catch unreachable` is currently used everywhere in `std.build` but it's still semantically incorrect and I don't want anyone to read this standard library code and think the style is ok in general.

marler8997: Agreed, changed.

lib/std/build.zig
```diff
@@ -132,7 +132,7 @@ pub const Builder = struct {
         include_dir: ?[]const u8 = null,
     };
 
-    pub fn create(
+    pub fn createBuilder(
```
andrewrk, line 135: Before, the fully qualified namespace for this function was: `Builder.create`. With this change, it is `Builder.createBuilder`.

lib/std/build.zig
```diff
@@ -242,6 +242,13 @@ pub const Builder = struct {
             null;
     }
 
+    /// Create the given object with the builder allocator
+    pub fn create(self: Builder, obj: anytype) *@TypeOf(obj) {
```
andrewrk, line 246: obj: anytype makes me nervous because you really should be telling the allocator exactly what type you want to allocate, and then we would want the initialization value to coerce to it. But there's a reason that Allocator does not accept an initialization value - it used to but then we realized that a common use case was to write data into place rather than copying it.

andrewrk (changes requested):
ehhhhhh I'm not sold on this change. I think abstracting the Allocator interface create function in this way is not ultimately going to lead to better code or more convenience.

marler8997:
> ehhhhhh I'm not sold on this change. I think abstracting the Allocator interface create function in this way is not ultimately going to lead to better code or more convenience.

Fair enough, closed.

lib/std/build.zig
marler8997: I would have thought RLS and "automatic parameter references" would elide any copy in this case?

PR #10063, opened 2021-10-30
C backend: implement signed trunc

rainbowbismuth:
Andrew was very close on the recent "adding to the C backend" stream. 

Unfortunately in C, `x << y` is undefined if `x < 0`. The code written on the stream would have worked without the undefined behavior sanitizer that `zig run` provides by default. Thankfully all we have to do is cast to unsigned and back to signed before the right shift.

All of the tests in `behavior/truncate.zig` passed without other changes.

Also, to verify the test cases I just plugged it into some handwritten LLVM IR in the compiler explorer: https://godbolt.org/z/vzvncr4a1

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -1389,7 +1389,14 @@ fn airTrunc(f: *Function, inst: Air.Inst.Index) !CValue {
             return local;
         },
         .signed => {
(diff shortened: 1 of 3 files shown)


rainbowbismuth:
My push fixed some (zig) formatting issues in the original.

andrewrk:
Thanks!

PR #10066, opened 2021-10-31
stage2 AArch64: introduce MIR

joachimschmidt557:
This PR transitions the AArch64 backend to using the new MIR. Most of the instructions available previously are available with the MIR intermediate step. Things that don't work yet and will be implemented in a follow-up PR:

- branch instructions
- debug info for function arguments

### DIFF
--- a/src/arch/aarch64/CodeGen.zig
+++ b/src/arch/aarch64/CodeGen.zig
@@ -315,6 +315,7 @@ pub fn generate(
 
     var emit = Emit{
         .mir = mir,
+        .bin_file = bin_file,
         .target = &bin_file.options.target,
         .code = code,
     };
@@ -335,6 +336,25 @@ fn addInst(self: *Self, inst: Mir.Inst) error{OutOfMemory}!Mir.Inst.Index {
     const result_index = @intCast(Air.Inst.Index, self.mir_instructions.len);
     self.mir_instructions.appendAssumeCapacity(inst);
     return result_index;
+}
+
+pub fn addExtra(self: *Self, extra: anytype) Allocator.Error!u32 {
+    const fields = std.meta.fields(@TypeOf(extra));
+    try self.mir_extra.ensureUnusedCapacity(self.gpa, fields.len);
+    return self.addExtraAssumeCapacity(extra);
+}
+
+pub fn addExtraAssumeCapacity(self: *Self, extra: anytype) u32 {
+    const fields = std.meta.fields(@TypeOf(extra));
+    const result = @intCast(u32, self.mir_extra.items.len);
+    inline for (fields) |field| {
+        self.mir_extra.appendAssumeCapacity(switch (field.field_type) {
+            u32 => @field(extra, field.name),
+            i32 => @bitCast(u32, @field(extra, field.name)),
+            else => @compileError("bad field type"),
+        });
+    }
+    return result;
 }
 
 fn gen(self: *Self) !void {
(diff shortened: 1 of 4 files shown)


src/arch/aarch64/CodeGen.zig
```diff
@@ -1789,25 +1493,17 @@ fn airCall(self: *Self, inst: Air.Inst.Index) !void {
...
-                    .target = .{ .global = n_strx },
-                    .addend = 0,
-                    .subtractor = null,
-                    .pcrel = true,
-                    .length = 2,
-                    .@"type" = @enumToInt(std.macho.reloc_type_arm64.ARM64_RELOC_BRANCH26),
+
+                _ = try self.addInst(.{
+                    .tag = .call_extern,
+                    .data = .{ .imm32 = n_strx },
```
kubkon, line 1506: Just a nit for the future, not necessarily now. `imm32` might be confusing as `n_strx` is not actually an immediate. Perhaps it would make sense to introduce another tagged variant such as `extern: u32`? This shouldn't increase the size of `data` in memory and would make the code slightly more readable.

src/arch/aarch64/Mir.zig
```diff
@@ -0,0 +1,208 @@
...
+        /// Load Register Byte
+        // TODO: split into ldrb_immediate and ldrb_register
+        ldrb,
+        /// Load Register Halfword
+        // TODO: split into ldrh_immediate and ldrh_register
+        ldrh,
+        /// Move (to/from SP)
+        mov_to_from_sp,
+        /// Move (register)
+        mov_register,
```
kubkon, line 63: I take it that there's no way in making these two pseudo-inst common?

src/arch/aarch64/Emit.zig
```diff
@@ -0,0 +1,403 @@
...
+            .offset = offset,
+            .target = .{ .global = n_strx },
+            .addend = 0,
+            .subtractor = null,
+            .pcrel = true,
+            .length = 2,
+            .@"type" = @enumToInt(std.macho.reloc_type_arm64.ARM64_RELOC_BRANCH26),
+        });
+    } else {
+        @panic("Implement call_extern for linking backends != MachO");
```
kubkon, line 258: How about returning an explicit error here? I think a runtime trap generating a stack trace might obfuscate the actual error which is a handled case, rather than an internal compiler error.

kubkon (changes requested):
All in all, this looks great - fantastic job! One request I have is to return explicit errors where functionality is missing rather than panicking. Also, do you think it would make sense to tie those errors with the compiler's error mechanism like in [here](https://github.com/ziglang/zig/blob/969bcb6a59eadb2ef46a9784286728a25d275c24/src/codegen.zig#L252)?

src/arch/aarch64/CodeGen.zig
joachimschmidt557: Makes sense, I'll do this!

src/arch/aarch64/Mir.zig
joachimschmidt557: I separated them because they are two separate instructions according to the ARM Architecture Reference Manual.

src/arch/aarch64/Emit.zig
joachimschmidt557: I agree, the `ErrorMsg` pattern used in `CodeGen.zig` can also be applied here.

kubkon (approved):
Thanks for addressing the comments, I think it's ready to merge!

PR #10068, opened 2021-10-31
stage2: tracy integration improvements

leecannon:
This PR extends the tracy integration in stage2 to include:

- More features of tracy are exposed in tracy.zig including but not limited to adding text/values to zones, message emission, frame marking, etc
- Added new `zig build` option `-Dtracy-callstack` which will make tracy grab call stacks. Currently hardcoded in tracy.zig to a depth of 10, this could be made configurable later.
- Use named discontinuous frames to mark the various stages of compilation in Compilation.zig; astgen, liveness, sema, etc.
- When in `--watch` mode every recompilation marks a new frame.
- Added new `zig build` option `-Dtracy-allocation` which will wrap the gpa allocator passed to `mainArgs` from `main` in a `tracy.TracyAllocator`. This provides memory allocation data, if the `-Dtracy-callstack` option is given in addition to this then allocation events will also capture call stack data.


andrewrk:
Very nice!

PR #10071, opened 2021-10-31
C backend: implement ?void

rainbowbismuth:
Since there is no type in C that can take up zero bytes in a struct, I've implemented `?void` as a simple `bool` instead of something like `typedef struct { bool is_null; } zig_Q_void;` This requires some special casing, but I figure simpler generated code is better.

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -308,6 +308,11 @@ pub const DeclGen = struct {
                 if (ty.isPtrLikeOptional()) {
                     return dg.renderValue(writer, payload_type, val);
                 }
+                const target = dg.module.getTarget();
+                if (payload_type.abiSize(target) == 0) {
+                    const is_null = val.castTag(.opt_payload) == null;
+                    return writer.print("{}", .{is_null});
+                }
                 try writer.writeByte('(');
                 try dg.renderType(writer, ty);
                 try writer.writeAll("){");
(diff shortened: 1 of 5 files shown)


g-w1:
What about generating `?struct {}`, also zero bytes. Maybe instead of checking if it is void, check if the type is zero bytes?

rainbowbismuth:
Excellent point, I will try that.

src/type.zig
```diff
@@ -3474,6 +3474,9 @@ pub const Type = extern union {
     pub fn structFieldOffset(ty: Type, index: usize, target: Target) u64 {
         switch (ty.tag()) {
             .@"struct" => {
+                if (index == 0) {
```
rainbowbismuth, line 3477: I misunderstood what this function does, I think it would be better to change `abiSize()` directly

src/type.zig
```diff
@@ -1805,7 +1805,13 @@ pub const Type = extern union {
             .void,
             => 0,
 
-            .@"struct" => return self.structFieldOffset(self.structFieldCount(), target),
+            .@"struct" => {
+                const field_count = self.structFieldCount();
+                if (field_count == 0) {
+                    return 0;
+                }
+                return self.structFieldOffset(self.structFieldCount(), target);
```
andrewrk, line 1813: Looks like you meant to do this: 

```suggestion
                return self.structFieldOffset(field_count, target);
```

src/codegen/c.zig
```diff
@@ -308,6 +308,10 @@ pub const DeclGen = struct {
                 if (ty.isPtrLikeOptional()) {
                     return dg.renderValue(writer, payload_type, val);
                 }
+                if (payload_type.abiSize(builtin.target) == 0) {
```
andrewrk, line 311: Be sure to pass in the target that is chosen to build for, not the host target that the compiler is running on! Look for `const target = ` elsewhere in this file. I see other instances of `builtin.target` in this patch too.

andrewrk (changes requested):
Thanks for the patch! I see a couple of cross-compilation bugs that should be trivial to resolve.

rainbowbismuth:
Thanks for the pointers Andrew :)

PR #10073, opened 2021-11-01
more haiku support 

hoanga:
Hello and happy halloween!

I was able to finally get some time to sit down and do some more continued work on improving Haiku support for Zig. The following changesets should allow

`zig build in a (simple) zig project works
`

Running make install to install Zig after compiling probably doesn't work and is on the list of things todo. Debug still remains on the open list of items to tackle at some point but wanted to attempt with smaller changesets. Also included in the changes are some better error checking for some syscalls to Haiku and an attempted cleanup of some defines and constants in the Haiku support section that came across from an older copy and paste. 

Getting Zig to compile on Haiku also meant using the current used version of LLVM (13) which is currently a (hopefully) pending port for Haiku. I am attaching some patches that I used to get LLVM to build on Haiku as well for reference but not including it in the PR changesets directly.

Sample Runs

```
~/src/git/zig/build> uname -a
Haiku shredder 1 hrev55602 Oct 29 2021 07:10:38 x86_64 x86_64 Haiku

~/src/git/zig/build> ./zig version
0.9.0-dev.1463+d97928bf5

## compiling with  simple program
~/src/git/zig/build> cd hellohaiku

~/src/git/zig/build/hellohaiku> cat src/main.zig 
const std = @import("std");
const thread = std.Thread;

pub fn main() anyerror!void {
    std.log.info("Num CPUs: {d}.", .{thread.getCpuCount()});
}

~/src/git/zig/build/hellohaiku> ../zig build run
info: All your codebase are belong to us.
info: Num CPUs: 16.


### try using zig build on some other zig codebases

~/src/git/zig/build/ziglings> ../zig build 8
Compiling 008_quiz.zig...
Checking 008_quiz.zig...
PASSED: Program in Zig!


~/src/git/zig/build/ziglings> ../zig build 31
Compiling 031_switch2.zig...
Checking 031_switch2.zig...
PASSED: ZIG!!
```

### DIFF
--- a/lib/std/c/haiku.zig
+++ b/lib/std/c/haiku.zig
@@ -145,28 +145,29 @@ pub const Kevent = extern struct {
 // include/dlfcn.h
 
 pub const POLL = struct {
-    pub const IN = 0x0001;
-    pub const ERR = 0x0004;
-    pub const NVAL = 0x1000;
-    pub const HUP = 0x0080;
+    /// input available
+    pub const IN = 70;
+    /// output available
+    pub const OUT = 71;
+    /// input message available
+    pub const MSG = 72;
+    /// I/O error
+    pub const ERR = 73;
+    /// high priority input available
+    pub const PRI = 74;
+    /// device disconnected
+    pub const HUP = 75;
 };
 
 pub const RTLD = struct {
-    /// Bind function calls lazily.
-    pub const LAZY = 1;
-    /// Bind function calls immediately.
-    pub const NOW = 2;
-    pub const MODEMASK = 0x3;
-    /// Make symbols globally available.
-    pub const GLOBAL = 0x100;
-    /// Opposite of GLOBAL, and the default.
+    /// relocations are performed as needed
+    pub const LAZY = 0;
+    /// the file gets relocated at load time
+    pub const NOW = 1;
+    /// all symbols are available
+    pub const GLOBAL = 2;
+    /// symbols are not available for relocating any other object
     pub const LOCAL = 0;
-    /// Trace loaded objects and exit.
-    pub const TRACE = 0x200;
-    /// Do not remove members.
-    pub const NODELETE = 0x01000;
-    /// Do not load if not already loaded.
-    pub const NOLOAD = 0x02000;
 };
 
 pub const dl_phdr_info = extern struct {
@@ -407,18 +408,11 @@ pub const sockaddr = extern struct {
     };
 };
 
-pub const CTL = struct {
-    pub const KERN = 1;
-    pub const DEBUG = 5;
-};
+pub const CTL = struct {};
 
-pub const KERN = struct {
-    pub const PROC = 14; // struct: process entries
-    pub const PROC_PATHNAME = 12; // path to executable
-    pub const IOV_MAX = 1024;
-};
+pub const KERN = struct {};
 
-pub const IOV_MAX = KERN.IOV_MAX;
+pub const IOV_MAX = 1024;
 
 pub const PATH_MAX = 1024;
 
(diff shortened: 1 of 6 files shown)


hoanga:
patches applied to (zig-bootstrap) LLVM
[llvm13-diff.patch.zip](https://github.com/ziglang/zig/files/7450509/llvm13-diff.patch.zip)

andrewrk:
CI is failing on `zig fmt`

andrewrk:
CC @waddlesplash who gave me permission on IRC to ask for help reviewing :-)

hoanga:
oops, sorry about the fmt miss. i have gone through and double checked that all files that have been changed have had fmt applied to them and pushed updates, hopefully CI will find that more agreeable.

lib/std/c/haiku.zig
```diff
@@ -400,18 +440,12 @@ pub const MAP = struct {
     pub const FAILED = @intToPtr(*c_void, maxInt(usize));
     pub const SHARED = 0x0001;
     pub const PRIVATE = 0x0002;
-    pub const FIXED = 0x0010;
+    pub const FIXED = 0x0004;
     pub const STACK = 0x0400;
```
waddlesplash, line 444: We don't have some of these flags (and we do have other flags not listed here.) Why not just syncronize the full list with the mmap flags we actually have?

lib/std/c/haiku.zig
```diff
@@ -282,15 +292,29 @@ pub const dirent = extern struct {
     }
 };
 
+pub const area_info = extern struct {
+    area: u32,
+    name: [32]u8,
```
waddlesplash, line 297: [32] is from the B_OS_NAME_LENGTH constant. I don't know if that should be preserved here or not...

lib/std/c/haiku.zig
```diff
@@ -535,9 +569,6 @@ pub const O = struct {
     pub const RDWR = 0x0002;
     pub const ACCMODE = 0x0003;
 
-    pub const SHLOCK = 0x0010;
-    pub const EXLOCK = 0x0020;
-
     pub const CREAT = 0x0200;
     pub const EXCL = 0x0800;
     pub const NOCTTY = 0x8000;
```
waddlesplash, line 574: These constants, also, do not seem to properly correspond to ours. They should probably all be fixed. (I think a lot of the constants not touched in these commits have similar problems, maybe a general review of the constants in this file is in order.)

lib/std/c/haiku.zig
hoanga: the mmap flags that i am aware of are the following:

```
MAP_SHARED
MAP_PRIVATE
MAP_FIXED
MAP_ANONYMOUS
MAP_ANON
MAP_NORESERVE
```

i can work on syncing these and see how it goes. if there is any other flags that should be included, let me know.

i can go through another look over the other constants when i get some more spare time. if there are particular ones to focus on, am open to suggestions.

lib/std/c/haiku.zig
hoanga: i would be open to suggestions rather sticking with this hard-coded value (perhaps use a constant with corresponding name?). another suggestion was to include many of the B_* constants that after taking a brief look at it, it might be better split off into another more directed changeset.

lib/std/c/haiku.zig
hoanga: see above

lib/std/c/haiku.zig
waddlesplash: Yes, those are all the ones currently in mman.h.

I would look at the error constants especially.

andrewrk:
Let's please correct all the constants in this PR, and delete any constants that have not been audited. I don't want any incorrect haiku constants in the zig standard library. Let me know when that's done and I'll look into merging the PR.

hoanga:
thank you for the feedback. when i get some time (hopefully this weekend), i will try to look over the constants and get some updates in.

hoanga:
have went through and cleaned up as many of the constants so they should match the haiku definitions that i could find. i think it is ready for review (assuming the build is okay with it too)

lib/std/c/haiku.zig
hoanga: updated

andrewrk:
Thanks @hoanga!

PR #10076, opened 2021-11-01
Replace ArrayList.init/ensureTotalCapacity pairs with initCapacity

squeek502:
Because ArrayList.initCapacity uses 'precise' capacity allocation, this should save memory on average, and definitely will save memory in cases where ArrayList is used where a regular allocated slice could have also been used (i.e. in cases where the ArrayList is guaranteed to never grow past the initial capacity).

See https://github.com/ziglang/zig/issues/9775#issuecomment-956026454 for a bit more context

### DIFF
--- a/lib/std/unicode.zig
+++ b/lib/std/unicode.zig
@@ -551,10 +551,9 @@ fn testDecode(bytes: []const u8) !u21 {
 
 /// Caller must free returned memory.
 pub fn utf16leToUtf8Alloc(allocator: *mem.Allocator, utf16le: []const u16) ![]u8 {
-    var result = std.ArrayList(u8).init(allocator);
-    errdefer result.deinit();
     // optimistically guess that it will all be ascii.
-    try result.ensureTotalCapacity(utf16le.len);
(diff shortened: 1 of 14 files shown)


squeek502:
Specifically, this is the pattern of allocation when using an `init/ensureTotalCapacity` pair:

```
1...8 => 8
9...20 => 20
21...38 => 38
39...65 => 65
66...105 => 105
106...165 => 165
166...255 => 255
256...390 => 390
391...593 => 593
594...897 => 897
898...1353 => 1353
1354...2037 => 2037
2038...3063 => 4096
3064...6911 => 8192
6912...10374 => 12288
10375...15569 => 16384
15570...23361 => 24576
23362...35049 => 36864
35050...52581 => 53248
52582...78879 => 81920
78880...118326 => 118784
118327...177497 => 180224
177498...266253 => 270336
```

When using `initCapacity`, the resulting capacity is equal to the requested capacity until you hit ~1/2 page size (then the allocator starts giving back page size allocations):

```
2049...4096 => 4096
4097...8192 => 8192
8193...12288 => 12288
12289...16384 => 16384
```

PR #10069, opened 2021-10-31
stage2: optimize AstGen.callExpr

leecannon:
<details>
<summary>tldr</summary>

| type | average improvement |
| --- | --- |
| debug no cache | ~549ms |
| debug cached | ~1.7ms |
| release no cache |  ~196ms |
| release cached | ~0.1ms |

</details>

While working on #10068 I happened to notice a sizeable number of short term allocations in this function.

I collected data regarding the number of arguments in call expressions by compiling a file with only an empty main function with no local or global cache and stashing the data from `AstGen.callExpr` which resulted in:

| Length | Count | Percentage |
| --- | --- | --- |
| 1 | 19480 | 45.80% |
| 2 | 7962 | 18.72% |
| 0 | 7140 | 16.79% |
| 3 | 4152 | 9.76% |
| 4 | 2042 | 4.80% |
| 5 | 1036 | 2.44% |
| 6 | 332 | 0.78% |
| 7 | 213 | 0.50% |
| 10 | 158 | 0.37% |
| 11 | 5 | 0.01% |
| 9 | 5 | <0.01% |
| 8 | 3 | <0.01% |
| 12 | 2 | <0.01% |

Using this information to prevent allocations for call expressions with less than 5 arguments seems to show good improvement (at least for compiling a file with nothing but an empty main)

the tldr at the top contains a summary of the below information

> :warning: **the below no cache timings include the time to delete the cached files and folders**

## debug no cache
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `rm -rf zig-cache && rm -rf simple && rm -rf ~/.cache/zig && ~/src/zig/zig-out/changed/bin/zig build-exe simple.zig` | 2.364 ± 0.020 | 2.326 | 2.394 | 1.00 |
| `rm -rf zig-cache && rm -rf simple && rm -rf ~/.cache/zig && ~/src/zig/zig-out/normal/bin/zig build-exe simple.zig` | 2.913 ± 0.029 | 2.877 | 2.953 | 1.23 ± 0.02 |

## debug cached
| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `~/src/zig/zig-out/changed/bin/zig build-exe simple.zig` | 66.4 ± 4.7 | 55.1 | 76.7 | 1.00 |
[...]

PR #10079, opened 2021-11-01
stage2: Reduce temporary allocations in AstGen

mattbork:
This is a large set of changes to reduce temporary allocations in `AstGen.zig` that I've been sitting on for a few days because I wasn't sure they were worth the disruption. However, inspired by the improvement in #10069, I thought I should see what other people think. They start off simple but especially the last one may be too much of a burden to maintain for the minor gains in performance. I've tried to split the commits up to make it easier to pick and choose, but there's still a lot to wade through.

For the first commit, the main observation is that some functions that make temporary ArrayLists of `Zir.Inst.Ref` know the length of those lists ahead of time and do nothing with the lists other than copying them to `astgen.extra` as trailing payload data. The temporary allocations can be avoided by simply reserving space in `astgen.extra` (before any recurrence through `expr`) and building the payloads in place. This covers the `arrayInit*` and `structInit*` functions, `callExpr`, `typeOf`, and the `.compile_log` branch of `builtinCall`. The remaining change is in `errorSetDecl`. While it does not know its payload length ahead of time, it only collects the `string_bytes` indexes of identifiers and calls nothing that could touch `astgen.extra`. Thus instead of using a temporary list, it can also build its payload in place. The next commit just avoids always calling `identifierTokenString` from `identifier` when the result is only needed for error messages; currently, `identifier` always does two allocs for names using the `@""` syntax.

The third commit reworks `switchExpr` to collect all case payloads in one ArrayList rather than three. Because the different case types are now interleaved in this ArrayList, a small table of one `u32` per case is reserved at the front and then used to save indexes to each case in the list in payload order. At the end of `switchExpr`, this table is looped over to build the payload in `astgen.extra`. Because cases are now handled more uniformly, this also reduces a fair amount of repetition. One other thing worth mentioning is that the change to capture indices in #10038 was inadvertantly increasing `scalar_case_index` whenever the current case was not a multi-case. Thus, scalar cases with captures that came after an `else` or `_` case would have a `capture_index` that was one too large. These changes also fix that.

The fourth commit rewrites `WipDecls` as `WipMembers`, which now handles collecting fields as well as decls (and their respective bit sets) in one allocation, rather than the (up to) four ArrayLists used now. The next commit just simplifies the container functions by pulling out the common member processing into its own function. One thing to mention is that currently `containerDecl` always appends at least one u32 of field bits to its payload on the assumption that an empty enum has already been checked for. This misses the case of `enum(T) { _ }` which isn't empty but has no fields and will therefore end up with an extra trailing u32 in its payload. The fourth commit also fixes this.

The sixth commit introduces a scratch buffer shared throughout AstGen that both `switchExpr` and `WipMembers` use to avoid temporary allocations. There may be other functions that could also be reworked to use the scratch buffer, but now it's just these two.

The last two commits are the most dramatic change. The main observation is that while every `GenZir` has its own ArrayList for collecting instructions, use of those lists mostly follows the same strict nesting as the scopes themselves. For example, `fnDecl` makes the GenZirs `decl_gz`, `fn_gz`, `param_gz`, and `ret_gz` but the pattern of use looks like:

1. A new `param_gz` is made for each parameter and used for type expressions
2. A param instruction is added to `decl_gz` but only after we're done with `param_gz`
3. The alignment, calling convention, etc. are added to `decl_gz`
4. `ret_gz` is made for the return type but at this point nothing further will be added to `decl_gz`
5. `func_gz` is made for the function body and at this point nothing further will be added to `ret_gz`

Thus the GenZirs can share one `instructions` ArrayList like a stack so long as care is taken to save and restore the top of the stack at the beginning and end of the use of each GenZir.

In practice, there are a few more complications. The first is that the current handling of `closure_capture` instructions breaks this nesting because new instructions can be added to an enclosing block from abitrarily deep sub-blocks. Fortunately because only containers create Namespaces, their declaring_gz is always the GenZir passed to containerDecl, and containerDecl always adds exactly one instruction (an extended `*_decl`) to that GenZir. Thus, `closure_capture` instructions are always lined up immediately after a container decl instruction. Rather than adding them at the point of first mention, where we're nested arbitrarily deep, we can simply walk through the Namespace captures hash map at the end of each containerDecl branch and add them then. The seventh commit does this.

Other complications have to do with helper functions that may want to add instructions to GenZirs that are now stacked on top of each other. One example is `finishThenElseBlock`. Because the instructions for `else_scope` will be stacked on top of those for `then_scope`, a break cannot be added directly to `then_scope`. Instead, if a break is necessary, it is created without being added to any GenZir, then passed along to `setCondBrPayload` or the `*ElideBlockStorePtr` variant, where it is copied into the `astgen.extra` payload as appropriate. For symmetry, the break possibly added to `else_scope` is handled the same way. In a more complex situation where multiple instructions might be added to either scope, `finishThenElseBlock` could instead create additional GenZirs that to hold them and pass those along to `setCondBrPayload`. The necessary logic would only be a bit more complicated.

Overall, I think the changes are not terribly burdensome, but this does add additional subtlety to way AstGen functions must be written. (To this point, I also renamed GenZir's `addBlock` to `makeBlockInst` to emphasize that the block instruction is *not* being added to the GenZir's instruction list. Calls to it do not need to worry about GenZir stacking, only whatever later point where it actually gets added to a GenZir does.) There is an escape hatch, as my implementation of this idea gives each `GenZir` a pointer to the ArrayList into which it should collect instructions. This way if more complicated scenarios come up where GenZir instruction collecting is not strictly nested and interleaving is unavoidable, then a separate ArrayList can be created on the stack and a `GenZir` can take a pointer to that list instead. This is no worse than the present situation where every GenZir by default creates a new ArrayList.

For testing, I wrote a bit of scaffolding to do a deep comparison between the `Zir` returned from `generate` before and after these changes and ran this against every .zig file in this repo. With the exception of the two fixes mentioned above, they were identical, modulo the order of some payloads in `extra`. Then for all files > 5Kb, I used a simple counting allocator to count the alloc calls and the peak memory usage, and I also did a benchmark by averaging 150 runs of `generate` (after 50 warm-up runs) in a release build of the compiler. The full results are [attached here](https://github.com/ziglang/zig/files/7453784/results-5kb.txt).

The absolute differences in performance are small because AstGen generally takes only a few milliseconds, but in relative terms the speed-up is in the range of 10% to 15% generally. The number of calls to alloc is in many cases greatly reduced. For example, Sema.zig goes from 18586 allocs to 1454, parse.zig from 3986 to 240, tokenizer.zig from 1887 to 58, and AstGen.zig from 9917 to 874. The naive measure of peak memory use tends to be slightly higher and in a few instances significantly higher than before because once the shared `gz_instruction` or `scratch` ArrayLists grow to some size, that size is retained until AstGen is finished. This is the trade-off between fewer allocations and a finer-grained control over the (obstensible) amount of memory currently in use, and may be compensated for by less heap fragmentation, though that depends on the allocator certainly. The worst case is Sema.zig, which rises from 2,627,769 bytes to 3,039,394 or +15.66%, and target/arm.zig is next worst, going from 185,259 bytes to 213,827 or +15.42%. Most files have much smaller increases, in the range of 1% or 2%, but 5% is not too uncommon. If this became a problem, logic could be added to shrink either ArrayList when the unused capacity exceeds some threshold. For example, with Sema.zig, the extreme of calling `shrinkAndFree` after every use of `scratch` elimates ~385Kb of the ~400Kb increase at the cost of 168 additional calls to alloc, so some middle ground could be reached.

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -6029,16 +6029,15 @@ fn switchExpr(
     const cond_ty_inst = try parent_gz.addUnNode(.typeof, cond, operand_node);
     const item_rl: ResultLoc = .{ .ty = cond_ty_inst };
 
-    // These contain the data that goes into the `extra` array for the SwitchBlock/SwitchBlockMulti.
-    // This is the optional else prong body.
-    var special_case_payload = ArrayListUnmanaged(u32){};
-    defer special_case_payload.deinit(gpa);
-    // This is all the scalar cases.
-    var scalar_cases_payload = ArrayListUnmanaged(u32){};
-    defer scalar_cases_payload.deinit(gpa);
-    // Same deal, but this is only the `extra` data for the multi cases.
-    var multi_cases_payload = ArrayListUnmanaged(u32){};
-    defer multi_cases_payload.deinit(gpa);
+    // This contains the data that goes into the `extra` array for the SwitchBlock/SwitchBlockMulti,
+    // except the first cases_nodes.len slots are a table that indexes payloads later in the array, with
+    // the special case index coming first, then scalar_case_len indexes, then multi_cases_len indexes
+    var payloads = ArrayListUnmanaged(u32){};
+    defer payloads.deinit(gpa);
+    const scalar_case_table: u32 = @boolToInt(special_prong != .none);
+    const multi_case_table = scalar_case_table + scalar_cases_len;
+    const case_table_len = multi_case_table + multi_cases_len;
+    try payloads.resize(gpa, case_table_len);
 
     var block_scope = parent_gz.makeSubBlock(scope);
     block_scope.setBreakResultLoc(rl);
@@ -6069,13 +6068,6 @@ fn switchExpr(
 
         var capture_val_scope: Scope.LocalVal = undefined;
         const sub_scope = blk: {
-            const capture_index = if (is_multi_case) ci: {
-                multi_case_index += 1;
-                break :ci multi_case_index - 1;
-            } else ci: {
-                scalar_case_index += 1;
-                break :ci scalar_case_index - 1;
-            };
             const payload_token = case.payload_token orelse break :blk &case_scope.base;
             const ident = if (token_tags[payload_token] == .asterisk)
                 payload_token + 1
@@ -6109,6 +6101,7 @@ fn switchExpr(
                     0b10 => .switch_capture_multi,
                     0b11 => .switch_capture_multi_ref,
                 };
+                const capture_index = if (is_multi_case) multi_case_index else scalar_case_index;
                 break :capture try case_scope.add(.{
                     .tag = capture_tag,
                     .data = .{ .switch_capture = .{
@@ -6129,10 +6122,11 @@ fn switchExpr(
             break :blk &capture_val_scope.base;
         };
 
-        if (is_multi_case) {
-            // items_len, ranges_len, body_len
-            const header_index = multi_cases_payload.items.len;
-            try multi_cases_payload.resize(gpa, multi_cases_payload.items.len + 3);
+        const header_index = @intCast(u32, payloads.items.len);
+        const body_len_index = if (is_multi_case) blk: {
+            payloads.items[multi_case_table + multi_case_index] = header_index;
+            multi_case_index += 1;
+            try payloads.resize(gpa, header_index + 3); // items_len, ranges_len, body_len
 
             // items
             var items_len: u32 = 0;
@@ -6141,7 +6135,7 @@ fn switchExpr(
                 items_len += 1;
 
                 const item_inst = try comptimeExpr(parent_gz, scope, item_rl, item_node);
-                try multi_cases_payload.append(gpa, @enumToInt(item_inst));
+                try payloads.append(gpa, @enumToInt(item_inst));
             }
 
             // ranges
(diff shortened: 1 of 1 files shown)


src/AstGen.zig
```diff
@@ -6029,16 +6029,15 @@ fn switchExpr(
...
-    defer multi_cases_payload.deinit(gpa);
+    // This contains the data that goes into the `extra` array for the SwitchBlock/SwitchBlockMulti,
+    // except the first cases_nodes.len slots are a table that indexes payloads later in the array, with
+    // the special case index coming first, then scalar_case_len indexes, then multi_cases_len indexes
+    var payloads = ArrayListUnmanaged(u32){};
+    defer payloads.deinit(gpa);
+    const scalar_case_table: u32 = @boolToInt(special_prong != .none);
+    const multi_case_table = scalar_case_table + scalar_cases_len;
+    const case_table_len = multi_case_table + multi_cases_len;
+    try payloads.resize(gpa, case_table_len);
```
squeek502: This will usually over-allocate (see https://github.com/ziglang/zig/pull/10076 for more details). It would be better to call `initCapacity` instead of `resize` so that `resize` doesn't over-allocate (because `resize` will calculate a `better_capacity` larger than the requested capacity, while `initCapacity` won't).

```suggestion
    const scalar_case_table: u32 = @boolToInt(special_prong != .none);
    const multi_case_table = scalar_case_table + scalar_cases_len;
    const case_table_len = multi_case_table + multi_cases_len;
    var payloads = ArrayListUnmanaged(u32).initCapacity(gpa, case_table_len);
    defer payloads.deinit(gpa);
    payloads.items.len = case_table_len;
```

andrewrk:
Thanks for this! After perusing the changes, I agree with your conclusion that the changes from a maintainability perspective are reasonable.

Inspired by this, as well as yours and others' recent performance improvements, and the fact that we have new CI hardware available, I spent the last 2 days rebooting [gotta-go-fast](https://github.com/ziglang/gotta-go-fast/). There is now a [performance tracking web page](https://ziglang.org/perf/) and I'm in the process of hooking it up to be automatically populated by the new CI system.

I also took the time to add the 3 benchmarks at the bottom, which collect measurements for `zig ast-check lib/std/os.zig`, `zig ast-check src/AstGen.zig`, and `zig ast-check src/Sema.zig`, respectively. As a caveat they test running these commands as a child process which can be a bit noisy, and one improvement I have in mind is to change them to a single benchmark that covers more .zig files and does not include the child process overhead. Even so, I suspect we might see some interesting points in the graphs soon :-)

PR #10080, opened 2021-11-01
stage2 AArch64: implement unconditional branches

joachimschmidt557:
This PR introduces a new algorithm for lowering branches in MIR (which reference another MIR instruction as the target) to real branches (which include a code offset).


andrewrk:
Looks good :+1:

Issue #6573, opened 2020-10-05
llvm11: clang error when building c++ hello world on arm-linux-musleabi -mcpu=generic+v6kz (__libcpp_signed_lock_free)


FireFox317:
When building a simple c++ hello world with llvm11 branch (llvm 11.0.0-rc5) (`zig c++ example-cc.cpp -target arm-linux-musleabi -mcpu=generic+v6kz`):
```cpp
#include <iostream>
int main() {
std::cout << "Hello\n";
}
```
We get the following error:
```sh
In file included from /home/timonkr/zig-bootstrap/out/host/lib/zig/libcxx/include/memory:681:
/home/timonkr/zig-bootstrap/out/host/lib/zig/libcxx/include/atomic:2780:16: error: use of undeclared identifier '__libcpp_signed_lock_free'
typedef atomic<__libcpp_signed_lock_free> atomic_signed_lock_free;
               ^
/home/timonkr/zig-bootstrap/out/host/lib/zig/libcxx/include/atomic:2781:16: error: use of undeclared identifier '__libcpp_unsigned_lock_free'
typedef atomic<__libcpp_unsigned_lock_free> atomic_unsigned_lock_free;
```
Related: https://reviews.llvm.org/D75183

cc @LemonBoy

FireFox317:
Issue has been found: https://github.com/ziglang/zig/blob/7067764ed3f85eca17be7310b848ad97bd8af52e/lib/libcxx/include/atomic#L2764-L2781
Apparently on arm none of the `ATOMIC_{}_LOCK_FREE` are set to 2 thus there is no support for signed/unsigned lock-free types. But in that case `typedef atomic<__libcpp_signed_lock_free> atomic_signed_lock_free;` should not be defined. 

llvm-project trunk does not have a fix for this, so the following patch has to be applied:
```diff
diff --git a/zig/lib/libcxx/include/atomic b/zig/lib/libcxx/include/atomic
index 9c2898653..bf037726f 100644
--- a/zig/lib/libcxx/include/atomic
+++ b/zig/lib/libcxx/include/atomic
@@ -2775,10 +2775,15 @@ typedef conditional<_LIBCPP_CONTENTION_LOCK_FREE, __cxx_contention_t, char>::typ
 typedef conditional<_LIBCPP_CONTENTION_LOCK_FREE, __cxx_contention_t, unsigned char>::type      __libcpp_unsigned_lock_free;
 #else
     // No signed/unsigned lock-free types
[...]

PR #10083, opened 2021-11-01
Normalized C++ compilation options for single-threaded targets

nuald:
The PR contains two major changes (as two commits to drop the second one if needed):

1. C++ is compiled without threading support for single-threaded targets (was already partially implemented for WASM, now it's applied to all single-threaded targets).
2. RunStep has been extended to run artifacts with the external executors if enabled.

The reason for (1) is that the current functionality is a little bit confusing as C++ standard library is compiled with the threading support even if an artifact is explicitly flagged to use the single-threaded mode. I've discovered that during dealing with #6573. The original LLVM commit referenced there is being abandoned, and the only way to use C++ for ARM in Zig is using modified standard library, hence that PR.

There are few reason for (2), but none of them are critical, therefore the change could be dropped:

- `.enable_qemu` and similar flags could be applied to the executable, but the corresponding RunStep if invoked would give the `IncorrectExe` error. It's not immediately obvious that the RunStep is applicable for native targets only, and `.enable_qemu` is no-op for executables, but for tests only. In theory, those flags should be applied to tests and run steps only, but it's out of scope of this PR.
- some tests involve running the actual executables, e.g. the integration tests like the `c_compiler` modified in this PR. It's handy to have this ability out-of-box.
- some build processes require cross-platform runs. One particular example here is one of my project that has the pure Zig build process for LuaJIT: `buildvm` should be invoked on the target platform (please see https://github.com/LuaJIT/LuaJIT/issues/664 for the reference). I could successfully compile and use LuaJIT on ARM, but need to use the external executor for `buildvm`. The first commit uses `ExecutorRunStep` that is partially taken from the project.

Testing done by:
 - running the full test suite: `zig build test`
 - running the `c_compiler` test with 3 targets (x86_64 Linux host, arm-linux-musleabi, wasm32-wasi):

```sh
cd test/standalone/c_compiler/
zig build test
zig build test -Dtarget=arm-linux-musleabi -Denable-qemu
zig build test -Dtarget=wasm32-wasi -Denable-wasmtime
```

Additional notes:

- for WASM target the `-fno-exceptions` flag is applied to C++ standard library only, not the final executables. If could be addressed in this or another PR if you want;
- macos doesn't properly support C++ exceptions, additional work is required (probably libunwind and DWARF CFI should be added), and for now I've just disabled testing it; 
- `_LIBCPP_HAS_NO_THREADS` is added for `buildLibCXXABI`  for single-threaded targets too as it uses C++ standard library;
- `test.c/test.cpp` code have mixed spaces/tabs, I've restyled the code for the consistency;
- RunStep.getExternalExecutor is a little bit verbose, but I anticipate that other changes could be required so left it as is for now


nuald:
Messed up the commits, closing it and will reopen a new PR.

PR #10084, opened 2021-11-01
omit frame pointer logic bug

leecannon:
In #9874 I forgot to remove the old logic in `addCCArgs` regarding whether to omit the frame pointer or not. Currently both pieces of logic are in there.

Old logic:
https://github.com/ziglang/zig/blob/83dcfd62053a80024b1cb282fddf399a36c9c58e/src/Compilation.zig#L3389-L3393

New logic:
https://github.com/ziglang/zig/blob/83dcfd62053a80024b1cb282fddf399a36c9c58e/src/Compilation.zig#L3411-L3415

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -3386,12 +3386,6 @@ pub fn addCCArgs(
                 try argv.append("-mthumb");
             }
 
-            if (comp.haveFramePointer()) {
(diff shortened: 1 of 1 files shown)


andrewrk:
Thanks for the follow-up :+1:

PR #10088, opened 2021-11-01
std: add Linux perf syscall bits

andrewrk:
Example usage:

```zig
const std = @import("std");
const PERF = std.os.linux.PERF;

test "perf" {
    var attr: std.os.linux.perf_event_attr = .{
        .type = PERF.TYPE.HARDWARE,
        .config = @enumToInt(PERF.COUNT.HW.INSTRUCTIONS),
        .flags = .{
            .disabled = true,
            .exclude_kernel = true,
            .exclude_hv = true,
        },
    };
    const fd = try std.os.perf_event_open(&attr, 0, -1, -1, PERF.FLAG.FD_CLOEXEC);
    defer std.os.close(fd);

    _ = std.os.linux.ioctl(fd, PERF.EVENT_IOC.RESET, 0);
    _ = std.os.linux.ioctl(fd, PERF.EVENT_IOC.ENABLE, 0);

    long();

    _ = std.os.linux.ioctl(fd, PERF.EVENT_IOC.DISABLE, 0);

    var result: usize = 0;
    assert((try std.os.read(fd, std.mem.asBytes(&result))) == @sizeOf(usize));

    std.debug.print("instruction count: {d}\n", .{result});
}

fn long() void {
    var i: usize = 0;
    while (i < 100000) : (i += 1) {}
}
```

### DIFF
--- a/lib/std/c/darwin.zig
+++ b/lib/std/c/darwin.zig
@@ -1703,8 +1703,6 @@ pub const CLOCK = struct {
 /// Max open files per process
 /// https://opensource.apple.com/source/xnu/xnu-4903.221.2/bsd/sys/syslimits.h.auto.html
 pub const OPEN_MAX = 10240;
-pub const RUSAGE_SELF = 0;
-pub const RUSAGE_CHILDREN = -1;
 
 pub const rusage = extern struct {
     utime: timeval,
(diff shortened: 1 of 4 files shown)


andrewrk:
cc @ifreund I would submit this as an example of the OS bits namespacing working pretty well.

lib/std/c/darwin.zig
```diff
@@ -1723,6 +1721,9 @@ pub const rusage = extern struct {
     nsignals: isize,
     nvcsw: isize,
     nivcsw: isize,
+
+    pub const SELF = 0;
+    pub const CHILDREN = -1;
```
daurnimator, line 1726: This looks like it should be an enum:
```zig
enum(u32) {
    self = 0,
    children = -1,
    _,
}
```

Issue #10061, opened 2021-10-29
improve the C backend lowering of Zig types to C types


andrewrk:
There are two functions where we lower Zig types to C types:

https://github.com/ziglang/zig/blob/83a4bb6e6957325b744a08e16cab6b2a4a045f3e/src/codegen/c.zig#L727-L733

https://github.com/ziglang/zig/blob/83a4bb6e6957325b744a08e16cab6b2a4a045f3e/src/codegen/c.zig#L471

Right now it is hacky and not correct for more complex types. It needs to be improved to fully support C types, particularly regarding function pointers.

Thankfully, @Vexu already solved this problem in [arocc](https://github.com/Vexu/arocc) so we can borrow from this project:

https://github.com/Vexu/arocc/blob/361274f0e4b5a4b1f1cab029077ac968914fa255/src/Type.zig#L1285

Part of this issue is coming up with complex Zig types as test cases and making sure they properly lower to C types.

PR #10091, opened 2021-11-02
C backend:  Improve lowering of Zig types to C types

thomasives:
Working towards #10061.

I think when rendering Zig types as C types there are three different "kinds" of C types we render to:
1. primitive non-pointer C types (e.g. uint8_t).  We just print these out.
2. typedef'd types.  We also just print out the names of these when rendering a type, however, we need to make sure that we have a copy of the rendered typedef in the TypedefMap
3/ pointers -- this is the recursive case, as we have to first render the element type of the pointer

The only issue with this is that function pointers have weird syntax.  However, this can be avoided if we typedef function pointers.  Then they can be composed with pointers just like any other type, e.g. we can render "a pointer to a const pointer to a function pointer" as

```
typedef void (*func)(void);
func * const * f;
```
I think this approach should let us straightforwardly render any Zig type without too many special cases.

I want to take another pass through the code to tidy things up and look to add some tests, but I thought I would get some feedback to check I am on the right track and I haven't missed anything.  Thanks for your help in advance :)

So far I have:

1. Changed function pointers to be typedef'd so then we can treat them the same as other types.
2. Distinguished between const slices (zig_L prefix) and mut slices (zig_M prefix).  I think we need to do this if we want to avoid too much special casing.
3. Changed Zig "const pointers" so that they are lowered to C "pointers to const" (e.g. const char *) rather than C "const pointers" (e.g. char * const).  Please let me know if I have misunderstood Zig's semantics here.
4. Ensured that all typedefs are "linked" even if the decl doesn't have any forward declarations -- not sure if this is the correct thing to do

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -179,6 +179,14 @@ pub const DeclGen = struct {
         return error.AnalysisFail;
     }
 
+    fn getTypedefName(dg: *DeclGen, t: Type) ?[]const u8 {
+        if (dg.typedefs.get(t)) |some| {
+            return some.name;
+        } else {
+            return null;
+        }
+    }
+
     fn renderDeclValue(
         dg: *DeclGen,
         writer: anytype,
@@ -200,14 +208,20 @@ pub const DeclGen = struct {
             return;
         }
 
-        // Determine if we must pointer cast.
         assert(decl.has_tv);
-        if (ty.eql(decl.ty)) {
-            try writer.writeByte('&');
-        } else {
-            try writer.writeAll("(");
-            try dg.renderType(writer, ty);
-            try writer.writeAll(")&");
+        // We shouldn't cast C function pointers as this is UB (when you call
+        // them).  The analysis until now should ensure that the C function
+        // pointers are compatible.  If they are not, then there is a bug
+        // somewhere and we should let the C compiler tell us about it.
+        if (ty.castPtrToFn() == null) {
+            // Determine if we must pointer cast.
+            if (ty.eql(decl.ty)) {
+                try writer.writeByte('&');
+            } else {
+                try writer.writeAll("(");
+                try dg.renderType(writer, ty);
+                try writer.writeAll(")&");
+            }
         }
         try dg.renderDeclName(decl, writer);
     }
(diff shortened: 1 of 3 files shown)


thomasives:
I have realised that we also need to render the zig Fn type as a C function pointer.  I am not sure if we still need to special case a pointer to a function and render it as a C function pointer.  I think I am slightly confused about Zig's semantics here.

I am also not sure if a pointer to a function type delegate to the function type for rendering.  I suspect it probably should but for now I have left it creating a typedef with a different name to the same C function pointer type.

Even with the above issue, I am becoming reasonably confident that this strategy will allow us to render any Zig type comfortably.  I will work on adding some tests to demonstrate this.  I have already added a couple of easy ones which now pass with these changes

SpexGuy:
I think the way to lower a Zig function pointer is as a C struct containing a C function pointer.  This allows you to correctly create pointers to it.

thomasives:
I don't think lowering a Zig function pointer to a C struct helps.  You can take a pointer to a function pointer in C.  What you cannot do it take a pointer to the name of a function, because that name is just a label, there is no memory location in the program where its value is stored.   Typedefing a function pointer is enough, then it should behave just like any other type.

The issue I am having is that there is an inconsistency with how variable function pointers and struct field function pointers are represented in the AIR.   Take these two examples:

```
fn f3(x: bool) []const u8 {
    var wrapper: FnPtrWrapper = .{
        .fn_ptr = fB,
    };

    if (x) {
        wrapper.fn_ptr = fA;
    }

    return wrapper.fn_ptr();
}

const FnPtrWrapper = struct {
    fn_ptr: fn () []const u8,
};

fn f4(x: bool) []const u8 {
    var fn_ptr = fB;

    if (x) {
        fn_ptr = fA;
    }

    return fn_ptr();
}
```

In f3, we have a struct containing a function pointer and the AIR generated for the initial store  (`.fn_ptr = fB`) is:

```
# Begin Function AIR: f3:
  ...
  %13 = constant(*const fn() callconv(.Unspecified) []const u8, (decl ref))
  ...
  %5 = alloc(*tri_test.FnPtrWrapper)
  %7 = struct_field_ptr_index_0(**const fn() callconv(.Unspecified) []const u8, %5)
  %14!= store(%7!, %13!)
  ...
```
Here we are taking a pointer to `fn_ptr` field in %7, notice that the type here is a `**fn`.  In %14 we then store the function pointer and notice that the constant we are using has type `*fn`.

Now for f4, we have a variable function pointer and the AIR generated for the initial store (`var fn_ptr = fB`) is:

```
# Begin Function AIR: f4:
  ...
  %7 = constant(fn() callconv(.Unspecified) []const u8, (function 'fB'))
  ...
  %3 = alloc(*fn() callconv(.Unspecified) []const u8)
  %9 = bitcast(*fn() callconv(.Unspecified) []const u8, %3)
  %10!= store(%9!, %7!)
  ...
```

Here we are allocating the function pointer on the stack with %3, notice that the type of the variable is just `fn`, no pointer.  I believe the bitcast instruction should be interpreted as "take a reference to %3", notice here that the type of %9 is `*fn`, but this should be an actual pointer referring to a memory location where the variable %3 is store, not a function pointer that we can call.  In %10 we are storing to %3 via the reference in %9, and notice that the constant we are using now has type `fn`.

I don't understand how I can reconcile this.  In some instances I am being asked to treat `*fn`as a pointer to a memory location where a function pointer is stored.  In other instances I am being asked to treat `*fn` as a function pointer.   Can anyone help me understand what I am missing here?

Interestingly, the master branch fails compiling f4 because renderType had marked the .Fn case as unreachable.   There was a comment saying that .Fn corresponds to a function body and not a function pointer.  I wonder if the AIR being generated for f4 is incorrect and %3 should have type `**fn`.

thomasives:
I think I have been getting confused and I am trying to expand the scope of the ticket.  My `f4` example above doesn't compile on master and I don't think I can make it compile by changing how Zig types are lowered to C types.  So I have reverted to marking the `.Fn` tag as `unreachable` in `renderType`.   Now only `.Pointer` to `.Fn` are rendered as C function pointers.

Assuming the checks pass, I think this can now be merged.

andrewrk:
Thank you @thomasives and excellent work!

PR #10094, opened 2021-11-03
Add more entries to the list of stage1+stage2+CBE passing tests

jmc-88:
All of these seem to be passing with both `zig test -fLLVM ${test}` and `zig test -ofmt=c ${test}` on a stage2 build.


andrewrk:
Thanks!

PR #10096, opened 2021-11-04
add support for compiling Objective-C++ code

slimsag:
Prior to this change, calling `step.addCSourceFiles` with Obj-C++ file extensions (`.mm`) would result in an error due to Zig not being aware of that extension. Clang supports an `-ObjC++` compilation mode flag, but it was only possible to use if you violated standards and renamed your `.mm` Obj-C++ files to `.m` (Obj-C) to workaround Zig being unaware of the extension.

This change makes Zig aware of `.mm` files so they can be compiled, enabling compilation of projects such as [Google's Dawn WebGPU implementation](https://dawn.googlesource.com/dawn/) using only a `build.zig` file.

Helps hexops/mach#21

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -3307,13 +3307,17 @@ pub fn addCCArgs(
     try argv.appendSlice(&[_][]const u8{ "-target", llvm_triple });
 
     switch (ext) {
-        .c, .cpp, .m, .h => {
+        .c, .cpp, .m, .mm, .h => {
             try argv.appendSlice(&[_][]const u8{
                 "-nostdinc",
                 "-fno-spell-checking",
             });
             if (comp.bin_file.options.lto) {
                 try argv.append("-flto");
+            }
+
+            if (ext == .mm) {
+                try argv.append("-ObjC++");
             }
 
             // According to Rich Felker libc headers are supposed to go before C language headers.
@@ -3599,6 +3603,7 @@ pub const FileExt = enum {
     cpp,
     h,
     m,
+    mm,
     ll,
     bc,
     assembly,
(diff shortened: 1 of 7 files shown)


kubkon:
Wow, cool! Out of curiosity, does ObjC++ link fine with zld?

slimsag:
@kubkon Yep! So far so good, Dawn has ~5.5k lines of ObjC++ across ~20 files and it links and runs that just fine without complaint. I did run into #10103 but it's not related to ObjC++

kubkon (changes requested):
The changes look great. I just have one tiny request: would you be able to add some basic standalone tests that are enabled on macOS with SDK present only (we have a flag for that `-Denable-macos-sdk`)? The test would then link against some frameworks or what not, unless you think this is unnecessary.

slimsag:
I'd be very happy to do that! I dug around a bit and it seems like I would basically need to add a directory [like this one](https://github.com/ziglang/zig/tree/master/test/standalone/link_frameworks) ("standalone test") with something like a hello-world ObjC++ file called from Zig, plus maybe linking a framework in? Does that sound about right?

kubkon:
> I'd be very happy to do that! I dug around a bit and it seems like I would basically need to add a directory [like this one](https://github.com/ziglang/zig/tree/master/test/standalone/link_frameworks) ("standalone test") with something like a hello-world ObjC++ file called from Zig, plus maybe linking a framework in? Does that sound about right?

Yep, sounds right. And then in `standalone.zig` you can enable it only in the presence of macOS SDK (I don't think the test would make much sense on a different platform).

slimsag:
@kubkon PTAL - I rebased against `master` and added a standalone test (I based it on the existing `objc` standalone test, just adapted it for ObjC++)

kubkon:
> @kubkon PTAL - I rebased against `master` and added a standalone test (I based it on the existing `objc` standalone test, just adapted it for ObjC++)

Thanks @slimsag! Just an FYI that in terms of the failing macOS CI, that's my doing and I'll fix ASAP :-)

kubkon:
@slimsag if you rebase against master, macOS CI should be green again :-)

slimsag:
Done! 🤞

PR #10097, opened 2021-11-04
stage2: Implement Sema.floatToInt

zenith391:
This is my first attempt to contribute to Sema so I surely got a few things wrong.

As a quick overview of my changes:
- The code in `zirFloatToInt` is made by inspiring a lot from `zirIntToFloat`
- Both `zirFloatToInt` and `coerceNum` will handle the `error.FloatCannotFit` error by printing an error message similar to what stage1 printed.

`Value.floatToInt` is implemented by first checking if its NaN or Infinity (but I found that stage1 didn't check it and accepted compiling `@floatToInt(u64, std.math.inf_f64)`) making the float an absolute number and flooring the float.
Then, all it does is convert the floored float into a `std.math.big.Rational` which will perform necessary computations and reduce it. So that we can just take the `rational.p` field and we have our converted big int.

As an arbitrary milestone, this pull request allows compiling the `std.math.cos` and `std.math.sin` functions (alongside some other math functions) with the LLVM backend and the C backend (with my C backend float support pull request).

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -9659,7 +9659,7 @@ fn zirFloatToInt(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!
             error.FloatCannotFit => {
                 return sema.fail(block, operand_src, "integer value {d} cannot be stored in type '{}'", .{ std.math.floor(val.toFloat(f64)), dest_ty });
             },
-            else => |e| return e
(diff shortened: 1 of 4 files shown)


zenith391:
The floatToInt implementation passes the test ! More test coverage for stage2

andrewrk:
Nice work!

PR #10098, opened 2021-11-04
std: Fix path resolution on Windows

### LINKED ISSUE #10093: Zig compiler unable to open files in subst virtual drive on Windows 10
### Remember to search before filing a new report

- [X] I searched for this bug and did not find it in the issue tracker, and furthermore, the title I used above will make this new bug report turn up in the search results for my query.

### Zig Version

0.9.0-dev.1553+f951bf8ae

### Steps to Reproduce

Platform: Windows 10 AMD x64

1. Make a new folder
2. Make a new .zig file that contains something trivial (like a Hello World! program)
3. Subst the folder where the zig file is contained to a new virtual drive (in my case I `subst`ed it onto the X: drive)
4. Use the `zig run main.zig` command in a terminal
5. The compiler will do some work and then print out `unable to open 'X:\\main.zig': invalid filename`

If I pass the real path to the zig file, it works as expected. 

### Expected Behavior

The compiler should be able to work with files in any arbitrary drive, including Windows virtual drives created through the `subst` command

### Actual Behavior

Compiler is unable to open the zig file, printing out a `unable to open 'X:\\main.zig': invalid filename` error.

LemonBoy:
GetCurrentDirectory returns a path with a trailing slash iff the cwd is
a root directory, making the code in `resolveWindows` return an invalid
path with two consecutive slashes.

Closes #10093

### DIFF
--- a/src/stage1/os.cpp
+++ b/src/stage1/os.cpp
@@ -473,6 +473,11 @@ static Buf os_path_resolve_windows(Buf **paths_ptr, size_t paths_len) {
         result_disk_designator = parsed_cwd.disk_designator;
         if (parsed_cwd.kind == WindowsPathKindDrive) {
             result.ptr[0] = asciiUpper(result.ptr[0]);
+            // Remove the trailing slash if present, eg. if the cwd is a root
+            // directory.
+            if (buf_ends_with_mem(&cwd, "\\", 1)) {
+                result_index -= 1;
+            }
         }
         have_drive_kind = parsed_cwd.kind;
     }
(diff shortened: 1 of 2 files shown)


src/stage1/os.cpp
```diff
@@ -473,6 +473,11 @@ static Buf os_path_resolve_windows(Buf **paths_ptr, size_t paths_len) {
         result_disk_designator = parsed_cwd.disk_designator;
         if (parsed_cwd.kind == WindowsPathKindDrive) {
             result.ptr[0] = asciiUpper(result.ptr[0]);
+            // Remove the trailing slash if present, eg. if the cwd is a root
+            // directory.
+            if (buf_ends_with_mem(&cwd, "\\", 1)) {
```
Mouvedia, line 478: Can't we have `\\\` or more? If so replace the `if` with a `while`.

Issue #1923, opened 2019-02-06
when handling a returned error, if the expression/block does not try/return error, reset error return trace index to 0


Sahnvour:
When trying to run the first example code from #1813 with an executable that can't be found, the error trace is weird. It looks like the return address array is filled with addresses from `windowsCreateProcess`.
I believe this is due to the loop in `spawnWindows` that tries every directory from `PATH` and returns a lot of errors because the file doesn't exist.

As for why it is sometimes `PATH_NOT_FOUND` and others `FILE_NOT_FOUND`, I assume my `PATH` contains some old directories that don't exist anymore.

Example output:
```
$ zig run childprocess.zig
error: Unexpected
D:\DEV\Install\zig\lib\zig\std\os\child_process.zig:693:45: 0x7ff663ac4a97 in windowsCreateProcess (run.obj)
            windows.ERROR.PATH_NOT_FOUND => return error.FileNotFound,
                                            ^
D:\DEV\Install\zig\lib\zig\std\os\child_process.zig:692:45: 0x7ff663ac4a7a in windowsCreateProcess (run.obj)
            windows.ERROR.FILE_NOT_FOUND => return error.FileNotFound,
                                            ^
D:\DEV\Install\zig\lib\zig\std\os\child_process.zig:693:45: 0x7ff663ac4a97 in windowsCreateProcess (run.obj)
            windows.ERROR.PATH_NOT_FOUND => return error.FileNotFound,
                                            ^
D:\DEV\Install\zig\lib\zig\std\os\child_process.zig:693:45: 0x7ff663ac4a97 in windowsCreateProcess (run.obj)
            windows.ERROR.PATH_NOT_FOUND => return error.FileNotFound,
                                            ^
D:\DEV\Install\zig\lib\zig\std\os\child_process.zig:692:45: 0x7ff663ac4a7a in windowsCreateProcess (run.obj)
            windows.ERROR.FILE_NOT_FOUND => return error.FileNotFound,
                                            ^
D:\DEV\Install\zig\lib\zig\std\os\child_process.zig:692:45: 0x7ff663ac4a7a in windowsCreateProcess (run.obj)
            windows.ERROR.FILE_NOT_FOUND => return error.FileNotFound,
                                            ^
D:\DEV\Install\zig\lib\zig\std\os\child_process.zig:692:45: 0x7ff663ac4a7a in windowsCreateProcess (run.obj)
            windows.ERROR.FILE_NOT_FOUND => return error.FileNotFound,
[...]

PR #10102, opened 2021-11-05
C backend: while, struct tests, better undefined global handling

rainbowbismuth:
Three changes to `codegen/c.zig`:

1. Function signatures that return a no member struct return `void`
2. Undefined var decls don't get a value generated for them
3. Don't generate bitcast code if the result isn't used, since bitcast is a pure function. Right now struct handling code generates some weird unused bitcast AIR, and this optimization side steps that issue.

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -448,7 +448,12 @@ pub const DeclGen = struct {
                 try w.writeAll("ZIG_COLD ");
             }
         }
-        try dg.renderType(w, dg.decl.ty.fnReturnType());
+        const return_ty = dg.decl.ty.fnReturnType();
+        if (return_ty.hasCodeGenBits()) {
+            try dg.renderType(w, return_ty);
+        } else {
+            try w.writeAll("void");
+        }
         try w.writeAll(" ");
         try dg.renderDeclName(dg.decl, w);
         try w.writeAll("(");
@@ -946,6 +951,10 @@ pub fn genDecl(o: *Object) !void {
             try o.dg.renderDeclName(o.dg.decl, fwd_decl_writer);
         }
         try fwd_decl_writer.writeAll(";\n");
+
+        if (variable.init.isUndef()) {
+            return;
+        }
 
         try o.indent_writer.insertNewline();
         const w = o.writer();
@@ -1886,6 +1895,9 @@ fn airBr(f: *Function, inst: Air.Inst.Index) !CValue {
 }
 
 fn airBitcast(f: *Function, inst: Air.Inst.Index) !CValue {
+    if (f.liveness.isUnused(inst))
+        return CValue.none;
+
     const ty_op = f.air.instructions.items(.data)[inst].ty_op;
     const operand = try f.resolveInst(ty_op.operand);
 
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -451,6 +451,8 @@ pub const DeclGen = struct {
         const return_ty = dg.decl.ty.fnReturnType();
         if (return_ty.hasCodeGenBits()) {
             try dg.renderType(w, return_ty);
+        } else if (return_ty.zigTypeTag() == .NoReturn) {
+            try w.writeAll("zig_noreturn void");
         } else {
             try w.writeAll("void");
         }
--- a/test/behavior.zig
+++ b/test/behavior.zig
@@ -21,12 +21,15 @@ test {
     _ = @import("behavior/hasdecl.zig");
     _ = @import("behavior/hasfield.zig");
     _ = @import("behavior/if.zig");
+    _ = @import("behavior/struct.zig");
+    _ = @import("behavior/truncate.zig");
     _ = @import("behavior/null.zig");
     _ = @import("behavior/ptrcast.zig");
     _ = @import("behavior/pub_enum.zig");
     _ = @import("behavior/truncate.zig");
     _ = @import("behavior/underscore.zig");
     _ = @import("behavior/usingnamespace.zig");
+    _ = @import("behavior/while.zig");
 
     if (builtin.object_format != .c) {
         // Tests that pass for stage1 and stage2 but not the C backend.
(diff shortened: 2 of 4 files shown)


rainbowbismuth:
I'm a bit surprised by these failures this morning. Don't have time to investigate all of it until the evening, but it looks like out of memory errors?

My `../zig-stage2/bin/zig test test/behavior.zig -I test -ofmt=c -femit-bin=test.c` and `build/zig test -I test test/behavior.zig` pass for me on my machine. 

Let me know if there is other testing I should be doing locally for changes like this.

rainbowbismuth:
Oh and also, I figure I should explain what I mean by weird AIR. "Weird" as in weird to me, because I haven't entirely grokked the IR yet, or looked much into AstGen.

Here's the relevant zig source:

```zig
const Val = struct {
    x: i32,
};

test "struct initializer" {
    const val = Val{ .x = 42 };
    try expect(val.x == 42);
}
```

```
# Begin Function AIR: test.struct initializer:
# Total AIR+Liveness bytes: 730B
# AIR Instructions:         42 (378B)
# AIR Extra Data:           30 (120B)
# AIR Values Bytes:         10 (80B)
# Liveness tomb_bits:       24B
# Liveness Extra Data:      2 (8B)
# Liveness special table:   1 (8B)
  %1 = const_ty((inferred_alloc_const))
  %3 = const_ty(*const type)
  %4 = constant(*const type, (decl ref))
  %5 = constant(type, behavior.struct.Val)
  %6 = const_ty(behavior.struct.Val)
  %8 = const_ty(*behavior.struct.Val)
  %10 = const_ty(*i32)
  %12 = constant(comptime_int, 42)
  %13 = constant(i32, 42)
  %17 = const_ty(*const fn(bool) callconv(.Unspecified) (inferred error set of expect)!void)
  %18 = constant(*const fn(bool) callconv(.Unspecified) (inferred error set of expect)!void, (decl ref))
  %19 = const_ty(fn(bool) callconv(.Unspecified) (inferred error set of expect)!void)
  %20 = constant(fn(bool) callconv(.Unspecified) (inferred error set of expect)!void, (function 'expect'))
  %21 = const_ty(behavior.struct.Val)
  %24 = constant(comptime_int, 42)
  %25 = const_ty(i32)
  %26 = constant(i32, 42)
  %32 = const_ty((inferred error set of expect))
  %35 = const_ty((inferred error set of test.struct initializer)!void)
  %39 = const_ty((inferred error set of test.struct initializer)!void)
  %40 = constant((inferred error set of test.struct initializer)!void, (eu_payload) {})

  %0!= dbg_stmt(45:5)
  %2 = alloc(*behavior.struct.Val)
###########################################
# This line right here, %7!, why would it be trying to
# bitcast a void value into this struct?
###########################################
  %7!= bitcast(behavior.struct.Val, @Ref.void_value)
  %9 = bitcast(*behavior.struct.Val, %2)
  %11 = struct_field_ptr_index_0(*i32, %9!)
  %14!= store(%11!, %13!)
  %15!= dbg_stmt(46:5)
  %16!= block(void, {
    %22 = load(behavior.struct.Val, %2!)
    %23 = struct_field_val(%22!, 0)
    %27 = cmp_eq(%23!, %26!)
    %28 = call(%20!, [%27!])
    %29 = is_non_err(%28)
    %38!= cond_br(%29!, {
      %30 = unwrap_errunion_payload(void, %28!)
      %31!= br(%16, %30!)
    }, {
      %33 = unwrap_errunion_err((inferred error set of expect), %28!)
      %34 = bitcast(anyerror, %33!)
      %36 = wrap_errunion_err((inferred error set of test.struct initializer)!void, %34!)
      %37!= ret(%36!)
    })
  })
  %41!= ret(%40!)
# End Function AIR: test.struct initializer
```

`codegen/c.zig` wasn't haven't a good time trying to convert the `@Ref.void_value` ref into an actual value to render with the `memcpy`, of course.

If this AIR is truly unintended and not extremely obvious, I'd love to start digging into AstGen to try and solve it myself, if someone could give me some pointers.

rainbowbismuth:
I looked into one of the failed builds, and saw that it runs the `build/zig build test-toolchain -Denable-macos-sdk` command. I went ahead and ran that locally since I'm on Mac OS, and it failed in the exact same way. 

I was able to make that build succeed by restoring the handling of `.NoReturn` that I broke in function signatures. I'm not sure how that lead to the out of memory errors though.

squeek502:
The OutOfMemory errors are almost certainly not real, they show up due to a combination of https://github.com/ziglang/zig/issues/1923 and `Allocator.reallocBytes`

Luukdegram:
The previous error was actually in 
```
src/Module.zig:3061:13: 0x5a2b4f in Module.ensureDeclAnalyzed (test)
            return error.AnalysisFail;
```
Unfortunately, Stage1 stack traces contain a lot of incorrect `error.OutOfMemory` you can mostly ignore, as the error is often further down the trace.

I can also really recommend looking at the new [Glossery](https://github.com/ziglang/zig/wiki/Glossary) to get yourself familiar with the pipeline and the terminology. (Should at least help you figure out that AIR is produced by Sema, rather than AstGen :)).

rainbowbismuth:
Oh, it is `Sema.zig`! I got the files mixed up when searching around for `Ref.void_value` to see if I could spot the issue :) since there are references in both places. 

Thanks to both of you for the additional context on that error.

andrewrk:
Thanks for this!

After merging, I noticed the number of behavior tests jumped up from 1060 to 1062 and discovered that two tests were duplicated across struct_llvm.zig and struct.zig. Please be very careful when moving behavior tests around! We don't want to inadvertently lose test coverage. Fixed in c77698d69ea5c3e09b1b9404f634e512627ad321.

> This line right here, `%7!`, why would it be trying to
> bitcast a void value into this struct?

That indeed looks sus. Here are some steps to understand it:

1. Look at the ZIR that is generated for that test case. You can get this with `zig ast-check -t foo.zig`. This is the output of AstGen.zig and the input to Sema.zig.
2. Look at the AIR (you have already done this step and highlighted the questionable AIR). This is the output of Sema.zig and the input to Codegen.
3. Find the corresponding logic in Sema that is producing that bitcast instruction, and try to understand why it is happening. Start with that giant switch expression near the top of Sema.zig; it has a similar structure to Codegen where it switches on the tag and the functions are named `zirFoo` after the tags.

rainbowbismuth:
Thanks for the fix Andrew, I'll make sure to be more careful in the future, and check the number of tests run. And I'll see what I can do and investigate the strange AIR.

andrewrk:
Ah forgot to mention yesterday, I happened to notice the issue when tackling another problem, and it is fixed in cb785b9c6ba705fcb507fdfa470e27b5e82ce1c5. The bitcast was a dummy value having to do with inferred types. Now it emits to a temporary block instead of the current block and then just gets thrown away.

PR #10105, opened 2021-11-05
Fix for Issue #8051: std.Thread.Condition: error: no member named 'unlock' in struct 'std.Thread.Mutex'

geo-ant:
Hi,
this is a potential fix for [Issue 8051](https://github.com/ziglang/zig/issues/8051). It is a very low tech solution, but I did add a test case (which is also pretty low tech).

The test case will succeed if signalling the condition variable unblocks the waiting thread. However, it would also (often) succeed when waiting on a condition variable has no effect. I'd be interested to get a hint on how I can test that waiting on the condition variable does really block the waiting thread.

There is also [PR 9437](https://github.com/ziglang/zig/pull/9437) which tackles the same issue but hasn't been active for a while.

Cheers
Geo

### DIFF
--- a/lib/std/Thread/Condition.zig
+++ b/lib/std/Thread/Condition.zig
@@ -194,61 +194,97 @@ pub const AtomicCondition = struct {
     }
 };
 
-// verify that the condition variable unblocks when signalled
-test "AtomicCondition" {
-    if (!builtin.single_threaded) {
-        var wait_thread_alive = std.atomic.Atomic(bool).init(false);
-        var wait_thread_finished = std.atomic.Atomic(bool).init(false);
-        var run_condition = std.atomic.Atomic(bool).init(false);
-        var condvar = Condition{};
-        const test_thread = try std.Thread.spawn(.{}, conditionWaitThread, .{ &wait_thread_alive, &wait_thread_finished, &run_condition, &condvar });
-        test_thread.detach();
+const testing = std.testing;
 
-        // we give the waiting thread generous time to become alive, but
-        // in case it does not, we fail here rather than hang indefinitely
-        try waitUntilTrue(&wait_thread_alive, 10);
+test "Condition - wait/signal" {
+    if (builtin.single_threaded) return error.SkipZigTest;
 
-        // this does not really tell us much, but we might as well check it
-        try std.testing.expect(!wait_thread_finished.load(.SeqCst));
+    const Context = struct {
+        lock: Mutex = .{},
+        cond: Condition = .{},
+        signaled: bool = false,
 
-        run_condition.store(true, .SeqCst);
-        condvar.signal();
+        fn doWait(self: *@This()) void {
+            const held = self.lock.acquire();
+            defer held.release();
 
-        // similar as above we let the thread indicate it is finished or fail
-        try waitUntilTrue(&wait_thread_finished, 10);
+            while (!self.signaled) {
+                self.cond.wait(&self.lock);
+            }
+        }
+
+        fn doSignal(self: *@This(), do_broadcast: bool) void {
+            const held = self.lock.acquire();
+            defer held.release();
+
+            self.signaled = true;
+            switch (do_broadcast) {
+                true => self.cond.signal(),
+                else => self.cond.broadcast(),
+            }
+        }
+    };
+
+    for ([_]bool{ false, true }) |do_broadcast| {
+        var context = Context{};
+        const wait_signal = try std.Thread.spawn(.{}, Context.doWait, .{&context});
+
+        context.doSignal(do_broadcast);
+        wait_signal.join();
     }
 }
 
-/// a primitive helper method that blocks until an atomic boolean becomes true
-/// if the bool does not become true within the given amount of max seconds, this function failes
-fn waitUntilTrue(boolean: *std.atomic.Atomic(bool), max_wait_time_seconds: u32) !void {
-    var current_time: std.os.timespec = undefined;
-    try std.os.clock_gettime(std.os.CLOCK.REALTIME, &current_time);
-    const start_time = current_time;
-    while (!boolean.load(.SeqCst)) {
-        std.os.nanosleep(1, 250 * 1000);
-        try std.os.clock_gettime(std.os.CLOCK.REALTIME, &current_time);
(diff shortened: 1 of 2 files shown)


lib/std/Thread/Condition.zig
```diff
@@ -152,9 +152,9 @@ pub const AtomicCondition = struct {
             @atomicStore(bool, &cond.pending, true, .SeqCst);
         }
 
-        mutex.unlock();
+        mutex.releaseDirect();
         waiter.data.wait();
-        mutex.lock();
+        _ = mutex.acquire();
```
matu3ba, line 157: same argument as for line 155.

matu3ba (commented):
The change from `mutex.lock()` -> `mutex.releaseDirect()` should be documented. Check, if `RWLock.zig` got removed from `Mutex.zig` and does the right thing for all edge cases.

matu3ba:
> The test case will succeed if signalling the condition variable unblocks the waiting thread. However, it would also (often) succeed when waiting on a condition variable has no effect. I'd be interested to get a hint on how I can test that waiting on the condition variable does really block the waiting thread.

Can you clarify what you mean with
```zig
    while (!run_condition.load(.SeqCst)) {
        condvar.wait(&mutex);
```
has no effect?

> how I can test that waiting on the condition variable does really block the waiting thread.

[It is not easy/portable](https://stackoverflow.com/a/4952672/9306292). 

One approach, if you can disable preemption/rely on no scheduler not preemting: get the system time before and after the section against a minimal architecture-specific constant.
(A debugger is essentially a scheduler that can stop execution to trace output.)

Another approach would be to "invoke a debugger" (doesnt work in CI though) or the approach mentioned on stackoverflow.

lib/std/Thread/Condition.zig
geo-ant: My reasoning here was this: since we expect a locked mutex to be passed into the `wait` method of the condition variable, we already have an instance of `Held` for that mutex in the calling scope. I thought to change the interface for `wait` to return the `Held` instance, but that felt confusing to me, because we already have an instance in the calling scope associated with the mutex. Very likely we have code like this:
```
// the mutex exists somewhere, maybe outside this scope
const held = mutex.acquire();
defer held.release();
while(! some_condition) {
  condvar.wait(&mutex);
}
```
Since the mutex is guaranteed to be locked when `wait` returns, it is semantically correct to release the mutex using the `held` instance. At least I think so.

Now if I make `wait` return the `Held` instance, I would have two instances pointing to the same mutex, which is confusing at best. Let me know if that makes sense.

lib/std/Thread/Condition.zig
```diff
@@ -152,9 +152,9 @@ pub const AtomicCondition = struct {
             @atomicStore(bool, &cond.pending, true, .SeqCst);
         }
 
-        mutex.unlock();
+        mutex.releaseDirect();
```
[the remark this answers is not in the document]
geo-ant: Hi, thanks for taking the time to review the PR. My reasoning was simple (maybe simplistic): since no instance of `Held` is available to me, I just used the `Mutex` API. Would it make sense to extend the interface of `wait` to include the `Held` instance? This would feel repetitive and error prone, but there's probably something I am not seeing.

geo-ant:
> Can you clarify what you mean with
> 
> ```zig
>     while (!run_condition.load(.SeqCst)) {
>         condvar.wait(&mutex);
> ```
> 
> has no effect?

Yeah, my wording was not very precise. What I meant is that my test would succeed even if `condvar.wait(...)` was a no-op. This is independent of the surrounding while loop, which I just put there against spurious wake-ups.

> > how I can test that waiting on the condition variable does really block the waiting thread.
> 
> [It is not easy/portable](https://stackoverflow.com/a/4952672/9306292).
> 
> One approach, if you can disable preemption/rely on no scheduler not preemting: get the system time before and after the section against a minimal architecture-specific constant. (A debugger is essentially a scheduler that can stop execution and read state.)
> 
> Another approach would be to "invoke a debugger" (doesnt work in CI though) or the approach mentioned on stackoverflow.

Argh, I feared that would be trouble. Do I understand correctly that none of the approaches is portable or work on CI? Can you suggest a way forward?

lib/std/Thread/Condition.zig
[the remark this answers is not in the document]
matu3ba: > Would it make sense to extend the interface of wait to include the Held instance?

`Held` is defined in `Mutex.zig`, but not needed as we dont expose it to the user. I asked about `unlock`, but I checked the git log and nothing shows up so it looks like it was broken from the commit.

lib/std/Thread/Condition.zig
matu3ba: > Let me know if that makes sense.

yes. looks good. The overall code in `Thread` looks good, but there is no nice overview how stuff works for good code navigation. So it took me a while to figure things out.

matu3ba:
@geo-ant From kprotty:
```txt
t1:
  lock()
  defer unlock()
  cond = false
  while !cond:
    wait(&lock)

t2:
  lock()
  defer unlock()
  cond = true
  notify()
```
and [tests in his rework PR](https://github.com/ziglang/zig/blob/d4377c0c2c9dd35d92b925ca8544257348cca921/lib/std/Thread/Condition.zig).

This were be taken from either #8084 or #7800 (a rewrite was planned to be done eventually, but tests are always valuable).

geo-ant:
> @geo-ant From kprotty:
> 
> ```
> t1:
>   lock()
>   defer unlock()
>   cond = false
>   while !cond:
>     wait(&lock)
> 
> t2:
>   lock()
>   defer unlock()
>   cond = true
>   notify()
> ```
> 
> and [tests in his rework PR](https://github.com/ziglang/zig/blob/d4377c0c2c9dd35d92b925ca8544257348cca921/lib/std/Thread/Condition.zig).
> 
> These were be taken from either #8084 or #7800 (a rewrite was planned to be done eventually, but tests are always valuable).
> 
> You may also want to glance at #9136.

Thanks, I'll follow up on this. I like kprotty's tests more than mine and I'll try to get them working for me, without the implementation changes that protty made to the condition variable.

PR #10107, opened 2021-11-05
Implement `chmod` and `chown`

ominitay:
I've noticed that std's filesystem API lacks `chmod` and `chown` functionality, so I thought I'd add this myself. I haven't written much cross-platform code in std before, so please make sure I've done this right, and any feedback is appreciated! This should be ready to merge once I write some unit tests.

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -254,6 +254,87 @@ pub fn close(fd: fd_t) void {
     }
 }
 
+pub const FChmodError = error{
+    AccessDenied,
+    InputOutput,
+    SymLinkLoop,
+    FileNotFound,
+    SystemResources,
+    ReadOnlyFileSystem,
+} || UnexpectedError;
+
+/// Changes the mode of the file referred to by the file descriptor.
+/// The process must have the correct privileges in order to do this
+/// successfully, or must have the effective user ID matching the owner
+/// of the file.
+pub fn fchmod(fd: fd_t, mode: mode_t) FChmodError!void {
+    if (builtin.os.tag == .windows or builtin.os.tag == .wasi)
+        @compileError("Unsupported OS");
+
+    while (true) {
+        const res = system.fchmod(fd, mode);
+
+        switch (system.getErrno(res)) {
+            .SUCCESS => return,
+            .INTR => continue,
+            .BADF => unreachable, // Can be reached if the fd refers to a directory opened without `OpenDirOptions{ .iterate = true }`
+
+            .FAULT => unreachable,
+            .INVAL => unreachable,
+            .ACCES => return error.AccessDenied,
+            .IO => return error.InputOutput,
+            .LOOP => return error.SymLinkLoop,
+            .NOENT => return error.FileNotFound,
+            .NOMEM => return error.SystemResources,
+            .NOTDIR => return error.FileNotFound,
+            .PERM => return error.AccessDenied,
+            .ROFS => return error.ReadOnlyFileSystem,
+            else => |err| return unexpectedErrno(err),
+        }
+    }
+}
+
+pub const FChownError = error{
+    AccessDenied,
+    InputOutput,
+    SymLinkLoop,
+    FileNotFound,
+    SystemResources,
+    ReadOnlyFileSystem,
+} || UnexpectedError;
+
+/// Changes the owner and group of the file referred to by the file descriptor.
+/// The process must have the correct privileges in order to do this
+/// successfully. The group may be changed by the owner of the directory to
+/// any group of which the owner is a member. If the owner or group is
+/// specified as `null`, the ID is not changed.
+pub fn fchown(fd: fd_t, owner: ?uid_t, group: ?gid_t) FChownError!void {
+    if (builtin.os.tag == .windows or builtin.os.tag == .wasi)
+        @compileError("Unsupported OS");
+
+    while (true) {
+        const res = system.fchown(fd, owner orelse @as(u32, 0) -% 1, group orelse @as(u32, 0) -% 1);
+
+        switch (system.getErrno(res)) {
+            .SUCCESS => return,
+            .INTR => continue,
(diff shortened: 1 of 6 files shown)


lib/std/os.zig
```diff
@@ -254,6 +254,101 @@ pub fn close(fd: fd_t) void {
...
+        };
+
+        switch (res) {
+            .SUCCESS => return,
+            .INTR => continue,
+
+            .FAULT => unreachable,
+            .INVAL => unreachable,
+            .ACCES => return error.AccessDenied,
+            .BADF => return error.InvalidFileDescriptor,
```
marler8997, line 291: I believe this should be `unreachable` and same thing with fchmod.

lib/std/os.zig
```diff
@@ -254,6 +254,101 @@ pub fn close(fd: fd_t) void {
     }
 }
 
+pub const FChmodError = error{
+    AccessDenied,
+    InvalidFileDescriptor,
```
marler8997, line 259: I believe this error should should be removed and same thing with FChownError.

lib/std/os.zig
ominitay: Which should be unreachable? Access denied should surely be an error to be handled, and I've seen other code in `os.zig` use a similar pattern for `BADF`.

lib/std/os.zig
ominitay: The rest of `os.zig` does the same as this

lib/std/os.zig
```diff
@@ -254,6 +254,101 @@ pub fn close(fd: fd_t) void {
...
+} || UnexpectedError;
+
+/// Changes the mode of the file referred to by the file descriptor.
+/// The process must have the correct privileges in order to do this
+/// successfully, or must have the effective user ID matching the owner
+/// of the file.
+pub fn fchmod(fd: fd_t, mode: mode_t) FChmodError!void {
+    while (true) {
+        const res = if (builtin.link_libc) blk: {
+            const rc = std.c.fchmod(fd, mode);
```
LemonBoy, line 274: Can't this and the cases below be folded into a single one reaching into `system`?

LemonBoy:
No tests, no go.

ominitay:
> No tests, no go.

I did say that myself lol

lib/std/os.zig
ominitay: Oh yep, I can do that now.

lib/std/os.zig
marler8997: Talking about `BADF`.  Almost everything uses unreachable.  There may be times where it's not used and I think those are times when `BADF` is overloaded to mean multiple things (i.e. not just a bad file descriptor).

lib/std/os.zig
marler8997: I think it's the opposite.  Most don't have this error.  I think there's 1 case that does.  May have been a mistake that got past review.

ominitay: In this case though, `BADF` can be raised if a directory has been opened without the `.iterate` option. That isn't an API violation, right?

lib/std/os.zig
ominitay: Let's talk about `BADF` in just one conversation.

andrewrk:
I'm sure the comment was as much directed at me, as I can be trigger happy with the merge button sometimes :-)

ominitay:
Right, I've got a test for `chmod` going, but I don't see a feasible way to test `chown`, since it would involve relying upon the system configuration, and/or requiring root permissions. What are your thoughts?

kubkon:
I haven't looked in-depth into the changes in this PR, but have you considered Windows and WASI in the mix here? At the very least, we should throw a compile error for these two platforms if you don't intend to implement the proposed functionality for those platforms.

ominitay:
@kubkon Yep, I'm intending to do so, will get that pushed in a few minutes. I don't think the chmod or chown concepts exist on Windows or WASI, but I could be wrong.

ominitay:
Have added null options for `chown` to keep the ID the same, have added some tests for `chown` which cover pretty much everything we conceivably can. Also have improved documentation a bit. I'm pretty much happy with the PR now.

lib/std/os.zig
marler8997: It looks like that is an API violation:
https://github.com/ziglang/zig/blob/0fb26b03690eb327164bbc7a21ba2d69d3b1a5b8/lib/std/fs.zig#L577

ominitay: Yeah, maybe so. It feels wrong that that should be an unreachable, but if that's what the rest of std does then I'll change this once I see whether these CI checks are passing. @andrewrk What are your thoughts on that?

lib/std/os.zig
```diff
@@ -254,6 +254,85 @@ pub fn close(fd: fd_t) void {
...
+        @compileError("Unsupported OS");
+
+    while (true) {
+        const res = system.fchmod(fd, mode);
+
+        switch (system.getErrno(res)) {
+            .SUCCESS => return,
+            .INTR => continue,
+
+            .FAULT => unreachable,
```
marler8997, line 281: Rather than remove `BADF` and fall down to the `unexpectedErrno` case, I believe the convention is to handle known error codes with `unreachable`.  These 2 ways of handling the error have different affects.  In the latter case, it would prompt developers (like myself) to think that the std library is missing an error code and go in and add it.

ominitay: Oh, that was a complete mistake on my part, apologies for that. I'll fix that now 🤦

ominitay:
@marler8997 @LemonBoy I can't re-request reviews, but is this all okay to be merged now?

ominitay:
That CI failure in Azure isn't to do with this PR, not sure why it failed like that, but all other CI tests are passing.

marler8997:
I don't see any more issues.

ominitay:
Rebased onto master, and added a note to help people out if their code reached `.BADF`

ominitay:
@andrewrk Is there anything blocking this from being merged?

PR #10109, opened 2021-11-06
Fix Typo in docs

Validark:


ominitay:
That's not a typo...

leecannon:
To clarify in `\r\n?  |  [\n\u0085\u2028\u2029]` (whitespace for clarity), the first term `\r\n?` matches windows `\r\n` and also old mac `\r` and the second term matches `\n`as well as various unicode line/paragraph seperators.

With your chage the old mac way of using use a `\r` is not matched.

Validark:
My mistake, I didn't realize old Macs used just an `/r`.

PR #10116, opened 2021-11-07
Cmake: Specify LLVM versions

martinhath:
Systems with multiple LLVM toolchains installed (e.g. one globally and one in `$HOME/local`) would get confused and fail to compile.  Being explicit about the version required will force CMake to find the right version of LLVM.

Not sure this is the intended way of doing this (I try my best to avoid CMake).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,9 +104,9 @@ else()
     set(ZIG_USE_LLVM_CONFIG OFF CACHE BOOL "use llvm-config to find LLVM libraries")
 endif()
 
-find_package(llvm)
-find_package(clang)
-find_package(lld)
+find_package(llvm 13)
+find_package(clang 13)
+find_package(lld 13)
 
 if(ZIG_STATIC_ZLIB)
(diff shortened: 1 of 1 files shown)


Jan200101:
Does this work?

find_package(llvm) loads in Findllvm.cmake which looks for LLVM on hardcoded paths so I don't understand how it would find an LLVM installation in `$HOME/local`

martinhath:
> Does this work?
> 
> find_package(llvm) loads in Findllvm.cmake which looks for LLVM on hardcoded paths so I don't understand how it would find an LLVM installation in `$HOME/local`

Not automatically, you've also got to supply the path to the local installation in `-DCMAKE_PREFIX_PATH`, [as written in the wiki](https://github.com/ziglang/zig/wiki/Building-Zig-From-Source). The problem I had was that when doing so CMake tried to use the wrong toolchain, which in my case was the system wide, which was a different version. Nothing in the CMake file signals that we actually require LLVM 13, hence this PR. 

But again, I'm not sure this is how it's supposed to be done.

andrewrk:
Thanks!

Issue #8940, opened 2021-05-30
stop invoking lld when building one object file


mattnite:
I have the following zig code:
```rs
export fn foo() c_int {
    return 0;
}
```

If I try to build it with `zig build-obj probe.zig -target bpfel-freestanding -OReleaseFast` I get:
```
error(link): /home/mknight/zig/0.8.0-dev.2707+322215bf0/files/zig terminated with stderr:

error: LLDCrashed
```

mattnite:
I'm sorry for the lack of info on this one, I gave up trying to build LLVM12 locally and I've just been waiting for it to be released on arch before working on the compiler again, so I don't know how to debug this.

LemonBoy:
[lld is getting](https://reviews.llvm.org/D101336) support for BPF as we speak.
The real problem here is that Zig should really stop invoking lld when building object files.

andrewrk:
Keeping the issue open to track these TODOs:

https://github.com/ziglang/zig/blob/a2c546fea30de2caf1efb454d327e70270c07338/src/link/Elf.zig#L1385-L1391

andrewrk:
Also as @LemonBoy pointed out in https://github.com/ziglang/zig/pull/10117#issuecomment-962662345 the `zig cc` case needs to be solved as well.

PR #10117, opened 2021-11-07
Don't invoke linker when just building an object

mattnite:
This is an attempt to address #8940, before this patch if you try to make a BPF object you get the following error:
```
ld.lld: error: internal linker error: cannot read addend for relocation R_BPF_64_NODYLD32
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace.


ld.lld: error: probe:(.debug_info+0x6): relocation R_BPF_64_ABS32 out of range: 276824064 is not in [-524288, 524287]; consider recompiling with -fdebug-types-section to reduce size of debug sections
ld.lld: error: /home/mknight/code/zig-bpf-intro/zig-cache/o/5b6131127baab3cca09863748393a523/probe.o:(.debug_str): offset is outside the section
```
And now it's working as expected. In the past LLD was just crashing and I think we're getting more information due to improvements that came with llvm13. I think the specifics of this linking error is a red herring because we shouldn't be invoking the linker at all.

I've made this change apply to all linking formats but I'm not sure if that's correct.

### DIFF
--- a/src/link.zig
+++ b/src/link.zig
@@ -474,6 +474,10 @@ pub const File = struct {
             try fs.cwd().copyFile(cached_pp_file_path, fs.cwd(), full_out_path, .{});
             return;
         }
+
+        if (base.options.output_mode == .Obj)
+            return;
+
         const use_lld = build_options.have_llvm and base.options.use_lld;
         if (use_lld and base.options.output_mode == .Lib and base.options.link_mode == .Static) {
             return base.linkAsArchive(comp);


LemonBoy:
The change is a little more invasive than this as it should also work for the `zig cc` case, check the `if` branch before the one you've just added.

kubkon:
Hmm, this won't work in its current form as stage2 linker is expected to be invoked when building relocatables. First of, as far as I understand, the patch is for stage1 and ELF, correct? If so, it should be put inside the `linkWithLLD` function and not `flush`. Secondly, there is already a bit that handles building objects inside the linker (stage1): [Elf.zig#L1384](https://github.com/ziglang/zig/blob/0fb26b03690eb327164bbc7a21ba2d69d3b1a5b8/src/link/Elf.zig#L1384). Thirdly, shouldn't the change perhaps be handled higher up the toolchain, either in `main.zig` or `Compilation.zig`?

Is the repro from #8940 still valid? I can have a look tomorrow or after Handmade and assist in making this change but I'll need some guidance at reproing the exact issue.

PR #10119, opened 2021-11-08
std: ppoll: cast number of fds to nfds_t

gpanders:
On some systems, the type of the length of a slice is different from the `nfds_t` type, so cast the slice length to `nfds_t`. This is already done in poll, so just copy that implementation for ppoll.


andrewrk:
Thanks!

PR #10122, opened 2021-11-09
Treat x86_64 tests as native under the Rosetta 2 on M1 Macs

kubkon:


lib/std/zig/cross_target.zig
```diff
@@ -626,6 +626,11 @@ pub const CrossTarget = struct {
                 return .native;
             }
         }
+        // If the OS match and OS is macOS and CPU is arm64, treat always as native
+        // since we'll be running the foreign architecture tests using Rosetta2.
+        if (os_match and os_tag == .macos and builtin.cpu.arch == .aarch64) {
```
LemonBoy, line 631: Missing a `cpu_arch == .x86_64` ?

andrewrk:
I intentionally don't have rosetta installed on my M1, so for me this would start breaking things. Could this perhaps be treated like qemu, wine, wasmtime, etc, with a `-Denable-rosetta` flag?

lib/std/zig/cross_target.zig
kubkon: No, I left it out intentionally. Rosetta is only available on arm64 macOS hosts.

kubkon:
> I intentionally don't have rosetta installed on my M1, so for me this would start breaking things. Could this perhaps be treated like qemu, wine, wasmtime, etc, with a `-Denable-rosetta` flag?
> 
> Edit: ah yes this is more than just a compiler detail; this API applies to `zig build` in general, so I do think we should care about this use case.

Mhm, and Rosetta for you is not available, correct? Lemme figure out if there's a way to check for it on macOS.

lib/std/zig/cross_target.zig
LemonBoy: I really doubt Rosetta is able to run binaries other than those targeting x86...

kubkon: You're right. What I'm getting at is we don't officially support anything but x86_64 and aarch64 for macOS.

kubkon: I've added a preemptive check below, if you could have a look.

kubkon:
@andrewrk I've added a check which according to Apple docs (unless I misunderstood something) should tell us if Rosetta is installed on the host. If you could test it out on your M1, that'd be awesome!

andrewrk:
I got this error:

```
andy@MacBook-Pro build % ./zig build test
Test [1/71] test ""... compiler [723/791] darwin hello world with updates [3/7] update [3/3]
update_index=2 The following command failed with Unexpected:
./test_case
compiler [723/791] darwin hello world with updates [4/7] test 'darwin hello world with updates' failed: ChildProcessExecution

compiler [724/791] corner case - update existing, singular TextBlock [1/2] updat...
update_index=0 The following command failed with Unexpected:
./test_case
compiler [724/791] corner case - update existing, singular TextBlock [2/2] test 'corner case - update existing, singular TextBlock' failed: ChildProcessExecution

compiler [769/791] 2 tests failed
Test [2/71] test ""... FAIL (TestFailed)
```

kubkon:
> I got this error:
> 
> ```
> andy@MacBook-Pro build % ./zig build test
> Test [1/71] test ""... compiler [723/791] darwin hello world with updates [3/7] update [3/3]
> update_index=2 The following command failed with Unexpected:
> ./test_case
> compiler [723/791] darwin hello world with updates [4/7] test 'darwin hello world with updates' failed: ChildProcessExecution
> 
> compiler [724/791] corner case - update existing, singular TextBlock [1/2] updat...
> update_index=0 The following command failed with Unexpected:
> ./test_case
> compiler [724/791] corner case - update existing, singular TextBlock [2/2] test 'corner case - update existing, singular TextBlock' failed: ChildProcessExecution
> 
> compiler [769/791] 2 tests failed
> Test [2/71] test ""... FAIL (TestFailed)
> ```

I think I've had the error conditions mismatched which I've corrected for now. Could you try now please?

andrewrk:
```
andy@Andrews-MBP build % ./zig build test
Test [1/71] test ""... compiler [725/793] darwin hello world with updates [3/7] update [3/3]
update_index=2 The following command failed with Unexpected:
./test_case
compiler [725/793] darwin hello world with updates [4/7] test 'darwin hello world with updates' failed: ChildProcessExecution

compiler [726/793] corner case - update existing, singular TextBlock [1/2] updat...
update_index=0 The following command failed with Unexpected:
./test_case
compiler [726/793] corner case - update existing, singular TextBlock [2/2] test 'corner case - update existing, singular TextBlock' failed: ChildProcessExecution

compiler [771/793] 2 tests failed
Test [2/71] test ""... FAIL (TestFailed)
```

kubkon:
@andrewrk I've now fixed the merge conflicts, and redone the PR to follow the declarative mechanism we already use for QEMU and Wine as we agreed offline. We thus don't actively check if Rosetta is available; instead we rely on the user to declaratively say it is, and we then proceed assuming that it is indeed installed. This intent can be specified  (unsurprisingly) via the `-Denable-rosetta` flag.

lib/std/build.zig
```diff
@@ -2530,6 +2533,9 @@ pub const LibExeObjStep = struct {
             }
         } else switch (self.target.getExternalExecutor()) {
             .native, .unavailable => {},
+            .rosetta => if (self.enable_rosetta) {
+                try zig_args.append("--test-cmd-bin");
```
andrewrk, line 2537: Wouldn't this just be the same as `native`, i.e. no args appended?

kubkon: Yeah, I was wondering about that one, and I thought it was the case, but then note that `.native` and `.unavailable` share the prong which suggests they won't be run. In fact, I think in this case we could easily say here `.native => unreachable` since I guess it should have been handled in the first `if` block above?

kubkon: Hmm, no, scratch that, I think you're right. I think we can simply have `.native, .unavailable, .rosetta => {},`.

kubkon: OK, I just verified that:

```
$ zig test main.zig -target x86_64-macos
warning: created src/zig-cache/o/493cf4b056dc7e9721da37e880cfb846/test but skipping execution because it is non-native
```

while:

```
$ zig test main.zig -target x86_64-macos --test-cmd-bin
All 1 tests passed.
```

So I think we do want to check if Rosetta is enabled declaratively and then append `--test-cmd-bin` if so. Thoughts?

andrewrk:
Tests passed for me locally on my m1 which does not have rosetta installed.

PR #10125, opened 2021-11-09
macho: use start.zig for macOS entrypoint

kubkon:
This effectively allows us to compile

```zig
pub fn main() void {}
```

which then calls into `std.start`.

Changes required to make this happen:
* handle signed int to immediate in x86_64 and aarch64 codegen
* ensure that on arm64 macOS, `.x19` is a caller-preserved register - I'm not sure about that one at all and would like to brainstorm it with anyone interested and especially @joachimschmidt557. Without this hack, a call to `libSystem` after leaving `main2` in [start.zig#L86](https://github.com/ziglang/zig/blob/b24e4757176b0f0b110b6f726d92bbfb820924e3/lib/std/start.zig#L86) would cause a segfault as it expected `.x19` to be somehow preserved. The thing though is, I cannot find any reference to `.x19` being reserved or caller preserved in Apple docs.
* finally, fix a bug in the linker - mark new got entry as dirty upon atom growth.

### DIFF
--- a/src/arch/aarch64/CodeGen.zig
+++ b/src/arch/aarch64/CodeGen.zig
@@ -2386,6 +2386,9 @@ fn genTypedValue(self: *Self, typed_value: TypedValue) InnerError!MCValue {
         },
         .Int => {
             const info = typed_value.ty.intInfo(self.target.*);
+            if (info.bits <= ptr_bits and info.signedness == .signed) {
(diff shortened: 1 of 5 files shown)


andrewrk (approved):
:rocket:

joachimschmidt557 (approved):
Looks good to me!

PR #10128, opened 2021-11-10
io_uring: add poll_update

lithdew:
Add method from liburing to queue (but not submit) a SQE to update the user data of an existing poll operation.

cc @ifreund @jorangreef

### DIFF
--- a/lib/std/os/linux/io_uring.zig
+++ b/lib/std/os/linux/io_uring.zig
@@ -634,6 +634,22 @@ pub const IO_Uring = struct {
         return sqe;
     }
 
+    /// Queues (but does not submit) an SQE to update the user data of an existing poll
+    /// operation. Returns a pointer to the SQE.
+    pub fn poll_update(
+        self: *IO_Uring,
+        user_data: u64,
+        old_user_data: u64,
+        new_user_data: u64,
+        poll_mask: u32,
+        flags: u32,
+    ) !*io_uring_sqe {
+        const sqe = try self.get_sqe();
+        io_uring_prep_poll_update(sqe, old_user_data, new_user_data, poll_mask, flags);
+        sqe.user_data = user_data;
+        return sqe;
+    }
+
     /// Queues (but does not submit) an SQE to perform an `fallocate(2)`.
     /// Returns a pointer to the SQE.
     pub fn fallocate(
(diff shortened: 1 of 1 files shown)


lib/std/os/linux/io_uring.zig
```diff
@@ -634,6 +634,15 @@ pub const IO_Uring = struct {
         return sqe;
     }
 
+    /// Queues (but does not submit) an SQE to update the user data of an existing poll
+    /// operation. Returns a pointer to the SQE.
+    pub fn poll_update(self: *IO_Uring, user_data: u64, old_user_data: u64, new_user_data: u64, poll_mask: u32, flags: u32) !*io_uring_sqe {
```
jorangreef, line 639:
```suggestion
    pub fn poll_update(self: *IO_Uring, user_data: u64, old_user_data: u64, new_user_data: u64, poll_mask: u32, flags: u32,) !*io_uring_sqe {
```

lib/std/os/linux/io_uring.zig
```diff
@@ -1109,6 +1118,17 @@ pub fn io_uring_prep_poll_remove(
...
 
+pub fn io_uring_prep_poll_update(
+    sqe: *io_uring_sqe,
+    old_user_data: u64,
+    new_user_data: u64,
+    poll_mask: u32,
+    flags: u32,
+) void {
+    io_uring_prep_rw(.POLL_REMOVE, sqe, -1, old_user_data, flags, new_user_data);
+    sqe.rw_flags = std.mem.nativeToLittle(u32, poll_mask);
```
jorangreef, line 1129: What do you think of wrapping this in `io_uring_prep_poll_mask()` as liburing does, so that we can reuse for `io_uring_poll_add()` above, but more importantly also add a comment on why we do this, i.e. because poll flags were once-upon-a-time defined in the flags union as 16-bit but subsequently resized to 32-bit?

References:
https://github.com/axboe/liburing/blob/master/src/include/liburing.h#L402
https://github.com/axboe/liburing/blob/master/src/include/liburing.h#L367-L373
https://www.spinics.net/lists/io-uring/msg02848.html

lithdew: Sounds good to me. Wrapped the endian conversion and added a small doc note explaining why the conversion is needed.

jorangreef: Nice, looks good.

PR #10130, opened 2021-11-10
x86_64/Emit: don't misisel push/pop for extended regs

g-w1:
cc @kubkon


kubkon (approved):
Good catch, thanks @g-w1!

Issue #8334, opened 2021-03-22
Allow coercion of comptime_float to integer type


Tetralux:
You cannot currently do something like:
```zig
var i: u64 = 44_000_000_000;
var n = i / 1e9; // error: float value 1000000000.000000 cannot be coerced to type 'u64'
assert(n == 44);
```

`comptime_int`s are currently coerced to `comptime_float` if there is no precision loss.

`comptime_float`s could also coerce to `comptime_int` if there is no precision loss, as is the case in this example.

Vexu:
An implicit cast from `comptime_float` to `comptime_int` cannot happen in divisions because the intention is ambiguous as explained in #2568.

matu3ba:
@Tetralux What performance difference does make computing things with floats and in ints in comptime?
Here is [what things are checked in python](https://stackoverflow.com/questions/21232733/why-is-float-faster-than-int).

Tetralux:
@Vexu There's doesn't seem to be any ambiguity here, as the compile error already suggests.

@matu3ba This issue is just about what type the literal should become, and less about what you are doing with it - kinda similar to peer type resolution with ternary if expressions, perhaps.

Vexu:
> There's doesn't seem to be any ambiguity here

In this case there isn't any since the remainder is zero. To apply this to all cases the compiler would need to check the remainder in the division which would add some complexity, but I guess that could be acceptable.

Tetralux:
As it turns out, this example already works:
```zig
const divisor: u64 = 1e9; // already coerces to u64
var n = i / divisor; // integer division
```
So this coercion is presumably already present, at least in that case.

---
Exactly what should happen regarding the logic from #2568 is not entirely clear to me, though.

One option is that peer type resolution kicks in, and decides to promote the comptime type to the concrete type. (`comptime_float` to `u64`)

The concrete type could be used a type hint, since:
1) comptime types normally coerce to any concrete type it fits into.
2) the concrete type is what would determine what operation actually happens here.
3) if you wanted floating-point division, they would have casted `i` to a float at some point; either in the division expression, or before it; thus they wanted integer division.

Counterpoint: `i / @as(u64, 1e9)` - though that does require naming the type of thing you already have, which seems a little redundant.

andrewrk:
Working as designed, because it's ambiguous whether the result type should be comptime_float or comptime_int.

PR #10133, opened 2021-11-11
std.math: implement generic float constant derivation functions, cleanup `fabs` / `isNormal`, fix rare panic

viriuwu:
## @ Maintainers: Summary of PR here https://github.com/ziglang/zig/pull/10133#issuecomment-1000925183

The behaviour of `isNormal` is to add `1 << mantissa_bits` to the float reinterpreted as an unsigned integer (which would add 1 to the exponent), remove the sign bit, and check if it's less than `1 << (mantissa_bits+1)` which is only possible if the exponent is all zeroes (subnormal) turning into `...00001` or all ones (inf, nan) and overflowed into the sign bit which was then removed either way. This is a really nice hack to simplify the function, however, what happens during the first bit if adding `1 << mantissa_bits` would overflow the integer the float was reinterpreted as? Well...

```zig
const std = @import("std");
const math = std.math;

pub fn main() void {
    const bad = @bitCast(f32, @as(u32, math.maxInt(u32)));
    std.debug.print("Oh no: {}\n", .{math.isNormal(bad)});
}
```

```
zig build-exe isn.zig && ./isn
thread 56137 panic: integer overflow
/usr/lib/zig/std/math/isnormal.zig:21:26: 0x23122b in std.math.isnormal.isNormal (isn)
            return (bits + (1 << 23)) & (maxInt(u32) >> 1) >= (1 << 24);
                         ^
/home/vi/repos/scrap/isn.zig:6:51: 0x229f84 in main (isn)
    std.debug.print("Oh no: {}\n", .{math.isNormal(bad)});
                                                  ^
...
```

The simple way to fix this would be to remove the sign bit, add `1 << mantissa_bits`, remove the sign bit again, and compare against `1 << (mantissa_bits+1)`, however, that's not entirely necessary as letting it overflow would achieve the same thing, and in fact LLVM optimises it out entirely, so `+%` is much more clear here than `((x & mask) + constant) & mask`. The function also previously had a really weird mix of derived and hardcoded constants for each type of float and it wasn't necessary at all so it's been nicely simplified, along with tests for this weird edge-case.


viriuwu:
Looks like `std.math.fabs` has the same sort of weird mix of hardcoded and semi-derived constants and it's near identical to `isNormal` internals so I'll put a cleanup for that on this PR.

lib/std/math/fabs.zig
```diff
@@ -16,86 +15,28 @@ const maxInt = std.math.maxInt;
...
+    if (@typeInfo(T) != .Float) {
+        @compileError("fabs not implemented for " ++ @typeName(T));
+    }
 
-fn fabs128(x: f128) f128 {
-    var u = @bitCast(u128, x);
-    u &= maxInt(u128) >> 1;
-    return @bitCast(f128, u);
+    const remove_sign = @as(IntT, math.maxInt(IntT)) >> 1;
+    return @bitCast(T, @bitCast(IntT, x) & remove_sign);
```
matu3ba, line 24:
```zig
const uintT = @bitCast(IntT, x);
return @bitCast(T, uintT & remove_sign);
```
is more readable as mixing 2 comptime executions with 1 runtime `&` in a nested scope is not less trivial to read.

lib/std/math/fabs.zig
```diff
@@ -16,86 +15,28 @@ const maxInt = std.math.maxInt;
...
-    u &= 0x7FFFFFFF;
-    return @bitCast(f32, u);
-}
-
-fn fabs64(x: f64) f64 {
-    var u = @bitCast(u64, x);
-    u &= maxInt(u64) >> 1;
-    return @bitCast(f64, u);
-}
+    const IntT = std.meta.Int(.unsigned, @bitSizeOf(T));
```
matu3ba, line 18: If its unsigned, it should be something along `UintT` or `UT` to have consistency with `u32,i32` etc.

lib/std/math/fabs.zig
```diff
@@ -16,86 +15,28 @@ const maxInt = std.math.maxInt;
...
-
-test "math.fabs64" {
-    try expect(fabs64(1.0) == 1.0);
-    try expect(fabs64(-1.0) == 1.0);
-}
-
-test "math.fabs128" {
-    try expect(fabs128(1.0) == 1.0);
-    try expect(fabs128(-1.0) == 1.0);
-}
```
matu3ba, line 77: you did not convert these to comptime and did not make an argument why.

lib/std/math/isnormal.zig
```diff
@@ -1,32 +1,20 @@
...
-        },
-        else => {
-            @compileError("isNormal not implemented for " ++ @typeName(T));
-        },
+
+    if (@typeInfo(T) != .Float) {
+        @compileError("isNormal not implemented for " ++ @typeName(T));
     }
+
+    const IntT = std.meta.Int(.unsigned, @bitSizeOf(T));
```
matu3ba, line 13: Again `UT` or `UintT` for readability.

lib/std/math/isnormal.zig
```diff
@@ -1,32 +1,20 @@
...
+    const overflow = 1 << math.floatMantissaBits(T);
+    const remove_sign = math.maxInt(IntT) >> 1;
+
+    return (@bitCast(IntT, x) +% overflow) & remove_sign >= (overflow << 1);
```
matu3ba, line 17: Declutter this abit. Also add a brief and comprehensive comment why the lhs works.
You mentioned `hack`, so readers can not validate its correctness/follow easily.
```zig
// your cool explanation
var lhs = (@bitCast(IntT, x) +% overflow) & remove_sign;
return lhs >= (overflow << 1);
```

matu3ba (changes requested):
1. some tests were removed and there is no explanation why
2. `IntT` for an unsigned int is very confusing.
3. the hack should have proper (likely abit more verbose) comments.
4. (personally view) mixing comptime and runtime and multiple nestings is something I find confusing. More than 3 nestings of statements or expressions also.

lib/std/math/fabs.zig
viriuwu: These were removed because those individual functions do not exist anymore, the function returns `@TypeOf(x)` and doesn't need to be tested for a `switch (@TypeOf(x))` working now. Unless I misunderstand what you're asking.

lib/std/math/fabs.zig
viriuwu: I used `IntT` to be consistent with the rest of the standard library, ex.

https://github.com/ziglang/zig/blob/d29871977f97b50fe5e3f16cd9c68ebeba02a562/lib/std/fmt/parse_hex_float.zig#L12-L15

I think specifying that it's for bit hacks is better though, as I don't need to use `maxInt` to get all ones, so I'll go with with `TBits` to mean a raw bit interpretation of the float, which is what it is.

lib/std/math/fabs.zig
viriuwu: Done on both fabs and isNormal.

lib/std/math/isnormal.zig
```diff
@@ -1,45 +1,43 @@
 const std = @import("../std.zig");
 const math = std.math;
 const expect = std.testing.expect;
-const maxInt = std.math.maxInt;
 
 // Returns whether x has a normalized representation (i.e. integer part of mantissa is 1).
```
matu3ba, line 5: This should have a better explanation or at least mention the standard paper ["What Every Computer Scientist Should Know About Floating-Point Arithmetic" by David Goldberg](https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html#936).

If you feel fancy add the [most relevant definition bits](https://www.volkerschatz.com/science/float.html) to math.zig that are missing:`MANT_DIG, MIN_EXP and MAX_EXP`.

```txt
EPSILON | 0.0625 | Difference between 1 and next larger representable value
-- | -- | --
MIN | 0.25 | Smallest normalised representable value
MAX | 15.5 | Largest positive representable value
MANT_DIG | 5 | Mantissa bit size including implied leading 1
MIN_EXP | -2 | Minimum exponent for normalised number; also implied for subnormals
MAX_EXP | 3 | Maximum exponent
TRUE_MIN | 0.015625 | Smallest positive representable value
```

matu3ba (commented):
Looks good now.
Would be nice, if you can add some comments on the overall methods and a few more test cases.

If you want to go the long road:
1. add the 3 missing most important `Floating-point type property constants` or make a TODO for it in math.zig
2. compute some values at the edge of being subnormal and not.

viriuwu:
Sorry for the delay. I've rebased the commits again and addressed all the additional tests by adding the smallest and biggest subnormals, +-0.0, +-Inf, and NaN, improved the docstring to make a little more sense if you understand basic float categories, which is what the function really is for, categorising.

As for the constants, `MANT_DIG` can already be derived from `std.math.floatMantissaBits`, but since the other mathematical constants are being added, it makes sense to add this one as well, especially since upon a bit of research pretty much every low-level language provides these. `MIN_EXP` and `MAX_EXP` on the other hand are absolutely needed to simplify stuff like this:

https://github.com/ziglang/zig/blob/d29871977f97b50fe5e3f16cd9c68ebeba02a562/lib/std/fmt/parse_hex_float.zig#L22-L24

https://github.com/ziglang/zig/blob/b26b72f54051b89e6fc349f453aafe9f3a3135a5/lib/std/math/scalbn.zig#L25-L27

... so I put that on this PR, along with switching it in those functions.

Should be good to merge now.

matu3ba:
cc @tiehuis @LewisGaul 
What do you think about the test coverage and the PR? I did not implement floating stuff yet.

viriuwu:
Hey, good to see some activity, I was actually going to update this PR today or tomorrow, with:

1) I notice that `scalbn` got moved to `ldexp` so that's a rebase
2) The whole thing with `isNormal` is present here too:
https://github.com/ziglang/zig/blob/3bd0575cfe685d52be79e1757c883c5f7e26ac0d/lib/std/math/isinf.zig#L6-L30
3) The new `floatMantissaDigits` **is** actually already usable in that very same function to better document the otherwise unexplained +1:
https://github.com/ziglang/zig/blob/3bd0575cfe685d52be79e1757c883c5f7e26ac0d/lib/std/math/ldexp.zig#L31

Anyways I'll wait for the @'ed people.

tiehuis:
The current PR looks good to me, both the method used and the tests. Feel free to perform those other changes in another PR as well. The scope for improving the C translated code is pretty wide.

lib/std/math/isnormal.zig
```diff
@@ -1,45 +1,48 @@
...
-        },
+    const TBits = std.meta.Int(.unsigned, @bitSizeOf(T));
+    if (@typeInfo(T) != .Float) {
+        @compileError("isNormal not implemented for " ++ @typeName(T));
     }
+
+    const increment_exp = 1 << math.floatMantissaBits(T);
+    const remove_sign = ~@as(TBits, 0) >> 1;
+
+    // Hack: 1 is added to the exponent, and if it overflows to 0 or becomes 1,
```
LewisGaul, line 16: Should this really be considered a "hack"? This kind of bit manipulation seems to be quite common in the implementation of maths functions. Just wouldn't want future contributors to see this and think it's something that needs improving, if we're happy with this approach.

lib/std/math/isnormal.zig
```diff
@@ -1,45 +1,48 @@
...
-    try expect(!isNormal(math.nan(f128)));
-    try expect(!isNormal(@as(f16, 0)));
-    try expect(!isNormal(@as(f32, 0)));
-    try expect(!isNormal(@as(f64, 0)));
-    try expect(!isNormal(@as(f128, 0)));
-    try expect(isNormal(@as(f16, 1.0)));
-    try expect(isNormal(@as(f32, 1.0)));
-    try expect(isNormal(@as(f64, 1.0)));
-    try expect(isNormal(@as(f128, 1.0)));
+    inline for ([_]type{ f16, f32, f64, f128 }) |T| {
```
LewisGaul, line 27: What's the situation with `c_longdouble`? I *think* this implementation should support it (since it's just represented as `Float{.bits=80}` in the compiler), but there's limited support for them in general in the (stage1) compiler such that the testcase won't pass? Should a note/TODO be added here to say `c_longdouble` should be added to the list of types being tested?

lib/std/math/isnormal.zig
```diff
@@ -1,45 +1,48 @@
...
+        const TBits = std.meta.Int(.unsigned, @bitSizeOf(T));
+        const mantissa_bits = math.floatMantissaBits(T);
+
+        // normals
+        try expect(isNormal(@as(T, 1.0)));
```
LewisGaul, line 32: Maybe add the max and min normals, e.g. `f128_max` and `f128_min` (add a function that takes the float type in `math.zig`?).

lib/std/math/isnormal.zig
```diff
@@ -1,45 +1,48 @@
...
+
+        // subnormals
+        try expect(!isNormal(@as(T, -0.0)));
+        try expect(!isNormal(@as(T, 0.0)));
+        try expect(!isNormal(@bitCast(T, @as(TBits, 1))));
```
LewisGaul, line 37: This is just the 'true min', e.g. `f128_true_min` - maybe preferable to use the variable name/function to make this clearer? `floatTrueMin(T)`

lib/std/math/isnormal.zig
```diff
@@ -1,45 +1,48 @@
...
+        try expect(!isNormal(@bitCast(T, ~(~@as(TBits, 0) << mantissa_bits))));
```
LewisGaul, line 38: It's not obvious what's going on here... Looks like the first subnormal to me, which might be clearer to write as `@bitCast(T, @bitCast(TBits, floatMin(T)) - 1)`? Or at least a comment would be helpful :)

LewisGaul (commented):
A few comments/suggestions on clarity in the testing, but the changes look good :)

lib/std/math/isnormal.zig
viriuwu: I don't think there *is* a `floatTrueMin`(?), but adding one sounds like a good idea.

lib/std/math/isnormal.zig
viriuwu: I guess you're right! Wording is important, I more meant like a bit-hack haha. I'll adjust the wording.

lib/std/math/isnormal.zig
LewisGaul: Indeed, I was suggesting adding, as in my other comment :)
> Maybe add the max and min normals, e.g. `f128_max` and `f128_min` (add a function that takes the float type in `math.zig`?).

lib/std/math.zig
```diff
@@ -294,8 +293,13 @@ pub fn floatMantissaBits(comptime T: type) comptime_int {
...
-/// Returns the number of bits in the exponent of floating point type
-/// T.
+/// Returns the number of bits in the mantissa of floating point type T,
+/// including the implied leading 1 or 0.
+pub fn floatMantissaDigits(comptime T: type) comptime_int {
+    return floatMantissaBits(T) + 1;
+}
+
+/// Returns the number of bits in the exponent of floating point type T.
 pub fn floatExponentBits(comptime T: type) comptime_int {
```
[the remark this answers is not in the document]
viriuwu: It would not test my changes but it sounds like the only way to reason about "testing" the magic constants in the functions the others derive from - will do

lib/std/math/isnormal.zig
viriuwu: Apologies for the delay, I saw that first thing when I woke up and had to leave. I understand what you mean now. Here's a thought - right now float mins and maxes are just done as magic constants, which isn't very Zig-y and this PR is kind of trying to fix that. How about adding a new file, `math/float.zig`, where all these constant derivation functions can exist instead of being scattered throughout `math.zig` and individual files like `math/epsilon.zig`, `math/inf.zig`, etc., and have the existing constants like `f32_max` be defined as for example `floatMax(f32)` (since replacing usages of it is **really** outside the scope of this PR)? This file would also be a logical place for the aforementioned test cases to exist.

LewisGaul: Yep, big fan of that plan :) Fine if you think the change should be done in a separate PR (or I'd be happy to do it).

I put together https://github.com/LewisGaul/zig/wiki/Maths-APIs today, where these suggested functions are mentioned. Note the existence of `maxInt()` and `minInt()`, which should maybe be renamed to `intMax()` and `intMin()` at some point to be consistent with the float equivalents (where I don't think `trueMinFloat()` would be a great choice of name).

I think all of `inf()`, `nan()`, `snan()`, `floatMax()`, `floatMin()`, `floatTrueMin()`, `floatEps()` should be implemented generically in terms of `floatMantissaBits()` and `floatExponentBits()` (where required) or just simpler bit manipulations. Then it would make sense for the constants to be defined using the functions IMO (and wouldn't suggest removing the constants).

viriuwu: Glad we're on the same wavelength, I had the exact same thing where I tried to match `maxInt` and indeed ended up with `trueMinFloat` and decided against it hahaha. I was thinking it should be in terms of exponent and mantissa bits too and I like your proposal, the only thing that sticks out to me is the `floatToint` constant which is a name borrowed from C code, meaning "to int" (it is 1/ε).

Other than that, I'd be glad to rebase the PR as soon as I get home, if that sounds alright.

lib/std/math.zig
```diff
@@ -36,33 +36,39 @@ pub const sqrt2 = 1.414213562373095048801688724209698079;
...
+pub const f128_epsilon = floatEps(f128);
 pub const f16_toint = 1.0 / f16_epsilon;
-
-pub const epsilon = @import("math/epsilon.zig").epsilon;
+pub const f32_toint = 1.0 / f32_epsilon;
+pub const f64_toint = 1.0 / f64_epsilon;
+pub const f128_toint = 1.0 / f128_epsilon;
+pub fn epsilon(comptime T: type) T {
+    return floatEps(T);
+}
```
LewisGaul, line 71: Is there a reason for making the float functions return `comptime_float` rather than `T`? I would expect the constants above to be the type that's in the name (e.g. `f32_max` of type `f32`), and then `epsilon()` could just be an alias for `floatEps()` (planned to be removed?).

lib/std/math/isnormal.zig
LewisGaul: This comment (and a few others) were marked resolved but the code changes don't seem to be present, just checking this wasn't lost through some git operation or anything like that :)

lib/std/math.zig
viriuwu: I went with what std currently does:

- `epsilon` coerces to `T`, not matching this behaviour caused some random testcases to flip out
- the other ones are `comptime_float`

I also checked if all of them coerce correctly to their concrete types, which they appear to do. Do you think I should make them all `T`, and have the compatibility constants re-coerce to `comptime_float`?

lib/std/math/isnormal.zig
viriuwu: I had some unforeseen issues with running the test suite, using a new environment. Threw it at the CI and got some rest. I was gonna reread all of these and have a big summary typed up for the PR, given it worked.

lib/std/math.zig
LewisGaul: Ah I hadn't thought that the constants were already `comptime_float`, that should've been obvious to me! Do things break if the constants are made `T` instead of `comptime_float`? (or is there some other subtle reason for making them comptime that I'm overlooking?). I'm happy with what you have now, was just wondering really :)

viriuwu: > Do things break if they're made `T`?

Surprisingly yes, see [this CI build](https://ci.ziglang.org/ziglang/zig/1009/1/4).

I just noticed however that the `f128` constants are in fact *not* comptime due to being consteucted from bit repr, so maybe it'd be best to go with `T` and do `@as(comptime_float, ...)` for the other constants, as I do think it makes more sense to receive a `T` too, I think pretty much every usecase would be coercing them to `T` for usage.

LewisGaul: The CI failure you linked seems to be a problem with `epsilon()` returning a `comptime_float`, not with the constants being type `T` (e.g. `f32`)? FWIW the problem is that multiplying a `comptime_float` and a `comptime_int` gives a compile error - relevant issue: #8334. Have you seen failures when making the consts `T`, as opposed to this error with `epsilon()` being `comptime_float`?

viriuwu: You're right, that's kind of the opposite issue. I don't believe I saw anything break with the mins and maxes since it wouldn't be hard to imagine it breaking a simple `@TypeOf(x)` for someone so I just kept this PR entirely non-breaking.

LewisGaul: Why would anyone be doing a `@TypeOf(f64_min)` and expecting `comptime_float`, given `f64` is even in the name? Thinking about it I do feel like the constants should be made `T` rather than `comptime_float` - as you point out the `f128` constants already are, so this would give consistency with that. I suspect this was more something that was overlooked rather than intentional.

viriuwu: > I suspect this was more something that was overlooked rather than intentional.

I have the same feeling from `f128` being a concrete type while the rest just aren't. I'm probably overthinking it, I'll change the constants.

lib/std/math/isnormal.zig
viriuwu: Looks like even when the `T` is directly constructed instead of it being `comptime_float`-ified, it's a stage1 panic as expected. I left a TODO here and for `fabs` too.

lib/std/math.zig
viriuwu: Nevermind, it breaks [compiler_rt](https://ci.ziglang.org/ziglang/zig/1014/1/4) of all things, definitely not touching that. I'll make the old constants be what they were before.

viriuwu:
Alright, here's a summary of the PR:

1) Improve the floating point constants to be dynamically derived.

    The new implementation and public API is located in `math/float.zig` (that being `floatTrueMin(T)`, `floatMin(T)`, `floatMax(T)`, and `floatEps(T)`), and it's all derived from `floatExponentBits(T)` and `floatMantissaBits(T)`. The current implementation uses some magic numbers like these:

    https://github.com/ziglang/zig/blob/3763580e1e734b558499df5512f6a165f6db988b/lib/std/math.zig#L40-L43

    https://github.com/ziglang/zig/blob/3763580e1e734b558499df5512f6a165f6db988b/lib/std/math.zig#L47-L50

    The latter being particularly troublesome with precision, since it's `comptime_float` literals:

    ```
    @as(f128, math.f64_max)  = 1000011111111101111111111111111111111111111111111111111111111111111000000000000000110001101010000000111111001101011011111001111
    @as(f128, floatMax(f64)) = 1000011111111101111111111111111111111111111111111111111111111111111000000000000000000000000000000000000000000000000000000000000
    ```

    Existing constants such as `math.f32_min` are now made to be `floatMin(f32)`, meaning that they can slowly be phased out for the new API. Replacing all of those usages is definitely outside the scope of this PR, however.

    The `fXX_toint` constants did **not** get an equivalent API, since they have a very confusing name, are barely ever used, and can be derived on demand as `1.0 / floatEps(T)`. As for `inf(T)`, a generic implementation was added in `float.zig`, but removing `inf.zig` messes with the stage1 build, so that's out of scope, along with `nan(T)` which was untouched too, due to sNaN and qNaN not being clearly defined yet (reasonable to assume that's due to them being hardware dependent).

    Unfortunately, trying to use the mysterious `c_longdouble` with the new functions causes stage1 to freak out, despite the numbers being right. Once support for it is improved, it should work on its own with no changes.

2) Refactor `math.isNormal` and `math.fabs` to be generic (no magic numbers), improve tests,
and fix a rare panic in the current implementation along with it (see first post in this PR).

3) Use the new float derivation functions for `std.fmt.parse_hex_float` and `std.math.ldexp`,
since I knew they figured it out "manually" before. Small example of a refactor doable with the new APIs.

LewisGaul (approved):
Thanks for iterating in response to my comments, the scope ended up increasing but I'm personally happy with the changes!

LewisGaul:
Looks like the Windows CI timed out... Doesn't seem to be an actual problem with the PR.

LewisGaul:
Would be great to get this merged to allow for further float tidy-up, e.g. the implementation of inf and NaN (related: https://github.com/ziglang/zig/pull/10432). @matu3ba did you have any further comments?

matu3ba:
No, LGTM.

lib/std/math.zig
```diff
@@ -36,33 +36,38 @@ pub const sqrt2 = 1.414213562373095048801688724209698079;
...
+pub const floatMantissaBits = @import("math/float.zig").floatMantissaBits;
+pub const floatExponentMin = @import("math/float.zig").floatExponentMin;
+pub const floatExponentMax = @import("math/float.zig").floatExponentMax;
+pub const floatTrueMin = @import("math/float.zig").floatTrueMin;
+pub const floatMin = @import("math/float.zig").floatMin;
+pub const floatMax = @import("math/float.zig").floatMax;
+pub const floatEps = @import("math/float.zig").floatEps;
+
+// TODO These should eventually be removed, see #10133
+pub const f16_true_min: comptime_float = floatTrueMin(f16);
```
matu3ba, line 49: Nit: add the values as comments behind, so they can be used for tests later.

matu3ba:
Just a nit, which could make testing the later refactoring simpler. But which is not necessary.

lib/std/math.zig
viriuwu: Do you mean adding the deprecated definitions, like this?

```zig
pub const f16_true_min: comptime_float = floatTrueMin(f16); // 0.000000059604644775390625
```

... or do you mean an equation that is equivalent to what the value is, like:

```zig
pub const f16_true_min: comptime_float = floatTrueMin(f16); // 2^-15
```

... or something else? Not sure what you mean.

matu3ba: you can also add both, if you feel like it. Though the first should be sufficient.

viriuwu:
It's a good nit. No tests broke, but it's helpful to have it there.

viriuwu:
Looks like we racked up a lot of merge conflicts since december from two PRs that did what this originally did (which were merged) and `f80` support landing. I updated the logic to work correctly for extended-precision floats, and enabled tests for them in `fabs` and `isNormal`. Having to support extended-precision, it made  sense to add `floatMantissaDigits(T)` now (discussed as `MANT_DIG` earlier) since you can't derive that as bits+1 due to it storing an explicit integer part unlike other floating point types.

ifreund:
Looks like a test regarding isNormal failed on the lavahut CI: 
```
685/2198 math.isnormal.test "std-native-Debug--multi math.isNormal"... FAIL (TestUnexpectedResult)
/usr/home/build/zig/lib/std/math/isnormal.zig:32:9: 0x4481e4 in math.isnormal.test "std-native-Debug--multi math.isNormal" (test)
        try expect(isNormal(@as(T, 1.0)));
```

viriuwu:
> Looks like a test regarding isNormal failed on the lavahut CI

Indeed, not quite sure what's up with that (or the surrounding errors about io.Reader...) I'll see what the other CIs think and then try to figure it out, although I have very limited experience with FreeBSD but I can just manually unroll the `inline for` since the current error messages aren't too helpful. Guessing it's to do with f80.

viriuwu:
Added a test skip for the FreeBSD due to #11391 weirdness and all the CIs are happy now!

lib/std/math.zig
```diff
@@ -36,38 +36,44 @@ pub const sqrt2 = 1.414213562373095048801688724209698079;
...
+pub const floatMantissaBits = @import("math/float.zig").floatMantissaBits;
+pub const floatMantissaDigits = @import("math/float.zig").floatMantissaDigits;
+pub const floatExponentMin = @import("math/float.zig").floatExponentMin;
+pub const floatExponentMax = @import("math/float.zig").floatExponentMax;
+pub const floatTrueMin = @import("math/float.zig").floatTrueMin;
+pub const floatMin = @import("math/float.zig").floatMin;
+pub const floatMax = @import("math/float.zig").floatMax;
+pub const floatEps = @import("math/float.zig").floatEps;
+
+// TODO These should eventually be removed, see #10133
```
ifreund, line 49:
```suggestion
// TODO Replace with @compileError("deprecated for foobar") after 0.10.0 is released.
```

lib/std/math/isnormal.zig
```diff
@@ -1,57 +1,55 @@
...
-        },
-        else => {
-            @compileError("isNormal not implemented for " ++ @typeName(T));
-        },
+    const TBits = std.meta.Int(.unsigned, @bitSizeOf(T));
+    if (@typeInfo(T) != .Float) {
+        @compileError("isNormal not implemented for " ++ @typeName(T));
     }
+
+    // self-explanatory constants used below
```
ifreund, line 13:
```suggestion
```

ifreund (approved):
LGTM, thanks for your patience here.

Just one improvement to the deprecation TODO then this can finally be merged.

viriuwu:
Done and done. I also realised while quickly glancing over the last commit which includes a nit about `IntT` that I implied I updated it in two places, when I only did it one. So I fixed that now too. Thank you for reviewing, excited to get this merged!

lib/std/math/isnormal.zig
viriuwu: yeah that's pretty bad honestly not sure why i put that in

PR #10135, opened 2021-11-11
docs: reorganize `@truncate` and `@intCast` for clarity

praschke:
the code example immediately following `@truncate`'s introduction is a demonstration of `@intCast`'s undefined behavior on oversized values. this misled myself and another person when trying to answer a question, so here's a clearer arrangement with links between `@truncate` and `@intCast`.


ominitay:
Thank you so much! I think I was the person you are talking about :)

Issue #7360, opened 2020-12-09
make `zig cc` use clang-compatible -target and -mcpu parameters


Vurich:
If you run clang with `--target=arm64-unknown-linux-gnu` it will understand it to mean the same as `arm64-linux-gnu`, but with `zig cc` it returns `UnknownOperatingSystem`. I wanted to use `zig cc` in a Rust build script, and this is an issue as Rust _only_ uses 4-part target triples - and so `<arch>-unknown-<sys>-<abi>` is by far the most common format for targets. While I can work around this in my particular case, I wouldn't expect it to be particularly difficult to implement in Zig and it would make `zig cc` compatible with clang in more contexts.

LemonBoy:
I've hit this very same problem today and indeed having to patch the triple every time is annoying.
This said, Zig has no concept of vendors yet so we could:
* Parse and ignore the field,
* Parse the field and relay it to LLVM, Zig is still oblivious to the vendor part,
* Same as above, but only when running `zig cc`,
* Enhance Zig's triple to include the vendor part too.

andrewrk:
Currently the `-target` and `-mcpu` CLI parameters are parsed according to zig's understanding of these values rather than LLVM's, including in `zig cc` mode. This is intentional because Zig has a different set of targets than LLVM does, and we do want you to be able to specify operating systems, architectures, and CPU features according to zig's model rather than LLVM's model.

Of course it does have the downside noted by this issue.

It's annoying but I think the best solution to the problem is to parse the triple according to LLVM rules, *only in zig cc* (as @LemonBoy notes in the 3rd option), which involves a more onerous process than simply omitting the vendor, because the CPU model names and CPU features have different syntax (underscore instead of dash, for instance, but also the sets of features are not the same).

Vurich:
Maybe if there was some standard way to convert between LLVM and Zig triples (and `-mcpu` arguments) that existed outside of Zig then it wouldn't be so bad? You could then have a really thin script called, like `zig clang` or whatever that does the conversion and passes it to `zig cc`. Seems kinda silly to convert twice but it could lead to more flexibility in `zig cc`'s argument syntax if you have an entirely separate binary whose only purpose is converting the arguments.

andrewrk:
Duplicate of #4911

PR #10139, opened 2021-11-12
cross_target: Fixed targets with vendor for Clang

RossSmyth:
Should be at least a temporary fix for #7360.

Of LemonBoy's options listed in his comment, I picked the second. parse it, but Zig doesn't do anything with it.

### DIFF
--- a/lib/std/zig/cross_target.zig
+++ b/lib/std/zig/cross_target.zig
@@ -41,6 +41,9 @@ pub const CrossTarget = struct {
     /// based on the `os_tag`.
     dynamic_linker: DynamicLinker = DynamicLinker{},
 
+    /// `null` means no vendor was passed
+    vendor_tag: ?Target.Vendor = null,
+
     pub const CpuModel = union(enum) {
         /// Always native
         native,
@@ -165,6 +168,7 @@ pub const CrossTarget = struct {
     pub fn toTarget(self: CrossTarget) Target {
         return .{
             .cpu = self.getCpu(),
+            .vendor = self.vendor_tag,
             .os = self.getOs(),
             .abi = self.getAbi(),
         };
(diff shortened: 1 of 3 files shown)


lib/std/zig/cross_target.zig
```diff
@@ -694,6 +718,14 @@ pub const CrossTarget = struct {
         set.removeFeatureSet(self.cpu_features_sub);
     }
 
+    /// This is a dumb parser to essentially validate that the vendor provided is one Clang will accept.
+    fn parseVendor(result: *CrossTarget, diags: *ParseOptions.Diagnostics, text: []const u8) !void {
+        result.vendor_tag = std.meta.stringToEnum(Target.Vendor, text) orelse
+            return error.UnknownVendor;
```
RossSmyth, line 724: Do I need to add this to the enum defined [here](https://github.com/ziglang/zig/blob/30d01c8fea68baab396081da040f49defa494088/src/stage1.zig#L182-L264)? It is labeled ABI which makes me think it would also need to be added [here](https://github.com/ziglang/zig/blob/c12704a3395a11c28569ebb160e6ac33091d983a/src/stage1/error.cpp#L11-L93) and [here](https://github.com/ziglang/zig/blob/c12704a3395a11c28569ebb160e6ac33091d983a/src/stage1/stage2.h#L35-L117). It compiled and ran the tests without doing so, so I didn't worry about it for the initial commit.

andrewrk:
Thank you for the patch-

Please see https://github.com/ziglang/zig/issues/7360#issuecomment-969549105. Modifying std.Target and zig's general parsing of `-target` and `-mcpu` is not going to be the best solution to this problem. Instead what has been proposed (by me in that comment) but not accepted yet is to have an additional layer as part of `zig cc` *only* that translates Clang-compatible `-target` and `-mcpu` syntax to zig's target awareness.

Issue #10138, opened 2021-11-11
stage2: registers get overwritten when ignoring return


g-w1:
### Remember to search before filing a new report

- [X] I searched for this bug and did not find it in the issue tracker, and furthermore, the title I used above will make this new bug report turn up in the search results for my query.

### Zig Version

0.9.0-dev.1599+008b0ec5e

### Steps to Reproduce

```
/tmp via ↯ v0.9.0-dev.1599+008b0ec5e via ❄️  impure
❯ cat test.zig
pub fn main() void {
    const fd = open();
    _ = write(fd, "a", 1);
    _ = close(fd);
}

fn open() usize {
    return 42;
}

fn write(fd: usize, a: [*]const u8, len: usize) usize {
    _ = fd;
    _ = a;
    return len;
}

fn close(fd: usize) usize {
    _ = fd;
    @breakpoint();
    return 0;
}
/tmp via ↯ v0.9.0-dev.1599+008b0ec5e via ❄️  impure
❯ zig build-exe test.zig -fno-stage1 -fno-LLVM
error(link): TODO implement .debug_info for type '[*]const u8'
/tmp via ↯ v0.9.0-dev.1599+008b0ec5e via ❄️  impure
❯ gdb ./test
GNU gdb (GDB) 9.2
BFD: /tmp/test: invalid string offset 12628 >= 163 for section `.shstrtab'
BFD: /tmp/test: invalid string offset 2307416 >= 163 for section `.shstrtab'
(gdb) r
Starting program: /tmp/test

Program received signal SIGTRAP, Trace/breakpoint trap.
close (fd=1) at test.zig:36
(gdb)
```

### Expected Behavior

fd is 42, when viewed in gdb

### Actual Behavior

We can see in gdb that `fd` is one, while it should be 42. This is because write returns 1, and overwrites the memory location of fd:
```asm
0000000008000062 <main>:
 8000062:	55                   	push   rbp
 8000063:	48 89 e5             	mov    rbp,rsp
 8000066:	48 83 ec 00          	sub    rsp,0x0
 800006a:	ff 14 25 38 00 00 04 	call   QWORD PTR ds:0x4000038 ; open
 8000071:	48 8b f8             	mov    rdi,rax
 8000074:	48 8b 34 25 48 00 00 	mov    rsi,QWORD PTR ds:0x4000048
 800007b:	04
 800007c:	48 c7 c2 01 00 00 00 	mov    rdx,0x1
[...]

PR #10140, opened 2021-11-12
stage2 x86_64 codegen: don't count return registers as callee-preserved

g-w1:
Attempts to close https://github.com/ziglang/zig/issues/10138
It fixes the test case in #10138 (now in close fd is 42)

### DIFF
--- a/src/arch/x86_64/CodeGen.zig
+++ b/src/arch/x86_64/CodeGen.zig
@@ -349,6 +349,13 @@ pub fn addExtraAssumeCapacity(self: *Self, extra: anytype) u32 {
 fn gen(self: *Self) InnerError!void {
     const cc = self.fn_type.fnCallingConvention();
     if (cc != .Naked) {
+        // push the callee_preserved_regs that were used
+        const backpatch_push_callee_preserved_regs_i = try self.addInst(.{
+            .tag = .push_regs_from_callee_preserved_regs,
+            .ops = undefined,
+            .data = .{ .regs_to_push_or_pop = undefined }, // to be backpatched
+        });
+
         _ = try self.addInst(.{
             .tag = .push,
             .ops = (Mir.Ops{
@@ -422,6 +429,22 @@ fn gen(self: *Self) InnerError!void {
                 .reg1 = .rbp,
             }).encode(),
             .data = undefined,
+        });
+        // calculate the data for callee_preserved_regs to be pushed and popped
+        var callee_preserved_regs_push_data: u32 = 0x0;
+        inline for (callee_preserved_regs) |reg, i| {
+            if (self.register_manager.isRegAllocated(reg)) {
+                callee_preserved_regs_push_data |= 1 << @intCast(u5, i);
+            }
+        }
+        const data = self.mir_instructions.items(.data);
+        // backpatch the push instruction
+        data[backpatch_push_callee_preserved_regs_i].regs_to_push_or_pop = callee_preserved_regs_push_data;
+        // pop the callee_preserved_regs
+        _ = try self.addInst(.{
+            .tag = .pop_regs_from_callee_preserved_regs,
+            .ops = undefined,
+            .data = .{ .regs_to_push_or_pop = callee_preserved_regs_push_data },
         });
         _ = try self.addInst(.{
             .tag = .ret,
--- a/src/arch/x86_64/CodeGen.zig
+++ b/src/arch/x86_64/CodeGen.zig
@@ -1902,10 +1902,10 @@ fn airCall(self: *Self, inst: Air.Inst.Index) !void {
                     try self.register_manager.getReg(reg, null);
                     try self.genSetReg(arg_ty, reg, arg_mcv);
                 },
-                .stack_offset => {
+                .stack_offset => |off| {
                     // Here we need to emit instructions like this:
                     // mov     qword ptr [rsp + stack_offset], x
-                    return self.fail("TODO implement calling with parameters in memory", .{});
+                    try self.genSetStack(arg_ty, off, arg_mcv);
                 },
                 .ptr_stack_offset => {
                     return self.fail("TODO implement calling with MCValue.ptr_stack_offset arg", .{});
--- a/src/arch/x86_64/Emit.zig
+++ b/src/arch/x86_64/Emit.zig
@@ -142,6 +142,9 @@ pub fn emitMir(emit: *Emit) InnerError!void {
             .dbg_epilogue_begin => try emit.mirDbgEpilogueBegin(inst),
             .arg_dbg_info => try emit.mirArgDbgInfo(inst),
 
+            .push_regs_from_callee_preserved_regs => try emit.mirPushPopRegsFromCalleePreservedRegs(.push, inst),
+            .pop_regs_from_callee_preserved_regs => try emit.mirPushPopRegsFromCalleePreservedRegs(.pop, inst),
+
             else => {
                 return emit.fail("Implement MIR->Isel lowering for x86_64 for pseudo-inst: {s}", .{tag});
             },
@@ -242,6 +245,39 @@ fn mirPushPop(emit: *Emit, tag: Mir.Inst.Tag, inst: Mir.Inst.Index) InnerError!v
             }
         },
         0b11 => unreachable,
+    }
+}
+fn mirPushPopRegsFromCalleePreservedRegs(emit: *Emit, tag: Mir.Inst.Tag, inst: Mir.Inst.Index) InnerError!void {
+    const callee_preserved_regs = bits.callee_preserved_regs;
+    // PUSH/POP reg
+    const opc: u8 = switch (tag) {
+        .push => 0x50,
+        .pop => 0x58,
+        else => unreachable,
+    };
+
+    const regs = emit.mir.instructions.items(.data)[inst].regs_to_push_or_pop;
+    if (tag == .push) {
+        for (callee_preserved_regs) |reg, i| {
+            if ((regs >> @intCast(u5, i)) & 1 == 0) continue;
+            const encoder = try Encoder.init(emit.code, 2);
+            encoder.rex(.{
+                .b = reg.isExtended(),
+            });
+            encoder.opcode_withReg(opc, reg.lowId());
+        }
+    } else {
+        // pop in the reverse direction
+        var i = callee_preserved_regs.len;
+        while (i > 0) : (i -= 1) {
+            const reg = callee_preserved_regs[i - 1];
+            if ((regs >> @intCast(u5, i - 1)) & 1 == 0) continue;
+            const encoder = try Encoder.init(emit.code, 2);
+            encoder.rex(.{
+                .b = reg.isExtended(),
+            });
+            encoder.opcode_withReg(opc, reg.lowId());
+        }
     }
 }
 
--- a/src/arch/x86_64/bits.zig
+++ b/src/arch/x86_64/bits.zig
@@ -84,15 +84,13 @@ pub const Register = enum(u7) {
     /// Returns the index into `callee_preserved_regs`.
     pub fn allocIndex(self: Register) ?u4 {
         return switch (self) {
-            .rax, .eax, .ax, .al => 0,
-            .rcx, .ecx, .cx, .cl => 1,
-            .rdx, .edx, .dx, .dl => 2,
-            .rsi, .esi, .si  => 3,
-            .rdi, .edi, .di => 4,
-            .r8, .r8d, .r8w, .r8b => 5,
-            .r9, .r9d, .r9w, .r9b => 6,
-            .r10, .r10d, .r10w, .r10b => 7,
-            .r11, .r11d, .r11w, .r11b => 8,
+            .rcx, .ecx, .cx, .cl => 0,
+            .rsi, .esi, .si  => 1,
+            .rdi, .edi, .di => 2,
+            .r8, .r8d, .r8w, .r8b => 3,
+            .r9, .r9d, .r9w, .r9b => 4,
+            .r10, .r10d, .r10w, .r10b => 5,
+            .r11, .r11d, .r11w, .r11b => 6,
             else => null,
         };
     }
@@ -145,7 +143,8 @@ pub const Register = enum(u7) {
 // zig fmt: on
 
 /// These registers belong to the called function.
-pub const callee_preserved_regs = [_]Register{ .rax, .rcx, .rdx, .rsi, .rdi, .r8, .r9, .r10, .r11 };
+/// TODO should the return_regs be in this array?
+pub const callee_preserved_regs = [_]Register{ .rcx, .rsi, .rdi, .r8, .r9, .r10, .r11 };
 pub const c_abi_int_param_regs = [_]Register{ .rdi, .rsi, .rdx, .rcx, .r8, .r9 };
 pub const c_abi_int_return_regs = [_]Register{ .rax, .rdx };
 
--- a/src/arch/x86_64/Mir.zig
+++ b/src/arch/x86_64/Mir.zig
@@ -264,8 +264,21 @@ pub const Inst = struct {
 
         /// arg debug info
         arg_dbg_info,
-    };
 
+        /// push registers from the callee_preserved_regs
+        /// data is the bitfield of which regs to push 
+        /// for example on x86_64, the callee_preserved_regs are [_]Register{ .rcx, .rsi, .rdi, .r8, .r9, .r10, .r11 };    };
+        /// so to push rcx and r8 one would make data 0b00000000_00000000_00000000_00001001 (the first and fourth bits are set)
+        /// ops is unused
+        push_regs_from_callee_preserved_regs,
+
+        /// pop registers from the callee_preserved_regs
+        /// data is the bitfield of which regs to pop
+        /// for example on x86_64, the callee_preserved_regs are [_]Register{ .rcx, .rsi, .rdi, .r8, .r9, .r10, .r11 };    };
+        /// so to pop rcx and r8 one would make data 0b00000000_00000000_00000000_00001001 (the first and fourth bits are set)
+        /// ops is unused
+        pop_regs_from_callee_preserved_regs,
+    };
     /// The position of an MIR instruction within the `Mir` instructions array.
     pub const Index = u32;
 
@@ -284,6 +297,8 @@ pub const Inst = struct {
         got_entry: u32,
         /// Index into `extra`. Meaning of what can be found there is context-dependent.
         payload: u32,
+        /// A bitfield of which callee_preserved_regs to push
+        regs_to_push_or_pop: u32,
     };
 
     // Make sure we don't accidentally make instructions bigger than expected.
--- a/src/link/Elf.zig
+++ b/src/link/Elf.zig
@@ -2604,7 +2604,7 @@ fn addDbgInfoType(self: *Elf, ty: Type, dbg_info_buffer: *std.ArrayList(u8)) !vo
                 // DW.AT.name,  DW.FORM.string
                 try dbg_info_buffer.writer().print("{}\x00", .{ty});
             } else {
-                log.err("TODO implement .debug_info for type '{}'", .{ty});
+                log.warn("TODO implement .debug_info for type '{}'", .{ty});
                 try dbg_info_buffer.append(abbrev_pad1);
             }
         },
--- a/test/cases.zig
+++ b/test/cases.zig
@@ -1819,4 +1819,36 @@ pub fn addCases(ctx: *TestContext) !void {
             ":2:28: error: cannot set address space of local variable 'foo'",
         });
     }
+    {
+        var case = ctx.exe("issue 10138: callee preserved regs working", linux_x64);
+        case.addCompareOutput(
+            \\pub fn main() void {
+            \\    const fd = open();
+            \\    _ = write(fd, "a", 1);
+            \\    _ = close(fd);
+            \\}
+            \\
+            \\fn open() usize {
+            \\    return 42;
+            \\}
+            \\
+            \\fn write(fd: usize, a: [*]const u8, len: usize) usize {
+            \\    return syscall4(.WRITE, fd, @ptrToInt(a), len);
+            \\}
+            \\
+            \\fn syscall4(n: enum { WRITE }, a: usize, b: usize, c: usize) usize {
+            \\    _ = n;
+            \\    _ = a;
+            \\    _ = b;
+            \\    _ = c;
+            \\    return 23;
+            \\}
+            \\
+            \\fn close(fd: usize) usize {
+            \\    if (fd != 42)
+            \\        unreachable;
+            \\    return 0;
+            \\}
+        , "");
+    }
 }


g-w1:
Ahh, I see the problem here: now the actual problem is that callee_preserved_registers is not implemented in the x86_64 backend. I'll do the naive solution, for now.

src/arch/x86_64/bits.zig
```diff
@@ -145,7 +143,7 @@ pub const Register = enum(u7) {
 // zig fmt: on
 
 /// These registers belong to the called function.
-pub const callee_preserved_regs = [_]Register{ .rax, .rcx, .rdx, .rsi, .rdi, .r8, .r9, .r10, .r11 };
+pub const callee_preserved_regs = [_]Register{ .rcx, .rsi, .rdi, .r8, .r9, .r10, .r11 };
```
kubkon, line 146: According to [this guide](http://cons.mit.edu/sp17/x86-64-architecture-guide.html), callee saved registers are `.rbx, .rbp, .r12, .r13, .r14, .r15`, while the rest belongs to the callee including `.rax` and `.rdx`. Is there a reason why you removed them from the list?

g-w1: I was following the guidance of @joachimschmidt557 https://discord.com/channels/605571803288698900/785499283368706060/908118081434435614 , and it fixed the test case in #10138 . It also didn't make sense how if they needed to be pushed and popped in every function that they could be returned. What do you think?

kubkon: Ah, I see. My suggestion would be to go with your changes for the moment, but drop a TODO comment to figure it out in such a way so that we adhere to the amd64 manual (i.e., having the full registers set).

g-w1: Ok, thanks for the suggestion!

src/arch/x86_64/CodeGen.zig
```diff
@@ -382,8 +382,32 @@ fn gen(self: *Self) InnerError!void {
             .data = undefined,
         });
 
+        // push the callee_preserved_regs that were used
+        _ = try self.addInst(.{
+            .tag = .push_regs_from_callee_preserved_regs,
+            .ops = undefined,
+            .data = undefined, // to be backpatched
+        });
+        const backpatch_push_callee_preserved_regs_i = self.mir_instructions.len;
```
joachimschmidt557, line 391:
```suggestion
        const backpatch_push_callee_preserved_regs_i = try self.addInst(.{
            .tag = .push_regs_from_callee_preserved_regs,
            .ops = undefined,
            .data = undefined, // to be backpatched
        });
```

Then you can use the index (without `- 1`) in the following code.

src/arch/x86_64/bits.zig
joachimschmidt557: I don't see how the information provided in the guide contradicts removing `rax` and `rdx`. The guide describes these registers as not "saved across calls", which is what I understand to be a `callee_preserved_register`. The definition I currently have for a `callee_preserved_register` is that such a register is not affected in any way when calling another function (whether the other function itself overwrites it or when this register is used as a parameter to that function and is thus overwritten etc.). As far as I can tell, this is also the definition the guide uses.

kubkon: Mhm, that sounds about right. My argument here was to investigate why there is a need to explicitly remove two registers from the list thus making them not used automatically by the register allocator especially if they are callee preserved.

src/arch/x86_64/Emit.zig
```diff
@@ -244,6 +247,25 @@ fn mirPushPop(emit: *Emit, tag: Mir.Inst.Tag, inst: Mir.Inst.Index) InnerError!v
...
+
+    const regs = emit.mir.instructions.items(.data)[inst].regs_to_push_or_pop;
+    for (callee_preserved_regs) |reg, i| {
+        if ((regs >> @intCast(u5, i)) & 1 == 0) continue;
+        const encoder = try Encoder.init(emit.code, 2);
+        encoder.rex(.{
+            .b = reg.isExtended(),
+        });
+        encoder.opcode_withReg(opc, reg.lowId());
+    }
```
joachimschmidt557, line 267: I think there may be an issue here: The order in which the registers are pushed to the stack must be the exact opposite of the order in which the registers are popped iirc, i.e. the last register pushed must be the first register popped.

g-w1: this makes a lot of sense. thanks

g-w1:
This causes this test case to fail; add does not produce the right number, it produces 561, when it should produce 1221. I think this is because the return registers were taken away and cannot be used.

g-w1:
Found the offending bit of assembly:
```asm
add:
push   rcx
push   rsi
push   rdi
push   r8
push   r9
push   r10
push   r11
push   rbp
mov    rbp,rsp
sub    rsp,0x10
mov    ecx,esi
add    ecx,edi
mov    r8,rcx
add    r8d,edi
add    esi,r8d
mov    r9d,esi
add    r9,r8
mov    r10,r9
add    r10d,esi
mov    r11,r10
add    r11,r9
mov    DWORD PTR [rbp-0x4],ecx
mov    rcx,r11
add    rcx,r10
mov    DWORD PTR [rbp-0x8],ecx
mov    rcx,r8 ; rcx is 100 here, r8 is 10 here https://github.com/ziglang/zig/blob/1d55705fa42d83556346f8fcb3a8fce0d55f06ad/test/cases.zig#L646 should be generated, instead the next line gets generated and 20 is generated for rcx, which screws everything else up
add    rcx,rcx
mov    DWORD PTR [rbp-0xc],ecx
mov    rcx,QWORD PTR [rbp-0x8]
add    rcx,rcx
mov    DWORD PTR [rbp-0x10],ecx
mov    rcx,QWORD PTR [rbp-0xc]
add    rcx,rcx
add    DWORD PTR [rbp-0x4],ecx
add    DWORD PTR [rbp-0x4],r8d
add    DWORD PTR [rbp-0x4],esi
add    DWORD PTR [rbp-0x4],r9d
add    DWORD PTR [rbp-0x4],r10d
add    DWORD PTR [rbp-0x4],r11d
mov    rsi,QWORD PTR [rbp-0x4]
add    rsi,QWORD PTR [rbp-0x8]
add    rsi,QWORD PTR [rbp-0xc]
add    rsi,QWORD PTR [rbp-0x10]
add    rsi,rcx
add    esi,edi
add    esi,edi
mov    eax,esi
add    rsp,0x10
pop    rbp
pop    r11
pop    r10
pop    r9
pop    r8
pop    rdi
pop    rsi
pop    rcx
ret
```
If someone could help me with debugging this, it would be very much appreciated, as this is the only test that is failing (the mul works fine).

kubkon:
> Found the offending bit of assembly:
> 
> ```assembly
> add:
> push   rcx
> push   rsi
> push   rdi
> push   r8
> push   r9
> push   r10
> push   r11
> push   rbp
> mov    rbp,rsp
> sub    rsp,0x10
> mov    ecx,esi
> add    ecx,edi
> mov    r8,rcx
> add    r8d,edi
> add    esi,r8d
> mov    r9d,esi
> add    r9,r8
> mov    r10,r9
> add    r10d,esi
> mov    r11,r10
> add    r11,r9
> mov    DWORD PTR [rbp-0x4],ecx
> mov    rcx,r11
> add    rcx,r10
> mov    DWORD PTR [rbp-0x8],ecx
> mov    rcx,r8 ; rcx is 100 here, r8 is 10 here https://github.com/ziglang/zig/blob/1d55705fa42d83556346f8fcb3a8fce0d55f06ad/test/cases.zig#L646 should be generated, instead the next line gets generated and 20 is generated for rcx, which screws everything else up
> add    rcx,rcx
> mov    DWORD PTR [rbp-0xc],ecx
> mov    rcx,QWORD PTR [rbp-0x8]
> add    rcx,rcx
> mov    DWORD PTR [rbp-0x10],ecx
> mov    rcx,QWORD PTR [rbp-0xc]
> add    rcx,rcx
> add    DWORD PTR [rbp-0x4],ecx
> add    DWORD PTR [rbp-0x4],r8d
> add    DWORD PTR [rbp-0x4],esi
> add    DWORD PTR [rbp-0x4],r9d
> add    DWORD PTR [rbp-0x4],r10d
> add    DWORD PTR [rbp-0x4],r11d
> mov    rsi,QWORD PTR [rbp-0x4]
> add    rsi,QWORD PTR [rbp-0x8]
> add    rsi,QWORD PTR [rbp-0xc]
> add    rsi,QWORD PTR [rbp-0x10]
> add    rsi,rcx
> add    esi,edi
> add    esi,edi
> mov    eax,esi
> add    rsp,0x10
> pop    rbp
> pop    r11
> pop    r10
> pop    r9
> pop    r8
> pop    rdi
> pop    rsi
> pop    rcx
> ret
> ```
> 
> If someone could help me with debugging this, it would be very much appreciated, as this is the only test that is failing (the mul works fine).

Another concerning bit here is the fact that we perform `edi <- esi + edi` twice for some reason.

src/arch/x86_64/CodeGen.zig
```diff
@@ -349,6 +349,13 @@ pub fn addExtraAssumeCapacity(self: *Self, extra: anytype) u32 {
 fn gen(self: *Self) InnerError!void {
     const cc = self.fn_type.fnCallingConvention();
     if (cc != .Naked) {
+        // push the callee_preserved_regs that were used
+        const backpatch_push_callee_preserved_regs_i = try self.addInst(.{
```
kubkon, line 353: This might be obvious but why do we need a special pseudo-instruction for register allocation?

g-w1: This is not register allocation, but pushing/popping registers. The reason that it is necessary is that before we gen the body, we don't know which registers are used. So we need to backpatch them. The reason why multiple push instructions can't be done is that we don't know which registers need to be pushed, and therefore we don't know how many mir indexes it will take.

g-w1:
For context, this is the old assembly
```asm
00000000080000e0 <add>: old
push   rbp
mov	rbp,rsp
sub	rsp,0x10
mov	eax,esi
add	eax,edi
mov	rcx,rax
add	ecx,edi
add	esi,ecx
mov	edx,esi
add	rdx,rcx
mov	r8,rdx
add	r8d,esi
mov	r9,r8
add	r9,rdx
mov	r10,r9
add	r10,r8
mov	r11,rcx
add	r11,r10
mov	DWORD PTR [rbp-0x4],eax
mov	rax,r11
add	rax,r10
mov	DWORD PTR [rbp-0x8],eax
add	rax,r11
add	DWORD PTR [rbp-0x4],eax
add	DWORD PTR [rbp-0x4],ecx
add	DWORD PTR [rbp-0x4],esi
add	DWORD PTR [rbp-0x4],edx
add	DWORD PTR [rbp-0x4],r8d
add	DWORD PTR [rbp-0x4],r9d
add	DWORD PTR [rbp-0x4],r10d
add	DWORD PTR [rbp-0x4],r11d
mov	rcx,QWORD PTR [rbp-0x4]
add	rcx,QWORD PTR [rbp-0x8]
add	rcx,rax
add	ecx,edi
add	ecx,edi
mov	eax,ecx
add	rsp,0x10
pop	rbp
ret
```
Some addition is done twice here too.

src/arch/x86_64/CodeGen.zig
kubkon: Oh OK, I think I get it now. So we basically need a placeholder within MIR to be able to say that in this place in code we will need to preserve a couple registers and then pop them in reverse order but we don't yet know which ones until we finish generating codegen for the given block / scope, correct? Hmm, there might not be a better way to do this in one pass than what you suggested here, but I'm wondering if it was worth considering an alternative: for every block / scope that codegen needs to generate, we would open a new "subslice" of MIR instructions which we would then be free to fixup with the correct number of push / pop instructions and only then we would append the entire thing to the global `mir_inststructions` multi arraylist. I think this approach might be worth considering since it would remove the need for a compound MIR instruction such as `push_regs_from_callee_preserved_regs`, and the resulting MIR slice would only consist of granual push / pop instructions. I haven't given it that much thought yet but this might be a slightly better input for any subsequent optimisation passes we'll want to perfom on the codegen in the future perhaps? @g-w1 @joachimschmidt557 lemme know what you guys think about this proposal.

kubkon:
> For context, this is the old assembly
> 
> ```assembly
> 00000000080000e0 <add>: old
> push   rbp
> mov	rbp,rsp
> sub	rsp,0x10
> mov	eax,esi
> add	eax,edi
> mov	rcx,rax
> add	ecx,edi
> add	esi,ecx
> mov	edx,esi
> add	rdx,rcx
> mov	r8,rdx
> add	r8d,esi
> mov	r9,r8
> add	r9,rdx
> mov	r10,r9
> add	r10,r8
> mov	r11,rcx
> add	r11,r10
> mov	DWORD PTR [rbp-0x4],eax
> mov	rax,r11
> add	rax,r10
> mov	DWORD PTR [rbp-0x8],eax
> add	rax,r11
> add	DWORD PTR [rbp-0x4],eax
> add	DWORD PTR [rbp-0x4],ecx
> add	DWORD PTR [rbp-0x4],esi
> add	DWORD PTR [rbp-0x4],edx
> add	DWORD PTR [rbp-0x4],r8d
> add	DWORD PTR [rbp-0x4],r9d
> add	DWORD PTR [rbp-0x4],r10d
> add	DWORD PTR [rbp-0x4],r11d
> mov	rcx,QWORD PTR [rbp-0x4]
> add	rcx,QWORD PTR [rbp-0x8]
> add	rcx,rax
> add	ecx,edi
> add	ecx,edi
> mov	eax,ecx
> add	rsp,0x10
> pop	rbp
> ret
> ```
> 
> Some addition is done twice here too.

Yeah, I actually replicated your problem locally on the M1 macbook. It's kinda cool that this test case can be tested verbatim on `x86_64-macos` now too :-)

kubkon:
So the problem here seems that `rcx` acts as an accumulator in the modified code and it's clobbered with the contents of `r8` in the line highlighted by you, correct?

src/arch/x86_64/CodeGen.zig
g-w1: The point about further optimization passes makes a lot of sense. I don't really understand how the "subslice" would work with prepending it to the function's Mir.

kubkon: Oh, sorry @g-w1, that's me being a poor communicator. What I meant was that we need to do this backpatching any time we sort of open a new scope. Then, we would simply accumulate the MIR instructions local to this scope in a temp multi arraylist (or similar), which we would then fixup using your logic while leaving the scope, and we would append the fixed-up list to the global `mir_instructions`. Something along the lines of:

```zig
var local_mir = std.MultiArrayList(Mir).init(gpa);
defer local_mir.deinit();

// do all the logic here like the logic from gen() or similar

// since local_mir is scoped to this local block, we can now backpatch push/pops granularly like

inline for (...) |reg| {
    try self.addInst(push(reg));
}

try self.mir_instructions.appendSlice(local_mir);

inline for (...) |reg| {
    try self.addInst(pop(reg));
}
```

Lemme know if that makes sense! Sorry for the pseudo-zig but I'm on my phone ATM so trying to just get the message across ;-)

g-w1: Ahh, this totally makes sense. I'll switch to it.

kubkon: Nice! Try it out and lemme know how this goes. In the meantime, when I have some time to access my laptop, I'll help you in debugging the clobbering issue too.

joachimschmidt557: Introducing MIR sub-slices may introduce a new issue: In branches, we reference other MIR instructions by their absolute index. When a fresh MIR sub-slice needs to be appended to the global MIR list, we need to rewrite all of those references to the new, global indices. Alternatively, we could use relative offsets (which would then need to be signed integers to enable backwards jumps).

kubkon: Mhm, that's a good point. Having said that though, push and pop instructions in the scopes prologue and epilogue should be fairly standard and would not have a jump to any one instruction, right? That is, any jump reloc from within the block would not end up in those regions but rather either after the scope's prologue or before the epilogue, correct? Perhaps it'd be worth saving the relative indices of jmp instructions that require a reloc in a separate arraylist side-by-side with the MIR subslice, and after working out the push/pop prologue/epilogue just translating the jmp targets by the offset equivalent to the prologue length? I think this should be doable and fairly efficient, requiring surgical fixups of a couple MIR instructions only. However, if it becomes overly complicated, then I'm happy to revisit this in the future instead.

g-w1: Here is my diff for this feature 
```diff
diff --git a/lib/std/multi_array_list.zig b/lib/std/multi_array_list.zig
index fd387513b..7ec67785b 100644
--- a/lib/std/multi_array_list.zig
+++ b/lib/std/multi_array_list.zig
@@ -174,6 +174,18 @@ pub fn MultiArrayList(comptime S: type) type {
             self.set(self.len - 1, elem);
         }
 
+        /// Append the slice of items to the list. Allocates more
+        /// memory as necessary.
+        /// TODO this can be optimised for speed
+        pub fn appendSlice(self: *Self, gpa: *Allocator, s: Slice) !void {
+            try self.ensureUnusedCapacity(gpa, s.len);
+            var i: usize = 0;
+            const slice_marlist = s.toMultiArrayList();
+            while (i < s.len) : (i += 1) {
+                self.appendAssumeCapacity(slice_marlist.get(i));
+            }
+        }
+
         /// Extend the list by 1 element, asserting `self.capacity`
         /// is sufficient to hold an additional item.  Returns the
         /// newly reserved index with uninitialized data.
diff --git a/src/arch/x86_64/CodeGen.zig b/src/arch/x86_64/CodeGen.zig
index 831093a87..e1a7c2551 100644
--- a/src/arch/x86_64/CodeGen.zig
+++ b/src/arch/x86_64/CodeGen.zig
@@ -86,6 +86,10 @@ next_stack_offset: u32 = 0,
 /// Debug field, used to find bugs in the compiler.
 air_bookkeeping: @TypeOf(air_bookkeeping_init) = air_bookkeeping_init,
 
+/// How much we should add to any Data with `inst` tag set, to offset for
+/// instructions that we backpatch.
+add_to_jmp_reloc: u32 = undefined,
+
 const air_bookkeeping_init = if (std.debug.runtime_safety) @as(usize, 0) else {};
 
 pub const MCValue = union(enum) {
@@ -349,12 +353,10 @@ pub fn addExtraAssumeCapacity(self: *Self, extra: anytype) u32 {
 fn gen(self: *Self) InnerError!void {
     const cc = self.fn_type.fnCallingConvention();
     if (cc != .Naked) {
-        // push the callee_preserved_regs that were used
-        const backpatch_push_callee_preserved_regs_i = try self.addInst(.{
-            .tag = .push_regs_from_callee_preserved_regs,
-            .ops = undefined,
-            .data = .{ .regs_to_push_or_pop = undefined }, // to be backpatched
-        });
+        const global_instructions = self.mir_instructions;
+        var local_mir = std.MultiArrayList(Mir.Inst){};
+        defer local_mir.deinit(self.gpa);
+        self.mir_instructions = local_mir;
 
         _ = try self.addInst(.{
             .tag = .push,
@@ -430,22 +432,35 @@ fn gen(self: *Self) InnerError!void {
             }).encode(),
             .data = undefined,
         });
-        // calculate the data for callee_preserved_regs to be pushed and popped
-        var callee_preserved_regs_push_data: u32 = 0x0;
-        inline for (callee_preserved_regs) |reg, i| {
+        self.mir_instructions = global_instructions;
+        // push callee_preserved_regs that were used
+        inline for (callee_preserved_regs) |reg| {
             if (self.register_manager.isRegAllocated(reg)) {
-                callee_preserved_regs_push_data |= 1 << @intCast(u5, i);
+                _ = try self.addInst(.{
+                    .tag = .push,
+                    .ops = (Mir.Ops{
+                        .reg1 = reg,
+                    }).encode(),
+                    .data = undefined,
+                });
             }
         }
-        const data = self.mir_instructions.items(.data);
-        // backpatch the push instruction
-        data[backpatch_push_callee_preserved_regs_i].regs_to_push_or_pop = callee_preserved_regs_push_data;
-        // pop the callee_preserved_regs
-        _ = try self.addInst(.{
-            .tag = .pop_regs_from_callee_preserved_regs,
-            .ops = undefined,
-            .data = .{ .regs_to_push_or_pop = callee_preserved_regs_push_data },
-        });
+        self.add_to_jmp_reloc = @intCast(u32, global_instructions.len);
+        // add the body
+        try self.mir_instructions.appendSlice(self.gpa, local_mir.slice());
+        // pop them
+        inline for (callee_preserved_regs) |reg| {
+            if (self.register_manager.isRegAllocated(reg)) {
+                _ = try self.addInst(.{
+                    .tag = .pop,
+                    .ops = (Mir.Ops{
+                        .reg1 = reg,
+                    }).encode(),
+                    .data = undefined,
+                });
+            }
+        }
+
         _ = try self.addInst(.{
             .tag = .ret,
             .ops = (Mir.Ops{
diff --git a/src/arch/x86_64/Emit.zig b/src/arch/x86_64/Emit.zig
index 7fd81d46c..8f7995474 100644
--- a/src/arch/x86_64/Emit.zig
+++ b/src/arch/x86_64/Emit.zig
@@ -285,7 +285,7 @@ fn mirJmpCall(emit: *Emit, tag: Mir.Inst.Tag, inst: Mir.Inst.Index) InnerError!v
     const ops = Mir.Ops.decode(emit.mir.instructions.items(.ops)[inst]);
     const flag = @truncate(u1, ops.flags);
     if (flag == 0) {
-        const target = emit.mir.instructions.items(.data)[inst].inst;
+        const target = emit.mir.instructions.items(.data)[inst].inst + emit.mir.function.add_to_jmp_reloc;
         const opc: u8 = switch (tag) {
             .jmp => 0xe9,
             .call => 0xe8,
@@ -441,7 +441,7 @@ inline fn getCondOpCode(tag: Mir.Inst.Tag, cond: CondType) u8 {
 
 fn mirCondJmp(emit: *Emit, tag: Mir.Inst.Tag, inst: Mir.Inst.Index) InnerError!void {
     const ops = Mir.Ops.decode(emit.mir.instructions.items(.ops)[inst]);
-    const target = emit.mir.instructions.items(.data)[inst].inst;
+    const target = emit.mir.instructions.items(.data)[inst].inst + emit.mir.function.add_to_jmp_reloc;
     const cond = CondType.fromTagAndFlags(tag, ops.flags);
     const opc = getCondOpCode(tag, cond);
     const source = emit.code.items.len;
```
IMO, we should stick to the backpatch strategy for now, as this is pretty complicated (and still does not work for now).
Also, copying the local slice into the mir_instructions is O(n), which is not ideal.

kubkon: Mhm, sounds good to me! Let's stick to the backpatch strategy for now.

kubkon:
After some debugging and playing around with the code, the problem comes from `genBinMathOp` logic which effectively spills `rcx` onto the stack but then fails to incorporate the value back into the final result. At first glance this looks like an issue with lack of free registers; but then, we have only removed 2 additional registers in this PR so I'm wondering if there is some higher level logic that keeps track of when the result of a binary operator should end up on the stack and when in a register. Anyhow, if you're happy with it @g-w1, I'm happy to take over this PR and work out all the nitty gritty details.

g-w1:
Thanks for debugging this @kubkon! I'd certainly be happy if you took this over.

kubkon:
> Thanks for debugging this @kubkon! I'd certainly be happy if you took this over.

Roger that. If you're about, please hang around during stage2 meetup today - I'll put this PR and general register allocation on the agenda.

kubkon:
Closed by 18a909f61d87d6b8e89147b7b46d8d0785b24b4d. Hoped for it to be closed automatically, but I messed something up evidently.

PR #10141, opened 2021-11-13
CBE: memset(..., 0xaa, ...) undefined values

jmc-88:
This commit makes airStore() handle undefined values directly instead of
delegating to renderValue(): the call to renderValue() happens too late,
when "dest = " has already been written to the stream, at which point
there's no sane way to initialize e.g. struct values by assignment.

Instead, we make airStore() use memset(dest, 0xaa, sizeof(dest)), which
should transparently handle all types.

Also moves the newly-passing tests to the top of test/behavior.zig.


andrewrk:
Nice work!

PR #10143, opened 2021-11-13
Normalized C++ compilation options for single-threaded targets

nuald:
The PR contains two major changes:

1. C++ is compiled without threading support for single-threaded targets (was already partially implemented for WASM, now it's applied to all single-threaded targets).
2. RunStep has been extended to run artifacts with the external executors if enabled.

The reason for (1) is that the current functionality is a little bit confusing as C++ standard library is compiled with the threading support even if an artifact is explicitly flagged to use the single-threaded mode. I've discovered that during dealing with #6573. The original LLVM commit referenced there is being abandoned, and the only way to use C++ for ARM in Zig is using modified standard library, hence this PR.

There are few reason for (2), but none of them are critical, therefore the change could be dropped:

- `.enable_qemu` and similar flags could be applied to the executable, but the corresponding RunStep if invoked would give the `IncorrectExe` error. It's not immediately obvious that the RunStep is applicable for native targets only, and `.enable_qemu` is no-op for executables, but for tests only. In theory, those flags should be applied to tests and run steps only, but it's out of scope of this PR.
- some tests involve running the actual executables, e.g. the integration tests like the `c_compiler` modified in this PR. It's handy to have this ability out-of-box.
- some build processes require cross-platform runs. One particular example here is one of my project that has the pure Zig build process for LuaJIT: `buildvm` should be invoked on the target platform (please see https://github.com/LuaJIT/LuaJIT/issues/664 for the reference). I could successfully compile and use LuaJIT on ARM, but need to use the external executor for `buildvm`. The first commit uses `ExecutorRunStep` that is partially taken from the project.

Testing done by:
 - running the full test suite: `zig build test`
 - running the `c_compiler` test with 3 targets (x86_64 Linux host, arm-linux-musleabi, wasm32-wasi):

```sh
cd test/standalone/c_compiler
zig build test
zig build test -Dtarget=arm-linux-musleabi -Denable-qemu -Dsingle-threaded
zig build test -Dtarget=wasm32-wasi -Denable-wasmtime -Dsingle-threaded
zig build test -Dtarget=x86_64-windows-gnu -Denable-wine
zig build test -Dtarget=x86_64-windows-gnu -Denable-wine -Dsingle-threaded
zig build test -Dtarget=x86_64-macos-gnu -Denable-darling
zig build test -Dtarget=x86_64-macos-gnu -Denable-darling -Dsingle-threaded
```

Additional notes:

- for WASM target the `-fno-exceptions` flag is applied to C++ standard library only, not the final executables. If could be addressed in this or another PR if you want;
- macos doesn't properly support C++ exceptions, additional work is required (probably libunwind and DWARF CFI should be added), and for now I've just disabled testing it; 
- `_LIBCPP_HAS_NO_THREADS` is added for `buildLibCXXABI`  for single-threaded targets too as it uses C++ standard library;
- `test.c/test.cpp` code have mixed spaces/tabs, I've restyled the code for the consistency;
- RunStep.getExternalExecutor is a little bit verbose, but I anticipate that other changes could be required so left it as is for now

### DIFF
--- a/lib/std/build/RunStep.zig
+++ b/lib/std/build/RunStep.zig
@@ -1,7 +1,6 @@
 const std = @import("../std.zig");
 const builtin = @import("builtin");
 const build = std.build;
-const CrossTarget = std.zig.CrossTarget;
 const Step = build.Step;
 const Builder = build.Builder;
 const LibExeObjStep = build.LibExeObjStep;
@@ -143,23 +142,6 @@ pub fn expectStdErrEqual(self: *RunStep, bytes: []const u8) void {
     self.stderr_action = .{ .expect_exact = self.builder.dupe(bytes) };
 }
 
-/// Returns true if the step could be run, otherwise false
-pub fn isRunnable(
-    self: *RunStep,
-) bool {
-    for (self.argv.items) |arg| {
-        switch (arg) {
-            .artifact => |artifact| {
-                _ = self.getExternalExecutor(artifact) catch {
-                    return false;
-                };
-            },
-            else => {},
-        }
-    }
-    return true;
-}
-
 pub fn expectStdOutEqual(self: *RunStep, bytes: []const u8) void {
     self.stdout_action = .{ .expect_exact = self.builder.dupe(bytes) };
 }
@@ -170,57 +152,6 @@ fn stdIoActionToBehavior(action: StdIoAction) std.ChildProcess.StdIo {
         .inherit => .Inherit,
         .expect_exact, .expect_matches => .Pipe,
     };
-}
-
-fn getExternalExecutor(self: *RunStep, artifact: *LibExeObjStep) !?[]const u8 {
-    const need_cross_glibc = artifact.target.isGnuLibC() and artifact.is_linking_libc;
-    const executor = self.builder.host.getExternalExecutor(artifact.target_info, .{
-        .qemu_fixes_dl = need_cross_glibc and self.builder.glibc_runtimes_dir != null,
-        .link_libc = artifact.is_linking_libc,
-    });
-    switch (executor) {
-        .bad_dl, .bad_os_or_cpu => {
-            return error.NoExecutable;
-        },
-        .native => {
-            return null;
-        },
-        .rosetta => {
-            if (self.builder.enable_rosetta) {
-                return null;
-            } else {
-                return error.RosettaNotEnabled;
-            }
-        },
-        .qemu => |bin_name| {
-            if (self.builder.enable_qemu) {
-                return bin_name;
-            } else {
-                return error.QemuNotEnabled;
-            }
-        },
-        .wine => |bin_name| {
-            if (self.builder.enable_wine) {
-                return bin_name;
-            } else {
-                return error.WineNotEnabled;
-            }
-        },
-        .wasmtime => |bin_name| {
-            if (self.builder.enable_wasmtime) {
-                return bin_name;
-            } else {
-                return error.WasmtimeNotEnabled;
-            }
-        },
-        .darling => |bin_name| {
-            if (self.builder.enable_darling) {
-                return bin_name;
-            } else {
-                return error.DarlingNotEnabled;
-            }
-        },
-    }
 }
 
 fn make(step: *Step) !void {
@@ -237,9 +168,6 @@ fn make(step: *Step) !void {
                 if (artifact.target.isWindows()) {
                     // On Windows we don't have rpaths so we have to add .dll search paths to PATH
                     self.addPathForDynLibs(artifact);
-                }
-                if (try self.getExternalExecutor(artifact)) |executor| {
-                    try argv_list.append(executor);
                 }
                 const executable_path = artifact.installed_path orelse artifact.getOutputSource().getPath(self.builder);
                 try argv_list.append(executable_path);
--- a/lib/std/build/RunStep.zig
+++ b/lib/std/build/RunStep.zig
@@ -1,6 +1,7 @@
 const std = @import("../std.zig");
 const builtin = @import("builtin");
 const build = std.build;
+const CrossTarget = std.zig.CrossTarget;
 const Step = build.Step;
 const Builder = build.Builder;
 const LibExeObjStep = build.LibExeObjStep;
@@ -142,6 +143,23 @@ pub fn expectStdErrEqual(self: *RunStep, bytes: []const u8) void {
     self.stderr_action = .{ .expect_exact = self.builder.dupe(bytes) };
 }
 
+/// Returns true if the step could be run, otherwise false
+pub fn isRunnable(
+    self: *RunStep,
+) bool {
+    for (self.argv.items) |arg| {
+        switch (arg) {
+            .artifact => |artifact| {
+                _ = self.getExternalExecutor(artifact) catch {
+                    return false;
+                };
+            },
+            else => {},
+        }
+    }
+    return true;
+}
+
 pub fn expectStdOutEqual(self: *RunStep, bytes: []const u8) void {
     self.stdout_action = .{ .expect_exact = self.builder.dupe(bytes) };
 }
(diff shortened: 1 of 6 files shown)


lib/std/build/RunStep.zig
```diff
@@ -136,6 +137,23 @@ pub fn expectStdErrEqual(self: *RunStep, bytes: []const u8) void {
     self.stderr_action = .{ .expect_exact = self.builder.dupe(bytes) };
 }
 
+/// Returns true if the step could be run, otherwise false
+pub fn isRunnable(
```
kubkon, line 141: I might have missed that, but why is this logic duplicated from [`cross_target.zig`](https://github.com/ziglang/zig/blob/1d55705fa42d83556346f8fcb3a8fce0d55f06ad/lib/std/zig/cross_target.zig#L608)?

lib/std/build/RunStep.zig
```diff
@@ -148,6 +166,42 @@ fn stdIoActionToBehavior(action: StdIoAction) std.ChildProcess.StdIo {
     };
 }
 
+fn getExternalExecutor(artifact: *LibExeObjStep) !?[]const u8 {
```
kubkon, line 169: Again, like in the comment above, why do we need to duplicate `getExternalExecutor` here?

lib/std/build/RunStep.zig
nuald: I'm not quite sure what exactly is duplicated here as CrossTarget declares the executors to use, and this code is just using those. In particular this helper function could be used in the RunStep to check if the step is viable (going to work) to avoid runtime errors.

lib/std/build/RunStep.zig
nuald: It doesn't duplicate the code but the name. I guess I can rename it to avoid confusion. This function utilizes CrossTarget's `getExternalExecutor()`, but besides that have a check if the executor is enabled. It's similar to the logic for [tests](https://github.com/ziglang/zig/blob/master/lib/std/build.zig#L2520).

lib/std/build/RunStep.zig
kubkon: Oh sorry, what I meant to point out is that we already have the ability to run (cross-target) tests that require an emulator or similar, therefore I don't really see the need for this in `RunStep.zig`. See for instance output of `zig test --help`:

```
...
Test Options:
  --test-filter [text]           Skip tests that do not match filter
  --test-name-prefix [text]      Add prefix to all tests
  --test-cmd [arg]               Specify test execution command one arg at a time
  --test-cmd-bin                 Appends test binary path to test cmd args
  --test-evented-io              Runs the test in evented I/O mode
  --test-no-exec                 Compiles test binary without running it
...
```

We already have a way of specifying a custom execution command which is exactly how we invoke foreign tests on some host for instace, like Wasm in Wasmtime.

kubkon:
Just an FYI that C++ exceptions require additional functionality in the zld linker to work on macOS. It's on my roadmap but didn't have time to add that in yet.

lib/std/build/RunStep.zig
nuald: Yes, but as I mentioned in the comment, the current RunStep behavior is inconsistent. I guess if it would be called NativeRunStep then it will be less confusing, but right now, then one specifies that it's possible to use QEMU, but RunStep gives IncorrectExe error, it's not immediately obvious that RunStep works with native target only.

Moreover, sometimes it's required to run the exe (not tests) in the emulator. Few use-cases for myself are:
 - complicated build processes (like LuaJIT compilation I've mentioned in the PR comment). `buildvm` (compiled from the source code using Zig in my toolchain) that is used for building LuaJIT VM doesn't support cross-compilation, and it's required to be invoked with the target architecture (i.e. to produce 32-bit LuaJIT, I have to use 32-bit `buildvm`). It can't be run as a test as it's standalone program and doesn't export anything to be linked into tests. Please take a look to [build.zig](https://gitlab.com/slpm/slpm/-/blob/main/modules/native/luajit/build.zig) for the reference.
 - cross-platform benchmarks. I'm working with the embedded systems, and need to measure the libraries footprint (e.g. pcre vs hyperscan for regexes). The benchmarks are compiled into executables for the deployment on the target platforms, and it's handy to run them first with QEMU to verify file sizes and memory consumption. While I could definitely use shell scripting for that, using extended RunStep functionality provided in this PR, makes the task much easier.

However, as I said, it's not critical as I have the workaround (like `ExecutorRunStep` in the referenced sources above). If you still have doubts, I can just remove it.

nuald:
> Just an FYI that C++ exceptions require additional functionality in the zld linker to work on macOS. It's on my roadmap but didn't have time to add that in yet.

Yes, I've noticed that you read `__eh_info` from `__TEXT` segment instead of `__unwind_info` and even have [a comment](https://github.com/ziglang/zig/blob/master/src/link/MachO.zig#L1486) about that, but the change seems out of scope for this PR. I was planning to fix it later, but macOS support for my projects are not priority and I guess you'll fix it faster :)

lib/std/build/RunStep.zig
kubkon: Ah gotcha. Hmm, ok, I'll need to have @andrewrk take another look at this as we might already have a way to do this that I'm not aware of.

lib/std/build.zig
```diff
@@ -2625,6 +2625,12 @@ pub const LibExeObjStep = struct {
             try zig_args.append(c_macro);
         }
 
+        const single_threaded = self.single_threaded or
+            self.target.getCpuArch().isWasm();
+        if (self.is_linking_libcpp and single_threaded) {
+            try zig_args.append("-D_LIBCPP_HAS_NO_THREADS");
+        }
+
```
andrewrk, line 2633: This define would be better added in `addCCArgs` in Compilation.zig rather than in the build system.

There, the single_threaded flag is already using such logic as:

```zig
        const single_threaded = options.single_threaded or target_util.isSingleThreaded(options.target);
```

```zig
pub fn isSingleThreaded(target: std.Target) bool {
    return target.isWasm();
}
```

Is this define supposed to affect when building libc++ from source, or when including libc++ headers, or both?

test/standalone/c_compiler/build.zig
```diff
@@ -1,20 +1,25 @@
...
-fn isRunnableTarget(t: CrossTarget) bool {
-    if (t.isNative()) return true;
+fn isSingleThreadedTarget(t: CrossTarget) bool {
+    // WASM target will be detected automatically as single threaded
 
-    return (t.getOsTag() == builtin.os.tag and
-        t.getCpuArch() == builtin.cpu.arch);
+    // C++ lib doesn't work on ARM on multi-threaded mode yet:
+    // https://reviews.llvm.org/D75183 (abandoned, used as a ref)
+    return t.getCpuArch() == .arm;
```
andrewrk, line 10: This logic should be moved to Compilation.zig where it deals with the Options struct to affect the choice of `single_threaded`.

If the user passes, for example, `-lc++ -target arm-linux`, then they should get single_threaded false by default. If they do not link libc++ then it should default to multi-threaded (status quo). Finally if the user requests both multi-threaded, libc++, and arm, then they should get an error: `-lc++ -target arm-linux -fno-single-threaded` should cause "error: libc++ does not work on multi-threaded ARM yet".

test/standalone/c_compiler/build.zig
```diff
@@ -23,15 +28,30 @@ pub fn build(b: *Builder) void {
     exe_c.setBuildMode(mode);
     exe_c.setTarget(target);
     exe_c.linkLibC();
+    exe_c.enable_wine = is_wine_enabled;
+    exe_c.enable_qemu = is_qemu_enabled;
+    exe_c.enable_wasmtime = is_wasmtime_enabled;
+    exe_c.enable_darling = is_darling_enabled;
```
andrewrk, line 34: I wonder why I decided to make this per-artifact instead of global to Builder. I think I'd like to explore changing this (myself, I will make changes to master branch), as a prerequisite to merging this PR.

andrewrk (changes requested):
Thank you for these changes, @nuald. I think we can definitely get this to a mergeable point, and it will be a nice improvement. I have a couple of requests which involve moving logic from the build system to `src/Compilation.zig`.

But also I want to make some build system improvements prior to merging this. I hope you don't mind if I go ahead and work on those, and then ask you to rebase this.

lib/std/build.zig
nuald: This particular define affects the build process when including libc++ headers. `src/libcxx.zig` contains the change that uses the same define to fix libc++ compilation.

As for other comments, I'll try to address them and wait for your changes to rebase.

test/standalone/c_compiler/build.zig
nuald: Done, verified by:
```
cd test/standalone/c_compiler
zig build-exe -lc++ -target arm-linux-musleabi test.cpp
zig build-exe -lc++ -target arm-linux-musleabi -fno-single-threaded test.cpp
zig build-exe -lc++ -target arm-linux-musleabi -fsingle-threaded test.cpp
```

Only the latter works, other fail with:
```
warning(compilation): libc++ does not work on multi-threaded ARM yet.
For more details: https://github.com/ziglang/zig/issues/6573
error: unable to create compilation: TargetRequiresSingleThreaded
```

test/standalone/c_compiler/build.zig
```diff
@@ -23,15 +20,30 @@ pub fn build(b: *Builder) void {
...
     exe_cpp.setTarget(target);
-    exe_cpp.linkSystemLibrary("c++");
+    exe_cpp.linkLibCpp();
+    exe_cpp.single_threaded = single_threaded;
+    const os_tag = target.getOsTag();
+    // macos C++ exceptions could be compiled, but not being catched,
+    // additional support is required, possibly unwind + DWARF CFI
+    if (target.getCpuArch().isWasm() or os_tag == .macos) {
+        exe_cpp.defineCMacro("_LIBCPP_NO_EXCEPTIONS", null);
+    }
```
andrewrk, line 40: What's going on here? this define logic belongs in the compiler, not in this build.zig file.

andrewrk (changes requested):

andrewrk:
Alright I did the aforementioned prerequisite changes in 1cac99c90a34bc2cab96d75017a337ddd3479b50. If you don't mind, can you rework your changes against latest master and let's go from there?

test/standalone/c_compiler/build.zig
nuald: As you know, C++ standard library (use older STL acronym later) heavily uses exceptions, and I've updated the test to use them (as it could be a key functionality for C++ applications like in my use-case). Unfortunately, wasm and macOS (currently) has one issue that I'm not quite sure how to overcome: the code is compiled, but doesn't work correctly. I guess the proper solution would be to stop compiling the code if compiler detects that exceptions are used for wasm and macOS, but it means that C++ STL won't be compiled at all on those platforms, and that would be quite breaking change (as STL still could be used with incorrectly working exceptions if used carefully). I didn't want to risk introducing it as it could affect way too many users.

So, for now I've just declared a macro (it could have any name though) that disables exception code in the test. I was going to fix macOS exception handling later if you team won't have time for that, but for now I've just inserted that stub.

nuald:
@andrewrk Could you please re-review the changes?

andrewrk:
Sorry for the delay, I have this in my high priority list for next week.

PR #10148, opened 2021-11-14
io_uring: implement register_files_update

vrischmann:
Implements updating the registered file descriptors (see [this manpage](https://manpages.debian.org/unstable/liburing-dev/io_uring_register.2.en.html#IORING_REGISTER_FILES_UPDATE)). It's equivalent to [io_uring_register_files_update](https://github.com/axboe/liburing/blob/master/src/include/liburing.h#L151-L152) from liburing.

This time I tried to take into account older kernels in the test: as far as I understand sparse entry in the file descriptors array was introduced in 5.5 (see [this comment](https://github.com/ziglang/zig/blob/008b0ec5e58fc7e31f3b989868a7d1ea4df3f41d/lib/std/os/linux/io_uring.zig#L1242-L1243), and according to the manpage `IORING_REGISTER_FILES_UPDATE` was added in 5.5 as well.
The test relies on the fact that `register_files` with a sparse entry will fail on kernels < 5.5 to skip the whole test.

### DIFF
--- a/lib/std/os/linux/io_uring.zig
+++ b/lib/std/os/linux/io_uring.zig
@@ -724,6 +724,37 @@ pub const IO_Uring = struct {
         try handle_registration_result(res);
     }
 
+    /// Updates registered file descriptors.
+    ///
+    /// Updates are applied starting at the provided offset in the original file descriptors slice.
+    /// There are three kind of updates:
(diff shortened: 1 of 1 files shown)


andrewrk:
Another io_uring PR was merged recently; mind rebasing?

vrischmann:
done

PR #10149, opened 2021-11-14
Add f128 support and tests in std.math.exp

LewisGaul:
Sibling PR to https://github.com/ziglang/zig/pull/10046 that adds `f128` support to `exp2()` - this adds support to `exp()`.

---
The CI is failng on `std-i386-linux-gnu-Debug-c-multi`, hitting a segfault when trying to add/subtract two 128-bit floats. This appears to be *uncovered* by my changes, as opposed to *caused by*.

Testing performed locally:
```
$./zig test --zig-lib-dir lib/ lib/std/math/exp.zig
All 13 tests passed.
```

Also tested using property-based testing, comparing random values to GCC's libquadmath (see https://github.com/LewisGaul/zig-f128math/blob/main/hypothesis/test.py).

---
As in the linked PR, some points for review:
- Handling of NaN needs working out and tidying up (just want to match GCC/musl?), but leaving this for now
- May want to reduce repetition in the declaration of the testcases, but can do this separately
- Would it be valuable to add any more information (in comments) about where the code comes from (and/or comments about how it works)? E.g. stuff from notes at https://github.com/LewisGaul/zig-f128math/blob/main/README.md#implementation-of-maths-functions
- Did I get the copyright header from the upstream code correct (is it needed, did I put it in the right place...)?
- May want to move to using latest musl (or other) implementation at some point, which uses a different slightly faster method...


lib/std/math/exp.zig
```diff
@@ -36,7 +43,7 @@ fn exp32(x_: f32) f32 {
     hx &= 0x7FFFFFFF;
 
     if (math.isNan(x)) {
-        return x;
+        return math.nan(f32);
```
tiehuis, line 46: This results in slightly different behavior in that the nan bit-representation returned will not necessarily be the same as the provided nan since there are many possible bit-representations. Probably won't cause an issue in practice but probably not worth changing to avoid surprises.

it's interesting to note the following the C standard

_Recommended practice
13 If a function with one or more NaN arguments returns a NaN result, the result should be the same
as one of the NaN arguments (after possible type conversion), except perhaps for the sign_

tiehuis:
_Did I get the copyright header from the upstream code correct (is it needed, did I put it in the right place...)?_

I previously added the link to the source on the upstream repo for musl ports so that the implementation that it was ported from was known and there was a reference of what license this was originally under. I'm not well versed in best-practice on this but since it wasn't a strict re-distribution I opted for this for brevity.

_May want to move to using latest musl (or other) implementation at some point, which uses a different slightly faster method..._

I think it would be good to use musl for this, solely based on the fact that other math functions are all based on musl (or our own) everywhere. It seems fairly simple as well from what I can see.

lib/std/math/exp.zig
LewisGaul: Yes, you're right, I should revert this change. After opening this PR I performed the property-based testing against GCC and found that this implementation disagrees with GCC (and Musl) on the NaN handling. The reason I changed it is I think the Zig handling was already inconsistent, so I was just trying to make it consistent! I'll take another pass over this PR sometime hopefully soon.

LewisGaul:
Since there seems to be a problem in the CI with 128-bit floats (https://github.com/ziglang/zig/issues/10199) I was thinking I might pivot focus slightly and try to get the testing additions in before the 128-bit support. This way I could take a broader view across all maths functions when coming up with the test structure, and could also look at making the implementations of things more consistent/idiomatic Zig.

I'll aim to get a separate PR up over the next week or so that includes the testing changes from this PR and https://github.com/ziglang/zig/pull/10046, and will then rebase these PRs so that they're simply adding 128-bit support and nothing else.

lib/std/math/exp.zig
LewisGaul: This will be resolved by https://github.com/ziglang/zig/pull/10415

andrewrk:
https://github.com/ziglang/zig/pull/10046#issuecomment-1122985940

Issue #10008, opened 2021-10-23
failing CI


andrewrk:
The CI has been especially unreliable lately. Apologies to everyone who has been opening pull requests and seen false negatives.

I'm in the process of acquiring our own hardware to run the CI stuff on, to increase reliability and make it easier to troubleshoot when a flaky test failure happens.

It's going to be annoying for another week or so and then hopefully we will have something sorted out.

Thanks for understanding, I know it's been really annoying lately for everyone.

andrewrk:
Update: CI is getting closer to ready. Meanwhile, I rebooted the [gotta-go-fast](https://github.com/ziglang/gotta-go-fast/) repository and added some `zig ast-check` benchmarks. Have a look at the README to see the changes of how it works now. The next steps will be:

 * Backfilling some performance measurements
 * Adding html/javascript to ziglang.org/perf which download the data and display interactive graphs
 * Integrating with the new CI system so that performance measurements happen with every commit

andrewrk:
[we have our first self-hosted CI run](https://ci.ziglang.org/ziglang/zig/1)

andrewrk:
x86_64 Linux has now been moved over to ci.ziglang.org. This should already improve stability of the CI and we are still planning on moving more over.

Meanwhile, https://ziglang.org/perf is live. Adjustments are still being made to integrate the page with the rest of the website as well as to the graphs.

RossSmyth:
Having a Zig [Bors](https://github.com/graydon/bors)/[Homu](https://github.com/barosl/homu) may be a good idea so that all PR's are tested as the head of master before merging.

matu3ba:
@RossSmyth The Azure build jobs on Windows are (yet) too unstable to do that unless Windows is excluded, lol. For example [10172](https://dev.azure.com/ziglang/zig/_build/results?buildId=17960&view=results) [10168](https://dev.azure.com/ziglang/zig/_build/results?buildId=17959&view=results) [10161](https://dev.azure.com/ziglang/zig/_build/results?buildId=17959&view=results) [10151](https://dev.azure.com/ziglang/zig/_build/results?buildId=17885&view=results) all failed due to Windows timing out after 6 hours.

andrewrk:
This particular issue is resolved. We have moved some things over to our self-hosted CI, reduced memory usage, and disabled flaky tests, and now CI is in an OK state.

andrewrk:
[...]

PR #10151, opened 2021-11-15
os/linux/io_uring: add link_timeout

hnakamur:
Add method from liburing for [io_uring support for linked timeouts \[LWN.net\]](https://lwn.net/Articles/803932/)
Test code is ported from https://github.com/axboe/liburing/blob/b7159e46bc002c7a174e184d94cfa61552f92e69/test/link-timeout.c#L539-L628

### DIFF
--- a/lib/std/os/linux/io_uring.zig
+++ b/lib/std/os/linux/io_uring.zig
@@ -607,6 +607,34 @@ pub const IO_Uring = struct {
         return sqe;
     }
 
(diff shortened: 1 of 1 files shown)


kprotty:
Seems fine and the failing CI is from the windows portion timing out which is a [separate known issue](https://github.com/ziglang/zig/issues/10008#issuecomment-974476617) so going to merge.

PR #10152, opened 2021-11-15
C backend: fix most cast and all pointer+generics behavior tests

drew-gpf:
This PR adds the necessary functionality so that most of the cast and all of the pointer and generics behavior tests now pass (tested on windows under x86_64 using zig cc). To support this,

-cast_llvm.zig now contains only the test cases that need a lot more work to pass; cast.zig contains every other test case.
-some test cases have been expanded slightly
-128-bit integers can now have their types and constants rendered (including undefined u128/i128 vars in Debug mode)
-additional test cases were introduced for 128-bit ints
-airPtrToInt now casts to the integer type from the pointer type
-airAlloc for an array type will now create a local and not local_ref CValue as the local will already refer to the array, and so referring to it is typically ill-formed
-implement airPtrElemVal/Ptr
-implement ptr_add/sub AIR handlers separately due to the fact that some pointer arithmetic can be UB (specifically, if the LHS is NULL or the arithmetic results in a NULL value). I haven't fully scanned the C standard as to if the alternate method results in UB but it at least doesn't raise errors with UBSan on GCC and Clang.
-fixed a bug where rendering an array's type would result in one extra layer of indirection. I'm not 100% sure this is necessarily the correct thing to do but the previous behavior is definitely incorrect.
-support pointer types holding a u64 constant
-fix how 'zero' and 'one' pointers are rendered; 'one' must be cast to the pointer type, and 'zero' should be the int literal '0' which functions the same as NULL but also works for <, >, etc ordering comparisons (although doing so in the first place raises warnings under -Wpedantic)
-all unsigned int constants are appended with 'u' to avoid warnings if the constant is too large
-airLoad and airStore will properly report an error if an array is used, as opposed to letting the C compiler error out. If fixed then the for.zig behavior tests will also work.
-airStoreUndefined now works with array types (the check in airStore is performed after)
-structFieldPtr now works with array types


doppioandante:
Hi Drew,
thanks for the PR, a few of us on discord have started [this wiki[(https://github.com/ziglang/zig/wiki/C-Backend-Behavioral-Tests-signup-sheet) in order to coordinate the work on CBE behavioral tests.
I have already marked the tests you're working on in this PR as taken. If you want to work on more, it would be nice if you checked their box and put your gh handle along with it. When a test is passing, we just delete the corresponding line.

drew-gpf:
Thanks!

andrewrk:
> -cast.zig now contains only the test cases that need a lot more work to pass; cast_c.zig contains every other test case.

Could you please do it the other way around? `cast.zig` should be the one that works for everything, and `cast_llvm.zig` should be the one that work for LLVM and stage1 but not the C backend.

Also the CI is failing because this PR makes the LLVM backend is exercising a new test case. I believe this was unintentional. Please be very careful when moving behavior tests around.

drew-gpf:
@andrewrk 
> Could you please do it the other way around? `cast.zig` should be the one that works for everything, and `cast_llvm.zig` should be the one that work for LLVM and stage1 but not the C backend.

Sorry about that :-) I'll get that in the next commit.

> Also the CI is failing because this PR makes the LLVM backend exercise a new test case. I believe this was unintentional. Please be very careful when moving behavior tests around- we don't want to accidentally lose test coverage due to a simple copy+paste mistake!

Yeah, it looks like every (quite useful) test case I added ended up breaking the LLVM backend in one way or another. I neglected doing those tests as it's broken for me on Windows to begin with although with WSL I can get something to work (but running the math behavior tests causes link errors with LLD, so I can only test individual files right now).

How should I go about this? Should I segregate these tests into *_c.zig files that uniquely pass for the C backend and stage1 but not the LLVM backend, or should I just forego them?

Thanks for responding.

andrewrk:
If the problem is because of new test cases added, that's great :+1: 

In this case I'd be happy to look at this myself and add a commit to make the new test cases pass in the LLVM backend too. I'm sure it's not too big of a change.

Forgot to mention, nice work!

drew-gpf:
From fixing a bug after finding out setRuntimeSafety isn't honored it looks like the LLVM tests are failing in cast.zig and int128.zig because they aren't initializing undefined variable bytes to 0xAA when set to undefined at runtime; other than that, all other tests pass. In pointers.zig it won't compile because the .one AIR value for pointers isn't yet recognized ("implement const of pointer type '[*c]u8' (Tag.one)").

andrewrk (approved):
Got it, thanks for the info. The changes look good - I've merged this locally and now I'm working on the LLVM backend improvements so that this can be merged.

PR #10153, opened 2021-11-15
Stage2: wasm - Implement the MIR pass

Luukdegram:
This implements the MIR pass for the wasm backend. Right now it re-uses the values of the opcode for each MIR tag, which allows us to easily write its value to memory without having to map from MIR to a wasm opcode. We will however have to add synthetic tags such as debug lines etc. Although not all 256 values are being used, the spec does reserve them and have in the past added new opcodes to those reserved slots. However, an option could be to have synthetic tags use a `secondary_tag` field, as we will need those anyway for some of the wasm opcodes that are beyond `u8`. This has the added benefit that each `MIR` instruction will remain 5 bytes max, as we can keep `Tag` 1 byte, rather than 2.


src/arch/wasm/CodeGen.zig
```diff
@@ -0,0 +1,1710 @@
...
+const LazySrcLoc = Module.LazySrcLoc;
+const link = @import("../../link.zig");
+const TypedValue = @import("../../TypedValue.zig");
+const Air = @import("../../Air.zig");
+const Liveness = @import("../../Liveness.zig");
+const Mir = @import("Mir.zig");
+const Emit = @import("Emit.zig");
+
+/// Wasm Value, created when generating an instruction
+const WValue = union(enum) {
```
kubkon, line 24: A minor nit but do we save much by abbreviating `WasmValue` to `WValue`?

Luukdegram: Back then I abbreviated it to keep it consistent with `MCValue` and `CValue`. Perhaps now is a good time to change it to `WasmValue`. I don't particularly mind either way.

kubkon: I'm fine either way. Just wanted to point out that we don't really save that much after all :-)

kubkon:
Awesome work!

PR #10155, opened 2021-11-15
stage1: Fix caching of LLVM builtin fns

### LINKED ISSUE #10147: Compiler error using `@clz` on `Vector(x, {u32,i32})`
### Remember to search before filing a new report

- [X] I searched for this bug and did not find it in the issue tracker, and furthermore, the title I used above will make this new bug report turn up in the search results for my query.

### Zig Version

0.9.0-dev.1625+d3a099c14

### Steps to Reproduce

```zig
// Same thing happens with i32, but no other type I tried
const Int = u32;
const V = Vector(4, Int); // <-- seems `len` of the vector doesn't matter for the crash

pub fn zeroes() V {
    return [_]Int{0, 0, 0, 0};
}

pub fn main() !void {
    // This works!
    // const v = [_]Int{0, 0, 0, 0};
    const v = zeroes();

    // @ctz and @popCount works
    const x = @clz(Int, v);
    debug.print("{d}", .{x});
}
```

### Expected Behavior

No compiler error.

### Actual Behavior

```
broken LLVM module found: Call parameter type does not match function signature!
  %5 = extractvalue { i32, i1 } %4, 0, !dbg !17320
 <4 x i32>  %7 = call <4 x i32> @llvm.ctlz.v4i32(i32 %5, i1 false), !dbg !17321
trunc source and destination must both be a vector or neither
  %8 = trunc <4 x i32> %7 to i6, !dbg !17321

This is a bug in the Zig compiler.thread 3267800 panic:
```

LemonBoy:
The cache entry must take into account the fact some functions operate on scalar types and some other on vectors of scalar types.
Adding a test case where `@clz(u32)` and `@clz(@Vector(N, u32))` are both used is left as an exercise for the person merging this.

Fixes #10147

### DIFF
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -5175,11 +5175,13 @@ static LLVMValueRef get_int_builtin_fn(CodeGen *g, ZigType *expr_type, BuiltinFn
         n_args = 2;
         key.id = ZigLLVMFnIdCtz;
         key.data.ctz.bit_count = (uint32_t)int_type->data.integral.bit_count;
+        key.data.ctz.vector_len = vector_len;
     } else if (fn_id == BuiltinFnIdClz) {
         fn_name = "ctlz";
         n_args = 2;
         key.id = ZigLLVMFnIdClz;
         key.data.clz.bit_count = (uint32_t)int_type->data.integral.bit_count;
+        key.data.clz.vector_len = vector_len;
     } else if (fn_id == BuiltinFnIdPopCount) {
         fn_name = "ctpop";
         n_args = 1;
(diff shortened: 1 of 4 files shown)


src/stage1/codegen.cpp
```diff
@@ -5197,6 +5199,7 @@ static LLVMValueRef get_int_builtin_fn(CodeGen *g, ZigType *expr_type, BuiltinFn
         n_args = 1;
         key.id = ZigLLVMFnIdBitReverse;
         key.data.bit_reverse.bit_count = (uint32_t)int_type->data.integral.bit_count;
+        key.data.bit_reverse.vector_len = vector_len;
```
matu3ba, line 5202:
```cpp
        struct {
            uint32_t bit_count;
        } bit_reverse;
```
in all_types.hpp needs to be changed or this removed. I guess removing this should work.

PR #10162, opened 2021-11-16
stage2 ARM: Introduce MIR

joachimschmidt557:


src/arch/arm/Mir.zig
```diff
@@ -0,0 +1,218 @@
...
+        ///
+        /// Used by e.g. push
+        register_list: bits.Instruction.RegisterList,
+        /// Debug info: line and column
+        ///
+        /// Used by e.g. dbg_line
+        dbg_line_column: struct {
+            line: u32,
+            column: u32,
+        },
```
Luukdegram, line 182: This makes `Data` 8 bytes instead of 4 as the doc comments above Data suggest.

andrewrk: why not use the commented out comptime block below to assert that the size of Data is what it is expected to be?

joachimschmidt557: Yes, `Data` is supposed to be 8 bytes. However, due to the use of tagged unions such as `bits.Instruction.Operand` etc. this is not the case yet and `Data` is currently 12 bytes large. Should I flatten down `Data` to 8 bytes here or can I do that in a follow-up PR?

andrewrk: Fine by me to do in a follow-up PR, as long as you see it as a to-do item :)

joachimschmidt557: Ok, I corrected the doc comment and added a TODO comment for fixing the size of `Data` in future.

Issue #7265, opened 2020-12-01
consolidate libc and libssp into compiler-rt


andrewrk:
Right now, zig generates a "c" library that has functions such as `memcpy` and `sqrt` because LLVM generates code that on those symbols. I propose to make zig's compiler-rt have *all* the symbols that codegen depends on, regardless of what library will typically provide them. Then our existing trick of using weak exports (which can be overridden by e.g. linking against libc) will make sure we don't provide too many functions. We already use function-sections to omit unused symbols.

The benefits of this are simplicity, which will become even more clear when we have 2 competing backends: the LLVM one and the self-hosted zig one, which will each require a different set of compiler-rt symbols. We may even want to expose a build option to compiler-rt.zig (or maybe a different entry point?) so that compiler-rt source can know which backend it needs to provide symbols for.

The simplicity is not only for zig compiler developers, but for users as well. For example, this would simplify the usage of #7264 - we would not need equivalent CLI flags for these other 2 libs, and users would not need to deal with all 3 things.

Issue #1485, opened 2018-09-07
u128 is largest unsigned type?


catb0t:
```zig
const debug = @import("std").debug;

const uType = u128;

pub fn main () void {
  debug.warn("{} {}", @intCast(uType, @maxValue(uType)), @intCast(uType, r()) * 2);
}

pub fn r () uType {
  return @intCast(uType, 1844674407370955161611); // higher than 2 ** 64
}
```

-> `340282366920938463463374607431768211455 3689348814741910323222`

if you change `uType` to anything higher than `u128` compiling fails with `LLVM ERROR: Unsupported library call operation!` (lower works fine though)

Maximum width for unsigned types isn't mentioned in the docs (anyway there is no explicit list of unsigned types).

Is 128 as max intentional? I might like a u256...

andrewrk:
The short answer to this and #1534  is that we need to send a patch to LLVM to make it emit library calls for this even though the scope of those library calls is outside of the standardized compiler-rt.

andrewrk:
It would be a solution, but ideally we could use upstream without any patches. I think the feature is not so critical that we can't wait for one llvm release cycle (after getting the patch into llvm)

data-man:
Inefficient workaround
```zig
const std = @import("std");
const Int = std.math.big.Int;
const warn = std.debug.warn;

pub fn main() void {
  var alloc = std.heap.direct_allocator;

  const a: u1024 = 1234;
  const b: u1024 = 5678;
  const c = a + b;

  const i = Int.initSet(alloc, c);

  warn("{}\n", i);
}
```

mikdusan:
an upstream bug report looking like there is no planned LLVM solution: https://bugs.llvm.org/show_bug.cgi?id=45649

andrewrk:
We can and should lower to our own lib calls instead of LLVM arithmetic for cases where we know LLVM cannot lower them.

Vexu:
LLVM15 is able to correctly lower addition, substraction, multiplication, all bitwise operations and comparisions for all integer types.

PR #10163, opened 2021-11-17
compiler_rt: export floorl

jcmoyer:
This fixes a linker error when compiling with VS2022 on Windows:

```
zig1.obj : LNK2019: unresolved external symbol floorl referenced in function Sema.coerceInMemoryAllowedPtrs [D:\zig\build\zig.vcxproj]
```

I'm not entirely sure this is the correct place to export or if it should go into one of the conditional branches above.

E: So actually this is the classic C thing where long double isn't necessarily bigger than double, this will probably require some more thought. I'm going to have a closer look at what andrew suggested in IRC (related: #7265) since it looks like there's a better way to write this already in `std/special/c_stage1.zig`.

### DIFF
--- a/lib/std/special/c_stage1.zig
+++ b/lib/std/special/c_stage1.zig
@@ -639,21 +639,6 @@ export fn fmod(x: f64, y: f64) f64 {
     return generic_fmod(f64, x, y);
 }
 
-// TODO add intrinsics for these (and probably the double version too)
-// and have the math stuff use the intrinsic. same as @mod and @rem
-export fn floorf(x: f32) f32 {
-    return math.floor(x);
-}
-export fn floor(x: f64) f64 {
-    return math.floor(x);
-}
-export fn floorl(x: c_longdouble) c_longdouble {
-    if (!long_double_is_f128) {
-        @panic("TODO implement this");
-    }
-    return math.floor(x);
-}
-
 export fn ceilf(x: f32) f32 {
     return math.ceil(x);
 }
--- a/lib/std/special/compiler_rt.zig
+++ b/lib/std/special/compiler_rt.zig
@@ -630,11 +630,31 @@ comptime {
         _ = @import("compiler_rt/atomics.zig");
 
         @export(fmaq, .{ .name = "fmaq", .linkage = linkage });
+        @export(floorf, .{ .name = "floorf", .linkage = linkage });
+        @export(floor, .{ .name = "floor", .linkage = linkage });
+        @export(floorl, .{ .name = "floorl", .linkage = linkage });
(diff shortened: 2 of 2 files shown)


jcmoyer:
Is it correct to delete code from c_stage1.zig as it is moved to compiler_rt or will this cause problems elsewhere?

matu3ba:
@jcmoyer As compiler_rt is intended to match LLVM (see the README), this does not look like the right place to me.
Further not all architectures support floats.

andrewrk (approved):
This looks OK to merge to me. What's the problem as it stands?

andrewrk:
> compiler_rt is intended to match LLVM

It's not intended to match LLVM exactly. It already has extra things for Zig to rely on that are not provided by LLVM's compiler-rt. See  #7265 for more details.

matu3ba:
> It's not intended to match LLVM exactly.

I find it confusing to call the document `compiler_rt.zig`, when it is not describing/abstracting the files in the folder `compiler_rt/` (which is ported and reimplemented from LLVM).
[In llvm you have this](https://compiler-rt.llvm.org/) with a differentiation between builtins, sanitizer runtimes, profile and BlocksRuntime.

A README of the folder structure would be nice and personally I would finding a rename of the folder compiler_rt to `builtins` more clear on what is happening.

LemonBoy:
> @LemonBoy What is your take on this?

IMO compiler-rt must follow LLVM's structure (of course with the only exception of our custom stack probes) which in turn (loosely) follows libgcc, having one extra CU is better than shoving everything into a single huge file (`ir.cpp` flashbacks intensify).

But let's take one step back. You don't want `floorl`, what you want is to get [around this LLVM problem](https://reviews.llvm.org/D112388) by implementing some custom lowering in stage1/stage2 like it's being done for `fmaq`.

#7265 arguments are weak, merging everything together doesn't simplify anything for developers nor for the average user.

andrewrk:
> merging everything together doesn't simplify anything for developers nor for the average user.

one thing is simpler than two things. what's the argument for having two things? is it to follow some standard? but already we see that llvm's compiler-rt, libgcc, and msvc's equivalent are subtly different. the only realistic way to make sure everything works is to provide every symbol ourselves, but allow overriding (which is why we use weak_odr linkage).

plus we have zig-specific stuff that requires non-"standard" symbols, such as math with arbitrary sized integers which we will need to lower to a compiler-rt call of a function that llvm's compiler-rt will never have.  #1485

> what you want is to get [around this LLVM problem](https://reviews.llvm.org/D112388)

Thank you for making that patch upstream btw.

PR #10165, opened 2021-11-17
CBE: add .optional_single_mut_pointer and .optional_single_const_pointer to Type.childType

jmc-88:
Also adds newly-passing behavioral tests `align.zig` and `error.zig` to the top of the list.

PR #10166, opened 2021-11-18
C backend: errors and optionals

Scibuild:
Hi this is my first ever pull request, let alone contributing to the zig compiler, so I'm really sorry if I've made some mistakes.

Both `test/behavior/error.zig` and a modified `test/behavior/optional.zig` are now passing. The only test removed from `optional.zig` and included in `optional_llvm.zig` relied on self referential structs, a bigger problem I will try and work on later.

Changes include:
- finding the child type of optional pointers with `Type.childType`
- equality checking for optional types, e.g. `?i32 != ?i32`
- distinguishing between `.unwrap_errunion_payload_ptr` and `.unwrap_errunion_payload` when taking the address of the operand
- using `isPtrAtRuntime()` over `.zigTypeTag() == .Pointer` in `bitcast` operation, might be more locations where this change is needed

Thank you everyone who made contributing to the C backend a lot easier :)

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -1162,12 +1162,13 @@ fn genBody(f: *Function, body: []const Air.Inst.Index) error{ AnalysisFail, OutO
 
             .slice => try airSlice(f, inst),
 
-            .cmp_eq  => try airEquality(f, inst, .cmp_eq),
             .cmp_gt  => try airBinOp(f, inst, " > "),
             .cmp_gte => try airBinOp(f, inst, " >= "),
             .cmp_lt  => try airBinOp(f, inst, " < "),
             .cmp_lte => try airBinOp(f, inst, " <= "),
-            .cmp_neq => try airEquality(f, inst, .cmp_neq),
+
+            .cmp_eq  => try airEquality(f, inst, "((", "=="),
+            .cmp_neq => try airEquality(f, inst, "!((", "!="),
 
             // bool_and and bool_or are non-short-circuit operations
             .bool_and        => try airBinOp(f, inst, " & "),
@@ -1908,9 +1909,13 @@ fn airBinOp(f: *Function, inst: Air.Inst.Index, operator: [*:0]const u8) !CValue
     return local;
 }
 
-fn airEquality(f: *Function, inst: Air.Inst.Index, op: Air.Inst.Tag) !CValue {
-    if (f.liveness.isUnused(inst))
-        return CValue.none;
+fn airEquality(
+    f: *Function,
+    inst: Air.Inst.Index,
+    negate_prefix: []const u8,
+    eq_op_str: []const u8,
+) !CValue {
+    if (f.liveness.isUnused(inst)) return CValue.none;
 
     const bin_op = f.air.instructions.items(.data)[inst].bin_op;
     const lhs = try f.resolveInst(bin_op.lhs);
@@ -1927,7 +1932,7 @@ fn airEquality(f: *Function, inst: Air.Inst.Index, op: Air.Inst.Tag) !CValue {
         // (A && B)  || (C && (A == B))
         // A = lhs.is_null  ;  B = rhs.is_null  ;  C = rhs.payload == lhs.payload
 
-        try writer.writeAll(if (op == .cmp_eq) "((" else "!((");
+        try writer.writeAll(negate_prefix);
         try f.writeCValue(writer, lhs);
         try writer.writeAll(".is_null && ");
         try f.writeCValue(writer, rhs);
@@ -1944,9 +1949,8 @@ fn airEquality(f: *Function, inst: Air.Inst.Index, op: Air.Inst.Tag) !CValue {
         return local;
     }
 
-    const operator = if (op == .cmp_eq) "==" else "!=";
     try f.writeCValue(writer, lhs);
-    try writer.print("{s}", .{operator});
+    try writer.writeAll(eq_op_str);
     try f.writeCValue(writer, rhs);
     try writer.writeAll(";\n");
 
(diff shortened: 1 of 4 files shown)


Scibuild:
#10165 does one of the things, submitted only minutes before me 😖

test/behavior.zig
```diff
@@ -37,6 +39,7 @@ test {
     _ = @import("behavior/while.zig");
     _ = @import("behavior/this.zig");
     _ = @import("behavior/member_func.zig");
+    _ = @import("behavior/error.zig");
```
jmc-88: Your changes should be enough for `behavior/align.zig` to also pass, give it another look.

src/codegen/c.zig
```diff
@@ -1162,12 +1162,12 @@ fn genBody(f: *Function, body: []const Air.Inst.Index) error{ AnalysisFail, OutO
 
             .slice => try airSlice(f, inst),
 
-            .cmp_eq  => try airBinOp(f, inst, " == "),
+            .cmp_eq  => try airEquality(f, inst, true),
```
miscco, line 1165: true / false are almost always code smells. Passing the operation in is much clearer and you can easily derive the boolean inside the method call

test/behavior/optional_llvm.zig
```diff
@@ -0,0 +1,139 @@
...
+        break :blk2 @as(f64, 5.0);
+    };
+    _ = blk1: {
+        mutable_state += 1;
+        break :blk1 @as(f64, 10.0);
+    } != blk2: {
+        try expect(mutable_state == 2);
+        break :blk2 @as(?f64, 5.0);
+    };
+}
```
miscco, line 139: Should the be an empty line at the end?

miscco: Seems like no is the answer

src/codegen/c.zig
Scibuild: Thank you :) should be good now

test/behavior/optional_llvm.zig
Scibuild: Is this resolved then?

miscco: totally, sorry for the noise

drew-gpf:
I don't think the align.zig tests should be counted as passing because the CBE doesn't handle any alignment. In my PR I thought it passed too but removed it because the output had no mention of alignment anywhere.

Something like this could be used somewhere to see if alignment is implemented correctly (although it might randomly pass if the stack happens to be aligned to 0x1000):
```zig
test "page aligned array on stack" {
    var array align(0x1000) = [_]u8{ 1, 2, 3, 4, 5, 6, 7, 8 };
    try expect(@ptrToInt(&array[0]) & 0x1000 == 0);
}
```
stage1 -OReleaseFast will generate:
image
Which always passes, while the CBE will unfold to (compiled with zig cc -fno-sanitize=undefined -O3):
image
Which usually won't pass because it doesn't try to align the stack pointer since there's no alignas directive on the array in the generated C code.

IMO align.zig should have a test like that since right now incorrect alignment will silently pass tests on a lot of architectures. I'd like to hear everyone's thoughts about this.

matu3ba:
@drew-gpf What do you suggest on how this could be fixed? Testing for the occurrence of a certain value works from my point of view only against concrete and known (C) code.

Scibuild:
We can for now take the `align.zig` behavior test out of this PR, it's probably a bigger issue that wasn't ever really the goal here. Also, I'm not sure why the test started failing with the last commit...

drew-gpf:
Hi @matu3ba, sorry for the late reply. To be honest I'm not entirely sure how to fix the issue other than to add alignment checks for variables aligned far beyond typical natural alignment with the additional stipulation that implementers may want to look at generated code/disassembly to make sure what they did was correct. I think at least a warning should be put somewhere so more people don't make the same mistake.

matu3ba:
> add alignment checks for variables aligned far beyond typical natural alignment with the additional stipulation that implementers may want to look at generated code/disassembly to make sure what they did was correct.

The whole point of the checks is to put them into CI for automatic checks (so no regressions occur) and to not do stuff manually. I think it will be simpler to compare string equality of expected and generated C code for these cases and for the alignment of bitpacked C code.
Probably this should be extended to all special "might be necessary on deviation of standard behavior of C code".

On the upside of this, the tests can be shared with [arocc](https://github.com/Vexu/arocc).

> I think at least a warning should be put somewhere so more people don't make the same mistake.

Warnings are an overall anti-pattern. Its better to provide examples "to copy paste and adapt to use case".

src/codegen/c.zig
```diff
@@ -2517,9 +2562,12 @@ fn airUnwrapErrUnionPay(f: *Function, inst: Air.Inst.Index) !CValue {
         return CValue.none;
     }
 
+    // cannot reliably determine from the return type whether we need to do the addrof ourselves
+    // e.g. *(E!T) -> *T   !=   E!*T -> *T
+    const inst_is_ptr = f.air.instructions.items(.tag)[inst] == .unwrap_errunion_payload_ptr;
```
andrewrk, line 2567: suggestion: pass `inst_is_ptr` as a parameter to this function rather than doing this check again. Better yet, pass `maybe_addrof` instead.

andrewrk (changes requested):
Hi @Scibuild, thanks for the patch! It looks ready to merge, except for one problem:

I count 1 test deleted and 8 tests added. Looks like there are some accidentally duplicated tests.

andrewrk:
@drew-gpf thanks for the additional align.zig behavior test case; I pushed a new test case inspired by it to master branch in e8b99428737c401afeb118fa277da362e903b3ca.

andrewrk:
> @drew-gpf thanks for the additional align.zig behavior test case; I pushed a new test case inspired by it to master branch in [e8b9942](https://github.com/ziglang/zig/commit/e8b99428737c401afeb118fa277da362e903b3ca).

I also made it pass with the LLVM backend in e08b6149ab8fb80ad8fa4983ad3aca8ef3303a9f.

andrewrk:
Thanks @Scibuild!

PR #10168, opened 2021-11-18
std: add `writer` methods on all crypto.hash types

nektro:
This enables many use cases but the one piqued my interest in this is if you're hashing many files in sequence. since the hash types are able to take in a continuous stream of bytes through `.update`, one can setup a `std.fifo.LinearFifo` and hash the files' contents with zero allocations by doing `try fifo.pump(file.reader(), hasher.writer());`

### DIFF
--- a/lib/std/crypto/blake2.zig
+++ b/lib/std/crypto/blake2.zig
@@ -184,6 +184,18 @@ pub fn Blake2s(comptime out_bits: usize) type {
                 r.* ^= v[i] ^ v[i + 8];
             }
         }
+
+        pub const Error = error{};
+        pub const Writer = std.io.Writer(*Self, Error, write);
+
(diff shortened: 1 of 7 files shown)


nektro:
adding this worked too

```zig
fn HashWriter(comptime T: type) type {
    return struct {
        h: *T,

        const Self = @This();
        pub const Error = error{};
        pub const Writer = std.io.Writer(*Self, Error, write);

        fn write(self: *Self, bytes: []const u8) Error!usize {
            self.h.update(bytes);
            return bytes.len;
        }

        pub fn writer(self: *Self) Writer {
            return .{ .context = self };
        }
    };
}```

jedisct1:
No love for Gimli and SipHash?

nektro:
oh good catch, I thought I'd added those too

PR #10172, opened 2021-11-19
std.bounded_array: fix `self` parameter type in `constSlice`

rohlem:
A minimal fix for a correctness issue I noticed, though it didn't lead to a bug for me (yet).
Because `.buffer` is an inline array field, we actually require `self` to be passed as a pointer to return a slice from it.
If the compiler decided to pass the object by copying, we would return a pointer to to-be-destroyed stack memory.

All other methods looked fine to me.
From a performance perspective, `*const Self` might be the saner default to use for `get`, but it's correct as-is, so I refrained from changing that one on that hunch alone.


jedisct1:
Sure!

Thank you @rohlem!

jedisct1 (approved):

PR #10177, opened 2021-11-19
Change Darwin term V const to match Linux

dantecatalfamo:
Currently some terminal constants for Darwin are stored as `std.os.system.VMIN`, where on Linux they're stored as `std.os.system.V.MIN`. This fixes that.

https://github.com/dantecatalfamo/zig/blob/darwin-v-term-const/lib/std/os/linux.zig#L4018


andrewrk:
Thanks!

PR #10188, opened 2021-11-20
stage2: wasm - implement the stack

Luukdegram:
This PR implements the stack for wasm. The stack is currently placed at the end of the read-only data segment, and before the program's memory. In the future we should instead place this at the beginning of the data section, to prevent overwriting the read-only segment. However, before that can be done, we need to remove the offset table and move its logic to the linker instead.
Picture for visualization:
image

The way the stack works is by having a global mutable variable (*not*-exported) that gets initialized at initialization. During codegen, we simply emit the symbol of this global to retrieve the stack pointer's value. In the future, the linker will resolve the actual symbol index by performing relocations. (For now, we hardcode global index 0 as we do not emit any other globals).
Next, whenever we find an `alloc` instruction, we move this stack pointer *downwards* based on the ABI size of the type. The reason we 'grow' downwards is because of 3 reasons:
- The `offset` field of a memory instruction is unsigned which means that for structs we can get the 2nd field, by calculating its offset from the first field.
- We cannot overwrite program data.
- Once we place the stack first, runtimes can properly panic/error on stack overflow as we reach address `0`.

#### Why?
There's 2 reasons why we would want to emulate a stack for wasm:
- Implement wasm's C ABI as specified by https://github.com/WebAssembly/tool-conventions/blob/main/BasicCABI.md
    - This allows us to interop with wasm 'libraries' that have been emit by other compilers than Zig. 
    - So the wasm backend emits the same function signatures for exports/externs as the llvm backend.
- To allow pointers in your Zig program such as:
```zig
pub export fn _start() u32 {
    var x: u32 = 0;

    foo(&x);
    return x;
}

fn foo(x: *u32) void {
    x.* = 2;
}
```


Luukdegram:
Rebased after 05330bbe0ddd3057a1badea92a7823a2b1f4d68a for that sweet, sweet, green CI badge.

andrewrk:
Great work!!

andrewrk:
By the way, how close are we to having this work?

#### foo.zig
```
test "empty" {}
```
#### sh
```
zig test foo.zig -target wasm32-wasi --test-cmd wasmtime --test-cmd-bin
```

Once this works, we can start using the behavior tests for wasm test coverage too!

Luukdegram:
> By the way, how close are we to having this work?
> 
> #### foo.zig
> ```
> test "empty" {}
> ```
> 
> #### sh
> ```
> zig test foo.zig -target wasm32-wasi --test-cmd wasmtime --test-cmd-bin
> ```
> 
> Once this works, we can start using the behavior tests for wasm test coverage too!

Really close, all we need to implement is those 2:
- Add temporary stage2 wasi entry to lib/std/start.zig
- Implement slices in wasm backend

I'll get a PR ready this week.

Issue #10064, opened 2021-10-30
Pointers for NVPTX support


gwenzek:
NVPTX backend is currently Tier 4 in Zig, 
yet there still might be people interested in improving support for it (like myself)
Here I'm sharing a few pointers provided by @Snektron to get started, complemented by information I've glanned from LLVM and Nvidia documentation.

I've also started a branch with an ongoing implementation.
I'm learning both about the Zig compiler and LLVM at the same time so beware ! https://github.com/gwenzek/zig/pull/1

# Adress spaces

PTX format defines a VM for the GPU and an ISA. The VM make uses of different memory spaces.
Address space support has been added to Stage 2, so you'll need to use stage 2.
Only a few generic adress spaces have been defined, but more should be added for Nvidia GPU:

https://github.com/ziglang/zig/blob/70ef9bc75c42ec00e9d4231a2e1f1dca84144748/lib/std/builtin.zig#L172-L177

And defaults here:

https://github.com/ziglang/zig/blob/70ef9bc75c42ec00e9d4231a2e1f1dca84144748/src/target.zig#L573-L577 

(for nvptx you probably want the .constant addrspace for constants for example, also .local for locals, etc
Then convert Zig address spaces to llvm address spaces here:

https://github.com/ziglang/zig/blob/70ef9bc75c42ec00e9d4231a2e1f1dca84144748/src/codegen/llvm.zig#L717

The Zig bindings already have definition for Nvidia memory spaces:
https://github.com/ziglang/zig/blob/70ef9bc75c42ec00e9d4231a2e1f1dca84144748/src/codegen/llvm/bindings.zig#L1283-L1290

To better understand what Zig needs to do you can look at a sample kernel written in LLVM IR:
https://llvm.org/docs/NVPTXUsage.html#the-kernel
_I guess it would also be helpful to be able to generate LLVM IR from an arbitrary kernel 
apparently Clang can do so: https://www.llvm.org/docs/CompileCudaWithLLVM.html

Also one would need to implement support for Cuda special variables gridDim, gridIdx, ... which are stored in specials registers:
https://llvm.org/docs/NVPTXUsage.html#id7
as well as the intrinsic for the block barrier `__syncthreads`

I'm not sure how to do that yet.

Snektron:
[...]

PR #10189, opened 2021-11-20
Add Stage2 support for Nvptx target

gwenzek:
Nvptx aka [NVidia Parallel Thread Execution](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html) is a high-level assembly for Nvidia GPUs.
This PR aims to allow Stage 2 to generate this format, and therefore enable GPU programming in Zig.
This is a follow up on issue #10064 , and contains work presented at last [Zig Rush](https://youtu.be/rvfsWm6TckA?t=5351)

## Overview

To generate the .ptx file we leverage the Nvptx LLVM backend.
The main thing we need to do is to ask LLVM to produce the assembly instead of the binary.
That's why we have a custom Linker that just pass most function calls to LLVM. 
Only we intercept `flushModule` and modify the `Compilation` object to ask for assembly instead of binary.

In NVPTX functions that can be launched from the host (CPU) are named kernel and need to be declared as such.
For this LLVM expects the function to use the `PTX_Kernel` calling convention.
Therefore I added the `.PtxKernel`  calling convention to Zig (both Stage1 and Stage2)

The main trick here is that kernel need to be called from CPU code, so you want to be able to access the signature of the kernel from the CPU code. That's why I convert the calling convention `.PtxKernel` into `.Fast` when compiled for another target.
The function itself will still probably be impossible to compile, but at least the CPU code can import it.

## Help needed

I was only able to implement the CC conversion for stage1 using `get_llvm_cc`.
But in Stage2, implementing the equivalent in `toLlvmCallConv` doesn't work as expected because
it doesn't change the `.cc` attribute of the function. Is it possible ? Or is it a better place to do that ?
Right now I need to compile device code with Stage2 and host code with Stage1.

## Demo 

You can find sample code that leverage this PR here:

Kernel (compiled with stage2 for Nvptx target): https://github.com/gwenzek/cudaz/blob/4dd5a6b2eef966afa11135fbe286e51c1fe5056d/CS344/src/hw1_pure_kernel.zig#L5

Caller code (compiled with stage1 for : https://github.com/gwenzek/cudaz/blob/4dd5a6b2eef966afa11135fbe286e51c1fe5056d/CS344/src/hw1_pure.zig#L44

Note that PTX has a lot of intrinsics but for now I didn't need to add them to Zig language, because I can use Zig inline assembly to generate the corresponding PTX code directly.

Tagging @Snektron that helped me previously on this topic.

### DIFF
--- a/src/link/NvPtx.zig
+++ b/src/link/NvPtx.zig
@@ -1,0 +1,115 @@
+//! NVidia PTX (Paralle Thread Execution)
+//! https://docs.nvidia.com/cuda/parallel-thread-execution/index.html
+//! For this we rely on the nvptx backend of LLVM
+//! Kernel functions need to be marked both as "export" and "callconv(.PtxKernel)"
+
+const NvPtx = @This();
+
+const std = @import("std");
+const builtin = @import("builtin");
+
+const Allocator = std.mem.Allocator;
+const assert = std.debug.assert;
+const log = std.log.scoped(.link);
+
+const Module = @import("../Module.zig");
+const Compilation = @import("../Compilation.zig");
+const link = @import("../link.zig");
+const trace = @import("../tracy.zig").trace;
+const build_options = @import("build_options");
+const Air = @import("../Air.zig");
+const Liveness = @import("../Liveness.zig");
+const LlvmObject = @import("../codegen/llvm.zig").Object;
+
+base: link.File,
+llvm_object: *LlvmObject,
+
+pub fn createEmpty(gpa: Allocator, options: link.Options) !*NvPtx {
+    const nvptx = try gpa.create(NvPtx);
+    nvptx.* = .{
+        .base = .{
+            .tag = .nvptx,
+            .options = options,
+            .file = null,
+            .allocator = gpa,
+        },
+        .llvm_object = undefined,
+    };
+
+    if (!build_options.have_llvm) return error.TODOArchNotSupported;
+
+    switch (options.target.cpu.arch) {
+        .nvptx, .nvptx64 => {},
+        else => return error.TODOArchNotSupported,
+    }
+
+    switch (options.target.os.tag) {
+        // TODO: does it also work with nvcl ?
+        .cuda => {},
+        else => return error.TODOOsNotSupported,
+    }
+
+    return nvptx;
+}
+
+pub fn openPath(allocator: Allocator, sub_path: []const u8, options: link.Options) !*NvPtx {
+    assert(options.object_format == .nvptx);
+    if (!build_options.have_llvm or !options.use_llvm) return error.NvptxRequiresLlvm;
+
+    const nvptx = try createEmpty(allocator, options);
+    errdefer nvptx.base.destroy();
+    log.info("Opening .ptx target file {s}", .{sub_path});
+    nvptx.llvm_object = try LlvmObject.create(allocator, options);
+    return nvptx;
+}
+
+pub fn deinit(self: *NvPtx) void {
+    self.llvm_object.destroy(self.base.allocator);
+}
+
+pub fn updateFunc(self: *NvPtx, module: *Module, func: *Module.Fn, air: Air, liveness: Liveness) !void {
+    try self.llvm_object.updateFunc(module, func, air, liveness);
+}
+
+pub fn updateDecl(self: *NvPtx, module: *Module, decl: *Module.Decl) !void {
+    return self.llvm_object.updateDecl(module, decl);
+}
+
+pub fn updateDeclExports(
+    self: *NvPtx,
+    module: *Module,
+    decl: *const Module.Decl,
+    exports: []const *Module.Export,
+) !void {
+    if (build_options.skip_non_native and builtin.object_format != .nvptx) {
+        @panic("Attempted to compile for object format that was disabled by build configuration");
+    }
+    return self.llvm_object.updateDeclExports(module, decl, exports);
+}
+
+pub fn freeDecl(self: *NvPtx, decl: *Module.Decl) void {
+    return self.llvm_object.freeDecl(decl);
+}
+
+pub fn flush(self: *NvPtx, comp: *Compilation) !void {
+    return self.flushModule(comp);
+}
+
+pub fn flushModule(self: *NvPtx, comp: *Compilation) !void {
+    if (build_options.skip_non_native or !build_options.have_llvm) {
+        @panic("Attempted to compile for architecture that was disabled by build configuration");
+    }
+    const tracy = trace(@src());
+    defer tracy.end();
+
+    var hack_comp = comp;
+    if (comp.bin_file.options.emit) |emit| {
+        hack_comp.emit_asm = .{
+            .directory = emit.directory,
+            .basename = comp.bin_file.intermediary_basename.?,
+        };
+        hack_comp.bin_file.options.emit = null;
+    }
+
+    return try self.llvm_object.flushModule(hack_comp);
+}
--- a/src/link/NvPtx.zig
+++ b/src/link/NvPtx.zig
@@ -25,6 +25,8 @@ base: link.File,
 llvm_object: *LlvmObject,
 
 pub fn createEmpty(gpa: Allocator, options: link.Options) !*NvPtx {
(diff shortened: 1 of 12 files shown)


Snektron:
> The main trick here is that kernel need to be called from CPU code, so you want to be able to access the signature of the kernel from the CPU code. That's why I convert the calling convention .PtxKernel into .Fast when compiled for another target.
The function itself will still probably be impossible to compile, but at least the CPU code can import it.

I don't really think that this is the right approach - a calling convention valid for only one target should not be silently translated into another calling convention. It seems to me from your examples that this is not actually required for any nvptx functionality, but is rather used for generating an interface and for calling a debug-version of the kernel compiled to CPU code. 

Would this also be resolved by implementing something as follows?
```zig
const kernel_cc: CallingConvention = if (target.cpu.arch == .nvptx) .PtxKernel else .Unspecified;

pub export fn rgba_to_greyscale(rgbaImage: []u8, greyImage: []u8) callconv(kernel_cc) void {
  ...
}
```

This still leaves the problem where you want to access the kernel prototype without emitting the cpu-kernel in the binary, or actually having the compiler analyze it at all. One solution might be to make the compiler not analyze a function's body if only the prototype is required (and the prototype does not depend on the function's body, such as with generics or inferred error sets). Kernels can then be conditionally exported using `@export`, depending on the target. I do not know how that currently works though.

src/Compilation.zig
```diff
@@ -1890,6 +1890,7 @@ fn emitOthers(comp: *Compilation) !void {
         if (out.emit) |loc| {
             if (loc.directory) |directory| {
                 const src_path = try std.fmt.allocPrint(comp.gpa, "{s}{s}", .{ basename, out.ext });
+                std.log.warn("Will copy {s} to {s}", .{ src_path, directory.handle });
```
Snektron, line 1893: Stray debug print?

src/stage1/analyze.cpp
```diff
@@ -1001,6 +1002,7 @@ bool calling_convention_allows_zig_types(CallingConvention cc) {
         case CallingConventionUnspecified:
         case CallingConventionAsync:
         case CallingConventionInline:
+        case CallingConventionPtxKernel:
```
Snektron, line 1005: Zig types have no ABI layout, and since PtxKernel is not a zig-internal calling convention this should return false at least when compiling for nvptx.

src/link/NvPtx.zig
```diff
@@ -0,0 +1,115 @@
...
+    const tracy = trace(@src());
+
+    var hack_comp = comp;
+    if (comp.bin_file.options.emit) |emit| {
+        hack_comp.emit_asm = .{
+            .directory = emit.directory,
+            .basename = self.llvm_object.sub_path,
+        };
+        hack_comp.bin_file.options.emit = null;
+    }
```
Snektron, line 111: Another option is to just override the values for `emit_bin` and `emit_asm` set in main.zig, though that might get just as messy. What does clang even do when you do not pass `-femit-asm`?

gwenzek: so without `-femit-asm` Clang will create a .o object that contains:

* the .ptx embedded in a binary format (.fatbin)
* some functions to init the Cuda module that mostly delegate to `cudart` (Cuda Runtime library) 
* stubs for the device functions

But note this is a Clang thing, LLVM doesn't know about cudart, and will fail if you pass it `emit-bin`: "LLVM failed to emit asm=(none) bin=kernel.ptx.ptx ir=(none) bc=(none): TargetMachine can't emit an object file"

src/Compilation.zig
gwenzek: fixed.

src/stage1/analyze.cpp
gwenzek: I'm not sure if it's feasible, but I'd really like to be able to have GPU code with Zig objects as argument, like slices.

PTX doesn't enforce one calling convention: https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#abstracting-abi
You can annotate the function parameters if they need to be read from registers.
The ptx also contains the size of each argument. 
So it may generate .ptx files that would be hard to launch from C, but I think this should be handled separately.
I'm fine if the .ptx is only possible to use with a specific version of Zig, as I'm likely to embed it inside the executable anyway.

gwenzek:
> I don't really think that this is the right approach. [...] It seems to me from your examples that this is not actually required for any nvptx functionality, but is rather used for generating an interface and for calling a debug-version of the kernel compiled to CPU code.

Actually one of my intermediate versions was using `const kernel_cc: CallingConvention = if (target.cpu.arch == .nvptx) .PtxKernel else .Unspecified;` on the client side. 
But it become mandatory because you'll need some CPU code to start the kernel, 
and for that you need to know the kernel signature, and you need to import it. 
And since compiling a `callconv(.PtxKernel)` for any target beside Nvptx doesn't make sense, I thought it would be better to nicely import it, instead of yielding a compile error.

So do you think it could be a valid use case of having a calling convention that have a different meaning on different target  ? (.Vectorcall seems to be a precedent)

Otherwise I may not even need a Calling convention at all. I just want to mark some of the function as "entry_point" in the `.ptx`.
LLVM generate this only for exported but it forces you to chose a calling convention. So I had to create a new one, to still be able to pass Zig objects. So if there is another way to annotate function as "entry_point" in Zig, I may not have to specify the CC/nor the export.

gwenzek:
I probably hasn't been really clear in my last post, and I got more time to think about this.

First what do we mean by "Ptx" target ? 
1. we want to generate a `.ptx` file that can be called by any language using the cuda C API
2. we want to generate an adhoc CPU program that can interact with the GPU

For instance Clang went with option 2. The generation of the `.ptx` is only an intermediate step of compiling a larger CPU program with GPU acceleration. The `.ptx` is embedded in the binary and isn't visible from the user. 
This approach has the benefit of not having to think about ABI: you only need the GPU code and the CPU code to agree on memory layout. The downside is that there is more work for the compiler, [Clang actually run two compilations passes](https://www.llvm.org/docs/CompileCudaWithLLVM.html#compilation-models). Some of the code will be compiled both for the CPU and the GPU. This allows to check that the GPU code is launched with correct parameter types.

My original approach was more like option 1: generate a `.ptx` file as a target, and `@embed` to include the `.ptx` in another compilation unit. And I think this is the approach you're favoring @Snektron IIUC.

But since I want to have a type-safe way of launching the kernels I also want to `@import`  the zig code targeted for the device. And I also want to be able to pass Zig structs between CPU/GPU, so my goal is more to provide a user experience similar to Clang to the user  (option 2). 
The good thing is my implementation only need minimal changes to the Zig compiler (this PR) to allow generating the `.ptx`, and some library code which doesn't need to be part of Zig.

So to get back to the two questions raised by @Snektron : 

## should we allow the .PtxKernel calling convention in code compiled for x86 ?

* For me the CPU code always need to know what are the signature of the GPU functions. So we will always want to "import" the device code into the CPU compilation unit. (The alternative being to manually declare the signature for each kernel.) But it raises the question of how to handle function with this calling convention on CPU.

## should we allow the .PtxKernel to use Zig object ?

* For me yes. For example Pytorch has lot of GPU functions that directly takes their Tensor type as input which provide a convenient  way to read/write to them. Having to go back to raw pointers would be a serious drawback

* If users want to have a `.ptx` compatible with other languages then they can just restrict their function to C types or use `extern struct` which have a guaranteed layout. We can introduce a `.PtxKernelC` calling convention if we want to be more explicit.

gwenzek:
I don't understand what's going on there. I can't reproduce the CI issues locally.

gwenzek:
Youhou tests passes!  

Previous version had surreptitiously introduced a "break" statement in a switch statement that was only triggering for other archs. Similarly I needed to put all usage of LLVMObject behind a comptime guard to avoid Zig to analyse it when not building with LLVM. (Thanks Meghan to point me to that)

To come back to the design questions, in stage 2 meeting we agreed to:
* restrict PtxKernel call conv to the nvptx targets,
* pass arbitrary Zig structs to those kernels

Vexu (approved):

gwenzek:
Thanks for the review!

PR #10190, opened 2021-11-20
stage1: improve packed struct array padding error message

scheibo:
I've been getting the `"array of 'Foo' not allowed in packed struct due to padding bits"` error message a bunch during development and usually end up adding a `@compileLog(@sizeOf(Foo), @bitSizeOf(Foo))` to my code before rerunning the compiler to have it tell me what it doesn't like. Given that the compiler has this information already I'm hoping we can expose it through the error message to begin with. Extremely indifferent to the wording of the hint (open to suggestions!), I'm just hoping to expose the two sizes.

(edit: failing CI test is same general OOM that is already failing at HEAD)


andrewrk:
Thanks!

PR #10195, opened 2021-11-21
feat(uefi): add virtual addressing helpers

stakach:
based off definitions in https://uefi.org/sites/default/files/resources/UEFI%20Spec%202_6.pdf


andrewrk:
Thanks!

PR #10197, opened 2021-11-21
io_uring: implement automatic buffer selection

vrischmann:
This implements [automatic buffer selection](https://lwn.net/Articles/815491/) for `io_uring`.

The current api of `read` and `recv` is not great when using automatic buffer selection, we have to write relatively ugly code like this:
```zig
var slice: []u8 = "";
slice.len = 128;

var sqe = try ring.read(0xdfdfdfdf, fd, slice[0..], 0);
sqe.flags |= linux.IOSQE_BUFFER_SELECT;
sqe.buf_index = group_id;
```
because `read` takes a slice and gives its `ptr` and `len` to the kernel via [these fields](https://github.com/ziglang/zig/blob/master/lib/std/os/linux.zig#L3503-L3504).

This is the reason why I added `read_raw` and `recv_raw` which take a `?[*]u8` and a separate length, so we can write code like this instead:
```zig
var sqe = try ring.read_raw(0xdfdfdfdf, fd, null, buffer_len, 0);
sqe.flags |= linux.IOSQE_BUFFER_SELECT;
sqe.buf_index = group_id;
```
which is better imo.

I'm not 100% happy with this API either, because:
* it feels redundant given the super small code difference between the two methods
* future operations which take a buffer as input and support automatic buffer selection will also need to implement two different methods.

There are other solutions but they require breaking existing code using `read` and `recv`.

1)  define `read` the same as `read_raw`
```zig
pub fn read(self: *IO_Uring, user_data: u64, fd: os.fd_t, buffer: ?[*]u8, len: usize, offset: u64) !*io_uring_sqe
```
and let the caller give the correct pointer/len.

2) we could add a buffer type, something like this:
```zig
pub const ReadBuffer = union(enum) {
    slice: []u8,
    buffer_select: usize,
};

pub fn read(self: *IO_Uring, user_data: u64, fd: os.fd_t, buffer: ReadBuffer, offset: u64) !*io_uring_sqe

var buffer: [128]u8 = undefined;
var sqe = try ring.read(0, fd, .{ .slice = &buffer }, 0);

var sqe = try ring.read(0, fd, .{ .buffer_select = buffer_len }, 0);
```

Any input on this is appreciated, I might have missed something obvious.

### DIFF
--- a/lib/std/os/linux/io_uring.zig
+++ b/lib/std/os/linux/io_uring.zig
@@ -1990,53 +1990,15 @@ test "accept/connect/send/recv" {
     };
     defer ring.deinit();
 
-    const address = try net.Address.parseIp4("127.0.0.1", 3131);
-    const kernel_backlog = 1;
-    const server = try os.socket(address.any.family, os.SOCK.STREAM | os.SOCK.CLOEXEC, 0);
-    defer os.close(server);
-    try os.setsockopt(server, os.SOL.SOCKET, os.SO.REUSEADDR, &mem.toBytes(@as(c_int, 1)));
-    try os.bind(server, &address.any, address.getOsSockLen());
-    try os.listen(server, kernel_backlog);
+    const socket_test_harness = try createSocketTestHarness(&ring);
+    defer socket_test_harness.close();
 
     const buffer_send = [_]u8{ 1, 0, 1, 0, 1, 0, 1, 0, 1, 0 };
     var buffer_recv = [_]u8{ 0, 1, 0, 1, 0 };
 
-    var accept_addr: os.sockaddr = undefined;
-    var accept_addr_len: os.socklen_t = @sizeOf(@TypeOf(accept_addr));
-    _ = try ring.accept(0xaaaaaaaa, server, &accept_addr, &accept_addr_len, 0);
-    try testing.expectEqual(@as(u32, 1), try ring.submit());
-
-    const client = try os.socket(address.any.family, os.SOCK.STREAM | os.SOCK.CLOEXEC, 0);
-    defer os.close(client);
-    _ = try ring.connect(0xcccccccc, client, &address.any, address.getOsSockLen());
-    try testing.expectEqual(@as(u32, 1), try ring.submit());
-
-    var cqe_accept = try ring.copy_cqe();
-    if (cqe_accept.err() == .INVAL) return error.SkipZigTest;
-    var cqe_connect = try ring.copy_cqe();
-    if (cqe_connect.err() == .INVAL) return error.SkipZigTest;
-
-    // The accept/connect CQEs may arrive in any order, the connect CQE will sometimes come first:
-    if (cqe_accept.user_data == 0xcccccccc and cqe_connect.user_data == 0xaaaaaaaa) {
-        const a = cqe_accept;
-        const b = cqe_connect;
-        cqe_accept = b;
-        cqe_connect = a;
-    }
-
-    try testing.expectEqual(@as(u64, 0xaaaaaaaa), cqe_accept.user_data);
-    if (cqe_accept.res <= 0) std.debug.print("\ncqe_accept.res={}\n", .{cqe_accept.res});
-    try testing.expect(cqe_accept.res > 0);
-    try testing.expectEqual(@as(u32, 0), cqe_accept.flags);
-    try testing.expectEqual(linux.io_uring_cqe{
-        .user_data = 0xcccccccc,
-        .res = 0,
-        .flags = 0,
-    }, cqe_connect);
-
-    const send = try ring.send(0xeeeeeeee, client, buffer_send[0..], 0);
+    const send = try ring.send(0xeeeeeeee, socket_test_harness.client, buffer_send[0..], 0);
     send.flags |= linux.IOSQE_IO_LINK;
-    _ = try ring.recv(0xffffffff, cqe_accept.res, .{ .buffer = buffer_recv[0..] }, 0);
+    _ = try ring.recv(0xffffffff, socket_test_harness.server, .{ .buffer = buffer_recv[0..] }, 0);
     try testing.expectEqual(@as(u32, 2), try ring.submit());
 
     const cqe_send = try ring.copy_cqe();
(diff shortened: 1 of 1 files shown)


andrewrk:
This old PR had an Azure CI failure and the log was deleted since it is so old, so I rebased this against master.

lib/std/os/linux/io_uring.zig
```diff
@@ -857,6 +895,41 @@ pub const IO_Uring = struct {
...
+    /// Returns a pointer to the SQE.
+    ///
+    /// Provided buffers can be used in `read`, `recv` or `recvmsg` commands by setting the IOSQE_BUFFER_SELECT flag on the SQE.
+    ///
+    /// The kernel expects a contiguous block of memory of size (num * buffer_len).
+    pub fn provide_buffers(
+        self: *IO_Uring,
+        user_data: u64,
+        buffers: [*]u8,
+        num: usize,
```
jorangreef, line 908:
```suggestion
        buffers_count: usize,
```

lib/std/os/linux/io_uring.zig
```diff
@@ -1508,6 +1592,28 @@ pub fn io_uring_prep_linkat(
     sqe.rw_flags = flags;
 }
 
+pub fn io_uring_prep_provide_buffers(
```
jorangreef, line 1595: Nice to see the same prep method names coming through from liburing.

lib/std/os/linux/io_uring.zig
```diff
@@ -522,6 +541,25 @@ pub const IO_Uring = struct {
...
+    /// The only difference with `recv` is that this takes a nullable many-item pointer and a length separately.
+    /// This lets the caller pass null as the buffer, which is required when using the automatic buffer selection.
+    pub fn recv_raw(
+        self: *IO_Uring,
+        user_data: u64,
+        fd: os.fd_t,
+        buffer: ?[*]u8,
+        len: usize,
+        flags: u32,
+    ) !*io_uring_sqe {
```
jorangreef, line 556:
```suggestion
    pub fn recv_with_automatic_buffer_selection(
        self: *IO_Uring,
        user_data: u64,
        fd: os.fd_t,
        group_id: usize,
        flags: u32,
    ) !*io_uring_sqe {
```

@vrischmann What do you think if this interface is explicitly only used for automatic buffer selection, not for normal recv/read calls where buffers are provided, since passing by slice there provides more safety in those cases?

This could then also set the `IOSQE_BUFFER_SELECT` in the prep routine, to ensure there's no confusion about where this should be done. I know this is different to how liburing does it, but I think that makes sense for C, and perhaps not Zig, which has richer types.

This interface was the one thing I wasn't sure about when I did the original io_uring PR.

kprotty: Another idea for consideration: have read/recv take a `Buffer` type which is
```zig
union(enum) {
  bytes: []u8,
  iovecs: []const os.iovec,
  buffer_group_id: usize,
}
```
The variant would then call the correct prep() function. Not sure if this is any better than just having separate functions.

vrischmann: Thanks for your feedback.

I like the idea of using a union for the buffer type, I was hesitant to do this because it will break all existing code using io_uring. If it's acceptable I'll get back into this PR and look at implementing it.

vrischmann:
I changed the PR, now `IO_Uring.read()` takes a `ReadBuffer` which can be used for simple reads or reads using buffer selection.

Both `IO_Uring.readv()` and `IO_Uring.read_fixed()` could be handled in the same way but it should be maybe go in another PR ?

jorangreef (commented):
Perhaps it makes sense to do the breaking changes in one go? Would be really nice to get `recv()` into this PR since it's so similar to `read()`.

vrischmann:
> Perhaps it makes sense to do the breaking changes in one go? Would be really nice to get `recv()` into this PR since it's so similar to `read()`.

Ok, I can do that.

vrischmann:
I've added support for buffer selection for `recv`.
I've also removed the `readv` method; `read` can now call `io_uring_prep_readv` for `ReadBuffer.iovecs`.

This means that I have to handle the `iovecs` case in `recv` and there's no API to use a `iovec` for `recv` so I simply make a slice from the first buffer in the `iovec`. 
Another option might be to just return a "not supported" error ?

lib/std/os/linux/io_uring.zig
```diff
@@ -2843,3 +2811,453 @@ test "linkat" {
...
+        os.close(self.listener);
+    }
+};
+
+fn createSocketTestHarness(ring: *IO_Uring) !SocketTestHarness {
+    // Create a TCP server socket
+
+    const address = try net.Address.parseIp4("127.0.0.1", 3131);
+    const kernel_backlog = 1;
+    const listener_socket = try os.socket(address.any.family, os.SOCK.STREAM | os.SOCK.CLOEXEC, 0);
```
kprotty, line 3216: For ceremony's sake, could add `errdefer os.closeSocket(listener_socket)`. Ditto for `client` socket below.

vrischmann: Thanks, done

jorangreef:
> I've added support for buffer selection for recv.
> I've also removed the readv method; read can now call io_uring_prep_readv for ReadBuffer.iovecs.

Thanks for doing this!

> This means that I have to handle the iovecs case in recv and there's no API to use a iovec for recv
> so I simply make a slice from the first buffer in the iovec.
> Another option might be to just return a "not supported" error ?

Rather than using `ReadBuffer` also for `recv()`, I think it would be better to use a separate, more restricted `RecvBuffer` type to eliminate this case completely.

vrischmann:
> Rather than using ReadBuffer also for recv(), I think it would be better to use a separate, more restricted RecvBuffer type to eliminate this case completely.
Ok makes sense, I'll change it

vrischmann:
I've pushed an update which uses the `RecvBuffer` type for `recv()`.

andrewrk:
Thanks @vrischmann and reviewers!

PR #10202, opened 2021-11-22
libc: update macOS headers to macOS 12.0 Monterey

### LINKED ISSUE #10200: aarch64-macos-gnu, any-macos-any headers need to be updated for macOS 12.0 Monterey
### Zig Version

0.9.0-dev.1696+ff9a44a9f

### Steps to Reproduce

I believe we need to regenerate the macOS libc headers to include changes from macOS 12.0 Monterey.

---

(1) libc headers for aarch64-macos-gnu [report only macOS 11.1 being available](https://github.com/ziglang/zig/blob/b644d493658d8fe4eddeeb3ec6d13593f89739bb/lib/libc/include/aarch64-macos-gnu/AvailabilityVersions.h#L61) (12.0 Monterey is the newest), which can cause issues when trying to compile some code. ~~I believe I'm facing this in https://github.com/hexops/mach/issues/84~~

(2) [`sys/cdefs.h`](https://github.com/ziglang/zig/blob/master/lib/libc/include/any-macos-any/sys/cdefs.h) currently does not define these as empty, which the latest header file from the macOS 12.0 SDK does:

```
#define __kernel_ptr_semantics
#define __kernel_data_semantics
#define __kernel_dual_semantics
``` 

This makes it impossible to supply macOS 12.0 frameworks and compile their headers, as e.g. definitions in `IOKit/IOTypes.h` become syntactically invalid:

```
typedef mach_vm_address_t       IOVirtualAddress __kernel_ptr_semantics;
```

### Expected Behavior

The headers reflect macOS 12

### Actual Behavior

They do not reflect macOS 12

slimsag:
updates libc headers for macos to reflect the latest version present on macOS 12 Monterey

This is the result of running the following after https://github.com/ziglang/fetch-them-macos-headers/pull/15:

```
rm -rf lib/libc/include/any-macos-any lib/libc/include/x86_64-macos-gnu lib/libc/include/aarch64-macos-gnu
fetch_them_macos_headers generate
```

Fixes #10200


slimsag:
Although I am confident this PR does what it describes, I wanted to call out a question: is this the right solution?

https://github.com/ziglang/zig/issues/10200 makes a clear case for why we need the latest macOS libc headers (without them, we cannot link frameworks from the latest macOS SDK when cross compiling, only when building natively as Zig uses the system libc in that case I think)

At the same time Andrew makes a case in [this issue](https://github.com/ziglang/fetch-them-macos-headers/issues/2#issuecomment-729144735) for why we should only use the minimum supported macOS version's headers.

A different solution might be to say something like this:

> If you intend to provide a Framework for Zig to compile, it must either be compatible with the lowest macOS version supported by Zig OR you must also provide the macOS SDK version you wish to target

slimsag:
I believe I no longer _need_ this change for my work, I can stick on the older version of the macOS SDK I believe, so we can defer making a decision here.

---

I think the issue is still valid and worth figuring out, however. I see these potential paths forward:

1. **Do nothing: ship the lowest-supported macOS version** headers (and framework stubs in the future). By doing this, using some newer frameworks _when cross compiling_ is not possible.
    * For example, it would not be possible to use [SharePlay/GroupActivities (added in macOS 12), or new Metal/CreateML features](https://developer.apple.com/macos/whats-new/), _when cross-compiling._
    * One particularly nasty problem here is answering "what is the lowest-supported macOS version"? I think ideally we track what has most market share, but there are no public numbers on this that seem reliable / up-to-date anywhere.
    * Another problem with this is that macOS tends to lag behind iOS/iPadOS/etc, and we must tie all of these together.
2. **Ship the absolute latest macOS version.**
    * The upside here is you can always use the latest APIs/frameworks.
    * We don't have to decide which minimum version to support.
    * In theory, due to dynamic linking against frameworks, as long as you do not use newer macOS APIs then binaries produced with the latest SDK will run on older macOS/iOS/etc versions I believe.
3. **Ship every supported version, allow users to choose**
    * The worst option IMO. Increased complexity on the Zig side.
    * Would still need to decide which version to default to.
    * Users would need to pick versions, some libraries would support only older versions while others would support only latest.
4. **Middle-ground solution, mix and match SDK versions?**
    * By default, Zig ships lowest-supported macOS version like we do today and targets that. This ensures all Zig libs in the ecosystem target that version by default.
    * Zig libs that need a newer macOS SDK version can supply the macOS SDK for cross-compilation. That SDK would be used for compiling _that Zig library only_, including using it for libc headers. Unlike setting `b.sysroot` which would try to use that SDK for compilation of all Zig libraries, this would occur for that single Zig library only.

kubkon:
Thanks for the PR and issue @slimsag. I am fine with trying to keep macOS libc headers at the latest version given that Apple encourages their users to update their OSes to the latest possible version every year. That said, this means the libc headers we ship will be incompatible with anything but the latest sysroot (well, they might work with the previous sets but we can never guarantee that I guess?). I think I am still fine with that personally. Also, if anyone would like to target an older version of macOS from a foreign host they'd just need to provide the entire sysroot including the libc headers (which are part of every macOS SDK anyhow). Finally, as you already pointed out, targeting the latest macOS version is the simplest thing to maintain on our side too.

I'd like to hear @andrewrk's opinion on this before committing to any one particular strategy.

kubkon:
Oh and also, sticking to the latest libc headers would go hand-in-hand in us tracking the latest LLVM release I guess.

slimsag:
Conversation shifted to over here: https://github.com/ziglang/fetch-them-macos-headers/pull/15#issuecomment-976091688

kubkon:
Superseded by #10215, closing.

PR #10205, opened 2021-11-23
std/math: add ldexp and make scalbn an alias

### LINKED ISSUE #9799: Missing ldexp() in std.math
As far as I can tell std.math does not provide ldexp(). I suppose there is no particular reason for this, and it just wasn't done yet? Or did I just miss it somehow?

EDIT: In practice I can just use scalbn() I suppose?

tiehuis:
We assume we are compiled on a base-2 radix floating point system. This
is a reasonable assumption. musl libc as an example also assumes this.

We implement scalbn as an alias for ldexp, since ldexp is defined as 2
regardless of the float radix. This is opposite to musl which defines
scalbn in terms of ldexp.

Closes #9799.

### DIFF
--- a/lib/std/math/ldexp.zig
+++ b/lib/std/math/ldexp.zig
@@ -1,0 +1,92 @@
+// Ported from musl, which is licensed under the MIT license:
+// https://git.musl-libc.org/cgit/musl/tree/COPYRIGHT
+//
+// https://git.musl-libc.org/cgit/musl/tree/src/math/ldexpf.c
+// https://git.musl-libc.org/cgit/musl/tree/src/math/ldexp.c
+
+const std = @import("std");
+const math = std.math;
+const assert = std.debug.assert;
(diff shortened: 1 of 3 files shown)


lib/std/math/ldexp.zig
```diff
@@ -0,0 +1,92 @@
...
+
+/// Returns x * 2^n.
+pub fn ldexp(x: anytype, n: i32) @TypeOf(x) {
+    var base = x;
+    var shift = n;
+
+    const T = @TypeOf(base);
+    const IntT = std.meta.Int(.unsigned, @bitSizeOf(T));
+    if (@typeInfo(T) != .Float) {
+        @compileError("ldexp not implemented for " ++ @typeName(T));
```
tiehuis, line 20: This may be confusing to callers of scalbn if they pass a different float type and get this message.

andrewrk (approved):
Thanks!

PR #10209, opened 2021-11-23
use safe invalidation only in safe modes

marler8997:
I found 2 cases in std where there are "safer" values than "undefined" when it comes invalidation.  I've modified those 2 cases to only do this when runtime_safety is enabled, so we can have the extra safety without paying the cost in release mode.

Note that setting values to `undefined` can aid in "optimize-ability".

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -182,11 +182,11 @@ pub const AtomicFile = struct {
     pub fn deinit(self: *AtomicFile) void {
         if (self.file_open) {
             self.file.close();
-            self.file_open = false;
+            self.file_open = if (std.debug.runtime_safety) false else undefined;
         }
         if (self.file_exists) {
             self.dir.deleteFile(&self.tmp_path_buf) catch {};
-            self.file_exists = false;
+            self.file_exists = if (std.debug.runtime_safety) false else undefined;
         }
         if (self.close_dir_on_deinit) {
             self.dir.close();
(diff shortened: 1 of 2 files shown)


ifreund (commented):
On further thought, I personally don't think this should be merged as I don't think the supposed performance gain from this change will be measurable if even existent. If I'm right and there is no performance gain, this patch does only two things, neither of which are desirable:

1. Make the standard library code more complex.
2. Make release fast/small build modes less safe.

marler8997:
Would you be for it if it did have performance gains?

daurnimator:
Only `undefined` has the special detecting properties of valgrind. This is worse depending on the tools you use.

marler8997:
@daurnimator sorry I'm not sure I understand your comment, could you elaborate for me?  Are you saying that valgrind knows about `undefined` and not using `undefined` in all cases is "worse" in some way?  Are you saying we should invalidate with `undefined` in all cases?

daurnimator:
> could you elaborate for me?

Whenever we assign `undefined`, we also emit valgrind instrumentation (special sequence of NOP instructions) that valgrind picks up so that it can warn you about e.g. branching on undefined memory.

>  Are you saying that valgrind knows about undefined and not using undefined in all cases is "worse" in some way?

yes.

andrewrk:
BTW regarding performance optimizations, there is a new policy: the benchmark has to be covered by https://github.com/ziglang/gotta-go-fast/ so that we can watch the graph change after the perf improvement is merged to master branch.

marler8997:
I don't anticipate this change would have a measurable affect on performance.  It was more to establish a "convention" since people look at the `std` library for conventions.  If we require performance changes to be measurable then this is DOA.  A variation on this where we always set to `undefined` could be considered more than a performance change, I'll leave that to someone else if they want to make that change.

Issue #89, opened 2016-01-26
self-hosted compiler: ship it!


andrewrk:
This is a tracking issue for when we can completely replace stage1 with self-hosted in official builds of Zig.

 * [x] add debug info to the LLVM backend
   - [x] initial implementation #11085
   - [x]  #11095
   - [x]  #11096
   - [x]  #11097
   - [x]  #11145
   - [x]  #11158
   - [x]  #11160
 * [x] #11674
 * [x] get compilation errors into an acceptable state
   - [x] #11288
   - [x] #11632
 * [x] adjust the default backend to be LLVM until other backends reach full behavior test coverage
 * [ ] ~#6025~ async/await/suspend/resume will not be available in 0.10.0
 * [x] #6034
 * [x] #11159
 * [x] #11169
 * [x] #11251
 * [x] #11259
 * [x] all std lib tests passing with the LLVM backend
   - [x] #11898
   - [x] #11897
   - [x] #11857
   - [x] #11856
   - [x] #11853
   - [x] #12029
 * [x] #11450
 * [x] #11498
 * [x] #11767
 * [x] #11274
 * [x] full test suite passing with the LLVM backend
 * [ ] ~check all the TODO comments in stage2 source code and convert relevant ones into issues blocking this one~ #363
 * [x] Eliminate use of conditional compilation on zig_backend in compiler-rt, freestanding libc, test runner, and default panic
 * [x] use the real start code
 * [x] eliminate conditional compilation on zig_backend in the std lib
 * [x] Eliminate the other TODO items from generated builtin.zig file
 * [ ] work with individual third party projects to make sure the stage2 transition completes successfully
   - [ ] Bun (cc @Jarred-Sumner)
   - [ ] zig-gamedev (cc @michal-z)
   - [ ] mach (cc @slimsag)
   - [ ] River (cc @ifreund)
   - [ ] aquila (cc @nektro)
   - [ ] https://github.com/ominitay/zigvale-barebones (cc @ominitay)
   - [ ] [ncdu](https://dev.yorhel.nl/ncdu)
   - [ ] https://github.com/ZigEmbeddedGroup/microzig (`zig build -Drelease-small`)
   - [ ] https://github.com/mattnite/gyro (`zig build`, good amount of zig and C code in here)
   - [ ] [ZLS](https://github.com/zigtools/zls)
   - [ ] https://github.com/sin-ack/zigself
   - [ ] https://github.com/floooh/sokol-zig

andrewrk:
[...]

PR #10214, opened 2021-11-24
File.close invalidates

marler8997:
This change is the polar opposite of https://github.com/ziglang/zig/pull/10213 but one that Andrew has expressed he would like.  Instead of providing 2 ways to cleanup a file (one that invalidates and one that doesn't), this PR modifies `File.close` to always invalidate.  Note this is a breaking change that now requires all instances of `File` that you want to close to be mutable.  This means no more `const file = `, you must always use `var file =`.  I've argued this reduces code readability, but that's my opinion, you can see for yourselves whether or not you agree and whether it would justify a second way to cleanup files (i.e. adding a `closeConst` for const files).

Also note that if we start requiring files to be mutable, we will be introducing "ownership" semantics to `File`.  Namely, passing a `File` by value vs by reference has a new meaning, and having a `File` field as a value or a reference has new meaning.  This adds complexity to how `File` is interpreted that doesn't exist today with immutable instances of `File`.

### DIFF
--- a/lib/std/Thread.zig
+++ b/lib/std/Thread.zig
@@ -80,7 +80,7 @@ pub fn setName(self: Thread, name: []const u8) SetNameError!void {
             var buf: [32]u8 = undefined;
             const path = try std.fmt.bufPrint(&buf, "/proc/self/task/{d}/comm", .{self.getHandle()});
 
-            const file = try std.fs.cwd().openFile(path, .{ .write = true });
+            var file = try std.fs.cwd().openFile(path, .{ .write = true });
             defer file.close();
 
             try file.writer().writeAll(name);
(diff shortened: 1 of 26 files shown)


andrewrk:
I recommend to run the full test suite locally for a big breaking change such as this to make sure it will pass the CI.

andrewrk:
Looks like this was never passing the CI, and now it has a bunch of merge conflicts. I'm closing this since it's an old PR. I think for this particular change it's a bit of a tricky one - we're going to want to pick a strategic time to make it.

Personally this is not a priority to me until we start focusing on the standard library, which I expect to be in a couple release cycles from now, after we ship self-hosted (#89), and after we have the package manager (#943).

andrewrk:
I see the similarities between these cases. I don't want to make this change to the allocator interface. I see your point. I don't have a conclusion or any guidance at this time. I'm focused on other parts of the project and am postponing these API consistency considerations. I think we have other priorities currently and the resolution to this problem will reveal itself once we focus on it.

PR #10220, opened 2021-11-25
std.mem.indexOfPos should return start_index when needle length is zero

### LINKED ISSUE #10216: std.mem.indexOfPos should return start_index when needle length is zero
### Zig Version

0.9.0-dev.1712+9836f1b2f

### Steps to Reproduce

Search for an empty slice inside another:
```zig
const std = @import("std");
const indexOfPos = std.mem.indexOfPos;

test "should return start index" {
    const index = indexOfPos(u8, "abracadabra", 5, "") orelse unreachable;
    try expect(index == 5); // fails
}
```

### Expected Behavior

I expect an empty slice to be always found exactly at the specified start index.

### Actual Behavior

The `indexOfPos` function always returns `0` if the needle has no length.

fabioarnold:
Closes #10216


andrewrk (approved):
Thanks!

PR #10224, opened 2021-11-26
zig fmt: Fix performance issue with nested arrays

### LINKED ISSUE #10194: `zig fmt` takes exponential time to render nested arrays/tuples with trailing commas
### Zig Version

0.9.0-dev.1678+7747bf07c

### Steps to Reproduce

Run `zig fmt` on this sequence of inputs.

- `const x = .{.{},};`
- `const x = .{.{.{},},};`
- `const x = .{.{.{.{},},},};`

or

- `const x = T{T{},};`
- `const x = T{T{T{},},};`
- `const x = T{T{T{T{},},},};`
- `And so on...`

### Expected Behavior

Formatter goes fast

### Actual Behavior

Each iteration takes twice as long as the previous. It takes about 20 iterations to get to 1 second execution time. Deleting the commas fixes this problem.

chwayne:
Fixes #10194 

Makes changes to `renderArrayInit` within `lib/std/zig/render.zig`. This new code works, except for one minor case I've found (shown below) where it doesn't format code correctly. It just needs a few more tweaks to make it correct.

Specifically, input like this:
```zig
const x = .{
    a,

    // Comment 1
    b, // Comment 2
};
```

should not change, but intstead gets formatted to this:
```zig
const x = .{
    a,
    // Comment 1
    b, // Comment 2
};
```

### DIFF
--- a/lib/std/zig/render.zig
+++ b/lib/std/zig/render.zig
@@ -1747,6 +1747,9 @@ fn renderArrayInit(
         var sub_expr_buffer = std.ArrayList(u8).init(gpa);
         defer sub_expr_buffer.deinit();
 
+        const sub_expr_buffer_starts = try gpa.alloc(usize, section_exprs.len + 1);
+        defer gpa.free(sub_expr_buffer_starts);
+
         var auto_indenting_stream = Ais{
             .indent_delta = indent_delta,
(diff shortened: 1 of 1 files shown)


andrewrk:
Thanks!

PR #10226, opened 2021-11-26
macos: improved SDK detection and linker integration with _mh_execute_header

### LINKED ISSUE #1982: creating dynamic libraries targeting macosx results in unusable .dylib files
* Host: `aarch64-macos.12.0.1`
 * Zig Version: `0.9.0-dev.1759+c9352ef9d`

```zig
const std = @import("std");
export fn base_address() usize {
    return std.process.getBaseAddress();
}
```

```
$ zig build-lib lib.zig -dynamic
error(link): undefined reference to symbol '__mh_execute_header'
error(link):   first referenced in '/Users/andy/tmp/zig-cache/o/e365c75106ca0ae0e6932a78cf567847/lib.o'
error: UndefinedSymbolReference
```

That's the first problem. But let's work around it by adjusting the zig code:

```zig
const std = @import("std");
export fn base_address() usize {
    return 1234;
}
```

```c
#include <stdio.h>

size_t base_address(void);

int main(int argc, char **argv) {
    fprintf(stderr, "base=%zu\n", base_address());
}
```

```
$ cc -o test test.c liblib.dylib
ld: warning: dylib (liblib.dylib) was built for newer macOS version (12.0.1) than being linked (12.0)
$ ./test
dyld[51244]: Library not loaded: @rpath/liblib.dylib
  Referenced from: /Users/andy/tmp/test
  Reason: tried: '/usr/local/lib/liblib.dylib' (no such file), '/usr/lib/liblib.dylib' (no such file)
zsh: abort      ./test
```

Related issue regarding static libraries: #1981

### LINKED ISSUE #10217: (regression) building macOS binary without XCode / Command Line Tools installed fails
### Zig Version

0.9.0-dev.1696+ff9a44a9f

### Steps to Reproduce

1. On a Mac machine without XCode / XCode Command Line Tools installed
2. `zig build` any Zig project without cross-compilation
3. `zig build -D=target=aarch64-macos` with cross-compilation (to the same target as the host)

### Expected Behavior

Both commands should successfully build a macOS binary without XCode / Command Line Tools being installed.

Given that Zig can cross-compile for macOS without those being installed on other targets, I think it is a super nice property to maintain that if you don't have those installed on a mac host that you don't have to install them - specifically because they are multi-GB downloads that can often take a half hour to install.

In prior versions of Zig (several months after 0.8 was released) both commands would succeed without XCode / Command Line Tools being installed. After the introduction of calling `xcrun` in Zig, however, these began to fail. I've been meaning to file this for a while now. @kubkon 

### Actual Behavior

Both commands fail with the following:

```
Zig attempted to find the path to native system libc headers by executing this command:
cc -E -Wp,-v -xc /dev/null
Output:
==========
xcode-select: note: no developer tools were found at '/Applications/Xcode.app', requesting install. Choose an option in the dialog to download the command line developer tools.
```

After running, the `xcrun` dialog box prompting to install XCode pops up. If you dismiss it and re-run the `zig` command multiple times, it eventually succeeds in producing the binary without XCode.

kubkon:
~~If Zig didn't detect native SDK, always use shipped libc headers when targeting macOS.~~

~~Fixes #10217~~

~~cc @slimsag~~

EDIT: Turns out this PR leads naturally to a fix of another long-standing issue, so I've now converged the two, one on top of the other.

Changes / fixes:
* always use Zig shipped libc headers when no native SDK - if Zig didn't detect native SDK, always use shipped libc headers when targeting macOS - fixes #10217 
* extract native SDK path *and* version info, and route that info all the way to the linker so that it can correctly populate `LC_BUILD_VERSION` load command
* fix version info encoded as part of `LC_BUILD_VERSION` load command - Apple strips off patch info from platform and SDK version info - without this, non-Zig tools may complain that the there's a mismatch between OS and platform as reported in #1982 
* define `__mh_execute_header` as a linker synthetic global symbol - fixes #1982 
* EDIT2: add a helper `std.zig.system.darwin.isDarwinSDKInstalled` which checks for the presence of the SDK tooling without triggering the CLT installation popup - I believe this is a much better UX on a bare macOS system

cc @andrewrk @slimsag

### DIFF
--- a/lib/std/zig/system/darwin.zig
+++ b/lib/std/zig/system/darwin.zig
@@ -2,14 +2,15 @@ const std = @import("std");
 const mem = std.mem;
 const Allocator = mem.Allocator;
 const Target = std.Target;
+const Version = std.builtin.Version;
 
 pub const macos = @import("darwin/macos.zig");
 
-/// Detect SDK path on Darwin.
-/// Calls `xcrun --sdk <target_sdk> --show-sdk-path` which result can be used to specify
-/// `--sysroot` of the compiler.
-/// The caller needs to free the resulting path slice.
-pub fn getSDKPath(allocator: *Allocator, target: Target) !?[]u8 {
+/// Detect SDK on Darwin.
+/// Calls `xcrun --sdk <target_sdk> --show-sdk-path` which fetches the path to the SDK sysroot (if any).
+/// Subsequently calls `xcrun --sdk <target_sdk> --show-sdk-version` which fetches version of the SDK.
+/// The caller needs to deinit the resulting struct.
+pub fn getDarwinSDK(allocator: *Allocator, target: Target) !?DarwinSDK {
     const is_simulator_abi = target.abi == .simulator;
     const sdk = switch (target.os.tag) {
         .macos => "macosx",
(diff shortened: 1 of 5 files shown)


andrewrk (approved):
Excellent work! Super sweet fix with _mh_execute_header!

slimsag:
Works beautifully, thanks @kubkon !

PR #10229, opened 2021-11-26
stage2: Upstream zwld (partly) into wasm backend

Luukdegram:
This upstreams a lot of logic from https://github.com/Luukdegram/zwld.
For now, I have chosen a hybrid method of linking, where easy-to-track and small computations are done using incremental style but hard-to-track linking is delayed until `flushModule()`. This has the benefits of having more information available during the linking process, and not having to worry too much about updating all indexes constantly. Besides, this will allow us to merge the remaining parts of zwld (once it gains the missing features), with almost no effort, meaning we will then be able to link both stage1 as well as stage2 code with external object files.

Prior to this PR, we were using a userland implementation of an offset table where we placed it inside wasm's memory. After this PR, we will generate and resolve relocations properly. We now have all the information available to also produce object files from the wasm backend, rather than only executables (once the emitting parts are implemented).

cc @kubkon

### DIFF
--- a/lib/std/wasm.zig
+++ b/lib/std/wasm.zig
@@ -1,4 +1,8 @@
-const testing = @import("std.zig").testing;
+///! Contains all constants and types representing the wasm
+///! binary format, as specified by:
+///! https://webassembly.github.io/spec/core/
+const std = @import("std.zig");
+const testing = std.testing;
 
 // TODO: Add support for multi-byte ops (e.g. table operations)
 
@@ -222,6 +226,18 @@ pub fn valtype(value: Valtype) u8 {
     return @enumToInt(value);
 }
 
+/// Reference types, where the funcref references to a function regardless of its type
+/// and ref references an object from the embedder.
+pub const RefType = enum(u8) {
+    funcref = 0x70,
+    externref = 0x6F,
+};
+
+/// Returns the integer value of a `Reftype`
+pub fn reftype(value: RefType) u8 {
+    return @enumToInt(value);
+}
+
 test "Wasm - valtypes" {
     const _i32 = valtype(.i32);
     const _i64 = valtype(.i64);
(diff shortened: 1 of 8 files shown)


andrewrk:
Nice work!

CI failure is due to compile errors from trying to build the compiler on 32-bit systems. You should be able to reproduce (even in zwld) by throwing a `-Dtarget=arm-linux-musleabihf` on there.

```
+ /workspace/_debug/staging/bin/zig build -Dtarget=arm-linux-musleabihf
./src/link/Wasm.zig:501:32: error: expected type 'usize', found 'u64'
        if (!std.mem.isAligned(initial_memory, page_size)) {
                               ^
./src/link/Wasm.zig:501:32: note: unsigned 32-bit int cannot represent all possible unsigned 64-bit values
        if (!std.mem.isAligned(initial_memory, page_size)) {
                               ^
./src/link/Wasm.zig:584:5: note: referenced here
    try self.setupMemory();
    ^
./src/Compilation.zig:1877:5: note: referenced here
    try self.bin_file.flush(self);
    ^
./src/main.zig:2689:5: note: referenced here
    try comp.update();
    ^
./src/main.zig:2370:35: note: referenced here
                                  ^
```

Luukdegram:
> Nice work!
> 
> The CI failure is due to compile errors from trying to build the compiler on 32-bit systems. You should be able to reproduce (even in zwld) by throwing a `-Dtarget=arm-linux-musleabihf` on there.
> 
> ```
> + /workspace/_debug/staging/bin/zig build -Dtarget=arm-linux-musleabihf
> ./src/link/Wasm.zig:501:32: error: expected type 'usize', found 'u64'
>         if (!std.mem.isAligned(initial_memory, page_size)) {
>                                ^
> ./src/link/Wasm.zig:501:32: note: unsigned 32-bit int cannot represent all possible unsigned 64-bit values
>         if (!std.mem.isAligned(initial_memory, page_size)) {
>                                ^
> ```

Thanks for the pointer on the target. Fixed it in 6e88df4

PR #10232, opened 2021-11-27
compiler_rt: implement __isPlatformVersionAtLeast (Objective-C @available expressions) for Darwin

### LINKED ISSUE #10227: compiler-rt port does not support Objective-C @available expressions
### Zig Version

0.9.0-dev.1713+07cc51e04

### Steps to Reproduce

(Apologies for not having a minimal reproduction ready, I will update the issue to include one very soon.)

When using Zig to compile any Objective-C code which includes an `@available` check, as many serious Objective-C codebases do, [such as Google's Dawn WebGPU implementation](https://github.com/hexops/dawn/blob/da58a7b2db6fd2dc2fbaf1857fa2dbb4c7946a9d/src/dawn_native/metal/BackendMTL.mm#L220), clang will emit calls to `___isPlatformVersionAtLeast` and `__isOSVersionAtLeast` which, as of Oct 2020, are expected to be provided by compiler-rt: https://reviews.llvm.org/D90367

Zig's compiler-rt port however is not up-to-date with this change (see also #1504) and so it results in undefined references when attempting to cross-compile such code:

```
error(link): undefined reference to symbol '___isPlatformVersionAtLeast'
error(link):   first referenced in '/Users/slimsag/Desktop/hexops/mach/gpu/zig-cache/o/a4898ce0dd88be1c63f03f62251f9adb/libdawn-native.a(/Users/slimsag/Desktop/hexops/mach/gpu/zig-cache/o/22c50fc6233f5284bfe9a440d8943d61/BackendMTL.o)'
thread 7413503 panic: attempt to unwrap error: UndefinedSymbolReference
```

### Expected Behavior

compiler-rt includes `___isPlatformVersionAtLeast` and `__isOSVersionAtLeast` definitions so Objective-C code using `@available` OS version checks links successfully.

### Actual Behavior

Missing symbols

slimsag:
Objective-C code in the wild will frequently use `@available` expressions to check if a given macOS, iOS, etc. (even Android OS) version is available before using a newer API. You can see an example of what these look like in [Google's Dawn WebGPU codebase here](https://github.com/hexops/dawn/blob/da58a7b2db6fd2dc2fbaf1857fa2dbb4c7946a9d/src/dawn_native/metal/BackendMTL.mm#L220).

When `@available` is used, clang emits a call to `___isPlatformVersionAtLeast` which is implemented in compiler-rt. The LLVM 13 implementation is available here: https://github.com/llvm/llvm-project/blob/llvmorg-13.0.0/compiler-rt/lib/builtins/os_version_check.c

In this PR, I port the LLVM implementation with two caveats:

1. Android targets, when Objective C code is compiled for Android `@available` can be used to determine the Android OS version. I did not port this code, but made it easy and clear how the next person would do so.
2. I did not retain the backwards compatibility logic LLVM currently retains for older macOS and clang versions:
    * macOS (and equivalent iOS/etc versions) 10.15+ expose a native `_availability_version_check` function we can simply invoke that does all of the hard work for us. The LLVM implementation also uses this, when available.
    * The backwards compatibility logic is fairly complex: dynamic library loading, system plist parsing logic, etc. and given Zig supports only the last three macOS versions porting this did not seem useful so I opted to keep the implementation super simple instead.

Fixes #10227


lib/std/special/compiler_rt/os_version_check.zig
```diff
@@ -0,0 +1,76 @@
...
+// the newer codepath, which merely calls out to the Darwin _availability_version_check API which is
+// available on macOS 10.15+, iOS 13+, tvOS 13+ and watchOS 6+.
+
+inline fn constructVersion(major: u32, minor: u32, subminor: u32) u32 {
+    return ((major & 0xffff) << 16) | ((minor & 0xff) << 8) | (subminor & 0xff);
+}
+
+// Darwin-only
+pub fn __isPlatformVersionAtLeast(platform: u32, major: u32, minor: u32, subminor: u32) callconv(.C) i32 {
+    return if (_availability_version_check(1, &[_]dyld_build_version_t{
```
LemonBoy, line 22: Nit, use `@boolToInt` and break up the long tuple by adding a trailing comma.

lib/std/special/compiler_rt/os_version_check.zig
```diff
@@ -0,0 +1,76 @@
...
+        .{ .platform = platform, .version = constructVersion(major, minor, subminor) },
+    })) 1 else 0;
+}
+
+// TODO(future): If anyone wishes to build Objective-C code for Android, we would need to port this
```
LemonBoy, line 27: IMO this can be removed as it may get out of sync with the upstream impl by the time somebody tries to implement it.

lib/std/special/compiler_rt/os_version_check_test.zig
```diff
@@ -0,0 +1,10 @@
+const __isPlatformVersionAtLeast = @import("os_version_check.zig").__isPlatformVersionAtLeast;
+const testing = @import("std").testing;
+
+test "isPlatformVersionAtLeast" {
```
LemonBoy, line 4: Move the test inside `os_version_check.zig` and have it execute only if the target OS is macos.

slimsag: Ah good call on only executing if target is macOS

normally I would've done test in same file, but I noticed compiler_rt had an unusual amount of `_test.zig` files and thought maybe some trickery was going on there.

slimsag:
@LemonBoy done :)

lib/std/special/compiler_rt/os_version_check_test.zig
andrewrk: The reason for _test.zig files is that they are excluded from installation. Compiler-rt has a lot of MiB of tests.

PR #10233, opened 2021-11-27
Add missing macOS libc headers

### LINKED ISSUE #9395: Some macOS `#include`s fail when cross-compiling using Zig
CoreCLR contains some macOS `#include`s here: https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/coreclr/gc/unix/gcenv.unix.cpp#L65-L76

Of these, two fail when building with Zig:

```
fatal error : 'mach/thread_state.h' file not found
fatal error : 'mach/vm_param.h' file not found
```

Replacing them with `mach/machine/...` equivalents seems to work, but this is probably still a bug in Zig.

kubkon:
* mach/thread_state.h
* mach/vm_param.h
* objc/objc-runtime.h

Fixes #9395


kubkon:
@andrewrk if you notice, we keep two copies of `vm_param.h`: one in `x86_64.macos.10-gnu` and one in `any-macos-any`. The only difference between them is change in the year range from 2008 to 2020. I'm wondering if we should do a manual intervention for this one in `fetch-them-macos-headers` to make them the same across all versions. What do you think?

kubkon:
I also should point out that it's not a problem we potentially have two headers with the same name included in the `zig cc` invocation as we always include the cpu and os version-specific include dir first so `x86_64-macos.10 < any-macos`.

andrewrk:
Nah I don't think we need any intervention. Thanks!

Issue #10145, opened 2021-11-14
RmdirZ does not handle all returned syscall errors


Daimanta:
### Remember to search before filing a new report

- [X] I searched for this bug and did not find it in the issue tracker, and furthermore, the title I used above will make this new bug report turn up in the search results for my query.

### Zig Version

0.8.1

### Steps to Reproduce

```
try std.os.rmdirZ(".");
```

### Expected Behavior

An error is returned, something like 'IncorrectDir'

### Actual Behavior

```EINVAL => unreachable,```
is reached which prevents handling of the error.

As a side note, 'EFAULT' is also not handled while the rmdir syscall can return this
_"EFAULT pathname points outside your accessible address space."_

**Comment by leecannon** [comment]
Looks like this issue should have been closed with #10236.

`error.BadPathName` is now returned for the provided code:
```zig
$ zig build run                                                                                                                                                                
error: BadPathName
/home/lee/src/github.com/leecannon/zig/lib/std/os.zig:2565:19: 0x23469f in std.os.rmdirZ (fff)
        .INVAL => return error.BadPathName,
                  ^
/home/lee/src/fff/src/main.zig:6:5: 0x22d225 in main (fff)
    try std.os.rmdirZ(".");
    ^
```

sskras:
Fixed by the [89afd4b](https://github.com/ziglang/zig/commit/89afd4bd33c72c2e608974d182f4624078da3a7f) commit [on Dec 1, 2021](https://github.com/ziglang/zig/pull/10236#event-5695776026) ?

PR #10236, opened 2021-11-28
libstd: handle all rmdirZ syscall errors (#10145)

jfilling:
Both FAULT and INVAL were marked unreachable which prevents handling
of the error at a higher level.

It seems like they should map to BadPathError based on the man page for
rmdir (and an incomplete understanding of DeleteDirError), which says:

```
EFAULT pathname points outside your accessible address space.
EINVAL pathname has .  as last component.
```


daurnimator:
> `EFAULT pathname points outside your accessible address space.`

This is intentionally `unreachable`, it means something has gone terribly wrong in your program and you have an invalid pointer.

jfilling:
I just noticed another bug discussing EINVAL from other syscalls, and realized that EFAULT should remain unreachable for rmdir. Will pull this request and fix if worthwhile.

PR #10239, opened 2021-11-28
macho: fix regression in handling SIGNED_X relocs on x86_64

### LINKED ISSUE #10207: macOS Intel: crash in release modes on the C side (LLVM codegen issue?)
### Zig Version

0.9.0-dev.1722+ed70f9981

### Steps to Reproduce

cc @kubkon: here's the promised ticket from https://github.com/ziglang/zig/issues/9747#issuecomment-974805026)

Ok, this is a weird problem, brace yourself:

On an Intel Mac (my config: Big Sur 11.6.2, Xcode 13.1, macOS SDK 12.0, Zig installed via nightly download, but ```brew install --HEAD``` has the same behaviour):

```sh
git clone https://github.com/floooh/sokol-zig
cd sokol-zig
zig build -Drelease-safe run-clear
```
The executable crashes on the C side with:

```
Assertion failed: ((desc->buffer_pool_size > 0) && (desc->buffer_pool_size < _SG_MAX_POOL_SIZE)), function _sg_setup_pools, file sokol_gfx.h, line 13332.
```

Or alternatively:
```
zig build -Drelease-fast run-clear
```
In this case the crash happens when trying to allocate a large piece of memory on the C side (which would have been caught by the assert in release-safe mode):

```
clear(5067,0x116afae00) malloc: can't allocate region
:*** mach_vm_map(size=18446744065119621120, flags: 100) failed (error code=3)
clear(5067,0x116afae00) malloc: *** set a breakpoint in malloc_error_break to debug
```

When buiding without release-safe or release-fast, it works fine.

The problem seems Intel Mac specific, it doesn't happen on M1 Macs or on Windows (haven't tested on Linux).

The problem occurs inside this call:

https://github.com/floooh/sokol-zig/blob/603a4abf693002d0cc2fd1b391ff111f78b698e0/src/examples/clear.zig#L13-L15

The provided structure is default-initialized to zero:

https://github.com/floooh/sokol-zig/blob/603a4abf693002d0cc2fd1b391ff111f78b698e0/src/sokol/gfx.zig#L641-L654

The function call into C goes through this stub where the address of the struct that's passed by value in the Zig call is taken when calling into the C side:

https://github.com/floooh/sokol-zig/blob/603a4abf693002d0cc2fd1b391ff111f78b698e0/src/sokol/gfx.zig#L655-L658

Eventually the call ends up here on the C side:

https://github.com/floooh/sokol-zig/blob/603a4abf693002d0cc2fd1b391ff111f78b698e0/src/sokol/c/sokol_gfx.h#L14773

Adding debug-printfs on the Zig- and C-side reveals that the structure content *doesn't* get corrupted on its way from Zig to C (which was my first suspicion), instead the breakage happens in this C code block which "patches" default values into zero-initialized struct members:

https://github.com/floooh/sokol-zig/blob/603a4abf693002d0cc2fd1b391ff111f78b698e0/src/sokol/c/sokol_gfx.h#L14790-L14800

Those ```_sg_def``` macros resolve to:
```
_sg.desc.buffer_pool_size = (((_sg.desc.buffer_pool_size) == 0) ? (default_value) : (_sg.desc.buffer_pool_size));
```

Interestingly, adding debug-printf-calls which print _sg.desc.buffer_pool_size before that construct makes the problem go away (great example of a Heisenbug I guess).

I can workaround the problem by **not** using the same left-hand-side and right-hand-side-values:
```
_sg.desc.buffer_pool_size = (((desc->buffer_pool_size) == 0) ? (default_value) : (desc->buffer_pool_size));
```

No idea how to proceed from here. It looks to me like perfectly valid C code. Is this a side effect of the last LLVM update perhaps or different compilation flags for C code? But I don't see the same problem with Clang in C projects, I'm also using -fstrict-aliasing for my C/C++ code anywhere.

### Expected Behavior

...

### Actual Behavior

...

kubkon:
Fixes #10207

### DIFF
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -4996,17 +4996,9 @@ fn writeSymbolTable(self: *MachO) !void {
         }
     }
 
-    var undefs = std.ArrayList(macho.nlist_64).init(self.base.allocator);
-    defer undefs.deinit();
-
-    for (self.undefs.items) |sym| {
-        if (sym.n_strx == 0) continue;
-        try undefs.append(sym);
-    }
-
     const nlocals = locals.items.len;
     const nexports = self.globals.items.len;
-    const nundefs = undefs.items.len;
+    const nundefs = self.undefs.items.len;
 
     const locals_off = symtab.symoff;
     const locals_size = nlocals * @sizeOf(macho.nlist_64);
@@ -5021,7 +5013,7 @@ fn writeSymbolTable(self: *MachO) !void {
     const undefs_off = exports_off + exports_size;
     const undefs_size = nundefs * @sizeOf(macho.nlist_64);
     log.debug("writing undefined symbols from 0x{x} to 0x{x}", .{ undefs_off, undefs_size + undefs_off });
-    try self.base.file.?.pwriteAll(mem.sliceAsBytes(undefs.items), undefs_off);
+    try self.base.file.?.pwriteAll(mem.sliceAsBytes(self.undefs.items), undefs_off);
 
     symtab.nsyms = @intCast(u32, nlocals + nexports + nundefs);
     seg.inner.filesize += locals_size + exports_size + undefs_size;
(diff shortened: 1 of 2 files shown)


kubkon:
cc @floooh

PR #10240, opened 2021-11-28
Stage2: wasm - Implement 'zig test'

Luukdegram:
This implements all the required instructions and declaration codegen to be able to use `zig test` using `wasm32-wasi` as a target.

A simple example:
```zig
// test.zig
test {
    var i: u32 = 5;
    if (i != 5) unreachable;
}
```
```sh
zig test test.zig -target wasm32-wasi --test-cmd wasmtime --test-cmd-bin
0
```

The temporary _start supports both `u8` and `void` as return type, to also support our old wasm stage2 testing backend. In a followup PR I'll be updating the `behaviour.zig` file to track progress on the wasm backend.
This is a huge milestone for the wasm backend and the backend is now in a state where contributing
is a lot easier.

### DIFF
--- a/lib/std/start.zig
+++ b/lib/std/start.zig
@@ -30,6 +30,8 @@ comptime {
                 }
             } else if (builtin.os.tag == .windows) {
                 @export(wWinMainCRTStartup2, .{ .name = "wWinMainCRTStartup" });
+            } else if (builtin.os.tag == .wasi) {
+                @export(wasmMain2, .{ .name = "_start" });
             } else {
                 if (!@hasDecl(root, "_start")) {
                     @export(_start2, .{ .name = "_start" });
@@ -96,6 +98,11 @@ fn callMain2() noreturn {
     @setAlignStack(16);
     root.main();
     exit2(0);
+}
+
+fn wasmMain2() u8 {
+    root.main();
+    return 0;
 }
 
 fn wWinMainCRTStartup2() callconv(.C) noreturn {
--- a/lib/std/start.zig
+++ b/lib/std/start.zig
@@ -101,8 +101,19 @@ fn callMain2() noreturn {
 }
 
 fn wasmMain2() u8 {
-    root.main();
-    return 0;
+    switch (@typeInfo(@typeInfo(@TypeOf(root.main)).Fn.return_type.?)) {
+        .Void => {
+            root.main();
+            return 0;
+        },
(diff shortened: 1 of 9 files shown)


Luukdegram:
Errors seem related to my PR. Though, not sure yet what's causing the change of behavior in the Darwin tests.

kubkon:
> Errors seem related to my PR. Though, not sure yet what's causing the change of behavior in the Darwin tests.

How do you know it's darwin specifically?

Luukdegram:
> > Errors seem related to my PR. Though, not sure yet what's causing the change of behavior in the Darwin tests.
> 
> How do you know it's darwin specifically? It may well be wasm tests cross-compiled on darwin too.

`test 'darwin hello world with updates' failed: WrongCompileErrors`
But upon further inspection, it seems to happen for `x86_64` as well.

It now seems to output the file path as well.

lib/std/start.zig
```diff
@@ -98,6 +100,22 @@ fn callMain2() noreturn {
...
 
+fn wasmMain2() u8 {
+    switch (@typeInfo(@typeInfo(@TypeOf(root.main)).Fn.return_type.?)) {
+        .Void => {
+            root.main();
+            return 0;
+        },
+        .Int => |info| {
+            if (info.bits != 8 or info.signedness == .signed) {
+                @compileError(bad_main_ret);
```
kubkon, line 111: What is `bad_main_ret`? Could this be the source of failures on other targets?

Luukdegram: https://github.com/ziglang/zig/blob/35f0ab613579efae5801e81d426963b9603db050/lib/std/start.zig#L469

kubkon: Gotcha, thanks!

Luukdegram:
We found the culprit: Had to update the line numbers in the compilation error checks (since I updated the `start.zig` file).

Luukdegram:
Will fix the 32bit test error tomorrow :)

andrewrk:
> We found the culprit: Had to update the line numbers in the compilation error checks (since I updated the `start.zig` file).

Sorry about that. That's really annoying to have to do that. I should figure out how to get that test coverage without depending on the line numbers of start.zig 🤔

andrewrk (approved):
Changes look good. Super exciting milestone!

PR #10255, opened 2021-12-01
compiler_rt: add __paritysi2, __paritydi2, __parityti2

matu3ba:
- use Bit Twiddling Hacks: Compute parity in parallel
- test cases derived from popcount.zig
- tests: compare naive approach 10_000 times with random numbers created
  from naive seed 42
- compiler_rt.zig: sort by LLVM builtin order and add comments to improve structure

See #1290


matu3ba:
As in #9687 I tested locally the whole range of `u32`:
```zig
var a: u32 = 0;
while (a < 0xffffffff) {
    const x = @bitCast(i32, a);
    try test__paritysi2(x);
    a += 1;
}
const x = @bitCast(i32, a);
try test__paritysi2(x); // a == 0xffffffff
```

andrewrk (approved):
Nice work :+1:

PR #10256, opened 2021-12-01
Allocator: shrink can fail

leecannon:
Assuming that every allocator supports all shrinks is incorrect.

Imagine an allocator with a cache of 8 byte slabs and a cache of 16 byte slabs.
The user requests an allocation of 10 bytes, which is serviced from the 16 byte area, then if the user tries to shrink that allocation to 2 bytes the allocator has to reject that resize as when that memory is freed the slice will have a length associated with an 8 byte slab.
However if the resize was 10 to 9 bytes that would be successful as the information regarding the source slab size is not lost.

The only other way around this is to pass only the pointer of the slice into free and make it a requirement of allocators to function without the length. 

## Problems
Due to these problems this PR will stay draft

- `toOwnedSlice`
Now that shrink can fail the only way to maintain the status quo signature of `toOwnedSlice` is `catch unreachable`, this doesn't seem very zig-like.
We can't `catch self.items` as when that is passed to free the length is not the length of the actual allocation.

- `analyzeExport`
What do we set `eo_gop.value_ptr`/`de_gop.value_ptr` to if the realloc fails?

### DIFF
--- a/lib/std/array_list.zig
+++ b/lib/std/array_list.zig
@@ -96,7 +96,7 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
         /// The caller owns the returned memory. Empties this ArrayList.
         pub fn toOwnedSlice(self: *Self) Slice {
             const allocator = self.allocator;
-            const result = allocator.shrink(self.allocatedSlice(), self.items.len);
+            const result = allocator.realloc(self.allocatedSlice(), self.items.len) catch unreachable;
             self.* = init(allocator);
             return result;
         }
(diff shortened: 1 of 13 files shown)


marler8997:
I'm not certain whether this is better or worse but I think it might be worse.  Here are some of my thoughts.

This will make it easier to implement the allocator in a few cases, however, this will make it more difficult for the code that uses allocators to shrink allocations since in general you'll have to support the case when shrink fails.  I see you've realized this with your comment about `toOwnedSlice`, but your solution of using `catch unreachable` is not equivalent to the current behavior.  What would be equivalent is tracking 2 sizes on the caller side.  Whereas before, the allocator would only track the size if it needed to, now the caller must ALWAYS track the size even if doing so is not necessary.  I estimate that the allocator does not need to track this size in 99% of cases, but with this change now all calling code must track the extra size in all cases.

This new interface does have the advantage that you can get more information out of the allocator, namely, whether shrinking a size would just result in wasted space.  Code that had this information could decide whether it's better to keep the wasted space or better to move the memory to a smaller area with less waste.  However, I estimate that keeping the current semantics that shrink always works, and providing a new interface to get this extra information would be simpler overall.  This is because in most cases allocators can support shrink at no cost and we only have to pay the extra cost when the allocator can't support it by default.

These are just my unsubstantiated thoughts not backed by data so feel free to prove them wrong :)

andrewrk:
closing old draft

PR #10258, opened 2021-12-01
gpa: fix leak in freeLarge and memory limit accounting in resize and resizeLarge

mattbork:
Fixes two issues with GeneralPurposeAllocator since the big allocator change:
1. `freeLarge` never actually calls free on the backing allocator. This may explain the memory leak observed in https://github.com/ziglang/zig/pull/10055#issuecomment-983305273.

2. Memory limit accounting in `resize` and `resizeLarge` used `errdefers` that will never be run since `resize` now returns an optional type instead of an error union. Only a little rearranging is needed.


leecannon:
The `freeLarge` issue is already fixed, good catch with the memory limit accounting.

Issue #10181, opened 2021-11-20
libc (std.c) termios related constants missing and other problems


user673679:
### Remember to search before filing a new report

- [x] I searched for this bug and did not find it in the issue tracker, and furthermore, the title I used above will make this new bug report turn up in the search results for my query.

### Zig Version

0.9.0-dev.1678+7747bf07c

### Steps to Reproduce

I'm trying to use `std.os.termios` with libc on linux, compiling the following code:

```
    const termios = std.os.tcgetattr(std.os.STDOUT_FILENO) catch {
        return error.TcGetAttrFailed;
    };
    
    var t = termios;
    t.iflag &= ~(std.os.BRKINT | std.os.ICRNL | std.os.INPCK | std.os.ISTRIP | std.os.IXON);
    t.oflag &= ~(std.os.OPOST);
    t.cflag |= std.os.linux.CS8;
    t.lflag &= ~(std.os.ECHO | std.os.ICANON | std.os.IEXTEN | std.os.ISIG);
    t.cc[std.os.VMIN] = 0;
    t.cc[std.os.VTIME] = 0;

    std.os.tcsetattr(std.os.STDOUT_FILENO, std.os.TCSA.NOW, t) catch {
        return error.TcSetAttrFailed;
    };
```

### Expected Behavior

The code should compile!

### Actual Behavior

There are several problems here:

 * Although `std.os.termios` exists, the constants (`BRKINT` etc.) don't. The fix for this issue: https://github.com/ziglang/zig/issues/9707 added `termios` to `std.c`, but didn't add any of the others.

 * The `TCSA` enum is also missing from `std.c` (and thus `std.os`). Thus `std.os.tcsetattr()` can't be used with libc, even if `std.os.linux.TCSA.NOW` is specified explicitly in the code.

 * <strike>Some constants are defined in `std.os.linux` (e.g. `BRKINT`), but some are not (they're missing altogether). In the above code `VMIN` and `VTIME` aren't defined (perhaps there are also others?). This means that the code above can't be compiled under linux *without* libc either.</strike> Actually, I see these in the `std.os.linux.V` enum. But should they also be brought into `std.c` and `std.os`?

[...]

PR #10266, opened 2021-12-03
Added an explicit type for the termios constants

ZapAnton:
Adds the `tcflag_t` type to the termios constants.
This is made to allow bitwise operations on the termios
constants without an integer cast, e.g.:

```zig
var raw = try std.os.tcgetattr(std.os.STDIN_FILENO);
raw.lflag &= ~(std.os.linux.ECHO | std.os.linux.ICANON);
```
instead of

```zig
var raw = try std.os.tcgetattr(std.os.STDIN_FILENO);
raw.lflag &= ~@intCast(u32, std.os.linux.ECHO | std.os.linux.ICANON);
```

Contributes to https://github.com/ziglang/zig/issues/10181


andrewrk (changes requested):
Thanks for the PR. Please search for these symbols in the project; there are other matching definitions for OpenBSD and macOS that also need to be updated.

ZapAnton:
Done

andrewrk (approved):

andrewrk:
Thanks!

PR #10267, opened 2021-12-03
Add single section dumping and output file padding to InstallRawStep

N00byEdge:
This removes the need for using tools like `objcopy` to copy out a single section, and `truncate` to pad binaries to the required size while building.

### DIFF
--- a/lib/std/build/InstallRawStep.zig
+++ b/lib/std/build/InstallRawStep.zig
@@ -2,7 +2,7 @@ const std = @import("std");
 
 const Allocator = std.mem.Allocator;
 const ArenaAllocator = std.heap.ArenaAllocator;
-const ArrayList = std.ArrayList;
+const ArrayListUnmanaged = std.ArrayListUnmanaged;
 const Builder = std.build.Builder;
 const File = std.fs.File;
 const InstallDir = std.build.InstallDir;
@@ -17,6 +17,7 @@ const BinaryElfSection = struct {
     elfOffset: u64,
     binaryOffset: u64,
     fileSize: usize,
+    name: ?[]const u8,
     segment: ?*BinaryElfSegment,
 };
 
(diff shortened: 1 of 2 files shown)


lib/std/build/InstallRawStep.zig
```diff
@@ -298,7 +334,26 @@ fn containsValidAddressRange(segments: []*BinaryElfSegment) bool {
     return true;
 }
 
-fn emitRaw(allocator: Allocator, elf_path: []const u8, raw_path: []const u8, format: RawFormat) !void {
+pub const Options = struct {
+    format: ?RawFormat = null,
+    only_section_name: ?[]const u8 = null,
```
MasterQ32, line 339: Can you allow filtering for more than a single section? Maybe just allow that field to be a slice that will filter out sections when set to non-zero length

marler8997:
Part of your PR conflicts with my PR here https://github.com/ziglang/zig/pull/9975

Looks like we both had the same idea to add `Options` to `InstallRawStep`.

N00byEdge:
@marler8997 Oh wow our code looks really similar. That's a fun realization. Uh, I guess I could solve the merge conflict if your PR gets merged. Thanks for the heads up.

lib/std/build/InstallRawStep.zig
N00byEdge: Sure. Do you think a callback `includeSection(name: []const u8) bool` would be even better, or just the list?

N00byEdge: Also, I'm not sure dumping multiple sections through this is a good idea. The small difference between this and dumping all of them is that this doesn't seek to the section base address before writing to the file.

andrewrk:
#9975 merged. Apologies to both of you for the delay on that one.

N00byEdge:
Okay, I'll resolve the conflicts

andrewrk:
```
/workspace/lib/std/build/InstallRawStep.zig:350:87: error: use of undeclared identifier 'Options'
/workspace/lib/std/build/InstallRawStep.zig:428:10: error: use of undeclared identifier 'Options'
```

N00byEdge:
Oh, sorry about that, merge fail

PR #10269, opened 2021-12-04
stage1: add a compile error for using @bitCast to convert an enum to an integer

### LINKED ISSUE #3647: add a compile error for using @bitCast to convert an enum to an integer
```zig
const std = @import("std");

const Number = enum(u32) {
    one,
    two,
};

test "passes but should be a compile error" {
    const x: Number = .one;
    const y = @bitCast(u32, x);
}
```

The proper cast is `@enumToInt`:

```zig
const std = @import("std");

const Number = enum(u32) {
    one,
    two,
};

test "enum to int" {
    const x: Number = .one;
    const y = @enumToInt(x);
}
```

scheibo:
This fixes #3647 for stage 1. AFAICT though **this is a breaking change** (the accepted proposal doesn't have the breaking label, I think that was an oversight?), so might have to be careful about merging this.


andrewrk:
Thanks for the patch. At this point we need a self-hosted implementation for this to be useful since zig 0.10.0 will ship with stage3. Feel free to open a fresh PR if you want to pursue that.

PR #10270, opened 2021-12-04
Stage2: Initial behavior tests succeeding

Luukdegram:
With this, we get the initial few behavior tests succeeding for the wasm backend.

Note: During the implementation of this, I came across 2 memory leaks in stage2. The fixes for those are both included in this PR.


src/Module.zig
```diff
@@ -1226,19 +1226,20 @@ pub const Fn = struct {
...
         if (ret_ty.tag() == .generic_poison) {
             return null;
         }
         if (ret_ty.zigTypeTag() == .ErrorUnion) {
             if (ret_ty.errorUnionSet().castTag(.error_set_inferred)) |payload| {
-                return &payload.data.map;
+                return &payload.data;
             }
         }
         return null;
```
Luukdegram, line 1245: @andrewrk Let me know if you're okay with those changes to fix the memory leak

src/Module.zig
```diff
@@ -1301,6 +1302,7 @@ pub const Namespace = struct {
             key.destroy(mod);
         }
         anon_decls.deinit(gpa);
+        ns.usingnamespace_set.deinit(gpa);
```
Luukdegram, line 1305: Same as above (to fix a memory leak)

src/Module.zig
```diff
@@ -1332,6 +1334,8 @@ pub const Namespace = struct {
             child_decl.destroy(mod);
         }
         anon_decls.deinit(gpa);
+
+        ns.usingnamespace_set.deinit(gpa);
```
Luukdegram, line 1338: Same as bove

Luukdegram (commented):
Implemented some fixes for the memory leaks, but as my knowledge of this part of the compiler is quite limited, I'd like to make sure this is being reviewed.

andrewrk (approved):
Both memory leak fixes look great! Sorry for the trouble and thank you for fixing them.

PR #10276, opened 2021-12-05
sin/cos/tan musl reimplementation

tiehuis:
This reimplements sin, cos and tan functions in the math library and bases them on musl instead of the previous golang port.

I've tested all f32 values against musl and confirmed that the results are bit-identical using the following simple program.

```zig
const std = @import("std");
const c = @cImport(@cInclude("math.h"));

pub fn assertEqual(comptime name: []const u8, in: f32, x: f32, y: f32) !void {
    const bin = @bitCast(u32, in);
    const bx = @bitCast(u32, x);
    const by = @bitCast(u32, y);

    if (bx != by) {
        std.debug.print(
            \\name: {s}
            \\in : {b} ({})
            \\expected {}, found {}
            \\zig: {b}
            \\c  : {b}
            \\
        , .{ name, bin, in, x, y, bx, by });
        return error.NotEqual;
    }
}

const start = 0;
const progress = 10_000_000;

const test_sin = true;
const test_cos = true;
const test_tan = true;

pub fn main() !void {
    var x: u32 = start;
    while (true) : (x += 1) {
        if (x % progress == 0) {
            std.debug.print("{}\n", .{x});
        }

        const y = @bitCast(f32, x);

        if (test_sin) {
            const z_sin = std.math.sin(y);
            const c_sin = c.sinf(y);
            assertEqual("sin", y, z_sin, c_sin) catch {};
        }

        if (test_cos) {
            const z_cos = std.math.cos(y);
            const c_cos = c.cosf(y);
            assertEqual("cos", y, z_cos, c_cos) catch {};
        }

        if (test_tan) {
            const z_tan = std.math.tan(y);
            const c_tan = c.tanf(y);
            assertEqual("tan", y, z_tan, c_tan) catch {};
        }

        if (x == std.math.maxInt(u32)) break;
    }
}
```

Built with

```
zig build-exe -target x86_64-linux-musl test.zig -lc -D ReleaseFast
```

I've also tested a few million f64 cases using afl++ and @squeek502's [fuzzing code](https://github.com/squeek502/zig-std-lib-fuzzing) and found no errors so far.

I haven't made an effort to make the code particular idiomatic to zig at this stage.

### DIFF
--- a/lib/std/math/cos.zig
+++ b/lib/std/math/cos.zig
@@ -1,11 +1,16 @@
-// Ported from go, which is licensed under a BSD-3 license.
-// https://golang.org/LICENSE
+// Ported from musl, which is licensed under the MIT license:
+// https://git.musl-libc.org/cgit/musl/tree/COPYRIGHT
 //
-// https://golang.org/src/math/sin.go
+// https://git.musl-libc.org/cgit/musl/tree/src/math/cosf.c
+// https://git.musl-libc.org/cgit/musl/tree/src/math/cos.c
 
 const std = @import("../std.zig");
 const math = std.math;
 const expect = std.testing.expect;
+
+const kernel = @import("__trig.zig");
+const __rem_pio2 = @import("__rem_pio2.zig").__rem_pio2;
+const __rem_pio2f = @import("__rem_pio2f.zig").__rem_pio2f;
 
 /// Returns the cosine of the radian value x.
 ///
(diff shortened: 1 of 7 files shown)


lib/std/math/cos.zig
```diff
@@ -15,109 +20,135 @@ const expect = std.testing.expect;
...
-    const epsilon = 0.000001;
-
-    try expect(math.approxEqAbs(f32, cos_(f32, 0.0), 1.0, epsilon));
-    try expect(math.approxEqAbs(f32, cos_(f32, 0.2), 0.980067, epsilon));
-    try expect(math.approxEqAbs(f32, cos_(f32, 0.8923), 0.627623, epsilon));
-    try expect(math.approxEqAbs(f32, cos_(f32, 1.5), 0.070737, epsilon));
-    try expect(math.approxEqAbs(f32, cos_(f32, -1.5), 0.070737, epsilon));
-    try expect(math.approxEqAbs(f32, cos_(f32, 37.45), 0.969132, epsilon));
-    try expect(math.approxEqAbs(f32, cos_(f32, 89.123), 0.400798, epsilon));
+    const epsilon = 0.00001;
```
tiehuis, line 121: I needed to reduce these epsilons since the calculation method changed and the test cases otherwise would need adjustment.

lib/std/math/cos.zig
```diff
@@ -15,109 +20,135 @@ const expect = std.testing.expect;
...
-const pi4c = 2.69515142907905952645E-15;
-const m4pi = 1.273239544735162542821171882678754627704620361328125;
-
-fn cos_(comptime T: type, x_: T) T {
-    const I = std.meta.Int(.signed, @typeInfo(T).Float.bits);
-
-    var x = x_;
-    if (math.isNan(x) or math.isInf(x)) {
-        return math.nan(T);
+pub fn cos32(x: f32) f32 {
```
tiehuis, line 29: These shouldn't be public.

andrewrk:
Thanks!

PR #10279, opened 2021-12-05
spelling mistakes

AvdN:
Past and perfect tense of to cast is cast, not casted (which isn't a correct modern English word)

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -1260,7 +1260,7 @@ test "expectEqual demo" {
 
     // The first argument to `expectEqual` is the known, expected, result.
     // The second argument is the result of some expression.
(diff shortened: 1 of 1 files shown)


andrewrk:
These changes are debatable, and not very helpful. I would respectfully ask you to not submit any more grammar/spelling patches.

AvdN:
You can always debate whether something is an error or just a change in language. I am a linguist by education, so I know. But especially as a non-native speaker, I get confused with unfamiliar words and have to look them up to find if they have special meaning, only to find they have no (agreed upon) meaning at all. 
Please keep your documentation accessible for non-native speakers, by keeping the spelling/grammar free of errors and neologisms.

Issue #9439, opened 2021-07-22
Subcompilation fails when build-exe executed in parallel


thirdsgames:
Executing the zig compiler (specifically `zig build-exe`) multiple times in parallel can *occasionally* cause subcompilations to fail. This issue is present on both Ubuntu and Windows, running from the master branch. A sample of error messages:

```
Locally on Windows:
error: sub-compilation of compiler_rt failed
    C:\Users\<name>\.quill\compiler-deps\zig\lib\zig\std\special\compiler_rt\compareXf2.zig:10:21: error: unable to load 'C:\\Users\\<name>\\AppData\\Local\\zig\\o\\1e050970caac7a47b47d926ff10336fa\builtin.zig': UnexpectedEndOfFile

CI on Ubuntu:
error: sub-compilation of compiler_rt failed
    /opt/hostedtoolcache/zig/zig-linux-x86_64-0.9.0-dev.453+7ef854682/x64/lib/std/special/compiler_rt/mulodi4.zig:6:25: error: unable to load '/home/runner/.cache/zig/o/7b7987c94a9484a14de4c027d3e08478/builtin.zig': UnexpectedEndOfFile

CI on Windows:
error: sub-compilation of libssp failed
    C:\hostedtoolcache\windows\zig\zig-windows-x86_64-0.9.0-dev.477+00e944f71\x64\lib\zig\std\start.zig:8:22: error: unable to load 'C:\\Users\\runneradmin\\AppData\\Local\\zig\\o\\0e84b4fe2fc2fbd9bba0df21f28dc564\builtin.zig': UnexpectedEndOfFile
```

This can happen during CI runs ([1](https://github.com/quill-lang/quill/runs/3043321440?check_suite_focus=true), [2](https://github.com/quill-lang/quill/runs/3110638911?check_suite_focus=true)). In this project, I'm using `zig` as essentially a portable linker, so the test suite calls `build-exe` dozens of times in different threads. Without changes to my code, the same test can succeed or fail seemingly at random.

At a guess, it looks like there might be a race condition between different invocations of the compiler writing to and reading from files.

alexrp:
Out of curiosity, could you try setting the `ZIG_LOCAL_CACHE_DIR` environment variable to a directory local to each compilation and see if it still happens?

thirdsgames:
Running the same tests (with unique `ZIG_LOCAL_CACHE_DIR`) about 10 times eventually caused an error. This time it was in the linking stage.

```
[...]

PR #10280, opened 2021-12-05
macho: add preliminary non-prealloc path to the linker

### LINKED ISSUE #9990: larger than expected macOS binaries
### Zig Version

0.8.1 and 0.9.0-dev.1433+4a76523b9

### Steps to Reproduce

mylib.c:

```c
#include <stdio.h>
#include <math.h>

int foobar (int n) {
  printf("Hello World");

  if (n <= 10) {
    return pow(2, n);
  } else {
    return pow(n, 2);
  }
}
```

build.zig:

```zig
const std = @import("std");

    const lib = b.addSharedLibrary("mylib", "mylib.c", b.version(0, 0, 1));
    lib.linkLibC();
    lib.setTarget(target);
    lib.setBuildMode(mode);
    lib.install();
}
```

Finally, run `zig build -Drelease-fast -Dtarget=x86_64-macos`

### Expected Behavior

The shared library should work properly on macOS (for me it's macOS 10.13).

### Actual Behavior

I tested in zig 0.8.1 and 0.9.0-dev.1433+4a76523b9, and the issue is different is these two versions.

With zig 0.8.1, when I try to use the shared library in macOS, I got an error: `Symbol not found: dyld_stub_binder`.

With zig 0.9.0-dev.1433+4a76523b9, I can use the shared library in macOS without errors, but the shared library size is almost 40 times larger. (with zig 0.8.1, `libmylib.0.0.1.dylib` is 8.7KB; and with zig 0.9.0-dev.1433+4a76523b9, `libmylib.0.0.1.dylib` is 345KB)

kubkon:
The non-prealloc path is currently enabled for stage1 by default, and in the future, it might also apply to release builds if we decide not to have incremental updates then. What it means is that, when the linker uses this codepath, it doesn't preallocate segment and section sizes, much like it's done in any traditional linker.

The reason I reintroduced this is to minimize the binary size blow up that many have observed when building with stage1. With this PR, we now get the following binary size improvements (linking speed is largely the same for both codepaths):

_(everything measured on my M1 MBA)_
* __stage2 (with `-Dlog -Denable-llvm -Dlink-snapshot` flags)__
    * 17.5% size reduction from 237.8MB down to 196.1MB
* __multi-file small C program__
    * 86.8% size reduction from 378KB down to 49KB
* Andrew's __zig-sdl__ library
    * 17.4% size reduction from 292.6MB down to 241.6MB

Fixes #9990

### DIFF
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -925,57 +925,21 @@ pub fn flushModule(self: *MachO, comp: *Compilation) !void {
             try self.allocateDataConstSegment();
             try self.allocateDataSegment();
             self.allocateLinkeditSegment();
-            try self.allocLocalSymbols();
+            try self.allocateLocals();
         }
 
-        try self.allocateGlobalSymbols();
+        try self.allocateGlobals();
 
-        log.debug("locals:", .{});
-        for (self.locals.items) |sym, id| {
-            log.debug("  {d}: {s}: {}", .{ id, self.getString(sym.n_strx), sym });
-        }
-        log.debug("globals:", .{});
-        for (self.globals.items) |sym, id| {
-            log.debug("  {d}: {s}: {}", .{ id, self.getString(sym.n_strx), sym });
-        }
-        log.debug("undefs:", .{});
-        for (self.undefs.items) |sym, id| {
-            log.debug("  {d}: {s}: {}", .{ id, self.getString(sym.n_strx), sym });
-        }
-        {
-            log.debug("resolver:", .{});
-            var it = self.symbol_resolver.iterator();
-            while (it.next()) |entry| {
-                log.debug("  {s} => {}", .{ self.getString(entry.key_ptr.*), entry.value_ptr.* });
-            }
+        if (build_options.enable_logging) {
+            self.logSymtab();
+            self.logSectionOrdinals();
         }
 
-        log.debug("GOT entries:", .{});
-        for (self.got_entries_map.keys()) |key| {
-            switch (key) {
-                .local => |sym_index| log.debug("  {} => {d}", .{ key, sym_index }),
-                .global => |n_strx| log.debug("  {} => {s}", .{ key, self.getString(n_strx) }),
-            }
+        if (use_stage1) {
+            try self.writeAllAtoms();
+        } else {
+            try self.writeAtoms();
         }
-
-        log.debug("stubs:", .{});
-        for (self.stubs_map.keys()) |key| {
-            log.debug("  {} => {s}", .{ key, self.getString(key) });
-        }
-
-        for (self.section_ordinals.keys()) |match, i| {
-            const seg = self.load_commands.items[match.seg].Segment;
-            const sect = seg.sections.items[match.sect];
-            log.debug("{d}: {d},{d} == {s},{s}", .{
-                i + 1,
-                match.seg,
-                match.sect,
-                commands.segmentName(sect),
-                commands.sectionName(sect),
-            });
-        }
-
-        try self.writeAtoms();
 
         if (self.bss_section_index) |idx| {
             const seg = &self.load_commands.items[self.data_segment_cmd_index.?].Segment;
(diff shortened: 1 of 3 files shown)


andrewrk:
Looks like this addresses #9990. Would you say it closes it?

kubkon:
> Looks like this addresses #9990. Would you say it resolves it?

Nice one, well spotted! Let me link this issue as resolved by this PR. Thanks Andrew!

andrewrk:
Possibly relevant to this, in a local branch I have this diff:

```diff
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -69,6 +69,8 @@ failed_c_objects: std.AutoArrayHashMapUnmanaged(*CObject, *CObject.ErrorMsg) = .
 /// Miscellaneous things that can fail.
 misc_failures: std.AutoArrayHashMapUnmanaged(MiscTask, MiscError) = .{},
 
+cache_mode: CacheMode,
+
 keep_source_files_loaded: bool,
 use_clang: bool,
 sanitize_c: bool,
@@ -632,6 +634,8 @@ pub const ClangPreprocessorMode = enum {
 
 pub const SystemLib = link.SystemLib;
 
+pub const CacheMode = enum { incremental, whole };
+
 pub const InitOptions = struct {
     zig_lib_directory: Directory,
     local_cache_directory: Directory,
@@ -668,6 +672,7 @@ pub const InitOptions = struct {
     /// is externally modified - essentially anything other than zig-cache - then
     /// this flag would be set to disable this machinery to avoid false positives.
     disable_lld_caching: bool = false,
+    cache_mode: CacheMode = .incremental,
     object_format: ?std.Target.ObjectFormat = null,
     optimize_mode: std.builtin.Mode = .Debug,
     keep_source_files_loaded: bool = false,
```

This is actually part of a bug fix to #9439, as well as a compilation speed performance enhancement - idea being that for some compilations we actually want to opt out of incremental compilation and just cache the whole frickin thing. For example, there is no sense in incrementally building compiler-rt, which we build using the LLVM backend, using `-OReleaseFast`. With incremental compilation enabled, every compilation is waiting for LLVM to optimize the exact same compiler-rt LLVM module. Same with libc and other commonly shared files.

Anyway point being, I wanted to alert you to this addition because I get the sense that the linker code may benefit from observing this `cache_mode` setting.

kubkon:
> Possibly relevant to this, in a local branch I have this diff:
> 
> ```diff
> --- a/src/Compilation.zig
> +++ b/src/Compilation.zig
> @@ -69,6 +69,8 @@ failed_c_objects: std.AutoArrayHashMapUnmanaged(*CObject, *CObject.ErrorMsg) = .
>  /// Miscellaneous things that can fail.
>  misc_failures: std.AutoArrayHashMapUnmanaged(MiscTask, MiscError) = .{},
>  
> +cache_mode: CacheMode,
> +
>  keep_source_files_loaded: bool,
>  use_clang: bool,
>  sanitize_c: bool,
> @@ -632,6 +634,8 @@ pub const ClangPreprocessorMode = enum {
>  
>  pub const SystemLib = link.SystemLib;
>  
> +pub const CacheMode = enum { incremental, whole };
> +
>  pub const InitOptions = struct {
>      zig_lib_directory: Directory,
>      local_cache_directory: Directory,
> @@ -668,6 +672,7 @@ pub const InitOptions = struct {
>      /// is externally modified - essentially anything other than zig-cache - then
>      /// this flag would be set to disable this machinery to avoid false positives.
>      disable_lld_caching: bool = false,
> +    cache_mode: CacheMode = .incremental,
>      object_format: ?std.Target.ObjectFormat = null,
>      optimize_mode: std.builtin.Mode = .Debug,
>      keep_source_files_loaded: bool = false,
> ```
> 
> This is actually part of a bug fix to #9439, as well as a compilation speed performance enhancement - idea being that for some compilations we actually want to opt out of incremental compilation and just cache the whole frickin thing. For example, there is no sense in incrementally building compiler-rt, which we build using the LLVM backend, using `-OReleaseFast`. With incremental compilation enabled, every compilation is waiting for LLVM to optimize the exact same compiler-rt LLVM module. Same with libc and other commonly shared files.
> 
> Anyway point being, I wanted to alert you to this upcoming addition because I get the sense that the linker code may benefit from observing this `cache_mode` setting.

Nice! That's going to be very relevant for disambiguating the two code paths in stage2 I think :-)

kubkon:
Hmm, I don't think that failing aarch64 drone CI is my fault so I'm going to merge and if the failure persists on master, I'll try investigating what's up.

PR #10281, opened 2021-12-06
std.math.sign: get the sign of a int, float or vector

travisstaloch:
- returns -1, 0 or 1
- supports vector types
- branchless

### DIFF
--- a/lib/std/math.zig
+++ b/lib/std/math.zig
@@ -611,15 +611,15 @@ pub fn IntFittingRange(comptime from: comptime_int, comptime to: comptime_int) t
     if (from == 0 and to == 0) {
         return u0;
     }
-    const sign: std.builtin.Signedness = if (from < 0) .signed else .unsigned;
+    const signedness: std.builtin.Signedness = if (from < 0) .signed else .unsigned;
     const largest_positive_integer = max(if (from < 0) (-from) - 1 else from, to); // two's complement
     const base = log2(largest_positive_integer);
     const upper = (1 << base) - 1;
     var magnitude_bits = if (upper >= largest_positive_integer) base else base + 1;
-    if (sign == .signed) {
+    if (signedness == .signed) {
         magnitude_bits += 1;
     }
-    return std.meta.Int(sign, magnitude_bits);
+    return std.meta.Int(signedness, magnitude_bits);
 }
 
 test "IntFittingRange" {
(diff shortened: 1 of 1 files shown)


SpexGuy:
As long as it's switching on type info, might as well add an implementation for floats too :P

travisstaloch:
added float support. left a TODO for vectors of floats which is blocked by @intToFloat support for vector types.

travisstaloch:
looks like ci failure was just a timeout

lib/std/math.zig
```diff
@@ -1484,3 +1484,66 @@ test "boolMask" {
...
+                    uTx2{ expected, neg * expected },
+                    actual,
+                );
+            }
+        }
+    }
+    comptime {
+        try std.testing.expectEqual(-1, sign(-10));
+        try std.testing.expectEqual(1, sign(10));
+        try std.testing.expectEqual(0, sign(0));
```
matu3ba, line 1544: If you want to add more edge cases:
Add -0.0, 0.0 for floats (yes -0.0 is a valid floating point number), -inf, inf.
If you want to go the extra mile, add 2 denormal numbers (+ and -), but it should be fine without.

matu3ba (changes requested):
Looks solid, but tests should be improved for the quirky floating numbers and small integers.

daurnimator:
I struggle to see a use case for this function
  - Returning `0` for zero doesn't really allow for the signed-ness of floats (where there is both a positive and negative 0)
  - When would I use this function instead of writing e.g. `x > 0`

iddev5:
In order to solve the float confusion, the function behavior can be changed as follows: sign() would return -1, 0 or 1 normally on integer inputs but for floats, it would return -1 for negative floats and -0, and it will return 1 for positive floats and 0. The float version would not return 0 under any case.
Behavior inspired from rust and java strictmath

SpexGuy:
Sign as defined originally (0.0 and -0.0 return 0) is a common primitive in shading languages. (Though metal additionally specifies that sign(-0.0) is -0.0).
HLSL: [documentation for sign](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-sign)
GLSL: [documentation for sign](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/sign.xhtml)
Metal: [language spec pdf](https://www.google.com/url?sa=t&source=web&rct=j&url=https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf&ved=2ahUKEwiKs76H2o72AhVMl2oFHXjoDqcQFnoECAQQAQ&usg=AOvVaw1vNHYjUoHCMkpvZn98onPs) (section 6.2)
The primitive is useful for branchless programing, especially on platforms where multiplication is cheap and float->int conversion or reinterpretation is free (aka GPUs).

andrewrk (changes requested):
I have taken the liberty to squash your 3 commits into 1 and rebase against latest master.

I would be happy to add `std.math.sign` as you have defined it here, but I would like the following enhancements first:

 * avoid using `inline for` in unit tests
 * use `inline` on the sign function, since "branchless" is one of the features

travisstaloch:
Thanks @andrewrk. I changed the loops from inline to comptime and made math.sign an inline function.  Hope this is what you meant.  Let me know if this needs additional changes.  :+1:

andrewrk (changes requested):
Well now it's only being tested at compile time, and not getting as much runtime coverage. Sorry for not being clear, what I mean is to **avoid using loops entirely in your unit tests.** This makes it easy to see exactly what cases are being tested, and comment some of them out, stack traces become more useful, etc.

Get rid of `input_expecteds`. Get rid of the list of types. Just hard code every test case directly, and don't over engineer it. You can use function calls to simplify the logic.

Here's an example of what I am looking for:

https://github.com/ziglang/zig/blob/090461a6959ccebbc0c79cacebcf52bc43d66b9e/lib/std/math.zig#L778-L809

One final request: please make this only a single commit, to make merging easier for me :pray:

andrewrk:
btw isn't this just `return copysign(1.0, x)` ?

travisstaloch:
Apologies for my sloppy git skills. I did my best to 'squash' them into one.  

> btw isn't this just `return copysign(1, x)` ?

Not sure about using copysign.  First i've heard of it.  Does it support vectors? If you think that would be better to use for non-vectors, sounds good to me.

iddev5:
> btw isn't this just return copysign(1, x) ?
The range of signum is {-1, 0, 1} while copysign(1, x) will just return -1 or +1 depending on sign. 

Copysign right now is unimplemented for integers and vectors too.

andrewrk:
@iddev5 I think all the math functions should be improved to take `anytype` and handle vectors and integers where appropriate, and that includes `copysign`. However, your point about the set of possible output values is spot on.

iddev5:
One more things is that signum would be defined different for float values than int values. See discussion above. Sadly different languages implement it differently (that behavior of +0 and -0). This is not something which signum suffers from in mathematics.

andrewrk (approved):
This is good to merge once the CI is green. Note to person merging: upon merging fix the commit message

PR #10285, opened 2021-12-06
std: Progress.completed_items doesn't need to be +1, fixes #10277

### LINKED ISSUE #10277: Incorrect number of ran tests
### Zig Version

0.9.0-dev.1815+20e19e75f

### Steps to Reproduce

1. Create a test file `test.zig` with code like this (it has 3 tests):
```zig
const std = @import("std");

test "expect this to fail" {
    try std.testing.expect(false);
}

test "expect this to succeed" {
    try std.testing.expect(true);
}

test "this will be skipped" {
    return error.SkipZigTest;
}
```
2. Run test file using `zig test ./test.zig`

### Expected Behavior

Test runner should show that **3/3** tests have been run.

### Actual Behavior

```zig
❯ zig test .\test.zig

Code Generation [684/1123] std.hash_map.HashMapUnmanaged(usize,std.heap.general_p
Test [2/3] test "expect this to fail"... FAIL (TestUnexpectedResult)
C:\zig-0.9.0\lib\std\testing.zig:306:14: 0x7ff777c62ddb in std.testing.expect (test.obj)
    if (!ok) return error.TestUnexpectedResult;
             ^
C:\Dev\zig\learn-zig\src\test.zig:4:5: 0x7ff777c617a3 in test "expect this to fail" (test.obj)
    try std.testing.expect(false);
    ^
Test [4/3] test "this will be skipped"... SKIP
1 passed; 1 skipped; 1 failed.
.\zig-cache\o\aaabc098d02d8cbe1113cc2a4aeec21f\test.exe C:\zig-0.9.0\zig.exe
```

Test runner reports that **4/3** have been run.

nektro:
`Progress.completed_items` was unnecessarily `+ 1`d making it so it would start at `2` and go up to `total + 1` after a single item had been completed

New output:

image

### DIFF
--- a/lib/std/Progress.zig
+++ b/lib/std/Progress.zig
@@ -266,11 +266,11 @@ fn refreshWithHeldLock(self: *Progress) void {
                 }
                 if (eti > 0) {
                     if (need_ellipse) self.bufWrite(&end, " ", .{});
-                    self.bufWrite(&end, "[{d}/{d}] ", .{ completed_items + 1, eti });
+                    self.bufWrite(&end, "[{d}/{d}] ", .{ completed_items, eti });
                     need_ellipse = false;
                 } else if (completed_items != 0) {
                     if (need_ellipse) self.bufWrite(&end, " ", .{});
-                    self.bufWrite(&end, "[{d}] ", .{completed_items + 1});
+                    self.bufWrite(&end, "[{d}] ", .{completed_items});
                     need_ellipse = false;
                 }
             }


nektro:
Alternatively, we can edit https://github.com/ziglang/zig/blob/master/lib/std/special/test_runner.zig and move the `node.end();` calls farther down after the line printing. Depends on if you want the general usage of `std.Progress` to start at `0` or `1` and where the `node.end();` be to obtain that result.

nektro:
I went with the approach of editing `std.Progress` to fix the issue since this way it lets the consumer decide if they want to start at `0` or `1` based on where they place `node.end()` in their code

iddev5:
Same as #10282

andrewrk:
merged with #10282

PR #10288, opened 2021-12-06
Add test executable builds to build.zig

SpexGuy:
This adds the addTestExe function, which creates step that will build an installable artifact for a test build.  The installed artifact can then be run in a debugger.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -320,6 +320,14 @@ pub const Builder = struct {
         return LibExeObjStep.createTest(self, "test", root_src.dupe(self));
     }
 
+    pub fn addTestExe(self: *Builder, name: []const u8, root_src: []const u8) *LibExeObjStep {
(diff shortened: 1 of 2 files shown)


lib/std/build.zig
```diff
@@ -2542,89 +2561,93 @@ pub const LibExeObjStep = struct {
...
-                .qemu => |bin_name| ok: {
-                    if (builder.enable_qemu) qemu: {
-                        const glibc_dir_arg = if (need_cross_glibc)
-                            builder.glibc_runtimes_dir orelse break :qemu
-                        else
-                            null;
+        if (self.kind == .@"test") {
+            if (self.exec_cmd_args) |exec_cmd_args| {
+                for (exec_cmd_args) |cmd_arg| {
+                    if (cmd_arg) |arg| {
```
SpexGuy, line 2567: This got mangled in the diff but it's just an indentation change, under the condition `self.kind == .@"test"`.  There's also an added else branch at the bottom.

PR #10289, opened 2021-12-06
ci: linux: supply on_master_sucess with VERSION

mikdusan:
- move on_master_success to the package step
- add required secret availability for packaging

This replaces a1b79ea5857a52332d9da49d261a7fd04e52a466 .


CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ if("${ZIG_VERSION}" STREQUAL "")
     endif()
 endif()
 message("Configuring zig version ${ZIG_VERSION}")
+file(WRITE ${CMAKE_BINARY_DIR}/zig_version.txt ${ZIG_VERSION})
```
andrewrk, line 67: I'm not ok with this, let's not rely on cmake to write the zig version. surely it can be solved a better way.

andrewrk (changes requested):

andrewrk:
why do we have so many steps in drone.yml? it should just do the on_master_success logic inside linux_package.sh.

mikdusan:
> why do we have so many steps in drone.yml? it should just do the on_master_success logic inside linux_package.sh.

Upon further investigation, here are our 3 options (Initially I did it at the pipeline level and did not see step level support for itt):

1. conditional/logic for on_master_success inside linux_package.sh
2. conditional for on_master_success at the pipeline `.yml` level, logic in `linux_on_master_success.sh`
3. conditional for on_master_success at the step `.yml` level, logic in `linux_on_master_success.sh`

num2 does not share the same pipeline volume as build/test steps. So that's ruled out as per the mess it is creating with re-computing zig version.

num3 should be easy enough to flip to. The value of having it as an individual step is it shows up as a step in drone ui and will only show up when master push. Negative is this is leveraging a specific ci feature, and although it may be clearer, it's not as portable as num1.

Looks like the choice is 1 or 3. I'll implement your preference. Thanks!

PR #10291, opened 2021-12-07
stage2: fix TracyAllocator bugs

leecannon:
As `TracyAllocator` is not in the stdlib and is only conditionally compiled in stage2 I missed it in some of the later allocgate commits.


andrewrk:
Thanks for the follow-up!

Issue #3600, opened 2019-11-05
Proposal: @prefetch builtin


data-man:
[Accepted Proposal](https://github.com/ziglang/zig/issues/3600#issuecomment-980481986)

-----

[LLVM docs](https://llvm.org/docs/LangRef.html#llvm-prefetch-intrinsic)
```llvm
@llvm.prefetch(i8* <address>, i32 <rw>, i32 <locality>, i32 <cache type>)
```

Possible syntax:
```zig

pub const FetchType = enum {
    Read,
    Write
}

pub const CacheType = enum {
    Instruction,
    Data
}

pub const Locality = enum {
    No,
    ...,
    ...,
    Extremely
}

@prefetch(comptime fetchType: FetchType,
          comptime cacheType: CacheType,
          comptime locality: Locality,
          ptr: var) noreturn
```

Related: #767

shawnl:
yeah I like it, except that isData should be a enum with `.Data` and `.Instruction`.

data-man:
Updated.

andrewrk:
To go into std.builtin:

```zig
/// This data structure is used by the Zig language code generation and
/// therefore must be kept in sync with the compiler implementation.
pub const Prefetch = struct {
    rw: Rw = .read,
    /// 0 means no locality.
    /// 3 means extremely local, keep in cache.
    locality: u2 = 3,
    type: Type = .data,

    pub const Rw = enum { read, write };
    pub const Type = enum { instruction, data };
};
```

`@prefetch(ptr: anytype, options: Prefetch) void`

Any objections?

andrewrk:
Landed in 97c0373fa7ee125d2d4e73f9b1dc8ea0e2c2e86a just in time for 0.9.0.

PR #10295, opened 2021-12-07
Implement @prefetch()

ifreund:
Implements #3600

I've made two small style changes to the proposed interface:

1. I went with `std.builtin.PrefetchOptions` instead of `std.builtin.Prefetch` to match `CallOptions`, `ExportOptions`, etc.
2. I called the third field `cache` instead of `type` as I think it is more clear. However we may actually want to remove this option outright for the time being...

It seems that passing 0 as the third argument to llvm's prefetch builtin to indicate that the prefetch should be performed on the instruction cache causes LLVM to error out with something like this:
```
LLVM ERROR: Cannot select: 0x562197ba83f8: ch = Prefetch<(load (s8) from %ir.9)> 0x56219b37a0b8, 0x562197ba8598, Constant:i32<0>, Constant:i32<0>, Constant:i32<0>, foo.zig:28:9
  0x562197ba8598: i64 = add 0x562197ba87a0, 0x562197ba8600, foo.zig:28:26
    0x562197ba87a0: i64,ch = CopyFromReg 0x56219b37a0b8, Register:i64 %5, foo.zig:28:26
      0x562197ba8870: i64 = Register %5
    0x562197ba8600: i64 = shl 0x562197ba8738, Constant:i8<2>, foo.zig:28:26
      0x562197ba8738: i64 = srl 0x562197ba8390, Constant:i8<2>, foo.zig:28:32
        0x562197ba8390: i64,ch = CopyFromReg 0x56219b37a0b8, Register:i64 %1, foo.zig:25:29
          0x562197ba8460: i64 = Register %1
        0x562197ba8328: i8 = Constant<2>
      0x562197ba8328: i8 = Constant<2>
  0x562197ba82c0: i32 = Constant<0>
  0x562197ba82c0: i32 = Constant<0>
  0x562197ba82c0: i32 = Constant<0>
In function: binary_search
zsh: abort      zig build-lib /tmp/foo.zig -OReleaseFast
```

I'm not sure what exactly this error means, but it seems to indicate lack of support for prefetch using the instruction cache. Note that gcc and clang don't expose this as an option in their `__builtin_prefetch()`: they only support prefetching for the data cache. I therefore propose that we remove the `cache` field from our prefetch options for now, we can always add it back in the future if llvm gains better support for it and someone has a use-case.


ifreund:
It seems that LLVM is able to compile the prefetch intrinsic targeting the instruction cache for some architectures, including aarch64 and powerpc64. The fact that it fails to do so when targeting x86 seems to be an LLVM bug which was reported in 2014: https://bugs.llvm.org/show_bug.cgi?id=21037

The plan for this PR is to keep the cache option and skip emitting the `llvm.prefetch` intrinsic when targeting x86 and any other targets that LLVM happens to fail on.

ifreund:
This should be ready for review now. I've added a behavior test and documentation for `@prefetch()` and implemented a workaround for the LLVM bug described above.

ifreund:
Welp, our wonderful CI matrix and the exhaustive behavior test I just added have revealed another case of this LLVM bug if a write prefetch is made targeting the instruction cache for the following target and mcpu:
```
-target arm-linux-none -mcpu=generic+has_v7+mp
```
I'll add this case to our workaround...

andrewrk (approved):

PR #10303, opened 2021-12-09
Add Vector documentation

bnprks:
I noticed Zig's SIMD support is quite far along now, so it seems like a good time to add a proper section in the language documentation. My main goal is to have *something* in the language docs, so I'd be happy to take suggestions on edits or have someone more experienced write the docs.

Summary of changes:
1. Create brief documentation on Zig vector types with code example
2. Get rid of the SIMD sub-heading under the main Vectors heading,
and update links accordingly
3. Add an example to the shuffle docs

Although the previous placeholder suggested to include "Advice on writing high perf software, how to abstract the best way", I've omitted that in this version as my original draft's SIMD tips seemed a little out of place for core language documentation. I've left the TODO for adding information on C ABI interop alone as I don't know much about that topic.

Preview of the documentation as built from this update:
https://htmlpreview.github.io/?https://gist.githubusercontent.com/bnprks/b6134ce638fbe1b10c7a871d6abda26b/raw/2c013d8649d93fd258f55118ff9e7a889e9a9fac/langref.html#Vectors

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -2483,23 +2483,84 @@ test "null terminated array" {
       {#header_open|Vectors#}
       <p>
       A vector is a group of booleans, {#link|Integers#}, {#link|Floats#}, or {#link|Pointers#} which are operated on
-      in parallel using a single instruction ({#link|SIMD#}). Vector types are created with the builtin function {#link|@Type#},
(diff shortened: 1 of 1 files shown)


andrewrk:
Thanks!

PR #10311, opened 2021-12-11
compiler_rt: add __bswapsi2, __bswapdi2 and __bswapti2

matu3ba:
- each byte gets masked, shifted and combined
- use boring masks instead of comptime for readability
- tests: bit patterns with reverse operation, if applicable

See #1290


matu3ba:
1. Integral Bit Manipulation is complete with this commit, see [here](https://github.com/ziglang/zig/issues/1290#issuecomment-819899869) for status. :) :)
2. I added `bswapti`, just because I find this [C behavior annoying](https://stackoverflow.com/questions/11656241/how-to-print-uint128-t-number-using-gcc) and because I found these answers on stackoverflow [about byteswapping sad](https://stackoverflow.com/questions/8004790/how-to-change-byte-order-of-128-bit-number).

andrewrk:
Nice work!

Issue #10299, opened 2021-12-08
[darwin aarch64 cgo] regression in 0.8.0-1366-gfc302f00a


motiejus:
### Zig Version

0.8.0-1921-g83a668195

### Steps to Reproduce

Identical to #10297, but substitute:
- `x86_64` with `aarch64`.
- `amd64` with `arm64`.

Step-by-step:

##  `hello.go`
```
package main

// #include <stdio.h>
// void helloworld() { printf("hello, world\n"); }
import "C"

func main() {
	C.helloworld()
}
```

## `~/test`
```
#!/bin/bash
set -xe

pushd $HOME/dev/zig/build
cmake ..
make -j$(nproc)
popd
GOOS=darwin GOARCH=arm64 CC=$HOME/zcc CGO_ENABLED=1 go build -a -buildmode=pie -ldflags "-s -w" $HOME/hello.go
```

## `~/zcc`
```
user@motiejus:~/dev/zig$ cat ~/zcc 
#!/bin/bash
zig=$HOME/dev/zig/build/zig
exec "$zig" cc -target aarch64-macos-gnu "$@"
```

# Bisect Result

fc302f00a9de5de0490f4a66720e75946763c695 is the first bad commit, which doesn't revert cleanly.

```
$ git bisect good dde0adcb363f3a3f306c0fc9eaec511cc3b74965
$ git bisect bad master
$ git bisect run ~/test
<...>
fc302f00a9de5de0490f4a66720e75946763c695 is the first bad commit
commit fc302f00a9de5de0490f4a66720e75946763c695
Author: Jakub Konka <kubkon@jakubkonka.com>
Date:   Sun Oct 10 10:33:15 2021 +0200

    macho: redo relocation handling and lazy bind globals

    * apply late symbol resolution for globals - instead of resolving
      the exact location of a symbol in locals, globals or undefs,
      we postpone the exact resolution until we have a full picture
      for relocation resolution.
    * fixup stubs to defined symbols - this is currently a hack rather
      than a final solution. I'll need to work out the details to make
      it more approachable. Currently, we preemptively create a stub
      for a lazy bound global and fix up stub offsets in stub helper
      routine if the global turns out to be undefined only. This is quite
      wasteful in terms of space as we create stub, stub helper and lazy ptr
      atoms but don't use them for defined globals.
    * change log scope to .link for macho.
    * remove redundant code paths from Object and Atom.
    * drastically simplify the contents of Relocation struct (i.e., it is
[...]

Issue #10158, opened 2021-11-16
Zig CC link errors (Scrt1.o and crtn.o not found)


uhthomas:
### Zig Version

0.9.0-dev.1414+cde3dd365

### Steps to Reproduce

Using [bazel-zig-cc](https://sr.ht/~motiejus/bazel-zig-cc/)@[361fc42b](https://git.sr.ht/~motiejus/bazel-zig-cc/tree/361fc42b2e1282ed078bf9c1ad281642e19858d4), build a cgo project with `zig cc` and `pure = off`.

```py3
go_binary(
    name = "some",
    embed = [":some_lib"],
    goarch = "amd64",
    goos = "linux",
    pure = "off",
    visibility = ["//visibility:public"],
)
```

### Expected Behavior

It should build.

### Actual Behavior

```
(11:17:01) ERROR: /private/var/tmp/_bazel_thomas/b6a3c999aae05297a54a9aa13d7a50c3/external/io_bazel_rules_go/BUILD.bazel:44:7: GoStdlib external/io_bazel_rules_go/stdlib_/pkg failed: (Exit 1): builder failed: error executing command bazel-out/host/bin/external/go_sdk/builder stdlib -sdk external/go_sdk -installsuffix linux_amd64 -out bazel-out/darwin-fastbuild-ST-64800c94ce45/bin/external/io_bazel_rules_go/stdlib_

Use --sandbox_debug to see verbose messages from the sandbox
# runtime/cgo
WARNING: flock not found, proceeding unsafely.
If build fails, retry it.
ld.lld: error: cannot open /var/folders/d5/hqd0d3f17h7gmntg94k251z80000gn/T//bazel-zig-cc/o/fdb8c1e82b0cc8d72c53e7cb12c0ef9b/Scrt1.o: No such file or directory
ld.lld: error: cannot open /var/folders/d5/hqd0d3f17h7gmntg94k251z80000gn/T//bazel-zig-cc/o/1e260ac65a7a7021ed17a6f508533e22/crti.o: No such file or directory
ld.lld: error: cannot open /var/folders/d5/hqd0d3f17h7gmntg94k251z80000gn/T//bazel-zig-cc/o/965602877839941573a56b32b0c704a7/crtn.o: No such file or directory
```

andrewrk:
Duplicate of #9439.

andrewrk:
btw what is going on here?

```
WARNING: flock not found, proceeding unsafely.
```

[...]

PR #10312, opened 2021-12-11
macho: fail if requested -framework is not found

motiejus:
If `-framework` is requested, but not found, the linker will err
instead of creating a strange executable.

https://github.com/ziglang/zig/issues/10299#issuecomment-990404953

Refs #9542
Refs #10299
Refs #10158

### DIFF
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -751,7 +751,6 @@ pub fn flushModule(self: *MachO, comp: *Compilation) !void {
                 }
             }
 
-            var framework_not_found = false;
             for (self.base.options.frameworks) |framework| {
                 for (&[_][]const u8{ ".tbd", ".dylib", "" }) |ext| {
                     if (try resolveFramework(arena, framework_dirs.items, framework, ext)) |full_path| {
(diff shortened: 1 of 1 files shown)


kubkon:
Thanks @motiejus! Do you mind if I add a few tweaks to your proposed solution? Namely, I don't want the linker fail prematurely until it resolves all of the symbols first, and then the priority should be to fail with `UndefinedSymbolReference` before failing with `FrameworkNotFound`. This should give the user a better idea what symbols were unresolved if any. In case we are able to resolve every symbol but fail to find a framework that might be used at runtime, we then fail with `FrameworkNotFound`.

motiejus:
> Thanks @motiejus! Do you mind if I add a few tweaks to your proposed solution? Namely, I don't want the linker fail prematurely until it resolves all of the symbols first, and then the priority should be to fail with `UndefinedSymbolReference` before failing with `FrameworkNotFound`. This should give the user a better idea what symbols were unresolved if any. In case we are able to resolve every symbol but fail to find a framework that might be used at runtime, we then fail with `FrameworkNotFound`.

Sure, that sounds like a great improvement!

Issue #8180, opened 2021-03-07
Proposal: fix shared library linking on macOS


haberman:
# Problem Statement

It is currently impossible to use Zig to build shared (dynamic) libraries on macOS where some symbols are left undefined. For example, consider the program:

```zig
extern fn bar() void;

export fn foo() void {
    bar();
}
```

On Linux we can build a shared library from this without error:

```
$ zig build-lib -dynamic test.zig
$ ls
libtest.so  test.zig  zig-cache
```

However the same program compiled with the same command line fails on macOS:

```
$ zig build-lib -dynamic test.zig
warning(link): unexpected LLD stderr:
Undefined symbol: zig-cache/o/c4e95092e3ee25bb54ad9a24fb2ee9a5/test.o: _bar
symbol(s) not found

error: FileNotFound
```

This has previously been reported in https://github.com/ziglang/zig/issues/3085 and https://github.com/ziglang/zig/issues/8016.

# Motivation

This is a blocker for creating loadable extension modules for Python, Ruby, Lua, Postgres, etc. -- basically any program that loads modules via `dlopen()` and expects that the symbols will be resolved at load time. If you only need your module to work on Linux, then Zig will work. But if you need macOS support, there appears to be no way to make this work, and no workaround that I could find.

# Background

Both macOS and Linux have command line options that control whether symbols may be undefined at link time. However the options are different and the two platforms have opposite defaults. By default, Linux is lax and macOS is strict.

## Linux

Linux has two different options:
* `--no-undefined`: causes the link to fail if unresolved symbols are present in the library currently being built. This is unset by default, so Linux *allows* undefined references by default, as we observed in the problem statement.
* `--no-allow-shlib-undefined`: causes the link to fail if unresolved symbols are present in any shared libraries *that we are linking against* (eg. pre-existing shared libraries). This defaults to unset (ie. allowing undefined references) when you are building a shared library, but the default is the opposite (failing for undefined references) if you are building an executable. This is a somewhat odd option that I don't know a real use case for, and macOS doesn't have an analogue to this that I am aware of.

(There are also a few other flags such as `--unresolved-symbols=method` and `-z defs`, which are just different porcelain around the same underlying options, AFAICT).

More info in [this StackOverflow thread](https://stackoverflow.com/a/2356407/77070)

## macOS

macOS has just one option for this:
* `-undefined treatment`, where `treatment` is one of: `error`, `warning`, `suppress`, or `dynamic_lookup`. The default is `error`.

[...]

PR #10314, opened 2021-12-11
macho: allow undefined symbols in dylibs

### LINKED ISSUE #3000: Zig and Node.js N-API example
Given Zig's promise of C header includes, it would be cool for the docs or blog to have an example of writing a "Hello Node.js" native addon for Node.js using N-API and Zig.

Node's N-API is fairly recent, and many existing native addons for Node written using Nan etc. are probably being ported. Now might be a great time for some of these native Node modules to consider using Zig instead of C/C++.

kubkon:
We now respect both `-fallow-shlib-undefined` and `-Wl,"-undefined=dynamic_lookup"` flags. This is the first step towards solving issues #8180. We currently do not expose any other ld64 equivalent flag for `-undefined` flag - we basically throw an error should the user specify a different flag. Support for those is conditional on closing #8180. As a result of this change, it is now possible to generate a valid native Node.js addon with Zig for macOS.

You can also check whether the option is on by requesting `--verbose-link` when building a shared object like so:

```
$ zig build-lib -dynamic example.zig -I$(brew --prefix node)/include/node -fallow-shlib-undefined -femit-bin=example.node --verbose-link
zig ld -dynamic -dylib -install_name @rpath/example.node zig-cache/o/f33f9be186fda290c12c65772584b1cb/example.o /Users/kubkon/.cache/zig/o/7d88bd701accf5f7b37248e431dd6be8/libcompiler_rt.a -o example.node -lSystem -lc -undefined dynamic_lookup
```

Fixes #3000

cc @jorangreef @ifreund

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1703,6 +1703,16 @@ fn buildOutputType(
                     }
                     emit_implib = .{ .yes = linker_args.items[i] };
                     emit_implib_arg_provided = true;
+                } else if (mem.eql(u8, arg, "-undefined")) {
+                    i += 1;
+                    if (i >= linker_args.items.len) {
+                        fatal("expected linker arg after '{s}'", .{arg});
+                    }
+                    if (mem.eql(u8, "dynamic_lookup", linker_args.items[i])) {
+                        linker_allow_shlib_undefined = true;
+                    } else {
+                        fatal("unsupported -undefined option '{s}'", .{linker_args.items[i]});
+                    }
                 } else {
                     warn("unsupported linker arg: {s}", .{arg});
                 }
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -444,6 +444,7 @@ pub fn flushModule(self: *MachO, comp: *Compilation) !void {
     const is_dyn_lib = self.base.options.link_mode == .Dynamic and is_lib;
     const is_exe_or_dyn_lib = is_dyn_lib or self.base.options.output_mode == .Exe;
     const stack_size = self.base.options.stack_size_override orelse 0;
+    const allow_undef = is_dyn_lib and (self.base.options.allow_shlib_undefined orelse false);
 
     const id_symlink_basename = "zld.id";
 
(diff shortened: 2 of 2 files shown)


src/main.zig
```diff
@@ -1703,6 +1703,16 @@ fn buildOutputType(
                     }
                     emit_implib = .{ .yes = linker_args.items[i] };
                     emit_implib_arg_provided = true;
+                } else if (mem.eql(u8, arg, "-undefined")) {
```
andrewrk, line 1706: Quick question- you mentioned that this supports `-Wl,"-undefined=dynamic_lookup"`, but it looks like this actually implements `-Wl,-undefined,dynamic_lookup`. Is this intentional?

andrewrk (approved):

src/main.zig
kubkon: Uhm, wait, did I make a typo somewhere? It should implement the former, `-Wl,"-undefined=dynamic_lookup"` unless I messed something up.

kubkon: Hmm, I think that's what I did. Note that there is another `if` statement after we match `-undefined`:

```zig
    5                 } else if (mem.eql(u8, arg, "-undefined")) {
     4                     i += 1;
     3                     if (i >= linker_args.items.len) {
     2                         fatal("expected linker arg after '{s}'", .{arg});
     1                     }
  1711                     if (mem.eql(u8, "dynamic_lookup", linker_args.items[i])) {
     1                         linker_allow_shlib_undefined = true;
     2                     } else {
     3                         fatal("unsupported -undefined option '{s}'", .{linker_args.items[i]});
     4                     }
     5                 } else {
```

andrewrk: Never mind- I forgot that we had this code up there:

```zig
                            // Handle nested-joined args like `-Wl,-rpath=foo`.
                            // Must be prefixed with 1 or 2 dashes.
                            if (linker_arg.len >= 3 and linker_arg[0] == '-' and linker_arg[2] != '-') {
                                if (mem.indexOfScalar(u8, linker_arg, '=')) |equals_pos| {
                                    try linker_args.append(linker_arg[0..equals_pos]);
                                    try linker_args.append(linker_arg[equals_pos + 1 ..]);
                                    continue;
                                }
                            }
```

all good :+1:

jorangreef:
Thanks so much @kubkon and @andrewrk ! Going to go link a cup of coffee now...

Issue #5893, opened 2020-07-17
Rename `anytype` to `anyval`


SpexGuy:
Sorry, this is kind of nitpicking, but it's been bothering me.

The value of a variable of type `anyerror` is some error.
The value of a variable of type `anyframe` is some frame.
The value of a variable of type `anytype` is usually not a type.

Said another way,
A variable that holds some unknown error is an `anyerror`.
A variable that holds some unknown frame is an `anyframe`.
A variable that holds some unknown type is a `type`.
A variable that holds some unknown value is an `anytype`?

`var` was a bad name because it was overloaded with the mutability specifier.  But aside from that, IMO it was an ok name.  I don't think we should restore it but I think `anytype` is also confusing.  `any` might be a good choice except that it is a common enough variable or function name that the language probably shouldn't reserve it.  `infer` was suggested previously, but doesn't convey that the type can change over the variable's lifetime.  Maybe `anyvalue`, `anyval`, or `vartype` would be better?

Sobeston:
I agree, and personally like `anyvalue` and `anyval`.

data-man:
Important part was missed - declaration.

```
i: u32 // i is u32 variable
T: type // T can be type only
writer: anytype // writer is variable with any type, not unknown value
```
What's problem?

EleanorNB:
+1 for `anyval`. I think that's much better.

timmyjose:
> I agree, and personally like `anyvalue` and `anyval`.

Seconded. I found it terribly confusing when the ZIg formatter started changing my `var`S to `anytype`S. I would have expectd `anytype` to really be a type, but it's apparently just the new name to `var`.

timmyjose:
> Important part was missed - declaration.
> 
> ```
> i: u32 // i is u32 variable
> T: type // T can be type only
> writer: anytype // writer is variable with any type, not unknown value
> ```
> 
> What's problem?

Something like so:

```
const std = @import("std");

fn printAnyType(obj: anytype) void {
    std.debug.print("{}, {}, {}\n", .{ obj.@"foo", obj.@"bar", obj.@"baz" });
}

test "printAnyType" {
    printAnyType(.{ .foo = "Hello", .bar = 1, .baz = false });
}

```

where I would have assumed `anytype` to be a type, not a value. Instead, it is sort of like a generic object, a *value*. In a purely functional dependently-typed language like Idris, maybe types and values are truly equivalent, but in an imperative language, it feels jarring.

[...]

PR #10316, opened 2021-12-12
stage1, stage2: rename c_void to anyopaque

### LINKED ISSUE #323: proposal: rename c_void to anyopaque
```
pub fn main() -> %void {
    var x: usize = 0x1000;
    var y: &void = @intToPtr(&void, x);
}
```

This causes the compiler to segfault.

ifreund:
zig fmt now replaces c_void with anyopaque to make updating
code easy.

Closes #323


ifreund:
The force push just fixed a typo in the commit message.

I had a full `zig build test` run pass for me locally so hopefully CI will be green when I wake up in the morning :)

ifreund:
Looks like the only CI failure was the windows build which timed out after 6 hours. Seems like it did make it to the last std test target though (`std-native-ReleaseSmall-bare-single`).

lib/std/zig/render.zig
```diff
@@ -2290,7 +2290,12 @@ fn renderToken(ais: *Ais, tree: Ast, token_index: Ast.TokenIndex, space: Space)
     const token_start = token_starts[token_index];
     const lexeme = tokenSliceForRender(tree, token_index);
 
-    try ais.writer().writeAll(lexeme);
+    // TODO remove this c_void -> anyopaque rewrite after the 0.10.0 release
+    if (mem.eql(u8, lexeme, "c_void")) {
+        try ais.writer().writeAll("anyopaque");
+    } else {
+        try ais.writer().writeAll(lexeme);
+    }
```
andrewrk, line 2298: Hmmm this is actually a *little* bit problematic because the following code will be incorrectly transformed:

```zig
const S = struct {
    c_void: i32,
    anyopaque: i32,
};
```

to

```zig
const S = struct {
    anyopaque: i32,
    anyopaque: i32,
};
```

causing a duplicate field compile error. `c_void` and `anyopaque` are primitives; not keywords, and so things such as block labels and field names are allowed to be named with those tokens.

Maaaaybe it's OK to let this slide?

But I think it would be wise instead to have a special handling for when the AST node is an identifier node and only do the transformation in such case.

andrewrk (changes requested):
Found a problem with the zig fmt auto fixer; other than that this looks ready to merge.

lib/std/zig/render.zig
ifreund: Good catch, probably would have been OK to let slide but I found a fairly clean way to do it and pushed a commit improving this to only rewrite identifier nodes.

ifreund:
Ironically, I think `anytype` would make a bit more sense for this primitive type than `anyopaque` does, if it wasn't already a keyword. 

IMO we should replace the `anytype` keyword with `ducktype` (https://github.com/ziglang/zig/issues/5893#issuecomment-992633920) and replace `c_void` with `anytype` instead of `anyopaque`.

andrewrk:
I'm going to pull this in because it *is* an improvement over status quo in that it makes the type no longer imply that it has to do with C or `void`, and given that `zig fmt` auto-fixes it, it's pretty harmless as far as breaking changes go.

We can evaluate follow-up proposals separately.

PR #10319, opened 2021-12-12
Ability to control what Zig emits with function

### LINKED ISSUE #6871: Ability to control what Zig emits in build.zig
As far as I understand, right now you can't specify what Zig should emit in `build.zig`. I'd like to have the equivalent  `build.zig` config for the command: `zig build-obj src/main.zig -femit-llvm-ir -fno-emit-bin`

Here's what I have in my `build.zig`
```

    b.setPreferredReleaseMode(builtin.Mode.ReleaseFast);

    const obj = b.addObject("bitcode", "src/main.zig");
    obj.setBuildMode(mode);

    // Theses don't exist
    // obj.shouldEmitLLVMIR(true);
    // obj.shouldEmitBin(false);

    // test steps, etc
}
```

After looking in `lib/std/build.zig`, I think adding something like:
```
pub fn shouldEmitLLVMIR(self: *LibExeObjStep, should_emit: bool) void {
    self.emit_llvm_ir = should_emit;
}

pub fn shouldEmitBin(self: *LibExeObjStep, should_emit: bool) void {
    self.emit_bin = should_emit;
}

// etc for all other emit types
```

Then in `make` everything already seems setup to add the args!

If this is something you'd like, I can submit a PR!

AliChraghi:
im not sure this is needed. but in future this will may be useful. closes #6871

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -1992,6 +1992,34 @@ pub const LibExeObjStep = struct {
         self.linkSystemLibraryName(name);
     }
 
+    pub fn setAnalysisEmit(self: *LibExeObjStep, emit_option: EmitOption) void {
+        self.emit_analysis = emit_option;
+    }
+    
+    pub fn setAsmEmit(self: *LibExeObjStep, emit_option: EmitOption) void {
+        self.emit_asm = emit_option;
+    }
+
+    pub fn setBinEmit(self: *LibExeObjStep, emit_option: EmitOption) void {
+        self.emit_bin = emit_option;
+    }
+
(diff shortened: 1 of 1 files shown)


matu3ba:
Why does Zig need setter functions, if the user can just set and read
```zig
    emit_analysis: EmitOption = .default,
    emit_asm: EmitOption = .default,
    emit_bin: EmitOption = .default,
    emit_docs: EmitOption = .default,
    emit_implib: EmitOption = .default,
    emit_llvm_bc: EmitOption = .default,
    emit_llvm_ir: EmitOption = .default,
```
in the `pub const LibExeObjStep = struct` ?

AliChraghi:
oops i see

PR #10320, opened 2021-12-12
Stage2: Allow specifying which symbols to export (wasm)

Luukdegram:
Notating a symbol to be exported in code will only tell the linker where to find this symbol, so other object files can find it. However, this does not mean said symbol will also be exported to the host environment. Currently, we 'fix' this by force exporting every single symbol that is visible by appending the `--export-all` flag. This creates bigger binaries and means host environments have access to symbols that they perhaps shouldn't have. Now, users can tell Zig which symbols are to be exported, meaning all other symbols that are not specified will not be exported. This decreases the binary sizes of wasm binaries tremendously as it will no longer export symbols from compiler_rt, etc.

Another change is we now support `-rdynamic` in the wasm linker as well, meaning all symbols will be put in the dynamic symbol table. This is the same behavior as with ELF. This means there's a 3rd strategy users will have to build their wasm binary.

In a future iteration, we could automatically append `--export` flags based on exported functions, similar to what stage1 did before it switched to the hybrid approach.


andrewrk:
I wonder, should we automatically add `--export=foo` lines to the linker for .zig code which has `export fn ...` or use of `@export`? :thinking:

andrewrk:
> In a future iteration, we could automatically append `--export` flags based on exported functions, similar to what stage1 did before it switched to the hybrid approach.

I just now noticed this. Sorry for missing it earlier!

PR #10321, opened 2021-12-12
compiler_rt: add __negsi2, __negdi2, __negti2

matu3ba:
- call it negXi2.zig to prevent confusion with negXf2.zig
- used for size optimized builds and machines without carry instruction
- tests: special cases 0, -INT_MIN
  * use divTrunc range and shift with constant offsets

See #1290


matu3ba:
The test data was generated with this script https://gist.github.com/matu3ba/3fdecf00934bb064657b080ae2ff27a2.

This completes **Integral arithmetic**. :)

Next are . 1 case for Integral arithmetic which returns if overflow (refactor only needs test data), 2. Integral comparison (only 2 impls)

**Integral arithmetic with trapping overflow** wasnt touched yet and **Integral / floating point conversion** looks now like a mess to be cleaned up later.

matu3ba:
This is a use case for #10321 to print binary and hex data of 2s complement how it actually looks in memory.
In absence of doing that I used a decimal representation.

PR #10326, opened 2021-12-12
Add ArrayList.clone

schmee:
This is a common operation and matches the API on other stdlib collections like HashMap and MultiArrayList.

cc @SpexGuy, as discussed on Discord.

### DIFF
--- a/lib/std/array_list.zig
+++ b/lib/std/array_list.zig
@@ -110,14 +110,9 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
 
         /// Creates a copy of this ArrayList, using the same allocator.
         pub fn clone(self: *Self) !Self {
-            var items_copy = try self.allocator.alloc(T, self.capacity);
-            mem.copy(T, items_copy, self.items);
-            items_copy.len = self.items.len;
-            return Self{
-                .items = items_copy,
-                .capacity = self.capacity,
-                .allocator = self.allocator,
-            };
+            var cloned = try ArrayList(T).initCapacity(self.allocator, self.capacity);
+            cloned.appendSliceAssumeCapacity(self.items);
+            return cloned;
         }
 
         /// Insert `item` at index `n` by moving `list[n .. list.len]` to make room.
@@ -503,13 +498,9 @@ pub fn ArrayListAlignedUnmanaged(comptime T: type, comptime alignment: ?u29) typ
 
         /// Creates a copy of this ArrayList.
         pub fn clone(self: *Self, allocator: Allocator) !Self {
-            var items_copy = try allocator.alloc(T, self.capacity);
-            mem.copy(T, items_copy, self.items);
-            items_copy.len = self.items.len;
-            return Self{
-                .items = items_copy,
-                .capacity = self.capacity,
-            };
+            var cloned = try ArrayListUnmanaged(T).initCapacity(allocator, self.capacity);
+            cloned.appendSliceAssumeCapacity(self.items);
+            return cloned;
         }
 
         /// Insert `item` at index `n`. Moves `list[n .. list.len]`
@@ -837,8 +828,8 @@ test "std.ArrayList/ArrayListUnmanaged.clone" {
         defer cloned.deinit();
 
         try testing.expectEqualSlices(i32, array.items, cloned.items);
-        try testing.expectEqual(array.capacity, cloned.capacity);
         try testing.expectEqual(array.allocator, cloned.allocator);
+        try testing.expect(cloned.capacity >= array.capacity);
 
         array.deinit();
 
@@ -856,7 +847,7 @@ test "std.ArrayList/ArrayListUnmanaged.clone" {
         defer cloned.deinit(a);
 
         try testing.expectEqualSlices(i32, array.items, cloned.items);
-        try testing.expectEqual(array.capacity, cloned.capacity);
+        try testing.expect(cloned.capacity >= array.capacity);
 
         array.deinit(a);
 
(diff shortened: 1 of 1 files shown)


InKryption:
Why not use `Allocator.dupe`?

schmee:
> Why not use Allocator.dupe?

If my understanding is correct, using `Allocator.dupe` would only copy `items.len` of the original array. It might be the case that `items.len < capacity`, and we want to make sure we include the reserved capacity when cloning.

SpexGuy:
Thanks for this! I think your implementation has a bug though, you need to set the length of the cloned array to match the length of the original.  It currently uses the capacity of the original IIUC.

schmee:
>  [...] you need to set the length of the cloned array to match the length of the original

D'oh, thanks for catching this! 🙏  Fixed and added to the test. Also had some arguments backwards in `expectEqual`, fixed that as well.

lib/std/array_list.zig
```diff
@@ -802,6 +825,55 @@ test "std.ArrayList/ArrayListUnmanaged.initCapacity" {
...
+        try array.append(5);
+
+        const cloned = try array.clone();
+        defer cloned.deinit();
+
+        var i: usize = 0;
+        while (i < array.items.len) : (i += 1) {
+            try testing.expectEqual(array.items[i], cloned.items[i]);
+        }
+        try testing.expectEqual(array.items.len, cloned.items.len);
```
squeek502, line 843: Nitpick: this could be replaced by `try testing.expectEqualSlices(i32, array.items, cloned.items);`.

schmee: Fixed 👍

Vexu:
Why not just do:
```zig
var copy = ArrayList(T).init(self.allocator);
try copy.appendSlice(self.items);
return copy;
```

schmee:
@Vexu If I understand correctly they are not equivalent, you would have to call `copy.ensureTotalCapacity(self.capacity)` as well, which means you may end up doing two allocations instead of one.

Vexu:
Right, but I still think it would be cleaner to implement this using existing functions. Especially since the current function will cause an invalid free if the allocator decides to give you more than `self.capacity` items.
```zig
var copy = ArrayList(T).init(self.allocator);
try copy.ensureTotalCapacity(self capacity);
copy.appendSliceAssumeCapacity(self.items);
return copy;
```

schmee:
Sure, I'm happy to switch to your implementation if that avoids an invalid free.

> Especially since the current function will cause an invalid free if the allocator decides to give you more than self.capacity items.

I had no idea that could happen! For my understanding, that could only happen if `alignment != @alignOf(T)`, right?

squeek502:
> ```zig
> var copy = ArrayList(T).init(self.allocator);
> try copy.ensureTotalCapacity(self capacity);
> copy.appendSliceAssumeCapacity(self.items);
> return copy;
> ```

If we were to go this route, I think this implementation would make more sense in order to avoid the `better_capacity` calculation in `ensureTotalCapacity`:

```zig
var copy = try ArrayList(T).initCapacity(self.allocator, self.capacity);
copy.appendSliceAssumeCapacity(self.items);
return copy;
```

(could also just use `ensureTotalCapacityPrecise` but I think `initCapacity` is cleaner)

The tests would also need check for `clone.capacity >= array.capacity` instead of equality (but maybe that's already the case regardless of the implementation?).

Vexu:
> I had no idea that could happen! For my understanding, that could only happen if `alignment != @alignOf(T)`, right? (if I'm reading [this comment](https://github.com/ziglang/zig/blob/master/lib/std/mem/Allocator.zig#L19-L20) correctly)

Reading this again, I was probably thinking about `realloc`, I just remember having issues because I assumed I was getting as many bytes as I asked for.

schmee:
So looking at the ArrayList code, the only place memory is allocated is `ensureTotalCapacityPrecise`, which uses `allocator.reallocAtLeast`. As is clear from the name, that will give you the bytes you ask for _or more_ if you're using a non-standard alignment. In light of this, I've updated to @squeek502's implementation which takes both these things into account.

Thank you @Vexu and @squeek502 for your careful review!

lib/std/array_list.zig
```diff
@@ -108,6 +108,13 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
             return result[0 .. result.len - 1 :sentinel];
         }
 
+        /// Creates a copy of this ArrayList, using the same allocator.
+        pub fn clone(self: *Self) !Self {
+            var cloned = try ArrayList(T).initCapacity(self.allocator, self.capacity);
```
wsengir, line 113: This will erase alignment upon cloning. We can avoid that by using `Self`:
```suggestion
            var cloned = try Self.initCapacity(self.allocator, self.capacity);
```

lib/std/array_list.zig
```diff
@@ -489,6 +496,13 @@ pub fn ArrayListAlignedUnmanaged(comptime T: type, comptime alignment: ?u29) typ
             return result[0 .. result.len - 1 :sentinel];
         }
 
+        /// Creates a copy of this ArrayList.
+        pub fn clone(self: *Self, allocator: Allocator) !Self {
+            var cloned = try ArrayListUnmanaged(T).initCapacity(allocator, self.capacity);
```
wsengir, line 501: Same as here:
```suggestion
            var cloned = try Self.initCapacity(allocator, self.capacity);
```

wsengir (changes requested):

schmee:
Thanks @wsengir, updated! 👍

Issue #9485, opened 2021-07-29
glibc 2.27 or older: fcntl64 not found, but zig's glibc headers refer it


motiejus:
TLDR: zig is using "too new" glibc headers, which sometimes references undefined symbols. This fails compilation for at least sqlite and libuv when older glibc is selected, because it [redefines](https://sourceware.org/git/?p=glibc.git;a=blob;f=io/fcntl.h;h=6b0e9fa1fa2a2d8872b1adb5c0861410756a9e52;hb=3c03baca37fdcb52c3881e653ca392bba7a99c2b#l180) `fcntl` to `fcntl64`, which is present only in newer glibcs.

`main.c`
```
#define _FILE_OFFSET_BITS 64
#include <unistd.h>
#include <fcntl.h>
#include <stdio.h>

int main() {
    printf("address to fcntl: %p\n", fcntl);
}
```

This fails when target is glibc 2.27 or older:
```
$ zig cc --target=x86_64-linux-gnu.2.27  main.c
ld.lld: error: undefined symbol: fcntl64
>>> referenced by main.c
>>>               /home/motiejus/.cache/zig/o/921a4d8978936f8450e53a6103470e2c/main.o:(main)
>>> did you mean: fcntl64@GLIBC_2.28
>>> defined in: /home/motiejus/.cache/zig/o/6dd7f9446c261fd02c47b1aad02ab90b/libc.so.6
```

And works with glibc 2.28+:
```
$ zig cc --target=x86_64-linux-gnu.2.28  main.c
$ ./a.out 
address to fcntl: 0x7f46a328e330
```

This task gives a small reproducible test case; the problem was well explained in https://github.com/ziglang/zig/issues/5882#issuecomment-748974297 , includes a workaround (for x86_64 only though). While the workaround works, it may be nicer if zig provided headers of the requested version, and make this problem go away?

motiejus:
Also related: 39083c31a550ed80f369f60d35791e98904b8096

motiejus:
if https://patchwork.sourceware.org/project/glibc/patch/20220129023727.1496360-1-andrew@ziglang.org/ gets merged, I will attempt do to an equivalent thing for `fcntl64` for glibc <= 2.27.

etanol:
The solution provided in [5882][1] didn't fully work for me when building CPython.  Therefore, I tried something similar to 39083c31a550ed80f369f60d35791e98904b8096:

``` diff
[...]

PR #10330, opened 2021-12-13
stage2: improved glibc stubs

### LINKED ISSUE #5882: glibc undefined symbol: fcntl64 when using zig cc as a drop-in compiler
I'm trying to build a small test application with libuv embedded.  I want to use zig cc because the caching system makes it much faster.  The following command works for clang and gcc, but not zig cc.

```sh
zig cc -o main main.c libuv/src/fs-poll.c libuv/src/idna.c libuv/src/inet.c libuv/src/random.c libuv/src/strscpy.c libuv/src/threadpool.c libuv/src/timer.c libuv/src/uv-common.c libuv/src/uv-data-getter-setters.c libuv/src/version.c libuv/src/unix/async.c libuv/src/unix/core.c libuv/src/unix/dl.c libuv/src/unix/fs.c libuv/src/unix/getaddrinfo.c libuv/src/unix/getnameinfo.c libuv/src/unix/loop-watcher.c libuv/src/unix/loop.c libuv/src/unix/pipe.c libuv/src/unix/poll.c libuv/src/unix/process.c libuv/src/unix/random-devurandom.c libuv/src/unix/signal.c libuv/src/unix/stream.c libuv/src/unix/tcp.c libuv/src/unix/thread.c libuv/src/unix/tty.c libuv/src/unix/udp.c libuv/src/unix/proctitle.c libuv/src/unix/linux-core.c libuv/src/unix/linux-inotify.c libuv/src/unix/linux-syscalls.c libuv/src/unix/procfs-exepath.c libuv/src/unix/random-getrandom.c libuv/src/unix/random-sysctl-linux.c -pthread -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_GNU_SOURCE -D_POSIX_C_SOURCE=200112 -Os -ldl -lrt  -Ilibuv/include -Ilibuv/src
Compile C Objects [36/36] random-sysctl-linux.c...lld: error: undefined symbol: fcntl64
>>> referenced by core.c
>>>               zig-cache/o/a98yrC-16nzaiLvm0YSsj7Kr64jgWlKyw30Ca_aTuY8lBx8HcWfjUWLmNzWOQ59K/core.o:(uv__nonblock_fcntl)
>>> referenced by core.c
>>>               zig-cache/o/a98yrC-16nzaiLvm0YSsj7Kr64jgWlKyw30Ca_aTuY8lBx8HcWfjUWLmNzWOQ59K/core.o:(uv__nonblock_fcntl)
>>> referenced by core.c
>>>               zig-cache/o/a98yrC-16nzaiLvm0YSsj7Kr64jgWlKyw30Ca_aTuY8lBx8HcWfjUWLmNzWOQ59K/core.o:(uv__cloexec_fcntl)
>>> referenced by core.c
>>>               zig-cache/o/a98yrC-16nzaiLvm0YSsj7Kr64jgWlKyw30Ca_aTuY8lBx8HcWfjUWLmNzWOQ59K/core.o:(uv__cloexec_fcntl)
>>> referenced by pipe.c
>>>               zig-cache/o/Gp2nxP14BgUu9fMnuMMJ4gryoTCPAocIoqmgTAVdptPg5_FP1QjIdJRHfMpqm3ls/pipe.o:(uv_pipe_open)
>>> referenced by process.c
>>>               zig-cache/o/Z2n-GR8LD-1fXu4uIs37LmdvcK0K1mfCZ4iNwE1idOtrGfUhw6eXoJ8i5oMbsgSg/process.o:(uv__process_child_init)
>>> did you mean: fcntl64
>>> defined in: /home/tim/.cache/zig/stage1/h/pVSItZEIxIXx_ZQQwt9SLNNbPH8xxZp1plbirUK6SOmyZ4FTf2EOP08onozfGP03/libc.so.6.0.0
```

Is there something special I need to pass in for this to build?

### LINKED ISSUE #7667: relocation error: symbol pthread_sigmask version GLIBC_2.2.5 not defined in file libc.so.6 with link time reference

### LINKED ISSUE #8714: Bad logic in generating glibc stubs

### LINKED ISSUE #8896: glibc dynamic stubs are deeply flawed

andrewrk:
This commit upgrades glibc shared library stub-creating code to use the
new abilists file which is generated by the new glibc-abi-tool project:
https://github.com/ziglang/glibc-abi-tool/

The abilists file is different in these ways:
 * It additionally encodes whether a symbol is a function or an object,
   and if it is an object, it additionally encodes the size in bytes.
 * It additionally encodes migrations of symbols from one library to
   another between glibc versions.
 * It is binary data instead of ascii.
 * It is one file instead of three.
 * It is 165 KB instead of 200 KB.

This solves a handful of bugs.

Fixes #7667
Fixes #8714
Fixes #8896

cc @LemonBoy 

I have not yet verified that all these issues are fixed; I will do that before merging the PR.


motiejus:
I don't think it will fix #5882, as it's a header problem, not a glib-stub problem. Testing my example from #9485 with https://github.com/ziglang/zig/tree/glibc-stubs:

```
$ ~/code/zig/build/zig version
0.9.0-dev.1951+0d7331b6f
$ ~/code/zig/build/zig cc --target=x86_64-linux-gnu.2.27  main.c
ld.lld: error: undefined symbol: fcntl64
>>> referenced by main.c
>>>               /home/motiejus/.cache/zig/o/201c47b92a2605ce99b0da8a4c160de3/main.o:(main)
```

marler8997:
I confirmed what @motiejus said by verifying that issue #9485 still reproduces with this branch.  It appears that in order to really fix this issue, Zig would need a way to version the glibc headers as well (not just the libraries we link to).  Maybe a new project is in order to solve this?  I think it may be fairly ambitious to do this in an efficient way (i.e. not including an instance of the glibc headers for every version) but maybe we could start there and make incremental improvements?  Maybe compression would be a good start to see how feasible it would be?

PR #10331, opened 2021-12-13
Bcrypt pbkdf

daurnimator:
Tested with:
```zig
// Test against OpenBSD implementation https://github.com/openbsd/src/blob/6df1256b7792691e66c2ed9d86a8c103069f9e34/lib/libutil/bcrypt_pbkdf.c#L98

const std = @import("std");
const crypto = std.crypto;
const testing = std.testing;
const bcrypt = crypto.pwhash.bcrypt;
const Sha512 = crypto.hash.sha2.Sha512;

extern fn bcrypt_hash(
    sha2pass: *const [Sha512.digest_length]u8,
    sha2salt: *const [Sha512.digest_length]u8,
    out: *[32]u8,
) void;
test "bcrypt_hash" {
    var c_out: [32]u8 = undefined;
    bcrypt_hash("Hi" ** 32, "Salt" ** 16, &c_out);

    const zig_out = bcrypt.pbkdf_prf.hash("Hi".* ** 32, "Salt".* ** 16);

    try testing.expectEqualSlices(u8, &c_out, &zig_out);
}

extern fn bcrypt_pbkdf(
    pass: [*]const u8,
    passlen: usize,
    salt: [*]const u8,
    saltlen: usize,
    key: [*]u8,
    keylen: usize,
    rounds: c_uint,
) c_int;
test "bcrypt_pbkdf" {
    var c_out: [32]u8 = undefined;
    switch (bcrypt_pbkdf("password", 8, "salt", 4, &c_out, c_out.len, 15)) {
        0 => {},
        else => return error.TestFailed,
    }

    var zig_out: [32]u8 = undefined;
    try bcrypt.pbkdf("password", "salt", &zig_out, 15);

    try testing.expectEqualSlices(u8, &c_out, &zig_out);
}
```


lib/std/crypto/bcrypt.zig
```diff
@@ -178,6 +179,88 @@ pub fn bcrypt(
...
+        crypto.utils.secureZero(State, @as(*[1]State, &state));
+
+        return out;
+    }
+};
+
+/// bcrypt PBKDF2 implementation with variations to match OpenBSD
+/// https://github.com/openbsd/src/blob/6df1256b7792691e66c2ed9d86a8c103069f9e34/lib/libutil/bcrypt_pbkdf.c#L98
+///
+/// This particular variant is used in e.g. SSH
```
jedisct1, line 259: s/SSH/OpenSSH/

lib/std/crypto/bcrypt.zig
```diff
@@ -178,6 +179,88 @@ pub fn bcrypt(
...
 }
 
+const pbkdf_prf = struct {
+    const Self = @This();
+    pub const mac_length = 32;
+
+    hasher: Sha512,
+    sha2pass: [Sha512.digest_length]u8,
+
+    pub fn create(out: *[mac_length]u8, msg: []const u8, key: []const u8) void {
```
jedisct1, line 189: Would you mind adding a short `///` comment to the new public functions and types?

lib/std/crypto/bcrypt.zig
```diff
@@ -178,6 +179,88 @@ pub fn bcrypt(
     return ct[0..dk_length].*;
 }
 
+const pbkdf_prf = struct {
```
Mouvedia, line 182: > [ One modification from official pbkdf2. Instead of outputting key material linearly, we mix it.
pbkdf2 has a known weakness where if one uses it to generate (i.e.) 512 bits of key material for use as two 256 bit keys, an attacker can merely run once through the outer loop below, but the user always runs it twice. Shuffling output bytes requires computing the entirety of the key material to assemble any subkey. This is something a wise caller could do; we just do it for you.](https://github.com/grnet/python-bcrypt/blob/master/bcrypt/bcrypt_pbkdf.c#L146)

Something I stumbled on.

lib/std/crypto/bcrypt.zig
daurnimator: FYI this is not actually public public (the struct it is part of is private)

lib/std/crypto/bcrypt.zig
daurnimator: It's used by more implementations than just OpenSSH, e.g. see go's implementation at https://cs.opensource.google/go/x/crypto/+/e495a2d5:ssh/internal/bcrypt_pbkdf/bcrypt_pbkdf.go

PR #10340, opened 2021-12-15
std.bounded_array: support inserting a new value at the end

Arnavion:
Since `BoundedArray.insert` internally reserves space for the element
to be inserted, it can support inserting at the position that is
the current length of the array. Change the check for the insertion position
to allow this.


jedisct1:
`ArrayList` supports this, so `BoundedArray` should as well.

Thanks!

PR #10342, opened 2021-12-16
std.priority_{de,}queue: allow comparator to take context parameter

Arnavion:
Justification, from #zig on 2021-12-15 :

>&lt;Arnavion&gt; There's no way I can use a std.PriorityQueue with a compareFn that requires a sort context, is there? Well, other than stuffing the context inside every T
>&lt;andrewrk&gt; Arnavion, PriorityQueue should be changed to accept a compare function that has a context parameter

### DIFF
--- a/lib/std/priority_dequeue.zig
+++ b/lib/std/priority_dequeue.zig
@@ -8,27 +8,29 @@ const expectEqual = testing.expectEqual;
 const expectError = testing.expectError;
 
 /// Priority Dequeue for storing generic data. Initialize with `init`.
-/// Provide `compareFn` that returns `Order.lt` when its first
-/// argument should get min-popped before its second argument,
+/// Provide `compareFn` that returns `Order.lt` when its second
+/// argument should get min-popped before its third argument,
 /// `Order.eq` if the arguments are of equal priority, or `Order.gt`
-/// if the second argument should be min-popped first.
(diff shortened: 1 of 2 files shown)


lib/std/priority_dequeue.zig
```diff
@@ -8,27 +8,29 @@ const expectEqual = testing.expectEqual;
...
+/// argument should get min-popped before its third argument,
 /// `Order.eq` if the arguments are of equal priority, or `Order.gt`
-/// if the second argument should be min-popped first.
+/// if the third argument should be min-popped second.
 /// Popping the max element works in reverse. For example,
 /// to make `popMin` return the smallest number, provide
-/// `fn lessThan(a: T, b: T) Order { return std.math.order(a, b); }`
-pub fn PriorityDequeue(comptime T: type, comptime compareFn: fn (T, T) Order) type {
+/// `fn lessThan(context: void, a: T, b: T) Order { _ = context; return std.math.order(a, b); }`
+pub fn PriorityDequeue(comptime T: type, comptime Context: type, comptime compareFn: fn (context: Context, a: T, b: T) Order) type {
```
Arnavion, line 18: Should `Context` be named something that specifically associates it with the comparator? `CompareFnContext` ? `SortContext` ?

andrewrk:
Thanks for doing this work!

PR #10344, opened 2021-12-16
std.time.epoch: fix issue in documentation

samhattangady:
std.time.epoch.DaySeconds.getHoursIntoDay said that hours goes from
0-11. should be 0-23.


andrewrk:
Thanks!

PR #10346, opened 2021-12-16
macho: handle -install_name option for dylibs/MachO

kubkon:
Given that this change is small when it comes to the amount of the required code edits, I've decided to go straight for a PR rather than submitting an issue first. I reckon with the code changes in this PR, it will be easier to discuss this proposal. This proposal changes the default behaviour for the ID of the created dynamic library targeting darwin in the following way:

The status quo for the `build.zig` build system is preserved in the sense that, if the user does not explicitly override
`dylib.setInstallName(...);` in their build script, the default of `@rpath/libname.dylib` applies. However, should they want to override the default behaviour, they can either:

1) unset it with

```dylib.setIntallName(null);```

2) set it to an explicit string with

```dylib.setInstallName("somename.dylib");```

When it comes to the command line however, the default is not to use `@rpath` for the install name when creating a dylib. The user will now be required to explicitly specify the `@rpath` as part of the desired install name should they choose so like so:

1) with `build-lib`

```
zig build-lib -dynamic foo.zig -install_name @rpath/libfoo.dylib
```

2) with `cc`

```
zig cc -shared foo.c -o libfoo.dylib -Wl,"-install_name=@rpath/libfoo.dylib"
```

I think this is a reasonable new default which effectively gives the user full control over the dylib's install name when building the library via the command line primitives such as `zig build-lib -dynamic` and `zig cc -shared`, and at the same time, Zig's build system still makes the inclusion of the `@rpath/` prefix in the generated dylib's install name the default.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -1537,6 +1537,9 @@ pub const LibExeObjStep = struct {
     /// Permit read-only relocations in read-only segments. Disallowed by default.
     link_z_notext: bool = false,
 
+    /// (Darwin) Install name for the dylib
+    install_name: ?[]const u8 = null,
+
     /// Position Independent Code
     force_pic: ?bool = null,
 
@@ -2450,6 +2453,16 @@ pub const LibExeObjStep = struct {
             if (self.version) |version| {
                 zig_args.append("--version") catch unreachable;
                 zig_args.append(builder.fmt("{}", .{version})) catch unreachable;
+            }
+
+            if (self.target.isDarwin()) {
+                const install_name = self.install_name orelse builder.fmt("@rpath/{s}{s}{s}", .{
+                    self.target.libPrefix(),
+                    self.name,
+                    self.target.dynamicLibSuffix(),
+                });
+                try zig_args.append("-install_name");
+                try zig_args.append(install_name);
             }
         }
 
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -780,6 +780,8 @@ pub const InitOptions = struct {
     enable_link_snapshots: bool = false,
     /// (Darwin) Path and version of the native SDK if detected.
     native_darwin_sdk: ?std.zig.system.darwin.DarwinSDK = null,
+    /// (Darwin) Install name of the dylib
+    install_name: ?[]const u8 = null,
 };
 
 fn addPackageTableToCacheHash(
@@ -1509,6 +1511,7 @@ pub fn create(gpa: Allocator, options: InitOptions) !*Compilation {
             .use_stage1 = use_stage1,
             .enable_link_snapshots = options.enable_link_snapshots,
             .native_darwin_sdk = options.native_darwin_sdk,
+            .install_name = options.install_name,
         });
         errdefer bin_file.destroy();
         comp.* = .{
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: lib/std/build.zig [hunk 1]
```diff
@@ -2451,6 +2464,15 @@ pub const LibExeObjStep = struct {
...
             }
+
+            if (!self.install_name.overriden or self.install_name.name != null) {
+                const target = self.target.toTarget();
+                const install_name = self.install_name.name orelse builder.fmt("@rpath/{s}{s}{s}", .{
+                    target.libPrefix(), self.name, target.dynamicLibSuffix(),
+                });
+                try zig_args.append("-install_name");
+                try zig_args.append(install_name);
+            }
```
andrewrk, line 2475: I don't think that the build system should accept a differently formatted install_name than the CLI. Also calling toTarget() here is problematic.

```suggestion
            if (self.install_name.name) |install_name| {
                try zig_args.append("-install_name");
                try zig_args.append(install_name);
            }
```

However we still need to inject this rpath/install name thing when an executable dynamically links a library that resides in the cache system, so that running the exe directly from the cache system works. I think the build system should set different state outside of `install_name` in order to satisfy this requirement.

andrewrk (changes requested):
I'm happy with the CLI and compiler changes here but I have some opinions about the build system.

lib/std/build.zig
```diff
@@ -1537,6 +1537,12 @@ pub const LibExeObjStep = struct {
     /// Permit read-only relocations in read-only segments. Disallowed by default.
     link_z_notext: bool = false,
 
+    /// (Darwin) Install name for the dylib
+    install_name: struct {
+        name: ?[]const u8 = null,
+        overriden: bool = false,
+    } = .{},
```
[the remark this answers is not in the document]
kubkon: So that's what I had initially, but do we ever permit the user to unset the install name? Would that ever be a possibility?

lib/std/build.zig
kubkon: Hmm, ok, so where do you suggest we should populate the default install name for the build system?

lib/std/build.zig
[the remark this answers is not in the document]
andrewrk: Yes, `null` means unset.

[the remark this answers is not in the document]
kubkon: After some offline discussion, we agreed that if the user wants to tweak the `install_name` they should access the field directly. Additionally, `null` means default in this case which means that the builder will populate the value with `@rpath/libname.dylib` by default (where `libname.dylib` equals the actual name of the dynamic library).

lib/std/build.zig
kubkon: After some offline discussion with Andrew, the logic for populating the `install_name` stays in this place, however, there is no need to invoke `CrossTarget.toTarget()` function; `CrossTarget` features both `libPrefix()` and `dynamicLibSuffix()` helper functions, so use that instead of prematurely acquiring the actual `std.Target` object.

lib/std/build.zig
```diff
@@ -2451,6 +2454,14 @@ pub const LibExeObjStep = struct {
                 zig_args.append("--version") catch unreachable;
                 zig_args.append(builder.fmt("{}", .{version})) catch unreachable;
             }
+
+            const install_name = self.install_name orelse builder.fmt("@rpath/{s}{s}{s}", .{
```
andrewrk, line 2458: This should only be emitted when the target is darwin, yeah?

kubkon: Good catch, thanks!

PR #10347, opened 2021-12-17
Actually installs the implib for DLLs when installing the artifact

MasterQ32:


andrewrk (approved):

kubkon:
@MasterQ32 so what's the implication of this change for Windows? Would you mind providing an illustrative example so that I can learn about this about more please?

MasterQ32:
> @MasterQ32 so what's the implication of this change for Windows? Would you mind providing an illustrative example so that I can learn about this about more please?

Windows requires to link a "implib" if you want to use a DLL instead of the actual DLL. The lib is just telling the linker "there's a DLL named lib and it has these symbols"

You can generate this lib from existing DLLs, but it's typical to ship them with a SDK so people can just link against that lib and use your DLL. 

Right now zig generated the right thing, but the build system didn't install it to the right place. This PR fixes this and allows people to easily ship precompiled DLLs

kubkon:
NIce, thanks for the explanation!

kubkon (approved):

PR #10351, opened 2021-12-17
plan9 linker: fix off by 1 error

g-w1:
Before:
sbad
After:
sgood

### DIFF
--- a/src/link/Plan9.zig
+++ b/src/link/Plan9.zig
@@ -406,14 +406,14 @@ pub fn flushModule(self: *Plan9, comp: *Compilation) !void {
     defer linecountinfo.deinit();
     // text
     {
-        var linecount: u32 = 0;
+        var linecount: i64 = -1;
         var it_file = self.fn_decl_table.iterator();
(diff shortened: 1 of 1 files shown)


src/link/Plan9.zig
```diff
@@ -406,14 +406,14 @@ pub fn flushModule(self: *Plan9, comp: *Compilation) !void {
     defer linecountinfo.deinit();
     // text
     {
-        var linecount: u32 = 0;
```
kubkon, line 409: Does that mean `linecount` can actually be negative? Is that even possible?

g-w1: It doesn't mean linecount is negative, but I think the interpreter starts with it as -1.

andrewrk: It doesn't seem like it should be signed, since a line count can't be negative. Also it's int casted down to an i32 below, so not sure why it is 64 bits.

Perhaps a `-1` is warranted somewhere else?

But you're the owner of this code, so if you want it merged I'll merge it :)

PR #10352, opened 2021-12-18
Allocator: `allocBytes` and `reallocBytes`

### LINKED ISSUE #10348: Allocgate missing runtime-known alignment functions
To quote @andrewrk on #10112:

> [...] I agree that this use case will need to continue to be supported - runtime-known alignment.

However, the new allocator interface does not feature any high-level functions to support this. Ideally there would be a `reallocAdvancedWithRetAddrRuntimeAlignment` (though that name is rather unwieldy).

leecannon:
Reimplement `reallocBytes` after regression from #9394/#10112
Also adds an analogous `allocBytes` function.

@silversquirl / @mlugg: do these functions satisfy your requirements?
As the these functions don't return a slice with the alignment as part of the type I choose not to name them something like `reallocAdvancedWithRetAddrRuntimeAlignment` to seperate them from the other `alloc`/`realloc` functions.

closes #10348

### DIFF
--- a/lib/std/mem/Allocator.zig
+++ b/lib/std/mem/Allocator.zig
@@ -507,15 +507,126 @@ pub fn dupeZ(allocator: Allocator, comptime T: type, m: []const T) ![:0]T {
     return new_buf[0..m.len :0];
 }
 
+/// This function allows a runtime `alignment` value. Callers should generally prefer
+/// to call the `alloc*` functions.
+pub fn allocBytes(
+    self: Allocator,
+    /// Must be >= 1.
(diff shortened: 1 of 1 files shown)


mlugg:
Quickly skimming through, it looks like that should be fine for my use case yeah. Thanks!

andrewrk (changes requested):
Thanks for the patch!

Needs simple test coverage and then it's good to go.

leecannon:
Tests added

andrewrk (approved):
Thank you :+1:

PR #10370, opened 2021-12-20
stage2: Make page_allocator work

Snektron:
This PR includes a bunch of things that are required for `page_allocator` to compile under stage 2. This includes the allocator itself as well as the allocation function:
```zig
const std = @import("std");

test "page allocator" {
    const msg = "hello from heap memory\n";
    const allocator = std.heap.page_allocator;
    const mem = try allocator.alloc(u8, msg.len);
    for (msg) |c, i| mem[i] = c;
    asm volatile ("syscall"
        :
        : [number] "{rax}" (1),
          [arg1] "{rdi}" (1),
          [arg2] "{rsi}" (@ptrToInt(mem.ptr)),
          [arg3] "{rdx}" (msg.len),
        : "rcx", "r11", "memory"
    );
}
```
(freeing does not work yet)

Error set related changes:
* Replaced the plain arrays used for the error values of `Module.ErrorSet` and Type `merged_error_set` with hash maps. This is to make error set subset checks easier, though it might become a problem for error set hashing.
* Moved error set state into `Module.Fn`. This was previously causing some problems after an error set was copied, losing a proper reference to the original error set as well as creating memory errors. Note that any individual function might have multiple inferred error sets associated with it; its own for runtime invocations of the function and those of inline/comptime functions which return inferred error sets.
* Error set coercion.
* Allow merged error sets to cast to `E!T`.

Other stage 2 additions:
* Allow `@ptrToInt` on anything that is a pointer at runtime, as `@ptrToInt(pointer_like_optional)` was used somewhere.
* Implemented built-in functions `@returnAddress()`, `@addWithOverflow()`, `@mulWithOverflow()`, `@subWithOverflow()` and `@shlWithOverflow` (sema and LLVM codegen). Each also has a corresponding new AIR instruction.
* Disabled a standard library feature for stage 2 builds where a message would be printed when an unexpected errno was returned from a syscall. This relieves a dependency on `std.fmt` and stack trace related code for `page_allocator` on stage 2 builds.
* Made `anyopaque` a sized type for the time being, as this was causing `*anyopaque` to be recognized as zero-sized type. I guess this isn't really the proper fix for this, though.

Some bugs fixed:
* ZIR function prototypes are now wrapped in blocks with their `param` instructions. The parameters were prior to this emitted in a block higher, which caused function pointers in the same block to pick up each others parameters.
* A similar problem with the `params` list of Block. My current fix just includes saving and restoring the list on any `block_inline` and `block` ZIR instruction, though it feels kind of hacky and a better solution is probably required for this in time.
* `addrspace` pointer information was not yet written into pointer `@typeInfo`, which caused some problems.

Tests pass locally:
```
$ ./zig build -p zig-out -Denable-llvm -Dskip-compile-errors test-stage2 -fqemu -fwasmtime -fwine
All 90 tests passed.
$ ./zig-out/bin/zig test ../test/behavior.zig -I ../test -fLLVM
$ ./zig-out/bin/zig test ../test/behavior.zig -I ../test -ofmt=c
```

### DIFF
--- a/lib/std/math/big/int.zig
+++ b/lib/std/math/big/int.zig
@@ -443,12 +443,12 @@ pub const Mutable = struct {
         }
     }
 
-    /// r = a + b with 2s-complement wrapping semantics.
+    /// r = a + b with 2s-complement wrapping semantics. Returns whether overflow occurred.
     /// r, a and b may be aliases
     ///
     /// Asserts the result fits in `r`. An upper bound on the number of limbs needed by
     /// r is `calcTwosCompLimbCount(bit_count)`.
-    pub fn addWrap(r: *Mutable, a: Const, b: Const, signedness: Signedness, bit_count: usize) void {
+    pub fn addWrap(r: *Mutable, a: Const, b: Const, signedness: Signedness, bit_count: usize) bool {
         const req_limbs = calcTwosCompLimbCount(bit_count);
 
         // Slice of the upper bits if they exist, these will be ignored and allows us to use addCarry to determine
@@ -463,6 +463,7 @@ pub const Mutable = struct {
             .limbs = b.limbs[0..math.min(req_limbs, b.limbs.len)],
         };
 
+        var carry_truncated = false;
         if (r.addCarry(x, y)) {
             // There are two possibilities here:
             // - We overflowed req_limbs. In this case, the carry is ignored, as it would be removed by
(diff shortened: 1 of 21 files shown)


Snektron:
Oh and i forgot to mention: i have made `@mulWithOverflow`, `@addWithOverflow` return `undefined` when both arguments are comptime-undefined rather than generate a compile error. It seemed to me that there is no case in which this could invoke illegal behavior, so this seemed like the proper implementation to me.

andrewrk (approved):
Great work. We can address the TODO questions and comments in follow-up commits.

PR #10371, opened 2021-12-20
Change `ArgIterator.next()` return type

ominitay:
I'm making this PR after receiving positive feedback regarding this from people on the Discord, including @SpexGuy and @N00byEdge. 

This changes the return type of `ArgIterator.next()` from `?(NextError![:0]u8)` to `NextError!?[:0]u8`. This change makes the `ArgIterator` much cleaner to use in `while` loops or `if` statements. For example:
```zig
while (it.next(allocator)) |arg_or_err| {
    const arg = try arg_or_err;
```
Is now:
```zig
while (try it.next(allocator)) |arg| {
```

This is much better, as it storing an error union in a variable is usually a bad thing to do, and discourages handling the error immediately. Additionally, the convention in the std library is to use this order (`!?` as opposed to `?!`) -- I couldn't find any examples of the opposite order besides this one with grep.

### DIFF
--- a/doc/docgen.zig
+++ b/doc/docgen.zig
@@ -27,18 +27,18 @@ pub fn main() !void {
 
     if (!args_it.skip()) @panic("expected self arg");
 
-    const zig_exe = try (args_it.next(allocator) orelse @panic("expected zig exe arg"));
+    const zig_exe = (try args_it.next(allocator)) orelse @panic("expected zig exe arg");
     defer allocator.free(zig_exe);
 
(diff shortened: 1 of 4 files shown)


N00byEdge:
Nice. I explicitly broke my non-linux builds by grabbing the underlying non-erroring iterator just because of this interface. Would be nice to theoretically make my code work on non-linux too.

ominitay:
@N00byEdge I think I have done the same in the past :joy:

PR #10376, opened 2021-12-21
Add BSD Authentication constants

dantecatalfamo:
Add the constants used to interact with BSD Authentication

https://blog.lambda.cx/posts/how-bsd-authentication-works/

### DIFF
--- a/lib/std/c/openbsd.zig
+++ b/lib/std/c/openbsd.zig
@@ -708,7 +708,7 @@ pub const auth_item_t = c_int;
 
 pub const AUTHV = struct {
     pub const ALL: auth_item_t = 0;
-    pub const CHALLENGEL: auth_item_t = 1;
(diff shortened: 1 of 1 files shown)


lib/std/c/openbsd.zig
```diff
@@ -703,8 +703,48 @@ pub const T = struct {
     pub const IOCXMTFRAME = 0x80087444;
 };
 
+// BSD Authentication
+pub const auth_item_t = c_int;
+
+pub const AUTHV = struct {
+    pub const ALL: auth_item_t = 0;
+    pub const CHALLENGEL: auth_item_t = 1;
```
semarie, line 711: typo: CHALLENGE

dantecatalfamo: Nice catch! Fixed :slightly_smiling_face:

Issue #6029, opened 2020-08-11
put constant data in .rodata section instead of .text


andrewrk:

andrewrk:
cc @kubkon - I noticed you mentioned this in your talk :-)

watchpoints:
how

Ersikan:
Hello ! 
I finally have some time back to contribute to Zig, and I'm excited to start working back on this issue.
Is there any new context on this issue I need to know before starting back working on it ? Who can I talk to if I have any questions on the ELF infrastructure ?

Is it possible to reopen my PR #8296 and put it in draft status, or should I create a new one ?

PR #10379, opened 2021-12-21
stage2 ARM: Refactor airStore

joachimschmidt557:
This _almost_ implements storing to global variables, but more support in the linking backends is needed as global variables are currently assigned to the read-only `.text` section (e.g. for ELF).

### DIFF
--- a/src/arch/arm/CodeGen.zig
+++ b/src/arch/arm/CodeGen.zig
@@ -1282,6 +1282,15 @@ fn load(self: *Self, dst_mcv: MCValue, ptr: MCValue, ptr_ty: Type) InnerError!vo
                         } },
                     });
                 },
+                .stack_offset => |off| {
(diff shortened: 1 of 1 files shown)


andrewrk:
Related:  #6029

andrewrk (approved):

andrewrk:
Looks like global variables might work now that e15a267668bae168f2bba06cc3706c54bc062522 landed?

joachimschmidt557:
Afaik, _reading_ from global variables was already working as they were assigned to the read-only `.text` section. The problem is _writing_ to global variables. I discussed this with @kubkon earlier, for read+write global variables, it is necessary to put them into the `.bss` or `.data` section, which requires further support from the linking backend.

But putting read-only global variables into `.rodata` instead of `.text` is already a huge step into the right direction imo :tada:

PR #10383, opened 2021-12-21
Only check the file's length once in `pdb.Msf.init`

gereeter:
I was debugging an unrelated issue with `procmon` and noticed that zig was doing this lookup to the filesystem a few thousand times. I doubt this has much performance impact in the grand scale of things, but removing that many calls into the system can't hurt, and it at least cleans up the `procmon` output.


andrewrk (approved):
Thanks!

PR #10388, opened 2021-12-22
std housekeeping

ominitay:
Spotted a few TODOs, thought it best to act on them now 0.9.0 is released.

### DIFF
--- a/lib/std/zig/parser_test.zig
+++ b/lib/std/zig/parser_test.zig
@@ -72,21 +72,6 @@ test "zig fmt: rewrite callconv(.Inline) to the inline keyword" {
     );
 }
 
-// TODO Remove this after zig 0.9.0 is released.
-test "zig fmt: rewrite @byteOffsetOf to @offsetOf" {
-    try testTransform(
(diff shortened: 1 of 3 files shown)


ominitay:
Seems that uncovered some broken code previously undetected thanks to the compileErrors

ominitay:
Rebased. This should be merged or rejected before it becomes stale.

lib/std/zig/parser_test.zig
```diff
@@ -72,21 +72,6 @@ test "zig fmt: rewrite callconv(.Inline) to the inline keyword" {
     );
 }
 
-// TODO Remove this after zig 0.9.0 is released.
-test "zig fmt: rewrite @byteOffsetOf to @offsetOf" {
```
Vexu, line 76: The actual rewriting this is testing should also be removed.

ominitay: Ah I see, will sort that out in a sec.

ominitay: Should be ok now :)

PR #10393, opened 2021-12-22
stage1: fix saturating arithmetic on type comptime_int

### LINKED ISSUE #10365: Unexpected result on saturated arithmetic on two comptime_int's
### Zig Version

0.9.0-dev.2006+ef0566df7

### Steps to Reproduce

`try std.testing.expectEqual(@as(u32, 2), 1 +| 1);`

It doesn't matter if the comptime_int's are literals or constants.

### Expected Behavior

Expected either the test to pass, or a compilation error as saturated addition of two comptime_int's is indicative of a bug. 

### Actual Behavior

Test fails, as arithmetic on two comptime_int's is always zero.

riverbl:
Makes wrapping or saturating arithmetic operations map to their non wrapping or saturating equivalents when the operands have type comptime_int (previously only wrapping operations were mapped).

Makes saturating left shift map to exact left shift when the left operand has type comptime_int (previously saturating left shift was banned in this case).

Fixes #10365.


andrewrk (changes requested):
Test coverage please!

riverbl:
Fixed some issues with `BigInt` in stage1. Fixes the following bugs:

```zig
const std = @import("std");

pub fn main() void {
    std.debug.print("{}\n", .{ 593423721213448152027139550640105366508 - 7 == 593423721213448152027139550640105366501 });
}
```
Should print `true`, prior to fix printed `false` (printing large numbers of type `comptime_int` is still broken).

```zig
pub fn main() void {
    const foo: i8 = -128;
    const bar: i8 = -128;

    _ = foo +% bar;
}
```
Should compile, prior to fix compilation failed with "error: operation caused overflow".

Fixed stage2 bug where performing wrapping or saturating operations on type `comptime_int` resulted in the compiler reaching `unreachable`. Wrapping or saturating operations now map to their non wrapping or saturating equivalents.

Added some tests!

I think function `analyzeArithmetic` in Sema.zig should be refactored to move the code in each switch case to separate functions. If this is desired I'll add it to this pull request or submit another pull request.

Vexu (approved):

PR #10394, opened 2021-12-23
stage2: rewrite MIR -> Isel layer for x86_64

kubkon:
This is a pretty hefty PR which aims at cleaning up outright wrong lowerings that I've created when adding the MIR layer for `x86_64` a couple of weeks ago. Hopefully, with this PR landing, it will be easier and much clearer when anyone will want to add new codegen bits to the compiler.

The main changes worthy of mention include:
* every MIR callsite in the codegen code now follows Intel syntax `dst <- src`
* every MIR -> Isel selection function in `Emit.zig` is implemented via a lowering function specific to the desired instruction encoding - this is also where we will only ever use the `Encoder` struct, in the lowering functions (functions prefixed with `lower`) ~~(**still working on it**)~~
* every lowering function is now tested for correctness when emitting a single instruction in the desired encoding ~~(**still working on it**)~~
* a mismatch in sizes between operands is a hard-error in the lowering functions - the errors are currently compressed into a single `error.EmitFail` but this will be a perfect opportunity for some future refactoring effort

EDIT: 
OK, first, mergable draft finished! There's a lot of potential for further refactoring between encodings, more test coverage to add, figure out more efficient lowerings in certain cases like `imul`, and add handling of the scale-index-base encoding modifications (this is currently not handled yet).

### DIFF
--- a/src/arch/x86_64/Mir.zig
+++ b/src/arch/x86_64/Mir.zig
@@ -380,14 +380,14 @@ pub const Ops = struct {
     }
 };
 
-pub fn extraData(mir_extra: []const u32, comptime T: type, index: usize) struct { data: T, end: usize } {
+pub fn extraData(mir: Mir, comptime T: type, index: usize) struct { data: T, end: usize } {
     const fields = std.meta.fields(T);
     var i: usize = index;
     var result: T = undefined;
     inline for (fields) |field| {
         @field(result, field.name) = switch (field.field_type) {
-            u32 => mir_extra[i],
-            i32 => @bitCast(i32, mir_extra[i]),
+            u32 => mir.extra[i],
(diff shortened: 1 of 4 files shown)


andrewrk (approved):

src/arch/x86_64/Mir.zig
```diff
@@ -292,6 +293,7 @@ pub const Inst = struct {
         /// Another instruction.
         inst: Index,
         /// A 32-bit immediate value.
+        /// TODO we should add support for 16- and 8-bit immediate values.
```
andrewrk, line 296: are you sure? it wouldn't save any bytes in the MIR, and we can just check the range of the value when lowering the MIR to machine code.

```suggestion
```

PR #10395, opened 2021-12-23
std.bit_set: add setRangeValue(Range, bool)

Arnavion:
For large ranges, this is faster than having the caller call setValue() for
each index in the range. Masks wholly covered by the range can be set to
the new mask value in one go, and the two masks at either end that are
partially covered can each set the covered range of bits in one go.

### DIFF
--- a/lib/std/bit_set.zig
+++ b/lib/std/bit_set.zig
@@ -110,6 +110,31 @@ pub fn IntegerBitSet(comptime size: u16) type {
             self.mask |= maskBit(index);
         }
 
+        /// Changes the value of all bits in the specified range to
+        /// match the passed boolean.
+        pub fn setRangeValue(self: *Self, range: Range, value: bool) void {
+            assert(range.end <= bit_length);
+            assert(range.start <= range.end);
+            if (range.start == range.end) return;
+            if (MaskInt == u0) return;
+
+            const start_bit = @intCast(ShiftInt, range.start);
+
+            var mask = std.math.boolMask(MaskInt, true) << start_bit;
+            if (range.end != bit_length) {
+                const end_bit = @intCast(ShiftInt, range.end);
+                mask &= std.math.boolMask(MaskInt, true) >> @truncate(ShiftInt, @as(usize, @bitSizeOf(MaskInt)) - @as(usize, end_bit));
+            }
+            self.mask &= ~mask;
+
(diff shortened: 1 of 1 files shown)


lib/std/bit_set.zig
```diff
@@ -345,6 +370,51 @@ pub fn ArrayBitSet(comptime MaskIntType: type, comptime size: usize) type {
...
+
+            const start_mask_index = maskIndex(range.start);
+            const start_bit = @truncate(ShiftInt, range.start);
+
+            const end_mask_index = maskIndex(range.end);
+            const end_bit = @truncate(ShiftInt, range.end);
+
+            if (start_mask_index == end_mask_index) {
+                var mask1 = std.math.boolMask(MaskInt, true) << start_bit;
+                var mask2 = std.math.boolMask(MaskInt, true) >> (mask_len - 1) - (end_bit - 1);
```
Arnavion, line 389: This `(a - 1) - (b - 1) == a - b` hack would be nice to avoid, but I couldn't find a way. The problem is that `end_bit` has to be `>= 1` at this point, so it's guaranteed that `mask_len - end_bit` will fit in a `ShiftInt`, but the compiler doesn't know that and complains that `mask_len` cannot fit in a `ShiftInt`.

lib/std/bit_set.zig
```diff
@@ -110,6 +110,31 @@ pub fn IntegerBitSet(comptime size: u16) type {
...
+            assert(range.start <= range.end);
+            if (range.start == range.end) return;
+            if (MaskInt == u0) return;
+
+            const start_bit = @intCast(ShiftInt, range.start);
+
+            var mask = std.math.boolMask(MaskInt, true) << start_bit;
+            if (range.end != bit_length) {
+                const end_bit = @intCast(ShiftInt, range.end);
+                mask &= std.math.boolMask(MaskInt, true) >> @truncate(ShiftInt, @as(usize, @bitSizeOf(MaskInt)) - @as(usize, end_bit));
```
Arnavion, line 126: This `@truncate(@as(usize) - @as(usize))` hack would be nice to avoid, but I couldn't find a way. If I use the `(@bitSizeOf(MaskInt) - 1) - (end_bit - 1)` method as in the other sets, the compiler complains that `1` isn't a valid `u0`, because it doesn't understand the early-return when `MaskInt` is a `u0`.

andrewrk (approved):

PR #10399, opened 2021-12-23
Make mem.Allocator functions generic

locriacyber:
Now the interface functions can be described shorter, like `std.mem.Allocator.AllocFnType(Foo)`.


Vexu:
Closing as stale.

PR #10401, opened 2021-12-23
stage2: make tests/behaviour/void.zig work with c backend

mattyhall:
* fix initialisation of void* fields of structs (initialises to NULL rather than {})
* don't generate fields where the type does not have codegen bits
* in airAlloc generate a void* local if the element type does not have codegen bits

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -302,7 +302,11 @@ pub const DeclGen = struct {
                         else => return dg.fail("TODO float types > 64 bits are not support in renderValue() as of now", .{}),
                     }
                 },
-
+                .Pointer => switch (dg.module.getTarget().cpu.arch.ptrBitWidth()) {
+                    32 => return writer.writeAll("(void *)0xaaaaaaaa"),
+                    64 => return writer.writeAll("(void *)0xaaaaaaaaaaaaaaaa"),
+                    else => unreachable,
+                },
                 else => {
                     // This should lower to 0xaa bytes in safe modes, and for unsafe modes should
                     // lower to leaving variables uninitialized (that might need to be implemented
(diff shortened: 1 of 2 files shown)


andrewrk:
the CI failure will be fixed by running `zig fmt` on your changes :)

mattyhall:
Whoops, thanks! I'm not sure why the new failure has happened unfortunately :(

src/codegen/c.zig
```diff
@@ -302,7 +302,7 @@ pub const DeclGen = struct {
                         else => return dg.fail("TODO float types > 64 bits are not support in renderValue() as of now", .{}),
                     }
                 },
-
+                .Pointer => return writer.writeAll("NULL"),
```
andrewrk, line 305: This is supposed to produce a pointer whose address is 0xaa bytes, not 0 bytes.

src/codegen/c.zig
```diff
@@ -1399,9 +1400,28 @@ fn airAlloc(f: *Function, inst: Air.Inst.Index) !CValue {
...
+        try f.object.dg.writeCValue(writer, local_value);
+        try writer.writeAll(" = (void *)");
+        switch (target.cpu.arch.ptrBitWidth()) {
+            32 => try writer.writeAll("0xaaaaaaaa"),
+            64 => try writer.writeAll("0xaaaaaaaaaaaaaaaa"),
+            else => unreachable,
+        }
+        try writer.writeAll(";\n");
+        return CValue{ .local_ref = local_value.local };
+    }
```
andrewrk, line 1422: There should be no local variable allocated in this case. See the `airAlloc` function in the LLVM backend for inspiration.

andrewrk (changes requested):

andrewrk:
Nice work!

PR #10403, opened 2021-12-23
compiler_rt: add __absvsi2, __absvdi2, __absvti2

matu3ba:
- abs can only overflow, if a == MIN
- comparing the sign change from wrapping addition is branchless
- tests: MIN, MIN+1,..MIN+4, -42, -7, -1, 0, 1, 7..

See #1290


matu3ba:
There is no panic handler for testing panics yet, so I left a TODO.

I tested manually the overflow case and complete range of i32.
And I checked with stage1 compiler the compiler_rt tests.

I have the other overflow cases also done, but I still need to generate test cases (especially for addv, subv and mulv) and the panic handler for testing panics.

PR #10404, opened 2021-12-24
std: Fix using `fs.Dir.Iterator` twice

### LINKED ISSUE #10317: `std.fs.Dir.walk` only works once
### Zig Version

0.9.0-dev.1815+20e19e75f

### Steps to Reproduce

```
$ zig init-exe
```

place the following contents in `src/main.zig`

```zig
const std = @import("std");

test {
    const allocator = std.testing.allocator;
    var dir = try std.fs.cwd().openDir("src", .{ .iterate = true });
    defer dir.close();
    const expected = try dirEntryCount(allocator, dir);
    const actual = try dirEntryCount(allocator, dir);
    try std.testing.expectEqual(expected, actual);
}

fn dirEntryCount(allocator: *std.mem.Allocator, dir: std.fs.Dir) !usize {
    var count: usize = 0;
    var walker = try dir.walk(allocator);
    defer walker.deinit();
    while (try walker.next()) |_| {
        count += 1;
    }
    return count;
}
```

```
$ zig build
```

### Expected Behavior

```
All 1 tests passed.
```

### Actual Behavior

```
Test [1/1] test ""... expected 1, found 0
Test [2/1] test ""... FAIL (TestExpectedEqual)
/home/snuc/.local/share/zig/lib/std/testing.zig:79:17: 0x20c645 in std.testing.expectEqual (test)
                ^
/home/snuc/dev/aquila/test.zig:9:5: 0x20b97a in test "" (test)
    try std.testing.expectEqual(expected, actual);
    ^
```

ominitay:
This fixes the use of multiple `Iterator`s in a row on a directory. Previously, on many platforms, using an `Iterator` on an already-iterated directory would give no entries.

Fixing this involved seeking to the beginning of the directory on the first call of `next()`.

cc: @nektro @squeek502 
fixes #10317

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -300,6 +300,7 @@ pub const Dir = struct {
             buf: [8192]u8, // TODO align(@alignOf(os.system.dirent)),
             index: usize,
             end_index: usize,
+            first_iter: bool,
 
             const Self = @This();
 
@@ -319,6 +320,10 @@ pub const Dir = struct {
             fn nextDarwin(self: *Self) !?Entry {
                 start_over: while (true) {
                     if (self.index >= self.end_index) {
+                        if (self.first_iter) {
+                            std.os.lseek_SET(self.dir.fd, 0) catch unreachable; // EBADF here likely means that the Dir was not opened with iteration permissions
+                            self.first_iter = false;
+                        }
                         const rc = os.system.__getdirentries64(
                             self.dir.fd,
                             &self.buf,
@@ -369,6 +374,10 @@ pub const Dir = struct {
             fn nextSolaris(self: *Self) !?Entry {
                 start_over: while (true) {
                     if (self.index >= self.end_index) {
+                        if (self.first_iter) {
+                            std.os.lseek_SET(self.dir.fd, 0) catch unreachable; // EBADF here likely means that the Dir was not opened with iteration permissions
+                            self.first_iter = false;
+                        }
                         const rc = os.system.getdents(self.dir.fd, &self.buf, self.buf.len);
                         switch (os.errno(rc)) {
                             .SUCCESS => {},
@@ -423,6 +432,10 @@ pub const Dir = struct {
             fn nextBsd(self: *Self) !?Entry {
                 start_over: while (true) {
                     if (self.index >= self.end_index) {
+                        if (self.first_iter) {
+                            std.os.lseek_SET(self.dir.fd, 0) catch unreachable; // EBADF here likely means that the Dir was not opened with iteration permissions
+                            self.first_iter = false;
+                        }
                         const rc = if (builtin.os.tag == .netbsd)
                             os.system.__getdents30(self.dir.fd, &self.buf, self.buf.len)
                         else
@@ -479,6 +492,7 @@ pub const Dir = struct {
             buf: [8192]u8, // TODO align(@alignOf(os.dirent64)),
             index: usize,
             end_index: usize,
+            first_iter: bool,
 
             const Self = @This();
 
@@ -491,6 +505,10 @@ pub const Dir = struct {
                     // TODO: find a better max
                     const HAIKU_MAX_COUNT = 10000;
                     if (self.index >= self.end_index) {
+                        if (self.first_iter) {
+                            std.os.lseek_SET(self.dir.fd, 0) catch unreachable; // EBADF here likely means that the Dir was not opened with iteration permissions
+                            self.first_iter = false;
+                        }
                         const rc = os.system._kern_read_dir(
                             self.dir.fd,
                             &self.buf,
@@ -563,6 +581,7 @@ pub const Dir = struct {
             buf: [8192]u8 align(if (builtin.os.tag != .linux) 1 else @alignOf(linux.dirent64)),
             index: usize,
             end_index: usize,
+            first_iter: bool,
 
             const Self = @This();
             const linux = os.linux;
@@ -574,6 +593,10 @@ pub const Dir = struct {
             pub fn next(self: *Self) Error!?Entry {
                 start_over: while (true) {
                     if (self.index >= self.end_index) {
+                        if (self.first_iter) {
+                            std.os.lseek_SET(self.dir.fd, 0) catch unreachable; // EBADF here likely means that the Dir was not opened with iteration permissions
+                            self.first_iter = false;
+                        }
                         const rc = linux.getdents64(self.dir.fd, &self.buf, self.buf.len);
                         switch (linux.getErrno(rc)) {
                             .SUCCESS => {},
@@ -620,7 +643,7 @@ pub const Dir = struct {
             buf: [8192]u8 align(@alignOf(os.windows.FILE_BOTH_DIR_INFORMATION)),
             index: usize,
             end_index: usize,
-            first: bool,
+            first_iter: bool,
             name_data: [256]u8,
 
             const Self = @This();
@@ -645,9 +668,9 @@ pub const Dir = struct {
                             .FileBothDirectoryInformation,
                             w.FALSE,
                             null,
-                            if (self.first) @as(w.BOOLEAN, w.TRUE) else @as(w.BOOLEAN, w.FALSE),
+                            if (self.first_iter) @as(w.BOOLEAN, w.TRUE) else @as(w.BOOLEAN, w.FALSE),
                         );
-                        self.first = false;
+                        self.first_iter = false;
                         if (io.Information == 0) return null;
                         self.index = 0;
                         self.end_index = io.Information;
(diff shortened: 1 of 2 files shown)


kubkon:
I take it that #10317 is still not fixed hence why the CI fails with this test added? If that's the case, I see two solutions for this PR to be in mergable state: 1) if you're up for it, you could propose a fix to #10317, or 2) disable this test for the time being an leave a TODO comment referencing the issue. Lemme know what you think!

nektro:
i think `// TODO https://github.com/ziglang/zig/issues/10317` is a good route, there's precedence

ominitay:
Yep, sorry, my PRs have fallen behind. I'll get this sorted in a moment to be merged.

ominitay:
I'd like to see the Windows CI pass before this is merged, to be sure :)

ominitay:
Ok, ready to merge! @kubkon

lib/std/fs.zig
```diff
@@ -300,6 +300,7 @@ pub const Dir = struct {
             buf: [8192]u8, // TODO align(@alignOf(os.system.dirent)),
             index: usize,
             end_index: usize,
+            first: bool,
```
kubkon, line 303: Would you mind renaming the field to something less ambiguous like for example `first_scan` or `first_iteration`? IMHO this will greatly improve readability of this code.

lib/std/fs.zig
```diff
@@ -369,6 +370,10 @@ pub const Dir = struct {
             fn nextSolaris(self: *Self) !?Entry {
                 start_over: while (true) {
                     if (self.index >= self.end_index) {
+                        if (self.first) {
+                            std.os.lseek_SET(self.dir.fs, 0) catch unreachable;
```
kubkon, line 374: Are we sure this cannot fail?

ominitay: It definitely can't on Linux, I assumed POSIX compliant systems would be the same, though I just pushed this to see how the tests run.

lib/std/fs.zig
ominitay: Was doing this to match the existing Windows implementation, so should that be changed also?

kubkon: Yeah, change the field name in all children. I think in this case being verbose is a good thing.

lib/std/fs.zig
squeek502: I think an explanatory comment on each `catch unreacable` would be nice, but the possible errors for each platform are (from [`lseek(2)`](https://man7.org/linux/man-pages/man2/lseek.2.html)):

- `EBADF` when Dir.fd is invalid or wasn't opened with iteration. This `catch unreachable` would show up later anyway during the `next` call (see [here and other similar lines](https://github.com/ziglang/zig/blob/35423b005440c9f31a4cc6a53ccf6b7edd08b859/lib/std/fs.zig#L580)), so it being triggered during the `lseek_SET` just moves where it happens.
- `EINVAL` when "whence is not valid or the resulting file offset would be negative, or beyond the end of a seekable device.", but we're always using the known-to-be-valid `SEEK_SET` and the offset is always `0`
- `EOVERFLOW` when "The resulting file offset cannot be represented in an off_t", but we know `0` will be the resulting file offset
- `ENXIO` when lseek is using `SEEK_DATA` or `SEEK_HOLE` (actually, this isn't even a possible error within lseek_SET since we know we're using `SEEK_SET`)
- `ESPIPE` when "fd is associated with a pipe, socket, or FIFO.", but we know `fd` is a directory

WASI has an additional `ENOTCAPABLE` that I'm unsure about when/if it can be triggered.

kubkon: Hmm, I see. Curious to see if it applies to BSD/macOS too. I remember in the past I had some mixed experience with Dir iterators on macOS - the way the dir pointer works was completely different to how it works on Linux. But let's see if the tests pass. If so, then we're golden!

ominitay: Yeah, worth looking further into this.

lib/std/fs.zig
```diff
@@ -369,6 +370,10 @@ pub const Dir = struct {
             fn nextSolaris(self: *Self) !?Entry {
                 start_over: while (true) {
                     if (self.index >= self.end_index) {
+                        if (self.first) {
+                            std.os.lseek_SET(self.dir.fd, 0) catch unreachable;
```
squeek502, line 374: For better debugging, I think adding a comment similar to the ones in the `Iterator.next` functions here would be nice (and the same comment on all the `lseek_SET catch unreachable` calls):

```suggestion
                            std.os.lseek_SET(self.dir.fd, 0) catch unreachable; // EBADF here likely means that the Dir was opened without iteration ability
```

See [this comment](https://github.com/ziglang/zig/issues/10317#issuecomment-1012129796) for a comparison of the stack traces of an uniterable Dir when it happens during `lseek_SET` vs `Iterator.next`

lib/std/fs.zig
kubkon: > I think an explanatory comment on each `catch unreachable` would be nice, but the possible errors for each platform are (from [`lseek(2)`](https://man7.org/linux/man-pages/man2/lseek.2.html)):
> 
> * `EBADF` when Dir.fd is invalid or wasn't opened with iteration. This `catch unreachable` would show up later anyway during the `next` call (see [here and other similar lines](https://github.com/ziglang/zig/blob/35423b005440c9f31a4cc6a53ccf6b7edd08b859/lib/std/fs.zig#L580)), so it being triggered during the `lseek_SET` just moves where it happens.
> * `EINVAL` when "whence is not valid or the resulting file offset would be negative, or beyond the end of a seekable device.", but we're always using the known-to-be-valid `SEEK_SET` and the offset is always `0`
> * `EOVERFLOW` when "The resulting file offset cannot be represented in an off_t", but we know `0` will be the resulting file offset
> * `ENXIO` when lseek is using `SEEK_DATA` or `SEEK_HOLE` (actually, this isn't even a possible error within lseek_SET since we know we're using `SEEK_SET`)
> * `ESPIPE` when "fd is associated with a pipe, socket, or FIFO.", but we know `fd` is a directory
> 
> WASI has an additional `ENOTCAPABLE` that I'm unsure about when/if it can be triggered.

`ENOTCAPABLE` would be triggered when we want iterate a dir tree we weren't given capabilities to do so.

lib/std/fs.zig
```diff
@@ -369,6 +370,10 @@ pub const Dir = struct {
             fn nextSolaris(self: *Self) !?Entry {
                 start_over: while (true) {
                     if (self.index >= self.end_index) {
+                        if (self.first) {
```
squeek502, line 373: Aren't these conditions backwards? We only want to `lseek_SET` when it's *not* the first iteration, right?

ominitay: No. We should rewind the directory before reading entries, and only on the first iteration. When else would we do it?

lib/std/fs.zig
ominitay: Yep, good idea. Will wait for CI results to come through and then push this.

lib/std/fs.zig
squeek502: Ah, you're right, sorry, had things mixed up in my head.

ominitay:
Am confused what went on with the Linux CI there. Seems to have failed on code which I didn't touch :/
My code seemed to pass fine though.

ominitay:
Will just push this to rerun CI

nektro:
maybe rebase

ominitay:
> maybe rebase

Already did that. This is so odd. Maybe I'll try reverting the changes I did to the WASI impl, see if that does anything...

ominitay:
Seems to have done it. Somehow using `lseek_SET` in WASI broke unrelated things ...

squeek502:
It looks like `lseek_SET` is unecessary for WASI, and the rewinding is already taken care of by using `DIRCOOKIE_START` when creating the Iterator here:

https://github.com/ziglang/zig/blob/0e6d2184cacf2dd1fad7508b2f9ae99d78763148/lib/std/fs.zig#L789

`fd_readdir` and `dircookie` docs:

- https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#-fd_readdirfd-fd-buf-pointeru8-buf_len-size-cookie-dircookie---resultsize-errno
- https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#dircookie

The current implementation with no changes already passes the `Dir.Iterator twice` test case.

ominitay:
Fantastic, thank you for explaining that to me!

With that, this should be ready to merge.

lib/std/fs.zig
```diff
@@ -574,10 +593,14 @@ pub const Dir = struct {
             pub fn next(self: *Self) Error!?Entry {
                 start_over: while (true) {
                     if (self.index >= self.end_index) {
+                        if (self.first_iter) {
+                            std.os.lseek_SET(self.dir.fd, 0) catch unreachable; // Can't fail
```
squeek502, line 597: This should probably have the `// EBADF here likely means that the Dir was not opened with iteration permissions` comment as well

lib/std/fs.zig
```diff
@@ -574,10 +593,14 @@ pub const Dir = struct {
...
+                            self.first_iter = false;
+                        }
                         const rc = linux.getdents64(self.dir.fd, &self.buf, self.buf.len);
                         switch (linux.getErrno(rc)) {
                             .SUCCESS => {},
-                            .BADF => unreachable, // Dir is invalid or was opened without iteration ability
+                            .BADF => unreachable, // EBADF here likely means that the Dir was not opened with iteration permissions
```
squeek502, line 603: This comment can probably stay as it was before.

ominitay: Ah I confused the two lines ;)

Vexu (approved):

kubkon (approved):
Looks great @ominitay ! Thank you for seeing it through!

PR #10415, opened 2021-12-26
Add comprehensive tests to log and exp functions in std.math

LewisGaul:
Tracked at https://github.com/LewisGaul/zig/wiki/Maths-APIs. There are still plenty of TODOs there, but I consider this in a good state to get merged (especially as it includes multiple fixes, listed below).

This improves on the test framework added in https://github.com/ziglang/zig/pull/10149 and https://github.com/ziglang/zig/pull/10046 - I will rebase those PRs after this is merged so that the focus there can be solely on `f128` support.

- Add test framework (at `lib/std/math/test.zig`) for maths functions that take in a float and output a float of the same type, e.g. `exp()`
- Create tests for all exp/log functions under `lib/std/math/test/`
  - Sanity tests, taken from [libc-test](https://wiki.musl-libc.org/libc-test.html) (e.g. [exp.h](http://nsz.repo.hu/git/?p=libc-test;a=blob;f=src/math/sanity/exp.h;h=7cbdfcae8770955be175e824e92e535fabae4430;hb=HEAD))
  - Special cases (inf, nan, 0, ...)
  - Boundary cases (last value before inf, first subnormal, ...)
- Remove special case tests from function implementation modules to avoid duplication (leave the basic tests intact)
- Fixes:
  - `exp2_32()` flushing to zero too soon
  - `expm1_32()` giving incorrect results for a certain range of small negative inputs
  - `ln64()` giving incorrect results for positive subnormal inputs
  - Some functions returning canonical NaN instead of passing through input NaNs unchanged (as per Musl and GCC)
  - `log10_64()` returning 32-bit inf and NaN, which is fine as it's cast up, except casting NaN results in failure to pass through the input unchanged

Example output with the 'verbose' flag toggled for debugging (failing test added for illustration):
```
$./zig test --zig-lib-dir lib lib/std/math/test.zig --test-filter 'exp32() boundary'
Test [0/1] test.exp.test "math.exp32() boundary"...
 IN:  0x42B17217  0x1.62e42ep6      8.87228317e+01
OUT:  0x7F7FFF84  0x1.ffff08p127    3.40279851e+38

 IN:  0x42B17218  0x1.62e43p6       8.87228393e+01
OUT:  0x7F800000  inf               inf

 IN:  0x7F7FFFFF  0x1.fffffep127    3.40282346e+38
OUT:  0x7F800000  inf               inf

 IN:  0x00000001  0x0.000002p-126   1.40129846e-45
OUT:  0x3F800000  0x1p0             1.0e+00

 IN:  0x80000001  -0x0.000002p-126  -1.40129846e-45
OUT:  0x3F800000  0x1p0             1.0e+00

 IN:  0x00800000  0x1p-126          1.17549435e-38
OUT:  0x3F800000  0x1p0             1.0e+00

 IN:  0x80800000  -0x1p-126         -1.17549435e-38
OUT:  0x3F800000  0x1p0             1.0e+00

 IN:  0xC2CFF1B4  -0x1.9fe368p6     -1.03972076e+02
OUT:  0x00000001  0x0.000002p-126   1.40129846e-45

 IN:  0xC2CFF1B5  -0x1.9fe36ap6     -1.03972084e+02
OUT:  0x00000000  0x0.0p0           0.0e+00

 IN:  0xC2AEAC4F  -0x1.5d589ep6     -8.73365402e+01
OUT:  0x00800026  0x1.00004cp-126   1.17549967e-38

 IN:  0xC2AEAC50  -0x1.5d58ap6      -8.73365478e+01
OUT:  0x007FFFE6  0x0.ffffccp-126   1.17549070e-38

 IN:  0x3FE07FF7  0x1.c0ffeep0      1.75390517e+00
OUT:  0x40B8DE2A  0x1.71bc54p2      5.77711963e+00
EXP:  0x3FDF778F  0x1.beef1ep0      1.74583613e+00
FAILURE: expected exp(0x1.c0ffeep0)->0x1.beef1ep0, got 0x1.71bc54p2 (32-bit)

Subtest summary: 11 passed; 1 failed
Test [1/1] test.exp.test "math.exp32() boundary"... FAIL (Failure)
lib/std/math/test.zig:90:17: 0x21769a in Testcase(std.math.exp.exp,"exp",f32).run (test)
                ^
lib/std/math/test.zig:111:23: 0x21667c in runTests (test)
    if (failures > 0) return error.Failure;
                      ^
lib/std/math/test/exp.zig:76:5: 0x216055 in test.exp.test "math.exp32() boundary" (test)
    try runTests(cases);
    ^
lib/std/math/zig-cache/o/a0bcd24bc7b2dc984cfd0a4c01be1d2e/test ./zig
```


LewisGaul:
@matu3ba @tiehuis @viriuwu

andrewrk (approved):

LewisGaul:
Is there something that needs adding somewhere to make these tests run with `zig build test`? I'm currently looking for it...

LewisGaul:
> Is there something that needs adding somewhere to make these tests run with `zig build test`? I'm currently looking for it...

Added `_ = @import("math/test.zig");` in a test block in `lib/std/math.zig`, seems to have done the trick :)

lib/std/math/test/log1p.zig
```diff
@@ -0,0 +1,103 @@
...
+const nan64 = math.nan_f64;
+
+const Tc32 = Testcase(math.log1p, "log1p", f32);
+const tc32 = Tc32.init;
+
+const Tc64 = Testcase(math.log1p, "log1p", f64);
+const tc64 = Tc64.init;
+
+// Special-case tests shared between different float sizes, see genTests().
+const special_tests = .{
```
matu3ba, line 17: For review its easier to write the special mathematical properties of log1p as set notation where log1p is defined as what value either above the test cases or near the top:
```txt
// log1p(a) - natural logarithm of one plus the input array, element-wise
in -> out
[-inf .. -min] -> -inf
..
[max .. inf] -> inf
```
Also order numbers from min to max input for consistency to minimize risk of leaving any gaps in the wanted test set.

lib/std/math/test.zig
```diff
@@ -0,0 +1,177 @@
+const std = @import("../std.zig");
+const print = std.debug.print;
+const meta = std.meta;
+const math = std.math;
+const bitCount = meta.bitCount;
+const nan = math.nan;
+
+// Switch to 'true' to enable debug output.
+var verbose = false;
```
matu3ba, line 9: The verbose flag should be justified. Why not use comptime log scope here? Tests are relative fast to compile and you also dont obtain the value from the test runner via cmd flag or stuff like that.

lib/std/math/test/log1p.zig
```diff
@@ -0,0 +1,103 @@
...
+        // zig fmt: on
+    };
+    try runTests(cases);
+}
+
+test "math.log1p_64() sanity" {
+    const cases = [_]Tc64{
+        // zig fmt: off
+        tc64(-0x1.02239f3c6a8f1p+3,  nan64               ),
+        tc64( 0x1.161868e18bc67p+2,  0x1.ad1bdd1e9e686p+0), // Disagrees with GCC in last bit
```
matu3ba, line 71: If you have stuff like `Disagrees`, always put a `TODO` so you later dont forget.

lib/std/math/test/exp.zig
```diff
@@ -0,0 +1,131 @@
...
+
+test "math.exp64() special" {
+    const cases = test_utils.genTests(Tc64, special_tests) ++ test_utils.nanTests(Tc64);
+    try runTests(cases);
+}
+
+test "math.exp64() boundary" {
+    const cases = [_]Tc64{
+        // zig fmt: off
+        tc64( 0x1.62e42fefa39efp+9,    0x1.fffffffffff2ap+1023), // The last value before the result gets infinite
```
matu3ba, line 117: when I read stuff like this I am wondering, why no `-inf64`?
Why not sort from min to max input?

matu3ba (changes requested):
Good job so far! It looks like a pile of work.

Verbose needs explanation and the tests have no great documentation in what input range you want to test.

Take as example https://github.com/ziglang/zig/blob/405ff911dae3ca10af380e5dd8e2dfda4a570191/lib/std/special/compiler_rt/negsi2_test.zig#L22
where you can see how input range maps to what is the representation.

Other than that, this needs likely to be eventually split into multiple PRs.

lib/std/math/test/log1p.zig
LewisGaul: Well, this was explicitly *not* considered as a TODO, rather just a note - this value agrees with Musl (which our implementation is based on) but not with GCC.

Musl seems to say the guarantee is that the error for `log1p()` is always less than 1 ULP (see [here](https://git.musl-libc.org/cgit/musl/tree/src/math/log1p.c#n36)), which leaves room for the last bit being wrong - are you saying this is a violation of IEEE754 and a different implemetation should be used? I'm not sure whether the GCC guarantees are stronger.

lib/std/math/test/exp.zig
LewisGaul: My intention was the following:
- 'special' tests covering special inputs, e.g. `0`, `1`, `+/-inf`, various NaNs, special numbers such as `ln2` or `e`
- 'boundary' tests covering special outputs (first/last inputs giving results in the different parts of the output range), e.g. last input before result is inf, first input where result is inf, ...
- 'sanity' tests for some other random/regular inputs just to give a sanity check over whether the implementation is working for mainline cases (I don't consider the actual values that important, since the special/boundary tests should be covering the full domain/range)

I ended up also including boundary *input* values (not covered in the 'special' tests) in the 'boundary' tests too though, such as float max/min/true min, which are definitely worth covering (there are code branches for handling subnormal inputs for example). There didn't seem to be much value in including all boundary inputs if they result in the same output (e.g. `exp(x) -> 0`, where `x < -1p+7`, so why bother with `-floatMax()` when checking the first value that hits zero). However, I'm fine with including all those cases more systematically if you think that's worth doing.

lib/std/math/test.zig
LewisGaul: Yeah, I was thinking this could be hooked up better. Any advice on how to do so? I'm not hugely familiar with running the Zig test suite. What do you mean by comptime log scope? How would I obtain "the value" from the test runner?

lib/std/math/test/log1p.zig
LewisGaul: Ok, I'll add a comment at the top of each test file explaining the domain->range mapping. Sorting the tests seems sensible too.

LewisGaul:
> Not sure, if this should be split into multiple PRs. I tend to correct comments etc to clarify things, if I add another PR with the same type of stuff.

Sorry, I'm not sure what you meant by the second sentence above. How were you thinking this PR should be split? My thinking was that the test infra doesn't make sense without the tests, the tests don't pass without the code fixes, and the code fixes should be accompanied by tests! I guess it could be split by only adding the tests at first, with failing testcases commented out with TODOs, and then (a) separate PR(s) raised for the actual code fixes and uncommenting the failing testcases - would that be useful from a tracking perspective, e.g. for release notes?

> Deviation to GCC should be explainied, as the standard is explicit on errors: there must be none.

It would seem reasonable for the standard to say that, but I haven't seen that... could you point me to where it says that? My reading of the various methods for computing these functions (mainly based on comments at the top of the implementations) is that the max error is 1 ULP or less, but this does leave room for the last bit to be wrong. If this is the case then I'm not sure whether it's Musl or GCC that has the last-bit error in these cases! Note there are more that I've identified:
- `exp_32()`: https://github.com/LewisGaul/zig-f128math/blob/main/hypothesis/test.py#L124
- `log2_32()`: https://github.com/LewisGaul/zig-f128math/blob/main/hypothesis/test.py#L168
- `log2_64()`: https://github.com/LewisGaul/zig-f128math/blob/main/hypothesis/test.py#L177

lib/std/math/test/log1p.zig
matu3ba: You probably want this note then near the top level decl.
I checked the musl source and it looks like they ported all their tests to be out of tree into [libc-test](https://wiki.musl-libc.org/libc-test.html).
Their thoroughly tests are also GPLed, but most are MIT.
For example, log1p.h has 550 test cases heavily commented.

If you run `rg 02239f3c6a8f1` in their repo you get alot of matches on `-0x1.02239f3c6a8f1p+3` and a macro like this:
```c
T(RN,   -0x1.02239f3c6a8f1p+3,                     nan,          0x0p+0, INVALID)
```

Should we port this behavior or do you even want to write a helper script to convert that macro stuff?
To me, it sounds simpler to track all upstream test changes then and you dont need to manually port them.

matu3ba:
2 ideas for discussion

1. port [libc-test](https://wiki.musl-libc.org/libc-test.html) to zig
- 1.1. Port the test running functions from [libc-test](https://wiki.musl-libc.org/libc-test.html), which is the test library of musl. 
- 1.2. Use scripts to generate the test code from the musl test macros.
- more code to write, but less dependency on C/clang
- caveat, but actually a feature: Include the comments in the generated test code.

2. vendor [libc-test](https://wiki.musl-libc.org/libc-test.html) 
- 2.1 use zig stubs for the build scripts and test calls
- less code to write, but more dependency on C/clang

lib/std/math/test.zig
matu3ba: Use `log.info` like in `heap/logging_allocator.zig`.

Usage from the test runner would be then in the test block or file
`pub const log_level: std.log.Level = .info;`, but this should either be a comment or removed.
Make sure you dont get name clashes, ie prefix log with `std.log`.

matu3ba:
@LewisGaul
> It would seem reasonable for the standard to say that, but I haven't seen that

I checked again and 1lub deviation is accepted for regular IEEE floating point functions. Fastmath conforming floating stuff is accepted with up 4lub deviation.
Unfortunately they dont include how they came up with the [error bound for the specific input or who generated them etc](https://git.musl-libc.org/cgit/musl/tree/src/math/log1p.c#n36):
```txt
 * Accuracy:
 *      according to an error analysis, the error is always less than
 *      1 ulp (unit in the last place).
 ```
[Here is an example how to dervie them](https://cs.stackexchange.com/a/111329).

andrewrk:
@LewisGaul I'm very interested in getting this merged. Are you willing to do the rebase? Just need that green CI and then we're good to go.

andrewrk:
Closing because it's old, has conflicts, and not passing the CI. Above comment stands - would love to see this resurrected!

LewisGaul:
Hey @andrewrk definitely planning on coming back to this PR and the related ones that had been languishing. Closing these for now seems the right approach and I can rebase before creating new ones. BTW I think the CI issues were only due to an underlying f128 issue, I'm not sure whether that's been fixed yet.

andrewrk:
There have been a lot of `f80` and `f128` fixes in the last month. I'll pay closer attention if you open a fresh PR and get involved in fixing any bugs that you surface this time. I'm sorry for dropping the ball on that last time.

PR #10419, opened 2021-12-27
compiler_rt: add __negvsi2, __negvdi2, __negvti2

matu3ba:
- neg can only overflow, if a == MIN
- case `-0` is properly handled by hardware, so overflow check by comparing
  `a == MIN` is sufficient
- tests: MIN, MIN+1, MIN+4, -42, -7, -1, 0, 1, 7..

See #1290


matu3ba:
There is no panic handler for testing panics yet, so I left TODOs.

I tested manually the overflow case and complete range of i32.
And I checked with stage1 compiler the compiler_rt tests.

next: overflow testing with the fork workaround as suggested by Vesim.
thereafter: refactor the other ugly overflow code from LLVM

andrewrk (approved):

PR #10422, opened 2021-12-27
fchown: use the 32-bit uid/gid variant of the syscall on 32-bit linux targets.

vesim987:
Before this change it was truncating the uid and gid to u16 on 32-bit targets.


vesim987:
The CI issue on drone seems unrelated

daurnimator (approved):

daurnimator:
FYI @vesim987 it looks like you didn't sign your commit, but have set up your github profile to require it.
image

PR #10427, opened 2021-12-27
stage1: fix access of slice sentinel at comptime

### LINKED ISSUE #8286: Bug when coercing sentinel-terminated array pointer to sentinel-terminated slice
Extracted from [here](https://zigforum.org/t/sentinel-terminated-slice-creation/143/4?u=satisfiedboolean).

Consider [this playground](https://www.godbolt.org/z/KxodTo):

```zig
comptime {
    const str0 = "123";

    //_ = &str0[0]; //Uh-oh.

    const slice0: [:0]const u8 = str0;

    @compileLog(slice0.len);
    @compileLog(slice0[slice0.len]);
}
```
```zig
| 3
| 0
```

This works as expected.

Now, uncomment the line and you get an out of bounds error while indexing on `slice0`:

```zig
| 3
<source>:8:23: error: out of bounds slice
    @compileLog(slice0[slice0.len]);
                      ^
```

As if indexing on `str0` somehow removed its sentinel.

If you take an array literal instead of a string it straight up fails (probably related under the hood):

```zig
comptime {
    const str0 = &[_:0]u8 { '1', '2', '3' };

    const slice0: [:0]const u8 = str0;

    @compileLog(slice0.len);
    @compileLog(slice0[slice0.len]);
}
```
```zig
| 3
<source>:7:23: error: out of bounds slice
    @compileLog(slice0[slice0.len]);
                      ^
```

Edit: probably related to #8256

Vexu:
Closes #8286

### DIFF
--- a/src/stage1/ir.cpp
+++ b/src/stage1/ir.cpp
@@ -14982,12 +14982,16 @@ static Stage1AirInst *ir_analyze_instruction_elem_ptr(IrAnalyze *ira, Stage1ZirI
                         case ConstPtrSpecialSubArray:
                         case ConstPtrSpecialBaseArray:
                             {
+                                uint64_t array_len = ptr_field->data.x_ptr.data.base_array.array_val->type->data.array.len;
+                                if (ptr_field->data.x_ptr.data.base_array.array_val->type->data.array.sentinel != nullptr) {
+                                    array_len += 1;
+                                }
                                 size_t offset = ptr_field->data.x_ptr.data.base_array.elem_index;
                                 uint64_t new_index = offset + index;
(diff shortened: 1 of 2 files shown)


test/behavior/slice.zig
```diff
@@ -166,3 +166,21 @@ test "slicing zero length array" {
     try expect(mem.eql(u8, s1, ""));
     try expect(mem.eql(u32, s2, &[_]u32{}));
 }
+
+test "slice sentinel access at comptime" {
```
andrewrk, line 170: Looks like this test is not yet passing in stage2; if you move it to the slice_stage1.zig then you should be good to go.

Laremere:
Thanks and nice job on the quick response on this fix!

Issue #5177, opened 2020-04-26
replace `@setCold` with `@cold`


andrewrk:
Status quo:

```
@setCold(is_cold: bool) void
```
>  Tells the optimizer that a function is rarely called. 

Proposal:

```
@cold() void
```

Annotates that it is relatively uncommon for control flow to reach this point. Similar to `unreachable`, but only communicates probability. It communicates a willingness to compromise performance of the cold path in order to improve performance of the hot path.

This makes #489 unnecessary. Instead of:

```zig
if (@expect(foo, true)) {
   bar();
} else {
   baz();
}
```

With this proposal:

```zig
if (foo) {
    bar();
} else {
    @cold();
    baz();
}
```

SpexGuy:
How does this work with comptime-known coldness?  The old approach would have been:
```zig
fn foo() void {
    @setCold(comptime_foo_is_cold);
}
```
But with the new approach:
```
fn foo() void {
    if (comptime_foo_is_cold) {
        @cold(); // doesn't work, applies to the if body.
    }
    // function implementation
}
```

SpexGuy:
I guess this might work, since technically there's no scope:
fn foo() void {
    if (comptime_foo_is_cold) @cold();
    // function implementation
}

andrewrk:
That later example would still apply to the then-body of the if statement, communicating that `comptime_foo_is_cold` is unlikely. Which if the value is comptime-known is meaningless and ignored since the compiler knows precisely whether the branch is taken.

This proposal removes the ability to specify comptime-known coldness. I would be happy to reconsider if there is a convincing enough use case for comptime-known coldness.

fengb:
ZeeAlloc has manual size tuning using `@setCold`: https://github.com/fengb/zee_alloc/commit/b11ee8ee6747e292b375b2eae6df97dd2c866dbd#diff-d2841187941f93280c97ecaafd02aeb4R292

andrewrk:
Thanks for the example. OK I'll amend the proposal to have `@cold` accept a comptime argument, same as `@setCold`. So the proposal is now essentially:
 * `rename `@setCold` to `@cold`
 * make it apply per-branch-scope rather than per-function-scope

EleanorNB:
There's a solution over at #5239 -- simple configuration is concise (`@{cold};`), and doesn't look deceptively like a function call; also, comptime configuration is always possible, but not so easy that it encourages users to go insane with it.

andrewrk:
[...]

Issue #489, opened 2017-09-17
@expect() hint to optimizer


PavelVozenilek:
This is proposal for a small feature, local and isolated. Could improve code readability.

------------
Linux kernel uses macros `likely` and `unlikely` :

```
if (likely(x > 0)) { ... }
...
if (unlikely(y == NULL)) { ... }

```

These macros may improve performance a little bit and they serve as handy documentation.

Language Nim supports this functionality too, through its standard library ( https://nim-lang.org/docs/system.html#likely.t,bool ).

----------------------

**My proposal:**

Add `if+` and `if-` into the language. `if+` would signal that the path is rather likely, `if-` will suggest the flow will not go this way.

```
if+ (x < 0) { 
  ...  // likely goes here
}
if- (err) { 
  ... // unlikely
}

```

What is this good for:
1. It gives hint to source code reader, hint which will be almost never found in the documentation. This hint is very short, intuitive, and doesn't require extra pair of parenthesis.
2. It may slightly improve the performance, by rearranging instructions so that the likely path goes w/o jump, or by inserting hint for branch selector. 

How it could be implemented:
1. By doing nothing, just using it as documentation only feature.
2. By using IR opcode `llvm.expect`. This is how clang implements `__builtin_expect`, which is called by `likely`/`unlikely `macros. (GCC also provides `__builtin_expect`, MSVC has nothing similar.)

-------------------

Performance effects of `__builtin_expect` are hotly disputed on the internet. Many people claim programmers are invariably bad in such prediction and profile guided optimization will do much, much better. However, they never support their claim by benchmarks.

Even if performance is unaffected the documentation value remains. When one is stepping the code through debugger and the flow goes against the hint, one may be get more cautious a catch a bug.

raulgrell:
I do think that something to guide branch prediction can be pretty neat, but might be clearer with a builtin like @expect(expression, value) or @likely(condition).

This would be less surprising to someone who's familiar with llvm.expect and gcc __builtin_expect. A built in might also be better suited as it communicates a hint to the compiler as opposed to actual program logic.

```
if (@expect(x < 0, false)) {
    //
}
```

PavelVozenilek:
I use `likely`/`unlikely` in my code and the additional parenthesis make have impact on readability. Alternative syntax  could be something as
```

if (x > 0) {
  @likely
   ...
} else {
  ...
}

```
but this wastes one precious line.

raulgrell:
[...]

PR #10428, opened 2021-12-28
Improve stdlib's random float generation

mrakh:
Status quo is that to generate a random value in the half open interval [0, 1), a random value in [1, 2) is generated via bit fiddling, and then 1 is subtracted.

The problem is that, in IEEE-754, the number of distinct values in [1, 2) is significantly smaller than that of [0, 1). In the single-precision case, only 1/127 of the available range is covered. In the double-precision case, the ratio is even less, only 1/1023.

This commit fixes these shortcomings by directly generating the mantissa and exponent. The mantissa bits are extracted directly from the RNG, and `@clz` is used to apply an exponential bias for the exponent bits. With the new implementation, the available range significantly increases.


lib/std/rand.zig
```diff
@@ -247,18 +247,34 @@ pub const Random = struct {
 
     /// Return a floating point value evenly distributed in the range [0, 1).
     pub fn float(r: Random, comptime T: type) T {
-        // Generate a uniform value between [1, 2) and scale down to [0, 1).
-        // Note: The lowest mantissa bit is always set to 0 so we only use half the available range.
+        // Generate a uniformly random value between for the mantissa.
+        // Then generate an exponentially biased random value for the exponent.
+        // Over the previous method, this has the advantage of being able to represent
```
matu3ba, line 252: can you be more accurate here: how much more of the available range?

lib/std/rand.zig
```diff
@@ -247,18 +247,34 @@ pub const Random = struct {
...
+        // much more of the available range.
         switch (T) {
             f32 => {
-                const s = r.int(u32);
-                const repr = (0x7f << 23) | (s >> 9);
-                return @bitCast(f32, repr) - 1.0;
+                // Use the lowest 23 bits for the mantissa, and use the rest to calculate
+                // the exponent. Using only 41 bits for calculating the exponent instead
+                // of the full 128 means that the minimum generatable value is 2^-42, but
```
matu3ba, line 258: s/, but//
s/ it / to /

matu3ba (commented):
If you describe "Over the previous method, this has the advantage of being able to represent much more of the available range.", then you might want to explain this more properly in the comments.

Is there a way to test this stuff?

lib/std/rand.zig
```diff
@@ -247,18 +247,34 @@ pub const Random = struct {
...
+                // it minimizes entropy waste and avoids branching.
+                const rand = r.int(u64);
+                const rand_lz = @clz(u64, rand | 0x7FFFFF);
+                const mantissa = @truncate(u32, rand & 0x7FFFFF);
```
daurnimator, line 262: the `&` here is redundant with the `@truncate`.

either use (preferred)
```suggestion
                const mantissa = @truncate(u23, rand);
```
or
```suggestion
                const mantissa = @intCast(u32, rand & 0x7FFFFF);
```

daurnimator (changes requested):
Is this really still uniform over the range?

mrakh:
I've made some changes:

* Added some logic so that the randgen should now be able to generate every possible IEEE-754 value in the [0, 1) range, even the subnormals
* Added cold path test for random float function
* Added chi-square goodness of fit test for random float function

> If you describe "Over the previous method, this has the advantage of being able to represent much more of the available range.", then you might want to explain this more properly in the comments.
> 
> Is there a way to test this stuff?

The goodness of fit test that I added should now verify that the generated values come from a uniformly distributed RNG. The test currently checks that the p-value is > 0.05; feel free to adjust the numbers to your liking. I personally experimented with 100,000,000 random values and 10,000 buckets and got p-values between 0.2 and 0.7, depending on the RNG seed, but that took quite a few seconds for me to run.

> Is this really still uniform over the range?

Yes, it should be. For a given exponent value, the distribution of mantissa bits should be uniform, so the mantissa bits can be directly extracted from the RNG. The exponent value is exponentially distributed, since each successive exponent spans double the range of the previous one. In the [0, 1) interval, the exponent must be an integer in the interval [0, 127), with successive values being twice as likely. And if you compute the `clz` on an N-bit uniform RNG, you exponentially distributed RNG in the interval [0, N] where each successive number is half as likely. So if you compute the `clz` of a 126-bit number, then subtract it from 126, you should get exponent bits for a uniformly randomly chosen IEEE-754 number within [0, 1).

mrakh:
CI failure is due to timeout, but the relevant tests pass.

lib/std/rand.zig
```diff
@@ -15,6 +15,8 @@ const math = std.math;
 const ziggurat = @import("rand/ziggurat.zig");
 const maxInt = std.math.maxInt;
 
+const Dilbert = @import("rand/Dilbert.zig");
```
daurnimator, line 18: If you name the file with `_test.zig` in the name then it won't get installed as part of the standard library.

lib/std/rand.zig
```diff
@@ -247,18 +249,51 @@ pub const Random = struct {
...
-                const repr = (0x7f << 23) | (s >> 9);
-                return @bitCast(f32, repr) - 1.0;
+                // Use 23 random bits for the mantissa, and the rest for the exponent.
+                // If all 41 bits are zero, generate additional random bits, until a
+                // set bit is found, or 126 bits have been generated.
+                const rand = r.int(u64);
+                var rand_lz = @clz(u64, rand | 0x7FFFFF);
+                if (rand_lz == 41) {
+                    rand_lz += @clz(u64, r.int(u64));
+                    if (rand_lz == 41 + 64) {
```
daurnimator, line 265: I imagine this branching may have a significant speed penalty; has this been benchmarked at all?

lib/std/rand.zig
```diff
@@ -247,18 +249,51 @@ pub const Random = struct {
...
+                        // It is astronomically unlikely to reach this point.
```
r00ster91, line 266: Then wouldn't it be good to use `@setCold` or something here? It probably doesn't work here though because this is not in a function. Maybe if you extract this part into an `inline` function and use `@setCold(true)` there?

Vexu: `@setCold` is function specific and I don't know how well `cold` attributes work on inline functions, better to just add a comment waiting for #5177 or #489 to be implemented.

r00ster91: Ah, I didn't know about those. Yes, those seem perfect for this situation.
```suggestion
                        // TODO: when #5177 or #489 is implemented
                        // tell the compiler it is astronomically unlikely to reach this point.
```

lib/std/rand.zig
mrakh: Tested `std.rand.Random.float(f32)`, old version vs new version without the branch vs. new version with the branch. Units are in nanoseconds per call in the below results:

```
Non-inlined Random.float(f32):
old: mean = 11.0096, std_dev = 0.2327
new, no branching: mean = 13.0287, std_dev = 0.1954 (18.3% slower)
new, with branching: mean = 13.8511, std_dev = 0.2545 (25.8% slower)

Inlined Random.float(f32):
old: mean = 1.7341, std_dev = 0.0437
new, no branching: mean = 1.8877, std_dev = 0.0081 (8.8% slower)
new, with branching: mean = 2.3880, std_dev = 0.0118 (37.7% slower)
```

I experimented with changing the comparison so the CPU could compute it in parallel with `clz` and shorten the dependency chain:

```
  const rand = r.int(u64);
  var rand_lz = @clz(u64, rand | 0x7FFFFF);
- if (rand_lz == 41) {
+ if ((rand | 0x7FFFFF) == 0x7FFFFF) {
```

This made the inline case faster, but the non-inline case *slower*:

```
Non-inlined Random.float(f32): 
new, with updated branching: mean = 14.7841, std_dev = 0.0527 (34.2% slower)

Inlined Random.float(f32): 
new, with updated branching: mean = 2.3205, std_dev = 0.0093 (33.8% slower)
```

I might dig into this a little further with `perf`.

daurnimator:
So something I thought of today is that this opens up a timing attack that wasn't there previously: certain numbers will take different amounts of time to generate.
@jedisct1 got thoughts on this?

mrakh:
> So something I thought of today is that this opens up a timing attack that wasn't there previously: certain numbers will take different amounts of time to generate.
> @jedisct1 got thoughts on this?

I’m not sure what use case requires cryptographically secure generation of floating point values, but fair point. We can address the `f32` case by always requesting the worst-case amount of necessary entropy (126 + 23 = 149 bits).

But the `f64` worst-case entropy is much higher, at 1022 + 52 = 1074 bits. Requesting that many random bits for every `f64` we wish to generate would have a quite negative impact on performance. On the other hand, generating *n* fewer random bits would mean being unable to generate any value in the [0, 2^(*n*-1023)) interval. Perhaps requesting a fixed 128 random bits per `f64` is a good compromise - 52 random bits for the mantissa, and 76 for the exponent? This puts the minimum generatable `f64` value to 2^-77, which I would say is acceptably low.

jedisct1:
> So something I thought of today is that this opens up a timing attack that wasn't there previously: certain numbers will take different amounts of time to generate.
> @jedisct1 got thoughts on this?

Don't worry about it.

Floating point representation is never used in cryptography.

The only exception I can think of abuse floating point registers as additional registers to store 53-bit integers. They don't use an actual floating point representation.

andrewrk:
Thank you for the enhancement!

PR #10434, opened 2021-12-28
std.crypto add RIPEMD-160 impl.

hazeycode:
This is an implementation of [RIPEMD-160](https://homes.esat.kuleuven.be/~bosselae/ripemd160.html)

RIPEMD-160 is not as common as the SHA-1 and SHA-2 hash algorithms, however, an example of a well known usage is in conjunction with SHA-256 to create Bitcoin addresses.

### DIFF
--- a/lib/std/crypto/ripemd160.zig
+++ b/lib/std/crypto/ripemd160.zig
@@ -101,41 +101,11 @@ fn compress(hash: *[5]u32, words: [16]u32) void {
     round(bfn2, -3, 0x7a6d76e9, permutations[8], shifts[8], &right, words);
     round(bfn1, -4, 0x00000000, permutations[9], shifts[9], &right, words);
 
-    right[3] = blk: {
-        var x: u32 = right[3];
-        _ = @addWithOverflow(u32, x, left[2], &x);
-        _ = @addWithOverflow(u32, x, hash[1], &x);
-        break :blk x;
-    };
-
-    hash[1] = blk: {
-        var x: u32 = hash[2];
-        _ = @addWithOverflow(u32, x, left[3], &x);
-        _ = @addWithOverflow(u32, x, right[4], &x);
-        break :blk x;
-    };
-
-    hash[2] = blk: {
-        var x: u32 = hash[3];
-        _ = @addWithOverflow(u32, x, left[4], &x);
-        _ = @addWithOverflow(u32, x, right[0], &x);
-        break :blk x;
-    };
-
-    hash[3] = blk: {
-        var x: u32 = hash[4];
-        _ = @addWithOverflow(u32, x, left[0], &x);
-        _ = @addWithOverflow(u32, x, right[1], &x);
-        break :blk x;
-    };
-
-    hash[4] = blk: {
-        var x: u32 = hash[0];
-        _ = @addWithOverflow(u32, x, left[1], &x);
-        _ = @addWithOverflow(u32, x, right[2], &x);
-        break :blk x;
-    };
-
+    right[3] = right[3] +% left[2] +% hash[1];
+    hash[1] = hash[2] +% left[3] +% right[4];
+    hash[2] = hash[3] +% left[4] +% right[0];
+    hash[3] = hash[4] +% left[0] +% right[1];
+    hash[4] = hash[0] +% left[1] +% right[2];
     hash[0] = right[3];
 }
 
@@ -151,30 +121,11 @@ fn round(
     comptime var r = off;
     comptime var i: usize = 0;
     inline while (i < 16) : (i += 1) {
-        hash[@mod(r + 0, 5)] = blk: {
-            var x: u32 = hash[@mod(r + 0, 5)];
-            _ = @addWithOverflow(u32, x, bfn(hash[@mod(r + 1, 5)], hash[@mod(r + 2, 5)], hash[@mod(r + 3, 5)]), &x);
-            _ = @addWithOverflow(u32, x, words[p[i]], &x);
-            _ = @addWithOverflow(u32, x, constant, &x);
-            break :blk x;
-        };
-
-        hash[@mod(r + 0, 5)] = blk: {
-            var x: u32 = rol(hash[@mod(r + 0, 5)], shifts[i]);
-            _ = @addWithOverflow(u32, x, hash[@mod(r + 4, 5)], &x);
-            break :blk x;
-        };
-
-        hash[@mod(r + 2, 5)] = rol(hash[@mod(r + 2, 5)], 10);
-
+        hash[@mod(r + 0, 5)] = hash[@mod(r + 0, 5)] +% bfn(hash[@mod(r + 1, 5)], hash[@mod(r + 2, 5)], hash[@mod(r + 3, 5)]) +% words[p[i]] +% constant;
+        hash[@mod(r + 0, 5)] = std.math.rotl(u32, hash[@mod(r + 0, 5)], @intCast(u32, shifts[i])) +% hash[@mod(r + 4, 5)];
+        hash[@mod(r + 2, 5)] = std.math.rotl(u32, hash[@mod(r + 2, 5)], 10);
         r -= 1;
     }
-}
-
-/// cyclically rotates x over n bits to the left
-fn rol(x: u32, comptime n: usize) u32 {
-    if (n >= 32) @compileError("Invalid n. Valid range is 0 <= n < 32");
-    return (x << n) | (x >> (32 - n));
 }
 
 fn bfn1(x: u32, y: u32, z: u32) u32 {
@@ -199,10 +150,9 @@ fn bfn5(x: u32, y: u32, z: u32) u32 {
 
 fn testHashEql(expected: []const u8, in: []const u8) !void {
     const hash = Ripemd160.hash(in);
-    const hex_str = try std.fmt.allocPrint(std.testing.allocator, "{x}", .{std.fmt.fmtSliceHexLower(hash.bytes[0..])});
-    defer std.testing.allocator.free(hex_str);
-    //std.debug.print("\"{s}\" => \"{s}\", expected = \"{s}\"\n", .{ in, hex_str, expected });
-    try std.testing.expectEqualSlices(u8, expected, hex_str);
+    var hex_str: [40]u8 = undefined;
+    _ = try std.fmt.bufPrint(&hex_str, "{x}", .{std.fmt.fmtSliceHexLower(hash.bytes[0..])});
+    try std.testing.expectEqualSlices(u8, expected, hex_str[0..]);
 }
 
 test "RIPEMD-160 standard tests" {
--- a/lib/std/crypto/ripemd160.zig
+++ b/lib/std/crypto/ripemd160.zig
@@ -1,0 +1,218 @@
+// https://homes.esat.kuleuven.be/~bosselae/ripemd160.html
+
+const std = @import("std");
+
+pub const Ripemd160 = struct {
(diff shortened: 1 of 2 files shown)


lib/std/crypto/ripemd160.zig
```diff
@@ -0,0 +1,218 @@
...
+    return (x & z) | (y & ~z);
+}
+
+fn bfn5(x: u32, y: u32, z: u32) u32 {
+    return x ^ (y | ~z);
+}
+
+fn testHashEql(expected: []const u8, in: []const u8) !void {
+    const hash = Ripemd160.hash(in);
+    const hex_str = try std.fmt.allocPrint(std.testing.allocator, "{x}", .{std.fmt.fmtSliceHexLower(hash.bytes[0..])});
```
jedisct1, line 202: The output size is constant, so you don't need a dynamic allocation here.

jedisct1:
Thanks for implementing RIPEMD-160 in Zig!

This doesn't implement the same API as all other hash functions, though. All other functions support a state, that can be updated and finalized, as well as a writer interface. The `hash()` shortcut is also defined the same way for all hash functions, so it would be great to remain consistent :)

RIPEMD-160 is not considered safe to use any more, as collisions can be quickly found [the same way as MD4 and MD5](https://eprint.iacr.org/2004/199.pdf). It is also very slow compared to modern hash functions, and is not used in any standardized protocol. It also doesn't have any properties that existing functions in the standard library don't have.

RIPEMD-160 remains useful for specific use cases, bitcoin addresses being one. But these cases are a little bit of a niche, so maybe an externally maintained package would be a better fit than the standard library for that function.

lib/std/crypto/ripemd160.zig
```diff
@@ -0,0 +1,218 @@
...
+
+            chunk[14] = @intCast(u32, str.len << 3);
+            chunk[15] = @intCast(u32, (str.len >> 29) | (0 << 3));
+            compress(&temp, chunk);
+        }
+
+        { // write final hash
+            var i: usize = 0;
+            while (i < 160 / 8) : (i += 4) {
+                res.bytes[i] = @truncate(u8, temp[i >> 2]);
```
jedisct1, line 49: You can use `mem.writeIntLittle()` instead.

lib/std/crypto/ripemd160.zig
```diff
@@ -0,0 +1,218 @@
...
+        var ptr = str.ptr;
+
+        { // compress string in 16-word chunks
+            var i = str.len / 64;
+            while (i > 0) : (i -= 1) {
+                var chunk: [16]u32 = [_]u32{0} ** 16;
+                var j: usize = 0;
+                while (j < 16) : (j += 1) {
+                    std.mem.copy(u8, @ptrCast([*]u8, &chunk[j])[0..4], ptr[0..4]);
+                    ptr += 4;
```
jedisct1, line 20: That looks too much like a straight conversion from C code. You can simply use an index, and `mem.readIntLittle(u32, ...)` to read a 32-bit integer at a given offset.

lib/std/crypto/ripemd160.zig
```diff
@@ -0,0 +1,218 @@
...
+                compress(&temp, chunk);
+            }
+        }
+
+        { // compress remaining string, (str.len % 64) bytes
+            var chunk: [16]u32 = [_]u32{0} ** 16;
+            var i: usize = 0;
+            while (i < (str.len & 63)) : (i += 1) {
+                chunk[i >> 2] ^= (@intCast(u32, ptr[0]) << @intCast(u5, 8 * (i & 3)));
+                ptr += 1;
```
jedisct1, line 31: Maybe use an index rather than a pointer here as well.

lib/std/crypto/ripemd160.zig
```diff
@@ -0,0 +1,218 @@
...
+
+    round(bfn5, -0, 0x50a28be6, permutations[5], shifts[5], &right, words);
+    round(bfn4, -1, 0x5c4dd124, permutations[6], shifts[6], &right, words);
+    round(bfn3, -2, 0x6d703ef3, permutations[7], shifts[7], &right, words);
+    round(bfn2, -3, 0x7a6d76e9, permutations[8], shifts[8], &right, words);
+    round(bfn1, -4, 0x00000000, permutations[9], shifts[9], &right, words);
+
+    right[3] = blk: {
+        var x: u32 = right[3];
+        _ = @addWithOverflow(u32, x, left[2], &x);
```
jedisct1, line 106: No need for `@addWithOverflow` here, you can simply use the `+%` operator.

lib/std/crypto/ripemd160.zig
```diff
@@ -0,0 +1,218 @@
...
+    comptime shifts: [16]u5,
+    hash: *[5]u32,
+    words: [16]u32,
+) void {
+    comptime var r = off;
+    comptime var i: usize = 0;
+    inline while (i < 16) : (i += 1) {
+        hash[@mod(r + 0, 5)] = blk: {
+            var x: u32 = hash[@mod(r + 0, 5)];
+            _ = @addWithOverflow(u32, x, bfn(hash[@mod(r + 1, 5)], hash[@mod(r + 2, 5)], hash[@mod(r + 3, 5)]), &x);
```
jedisct1, line 156: Same here, use `+%`

lib/std/crypto/ripemd160.zig
```diff
@@ -0,0 +1,218 @@
+// https://homes.esat.kuleuven.be/~bosselae/ripemd160.html
+
+const std = @import("std");
+
+pub const Ripemd160 = struct {
```
jedisct1, line 5: RIPEMD is defined for 128, 160, 256 and 320 bit outputs.

So there should be a `Ripemd(comptime out_bits: usize) type` function that accepts the output size, and `Ripemd160` is `Ripemd(160)`. Similar to what was done for, e.g. Blake2.

lib/std/crypto/ripemd160.zig
```diff
@@ -0,0 +1,218 @@
...
+    bytes: [160 / 8]u8,
+
+    pub fn hash(str: []const u8) Ripemd160 {
```
jedisct1, line 8: Input is not assumed to be a string, so `str` may not be a good name for that parameter :)

lib/std/crypto/ripemd160.zig
```diff
@@ -0,0 +1,218 @@
...
+    inline while (i < 16) : (i += 1) {
+        hash[@mod(r + 0, 5)] = blk: {
+            var x: u32 = hash[@mod(r + 0, 5)];
+            _ = @addWithOverflow(u32, x, bfn(hash[@mod(r + 1, 5)], hash[@mod(r + 2, 5)], hash[@mod(r + 3, 5)]), &x);
+            _ = @addWithOverflow(u32, x, words[p[i]], &x);
+            _ = @addWithOverflow(u32, x, constant, &x);
+            break :blk x;
+        };
+
+        hash[@mod(r + 0, 5)] = blk: {
```
jedisct1, line 162: No need for `@mod`, you can use `%`

lib/std/crypto/ripemd160.zig
```diff
@@ -0,0 +1,218 @@
...
+        };
+
+        hash[@mod(r + 2, 5)] = rol(hash[@mod(r + 2, 5)], 10);
+
+        r -= 1;
+    }
+}
+
+/// cyclically rotates x over n bits to the left
+fn rol(x: u32, comptime n: usize) u32 {
```
jedisct1, line 175: There's `math.rotl` already.

hazeycode:
Thanks for taking the time to review @jedisct1!
I was under the impression that RIPEMD-160 specially was not susceptible to the same collision attacks that MD4 and other RIPEMD variants are susceptible to due to. Is this incorrect?
It was for this reason that I opted not to implement a generalized version that supports other size outputs and rather keep to the 160 bit output variant.

lib/std/crypto/ripemd160.zig
hazeycode: @mod instead of the % operator because the lhs is negative and the rhs is comptime - using % instead here yields undefined and is a compile error

jedisct1:
> I was under the impression that RIPEMD-160 specially was not susceptible to the same collision attacks that MD4 and other RIPEMD variants are susceptible to. Is this incorrect?

Oh, you're right, RIPEMD-160 is a tweaked version of the original RIPEMD [with better characteristics](https://eprint.iacr.org/2018/652.pdf).

I'm still a little bit on the fence about whether this is a good fit for the standard library. Since the only modern application using it is generation of Bitcoin addresses, it would be better integrated in a standalone library dedicated to bitcoin addresses generation/parsing/validation.

hazeycode:
> Oh, you're right, RIPEMD-160 is a tweaked version of the original RIPEMD [with better characteristics](https://eprint.iacr.org/2018/652.pdf).
> 
> I'm still a little bit on the fence about whether this is a good fit for the standard library. Since the only modern application using it is generation of Bitcoin addresses, it would be better integrated in a standalone library dedicated to bitcoin addresses generation/parsing/validation.

Thanks for the clarification!

Obviously, I'm of the opinion that it would be convenient to land this in the standard library and I don't see a very compelling reason to exclude it. Most people won't use it but it'd be there for those that do have use for it.

More specifically to the point about instead having a standalone library with functions relating to bitcoin, this is possibly something I will end up putting together at some point. However, my personal preference is to have general functions, not specifically for that purpose, ship with the language. There are possibly existing use cases of RIPEMD-160 that we are not aware of and potentially use cases that are yet to be realized.

jedisct1:
> I don't see a very compelling reason to exclude it. Most people won't use it.

You answered your own question :)

Zig has a rich standard library, and this is great! But functions that are rarely used, and are not required by the compiler itself won't get the same level of attention as the rest of the code.

Parts of `std` have already been [detached](https://github.com/ziglang/std-lib-orphanage), and I guess the intent is to remove more before 1.0, ensuring that what's left is a good subset of what most applications use. The larger a code base is, the most maintenance work it requires, and in practice, the less improvements individual parts get over time.

For `std.crypto` specifically, existing things such as `Gimli` may become external packages before Zig 1.0. Nothing's wrong with Gimli from a security perspective, but as it didn't pass the last round of the NIST LWC competition, and as other options that may eventually become more popular and can fulfill the same role, it will be better maintained as an external package.

An external package can get better integration with its actual use cases, can freely improve its public API anytime (in `std`, once 1.0 is tagged, it's set in stone forever), and can release new versions anytime as well.

And you can be confident that Zig's package manager is going to be awesome :)

jedisct1:
As the RIPEMD-160 document still defines variants with 256 and 320 bit output sizes (security level remains the same, but this is useful for key derivation as 160 bit keys are not very common), it may still be good to define a `Ripemd(comptime out_bits: usize)` function rather restrict it to a single specialized type. Implementation-wise, differences are minimal, and are a perfect fit for `comptime`.

I'd be happy to help you with this (RIPEMD-160 is just a small piece of the puzzle - a full library to handle bitcoin addresses in Zig would be great), but once again, a dedicated project may be a better place that `std`.

hazeycode:
@jedisct1 I've been reading up on elliptic curve cryptography and the p256 source code that you added to the Zig std lib and coming to realize how an implementation can be specialized to better suit the application. I was naive in thinking the generalized p256 code in the std lib would be suitable for implementing optimal operations on secp256k1; would it be correct to say that the code here is more suited to curves with random coefficients?
I've made a start on the project but I'm still very much at the stage of piecing together the puzzle, so to speak, learning cryptography from zero as I go.
Closing this PR with the intention of moving this code to the bitcoin library, when it's further along.

Issue #5853, opened 2020-07-12
rename `std.ChildProcess.exec` to avoid ambiguity with posix exec which replaces current image with new process


tysonclugg:
The [`std.ChildProcess.exec`](https://ziglang.org/documentation/master/std/#std;ChildProcess.exec) function doesn't follow the expected behaviour of *replacing* the executable with the new program being `exec`'ed.  See the SO question "[Differences between fork and exec](https://stackoverflow.com/q/1653340/253599)" for details on how `fork` and `exec` should behave.

Currently, `std.ChildProcess.exec` creates a `std.ChildProcess` instance and then calls its `spawn` method - returning an `ExecResult` struct:
https://github.com/ziglang/zig/blob/f23987db7d83403fc504b63171531e9ad1ffdc1a/lib/std/child_process.zig#L186-L206

I expect that `std.ChildProcess.exec` should not `fork` or `spawn`, and it should have a `!noreturn` result (perhaps `ExecError!noreturn` with `ExecError` being much the same as a `SpawnError`).

tgschultz:
However, coming from a Windows environment, this is not the behavior I expect. Should std.ChildProcess.exec be expected to follow POSIX definitions?

Perhaps instead of making it noreturn, it can simply be renamed to `ChildProcess.spawn` or `ChildProcess.create` to avoid confusion.

daurnimator:
Yeah it might be best if we avoid `exec` at all.

Give that the module is `ChildProcess`, I don't think any "replaceCurrentProcess" doesn't make sense and belongs elsewhere (`process.zig`?).

nektro:
any thoughts on `run`? doing some reworking in the above PR

tysonclugg:
Why not follow POSIX conventions?

Even in Windows, some form of POSIX has been available for the last 29 years (since Windows NT 3.1).

I didn't ask for exec to be renamed.

tysonclugg:
Regarding the POSIX question, I see a number of possible outcomes:
1. Zig closely follows the C/C++ precedent and provides a clean mapping of zig methods to various syscalls, including those defined by POSIX.  Zig provides an exec method that closely follows the POSIX spec (as per the intent of the issue as it was raised).
[...]

PR #10436, opened 2021-12-29
std: expose `ChildProcess`'s `exec` function

nektro:
found this as a cleaner solution than exposing `collectOutputPosix` and `collectOutputWindows` separately.

`spawnAndWait` alone does not solve this use case since it does not collect any output

what this does instead is rename `exec` to `initAndExec` and makes `exec` a standalone method to get the term and output

edit: feel free to squash commit this

### DIFF
--- a/doc/docgen.zig
+++ b/doc/docgen.zig
@@ -1326,7 +1326,7 @@ fn genHtml(
                         try shell_out.print("\n", .{});
 
                         if (expected_outcome == .BuildFail) {
-                            const result = try ChildProcess.exec(.{
+                            const result = try ChildProcess.initAndExec(.{
                                 .allocator = allocator,
                                 .argv = build_args.items,
                                 .env_map = &env_map,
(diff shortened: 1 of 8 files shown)


daurnimator:
If we're renaming this method, can we please move away from "exec" as a name? It suggests that the current process will be replaced.

See https://github.com/ziglang/zig/issues/5853

nektro:
what name should be used instead? i dont see any consensus.

fwiw i also disagree with the premise of the issue. exec is simply short for execute. perhaps run would be better?

daurnimator:
> what name should be used instead?

I think `spawn` seems the best suggestion in there; and it has precedence via e.g. `posix_spawn` in C.

nektro:
`spawn` is already a method and does something different

andrewrk (changes requested):
I agree with @daurnimator - this is a breaking change to rename a function which has an accepted proposal to avoid "exec" as the name, so the breaking change should follow this guidance.

Also, this is a breaking change that I'm not sure what the motivation is. Can I see an example of what didn't work very well without this breaking change?

PR #10446, opened 2021-12-29
stage2: more fixes to x86_64 codegen, mir and isel

kubkon:
* add separate tag for MI encoding - to request memory-immediate encoding at the MIR side, we are now required to use a new tag such as `mov_mem_imm` where the size of the memory pointer operand is encoded using the flags:

```zig
0b00 => .byte_ptr,
0b01 => .word_ptr,
0b10 => .dword_ptr,
0b11 => .qword_ptr,
```

* fix handling of `ah`, `bh`, `ch`, and `dh` registers (which are actually used as aliases to `dil`, etc. registers). Currenly, we treat them as aliases only meaning when we encounter `ah` we make sure to set the REX.W to promote the instruction to 64bits and use `dil` register instead - otherwise we might have mismatch between registers used in different parts of the codegen. In the future, we can and should use `ah`, etc. as upper 8bit halves of 16bit registers `ax`, etc.
* fix bug in `airCmp` where `.cmp` MIR instruction shouldn't force type `Bool` but let the type of the original type propagate downwards - we need this to make an informed choice of the target register size and hence choose the right encoding down the line.
* implement lowering of 1-byte and 2-byte values to stack and add matching stage2 tests for x86_64 codegen
* update `PrintMir` with latest instructions and Isel changes

### DIFF
--- a/src/arch/x86_64/CodeGen.zig
+++ b/src/arch/x86_64/CodeGen.zig
@@ -1646,7 +1646,7 @@ fn genBinMathOpMir(
                     _ = try self.addInst(.{
                         .tag = mir_tag,
                         .ops = (Mir.Ops{
-                            .reg1 = registerAlias(dst_reg, 4),
+                            .reg1 = dst_reg.to32(),
                         }).encode(),
                         .data = .{ .imm = @intCast(i32, imm) },
                     });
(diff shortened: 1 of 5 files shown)


andrewrk (approved):
Looks great :rocket:

PR #10455, opened 2021-12-30
stage2 ARM: basic slice + basic struct support

joachimschmidt557:
Further progress towards `zig test`

### DIFF
--- a/src/arch/arm/CodeGen.zig
+++ b/src/arch/arm/CodeGen.zig
@@ -1202,7 +1202,6 @@ fn airSliceElemVal(self: *Self, inst: Air.Inst.Index) !void {
     const bin_op = self.air.instructions.items(.data)[inst].bin_op;
     const result: MCValue = if (!is_volatile and self.liveness.isUnused(inst)) .dead else result: {
         const slice_mcv = try self.resolveInst(bin_op.lhs);
-        const index_mcv = try self.resolveInst(bin_op.rhs);
 
         const slice_ty = self.air.typeOf(bin_op.lhs);
         const elem_ty = slice_ty.childType();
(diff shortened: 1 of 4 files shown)


src/arch/arm/CodeGen.zig
```diff
@@ -1435,7 +1434,46 @@ fn load(self: *Self, dst_mcv: MCValue, ptr: MCValue, ptr_ty: Type) InnerError!vo
...
+                            .cond = .al,
+                            .data = .{ .rr_op = .{
+                                .rd = len_reg,
+                                .rn = .r0,
+                                .op = len_op,
+                            } },
+                        });
+
+                        // memcpy(src, dst, len)
+                        try self.genArmInlineMemcpy(src_reg, dst_reg, len_reg, count_reg, tmp_reg);
```
andrewrk, line 1476: Keep in mind that we also have the ability to call an external `memcpy` function if it makes sense to. It will be provided by libc, or in freestanding mode, zig provides a "freestanding libc" which has `memcpy` in it. Related: #7265

src/codegen.zig
```diff
@@ -372,11 +372,32 @@ pub fn generateSymbol(
             return Result{ .appended = {} };
         },
         .Struct => {
+            // TODO debug info
+            // TODO padding of struct members
```
andrewrk, line 376: I'll help with this in a future patch :)

andrewrk (approved):

src/arch/arm/CodeGen.zig
joachimschmidt557: Yes, in future, we should definitely do that. However, currently, extern function calls aren't supported by the Elf linking backend, so this is the best I could do :laughing: . Also (this was @kubkon's idea) a potential for optimization is analyzing how often memcpy is needed and if e.g. memcpy is needed only once, it could be inlined, else emit calls to the extern `memset`.

src/codegen.zig
joachimschmidt557: Thanks! It shouldn't be too hard as `Type.structFieldOffset()` already provides a lot of relevant information I think.

PR #10456, opened 2021-12-30
Support parsing tz timezone data

Aransentin:
I added support for parsing tz timezone database files. This will be useful for future date support in the standard library.

Some caveats:
~~It does not support the legacy pre-2005 format. That one stored timestamps as 32-bit signed integers, and as such will stop working completely in 2038. It does support "thick" tz files that have both legacy and modern data in the same file, by just skipping over the legacy cruft.~~

I prepared the leap second data structure to fit into 8 bytes (instead of 16, with padding), which strictly speaking violates the spec but will never matter in practice; it can still represent leap seconds nine million years in the future.

The parser rejects all "MUST NOT" from rfc8536, and one SHOULD NOT - timezone type names (e.g. EST, CET...) cannot be more than 6 characters, which is banned by POSIX anyway.


lib/std/tz.zig
```diff
@@ -0,0 +1,218 @@
...
+    footer: []const u8,
+
+    pub fn parse(allocator: std.mem.Allocator, data: []const u8) !Tz {
+        const header_size = 4 + 1 + 15 + 6 * 4;
+        if (data.len < header_size) return error.BadSize;
+
+        const magic_l = data[0..4];
+        const version_l = data[4];
+        if (!std.mem.eql(u8, magic_l, "TZif")) return error.BadHeader;
+        if (version_l != '2' and version_l != '3') return error.BadVersion;
```
daurnimator, line 50: `version_l == 0` also has this format up to the gmt flag at least.

lib/std/tz.zig
```diff
@@ -0,0 +1,218 @@
...
+};
+
+pub const Tz = struct {
+    allocator: std.mem.Allocator,
+    transitions: []const Transition,
+    timetypes: []const Timetype,
+    leapseconds: []const Leapsecond,
+    footer: []const u8,
+
+    pub fn parse(allocator: std.mem.Allocator, data: []const u8) !Tz {
```
daurnimator, line 43: Rather than taking an array of bytes, I think this could take a reader?

lib/std/tz.zig
```diff
@@ -0,0 +1,218 @@
...
+        if (data.len < header_size + data_block_size_legacy) return error.BadSize;
+
+        const data2 = data[header_size + data_block_size_legacy ..];
+        if (data2.len < header_size) return error.BadSize;
+
+        const magic = data2[0..4];
+        const version = data2[4];
+        if (!std.mem.eql(u8, magic, "TZif")) return error.BadHeader;
+        if (version != '2' and version != '3') return error.BadVersion;
+
```
daurnimator, line 69: Should verify that it is followed by 15 nulls

lib/std/tz.zig
```diff
@@ -0,0 +1,218 @@
...
+        self.allocator.free(self.footer);
+        self.allocator.free(self.leapseconds);
+        self.allocator.free(self.transitions);
+        self.allocator.free(self.timetypes);
+    }
+};
+
+test "parse" {
+    // Asia/Tokyo is good for embedding, as Japan only had DST for a short while during the US occupation
+    const data = @embedFile("tz/asia_tokyo.tzif");
```
daurnimator, line 210: Try and add a test file for each version. with and without legacy sections.

FWIW `America/Godthab` should be a small file.

daurnimator (changes requested):
Make sure you add .tzif to `exclude_extensions` in build.zig


FYI I wrote a tzfile parser in lua over at https://github.com/daurnimator/luatz/blob/master/luatz/tzfile.lua
I'd like to think its pretty easy to read.

lib/std/tz.zig
Aransentin: Done. Though Godthab isn't *that* small, so I added a fat version of Antarctica/Davis as a second test instead.

lib/std/tz.zig
Aransentin: RFC8536 doesn't actually say that they should be null, just that they are reserved.

It's even implied that you shouldn't care about them - the spec depends on that legacy parsers that doesn't understand the "version" field not bailing when it sees a '2' or '3' in that byte, otherwise the "fat" version of the files wouldn't work.

lib/std/tz.zig
Aransentin: Fix'd

lib/std/tz.zig
Aransentin: Not sure what you are trying to say. I don't support version 0 at all and just bail as soon as possible if I see that.

Aransentin:
The CI failure isn't my fault, I don't think. Testing "cmpxchg on a global variable" has nothing to do with this, and similar errors are happening on other PRs.

lib/std/tz.zig
```diff
@@ -0,0 +1,233 @@
...
+        };
+
+        const Counts = extern struct {
+            isutcnt: u32,
+            isstdcnt: u32,
+            leapcnt: u32,
+            timecnt: u32,
+            typecnt: u32,
+            charcnt: u32,
+        };
```
daurnimator, line 57: These might be useful to hoist out of the function and have in the `struct`.

lib/std/tz.zig
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        // Parse standard/wall indicators
+        i = 0;
+        while (i < counts.isstdcnt) : (i += 1) {
+            const stdtime = try reader.readByte();
+            if (stdtime == 1) {
+                timetypes[i].flags |= 0x02;
+            }
+        }
```
daurnimator, line 169:
```suggestion
        for (timetypes) |*tt| {
            const stdtime = try reader.readByte();
            if (stdtime == 1) {
                tt.* |= 0x02;
            }
        }
```

lib/std/tz.zig
```diff
@@ -0,0 +1,233 @@
...
+const builtin = @import("builtin");
+
+pub const Transition = struct {
+    ts: i64,
+    timetype: *Timetype,
+};
+
+pub const Timetype = struct {
+    offset: i32,
+    flags: u8,
```
daurnimator, line 11: Why have this as `flags` instead of `bool` fields for each?

lib/std/tz.zig
```diff
@@ -0,0 +1,233 @@
...
+            timecnt: u32,
+            typecnt: u32,
+            charcnt: u32,
+        };
+
+        // Parse and skip the legacy header and data
+        {
+            const header = try reader.readStruct(Header);
+            if (!std.mem.eql(u8, &header.magic, "TZif")) return error.BadHeader;
+            if (header.version == 0) return error.UnsupportedLegacyFormat;
```
daurnimator, line 63: Why not support version 0? it's quite trivial and IIRC I ran into it on a few systems.

lib/std/tz.zig
Aransentin: Not sure they would be very useful for anything other than the function itself. I mostly added them as an optimization for the `reader`, so that it could fetch multiple values at the same time to reduce all the error branching.

lib/std/tz.zig
Aransentin: `counts.isstdcnt` is allowed to be zero, which would mess this up.

lib/std/tz.zig
daurnimator: Yeah the only use-case I could think of was a tzfile writer.

lib/std/tz.zig
Aransentin: It makes `Timetype` pack better; 12 bytes instead of 16. I can change it if the optimization isn't worth the complexity of having getters for each value.

lib/std/tz.zig
Aransentin: I imagined that it would cause an entire new class of possible errors for the future date library, e.g. "You've tried to process a time after 2038" failures that need to be taken into account.

If there actually are systems that only use the legacy format then I guess we should support it as well though; at least until 2038...

Aransentin: Added legacy file support.

lib/std/tz.zig
Aransentin: Done.

daurnimator (approved):
LGTM.

Though without algorithms that *use* the data it's all a bit useless right?
I assume you aim to have a followup PR in mind to make use of this code?

Aransentin:
> Though without algorithms that _use_ the data it's all a bit useless right? I assume you aim to have a followup PR in mind to make use of this code?

I want to write some Date/Calendar utilities, yeah. Even if that doesn't go anywhere splitting up the task into smaller subtasks (like the timezone parsing here) means some of the work will already be done for the next person who picks up the torch.

PR #10464, opened 2021-12-31
Report more errors upfront when possible

Jarred-Sumner:
Some errors do not need to halt ast check eagerly. In those cases, it's a better developer experience to show more errors upfront rather than wait for previous errors to be fixed before showing more.

This applies to the following errors:
- `"unused for loop label"`
- `"unused while loop label"`
- `enum declarations must have at least one tag`
- `non-exhaustive enum missing integer tag type` (is this one safe?)
- `'comptime const' is redundant; instead wrap the initialization expression with 'comptime'`
- `"unused {s}"`
- `unused block label`
- `redundant nosuspend block`
- `unreachable code` (there is a case to be made that reporting errors in unreachable code is unnecessary)

| Before (37) | After (69) |
:-------------------------:|:-------------------------:
 |

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -319,7 +319,7 @@ fn reachableExpr(
 ) InnerError!Zir.Inst.Ref {
     const result_inst = try expr(gz, scope, rl, node);
     if (gz.refIsNoReturn(result_inst)) {
-        return gz.astgen.failNodeNotes(reachable_node, "unreachable code", .{}, &[_]u32{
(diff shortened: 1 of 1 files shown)


bradms:
I think this is super useful and cuts down on the "edit-compile-run" cycle by showing even more errors the first time around.

andrewrk (approved):
Nice!

Vexu:
So has the consensus on redundant compile errors changed since #9377?

PR #10486, opened 2022-01-02
std: Implement cross-platform metadata API

### LINKED ISSUE #10441: Implement a better file metadata API
Currently, as highlighted in #6600, `std.fs` is very posix-based, which is harmful to cross-platform functionality. To improve the situation, I'm working on abstracting `stat` and `chmod` (maybe `chown` too) to a lowest-common-denominator, with the option to use more fine-grained platform-specific functionality if desired.

I'm making this issue so that nobody else works on this at the same time as myself, and wastes my or their time :)

ominitay:
This PR implements a cross-platform metadata API, aiming to reduce unnecessary Unix-dependence of the Zig `std.fs` api. Presently, all OSes beside Windows are treated as Unix; I think that this is the best way to treat things by default, instead of explicitly listing each Posix-compliant OS.

Adds:

- File.setPermissions() : Sets permission of a file according to a `Permissions` struct (not available on WASI)

- File.Metadata : A cross-platform representation of file metadata
  - Metadata.size() : Returns the size of a file
  - Metadata.permissions() : Returns a `Permissions` struct, representing permissions on the file
  - Metadata.kind() : Returns the `Kind` of the file
  - Metadata.time() : Returns the time the file was accessed, modified, or created

- File.Permissions : A cross-platform representation of file permissions
  - Permissions.readOnly() : Returns whether the file is read-only
  - Permissions.setReadOnly() : Sets whether the file is read-only
  - Permissions.unixSet() : Sets permissions for a class (UNIX-only)
  - Permissions.unixGet() : Checks a permission for a class (UNIX-only)
  - Permissions.unixNew() : Returns a new `Permissions` struct representing the passed mode (UNIX-only)

I don't quite think this is complete yet, and would like some feedback on this. Additionally, I haven't been able to test whether this works on Windows (it compiles, but not sure if tests pass), as Wine doesn't seem to want to work. If anyone could run tests on a Windows machine for me, it would be much appreciated.

closes #10441 
cc #6600

### DIFF
--- a/lib/std/fs/file.zig
+++ b/lib/std/fs/file.zig
@@ -398,6 +398,485 @@ pub const File = struct {
         try os.fchown(self.handle, owner, group);
     }
 
+    /// Cross-platform representation of permissions on a file.
+    /// The `readonly` and `setReadonly` are the only methods available across all platforms.
+    /// Platform-specific functionality is available through the `inner` field.
+    pub const Permissions = struct {
+        /// You may use the `inner` field to use platform-specific functionality
+        inner: switch (builtin.os.tag) {
+            .windows => PermissionsWindows,
+            else => PermissionsUnix,
+        },
+
+        const Self = @This();
+
+        /// Returns `true` if permissions represent an unwritable file.
+        /// On Unix, `true` is returned only if no class has write permissions.
+        pub fn readOnly(self: Self) bool {
+            return self.inner.readOnly();
+        }
+
+        /// Sets whether write permissions are provided.
+        /// On Unix, this affects *all* classes. If this is undesired, use `unixSet`
+        /// This method *DOES NOT* set permissions on the filesystem: use `File.setPermissions(permissions)`
+        pub fn setReadOnly(self: *Self, read_only: bool) void {
+            self.inner.setReadOnly(read_only);
+        }
+    };
+
+    pub const PermissionsWindows = struct {
+        attributes: os.windows.DWORD,
+
+        const Self = @This();
+
+        /// Returns `true` if permissions represent an unwritable file.
+        pub fn readOnly(self: Self) bool {
+            return self.attributes & os.windows.FILE_ATTRIBUTE_READONLY != 0;
+        }
+
+        /// Sets whether write permissions are provided.
+        /// This method *DOES NOT* set permissions on the filesystem: use `File.setPermissions(permissions)`
+        pub fn setReadOnly(self: *Self, read_only: bool) void {
+            if (read_only) {
+                self.attributes |= os.windows.FILE_ATTRIBUTE_READONLY;
+            } else {
+                self.attributes &= ~@as(os.windows.DWORD, os.windows.FILE_ATTRIBUTE_READONLY);
+            }
+        }
+    };
+
+    pub const PermissionsUnix = struct {
+        mode: Mode,
+
+        const Self = @This();
+
+        /// Returns `true` if permissions represent an unwritable file.
+        /// `true` is returned only if no class has write permissions.
+        pub fn readOnly(self: Self) bool {
+            return self.mode & 0o222 == 0;
+        }
+
+        /// Sets whether write permissions are provided.
+        /// This affects *all* classes. If this is undesired, use `unixSet`
+        /// This method *DOES NOT* set permissions on the filesystem: use `File.setPermissions(permissions)`
+        pub fn setReadOnly(self: *Self, read_only: bool) void {
+            if (read_only) {
+                self.mode &= ~@as(Mode, 0o222);
+            } else {
+                self.mode |= @as(Mode, 0o222);
+            }
+        }
+
+        pub const Class = enum(u2) {
+            user = 2,
+            group = 1,
+            other = 0,
+        };
+
+        pub const Permission = enum(u3) {
+            read = 0o4,
+            write = 0o2,
+            execute = 0o1,
+        };
+
+        /// Returns `true` if the chosen class has the selected permission.
+        /// This method is only available on Unix platforms.
+        pub fn unixHas(self: Self, class: Class, permission: Permission) bool {
+            const mask = @as(Mode, @enumToInt(permission)) << @as(u3, @enumToInt(class)) * 3;
+            return self.mode & mask != 0;
+        }
+
+        /// Sets the permissions for the chosen class. Any permissions set to `null` are left unchanged.
+        /// This method *DOES NOT* set permissions on the filesystem: use `File.setPermissions(permissions)`
+        pub fn unixSet(self: *Self, class: Class, permissions: struct {
+            read: ?bool = null,
+            write: ?bool = null,
+            execute: ?bool = null,
+        }) void {
+            const shift = @as(u3, @enumToInt(class)) * 3;
+            if (permissions.read) |r| {
+                if (r) {
+                    self.mode |= @as(Mode, 0o4) << shift;
+                } else {
+                    self.mode &= ~(@as(Mode, 0o4) << shift);
+                }
+            }
+            if (permissions.write) |w| {
+                if (w) {
+                    self.mode |= @as(Mode, 0o2) << shift;
+                } else {
+                    self.mode &= ~(@as(Mode, 0o2) << shift);
+                }
+            }
+            if (permissions.execute) |x| {
+                if (x) {
+                    self.mode |= @as(Mode, 0o1) << shift;
+                } else {
+                    self.mode &= ~(@as(Mode, 0o1) << shift);
+                }
+            }
+        }
+
+        /// Returns a `Permissions` struct representing the permissions from the passed mode.
+        pub fn unixNew(new_mode: Mode) Self {
+            return Self{
+                .mode = new_mode,
+            };
+        }
+    };
+
+    pub const SetPermissionsError = ChmodError;
+
+    /// Sets permissions according to the provided `Permissions` struct.
+    /// This method is *NOT* available on WASI
+    pub fn setPermissions(self: File, permissions: Permissions) SetPermissionsError!void {
+        switch (builtin.os.tag) {
+            .windows => {
+                var io_status_block: windows.IO_STATUS_BLOCK = undefined;
+                var info = windows.FILE_BASIC_INFORMATION{
+                    .CreationTime = 0,
+                    .LastAccessTime = 0,
+                    .LastWriteTime = 0,
+                    .ChangeTime = 0,
+                    .FileAttributes = permissions.inner.attributes,
+                };
+                const rc = windows.ntdll.NtSetInformationFile(
+                    self.handle,
+                    &io_status_block,
+                    &info,
+                    @sizeOf(windows.FILE_BASIC_INFORMATION),
+                    .FileBasicInformation,
+                );
+                switch (rc) {
+                    .SUCCESS => return,
+                    .INVALID_HANDLE => unreachable,
+                    .ACCESS_DENIED => return error.AccessDenied,
+                    else => return windows.unexpectedStatus(rc),
+                }
+            },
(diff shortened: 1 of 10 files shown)


ominitay:
This is **NOT** ready to be merged!!

ominitay:
WASI isn't supported because of a lack of `chmod`. Is this something that could be implemented on WASI?

andreafeletto:
Regarding the {a,m,c}time functions, I would suggest merging them into one function (maybe called time) which would accept an enum of value .accessed, .modified or .created.

ominitay:
> Regarding the {a,m,c}time functions, I would suggest merging them into one function (maybe called time) which would accept an enum of value .accessed, .modified or .created.

I like that idea! I'll get that done in a bit.

ominitay:
Don't understand why Windows is throwing an `AccessDenied` error...

lib/std/fs/file.zig
```diff
@@ -385,6 +385,274 @@ pub const File = struct {
...
+    /// Unix-only functionality is available through the `unixHas` and `unixSet` methods.
+    pub const Permissions = switch (builtin.os.tag) {
+        .windows => struct {
+            attributes: os.windows.DWORD,
+
+            const Self = @This();
+
+            /// Returns `true` if permissions represent an unwritable file.
+            /// On Unix, `true` is returned only if no class has write permissions.
+            pub fn readonly(self: Self) bool {
```
squeek502, line 399: Nitpick: I think these functions would be better capitalized as `readOnly` / `setReadOnly`

lib/std/fs/test.zig
```diff
@@ -1068,3 +1068,72 @@ test "chown" {
...
+    try testing.expect(!permissions.readonly());
+    permissions.setReadonly(true);
+    try testing.expect(permissions.readonly());
+
+    try file.setPermissions(permissions);
+
+    const new_metadata = try file.metadata();
+    const new_permissions = new_metadata.permissions();
+
+    try testing.expect(new_permissions.readonly());
```
squeek502, line 1107:
```suggestion
    try testing.expect(new_permissions.readonly());

    // Must set it back to non-read-only in order to delete it
    permissions.setReadonly(false);
    try file.setPermissions(permissions);
```

This will fix the test on Windows (it was running into an unexpected `CANNOT_DELETE` status during the `defer tmp.cleanup()`). I'll make a separate PR to fix the unexpected status part of this.

ominitay: Thanks for spotting this :)

lib/std/fs/file.zig
ominitay: I agree

ominitay:
Passing tests on Windows now!! 🎉

lib/std/fs/file.zig
```diff
@@ -385,6 +385,274 @@ pub const File = struct {
...
+            /// This method is only available on Unix platforms.
+            pub fn unixNew(new_mode: Mode) Self {
+                return Self{
+                    .mode = new_mode,
+                };
+            }
+        },
+    };
+
+    pub const SetPermissionsError = if (is_windows) error{AccessDenied} || os.UnexpectedError else ChmodError;
```
Vexu, line 497:
```suggestion
    pub const SetPermissionsError = ChmodError;
```
The error set should contain all possible errors for all possible targets.

ominitay: Not sure why you're suggesting to remove the possible errors for Windows...

ominitay: Oh, I see now. `FChmodError`, by coincidence, contains the possible errors thrown on Windows. Is this a sensible thing to assume though?

ominitay: Actually that makes sense, I'm just way too tired to be working on this

ominitay:
Soooo `ctime` on Unix and `ctime` on Windows mean different things. Time to redo a lot of things :)

ominitay:
> Regarding the {a,m,c}time functions, I would suggest merging them into one function (maybe called time) which would accept an enum of value .accessed, .modified or .created.

This won't be happening now, sorry :(

lib/std/fs/file.zig
```diff
@@ -385,6 +385,274 @@ pub const File = struct {
...
+            pub fn modified(self: Self) i128 {
+                return self.stat.mtime;
+            }
+        },
+    };
+
+    pub const MetadataError = StatError;
+
+    /// Returns a `Metadata` struct, representing the permissions on the file
+    pub fn metadata(self: File) MetadataError!Metadata {
```
marler8997, line 616: How about `getMetadata`?  I think Zig's std convention is usually to use verbs for getting/setting things?

lib/std/fs/file.zig
```diff
@@ -385,6 +385,274 @@ pub const File = struct {
...
+                try self.chmod(permissions.mode);
+            },
+        }
+    }
+
+    /// Cross-platform representation of file metadata.
+    /// Can be obtained with `File.metadata()`
+    pub const Metadata = switch (builtin.os.tag) {
+        .windows => struct {
+            attributes: windows.DWORD,
```
marler8997, line 537: I wonder if we should prefix all platform specific fields with a platform identifier?  So in this case `windowsAttributes`?  The reason I suggest this is to make it obvious what parts of the API are platform agnostic and which aren't.   For example,
```zig
const m = try file.metadata();
std.log.info("file attributes={d}", .{m.attributes});
```
This code snippet looks like it's platform agnostic but it's not.  If the developer who wrote this only compiles it for Windows then they won't know the mistake.  This would become much more obvious if it looked like this:

```zig
const m = try file.metadata();
std.log.info("file attributes={d}", .{m.windowsAttributes});
```

Now it's alot more obvious that this field is specific to Windows and if they want their code to work elsewhere to surround it with a `builtin.os.tag` check.  We also use this strategy for platform specific functions like you did with `unixHas` etc.

ominitay: I'm starting to rethink a lot of this really, and one idea I keep coming back do is having `File.Metadata`, which consistently has the same functions, and has a field `inner`, which contains the actual platform-specific implementation (`MetadataLinux` etc), and has methods which aren't shared across platforms. This would help discourage direct use of member fields as well as prefixing them with the platform name, and additionally keep documentation for `File.Metadata` exclusively to cross-platform methods.

lib/std/fs/file.zig
ominitay: When discussing this in the Discord server, @ifreund said to me the opposite -- that I should drop the 'get' prefix which I was using previously. That matches with `.allocator()` too.

lib/std/fs/file.zig
marler8997: Yeah that's an improvement IMO.

ominitay:
Have been doing some research in the Discord server regarding how OSes expose creation time to us. So far, I have found:

When there is no creation time (i.e. the filesystem does not support it):
- Linux zeroes a mask bit with statx
  - This is nice -- we can easily detect this, and don't lose any otherwise valid times.
- FreeBSD sets birthtime seconds to -1, and nsecs to 0
- NetBSD sets birthtime seconds to 0, and nsecs to 0
- OpenBSD sets birthtime seconds to 0, and nsecs to 0
  - The *BSDs are less nice -- we can easily detect this, but we lose otherwise valid times.
- MacOS sets birthtime to ctime
  - This is awful -- we can't detect this.

Obtaining the creation time:
- Linux goes its own way with the added `statx` syscall, but is fairly nice to use, so I'm not upset.
- Other Unices either have a field in their `stat` structs, or don't provide it at all. This is not so good because of the above point. Additionally, they may be named differently, so I intend to change the declarations to fix this.
- Windows is Windows, and has its own way, putting CreationTime in BasicFileInfo.

ominitay:
This means we have two options for `Metadata.created()`:

Firstly, as it has to return an optional anyway (we can't know whether the filesystem supports it or not until runtime), we could return null if the target doesn't have `Stat.btim`. This would have the disadvantage of leaving this to be dealt with at compile-time to be dealt with at runtime.

On the other hand, as we know for certain whether the target has `Stat.btim` or not at compile-time, we could instead throw a compile error. This would have the disadvantage of meaning that we can't do `Metadata.created()`, only `Metadata.inner.created()`. 

My personal preference is for the first, as I think it makes the API more flexible, but would like to see what you think :)

lib/std/fs/file.zig
```diff
@@ -385,6 +385,460 @@ pub const File = struct {
...
+            if (!@hasDecl(@TypeOf(self.stat), "birthtime")) return null;
+            const birthtime = self.stat.birthtime();
+
+            // If the filesystem doesn't support this the value *should* be:
+            // On FreeBSD: tv_nsec = 0, tv_sec = -1
+            // On NetBSD and OpenBSD: tv_nsec = 0, tv_sec = 0
+            // On MacOS, it is set to ctime -- we cannot detect this!!
+            // Additionally, we don't know how filesystems themselves may deal with this -- they may
+            // in-fact return a different value, so this check is fairly permissive.
+            if (birthtime.tv_sec == -1 or birthtime.tv_nsec <= -1 or (birthtime.tv_sec == 0 and birthtime.tv_nsec == 0)) // wtf i hate this
```
ominitay, line 649: I anticipate that people will want to examine this line in particular.

lib/std/fs/file.zig
ominitay: This should now be resolved by the pushed changes, though I haven't yet added any platform-specific operations :)

ominitay:
If someone is still running it, could someone test this on a machine with a kernel earlier than 4.11?

squeek502:
> If someone is still running it, could someone test this on a machine with a kernel earlier than 4.11?

I downloaded the latest master tarball of zig, replaced `lib/std/fs.zig`/`lib/std/fs/file.zig`/`lib/std/fs/test.zig` with the versions from this branch, added a `debug.print` call to the metadata test to print out the accessed/modified/created times and got:

```
$ uname -r
4.4.0-210-generic
$ ./zig test lib/std/fs/test.zig
Test [36/39] test "File.metadata"... accessed: 1641697197363780756 modified: 1641697197363780756 created: null
All 39 tests passed.
```

Let me know if there's something more specific that you're looking to check.

ominitay:
That looks good! Thank you :)

lib/std/fs/file.zig
Vexu: I'd prefer this check to broken up per OS as in the comment.

Vexu (approved):

lib/std/fs/file.zig
```diff
@@ -385,6 +385,462 @@ pub const File = struct {
         try os.fchown(self.handle, owner, group);
     }
 
+    /// Cross-platform representation of permissions on a file.
+    /// The `readonly` and `setReadonly` are the only methods available across all platforms.
+    /// Unix-only functionality is available through the `unixHas` and `unixSet` methods.
+    pub const Permissions = switch (builtin.os.tag) {
```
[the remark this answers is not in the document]
ominitay: Agreed. I'll do that :)

ominitay:
> I'd prefer this check to broken up per OS as in the comment.

Not sure why I can't reply to this in a conversation. My thought process was to make this check fairly loose, so that if the filesystem is being weird and setting this value to a different (also invalid) value, we'd still deal with it. Though I think you are right about this the further I've thought about it.

ominitay:
I think it's best to compile error if a switch branch isn't specified though, in order to force future contributors to update this properly.

ominitay:
So I can't find what Haiku sets it to by default. This PR would still be mergeable without a branch for it, but means that Haiku users would encounter a compile error when trying to use `Metadata.created()`.

ominitay:
Solved all that, notwithstanding the Haiku question. Should be ready to merge.

PR #10488, opened 2022-01-02
plan9: fix .z symbol in debuginfo

g-w1:
This allows the `acid` debugger on
plan9 to be used to debug a zig source
file without patching `acid`!

The patch adds a second `z` symbol. This z
symbol has a value of 0, which means that it
pops the history stack. We put a very large
number for the value of the second symbol because
it has to be at least as large as the linecount of
the file. The debuginfo format is meant to be used
with c files, where the stack would look something
like this:
```
->  Line: 0x1 (1)  Name: 0x1/0x2/0x3/0xe/0x13/0x1b (/sys/src/libc/port/malloc.c)
->  Line: 0x2 (2)  Name: 0x1/0x6/0x7/0x8 (/amd64/include/u.h)
->  Line: 0x4f (79)  Name:  ()
->  Line: 0x50 (80)  Name: 0x1/0x2/0x7/0x9 (/sys/include/libc.h)
->  Line: 0x358 (856)  Name:  ()
->  Line: 0x359 (857)  Name: 0x1/0x2/0x7/0x1c (/sys/include/pool.h)
->  Line: 0x392 (914)  Name:  ()
->  Line: 0x393 (915)  Name: 0x1/0x2/0x7/0x1d (/sys/include/tos.h)
->  Line: 0x3ab (939)  Name:  ()
->  Line: 0x4eb (1259)  Name:  ()
```
however in zig, we do not use includes and .h files,
so we only need the first and last items in the stack:
the source file that the symbols belong to, and the pop
symbol with a null name and a value of the total linecount of the
preprocessed source. Since there is no preprocessing in zig, we
just make the linecount very large. There do not appear to be
any downsides to this approach. If this causes a bug in the future,
a simple fix would be to make the pop symbol just have the value
of how many newlines are in the source file.

### DIFF
--- a/src/link/Plan9.zig
+++ b/src/link/Plan9.zig
@@ -173,6 +173,7 @@ fn putFn(self: *Plan9, decl: *Module.Decl, out: FnDeclOutput) !void {
         fn_map_res.value_ptr.* = .{
             .sym_index = blk: {
                 try self.syms.append(gpa, undefined);
+                try self.syms.append(gpa, undefined);
                 break :blk @intCast(u32, self.syms.items.len - 1);
             },
         };
@@ -194,10 +195,16 @@ fn putFn(self: *Plan9, decl: *Module.Decl, out: FnDeclOutput) !void {
         // null terminate
         try a.append(0);
         const final = a.toOwnedSlice();
-        self.syms.items[fn_map_res.value_ptr.sym_index] = .{
+        self.syms.items[fn_map_res.value_ptr.sym_index - 1] = .{
             .type = .z,
             .value = 1,
             .name = final,
+        };
+        self.syms.items[fn_map_res.value_ptr.sym_index] = .{
+            .type = .z,
+            // just put a giant number, no source file will have this many newlines
+            .value = std.math.maxInt(u32),
+            .name = &.{ 0, 0 },
         };
     }
 }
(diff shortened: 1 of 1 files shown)


g-w1:
Next step: get x86_64 backend working!

kubkon:
> Next step: get x86_64 backend working!

Just an FYI that I am currently halfway through adding `zig test` support for x86_64 and you should expect conflicts for x86_64 codegen and isel if you make any changes. Other than that, great work!

src/link/Plan9.zig
```diff
@@ -172,6 +172,7 @@ fn putFn(self: *Plan9, decl: *Module.Decl, out: FnDeclOutput) !void {
         // each file gets a symbol
         fn_map_res.value_ptr.* = .{
             .sym_index = blk: {
+                try self.syms.append(gpa, undefined);
```
kubkon, line 175: Is this intentional that you append `undefined` twice in a row to the same container?

g-w1:
> > Next step: get x86_64 backend working!
> 
> Just an FYI that I am currently halfway through adding `zig test` support for x86_64 and you should expect conflicts for x86_64 codegen and isel if you make any changes. Other than that, great work!

Oh, I just meant that in the future that is what I will be trying to do, this pr is done. :)

src/link/Plan9.zig
g-w1: Yes, they both are filled in with .z symbols later.
The first one has the name of the file, while the second one has a null name for pop off the stack.

PR #10494, opened 2022-01-03
CBE: Add basic slice support

mshockwave:
Details:
  - Add codegen support for slicing of element pointers.
  - Basic Type.hash for .Pointer types.
  - Fix incorrect codegen on global array variables. Nevertheless, the initialization for such variables is still buggy.
  - Zero-length global array will not be codegen-ed by stage2, thus updating related test cases to use sentinel array instead.
  - Use HashMap to avoid duplicated typedef bodies.

Now test/behavior/slice.zig (as well as other existing tests) should pass.
I bumped into the problem mentioned in the third bullet point, I can file a separate issue for that.

### DIFF
--- a/src/type.zig
+++ b/src/type.zig
@@ -694,8 +694,8 @@ pub const Type = extern union {
             => {}, // The zig type tag is all that is needed to distinguish.
 
             .Pointer => {
+                // Basic hashing for simple types.
                 // TODO implement more pointer type hashing
-                // Basic hashing for small types.
                 std.hash.autoHash(hasher, self.tag());
             },
             .Int => {
--- a/src/type.zig
+++ b/src/type.zig
@@ -695,6 +695,8 @@ pub const Type = extern union {
 
             .Pointer => {
                 // TODO implement more pointer type hashing
+                // Basic hashing for small types.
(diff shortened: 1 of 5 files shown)


src/type.zig
```diff
@@ -694,7 +694,9 @@ pub const Type = extern union {
             => {}, // The zig type tag is all that is needed to distinguish.
 
             .Pointer => {
+                // Basic hashing for simple types.
                 // TODO implement more pointer type hashing
+                std.hash.autoHash(hasher, self.tag());
```
andrewrk, line 699: This would unfortunately cause a miscompilation and the real implementation is a bit nuanced. I'll work on this to unblock you.

andrewrk (changes requested):

src/type.zig
mshockwave: Oh okay, then I guess I’ll just wait for now

andrewrk: I have implemented hashing for pointers in a recent master branch commit. If you rebase this PR on master you should be unblocked.

mshockwave: Cool, thanks for the heads up :-) I'll open a new one

PR #10497, opened 2022-01-03
Simplified Once synchronization primitive to use atomics

igor84:
Using Mutex can be slow since it has to make a syscall so using
@cmpxchgStrong atomic should be faster and it makes code much simpler


lithdew:
The Mutex is used so that all concurrent callers would suspend themselves until the one caller that had successfully acquired the mutex has completed its critical section.

This is generally desired when using Once as we want only a single caller to initialize some singleton instance, and have all other callers block until initialization is complete.

igor84:
Uh, I misunderstood its purpose. I'm closing the pull request then.

PR #10498, opened 2022-01-03
std.json: allow field missing when the field type is `Optional`

hronro:
When we say optional field in JSON, it usually means:

1. the value of the field may be `null`.
2. the field may not exist at all.

Currently the std library only support option 1 when parsing, this patch make it support option 2 as well.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1795,6 +1795,8 @@ fn parseInternal(
                         if (!field.is_comptime) {
                             @field(r, field.name) = default;
                         }
+                    } else if (@typeInfo(field.field_type) == .Optional) {
+                        @field(r, field.name) = null;
                     } else {
                         return error.MissingField;
                     }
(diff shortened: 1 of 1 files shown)


tauoverpi:
This usecase can currently be handled by setting the default value to `null` for the optional field which communicates the value that will be set if the field is missing. I feel that the current behaviour is better since the same applies to any field type rather than just optional fields.

example:
```zig
test {
    const std = @import("std");

    const text =
        \\{
        \\  "int": 4
        \\}
    ;

    const T = struct {
        uint: u8 = 5,
        opt: ?[]const u8 = null,
        int: i8,
    };

    var stream = std.json.TokenStream.init(text);

    const result = try std.json.parse(T, &stream, .{});

    try std.testing.expect(result.opt == null);
    try std.testing.expect(result.uint == 5);
    try std.testing.expect(result.int == 4);
}

daurnimator:
As @tauoverpi notes, the intended solution to optional fields is to have a default value.

PR #10517, opened 2022-01-05
Stage2: wasm-linker - Only export symbols notated as such

### LINKED ISSUE #2910: wasm output contains compiler_rt exports
foo.c:
```C
int foo(void) {
    return 42;
}
```

```
$ zig build-lib -target wasm32-freestanding --c-source foo.c
$ wasm-dis foo.wasm  | grep '(func' | tail
 (func $__udivmodti4 (; 265 ;) (type $50) (param $0 i32) (param $1 i64) (param $2 i64) (param $3 i64) (param $4 i64) (param $5 i32)
 (func $__umodti3 (; 266 ;) (type $28) (param $0 i32) (param $1 i64) (param $2 i64) (param $3 i64) (param $4 i64)
 (func $__muloti4 (; 267 ;) (type $50) (param $0 i32) (param $1 i64) (param $2 i64) (param $3 i64) (param $4 i64) (param $5 i32)
 (func $__mulodi4 (; 268 ;) (type $23) (param $0 i64) (param $1 i64) (param $2 i32) (result i64)
 (func $compiler_rt.mulXf3.S.loWord (; 269 ;) (type $35) (param $0 i64) (result i64)
 (func $compiler_rt.mulXf3.S.hiWord (; 270 ;) (type $35) (param $0 i64) (result i64)
 (func $compiler_rt.mulXf3.S.Word_1 (; 271 ;) (type $48) (param $0 i64) (param $1 i64) (result i64)
 (func $compiler_rt.mulXf3.S.Word_2 (; 272 ;) (type $48) (param $0 i64) (param $1 i64) (result i64)
 (func $compiler_rt.mulXf3.S.Word_3 (; 273 ;) (type $48) (param $0 i64) (param $1 i64) (result i64)
 (func $compiler_rt.mulXf3.S.Word_4 (; 274 ;) (type $48) (param $0 i64) (param $1 i64) (result i64)
```

Related #2062 #2290 #2369

### LINKED ISSUE #7133: wasm automatically links and re-exports all of libc
```
$ zig build-lib empty.zig -OReleaseSmall -target wasm32-freestanding

$ ls -l empty.*
-rwxr-xr-x  1 fengb  staff  48776 Nov 16 12:06 empty.wasm
-rw-r--r--  1 fengb  staff      0 Nov 16 12:06 empty.zig

$ twiggy top empty.wasm 
 Shallow Bytes │ Shallow % │ Item
───────────────┼───────────┼────────────────────────────────────────────────
          4255 ┊     8.72% ┊ data[0]
          2851 ┊     5.85% ┊ "function names" subsection
          1884 ┊     3.86% ┊ __divtf3
          1477 ┊     3.03% ┊ compiler_rt.addXf3.addXf3.27
          1384 ┊     2.84% ┊ __multf3
           941 ┊     1.93% ┊ __divdf3
           879 ┊     1.80% ┊ compiler_rt.udivmod.udivmod.64
           864 ┊     1.77% ┊ fma
           803 ┊     1.65% ┊ __muldf3
           778 ┊     1.60% ┊ compiler_rt.addXf3.addXf3.26
           767 ┊     1.57% ┊ __udivmoddi4
           675 ┊     1.38% ┊ __divsf3
           639 ┊     1.31% ┊ compiler_rt.addXf3.addXf3
           592 ┊     1.21% ┊ __mulsf3
           551 ┊     1.13% ┊ sqrt
           519 ┊     1.06% ┊ fmod
           512 ┊     1.05% ┊ expf
[...]
```

Luukdegram:
This exposes a function from stage2 to stage1 to append symbols to automatically export them.
Stage1 will only call this function when the target is wasm.

The linker will then only append `--export=<symbol>` for when the user has not done so manually and did not provide the `-rdynamic` flag also. This allows the user to still override the default behavior do they wish to.

I also went ahead and implemented the same behavior for the LLVM backend when using stage2, meaning we can remove all stage1 related code for this without losing the behavior when stage2 becomes the default compiler.

Closes #2910 
Closes #7133

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1882,9 +1882,9 @@ pub fn destroy(self: *Compilation) void {
     self.astgen_wait_group.deinit();
 
     for (self.export_symbol_names.items) |symbol_name| {
-        self.gpa.free(symbol_name);
+        gpa.free(symbol_name);
     }
-    self.export_symbol_names.deinit(self.gpa);
+    self.export_symbol_names.deinit(gpa);
 
     // This destroys `self`.
     self.arena_state.promote(gpa).deinit();
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -165,6 +165,10 @@ emit_docs: ?EmitLoc,
 work_queue_wait_group: WaitGroup,
 astgen_wait_group: WaitGroup,
 
+/// Exported symbol names. This is only for when the target is wasm.
+/// TODO: Remove this when Stage2 becomes the default compiler as it will already have this information.
+export_symbol_names: std.ArrayListUnmanaged([]const u8) = .{},
+
 pub const SemaError = Module.SemaError;
 
 pub const CRTFile = struct {
@@ -1876,6 +1880,11 @@ pub fn destroy(self: *Compilation) void {
 
     self.work_queue_wait_group.deinit();
     self.astgen_wait_group.deinit();
+
+    for (self.export_symbol_names.items) |symbol_name| {
+        self.gpa.free(symbol_name);
+    }
+    self.export_symbol_names.deinit(self.gpa);
 
     // This destroys `self`.
     self.arena_state.promote(gpa).deinit();
--- a/src/link/Wasm.zig
+++ b/src/link/Wasm.zig
@@ -1163,7 +1163,7 @@ fn linkWithLLD(self: *Wasm, comp: *Compilation) !void {
         };
     }
 
-    if (self.base.options.output_mode == .Obj) {
+    if (is_obj) {
         // LLD's WASM driver does not support the equivalent of `-r` so we do a simple file copy
         // here. TODO: think carefully about how we can avoid this redundant operation when doing
         // build-obj. See also the corresponding TODO in linkAsArchive.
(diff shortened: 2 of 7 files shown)


Luukdegram:
This is interesting. The error in the CI actually only happens when using `wasmtime`, where as `wasmer` executes the tests perfectly. For some reason, wasmtime assume an exported `u32` global is a function. (`wasm-objdump` correctly shows it as a global, that points to the symbol's value in the data section).
I'll investigate more tomorrow and create an issue downstream if it's indeed a bug on their end.

kubkon:
> This is interesting. The error in the CI actually only happens when using `wasmtime`, where as `wasmer` executes the tests perfectly. For some reason, wasmtime assume an exported `u32` global is a function. (`wasm-objdump` correctly shows it as a global, that points to the symbol's value in the data section). I'll investigate more tomorrow and create an issue downstream if it's indeed a bug on their end.

Note that this very well may be a bug with Wasmtime.

andrewrk:
Should we switch to wasmer in the CI, or perhaps upgrade to a newer wasmtime version?

Luukdegram:
It appears that under the current [WASI application ABI](https://github.com/WebAssembly/WASI/blob/main/design/application-abi.md#current-unstable-abi), instances may assume exports are not accessed when using the `command` execution model. This is exactly the model we use when running tests. Wasmer seems to simply ignore unexpected exports, whereas wasmtime is a bit more strict about it and errors. (There's a flag `--allow-unknown-exports` available to allow those).

Here's how I think we should solve it for now:
Append the `--allow-unknown-exports` flag to wasmtime, and revisit this flag when stage2 becomes the default compiler. (As we will have much more information about the export decls, as well as the execution model we're using).

kubkon:
Sounds reasonable to me!

src/Compilation.zig
```diff
@@ -1877,6 +1881,11 @@ pub fn destroy(self: *Compilation) void {
     self.work_queue_wait_group.deinit();
     self.astgen_wait_group.deinit();
 
+    for (self.export_symbol_names.items) |symbol_name| {
+        self.gpa.free(symbol_name);
+    }
+    self.export_symbol_names.deinit(self.gpa);
```
andrewrk, line 1887:
```suggestion
        gpa.free(symbol_name);
    }
    self.export_symbol_names.deinit(gpa);
```

src/link/Wasm.zig
```diff
@@ -1233,14 +1233,45 @@ fn linkWithLLD(self: *Wasm, comp: *Compilation) !void {
...
+            auto_export_symbols = false;
+        }
+
+        if (auto_export_symbols) {
+            if (self.base.options.module) |module| {
+                // when we use stage1, we use the exports that stage1 provided us.
+                // For stage2, we can directly retrieve them from the module.
+                const use_stage1 = build_options.is_stage1 and self.base.options.use_stage1;
+                if (use_stage1) {
+                    for (comp.export_symbol_names.items) |symbol_name| {
```
andrewrk, line 1255: Don't forget to put `export_symbol_names` in the cache hash up with the other hashed stuff :point_up: :point_up:

andrewrk (changes requested):
Nice work! Just need an additional cache input and then this is good to go.

src/link/Wasm.zig
Luukdegram: Good point, done :)

andrewrk (approved):
After making the previous review, I realized my suggestion about the cache hash is incorrect, since `export_symbol_names` depends only on the source files and cannot be overridden by CLI args.

This is good to merge.

andrewrk:
The Windows failure does look potentially relevant. The failing test is using the wasm32-wasi target.

Luukdegram:
> The Windows failure does look potentially relevant. The failing test is using the wasm32-wasi target.

Failure definitely seems to be related to this change, as it is reproducible on the CI. However, I fail to see why this is happening on Windows only. I'll see if I can rent out some cheap Windows VM as I don't own any Windows machine nor have the RAM to run a VM that hosts Windows and is able to build the compiler. Will need some more time to get this PR ready, but we'll get there!

Luukdegram:
Interestingly, commenting out the call to `stage2_append_symbol` does make the test cases for Windows succeed. This makes me think it may be some ABI issue, tho the function definition does seem alright.

kubkon:
@Luukdegram I got my local Win10 installation setup for development today so can have a look at this failure tomorrow. Will keep you posted!

PR #10519, opened 2022-01-06
lint: duplicate import

nektro:
found with https://github.com/nektro/ziglint and manually updated

### DIFF
--- a/lib/std/special/compiler_rt/int.zig
+++ b/lib/std/special/compiler_rt/int.zig
@@ -1,8 +1,9 @@
 // Builtin functions that operate on integer types
 const builtin = @import("builtin");
-const testing = @import("std").testing;
-const maxInt = @import("std").math.maxInt;
-const minInt = @import("std").math.minInt;
+const std = @import("std");
+const testing = std.testing;
+const maxInt = std.math.maxInt;
+const minInt = std.math.minInt;
 
 const udivmod = @import("udivmod.zig").udivmod;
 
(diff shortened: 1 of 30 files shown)


lib/std/special/compiler_rt/int.zig
```diff
@@ -1,8 +1,9 @@
 // Builtin functions that operate on integer types
 const builtin = @import("builtin");
-const testing = @import("std").testing;
-const maxInt = @import("std").math.maxInt;
-const minInt = @import("std").math.minInt;
+const std = @import("std");
+const testing = std.testing;
+const maxInt = std.math.maxInt;
+const minInt = std.math.minInt;
```
marler8997, line 6: So this is considered bad?

nektro: not inherently, but i think it helps readability

src/main.zig
```diff
@@ -23,6 +23,10 @@ const Cache = @import("Cache.zig");
 const target_util = @import("target.zig");
 const ThreadPool = @import("ThreadPool.zig");
 const crash_report = @import("crash_report.zig");
+const Module = @import("Module.zig");
+const AstGen = @import("AstGen.zig");
+const Zir = @import("Zir.zig");
+const llvm = @import("codegen/llvm/bindings.zig");
```
andrewrk, line 29: I don't want these in scope for the rest of the file. Don't want any accidental dependencies on these namespaces.

test/behavior/import.zig
```diff
@@ -12,7 +13,7 @@ test "importing the same thing gives the same import" {
 
 test "import in non-toplevel scope" {
     const S = struct {
-        usingnamespace @import("import/a_namespace.zig");
+        usingnamespace a_namespace;
```
andrewrk, line 16: notice the name of the test

andrewrk (changes requested):
Overall looks good- thank you for doing the lint. Just a couple fixes and then this is good to go.

PR #10525, opened 2022-01-07
Plan9 zig test

g-w1:


src/arch/x86_64/Isel.zig
```diff
@@ -753,7 +753,16 @@ fn dbgAdvancePCAndLine(isel: *Isel, line: u32, column: u32) InnerError!void {
...
             if (d_pc_p9 > 0) {
                 // minus one because if its the last one, we want to leave space to change the line which is one quanta
-                try dbg_out.dbg_line.append(@intCast(u8, @divExact(d_pc_p9, quant) + 128) - quant);
+                var diff = @divExact(d_pc_p9, quant) - quant;
+                while (diff > 0) {
+                    if (diff < 64) {
+                        try dbg_out.dbg_line.append(@intCast(u8, diff + 128));
+                        diff = 0;
+                    } else {
+                        try dbg_out.dbg_line.append(@intCast(u8, 64 + 128));
```
kubkon, line 762: Should this not be `@intCast(u8, diff + 128)`?

g-w1: No, the highest you can add to the pc is 64.
> A byte value of 129
          to 255 is reduced by 129, multiplied by the quantum of
          instruction size (e.g. two on the MC680X0), and added to the
          current PC without changing the offset.

kubkon (approved):

PR #10526, opened 2022-01-07
Move zig exe path to formal test args

SpexGuy:
This is one attempt at removing the implicit zig exe path parameter for tests.  It creates a more general system of "test arguments" which are specified at runtime for the test runner, through the command line.  This means the zig exe parameter only exists for the zig compiler tests (as configured by build.zig), not all tests built by zig.

It also splits the binary from the arguments for custom test executors.  This allows you to construct a command where it must be split, like `emulator <test_bin> <emulator_args> -- <test_args>`.

If this is the wrong direction, I'm happy to make changes.  But it seems much easier to me than getting test executors to honor environment variables or other methods of passing in information.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -45,6 +45,8 @@ pub fn build(b: *Builder) !void {
     test_stage2.addPackagePath("test_cases", "test/cases.zig");
     test_stage2.single_threaded = single_threaded;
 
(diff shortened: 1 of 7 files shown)


SpexGuy:
Closed in favor of #11633

PR #10528, opened 2022-01-07
stage2: EmitAsm

g-w1:


src/link/MachO.zig
```diff
@@ -3958,6 +3961,12 @@ pub fn getDeclVAddr(self: *MachO, decl: *const Module.Decl) u64 {
     return self.locals.items[decl.link.macho.local_sym_index].n_value;
 }
 
+pub fn getGot(self: *MachO, ally: Allocator) !link.File.GotWithAddr {
+    _ = self;
+    _ = ally;
+    @panic("TODO GET GOT for MACOS");
```
g-w1, line 3967: @kubkon I would appreciate help on how to get the got on macos to allow emit_asm to work on macos also.

src/link.zig
```diff
@@ -432,6 +435,23 @@ pub const File = struct {
         }
     }
 
+    pub const GotWithAddr = struct { code: []const u8, addr: usize };
+    /// Gets the GOT in it's current form as an allocated string if needed
+    /// Asserts backend uses a got
+    pub fn getGot(base: *File, ally: Allocator) !GotWithAddr {
```
kubkon, line 441: Is the `Allocator` really needed here? I think you pass to each backend the allocator instance that they define.

src/EmitAsm.zig
```diff
@@ -0,0 +1,137 @@
...
+    for (self.text_decl_to_asm_map.items) |*item| {
+        if (item.decl == decl) {
+            item.assembly = s;
+            return;
+        }
+    }
+    try self.text_decl_to_asm_map.append(self.mod.gpa, .{ .decl = decl, .assembly = s });
+}
+
+pub fn updateDataDecl(self: *EmitAsm, decl: *Module.Decl, code: []const u8) !void {
```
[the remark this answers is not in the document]
g-w1: It is anything that is not a zig function.

g-w1:
Hmm, I think hashmaps are very necessary to implement >2 sections, but when trying to use hashmaps, I get this dependency loop:
```zig
/home/jacob/.local/zig/0.10.0-dev.203+0772fb051/files/lib/std/hash_map.zig:178:33: error: dependency loop detected
            const hash = Context.hash;
                                ^
/home/jacob/.local/zig/0.10.0-dev.203+0772fb051/files/lib/std/array_hash_map.zig:465:40: note: called from here
    comptime std.hash_map.verifyContext(Context, K, K, u32);
                                       ^
/home/jacob/.local/zig/0.10.0-dev.203+0772fb051/files/lib/std/array_hash_map.zig:23:33: note: called from here
    return ArrayHashMapUnmanaged(K, V, AutoContext(K), !autoEqlIsCheap(K));
                                ^
./src/Module.zig:1388:46: note: called from here
    anon_decls: std.AutoArrayHashMapUnmanaged(*Decl, void) = .{},
                                             ^
./src/Module.zig:1388:46: note: called from here
    anon_decls: std.AutoArrayHashMapUnmanaged(*Decl, void) = .{},
                                             ^
/home/jacob/.local/zig/0.10.0-dev.203+0772fb051/files/lib/std/array_hash_map.zig:465:40: note: referenced here
    comptime std.hash_map.verifyContext(Context, K, K, u32);
                                       ^
/home/jacob/.local/zig/0.10.0-dev.203+0772fb051/files/lib/std/array_hash_map.zig:23:33: note: referenced here
    return ArrayHashMapUnmanaged(K, V, AutoContext(K), !autoEqlIsCheap(K));
                                ^
./src/Module.zig:1388:46: note: referenced here
    anon_decls: std.AutoArrayHashMapUnmanaged(*Decl, void) = .{},
                                             ^
./src/Module.zig:362:5: note: while checking this field
    src_namespace: *Namespace,
    ^
/home/jacob/.local/zig/0.10.0-dev.203+0772fb051/files/lib/std/array_hash_map.zig:2243:64: note: referenced here
pub fn getAutoHashFn(comptime K: type, comptime Context: type) (fn (Context, K) u32) {
                                                               ^
/home/jacob/.local/zig/0.10.0-dev.203+0772fb051/files/lib/std/array_hash_map.zig:2238:39: note: referenced here
        pub const hash = getAutoHashFn(K, @This());
                                      ^
./src/EmitAsm.zig:10:52: note: referenced here
text_decl_to_asm_map: std.AutoArrayHashMapUnmanaged(*Module.Decl, []const u8) = .{},
                                                   ^
./src/link.zig:68:5: note: while checking this field
    emit_asm: ?EmitAsm,
    ^
./src/link.zig:198:5: note: while checking this field
    options: Options,
    ^
./src/Compilation.zig:44:1: note: while checking this field
bin_file: *link.File,
^
./src/Module.zig:34:1: note: while checking this field
comp: *Compilation,
^
./src/Sema.zig:8:1: note: while checking this field
mod: *Module,
^
./src/crash_report.zig:39:5: note: while checking this field
    sema: En(*Sema),
    ^
./src/crash_report.zig:66:28: note: referenced here
threadlocal var zir_state: ?*AnalyzeBody = if (is_enabled) null else @compileError("Cannot use zir_state if crash_report is disabled.");
                           ^
./src/crash_report.zig:85:18: note: referenced here
    const anal = zir_state orelse return;
                 ^
./src/crash_report.zig:440:28: note: referenced here
        dumpStatusReport() catch |err| {
                           ^
```
Does this have something to do with the crash report?

andrewrk:
> Edit: just updated to use ArrayList instead of HashMap. It is less efficient, but at least it compiles.

This is not really an acceptable workaround. If the use case calls for a HashMap then use a hash map, and don't settle for an ArrayList. There's other work arounds to get around the compiler dependency loop. Worst case scenario, this PR is blocking on a stage1 bugfix to eliminate the false positive dependency loop.

g-w1:
> > Edit: just updated to use ArrayList instead of HashMap. It is less efficient, but at least it compiles.
> 
> This is not really an acceptable workaround. If the use case calls for a HashMap then use a hash map, and don't settle for an ArrayList. There's other work arounds to get around the compiler dependency loop. Worst case scenario, this PR is blocking on a stage1 bugfix to eliminate the false positive dependency loop.

I agree with this, do you have any ideas for getting around this dependency loop?

src/arch/x86_64/CodeGen.zig
```diff
@@ -326,17 +326,30 @@ pub fn generate(
...
+        defer buf.deinit();
+        const w = buf.writer();
+        const print = PrintMir{
+            .mir = mir,
+            .bin_file = bin_file,
+            .emit_asm = true,
+        };
+        try print.printMir(w, null, null);
+        // TODO make this __TEXT on macos etc...
+        try e.updateTextDecl(module_fn.owner_decl, buf.items, "section .text\n");
```
g-w1, line 351: I think we need to have a function on a decl called `getSection()` so this could be `try e.updateTextDecl(module_fn.owner_decl, buf.items, module_fn.owner_decl.getSection());`

kubkon:
Thanks for taking the time to work on this, and I apologise I didn't get round to reviewing it properly until now! I don't think this PR is in mergeable state as-is. Here are some of my thoughts why I think so.

In order to merge this PR, I believe there is quite a bit of work remaining to be done. Firstly, we should redo storing `*Module.Decl` into storing actual indexes after Andrew's latest changes to tracking decls which now rely on passing indexes rather than heap-allocated pointers. But more importantly, I spend some time yesterday playing with the generated output of `-femit-asm` using this PR and I noticed that the output is non-conformant with clang's syntax for assembly output - if we'd like to merge this new functionality I think it makes sense to merge something that works even for a minimal source file which it currently unfortunately doesn't.

Issue #6706, opened 2020-10-17
change pointers to zero-bit types to be actual pointers with addresses


tadeokondrak:
I don't think this optimization pulls its weight.

https://github.com/ziglang/zig/pull/6239 https://github.com/ziglang/zig/issues/1553 https://github.com/ziglang/zig/issues/4537 https://github.com/ziglang/zig/issues/6700

With https://github.com/ziglang/zig/pull/6432, you can't cast `*[1]u8` to `*[0]u8`.
I think that's a reasonable cast, and standard library relies on this working.

There are special cases that happen for *all* zero-bit types and should still be handled by generic code, but when some generic code is using a type solely as a pointer, it still needs to deal with them.

I think the main user of the `void` type in generic code is HashMap, and it doesn't seem to use `*V` anywhere.

tadeokondrak:
Regarding the actual value of the pointer, it should be legal to have any number but 0, since the compiler won't let you dereference it. To get a pointer to one, you could use `@intToPtr(*T, 1)`.
As far as I know this is what Rust does.

Snektron:
I would also like to note that this optimization makes Zig more complex, and makes it harder to communicate intent.

theInkSquid:
I totally agree with this. It's such an edge-case optimization that reduces the general consistency of the language.

katesuyu:
Another case where this optimization is fairly annoying is generic code that performs type erasure and requires an opaque "context pointer." In such a case, the 8 bytes are already "wasted", so the optimization does not help in this case; on the contrary, it causes strange-looking special casing involving `@bitSizeOf(T) == 0`.

N00byEdge:
What does this mean for addresses of zero sized struct members? Are `a` and `b` be defined to have the same address here, or will that be undefined?
```
pub S = struct {
  a: u0,
  b: u8,
};
```

SpexGuy:
Some notes on this decision:

The original motivation behind zero-sized pointers was not so much an optimization, but the result of an observation that a pointer is how you find a thing, and since zero-sized values can be anywhere, you don't need any information to find one.  The insight that we found most persuasive in reversing this is that pointers are also how you find a part of something larger, and you do need information to find a zero-sized piece of a larger structure or array.

[...]

PR #10532, opened 2022-01-07
Stage2 bit_shifting.zig passing

Hejsil:
This required some llvm and struct+union work. Hope I did it right. Each commit contains reasoning for why this change was done

### DIFF
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -2647,7 +2647,7 @@ pub const FuncGen = struct {
         switch (struct_ty.zigTypeTag()) {
             .Struct => {
                 var ptr_ty_buf: Type.Payload.Pointer = undefined;
-                const llvm_field_index = llvmFieldIndex(struct_ty, field_index, target, &ptr_ty_buf);
+                const llvm_field_index = llvmFieldIndex(struct_ty, field_index, target, &ptr_ty_buf).?;
                 const field_ptr = self.builder.buildStructGEP(struct_llvm_val, llvm_field_index, "");
                 const field_ptr_ty = Type.initPayload(&ptr_ty_buf.base);
                 return self.load(field_ptr, field_ptr_ty);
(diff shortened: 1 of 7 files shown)


src/codegen/llvm.zig
```diff
@@ -4761,19 +4761,23 @@ fn llvmFieldIndex(
...
+                .data = .{
+                    .pointee_type = if (field.ty.hasCodeGenBits())
+                        field.ty
+                    else
+                        Type.initTag(.i8),
+                    .@"align" = field.normalAlignment(target),
+                    .@"addrspace" = .generic,
+                },
+            };
+            return llvm_field_index;
```
Hejsil, line 4780: Actually, what are the semantics of taking a pointer to zero sized fields? Right now, I'm doing a pointer to the next none zero sized field, but it might as well just be a pointer to the start of the struct. I guess the semantics are "a pointer to an unspecified offset into the struct" so we can choose whatever. Maybe a pointer to the start of the struct is less buggy then, idk :shrug:

Hejsil: Wait, no, this gives you a pointer to the previous none zero sized field. Truly, I understand my own code well.

andrewrk: In stage2, `@sizeOf(void) == @sizeOf(i32)` because it implements  #6706

src/value.zig
```diff
@@ -2635,9 +2635,17 @@ pub const Value = extern union {
         var lhs_space: Value.BigIntSpace = undefined;
         const lhs_bigint = lhs.toBigInt(&lhs_space);
         const shift = @intCast(usize, rhs.toUnsignedInt());
+
+        const result_limps = lhs_bigint.limbs.len -| (shift / (@sizeOf(std.math.big.Limb) * 8));
```
Snektron, line 2639: Typo

src/codegen/llvm.zig
```diff
@@ -2647,7 +2647,7 @@ pub const FuncGen = struct {
         switch (struct_ty.zigTypeTag()) {
             .Struct => {
                 var ptr_ty_buf: Type.Payload.Pointer = undefined;
-                const llvm_field_index = llvmFieldIndex(struct_ty, field_index, target, &ptr_ty_buf);
+                const llvm_field_index = llvmFieldIndex(struct_ty, field_index, target, &ptr_ty_buf).?;
```
Hejsil, line 2650: I'm assuming here that `airStructFieldVal` should never be emitted when the field is zero sized, as then the result would always be comptime know, so the instruction would never make it to llvm.

andrewrk: That assumption is sound :+1:

andrewrk (approved):

PR #10533, opened 2022-01-07
Add missing package to the langref that's always available

r00ster91:
I noticed this is missing and thought I should add it.


andrewrk:
Thanks!

PR #10538, opened 2022-01-08
Fix a bug in std.Thread.Condition and add a basic Condition test.

afranchuk:
Condition in 0.9 is broken. This fixes it and adds a test to hopefully not have that happen again.

### DIFF
--- a/lib/std/Thread.zig
+++ b/lib/std/Thread.zig
@@ -1151,3 +1151,36 @@ test "Thread.detach" {
     event.wait();
     try std.testing.expectEqual(value, 1);
 }
+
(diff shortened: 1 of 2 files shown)


andrewrk (changes requested):
Thanks for the patch. The fix looks good,however, the new test case failed on the FreeBSD CI run.

afranchuk:
I almost expected as much since there were no tests at all before. I'll see if I can figure it out.

afranchuk:
I expect there's a race at the end with the thread still up and waiting to unlock the mutex after the test body is done; should be a simple fix once I get back on my computer. Not related to bsd systems alone.

andrewrk:
Thanks for the follow-up!

PR #10541, opened 2022-01-08
std: make `fs.openFileZ` `unreachable` if both `write` and `read` are…

### LINKED ISSUE #10540: `std.fs.openFileZ` wrongly assumes that a file is read-only when `OpenFlags.read` is set to `false`.
### Zig Version

0.10.0-dev.203+0772fb051

### Steps to Reproduce

Run `zig run file.zig` with the following file:

---

file.zig:
```zig
const std = @import("std");

pub fn main() !void {
    const file = try std.fs.cwd().openFile("file.zig", .{ .read = false });

    var buffer: [std.mem.page_size]u8 = undefined;
    _ = try file.read(&buffer);

    try stdout.print("{s}", .{buffer});
}
```

### Expected Behavior

The file shouldn't be readable.
https://github.com/ziglang/zig/blob/81fa31c05456facea1d1963a1e7f665351fb248d/lib/std/fs.zig#L991-L996
The following code is wrong because if both `.read` and `.write` is set to false, it should hit an `unreachable`.

As per https://man7.org/linux/man-pages/man2/open.2.html:
```
       The argument flags must include one of the following access
       modes: O_RDONLY, O_WRONLY, or O_RDWR.  These request opening the
       file read-only, write-only, or read/write, respectively.
```

---

I have not checked if this behavior would be considered incorrect for platforms other than Linux.

### Actual Behavior

The file is treated as readable, and thus, produces a quine:
```sh-session
$ zig run file.zig
const std = @import("std");

pub fn main() !void {
    const file = try std.fs.cwd().openFile("file.zig", .{ .read = false });

    var buffer: [std.mem.page_size]u8 = undefined;
    _ = try file.read(&buffer);

    try stdout.print("{s}", .{buffer});
}
```

sagehane:
Closes #10540.

Needs checking to make sure this behavior is correct for platforms other than Linux.

### DIFF
--- a/lib/std/Thread.zig
+++ b/lib/std/Thread.zig
@@ -79,7 +79,7 @@ pub fn setName(self: Thread, name: []const u8) SetNameError!void {
             var buf: [32]u8 = undefined;
             const path = try std.fmt.bufPrint(&buf, "/proc/self/task/{d}/comm", .{self.getHandle()});
 
-            const file = try std.fs.cwd().openFile(path, .{ .write = true });
+            const file = try std.fs.cwd().openFile(path, .{ .read = false, .write = true });
             defer file.close();
 
             try file.writer().writeAll(name);
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -992,8 +992,10 @@ pub const Dir = struct {
             @as(u32, os.O.RDWR)
         else if (flags.write)
             @as(u32, os.O.WRONLY)
+        else if (flags.read)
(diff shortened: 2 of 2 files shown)


leecannon:
This issue is caused by the fact that the `read` and `write` booleans in `File.OpenFlags` are not completely independant, would it be better to replace them with an enum?

```zig
pub const OpenMode = enum {
    read_only,
    write_only,
    read_write,
};
```

Then the unreachable code path is not representable.

nektro (commented):
reject

PR #10542, opened 2022-01-08
Stage2 bitOffsetOf and offsetOf builtin functions

Hejsil:
This PR is a precursor to implementing more complete `@bitCast` at comptime.


andrewrk:
Nice!

I took the time to type up related proposal #10547 which I am leaning towards accepting. Just a heads up because it is very relevant to implementing `@bitCast` at comptime. I would welcome your comments on that proposal.

PR #10545, opened 2022-01-08
Stage2: wasm - implement arrays

Luukdegram:
This gets all the array behavior tests passing.
Besides that, this also implements lowering struct constants which means `4560.zig` is passing once again.

### DIFF
--- a/src/arch/wasm/CodeGen.zig
+++ b/src/arch/wasm/CodeGen.zig
@@ -915,7 +915,7 @@ fn genTypedValue(self: *Self, ty: Type, val: Value) InnerError!Result {
             },
             .array => {
                 const elem_vals = val.castTag(.array).?.data;
-                const elem_ty = ty.elemType();
+                const elem_ty = ty.childType();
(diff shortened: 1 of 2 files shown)


src/arch/wasm/CodeGen.zig
```diff
@@ -2012,9 +2085,13 @@ fn emitUndefined(self: *Self, ty: Type) InnerError!void {
         },
         // As arrays point to linear memory, we cannot use 0xaaaaaaaa as the wasm
         // validator will not accept it due to out-of-bounds memory access);
-        .Array => try self.addImm32(@bitCast(i32, @as(u32, 0xaa))),
+        .Array => {
```
miscco, line 2088: Does it make sense to combine both cases?

Luukdegram: Good call. Also went ahead and made sure it writes 0xaa to its memory.

Luukdegram:
Rebased after cc5c25d to resolve merge conflict.

PR #10548, opened 2022-01-09
fs: Use `OpenMode` enum instead of read/write flags.

### LINKED ISSUE #10540: `std.fs.openFileZ` wrongly assumes that a file is read-only when `OpenFlags.read` is set to `false`.

sagehane:
Closes #10540, and is an alternate solution to #10541.

### DIFF
--- a/lib/std/Thread.zig
+++ b/lib/std/Thread.zig
@@ -79,7 +79,7 @@ pub fn setName(self: Thread, name: []const u8) SetNameError!void {
             var buf: [32]u8 = undefined;
             const path = try std.fmt.bufPrint(&buf, "/proc/self/task/{d}/comm", .{self.getHandle()});
 
-            const file = try std.fs.cwd().openFile(path, .{ .mode = .read_write });
+            const file = try std.fs.cwd().openFile(path, .{ .mode = .write_only });
             defer file.close();
 
             try file.writer().writeAll(name);
--- a/lib/std/Thread.zig
+++ b/lib/std/Thread.zig
@@ -79,7 +79,7 @@ pub fn setName(self: Thread, name: []const u8) SetNameError!void {
             var buf: [32]u8 = undefined;
             const path = try std.fmt.bufPrint(&buf, "/proc/self/task/{d}/comm", .{self.getHandle()});
 
-            const file = try std.fs.cwd().openFile(path, .{ .write = true });
+            const file = try std.fs.cwd().openFile(path, .{ .mode = .read_write });
             defer file.close();
 
             try file.writer().writeAll(name);
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -938,10 +938,10 @@ pub const Dir = struct {
         const w = os.wasi;
         var fdflags: w.fdflags_t = 0x0;
         var base: w.rights_t = 0x0;
-        if (flags.read) {
+        if (flags.isRead()) {
             base |= w.RIGHT.FD_READ | w.RIGHT.FD_TELL | w.RIGHT.FD_SEEK | w.RIGHT.FD_FILESTAT_GET;
         }
-        if (flags.write) {
+        if (flags.isWrite()) {
             fdflags |= w.FDFLAG.APPEND;
             base |= w.RIGHT.FD_WRITE |
                 w.RIGHT.FD_TELL |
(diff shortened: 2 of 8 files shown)


lib/std/fs/file.zig
```diff
@@ -118,6 +123,20 @@ pub const File = struct {
         /// Set this to allow the opened file to automatically become the
         /// controlling TTY for the current process.
         allow_ctty: bool = false,
+
+        fn isRead(self: OpenFlags) bool {
```
sagehane, line 127: Should I have taken a different approach than this?

sagehane:
I don't see anything obviously wrong with the logic, but I'm not confident with the code style so I'd love some reviews.

lib/std/fs.zig
```diff
@@ -938,10 +938,10 @@ pub const Dir = struct {
         const w = os.wasi;
         var fdflags: w.fdflags_t = 0x0;
         var base: w.rights_t = 0x0;
-        if (flags.read) {
+        if (flags.isRead()) {
```
sagehane, line 941: In hindsight, I could just have done `flags.mode != .write_only`. Would that be better? Or less legible?

lib/std/Thread.zig
```diff
@@ -79,7 +79,7 @@ pub fn setName(self: Thread, name: []const u8) SetNameError!void {
             var buf: [32]u8 = undefined;
             const path = try std.fmt.bufPrint(&buf, "/proc/self/task/{d}/comm", .{self.getHandle()});
 
-            const file = try std.fs.cwd().openFile(path, .{ .write = true });
+            const file = try std.fs.cwd().openFile(path, .{ .mode = .read_write });
```
Hejsil, line 82: This seems to only actually want `.mode = .write`

sagehane: I assumed that it wants both because `.read` is `true` by default, as per this commit:
https://github.com/ziglang/zig/blob/85de022c5671d777f62ddff254a814dab05242fc/lib/std/fs/file.zig#L75-L77
Are you implying that the existing code is wrong? Will check a bit more on it.

lib/std/fs/test.zig
```diff
@@ -319,9 +319,9 @@ test "file operations on directories" {
             try testing.expectError(error.IsDir, tmp_dir.dir.readFileAlloc(testing.allocator, test_dir_name, std.math.maxInt(usize)));
         },
     }
-    // Note: The `.write = true` is necessary to ensure the error occurs on all platforms.
+    // Note: The `.mode = .read_write` is necessary to ensure the error occurs on all platforms.
```
sagehane, line 322: I guess I should check if this works with `.write_only`, and perhaps make the comment a bit clearer.

lib/std/fs/watch.zig
```diff
@@ -678,7 +678,7 @@ fn testWriteWatchWriteDelete(allocator: Allocator) !void {
     };
 
     // overwrite line 2
-    const file = try std.fs.cwd().openFile(file_path, .{ .read = true, .write = true });
```
sagehane, line 681: I'm starting to question if this needs `.read = true` too. Not quite sure how to run the tests to make sure.
```sh-session
$ ~/repos/zig/build/zig test --zig-lib-dir lib lib/std/fs/watch.zig
Test [1/1] test "write a file, watch it, write it again, delete it"... SKIP
0 passed; 1 skipped; 0 failed.
```
This clearly doesn't seem to be the way to go.

Vexu (approved):

sagehane:
I'm still curious if the `watch.zig` test needs read permissions. Although I presume the other test is fine.

sagehane:
I thought this PR was a decent improvement over the other one, could you explain what you liked more about the other one?

PR #10551, opened 2022-01-09
stage2: enable zig test on x86_64-macos

kubkon:
Put `Decl`s in different MachO sections.

Use `getDeclVAddrWithReloc` when targeting MachO backend rather than `getDeclVAddr` - this fn returns a zero vaddr and instead creates a relocation on the linker side which will get automatically updated whenever the target decl is moved in memory. This fn also records a rebase of the target pointer so that its value is correctly slid in presence of ASLR.

This commit enables `zig test` on x86_64-macos.

### DIFF
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -220,7 +220,7 @@ managed_atoms: std.ArrayListUnmanaged(*Atom) = .{},
 /// We store them here so that we can properly dispose of any allocated
 /// memory within the atom in the incremental linker.
 /// TODO consolidate this.
-decls: std.AutoArrayHashMapUnmanaged(*Module.Decl, void) = .{},
+decls: std.AutoArrayHashMapUnmanaged(*Module.Decl, ?MatchingSection) = .{},
 
 /// Currently active Module.Decl.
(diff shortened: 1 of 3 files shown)


src/link/MachO.zig
```diff
@@ -3656,19 +3656,121 @@ pub fn updateDecl(self: *MachO, module: *Module, decl: *Module.Decl) !void {
...
+                .sect = self.text_section_index.?,
+            };
+        }
+        if (mem.allEqual(u8, code, 0)) {
+            break :blk MatchingSection{
+                .seg = self.data_segment_cmd_index.?,
+                .sect = self.bss_section_index.?,
+            };
+        }
+        if (mem.allEqual(u8, code, 0xaa)) {
```
andrewrk, line 3678: This would make the following code miscompile:

```zig
var x: u16 = 0xaaaa;
```

kubkon: Good point!

kubkon: I cleaned the logic up a bit, but I'm certain is not 100% correct nor finished up yet - I'm sure we will be able to tweak and rework it in tree.

PR #10553, opened 2022-01-09
stage2: handle field_ptr in C-backend renderValue

mattyhall:
Makes the tests in null_llvm.zig pass.


src/codegen/c.zig
```diff
@@ -392,6 +392,24 @@ pub const DeclGen = struct {
                     try dg.renderType(writer, ty);
                     try writer.print(")0x{x}u)", .{val.toUnsignedInt()});
                 },
+                .field_ptr => {
+                    const field_ptr = val.castTag(.field_ptr).?.data;
+                    const index = field_ptr.field_index;
+                    const container_ptr = field_ptr.container_ptr;
+                    const decl = switch (container_ptr.tag()) {
+                        .decl_ref => container_ptr.castTag(.decl_ref).?.data,
```
mattyhall, line 400: I'm not quite sure what values I need to handle here. I've seen in other places in this file it handles a subset of tags that you might expect (eg decl_ref but not decl_ref_mut)

jmc-88:
I believe this PR is now obsolete, @topolarity fixed this via `renderChildPtr()` in https://github.com/ziglang/zig/commit/1639fd4c45ee816c7c67963fbff3513a32f5737b.

PR #10556, opened 2022-01-09
lint: unused local declarations

nektro:
found with https://github.com/nektro/ziglint and manually fixed


lib/std/c/freebsd.zig
```diff
@@ -944,8 +944,6 @@ pub const winsize = extern struct {
     ws_ypixel: u16,
 };
 
-const NSIG = 32;
```
andrewrk, line 947: The fix for this one is to make it `pub`. Same with all the other instances.

lib/std/crypto/siphash.zig
```diff
@@ -46,8 +46,6 @@ fn SipHashStateless(comptime T: type, comptime c_rounds: usize, comptime d_round
 
     return struct {
         const Self = @This();
-        const block_length = 64;
-        const digest_length = 64;
```
andrewrk, line 50: These are observable from outside siphash.zig. They should be `pub`.

lib/std/start.zig
```diff
@@ -1,12 +1,12 @@
 // This file is included in the compilation unit when exporting an executable.
+//
```
andrewrk, line 2: stray `//`

lib/std/zig/system/linux.zig
```diff
@@ -268,6 +267,7 @@ const ArmCpuinfoImpl = struct {
                 0x43 => &Cavium,
                 0x46 => &Fujitsu,
                 0x48 => &HiSilicon,
+                0x4e => &Nvidia,
```
andrewrk, line 270: Nice fix

test/behavior/array.zig
```diff
@@ -58,9 +57,6 @@ test "array dot len const expr" {
     });
 }
 
-const ArrayDotLenConstExpr = struct {
```
andrewrk, line 61: please check the git history of this and make sure a test that referenced this decl was not accidentally deleted

test/behavior/sizeof_and_typeof.zig
```diff
@@ -1,10 +1,10 @@
 const std = @import("std");
 const expect = std.testing.expect;
-const expectEqual = std.testing.expectEqual;
 
 test "@sizeOf and @TypeOf" {
     const y: @TypeOf(x) = 120;
     try expect(@sizeOf(@TypeOf(y)) == 2);
+    try expect(@sizeOf(@TypeOf(z)) == 2);
```
andrewrk, line 7:
```suggestion
    try expect(@TypeOf(z) == u16);
```

test/behavior/typename.zig
```diff
@@ -1,5 +1,5 @@
+//
```
andrewrk, line 1: stray line

andrewrk (changes requested):
Thanks for the PR. Left some requested changes.

test/behavior/typename.zig
nektro: needed for source sensitive test

lib/std/start.zig
nektro: needed for source sensitive test

lib/std/crypto/siphash.zig
nektro: https://discord.com/channels/697519923273924658/713755801583616100/929803897252511794

andrewrk: Can you post the content here?

lib/std/os/linux/start_pie.zig
```diff
@@ -16,6 +15,7 @@ const R_RELATIVE = switch (builtin.cpu.arch) {
     .arm => R_ARM_RELATIVE,
     .aarch64 => R_AARCH64_RELATIVE,
     .riscv64 => R_RISCV_RELATIVE,
+    .sparc => R_SPARC_RELATIVE,
```
nektro, line 18: should this be `.sparc` or `.sparcv9`

lib/std/start.zig
ifreund: > needed for source sensitive test

what does this mean?

andrewrk: Ah, the stack trace test checks lines and columns and unfortunately that means the test case breaks whenever lines are added or removed from start.zig.

The annoying but proper fix is to adjust the test case expected line number. Or just don't mess with this file in this PR.

test/behavior/typename.zig
andrewrk: not sure what you're talking about on this one

lib/std/os/linux/start_pie.zig
andrewrk: definitely at least `.sparc`. looks like also `.sparcv9`

test/behavior/typename.zig
Hejsil: I think this refers to the fact that these files source-location sensitive, as some of these type names contains source lines and columns in the names.

test/behavior/array.zig
nektro: <https://github.com/ziglang/zig/commit/581edd643> first added in its modern form here
<https://github.com/ziglang/zig/commit/0cd361219> moved to `_stage1.zig` here
<https://github.com/ziglang/zig/commit/6d6cf5984> implemented here

nektro: asserting that `y: [some_array.len]u8` properly turned to `y: [4]u8` is likely what it was meant to test, but never added

nektro: <https://github.com/ziglang/zig/commit/6b3ce918d> first added ever here
<https://github.com/ziglang/zig/commit/dab3ddab4> getting closer to modern zig

Issue #9594, opened 2021-08-20
`std.io.Reader.readUntilDelimiter*` functions lose last byte if it is past the max size


Vexu:
Seems like this could be resolved by requiring space for the delimiter.

More tests are also needed.

PR #10557, opened 2022-01-09
readUntilDelimiter*: read only if buffer not full

scorphus:
This is one way I see #9594 being solved. I hope I didn't get it too far off. I'm more than happy making any changes (and learn in the process)

I decided to start with `readUntilDelimiter` and `readUntilDelimiterOrEof` and submit this as draft to collect early feedbacks — which are more than welcome 🙂 I'll continue to the other `readUntilDelimiter*` functions.

Happy to propose my first contribution to Zig! 🎉 Thanks to @SpexGuy for suggesting me this issue.

### DIFF
--- a/lib/std/io/reader.zig
+++ b/lib/std/io/reader.zig
@@ -107,14 +107,14 @@ pub fn Reader(
         ) !void {
             array_list.shrinkRetainingCapacity(0);
             while (true) {
+                if (array_list.items.len == max_size) {
+                    return error.StreamTooLong;
+                }
+
                 var byte: u8 = try self.readByte();
 
                 if (byte == delimiter) {
                     return;
-                }
-
-                if (array_list.items.len == max_size) {
-                    return error.StreamTooLong;
                 }
 
                 try array_list.append(byte);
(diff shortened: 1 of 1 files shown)


SpexGuy:
Not all readers support seekBy.  I think a better solution would be to avoid reading at all if the output buffer is full.

scorphus:
> Not all readers support seekBy. I think a better solution would be to avoid reading at all if the output buffer is full.

Yeah, it's so true that most streams don't support rewinding.

I actually thought about that a few iterations before. Without reading the next byte, how's it going to be able to stop and return what can be a full buffer?

scorphus:
For instance:

```zig
test "Reader.readUntilDelimiter EndOfStream 2" {
    var buf: [4]u8 = undefined;
    const reader = std.io.fixedBufferStream("1234\n").reader();
    try std.testing.expectEqualStrings("1234", try reader.readUntilDelimiter(&buf, '\n'));
    try std.testing.expectError(error.EndOfStream, reader.readUntilDelimiter(&buf, '\n'));
}
```

SpexGuy:
That test would no longer pass, the buffer would need to be at least 5 bytes long.  An easy way to communicate that might be to say that the output buffer needs to contain (or be terminated by) the delimiter if one was found.

scorphus:
Thanks, @SpexGuy! PR updated. I would appreciate reading from you again, while I write tests for the remaining `readUntilDelimiter*` and start checking namespaces that use them (e.g. `std.net`).

lib/std/io/reader.zig
```diff
@@ -144,13 +144,14 @@ pub fn Reader(
...
-
-                if (byte == delimiter) return buf[0..index];
                 if (index >= buf.len) return error.StreamTooLong;
 
+                const byte = try self.readByte();
+
                 buf[index] = byte;
                 index += 1;
+
+                if (byte == delimiter) return buf[0..index];
```
SpexGuy, line 154: This should probably be `byte[0..index-1]`, or placed before incrementing the index.  Callers don't usually expect the delimiter to be within the len.  But we should still write it to the buffer before returning.

SpexGuy (changes requested):
Thanks for doing this work!

scorphus:
You're welcome. I'm happy to contribute 🙂 And this is fun! It makes much sense after I thought about what you wrote. Thanks, @SpexGuy!

lib/std/io/reader.zig
```diff
@@ -144,12 +144,13 @@ pub fn Reader(
         pub fn readUntilDelimiter(self: Self, buf: []u8, delimiter: u8) ![]u8 {
```
SpexGuy, line 144: The documentation should be updated to say something like "The delimiter byte is written to the output buffer, but is not included in the returned slice."

SpexGuy (changes requested):
Just noticed one more thing :)

lib/std/io/reader.zig
scorphus: Oh yes, definitely! I'm only waiting until we have a resolute "new behavior" so I can update all those docs ;-)

Vexu (approved):

PR #10563, opened 2022-01-10
Build fs/filesystem libcxx module when targeting GNU Win

kubkon:


SpexGuy (approved):

src/libcxx.zig
```diff
@@ -114,8 +114,8 @@ pub fn buildLibCXX(comp: *Compilation) !void {
     for (libcxx_files) |cxx_src| {
         var cflags = std.ArrayList([]const u8).init(arena);
 
-        if (target.os.tag == .windows or target.os.tag == .wasi) {
-            // Filesystem stuff isn't supported on WASI and Windows.
+        if ((target.os.tag == .windows and target.abi == .msvc) or target.os.tag == .wasi) {
```
marler8997, line 117: I think this may have broke zig for windows.  This change causes `src/filesystem/int128_builtins.cpp` to be included in `c++.lib` on windows, but the implementation causes a segfault on Windows x86_64.  This file provides `__muloti4` but that function is already provided by `compiler_rt`.  I created a PR to remove `src/filesystem/int128_builtins.cpp` for all targets here (https://github.com/ziglang/zig/pull/10729) since it should be provided by compiler_rt in all platforms as far as I know.  I also confirmed this change fixed the crash on windows.

kubkon: Thanks for working this out @marler8997 !

PR #10570, opened 2022-01-11
Make prominent_compile_errors also work for test steps

Hejsil:
Seems like the right solution, but who knows :shrug:


andrewrk:
I poked around. Seems fine :+1:

Issue #10521, opened 2022-01-06
std.net.Address.resolveIp fails on macOS 12.1 on M1 Macbook


dantecatalfamo:
### Zig Version

0.10.0-dev.203+0772fb051

### Steps to Reproduce

```zig
const std = @import("std");

test "resolve ip" {
    try std.net.Address.resolveIp("google.com", 80);
}
```
```
zig test test.zig
```

### Expected Behavior

The test passes

### Actual Behavior

```
$ zig test test.zig
/opt/homebrew/Cellar/zig/HEAD-0772fb0/lib/zig/std/os.zig:83:30: error: container 'std.c' has no member called 'IFNAMESIZE'
pub const IFNAMESIZE = system.IFNAMESIZE;
                             ^
/opt/homebrew/Cellar/zig/HEAD-0772fb0/lib/zig/std/net.zig:427:32: note: referenced here
        var scope_id_value: [os.IFNAMESIZE - 1]u8 = undefined;
                               ^
/opt/homebrew/Cellar/zig/HEAD-0772fb0/lib/zig/std/net.zig:85:32: note: referenced here
        return Address{ .in6 = try Ip6Address.resolve(buf, port) };
                               ^
/opt/homebrew/Cellar/zig/HEAD-0772fb0/lib/zig/std/net.zig:58:9: note: referenced here
        if (resolveIp6(name, port)) |ip6| return ip6 else |err| switch (err) {
        ^
./test.zig:4:5: note: referenced here
    try std.net.Address.resolveIp("google.com", 80);
    ^
```

The-King-of-Toasters:
This is [apparently 16](https://github.com/ziglang/fetch-them-macos-headers/blob/main/headers/x86_64-macos.12-gnu/net/if.h#L66) according to the system headers. I'd submit a pr myself but I don't have a macOS install handy.

PR #10576, opened 2022-01-12
Use libc if_nametoindex for macOS when parsing IPs

schmee:
Fixes https://github.com/ziglang/zig/issues/10521 and makes a couple of additional tests pass ~when linking libc~ for macOS.

### DIFF
--- a/lib/std/net.zig
+++ b/lib/std/net.zig
@@ -650,7 +650,7 @@ fn if_nametoindex(name: []const u8) !u32 {
         return @bitCast(u32, ifr.ifru.ivalue);
     }
 
-    if (builtin.link_libc) {
+    if (comptime builtin.target.os.tag.isDarwin()) {
         if (name.len >= os.IFNAMESIZE)
             return error.NameTooLong;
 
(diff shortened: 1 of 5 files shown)


schmee:
Judging by the CI, seems like checking for `builtin.link_libc` instead of `builtin.target.os.tag == .macos was a bit optimistic. Fix incoming.

lib/std/net.zig
```diff
@@ -636,17 +636,35 @@ pub fn connectUnixSocket(path: []const u8) !Stream {
...
+
+        // TODO investigate if this needs to be integrated with evented I/O.
+        try os.ioctl_SIOCGIFINDEX(sockfd, &ifr);
 
-    std.mem.copy(u8, &ifr.ifrn.name, name);
-    ifr.ifrn.name[name.len] = 0;
+        return @bitCast(u32, ifr.ifru.ivalue);
+    }
+
+    if (builtin.target.os.tag == .macos) {
```
schmee, line 653: This could also check for any of `.macos`, `.ios`, `.tvos`, `.watchos`, as is done [here](https://github.com/ziglang/zig/blob/master/lib/std/c.zig#L45) if that is preferable.

kubkon: How about `builtin.target.isDarwin()` then?

schmee: Perfect, I wasn't aware of that function! Will update 👍

schmee: @kubkon updated!

PR #10580, opened 2022-01-12
Thread: replaced SpinFutex with Thread.Futex for macOS in StaticResetEvent

kristoff-it:
also updated the interfaces of the other Futex implementations
private to StaticResetEvent to match the one of std.Thread.Futex

### DIFF
--- a/lib/std/Thread/StaticResetEvent.zig
+++ b/lib/std/Thread/StaticResetEvent.zig
@@ -114,7 +114,7 @@ pub const DebugEvent = struct {
 };
 
 pub const AtomicEvent = struct {
-    waiters: u32 = 0,
+    waiters: std.atomic.Atomic(u32) = std.atomic.Atomic(u32).init(0),
 
     const WAKE = 1 << 0;
     const WAIT = 1 << 1;
@@ -132,7 +132,7 @@ pub const AtomicEvent = struct {
     }
 
     pub fn set(ev: *AtomicEvent) void {
-        const waiters = @atomicRmw(u32, &ev.waiters, .Xchg, WAKE, .Release);
+        const waiters = ev.waiters.swap(WAKE, .Release);
         if (waiters >= WAIT) {
             return Futex.wake(&ev.waiters, waiters >> 1);
         }
@@ -146,10 +146,10 @@ pub const AtomicEvent = struct {
     }
 
     pub fn timedWait(ev: *AtomicEvent, timeout: ?u64) TimedWaitResult {
-        var waiters = @atomicLoad(u32, &ev.waiters, .Acquire);
+        var waiters = ev.waiters.load(.Acquire);
         while (waiters != WAKE) {
-            waiters = @cmpxchgWeak(u32, &ev.waiters, waiters, waiters + WAIT, .Acquire, .Acquire) orelse {
-                if (Futex.wait(&ev.waiters, timeout)) |_| {
+            waiters = ev.waiters.tryCompareAndSwap(waiters, waiters + WAIT, .Acquire, .Acquire) orelse {
+                if (Futex.wait(&ev.waiters, WAIT, timeout)) |_| {
                     return .event_set;
                 } else |_| {
                     return .timed_out;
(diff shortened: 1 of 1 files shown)


kristoff-it:
@kprotty I should also point out that the new API has a `expected` parameter that was not there before and you should double check that I didn't put anything stupid there. I tested my changes (both compiled and run) on all 3 platforms so it should be fine, but still.

lib/std/Thread/StaticResetEvent.zig
```diff
@@ -114,7 +114,7 @@ pub const DebugEvent = struct {
 };
 
 pub const AtomicEvent = struct {
```
kprotty, line 116: This struct makes me sad and is what prompted a [sync primitive rewrite](https://github.com/kprotty/zig/tree/posix-simple/lib/std/Thread). Tracking the waiters for set() was to account for window's NtKeyedEvent's API only waking one waiter at a time. `std.Thread.Futex` doesn't use that now so it can be made simpler. 

Trying to track waiters while using Futex isn't ideal as the value it waits on needs to stay the same until the set() happens. Each new thread wait()'ing bumps the waiters WAIT count potentially ending up in the waiting threads spuriously waking up to now wait on the new waiters count.

Either ways, a quick fix should be down below.

lib/std/Thread/StaticResetEvent.zig
```diff
@@ -160,27 +160,28 @@ pub const AtomicEvent = struct {
...
 
     pub fn reset(ev: *AtomicEvent) void {
-        @atomicStore(u32, &ev.waiters, 0, .Monotonic);
+        ev.waiters.store(0, .Monotonic);
     }
 
     pub const Futex = switch (builtin.os.tag) {
         .windows => WindowsFutex,
         .linux => LinuxFutex,
+        .macos => std.Thread.Futex,
```
kprotty, line 169: Futex.wait() usage in `timedWait()` is expected to not have spurious wakeups but `std.Thread.Futex`'s API does. Convert it into a loop to account for this. See the following pseudo code

```zig
struct:
  wake(waiters, count):
    Futex.wake(waiters, count)
    
  wait(waiters, timeout):
    deadline = timeout != null ? nanoTimestamp() : null
    while true:
      w = waiters.load(Acquire)
      if w == WAKE: return
      Futex.wait(waiters, w, blk:
        _ = timeout orelse break null
        now = nanoTimestamp()
        if now >= deadline: return error.TimedOut
        break deadline - now
```

Not really supposed to use `nanoTimestamp()` for tracking elapsed time, but whatever. Also might as well use that pseudo code for all platforms not just macos (since std.Thread.Futex is portable and doesn't spin)

kprotty (changes requested):

kprotty (approved):
Looks about right. Would recommend also replacing `SpinFutex` with `ThreadFutexWrapper` as well, but this is fine for now.

kristoff-it:
@kprotty this test failed in CI, ideas?

```
338/1081 behavior.atomics.test "behavior-native-ReleaseSafe-c-multi atomicrmw with floats"... behavior.atomics.test "behavior-native-ReleaseSafe-c-multi atomicrmw with floats"... FAIL (TestUnexpectedResult)

FAIL (TestUnexpectedResult)

/drone/src/test/behavior/defer.zig:42:13: 0x2324a3 in behavior.defer.test "behavior-native-ReleaseSafe-c-multi errdefer does not apply to fn inside fn" (test)

            return error.Bad;

            ^

/drone/src/test/behavior/defer.zig:45:5: 0x2324ab in behavior.defer.test "behavior-native-ReleaseSafe-c-multi errdefer does not apply to fn inside fn" (test)

    return S.baz();

    ^

/drone/src/test/behavior/error.zig:27:5: 0x2324e7 in behavior.error.test "behavior-native-ReleaseSafe-c-multi error binary operator" (test)

    return if (x) error.ItBroke else @as(isize, 10);

    ^

/drone/src/test/behavior/try.zig:31:9: 0x232607 in behavior.try.test "behavior-native-ReleaseSafe-c-multi try without vars" (test)

        return error.ItBroke;

        ^

/drone/src/test/behavior/try.zig:31:9: 0x23261f in behavior.try.test "behavior-native-ReleaseSafe-c-multi try then not executed with assignment" (test)

        return error.ItBroke;

        ^

/drone/src/test/behavior/defer_stage1.zig:47:13: 0x23b03f in behavior.defer_stage1.test "behavior-native-ReleaseSafe-c-multi errdefer with payload" (test)

            return error.One;

            ^

/drone/src/test/behavior/while.zig:61:5: 0x2327c7 in behavior.while.test "behavior-native-ReleaseSafe-c-multi while with error union condition" (test)

    return if (numbers_left == 0) error.OutOfNumbers else x: {

    ^

/drone/src/test/behavior/merge_error_sets.zig:10:5: 0x2349bf in behavior.merge_error_sets.test "behavior-native-ReleaseSafe-c-multi merge error sets" (test)

    return error.NotDir;

    ^

/drone/src/test/behavior/atomics.zig:95:42: 0x232e4b in behavior.atomics.test "behavior-native-ReleaseSafe-c-multi 128-bit cmpxchg" (test)

        if (builtin.cpu.arch != .x86_64) return error.SkipZigTest;

                                         ^

/drone/src/test/behavior/atomics.zig:86:5: 0x232e53 in behavior.atomics.test "behavior-native-ReleaseSafe-c-multi 128-bit cmpxchg" (test)

    try test_u128_cmpxchg();

    ^

/drone/src/lib/std/testing.zig:0:5: 0x2330a3 in behavior.atomics.test "behavior-native-ReleaseSafe-c-multi atomicrmw with floats" (test)

/drone/src/lib/std/testing.zig:0:5: 0x2330ab in behavior.atomics.test "behavior-native-ReleaseSafe-c-multi atomicrmw with floats" (test)

/drone/src/test/behavior/atomics.zig:156:5: 0x2330b3 in behavior.atomics.test "behavior-native-ReleaseSafe-c-multi atomicrmw with floats" (test)

    try testAtomicRmwFloat();

    ^
```

Vexu:
Closed by #11523

PR #10582, opened 2022-01-12
stage2: detection of comptime array literals

andrewrk:
Introduce `validate_array_init_comptime`, similar to
`validate_struct_init_comptime` introduced in
713d2a9b3883942491b40738245232680877cc66.

`zirValidateArrayInit` is improved to detect comptime array literals and
emit AIR accordingly. This code is very similar to the changes
introduced in that same commit for `zirValidateStructInit`.

The C backend needed some improvements to continue passing the same set
of tests:
 * `resolveInst` for arrays now will add a local `static const` with the
   array value and so then `elem_val` instructions reference that local.
   It memoizes accesses using `value_map`, which is changed to use
   `Air.Inst.Ref` as the key rather than `Air.Inst.Index`.
 * This required a mechanism for writing to a "header" which is lines
   that appear at the beginning of a function body, before everything
   else.
 * dbg_stmt output comments rather than `#line` directives.
   TODO comment reproduced here:

We need to re-evaluate whether to emit these or not. If we naively emit
these directives, the output file will report bogus line numbers because
every newline after the #line directive adds one to the line.
We also don't print the filename yet, so the output is strictly unhelpful.
If we wanted to go this route, we would need to go all the way and not output
newlines until the next dbg_stmt occurs.
Perhaps an additional compilation option is in order?

`Value.elemValue` is improved to support `elem_ptr` values.

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -1418,7 +1418,13 @@ fn arrayInitExprRlPtrInner(
         extra_index += 1;
         _ = try expr(gz, scope, .{ .ptr = elem_ptr }, elem_init);
     }
-    _ = try gz.addPlNodePayloadIndex(.validate_array_init, node, payload_index);
+
+    const tag: Zir.Inst.Tag = if (gz.force_comptime)
+        .validate_array_init_comptime
+    else
+        .validate_array_init;
+
+    _ = try gz.addPlNodePayloadIndex(tag, node, payload_index);
     return .void_value;
 }
 
@@ -2317,6 +2323,7 @@ fn unusedResultExpr(gz: *GenZir, scope: *Scope, statement: Ast.Node.Index) Inner
             .validate_struct_init,
             .validate_struct_init_comptime,
             .validate_array_init,
+            .validate_array_init_comptime,
             .set_align_stack,
             .set_cold,
             .set_float_mode,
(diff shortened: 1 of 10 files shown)


andrewrk:
Here's a repro for failing wasm test:

```zig
const std = @import("std");

export fn entry() bool {
    doTheTest() catch return false;
    return true;
}

fn doTheTest() !void {
    const array_of_strings = [_][]const u8{ "hello", "this", "is", "my", "thing" };
    for (array_of_strings) |s, i| {
        if (i == 0) try expect(mem.eql(u8, s, "hello"));
        if (i == 1) try expect(mem.eql(u8, s, "this"));
        if (i == 2) try expect(mem.eql(u8, s, "is"));
        if (i == 3) try expect(mem.eql(u8, s, "my"));
        if (i == 4) try expect(mem.eql(u8, s, "thing"));
    }
}
```

```
./zig-out/bin/zig build-obj test.zig  --verbose-air --verbose-llvm-ir  -target wasm32-wasi -femit-bin=test.wasm
```

It produces output but miscompiled. I suspect in the lowering of `array_elem_val`.

andrewrk:
Disabled the 2 failing wasm behavior tests for now. After that, all tests passed locally for me.

PR #10583, opened 2022-01-12
net: Use allocPrintZ to avoid needing assumeSentinel

bryfry:
Removes a hard coded null terminator and the associated assumeSentinel.  Replaced with the std.fmt.allocPrintZ which makes it clear what/why it's all happening.


andrewrk:
Thanks!

PR #10584, opened 2022-01-12
zld: a couple of fixes which result in better rustc support

kubkon:
By `rustc` support, I mean that this a step closer to using `zig cc` (or in fact, `zig ld`) as a linker replacement for Rust. The only thing missing now is compact unwinding support (which incidentally is also missing for C++). I shall focus on that next when working on MachO!

Changes:
* fix num nlist calculation when there's no dynsymtab available - off by one error
* handle `__DATA,.rustc` section containing `rustc` metadata - this is required to get crates like `serde_derive` link properly. I should note that this special section has to be copied __verbatim__ from the relocatable object file - this includes preserving its size regardless of alignment requirements.
* support `-Wl,-force_load=<archive_path>` flag - this actually enables using `zig cc` as a linker for `cargo test` (minus setting up panic handlers which are a TODO and related to unwinding support).
* add parsing of addend from the instruction for `BRANCH` reloc on x86_64 - turns out, it can be non-zero!

EDIT: fixes a segfault issue reported by our user in Zig Forum [here](https://zigforum.org/t/cross-compiled-darwin-binary-segfaults/1041)


src/Compilation.zig
```diff
@@ -699,6 +699,7 @@ pub const InitOptions = struct {
     rpath_list: []const []const u8 = &[0][]const u8{},
     c_source_files: []const CSourceFile = &[0]CSourceFile{},
     link_objects: []const []const u8 = &[0][]const u8{},
+    must_link_objects: []const []const u8 = &[0][]const u8{},
```
andrewrk, line 702: There are a few places throughout the codebase that the `link_objects` array (1 line above) is used for various things; to calculate the root name, choose the linker backend, caching stuff, and probably more that I'm not thinking of. In order to avoid introducing subtle bugs, I suggest this change:

```diff
-    link_objects: []const []const u8 = &[0][]const u8{},
-    must_link_objects: []const []const u8 = &[0][]const u8{}, 
+    link_objects: []const LinkObject = &.{},
```

With `LinkObject` defined like this:

```zig
const LinkObject = struct {
    path: []const u8,
    must_link: bool = false,
};
```

You'll have to go and edit a bunch of places in the code to make this change but I think it will reveal problems with this patch, and make it straightforward to fix and maintain.

andrewrk (changes requested):
Nice work! I spotted one issue but I think it should be straightforward to solve with my suggestion.

src/Compilation.zig
kubkon: Changes introduced in 59ef553 - have a look and lemme know if that's what you had in mind!

PR #10585, opened 2022-01-13
std.json: optional fields

dreinharth:
Do not require optional fields when parsing JSON into a struct


daurnimator:
If you want to be able to omit a value, set a default value.

Duplicate of https://github.com/ziglang/zig/pull/10498

Issue #10512, opened 2022-01-04
another zig cc regression in 0.9.0


travisstaloch:
The following runs fine with clang and previous versions of zig.  Not sure when it stopped working but seems to have happened in the past month as i first noticed this from a CI failure around 3 weeks ago (approx 12/7/21).  Seems related to #10386.  I am on a linux debian system: Linux 5.6.10-antix.1-amd64-smp

```c++
// tmp.cpp
#include <unordered_map>
#include <string>
#include <stdio.h>

const static std::unordered_map<char, int> ranks = {
  {'0', 0}, {'1', 1}, {'2', 2}, {'3', 3},
};

int main(int argc, char const *argv[])
{ 
  printf("ranks.at('2'): %d\n", ranks.at('2'));
  return 0;
}
```

```console
$ zig run tmp.cpp -lc++
libc++abi: terminating with uncaught exception of type std::out_of_range: unordered_map::at: key not found
Aborted
$ zig version
0.10.0-dev.63+6b8e33d14
``` 

This works fine when targeting musl: 
```console
$ zig run tmp.cpp -lc++ -target native-native-musl
ranks.at('2'): 2
```

Moving the the map to a non-static local in main also makes it work (without musl):

```cpp
// tmp2.cpp: works fine 
#include <unordered_map>
#include <string>
#include <stdio.h>


int main(int argc, char const *argv[])
{ 
  const std::unordered_map<char, int> ranks = {
    {'0', 0}, {'1', 1}, {'2', 2}, {'3', 3},
  };
  printf("ranks.at('2'): %d\n", ranks.at('2'));
  return 0;
}
```

sagehane:
I wasn't sure if this stems from the same issue, but it seemed relevant so here it is.
I was trying to compile the Hello world program here: https://www.tutorialspoint.com/cplusplus-hello-world-program.

Fails on latest Zig:
```sh-session
$ zig version
0.10.0-dev.237+ada8e1713
$ zig run src/main.cpp -lc++
Segmentation fault (core dumped)
```

Works with `-target x86_64-linux-musl`:
```sh-session
$ zig run src/main.cpp -lc++ -target x86_64-linux-musl
Hello, World!
```

Works on Zig 0.8.1:
```sh-session
$ zig version
0.8.1
$ zig run src/main.cpp -lc++
Hello, World!
```

---

In case this matters, the `ldd` outputs of the binaries:
Master:
```sh-session
$ ldd a.out
[...]

Issue #10386, opened 2021-12-22
zig cc regression in 0.9.0


motiejus:
### Zig Version

0.9.0

### Steps to Reproduce

I am running on debian buster (glibc 2.28).

```
git clone https://github.com/protocolbuffers/protobuf/
mkdir -p protobuf/cmake/build
cd protobuf/cmake/build
CFLAGS=-g CC=~/dev/zig/build/zig\ cc\ -target\ x86_64-linux-gnu.2.28 CXX=~/dev/zig/build/zig\ c++\ -target\ x86_64-linux-gnu.2.28 cmake ..
make -j$(nproc) # parallel make has a tendency to fail with many cores, thus ...
make
./protoc -h | head -1
```

git bisect points to 19ca2415f26b07b399a60fd17a4cf85f290da0fd

```
19ca2415f26b07b399a60fd17a4cf85f290da0fd is the first bad commit
commit 19ca2415f26b07b399a60fd17a4cf85f290da0fd                                                                           
Author: Andrew Kelley <andrew@ziglang.org>                     
Date:   Wed Dec 15 00:22:38 2021 -0700                       
                                                                                                                     
    update glibc start files to 2.34                                  
                                                              
    This commit introduces tools/update_glibc.zig to update the start files       
    for next time.                                                                                                                                                   
                                                                
    Some notable changes in recent glibc:                      
                                                                                                                          
     * abi-note.S has been changed to abi-note.c but we resist the change to
       keep it easier to compile the start files.            
     * elf-init.c has been deleted upstream. Further testing should be done                                          
       to verify that binaries against glibc omitting elf-init.c still run
       properly on oldel glibc linux systems.                 
                                                                                  
    Closes #4926
```

### Expected Behavior

```
./protoc -h | head -1
Usage: ./protoc [OPTION] PROTO_FILES
```

### Actual Behavior

```
./protoc -h
Segmentation fault
```

Here is a backtrace:
```
user@motiejus:~/code/protobuf/cmake/build$ gdb ./protoc
GNU gdb (Debian 8.2.1-2+b3) 8.2.1
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
[...]

PR #10587, opened 2022-01-13
restore compatibility with glibc<=2.33 for global initializers

xxxbxxx:
Tentative fix for #10386 #10512

I've just conditionally restored some code from before 19ca2415f26b07b399a60fd17a4cf85f290da0fd 
to get _start to pass an init() fonction to __libc_start_main.

appears to work, but I've just tested with system glibc 2.33.  (didn't test for no regression in 2.34 for ex.)

Is this the correct thing to do? 

I've manually restored the old files, expecting this to be done 'once' and the files kept unchanged for future glibc versions. But maybe it is needed to update the tools/upadte_glibc?


andrewrk:
This looks reasonable, thanks! I'll add another commit with some minor organization and then merge.

motiejus:
Thanks for this! It unblocked bazel-zig-cc[1] to use post-0.9.0 zig. Please let me know your name/alias if you want better credit than this[2].

[1]: https://lists.sr.ht/~motiejus/bazel-zig-cc/%3C20220117151129.kdvzdw3w2dnuaxrj%40mtpad.i.jakstys.lt%3E
[2]: https://git.sr.ht/~motiejus/bazel-zig-cc#closed-upstream-issues

PR #10589, opened 2022-01-13
stage2 AArch64: split ldr/str into {ldr,str}_register and _immediate

joachimschmidt557:
6 less TODO comments

### DIFF
--- a/src/arch/aarch64/Emit.zig
+++ b/src/arch/aarch64/Emit.zig
@@ -112,12 +112,19 @@ pub fn emitMir(
             .strb_stack => try emit.mirLoadStoreStack(inst),
             .strh_stack => try emit.mirLoadStoreStack(inst),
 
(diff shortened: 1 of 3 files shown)


kubkon:
@joachimschmidt557 did you check that none of the changes to arm64 bits affected `zig ld` when targeting arm64? `zig ld` uses the instruction encoding internally :-)

joachimschmidt557:
@kubkon Yes, as this is mostly a change in MIR and the `bits.zig` has only been very slightly refactored, this shouldn't be an issue. I also confirmed this by running the test suite on an M1 mac.

kubkon:
> @kubkon Yes, as this is mostly a change in MIR and the `bits.zig` has only been very slightly refactored, this shouldn't be an issue. I also confirmed this by running the test suite on an M1 mac.

Excellent, thank you so much!

PR #4608, opened 2020-03-02
make std.os.getenvW case insensitive

ilmaria:
Fixes partially #4603. Fixing std.process.getEnvMap is NOT included in this commit.

Issue #4603, opened 2020-03-02
std.process.getEnvVarOwned is case sensitive on Windows


ilmaria:
On Windows, environment variables should be case insensitive but `std.process.getEnvVarOwned` returns only case sensitive matches. For example, on my Windows 10 the path variable is "Path" and not "PATH" so `std.ChildProcess.exec` doesn't work when the argument is an executable in the path.

For me this prints "error: FileNotFound" :
```
const std = @import("std");

pub fn main() !void {
    var arena = std.heap.ArenaAllocator.init(std.heap.page_allocator);
    const args = [_][]const u8{
        "ipconfig",
    };
    _ = try std.ChildProcess.exec(&arena.allocator, args[0..], null, null, std.math.maxInt(usize));
}
```

mikdusan:
related: `pub fn std.process.getEnvMap(allocator: *Allocator) !BufMap` needs a specialized `BufMap` for Windows with case-insensitive keys or maybe `getEnvMap()` can convert all env vars to uppercase but that seems hacky, and I'm not sure if losing case-retention will have consequences.

suirad:
Should probably be closed.

andrewrk:
The commit says:

> partially addresses #4603
> Fixing std.process.getEnvMap is NOT included in this commit.

PR #10595, opened 2022-01-15
os.getenvW: Fix case-insensitivity for Unicode env var names

squeek502:
Windows does Unicode-aware case-insensitive comparisons for environment variable names. Before, os.getenvW was only doing ASCII case-insensitivity. We can take advantage of [RtlEqualUnicodeString in NtDll](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-rtlequalunicodestring) to get the proper Unicode case insensitivity (note that this is what [ReactOS uses for its environment variable case insensitivity](https://github.com/reactos/reactos/blob/3fa57b8ff7fcee47b8e2ed869aecaf4515603f3f/sdk/lib/rtl/env.c#L560)).

This should be a more complete fix for part of https://github.com/ziglang/zig/issues/4603, and an improvement on https://github.com/ziglang/zig/pull/4608.

Example with the following test code and the env var `СЛАВЬСЯ` (set as all uppercase, but the test code uses all lowercase):

```zig
const std = @import("std");
const testing = std.testing;

test {
    const cyrillic = try std.process.getEnvVarOwned(testing.allocator, "славься");
    defer testing.allocator.free(cyrillic);

    std.debug.print("value is {s}\n", .{cyrillic});
}
```

Before this PR, the test would fail:

```
> set СЛАВЬСЯ=1
> echo %славься%
1
> zig test env.zig
Test [0/1] testTest [1/1] test ""... FAIL (EnvironmentVariableNotFound)
FAIL (EnvironmentVariableNotFound)
C:\Users\Ryan\Programming\Zig\zig\lib\std\process.zig:163:53: 0x7ff73c352430 in std.process.getEnvVarOwned (test.obj)
            break :blk std.os.getenvW(key_w) orelse return error.EnvironmentVariableNotFound;
                                                    ^
C:\Users\Ryan\Programming\Zig\tmp\env.zig:5:22: 0x7ff73c351c59 in test "" (test.obj)
    const cyrillic = try std.process.getEnvVarOwned(testing.allocator, "славься");
                     ^
```

After this PR, the test passes:

```
> set СЛАВЬСЯ=1
> echo %славься%
1
> zig test env.zig
Test [0/1] test ""... 1/1 test ""... value is 1
OK
All 1 tests passed.
```

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -1715,15 +1715,13 @@ pub fn getenvZ(key: [*:0]const u8) ?[]const u8 {
 
 /// Windows-only. Get an environment variable with a null-terminated, WTF-16 encoded name.
 /// See also `getenv`.
-/// This function first attempts a case-sensitive lookup. If no match is found, and `key`
-/// is ASCII, then it attempts a second case-insensitive lookup.
+/// This function performs a Unicode-aware case-insensitive lookup using RtlEqualUnicodeString.
 pub fn getenvW(key: [*:0]const u16) ?[:0]const u16 {
     if (builtin.os.tag != .windows) {
         @compileError("std.os.getenvW is a Windows-only API");
     }
     const key_slice = mem.sliceTo(key, 0);
     const ptr = windows.peb().ProcessParameters.Environment;
-    var ascii_match: ?[:0]const u16 = null;
     var i: usize = 0;
     while (ptr[i] != 0) {
         const key_start = i;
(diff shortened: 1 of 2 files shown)


squeek502:
Just realized that this contradicts what @andrewrk put in 8bf7cffe29b520cda094756309bca1d354af0d6b:

> It is not planned to support full Unicode case-insensitivity on Windows, and in fact relying on non-ASCII case-insensitive environment variables is fundamentally problematic.

Unsure if the approach in this PR avoids or succumbs to the problematic parts referenced in that message.

andrewrk:
Looks great. Big thanks for this change, and sorry for the delayed review!

PR #10596, opened 2022-01-15
stage2 ARM: store all function parameters to stack for unspecified cc

joachimschmidt557:
As we copy function parameters to the stack anyways for debugging purposes, this makes things easier.


andrewrk:
I tried playing around with this but ran into #10597 when I tried to look at the disassembly. How are you doing this without having `-femit-asm` working yet?

PR #10598, opened 2022-01-15
stage2: rename Isel back to Emit, and fix immediate casts in Emit for x86 64

kubkon:
* refactor handling of immediates in x86_64 backend
* rename Isel to Emit for x86_64
* implement signed compare

### DIFF
--- a/test/stage2/x86_64.zig
+++ b/test/stage2/x86_64.zig
@@ -1767,6 +1767,40 @@ pub fn addCases(ctx: *TestContext) !void {
                 \\    if (!ok) unreachable;
                 \\}
             , "");
+            case.addCompareOutput(
+                \\pub fn main() void {
+                \\    var x: u8 = undefined;
+                \\    const maybe_x = byPtr(&x);
+                \\    assert(maybe_x != null);
+                \\    maybe_x.?.* = 255;
+                \\    assert(x == 255);
+                \\}
+                \\
+                \\fn byPtr(x: *u8) ?*u8 {
+                \\    return x;
+                \\}
+                \\
+                \\fn assert(ok: bool) void {
+                \\    if (!ok) unreachable;
+                \\}
+            , "");
+            case.addCompareOutput(
+                \\pub fn main() void {
+                \\    var x: i8 = undefined;
+                \\    const maybe_x = byPtr(&x);
+                \\    assert(maybe_x != null);
+                \\    maybe_x.?.* = -1;
+                \\    assert(x == -1);
+                \\}
+                \\
+                \\fn byPtr(x: *i8) ?*i8 {
+                \\    return x;
+                \\}
+                \\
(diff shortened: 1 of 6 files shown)


g-w1:
Just curious, why was Isel.zig renamed back to Emit.zig?

test/stage2/x86_64.zig
```diff
@@ -1700,6 +1700,36 @@ pub fn addCases(ctx: *TestContext) !void {
                 \\    if (!ok) unreachable;
                 \\}
             , "");
+            case.addCompareOutput(
```
andrewrk, line 1703: why not make these behavior tests instead? the intent of these tests is to verify incremental compilation, but it doesn't look like this exercises any new incremental compilation code; it only tests behavior.

joachimschmidt557:
> Just curious, why was Isel.zig renamed back to Emit.zig?

According to [Wikipedia](https://en.wikipedia.org/wiki/Instruction_selection), instruction selection is not really what the Emit phase is doing. This might be confusing to contributors which expect to find the generally accepted definition of instruction selection.

The name `emit` has precedent in the `cranelift` part of wasmtime: https://github.com/bytecodealliance/wasmtime/blob/df37074218fe1e94a767e780e2138dde4dbd264c/cranelift/codegen/src/isa/aarch64/inst/emit.rs

andrewrk:
> According to [Wikipedia](https://en.wikipedia.org/wiki/Instruction_selection), instruction selection is not really what the Emit phase is doing. This might be confusing to contributors which expect to find the generally accepted definition of instruction selection.

This prompts the question, should "Codegen" be renamed to "ISel"?

test/stage2/x86_64.zig
kubkon: That's a great suggestion, thanks!

> > According to [Wikipedia](https://en.wikipedia.org/wiki/Instruction_selection), instruction selection is not really what the Emit phase is doing. This might be confusing to contributors which expect to find the generally accepted definition of instruction selection.
> 
> This prompts the question, should "Codegen" be renamed to "ISel"?

How about we bring it up at next stage2 meetup and discuss?

andrewrk (approved):

kubkon:
> > According to [Wikipedia](https://en.wikipedia.org/wiki/Instruction_selection), instruction selection is not really what the Emit phase is doing. This might be confusing to contributors which expect to find the generally accepted definition of instruction selection.
> 
> This prompts the question, should "Codegen" be renamed to "ISel"?

How about we discuss this at the upcoming stage2 meetup?

test/stage2/x86_64.zig
kubkon: I might land this as is and then do a sweep for any tests that should really end up in behavior test suite. Either way, why did I not think of it before - thanks again Andrew!

PR #10599, opened 2022-01-15
stage2 ARM: implement load for stack_offset + stack_argument_offset

joachimschmidt557:
1 more behavior test passing


andrewrk:
By the way, feel free to move files to the "passing for arm" section, and then disable each test within independently with a prefix like this:

```zig
if (@import("builtin").zig_backend == .stage2_arm) return error.SkipZigTest; // TODO
```

Then you could start enabling them one at a time, instead of one file at a time.

andrewrk (approved):

Issue #7298, opened 2020-12-04
Clearer documentation for `errdefer` in blocks


alunbestor:
[The current documentation example for `errdefer`](https://ziglang.org/documentation/master/#errdefer) only shows it being used at function scope, where its behaviour is pretty intuitive. But the behaviour in if/while/for etc blocks - i.e. that the `errdefer` goes out of scope at the end of the block - is subtle and could use more documentation. 

In particular, blocks silently break the pattern described in the documentation where "The deallocation code is always directly following the allocation code." A couple of footguns:

```zig
const std = @import("std");

const Error = error {
    FlagrantViolation,
};

// Example 1
fn errdeferAtFunctionScope(allocator: *std.mem.Allocator) !void {
    const ptr = try allocator.alloc(u8, 1);
    
    // Does not leak: the errdefer is at the same scope as the return.
    errdefer allocator.free(ptr);
    
    return Error.FlagrantViolation;
}

// Example 2
fn errdeferInBlock(allocator: *std.mem.Allocator) !void {
    {
        const ptr = try allocator.alloc(u8, 1);

        // Leaks: by the time the error is returned, errdefer has gone out of scope.
        errdefer allocator.free(ptr);
    }
    
    return Error.FlagrantViolation;
}

// Example 3
fn errdeferInLoop(allocator: *std.mem.Allocator) !void {
    var count: usize = 0;
    
    while (true) {
        const ptr = try allocator.alloc(u8, 1);

        // Leaks on the second iteration of the loop: 
        // the errdefer from the first iteration will have gone out of scope.
        errdefer allocator.free(ptr);
        
        count += 1;
        if (count == 2) {
            return Error.FlagrantViolation;
        }
    }
}

test "errdefer at function scope" {
    // Passes
    std.testing.expectError(Error.FlagrantViolation, errdeferAtFunctionScope(std.testing.allocator));
}

test "errdefer in block" {
    // Fails with a leak
    std.testing.expectError(Error.FlagrantViolation, errdeferInBlock(std.testing.allocator));
}

test "errdefer in loop" {
    // Fails with a leak
    std.testing.expectError(Error.FlagrantViolation, errdeferInLoop(std.testing.allocator));
}
```

The specific case that bit me was Example 3: repeatedly allocating in a failable loop, and assuming that I could keep an `errdefer` next to the `alloc` call and undo all previous allocations that way. It makes sense in retrospect that it doesn't work that way, but some scope examples in the docs could help users arrive at that mental model.

[...]

PR #10607, opened 2022-01-16
Added documentation for common mistakes in errdefer scoping

Exonorid:
This PR adds a subsection to the Error Union Type section of the documentation to inform readers about common mistakes caused by errdefer scoping (e.g. allocated memory leaving the scope of an errdefer and being leaked) to hopefully save some people from shooting themselves in the foot. In commit e9c7987, I added a header to separate this from the errdefer subsection because I felt it was distinct enough, but it's completely optional imo

This PR is intended to resolve issue #7298

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -5297,6 +5297,97 @@ fn createFoo(param: i32) !Foo {
       the verbosity and cognitive overhead of trying to make sure every exit path
       is covered. The deallocation code is always directly following the allocation code.
       </p>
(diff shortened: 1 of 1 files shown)


Vexu (approved):

andrewrk:
Reverted in 3ce6de87657e67da358e9c7869a87c61cc9f1305. This kind of thing does not belong in the language reference.

linusg:
#7298 should probably be reopened then.

Issue #10609, opened 2022-01-16
add segfault handling support for macOS


andrewrk:
In debug builds of Zig, for most operating systems, we have it set up so that segfaults print a stack trace. However, macOS is not yet supported!

https://github.com/ziglang/zig/blob/97063efb69d11bfe7210555948b1f59ab6fb9d08/lib/std/debug.zig#L1574-L1612

PR #10615, opened 2022-01-17
Implement segfault handler for macOS x86_64

schmee:
Fixes part of https://github.com/ziglang/zig/issues/10609. It is still not implemented for ARM, but in a few weeks I will have access to an M1 MacBook and can hopefully continue from there.

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -1574,6 +1574,7 @@ fn getDebugInfoAllocator() mem.Allocator {
 /// Whether or not the current target can print useful debug information when a segfault occurs.
 pub const have_segfault_handling_support = switch (native_os) {
     .linux, .netbsd, .solaris => true,
+    .macos => native_arch == .x86_64,
     .windows => true,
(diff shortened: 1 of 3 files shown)


lib/std/debug.zig
```diff
@@ -1601,7 +1602,7 @@ pub fn attachSegfaultHandler() void {
         return;
     }
     var act = os.Sigaction{
-        .handler = .{ .sigaction = handleSegfaultLinux },
+        .handler = .{ .sigaction = handleSegfaultLinuxAndMacos },
```
andrewrk, line 1605:
```suggestion
        .handler = .{ .sigaction = handleSegfaultPosix },
```

andrewrk (changes requested):
One minor rename request for ya.

cc @SpexGuy - why do we have this code duplicated in src/crash_report.zig? looks like some more code should be shared, no?

andrewrk (approved):
Thanks! Looking forward to this working :)

PR #10616, opened 2022-01-17
stage2: implement airArrayToSlice for x86_64

kubkon:
* implement `genSetStack` for `ptr_stack_offset`
* handle `ptr_add`
* implement storing from register into pointer in register
* split alignment and array tests into those that pass on x86_64 and those that do not
* pass more tests on x86_64


test/behavior.zig
```diff
@@ -17,195 +17,201 @@ test {
...
-        _ = @import("behavior/try.zig");
-        _ = @import("behavior/undefined.zig");
-        _ = @import("behavior/underscore.zig");
-        _ = @import("behavior/usingnamespace.zig");
-        _ = @import("behavior/void.zig");
-        _ = @import("behavior/while.zig");
+    if (builtin.zig_backend != .stage2_arm) {
+        // Tests that pass for stage1, llvm backend, C backend, wasm backend and x86_64 backend.
+        _ = @import("behavior/align_simple.zig");
+        _ = @import("behavior/array_simple.zig");
```
andrewrk, line 23: To avoid all this churn in test/behavior.zig I propose that we start doing this strategy:

https://github.com/ziglang/zig/commit/0f3938e498624868a951423a0b1ca9749301130a

andrewrk (approved):

test/behavior.zig
kubkon: Sounds great!

src/arch/x86_64/CodeGen.zig
```diff
@@ -3618,10 +3631,21 @@ fn airBitCast(self: *Self, inst: Air.Inst.Index) !void {
...
 fn airArrayToSlice(self: *Self, inst: Air.Inst.Index) !void {
     const ty_op = self.air.instructions.items(.data)[inst].ty_op;
-    const result: MCValue = if (self.liveness.isUnused(inst))
-        .dead
-    else
-        return self.fail("TODO implement airArrayToSlice for {}", .{self.target.cpu.arch});
+    const ptr_ty = self.air.typeOf(ty_op.operand);
+    const ptr = try self.resolveInst(ty_op.operand);
+    const result: MCValue = if (self.liveness.isUnused(inst)) .dead else blk: {
+        switch (ptr) {
```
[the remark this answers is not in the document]
kubkon: Good point! Thanks!

test/behavior.zig
kubkon: Done in ab204f8 - have a look and lemme know if this is what you had in mind @andrewrk!

andrewrk (approved):

src/arch/x86_64/CodeGen.zig
```diff
@@ -3125,7 +3137,6 @@ fn setRegOrMem(self: *Self, ty: Type, loc: MCValue, val: MCValue) !void {
 fn genSetStack(self: *Self, ty: Type, stack_offset: u32, mcv: MCValue) InnerError!void {
     switch (mcv) {
         .dead => unreachable,
-        .ptr_stack_offset => unreachable,
```
joachimschmidt557, line 3128: Hmm, as far as I know, `genSetStack`in a `ptr_stack_offset` shouldn't be possible, right? Just like `genSetStack`in an `immediate`. `ptr_stack_offset` represents not the value at that stack offset, but the stack offset itself.

What exactly triggers a `genSetStack` for `ptr_stack_offset`?

joachimschmidt557: > Just like `genSetStack`ing an `immediate`.

Oh, nevermind. That's perfectly possible. But shouldn't the code paths for `stack_offset` and `ptr_stack_offset` be different?

kubkon: So the codepaths share the bit until abi_size is within 8 bytes (for ptr to stack offset, that's obv always 8 bytes), but I take your point that having this more explicit will not be as confusing to newcomers. I will change it so that it's easier to reason about, and when we have enough implemented, we can start trimming it down and refactoring.

PR #10617, opened 2022-01-17
translate-c: reserve space for bitfield.

TwoClocks:
This patch attempts to reserve the correct amount of space in a record containing a bitfield but does not implement bitfields. So this :
```c
struct my_struct {
    unsigned a: 1;
    unsigned b: 28;
};
```
becomes
```zig
pub const struct_my_struct = extern struct {
    bitfield0: c_uint,
};
```

The motivation was working w/ DPDK, which has tons of bitfields in the hierarchy, and I didn't want to fix them up all by hand. This works.

There are likely more reasons not to accept this patch than to accept it. They include:
This only works if the C was intended to be compiled w/ clang.
Any generated zig code that reference the individual bitfields is now invalid.
likely x86 specific.
it's a hack.

If there is any interest in this PR, things that would need to be done:
make it opt in. Not sure how to get params to `translate-c`
Make the code in `test/transte_c.zig` aware of the param, and have different tests for this option.
other?
Happy to do this work with some guidance if there is interest.

FWIW : For these deep DPDK structs with bitfields, the records generated are the same size as clang generates.

### DIFF
--- a/src/clang.zig
+++ b/src/clang.zig
@@ -476,6 +476,12 @@ pub const FieldDecl = opaque {
     pub const isBitField = ZigClangFieldDecl_isBitField;
     extern fn ZigClangFieldDecl_isBitField(*const FieldDecl) bool;
 
+    pub const getBitWidthValue = ZigClangFieldDecl_getBitWidthValue;
+    extern fn ZigClangFieldDecl_getBitWidthValue(*const FieldDecl, *const ASTContext) c_uint;
+
+    pub const isZeroLengthBitField = ZigClangFieldDecl_isZeroLengthBitField;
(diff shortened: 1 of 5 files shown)


TwoClocks:
I got a few questions about this on discord.

This patch is very conservative in what it will try to reserve space for. If the criteria isn't met, it's demoted to opaque, like the old behavior.

- The field type must be unsigned.
- You must define enough bits so less than 8 bits remain in the type. (e.g. if the type is 32 bits, you must define at least 25).
- A field can not straddle a type, even if the next type is also a bitfield.

I think the above are all valid in C, but they make layout hard. So the record will be demoted.

So it's only useful for well-behaved usages of bitfields. The structs in "ip.h" do not work, for example.

PR #10619, opened 2022-01-18
std: add meta.DeclEnum

nektro:
counterpart to `std.meta.FieldEnum` but for declarations

### DIFF
--- a/lib/std/meta.zig
+++ b/lib/std/meta.zig
@@ -594,6 +594,47 @@ test "std.meta.FieldEnum" {
     try expectEqualEnum(enum { a, b, c }, FieldEnum(union { a: u8, b: void, c: f32 }));
 }
 
+pub fn DeclEnum(comptime T: type) type {
+    const fieldInfos = std.meta.declarations(T);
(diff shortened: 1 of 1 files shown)


daurnimator:
Do you have an example use case for this?

Would you be able to write a doc-comment?

nektro:
<https://github.com/nektro/limitless/blob/master/tools/os-list-arches.zig>
<https://github.com/nektro/limitless/blob/master/tools/shared.zig>

made a PR because i saw @Jarred-Sumner also asked for it and so imagined others might need something like it as well

nektro:
otherwise i can keep it external and put it in https://github.com/nektro/zig-extras

ehaas:
> Do you have an example use case for this?

[Here's an example](https://github.com/Vexu/arocc/blob/aaefccfd30dc5845a68825d1c32318634005f645/src/Diagnostics.zig#L40) from arocc - the idea is to create a tagged union at comptime, with behavior declaratively controlled by ... declarations ... in the individual structs. This avoids having to create a giant switch for each thing that can vary per tag (active union field, default diagnostic level, error message format string, relevant compiler option flag, etc)

PR #10621, opened 2022-01-18
ignore target lib dirs when invoked with -feach-lib-rpath

Jan200101:
since the native paths are expected to be linked by default adding them to the rpath is pointless.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -2077,10 +2077,6 @@ fn buildOutputType(
             clang_argv.appendAssumeCapacity(framework_dir);
             framework_dirs.appendAssumeCapacity(framework_dir);
         }
-
-        for (paths.lib_dirs.items) |lib_dir| {
-            try lib_dirs.append(lib_dir);
-        }
         for (paths.rpaths.items) |rpath| {
             try rpath_list.append(rpath);
         }


src/link/Elf.zig
```diff
@@ -1631,7 +1631,20 @@ fn linkWithLLD(self: *Elf, comp: *Compilation) !void {
         if (self.base.options.each_lib_rpath) {
             var test_path = std.ArrayList(u8).init(self.base.allocator);
             defer test_path.deinit();
+
+            const target_info = try std.zig.system.NativeTargetInfo.detect(self.base.allocator, .{});
```
Jan200101, line 1635: NativeTargetInfo.detect takes a CrossTarget as the second argument. I could probably rebuild it from scratch using what is available, but there should be a better way of doing this that I am not aware of.

src/link/Elf.zig
```diff
@@ -1631,7 +1631,20 @@ fn linkWithLLD(self: *Elf, comp: *Compilation) !void {
...
             for (self.base.options.lib_dirs) |lib_dir_path| {
+                var is_target_lib_dir = false;
+                for (paths.lib_dirs.items) |lib_dir| {
+                    if (mem.eql(u8, lib_dir, lib_dir_path)) {
+                        is_target_lib_dir = true;
+                    }
+                }
+                if (is_target_lib_dir) {
+                    continue;
+                }
```
Jan200101, line 1647: There is probably a way to do this better.

Jan200101: ended up force pushing twice because I was thrown for a loop around this.

I think this should be the best implementation for this
```zig
                var is_target_lib_dir = false;
                for (paths.lib_dirs.items) |lib_dir| {
                    if (mem.eql(u8, lib_dir, lib_dir_path)) {
                        is_target_lib_dir = true;
                        break;
                    }
                }
                if (is_target_lib_dir) {
                    continue;
                }
```

it might be worth to use labeled loops for this, but I am unsure

src/link/Elf.zig
```diff
@@ -1631,7 +1631,22 @@ fn linkWithLLD(self: *Elf, comp: *Compilation) !void {
         if (self.base.options.each_lib_rpath) {
             var test_path = std.ArrayList(u8).init(self.base.allocator);
             defer test_path.deinit();
+
+            const cross_target = std.zig.CrossTarget.fromTarget(target);
+            const target_info = try std.zig.system.NativeTargetInfo.detect(self.base.allocator, cross_target);
```
kubkon, line 1636: Using `std.zig.system.NativeTargetInfo.detect` or related does not belong in the linker code and should only ever be invoked in `main.zig`. Please have a look at how we handle detection of macOS SDK for example (see [this line](https://github.com/ziglang/zig/blob/04f3d9301798e41ac8272822328914073e01a6ab/src/main.zig#L2064)). If you want to filter the `native_paths` based on the detected native paths, this should be done in `main.zig` when passing the info to `Compilation.zig`. Lemme know if that makes sense, or if you need any guidance or clarification.

kubkon (changes requested):

src/link/Elf.zig
Jan200101: Yeah, apologies for that.

At the time I couldn't find the starting point I was looking for and took this as the best location to get something off the ground.

I've been delayed a bit from working on this again, but I've now looked through the code and figured its better to remove what was the cause of this instead of adding new code

PR #10628, opened 2022-01-18
MultiArrayList: Fix error when struct is 0 sized

### LINKED ISSUE #10618: AutoArrayHashMap: Alignment error when key type is single-member enum and value type is void
### Zig Version

0.10.0-dev.295+f19b5ecf4

### Steps to Reproduce

This occurs when the `AutoArrayHashMap` key type is an enum with a single member **and** the value type is `void`. 

This construct is unlikely to be seen in real-world code, but occurred in a test-case where the hashmap was part of a larger structure taking an enum type argument.

```zig
const std = @import("std");

test "alignment error" {
    const E = enum { a };
    var map = std.AutoArrayHashMap(E, void).init(std.testing.allocator);
    defer map.deinit();
}
```

The compile error disappears if a) another enum member is added, or b) a value type is specified.

### Expected Behavior

Successful compilation

### Actual Behavior

Fails with

```bash
build/lib/zig/std/multi_array_list.zig:19:25: error: alignment must be >= 1
        bytes: [*]align(@alignOf(S)) u8 = undefined,
                        ^
build/lib/zig/std/multi_array_list.zig:19:9: note: referenced here
        bytes: [*]align(@alignOf(S)) u8 = undefined,
        ^
./repro.zig:5:44: note: referenced here
    var map = std.AutoArrayHashMap(E, void).init(std.testing.allocator);
```

riverbl:
Also fixes error with ArrayHashMap when both key and value are 0 sized.

Fixes #10618.

### DIFF
--- a/lib/std/array_hash_map.zig
+++ b/lib/std/array_hash_map.zig
@@ -2215,6 +2215,35 @@ test "auto store_hash" {
     try testing.expect(meta.fieldInfo(HasExpensiveEqlUn.Data, .hash).field_type != void);
 }
 
+test "0 sized key" {
+    var map = AutoArrayHashMap(u0, i32).init(std.testing.allocator);
+    defer map.deinit();
+
+    try testing.expectEqual(map.get(0), null);
+
+    try map.put(0, 5);
+    try testing.expectEqual(map.get(0), 5);
+
+    try map.put(0, 10);
+    try testing.expectEqual(map.get(0), 10);
+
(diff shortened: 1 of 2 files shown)


andrewrk:
Reverted in 913d61ebb95567fe67b0bfad99694892aac841e2. I'd still like to merge the PR, but I want to solve it a different way - with an improvement to the language/compiler rather than a userland workaround such as this.

PR #10635, opened 2022-01-19
stage2: fix passing arguments on the stack on x86_64

kubkon:
* push the arguments in reverse order
* add logic for pushing args of any abi size to stack - very similar to `genSetStack` however uses `.rsp` as the base register
* increment and decrement `.rsp` if we called a function with args on the stack in `airCall`
* add logic for recovering args from the caller's stack in the callee
* add tweak to please Apple re stack alignment when calling extern functions - this is more like a temp hack than anything else as I think the mechanism we use for adjusting the stack when pushing function args onto the stack could/should be reused - i.e., we should just align the stack before each call and then reset the `.rsp` when we return from the callee rather than relying on the current hack

### DIFF
--- a/src/arch/x86_64/CodeGen.zig
+++ b/src/arch/x86_64/CodeGen.zig
@@ -512,9 +512,11 @@ fn gen(self: *Self) InnerError!void {
             });
         }
         while (self.stack_args_relocs.popOrNull()) |index| {
+            // TODO like above, gotta figure out the alignment shenanigans for macOS, etc.
+            const adjustment = if (self.target.isDarwin()) 2 * stack_adjustment else stack_adjustment;
             // +16 bytes to account for saved return address of the `call` instruction and
             // `push rbp`.
-            self.mir_instructions.items(.data)[index].imm += stack_adjustment + aligned_stack_end + 16;
+            self.mir_instructions.items(.data)[index].imm += adjustment + aligned_stack_end + 16;
         }
     } else {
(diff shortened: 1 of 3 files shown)


src/arch/x86_64/CodeGen.zig
```diff
@@ -4252,3 +4343,7 @@ fn registerAlias(reg: Register, size_bytes: u32) Register {
         else => unreachable,
     }
 }
+
+fn abiSize(self: *Self, ty: Type) u64 {
+    return if (ty.isSlice()) 16 else ty.abiSize(self.target.*);
+}
```
andrewrk, line 4349: This should be fixed in type.zig, not hacked here in this x86 codegen file.

andrewrk (changes requested):

src/arch/x86_64/CodeGen.zig
kubkon: Oh, this is a bug in type.zig? Oops, my bad! I thought this was by design. Lemme provide a fix!

andrewrk:
If you rebase on top of 7c6981e0c0f2412c49457a900e5a6a4db96be2e8 then you can remove all the slice workaround stuff.

src/arch/x86_64/CodeGen.zig
kubkon: Done!

PR #10641, opened 2022-01-19
glibc: fix passing of __GNU_MINOR__

### LINKED ISSUE #7508: about __GLIBC__  and __GLIBC_MINOR__
Zig use glibc 2.32. 

Input v.c
```v.c
#include <stdio.h>

int main()
{
  printf("libc %d.%d\n", __GLIBC__, __GLIBC_MINOR__);
  printf("gnuc %d.%d\n", __GNUC__, __GNUC_MINOR__);
  return 0;
}
```

And compile it with `zig cc -target x86_64-linux-gnu.2.15 v.c`, we get
```
libc 2.32
gnuc 4.2
```

When compile it with `zig cc -target x86_64-linux-gnu.2.15 v.c`, we hope `__GLIBC__` , `__GLIBC_MINOR__` change 2 and 15.
So need output is
```
libc 2.15
gnuc 4.2
```

but got
```
libc 2.32
gnuc 4.2
```

Will Zig team to consider this?

### LINKED ISSUE #10614: `zig c++`: Undefined symbol `pthread_cond_clockwait`
### Zig Version

0.10.0-dev.290+3901b6fb0

### Steps to Reproduce

Try to compile the following example with `zig c++`:

```cpp
#include <condition_variable>
#include <mutex>

int main()
{
    std::mutex mutex;
    std::unique_lock<std::mutex> lock(mutex);

    std::condition_variable cv;
    cv.wait_for(lock, std::chrono::seconds(1));

    return 0;
}
```

### Expected Behavior

The example should compile.

### Actual Behavior

Everything works fine with default settings / musl. When trying to compile with glibc (`zig c++ -target x86_64-linux-gnu main.cpp`), we get an undefined symbol.

```
ld.lld: error: undefined symbol: pthread_cond_clockwait
```

The symbol in question is guarded by `__USE_GNU`, so I guess glibc should be compiled with `_GNU_SOURCE`, but is not.

saierd:
This was originally introduced in 4d48948b526337947ef59a83f7dbc81b70aa5723 but broken immediately afterwards in c8af00c66e8b6f62e4dd6ac4d86a3de03e9ea354.

In my use case, this fixes an undefined symbol when compiling libc++ with older versions of glibc (see #10614). libc++ checks for the existence of features using `__GLIBC_PREREQ`, which always passed since `__GNU_MINOR__` was hardcoded in the header.

I would also like to add a test for this, but I only found a single test that compiles a C++ file in `test/standalone/c_compiler`. Are you interested in more small tests which try to compile code with different versions of glibc?

Closes #10614 and if I am not mistaken the original change already closed #7508.

### DIFF
--- a/lib/libc/include/generic-glibc/features.h
+++ b/lib/libc/include/generic-glibc/features.h
@@ -477,7 +477,7 @@
 /* Major and minor version number of the GNU C library package.  Use
    these macros to test for features in specific releases.  */
 #define	__GLIBC__	2
-#define	__GLIBC_MINOR__	34
+/* Zig patch: we pass `-D__GLIBC_MINOR__=XX` depending on the target. */
 
 #define __GLIBC_PREREQ(maj, min) \
 	((__GLIBC__ << 16) + __GLIBC_MINOR__ >= ((maj) << 16) + (min))


andrewrk:
Thank you for this!

> Are you interested in more small tests which try to compile code with different versions of glibc?

Such tests would be much appreciated.

PR #10646, opened 2022-01-20
Fix overflow in std.math.isNormal when applied to -Inf or a negative NaN

gereeter:
These cases had all the top bits set, so adding would cause an overflow, despite the high bits being immediately masked out.

I'm having trouble with fully building and testing my local copy of `zig`, so apologies if I've made basic errors, but I figured this was a simple enough fix that it more worth submitting an untested PR than opening an issue; I also tested outside of `std`.


Vexu:
Could you add a test for negative NaN as well?

PR #10648, opened 2022-01-20
stage1: remove the "referenced here" error note

andrewrk:
It's generally noise. The parts where it is useful will need to be
redone to not be annoying for the general case.


Vexu:
Alternative to #9634?

PR #10650, opened 2022-01-20
Add `process.EnvMap`, a cross-platform environment variable map

### LINKED ISSUE #4603: std.process.getEnvVarOwned is case sensitive on Windows

### LINKED ISSUE #10561: `Builder.findProgram()` fails on Windows when env path is `Path` not `PATH`.
### Zig Version

0.10.0-dev.207+c08b190c6

### Steps to Reproduce

Here's a repro on my box.

```
C:\tmp>cat findDir.zig
const std = @import("std");
const Builder = std.build.Builder;

test "builder.findProgram fails on windows" {
    var arena = std.heap.ArenaAllocator.init(std.testing.allocator);
    defer arena.deinit();
    var b = try Builder.create(
        arena.allocator(),
        "test",
        "test",
        "test",
        "test",
    );

    _ = try b.findProgram(&[_][]const u8{
        "calc",
    }, &[_][]const u8{});
}

C:\tmp>set | rg -i Path=
Path=C:\SDK\Conda;.......
```

Basically this happens when env is `Path` not `PATH` on windows. 

I can track down it's this particular line checking for `PATH`
<https://github.com/ziglang/zig/blob/master/lib/std/build.zig#L1140>

It's trivia to fix but not sure how should I do it though. Should I check for path case insensitive on windows?
Thanks in advance!

### Expected Behavior

`Builder.findProgram` should work when it's `Path` on Windows.

### Actual Behavior

`Builder.findProgram` failed to find when it's `Path` on Windows.

squeek502:
EnvMap provides the same API as the previously used BufMap (besides `putMove` and `getPtr`), so usage sites of `getEnvMap` can usually remain unchanged.

For non-Windows, EnvMap is a wrapper around BufMap. On Windows, it uses a new EnvMapWindows to handle some Windows-specific behavior:

- Lookups use Unicode-aware case insensitivity (but `get` cannot return an error because EnvMapWindows has an internal buffer to use for lookup conversions)
- Canonical names are returned when iterating the EnvMap

Fixes #10561, closes #4603 (alongside #10595)

### DIFF
--- a/lib/std/unicode.zig
+++ b/lib/std/unicode.zig
@@ -712,6 +712,29 @@ pub fn utf8ToUtf16Le(utf16le: []u16, utf8: []const u8) !usize {
     return dest_i;
 }
 
+pub fn utf8ToUtf16LeWriter(writer: anytype, utf8: []const u8) !usize {
+    var src_i: usize = 0;
+    var bytes_written: usize = 0;
+    while (src_i < utf8.len) {
+        const n = utf8ByteSequenceLength(utf8[src_i]) catch return error.InvalidUtf8;
+        const next_src_i = src_i + n;
+        const codepoint = utf8Decode(utf8[src_i..next_src_i]) catch return error.InvalidUtf8;
+        if (codepoint < 0x10000) {
+            const short = @intCast(u16, codepoint);
+            try writer.writeIntLittle(u16, short);
+            bytes_written += 2;
+        } else {
+            const high = @intCast(u16, (codepoint - 0x10000) >> 10) + 0xD800;
+            const low = @intCast(u16, codepoint & 0x3FF) + 0xDC00;
+            try writer.writeIntLittle(u16, high);
+            try writer.writeIntLittle(u16, low);
+            bytes_written += 4;
+        }
+        src_i = next_src_i;
+    }
+    return bytes_written;
+}
+
 test "utf8ToUtf16Le" {
     var utf16le: [2]u16 = [_]u16{0} ** 2;
     {
--- a/doc/docgen.zig
+++ b/doc/docgen.zig
@@ -1708,7 +1708,7 @@ fn genHtml(
     }
 }
 
-fn exec(allocator: Allocator, env_map: *std.BufMap, args: []const []const u8) !ChildProcess.ExecResult {
+fn exec(allocator: Allocator, env_map: *process.EnvMap, args: []const []const u8) !ChildProcess.ExecResult {
     const result = try ChildProcess.exec(.{
         .allocator = allocator,
         .argv = args,
@@ -1732,7 +1732,7 @@ fn exec(allocator: Allocator, env_map: *std.BufMap, args: []const []const u8) !C
     return result;
 }
 
-fn getBuiltinCode(allocator: Allocator, env_map: *std.BufMap, zig_exe: []const u8) ![]const u8 {
+fn getBuiltinCode(allocator: Allocator, env_map: *process.EnvMap, zig_exe: []const u8) ![]const u8 {
     const result = try exec(allocator, env_map, &[_][]const u8{ zig_exe, "build-obj", "--show-builtin" });
     return result.stdout;
 }
--- a/lib/std/buf_map.zig
+++ b/lib/std/buf_map.zig
@@ -82,7 +82,7 @@ pub const BufMap = struct {
     }
 
     /// Returns the number of KV pairs stored in the map.
-    pub fn count(self: BufMap) usize {
+    pub fn count(self: BufMap) BufMapHashMap.Size {
         return self.hash_map.count();
     }
 
--- a/lib/std/buf_map.zig
+++ b/lib/std/buf_map.zig
@@ -9,7 +9,7 @@ const testing = std.testing;
 pub const BufMap = struct {
     hash_map: BufMapHashMap,
 
-    const BufMapHashMap = StringHashMap([]const u8);
+    pub const BufMapHashMap = StringHashMap([]const u8);
 
     /// Create a BufMap backed by a specific allocator.
     /// That allocator will be used for both backing allocations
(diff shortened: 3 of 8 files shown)


squeek502:
cc @marler8997 since he was interested in this.

lib/std/unicode.zig
```diff
@@ -712,6 +712,29 @@ pub fn utf8ToUtf16Le(utf16le: []u16, utf8: []const u8) !usize {
...
+            const high = @intCast(u16, (codepoint - 0x10000) >> 10) + 0xD800;
+            const low = @intCast(u16, codepoint & 0x3FF) + 0xDC00;
+            try writer.writeIntLittle(u16, high);
+            try writer.writeIntLittle(u16, low);
+            bytes_written += 4;
+        }
+        src_i = next_src_i;
+    }
+    return bytes_written;
+}
```
squeek502, line 736: This is something I'm the most unsure about as it might be a little strange. I needed something that:

- Is able to return an error on overflowing the provided buffer
- And is able to use exactly the buffer provided (i.e. no over-allocations)

None of the existing utf8 -> utf16le conversion functions provide both of those:

- `utf8ToUtf16LeWithNull` uses an ArrayList so providing a `FixedBufferAllocator` to it can fail even if the converted string would fit in the buffer because it can over-allocate when resizing
- `utf8ToUtf16Le` asserts that the buffer is large enough

Using a `Writer` got me both of my requirements essentially for free, but I'm not sure if it's out-of-place here or not.

marler8997:
The EnvMap for windows seems like it might be storing more than necessary.  I was thinking you could just store the original key with the original casing, then modify the hash function to ignore case.

squeek502:
> The EnvMap for windows seems like it might be storing more than necessary. I was thinking you could just store the original key with the original casing, then modify the hash function to ignore case.

This is admittedly retroactive justification, but I'm not so sure that this would be an objective improvement.

If we store just the canonical names as UTF-16:

- For hashing, we'd need to convert keys to uppercase during `get` (this is already the case)
- For `eql`, we'd need to use `RtlEqualUnicodeString` which [iterates both strings, does uppercase-conversions for each u16 and compares them](https://doxygen.reactos.org/df/d18/sdk_2lib_2rtl_2unicode_8c.html#a56e217e09c579d0681c35c732a3572bc) (this is new, currently that case conversion of the lookup key done once in `put` instead)
- For iterating, we'd need to convert names to UTF-8 (this is new, currently this is done once in `put` instead)

If we store just the canonical names as UTF-8:

- For hashing, we'd need to convert keys to uppercase UTF-16 and then back to UTF-8 I think
- For `eql`, we'd need to convert both to UTF-16 and then use `RtlEqualUnicodeString` I think
- For iterating, we'd need nothing special

So, by storing more than we need to, we save some processing during `get` and iterating, which to me seems like a potentially worthwhile tradeoff. I could definitely be wrong about that, though.

squeek502:
In attempting to implement a version that only stores the names in one representation, I think my previous comment was correct--they introduce some new problems and I couldn't come up with any solutions I was happy with.

- If the names are only stored in the canonical case with a UTF-16 representation, then `Iterator.next` needs to allocate to be able to return them as UTF-8, which introduces the possibility of failure which doesn't seem ideal. To get around the allocation we'd have to do something like keep a UTF-8 buffer around that's big enough for all the keys, but that'd also mean doing dummy conversions of the UTF-16 names to UTF-8 on each `put` to know if that buffer needs to be expanded, which also doesn't seem ideal.
- If the names are only stored as the canonical case with a UTF-8 representation, then we remove the allocation-in-iterator problem, but start needing to do UTF-8 -> UTF-16 conversions for both strings in the hash map's `eql` function. This seems like it'd have a performance impact worth avoiding, but I haven't investigated this fully.

I think it's possible there's a more efficient implementation possible that has nice solutions to the problems I perceive here, but I think that could be done later, and would need some benchmarking to validate the extent of the improvement.

marler8997:
> So, by storing more than we need to, we save some processing during get and iterating, which to me seems like a potentially worthwhile tradeoff. I could definitely be wrong about that, though.

One thing us programmers learn about modern hardware the more experienced we get is that memory is much more expensive that re-computing things.  It's not the instructions that are slow, it's the cache misses that really slow us down.

squeek502:
Fair enough, I don't think that this should block an initial implementation though.

marler8997:
Here's more what I was thinking, but I left a couple of TODOs in there to implement the UNICODE support: https://github.com/marler8997/zig/commit/bef8bfb9339269f3a2041502dfb09a2e8b94f1c9

For the UNICODE support, during the hash we just need to use a utf8 iterator (I think there is one in std).

squeek502:
I don't think those TODOs are solvable without an implementation more like mine, as Andrew has said that there are no plans to ever support including the Unicode data necessary for something like case conversion in the standard library. Using `RtlUpcastUnicodeString`/`RtlEqualUnicodeString` allows us to get that support without implementation case conversion ourselves, but means that we need UTF-16 strings to pass to them.

marler8997:
I see your "Andrew twitter post" and I raise you an "Implementation" :)

https://github.com/marler8997/zig/commit/655ecb1c768f61b9acc50ae5ae09b81716cb5e6d

squeek502:
Nice. Might want to check that codepoints >= `0x10000` don't cause problems, as it seems like they'd overflow the u16. I believe It'd be safe to just compare those codepoints directly without conversion, since that's what `RtlUpcaseUnicodeString`/`RtlEqualUnicodeString` does I believe (they just ignore surrogate pairs).

squeek502:
Closing in favor of https://github.com/ziglang/zig/pull/10796 which seems to be preferred.

PR #10652, opened 2022-01-20
Add test coverage for recent glibc version problems with `zig cc`

saierd:
Adds test coverage for #10641 and the related issues. Both tests fail without that fix.

I am not sure whether I integrated this into the overall test system correctly, since I only ran the tests individually.

### DIFF
--- a/test/standalone.zig
+++ b/test/standalone.zig
@@ -60,6 +60,9 @@ pub fn addCases(cases: *tests.StandaloneContext) void {
         .build_modes = true,
         .cross_targets = true,
     });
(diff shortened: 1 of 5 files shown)


andrewrk:
Thanks for the effort. Looks like this was never passing the CI, and now is abandoned. Feel free to re-open a fresh PR if you want to continue the effort.

saierd:
Yeah. Both tests ran when I executed them individually. But I didn't find out what the problem in CI was, since I don't really know the zig build system.

PR #10654, opened 2022-01-20
c backend: Implement aligning fields and local/global variables

Hejsil:
There are some restrictions here.

- We either need C11 or a compiler that supports the aligned attribute
- We cannot provide align less than the type's natural C alignment.


src/link/C/zig.h
```diff
@@ -26,6 +26,16 @@
...
 #endif
 
+#if __STDC_VERSION__ >= 201112L
+#include <stdalign.h>
+#define ZIG_ALIGN(alignment) alignas(alignment)
+#elif defined(__GNUC__)
+#define ZIG_ALIGN(alignment) __attribute__((aligned(alignment)))
+#else
+/* TODO: Can we somehow align stack variables without help from the compiler? */
+#define ZIG_ALIGN(alignment)
```
andrewrk, line 36: Here's an idea:

```suggestion
#define ZIG_ALIGN(alignment) zig_compile_error("the C compiler being used does not support aligning variables")
```

And then don't define `zig_compile_error` anywhere (that's how it will be a compile error).

What do you think?

Hejsil: Sounds good to me.

PR #10657, opened 2022-01-21
fix sigaction double panic

### LINKED ISSUE #8357: Infinite loop if sigaction fails on startup
`callMainWithArgs` calls `std.debug.maybeEnableSegfaultHandler`, which calls `std.debug.attachSegfaultHandler`.

`attachSegfaultHandler` calls `os.sigaction`, which if fails panics with `unreachable`.

`std.builtin.default_panic` calls `std.debug.panicExtra`, which calls `resetSegfaultHandler`.

Here is the issue: `resetSegfaultHandler` will then call `os.sigaction`, which may fail again and recurse infinitely since we just came from a failed `os.sigaction`.

zseri:
Fixes #8357
I couldn't really test it locally, both because I don't have a system with a failing `sigaction`.

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -1592,6 +1592,12 @@ pub fn maybeEnableSegfaultHandler() void {
 
 var windows_segfault_handle: ?windows.HANDLE = null;
 
+pub fn updateSegfaultHandler(act: ?*const os.Sigaction) error{OperationNotSupported}!void {
+    try os.sigaction(os.SIG.SEGV, act, null);
+    try os.sigaction(os.SIG.ILL, act, null);
+    try os.sigaction(os.SIG.BUS, act, null);
+}
+
 /// Attaches a global SIGSEGV handler which calls @panic("segmentation fault");
 pub fn attachSegfaultHandler() void {
     if (!have_segfault_handling_support) {
@@ -1607,9 +1613,9 @@ pub fn attachSegfaultHandler() void {
         .flags = (os.SA.SIGINFO | os.SA.RESTART | os.SA.RESETHAND),
     };
 
-    os.sigaction(os.SIG.SEGV, &act, null);
-    os.sigaction(os.SIG.ILL, &act, null);
-    os.sigaction(os.SIG.BUS, &act, null);
+    updateSegfaultHandler(&act) catch {
+        @panic("unable to install segfault handler, maybe adjust have_segfault_handling_support in std/debug.zig");
+    };
 }
 
 fn resetSegfaultHandler() void {
@@ -1625,9 +1631,8 @@ fn resetSegfaultHandler() void {
         .mask = os.empty_sigset,
         .flags = 0,
     };
-    os.sigaction(os.SIG.SEGV, &act, null);
-    os.sigaction(os.SIG.ILL, &act, null);
-    os.sigaction(os.SIG.BUS, &act, null);
+    // do nothing if an error happens to avoid a double-panic
+    updateSegfaultHandler(&act) catch {};
 }
 
 fn handleSegfaultPosix(sig: i32, info: *const os.siginfo_t, ctx_ptr: ?*const anyopaque) callconv(.C) noreturn {
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -5012,11 +5012,10 @@ pub fn sigaltstack(ss: ?*stack_t, old_ss: ?*stack_t) SigaltstackError!void {
 }
 
 /// Examine and change a signal action.
-pub fn sigaction(sig: u6, act: ?*const Sigaction, oact: ?*Sigaction) void {
+pub fn sigaction(sig: u6, noalias act: ?*const Sigaction, noalias oact: ?*Sigaction) error{OperationNotSupported}!void {
     switch (errno(system.sigaction(sig, act, oact))) {
         .SUCCESS => return,
-        .FAULT => unreachable,
-        .INVAL => unreachable,
+        .INVAL, .NOSYS => return error.OperationNotSupported,
         else => unreachable,
     }
 }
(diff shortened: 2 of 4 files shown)


zseri:
Two questions remain:
* should `os.sigaction` be replaced with the `gracefulSigaction` introduced here earlier?
* should `src/crash_report.zig` also be adjusted, and why doesn't it reuse `std.debug.attachSegfaultHandler`?

zseri:
ok, I'm pretty irritated why large amounts of code in `src/crash_report.zig` are just duplicated from `libs/std/debug.zig`, but `src/crash_report.zig` also imports `std.debug`, so avoiding using code from it doesn't make sense...

src/crash_report.zig
```diff
@@ -22,7 +22,7 @@ pub const root_decls = struct {
 /// Install signal handlers to identify crashes and report diagnostics.
 pub fn initialize() void {
     if (is_enabled and debug.have_segfault_handling_support) {
-        attachSegfaultHandler();
+        debug.attachSegfaultHandler();
```
schmee, line 25: This breaks `crash_report.zig`, see below.

src/crash_report.zig
```diff
@@ -159,145 +159,6 @@ pub fn compilerPanic(msg: []const u8, error_return_trace: ?*std.builtin.StackTra
...
-    };
-
-    os.sigaction(os.SIG.SEGV, &act, null);
-    os.sigaction(os.SIG.ILL, &act, null);
-    os.sigaction(os.SIG.BUS, &act, null);
-}
-
-fn handleSegfaultPosix(sig: i32, info: *const os.siginfo_t, ctx_ptr: ?*const anyopaque) callconv(.C) noreturn {
-    // TODO: use alarm() here to prevent infinite loops
-    PanicSwitch.preDispatch();
```
schmee, line 184: This code is similar to the one in `debug.zig` but not exactly duplicated, so it can't be straight up deleted. I ran into the same issue in https://github.com/ziglang/zig/pull/10615 and I think the fix is to make `debug.zig` use the implementation in `crash_report.zig`, since it is thread-safe and handles more edge cases (like double panic).

zseri: ok, does someone know where exactly the difference is?

schmee: The one commented on here is one, the way that the Windows sigfault handler is set up in `attachSegfaultHandler` is another, there might be more. Point is that unifying these two implementations is more complicated than just removing and/or copy-pasting code from one place to the other.

zseri: I removed my removal for now, but reused the `updateSegfaultHandler` to avoid a further diversion of the codes.

lib/std/os.zig
```diff
@@ -5012,12 +5012,12 @@ pub fn sigaltstack(ss: ?*stack_t, old_ss: ?*stack_t) SigaltstackError!void {
 }
 
 /// Examine and change a signal action.
-pub fn sigaction(sig: u6, act: ?*const Sigaction, oact: ?*Sigaction) void {
+pub fn sigaction(sig: u6, noalias act: ?*const Sigaction, noalias oact: ?*Sigaction) error{OperationNotSupported}!void {
```
zseri, line 5015: This is a breaking change, unsure how exactly that should be handled.

lib/std/debug.zig
```diff
@@ -1592,6 +1592,12 @@ pub fn maybeEnableSegfaultHandler() void {
 
 var windows_segfault_handle: ?windows.HANDLE = null;
 
+pub fn updateSegfaultHandler(act: ?*const os.Sigaction) error{OperationNotSupported}!void {
```
zseri, line 1595: `pub` because it's used by `src/crash_report.zig`

zseri:
The `azure/ziglang.zig` test failure seems to stem from the known timeout issue, and, as the set of tests seem to pass otherwise, I it can be ignored. It might be nice to add an unit test for this, but I have no idea how / what would be an appropriate method for mocking the `os` interface.

lib/std/os.zig
```diff
@@ -5012,12 +5012,12 @@ pub fn sigaltstack(ss: ?*stack_t, old_ss: ?*stack_t) SigaltstackError!void {
...
     switch (errno(system.sigaction(sig, act, oact))) {
         .SUCCESS => return,
-        .FAULT => unreachable,
-        .INVAL => unreachable,
-        else => unreachable,
+        // .FAULT => unreachable,
```
matu3ba, line 5018: Dead code is rather uncool, so you should make a comment, why the code stays for now.

matu3ba (commented):
nit: no commenting of code without a comment why the dead code remains

matu3ba:
> I have no idea how / what would be an appropriate method for mocking the os interface.

Why does calling the panic handler in the panic handler not work? Or in this case a custom segfault handler that segfaults?

zseri:
because we want to test a failing `sigaction` call, not a panicking handler.

lib/std/os.zig
```diff
@@ -5012,11 +5012,10 @@ pub fn sigaltstack(ss: ?*stack_t, old_ss: ?*stack_t) SigaltstackError!void {
 }
 
 /// Examine and change a signal action.
-pub fn sigaction(sig: u6, act: ?*const Sigaction, oact: ?*Sigaction) void {
+pub fn sigaction(sig: u6, noalias act: ?*const Sigaction, noalias oact: ?*Sigaction) error{OperationNotSupported}!void {
     switch (errno(system.sigaction(sig, act, oact))) {
         .SUCCESS => return,
-        .FAULT => unreachable,
```
zseri, line 5018: I removed the `.FAULT` case (for now) completely, as I don't know why was there in the first place given that the `else` branch also satisfies it. There may be some deeper reason behind it, but idk.

`INVAL` and `NOSYS` should catch most systems with unexpected behavoir without looping. But it might be appropriate to return `Unexpected` instead of using `unreachable`.

matu3ba:
Ah. @BinaryWarlock any idea how this could be tested in CI? There should be a way to tell qemu to emulate targets without signal handlers.

PR #10661, opened 2022-01-21
std: add http definitions for Method and Status

nektro:
starts again the `std.http` namespace and add definitions for the `Method` and `Status` primitives for use in further work and 3rd party libraries

additionally will be used by 910, 2007, 943

### DIFF
--- a/lib/std/http/method.zig
+++ b/lib/std/http/method.zig
@@ -14,4 +14,87 @@ pub const Method = enum {
     OPTIONS,
     TRACE,
     PATCH,
+
+    /// Returns true if a request of this method is allowed to have a body
+    /// Returns null if it also depends on other conditions
+    pub fn requestHasBody(self: Method) ?bool {
+        return switch (self) {
+            .GET => false,
+            .HEAD => false,
+            .POST => true,
+            .PUT => true,
+            .DELETE => null,
+            .CONNECT => false,
+            .OPTIONS => false,
+            .TRACE => false,
+            .PATCH => true,
+        };
+    }
+
+    /// Returns true if a response to this method is allowed to have a body
+    /// Returns null if it also depends on other conditions
+    pub fn responseHasBody(self: Method) ?bool {
+        return switch (self) {
+            .GET => true,
+            .HEAD => false,
+            .POST => true,
+            .PUT => false,
+            .DELETE => null,
+            .CONNECT => true,
+            .OPTIONS => true,
+            .TRACE => false,
+            .PATCH => true,
+        };
+    }
+
+    /// An HTTP method is safe if it doesn't alter the state of the server.
+    /// https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP
+    /// https://datatracker.ietf.org/doc/html/rfc7231#section-4.2.1
+    pub fn safe(self: Method) bool {
+        return switch (self) {
+            .GET => true,
+            .HEAD => true,
+            .POST => false,
+            .PUT => false,
+            .DELETE => false,
+            .CONNECT => false,
+            .OPTIONS => true,
+            .TRACE => true,
+            .PATCH => false,
+        };
+    }
+
+    /// An HTTP method is idempotent if an identical request can be made once or several times in a row with the same effect while leaving the server in the same state.
+    /// https://developer.mozilla.org/en-US/docs/Glossary/Idempotent
+    /// https://datatracker.ietf.org/doc/html/rfc7231#section-4.2.2
+    pub fn idempotent(self: Method) bool {
+        return switch (self) {
+            .GET => true,
+            .HEAD => true,
+            .POST => false,
+            .PUT => true,
+            .DELETE => true,
+            .CONNECT => false,
+            .OPTIONS => true,
+            .TRACE => true,
+            .PATCH => false,
+        };
+    }
+
+    /// A cacheable response is an HTTP response that can be cached, that is stored to be retrieved and used later, saving a new request to the server.
+    /// https://developer.mozilla.org/en-US/docs/Glossary/cacheable
+    /// https://datatracker.ietf.org/doc/html/rfc7231#section-4.2.3
+    pub fn cacheable(self: Method) bool {
+        return switch (self) {
+            .GET => true,
+            .HEAD => true,
+            .POST => false,
+            .PUT => false,
+            .DELETE => false,
+            .CONNECT => false,
+            .OPTIONS => false,
+            .TRACE => false,
+            .PATCH => false,
+        };
+    }
 };
--- a/lib/std/http/method.zig
+++ b/lib/std/http/method.zig
@@ -38,15 +38,8 @@ pub const Method = enum {
     /// https://datatracker.ietf.org/doc/html/rfc7231#section-4.2.1
     pub fn safe(self: Method) bool {
         return switch (self) {
-            .GET => true,
-            .HEAD => true,
-            .POST => false,
-            .PUT => false,
-            .DELETE => false,
-            .CONNECT => false,
-            .OPTIONS => true,
-            .TRACE => true,
-            .PATCH => false,
+            .GET, .HEAD, .OPTIONS, .TRACE => true,
+            .POST, .PUT, .DELETE, .CONNECT, .PATCH => false,
         };
     }
 
@@ -55,15 +48,8 @@ pub const Method = enum {
     /// https://datatracker.ietf.org/doc/html/rfc7231#section-4.2.2
     pub fn idempotent(self: Method) bool {
         return switch (self) {
-            .GET => true,
-            .HEAD => true,
-            .POST => false,
-            .PUT => true,
-            .DELETE => true,
-            .CONNECT => false,
-            .OPTIONS => true,
-            .TRACE => true,
-            .PATCH => false,
+            .GET, .HEAD, .PUT, .DELETE, .OPTIONS, .TRACE => true,
+            .CONNECT, .POST, .PATCH => false,
         };
     }
 
@@ -72,15 +58,8 @@ pub const Method = enum {
     /// https://datatracker.ietf.org/doc/html/rfc7231#section-4.2.3
     pub fn cacheable(self: Method) bool {
         return switch (self) {
-            .GET => true,
-            .HEAD => true,
-            .POST => false,
-            .PUT => false,
-            .DELETE => false,
-            .CONNECT => false,
-            .OPTIONS => false,
-            .TRACE => false,
-            .PATCH => false,
+            .GET, .HEAD => true,
+            .POST, .PUT, .DELETE, .CONNECT, .OPTIONS, .TRACE, .PATCH => false,
         };
     }
 };
--- a/lib/std/http/method.zig
+++ b/lib/std/http/method.zig
@@ -1,0 +1,17 @@
+//! HTTP Methods
+//! https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
+
+// Style guide is violated here so that @tagName can be used effectively
+/// https://datatracker.ietf.org/doc/html/rfc7231#section-4 Initial definiton
+/// https://datatracker.ietf.org/doc/html/rfc5789#section-2 PATCH
+pub const Method = enum {
+    GET,
+    HEAD,
+    POST,
+    PUT,
+    DELETE,
+    CONNECT,
+    OPTIONS,
+    TRACE,
+    PATCH,
+};
--- a/lib/std/http/method.zig
+++ b/lib/std/http/method.zig
@@ -16,34 +16,20 @@ pub const Method = enum {
     PATCH,
 
     /// Returns true if a request of this method is allowed to have a body
-    /// Returns null if it also depends on other conditions
-    pub fn requestHasBody(self: Method) ?bool {
+    /// Actual behavior from servers may vary and should still be checked
+    pub fn requestHasBody(self: Method) bool {
         return switch (self) {
-            .GET => false,
-            .HEAD => false,
-            .POST => true,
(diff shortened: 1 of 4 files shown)


lib/std/http/status.zig
```diff
@@ -0,0 +1,193 @@
...
+
+    pub const Unknown = struct {
+        code: u16,
+        phrase: []const u8,
+    };
+
+    pub fn phrase(self: Status) []const u8 {
+        return switch (self) {
+            .standard => |v| v.phrase(),
+            .bare => |v| v.phrase,
```
zseri, line 158:
```suggestion
            .unknown => |v| v.phrase,
```

lib/std/http/status.zig
```diff
@@ -0,0 +1,193 @@
...
+        redirect,
+        client_error,
+        server_error,
+        invalid,
+    };
+
+    pub fn class(self: Status) Class {
+        const code = switch (self) {
+            .standard => @enumToInt(self.standard),
+            .unknown => self.unknown.code,
```
zseri, line 174:
```suggestion
            .standard => |v| @enumToInt(v),
            .unknown => |v| v.code,
```

zseri (changes requested):
`switch` on `Status` is currently inconsistent

lib/std/http/status.zig
nektro: good catch, will add a test

lib/std/http/status.zig
```diff
@@ -0,0 +1,203 @@
+//! HTTP Status
+//! https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
+
+const std = @import("../std.zig");
+
+pub const Status = union(enum) {
+    standard: Standard,
+    unknown: Unknown,
+
+    pub const Standard = enum(u16) {
```
daurnimator, line 10: Why `u16`?
For HTTP 1.x syntax compat, status can only be 100-999; i.e. 899 possible values, which would fit in a `u10`

lib/std/http/status.zig
```diff
@@ -0,0 +1,203 @@
...
+        bad_gateway = 502, // RFC7231, Section 6.6.3
+        service_unavailable = 503, // RFC7231, Section 6.6.4
+        gateway_timeout = 504, // RFC7231, Section 6.6.5
+        http_version_not_supported = 505, // RFC7231, Section 6.6.6
+        variant_also_negotiates = 506, // RFC2295
+        insufficient_storage = 507, // RFC4918
+        loop_detected = 508, // RFC5842
+        not_extended = 510, // RFC2774
+        network_authentication_required = 511, // RFC6585
+
```
daurnimator, line 77: This should be an extensible enum
```suggestion
+        _,

```

lib/std/http/method.zig
```diff
@@ -0,0 +1,17 @@
+//! HTTP Methods
+//! https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
+
+// Style guide is violated here so that @tagName can be used effectively
+/// https://datatracker.ietf.org/doc/html/rfc7231#section-4 Initial definiton
+/// https://datatracker.ietf.org/doc/html/rfc5789#section-2 PATCH
+pub const Method = enum {
+    GET,
+    HEAD,
+    POST,
```
daurnimator, line 10: For later optimisations I'd suggest matching the 2 entries in the HPACK static table (https://httpwg.org/specs/rfc7541.html#static.table.definition).
`GET = 1`
`POST = 2`

daurnimator (changes requested):

lib/std/http/status.zig
nektro: no, that's what `Status.unknown` is for

lib/std/http/status.zig
nektro: `u10` is an odd bit size, didn't know if it would hurt perf

lib/std/http/method.zig
nektro: you mean 2 and 3 ?

nektro: not sure if assigning numbers to those two would help, since `@enumToInt` would not apply to any of the others

daurnimator: > you mean 2 and 3 ?

Ah yep sorry; brain fail. I edited my comment

lib/std/http/status.zig
daurnimator: That would mean you get an ABI break every time a new status code is added.
Get rid of the `unknown` enum and use an extensible enum

nektro: why would it cause a break? there's no reason why someone would switch on it.

`Status.unknown` is a union option that is a struct so that if someone tries to parse a non standard code, the phrase text is not lost. this prevents needing multiple status types

lib/std/http/status.zig
```diff
@@ -0,0 +1,203 @@
...
+                .loop_detected => "Loop Detected",
+                .not_extended => "Not Extended",
+                .network_authentication_required => "Network Authentication Required",
+            };
+        }
+    };
+
+    pub const Unknown = struct {
+        code: u16,
+        phrase: []const u8,
```
daurnimator, line 152: Note that phrases are deprecated in HTTP. They do not exist since http2; so this would at least need to be optional.

nektro: why not always leave it as empty string then in that mode? because its not optional in 1.x

daurnimator: Because an empty string is distinct from a missing string

daurnimator: I'd recommend you stop the reason phrase entirely. See https://github.com/httpwg/http2-spec/issues/202#issuecomment-362946819

Or at least I don't think it belongs in the zig standard library.

nektro: it's needed when constructing and parsing a response, which is going to be important when we add a server or client implementation down the line

nektro: unless we don't want to support being able to parse a reason when using non-standard codes, which is fine, but a call we have to make

daurnimator: > unless we don't want to support being able to parse a reason when using non-standard codes, which is fine, but a call we have to make

IMO the low level http1-specific method would do so; but not the higher level layers that normal users would use.

I also advise e.g. exposing method and status as "just another header".

nektro: okay

lib/std/http/method.zig
```diff
@@ -14,4 +14,87 @@ pub const Method = enum {
...
+    }
+
+    /// Returns true if a response to this method is allowed to have a body
+    /// Returns null if it also depends on other conditions
+    pub fn responseHasBody(self: Method) ?bool {
+        return switch (self) {
+            .GET => true,
+            .HEAD => false,
+            .POST => true,
+            .PUT => false,
```
daurnimator, line 41: huh? `PUT` can have a body

lib/std/http/method.zig
```diff
@@ -14,4 +14,87 @@ pub const Method = enum {
     OPTIONS,
     TRACE,
     PATCH,
+
+    /// Returns true if a request of this method is allowed to have a body
+    /// Returns null if it also depends on other conditions
+    pub fn requestHasBody(self: Method) ?bool {
+        return switch (self) {
+            .GET => false,
```
daurnimator, line 22: `GET` can have a body.

lib/std/http/method.zig
```diff
@@ -14,4 +14,87 @@ pub const Method = enum {
...
+            .HEAD => false,
+            .POST => true,
+            .PUT => true,
+            .DELETE => null,
+            .CONNECT => false,
+            .OPTIONS => false,
+            .TRACE => false,
```
daurnimator, line 29: I think `TRACE` can have a body too?

lib/std/http/method.zig
```diff
@@ -14,4 +14,87 @@ pub const Method = enum {
     OPTIONS,
     TRACE,
     PATCH,
+
+    /// Returns true if a request of this method is allowed to have a body
+    /// Returns null if it also depends on other conditions
```
daurnimator, line 19: It always depends on other conditions. even a POST might have no body.

lib/std/http/method.zig
```diff
@@ -14,4 +14,87 @@ pub const Method = enum {
...
+    pub fn requestHasBody(self: Method) ?bool {
+        return switch (self) {
+            .GET => false,
+            .HEAD => false,
+            .POST => true,
+            .PUT => true,
+            .DELETE => null,
+            .CONNECT => false,
```
daurnimator, line 27: What do you call the contents of a CONNECT if not a body?

lib/std/http/method.zig
```diff
@@ -14,4 +14,87 @@ pub const Method = enum {
...
+            .OPTIONS => true,
+            .TRACE => true,
+            .PATCH => false,
+        };
+    }
+
+    /// A cacheable response is an HTTP response that can be cached, that is stored to be retrieved and used later, saving a new request to the server.
+    /// https://developer.mozilla.org/en-US/docs/Glossary/cacheable
+    /// https://datatracker.ietf.org/doc/html/rfc7231#section-4.2.3
+    pub fn cacheable(self: Method) bool {
```
daurnimator, line 87: I have a feeling this method could be very misleading (e.g. it might then depend on `Vary` header)

lib/std/http/status.zig
```diff
@@ -0,0 +1,178 @@
...
+
+    _,
+
+    pub fn phrase(self: Status) ?[]const u8 {
+        return switch (self) {
+            .@"continue" => "Continue",
+            .switching_protcols => "Switching Protocols",
+            .processing => "Processing",
+            .early_hints => "Early Hints",
+            .ok => "OK",
```
daurnimator, line 82: Should this be down one line after the empty line?

Might be helpful to separate sections with comments `// 1xx statuses` etc.

lib/std/http/method.zig
marler8997: Huh I didn't know some methods couldn't have bodies.  Do you know where this is in the HTTP spec?

lib/std/http/method.zig
zseri: idk, I would return `null`, since it seems similar to ws-upgrade, I wouldn't count that as body, because it's a stream I suppose...

zseri:
I think it might make sense to collapse the `switch` bodies like it's done at other places in the standard library, where all the branches which return `true` and the branches which return `false` are grouped; imo this would also make the code more readable, because not as much scrolling is necessary, e.g.
```zig
    /// An HTTP method is safe if it doesn't alter the state of the server.
    /// https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP
    /// https://datatracker.ietf.org/doc/html/rfc7231#section-4.2.1
    pub fn safe(self: Method) bool {
        return switch (self) {
            .GET, .HEAD, .OPTIONS, .TRACE => true,

            .POST, .PUT, .DELETE, .CONNECT, .PATCH => false,
        };
    }
```

lib/std/http/method.zig
nektro: its about *shouldn't*, something the client/server implementation can use in debug mode

lib/std/http/method.zig
nektro: its about *shouldn't*, something the client/server implementation can use in debug mode, the info is from MDN

lib/std/http/method.zig
nektro: its about *shouldn't*, something the client/server implementation can use in debug mode, the info is from MDN

nektro: gonna rework the doc comment to reflect this

lib/std/http/method.zig
nektro: gonna rework the doc comment to reflect this

lib/std/http/method.zig
nektro: gonna rework the doc comment to reflect this

lib/std/http/method.zig
nektro: gonna rework the doc comment to reflect this

lib/std/http/status.zig
nektro: added

lib/std/http/method.zig
```diff
@@ -0,0 +1,65 @@
...
+            .GET, .HEAD, .DELETE, .CONNECT, .OPTIONS, .TRACE => false,
+        };
+    }
+
+    /// Returns true if a response to this method is allowed to have a body
+    /// Actual behavior from clients may vary and should still be checked
+    pub fn responseHasBody(self: Method) bool {
+        return switch (self) {
+            .GET, .POST, .DELETE, .CONNECT, .OPTIONS, .PATCH => true,
+            .HEAD, .PUT, .TRACE => false,
```
daurnimator, line 32: I'm not sure why `PUT` is in this list?

Infact *most* `PUT` responses I've seen have a body on them.

nektro: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT

bdoner: Here's an [example of PUT returning a body.](https://github.com/tick/tick-api/blob/master/sections/entries.md#update-entry)

I would still say the stdlib should follow the http standard, but it should not block the ability to get a response body from a PUT request.

nektro: it wouldnt stop a client from receiving a PUT body, but it could raise a warning in the server implementation from sending a body

Inve1951:
Was `499` considered? It's not (yet) standard but commonly used for cancellation. E.g. when the visitor hits ESC in their browser.

PR #10665, opened 2022-01-22
spir-v improvements

Snektron:
This pull request brings a bunch of stuff that i worked on a few months ago. I was reasonably satisfied with the changes back then, except that they seemed to put a lot of pressure on the stage 1 compiler. While this was mainly due to the inline assembly part, the remainder is still slightly reworked.

The spir-v spec generator is updated to emit richer information, including a Zig type describing an opcode's operands. This type is then used to provide a nice interface to emitting instructions, which should prevent some mistakes, and (in my opinion) makes to code a bit nicer as well.

Furthermore, this pull request also introduces a new type- and module system for spir-v. This is mainly modelled after stage 2's own type and module system, but is tailored for spir-v in specific. The Module struct will be used to manage module-wide information, such as the global type cache, file name string op cache, and some others in the future. A module also consists of a number of sections structs, each of which model a particular logical section in the final spir-v binary.

There are two main reasons for introducing a dedicated type tree for spir-v types: First, non-aggregate non-pointer types must be uniquely declared. If multiple zig types translate to the same spir-v type, some extra layer needs to make sure that there is only one spir-v type instruction emitted for those if applicable. The second reason has to do with inline assembly: If we want to provide a decent experience for that, we are going to need to untangle the instructions in an inline assembly statement and put each in the appropriate section. If the user constructs a spir-v specific type which cannot be represented by Zig, it might also need to be deduplicated with a separate system.


lib/std/hash.zig
```diff
@@ -1,7 +1,7 @@
 const adler = @import("hash/adler.zig");
 pub const Adler32 = adler.Adler32;
 
-const auto_hash = @import("hash/auto_hash.zig");
+pub const auto_hash = @import("hash/auto_hash.zig");
```
Vexu, line 4: `auto_hash.hash` == `autoHashStrat`

PR #10668, opened 2022-01-22
Stop using LiniarFifo in BufferedWriter

Hejsil:
Looking at the BufferedWriter assembly generated, one can see that is
has to do a lot of work, just to copy over some bytes and increase an
offset. This is because the LiniarFifo is a much more general construct
than what BufferedWriter needs and the optimizer cannot prove that we
don't need to do this extra work.

Here are some benchmarks:

``` zig
const std = @import("std");

const fmt = std.fmt;
const heap = std.heap;
const io = std.io;
const process = std.process;

pub fn main() anyerror!void {
    var arena = heap.ArenaAllocator.init(heap.page_allocator);
    defer arena.deinit();
    const args = try process.argsAlloc(arena.allocator());
    const start = try fmt.parseUnsigned(usize, args[1], 10);
    const end = try fmt.parseUnsigned(usize, args[2], 10);

    var writer = io.bufferedWriter(io.getStdOut().writer());

    var i = start;
    while (i < end) : (i += 1)
        try writer.writer().print("{}\n", .{i});

    try writer.flush();
}
```

``` sh
$ # Master branch
$ zig build -Drelease-fast && perf stat zig-out/bin/benchy 10000 100000000 > /dev/null

 Performance counter stats for 'zig-out/bin/benchy 10000 100000000':

          1,832.49 msec task-clock                #    1.000 CPUs utilized
                 5      context-switches          #    2.729 /sec
                 0      cpu-migrations            #    0.000 /sec
                 9      page-faults               #    4.911 /sec
     7,830,620,192      cycles                    #    4.273 GHz
         4,024,945      stalled-cycles-frontend   #    0.05% frontend cycles idle
     3,083,214,556      stalled-cycles-backend    #   39.37% backend cycles idle
    30,755,900,740      instructions              #    3.93  insn per cycle
                                                  #    0.10  stalled cycles per insn
     4,492,432,054      branches                  #    2.452 G/sec
         1,703,301      branch-misses             #    0.04% of all branches

       1.832929489 seconds time elapsed

       1.769826000 seconds user
       0.060933000 seconds sys

# With this change
$ ~/repo/zig/build/zig build -Drelease-fast && perf stat zig-out/bin/benchy 10000 100000000 > /dev/null

 Performance counter stats for 'zig-out/bin/benchy 10000 100000000':

          1,241.51 msec task-clock                #    0.999 CPUs utilized
               126      context-switches          #  101.489 /sec
                 3      cpu-migrations            #    2.416 /sec
                 8      page-faults               #    6.444 /sec
     5,317,868,613      cycles                    #    4.283 GHz
         1,915,316      stalled-cycles-frontend   #    0.04% frontend cycles idle
     2,167,800,635      stalled-cycles-backend    #   40.76% backend cycles idle
    19,430,694,472      instructions              #    3.65  insn per cycle
                                                  #    0.11  stalled cycles per insn
     2,717,122,389      branches                  #    2.189 G/sec
           669,062      branch-misses             #    0.02% of all branches

       1.242325483 seconds time elapsed

       1.228096000 seconds user
       0.011975000 seconds sys
```

### DIFF
--- a/lib/std/io/buffered_writer.zig
+++ b/lib/std/io/buffered_writer.zig
@@ -1,24 +1,22 @@
 const std = @import("../std.zig");
+
 const io = std.io;
+const mem = std.mem;
 
 pub fn BufferedWriter(comptime buffer_size: usize, comptime WriterType: type) type {
     return struct {
         unbuffered_writer: WriterType,
-        fifo: FifoType = FifoType.init(),
+        buf: [buffer_size]u8 = undefined,
+        end: usize = 0,
 
         pub const Error = WriterType.Error;
         pub const Writer = io.Writer(*Self, Error, write);
 
         const Self = @This();
-        const FifoType = std.fifo.LinearFifo(u8, std.fifo.LinearFifoBufferType{ .Static = buffer_size });
 
         pub fn flush(self: *Self) !void {
-            while (true) {
-                const slice = self.fifo.readableSlice(0);
-                if (slice.len == 0) break;
-                try self.unbuffered_writer.writeAll(slice);
-                self.fifo.discard(slice.len);
-            }
+            try self.unbuffered_writer.writeAll(self.buf[0..self.end]);
+            self.end = 0;
         }
 
         pub fn writer(self: *Self) Writer {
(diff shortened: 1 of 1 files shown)


daurnimator:
An important test is to write more than the buffer fits (autoflush), then keep going.

I suggest that any good optimizations can be made in `LinearFifo`

Hejsil:
> I suggest that any good optimizations can be made in `LinearFifo`

I have my doubts we can get the same kinds of performance by optimizing LiniarFifo. As I said, this contruct is more general than what BufferedWriter needs, so it just has to do more work because of that. I'll see if there are some things that can be done though.

Hejsil:
Also, it's not like using this abstraction made the code simpler or easier to read. What do we actually gain from using it here?

Hejsil:
> An important test is to write more than the buffer fits (autoflush), then keep going.

Also, this benchmark writes a total of `888840000` bytes to stdout, way more than what fits in the buffer. Autoflushing is definitly being tested here.

kubkon (approved):

zseri:
could you fix the typo `LiniarFifo` (vs `LinearFifo`) in the commit message?

Issue #10666, opened 2022-01-22
macOS: test failure causes panic in machoSearchSymbols


hazeycode:
### Zig Version

0.9.0

### Steps to Reproduce

1. Create a new project
```sh
mkdir test && cd test
zig init-lib
```

2. Modify src/main.zig to
```zig
const testing = @import("std").testing;

const other = @import("other.zig");

test {
    testing.refAllDecls(@This());
}
```

3. Create new file src/other.zig
```zig
const testing = @import("std").testing;

test {
    try testing.expect(false);
}
```

4. run `zig build test`

### Expected Behavior

Zig should output a stack trace of where the test expectation failed

### Actual Behavior

Zig panics and output the following:

```
chris@chris_mbp testtest % zig build test
Test [2/2] other.test ""... FAIL (TestUnexpectedResult)
thread 702619 panic: index out of bounds
/Users/chris/Desktop/zig/build/lib/zig/std/debug.zig:566:30: 0x1010ce054 in std.debug.machoSearchSymbols (test)
        const next = &symbols[mid + 1];
                             ^
/Users/chris/Desktop/zig/build/lib/zig/std/debug.zig:1417:50: 0x1010bd999 in std.debug.ModuleDebugInfo.getSymbolAtAddress (test)
                const symbol = machoSearchSymbols(self.symbols, relocated_address) orelse
                                                 ^
/Users/chris/Desktop/zig/build/lib/zig/std/debug.zig:595:54: 0x1010bc8f0 in std.debug.printSourceAtAddress (test)
    const symbol_info = try module.getSymbolAtAddress(address);
                                                     ^
/Users/chris/Desktop/zig/build/lib/zig/std/debug.zig:356:33: 0x1010b3f61 in std.debug.writeStackTrace (test)
        try printSourceAtAddress(debug_info, out_stream, return_address - 1, tty_config);
                                ^
/Users/chris/Desktop/zig/build/lib/zig/std/debug.zig:206:24: 0x1010b05b5 in std.debug.dumpStackTrace (test)
        writeStackTrace(stack_trace, stderr, getDebugInfoAllocator(), debug_info, detectTTYConfig()) catch |err| {
                       ^
/Users/chris/Desktop/zig/build/lib/zig/std/special/test_runner.zig:98:45: 0x1010af320 in std.special.main (test)
                    std.debug.dumpStackTrace(trace.*);
                                            ^
[...]

PR #10672, opened 2022-01-22
debug: fix edge cases in macOS debug symbol lookup

schmee:
This commit fixes two related things:

1. If the loop goes all the way through the slice without a match, on
   the last iteration `mid == symbols.len - 1` which causes
   `&symbols[mid + 1]` to be out of bounds. Handle this by clamping
   the value to the length of the slices.

2. If the address we're looking for is greater than the address of the
   last symbol in the `symbols` slice, we now match it to that symbol. Previously,
   we would miss this case since we only matched if the address was _in
   between_ the address of two symbols.

Closes https://github.com/ziglang/zig/issues/10666.

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -6,6 +6,7 @@ const io = std.io;
 const os = std.os;
 const fs = std.fs;
 const process = std.process;
+const testing = std.testing;
 const elf = std.elf;
 const DW = std.dwarf;
 const macho = std.macho;
@@ -568,7 +569,7 @@ pub const TTY = struct {
 
 fn machoSearchSymbols(symbols: []const MachoSymbol, address: usize) ?*const MachoSymbol {
     var min: usize = 0;
-    var max: usize = symbols.len;
+    var max: usize = symbols.len - 1;
     while (min < max) {
         const mid = min + (max - min) / 2;
         const curr = &symbols[mid];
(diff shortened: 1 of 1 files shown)


lib/std/debug.zig
```diff
@@ -563,7 +563,7 @@ fn machoSearchSymbols(symbols: []const MachoSymbol, address: usize) ?*const Mach
     while (min < max) {
         const mid = min + (max - min) / 2;
         const curr = &symbols[mid];
-        const next = &symbols[mid + 1];
+        const next = &symbols[@minimum(mid + 1, symbols.len - 1)];
```
kubkon, line 566: Out of curiosity, did you check if the same effect can be achieved by initially setting `max = len - 1` instead of `max = len`?

kubkon (changes requested):
I think we ought to have an inline test of `machoSearchSymbols` to future proof any changes to it, etc. Is this something you'd be willing to help with, ideally as part of this PR?

lib/std/debug.zig
schmee: IIRC I did try that and it didn't work for some reason, let me try again to confirm 👍

schmee:
@kubkon sure, I'll see what I can cook up!

lib/std/debug.zig
schmee: Seems like this does work, probably I tried it before I added the second half of the fix. Updated 👍

schmee:
@kubkon updated and added tests.

lib/std/debug.zig
```diff
@@ -581,9 +582,36 @@ fn machoSearchSymbols(symbols: []const MachoSymbol, address: usize) ?*const Mach
...
 }
 
+test "machoSearchSymbols" {
+    const symbols = [_]MachoSymbol{
+        .{ .addr = 100, .strx = undefined, .size = undefined, .ofile = undefined },
+        .{ .addr = 200, .strx = undefined, .size = undefined, .ofile = undefined },
+        .{ .addr = 300, .strx = undefined, .size = undefined, .ofile = undefined },
+    };
+
+    try testing.expectEqual(@as(?*const MachoSymbol, null), machoSearchSymbols(&symbols, 0));
```
kubkon, line 600: Nit for the future: you could also just do

```zig
try testing.expect(machoSearchSymbols(&symbols, 99) == null);
```

kubkon (approved):
Looks great! Thank you so much for further investigating my suggestions @schmee!

PR #10673, opened 2022-01-23
std: define static error set for fs.Dir.copyFile

nektro:
complex and needed to define set in consumer functions

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -2137,7 +2137,7 @@ pub const Dir = struct {
         defer atomic_file.deinit();
 
         try copy_file(in_file.handle, atomic_file.file.handle);
-        return try atomic_file.finish();
+        try atomic_file.finish();
     }
 
     pub const AtomicFileOptions = struct {
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -193,7 +193,9 @@ pub const AtomicFile = struct {
         self.* = undefined;
     }
 
-    pub fn finish(self: *AtomicFile) !void {
+    pub const FinishError = std.os.RenameError;
+
+    pub fn finish(self: *AtomicFile) FinishError!void {
         assert(self.file_exists);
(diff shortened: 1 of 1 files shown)


lib/std/fs.zig
```diff
@@ -2139,7 +2137,7 @@ pub const Dir = struct {
         defer atomic_file.deinit();
 
         try copy_file(in_file.handle, atomic_file.file.handle);
-        return atomic_file.finish();
+        return try atomic_file.finish();
```
zseri, line 2140: why is that necessary? it seems like a nop change to me... either way, wouldn't instead replacing `return` with `try` be more appropriate?, e.g.
```suggestion
        try atomic_file.finish();
```

lib/std/fs.zig
```diff
@@ -2114,17 +2116,13 @@ pub const Dir = struct {
...
     /// there is a possibility of power loss or application termination leaving temporary files present
     /// in the same directory as dest_path.
-    pub fn copyFile(
-        source_dir: Dir,
-        source_path: []const u8,
-        dest_dir: Dir,
-        dest_path: []const u8,
-        options: CopyFileOptions,
-    ) !void {
+    pub fn copyFile(source_dir: Dir, source_path: []const u8, dest_dir: Dir, dest_path: []const u8, options: CopyFileOptions) CopyFileError!void {
```
zseri, line 2125: I don't think that it's a good idea to fit all the arguments in a single line, the multi-line version is more readable.

lib/std/fs.zig
nektro: it makes it more clear to the reader who may or may not be at an editor that the function can error, before i thought it did not and it confused me

zseri: ok, but as the function otherwise returns void, it makes more sense to just replace `return` with `try` instead of using both.

zseri (approved):

PR #10679, opened 2022-01-23
Stage2: wasm - Implement unions

Luukdegram:
This PR implements unions for the wasm backend.

Besides unions, the following was done as well:
- genTypedValue for enums (needed for tagged unions and was previously always emitting 'undefined' as value).
- Directly use Type.abiSize() to determine load/store instruction rather than an own variant.
(Thanks to previous refactor PR this was now possible).
- Implement the `rem` instruction: Only had to be enabled and will allow better test runner features.
- Function calling through decl_ref.

Note that this moves all the behavior tests to the outer branch and each test case has been disabled individually for all non-passing backends.

### DIFF
--- a/src/arch/wasm/CodeGen.zig
+++ b/src/arch/wasm/CodeGen.zig
@@ -722,9 +722,9 @@ fn typeToValtype(ty: Type, target: std.Target) wasm.Valtype {
             if (info.bits > 32 and info.bits <= 64) break :blk wasm.Valtype.i64;
             break :blk wasm.Valtype.i32; // represented as pointer to stack
         },
-        .Enum => switch (ty.tag()) {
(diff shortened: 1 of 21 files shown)


Luukdegram:
Will resolve the merge conflicts introduced in f1b91bb41b2d810ecabf4c69cad91b24b3846b77 tomorrow and reverify the passing tests.

Luukdegram:
Rebased after master now that cbe also passes unions.

Luukdegram:
Rebased on top of master again to resolve merge conflicts from 0817d6b2150a00ab0d552888fc4822fced8e0f5f.

andrewrk:
Excellent work :+1:

PR #10680, opened 2022-01-23
stage2: always use GPA in debug builds

leecannon:
This PR makes debug builds of the compiler use `GeneralPurposeAllocator` even when linking libc.

Without this change debug builds on targets that always link libc or when the compiler links libc for tracy would use `c_allocator` meaning the issues checked by `GeneralPurposeAllocator.deinit()` are not caught.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -112,6 +112,7 @@ pub fn build(b: *Builder) !void {
     const tracy = b.option([]const u8, "tracy", "Enable Tracy integration. Supply path to Tracy source");
     const tracy_callstack = b.option(bool, "tracy-callstack", "Include callstack information with Tracy data. Does nothing if -Dtracy is not provided") orelse false;
     const tracy_allocation = b.option(bool, "tracy-allocation", "Include allocation information with Tracy data. Does nothing if -Dtracy is not provided") orelse false;
+    const force_gpa = b.option(bool, "force-gpa", "Force the compiler to use GeneralPurposeAllocator") orelse false;
     const link_libc = b.option(bool, "force-link-libc", "Force self-hosted compiler to link libc") orelse enable_llvm;
     const strip = b.option(bool, "strip", "Omit debug information") orelse false;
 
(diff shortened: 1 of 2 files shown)


andrewrk (changes requested):
Need to make sure the API boundaries between LLVM and stage2 are using the correct allocator, as well as stage1 and stage2.

For example maybe one side allocs and another frees; they need to use the same allocator. Not sure if there is a problem, but it needs to be audited before merging this.

leecannon:
Looks fine at the stage2 <-> stage1 boundary:
-  `zig_stage1_create` has a matching `zig_stage1_destroy` that frees with the same allocator. 
- `zig_stage1_build_object` - uses `buf_create_from_mem` and `buf_init_from_mem` to duplicate the slices given to it by stage2, no other allocated memory is passed to `zig_stage1_build_object` except `stage1->main_pkg` which is allocated into an `ArenaAllocator` in `updateStage1Module` and not freed on the c++ side

My knowledge of the boundary between stage2 and LLVM is a too lacking to comfortably say if all is well. But looking through the code I can't find anywhere we either we pass memory allocated by LLVM to a zig allocator or where we pass memory allocated by a zig allocator to LLVM without there also being a matching free of that memory on the zig side.

andrewrk:
Thanks for looking at the API boundaries.

After spending a ton of time debugging issues during #10656, I don't want to unconditionally use GPA in debug mode. It is useful to use valgrind sometimes. I used this several times to troubleshoot something. However, I also used GPA to troubleshoot stuff sometimes.

I think the defaults should stay what they are, and overriding to use GPA should be a new `-Dforce-gpa` option to the build script. So default allocator will be chosen like this:

1. `-Dforce-gpa` can be used to force GeneralPurposeAllocator (this is what I am requesting you to implement)
2. `-Dforce-link-libc` can be used to force libc allocator (this already exists)
3.  If linking libc, libc allocator
4. otherwise, GPA

leecannon:
Added `-Dforce-gpa` as requested

andrewrk:
Thanks :+1:

PR #10681, opened 2022-01-23
stage2: x64: fix printing of callee preserved regs instructions

mparadinha:
These MIR instruction were getting printed all on the same line.
before:
```
  mov qword ptr [rbp + 4294967288], rcxmov qword ptr [rbp + 4294967280], rdi
```
after
```
  mov qword ptr [rbp + 4294967288], rcx
  mov qword ptr [rbp + 4294967280], rdi
```


kubkon:
Nice, thanks!

PR #10688, opened 2022-01-24
stage2: Add `union` support to C backend

topolarity:
This PR adds basic support for `union` to the C backend.

There are some differences vs. the union encoding in the LLVM backend:
  - The payload/tag fields are not ordered based on their alignment
  - Tagged unions with a 0-bit payload are not lowered directly to their tag type. 
      Instead, they become a struct with an empty union as their payload field

Some additional fixes were needed to get `test/behavior/union.zig` to pass:
 - Rendering `.elem_ptr` + `.enum_numbered` types
 - (Not) rendering `void` function args
 - Add `__` prefix to rendered identifiers, to avoid conflicts with reserved C keywords
 
This is also my first Zig PR, so I hope I haven't missed any important steps/guidelines.
Please let me know if anything needs fixing 👍

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -408,6 +408,18 @@ pub const DeclGen = struct {
                     try dg.renderValue(writer, Type.usize, slice.len);
                     try writer.writeAll("}");
                 },
+                .elem_ptr => {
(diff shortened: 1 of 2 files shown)


andrewrk:
Nice work!

PR #10691, opened 2022-01-25
std.crypto.random: `Random`s are no longer passed by reference

nektro:
ever since 10045


daurnimator (approved):
Why did this not get picked up in tests?

ominitay:
> Why did this not get picked up in tests?

There was no test this would affect, as Zig automatically dereferences this, so very easy to miss.

PR #10696, opened 2022-01-26
stage2: x64: implement airTrunc

mparadinha:
Some bugs in other parts of the compiler lead to some of the truncate tests not working properly.
Specifically:
1. comparing signed integers of non power-of-two sizes
2. using `u0` or `i0` seems to crash the stage2 compiler

But ignoring those, all the other tests that needed `airTrunc` in both `test/behavior/basic.zig` and `test/behavior/truncate.zig`
now pass.


kubkon:
@mparadinha thank you so much for this contribution - I've merged it in 075e2ea commit after I refactor your solution a little so that it matches the style of the rest of codegen code. I have also fixed (hopefully for good) the signedness problem when loading values into registers and comparing them via `cmp` instruction. With this done and your `airTrunc` solution, I've now enabled the entire `test/behavior/truncate.zig` which passes in its entirety on the `x86_64` backend! :tada:

PR #10698, opened 2022-01-26
improve comptime windows GUID.parse performance

marler8997:
I found that after switching from my custom Guid parser to the one in std that it increased zigwin32 build times substantially (from 40 seconds to over 10 minutes).  More information can be found in the benchmark PR I created here: https://github.com/ziglang/gotta-go-fast/pull/21 .  This PR ports my GUID parser to std so all projects can leverage the faster comptime performance.

I also added a `parseNoBraces` function (which `parse` calls into) that supports parsing guids that don't have surrounding braces.

### DIFF
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -2765,20 +2765,12 @@ pub const GUID = extern struct {
         assert(s[13] == '-');
         assert(s[18] == '-');
         assert(s[23] == '-');
-        const GuidUnion = extern union {
-            typed: GUID,
-            bytes: [16]u8,
-        };
-        // GuidUnion only works if @sizeOf(GUID) == 16
-        comptime {
-            std.debug.assert(@sizeOf(GUID) == 16);
-        }
-        var guid = GuidUnion{ .bytes = undefined };
+        var bytes: [16]u8 = undefined;
         for (hex_offsets) |hex_offset, i| {
-            guid.bytes[i] = (try std.fmt.charToDigit(s[hex_offset], 16)) << 4 |
+            bytes[i] = (try std.fmt.charToDigit(s[hex_offset], 16)) << 4 |
                 try std.fmt.charToDigit(s[hex_offset + 1], 16);
         }
-        return guid.typed;
+        return @bitCast(GUID, bytes);
     }
 };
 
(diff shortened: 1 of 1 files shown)


lib/std/os/windows.zig
```diff
@@ -2738,50 +2738,73 @@ pub const GUID = extern struct {
...
-        return guid;
+        var guid = GuidUnion{ .bytes = undefined };
+        for (hex_offsets) |hex_offset, i| {
+            guid.bytes[i] = decodeHexByte([2]u8{ s[hex_offset], s[hex_offset + 1] });
+        }
+        return guid.typed;
     }
 };
 
+// TODO: is this already somewhere else?
```
daurnimator, line 2784: `std.fmt.charToDigit`?

lib/std/os/windows.zig
```diff
@@ -2738,50 +2738,73 @@ pub const GUID = extern struct {
...
+fn hexVal(c: u8) u4 {
+    if (c <= '9') return @intCast(u4, c - '0');
+    if (c >= 'a') return @intCast(u4, c + 10 - 'a');
+    return @intCast(u4, c + 10 - 'A');
+}
+
+// TODO: is this already somewhere else?
```
daurnimator, line 2791: `std.fmt.hexToBytes`?

lib/std/os/windows.zig
```diff
@@ -2738,50 +2738,62 @@ pub const GUID = extern struct {
...
+        return parseNoBraces(s[1 .. s.len - 1]) catch @panic("invalid GUID string");
+    }
+
+    pub fn parseNoBraces(s: []const u8) !GUID {
+        assert(s.len == 36);
+        assert(s[8] == '-');
+        assert(s[13] == '-');
+        assert(s[18] == '-');
+        assert(s[23] == '-');
+        const GuidUnion = extern union {
```
xackus, line 2768: `mem.bytesToValue`? or is it too slow?

marler8997: Thanks for the suggestion.  I tried it out and it seems to be slightly slower but not by too much.  All things being equal I would go with the faster one, is there reasons why this would not be preferable?

xackus: I think it would be marginally more readable with `mem.bytesToValue`.
But now that I thought about it more, a regular `@bitCast` should work.

PR #10699, opened 2022-01-26
[linux headers] rename arm64 to aarch64

motiejus:
Zig calls it aarch64. Linux calls it arm64. Currently lib/libc/include
has both arm64 and aarch64, which is quite confusing.

tools/update-linux-headers.zig was executed against the latest stable
linux patch version, therefore some other minor header updates. I will
update the wiki on how to do it once this PR is accepted.


andrewrk (approved):

motiejus:
Updated https://github.com/ziglang/zig/wiki/Updating-libc:

- build dependencies.
- how to get a latest stable kernel's patch version.

PR #10701, opened 2022-01-26
stage2 regalloc: Add freezeRegs/unfreezeRegs API

joachimschmidt557:
The freeze/unfreeze API replaces the exceptions API for hopefully preventing bugs in codegen code using the RegisterManager. The exceptions API is still available for backwards compatibility and will be removed once all backends transition to the new freeze/unfreeze API.

This topic was discussed in the [stage2 meeting on 2022-01-13](https://github.com/ziglang/zig/wiki/Self-Hosted-Compiler-Meetings#2022-01-13).

### DIFF
--- a/src/register_manager.zig
+++ b/src/register_manager.zig
@@ -6,7 +6,6 @@ const Allocator = std.mem.Allocator;
 const Air = @import("Air.zig");
 const Type = @import("type.zig").Type;
 const Module = @import("Module.zig");
-const LazySrcLoc = Module.LazySrcLoc;
 const expect = std.testing.expect;
 const expectEqual = std.testing.expectEqual;
 const expectEqualSlices = std.testing.expectEqualSlices;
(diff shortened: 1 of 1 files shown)


kubkon:
Just to get more context on this (so that I can apply the new API in the correct way in x86_64). Say I want to generate inline memcpy, and I need to use a couple of registers just for this operation alone; I would then allocate the required number of registers (potentially spilling the instructions if any), and freeze the allocated set of reqisters, deferring auto-free until the end of the current lexical scope. Is that more or less how we discussed it at the last meeting and how you envisioned it?

joachimschmidt557:
@kubkon Yes, exactly. For the current inline `memcpy`, the use of exceptions is not strictly necessary because `genInlineMemcpy` won't allocate new registers. But let's assume that it does allocate some registers. Then the method that you described will prevent `genInlineMemcpy` from accidentally spilling e.g. the register containing the length of the memory region.

I think a good recommendation is to always freeze registers once you are sure that you will need them further down the line, even (like in the case of inline `memcpy`) if no new registers will be allocated for this AIR instruction. This way, in future, the code can be extended to include code that will allocate further registers. So the code is future-proof and bugs that would have been easily accidentally introduced are prevented.

kubkon:
Excellent! This sounds perfect! Thanks @joachimschmidt557 and excellent work!

Issue #10634, opened 2022-01-19
`zig c++`: Compile CUDA code


saierd:
### Zig Version

0.10.0-dev.290+3901b6fb0

### Steps to Reproduce

Try to compile a `.cu` file with `zig c++`.

### Expected Behavior

`clang++` can compile CUDA code as described here: https://llvm.org/docs/CompileCudaWithLLVM.html It would be nice if zig could also expose this functionality.

I am not sure if this is a simple matter of adding the file extension to treat it as compilable code or if there needs to be some other logic for forwarding all of the additional arguments.

### Actual Behavior

zig tries to link using the given arguments, because it does not know the file extension.

andrewrk:
Looks like your PR (landed in 2f41bd3be438dae2a188cfae3295dc28f4e9d434) solved the issue. Let me know if you think there is still an issue.

saierd:
No, it did not really solve this issue. As I explained in [this comment](https://github.com/ziglang/zig/pull/10704#issuecomment-1023616464) the PR allows to use `zig cc` as a host compiler for `nvcc`. `clang`'s own CUDA mode does not work yet.

I just tried to run `zig c++ main.cu --cuda-gpu-arch=sm_35 -mcpu=sm_35` on a small example as described in the link above, but it failed with

```
error: Unknown CPU: 'sm_35'
```

PR #10704, opened 2022-01-27
zig cc: Treat cu files as C++ source files

saierd:
First step towards #10634.

Treating stub files as C++ allows to use zig c++ as a host
compiler for nvcc.

Treating cu files as C++ allow using zig c++ as a host compiler in
CMake. CMake calls the host compiler with -E on a cu file to identify
the compiler.

Using zig c++ to directly compile CUDA code is untested.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -4198,7 +4198,10 @@ pub fn hasCppExt(filename: []const u8) bool {
     return mem.endsWith(u8, filename, ".C") or
         mem.endsWith(u8, filename, ".cc") or
         mem.endsWith(u8, filename, ".cpp") or
-        mem.endsWith(u8, filename, ".cxx");
+        mem.endsWith(u8, filename, ".cxx") or
+        mem.endsWith(u8, filename, ".cu") or
+        // .stub files are compiled by nvcc when using `zig c++` as the host compiler. They contain C++ code.
+        mem.endsWith(u8, filename, ".stub");
 }
 
 pub fn hasObjCExt(filename: []const u8) bool {


andrewrk:
Are `.cu` files and `.stub` files strictly C++ code? Or are they something different, that is also passed to a c++ compiler?

If so, we should add an additional enum tag to `FileExt` for it.

saierd:
Thank you for merging. Will this change be included in 0.9.1 or is there already a branch for that version?

---

To answer your question:

The `.stub` files are actually C++ as far as I can tell.

`.cu` files are not really C++, they contain some [language extensions](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#c-language-extensions). As I wrote in #10634, clang can compile `.cu` files itself but this probably needs some  more work to pass the correct parameters to clang (which then requires an additional enum tag).

That is not what is usually done, though. When using Nvidia's `nvcc` for compiling a `.cu` file it splits out the C++ code into a separate file and calls a host compiler to process it, which can be `zig c++` now. See the [`nvcc` documentation](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#cuda-compilation-trajectory) for more information on how this works. In principle `zig c++` should never see the `.cu` file in this mode. When using CMake, though, it tries to identify the host compiler by executing it with `-E` on a `.cu` file directly. That is what this pull request aimed to fix as a first step.

In summary, with this pull request `zig c++` can be used as a host compiler for CUDA using CMake and `nvcc`. The direct compilation of `.cu` files using clang's CUDA mode probably does not work yet and needs some additional handling of `.cu` files.

andrewrk:
This is eligible for being included in the bug fix release.

gwenzek:
@saierd  If you're interested in Cuda programming with Zig you should look at #10189 and [cudaz](https://github.com/gwenzek/cudaz)

saierd:
> @saierd If you're interested in Cuda programming with Zig you should look at #10189 and [cudaz](https://github.com/gwenzek/cudaz)

Thank you for the reference, this is interesting. I already have code in C++, though, I just want to compile it :wink:

PR #10715, opened 2022-01-28
autodoc: generate WalkResults for ref types

g-w1:
this fixes this case:
```zig
const Foo = usize;
const Bar = u6;
```
@kristoff-it
I also cleaned up to not hardcode /home/kristoff
I can't test it in a browser since you didn't push the js.


kristoff-it:
Thank you very much! All makes sense, I'll try to bring the js stuff into the branch soon so we can have a complete feedback loop. Haven't done so yet because I haven't looked into how Zig install artifacts.

PR #10717, opened 2022-01-29
Add Seccomp bits for linux

The-King-of-Toasters:
Full functionality necessitates a classic BPF implementation, so I did that too. Included is a simulator for debugging/testing purposes. It's been fuzzed for a couple days and I haven't found any errors. It may be worth de-duplicating the shared constants in `std.os.linux.BPF` and `std.x.net.bpf` later on.

Most of the effort was spent writing the documentation for the bpf/seccomp bits, so I hope that everything is clear and understandable.

There's a small issue relating to `@offsetOf` that I'll open a new issue for.


lib/std/os/linux/seccomp.zig
```diff
@@ -0,0 +1,187 @@
...
+    /// Allow.
+    pub const ALLOW = 0x7fff0000;
+
+    // Masks for the return value sections.
+    pub const ACTION_FULL = 0xffff0000;
+    pub const ACTION = 0x7fff0000;
+    pub const DATA = 0x0000ffff;
+};
+
+pub const SECCOMP_USER_NOTIF_FLAG_CONTINUE = 1 << 0;
```
matu3ba, line 127: missing a comment. `man seccomp` does not hint what this is for. There is `SECCOMP_RET_USER_NOTIF`, but its not evident what `CONTINUE` is for.

lib/std/os/linux/seccomp.zig
```diff
@@ -0,0 +1,187 @@
...
+
+pub const ADDFD_FLAG = struct {
```
matu3ba, line 129: I think this was lately added: "Add seccomp notifier ioctl that enables adding fds" https://lwn.net/Articles/821351/
and does not show up in the man page. If you dont want to comment this here, provide a link or other authoritative ressources how to understand the bits.

matu3ba (commented):
just a nit on comments. do you have a plan how and when to test this in CI to prevent bitrot and tracking upstream Kernel?

Side question: Are there other OSes with bpf support? Otherwise `bpf.zig` belongs into `os/linux`.

The-King-of-Toasters:
BPF is supported on many different OSs and all implementations - besides Linux - share code. This is the problem with OS testing. For example, here's how to attach a filter to an interface on [OpenBSD](https://man.openbsd.org/bpf.4):
- Open `/dev/bpf` `O_RDONLY`.
- Issue a bunch of `ioctl`s to the device, including `BIOCSETIF`.
- Set the filter using the `BIOCSETF` `ioctl` command. The filter is contained in a `bpf_program` struct:
  ```c
  struct bpf_program {
      u_int bf_len;
      struct bpf_insn *bf_insns;
  };
  ```
- Loop reading the bpf device to get new `bpf_hdr` structs.

On Linux, the process is:
- Create an `AF_PACKET` socket.
- Bind the socket to the interface using a `sockaddr_ll` struct.
- Set the filter using the `SO_ATTACH_FILTER` `setsockopt` option. The filter is contained in a `sock_fprog` struct:
  ```c
  struct sock_fprog {
      unsigned short     len;
      struct sock_filter *filter;
  };
- Loop `recv`ing the socket to get packets.

On top of the OS-level differences, network packet capturing requires root privilege on the *BSDs and the `CAP_NET_RAW` capability on Linux.

Note, this is from a surface-level reading of tcpdump/libpcap. There are most likely errors.

Seccomp testing shouldn't be too hard, I left it out because it requires a bit of work:
- Create a pipe/socketpair.
- fork/clone to create a new process.
- Set a seccomp filter in the child process that returns an errno for invalid syscalls.
- Test for passing/failing syscalls and write the errno on the pipe.
- The parent reads the errnos and asserts they're the correct values.
- The parent kills the child.

The-King-of-Toasters:
I've tweaked and added to the seccomp documentation.

The-King-of-Toasters:
I've just touched up the docs one last time. @matu3ba, do you require any changes or is it good to merge as is?

matu3ba:
I like the tests and the documentation is very nice to read. Since the socket/pipe is currently untested in between processes, this should be good to merge.

lib/std/os/linux/seccomp.zig
```diff
@@ -0,0 +1,215 @@
...
+//! 2. Many syscalls take an `unsigned long` or `size_t` argument, the size of
+//!    which is dependant on the ABI. Since BPF programs execute in a 32-bit
+//!    machine, validation of 64-bit arguments necessitates two load-and-compare
+//!    instructions for the upper and lower words.
+//! 3. A further wrinkle to the above issue is CPU endianess. Unlike network
+//!    packets, syscall data shares the endianess of the target machine. A
+//!    filter compiled on a little-endian machine will not work on a big-endian
+//!    one, and vice-versa. For example: Checking the upper 32-bits of
+//!    `data.args[1]` requires a load at `@offsetOf(data, "args[1]") + 4` on
+//!    big-endian systems and `@offsetOf(data, "args[1]")` on little-endian
```
Vexu, line 39: `@offsetOf(data, "args[1]")` syntax is not likely to be added and should not be referenced.

Vexu (changes requested):

matu3ba:
@The-King-of-Toasters I was wrong on sockets on Linux (in `x/os/net.zig`). You can find a simple child process with pipe example for a temporary generated program here: https://gist.github.com/matu3ba/0add09568327bad64e7181f9faadca40

1. However, the paths are wrong, if I use a test block instead of main function.
2. The pipe test is in `os/test.zig` under `test "pipe"` and requires similar adapations.
3. It would be nice, if build.zig would have functionality for that as to not rely on hacky path workarounds.

The-King-of-Toasters:
I'm open to writing proper tests, however I feel writing a new style of test is out of scope for this addition, and should be worked on in its own PR.

I've also removed the few remaining references to `args[n]`, so I think everything is in place.

lib/std/os/linux.zig
```diff
@@ -5369,3 +5374,55 @@ pub const PERF = struct {
...
     pub const IOC_FLAG_GROUP = 1;
 };
+
+// TODO: Add the rest of the AUDIT defines?
+pub const AUDIT = struct {
+    pub const ARCH = enum(u32) {
+        const _64BIT = 0x80000000;
+        const _LE = 0x40000000;
+
+        pub const current = switch (native_arch) {
```
andrewrk, line 5384: missing type annotation (should have type `ARCH`)

lib/std/os/linux.zig
```diff
@@ -5369,3 +5374,55 @@ pub const PERF = struct {
...
+            .aarch64 => .AARCH64,
+            .arm, .thumb => .ARM,
+            .riscv64 => .RISCV64,
+            .sparcv9 => .SPARC64,
+            .mips => .MIPS,
+            .mipsel => .MIPSEL,
+            .powerpc => .PPC,
+            .powerpc64 => .PPC64,
+            .powerpc64le => .PPC64LE,
+            else => undefined,
```
andrewrk, line 5396: should be `else => @compileError("unsupported architecture")`

andrewrk (commented):
Thanks for the patch. I found a couple issues, but nothing so problematic as to deserve a revert.

The-King-of-Toasters:
Thanks for the feedback, will follow up with a commit soon.

PR #10720, opened 2022-01-29
windows: avoid runtime crash in fromSysTime

marler8997:
NOTE: currently a draft because I may have a fix for the root cause soon

A workaround for issue ziglang#10719.  Removes the temporary `i128` local variable `adjusted_epoch` which prevents a segfault from occuring on my machine.  It's unclear as to why. See issue ziglang#10719 for more details.

While I was editing this function I also added a bit of clarity in the doc comment to make it clear that "file time" was referring the SysTime in the function name `fromSysTime`.

### DIFF
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -1760,14 +1760,16 @@ pub fn peb() *PEB {
     return teb().ProcessEnvironmentBlock;
 }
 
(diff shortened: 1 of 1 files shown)


marler8997:
Closing in favor of the real fix here: https://github.com/ziglang/zig/pull/10729

PR #10721, opened 2022-01-29
Fix preadv/pwritev bug on 64bit platform

tw4452852:
On 64bit platform, we shouldn't split `offset` argument because the register is big enough to hold it.
Linux kernel also expects this assumption.

### DIFF
--- a/lib/std/os/linux.zig
+++ b/lib/std/os/linux.zig
@@ -440,26 +440,30 @@ pub fn read(fd: i32, buf: [*]u8, count: usize) usize {
 }
 
 pub fn preadv(fd: i32, iov: [*]const iovec, count: usize, offset: i64) usize {
-    const offset_halves = splitValueLE64(offset);
+    const offset_u = @bitCast(u64, offset);
     return syscall5(
         .preadv,
         @bitCast(usize, @as(isize, fd)),
         @ptrToInt(iov),
         count,
-        offset_halves[0],
-        offset_halves[1],
+        // Kernel expects the offset is splitted into largest natural word-size.
+        // See following link for detail:
+        // https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=601cc11d054ae4b5e9b5babec3d8e4667a2cb9b5
+        @truncate(usize, offset_u),
+        if (usize_bits < 64) @truncate(usize, offset_u >> 32) else 0,
     );
 }
 
 pub fn preadv2(fd: i32, iov: [*]const iovec, count: usize, offset: i64, flags: kernel_rwf) usize {
-    const offset_halves = splitValue64(offset);
+    const offset_u = @bitCast(u64, offset);
     return syscall6(
         .preadv2,
         @bitCast(usize, @as(isize, fd)),
         @ptrToInt(iov),
         count,
-        offset_halves[0],
-        offset_halves[1],
+        // See comments in preadv
+        @truncate(usize, offset_u),
+        if (usize_bits < 64) @truncate(usize, offset_u >> 32) else 0,
         flags,
     );
 }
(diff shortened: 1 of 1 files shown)


ifreund:
>  Linux kernel also expects this assumption.

Do you have a source on this? As far as I can tell from the source linux always expects the offset argument to be split
```c
SYSCALL_DEFINE5(preadv, unsigned long, fd, const struct iovec __user *, vec,
		unsigned long, vlen, unsigned long, pos_l, unsigned long, pos_h)
{
	loff_t pos = pos_from_hilo(pos_h, pos_l);

	return do_preadv(fd, vec, vlen, pos, 0);
}
```

tw4452852:
> There are multiple architectures that have both 32 and 64-bit versions and the 64-bit kernel must support system calls from 32-bit user-space programs. Those programs will put 64-bit arguments into two registers, but the 64-bit kernel will expect that argument in a single register. 

https://lwn.net/Articles/311630/

ifreund:
@tw4452852 If I've read the article correctly, that bit you quoted explains why they chose to *always* use two registers even on 64 bit targets. The only truth is in the linux source code.

ifreund:
Another lkml message confirming what I've been saying: https://lkml.kernel.org/lkml/1232976389-13021-4-git-send-email-kraxel@redhat.com/

tw4452852:
@ifreund 
On 64bit kernel `pos_from_lohi` will ignore high end.

ifreund:
> @ifreund On 64bit kernel `pos_from_hilo` will ignore high end.

If it did that, calling `preadv()` on a system with 64 bit kernel and 32 bit userspace wouldn't work, which is the point of splitting the arguments like this.

ifreund:
@tw4452852 Ah, just found this: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=601cc11d054ae4b5e9b5babec3d8e4667a2cb9b5

It seems this behavior was changed since preadv/pwritev was introduced and you're correct that linux now will infact ignore the high argument if the kernel and userland share the word size. Sorry for the confusion.

lib/std/os/linux.zig
```diff
@@ -440,28 +440,51 @@ pub fn read(fd: i32, buf: [*]u8, count: usize) usize {
...
-    const offset_halves = splitValueLE64(offset);
-    return syscall5(
-        .preadv,
-        @bitCast(usize, @as(isize, fd)),
-        @ptrToInt(iov),
-        count,
-        offset_halves[0],
-        offset_halves[1],
-    );
+    if (usize_bits < 64) {
```
ifreund, line 443: It would better match the kernel logic if instead of branching here we just passed `@truncate(usize, @bitCast(u64, offset))` and `@truncate(usize, @bitCast(u64, offset) >> 32)`.

ifreund, line 443: It would better match the kernel logic if instead of branching here we just passed `@truncate(usize, @bitCast(u64, offset))` and `@truncate(usize, @bitCast(u64, offset) >> 32)`.

ifreund, line 443: It would better match the kernel logic if instead of branching here we just passed `@truncate(usize, @bitCast(u64, offset))` and `@truncate(usize, @bitCast(u64, offset) >> 32)`.

ifreund: Also it would be very nice to link the linux commit I found with the explanation in the source code to prevent anyone being further confused by this in the future.

tw4452852: @ifreund Thanks for the comment.
For the high end, do you think this would be better?
`if (usize_bits < 64) @truncate(usize, @bitCast(u64, offset) >> usize_bits) else 0`

I think this would cause less confusion.

ifreund: Yeah, that seems nice to me. A good comment would also go a long way to prevent confusion.

tw4452852: Agree. I just uploaded a new version.

Vexu (approved):

Issue #10719, opened 2022-01-29
segfault running "zig build" when zig is built in debug mode


marler8997:
### Zig Version

0.10.0-dev.5230+4613499fa

### Steps to Reproduce

Build a debug version of Zig on Windows.  Run it with "zig build".

### Expected Behavior

Not to crash

### Actual Behavior

Crashes with:
```
Segmentation fault at address 0xffffffffffffffff
```

Using a debugger I can see the crash occurs in `__muloti4`.  This is normally provided by `compiler_rt`, however, in the Zig executable it's shadowed by the one from `c++.lib`.  I can reproduce this by compiling a simple test program and adding `-lc++` which will cause it to link to `__muloti4` from `c++lib` and reproduce the crash.

Note that this call to `__muloti4` comes from `std.os.windows.fromSysTime`.

```zig
pub fn fromSysTime(hns: i64) i128 {
    const adjusted_epoch: i128 = hns + std.time.epoch.windows * (std.time.ns_per_s / 100);
    return adjusted_epoch * 100;
}
```

If I remove the `adjusted_epoch` local variable, the call to `_muloti4` goes away and so does the crash.  I've included the assembly for both of these functions below.

The segfault occurs early in `__muloti4` which makes me guess the segfault probably occurs here:

https://github.com/ziglang/zig/blob/aca665cebd2b6ec9ec3db669cf5446ee45bbb5d0/lib/std/special/compiler_rt/mulo.zig#L21

My guess is that zig isn't able to handle a `callconv (.C)` function with 2 `i128` followed by a pointer, and the pointer ends up being invalid. Note that this is just "integers" and "pointers" which #1481 says it supports, but this might be a regression of it.

# Supplemental Info

Not sure if this helps but here are the values of each variable in `fromSysTime` if that helps.

| expr | value |
|------|-------|
| `hns` (from debugger) | `132879187288034414` |
| `std.time.epoch.windows * (std.time.ns_per_s / 100)` (calculated by hand, these are comptime-known) | `-116444736000000000` |
| `adjusted_epoch` | `16434451288034414` |
| should return | `1643445128803441400` |

```asm
__muloti4:
00007FF60FF04180  push        rbp  
00007FF60FF04181  push        r15  
00007FF60FF04183  push        r14  
00007FF60FF04185  push        r13  
00007FF60FF04187  push        r12  
00007FF60FF04189  push        rsi  
00007FF60FF0418A  push        rdi  
00007FF60FF0418B  push        rbx  
[...]

PR #10729, opened 2022-01-29
remove __muloti4 from libc++

marler8997:
fixes https://github.com/ziglang/zig/issues/10719

compiler_rt already provides __muloti4 but libc++ is also providing it and when linking libc++ it causes a crash on my windows x86_64 machine.

### DIFF
--- a/src/libcxx.zig
+++ b/src/libcxx.zig
@@ -43,7 +43,9 @@ const libcxx_files = [_][]const u8{
     "src/exception.cpp",
     "src/experimental/memory_resource.cpp",
     "src/filesystem/directory_iterator.cpp",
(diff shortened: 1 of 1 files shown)


andrewrk:
Thank you for leaving that comment behind to prevent future accidents :-)

andrewrk (approved):

marler8997:
I was able to build zig with and without this PR and used it to compile a simple test program to reproduce the segfault and then see the segfault go away.  See the code/build commands here: https://github.com/ziglang/zig/issues/10719#issuecomment-1024990891

kubkon (approved):

PR #10730, opened 2022-01-30
stage2 x64: move some already passing tests to correct place

mparadinha:


mparadinha:
I just noticed the `if` check for bot `arm` *and* `x86_64`, and it these seem to crash when trying to compile for `arm-linux-none`. I'm closing the PR.

mparadinha:
I just noticed the `if` checks for both `arm` *and* `x86_64`, and these tests seem to crash when trying to compile for `arm-linux-none`. I'm closing the PR.

PR #10732, opened 2022-01-30
stage2: llvm: simplify behavior.zig test tree

nektro:
simplifies the logic of part of the `test/behavior.zig` tree. starts with the llvm backend

this method of ignoring backends allows them to be independently developed asynchronously at their own pace and allows for more complicated logic than would otherwise be possible to represent in `test/behavior.zig` alone

additionally it reduces the probability for merge conflicts as any pr that unblocks tests would be only removing the respective if statement in the affected files, rather than updating the main tree in `behavior.zig`


nektro:
did not move `test/behavior/misc.zig` since it hits an `unreachable` here https://github.com/ziglang/zig/blob/f8e418c/src/codegen/llvm.zig#L1774

nektro:
output after this change

```
$ ./zig-out/bin/zig test test/behavior.zig -I test -fno-stage1 -fLLVM
652 passed; 379 skipped; 0 failed.
```

andrewrk:
Reviewing this PR is the same as doing the work myself. Please leave the shuffling around of behavior tests to the commits that actually are implementing improvements to the compiler.

PR #10734, opened 2022-01-30
std.os: throw compile error for `argv` on Windows

ominitay:
On Windows, `argv` is not populated by start code, and instead left as undefined. This is problematic, and can lead to incorrect programs compiling, but panicking when trying to access `argv`. This change causes these programs to produce a compile error on Windows instead, which is far preferable to a runtime panic.

related: #6600
cc: @enkore

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -217,9 +217,13 @@ pub const socket_t = if (builtin.os.tag == .windows) windows.ws2_32.SOCKET else 
 pub var environ: [][*:0]u8 = undefined;
 
 /// Populated by startup code before main().
-/// Not available on Windows. See `std.process.args`
-/// for obtaining the process arguments.
-pub var argv: [][*:0]u8 = undefined;
+/// Not available on WASI or Windows without libc. See `std.process.argsAlloc`
+/// or `std.process.argsWithAllocator` for a cross-platform alternative.
+pub var argv: [][*:0]u8 = if (builtin.link_libc) undefined else switch (builtin.os.tag) {
+    .windows => @compileError("argv isn't supported on Windows: use std.process.argsAlloc instead"),
+    .wasi => @compileError("argv isn't supported on WASI: use std.process.argsAlloc instead"),
+    else => undefined,
+};
 
 /// To obtain errno, call this function with the return value of the
 /// system function call. For some systems this will obtain the value directly


ominitay:
Hm, that compile error isn't very helpful in finding *where* it was referenced.

Vexu:
Try reverting #10648

ominitay:
> Try reverting #10648

Wow, why was that merged? That makes debugging compile errors so much harder.

ominitay:
CI fail was the usual timeout on Windows. Will rebase and squash -- hopefully the CI decides to function correctly this time.

ominitay:
Windows CI timed out again, another rebase until this starts working :^)

ominitay:
Windows tests now passing :)
Ready to merge if maintainers approve

lib/std/os.zig
```diff
@@ -219,7 +219,7 @@ pub var environ: [][*:0]u8 = undefined;
 /// Populated by startup code before main().
 /// Not available on Windows. See `std.process.args`
 /// for obtaining the process arguments.
-pub var argv: [][*:0]u8 = undefined;
+pub var argv: [][*:0]u8 = if (builtin.os.tag != .windows) undefined else @compileError("argv isn't supported on Windows: use std.process.args instead");
```
ifreund, line 222: It seems to me that this should be a switch, zig can target other platforms besides windows that don't have argv/aren't posixy. For exampl, WASI.

ominitay: Yep, it should be. I'll get that changed in a moment.

ominitay:
Just spotted a couple of changes I'd like to make to the startup code, will get that done in a moment.

ominitay:
Actually, it all seems fine to me. Ready to merge :)

lib/std/os.zig
```diff
@@ -217,9 +217,13 @@ pub const socket_t = if (builtin.os.tag == .windows) windows.ws2_32.SOCKET else
 pub var environ: [][*:0]u8 = undefined;
 
 /// Populated by startup code before main().
-/// Not available on Windows. See `std.process.args`
-/// for obtaining the process arguments.
-pub var argv: [][*:0]u8 = undefined;
+/// Not available on WASI without libc, or on Windows. See `std.process.argsAlloc`
+/// or `std.process.argsWithAllocator` for obtaining the process arguments.
+pub var argv: [][*:0]u8 = switch (builtin.os.tag) {
+    .windows => @compileError("argv isn't supported on Windows: use std.process.argsAlloc instead"),
```
Vexu, line 223: Looking at this again, Windows should also have `argv` if linking libc.

Vexu (changes requested):

lib/std/os.zig
ominitay: Thanks for pointing this out :)

lib/std/start.zig
```diff
@@ -437,7 +437,7 @@ fn expandStackSize(phdrs: []elf.Phdr) void {
 }
 
 fn callMainWithArgs(argc: usize, argv: [*][*:0]u8, envp: [][*:0]u8) u8 {
-    std.os.argv = argv[0..argc];
+    if (builtin.os.tag != .windows or builtin.link_libc) std.os.argv = argv[0..argc];
```
Vexu, line 440: This will never be called when `builtin.os.tag == .windows and !builtin.link_libc`

src/stage1.zig
```diff
@@ -29,7 +29,7 @@ pub const log = stage2.log;
 pub const log_level = stage2.log_level;
 
 pub fn main(argc: c_int, argv: [*][*:0]u8) callconv(.C) c_int {
-    std.os.argv = argv[0..@intCast(usize, argc)];
+    if (builtin.os.tag != .windows or builtin.link_libc) std.os.argv = argv[0..@intCast(usize, argc)];
```
Vexu, line 32: Stage1 is always built with libc as LLVM depends on it.

Issue #10248, opened 2021-11-30
make overflow arithmetic builtins return a tuple instead of using a pointer parameter and bool return value


andrewrk:
Let's examine integer parsing without `@addWithOverflow` and `@mulWithOverflow`:

```zig
const maxInt = std.math.maxInt;

fn parseInt(comptime T: type, buf: []const u8, radix: u8) !T {
    const Big = @Type( .{ .Int = .{
        .signedness = false,
        .bits = @typeInfo(T).Int.bits * 2,
    }});
    var x: Big = 0;

    for (buf) |c| {
        const digit = switch (c) {
            '0'...'9' => c - '0',
            'A'...'Z' => c - 'A' + 10,
            'a'...'z' => c - 'a' + 10,
            else => return error.InvalidCharacter,
        };
        x *= radix;
        if (x > maxInt(T)) return error.Overflow;

        x += digit;
        if (x > maxInt(T)) return error.Overflow;
    }

    return @intCast(T, x);
}
```

With `@addWithOverflow` and `@mulWithOverflow`:

```zig
fn parseInt(comptime T: type, buf: []const u8, radix: u8) !T {
    var x: T = 0;

    for (buf) |c| {
        const digit = switch (c) {
            '0'...'9' => c - '0',
            'A'...'Z' => c - 'A' + 10,
            'a'...'z' => c - 'a' + 10,
            else => return error.InvalidCharacter,
        };
        if (@mulWithOverflow(T, x, radix, &x)) return error.Overflow;
        if (@addWithOverflow(T, x, digit, &x)) return error.Overflow;
    }

    return x;
}
```

[godbolt link](https://godbolt.org/z/zezfzMbhK)

Observations: Even with `-OReleaseFast` optimizations on, LLVM is able to generate better code with `@addWithOverflow` and `@mulWithOverflow`. The reason for this is that the backend that generates machine code can special-case the result of these functions, and turn struct field accesses followed by jumps into `jo` rather than actually doing multiplication with more bits.

Certainly in a debug build, this has the potential to generate much better runtime code.

So that's why these builtins exist: they help the programmer avoid integer overflow bugs, while generating efficient code.

There is a problem though, which is that the current function signature sabotages the potential for debug code to be efficient. The status quo function signature is:

```zig
@addWithOverflow(comptime T: type, a: T, b: T, result: *T) bool
```

The problem is the result pointer. As a point of comparison, here is the corresponding LLVM builtin signature:

```llvm
[...]

PR #10740, opened 2022-01-31
stage2: add more float arithmetic and f80 support

andrewrk:
AstGen: Fixed bug where f80 types in source were triggering illegal
behavior.

Value: handle f80 in floating point arithmetic functions.

Value: implement floatRem and floatMod


andrewrk:
cc @matu3ba - have a look at eb82fdf96c3392ec9eabacbaa98c976f9ccd264e. If you're up for it, implementations of these compiler-rt functions would unblock progress on passing more float test cases.

matu3ba:
@andrewrk 

```txt
.float_80 => @panic("TODO implement __remx in compiler-rt")
.float_128 => @panic("TODO implement fmodl in compiler-rt")
80 => @panic("TODO implement compiler_rt __modx");
80 => @panic("TODO implement compiler_rt __divxf3");
80 => @panic("TODO implement compiler_rt __floorx")
80 => @panic("TODO implement compiler_rt __truncx");
80 => @panic("TODO implement compiler_rt __mulxf3");
```

1. These are [softfloat routines](https://gcc.gnu.org/onlinedocs/gccint/Soft-float-library-routines.html) and it would be nice to document where they are ported from in the README of folder compiler_rt. Should I do that or move them to another folder `softfloat` next to the folder `compiler_rt` ?
2. Where do you want to have the Zig-internal generic builtin routines for backing #10248 ? In another folder `internal` next to folder ´compiler_rt` indicating their unstable abi?

andrewrk:
> These are [softfloat routines](https://gcc.gnu.org/onlinedocs/gccint/Soft-float-library-routines.html) and it would be nice to document where they are ported from in the README of folder compiler_rt. Should I do that or move them to another folder `softfloat` next to the folder `compiler_rt` ?

Here are two other possible places to port them from. You may find the licenses more amenable:

 * http://www.jhauser.us/arithmetic/SoftFloat-3/doc/SoftFloat.html
 * musl libc. The trick is to follow the correct ifdefs on the x86 platform for `long double` and you'll get f80 routines.

Everything in compiler-rt could be considered "soft routines" but so far we do not have any kind of organization and categorization into subfolders. Personally, I would keep it all flat. However, you have been doing a lot of work on compiler-rt, so if you want to do some organization in a way that makes sense to you, I would support those changes.

> Where do you want to have the Zig-internal generic builtin routines for backing #10248?

I think you misunderstand the purpose of this issue. **I do not expect any changes to compiler-rt for this issue**

Issue #10737, opened 2022-01-30
standard library hex float parsing incorrect value


andrewrk:
### Zig Version

0.10.0-dev.493+91ad96b88

### Steps to Reproduce

Apply this diff to get a new test case:

```diff
--- a/lib/std/fmt/parse_hex_float.zig
+++ b/lib/std/fmt/parse_hex_float.zig
@@ -338,6 +338,7 @@ test "f128" {
         // // Min denormalized value.
         .{ .s = "0x1p-16494", .v = math.f128_true_min },
         .{ .s = "-0x1p-16494", .v = -math.f128_true_min },
+        .{ .s = "0x1.edcb34a235253948765432134674fp-1", .v = 0x1.edcb34a235253948765432134674fp-1 },
     };
 
     for (cases) |case| {
```

Run the standard library tests:

```
$ ./zig test ../lib/std/std.zig 
```

### Expected Behavior

All tests passed.

### Actual Behavior

New test case fails:

```
Test [876/2176] fmt.parse_hex_float.test "f128"... expected 85065030167199919517705913334175581812, found 85065030167199919517705913334175581813
Test [877/2176] fmt.parse_hex_float.test "f128"... FAIL (TestExpectedEqual)
```

schmee:
Closed by https://github.com/ziglang/zig/pull/10743?

PR #10743, opened 2022-01-31
std: correct rounding in parse_hex_float.zig

m-radomski:
Addressing #10737, corrected the rounding logic in float128 parsing. Using [golang logic](https://github.com/golang/go/blob/a5c0b190809436fd196a348f85eca0416f4de7fe/src/strconv/atof.go#L528) (original source as the comment says) no longer produces the off-by-one error.

### DIFF
--- a/lib/std/fmt/parse_hex_float.zig
+++ b/lib/std/fmt/parse_hex_float.zig
@@ -206,13 +206,11 @@ pub fn parseHexFloat(comptime T: type, s: []const u8) !T {
     //   - we've truncated more than 0.5ULP (R=S=1)
     //   - we've truncated exactly 0.5ULP (R=1 S=0)
     // Were are going to increase the mantissa (round up)
-    var exactly_half = (mantissa & 0b11) == 0b10;
-    var more_than_half = (mantissa & 0b11) == 0b11;
+    const guard_bit_and_half_or_more = (mantissa & 0b110) == 0b110;
     mantissa >>= 2;
-    var guardBit = mantissa & 1 == 1;
     exponent += 2;
 
-    if (guardBit and (exactly_half or more_than_half)) {
+    if (guard_bit_and_half_or_more) {
         mantissa += 1;
     }
 
--- a/lib/std/fmt/parse_hex_float.zig
+++ b/lib/std/fmt/parse_hex_float.zig
@@ -202,18 +202,21 @@ pub fn parseHexFloat(comptime T: type, s: []const u8) !T {
         exponent += 1;
     }
 
-    // There are two cases to handle:
-    // - We've truncated more than 0.5ULP (R=S=1), increase the mantissa.
-    // - We've truncated exactly 0.5ULP (R=1 S=0), increase the mantissa if the
-    //   result is odd (G=1).
-    // The two checks can be neatly folded as follows.
-    mantissa |= @boolToInt(mantissa & 0b100 != 0);
-    mantissa += 1;
-
+    // Whenever the guard bit is one (G=1) and:
+    //   - we've truncated more than 0.5ULP (R=S=1)
+    //   - we've truncated exactly 0.5ULP (R=1 S=0)
+    // Were are going to increase the mantissa (round up)
+    var exactly_half = (mantissa & 0b11) == 0b10;
+    var more_than_half = (mantissa & 0b11) == 0b11;
     mantissa >>= 2;
+    var guardBit = mantissa & 1 == 1;
     exponent += 2;
 
-    if (mantissa & (1 << (mantissa_bits + 1)) != 0) {
+    if (guardBit and (exactly_half or more_than_half)) {
+        mantissa += 1;
+    }
+
+    if (mantissa == (1 << (mantissa_bits + 1))) {
         // Renormalize, if the exponent overflows we'll catch that below.
         mantissa >>= 1;
         exponent += 1;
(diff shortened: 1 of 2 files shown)


lib/std/fmt/parse_hex_float.zig
```diff
@@ -207,16 +207,19 @@ pub fn parseHexFloat(comptime T: type, s: []const u8) !T {
     // - We've truncated exactly 0.5ULP (R=1 S=0), increase the mantissa if the
     //   result is odd (G=1).
     // The two checks can be neatly folded as follows.
-    mantissa |= @boolToInt(mantissa & 0b100 != 0);
-    mantissa += 1;
-
+    var round: u128 = 0;
+    round = @boolToInt(mantissa & 0b011 != 0);
     mantissa >>= 2;
+    round |= @boolToInt(mantissa & 0b001 != 0);
```
Vexu, line 213: `u1 | u1` will never be `0b11`, here's an accurate translation:
```suggestion
    var round = @truncate(u2, mantissa);
    mantissa >>= 2;
    round |= @truncate(u1, mantissa); 
```

m-radomski:
Rounding is done according to [this](https://raw.githubusercontent.com/CppCon/CppCon2015/master/Presentations/Demystifying%20Floating%20Point/Demystifying%20Floating%20Point%20-%20John%20Farrier%20-%20CppCon%202015.pdf) presentation where only when the guard bit is non-zero we are going to round up.

lib/std/fmt/parse_hex_float.zig
```diff
@@ -341,7 +341,6 @@ test "f128" {
         // // Min denormalized value.
         .{ .s = "0x1p-16494", .v = math.f128_true_min },
         .{ .s = "-0x1p-16494", .v = -math.f128_true_min },
-        .{ .s = "0x1.edcb34a235253948765432134674fp-1", .v = 0x1.edcb34a235253948765432134674fp-1 },
```
andrewrk: I wouldn't consider this a duplicate. This case is not covered by the tests in this file. The test for the self-hosted compiler does not count; it's testing something different and it's not part of the standard library tests.

lib/std/fmt/parse_hex_float.zig
```diff
@@ -202,18 +202,21 @@ pub fn parseHexFloat(comptime T: type, s: []const u8) !T {
...
-    mantissa += 1;
-
+    // Whenever the guard bit is one (G=1) and:
+    //   - we've truncated more than 0.5ULP (R=S=1)
+    //   - we've truncated exactly 0.5ULP (R=1 S=0)
+    // Were are going to increase the mantissa (round up)
+    var exactly_half = (mantissa & 0b11) == 0b10;
+    var more_than_half = (mantissa & 0b11) == 0b11;
     mantissa >>= 2;
+    var guardBit = mantissa & 1 == 1;
```
andrewrk, line 212: This is equivalent to `mantissa & true` which is just going to give you a compile error

m-radomski: Get the last bit of mantissa as bool.

lib/std/fmt/parse_hex_float.zig
m-radomski: Reverted in db4c4cfb

lib/std/fmt/parse_hex_float.zig
```diff
@@ -202,18 +202,21 @@ pub fn parseHexFloat(comptime T: type, s: []const u8) !T {
...
-    //   result is odd (G=1).
-    // The two checks can be neatly folded as follows.
-    mantissa |= @boolToInt(mantissa & 0b100 != 0);
-    mantissa += 1;
-
+    // Whenever the guard bit is one (G=1) and:
+    //   - we've truncated more than 0.5ULP (R=S=1)
+    //   - we've truncated exactly 0.5ULP (R=1 S=0)
+    // Were are going to increase the mantissa (round up)
+    var exactly_half = (mantissa & 0b11) == 0b10;
```
marler8997, line 209: seems like some of these should be `const`?

lib/std/fmt/parse_hex_float.zig
m-radomski: Would something like this be considered more in zig ethos?
```suggestion
    var guardBit = @bitCast(bool, @truncate(u1, mantissa));
```

andrewrk: Ah I think I just got confused by order of operations, my mistake. Can you just put parens on it?

Related: #114

andrewrk: Oh and probably the simplest way to do this in zig is

```suggestion
    const guard_bit = @truncate(u1, mantissa) != 0;
```

lib/std/fmt/parse_hex_float.zig
```diff
@@ -202,18 +202,21 @@ pub fn parseHexFloat(comptime T: type, s: []const u8) !T {
         exponent += 1;
     }
 
-    // There are two cases to handle:
-    // - We've truncated more than 0.5ULP (R=S=1), increase the mantissa.
-    // - We've truncated exactly 0.5ULP (R=1 S=0), increase the mantissa if the
-    //   result is odd (G=1).
-    // The two checks can be neatly folded as follows.
-    mantissa |= @boolToInt(mantissa & 0b100 != 0);
```
DraagrenKirneh, line 210: Changing this line to:
mantissa += @boolToInt(mantissa & 0b100 != 0);

seems to fix the issue without needing to check both sides of the truncation.

DraagrenKirneh: and then removing the next line: `mantissa += 1`

andrewrk: I did a cleanup commit + merge. If you have an additional improvement to make, would you mind submitting it as a follow-up PR?

andrewrk:
Thanks @m-radomski!

PR #10751, opened 2022-02-01
stage2: cmp between untyped undefines values results in undef bool

mitchellh:
This makes the `"compare undefined literal with comptime_int"` behavior test pass.

This focused on just fixing this one test, but in the process I found a lot of weird behavior around "undefined" with stage1 (and therefore was unsure what the expected behavior was in stage2). I'll open another issue about that. 

This is my first time making a Zig compiler PR so please let me know if I should run any other commands. The only commands I ran was `test-stage2` and then the stage2 behavior tests runner with `zig2 test ../test/behavior.zig ...`

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -9244,9 +9244,9 @@ fn cmpSelf(
     const resolved_type = sema.typeOf(casted_lhs);
     const runtime_src: LazySrcLoc = src: {
         if (try sema.resolveMaybeUndefVal(block, lhs_src, casted_lhs)) |lhs_val| {
-            if (lhs_val.isUndef()) return sema.addConstUndef(resolved_type);
+            if (lhs_val.isUndef()) return sema.addConstUndef(Type.initTag(.bool));
             if (try sema.resolveMaybeUndefVal(block, rhs_src, casted_rhs)) |rhs_val| {
-                if (rhs_val.isUndef()) return sema.addConstUndef(resolved_type);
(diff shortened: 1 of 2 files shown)


src/Sema.zig
```diff
@@ -9244,9 +9244,9 @@ fn cmpSelf(
     const resolved_type = sema.typeOf(casted_lhs);
     const runtime_src: LazySrcLoc = src: {
         if (try sema.resolveMaybeUndefVal(block, lhs_src, casted_lhs)) |lhs_val| {
-            if (lhs_val.isUndef()) return sema.addConstUndef(resolved_type);
+            if (lhs_val.isUndef()) return sema.addConstUndef(Type.initTag(.bool));
```
andrewrk, line 9247:
```suggestion
            if (lhs_val.isUndef()) return sema.addConstUndef(Type.bool);
```

andrewrk (approved):
Nice! And... derp, silly mistake that I made there. Thanks for the fix!

I'll make this little fixup on merge, no need to push another commit on your end.

src/Sema.zig
mitchellh: Got it, thanks! You use `initTag(.bool)` one other place in Sema if you want to fix that up too at the same time, that's where I cargo culted this from.

andrewrk: Will do :+1:

PR #10756, opened 2022-02-01
compiler_rt: Implement atomic operations and __eabi_read_tp for ARM older than v7.

vesim987:
This PR implements most of the builtins from https://gcc.gnu.org/onlinedocs/gcc-4.1.0/gcc/Atomic-Builtins.html.
I omitted big-endian support and: 
- `__sync_lock_release`
- `__sync_bool_compare_and_swap`
I also implemented additional ones that i found in the LLVM test cases:
- `__sync_fetch_and_max`
- `__sync_fetch_and_umax`
- `__sync_fetch_and_min`
- `__sync_fetch_and_umin`

And also added support for thumb2 and <ARMv6 in `__aeabi_read_tp` using the `kuser_get_tls` helper

### DIFF
--- a/lib/std/special/compiler_rt/arm.zig
+++ b/lib/std/special/compiler_rt/arm.zig
@@ -41,10 +41,28 @@ pub fn __aeabi_unwind_cpp_pr2() callconv(.C) void {}
 // This function can only clobber r0 according to the ABI
 pub fn __aeabi_read_tp() callconv(.Naked) void {
     @setRuntimeSafety(false);
-    asm volatile (
-        \\ mrc p15, 0, r0, c13, c0, 3
-        \\ bx lr
-    );
+
+    // for thumb2, ARMv2, ARMv3, ARMv4 and ARMv5 use the kuser_get_tls linux helper which is at address 0xffff0fe0
+    if (comptime builtin.cpu.arch.isThumb() and std.Target.arm.featureSetHas(builtin.cpu.features, .thumb2)) {
+        asm volatile (
+            \\ movt r0, #0xffff
+            \\ movw r0, #0x0fe0
+            \\ bx r0
+        );
+    } else if (comptime builtin.cpu.arch.isARM() and
+        std.Target.arm.featureSetHasAny(builtin.cpu.features, .{ .v2, .v2a, .v3, .v3m, .v4, .v4t, .v5t, .v5te, .v5tej, .v6, .v6j, .v6m, .v6sm, .v6t2 }))
+    {
+        asm volatile (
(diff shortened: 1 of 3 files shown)


matu3ba:
1. How did you test this? (belongs into PR message)
2. Does any qemu runner work reliably for this to be tested in CI?

vesim987:
@matu3ba
> 1. How did you test this? (belongs into PR message+commit)

I manually extracted and run couple of the tests form `std.Thread.Mutex` and `std.atomic` on LicheePi Nano with ARM926EJ-S core, i couldn't run the whole std lib test suite on this device, because it has only 32MB of RAM. 
And also for `-target arm-linux-musleabi -mcpu arm926ej_s` LLVM is crashing on `std.Atomic(f32).init(5.0).load(.SeqCst)`, but I experienced similar crash when compiling musl. I think it is related to soft floats being broken on this target for some reasons, this problem can be reproduced on standard clang-13 builds from NixOS.
> 2. Does any qemu runner work reliably for this to be tested in CI?

`qemu-arm -cpu arm926` is crashing a lot, i ended up with ~270 core dumps while developing this. Maybe the `any` CPU can be used. the `kuser` helpers should be available on it.
> 3. Are additionally tests for the targets necessary?

I don't think so, the one in the for `std.atomic.Atomic` and `std.Thread.Mutex` are covering most of the builtins implemented in this PR.

vesim987:
Please do not merge, i reviewed my code and found one issue, atomic operations on unaligned `u8` and `u16` are broken, i forgot to shift the values.

lib/std/special/compiler_rt/arm.zig
```diff
@@ -40,10 +41,28 @@ pub fn __aeabi_unwind_cpp_pr2() callconv(.C) void {}
...
 pub fn __aeabi_read_tp() callconv(.Naked) void {
     @setRuntimeSafety(false);
-    asm volatile (
-        \\ mrc p15, 0, r0, c13, c0, 3
-        \\ bx lr
-    );
+
+    // for thumb2, ARMv2, ARMv3, ARMv4 and ARMv5 use the kuser_get_tls linux helper which is at address 0xffff0fe0
+    if (comptime builtin.cpu.arch.isThumb() and std.Target.arm.featureSetHas(builtin.cpu.features, .thumb2)) {
+        asm volatile (
```
daurnimator, line 47: Should this be conditional on the OS?

Should we hoist the magic constant out and give it a name?

Vexu:
Closing as stale. Open a new pr if you fix the issue with `u8` and `u16`.

PR #10757, opened 2022-02-01
Import SPARCv9 libunwind

koachan:
Import LLVM's [D32450](https://reviews.llvm.org/D32450)/[D116857](https://reviews.llvm.org/D116857) patch to enable unwinding support on SPARCv9 systems.

For now, this is just a stopgap solution, I'll update it again when upstream patch finally lands.


koachan:
Update it to the final version of the patch. Additionally, import [patch D119068](https://reviews.llvm.org/D119068) to enable `_Undwind_Backtrace`.

PR #10760, opened 2022-02-01
stage2: cmp between nans

m-radomski:
Catching the cmp in Sema stops it's being ordered against zero (which results in a hard crash). Also supported the logic of `(nan != nan) == true` is easier and less hacky compred to if it was to be handled lower down the chain. `float80` has been ommited as it does not seem to be supported that well in stage2 (crashes trying to access to `math.nan_f80`, no ability to `@bitCast` into `f80`).


andrewrk:
Thanks!

PR #10762, opened 2022-02-01
stage2: handle more MCValue types in `struct_field_ptr` in x86_64 and pad out nonpacked struct fields when lowering to bytes (all targets incl wasm32)

kubkon:
* start moving to new regalloc freeze API
* handle `struct_field_ptr` for `register` mcv
* handle `struct_field_ptr` for `stack_offset` mcv
* pass more behaviour tests

UPDATE:
Additionally,
* pad out (non-packed) struct fields when lowering to bytes to be saved in the binary - prior to this change, fields would be saved at non-aligned addresses leading to wrong accesses
* add a matching test case to `behavior/struct.zig` tests
* fix offset to field calculation in `struct_field_ptr` on `x86_64`

### DIFF
--- a/src/arch/x86_64/CodeGen.zig
+++ b/src/arch/x86_64/CodeGen.zig
@@ -680,6 +680,9 @@ fn genBody(self: *Self, body: []const Air.Inst.Index) InnerError!void {
             .wrap_errunion_err     => try self.airWrapErrUnionErr(inst),
             // zig fmt: on
         }
+
+        assert(!self.register_manager.frozenRegsExist());
+
         if (std.debug.runtime_safety) {
             if (self.air_bookkeeping < old_air_bookkeeping + 1) {
                 std.debug.panic("in codegen.zig, handling of AIR instruction %{d} ('{}') did not do proper bookkeeping. Look for a missing call to finishAir.", .{ inst, air_tags[inst] });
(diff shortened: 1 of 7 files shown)


src/arch/x86_64/CodeGen.zig
```diff
@@ -1795,22 +1791,62 @@ fn airStructFieldPtrIndex(self: *Self, inst: Air.Inst.Index, index: u8) !void {
...
+    const struct_field_offset = @intCast(u32, struct_ty.structFieldOffset(index, self.target.*));
+    const struct_field_ty = struct_ty.structFieldType(index);
+    const struct_field_size = @intCast(u32, struct_field_ty.abiSize(self.target.*));
+    const offset_to_field = struct_size - struct_field_offset - struct_field_size;
+
+    const dst_mcv: MCValue = result: {
         switch (mcv) {
+            .stack_offset => {
+                const offset_reg = try self.copyToTmpRegister(ptr_ty, .{
+                    .immediate = offset_to_field,
```
joachimschmidt557, line 1810: Hmm, I think this should be `struct_field_offset`. `mcv` contains the exact run-time address of the struct in memory, so getting a field from that struct should just be `base + offset` operation.

```suggestion
                    .immediate = struct_field offset,
```

kubkon: Hmm, right, lemme double check with multiple struct fields.

kubkon: You were absolutely right, thanks @joachimschmidt557. While at it, I've also fixed how we lower the struct to bytes in the final binary - prior to cf503a5 we would not correctly save each field at its required alignment leading to wrong field accesses in the generated machine code.

kubkon:
@Luukdegram I apologise for pulling what turns out to be all targets into one PR, but while adding a test for correct struct padding, I've encountered the same bug in wasm32 backend and went ahead and fixed it - structs in wasm32 are correctly padded out. I was hoping you could have a look and lemme know that you're fine with my changes to wasm32 codegen cbfa89d.

Luukdegram:
> @Luukdegram I apologise for pulling what turns out to be all targets into one PR, but while adding a test for correct struct padding, I've encountered the same bug in wasm32 backend and went ahead and fixed it - structs in wasm32 are correctly padded out. I was hoping you could have a look and lemme know that you're fine with my changes to wasm32 codegen d8b0068.

LGTM

PR #10770, opened 2022-02-02
stage 1: improve error message if error union is cast to payload

SebastianKeller:
@kristoff-it  mentioned that this error message could be improved.
Most often the code is missing a 'try'. The error message should reflect that.

before:

```
./src/main.zig:6:24: error: expected type '[]const u8', found '@typeInfo(@typeInfo(@TypeOf(getValue)).Fn.return_type.?).ErrorUnion.error_set![]const u8'
    printValue(getValue());
```
after:

```
./src/main.zig:6:24: error: Can not convert error union to payload type. Missing 'try'?
Expected type '[]const u8', found '@typeInfo(@typeInfo(@TypeOf(getValue)).Fn.return_type.?).ErrorUnion.error_set![]const u8'
    printValue(getValue());
```

code
```zig
const std = @import("std");
var gpa = std.heap.GeneralPurposeAllocator(.{}){};

pub fn main() anyerror!void {
    printValue(getValue());
}

fn printValue(str: []const u8) void {
    std.debug.print("{s}", .{str});
}

fn getValue() ![]const u8 {
    return try gpa.allocator().dupe(u8, "test");
}
```

### DIFF
--- a/src/stage1/ir.cpp
+++ b/src/stage1/ir.cpp
@@ -8242,6 +8242,20 @@ static Stage1AirInst *ir_analyze_cast(IrAnalyze *ira, Scope *scope, AstNode *sou
         return ir_implicit_cast2(ira, scope, source_node, cast1, wanted_type);
     }
 
+    // E!T to T
+    if (actual_type->id == ZigTypeIdErrorUnion) {
+        if (types_match_const_cast_only(ira, actual_type->data.error_union.payload_type, wanted_type,
+            source_node, false).id == ConstCastResultIdOk)
+        {
+            ErrorMsg *parent_msg = ir_add_error_node(ira, source_node,
+                buf_sprintf("cannot convert error union to payload type. consider using `try`, `catch`, or `if`. expected type '%s', found '%s'",
+                    buf_ptr(&wanted_type->name),
+                    buf_ptr(&actual_type->name)));
+            report_recursive_error(ira, source_node, &const_cast_result, parent_msg);
+            return ira->codegen->invalid_inst_gen;
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


src/stage1/ir.cpp
```diff
@@ -8242,6 +8242,20 @@ static Stage1AirInst *ir_analyze_cast(IrAnalyze *ira, Scope *scope, AstNode *sou
         return ir_implicit_cast2(ira, scope, source_node, cast1, wanted_type);
     }
 
+    // E!T to T
+    if (actual_type->id == ZigTypeIdErrorUnion) {
+        if (types_match_const_cast_only(ira, actual_type->data.error_union.payload_type, wanted_type,
+            source_node, false).id == ConstCastResultIdOk)
+        {
+            ErrorMsg *parent_msg = ir_add_error_node(ira, source_node,
+                buf_sprintf("Can not convert error union to payload type. Missing 'try'?\nExpected type '%s', found '%s'",
```
Vexu, line 8251:
```suggestion
                buf_sprintf("cannot convert error union to payload type. consider using `try`, `catch`, or `if`. expected type '%s', found '%s'",
```
No newlines in error messages, we should probably have a separate hint message type for suggestions but that can be an improvement in stage2.

Vexu (changes requested):
Needs a test case in `test/compile_errors.zig` so that it isn't forgotten in stage2.

Adding a similar error for `?T to T` should also be pretty simple with `consider using '.?', 'orelse' or 'if'` as the hint.

SebastianKeller:
Hi Vexu, thanks for the feedback! I've changed the message, added the `?T -> T` case and updated the `compile_errors.zig`.

Vexu (approved):

PR #10771, opened 2022-02-02
C Backend: f16 and f128 support

jean-dao:
- 128-bit floats: use `__float128` type, re-use 128-bit int literal trick for floats
- 16-bit floats: use either `__fp16` or `_Float16` when supported

Compiler support is not super robust, but since `__int128` is also used without any `#if defined` guard, I did the same for `__float128`.

Support for 16-bit floats is a bit more documented from what I found and I tried to make the `#if defined` guards match actual compiler support. See https://gcc.gnu.org/onlinedocs/gcc/Half-Precision.html and https://clang.llvm.org/docs/LanguageExtensions.html#id12. (I didn't found any doc for the Microsoft compiler.)

Note/Question: The `TODO: f128 ?` comment in `Value.floatToInt()` leads me to think there was maybe a reason (performance related?) not to directly use a f128 instead of a f64, so it may require some rework.

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -154,6 +154,7 @@ const reserved_idents = std.ComptimeStringMap(void, .{
     .{ "enum", {} },
     .{ "extern ", {} },
     .{ "float", {} },
+    .{ "float128_t", {} },
     .{ "for", {} },
     .{ "fortran", {} },
     .{ "goto", {} },
@@ -420,6 +421,7 @@ pub const DeclGen = struct {
         const int_info = @typeInfo(@TypeOf(int_val)).Int;
         const is_signed = int_info.signedness == .signed;
         const is_neg = int_val < 0;
+
         comptime assert(int_info.bits > 64 and int_info.bits <= 128);
 
         // TODO support big endian arch
@@ -433,13 +435,27 @@ pub const DeclGen = struct {
         const low = @truncate(u64, magnitude);
 
         // (int128_t)/<->( ( (uint128_t)( val_high << 64 )u ) + (uint128_t)val_low/u )
-        if (is_signed) try writer.writeAll("(int128_t)");
-        if (is_neg) try writer.writeByte('-');
+        if (is_signed) {
+            try writer.writeAll("(int128_t)");
+            if (is_neg)
+                try writer.writeByte('-');
+        } else {
+            try writer.writeAll("(uint128_t)");
+        }
 
-        try writer.print("(((uint128_t)0x{x}u<<64)", .{high});
+        if (magnitude == 0)
+            return writer.print("0", .{});
+
+        try writer.print("(", .{});
+
+        if (high > 0)
+            try writer.print("((uint128_t)0x{x}u << 64)", .{high});
+
+        if (high > 0 and low > 0)
+            try writer.print(" + ", .{});
 
         if (low > 0)
-            try writer.print("+(uint128_t)0x{x}u", .{low});
+            try writer.print("(uint128_t)0x{x}u", .{low});
 
         return writer.writeByte(')');
     }
@@ -539,6 +555,35 @@ pub const DeclGen = struct {
         }
     }
 
+    fn renderSmallFloat(
+        dg: *DeclGen,
+        comptime T: type,
+        writer: anytype,
+        val: Value,
+    ) error{ OutOfMemory, AnalysisFail }!void {
+        const v = val.toFloat(T);
+        if (std.math.isNan(v) or std.math.isInf(v)) {
+            // just generate a bit cast (exactly like we do in airBitcast)
+            switch (T) {
+                f16 => return dg.fail("TODO: C backend: implement 16-bit float lowering", .{}),
+                f32 => return writer.print("zig_bitcast_f32_u32(0x{x})", .{v}),
+                f64 => return writer.print("zig_bitcast_f64_u64(0x{x})", .{v}),
+                else => unreachable,
+            }
+        } else {
+            return writer.print("{x}", .{v});
+        }
+    }
+
+    fn renderFloat128(
+        writer: anytype,
+        val: f128,
+    ) error{ OutOfMemory, AnalysisFail }!void {
+        try writer.print("zig_bitcast_f128_u128(", .{});
+        try renderInt128(writer, @bitCast(u128, val));
+        try writer.print(")", .{});
+    }
+
     fn renderValue(
         dg: *DeclGen,
         writer: anytype,
@@ -605,19 +650,13 @@ pub const DeclGen = struct {
                 },
             },
             .Float => {
-                if (ty.floatBits(dg.module.getTarget()) <= 64) {
-                    if (std.math.isNan(val.toFloat(f64)) or std.math.isInf(val.toFloat(f64))) {
-                        // just generate a bit cast (exactly like we do in airBitcast)
-                        switch (ty.tag()) {
-                            .f32 => return writer.print("zig_bitcast_f32_u32(0x{x})", .{@bitCast(u32, val.toFloat(f32))}),
-                            .f64 => return writer.print("zig_bitcast_f64_u64(0x{x})", .{@bitCast(u64, val.toFloat(f64))}),
-                            else => return dg.fail("TODO float types > 64 bits are not support in renderValue() as of now", .{}),
-                        }
-                    } else {
-                        return writer.print("{x}", .{val.toFloat(f64)});
-                    }
+                switch (ty.tag()) {
+                    .f16 => return dg.renderSmallFloat(f16, writer, val),
+                    .f32 => return dg.renderSmallFloat(f32, writer, val),
+                    .f64 => return dg.renderSmallFloat(f64, writer, val),
+                    .f128 => return renderFloat128(writer, val.toFloat(f128)),
+                    else => unreachable,
                 }
-                return dg.fail("TODO: C backend: implement lowering large float values", .{});
             },
             .Pointer => switch (val.tag()) {
                 .null_value => try writer.writeAll("NULL"),
@@ -1295,11 +1334,11 @@ pub const DeclGen = struct {
             },
             .Float => {
                 switch (t.tag()) {
+                    .f16 => return dg.fail("TODO: C backend: implement float type f16", .{}),
                     .f32 => try w.writeAll("float"),
                     .f64 => try w.writeAll("double"),
+                    .f128 => try w.writeAll("float128_t"),
                     .c_longdouble => try w.writeAll("long double"),
-                    .f16 => return dg.fail("TODO: C backend: implement float type f16", .{}),
-                    .f128 => return dg.fail("TODO: C backend: implement float type f128", .{}),
                     else => unreachable,
                 }
             },
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -155,6 +155,7 @@ const reserved_idents = std.ComptimeStringMap(void, .{
     .{ "extern ", {} },
     .{ "float", {} },
     .{ "float128_t", {} },
+    .{ "float16_t", {} },
     .{ "for", {} },
     .{ "fortran", {} },
     .{ "goto", {} },
@@ -556,7 +557,6 @@ pub const DeclGen = struct {
     }
 
     fn renderSmallFloat(
-        dg: *DeclGen,
         comptime T: type,
         writer: anytype,
         val: Value,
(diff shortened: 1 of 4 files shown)


src/codegen/c.zig
```diff
@@ -386,52 +386,97 @@ pub const DeclGen = struct {
         try dg.renderDeclName(decl, writer);
     }
 
-    fn renderInt128(
+    fn renderBigIntConst(
```
matu3ba, line 389: What is the reason to rename it like it would accept any integer sizes and then do `assert(bits <= 128);` in the fn body?

src/codegen/c.zig
```diff
@@ -386,52 +386,97 @@ pub const DeclGen = struct {
...
 
         if (low > 0)
-            try writer.print("+(uint128_t)0x{x}u", .{low});
+            try writer.print("(uint128_t)0x{x}u", .{low});
 
         return writer.writeByte(')');
     }
 
-    fn renderBigIntConst(
+    fn renderInt128(
```
matu3ba, line 438: I dont understand why this fn is necessary.

matu3ba (changes requested):
Some things need clarification.

A few more tests would be great, both for floatop.zig and the C side.
floatop.zig has very few test cases, which do not involve negative numbers or the floating point number edge cases.

Did you look into the CI failures?

src/codegen/c.zig
```diff
@@ -386,52 +386,97 @@ pub const DeclGen = struct {
...
+                try writer.writeByte('-');
+        } else {
+            try writer.writeAll("(uint128_t)");
+        }
+
+        try writer.print("(", .{});
 
-        assert(high > 0);
-        try writer.print("(((uint128_t)0x{x}u<<64)", .{high});
+        if (high > 0)
```
matu3ba, line 426: What about negative numbers with 128 bit width?

src/codegen/c.zig
jean-dao: The reason I did those changes (`renderInt128`/`renderBigIntConst` refactoring/renaming) is not obvious anymore, I'll remove unneeded changes.

src/codegen/c.zig
```diff
@@ -386,52 +386,97 @@ pub const DeclGen = struct {
...
+        // hopefully unfold them if we construct one manually.
+        const magnitude = if (val.positive)
+            val.to(u128) catch unreachable
+        else
+            std.math.absCast(val.to(i128) catch unreachable);
 
         const high = @truncate(u64, magnitude >> 64);
         const low = @truncate(u64, magnitude);
 
         // (int128_t)/<->( ( (uint128_t)( val_high << 64 )u ) + (uint128_t)val_low/u )
```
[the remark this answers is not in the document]
jean-dao: Note that I didn't write the int128 handling code, just modifying the function to also accept smaller values. I can put a TODO comment, but I think it should be part of another PR.

src/codegen/c.zig
jean-dao: The reason I did those changes (renderInt128/renderBigIntConst refactoring/renaming) is not obvious anymore, I'll remove unneeded changes.

jean-dao:
> Some things need clarification. Better split this in 2 PRs, as this also includes 128bit integer support. Tests for the c codegen them are also missing.

I didn't write the 128-bit integer support, this was already here. I realize there are a bunch of unneeded changes in this PR, so I'll clean it up. I did have to change a bit the 128-bit int function to handle smaller values as well.

> A few more tests would be great, both for floatop.zig and the C side. floatop.zig has very few test cases, which do not involve negative numbers or the floating point number edge cases.

Agreed, will write more tests.

Thanks for the review!

src/codegen/c.zig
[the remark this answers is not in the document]
matu3ba: Better would be a compilation error with comptime.

jean-dao:
I updated my branch with the following changes:
- 128-bit floats are not available on all platform (e.g. not available on freebsd or macos), don't use `__float128` type in this case
- remove some unnecessary changes (functions renaming and such)
- add more tests
- fix an unrelated bug I found when writing tests ([829ea92](https://github.com/ziglang/zig/pull/10771/commits/829ea92127b8fad995bf4c658d8b511fb6048787))
- add a comptime error when using `renderInt128()` on big endian arch. I'm not 100% sure this is actually needed, I will look more into it.

drew-gpf:
Hi, I wrote the original renderInt128(). The endianness check should not be needed because you're just taking integer values and shifting them; the underlying representation only matters when you put it into memory, although given that there's doubt about it it would be worth it if someone could test this with a big endian arch.

matu3ba:
@jean-dao Can you rebase to resolve the conflicts?

test/behavior/floatop.zig
```diff
@@ -697,3 +697,38 @@ test "f128 at compile time is lossy" {
 
     try expect(@as(f128, 10384593717069655257060992658440192.0) + 1 == 10384593717069655257060992658440192.0);
 }
+
+test "f16 literals" {
+    if (builtin.zig_backend != .stage2_c) return error.SkipZigTest; // TODO
```
jmc-88, line 702: Why are this test and the next being skipped for the C backend? I would assume them to be passing after your changes.

jean-dao: You're right, only the f128 tests are not passing, I'll update again.

jean-dao:
``` 
stage2/bin/zig test test/behavior.zig -I test -target aarch64-linux --test-cmd qemu-aarch64 --test-cmd-bin
test/behavior/floatop.zig:701:1: error: TODO implement const of type 'f16'
```
Actually it's not passing, although for a different reason than f128. So I think it should still be skipped for backend != c.

src/codegen/c.zig
```diff
@@ -502,6 +527,34 @@ pub const DeclGen = struct {
...
+    fn renderSmallFloat(
+        comptime T: type,
+        writer: anytype,
+        val: Value,
+    ) error{ OutOfMemory, AnalysisFail }!void {
+        const v = val.toFloat(T);
+        if (std.math.isNan(v) or std.math.isInf(v)) {
+            // just generate a bit cast (exactly like we do in airBitcast)
+            switch (T) {
+                f16 => return writer.print("zig_bitcast_f16_u16(0x{x})", .{v}),
```
topolarity, line 539: Is a bitCast of `v` missing here?

src/link/C/zig.h
```diff
@@ -163,6 +163,30 @@
 
 #define int128_t __int128
 #define uint128_t unsigned __int128
+
+#if defined(__aarch64__)
+#define ZIG_HAS_FLOAT128
+#define float128_t long double
+#elif defined(__FreeBSD__) || defined(__APPLE__)
+#define float128_t float128_t_is_not_supported_on_this_target
```
matu3ba, line 171: Just curious: Why not `#error float128_t_is_not_supported_on_this_target` ?

src/link/C/zig.h
```diff
@@ -163,6 +163,30 @@
...
+#define ZIG_HAS_FLOAT16
+#define float16_t __fp16
+#elif defined(__GNUC__) && (defined(__arm__) || defined(__aarch64__))
+#define ZIG_HAS_FLOAT16
+#define float16_t __fp16
+#elif defined(__GNUC__) && defined(__i386__) && defined(__SSE2__)
+#define ZIG_HAS_FLOAT16
+#define float16_t _Float16
+#else
+#define float16_t float16_t_is_not_supported_on_this_target
```
matu3ba, line 187: same question: Why not `#error float128_t_is_not_supported_on_this_target` ?

test/behavior/floatop.zig
```diff
@@ -697,3 +697,38 @@ test "f128 at compile time is lossy" {
...
+
+    var buf: f16 = math.f16_min;
+    try expect(math.f16_min == buf);
+
```
matu3ba, line 706: missing test cases +-inf, -0 with at least 1 example with negative sign.
Not sure, if you also want to check `f16_true_min` or what the expected behavior for denormals is.

test/behavior/floatop.zig
```diff
@@ -697,3 +697,38 @@ test "f128 at compile time is lossy" {
...
+    buf = 0;
+    try expect(@as(f16, 0) == buf);
+}
+
+test "f128 tests" {
+    if (builtin.zig_backend != .stage2_c) return error.SkipZigTest; // TODO
+
+    const f128_min = @bitCast(f128, @as(u128, 0x00010000000000000000000000000000));
+    const f128_max = @bitCast(f128, @as(u128, 0x7FFEFFFFFFFFFFFFFFFFFFFFFFFFFFFF));
```
matu3ba, line 718: same story: at least 1 example with negative sign.

matu3ba (changes requested):
add a few examples to test the relevant cases.

test/behavior/floatop.zig
```diff
@@ -697,3 +697,38 @@ test "f128 at compile time is lossy" {
...
+    try expect(math.f16_min == buf);
+
+    buf = math.f16_max;
+    try expect(math.f16_max == buf);
+
+    buf = 0;
+    try expect(@as(f16, 0) == buf);
+}
+
+test "f128 tests" {
```
andrewrk, line 714: Can you please be more specific? What exactly is this testing?

src/codegen/c.zig
```diff
@@ -412,7 +412,19 @@ pub const DeclGen = struct {
         const int_info = @typeInfo(@TypeOf(int_val)).Int;
         const is_signed = int_info.signedness == .signed;
         const is_neg = int_val < 0;
-        comptime assert(int_info.bits > 64 and int_info.bits <= 128);
```
andrewrk, line 415: why is this comptime assert deleted?

test/behavior/floatop.zig
andrewrk, line 702: Please at least run the behavior tests with the LLVM backend (which is now the default backend) and don't disable it for the LLVM backend. Look at the `@sqrt` test in this file for a preamble that looks appropriate here.

andrewrk (changes requested):
Thanks for your patch. @topolarity's review comment looks like a merge blocker to me and I found a few other issues as well-

src/link/C/zig.h
jmc-88: +1 to @matu3ba's point, please use:

```
#error "float128_t is not supported on this target"
```

jean-dao: From my understanding, using `#error` will make all compilations fail on unsupported targets, even if `float128_t` is not used. I agree this is less clear though.

matu3ba: Can you attach a TODO and make this something along `ZIG_UNSUPPORTED_SYMBOL_ON_TARGET_float128_t` ?
I would envision a central example header for the user to invoke such that the user can check these kind of things and also includes stuff like abi checks as macros: https://github.com/marler8997/ziglibc/issues/1

Getting a convention for what stuff is unsupported sounds very relevant to me.

src/codegen/c.zig
jean-dao: This was wrongly deleted I reverted it, thanks.

test/behavior/floatop.zig
jean-dao: For some reason it was failing on the LLVM backend in the past, it is not the case anymore so I removed the conditional.

jean-dao:
The new update changelog:
- use bitcast to unsigned int type before printing hex value in `renderSmallFloat`
- more test values (inf, -inf, 0, -0, -1, nan, min, true_min, max, -max)
- better test names
- handle C limited support for f16: C (well at least [clang](https://clang.llvm.org/docs/LanguageExtensions.html#id14)) only supports f16 as a storage type, so it cannot be used for e.g. function return types or parameters.

jmc-88:
The CI run crashed. :frowning_face:  Can you reproduce the crash locally?

jean-dao:
@jmc-88 I cannot reproduce it locally. I thought maybe it could be related to musl libc vs glibc, but I couldn't yet reproduce the ci setup on my local setup. Since there is no output from the `zig run` subcommand that fail it's hard to know how to debug :/

topolarity:
I was able to reproduce the failure locally. This is the failing test case from `basic.zig`:
```zig
const expect = @import("std").testing.expect;

test "take address of parameter" {
    try testTakeAddressOfParameter(12.34);
}
fn testTakeAddressOfParameter(f: f32) !void {
    const f_ptr = &f;
    try expect(f_ptr.* == 12.34);
}
```

Looks like the problem is that 12.34 is rendered in decimal fp `1.23400001e+01` instead of hex fp `0x1.8ae148p3`, causing a rounding error.

jean-dao:
@topolarity thank you for your help!

On another topic, it seems the f128 tests are segfaulting on i386. I'm not quite sure how it passed before, I couldn't reproduce it (= passing test), even on an older master branch. I made the f128 test skipped on i386, not sure if that is acceptable.

topolarity:
Good news, it doesn't like the segfault is related to these changes! 👍 

Bug filed here, with root cause: https://github.com/ziglang/zig/issues/11383

src/codegen/c.zig
```diff
@@ -420,8 +422,12 @@ pub const DeclGen = struct {
         const int_info = @typeInfo(@TypeOf(int_val)).Int;
         const is_signed = int_info.signedness == .signed;
         const is_neg = int_val < 0;
+
         comptime assert(int_info.bits > 64 and int_info.bits <= 128);
 
+        // TODO support big endian arch
+        comptime assert(builtin.cpu.arch.endian() == .Little);
```
andrewrk, line 429: This is checking the endianness of the compiler, not the target. Also can you just fix the bug, if any, rather than asserting? It's not that hard to solve.

src/link/C/zig.h
```diff
@@ -165,6 +165,27 @@
 
 #define int128_t __int128
 #define uint128_t unsigned __int128
+
+#if defined(__aarch64__)
+#define ZIG_HAS_FLOAT128
+#define float128_t long double
+#elif defined(__FreeBSD__) || defined(__APPLE__)
+#define float128_t ZIG_UNSUPPORTED_SYMBOL_ON_TARGET_float128_t
```
andrewrk, line 173: In this case make it a 128 bit integer and emit instructions to perform softfloat on it. All the softfloat functions are available in compiler-rt.

andrewrk (changes requested):

andrewrk:
This PR seems to be abandoned. Please open a fresh one if you want to continue the effort.

PR #10779, opened 2022-02-03
stage2 ARM: optimize airSliceElemVal for elem_size 1 or 4

joachimschmidt557:
In these cases, the AIR inst can be lowered to only one ldr instruction.

Also fixes shifts in arm.bits.Offset

### DIFF
--- a/src/arch/arm/CodeGen.zig
+++ b/src/arch/arm/CodeGen.zig
@@ -1222,8 +1222,15 @@ fn airPtrSlicePtrPtr(self: *Self, inst: Air.Inst.Index) !void {
 fn airSliceElemVal(self: *Self, inst: Air.Inst.Index) !void {
     const is_volatile = false; // TODO
     const bin_op = self.air.instructions.items(.data)[inst].bin_op;
-    const result: MCValue = if (!is_volatile and self.liveness.isUnused(inst)) .dead else result: {
+
+    if (!is_volatile and self.liveness.isUnused(inst)) return self.finishAir(inst, .dead, .{ bin_op.lhs, bin_op.rhs, .none });
+    const result: MCValue = result: {
         const slice_mcv = try self.resolveInst(bin_op.lhs);
+
+        // TODO optimize for the case where the index is a constant,
+        // i.e. index_mcv == .immediate
+        const index_mcv = try self.resolveInst(bin_op.rhs);
+        const index_is_register = index_mcv == .register;
 
         const slice_ty = self.air.typeOf(bin_op.lhs);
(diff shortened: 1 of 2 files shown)


src/arch/arm/CodeGen.zig
```diff
@@ -1246,61 +1249,67 @@ fn airSliceElemVal(self: *Self, inst: Air.Inst.Index) !void {
...
-            self.register_manager.unfreezeRegs(&.{ base_mcv.register, offset_mcv.register });
+                try self.genArmBinOpCode(addr_reg, base_mcv, offset_mcv, false, .add, .unsigned);
 
-            try self.load(dst_mcv, .{ .register = addr_reg }, slice_ptr_field_type);
+                // I know we will unfreeze these registers at the end of
+                // the scope of :result. However, at this point in time,
+                // neither the base register nor the offset register
+                // contains any valuable data anymore. In order to reduce
+                // register pressure, unfreeze them prematurely
+                self.register_manager.unfreezeRegs(&.{ base_mcv.register, offset_mcv.register });
```
kubkon, line 1307: Hmm, this doesn't feel right to me. We should perhaps figure out a way to do this automatically. Also, we only make a call to `load` subsequently, so I guess reducing pressure here won't have a measurable effect?

joachimschmidt557: Yes, this does in fact have a measurable impact: Without this, the register manager will quickly run out of usable registers (of which there are only 6, see code snippet below) in certain test cases which involve inline `memcpy`.

https://github.com/ziglang/zig/blob/95fbce2b958395a367a82ce33170edd93e686173/src/arch/arm/bits.zig#L189

The long term solution to this consists of

- calling extern `memcpy`
- use caller-preserved registers as well

But for now (and possibly also for the future for reducing register pressure), this tiny optimization really does help.

kubkon: Right, agreed. So in this case, I believe your intent will be better communicated if you remove the defers in favour of explicit management of unfreezing (in this block). Then, there is no need for the comment here. Rather, we could add a comment to this effect to the `RegisterManager`'s docs to better document different idioms and uses. Lemme know what you think about this proposal!

joachimschmidt557: Ok, will do!

kubkon: @joachimschmidt557 I'm gonna merge this as-is and you can follow up with our proposal in a subsequent PR. Hope that sounds good to you!

PR #10780, opened 2022-02-03
stage2: Wasm - Account for stack alignment

Luukdegram:
The virtual stack has been slightly refactored to account for the alignment of the local variables.
This means that a local variable such as
```zig
var x: u32 align(128) = 10;
```
Will now work correctly.

We do this by calculating the total stack size that will then be forward aligned by the alignment of the largest type.
For the curious reader, the full explanation can be read in the commit message of e35414bf5c356798f201be85303101f59220326c

Besides this change, it also implements `elem_ptr` and `vector_init` instructions which makes the remaining array and slice behavior tests passing.


kubkon:
This is so cool! Great progress!

andrewrk (approved):
Excellent work!

PR #10781, opened 2022-02-03
turn echo back on in windows_msvc_script.bat

marler8997:
Zig turns echo on at the start of the msvc build script, but then `vcvarsall.bat` turns echo off (because otherwise it would print thousands of commands to the screen).  This PR turns it back after calling `vcvarsall.bat` on so we can see the rest of the batch commands in the CI logs.

This will help triage an intemittent failure on Windows where the CI sometimes gets stuck somewhere between "test-std" and "docs".  It's unclear exactly where but this change should make it clear and help us triage other CI failures in the future.


marler8997:
With this change we can see that Windows CI is getting stuck in the "zig build docs" step (see below).  Without it the CI doesn't echo the command so we don't know where it's stuck.

image

marler8997:
This PR no longer applies since https://github.com/ziglang/zig/commit/846eb701821a3f2af514bbad770478e3276b2d89 was integrated

Issue #10750, opened 2022-02-01
avoid depending on child process execution when the target OS does not support it


andrewrk:
There are a few places in the Zig compiler where we invoke a child process, but none of them are *strictly* necessary. This issue is to detect when the OS does not support child process execution and produce a build of the compiler that does not attempt to use child processes.

For now the detection logic can be simply:

```zig
const can_spawn = switch (builtin.os.tag) {
    .wasi => false,
    else => true,
};
```

Here is a complete list of everywhere Zig might want to spawn a child process:

 * `zig test` - instead print a message about not being able to spawn child processes and exit success.
 * `zig run` - instead print "unable to spawn child processes" and exit failure.
 * for clang we can invoke clang_main directly instead of as a child process
 * for lld we can invoke the lld library functions instead of as child process
 * same thing for `zig ar`, `zig dlltool`, etc.
 * for detecting native libc, just return error.LibcNotDetected or whatever the appropriate error is.

This is a prerequisite for #10716.

marler8997:
I'd be curious to see how significant the performance cost is for sub processes on platforms that support it.  With this in place we could easily find out!

zigazeljko:
Doesn't `zig build` also spawn child processes?

zseri:
Why should `zig test` exit with success?

marler8997:
> Doesn't zig build also spawn child processes?

Oh yes. To compile anything zig build will spawn a subprocess like `zig build-exe ...`.

matu3ba:
This should probably be first implemented before #1356 with potentially panicking subprocess being started from the subprocess that supervises test results.

andrewrk:
Landed in 21135387fb7c2dbaf70a72f2c97341e9c1307045

PR #10782, opened 2022-02-03
Avoid depending on child process execution when not supported by host OS

topolarity:
In accordance with the requesting issue (#10750):
- `zig test` skips any tests that it cannot spawn, returning success
- `zig run` and `zig build` exit with failure, reporting the command the cannot be run
- `zig clang`, `zig ar`, etc. already punt directly to the appropriate clang/lld main(), even before this change
- Native `libc` Detection is not supported

Additionally, `exec()` and related Builder functions error at run-time, reporting the command that cannot be run

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -229,6 +229,10 @@ pub fn build(b: *Builder) !void {
     const version = if (opt_version_string) |version| version else v: {
         const version_string = b.fmt("{d}.{d}.{d}", .{ zig_version.major, zig_version.minor, zig_version.patch });
 
+        if (!std.process.can_spawn) {
+            std.debug.print("error: version info cannot be retrieved from git. Zig version must be provided using -Dversion-string\n", .{});
+            std.process.exit(1);
+        }
         var code: u8 = undefined;
         const git_describe_untrimmed = b.execAllowFail(&[_][]const u8{
             "git", "-C", b.build_root, "describe", "--match", "*.*.*", "--tags",
@@ -542,6 +546,9 @@ fn addCxxKnownPath(
     errtxt: ?[]const u8,
     need_cpp_includes: bool,
 ) !void {
+    if (!std.process.can_spawn)
+        return error.RequiredLibraryNotFound;
+
     const path_padded = try b.exec(&[_][]const u8{
         ctx.cxx_compiler,
         b.fmt("-print-file-name={s}", .{objname}),
--- a/lib/std/child_process.zig
+++ b/lib/std/child_process.zig
@@ -124,6 +124,10 @@ pub const ChildProcess = struct {
 
     /// On success must call `kill` or `wait`.
     pub fn spawn(self: *ChildProcess) SpawnError!void {
+        if (!std.process.can_spawn) {
+            @compileError("the target operating system cannot spawn processes");
+        }
+
         if (builtin.os.tag == .windows) {
             return self.spawnWindows();
         } else {
--- a/lib/std/process.zig
+++ b/lib/std/process.zig
@@ -950,7 +950,13 @@ pub fn getSelfExeSharedLibPaths(allocator: Allocator) error{OutOfMemory}![][:0]u
 
 /// Tells whether calling the `execv` or `execve` functions will be a compile error.
 pub const can_execv = switch (builtin.os.tag) {
-    .windows, .haiku => false,
+    .windows, .haiku, .wasi => false,
+    else => true,
+};
+
+/// Tells whether spawning child processes is supported (e.g. via ChildProcess)
+pub const can_spawn = switch (builtin.os.tag) {
+    .wasi => false,
     else => true,
 };
 
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -25,6 +25,7 @@ const libunwind = @import("libunwind.zig");
 const libcxx = @import("libcxx.zig");
 const wasi_libc = @import("wasi_libc.zig");
 const fatal = @import("main.zig").fatal;
+const clangMain = @import("main.zig").clangMain;
 const Module = @import("Module.zig");
 const Cache = @import("Cache.zig");
 const stage1 = @import("stage1.zig");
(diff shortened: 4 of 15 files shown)


matu3ba (commented):
just a nit on `can_spawn` else looking good.

Did you check what is still missing?
Any of translate-c, ar, dlltool, lib, ranlib ?

lib/std/process.zig
```diff
@@ -936,6 +936,12 @@ pub const can_execv = switch (builtin.os.tag) {
     else => true,
 };
 
+/// Tells whether spawning child processes is supported (e.g. via ChildProcess)
+pub const can_spawn = switch (builtin.os.tag) {
```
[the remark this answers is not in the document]
topolarity: Do you mean that `can_spawn` will still need to be updated to be correct for other targets? If so, more than happy to add the TODO

lib/std/child_process.zig
```diff
@@ -126,8 +126,10 @@ pub const ChildProcess = struct {
     pub fn spawn(self: *ChildProcess) SpawnError!void {
         if (builtin.os.tag == .windows) {
             return self.spawnWindows();
-        } else {
+        } else if (std.process.can_spawn) {
             return self.spawnPosix();
+        } else {
+            unreachable;
```
andrewrk, line 132: I think this spawn function should be untouched, or if you want to make the compile error clearer, put at the top of it:

```zig
if (!std.process.can_spawn) {
    @compileError("the target operating system cannot spawn processes");
}
```

lib/std/process.zig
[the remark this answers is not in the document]
andrewrk: Prefer to make the TODO in a github issue rather than a comment, please :-)

build.zig
```diff
@@ -229,6 +229,11 @@ pub fn build(b: *Builder) !void {
     const version = if (opt_version_string) |version| version else v: {
         const version_string = b.fmt("{d}.{d}.{d}", .{ zig_version.major, zig_version.minor, zig_version.patch });
 
+        // If we can't spawn a process for git, just trust the version string
```
andrewrk, line 232: if we can't spawn a process for git, emit a fatal error here saying that it must be provided with `-Dversion-string`.

src/Compilation.zig
```diff
@@ -978,6 +979,9 @@ pub fn create(gpa: Allocator, options: InitOptions) !*Compilation {
             if (!build_options.have_llvm)
                 break :blk false;
 
+            if (!std.process.can_spawn)
+                break :blk false;
```
andrewrk, line 983: This should be removed. We can still invoke LLD and Clang even if we cannot spawn processes.

src/main.zig
```diff
@@ -2032,6 +2032,9 @@ fn buildOutputType(
         if (opt and cross_target.isDarwin()) {
             fatal("LLD requested with Mach-O object format. Only the self-hosted linker is supported for this target.", .{});
         }
+        if (opt and !std.process.can_spawn) {
+            fatal("LLD requested, but is unsupported on {s} due to inability to spawn child processes. Only the self-hosted linker is supported for this target.", .{@tagName(builtin.os.tag)});
+        }
```
andrewrk, line 2037: This is incorrect. We can still call LLD without doing it as a child process.

src/main.zig
```diff
@@ -4081,7 +4093,7 @@ extern "c" fn ZigClang_main(argc: c_int, argv: [*:null]?[*:0]u8) c_int;
 extern "c" fn ZigLlvmAr_main(argc: c_int, argv: [*:null]?[*:0]u8) c_int;
 
 /// TODO https://github.com/ziglang/zig/issues/3257
-fn punt_to_clang(arena: Allocator, args: []const []const u8) error{OutOfMemory} {
+pub fn punt_to_clang(arena: Allocator, args: []const []const u8) error{OutOfMemory} {
```
andrewrk, line 4096: I don't think these should be made public or used outside this file

andrewrk (changes requested):
Thanks! Left you some review comments.

topolarity:
Thanks for the review, folks!

> Did you check what is still missing? Any of translate-c, ar, dlltool, lib, ranlib ?

These are already covered without this change :+1: Zig calls directly into `ZigLlvmAr_main`

As for what's left: I thought I had covered all of the usage of `ChildProcess` with my most recent commit, but @andrewrk's request to make referencing `spawn()` a compile-time error (probably a good idea, I agree) is still triggering on an unknown reference I haven't been able to suss out yet.

I'll dive in deeper tomorrow and see if I can eliminate the lingering references to `ChildProcess.spawn`, and address the rest of your comments.

src/Compilation.zig
topolarity: Hmm, I had assumed that LLD/Clang couldn't be safely called multiple times, or may never return, due to [a comment in the linker](https://github.com/ziglang/zig/blob/71e0cca7a7957e2f024d2985318e478aa6fb1451/src/link/Elf.zig#L1613-L1615)
> LLD does not behave properly as a library - it calls exit() and does not reset all global data between invocations.

Is that incorrect after all?

andrewrk: It mostly works but it's buggy. I think we should print a warning when invoking LLD for the second time in the same process execution.

andrewrk: Check out 72f6c6e6345030392a3f8cac79862d58359f1e76

topolarity: Great, thanks for the reference. So the preference is to support these calls with warnings, despite the dangers, rather than to block this functionality completely - Makes sense.

I'll go ahead and add non-terminating equivalents of `punt_to_clang` and friends to `src/codegen/llvm.zig`, and update the compiler and linkers to use those on non-spawning OS's

Let me know if you see any issues with that

andrewrk: Sounds mostly good. One suggestion:

Extract the part in `punt_to_clang` that doesn't call exit() into a function called `clangMain`. Make sure the arena memory is freed because we're not going to call exit(). Make *that* function public, leaving `punt_to_clang` private. It can live in main.zig, that's a better place than src/codegen/llvm.zig, which is specifically for lowering AIR to LLVM IR.

andrewrk:
> I'll dive in deeper tomorrow and see if I can eliminate the lingering references to `ChildProcess.spawn`, and address the rest of your comments.

This is a deficiency of zig compile errors; would be nice if it could explain how this dependency is being dragged in. Someday, we'll have it. For now, a workaround is to temporarily delete the spawn function from the std lib and then it tends to generate more helpful errors.

topolarity:
> This is a deficiency of zig compile errors; would be nice if it could explain how this dependency is being dragged in. Someday, we'll have it. For now, a workaround is to temporarily delete the spawn function from the std lib and then it tends to generate more helpful errors.

This tip was a huge help; ended up being short work to track things down.

I believe I've addressed all the feedback in the latest changes now. Let me know if anything still needs work :+1:

zigazeljko:
Why is `argvCmd` needed? Doesn't `std.mem.join` do exactly the same?

topolarity:
> Why is `argvCmd` needed? Doesn't `std.mem.join` do exactly the same?

Nice catch. I updated the PR to use `std.mem.join`

andrewrk:
Excellent work. Congrats on landing your first Zig PR :tada:

PR #10783, opened 2022-02-04
stage1: Fix missing LLD library

splhack:
### Problem and repro step
1. Build LLVM, LLD, and Clang, release 13.0.1 by following https://github.com/ziglang/zig/wiki/How-to-build-LLVM,-libclang,-and-liblld-from-source#posix 
2. Build stage 1 by following https://github.com/ziglang/zig/wiki/Building-Zig-From-Source#instructions
3. linker failed due to undefined symbols in lldMachO.

### Fix
link lldMachO.


andrewrk:
Thanks for taking the time to make a patch.

Zig does not depend on LLD for Mach-O at all. This library is intentionally omitted from Findlld.cmake. If you share the undefined symbols that you are seeing, perhaps we can figure out why you are getting an error.

splhack:
https://gist.github.com/splhack/b4cf053a225e22ad432e6ef35ffc345c
looks like liblldDriver.a (DarwinLdDriver.cpp) depends on liblldMachO.a
https://github.com/splhack/zig/blob/master/cmake/Findlld.cmake#L41

CMake dependency: lld → lib → Driver → DarwinLdDriver.cpp → liblldMachO.a
https://github.com/llvm-mirror/lld/blob/master/CMakeLists.txt#L213
https://github.com/llvm-mirror/lld/blob/master/lib/CMakeLists.txt#L2
https://github.com/llvm-mirror/lld/blob/master/lib/Driver/CMakeLists.txt#L6

andrewrk:
Thank you for the additional information. I think it would be OK to add your patch as a workaround. What OS are you on and what linker are you using? I am interested to learn why you are the first one to run into this problem. It does not occur on any of the systems I have built Zig on.

splhack:
GCC 9 and GNU ld(Binutils) 2.32 on CentOS7

andrewrk:
I guess ld got better at garbage collecting between 2.32 and 2.35.2 :shrug:

PR #10788, opened 2022-02-04
std.process: add option to support single quotes to ArgIteratorGeneral

### LINKED ISSUE #10787: Support single quotes in response files
### Zig Version

0.10.0  (64f7231f86d4b8a155f48087b3f173d8e41b620c)

### Steps to Reproduce

# Prepare response files and test C code

argsfile 

    cat <<EOF > argsfile
    '-o'
    test
    test.c
    EOF

argsfile2

    cat <<EOF > argsfile2
    -o
    'te st'
    test.c
    EOF

test.c

    cat <<EOF > test.c
    int main() {return 0;}
    EOF

### Expected Behavior

# Verify the response files with gcc or Clang

I did with gcc 8.5.0, but most of the gcc versions and also Clang should work

    $ gcc @argsfile; echo $?
    0
    $ gcc @argsfile2; echo $?
    0

### Actual Behavior

# Test zig cc with the response files

    zig version
    zig cc @argsfile; echo $?
    zig cc @argsfile2; echo $?

result

    $ zig version
    0.10.0

    $ zig cc @argsfile; echo $?
    warning: FileNotFound: /tmp/'-o'
    error: FileNotFound
    1

    $ zig cc @argsfile2; echo $?
    warning: FileNotFound: /tmp/st'
    error: FileNotFound
    1

# Context
This exact issue was reported in #4833.
I attempted to test it with #9654 that should be merged on to trunk https://github.com/ziglang/zig/commit/8a97807d6812f62db4c3088fecd04a98a84b9943

Vexu:
Closes  #10787

Not sure whether we should accept single quotes for windows, I'm guessing not?

### DIFF
--- a/lib/std/process.zig
+++ b/lib/std/process.zig
@@ -303,7 +303,8 @@ pub const ArgIteratorWasi = struct {
 
 /// Optional parameters for `ArgIteratorGeneral`
 pub const ArgIteratorGeneralOptions = struct {
-    comments_supported: bool = false,
+    comments: bool = false,
+    single_quotes: bool = false,
 };
 
 /// A general Iterator to parse a string into a set of arguments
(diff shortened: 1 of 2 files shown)


splhack:
verified this (on top of 64f7231f86d4b8a155f48087b3f173d8e41b620c) fixed #10787 🎉

andrewrk (approved):
> Not sure whether we should accept single quotes for windows, I'm guessing not?

I don't know the answer to this either, but your patch makes it as simple as toggling a flag, so it's progress either way!

PR #10791, opened 2022-02-04
Dry run of spawn

syrusakbary:
This PR is in-progress for fixing #10750.
Places where spawn should now work:
* [ ] `zig test` - instead print a message about not being able to spawn child processes and exit success.
* [x] `zig run` - instead print "unable to spawn child processes" and exit failure.
* [ ] for clang we can invoke clang_main directly instead of as a child process
* [ ] for lld we can invoke the lld library functions instead of as child process
* [ ] same thing for `zig ar`, `zig dlltool`, etc.
* [ ] for detecting native libc, just return error.LibcNotDetected or whatever the appropriate error is.

Note: This is the first time I'm trying Zig so I want to make sure I do things according to what's expected, looking forward to the feedback!

### DIFF
--- a/lib/std/build/RunStep.zig
+++ b/lib/std/build/RunStep.zig
@@ -153,6 +153,10 @@ fn stdIoActionToBehavior(action: StdIoAction) std.ChildProcess.StdIo {
 }
 
 fn make(step: *Step) !void {
+    if (!os.can_spawn()) {
+        std.debug.print("Unable to spawn {s}: the current OS doesn't support it\n", .{argv[0]});
+        return;
+    }
     const self = @fieldParentPtr(RunStep, "step", step);
 
     const cwd = if (self.cwd) |cwd| self.builder.pathFromRoot(cwd) else self.builder.build_root;
--- a/lib/std/build/RunStep.zig
+++ b/lib/std/build/RunStep.zig
@@ -176,7 +176,7 @@ fn make(step: *Step) !void {
 
     const argv = argv_list.items;
 
-    if (!os.can_spawn()) {
+    if (!os.can_spawn) {
         std.debug.print("Unable to spawn {s}: the current OS doesn't support it\n", .{argv[0]});
         return;
     }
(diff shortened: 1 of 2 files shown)


lib/std/build/RunStep.zig
```diff
@@ -153,6 +153,10 @@ fn stdIoActionToBehavior(action: StdIoAction) std.ChildProcess.StdIo {
 }
 
 fn make(step: *Step) !void {
+    if (!os.can_spawn()) {
+        std.debug.print("Unable to spawn {s}: the current OS doesn't support it\n", .{argv[0]});
+        return;
+    }
```
syrusakbary, line 159: I'm assuming the compiler will throw everything else away, so spawn is not called and used. But I'd love to get confirmation to see if we need to wrap things in an `else`.

lib/std/build/RunStep.zig
```diff
@@ -175,6 +176,11 @@ fn make(step: *Step) !void {
 
     const argv = argv_list.items;
 
+    if (!os.can_spawn()) {
+        std.debug.print("Unable to spawn {s}: the current OS doesn't support it\n", .{argv[0]});
+        return;
+    }
```
syrusakbary, line 182: I'm assuming the compiler will throw everything else away, so spawn is not called and used. But I'd love to get confirmation to see if we need to wrap things in an else.

matu3ba:
Hey @syrusakbary, did you take a look into #10782 trying to fix the same issue?

lib/std/build/RunStep.zig
```diff
@@ -175,6 +176,11 @@ fn make(step: *Step) !void {
 
     const argv = argv_list.items;
 
+    if (!os.can_spawn) {
+        std.debug.print("Unable to spawn {s}: the current OS doesn't support it\n", .{argv[0]});
+        return;
+    }
+
```
syrusakbary, line 183: I'm assuming the compiler will throw everything else away, so spawn is not called and used. But I'd love to get confirmation to see if we need to wrap things in an else.

syrusakbary:
Ooops, didn't realize of that PR! Thanks for the ping, I'll close this one!

PR #10794, opened 2022-02-05
compiler_rt: optimize mulo

matu3ba:
- use usize to decide if register size is big enough to store
  multiplication result or if division is necessary
- multiplication routine with check of integer bounds
- wrapping multipliation and division routine from Hacker's Delight

### DIFF
--- a/lib/std/special/compiler_rt/mulo.zig
+++ b/lib/std/special/compiler_rt/mulo.zig
@@ -1,67 +1,68 @@
 const builtin = @import("builtin");
+const std = @import("std");
+const math = std.math;
 
 // mulo - multiplication overflow
-// - muloXi4_generic for unoptimized version
+// * return a*b.
+// * return if a*b overflows => 1 else => 0
+// - muloXi4_genericSmall as default
+// - muloXi4_genericFast for 2*bitsize <= usize
 
-// return a*b.
-// return if a*b overflows => 1 else => 0
-// see https://stackoverflow.com/a/26320664 for possible implementations
-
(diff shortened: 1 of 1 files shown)


matu3ba:
Justification: speedup by 17x on ReleaseFast vs the llvm implementation.
[Benchmarks here](https://gist.github.com/matu3ba/eb52a8132cd4fc39b1836275341b8ce5)

```txt
Benchmark 1: ./mulo_fast
  Time (mean ± σ):      1.008 s ±  0.009 s    [User: 1.006 s, System: 0.001 s]
  Range (min … max):    0.996 s …  1.027 s    10 runs
Benchmark 2: ./mulo_hack
  Time (mean ± σ):      1.256 s ±  0.004 s    [User: 1.250 s, System: 0.001 s]
  Range (min … max):    1.248 s …  1.260 s    10 runs

Benchmark 3: ./mulo_llvm
  Time (mean ± σ):     17.403 s ±  0.155 s    [User: 17.363 s, System: 0.005 s]
  Range (min … max):   17.314 s … 17.836 s    10 runs
  ```
  and for 1_000_000_000 elements (too slow to wait for llvm):
  ```txt
  enchmark 1: ./mulo_fast
  Time (mean ± σ):     21.826 s ±  1.182 s    [User: 21.791 s, System: 0.001 s]
  Range (min … max):   20.844 s … 24.191 s    10 runs

Benchmark 2: ./mulo_hack
 ⠙ Current estimate: 26.246 s
  ```

Issue #10784, opened 2022-02-04
make AIR instruction for assembly not reference ZIR


andrewrk:
https://github.com/ziglang/zig/blob/1b6a1e691fab75ce40c9d0c4015c9f4a46b72aa4/src/Air.zig#L665-L669

This is problematic because it causes

```
thread 345550 panic: access of inactive union field
/home/andy/dev/zig/src/Liveness.zig:434:83: 0x319d323 in Liveness.analyzeInst (zig)
            const extended = a.zir.instructions.items(.data)[extra.data.zir_index].extended;
                                                                                  ^
```

Problem is that the ZIR we pass to Liveness analysis (and to codegen later) is for a particular Decl that could have inlined code, causing an assembly ZIR instruction from a *different file* to be in the AIR for the Decl. This means we are looking up a ZIR instruction index in the wrong ZIR object.

This is the only AIR instruction that references a ZIR instruction.

The solution is pretty straightforward: instead of trying to cut corners, change the AIR encoding to duplicate all the info from the ZIR that it needs, and uphold the property that ZIR is not needed in order to decode AIR.

PR #10795, opened 2022-02-05
add Aro as a C frontend

andrewrk:
Problems to solve in this branch before merging:

 * Liveness and AIR printing want a ZIR instance. This issue is
   addressed by #10784.
 * linker: updateFunc / updateDecl want a Module instance which
   currently represents only Zig code. These functions should be changed
   to accept a Compilation, not a Module. The relevant state that it
   needed to touch on Module should be moved to Compilation instead.
 * `Decl` needs to be shared by both the C frontend and the Zig
   frontend. It currently has Zig-only fields and the functions to
   create Decl objects don't really fit the API that the C frontend
   needs.
 * A lot of the incremental compilation infrastructure doesn't quite
   match up, for example Decl objects being owned by Namespaces which is
   a Zig concept. Maybe there could be 1 Namespace globally across all C
   files and 1 more Namespace per C file for functions and globals
   declared `static`.
 * There is common code to be extracted from Sema that is shared with
   Aro frontend. All the helper functions having to do with the
   air_instructions and air_extra arrays.

There are many instances of calling `@panic("TODO")` that need to be
cleaned up.

There is a lot of compiler options that Zig is not communicating to Aro,
for example we do not pass the -D switches that we pass to Clang yet.
Need to audit `addCCArgs` and do the equivalent things for setting up
the Aro Compilation.

The Aro code is copied from https://github.com/Vexu/arocc, commit
7a0e54227e1ea2b2df8aa7bd2b7075f735109317. The only patches are to delete
Codegen and replace it with our own, and to delete main.zig, and add
Type and Value to lib.zig.

I believe the current plan is for main Aro development to happen
upstream on the arocc repository and periodically sync it downstream
with the Zig repository. It's up to @Vexu whether to keep that process
or change it in the future.


andrewrk:
Current status on hello world:

```
$ ./zig-out/bin/zig build-exe hello.c -lc empty.zig --verbose-air
# Begin Function AIR: main:
# Total AIR+Liveness bytes: 257B
# AIR Instructions:         9 (81B)
# AIR Extra Data:           6 (24B)
# AIR Values Bytes:         4 (32B)
# Liveness tomb_bits:       8B
  %0 = const_ty(*const fn([*c]i8, ...) callconv(.C) c_int)
  %1 = constant(*const fn([*c]i8, ...) callconv(.C) c_int, (decl ref))
  %2 = const_ty([15]u8)
  %3 = constant([15]u8, "Hello, World!\n\x00")
  %4 = const_ty(*const [15:0]u8)
  %5 = constant(*const [15:0]u8, (decl ref))
  %7 = constant(c_int, 0)

  %6!= call(%1!, [%5!])
  %8!= ret(%7!)
# End Function AIR: main

thread 489784 panic: Segmentation fault at address 0x8
/home/andy/dev/zig/src/Module.zig:751:34: 0x3faba9 in Module.Decl.getFileScope (zig)
        return decl.src_namespace.file_scope;
                                 ^
/home/andy/dev/zig/src/Module.zig:614:44: 0x59e852 in Module.Decl.nodeOffsetSrcLoc (zig)
            .file_scope = decl.getFileScope(),
                                           ^
/home/andy/dev/zig/src/Module.zig:609:37: 0x59a205 in Module.Decl.srcLoc (zig)
        return decl.nodeOffsetSrcLoc(0);
                                    ^
/home/andy/dev/zig/src/link/Elf.zig:2690:69: 0x706c57 in link.Elf.updateFunc (zig)
    const res = try codegen.generateFunction(&self.base, decl.srcLoc(), func, air, liveness, &code_buffer, .{
                                                                    ^
/home/andy/dev/zig/src/link.zig:453:77: 0x6fde29 in link.File.updateFunc (zig)
            .elf   => return @fieldParentPtr(Elf,   "base", base).updateFunc(module, func, air, liveness),
                                                                            ^
/home/andy/dev/zig/src/aro/Codegen.zig:314:26: 0x739e16 in aro.Codegen.genFn (zig)
    c.bin_file.updateFunc(mod, module_fn, air, liveness) catch |err| switch (err) {
                         ^
/home/andy/dev/zig/src/aro/Codegen.zig:84:27: 0x7108b1 in aro.Codegen.generateTree (zig)
            => try c.genFn(decl_node),
                          ^
/home/andy/dev/zig/src/Compilation.zig:3222:33: 0x6fea91 in Compilation.compileWithAro (zig)
    try aro.Codegen.generateTree(comp, &aro_comp, tree, arena);
                                ^
/home/andy/dev/zig/src/Compilation.zig:2832:32: 0x58df44 in Compilation.processOneJob (zig)
            comp.compileWithAro(c_source_file) catch |err| switch (err) {
                               ^
/home/andy/dev/zig/src/Compilation.zig:2691:30: 0x57a716 in Compilation.performAllTheWork (zig)
            try processOneJob(self, work_item, main_progress_node);
                             ^
```

I'm starting to wonder... should we perhaps lower to ZIR instead of AIR? The integration would be a lot less messy, and it's much more obvious how to parallelize it. One .c file would produce exactly one corresponding ZIR file. :thinking: :thinking: :thinking:

andrewrk:
Keeping the branch alive but closing the PR as I don't intend to merge this any time soon.

PR #10796, opened 2022-02-05
Envmap

marler8997:
This is my take on fixing #4603 which is an alternative to https://github.com/ziglang/zig/pull/10650

This PR includes commits from that PR and modifies it to a different design.  In this PR, instead of having `EnvMap` store environment variables in both their original casing and a canonicalized casing, this one keeps the original casing and modifies the hash map to ignore case.

### DIFF
--- a/lib/std/os/windows/ntdll.zig
+++ b/lib/std/os/windows/ntdll.zig
@@ -229,13 +229,13 @@ pub extern "ntdll" fn RtlEqualUnicodeString(
     CaseInSensitive: BOOLEAN,
 ) callconv(WINAPI) BOOLEAN;
 
-pub extern "NtDll" fn RtlUpcaseUnicodeString(
+pub extern "ntdll" fn RtlUpcaseUnicodeString(
     DestinationString: *UNICODE_STRING,
     SourceString: *const UNICODE_STRING,
     AllocateDestinationString: BOOLEAN,
 ) callconv(WINAPI) NTSTATUS;
 
-pub extern "NtDll" fn RtlUpcaseUnicodeChar(
+pub extern "ntdll" fn RtlUpcaseUnicodeChar(
     SourceCharacter: u16,
 ) callconv(WINAPI) u16;
 
--- a/lib/std/os/windows/ntdll.zig
+++ b/lib/std/os/windows/ntdll.zig
@@ -235,6 +235,10 @@ pub extern "NtDll" fn RtlUpcaseUnicodeString(
     AllocateDestinationString: BOOLEAN,
 ) callconv(WINAPI) NTSTATUS;
 
+pub extern "NtDll" fn RtlUpcaseUnicodeChar(
+    SourceCharacter: u16,
+) callconv(WINAPI) u16;
+
 pub extern "ntdll" fn NtLockFile(
     FileHandle: HANDLE,
     Event: ?HANDLE,
--- a/lib/std/os/windows/ntdll.zig
+++ b/lib/std/os/windows/ntdll.zig
@@ -229,12 +229,6 @@ pub extern "ntdll" fn RtlEqualUnicodeString(
     CaseInSensitive: BOOLEAN,
 ) callconv(WINAPI) BOOLEAN;
 
-pub extern "ntdll" fn RtlUpcaseUnicodeString(
-    DestinationString: *UNICODE_STRING,
-    SourceString: *const UNICODE_STRING,
-    AllocateDestinationString: BOOLEAN,
-) callconv(WINAPI) NTSTATUS;
-
 pub extern "ntdll" fn RtlUpcaseUnicodeChar(
     SourceCharacter: u16,
 ) callconv(WINAPI) u16;
--- a/lib/std/os/windows/ntdll.zig
+++ b/lib/std/os/windows/ntdll.zig
@@ -229,6 +229,12 @@ pub extern "ntdll" fn RtlEqualUnicodeString(
     CaseInSensitive: BOOLEAN,
 ) callconv(WINAPI) BOOLEAN;
 
+pub extern "NtDll" fn RtlUpcaseUnicodeString(
+    DestinationString: *UNICODE_STRING,
+    SourceString: *const UNICODE_STRING,
+    AllocateDestinationString: BOOLEAN,
+) callconv(WINAPI) NTSTATUS;
+
 pub extern "ntdll" fn NtLockFile(
     FileHandle: HANDLE,
     Event: ?HANDLE,
--- a/lib/std/process.zig
+++ b/lib/std/process.zig
@@ -255,19 +255,21 @@ pub fn getEnvMap(allocator: Allocator) !EnvMap {
 
             while (ptr[i] != 0 and ptr[i] != '=') : (i += 1) {}
             const key_w = ptr[key_start..i];
+            const key = try std.unicode.utf16leToUtf8Alloc(allocator, key_w);
+            errdefer allocator.free(key);
 
             if (ptr[i] == '=') i += 1;
 
             const value_start = i;
             while (ptr[i] != 0) : (i += 1) {}
             const value_w = ptr[value_start..i];
-
-            try result.storage.putUtf16NoClobber(key_w, value_w);
+            const value = try std.unicode.utf16leToUtf8Alloc(allocator, value_w);
+            errdefer allocator.free(value);
 
             i += 1; // skip over null byte
-        }
 
-        try result.storage.reallocUppercaseBuf();
+            try result.putMove(key, value);
+        }
         return result;
     } else if (builtin.os.tag == .wasi and !builtin.link_libc) {
         var environ_count: usize = undefined;
--- a/lib/std/process.zig
+++ b/lib/std/process.zig
@@ -63,26 +63,42 @@ pub const EnvMap = struct {
     );
 
     pub const EnvNameHashContext = struct {
+        fn upcase(c: u21) u21 {
+            if (c <= std.math.maxInt(u16))
+                return std.os.windows.ntdll.RtlUpcaseUnicodeChar(c);
+            return c;
+        }
+
         pub fn hash(self: @This(), s: []const u8) u64 {
             _ = self;
             if (builtin.os.tag == .windows) {
                 const h = std.hash.Wyhash.init(0);
-                // TODO: improve this, instead of iterating over ascii,
-                //       iterate over with unicode
-                for (s) |c| {
-                    var s_upper = [_]u8 { std.ascii.toLower(c) };
-                    h.update(s_upper);
+                var it = std.unicode.Utf8View(s).iterator();
+                while (it.nextCodepoint()) |cp| {
+                    const cp_upper = upcase(cp);
+                    h.update(&[_]u8{
+                        @intCast(u8, (cp_upper >> 16) & 0xff),
+                        @intCast(u8, (cp_upper >> 8) & 0xff),
+                        @intCast(u8, (cp_upper >> 0) & 0xff),
+                    });
                 }
                 return h.final();
             }
             return std.hash_map.hashString(s);
         }
+
         pub fn eql(self: @This(), a: []const u8, b: []const u8) bool {
             _ = self;
             if (builtin.os.tag == .windows) {
-                // TODO: improve this, instead of comparing ascii
-                //       compare with unicode
-                return std.ascii.eqlIgnoreCase(a, b);
+                var it_a = std.unicode.Utf8View(a).iterator();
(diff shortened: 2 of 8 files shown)


lib/std/process.zig
```diff
@@ -53,9 +52,194 @@ test "getCwdAlloc" {
...
+            if (builtin.os.tag == .windows) {
+                var it_a = std.unicode.Utf8View.initUnchecked(a).iterator();
+                var it_b = std.unicode.Utf8View.initUnchecked(b).iterator();
+                while (true) {
+                    const c_a = it_a.nextCodepoint() orelse break;
+                    const c_b = it_b.nextCodepoint() orelse return false;
+                    if (upcase(c_a) != upcase(c_b))
+                        return false;
+                }
+                if (it_b.nextCodepoint()) |_| return false;
```
squeek502, line 103:
```suggestion
                if (it_b.nextCodepoint()) |_| return false;
                return true;
```

Without the `return true` here, it falls down to `eqlString` which is case sensitive.

squeek502, line 103:
```suggestion
                if (it_b.nextCodepoint()) |_| return false;
                return true;
```

Without the `return true` here, it falls down to `eqlString` which is case sensitive.

squeek502, line 103:
```suggestion
                if (it_b.nextCodepoint()) |_| return false;
                return true;
```

Without the `return true` here, it falls down to `eqlString` which is case sensitive.

squeek502, line 103:
```suggestion
                if (it_b.nextCodepoint()) |_| return false;
                return true;
```

Without the `return true` here, it falls down to `eqlString` which is case sensitive.

squeek502, line 103:
```suggestion
                if (it_b.nextCodepoint()) |_| return false;
                return true;
```

Without the `return true` here, it falls down to `eqlString` which is case sensitive.

squeek502, line 103:
```suggestion
                if (it_b.nextCodepoint()) |_| return false;
                return true;
```

Without the `return true` here, it falls down to `eqlString` which is case sensitive.

squeek502, line 103:
```suggestion
                if (it_b.nextCodepoint()) |_| return false;
                return true;
```

Without the `return true` here, it falls down to `eqlString` which is case sensitive.

squeek502, line 103:
```suggestion
                if (it_b.nextCodepoint()) |_| return false;
                return true;
```

Without the `return true` here, it falls down to `eqlString` which is case sensitive.

squeek502, line 103:
```suggestion
                if (it_b.nextCodepoint()) |_| return false;
                return true;
```

Without the `return true` here, it falls down to `eqlString` which is case sensitive.

lib/std/process.zig
```diff
@@ -53,9 +52,194 @@ test "getCwdAlloc" {
...
+        const is_an_expected_name = std.mem.eql(u8, "SOMETHING_NEW", entry.key_ptr.*) or std.mem.eql(u8, "SOMETHING_NEW_AND_LONGER", entry.key_ptr.*);
+        try testing.expect(is_an_expected_name);
+        count += 1;
+    }
+    try testing.expectEqual(@as(EnvMap.Size, 2), count);
+
+    env.remove("SOMETHING_NEW");
+    try testing.expect(env.get("SOMETHING_NEW") == null);
+
+    try testing.expectEqual(@as(EnvMap.Size, 1), env.count());
```
squeek502, line 234:
```suggestion
    try testing.expectEqual(@as(EnvMap.Size, 1), env.count());

    // test Unicode case-insensitivity on Windows
    if (builtin.os.tag == .windows) {
        try env.put("КИРиллИЦА", "something else");
        try testing.expectEqualStrings("something else", env.get("кириллица").?);
    }
```

squeek502:
Did some benchmarking comparing the implemention of this PR vs #10650 (some context: https://github.com/ziglang/zig/pull/10650#issuecomment-1029782259)

This PR is in blue, #10650 is in orange:

insert-2022-02-07
get-found-2022-02-07
get-missing-2022-02-07
remove-2022-02-07
memory-2022-02-07

Benchmark code is here: https://gist.github.com/squeek502/13fdaa4255c6c5e926ceb00a728e1ee1

Some notes:
- On my machine, the only benchmark that this PR wins in is `get` of missing keys when using Unicode strings. For all the rest, 10650 either wins or has comparable performance.
- 10650 uses about ~2x more memory across the board.
- The Unicode benchmarks insert random UTF-8 strings and call `get` with UTF-8 strings, and likewise the ASCII benchmarks insert random ASCII strings and call `get` with ASCII strings. That is, the benchmarks do not mix/match ASCII/UTF-8. Doing so might better simulate real-world use-cases (i.e. an EnvMap with mostly/all ASCII keys and then seeing how fast lookups of (missing) UTF-8 strings are).

marler8997:
These benchmarks are good data. I feel I should also share some of the things I think matter when it comes to deciding on a design like this.

* Perfomance
* Code size
* Runtime memory usage
* Code simplicity/maintainability

In the case of environment variables, I would actually put runtime perfomance lower in the list of importance than with other types of modules because I don't think managing environment variables would usually be in the critical path of performance. Still important, but not the only criteria to consider. When I was younger I use to always put performance at the top of my list, but now I realize how complex performance is and now I tend to start with the "simplest" design and only optimize for performance when the simplest design isn't fast enough. This not only results in less work, it also ends up having faster code overall because once you have a working design, you don't need to guess what will be faster, you can actually benchmark the application that isn't performing fast enough. This focuses your efforts on optimizing code that will actually make you application faster instead of trying to shoot fish in a barrel.

All this being said, I find the benchmarks interesting and we can use that data to help guide us if we find that environment variables management performance becomes an issue.  Adding a mechanism to cache string conversion like you've done trades higher memory usage and higher complexity for less calculations, which can increase performance in certain situations. But to reiterate, I think it would be best to leave a performance optimization like that out until we see that it has a noticeable benefit, and not at the cost of making other usages slower if those usages are just as common.

matu3ba:
Unifying directory path traversal (without opening directories for perf) on Linux and Windows has the same problem, so it looks to me like the performance optimization could go into a lib for doing string stuff on paths (without the syscall overhead).

squeek502:
@matu3ba could you be more specific? I'm unsure what performance optimization you're talking about or how it would apply to path traversal.

matu3ba:
@squeek502 My intention was to provide a faster iterator that works forward and backwards on paths `/home/user/bla/blubb\/bla/../file` represented as strings to get the directory names without asking the Kernel first.

matu3ba:
> Faster than what?

I think the idea is stupid, so never mind.

lib/std/process.zig
```diff
@@ -53,9 +52,200 @@ test "getCwdAlloc" {
...
+            if (c <= std.math.maxInt(u16))
+                return std.os.windows.ntdll.RtlUpcaseUnicodeChar(@intCast(u16, c));
+            return c;
+        }
+
+        pub fn hash(self: @This(), s: []const u8) u64 {
+            _ = self;
+            if (builtin.os.tag == .windows) {
+                var h = std.hash.Wyhash.init(0);
+                var it = std.unicode.Utf8View.initUnchecked(s).iterator();
```
Vexu, line 78: Using `initUnchecked` in the hash context without validating the key in `put`/`get` is a really bad idea as `Utf8Iterator` does a whole bunch of `catch unreachable`.

Vexu (changes requested):

lib/std/process.zig
marler8997: This is intentional because the utf8 string in this map come from decoding utf16 strings on Windows.

Vexu: `child_process.env_map.put(user_provided_key, some_value)` seems like something you could find. It should at least be documented that on Windows the keys must be valid utf-8.

marler8997: Ok I've documented that the `key` argument for all functions must be valid UTF8 on windows.

lib/std/os/windows/ntdll.zig
```diff
@@ -223,6 +223,10 @@ pub extern "ntdll" fn RtlWaitOnAddress(
     Timeout: ?*const LARGE_INTEGER,
 ) callconv(WINAPI) NTSTATUS;
 
+pub extern "NtDll" fn RtlUpcaseUnicodeChar(
```
Vexu, line 226:
```suggestion
pub extern "ntdll" fn RtlUpcaseUnicodeChar(
```
3c2eddae5a4e89ec428a7b3b9f089c32f57b17af

Vexu (approved):

squeek502:
@Vexu if the usage of `ntdll`'s `Rtl` Unicode functions for Windows environment variables is accepted, then #10595 can act as a companion to these EnvMap changes to fully address #4603

marler8997:
Yeah I think we'll need clarification from @andrewrk on what he meant by https://github.com/ziglang/zig/commit/8bf7cffe29b520cda094756309bca1d354af0d6b:

> It is not planned to support full Unicode case-insensitivity
on Windows, and in fact relying on non-ASCII case-insensitive
environment variables is fundamentally problematic.

This PR uses `RtlUpcaseUnicodeChar` to make string comparison case insensitive for environment variable keys on Windows.  I'm not sure why this is problematic.

squeek502:
> Either Andrew knows something I'm not aware of, or he had another solution in mind that does have a problem.

My assumption is that he was assuming it would require putting Unicode data in the standard library.

marler8997:
> My assumption is that he was assuming it would require putting Unicode data in the standard library.

Yeah that could definitely be it.  Out of curiosity, does anyone know how Windows implements `RtlUpcaseUnicodeChar`?  Where is it keeping all the unicode data at runtime?  Maybe it's in the kernel or something?

squeek502:
> Yeah that could definitely be it. Out of curiosity, does anyone know how Windows implements `RtlUpcaseUnicodeChar`? Where is it keeping all the unicode data at runtime? Maybe it's in the kernel or something?

My guess is [`PEB.UnicodeCaseTableData`](https://github.com/ziglang/zig/blob/8841a71aa675f76c0ff7658339872a5faa5e4d5b/lib/std/os/windows.zig#L3342).

andrewrk:
I was unaware that NtDll had functions for environment variables when I made that comment. This changes my position.

What I do not want to do is have the Zig std lib depend on shipping with Unicode.txt or similar data. However, if NtDll provides us with an API that allows us to match the Windows-native method of accessing environment variables, then I am in favor of using that API. This really has nothing to do with proper Unicode support; really what's happening is that Windows has some set of weird rules about what environment variables do or don't match via string equality, and we just have to match those rules by calling the appropriate NtDll functions.

For example, the `RtlUpcaseUnicodeChar` function is completely wrong for trying to implement proper Unicode support because it lacks awareness of various things, but it may be correct in matching environment variable name string equality to be what is expected behavior on Windows.

squeek502:
> Can we confirm that is indeed the case?

Unsure if this is what you're looking for, but for ReactOS:

- [`GetEnvironmentVariableW` calls `RtlQueryEnvironmentVariable_U`](https://github.com/reactos/reactos/blob/37bc0a5a08613cded15bea71c27d0d06085b42a7/reactos/dll/win32/kernel32/client/environ.c#L194)
- [`RtlQueryEnvironmentVariable_U` calls `RtlEqualUnicodeString`](https://github.com/reactos/reactos/blob/3fa57b8ff7fcee47b8e2ed869aecaf4515603f3f/sdk/lib/rtl/env.c#L560)
- [`RtlEqualUnicodeString` calls `RtlCompareUnicodeString`](https://github.com/reactos/reactos/blob/3fa57b8ff7fcee47b8e2ed869aecaf4515603f3f/sdk/lib/rtl/unicode.c#L411)
- [`RtlCompareUnicodeString` uses `RtlpUpcaseUnicodeChar`](https://github.com/reactos/reactos/blob/3fa57b8ff7fcee47b8e2ed869aecaf4515603f3f/sdk/lib/rtl/unicode.c#L2199), which is an internal version of [`RtlUpcaseUnicodeChar`, which in turn just calls `RtlpUpcaseUnicodeChar`](https://github.com/reactos/reactos/blob/3fa57b8ff7fcee47b8e2ed869aecaf4515603f3f/sdk/lib/rtl/nls.c#L745)

This PR uses `RtlUpcaseUnicodeChar` directly which should be equivalent.

andrewrk:
@squeek502 I am satisfied with this. Thank you for digging!

marler8997:
> 2. What was the CI failure?

Just rebased and repushed, looks like the Azure CI failure is no more.  Probably an intermittent failure from long ago.

PR #10801, opened 2022-02-05
stage2: implement @sqrt for f{16,32,64}

schmee:
This is my first contribution to stage2 and hopefully it's not a complete mess, all feedback is appreciated!

Support for f128, comptime_float, and c_longdouble require improvements to compiler_rt and will implemented in a later PR. Some of the code in this commit could be made more generic, for instance `llvm.airSqrt` could probably be `llvm.airUnaryMath`, but let's cross that bridge when we get to it.

My tentative plan going forward is:

1. Implement a bunch more unary operators (`@floor`, `@ceil`, `@sin`, `@cos`) for `f{16,32,64}` to get a feel for how it's going to look with multiple operators implemented.
2. Try to understand how compiler_rt works, and make the necessary changes to support f128 for the operators implemented in Step 1.
3. Implement the remaining operators for all float types.

Does that sound reasonable @andrewrk?

### DIFF
--- a/src/arch/riscv64/CodeGen.zig
+++ b/src/arch/riscv64/CodeGen.zig
@@ -507,6 +507,8 @@ fn genBody(self: *Self, body: []const Air.Inst.Index) InnerError!void {
             .max             => try self.airMax(inst),
             .slice           => try self.airSlice(inst),
 
+            .sqrt            => try self.airUnaryMath(inst),
+
             .add_with_overflow => try self.airAddWithOverflow(inst),
             .sub_with_overflow => try self.airSubWithOverflow(inst),
             .mul_with_overflow => try self.airMulWithOverflow(inst),
@@ -1164,6 +1166,15 @@ fn airPopcount(self: *Self, inst: Air.Inst.Index) !void {
     const ty_op = self.air.instructions.items(.data)[inst].ty_op;
     const result: MCValue = if (self.liveness.isUnused(inst)) .dead else return self.fail("TODO implement airPopcount for {}", .{self.target.cpu.arch});
     return self.finishAir(inst, result, .{ ty_op.operand, .none, .none });
+}
+
+fn airUnaryMath(self: *Self, inst: Air.Inst.Index) !void {
+    const un_op = self.air.instructions.items(.data)[inst].un_op;
+    const result: MCValue = if (self.liveness.isUnused(inst))
+        .dead
+    else
+        return self.fail("TODO implement airUnaryMath for {}", .{self.target.cpu.arch});
+    return self.finishAir(inst, result, .{ un_op, .none, .none });
 }
 
 fn reuseOperand(self: *Self, inst: Air.Inst.Index, operand: Air.Inst.Ref, op_index: Liveness.OperandInt, mcv: MCValue) bool {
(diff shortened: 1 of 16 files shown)


m-radomski:
Dang! You outran me by 40 minutes, I just started working on this today :D I just want to comment on f128 issue, I noticed that the builtin `@sqrt` does not work on f128 types inside the stage2 compiler. By that I mean values that are f16-f64 are squared fine, but values that are comptime_float or f128 fail to sqaure.

```
here doing sqrt at comp time sqrt(1.44e+02) = 1.2e+01   // f16
here doing sqrt at comp time sqrt(1.69e+02) = 1.3e+01   // f32
here doing sqrt at comp time sqrt(1.69e+02) = 1.3e+01   // f64
here doing sqrt at comp time sqrt(1.96e+02) = 1.96e+02  // f128
```

Because of that the lower part of `@sqrt` test fails:
```
const x = 14.0;
const y = x + 0.4;
const z = @round(y);
comptime try expectEqual(x, z); // z=196.0 for some reason
```

By the looks of `ir_eval_float_op` I presume that squares of f128 ought to work.

schmee:
@m-radomski haha, damn! 😄 Fortunately there's many more ops to implement, maybe we can sync up so we don't redo each other's work?

Indeed, I noticed the same thing about f128 and came to the same conclusion. I'll try to figure it out and resolve it in a future PR, if you have any ideas or want to discuss I have the same username in the Zig Discord 👍

src/arch/riscv64/CodeGen.zig
```diff
@@ -1159,6 +1161,12 @@ fn airPopcount(self: *Self, inst: Air.Inst.Index) !void {
     return self.finishAir(inst, result, .{ ty_op.operand, .none, .none });
 }
 
+fn airUnaryMath(self: *Self, inst: Air.Inst.Index) !void {
+    const un_op = self.air.instructions.items(.data)[inst].un_op;
+    const result: MCValue = if (self.liveness.isUnused(inst)) .dead else return self.fail("TODO implement airUnaryMath for {}", .{self.target.cpu.arch});
```
kubkon, line 1166: Could you please re-format this into multiple lines like you did for `x86_64` backend?

schmee: I made this function match whatever the other ones were like for the respective backends. Would you like me to re-format anyway?

schmee:
Just discovered that `llvm.zig` has a `getIntrinsics` function, gonna update the code to use that 👍

src/arch/riscv64/CodeGen.zig
kubkon: Yes please.

andrewrk:
Excellent work.

I pushed a fixup commit on top with some guidance on how to proceed with the other unary math operations.

It might make sense to change the AIR instruction from `sqrt` to being `unary_float_op` because as is we're only using 1 of the available 2 fields. Instead of `un_op` there could be a new `float_op` union field that has the operand and an enum tag telling which float function should be invoked. This will avoid using too many AIR tags (in our space of 256 possible values) at the cost of 1 additional branch in the backends (since it will have to branch on the AIR tag and then also the float op after that).

PR #10802, opened 2022-02-05
stage2 regalloc: Introduce error.OutOfRegisters

joachimschmidt557:
>  * Runtime crashes are better than bugs.

Previously, when the register manager ran out of registers, it would just return registers as if nothing went wrong, except that the registers which couldn't be allocated were `undefined`.

Also improves documentation for `freezeRegs` and `unfreezeRegs` with some comprehensive examples.


kubkon (approved):
Very cool!

PR #10803, opened 2022-02-05
stage2: store externs lib name as part of decl

kubkon:
* add new `Decl` subtype, `Decl.ExternFn` - `ExternFn` will contain a maybe-lib-name if it was defined with the `extern` keyword like so

```zig
```

`lib_name` will live as long as `ExternFn` decl does.

---

TODO
- [x] handle lib name in extern variable

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -510,6 +510,7 @@ pub const Decl = struct {
             gpa.destroy(func);
         }
         if (decl.getVariable()) |variable| {
+            variable.deinit(gpa);
             gpa.destroy(variable);
         }
         if (decl.value_arena) |arena_state| {
@@ -694,6 +695,8 @@ pub const Decl = struct {
         return func;
     }
 
+    /// If the Decl has a value and it is an extern function, returns it,
+    /// otherwise null.
     pub fn getExternFn(decl: *const Decl) ?*ExternFn {
         if (!decl.owns_tv) return null;
         const extern_fn = (decl.val.castTag(.extern_fn) orelse return null).data;
@@ -701,6 +704,8 @@ pub const Decl = struct {
         return extern_fn;
     }
 
+    /// If the Decl has a value and it is a variable, returns it,
+    /// otherwise null.
     pub fn getVariable(decl: *Decl) ?*Var {
         if (!decl.owns_tv) return null;
         const variable = (decl.val.castTag(.variable) orelse return null).data;
@@ -1469,9 +1474,20 @@ pub const Var = struct {
     init: Value,
     owner_decl: *Decl,
 
+    /// Library name if specified.
+    /// For example `extern "c" var stderrp = ...` would have 'c' as library name.
+    /// Allocated with Module's allocator; outlives the ZIR code.
+    lib_name: ?[*:0]const u8,
+
     is_extern: bool,
     is_mutable: bool,
     is_threadlocal: bool,
+
+    pub fn deinit(variable: *Var, gpa: Allocator) void {
+        if (variable.lib_name) |lib_name| {
+            gpa.free(mem.sliceTo(lib_name, 0));
+        }
+    }
 };
 
 /// The container that structs, enums, unions, and opaques have.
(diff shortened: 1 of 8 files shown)


Luukdegram:
The use case where this is very useful is for wasm extern functions and globals.
Wasm has a section called "imports" where we define objects we want to import from the host environment, such as globals and functions. Those imports require a `module name` and a regular `name`. 

For instance, when we take the following function:
```zig
pub extern "wasi_snapshot_preview1" fn fd_write(...) void;
```
we want to import the function from the host environment. (WASI in this case).
For this particular function, the `module name` is `wasi_snapshot_preview1` and the `name` is `fd_write`.

Before this PR, we had to manually reparse the AST and find the library name that way. However, now we can simply find it from the ZIR.

Note, it is not mandatory to annotate an extern with a library name. Wasm uses `env` as a default.

kubkon:
Another use-case, for non-Wasm targets, is better error reporting for traditional targets such as MachO and ELF, where if the import is *not* found in the specified library by name, we will report an error. Please note, this is still a TODO.

src/Module.zig
```diff
@@ -383,6 +383,12 @@ pub const Decl = struct {
     /// For anonymous decls and also the root Decl for a File, this is 0.
     zir_decl_index: Zir.Inst.Index,
 
+    /// Whether the extern decl has a lib name populated.
+    /// Populated when tag is `extern_fn` or `extern`.
+    has_lib_name: bool,
```
andrewrk, line 388: This does not need to exist because a `0` value for a `Zir.Inst.Index` can represent absence. As this is written we would be paying 4 bytes for 1 bit of data, per Decl, of which there are many.

However there is a bigger issue with this patch; see below...

src/Sema.zig
```diff
@@ -12404,6 +12404,8 @@ fn zirVarExtended(
 
     const lib_name: ?[]const u8 = if (small.has_lib_name) blk: {
         const lib_name = sema.code.nullTerminatedString(sema.code.extra[extra_index]);
+        sema.owner_decl.has_lib_name = true;
+        sema.owner_decl.zir_lib_name_index = @intCast(Zir.Inst.Index, extra_index);
```
andrewrk, line 12408: This is storing a ZIR index into a Decl. A ZIR index changes when source files are updated - even when unrelated changes are made earlier on in the file. However a Decl is a long-lived object that persists beyond multiple updates. So if you want to hang on to a ZIR index then you have to take it into account into the `Module.mapOldZirToNew` mechanism, which tries to figure out how ZIR indexes have moved when a source file is updated.

The much better solution than doing this, would be to not store a new ZIR index into a Decl at all.

If you look further down this file, you can see this:

```zig
    if (lib_name != null) {
        // Look at the sema code for functions which has this logic, it just needs to
        // be extracted and shared by both var and func
        return sema.fail(block, src, "TODO: handle var with lib_name in Sema", .{});
    }
```

I suggest to take this comment's advice and solve that related problem while working on this problem.

And then finally, storing a extern lib name string into the `Module.Var` object in this function looks like a much more robust solution to the problem than messing with ZIR indexes or ZIR string table indexes. Be sure to copy the string data of `lib_name` so that it is not a reference to the ZIR memory, which gets freed with every update to the source file. You can copy it to the Decl's arena, or you can copy it with gpa, but then you will need to introduce a deinit() function for Module.Var objects.

As for functions, they should be handled pretty much the same thing as variables. Storing the lib name, if any, should be done on `Module.Fn` and `Module.Var`, not on `Module.Decl`.

andrewrk (changes requested):

src/Sema.zig
kubkon: Done! Introduced new type called `Module.ExternFn` which contains a maybe-lib-name field that outlives ZIR and is tied to the liveness of `Module`. Similarly, I added a maybe-lib-name to `Module.Var` which the same liveness, and I've refactored out common  code path for analysing the parsed library name in Sema (for extern fns and vars). Have a look and let me know what you think!

PR #10805, opened 2022-02-05
ci: azure: split build-and-test step

hdorio:
Changes to the BuildWindows steps:
Split the step "Build and Test" into 3 steps "Build", "Test" and "Documentation"
Convert everything into Powershell 7 (core).
Avoid the workaround about symlinks (decompressing 2 times the llvm release) for `tar` by replacing it by `7zip`.
Avoid calling a script within a script (use a pipeline step instead).
Avoid installing tools already installed in the windows-2019 Azure VM image (`wget`, `xz`, `tar`, ...).
Only install `s3cmd` for runs which are not triggered by pull request.


marler8997:
Why replace batch with pwsh?

hdorio:
> Why replace batch with pwsh?

It's easier to deal with error outputs.
Do you have any reasons to want .bat scripts?
 
> Also, are the scripts done the way they are so developers can easily run the CI scripts on their own machine

This is one of the many questions I want to ask whoever is responsible for CI.
But by the look of it, I would say it's not meant to be.

> will that still work by moving more things into separate steps?

Not sure what you mean.
I have tested the pwsh scripts just by copy/pasting the steps and replacing some variables.
At least to me, as a user trying to run the script on my machine, it's easier to see what is going on by looking only at the pipeline.yml file.

hdorio:
@marler8997 I closed this pull request in an attempt to cancel a useless job on the CI but it didn't work.
Do you know someone who has the right to cancel it? (and reopen this pull request).

marler8997:
> No need to end a command with || exit /b because $ErrorActionPreference = 'stop' is added to each step.
> Do you have any reasons to want .bat scripts?

Ah exit on error is a nice feature.  I'm asking why the switch because if we ever make a change we need to have a good reason so nobody switches it back to the original, then back to the new one, then back again etc.  I myself avoid powershell because it takes too long to startup.  I recall waiting 5 to 10 seconds for it to start and immediately closed it and wrote it off as a tool that I myself would not be using.  It does have some nice features I've heard though.

> Do you know someone who has the right to cancel it? (and reopen this pull request).

I don' know who is able to re-open pull requests, maybe those with "Collaborator" status, but I'm not one of those people.  I suggest creating a new PR to triage the CI issues and not closing it in the future.  If you push a new change, the current CI run is automatically cancelled.  You may also be able to cancel your own CI jobs on azure if you create an account and login and it's able to tell that the PR being tested is connected to your github account.

hdorio:
> If you push a new change, the current CI run is automatically cancelled.

Yes, but for unknown reasons my push didn't cancel the build this time :disappointed: so I tried to close the pull request.

PR #10806, opened 2022-02-05
stage2: remdiv

m-radomski:
Every type except `float128` works, I suspect that the issue is the same as in #10801.


andrewrk:
Thanks but looks redundant with  #10740 which is about to be merged.

PR #10813, opened 2022-02-06
child_process: collectOutputWindows handle broken_pipe from ReadFile

marler8997:
This was found on a user's machine when calling "git" as a child process from msys.  Instead of getting `BROKEN_PIPE` on `GetOverlappedResult` like we currently expect, it would occur on `ReadFile` which we are currently ignoring, which would then cause collectOutputWindows to hang because the async operation was never started.

With this patch, `zig build` no longer hangs on their machine when collecting output from child processes.  This bug could also be causing intermittent "hangs" in the CI on windows.

### DIFF
--- a/lib/std/child_process.zig
+++ b/lib/std/child_process.zig
@@ -288,10 +288,8 @@ pub const ChildProcess = struct {
                 else => |err| return windows.unexpectedError(err),
             }
         }
-        if (wait_object_count == 0)
-            return;
 
-        while (true) {
+        while (wait_object_count > 0) {
             const status = windows.kernel32.WaitForMultipleObjects(wait_object_count, &wait_objects, 0, windows.INFINITE);
             if (status == windows.WAIT_FAILED) {
                 switch (windows.kernel32.GetLastError()) {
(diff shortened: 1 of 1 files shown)


marler8997:
I've converted back to a draft for now because it looks like the call to `ReadFile` can actually succeed?  This needs to be handled, looking into it.

marler8997:
Ok `ReadFileEx` uses a callback mechanism but I'm not sure that's what we want.  Instead, I confirmed that calling `ReadFile` on an asynchronous file with an asynchronous OVERLAPPED pointer can still complete synchronously (see https://docs.microsoft.com/en-US/troubleshoot/windows/win32/asynchronous-disk-io-synchronous#asynchronous-io-still-appears-to-be-synchronous).  This means that if `ReadFile` returns success then that call was already completed synchronously and we need to issue another call in order to get one that is asynchronous.  I've abstracted this detail into a new function `windowsAsyncRead` which will continue to call `ReadFile` until it triggers an asynchronous read.

andrewrk:
Wow, tricky issue. Thanks for diving into this.

PR #10814, opened 2022-02-06
CLI: remove remainders of --verbose-ast and --verbose-tokenize

kyrias:
These options were removed in 5e63baae8 but some remainders were left in.


andrewrk:
Thanks!

PR #10824, opened 2022-02-06
compiler_rt: add addo

matu3ba:
- approach by Hacker's Delight with wrapping addition
- ca. 1.10x perf over the standard approach on my laptop
- tests with all combinations of min,max with -1,0,+1 and combinations of
  sequences +-1,2,4..,max

### DIFF
--- a/lib/std/special/compiler_rt/addosi4_test.zig
+++ b/lib/std/special/compiler_rt/addosi4_test.zig
@@ -34,7 +34,6 @@ test "addosi4" {
         try test__addosi4(-i, i);
     }
 
-    // zig fmt: off
     // edge cases
     // 0   + 0     = 0
     // MIN + MIN   overflow
@@ -45,7 +44,6 @@ test "addosi4" {
     // MAX + 0     MAX
     // MIN + MAX   -1
     // MAX + MIN   -1
-    // zig fmt: on
     try test__addosi4(0, 0);
     try test__addosi4(min, min);
     try test__addosi4(max, max);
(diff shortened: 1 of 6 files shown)


matu3ba:
benchmark results: https://gist.github.com/matu3ba/f848765e3c67ee74781b57858049d692

see also #10248, which will be follow-up PRs.
If the intention is to not offer `addo` and `subo` to be used with external code for perf reasons, then this PR should be closed.

lib/std/special/compiler_rt/addosi4_test.zig
```diff
@@ -0,0 +1,82 @@
...
+    try test__addosi4(min, min);
+    try test__addosi4(max, max);
+    try test__addosi4(0, min);
+    try test__addosi4(0, max);
+    try test__addosi4(min, 0);
+    try test__addosi4(max, 0);
+    try test__addosi4(min, max);
+    try test__addosi4(max, min);
+
+    // zig fmt: off
```
g-w1, line 59: Why include these zig fmt directives? I think they are not necessary

matu3ba: zig fmt otherwise removes the alignment of the comments.

matu3ba: You are correct and I will remove them. I get confused with the behaviro of clang-fmt.

andrewrk:
To be clear, #10248 has to do with the zig language; I believe this issue has no effect on the corresponding compiler-rt functions.

andrewrk (approved):

Issue #10705, opened 2022-01-27
eliminate anytype fields from the language


andrewrk:
## Motivation

Simplify the language. Most people probably aren't even aware that this is possible:

```zig
const S = struct {
    x: anytype,
};
```

Similarly, the simplified language will be simpler to implement compilers for. Such as, you know, the one that is the main focus of the 0.10.0 milestone.

## Proposed changes

Mainly, make it a compile error to use `anytype` as a field name. However this poses some problems to solve, such as how to deal with `@typeInfo` for various things. For example, the sentinel field of pointers:

https://github.com/ziglang/zig/blob/0.9.0/lib/std/builtin.zig#L239-L252

Likewise the `sentinel` field of arrays, the `default_value` field of structs, and the `function` field of async frames. But that's it. I can't find any other dependencies on this feature in the entire standard library.

In order to address this problem, this proposal suggests to use compile-time type erasure by using `?*const anyopaque` instead of `anytype`. This is something that the language is already required to support, so we are exploiting it here rather than introducing yet another way of doing the same thing via `anytype` struct fields.

```diff
--- a/lib/std/builtin.zig
+++ b/lib/std/builtin.zig
@@ -222,11 +222,10 @@ pub const TypeInfo = union(enum) {
         child: type,
         is_allowzero: bool,
 
-        /// This field is an optional type.
         /// The type of the sentinel is the element type of the pointer, which is
         /// the value of the `child` field in this struct. However there is no way
-        /// to refer to that type here, so we use `anytype`.
-        sentinel: anytype,
+        /// to refer to that type here, so we use pointer to anyopaque.
+        sentinel: ?*const anyopaque,
 
         /// This data structure is used by the Zig language code generation and
         /// therefore must be kept in sync with the compiler implementation.
@@ -244,11 +243,10 @@ pub const TypeInfo = union(enum) {
         len: comptime_int,
         child: type,
 
-        /// This field is an optional type.
         /// The type of the sentinel is the element type of the array, which is
         /// the value of the `child` field in this struct. However there is no way
-        /// to refer to that type here, so we use `anytype`.
-        sentinel: anytype,
+        /// to refer to that type here, so we use a pointer to anyopaque.
+        sentinel: ?*const anyopaque,
     };
 
     /// This data structure is used by the Zig language code generation and
@@ -264,7 +262,7 @@ pub const TypeInfo = union(enum) {
     pub const StructField = struct {
         name: []const u8,
         field_type: type,
-        default_value: anytype,
[...]

PR #10849, opened 2022-02-09
std.json: fix compile error for comptime fields

sharpobject:
This is covered by an existing test which was already failing.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1766,7 +1766,7 @@ fn parseInternal(
                                     }
                                 }
                                 if (field.is_comptime) {
-                                    if (!try parsesTo(field.field_type, field.default_value.?, tokens, child_options)) {
+                                    if (!try parsesTo(field.field_type, @ptrCast(*const field.field_type, field.default_value.?).*, tokens, child_options)) {
                                         return error.UnexpectedValue;
                                     }
                                 } else {
--- a/lib/std/testing.zig
+++ b/lib/std/testing.zig
@@ -466,6 +466,8 @@ test {
 pub fn refAllDecls(comptime T: type) void {
     if (!builtin.is_test) return;
     inline for (comptime std.meta.declarations(T)) |decl| {
+        if (decl.is_pub and @typeInfo(@TypeOf(@field(T, decl.name))) == .Struct)
+            _ = @hasDecl(@field(T, decl.name), "foo");
         _ = decl;
     }
 }
--- a/lib/std/zig/c_translation.zig
+++ b/lib/std/zig/c_translation.zig
@@ -166,7 +166,7 @@ pub fn sizeof(target: anytype) usize {
                 const array_info = @typeInfo(ptr.child).Array;
                 if ((array_info.child == u8 or array_info.child == u16) and
                     array_info.sentinel != null and
-                    array_info.sentinel.? == 0)
+                    @ptrCast(*const array_info.child, array_info.sentinel.?).* == 0)
                 {
                     // length of the string plus one for the null terminator.
                     return (array_info.len + 1) * @sizeOf(array_info.child);
--- a/lib/std/zig/parser_test.zig
+++ b/lib/std/zig/parser_test.zig
@@ -585,15 +585,6 @@ test "zig fmt: asm expression with comptime content" {
     );
 }
 
-test "zig fmt: anytype struct field" {
-    try testCanonical(
-        \\pub const Pointer = struct {
-        \\    sentinel: anytype,
-        \\};
-        \\
-    );
-}
-
 test "zig fmt: array types last token" {
     try testCanonical(
         \\test {
@@ -4146,6 +4137,17 @@ test "zig fmt: container doc comments" {
         \\//! Nothing here
         \\
     );
+}
+
+test "zig fmt: anytype struct field" {
+    try testError(
+        \\pub const Pointer = struct {
+        \\    sentinel: anytype,
+        \\};
+        \\
+    , &[_]Error{
+        .expected_type_expr,
+    });
 }
 
 test "zig fmt: extern without container keyword returns error" {


daurnimator:
I don't understand why this would be needed..... is this working around a compiler bug?

sharpobject:
I don't know what the expected behavior is when you pass a `*const anyopaque` to a function that takes a `T`. If that is expected to work, then it's working around a compiler bug. If not, then it's not.

The stdlib tests for this file are currently failing because of this issue. Is that not what you see if you run them?

sharpobject:
Here's what I get over here on master

```
$ ./build/zig test lib/std/json.zig --zig-lib-dir lib --main-pkg-path lib/std
./lib/std/json.zig:1769:92: error: expected type 'i32', found '*const anyopaque'
                                    if (!try parsesTo(field.field_type, field.default_value.?, tokens, child_options)) {
                                                                                           ^
./lib/std/json.zig:1906:32: note: called from here
    const r = try parseInternal(T, token, tokens, options);
                               ^
./lib/std/json.zig:2106:66: note: called from here
        try testing.expectEqual(T{ .a = 0, .b = true }, try parse(T, &TokenStream.init(
                                                                 ^
./lib/std/json.zig:2100:34: note: called from here
test "parse with comptime field" {
                                 ^
./lib/std/json.zig:1769:92: error: expected type 'f64', found '*const anyopaque'
                                    if (!try parsesTo(field.field_type, field.default_value.?, tokens, child_options)) {
                                                                                           ^
./lib/std/json.zig:1906:32: note: called from here
    const r = try parseInternal(T, token, tokens, options);
                               ^
./lib/std/json.zig:2246:42: note: called from here
    try testing.expectEqual(t3, try parse(T3, &TokenStream.init(str), options_first));
                                         ^
./lib/std/json.zig:2220:47: note: called from here
test "parse into struct with duplicate field" {
                                              ^
./lib/std/json.zig:1769:92: error: expected type '[]const u8', found '*const anyopaque'
                                    if (!try parsesTo(field.field_type, field.default_value.?, tokens, child_options)) {
                                                                                           ^
./lib/std/json.zig:1700:38: note: called from here
                    if (parseInternal(u_field.field_type, token, &tokens_copy, options)) |value| {
                                     ^
./lib/std/json.zig:1906:32: note: called from here
    const r = try parseInternal(T, token, tokens, options);
                               ^
./lib/std/json.zig:2130:28: note: called from here
        const r = try parse(T, &TokenStream.init(
                           ^
./lib/std/json.zig:2100:34: note: called from here
test "parse with comptime field" {
                                 ^
./lib/std/json.zig:1769:92: error: expected type '[]const u8', found '*const anyopaque'
                                    if (!try parsesTo(field.field_type, field.default_value.?, tokens, child_options)) {
                                                                                           ^
./lib/std/json.zig:1700:38: note: called from here
                    if (parseInternal(u_field.field_type, token, &tokens_copy, options)) |value| {
                                     ^
./lib/std/json.zig:1906:32: note: called from here
    const r = try parseInternal(T, token, tokens, options);
                               ^
./lib/std/json.zig:2130:28: note: called from here
        const r = try parse(T, &TokenStream.init(
                           ^
./lib/std/json.zig:2100:34: note: called from here
test "parse with comptime field" {
                                 ^
```

matu3ba:
> working around a compiler bug

Please be descriptive on what semantics your code change has. Before the change the compiler (I do assume stage1) crashed on the comptime field.
To fix this you are suggesting to use the default value.

> This is covered by an existing test which was already failing.

Which test? I dont see any disabled one inside json.zig and json/test.zig

Finally and most important: Can you provide a use case that justify this hack to derive behavior from a opaque pointer meant to only be used on the [external interfacing code](https://interrupt.memfault.com/blog/opaque-pointers#the-opaque-pointer-pattern) ?

daurnimator:
> The stdlib tests for this file are currently failing because of this issue. Is that not what you see if you run them?

tests currently pass for me; and on every CI run for the repository.
Are you on some esoteric platform?
Otherwise you likely have corrupted your zig installation/checkout somehow.

sharpobject:
>Please be descriptive on what semantics your code change has. Before the change the compiler (I do assume stage1) crashed on the comptime field.
>To fix this you are suggesting to use the default value.
Prior to the change in which anytime fields were removed from the language, this line already used the default value. The intent of this line is to check that the value of the field in the json is equal to the value of the comptime field.

Can you be descriptive of the intended semantics of the current behavior of passing a `*const anyopaque` where an `f64` is expected?

If there is some philosophical objection to the idea of accessing the default value (which is where the value of the comptime field is stored) ever for any reason, I guess we will have to give up on the idea of having a generic json parser that can verify that values in json objects correspond to comptime fields. In that case, I can submit a different patch that completely removes the feature.

> Which test? I dont see any disabled one inside json.zig and json/test.zig
It's not disabled. I posted the output and the command used to generate the output in the comment immediately preceding yours.

> Are you on some esoteric platform?
The above is the failure I observe when I run precisely these commands on my M1 mac, starting with no zig2 folder:
```
git clone git@github.com:ziglang/zig.git zig2
cd zig2
mkdir build
cd build
cmake .. -DCMAKE_PREFIX_PATH=$(brew --prefix llvm)
make
cd ..
./build/zig test lib/std/json.zig --zig-lib-dir lib --main-pkg-path lib/std
```
I haven't succeeded in building llvm 13 on linux on x64, but I am also able to reproduce this test failure on linux on x64 by running the test using a zig binary obtained from snap against a fresh clone of the repository. the zig binary obtained from snap reports its version as 0.10.0-dev.662+e139c41fd. That is, after acquiring the zig from snap, I can run these commands on linux on x64 to also produce the output I posted above:
```
git clone git@github.com:ziglang/zig.git zig2
cd zig2
zig test lib/std/json.zig --zig-lib-dir lib --main-pkg-path lib/std
```

I am attempting to follow the instructions given at https://github.com/ziglang/zig/wiki/Building-Zig-From-Source and https://github.com/ziglang/zig/blob/master/CONTRIBUTING.md about building zig (in the case of the M1 mac) and running stdlib tests. Have I done so incorrectly?

daurnimator:
> Have I done so incorrectly?

Ah it seems I tested against non-latest zig. updating to master I indeed see the same failure.




> anytype fields were removed from the language

Ah ha! I now understand where the change comes from.

daurnimator (approved):

daurnimator:
I wonder if this also needs an `@alignCast`?
https://github.com/ziglang/zig/issues/10705#issuecomment-1023995206

PR #10850, opened 2022-02-09
elf: pad out file to the required size when init data

kubkon:
We need to pad out the file to the required maximum size equal the
final section's offset plus the section's size. We only need to
this when populating initial metadata and only when section header
was updated.


andrewrk (commented):
Nice thank you!

PR #10858, opened 2022-02-10
stage2 sema: Fix sign handling of exotic integers in `@bitCast`

topolarity:
**Summary:** This PR fixes a small sign-handling bug in `readTwosComplement` that affects `@bitCast` for non power-of-2 integer types. As an extra bonus, it adds support for big integers of arbitrary length.

Previous implementation would incorrectly result in a positive value for exotic integers, e.g. `@bitCast(i48, @as(u48, 0x800000000000))`. I added some tests to cover more big and exotic integers, since this was missing in the existing coverage.

**Note:** ~~This change includes endian-specific logic that I'm not able to test on my little-endian machine. If anyone can point to a CI log or emulator for me to use, I'd rest a lot easier knowing this was tested on some BE targets.~~ Getting started with QEMU now

### DIFF
--- a/src/value.zig
+++ b/src/value.zig
@@ -1046,7 +1046,7 @@ pub const Value = extern union {
                 var bigint_buffer: BigIntSpace = undefined;
                 const bigint = val.toBigInt(&bigint_buffer);
                 const bits = ty.intInfo(target).bits;
-                const abi_size = ty.abiSize(target);
+                const abi_size = @intCast(usize, ty.abiSize(target));
                 bigint.writeTwosComplement(buffer, bits, abi_size, target.cpu.arch.endian());
             },
             .Enum => {
@@ -1055,7 +1055,7 @@ pub const Value = extern union {
                 var bigint_buffer: BigIntSpace = undefined;
                 const bigint = int_val.toBigInt(&bigint_buffer);
                 const bits = ty.intInfo(target).bits;
-                const abi_size = ty.abiSize(target);
+                const abi_size = @intCast(usize, ty.abiSize(target));
                 bigint.writeTwosComplement(buffer, bits, abi_size, target.cpu.arch.endian());
             },
             .Float => switch (ty.floatBits(target)) {
@@ -1098,7 +1098,7 @@ pub const Value = extern union {
                 const Limb = std.math.big.Limb;
                 const limb_count = (buffer.len + @sizeOf(Limb) - 1) / @sizeOf(Limb);
                 const limbs_buffer = try arena.alloc(Limb, limb_count);
-                const abi_size = ty.abiSize(target);
+                const abi_size = @intCast(usize, ty.abiSize(target));
                 var bigint = BigIntMutable.init(limbs_buffer, 0);
                 bigint.readTwosComplement(buffer, int_info.bits, abi_size, endian, int_info.signedness);
                 return fromBigInt(arena, bigint.toConst());


topolarity:
There's a bit of spec ambiguity to call out here:

It's not immediately clear what the read/written size of the `[]u8` memory buffer should be for non-power-of-two types. The amount of buffer we use matters because it affects sign extension and, on big-endian targets, the significance of each byte read. However, the implementations currently diverge for non-power-of-2 integer types:
 - The stage1 implementation uses a "tight" buffer length of `divCeil(bits, 8)`
 - The existing stage2 implementation will do some over-reading/writing, in order to align with a power-of-two integer size within a limb
 - `mem.readInt` and `mem.readWrite` also use a "tight" buffer length of `divCeil(bits, 8)`
 - The [documentation for @byteSwap](https://ziglang.org/documentation/0.9.0/#byteSwap) mentions that exotic integers like `u24` should be swapped with their ABI size, which both suggests that they have a well-defined layout and disagrees with the implementations above

@andrewrk mentioned in the stage2 meeting that it's not clear whether exotic integers will have a well-defined layout, in which case the above differences are probably not observable by defined behavior. 
 
This PR does not try to resolve any of this for now. All we guarantee is that a round-trip through read/writeTwosComplement will restore the same big integer. Whenever the spec is clarified about whether these integer types have well-defined layouts, we can go back and fix this padding problem.

topolarity:
Oops, looks like big-endian is an issue, after all 😅 

Currently spinning up QEMU to investigate

lib/std/math/big/int_test.zig
```diff
@@ -2486,3 +2486,28 @@ test "big int popcount" {
...
+        // Writing to buffer and back should not change anything
+        a.toConst().writeTwosComplement(buffer1, 493, endian);
+        m.readTwosComplement(buffer1, 493, endian, .unsigned);
+        try testing.expect(m.toConst().order(a.toConst()) == .eq);
+
+        // Equivalent to @bitCast(i493, @as(u493, intMax(u493))
+        a.toConst().writeTwosComplement(buffer1, 493, endian);
+        m.readTwosComplement(buffer1, 493, endian, .signed);
+        try testing.expect(m.toConst().orderAgainstScalar(-1) == .eq);
+    }
```
matu3ba, line 2512: 1. Can you add another test to ensure representation of endianess is correct? 
Something along  `@bswap(little) == big`.

2. Another nice test is to use a combination of bitpatterns like those in [bswap](https://github.com/ziglang/zig/blob/master/lib/std/special/compiler_rt/bswapti2_test.zig), but mixing them to minimize any test is correct by chance.

test/behavior/bitcast.zig
```diff
@@ -11,21 +12,119 @@ test "@bitCast i32 -> u32" {
 }
 
 fn testBitCast_i32_u32() !void {
-    try expect(conv(-1) == maxInt(u32));
-    try expect(conv2(maxInt(u32)) == -1);
+    try expect(conv_i32(-1) == maxInt(u32));
+    try expect(conv_u32(maxInt(u32)) == -1);
+    try expect(conv_u32(0x8000_0000) == minInt(i32));
+    try expect(conv_i32(minInt(i32)) == 0x8000_0000);
```
matu3ba, line 18: no maxInt(i32), minInt(u32), -0 etc?
Is it possible to test error cases?

test/behavior/bitcast.zig
```diff
@@ -11,21 +12,119 @@ test "@bitCast i32 -> u32" {
...
+fn conv_u27(x: u27) i27 {
+    return @bitCast(i27, x);
+}
+
+test "@bitCast i512 -> u512" {
+    try testBitCast_i512_u512();
+    comptime try testBitCast_i512_u512();
+}
+
+fn testBitCast_i512_u512() !void {
```
matu3ba, line 81: `2^{9} = 512`, so 512 is no exotic integer. Is this intended, because there were no tests for regular integers?

test/behavior/bitcast.zig
```diff
@@ -11,21 +12,119 @@ test "@bitCast i32 -> u32" {
...
+    try testBitCast_i493_u493();
+    comptime try testBitCast_i493_u493();
+}
+
+fn testBitCast_i493_u493() !void {
+    if (builtin.zig_backend == .stage2_wasm) return error.SkipZigTest;
+    if (builtin.zig_backend == .stage2_c) return error.SkipZigTest;
+    if (builtin.zig_backend == .stage2_x86_64) return error.SkipZigTest;
+
+    try expect(conv_i493(-1) == maxInt(u493));
```
matu3ba, line 114: create a few values in between. formula is `2s complement(number) = ~number + 1` except for 0 handled by the hardware and min.

also add `-0` as edge case.
you could also use the formula directly except for minInt and 0.

matu3ba (changes requested):
Structure looks good, though tests dont cover some edge cases yet.
Especially the endianess should have more test coverage, ie with reading bit patters and some sort of byteswap.

Somebody, who did big int stuff should look into the implementation.

test/behavior/bitcast.zig
topolarity: This is intended, because there were no tests for integers larger than a Limb (`usize`)

lib/std/math/big/int_test.zig
topolarity: Great suggestion - I added a few explicit tests which compare the written buffer against a known good output, including endianness coverage.

test/behavior/bitcast.zig
matu3ba: 1. out of range stuff for longer aligned integer could be broken.
one option to test this is to write a bit pattern to a buffer, cast a subpart (or whole) region to an exotic integer and test, if the casted integer fails the bitcast.

2. another option is to test, if the bitcast only applies to a subpart of a buffer by writing to it and check expected results with the manual combination of bit masks or their results.

Option 2 with bit mask (fun) feels like it is in scope of this PR, Option 1 feels like it belong into the other PR.

topolarity:
I think the ABI details have been resolved with https://github.com/ziglang/zig/pull/10858/commits/7b72fc6bbc5554643bc27933310899e32783b81b

These assumptions are made w.r.t. integer layout:
- Any required padding bits are assumed to be on the MSB end
- An explicit `abi_size` is provided to read/writeTwosComplement

With this information, the location and significance of any bits in the read/written buffers are well-defined, even for exotic integers. As an implementation detail, I also sign extend when writing padding bits in `writeTwosComplement`.

Note that power-of-two integers maintain the same layout as before (they have no padding bits). The main difference is that exotic integers now read/write to a sign-extended buffer corresponding to their ABI size.

Let me know if there are any bad assumptions in the above :+1:

test/behavior/bitcast.zig
matu3ba: > write undefined values to the output buffer/integer, if we're concerned about incomplete updates

Uh, I think I got confused with `@truncate` and was thinking `@bitCast` could do the same.
Is there an operation to get a subslice at bit-boundary, which `@bitCast` can operate on or are all integers
aligned at byte boundary (via `@truncate`)?

(probably not relevant)
Then it would look like 1. Use `const buf: []const u8 : .{ bitpattern};`, 2. take a subslice, 3. convert to the exotic integer, 4. make an arithmetic operation (negation, abs, shift left/right, addition, subraction, multiplication, division) and 5. compare to the expected bitpattern of the whole slice `buf`.

Mark this as resolved, if you think this does not apply.

topolarity: Ah, no problem.

I don't think this applies - read/writeTwosComplement have no functionality to read/write a subset of the bits backing a big integer

topolarity:
Any next steps for me here? Are we waiting on an additional reviewer?

andrewrk:
> Any next steps for me here? Are we waiting on an additional reviewer?

Yeah just waiting on a review from a core team member. I'm looking at it now - sorry for the delay!

topolarity:
> Really nice work @topolarity. Thank you for dealing with the spec ambiguities - I know that's a particularly tricky situation when contributing.
> 
> After we're finally shipping stage2, I think one of the highest priority items for me will be getting a first draft of the lang spec out the door, which we can then iterate on.

That's great to hear, and no worries about the spec - sometimes ambiguity is fun

Still, that's one more reason to push stage2 to feature-complete 🚀

Issue #10773, opened 2022-02-03
Incorrect number logged when a test fails 


m0vsb:
### Zig Version

0.10.0-dev.450+63ee6e662

### Steps to Reproduce

1. Create a test file with the following contents.

```zig
const std = @import("std");

pub fn add(x: i32, y: i32) i32 {
    return x + y;
}

test "basic test" {
    try std.testing.expectEqual(10, 3 + 7);
}

test "Test should fail" {
    try std.testing.expectEqual(add(5, 5), 15);
}
```
2. Run the test file `zig test test.zig`

### Expected Behavior

Test runner should show the correct test number when the test fails.
Example:
```
Test [2/2] test "Test should fail"... expected 10, found 15
Test [2/2] test "Test should fail"... FAIL (TestExpectedEqual)
```

### Actual Behavior

```
Test [1/2] test "Test should fail"... expected 10, found 15
Test [2/2] test "Test should fail"... FAIL (TestExpectedEqual)
/usr/lib/zig/std/testing.zig:79:17: 0x20801f in std.testing.expectEqual (test)
                return error.TestExpectedEqual;
                ^
/home/kenobi/code/sterling/src/test.zig:8:5: 0x207a8f in test "Test should fail" (test)
    try std.testing.expectEqual(5 + 5, 15);
    ^
1 passed; 0 skipped; 1 failed.
error: the following test command failed with exit code 1:
src/zig-cache/o/193861cd31ad5908c0ceb539270824aa/test /usr/bin/zig
```

SebastianKeller:
@m0vsb With #10859 being merged, the output is as expected

<details>
<summary>output</summary>

```zig
const std = @import("std");

pub fn add(x: i32, y: i32) i32 {
    return x + y;
}

test "basic test" {
    try std.testing.expectEqual(10, 3 + 7);
}

test "Test should fail" {
    try std.testing.expectEqual(add(5, 5), 15);
}
```
```
[I] ➜ zig test test.zig 
Test [2/2] test "Test should fail"... expected 10, found 15
Test [2/2] test "Test should fail"... test "Test should fail"... FAIL (TestExpectedEqual)
/home/basti/local/lib/zig/std/testing.zig:79:17: 0x20805d in std.testing.expectEqual (test)
                return error.TestExpectedEqual;
                ^
/home/basti/temp/test.zig:12:5: 0x207a69 in test "Test should fail" (test)
    try std.testing.expectEqual(add(5, 5), 15);
    ^
[...]

PR #10859, opened 2022-02-10
Fixed progress indicator for `zig test`

SebastianKeller:
This is a fix for https://github.com/ziglang/zig/issues/10773
Now the progress always starts with 1. The counter is updated last, after every test related message has been printed.

### DIFF
--- a/lib/std/special/test_runner.zig
+++ b/lib/std/special/test_runner.zig
@@ -82,18 +82,18 @@ pub fn main() void {
         } else |err| switch (err) {
             error.SkipZigTest => {
                 skip_count += 1;
-                test_node.end();
                 progress.log("{s}... SKIP\n", .{test_fn.name});
                 if (!have_tty) std.debug.print("SKIP\n", .{});
+                test_node.end();
             },
             else => {
                 fail_count += 1;
-                test_node.end();
                 progress.log("{s}... FAIL ({s})\n", .{ test_fn.name, @errorName(err) });
                 if (!have_tty) std.debug.print("FAIL ({s})\n", .{@errorName(err)});
                 if (@errorReturnTrace()) |trace| {
                     std.debug.dumpStackTrace(trace.*);
                 }
+                test_node.end();
             },
         }
     }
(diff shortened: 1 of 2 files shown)


lib/std/special/test_runner.zig
```diff
@@ -35,6 +35,8 @@ pub fn main() void {
         .dont_print_on_dumb = true,
     };
     const root_node = progress.start("Test", test_fn_list.len);
+    root_node.completeOne();
```
andrewrk, line 38: incorrect API usage; it didn't actually complete one task, did it?

andrewrk (changes requested):

lib/std/special/test_runner.zig
SebastianKeller: It did not. But the test runner is executing the first test with `[0/x]`, where x is the number of tests. I guess the first test should be displayed as `[1/x]`.

Is there another way to accomplish that?

SebastianKeller: Would you prefer a `start_with_index_one` flag in the progress struct? This would allow to choose between 'start with 0' or 'start with 1'.

andrewrk: I don't think such thing should be configurable. I'm OK with changing the display to be more intuitive however.

SebastianKeller: Ok, cool. I'll change that.

Snektron:
This change introduces an inconsistency: When the last item being worked on is printed, the count will be one more than the total count, and so the line will briefly display something like `Progress [66/65]`. Normally this is easily missed, however, it appears in a scenario where the time between updates is relatively long, and so when the final `completeOne()` is called it determines an update is required, and prints the incorrect message.

I would propose to either revert the change on this line:
https://github.com/ziglang/zig/blob/33cf6ef621114daad63d14067b6ff374e664d410/lib/std/Progress.zig#L267
as the problem in the issue seems to be fixed by the other parts of this change, or to replace the final value with some special indicator, like `Progress [done]`.

Issue #10114, opened 2021-11-07
Stack iterator may segfault on non-standard stack frames


silversquirl:
### Remember to search before filing a new report

- [X] I searched for this bug and did not find it in the issue tracker, and furthermore, the title I used above will make this new bug report turn up in the search results for my query.

### Zig Version

0.9.0-dev.1561+5ebdc8c46

### Steps to Reproduce

Run this test:

```zig
const std = @import("std");

test "StackIterator segfault" {
    weirdStackFrame();
}

comptime {
    asm (
        \\weirdStackFrame:
        // Random padding for 16-byte alignment
        \\  pushq $0x123abc
        // Save rbp
        \\  pushq %rbp
        // Push an invalid address that's above the stack
        \\  movq $0xffffaaaa0000, %rax
        \\  pushq %rax
        // Set rbp so it points to our invalid address
        \\  mov %rsp, %rbp
        // Iterate over the stack
        \\  call doStackIter
        // Clean up
        \\  add $8, %rsp
        \\  popq %rbp
        \\  add $8, %rsp
        \\  ret
    );
}
extern fn weirdStackFrame() void;

export fn doStackIter() void {
    var it = std.debug.StackIterator.init(null, null);
    while (it.next()) |addr| {
        std.math.doNotOptimizeAway(addr);
    }
}
```

It's also possible to reproduce this by having certain system libraries call into a `std.heap.GeneralPurposeAllocator` in debug mode. Since GPA uses `StackIterator` to capture stack frames, libraries that accept allocation callbacks and have weird internal stack layouts can trigger this bug.
One example of a library that exhibits these properties is `libvulkan_radeon` on Linux.

### Expected Behavior

The test should pass

### Actual Behavior

A segmentation fault occurs inside `std.debug.StackIterator.next_internal`:
```
Test [1/1] test "StackIterator segfault"... Segmentation fault at address 0x0
/home/silver/.opt/zig/lib/std/debug.zig:430:67: 0x2091ac in std.debug.StackIterator.next_internal (test)
        const new_fp = math.add(usize, @intToPtr(*const usize, fp).*, fp_bias) catch return null;
                                                                  ^
/home/silver/.opt/zig/lib/std/debug.zig:407:41: 0x207fa0 in std.debug.StackIterator.next (test)
        var address = self.next_internal() orelse return null;
                                        ^
/home/silver/code/play-zig/test.zig:32:19: 0x207ef1 in doStackIter (test)
    while (it.next()) |addr| {
[...]

PR #10863, opened 2022-02-11
std: validate frame-pointer address in stack walking

m-radomski:
Before we are going to de-reference the frame-pointer check if the memory is mapped. Inspired by `libunwind`. For now this is just a draft, main thing that is not finished is the same behavior on non-POSIX systems (Windows). I'm open to input :+]

Closes: #10114.

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -424,6 +424,28 @@ pub const StackIterator = struct {
         return address;
     }
 
+    fn isValidMemory(address: u64) bool {
+        if (native_os != .windows) {
+            var res = true;
+            const length = 2 * mem.page_size;
+            const aligned_address = address & ~@intCast(u64, (mem.page_size - 1));
+            const aligned_memory = @intToPtr([*]align(mem.page_size) u8, aligned_address)[0..length];
+
+            os.msync(aligned_memory, os.MSF.ASYNC) catch |err| {
+                switch (err) {
+                    os.MSyncError.UnmappedMemory => {
+                        res = false;
+                    },
+                    else => unreachable,
+                }
+            };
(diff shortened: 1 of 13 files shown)


Vexu (approved):

andrewrk:
Nice improvement!

lib/std/debug.zig
```diff
@@ -424,6 +424,52 @@ pub const StackIterator = struct {
         return address;
     }
 
+    fn isValidMemory(address: usize) bool {
+        const aligned_address = address & ~@intCast(usize, (mem.page_size - 1));
+
+        // If the address does not span 2 pages, query only the first one
+        const length: usize = if (aligned_address == address) mem.page_size else 2 * mem.page_size;
```
kubkon, line 431: @m-radomski I'm trying to understand post factum, what is the argument here for checking if not only the current page but also the subsequent one are currently mapped by the OS? Put it another way, `msync` will fail should a single page out of the two we are syncing were not mapped. Are we guaranteed this will never happen? On arm64 macOS for instance the page size is a tremendous 16KB meaning we might not map two subsequent pages and as a result not printing the stack trace on encountering a trap since `msync` and by extension `isValidMemory` return error/false. Additionally, what is the point of syncing more than one page given that we will issue a call to `msync` on every iteration of the stack trace walker?

m-radomski: Hi, sorry for the late response, but the logic was ported from libunwind and after looking closely I think you are right. It ought to check just the single page. I think my go-to-definition got lost and transported me to a file for loongarch64 instead of x86. For anyone that might still be interested, search for "validate_mem()" function in libunwind.

PR #10866, opened 2022-02-11
stage1+target: Bringup SerenityOS

sin-ack:
Well hello friends! This is a Zig target for SerenityOS which allows you to build C, C++ and Zig binaries for SerenityOS. To build a binary for the Serenity target, you will need a program (WIP) which can generate the current constants from the SerenityOS repository. This process will be done automatically when using the package script when building the zig port on the Serenity repository.

This PR also contains a few auxiliary fixes which fix some compiler bugs and fix a segfault on i386 targets.

### DIFF
--- a/lib/std/Thread.zig
+++ b/lib/std/Thread.zig
@@ -584,7 +584,7 @@ const PosixThreadImpl = struct {
                 };
                 return @intCast(usize, count);
             },
-            .solaris => {
+            .solaris, .serenity => {
                 // The "proper" way to get the cpu count would be to query
                 // /dev/kstat via ioctls, and traverse a linked list for each
                 // cpu.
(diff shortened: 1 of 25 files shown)


sin-ack:
Dropped the f80 commit; this PR depends on #10861 now.

lib/std/Thread.zig
```diff
@@ -584,7 +584,7 @@ const PosixThreadImpl = struct {
                 };
                 return @intCast(usize, count);
             },
-            .solaris => {
+            .solaris, .serenity => {
```
The-King-of-Toasters, line 587: FWIW: I did the Solaris port, and this comment refers to a unique device. I think modifying the comment to add "On Solaris ...", or creating a serenity-specific branch in the case would be better.

sin-ack:
I think I'll close it for the time being since I won't be attempting this until stage2 is ready to ship (because it will cause a lot of changes on the SerenityOS side too).

PR #10868, opened 2022-02-11
stage2 ARM: lower const slices

joachimschmidt557:
Follow-up to e1a535360fb9ed08fc48018571b9702ab12a5876 (#10862) for ARM

This also fixes some stack offset calculation bugs

cc @kubkon


kubkon:
Is the CI failure related by any chance? It didn't happen before 🤔

joachimschmidt557:
> Is the CI failure related by any chance? It didn't happen before :thinking:

I currently don't see how it could be related, the error is in a std test for `Mutex`.

joachimschmidt557:
> Hmm, it passes now which makes me think we might have some hidden race condition somewhere.

Should I open an issue for this? I still have the link to the "original" CI failure here: https://ci.ziglang.org/ziglang/zig/2278/1/4

PR #10871, opened 2022-02-12
stage2: add type checking for @bitCast

schmee:
I wasn't too sure about how to split up the check between `sema.zirBitcast` and `sema.bitCast`, so it's very possible that things need to move around a bit.

I added more detailed error messages for a couple of cases cause I think it will be very common for people to mix up `@bitCast` and `@ptrCast`. As far as I know the compiler error tests are not implemented yet for stage2 so sadly there's no tests in this PR. I'll happily add them once these tests are enabled for stage2.

One improvement I would like to make but couldn't figure out is to elide the entire call to `@bitCast` if the thing you're bitcasting already has the correct type. So in other words,

```zig
if (operand_ty == dest_ty) {
  // What do I return here???
}
```

If anyone has any hints of what to do there please let me know!

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -6765,8 +6765,71 @@ fn zirBitcast(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air
     const dest_ty_src: LazySrcLoc = .{ .node_offset_builtin_call_arg0 = inst_data.src_node };
     const operand_src: LazySrcLoc = .{ .node_offset_builtin_call_arg1 = inst_data.src_node };
     const extra = sema.code.extraData(Zir.Inst.Bin, inst_data.payload_index).data;
+    const target = sema.mod.getTarget();
 
     const dest_ty = try sema.resolveType(block, dest_ty_src, extra.lhs);
+    switch (dest_ty.zigTypeTag()) {
+        .Type,
+        .Void,
+        .NoReturn,
+        .ComptimeFloat,
+        .ComptimeInt,
+        .Undefined,
+        .Null,
+        .Optional,
+        .ErrorUnion,
+        .ErrorSet,
+        .Opaque,
+        .Frame,
+        .AnyFrame,
+        .EnumLiteral,
+        .Union,
+        .Fn,
+        => return sema.fail(block, dest_ty_src, "invalid type '{}' for @bitCast", .{dest_ty.fmt(target)}),
+
+        .Pointer => {
+            const msg = msg: {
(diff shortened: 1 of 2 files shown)


matu3ba:
out of range stuff for longer aligned integer could be broken.
one option to test this is to write a bit pattern to a buffer, cast a subpart (or whole) region to an exotic integer and test, if the casted integer fails the bitcast.

Edge cases would be here `u/i0` and `u/i65535`, although maybe to not waste CI time a smaller integer should be used.

src/Sema.zig
```diff
@@ -6294,6 +6294,71 @@ fn zirBitcast(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air
...
+                try sema.errNote(block, dest_ty_src, msg, "to cast to a pointer type, use @ptrCast({}, ...)", .{dest_ty});
+                try sema.errNote(block, dest_ty_src, msg, "to cast to a non-pointer type, use @bitCast({}, ...)", .{pointee_ty});
+                break :msg msg;
+            };
+            return sema.failWithOwnedErrorMsg(msg);
+        },
+
+        .Fn,
+        .BoundFn,
+        => return sema.fail(block, dest_ty_src, "cannot @bitCast to '{}', use @ptrCast to cast to a function", .{dest_ty}),
```
Vexu, line 6330: Stage2 handles function pointers differently from stage1; `.Fn` should be with the other invalid types and `.BoundFn` should be in a prong with `@panic("TODO remove this type from the language and compiler")`.

src/Sema.zig
```diff
@@ -6294,6 +6294,71 @@ fn zirBitcast(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air
...
+                        msg,
+                        "consider using 'packed struct' or 'extern struct' for a specified layout.",
+                        .{},
+                    );
+                    break :msg msg;
+                };
+                return sema.failWithOwnedErrorMsg(msg);
+            }
+        },
+        else => {},
```
Vexu, line 6355: For readability it would be nice to explicitly list the six other valid types here.

Vexu (changes requested):

Vexu (approved):

matu3ba:
Do you believe exotic integer checks should be part of this PR or not? https://github.com/ziglang/zig/pull/10858#discussion_r805402021 pointed me to this PR, but they could also be checked as part of another PR.

schmee:
@matu3ba I think I'll stick to the basics in this PR and we can add further improvements in future PRs 👍

PR #10882, opened 2022-02-13
stage2: refactor LLVM bitcount ops

schmee:
- Use `llvm.getIntrinsic` instead of `llvm.getNamedFunction`
- Merge `airClzCtz` and `airPopCount` to reduce duplication and to match SEMA.

### DIFF
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -2203,9 +2203,9 @@ pub const FuncGen = struct {
                 .memcpy         => try self.airMemcpy(inst),
                 .set_union_tag  => try self.airSetUnionTag(inst),
                 .get_union_tag  => try self.airGetUnionTag(inst),
-                .clz            => try self.airBitCount(inst, "llvm.ctlz", true),
-                .ctz            => try self.airBitCount(inst, "llvm.cttz", true),
-                .popcount       => try self.airBitCount(inst, "llvm.ctpop", false),
+                .clz            => try self.airClzCtz(inst, "llvm.ctlz"),
+                .ctz            => try self.airClzCtz(inst, "llvm.cttz"),
+                .popcount       => try self.airPopCount(inst),
                 .tag_name       => try self.airTagName(inst),
                 .error_name     => try self.airErrorName(inst),
                 .splat          => try self.airSplat(inst),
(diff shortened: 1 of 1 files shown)


src/codegen/llvm.zig
```diff
@@ -4320,79 +4320,40 @@ pub const FuncGen = struct {
...
+            params.append(llvm_i1.constNull()) catch unreachable;
+        }
 
-        const params = [_]*const llvm.Value{ operand, llvm_i1.constNull() };
-        const wrong_size_result = self.builder.buildCall(fn_val, &params, params.len, .C, .Auto, "");
+        const operand_llvm_ty = try self.dg.llvmType(operand_ty);
+        const fn_val = self.getIntrinsic(llvm_fn_name, &.{operand_llvm_ty});
+        const wrong_size_result = self.builder.buildCall(
+            fn_val,
+            params.constSlice().ptr,
```
schmee, line 4347: Without the `.ptr` here, compilation fails with:
```
./src/codegen/llvm.zig:4347:30: error: expected type '[*]const *const codegen.llvm.bindings.Value', found '[]const *const codegen.llvm.bindings.Value'
            params.constSlice(),
```
Is this supposed to be an error or is it supposed to coerce? 🤔

src/codegen/llvm.zig
```diff
@@ -4320,79 +4320,40 @@ pub const FuncGen = struct {
...
-        const llvm_i1 = self.context.intType(1);
-        const fn_val = self.dg.object.llvm_module.getNamedFunction(llvm_fn_name) orelse blk: {
-            const operand_llvm_ty = try self.dg.llvmType(operand_ty);
-            const param_types = [_]*const llvm.Type{ operand_llvm_ty, llvm_i1 };
-            const fn_type = llvm.functionType(operand_llvm_ty, &param_types, param_types.len, .False);
-            break :blk self.dg.object.llvm_module.addFunction(llvm_fn_name, fn_type);
-        };
+        var params = std.BoundedArray(*const llvm.Value, 2).init(0) catch unreachable;
+        params.append(operand) catch unreachable;
+        if (zero_poison_flag) {
```
matu3ba, line 4338: could be nice to explain briefly the zero_poison_flag, which applies on some targets for `input=0` for `clz` and `ctz` due to how the hardware works ie `//assert: input != 0 for some targets`.
Or do you think that this is self-evident from reading `compiler_rt`?

andrewrk: Documentation for this is found in the doc comments for the AIR instruction of ctz and clz, as well as the LLVM language reference.

src/codegen/llvm.zig
```diff
@@ -4320,79 +4320,40 @@ pub const FuncGen = struct {
...
-                prefix, bits,
-            }) catch unreachable;
-        const llvm_i1 = self.context.intType(1);
-        const fn_val = self.dg.object.llvm_module.getNamedFunction(llvm_fn_name) orelse blk: {
-            const operand_llvm_ty = try self.dg.llvmType(operand_ty);
-            const param_types = [_]*const llvm.Type{ operand_llvm_ty, llvm_i1 };
-            const fn_type = llvm.functionType(operand_llvm_ty, &param_types, param_types.len, .False);
-            break :blk self.dg.object.llvm_module.addFunction(llvm_fn_name, fn_type);
-        };
+        var params = std.BoundedArray(*const llvm.Value, 2).init(0) catch unreachable;
```
andrewrk, line 4336: You've reduced duplication, but increased branching. Also we are now using a more complex API along with `catch unreachable`. Is this an improvement? It certainly does not match my preference.

schmee: I'm fine either way, so if it's your preference I can split them out to two separate functions again. The `getIntrinsic` part still applies to both 👍

andrewrk:
Really nice improvement :+1:

PR #10883, opened 2022-02-13
stage2 regalloc: track Inst instead of ?Inst in register mapping

joachimschmidt557:
The information whether a register is allocated to an instruction is
already encoded in the `free_registers` "bitmap". Duplicating that
information in the `registers` map is unnecessary and may lead to
performance degradations.


kubkon:
Apologies for closing, hit the wrong button!

PR #10884, opened 2022-02-13
enable Gpu address spaces

gwenzek:
Follow up on #10189 , this enable GPU specific address spaces leveraging previous work from @Snektron in #9649.

Right now this is directly mapping to the LLVM address spaces, but with Snektron we didn't found good reason to deviate from this and we believe it should be compatible across different architectures.

Note: this is still experimental feature because Zig doesn't have any officially supported GPU target.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -18006,10 +18006,14 @@ pub fn analyzeAddrspace(
     const address_space = addrspace_tv.val.toEnum(std.builtin.AddressSpace);
     const target = sema.mod.getTarget();
     const arch = target.cpu.arch;
+    const is_gpu = arch == .nvptx or arch == .nvptx64;
 
     const supported = switch (address_space) {
         .generic => true,
         .gs, .fs, .ss => (arch == .i386 or arch == .x86_64) and ctx == .pointer,
+        // TODO: check that .shared and .local are left uninitialized
+        .global, .param, .shared, .local => is_gpu,
+        .constant => is_gpu and (ctx == .constant),
     };
 
     if (!supported) {
@@ -18020,7 +18024,6 @@ pub fn analyzeAddrspace(
             .constant => "constant values",
             .pointer => "pointers",
         };
-
         return sema.fail(
             block,
             src,
--- a/lib/std/builtin.zig
+++ b/lib/std/builtin.zig
@@ -157,6 +157,12 @@ pub const AddressSpace = enum {
     gs,
     fs,
     ss,
+    // GPU address spaces
+    global,
+    constant,
+    param,
+    shared,
+    local,
 };
 
 /// This data structure is used by the Zig language code generation and
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -801,6 +801,16 @@ pub const DeclGen = struct {
                 .gs => llvm.address_space.x86.gs,
                 .fs => llvm.address_space.x86.fs,
                 .ss => llvm.address_space.x86.ss,
+                else => unreachable,
+            },
+            .nvptx, .nvptx64 => switch (address_space) {
+                .generic => llvm.address_space.default,
+                .global => llvm.address_space.nvptx.global,
+                .constant => llvm.address_space.nvptx.constant,
+                .param => llvm.address_space.nvptx.param,
+                .shared => llvm.address_space.nvptx.shared,
+                .local => llvm.address_space.nvptx.local,
+                else => unreachable,
             },
             else => switch (address_space) {
                 .generic => llvm.address_space.default,
--- a/test/cases.zig
+++ b/test/cases.zig
@@ -16,4 +16,5 @@ pub fn addCases(ctx: *TestContext) !void {
     try @import("stage2/riscv64.zig").addCases(ctx);
     try @import("stage2/plan9.zig").addCases(ctx);
     try @import("stage2/x86_64.zig").addCases(ctx);
+    try @import("stage2/nvptx.zig").addCases(ctx);
 }
--- a/test/stage2/nvptx.zig
+++ b/test/stage2/nvptx.zig
@@ -1,0 +1,57 @@
+const std = @import("std");
+const TestContext = @import("../../src/test.zig").TestContext;
+
+const nvptx = std.zig.CrossTarget{
+    .cpu_arch = .nvptx64,
+    .os_tag = .cuda,
+};
+
+pub fn addCases(ctx: *TestContext) !void {
+    {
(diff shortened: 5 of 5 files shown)


gwenzek:
Tagging @Vexu who reviewed the previous PR (I'm not able to assign reviewers, not sure if it's intended or not)

Snektron:
One thing you might also want to implement is the logic for obtaining the default address space in a given context for this architecture:
https://github.com/ziglang/zig/blob/f5068107cdf54f143a150feb5d500962ba98e963/src/target.zig#L598-L614
Currently this just returns the generic address space, but since this is not allowed for SPIR-V, the idea was to assign a default address space for each context. This would for example set `const x = y;` at global scope to `constant`, and function-local variables to `local`. Do you happen to know if theres a big difference between the performance of different address spaces? From the [LLVM NVPTX Docs](https://llvm.org/docs/NVPTXUsage.html#address-spaces), it looks like the the default address space is generic, and I assume this is what CUDA also uses, but I'm not very sure about that. If CUDA instead makes function-locals `local` (address space 5) instead, it might be best to replicate that in Zig.

gwenzek:
> Do you happen to know if there's a big difference between the performance of different address spaces?

So when you dereference a generic pointer, the GPU runtime need to first identify what kind of pointer it is before dereferencing.
I don't know how costly it is, but it's certainly better to use qualified pointers as much as possible.

The slowest memory are local and global, then constant memory is a bit faster, then shared memory is significantly faster and finally registers is the fastest memory.

* the shared memory is on chip memory, and compete with memory of the L1 cache (IIRC, need to find a source), it's generally act as a user controlled cache.
* `constant` address space can be the same than `global` on some Nvidia GPUs, but is otherwise faster because it has dedicated caching
* local memory is generally only used for kernels that require too many registers.

> you might also want to implement is the logic for obtaining the default address space

I've tried based on your recommendation, but I think it will require more work. It tend to generate compile errors when 
generating the builtin.zig for nvptx64

`error: expected *const target.Target.Cpu.Model, found *addrspace(.constant) const target.Target.Cpu.Model`

I've added the commit for reference, but will probably rollback since it's seems mostly useless for now:

https://github.com/ziglang/zig/blob/aa7acf057c832ff8d897d746722a913bcaae2b3c/src/target.zig#L604-L614

What would be useful would be to mark kernel parameters that are pointers as pointers to global memory, but I haven't found a way to do so.

I'm also not sure if we should expose the `.local` address space. For Cuda at least, it's not something we should use. It's only use when the device doesn't have enough registers to compute a given kernel, but it's quite slow.

Snektron:
> > you might also want to implement is the logic for obtaining the default address space
> 
> I've tried based on your recommendation, but I think it will require more work. It tend to generate compile errors when generating the builtin.zig for nvptx64
> 
> `error: expected *const target.Target.Cpu.Model, found *addrspace(.constant) const target.Target.Cpu.Model`
> 
> I've added the commit for reference, but will probably rollback since it's seems mostly useless for now:

Hm, maybe need to experiment a bit more with that. Let's address that another time.

> What would be useful would be to mark kernel parameters that are pointers as pointers to global memory, but I haven't found a way to do so.

Do you mean automatically? Perhaps it could be inferred based on the function calling convention, I am having second thoughts about the whole "automatically inferring address spaces" thingy anyway. Perhaps things should remain more explicit?

In the explicit (and current) case it would look like
```zig
export fn kernel(ptr: *addrspace(.global) u32) callconv(.PtxKernel) void {
  ...
}
```

> I'm also not sure if we should expose the `.local` address space. For Cuda at least, it's not something we should use. It's only use when the device doesn't have enough registers to compute a given kernel, but it's quite slow.

Hm i see. The AMDGPU backend also exposes a `local` address space, and it seems that any `alloca` instruction automatically gains this address space (so in that sense its the correct mental model for local variables). How does this work for PTX? Do we need to avoid `alloca` all together?

src/Sema.zig
```diff
@@ -17578,10 +17578,13 @@ pub fn analyzeAddrspace(
     const address_space = addrspace_tv.val.toEnum(std.builtin.AddressSpace);
     const target = sema.mod.getTarget();
     const arch = target.cpu.arch;
+    const is_gpu = arch == .nvptx or arch == .nvptx64;
 
     const supported = switch (address_space) {
         .generic => true,
         .gs, .fs, .ss => (arch == .i386 or arch == .x86_64) and ctx == .pointer,
+        .global, .shared, .local => is_gpu,
+        .constant => is_gpu and (ctx == .constant),
```
Snektron, line 17587: I think that `constant` should also be valid for pointers. The context `.constant` refers to a global value that is constant in this case. Im also not sure if `global` should be valid for anything thats not a pointer. It should in theory be possible to declare a `local` and `shared` global value, but im not sure whether it can be initialized at all.

Snektron (changes requested):

gwenzek:
>  The AMDGPU backend also exposes a local address space, and it seems that any alloca instruction automatically gains this address space. How does this work for PTX? 

PTX does also have `alloca` but it's considered an experimental feature: https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#stack-manipulation-instructions-alloca
The returned pointer is effectively in .local space.
The code that I saw generated is always fully inlined, and there is no function calls nor alloca I don't know if it's because my kernels are too simples, or is that's just a common practice for PTX format.

Vexu (approved):

src/Sema.zig
```diff
@@ -17592,7 +17595,7 @@ pub fn analyzeAddrspace(
             .constant => "constant values",
             .pointer => "pointers",
         };
-
+        log.err("{s} with address space '{s}' are not supported on {s}", .{ entity, @tagName(address_space), arch.genericName() });
```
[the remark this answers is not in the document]
gwenzek: removed

src/Sema.zig
gwenzek: followed up in the main thread

gwenzek:
>  Im also not sure if global should be valid for anything thats not a pointer. It should in theory be possible to declare a local and shared global value, but im not sure whether it can be initialized at all.

So PTX can initialize .global and .constant. It seems to also be possible with  AMDGPU, but it's a bit harder.
Maybe it won't be possible on all platforms but let's be permissive for now.
.local and .shared can't be preinitialized, but I haven't found a way to prevent this in the Zig frontend, so the errors are triggered in the backend.

relevant ptx documentation
https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#state-spaces__properties-state-spaces

gwenzek:
Thanks for the insightful exchanges Robin, 
and thanks for the reviews !

Time to go back making more examples of GPU Zig programming :-)

PR #10893, opened 2022-02-14
stage2 AArch64: get zig test working; enable behavior tests

joachimschmidt557:
Behavior testing for `aarch64-linux` is now enabled, further work for `aarch64-macos` is required though.

cc @kubkon


kubkon (approved):
Fantastic work!

Issue #9743, opened 2021-09-12
std.lib: inconsistency between std.json and compiler when creating strings.


jdmichaud:
`std.json.parser` creates slices when encountering a string:
```zig
const std = @import("std");

const Foo = struct {
  bar: []u8,
};

pub fn main() !void {
  const options = std.json.ParseOptions{ .allocator = std.testing.allocator };
  const foo = try std.json.parse(Foo, &std.json.TokenStream.init("{\"bar\":\"bar\"}"), options);
  std.debug.print("{}\n", .{ foo.bar[3] }); // Index out of bound
}
```
The slice is *not* null-terminated. It is counter-intuitive as a string, when created through a literal, is null-terminated. There is a lack of consistency between the compiler and the standard library.

```zig
const std = @import("std");

pub fn main() !void {
  const bar = "bar";
  std.debug.print("{}\n", .{ bar[3] }); // This would print 0
}
```

Trying to change the receiving type:
```zig
const Foo = struct {
  bar: [:0]u8,
};
```
Zig fails with:
```
/opt/compiler-explorer/zig-master/lib/std/json.zig:1845:58: error: expected type '[:0]u8', found '[]u8'
                            return arraylist.toOwnedSlice();
```

PR #10895, opened 2022-02-15
Make json parser return a null terminated string for []u8 types

bsansouci:
Hey all, this is my first time contributing to Zig so adjust your expectations accordingly heh.
I wanted to contribute and picked this issue [#9743](https://github.com/ziglang/zig/issues/9743) to fix.
My fix only partially fixes the issue, because there's the 2nd part where if you mark the struct field as a null terminating slice `u8[:0]` you get yelled at that the types don't match with what the function returns:
`./zig/build/lib/zig/std/json.zig:1874:58: error: expected type '[:0]u8', found '[]u8'` <- something like that 

I don't know enough about the language to tell exactly what the expectation is, and how to handle a polymorphic case like this. But I can fix that part in another PR.

I'd love some feedback on how to improve this!

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -2248,13 +2248,13 @@ test "parse into struct with strings and arrays with sentinels" {
     ), options);
     defer parseFree(T, r, options);
 
-    try testing.expectEqualSlicesSentinel("zig", r.language);
+    try testing.expectEqualSentinel("zig", r.language);
 
     const data = [_:99]i32{ 1, 2, 3 };
     const simple_data = [_]i32{ 4, 5, 6 };
 
-    try testing.expectEqualSlicesSentinel(data[0..data.len], r.data);
-    try testing.expectEqualSlicesSentinel(simple_data, r.simple_data);
+    try testing.expectEqualSentinel(data[0..data.len], r.data);
+    try testing.expectEqualSentinel(simple_data, r.simple_data);
 }
 
 test "parse into struct with duplicate field" {
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1876,15 +1876,15 @@ fn parseInternal(
                         .String => |stringToken| {
                             if (ptrInfo.child != u8) return error.UnexpectedToken;
                             const source_slice = stringToken.slice(tokens.slice, tokens.i - 1);
+                            const len = stringToken.decodedLength();
+                            const output = try allocator.alloc(u8, len + 1);
+                            errdefer allocator.free(output);
                             switch (stringToken.escapes) {
-                                .None => return allocator.dupe(u8, source_slice),
-                                .Some => {
-                                    const output = try allocator.alloc(u8, stringToken.decodedLength());
-                                    errdefer allocator.free(output);
-                                    try unescapeValidString(output, source_slice);
-                                    return output;
-                                },
+                                .None => mem.copy(u8, output, source_slice),
+                                .Some => try unescapeValidString(output, source_slice),
                             }
+                            output[len] = 0;
+                            return output;
                         },
                         else => return error.UnexpectedToken,
                     }
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1871,19 +1871,33 @@ fn parseInternal(
                                 const v = try parseInternal(ptrInfo.child, tok, tokens, options);
                                 arraylist.appendAssumeCapacity(v);
                             }
+
+                            if (ptrInfo.sentinel) |some| {
+                                const char = @ptrCast(*const u8, some).*;
+                                try arraylist.append(char);
+                                const output = arraylist.toOwnedSlice();
+                                return output[0 .. output.len - 1 :char];
+                            }
+
                             return arraylist.toOwnedSlice();
                         },
                         .String => |stringToken| {
                             if (ptrInfo.child != u8) return error.UnexpectedToken;
                             const source_slice = stringToken.slice(tokens.slice, tokens.i - 1);
                             const len = stringToken.decodedLength();
-                            const output = try allocator.alloc(u8, len + 1);
+                            const output = try allocator.alloc(u8, len + @boolToInt(ptrInfo.sentinel != null));
                             errdefer allocator.free(output);
                             switch (stringToken.escapes) {
                                 .None => mem.copy(u8, output, source_slice),
                                 .Some => try unescapeValidString(output, source_slice),
                             }
-                            output[len] = 0;
+
+                            if (ptrInfo.sentinel) |some| {
+                                const char = @ptrCast(*const u8, some).*;
+                                output[len] = char;
+                                return output[0..len :char];
+                            }
+
                             return output;
                         },
                         else => return error.UnexpectedToken,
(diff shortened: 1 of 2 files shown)


lib/std/json.zig
```diff
@@ -1876,15 +1876,15 @@ fn parseInternal(
                         .String => |stringToken| {
                             if (ptrInfo.child != u8) return error.UnexpectedToken;
                             const source_slice = stringToken.slice(tokens.slice, tokens.i - 1);
+                            const len = stringToken.decodedLength();
+                            const output = try allocator.alloc(u8, len + 1);
```
Vexu, line 1880:
```suggestion
                            const output = try allocator.alloc(u8, len + @boolToInt(ptrInfo.sentinel != null));
```

lib/std/json.zig
```diff
@@ -1876,15 +1876,15 @@ fn parseInternal(
...
-                                .Some => {
-                                    const output = try allocator.alloc(u8, stringToken.decodedLength());
-                                    errdefer allocator.free(output);
-                                    try unescapeValidString(output, source_slice);
-                                    return output;
-                                },
+                                .None => mem.copy(u8, output, source_slice),
+                                .Some => try unescapeValidString(output, source_slice),
                             }
+                            output[len] = 0;
```
Vexu, line 1886:
```suggestion
                            if (ptrInfo.sentinel) |some| output[len] = @ptrCast(*u8, some).*;
```

Vexu (changes requested):
This needs a few adjustments since we don't always want a sentinel.

Also needs a test with different sentinels like `[:0]const u8` and `[:1]const u8`, you can get a value of `[:1]const u8` with `&[3:1]u8{ 'f', 'o', 'o'}`.

marler8997:
It really helps if you could add a test.  You can add a test directly in `json.zig` to make sure it respects sentinel-terminated strings when the field type requires it.

bsansouci:
Hey all, I just added a very simple test and a new testing helper. I'd love to hear what you think of that helper.
I decided that adding a new helper was easier, because I'm not sure what the best way to update the existent `expectEqualSlices` in a way that checks that the `expected` and `actual` are the same type, while allowing the implicit coercion to happen. What I mean is that we'd need to change its type signature to accept `anytype` because there is no easy way to have a parameter type that contains an arbitrary sentinel (as far as I know), and once you pass anytype you need to add some logic to check the types of the two values passed: it shouldn't accept a `u8` and a `[]u8`, but it should accept a `[]u8` and a `*const [3]u8` because one can be implicitly casted into the other, something that happens each time we use a string literal with this function.
If you have some tips on how to do that, we could fixup the `expectEqualSlices` function to support optionally checking the sentinel.

lib/std/testing.zig
```diff
@@ -299,6 +299,41 @@ pub fn expectEqualSlices(comptime T: type, expected: []const T, actual: []const
     }
 }
 
+/// This function is intended to be used only in tests. Checks that two slices or two arrays are equal,
+/// including that their sentinel (if any) are the same. Will error if given another type.
+pub fn expectEqualSlicesSentinel(expected: anytype, actual: anytype) !void {
+    if (expected.len != actual.len) {
+        std.debug.print("slice lengths differ. expected {d}, found {d}\n", .{ expected.len, actual.len });
```
marler8997, line 306:
```zig
std.debug.print("slice lengths differ in length. expected {d}, found {d}\n", .{ expected.len, actual.len });
```

lib/std/testing.zig
```diff
@@ -299,6 +299,41 @@ pub fn expectEqualSlices(comptime T: type, expected: []const T, actual: []const
...
+        return error.TestExpectedEqual;
+    }
+    var i: usize = 0;
+    while (i < expected.len) : (i += 1) {
+        if (!std.meta.eql(expected[i], actual[i])) {
+            std.debug.print("index {} incorrect. expected {any}, found {any}\n", .{ i, expected[i], actual[i] });
```
marler8997, line 312:
```zig
std.debug.print("slices differ at index {} incorrect. expected {any}, found {any}\n", .{ i, expected[i], actual[i] });
```

lib/std/testing.zig
```diff
@@ -299,6 +299,41 @@ pub fn expectEqualSlices(comptime T: type, expected: []const T, actual: []const
     }
 }
 
+/// This function is intended to be used only in tests. Checks that two slices or two arrays are equal,
+/// including that their sentinel (if any) are the same. Will error if given another type.
+pub fn expectEqualSlicesSentinel(expected: anytype, actual: anytype) !void {
```
marler8997, line 304: How about making this more generic, just `expectEqualSlices`, and you check the sentinel if the slice type indicates there is a sentinel?

lib/std/testing.zig
```diff
@@ -299,6 +299,41 @@ pub fn expectEqualSlices(comptime T: type, expected: []const T, actual: []const
...
+            std.debug.print("index {} incorrect. expected {any}, found {any}\n", .{ i, expected[i], actual[i] });
+            return error.TestExpectedEqual;
+        }
+    }
+
+    const expected_sentinel = std.meta.sentinel(@TypeOf(expected));
+    const actual_sentinel = std.meta.sentinel(@TypeOf(actual));
+
+    if (!std.meta.eql(expected_sentinel, actual_sentinel)) {
+        std.debug.print("slice sentinel not equal in the type information. expected {}, found {}\n", .{ expected_sentinel, actual_sentinel });
```
marler8997, line 321: Since this is an issue at comptime, this should be a `@compileError`, something like:
```zig
@compileError("expectSlicesEqual cannot compare " ++ @typeName(@TypeOf(expected)) ++ " and " ++ @typeName(@TypeOf(actual)) ++ " because their sentinel values differ");
```

lib/std/testing.zig
```diff
@@ -299,6 +299,41 @@ pub fn expectEqualSlices(comptime T: type, expected: []const T, actual: []const
...
+        return error.TestExpectedEqual;
+    }
+
+    switch (@typeInfo(@TypeOf(expected))) {
+        .Pointer => {
+            if (expected_sentinel != null and expected[expected.len] != actual[actual.len]) {
```
marler8997, line 327: Testing that these values are true doesn't ensure they are equal to the expected sentinel value.  You also need to check that `actual[actual.len] == expected_sentinel`.

lib/std/testing.zig
marler8997: In fact, you could just extend `expectEqual` to do all this as well.

bsansouci: We talked on Discord, and tried to come up with a way to simply change `expectEqualSlices` but didn't find something simple enough that was worth making.
I also tried simply calling `expectEqualSlices` instead of re-writing it, but it turned into a bit of a mess to make sure that each argument (which can be a pointer or an array) was properly cast-ed to pass to `expectEqualSlices`. That turned out to be more complex code than just doing the same iteration as `expectEqualSlices`, so I ended up going with that method.

lib/std/testing.zig
```diff
@@ -299,6 +299,43 @@ pub fn expectEqualSlices(comptime T: type, expected: []const T, actual: []const
...
+            },
+            .Array => |array_info| {
+                const indexable_outside_of_bounds = @as([]const array_info.child, &actual);
+                break :blk indexable_outside_of_bounds[indexable_outside_of_bounds.len];
+            },
+            else => {},
+        }
+    };
+
+    if (!std.meta.eql(expected_value_sentinel, actual_value_sentinel)) {
```
marler8997, line 333: You haven't checked whether `expected_value_sentinel == sentinel`

bsansouci: Fixed!

lib/std/testing.zig
```diff
@@ -299,6 +299,53 @@ pub fn expectEqualSlices(comptime T: type, expected: []const T, actual: []const
...
+
+    if (!std.meta.eql(sentinel, actual_value_sentinel)) {
+        std.debug.print("expectEqualSentinel: 'actual' sentinel in memory is different from its type sentinel. type sentinel {}, in memory sentinel {}\n", .{ sentinel, actual_value_sentinel });
+        return error.TestExpectedEqual;
+    }
+
+    if (!std.meta.eql(expected_value_sentinel, actual_value_sentinel)) {
+        std.debug.print("expectEqualSentinel: 'expected' and 'actual' parameters have different sentinels in memory. expected {}, found {}\n", .{ expected_value_sentinel, actual_value_sentinel });
+        return error.TestExpectedEqual;
+    }
```
marler8997, line 346: This condition should now be impossible

bsansouci:
Thanks for merging :)

Issue #4221, opened 2020-01-18
Proposal: decimal floating-point types


SaintCharlesCode:
What are your thoughts on Zig having decimal floating-point types built into the language? http://speleotrove.com/decimal/
They could have types df32 df64 df128 or something like that.

JesseRMeyer:
There's also the general possibility of user-space defined Fixed Point storage.

shawnl:
you should probably bring this up with llvm. This is a bunch of work (it needs a full soft float library) and that work would probably belong in llvm.

SaintCharlesCode:
That's true. Also clang does not have "(_Decimal32 and friends)" according to the manual https://clang.llvm.org/docs/UsersManual.html It is probably too much of a niche feature.

DanB91:
Came across this article and adding this in could maybe help in places where COBOL is used: https://medium.com/the-technical-archaeologist/is-cobol-holding-you-hostage-with-math-5498c0eb428b

matu3ba:
This is identical to the more poorly named #1974.
Look especially at this nice comment, which elaborates quite alot: https://github.com/ziglang/zig/issues/1974#issuecomment-620336282

Though I disagree that a separate type should be reserved for it, since the type consists of a tuple.

zzyxyzz:
At the moment, #1974 is only about *binary* fixed-point numbers, so it wouldn't help with this issue.

matu3ba:
> At the moment, #1974 is only about _binary_ fixed-point numbers, so it wouldn't help with this issue.

They are very closely related: https://www.tutorialspoint.com/fixed-point-and-floating-point-number-representations,
so I dont see a special reason to separate them.
Since we have comptime, they could probably even be done together.

If you want to be fancy you could also use [binary-coded decimals](https://en.wikipedia.org/wiki/Binary-coded_decimal).

zzyxyzz:
@matu3ba,
Fixed vs floating point and binary vs decimal are orthogonal problems, as I understand them. The former trades range for uniformity of representable numbers (and sometimes also performance). The latter determines which fractional numbers are representable *exactly* and which aren't. E.g., `0.1` is not representable as a binary fraction, regardless of whether it is fixed or floating. Although the rounding errors are usually small, they sometimes do matter in financial and accounting software. And they can compound quickly, with the classic example being:
```zig
var x = 0.01; // type omitted
var i: i32 = 0;
while (i < 12) : (i += 1) {
[...]

Issue #1974, opened 2019-02-17
Fixed-point number support?


DaZombieKiller:
I discovered Zig yesterday and have spent pretty much all my time since then delving into it and reading various materials on it. Loving it so far.

One feature that I feel is missing currently is built-in support for fixed-point numbers. As the language doesn't allow operator overloading, this isn't something I could implement in user code and use like any other numeric type. For reference, the [Embedded C standard](http://www.open-std.org/JTC1/SC22/WG14/www/docs/n1169.pdf) specifies support for such numbers (refer to Annex A), although I'm not aware of any C compiler that supports them for a desktop target.

I found it interesting that Zig offers support for variable-width integers through `i#` and `u#`. I wonder if this could be applied to fixed-point numbers by supporting some form of Q-notation, perhaps?

andrewrk:
LLVM has 
 * `@llvm.smul.fix.*`
 * `@llvm.umul.fix.*`

https://llvm.org/docs/LangRef.html#fixed-point-arithmetic-intrinsics

But that's it. Interesting. I'll have to learn why these intrinsics are provided. It may be that some architectures provide some fixed point multiplication operations.

The null hypothesis here is that you would have to implement fixed-point in userland, and yes operations would be function calls rather than overloaded operators.

ryanavella:
Several architectures do indeed provide fixed-point multiplication operations. For instance, Atmel's AVR UC3 provides fixed-point DSP arithmetic, but has no support for floating point representations. This means that e.g. FFT's are blazingly fast. This is fairly typical for a MCU catering to DSP applications.

Diltsman:
Most processors that I use at work don't even have a hardware multiply.  Generally we work in ADC counts, but sometimes we will work in fixed-point encoded in 8 or 16 bit integers.

It seems to me that, if LLVM IR has minimal support for fixed-point numbers (only data type, no operations), that it wouldn't be terribly difficult to map fixed-point operations to sequences of bit-shifts and integer operations.

floopfloopfloopfloopfloop:
# Fixedpoint Numbers
[wow-very-large-such-number]: #wow-very-large-such-number

An approximation of real numbers using only integer hardware.

# Why do we want this?
[motivation-such-an-aggravation]: #motivation-such-an-aggravation

They are useful for
a) software running on hardware with no floating point support
b) software that benefits from both cross platform determinism and fractional precision, such as video games using lockstep networking
c) software designed to maximize performance of small decimal numbers using simd operations

[...]

PR #10897, opened 2022-02-15
compiler_rt: specify goals, organize README and compiler_rt.zig

matu3ba:
* goals
  - zig as linker for object files generated by other compilers
  - zig-specific runtime features for eventual standardisation

* changes
  - missing routines are marked with `missing`
  - structure inspired by libgcc docs, but improved order and wording
  - rename misspelled functions
  - reorder and rephrase compiler_rt.zig to reflect documentation
  - potential decimal float or fixed-point arithmetic support:
    * 'Decimal float library routines' ca. 120 functions
    * 'Fixed-point fractional library routines' ca. 300 functions

thanks to @Vexu for multiple reviews and @scheibo for review

### DIFF
--- a/lib/std/special/compiler_rt/README.md
+++ b/lib/std/special/compiler_rt/README.md
@@ -1,15 +1,267 @@
-This compiler-rt library is ported from [LLVM](http://compiler-rt.llvm.org/).
+If hardware lacks basic or specialized functionality, compiler-rt adds such functionality
+for basic arithmetic(s).
+One such example is 64-bit integer multiplication on 32-bit x86.
 
-It's needed because LLVM emits library calls to compiler-rt when hardware lacks
-functionality, for example, 64-bit integer multiplication on 32-bit x86.
+Goals:
+1. zig as linker for object files produced by other compilers
+   => `function compatibility` to compiler-rt and libgcc for same-named functions
+   * compatibility conflict between compiler-rt and libgcc: prefer compiler-rt
+2. `symbol-level compatibility` low-priority compared to emitted calls by llvm
+   * symbol-level compatibility: libgcc even lower priority
+3. add zig-specific language runtime features, see #7265
+   * example: arbitrary bit width integer arithmetic
+   * lower to call those functions for e.g. multiplying two i12345 numbers together
+   * proper naming + documention for standardizing (allow languages to follow our exmaple)
+
+Current status (tracking libgcc documentation):
+- Integer library routines => almost implemented
+- Soft float library routines => only f80 routines missing
+- Decimal float library routines => unimplemented (~120 functions)
+- Fixed-point fractional library routines => unimplemented (~300 functions)
+- Exception handling routines => unclear, if supported (~32+x undocumented functions)
+- Miscellaneous routines => unclear, if supported (cache control and stack function)
+- No zig-specific language runtime features in compiler-rt yet
 
 This library is automatically built as-needed for the compilation target and
 then statically linked and therefore is a transparent dependency for the
 programmer.
+For details see `../compiler_rt.zig`.
 
-Any bugs should be solved by trying to duplicate the bug upstream.
- * If the bug exists upstream, get it fixed with the LLVM team and then port
-   the fix downstream to Zig.
- * If the bug only exists in Zig, something went wrong porting the code,
-   and you can run the C code and Zig code side by side in a debugger
-   to figure out what's happening differently.
+The routines in this folder are listed below.
+Routines are annotated as `type source routine // description`, with `routine`
+being the name used in aforementioned `compiler_rt.zig`.
+`dev` means deviating from compiler_rt, `port` ported, `source` is the
+information source for the implementation, `none` means unimplemented.
+Some examples for the naming convention are:
+- dev source name_routine, name_routine2 various implementations for performance, simplicity etc
+- port llvm compiler-rt library routines from [LLVM](http://compiler-rt.llvm.org/)
+   * LLVM emits library calls to compiler-rt, if the hardware lacks functionality
+- port musl libc routines from [musl](https://musl.libc.org/)
+If the library or information source is uncommon, use the entry `other` for `source`.
+Please do not break the search by inserting entries in another format than `impl space source`.
+
+Bugs should be solved by trying to duplicate the bug upstream, if possible.
+ * If the bug exists upstream, get it fixed upstream and port the fix downstream to Zig.
+ * If the bug only exists in Zig, use the corresponding C code and debug
+   both implementations side by side to figure out what is wrong.
+
+## Integer library routines
+
+#### Integer Bit operations
+- dev HackersDelight __clzsi2          // count leading zeros
+- dev HackersDelight __clzdi2          // count leading zeros
+- dev HackersDelight __clzti2          // count leading zeros
+- dev HackersDelight __ctzsi2          // count trailing zeros
+- dev HackersDelight __ctzdi2          // count trailing zeros
+- dev HackersDelight __ctzti2          // count trailing zeros
+- dev __ctzsi2 __ffssi2                // find least significant 1 bit
+- dev __ctzsi2 __ffsdi2                // find least significant 1 bit
+- dev __ctzsi2 __ffsti2                // find least significant 1 bit
+- dev BitTwiddlingHacks __paritysi2    // bit parity
+- dev BitTwiddlingHacks __paritydi2    // bit parity
+- dev BitTwiddlingHacks __parityti2    // bit parity
+- dev TAOCP __popcountsi2              // bit population
+- dev TAOCP __popcountdi2              // bit population
+- dev TAOCP __popcountti2              // bit population
+- dev other __bswapsi2                 // a byteswapped
+- dev other __bswapdi2                 // a byteswapped
+- dev other __bswapti2                 // a byteswapped
+
+#### Integer Comparison
+- port llvm __cmpsi2        // (a<b)=>output=0, (a==b)=>output=1, (a>b)=>output=2
+- port llvm __cmpdi2
+- port llvm __cmpti2
+- port llvm __ucmpsi2       // (a<b)=>output=0, (a==b)=>output=1, (a>b)=>output=2
+- port llvm __ucmpdi2
+- port llvm __ucmpti2
+
+#### Integer Arithmetic
+- none none __ashlsi3              // a << b unused in llvm, missing (e.g. used by rl78)
+- port llvm __ashldi3              // a << b
+- port llvm __ashlti3              // a << b
+- none none __ashrsi3              // a >> b  arithmetic (sign fill) missing (e.g. used by rl78)
+- port llvm __ashrdi3              // a >> b  arithmetic (sign fill)
+- port llvm __ashrti3              // a >> b  arithmetic (sign fill)
+- none none __lshrsi3              // a >> b  logical    (zero fill) missing (e.g. used by rl78)
+- port llvm __lshrdi3              // a >> b  logical    (zero fill)
+- port llvm __lshrti3              // a >> b  logical    (zero fill)
+- port llvm __negdi2               // -a symbol-level compatibility: libgcc
+- port llvm __negti2               // -a unnecessary: unused in backends
+- port llvm __mulsi3               // a * b  signed
+- port llvm __muldi3               // a * b  signed
+- port llvm __multi3               // a * b  signed
+- port llvm __divsi3               // a / b  signed
+- port llvm __divdi3               // a / b  signed
+- port llvm __divti3               // a / b  signed
+- port llvm __udivsi3              // a / b  unsigned
+- port llvm __udivdi3              // a / b  unsigned
+- port llvm __udivti3              // a / b  unsigned
+- port llvm __modsi3               // a % b  signed
+- port llvm __moddi3               // a % b  signed
+- port llvm __modti3               // a % b  signed
+- port llvm __umodsi3              // a % b  unsigned
+- port llvm __umoddi3              // a % b  unsigned
+- port llvm __umodti3              // a % b  unsigned
+- port llvm __udivmoddi4           // a / b, rem.* = a % b  unsigned
+- port llvm __udivmodti4           // a / b, rem.* = a % b  unsigned
+- port llvm __udivmodsi4           // a / b, rem.* = a % b  unsigned
+- port llvm __divmodsi4            // a / b, rem.* = a % b  signed, ARM
+
+#### Integer Arithmetic with trapping overflow
+- dev BitTwiddlingHacks __absvsi2  // abs(a)
+- dev BitTwiddlingHacks __absvdi2  // abs(a)
+- dev BitTwiddlingHacks __absvti2  // abs(a)
+- port llvm __negvsi2              // -a symbol-level compatibility: libgcc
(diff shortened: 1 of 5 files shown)


lib/std/special/compiler_rt/README.md
```diff
@@ -1,15 +1,258 @@
...
+- dev HackersDelight __mulodi4     // a * b, overflow=>ov.*=1 else 0 (required by llvm)
+- dev HackersDelight __muloti4     // a * b, overflow=>ov.*=1 else 0 (required by llvm)
+
+## Soft float library routines
+
+#### Conversion
+
+- __extendsfdf2  // extend a f32 => f64
+- __extendsftf2  // extend a f32 => f128
+- __extendsfxf2  // extend a f32 => f80 missing
```
Vexu, line 135: This and `__truncxfsf2` exist but I misnamed them using `f` instead of `s`.

lib/std/special/compiler_rt/README.md
```diff
@@ -1,15 +1,258 @@
...
+- __fixunsxfdi   // missing
+- __fixunssfti   // convert to u128, rounding towards zero. negative values become 0.
+- __fixunsdfti   //
+- __fixunstfdi   //
+- __fixunsxfti   // missing
+
+- __floatsisf    // convert i32 to floating point
+- __floatsidf    //
+- __floatsitf    // missing
+- __floatsixf    //
```
Vexu, line 172:
```suggestion
- __floatsitf    //
- __floatsixf    // missing
```

lib/std/special/compiler_rt/README.md
```diff
@@ -1,15 +1,258 @@
...
+- __floatuntitf  //
+- __floatuntixf  // missing
+
+#### Comparison
+- __cmpsf2       // missing return (a<b)=>-1,(a==b)=>0,(a>b)=>1,Nan=>1 dont rely on this
+- __cmpdf2       // missing
+- __cmptf2       // missing
+- __unordsf2     // missing (input==NaN) => out!=0 else out=0,
+- __unorddf2     // missing __only reliable for (input!=Nan)__
+- __unordtf2     // missing
```
Vexu, line 200: All of these exist?

Vexu (commented):
The formatting seems fine to me. I'd assume fixed-point and decimal float support depend on #1974 and #4221.

lib/std/special/compiler_rt/README.md
matu3ba: They are then misnamed. For example grepping for `__cmpsf2` I found
```txt
[user@pc compiler_rt]$ rg __cmpsf2
comparesf2_test.zig
23:fn test__cmpsf2(vector: TestVector) bool {
99:        try std.testing.expect(test__cmpsf2(vector));

README.md
195:- __cmpsf2       // missing return (a<b)=>-1,(a==b)=>0,(a>b)=>1,Nan=>1 dont rely on this
```

Vexu: The first three are just exports of a compatible function under a different name:
https://github.com/ziglang/zig/blob/cf5009f9af118716cbe97a28957cb1ca8d047f44/lib/std/special/compiler_rt.zig#L98-L100

The others are handled like any other functions:
https://github.com/ziglang/zig/blob/cf5009f9af118716cbe97a28957cb1ca8d047f44/lib/std/special/compiler_rt.zig#L239-L244

lib/std/special/compiler_rt/README.md
matu3ba: adjusted

lib/std/special/compiler_rt/README.md
matu3ba: renamed.

lib/std/special/compiler_rt/README.md
matu3ba: ah, good catch. I added that.

lib/std/special/compiler_rt/README.md
Vexu, line 135: This still says it's missing.

lib/std/special/compiler_rt/README.md
```diff
@@ -1,15 +1,258 @@
...
+- __multf3       // a * b
+- __mulxf3       // a * b missing
+- __divsf3       // a / b
+- __divdf3       // a / b
+- __divtf3       // a / b
+- __divxf3       // a / b missing
+- __negsf2       // -a
+- __negdf2       // -a
+- __negtf2       // -a missing
+- __negxf2       // -a missing
```
Vexu, line 244: These don't exist because calls to them are never generated. LLVM lowers `__negtf2` directly into an xor and we do the same for `f80` in `ir_gen_soft_f80_neg` which regressed with #10861.

Here's a patch to fix it:
```diff
diff --git a/src/stage1/codegen.cpp b/src/stage1/codegen.cpp
index dc8f5e590..8bb5430ae 100644
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -4115,13 +4115,9 @@ static LLVMValueRef ir_render_binary_not(CodeGen *g, Stage1Air *executable,
 static LLVMValueRef ir_gen_soft_f80_neg(CodeGen *g, ZigType *op_type, LLVMValueRef operand) {
     uint32_t vector_len = op_type->id == ZigTypeIdVector ? op_type->data.vector.len : 0;
 
-    uint64_t buf[2] = {0, 0};
-    if (g->is_big_endian != native_is_big_endian) {
-        buf[1] = 0x8000000000000000;
-    } else {
-        buf[1] = 0x8000;
-    }
-    LLVMValueRef sign_mask = LLVMConstIntOfArbitraryPrecision(LLVMInt128Type(), 2, buf);
+    LLVMTypeRef llvm_i80 = LLVMIntType(80);
+    LLVMValueRef sign_mask = LLVMConstInt(llvm_i80, 1, false);
+    sign_mask = LLVMConstShl(sign_mask, LLVMConstInt(llvm_i80, 79, false));
 
     LLVMValueRef result;
     if (vector_len == 0) {
```

lib/std/special/compiler_rt/README.md
```diff
@@ -1,15 +1,258 @@
...
+The routines in this folder are listed below.
+Routines are annotated as `type source routine // desription`, with `routine`
+being the name used in aforementioned `compiler_rt.zig`.
+`dev` means deviating from compiler_rt, `port` ported, `source` is the
+information source for the implementation, `none` means unimplemented.
+In the following a brief overview of is given:
+- dev source name_routine, name_routine2 various implementations for performance, simplicity etc
+- port llvm compiler-rt library routines from [LLVM](http://compiler-rt.llvm.org/)
+   * LLVM emits library calls to compiler-rt, if the hardware lacks functionality
+- port softfloat floating point routines from [berkeley-softfloat](https://github.com/ucb-bar/berkeley-softfloat-3)
```
Vexu, line 26: Note that currently no functions are ported from berkley softlfoat, the ones I added were adapted from the ones ported from LLVM.

Vexu (changes requested):

lib/std/special/compiler_rt/README.md
matu3ba: I can make a PR, if I know what to test for. Do you have the test cases in arocc?

lib/std/special/compiler_rt/README.md
matu3ba: Sorry about that. Should be now fixed.

lib/std/special/compiler_rt/README.md
matu3ba: I added the parts of negtf2 and negxf2 being unnecessary with your written reason.

Vexu:
I looked into LLVM and no calls to of the `__neg` functions, integer or float, are being generated and they are there only for libgcc ABI compatibility which I'm not sure is something we care about.

matu3ba:
@Vexu Did you also check what (for example `__neg` routines are used for in gcc? 
For example the integer routines in those files use them (but the float stuff is also used in microcontroller routines):
```txt
libgcc/config/avr/lib1funcs-fixed.S
libgcc/config/avr/lib1funcs.S
libgcc/config/avr/libf7/libf7-asm.sx
libgcc/config/cr16/lib1funcs.S
libgcc/config/nds32/lib1asmsrc-mculib.S
```
Of course we could also implement it in assembly for each target, but 1. this can increase initial burden for contributing/upstreaming a target depending on it and 2. (not sure about this and if it applies to cheap hardware) there might be an efficiency loss due to suboptimal register choice (as there should be other registers to spare).
What do you think on this?

matu3ba:
@Vexu Sorry for the rebase. The commit log got too long and cluttered.

For the integer stuff I checked gcc and it looks like they are legacy and unused function.
For float stuff `__negsf2` is used by `rl78`, so I added comments like `__negsf2       // -a (e.g. used by rl78)` for each "simple" routine.

Do you also want me to apply your patch without tests?

matu3ba:
The drone runner looks broke and I am not sure why. Log:
```txt
1 ./ci/zinc/linux_probe.sh
0s
2/bin/sh: 39: ./ci/zinc/linux_probe.sh: not found
0s
```

lib/std/special/compiler_rt/README.md
matu3ba: @Vexu applied the patch in 6235afc6, so this is fixed.

matu3ba:
@Vexu Do you see further problems? I am open to remove `__negsf2`, if you give me a reason why it is used in `gcc` and unnecessary for Zig. Otherwise, I would prefer to do this in another PR.

Vexu (approved):
> I am open to remove `__negsf2` or other things, if you give me a reason why it is used in `gcc`

I doubt it is used, it's likely just left for ABI backwards compatibility

> and unnecessary for Zig.

`__negXf2` is just a fancy XOR.

---

@andrewrk this removes some functions from compiler-rt for which LLVM doesn't generate calls to, making it slightly less compatible with LLVM's compiler-rt and libgcc but I'm not sure if that's even a goal.

andrewrk:
> this removes some functions from compiler-rt for which LLVM doesn't generate calls to, making it slightly less compatible with LLVM's compiler-rt and libgcc but I'm not sure if that's even a goal.

I put some thought into it just now and my conclusions are:

 * we do want to be compatible with compiler-rt and libgcc for the same-named functions. This allows zig to act as a linker for object files produced by other compilers. If there is a compatibility conflict between compiler-rt and libgcc, prefer compiler-rt.
 * compiler-rt symbol-level compatibility is low priority compared to implementing the functions that LLVM will emit calls to in practice. libgcc compatibility is even lower priority than that.
 * we will also extend compiler-rt to add more zig-specific language runtime features such as arbitrary bit width integer arithmetic, so that we can lower to call those functions for e.g. multiplying two `i12345` numbers together. Related: #7265
    - let's try to name these extensions well and document them clearly so that we have a chance of other languages following our example and standardizing on these extensions.

lib/std/special/compiler_rt/README.md
```diff
@@ -1,15 +1,257 @@
...
 
-It's needed because LLVM emits library calls to compiler-rt when hardware lacks
-functionality, for example, 64-bit integer multiplication on 32-bit x86.
+Current status (tracking libgcc documentation):
+- Integer library routines => almost implemented
+- Soft float library routines => only f80 routines missing
+- Decimal float library routines => unimplemented (~120 functions)
+- Fixed-point fractional library routines => unimplemented (~300 functions)
+- Exception handling routines => unclear, if supported (~32+x undocumented functions)
+- Miscellaneous routines => unclear, if supporetd (cache control and stack function)
```
scheibo, line 11:
```suggestion
- Miscellaneous routines => unclear, if supported (cache control and stack function)
```

lib/std/special/compiler_rt/README.md
```diff
@@ -1,15 +1,257 @@
...
+For details see `../compiler_rt.zig`.
 
-Any bugs should be solved by trying to duplicate the bug upstream.
- * If the bug exists upstream, get it fixed with the LLVM team and then port
-   the fix downstream to Zig.
- * If the bug only exists in Zig, something went wrong porting the code,
-   and you can run the C code and Zig code side by side in a debugger
-   to figure out what's happening differently.
+The routines in this folder are listed below.
+Routines are annotated as `type source routine // desription`, with `routine`
```
scheibo, line 19:
```suggestion
Routines are annotated as `type source routine // description`, with `routine`
```

lib/std/special/compiler_rt/README.md
```diff
@@ -1,15 +1,257 @@
...
+- port musl libc routines from [musl](https://musl.libc.org/)
+If the library or information source is uncommon, use the entry `other` for `source`.
+Please do not break the search by inserting entries in another format than `impl space source`.
+
+Bugs should be solved by trying to duplicate the bug upstream, if possible.
+ * If the bug exists upstream, get it fixed upstream and port the fix downstream to Zig.
+ * If the bug only exists in Zig, use the corresponding C code and debug
+   both implementations side by side to figure out what is wrong.
+
+impl source routine // description
```
scheibo, line 36:
```suggestion
```

lib/std/special/compiler_rt/README.md
```diff
@@ -1,15 +1,257 @@
...
-   the fix downstream to Zig.
- * If the bug only exists in Zig, something went wrong porting the code,
-   and you can run the C code and Zig code side by side in a debugger
-   to figure out what's happening differently.
+The routines in this folder are listed below.
+Routines are annotated as `type source routine // desription`, with `routine`
+being the name used in aforementioned `compiler_rt.zig`.
+`dev` means deviating from compiler_rt, `port` ported, `source` is the
+information source for the implementation, `none` means unimplemented.
+In the following a brief overview of is given:
```
scheibo, line 23: I'm having difficult parsing this sentence - I think you may be missing a word after "of" ("In the following" also seems perhaps redundant)?

matu3ba: good catch.

lib/std/special/compiler_rt/README.md
matu3ba: fixed.

lib/std/special/compiler_rt/README.md
matu3ba: fixed.

lib/std/special/compiler_rt/README.md
matu3ba: fixed.

matu3ba: fixed.

matu3ba:
I did add the following and restored `negvX` and `negX`, since they are "symbol-level compatibility" and partially "function-level compatibility" as I understand it.

```txt
Goals:
1. zig as linker for object files produced by other compilers
   => `function compatibility` to compiler-rt and libgcc for same-named functions
   * compatibility conflict between compiler-rt and libgcc: prefer compiler-rt
2. `symbol-level compatibility` low-priority compared to emitted calls by llvm
   * symbol-level compatibility: libgcc even lower priority
3. add zig-specific language runtime features, see #7265
   * example: arbitrary bit width integer arithmetic
   * lower to call those functions for e.g. multiplying two i12345 numbers together
   * proper naming + documention for standardizing (allow languages to follow our exmaple)
```

@andrewrk Is this good to merge?

andrewrk:
Thanks @matu3ba!

PR #10899, opened 2022-02-15
aarch64,macos: handle GOT and direct loads in codegen

kubkon:
* disable failing `aarch64-macos` behavior tests
* introduce `MCValue.got_load` and `MCValue.direct_load` - this matches the current design in `x86_64` backend and simplifies handling of PIE targets in `aarch64` backend
* add `lowerUnnamedConst` glue to codegen

cc @joachimschmidt557

### DIFF
--- a/test/behavior/basic.zig
+++ b/test/behavior/basic.zig
@@ -197,6 +197,9 @@ test "multiline string comments at multiple places" {
 }
 
 test "string concatenation" {
+    if (builtin.zig_backend == .stage2_aarch64 and builtin.os.tag == .macos) return error.SkipZigTest;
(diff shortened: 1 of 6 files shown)


test/behavior/basic.zig
```diff
@@ -197,6 +197,8 @@ test "multiline string comments at multiple places" {
 }
 
 test "string concatenation" {
+    if (builtin.zig_backend == .stage2_aarch64) return error.SkipZigTest;
```
joachimschmidt557, line 200: As far as I know, this test passes on `aarch64-linux`.

joachimschmidt557 (approved):

test/behavior/basic.zig
joachimschmidt557: Apparently this test fails when compiling on macOS, regardless of the backend.

PR #10900, opened 2022-02-15
stage2: emit AIR for decl scanning when --verbose-air is specified

mitchellh:
This modifies `--verbose-air` to also output the AIR generated during decl analysis. 

The AIR generated here is typically (always?) a series of 1 or more `.constant` types because this scope is forced comptime. I've still found dumping this to be useful in learning how the frontend of stage2 works and debugging some comptime behavior.

Generally, I don't think this is super useful since you could always debug comptime stuff by pulling it into a function and wrapping it in `comptime` so I don't _think_ this enables anything new.

As a final warning: this makes `--verbose-air` output very noisy, since it always dumps the AIR for every decl imported since all decls are scanned, unlike function bodies which are only analyzed if they're referenced/exported/etc. If this is desired, we may want a separate flag, or to change `--verbose-air` to take an integer verbosity level or something.

Opening this PR in case this is desired, completely understand if it isn't wanted. 😄

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -3703,6 +3703,44 @@ fn semaDecl(mod: *Module, decl: *Decl) !bool {
         };
     };
 
+    if (mod.comp.verbose_air) {
+        // Duplicate everything so that we can make modifications. It doesn't
+        // look like the rest of this fn uses this so we could modify it in
+        // place but this makes it more future-proof to any semaDecl changes.
+        var air_extra = try sema.air_extra.clone(gpa);
+        var air_instructions = try sema.air_instructions.clone(gpa);
+        var air_values = try sema.air_values.clone(gpa);
+        defer air_extra.deinit(gpa);
+        defer air_instructions.deinit(gpa);
+        defer air_values.deinit(gpa);
+
+        // Copy the block into place and mark that as the main block. This is
+        // required for print_air to find the "main" block of the AIR.
+        try air_extra.ensureUnusedCapacity(gpa, @typeInfo(Air.Block).Struct.fields.len +
+            block_scope.instructions.items.len);
+        const main_block_index = @intCast(u32, air_extra.items.len);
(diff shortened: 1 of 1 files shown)


src/Module.zig
```diff
@@ -3703,6 +3703,44 @@ fn semaDecl(mod: *Module, decl: *Decl) !bool {
         };
     };
 
+    if (mod.comp.verbose_air) {
```
andrewrk, line 3706:
```suggestion
    if (builtin.mode == .Debug and mod.comp.verbose_air) {
```

This makes us not have to pay the cost of the code in this block for release builds

src/Module.zig
```diff
@@ -3703,6 +3703,44 @@ fn semaDecl(mod: *Module, decl: *Decl) !bool {
...
+
+        var air = Air{
+            .instructions = air_instructions.toOwnedSlice(),
+            .extra = air_extra.toOwnedSlice(gpa),
+            .values = air_values.toOwnedSlice(gpa),
+        };
+        defer air.deinit(gpa);
+
+        var liveness = try @import("Liveness.zig").analyze(gpa, air, zir);
+        defer liveness.deinit(gpa);
```
andrewrk, line 3734: We don't really want to analyze Liveness in this case - doing it just so that we can print the AIR is a bit misleading and is triggering code paths that were never meant to be covered.

I get that this would then require a bunch more effort to make AIR printable without liveness information attached to it.

andrewrk (changes requested):
Thanks for the patch!

There are a couple things I'd like to address here before merging this.

To be honest I'm not sure it's worth the effort here, especially considering I eventually want to migrate from this constant value system to [InternPool](https://github.com/ziglang/zig/blob/e620b692c6c9f7c4b655f30aae5e51a87dc56e91/src/InternArena.zig) which would cause a rewrite of this block of code anyway.

mitchellh:
Let's drop it! I can understand why this gets more complex and its not worth it imo.

PR #10906, opened 2022-02-16
stage2 CBE: Implement 2D array support

topolarity:
This PR adds basic multi-dimensional array support to the C backend.

The core of the change is lower pointers to 1D arrays as `[*c][N]T`, rather than `[*c]T`. This allows correct pointer arithmetic based on element pointers in 2D arrays

Higher-dimensional arrays are also supported (e.g. pointer to 2D array is lowered as `[*c][N][M]T`, etc.)

Also includes miscellaneous changes to:
     - refactor CValue dereferencing logic
     - fix up elem_ptr and field_ptr handling, including to undefined pointers
     - add `renderTypecast` (renders in C typecast format, e.g. "int* [10]")

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -291,6 +291,19 @@ pub const Function = struct {
         }
     }
 
+    fn writeCValueDeref(f: *Function, w: anytype, c_value: CValue) !void {
+        switch (c_value) {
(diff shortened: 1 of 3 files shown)


andrewrk:
Great work!

PR #10907, opened 2022-02-16
CI: windows: depend on Zig instead of CMake and MSVC

andrewrk:
Instead of assuming that the zig version used with `zig build` is appropriate for building the self-hosted compiler component, build.zig now follows the same path as the cmake build script and build zig0, using that to produce zig1.o, re-linking to produce stage1.

This allows an arbitrarily old Zig version and corresponding build.zig script to build all future versions of Zig from source via the bootstrap compiler. In other words, it allows us to use `zig build` as an alternative to cmake when bootstrapping. We copy build.zig to the ci/azure directory and make it compatible with 0.9.1 so that it continues to function properly even as we make breaking changes to the zig language, standard library, and build system.

The motivation for this change was updating the CI tarball from LLVM 13.0.0 to 13.0.1. Now, instead of wrangling MSVC on a Windows computer in order to update the CI tarballs whenever LLVM updates, the process is to use [zig-bootstrap](https://github.com/ziglang/zig-bootstrap) and [release-cutter](https://github.com/ziglang/release-cutter) to produce a new Windows dev kit zip file. This makes it possible to do from any operating system and MSVC is no longer part of the process.

It also makes Windows dev kits much more resilient to breaking Zig changes. A follow-up change could add a new flag to `zig build` called `--override-build-script` which allows passing in a build.zig file from a different location, without changing the build root directory to match. This would make a Windows dev kit immune to breaking changes in the Zig language and standard library.


ci/azure/windows_script
```diff
@@ -3,11 +3,60 @@
...
+  -Dstrip \
+  -Duse-zig-libcxx \
+  -Dtarget=$TARGET
+
+# Now that we have built an up-to-date zig.exe, we restore the original
+# build script from master branch.
+mv build.zig.master build.zig
+
+#dist/bin/zig.exe build test-toolchain -Dskip-non-native -Dskip-stage2-tests
+#dist/bin/zig.exe build test-std -Dskip-non-native
```
andrewrk, line 54: oops, I meant to re-enable these. they did pass already once in a CI run so I know they're good. If the rest of the CI tests pass for this branch, I'll re-enable these and force push just before merging.

kubkon (approved):

PR #10909, opened 2022-02-17
refAllDecls handle pub instances of structs

### LINKED ISSUE #10908: refAllDecls does not handle pub instances of structs
### Zig Version

0.10.0-dev.760+1a84c23d6

### Steps to Reproduce

Test the below file.

```zig
const std = @import("std");

pub const SInstance: S = .{};

pub const S = struct {
    a: usize = 0,
    b: bool = false,
};

comptime {
}
```

### Expected Behavior

No compile error.

### Actual Behavior

```
zig test src/main.zig                                                                                                                                                 237ms  Thu 17 Feb 2022 00:54:08 GMT
/home/lee/src/github.com/leecannon/zig/lib/std/testing.zig:470:26: error: expected type 'type', found 'S'
            _ = @hasDecl(@field(T, decl.name), "foo");
                         ^
./src/main.zig:11:28: note: called from here
                           ^
./src/main.zig:10:10: note: called from here
comptime {
         ^
```

leecannon:
To prevent calling `@hasDecl` on an instance of a struct we need to check the `@TypeOf` the decl is actually `type`.

This PR closes #10908

### DIFF
--- a/lib/std/testing.zig
+++ b/lib/std/testing.zig
@@ -466,13 +466,6 @@ test {
 pub fn refAllDecls(comptime T: type) void {
     if (!builtin.is_test) return;
     inline for (comptime std.meta.declarations(T)) |decl| {
-        if (decl.is_pub) {
-            const decl_value = @field(T, decl.name);
-            const type_of_decl = @TypeOf(decl_value);
-            if (type_of_decl == type and @typeInfo(type_of_decl) == .Struct) {
-                _ = @hasDecl(decl_value, "foo");
-            }
-        }
-        _ = decl;
+        if (decl.is_pub) _ = @field(T, decl.name);
     }
 }
--- a/lib/std/testing.zig
+++ b/lib/std/testing.zig
@@ -466,8 +466,13 @@ test {
 pub fn refAllDecls(comptime T: type) void {
     if (!builtin.is_test) return;
     inline for (comptime std.meta.declarations(T)) |decl| {
-        if (decl.is_pub and @typeInfo(@TypeOf(@field(T, decl.name))) == .Struct)
-            _ = @hasDecl(@field(T, decl.name), "foo");
+        if (decl.is_pub) {
+            const decl_value = @field(T, decl.name);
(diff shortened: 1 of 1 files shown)


lib/std/testing.zig
```diff
@@ -466,8 +466,13 @@ test {
 pub fn refAllDecls(comptime T: type) void {
     if (!builtin.is_test) return;
     inline for (comptime std.meta.declarations(T)) |decl| {
-        if (decl.is_pub and @typeInfo(@TypeOf(@field(T, decl.name))) == .Struct)
-            _ = @hasDecl(@field(T, decl.name), "foo");
+        if (decl.is_pub) {
+            const decl_value = @field(T, decl.name);
+            const type_of_decl = @TypeOf(decl_value);
+            if (type_of_decl == type and @typeInfo(type_of_decl) == .Struct) {
```
nektro, line 472: enums and unions can also be containers

Vexu, line 472: This is always false but it seems that just doing `@field(T, decl.name)` is enough to force the declaration to be analyzed.

InKryption:
> @Vexu your fix is 100x simpler than mine:
> 
> ```zig
> pub fn refAllDecls(comptime T: type) void {
>     if (!builtin.is_test) return;
>     inline for (comptime std.meta.declarations(T)) |decl| {
>         if (decl.is_pub) {
>             const decl_value = @field(T, decl.name);
>             if (@TypeOf(decl_value) == type) {
>                 switch (@typeInfo(decl_value)) {
>                     .Struct, .Union, .Enum, .Opaque => _ = @hasDecl(decl_value, "foo"),
>                     else => {},
>                 }
>             }
>         }
>         _ = decl;
>     }
> }
> ```

Just to note, you could also use `std.meta.trait.isContainer` here, instead of doing the switch manually.

leecannon:
@InKryption I though I remembered an `isContainer` but I only looked in `std.meta` instead of `std.meta.trait`

PR #10913, opened 2022-02-17
further parser error improvements

### LINKED ISSUE #416: Improve the error message on missing ':' in while statement
As shown in the following example.

Was expecting it to complain on the `while (...) (...) {`.

```
// -*- indent-tabs-mode:nil; -*-
const std = @import("std");
const print = std.io.stdout.printf;

pub fn main() -> %void {
    const buf : [3]u8 = []u8{'a','b','c'};

    var i : usize = 0;
    const count : usize = buf.len;
    while (i < count) : (i += 1) {
            %%print("{}/{}/{}\n", i, buf[i], count);
    }
}

// zig commit 221e5b

// step 1
// original working program
// build_exe and run
// output:
// 0/97/3
// 1/98/3
// 2/99/3

// step 2
// remove ':' from while line (or forget to put it in)
// while (i < count) (i += 1)
// build_exe
// y.zig:11:52: error: expected token '}', found ';'
//            %%print("{}/{}/{}\n", i, buf[i], count);
//                                                   ^

// step 3
// remove ';' as indicated by error message
// build_exe and run (panics)
// output:
// 1/98/3
// 2/99/3
// index out of bounds
// : snip, snip
// @import("std").debug.panic("{}", message_ptr[0..message_len]);

```

### LINKED ISSUE #5739: Unhelpful/misleading error message on extra symbols between struct functions
Given the following test `.zig` file:
```zig
const Example = struct {
    some_var: i32,

    fn a() void { }
    fn b() void { }
    fn c() void { } e
    fn d() void { }
};

pub fn main() void { }
```

Zig throws the following compilation error:
```
> zig build-exe play.zig
.\play.zig:4:5: error: declarations are not allowed between container fields
    fn a() void { }
    ^
```

Unfortunately, this is completely not helpful and misleading, since it points to the first occurrence of the first function within a struct rather than the obviously erroneous `e` symbol after the `fn c()` declaration.
This is not a problem in such a small example but quickly becomes an issue in a huge source file when it is not obvious what the error is really caused by.

The best possible possible error in this scenario would be something like:
```
> zig build-exe play.zig
.\play.zig:6:21: error: unexpected symbol 'e' between container declarations
    fn c() void { } e
                    ^
```

### LINKED ISSUE #10511: bug: zig fmt accepts while with bad continue expression
### Zig Version

0.9.0-dev.2025+ecf0050a9

### Steps to Reproduce

write a while loop with a continue expression without a colon and at the end of the while loop write a semicolon
```zig
pub fn main() void {
    const somevalue = true;
    while (true) (somevalue){};
}
```
and run `zig fmt`.
you can also instead of the semicolon write an else expression
```zig
pub fn main() void {
    const somevalue = true;
    while (true) (somevalue){} else false;
}
```

### Expected Behavior

the call to `zig fmt` would print an error mentioning the missing colon. something like
```
file.zig:4:1: error: expected ':' or '{', found '('
    while (true) (somevalue){};
                       ^
```

### Actual Behavior

considering the file
```zig
pub fn main() void {
    const somevalue = true;
    while (true) (somevalue){};
}
```
the call to `zig fmt` accepts the input and formats the document.
if calling `zig build-exe` with the same file, it compiles until it tries to analyze the loop and errors with this error
```
.\file.zig:5:19: error: expected type 'type', found 'bool'
    while (true) (somevalue){};
                  ^
.\file.zig:5:29: note: referenced here
    while (true) (somevalue){};
```
when removing the semicolon at the end of the loop
```
.\file.zig:4:1: error: expected ';' or 'else', found '}'
}
```

### LINKED ISSUE #10529: Useless compile error for array literal and slice type mismatch (semantic unification fail during type analysis)
### Zig Version

0.10.0-dev.40+303bad998

### Steps to Reproduce

If I try to compile this
```
fn satisfyme(_: []const []const u8) void {}

pub fn main() void {
    satisfyme(.{});
}
```
I get
```
> zig build-exe test.zig
./test.zig:4:14: error: array literal requires address-of operator to coerce to slice type '[]const []const u8'
    satisfyme(.{});
             ^
./test.zig:4:16: note: referenced here
    satisfyme(.{});
```

...which is not useful at all.

## Common "wrong" solutions

I'm pretty sure that's an array
```
./test.zig:4:18: error: slice of non-array type 'struct:4:16'
    satisfyme(.{}[0..]);
                 ^
```

Parser error (I'm sure all of the things inside `(...)` is left-binding)
```
test.zig:4:35: error: expected ',', found '['
    satisfyme([]const []const u8{}[0..]);
                                  ^
```

That's literally a slice literal with address-of operator
```
./test.zig:4:17: error: array literal requires address-of operator to coerce to slice type '[]const []const u8'
    satisfyme(&([]const []const u8{}));
                ^
./test.zig:4:35: note: referenced here
    satisfyme(&([]const []const u8{}));
                                  ^
```

Assign to variable doesn't work either?
```
./test.zig:4:19: error: array literal requires address-of operator to coerce to slice type '[]const []const u8'
    const args = ([]const []const u8{});
                  ^
./test.zig:4:37: note: referenced here
    const args = ([]const []const u8{});
                                    ^
```

This one I kind of understand, but then I don't
```
test.zig:4:21: error: pointer modifier 'const' not allowed on array child type
    const args = [_]const[]const u8 { };
                    ^
```

## Right solutions

```
fn satisfyme(_: []const []const u8) void {}

pub fn main() void {
    satisfyme(([_][]const u8 {})[0..]);
}
```

```
fn satisfyme(_: []const []const u8) void {}

pub fn main() void {
    satisfyme(&([_][]const u8 {}));
}
```

### Expected Behavior

The error message should be clearer as to what I should do.

### Actual Behavior

It no work

Vexu:
Continuation to #10879.

Also included is better error message when casting tuple to slice/ptr to array discards const qualifier.

Closes  #416
Closes #5739
Closes  #10511
Closes  #10529

### DIFF
--- a/lib/std/zig/parse.zig
+++ b/lib/std/zig/parse.zig
@@ -2943,7 +2943,12 @@ const Parser = struct {
 
     /// WhileContinueExpr <- COLON LPAREN AssignExpr RPAREN
     fn parseWhileContinueExpr(p: *Parser) !Node.Index {
-        _ = p.eatToken(.colon) orelse return null_node;
+        _ = p.eatToken(.colon) orelse {
+            if (p.token_tags[p.tok_i] == .l_paren and
+                p.tokensOnSameLine(p.tok_i - 1, p.tok_i))
+                return p.fail(.expected_continue_expr);
+            return null_node;
+        };
         _ = try p.expectToken(.l_paren);
         const node = try p.parseAssignExpr();
         if (node == 0) return p.fail(.expected_expr_or_assignment);
--- a/lib/std/zig/parse.zig
+++ b/lib/std/zig/parse.zig
@@ -147,11 +147,6 @@ const Parser = struct {
         return result;
     }
 
-    fn warn(p: *Parser, tag: Ast.Error.Tag) error{OutOfMemory}!void {
-        @setCold(true);
-        try p.warnMsg(.{ .tag = tag, .token = p.tok_i });
-    }
-
     fn warnExpected(p: *Parser, expected_token: Token.Tag) error{OutOfMemory}!void {
         @setCold(true);
         try p.warnMsg(.{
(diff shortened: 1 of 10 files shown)


lib/std/zig/parse.zig
```diff
@@ -3590,7 +3583,7 @@ const Parser = struct {
         _ = try p.parsePtrPayload();
 
         const then_expr = try bodyParseFn(p);
-        if (then_expr == 0) return p.fail(.invalid_token);
+        assert(then_expr != 0);
```
matu3ba, line 3586: how do you rule out that code can trip this assertion?

lib/std/zig/parse.zig
```diff
@@ -3602,7 +3595,7 @@ const Parser = struct {
         });
         _ = try p.parsePayload();
         const else_expr = try bodyParseFn(p);
-        if (else_expr == 0) return p.fail(.invalid_token);
+        assert(then_expr != 0);
```
matu3ba, line 3598: again: how to rule out code can trip assertion?

matu3ba (commented):
The change looks like good usability improvement.
Can you explain, how invalid code can not trip the assertions? Maybe as comment?

lib/std/zig/parse.zig
Vexu: I changed the calls to `parseIf` to use `expectExpr` which return `error.ParseError` if no node was found.

lib/std/zig/parse.zig
Vexu: Same as above.

src/main.zig
```diff
@@ -4041,66 +4037,95 @@ fn fmtPathFile(
     }
 }
 
-fn printErrMsgToStdErr(
+fn printErrsMsgToStdErr(
```
andrewrk, line 4040: Would this logic perhaps be better moved to `Compilation.getAllErrorsAlloc` ?

andrewrk (approved):

Issue #10180, opened 2021-11-19
stage2: function params get clobbered


g-w1:
### Remember to search before filing a new report

- [X] I searched for this bug and did not find it in the issue tracker, and furthermore, the title I used above will make this new bug report turn up in the search results for my query.

### Zig Version

0.9.0-dev.1686+696e51be4

### Steps to Reproduce

compile and run this with the latest master stage2:
```zig
const plan9 = @import("std").os.plan9;
pub fn main() void {
    const file = "a_file.f";
    const str = "you got created!\n";
    const fd = create(file, .ORDWR, 0o777);
    _ = pwrite(fd, str, str.len, 0);
    _ = close(fd);
}
pub fn pwrite(fd: usize, buf: [*]const u8, count: usize, offset: usize) usize {
    return syscall4(.PWRITE, fd, @ptrToInt(buf), count, offset);
}
pub fn syscall4(sys: plan9.SYS, arg0: usize, arg1: usize, arg2: usize, arg3: usize) usize {
    _ = sys;
    _ = arg0;
    _ = arg1;
    _ = arg2;
    _ = arg3;
    if (arg1 == arg2) unreachable;

    return arg1; // len
}

pub fn create(path: [*:0]const u8, omode: enum(usize) { ORDWR }, perms: usize) usize {
    _ = path;
    _ = omode;
    _ = perms;
    return 42;
}
pub fn close(fd: usize) usize {
    if (fd != 42) unreachable;
    return 0;
}
```

### Expected Behavior

buf (arg1) and count (arg2) not to be the same in syscall4

### Actual Behavior

buf (arg1) and count (arg2) are the same in syscall4:
![image](https://user-images.githubusercontent.com/58830309/142704375-ef901794-a855-4362-a2d1-a14dd96101fc.png)

Really interestingly, this issue goes away if we replace `sys: plan9.SYS` with `sys: enum(usize) { PWRITE }`, which is basically what std.os.plan9.SYS is:
![image](https://user-images.githubusercontent.com/58830309/142704524-7fa09371-58d8-4662-ab49-998bfa37c6d0.png)

The two assembly versions and the diff between them:
```
❯ cat zig/build-release/oldd
push   rcx
push   rsi
push   rdi
push   rbp
mov    rbp,rsp
sub    rsp,0x20
mov    DWORD PTR [rbp-0x10],edi
mov    DWORD PTR [rbp-0x18],esi
mov    DWORD PTR [rbp-0x20],ecx
[...]

PR #10916, opened 2022-02-17
stage2,x64: pass all args on stack in debug and if not extern fn

kubkon:
If the function call is not extern and we are compiling in debug,
pass function params always on stack. This will improve debugging
capabilities since the params will not be volatile and possibly
clobbered by the procedure code.

### DIFF
--- a/src/arch/x86_64/CodeGen.zig
+++ b/src/arch/x86_64/CodeGen.zig
@@ -883,7 +883,8 @@ fn copyToTmpRegister(self: *Self, ty: Type, mcv: MCValue) !Register {
 /// Allocates a new register and copies `mcv` into it.
 /// `reg_owner` is the instruction that gets associated with the register in the register table.
 /// This can have a side effect of spilling instructions to the stack to free up a register.
-fn copyToNewRegister(self: *Self, reg_owner: Air.Inst.Index, ty: Type, mcv: MCValue) !MCValue {
+/// WARNING make sure that the allocated register matches the returned MCValue from an instruction!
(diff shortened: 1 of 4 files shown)


src/arch/x86_64/CodeGen.zig
```diff
@@ -4700,7 +4749,9 @@ const CallMCValues = struct {
 };
 
 /// Caller must call `CallMCValues.deinit`.
-fn resolveCallingConventionValues(self: *Self, fn_ty: Type) !CallMCValues {
+fn resolveCallingConventionValues(self: *Self, fn_ty: Type, opts: struct {
+    is_extern_fn: bool,
```
andrewrk, line 4753: You shouldn't need to care whether it's an extern function or not. The frontend should have decided the calling convention by now, and the calling convention is the only information you need to look at to determine where parameters are stored.

andrewrk (changes requested):

src/arch/x86_64/CodeGen.zig
kubkon: Oh wait, I could do that by correctly observing C ABI rather than collapsing C ABI with `unspecified`. I obviously need to pass args in registers first when making an extern function call, but I forgot we actually folded the two calling conventions. Thanks for pointing this out!

g-w1:
I think this solves https://github.com/ziglang/zig/issues/10180 (at least in debug mode and I haven't tested it yet) \o/ 🎉🎉🎉

kubkon:
> I think this solves #10180 (at least in debug mode and I haven't tested it yet) \o/ tadatadatada

Great, although this is just a happy byproduct as we pass everything on stack. If the args get clobbered this means we don't track register spilling correctly, and still needs to be investigated as in release mode or if we are issuing a C ABI call, this issue will remain.

PR #10917, opened 2022-02-17
stage2,x64: implement more slice and array instructions

kubkon:
* implement `slice_elem_ptr`
* implement `array_elem_val` when array fits in register
* implement `array_elem_val` when array is stored in memory


kubkon:
Superseded by #10916

PR #10922, opened 2022-02-18
Fix incorrect documentation for std.HashMap.remove()

chip2n:
I think this documentation was copied over from fetchRemove() which does return the removed element.


marler8997 (approved):

andrewrk:
Thanks @chip2n

PR #10927, opened 2022-02-19
stage2,x64: implement integer division

kubkon:
* add lowering for single operand `imul` and `idiv`
* implement `div_exact` for signed and unsigned integers
* add unsigned `div`
* implement the rest of builtins for integers: `div_trunc`, `div_floor`, `mod`, and `rem`
* pull out integer division tests into a standalone test unit

### DIFF
--- a/test/behavior/int_div.zig
+++ b/test/behavior/int_div.zig
@@ -1,0 +1,93 @@
+const std = @import("std");
+const builtin = @import("builtin");
+const expect = std.testing.expect;
+
+test "integer division" {
+    if (builtin.zig_backend == .stage2_aarch64) return error.SkipZigTest;
+    if (builtin.zig_backend == .stage2_c) return error.SkipZigTest;
+    if (builtin.zig_backend == .stage2_wasm) return error.SkipZigTest;
+    if (builtin.zig_backend == .stage2_arm) return error.SkipZigTest;
+
+    try testDivision();
+    comptime try testDivision();
+}
+fn testDivision() !void {
+    try expect(div(u32, 13, 3) == 4);
+    try expect(div(u64, 13, 3) == 4);
+    try expect(div(u8, 13, 3) == 4);
+
+    try expect(divExact(u32, 55, 11) == 5);
+    try expect(divExact(i32, -55, 11) == -5);
+    try expect(divExact(i64, -55, 11) == -5);
+    try expect(divExact(i16, -55, 11) == -5);
+
+    try expect(divFloor(i8, 5, 3) == 1);
+    try expect(divFloor(i16, -5, 3) == -2);
+    try expect(divFloor(i64, -0x80000000, -2) == 0x40000000);
+    try expect(divFloor(i32, 0, -0x80000000) == 0);
+    try expect(divFloor(i64, -0x40000001, 0x40000000) == -2);
+    try expect(divFloor(i32, -0x80000000, 1) == -0x80000000);
+    try expect(divFloor(i32, 10, 12) == 0);
+    try expect(divFloor(i32, -14, 12) == -2);
+    try expect(divFloor(i32, -2, 12) == -1);
+
+    try expect(divTrunc(i32, 5, 3) == 1);
+    try expect(divTrunc(i32, -5, 3) == -1);
+    try expect(divTrunc(i32, 9, -10) == 0);
+    try expect(divTrunc(i32, -9, 10) == 0);
+    try expect(divTrunc(i32, 10, 12) == 0);
+    try expect(divTrunc(i32, -14, 12) == -1);
+    try expect(divTrunc(i32, -2, 12) == 0);
+
+    try expect(mod(u32, 10, 12) == 10);
+    try expect(mod(i32, 10, 12) == 10);
+    try expect(mod(i64, -14, 12) == 10);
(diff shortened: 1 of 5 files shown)


test/behavior/int_div.zig
```diff
@@ -0,0 +1,93 @@
...
+test "integer division" {
+    if (builtin.zig_backend == .stage2_aarch64) return error.SkipZigTest;
+    if (builtin.zig_backend == .stage2_c) return error.SkipZigTest;
+    if (builtin.zig_backend == .stage2_wasm) return error.SkipZigTest;
+    if (builtin.zig_backend == .stage2_arm) return error.SkipZigTest;
+
+    try testDivision();
+    comptime try testDivision();
+}
+fn testDivision() !void {
```
matu3ba, line 14: potential edge cases: 
`0/1`,`-0/1`, `1/1`,`-1/-1`
`max/max`,`min/min`,`max/min`,`min/max`

if it is possible, add a TODO for `0/0`,`anynumber/0`

test/behavior/int_div.zig
```diff
@@ -0,0 +1,93 @@
...
+    try expect(div(u32, 13, 3) == 4);
+    try expect(div(u64, 13, 3) == 4);
+    try expect(div(u8, 13, 3) == 4);
```
matu3ba, line 17: maybe add here the complete sequence `15,14,13,12` each divided by 3 with expected result?

test/behavior/int_div.zig
```diff
@@ -0,0 +1,93 @@
...
+
+    try expect(divExact(u32, 55, 11) == 5);
+    try expect(divExact(i32, -55, 11) == -5);
+    try expect(divExact(i64, -55, 11) == -5);
+    try expect(divExact(i16, -55, 11) == -5);
```
matu3ba, line 22: what about `55/-11` and the other combinations?

test/behavior/int_div.zig
```diff
@@ -0,0 +1,93 @@
...
+
+    try expect(divFloor(i8, 5, 3) == 1);
+    try expect(divFloor(i16, -5, 3) == -2);
+    try expect(divFloor(i64, -0x80000000, -2) == 0x40000000);
+    try expect(divFloor(i32, 0, -0x80000000) == 0);
+    try expect(divFloor(i64, -0x40000001, 0x40000000) == -2);
+    try expect(divFloor(i32, -0x80000000, 1) == -0x80000000);
```
matu3ba, line 29: I get `int(0x80000000)=2147483648` and `2**31=2147483648`.
I might be wrong, but `-(-2147483648)=2147483648 > i32max = 2147483647 = 2^{31}-1` (overflow)
Can you use `minInt` from `int.zig` or annotate as comment which number is meant?
`-hex` is very non-trivial to read (and I hate std.fmt to give such an output).

test/behavior/int_div.zig
```diff
@@ -0,0 +1,93 @@
...
+    try expect(divFloor(i32, 10, 12) == 0);
+    try expect(divFloor(i32, -14, 12) == -2);
+    try expect(divFloor(i32, -2, 12) == -1);
```
matu3ba, line 32: `-2/-12` ? What about edge case `0/12`,`0/-12`,`-0/12`,`-0/12` ?

matu3ba (changes requested):
edge cases for tests would be great and clarification of potential overflow in test

would be nice to add combinations of `+-0`,`+-min,+-max` and values in between +-[0..max/min] with testing the respective features (trunc,floor) if existing.
Same testing strategy can be applied to remainder.

Probably useful to add as comment would be to mention `cld(x,y) = -fld(-x,y)` somewhere inside `int.zig`. https://github.com/JuliaLang/julia/issues/7132

test/behavior/int_div.zig
kubkon: Overflow guards are currently not there and need Sema improvements. If you are asking if these are correct, I did simply pull them from `behavior/math.zig` tests. Also, they don't overflow with stage1.

kubkon:
> edge cases for tests would be great and clarification of potential overflow in test
> 
> would be nice to add combinations of `+-0`,`+-min,+-max` and values in between +-[0..max/min] with testing the respective features (trunc,floor) if existing. Same testing strategy can be applied to remainder.
> 
> Probably useful to add as comment would be to mention `cld(x,y) = -fld(-x,y)` somewhere inside `int.zig`. [JuliaLang/julia#7132](https://github.com/JuliaLang/julia/issues/7132)

Agreed but we should probably add these to `behavior/math.zig` test. I actually pulled this test from there pretty much verbatim. I am happy to drop some TODOs so that we don't forget to add more complete tests in the future.

test/behavior/int_div.zig
kubkon: Testing for ReleaseSafe is pointless at this stage (for native stage2 backends) as we don't have safety support of Sema implemented yet.

kubkon: Also, any UB checks will need to be postponed too until we get some safety guards emitted by Sema into codegen.

kubkon:
Actually, @matu3ba would you mind creating an issue with your comments from here please? It's much easier to track it on GH then TODOs in the code.

matu3ba:
@kubkon done in #10930.

PR #10928, opened 2022-02-19
added 'negation' and 'mulitply by i' for complex numbers

BlueAlmost:
Added methods to the complex number structure:

1) negation - simply negate the real part, and negate the imaginary part.  This is needed for stringing together mathematical operations without invoking a complex multiply (which may be costly).

2) mulbyi - multiplies a complex number by i=sqrt(-1).  BUT, this is done without invoking a complex multiply (which may be costly with its 4 required operations).  Effectively, this "rotates" any number on the complex plane by 90 degrees.  The operation is found often (such as in FFT routines).  Can be common in inner loops.  

NOTE A: I think that the name "mulbyi" is not optimal, but other names I had considered (muli, rot90, imul, imult) all were a bit underwhelming.

NOTE B: It understood that one may create their own complex struct and implement these additions themselves... but splintering off the standard complex struct for these basic, and useful mathematical operations is not attractive.

### DIFF
--- a/lib/std/math/complex.zig
+++ b/lib/std/math/complex.zig
@@ -166,7 +166,7 @@ test "complex.neg" {
     const a = Complex(f32).init(5, 3);
     const c = a.neg();
 
-    try testing.expect(c.re == -5 and c.im == 3);
+    try testing.expect(c.re == -5 and c.im == -3);
 }
 
 test "complex.mulbyi" {
--- a/lib/std/math/complex.zig
+++ b/lib/std/math/complex.zig
@@ -103,7 +103,7 @@ pub fn Complex(comptime T: type) type {
                 .im = self.re,
             };
         }
-        
+
         /// Returns the reciprocal of a complex number.
         pub fn reciprocal(self: Self) Self {
             const m = self.re * self.re + self.im * self.im;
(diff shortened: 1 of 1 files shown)


rohlem:
Minor nitpick: [The documentation's style guide recommends camel cased function names](https://ziglang.org/documentation/master/#Names) (unless they return a type).
That would lead to `mulByI` instead of `mulbyi`.

BlueAlmost:
> Minor nitpick: [The documentation's style guide recommends camel cased function names](https://ziglang.org/documentation/master/#Names) (unless they return a type). That would lead to `mulByI` instead of `mulbyi`.

You raise a good point.

To be honest, I am still interested in finding a better name for this, mulbyi is not all that good.  And "mulByI" just looks ugly to me personally and is visually confusing (although it would agree with the style guide).  

Multiply by "i", is quite a common operation, but does not have a name (as far as I know).  Mathematically, it truly _**is**_ a multiply.  But computationally, it does not make sense to do the 4 operation complex multiply literally.  Hence, this simple (and efficient) method is being proposed.

I considered: 

muli, imul, (too close to the standard "mul")
rot90, rotpi2  (a bit arcane perhaps)

maybe "itimes"?

PR #10929, opened 2022-02-19
stage2: C backend: fix codegen for field_ptr/elem_ptr values

davidmrt98:
Codegen for `field_ptr` and `elem_ptr` resulted in invalid C programs with misplaced referencing/dereferencing operators. This is fixed by this pull request. As a result, a few more behavior tests are passing now.

Additionally, I changed the order of parameters in `renderDeclName` so it is consistent with the other `render` functions.

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -464,7 +464,7 @@ pub const DeclGen = struct {
                     .variable => ptr_val.castTag(.variable).?.data.owner_decl,
                     else => unreachable,
                 };
-                try dg.renderDeclValue(writer, decl.ty, ptr_val, decl);
+                try dg.renderDeclName(decl, writer);
                 return decl.ty;
             },
             .field_ptr => {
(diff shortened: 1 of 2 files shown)


src/codegen/c.zig
```diff
@@ -464,7 +464,7 @@ pub const DeclGen = struct {
                     .variable => ptr_val.castTag(.variable).?.data.owner_decl,
                     else => unreachable,
                 };
-                try dg.renderDeclValue(writer, decl.ty, ptr_val, decl);
+                try dg.renderDeclName(writer, decl);
                 return decl.ty;
             },
             .field_ptr => {
```
topolarity, line 470: I think I'm the cause of this bug, so thanks for fixing this 🙂 

It seems like `.field_ptr` would need to a dereference too, right?

davidmrt98: You're right, thanks for pointing that out. The PR was merged before I noticed your comment, so I'm just going to open a new PR with the fix :+1:

Vexu: > The PR was merged before I noticed your comment

Sorry about that, I misread topolarity's comment.

PR #10937, opened 2022-02-20
translate-c: add missing builtins used by CRuby

FnControlOption:
I added these builtins used in the [CRuby header files](https://github.com/ruby/ruby/tree/master/include):

- `__has_builtin` - Always returns true because unimplemented builtins give an "undefined identifier" error before it can be called.
- `__builtin_assume`
- `__builtin_unreachable`
- `__builtin_constant_p` - With a C compiler, this returns a non-zero value (`true`) if the given expression is known to be constant at compile time. AFAICT it's not possible to check for this in Zig, so we just return `false` (zero).
- `__builtin_mul_overflow`

### DIFF
--- a/lib/std/zig/c_builtins.zig
+++ b/lib/std/zig/c_builtins.zig
@@ -245,31 +245,8 @@ pub inline fn __builtin_constant_p(expr: anytype) c_int {
     _ = expr;
     return @boolToInt(false);
 }
-
-fn multiplyAs(a: anytype, b: anytype, comptime T: type) !T {
-    const allocator = std.heap.c_allocator;
-    var big_a = try std.math.big.int.Managed.initSet(allocator, a);
-    defer big_a.deinit();
-    var big_b = try std.math.big.int.Managed.initSet(allocator, b);
-    defer big_b.deinit();
-    var product = try std.math.big.int.Managed.init(allocator);
-    defer product.deinit();
-    try product.mul(big_a.toConst(), big_b.toConst());
-    return try product.to(T);
-}
-
-pub fn __builtin_mul_overflow(a: anytype, b: anytype, result: anytype) c_int {
-    const ResultType = @TypeOf(result);
-    const result_type_info = @typeInfo(ResultType);
-    if (result_type_info != .Pointer) {
-        @compileError("Expected pointer, found " ++ @typeName(ResultType));
-    }
-    if (multiplyAs(a, b, result_type_info.Pointer.child)) |product| {
-        result.* = product;
-        return @boolToInt(false);
-    } else |_| {
-        return @boolToInt(true);
-    }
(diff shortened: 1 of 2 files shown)


lib/std/zig/c_builtins.zig
```diff
@@ -228,6 +228,50 @@ pub inline fn __builtin_isinf_sign(x: anytype) c_int {
...
+    const result_type_info = @typeInfo(ResultType);
+    if (result_type_info != .Pointer) {
+        @compileError("Expected pointer, found " ++ @typeName(ResultType));
+    }
+    if (multiplyAs(a, b, result_type_info.Pointer.child)) |product| {
+        result.* = product;
+        return @boolToInt(false);
+    } else |_| {
+        return @boolToInt(true);
+    }
```
Vexu, line 272:
```suggestion
pub fn __builtin_mul_overflow(a: anytype, b: anytype, result: *@TypeOf(a, b)) c_int {
    return @mulWithOverflow(@TypeOf(a, b), a, b, result);
```

FnControlOption: `a`, `b`, and `result` can have different types:

- https://clang.llvm.org/docs/LanguageExtensions.html#checked-arithmetic-builtins
- https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html

ehaas:
As it's currently implemented here, `__builtin_mul_overflow` will have different behavior than clang's on overflow, since it won't set result to `the unique value that is equivalent to the mathematically-correct result modulo two raised to the k power, where k is the number of bits in the result type`

I think `@mulWithOverflow` will give us the behavior we need, which has the added benefit of not requiring an allocator. You can use the multi-argument form of `@TypeOf` to determine the first argument to `@mulWithOverflow`

lib/std/zig/c_builtins.zig
```diff
@@ -228,6 +228,27 @@ pub inline fn __builtin_isinf_sign(x: anytype) c_int {
...
+
+pub inline fn __builtin_unreachable() noreturn {
+    unreachable;
+}
+
+pub inline fn __builtin_constant_p(expr: anytype) c_int {
+    _ = expr;
+    return @boolToInt(false);
+}
+pub fn __builtin_mul_overflow(a: anytype, b: anytype, result: *@TypeOf(a, b)) c_int {
```
ehaas, line 248: `result` could be a pointer to a type that is distinct from `a` and `b` - for example  `a` and `b` could be `int` and `result` could be a `char *`. When it's smaller than `@TypeOf(a,b)` you'll need to truncate the result.

ehaas:
I would also suggest adding a test in `test/run_translated_c.zig` that exercises `__builtin_mul_overflow` with various sizes of integers and checks that `result` has the expected value.

Issue #9484, opened 2021-07-29
eliminate the BoundFn type from the language and the compiler implementations


andrewrk:
Currently the Zig type system has a "BoundFn" which can be observed thusly:

```zig
test "example" {
    var a: A = .{};
    @compileLog(@TypeOf(a.b));
}

const A = struct {
    fn b() void {}
};
```

This type is to be removed from the language. Instead, method call syntax will be explicitly recognized in the parser. Here's an example which currently passes but will become a compile error:

```zig
test "example" {
    var a1: A = .{};
    var a2: A = .{};
    (if (true) a1.b else a2.b)();
}

const A = struct {
    fn b(a: A) void {
        _ = a;
    }
};
```

After this change, this will give a compile error: struct `A` has no field named `b`.

Related: #7942

Here's a diff to get you started:

```diff
diff --git a/lib/std/builtin.zig b/lib/std/builtin.zig
index 9d432a3a0..f4daf5354 100644
--- a/lib/std/builtin.zig
+++ b/lib/std/builtin.zig
@@ -204,7 +204,6 @@ pub const TypeInfo = union(enum) {
     Enum: Enum,
     Union: Union,
     Fn: Fn,
-    BoundFn: Fn,
     Opaque: Opaque,
     Frame: Frame,
     AnyFrame: AnyFrame,
diff --git a/src/stage1/all_types.hpp b/src/stage1/all_types.hpp
index dfebc66cf..d36396c68 100644
--- a/src/stage1/all_types.hpp
+++ b/src/stage1/all_types.hpp
@@ -310,12 +310,6 @@ struct ConstErrValue {
     ZigValue *payload;
 };
 
-struct ConstBoundFnValue {
-    ZigFn *fn;
-    Stage1AirInst *first_arg;
-    AstNode *first_arg_src;
-};
-
 struct ConstArgTuple {
     size_t start_index;
     size_t end_index;
@@ -510,7 +504,6 @@ struct ZigValue {
         double x_f64;
         float128_t x_f128;
         bool x_bool;
-        ConstBoundFnValue x_bound_fn;
         ZigType *x_type;
         ZigValue *x_optional;
         ConstErrValue x_err_union;
@@ -1492,10 +1485,6 @@ struct ZigTypeFn {
     ZigType *bound_fn_parent;
 };
 
-struct ZigTypeBoundFn {
-    ZigType *fn_type;
-};
-
 // Needs to have the same memory layout as ZigTypeArray
[...]

PR #10940, opened 2022-02-20
[std.ArrayList] return explicit errors

motiejus:
All errors from std.ArrayList are allocation errors. Mark them as such.
This is helpful when one wants to obtain a function pointer.

However, I was not able to get a "simple" signature: help would be appreciated
on how to type it:

```
$ zig test array_list.zig
./array_list.zig:1451:41: error: expected type 'fn(u8) std.mem.Allocator.Error!void', found '(bound fn(*ArrayListAligned(u8,null), u8) std.mem.Allocator.Error!void)'
    const appendFn: appendProto = buffer.append;
                                        ^
```

### DIFF
--- a/lib/std/array_list.zig
+++ b/lib/std/array_list.zig
@@ -58,7 +58,7 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
         /// Initialize with capacity to hold at least `num` elements.
         /// The resulting capacity is likely to be equal to `num`.
         /// Deinitialize with `deinit` or use `toOwnedSlice`.
-        pub fn initCapacity(allocator: Allocator, num: usize) !Self {
+        pub fn initCapacity(allocator: Allocator, num: usize) Allocator.Error!Self {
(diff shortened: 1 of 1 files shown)


Vexu:
The test is failing because you're accessing the function from an instance of the type instead of the type directly (see #9484).

But this change isn't even needed, Zig is able to cast a function with an inferred error set just fine.
```zig
test {
    var list = std.ArrayList(u8).init(std.testing.allocator);
    defer list.deinit();
    var a: fn(*std.ArrayList(u8), u8) std.mem.Allocator.Error!void = @TypeOf(list).append;
    try a(&list, 1);
    try testing.expect(list.items.len == 1);
}
```

motiejus:
> But this change isn't even needed, Zig is able to cast a function with an inferred error set just fine.

I had a slightly different use case in mind: I wanted to pass a function pointer which matches the signature, but does not depend on the underlying implementation. But just realized I can do so with a simple closure.

Updated the PR to only match the types.

PR #10941, opened 2022-02-20
std.mem.bytesAsValue for byte-aligned packed structs

motiejus:
Currently `bytesAsValue` works with only word-aligned structs; technically, when `@sizeOf(T) == @divExact(@bitSizeOf(T), 8)`.

This diff enables converting bytes to non-word-aligned struct pointers. Though still requires them to be byte-aligned.

`asBytes` is not so easy; I will file a separate task for it.

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -2610,7 +2610,7 @@ fn AsBytesReturnType(comptime P: type) type {
     if (!trait.isSingleItemPtr(P))
         @compileError("expected single item pointer, passed " ++ @typeName(P));
 
-    const size = @divExact(@bitSizeOf(meta.Child(P)), 8);
+    const size = @sizeOf(meta.Child(P));
 
     return CopyPtrAttrs(P, .One, [size]u8);
 }
@@ -2758,15 +2758,11 @@ test "bytesAsValue preserves pointer attributes" {
     try testing.expectEqual(in.alignment, out.alignment);
 }
 
-test "bytesAsValue and asBytes work with non-word-aligned structs" {
-    const Padded = packed struct { a: u12, b: u12 };
+test "bytesAsValue works with non-word-aligned structs" {
+    const padded = packed struct { a: u14, b: u10 };
     // for test to be meaningful @sizeOf must think the struct is padded.
-    assert(@sizeOf(Padded) != @divExact(@bitSizeOf(Padded), 8));
-    const padded = Padded{ .a = 42, .b = 1 };
-    const bytes = asBytes(&padded);
-    try testing.expectEqual(bytes.len, 3);
-    const padded2 = bytesAsValue(Padded, bytes[0..bytes.len]);
-    try testing.expectEqual(padded, padded2.*);
+    assert(@sizeOf(padded) != @divExact(@bitSizeOf(padded), 8));
+    _ = bytesAsValue(padded, &[3]u8{ 0, 0, 0 });
 }
 
 /// Given a pointer to an array of bytes, returns a value of the specified type backed by a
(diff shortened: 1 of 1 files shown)


motiejus:
re. drone-ci failures: this command seems to have worked for me:

```
$ ./build/zig build -Dtarget=arm-linux-musleabihf
<...>
$ file zig-out/bin/zig
zig-out/bin/zig: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), statically linked, with debug_info, not stripped
$ 
```

Vexu:
The test you added passes without changing the `@sizeOf`, do you have a test case that doesn't currently pass?

> re. drone-ci failures: this command seems to have worked for me:

It is caused by a broken commit in master, don't know why the CI is using master instead of your branch.

motiejus:
> The test you added passes without changing the `@sizeOf`, do you have a test case that doesn't currently pass?

This is a mixup, thanks for noticing. Will do so likely tomorrow.

motiejus:
> > The test you added passes without changing the `@sizeOf`, do you have a test case that doesn't currently pass?
> 
> This is a mixup, thanks for noticing. Will do so likely tomorrow.

Updated. It wasn't easy to find a packed struct that satisfies the assertion; I would love to learn more about ([in `@sizeOf` docs](https://ziglang.org/documentation/master/#sizeOf)) though:

```
This size may contain padding bytes. 
```

motiejus:
Changing `asBytes` to assume byte-aligned data fails to compile stage1, as there are some non-byte-aligned structs. So `asBytes` cannot be changed as-is. I also spent an hour or so mirroring the change to `asBytes`, to no avail.

I will leave this PR to `bytesAsValue` alone and file an issue for `asBytes` to discuss it further.

lib/std/mem.zig
```diff
@@ -2758,6 +2758,13 @@ test "bytesAsValue preserves pointer attributes" {
     try testing.expectEqual(in.alignment, out.alignment);
 }
 
+test "bytesAsValue works with non-word-aligned structs" {
```
andrewrk, line 2761: Please use a different word than "word" which is ambiguous.

lib/std/mem.zig
```diff
@@ -2758,6 +2758,13 @@ test "bytesAsValue preserves pointer attributes" {
...
+    const padded = packed struct { a: u14, b: u10 };
+    // for test to be meaningful @sizeOf must think the struct is padded.
+    assert(@sizeOf(padded) != @divExact(@bitSizeOf(padded), 8));
```
andrewrk, line 2764: This assertion trips on latest master:

```
$ ./stage1/bin/zig test test.zig 
Test [1/1] test "example"... thread 2925231 panic: reached unreachable code
/home/andy/Downloads/zig/lib/std/debug.zig:260:14: 0x206aab in std.debug.assert (test)
             ^
/home/andy/Downloads/zig/build-release/test.zig:7:11: 0x2076de in test "example" (test)
    assert(@sizeOf(padded) != @divExact(@bitSizeOf(padded), 8));
          ^
```

lib/std/mem.zig
```diff
@@ -2758,6 +2758,13 @@ test "bytesAsValue preserves pointer attributes" {
...
+    _ = bytesAsValue(padded, &[3]u8{ 0, 0, 0 });
```
andrewrk, line 2765: The test should write through the pointer and then observe an expected outcome

andrewrk (changes requested):

motiejus:
I think I no longer want to do this, because:
- we cannot make a symmetric change to `asBytes` (most importantly).
- this is only for stage1 (at the time I did not test stage2).
- I am adding the assertion to my packed structs: `@sizeOf(S) * 8 == @bitSizeOf(S)` and life is easy.

PR #10948, opened 2022-02-20
Support -Wl,--soname

sfackler:
Some projects ([elfutils](https://sourceware.org/elfutils/) in my case) pass `-Wl,--soname` rather than `-Wl,-soname` to the C compiler. `--soname` appears to be supported by GNU ld but undocumented, so I've just updated the linker argument processing logic to handle `-soname` and `--soname` the same way.

There were also a couple of other linker args that weren't supported but didn't seem to negatively affect the final output:
* `-rpath-link`
* `--build-id`
* `--no-undefined`
* `--enable-new-dtags`

I may circle back and add support for those as well if there is interest.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1527,7 +1527,9 @@ fn buildOutputType(
             var i: usize = 0;
             while (i < linker_args.items.len) : (i += 1) {
                 const arg = linker_args.items[i];
-                if (mem.eql(u8, arg, "-soname")) {
+                if (mem.eql(u8, arg, "-soname") or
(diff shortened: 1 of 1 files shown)


andrewrk:
Thanks!

vazub:
@sfackler 
> I may circle back and add support for those as well if there is interest.
Would you be so kind as to add support for those remaining as well? (except for --no-undefined, it has been added already).

Specifically looking at -rpath-link, as it is required to build libdrm and maybe others with zig cc.

PR #10949, opened 2022-02-20
stage2,x64: print test_runner results

kubkon:
I really wanna see the test results printed to `stderr`. I really do.

TODO:
- [x] `airSwitch`
- [x] return values bigger than register size
- [x] lower `error_set` and `error_union`
- [x] lower `repeated` and `empty_with_sentinel` array types
- [x] `airMin`
- [x] `airShl`
- [x] `airUnwrapErrPayload` and `airUnwrapErrErr`
- [x]  `airWrapErrPayload` and `airWrapErrErr`

### DIFF
--- a/test/behavior/src.zig
+++ b/test/behavior/src.zig
@@ -1,7 +1,10 @@
 const std = @import("std");
+const builtin = @import("builtin");
 const expect = std.testing.expect;
 
 test "@src" {
+    if (builtin.zig_backend != .stage1) return error.SkipZigTest;
+
     try doTheTest();
 }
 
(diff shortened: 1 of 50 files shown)


test/behavior/src.zig
```diff
@@ -3,9 +3,10 @@ const builtin = @import("builtin");
 const expect = std.testing.expect;
 
 test "@src" {
-    if (builtin.zig_backend != .stage1) return error.SkipZigTest;
+    // TODO why is this failing on stage1?
```
Vexu, line 6: It's because you changed the line numbers.

kubkon: Mhm, makes sense. I'll push a fix later tomorrow morning then unless you beat me to it. Thanks for pointing this one out Vexu!

Issue #10693, opened 2022-01-25
zig build returns error.SystemResources on MacOS when build-lib command exceeds ARG_MAX


fubark:
### Zig Version

0.9.0

### Steps to Reproduce

Create build.zig with a static lib step that adds many c source files with lib.addCSourceFile (and many c flags).
run: zig build

### Expected Behavior

zig build should work and not have this limit or at the very least return a more useful error.

### Actual Behavior

Prints error.SystemResources and quits.

Since zig build creates a build-lib command the args list can dramatically go up with many c source files/flags. 
On linux: getconf ARG_MAX returns 2097152 so I did not run into this issue before but on MacOS this number is much lower: 262144.

fubark:
This is even worse on Windows where CreateProcessW has a max char limit of 32,766.

The error returned was useful though:
error.Unexpected: GetLastError(206): The filename or extension is too long.

Maybe zig build-lib should also be able to read input from a generated file instead of relying on args?

Vexu:
We're getting response file support directly in Zig soon #9654, so we could also use it for `build-lib`, `build-obj`, `build-exe` and `run`.

andrewrk:
Reopened as the improvement introduced a regression and was reverted in 5ab5e2e6731a9f1198df6c53134545ccc6a6bbd3.

andrewrk:
Fixed by #11065

PR #10950, opened 2022-02-20
Do not fail to build if 'zig build-lib' etc. arguments exceed OS limits

slimsag:
These patches fix a major issue which prevents using Zig's build system from being viable in larger projects with lots of source files on Windows (and macOS + Linux, to a lesser degree.)

In [Mach engine](https://github.com/hexops/mach) we're seeing command line arguments to `zig build-lib` exceed the 32 KiB limit that Windows imposes in the `CreateProcess` API, due to the number of source files and compiler flags we need in order to build WebGPU/Dawn.

I've fixed this issue in three steps (best reviewed as individual commits):

1. Refactored `zig build-lib [args]` args handling to use an iterator, in anticipation of response file support.
2. Added response file support to `zig build-lib @args.rsp` (and `build-exe`, etc. friends) so you can pass them a text file of arguments instead.
3. Changed the `Builder` to check if the arguments it would pass to any `zig` build command are >30 KiB and, if so, use e.g. a `zig build-lib @zig-cache/args/<SHA2 of args>` response file instead.

Fixes https://github.com/ziglang/zig/issues/10693
Fixes https://github.com/hexops/mach/issues/167

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -17,6 +17,7 @@ const fmt_lib = std.fmt;
 const File = std.fs.File;
 const CrossTarget = std.zig.CrossTarget;
 const NativeTargetInfo = std.zig.system.NativeTargetInfo;
+const Sha256 = std.crypto.hash.sha2.Sha256;
 
 pub const FmtStep = @import("build/FmtStep.zig");
 pub const TranslateCStep = @import("build/TranslateCStep.zig");
@@ -2888,6 +2889,48 @@ pub const LibExeObjStep = struct {
 
         try zig_args.append("--enable-cache");
 
+        const is_build = switch (self.kind) {
+            .lib => true,
+            .exe => true,
+            .obj => true,
+            else => false,
+        };
+        if (is_build) {
+            // Windows has an argument length limit of 32,766 characters, macOS 262,144 and Linux
+            // 2,097,152. If our args exceed 30 KiB, we instead write them to a "response file" and
+            // pass that to zig, e.g. via 'zig build-lib @args.rsp'
+            var args_length: usize = 0;
+            for (zig_args.items) |arg| {
+                args_length += arg.len;
+            }
+            if (args_length >= 30 * 1024) {
+                const args_dir = try fs.path.join(
+                    builder.allocator,
+                    &[_][]const u8{ builder.pathFromRoot("zig-cache"), "args" },
+                );
+                try std.fs.cwd().makePath(args_dir);
+
+                // Write the args to zig-cache/args/<SHA256 hash of args> to avoid conflicts with
+                // other zig build commands running in parallel.
+                const partially_quoted = try std.mem.join(builder.allocator, "\" \"", zig_args.items[2..]);
+                const args = try std.mem.concat(builder.allocator, u8, &[_][]const u8{ "\"", partially_quoted, "\"" });
+
+                var args_hash: [Sha256.digest_length]u8 = undefined;
+                Sha256.hash(args, &args_hash, .{});
+                var args_hex_hash: [Sha256.digest_length * 2]u8 = undefined;
+                _ = try std.fmt.bufPrint(
+                    &args_hex_hash,
+                    "{s}",
+                    .{std.fmt.fmtSliceHexLower(&args_hash)},
+                );
+
+                const args_file = try fs.path.join(builder.allocator, &[_][]const u8{ args_dir, args_hex_hash[0..] });
+                try std.fs.cwd().writeFile(args_file, args);
+
+                zig_args.shrinkRetainingCapacity(2);
+                try zig_args.append(try std.mem.concat(builder.allocator, u8, &[_][]const u8{ "@", args_file }));
+            }
+        }
         const output_dir_nl = try builder.execFromStep(zig_args.items, &self.step);
         const build_output_dir = mem.trimRight(u8, output_dir_nl, "\r\n");
 
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -2901,7 +2901,7 @@ pub const LibExeObjStep = struct {
             // pass that to zig, e.g. via 'zig build-lib @args.rsp'
             var args_length: usize = 0;
             for (zig_args.items) |arg| {
-                args_length += arg.len;
+                args_length += arg.len + 1; // +1 to account for null terminator
             }
             if (args_length >= 30 * 1024) {
                 const args_dir = try fs.path.join(
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -2889,48 +2889,41 @@ pub const LibExeObjStep = struct {
 
         try zig_args.append("--enable-cache");
 
-        const is_build = switch (self.kind) {
-            .lib => true,
-            .exe => true,
-            .obj => true,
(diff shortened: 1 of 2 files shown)


kubkon:
`CreateProcess` is a Win32 function, correct? Perhaps you'd be willing to investigate if using `NT` directly would not impose such a limit in Windows?

src/main.zig
```diff
@@ -765,85 +765,102 @@ fn buildOutputType(
...
+                        if (next_arg == null) fatal("expected [auto|on|off] after --color", .{});
+                        color = std.meta.stringToEnum(Color, next_arg.?) orelse {
+                            fatal("expected [auto|on|off] after --color, found '{s}'", .{next_arg.?});
                         };
                     } else if (mem.eql(u8, arg, "--subsystem")) {
-                        if (i + 1 >= args.len) fatal("expected parameter after {s}", .{arg});
-                        i += 1;
-                        if (mem.eql(u8, args[i], "console")) {
+                        const next_arg = getNextArg(&process_args_iter, &resp_file_args_iter);
+                        if (next_arg == null) fatal("expected parameter after {s}", .{arg});
```
ifreund, line 844: You could get rid of all the `.?`s below by using orelse here:
```suggestion
                        const next_arg = getNextArg(&process_args_iter, &resp_file_args_iter) orelse {
                            fatal("expected parameter after {s}", .{arg});
                        };
```
One could arguably use `orelse` in many other places in your patch as well, but this is the place where I see the biggest improvement.

src/main.zig
```diff
@@ -765,85 +765,102 @@ fn buildOutputType(
...
-            var i: usize = 0;
-            args_loop: while (i < args.len) : (i += 1) {
-                const arg = args[i];
-                if (mem.startsWith(u8, arg, "-")) {
+            var process_args_iter = try process.argsWithAllocator(arena);
+            _ = process_args_iter.skip(); // "zig"
+            _ = process_args_iter.skip(); // e.g. "build-lib"
+            var resp_file_args_iter: ?ArgIteratorResponseFile = null;
+
+            var getNextArg = (struct {
```
ifreund, line 773:
```suggestion
            const getNextArg = (struct {
```

slimsag:
@kubkon 

> Perhaps you'd be willing to investigate if using NT directly would not impose such a limit in Windows?

I was not able to find such an API, maybe it'd be something undocumented?

In any case, I should note that macOS and Linux also have limitations here (256 KiB and 256 MiB, respectively): I think it would be advisable to have response file support like this in place, as clang and others use the same approach to solve this issue.

lib/std/build.zig
```diff
@@ -2888,6 +2889,48 @@ pub const LibExeObjStep = struct {
...
+            .obj => true,
+            else => false,
+        };
+        if (is_build) {
+            // Windows has an argument length limit of 32,766 characters, macOS 262,144 and Linux
+            // 2,097,152. If our args exceed 30 KiB, we instead write them to a "response file" and
+            // pass that to zig, e.g. via 'zig build-lib @args.rsp'
+            var args_length: usize = 0;
+            for (zig_args.items) |arg| {
+                args_length += arg.len;
```
Vexu, line 2904: `+ 1` for the null terminator.

lib/std/build.zig
```diff
@@ -2888,6 +2889,48 @@ pub const LibExeObjStep = struct {
 
         try zig_args.append("--enable-cache");
 
+        const is_build = switch (self.kind) {
+            .lib => true,
+            .exe => true,
+            .obj => true,
+            else => false,
+        };
+        if (is_build) {
```
Vexu, line 2898: Why would this not apply to `test`?

src/main.zig
```diff
@@ -765,85 +765,102 @@ fn buildOutputType(
...
 
             soname = .yes_default_value;
-            const args = all_args[2..];
-            var i: usize = 0;
-            args_loop: while (i < args.len) : (i += 1) {
-                const arg = args[i];
-                if (mem.startsWith(u8, arg, "-")) {
+            var process_args_iter = try process.argsWithAllocator(arena);
+            _ = process_args_iter.skip(); // "zig"
+            _ = process_args_iter.skip(); // e.g. "build-lib"
```
Vexu, line 770: No need to reallocate all the arguments just to get an iterator, it's easy enough to make your own:
```zig
const Iterator = struct {
    resp_file: ?ArgIteratorResponseFile = null,
    args: []const []const u8,
    i: usize = 0,
    fn next(it: *Iterator) ?[]const u8 {
        if (it >= it.args.len) {
            if (it.resp_file) |*resp| return resp.next();
            return null;
        }
        defer it.i += 1;
        return it.args[it.i];
    }
};
```

Vexu (changes requested):

lib/std/build.zig
slimsag: Ah, good point, it should. Not sure why I did that.

src/main.zig
slimsag: Ah, this is great, thanks so much for this!

slimsag:
Addressed all the feedback :)

src/main.zig
```diff
@@ -765,85 +765,106 @@ fn buildOutputType(
...
                         cur_pkg = cur_pkg.parent orelse
                             fatal("encountered --pkg-end with no matching --pkg-begin", .{});
                     } else if (mem.eql(u8, arg, "--main-pkg-path")) {
-                        if (i + 1 >= args.len) fatal("expected parameter after {s}", .{arg});
-                        i += 1;
-                        main_pkg_path = args[i];
+                        const next_arg = args_iter.next() orelse {
+                            fatal("expected parameter after {s}", .{arg});
+                        };
+                        main_pkg_path = next_arg;
```
Vexu, line 828:
```suggestion
                        main_pkg_path = args_iter.next() orelse {
                            fatal("expected parameter after {s}", .{arg});
                        };
```

Style nit but I'd prefer these to be assigned directly.

Vexu (approved):

slimsag:
friendly ping :) I have some quite cool tech demos that will be blocked on this fix landing

andrewrk:
Thanks for the fix! Looks like this introduce a regression in `zig run`:

```
+ /Users/runner/work/1/s/build.host/release/bin/zig run ../doc/docgen.zig -- /Users/runner/work/1/s/build.host/release/bin/zig ../doc/langref.html.in release/docs/langref.html
thread 135918 panic: expected output arg
/Users/runner/work/1/s/doc/docgen.zig:36:49: 0x100c16d2f in main (docgen)
    const out_file_name = args_it.next() orelse @panic("expected output arg");
                                                ^
/Users/runner/work/1/s/build.host/release/lib/zig/std/start.zig:561:37: 0x100c88098 in std.start.callMain (docgen)
                                    ^
/Users/runner/work/1/s/build.host/release/lib/zig/std/start.zig:495:12: 0x100c27577 in std.start.callMainWithArgs (docgen)
           ^
/Users/runner/work/1/s/build.host/release/lib/zig/std/start.zig:460:12: 0x100c274b5 in std.start.main (docgen)
           ^
```

andrewrk:
Reverted in 5ab5e2e6731a9f1198df6c53134545ccc6a6bbd3

PR #10952, opened 2022-02-20
std: clarify that os.getenv unsupported on Windows

### LINKED ISSUE #8456: Windows std.os.getenv(): lld-link: error: undefined symbol: environ
I have an executable that I'm try to build linked to libc on Windows. If I try to use `std.os.getenv` I get this linker error:

```
lld-link: error: undefined symbol: environ
>>> referenced by D:\...\zig-install\lib\zig\std\os.zig:1388
>>>               D:\...\zig-cache\o\4e0efec30ee2fdf8e70f3ef88ccff625\main.obj:(std.os.getenv)
error: LLDReportedFailure
```

From a quick glance at [the source](https://github.com/ziglang/zig/blob/f253822415304fc069f68452f7f4abbded58a24e/lib/std/os.zig#L1376-L1421) I see that I shouldn't(?) be using this function, but the `@compileError` didn't trigger. I'm not sure what should happen since I **am** linking libc and [`environ` is apparently available on Windows](https://docs.microsoft.com/en-us/cpp/c-runtime-library/environ-wenviron) (albeit with a leading underscore).

slimsag:
The compile error here would not be reached prior to a linker error:

```
lld-link: error: undefined symbol: environ
```

Fixed this, and clarified that `std.process.getEnvVarOwned` is the cross-platform API.

Fixes #8456

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -1655,7 +1655,10 @@ pub fn execvpeZ(
 
 /// Get an environment variable.
 /// See also `getenvZ`.
+///
+/// This API is not supported on Windows, see std.process.getEnvVarOwned for cross-platform API or std.os.getenvW for Windows-specific API.
 pub fn getenv(key: []const u8) ?[]const u8 {
+    if (builtin.os.tag == .windows) @compileError("std.os.getenv is unavailable for Windows because environment string is in WTF-16 format. See std.process.getEnvVarOwned for cross-platform API or std.os.getenvW for Windows-specific API.");
     if (builtin.link_libc) {
         var small_key_buf: [64]u8 = undefined;
         if (key.len < small_key_buf.len) {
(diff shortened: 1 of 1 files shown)


marler8997:
Are we sure `environ` doesn't work on windows?  The docs say it's deprecated but apparently it's defined somewhere (as `char**` so not WTF-16) https://docs.microsoft.com/en-us/cpp/c-runtime-library/environ-wenviron?view=msvc-170&viewFallbackFrom=vs-2017

The error you're getting could just be a problem with our mingw lib files or something.  We should double check what the msvc toolchain does here.

slimsag:
Ah, I see what you're saying, this may be working with msvc just not the default gnu target.. makes sense.

I won't have time to dig into this further soon, so I'll close for now.

PR #10954, opened 2022-02-21
Use lowercase in shasum for windows builds

charlieman:
The shasum generated for windows builds is showing as uppercase, this change makes it lowercase like all the other builds.

ziglang org download index json 2022-02-20


andrewrk:
Thanks!

PR #10964, opened 2022-02-21
translate-c: Add support for cast-to-union

### LINKED ISSUE #10955: translate-c: Incorrect code generated for cast-to-union
### Zig Version

0.10.0-dev.870+65aa33319

### Steps to Reproduce

Run `zig translate-c` on the following:

```C
union U {
    long l;
    double d;
};

void foo(void) {
    union U u = (union U) 42L;
}
```

### Expected Behavior

The union cast should be translated into something like this:
```zig
var u: union_U = union_U{ .l = @as(c_long, 42) };
```

### Actual Behavior

The union cast is translated like this, which does not compile:
```zig
var u: union_U = @as(union_U, @as(c_long, 42));
```

ehaas:
Fixes #10955

### DIFF
--- a/lib/std/zig/c_translation.zig
+++ b/lib/std/zig/c_translation.zig
@@ -30,6 +30,12 @@ pub fn cast(comptime DestType: type, target: anytype) DestType {
                 else => {},
             }
         },
+        .Union => |info| {
+            inline for (info.fields) |field| {
+                if (field.field_type == SourceType) return @unionInit(DestType, field.name, target);
+            }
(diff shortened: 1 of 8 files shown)


ehaas:
This appears to be triggering a stage2 TODO when running the new test in `translate_c_macros.zig`:

```
➜  build (cast-to-union) ./stage2/bin/zig test ../test/behavior.zig -I../test -fLLVM
/Users/ehaas/source/zig/build/stage2/lib/zig/std/zig/c_translation.zig:35:60: error: TODO: Sema.zirUnionInitPtr
                if (field.field_type == SourceType) return @unionInit(DestType, field.name, target);
                                                           ^
```

PR #10965, opened 2022-02-21
wasm: implement shared memory

xReveres:
This implements WebAssembly LLVM argument `--shared-memory`.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -2020,7 +2020,7 @@ fn buildOutputType(
         if (!target_info.target.cpu.features.isEnabled(@enumToInt(std.Target.wasm.Feature.atomics)) or
             !target_info.target.cpu.features.isEnabled(@enumToInt(std.Target.wasm.Feature.bulk_memory)))
         {
-            fatal("'atomics' and 'bulk-memory' features must be used in order to use shared memory", .{});
+            fatal("'atomics' and 'bulk-memory' features must be enabled to use shared memory", .{});
         }
     }
 
--- a/src/main.zig
+++ b/src/main.zig
@@ -2012,6 +2012,18 @@ fn buildOutputType(
             link_libcpp = true;
     }
 
+    if (target_info.target.cpu.arch.isWasm() and linker_shared_memory) {
+        if (output_mode == .Obj) {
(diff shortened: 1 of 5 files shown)


Luukdegram (changes requested):
Hi! Thank you for your contribution. This is looking pretty good so far. I do however have some feedback.
`--shared-memory` is only allowed when either the feature `atomics` or `bulk-memory` is enabled. By default we use the cpu features `generic` which currently only includes the MVP features.  https://github.com/ziglang/zig/blob/master/lib/std/target/wasm.zig#L101-L105. Apart from this, it's also not allowed to enable this flag when creating a relocatable object file.

So to get this merged, I ask you to ensure the following checks are included (preferably seperately in the main.zig file so the end users receives the error in the earliest stage):
- Atomics OR bulk-memory feature is enabled
- outputmode is not an object file

xReveres:
As I tested, LLVM requires both features `atomics` and `bulk-memory` to compile WebAssembly with shared memory, so I made checks for both features.
I am not sure I understood correctly object checking.

src/main.zig
```diff
@@ -2012,6 +2012,18 @@ fn buildOutputType(
...
 
+    if (target_info.target.cpu.arch.isWasm() and linker_shared_memory) {
+        if (output_mode == .Obj) {
+            fatal("shared memory is not allowed in object files", .{});
+        }
+
+        if (!target_info.target.cpu.features.isEnabled(@enumToInt(std.Target.wasm.Feature.atomics)) or
+            !target_info.target.cpu.features.isEnabled(@enumToInt(std.Target.wasm.Feature.bulk_memory)))
+        {
+            fatal("'atomics' and 'bulk-memory' features must be used in order to use shared memory", .{});
```
Luukdegram, line 2023: Small nitpick:
Replace with `"'atomics' and 'bulk-memory' features must be enabled to use shared memory"`

Luukdegram (changes requested):
Thank you for making the requested changes. One small textual enhancement left and this will be good to go.

Luukdegram (approved):
Thanks, LGTM

PR #10972, opened 2022-02-22
time: introduce Instant

kprotty:
`std.time.Instant` is an abstraction which helps represent elapsed time within a program. The name takes inspiration from Rust's [`std::time::Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html).

### Motivation

The motive is to delegate `std.time.timestamp` and friends for *telling time* (e.g. dates) while `std.time.Timer/std.time.Instant` are for *measuring time* (e.g. elapsed durations). 

The difference between Instant and Timer is that the former isn't required to *necessarily* be monotonic nor be representable in wall clock time like nanoseconds. This API allows for slightly faster sampling by deferring time conversions to a separate function. It also makes it trivial for Timer to be built on-top of it, providing the monotonicity guarantee in userspace instead of relying on the OS or hardware which [hasn't always been correct](https://doc.rust-lang.org/src/std/time.rs.html#247).

### Breaking Changes

This changes the structure of `Timer` for it to use `Instant` internally. Previously, fields like `resolution` and `frequency` were documented and probably publicly exposed. Those don't exist with this implementation.

The `Timer` also used `CLOCK_MONOTONIC` and `mach_absolute_time` in the past.  On darwin and linux these time sources don't tick when the system is suspended unlike windows `QueryPerformanceCounter` so the behavior differed edge-case-wise per platform. Now it uses `CLOCK_BOOTTIME` and (internally) `mach_continuous_time` respectively which *do* tick during suspend.

### DIFF
--- a/lib/std/time.zig
+++ b/lib/std/time.zig
@@ -155,7 +155,14 @@ pub const s_per_week = s_per_day * 7;
 /// It also tries to be monotonic, but this is not a guarantee due to OS/hardware bugs.
 /// If you need monotonic readings for elapsed time, consider `Timer` instead.
 pub const Instant = struct {
-    timestamp: if (!(builtin.os.tag == .wasi and !builtin.link_libc)) os.timespec else u64,
+    timestamp: if (is_posix) os.timespec else u64,
+
+    // true if we should use clock_gettime()
+    const is_posix = switch (builtin.os.tag) {
+        .wasi => builtin.link_libc,
+        .windows => false,
+        else => true,
+    };
 
     /// Queries the system for the current moment of time as an Instant.
     /// This is not guaranteed to be monotonic or steadily increasing, but for most implementations it is.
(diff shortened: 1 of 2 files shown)


lib/std/time.zig
```diff
@@ -127,149 +147,155 @@ pub const s_per_hour = s_per_min * 60;
...
-    /// https://github.com/ziglang/zig/pull/933
-    const monotonic_clock_id = os.CLOCK.MONOTONIC;
-
-    /// Initialize the timer structure.
-    /// Can only fail when running in a hostile environment that intentionally injects
-    /// error values into syscalls, such as using seccomp on Linux to intercept
-    /// `clock_gettime`.
+    /// Initialize the timer by querying for a supported clock.
+    /// Returns `error.TimerUnsupported` when such a clock is unavailable.
+    /// This should only fail in hostile environements such as linux seccomp misuse.
```
r00ster91, line 262:
```suggestion
    /// This should only fail in hostile environments such as linux seccomp misuse.
```

PR #10977, opened 2022-02-23
stage2 AArch64: more support for PIE targets (Mach-O)

joachimschmidt557:
This enables a handful of behavior tests previously disabled for `aarch64-macos`.

cc @kubkon

### DIFF
--- a/src/arch/aarch64/Emit.zig
+++ b/src/arch/aarch64/Emit.zig
@@ -110,6 +110,8 @@ pub fn emitMir(
 
             .load_memory_got => try emit.mirLoadMemoryPie(inst),
             .load_memory_direct => try emit.mirLoadMemoryPie(inst),
+            .load_memory_ptr_got => try emit.mirLoadMemoryPie(inst),
+            .load_memory_ptr_direct => try emit.mirLoadMemoryPie(inst),
 
             .ldp => try emit.mirLoadStoreRegisterPair(inst),
             .stp => try emit.mirLoadStoreRegisterPair(inst),
@@ -208,6 +210,8 @@ fn instructionSize(emit: *Emit, inst: Mir.Inst.Index) usize {
     switch (tag) {
         .load_memory_got,
         .load_memory_direct,
+        .load_memory_ptr_got,
+        .load_memory_ptr_direct,
         => return 2 * 4,
         .pop_regs, .push_regs => {
             const reg_list = emit.mir.instructions.items(.data)[inst].reg_list;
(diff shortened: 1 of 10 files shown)


src/arch/aarch64/Emit.zig
```diff
@@ -209,17 +210,9 @@ fn instructionSize(emit: *Emit, inst: Mir.Inst.Index) usize {
     switch (tag) {
         .load_memory_got,
         .load_memory_direct,
+        .load_memory_ptr_got,
+        .load_memory_ptr_direct,
```
kubkon, line 214: Why do we need all 4? Would just `load_memory_got` and `load_memory_direct` not suffice? Or do you want to flag a potential for having a sequence `adrp, ldr` vs `adrp, add, ldr`?

joachimschmidt557: For types that don't fit into a register, the `adrp, ldr` sequence doesn't work, right? In those cases we emit `adrp, add` and `memcpy`.

kubkon: Mhm, right. Also, I've seen multiple times cases of `adrp, add, ldr`, or `adrp, ldr, ldr`. The later sometimes qualifies for an instruction rewrite on the linker side depending on the relocation type of the pair `adrp, ldr` preceding `ldr`.

joachimschmidt557: I'm unsure about the case `adrp, add, ldr`, but the case `adrp, ldr, ldr` seems to me like the contents of the memory location itself is a pointer to something else (maybe a global variable `x` which is a `*u64` and we want to access `x.*` ?)

kubkon: Yep, correct. However, MachO does certain optimisations such as using a different mechanism for TLS pointer loading, which implies a rewrite of `adrp, ldr, ldr` into `adrp, add, ldr` (note that the linker doesn't touch a non-relocated intstruction, in this case the last `ldr`). Anyhow, I digress :-)

joachimschmidt557: At the moment, the only time `load_memory_ptr_got` and `load_memory_ptr_direct` are emitted are for memcpy. If there is a `adrp, add, ldr` sequence I'm pretty sure that the last `ldr` is unrelated to loading the address.

kubkon: Fair, but something to bear in mind for sure though.

Issue #4812, opened 2020-03-25
MAX_PATH_BYTES is fiction


daurnimator:
> `MAX_PATH_BYTES` is a fiction (see e.g. https://eklitzke.org/path-max-is-tricky). It would probably make sense to loop until we don't get `ENAMETOOLONG`?

_Originally posted by @daurnimator in https://github.com/ziglang/zig/pull/4811

> @daurnimator can you open a separate issue to discuss `MAX_PATH_BYTES`?

_Originally posted by @andrewrk in https://github.com/ziglang/zig/pull/4811#issuecomment-604107731_

andrewrk:
I experimented with this and found it to be fact - for example the Linux kernel itself will give `ENAMETOOLONG` when syscalls are done with paths longer than `PATH_MAX`.

You can have paths that cannot be accessed with an absolute pathname, because the syscall will reject it, and still access them if you use syscalls based on directory handles. This is one reason zig's std lib has been transitioning to use directory handles for relevant operations.

The article says it could be a
> filesystem limitation, or at the very least a kernel parameter

Admittedly I did all my testing on ext4/linux, however, I'd like to see evidence / a demonstration of this before proceeding further.

andrewrk:
Oops, I misunderstood that the entire article was on this topic and so only read 1 heading. Now I finished reading the article, and it says: 

> On Linux, system calls like open(2) perform this check via getname_flags(). The check is performed using the 4096 byte PATH_MAX value, which, as I just finished explaining, is not really a file path limit!

This is what I already believed to be true, and it means zig's `PATH_MAX` (posix) and `MAX_PATH_BYTES` (posix+windows) are correct.

Now I will say that some std lib abstractions (the ones higher level than `std.os`) should be improved so that `error.NameTooLong` is no longer in the error set. That would be a welcome improvement to the std lib.

gereeter:
> Now I will say that some std lib abstractions (the ones higher level than `std.os`) should be improved so that `error.NameTooLong` is no longer in the error set.

[...]

PR #10988, opened 2022-02-25
Add assumptions for `resolveWindows`, `resolvePosix`and `realpath`

matu3ba:
* doc: `realpath` description
* doc: `resolveWindows`, `resolvePosix` do not work with relative paths

### DIFF
--- a/lib/std/fs/path.zig
+++ b/lib/std/fs/path.zig
@@ -468,6 +468,10 @@ fn asciiEqlIgnoreCase(s1: []const u8, s2: []const u8) bool {
 }
 
 /// On Windows, this calls `resolveWindows` and on POSIX it calls `resolvePosix`.
+/// Note: Symlinks currently dont work and should not be used,
+/// because Windows and Posix having different behavior.
+/// On Windows, symlinks are resolved after `..` is resolved.
+/// On POSIX, symlinks are resolved eagerly.
 pub fn resolve(allocator: Allocator, paths: []const []const u8) ![]u8 {
     if (native_os == .windows) {
         return resolveWindows(allocator, paths);
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -1438,7 +1438,7 @@ var wasi_cwd = if (builtin.os.tag == .wasi and !builtin.link_libc) struct {
 /// Note that `cwd_init` corresponds to a Preopen directory, not necessarily
 /// a POSIX path. For example, "." matches a Preopen provided with `--dir=.`
 ///
-/// This must be called before using any relative or absolute paths with `std.os` 
+/// This must be called before using any relative or absolute paths with `std.os`
 /// functions, if you are on WASI without linking libc.
 ///
 /// `alloc` must not be a temporary or leak-detecting allocator, since `std.os`
@@ -1472,7 +1472,7 @@ pub fn initPreopensWasi(alloc: Allocator, cwd_init: ?[]const u8) !void {
 
 /// Resolve a relative or absolute path to an handle (`fd_t`) and a relative subpath.
 ///
-/// For absolute paths, this automatically searches among available Preopens to find 
+/// For absolute paths, this automatically searches among available Preopens to find
 /// a match. For relative paths, it uses the "emulated" CWD.
 pub fn resolvePathWasi(path: []const u8, out_buffer: *[MAX_PATH_BYTES]u8) !RelativePathWasi {
     // Note: Due to WASI's "sandboxed" file handles, operations with this RelativePathWasi
@@ -5091,6 +5091,10 @@ pub const RealPathError = error{
 /// extra `/` characters in `pathname`.
 /// The return value is a slice of `out_buffer`, but not necessarily from the beginning.
 /// See also `realpathZ` and `realpathW`.
+/// Note: All /-separated components of the supplied path must resolve to a
+/// directory, or a symlink to a directory, except for the last component
+/// Use `fs.path.resolve` or `fs.path.relative` to get dir of file (`file/..`).
+/// assume: if relative path used, then they are relative to `getcwd()`
 pub fn realpath(pathname: []const u8, out_buffer: *[MAX_PATH_BYTES]u8) RealPathError![]u8 {
     if (builtin.os.tag == .windows) {
         const pathname_w = try windows.sliceToPrefixedFileW(pathname);


matu3ba:
This was quite abit surprising behavior on figuring out the problem with workarounds of https://github.com/matu3ba/chepa/issues/1

topolarity:
I'm not sure the realpath description is accurate either, or maybe I'm just misunderstanding it.

`realpath` currently gives me these result in the Zig repos (both with and without libc):

```bash
realpath("lib") = "/home/topolarity/repos/zig/lib"
realpath("lib/..") = "/home/topolarity/repos/zig"
realpath("build.zig") = "/home/topolarity/repos/zig/build.zig"
realpath("build.zig/..") = error.NotDir
```

Is the comment pointing out that "build.zig/.." fails, or are you hitting a different problem?

matu3ba:
> Is the comment pointing out that "build.zig/.." fails, or are you hitting a different problem?

Yes. Mostly its about pointing out that "build.zig/.." fails.

topolarity:
I don't personally find that behavior surprising. However, I do see how that could depend on one's familiarity with POSIX

I don't think the comment needs to mention "/" at all - Maybe clearer phrasing would be:
> Note: All `/`-separated components of the supplied path must resolve to a 
> directory, or a symlink to a directory, except for the last component

matu3ba:
@topolarity Do you think this good enough?
I dont want to mess with the other text right now.
```txt
+/// Note: All /-separated components of the supplied path must resolve to a
+/// directory, or a symlink to a directory, except for the last component
+/// Use `fs.path.resolve` or `fs.path.relative` to get dir of file (`file/..`).
+/// assume: if relative path used, then they are relative to `getcwd()`
```

topolarity:
Looking better :+1: Last nit is that I think we should either mention that `path.resolve` and `.relative` do not resolve symlinks, or we should not recommend them here because of that behavior difference

FWIW, the reason file-relative paths work on Windows is that symlinks are resolved after ".." is resolved (roughly equivalent to doing path.resolve, followed by realpath). However, this changes how symlinks interact with `..`: "./symlink/.." resolves to "." (the normal Windows behavior), rather than the parent of the symlink target (which is the normal POSIX behavior)

WASI is tackling the same behavioral differences in https://github.com/WebAssembly/wasi-filesystem/issues/26 and so far seems to be leaning towards leaving this up to the Host OS

lib/std/fs/path.zig
```diff
@@ -484,6 +484,7 @@ pub fn resolve(allocator: Allocator, paths: []const []const u8) ![]u8 {
 /// Path separators are canonicalized to '\\' and drives are canonicalized to capital letters.
 /// Note: all usage of this function should be audited due to the existence of symlinks.
 /// Without performing actual syscalls, resolving `..` could be incorrect.
+/// assume: paths describes network path (prefix `\\`) or root path (prefix `X:` or `x:`)
```
topolarity, line 487: This comment still looks incorrect to me - This function intends to support relative paths, resolving versus the CWD if necessary

Same for `resolvePosix`

matu3ba: I will change it then to `TODO: support relative paths`

topolarity: What's the case that's failing? There a couple of basic relative path tests that seem to be passing already

matu3ba: added the TODOs.

lib/std/fs/path.zig
```diff
@@ -468,6 +468,9 @@ fn asciiEqlIgnoreCase(s1: []const u8, s2: []const u8) bool {
 }
 
 /// On Windows, this calls `resolveWindows` and on POSIX it calls `resolvePosix`.
+/// Note: Dont use for symlinks, because Windows and Posix having different behavior.
```
topolarity, line 471: This is misleading - `resolve` currently makes no attempt to resolve symlinks on any platform

lib/std/fs/path.zig
topolarity: There are already working test cases at https://github.com/ziglang/zig/blob/f59cbd89e349fb3002500cb2a1d69ca5e6063338/lib/std/fs/path.zig#L736-L751

If you're noticing a failure, I think that should be filed as a separate issue

lib/std/fs/path.zig
matu3ba: added
```txt
/// Note: Symlinks currently dont work and should not be used,
/// because Windows and Posix having different behavior.
```

lib/std/fs/path.zig
```diff
@@ -468,6 +468,10 @@ fn asciiEqlIgnoreCase(s1: []const u8, s2: []const u8) bool {
 }
 
 /// On Windows, this calls `resolveWindows` and on POSIX it calls `resolvePosix`.
+/// Note: Symlinks currently dont work and should not be used,
+/// because Windows and Posix having different behavior.
+/// On Windows, symlinks are resolved after `..` is resolved.
```
topolarity, line 473: I think "Note: does not expand symlinks" is sufficient for the note here

The details on symlink resolution between platforms have to do with `realpath` and the other `std.os` calls, not this function

topolarity, line 473: I think "Note: does not expand symlinks" is sufficient for the note here

The details on symlink resolution between platforms have to do with `realpath` and the other `std.os` calls, not this function

matu3ba: > I think "Note: does not expand symlinks" is sufficient for the note here

agreed.

> The details on symlink resolution between platforms have to do with realpath and the other std.os calls, not this function

Why does this not belong into the description of the portable method, if it is a portability restriction? 
I would expect that users tend to do tests on one platform with the most portable method and extrapolate to the others.

lib/std/os.zig
```diff
@@ -5091,6 +5091,10 @@ pub const RealPathError = error{
 /// extra `/` characters in `pathname`.
 /// The return value is a slice of `out_buffer`, but not necessarily from the beginning.
 /// See also `realpathZ` and `realpathW`.
+/// Note: All /-separated components of the supplied path must resolve to a
+/// directory, or a symlink to a directory, except for the last component
+/// Use `fs.path.resolve` or `fs.path.relative` to get dir of file (`file/..`).
```
kubkon, line 5096: `realpath` should fully resolve any path containing relative `.` and `..` components on any POSIX-compliant system as far as I remember. It is only Windows that is the problem where the NT kernel call we issue is unable to deal with those which implies it is up to the user to `resolve` the relative components first.

matu3ba: The non-working semantics (`pathtofolder/file/..`) of `fs.path.relative` can be seen here https://github.com/matu3ba/chepa/commit/f34084cbf103ca9bbf75db94bf4b1835a8c9b6aa#diff-d924ca21d81d7d5a59eb9e10d3e2689c4c58a7e28e6ecce6a064701666f5a730L173

> It is only Windows that is the problem where the NT kernel call we issue is unable to deal with those which implies it is up to the user to resolve the relative components first.

Could we on Windows work around the problem with calling relative on the subpath and then call resolve for each extending subpath? If you can point me to the things I need for testing, I can play abit to see if that could work.
Kind related is https://github.com/ziglang/zig/issues/4812.

andrewrk:
The changes here are incorrect and unhelpful. I'll scrutinize these doc comments at some point, but at a glance this PR is not an improvement.

PR #10991, opened 2022-02-25
macho,x64,aarch64: misc fixes

kubkon:
* aarch64: fix alignment issues for consts with natural ABI alignment not matching that of the `ldr` instruction in `aarch64` - solved by preceding the `ldr` with an additional `add` instruction to form  the full address before dereferencing the pointer.
* macho: redo selection of segment/section for decls and consts based on combined type and value
* aarch64: check if type has runtime bits before allocating mem ptr
* x64+aarch64: check for pointer to zero-bit type when lowering decl - unless the pointer is a pointer to a function, if the pointee type has zero-bits, we need to return `MCValue.none` as the `Decl` has not been lowered to memory, and therefore, any GOT reference will be wrong.

cc @joachimschmidt557

### DIFF
--- a/src/arch/aarch64/Emit.zig
+++ b/src/arch/aarch64/Emit.zig
@@ -208,8 +208,8 @@ fn instructionSize(emit: *Emit, inst: Mir.Inst.Index) usize {
     }
 
     switch (tag) {
+        .load_memory_direct => return 3 * 4,
         .load_memory_got,
-        .load_memory_direct,
         .load_memory_ptr_got,
         .load_memory_ptr_direct,
         => return 2 * 4,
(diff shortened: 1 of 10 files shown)


src/arch/aarch64/Emit.zig
```diff
@@ -654,15 +654,31 @@ fn mirLoadMemoryPie(emit: *Emit, inst: Mir.Inst.Index) !void {
...
+        .load_memory_got => {
+            // ldr reg, reg, offset
+            try emit.writeInstruction(Instruction.ldr(
+                reg,
+                reg,
+                Instruction.LoadStoreOffset.imm(0),
+            ));
+        },
+        .load_memory_direct => {
+            // We cannot load the offset directly as it may not be aligned properly.
```
joachimschmidt557, line 672: I think another long-term solution for this is to emit `ldrb`, `ldrh`, `ldrsb`, and `ldrsh` when appropriate. While `ldr` requires the address to be aligned to 8 bytes, `ldrh`/`ldrsh` require an alignment of 2 and `ldrb`/`ldrsb` require an alignment of 1 (no alignment).

Of course, these instructions are meant for loading integers which have natural alignment. For types with different alignments, we can fall back to the `adrp, add, ldr` or have the linker guarantee larger alignment so the `add` is not necessary anymore.

But for now, this is a perfectly fine solution.

joachimschmidt557 (commented):
Looks good to me!

src/arch/aarch64/Emit.zig
kubkon: Agreed on different `ldr` variants. As for guaranteeing larger alignment by the linker, that's a no-go: the linker should only ever put the decl/const at the required alignment. However, it can optimise the `add` out and replace with an `ldr` if the target address satisfies the alignment requirements of the `ldr` instruction. It can then also `nop` the 3rd `ldr` in that case; however note it won't be able to remove it as this would potentially destroy subsequent relocations as the offsets would change potentially resulting in misaligned accesses etc.

joachimschmidt557: > Agreed on different `ldr` variants. As for guaranteeing larger alignment by the linker, that's a no-go: the linker should only ever put the decl/const at the required alignment.

That makes sense, yes.

> However, it can optimise the `add` out and replace with an `ldr` if the target address satisfies the alignment requirements of the `ldr` instruction. It can then also `nop` the 3rd `ldr` in that case; however note it won't be able to remove it as this would potentially destroy subsequent relocations as the offsets would change potentially resulting in misaligned accesses etc.

Agreed. After Emit, we shouldn't be allowed to delete instructions.

PR #10994, opened 2022-02-26
stage2: `@TypeOf` improvements

Vexu:
Arguments to `@TypeOf` need to be analyzed in their own block so that we don't end up with trash instructions in codegen. This also makes it easy to avoid issues with some instructions requiring a runtime block when we only care about their type.


andrewrk:
Nice, looks like the tree form of ZIR/AIR is working out nicely for this case at least.

PR #11000, opened 2022-02-27
std.HashMap: return explicit errors

motiejus:
All errors from std.ArrayList are allocation errors. Mark them as
such. This is helpful when one wants to return explicit errors where
HashMap is used.


motiejus:
For future reference for self, in vim:

```
%s/) !\([^=]\)/) Allocator.Error!\1/g
```

PR #11002, opened 2022-02-27
stage2: Add comptime result for certain unsigned/comptime comparisons

topolarity:
This PR implements a `comptime` result for certain comparisons between unsigned integers and comptime values.

For example:

```
 ( 0 <= (unsigned runtime value)) => true
 (-1 <  (unsigned runtime value)) => true
 ((unsigned runtime value) < -15) => false
 ((unsigned runtime value) <  15) => (runtime comparison)
 ```

**Note:** Along the way, I noticed that `test/behavior/eval.zig` was failing when executed on its own. This seems to be caused by an unresolved union layout in zirTagName. I've added the fix to this PR, but unfortunately I wasn't able to add any tests, because I'm not sure how to reliably create a union with un-resolved layout for testing.


andrewrk:
Nice work!

PR #11006, opened 2022-02-27
stage2: Resolve alignment for union field in `@TypeInfo`

topolarity:
This also includes two other small fixes:
 1. Instantiate void TypeInfo fields as `void`
 2. Return an error in `type.comptimeOnly` when encountering unresolved comptime requirements

The problem behind (2) is that type.hasRuntimeBits() can hit an `unreachable` because "requires_comptime" is not resolved for some child structs.

Since we already approximate (i.e. return `true`, even though the type may end up comptime-only) when encountering unknown comptime requirements in other branches, this PR updates the function to approximate in all branches.

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -1125,6 +1125,8 @@ pub const Union = struct {
         abi_align: Value,
 
         /// Returns the field alignment, assuming the union is not packed.
+        /// Keep implementation in sync with `Sema.unionFieldAlignment`.
+        /// Prefer to call that function instead of this one during Sema.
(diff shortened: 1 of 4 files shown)


topolarity:
The alternative change here would be to make sure that `requires_comptime` is reliably resolved for child structs.

Specifically, sema.analyzeStructDecl() and sema.typeRequiresComptime() would have to be changed to no longer eagerly exit without updating requires_comptime for all child structs.

andrewrk:
I solved the problem a slightly different way before merging - hope you don't mind

PR #11010, opened 2022-02-28
stage2: peer resolve error sets and error unions 

mitchellh:
I surprisingly could only find 1 test that passes thanks to this (no new failures). But I added a number of new cases to better cover error union/set peer resolution in various cases and these also pass in stage1.

One question: note the TODO in Sema L17404 I added. A generic poison is making its way through somehow if I don't guard against it. Is the way I'm guarding fine?

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -17814,8 +17814,8 @@ fn resolvePeerTypes(
                 err_set_ty = try err_set_ty.?.errorSetMerge(sema.arena, candidate_ty);
                 continue;
             },
-            .ErrorUnion => {
-                if (chosen_ty_tag == .ErrorSet) {
+            .ErrorUnion => switch (chosen_ty_tag) {
+                .ErrorSet => {
                     if (err_set_ty.?.isAnyError()) {
                         chosen = candidate;
                         chosen_i = candidate_i + 1;
@@ -17854,9 +17854,9 @@ fn resolvePeerTypes(
                     chosen = candidate;
                     chosen_i = candidate_i + 1;
                     continue;
-                }
+                },
 
-                if (chosen_ty_tag == .ErrorUnion) {
+                .ErrorUnion => {
                     const chosen_payload_ty = chosen_ty.errorUnionPayload();
                     const candidate_payload_ty = candidate_ty.errorUnionPayload();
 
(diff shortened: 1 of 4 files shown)


src/Sema.zig
```diff
@@ -17499,29 +17502,269 @@ fn resolvePeerTypes(
...
+                    // If candidate is superset of chosen, switch it.
+                    // If neither is a superset, merge errors.
+                    var prev_is_superset = true;
+                    for (candidate_ty.errorSetNames()) |name| {
+                        if (!err_set_ty.?.errorSetHasField(name)) {
+                            prev_is_superset = false;
+                            break;
+                        }
+                    }
+                    if (prev_is_superset) continue; // use previous
```
Vexu, line 17536: Instead of a pattern like this you should label the outer loop and use the labeled continue syntax.

mitchellh: Sorry, can you clarify what you're asking? Would you prefer this?

```zig
var prev_is_superset = blk: {
    for (candidate_ty.errorSetNames()) |name| 
        if (!err_set_ty.?.errorSetHasField(name)) break :blk false;

    break :blk true;
}
if (prev_is_superset) continue; 
```

Vexu: I meant:
```zig
candidate_loop: for (instructions[1..]) |candidate, candidate_i| {
...
for (candidate_ty.errorSetNames()) |name| {
    if (!err_set_ty.?.errorSetHasField(name)) {
        continue :main_loop; // use previous
    }
}
```
Using a flag variable like this is not necessary in Zig thanks to labeled loops allowing breaking from nested loops.

mitchellh: Oh yes, I would do that, but the logic is opposite: we only want to continue the main loop if we make it all the way through (_not_ break).

mitchellh: Actually, I think I've identified a way to clean this up.

Vexu: Oh I see, in that case you'll want to use the else branch:
```zig
for (candidate_ty.errorSetNames()) |name| {
    if (!err_set_ty.?.errorSetHasField(name)) break;
} else continue :candidate_loop;
```

mitchellh: Ah hah! I misunderstood `else` to only work for optionals, not generally for any loop done semantics. Perfect. I updated it, tests pass, and pushed the commit.

src/Sema.zig
```diff
@@ -17398,6 +17398,21 @@ fn resolvePeerTypes(
     const target = sema.mod.getTarget();
 
     var chosen = instructions[0];
+    var err_set_ty: ?Type = blk: {
+        const chosen_ty = sema.typeOf(chosen);
+
+        // TODO: is this the right handling of generic poison?
+        if (chosen_ty.tag() == .generic_poison or chosen_ty.zigTypeTag() != .ErrorSet)
```
andrewrk, line 17405: pass it up the stack:

```suggestion
        const chosen_ty_tag = try chosen_ty.zigTypeTagOrPoison();
        if (chosen_ty_tag != .ErrorSet)
```

This mechanism is (at least for now) how we obtain a function type for an un-instantiated generic function. We try to obtain parameter types, but it might not be possible if the type depends on something that is only available in the instantiation. Example:

```zig
fn foo(T: type, x: T, other: i32) void {}
```

In this example when we look at `@TypeOf(foo)`, the parameter `x` will cause generic poison, but `other` will return the type `i32` for that parameter.

Hope that helps clear things up.

andrewrk (changes requested):
The rest of the logic looks good at first glance - I'll give it a closer inspection on the second round.

src/Sema.zig
mitchellh: Thanks, done!

andrewrk:
I did a bunch of work on `Type.hash` and `Type.eql` today. I noticed that the implementation for error sets for both is incomplete. If you were planning on working on more test cases related to errors, I bet it would help to make the next step fully implementing these functions:

https://github.com/ziglang/zig/blob/eaf1c97ce80606505768b8dd8daf8b479aec91a7/src/type.zig#L558-L587

PR #11017, opened 2022-02-28
stage2 ARM: implement return types with abi size > 4

joachimschmidt557:


src/arch/arm/CodeGen.zig
```diff
@@ -2491,42 +2527,59 @@ fn airCall(self: *Self, inst: Air.Inst.Index) !void {
...
+                arg_ty,
+                info.stack_byte_count - offset,
+                arg_mcv,
+            ),
         }
+    }
 
+    // Due to incremental compilation, how function calls are generated depends
+    // on linking.
+    if (self.bin_file.tag == link.File.Elf.base_tag or self.bin_file.tag == link.File.Coff.base_tag) {
```
andrewrk, line 2582: suggestion: `switch (self.bin_file.tag)`

andrewrk (approved):

src/arch/arm/CodeGen.zig
joachimschmidt557: I just moved some old code, but this is a good catch, I converted the relevant code into a switch.

PR #11018, opened 2022-03-01
std.BufSet.clone: fix key ownership

motiejus:
This was introduced in d1a46548349a902c30057b3ba66ebad9bc25bdd2: when a
BufSet clones the keys, it used to assign the new pointers to the old
struct. Fix that by assigning the pointers to the correct, i.e. the new,
struct.

This caused double-free when using arena allocator for the new struct,
also in the test case.


andrewrk:
Thanks for the follow-up!

Issue #5019, opened 2020-04-12
rename std.os to std.posix


andrewrk:
`std.os.windows` and similar will remain the same, but everything else, e.g. `std.os.write` would become `std.posix.write`.

anosovic:
why do we want the `posix` namespace by itself and not under `os`?

karrick:
Is there merit to having a OS agnostic library, which makes calls into posix and window libraries?

That way most programmers can write OS agnostic code, and standard library writers can focus on how to implement that code using both posix and windows sub-libraries?

andrewrk:
> Is there merit to having a OS agnostic library

All of the std lib is OS agnostic except what is namespaced under `std.os`

Flaminator:
> All of the std lib is OS agnostic except what is namespaced under `std.os`
Will this still be that case after moving the POSIX parts from `std.os`  to `std.posix` or will that statement move to also include what is namespaced under `std.posix`? As POSIX isn't os agnostic.

kubkon:
> All of the std lib is OS agnostic except what is namespaced under `std.os`
> Will this still be that case after moving the POSIX parts from `std.os` to `std.posix` or will that statement move to also include what is namespaced under `std.posix`? As POSIX isn't os agnostic.

@andrewrk might correct me here but I guess `std.posix` wouldn't be platform agnostic in its entirety, although for the most part we could probably successfully emulate POSIX behaviour in say Windows for example (with a few exceptions ofc). In fact, I'd imagine that with #1840 landing, we'd get really close to achieving that goal since hooking directly into NT API seems to offer a lot more flexibility *plus* it does allow for something close to `*at` syscalls you'd normally find in POSIX.

Flaminator:
@kubkon I know I am no authority when it comes to stuff like this but I don’t really think trying to emulate POSIX behaviour on non-POSIX platforms is a good or a preferable thing. It might make it easier in someways but mostly towards one group.

I have noticed over the last 10-15 years with all these new languages popping up that most of these languages are primarily developed by people that use Linux or a different Unix-like operating system. Which of course means that a lot of the interfaces, apis and tooling are heavily focussed on POSIX ideas of how things work and should work. This brings some annoyances to users on operating systems like Windows where things just work differently. I have seen it happen in Go, Rust, the filesystem api in C++ and even in some places in Zig.

andrewrk:
[...]

PR #11021, opened 2022-03-01
stdlib: Add emulated CWD to std.os for WASI targets

### LINKED ISSUE #10749: add the concept of cwd to standard library file system abstraction for WASI
Here's how they do it in WASI libc:
 * https://github.com/WebAssembly/wasi-libc/pull/214
 * https://github.com/WebAssembly/wasi-libc/pull/230

In the Zig standard library, what we currently do for WASI is this:

https://github.com/ziglang/zig/blob/abbcf4032770676aee0e4a37f19d9e9ad2bdd992/lib/std/fs.zig#L2242-L2245

Notice that for WASI libc however they have an "emulated FD_ATCWD". It looks like it just treats `/` as the cwd.

Instead of erroring out, let's follow this convention by returning fd = `std.os.wasi.AT.FDCWD` which matches the WASI libc convention of `-2`.

Then we have to check for FD_ATCWD at every std.os function (see similar code in the WASI libc PR linked above). For example:

```c
int fstatat(int dirfd, const char *__restrict pathname, struct stat *__restrict statbuf, int flags) {
    if (dirfd == AT_FDCWD || pathname[0] == '/') {
        return __wasilibc_stat(pathname, statbuf, flags);
    }

    return __wasilibc_nocwd_fstatat(dirfd, pathname, statbuf, flags);
}
```

This is a prerequisite for #10716.

topolarity:
This PR provides emulated relative and absolute path support for `std.os` on WASI targets (resolves #10749). 

"os.initPreopensWasi()" must be called before `std.os` functions can resolve relative or absolute paths, which is asserted at runtime.

Support has been added for: `open`, `rename`, `mkdir`, `rmdir`, `chdir`, `fchdir`, `link`, `symlink`, `unlink`, `readlink`, `fstatat`, `access`, and `faccessat`.

This also includes limited support for `getcwd()` and `realpath()`. These return an error if the CWD does not correspond to a Preopen with an absolute path. They also do not currently expand symlinks. Tests have been updated to expect the CWD at "/cwd" rather than ".", so that "realpath" will work for CWD-relative paths.

### DIFF
--- a/lib/std/fs/wasi.zig
+++ b/lib/std/fs/wasi.zig
@@ -194,6 +194,17 @@ pub const PreopenList = struct {
         return null;
     }
 
+    /// Find preopen by fd. If the preopen exists, return it.
+    /// Otherwise, return `null`.
+    pub fn findByFd(self: Self, fd: fd_t) ?Preopen {
+        for (self.buffer.items) |preopen| {
+            if (preopen.fd == fd) {
+                return preopen;
+            }
+        }
+        return null;
+    }
+
     /// Find preopen by type. If the preopen exists, return it.
     /// Otherwise, return `null`.
     pub fn find(self: Self, preopen_type: PreopenType) ?*const Preopen {
@@ -224,6 +235,6 @@ test "extracting WASI preopens" {
 
     try preopens.populate();
 
-    const preopen = preopens.find(PreopenType{ .Dir = "/cwd" }) orelse unreachable;
-    try std.testing.expect(preopen.@"type".eql(PreopenType{ .Dir = "/cwd" }));
+    const preopen = preopens.find(PreopenType{ .Dir = "." }) orelse unreachable;
+    try std.testing.expect(preopen.@"type".eql(PreopenType{ .Dir = "." }));
 }
--- a/lib/std/fs/wasi.zig
+++ b/lib/std/fs/wasi.zig
@@ -25,10 +25,26 @@ pub const PreopenType = union(PreopenTypeTag) {
     const Self = @This();
 
     pub fn eql(self: Self, other: PreopenType) bool {
-        if (!mem.eql(u8, @tagName(self), @tagName(other))) return false;
+        if (std.meta.activeTag(self) != std.meta.activeTag(other)) return false;
 
         switch (self) {
             PreopenTypeTag.Dir => |this_path| return mem.eql(u8, this_path, other.Dir),
+        }
+    }
+
+    // Checks whether `other` refers to a subdirectory of `self` and, if so,
+    // returns the relative path to `other` from `self`
+    pub fn getRelativePath(self: Self, other: PreopenType) ?[]const u8 {
+        if (std.meta.activeTag(self) != std.meta.activeTag(other)) return null;
+
+        switch (self) {
+            PreopenTypeTag.Dir => |this_path| {
+                const other_path = other.Dir;
+                if (mem.indexOfDiff(u8, this_path, other_path)) |index| {
+                    if (index < this_path.len) return null;
+                }
+                return other_path[this_path.len..];
+            },
         }
     }
 
@@ -60,6 +76,15 @@ pub const Preopen = struct {
             .@"type" = preopen_type,
         };
     }
+};
+
+/// WASI resource identifier struct. This is effectively a path within
+/// a WASI Preopen.
+pub const PreopenUri = struct {
+    /// WASI Preopen containing the resource.
+    base: Preopen,
+    /// Path to resource within `base`.
+    relative_path: []const u8,
 };
 
 /// Dynamically-sized array list of WASI preopens. This struct is a
@@ -137,10 +162,36 @@ pub const PreopenList = struct {
                 .SUCCESS => {},
                 else => |err| return os.unexpectedErrno(err),
             }
+
             const preopen = Preopen.new(fd, PreopenType{ .Dir = path_buf });
             try self.buffer.append(preopen);
             fd = try math.add(fd_t, fd, 1);
         }
+    }
+
+    /// Find a preopen which includes access to `preopen_type`.
+    ///
+    /// If the preopen exists, `relative_path` is updated to point to the relative
+    /// portion of `preopen_type` and the matching Preopen is returned. If multiple
+    /// preopens match the provided resource, the most recent one is used.
+    pub fn findContaining(self: Self, preopen_type: PreopenType) ?PreopenUri {
+        // Search in reverse, so that most recently added preopens take precedence
+        var k: usize = self.buffer.items.len;
+        while (k > 0) {
+            k -= 1;
+
+            const preopen = self.buffer.items[k];
+            if (preopen.@"type".getRelativePath(preopen_type)) |rel_path_orig| {
+                var rel_path = rel_path_orig;
+                while (rel_path.len > 0 and rel_path[0] == '/') rel_path = rel_path[1..];
+
+                return PreopenUri{
+                    .base = preopen,
+                    .relative_path = if (rel_path.len == 0) "." else rel_path,
+                };
+            }
+        }
+        return null;
     }
 
     /// Find preopen by type. If the preopen exists, return it.
@@ -173,8 +224,6 @@ test "extracting WASI preopens" {
 
     try preopens.populate();
 
-    try std.testing.expectEqual(@as(usize, 1), preopens.asSlice().len);
-    const preopen = preopens.find(PreopenType{ .Dir = "." }) orelse unreachable;
-    try std.testing.expect(preopen.@"type".eql(PreopenType{ .Dir = "." }));
-    try std.testing.expectEqual(@as(i32, 3), preopen.fd);
+    const preopen = preopens.find(PreopenType{ .Dir = "/cwd" }) orelse unreachable;
+    try std.testing.expect(preopen.@"type".eql(PreopenType{ .Dir = "/cwd" }));
 }
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -21,9 +21,13 @@ const assert = std.debug.assert;
 const math = std.math;
 const mem = std.mem;
 const elf = std.elf;
+const fs = std.fs;
 const dl = @import("dynamic_library.zig");
 const MAX_PATH_BYTES = std.fs.MAX_PATH_BYTES;
 const is_windows = builtin.os.tag == .windows;
+const Allocator = std.mem.Allocator;
+const Preopen = std.fs.wasi.Preopen;
+const PreopenList = std.fs.wasi.PreopenList;
 
 pub const darwin = std.c;
 pub const dragonfly = std.c;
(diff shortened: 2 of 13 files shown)


lib/std/fs/wasi.zig
```diff
@@ -25,13 +25,29 @@ pub const PreopenType = union(PreopenTypeTag) {
...
+        if (std.meta.activeTag(self) != std.meta.activeTag(other)) return false;
 
         switch (self) {
             PreopenTypeTag.Dir => |this_path| return mem.eql(u8, this_path, other.Dir),
         }
     }
 
+    // Checks whether `other` refers to a subdirectory of `self` and, if so,
+    // returns the relative path to `other` from `self`
+    pub fn get_relative_path(self: Self, other: PreopenType) ?[]const u8 {
```
kubkon, line 37:
```suggestion
    pub fn getRelativePath(self: Self, other: PreopenType) ?[]const u8 {
```

The rule of thumb for naming is: for calls corresponding 1:1 with syscalls, etc., preserve the name of the source if possible, for any Zig abstraction, camelCase.

lib/std/fs/wasi.zig
```diff
@@ -62,6 +78,15 @@ pub const Preopen = struct {
     }
 };
 
+/// WASI resource identifier struct. This is effectively a path within
+/// a WASI Preopen.
+pub const PreopenURI = struct {
```
kubkon, line 83: This is a stylistic nit more than anything else:

```suggestion
pub const PreopenUri = struct {
```

lib/std/fs/wasi.zig
```diff
@@ -131,18 +156,44 @@ pub const PreopenList = struct {
                 else => |err| return os.unexpectedErrno(err),
             }
             const preopen_len = buf.u.dir.pr_name_len;
-            const path_buf = try self.buffer.allocator.alloc(u8, preopen_len);
+            var path_buf = try self.buffer.allocator.alloc(u8, preopen_len);
```
kubkon, line 159: Is this change needed here? I don't see you re-assigning `path_buf` anywhere. Note that `allocator.alloc` returns a pointer so you can still do interior mutability on a `const`.

lib/std/fs/wasi.zig
```diff
@@ -131,18 +156,44 @@ pub const PreopenList = struct {
...
             fd = try math.add(fd_t, fd, 1);
         }
     }
 
+    /// Find a preopen which includes access to `preopen_type`.
+    ///
+    /// If the preopen exists, `relative_path` is updated to point to the relative
+    /// portion of `preopen_type` and the matching Preopen is returned. If multiple
+    /// preopens match the provided resource, the most recent one is used.
+    pub fn find_containing(self: Self, preopen_type: PreopenType) ?PreopenURI {
```
kubkon, line 177: Ditto

```suggestion
    pub fn findContaining(self: Self, preopen_type: PreopenType) ?PreopenURI {
```

lib/std/fs/wasi.zig
```diff
@@ -173,8 +224,6 @@ test "extracting WASI preopens" {
 
     try preopens.populate();
 
-    try std.testing.expectEqual(@as(usize, 1), preopens.asSlice().len);
-    const preopen = preopens.find(PreopenType{ .Dir = "." }) orelse unreachable;
-    try std.testing.expect(preopen.@"type".eql(PreopenType{ .Dir = "." }));
-    try std.testing.expectEqual(@as(i32, 3), preopen.fd);
+    const preopen = preopens.find(PreopenType{ .Dir = "/cwd" }) orelse unreachable;
```
kubkon, line 227: So what's the difference between `.` and `/cwd`?

lib/std/os.zig
```diff
@@ -210,6 +219,13 @@ pub const LOG = struct {
     pub const DEBUG = 7;
 };
 
+pub const RelativePath = struct {
```
kubkon, line 222: This looks an WASI-only abstraction; should it not go into `wasi.zig` submodule? Perhaps together with preopens machinery?

kubkon (changes requested):
A couple of comments to get the ball rolling. I'm still reading over the changes, but I'll try to leave feedback incrementally to make it more manageable.

lib/std/fs/wasi.zig
topolarity: Oops, this is a leftover from some hacking - I'll clean it up!

lib/std/fs/wasi.zig
topolarity: It's purely a convention for the Preopens, but the primary advantage is that "/cwd" is an absolute path that supports `getcwd` and `realpath`. 

For instance, if the CWD is set to a preopen backed by "/cwd", then realpath("./some/file/path") will return "/cwd/some/file/path".

If the CWD is backed by just ".", then realpath has to error out instead.

lib/std/os.zig
topolarity: If I understood @andrewrk correctly on Discord, it sounds like only purely native functionality should be available in `std/os/wasi.zig`. This means that platform-specific code for implementing POSIX support has to stay in std/os.zig, just along with the existing platform-specific POSIX logic (e.g. openatWasi and openOptionsFromFlagsWindows)

Ideally, `std/os/wasi.zig` would exist for the purely native stuff, and then `std/posix/wasi.zig` could be a home for the WASI-specific implementation of POSIX-like behavior, but without #5019 we don't have that separation

Let me know if I got any of that wrong - Happy to fix it up if there's a mistake here :+1:

topolarity:
Thanks for the review!

Just pushed a small fix for a Windows-specific test failure and corrected the stylistic problems you pointed out.

lib/std/fs/wasi.zig
topolarity: There's an alternative I hadn't considered: `/preopen/fd/<N>/` could be used as a special absolute path corresponding to a preopen file descriptor.

realpath("./some/file/path") would return "/preopen/fd/\<N\>/some/file/path" and `resolvePathWasi` would check for the special `/preopen/fd/` prefix when resolving absolute paths, so that it can return the correct file descriptor. I'm not sure whether that's worth implementing right now, but I'm open to feedback 

In any case, this territory is a bit uncharted - wasi-libc doesn't currently emulate getcwd() or realpath()

kubkon: I like the sound of the latter; however, this seems like a hefty amount of work (unless I am mistaken) so we leave that out for a subsequent PR. Would you agree?

lib/std/os.zig
kubkon: Mhm, that's fair. But given that it's only used by WASI target, perhaps a more meaningful name such as `RelativePathWasi` or `WasiRelativePath` would make sense? Also, if you could drop a general doc-style comment for the struct - its intended use, etc. - that would great!

topolarity: Sure thing!

TBH, I think the clarity of the `*at` functions would be improved if they accepted a `RelativePath` rather than separate fd/path parameters related by name (`symlinkat` is particularly confusing). That's a separate change though and perhaps too far away from the standard POSIX interface, so I'll just document this as WASI-specific for now.

kubkon: Actually, that's a neat idea. Would you be willing to write it up as a proposal issue?

topolarity: Absolutely - I'll file something this evening when I come around to this PR again

kubkon (approved):
LGTM! Thanks for this excellent contribution @topolarity !

lib/std/fs/wasi.zig
topolarity: Hope it's not a bother to ask for a quick re-review  - I ended up getting this one done in [f357598](https://github.com/ziglang/zig/pull/11021/commits/f3575983ebd10b01fb9e4af427355f176ebd0ba6)

zigazeljko:
@topolarity How is `initPreopensWasi` supposed to be used -- in other words, what should `init_cwd` be set to to make path resolution work in a "normal" way?

zigazeljko:
To clarify my previous comment, let me give an example. Assume the program is given these preopens:

* fd `3` as `/foo`,
* fd `4` as `./foo`,
* fd `5` as `./bar`,
* fd `6` as `./bar/quux`.

Intuitively, the user would expect the path resolution to work like this:

* `/foo/aaa` resolves to `(fd=3, rel_path="aaa")`,
* `foo/bbb` resolves to `(fd=4, rel_path="bbb")`,
* `./foo/ccc` resolves to `(fd=4, rel_path="ccc")`,
* `./bar/ddd` resolves to `(fd=5, rel_path="ddd")`,
* `bar/eee/../quux/fff` resolves to `(fd=6, rel_path="eee/fff")`.

This is exactly how path resolution used to work in `wasi-libc`, before they introduced CWD emulation. If I understand the code correctly, such behavior is not possible with the implementation in this PR.

When `init_cwd` is `null`, `resolvePathAndGetWasiPreopen` behaves as if CWD was set to `"/"`, which means:

* `foo/bbb` is treated as `/foo/bbb` and resolves to `(fd=3, rel_path="bbb")`,
* `./foo/ccc` is treated as `/foo/ccc` and resolves to `(fd=3, rel_path="ccc")`,
* `./bar/ddd` and `bar/eee/../quux/fff` are not found.

When `init_cwd` is `"."`, `initPreopensWasi` fails, since there is no preopen for `"."`, only for its subdirectories.

When `init_cwd` is (let's say) `"./foo"`, `initPreopensWasi` works, but:

* `foo/bbb` resolves to `(fd=4, rel_path="foo/bbb")`,
* similarly for `./foo/ccc`, `./bar/ddd`, and `bar/eee/../quux/fff`,
* other CWD-relative preopens are not reachable with normal paths.

topolarity:
Thanks for the detailed input @zigazeljko.

I'm planning on opening an additional PR this evening to bring us up to parity with the "wasi-libc" behavior. `initPreopensWasi(..., null)` will give you the wasi-libc behavior by default. Under those rules, "./foo", "/foo", and "foo" preopens would be equivalent - the highest `fd` would take priority. This behavior is necessary to support wasmer-style Preopens, which are prefix-stripped by the runtime (no leading `.` or `/`). Nested preopens will need to be fixed upin that PR, too.

The only bit I'm unsure about is whether we can achieve the "intuitive" behavior of distinguishing `./foo/bar` vs. `/foo/bar/`, especially since some runtimes (`wasmer`) do not support this. 

I'm going to cc you there so that we can discuss in more detail :-)

zigazeljko:
> which are prefix-stripped by the runtime (no leading `.` or `/`)

Is this documented anywhere and/or have you tested it?

I just tested this in wasmer (versions 1.0.0 and 2.2.0) and while leading `/` is stripped, leading `.` is **not** stripped. For my example above, I get:

```
$ wasmer --mapdir=/foo:foo --mapdir=./foo:foo --mapdir=./bar:bar --mapdir=./bar/quux:quux preopens.wasm
3 -> /
4 -> foo
5 -> ./foo
6 -> ./bar
7 -> ./bar/quux
```

As far as I tested, wasmer only trims all leading `/`, but does no other path normalization (i.e. it keeps leading `.`, trailing `/`, repeated `/`, and `.` and `..` components).

Additionally, the first preopen (`3 -> /`) is always present, and is also the only preopen that can begin with a `/` (using `--mapdir=/:dir` gives you a preopen with an empty name).

topolarity:
You're right, I had only tested absolute paths. Thanks for the correction - That's an important detail.

Let's continue this discussion in the new PR, once its open?

PR #11024, opened 2022-03-01
stage2: Add limited WASI support for selfExePath and globalCacheDir

### LINKED ISSUE #10716: Compile Zig (compiler) to Wasm/WASI
It would be great to be able to compile Zig (the compiler) to the `wasm32` target.

From what I understood this are the required things to make it happen:
* One of:
     * [ ] Flock support in WASI, OR
     * [ ] Print warning instead of calling flock
* [ ] CWD support in WASI Zig (it's already supported in wasi-libc: https://github.com/WebAssembly/wasi-libc/pull/214 )
* One of:
    * [ ] Ship Zig just with the Zig syntax (no support for Clang at the beginning)
    * [ ] arocc merged upstream https://github.com/Vexu/arocc, OR
    * [ ] Compile LLVM to WASI https://github.com/binji/llvm-project (it's version 8.0, not sure if that would work or not, it should be easy to upgrade!)

topolarity:
Dependent on #11053. Resolves #10716.

This PR adds support for locating the Zig executable and the library and global cache directories, based on looking in the fixed "/zig" and "/cache" directories. It also bypasses the file-locking in "src/Cache.zig" and initializes the WASI preopens on startup.

The expected directories can be provided on the command-line using `--mapdir`, as follows:

```
wasmtime --mapdir=/cwd::. --mapdir=/cache::"$HOME/.cache/zig" --mapdir=/zig::./stage2-wasm/ ./stage2-wasm/bin/zig.wasm
```

The reason for the fixed directories is that argv[0] on WASI is just the basename (any absolute/relative path information is deleted by the runtime) and there is no WASI-equivalent to procfs.

This is enough to get stage2 running without LLVM on WASI:
```bash
$ ./build/zig build -Dtarget=wasm32-wasi --zig-lib-dir lib --prefix $(pwd)/stage2-wasm -Dsingle-threaded -Drelease
$ alias zigwasm="wasmtime --mapdir=/cwd::\$(pwd) --mapdir=/cache::\"$HOME/.cache/zig\" --mapdir=/zig::$(pwd)/stage2-wasm/ $(pwd)/stage2-wasm/bin/zig.wasm"
$ zigwasm build-exe ./simple_hello.zig -target x86_64-linux
$ chmod +x ./simple_hello
$ ./simple_hello
Hello, world!
```

This should be considered highly experimental - I haven't even checked the behavior tests or anything.

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -2547,6 +2547,15 @@ pub const SelfExePathError = os.ReadLinkError || os.SysCtlError || os.RealPathEr
 /// `selfExePath` except allocates the result on the heap.
 /// Caller owns returned memory.
 pub fn selfExePathAlloc(allocator: Allocator) ![]u8 {
+    if (builtin.os.tag == .wasi) {
+        var args = try std.process.argsWithAllocator(allocator);
+        defer args.deinit();
+        // On WASI, argv[0] is always just the basename of the current executable
+        const exe_name = args.next() orelse return error.FileNotFound;
+
+        var buf: [MAX_PATH_BYTES]u8 = undefined;
+        return allocator.dupe(u8, try selfExePathWasi(&buf, exe_name));
+    }
     // Use of MAX_PATH_BYTES here is justified as, at least on one tested Linux
     // system, readlink will completely fail to return a result larger than
     // PATH_MAX even if given a sufficiently large buffer. This makes it
@@ -2643,8 +2652,43 @@ pub fn selfExePath(out_buffer: []u8) SelfExePathError![]u8 {
             const end_index = std.unicode.utf16leToUtf8(out_buffer, utf16le_slice) catch unreachable;
             return out_buffer[0..end_index];
         },
+        .wasi => @compileError("std.fs.selfExePath not supported for WASI. Use std.fs.selfExePathAlloc instead."),
         else => @compileError("std.fs.selfExePath not supported for this target"),
     }
+}
+
+/// WASI-specific implementation of selfExePath
+///
+/// On WASI argv0 is always just the executable basename, so this function relies
+/// using a fixed executable directory path: "/zig"
+///
+/// This path can be configured in wasmtime using `--mapdir=/zig::/path/to/zig/dir/`
+fn selfExePathWasi(out_buffer: []u8, exe_name: []const u8) SelfExePathError![]const u8 {
+    var allocator = std.heap.FixedBufferAllocator.init(out_buffer);
+    var alloc = allocator.allocator();
+
+    // Check these paths:
+    //  1. "/zig/{exe_name}"
+    //  2. "/zig/bin/{exe_name}"
+    const base_paths_to_check = &[_][]const u8{ "/zig", "/zig/bin" };
+
+    for (base_paths_to_check) |base_path| {
+        const test_path = path.join(alloc, &.{ base_path, path.basename(exe_name) }) catch continue;
+
+        // Make sure it's a file we're pointing to
+        const file = os.fstatat(os.wasi.AT.FDCWD, test_path, 0) catch continue;
+        if (file.filetype != .REGULAR_FILE) continue;
+
+        // Path seems to be valid, let's try to turn it into an absolute path
+        var real_path_buf: [MAX_PATH_BYTES]u8 = undefined;
+        if (os.realpath(test_path, &real_path_buf)) |real_path| {
+            if (real_path.len > out_buffer.len)
+                return error.NameTooLong;
+            mem.copy(u8, out_buffer, real_path);
+            return out_buffer[0..real_path.len];
+        } else |_| continue;
+    }
+    return error.FileNotFound;
 }
 
 /// The result is UTF16LE-encoded.
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -2663,7 +2663,7 @@ pub fn selfExePath(out_buffer: []u8) SelfExePathError![]u8 {
 /// using a fixed executable directory path: "/zig"
 ///
 /// This path can be configured in wasmtime using `--mapdir=/zig::/path/to/zig/dir/`
-fn selfExePathWasi(out_buffer: []u8, exe_name: []const u8) SelfExePathError![]const u8 {
+fn selfExePathWasi(out_buffer: []u8, argv0: []const u8) SelfExePathError![]const u8 {
     var allocator = std.heap.FixedBufferAllocator.init(out_buffer);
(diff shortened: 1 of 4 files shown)


Jarred-Sumner:
The nice thing about WASI is it's not too hard to emulate a subset of it in browsers

topolarity:
I've hit what seems to be an instability in the CI tests a couple times now, related to the "POSIX file locking with fcntl".

Maybe the 1ms wait in that test is too low? Is this a known problem?

src/main.zig
```diff
@@ -162,6 +162,13 @@ pub fn main() anyerror!void {
         return mainArgs(gpa_tracy.allocator(), arena, args);
     }
 
+    // On WASI, the CWD and absolute paths are emulated w.r.t. the available preopens.
+    // Here we initialize that machinery and set the CWD to ".", where the user is
+    // expected to map their current directory using "--dir=."
```
matu3ba, line 167: Reordering the sentence makes it more understandable:
"WASI: The --dir= option instructs the web assmebly runtime to preopen a directory, and make it available to the program as a capability which can be used to open files inside that directory. The CWD and absolute paths are emulated w.r.t. the available preopens." [source](https://github.com/bytecodealliance/wasmtime/blob/main/docs/WASI-tutorial.md)

matu3ba (commented):
nit to reorder sentence and question on tests with recommended preopen

src/main.zig
```diff
@@ -162,6 +162,13 @@ pub fn main() anyerror!void {
...
+    if (builtin.os.tag == .wasi) {
+        try std.os.initPreopensWasi(std.heap.page_allocator, ".");
```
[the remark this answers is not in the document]
topolarity: Yeah, it is :+1:

I was trying to avoid the dependency on #11053 for now, but whichever PR lands last will need to update this to `os.initPreopensWasi(std.heap.page_allocator, "/preopens/cwd")`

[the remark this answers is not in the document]
matu3ba: Correction: https://github.com/WebAssembly/wasi-filesystem/issues/24#issuecomment-705621285 `.` and ´/` can be mapped differently: "(Ex: In Wasmer executables installed by the package manager come with a manifest; the author can request / and . to be mapped, creating a POSIX-like environment.)"

topolarity:
Chose to make this dependent on #11053 - Let's try to get that one merged first

src/Cache.zig
```diff
@@ -508,8 +508,10 @@ pub const Manifest = struct {
 
         // Next we will check the globally shared Cache timestamp, which is accessed
         // from multiple threads.
-        man.cache.mutex.lock();
-        defer man.cache.mutex.unlock();
+        if (builtin.os.tag != .wasi or !builtin.single_threaded) { // WASI does not support flock
```
kubkon, line 511: Should this not be

```suggestion
        if (builtin.os.tag != .wasi and !builtin.single_threaded) { // WASI does not support flock
```

If either WASI or single-threaded is true, we shouldn't need to lock anything, am I right?

kubkon (changes requested):

src/Cache.zig
topolarity: I think this is a mistake on my part - I probably thought this was locking the manifest file, but we can actually leave the code as-is. The standard library already ensures Mutex lock/unlock is a noop in single-threaded mode 🙂 

Pushing an update now.

(While at it, I also cleaned up the file locks to use `can_spawn` to determine whether `flock` is needed to contend with multiple processes.)

kubkon: Oh nice one! Thanks for investigating further and updating the PR!

kubkon (approved):

src/Cache.zig
```diff
@@ -762,18 +762,22 @@ pub const Manifest = struct {
...
     }
 
     fn upgradeToExclusiveLock(self: *Manifest) !void {
         if (self.have_exclusive_lock) return;
-        const manifest_file = self.manifest_file.?;
-        // Here we intentionally have a period where the lock is released, in case there are
-        // other processes holding a shared lock.
-        manifest_file.unlock();
-        try manifest_file.lock(.Exclusive);
+        if (std.process.can_spawn or !builtin.single_threaded) { // Some targets (WASI) do not support flock
```
andrewrk, line 774: This doesn't seem like it's checking the correct condition. Shouldn't it be something like "if the target supports file locking"?

Also I'm pretty sure file locking can be implemented in user space when it is not provided by the OS layer.

lib/std/fs.zig
```diff
@@ -2643,10 +2652,48 @@ pub fn selfExePath(out_buffer: []u8) SelfExePathError![]u8 {
...
     }
 }
 
+/// WASI-specific implementation of selfExePath
+///
+/// On WASI argv0 is always just the executable basename, so this function relies
+/// using a fixed executable directory path: "/zig"
+///
+/// This path can be configured in wasmtime using `--mapdir=/zig::/path/to/zig/dir/`
+fn selfExePathWasi(out_buffer: []u8, argv0: []const u8) SelfExePathError![]const u8 {
```
andrewrk, line 2666: I don't really think this belongs in the std lib. I think this should be solved in the compiler code not the std lib.

src/Cache.zig
topolarity: Just to explain the thinking: This check is intending to verify that we won't have multiple processes or threads contending for the manifest file, in which case `flock` can be safely elided.

The user-space implementations that I know of are a wrapper around locks provided by `fcntl`, which are also unavailable on WASI. AFAIK, there's no way on WASI to implement file locking yet (both child process spawning and file locking are on the roadmap for consideration, though)

lib/std/fs.zig
topolarity: I agree - Should have noticed that none of the other `selfExePath` functions here make Zig-specific assumptions.

Will add this to the list for follow-up.

src/Cache.zig
andrewrk: It's entirely possible to use a shell to execute two different processes that are each single threaded and cannot spawn subprocesses but they each try to access the same file from the file system, right?

topolarity: Yeah, it is - Sounds like it will be best to constrain this workaround to WASI-only, and add a note referencing the open wasi-filesystem issue where file locking is being considered

andrewrk:
> I tried making a build, but it seems to be broken now.

https://github.com/ziglang/zig/issues/13229

Fixed in a4eb221b9ef7dab8fa1c6cc07c7891e25d98d2b6

PR #11025, opened 2022-03-01
stage2: implement `@extern`

Vexu:


src/Air.zig
```diff
@@ -583,6 +583,10 @@ pub const Inst = struct {
         /// Uses the `ty_pl` field.
         field_parent_ptr,
 
+        /// Implements @extenr builtin.
```
joachimschmidt557, line 586:
```suggestion
        /// Implements @extern builtin.
```

small typo here

andrewrk:
Shouldn't this be lowered as a constant with `Value.Tag.decl_ref` (pointing to an extern Decl created by Sema) rather than introducing a new AIR instruction?

andrewrk:
Looks great

PR #11026, opened 2022-03-01
codegen: lower field_ptr to memory across linking backends

kubkon:
This requires generating an addend for the target relocation as
the field pointer might point at a field inner to the container.

### DIFF
--- a/src/arch/x86_64/CodeGen.zig
+++ b/src/arch/x86_64/CodeGen.zig
@@ -2565,7 +2565,8 @@ fn store(self: *Self, ptr: MCValue, value: MCValue, ptr_ty: Type, value_ty: Type
 
                     const payload = try self.addExtra(Mir.ImmPair{
                         .dest_off = 0,
-                        .operand = @intCast(u32, imm),
+                        // TODO check if this logic is correct
+                        .operand = @truncate(u32, imm),
                     });
                     const flags: u2 = switch (abi_size) {
                         1 => 0b00,
--- a/src/codegen.zig
+++ b/src/codegen.zig
@@ -142,11 +142,11 @@ pub fn generateFunction(
 
 pub fn generateSymbol(
     bin_file: *link.File,
-    parent_atom_index: u32,
     src_loc: Module.SrcLoc,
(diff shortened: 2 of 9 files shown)


Luukdegram:
This addition is great! The addition of addends makes `codegen.zig` Feature compatible with the Wasm backend’s ‘generateSymbol’. I’ll try to convert the Wasm backend to use this instead. It seems to only miss 128bit integers, floats and elem_ptr’s but perhaps I can add those in while I’m at it.

kubkon:
> This addition is great! The addition of addends makes `codegen.zig` Feature compatible with the Wasm backend’s ‘generateSymbol’. I’ll try to convert the Wasm backend to use this instead. It seems to only miss 128bit integers, floats and elem_ptr’s but perhaps I can add those in while I’m at it.

Sounds good, although last time I saw, this will require a significant rewrite in the Wasm codegen and linker as it's handled differently to how Wasm handles it. My suggestion would be to regress Wasm first and try using the `codegen` and `link` APIs and see how that pans out. Afterwards, we can add the rest.

Luukdegram:
> Sounds good, although last time I saw, this will require a significant rewrite in the Wasm codegen and linker as it's handled differently to how Wasm handles it. My suggestion would be to regress Wasm first and try using the `codegen` and `link` APIs and see how that pans out. Afterwards, we can add the rest.

Sounds great. Rather regress it now than after we get all tests passing :) (I had already started changing some of the linker API to make it a little more similar to the others).

kubkon:
> > Sounds good, although last time I saw, this will require a significant rewrite in the Wasm codegen and linker as it's handled differently to how Wasm handles it. My suggestion would be to regress Wasm first and try using the `codegen` and `link` APIs and see how that pans out. Afterwards, we can add the rest.
> 
> Sounds great. Rather regress it now than after we get all tests passing :) (I had already started changing some of the linker API to make it a little more similar to the others).

So here's an added bonus: if we manage to unify the API of Wasm with the other linking backends, you'll almost get DWARF for free. And that's a promise ;-)

Issue #10678, opened 2022-01-23
make std.fs absolute functions lower to the non-"at" variants of syscalls


synaptiko:
### Zig Version

0.9.0

### Steps to Reproduce

I have the following code:
```
const file = try std.fs.openFileAbsolute(list_path, .{ .read = true, .write = true });
defer file.close();
```

And the linux system has `tmpfs` enabled for `/tmp`.
```
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,nr_inodes=1048576,inode64)
```

The content of `list_path` is `/home/user/some-file`.

### Expected Behavior

Everything works fine, no need to worry about from where the command is being run.

### Actual Behavior

When the command which uses the above code is run from `/tmp` directory, it will end up with "rename: Invalid cross-device link" error and `/home/user/some-file` will disappear. The content can be recovered from `/tmp` itself, as there is some random file created.

IMHO this is quite unexpected behavior and it took me really long to figure out. I only realized what's going on when I checked the implementation of `std.fs.openFileAbsolute()` and saw that it in fact calls `cwd().openFile()`.

The workaround is to call `cd && command` instead.

andrewrk:
`cwd().openFile()` is just doing this:

```
openat(AT_FDCWD, "/home/user/some-file", O_RDWR|O_NOCTTY|O_CLOEXEC) = 3
```

Does `openat` with `AT_FDCWD` not work with tmpfs? Seems like a bug in tmpfs.

That said, I think it would be reasonable to expect `openFileAbsolute` to lower to the `open` syscall instead of `openat`.

I-mikan-I:
> FWIW glibc uses the -at functions all the time now (even the `open` libc function goes to `openat`)
> 
> Linux kernel developers (at one time at least) were saying that going forward, only `-at` variants would be available and to considered the non -at variants to be for compatibility only.

Thanks for the info, I tested if it makes a difference whether we use AT_FDCWD or not, and could not replicate the original issue myself.

Here's my application's source code, compiled with zig 0.9.0 in Debug and ReleaseFast:
```zig
pub fn main() anyerror!void {
    const file = try std.fs.openFileAbsolute("/home/mikan/test", .{.read = true, .write = true});
    defer file.close();
    try file.writeAll("Hello, World!\n");
}
```

[...]

PR #11027, opened 2022-03-01
use open syscall when using an absolute path

I-mikan-I:
With these changes, opening files on linux using `openFileAbsolute` will translate to an `open` syscall instead of `openat`.

With how this is implemented, using `openFile` will also translate to open, when the path is absolute.
Since I am not sure if this was undesired or not, I implemented it in this way anyway. 
If only calls to `openFileAbsolute` should use `open`, we can add a new function, although some replication of boilerplate source code would probably be required.

Since this is my first PR, any feedback is very welcome :).

Closes: #10678

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -1023,17 +1023,21 @@ pub const Dir = struct {
         // Use open syscall instead of openat when we are dealing with absolute paths,
         // because openat with AT_FDCWD does not work with tmpfs.
         const is_absolute = path.isAbsoluteZ(sub_path);
-        const fd = if (flags.intended_io_mode != .blocking)
-            if (is_absolute) 
-                try std.event.Loop.instance.?.openZ(sub_path, os_flags, 0)
-            else 
-                try std.event.Loop.instance.?.openatZ(self.fd, sub_path, os_flags, 0)
-            
-        else
-            if (is_absolute) 
-                try os.openZ(sub_path, os_flags, 0)
-            else
-                try os.openatZ(self.fd, sub_path, os_flags, 0);
+        const fd = blk: {
+            if (flags.intended_io_mode != .blocking) {
+                if (is_absolute) {
+                    break :blk try std.event.Loop.instance.?.openZ(sub_path, os_flags, 0);
+                } else {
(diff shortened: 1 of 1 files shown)


I-mikan-I:
`openFileW` uses [ntdll](https://docs.microsoft.com/en-us/windows/win32/api/winternl/nf-winternl-ntcreatefile) and already sets the root directory attribute to `NULL` when an absolute file is passed: [snippet](https://github.com/ziglang/zig/blob/b90bfc8de673ea0ceb56184476bc29e42afbeaa7/lib/std/os/windows.zig#L91).

`openFileWasi` uses `path_open`, and I don't think there's an `open` equivalent.

matu3ba:
@I-mikan-I Did you check the wasi sources? https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md
I dont see anything beside `path_open`.

ifreund:
I don't think there's a good reason for this complexity increase, at least not on linux. In the linux kernel implementation `open()` and `open_at(AT_FDCWD)` are 100% identical.
```c
SYSCALL_DEFINE3(open, const char __user *, filename, int, flags, umode_t, mode)
{
	if (force_o_largefile())
		flags |= O_LARGEFILE;

	return do_sys_open(AT_FDCWD, filename, flags, mode);
}

SYSCALL_DEFINE4(openat, int, dfd, const char __user *, filename, int, flags,
		umode_t, mode)
{
	if (force_o_largefile())
		flags |= O_LARGEFILE;

	return do_sys_open(dfd, filename, flags, mode);
}
```

I-mikan-I:
Since I could not reproduce the original issue when trying out a different approach, I will convert this PR to a draft.

I don't believe the issue is caused by using `openat`, and perhaps we need to do some more testing to identify the cause.  
Has anyone else encountered this issue before?

Vexu:
> Since I could not reproduce the original issue when trying out a different approach, I will convert this PR to a draft.
> 
> I don't believe the issue is caused by using `openat`, and perhaps we need to do some more testing to identify the cause. Has anyone else encountered this issue before?

Let's keep the discussion in the relevant issue and leave the pull request queue for changes that are ready to be merged.

PR #11029, opened 2022-03-02
stage2: reify error sets

mitchellh:
`@Type()` for error sets.

This also fixes a bug in @typeInfo for error sets where I wasn't wrapping the payload properly. This test also catches that.

Question: I create an `error_set_merged` since that only requires a name map. Should I be creating a full `error_set` decl for this? That requires a full `Module.ErrorSet`. I'm sure I can do that but wasn't sure what you wanted here.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -10397,7 +10397,13 @@ fn zirTypeInfo(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Ai
                         vals,
                     ),
                 );
-                break :v try Value.Tag.decl_ref.create(sema.arena, new_decl);
+
+                const new_decl_val = try Value.Tag.decl_ref.create(sema.arena, new_decl);
+                const slice_val = try Value.Tag.slice.create(sema.arena, .{
+                    .ptr = new_decl_val,
+                    .len = try Value.Tag.int_u64.create(sema.arena, vals.len),
+                });
+                break :v try Value.Tag.opt_payload.create(sema.arena, slice_val);
             } else Value.@"null";
 
             // Construct TypeInfo{ .ErrorSet = errors_val }
@@ -12317,7 +12323,31 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
             });
             return sema.addType(ty);
         },
-        .ErrorSet => return sema.fail(block, src, "TODO: Sema.zirReify for ErrorSet", .{}),
+        .ErrorSet => {
+            const payload_val = union_val.val.optionalValue() orelse
+                return sema.addType(Type.initTag(.anyerror));
+            const slice_val = payload_val.castTag(.slice).?.data;
+            const decl = slice_val.ptr.castTag(.decl_ref).?.data;
+            try sema.ensureDeclAnalyzed(decl);
+            const array_val = decl.val.castTag(.array).?.data;
+
+            var names: Module.ErrorSet.NameMap = .{};
+            try names.ensureUnusedCapacity(sema.arena, array_val.len);
+            for (array_val) |elem_val| {
+                const struct_val = elem_val.castTag(.@"struct").?.data;
+                // TODO use reflection instead of magic numbers here
+                // error_set: type,
+                const name_val = struct_val[0];
+
+                names.putAssumeCapacityNoClobber(
+                    try name_val.toAllocatedBytes(Type.initTag(.const_slice_u8), sema.arena),
+                    {},
+                );
+            }
+
+            const ty = try Type.Tag.error_set_merged.create(sema.arena, names);
+            return sema.addType(ty);
+        },
         .Enum => return sema.fail(block, src, "TODO: Sema.zirReify for Enum", .{}),
         .Union => return sema.fail(block, src, "TODO: Sema.zirReify for Union", .{}),
         .Fn => return sema.fail(block, src, "TODO: Sema.zirReify for Fn", .{}),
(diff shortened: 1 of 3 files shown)


src/Sema.zig
```diff
@@ -12189,7 +12195,35 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
             });
             return sema.addType(ty);
         },
-        .ErrorSet => return sema.fail(block, src, "TODO: Sema.zirReify for ErrorSet", .{}),
+        .ErrorSet => switch (union_val.val.tag()) {
```
andrewrk, line 12198: Check out `zirOptionalPayload` for an example of how to handle the comptime optional here:

https://github.com/ziglang/zig/blob/6f303c01f3e06fe8203563065ea32537f6eff456/src/Sema.zig#L5489-L5497

src/Sema.zig
```diff
@@ -12189,7 +12195,35 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
...
+                        try name_val.toAllocatedBytes(Type.initTag(.const_slice_u8), sema.arena),
+                        {},
+                    );
+                }
+
+                // TODO: should this be an error_set with a full decl?
+                const ty = try Type.Tag.error_set_merged.create(sema.arena, names);
+                return sema.addType(ty);
+            },
+            else => return sema.fail(block, src, "ErrorSet invalid value type", .{}),
```
andrewrk, line 12225: This is an internal compiler error - effectively an assertion failure - so it shouldn't use the normal compile error mechanism.

```suggestion
            else => unreachable,
```

If we wanted to make a more fail-safe build of the compiler we would probably do it by using -OReleaseSafe and having a custom panic handler that makes an error report. In such case the above `unreachable` is all it would take to trigger the appropriate mechanism.

But I think this comment becomes irrelevant after the other comment above is addressed.

andrewrk (changes requested):
Nice work. I have a requested change that I think will be worth doing before merging.

> Question: I create an `error_set_merged` since that only requires a name map. Should I be creating a full `error_set` decl for this? That requires a full `Module.ErrorSet`. I'm sure I can do that but wasn't sure what you wanted here.

Just the `error_set_merged` is best for now. In the future we may want to complicate it so that we can trace compile errors back to which decls errors came from, and attached doc comments for generated documentation or perhaps IDE integration, but for now sticking with the simplest thing that works is the way to go. Easy to add these mechanisms later when we are focusing directly on them.

src/Sema.zig
mitchellh: I took a look, I'm unsure exactly what you wanted me to do. See the added commit but I'm not sure this is what you intended. Sorry for the confusion, if you can clarify or reword I'll try again.

src/Sema.zig
```diff
@@ -12189,7 +12195,33 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
             });
             return sema.addType(ty);
         },
-        .ErrorSet => return sema.fail(block, src, "TODO: Sema.zirReify for ErrorSet", .{}),
+        .ErrorSet => {
+            if (union_val.val.isNull())
+                return sema.addType(Type.initTag(.anyerror));
+
+            const payload_val = union_val.val.castTag(.opt_payload).?;
+            const slice_val = payload_val.data.castTag(.slice).?.data;
```
andrewrk, line 12203: This change takes us almost there- only one more thing left to do:

```suggestion
            const payload_val = if (union_val.val.castTag(.opt_payload)) |p| p.data else union_val.val;
            const slice_val = payload_val.castTag(.slice).?.data;
```

Optionals can be represented both ways. This should probably be abstracted into a helper method just like `isNull` so that Sema code does not rely too heavily on checking tag(). (that's not going to block merging this though)

andrewrk (changes requested):
One last request and then this is good to go as soon as it passes the tests.

src/Sema.zig
mitchellh: Oooooooh. This explains why my "bug fix" in this PR also didn't cause issues unless it was deeply introspected using the raw values/types. Okay. Yes, I understand. I'll poke around and might do the helper too.

andrewrk: For some more context: comptime value representation upholds the property that if `coerceInMemoryAllowed` returns `ok` then it is safe to simply change the Type while keeping the Value the same. In other words `coerceInMemoryAllowed` applies to both runtime representation and comptime representation! As a consequence, optional pointer types such as `?*i32` support the same Value representation as non-optional pointers such as `*i32`.

mitchellh: Makes sense. Thank you. The change is in 👍

src/Sema.zig
```diff
@@ -12189,7 +12195,31 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
             });
             return sema.addType(ty);
         },
-        .ErrorSet => return sema.fail(block, src, "TODO: Sema.zirReify for ErrorSet", .{}),
+        .ErrorSet => {
+            if (union_val.val.optionalValue()) |payload_val| {
```
andrewrk, line 12199:
```suggestion
            const payload_val = union_val.val.optionalValue() orelse
                return sema.addType(Type.initTag(.anyerror));
```

andrewrk (approved):
logic looks right! this is good to merge when tests pass.

I left one piece of style feedback but it's not important.

src/Sema.zig
mitchellh: Thanks, made that change! Makes sense.

andrewrk:
Tests passed for me locally.

PR #11036, opened 2022-03-02
x64: handle more optional types

kubkon:
* codegen: lower optionals and floats across linking backends
* x64: rectify and missing optionals bits - `wrap_optional`, `optional_payload` and `isNull` helper
* x64: fix `int_cast` to properly clear out the destination register before copying the value over
* x64: fix incorrect calculation of spill stack location of the `rdi` register for backpatching the return value write locations

### DIFF
--- a/src/codegen.zig
+++ b/src/codegen.zig
@@ -141,26 +141,13 @@ pub fn generateFunction(
 }
 
 fn writeFloat(comptime F: type, f: F, target: Target, endian: std.builtin.Endian, code: []u8) void {
-    if (F == f80) {
-        switch (target.cpu.arch) {
-            .i386, .x86_64 => {
-                const repr = math.break_f80(f);
-                mem.writeIntLittle(u64, code[0..8], repr.fraction);
-                mem.writeIntLittle(u16, code[8..10], repr.exp);
-                // TODO set the rest of the bytes to undefined. should we use 0xaa
-                // or is there a different way?
-                return;
-            },
-            else => {},
-        }
-    } else {
-        const Int = @Type(.{ .Int = .{
-            .signedness = .unsigned,
-            .bits = @typeInfo(F).Float.bits,
-        } });
-        const int = @bitCast(Int, f);
-        mem.writeInt(Int, code[0..@sizeOf(Int)], int, endian);
-    }
+    _ = target;
+    const Int = @Type(.{ .Int = .{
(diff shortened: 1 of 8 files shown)


src/codegen.zig
```diff
@@ -140,6 +140,29 @@ pub fn generateFunction(
     }
 }
 
+fn writeFloat(comptime F: type, f: F, target: Target, endian: std.builtin.Endian, code: []u8) void {
+    if (F == f80) {
```
Vexu, line 144: I think this should be handled same as every other float.

src/codegen.zig
```diff
@@ -171,6 +196,18 @@ pub fn generateSymbol(
                 ),
             };
         },
+        .Float => {
+            const float_bits = typed_value.ty.floatBits(target);
+            switch (float_bits) {
+                16 => writeFloat(f16, typed_value.val.toFloat(f16), target, endian, try code.addManyAsArray(2)),
+                32 => writeFloat(f32, typed_value.val.toFloat(f32), target, endian, try code.addManyAsArray(4)),
+                64 => writeFloat(f64, typed_value.val.toFloat(f64), target, endian, try code.addManyAsArray(8)),
+                80 => writeFloat(f80, typed_value.val.toFloat(f80), target, endian, try code.addManyAsArray(10)),
```
Vexu, line 205: Shouldn't the 10 match `f80`'s ABI size (12 on i386, 16 elsewhere)?

src/codegen.zig
kubkon: As below, the source is here: https://github.com/ziglang/zig/blob/7cfc3f0cfa626abe25c8318a7852977cbc1c723b/src/value.zig#L1273 I actually don't really know yet how to handle `f80` at all. I can leave is a todo instead then.

kubkon: `f80` is now explicitly unhandled in b0bb158. Happy to come back to this discussion when I start working on floats in x64. Until then, lowering floats was an unfortunate prerequisite to not regress comptime tests it seems.

src/codegen.zig
kubkon: `f80` is now explicitly unhandled in b0bb158. Happy to come back to this discussion when I start working on floats in x64. Until then, lowering floats was an unfortunate prerequisite to not regress comptime tests it seems.

PR #11038, opened 2022-03-02
codegen: handle elem_ptr when lowering to memory

kubkon:
* x64: handle storing from-to non-stack memory

502 tests passing on x64! :tada: 
image

### DIFF
--- a/src/codegen.zig
+++ b/src/codegen.zig
@@ -392,6 +392,30 @@ pub fn generateSymbol(
                     },
                 }
             },
+            .elem_ptr => {
+                const elem_ptr = typed_value.val.castTag(.elem_ptr).?.data;
+                const elem_size = typed_value.ty.childType().abiSize(target);
+                const addend = @intCast(u32, elem_ptr.index * elem_size);
+                const array_ptr = elem_ptr.array_ptr;
+
+                switch (array_ptr.tag()) {
+                    .decl_ref => {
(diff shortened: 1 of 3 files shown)


src/codegen.zig
```diff
@@ -392,6 +392,30 @@ pub fn generateSymbol(
                     },
                 }
             },
+            .elem_ptr => {
+                const elem_ptr = typed_value.val.castTag(.elem_ptr).?.data;
+                const elem_size = typed_value.ty.childType().abiSize(target);
+                const addend = @intCast(u32, elem_ptr.index * elem_size);
+                const array_ptr = elem_ptr.array_ptr;
+
+                switch (array_ptr.tag()) {
```
andrewrk, line 401: we need some kind of recursive thing here. this could be another elem_ptr, a field_ptr, or other kind of pointer.

check out what we do in the llvm backend, specifically the `lowerParentPtr` function:

https://github.com/ziglang/zig/blob/446324a1d82514157f5a11a6ca775711bc48c419/src/codegen/llvm.zig#L1360-L1380

kubkon: Yep, correct, but one step at a time. Before I recurse, I gotta see the examples ;-)

kubkon: Oh, you can be sure I shall be cleaning this up soon-ish when I have the bigger picture firmly lodged in my head.

PR #11039, opened 2022-03-02
stage2: implement @errSetCast

mitchellh:
This adds `@errSetCast` for both comptime and runtime. Safety checks are TODO.

This also adds value hashing for error sets and error unions.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -12460,7 +12460,8 @@ fn zirErrSetCast(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!
         return sema.addConstant(dest_ty, val);
     }
 
-    return sema.fail(block, src, "TODO: implement runtime errSetCast", .{});
+    try sema.requireRuntimeBlock(block, src);
+    return block.addTyOp(.errcast, dest_ty, operand);
 }
 
 fn zirPtrCast(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.Inst.Ref {
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -12417,7 +12417,50 @@ fn zirIntToPtr(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Ai
 fn zirErrSetCast(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.Inst.Ref {
     const inst_data = sema.code.instructions.items(.data)[inst].pl_node;
     const src = inst_data.src();
-    return sema.fail(block, src, "TODO: Sema.zirErrSetCast", .{});
+    const dest_ty_src: LazySrcLoc = .{ .node_offset_builtin_call_arg0 = inst_data.src_node };
+    const operand_src: LazySrcLoc = .{ .node_offset_builtin_call_arg1 = inst_data.src_node };
+    const extra = sema.code.extraData(Zir.Inst.Bin, inst_data.payload_index).data;
+
+    const dest_ty = try sema.resolveType(block, dest_ty_src, extra.lhs);
+    if (dest_ty.zigTypeTag() != .ErrorSet) {
+        return sema.fail(
+            block,
+            dest_ty_src,
+            "expected error set type, found '{}'",
+            .{dest_ty},
+        );
+    }
+
+    const operand = sema.resolveInst(extra.rhs);
+    const operand_ty = sema.typeOf(operand);
+    if (operand_ty.zigTypeTag() != .ErrorSet) {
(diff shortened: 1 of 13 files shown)


src/Sema.zig
```diff
@@ -12417,7 +12417,55 @@ fn zirIntToPtr(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Ai
 fn zirErrSetCast(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.Inst.Ref {
     const inst_data = sema.code.instructions.items(.data)[inst].pl_node;
     const src = inst_data.src();
-    return sema.fail(block, src, "TODO: Sema.zirErrSetCast", .{});
+    const dest_ty_src: LazySrcLoc = .{ .node_offset_builtin_call_arg0 = inst_data.src_node };
+    const operand_src: LazySrcLoc = .{ .node_offset_builtin_call_arg1 = inst_data.src_node };
+    const extra = sema.code.extraData(Zir.Inst.Bin, inst_data.payload_index).data;
+
+    const dest_ty = try sema.resolveType(block, dest_ty_src, extra.lhs);
+    if (dest_ty.zigTypeTag() != .ErrorSet) {
```
andrewrk, line 12425: There's a pattern here we can follow to slightly reduce duplication, see some of these other functions for example:

```
fn checkIntType(
fn checkPtrType(
fn checkVectorElemType(
fn checkFloatType(
fn checkNumericType(
fn checkAtomicOperandType(
fn checkIntOrVector(
```

Looks like operand_ty below wants the exact same check.

src/Sema.zig
```diff
@@ -12417,7 +12417,55 @@ fn zirIntToPtr(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Ai
...
+    if (operand_ty.zigTypeTag() != .ErrorSet) {
+        return sema.fail(
+            block,
+            operand_src,
+            "expected error set type, found '{}'",
+            .{operand_ty},
+        );
+    }
+
+    if (try sema.resolveMaybeUndefVal(block, operand_src, operand)) |val| {
```
andrewrk, line 12445: This should be `resolveDefinedValue` because it is a compilation error to try to `@errSetCast` an undefined value:

```suggestion
    if (try sema.resolveDefinedValue(block, operand_src, operand)) |val| {
```

src/Sema.zig
```diff
@@ -12417,7 +12417,55 @@ fn zirIntToPtr(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Ai
...
+
+        return sema.addConstant(dest_ty, val);
+    }
+
+    try sema.requireRuntimeBlock(block, src);
+    if (block.wantSafety()) {
+        // TODO
+    }
+
+    return operand;
```
andrewrk, line 12468: The type of the returned Air.Inst.Ref of this function should be dest_ty. Here it will incorrectly have the same error set as the operand (hopefully the behavior test would catch this).

test/behavior/error.zig
```diff
@@ -208,8 +208,29 @@ fn testErrorSetType() !void {
     }
 }
 
+test "error set cast at comptime" {
```
andrewrk, line 211: no need for this extra test. `comptime try testExplicitErrorSetCast(Set1.A); ` below is already doing a comptime version of the same test.

andrewrk (changes requested):
Here are a couple lines you can add to the existing behavior tests to catch the mistake in `zirErrSetCast`.

```diff
--- a/test/behavior/error.zig
+++ b/test/behavior/error.zig
@@ -220,7 +220,9 @@ const Set2 = error{ A, C };
 
 fn testExplicitErrorSetCast(set1: Set1) !void {
     var x = @errSetCast(Set2, set1);
+    try expect(@TypeOf(x) == Set2);
     var y = @errSetCast(Set1, x);
+    try expect(@TypeOf(y) == Set1);
     try expect(y == error.A);
 }
 ```

test/behavior/error.zig
mitchellh: Yeah I thought so too, but it wasn't because of the `var` usage? Maybe that's a bug in stage2. I'll remove this test for now.

src/Sema.zig
mitchellh: Thanks, got it.

mitchellh:
Feedback addressed. Thank you!

andrewrk (approved):
Looks great :+1:

andrewrk:
CI failure is my fault, introduced in master branch in 1c8a86f063b65e95c2e23ceaa40843069adfdc23. I'm on it.

PR #11043, opened 2022-03-03
stage2: Improve `@ptrCast` support for sliced/optional operands

topolarity:
Various changes to fix some small tests in `behavior/optional.zig` and `behavior/ptrcast.zig`

**Aside:** The last remaining test in ptrcast seems quite tricky: A field_ptr fails to be dereferenced because Sema.beginComptimePtrLoad assumes that it can recurse to a root decl_ref and ask for its type to get the parent struct/array type. However, this seems incorrect for reinterpreted memory, where a pointer intentionally disagrees about its target vs. the containing decl

### DIFF
--- a/src/type.zig
+++ b/src/type.zig
@@ -5287,8 +5287,10 @@ pub const Type = extern union {
         // type, we change it to 0 here. If this causes an assertion trip because the
         // pointee type needs to be resolved more, that needs to be done before calling
         // this ptr() function.
-        if (d.@"align" != 0 and d.@"align" == d.pointee_type.abiAlignment(target)) {
-            d.@"align" = 0;
+        if (d.@"align" != 0) {
+            if (d.pointee_type.zigTypeTag() == .Opaque or d.@"align" == d.pointee_type.abiAlignment(target)) {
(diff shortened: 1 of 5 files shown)


src/type.zig
```diff
@@ -5287,8 +5287,10 @@ pub const Type = extern union {
         // type, we change it to 0 here. If this causes an assertion trip because the
         // pointee type needs to be resolved more, that needs to be done before calling
         // this ptr() function.
-        if (d.@"align" != 0 and d.@"align" == d.pointee_type.abiAlignment(target)) {
-            d.@"align" = 0;
+        if (d.@"align" != 0) {
+            if (d.pointee_type.zigTypeTag() == .Opaque or d.@"align" == d.pointee_type.abiAlignment(target)) {
+                d.@"align" = 0;
+            }
```
andrewrk, line 5293: This will make `*align(N) T == *T` when T is an opaque type, which is incorrect. Perhaps instead this should be reverted and abiAlignment should return 1 for opaque types?

andrewrk (changes requested):
I identified one problem with opaque pointers-

PR #11044, opened 2022-03-03
std.BoundedArray: return explicit errors

motiejus:
Makes it easier to mark explicit errors when using BoundedArray
downstream.

Per @jedisct1 confirmation, also removing `IndexOutOfBounds` in favor of `Overflow`, making it the only error this module can return.


motiejus:
(happy to submit another PR if you agree that it should)

jedisct1:
Thanks!

Returning `Overflow` instead of `IndexOutOfBounds` in `insert()` makes sense.

Would you mind adding that change to the PR?

PR #11048, opened 2022-03-03
stage2: Implement wasm builtins

Luukdegram:
This implements the [@wasmMemorySize](https://ziglang.org/documentation/master/#wasmMemorySize) and [@wasmMemoryGrow](https://ziglang.org/documentation/master/#wasmMemoryGrow) builtins. Note that the implementation differs from stage1 whereas in stage2 it ensures the `index` argument is a comptime-known value. This is required, as the index is part of the instruction encoding, and therefore cannot be a runtime value. For stage1, this is only enforced by AstGen whereas in stage2 Sema also resolves it as a comptime value. I figured it wouldn't be worth the effort to correct this in stage1 when AstGen already enforces it there now as well. (This does however make this PR a breaking change).

Unfortunately, I have to perform quite some fixes within the self-hosted linker to ensure the LLVM backend can target wasm. For this reason, I'm leaving this builtin as a TODO for the LLVM backend. I plan on revisiting this soon.

@andrewrk I would appreciate it if you could review this PR to ensure it's in line with your thoughts.

### DIFF
--- a/src/Air.zig
+++ b/src/Air.zig
@@ -584,11 +584,11 @@ pub const Inst = struct {
         field_parent_ptr,
 
         /// Implements @wasmMemorySize builtin.
-        /// Uses the `ty_pl` field, payload is `WasmMemoryIndex`.
+        /// Uses the `ty_pl` field, payload represents the index of the target memory.
         wasm_memory_size,
 
         /// Implements @wasmMemoryGrow builtin.
-        /// Uses the `pl_op` field, payload is `WasmMemoryIndex`.
+        /// Uses the `pl_op` field, payload represents the index of the target memory.
         wasm_memory_grow,
 
         pub fn fromCmpOp(op: std.math.CompareOperator) Tag {
@@ -723,12 +723,6 @@ pub const Bin = struct {
 pub const FieldParentPtr = struct {
     field_ptr: Inst.Ref,
     field_index: u32,
-};
-
-/// Wasm's memory instructions require a comptime-known index
-/// which represents the memory it operates on.
-pub const WasmMemoryIndex = struct {
-    index: u32,
 };
 
 /// Trailing:
--- a/src/Air.zig
+++ b/src/Air.zig
@@ -584,10 +584,13 @@ pub const Inst = struct {
         field_parent_ptr,
 
         /// Implements @wasmMemorySize builtin.
-        /// Uses the `ty_pl` field, payload represents the index of the target memory.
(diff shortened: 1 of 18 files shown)


src/Air.zig
```diff
@@ -583,6 +583,14 @@ pub const Inst = struct {
         /// Uses the `ty_pl` field.
         field_parent_ptr,
 
+        /// Implements @wasmMemorySize builtin.
+        /// Uses the `ty_pl` field, payload is `WasmMemoryIndex`.
```
Vexu, line 587: It should be ok to directly put the index in the `payload` field like `constant` does.

src/codegen/c.zig
```diff
@@ -1758,6 +1758,9 @@ fn genBody(f: *Function, body: []const Air.Inst.Index) error{ AnalysisFail, OutO
             .wrap_errunion_payload       => try airWrapErrUnionPay(f, inst),
             .wrap_errunion_err           => try airWrapErrUnionErr(f, inst),
             .errunion_payload_ptr_set    => try airErrUnionPayloadPtrSet(f, inst),
+
+            .wasm_memory_size => unreachable,
+            .wasm_memory_grow => unreachable,
```
Vexu, line 1763: These aren't unreachable, you can do `-target wasm32-wasi -ofmt=c`.

src/Air.zig
Luukdegram: I thought of that, but I felt like I was abusing the intent of the `ty_pl` & `pl_op` instructions. Considering we're already doing this for constants, I'll apply your suggestion instead. Thanks!

src/codegen/c.zig
Luukdegram: @Vexu Would you mind reviewing the changes to CBE with regards to this?

test/behavior.zig
```diff
@@ -96,6 +96,10 @@ test {
     _ = @import("behavior/void.zig");
     _ = @import("behavior/while.zig");
 
+    if (builtin.zig_backend == .stage2_wasm) {
```
andrewrk, line 99: I suggest to delete the other import in this file and do this:

```suggestion
    if (builtin.target.cpu.arch == .wasm32) {
```

andrewrk (changes requested):
Looks great! I would be happy to make the LLVM enhancements as well as the requested changes below, just give me the word, and I'll take it from here.

PR #11053, opened 2022-03-04
stdlib std.os: Update CWD emulation to match wasi-libc behavior

topolarity:
Dependent on #11021 .

Some more work is needed to align our CWD treatment on WASI with wasi-libc. This is particularly important for runtimes like wasmer, which inserts a special "/" Preopen by default and also strips leading slashes from other Preopens

Here's the idea behind these changes:
1. We store the CWD as a simple `[]const u8`. We lookup Preopens for every absolute or CWD-referenced file operation, based on the Preopen with the longest match (i.e. most specific path)
2. Preorders are normalized to POSIX absolute paths at init time, based on the `cwd_root` parameter of `initPreopensWasi`.  cwd_root is used _only_ for any Preopens that start with "." so that:
```
   "./foo/bar" ~ canonicalizes to ~> "{cwd_root}/foo/bar"
   "foo/bar"   ~ canonicalizes to ~> "/foo/bar"
   "/foo/bar"  ~ canonicalizes to ~> "/foo/bar"
```

The intention is that:
 - Using "/" as `cwd_root` gives behavior similar to wasi-libc
 - Applications can provide a unique prefix, such as "/preopens/cwd" to prevent name collisions, esp. with the "." Preopen

These changes come at a performance cost, since we no longer cache any matching Preopen for the CWD, but the implementation is also much simpler and the compatibility gain from matching wasi-libc's behavior seems worth it.

Another consequence is that we no longer support `fchdir` on WASI

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -11279,7 +11279,7 @@ pub fn main() !void {
     var preopens = PreopenList.init(gpa);
     defer preopens.deinit();
 
-    try preopens.populate();
+    try preopens.populate(null);
 
     for (preopens.asSlice()) |preopen, i| {
         std.debug.print("{}: {}\n", .{ i, preopen });
--- a/lib/std/fs/path.zig
+++ b/lib/std/fs/path.zig
@@ -735,7 +735,7 @@ pub fn resolvePosix(allocator: Allocator, paths: []const []const u8) ![]u8 {
 
 test "resolve" {
     if (native_os == .wasi and builtin.link_libc) return error.SkipZigTest;
-    if (native_os == .wasi and !builtin.link_libc) try os.initPreopensWasi(std.heap.page_allocator, ".");
+    if (native_os == .wasi and !builtin.link_libc) try os.initPreopensWasi(std.heap.page_allocator, "/");
 
     const cwd = try process.getCwdAlloc(testing.allocator);
     defer testing.allocator.free(cwd);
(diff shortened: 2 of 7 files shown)


topolarity:
cc @zigazeljko

Can you take a peek at b6ea3b27345780b92d9a8c5cda2a7361652f224e here and see what you think of the new approach? I'm open to any alternatives or concerns that might jump out to you.

zigazeljko:
Looks good to me.

kubkon:
#11021 is now merged - would you mind rebasing onto master please?

matu3ba:
1. Question for edge case: Does `"../foo/bar" ~ canonicalize to ~> "{cwd_root}/foo/bar"` or is it invalid?
A test for this behavior would be useful.

2. Does this also solve the shortcoming of `realpath` requiring a prefixed `/` or will this be a separate commit? see #10988

3. Can you add to the doc comments of function `realpath` `assume: relative path is given from from cwd() of process` ?
And `assume: `/` followed by file description can be `cd`ed to` ?

topolarity:
> 1. Question for edge case: Does `"../foo/bar" ~ canonicalize to ~> "{cwd_root}/foo/bar"` or is it invalid/creates an error?

A very good question.  This would be canonicalized as `{cwd_root}/../foo/bar` after the substitution of "cwd_root". So if cwd_root is "/", this resolves to "/foo/bar". If cwd_root is "/preopens/cwd", this resolves to "/preopens/foo/bar".

Whether this behavior is "right" probably depends on how https://github.com/WebAssembly/wasi-filesystem/issues/24 is resolved. I don't have a strong argument either way, right now

> 2. Does this also solve the shortcoming of `realpath` requiring a prefixed `/` or will this be a separate commit? see [Add assumptions for `resolveWindows`, `resolvePosix`and `realpath` #10988](https://github.com/ziglang/zig/pull/10988)

The WASI implementation for `realpath` added here does not require a prefixed `/`, and the resolve always includes an absolute path (i.e. the absolute CWD)

> 3. Can you add to the doc comments of function `realpath` "assume: relative path is given from from cwd() of process" ?
>    And "assume: `/` followed by file description can be `cd`ed to" ?

That seems like it should be tackled in your existing PR #10988

PR #11054, opened 2022-03-04
Implement `@mulAdd` for scalar floats

schmee:
Some unresolved questions:

- Word on the street is that `.coerced_ty` is the new hotness in AstGen. I tried to use it but I couldn't figure out what I needed to do in Sema to make it work, any suggestions?
- `math.fma` is not implemented for `f16`. Is it okay to leave it as a TODO for now?

Vector support will be a follow-up PR!


src/AstGen.zig
```diff
@@ -7303,6 +7303,7 @@ fn builtinCall(
             const mulend2 = try expr(gz, scope, .{ .ty = float_type }, params[2]);
             const addend = try expr(gz, scope, .{ .ty = float_type }, params[3]);
             const result = try gz.addPlNode(.mul_add, node, Zir.Inst.MulAdd{
+                .ty = float_type,
```
andrewrk, line 7306: no need to add the float type here- all 3 operands have been coerced to it so Sema can get it from there.

src/Sema.zig
```diff
@@ -13518,8 +13518,68 @@ fn zirAtomicStore(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError
...
+        },
+        else => return sema.fail(block, src, "expected vector of floats or float type, found '{}'", .{ty}),
+    }
+
+    const mulend1 = sema.resolveInst(extra.mulend1);
+    const mulend2 = sema.resolveInst(extra.mulend2);
+    const addend = sema.resolveInst(extra.addend);
+    const mulend1_coerced = try sema.coerce(block, ty, mulend1, mulend1_src);
+    const mulend2_coerced = try sema.coerce(block, ty, mulend2, mulend2_src);
+    const addend_coerced = try sema.coerce(block, ty, addend, addend_src);
```
andrewrk, line 13547: No need to coerce because in AstGen uses `ResultLoc.ty` for these operands. These would be needed if you used `coerced_ty` instead.

src/Sema.zig
```diff
@@ -13518,8 +13518,68 @@ fn zirAtomicStore(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError
...
+
+    const target = sema.mod.getTarget();
+    switch (ty.zigTypeTag()) {
+        .ComptimeFloat, .Float => {
+            const maybe_mulend1 = try sema.resolveMaybeUndefVal(block, mulend1_src, mulend1_coerced);
+            const maybe_mulend2 = try sema.resolveMaybeUndefVal(block, mulend2_src, mulend2_coerced);
+            const maybe_addend = try sema.resolveMaybeUndefVal(block, addend_src, addend_coerced);
+            if (maybe_mulend1) |mulend1_val| {
```
andrewrk, line 13555: don't forget the case when mulend2 is comptime known to be undef but mulend1 is runtime-known. same deal with the addend.

src/value.zig
```diff
@@ -4020,6 +4020,40 @@ pub const Value = extern union {
         }
     }
 
+    pub fn mulAdd(float_type: Type, mulend1: Value, mulend2: Value, addend: Value, arena: Allocator, target: Target) Allocator.Error!Value {
+        switch (float_type.floatBits(target)) {
+            16 => {
+                if (true) {
+                    @panic("TODO implement mulAdd for f16");
```
andrewrk, line 4027: why not handle f16 and f80?

src/codegen/llvm.zig
```diff
@@ -3834,6 +3835,45 @@ pub const FuncGen = struct {
...
+        const fn_val = switch (ty.floatBits(target)) {
+            16, 32, 64 => blk: {
+                break :blk self.getIntrinsic("llvm.fma", &.{llvm_ty});
+            },
+            80 => return self.dg.todo("implement mulAdd for f80", .{}),
+            128 => blk: {
+                // LLVM incorrectly lowers the fma builtin for f128 to fmal, which is for
+                // `long double`. On some targets this will be correct; on others it will be incorrect.
+                if (target.longDoubleIsF128()) {
+                    break :blk self.getIntrinsic("llvm.fma", &.{llvm_ty});
```
andrewrk, line 3862: long double is sometimes f80; in this case we could also use the llvm intrinsic.

src/codegen/llvm.zig
```diff
@@ -3834,6 +3835,45 @@ pub const FuncGen = struct {
...
+
+        const ty = self.air.typeOfIndex(inst);
+        const llvm_ty = try self.dg.llvmType(ty);
+        const target = self.dg.module.getTarget();
+
+        const fn_val = switch (ty.floatBits(target)) {
+            16, 32, 64 => blk: {
+                break :blk self.getIntrinsic("llvm.fma", &.{llvm_ty});
+            },
+            80 => return self.dg.todo("implement mulAdd for f80", .{}),
```
andrewrk, line 3857: why todo?

andrewrk (changes requested):
nice work - I left some review comments.

schmee:
@andrewrk thanks for the review! I've pushed a new round of commits that hopefully addresses your comments 👍

src/codegen/llvm.zig
schmee: I tried using `llvm.fma` for this as well but it caused CI to blow up: https://ci.ziglang.org/ziglang/zig/2971/1/2. Disabling the f80 test for now and leaving a comment that it needs further investigation.

schmee:
Hmm, CI failure seems entirely unrelated, looks like a flaky test. Is it OK to disable this test and open an issue for it?

```
956/2191 os.test.test "std-native-ReleaseSmall-c-multi POSIX file locking with fcntl"... expected error.DeadLock, found 0
expected error.Locked, found 0
FAIL (TestUnexpectedError)
FAIL (TestUnexpectedError)
```

https://cloud.drone.io/ziglang/zig/16154/1/7

topolarity:
> Hmm, CI failure seems entirely unrelated, looks like a flaky test. Is it OK to disable this test and open an issue for it?
> 
> ```
> 956/2191 os.test.test "std-native-ReleaseSmall-c-multi POSIX file locking with fcntl"... expected error.DeadLock, found 0
> expected error.Locked, found 0
> FAIL (TestUnexpectedError)
> FAIL (TestUnexpectedError)
> ```
> 
> https://cloud.drone.io/ziglang/zig/16154/1/7

I hit a problem with the same failing test twice on #11024, once on aarch64-linux just like you and another time on FreeBSD (https://builds.hut.lavatech.top/~andrewrk/job/20704). Seems like the test is indeed flaky

andrewrk:
> Hmm, CI failure seems entirely unrelated, looks like a flaky test. Is it OK to disable this test and open an issue for it?

Yes. Generally if a flaky test failure is encountered, that is exactly our process. File an issue, then disable the flaky test with a link to the open issue.

andrewrk:
@schmee I force-pushed to your branch; if this passes the CI tests then we're good to merge.

andrewrk:
All tests passed locally as well.

PR #11056, opened 2022-03-04
stage2: test changed behavior of c pointer resolution from stage1

mitchellh:
stage1 peer resolves the given test to `[*c]u8` but stage2 makes that a
const u8. I believe stage2 behavior is correct since the pointer itself
is const, but I'm not sure.

Please advise what the proper behavior here should be and I can make it so.

### DIFF
--- a/test/behavior/pointers.zig
+++ b/test/behavior/pointers.zig
@@ -142,6 +142,15 @@ test "peer type resolution with C pointers" {
     try expect(@TypeOf(x4) == [*c]u8);
 }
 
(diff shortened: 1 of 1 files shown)


andrewrk:
Proper behavior is indeed to have the result type be `const` qualified.

I suggest to isolate the different behavior into its own test and then skip the test with stage1 (rather than having the test marked as "passing")

mitchellh:
Done

andrewrk (approved):

PR #11057, opened 2022-03-05
stage2: decl equality so &a == &a in comptime

mitchellh:
In stage2, decl ref equality was `decl == decl2` (pointer comparison).
But it is possible for two identical decls to have two separate pointers
via the `ref` ZIR. Each `ref` ZIR creates a new anonymous decl.
Therefore, `&a == &a` would return false in comptime.

This modifies the comparison so that if the pointers are not equivalent
(they are in the fast path), then we check if they were created in the
same scope from the same AST node. If so, we assume they're equivalent.

A better solution would probably be to intern decls so that we don't create
unnecessary decls, but that felt a lot harder to do. 

This definitely feels like the _wrong_ solution or at the very least a 
highly _suspect_ solution but existing tests continue to pass and the new 
test I found failing works, so I'm not sure. I'm opening this PR for discussion,
I'm not confident in this one. 😄 

This was the exact test that was failing before for stage2, passing stage1:

```zig
test "comptime decl equality" {
  var a: i32 = undefined;
  try expect(&a == &a);
  try expect(@ptrToInt(&a) == @ptrToInt(&a));
}
```

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -810,6 +810,21 @@ pub const Decl = struct {
         // as also alive, so that any Decl referenced does not get garbage collected.
         decl.val.markReferencedDeclsAlive();
     }
+
+    /// Two decls are considering equal if they were created in the same
+    /// scope from the exact same AST node.
+    pub fn eql(decl: *Decl, decl2: *Decl) bool {
+        // In most cases, the decls are pointer equivalent. Trivial.
+        if (decl == decl2) return true;
+
+        // In some cases, decls aren't pointer equiv but still equal.
+        // Example: &a == &a creates two decls, but are equal so long
+        // as its created in the same scope from the same node.
+        // TODO would be much better to intern decls rather than store
+        // duplicates and perform this comparison.
+        return decl.src_scope == decl2.src_scope and
+            decl.src_node == decl2.src_node;
+    }
 };
 
 /// This state is attached to every Decl when Module emit_h is non-null.
--- a/src/value.zig
+++ b/src/value.zig
@@ -1788,8 +1788,8 @@ pub const Value = extern union {
         if (lhs.pointerDecl()) |lhs_decl| {
             if (rhs.pointerDecl()) |rhs_decl| {
                 switch (op) {
-                    .eq => return lhs_decl == rhs_decl,
-                    .neq => return lhs_decl != rhs_decl,
+                    .eq => return lhs_decl.eql(rhs_decl),
+                    .neq => return !lhs_decl.eql(rhs_decl),
                     else => {},
                 }
             } else {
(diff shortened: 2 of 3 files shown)


SpexGuy:
I think you can create a test that fails if you depend on the hash of one of these pointers.  For example you may be able to observe a memoization cache miss if you pass one of these pointers into a const function.

mitchellh:
Indeed, I found a failing case that this definitely doesn't solve. `@ptrToInt(&a) == @ptrToInt(&a)`. So this isn't the right solution. I'll keep plugging away at it.

mitchellh:
> I think you can create a test that fails if you depend on the hash of one of these pointers. For example you may be able to observe a memoization cache miss if you pass one of these pointers into a const function.

Yep, this is correct, `hash` uses `hashPtr` for pointers and decl ref uses the pointer of the `*Module.Decl`. So there would be bugs there without changes.

What is the correct solution here? Should we be doing decl interning instead?

SpexGuy:
Interning would work, but IMO the most "correct" solution is for decls to already *be* pointers implicitly, and massage all the code that uses them to auto deref.  Might be kind of a big change though.

src/Module.zig
```diff
@@ -810,6 +810,21 @@ pub const Decl = struct {
         // as also alive, so that any Decl referenced does not get garbage collected.
         decl.val.markReferencedDeclsAlive();
     }
+
+    /// Two decls are considering equal if they were created in the same
```
komuw, line 814:
```suggestion
    /// Two decls are considered equal if they were created in the same
```

andrewrk (changes requested):
This is definitely not the correct fix. I'll help look into the underlying problem, but wanted to leave a review to prevent any trigger happy merging on this one.

mitchellh:
No problem! I’ll make it easier and close this. I figured it was not correct, but wasn’t sure what exactly the right approach was. If you have suggestions I’d be happy to take a stab.

andrewrk:
The `ref` ZIR instruction is working as intended - it takes a value and stuffs it into a new location, which now can have a pointer taken to it. `ref` does not provide the guarantee that the same value will be placed into the same memory location. In the example code, I don't see why this problem would be happening because the ZIR generated is:

```
      %69 = alloc_mut(@Ref.i32_type) node_offset:18:5
      %70 = store_node(%69, @Ref.undef) node_offset:18:18
...
        %73 = cmp_eq(%69, %69) node_offset:19:19
```

There are no `ref` instructions here, and so `alloc_mut` should create a memory location, and then clearly the comparison will be looking at the same memory location twice.

mitchellh:
I’m not in front of a computer right now but my test case I put in the original issue was reliably failing with stage 2. I extracted that from another test though, I can find the exact ones and get the ZIR when I am back at a pc. It was assigning two refs to two intermediates so the decls were not matching. In my ZIR I saw two separate Ref instructions and that was the key. 

Also note this issue is comptime only.

PR #11061, opened 2022-03-05
stage2: LLVM - Wasm improvements/fixes

Luukdegram:
- This fixes 2 entry points within the self-hosted wasm linker that would be called for the LLVM backend, whereas we should simply call into the LLVM backend to perform such action. i.e. not allocate a decl index when we have an LLVM object, and when flushing a module, we should be calling it on LLVM's object, rather than have the wasm linker perform the operation.
- Also, fixes were done for the wasm intrinsics `wasm.memory.size` and `wasm.memory.grow`.
- Lastly, this PR ensures that when an extern function is being resolved, we tell LLVM how to import such function.

Those fixes make all the behavior tests pass for the LLVM backend when targeting WebAssembly.

### DIFF
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -708,6 +708,14 @@ pub const DeclGen = struct {
         if (!is_extern) {
             llvm_fn.setLinkage(.Internal);
             llvm_fn.setUnnamedAddr(.True);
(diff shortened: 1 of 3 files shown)


andrewrk:
Nice! I'll enable CI coverage for this then :)

PR #11065, opened 2022-03-05
Do not fail to build if 'zig build-lib' etc. arguments exceed OS limits (take 2)

slimsag:
This brings back #10950, which was reverted in 5ab5e2e6731a9f1198df6c53134545ccc6a6bbd3 because it [introduced a regression in `zig run`](https://github.com/ziglang/zig/pull/10950#issuecomment-1049481212) where the runtime arguments passed were incorrect (sorry for not catching that sooner!)

I've fixed the issue, and notably this was the only location where we directly relied on accessing arguments by index in this code still (all other locations use the iterator proper) and so we should be all good to go now.

cc @andrewrk sorry for the regression!


andrewrk:
Thanks for the follow-up!

PR #11070, opened 2022-03-06
stage2: wasm - unify codegen with other backends

Luukdegram:
Feels good to remove code :)

The wasm linker and backend now use `generateSymbol` and `generate` to lower decls and perform codegen, similar to the other backends and linkers.


andrewrk:
Very nice!!

kubkon (approved):

PR #11079, opened 2022-03-07
stage2: make references to const allocs const

Vexu:


andrewrk:
We already have `_mut` versions of all the `alloc` ZIR instructions; I think we should be able to make that work without an additional instruction. `coerce_result_ptr` already makes an `alloc` mutable. `field_base_ptr` and `array_base_ptr` should probably do the same. Then the later references to the `alloc` instructions will have the appropriate constness. Can we start with something like this?

```diff
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -600,12 +600,12 @@ fn analyzeBodyInner(
         });
         const air_inst: Air.Inst.Ref = switch (tags[inst]) {
             // zig fmt: off
-            .alloc                        => try sema.zirAlloc(block, inst),
+            .alloc                        => try sema.zirAlloc(block, inst, false),
+            .alloc_mut                    => try sema.zirAlloc(block, inst, true),
             .alloc_inferred               => try sema.zirAllocInferred(block, inst, Type.initTag(.inferred_alloc_const)),
             .alloc_inferred_mut           => try sema.zirAllocInferred(block, inst, Type.initTag(.inferred_alloc_mut)),
             .alloc_inferred_comptime      => try sema.zirAllocInferredComptime(inst, Type.initTag(.inferred_alloc_const)),
             .alloc_inferred_comptime_mut  => try sema.zirAllocInferredComptime(inst, Type.initTag(.inferred_alloc_mut)),
-            .alloc_mut                    => try sema.zirAllocMut(block, inst),
             .alloc_comptime_mut           => try sema.zirAllocComptime(block, inst),
             .anyframe_type                => try sema.zirAnyframeType(block, inst),
             .array_cat                    => try sema.zirArrayCat(block, inst),
@@ -2426,7 +2426,12 @@ fn zirAllocInferredComptime(
     );
 }
 
-fn zirAlloc(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.Inst.Ref {
+fn zirAlloc(
+    sema: *Sema,
+    block: *Block,
+    inst: Zir.Inst.Index,
+    is_mut: bool,
+) CompileError!Air.Inst.Ref {
     const tracy = trace(@src());
     defer tracy.end();
 
@@ -2437,31 +2442,13 @@ fn zirAlloc(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
     if (block.is_comptime) {
         return sema.analyzeComptimeAlloc(block, var_ty, 0, ty_src);
     }
-    const target = sema.mod.getTarget();
-    const ptr_type = try Type.ptr(sema.arena, target, .{
-        .pointee_type = var_ty,
-        .@"addrspace" = target_util.defaultAddressSpace(target, .local),
-    });
-    try sema.requireRuntimeBlock(block, var_decl_src);
-    try sema.resolveTypeFully(block, ty_src, var_ty);
-    return block.addTy(.alloc, ptr_type);
-}
-
-fn zirAllocMut(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.Inst.Ref {
-    const tracy = trace(@src());
-    defer tracy.end();
-
-    const inst_data = sema.code.instructions.items(.data)[inst].un_node;
-    const var_decl_src = inst_data.src();
-    const ty_src: LazySrcLoc = .{ .node_offset_var_decl_ty = inst_data.src_node };
-    const var_ty = try sema.resolveType(block, ty_src, inst_data.operand);
-    if (block.is_comptime) {
-        return sema.analyzeComptimeAlloc(block, var_ty, 0, ty_src);
+    if (is_mut) {
+        try sema.validateVarType(block, ty_src, var_ty, false);
     }
-    try sema.validateVarType(block, ty_src, var_ty, false);
     const target = sema.mod.getTarget();
     const ptr_type = try Type.ptr(sema.arena, target, .{
         .pointee_type = var_ty,
+        .mutable = is_mut,
         .@"addrspace" = target_util.defaultAddressSpace(target, .local),
     });
     try sema.requireRuntimeBlock(block, var_decl_src);
```

Vexu:
> We already have `_mut` versions of all the `alloc` ZIR instructions; I think we should be able to make that work without an additional instruction. `coerce_result_ptr` already makes an `alloc` mutable. `field_base_ptr` and `array_base_ptr` should probably do the same. Then the later references to the `alloc` instructions will have the appropriate constness. Can we start with something like this?

I tried that approach first but it got quite intrusive having to make `{field,elem}_ptr` and `validate_{struct,array}_init` able to write through const, I think that could mostly be solved by calling a version of `{array,field}_base_ptr` for all struct/array inits not just the anonymous ones? But then I'd also have to be able to detect if the pointer is const because it's trying to assign to a constant variable or if it's just trying to initialize one? Feels much simpler to just give the init expr a mutable pointer and everything else a constant one.

andrewrk:
Alright, if it's simpler and it gets the job done, sounds good. We can always revisit later, armed with real world data about which ZIR instructions are the most expensive in practice.

PR #11082, opened 2022-03-07
std: SIMD utility functions

tecanec:
I've made a bunch of useful functions for working with SIMD. I hope that they'll make SIMD more accessible and practical to use. Performance-wise, these functions should all be quite easy on the run-time, as most of their functionality comes from compile-time operations. Example features include:

- `iota`, which generates a vector in which each element contains its own index.
- Element-wise shifts and rotates.
- `prefixScan`, which lets you generate a vector of sums/products/XORs/whatever of all earlier elements in the input vector. Kinda like atomic read-modify-write operations, but for SIMD instead of multi-threading.
- A function that suggests the recommended vector size for a given element type and CPU. (That particular feature is only implemented for x86_64 so far, and its solution may not be perfect.)

I decided to avoid high risks or runtime performance costs, so a lot of stuff in here is comptime-only. Much of it could be made possible in runtime if we had runtime shuffles, though, but we unfortunately don't. These functions should still be helpful for very linear algorithms, though.

This is my first time actually making a pull request to zig, so if I did something wrong or you think that I should change something, please tell me.


tecanec:
Ah. It seems to be broken. I might take a look sometime tomorrow.

tecanec:
Turns out some functions don't work on MIPS. The problem seems to be with `@shuffle` itself, so I consider fixing those out of scope for this PR. Those functions will trigger compile errors on MIPS, though (as noted by half a dozen comments...).

lib/std/simd.zig
```diff
@@ -0,0 +1,369 @@
...
+/// Suggests a vector size for a given type and CPU, or null if it doesn't support SIMD natively.
+/// Not yet implemented for every CPU architecture.
+pub fn suggestVectorSizeForCpu(comptime T: type, cpu: std.Target.Cpu) ?usize {
+    switch (cpu.arch) {
+        .x86_64 => {
+            if (T == bool and std.Target.x86.featureSetHas(.prefer_mask_registers)) return 64;
+
+            const vector_bit_size = if (std.Target.x86.featureSetHas(.avx512f)) 512 else if (std.Target.x86.featureSetHas(.prefer_256_bit)) 256 else if (std.Target.x86.featureSetHas(.prefer_128_bit)) 128 else return null;
+            const element_bit_size = std.math.max(8, std.math.ceilPowerOfTwo(T, @bitSizeOf(T)));
+            return vector_bit_size / element_bit_size;
```
matu3ba, line 18: This should probably be https://ziglang.org/documentation/master/#toc-divTrunc, since comptime_int is signed on default.

lib/std/simd.zig
```diff
@@ -0,0 +1,369 @@
...
+pub const Vector = std.meta.Vector;
+
+/// Suggests a vector size for a given type and CPU, or null if it doesn't support SIMD natively.
+/// Not yet implemented for every CPU architecture.
+pub fn suggestVectorSizeForCpu(comptime T: type, cpu: std.Target.Cpu) ?usize {
+    switch (cpu.arch) {
+        .x86_64 => {
+            if (T == bool and std.Target.x86.featureSetHas(.prefer_mask_registers)) return 64;
+
+            const vector_bit_size = if (std.Target.x86.featureSetHas(.avx512f)) 512 else if (std.Target.x86.featureSetHas(.prefer_256_bit)) 256 else if (std.Target.x86.featureSetHas(.prefer_128_bit)) 128 else return null;
```
matu3ba, line 16: `else return unreachable`
Also, please split this in more lines.

lib/std/simd.zig
```diff
@@ -0,0 +1,369 @@
...
+            const element_bit_size = std.math.max(8, std.math.ceilPowerOfTwo(T, @bitSizeOf(T)));
+            return vector_bit_size / element_bit_size;
+        },
+        else => @compileError("No vector sizes for this CPU architecture have yet been recommended"),
+    }
+}
+
+/// Suggests a vector size for a given type, or null if the target CPU doesn't support SIMD natively.
+/// Not yet implemented for every CPU architecture.
+pub fn suggestVectorSize(comptime T: type) ?usize {
```
matu3ba, line 26: This function looks redundant. Why is it needed?
If it is needed, only comment on the more likely used (shorter) function that wraps the longer one.

lib/std/simd.zig
```diff
@@ -0,0 +1,369 @@
...
+        const b_indices = shiftElementsUp(count_up - cycle * @splat(len, @intCast(i32, a_vec_count)), a_vec_count, 0);
+        break :blk @select(i32, select_mask, a_indices, ~b_indices);
+    };
+
+    return @shuffle(Child, a, b, indices);
+}
+
+/// The contents of braided is evenly split between vec_count vectors that are returned as an array. They "take turns",
+/// recieving one element from braided at a time.
+pub fn unbraid(comptime vec_count: usize, braided: anytype) [vec_count]Vector(vectorLength(@TypeOf(braided)) / vec_count, std.meta.Child(@TypeOf(braided))) {
```
matu3ba, line 119: newline can be created with comma after last argument.

lib/std/simd.zig
```diff
@@ -0,0 +1,369 @@
...
+    comptime var i: usize = 0; // for doesn't work, apparently.
+    inline while (i < out.len) : (i += 1) {
+        const indices = comptime countUp(i32, vec_len) * @splat(vec_len, @intCast(i32, vec_count)) + @splat(vec_len, @intCast(i32, i));
+        out[i] = @shuffle(Child, braided, undefined, indices);
+    }
+
+    return out;
+}
+
+pub fn extract(vec: anytype, comptime first: VectorIndex(@TypeOf(vec)), comptime count: VectorCount(@TypeOf(vec))) Vector(count, std.meta.Child(@TypeOf(vec))) {
```
matu3ba, line 134: newline same here.

lib/std/simd.zig
```diff
@@ -0,0 +1,369 @@
...
+
+    try std.testing.expectEqual(@as(?u3, 1), firstIndexOfValue(base, 4));
+    try std.testing.expectEqual(@as(?u3, 4), lastIndexOfValue(base, 4));
+    try std.testing.expectEqual(@as(?u3, null), lastIndexOfValue(base, 99));
+    try std.testing.expectEqual(@as(u4, 3), countElementsWithValue(base, 4));
+}
+
+/// Same as accum, but with a user-provided function that mathematically must be associative.
+/// The function must also do nothing when called with identity as either argument.
+pub fn accumWithFunc(comptime hop: isize, vec: anytype, comptime func: fn (@TypeOf(vec), @TypeOf(vec)) @TypeOf(vec), comptime identity: std.meta.Child(@TypeOf(vec))) @TypeOf(vec) {
```
matu3ba, line 269: newline same here.

lib/std/simd.zig
```diff
@@ -0,0 +1,369 @@
...
+    // I haven't debugged this, but it might be a cousin of sorts to what's going on with braid.
+    comptime if (builtin.cpu.arch.isMIPS() and !builtin.is_test) @compileError("TODO: Find out why accum doesn't work on MIPS");
```
matu3ba, line 271: Does it only break without test mode? You did not mention this in the associated text.

lib/std/simd.zig
```diff
@@ -0,0 +1,369 @@
...
+    if (!builtin.cpu.arch.isMIPS()) {
+        try std.testing.expectEqual([8]u32{ 10, 55, 20, 66, 30, 77, 40, 88 }, braid(.{ base, other_base }));
+
+        const small_braid = braid(small_bases);
+        try std.testing.expectEqual([10]u8{ 0, 2, 4, 6, 8, 1, 3, 5, 7, 9 }, small_braid);
+        try std.testing.expectEqual(small_bases, unbraid(small_bases.len, small_braid));
+    }
+}
+
+pub fn shiftElementsUp(vec: anytype, comptime amount: VectorCount(@TypeOf(vec)), shift_in: std.meta.Child(@TypeOf(vec))) @TypeOf(vec) {
```
matu3ba, line 168: description? what is "up shifting"? shifting each element left? increasing each element with a shift?!

lib/std/simd.zig
```diff
@@ -0,0 +1,369 @@
+// Note: Some functions here are known to malfunction on MIPS.
+// Functions whose tests fail on MIPS will trigger compile errors when compiled for that platform.
```
matu3ba, line 2: Super-brief introduction of SIMD and error behavior of SIMD belongs into the top level comment.

matu3ba (changes requested):
1. SIMD has no error handling, so testing error behavior and edge is not very useful. Is this correct? Absence of errors should be documented in a top-level comment.
2. `shiftElementsUp` is from my point of view an ambiguous/non-optimal naming, since shifting regularly is understood to be leftwards or rightwards. If there is additional meaning, the functions should be documented
3. add newlines and check, if `return unreachable` is used for comptime stuff and otherwise a sane default.

lib/std/simd.zig
tecanec: It's shifting elements to higher indices. I'm not sure whether that'd be called a left- or a right-shift (it's towards the right of an array literal, but the bit positions of elements within the vector are technically left-shifted), so I went with up-shift. Thinking back, though, I must admit that that might've been a poor choice. Should I call it left or right?

lib/std/simd.zig
tecanec: Are you saying that you want me to write one? Thanks for the review, by the way.

lib/std/simd.zig
tecanec: Can I split this line up without zig fmt reverting it?

matu3ba: using brackets should work. Not sure, if comma like for function arguments also works.

lib/std/simd.zig
matu3ba: > Are you saying that you want me to write one?

Yup. It does not need to be perfect. Just summarizing what you are using.

lib/std/simd.zig
tecanec: I thought that needing a suggested vector size for the target platform would be the most common use-case, but that there might be other use-cases that I haven't foreseen. So `suggestVectorSize` is just a shortcut.

lib/std/simd.zig
tecanec: This is just me struggling to stop the compile error from triggering under the wrong circumstances. Not having the `comptime` keyword caused the compile error to trigger on every platform. Admittedly, it was a bit rushed, so I'll try removing the `!builtin.is_test` and see what happens once I get home.

lib/std/simd.zig
matu3ba: [I found this description for bi-endian](https://llvm.org/devmtg/2014-10/Slides/Schmidt-SupportingVectorProgramming.pdf),
so shifting up sounds reasonable if you add a comment (and ideally some ascii art explaining things for x86).
Do you know, if bi-endian architectures can change operation endianess at runtime?

I think MIPS fails, because it is used as big endian in CI.

tecanec: I'm staying away from raw memory access since it seems a bit risky in terms of platform-dependent undefined behavior.

lib/std/simd.zig
matu3ba: I dont think `suggestVectorSize` is used that frequently to justify the additional function, which only adds `std.Target.Cpu`.
SIMD code will likely require to handle non-SIMD for portability, which requires usage of `std.Target`.
So I dont see the benefit.

tecanec: My rationale is that SIMD code will do either of these:

- Use the same vector sizes unconditionally. Zig lets you use any vector size you want, even if it's not native, much like with ints such as `u128`. This is simple to work with, and is fitting if you only need SIMD for a small, known amount of data that you'd expect to fit into one or two native vector registers. In this case, `suggestVectorSize` is not used at all.
- Adapt to the native vector size of the target platform, or avoid SIMD altogether if it's not native. `suggestVectorSize` will tell both if and how large vectors should be used, so I'm not sure why `std.Target` would be needed.

The reason I added both `suggestVectorSize` was to support the latter case, and I then decided to split it into `suggestVectorSizeForCpu` because I thought it'd be a waste if someone needed but couldn't get the suggested vector size for a different CPU. I'm not opposed to removing `suggestVectorSize` and having programmers use `suggestVectorSizeForCpu` directly. I simply thought that having it seemed more in style with the rest of stdlib, but I could be wrong.

lib/std/simd.zig
```diff
@@ -0,0 +1,395 @@
...
+
+        acc = if (ErrorType == void) func(acc, shifted) else try func(acc, shifted);
+    }
+    return acc;
+}
+
+/// Returns a vector whose elements are the result of performing the specified operation on the corresponding
+/// element of the input vector and every hop'th element that came before it (or after, if hop is negative).
+/// Supports the same operations as the @reduce() builtin. Takes O(logN) to compute.
+pub fn accum(comptime op: std.builtin.ReduceOp, comptime hop: isize, vec: anytype) @TypeOf(vec) {
```
Snektron, line 317: This operation is usually either called prefix sum (for addition) or prefix scan (generic). Its also worth noting here whether the operation is inclusive or exclusive.

lib/std/simd.zig
```diff
@@ -0,0 +1,395 @@
...
+    const VecType = @TypeOf(vec);
+    const Child = std.meta.Child(VecType);
+    const len = vectorLength(VecType);
+
+    const identity = comptime switch (@typeInfo(Child)) {
```
Snektron, line 322: Another option would be to pass the identity explicitly here as well, though i don't know of many situations where that is actually useful. Its probably fine.

lib/std/simd.zig
```diff
@@ -0,0 +1,395 @@
...
+}
+
+/// Same as accum, but with a user-provided function that mathematically must be associative.
+/// The function must also do nothing when called with identity as either argument.
+/// Set ErrorType to void if func doesn't return an error union.
+pub fn accumWithFunc(
+    comptime hop: isize,
+    vec: anytype,
+    comptime ErrorType: type,
+    comptime func: fn (@TypeOf(vec), @TypeOf(vec)) if (ErrorType == void) @TypeOf(vec) else ErrorType!@TypeOf(vec),
```
Snektron, line 293: I would say that allowing the callback to return an error set here is a little overkill. Did you have a specific use case for this in mind?

lib/std/simd.zig
```diff
@@ -0,0 +1,395 @@
...
+    const Child = std.meta.Child(@TypeOf(a));
+    const a_len = vectorLength(@TypeOf(a));
+    const b_len = vectorLength(@TypeOf(b));
+
+    return @shuffle(Child, a, b, @as([a_len]i32, countUp(i32, a_len)) ++ @as([b_len]i32, ~countUp(i32, b_len)));
+}
+
+/// Returns a vector whose elements alternates between those of each input vector.
+/// For example, braid(.{[4]u32{11, 12, 13, 14}, [4]u32{21, 22, 23, 24}}) returns a vector containing {11, 21, 12, 22, 13, 23, 14, 24}.
+pub fn braid(vecs: anytype) Vector(vectorLength(@TypeOf(vecs[0])) * vecs.len, std.meta.Child(@TypeOf(vecs[0]))) {
```
Snektron, line 86: Other options: (de)interlace, (un)zip.

lib/std/simd.zig
```diff
@@ -0,0 +1,395 @@
...
+}
+
+/// Returns the smallest type of an unsigned int capable of holding the length of the given vector type.
+pub fn VectorCount(comptime VectorType: type) type {
+    return std.math.IntFittingRange(0, vectorLength(VectorType));
+}
+
+/// Returns a vector containing the integer sequence starting at 0.
+/// For example, countUp(i32, 8) will return a vector containing {0, 1, 2, 3, 4, 5, 6, 7}.
+pub fn countUp(comptime T: type, comptime len: usize) Vector(len, T) {
```
Snektron, line 55: Other options include 'iota', or maybe something like 'indices'?

lib/std/simd.zig
Snektron: I would probably call it 'shifting left' and 'shifting right'.

Snektron (commented):
These functions seem pretty handy to have. I do have some concerns whether the quality of generated code will be good, since the implementations are in general pretty abstract.

Regarding the MIPS stuff, if this is a Zig problem, I suggest making an issue like "SIMD stuff doesnt work on MIPS" and then linking an url with "TODO" somewhere in this file.

lib/std/simd.zig
tecanec: Mainly functions that can return `error.Overflow`.

lib/std/simd.zig
tecanec: > I would probably call it 'shifting left' and 'shifting right'.
(Somehow can't reply to the right comment)
So, what'd you consider "left" and "right"?

Snektron: Arrays are usually noted from left to right, when you write `[a, b, c]` index 0 is written on the left side, and so it seems to me that 'shifting left' is the same as 'shifting up'. Another option would be to generalize the two and use a signed value for the shift, though i must admit that i always forget whether to use positive or negative values for a particular rotate.

tecanec: How about `getIndices` or `genIndices`?

tecanec: I'm unsure what'd be left or right because on little-endian platforms like x86, shifting elements up is a left-shift in terms of bits.

lib/std/simd.zig
tecanec: That's not how `identity` is meant to be used, and it's already a somewhat hacky solution, so I think it's best if it's made no more involved than it needs to be. That's also why it's marked with comptime in the callback version, even though it technically doesn't need to be. My philosophy is that hacky solutions are okay if they only cover a minimal amount of code and are known to work.

lib/std/simd.zig
```diff
@@ -0,0 +1,395 @@
...
+    const len = vectorLength(@TypeOf(vec));
+
+    std.debug.assert(amount <= len);
+
+    const indices = comptime @mod(iota(i32, len) + @splat(len, @as(i32, amount)), @splat(len, @intCast(i32, len)));
+
+    return @shuffle(Child, vec, undefined, indices);
+}
+
+/// Elements are shifted to higher indices. Elements that pass the top return to the bottom.
```
matu3ba, line 207: Phrasing could be improved. Maybe something like the following:
-Elements are shifted to higher indices.
+Vector elements are shifted rightwards.
-Elements that pass the top return to the bottom.
+Elements that pass the highest index are returned relative to the offset starting at index 0.

lib/std/simd.zig
```diff
@@ -0,0 +1,395 @@
+//! This module provides functions for working conveniently with SIMD (Single Instruction; Multiple Data),
+//! which may offer a potential boost in performance on some targets by performing the same operations on
+//! multiple elements at once.
+//! Please be aware that some functions are known to not work on MIPS.
+
+const std = @import("std");
+const builtin = @import("builtin");
+
+pub const Vector = std.meta.Vector;
+
```
matu3ba, line 10: I tend to explain the basic data structures as regular comment, so that first users know how to deal with stuff
and to not clutter the top level decl with information.
```zig
// A Vector is continuous and has comptime-known length and element sizes with the following layout:
// data  A B C D
// index 0 1 2 3
// -> right
// <- left
```

lib/std/simd.zig
```diff
@@ -0,0 +1,395 @@
...
+}
+
+/// Elements are shifted to lower indices. Elements that pass the bottom are discarded.
+pub fn shiftElementsLeft(vec: anytype, comptime amount: VectorCount(@TypeOf(vec)), shift_in: std.meta.Child(@TypeOf(vec))) @TypeOf(vec) {
+    const len = vectorLength(@TypeOf(vec));
+
+    return join(extract(vec, amount, len - amount), @splat(amount, shift_in));
+}
+
+/// Elements are shifted to lower indices. Elements that pass the bottom return to the top.
```
matu3ba, line 195: Phrasing could be improved. Maybe something like the following:
-Elements are shifted to lower indices.
+Vector elements are shifted leftwards.
-Elements that pass the bottom return to the top.
+Elements that pass the index 0 are returned relative to the offset starting at the highest index.

lib/std/simd.zig
```diff
@@ -0,0 +1,395 @@
...
+    if (comptime !builtin.cpu.arch.isMIPS()) {
+        try std.testing.expectEqual([8]u32{ 10, 55, 20, 66, 30, 77, 40, 88 }, interlace(.{ base, other_base }));
+
+        const small_braid = interlace(small_bases);
+        try std.testing.expectEqual([10]u8{ 0, 2, 4, 6, 8, 1, 3, 5, 7, 9 }, small_braid);
+        try std.testing.expectEqual(small_bases, deinterlace(small_bases.len, small_braid));
+    }
+}
+
+/// Elements are shifted to higher indices. Elements that pass the top are discarded.
```
matu3ba, line 183: also phrasing here.

lib/std/simd.zig
```diff
@@ -0,0 +1,395 @@
...
+pub fn shiftElementsRight(vec: anytype, comptime amount: VectorCount(@TypeOf(vec)), shift_in: std.meta.Child(@TypeOf(vec))) @TypeOf(vec) {
+    return join(@splat(amount, shift_in), extract(vec, 0, vectorLength(@TypeOf(vec)) - amount));
+}
+
+/// Elements are shifted to lower indices. Elements that pass the bottom are discarded.
```
matu3ba, line 188: also phrasing here.

matu3ba (approved):
Looking good, though the comments dont match yet the shift functions and what a vector is should be explained.

lib/std/simd.zig
tecanec: I say "higher indices" because left versus right is somewhat ambigious, as discussed earlier. Will "shifted rightwards (towards higher indicies)" work?

matu3ba: That would also work. Only relevant things is make it clear what is meant with the terminology. see my comment below ` pub const Vector = std.meta.Vector;` as the basic structure.

lib/std/simd.zig
tecanec: This seems like the responsibility of the language docs, since SIMD vectors are a language-provided feature. Am I wrong?

matu3ba: good catch. Can you add it to the section `Vectors` in the langref or create an issue for this?

tecanec: They already seem to be doing fine, to me.

lib/std/simd.zig
```diff
@@ -381,7 +393,7 @@ test "vector prefix scan" {
     try std.testing.expectEqual(Vector(4, i32){ 11, 11, 9, -21 }, prefixScan(.Min, 1, int_base));
     try std.testing.expectEqual(Vector(4, i32){ 11, 23, 23, 23 }, prefixScan(.Max, 1, int_base));
 
-    // Not testing add and mul because I'm not gonna try predicting those rounding errors.
+    // Trying to predict the rounding errors of addition and multiplication of floats would be a mess, so we don't test those.
```
matu3ba, line 396: Error range prediction is not that hard, if the arithmetic is simple. The hard part is to get the error ranges for each SIMD instruction + have a standard way to extract the user-set error range (see https://composter.com.ua/documents/AVX512.pdf or https://developer.apple.com/documentation/accelerate/simd_float4).
IEE754 has yet no standard that includes the NaN behavior, ie denormals and +-infty https://www.agner.org/optimize/nan_propagation.pdf so the programmer must ensure this does not happen or handle it on a per hardware case.

lib/std/simd.zig
```diff
@@ -314,6 +325,7 @@ pub fn prefixScanWithFunc(
 /// Returns a vector whose elements are the result of performing the specified operation on the corresponding
 /// element of the input vector and every hop'th element that came before it (or after, if hop is negative).
 /// Supports the same operations as the @reduce() builtin. Takes O(logN) to compute.
+/// Note that the algorithm used by this function might affect floating point rounding errors.
```
matu3ba, line 328: How does it affect floating point rounding errors? increase, decrease?
You could explain that its the programmer's job to ensure errors are in expected range + add mention that there is no function yet to capture those per platform.

lib/std/simd.zig
matu3ba: Often SIMD does not conform IEE754 at all for performance.

matu3ba (commented):
if you want to add usage information for the programmer, try to be very specific on the necessary background information.

having an api to get the error per simd operation would also be wishful: https://developer.apple.com/documentation/accelerate/simd_float4
though I am unsure how feasible it is to extract the error on a per hardware base.

lib/std/simd.zig
matu3ba: I think it is fine like this and my stuff belongs into the specification.

lib/std/simd.zig
tecanec: I meant it more as a warning for those who need deterministic behavior. Kinda like `@setFloatMode`.
In theory, this function should actually reduce rounding errors compared to a linear, sequential scan, as the values are more likely to be similar.

lib/std/simd.zig
tecanec: Are you telling me to change the comment or to implement the test?

matu3ba: nope. this is just missing a reason why its a mess and what we could (not) do about it.

PR #11090, opened 2022-03-08
stage2: correct comptime-known inferred alloc type, plus comptime array init sentinels

mitchellh:
Two changes in here, both needed to get the tests passing in this PR.

At least 3 more passing tests!

1. For inferred allocs with values known at comptime _but not forced comptime_ (i.e. not in a `comptime {}` block), we were setting the incorrect type, causing cascading errors. We can now set the type to the bitcast type. The previous comment noted we can't do that due to constness, but @Vexu's recent `make_const_ptr` work fixes this.

2. A mistake from my prior PR on sentinel array initialization, I was not adding the sentinel to the end of the array value for comptime. Bonus: this solidifies that adding the sentinel as the final op in the `array_init_sent` ZIR is correct.

Tests used during debugging/development are below. First two are the inferred alloc commit, last one is the sentinel bug. The real behavior tests exercise both at once.

```zig
test "inferred alloc with sentinel" {
    const ary = [_:0]u8{42};
    const ptr: [*:0]const u8 = &ary;
    try expect(ptr[1] == 0);
}

test "inferred alloc with sentinel at comptime" {
    comptime {
        const ary = [_:0]u8{42};
        const ptr: [*:0]const u8 = &ary;
        try expect(ptr[1] == 0);
    }
}

test "comptime sentinel array init" {
    comptime {
        var ptr: [*:0]const u8 = &[_:0]u8{1};
        try expect(ptr[1] == 0);
    }
}
```

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -11827,21 +11827,16 @@ fn zirArrayInit(
         });
     };
 
-    const elems = if (!is_sent)
-        resolved_args
-    else
-        resolved_args[0 .. resolved_args.len - 1];
-
-    const opt_runtime_src: ?LazySrcLoc = for (elems) |arg| {
+    const opt_runtime_src: ?LazySrcLoc = for (resolved_args) |arg| {
         const arg_src = src; // TODO better source location
         const comptime_known = try sema.isComptimeKnown(block, arg_src, arg);
         if (!comptime_known) break arg_src;
     } else null;
 
     const runtime_src = opt_runtime_src orelse {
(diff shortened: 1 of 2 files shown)


andrewrk:
Nice work!

> Tests used during debugging/development are below. First two are the inferred alloc commit, last one is the sentinel bug. The real behavior tests exercise both at once.

For future reference, I would consider it an improvement if you took any behavior tests that test too much at once and meaningfully split them into multiple independent tests like this.

Issue #11022, opened 2022-03-01
proposal: error set equality behavior


mitchellh:
Tested in Zig versions: 0.9.0 released, 0.10.0-dev.1037+331cc810d

**I know you said no new language proposals.** I'm only bringing this up because there is an explicit TODO in the stage2 codebase to discuss this...

I was exploring how error set type equality works in stage1, so that I could implement it in stage2, and I noticed the behavior is perhaps not consistent with other type comparisons and I would argue not generally useful in its current form. I am proposing changing it for stage2. 

# Current Stage1 Behavior

The test below demonstrates:

```zig
// This is the current stage1 behavior
test "error set equality" {
	const a = error{One};
	const b = error{One};
	
	try expect(a == a);
	try expect(a != b);
	try expect(a != error{One});
}
```

The error must be exactly the error declaration, it doesn't match structurally. 

## Inconsistent?

This is inconsistent with `anyerror`, which is itself an error set. (Edit: as noted in comments, this is not actually inconsistent if you consider `anyerror` is just a declaration to the "global error set.")

```zig
// But...
test "anyerror" {
  const a = anyerror;
  const b = anyerror;
  
  try expect(a == anyerror);
  try expect(a == a);
  try expect(a == b);
}
```

Further, this is inconsistent with other types:

```zig
// But...
test "alias u8" {
  const a = u8;
  const b = u8;
  
  try expect(a == u8);
  try expect(a == a);
  try expect(a == b);
}
```

Notice that as long as the type _values_ match, it is true. It doesn't matter where it was declared.

**But, it _is consistent_ with structs.** If we consider an `error` a structural type, in spirit.

# Proposed Behavior

I propose that the following test should be the working behavior.

**However,** Zig doesn't structurally match struct types. That is a good argument against. 

```zig
// Suggestion, does NOT work in stage1 or stage2 right now
test "error set equality" {
	const a = error{One};
	const b = error{One};
	
	try expect(a == a);
	try expect(a == b);
	try expect(a == error{One});
	
	// should treat as a set
	const c = error{One,Two};
	const d = error{Two,One};
	
	try expect(c == d);
}
```

MasterQ32:
Related issues: 
- #7533
- #9573
- #7893

Hejsil:
Just want to clarify that currently the error set declaration works very much like an enum declaration here. So if we replace `error` with `enum` we can see that the behavior is the same:
```zig
test "enum set equality" {
	const a = enum{One};
[...]

PR #11098, opened 2022-03-09
stage2: error set type equality, error and error union value equality

mitchellh:
This implements type equality for merged error types. This is done through element-wise error set comparison. This is a bit odd since no other error type is structurally compared (see #11022) but it appears to be the same behavior that stage1 also expects based on tests.

Also took the opportunity to clean up the style of error set comparisons.

This gets 6 more error-specific tests passing!

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -824,7 +824,7 @@ pub const ErrorSet = struct {
     /// Offset from Decl node index, points to the error set AST node.
     node_offset: i32,
     /// The string bytes are stored in the owner Decl arena.
-    /// They are in the same order they appear in the AST.
+    /// These must be in sorted order. See sortNames.
     names: NameMap,
 
     pub const NameMap = std.StringArrayHashMapUnmanaged(void);
@@ -835,6 +835,18 @@ pub const ErrorSet = struct {
             .parent_decl_node = self.owner_decl.src_node,
             .lazy = .{ .node_offset = self.node_offset },
         };
+    }
+
+    /// sort the NameMap. This should be called whenever the map is modified.
+    /// alloc should be the allocator used for the NameMap data.
+    pub fn sortNames(names: *NameMap) void {
+        const Context = struct {
+            keys: [][]const u8,
+            pub fn lessThan(ctx: @This(), a_index: usize, b_index: usize) bool {
+                return std.mem.lessThan(u8, ctx.keys[a_index], ctx.keys[b_index]);
+            }
+        };
+        names.sort(Context{ .keys = names.keys() });
     }
 };
 
--- a/src/type.zig
+++ b/src/type.zig
@@ -556,37 +556,36 @@ pub const Type = extern union {
                 return info_a.signedness == info_b.signedness and info_a.bits == info_b.bits;
             },
 
+            .error_set_inferred => {
+                // Inferred error sets are only equal if both are inferred
+                // and they originate from the exact same function.
+                const a_set = a.castTag(.error_set_inferred).?.data;
+                const b_set = (b.castTag(.error_set_inferred) orelse return false).data;
+                return a_set.func == b_set.func;
+            },
+
+            .anyerror => {
+                return b.tag() == .anyerror;
+            },
+
             .error_set,
             .error_set_single,
-            .anyerror,
-            .error_set_inferred,
             .error_set_merged,
             => {
-                if (b.zigTypeTag() != .ErrorSet) return false;
-
-                // inferred error sets are only equal if both are inferred
-                // and they originate from the exact same function.
-                if (a.castTag(.error_set_inferred)) |a_pl| {
-                    if (b.castTag(.error_set_inferred)) |b_pl| {
-                        return a_pl.data.func == b_pl.data.func;
-                    }
-                    return false;
+                switch (b.tag()) {
+                    .error_set, .error_set_single, .error_set_merged => {},
+                    else => return false,
                 }
-                if (b.tag() == .error_set_inferred) return false;
 
-                // anyerror matches exactly.
-                const a_is_any = a.isAnyError();
-                const b_is_any = b.isAnyError();
-                if (a_is_any or b_is_any) return a_is_any and b_is_any;
-
-                // two resolved sets match if their error set names match.
+                // Two resolved sets match if their error set names match.
+                // Since they are pre-sorted we compare them element-wise.
                 const a_set = a.errorSetNames();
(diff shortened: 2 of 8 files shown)


src/type.zig
```diff
@@ -567,24 +567,30 @@ pub const Type = extern union {
...
                 // for error set types.
 
-                if (a.tag() == .anyerror and b.tag() == .anyerror) {
-                    return true;
-                }
-
-                if (a.tag() == .error_set and b.tag() == .error_set) {
-                    return a.castTag(.error_set).?.data.owner_decl == b.castTag(.error_set).?.data.owner_decl;
-                }
+                if (a.tag() != b.tag()) return false;
```
andrewrk, line 570: this looks wrong - why can't the tags be different?

andrewrk (changes requested):

src/type.zig
mitchellh: This is the same logic as before (see split diff!)

andrewrk: the `return false` at the end of master branch is wrong - it's essentially returning a false negative as a TODO.

mitchellh: I can turn this into `unreachable` for now if that'll unblock this PR, since it feels like merging the changes of #11022 into this might be heavy (I have those in a branch, but requires more changes i.e. due to resolving inferred error sets).

andrewrk: Alright, #11022 is marked "accepted" now, hope that helps :-)

I'm fine with however you want to structure the changesets.

Also I think we can probably get away with implementing this proposal only in stage2 and leaving stage1 to rot for now. In case you haven't seen it, this is the plan to deal with that: #6378

mitchellh: Hah okay, stand by. I'll re-request review once those are in this PR.

mitchellh:
I added the logic for #11022. I'm not sure why the drone CI is failing on the `-ofmt=c` run.

src/type.zig
mitchellh: Added it, note that even though the changes are only in stage2, I don't think anything was or is exercising the changed behavior in the compiler (besides new tests). So in regards to #6378 I don't think this will negatively impact bootstrapping in any way.

andrewrk:
Likely the different value returned for equality caused some different codepath somewhere and uncovered a missing feature or bug in the C backend. Probably it's just a couple test cases that can now be regressed. I'd be happy to take it from here if you want.

mitchellh:
Yes please all yours! since I don't have the local env setup (or knowledge yet) to debug the C backend. 😄

src/type.zig
```diff
@@ -564,27 +564,41 @@ pub const Type = extern union {
...
-                // TODO: revisit the language specification for how to evaluate equality
-                // for error set types.
-
-                if (a.tag() == .anyerror and b.tag() == .anyerror) {
-                    return true;
-                }
+                // anyerror matches exactly.
+                const a_is_any = a.isAnyError();
+                const b_is_any = b.isAnyError();
+                if (a_is_any or b_is_any) return a_is_any and b_is_any;
```
andrewrk, line 570: This can return a false negative:

```zig
    /// Returns true if it is an error set that includes anyerror, false otherwise.
    /// Note that the result may be a false negative if the type did not get error set
    /// resolution prior to this call.
    pub fn isAnyError(ty: Type) bool {
```

src/type.zig
```diff
@@ -564,27 +564,41 @@ pub const Type = extern union {
...
+                // there are cases the caller has to compare type equality
+                // without being able to resolve inferred error sets
+                // (i.e. comparing function type signatures for generic
+                // instantiation) so we DO NOT require inferred set resolution
+                if (a.castTag(.error_set_inferred)) |a_pl| {
+                    if (b.castTag(.error_set_inferred)) |b_pl| {
+                        // We can verify equality without resolution if data matches exactly
+                        if (a_pl.data == b_pl.data) return true;
+
+                        if (!b_pl.data.is_resolved) return false;
```
andrewrk, line 582: This could be a false negative.

andrewrk (changes requested):
I think `Type.eql` should never return false negatives because that would cause subtle bugs for us down the road. There's also `Type.hash` to implement along with these changes.

src/type.zig
mitchellh: Ah yes, I should do this later in the call.

andrewrk:
I suggest to delete this from test/stage2/x86_64.zig:

```zig
        {
            var case = ctx.exe("error set equality", target);

            case.addCompareOutput(
                \\pub fn main() void {
                \\    assert(@TypeOf(error.Foo) == @TypeOf(error.Foo));
                \\    assert(@TypeOf(error.Bar) != @TypeOf(error.Foo));
                \\    assert(anyerror == anyerror);
                \\    assert(error{Foo} != error{Foo});
                \\    // TODO put inferred error sets here when @typeInfo works
                \\}
                \\fn assert(b: bool) void {
                \\    if (!b) unreachable;
                \\}
            ,
                "",
            );
        }
```

it doesn't belong there, and we already have coverage for this in behavior tests. these kind of tests are supposed to be for testing incremental compilation and other complex invocations, but the test harness got abused early on before we had `zig test` working.

src/Module.zig
```diff
@@ -836,6 +836,31 @@ pub const ErrorSet = struct {
...
+            while (j > 0 and std.mem.lessThan(u8, key, keys[j - 1])) : (j -= 1) {
+                keys[j] = keys[j - 1];
+                hashes[j] = hashes[j - 1];
+            }
+            keys[j] = key;
+            hashes[j] = hash;
+        }
+
+        try names.reIndex(alloc);
+    }
```
andrewrk, line 863: I made #11109 so that this can be improved:

```suggestion
    pub fn sortNames(names: *NameMap) void {
        const Context = struct {
            keys: [][]const u8,

            pub fn lessThan(ctx: @This(), a_index: usize, b_index: usize) bool {
                return std.mem.lessThan(u8, ctx.keys[a_index], ctx.keys[b_index]);
            }
        };
        names.sort(Context{ .keys = names.keys() });
    }
```

andrewrk (changes requested):

src/Module.zig
mitchellh: Looks great! Once #11109 merges I'll make those changes.

mitchellh: Done!

mitchellh:
Using the new sort method, should be good to go.

PR #11102, opened 2022-03-09
stage2: Tests passing for LLVM backend 

curtiswilkinson:
Found extra passing tests (LLVM backend) while working and figured I'd move the goal slightly closer.

These pass **at least** on the LLVM backend, I was conservative with the skips and some of these would potentially pass on the other backends. 

- Stage2 compiled with `./build/zig build --zig-lib-dir lib --prefix $(pwd)/stage2 -Denable-llvm -Dconfig_h=build/config.h`
- Tests emit `898 passed; 113 skipped; 0 failed.` when run with `./stage2/bin/zig test ./test/behavior.zig -I ./test -fLLVM`


andrewrk:
Appreciate the effort but please do this kind of meta work only along with real work. When I've been moving the tests to the "passing for llvm" status, I've been testing all the backends and not skipping ones unnecessarily. So I'll close this and test them myself and only skip the ones that need to be skipped. Thanks anyways - sincerely.

PR #11105, opened 2022-03-09
stage2: wasm - miscellaneous improvements

Luukdegram:
This improve the lowering of switches as well as implements some additional instructions:
- fpext (only f32->f64)
- fptrunc (only f64->f32)
- errunion_payload_ptr_set
- field_parent_ptr

Lastly it enables the test cases that are now passing thanks to above fixes/implementations as well as tests that were already passing.


andrewrk (approved):
Looks great :+1:

PR #11109, opened 2022-03-10
std: add sort method to ArrayHashMap and MultiArrayList

andrewrk:
This also adds `std.sort.sortContext` and
`std.sort.insertionSortContext` which are more advanced methods that
allow overriding the `swap` method. The former calls the latter for now
because reworking the main sort implementation is a big task that can be
done later without any changes to the API.

### DIFF
--- a/lib/std/multi_array_list.zig
+++ b/lib/std/multi_array_list.zig
@@ -392,6 +392,34 @@ pub fn MultiArrayList(comptime S: type) type {
             return result;
         }
 
+        /// `ctx` has the following method:
+        /// `fn lessThan(ctx: @TypeOf(ctx), a_index: usize, b_index: usize) bool`
+        pub fn sort(self: Self, ctx: anytype) void {
+            const SortContext = struct {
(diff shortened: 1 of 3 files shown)


lib/std/multi_array_list.zig
```diff
@@ -392,6 +392,34 @@ pub fn MultiArrayList(comptime S: type) type {
...
+        /// `ctx` has the following method:
+        /// `fn lessThan(ctx: @TypeOf(ctx), a_index: usize, b_index: usize) bool`
+        pub fn sort(self: Self, ctx: anytype) void {
+            const SortContext = struct {
+                sub_ctx: @TypeOf(ctx),
+                slice: Slice,
+
+                pub fn swap(sc: @This(), a_index: usize, b_index: usize) void {
+                    inline for (fields) |field_info, i| {
+                        if (@sizeOf(field_info.field_type) != 0) {
```
mitchellh, line 404: Love this little optimization.

mitchellh (approved):
Thank you!

PR #11113, opened 2022-03-10
stage2: if generic function evaluates to another generic function call it inline

Vexu:


andrewrk:
How can a generic function evaluate to another generic function?

Vexu:
It's generic as in having an anytype parameter or a comptime parameter or a parameter with a comptime only type.

A generic function cannot evaluate to the first two but if you use `@TypeOf` on an anytype parameter you could end up with the third kind as demonstrated in the commit message:
```zig
fn foo(a: anytype, b: @typeof(a)) void { _ = b; }
test {
    // foo evaluates to `fn (type) void` and must be called inline
    foo(u32, u32);
}
```

Of course I could just be misunderstanding something, I haven't spend that much time trying to understand the way generic functions work in stage2.

andrewrk:
Ah I didn't see the commit message at first, sorry about that.

This makes sense, thanks!

Vexu:
Np, I should have copied it into the description.

PR #11118, opened 2022-03-10
std: Add elf.EM, coff.MachineType to target.Arch

moosichu:
target.Arch already supports finding the correct encoding for either
target, so being able to do the inverse has use cases for when parsing
files of an unknown target (i.e. for zar).

### DIFF
--- a/lib/std/coff.zig
+++ b/lib/std/coff.zig
@@ -62,6 +62,21 @@ pub const MachineType = enum(u16) {
     Thumb = 0x1c2,
     /// MIPS little-endian WCE v2
     WCEMIPSV2 = 0x169,
+
+    pub fn toTargetCpuArch(machine_type: MachineType) ?std.Target.Cpu.Arch {
+        return switch (machine_type) {
+            .ARM => .arm,
+            .POWERPC => .powerpc,
+            .RISCV32 => .riscv32,
+            .Thumb => .thumb,
+            .I386 => .i386,
+            .ARM64 => .aarch64,
+            .RISCV64 => .riscv64,
+            .X64 => .x86_64,
+            // there's cases we don't (yet) handle
+            else => null,
+        };
+    }
 };
 
 // OptionalHeader.magic values
--- a/lib/std/elf.zig
+++ b/lib/std/elf.zig
@@ -1481,6 +1481,36 @@ pub const EM = enum(u16) {
     _BPF = 247,
 
     _,
+
+    pub fn toTargetCpuArch(em: EM) ?std.Target.Cpu.Arch {
+        return switch (em) {
+            ._AVR => .avr,
+            ._MSP430 => .msp430,
+            ._ARC => .arc,
+            ._ARM => .arm,
+            ._HEXAGON => .hexagon,
+            ._68K => .m68k,
+            ._MIPS => .mips,
+            ._MIPS_RS3_LE => .mipsel,
+            ._PPC => .powerpc,
+            ._SPARC => .sparc,
+            ._386 => .i386,
+            ._XCORE => .xcore,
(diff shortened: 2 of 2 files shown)


moosichu:
The motivating example for this PR is that zar is using functionality in zld to parse object files - but it needs to be able to infer the target cpu architecture from those files as the parser takes that as an argument. As the standard library already includes a conversion going one way, having the conversion that goes the other way available seems to make sense.

lib/std/coff.zig
```diff
@@ -62,6 +62,21 @@ pub const MachineType = enum(u16) {
     Thumb = 0x1c2,
     /// MIPS little-endian WCE v2
     WCEMIPSV2 = 0x169,
+
+    pub fn toTargetarch(machine_type: MachineType) std.target.Arch {
```
kubkon, line 66: I think this a typo, isn't it?
```suggestion
    pub fn toTargetCpuArch(machine_type: MachineType) std.Target.Cpu.Arch {
```

lib/std/elf.zig
```diff
@@ -1481,6 +1481,38 @@ pub const EM = enum(u16) {
     _BPF = 247,
 
     _,
+
+    pub fn toTargetarch(em: EM) std.target.Arch {
```
kubkon, line 1485: Ditto.
```suggestion
    pub fn toTargetCpuArch(em: EM) std.Target.Cpu.Arch {
```

moosichu: Cheers! Checking in fixes now! Always miss this kind of stuff with simple changes :facepalm:

lib/std/coff.zig
```diff
@@ -62,6 +62,21 @@ pub const MachineType = enum(u16) {
...
+    pub fn toTargetCpuArch(machine_type: MachineType) std.target.Arch {
+        return switch (machine_type) {
+            .ARM => .arm,
+            .POWERPC => .powerpc,
+            .RISCV32 => .riscv3,
+            .Thumb => .thumb,
+            .Thumb => .thumbeb,
+            .I386 => .i386,
+            .ARM64 => .aarch64_32,
+            .ARM64 => .aarch64,
```
joachimschmidt557, line 75: There is a duplicate switch item here. Probably it should just be `.aarch64`?

moosichu: you are correct! I'm surprised that isn't a compiler error.

joachimschmidt557: Semantic Analysis should detect these duplicate switch branches and report an error, but I think this function is not referenced so it isn't analyzed.

moosichu: hmm - is there something I should be doing to force reference these then as part of the standard library? Either way I have checked a fix for this issue in now.

joachimschmidt557: There is https://github.com/ziglang/zig/blob/5fbae9cd6fe7a53d85ad6dd20cab75f7b835f0ad/lib/std/testing.zig#L508 but I don't know the conventions of the standard library well enough to say if its use is ok or if its not recommended.

joachimschmidt557 (changes requested):

lib/std/elf.zig
```diff
@@ -1481,6 +1481,36 @@ pub const EM = enum(u16) {
     _BPF = 247,
 
     _,
+
+    pub fn toTargetCpuArch(em: EM) std.target.Arch {
+        return switch (em) {
+            ._AVR => .avr,
```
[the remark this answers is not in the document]
moosichu: yeah - I actually finally have access to a computer again, so will properly be able to resolve this stuff!

joachimschmidt557 (approved):

PR #11122, opened 2022-03-11
stage2: @hasField for anon structs

mitchellh:
Subject self-explanatory.

Augmented behavior test, but for extra motivation, this is exercised by `lib/std/mem.zig` which is where I found it.

PR #11123, opened 2022-03-11
stage2: Array len field should be a usize not comptime_int

mitchellh:
Matches behavior of stage1, makes `std.mem` tests move forward a bit more when paired with #11122

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -14729,7 +14729,7 @@ fn fieldVal(
         .Array => {
             if (mem.eql(u8, field_name, "len")) {
                 return sema.addConstant(
-                    Type.comptime_int,
+                    Type.usize,
                     try Value.Tag.int_u64.create(arena, inner_ty.arrayLen()),
                 );
             } else {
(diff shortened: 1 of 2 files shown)


Vexu:
There are other places that need to be changed as you can see in https://github.com/Vexu/zig/commit/73bf418eb16a3e385218400cdc184f5d5e5aeaad but changing those seems to break some tests so you'll need to do some investigating.

mitchellh:
Thanks, I can try to take a look. I was just whack-a-mole-ing issues as I was chasing down `std.mem.Allocator` tests.

mitchellh:
@Vexu I brought in those changes and all behavior tests still pass for me locally. I'll wait to see what the CI shows maybe its backend specific. Or maybe other FE improvements fixed those...

Vexu:
Did you run `zig build test-stage2`? That's where the errors were occurring last time if I remember correctly.

mitchellh:
> Did you run `zig build test-stage2`? That's where the errors were occurring last time if I remember correctly.

Yes, passes for my environment (`linux/aarch64`).

andrewrk (approved):

PR #11127, opened 2022-03-11
Added top-level doc comments for base64

guidoism:


lib/std/base64.zig
```diff
@@ -1,3 +1,34 @@
+//! This module provides functions for encoding binary data to
+//! printable ASCII characters and decoding such encodings back to
+//! binary data.
+//!
+//! The encodings are described in RFC 4648.
+//!
+//! There are four variants defined here:
```
matu3ba, line 7: more than 3 things should have a number

lib/std/base64.zig
```diff
@@ -1,3 +1,34 @@
...
+//!
+//! standard:
```
matu3ba, line 9: why waste space here? You can make it very dense without losing readability, ie
```txt
1. standard:         Standard Base64 codecs + padding
2. standard_no_pad:  Standard Base64 codecs - padding
3. url_safe:         URL-safe Base64 codecs + padding
4. url_safe_no_pad:  URL-safe Base64 codecs - padding
```
or use `with` and `without`

lib/std/base64.zig
```diff
@@ -1,3 +1,34 @@
...
+//! standard_no_pad:
+//!   Standard Base64 codecs, without padding
+//!
+//! url_safe:
+//!   URL-safe Base64 codecs, with padding
+//!
+//! url_safe_no_pad:
+//!   URL-safe Base64 codecs, without padding
+//!
+//! Example:
```
matu3ba, line 21: we have fn doc for this.
The example can reference a specific function with
```zig
test fnname {
    coolexample
    ...
}

Code should be always tested, if possible. No dead code.

**Review by matu3ba — CHANGES_REQUESTED** [review]
make stuff more dense and no untested examples

File: lib/std/base64.zig (hunk 3 above)
Comment (guidoism) [reply lib/std/base64.zig]: Roger. That is a good policy. I saw that there was talk of the roughness of the docs keeping the newbs out before stuff is ready for them. I started writing these docs to provide easy onboarding for new users. Maybe we don't want that yet?

Comment (matu3ba) [reply lib/std/base64.zig]: > I started writing these docs to provide easy onboarding for new users. Maybe we don't want that yet?

Comments have a specific purpose, ie top level docs giving an overview if it is feasible and meaningful.
Personally I prefer to keep top level docs very short with text descriptions, but this can be often seen as a taste thing.

**Comment by matu3ba** [comment]
the test block is broken.

File: lib/std/base64.zig [hunk 4]
```diff
@@ -313,6 +326,8 @@ test "base64 url_safe_no_pad" {
     comptime try testAllApis(url_safe_no_pad, "comptime", "Y29tcHRpbWU");
 }
 
+test ""
```
matu3ba, line 329: what is this? remove it, if unclear.

matu3ba:
fmt --check is still failing.

[2/3] Install the project...
-- Install configuration: "Debug"
-- Installing: /workspace/_debug/staging/bin/zig
-- Installing: /workspace/_debug/staging/lib
+ cd /workspace
+ /workspace/_debug/staging/bin/zig fmt --check .
./_debug/staging/lib/zig/std/base64.zig
./lib/std/base64.zig

andrewrk:
Two months old; failing CI. Closing.

PR #11130, opened 2022-03-12
stage2: zirReify for enums

mitchellh:
I wasn't 100% sure I used all the correct allocators all over the place, so please take a close look there.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -12548,7 +12548,7 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
             const slice_val = fields_val.castTag(.slice).?.data;
             const decl = slice_val.ptr.castTag(.decl_ref).?.data;
             try sema.ensureDeclAnalyzed(decl);
-            const fields_len = decl.ty.arrayLen();
+            const fields_len = @intCast(usize, decl.ty.arrayLen());
             if (fields_len > 0) {
                 try enum_obj.fields.ensureTotalCapacity(new_decl_arena_allocator, fields_len);
                 try enum_obj.values.ensureTotalCapacityContext(new_decl_arena_allocator, fields_len, .{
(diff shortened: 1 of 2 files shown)


src/Sema.zig
```diff
@@ -12470,7 +12470,121 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
...
+
+            // Enum tag type
+            var buffer: Value.ToTypeBuffer = undefined;
+            enum_obj.tag_ty = try tag_type_val.toType(&buffer).copy(new_decl_arena_allocator);
+
+            // Fields
+            const slice_val = fields_val.castTag(.slice).?.data;
+            const decl = slice_val.ptr.castTag(.decl_ref).?.data;
+            try sema.ensureDeclAnalyzed(decl);
+            const fields_len = @intCast(usize, decl.ty.arrayLen());
```
andrewrk, line 12551: check out `usizeCast` so that this is a compile error instead of UB

mitchellh:
Feedback addressed, also found a bug in the error set reification (along with this) that I just caught: have to use `pointerDecl` to get the decl for a slice to account for non-const pointers.

src/Sema.zig
```diff
@@ -12470,11 +12470,171 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
...
+            // enum layout is always auto
+            const layout = layout_val.toEnum(std.builtin.Type.ContainerLayout);
+            if (layout != .Auto) {
+                return sema.fail(block, src, "reified enums must have a layout .Auto", .{});
+            }
+
+            // Decls
+            const decls_slice_val = decls_val.castTag(.slice).?.data;
+            const decls_decl = decls_slice_val.ptr.pointerDecl().?;
+            try sema.ensureDeclAnalyzed(decls_decl);
```
andrewrk, line 12497: this is a no-op because creating an anonymous decl creates it with already `complete` status.

andrewrk (approved):
Arena usage looks good.

I think in the future we may want to add new Type tag variants for reified_enum, reified_opaque, etc., so that we can avoid creating unnecessary Decl objects for these. That change will likely require touching a bunch of different places in the codebase.

mitchellh:
> I think in the future we may want to add new Type tag variants for reified_enum, reified_opaque, etc., so that we can avoid creating unnecessary Decl objects for these. That change will likely require touching a bunch of different places in the codebase.

Makes sense.

PR #11093, opened 2022-03-08
std: add test for child_process

matu3ba:
- Cli operations should be refactored, since the standard test runner
  has an expected argument structure. This would also ensure that the
  test cli is usable as tested library with checks for subprocess
  error or success instead of "hacky shell script interfaces".
- Default paths generation based on tmpDir would also be useful.
- Anonymous pipes on windows are generated from named pipes
- Async IO does not work on anonymous pipes
- Remove finished TODO

PR #11138, opened 2022-03-12
std.testing: add methods tmpDirPath, getTestArgs, buildExe

matu3ba:
continuation of #11093 to simplify testing IPC

* use cases
  - get path to temporary directory
  - get the test arguments inside test block for reusage
  - build executables from text within test blocks, ie to test IPC
* missing conventions
  - how to name and debug test cases
  - where do simple+repititve build commands for testing belong

### DIFF
--- a/lib/std/testing.zig
+++ b/lib/std/testing.zig
@@ -400,6 +400,56 @@ pub fn tmpDir(opts: std.fs.Dir.OpenDirOptions) TmpDir {
     };
 }
 
+/// caller owns memory
+pub fn tmpDirPath(alloc: std.mem.Allocator, tmpdir: *TmpDir) ![]u8 {
+    const cwd_str = try std.process.getCwdAlloc(alloc);
+    defer alloc.free(cwd_str);
+    const path = try std.fs.path.join(alloc, &[_][]const u8{
+        cwd_str,
+        "zig-cache",
+        "tmp",
+        &tmpdir.sub_path,
+    });
+    return path;
+}
+
+const TestArgs = struct {
+    testexec: [:0]const u8 = undefined,
+    zigexec: [:0]const u8 = undefined,
+};
+
+/// Get test arguments inside test block by regular test runner ('zig test file.zig')
+/// Caller must provide backing ArgIterator
+pub fn getTestArgs(it: *std.process.ArgIterator) !TestArgs {
+    var testargs = TestArgs{};
+    testargs.testexec = it.next() orelse unreachable;
+    testargs.zigexec = it.next() orelse unreachable;
+    try expect(!it.skip());
+    return testargs;
+}
+
+test "getTestArgs" {
+    var it = try std.process.argsWithAllocator(allocator);
+    const testargs = try getTestArgs(&it);
+    defer it.deinit(); // no-op unless WASI or Windows
+    try expect(testargs.testexec.len > 0); // zig compiler executable path
+    try expect(testargs.zigexec.len > 0); // test runner executable path
+}
+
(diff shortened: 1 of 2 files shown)


matu3ba:
Question for next PR: Should there be a method to create a zig program from a string inside tests or where should it reside?
I woudl envision such a method to contain something along the following with an interface `buildExeInTmp(allocator, progstr, binfile) ![]const u8` and the exe as return value.
```zig
    var it = try std.process.argsWithAllocator(allocator);
    const testargs = try testing.getTestArgs(&it);
    defer it.deinit(); // no-op unless WASI or Windows

    var tmp = testing.tmpDir(.{ .no_follow = true }); // ie zig-cache/tmp/8DLgoSEqz593PAEE
    defer tmp.cleanup();
    const tmpdirpath = try testing.tmpDirPath(allocator, &tmp);
    defer allocator.free(tmpdirpath);
    const child_name = "child"; // no need for suffixes (.exe, .wasm) due to '-femit-bin'
    const suffix_zig = ".zig";
    const child_path = try fs.path.join(allocator, &[_][]const u8{ tmpdirpath, child_name });
    defer allocator.free(child_path);
    const child_zig = try mem.concat(allocator, u8, &[_][]const u8{ child_path, suffix_zig });
    defer allocator.free(child_zig);

    try tmp.dir.writeFile("child.zig", childstr);
    try testing.buildExe(testargs.zigexec, child_zig, child_path);
```
That way, building (multiple) executables to a tmp folder would be exactly 1 properly checked function from within test blocks.

Alternatively, I will implement a shortcut to obtain `child_zig` and `child_path` (that means absolute path to zig file and absolute path to binary).

matu3ba:
Some functionality is redundant to `src/test.zig`, but unfortunately this is no available for user programs, because it is not part of `lib/std`.
Any advice how to refactor `src/test.zig` or how to proceed would be nice.

lib/std/testing.zig
```diff
@@ -400,6 +400,56 @@ pub fn tmpDir(opts: std.fs.Dir.OpenDirOptions) TmpDir {
     };
 }
 
+/// caller owns memory
+pub fn tmpDirPath(alloc: std.mem.Allocator, tmpdir: *TmpDir) ![]u8 {
```
squeek502, line 404: Any reason why this isn't in the `TmpDir` struct as something like:

```zig
pub fn getFullPath(self: TmpDir, alloc: std.mem.Allocator) ![]u8
```

matu3ba: fixed.

iddev5:
> Question for next PR:

Its a good suggestion but eventually similar functions would be needed for all other zig operations like run, buildLib etc.
What could instead be done is a function ``fn writeZigFileInTmp(allocator, progstr, binfile) ![]const u8`` which would return the final path of file (which is child_zig) in your test, on which we can then manually do ``std.fs.path.dirname(child_zig).?`` to get child_path.

iddev5 (approved):

matu3ba:
@iddev5 With your proposed solution I get
```zig
    const zigfile_path = try tmp.writeZigFile(std.testing.allocator, tmp, childstr, child_name);
    defer std.testing.allocator.free(zigfile_path);
    const binary = zigfile_path[0 .. zigfile_path.len - 4]; // '.zig' is 4 characters
    try testing.buildExe(testargs.zigexec, zigfile_path, binary);
```
because we still need the path to the resulting binary for `buildExe`.
I think requiring to copy paste from users with an example for the subslice should be fine, since
we cant do it more efficiently than the subslice.

PR #11142, opened 2022-03-13
zig-fmt: Resolve #11131 loss of comment on switch cases 

### LINKED ISSUE #11131: zig fmt: lossy switch
### Zig Version

0.10.0-dev.1253+86a98b172

### Steps to Reproduce

```zig
const std = @import("std");

const E = enum { a, b };

pub fn main() !void {
    var e = E.a;
    switch (e) {
        .a,
        // .b,
        => {},
    }
}
```

### Expected Behavior

No change.

### Actual Behavior

```zig
const std = @import("std");

const E = enum { a, b };

pub fn main() !void {
    var e = E.a;
    switch (e) {
        .a => {},
    }
}
```

curtiswilkinson:
This now checks for the presence of a comment before throwing everything on one line and dropping the comment from the switch case entirely, as mentioned in https://github.com/ziglang/zig/issues/11131

After verifying the issue being fixed, I tested for potential regressions by running it on `Sema.zig` and seeing if there was any formatting changes in the git diff. 

~~If there is a test suite (I was unable to find any sorry) for `zig fmt`, please let me know!~~

### DIFF
--- a/lib/std/zig/render.zig
+++ b/lib/std/zig/render.zig
@@ -1505,7 +1505,7 @@ fn renderSwitchCase(
     const node_tags = tree.nodes.items(.tag);
     const token_tags = tree.tokens.items(.tag);
     const trailing_comma = token_tags[switch_case.ast.arrow_token - 1] == .comma;
-    const has_comment = blk: {
+    const has_comment_before_arrow = blk: {
         if (switch_case.ast.values.len == 0) break :blk false;
         break :blk hasComment(tree, tree.firstToken(switch_case.ast.values[0]), switch_case.ast.arrow_token);
     };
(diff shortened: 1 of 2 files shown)


g-w1:
The tests are here https://github.com/ziglang/zig/blob/master/lib/std/zig/parser_test.zig#L342-L354

curtiswilkinson:
Ahhh sure. Thanks for pointing me to those @g-w1! 

I've added a corresponding test-case to cover this and verified it was failing prior to my commit.

lib/std/zig/render.zig
```diff
@@ -1505,16 +1505,18 @@ fn renderSwitchCase(
     const node_tags = tree.nodes.items(.tag);
     const token_tags = tree.tokens.items(.tag);
     const trailing_comma = token_tags[switch_case.ast.arrow_token - 1] == .comma;
+    const has_comment = blk: {
```
ifreund, line 1508: I think we should make this variable name more descriptive to avoid future misuse.
```suggestion
    const has_comment_before_arrow = blk: {
```

ifreund (commented):
LGTM aside from one small suggestion.

lib/std/zig/render.zig
curtiswilkinson: Agreed! Resolved in latest commit. Thanks for the review!

ifreund (approved):
Thanks!

PR #11144, opened 2022-03-13
stage2: reify structs and tuples

mitchellh:
Implements `@Type` for structs, anon structs, and tuples. This is another place that would probably benefit from a `.reified_struct` type tag but will defer for later in the interest of getting tests passing first.

I'm not sure I got all the struct status fields right so please check that carefully...

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -12478,17 +12478,16 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
             return sema.addType(ty);
         },
         .Struct => {
-            const struct_val = union_val.val.castTag(.@"struct").?.data;
             // TODO use reflection instead of magic numbers here
-            // error_set: type,
-            // layout: ContainerLayout,
+            const struct_val = union_val.val.castTag(.@"struct").?.data;
+            // layout: containerlayout,
             const layout_val = struct_val[0];
-            // fields: []const EnumField,
+            // fields: []const enumfield,
             const fields_val = struct_val[1];
-            // decls: []const Declaration,
+            // decls: []const declaration,
             const decls_val = struct_val[2];
             // is_tuple: bool,
-            // TODO const is_tuple_val = struct_val[4];
+            const is_tuple_val = struct_val[3];
 
             // Decls
             const decls_slice_val = decls_val.castTag(.slice).?.data;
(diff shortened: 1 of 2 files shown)


mitchellh:
There a missing usize cast. I’ll get it tomorrow.

src/Sema.zig
```diff
@@ -12468,10 +12477,34 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
...
+            const decls_decl = decls_slice_val.ptr.pointerDecl().?;
+            try sema.ensureDeclAnalyzed(decls_decl);
+            if (decls_decl.ty.arrayLen() > 0) {
+                return sema.fail(block, src, "reified structs must have no decls", .{});
+            }
+
+            return if (is_tuple_val.toBool())
+                sema.reifyTuple(block, src, fields_val)
+            else
+                sema.reifyStruct(block, inst, src, layout_val, fields_val);
```
andrewrk, line 12503: It's good hygiene to put the `try` in each of the prongs here like this

```suggestion
                try sema.reifyTuple(block, src, fields_val)
            else
                try sema.reifyStruct(block, inst, src, layout_val, fields_val);
```

because in an error return trace, it will point at the `try` in the stack frame, telling you which prong was taken. This is similarly why you see `return error.Foo` in each prong of switch cases rather than `return switch(...`

src/Sema.zig
```diff
@@ -12468,10 +12477,34 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
...
+            // decls: []const declaration,
+            const decls_val = struct_val[2];
+            // is_tuple: bool,
+            const is_tuple_val = struct_val[3];
+
+            // Decls
+            const decls_slice_val = decls_val.castTag(.slice).?.data;
+            const decls_decl = decls_slice_val.ptr.pointerDecl().?;
+            try sema.ensureDeclAnalyzed(decls_decl);
+            if (decls_decl.ty.arrayLen() > 0) {
```
andrewrk, line 12496: This code makes too many assumptions. Just check the comptime slice length:

```suggestion
            if (decls_val.sliceLen() > 0) {
```

If I remember correctly this affects other reification code too.

src/Sema.zig
```diff
@@ -12637,6 +12670,180 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
...
 
+fn reifyTuple(
+    sema: *Sema,
+    block: *Block,
+    src: LazySrcLoc,
+    fields_val: Value,
+) CompileError!Air.Inst.Ref {
+    const slice_val = fields_val.castTag(.slice).?.data;
+    const decl = slice_val.ptr.pointerDecl().?;
+    try sema.ensureDeclAnalyzed(decl);
```
andrewrk, line 12681: You're not guaranteed to have a `slice` tag, nor a `Decl` here. Just use `fields_val.slicePtr()` and `fields_val.sliceLen()`.

Note there is a handy `empty_struct_literal` for the case of empty tuple.

src/Sema.zig
```diff
@@ -12637,6 +12670,180 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
...
+        },
+    };
+
+    // Fields
+    const slice_val = fields_val.castTag(.slice).?.data;
+    const decl = slice_val.ptr.pointerDecl().?;
+    try sema.ensureDeclAnalyzed(decl);
+    const fields_len = try sema.usizeCast(block, src, decl.ty.arrayLen());
+    if (fields_len > 0) {
+        struct_obj.status = .have_field_types;
```
andrewrk, line 12795: status looks good :+1:

src/Sema.zig
```diff
@@ -12637,6 +12670,180 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
...
+            .ty = struct_ty,
+            .file_scope = block.getFileScope(),
+        },
+    };
+
+    // Fields
+    const slice_val = fields_val.castTag(.slice).?.data;
+    const decl = slice_val.ptr.pointerDecl().?;
+    try sema.ensureDeclAnalyzed(decl);
+    const fields_len = try sema.usizeCast(block, src, decl.ty.arrayLen());
```
andrewrk, line 12793: same deal here, use the slice methods of Value. Also if there end up being no fields, then I think you can just return `Type.initTag(.empty_struct)`, no Decl required. There's even an `Air.Inst.Ref` tag for it. You can straight up return that from the function.

andrewrk (changes requested):
Found some stuff that I think is worth changing-

src/Sema.zig
mitchellh: Ah hah! TIL, thank you, this felt wrong! Fixed it up everywhere I could find it.

mitchellh:
Thank you!

Yes, all the array deref stuff felt really wrong. I wasn't aware of the value funcs around that and I'm glad they already exist. I swapped all usage with that. I addressed all the other feedback too. Thank you.

src/Sema.zig
```diff
@@ -12637,6 +12651,174 @@ fn zirReify(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.I
...
 }
 
+fn reifyTuple(
+    sema: *Sema,
+    block: *Block,
+    src: LazySrcLoc,
+    fields_val: Value,
+) CompileError!Air.Inst.Ref {
+    const fields_len = try sema.usizeCast(block, src, fields_val.sliceLen());
+    if (fields_len == 0) return Air.Inst.Ref.empty_struct;
```
andrewrk, line 12661: The only problem with this is that it returns `false` for isTuple:

```zig
    pub fn isTuple(ty: Type) bool {
        return switch (ty.tag()) {
            .tuple, .empty_struct_literal => true,
            else => false,
        };
    }
```

Two options to fix this:

1. Add a new Air.Inst.Ref tag for `empty_struct_literal`
2. Do the `addConstant` dance.

----

After typing this up I realized a mistake I made in my previous comment. `Air.Inst.Ref.empty_struct` is a *value*, not a *type*. So that won't work. Both cases will have to use `addConstant(Type.type, ...)`, with `Value.Tag.ty`. (either that or adding more Value Tags such as `empty_struct_type` and `empty_tuple_type`.

mitchellh: Ah yes. Thanks. I’ll fix this up tomorrow.

mitchellh: Okay I got this fixed. 

For tuples, I just didn't specially handle `fields_len == 0` because tuples don't require a decl anyways so the logic below that is more or less free and not worth the special handling.

For structs, I created a `empty_struct_literal` type.

mitchellh:
Ugh, actually, the behavior doesn't exactly match stage1. With stage1 `@Type(@typeInfo(struct {}))` marks `is_tuple` as false, but `empty_struct_literal` has it as `true` in stage2. I'll have to clean this up tomorrow, I'm sorry!

PR #11147, opened 2022-03-13
stage2: add debug info for globals in the LLVM backend

### LINKED ISSUE #11097: implement debug info for global variables in the llvm backend
hello.zig

```zig
var blah: i32 = 1;

pub fn main() void {
    foo();
    blah = add(blah, 2);
}

fn foo() void {}

fn add(a: i32, b: i32) i32 {
    const result = a + b;
    return result;
}
```

```
gdb ./hello
(gdb) break hello.main
Breakpoint 1 at 0x2011b4: file hello.zig, line 4.
(gdb) run
Starting program: /home/andy/dev/zig/build-release/hello 

Breakpoint 1, hello.main () at hello.zig:4
4	    foo();
(gdb) info variables
All defined variables:

Non-debugging symbols:
...
0x0000000000202254  hello.blah
(gdb) p 'hello.blah'
'hello.blah' has unknown type; cast it to its declared type
(gdb) p (i32)'hello.blah'
$2 = 1
```

It looks like gdb is cleverly noticing symbols and treating them like typeless global variables. But to solve this issue we would expect to see the global variable in the "All defined variables" list, and it should have a type associated with it so that we do not have to manually cast it.

We would also expect to see some output from:

```
readelf --debug-dump=info ./hello | grep variable
```

wsengir:
LLVM backend: generate `DIGlobalVariable`s for non-function globals and
rename linkage names when exporting functions and globals.

`zig_llvm.cpp`: add some wrappers to convert a handful of DI classes
into `DINode`s since `DIGlobalVariable` is not a `DIScope` like the others.

`zig_llvm.cpp`: add some wrappers to allow replacing the `LinkageName` of
`DISubprogram` and `DIGlobalVariable`.

`zig_llvm.cpp`: fix DI class mixup causing nonsense reinterpret_cast.

The end result is that GDB is now usable since you now no longer need
to manually cast every global nor fully qualify every export.

Closes #11097.

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -663,7 +663,7 @@ pub const Decl = struct {
         return (try decl.typedValue()).val;
     }
 
-    pub fn isFunction(decl: *Decl) !bool {
+    pub fn isFunction(decl: Decl) !bool {
         const tv = try decl.typedValue();
         return tv.ty.zigTypeTag() == .Fn;
     }
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -164,8 +164,9 @@ pub const Object = struct {
     di_builder: ?*llvm.DIBuilder,
     /// One of these mappings:
     /// - *Module.File => *DIFile
-    /// - *Module.Decl => *DISubprogram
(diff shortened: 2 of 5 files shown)


andrewrk:
Excellent work!

Issue #11152, opened 2022-03-13
Function only called at comptime is emitted to binary


Vexu:
Both stage1 and stage2 will emit an error for `foo` because they try to emit is as a runtime function even though it is only ever used at comptime.
```zig
fn foo() void {
    var a = false;
    if (a) @compileError("analyzed as runtime");
}
comptime {
    _ = foo();
}
```

andrewrk:
Duplicate of #6256.

Related:
 * #425
 * #7772

PR #11153, opened 2022-03-13
stage2: add error for returning pointer to local variable

### LINKED ISSUE #2646: Add Compiler Error when returning pointer to stack variable
I'm new to zig and low level programming in general. Naturally when I encountered the code for setting up an allocater I tried to encapsulate it.

```
    var direct = std.heap.DirectAllocator.init();
    defer direct.deinit();
    var arena = std.heap.ArenaAllocator.init(&direct.allocator);
    defer arena.deinit();
```

I tried to (in retrospect naively) create a struct to take care of the repetition in my tests:

```
const MyStruct = struct {
    
    _alloc1: *std.heap.DirectAllocator,
    _alloc2: *std.heap.ArenaAllocator,

    pub fn init() !MyStruct {
        var direct = std.heap.DirectAllocator.init();
        var arena = std.heap.ArenaAllocator.init(&direct.allocator);
        return MyStruct{
            ._alloc1 = &direct,
            ._alloc2 = &arena,
        };
    }
    pub fn deinit(self: MyStruct) void {
        self._alloc2.deinit();
        self._alloc1.deinit();
    }
};
```

When using this I get a segfault though. 

In #zig @shawnl was helpful enough to explain the issue. The allocators are created on the stack of the `init` function and so the pointers to them that are set on the struct are invalid once the `init` function returns.

I am opening this issue because @shawnl claims there may be a way to figure out when such pointers are created and survive beyond the lifetime of the stack variables.

Vexu:
The check does not apply to inline functions and wont apply to functions called at comptime depending on #11152.
Most simple cases are caught but it is easy to bypass if needed.

Some cases might be too aggressive but they are easy to remove if deemed so.

Closes  #2646

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -64,6 +64,8 @@ comptime_args_fn_inst: Zir.Inst.Index = 0,
 /// Sema will set this to null when it takes ownership.
 preallocated_new_func: ?*Module.Fn = null,
 
+stack_pointers: std.AutoHashMapUnmanaged(Air.Inst.Ref, LazySrcLoc) = .{},
+
 const std = @import("std");
 const mem = std.mem;
 const Allocator = std.mem.Allocator;
@@ -506,6 +508,7 @@ pub fn deinit(sema: *Sema) void {
     sema.air_values.deinit(gpa);
     sema.inst_map.deinit(gpa);
     sema.decl_val_table.deinit(gpa);
+    sema.stack_pointers.deinit(gpa);
     sema.* = undefined;
 }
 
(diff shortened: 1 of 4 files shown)


andrewrk:
One thing we need to do here is make sure we can manage to make the language specification compatible with this compile error.

cc @SpexGuy

Jarred-Sumner:
> This is valid zig code

I do this currently when implementing a large interface for something function by function

Just so the code compiles even though the implementation is incomplete

andrewrk:
@Jarred-Sumner why not instead `@panic("TODO")`? Or even just `unreachable`?

andrewrk:
Closing since we currently have no workable plan to make this a legal compile error according to the lang spec. We can revisit this patch when dealing with #2646 which now has a link to this closed PR. **The next step towards landing this patch is coming up with a lang spec proposal that allows this compile error.** Please make such proposals as comments on #2646.

PR #11156, opened 2022-03-14
Error when binary ops don't have matching whitespace on both sides. 

### LINKED ISSUE #7399: Proposal: Require whitespace on both sides of - or neither
This may be out of scope for the language, but it's in line with other accepted proposals like #35 and #114 which aim to avoid misinterpretation.
Here's an example of the problem this prevents:
```zig
const directions = [_]isize{
    -pitch-1,
    -pitch,
    -pitch+1
    -1,
    1,
    pitch-1,
    pitch,
    pitch+1,
};
```
See the bug?  It took me a minute.

<details>
  <summary>View Answer</summary>
  Here the `,` is missing from the third item in the list.  This makes the list 7 items long instead of 8, and the third item `-pitch+1-1`.  If you look closely, you can see that GitHub's syntax highlighting parser actually identifies the bug, and gives the `-` on line 4 a different color to indicate that it's not unary.
</details>

My proposal to solve this is to require whitespace either both before and after `-`, or neither before nor after.  We may also want to apply this to other operators for consistency, but I think `-` is the one that really causes problems.  Here are some examples under this rule:
```zig
_ = a - b; // fine
_ = a-b; // fine
_ = a -b; // error
_ = a- b; // error
_ = some_really_long_thing -
    some_other_really_long_thing; // ok
_ = some_really_long_thing
    - some_other_really_long_thing; // ok
_ = some_really_long_thing-
    some_other_really_long_thing; // error
_ = some_really_long_thing
    -some_other_really_long_thing; // error
```

danielchasehooper:
Fixes #7399

This change also moves the warning about `&&` from the AstGen into the parser so that the `&&` warning can supersede the whitespace warning.

This is my first PR for Zig, so be gentle :) Let me know if there is anything that can be improved.


ifreund (changes requested):
I don't think moving the error for `&&` to the parser is ok, at least not with the current implementation. `&&foo` where the `&`s are the unary address of operator is valid. See #8046 and #9114.

danielchasehooper:
@ifreund `&&foo` is parsed as `bitwise_and address_of foo`, and not `address_of address_of foo` correct? I didn't think you could get the address of an address. You can write it as `bar & &foo` to get the compiler to accept the code, which i think is clearer anyway. 

I could make the error message clearer so that if the right hand side of bitwise_and parses as an address_of, it could suggest adding a space '& &' , otherwise it would suggest using `and`

danielchasehooper:
FWIW: this is what the current behavior of the compiler is for `&&foo`

```
./addressof.zig:6:51: error: `&&` is invalid. Note that `and` is boolean AND
    &&foo
     ^
```

This PR doesn't change that behavior.

danielchasehooper:
Ok, so this is somewhat complicated by the fact that the *existing* code in AstGen.zig currently doesn't accept `bar&&foo` or `bar && foo`, even when you mean `bitwise_and address_of`. You currently have to write it as `bar& &foo` to appease the compiler.
https://github.com/ziglang/zig/blob/5ea94e7715607e986298908536cdd3d9dfdd0ce9/src/AstGen.zig#L672-L688

I don't think the compiler can recommend `'&&' should be 'and'` for `foo&&bar` until types are resolved, because without types you don't know if it should be interpreted as `boolean_expr logical_and boolean_expr` or `expr bitwise_and addressof expr` which, I'm sure, is why `&&` was changed to `and` in the first place. 

I see two different approaches
1. Make the parser require whitespace before and after the bitwise_and operator if it is followed by an address_of operator. Also, require there be no whitespace after address_of `&`. If the whitespace is wrong for bitwise_and it will give you the two options: use `and` or change the whitespace. 
2. Revert my AstGen change, remove my special case for `&&` in the parser. Make the tests that previously expected a `'&&' should be 'and'` error to expect a `whitespace mismatch` error if they are testing `bar &&foo`, but if they are testing `bar&&foo`, they will continue to get the existing `&&` error from AstGen.zig. 

I think approach 1 is correct. Either way, I think a separate issue should be created to move the `'&&' should be 'and'` recommendation till after types are resolved.

Thoughts?

danielchasehooper:
Ok, `&&foo` works and `foo&&bar` requires either using `and` or ` & &`.

matu3ba:
The zig fmt check fails, which is only run in `zinc/linux_test.sh`. ie ci.ziglang/dron/pr

Is this also failing locally for you?

danielchasehooper:
@matu3ba Couple questions:
1. (maybe stupid question) I'm on macOS, is `zinc/linux_test.sh` linux only? 
2. How do I run it? It looks like it depends on specific environment variables/paths to be set up to work
3. Are the CI tests run on my branch, or on what would result from the merge of this PR into master?
4. Is there a way to see if `zinc/linux_test.sh` was passing before my changes?

andrewrk:
Hi @danielchasehooper - check CONTRIBUTING.md for some instructions on how to run the tests locally.

danielchasehooper:
@matu3ba I misread your message at first. Yes it was failing locally for me, but I've run zig fmt and we're all good now.

@andrewrk `zig fmt --check` doesn't seem to get run in local tests like it is in the `zinc/linux_test.sh`, so running local tests wouldn't have caught this for me. Does that sound correct to you?

Additionally, `zig build test` is failing on my arm64 macOS machine as of `cb3b1dd6ddee65d1811fb4058b5cc0f2c06d0139` (i.e. before any of my changes). So I couldn't even run the full tests to check my changes. Do you know anything about this? Output of failing test below:

```

$ zig build test

...passing tests removed for brevity...

Test 13/13 stack-trace dumpCurrentStackTrace (Debug)...
source.zig:7:8: [address] in foo (test)
    bar();
       ^
source.zig:10:8: [address] in main (test)
    foo();
       ^

error: TestFailed
/Users/daniel/Developer/zig/lib/std/fs.zig:1285:9: 0x1021a9156 in std.fs.Dir.makeDir (build)
        try os.mkdirat(self.fd, sub_path, default_new_dir_mode);
        ^
/Users/daniel/Developer/zig/lib/std/os.zig:2757:19: 0x1021a95a0 in std.os.mkdiratZ (build)
        .EXIST => return error.PathAlreadyExists,
                  ^
/Users/daniel/Developer/zig/lib/std/os.zig:2717:9: 0x1021a92a8 in std.os.mkdirat (build)
        return mkdiratZ(dir_fd, &sub_dir_path_c, mode);
        ^
/Users/daniel/Developer/zig/lib/std/fs.zig:1285:9: 0x1021a9156 in std.fs.Dir.makeDir (build)
        try os.mkdirat(self.fd, sub_path, default_new_dir_mode);
        ^
/Users/daniel/Developer/zig/lib/std/os.zig:2757:19: 0x1021a95a0 in std.os.mkdiratZ (build)
        .EXIST => return error.PathAlreadyExists,
                  ^
/Users/daniel/Developer/zig/lib/std/os.zig:2717:9: 0x1021a92a8 in std.os.mkdirat (build)
        return mkdiratZ(dir_fd, &sub_dir_path_c, mode);
        ^
/Users/daniel/Developer/zig/lib/std/fs.zig:1285:9: 0x1021a9156 in std.fs.Dir.makeDir (build)
        try os.mkdirat(self.fd, sub_path, default_new_dir_mode);
        ^
/Users/daniel/Developer/zig/test/tests.zig:853:17: 0x102289114 in .@build.test.tests.RunAndCompareStep.make (build)
                return error.TestFailed;
                ^
/Users/daniel/Developer/zig/lib/std/build.zig:3297:9: 0x1021a9ebe in std.build.Step.make (build)
        try self.makeFn(self);
        ^
/Users/daniel/Developer/zig/lib/std/build.zig:506:9: 0x1021a8ad1 in std.build.Builder.makeOneStep (build)
        try s.make();
        ^
/Users/daniel/Developer/zig/lib/std/build.zig:500:17: 0x1021a8a55 in std.build.Builder.makeOneStep (build)
                return err;
                ^
/Users/daniel/Developer/zig/lib/std/build.zig:500:17: 0x1021a8a55 in std.build.Builder.makeOneStep (build)
                return err;
                ^
/Users/daniel/Developer/zig/lib/std/build.zig:500:17: 0x1021a8a55 in std.build.Builder.makeOneStep (build)
                return err;
                ^
/Users/daniel/Developer/zig/lib/std/build.zig:461:13: 0x102198c5b in std.build.Builder.make (build)
            try self.makeOneStep(s);
            ^
/Users/daniel/Developer/zig/lib/std/special/build_runner.zig:209:21: 0x102193dcd in main (build)
            else => return err,
                    ^
/Users/daniel/Developer/zig/lib/std/os.zig:2757:19: 0x1021a95a0 in std.os.mkdiratZ (build)
        .EXIST => return error.PathAlreadyExists,
                  ^
/Users/daniel/Developer/zig/lib/std/os.zig:2717:9: 0x1021a92a8 in std.os.mkdirat (build)
        return mkdiratZ(dir_fd, &sub_dir_path_c, mode);
        ^
/Users/daniel/Developer/zig/lib/std/fs.zig:1285:9: 0x1021a9156 in std.fs.Dir.makeDir (build)
        try os.mkdirat(self.fd, sub_path, default_new_dir_mode);
        ^
/Users/daniel/Developer/zig/lib/std/os.zig:2757:19: 0x1021a95a0 in std.os.mkdiratZ (build)
        .EXIST => return error.PathAlreadyExists,
                  ^
/Users/daniel/Developer/zig/lib/std/os.zig:2717:9: 0x1021a92a8 in std.os.mkdirat (build)
        return mkdiratZ(dir_fd, &sub_dir_path_c, mode);
        ^
/Users/daniel/Developer/zig/lib/std/fs.zig:1285:9: 0x1021a9156 in std.fs.Dir.makeDir (build)
        try os.mkdirat(self.fd, sub_path, default_new_dir_mode);
        ^
/Users/daniel/Developer/zig/lib/std/os.zig:2757:19: 0x1021a95a0 in std.os.mkdiratZ (build)
        .EXIST => return error.PathAlreadyExists,
                  ^
/Users/daniel/Developer/zig/lib/std/os.zig:2717:9: 0x1021a92a8 in std.os.mkdirat (build)
        return mkdiratZ(dir_fd, &sub_dir_path_c, mode);
        ^
/Users/daniel/Developer/zig/lib/std/fs.zig:1285:9: 0x1021a9156 in std.fs.Dir.makeDir (build)
        try os.mkdirat(self.fd, sub_path, default_new_dir_mode);
        ^
/Users/daniel/Developer/zig/lib/std/os.zig:2757:19: 0x1021a95a0 in std.os.mkdiratZ (build)
        .EXIST => return error.PathAlreadyExists,
                  ^
/Users/daniel/Developer/zig/lib/std/os.zig:2717:9: 0x1021a92a8 in std.os.mkdirat (build)
        return mkdiratZ(dir_fd, &sub_dir_path_c, mode);
        ^
/Users/daniel/Developer/zig/lib/std/fs.zig:1285:9: 0x1021a9156 in std.fs.Dir.makeDir (build)
        try os.mkdirat(self.fd, sub_path, default_new_dir_mode);
        ^
/Users/daniel/Developer/zig/lib/std/os.zig:2757:19: 0x1021a95a0 in std.os.mkdiratZ (build)
        .EXIST => return error.PathAlreadyExists,
                  ^
/Users/daniel/Developer/zig/lib/std/os.zig:2717:9: 0x1021a92a8 in std.os.mkdirat (build)
        return mkdiratZ(dir_fd, &sub_dir_path_c, mode);
        ^
/Users/daniel/Developer/zig/lib/std/fs.zig:1285:9: 0x1021a9156 in std.fs.Dir.makeDir (build)
        try os.mkdirat(self.fd, sub_path, default_new_dir_mode);
        ^
/Users/daniel/Developer/zig/lib/std/os.zig:2757:19: 0x1021a95a0 in std.os.mkdiratZ (build)
        .EXIST => return error.PathAlreadyExists,
                  ^
/Users/daniel/Developer/zig/lib/std/os.zig:2717:9: 0x1021a92a8 in std.os.mkdirat (build)
        return mkdiratZ(dir_fd, &sub_dir_path_c, mode);
        ^
/Users/daniel/Developer/zig/zig-cache/o/0805107800bfebdbbdeee4dd304f1974/build /Users/daniel/Developer/zig/build/zig /Users/daniel/Developer/zig /Users/daniel/Developer/zig/zig-cache /Users/daniel/.cache/zig test

```

schmee:
@danielchasehooper `zig build test` will run all the tests on all the backends in all the release modes, which will probably not work. Feel free to ask build questions in `#stage2-meetings` in Discord or ping me directly (same name as here) and I'll try to help you get up and running.

lib/std/zig/Ast.zig
```diff
@@ -328,7 +328,12 @@ pub fn renderError(tree: Ast, parse_error: Error, stream: anytype) !void {
         .expected_initializer => {
             return stream.writeAll("expected field initializer");
         },
-
+        .mismatched_binary_op_whitespace => {
+            return stream.print("binary operator `{s}` has whitespace on one side, but not the other.", .{token_tags[parse_error.token].lexeme().?});
+        },
+        .invalid_ampersand_ampersand => {
+            return stream.writeAll("Ambiguous use of '&&'. Instead, use 'and' for logical and, or change whitespace to ' & &' for bitwise and");
```
Vexu, line 335:
```suggestion
            return stream.writeAll("ambiguous use of '&&'; use 'and' for logical AND, or change whitespace to ' & &' for bitwise AND");
```
Error messages should be lower case to match rest of the compiler.

Vexu (approved):

PR #11173, opened 2022-03-15
std.dwarf: implement basic DWARF 5 parsing (and fix stack traces!)

wsengir:
DWARF 5 moves around some fields and adds a few new ones that can't be
parsed or ignored by our current DWARF 4 parser. This isn't a complete
implementation of DWARF 5, but this is enough to make stack traces
mostly work. Line numbers from C++ don't show up, but I know the info
is there. I think the answer is to iterate through `.debug_line_str` in
`getLineNumberInfo`, but I didn't want to fall into an even deeper rabbit
hole tonight.


andrewrk (approved):
Nice work!

PR #11177, opened 2022-03-15
Add debug info for inlined calls

### LINKED ISSUE #11158: make function calls contain debug stmt line/column info
Zig version: `0.10.0-dev.1311+5ea94e771`

```zig
const std = @import("std");

pub fn main() void {
    var x: i32 = 1234;
    x += foo() +
        bar();
}

fn foo() i32 {
    return 1;
}

fn bar() i32 {
    @panic("crash");
}
```

Here you can see in a debugging session, the traceback points back to the wrong function call:

```
$ ./stage2/bin/zig build-exe test3.zig -fLLVM
$ gdb ./test3 -ex run
Program received signal SIGTRAP, Trace/breakpoint trap.
0x00000000002019e7 in builtin.default_panic (msg=..., error_return_trace=<optimized out>) at /home/andy/dev/zig/lib/std/builtin.zig:758
758	            @breakpoint();
(gdb) up
#1  0x0000000000202127 in test3.bar () at test3.zig:14
14	    @panic("crash");
(gdb) 
#2  0x0000000000201b42 in test3.main () at test3.zig:5
5	    x += foo() +
(gdb) 
```

This is because in the ZIR we only emit `dbg_stmt` for the var decl, not each function call:

```
      %11 = dbg_stmt(3, 5)
      %12 = load(%7) node_offset:5:7
      %13 = typeof(%12) node_offset:5:7
      %14 = decl_val("foo") token_offset:5:10
      %15 = call(.auto, %14, []) node_offset:5:13
      %16 = decl_val("bar") token_offset:6:9
      %17 = call(.auto, %16, []) node_offset:6:12
      %18 = add(%15, %17) node_offset:5:16
      %19 = add(%12, %18) node_offset:5:7
      %20 = store(%7, %19)
      %21 = ret_tok(@Ref.void_value) token_offset:7:1
```

The two `call` instructions do have a source node, but only the index is communicated; not line/column info for use by debug info.

We need to augment the call ZIR instruction with the same information as provided by `dbg_stmt` and then treat each function call as an implicit dbg_stmt instruction.

### LINKED ISSUE #11160: make else if conditions contain debug stmt info
Zig version: `0.10.0-dev.1311+5ea94e771`

```zig
const std = @import("std");

pub fn main() void {
    var x: i32 = 1234;
    if (x + 1 < 10) {
        x += 1;
    } else if (x / 2 == 0) {
        x += 2;
    }
    var y = x;
    _ = y;
}
```

Here you can see in a debugging session, the debugger regrettably steps over the `x / 2 == 0` expression:

```
$ ./stage2/bin/zig build-exe test3.zig -fLLVM
$ gdb ./test3
(gdb) break test3.main
Breakpoint 1 at 0x201b24: file test3.zig, line 4.
(gdb) run
Starting program: /home/andy/dev/zig/build-release/test3 

Breakpoint 1, test3.main () at test3.zig:4
4	    var x: i32 = 1234;
(gdb) step
5	    if (x + 1 < 10) {
(gdb) 
10	    var y = x;
(gdb) 
```

This is because we do not emit a `dbg_stmt` for it in the ZIR:

```
        %17 = condbr(%16, {
          %19 = dbg_stmt(4, 9)
          %20 = load(%7) node_offset:6:11
          %21 = typeof(%20) node_offset:6:11
          %22 = add(%20, @Ref.one) node_offset:6:11
          %23 = store(%7, %22)
          %39 = break(%18, @Ref.void_value)
        }, {
          %30 = block({
            %24 = load(%7) node_offset:7:16
            %25 = int(2)
            %26 = div(%24, %25) node_offset:7:18
            %27 = cmp_eq(%26, @Ref.zero) node_offset:7:22
            %28 = as_node(@Ref.bool_type, %27) node_offset:7:22
            %29 = condbr(%28, {
              %31 = dbg_stmt(6, 9)
              %32 = load(%7) node_offset:8:11
              %33 = typeof(%32) node_offset:8:11
              %34 = int(2)
              %35 = add(%32, %34) node_offset:8:11
              %36 = store(%7, %35)
              %37 = break(%30, @Ref.void_value)
            }, {
              %38 = break(%30, @Ref.void_value)
            }) node_offset:7:12
          }) node_offset:7:12
```

If you ctrl+f to highlight `dbg_stmt` you can see that one is missing from the `else if` branch, just before the condition.

Vexu:
Add `dbg_inline_{begin,end}` emitted around an inlined call so that inlined `dbg_line` instructions will emit debug info pointing to the correct source function.
```zig
pub fn main() void {
    var a: u32 = 0;
    a += 
    // dbg_inline_begin foo
        @import("b.zig").foo();
    // dbg_inline_end main
    if (a != 15) @breakpoint();
}
// b.zig
pub inline fn foo() u32 {
    var c: u32 = 3;
    return c * (c + 2);
}
```

I'm not sure if this works currently... I managed to break my system to a point where I can't build master branch stage2  with a clean cache without an assertion failure.

Closes  #11158
Closes  #11160


andrewrk:
> rename dbg_stmt to dbg_line

In DWARF, there is a flag `is_stmt` that goes into the line/column info that marks whether a location is a statement. It looks like LLVM does not expose this, but I think the other backends probably care about it.

Should we make it `dbg_line` instead of `dbg_stmt`? Maybe. I'm not against it. But I do think that having `dbg_stmt` correspond to setting `is_stmt=true` in DWARF is a reasonable way to go. I get that the condition of an if is not really a statement according to Zig but maybe making it a statement according to debug info is the way that makes it work well? :shrug:

Vexu:
Got this to work. Not sure what that syntax error is about? Also should symbols from the parent function be visible in the inlined function and vice-versa? Currently they aren't but it should be easy to change by removing the lexical scopes.
```
(gdb) b a.main
Breakpoint 1 at 0x201b54: file a.zig, line 15.
(gdb) r
Starting program: /home/vexu/Documents/zig/zig/a 

Breakpoint 1, a.main () at a.zig:15
15          var a: u32 = 0;
(gdb) s
16          a +=
(gdb) p a
$1 = 0
(gdb) p c
No symbol "c" in current context.
(gdb) s
18              @import("b.zig").foo();
(gdb) s
b.foo () at b.zig:2
2           var c: u32 = 3;
(gdb) p a
A syntax error in expression, near `'.
(gdb) p c
$2 = 0
(gdb) s
3           return c * (c + 2);
(gdb) 
a.main () at a.zig:20
20          if (a != 15) @breakpoint();
(gdb) p a
$3 = 15
(gdb) p c
No symbol "c" in current context.
```

ifreund:
> Also should symbols from the parent function be visible in the inlined function and vice-versa? Currently they aren't but it should be easy to change by removing the lexical scopes.

I think making symbols from the parent visible would be confusing. Also there might be symbol conflicts right? In your example code it would be perfectly valid to rename `c` to `a`.

Vexu:
It seems stage1 and 2 behave the same on this, I think I just got confused by the different behavior when the variable is not in scope:
```
// stage1
(gdb) p c
$2 = c
// stage2
(gdb) p c
No symbol "c" in current context.
```

Seems like we should also add `dbg_block_{begin,end}` to be able match stage1:
```zig
pub fn main() void {
    var a: u32 = 0;
    {
        var b: u32 = a;
        if (b != 0) unreachable;
    }
    if (a != 0) unreachable;
}
```
```
// stage1
29          if (a != 0) unreachable;
(gdb) p a
$1 = 0
(gdb) p b
No symbol "b" in current context.
// stage2 
29          if (a != 0) unreachable;
(gdb) p a
$1 = 0
(gdb) p b
$2 = 0
```

PR #11184, opened 2022-03-15
Optional TypeScript for Autodocs development

ABuffSeagull:
Okay, so this *should* be a minimal working setup to check the Autodocs javascript file for type errors.
The `analysis.d.ts` file is for declaring what the structure of the eventual `zigAnalysis` object will be, and currently needs to be hand edited and checked for consistency with the actual `DocData` struct. I did preemptively include a type comment for the generated file, although I'm not sure if that will be the best way to ensure consistency.
In order to install and use TypeScript, just go to the `lib/std/special/docs` folder and run
```sh
npm install
```
then
```sh
npm exec tsc
```
in order to get it to type check the `main.js` file.
If you want it in watch mode:
```sh
npm exec tsc -- --watch
```
Also fixed the problem causing strict mode to error.
Also also note that there has been no fixes for the `main.js` just yet, so expect hundreds of errors.

### DIFF
--- a/lib/std/special/docs/main.js
+++ b/lib/std/special/docs/main.js
@@ -1,5 +1,184 @@
 'use strict';
 
+/**
+ * @typedef {
+    | "Type"
+    | "Void"
+    | "Bool"
+    | "NoReturn"
+    | "Int"
+    | "Float"
+    | "Pointer"
+    | "Array"
+    | "Struct"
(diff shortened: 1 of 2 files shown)


ABuffSeagull:
oh, I should probably ping @kristoff-it

kristoff-it:
Thank you!!

Issue #11159, opened 2022-03-14
stage2: array-like tuple initialization gets LLVM verification panic


mitchellh:
```
test {
    const T = @TypeOf(.{ @as(i32, 0), @as(u32, 0) });
    var a: T = .{ -1234, 5678 };
    _ = a;
}
```

result:

```
Stored value type does not match pointer operand type!
  store i32 -1234, i8* getelementptr inbounds (i8, i8* inttoptr (i64 -6148914691236517206 to i8*), i64 1), align 4, !dbg !647
 i32Stored value type does not match pointer operand type!
  store i32 5678, i8* getelementptr inbounds (i8, i8* inttoptr (i64 -6148914691236517206 to i8*), i64 1), align 4, !dbg !647
 i32
thread 711227 panic: LLVM module verification failed
```

Works on stage1.

topolarity:
Based on inspecting the AIR, looks like the bug here is that either Sema needs to avoid generating a store to a comptime field, or that the LLVM backend needs to not generate a store when the target has no runtime bits.

mitchellh:
Note, as @topolarity pointed out, this doesn't actually _work_ on stage1, but it has no error.

In stage1, both `a[0]` and `a[1]` are set to 0. 

However, an LLVM module verification panic is probably not the desired outcome in stage2 regardless of correct behavior. :)

topolarity:
That's me :-) 

This does actually work on stage1, but the test case above just happens to overlap with #11162. Here's one that doesn't:
```zig
test {
    const T = @TypeOf(.{ @as(i32, -1234), @as(u32, 5678) });
    var a: T = .{ -1234, 5678 };
    _ = a;
}
```
This works correctly on stage1 but panics on stage2, because of an AIR store to a comptime field:
```zig
  %16 = constant(i32, -1234)
...
  %10 = alloc(*tuple{comptime i32 = -1234, comptime u32 = 5678})
  %13 = struct_field_ptr_index_0(*i32, %10)
  %17!= store(%13!, %16!)
```

I think the bug is that Sema should not be generating that store of an `i32` to a `comptime` field.

Issue #11162, opened 2022-03-14
Aggregate initializers should allow initializing comptime fields, verifying equality


topolarity:
### Zig Version

0.10.0-dev.1246+5fbae9cd6

### Steps to Reproduce

cc @ifreund and @mitchellh , who provided these test cases

```zig
test {
    const T = @TypeOf(.{ @as(i32, 0), @as(u8, 123) });
    var a: T = .{-1234, 0};
    try testing.expectEqual(@as(i32, 0), a[0]);
    try testing.expectEqual(@as(u8, 123), a[1]);
}

test {
    const T = struct {
        comptime x: i32 = 0,
        comptime y: u8 = 123,
    };
    var a: T = .{.x = -1234, .y = 0};
    try testing.expectEqual(@as(i32, 0), a.x);
    try testing.expectEqual(@as(u8, 123), a.y);
}
```

### Expected Behavior

Test should fail - the provided field does not match the value of the comptime field.

Equality should be checked for the init value of comptime fields, in both structs and arrays.

### Actual Behavior

The test passes on stage1.

It fails on stage2 due to #11159, but once that bug is fixed it will incorrectly pass.

topolarity:
The second test already fails on stage2 with "error: cannot assign to constant", but it sounds like this behavior will need to be relaxed to support initializing tuple types with comptime fields.

topolarity:
On stage1, initializers fail when they include some comptime and some runtime fields:

```zig
const std = @import("std");
const testing = std.testing;

test {
    var x: u32 = 15;
    const T = @TypeOf(.{ @as(i32, -1234), @as(u32, 5678), x });
    var a: T = .{ -1234, 5678, x + 1 };
    _ = a;
}
```

```
./test2.zig:7:19: error: cannot assign to constant
    var a: T = .{ -1234, 5678, x + 1 };
                  ^
./test2.zig:7:26: error: cannot assign to constant
    var a: T = .{ -1234, 5678, x + 1 };
                         ^
```

topolarity:
Looks like we're still missing the verify equality check for comptime tuple fields.

This test is still passing on master:
```zig
test {
    const T = @TypeOf(.{ @as(i32, 0), @as(u8, 123) });
    var a: T = .{-1234, 0};
    try testing.expectEqual(@as(i32, 0), a[0]);
    try testing.expectEqual(@as(u8, 123), a[1]);
}
```

topolarity:
[...]

Issue #11182, opened 2022-03-15
stage2: Panic when initializing Tuple with explicit type


topolarity:
### Zig Version

0.10.0-dev.1335+c64279b15

### Steps to Reproduce

```zig
test {
    const T = @TypeOf(.{ @as(i32, 0), @as(u32, 0) });
    var a = T{ 0, 0 };
    _ = a;
}
```

### Expected Behavior

Test should pass.

Test is a minor variant of #11159, but the root cause is quite different. In this case, AstGen and the AIR expect to be able to refer to the "elem_type" of the initialized array.

Tuples are initialized like arrays but their fields are heterogeneous like structs, so this fails in Sema.

### Actual Behavior

Test crashes:

```
thread 16327 panic: reached unreachable code
Analyzing testme.zig: testme.zig:test_0
      %2 = dbg_stmt(2, 5)
      %3 = typeof_builtin({
        %4 = as_node(@Ref.i32_type, @Ref.zero) node_offset:2:35
        %5 = as_node(@Ref.u32_type, @Ref.zero) node_offset:2:48
        %6 = array_init_anon(.{%4, %5}) node_offset:2:24
        %7 = break_inline(%3, %6)
      }) node_offset:2:15
      %8 = dbg_var_val(%3, "T")
      %9 = dbg_stmt(3, 5)
      %10 = alloc_inferred_mut() node_offset:3:5
      %11 = validate_array_init_ty(%3) node_offset:3:14
    > %12 = elem_type(%3) node_offset:3:13
      %13 = coerce_result_ptr(%3, %10)
      %14 = elem_ptr_imm(%13, 0) node_offset:3:16
      %15 = store_node(%14, @Ref.zero) node_offset:3:16
      %16 = elem_ptr_imm(%13, 1) node_offset:3:19
      %17 = store_node(%16, @Ref.zero) node_offset:3:19
      %18 = validate_array_init({
        %14 = elem_ptr_imm(%13, 0) node_offset:3:16
        %16 = elem_ptr_imm(%13, 1) node_offset:3:19
      }) node_offset:3:14
      %19 = resolve_inferred_alloc(%10) node_offset:3:5
      %20 = dbg_var_ptr(%10, "a")
      %21 = dbg_stmt(4, 5)
      %22 = load(%10) node_offset:4:9
      %23 = ensure_result_non_error(%22) node_offset:4:9
      %24 = ret_tok(@Ref.void_value) token_offset:5:1
    For full context, use the command
[...]

PR #11185, opened 2022-03-15
stage2: resolve panic on array-like tuple initialization

topolarity:
This resolves https://github.com/ziglang/zig/issues/11159

Assorted fixes here:
  1. Comptime fields should not affect whether a type has runtime bits
  2. We were not correctly deleting the field stores after recognizing that an array initializer was a comptime-known value.
  3. LLVM was not checking that the stored type had no runtime bits, and so would generate an invalid store.

This also adds several test cases for related bugs, just to check these in for later work. Initializing tuples with mixed comptime/runtime fields is unfortunately still broken (on both stage1 and stage2) - that will have to be tackled later, probably with https://github.com/ziglang/zig/issues/11162

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -20860,6 +20860,7 @@ pub fn typeHasOnePossibleValue(
             const resolved_ty = try sema.resolveTypeFields(block, src, ty);
             const s = resolved_ty.castTag(.@"struct").?.data;
             for (s.fields.values()) |value| {
+                if (value.is_comptime) continue;
                 if ((try sema.typeHasOnePossibleValue(block, src, value.ty)) == null) {
                     return null;
                 }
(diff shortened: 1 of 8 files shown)


src/Sema.zig
```diff
@@ -3218,6 +3213,12 @@ fn zirValidateArrayInit(
             break :inst block.instructions.items[block_index + 1];
         };
 
+        // Array has one possible value, so value is always comptime-known
```
topolarity, line 3216: This usage of OPV seems like it might be incorrect for tuples, since they can have OPV without that being shared by all their elements.

I kept it all the same for now, since this change is just to make sure that the calculation of `first_block_index` is not skipped.

I think we can tackle the array/tuple init differences in another PR, along with #11182

PR #11186, opened 2022-03-15
stage2: move duplicate error set check to AstGen

mitchellh:


src/AstGen.zig
```diff
@@ -4714,14 +4714,36 @@ fn errorSetDecl(gz: *GenZir, rl: ResultLoc, node: Ast.Node.Index) InnerError!Zir
     const payload_index = try reserveExtra(astgen, @typeInfo(Zir.Inst.ErrorSetDecl).Struct.fields.len);
     var fields_len: usize = 0;
     {
+        var idents: std.AutoHashMapUnmanaged(u32, Ast.TokenIndex) = .{};
+        defer idents.deinit(gpa);
```
mitchellh, line 4718: Don't know if there some base capacity I should set here, but we can't know the capacity for sure until after the loop.

mitchellh:
I can reproduce the build failure locally but I'm not sure what causes that. Any pointers so I can take a look?

andrewrk:
```
=========== Expected compile error: ============
tmp.zig:3:5: error: duplicate error: 'Bar'
================= Full output: =================
./zig-cache/tmp/tCA9n4N5IHa76hgN/tmp.zig:3:5: error: duplicate error set field 'Bar'
./zig-cache/tmp/tCA9n4N5IHa76hgN/tmp.zig:2:5: note: previous declaration here

test 'duplicate error value in error set' failed: TestFailed
```

What's happening here is that we actually run stage2 AstGen unconditionally for all files even when building with stage1. So the mismatch between stage1 compile error message and stage2 compile error message is causing the problem here. This is a test that was previously checking that stage1 emitted the correct compile error, and now it is testing stage2 AstGen instead. Great!

Just change either the expected compile error in the test case, or the compile error message that AstGen is emitting. I'll leave it up to you which message you think is better.

mitchellh:
> Just change either the expected compile error in the test case, or the compile error message that AstGen is emitting. I'll leave it up to you which message you think is better.

I changed the test, because the error output I chose more closely matches other similar errors output out of stage2!

andrewrk (approved):

PR #11187, opened 2022-03-15
CBE: better handling of sentineled slices/arrays

jmc-88:
Far from perfect I'm sure, but enables printing of test stats which is useful, and makes a few new tests pass.

As a bonus, while doing a pass on existing tests that were being skipped for the C backend I found more that are now passing, not necessarily due to my changes here..

```
$ build/zig build --zig-lib-dir lib --prefix stage2 -Denable-llvm -Dconfig_h=build/config.h && stage2/bin/zig test -I test/ -lc -ofmt=c test/behavior.zig
694 passed; 339 skipped; 0 failed.
```

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -616,27 +616,33 @@ pub const DeclGen = struct {
             .Array => {
                 // First try specific tag representations for more efficiency.
                 switch (val.tag()) {
-                    .undef, .empty_struct_value, .empty_array => try writer.writeAll("{}"),
+                    .undef, .empty_struct_value, .empty_array => {
+                        try writer.writeByte('{');
+                        const ai = ty.arrayInfo();
+                        if (ai.sentinel) |s| {
+                            try dg.renderValue(writer, ai.elem_type, s);
+                        }
+                        try writer.writeByte('}');
+                    },
                     else => {
                         // Fall back to generic implementation.
                         var arena = std.heap.ArenaAllocator.init(dg.module.gpa);
                         defer arena.deinit();
                         const arena_allocator = arena.allocator();
 
-                        try writer.writeAll("{");
+                        try writer.writeByte('{');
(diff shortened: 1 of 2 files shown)


jmc-88:
Moved this to a draft for now, as there's some inexplicable failures when running tests without `-lc` (which Drone CI is also hitting).

jmc-88:
For my own future reference: using `-lc` in the test command will alias `std.os.system` to `std.c`, skipping the native system layer e.g. `std.os.linux`: https://github.com/ziglang/zig/blob/2c434cddd6624cfd5c71f081b9445936d25c7703/lib/std/os.zig#L64-L65

topolarity:
No need to have the sentinel as a runtime field, since it's contained in the array, right?

I think it should be enough just to include the sentinel in the typedef names, so that we don't get conflicting redeclarations.

jmc-88:
> No need to have the sentinel as a runtime field, since it's contained in the array, right?
> 
> I think it should be enough just to include the sentinel in the typedef names, so that we don't get conflicting redeclarations.

Yes, I'm currently leaning on the same interpretation.

But, notice I converted this PR to a draft since it's not doing what it claimed it did -- it doesn't unblock printing out test stats :slightly_smiling_face: For that to work I need to have at least the `asm` directives working, and I hit some stage2 bugs with that. I might split the sentineled slices/arrays commit to its own PR and leave the rest for later...

topolarity:
> I might split the sentineled slices/arrays commit to its own PR and leave the rest for later...

Sounds like the right idea to me :+1:

P.S. Nice work on the C backend lately! It's been great to watch your changes coming in

jmc-88:
As mentioned in the above [comment](https://github.com/ziglang/zig/pull/11187#issuecomment-1069667943), I've reduced the scope of this change to just the handling of sentineled slices/arrays. I'll defer the test runner changes for later.

PR #11189, opened 2022-03-16
std: introduce posix_spawn as an alt to fork-exec

kubkon:
Currently, the new API will only be available on macOS with the intention of adding more POSIX systems to it incrementally (such as Linux, etc.). A cool new feature is the ability for the `ChildProcess` have the ASLR disabled on request:

```zig
var child = ChildProcess.init(argv.items, arena);
child.disable_aslr = true;
try child.spawn();
return child.pid;
```

Changes:
* add `posix_spawn` wrappers in a separate container in `os/posix_spawn.zig`
* rewrite `ChildProcess.spawnPosix` using `posix_spawn` targeting macOS as `ChildProcess.spawnMacos`
* introduce a `posix_spawn` specific `std.c.waitpid` wrapper which does return an error in case the child process failed to exec - this is required for any process that was spawned using `posix_spawn` mechanism as, by definition, the errors returned by `posix_spawn` routine cover only the `fork`-equivalent; `pre-exec()` and `exec()` steps are covered by a catch-all error `ECHILD` returned by `waitpid` on unsuccessful execution, e.g., no such file error, etc.

### DIFF
--- a/lib/std/os/posix_spawn.zig
+++ b/lib/std/os/posix_spawn.zig
@@ -1,0 +1,282 @@
+const std = @import("std");
+const builtin = @import("builtin");
+
+const os = @import("../os.zig");
+const system = os.system;
+const errno = system.getErrno;
+const fd_t = system.fd_t;
+const mode_t = system.mode_t;
+const pid_t = system.pid_t;
+const unexpectedErrno = os.unexpectedErrno;
+const UnexpectedError = os.UnexpectedError;
+const toPosixPath = os.toPosixPath;
+const WaitPidResult = os.WaitPidResult;
+
+pub usingnamespace posix_spawn;
+
+pub const Error = error{
+    SystemResources,
+    InvalidFileDescriptor,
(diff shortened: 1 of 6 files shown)


andrewrk (approved):

lib/std/os/posix_spawn.zig
```diff
@@ -0,0 +1,282 @@
...
+const errno = system.getErrno;
+const fd_t = system.fd_t;
+const mode_t = system.mode_t;
+const pid_t = system.pid_t;
+const unexpectedErrno = os.unexpectedErrno;
+const UnexpectedError = os.UnexpectedError;
+const toPosixPath = os.toPosixPath;
+const WaitPidResult = os.WaitPidResult;
+
+pub usingnamespace posix_spawn;
```
andrewrk, line 15: I would love to see this gotten rid of somehow, but I won't request to block the PR merge for it :-)

PR #11191, opened 2022-03-16
Zig gdb improvements

Snektron:
This pull request adds some nice pretty printers for GDB. In particular:
* A pretty printer for `ArrayList(Unmanaged)`:
  ```
  $1 = ArrayList(i32) of length 2, capacity 1024 = {123, 1234}
  ```
* A pretty printer for `ArrayHashMap(Unmanaged)`:
  ```
  $1 = AutoArrayHashMapUnmanaged(i32,i32) of length 2 = {[123] = 1, [1234] = 1}
  ```
* A pretty printer for `MultiArrayList`:
  ```
  $1 = MultiArrayList(X) of length 2, capacity 8 = {[a] = {"zero", "one"}, [b] = {1, 2}}
  ```
* A pretty printer for `HashMap(Unmanaged)`:
  ```
  $1 = AutoHashMapUnmanaged(i32,[]const u8) of length 2, capacity 8 = {[1] = "python", [2] = "boa"}
  ```
* A pretty printer for `Type` (from selfhosted):
  ```
  (gdb) p param_types
  $1 = {{tag = .single_mut_pointer, payload = .i32}, {tag = .array, payload = {len = 3, elem_type = .i32}}}
  (gdb) set print pretty on
  (gdb) p param_types
  $2 = {{
      tag = .single_mut_pointer,
      payload = .i32
    }, {
      tag = .array,
      payload = {
        len = 3,
        elem_type = .i32
      }
    }}
   ```
* A pretty printer for `Value` (from selfhosted):
  ```
  $1 = {tag = .int_u64, payload = 123}
  ```
  
Note that the pretty printers for `Type` and `Value` must be kept up-to-date with the respective `Tag` enum. For some other types, like MultiArrayLists and HashMaps, gdb needs to be able to access some internal/child type. It looks like Zig types aren't exported in the same way that GDB expects them, so `gdb.lookup_type` wasn't possible here. `gdb.lookup_symbol` seems to work fine for functions of generic types though, so i've added a helper method (only exported in debug mode) which is traversed by gdb to find the type.


Snektron:
One drawback of this is that some types are not structured as advertised. For example, one cannot access the `tag` of a selfhosted `Type`. I think GDB offers ways around this using Xmethods, but those are for another time.

PR #11128, opened 2022-03-11
stage2: Track parent type for `.elem_ptr`, `.field_ptr`, and `.*_payload_ptr`

topolarity:
This PR resolves the "tricky" situation noted in #11043. Also resolves #11139.

Parent pointer values (`.elem_ptr`, `.field_ptr`, `.eu_payload_ptr`, and `.union_payload_ptr`) don't have any way of recovering the type that they were constructed with. If you create a .field_ptr from the result of a bit-cast (e.g. `&@bitCast(T, ptr).foo`), it can't be lowered or dereferenced correctly, since we cannot recover `T`

This ended up being a pretty scruffy yak. Major changes:
   - Add `container_ty`/`elem_ty` to parent pointer Values
   - Implement `Type.hasWellDefinedLayout` and `Sema.typeHasWellDefinedLayout`, so that we can anticipate when bitcasting in Sema is legal
   - Update Sema `pointerDeref` to bitcast if parent pointer type diverges from the backing decl type
   - Update LLVM `lowerParentPtr` to bitcast if parent pointer type diverges from the backing decl type
   
I haven't gotten around to updating `beginComptimePtrMutation` - its behavior is unchanged since it doesn't use the tracked container type information yet.

PR #11193, opened 2022-03-16
stage2: elem_ptr needs to know if slice or direct access

mitchellh:
This fixes one of the major issues plaguing the `std.sort` comptime tests.
The high level issue is that at comptime, we need to know whether `elem_ptr` is
being used to subslice an array-like pointer or access a child value. High-level
example:

    var x: [2][2]i32 = undefined;
    var a = &x[0]; // elem_ptr, type *[2]i32

    var y: [5]i32 = undefined;
    var b = y[1..3]; // elem_ptr, type *[2]i32

`a` is pointing directly to the 0th element of `x`. But `b` is
subslicing the 1st and 2nd element of `y`. At runtime with a well
defined memory layout, this is an inconsequential detail. At comptime,
the values aren't laid out exactly in-memory so we need to know the
difference.

This becomes an issue specifically in this case:

    var c: []i32 = a;
    var d: []i32 = b;

When converting the `*[N]T` to `[]T` we need to know what array to point
to. For runtime, its all the same. For comptime, we need to know if its
the parent array or the child value.

See the behavior tests for more details.

This commit fixes this by adding a boolean to track this on the
`elem_ptr`. We can't just immediately deref the child for `&x[0]`
because it is legal to ptrCast it to a many-pointer, do arithmetic, and
then cast it back (see behavior test) so we need to retain access to the
"parent" indexable.

I don't love the nomenclature of "direct" here, happy to change that if
there are better ideas around what to call it.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -10388,25 +10388,32 @@ fn zirTypeInfo(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Ai
                 )).?;
                 try sema.mod.declareDeclDependency(sema.owner_decl, fn_info_decl);
                 try sema.ensureDeclAnalyzed(fn_info_decl);
+                var fn_ty_buffer: Value.ToTypeBuffer = undefined;
+                const fn_ty = fn_info_decl.val.toType(&fn_ty_buffer);
                 const param_info_decl = (try sema.namespaceLookup(
                     block,
                     src,
-                    fn_info_decl.val.castTag(.ty).?.data.getNamespace().?,
+                    fn_ty.getNamespace().?,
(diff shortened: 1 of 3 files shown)


topolarity:
Great find.

I had to work around something similar in #11128 - In that case I used the type of the elem_ptr to infer whether the load should "over-read" at comptime (to adjacent elements in the parent array).

This feels much more correct. Maybe a "direct elem_ptr" is worth calling an "offset_ptr" or similar

andrewrk:
Whew, that was a tricky one. I ended up solving it a different way, which I suspect you will be happy with. Thanks for the test cases!

PR #11195, opened 2022-03-16
stage2: x86_64: implement `@floatToInt` for `f32` and `f64`

mparadinha:
stage1 and LLVM backend both use the `vcvt` group of instructions for these operations. Those require AVX512, so I opted `fisttp`/`fld` instead.

the `@floatToInt` test from `test/behavior/cast.zig` now passes for x86_64 backend.

### DIFF
--- a/src/arch/x86_64/CodeGen.zig
+++ b/src/arch/x86_64/CodeGen.zig
@@ -5437,11 +5437,56 @@ fn airIntToFloat(self: *Self, inst: Air.Inst.Index) !void {
 
 fn airFloatToInt(self: *Self, inst: Air.Inst.Index) !void {
     const ty_op = self.air.instructions.items(.data)[inst].ty_op;
-    const result: MCValue = if (self.liveness.isUnused(inst))
-        .dead
-    else
-        return self.fail("TODO implement airFloatToInt for {}", .{self.target.cpu.arch});
-    return self.finishAir(inst, result, .{ ty_op.operand, .none, .none });
+    if (self.liveness.isUnused(inst))
+        return self.finishAir(inst, .dead, .{ ty_op.operand, .none, .none });
+
+    const src_ty = self.air.typeOf(ty_op.operand);
+    const dst_ty = self.air.typeOfIndex(inst);
+    const operand = try self.resolveInst(ty_op.operand);
+
+    // move float src to ST(0)
+    const stack_offset = switch (operand) {
+        .stack_offset, .ptr_stack_offset => |offset| offset,
+        else => blk: {
+            const offset = @intCast(i32, try self.allocMem(
+                inst,
+                @intCast(u32, src_ty.abiSize(self.target.*)),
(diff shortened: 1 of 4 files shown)


andrewrk:
Note that the [target value](https://github.com/ziglang/zig/blob/312536540baf26728a56304811f63f01a7414b7a/src/arch/x86_64/CodeGen.zig#L49) has a complete set of [CPU feature flags](https://github.com/ziglang/zig/blob/312536540baf26728a56304811f63f01a7414b7a/lib/std/target/x86.zig#L7-L150) which tells whether AVX512 is available.

doesthisusername:
I haven't been keeping up with this project for a while, so I might be missing something, but x86 has had SIMD float conversion instructions for ages (depending on your definition). While AVX512 does introduce some more `vcvt*` instructions, those appear to work like the old SSE/AVX ones, but with the AVX512 write-mask and increased operand width.
I don't actually know if those x87 instructions are worse than the SSE/AVX ones (`[v]cvtss2si`, `[v]cvtsd2si`, etc.), but both should be supported by virtually every x86_64 CPU made in 2011 or later, with x87 being older.

In any case, I just want to make it clear that AVX512 isn't required for a SIMD/not-40-year-old `@floatToInt` implementation in the event you missed that (which is fair, I got confused as well seeing `cvt*` instructions on ISA listings when double checking this, yet seeing `vcvt*` in disassembly, but it turns out that that `v` prefix just means it's using plain AVX instead of SSE ... or so I think).

2011 might still be too new, I can't make that call, though other compilers seem to use it, and honestly I'm a bit confused in terms of SSE and AVX and whether a CPU with SSE (basically anything in the 2000s) can also run these instructions, or if you need AVX or something.

Here's a disassembly example on [godbolt](https://godbolt.org/z/WG8Kxa9ef).

LemonBoy:
> both should be supported by virtually every x86_64 CPU made in 2011 or later, with x87 being older.

X86_64 by definition _always_ supports at least SSE2 and even on older X86 CPUs using the x87 FPU is going to give you wildly different results (that's also why I bumped the LLVM CPU model for the x86 target to `pentium4`) and make the test suite fail.

mparadinha:
Thank you all for this information! :smile:
@doesthisusername Checking the instruction reference I now understand what you mean. I did in fact get confused with the `v` prefix, and though that meant AVX512 was required for this conversion (which in hindsight doesn't seem very reasonable hahaha)

> and even on older X86 CPUs using the x87 FPU is going to give you wildly different results
@LemonBoy I didn't expect the x87 FPU to give different results, is this somehow related to the FPU registers having 80 bits?

So, I think I should probably wait until the `xmm` family of registers gets introduced into the x86_64 backend and then reimplement `@floatToInt` using the SSE2 instructions.

andrewrk:
Here's the [godbolt link with sse2 disabled](https://godbolt.org/z/r5qfxs13P)

PR #11196, opened 2022-03-16
Implement `@mulAdd` for vectors

schmee:
Little bit unsure of the LLVM part for the`libc` strat for vectors, but it seems to work.

Using `expectEquals` on vectors runs into the following error:

```
$ zig test test/behavior/muladd.zig -Itest -fLLVM
/Users/<USER>/repos/zig/stage2/lib/zig/std/fmt.zig:765:18: error: unreachable else prong; all cases already handled
            else => |e| return e,
```

Not sure if the else-prong actually is unreachable or if there is a bug, but I'll take a look and see what I find.


andrewrk:
Nice work!

PR #11198, opened 2022-03-17
stage2: bit_not on u0 is always 0

mitchellh:
A simple one, exercised by some stdlib tests (bit sets). 

Might be more useful to fix `std.math.big.int` to not crash in this case, but this is probably something that should avoid all the BigInt machinery anyways (as the TODO below my changes implies...)


mitchellh:
Context...

CleanShot 2022-03-16 at 21 57 33@2x

PR #11200, opened 2022-03-17
stage2: wasm - Implement memcpy instruction

Luukdegram:
This implements the `memcpy` instruction, as well as improves the inline memcpy function.
We feature-detect for the availability of wasm's built-in instruction `memory.copy`. When the feature 'bulk-memory' is not enabled, we perform the operation ourselves.
When the length is comptime-known, we perform an inline loop, whereas we emit a regular runtime loop into the binary when the length is runtime known.

The same changes were made to the inline memset function and instructions such as `airWrapErrUnionErr` now use this to set the entire payload to 'undefined'.


Luukdegram:
Seems the CI error is due to a memory leak in Sema caused by the test in https://github.com/ziglang/zig/blob/master/test/behavior/bugs/7047.zig
Honestly a little surprised this isn't being caught by any other CI runs as I can reproduce this from any commit since f9e4344b locally.

The memory leak is caused by the parameter names of generic functions not being freed. I'll implement a fix later today after work.

andrewrk:
Looks great!

PR #11203, opened 2022-03-17
stage2: add safety checks for index out of bounds

wsengir:
Sorry for the noisy diff, but it was really weird that each branch on `elem{Ptr,Val}` had a different argument order, so I normalized them. I also used the term `indexable` instead of `array` when it wasn't specifically referring to a `[N]` type.

Broke out indexing into empty errors since they could be checked with less resolved information. Also moved comptime index bound checks for arrays to run them before bubbling up `undefined`.

Added ` +1 (sentinel)` to the length printout for sentinel-terminated arrays/slices since it was really weird & inconsistent that `index X outside array of length X` would only sometimes be an error.

Made `elemPtr{Array,Slice}` allow an `undefined` array/slice since it's not actually dereffing any pointers at this point in time. Instead, any attempt to use returned `undefined` pointer will be responsible for catching the `.undef`.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -3201,7 +3201,7 @@ fn zirValidateArrayInit(
         // any ZIR instructions at comptime; we need to do that here.
         if (array_ty.sentinel()) |sentinel_val| {
             const array_len_ref = try sema.addIntUnsigned(Type.usize, array_len);
-            const sentinel_ptr = try sema.elemPtrArray(block, init_src, array_ptr, array_len_ref, init_src);
+            const sentinel_ptr = try sema.elemPtrArray(block, init_src, array_ptr, init_src, array_len_ref);
             const sentinel = try sema.addConstant(array_ty.childType(), sentinel_val);
             try sema.storePtr2(block, init_src, sentinel_ptr, init_src, sentinel, init_src, .store);
         }
@@ -15600,6 +15600,7 @@ pub const PanicId = enum {
     cast_to_null,
     incorrect_alignment,
     invalid_error_code,
+    index_out_of_bounds,
 };
 
 fn addSafetyCheck(
@@ -15721,6 +15722,7 @@ fn safetyPanic(
         .cast_to_null => "cast causes pointer to be null",
         .incorrect_alignment => "incorrect alignment",
         .invalid_error_code => "invalid error code",
+        .index_out_of_bounds => "attempt to index out of bounds",
     };
 
     const msg_inst = msg_inst: {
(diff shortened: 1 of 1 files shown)


src/Sema.zig
```diff
@@ -16894,57 +16868,221 @@ fn tupleField(
...
+    }
+
+    const maybe_undef_array_ptr_val = try sema.resolveMaybeUndefVal(block, array_ptr_src, array_ptr);
+    // index must be defined since it can index out of bounds
+    const maybe_index_val = try sema.resolveDefinedValue(block, elem_index_src, elem_index);
+
+    if (maybe_index_val) |index_val| {
+        const index = @intCast(usize, index_val.toUnsignedInt());
+        if (index >= array_len_s) {
+            const sentinel_label: []const u8 = if (array_sent) ":+1" else "";
```
Vexu, line 16949:
```suggestion
            const sentinel_label: []const u8 = if (array_sent) " +1 (sentinel)" else "";
```
`:+1` by itself seems pretty confusing, maybe adding (sentinel) would make it clearer?

src/Sema.zig
```diff
@@ -16727,90 +16717,66 @@ fn elemVal(
...
 ) CompileError!Air.Inst.Ref {
-    const array_src = src; // TODO better source location
-    const array_ty = sema.typeOf(array);
+    const indexable_src = src; // TODO better source location
+    const indexable_ty = sema.typeOf(indexable);
 
-    if (!array_ty.isIndexable()) {
-        return sema.fail(block, src, "array access of non-indexable type '{}'", .{array_ty});
+    if (!indexable_ty.isIndexable()) {
+        return sema.fail(block, src, "indexing of non-indexable type '{}'", .{indexable_ty});
```
Vexu, line 16728: I'd prefer to keep calling this `array access` the same way `{ 1, 2 }` is called `array initialization syntax` even though it works for vectors and tuples too, and `.{ .foo = 1 }` is called `struct initialization syntax` when it can also be used for unions.

src/Sema.zig
wsengir: Sounds great to me. I was originally trying to make it symmetrical with sentinel type syntax, but I think in the case of an error message, we should be as clear as possible.

src/Sema.zig
wsengir: Looking at the docs, we seem to use a mix of "indexing" and "array indexing". It also refers to "array initialization syntax" instead as "array literals" and "list literals". I don't mind keeping it as "array access", but I also would like to stay consistent where possible.

andrewrk:
Great work :+1:

PR #11208, opened 2022-03-17
stage2: Fix memory leak in Sema

Luukdegram:
When a generic call evaluates to a generic type, the call will be re-generated. However, the old function was not freed before being re-generated, causing a memory leak. So rather than only returning an error, we first free the old value.


andrewrk:
Wow, clean fix!

PR #11212, opened 2022-03-18
std: change error union hashing to use if/else format

mitchellh:
Test fails on stage2 without this change, but logically it should
be equivalent. I wasn't able to reduce this to an isolated case
demonstrating any issues with the compiler. I'm going to keep seeking
that out, but in the mean time I felt it was probably worthwhile for the
auto_hash tests to pass since we use auto hash significantly in the
compiler so this should help it compile itself.

### DIFF
--- a/src/type.zig
+++ b/src/type.zig
@@ -1788,8 +1788,14 @@ pub const Type = extern union {
                     continue;
                 },
                 .error_set => {
-                    const error_set = ty.castTag(.error_set).?.data;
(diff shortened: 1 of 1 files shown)


mitchellh:
Root issue: #11213

mitchellh:
#11218 fixes the root issue of this PR, but this PR also contains improved error formatting for `error_set` values based on the new error equality work, so I recommend still merging this if its acceptable.

andrewrk:
The improved error formatting for `error_set` values looks nice, but can we drop the unnecessary changes to the std lib?

PR #11214, opened 2022-03-18
std: explicitly handle error.UnexpectedExitCode in build_runner

### LINKED ISSUE #10381: zig build run produces error trace when exit code is non-zero
### Zig Version

0.9.0

### Steps to Reproduce

1. Use std.build based project
2. Have a run-step
3. Have the program return non-zero exit, e.g. 1

### Expected Behavior

Error message that the program exited with code X. This was the case in 0.8.1.

### Actual Behavior

Long error trace:

```
$ zig build run
...
The following command exited with error code 1 (expected 0):
cd ... && .../zig-out/bin/...
error: UnexpectedExitCode
/usr/lib/zig/std/os.zig:2417:19: 0x2aa37b in std.os.mkdiratZ (build)
        .EXIST => return error.PathAlreadyExists,
                  ^
/usr/lib/zig/std/os.zig:2379:9: 0x2aa263 in std.os.mkdirat (build)
        return mkdiratZ(dir_fd, &sub_dir_path_c, mode);
        ^
/usr/lib/zig/std/fs.zig:1236:9: 0x2aa156 in std.fs.Dir.makeDir (build)
        try os.mkdirat(self.fd, sub_path, default_new_dir_mode);
        ^
/usr/lib/zig/std/build/RunStep.zig:238:17: 0x2cfd10 in std.build.RunStep.make (build)
                return error.UnexpectedExitCode;
                ^
/usr/lib/zig/std/build.zig:3192:9: 0x2ab58e in std.build.Step.make (build)
        try self.makeFn(self);
        ^
/usr/lib/zig/std/build.zig:502:9: 0x2a9c2c in std.build.Builder.makeOneStep (build)
        try s.make();
        ^
/usr/lib/zig/std/build.zig:496:17: 0x2a9bd8 in std.build.Builder.makeOneStep (build)
                return err;
                ^
/usr/lib/zig/std/build.zig:457:13: 0x29ae41 in std.build.Builder.make (build)
            try self.makeOneStep(s);
            ^
/usr/lib/zig/std/special/build_runner.zig:213:21: 0x28f434 in main (build)
            else => return err,
                    ^
```

(also, is it just me or does the error trace above RunStep.zig:238:17 seem nonsensical? It almost looks like there is an older error trace of a internally handled error which was not cleared correctly)

iddev5:
```
RunStep on unexpected exit code used to return error.UncleanExit, which
was confusing and unclear. When it was changed, the error handling code
in build_runner was not modified, which produced an error trace.

This commit explicitly handles error.UnexpectedExitCode in build_runner
so that the behavior now matches that of zig 0.8.1 after which it was
regressed.
```
I think this fix is the correct behavior since an unexpected exit code means a failure on the part of the program which did not arose from zig side, and showing an error trace is unnecessary here, though it still returns 1 indication that something somewhere went wrong.

Closes #10381


matu3ba:
Can you please attach the commit, which regressed the behavior?
Ideally there would be also a test that checks the expected return code(s) to prevent future regressions.
You could use for this from https://github.com/ziglang/zig/pull/11138/files `test "build and call child_process"` with setting the expected return code like here: https://github.com/matu3ba/chepa/blob/52ec689b65e53bf616898d56b6f2aa7324d4b8c5/build.zig#L31

Unfortunately running `zig build` from zig code is still a pain in the butt, which could be solved either with https://github.com/ziglang/zig/pull/11138 or moving stuff from `src/test.zig` into libstd.

iddev5:
The commit which regressed it was https://github.com/ziglang/zig/commit/da2b615efb14f662bf3aac1f9050174966acd17d

I am not sure how a test case will work here, because with or without this commit, the process will return 1.

matu3ba:
> Adding a test would mean running the build runner for a test application and checking if it is throwing a "zig error", unless there can be any other kind of test. Are there even any test/testing framework for build runner?

Mhm, is it unfeasible to recreate `error.UnexpectedExitCode` for each platform?
You can build the builder that builds hello world/something broken and check the return status + stderr, which must be empty instead of having the error trace.

iddev5:
After thinking for a bit, I think that's a fair solution. With it, a whole line of tests can be created for build_runner and I m interested in doing it. Though that makes this PR depending on your PR #11138 (tmp.getFullPath()). For now I will be doing that in separate which can then be rebased once 11138 is merged.

iddev5:
After doing my tests, I have concluded that it's not the right solution. The build runner in any case writes the error line that it is unexpected code to stderr along with the command executed, which is correct. It's just that stack trace is extra to it. This means that asserting stderr != 0 won't work as it will never be zero.

I am fairly confident that it's not possible to cleanly test it.

iddev5:
I managed to make a test here: https://github.com/iddev5/zig/tree/ay-build-runner-testing-fw but this depends on #11138

lib/std/testing.zig
```diff
@@ -450,6 +450,33 @@ pub fn buildExe(zigexec: []const u8, zigfile: []const u8, binfile: []const u8) !
     try expectEqual(ret_val, .{ .Exited = 0 });
 }
 
+/// Spawns a zig build runner process 'zigexec build subcmd' and
```
matu3ba, line 453: This does not mention what `cwd` and `lib_dir` are used for.
For `test` commands the structure is always the same, but for `cwd` and `lib_dir` we have significant variability or what do you think?
The user could also set the `lib_dir` itself, so this sounds not too convincing yet to me.

The main advantage to justify `runZigBuild` I see from offering `cwd`.

matu3ba (commented):
progstr could be made shorter and I dont understand the use case for `lib_dir`

I dont understand which `lib_dir` is meant (its not explained in the doc comments) and per se I would assume providing the path to libstd is not a frequent use case.
Explain.

lib/std/special/build_runner.zig
```diff
@@ -331,3 +331,57 @@ fn argsRest(args: [][]const u8, idx: usize) ?[][]const u8 {
     if (idx >= args.len) return null;
     return args[idx..];
 }
+
+const progstr =
+    \\ const std = @import("std");
```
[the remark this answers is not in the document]
iddev5: Fixed, thanks

lib/std/testing.zig
iddev5: I m not sure I understand your question correctly. Are you talking about lack of doc comments explaining what does what?

For now, cwd is used to mention the directory in which ``zig build`` will run, this is absolutely needed. lib_dir is actually used to specify the path to the zig library which will be used. This is needed because we cannot assume which zig compiler is being used to perform the test, so its better to default it to use the local stdlib present in the source directory.

matu3ba: Ah, sounds like a frequent operation to test or ensure compatibility.
Please describe both very briefly in the doc comment.

matu3ba (approved):
With the added doc comment it looks good to me.

matu3ba:
I have another use case for this:
testing custom compare function of output, in my case for testing paths on user system (line-wise string from EOL to start):

The code is roughly looking like this (not sure, if calling the build.zig functions directly without setting a pile of things is feasible):
```zig
const min_prog =
    \\const stdout = @import("std").io.getStdOut();
    \\pub fn main() {
    \\  stdout.writeAll("/bruh/ok");
    \\  return 0;
    \\}
;

// TODO simpler way to directly invoke the build.zig functions with necessary arguments?
const build_zig =
    \\  const target = b.standardTargetOptions(.{});
    \\  const mode = b.standardReleaseOptions();
    \\  const exe = b.addExecutable("chepa", "src/min_prog.zig");
    \\  exe.setTarget(target);
    \\  exe.setBuildMode(mode);
    \\  const run_cmd = exe.run();
    \\  run_cmd.step.dependOn(b.getInstallStep());
    \\  const run_cmd_step = b.step("run", "Run the app"); // runs min_prog written as zig file in test block
    \\  run_cmd_step.dependOn(&run_cmd.step);
;

test "custom user compare function" {
    // TODO set cwd to have proper zig build dir
}
```

This also makes me wonder, if path to libstd should be available in test blocks or what other things from the caller.

lib/std/testing.zig
iddev5: Hopefully clear enough now, not putting use case info into doc comments because thats not what they are for, generally.

lib/std/special/build_runner.zig
```diff
@@ -331,3 +331,56 @@ fn argsRest(args: [][]const u8, idx: usize) ?[][]const u8 {
...
+    \\ }
+;
+
+test "issue 10381" {
+    const testing = std.testing;
+    const allocator = testing.allocator;
+
+    var it = try std.process.argsWithAllocator(allocator);
+    const testargs = try testing.getTestArgs(&it);
+    defer it.deinit();
```
Vexu, line 359: This defer should be one line up, though currently it doesn't matter as this test isn't even being referenced. Maybe `std/build.zig` would be a better place for it?

iddev5: Oh no, that was an accident, the test is meant to be referenced. I have moved it to std.build and pushed the changes

lib/std/build.zig
```diff
@@ -3492,3 +3492,58 @@ test "LibExeObjStep.addPackage" {
     const dupe = exe.packages.items[0];
     try std.testing.expectEqualStrings(pkg_top.name, dupe.name);
 }
+
+test "build_runner issue 10381" {
```
andrewrk, line 3496: Please specify what is being tested rather than referencing a github issue number

lib/std/testing.zig
```diff
@@ -450,6 +450,36 @@ pub fn buildExe(zigexec: []const u8, zigfile: []const u8, binfile: []const u8) !
...
+/// expects success
+/// If specified, runs zig build in the cwd path
+/// If specified, uses the specified lib_dir for zig standard library
+/// instead of compiler's default library directory
+pub fn runZigBuild(zigexec: []const u8, options: struct {
```
andrewrk, line 458: I don't think `runZigBuild` belongs in std.testing. We already have `test/standalone/*` for this.

lib/std/testing.zig
andrewrk: @matu3ba I would like you refrain from reviewing code that is not directly related to changes that you are responsible for. Generally, your code reviews have been confusing, and thereby misleading to newcomers. Please leave nearly all code reviews to others.

andrewrk (commented):
Reverting momentarily. If you would like to continue to work on these changes, please re-open and then request me as a reviewer.

lib/std/testing.zig
matu3ba: ok. no more reviews from me then.

PR #11217, opened 2022-03-18
CBE: add support for tuples

jmc-88:
Very simplistic but it seems to get the job done, at least good enough for `test/behavior/tuple.zig` to pass. Let me know of any comments.

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -1000,6 +1000,46 @@ pub const DeclGen = struct {
         return name;
     }
 
+    fn renderTupleTypedef(dg: *DeclGen, t: Type) error{ OutOfMemory, AnalysisFail }![]const u8 {
+        const tuple = t.tupleFields();
+
+        var buffer = std.ArrayList(u8).init(dg.typedefs.allocator);
+        defer buffer.deinit();
+        const writer = buffer.writer();
+
+        try buffer.appendSlice("typedef struct {\n");
+        {
+            for (tuple.types) |field_ty, i| {
+                const val = tuple.values[i];
+                if (val.tag() != .unreachable_value) continue;
+
+                var name = std.ArrayList(u8).init(dg.gpa);
+                defer name.deinit();
+                try name.writer().print("field_{d}", .{i});
+
+                try buffer.append(' ');
+                try dg.renderTypeAndName(writer, field_ty, .{ .bytes = name.items }, .Mut, Value.initTag(.abi_align_default));
+                try buffer.appendSlice(";\n");
+            }
+        }
+        try buffer.appendSlice("} ");
+
+        const name_start = buffer.items.len;
+        try writer.print("zig_T_{};\n", .{typeToCIdentifier(t)});
+
+        const rendered = buffer.toOwnedSlice();
+        errdefer dg.typedefs.allocator.free(rendered);
+        const name = rendered[name_start .. rendered.len - 2];
+
+        try dg.typedefs.ensureUnusedCapacity(1);
+        dg.typedefs.putAssumeCapacityNoClobber(
+            try t.copy(dg.typedefs_arena),
+            .{ .name = name, .rendered = rendered },
+        );
+
+        return name;
+    }
+
     fn renderUnionTypedef(dg: *DeclGen, t: Type) error{ OutOfMemory, AnalysisFail }![]const u8 {
         const union_ty = t.cast(Type.Payload.Union).?.data;
         const fqn = try union_ty.getFullyQualifiedName(dg.typedefs.allocator);
@@ -1276,7 +1316,9 @@ pub const DeclGen = struct {
                 return w.writeAll(name);
             },
             .Struct => {
-                const name = dg.getTypedefName(t) orelse
+                const name = dg.getTypedefName(t) orelse if (t.isTuple())
+                    try dg.renderTupleTypedef(t)
+                else
                     try dg.renderStructTypedef(t);
 
                 return w.writeAll(name);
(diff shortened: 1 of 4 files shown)


test/behavior/bugs/3779.zig
```diff
@@ -7,9 +7,9 @@ const ptr_tag_name: [*:0]const u8 = tag_name;
 
 test "@tagName() returns a string literal" {
     if (builtin.zig_backend == .stage1) return error.SkipZigTest; // stage1 gets the type wrong
-    try std.testing.expectEqual(*const [13:0]u8, @TypeOf(tag_name));
-    try std.testing.expectEqualStrings("TestEnumValue", tag_name);
-    try std.testing.expectEqualStrings("TestEnumValue", ptr_tag_name[0..tag_name.len]);
+    try std.testing.expect(*const [13:0]u8 == @TypeOf(tag_name));
```
matu3ba, line 10: more of a question: whats missing for `expectEqual` and `expectEqualStrings` to work?

matu3ba (commented):
only some nits on comments and unexplained `rendered.len - 2`.

src/codegen/c.zig
```diff
@@ -1000,6 +1000,46 @@ pub const DeclGen = struct {
...
+            }
+        }
+        try buffer.appendSlice("} ");
+
+        const name_start = buffer.items.len;
+        try writer.print("zig_T_{};\n", .{typeToCIdentifier(t)});
+
+        const rendered = buffer.toOwnedSlice();
+        errdefer dg.typedefs.allocator.free(rendered);
+        const name = rendered[name_start .. rendered.len - 2];
```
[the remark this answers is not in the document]
jmc-88: To trim the trailing `";\n"` that's appended to the name above (the code records the `name_start`, appends the type name and trailing bits, then we chop them off here when slicing).

Look around `c.zig` and you'll see this is done in most of the `render*Typedef()` functions for pretty much the same reasons.

src/codegen/c.zig
```diff
@@ -1000,6 +1000,46 @@ pub const DeclGen = struct {
         return name;
     }
 
+    fn renderTupleTypedef(dg: *DeclGen, t: Type) error{ OutOfMemory, AnalysisFail }![]const u8 {
+        const tuple = t.tupleFields(); // Handle 0 bit types elsewhere.
```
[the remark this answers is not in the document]
jmc-88: Eh, apologies, this is the remnant of a copy-past of this line from `renderStructTypedef()` above:
```
        const struct_obj = t.castTag(.@"struct").?.data; // Handle 0 bit types elsewhere.
```
I don't think this is needed, I'll remove it.

test/behavior/bugs/3779.zig
jmc-88: Good question, and might be a useful note for posterity:

`std.testing.expectEqual*()` functions try to be helpful to the user by printing out, when a difference is found, what the difference was or where inside the string/slice it began. Unfortunately this means that they require `std.debug.print()` to work, which then carries with it a lot more of the stdlib than strictly necessary for equality testing. In the case of the C backend, this means that in order for it to work freestanding it requires `asm()` statements to compile, which are currently not working in stage2. I confirmed with Andrew on Discord chat that it's OK to move behavior tests to a plain `std.testing.expect()` if needed:

image


As a side note: printing _does_ already work if we link with `-lc`, which makes `std.os.system == std.c`: https://github.com/ziglang/zig/blob/2c434cddd6624cfd5c71f081b9445936d25c7703/lib/std/os.zig#L64-L65

src/codegen/c.zig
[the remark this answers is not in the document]
jmc-88: Removed -- resolving this thread.

matu3ba (approved):

src/codegen/c.zig
```diff
@@ -3149,9 +3201,18 @@ fn airStructFieldVal(f: *Function, inst: Air.Inst.Index) !CValue {
...
     const struct_byval = try f.resolveInst(extra.struct_operand);
     const struct_ty = f.air.typeOf(extra.struct_operand);
+    var buf = std.ArrayList(u8).init(f.object.dg.gpa);
+    defer buf.deinit();
     const field_name = switch (struct_ty.tag()) {
         .@"struct" => struct_ty.structFields().keys()[extra.field_index],
         .@"union", .union_tagged => struct_ty.unionFields().keys()[extra.field_index],
+        .tuple => blk: {
+            const tuple = struct_ty.tupleFields();
+            if (tuple.values[extra.field_index].tag() != .unreachable_value) return CValue.none;
```
topolarity, line 3211: I might be mistaken, but I don't think the backend should ever be asked to produce the field value of a comptime field.

Is it possible to assert here that the field is not comptime instead?

src/codegen/c.zig
```diff
@@ -1000,6 +1000,46 @@ pub const DeclGen = struct {
...
+            for (tuple.types) |field_ty, i| {
+                const val = tuple.values[i];
+                if (val.tag() != .unreachable_value) continue;
+
+                var name = std.ArrayList(u8).init(dg.gpa);
+                defer name.deinit();
+                try name.writer().print("field_{d}", .{i});
+
+                try buffer.append(' ');
+                try dg.renderTypeAndName(writer, field_ty, .{ .identifier = name.items }, .Mut, Value.initTag(.abi_align_default));
```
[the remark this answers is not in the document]
jmc-88: Good point, amended.

src/codegen/c.zig
jmc-88: I don't think that's necessary, see:

https://github.com/ziglang/zig/blob/f3f5a5d05b7056aceb408b701613242d053019ab/src/type.zig#L5528-L5529

`tuple.values[i] == .unreachable_value` should be sufficient to tell that the value is runtime-known.

topolarity: Yeah - I meant to recommend you assert the tag is `.unreachable_value` (rather than returning `CValue.none`)

However, I just checked that `CValue.none` already asserts if we try to render it anywhere, so this is fine as-is :+1:

PR #11218, opened 2022-03-18
AstGen: labeled blocks should always complete with a normal break

### LINKED ISSUE #11213: stage2: test produces invalid results when inside a labeled block
### Zig Version

0.10.0-dev.1211+0f0d27ce2

### Steps to Reproduce

Run the test below. Some notes:

* If you end the switch case with a noreturn value (like a panic), it passes.
* If you remove the labeled block, it passes.
* If you remove the labeled block to the outside of the switch and make the `else` break it (so you use the label), it passes.

My conclusion is that something about being (1) inside a labeled block expr and (2) NOT ending with a noreturn, is causing issues.

```zig
const std = @import("std");
const testing = std.testing;

test {
    const g: error{Test}!void = error.Test;

    var v: u32 = 0;
    hash(&v, g);
    try testing.expect(v == 1); // what we actually expect
}

fn hash(v: *u32, key: anytype) void {
    const Key = @TypeOf(key);

    if (@typeInfo(Key) == .ErrorSet) {
        v.* += 1;
        return;
    }

    switch (@typeInfo(Key)) {
        .ErrorUnion => blk: {
            const payload = key catch |err| {
                hash(v, err);
                break :blk;
            };

            hash(v, payload);

            // uncomment this and the test passes
            // @panic("Hello");
        },

        else => unreachable,
    }
}
```

### Expected Behavior

Pass

### Actual Behavior

Fail

mitchellh:
They aren't inline blocks by nature of being labeled.

I'm not sure I fully grokked `break_inline` properly in this case, so I'm not 100% confident in this fix. It fixes the previous issue and all previous behavior tests continue to pass.

Fixes #11213

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -1939,7 +1939,7 @@ fn blockExpr(
     if (token_tags[lbrace - 1] == .colon and
         token_tags[lbrace - 2] == .identifier)
     {
-        return labeledBlockExpr(gz, scope, rl, block_node, statements, .block);
+        return labeledBlockExpr(gz, scope, rl, block_node, statements);
     }
 
     try blockExprStmts(gz, scope, statements);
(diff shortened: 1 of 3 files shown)


mitchellh:
Note in the course of fixing this I did find another issue that exists with or without this PR. The following crashes the compiler. **This is the same behavior as before this PR, as far as I can tell.** So this PR does not cause this and we should look into this separately.

```

test {
    const a: i32 = blk: {
        if (false) break :blk 24;
        break :blk "foo"; // with or without this, a crash.
    };
    _ = a;
}
```

andrewrk:
The fix looks exactly right :+1:

PR #10159, opened 2021-11-16
CBE: pass math.zig and popcount.zig

doppioandante:
Summary:
- every commit is reviewable independently
- math.zig is till not passing because of f16 and f128; those are my next TODO, i can either continue here and wait for this to be merged or do a followup, as the reviewer prefers
- @clz and @popcount are implemented as naive functions in zig.h, I hope not to offend anyone, I'll open a follow up issues for converting them to a more efficient implementation that use intrinsics/LUTs (reference to: https://github.com/ziglang/zig/pull/9687)
- floored division and mod for integers are implemented ad hoc in zig.h, implementation is not efficient but correct
- the commit that substitutes the clz intrinsic in zig_sub_sat_u/s is droppable (cc @travisstaloch)

The last commit fixes the following scenario that occurred in math.zig:
```zig
test "xor" () {
}
fn test_xor() {
// ...
}
```
Currently those two get converted to the same function name and it won't compile. I'm not really sure about the fix, particularly whether it could happen for other reason, so I'm open for a better solution

PR #11229, opened 2022-03-19
CBE: implement mod, divFloor, divTrunc, and enable more passing tests

jmc-88:
Also enables the newly-passing `math.zig` tests, while splitting off the half-precision floating point tests until https://github.com/ziglang/zig/pull/10771 is merged.

### DIFF
--- a/src/link/C/zig.h
+++ b/src/link/C/zig.h
@@ -750,3 +750,81 @@ static inline uint128_t zig_bit_reverse_u128(uint128_t value, uint8_t zig_type_b
 }
 
 #define zig_bit_reverse_i128 zig_bit_reverse_u128
+
+static inline float zig_div_truncf(float numerator, float denominator) {
+    return __builtin_truncf(numerator / denominator);
+}
+
+static inline double zig_div_trunc(double numerator, double denominator) {
+    return __builtin_trunc(numerator / denominator);
+}
+
+static inline long double zig_div_truncl(long double numerator, long double denominator) {
+    return __builtin_truncf(numerator / denominator);
+}
+
+#define zig_div_trunc_f16  zig_div_truncf
+#define zig_div_trunc_f32  zig_div_truncf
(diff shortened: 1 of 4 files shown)


src/link/C/zig.h
```diff
@@ -750,3 +750,97 @@ static inline uint128_t zig_bit_reverse_u128(uint128_t value, uint8_t zig_type_b
...
+#define zig_div_trunc_f16  zig_div_truncf
+#define zig_div_trunc_f32  zig_div_truncf
+#define zig_div_trunc_f64  zig_div_trunc
+#define zig_div_trunc_f80  zig_div_truncl
+#define zig_div_trunc_f128 zig_div_truncl
+
+#define zig_div_trunc_int(ZigType, CType) \
+    static inline CType zig_div_trunc_##ZigType(CType numerator, CType denominator) { \
+        return numerator / denominator; \
+    }
```
andrewrk, line 775: If this is how div trunc for integers is defined then why don't we simply emit `a / b` instead of a call to this function? Looks like before this change, that's exactly what we did. Was there a problem with it?

andrewrk (changes requested):

src/link/C/zig.h
jmc-88: Very true, this was only convenient for a unified interface but we shouldn't need this code at all.

I changed the code in `genBody()` to instead selectively dispatch to `airBinOp(..., " / ")` or `airBinOpBuiltinCall(..., "div_trunc")` depending on the type, PTAL.

doppioandante:
When I did some work in this I had added [some more tests](https://github.com/ziglang/zig/pull/10159/files#diff-04d4ce948bc351d4b34d82e3d0488a98fb6aa143b81794c807bd6319d9d70026), maybe they could be useful (unfortunately I don't remember if I did for a specific reason lol).

PR #11239, opened 2022-03-20
Adding "value of" to clarify "expected type _", found _" error message

### LINKED ISSUE #9287: fix error message (add word "value")
```
const std = @import("std");

pub fn my_error_func(num: i32) void {
  std.log.info("num = {}", .{num});
}

pub fn main() anyerror!void {
  my_error_func(i32);
}
```
> error: expected type 'i32', found 'type' i32, 

if i send type as argument i give error.  I suggest changing the error message to this: 
> error: expected **value of** type 'i32', found 'type' i32,

daneelsan:
Fixes #9287.

Adds the words "value of" to clarify error messages of the form `"... expected type _, found _ ..."` (stage1).

Before:
```zig
./build/test.zig:7:19: error: expected type 'i32', found 'type'
    my_error_func(i32);
                  ^
```

After:
```zig
./build/test.zig:7:19: error: expected value of type 'i32', found 'type'
    my_error_func(i32);
                  ^
```

I don't think it is necessary to include the encountered value (e.g. `i32`) in the error message as the `^` already points to it.

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -3320,7 +3320,7 @@ test "pointer to non-byte-aligned field" {
       However, the pointer to a non-byte-aligned field has special properties and cannot
       be passed when a normal pointer is expected:
       </p>
-      {#code_begin|test_err|expected type#}
+      {#code_begin|test_err|expected value of type#}
 const std = @import("std");
 const expect = std.testing.expect;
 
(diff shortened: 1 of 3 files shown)


daneelsan:
The last part of the sentence could also be: `"..., found 'type' instead"`. But maybe that is just more verbose for no good reason.

andrewrk:
Thanks for the patch. See my comment here: https://github.com/ziglang/zig/issues/9287#issuecomment-1124319728

PR #11245, opened 2022-03-20
Add sem_open sem_close C definition to standard library

locriacyber:
Current `Thread.Semaphore` in standard library cannot be used across process.

Here's a simple example for using semaphore with current API:
```zig
const std = @import("std");
const PROT = std.os.PROT;
const MAP = std.os.MAP;
const sem_t = std.c.sem_t;

// make sure child always prints first
pub fn main() anyerror!void {
    var shm_addr = try std.os.mmap(null, @sizeOf(sem_t), PROT.READ | PROT.WRITE, MAP.SHARED | MAP.ANONYMOUS, 0, 0);
    var semaphore = @ptrCast(*sem_t, &shm_addr[0]);
    if (std.c.sem_init(semaphore, 1, 0) != 0) return error.SemaphoreFailed;
    const pid = try std.os.fork();
    if (pid == 0) {
        std.debug.print("child\n", .{});
        if (std.c.sem_post(semaphore) != 0) return error.SemaphoreFailed;
    } else {
        if (std.c.sem_wait(semaphore) != 0) return error.SemaphoreFailed;
        std.debug.print("parent\n", .{});
    }
}
```

I'd like to know how to achieve similar result without depending on libc or libpthread (at least on Linux). The following doesn't work:

```zig
const std = @import("std");
const PROT = std.os.PROT;
const MAP = std.os.MAP;
const Semaphore = std.Thread.Semaphore;

// make sure child always prints first
// now with zig's semaphore
pub fn main() anyerror!void {
    var shm_addr = try std.os.mmap(null, @sizeOf(Semaphore), PROT.READ | PROT.WRITE, MAP.SHARED | MAP.ANONYMOUS, -1, 0);
    var semaphore = @ptrCast(*Semaphore, &shm_addr[0]);
    semaphore.* = Semaphore{};
    
    const pid = try std.os.fork();
    if (pid == 0) {
        std.debug.print("child\n", .{});
        semaphore.post();
    } else {
        semaphore.wait();
        std.debug.print("parent\n", .{});
    }
}
```

### DIFF
--- a/lib/std/c.zig
+++ b/lib/std/c.zig
@@ -239,6 +239,8 @@ pub extern "c" fn pthread_getspecific(key: c.pthread_key_t) ?*anyopaque;
 pub extern "c" fn pthread_setspecific(key: c.pthread_key_t, value: ?*anyopaque) c_int;
 pub extern "c" fn sem_init(sem: *c.sem_t, pshared: c_int, value: c_uint) c_int;
 pub extern "c" fn sem_destroy(sem: *c.sem_t) c_int;
+pub extern "c" fn sem_open(name: [*:0]const u8, flag: c_int, mode: c.mode_t, value: c_uint) *c.sem_t;
+pub extern "c" fn sem_close(sem: *c.sem_t) c_int;
 pub extern "c" fn sem_post(sem: *c.sem_t) c_int;
 pub extern "c" fn sem_wait(sem: *c.sem_t) c_int;
 pub extern "c" fn sem_trywait(sem: *c.sem_t) c_int;


locriacyber:
What would the API be? Make the user map shared memory?

kprotty:
We could by removing the `FUTEX_PRIVATE_FLAG` [here](https://github.com/ziglang/zig/blob/master/lib/std/Thread/Futex.zig#L144-L157), but this would only allow cross-process usage specifically for non-libc linux targets; The libc targets use pthread primitives atm, and those require special setup iirc for cross-process synchronization. Removing `FUTEX_PRIVATE_FLAG` would also make normal (single-process) waits/wakes slower. 

IMO this should be its own thing (maybe outside the stdlib?) due to the varying apis needed for it be cross-platform and its usefulness vs a named/anonymous/unix pipe(). Adding the posix sem functions to `std.c` as done currently doesn't hurt though.

locriacyber:
I tried using pthread with mmap (see pull request description). The semaphore just need to be placed in shared memory region.

locriacyber:
Also, is there a reason to fall back to libc if platform-specific implementation is available? Is it better to use as little libc even if libc is available?

kprotty:
> I tried using pthread with mmap (see pull request description)

I don't see the use of pthread in shared mode in the PR description. You need to initialize the pthread types with things like [pthread_mutexattr_setpshared](https://linux.die.net/man/3/pthread_mutexattr_setpshared) in order to use it across shared memory. This is the "special setup" I was talking about. 

> is there a reason to fall back to libc if platform-specific implementation is available?

In this case, I don't think so. The only one interacting with the semaphore is you. Sometimes it's important for users who are using libc stuff that expect the stdlib to perform required setup: e.g. spawning threads with `pthread_create` instead of the `clone` syscall (which `pthread_create` calls internally) in order for `pthread_getspecific` to work as thread locals.

> I think I'll write this as a third-party library first.

Sure thing. Is there anything else you want to add to this PR before its merged/closed?

locriacyber:
> I don't see the use of pthread in shared mode in the PR description. 
`sem_init` is backed by libpthread on Linux from POSIX.1-2001. It's simpler to use than `pthread_....` from later POSIX versions. I'd rather use libuv than to directly use the later POSIX API - libuv is easier to use.

> Is there anything else you want to add to this PR before its merged/closed?
No.

Issue #10920, opened 2022-02-17
stage1: miscompilation on writing with comptime-computed offsets into comptime constructed buffer


matu3ba:
### Zig Version

0.10.0-dev.787+5aa35f62c

### Steps to Reproduce

Run
```zig
zig build-exe comptimeMiscompilation.zig
./comptimeMiscompilation
```
Inspect with an editor the binary `comptimeMiscompilation`

File `comptimeMiscompilation.zig`:
```zig
const std = @import("std");
const mem = std.mem;

const strings = [_][]const u8{ "Hello", "There" };
pub fn main() !void {
    comptime var buf_size = 0;
    inline for (strings) |str| {
        buf_size += str.len;
    }
    const len_offsets = strings.len;
    comptime var buf = [_]u8{0} ** buf_size;
    comptime var offsets = [_]u64{0} ** len_offsets;
    //@compileLog(strings.len);
    //@compileLog(strings[0].len);
    //@compileLog(strings[1].len);
    //@compileLog(buf);
    //@compileLog(offsets);
    comptime var cur_offset = 0;
    inline for (strings) |str, i| {
        //@compileLog(cur_offset);
        offsets[i] = cur_offset;
        //@compileLog(cur_offset);
        //@compileLog(str.len);
        //@compileLog(cur_offset + str.len);
        //@compileLog(str.len);
        //@compileLog(buf[cur_offset .. cur_offset + str.len].len);
        mem.copy(u8, buf[cur_offset .. cur_offset + str.len], str);
        cur_offset += str.len;
    }
    std.debug.print("{d}\n", .{buf_size});
    std.debug.print("{d}\n", .{len_offsets});
    std.debug.print("{s}\n", .{buf});
    std.debug.print("{d}\n", .{offsets});
    std.debug.print("{d}\n", .{cur_offset});
}
```

### Expected Behavior

1. No error. I checked the offsets with the debug logs and they are correct. Without `mem.copy` the program works.
2. Even if not, the program should crash at comptime.
3. If comptime fails (ie due to versatility of stage1), it should crash instead of writing the crash data into the binary that is being linked.

### Actual Behavior

Output of the broken comptime execution is written into the resulting elf binary (on Linux) stripped by the nonvisual parts:
```txt
{ ... }ELF/proc/self/exe.debug_info.debug_abbrev.debug_str.debug_line.debug_rangessentinel mismatch
attempt to cast negative value to unsigned integerPanicked during a panic. Aborting.
[...]

Issue #10684, opened 2022-01-24
Unexpected behavior when creating a slice of another slice


hronro:
### Zig Version

0.9.0

### Steps to Reproduce

```zig
const foo = "12345678";
const bar = foo[2..7]; // "34567"
const baz = bar[1..3];
```

### Expected Behavior

`baz` is `"45"`.

### Actual Behavior

`baz` is `"34"`.

squeek502:
Note: changing `bar` to `var` makes it work as expected:
```zig
const foo = "12345678";
var bar = foo[2..7]; // "34567"
const foobar = bar[1..3]; // "45"
```

nektro:
using slices also makes it work
```zig
const foo: []const u8 = "hello world";
const bar: []const u8 = foo[2..7];
const baz: []const u8 = bar[1..3];
```

but the bug remains when using the default types
```zig
const foo: *const [11:0]u8 = "hello world";
const bar: *const [5]u8 = foo[2..7];
const baz: *const [2]u8 = bar[1..3];
```

matu3ba:
More awkwardness:
```zig
const std = @import("std");
pub fn main() !void {
    const foo = "1234";
    const bar = foo[0..4]; // "1234"
    const s2 = bar[1..2]; // "2"
    std.debug.print("s2: {s} {s}\n", .{ s2, @TypeOf(s2) }); // outputs 1 with type [1]u8
    const s3 = bar[2..3]; // "3"
    std.debug.print("s3: {s} {s}\n", .{ s3, @TypeOf(s3) }); // outputs 1 with type [1]u8
    const s4 = bar[3..4]; // "4"
    std.debug.print("s4: {s} {s}\n", .{ s4, @TypeOf(s4) }); // outputs 1 with type [1:0]u8
    const s5 = bar[2..4]; // "3"
    std.debug.print("s5: {s} {s}\n", .{ s5, @TypeOf(s5) }); // outputs 12 with type [2:0]u8
    const s6 = bar[3..4]; // "3"
    std.debug.print("s6: {s} {s}\n", .{ s6, @TypeOf(s6) }); // outputs 1 with type [1:0]u8
}
```

PR #11249, opened 2022-03-21
stage2: slice bounds checks at comptime, slicing with sentinel matching stage1

mitchellh:
This PR has two changes split by commit, respectively.

First, it introduces out of bounds checking for comptime slicing. I matched the error messages with stage1.

Second, **this one I'm unsure if we want to match stage1 behavior** is to make it so that the following is true:

```zig
test {
    comptime {
        var array = [0:0]u8{};
        var src_slice: [:0]u8 = &array;
        var slice = src_slice[0..0];
        try testing.expect(@TypeOf(slice) == *[0]u8);
        // try testing.expect(@TypeOf(slice) == *[0:0]u8); prior stage2 behavior
    }
}
```

This is the stage1 behavior. The stage 2 behavior was to create a `*[0:0]u8`. 

What makes this weird is that arrays behave differently in stage1. For arrays, if the end of a slice equals the length and the source type has a sentinel, it adds the sentinel. Stage2 previously did this behavior for BOTH arrays and slices. Do we want to match stage1? Or keep the stage2 behavior?

I split the behavior by commit so if we want to keep the stage2 behavior, you can just cherry pick my first commit. Otherwise, take both.

No test because the behavior test in `slice.zig` tests it, but we can't enable it yet because there is one more separate stage2 issue.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -19553,6 +19553,14 @@ fn analyzeSlice(
             if (!end_is_len) {
                 const end = try sema.coerce(block, Type.usize, uncasted_end_opt, end_src);
                 if (try sema.resolveMaybeUndefVal(block, end_src, end)) |end_val| {
+                    if (end_val.compare(.gt, len_val, Type.usize)) {
+                        return sema.fail(
+                            block,
+                            end_src,
+                            "end index {} out of bounds for array of length {}",
+                            .{ end_val.fmtValue(Type.usize), len_val.fmtValue(Type.usize) },
+                        );
+                    }
                     if (end_val.eql(len_val, Type.usize)) {
                         end_is_len = true;
                     }
@@ -19571,6 +19579,14 @@ fn analyzeSlice(
                             .data = slice_val.sliceLen(),
                         };
                         const slice_len_val = Value.initPayload(&int_payload.base);
+                        if (end_val.compare(.gt, slice_len_val, Type.usize)) {
+                            return sema.fail(
+                                block,
+                                end_src,
+                                "end index {} out of bounds for slice of length {}",
+                                .{ end_val.fmtValue(Type.usize), slice_len_val.fmtValue(Type.usize) },
+                            );
+                        }
                         if (end_val.eql(slice_len_val, Type.usize)) {
                             end_is_len = true;
                         }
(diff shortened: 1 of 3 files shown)


matu3ba:
> What makes this weird is that arrays behave differently in stage1. For arrays, if the end of a slice equals the length and the source type has a sentinel, it adds the sentinel. Stage2 previously did this behavior for BOTH arrays and slices. Do we want to match stage1? Or keep the stage2 behavior?

Use case: Stage1 has several serious issues with offsets which break utterly in #10920 with a simplification in #10684 Do both compile properly with this change or what are the missing bits?

Questions for edge cases: 1.+2. What do you believe should be the behavior for overlapping (comptime) slices?
3.+4. What should be behavior for overlapping (comptime) arrays?

mitchellh:
Okay, okay yeah. I can help clean up a lot of this (definitely all the comptime stuff, at least) if we figure out what the expect behavior is. Here is another example (not touched by this PR):

```zig
const std = @import("std");
const testing = std.testing;

test {
    comptime {
        var array = [5:0]u8{ 1, 2, 3, 4, 5 };
        var slice: [:0]u8 = &array;
        
        // stage2
        try testing.expect(@TypeOf(slice[1..]) == *[4:0]u8);
        
        // stage1
        try testing.expect(@TypeOf(slice[1..]) == [:0]u8);
    }

    var array = [5:0]u8{ 1, 2, 3, 4, 5 };
    var slice: [:0]u8 = &array;
    comptime try testing.expect(@TypeOf(slice[1..]) == [:0]u8); // stage1 and stage2
}
```

andrewrk:
> Do we want to match stage1? Or keep the stage2 behavior?

Keep stage2 behavior. For this example, the result type of `*[0:0]u8` is correct and has more metadata than `*[0]u8`, which failed to understand that there is indeed a 0 byte being pointed at.

src/Sema.zig
```diff
@@ -19553,6 +19553,9 @@ fn analyzeSlice(
             if (!end_is_len) {
                 const end = try sema.coerce(block, Type.usize, uncasted_end_opt, end_src);
                 if (try sema.resolveMaybeUndefVal(block, end_src, end)) |end_val| {
+                    if (end_val.compare(.gt, len_val, Type.usize)) {
+                        return sema.fail(block, end_src, "out of bounds slice", .{});
```
andrewrk, line 19557: Can we match the corresponding array message? "index {d} out of bounds of slice of length {d} (+1 sentinel)"

src/Sema.zig
```diff
@@ -19601,6 +19607,15 @@ fn analyzeSlice(
         break :s null;
     };
 
+    // requirement: start <= end
```
andrewrk, line 19616: does this handle sentinels?

andrewrk (changes requested):
A couple requests here, plus new error messages should get tests. Right now they go into test/compile_errors.zig

src/Sema.zig
mitchellh: I'm sure we can, might wait for Vexu's #11253 which I'm guessing you'll look at today. That way we can use that since we have the type.

andrewrk: Merged!

mitchellh:
@andrewrk Okay, all cleaned up with tests. Sentinel slicing/errors work fine. Decided to keep this to just bounds checking to keep it easy, will do the type cleanups we discussed in a separate PR.

PR #11250, opened 2022-03-21
macho: extend CodeSignature to accept entitlements

kubkon:
With this change, we can now bake in entitlements into the binary.
Additionally, I see this as the first step towards full code signature
support which includes baking in Apple issued certificates for
redistribution, etc.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -815,6 +815,8 @@ pub const InitOptions = struct {
     native_darwin_sdk: ?std.zig.system.darwin.DarwinSDK = null,
     /// (Darwin) Install name of the dylib
     install_name: ?[]const u8 = null,
+    /// (Darwin) Path to entitlements file
+    entitlements: ?[]const u8 = null,
 };
 
 fn addPackageTableToCacheHash(
@@ -1624,6 +1626,7 @@ pub fn create(gpa: Allocator, options: InitOptions) !*Compilation {
             .enable_link_snapshots = options.enable_link_snapshots,
             .native_darwin_sdk = options.native_darwin_sdk,
             .install_name = options.install_name,
+            .entitlements = options.entitlements,
         });
         errdefer bin_file.destroy();
         comp.* = .{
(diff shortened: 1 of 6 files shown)


src/main.zig
```diff
@@ -1036,6 +1037,10 @@ fn buildOutputType(
                         } else {
                             enable_link_snapshots = true;
                         }
+                    } else if (mem.eql(u8, arg, "--entitlements")) {
```
andrewrk, line 1040: don't forget to update the corresponding usage text above this function

andrewrk (changes requested):
One minor thing, and a question: what is the clang / apple tooling way to specify an entitlements file? I didn't find an `--entitlements` command line parameter for clang. Does it have a corresponding flag? If so we should adjust `zig cc` to recognize it too.

kubkon:
> One minor thing, and a question: what is the clang / apple tooling way to specify an entitlements file? I didn't find an `--entitlements` command line parameter for clang. Does it have a corresponding flag? If so we should adjust `zig cc` to recognize it too.

It doesn't, and most likely never will. Linker only creates the minimal adhoc signature, while `codesign` tool is used as part of the full Xcode toolchain to embed advanced code signature objects which include entitlements, etc.

src/Compilation.zig
```diff
@@ -2302,6 +2305,7 @@ fn addNonIncrementalStuffToCacheManifest(comp: *Compilation, man: *Cache.Manifes
 
     man.hash.addListOfBytes(comp.clang_argv);
 
+    man.hash.addOptionalBytes(comp.bin_file.options.entitlements);
```
andrewrk, line 2308: there's a "Mach-O specific stuff" section about 30 lines below. Also this is a file so it needs to be added to the manifest as a file dependency:

```suggestion
    man.hash.addOptionalFile(comp.bin_file.options.entitlements);
```

Same deal with the other cache API usage in the linker.

src/main.zig
```diff
@@ -433,6 +433,7 @@ const usage_build_generic =
     \\  -framework [name]              (Darwin) link against framework
     \\  -F[dir]                        (Darwin) add search path for frameworks
     \\  -install_name=[value]          (Darwin) add dylib's install name
+    \\  --entitlements=[path]          (Darwin) add path to entitlements file for embedding in code signature
```
andrewrk, line 436: Looks like it does not parse with an equal sign:

```suggestion
    \\  --entitlements [path]          (Darwin) add path to entitlements file for embedding in code signature
```

andrewrk (changes requested):

andrewrk (approved):

PR #11254, opened 2022-03-21
Add support for f128 values in std.fmt.parseFloat. 

### LINKED ISSUE #11169: `std.fmt.parseFloat` does not parse 9007199254740993.0 correctly
Implementation is here: https://github.com/ziglang/zig/blob/master/lib/std/fmt/parse_float.zig

```diff
--- a/lib/std/fmt/parse_float.zig
+++ b/lib/std/fmt/parse_float.zig
@@ -425,4 +425,6 @@ test "fmt.parseFloat" {
             try expect(approxEqAbs(T, try parseFloat(T, "2.71828182845904523536"), @as(T, 2.718281828459045), epsilon));
         }
     }
+
+    try expectEqual(try parseFloat(f128, "9007199254740993.0"), 9007199254740993.0);
 }
```

```
$ ./zig test ../lib/std/fmt/parse_float.zig 
Test [1/1] test "parseFloat"... expected 9.007199254740994e+15, found 9.007199254740992e+15
Test [1/1] test "parseFloat"... FAIL (TestExpectedEqual)
```

This is causing a behavior test failure in the self-hosted compiler, which relies on `std.fmt.parseFloat`:

https://github.com/ziglang/zig/blob/9eceba248511849ff77a90876424f9930bf7f9f9/test/behavior/floatop.zig#L679

danielchasehooper:
Add f128 support to std.fmt.parseFloat by transitioning it to sqlite's float parser, adapted to support 128 floats. The old implementation worked in f64 internally, and then upcast to f128 at the last moment.

Additionally, this fixes a typo in parse_f128.c - my initial approach for this PR was to port *that* parser to zig as recommended in https://github.com/ziglang/zig/issues/11169#issuecomment-1067629972, but found it had issues parsing large f64 values like 144115188075855870, even after fixing the typo. I suspect there may be other issues with parse_f128.c, but due to the way its style, it's hard to know what else might be wrong. I also explored adapting the existing parser to truly support f128 (and not just upcast from f64 to f128), but in early tests it exhibited incorrect behavior for parsing large f64 values, and it relied on some custom 96bit-integer code that made it difficult to adapt to f128. Sqlite's f64 parser behaved correctly out of the box and is much easier to read, so was easy to adapt to support f128.

I'm not sure what the implications are for relying on f128 math in parseFloat now - vs working in f64 and then casting to f16/f32/f64/f128 at the last moment like the old implementation did. Is f128 math emulated in software for targets that don't natively support it? Should we make parseFloat work in f64 when f128 isn't the requested return type?

Fixes #11169

### DIFF
--- a/lib/std/fmt/parse_float.zig
+++ b/lib/std/fmt/parse_float.zig
@@ -1,336 +1,23 @@
-// Adapted from https://github.com/grzegorz-kraszewski/stringtofloat.
-
-// MIT License
 //
-// Copyright (c) 2016 Grzegorz Kraszewski
+// Adapted from sqlite3's sqlite3AtoF()
+// https://github.com/mackyle/sqlite/blob/ae5d3aa91a794f6f1486b3f453ec44c0aa4c086b/src/util.c#L375-L571
 //
-// Permission is hereby granted, free of charge, to any person obtaining a copy
-// of this software and associated documentation files (the "Software"), to deal
-// in the Software without restriction, including without limitation the rights
-// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-// copies of the Software, and to permit persons to whom the Software is
-// furnished to do so, subject to the following conditions:
+// adds support for parsing f128 floats, "nan", "inf"
 //
-// The above copyright notice and this permission notice shall be included in all
-// copies or substantial portions of the Software.
+// Returns a float if the string is a valid number, or error.invalidCharacter
+// if the string is empty or contains extraneous text.  Valid numbers
+// are in one of these formats:
 //
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-// SOFTWARE.
+//    [+-]digits[E[+-]digits]
+//    [+-]digits.[digits][E[+-]digits]
+//    [+-].digits[E[+-]digits]
 //
-
-// Be aware that this implementation has the following limitations:
+// underscore characters may appear zero or more times between
+// digits and are ignored by the parser: 500_000
 //
-// - Is not round-trip accurate for all values
-// - Only supports round-to-zero
-// - Does not handle denormals
 
 const std = @import("std");
 const ascii = std.ascii;
-
-// The mantissa field in FloatRepr is 64bit wide and holds only 19 digits
-// without overflowing
-const max_digits = 19;
-
-const f64_plus_zero: u64 = 0x0000000000000000;
-const f64_minus_zero: u64 = 0x8000000000000000;
-const f64_plus_infinity: u64 = 0x7FF0000000000000;
-const f64_minus_infinity: u64 = 0xFFF0000000000000;
-
-const Z96 = struct {
-    d0: u32,
-    d1: u32,
-    d2: u32,
-
-    // d = s >> 1
-    inline fn shiftRight1(d: *Z96, s: Z96) void {
-        d.d0 = (s.d0 >> 1) | ((s.d1 & 1) << 31);
-        d.d1 = (s.d1 >> 1) | ((s.d2 & 1) << 31);
-        d.d2 = s.d2 >> 1;
-    }
-
-    // d = s << 1
-    inline fn shiftLeft1(d: *Z96, s: Z96) void {
-        d.d2 = (s.d2 << 1) | ((s.d1 & (1 << 31)) >> 31);
-        d.d1 = (s.d1 << 1) | ((s.d0 & (1 << 31)) >> 31);
-        d.d0 = s.d0 << 1;
-    }
-
-    // d += s
-    inline fn add(d: *Z96, s: Z96) void {
-        var w = @as(u64, d.d0) + @as(u64, s.d0);
-        d.d0 = @truncate(u32, w);
-
-        w >>= 32;
-        w += @as(u64, d.d1) + @as(u64, s.d1);
-        d.d1 = @truncate(u32, w);
-
-        w >>= 32;
-        w += @as(u64, d.d2) + @as(u64, s.d2);
-        d.d2 = @truncate(u32, w);
-    }
-
-    // d -= s
-    inline fn sub(d: *Z96, s: Z96) void {
-        var w = @as(u64, d.d0) -% @as(u64, s.d0);
-        d.d0 = @truncate(u32, w);
-
-        w >>= 32;
-        w += @as(u64, d.d1) -% @as(u64, s.d1);
-        d.d1 = @truncate(u32, w);
-
-        w >>= 32;
-        w += @as(u64, d.d2) -% @as(u64, s.d2);
-        d.d2 = @truncate(u32, w);
-    }
-};
-
-const FloatRepr = struct {
-    negative: bool,
-    exponent: i32,
-    mantissa: u64,
-};
-
-fn convertRepr(comptime T: type, n: FloatRepr) T {
-    const mask28: u32 = 0xf << 28;
-
-    var s: Z96 = undefined;
-    var q: Z96 = undefined;
-    var r: Z96 = undefined;
-
-    s.d0 = @truncate(u32, n.mantissa);
-    s.d1 = @truncate(u32, n.mantissa >> 32);
-    s.d2 = 0;
-
-    var binary_exponent: i32 = 92;
-    var exp = n.exponent;
-
-    while (exp > 0) : (exp -= 1) {
-        q.shiftLeft1(s); // q = p << 1
-        r.shiftLeft1(q); // r = p << 2
-        s.shiftLeft1(r); // p = p << 3
-        s.add(q); // p = (p << 3) + (p << 1)
-
-        while (s.d2 & mask28 != 0) {
-            q.shiftRight1(s);
-            binary_exponent += 1;
-            s = q;
-        }
-    }
-
-    while (exp < 0) {
-        while (s.d2 & (1 << 31) == 0) {
-            q.shiftLeft1(s);
-            binary_exponent -= 1;
-            s = q;
-        }
-
-        q.d2 = s.d2 / 10;
-        r.d1 = s.d2 % 10;
-        r.d2 = (s.d1 >> 8) | (r.d1 << 24);
-        q.d1 = r.d2 / 10;
-        r.d1 = r.d2 % 10;
-        r.d2 = ((s.d1 & 0xff) << 16) | (s.d0 >> 16) | (r.d1 << 24);
-        r.d0 = r.d2 / 10;
-        r.d1 = r.d2 % 10;
-        q.d1 = (q.d1 << 8) | ((r.d0 & 0x00ff0000) >> 16);
-        q.d0 = r.d0 << 16;
-        r.d2 = (s.d0 *% 0xffff) | (r.d1 << 16);
-        q.d0 |= r.d2 / 10;
-        s = q;
-
-        exp += 1;
-    }
-
-    if (s.d0 != 0 or s.d1 != 0 or s.d2 != 0) {
-        while (s.d2 & mask28 == 0) {
-            q.shiftLeft1(s);
-            binary_exponent -= 1;
-            s = q;
-        }
-    }
-
-    binary_exponent += 1023;
-
-    const repr: u64 = blk: {
-        if (binary_exponent > 2046) {
-            break :blk if (n.negative) f64_minus_infinity else f64_plus_infinity;
-        } else if (binary_exponent < 1) {
-            break :blk if (n.negative) f64_minus_zero else f64_plus_zero;
-        } else if (s.d2 != 0) {
-            const binexs2 = @intCast(u64, binary_exponent) << 52;
-            const rr = (@as(u64, s.d2 & ~mask28) << 24) | ((@as(u64, s.d1) + 128) >> 8) | binexs2;
-            break :blk if (n.negative) rr | (1 << 63) else rr;
-        } else {
-            break :blk 0;
-        }
-    };
-
-    const f = @bitCast(f64, repr);
-    return @floatCast(T, f);
-}
-
-const State = enum {
-    MaybeSign,
-    LeadingMantissaZeros,
-    LeadingFractionalZeros,
-    MantissaIntegral,
-    MantissaFractional,
-    ExponentSign,
-    LeadingExponentZeros,
-    Exponent,
-};
-
-const ParseResult = enum {
-    Ok,
-    PlusZero,
-    MinusZero,
-    PlusInf,
-    MinusInf,
-};
-
-fn parseRepr(s: []const u8, n: *FloatRepr) !ParseResult {
-    var digit_index: usize = 0;
-    var negative_exp = false;
-    var exponent: i32 = 0;
-
-    var state = State.MaybeSign;
-
-    var i: usize = 0;
-    while (i < s.len) {
-        const c = s[i];
-
-        switch (state) {
-            .MaybeSign => {
-                state = .LeadingMantissaZeros;
-
-                if (c == '+') {
-                    i += 1;
-                } else if (c == '-') {
-                    n.negative = true;
-                    i += 1;
-                } else if (ascii.isDigit(c) or c == '.') {
-                    // continue
-                } else {
-                    return error.InvalidCharacter;
-                }
-            },
-            .LeadingMantissaZeros => {
-                if (c == '0') {
-                    i += 1;
-                } else if (c == '.') {
-                    i += 1;
-                    state = .LeadingFractionalZeros;
-                } else if (c == '_') {
-                    i += 1;
-                } else {
-                    state = .MantissaIntegral;
-                }
-            },
-            .LeadingFractionalZeros => {
-                if (c == '0') {
-                    i += 1;
-                    if (n.exponent > std.math.minInt(i32)) {
-                        n.exponent -= 1;
-                    }
-                } else {
-                    state = .MantissaFractional;
-                }
-            },
-            .MantissaIntegral => {
-                if (ascii.isDigit(c)) {
-                    if (digit_index < max_digits) {
-                        n.mantissa *%= 10;
-                        n.mantissa += c - '0';
-                        digit_index += 1;
-                    } else if (n.exponent < std.math.maxInt(i32)) {
-                        n.exponent += 1;
-                    }
-
-                    i += 1;
-                } else if (c == '.') {
-                    i += 1;
-                    state = .MantissaFractional;
-                } else if (c == '_') {
-                    i += 1;
-                } else {
-                    state = .MantissaFractional;
-                }
-            },
-            .MantissaFractional => {
-                if (ascii.isDigit(c)) {
-                    if (digit_index < max_digits) {
-                        n.mantissa *%= 10;
-                        n.mantissa += c - '0';
-                        n.exponent -%= 1;
-                        digit_index += 1;
-                    }
-
-                    i += 1;
-                } else if (c == 'e' or c == 'E') {
-                    i += 1;
-                    state = .ExponentSign;
-                } else if (c == '_') {
-                    i += 1;
-                } else {
-                    state = .ExponentSign;
-                }
-            },
-            .ExponentSign => {
-                if (c == '+') {
-                    i += 1;
-                } else if (c == '_') {
-                    return error.InvalidCharacter;
-                } else if (c == '-') {
-                    negative_exp = true;
-                    i += 1;
-                }
-
-                state = .LeadingExponentZeros;
-            },
-            .LeadingExponentZeros => {
-                if (c == '0') {
-                    i += 1;
-                } else if (c == '_') {
-                    i += 1;
-                } else {
-                    state = .Exponent;
-                }
-            },
-            .Exponent => {
-                if (ascii.isDigit(c)) {
-                    if (exponent < std.math.maxInt(i32) / 10) {
-                        exponent *= 10;
-                        exponent += @intCast(i32, c - '0');
-                    }
-
-                    i += 1;
-                } else if (c == '_') {
-                    i += 1;
-                } else {
-                    return error.InvalidCharacter;
-                }
-            },
-        }
-    }
-
-    if (negative_exp) exponent = -exponent;
-    n.exponent += exponent;
-
-    if (n.mantissa == 0) {
-        return if (n.negative) .MinusZero else .PlusZero;
-    } else if (n.exponent > 309) {
-        return if (n.negative) .MinusInf else .PlusInf;
-    } else if (n.exponent < -328) {
-        return if (n.negative) .MinusZero else .PlusZero;
-    }
-
-    return .Ok;
-}
 
 fn caseInEql(a: []const u8, b: []const u8) bool {
     if (a.len != b.len) return false;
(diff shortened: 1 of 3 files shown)


danielchasehooper:
Not sure what's going on with the test failure. `zig test test/behavior/union.zig` works locally for me (arm64 macOS). Any insight would be appreciated.

matu3ba:
> but found it had issues parsing large f64 values like 144115188075855870, even after fixing the typo. I suspect there may be other issues with parse_f128.c, but due to its style, it's hard to know what else might be wrong.

1. Do you have the test examples anywhere, which break/are wrong with parse_f128.c?
2. How did you ensure that the deviation from f64 was not due to better accuracy?
3. Did you cross-compare with other implementations like these https://nigeltao.github.io/blog/2020/eisel-lemire.html ?
4. Any information on test data would help on this.

Also: I found this library, which is MIT licensed: http://www.kerbaya.com/ieee754lib/index.html
Also: You can use http://www.jhauser.us/arithmetic/TestFloat-3/doc/testfloat.html for testing floats/the implementation.
Also: There appears to be only proprietary licensed test programs for 128bit floats. :(
"IEEE 754 has no official tests for floating-point but there are well-known third party tools to check such as John Hauser's TestFloat." https://libre-soc.org/resources/

>  casting to f16/f32/f64/f128 at the last moment like the old implementation did

Accuracy is lower with less bits, so one wants to use all of them and not upcast things.

danielchasehooper:
@matu3ba You probably know this, but for clarity I'll mention it: parse_f128.c and the previous implementation of parseFloat that this PR replaces are two separate parsers. This PR is not directly about parse_f128.c - it's about making parseFloat pass tests that were previously failing. Those tests now pass.

> 1. Do you have the test examples anywhere, which break/are wrong with parse_f128.c?

Mentioning parse_f128.c may have been a distraction. I only mentioned it because at first my approach for this PR was to port it to zig in order to make parseFloat support f128. Once I noticed that parse_f128.c had typos and my port had strange behavior for some numbers, I abandoned that path because the port was hard enough even when I had faith in the source material. It's possible that parse_f128.c is perfect (*besides that one typo this PR corrects*) and the issue was in my zig port of it. But the code is so unreadable I abandoned it. Regardless, any potential issues with it are separate from this PR. I'd be happy to write a new GitHub issue about testing parse_f128.c if you think it's worth it. Probably best to forget I mentioned parse_f128.c at all. 

> 2. How did you ensure that the deviation from f64 was not due to better accuracy?
 
144115188075855870 can be represented exactly by both f64 and f128. The previous implementation of parseFloat didn't get it right when parsing to f64 OR f128. Even if it was able to get that specific number correct, the previous parseFloat used f64s internally, and so you always got f64's range and accuracy, even if asking for a f128.

> 3. Did you cross-compare with other implementations like these https://nigeltao.github.io/blog/2020/eisel-lemire.html ?

I compared the behavior of this PR against c's atof and found that they matched for 64 bit floats. atof and that implementation you linked to do not support f128, so I wasn't able to verify any f128 numbers. One of the tests for f128 I manually figured out the 128 bits the number should be parsed as - the test has that written as hex.

> Do you [think testfloat](http://www.jhauser.us/arithmetic/TestFloat-3/doc/testfloat.html) can be hacked up to generate useful test cases for parsing?

I'll look into this... more rigorous testing beyond what we had for the previous implementation of parseFloat may be a separate task. 

> Accuracy is lower with less bits, so one wants to use all of them and not upcast things.
I agree, and the parser working in a higher bit depth is the point of this PR. I was suggesting that maybe it would be worthwhile to make parseFloat be written generically so that its calculations happen in f64 when an f64 number is requested. This would have benefits for targets which f128 is software emulated/not supported. If a f128 type was requested, then yes, it would work in the higher bit depth for accuracy.

topolarity:
Really clean and readable implementation - Very nice work :+1:

Can you clarify the rounding behavior, and any support for subnormals? Based on the code, those probably depend on the configuration of the FPU/softfloat library at the time of parsing?

lib/std/fmt/parse_float.zig
```diff
@@ -425,4 +248,22 @@ test "fmt.parseFloat" {
...
+
+    try expectEqual(@bitCast(u64, try parseFloat(f64, "9007199254740993")), 0x4340000000000000); // rounded down
+
+    // test precision of f128
+    try expectEqual(@bitCast(u128, try parseFloat(f128, "9007199254740993")), 0x40340000000000000800000000000000); // exact
+
+    // test range of f128
+    // at time of writing (Mar 2021), zig prints f128 values larger than f64 as "inf",
+    // so I'm not 100% sure this hex literal is the corrent parse of 1e4930
+    try expectEqual(@bitCast(u128, try parseFloat(f128, "1e4930")), 0x7ff8136c69ce8adff4397b050cae44c7);
```
topolarity, line 268: This looks like it's off by 1 bit to me - should be `0x7ff8136c69ce8adff4397b050cae44c6`

Based on a rudimentary calculation in Python, intended to get the error in ulp:
```python
>>> exp = 0x7ff8 - 0x3fff
>>> (10**4930 - 0x1136c69ce8adff4397b050cae44c7 * 2**(exp-112)) / 2**(exp-112)
-1.2945749668158448
>>> (10**4930 - 0x1136c69ce8adff4397b050cae44c6 * 2**(exp-112)) / 2**(exp-112)
-0.2945749668158449
``` 
I might have made a mistake in the calculation, of course

andrewrk:
> Not sure what's going on with the test failure: `992/1116 test.comptime union field access... FAIL (TestUnexpectedResult)`
> 
> `zig test test/behavior/union.zig` works locally for me (arm64 macOS). Any insight would be appreciated.

Disable the non-LLVM backends in the newly passing test case:

```
    if (builtin.zig_backend == .stage2_wasm) return error.SkipZigTest; // TODO
    if (builtin.zig_backend == .stage2_c) return error.SkipZigTest; // TODO
    if (builtin.zig_backend == .stage2_x86_64) return error.SkipZigTest; // TODO
    if (builtin.zig_backend == .stage2_arm) return error.SkipZigTest; // TODO
    if (builtin.zig_backend == .stage2_aarch64) return error.SkipZigTest; // TODO
```

These other backends do not handle f128 yet.

lib/std/fmt/parse_float.zig
```diff
@@ -1,377 +1,195 @@
...
-    d1: u32,
-    d2: u32,
-
-    // d = s >> 1
-    inline fn shiftRight1(d: *Z96, s: Z96) void {
-        d.d0 = (s.d0 >> 1) | ((s.d1 & 1) << 31);
-        d.d1 = (s.d1 >> 1) | ((s.d2 & 1) << 31);
-        d.d2 = s.d2 >> 1;
-    }
+fn caseInEql(a: []const u8, b: []const u8) bool {
```
andrewrk, line 22: this function is available with `ascii.eqlIgnoreCase`

danielchasehooper: done.

danielchasehooper:
@andrewrk 
> Disable the non-LLVM backends in the newly passing test case:

I'm confused, The failing test cases on the ci server seem unrelated to the `behavior/floatop.zig "float literal at compile time not lossy"` test - How does disabling this test make other tests start succeeding in the ci (`test.comptime union field access`)?

Additionally, parseFloat now requires f128 internally - is that going to be an issue for the backends that don't support f128? My musings from my previous comment are related, and I'm curious to get your take:

>maybe it would be worthwhile to make parseFloat be written generically so that its calculations happen in f64 when an f64 number is requested. This would have benefits for targets which f128 is software emulated/not supported. If a f128 type was requested, then yes, it would work in the higher bit depth for accuracy.

andrewrk:
> I'm confused, The failing test cases on the ci server seem unrelated to the `behavior/floatop.zig "float literal at compile time not lossy"` test - How does disabling this test make other tests start succeeding in the ci (`test.comptime union field access`)?

I didn't look closely when I made this suggestion. You would have hit the problem that I mentioned next if you weren't already hitting this one.

The command that is failing, on both x86_64-macos and x86_64-linux, is:

```
+ stage2/bin/zig test test/behavior.zig -I test -fLLVM
553/1116 test.anytype params... FAIL (TestUnexpectedResult)
838/1116 test.fully anonymous struct... FAIL (TestUnexpectedResult)
839/1116 test.fully anonymous list literal... FAIL (TestUnexpectedResult)
992/1116 test.comptime union field access... FAIL (TestUnexpectedResult)
1036/1116 test.anonymous union literal syntax... FAIL (TestUnexpectedResult)
```

All of these tests have float literals in them.

topolarity:
> However, it would likely be more efficient to limit use to f64 when parsing an f64, and limit use to f32 when parsing an f32, and I would like to see that improvement made eventually. That improvement could come as a follow-up to this PR, or it could be part of this one.

Another note for future optimization: A lot of the fastest float parsers seem to favor a "fast path" where the math can be done easily in smaller types, and then fall back to slower options only for extremely long/precise literals. This tries to give the "best of both worlds": exact precision/rounding in the difficult case, and fast performance in the common case

The fast path can usually be chosen based on the number of digits in the literal, and the decimal exponent.  [fast_double_parser](https://github.com/lemire/fast_double_parser/blob/master/include/fast_double_parser.h) takes this approach, for example

andrewrk:
I would like to repeat my suggestion to port musl's float parsing code (which has already been ported to self-contained .c code in parse_f128.c). If you want to port a different implementation instead (such as the one in this PR), can you provide some benchmark results to compare perf against it?

danielchasehooper:
@andrewrk 
> can you provide some benchmark results to compare perf against it?

Sure! Since I don't have a zig port of musl, I put both the sqlite and musl parsers in a .c file and had them parse `argv[1]` a million times. sqlite's implementation is 3 to 100 times faster than musl's, depending on the number being parsed. I believe sqlite is faster because it uses the FPU to construct the float, rather than doing a bunch of bit twiddling, but then again musl's implementation is greek to me. I included the source for my test below for anyone that would like to check my work. 

If you're satisfied with this, I'll get to making the tests pass and look into making it work in f64/f32 when f128 isn't requested.

```
~/Developer/float_parse_perf $ clang -O3 main.c
~/Developer/float_parse_perf $ ./a.out 3.141592654
sql 1.0271s
musl 2.59674s
~/Developer/float_parse_perf $ ./a.out 3141592654
sql 0.020399s
musl 1.78777s
~/Developer/float_parse_perf $ ./a.out +1.53e238
sql 1.88123s
musl 14.0651s
~/Developer/float_parse_perf $ ./a.out -1.53e-238
sql 1.99978s
musl 5.0289s
```
<details>
  <summary>main.c source</summary>
  
```
#include <stdint.h>
#include <stdio.h>
#include <math.h>
#include <float.h>
#include <limits.h>
#include <errno.h>
#include <ctype.h>
#include <stdbool.h>
#include <string.h>
#include <assert.h>
#include <time.h>
#include <stdlib.h>

#if LDBL_MANT_DIG == 53 && LDBL_MAX_EXP == 1024

#define LD_B1B_DIG 2
#define LD_B1B_MAX 9007199, 254740991
#define KMAX 128

#elif LDBL_MANT_DIG == 64 && LDBL_MAX_EXP == 16384

#define LD_B1B_DIG 3
#define LD_B1B_MAX 18, 446744073, 709551615
#define KMAX 2048

#elif LDBL_MANT_DIG == 113 && LDBL_MAX_EXP == 16384

#define LD_B1B_DIG 4
#define LD_B1B_MAX 10384593, 717069655, 257060992, 658440191
#define KMAX 2048

#else
#error Unsupported long double representation
#endif
#define MASK (KMAX-1)

typedef struct {
    const char *str;
    uint64_t off;
    uint64_t len;
} StringFile;

inline void shunget(StringFile* f) {
    f->off--;
    if (f->off < 0) f->off = 0;
}

inline int shgetc(StringFile* f) {
    if (f->off >= f->len) return 0;
    int res = f->str[f->off];
    f->off++;
    return res;
}

static long long scanexp(StringFile *f)
{
    int c;
    int x;
    long long y;
    int neg = 0;
    
    c = shgetc(f);
    if (c=='+' || c=='-') {
        neg = (c=='-');
        c = shgetc(f);
        if (c-'0'>=10U) shunget(f);
    }
    if (c-'0'>=10U) {
        shunget(f);
        return LLONG_MIN;
    }
    for (x=0; c-'0'<10U && x<INT_MAX/10; c = shgetc(f))
        x = 10*x + c-'0';
    for (y=x; c-'0'<10U && y<LLONG_MAX/100; c = shgetc(f))
        y = 10*y + c-'0';
    for (; c-'0'<10U; c = shgetc(f));
    shunget(f);
    return neg ? -y : y;
}

static long double decfloat(StringFile *f, int c, int bits, int emin, int sign)
{
    uint32_t x[KMAX];
    static const uint32_t th[] = { LD_B1B_MAX };
    int i, j, k, a, z;
    long long lrp=0, dc=0;
    long long e10=0;
    int lnz = 0;
    int gotdig = 0, gotrad = 0;
    int rp;
    int e2;
    int emax = -emin-bits+3;
    int denormal = 0;
    long double y;
    long double frac=0;
    long double bias=0;
    static const int p10s[] = { 10, 100, 1000, 10000,
        100000, 1000000, 10000000, 100000000 };

    j=0;
    k=0;


    /* Don't let leading zeros consume buffer space */
    for (; c=='0'; c = shgetc(f)) gotdig=1;
    if (c=='.') {
        gotrad = 1;
        for (c = shgetc(f); c=='0'; c = shgetc(f)) gotdig=1, lrp--;
    }

    x[0] = 0;
    for (; c-'0'<10U || c=='.'; c = shgetc(f)) {
        if (c == '.') {
            if (gotrad) break;
            gotrad = 1;
            lrp = dc;
        } else if (k < KMAX-3) {
            dc++;
            if (c!='0') lnz = dc;
            if (j) x[k] = x[k]*10 + c-'0';
            else x[k] = c-'0';
            if (++j==9) {
                k++;
                j=0;
            }
            gotdig=1;
        } else {
            dc++;
            if (c!='0') {
                lnz = (KMAX-4)*9;
                x[KMAX-4] |= 1;
            }
        }
    }
    if (!gotrad) lrp=dc;

    if (gotdig && (c|32)=='e') {
        e10 = scanexp(f);
        if (e10 == LLONG_MIN) {
            shunget(f);
            e10 = 0;
        }
        lrp += e10;
    } else if (c>=0) {
        shunget(f);
    }
    if (!gotdig) {
        errno = EINVAL;
        // shlim(f, 0);
        return 0;
    }

    /* Handle zero specially to avoid nasty special cases later */
    if (!x[0]) return sign * 0.0;

    /* Optimize small integers (w/no exponent) and over/under-flow */
    if (lrp==dc && dc<10 && (bits>30 || x[0]>>bits==0))
        return sign * (long double)x[0];
    if (lrp > -emin/2) {
        errno = ERANGE;
        return sign * LDBL_MAX * LDBL_MAX;
    }
    if (lrp < emin-2*LDBL_MANT_DIG) {
        errno = ERANGE;
        return sign * LDBL_MIN * LDBL_MIN;
    }

    /* Align incomplete final B1B digit */
    if (j) {
        for (; j<9; j++) x[k]*=10;
        k++;
        j=0;
    }

    a = 0;
    z = k;
    e2 = 0;
    rp = lrp;

    /* Optimize small to mid-size integers (even in exp. notation) */
    if (lnz<9 && lnz<=rp && rp < 18) {
        if (rp == 9) return sign * (long double)x[0];
        if (rp < 9) return sign * (long double)x[0] / p10s[8-rp];
        int bitlim = bits-3*(int)(rp-9);
        if (bitlim>30 || x[0]>>bitlim==0)
            return sign * (long double)x[0] * p10s[rp-10];
    }

    /* Drop trailing zeros */
    for (; !x[z-1]; z--);

    /* Align radix point to B1B digit boundary */
    if (rp % 9) {
        int rpm9 = rp>=0 ? rp%9 : rp%9+9;
        int p10 = p10s[8-rpm9];
        uint32_t carry = 0;
        for (k=a; k!=z; k++) {
            uint32_t tmp = x[k] % p10;
            x[k] = x[k]/p10 + carry;
            carry = 1000000000/p10 * tmp;
            if (k==a && !x[k]) {
                a = (a+1 & MASK);
                rp -= 9;
            }
        }
        if (carry) x[z++] = carry;
        rp += 9-rpm9;
    }

    /* Upscale until desired number of bits are left of radix point */
    while (rp < 9*LD_B1B_DIG || (rp == 9*LD_B1B_DIG && x[a]<th[0])) {
        uint32_t carry = 0;
        e2 -= 29;
        for (k=(z-1 & MASK); ; k=(k-1 & MASK)) {
            uint64_t tmp = ((uint64_t)x[k] << 29) + carry;
            if (tmp > 1000000000) {
                carry = tmp / 1000000000;
                x[k] = tmp % 1000000000;
            } else {
                carry = 0;
                x[k] = tmp;
            }
            if (k==(z-1 & MASK) && k!=a && !x[k]) z = k;
            if (k==a) break;
        }
        if (carry) {
            rp += 9;
            a = (a-1 & MASK);
            if (a == z) {
                z = (z-1 & MASK);
                x[z-1 & MASK] |= x[z];
            }
            x[a] = carry;
        }
    }

    /* Downscale until exactly number of bits are left of radix point */
    for (;;) {
        uint32_t carry = 0;
        int sh = 1;
        for (i=0; i<LD_B1B_DIG; i++) {
            k = (a+i & MASK);
            if (k == z || x[k] < th[i]) {
                i=LD_B1B_DIG;
                break;
            }
            if (x[a+i & MASK] > th[i]) break;
        }
        if (i==LD_B1B_DIG && rp==9*LD_B1B_DIG) break;
        /* FIXME: find a way to compute optimal sh */
        if (rp > 9+9*LD_B1B_DIG) sh = 9;
        e2 += sh;
        for (k=a; k!=z; k=(k+1 & MASK)) {
            uint32_t tmp = x[k] & (1<<sh)-1;
            x[k] = (x[k]>>sh) + carry;
            carry = (1000000000>>sh) * tmp;
            if (k==a && !x[k]) {
                a = (a+1 & MASK);
                i--;
                rp -= 9;
            }
        }
        if (carry) {
            if ((z+1 & MASK) != a) {
                x[z] = carry;
                z = (z+1 & MASK);
            } else x[z-1 & MASK] |= 1;
        }
    }

    /* Assemble desired bits into floating point variable */
    for (y=i=0; i<LD_B1B_DIG; i++) {
        if ((a+i & MASK)==z) x[(z=(z+1 & MASK))-1] = 0;
        y = 1000000000.0L * y + x[a+i & MASK];
    }

    y *= sign;

    /* Limit precision for denormal results */
    if (bits > LDBL_MANT_DIG+e2-emin) {
        bits = LDBL_MANT_DIG+e2-emin;
        if (bits<0) bits=0;
        denormal = 1;
    }

    /* Calculate bias term to force rounding, move out lower bits */
    if (bits < LDBL_MANT_DIG) {
        bias = copysignl(scalbn(1, 2*LDBL_MANT_DIG-bits-1), y);
        frac = fmodl(y, scalbn(1, LDBL_MANT_DIG-bits));
        y -= frac;
        y += bias;
    }

    /* Process tail of decimal input so it can affect rounding */
    if ((a+i & MASK) != z) {
        uint32_t t = x[a+i & MASK];
        if (t < 500000000 && (t || (a+i+1 & MASK) != z))
            frac += 0.25*sign;
        else if (t > 500000000)
            frac += 0.75*sign;
        else if (t == 500000000) {
            if ((a+i+1 & MASK) == z)
                frac += 0.5*sign;
            else
                frac += 0.75*sign;
        }
        if (LDBL_MANT_DIG-bits >= 2 && !fmodl(frac, 1))
            frac++;
    }

    y += frac;
    y -= bias;

    if ((e2+LDBL_MANT_DIG & INT_MAX) > emax-5) {
        if (fabsl(y) >= 2/LDBL_EPSILON) {
            if (denormal && bits==LDBL_MANT_DIG+e2-emin)
                denormal = 0;
            y *= 0.5;
            e2++;
        }
        if (e2+LDBL_MANT_DIG>emax || (denormal && frac))
            errno = ERANGE;
    }

    return scalbnl(y, e2);
}

long double __floatscan(StringFile *f, int prec)
{
    int sign = 1;
    size_t i;
    int bits;
    int emin;
    int c;

    switch (prec) {
    case 0:
        bits = FLT_MANT_DIG;
        emin = FLT_MIN_EXP-bits;
        break;
    case 1:
        bits = DBL_MANT_DIG;
        emin = DBL_MIN_EXP-bits;
        break;
    case 2:
        bits = LDBL_MANT_DIG;
        emin = LDBL_MIN_EXP-bits;
        break;
    default:
        return 0;
    }

    while (isspace((c=shgetc(f))));

    if (c=='+' || c=='-') {
        sign -= 2*(c=='-');
        c = shgetc(f);
    }

    for (i=0; i<8 && (c|32)=="infinity"[i]; i++)
        if (i<7) c = shgetc(f);
    if (i==3 || i==8 || (i>3 && true)) {
        if (i!=8) {
            shunget(f);
            if (true) for (; i>3; i--) shunget(f);
        }
        return sign * INFINITY;
    }
    if (!i) for (i=0; i<3 && (c|32)=="nan"[i]; i++)
        if (i<2) c = shgetc(f);
    if (i==3) {
        if (shgetc(f) != '(') {
            shunget(f);
            return NAN;
        }
        for (i=1; ; i++) {
            c = shgetc(f);
            if (c-'0'<10U || c-'A'<26U || c-'a'<26U || c=='_')
                continue;
            if (c==')') return NAN;
            shunget(f);
            while (i--) shunget(f);
            return NAN;
        }
        return NAN;
    }

    if (i) {
        shunget(f);
        errno = EINVAL;
        // shlim(f, 0);
        return 0;
    }

    if (c=='0') {
        c = shgetc(f);
        if ((c|32) == 'x') {
            // hex floats aren't supported
            return 0;
        }
        shunget(f);
        c = '0';
    }

    return decfloat(f, c, bits, emin, sign);
}

#define LARGEST_INT64 INT64_MAX
#define LONGDOUBLE_TYPE long double

/*
** Compute 10 to the E-th power.  Examples:  E==1 results in 10.
** E==2 results in 100.  E==50 results in 1.0e50.
**
** This routine only works for values of E between 1 and 341.
*/
static LONGDOUBLE_TYPE sqlite3Pow10(int E){
#if defined(_MSC_VER)
  static const LONGDOUBLE_TYPE x[] = {
    1.0e+001,
    1.0e+002,
    1.0e+004,
    1.0e+008,
    1.0e+016,
    1.0e+032,
    1.0e+064,
    1.0e+128,
    1.0e+256
  };
  LONGDOUBLE_TYPE r = 1.0;
  int i;
  assert( E>=0 && E<=307 );
  for(i=0; E!=0; i++, E >>=1){
    if( E & 1 ) r *= x[i];
  }
  return r;
#else
  LONGDOUBLE_TYPE x = 10.0;
  LONGDOUBLE_TYPE r = 1.0;
  while(1){
    if( E & 1 ) r *= x;
    E >>= 1;
    if( E==0 ) break;
    x *= x;
  }
  return r; 
#endif
}

int sqlite3AtoF(const char *z, double *pResult, int length){
  int incr = 1;
  const char *zEnd = z + length;
  /* sign * significand * (10 ^ (esign * exponent)) */
  int sign = 1;    /* sign of significand */
  int64_t s = 0;       /* significand */
  int d = 0;       /* adjust exponent for shifting decimal point */
  int esign = 1;   /* sign of exponent */
  int e = 0;       /* exponent */
  int eValid = 1;  /* True exponent is either not used or is well-formed */
  double result;
  int nDigits = 0;
  int nonNum = 0;  /* True if input contains UTF16 with high byte non-zero */

  *pResult = 0.0;   /* Default return value, in case of an error */

  /* skip leading spaces */
  while( z<zEnd && isspace(*z) ) z+=incr;
  if( z>=zEnd ) return 0;

  /* get sign of significand */
  if( *z=='-' ){
    sign = -1;
    z+=incr;
  }else if( *z=='+' ){
    z+=incr;
  }

  /* copy max significant digits to significand */
  while( z<zEnd && isdigit(*z) && s<((LARGEST_INT64-9)/10) ){
    s = s*10 + (*z - '0');
    z+=incr; nDigits++;
  }

  /* skip non-significant significand digits
  ** (increase exponent by d to shift decimal left) */
  while( z<zEnd && isdigit(*z) ){ z+=incr; nDigits++; d++; }
  if( z>=zEnd ) goto do_atof_calc;

  /* if decimal point is present */
  if( *z=='.' ){
    z+=incr;
    /* copy digits from after decimal to significand
    ** (decrease exponent by d to shift decimal right) */
    while( z<zEnd && isdigit(*z) ){
      if( s<((LARGEST_INT64-9)/10) ){
        s = s*10 + (*z - '0');
        d--;
      }
      z+=incr; nDigits++;
    }
  }
  if( z>=zEnd ) goto do_atof_calc;

  /* if exponent is present */
  if( *z=='e' || *z=='E' ){
    z+=incr;
    eValid = 0;

    /* This branch is needed to avoid a (harmless) buffer overread.  The 
    ** special comment alerts the mutation tester that the correct answer
    ** is obtained even if the branch is omitted */
    if( z>=zEnd ) goto do_atof_calc;              /*PREVENTS-HARMLESS-OVERREAD*/

    /* get sign of exponent */
    if( *z=='-' ){
      esign = -1;
      z+=incr;
    }else if( *z=='+' ){
      z+=incr;
    }
    /* copy digits to exponent */
    while( z<zEnd && isdigit(*z) ){
      e = e<10000 ? (e*10 + (*z - '0')) : 10000;
      z+=incr;
      eValid = 1;
    }
  }

  /* skip trailing spaces */
  while( z<zEnd && isspace(*z) ) z+=incr;

do_atof_calc:
  /* adjust exponent by d, and update sign */
  e = (e*esign) + d;
  if( e<0 ) {
    esign = -1;
    e *= -1;
  } else {
    esign = 1;
  }

  if( s==0 ) {
    /* In the IEEE 754 standard, zero is signed. */
    result = sign<0 ? -(double)0 : (double)0;
  } else {
    /* Attempt to reduce exponent.
    **
    ** Branches that are not required for the correct answer but which only
    ** help to obtain the correct answer faster are marked with special
    ** comments, as a hint to the mutation tester.
    */
    while( e>0 ){                                       /*OPTIMIZATION-IF-TRUE*/
      if( esign>0 ){
        if( s>=(LARGEST_INT64/10) ) break;             /*OPTIMIZATION-IF-FALSE*/
        s *= 10;
      }else{
        if( s%10!=0 ) break;                           /*OPTIMIZATION-IF-FALSE*/
        s /= 10;
      }
      e--;
    }

    /* adjust the sign of significand */
    s = sign<0 ? -s : s;

    if( e==0 ){                                         /*OPTIMIZATION-IF-TRUE*/
      result = (double)s;
    }else{
      /* attempt to handle extremely small/large numbers better */
      if( e>307 ){                                      /*OPTIMIZATION-IF-TRUE*/
        if( e<342 ){                                    /*OPTIMIZATION-IF-TRUE*/
          LONGDOUBLE_TYPE scale = sqlite3Pow10(e-308);
          if( esign<0 ){
            result = s / scale;
            result /= 1.0e+308;
          }else{
            result = s * scale;
            result *= 1.0e+308;
          }
        }else{ assert( e>=342 );
          if( esign<0 ){
            result = 0.0*s;
          }else{
#ifdef INFINITY
            result = INFINITY*s;
#else
            result = 1e308*1e308*s;  /* Infinity */
#endif
          }
        }
      }else{
        LONGDOUBLE_TYPE scale = sqlite3Pow10(e);
        if( esign<0 ){
          result = s / scale;
        }else{
          result = s * scale;
        }
      }
    }
  }

  /* store the result */
  *pResult = result;

  /* return true if number and no extra non-whitespace chracters after */
  return z==zEnd && nDigits>0 && eValid && nonNum==0;
}

int main(int argc, char *argv[]) {
    assert(argc==2);
    const char *num = argv[1];
    int len = strlen(num);

    const double ref_answer = strtod(num, NULL);

    const int iterations = 1000000;

    double musl_time, sql_time;
    {
        clock_t start = clock();

        for (int i=0; i < iterations; i++) {

            double result;
            sqlite3AtoF(num, &result, len);
            assert(ref_answer == result);
        }

        sql_time = ((double) (clock() - start)) / CLOCKS_PER_SEC;
    }
    
    {
        clock_t start = clock();

        StringFile f = {.str = num, .len = len};
        for (int i=0; i < iterations; i++) {
            f.off=0;
            double result = __floatscan(&f, 1);
            assert(ref_answer == result);
        }
        musl_time = ((double) (clock() - start)) / CLOCKS_PER_SEC;
    }

    printf("sql %gs\n", sql_time);
    printf("musl %gs\n", musl_time);
}
```
</details>

Test Notes
This source is the original f64 versions of the parsers, but I think sqlite will still be faster for f128 zig adaptations.
Performed on a 10 Core M1 Pro, (arm64, macos). Clang is producing a x86_64 binary - I expect performance implications hold true even in x86 rosetta mode.

andrewrk:
If it's faster and still correct that is a clear win. Nice!

andrewrk:
> x64 in rosetta mode

I'm taking this opportunity to make a general statement: I don't care about perf of x64 in rosetta mode. It's completely irrelevant to the near future of computing, in which Apple deprecates x86 entirely and we are left with only native ARM software on macOS.

Zig is young enough that we can aim a little bit for the future and not the present, and this is one of those times. arm64 perf matters. Native x86_64 perf matters. Rosetta-emulated x86_64 perf does not matter.

Same deal with effort spent on compatibility, QA, testing, etc. It's all a waste of time. Given that we have limited efforts to spend, we should spend our efforts on native x86_64 and native arm64 and forget that Rosetta exists.

andrewrk:
Closing stale PR. Feel free to open a new one if you decide to complete this.

PR #11255, opened 2022-03-21
stage2 CBE: Improve support for unions and error sets

topolarity:
Various fixes/improvements to the C backend to improve error/union support.

This also fixes up our handling of decls, where some decls were not correctly marked alive. Also had to tweak a few of the @ptrCast tests so that they avoid memory accesses that are well-defined in Zig but UB in C.

### DIFF
--- a/test/behavior/ptrcast.zig
+++ b/test/behavior/ptrcast.zig
@@ -39,7 +39,6 @@ fn testReinterpretWithOffsetAndNoWellDefinedLayout() !void {
 }
 
 test "reinterpret bytes inside auto-layout struct as integer with nonzero offset" {
-    if (builtin.zig_backend == .stage2_c) return error.SkipZigTest; // TODO
     if (builtin.zig_backend == .stage2_arm) return error.SkipZigTest; // TODO
     if (builtin.zig_backend == .stage2_aarch64) return error.SkipZigTest; // TODO
 
@@ -80,9 +79,9 @@ fn testReinterpretBytesAsExternStruct() !void {
     try expect(val == 5);
 }
 
-test "reinterpret bytes of an extern struct into another" {
+test "reinterpret bytes of an extern struct (with under-aligned fields) into another" {
     if (builtin.zig_backend == .stage1) return error.SkipZigTest;
-    if (builtin.zig_backend == .stage2_c) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_c) return error.SkipZigTest; // TODO: Under-aligned fields are not yet supported in the CBE
     if (builtin.zig_backend == .stage2_wasm) return error.SkipZigTest; // TODO
 
     try testReinterpretExternStructAsExternStruct();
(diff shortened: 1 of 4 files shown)


test/behavior/ptrcast.zig
```diff
@@ -91,14 +89,16 @@ test "reinterpret bytes of an extern struct into another" {
...
         c: u8,
     };
-    comptime var bytes align(2) = S1{ .a = 0, .b = 0, .c = 5 };
+    comptime var bytes: S1 = .{ .a = 0, .b = 0, .c = 5 };
 
     const S2 = extern struct {
-        a: u32 align(2),
+        a0: u32,
+        a1: u16,
+        a2: u16,
```
andrewrk, line 101: I don't see anything wrong with the previous version of this test case- it looks like valid Zig code to me. So this alteration of the behavior test is a regression. If you want to add additional coverage with this new case, that's fine. But if it's valid Zig code then it has to lower to valid C code in the C backend.

Same deal with the other altered behavior tests here.

andrewrk (changes requested):

test/behavior/ptrcast.zig
topolarity: The challenge here is that the CBE can't support some Zig features, specifically under-aligned fields.

Constructing the field pointer in the other test is also UB according to ubsan, since the parent pointer does not fit in the bounds of the pointee object. We can potentially work around that by using `offsetof` and pointer arithmetic.

Should we skip these tests for the CBE indefinitely, and add others that are within the range of defined behavior for C?

andrewrk: > The challenge here is that the CBE can't support some Zig features, specifically under-aligned fields.

That is incorrect, because C is turing complete. If we must lower some types as byte arrays then so be it.

There are also non-standard compiler extensions that we can take advantage of such as the [aligned attribute](https://gcc.gnu.org/onlinedocs/gcc-11.2.0/gcc/Common-Type-Attributes.html#Common-Type-Attributes).

> Constructing the field pointer in the other test is also UB according to ubsan, since the parent struct does not fit in the bounds of the pointee object.

The correct size of `S` is 6 bytes, the pointer cast is valid zig, and the C backend's job is to lower it, without invoking UB at the C language layer. I don't think a workaround with offsetof and pointer arithmetic is the way to go, instead, the C backend should lower the `S` type to something that is only 6 bytes.

> Should we skip these tests for the CBE indefinitely, and add others that are within the range of defined behavior for C?

No. We will bend C our will, and make them all pass.

topolarity: I see - A complete feature set is the hard requirement, and using canonical C features is simply best-effort :+1:

Is there any clear line about which compilers we are targeting and when we can use non-standard extensions?

FWIW, the field pointer problem shows up in a test like this:
```zig
const bytes align(4) = [_]u8{1, 2, 3, 4};
const S = struct {
    a: u16,
    b: u16,
    c: u32,
};

const x = @ptrCast(*const S, &bytes);
var p = &x.b; // triggers UB in the C output
_ = p;
```

I will add this test to the PR, so that we don't forget to support this in the CBE

andrewrk: > Is there any clear line about which compilers we are targeting and when we can use non-standard extensions?

The goal is to target C99, and any non-standard extensions are fair game as long as we also support the extensions not being available (via preprocessor checks). You can see some of this in zig.h.

test/behavior/ptrcast.zig
```diff
@@ -80,9 +79,34 @@ fn testReinterpretBytesAsExternStruct() !void {
...
+    };
+
+    const x = @ptrCast(*const S, &bytes);
+    var p = &x.b; // TODO: Triggers ubsan for CBE, since @sizeOf(S) > @sizeOf(bytes)
+
+    const expected = switch (native_endian) {
+        .Little => 0x0403,
+        .Big => 0x0304,
+    };
+    try expect(p.* == expected);
```
andrewrk, line 104: This is not a valid Zig code. The struct does not have a well-defined layout. What's the goal of adding this test case?

andrewrk (changes requested):

test/behavior/ptrcast.zig
topolarity: Should have been an extern struct, sorry about that!

andrewrk: With an extern struct, it's at best a grey area of the spec whether this will be allowed or not, for the same reason as it is not allowed in C.

topolarity: Would you prefer we leave it out for now, or a TODO to decide later whether this is spec'd or not?

andrewrk (approved):
Thanks for all the follow-ups :+1:

Issue #11091, opened 2022-03-08
doc/langref: replace `package-level` with `container level`


perillo:
The documentation uses the term `container level variable` in https://ziglang.org/documentation/master/#Container-Level-Variables, and in the grammar.

However in https://ziglang.org/documentation/master/#Top-Level-Doc-Comments, it uses the term `package-level comment`.

Vexu:
Feel free to make a pull request for a trivial change like this, the langref is generated from https://github.com/ziglang/zig/blob/master/doc/langref.html.in using the command `zig build docs`.

perillo:
@Vexu Thanks.
I have tried to read the CONTRIBUTING.md file but it was not very useful.  There is no information about the conventions to use for a PR, like the format of the commit header (the first line) or if the commit message should include the associated issue number.

I have experience in contributing to the Go project but I have never used the Github Flow.

Vexu:
> I have tried to read the CONTRIBUTING.md file but it was not very useful. There is no information about the conventions to use for a PR, like the format of the commit header (the first line) or if the commit message should include the associated issue number.

We're not too pedantic about details like that but generally the convention is that small changes do not require an issue beforehand, commit messages don't need to follow a format but more detailed is better, and if the PR addresses an issue link it using a [closing keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

I'll try to document this in CONTRIBUTING.md to make it more clear in the future.

dxps:
I'm a newbie in Zig, but pretty excited about it and very keen on contributing. :blush: 

Related to `langref.html.in` file, I have a first question about formatting. Let me present my case.

Since this has an HTML format, I am tempted to use it accordingly:

1. Inform VSCode (my editor of choice, by default) about it.
   - Instead of "Plain Text", selecting "HTML" I get a nice syntax highlighting.
   - I would also be able to use a proper formatting.
   - I'm using the default formatter
     ```json
     "[html]": {
        "editor.defaultFormatter": "vscode.html-language-features"
     },
     ```
     but, of course, if I'd do that pretty much the whole file gets reformatted.

2. Keep treating it as "Plain Text" is doable (not preferred, ofc), but I won't be able to collapse some sections.
   - For example, the `<style` section gets partially collapsed since there is a wrong indentation at line 55 (and everything below within the section).

[...]

PR #11256, opened 2022-03-21
chore: doc fixes, closes 11091

dxps:
It includes:
- `<style>` content indentation fix, so that content collapsing works (at least in VSCode) regardless of the selected file format.
- small rephrasing as described in #11091.

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -52,255 +52,255 @@
         #contents-wrapper {
           margin-left: calc(var(--nav-width) + var(--nav-margin-l));
         }
-    }
-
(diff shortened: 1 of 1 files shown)


dxps:
:information_source:  As an idea for future improvements:

Maybe I'm too picky or my OCD for consistency kicks in again, but probably we should use dashes everywhere there are mentions about `{scope}-level` references, whether container, local, or top.

The reason seems to be:
> It is hyphenated because it is an adjective.

dxps:
:partying_face:  Yay! My first and minimal contribution!

PR #11257, opened 2022-03-21
stage2: wasm - Implement popcount instruction

Luukdegram:
Implements the `@popCount` builtin instruction which means `std.debug.print` now works in the wasm backend :)


andrewrk (approved):
Nice!

PR #11258, opened 2022-03-22
CBE: enable more passing tests

jmc-88:


topolarity:
Uh oh, I think this overlaps with #11255

I should have coordinated with you better, since I know you've been active in the C backend - Sorry about that!

jmc-88:
> Uh oh, I think this overlaps with #11255
> 
> I should have coordinated with you better, since I know you've been active in the C backend - Sorry about that!

Ah no bother, I see your PR is more complete than mine so let's keep that one -- I rescoped this to just enabling some tests that pass on `master`.

PR #11264, opened 2022-03-22
Fix ucontext_t

locriacyber:
`link` might be null


ifreund (approved):
Thanks for the fix, LGTM. Just waiting for the CI to pass.

PR #11269, opened 2022-03-22
stage2: runtime safety checks for slicing

mitchellh:
I don't know if there is a way to add automated tests for these, I just manually tested comparing stage1 to stage2.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -19715,6 +19715,32 @@ fn analyzeSlice(
     });
 
     try sema.requireRuntimeBlock(block, src);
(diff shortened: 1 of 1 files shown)


andrewrk:
We have automated tests that run with `zig build test` (or to only run the one step, `zig build test-runtime-safety`) and that will provide coverage for this eventually - we just need to get `zig build` working with stage2 :-)

andrewrk:
Oops, forgot to clean up the merge conflicts on this one before merging. I'll do that in a follow-up commit to master right now.

Issue #863, opened 2018-03-27
Cast from slices to array pointers


Hejsil:
Array pointers are slices, but with a length known at compile time. This is powerful because it allows many of Zig's compile time constructs to work on them such as comptime bound check for constants ~~and `inline for`~~ (`inline for` can't unroll runtime arrays. We can use `inline while` to do it though).

I propose making it safer and easier to convert from slices to array pointers so that it encourages people to utilize the static guarantees that array pointers provide. 
```
const s : []const u8 = "ABCD";

// Currently, we can do:
comptime assert(@typeOf(@ptrCast(&const [1]u8, s[1..2].ptr)) == &const [1]u8);

// Option 1: Slicing with constant values should return an array pointer
comptime assert(@typeOf(s[1..2]) == &const [1]u8);

// Option 2: Explicit cast from slice to array pointer. The cast will assert that the slices length is >= the array
comptime assert(@typeOf((&const [1]u8)(s[1..])) == &const [1]u8);
```

The current method is no good. `ptrCast` is a scary builtin, and leaving out the `2` in `s[1..2]` will bypass the safety check that ensures that our slice is at least the size of the array we are casting too.

Option 1 has some upsides. It will seamlessly flow around array pointers, giving you all their benefits without you even trying. Plus, array pointers should implicitly cast to slices (currently, only `&const [1]u8` -> `[]const u8` works, not `&[1]u8` -> `[]u8`), so this shouldn't break much code. This does have the downside that changing a `const` declaration to `var` might break code that relies on these implicit array pointers.

Option 2 favors the programmer being more explicit about when they want the static guaranty of the array pointer. This allows a more clearly expressed intent, and code your code will only break in predictable ways when `const`s are changed to `var` (where the cast takes place).

Note, that indexing array pointers right now is a pain, as `[]` works on both pointers and arrays. #770 Fixes this.

andrewrk:
I like Option 1. Let's implement that and then see if we need to introduce anything else to make slicing seamless.

andrewrk:
Note after implementing this, audit all the calls to `readIntSlice*` and `writeIntSlice*` and make them call the array variants instead.

Hejsil:
Here is a fun thing this proposals will allow

```zig
a[1..4].* = b[2..5].*;
```

thejoshwolfe:
not sure if this was said already, but you can get a comptime len slice at a runtime offset with `a[index..][0..4]`.

daurnimator:
[...]

PR #11270, opened 2022-03-23
stage2: concat/mult of slices should yield a ptr to an array

topolarity:
hundo p for `test/behavior/slice.zig`


topolarity:
Before fixing, this test failure showed a src loc in the wrong file (thanks to @mitchellh for the heads up):
```bash
/home/topolarity/repos/zig/stage2/lib/zig/std/mem.zig:548:45: error: expected []const u8, found [8]u8
    try testing.expect(lessThan(u8, "abcd", "bee"));
```
I plan to try to fix that up in another PR soon

andrewrk:
I'd like to re-evaluate this as part of the language spec - this test precedes the changes from #863 which made slices with comptime-known start and end indexes return single-pointers to arrays instead of slices. In light of this I think the test cases should perhaps be changed to have `++` and `**` always return single-pointers to arrays.

topolarity:
Works for me! Updated with the new behavior :+1:

Issue #11157, opened 2022-03-14
stage2: Sema panic 


topolarity:
### Zig Version

0.10.0-dev.1309+5919b1004

### Steps to Reproduce

```zig
test {
    const S = extern struct {
        a: f32,
        b: [4] u8,
        c: f32,
    };

    comptime {
        const x = S{ 
            .a = 1.5,
            .b = [_]u8 { 1, 2, 3, 4 },
            .c = 2.6,
        };
        const ptr = &x.b[0];
        _ = @ptrCast([*c]const u8, ptr)[5]; // fails
    }
}
```

### Expected Behavior

Test should fail with a compile error or succeed, reinterpreting comptime memory according to the defined memory layout.

### Actual Behavior

```zig
Analyzing test.zig: test.zig:test_0
      %2 = dbg_stmt(2, 5)
      %3 = extended(struct_decl(known_non_opv, anon, Extern, {}, {%4..%6}, {
        a: @Ref.f32_type,
        b: %5,
        c: @Ref.f32_type,
      }) node_offset:2:22
      %7 = dbg_var_val(%3, "S")
      %8 = dbg_stmt(8, 5)
      %9 = alloc_inferred_comptime() node_offset:9:9
      %10 = validate_struct_init_ty(%3) node_offset:9:20
      %11 = coerce_result_ptr(%3, %9)
      %12 = field_ptr(%11, "a") node_offset:10:18
      %13 = float(1.5)
      %14 = store_node(%12, %13) node_offset:10:18
      %15 = field_ptr(%11, "b") node_offset:11:24
      %16 = int(4)
      %17 = array_type(%16, @Ref.u8_type)
      %18 = coerce_result_ptr(%17, %15)
      %19 = elem_ptr_imm(%18, 0) node_offset:11:26
      %20 = store_node(%19, @Ref.one) node_offset:11:26
      %21 = elem_ptr_imm(%18, 1) node_offset:11:29
      %22 = int(2)
      %23 = store_node(%21, %22) node_offset:11:29
      %24 = elem_ptr_imm(%18, 2) node_offset:11:32
      %25 = int(3)
      %26 = store_node(%24, %25) node_offset:11:32
      %27 = elem_ptr_imm(%18, 3) node_offset:11:35
      %28 = int(4)
      %29 = store_node(%27, %28) node_offset:11:35
      %30 = validate_array_init_comptime({
        %19 = elem_ptr_imm(%18, 0) node_offset:11:26
[...]

PR #11273, opened 2022-03-23
stage2: Properly "flatten" elem_ptrs before deref

### LINKED ISSUE #11272: failing stage2 test: comptime pointer load through elem_ptr
Zig Version: `0.10.0-dev.1533+8326ab7ad`

```zig
const std = @import("std");

test "comptime pointer load through elem_ptr" {
    const S = struct {
        x: usize,
    };

    comptime {
        var array: [10]S = undefined;
        for (array) |*elem, i| {
            elem.* = .{
                .x = i,
            };
        }
        var ptr = @ptrCast([*]S, &array);
        var x = ptr[0].x;
        assert(x == 0);
        ptr += 1;
        assert(ptr[1].x == 2);
    }
}
```

Expected to pass. Instead:

```
$ ./zig-out/bin/zig test test.zig 
/home/andy/dev/zig/lib/std/debug.zig:234:14: error: unable to resolve comptime value
             ^
test.zig:20:15: note: called from here
        assert(ptr[1].x == 2);
              ^
```

This is the next issue blocking std.zig.tokenizer tests from passing:

```
$ ./zig-out/bin/zig test ../lib/std/zig/tokenizer.zig 
/home/andy/dev/zig/lib/std/sort.zig:1147:59: error: comptime dereference requires comptime_string_map.ComptimeStringMap__struct_1284 to have a well-defined layout, but it does not.
    comptime lessThan: fn (@TypeOf(context), lhs: T, rhs: T) bool,
                                                          ^
/home/andy/dev/zig/lib/std/sort.zig:273:21: note: called from here
                swap(T, sliced_items, context, lessThan, &order, 1, 2);
                    ^
/home/andy/dev/zig/lib/std/comptime_string_map.zig:32:22: note: called from here
        std.sort.sort(KV, &sorted_kvs, {}, lenAsc);
                     ^
../lib/std/zig/tokenizer.zig:13:47: note: called from here
    pub const keywords = std.ComptimeStringMap(Tag, .{
                                              ^
```

cc @topolarity - the issue appears to be in the elem_ptr handler of `beginComptimePtrLoad`.

topolarity:
Resolves #11272 - `tokenizer` is 100% ✅ 

Sema.pointerDeref() assumes that elem_ptrs have been "flattened" when they were created. That is, an elem_ptr should never be the array_ptr of another elem_ptr if they share the same types.

Value.elemPtr does this already, but a couple of places in Sema were bypassing this logic.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -17045,17 +17045,9 @@ fn elemVal(
                     const indexable_val = maybe_indexable_val orelse break :rs indexable_src;
                     const index_val = maybe_index_val orelse break :rs elem_index_src;
                     const index = @intCast(usize, index_val.toUnsignedInt(target));
-                    const elem_ty = indexable_ty.elemType2();
-
-                    var payload: Value.Payload.ElemPtr = .{ .data = .{
-                        .array_ptr = indexable_val,
-                        .elem_ty = elem_ty,
-                        .index = index,
-                    } };
-                    const elem_ptr_val = Value.initPayload(&payload.base);
-
+                    const elem_ptr_val = try indexable_val.elemPtr(indexable_ty, sema.arena, index, target);
                     if (try sema.pointerDeref(block, indexable_src, elem_ptr_val, indexable_ty)) |elem_val| {
-                        return sema.addConstant(elem_ty, elem_val);
+                        return sema.addConstant(indexable_ty.elemType2(), elem_val);
                     }
                     break :rs indexable_src;
                 };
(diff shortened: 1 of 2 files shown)


topolarity:
Turns out this bug was also masking #11157, making that issue look resolved when it was not.

andrewrk:
Changes look good - the only thing missing is to add the test case from #11272 as a new behavior test. I will do that now upon merging.

topolarity:
Thanks for lending a hand! Noted for next time :+1:

PR #11280, opened 2022-03-23
stage2: wasm - `@errorName` and more

Luukdegram:
Implements the following instructions:
- error_name (runtime `@errorName`)
- lowering opt_payload_ptr for pointers
- int_to_float
- ptr_slice_ptr_ptr
- ptr_slice_len_ptr
- unwrap_errunion_err_ptr
- unwrap_errunion_payload_ptr
 
This also fixes switching on errors.

`@errorName` was really fun to implement and works by creating 2 symbols and corresponding atoms. The initial symbol contains a table in which each element consists of a slice where the `ptr` field points towards the error name and the `len` field contain the error name length without the sentinel.

The secondary symbol contains a list of all error names from the global error set.

During the error_name instruction, we first get a pointer to the first symbol.
Then based on the operand we perform pointer arithmetic, to get the correct index into this table.
e.g. error index 2 -> result = ptr + (2 * slice size). The result of this will be stored in a local and then returned as an instruction result.

During `flush()` we populate the error names table by looping over the global error set and creating a relocation for each error name. This relocation is appended to the table symbol. Then finally, this name is written to the names list itself.

Finally, both symbols' atoms are allocated within the rest of the binary. When no error name is referenced, the `error_name_symbol` is never set, and therefore no error name table will be emitted into the final binary.


andrewrk (approved):
Great work :grin:

PR #11284, opened 2022-03-24
Add primitive file support for compile error tests

topolarity:
This PR implements a rudimentary version of independent file support for compile error tests. Since these new `.zig` files would trip the formatter check, this also adds an `--exclude` parameter to "zig fmt"

Here's an example:
```zig
const U = union(enum(u2)) {
    A: u8,
    B: u8,
    C: u8,
    D: u8,
    E: u8,
};
export fn entry() void {
    _ = U{ .E = 1 };
}

// union with too small explicit unsigned tag type
//
// tmp.zig:1:22: error: specified integer tag type cannot represent every field
// tmp.zig:1:22: note: type u2 cannot fit values in range 0...4
```

The format is extremely basic. In a comment block at the end of the file:
1. Comment line 1 is the case name
2. Comment line 2 is a blank line
3. Comment line 3+ are the expected errors.

The mode of the test (obj/exe/test), as well as the target (stage1/stage2) is determined based on the directory containing the test file.

7f64f7c and 1de63ad contain the core logic changes. 0568b45 just moves the tests over


topolarity:
To make sure I didn't drop any tests in the process, I logged all obj/test/exe test cases from a run of "zig build test" and compared before/after this change - Everything matches, accounting for a few minor renames and one deleted duplicate test. You can find the [full diff here](https://gist.github.com/topolarity/8968f6456280c325e1202b31af85fa4f).

topolarity:
cc @mitchellh - I know you were looking at a more complete set of improvements in this area.

Hopefully this basic support gives you a good base to work from. I tried to keep the file format as simple as possible, so that these can be easily updated by script if we choose to update it later for more test magic 🪄

ci/zinc/linux_test.sh
```diff
@@ -45,7 +45,7 @@ cd $WORKSPACE
 
 # Look for non-conforming code formatting.
 # Formatting errors can be fixed by running `zig fmt` on the files printed here.
-$ZIG fmt --check .
+$ZIG fmt --check . --exclude test/compile_errors/
```
matu3ba, line 48: Why is the formatting check disabled for test cases in the CI? If there problems/bugs, then they should be linked here.

test/compile_errors.zig
```diff
@@ -2,181 +2,48 @@ const std = @import("std");
...
 
-    ctx.objErrStage1("@Type() union payload is undefined",
-        \\const Foo = @Type(.{
-        \\    .Struct = undefined,
-        \\});
-        \\comptime { _ = Foo; }
-    , &[_][]const u8{
-        "tmp.zig:1:20: error: use of undefined value here causes undefined behavior",
-    });
+            var exe_dir = try stage1_dir.openDir("exe", .{ .iterate = true });
```
matu3ba, line 36: add `.no_follow = true`

test/compile_errors.zig
```diff
@@ -2,181 +2,48 @@ const std = @import("std");
...
-        \\    D: u8,
-        \\};
-        \\export fn entry() void {
-        \\    _ = U{ .D = 1 };
-        \\}
-    , &[_][]const u8{
-        "tmp.zig:1:22: error: specified integer tag type cannot represent every field",
-        "tmp.zig:1:22: note: type i2 cannot fit values in range 0...3",
-    });
+            var test_dir = try stage1_dir.openDir("test", .{ .iterate = true });
```
matu3ba, line 41: add `.no_follow = true`

test/compile_errors.zig
```diff
@@ -2,181 +2,48 @@ const std = @import("std");
...
-        \\const e: E = .two;
-        \\comptime {
-        \\    @export(e, .{ .name = "e" });
-        \\}
-    , &.{
-        "tmp.zig:3:13: error: exported enum without explicit integer tag type",
-        "tmp.zig:7:13: error: exported enum value without explicit integer tag type",
-    });
+pub fn addCases(ctx: *TestContext) !void {
+    var parent_dir = try std.fs.cwd().openDir(std.fs.path.dirname(@src().file).?, .{});
```
matu3ba, line 14: add `no_follow=true`

test/compile_errors.zig
```diff
@@ -2,181 +2,48 @@ const std = @import("std");
...
+    defer parent_dir.close();
 
-    ctx.objErrStage1("issue #9346: return outside of function scope",
-        \\pub const empty = return 1;
-    , &.{"tmp.zig:1:19: error: 'return' outside function scope"});
+    var compile_errors_dir = try parent_dir.openDir("compile_errors", .{});
```
matu3ba, line 17: add `no_follow=true`

matu3ba (commented):
Quickly glimpsing over it, it looks good.
Though the browser review on github is annoyingly slow.

1. You changed `test/behavior.zig` in commit https://github.com/ziglang/zig/commit/f266a9df9d5447d93add917064d1f91d5b6142e2 with
```diff
-    _ = @import("behavior/translate_c_macros.zig");
+    //_ = @import("behavior/translate_c_macros.zig");
```
For some reason this change does not show up in the pushed commits. Did you rebase?
2. Question on CI
3. Forbidding symlinks sounds like a good thing to me. I have no idea, why the default for `openDir` is to follow symlinks (which have nonportable behavior).

mitchellh (commented):
Nice! At a high level, I'd say that this is a great improvement and anything I propose we can just continue to extend this work. 

It might require some churn on the comment formats, because one thing I'm trying to do is unify a handful of similar compiler tests into this directory structure: compiler errors, runtime safety, execute and compare output, assemble and link, etc. So, we may need to mechanically go through and update the comments to have say... `//! type=errors` (NOT THE REAL SYNTAX) or modify the comment structure itself. But like you said, I think we can script that...

ci/zinc/linux_test.sh
topolarity: This is because the tested compile errors are sometimes parsing errors. Some of these test files fail when run with zig fmt

Additionally, the existing expected error messages reference the line/column numbers of the sources. Formatting them would require a lot of work to update the expected error messages

topolarity:
> For some reason this change does not show up in the pushed commits. Did you rebase?

Yes, sorry about that - please disregard that change.

Thanks for the review. I'll update the directory handling later today

andrewrk:
> Thanks for the review. I'll update the directory handling later today

Let me know when it's ready for review

topolarity:
Sorry about the delay - I've updated the manifest format to better align with #11288.

PR is ready for review :+1:

andrewrk:
Huge quality of life improvement, implemented in a maintainable, straightforward manner. Love it!

Issue #3822, opened 2019-12-01
Add `readLine` and `readLineSlice` as methods to `InStream`


jayschwa:
`std.io.readLine` and friends are top-level functions that read from the stdin stream or passed-in stream. `InStream` already has a wide variety of methods on it, and it seems inconsistent that `readLine` would be top-level functions, while all the other `readFoo`-style functions are methods on `InStream`.

If implemented, the top-level `readLine` functions could become wrappers for the new methods, and eventually be deprecated and removed.

One of the nice things that `readLine` does is DRY the `getStdIn().inStream().stream` ceremony. Proposal #3823 addresses that.

MasterQ32:
`readLine` ist just a special case of `InStream.readUntilDelimiterAlloc` or `InStream.readUntilDelimiterBuffer`

See:
https://github.com/ziglang/zig/blob/4b6740e19d57454f3c4eac0c2e9a92ce08e7ec04/lib/std/io/in_stream.zig#L125
https://github.com/ziglang/zig/blob/4b6740e19d57454f3c4eac0c2e9a92ce08e7ec04/lib/std/io/in_stream.zig#L103

I think it's better this way as with a `readLine`-Function, the zig std will assume a defined "newline sequence" which may not be the same on all environments

jayschwa:
> `readLine` ist just a special case of `InStream.readUntilDelimiterAlloc` or `InStream.readUntilDelimiterBuffer`

Agreed, but why does it make sense for `readLine` to be standalone and for `readUntilDelimeter` to be a method? `readUntilDelimeter` and many other `readFoo`-style methods could just as easily be standalone functions like `readLine`. What is the distinction?

> the zig std will assume a defined "newline sequence" which may not be the same on all environments

I don't follow. How is the assumed newline sequence a point in favor of standalone function versus method?

andrewrk:
I see why this was problematic now, `but std.io.readLine` is named after [GNU readline](https://en.wikipedia.org/wiki/GNU_Readline) and is intended for the use case of an interactive command line, rather than streams.

daurnimator:
FWIW this will all get much cleaner if the stream mixin concept is integrated. Currently discussion of Mixins is going on in #3785

jayschwa:
I am closing this issue because it is no longer relevant. Standalone `std.io.readLine` no longer exists.

andrewrk:
[...]

PR #11287, opened 2022-03-24
Adds ability to iterate overs lines in a reader:

amirrezaask:
This PR will add LinesIterator struct and a method on `reader` called `readLinesAlloc` that will get an Allocator and returns an iterator that for each `.next()` call will alocate enough memory, check with `max_size` and read the line into alocated space and return that line to the user.
```zig
for (reader.readLinesAlloc()) |line| {
     // do smth with the line
}
```

### DIFF
--- a/lib/std/io/reader.zig
+++ b/lib/std/io/reader.zig
@@ -175,9 +175,12 @@ pub fn Reader(
                 return self.reader.readUntilDelimiterAlloc(self.allocator, '\n', self.max_size);
             }
         };
-        // Creates a LinesIterator using given allocator.
-        pub fn readLinesAlloc(self: Self, allocator: mem.Allocator) LinesIterator {
-            return LinesIterator.init(allocator, self, std.math.maxInt(usize));
+        /// Creates a new `LinesIterator` using given allocator. LinesIterator stores a reference
+        /// to the `io.Reader` and also the allocator, for each call to `next` method on Iterator 
+        /// it will allocate enough memory to read a line and if memory space for given line is 
+        /// larger than `max_size` it will return `error.StreamTooLong`.
+        pub fn readLinesAlloc(self: Self, allocator: mem.Allocator, max_size: usize) LinesIterator {
+            return LinesIterator.init(allocator, self, max_size);
         }
 
         /// Allocates enough memory to read until `delimiter` or end-of-stream.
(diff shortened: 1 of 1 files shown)


matu3ba:
What use case do you have for this change?

On file operations one has typically 3 operations: 1. read whole file at once from the meta-information of the file system, 2. predetermine maximum length in file + error out, 3. predetermined length with more of an exceptional code path using new/bigger allocations for the content.

jayschwa:
Related: #3822

amirrezaask:
> What use case do you have for this change?
> 
> On file operations one has typically 3 operations: 1. read whole file at once from the meta-information of the file system, 2. predetermine maximum length in file + error out, 3. predetermined length with more of an exceptional code path using new/bigger allocations for the content.

It's more of a high level kind of api as we see in high level programming languages to read any kind of reader imagine a file by lines and loop over lines easily

ominitay:
What does this achieve which isn't already done by `readUntilDelimiterAlloc`?

amirrezaask:
> What does this achieve which isn't already done by `readUntilDelimiterAlloc`?

This is a simpler interface when you want whole file lines and iterate over them, all other languages have some way of doing so, it's simpler and more readable than doing while(true) and read until for example EOF.

ominitay:
So, you're saying that this:
```zig
var iter = reader.readLinesAlloc(a, std.math.maxInt(usize));
while (iter.next()) |line| {
    ...
}
```
Is simpler than this?
```zig
while (reader.readUntilDelimiterAlloc(a, '\n', std.math.maxInt(usize))) |line| {
    ...
}
```
I don't really see the improvement.

ominitay:
Wait I've just blindly followed your code and trusted it, but this code is incorrect as it stands...

ominitay:
You don't ever return null -- you return an empty slice when you reach the end. So anyone trying to use this in a while loop will just have an infinite loop

amirrezaask:
Sorry, I had a problem on my setup I thought of the sample you provided but for some reason i was getting compile error, i am new to zig so I thought it wasn't possible, i will close this pr sry for the inconvenience.

ominitay:
No problem :) If you need any help, consider joining a Zig community, such as the Zig Discord, where you can get help!

amirrezaask:
> No problem :) If you need any help, consider joining a Zig community, such as the Zig Discord!

Already did but sometimes things seems so easy that you are embarrassed to ask questions

jayschwa:
> What does this achieve which isn't already done by `readUntilDelimiterAlloc`?

While the impl in this PR didn't handle it, one advantage for a distinct line-reading function is that the user doesn't have to think about CRLF vs LF. I'm not sure the state of CRLF in Windows these days, but it's still common in Internet protocols. [Go](https://pkg.go.dev/bufio#Reader.ReadLine) and [Rust](https://doc.rust-lang.org/std/io/trait.BufRead.html#method.lines) handle both in their buffered readers, for example.

PR #11289, opened 2022-03-24
stage2: implement `@select`

schmee:
@andrewrk I noticed that `airShuffle` does a bit of coercion to allow arrays as input, does that apply here as well?

### DIFF
--- a/src/Air.zig
+++ b/src/Air.zig
@@ -554,7 +554,7 @@ pub const Inst = struct {
         /// Uses the `ty_pl` field with payload `Shuffle`.
         shuffle,
         /// Constructs a vector element-wise from `a` or `b` based on `pred`.
-        /// Uses the `ty_pl` field with payload `Select`.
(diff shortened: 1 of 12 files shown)


src/Air.zig
```diff
@@ -553,6 +553,9 @@ pub const Inst = struct {
         /// Constructs a vector by selecting elements from `a` and `b` based on `mask`.
         /// Uses the `ty_pl` field with payload `Shuffle`.
         shuffle,
+        /// Constructs a vector element-wise from `a` or `b` based on `pred`.
+        /// Uses the `ty_pl` field with payload `Select`.
```
schmee, line 557: @andrewrk should this be a `pl_op` instead? 🤔 The type is not used for the LLVM backend, but I don't know if it might be needed for the other backends.

schmee:
@andrewrk updated 👍

andrewrk (approved):
Squeaky clean :sparkles:

Issue #11297, opened 2022-03-25
stage2: attempt to use null value when usingnamespace of result of function call


Vexu:
```zig
const Completion = struct {
    const condition = true;
    fn exportWhen() type {
        return if (condition) struct {} else struct {};
    }
    usingnamespace exportWhen();
};
test {
    _ = Completion.exportWhen;
}
```
This is blocking stage2 being able to do `zig build --help`.

Issue #11293, opened 2022-03-25
callsites lack result location types, causing unwanted compile errors when argument expressions make a runtime branch


andrewrk:
```zig
test {
    foo(if (runtime) .a else .b);
}
var runtime = true;
const E = enum { a, b };
fn foo(e: E) void {
    _ = e;
}
```

Expected:

```
$ ./zig test test3.zig 
All 1 tests passed.
```

Actual:

```
$ ./zig-out/bin/zig test test3.zig 
test3.zig:2:9: error: value with comptime only type '@TypeOf(.enum_literal)' depends on runtime control flow
    foo(if (runtime) .a else .b);
        ^
test3.zig:2:13: note: runtime control flow here
    foo(if (runtime) .a else .b);
            ^
```

This did work before 1cc5d4e758a95be373756e7c32f9bb46d21633c9 which added method call syntax and in the process of it removed the result location type from function call arguments:

```diff
     for (call.ast.params) |param_node, i| {
-        const param_type = try gz.add(.{
-            .tag = .param_type,
-            .data = .{ .param_type = .{
-                .callee = lhs,
-                .param_index = @intCast(u32, i),
-            } },
-        });
-        args[i] = try expr(gz, scope, .{ .coerced_ty = param_type }, param_node);
+        // Parameters are always temporary values, they have no
+        // meaningful result location.  Sema will coerce them.
+        args[i] = try expr(gz, scope, .none, param_node);
     }
```

 @SpexGuy did you have a suggestion on how to make this work? Seems like we might need to restructure it so that it builds function calls one piece at a time.

This is the next bug to get the self-hosted parser working.

SpexGuy:
My crazy idea was to insert special instructions at call sites and in parameter type evaluation that cause sema to ping-pong back and forth between the two contexts, evaluating a type in the signature and then a value at the call site.  Something like `%type = next_param_type %curr_param_value` and `%value = yield_param_type %type`.

matu3ba:
@SpexGuy user andyjansson on twitch pointed to type system resolving, so I checked and found "Complete and Easy Bidirectional Typechecking for Higher-Rank Polymorphism" by Jana Dunfield and Neelakantan R. Krishnaswami
which explains tradeoffs in 9. Related Work and Discussion (whole thing is 91 pages).

This paper could be related as it claims to simplify annotations https://i.cs.hku.hk/~bruno/papers/arguments-esop18.pdf

[...]

PR #11299, opened 2022-03-25
stage2 fixes on the way to `zig2 build --help` working

Vexu:
`zig2 build --help` is still blocked by (at least) #11297, #11293, #11306 and `TODO: get max errors in compilation` but this gets us a little closer.

### DIFF
--- a/test/behavior/switch.zig
+++ b/test/behavior/switch.zig
@@ -465,7 +465,10 @@ test "else prong of switch on error set excludes other cases" {
 }
 
 test "switch prongs with error set cases make a new error set type for capture value" {
-    if (builtin.zig_backend != .stage1) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_x86_64) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_arm) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_aarch64) return error.SkipZigTest; // TODO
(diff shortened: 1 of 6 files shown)


test/behavior/switch.zig
```diff
@@ -652,3 +656,8 @@ test "switch capture copies its payload" {
     try S.doTheTest();
     comptime try S.doTheTest();
 }
+
+test "switch on empty error set" {
+    var a: error{} = undefined;
+    switch (a) {}
```
andrewrk, line 662: Pretty sure this is Illegal Behavior because according to the proposed spec, even 0-bit types can theoretically hold the value `undefined` and branching on `undefined` is Illegal Behavior.

An empty error set is effectively the same thing as an empty enum, which is effectively the same thing as `noreturn`. Switching on an empty error set is equivalent to `unreachable`.

Alternate proposed behavior test:

```zig
test {
    if (foo()) |_| {
        // OK
    } else |err| switch (err) {}
}
fn foo() !void {}
```

andrewrk (changes requested):

test/behavior/switch.zig
Vexu: I did realize this after seeing the test failure (I didn't run the behavior tests on stage1 when developing) and realized this was actually the issue I was seeing. I tried to also apply the fix in #11126 for inferred error sets but ran into some issues doing so.

I'm guessing an empty switch should unconditionally be a compile error (if semantically analyzed)?

andrewrk (approved):

PR #11300, opened 2022-03-25
dwarf: add debug info for error sets and error unions

kubkon:
image

### DIFF
--- a/src/link/Dwarf.zig
+++ b/src/link/Dwarf.zig
@@ -468,6 +468,7 @@ pub fn commitDeclDebugInfo(
             value_ptr.off = @intCast(u32, dbg_info_buffer.items.len);
             try self.addDbgInfoType(
                 dbg_type_arena.allocator(),
+                module,
                 ty,
                 dbg_info_buffer,
                 dbg_info_type_relocs,
@@ -771,6 +772,7 @@ pub fn freeDecl(self: *Dwarf, decl: *Module.Decl) void {
 fn addDbgInfoType(
     self: *Dwarf,
     arena: Allocator,
+    module: *Module,
     ty: Type,
     dbg_info_buffer: *std.ArrayList(u8),
     dbg_info_type_relocs: *File.DbgInfoTypeRelocsTable,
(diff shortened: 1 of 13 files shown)


src/link/Dwarf.zig
```diff
@@ -1095,6 +1097,97 @@ fn addDbgInfoType(
...
+            dbg_info_buffer.appendAssumeCapacity(abbrev_enum_variant);
+            // DW.AT.name, DW.FORM.string
+            dbg_info_buffer.appendSliceAssumeCapacity(no_error);
+            dbg_info_buffer.appendAssumeCapacity(0);
+            // DW.AT.const_value, DW.FORM.data8
+            mem.writeInt(u64, dbg_info_buffer.addManyAsArrayAssumeCapacity(8), 0, target_endian);
+
+            const error_names = blk: {
+                if (ty.isAnyError())
+                    break :blk module.error_name_list.items;
```
andrewrk, line 1122: The problem here is is that the list is not complete until flush(), however updateDecl() and updateFunc() will be called many times with references to this type before it is complete.

Here is my counter-proposal:
 * For debug info purposes, treat all error sets as `anyerror`.
 * Populate the DWARF info for `anyerror` in flush().

This is simple, easy to get right, and provides a completely reasonable debugging experience.

andrewrk (changes requested):

andrewrk:
Nice work btw :grin:

src/link/Dwarf.zig
kubkon: OK, I think I've arrived at something I'm pretty happy with. tl;dr we generate `anyerror` in `flush` for errors that were either an `anyerror` or an inferred error set which wasn't resolved yet. From my local testing, works like a charm. This actually brings another topic on the agenda: DWARF testing! Gotta figure out a good test harness given how feature-full it is becoming.

andrewrk:
Wow, this looks like some serious progress on making DWARF support incremental compilation!

src/link/Dwarf.zig
```diff
@@ -31,27 +31,89 @@ dbg_line_fn_free_list: std.AutoHashMapUnmanaged(*SrcFn, void) = .{},
...
 
+/// TODO replace with InternArena
 /// Table of debug symbol names.
 strtab: std.ArrayListUnmanaged(u8) = .{},
 
-pub const DebugInfoAtom = struct {
+/// Lives only as long as the analysed Decl.
+/// Allocated with `initDeclState`.
+/// Freed with `commitDeclState`.
+decl_state: ?DeclState = null,
```
andrewrk, line 49: I think it would be better to eliminate this field. It does not belong here; it belongs as ephemeral data in the space between init() and commit(). init() should return some kind of work-in-progress type, and commit() should consume it.

It took me a long time to understand what the purpose of this field is based on its existence inside this struct because it does not belong here; the Dwarf struct is global to the whole compilation but this `decl_state` is for exactly one Decl.

Furthermore, it encourages buggy code, such as the problem I'm currently running into, where one of the `try` has triggered in Elf.zig between init() and commit(), causing the `decl_state` to be non-null and the assertion tripping next time init() is called.

I suggest to remove this field, and see where that naturally leads the code to be instead.

andrewrk: e39c86399da565d3855d54b242bf7b5940e3924d

kubkon: Thanks for pointing this out @andrewrk - fix landed in https://github.com/ziglang/zig/pull/11349

PR #11301, opened 2022-03-25
`Namespace.decls` use context to save memory

jagt:
Greetings!

This commit implements the TODO above `Module.Namespace.Decl` that uses hash map context to save memory.

There's a few things that I'm not sure:

1. Should I use `mem.sliceTo(s, 0)` or `mem.span(s)` to convert `[*:0]const u8` to slice?
2. Is this the intended way to "Save Memory"? Or maybe it's about doing the string table buffer approach as in AstGen?

Cheers!

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -1591,6 +1591,19 @@ pub const Var = struct {
     }
 };
 
+pub const DeclAdapter = struct {
+    pub fn hash(self: @This(), s: []const u8) u32 {
+        _ = self;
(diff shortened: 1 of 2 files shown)


Vexu (approved):
> * Should I use `mem.sliceTo(s, 0)` or `mem.span(s)` to convert `[*:0]const u8` to slice?

For `[*:0]const u8` they both do the same thing

> * Is this the intended way to "Save Memory"? Or maybe it's about doing the string table buffer approach as in AstGen?

This is exactly it, the value already contains the key so there is no need to store separately.

PR #11302, opened 2022-03-25
stage2: runtime safety checks for slicing a null C pointer and @intCast truncating bits

mitchellh:
I'll add a compiler error test once #11284 is figured out...

Note: this forces some behavior tests for the CBE to be disabled because it is failing at rendering `int_big_positive` correctly for the max int values that are generated during the safety checks.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -6782,6 +6782,20 @@ fn intCast(
     }
 
     if ((try sema.typeHasOnePossibleValue(block, dest_ty_src, dest_ty))) |opv| {
+        // requirement: intCast(u0, input) iff input == 0
+        if (runtime_safety and block.wantSafety()) {
+            try sema.requireRuntimeBlock(block, operand_src);
+            const target = sema.mod.getTarget();
+            const wanted_info = dest_ty.intInfo(target);
+            const wanted_bits = wanted_info.bits;
+
+            if (wanted_bits == 0) {
+                const zero_inst = try sema.addConstant(sema.typeOf(operand), Value.zero);
+                const is_in_range = try block.addBinOp(.cmp_eq, operand, zero_inst);
+                try sema.addSafetyCheck(block, is_in_range, .cast_truncated_data);
+            }
+        }
+
         return sema.addConstant(dest_ty, opv);
     }
 
@@ -6794,7 +6808,7 @@ fn intCast(
         const actual_bits = actual_info.bits;
         const wanted_bits = wanted_info.bits;
 
-        // requirement: operand can fit into bit size of destination type
+        // requirement: bitSizeOf(operand) <= bitSizeOf(destination type)
         if (actual_bits > wanted_bits) {
             const max_int = try dest_ty.maxInt(sema.arena, target);
             const max_int_inst = try sema.addConstant(operand_ty, max_int);
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -6781,14 +6781,27 @@ fn intCast(
         return sema.fail(block, operand_src, "unable to cast runtime value to 'comptime_int'", .{});
     }
 
-    // TODO insert safety check to make sure the value fits in the dest type
-    _ = runtime_safety;
-
     if ((try sema.typeHasOnePossibleValue(block, dest_ty_src, dest_ty))) |opv| {
         return sema.addConstant(dest_ty, opv);
     }
 
     try sema.requireRuntimeBlock(block, operand_src);
+    if (runtime_safety) {
+        const target = sema.mod.getTarget();
+        const operand_ty = sema.typeOf(operand);
+        const actual_info = operand_ty.intInfo(target);
+        const wanted_info = dest_ty.intInfo(target);
+        const actual_bits = actual_info.bits;
+        const wanted_bits = wanted_info.bits;
+
+        // requirement: operand can fit into bit size of destination type
+        if (actual_bits > wanted_bits) {
+            const max_int = try dest_ty.maxInt(sema.arena, target);
+            const max_int_inst = try sema.addConstant(operand_ty, max_int);
+            const is_in_range = try block.addBinOp(.cmp_lte, operand, max_int_inst);
+            try sema.addSafetyCheck(block, is_in_range, .cast_truncated_data);
+        }
+    }
     return block.addTyOp(.intcast, dest_ty, operand);
 }
 
@@ -16166,6 +16179,7 @@ pub const PanicId = enum {
     incorrect_alignment,
     invalid_error_code,
     index_out_of_bounds,
+    cast_truncated_data,
 };
 
 fn addSafetyCheck(
@@ -16288,6 +16302,7 @@ fn safetyPanic(
         .incorrect_alignment => "incorrect alignment",
         .invalid_error_code => "invalid error code",
         .index_out_of_bounds => "attempt to index out of bounds",
+        .cast_truncated_data => "integer cast truncated bits",
     };
 
     const msg_inst = msg_inst: {
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -6800,7 +6800,7 @@ fn intCast(
     }
 
     try sema.requireRuntimeBlock(block, operand_src);
-    if (runtime_safety) {
+    if (runtime_safety and block.wantSafety()) {
         const target = sema.mod.getTarget();
         const operand_ty = sema.typeOf(operand);
         const actual_info = operand_ty.intInfo(target);
@@ -6808,8 +6808,21 @@ fn intCast(
         const actual_bits = actual_info.bits;
         const wanted_bits = wanted_info.bits;
 
-        // requirement: bitSizeOf(operand) <= bitSizeOf(destination type)
-        if (actual_bits > wanted_bits) {
+        // requirement: signed to unsigned >= 0
+        if (actual_info.signedness == .signed and
+            wanted_info.signedness == .unsigned)
+        {
+            const zero_inst = try sema.addConstant(sema.typeOf(operand), Value.zero);
+            const is_in_range = try block.addBinOp(.cmp_gte, operand, zero_inst);
+            try sema.addSafetyCheck(block, is_in_range, .cast_truncated_data);
+        }
+
+        // requirement: unsigned int value fits into target type
+        if (actual_bits > wanted_bits or
+            (actual_bits == wanted_bits and
+            actual_info.signedness == .unsigned and
+            wanted_info.signedness == .signed))
+        {
             const max_int = try dest_ty.maxInt(sema.arena, target);
             const max_int_inst = try sema.addConstant(operand_ty, max_int);
             const is_in_range = try block.addBinOp(.cmp_lte, operand, max_int_inst);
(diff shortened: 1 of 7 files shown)


mitchellh:
I've got a reduction on the C issue, looking into it.

src/Sema.zig
```diff
@@ -6746,14 +6746,41 @@ fn intCast(
...
     try sema.requireRuntimeBlock(block, operand_src);
+    if (runtime_safety) {
+        const target = sema.mod.getTarget();
+        const operand_ty = sema.typeOf(operand);
+        const actual_info = operand_ty.intInfo(target);
+        const wanted_info = dest_ty.intInfo(target);
+        const actual_bits = actual_info.bits;
+        const wanted_bits = wanted_info.bits;
+
+        // requirement: operand can fit into bit size of destination type
```
matu3ba, line 6776: requirement: bitSizeOf(operand) <= bitSizeOf(destination type) ?

src/Sema.zig
```diff
@@ -16103,6 +16131,7 @@ fn safetyPanic(
         .incorrect_alignment => "incorrect alignment",
         .invalid_error_code => "invalid error code",
         .index_out_of_bounds => "attempt to index out of bounds",
+        .cast_truncated_data => "integer cast truncated bits",
```
matu3ba, line 16134: This error is not very explanatory, because there is also the trunace inbuild https://ziglang.org/documentation/master/#truncate

src/Sema.zig
```diff
@@ -6746,14 +6746,41 @@ fn intCast(
         return sema.fail(block, operand_src, "unable to cast runtime value to 'comptime_int'", .{});
     }
 
-    // TODO insert safety check to make sure the value fits in the dest type
-    _ = runtime_safety;
-
     if ((try sema.typeHasOnePossibleValue(block, dest_ty_src, dest_ty))) |opv| {
+        // requirement: intCast to u0 must be 0
```
matu3ba, line 6750: requirement: intCast(u0, input) == 0

matu3ba (commented):
some nits on requirements formulation and less ideal error message

src/Sema.zig
mitchellh: I copied the same error message as stage1. Do you have a suggestion on something else?

matu3ba: should be fine then, but adding a TODO would be nice.

src/Sema.zig
```diff
@@ -6746,14 +6746,54 @@ fn intCast(
...
+        // requirement: intCast(u0, input) iff input == 0
+        if (runtime_safety) {
+            try sema.requireRuntimeBlock(block, operand_src);
+            const target = sema.mod.getTarget();
+            const wanted_info = dest_ty.intInfo(target);
+            const wanted_bits = wanted_info.bits;
+
+            if (wanted_bits == 0) {
+                const zero_inst = try sema.addConstant(sema.typeOf(operand), Value.zero);
+                const is_in_range = try block.addBinOp(.cmp_eq, operand, zero_inst);
```
mitchellh, line 6759: I'm unsure why I need `zero_inst` here and can't just use `.zero`. I'm guessing due to typing?

andrewrk: `Air.Inst.Ref.zero` has type `usize`

src/Sema.zig
```diff
@@ -6746,14 +6746,54 @@ fn intCast(
         return sema.fail(block, operand_src, "unable to cast runtime value to 'comptime_int'", .{});
     }
 
-    // TODO insert safety check to make sure the value fits in the dest type
-    _ = runtime_safety;
-
     if ((try sema.typeHasOnePossibleValue(block, dest_ty_src, dest_ty))) |opv| {
+        // requirement: intCast(u0, input) iff input == 0
+        if (runtime_safety) {
```
andrewrk, line 6751:
```suggestion
        if (runtime_safety and block.wantSafety()) {
```

src/Sema.zig
```diff
@@ -6746,14 +6746,54 @@ fn intCast(
...
+                const is_in_range = try block.addBinOp(.cmp_eq, operand, zero_inst);
+                try sema.addSafetyCheck(block, is_in_range, .cast_truncated_data);
+            }
+        }
+
         return sema.addConstant(dest_ty, opv);
     }
 
     try sema.requireRuntimeBlock(block, operand_src);
+    if (runtime_safety) {
```
andrewrk, line 6768:
```suggestion
    if (runtime_safety and block.wantSafety()) {
```

src/Sema.zig
```diff
@@ -19767,6 +19809,14 @@ fn analyzeSlice(
                 slice_ty = ptr_ptr_child_ty;
                 array_ty = ptr_ptr_child_ty;
                 elem_ty = ptr_ptr_child_ty.childType();
+
+                if (ptr_ptr_child_ty.ptrSize() == .C) {
+                    if (try sema.resolveMaybeUndefVal(block, ptr_src, ptr_or_slice)) |ptr_val| {
+                        if (!ptr_val.isUndef() and ptr_val.isNull()) {
```
andrewrk, line 19815: This is a case for `resolveDefinedValue` instead of `resolveMaybeUndefVal` because according the proposed undef semantics, if there exists an adversarial operand value that would cause Illegal Behavior, and that operand is `undefined`, then Illegal Behavior is invoked. In other words, `isUndef()` is treated equivalent to `isNull()` in this case.

andrewrk (changes requested):
Just a couple requests and then this is good to go.

I spent a little bit of time thinking if the `@intCast` safety could be done with only a single branch but I couldn't come up with a solution. Would be fun to brainstorm a bit. Regardless, the double check here is an improvement on status quo.

src/Sema.zig
```diff
@@ -6746,14 +6746,54 @@ fn intCast(
...
+        if (actual_info.signedness == .signed and
+            wanted_info.signedness == .unsigned)
+        {
+            const zero_inst = try sema.addConstant(sema.typeOf(operand), Value.zero);
+            const is_in_range = try block.addBinOp(.cmp_gte, operand, zero_inst);
+            try sema.addSafetyCheck(block, is_in_range, .cast_truncated_data);
+        }
+
+        // requirement: int value fits into target type
+        if (actual_bits > wanted_bits or
```
matu3ba, line 6786: Why is `actual_bits > wanted_bits` only part of this branch?

I would expect that `intCast(u16, number_with_type_u32)` is also checked for maxInt and
`intCast(i16, number_with_type_i32)` is checked for minInt and maxInt.

Can you explain or enumerate briefly the possible cases?

matu3ba (commented):
Had more time for review now and I have some (probably stupid) questions and idea how to unify the two sequential if conditions

1. semantics wrt `intCast(u16, var_with_type_u32)` and `intCast(i16, var_with_type_i32)`
2. potential unification of branch with 2 entry jump table to the enum entries, reading xoring data
3. enumeration of all cases of intCast would be great, because I dont understand why only one branch needs `actual_bits > wanted_bits`

src/Sema.zig
mitchellh: I may be misunderstanding, but your examples would get checked, because 32 > 16 in both cases. At, least, in my local test, they do get checked fine.

src/Sema.zig
```diff
@@ -6746,14 +6746,54 @@ fn intCast(
...
+    if (runtime_safety) {
+        const target = sema.mod.getTarget();
+        const operand_ty = sema.typeOf(operand);
+        const actual_info = operand_ty.intInfo(target);
+        const wanted_info = dest_ty.intInfo(target);
+        const actual_bits = actual_info.bits;
+        const wanted_bits = wanted_info.bits;
+
+        // requirement: signed to unsigned >= 0
+        if (actual_info.signedness == .signed and
```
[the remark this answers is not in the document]
mitchellh: @andrewrk and @wsengir found a good alternative to this. I'll let them decide if they want to integrate it. See my latest comment (unfortunately I don't have time in the short term).

mitchellh:
Addressed all feedback, except making the runtime check more efficient amongst the two cases. I got to the diff below with the help of @wsengir's work but it still has one edge case it doesn't handle correctly. I will have limited computer access time this next week so if one of you want to take it over, that'd be great! Sorry.

```diff
diff --git a/src/Sema.zig b/src/Sema.zig
index beab95752..bed4bca02 100644
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -6805,29 +6805,39 @@ fn intCast(
         const operand_ty = sema.typeOf(operand);
         const actual_info = operand_ty.intInfo(target);
         const wanted_info = dest_ty.intInfo(target);
-        const actual_bits = actual_info.bits;
-        const wanted_bits = wanted_info.bits;
+        const actual_bits = actual_info.bits - @boolToInt(actual_info.signedness == .signed);
+        const wanted_bits = wanted_info.bits - @boolToInt(wanted_info.signedness == .signed);
+        const sign_loss = actual_info.signedness == .signed and
+            wanted_info.signedness == .unsigned;
+
+        if (wanted_bits < actual_bits) {
+            // requirement: int value fits into smaller target type
+            const actual_unsigned_ty = if (actual_info.signedness == .signed)
+                try Type.Tag.int_unsigned.create(sema.arena, actual_info.bits)
+            else
+                operand_ty;
+            const wanted_max = try dest_ty.maxInt(sema.arena, target);
+            const wanted_max_inst = try sema.addConstant(operand_ty, wanted_max);
+            const diff = try sema.analyzeArithmetic(block, .sub, wanted_max_inst, operand, operand_src, operand_src, operand_src);
+            const diff_operands: struct {
+                lhs: Air.Inst.Ref,
+                rhs: Air.Inst.Ref,
+            } = if (sign_loss) .{
+                .lhs = try sema.bitCast(block, actual_unsigned_ty, diff, operand_src),
+                .rhs = try sema.coerce(block, actual_unsigned_ty, wanted_max_inst, operand_src),
+            } else .{
+                .lhs = diff,
+                .rhs = wanted_max_inst,
+            };
 
-        // requirement: signed to unsigned >= 0
-        if (actual_info.signedness == .signed and
-            wanted_info.signedness == .unsigned)
-        {
+            const is_in_range = try block.addBinOp(.cmp_lte, diff_operands.lhs, diff_operands.rhs);
+            try sema.addSafetyCheck(block, is_in_range, .cast_truncated_data);
+        } else if (sign_loss) {
+            // requirement: unsigned int value fits into signed target type
             const zero_inst = try sema.addConstant(sema.typeOf(operand), Value.zero);
             const is_in_range = try block.addBinOp(.cmp_gte, operand, zero_inst);
             try sema.addSafetyCheck(block, is_in_range, .cast_truncated_data);
         }
-
-        // requirement: unsigned int value fits into target type
-        if (actual_bits > wanted_bits or
-            (actual_bits == wanted_bits and
-            actual_info.signedness == .unsigned and
-            wanted_info.signedness == .signed))
-        {
-            const max_int = try dest_ty.maxInt(sema.arena, target);
-            const max_int_inst = try sema.addConstant(operand_ty, max_int);
-            const is_in_range = try block.addBinOp(.cmp_lte, operand, max_int_inst);
-            try sema.addSafetyCheck(block, is_in_range, .cast_truncated_data);
-        }
     }
     return block.addTyOp(.intcast, dest_ty, operand);
 }
```

PR #11174, opened 2022-03-15
Sema: panic on int add/sub/mul overflow

wsengir:
This relies on `{add,sub,mul}_with_overflow`, so it will be incompatible
with most backends for now. Closes #6017.

PR #11311, opened 2022-03-26
stage2: Change semantics of AIR arithmetic overflow instructions

joachimschmidt557:
This partially implements #10248 but retains compatibility with the current signature and semantics of `@addWithOverflow` and similar builtins.

This paves the way for implementing these AIR instructions efficiently in self-hosted backends and should also help unblocking https://github.com/ziglang/zig/pull/11174.


matu3ba (approved):
Somehow, I messed up the doc comment for sub_with_overflow

PR #11320, opened 2022-03-27
 std.math.Complex: add `norm` and `argument`, `div` cleanup

leesongun:
This PR adds `norm` and `argument` to `std.math.Complex`, and simplifies `div`, `reciprocal` using `norm` and `conjugate`.

[Should I change doc-comment to algebraic norm?](https://math.stackexchange.com/questions/285398/what-is-the-norm-of-a-complex-number)


andrewrk:
closing old draft

PR #11325, opened 2022-03-27
std.mem.zeroInit: Fix behaviour with empty initialiser

### LINKED ISSUE #11323: std.mem.zeroInit fails to initialize a field
### Zig Version

0.9.0

### Steps to Reproduce

[This test case](https://gist.github.com/acarrico/ceced9788167c13a32e1992306a19ad7) illustrates a field initialization failure.

### Expected Behavior

All fields should be initialized to initializer, default value, or zero.

### Actual Behavior

A field with a default value is not being set to that value.

ominitay:
Closes #11323

Issue was that if an empty tuple was passed as the initialiser, `zeroInit` would return a zeroed struct. Correct behaviour is to return the struct with default field values, with the rest being zeroed. This PR fixes this behaviour and implements a test case to prevent future regressions.

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -515,6 +515,17 @@ test "zeroInit" {
         .b = 0,
         .a = 0,
     }, c);
+
+    const Foo = struct {
+        foo: u8 = 69,
+        bar: u8,
+    };
+
+    const f = zeroInit(Foo, .{});
+    try testing.expectEqual(Foo{
+        .foo = 69,
+        .bar = 0,
+    }, f);
 }
 
 /// Compares two slices of numbers lexicographically. O(n).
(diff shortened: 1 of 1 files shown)


Vexu:
I think it should first loop over the fields giving them their default values to match if the init value is a non tuple struct.

ominitay:
Huh, so is *that* the intended behaviour? That'll need a bit more work. I'd presumed that what was happening with tuples was intended behaviour, but just was not a correct conditional? Anyway will get to doing that.

ominitay:
Actually not sure if this is the right solution.

Jarred-Sumner:
fwiw, sometimes I use `std.mem.zeroes` to set values to zero for types which have non-zero default values

I don't know if that's relevant for `zeroInit` but they seem vaguely related

ominitay:
That's not relevant to `zeroInit`. `zeroInit` sets all values to default values where present, and the rest to zeroes. It also takes a tuple or struct to override specific fields.

ominitay:
> Actually not sure if this is the right solution.
Don't know what I was thinking here (was 11:30pm at the time) -- this is the correct way of doing this. If there's no concerns, then LGTM!

Issue #11306, opened 2022-03-25
stage2: is_non_error should always returns comptime true for empty inferred error set


Vexu:
#11126 but for inferred error sets.
```zig
fn foo() !void {}
test {
    foo() catch @compileError("fail");
}
```

I tried simply resolving the inferred error set in `Sema.analyzeIsNonErr` but that lead to a false positive compile error in: https://github.com/ziglang/zig/blob/fd067fbe8b14c445ed764da15fff5163156126df/lib/std/debug.zig#L1304

PR #11333, opened 2022-03-28
Stage2 fixes to get `zig2 build --help` working

### LINKED ISSUE #11326: AstGen tries to store result of catch in parent block
```zig
test {
    const config_h_text: u32 = blk: {
        var dir = open() catch unreachable;
        // error: expected u32, found a.S
        //     var dir = open(name) catch unreachable;
        //     ^
        //  %9 = decl_val("open") token_offset:6:19
        //   %10 = dbg_stmt(3, 23)
        //   %11 = call(.auto, %9, []) node_offset:6:23
        //   %12 = is_non_err(%11) node_offset:6:26
        //   %13 = condbr(%12, {
        //     %15 = err_union_payload_unsafe(%11) node_offset:6:26
        //     %16 = as(@Zir.Inst.Ref.u32_type, %15)
        //     %18 = break(%14, %16)
        //   }, {
        //     %17 = unreachable(safe) node_offset:6:32
        //   }) node_offset:6:26
        break :blk dir.bar();
    };
    _ = config_h_text;
}
fn open() error{A}!S {
    return undefined;
}
const S = struct {
    a: u32,
    fn bar(s: S) u32 {
        return s.a;
    }
};
```
This is blocking `zig2 build --help`.

### LINKED ISSUE #11327: stage2 produces invalid llvm ir for union init as function argument
```zig
const U = union {
    a: std.ArrayList([]const u8),
    b: u32,
};
fn bar(_: U) void {}
test {
    bar(.{ .a = std.ArrayList([]const u8).init(std.testing.allocator) });
    // Stored value type does not match pointer operand type!
    //   store %"array_list.ArrayListAligned([]const u8,null)"* %1, %"array_list.ArrayListAligned([]const u8,null)"* %11, align 8, !dbg !3902
    //  %"array_list.ArrayListAligned([]const u8,null)"*
}
```
This is blocking `zig2 build --help`.

Vexu:
With these fixes applied `zig2 build --help` works. ~~successfully miscompile most of the time.~~ It was branching on undefined values whose debug info was getting optimized away.

Closes  #11326
Closes  #11327

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -1885,7 +1885,9 @@ pub fn bufPrintZ(buf: []u8, comptime fmt: []const u8, args: anytype) BufPrintErr
 /// Count the characters needed for format. Useful for preallocating memory
 pub fn count(comptime fmt: []const u8, args: anytype) u64 {
     var counting_writer = std.io.countingWriter(std.io.null_writer);
-    format(counting_writer.writer(), fmt, args) catch |err| switch (err) {};
+    // TODO https://github.com/ziglang/zig/issues/11306
+    // format(counting_writer.writer(), fmt, args) catch |err| switch (err) {};
+    format(counting_writer.writer(), fmt, args) catch unreachable;
     return counting_writer.bytes_written;
 }
 
--- a/lib/std/zig/system/x86.zig
+++ b/lib/std/zig/system/x86.zig
@@ -30,6 +30,8 @@ pub fn detectNativeCpuAndFeatures(arch: Target.Cpu.Arch, os: Target.Os, cross_ta
         .features = Target.Cpu.Feature.Set.empty,
     };
 
+    if (@import("builtin").zig_backend != .stage1) return cpu; // TODO
+
     // First we detect features, to use as hints when detecting CPU Model.
     detectNativeFeatures(&cpu, os.tag);
 


Vexu:
```sh-session
$ zig2 build --help --zig-lib-dir lib
Usage: /home/vexu/Documents/zig/zig/zig-out/bin/zig build [steps] [options]

Steps:
  install (default)            Copy build artifacts to prefix path
  uninstall                    Remove build artifacts from prefix path
  docs                         Build documentation
  test-toolchain               Run the tests for the toolchain
  test-stage2                  Run the stage2 compiler tests
  test-fmt                     Run zig fmt against build.zig to make sure it works
  test-behavior                Run the behavior tests
  test-compiler-rt             Run the compiler_rt tests
  test-minilibc                Run the mini libc tests
  test-compare-output          Run the compare output tests
  test-standalone              Run the standalone tests
  test-stack-traces            Run the stack trace tests
  test-cli                     Test the command line interface
  test-asm-link                Run the assemble and link tests
  test-runtime-safety          Run the runtime safety tests
  test-translate-c             Run the C translation tests
  test-run-translated-c        Run the Run-Translated-C tests
  test-std                     Run the standard library tests
  test                         Run all the tests
...
```

lib/std/fmt.zig
```diff
@@ -1885,7 +1885,7 @@ pub fn bufPrintZ(buf: []u8, comptime fmt: []const u8, args: anytype) BufPrintErr
 /// Count the characters needed for format. Useful for preallocating memory
 pub fn count(comptime fmt: []const u8, args: anytype) u64 {
     var counting_writer = std.io.countingWriter(std.io.null_writer);
-    format(counting_writer.writer(), fmt, args) catch |err| switch (err) {};
+    format(counting_writer.writer(), fmt, args) catch unreachable;
```
andrewrk, line 1888: What's going on here? Ideally we would keep the previous code so that if a new error is introduced to the empty error, set we get a compile error here instead of asserting no error happens.

lib/std/fs.zig
```diff
@@ -1028,7 +1028,7 @@ pub const Dir = struct {
             .write_only => @as(u32, os.O.WRONLY),
             .read_write => @as(u32, os.O.RDWR),
         };
-        const fd = if (flags.intended_io_mode != .blocking)
+        const fd = if (builtin.zig_backend == .stage1 and flags.intended_io_mode != .blocking) // TODO
```
andrewrk, line 1031: `flags.intended_io_mode != .blocking` is supposed to evaluate to comptime `false` here, preventing referencing `std.event.Loop`. Are you sure we need the direct workaround? Let's look into why stage2 didn't figure out that this expression is comptime-known.

andrewrk (changes requested):
First few commits look golden. I have a few requests on some of the later ones-

lib/std/mem/Allocator.zig
```diff
@@ -299,7 +299,8 @@ pub fn allocAdvancedWithRetAddr(
     } else @alignOf(T);
 
     if (n == 0) {
-        return @as([*]align(a) T, undefined)[0..0];
+        var ptr = @as([*]align(a) T, undefined);
+        return ptr[0..0];
```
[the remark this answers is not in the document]
Vexu: This change was motivated by the fact that casting `@as(T, undefined) to E!T` results in the error bit being set to undefined, but looking at it again it seems that `analyzeSlice` is in the wrong here for returning `.undef` instead of `.empty_array`.

lib/std/fmt.zig
Vexu: This is #11306, I'll add a TODO comment.

lib/std/fs.zig
Vexu: `intended_io_mode` is coming from a non-comptime parameter here, if the condition is supposed to be comptime known then that's a bug in the function.

andrewrk: Look at the type definition though:

```
pub const ModeOverride = if (is_async) Mode else enum { blocking };
```

In our case, the enum has exactly 1 tag so it should be returning true for one-possible-value.

Vexu: I see, looks like the only `io_mode` related change that was needed was to swap the order in `windows.WriteFile`. And I guess those in `os.zig` could also be made to check `std.io.mode`.

andrewrk: This is a bit ugly but it does solve the problem:

```diff
--- a/lib/std/event/loop.zig
+++ b/lib/std/event/loop.zig
@@ -103,8 +103,13 @@ pub const Loop = struct {
         };
     };
 
-    var global_instance_state: Loop = undefined;
-    const default_instance: ?*Loop = switch (std.io.mode) {
+    const LoopOrVoid = switch (std.io.mode) {
+        .blocking => void,
+        .evented => Loop,
+    };
+
+    var global_instance_state: LoopOrVoid = undefined;
+    const default_instance: ?*LoopOrVoid = switch (std.io.mode) {
         .blocking => null,
         .evented => &global_instance_state,
     };

```

lib/std/os/windows.zig
```diff
@@ -528,7 +528,7 @@ pub fn WriteFile(
     offset: ?u64,
     io_mode: std.io.ModeOverride,
 ) WriteFileError!usize {
-    if (std.event.Loop.instance != null and io_mode != .blocking) {
+    if (io_mode != .blocking and std.event.Loop.instance != null) {
```
andrewrk, line 531: This change is not needed

src/Sema.zig
```diff
@@ -20235,7 +20235,16 @@ fn analyzeSlice(
 
         // Special case: @as([]i32, undefined)[x..x]
         if (new_len_int == 0) {
-            return sema.addConstUndef(return_ty);
+            var anon_decl = try block.startAnonDecl(src);
```
andrewrk, line 20238: let's try to make a reduction for this, I think the previous code should work fine.

andrewrk (changes requested):

src/Sema.zig
Vexu: Well, it does work, the issue comes from casting an undefined payload to an error union also setting the error code to undefined.
```zig
test {
    const a =  @as([*]u8, undefined)[0..0];
    var b: error{a}![]const u8 = a;
    _ = try b;
}
```

andrewrk: Thanks let me take a look. What I expect to happen here is that the coercion to `[]const u8` makes a *defined* slice, setting the ptr to undefined but the len to 0. Then `b` will have the error union be defined as well.

andrewrk: b59428e9f78b1f9f265c0ffdba79b128d77644d2

lib/std/child_process.zig
```diff
@@ -522,6 +522,7 @@ pub const ChildProcess = struct {
                 // has a value greater than 0
                 if ((fd[0].revents & std.os.POLL.IN) != 0) {
                     const err_int = try readIntFd(err_pipe[0]);
+                    @setRuntimeSafety(std.debug.runtime_safety and builtin.zig_backend == .stage1); // TODO
```
andrewrk, line 525: 05947ea870f95ab90a75e174079492f546baaf72

src/Sema.zig
Vexu: Seems like the same change should be applied to optionals.

lib/std/fmt.zig
andrewrk: c21f046a8b019c42aa0dbb0fb9c592b590edf977 :grin:

Vexu: Guess I'll just close this pr at this point and implement input variables in assembly.

PR #11335, opened 2022-03-28
zig cc: Don't link libc if provided `-nostdinc`

### LINKED ISSUE #11328: zig cc -nostdinc not working
### Zig Version

0.10.0-dev.1335+c64279b15

### Steps to Reproduce

`hello.c`
```c
#include <stdio.h>

int main(int argc, char *argv[])
{
  printf("hello\n");
  return 0;
}
```
```sh
$ zig cc -nostdinc hello.c
```

### Expected Behavior

Expect it to fail with this message:
```
hello.c:1:10: fatal error: 'stdio.h' file not found
#include <stdio.h>
         ^~~~~~~~~
1 error generated.
```

### Actual Behavior

Compiles succesfully.

topolarity:
Resolves #11328. Thanks to @marler8997 for the report and the test case

Linking libc in Zig pulls in the `#include` headers. To make sure these headers are not pulled in, we need to refrain from linking libc/libcpp if we were provided `-nostdinc`, `-nostdinc++` or synonymous flags.

I also noticed that we were treating `-nostdinc++` as if it were `-nostdlib++`, so this fixes that as well.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -970,8 +970,8 @@ pub fn create(gpa: Allocator, options: InitOptions) !*Compilation {
         const tsan = options.want_tsan orelse false;
         // TSAN is implemented in C++ so it requires linking libc++.
         const link_libcpp = options.link_libcpp or tsan;
-        const link_libc = link_libcpp or options.link_libc or options.link_libunwind or
-            target_util.osRequiresLibC(options.target);
+        const include_libc_headers = link_libcpp or options.link_libc or options.link_libunwind;
+        const link_libc = include_libc_headers or target_util.osRequiresLibC(options.target);
 
         const link_libunwind = options.link_libunwind or
             (link_libcpp and target_util.libcNeedsLibUnwind(options.target));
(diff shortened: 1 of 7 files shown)


topolarity:
The alternative fix would have been to update the Compilation interface to support linking libc/libcpp but _not_ include the header files - Let me know if that's the right way to go instead :+1:

marler8997:
> The alternative fix would have been to update the Compilation interface to support linking libc/libcpp but not include the header files - Let me know if that's the right way to go instead +1

Since there are separate flags for `-nostdinc` and `-nostdlib` (also `-nodefaultlibs`) this alternative seems like the more correct thing to do.  Thanks by the way for providing a fix for this so quickly.

topolarity:
> Since there are separate flags for `-nostdinc` and `-nostdlib` (also `-nodefaultlibs`) this alternative seems like the more correct thing to do. Thanks by the way for providing a fix for this so quickly.

I think I agree - Let me double-check how we're handling other includes/libraries and revise this PR this evening.

topolarity:
Pausing on the proper fix, at @andrewrk's recommendation. It touches a fair bit of compilation/linker code, so it may have to wait until a later date, if it's needed after all.

andrewrk:
closing old draft

PR #11341, opened 2022-03-29
std.testing: add writeZigFile for TmpDir

matu3ba:
* remove need for manual string concatenation for building binaries in test blocks
* include small program snippet to show how to get binary path with subslicing

### DIFF
--- a/lib/std/child_process.zig
+++ b/lib/std/child_process.zig
@@ -1353,23 +1353,16 @@ test "build and call child_process" {
     var it = try std.process.argsWithAllocator(allocator);
     defer it.deinit(); // no-op unless WASI or Windows
     const testargs = try testing.getTestArgs(&it);
-
     var tmp = testing.tmpDir(.{ .no_follow = true }); // ie zig-cache/tmp/8DLgoSEqz593PAEE
     defer tmp.cleanup();
-    const tmpdirpath = try tmp.getFullPath(allocator);
-    defer allocator.free(tmpdirpath);
     const child_name = "child"; // no need for suffixes (.exe, .wasm) due to '-femit-bin'
-    const suffix_zig = ".zig";
-    const child_path = try fs.path.join(allocator, &[_][]const u8{ tmpdirpath, child_name });
-    defer allocator.free(child_path);
-    const child_zig = try mem.concat(allocator, u8, &[_][]const u8{ child_path, suffix_zig });
-    defer allocator.free(child_zig);
+    const zigfile_path = try tmp.writeZigFile(allocator, childstr, child_name);
+    defer allocator.free(zigfile_path);
 
-    try tmp.dir.writeFile("child.zig", childstr);
-    try testing.buildExe(testargs.zigexec, child_zig, child_path);
-
+    const binary = zigfile_path[0 .. zigfile_path.len - 4]; // '.zig' is 4 characters
+    try testing.buildExe(testargs.zigexec, zigfile_path, binary);
     // spawn compiled file as child_process with argument 'hello world' + expect success
-    const args = [_][]const u8{ child_path, "hello world" };
+    const args = [_][]const u8{ binary, "hello world" };
     var child_proc = try ChildProcess.init(&args, allocator);
     defer child_proc.deinit();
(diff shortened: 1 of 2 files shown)


matu3ba:
Continuation from #11138 with advice from @iddev5 applied.

More testing of binary generation etc should be deferred until stuff from the `compiler test suite` is moved to minimize redundancy.

lib/std/child_process.zig
```diff
@@ -1353,23 +1353,16 @@ test "build and call child_process" {
...
-    const child_zig = try mem.concat(allocator, u8, &[_][]const u8{ child_path, suffix_zig });
-    defer allocator.free(child_zig);
-
-    try tmp.dir.writeFile("child.zig", childstr);
-    try testing.buildExe(testargs.zigexec, child_zig, child_path);
+    const zigfile_path = try tmp.writeZigFile(allocator, childstr, child_name);
+    defer allocator.free(zigfile_path);
 
+    const binary = zigfile_path[0 .. zigfile_path.len - 4]; // '.zig' is 4 characters
+    try testing.buildExe(testargs.zigexec, zigfile_path, binary);
```
iddev5, line 1363: The path to binary files can be calculated as:
```zig
const filename = std.fs.path.basename(zigfile_path);
const binary = :blk {
    const index = std.mem.lastIndexOfScalar(u8, filename, '.') orelse break :blk filename;
    if (index == 0) break :blk filename;
    break :blk filename[0..index];
}
```

This is a code snippet I created after modifying std.fs.path.extension() but just reversed and I succesfully use it in a project. It is as reliable as stdlib itself. This code can also be put inside buildExe in order to automatically deduce the binary name.

iddev5 (approved):

lib/std/child_process.zig
matu3ba: yes, however thats 6 LOC instead of 1 and we know that `.zig` is always encoded as ascii (cause we generated) and thus 4 bytes.

I can see the usage of a function instead of `const binary = zigfile_path[0 .. zigfile_path.len - 4]; // '.zig' is 4 characters` ie for static,dynamic etc libraries or other stuff.

Your suggestion belongs into a separate PR for `path.zig` as complement to 
the function `extension`. Maybe this is your next PR? ;-)

My reasoning for simplifying this is to have readable+reasonable short tests for testing the test runner (output).

iddev5: This is true, I was just suggesting a general purpose solution.

matu3ba:
CI issue unrelated, so I rebased on top of master.

lib/std/testing.zig
```diff
@@ -374,6 +374,42 @@ pub const TmpDir = struct {
...
+    pub fn writeZigFile(
+        self: *TmpDir,
+        alloc: std.mem.Allocator,
+        progstr: []const u8,
+        filename: []const u8,
+    ) ![]const u8 {
+        const tmpdir_path = try self.getFullPath(alloc);
+        defer alloc.free(tmpdir_path);
+        const suffix_zig = ".zig";
+        const zigfile_path = try std.mem.concat(alloc, u8, &[_][]const u8{
```
Vexu, line 402: Missing `errdefer` after this.

matu3ba: was fixed. good catch.

andrewrk:
Reverted in 3052597a734f87727fa7f1a0e92247f100df3e96

Vexu:
> Also I don't want any code snippets in doc strings. We have doctests for that.

Stage1 can't parse doctests currently.

matu3ba:
> This is an awfully specific API that I don't think belongs in std.testing.

I am fine with either decision, but I would prefer having some reason(s).

1. Are you mostly concerned about the restriction to write files with ending `.zig` ?
2. Or is there a more fundamental problem with tmp files generated and being run from the test runner?

2.1 In the latter case, more tailored/numerous IPC tests must have a 1. numbering schema or we will have 2. cluttered folder or 3. a lot duplicated code.
2.2 Alternatively, IPC is supposed to be (for now) not tested.

PR #11346, opened 2022-03-30
std: replace other uses of `std.meta.Vector` with `@Vector`

nektro:
followup to https://github.com/ziglang/zig/commit/7be340e3cc62285d77cbd13f19f4ff0a2a982e82


andrewrk:
Thanks!

PR #11349, opened 2022-03-30
dwarf: pass DeclState around instead of storing a temp global in Dwarf

kubkon:
Avoids many pitfalls connected with premature/early return in case
there are errors with Decl, etc. This is effectively bringing back
the old design however in a much nicer packaging, where every
mechanism related to tracking Decl's debug info is now nicely
wrapped in a single struct (aka the `DeclState`). This includes
relocation table, type arena, etc. It is now the caller's
responsibility to deinit the state (so that no memory is leaked)
after `Decl` has been analysed (or errored out). The caller here
is typically a linker such as `Elf` or `MachO`.

cc @andrewrk redone fix of https://github.com/ziglang/zig/commit/e39c86399da565d3855d54b242bf7b5940e3924d


andrewrk (approved):
This looks like a nice improvement to me. Thanks for taking my suggestion :slightly_smiling_face:

Issue #11340, opened 2022-03-29
stage2 LLVM miscompiles returning error union of struct with `u128`s


Vexu:
```zig
// pub export fn entry() void {
test {
    const blk = foo() catch unreachable;
    if (blk.a != 1) unreachable;
}
const S = struct {
    a: u128,
    b: u128,
    c: u128,
};
fn foo() error{}!S {
    return S{ .a = 1, .b = 2, .c = 3 };
}
```
Doesn't happen if the error union is changed for an optional or removed, also doesn't happen on `u64`s. 

The difference in the generated LLVM  IR quite minimal so I'm not sure what's at fault here:
```diff
diff --git a/b.ll b/a.ll
index 7125a8f9a..e46d5d58d 100644
--- a/b.ll
+++ b/a.ll
@@ -9,7 +9,7 @@ target triple = "x86_64-unknown-linux-gnu"
-%a.S = type { i64, i64, i64 }
+%a.S = type { i128, i128, i128 }

-@1 = private unnamed_addr constant %a.S { i64 1, i64 2, i64 3 }, align 8
+@1 = private unnamed_addr constant %a.S { i128 1, i128 2, i128 3 }, align 16
 
@@ -168,49 +168,49 @@ Block3:                                           ; preds = %Then1
 ; Function Attrs: nounwind
 define dso_local void @entry() #0 !dbg !462 {
 Entry:
-  %0 = alloca { i16, %a.S }, align 8
-  call fastcc void @a.foo({ i16, %a.S }* sret({ i16, %a.S }) %0), !dbg !474
+  %0 = alloca { i16, %a.S }, align 16
+  call fastcc void @a.foo({ i16, %a.S }* sret({ i16, %a.S }) %0), !dbg !475
 
 Block:                                            ; preds = %Then
-  %7 = load i64, i64* %6, align 8, !dbg !475
-  %8 = icmp ne i64 %7, 1, !dbg !475
+  %7 = load i128, i128* %6, align 16, !dbg !476
+  %8 = icmp ne i128 %7, 1, !dbg !476

-  call void @llvm.memcpy.p0i8.p0i8.i64(i8* align 8 %3, i8* align 8 bitcast (%a.S* @1 to i8*), i64 24, i1 false), !dbg !485
+  call void @llvm.memcpy.p0i8.p0i8.i64(i8* align 16 %3, i8* align 16 bitcast (%a.S* @1 to i8*), i64 48, i1 false), !dbg !486
 }
```

[...]

PR #11356, opened 2022-03-30
stage2: handle assembly input names

Vexu:
With this `zig2 build -Dskip-install-lib-files` works (on my machine) (most of the time).

`-Dskip-install-lib-files`  due to #11340.


andrewrk:
That CI failure is happening on master too. I've been trying to track it down but I haven't been able to reliably reproduce it yet.

PR #11357, opened 2022-03-30
Remove a std.debug.print from the dwarf.zig file

jamesmintram:
This was causing freestanding builds to fail due to the use of `std.debug.print`

Unsure if this is the correct path to take (as potentially valuable output is lost) - happy to accept feedback and adjust as needed.


ominitay:
If this is deemed to be important to keep, it should probably be replaced with `std.log.err`, which works fine on freestanding targets.

andrewrk:
Thanks!

PR #11365, opened 2022-03-31
Fix C include files not being in `whole` cache

### LINKED ISSUE #11063: cache is not invalidated for #include'd files
### Zig Version

0.10.0-dev.871+bb05a8a08

### Steps to Reproduce

run this:
```
printf "#include \"mytypes.h\"
my_float_type public_func(my_float_type a) {
    return a + 1.0f;
}">main.c

printf "typedef float my_float_type;" > mytypes.h

zig build-lib  --enable-cache  main.c      #get everything cached

printf "" > mytypes.h                             #remove the definition of my_float_type

zig build-lib  --enable-cache  main.c      #this should fail, but doesn't
```

### Expected Behavior

the second `zig build-lib` should fail because `my_float_type` is no longer defined in `mytypes.h`

### Actual Behavior

zig doesn't doesn't recompile since it doesn't seem to notice that `my types.h` changed and `my_float_type` is no longer defined.

you can make the caching system pick up the change and (as expected) fail the build by modifying `main.c`:
```
printf "\n">>main.c
zig build-lib  --enable-cache  main.c 
```

changing `main.c` results in the expected compile error.

This is more serious when using `zig build` and you don't have the option to turn off the cache. This bug is serious enough  to completely prevent using c source with zig – your code changes don't make it into the binary.

moosichu:
Adapting the logic used for @import and @embedFile, we needed to ensure that the include files generated by the clang dependency file are added to the whole cache manifest in addition to the c-object specific ones.

Due to the fact that these are scheduled on worker threads - the whole cache needs to be locked with a mutex on access, a heavy handed solution but workable in the short term.

This resolves  #11063

### DIFF
--- a/src/Cache.zig
+++ b/src/Cache.zig
@@ -665,7 +665,7 @@ pub const Manifest = struct {
         self.hash.hasher.update(&ch_file.bin_digest);
     }
 
-    fn addDepFilePostCommon(self: *Manifest, dir: fs.Dir, dep_file_basename: []const u8, comptime is_post_contents: bool) !void {
+    pub fn addDepFilePost(self: *Manifest, dir: fs.Dir, dep_file_basename: []const u8) !void {
         assert(self.manifest_file != null);
 
         const dep_file_contents = try dir.readFileAlloc(self.cache.gpa, dep_file_basename, manifest_file_size_max);
@@ -690,29 +690,7 @@ pub const Manifest = struct {
         while (true) {
             switch (it.next() orelse return) {
                 .target, .target_must_resolve => return,
-                .prereq => |file_path| if (is_post_contents) {
-                    const resolved_path = try fs.path.resolve(self.cache.gpa, &[_][]const u8{file_path});
-                    errdefer self.cache.gpa.free(resolved_path);
-
-                    const file = try fs.cwd().openFile(resolved_path, .{ .mode = .read_only });
-                    defer file.close();
-                    const file_stat = try file.stat();
-
-                    const contents = try self.cache.gpa.alloc(u8, @intCast(usize, file_stat.size));
-                    defer self.cache.gpa.free(contents);
-
-                    _ = try file.readAll(contents);
-
-                    try self.addFilePostContents(
-                        resolved_path,
-                        contents,
-                        .{
-                            .size = file_stat.size,
-                            .inode = file_stat.inode,
-                            .mtime = file_stat.mtime,
-                        },
-                    );
-                } else try self.addFilePost(file_path),
+                .prereq => |file_path| try self.addFilePost(file_path),
                 else => |err| {
                     try err.printError(error_buf.writer());
                     log.err("failed parsing {s}: {s}", .{ dep_file_basename, error_buf.items });
@@ -720,14 +698,6 @@ pub const Manifest = struct {
                 },
             }
         }
-    }
-
-    pub fn addDepFilePost(self: *Manifest, dir: fs.Dir, dep_file_basename: []const u8) !void {
-        try self.addDepFilePostCommon(dir, dep_file_basename, false);
-    }
-
-    pub fn addDepFilePostContents(self: *Manifest, dir: fs.Dir, dep_file_basename: []const u8) !void {
-        try self.addDepFilePostCommon(dir, dep_file_basename, true);
     }
 
     /// Returns a hex encoded hash of the inputs.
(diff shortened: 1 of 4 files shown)


moosichu:
I'm wrapping up for the day - unfortunately haven't figured out how C/C++ headers are currently added to the manifest once the work to compile them has been scheduled, and therefore my attempts to try and add them to the whole cache has proven unsuccessful. Will have another look with a fresh face in a few days hopefully.

moosichu:
Finally realised the "obvious" part I was missing & how clang was generating a dependency file from which we extract the header includes. Once that was done putting together an actual fix that worked was pretty simple. I've tested that at least with my project, it works in both the positive and negative cases (not modifying a file results in cache being used properly, modifying a file does trigger a rebuild as expected.).

moosichu:
I'm not entirely sure why the build is failing here - because I'm able to build stage 2 locally without any issues. (Although I'm not using the `-Duse-zig-libcxx` flag).

moosichu:
I think this is now ready for review :)

matu3ba:
This looks to be a nasty synchronisation issue. :o 
1. Did you use rr or how did you figure this out?
2. Can you add a test (ie from the issue or, if you want and can produce a simplified one)?
3. Did you look around abit and think other cache/synchronization stuff is missing basic tests?

moosichu:
> 1. how did you figure this out?

Looking at the code, understanding it and noticing it wasn't thread safe. When I implemented these changes the compiler stopped segfaulting when building stage 2.

> 2. Can you add a test (ie from the issue or, if you want and can produce a simplified one)?

Testing for race conditions is generally a diffucult problem without a systemic way for dealing with these things.

> 3. Did you look around abit and think other cache/synchronization stuff is missing basic tests?

Honestly this stuff is a bit above my experience with the zig project so far! I was just trying to fix what I thought would be a simple bug haha.

jagt:
Hey! I've checked out this fix in my local build and can confirm it fixed the issue for me.
I'll test the fix locally for a while and report back if anything's wrong.

moosichu:
How did the testing go? Any issues so far?

jagt:
@moosichu I can only say it's working well for my very small project in which I iterate on headers a lot. It indeed fixed the issue for me. Though I didn't spend time trying to understand the fix and cache logic yet.

kubkon:
cc'ing @andrewrk as he knows this bit of code best :-)

andrewrk:
Nice work, @moosichu!

> Although all this locking of the manifest has made me wonder about what happens if two zig processes are trying to access the global cache and if that case has also been considered.

Yes, this is the central use case that the entire cache system is designed around. The first process to obtain the file system lock on the manifest file will do the work, and subsequent processes will wait for the first process, then use the cached results.

Note that we do have `-fsanitize-thread` which can be used to detect race conditions at runtime.

The changes look good. I see some potential for improvement, such as moving the mutex lock to inside the `addDepFilePost` function, so that we do not hold the lock while reading the .d files from the filesystem. Generally, making the Cache abstraction thread-safe rather than handling the locking externally in Compilation looks like it can offer some performance enhancements by avoiding holding locks during file system access. However, this enhancement can be a separate issue, which I will file upon merging.

andrewrk:
I realized the fatal flaw in this solution just a moment too late. Pasting from the commit message of 06310e3d4eb47fed88b175891cb5865bb050f020:

    Wait a minute, I'm sorry, I need to revert this. The whole premise
    of this change is broken because the point of the hash is that it tells
    whether the same compilation has been done before. This requires items
    to be added to the hash in the same sequence every time. This means that
    introducing a lock is fundamentally broken because the order needs to be
    the same in future runs of the compiler, and not decided by threads
    racing against each other.
    
    The proper solution to this is to, in whole cache mode, append the hash
    inputs to some data structure, and then after the compilation is
    complete, do some kind of sorting on the hash inputs so that they will
    be the same order every time, then apply them in sequence. No lock on
    the Cache object is needed for this scheme.

moosichu:
Thanks for taking a look! Yeah - I just wanted to make sure the change itself was as simple as possible for resolving the fix locally rather than trying anything more complex instead it was incomplete/looking at the wrong thing not understanding the cache system super well (which has turned out to be the case). 

Bummer that it's not viable as-is! Unfortunately I don't have the bandwidth to look at this right now myself, so if anyone else is able to look at this it would be great.

PR #11366, opened 2022-03-31
std.simd: Fix suggestVectorSizeForCpu

ominitay:


ominitay:
Rebased to get tests passing, after forking branch from failing commit.

PR #11372, opened 2022-04-01
dwarf: emit debug info for local variables on x86_64

kubkon:
Add support for emitting debug info for local variables within a subprogram.
This required moving bits responsible for populating the debug info back to
`CodeGen` from `Emit` as we require the operand to be resolved at callsite
plus we need to know its type. Without enforcing this, we could end up
with a `dead` mcv.

cc @joachimschmidt557

### DIFF
--- a/src/arch/x86_64/CodeGen.zig
+++ b/src/arch/x86_64/CodeGen.zig
@@ -48,6 +48,7 @@ gpa: Allocator,
 air: Air,
 liveness: Liveness,
 bin_file: *link.File,
+debug_output: DebugInfoOutput,
 target: *const std.Target,
 mod_fn: *const Module.Fn,
 err_msg: ?*ErrorMsg,
@@ -337,6 +338,7 @@ pub fn generate(
         .liveness = liveness,
         .target = &bin_file.options.target,
         .bin_file = bin_file,
+        .debug_output = debug_output,
         .mod_fn = module_fn,
         .err_msg = null,
         .args = undefined, // populated after `resolveCallingConventionValues`
@@ -382,7 +384,6 @@ pub fn generate(
     };
 
     var mir = Mir{
-        .function = &function,
         .instructions = function.mir_instructions.toOwnedSlice(),
         .extra = function.mir_extra.toOwnedSlice(bin_file.allocator),
     };
@@ -391,7 +392,6 @@ pub fn generate(
     var emit = Emit{
         .mir = mir,
         .bin_file = bin_file,
-        .function = &function,
         .debug_output = debug_output,
         .target = &bin_file.options.target,
         .src_loc = src_loc,
@@ -3443,17 +3443,11 @@ fn airArg(self: *Self, inst: Air.Inst.Index) !void {
     const arg_index = self.arg_index;
     self.arg_index += 1;
 
+    const ty = self.air.typeOfIndex(inst);
     const mcv = self.args[arg_index];
-    const payload = try self.addExtra(Mir.ArgDbgInfo{
-        .air_inst = inst,
-        .arg_index = arg_index,
-        .max_stack = self.max_end_stack,
-    });
-    _ = try self.addInst(.{
-        .tag = .arg_dbg_info,
-        .ops = undefined,
-        .data = .{ .payload = payload },
-    });
+    const name = self.mod_fn.getParamName(arg_index);
+    const name_with_null = name.ptr[0 .. name.len + 1];
+
     if (self.liveness.isUnused(inst))
         return self.finishAirBookkeeping();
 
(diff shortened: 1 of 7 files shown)


src/arch/x86_64/CodeGen.zig
```diff
@@ -3903,13 +3933,99 @@ fn airDbgBlock(self: *Self, inst: Air.Inst.Index) !void {
 
 fn airDbgVar(self: *Self, inst: Air.Inst.Index) !void {
     const pl_op = self.air.instructions.items(.data)[inst].pl_op;
-    const name = self.air.nullTerminatedString(pl_op.payload);
     const operand = pl_op.operand;
-    // TODO emit debug info for this variable
-    _ = name;
+    const ty = self.air.typeOf(operand);
+
+    if (!self.liveness.operandDies(inst, 0)) {
```
kubkon, line 3939: This seems like a bug somewhere in the codegen code to require us to do this. I think I will investigate further and see if I can get rid of this check. For the record, if anyone is interested in why this check was needed, here's the example which leads to UB in the compiler as it tries to reference an anon decl that ceased to exist:

```zig
pub fn main() void {
    const T = struct { v: bool };
    const decls = [_][]const T{
        &[_]T{
            .{ .v = false },
        },
    };
    _ = decls;
}
```

kubkon: Here's the generated AIR for `main`:

```
# Begin Function AIR: main:                                                                                                                                                                                                                                                                                              
# Total AIR+Liveness bytes: 730B                                                                                                                                                                                                                                                                                         
# AIR Instructions:         46 (414B)                                                                                                                                                                                                                                                                                    
# AIR Extra Data:           13 (52B)                                                                                                                                                                                                                                                                                     
# AIR Values Bytes:         16 (128B)                                                                                                                                                                                                                                                                                    
# Liveness tomb_bits:       24B                                                                                                                                                                                                                                                                                          
 %2 = const_ty(*const type)                                                                                                                                                                                                                                                                                             
 %3 = constant(*const type, (decl ref 'main__struct_3'))                                                                                                                                                                                                                                                                
 %4 = constant(type, const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                               
 %6 = const_ty((inferred_alloc_const))                                                                                                                                                                                                                                                                                  
 %7 = constant(*[1][]const const_inferred_array.main__struct_3, (decl ref 'main__anon_5'))                                                                                                                                                                                                                              
 %8 = constant(comptime_int, 1)                                                                                                                                                                                                                                                                                         
 %9 = const_ty([]const const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                             
 %10 = constant(usize, 1)                                                                                                                                                                                                                                                                                               
 %11 = const_ty([1][]const const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                         
 %12 = const_ty([1][]const const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                         
 %14 = const_ty(*[1][]const const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                        
 %16 = constant(usize, 0)                                                                                                                                                                                                                                                                                               
 %17 = const_ty(*[]const const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                           
 %19 = constant(comptime_int, 1)                                                                                                                                                                                                                                                                                        
 %20 = constant(usize, 1)                                                                                                                                                                                                                                                                                               
 %21 = const_ty([1]const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                                 
 %22 = const_ty(struct{comptime v: bool = false})                                                                                                                                                                                                                                                                       
 %23 = constant(struct{comptime v: bool = false}, (aggregate))                                                                                                                                                                                                                                                          
 %24 = constant(bool, false)                                                                                                                                                                                                                                                                                            
 %25 = const_ty(const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                                    
 %26 = constant(const_inferred_array.main__struct_3, (aggregate))                                                                                                                                                                                                                                                       
 %27 = const_ty(*const [1]const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                          
 %28 = constant(*const [1]const_inferred_array.main__struct_3, (decl ref 'main__anon_4'))                                                                                                                                                                                                                               
 %29 = const_ty([]const const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                            
 %30 = constant([]const const_inferred_array.main__struct_3, (slice))                                                                                                                                                                                                                                                   
 %31 = const_ty([]const const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                            
 %33 = const_ty([1][]const const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                         
 %34 = constant([1][]const const_inferred_array.main__struct_3, (aggregate))                                                                                                                                                                                                                                            
 %35 = const_ty([1][]const const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                         
 %37 = const_ty(*const [1][]const const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                  
 %38 = constant(*const [1][]const const_inferred_array.main__struct_3, (decl ref 'main__anon_5'))                                                                                                                                                                                                                       
 %39 = const_ty(*const [1][]const const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                  
 %42 = const_ty([1][]const const_inferred_array.main__struct_3)                                                                                                                                                                                                                                                         
 %43 = constant([1][]const const_inferred_array.main__struct_3, (aggregate))                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                      
 %0!= dbg_block_begin()                                                                                                                                                                                                                                                                                                 
 %1!= dbg_stmt(2:5)                                                                                                                                                                                                                                                                                                     
 %5!= dbg_stmt(4:5)                                                                                                                                                                                                                                                                                                     
 %40!= dbg_var_ptr(%38!, decls)                                                                                                                                                                                                                                                                                         
 %41!= dbg_stmt(9:5)                                                                                                                                                                                                                                                                                                    
 %44!= dbg_block_end()                                                                                                                                                                                                                                                                                                  
 %45!= ret(@Zir.Inst.Ref.void_value)                                                                                                                                                                                                                                                                                    
# End Function AIR: main  
```

kubkon: If you notice, `%40!` references `%38!` which is only referenced by this instruction; however, resolving `%38` in `%40` tries lowering the composite type `*const [1][]const const_inferred_array.main___struct_3` into memory (`rodata` section) tied to the parent decl `main___anon_5` which by this time has already been GCed.

Checking for liveness like I did above seems to solve this. However, I am wondering if there is perhaps something I am missing? Should or could this be handled more elegantly in Sema, or perhaps should we overhaul `fn resolveInst()` in x64 codegen?

kubkon: Furthermore, if you notice the debug instruction `dbg_var_ptr` is the only instruction referencing the constant instruction `%38` which does not seem right - I would expect dbg instructions to only ever reference instructions referenced by other non-debug instructions. Otherwise, we make the already-dead-by-optimisation instruction (here `%38`) outlive itself.

Anyway, looking forward to what you think about this @andrewrk, @joachimschmidt557 and @Luukdegram!

Luukdegram: For codegen, we do not emit anything if the result is not referenced (dies). But since debug instructions are never referenced, I think it makes sense to check the operand instead.
Perhaps ideally Sema wouldn't even emit this debug instruction, but I don't think the solution you came up with is hacky or wrong in any way.

kubkon: I am going to make an exec decision here and just merge this as is, but with the intention of discussing this with Andrew, Luuk, and Joachim in person this week in Milan.

Issue #11375, opened 2022-04-01
stage2: defers are generated incorrectly in labeled blocks


Vexu:
```zig
test "incorrect" {
    _ = blk: {
        defer foo();
        break :blk try bar();
    };
}
test "correct" {
    defer foo();
    try bar();
}
fn foo() void {
    std.debug.print("foo called\n", .{});
}
fn bar() !void {
    std.debug.print("bar called\n", .{});
}
```
```sh-session
$ zig test a.zig
Test [1/2] test "incorrect"... bar called
foo called
Test [2/2] test "correct"... bar called
foo called
All 2 tests passed.
$ zig2 test a.zig
Test [1/2] test.incorrect... foo called
bar called
Test [2/2] test.correct... bar called
foo called
All 2 tests passed.
```

PR #11376, opened 2022-04-01
stage2 llvm: properly align error union payload

### LINKED ISSUE #11340: stage2 LLVM miscompiles returning error union of struct with `u128`s

Vexu:
I'm not sure if the order of the fields is important or not because this would likely be easier solved by swapping the order.

Closes #11340

`-Dskip-install-lib-files` is still required due to #11375.

### DIFF
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -1270,7 +1270,8 @@ pub const Object = struct {
                 offset = std.mem.alignForwardGeneric(u64, offset, payload_align);
                 const payload_offset = offset;
 
-                const fields: [2]*llvm.DIType = .{
+                var len: u8 = 2;
+                var fields: [3]*llvm.DIType = .{
                     dib.createMemberType(
                         fwd_decl.toScope(),
                         "tag",
@@ -1293,7 +1294,21 @@ pub const Object = struct {
                         0, // flags
                         try o.lowerDebugType(payload_ty, .full),
                     ),
+                    undefined,
                 };
+
+                const error_size = Type.anyerror.abiSize(target);
+                if (payload_align > error_size) {
+                    fields[2] = fields[1];
+                    const pad_len = @intCast(u32, payload_align - error_size);
+                    fields[1] = dib.createArrayType(
+                        pad_len * 8,
+                        8,
+                        try o.lowerDebugType(Type.u8, .full),
+                        @intCast(c_int, pad_len),
+                    );
+                    len += 1;
+                }
 
                 const full_di_ty = dib.createStructType(
                     compile_unit_scope,
@@ -1305,7 +1320,7 @@ pub const Object = struct {
                     0, // flags
                     null, // derived from
                     &fields,
-                    fields.len,
+                    len,
                     0, // run time lang
                     null, // vtable holder
                     "", // unique id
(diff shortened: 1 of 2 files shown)


kubkon:
So we're allowed to flip the fields around so that the payload may precede the error value? I think we should start documenting things like this so that we are consistent between the backends. For example, self-hosted backends currently enforce the ordering to always be error preceding the payload: https://github.com/ziglang/zig/blob/8b5d5f44e23c86082fba480f01092f2f7ffb3dfe/src/codegen.zig#L691-L743

Vexu:
> For example, self-hosted backends currently enforce the ordering to always be error preceding the payload:

That's also what the LLVM backend does and this pr doesn't change that. I was suggesting that if we flip the fields we wouldn't need the padding since the error code is always by value and will always have sufficient alignment. That is also how the LLVM backend currently handles optionals as opposed to self hosted backends which put the payload second if I understand correctly.

LemonBoy:
IMO the best way forward is to have the frontend decide the layour for {error,} unions instead of having to do this silly dance in every backend and hope they all pick the same order. Stage1 currently suffers from the same problem and keeping the logic in sync between the type layout and the value one was a pain and a constant source of bugs.

andrewrk:
> I'm not sure if the order of the fields is important or not because this would likely be easier solved by swapping the order.

It's not important - I agree that swapping the order would be a better solution so that we don't need the padding field sometimes. Although I think we would still need it for the case of a u8 payload.

Keep in mind the frontend code for calculating ABI size needs to be correct (and it can query which backend is being used if necessary). Currently you can see that it assumes error set first, payload second.

https://github.com/ziglang/zig/blob/3ee44ce949117e8e91348ef870b18b23571a408d/src/type.zig#L3199-L3204

> IMO the best way forward is to have the frontend decide the layour for {error,} unions instead of having to do this silly dance in every backend and hope they all pick the same order. Stage1 currently suffers from the same problem and keeping the logic in sync between the type layout and the value one was a pain and a constant source of bugs.

That's fine but then the backends would still have to query the layout from the frontend and still have logic to flip the fields sometimes. There's also this other trick that we're using:

https://github.com/ziglang/zig/blob/3ee44ce949117e8e91348ef870b18b23571a408d/src/type.zig#L6053-L6055

LemonBoy:
> That's fine but then the backends would still have to query the layout from the frontend and still have logic to flip the fields sometimes.

As it's already done for regular unions and structures, the codegen should have no say in where each field goes after the analysis phase is done.

andrewrk:
Merging with intention to make the frontend decide the field order in a future enhancement.

PR #11377, opened 2022-04-02
C backend: Fix array declarations

### LINKED ISSUE #11368: C backend cannot compile function calls with constant arrays as arguments
### Zig Version

0.10.0-dev.1693+243afdcdf

### Steps to Reproduce

To reproduce this bug, save the following Zig code to a file named `bug.zig`:
```zig
pub fn main() void {
    callee(.{1});
}

fn callee(x: [1]u8) void {
    _ = x;
}
```

Then, attempt to compile this with Zig's C backend:
```bash
zig build-exe bug.zig -fno-LLVM -ofmt=c
zig cc bug.c 
```

### Expected Behavior

It should compile successfully, as it does when using Zig's LLVM backend (`zig build-exe bug.zig`).

### Actual Behavior

What actually happens is that Zig emits C code seemingly without error, but the C code cannot be compiled by C compilers (`clang`, `gcc`, `zig cc`):
```
bug.c:841:13: error: expected expression
 bug_callee({1u});
            ^
1 error generated.
```

Here's the relevant snippet of Zig-emitted C code:
```c
static void bug_main(void) {
 /* file:2:5 */
 /* file:2:11 */
 bug_callee({1u});
 return;
}
```

I'm not very good with C, but I'm pretty sure this is supposed to be casted
```c
static void bug_main(void) {
 bug_callee((uint8_t[]){1u});
 return;
}
```
or declared as a variable first
```c
static void bug_main(void) {
 uint8_t a[] = {1u};
 bug_callee(a);
 return;
}
```

AnnikaCodes:
Hi, I've done my best to fix #11368, but there are a few things I'm uncertain about:
- how do I add test cases for the C backend? I have a test case and verified manually that this fixes it, but just adding it in `test/behavior/bugs` didn't actually seem to compile the C output.
- is there a better way to tell if we're in a function argument? I just passed in an enum to `renderValue`, but that adds a lot of noise to the rest of the code.

Thanks and sorry for my ignorance.

### DIFF
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -61,6 +61,11 @@ const FormatTypeAsCIdentContext = struct {
     mod: *Module,
 };
 
+const ValueRenderLocation = enum {
+    FunctionArgument,
+    Other,
+};
+
 /// TODO make this not cut off at 128 bytes
 fn formatTypeAsCIdentifier(
     data: FormatTypeAsCIdentContext,
@@ -259,7 +264,7 @@ pub const Function = struct {
                     0,
                 );
                 try writer.writeAll(" = ");
-                try f.object.dg.renderValue(writer, ty, val);
+                try f.object.dg.renderValue(writer, ty, val, .Other);
                 try writer.writeAll(";\n ");
                 return decl_c_value;
             },
@@ -298,7 +303,7 @@ pub const Function = struct {
             .constant => |inst| {
                 const ty = f.air.typeOf(inst);
                 const val = f.air.value(inst).?;
-                return f.object.dg.renderValue(w, ty, val);
+                return f.object.dg.renderValue(w, ty, val, .Other);
             },
             else => return f.object.dg.writeCValue(w, c_value),
         }
@@ -310,7 +315,7 @@ pub const Function = struct {
                 const ty = f.air.typeOf(inst);
                 const val = f.air.value(inst).?;
                 try w.writeAll("(*");
-                try f.object.dg.renderValue(w, ty, val);
+                try f.object.dg.renderValue(w, ty, val, .Other);
                 return w.writeByte(')');
             },
             else => return f.object.dg.writeCValueDeref(w, c_value),
@@ -384,7 +389,7 @@ pub const DeclGen = struct {
             try dg.renderTypecast(writer, ty);
             try writer.writeAll("){");
             var buf: Type.SlicePtrFieldTypeBuffer = undefined;
-            try dg.renderValue(writer, ty.slicePtrFieldType(&buf), val.slicePtr());
+            try dg.renderValue(writer, ty.slicePtrFieldType(&buf), val.slicePtr(), .Other);
             try writer.writeAll(", ");
             try writer.print("{d}", .{val.sliceLen(dg.module)});
             try writer.writeAll("}");
@@ -544,6 +549,7 @@ pub const DeclGen = struct {
         writer: anytype,
         ty: Type,
         val: Value,
+        location: ValueRenderLocation,
     ) error{ OutOfMemory, AnalysisFail }!void {
         const target = dg.module.getTarget();
         if (val.isUndefDeep()) {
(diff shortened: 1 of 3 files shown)


jmc-88:
> how do I add test cases for the C backend? I have a test case and verified manually that this fixes it, but just adding it in test/behavior/bugs didn't actually seem to compile the C output.

[`test/behavior.zig`](https://github.com/ziglang/zig/blob/a0a587ff85d3785f99c475d9e0d5f1eb9e27bd26/test/behavior.zig) is the entry point for all behavioral tests; in order to have new tests run automatically they need to be added to the list of tests there.

Though you may only need to enable the [existing test case from `fn.zig`](https://github.com/ziglang/zig/blob/a0a587ff85d3785f99c475d9e0d5f1eb9e27bd26/test/behavior/fn.zig#L351-L370=) to cover your bug fix, since it already exercises the "function call with anon literal" case.

jmc-88:
> is there a better way to tell if we're in a function argument? I just passed in an enum to `renderValue`, but that adds a lot of noise to the rest of the code.

Unfortunately `renderValue()` has no idea of the context it's called from (unless you tell it that, as in your change). Since assignment works fine as is, and the problem only occurs on function call arguments, it might be just enough to always add the cast to function call arguments, e.g. apply a diff like the following:

```diff
$ git diff
diff --git a/src/codegen/c.zig b/src/codegen/c.zig
index 408530594..89b552d7e 100644
--- a/src/codegen/c.zig
+++ b/src/codegen/c.zig
@@ -2751,7 +2751,11 @@ fn airCall(
             try writer.writeAll(", ")
         }
         if (f.air.value(arg)) |val| {
-            try f.object.dg.renderValue(writer, f.air.typeOf(arg), val);
+            const arg_ty = f.air.typeOf(arg);
+            try writer.writeByte('(');
+            try f.object.dg.renderTypecast(writer, arg_ty);
+            try writer.writeByte(')');
+            try f.object.dg.renderValue(writer, arg_ty, val);
         } else {
             const val = try f.resolveInst(arg);
             try f.writeCValue(writer, val);
```

Which makes the compound literal syntax work and produces a no-op cast to the destination type in other cases. Though this trades off a little bit of terseness in `src/codegen/c.zig` for more noisy output code... Your approach may be the more desirable one in the end.

AnnikaCodes:
> existing test case from `fn.zig`

I've tried that, and the test case does seem to cover my fix, but Zig doesn't seem to be compiling it for the C backend—when I enable it on master, it doesn't fail, even though the C code emitted for the equivalent function calls on their own would not compile.

jmc-88:
> > existing test case from `fn.zig`
> 
> I've tried that, and the test case does seem to cover my fix, but Zig doesn't seem to be compiling it for the C backend—when I enable it on master, it doesn't fail (when I run `./zig build test-behavior -Dskip-release`), even though the C code emitted for the equivalent function calls on their own would not compile.

I tested my local changes against that and it was falling before I applied the diff from https://github.com/ziglang/zig/pull/11377#issuecomment-1086743195. How are you running the tests? (We can move this discussion to Discord if you prefer, to reduce noise here)

AnnikaCodes:
I've just been running `make install` and `./zig build test -Dskip-release`. I've also tried compiling my own case with Zig's C backend and then compiling it with Clang.

I did find a bug in this where it adds extra casts to 2D arrays that cause compilation errors, so I've just pushed a fix for that.

jmc-88:
> I've just been running `make install` and `./zig build test -Dskip-release`. I've also tried compiling my own case with Zig's C backend and then compiling it with Clang.
> 
> I did find a bug in this where it adds extra casts to 2D arrays that cause compilation errors, so I've just pushed a fix for that.

My usual command line looks something like:

```
$ build/zig build --zig-lib-dir lib --prefix stage2 -Denable-llvm -Dconfig_h=build/config.h && \
  stage2/bin/zig test -I test/ -ofmt=c test/behavior.zig
```

This allows you to run only the behavioral tests (or individual ones from within `test/behavior/...`, if needed) that you care about and iterate faster; then you can test the whole thing before pushing, or leave it for the CI run.

AnnikaCodes:
> > I've just been running `make install` and `./zig build test -Dskip-release`. I've also tried compiling my own case with Zig's C backend and then compiling it with Clang.
> > I did find a bug in this where it adds extra casts to 2D arrays that cause compilation errors, so I've just pushed a fix for that.
> 
> My usual command line looks something like:
> 
> ```
> $ build/zig build --zig-lib-dir lib --prefix stage2 -Denable-llvm -Dconfig_h=build/config.h && \
>   stage2/bin/zig test -I test/ -ofmt=c test/behavior.zig
> ```
> 
> This allows you to run only the behavioral tests (or individual ones from within `test/behavior/...`, if needed) that you care about and iterate faster; then you can test the whole thing before pushing, or leave it for the CI run.

Awesome, thanks! This got me running the tests, which pass locally for me. Are they working for you now that I've fixed the bug with 2D arrays? Would you rather I switch to your idea of just casting all the function args?

jmc-88:
I'm personally OK with making the rendering context explicit for `renderValue()`, it seems necessary to know here and likely useful in the future too. I'll let Zig core members have the last word onit though 😃

andrewrk:
Thanks @AnnikaCodes!

PR #11379, opened 2022-04-03
Pull elf magic string out to re-used constant

moosichu:
With the work I'm doing on [zar](https://github.com/moosichu/zar) having this available in the standard library means I don't have to keep looking it up & pasting it in the places needed. This also matches what we already to with the magic for MachO, which has it's own constant as well.


kubkon (approved):

lib/std/elf.zig
```diff
@@ -305,6 +305,8 @@ pub const STT_ARM_16BIT = STT_HIPROC;
 pub const VER_FLG_BASE = 0x1;
 pub const VER_FLG_WEAK = 0x2;
 
+pub const ELF_MAGIC = "\x7fELF";
```
andrewrk, line 308: This declaration is in a namespace called "elf" already, so the prefix is redundant. It should be called instead `MAGIC`.

moosichu: Donzo :)

Issue #7738, opened 2021-01-10
std.ChildProcess doesn't allow to pipe two processes together.


vesim987:
Right now there is no way to pipe stdin from one process to stdout in another process because `spawnPosix` is always creating(and overwriting .stdin/out/err) pipes when `std*_behavior` is set to `Pipe`. Example code(provided by g_w1) affected by this, which should work:
```zig
const std = @import("std");
const ChildProcess = std.ChildProcess;
pub fn main() !void {
    const child1 = try ChildProcess.init(&[_][]const u8{ "cat", "test.txt" }, std.heap.page_allocator);
    defer child1.deinit();

    const child = try ChildProcess.init(&[_][]const u8{"cat"}, std.heap.page_allocator);
    defer child.deinit();

    child1.stdout_behavior = .Pipe;
    child.stdin_behavior = .Pipe;

    try child1.spawn();
    child.stdin = child1.stdout.?;
    try child.spawn();

    _ = try child1.wait();
    _ = try child.wait();
}
```

The one fix that I see for this is to check if the fields stdin/out/err are already set, and don't create pipes when they are valid `Files`. That would affect [this](https://github.com/ziglang/zig/blob/master/lib/std/child_process.zig#L413-L426).
The other fix is by creating new `StdIo` mode, something like `Passthrough` and handle it in `setUpChildIo` by just doing `os.dup2` like in the `.Pipe` case.

vesim987:
I tried working on that, but I don't have idea how to handle ownership of the pipes. In a PoC I used StdIo.Ignore to pass the pipe's file descriptor.
```diff
diff --git a/lib/std/child_process.zig b/lib/std/child_process.zig
index d37dd9fdf..77a1c8f39 100644
--- a/lib/std/child_process.zig
+++ b/lib/std/child_process.zig
@@ -492,9 +492,9 @@ pub const ChildProcess = struct {
         const pid_result = try os.fork();
         if (pid_result == 0) {
             // we are the child
-            setUpChildIo(self.stdin_behavior, stdin_pipe[0], os.STDIN_FILENO, dev_null_fd) catch |err| forkChildErrReport(err_pipe[1], err);
-            setUpChildIo(self.stdout_behavior, stdout_pipe[1], os.STDOUT_FILENO, dev_null_fd) catch |err| forkChildErrReport(err_pipe[1], err);
[...]

PR #11393, opened 2022-04-06
std.child_process: support arbitrary number of pipes

matu3ba:
* default remains with standard streams stdin, stdout, stderr

wip


matu3ba:
related #7738

matu3ba:
Summary on discussion with @MasterQ32 on portability of pipes in IPC:
Q(me): Are they treated "special" by OSes?
A(xq): yes, they are

Closing, because not portable in general and the default test_runner must be portable.

PR #11396, opened 2022-04-06
zig cc: support --subsystem linker flag

### LINKED ISSUE #11389: MinGW `-mwindows` flag support
I have this simple win32 program:

```cpp
#include <windows.h>

int WINAPI WinMain(HINSTANCE hInst, HINSTANCE hInstPrev, PSTR cmdline, int cmdshow)
{
    return MessageBox(NULL, "Hello, worl!", "Caption", 0);
}
```

Cross-compiling it from my Mac just works.

```
$ zig cc -o main.exe main.cpp -target x86_64-windows
```

(Which never ceases to amaze me.)

When starting the .exe from the Windows File Explorer, a terminal window shows up:

image

There are workarounds to make the terminal flash just briefly but that is still not ideal.

I found out there's a `-mwindows` flag I can pass when cross-compiling with MinGW:

```
$ x86_64-w64-mingw32-g++ -o main.exe main.cpp -mwindows
```

Without it, I have the same behaviour like with Zig, but _with_ it, there's no terminal window.

I wasn't able to determine what exactly that flag does. From little I was able to find it seems by default programs are built in the "console" mode and in this particular case we want the "GUI" mode.

Per:

```
$ x86_64-w64-mingw32-g++ --help | grep -- -m
Options starting with -g, -f, -m, -O, -W, or --param are automatically passed on to the various sub-processes invoked by x86_64-w64-mingw32-g++.
```

Perhaps it's:

```
$ x86_64-w64-mingw32-ld --help | grep -- -m
...
-m EMULATION                Set emulation
...
```

EDIT: it's not an ld flag. Possible options for `-m` are: `-mwindows`, `-mconsole`, and `-mdll`.

In any case, if it was desirable to expose this flag somehow, I'd be happy to look into it. If so, any pointers would be appreciated.

wojtekmach:
Example:

    $ zig cc -o main.exe main.cpp -target x86_64-windows -Wl,--subsystem,windows

Closes #11389

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -849,36 +849,7 @@ fn buildOutputType(
                         const next_arg = args_iter.next() orelse {
                             fatal("expected parameter after {s}", .{arg});
                         };
-                        if (mem.eql(u8, next_arg, "console")) {
-                            subsystem = .Console;
-                        } else if (mem.eql(u8, next_arg, "windows")) {
-                            subsystem = .Windows;
-                        } else if (mem.eql(u8, next_arg, "posix")) {
-                            subsystem = .Posix;
-                        } else if (mem.eql(u8, next_arg, "native")) {
-                            subsystem = .Native;
-                        } else if (mem.eql(u8, next_arg, "efi_application")) {
-                            subsystem = .EfiApplication;
-                        } else if (mem.eql(u8, next_arg, "efi_boot_service_driver")) {
-                            subsystem = .EfiBootServiceDriver;
-                        } else if (mem.eql(u8, next_arg, "efi_rom")) {
-                            subsystem = .EfiRom;
-                        } else if (mem.eql(u8, next_arg, "efi_runtime_driver")) {
-                            subsystem = .EfiRuntimeDriver;
-                        } else {
-                            fatal("invalid: --subsystem: '{s}'. Options are:\n{s}", .{
-                                next_arg,
-                                \\  console
-                                \\  windows
-                                \\  posix
-                                \\  native
-                                \\  efi_application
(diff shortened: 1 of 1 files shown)


wojtekmach:
Please let me know if `-Wl,--subsystem,windows` is fine or it should rather be:

```
-Wl,--subsystem=windows
               ^
```

src/main.zig
```diff
@@ -1467,6 +1438,11 @@ fn buildOutputType(
                                 mem.eql(u8, linker_arg, "-static"))
                             {
                                 force_static_libs = true;
+                            } else if (mem.eql(u8, linker_arg, "--subsystem")) {
+                                const next_arg = split_it.next() orelse {
```
kubkon, line 1442:
```suggestion
                                const next_arg = args_iter.next() orelse {
```

so that we are consistent with other linker args such as `--entry` or `--stack`. Lemme know what you think!

kubkon: Just realised you asked about this. Yeah, this will swap out `,` for `=` making it more consistent across the additional args IMHO.

kubkon: Please ignore my earlier code change suggestion - I misread where it was in `main.zig`. However, I'd still prefer we had `--subsystem=<...>` - as far as I can remember, you just need to defer parsing this arg until you collect them all into `linker_args` container.

wojtekmach: Sounds good, `—-subsystem=x` it is, stay tuned!

wojtekmach: @kubkon memory serves you well, fixed in b2344cc18e9fa5dfa1d19025caa2b0f583498734!

PR #11400, opened 2022-04-07
std.fs: prevent possible integer overflow in Dir.makePath

### LINKED ISSUE #11397: std.fs.cwd.makePath: integer overflow if CWD does not exist
### Zig Version

0.10.0-dev.1575+7f91be9c8

### Steps to Reproduce

Run `zig test` on the following:

```zig
const std = @import("std");

test "cwd.makePath with non-existent CWD" {
    var tmp_dir = std.testing.tmpDir(.{});
    try tmp_dir.dir.setAsCwd();
    tmp_dir.cleanup();

    const cwd = std.fs.cwd();
    try cwd.makePath("sub-path");
}
```

### Expected Behavior

The equivalent C code will set `errno` to `ENOENT` if trying to create a directory within a non-existent directory, so I would expect the test to fail with `error.NotFound`.

### Actual Behavior

```
Test [1/1] test "cwd.makePath with non-existent CWD"... thread 42590880 panic: integer overflow
/Users/ehaas/source/zig/build/lib/zig/std/fs.zig:1314:35: 0x108c24ce8 in std.fs.Dir.makePath (test)
                        end_index -= 1;
                                  ^
/Users/ehaas/test.zig:9:21: 0x108c235cc in test "cwd.makePath with non-existent CWD" (test)
    try cwd.makePath("sub-path");
                    ^
```

ehaas:
The call to `makeDir` for the top-level component of `sub_path`
can return `error.FileNotFound` if the directory represented by
`self` has been deleted.

Fixes #11397

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -1308,9 +1308,9 @@ pub const Dir = struct {
                     if (end_index == sub_path.len) return;
                 },
                 error.FileNotFound => {
-                    if (end_index == 0) return err;
                     // march end_index backward until next path component
                     while (true) {
+                        if (end_index == 0) return err;
                         end_index -= 1;
                         if (path.isSep(sub_path[end_index])) break;
                     }
--- a/lib/std/fs/test.zig
+++ b/lib/std/fs/test.zig
@@ -610,6 +610,16 @@ test "makePath, put some files in it, deleteTree" {
     }
 }
 
+test "makePath in a directory that no longer exists" {
+    if (builtin.os.tag == .windows) return error.SkipZigTest; // Windows returns FileBusy if attempting to remove an open dir
+
+    var tmp = tmpDir(.{});
+    defer tmp.cleanup();
+    try tmp.parent_dir.deleteTree(&tmp.sub_path);
+
+    try testing.expectError(error.FileNotFound, tmp.dir.makePath("sub-path"));
+}
+
 test "writev, readv" {
     var tmp = tmpDir(.{});
     defer tmp.cleanup();


matu3ba:
@ehaas 
```txt
2156 passed; 51 skipped; 0 failed.
926s
52540./lib/std/fs.zig:1477:13: error: changing cwd is not currently possible in WASI
936s
52541            @compileError("changing cwd is not currently possible in WASI");
936s
52542            ^
```
Remove WASI from the test.

Windows error looks unrelated and weird (could be flanky CI?):
```txt
Test 9/9 stack-trace try try return return (ReleaseSmall)...OK
.\lib\std\os.zig:3027:33: error: dependency on libc must be explicitly specified in the build command
            switch (errno(system.fchdir(dirfd))) {
                                ^
.\lib\std\c.zig:92:33: error: dependency on libc must be explicitly specified in the build command
        return @intToEnum(c.E, c._errno().*);
                                ^
.\lib\std\os.zig:3027:26: note: called from here
            switch (errno(system.fchdir(dirfd))) {
                         ^
.\lib\std\os.zig:3012:45: note: called from here
pub fn fchdir(dirfd: fd_t) FchdirError!void {
                                            ^
test...The following command exited with error code 1:
```

ehaas:
@matu3ba it looks like some of the CI's are complaining about `fchdir` at link time - any idea what sort of condition I should use for skipping tests so that those can pass?

ehaas:
Thanks for looking into it. For now I'll try changing the test to use `chdir` directly instead of indirectly using `fchdir`

Vexu (approved):
Does this need to be tested on CWD? Seems like it would be easier to use `tmp.dir`.

ehaas:
> Does this need to be tested on CWD? Seems like it would be easier to use `tmp.dir`.

Is something like this what you had in mind:

```zig
    var tmp = tmpDir(.{});
    tmp.parent_dir.deleteTree(&tmp.sub_path) catch {};
    try tmp.dir.makePath("sub-path");
```

That also reproduces the original bug without having to mess with CWD, so I could remove that and the Windows/WASI skip and see if they still pass.

Vexu:
> Is something like this what you had in mind:

Something like that yes, working on Windows/WASI would be a nice bonus.

PR #11401, opened 2022-04-07
math: simplify inf (+f80 support), deprecate old constants (followup for #10133)

viriuwu:
### Contents

- simplify `isFinite` to remove hardcoded constants + adds `f80` support + tests
- generalise `inf` the same way `floatMin` and others were in [#10133](https://github.com/ziglang/zig/pull/10133/)
- simplify `isInf`, `isPositiveInf`, `isNegativeInf` to remove hardcoded constants + add `f80` support + tests
- as part of preparing to deprecate the old `fXX_xxx` and `inf_fXX` constants, switch to new functions everywhere (except for one spot in tests/behaviour/pointers.zig, causes Sema to freak out with "zig compiler bug: needs source location" so I left it as-is for now)

### Notes

- the new `isPositiveInf` and `isNegativeInf` are so simple that they basically don't need to be functions anymore, it's not my call but I recommend looking at how simple the functions can be. If `isInf` was implemented in terms of `fabs` then it'd be a similar case, but I avoided the float -> int -> float conversion.
- while replacing the old constants in the codebase, I corrected two errors I found:
  1. there was a copy paste error in `round128`, testing for trapping with f64 in the f128 routine:
  https://github.com/ziglang/zig/blob/93e11b824a37a14fc392bfc64ed8f364f4fc7d46/lib/std/math/round.zig#L109-L114
  2. there was a test in compiler-rt testing for a `floatMax(f64)` to `i64` conversion in the **f32** -> i64 routine:
  https://github.com/ziglang/zig/blob/93e11b824a37a14fc392bfc64ed8f364f4fc7d46/lib/std/special/compiler_rt/fixsfdi_test.zig#L63
  this only worked due to some really strange behaviour with explicit `@as` catching out of bounds floats being coerced while in these cases it does not, investigating that


viriuwu:
Alright, maybe `c_longdouble` support for the functions on all platforms was a bit ambitious for now but it does work on like half of them, so that's good. Anyway, I've removed those tests, and CIs are looking happy.

cc @ifreund

lib/std/math/ceil.zig
```diff
@@ -62,6 +62,8 @@ fn ceil32(x: f32) f32 {
 }
 
 fn ceil64(x: f64) f64 {
+    const f64_toint = 1.0 / math.floatEps(f64);
```
matu3ba, line 65: There is no commit message, which explains that `ceil64` was broken and how it got fixed.

lib/std/math/floor.zig
```diff
@@ -98,6 +98,8 @@ fn floor32(x: f32) f32 {
 }
 
 fn floor64(x: f64) f64 {
+    const f64_toint = 1.0 / math.floatEps(f64);
```
matu3ba, line 101: same story: I dont see a commit message how `floor64` was broken and got fixed.

lib/std/math/isinf.zig
```diff
@@ -1,131 +1,60 @@
...
-    try expect(isInf(math.inf(f128)));
-    try expect(isInf(-math.inf(f128)));
+    // TODO remove when #11391 is resolved
+    if (@import("builtin").os.tag == .freebsd) return error.SkipZigTest;
+
+    inline for ([_]type{ f16, f32, f64, f80, f128 }) |T| {
+        try expect(!isInf(@as(T, 0.0)));
+        try expect(!isInf(@as(T, -0.0)));
+        try expect(isInf(math.inf(T)));
+        try expect(isInf(-math.inf(T)));
```
matu3ba, line 34: add +-NaN as test cases.

lib/std/math/isinf.zig
```diff
@@ -1,131 +1,60 @@
...
-    try expect(isPositiveInf(math.inf(f128)));
-    try expect(!isPositiveInf(-math.inf(f128)));
+    // TODO remove when #11391 is resolved
+    if (@import("builtin").os.tag == .freebsd) return error.SkipZigTest;
+
+    inline for ([_]type{ f16, f32, f64, f80, f128 }) |T| {
+        try expect(!isPositiveInf(@as(T, 0.0)));
+        try expect(!isPositiveInf(@as(T, -0.0)));
+        try expect(isPositiveInf(math.inf(T)));
+        try expect(!isPositiveInf(-math.inf(T)));
```
matu3ba, line 46: add +-NaN as test cases.

lib/std/math/isinf.zig
```diff
@@ -1,131 +1,60 @@
...
-    try expect(!isNegativeInf(math.inf(f128)));
-    try expect(isNegativeInf(-math.inf(f128)));
+    // TODO remove when #11391 is resolved
+    if (@import("builtin").os.tag == .freebsd) return error.SkipZigTest;
+
+    inline for ([_]type{ f16, f32, f64, f80, f128 }) |T| {
+        try expect(!isNegativeInf(@as(T, 0.0)));
+        try expect(!isNegativeInf(@as(T, -0.0)));
+        try expect(!isNegativeInf(math.inf(T)));
+        try expect(isNegativeInf(-math.inf(T)));
```
matu3ba, line 58: add +-NaN as test cases.

lib/std/special/compiler_rt/divtf3_test.zig
```diff
@@ -35,7 +35,7 @@ test "divtf3" {
     // NaN / any = NaN
     try test__divtf3(math.nan_f128, 0x1.23456789abcdefp+5, 0x7fff800000000000, 0);
     // inf / any = inf
```
matu3ba, line 37: nit: this comment is wrong:  inf/0 and inf/NaN are not inf, but NaN and could be tested.

matu3ba (changes requested):
add a few more test cases and please explain the often occuring changes like
`const f64_toint = 1.0 / math.floatEps(f64);` in the commit message with
what was broken and how it was fixed.

lib/std/math/ceil.zig
viriuwu: It was not broken, and it did not get fixed. The `fXX_toint` constants are getting deprecated (see #10133) as they have a very confusing name and are literally only used in the rounding functions and nowhere else, and it's just `1 / epsilon`. This is part of the cleanup.

matu3ba: ah, my bad then.

lib/std/special/compiler_rt/divtf3_test.zig
viriuwu: I'm pretty sure +inf/0 **is** +inf? Seems to be. I can add one for NaN, which indeed results in NaN.

lib/std/math/isinf.zig
viriuwu: good call!

andrewrk:
Nice work!

PR #11405, opened 2022-04-08
Reuse code in `Utf8Iterator.nextCodepoint`

r00ster91:
I noticed that this function was repeating code that was already written before: https://github.com/ziglang/zig/blob/6ae8fe193bd5397b9fceea98ba684d9d870286cf/lib/std/unicode.zig#L71

### DIFF
--- a/lib/std/unicode.zig
+++ b/lib/std/unicode.zig
@@ -269,14 +269,7 @@ pub const Utf8Iterator = struct {
 
     pub fn nextCodepoint(it: *Utf8Iterator) ?u21 {
         const slice = it.nextCodepointSlice() orelse return null;
-
-        switch (slice.len) {
-            1 => return @as(u21, slice[0]),
-            2 => return utf8Decode2(slice) catch unreachable,
-            3 => return utf8Decode3(slice) catch unreachable,
-            4 => return utf8Decode4(slice) catch unreachable,
-            else => unreachable,
-        }
+        return utf8Decode(slice) catch unreachable;
     }
 
     /// Look ahead at the next n codepoints without advancing the iterator.


lib/std/unicode.zig
```diff
@@ -269,14 +269,7 @@ pub const Utf8Iterator = struct {
 
     pub fn nextCodepoint(it: *Utf8Iterator) ?u21 {
         const slice = it.nextCodepointSlice() orelse return null;
-
-        switch (slice.len) {
-            1 => return @as(u21, slice[0]),
-            2 => return utf8Decode2(slice) catch unreachable,
-            3 => return utf8Decode3(slice) catch unreachable,
-            4 => return utf8Decode4(slice) catch unreachable,
```
matu3ba, line 277: the error conditions are here assumed to be unreachable, which is used for ReleaseFast and ReleaseSmall https://ziglang.org/documentation/master/#unreachable

For example `pub fn utf8Decode2(bytes: []const u8) Utf8Decode2Error!u21` has 2 checks
`if (bytes[1] & 0b11000000 != 0b10000000) return error.Utf8ExpectedContinuation;` and `if (value < 0x80) return error.Utf8OverlongEncoding;`, which are optimized away with the current approach.

As how I understand the semantics of unreachable, Your change will remove this optimization.

At least how I understand it, `utf8Decode2` does not need to be reexecuted for nextCodePointSlice, since that is what `utf8ValidateSlice` is exactly used for:
```zig
/// Utf8View is used to iterate the code points of an utf-8 encoded string.
/// Use `init()` to check validity of code points. Otherwise, use `initUnchecked()`.
///
/// ```
/// var utf8 = (try std.unicode.Utf8View.init("hi there")).iterator();
/// while (utf8.nextCodepointSlice()) |codepoint| {
///   std.debug.print("got codepoint {}\n", .{codepoint});
/// }
/// ```
```

Can you clarify?

r00ster91: > which are optimized away with the current approach.

Yes, with my approach as well. This is due to the `catch unreachable` which will catch *all* the errors and mark them as unreachable, which allows the compiler to optimize all the error conditions out.

```zig
pub fn utf8Decode(bytes: []const u8) Utf8DecodeError!u21 {
    return switch (bytes.len) {
        1 => @as(u21, bytes[0]),
        2 => utf8Decode2(bytes), // propagated
        3 => utf8Decode3(bytes), // propagated
        4 => utf8Decode4(bytes), // propagated
        else => unreachable, // immediately marked as unreachable
    };
}
```
All the propagated errors here are later marked as unreachable using the `catch unreachable` after the call to `utf8Decode`.

> At least how I understand it, `utf8Decode2` does not need to be reexecuted for `nextCodePointSlice`, since that is what utf8ValidateSlice is exactly used for:

I'm a bit confused about this second part. Is this a suggestion that could be applied to the code of `nextCodepoint` before this PR? If so, are you suggesting to remove the `2` prong in the `switch`?

matu3ba: > which will catch all the errors and mark them as unreachable

Yes, makes sense.

> I'm a bit confused about this second part.

The second part elaborated on the error condition, which dont apply here (because before the whole slice/array was validated).

Mark this as resolved.

matu3ba (approved):
LGTM.

PR #11406, opened 2022-04-08
Remove primitive values from keyword reference

scheibo:
5a53ab28 removed these as keywords, and the Primitive Values section
of the docs already exists to describe them.


ifreund (approved):
LGTM, Thanks!

Issue #11391, opened 2022-04-06
function call eats `f80` on freebsd


viriuwu:
### Steps to Reproduce

```zig
const std = @import("std");

pub fn main() void {
    std.debug.print("one: {}\n", .{ @as(f80, 1.0) });
}
```

### Expected Behavior

> one: 1.0e+00

### Actual Behavior

> one: 0.0e+00

It should be noted this isn't zero but a very small subnormal, the raw bits are `0000000000000000000000000000000000000000000000000011111111111111` if I try to print them out inside of a function call (may be garbage). The correct `1.0e+00` gets printed on Linux and Windows.

### FreeBSD & Zig Version

FreeBSD: 13.0 amd64
Zig: 0.10.0-dev.1727+ac873367b

Found this while investigating a [failed test on lavahut](https://builds.hut.lavatech.top/~andrewrk/job/22470) for #10133.

PR #11407, opened 2022-04-09
compiler_rt: Add `f80` support to `floatXiYf`/`fixXfYi`

topolarity:
This change:
 - Adds a generic implementation of the float -> integer conversion (`floatXiYf`), including support for f80
 - Updates the existing implementation of integer -> float conversion (`fixXfYi`) to support f16 and f80
 - Fixes the handling of the explicit integer bit in `__trunctfxf2`
 - Combines the test cases for `fixXfYi`/`floatXiYf` into a single file
 - Renames `fmodl` to `fmodq`, since it operates on 128-bit floats

The new implementation for floatXiYf has been benchmarked on my x86-64 machine:

#### `floatXiYf` Throughput (MiB/s) - Before
|    |    u32   |    i32   |    u64   |    i64   |   u128   |   i128   |
|-----|----------|----------|----------|----------|----------|----------|
| f16 |     none |     none |     none |     none |     none |     none |
| f32 |  2231.67 |  2001.19 |  1745.66 |  1405.77 |  2173.99 |  1874.63 |
| f64 |  1407.17 |  1055.83 |  2911.68 |  2437.21 |  1676.05 |  1476.67 |
| f80 |     none |     none |     none |     none |     none |     none |
|f128 |   327.56 |   321.25 |   645.92 |   654.52 |  1153.56 |  1096.27 |

#### `floatXiYf` Throughput (MiB/s) - After
 |   |    u32   |    i32   |    u64   |    i64   |   u128   |   i128   |
|-----|----------|----------|----------|----------|----------|----------|
| f16 |  1407.61 |  1637.25 |  3555.03 |  2594.56 |  3680.60 |  3063.34 |
| f32 |  2101.36 |  2122.62 |  3225.46 |  3123.86 |  2860.05 |  1985.21 |
| f64 |  1395.57 |  1314.87 |  2409.24 |  2196.30 |  2384.95 |  1908.15 |
| f80 |   475.53 |   457.92 |   884.50 |   812.12 |  1475.27 |  1382.16 |
|f128 |   359.60 |   350.91 |   723.08 |   706.80 |  1296.42 |  1198.87 |

`fixXfYi` performance is about the same as before, since only minor modifications to the existing generic implementation were needed. Full benchmarks are [available here](https://gist.github.com/topolarity/08e3f0aa6acdf00640edd0c5aa26b128).


topolarity:
Unfortunately, this does deactivate 3 tests for `@rem`/`@mod` on stage2. The problem is that LLVM incorrectly lowers the 128-bit operations to `fmodl`. I'm planning to follow up soon with another change that will lower these operations to the correct library calls, making these tests pass again.

viriuwu:
worth pointing out that #11391 is a thing at the moment and lavahut will not be happy with f80 tests

matu3ba:
cool and big change. 

I have several question regarding benchmarks:
1. can you provide the benchmark scripts? I would like to peek at the input sizes, numbers etc for below given reason
2. f32/u32 has lower throughput: 2231.67 -> 2101.36
3. f64/u64 also has lower throughput: 2911.68 -> 2409.24
4. f64/i64 also: 2437.21 -> 2196.30

It would be interesting, if thermal reasons are the cause or set of data was too small etc.
The change of the throughput is also significant more than a few %, so I am curious on the reason.

topolarity:
> worth pointing out that #11391 is a thing at the moment and lavahut will not be happy with f80 tests

Thanks for the heads up! That test case saved me a ton of time debugging.

I also think I was able to find the root cause, and I've pushed a fix :+1:

topolarity:
> It would be interesting, if thermal reasons are the cause or set of data was too small etc. The change of the throughput is also significant more than a few %, so I am curious on the reason.

I think that's because the previous implementations got to "cheat" a bit:
  - floatsiXf has a few rounding bugs (0x1FFFFFE is exactly representable but is rounded to 0x2000000, etc.)
  - floatdidf uses floating point operations internally. Since most relevant targets won't have hardware FPU support but x86-64 does, this gives it a big artificial boost in the benchmark.
  
The [benchmark script is here](https://gist.github.com/topolarity/ca760ca7fc660373f90aa81e6560efd2) and can be run with `zig run benchmark.zig -O ReleaseFast --zig-lib-dir lib -- 26214`

I did notice significant variation run-to-run, and I'm curious to hear if you find a way to tighten up the results.

matu3ba:
I played around with the benchmark script and the results are half-consistent. I think the benchmark script itself is abit flawed, as you want to measure the time for 1 or multiple operations.

However, `_ = timer.lap();` is called only once in the loop, whereas also stuff like
```zig
                for (in) |*val| {
                    val.* = rnd.random().int(Z);
                }
```
is run within the same loop.

Is there a particular reason, why stuff like https://gist.github.com/matu3ba/f848765e3c67ee74781b57858049d692#file-simplebench_addo-zig-L90 (but for floats) will not work as rough estimate?

This is much simpler than doing random number stuff, which should be called at comptime to not falsify the results (or the timer would need to be stopped and started only for the measurement).

Did you measure how much floating operations deviate in between different numbers to justify using of `random()`, ie by repeating the same floating operations 10_000 times and then another one to see the average performance?

topolarity:
I'm sure non-random benchmark data could work too. This is just the benchmark I happened to throw together.

FWIW, the measurement is between `timer.lap()` and `timer.read()`, which should not include random number gen.

andrewrk (approved):
Fantastic work. Thank you especially for providing the benchmark measurements.

Looks like I caused some conflicts for you by merging other pull requests first. Apologies for that. Do you mind doing the rebase?

lib/std/special/compiler_rt/fixXfYi.zig
```diff
@@ -0,0 +1,224 @@
...
+pub inline fn fixXfYi(comptime I: type, a: anytype) I {
+    @setRuntimeSafety(is_test);
+
+    const F = @TypeOf(a);
+    const float_bits = @typeInfo(F).Float.bits;
+    const int_bits = @typeInfo(I).Int.bits;
+    const rep_t = std.meta.Int(.unsigned, float_bits);
+    const sig_bits = math.floatMantissaBits(F);
+    const exp_bits = math.floatExponentBits(F);
+    const fractional_sig_bits = sig_bits - @boolToInt(F == f80);
```
viriuwu, line 15: There's an API for this condition as of 5b8ac9821dd25c3e5282130b4d93d6c5b7debb08

https://github.com/ziglang/zig/blob/17631cb2d30bf2d7b10401cf8f784a599bade5c5/lib/std/math/float.zig#L46

topolarity: Oh excellent - I'll use this to make the intent clearer, thanks

The `floatMantissaDigits`/`floatMantissaBits` naming is a bit confusing, isn't it? I wonder if `floatFractionalMantissaBits` would be a better function to provide, even if it is long-winded

viriuwu: I carried it over from what it's called in other languages but I was thinking too after posting this that maybe it could have a better name, I was thinking `floatFractionalBits` to match the rest of them (returning 1 less than it does right now).

viriuwu (approved):

lib/std/special/compiler_rt/fixXfYi.zig
topolarity: Perfect. Thanks for the input, I think the new name is much clearer

lib/std/math/float.zig
```diff
@@ -43,18 +43,18 @@ pub fn floatMantissaBits(comptime T: type) comptime_int {
 }
 
 /// Returns the number of binary digits in the mantissa of floating point type T.
```
viriuwu, line 45: I feel like this docstring mismatches the new function since it no longer includes the integer part.

topolarity: You're right, good catch. Let me fix that up real quick.

viriuwu (approved):

PR #11410, opened 2022-04-09
Add `std.fs.File.sync`

r00ster91:
This is a simple wrapper around `std.os.fsync` for convenience. It seemed like this could be added.

I thought it was simple enough that no test for this was needed in https://github.com/ziglang/zig/blob/master/lib/std/fs/test.zig and there's already a test [here](https://github.com/ziglang/zig/blob/f654e16d06460a5f44fd8089538a41fc1936ea0b/lib/std/os/test.zig#L678) so I hope it's okay.

### DIFF
--- a/lib/std/fs/file.zig
+++ b/lib/std/fs/file.zig
@@ -199,6 +199,14 @@ pub const File = struct {
             os.close(self.handle);
         }
     }
+    
+    pub const SyncError = os.SyncError;
+
+    /// Blocks until all pending file contents and metadata modifications
+    /// have been synchronized with the underlying filesystem.
+    pub fn sync(self: File) SyncError!void {
+        return os.fsync(self.handle);
+    }
 
     /// Test whether the file refers to a terminal.
     /// See also `supportsAnsiEscapeCodes`.


matu3ba:
zig fmt fails:
```txt
+ /workspace/_debug/staging/bin/zig fmt --check . --exclude test/compile_errors/
127s
263./_debug/staging/lib/zig/std/fs/file.zig
127s
264./lib/std/fs/file.zig
```


Can you add a brief explanation that **only file contents are synced**, but directory metadata not necessarily?
see https://unix.stackexchange.com/questions/414749/how-does-fsync-treat-directory-links
https://github.com/google/renameio/issues/11 and https://man7.org/linux/man-pages/man2/fdatasync.2.html

Ideally, you would add what one should use to ensure that directory metadata are updated as well in 1 short sentence.

r00ster91:
> zig fmt fails:

Ah, thanks. I took a look at the logs but wasn't quite sure what was going on.

I took a closer look at this whole thing and realized that there's another relevant function here: `fdatasync`. I added it as a faster alternative to the `fsync` wrapper.

r00ster91:
@matu3ba 
> Ideally, you would add what one should use to ensure that directory metadata are updated as well in 1 short sentence.

I think you would use `std.os.fsync` or `std.os.fdatasync` on the `fd` of your `std.fs.Dir`, right? Shouldn't we have something like a `sync` for `std.fs.Dir` too?

matu3ba:
> Shouldn't we have syncAll and syncContent (or only sync?) for std.fs.Dir too?

Personally, I dont see the advantage of having a redundant function that only exposes the syscall.
If you want `syncAll` as "convenient, but slower", then it must use `fsync` for both the file handle and the directory handle (both provided by the user.

`fdatasync` is unused on current Linux (does the same thing as `fsync`), so we dont want to expose wrong semantic guarantees to users https://www.informit.com/articles/article.aspx?p=23618&seqNum=5

r00ster91:
> Personally, I dont see the advantage of having a redundant function that only exposes the syscall.

Was this comment paragraph also meant for the functions that this PR adds? There are other such examples (in `std.fs.File` and elsewhere) where very much only a single syscall is exposed like this so I thought it'd be okay. If so, I don't see why it wouldn't be okay to do for `std.fs.Dir` too. Are you fundamentally against adding this?

> `fdatasync` is unused on current Linux (does the same thing as `fsync`), so we dont want to expose wrong semantic guarantees to users https://www.informit.com/articles/article.aspx?p=23618&seqNum=5

It's interesting that Linux is in fact one of the systems where both functions behave the same. Yes, I've read about that too, like in the Rust docs for example: https://doc.rust-lang.org/std/fs/struct.File.html#method.sync_data
> Note that some platforms may simply implement this in terms of [sync_all](https://doc.rust-lang.org/std/fs/struct.File.html#method.sync_all).

I don't the functions added here expose any wrong semantic guarantees to users because it's not a guarantee: it says that "it does not necessarily synchronize the file's metadata" but still could. I suppose I could add the "but still could" to make it clearer.

matu3ba:
> Was this comment paragraph also meant for the functions that this PR adds?

Yes and no. Personally, I think the docs should describe which functions to use instead of re-exposing them. However, in light of no convention etc, the change is fine. I should have been more accurate.

> it says that "it does not necessarily synchronize the file's metadata" but still could

With that in mind, I think the change is okay.
The second paragraphs can be simplified by mentioning `see also syncContent/syncAll` instead of
```zig
    ///
    /// Note that this does not necessarily ensure that metadata for the directory containing the file has also reached disk.`
```

lib/std/fs/file.zig
```diff
@@ -200,6 +200,24 @@ pub const File = struct {
         }
     }
 
+    pub const SyncError = os.SyncError;
+
+    /// Blocks until all pending file contents and metadata modifications
+    /// for the file have been synchronized with the underlying filesystem.
+    ///
+    /// Note that this does not necessarily ensure that metadata for the directory containing the file has also reached disk.
```
matu3ba: nit: long sentence+line and half-redundant. better: `see syncContent`
I think the paragraph is not necessary, but thats more of a style thing.

r00ster91:
But then again, on Windows there isn't a distinction between `fsync` and `fdatasync` either. Windows also always synchronizes both content and modification time, just like Linux.
Even if `fdatasync` didn't also immediately update the modification time, there wouldn't be a big speed up so maybe it is better after all to make it one simple `std.fs.File.sync` that updates both things (which is the case on probably most platforms anyway, as we learned here) and anyone that wants something else can always go to `std.os`.
@matu3ba what do you think?

lib/std/fs/file.zig
```diff
@@ -200,6 +200,24 @@ pub const File = struct {
...
+    pub fn syncAll(self: File) SyncError!void {
+        return os.fsync(self.handle);
+    }
+
+    /// This is the same as `syncAll` except that it does not necessarily synchronize the file's metadata.
+    /// Use this if you want to make sure that at least content is stored on disk and metadata is secondary.
+    ///
+    /// Note that this does not necessarily ensure that metadata for the directory containing the file has also reached disk.
```
matu3ba: same argument `see syncAll`

matu3ba (approved):
some nits.

matu3ba:
> what do you think about keeping only syncAll and naming it sync

That sounds better to me.

r00ster91:
@matu3ba 
Yes, it seems like this would work if that's what you mean:
```zig
    var relative_path: [std.fs.MAX_PATH_BYTES]u8 = undefined;
    const absolute_path = try std.fs.realpath("src", &relative_path);
    const directory = try std.fs.openDirAbsolute(absolute_path, .{});
    const file = std.fs.File{ .handle = directory.fd };
    file.close();
```
But one could also add `sync` to `std.fs.Dir`.

matu3ba:
> But one could also add sync to std.fs.Dir.

I think this would be nice, but this can also be part of another PR.

r00ster91:
Yes, I agree. I feel like this is good enough as-is. If someone needs it someday, they can add it and possibly link to it from these docs too if they want. I feel like people generally don't really do this if they just update a file inside the directory.

andrewrk:
Thanks!

PR #11412, opened 2022-04-11
std.event.WaitGroup: fix compilation (acquire->lock, release->unlock)

slimsag:
In 008b0ec5e58fc7e31f3b989868a7d1ea4df3f41d the `std.Thread.Mutex` API was changed from `acquire`/`release` to `lock`/`unlock`. `std.event.Lock` still uses `acquire`/`release`.

`std.event.WaitGroup` is using `std.Thread.Mutex`, though, not `std.event.Lock`, and it was not updated to use `lock`/`unlock`, and so compilation failed prior to this change.

### DIFF
--- a/lib/std/event/wait_group.zig
+++ b/lib/std/event/wait_group.zig
@@ -35,8 +35,8 @@ pub fn WaitGroupGeneric(comptime counter_size: u16) type {
 
         const Self = @This();
         pub fn begin(self: *Self, count: CounterType) error{Overflow}!void {
-            const held = self.mutex.acquire();
-            defer held.release();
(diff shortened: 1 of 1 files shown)


andrewrk:
Thanks for the fix! Are you using evented I/O for something?

slimsag:
Just using it for a few different one-off/hack projects. I have a little text search indexer I'm working on, and during querying I need some simple M:N thread scheduling to parallelize work, `std.event.Loop.instance.?.runDetached` fits the bill so I'm using that - but I needed waitgroup synchronization to work.

Issue #8362, opened 2021-03-26
Windows: Access denied while building project initialised with build-exe


Einh06:
Tested with`zig version: 0.7.1` and `zig version: 0.8.0-dev.1548+0d96a284e`
I'm currently running into an issue building on Windows where running `zig build` on a project generated with `zig init-exe` fails with `error: unable to build stage1 zig object: AccessDenied`. I ran the command both as user and administrator, to no avail.
The permissions on all folders also seem correct (`%LOCALAPPDATA%\zig` has all permisions, so does my directory with the zig binaries, and the project directory).
`zig build-exe main.zig` also fails, but `zig run main.zig` works.

csharpdf:
I just started learning Zig, and after trying to build a basic app I got this same issue. I tried both the current master branch, `zig-windows-x86_64-0.9.0-dev.1610+e5bc09240`, and the current stable release, `0.8.1`. I'm on Windows 10, and also ran `zig build` on an app initialized with `zig init-exe`. I'd really like to see this issue resolved so I can start working with the language.

csharpdf:
Update: It seems that `zig run src/main.zig` works for the default `zig init-exe` example, but again, the `zig build (either build.zig, src/main.zig, or no param)` still results in an error. This may help a bit @andrewrk

aotmr:
I'm seeing the same issue myself on the stable release, `0.8.1`, both for `zig build run` and `zig run src/main.zig`. This happens both for Powershell and CMD alike from within Visual Studio Code. The issue is intermittent--if I re-try the build, eventually it succeeds.

csharpdf:
It does not work, and even if it did, it's still a major issue. You should be able to build programs on the first attempt with no problems. I sincerely hope that this issue gets fixed. If I could help, I would.

aotmr:
@csharpdf I've been using the Windows builds, extracted to a folder on my drive and added to the PATH. I will see if the problem still exists with the new release.

aotmr:
I'm afraid I get the same issue with 0.9.0, but with a perhaps better error message:
```
error: unable to generate DLL import .lib file: AccessDenied
```

marler8997:
Doesn't reproduce on my machine:
```
C:\temp\b>zig init-exe
info: Created build.zig
info: Created src\main.zig
info: Next, try `zig build --help` or `zig build run`

C:\temp\b>zig build

C:\temp\b>zig version
0.9.0-dev.1815+20e19e75f
```

[...]

Issue #6364, opened 2020-09-17
Windows: rename fails for directories


squeek502:
Related to https://github.com/ziglang/zig/issues/6344 and https://github.com/ziglang/zig/issues/6344

> * On Windows, the implementation of both `renameAbsolute` and `Dir.rename` do not allow renaming directories.
>   * `renameAbsolute` [uses `MOVEFILE_REPLACE_EXISTING`](https://github.com/ziglang/zig/blob/61e9e82bdc10110b74bdeb973cc542c7b73a4ae2/lib/std/os.zig#L1968) which ["cannot be used if lpNewFileName or lpExistingFileName names a directory."](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexw)
>   * `Dir.rename` [uses `OpenFile`](https://github.com/ziglang/zig/blob/61e9e82bdc10110b74bdeb973cc542c7b73a4ae2/lib/std/os.zig#L2066-L2074) which fails with `error.IsDir`

Some more info about what I tried and ultimately abandoned:

Swapping out https://github.com/ziglang/zig/blob/2962be81359a6806f66d220474abcb0b1cf7edf0/lib/std/os.zig#L2066-L2075

for `NtCreateFile` like so:

```zig
    const path_len_bytes = @intCast(u16, old_path_w.len * 2);
    var nt_name = windows.UNICODE_STRING{
        .Length = path_len_bytes,
        .MaximumLength = path_len_bytes,
        // The Windows API makes this mutable, but it will not mutate here.
        .Buffer = @intToPtr([*]u16, @ptrToInt(old_path_w.ptr)),
    };

    var attr = windows.OBJECT_ATTRIBUTES{
        .Length = @sizeOf(windows.OBJECT_ATTRIBUTES),
        .RootDirectory = if (std.fs.path.isAbsoluteWindowsWTF16(old_path_w)) null else old_dir_fd,
        .Attributes = 0, // Note we do not use OBJ_CASE_INSENSITIVE here.
        .ObjectName = &nt_name,
        .SecurityDescriptor = null,
        .SecurityQualityOfService = null,
    };
    var io: windows.IO_STATUS_BLOCK = undefined;
    var src_fd: windows.HANDLE = undefined;
    var rc = windows.ntdll.NtCreateFile(
        &src_fd,
        windows.SYNCHRONIZE | windows.DELETE,
        &attr,
        &io,
        null,
        0,
[...]

PR #11413, opened 2022-04-11
std: `os.renameatW` properly fail with `PathAlreadyExists`

jagt:
Before this fix `os.renameatW` never fail with `PathAlreadyExists`. When dest path exists it fails with `AccessDenied`.
To me this causes most of my `zig build` stuck trouble described in here: <https://github.com/ziglang/zig/issues/8362>

A simple fixture that demonstrate current behavior:

```zig
const std = @import("std");
const testing = std.testing;

test "rename on windows" {
    const cwd = std.fs.cwd();

    cwd.makeDir("foo") catch {};
    cwd.makeDir("bar") catch {};

    try testing.expectError(error.AccessDenied, std.fs.rename(cwd, "foo", cwd, "bar"));
}
```

After this commit the test should be like this:

```zig
    try testing.expectError(error.PathAlreadyExists, std.fs.rename(cwd, "foo", cwd, "bar"));
```

I believe this should mitigate the `zig build` AccessDenied problem quite a lot on windows.

The concerning part is that this change added a new system call on the `AccessDenied` error code path. 

Another way to address the zig build issue is to change the behavior of `renameTmpIntoCache` so that it checks dest dir befor rename on windows.

Thanks in advance!

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -2694,7 +2694,28 @@ pub fn renameatW(
         .INVALID_HANDLE => unreachable,
         .INVALID_PARAMETER => unreachable,
         .OBJECT_PATH_SYNTAX_BAD => unreachable,
-        .ACCESS_DENIED => return error.AccessDenied,
+        .ACCESS_DENIED => {
+            const new_path_len_bytes = math.cast(u16, rename_info.FileNameLength) catch |err| switch (err) {
+                error.Overflow => return error.AccessDenied,
+            };
+            var nt_name = windows.UNICODE_STRING{
+                .Length = new_path_len_bytes,
+                .MaximumLength = new_path_len_bytes,
+                .Buffer = @as([*]u16, &rename_info.FileName),
+            };
+            var attr = windows.OBJECT_ATTRIBUTES{
+                .Length = @sizeOf(windows.OBJECT_ATTRIBUTES),
+                .RootDirectory = new_dir_fd,
+                .Attributes = 0, // Note we do not use OBJ_CASE_INSENSITIVE here.
+                .ObjectName = &nt_name,
+                .SecurityDescriptor = null,
+                .SecurityQualityOfService = null,
+            };
+            var basic_info: windows.FILE_BASIC_INFORMATION = undefined;
+            const status = windows.ntdll.NtQueryAttributesFile(&attr, &basic_info);
+            if (status == .SUCCESS) return error.PathAlreadyExists;
+            return error.AccessDenied;
+        },
         .OBJECT_NAME_NOT_FOUND => return error.FileNotFound,
         .OBJECT_PATH_NOT_FOUND => return error.FileNotFound,
         .NOT_SAME_DEVICE => return error.RenameAcrossMountPoints,


squeek502:
Note that rename for directories is broken more generally on Windows (#6364), so this could be a misdiagnosis of the problem you're encountering.

Also, this is different than how non-Windows platforms handle this particular rename, as on Linux renaming a dir onto an empty dir succeeds, and PathAlreadyExists only occurs if the destination is a non-empty directory:

```zig
const std = @import("std");
const testing = std.testing;

test "rename onto empty dir" {
    var tmp_dir = testing.tmpDir(.{});
    defer tmp_dir.cleanup();

    try tmp_dir.dir.makeDir("test_dir");
    try tmp_dir.dir.makeDir("target_dir");
    try tmp_dir.dir.rename("test_dir", "target_dir");

    // Ensure the directory was renamed
    try testing.expectError(error.FileNotFound, tmp_dir.dir.openDir("test_dir", .{}));
    var dir = try tmp_dir.dir.openDir("target_dir", .{});
    dir.close();
}

test "rename onto non-empty dir" {
    var tmp_dir = testing.tmpDir(.{});
    defer tmp_dir.cleanup();

    try tmp_dir.dir.makeDir("test_dir");

    var target_dir = try tmp_dir.dir.makeOpenPath("target_dir", .{});
    var file = try target_dir.createFile("test_file", .{ .read = true });
    file.close();
    target_dir.close();

    // Rename will fail with PathAlreadyExists if target_dir is non-empty
    try testing.expectError(error.PathAlreadyExists, tmp_dir.dir.rename("test_dir", "target_dir"));

    // Ensure the directory was not renamed
    var dir = try tmp_dir.dir.openDir("test_dir", .{});
    dir.close();
}
```
On Linux gives:

```
$ zig test rename.zig
All 2 tests passed.
```

There's also some test code in the Zig std library that tests some of this stuff (but is currently skipped on Windows):

https://github.com/ziglang/zig/blob/d2681d2537b630359657e117ed77cf12a15bb7df/lib/std/fs/test.zig#L436-L477

This also ties into https://github.com/ziglang/zig/issues/6600, since it's still slightly unclear how these types of platform differences should be handled by `std.os`.

jagt:
Thanks for the info! For the `rename onto empty dir` case it's pretty surprising.
I did noticed that there's a previous cl <https://github.com/ziglang/zig/commit/d3f87f8ac01039722197a13a12342fc747a90567> that explicitly allow `fs.rename` work on dir on windows.

I do agree that make fs things work is tricky. I'll fire for a separated PR to hack around `renameTmpIntoCache` for now since that's the thing I'm trying to fix.

jagt:
I've opened <https://github.com/ziglang/zig/pull/11419> in which addresses the same issue.

andrewrk:
Instead of solving this problem by adding more syscalls on Windows to try to match the POSIX interface, I would like to instead solve this by making `std.os.renameat` be a compile error on Windows. Any code paths that call `std.os.renameat` will have to be updated to have conditional compilation for Windows and handle it differently. This way Windows users will still get the ideal experience which does not make unnecessary syscalls.

I think this is a perfect example of what #6600 is intended to address.
See also: #5019

squeek502:
> Instead of solving this problem by adding more syscalls on Windows to try to match the POSIX interface, I would like to instead solve this by making std.os.renameat be a compile error on Windows. Any code paths that call std.os.renameat will have to be updated to have conditional compilation for Windows and handle it differently. This way Windows users will still get the ideal experience which does not make unnecessary syscalls.

Does this mean that something like`std.fs.Dir.rename` would have different behavior per-platform? Or would the idea be to make `std.fs.Dir.rename` not necessarily conform to POSIX `rename` either, but instead find some common-ground functionality that works efficiently-enough for all platforms? Or would this rule out `rename` from being able to be part of `std.fs` at all, and instead users would need to call the os-specific functions and handle the differences themselves?

andrewrk:
Different behavior per platform is one possible resolution to the problem. I would describe the design space like this:
* If there is a behavior definition that we can choose that lowers to a single syscall on all platforms and the differences are minor enough, then this is a nice approach. We can see this for a few functions in std.fs which document that they are atomic on some platforms and not on others.
* Otherwise, the next strategy is to make the functions more specific in their semantics. Then they will lower to one syscall on some platforms and multiple on others. The user will call the one with the matching semantics they need and get the ideal lowering for each platform.

For the case of rename, the second approach seems like it might be the way to go to me.

PR #11416, opened 2022-04-11
std/fs/test.zig: Add test for renaming a dir onto an empty dir

squeek502:
Realized this case wasn't explicitly tested in the current test code when writing https://github.com/ziglang/zig/pull/11413#issuecomment-1095571208.

Also split the "Dir.rename directories" test into 3 tests:
- General rename of a directory
- Rename of a directory onto an existing empty directory
- Rename of a directory onto an existing non-empty directory

The only actually new case here is the "rename onto an existing empty directory", but splitting the tests this way made them much more understandable.


andrewrk:
Thanks!

Issue #11288, opened 2022-03-24
proposal: improved compiler test suite (compiler errors, run and compare, translate-c, etc.)


mitchellh:
Requirement for #89 

# Proposal

An improved test harness for the compiler that supports the following:

- Utilizes a directory structure of real source files (Zig, C, and assembly) to define compiler test cases.
- Test settings and assertions are configured using a manifest file. The manifest file can be embedded in the first (top) comment of a Zig file or it can exist alongside a test file.
- Multiple test types:
    - `run` - Build and run to completion and compare output
    - `error` - Build and expect specific error output
    - `translate-c` - Build a “.c” file and assert an expected “zig” file is produced. The Zig file can further optionally define a subsequent test to run (i.e. `run`) so you can translate C then optionally assert the proper behavior also happens.
    - `incremental` - Build and incrementally update a program. Each incremental step is itself a test case so that we can support incremental updates introducing errors, and a subsequent update fixing the errors. [stage2-only]
    - `cli` - Run a specific `zig` CLI command (typically, `zig build test`) to test the Zig CLI or Zig build machinery.
- Concurrency. Test cases will run on a thread pool similar to compilation.
- Test filtering by name, test type, or backend.
- Test dependencies. Tests can note that they are dependent on another test in order to optimize parallelism or build order.
- Adding or modifying tests will not trigger a full rebuild of the compiler.

This will be built on the existing `test-stage2` harness (`src/test.zig`) and won’t a totally new implementation. This is almost purely iteratively improving the existing harness rather than writing anything new. 

The tests will continue to invoked using `zig build test-stage2` or similar targets.

**NOTE:** Some of these are not new features, but are features that are currently unimplemented in the stage2 test harness or just need slight improvement in the stage2 test harness.

## Backwards Compatibility

A limited amount of backwards compatibility with the existing compiler test API will be retained so we can incrementally introduce these test cases. In some cases, certain test types such as stack trace tests and certain native backend tests will continue to use the Zig API and *will not* support the file-based definition format. Therefore, we will continue to support the necessary subset of the Zig API.

Note that changes to these tests will not gain the benefits of the above harness. For example, changes will trigger a full rebuild. They won’t be parallelized as well, etc. However, these tests tend to have very few cases and/or are updated very infrequently.

# Test Definitions

Tests are now defined using files within a directory structure rather than a Zig API. This has benefits:

1. Adding, changing, or removing a test case does not cause the full test harness (including the compiler) to rebuild.
[...]

PR #11417, opened 2022-04-11
Add file support for incremental compile error tests

topolarity:
Compile error test cases can now be given as a sequence of files: `foo.1.zig, foo.2.zig, foo.3.zig, etc...`

As recommended in #11288, "foo.zig" may not co-exist with "foo.X.zig", and no numbers may be skipped in the test sequence. This implementation also has a couple of slight differences:
  - Test indices are 1-indexed
  - A suffix of ".X.zig" is used instead of "_X.zig"

1-indexing was chosen so that it will match update progress number "[x/N]" reported in the log. I've also updated a few other places where we report the update number, to make sure it is consistently 1-indexed.

If you'd prefer to make this 0-indexed or to use the original `_` suffix, let me know and I'm happy to update the changes :+1:

### DIFF
--- a/src/test.zig
+++ b/src/test.zig
@@ -720,14 +720,17 @@ pub const TestContext = struct {
                     // Move to beginning of line
                     while (cursor > 0 and src[cursor - 1] != '\n') cursor -= 1;
 
-                    // Check if line is non-empty and does not start with "//"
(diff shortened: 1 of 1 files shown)


andrewrk:
Looks great!

I would like to take you on your offer to make them zero-indexed. Mind if I hit merge and then that can be a follow-up PR along with (I'm guessing) migration of existing incremental compilation test cases to this new system?

PR #11419, opened 2022-04-12
fix `link.renameTmpIntoCache` on windows when dest dir exists.

jagt:
This should mitigate <https://github.com/ziglang/zig/issues/8362> on Windows.

I previously have another PR trying to fix it in `renameatW` <https://github.com/ziglang/zig/pull/11413>. I figured if that change is a bit risky then this one should be a good enough workaround for now. As comment suggested `renameTmpIntoCache` logic is temporary and I think it's ok to add platform workaround in here?

I've locally tested this a bit and it seems to work for me. I'll locally test this with <https://github.com/ziglang/zig/pull/11365> and see if it fixes zig build related issues on Windows.

Thanks in advance!

### DIFF
--- a/src/link.zig
+++ b/src/link.zig
@@ -717,19 +717,38 @@ pub const File = struct {
         // directly, and remove this function from link.zig.
         _ = base;
         while (true) {
-            std.fs.rename(
-                cache_directory.handle,
-                tmp_dir_sub_path,
-                cache_directory.handle,
-                o_sub_path,
-            ) catch |err| switch (err) {
-                error.PathAlreadyExists => {
+            if (builtin.os.tag == .windows) {
+                // workaround windows `renameW` can't fail with `PathAlreadyExists`
+                // See https://github.com/ziglang/zig/issues/8362
+                if (cache_directory.handle.access(o_sub_path, .{})) |_| {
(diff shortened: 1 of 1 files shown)


src/link.zig
```diff
@@ -717,19 +717,37 @@ pub const File = struct {
...
         while (true) {
-            std.fs.rename(
-                cache_directory.handle,
-                tmp_dir_sub_path,
-                cache_directory.handle,
-                o_sub_path,
-            ) catch |err| switch (err) {
-                error.PathAlreadyExists => {
+            if (builtin.os.tag == .windows) {
+                // workaround windows `renameW` can't fail with `PathAlreadyExists`
```
squeek502, line 721: Might be worth adding a link to https://github.com/ziglang/zig/issues/8362 and/or https://github.com/ziglang/zig/issues/6364

```suggestion
                // workaround windows `renameW` can't fail with `PathAlreadyExists`
                // see https://github.com/ziglang/zig/issues/8362
                // TODO: https://github.com/ziglang/zig/issues/6364
```

andrewrk:
Thank you for this pull request and the other one that I closed. This looks like a perfectly reasonable mitigation until #8362 is properly resolved via breaking standard library changes.

PR #11421, opened 2022-04-12
Use 0-indexing for incremental compile error tests

topolarity:
Follow-up to #11417.

I went ahead and reverted the reported `update_index` to also be 0-indexed.

P.S. I took a look to move over any existing incremental tests, but I don't think we have any yet for stage2

### DIFF
--- a/src/test.zig
+++ b/src/test.zig
@@ -656,10 +656,10 @@ pub const TestContext = struct {
 
     const Strategy = enum {
         /// Execute tests as independent compilations, unless they are explicitly
-        /// incremental ("foo.1.zig", "foo.2.zig", etc.)
+        /// incremental ("foo.0.zig", "foo.1.zig", etc.)
         independent,
         /// Execute all tests as incremental updates to a single compilation. Explicitly
-        /// incremental tests ("foo.1.zig", "foo.2.zig", etc.) still execute in order
+        /// incremental tests ("foo.0.zig", "foo.1.zig", etc.) still execute in order
         incremental,
     };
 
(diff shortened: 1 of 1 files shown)


andrewrk:
We do! Here are some, for example:

https://github.com/ziglang/zig/blob/99657dca1fac04e5015203699e3eb0d656d66b09/test/stage2/aarch64.zig#L61-L113

https://github.com/ziglang/zig/blob/99657dca1fac04e5015203699e3eb0d656d66b09/test/stage2/arm.zig#L45-L91

https://github.com/ziglang/zig/blob/99657dca1fac04e5015203699e3eb0d656d66b09/test/stage2/cbe.zig#L12-L57

https://github.com/ziglang/zig/blob/99657dca1fac04e5015203699e3eb0d656d66b09/test/stage2/llvm.zig#L31-L54

https://github.com/ziglang/zig/blob/99657dca1fac04e5015203699e3eb0d656d66b09/test/stage2/plan9.zig#L21-L54

https://github.com/ziglang/zig/blob/99657dca1fac04e5015203699e3eb0d656d66b09/test/stage2/wasm.zig#L10-L62

https://github.com/ziglang/zig/blob/99657dca1fac04e5015203699e3eb0d656d66b09/test/stage2/x86_64.zig#L24-L65

topolarity:
Oh sweet, I missed those. Let me see about moving them over later today.

topolarity:
Hmm, I think we'll have to implement more of the enhanced test suite proposal before these are ready to be moved over. They all either involve "run" tests or include specific targets/backends, neither of which are supported in manifests yet.

Feel free to ping me if I missed any other tests, though

andrewrk:
These were some of the earliest test cases we had for stage2. It should be possible to run most of them for all backends at this point.

PR #11425, opened 2022-04-13
std.math.float: fix f80-backed c_longdouble consts

viriuwu:
319555a6690a43de143698f13d6d107a7873dce0 introduced this change:
```diff
-    return if (T == f80) 1 << floatFractionalBits(T) else 0;
+    return if (floatMantissaDigits(T) == 64) 1 << 63 else 0;
```
However, this prevents an `f80` backed `c_longdouble` from getting the explicit integer part since it's not `== f80`. There's not a lot of test coverage for `c_longdouble` at the moment but I noticed this at a second glance so I thought I'd make a quick PR.


andrewrk:
Thanks for the follow-up!

PR #11426, opened 2022-04-13
Fix bug in PriorityQueue::removeIndex()

eshrews:
Fix to call siftDown on the removed index instead of always on index 0.
Updated test to a test that fails before and passes now.

PriorityDequeue does not have this issue.


ifreund (approved):
Hello, thanks for fixing this! Could you please run `zig fmt` on the files you changed? Other than that this looks good to me.

andrewrk:
Thanks @eshrews!

PR #11427, opened 2022-04-13
Add `std.unicode.replacement_character`

r00ster91:
This is a possible improvement I found while working on an `std.unicode.Char` and `std.unicode.Codepoint` (https://github.com/r00ster91/zig/commit/02da6dc26e011bcebc25b899f275e5d2c6c8ca44) which, after some discussion, seemed better not to pull. Instead, it's only this.

There are already 2 places within `std` where code related to this is kind of duplicated and I thought this constant could be useful to general applications as well.
For reference here is what this looks like in other languages:
* https://doc.rust-lang.org/std/primitive.char.html#associatedconstant.REPLACEMENT_CHARACTER
* https://crystal-lang.org/api/Char.html#REPLACEMENT

The problem this solves is that you don't have to memorize or look up this constant yourself. I would understand if that alone doesn't justify adding this to `std`.

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -968,10 +968,10 @@ pub fn formatUnicodeCodepoint(
     writer: anytype,
 ) !void {
     var buf: [4]u8 = undefined;
-    const len = std.unicode.utf8Encode(c, &buf) catch |err| switch (err) {
+    const len = unicode.utf8Encode(c, &buf) catch |err| switch (err) {
         error.Utf8CannotEncodeSurrogateHalf, error.CodepointTooLarge => {
-            // In case of error output the replacement char U+FFFD
-            return formatBuf(&[_]u8{ 0xef, 0xbf, 0xbd }, options, writer);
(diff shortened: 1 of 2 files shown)


viriuwu (approved):
> I would understand if that alone doesn't justify adding this to `std`.

... but it already was in `std`, in two different places! It's in one place now. LGTM, only thing is that before the result was hardcoded and now it's figured out at runtime but that can easily be remedied with `comptime var` if formatting invalid UTF-8 is a concern.

r00ster91:
@viriuwu unless I'm doing the comparison wrong, codegen seems to be the exact same:
Before: https://zig.godbolt.org/z/4v65c75fY
After: https://zig.godbolt.org/z/1KnoEf1Md
It seems the compiler is smart enough to figure it out without forcing anything with `comptime`.

viriuwu:
> It seems the compiler is smart enough to figure it out

That *does* look the same to me, never mind then! Thanks for the comparison.

PR #11429, opened 2022-04-13
stage2: fix (recursively) marking decls as alive, and improve DWARF for local vars

kubkon:
* x64: recursively mark decls as alive when lowering
* x64: generate debug info for local vars at hardcoded memory address (non-PIE)
* dwarf: generate debug info for arrays
* x64: deref memory address if referenced via GOT when computing the exprloc for a local variable stored in memory
* x64+macho: resolve debug info relocs for RIP-based addressing after symbols were allocated VM addresses - sometimes we will want to generate debug info for a constant that has been lowered to memory and not copied anywhere else. For this we will need to defer resolution on PIE platforms until all locals (including GOT entries) have been allocated.

cc @joachimschmidt557 @Luukdegram @andrewrk this fixes the compiler panic we discussed during the Zig Meetup in Milan.

### DIFF
--- a/src/arch/x86_64/CodeGen.zig
+++ b/src/arch/x86_64/CodeGen.zig
@@ -3950,7 +3950,7 @@ fn genVarDbgInfo(
                     leb128.writeILEB128(dbg_info.writer(), -off) catch unreachable;
                     dbg_info.items[fixup] += @intCast(u8, dbg_info.items.len - fixup - 2);
                 },
-                .memory => |addr| {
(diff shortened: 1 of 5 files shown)


kubkon:
Forgot to add, with this change, we get debug info for comptime values referenced directly via memory (or symbol indexes in case of PIE). Tested in both gdb and lldb :sunglasses:

andrewrk (approved):

joachimschmidt557 (approved):
Amazing work!

PR #11432, opened 2022-04-14
std/math.zig: resolve missed optimization in rotates

### LINKED ISSUE #11431: Missed Optimization with `std.math.rotr` and `std.math.rotl`
### Zig Version

0.10.0-dev.1741+d2681d253

### Steps to Reproduce

You can see what Zig currently produces [here on godbolt](https://godbolt.org/z/MKePeon3e)

### Expected Behavior

I expect trivial `std.math.rotr` and `std.math.rotl` calls to be optimized to their equivalent x86 instructions.

I'd imagine ideally the emitted assembly would look like: 
```asm
mov     ecx, esi
mov     eax, edi
ror     eax, cl
ret
```

### Actual Behavior

Currently, under ReleaseFast, `std.math.rotr` produces:
```asm
shrx    ecx, edi, esi
mov     eax, esi
neg     al
xor     edx, edx
test    sil, 31
shlx    eax, edi, eax
cmove   eax, edx
or      eax, ecx
ret
```

and `std.math.rotl` produces similar output.

paoda:
closes #11431

### DIFF
--- a/lib/std/math.zig
+++ b/lib/std/math.zig
@@ -548,8 +548,8 @@ pub fn rotr(comptime T: type, x: T, r: anytype) T {
     } else if (@typeInfo(T).Int.signedness == .signed) {
         @compileError("cannot rotate signed integer");
     } else {
-        const ar = @mod(r, @typeInfo(T).Int.bits);
-        return shr(T, x, ar) | shl(T, x, @typeInfo(T).Int.bits - ar);
+        const ar = @intCast(Log2Int(T), @mod(r, @typeInfo(T).Int.bits));
+        return x >> ar | x << (1 +% ~ar);
     }
 }
 
@@ -576,8 +576,8 @@ pub fn rotl(comptime T: type, x: T, r: anytype) T {
     } else if (@typeInfo(T).Int.signedness == .signed) {
         @compileError("cannot rotate signed integer");
     } else {
-        const ar = @mod(r, @typeInfo(T).Int.bits);
-        return shl(T, x, ar) | shr(T, x, @typeInfo(T).Int.bits - ar);
+        const ar = @intCast(Log2Int(T), @mod(r, @typeInfo(T).Int.bits));
+        return x << ar | x >> 1 +% ~ar;
     }
 }
 


Vexu (approved):

andrewrk:
Thank you for the patch!

daneelsan:
This snippet of code does not compile, but did on zig 0.9.0:
```zig
const std = @import("std");

pub fn main() anyerror!void {
    const x: u12 = 0o7777;
    std.debug.print("res: 0o{o:0>4}\n", .{std.math.rotl(u12, x, 1)});
}
```
The error:
```
error: RHS of shift is too large for LHS type
        return x << ar | x >> 1 +% ~ar;
                           ^
./test3.zig:5:56: note: called from here
    std.debug.print("res: 0o{o:0>4}\n", .{std.math.rotl(u12, x, 1)});
                                                       ^
./test3.zig:3:29: note: called from here
pub fn main() anyerror!void {
                            ^
/opt/homebrew/Cellar/zig/HEAD-3f11d1d_1/lib/zig/std/math.zig:623:26: note: type u12 has only 12 bits
        return x << ar | x >> 1 +% ~ar;
```

paoda:
Ugh, my bad. 

Fixing this basically means reverting this commit but I'd prefer not to give up the optimizations made for "native" types. 

I can imagine a potential fix would be to check to see if the amount of bits in `T` are a multiple of the native bit width at comptime and then if not, use `shr()` and `shl()` like before. 

I'm not quite sure how I can determine the native bit-width at compile time, however. I know it's most likely improper to assume that it will always be 8 like it is for x86 and ARM. My cursory glance in [lib/std/builtin.zig. ](https://github.com/ziglang/zig/blob/master/lib/std/builtin.zig) resulted in nothing useful but I've most likely just missed something.

daneelsan:
Thanks for the explanation. I'm afraid I also don't know how to determine the native bit-width at compile time...
I opened an issue to track this bug.

Vexu:
> I'm not quite sure how I can determine the native bit-width at compile time

Do you actually need to know the native bit-width or just that it is a multiple of a byte?

topolarity:
You shouldn't need to worry about the native bit-width: Zig's arithmetic does not auto-promote to the machine word size as in C. Instead, it behaves as if operating on the peer-resolved integer type.

So `-%x` (or equivalently `1 +% ~ar`) has the same value on all targets, and it should be enough just to check that you're working with a power-of-two integer size (`std.math.isPowerOfTwo`)

PR #11437, opened 2022-04-15
compiler_rt: atomics: Add TAS lock support for SPARC

koachan:
Some SPARC CPUs (particularly old and/or embedded ones) only has atomic TAS instruction available (`ldstub`). This adds support for emitting that instruction in the spinlock.

### DIFF
--- a/lib/std/special/compiler_rt/atomics.zig
+++ b/lib/std/special/compiler_rt/atomics.zig
@@ -54,14 +54,15 @@ const SpinlockTable = struct {
 
         fn acquire(self: *@This()) void {
             while (true) {
-                const flag = if (comptime arch.isSPARC())
-                    asm volatile ("ldstub [%[addr]], %[flag]"
+                const flag = if (comptime arch.isSPARC()) flag: {
+                    break :flag asm volatile ("ldstub [%[addr]], %[flag]"
                         : [flag] "=r" (-> @TypeOf(self.v)),
                         : [addr] "r" (&self.v),
                         : "memory"
-                    )
-                else
-                    @atomicRmw(@TypeOf(self.v), &self.v, .Xchg, .Locked, .Acquire);
+                    );
+                } else flag: {
+                    break :flag @atomicRmw(@TypeOf(self.v), &self.v, .Xchg, .Locked, .Acquire);
+                };
 
                 switch (flag) {
                     .Unlocked => break,
--- a/lib/std/special/compiler_rt/atomics.zig
+++ b/lib/std/special/compiler_rt/atomics.zig
@@ -1,6 +1,7 @@
 const std = @import("std");
 const builtin = @import("builtin");
-const arch = builtin.cpu.arch;
+const cpu = builtin.cpu;
+const arch = cpu.arch;
 
 const linkage: std.builtin.GlobalLinkage = if (builtin.is_test) .Internal else .Weak;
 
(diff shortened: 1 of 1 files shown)


lib/std/special/compiler_rt/atomics.zig
```diff
@@ -24,6 +24,13 @@ const supports_atomic_ops = switch (arch) {
 // load/store atomically.
 // Objects bigger than this threshold require the use of a lock.
 const largest_atomic_size = switch (arch) {
+    // On SPARC systems that lacks CAS and/or swap instructions, the only
+    // available atomic operation is a test-and-set (`ldstub`), so we force
+    // every atomic memory access to go through the lock.
+    // XXX: Check the presence of CAS/swap instructions and set this parameter
+    // accordingly.
+    .sparc, .sparcel, .sparcv9 => 0,
```
LemonBoy, line 32: V9 is safe, for the other two check for `hasleoncasa`.

lib/std/special/compiler_rt/atomics.zig
```diff
@@ -38,18 +45,35 @@ const SpinlockTable = struct {
     const Spinlock = struct {
         // Prevent false sharing by providing enough padding between two
         // consecutive spinlock elements
-        v: enum(usize) { Unlocked = 0, Locked } align(cache_line_size) = .Unlocked,
+        v: if (arch.isSPARC()) enum(u8) { Unlocked = 0, Locked = 255 } else enum(usize) { Unlocked = 0, Locked } align(cache_line_size) = .Unlocked,
```
LemonBoy, line 48: Break the overlong line by giving the `enum` its own type.
Add a small comment explaining why the second state is `255`.

lib/std/special/compiler_rt/atomics.zig
```diff
@@ -38,18 +45,35 @@ const SpinlockTable = struct {
...
         fn acquire(self: *@This()) void {
             while (true) {
-                switch (@atomicRmw(@TypeOf(self.v), &self.v, .Xchg, .Locked, .Acquire)) {
+                const flag = if (comptime arch.isSPARC())
+                    asm volatile ("ldstub [%[addr]], %[flag]"
+                        : [flag] "=r" (-> @TypeOf(self.v)),
+                        : [addr] "r" (&self.v),
+                        : "memory"
+                    )
+                else
```
LemonBoy, line 58: Nit: add brackets as done for `release` or remove them here.

koachan: I tried to put brackets here but the compiler seems to not like it if there's brackets in if-expressions, that's why I left it out:

```
/home/k/zig/src/zig/lib/std/special/compiler_rt/atomics.zig:53:21: error: expression value is ignored
                    asm volatile ("ldstub [%[addr]], %[flag]"
                    ^
/home/k/zig/src/zig/lib/std/special/compiler_rt/atomics.zig:62:17: error: else prong required when switching on type 'void'
                switch (flag) {
                ^
```

Vexu: You need to use a labeled block and a break:
```zig
const val = val: {
    break :val 1;
};
```

koachan: I see, thanks.

koachan:
Implemented the suggestions from @LemonBoy.

PR #11441, opened 2022-04-15
stage2 ARM: move genArgDbgInfo back to CodeGen

joachimschmidt557:
This removes the questionable Air -> Mir dependency that existed
before. The x86_64 backend also performed this change.

cc @kubkon

### DIFF
--- a/src/arch/arm/CodeGen.zig
+++ b/src/arch/arm/CodeGen.zig
@@ -49,6 +49,7 @@ gpa: Allocator,
 air: Air,
 liveness: Liveness,
 bin_file: *link.File,
+debug_output: DebugInfoOutput,
 target: *const std.Target,
 mod_fn: *const Module.Fn,
 err_msg: ?*ErrorMsg,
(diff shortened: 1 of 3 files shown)


src/arch/arm/CodeGen.zig
```diff
@@ -73,6 +74,12 @@ end_di_column: u32,
 /// which is a relative jump, based on the address following the reloc.
 exitlude_jump_relocs: std.ArrayListUnmanaged(usize) = .{},
 
+/// For every argument, we postpone the creation of debug info for
+/// later after all Mir instructions have been generated. Only then we
+/// will know saved_regs_stack_space which is necessary in order to
+/// address parameters passed on the stack.
+dbg_arg_relocs: std.ArrayListUnmanaged(DbgArgReloc) = .{},
```
kubkon, line 81: This is cool - I love we have two different ways of handling this between x64 and arm backends. If relocs work great, I will probably steal it and backport to x64 as it will clear up stack mgmt in the prologue.

kubkon (approved):

PR #11445, opened 2022-04-15
re-enable f80 tests on freebsd

viriuwu:
- Enables tests skipped due to #11391 being open, which was fixed by 319b5cbce5d4c43de8a6848ac5f5c8879ab9806e.
- Enables f80 testing on `fabs` which I did not add in cb019b80ac8ae8ffd7f7dd619c4da29a83668cde because that same issue was open.


andrewrk:
Thanks!

PR #11453, opened 2022-04-16
std.mem: add concatWithSentinel

fifty-six:
Adds a concatZ to match joinZ et al, useful when dealing with C apis expecting null termination.

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -1941,13 +1941,29 @@ test "joinZ" {
 
 /// Copies each T from slices into a new slice that exactly holds all the elements.
 pub fn concat(allocator: Allocator, comptime T: type, slices: []const []const T) ![]T {
-    if (slices.len == 0) return &[0]T{};
+    return concatMaybeSentinel(allocator, T, slices, null);
+}
+
+/// Copies each T from slices into a new slice that exactly holds all the elements.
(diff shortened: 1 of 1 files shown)


lib/std/mem.zig
```diff
@@ -1941,13 +1941,27 @@ test "joinZ" {
 
 /// Copies each T from slices into a new slice that exactly holds all the elements.
 pub fn concat(allocator: Allocator, comptime T: type, slices: []const []const T) ![]T {
-    if (slices.len == 0) return &[0]T{};
+    return concatMaybeZ(allocator, T, slices, false);
+}
+
+/// Copies each T from slices into a new slice that exactly holds all the elements.
+pub fn concatZ(allocator: Allocator, comptime T: type, slices: []const []const T) ![:0]T {
```
andrewrk, line 1948:
```suggestion
pub fn concatWithSentinel(allocator: Allocator, comptime T: type, slices: []const []const T, comptime s: T) ![:s]T {
```

andrewrk (changes requested):
API counter-proposal.

fifty-six:
Could also have `concatZ`, if you think that'd be a decent addition to the API, as a convenience/to match other methods which just defers to `concatWithSentinel`, although it's unnecessary

PR #11456, opened 2022-04-17
stage2: Do not strip by default for wasm

### LINKED ISSUE #11420: wasm: Missing 'name' section from binary
### Zig Version

0.10.0-dev.1756+17631cb2d

### Steps to Reproduce

Compile the following zig source using the below command
```zig
// test.zig
export fn foo() void{}
```
```
zig build-lib test.zig -target wasm32-freestanding -dynamic -femit-bin=test.wasm
```

Dump the headers of the wasm binary
```
wasm-objdump test.wasm -h
```

### Expected Behavior

```
test.wasm:	file format wasm 0x1

Sections:

     Type start=0x0000000a end=0x0000000e (size=0x00000004) count: 1
 Function start=0x00000010 end=0x00000012 (size=0x00000002) count: 1
    Table start=0x00000014 end=0x00000019 (size=0x00000005) count: 1
   Memory start=0x0000001b end=0x0000001e (size=0x00000003) count: 1
   Global start=0x00000020 end=0x00000028 (size=0x00000008) count: 1
   Export start=0x0000002a end=0x0000003a (size=0x00000010) count: 2
     Code start=0x0000003c end=0x00000041 (size=0x00000005) count: 1
   Custom start=0x00000043 end=0x00000064 (size=0x00000021) "name"
   Custom start=0x00000066 end=0x00000080 (size=0x0000001a) "producers"
```

### Actual Behavior

```
test.wasm:	file format wasm 0x1

Sections:

     Type start=0x0000000a end=0x0000000e (size=0x00000004) count: 1
 Function start=0x00000010 end=0x00000012 (size=0x00000002) count: 1
    Table start=0x00000014 end=0x00000019 (size=0x00000005) count: 1
   Memory start=0x0000001b end=0x0000001e (size=0x00000003) count: 1
   Global start=0x00000020 end=0x00000028 (size=0x00000008) count: 1
   Export start=0x0000002a end=0x0000003a (size=0x00000010) count: 2
     Code start=0x0000003c end=0x00000041 (size=0x00000005) count: 1
```

Luukdegram:
For wasm, we do not wish to strip by default, as it also strips away useful sections such as the 'name' section. This section provides functions with names, which can be used during debugging. Other tools also use this information to generate code (such as WAT).
DWARF info is out-of-scope for this PR, so we do strip debug info by default for wasm still.

Fixes #11420

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1201,7 +1201,6 @@ pub fn create(gpa: Allocator, options: InitOptions) !*Compilation {
             break :blk buf.items[0 .. buf.items.len - 1 :0].ptr;
         } else null;
 
-        const strip = options.strip or !target_util.hasDebugInfo(options.target);
         const red_zone = options.want_red_zone orelse target_util.hasRedZone(options.target);
         const omit_frame_pointer = options.omit_frame_pointer orelse (options.optimize_mode != .Debug);
         const linker_optimization: u8 = options.linker_optimization orelse switch (options.optimize_mode) {
@@ -1244,7 +1243,7 @@ pub fn create(gpa: Allocator, options: InitOptions) !*Compilation {
         cache.hash.add(omit_frame_pointer);
         cache.hash.add(link_mode);
         cache.hash.add(options.function_sections);
-        cache.hash.add(strip);
+        cache.hash.add(options.strip);
         cache.hash.add(link_libc);
         cache.hash.add(link_libcpp);
         cache.hash.add(link_libunwind);
@@ -1433,7 +1432,9 @@ pub fn create(gpa: Allocator, options: InitOptions) !*Compilation {
         };
         errdefer if (module) |zm| zm.deinit();
 
-        const error_return_tracing = !strip and switch (options.optimize_mode) {
+        const error_return_tracing = !options.strip and
+            target_util.hasDebugInfo(options.target) and
+            switch (options.optimize_mode) {
             .Debug, .ReleaseSafe => true,
             .ReleaseFast, .ReleaseSmall => false,
         };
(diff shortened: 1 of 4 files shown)


andrewrk (changes requested):
I'm a bit concerned with the changed definition of the meaning of `strip` with this change - I think `strip` should continue to mean "there is no debug info and no symbol table" and the `hasDebugInfo` function should continue to mean "there is no point to ever turning off strip for this target".

Perhaps this could be solved by adding exceptions for WebAssembly specifically rather than changing the meaning of `strip`? i think `hasDebugInfo` should return `true` for WebAssembly since `strip` will destroy some useful debug information (the name section).

Luukdegram:
> I'm a bit concerned with the changed definition of the meaning of `strip` with this change - I think `strip` should continue to mean "there is no debug info and no symbol table" and the `hasDebugInfo` function should continue to mean "there is no point to ever turning off strip for this target".
> 
> Perhaps this could be solved by adding exceptions for WebAssembly specifically rather than changing the meaning of `strip`? i think `hasDebugInfo` should return `true` for WebAssembly since `strip` will destroy some useful debug information (the name section).

I’m not convinced this changes the definition of `strip`. Passing the strip flag will correctly remove the symbol table (as well as debug info) for Wasm now, where previously it was always being removed as `strip` was always set to true when targeting Wasm. I do, however, agree that `hasDebugInfo` should return true for Wasm, but that pulls in std.debug which hasn’t implemented anything for Wasm just yet. I figured that would be a better fit for a seperate PR. (But perhaps they’re too dependent on eachother). 

Would you mind specifying on what we could apply the exceptions for Wasm?

andrewrk:
> but that pulls in std.debug which hasn’t implemented anything for Wasm just yet

This is where I suggest to make the exception for Wasm - in the std.debug code next to where it checks strip.

Luukdegram:
> > but that pulls in std.debug which hasn’t implemented anything for Wasm just yet
> 
> 
> 
> This is where I suggest to make the exception for Wasm - in the std.debug code next to where it checks strip.

Gotcha, let's go with that! Thanks for the input.

Luukdegram:
Superseded by #11461

PR #11458, opened 2022-04-18
std.build: Include path access specifiers

029onikyu:
This is mainly for ease of use when translating CMakeLists to Zig.
CMake offers three different types of include visibility, Interface, Public, and Private, which have been reproduced here.
These essentially control whether header files added to a target are visible to the source code of that target or to targets that link to it.
With this, linking to a target that has visible include directories will automatically pull in those include directories.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -1627,15 +1627,13 @@ pub const LibExeObjStep = struct {
         other_step: *LibExeObjStep,
     };
 
-    const IncludePathVisibility = packed struct {
-        visible_to_others: bool,
-        visible_to_self: bool,
-    };
+    pub const IncludePathAccess = packed struct {
+        pub const Interface = IncludePathAccess{ .by_others = true, .by_self = false };
+        pub const Public = IncludePathAccess{ .by_others = true, .by_self = true };
+        pub const Private = IncludePathAccess{ .by_others = false, .by_self = true };
 
-    const IncludePathVisibilityPresets = enum(u2) {
-        Interface = @bitCast(u2, IncludePathVisibility{ .visible_to_others = true, .visible_to_self = false }),
-        Public = @bitCast(u2, IncludePathVisibility{ .visible_to_others = true, .visible_to_self = true }),
-        Private = @bitCast(u2, IncludePathVisibility{ .visible_to_others = false, .visible_to_self = true }),
+        by_others: bool,
+        by_self: bool,
     };
 
     pub const Kind = enum {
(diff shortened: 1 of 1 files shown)


lib/std/build.zig
```diff
@@ -1626,6 +1627,17 @@ pub const LibExeObjStep = struct {
         other_step: *LibExeObjStep,
     };
 
+    const IncludePathVisibility = packed struct {
```
matu3ba, line 1630: s/IncludePathVisibility/IncludePathVisible ?

lib/std/build.zig
```diff
@@ -1626,6 +1627,17 @@ pub const LibExeObjStep = struct {
...
+        visible_to_others: bool,
```
matu3ba, line 1631: s/visible_to_others/to_others
s/visible_to_self/to_self

lib/std/build.zig
```diff
@@ -1626,6 +1627,17 @@ pub const LibExeObjStep = struct {
...
+        visible_to_self: bool,
+    };
+
+    const IncludePathVisibilityPresets = enum(u2) {
```
matu3ba, line 1635: something shorter please.

lib/std/build.zig
```diff
@@ -2192,7 +2208,16 @@ pub const LibExeObjStep = struct {
     }
 
     pub fn addIncludePath(self: *LibExeObjStep, path: []const u8) void {
-        self.include_dirs.append(IncludeDir{ .raw_path = self.builder.dupe(path) }) catch unreachable;
+        self.addIncludePathWithVisibility(path, .Private);
+    }
+
+    pub fn addIncludePathWithVisibility(self: *LibExeObjStep, path: []const u8, visibility: IncludePathVisibilityPresets) void {
```
matu3ba, line 2214: Can you think of a shorter synonym?

matu3ba (commented):
The names are very long. Would be nice to make them more ziggish.

lib/std/build.zig
029onikyu: changed to IncludePathAccess

ifreund:
> This is mainly for ease of use when translating CMakeLists to Zig.

Could you show a concrete example comparing the status quo `std.build` API to what you have here? I don't think we should add features just because Cmake has them.

andrewrk:
* missing tests
 * missing a description of the use case
 * no response to @ifreund for 22 days

Feel free to re-open this in the future if you want to pursue this. I suggest to first open an issue to propose an enhancement to the zig build system (not to be confused with a proposal to change the language).

Issue #11455, opened 2022-04-17
JSON stringify string setting affects keys


Sirius902:
### Zig Version

0.10.0-dev.1860+7b090df66

### Steps to Reproduce

```zig
const std = @import("std");

const Foo = struct {
    bar: []const u8,
};

pub fn main() !void {
    const foo = Foo{ .bar = "donuts" };
    const stdOut = std.io.getStdOut();
    try std.json.stringify(foo, .{ .string = .Array }, stdOut.writer());
}
```

### Expected Behavior

I would expect the output of the above program to be the following:
```
{"bar":[100,111,110,117,116,115]}
```

### Actual Behavior

The program outputs the following:
```
{[98,97,114]:[100,111,110,117,116,115]}
```
I would also like to note that  `std.json.parse` cannot parse the above string.

OfekShochat:
I can try doing that as a first contribution! In ~12 hours I can start on this?

OfekShochat:
lol building the compiler gives me a segfault

OfekShochat:
wow Im embarrassed in how much time that took. though it wasnt in actually programming, but trying to first make the compiler compile (which wasnt necessary and doesnt work, some weird link errors) #11459 should close this

andrewrk:
Would appreciate if you mentioned specifically a bug report that documents the compiler not compiling. There is nothing we can do about "it doesn't work" other than say, "it does work, actually". Bug reports need to have steps to reproduce, expected behavior, and actual behavior.

OfekShochat:
Hello, didn't mean it like that, it's just that I have problems like this often on my machine and I should just check it, just explaining why it took so long.

luehmann:
Hey @OfekShochat, I don't know if you are still working on it. I opened a pull-request with the solution I found. I hope that's fine for you. I'd be interested how you solved it

PR #11459, opened 2022-04-18
fixing unintended behaviour for #11455

OfekShochat:


luehmann:
Hi @OfekShochat, I think it would make sense to add tests for this. You can take a look the pull-request I opened before I saw yours.

OfekShochat:
Ok will do! Busy right now though, might take some time

lib/std/json.zig
```diff
@@ -3048,7 +3048,9 @@ pub fn stringify(
                         try out_stream.writeByte('\n');
                         try child_whitespace.outputIndent(out_stream);
                     }
-                    try stringify(Field.name, options, out_stream);
+                    try out_stream.writeByte('\"');
+                    try out_stream.writeAll(Field.name);
```
daurnimator, line 3052: This is not correct; it may need escaping.

daurnimator (changes requested):
Make sure you add a test case of a key that needs escaping in json.

andrewrk:
* CI tests failing
 * Conflicts with master branch
 * No activity for 28 days

Please open a fresh PR if you would like to continue this effort.

PR #11460, opened 2022-04-18
std.json.stringify: always use default options for keys

### LINKED ISSUE #11455: JSON stringify string setting affects keys

luehmann:
closes #11455

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -3048,7 +3048,8 @@ pub fn stringify(
                         try out_stream.writeByte('\n');
                         try child_whitespace.outputIndent(out_stream);
                     }
-                    try stringify(Field.name, options, out_stream);
+                    const key_options = StringifyOptions{};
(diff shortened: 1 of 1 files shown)


luehmann:
Missed there was already #11459

PR #11464, opened 2022-04-18
std.Thread.Futex improvements

kprotty:
- Use the native futex apis available on BSD systems
- Optimizes the userspace pthread implementation of futex
- Introduces `std.atomic.cache_line` for avoiding false sharing
- More comments on stuff

### DIFF
--- a/lib/std/Thread/Futex.zig
+++ b/lib/std/Thread/Futex.zig
@@ -64,7 +64,7 @@ const Impl = if (builtin.single_threaded)
     SerialImpl
 else if (builtin.os.tag == .windows)
     WindowsImpl
-else if (builtin.os.isDarwin())
+else if (builtin.os.tag.isDarwin())
     DarwinImpl
 else if (builtin.os.tag == .linux)
     LinuxImpl
@@ -237,27 +237,28 @@ const DarwinImpl = struct {
     }
 };
 
+// https://man7.org/linux/man-pages/man2/futex.2.html
 const LinuxImpl = struct {
     fn wait(ptr: *const Atomic(u32), expect: u32, timeout: ?u64) error{Timeout}!void {
         var ts: os.timespec = undefined;
-        var ts_ptr: ?*const os.timespec = null;
-
-        if (timeout) |delay| {
-            ts_ptr = &ts;
-            ts.tv_sec = @intCast(@TypeOf(ts.tv_sec), delay / std.time.ns_per_s);
-            ts.tv_nsec = @intCast(@TypeOf(ts.tv_nsec), delay % std.time.ns_per_s);
+        if (timeout) |timeout_ns| {
+            ts.tv_sec = @intCast(@TypeOf(ts.tv_sec), timeout_ns / std.time.ns_per_s);
+            ts.tv_nsec = @intCast(@TypeOf(ts.tv_nsec), timeout_ns % std.time.ns_per_s);
         }
 
         switch (os.linux.getErrno(os.linux.futex_wait(
             @ptrCast(*const i32, &ptr.value),
             os.linux.FUTEX.PRIVATE_FLAG | os.linux.FUTEX.WAIT,
             @bitCast(i32, expect),
-            ts_ptr,
+            if (timeout != null) &ts else null,
         ))) {
             .SUCCESS => {}, // notified by `wake()`
             .INTR => {}, // spurious wakeup
             .AGAIN => {}, // ptr.* != expect
-            .TIMEDOUT => return error.Timeout,
+            .TIMEDOUT => {
+                assert(timeout != null);
+                return error.Timeout;
+            },
             .INVAL => {}, // possibly timeout overflow
             .FAULT => unreachable,
             else => unreachable,
@@ -278,6 +279,7 @@ const LinuxImpl = struct {
     }
 };
 
+// https://www.freebsd.org/cgi/man.cgi?query=_umtx_op&sektion=2&n=1
 const FreebsdImpl = struct {
     fn wait(ptr: *const Atomic(u32), expect: u32, timeout: ?u64) error{Timeout}!void {
         var tm_size: usize = 0;
@@ -302,14 +304,17 @@ const FreebsdImpl = struct {
             @ptrToInt(tm_ptr),
         );
 
-        return switch (rc) {
+        switch (rc) {
             .SUCCESS => {},
             .FAULT => unreachable, // one of the args points to invalid memory
             .INVAL => unreachable, // arguments should be correct
-            .TIMEDOUT => error.Timeout,
+            .TIMEDOUT => {
+                assert(timeout != null);
+                return error.Timeout;
+            },
             .INTR => {}, // spurious wake
             else => unreachable,
-        };
+        }
     }
 
     fn wake(ptr: *const Atomic(u32), max_waiters: u32) void {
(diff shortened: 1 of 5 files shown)


lib/std/atomic.zig
```diff
@@ -80,3 +80,41 @@ test "spinLoopHint" {
...
+
+    // This platform reportedly has 256-byte cache lines
+    // - https://github.com/golang/go/blob/3dd58676054223962cd915bb0934d1f9f489d4d2/src/internal/cpu/cpu_s390x.go#L7
+    .s390x => 256,
+
+    // Other x86 and WASM platforms have 64-byte cache lines.
+    // The rest of the architectures are assumed to be similar.
+    // - https://github.com/golang/go/blob/dda2991c2ea0c5914714469c4defc2562a907230/src/internal/cpu/cpu_x86.go#L9
+    // - https://github.com/golang/go/blob/3dd58676054223962cd915bb0934d1f9f489d4d2/src/internal/cpu/cpu_wasm.go#L7
+    else => 64,
```
matu3ba, line 119: Why not explicitly writing the architectures + conditions here to error out, if user asks for non-provided architecture/condition?

lib/std/Thread/Futex.zig
```diff
@@ -183,58 +170,65 @@ const DarwinFutex = struct {
...
         };
 
         if (status >= 0) return;
         switch (@intToEnum(std.os.E, -status)) {
+            // Wait was interrupted by the OS or other spurious signalling.
             .INTR => {},
             // Address of the futex is paged out. This is unlikely, but possible in theory, and
             // pthread/libdispatch on darwin bother to handle it. In this case we'll return
             // without waiting, but the caller should retry anyway.
             .FAULT => {},
```
matu3ba, line 209: stupid question: how does the caller know if INTR or FAULT happened? Again reading `status` sounds wasteful or is it not?
It should be documented at the function doc that there is stuff to handle. Not sure, if also what cases exact.

lib/std/Thread/Futex.zig
```diff
@@ -243,332 +237,700 @@ const DarwinFutex = struct {
...
-        {
-            assert(std.c.pthread_mutex_lock(&bucket.mutex) == .SUCCESS);
-            defer assert(std.c.pthread_mutex_unlock(&bucket.mutex) == .SUCCESS);
+        switch (os.linux.getErrno(os.linux.futex_wait(
+            @ptrCast(*const i32, &ptr.value),
+            os.linux.FUTEX.PRIVATE_FLAG | os.linux.FUTEX.WAIT,
+            @bitCast(i32, expect),
+            if (timeout != null) &ts else null,
+        ))) {
+            .SUCCESS => {}, // notified by `wake()`
```
matu3ba, line 255: same argument: how does the caller know if INTR or FAULT happened? ...

lib/std/Thread/Futex.zig
```diff
@@ -243,332 +237,700 @@ const DarwinFutex = struct {
...
-                return;
-            }
+    fn wake(ptr: *const Atomic(u32), max_waiters: u32) void {
+        switch (os.linux.getErrno(os.linux.futex_wake(
+            @ptrCast(*const i32, &ptr.value),
+            os.linux.FUTEX.PRIVATE_FLAG | os.linux.FUTEX.WAKE,
+            std.math.cast(i32, max_waiters) catch std.math.maxInt(i32),
+        ))) {
+            .SUCCESS => {}, // successful wake up
+            .INVAL => {}, // invalid futex_wait() on ptr done elsewhere
```
matu3ba, line 275: same argument, omitting the rest.

matu3ba (commented):
`cache_line` to be available looks very useful for performance stuff, though I believe it should not have defaults, if possible (ie to not lead to bugs in future etc).

As I understand it, `status` or `errno` must be checked again to ensure everything worked fine and its not documented at function level as doc comment for almost all cases.
see comment.

lib/std/atomic.zig
kprotty: Because these are estimates. For example: not all x86_64 cpus have 128-byte cache lines, but most do at this point. Dynamically looking it up would make it difficult to use in static allocation instances e.g. on struct declarations.

lib/std/Thread/Futex.zig
kprotty: The caller shouldn't need to be aware of it for correctness when using the Futex API given spurious wakes are a possibility. Tis documented at the function level [here](https://github.com/kprotty/zig/blob/323c07084d6a4a2e2ed3eba690eaeba4e6da1c91/lib/std/Thread/Futex.zig#L20).

lib/std/Thread/Futex.zig
```diff
@@ -21,160 +19,149 @@ const spinLoopHint = std.atomic.spinLoopHint;
...
-        }
+pub fn wait(ptr: *const Atomic(u32), expect: u32) void {
+    @setCold(true);
 
-        // There are no other threads which could notify the caller on single_threaded.
-        // Therefor a wait() without a timeout would block indefinitely.
-        const timeout_ns = timeout orelse {
-            @panic("deadlock");
-        };
+    Impl.wait(ptr, expect, null) catch unreachable;
```
andrewrk, line 27: why catch unreachable?

I suggest to switch on the error even if there is only one error in the set, and especially if we end up making it `unreachable`. This will make it clear what error conditions are unreachable, or perhaps it will highlight the fact that the error condition should be ignored rather than unreachable.

lib/std/Thread/Futex.zig
```diff
@@ -21,160 +19,149 @@ const spinLoopHint = std.atomic.spinLoopHint;
...
 
-    fn wake(ptr: *const Atomic(u32), num_waiters: u32) void {
-        return unsupported(.{ ptr, num_waiters });
+    fn wake(ptr: *const Atomic(u32), max_waiters: u32) void {
+        return unsupported(.{ ptr, max_waiters });
     }
 
     fn unsupported(unused: anytype) noreturn {
-        @compileLog("Unsupported operating system", target.os.tag);
+        @compileLog("Unsupported operating system", builtin.target.os.tag);
```
andrewrk, line 94: Let's use the language features as they were intended to be used:

```suggestion
        @compileError("Unable to provide std.Thread.Futex implementation for target OS " ++ @tagName(builtin.target.os.tag));
```

And then on the side, let's discuss why `@compileLog` is more attractive for this use case than `@compileError` because the very fact that you are tempted to use the unintended builtin here means the language has a design flaw.

lib/std/atomic.zig
andrewrk: Note that we do have extra comptime CPU information for x86_64 which could be used here. You can see an example of this used for arm with `spinLoopHint`. We also have access to the CPU model, if provided. That said, I don't see it as super important and that can always be added later.

andrewrk (changes requested):
Nice work. I have a couple requested changes but nothing structural.

Also, the CI is timing out (6 hours) on macOS and that seems problematic. Could that have anything to do with the changes here?

lib/std/Thread/Futex.zig
kprotty: `@compileLog` is only more attractive due to having built-in varargs and serialization. Not sure if this could (or even should) be added to `@compileError`. Either way, it's a quick fix.

kprotty:
> Also, the CI is timing out (6 hours) on macOS and that seems problematic. Could that have anything to do with the changes here?

@andrewrk Seems like it. I messed up the `timeout_ns == 0` case in `Futex.timedWait()` and darwin's `__ulock_wait` uses a 0 timeout to mean "wait forever".

andrewrk (approved):

PR #11469, opened 2022-04-19
stage2: Move `selfExePathWasi` to `introspect.zig`

topolarity:
Follow-up to #11024. Also updates the guard around the `flock` bypass to be WASI-specific.

Thanks to @andrewrk for the input behind these changes.


andrewrk:
Thanks for the follow-up!

PR #11472, opened 2022-04-19
std.build: user-configurable IO compare

matu3ba:
* expect_custom as function pointer with return bool
* test with capturing and comparing stdout in build.zig
* justfication: enables expecting system-specific paths of running
  binaries or more use-case specific testing with dependencies

### DIFF
--- a/lib/std/build/RunStep.zig
+++ b/lib/std/build/RunStep.zig
@@ -47,6 +47,12 @@ pub const StdIoAction = union(enum) {
     ignore,
     expect_exact: []const u8,
     expect_matches: []const []const u8,
+    /// Custom comparison function for pipe output
+    /// It is recommended to also handle mismatch reporting
+    expect_custom: switch (builtin.zig_backend) {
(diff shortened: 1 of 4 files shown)


matu3ba:
**Do not merge**. This PR exists for now to get feedback on how to proceed, as there is no custom test logic in `./build.zig` to test `lib/std/build.zig`+test_runner.

matu3ba:
Ready for review.

andrewrk (changes requested):
Thank you for the patch, the explanation of the user experience, and for the test case.

Unfortunately, this is a different direction than I want to go with the Zig build system API. My goal is to keep it extremely declarative, keeping in mind the use case of executing an alternative build runner that merely collects the information about building the project, and then **implements its own logic for running (or not running) the steps**. This is crucial for IDEs and tools such as a language server. Adding in a function pointer like this is detrimental to this use case. It will become more clear later on after we start working on IDE tooling why I am making such a fuss about this, but I think it will pay off.

PR #11476, opened 2022-04-19
std.build: Include Path visibility

arBmind:
Extension to #11458


Vexu:
This is now identical to https://github.com/ziglang/zig/pull/11458

PR #11487, opened 2022-04-21
stage2: sparcv9: Implement basic function calls and returns

koachan:
This implements basic function calls/returns on SPARCv9, such that the basic implementation of `_start` from the stdlib works.

### DIFF
--- a/src/arch/sparcv9/CodeGen.zig
+++ b/src/arch/sparcv9/CodeGen.zig
@@ -1,5 +1,7 @@
 //! SPARCv9 codegen.
 //! This lowers AIR into MIR.
+//! For now this only implements medium/low code model with absolute addressing.
+//! TODO add support for other code models.
 const std = @import("std");
 const assert = std.debug.assert;
 const log = std.log.scoped(.codegen);
@@ -884,7 +886,14 @@ fn airCall(self: *Self, inst: Air.Inst.Index, modifier: std.builtin.CallOptions.
 
                 _ = try self.addInst(.{
                     .tag = .jmpl,
-                    .data = .{ .branch_link_indirect = .{ .reg = .o7 } },
+                    .data = .{
+                        .arithmetic_3op = .{
+                            .is_imm = false,
+                            .rd = .o7,
+                            .rs1 = .o7,
+                            .rs2_or_imm = .{ .rs2 = .g0 },
+                        },
+                    },
                 });
             } else if (func_value.castTag(.extern_fn)) |_| {
                 return self.fail("TODO implement calling extern functions", .{});
@@ -899,7 +908,14 @@ fn airCall(self: *Self, inst: Air.Inst.Index, modifier: std.builtin.CallOptions.
 
         _ = try self.addInst(.{
             .tag = .jmpl,
-            .data = .{ .branch_link_indirect = .{ .reg = .o7 } },
+            .data = .{
+                .arithmetic_3op = .{
+                    .is_imm = false,
+                    .rd = .o7,
+                    .rs1 = .o7,
+                    .rs2_or_imm = .{ .rs2 = .g0 },
+                },
+            },
         });
     }
 
--- a/src/arch/sparcv9/CodeGen.zig
+++ b/src/arch/sparcv9/CodeGen.zig
@@ -340,16 +340,15 @@ fn gen(self: *Self) !void {
     if (cc != .Naked) {
         // TODO Finish function prologue and epilogue for sparcv9.
 
-        // TODO Backpatch stack offset
-        // save %sp, -176, %sp
-        _ = try self.addInst(.{
+        // save %sp, stack_save_area, %sp
+        const save_inst = try self.addInst(.{
             .tag = .save,
             .data = .{
                 .arithmetic_3op = .{
                     .is_imm = true,
                     .rd = .sp,
                     .rs1 = .sp,
-                    .rs2_or_imm = .{ .imm = -176 },
+                    .rs2_or_imm = .{ .imm = -abi.stack_save_area },
                 },
             },
         });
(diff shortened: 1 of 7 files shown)


src/arch/sparcv9/CodeGen.zig
```diff
@@ -418,7 +420,7 @@ fn gen(self: *Self) !void {
         .data = .{ .dbg_line_column = .{
             .line = self.end_di_line,
             .column = self.end_di_column,
-        } },
+         } },
```
sin-ack: Missing `zig fmt`?

src/arch/sparcv9/CodeGen.zig
```diff
@@ -1378,12 +1379,79 @@ fn genSetStack(self: *Self, ty: Type, stack_offset: u32, mcv: MCValue) InnerErro
...
+            const biased_offset = stack_offset + abi.stack_bias;
+            const simm13 = math.cast(i13, biased_offset) catch
+                return self.fail("TODO larger stack offsets", .{});
+            return self.genStore(reg, .sp, i13, simm13, abi_size);
+        },
         .memory, .stack_offset => return self.fail("TODO implement memcpy", .{}),
     }
 }
 
-// Common helper functions
```
sin-ack: This line is introduced in the previous commit and removed in this commit.

src/arch/sparcv9/Mir.zig
```diff
@@ -94,6 +94,16 @@ pub const Inst = struct {
         // TODO add other operations.
         sllx,
 
+        /// A.54 Store Integer
+        /// Those uses the arithmetic_3op field.
```
sin-ack, line 96: Typo: Either "Those use" (though the doc comment should probably refer to its subject as this/these), or "This uses".

src/arch/sparcv9/CodeGen.zig
```diff
@@ -892,6 +892,12 @@ fn airCall(self: *Self, inst: Air.Inst.Index, modifier: std.builtin.CallOptions.
                         },
                     },
                 });
+
+                // TODO Find a way to fill this slot
+                _ = try self.addInst(.{
+                    .tag = .nop,
+                    .data = .{ .nop = {} },
+                });
```
sin-ack, line 911: Is this related to https://arcb.csc.ncsu.edu/~mueller/codeopt/codeopt00/notes/delaybra.html? If so it's probably a good idea to have it linked here so that it does not get forgotten.

src/arch/sparcv9/CodeGen.zig
```diff
@@ -771,8 +771,7 @@ fn airArg(self: *Self, inst: Air.Inst.Index) !void {
     // TODO Copy registers to the stack
     const mcv = result;
 
-    // TODO generate debug info
-    // try self.genArgDbgInfo(inst, mcv);
+    try self.genArgDbgInfo(inst, mcv, @intCast(u32, arg_index));
```
sin-ack, line 774: This TODO is introduced in the previous commit and removed right afterwards.

sin-ack (commented):
The final two commits do not preserve the atomicity of previous commits, I suggest squashing the patches to their respective locations.

src/arch/sparcv9/CodeGen.zig
koachan: Hmm, that is what `zig fmt` suggested to me, should I change it back?

src/arch/sparcv9/CodeGen.zig
koachan: Yes, the removal is intended, it was added wrongly, sorry.

src/arch/sparcv9/CodeGen.zig
koachan: Yes, the nops are there to fill the delay slot with something safe.
For the time being it's needed to be done everytime we emit a branch, do I need to add the note in all of them, or?

src/arch/sparcv9/CodeGen.zig
koachan: So, about this one, originally I realized that the dbg_arg instruction is currently useless so I removed it and replaced the emission part with that TODO.
However, a while later I took a read at other backends and found that the RISC-V backend has a working implementation, so I wrote a `genArgDbgInfo` based on that impl and replaced that TODO with a call to the real function.

Is there anything I should do about it?

src/arch/sparcv9/Mir.zig
koachan: Ah, thanks, I'll change the doc comments then.

koachan:
Okay so I'll rework the patches later today. For now, I'm marking this as a draft.

src/arch/sparcv9/CodeGen.zig
sin-ack: What I meant is that commits that come after this should not fix things; instead, all commits should only make the changes that are necessary.

src/arch/sparcv9/CodeGen.zig
sin-ack: I'm not quite sure honestly. Perhaps it's a good idea to add that link to the first occurrence and leave the rest without the link.

src/arch/sparcv9/CodeGen.zig
sin-ack: See my comment on the first review part. Commits should be atomic and shouldn't introduce lines in one commit and remove it in the next within the same PR.

koachan:
Okay, reworked the patch history, now it should be good for review.

src/arch/sparcv9/CodeGen.zig
koachan: Oh I forgot to say, so for this I put a comment in the `return` instruction in the epilogue but for the rest I'm keeping it as-is. Is that okay with you?

koachan:
Rebasing this to trigger the tests again, as the macOS one seems to be failing due to timeout (?)

kubkon:
Just a heads up we are abandoning the current way of testing self-hosted backends. See #11530.

src/arch/sparcv9/Mir.zig
sin-ack: Seems like you missed this one.

src/arch/sparcv9/CodeGen.zig
```diff
@@ -394,7 +394,11 @@ fn gen(self: *Self) !void {
             },
         });
 
-        // TODO Find a way to fill this slot
+        // Branches in SPARC has a delay slot, that is, the instruction
```
sin-ack, line 418: Typo: Branches in SPARC have ...

src/arch/sparcv9/Mir.zig
```diff
@@ -66,34 +66,34 @@ pub const Inst = struct {
...
 
         /// A.49 Shift
-        /// Those uses the shift field.
+        /// This uses the shift field.
         // TODO add other operations.
         sllx,
 
         /// A.54 Store Integer
-        /// Those uses the arithmetic_3op field.
+        /// This uses the arithmetic_3op field.
```
sin-ack, line 96: Please move this change to the first commit where StoreInteger was introduced.

sin-ack (commented):
Looking better :^)

kubkon (approved):
LGTM! While at it, I've rebased, fixed a typo and enabled QEMU support for `sparcv9`. I am wondering though, is `sparcv9` a `sparc64` with a certain CPU feature set? I am asking as I created a mapping `sparvc9 => qemu-sparc64` in `NativeTargetInfo` and seems to run on my Linux with QEMU without any problem. It seems to me like we should really rename `sparcv9` to `sparc64` and put any features specific to `sparcv9` chipset into CPU feature set which we can react to in codegen. Lemme know if this sounds right, or not! In the meantime, this PR looks mergable, but if my comment above makes sense to you, I'd like to work with you on removing `sparcv9` as an architecture, and switching with `sparc64` with CPU feature sets.

kubkon:
Thanks for clarifying! In that case, we should definitely rename the target arch to `sparc64` :-)

PR #11488, opened 2022-04-21
std.os.windows.kernel32: add GetProcessTimes binding

Morritz:
Added one missing Windows API.


kprotty:
Remember to run `zig fmt lib/std` before pushing. The CI is failing without this.

viriuwu:
Unless this is actually going to be used as part of the standard library APIs, I don't think there's much point in adding this (or, well, it'll be removed eventually). See discussion on #4426, specifically Andrew's call on the situation: https://github.com/ziglang/zig/issues/4426#issuecomment-858820245

Morritz:
> Unless this is actually going to be used as part of the standard library APIs, I don't think there's much point in adding this (or, well, it'll be removed eventually). See discussion on #4426, specifically Andrew's call on the situation: [#4426 (comment)](https://github.com/ziglang/zig/issues/4426#issuecomment-858820245)

Yes, It would be required for one of the os apis

PR #11491, opened 2022-04-22
std.math: simplify signbit and copysign

poinwn:
Simplified the implementations of `signbit` and `copysign`, based on `fabs`, also removed the explicit type argument. Works on f80 now!

### DIFF
--- a/lib/std/math/isnormal.zig
+++ b/lib/std/math/isnormal.zig
@@ -5,10 +5,7 @@ const expect = std.testing.expect;
 /// Returns whether x is neither zero, subnormal, infinity, or NaN.
 pub fn isNormal(x: anytype) bool {
     const T = @TypeOf(x);
-    const TBits = std.meta.Int(.unsigned, @bitSizeOf(T));
-    if (@typeInfo(T) != .Float) {
-        @compileError("isNormal not implemented for " ++ @typeName(T));
-    }
+    const TBits = std.meta.Int(.unsigned, @typeInfo(T).Float.bits);
 
     const increment_exp = 1 << math.floatMantissaBits(T);
     const remove_sign = ~@as(TBits, 0) >> 1;
--- a/lib/std/math/signbit.zig
+++ b/lib/std/math/signbit.zig
@@ -5,64 +5,19 @@ const expect = std.testing.expect;
 /// Returns whether x is negative or negative 0.
 pub fn signbit(x: anytype) bool {
     const T = @TypeOf(x);
-    return switch (T) {
-        f16 => signbit16(x),
-        f32 => signbit32(x),
-        f64 => signbit64(x),
-        f80 => signbit80(x),
-        f128 => signbit128(x),
-        else => @compileError("signbit not implemented for " ++ @typeName(T)),
-    };
-}
-
-fn signbit16(x: f16) bool {
-    const bits = @bitCast(u16, x);
-    return bits >> 15 != 0;
-}
-
-fn signbit32(x: f32) bool {
-    const bits = @bitCast(u32, x);
-    return bits >> 31 != 0;
-}
-
-fn signbit64(x: f64) bool {
-    const bits = @bitCast(u64, x);
-    return bits >> 63 != 0;
-}
-
-fn signbit80(x: f80) bool {
-    const bits = @bitCast(u80, x);
-    return bits >> 79 != 0;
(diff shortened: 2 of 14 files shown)


lib/std/math/signbit.zig
```diff
@@ -2,61 +2,25 @@ const std = @import("../std.zig");
...
-    const bits = @bitCast(u64, x);
-    return bits >> 63 != 0;
-}
-
-fn signbit128(x: f128) bool {
-    const bits = @bitCast(u128, x);
-    return bits >> 127 != 0;
+    const TBits = std.meta.Int(.unsigned, @bitSizeOf(T));
+    if (@typeInfo(T) != .Float) {
+        @compileError("copysign not implemented for " ++ @typeName(T));
```
viriuwu, line 10: LGTM, only thing is this error message mismatches the name of the function.

viriuwu (changes requested):

lib/std/math/signbit.zig
viriuwu, line 10: LGTM, only thing is this error message mismatches the name of the function.
```suggestion
        @compileError("signbit not implemented for " ++ @typeName(T));
```

viriuwu (changes requested):

lib/std/math/signbit.zig
poinwn: Thanks!

jibal:
```zig
   const TBits = std.meta.Int(.unsigned, @bitSizeOf(T));
    if (@typeInfo(T) != .Float) {
        @compileError("signbit not implemented for " ++ @typeName(T));
    }
```

```zig
    const TBits = std.meta.Int(.unsigned, @bitSizeOf(T));
    if (@typeInfo(T) != .Float) {
        @compileError("copysign not implemented for " ++ @typeName(T));
    }
```

Surely the assignment to `TBits` should be done after the error message so you get the correct error message for types for which `std.meta.Int(.unsigned, @bitSizeOf(T))` isn't valid (e.g., ` [8192]u8`). As a general strategy code should be localized as much as possible, performing actions (such as defining `TBits`) no sooner than necessary. This is especially important when there are error checks.

jibal:
Thanks for the change and going above and beyond to fix it in the other cases as well!

lib/std/math/signbit.zig
```diff
@@ -2,67 +2,25 @@ const std = @import("../std.zig");
...
-    return bits >> 79 != 0;
-}
-
-fn signbit128(x: f128) bool {
-    const bits = @bitCast(u128, x);
-    return bits >> 127 != 0;
+    if (@typeInfo(T) != .Float) {
+        @compileError("signbit not implemented for " ++ @typeName(T));
+    }
+    const TBits = std.meta.Int(.unsigned, @bitSizeOf(T));
```
andrewrk, line 11:
```suggestion
    const TBits = std.meta.Int(.unsigned, @typeInfo(T).Float.bits);
```

lib/std/math/signbit.zig
```diff
@@ -2,67 +2,25 @@ const std = @import("../std.zig");
 const math = std.math;
 const expect = std.testing.expect;
 
-/// Returns whether x is negative or negative 0.
+/// Returns whether x is a negative number.
```
andrewrk, line 5: This is more ambiguous than the previous wording.

lib/std/math/isnormal.zig
```diff
@@ -5,10 +5,10 @@ const expect = std.testing.expect;
 /// Returns whether x is neither zero, subnormal, infinity, or NaN.
 pub fn isNormal(x: anytype) bool {
     const T = @TypeOf(x);
-    const TBits = std.meta.Int(.unsigned, @bitSizeOf(T));
     if (@typeInfo(T) != .Float) {
         @compileError("isNormal not implemented for " ++ @typeName(T));
     }
+    const TBits = std.meta.Int(.unsigned, @bitSizeOf(T));
```
andrewrk, line 11:
```suggestion
    const TBits = std.meta.Int(.unsigned, @typeInfo(T).Float.bits);
```

andrewrk (changes requested):
Nice work! I'm in favor of these changes. Just a few small requests and then this is ready to be merged.

Do you mind rebasing against master branch for me?

PR #11492, opened 2022-04-22
CI: add non-LLVM backends to the test matrix

andrewrk:
We can't yet run the behavior tests with stage3, but at least we can run
them with stage2, and we can use the proper test matrix.

This commit also adds use_llvm and ofmt to the zig build system.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -445,6 +445,9 @@ pub fn build(b: *Builder) !void {
         false, // skip_single_threaded
         skip_non_native,
         skip_libc,
+        skip_stage1,
+        omit_stage2,
+        is_stage1,
     ));
 
     toolchain_step.dependOn(tests.addPkgTests(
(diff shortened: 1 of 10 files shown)


andrewrk:
Managed to reproduce the CI failure locally:

```
[nix-shell:~/dev/zig/build-release]$ ./stage2/bin/zig build -p stage3 -Dstatic-llvm -Dtarget=native-native-musl --search-prefix ~/dev/zig-bootstrap/out/x86_64-linux-musl-baseline
[nix-shell:~/dev/zig/build-release]$ ./stage3/bin/zig test ../test/behavior.zig -I../test
function definition may only have a distinct !dbg attachment
i16 ()* @behavior.test_0
scope points into the type hierarchy
!11425 = distinct !DILexicalBlock(scope: !11416, file: !11417, line: 3, column: 1)
Expected no forward declarations!
!11420 = <temporary!> !{}
...
```

andrewrk:
I reduced the ambitions of this branch. Once the two bugs noted in this patch are solved, we can try to use stage3 to run the behavior tests rather than stage2.

PR #11497, opened 2022-04-22
std.Thread: Mutex and Condition improvements

kprotty:
#### Main Goals

- Optimize Mutex and Condiiton implementations to use Futex based algorithms.
- More robust tests for Mutex and Condition.
- More documentation in general, but specifically about Condition semantics.

#### Side Goals

- Guarantee at most `u64` large sync primitives (if Futex is removed in the future, they can always be made [`usize` large](https://github.com/kprotty/usync)).
- Introduce `std.Thread.Futex.Deadline` for accurately waiting on a Futex in a loop with a relative timeout.
  - currently used by Condition's futex impl.
  - will be used by Semaphore and ResetEvent's futex impl.

### DIFF
--- a/lib/std/Thread/Condition.zig
+++ b/lib/std/Thread/Condition.zig
@@ -67,7 +67,7 @@ impl: Impl = .{},
 /// - a future call to `signal()` or `broadcast()` which has acquired the Mutex and is sequenced after this `wait()`.
 ///
 /// Given wait() can be interrupted spuriously, the blocking condition should be checked continuously
-/// irrespective of any notifications from `signal()` or `broadcast()`. 
+/// irrespective of any notifications from `signal()` or `broadcast()`.
 pub fn wait(noalias self: *Condition, noalias mutex: *Mutex) void {
     self.impl.wait(mutex, null) catch |err| switch (err) {
         error.Timeout => unreachable, // no timeout provided so we shouldn't have timed-out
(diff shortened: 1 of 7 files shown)


lib/std/Thread/Mutex.zig
```diff
@@ -1,288 +1,282 @@
...
-    /// held by the calling thread.
-    pub fn lock(m: *Dummy) void {
-        if (!m.tryLock()) {
-            @panic("deadlock detected");
+    inline fn lockFast(self: *Impl, comptime casFn: []const u8) bool {
+        // On x86, use `lock bts` instead of `lock cmpxchg` as:
+        // - they both seem to mark the cache-line as modified regardless: https://stackoverflow.com/a/63350048
+        // - `lock bts` is smaller instruction-wise which makes it better for inlining
+        if (comptime builtin.target.cpu.arch.isX86()) {
+            return self.state.bitSet(@ctz(u32, locked), .Acquire) == 0;
```
andrewrk, line 141: stage2 ignores the type parameter of `@clz` in preparation for an upcoming language change. Here's how to solve:

```suggestion
            return self.state.bitSet(@ctz(u32, @as(u32, locked)), .Acquire) == 0;
```

andrewrk:
```
error: <inline asm>:1:12: invalid operand for instruction
        lock btsl (%rax), %eax
                  ^~~~~
```

Is this as problem with stage2 inline assembly? If so, I can work on this to unblock you asap. However if it's possible to work around this with non-assembly code, you can check for `@import("builtin").zig_backend != .stage1` and do the fallback condition. Then I can remove the fallback in favor of the assembly after the stage2 issue is fixed, at my leisure.

kprotty:
@andrewrk Seems like stage2 either doesn't properly lower the `*p`/`*m` constraint or treats the `X` constraint incorrectly. 

Also, the CI also says `noalias` pointers aren't actually pointers. Is this a stage2 false positive?

gracefuu (approved):
LGTM! I can't really comment on the platform specific implementations, but the `FutexImpl`s seem fine.

lib/std/Thread/Condition.zig
```diff
@@ -1,411 +1,538 @@
-//! A condition provides a way for a kernel thread to block until it is signaled
-//! to wake up. Spurious wakeups are possible.
-//! This API supports static initialization and does not require deinitialization.
-
-impl: Impl = .{},
+//! Condition variables are used in tangent with a Mutex to efficiently wait for an arbitrary condition to occur.
```
komuw, line 1: Nitpick: does `in tangent with` mean the same thing as `together with`? if yes, I would use the latter phrase instead of the former.

PR #11499, opened 2022-04-22
`std.math.powi`: use standard definition of underflow/overflow, implement `u0`, `i0`, `i1` edge case

leesongun:
This PR changes `std.math.powi` 
1. use standard definition of underflow/overflow, and document it.
2. implement `u0`, `i0`, `i1` edge case
3. better short-circuit
4. more test cases

details : 
1. I think underflow means 'being to insignificant', rather than 'too close to -Infinity'. (IEEE754 and wikipedia)
2. `1` overflows for `u0`, `i0`, `i1`. 
3. Using `@bitSizeOf` instead of `8 * @sizeOf`. 
4. `pow(i32, -2, 31)` shouldn't overflow, while `pow(i31, -2, 30)` and `pow(i32, -2, 32)` should.


lib/std/math/powi.zig
```diff
@@ -10,108 +10,92 @@ const testing = std.testing;
...
+    if (exp == 1) {
+        if (@mulWithOverflow(T, acc, base, &acc)) {
+            return error.Overflow;
+        }
     }
+
+    return acc;
 }
 
 test "math.powi" {
```
matu3ba, line 92: Unfortunately this is missing a bunch of edge cases.
Using pseudo code related to python
```python
0**0 == 1 // u0,i0,u1,i1
1**1 == 1 // same
1**0 == 1
1**y == 1, y any
```

lib/std/math/powi.zig
```diff
@@ -10,108 +10,92 @@ const testing = std.testing;
...
-                    } else {
-                        return error.Underflow;
-                    }
-                }
-            }
+    if (x == 0) {
+        if (y > 0) {
+            return 0;
+        } else {
+            // not overflow in strict sense
```
matu3ba, line 53: I dont understand this. Please be more specific.

lib/std/math/powi.zig
```diff
@@ -10,108 +10,92 @@ const testing = std.testing;
...
-                if (@mulWithOverflow(T, base, base, &base)) {
-                    if (x > 0) {
-                        return error.Overflow;
-                    } else {
-                        return error.Underflow;
-                    }
-                }
-            }
+    if (x == 0) {
+        if (y > 0) {
```
matu3ba, line 50: It must be `y >= 0`, as 0^0 == 0.

lib/std/math/powi.zig
```diff
@@ -10,108 +10,92 @@ const testing = std.testing;
...
+///  - Underflow: Absolute value of result smaller than 1
 /// Special Cases:
-///  - powi(x, +-0)   = 1 for any x
-///  - powi(0, y)     = 0 for any y
-///  - powi(1, y)     = 1 for any y
-///  - powi(-1, y)    = -1 for y an odd integer
-///  - powi(-1, y)    = 1 for y an even integer
-///  - powi(x, y)     = Overflow for y >= @sizeOf(x) - 1 or y > 0
-///  - powi(x, y)     = Underflow for y > @sizeOf(x) - 1 or y < 0
+///  - powi(x, 0)   = 1 unless T is i1, i0, u0
```
matu3ba, line 17: nit: please use ascending sorting by x, then y.
T is not explained before `T is i1, i0, u0`. I guess its the type for x, but then I still dont understand what the result semantics are from the text.

lib/std/math/powi.zig
```diff
@@ -10,108 +10,92 @@ const testing = std.testing;
...
+///  - powi(0, x)   = 0
+///  - powi(0, x)   = Overflow when x < 0
+///  - powi(1, y)   = 1 unless T is i1, i0, u0
```
matu3ba, line 20: same argument.

lib/std/math/powi.zig
```diff
@@ -10,108 +10,92 @@ const testing = std.testing;
...
+///  - powi(-1, y)  = -1 for y an odd integer
+///  - powi(-1, y)  = 1 unless T is i1, i0, u0
```
matu3ba, line 22: same argument.

lib/std/math/powi.zig
```diff
@@ -10,108 +10,92 @@ const testing = std.testing;
...
 }!T) {
-    const info = @typeInfo(T);
+    const bit_size = @bitSizeOf(T);
 
     comptime assert(@typeInfo(T) == .Int);
 
-    //  powi(x, +-0)   = 1 for any x
-    if (y == 0 or y == -0) {
-        return 1;
+    // integer value 1 cannot be coerced to type 'i1'
```
matu3ba, line 33: I dont see the coercion here. please be more specific. As I understand it, you try to special case any operation on u0,i0,u1,i1.

However, `const t1: u1 = 0; const t2: u1 = 0;` with `0**0` and y being anything is a valid operation. And same for `i1`.

lib/std/math/powi.zig
```diff
@@ -10,108 +10,92 @@ const testing = std.testing;
...
+    if (x == 0) {
+        if (y > 0) {
+            return 0;
+        } else {
+            // not overflow in strict sense
+            return error.Overflow;
+        }
+    }
+    // x >= 2 or x <= -2
+    if (y >= bit_size) {
```
matu3ba, line 58: counterexample: `x**1` with `x=1` for `u1`.

lib/std/math/powi.zig
```diff
@@ -10,108 +10,92 @@ const testing = std.testing;
...
+        return error.Overflow;
+    }
+    if (y < 0) {
```
matu3ba, line 61: should be checked first to remove the `x==0 and `y<0` branch.

lib/std/math/powi.zig
```diff
@@ -10,108 +10,92 @@ const testing = std.testing;
...
-                        return error.Overflow;
-                    } else {
-                        return error.Underflow;
-                    }
-                }
+    var base = x;
+    var exp = y;
+    var acc: T = if (does_one_overflow) unreachable else 1;
+
+    while (exp > 1) {
```
matu3ba, line 69: can you write as comment the used approach? ideally with invariant(s) to show why this works?

lib/std/math/powi.zig
```diff
@@ -140,22 +124,38 @@ test "math.powi" {
...
 }
 
 test "math.powi.special" {
-    try testing.expectError(error.Underflow, powi(i8, -2, 8));
-    try testing.expectError(error.Underflow, powi(i16, -2, 16));
-    try testing.expectError(error.Underflow, powi(i32, -2, 32));
-    try testing.expectError(error.Underflow, powi(i64, -2, 64));
-    try testing.expectError(error.Underflow, powi(i17, -2, 17));
-    try testing.expectError(error.Underflow, powi(i42, -2, 42));
+    try testing.expectError(error.Overflow, powi(i8, -2, 8));
```
matu3ba, line 143: `-2**8` is in python -256, which clearly is < 1. But this gives Overflow.

matu3ba (changes requested):
Nice attempt. Math is tricky.

The semantics for `underflow` are abit brittle, as the return value may be <1 and additionally an overflow exists (which is not intuitive as people also refer to it as "integer underflow").

Aside, a bunch of test cases are missing, 1 optimization could be done and the "reasoning why this math works and correct" would be very nice..

lib/std/math/powi.zig
leesongun: `(-2)**8` is 256

lib/std/math/powi.zig
leesongun: `0^(-1)` should be an overflow, not an underflow?

lib/std/math/powi.zig
leesongun: `powi(u1, 1, 1)` is checked at the first branch, `if (x == 1`

lib/std/math/powi.zig
leesongun: i documented why it won't compile

lib/std/math/powi.zig
leesongun: It is ordered by priority, added comment

lib/std/math/powi.zig
leesongun: `0^0` is `1`

lib/std/math/powi.zig
matu3ba: my bad, mark as resolved then.

lib/std/math/powi.zig
leesongun: it is on the first branch, `if (x == 1 or y == 0`

lib/std/math/powi.zig
matu3ba: ups. my bad. mark as resolved then.

lib/std/math/powi.zig
matu3ba: Are those missing u1?

lib/std/math/powi.zig
matu3ba: Julia thinks `0^(-1)` is Inf.

lib/std/math/powi.zig
matu3ba: I dont understand. The special cases `u0,u1,i0,i1`, which you added in the PR, are not tested.

lib/std/math/powi.zig
leesongun: `u1` can represent `1`, so no need for special case

lib/std/math/powi.zig
leesongun: i misunderstood your comment, will add asap

lib/std/math/powi.zig
leesongun: which is overflow per the code's definition? I don't see how that would remove branch, as `x==0 and y<0` and `x !=0 and y<0` (and implictly `x != -1 and x !=1`, because it is already checked) should give different result (Overflow and Underflow respectively)

lib/std/math/powi.zig
leesongun: added one invariant. a direct invariant would require another tracked variable to be concise, so i hope this would be sufficient.

lib/std/math/powi.zig
leesongun: changed comment to 
`// Infinity/NaN, not overflow in strict sense`

lib/std/math/powi.zig
leesongun: sorry, i misunderstood your comment. 
now i added test cases on `u0, i0, i1`. `u1` is not special cases as it can represent `1`.

lib/std/math/powi.zig
matu3ba: correct, my bad then.

lib/std/math/powi.zig
matu3ba: did you forget to push? The rest is looking decent now.

lib/std/math/powi.zig
```diff
@@ -10,108 +10,96 @@ const testing = std.testing;
...
+///  - powi(T, x, y)   = Overflow when y >= @bitSizeOf(x)
+///  - powi(T, x, y)   = Underflow when y < 0
 pub fn powi(comptime T: type, x: T, y: T) (error{
     Overflow,
     Underflow,
 }!T) {
-    const info = @typeInfo(T);
+    const bit_size = @bitSizeOf(T);
 
     comptime assert(@typeInfo(T) == .Int);
```
andrewrk, line 32:
```suggestion
    const bit_size = @typeInfo(T).Int.bits;
```

lib/std/math/powi.zig
```diff
@@ -10,108 +10,96 @@ const testing = std.testing;
...
 
-    //  powi(x, +-0)   = 1 for any x
-    if (y == 0 or y == -0) {
-        return 1;
+    // `y & 1 == 0` won't compile when `does_one_overflow`.
+    const does_one_overflow = bit_size == 0 or T == i1;
```
andrewrk, line 35:
```suggestion
    const does_one_overflow = math.maxInt(T) < 1;
```

andrewrk (changes requested):
Two small fixups requested, other than that, looks ready to merge.

PR #11501, opened 2022-04-22
stage2: error return traces

### LINKED ISSUE #11084: out of bounds panic does not report the index or bounds
### Zig Version

0.10.0-dev.1120+0ea51f7f4

### Steps to Reproduce

1. Paste this code:
```zig
const std = @import("std");
test "oob!" {
    var foo: [123]u8 = undefined;
    foo[std.rand.DefaultPrng.init(100).random().uintAtMost(usize, 99999) + 123] = 123;
}
```
2. `zig test file.zig`

### Expected Behavior

Something more like this:

```zig
test "oob!" failed
oob.zig:4:5
  2 | test "oob!" {
  3 |    var foo: [123]u8 = undefined; 
  4 |    foo[std.rand.DefaultPrng.init(100).random().uintAtMost(usize, 99999) + 123] = 123;
             ^
panic: index out of bounds
  > index:  1209348
  > bounds: 122

Stack trace:
  test "oob!" | oob.zig:4:5

1 test failure
0 tests passed (0 expectations)
1ms elapsed
```

### Actual Behavior

```bash
❯ zig test /Users/jarred/Desktop/oob.zig
Test [1/1] test "ooob"... thread 5048573 panic: index out of bounds
/Users/jarred/Desktop/zig-cache/o/c5163ac39be41b38c5688f3aeff793b0/test /Users/jarred/Build/zig/zig
```

### LINKED ISSUE #11259: implement error return tracing in the self-hosted compiler
[Error Return Traces](https://ziglang.org/documentation/0.9.1/#Error-Return-Traces)

```zig
fn zirErrorReturnTrace(
    sema: *Sema,
    block: *Block,
    extended: Zir.Inst.Extended.InstData,
) CompileError!Air.Inst.Ref {
    const src: LazySrcLoc = .{ .node_offset = @bitCast(i32, extended.operand) };
    return sema.fail(block, src, "TODO: Sema.zirErrorReturnTrace", .{});
}
```

Vexu:
~~Currently disabled on all targets because LLVM is being a dick and I'm too tired to fight it.~~
```sh-session
$ zig2 test test/behavior.zig -I test -fLLVM    -target aarch64-linux
LLVM Emit Object... 
inlinable function call in a function with debug info must have a !dbg location
  call fastcc void @builtin.returnError(%builtin.StackTrace* %0) #13
inlinable function call in a function with debug info must have a !dbg location
  call fastcc void @builtin.returnError(%builtin.StackTrace* %0) #13
inlinable function call in a function with debug info must have a !dbg location
  call fastcc void @builtin.returnError(%builtin.StackTrace* %0) #13
inlinable function call in a function with debug info must have a !dbg location
  call fastcc void @builtin.returnError(%builtin.StackTrace* %0) #13
```
Not that it was that useful anyways:
```sh-session
$ zig2 run a.zig
error: Foo
???:?:?: 0x20b078 in ??? (???)
???:?:?: 0x20b145 in ??? (???)
```

I tried to do as much as possible in Sema without touching parameters, if that is also desired it should be done alongside the `begin_call/end_call` change.

Closes  #11259
Closes #11084

### DIFF
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -637,7 +637,7 @@ pub const Object = struct {
         const gpa = dg.gpa;
 
         const err_return_tracing = fn_info.return_type.isError() and
-            dg.module.comp.bin_file.options.error_return_tracing;
+            dg.module.comp.bin_file.options.error_return_tracing and false;
 
         const err_ret_trace = if (err_return_tracing)
             llvm_func.getParam(@boolToInt(ret_ptr != null))
@@ -1765,7 +1765,7 @@ pub const Object = struct {
                 }
 
                 if (fn_info.return_type.isError() and
-                    o.module.comp.bin_file.options.error_return_tracing)
+                    o.module.comp.bin_file.options.error_return_tracing and false)
                 {
                     var ptr_ty_payload: Type.Payload.ElemType = .{
                         .base = .{ .tag = .single_mut_pointer },
@@ -2018,7 +2018,7 @@ pub const DeclGen = struct {
         }
 
         const err_return_tracing = fn_info.return_type.isError() and
-            dg.module.comp.bin_file.options.error_return_tracing;
+            dg.module.comp.bin_file.options.error_return_tracing and false;
 
         if (err_return_tracing) {
             dg.addArgAttr(llvm_fn, @boolToInt(sret), "nonnull");
@@ -2484,7 +2484,7 @@ pub const DeclGen = struct {
                 }
 
                 if (fn_info.return_type.isError() and
-                    dg.module.comp.bin_file.options.error_return_tracing)
+                    dg.module.comp.bin_file.options.error_return_tracing and false)
                 {
                     var ptr_ty_payload: Type.Payload.ElemType = .{
                         .base = .{ .tag = .single_mut_pointer },
@@ -3796,7 +3796,7 @@ pub const FuncGen = struct {
         };
 
         if (fn_info.return_type.isError() and
-            self.dg.module.comp.bin_file.options.error_return_tracing)
+            self.dg.module.comp.bin_file.options.error_return_tracing and false)
         {
             try llvm_args.append(self.err_ret_trace.?);
         }
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -637,7 +637,7 @@ pub const Object = struct {
         const gpa = dg.gpa;
 
         const err_return_tracing = fn_info.return_type.isError() and
-            dg.module.comp.bin_file.options.error_return_tracing and false;
+            dg.module.comp.bin_file.options.error_return_tracing;
 
         const err_ret_trace = if (err_return_tracing)
             llvm_func.getParam(@boolToInt(ret_ptr != null))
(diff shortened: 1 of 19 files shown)


schmee:
@Vexu does this PR solve https://github.com/ziglang/zig/issues/1923 as well?

Vexu:
> @Vexu does this PR solve #1923 as well?

No, currently this just matches stage1.

andrewrk (changes requested):
I'll do a second review on the error return traces implementation, but I want to block the merge solely on the basis of 61d25c699ef3a501169399a5073c7462971e1715.

`build.zig` should not assume that there is a config.h file in an arbitrary directory called "build". The intended way that it works currently is if you run `zig build` from inside a build directory that has already been used via the cmake build process, then it will find config.h from there. On my machine, I have many build directories:

```
[nix-shell:~/dev/zig]$ ls build
build/              build-debug/        build-llvm13-debug/ build-release/      build-llvm14/
```

They each have a different config.h, and arbitrarily choosing build/config.h is incorrect.

As for the silent default, what I would expect is for it to search for LLVM via your system default prefix. I see that it is hitting these lines:

```zig
            // Here we are -Denable-llvm but no cmake integration.
            try addStaticLlvmOptionsToExe(exe);
            try addStaticLlvmOptionsToExe(test_stage2);
```

Looking at the implementation of these functions, I don't think there is anything to change. These are perfectly reasonable defaults for the case when one choosing to build from source without the cmake bootstrapping step. Indeed, this is the path we use for the CI as well as zig-bootstrap.

My suggestion for your workflow is to follow the [official building from source instructions](https://github.com/ziglang/zig/wiki/Building-Zig-From-Source) and execute the `zig build` commands from inside the appropriate cmake build directory, or to continue to use the `-Dconfig_h` option.

Vexu:
> Looking at the implementation of these functions, I don't think there is anything to change.

They do need a `ZIG_PREFER_CLANG_CPP_DYLIB` equivalent so I guess I'll implement that instead.

andrewrk:
Will that actually improve anyone's workflow? For your case, you have a build folder from cmake, correct? Then the correct way to build is using that same information, using the cwd method or with `-Dconfig_h`. Package maintainers building a zig package will do the same thing. People using zig-bootstrap will use static libs and never have a use case for dynamic clang.

Vexu:
Well, not currently at least, but if we end up ditching the bootstrap compiler of cleaned output of stage2 with cbe only then there would be no `config.h`?

andrewrk:
Agreed - at that point the logic for integrating with system LLVM, Clang, LLD will become the business of build.zig.

src/codegen/llvm.zig
```diff
@@ -1791,6 +1811,27 @@ pub const Object = struct {
             "", // unique id
         );
     }
+
+    fn getStackTraceType(o: *Object) Type {
```
andrewrk, line 1815: This looks a little fishy. I think it's OK for now, but before this function gets copy+pasted into the other backends, let's look into moving this logic into Sema. Perhaps for a Module in which error return tracing is enabled, it can create a root reference to this type and store it directly in Module.

andrewrk (approved):
Great work! I like how you were able to move some of the functions into userland.

I'm sorry for taking so long to review this. It's ready to be merged as far as I'm concerned.

andrewrk:
I hope you don't mind - I took the liberty to rebase against master branch and force push. I'm running the behavior tests locally and plan to press the merge button if they pass.

PR #11516, opened 2022-04-24
std.SegmentedList: add constIterator

natanalt:
Currently, the SegmentedList only has an `iterator` function, which requires the list to be available through a mutable pointer. Sometimes the user may only have a constant pointer to the list, so I added in a second variant, `constIterator` which creates an equivalent iterator, returning constant pointers to list elements instead of mutable ones.

Usage of `constIterator` is equivalent to `iterator`, and in fact both those iterators now share most of their implementations through an internal generic struct instantiated into `SegmentedList(T).Iterator` and `SegmentedList(T).ConstIterator`.

### DIFF
--- a/lib/std/segmented_list.zig
+++ b/lib/std/segmented_list.zig
@@ -294,71 +294,75 @@ pub fn SegmentedList(comptime T: type, comptime prealloc_item_count: usize) type
             }
         }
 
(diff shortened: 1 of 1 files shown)


andrewrk:
Nice work, thank you!

PR #11528, opened 2022-04-26
std.json: Added Vector type support to std.json.parse()

pncrawley:
Vector type support already existed for std.json.stringify(), but was missing for parse()
This change is basically a copy of the .Array code with minor modifications.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1621,6 +1621,11 @@ fn ParseInternalErrorImpl(comptime T: type, comptime inferred_types: []const typ
                 else => @compileError("Unable to parse into type '" ++ @typeName(T) ++ "'"),
             }
         },
+        .Vector => |vectorInfo| {
+            return error{ UnexpectedEndOfJson, UnexpectedToken } ||
+                ParseInternalErrorImpl(vectorInfo.child, inferred_types ++ [_]type{T}) ||
+                UnescapeValidStringError || TokenStream.Error;
+        },
         else => return error{},
     }
     unreachable;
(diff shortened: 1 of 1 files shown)


lib/std/json.zig
```diff
@@ -1905,6 +1910,28 @@ fn parseInternal(
...
+                    var vec = std.meta.Vector(vectorInfo.len, vectorInfo.child){};
+                    var idx: usize = 0;
+                    while (true) : (idx += 1) {
+                        const tok = (try tokens.next()) orelse return error.UnexpectedEndOfJson;
+                        switch (tok) {
+                            .ArrayEnd => break,
+                            else => {},
+                        }
+
+                        //we ignore extra data, should this be an option to error or ignore?
```
daurnimator, line 1925: I think it should always error.

pncrawley: I added a new error for this in both cases. Smaller and larger than expected vector length.

lib/std/json.zig
```diff
@@ -1959,6 +1994,12 @@ pub fn parseFree(comptime T: type, value: T, options: ParseOptions) void {
                 parseFree(arrayInfo.child, v, options);
             }
         },
+        .Vector => |vectorInfo| {
+            comptime var idx: usize = 0;
+            inline while (idx < vectorInfo.len) : (idx += 1) {
```
daurnimator, line 1999: I don't think this needs to be a comptime loop

lib/std/json.zig
```diff
@@ -1905,6 +1910,36 @@ fn parseInternal(
...
+                    while (true) : (idx += 1) {
+                        const tok = (try tokens.next()) orelse return error.UnexpectedEndOfJson;
+                        switch (tok) {
+                            .ArrayEnd => break,
+                            else => {},
+                        }
+
+                        if (idx < vectorInfo.len) {
+                            vec[idx] = try parseInternal(vectorInfo.child, tok, tokens, options);
+                        }
```
daurnimator, line 1933:
```suggestion
                        if (idx >= vectorInfo.len) return error.VectorLengthMismatch;
                        
                        vec[idx] = try parseInternal(vectorInfo.child, tok, tokens, options);
```
This prevents doing un-necessary work parsing array elements that would never fit in the vector

Issue #11137, opened 2022-03-12
`-lc` doesn't link `c_nonshared`


Vexu:
```sh-session
$ cat a.c
extern char __libc_single_threaded;
int main(void) {
    return __libc_single_threaded;
}
$ clang a.c
$ zig build-exe a.c -lc
ld.lld: error: undefined symbol: __libc_single_threaded
>>> referenced by a.c:3
>>>               /home/vexu/.cache/zig/o/00d791557252b46f5ed6668a930dc12a/a.o:(main)
error: LLDReportedFailure
```

Not sure how this works on other targets but on Arch Linux it means that building stage2 with system LLVM requires this patch:

```diff
diff --git a/build.zig b/build.zig
index 5896ab1a8..fb821ccf4 100644
--- a/build.zig
+++ b/build.zig
@@ -550,6 +550,7 @@ fn addCmakeCfgOptionsToExe(
                 else => |e| return e,
             };
             exe.linkSystemLibrary("unwind");
+            exe.linkSystemLibrary("c_nonshared");
         } else if (exe.target.isFreeBSD()) {
             try addCxxKnownPath(b, cfg, exe, "libc++.a", null, need_cpp_includes);
             exe.linkSystemLibrary("pthread");
```

daurnimator:
Arch uses glibc 2.35.

Looking through recent glibc changes related to `c_nonshared` I found https://github.com/bminor/glibc/commit/1a2f44a848663036c8a14671fe0faa3fed0b2a25. I wonder if this means that linking `pthread` previously implied `c_nonshared`, but it doesn't any more.

ifreund:
The proper fix is *probably* adding `-lc_nonshared` to `libcFullLinkFlags()` in `src/target.zig` if targeting glibc. I would like to understand exactly what's happening here though before doing that.

matu3ba:
At least for me a default build of glibc contains `libc_nonshared.a`, but `src`glibc.zig` does not contain it in the library list:
```zig
// The order of the elements in this array defines the linking order.
pub const libs = [_]Lib{
    .{ .name = "m", .sover = 6 },
    .{ .name = "pthread", .sover = 0 },
    .{ .name = "c", .sover = 6 },
    .{ .name = "dl", .sover = 2 },
    .{ .name = "rt", .sover = 1 },
    .{ .name = "ld", .sover = 2 },
    .{ .name = "util", .sover = 1 },
};
```
[...]

PR #11529, opened 2022-04-26
stage2: fix build on Arch Linux

mewmew:
Prior to this commit, building on Arch Linux would result in
the following linker error:

	undefined symbol: __libc_single_threaded

Updates ziglang/zig#11137.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -154,6 +154,7 @@ pub fn build(b: *Builder) !void {
 
     b.default_step.dependOn(&exe.step);
     exe.single_threaded = single_threaded;
+    exe.linkSystemLibrary("c_nonshared");
 
     if (target.isWindows() and target.getAbi() == .gnu) {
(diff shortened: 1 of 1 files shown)


mewmew:
This PR should most likely NOT BE MERGED.

ref: @ifreund
> the patch there is a workaround for some deeper issue, it shouldn’t be sent as a PR without further investigation and understanding of why it is needed on arch but not elsewhere

with regards to issue https://github.com/ziglang/zig/issues/11137

This PR can be used as a workaround to get stage2 to build on Arch Linux, until #11137 is resolved.

mewmew:
**Note**, this seems to give the build error `ld.lld: error: unable to find library -lc_nonshared` on the build bot of `
ci.ziglang.org/drone/pr`.

Vexu:
Closing since this is not the proper fix and the same patch can be found in the issue.

Issue #385, opened 2017-06-04
extern functions should know which library they are expected to be found in


andrewrk:
And the compiler should look for symbols in libraries and verify that the function exists or not.

And then instead of undefined symbol linker errors, we'll get nice compiler errors.

andrewrk:
I believe this will also enable us to generate .lib files, such as kernel32.lib, removing this potential barrier to cross compiling.

andrewrk:
Proposal:

```zig
extern "kernel32" {
    pub stdcallcc fn CryptAcquireContext(phProv: &HCRYPTPROV, pszContainer: LPCTSTR,
        pszProvider: LPCTSTR, dwProvType: DWORD, dwFlags: DWORD) -> bool;

    pub stdcallcc fn CryptReleaseContext(hProv: HCRYPTPROV, dwFlags: DWORD) -> bool;

    pub stdcallcc fn CryptGenRandom(hProv: HCRYPTPROV, dwLen: DWORD, pbBuffer: &BYTE) -> bool;
}
```

Change 

```zig
extern fn foo() {

}
```
to
```zig
ccc fn foo() { /// c calling convention

}
```

Couple monkey wrenches:

 * When linking against dynamic libraries we don't actually have the library available until runtime to check if all the functions resolve. If you have the libraries available and know them to be the correct version, then you can have the check. But you could be cross compiling or for other reasons not have them.
 * `@cImport(...)` there's not really a way to know, *per-function* which library things come from, if you have multiple includes in the expression, which is the intended way to use the feature. And we really need to know which library each function comes from so we can generate the .lib file on windows.

andrewrk:
This is already done. The library name is optional, however, since it's a feature of the C ABI that you don't have to know what library a function comes from.

PR #11530, opened 2022-04-26
test: move compare and run tests to new test harness

kubkon:
Builds on top of #11417 and #11421.

~~Submitting as a draft PR so that all interested parties can comment ahead of time.~~

Implements parsing of key-value config values in the test manifest as prescribed in https://github.com/ziglang/zig/issues/11288. With this PR landing, we currently support specifying a set of backends, targets, whether it is a test, output mode, as well as the type. For example, the following combination is now fully supported

```
// error
// output_mode=Exe
// backend=stage1,stage2,llvm
// target=aarch64-macos,x86_64-macos
//
// :1:2: error: WATTTT
```

We unpack this manifest into a Cartesian product of all possible combinations as you'd expect.

A couple of limitations of the current state-of-affairs:
* note that the `output_mode` has to be specified in Camelcase - this is due to the fact that the default parser for enums is using `std.meta.stringToEnum` and `std.builtin.OutputMode` defines variants using Camelcase. We can either fix the parser or, my preference, rename the variants to lowercase as it now idiomatic in Zig (AFAIK).
* when we detect a sequence of incremental updates (`name.X.zig`), in any update following the first one (`name.0.zig`), we only parse the type of the update (`run` or `error`), and then the trailing data. Any additional key-value config values will not be actionable. We can tweak this should there be a need to include other config values for subsequent updates.
* finally, currently, the default for `backend` option if unspecified is only `stage2` but I'd like it to change to `stage2,llvm` when both become more feature complete
* the default for `target` is the set of the mainstream targets: `arm-linux`,`x86_64-linux`,`aarch64-linux`,`x86_64-macos`,`aarch64-macos`,`wasm32-wasi`
* handling of `cli` type of tests is currently unimplemented, as well as `translate_c`

The next missing bit I want to implement is multi-source/object tests.

That's about it! Looking forward to your comments folks!

cc @topolarity @mitchellh


kubkon:
> Superb! Looks like this opens the door to move over a ton of the existing tests 👌
> 
> Is the plan is to temporarily support the "legacy" manifest format for compile error tests until those are updated in a follow-up? That works for me - just looking to confirm

I'll strive to update compile error tests to the updated manifest format too. There's a quite a bit more I'd like to explore before settling on the `TestManifest` parser before I'll mark this PR mergeable, but hopefully in a day or two, I should have something working.

src/test.zig
```diff
@@ -34,42 +34,28 @@ test {
...
-        std.fs.path.dirname(@src().file).?, "..", "test", "compile_errors",
-    });
-
-    var compile_errors_dir = try std.fs.cwd().openDir(compile_errors_dir_path, .{});
-    defer compile_errors_dir.close();
-
     {
-        var stage2_dir = try compile_errors_dir.openDir("stage2", .{ .iterate = true });
-        defer stage2_dir.close();
+        const dir_path = try std.fs.path.join(arena, &.{
```
matu3ba, line 38: missing defer arena.free(dir_path)

src/test.zig
```diff
@@ -34,42 +34,28 @@ test {
...
-        var stage1_dir = try compile_errors_dir.openDir("stage1", .{});
-        defer stage1_dir.close();
-
-        const Config = struct {
-            name: []const u8,
-            is_test: bool,
-            output_mode: std.builtin.OutputMode,
-        };
+    {
+        const dir_path = try std.fs.path.join(arena, &.{
```
matu3ba, line 51: missing defer arena.free(dir_path)

src/test.zig
```diff
@@ -34,42 +34,28 @@ test {
...
-    defer compile_errors_dir.close();
-
     {
-        var stage2_dir = try compile_errors_dir.openDir("stage2", .{ .iterate = true });
-        defer stage2_dir.close();
+        const dir_path = try std.fs.path.join(arena, &.{
+            std.fs.path.dirname(@src().file).?, "..", "test", "compile_errors",
+        });
+
+        var dir = try std.fs.cwd().openDir(dir_path, .{ .iterate = true });
```
matu3ba, line 42: suggestions: add .no_follow = true and to the other instances of openDir

src/test.zig
```diff
@@ -154,6 +140,261 @@ const ErrorMsg = union(enum) {
     }
 };
 
+/// Default config values for known test manifest key-value pairings.
```
matu3ba, line 143: "known test manifest" is unclear.
What does this mean? Why can the target triplet not being reused etc?

src/test.zig
```diff
@@ -154,6 +140,261 @@ const ErrorMsg = union(enum) {
...
+                .run => "Exe",
+                .cli => @panic("TODO test harness for CLI tests"),
+            };
+        } else if (std.mem.eql(u8, key, "is_test")) {
+            return "0";
+        } else unreachable;
+    }
+};
+
+/// Manifest syntax example:
```
matu3ba, line 184: spelling: examples
Its simpler to read with
````txt
type
config_map

trailing_bytes (the output)
-----
1-2 examples
```

src/test.zig
```diff
@@ -785,108 +1010,99 @@ pub const TestContext = struct {
...
+            if (std.mem.eql(u8, prev_parts.base_name, new_parts.base_name) and
+                std.mem.eql(u8, prev_parts.file_ext, new_parts.file_ext))
+            {
+                // This is "foo.X.zig" followed by "foo.Y.zig". Make sure that X = Y + 1
+                if (prev_parts.test_index == null) return error.InvalidIncrementalTestIndex;
+                if (new_parts.test_index == null) return error.InvalidIncrementalTestIndex;
+                if (new_parts.test_index.? != prev_parts.test_index.? + 1) return error.InvalidIncrementalTestIndex;
+            } else {
+                // This is not the same test sequence, so the new file must be the first file
+                // in a new sequence ("*.0.zig") or an independent test file ("*.zig")
```
matu3ba, line 1029: how is the errorneaous existence of both (.0.zig and .zig) ruled out?

src/test.zig
```diff
@@ -785,108 +1010,99 @@ pub const TestContext = struct {
...
+                            },
+                            .incremental => {
+                                case.addErrorNamed("update", src, errors);
+                            },
+                        }
                     },
-                    .incremental => {
-                        case.addErrorNamed(case_name, src, errors.items);
+                    .run => {
+                        var output = std.ArrayList(u8).init(ctx.arena);
```
matu3ba, line 1093: Does this create a new arena instance? Otherwise this looks like it would leak memory.
But maybe, I am just blind.

src/test.zig
```diff
@@ -785,108 +1010,99 @@ pub const TestContext = struct {
...
+                    const ext_index = std.mem.lastIndexOfScalar(u8, current_file.*, '.') orelse
+                        return error.InvalidFilename;
+                    const index = std.mem.lastIndexOfScalar(u8, current_file.*[0..ext_index], '.') orelse ext_index;
+                    break :blk current_file.*[0..index];
+                };
 
-                // If the second line is present, it should be blank
-                if (manifest_it.next()) |second_line| {
-                    if (std.mem.trim(u8, second_line[2..], " \t").len != 0) return error.SecondLineNotBlank;
+                // Cross-product to get all possible test combinations
```
matu3ba, line 1054: could allocation size be a future problem for development on potato systems?
It might be good to set an upper limit of combinations to have some control.

matu3ba (changes requested):
some smaller changes sound feasible and I added a few comments/questions on memory consumption etc.

How did you test against invalid inputs? Do you think there should be a path that does not execute anything for testing/fuzzing?

I dont understand the idea behind getConfigForKeyCustomParser. Can you add a comment for this?

src/test.zig
Luukdegram: `free` is a no-op for arena's. All memory is freed at once by calling `deinit`.

kubkon: As @Luukdegram already pointed out, we deinit the entire arena all at once when freeing `TestContext`. Normally, if this was public API, I'd have the function accept an allocator and do the bookkeeping manually, but since it's meant to always be used with `TestContext`, we implicitly assume the above.

src/test.zig
kubkon: For safety we could, or to avoid unintended bugs; however I cannot see us adding symlinks inside the test directories for compiler tests ever. Note that this test harness should never be used to test any standard library calls.

src/test.zig
kubkon: What do you mean by a target triplet?

src/test.zig
kubkon: This is a good question, and I can direct to the author of this bit of code: @topolarity could you shed some light on this please? Having said that, this is out of scope of this PR as this PR only abstracted the existing logic into a helper struct.

src/test.zig
kubkon: Let's revisit this once we get everything working on mainstream systems first.

kubkon:
> some smaller changes sound feasible and I added a few comments/questions on memory consumption etc.
> 
> How did you test against invalid inputs? Do you think there should be a path that does not execute anything for testing/fuzzing?

We throw an error cancelling the test run altogether on an invalid manifest input. Having said that though, this is an area I will definitely work a lot more on. This PR however is just a first step into porting existing tests to the new test harness.

> 
> I dont understand the idea behind getConfigForKeyCustomParser. Can you add a comment for this?

This is just a helper for returning a strongly typed config value for some key with a parsing function explicitly specified. I might remove it in the future we will only ever use default parsers such as `std.meta.stringToEnum`.

src/test.zig
matu3ba: I thought there might be a target triplet parser, which could be reused. However, currently I dont find one.

kubkon: Oh, but we have one and it's used here by default. Have a look at line #385 of the file:

```zig
            .Struct => if (comptime std.mem.eql(u8, @typeName(T), "CrossTarget")) return struct {
                fn parse(str: []const u8) ?T {
                    var opts = CrossTarget.ParseOptions{
                        .arch_os_abi = str,
                    };
                    return CrossTarget.parse(opts) catch null;
                }
            }.parse else @compileError("no default parser for " ++ @typeName(T)),
```

The reason `TestManifestConfigDefaults` returns triplets as a string is to feed uniformly into the config iterator and parsing function. This seemed like the cleanest and simplest way of the iterator *always* returning a stream for each required config value such `backend` and `target`.

src/test.zig
```diff
@@ -785,108 +1010,99 @@ pub const TestContext = struct {
...
+            // Parse the manifest
+            var manifest = try TestManifest.parse(ctx.arena, src);
 
-                        // We ignore all-whitespace lines following the comment block, but anything else
-                        // means that there is no manifest present.
-                        if (std.mem.trim(u8, src[cursor..manifest_end], " \r\n\t").len == 0) {
-                            manifest_end = cursor;
-                        } else break; // If it's not whitespace, there is no manifest
-                    }
+            if (cases.items.len == 0) {
```
topolarity, line 1041: IIUC, this skips over any backend/target/etc. entries in the manifest if the test case is part of an incremental update

Would it be better to check that these have been left unspecified, or else match the effective configuration?

kubkon: Yeah, this is on purpose. I don't want to created cases for incremental updates, rather I just want to update expected error/output conditions. Having said that, I think it'd be nice if we checked if the manifest changed significantly and flag an error if it did (I don't see any value in suddenly requesting a different target for an update - doesn't make any sense, right?). I didn't want to introduce too much new functionality from get-go into this PR hence why I didn't add the full manifest checking yet.

Also, note we only ignore the config values; the type of the update (`run` or `error`) as well as parsing of the trailing data is still performed as expected.

kubkon: One thing we could consider at a spec level though is: do we want to spell out full manifests for each incremental update test file so that we can run them as independent test cases, or not? For now, I assumed that latter, but we can easily change that in the future should we want more flexibility.

kubkon: Finally, this check `cases.items.len == 0` is equivalent to the previous `opt_case == null` check. The reason we need an array here is that we need to be able to unpack a manifest into potentially several cases that are a Cartesian product of the `backend`s and `target`s.

topolarity: For explicitly incremental updates (".0.zig", ".1.zig", etc.), I think that makes sense

It'd be nice to check that these config values are not provided in anything except the first config, so that folks are not confused when these are ignored. But I'm definitely OK with that being left as a TODO for now

The more important milestone will be when we try to run these tests in incremental mode, so that tests become "implicitly" incremental, too. I'd like to make sure we don't forget that we haven't resolved the configuration differences between test cases when we eventually try to turn that on for stage2. It's not totally clear to me what the current proposal is for handling backends, etc. when running in incremental mode

cc @mitchellh , since he may have some valuable input here

kubkon:
One thing I forgot to mention (and this will be of particular interest for you @topolarity), test names are taken to be the

`subpath/subpath/.../filename (backend, target)`

It's pretty verbose, but saves us explicitly specifying the test name in the manifest when we already have a unique sub path + filename + backend + target combo. One thing I didn't test yet though is test filtering which I'll start testing after we land this PR too for better testing experience.

src/test.zig
kubkon: Yep agreed. I actually left it unfinished for that very reason - so that we can discuss the implications and follow-up with a fix(es) if desired.

For incremental mode, we can run both `stage2` and `llvm` backends which we already do. `stage1` will assert there are updates upon execution.

Issue #11161, opened 2022-03-14
lower float functions to the `l`, `q`, or `x` suffixes as appropriate


andrewrk:
Many behavior tests are disabled for 128-bit floating point functions. Have a look for this pattern inside test/behavior/math.zig:

https://github.com/ziglang/zig/blob/5ea94e7715607e986298908536cdd3d9dfdd0ce9/test/behavior/math.zig#L1110-L1114

We cannot rely on LLVM to lower these intrinsics correctly for 128 bit floats on platforms where `long double` is not the same as `f128`.

The solution is done for example with `fma`:

https://github.com/ziglang/zig/blob/5ea94e7715607e986298908536cdd3d9dfdd0ce9/src/codegen/llvm.zig#L4943-L4959

In summary; lower to the `l` suffix when `long double` happens to be correct on the target platform. Otherwise, lower to `q` suffix for `f128` or to a `__` prefix and `x` suffix for `f80`. This requires having such functions available in freestanding libc. The implementations can be ported from musl libc. Just pick a set of `#ifdef`s for `long double` that match a target where `long double` happens to be `f80` or `f128` and port the corresponding code.

Finally we need to solve vectors. I think the best solution is to scalarize the vector float function calls when we know that LLVM will generate calls to the wrong function.

Of course it would be nice if LLVM would solve this problem but we may have a hard time convincing them to make the codegen rely on non-standard compiler-rt features. This is why we should also simultaneously work to enhance the compiler-rt standard to include our extensions.

andrewrk:
Solved by @topolarity and me, landed in 360ecc1a2f72967f3a3882b3327e130bdc4e18c0.

PR #11532, opened 2022-04-26
compiler-rt math functions reorg

andrewrk:
* unify the logic for exporting math functions from compiler-rt,
   with the appropriate suffixes and prefixes.
   - add all missing f128 and f80 exports. Functions with missing
     implementations call other functions and have TODO comments.
   - also add f16 implementations
 * move math functions from freestanding libc to compiler-rt (#7265)
 * enable all the f128 and f80 code in the stage2 compiler and behavior
   tests
 * update std lib to use builtins rather than `std.math`.
 * `tan` is moved to compiler-rt to match `@sin` and `@cos`, because all three implementations share a common dependency.

Merge checklist:
 * [x] #11161
 * [x] add `@tan` builtin.


lib/std/math/acosh.zig
```diff
@@ -29,15 +29,15 @@ fn acosh32(x: f32) f32 {
 
     // |x| < 2, invalid if x < 1 or nan
     if (i < 0x3F800000 + (1 << 23)) {
-        return math.log1p(x - 1 + math.sqrt((x - 1) * (x - 1) + 2 * (x - 1)));
+        return math.log1p(x - 1 + @sqrt((x - 1) * (x - 1) + 2 * (x - 1)));
     }
     // |x| < 0x1p12
     else if (i < 0x3F800000 + (12 << 23)) {
-        return math.ln(2 * x - 1 / (x + math.sqrt(x * x - 1)));
+        return @log(2 * x - 1 / (x + @sqrt(x * x - 1)));
```
fabioarnold, line 36: `math.ln` and `@log` are not equivalent. `ln` is base `e` and `log` is base `10`. You would need to divide by `@log(e)`.

tiehuis: `@log` does refer to the natural logarithm. See https://ziglang.org/documentation/0.9.1/#log.

fabioarnold: Thanks, I should have checked before commenting. Sorry for that!

andrewrk: It's still a fair point; we may want to play with the naming to make it less ambiguous. Although we are a bit limited by libc here since we do need to match the symbol names, and libc has `log` as natural log.

andrewrk:
Labeling as "breaking" because this removes `std.math` functions which have corresponding builtins, such as `@sqrt`.

src/codegen/llvm.zig
```diff
@@ -8055,6 +8188,26 @@ fn backendSupportsF80(target: std.Target) bool {
...
+/// and false if we expect LLVM to crash if it counters an f16 type or
+/// if it produces miscompilations.
+fn backendSupportsF16(target: std.Target) bool {
+    return switch (target.cpu.arch) {
+        else => true,
+    };
+}
+
+/// LLVM does not support all relevant intrinsics for all targets, so we
+/// may need to manually generate a libc call
```
matu3ba, line 8201: same idea: offering the opportunity to query from build.zig if intrinsics are used sounds useful.

matu3ba (commented):
idea (could also be part of later refactoring): expose those things as part of the target info to be usable in build.zig

src/codegen/llvm.zig
```diff
@@ -8055,6 +8188,26 @@ fn backendSupportsF80(target: std.Target) bool {
     };
 }
 
+/// This function returns true if we expect LLVM to lower f16 correctly
+/// and false if we expect LLVM to crash if it counters an f16 type or
+/// if it produces miscompilations.
```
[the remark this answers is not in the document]
andrewrk: No, it is best if the user does not check these conditions. Technically, they already can, because they can put this logic into their application code and check for `@import("builtin").zig_backend` being one of the LLVM backends. However, see the doc comment on the `CompilerBackend` enum for best practices related to this.

frmdstryr:
Would you be opposed to having backward compatible inline fn's included in math?

Code doing complex math is already difficult enough to follow as is... remembering which is builtin vs in math just adds unnecessary friction (and makes the code ugly imo).

andrewrk:
Not opposed

PR #11533, opened 2022-04-26
stage2: add referenced by trace to compile errors

Vexu:
This is useful information to have but the way stage1 handled it was quite noisy so I tried to make it as condensed as possible. If it is still too much a flag to disable them could be added for those who want it.
Screenshot from 2022-04-26 22-51-30

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -342,10 +342,17 @@ pub const AllErrors = struct {
             /// Does not include the trailing newline.
             source_line: ?[]const u8,
             notes: []Message = &.{},
+            reference_trace: []Message = &.{},
         },
         plain: struct {
             msg: []const u8,
             notes: []Message = &.{},
+        },
+        ref: struct {
+            decl_name: []const u8,
+            src_path: []const u8,
+            line: u32,
+            column: u32,
         },
 
         pub fn renderToStdErr(msg: Message, ttyconf: std.debug.TTY.Config) void {
@@ -395,6 +402,21 @@ pub const AllErrors = struct {
                     for (src.notes) |note| {
                         try note.renderToStdErrInner(ttyconf, stderr_file, "note:", .Cyan, indent);
                     }
+                    if (src.reference_trace.len != 0) {
+                        ttyconf.setColor(stderr, .Reset);
+                        ttyconf.setColor(stderr, .Dim);
+                        try stderr.print("referenced by:\n", .{});
+                        for (src.reference_trace) |reference| {
+                            try stderr.print("    {s}: {s}:{d}:{d}\n", .{
+                                reference.ref.decl_name,
+                                reference.ref.src_path,
+                                reference.ref.line + 1,
+                                reference.ref.column + 1,
+                            });
+                        }
+                        try stderr.writeByte('\n');
+                        ttyconf.setColor(stderr, .Reset);
+                    }
                 },
                 .plain => |plain| {
                     ttyconf.setColor(stderr, color);
@@ -407,6 +429,7 @@ pub const AllErrors = struct {
                         try note.renderToStdErrInner(ttyconf, stderr_file, "error:", .Red, indent + 4);
                     }
                 },
+                .ref => unreachable,
             }
         }
     };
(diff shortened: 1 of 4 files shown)


src/Compilation.zig
```diff
@@ -342,11 +342,18 @@ pub const AllErrors = struct {
             /// Does not include the trailing newline.
             source_line: ?[]const u8,
             notes: []Message = &.{},
+            reference_trace: []Message = &.{},
         },
         plain: struct {
             msg: []const u8,
             notes: []Message = &.{},
         },
+        ref: struct {
```
andrewrk, line 351: Any particular reason this is another tagged union field rather than a separate, named struct, and then having `reference_trace: []Ref` rather than `reference_trace: []Message` ?

src/Compilation.zig
```diff
@@ -395,6 +402,21 @@ pub const AllErrors = struct {
                     for (src.notes) |note| {
                         try note.renderToStdErrInner(ttyconf, stderr_file, "note:", .Cyan, indent);
                     }
+                    if (src.reference_trace.len != 0) {
```
andrewrk, line 405: We need to do one of the following things before merging:

1. Make sure that existing tools properly handle the "referenced by" format as you've created it. There are many tools that support the standard "error: file:line:column: message" format of stderr, such as IDE integration, vim plugins, terminal extensions, etc. We need to survey the landscape and make sure this format is machine-friendly.
2. Or, we can render to stderr with the "referenced by" trace using the standard "note: file:line:column: referenced here" format, when `ttyconf` has determined that the output is not a terminal.

src/Module.zig
```diff
@@ -386,6 +386,7 @@ pub const Decl = struct {
     /// the contents hash, followed by line, name, etc.
     /// For anonymous decls and also the root Decl for a File, this is 0.
     zir_decl_index: Zir.Inst.Index,
+    first_referenced_by: ReferencedBy = .{},
```
andrewrk, line 389: `Decl` memory is something we are concerned with keeping minimal. Given that both of these things are true...

 * referenced-by traces are noise in some cases and it may make sense to opt-in to this extra diagnostic info
 * keeping track of referenced-by traces may come with a performance cost in the compiler (even when no compile errors occur)

...I suggest that we keep all referenced-by trace information in a separate table that remains empty unless the compilation opts into additional compile-error diagnostics.

Also, I would expect to see some kind of modification to `Module.destroyDecl`, removing `ReferencedBy` objects. Have you done any testing with incremental compilation on this? We may want to get some improved debug/testing infrastructure going for this before attempting to implement referenced by diagnostics.

src/Module.zig
```diff
@@ -1953,7 +1959,15 @@ pub const EmbedFile = struct {
 pub const ErrorMsg = struct {
     src_loc: SrcLoc,
     msg: []const u8,
-    notes: []ErrorMsg = &.{},
+    notes: []Note = &.{},
```
andrewrk, line 1962: The intent here is that notes can have sub-notes. Are you interested in making a case against this?

andrewrk (changes requested):

src/Compilation.zig
Vexu: No, that would require less changes so I'm not sure why I didn't do that originally.

src/Compilation.zig
Vexu: I would assume most tooling just looks for `path:column:line`. Rust for example emits ` --> src/main.rs:2:14` on a separate line and I'm sure they have considered the IDEs and tooling.

src/Module.zig
Vexu: I just didn't find any examples of notes having notes attached, how would that differ from just having them all linearly?

src/Module.zig
Vexu: > Also, I would expect to see some kind of modification to `Module.destroyDecl`, removing `ReferencedBy` objects. Have you done any testing with incremental compilation on this?

I thought that removing the referencer would also make this decl unreferenced but that obviously only holds up if the decl is referenced from one place.

src/Compilation.zig
andrewrk: I would be satisfied with this reasoning if you double-check one thing for me: try piping the compiler output to a file (not a terminal) and see if the format is different.

src/Module.zig
andrewrk: One example: the referenced by trace could be implemented in this way, having one of the notes be the root referenced by note, and then having each one attached to that one. However I'm not against explicitly handling it with a different data type as you do here.

For terminals, the tree could be represented with indentation. For IDEs it could be represented as a tree view widget. We do have *some* prior art for tree view on compile errors in the case that compiler-rt has compile errors.

src/Compilation.zig
Vexu: Piping does result in the same error message minus the colors but one thing they have that we don't is the ability to emit errors in JSON format.

PR #11535, opened 2022-04-26
llvm: Manually lower softfloat routines when necessary

topolarity:
Pushing this as a mostly-complete draft for now, to better coordinate with the changes in #11532.

There a couple of pieces still missing for stage2: conversion routines (int<->float, float<->float) and f16 support. We'll also need to revise this to match the new naming for non-standard compiler-rt symbols in 11532

### DIFF
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -869,7 +869,7 @@ static LLVMValueRef get_float_fn(CodeGen *g, ZigType *type_entry, ZigLLVMFnId fn
         name = "fma";
         num_args = 3;
     } else if (fn_id == ZigLLVMFnIdFloatOp) {
-        name = float_op_to_name(op);
+        name = float_un_op_to_name(op);
         num_args = 1;
     } else {
         zig_unreachable();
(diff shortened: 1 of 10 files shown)


matu3ba:
ld.lld: error: undefined symbol: fmodq

andrewrk:
cherry-picked into #11532

src/stage1/codegen.cpp
```diff
@@ -3362,152 +3494,129 @@ static LLVMValueRef ir_render_soft_f80_bin_op(CodeGen *g, Stage1Air *executable,
...
-            func_name = "__floorx";
+            sprintf(fn_name, "__div%sf3", compiler_rt_type_abbrev);
             break;
         case IrBinOpRemRem:
-            param_count = 1;
-            func_name = "__remx";
-            break;
         case IrBinOpRemMod:
-            param_count = 1;
-            func_name = "__modx";
```
andrewrk, line 3430: What's up with these changes? This breaks floored division and modulus

andrewrk, line 3430: What's up with these changes? This breaks floored division and modulus

topolarity: I might have introduced a bug here.

This change was trying to re-align the softfloat logic to match `gen_div`, where we perform a division/modulus op and follow it up with extra operations to create the divFloor, etc.

I didn't have this part of compiler-rt up and running yet when I wrote this, so this hasn't been tested. Sorry about that!

andrewrk: gotcha - no problem, I'll fix it up :+1:

topolarity: FWIW, the original code seems strange to me too:

If I grok'd it correctly, it truncates the first operand directly to an integer without actually performing a division. I'm also not sure what functions __modx and __remx refer to, since they are treated as unary here

I might have misunderstood it, though.

andrewrk: oh well, the clock is ticking on this code anyway. For 0.10.0 it will be irrelevant, since we'll ship stage3 for the llvm backend instead of stage1, making llvm.zig the only important lowering for zig users.

PR #11536, opened 2022-04-27
naive attempt to port posix_spawn from musl

matu3ba:
* feasible, but some threading stuff needs to be ported first:
  - missing pieces in Zig libstd for finishing:
    + pthread_setcancelstate
    + pthread_sigmask
    
(or whatever Zig wants to use instead inside Thread.zig)

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -6901,3 +6901,218 @@ pub fn perf_event_open(
         else => |err| return unexpectedErrno(err),
     }
 }
+
+// ported from musl, TODO release
+// TODO make this cross platform (pull in other symbols depending on libc/darwin)
+pub const POSIX_SPAWN_RESETIDS = 0x0001;
+pub const POSIX_SPAWN_SETPGROUP = 0x0002;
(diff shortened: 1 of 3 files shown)


matu3ba:
After getting linker errors with libc, I tried to port posix_spawn to identify the missing pieces.

https://github.com/matu3ba/zig/tree/posix_spawn
`./build/zig test -lc lib/std/child_process.zig`
```
ld.lld: error: undefined symbol: posix_spawn_file_actions_addfchdir_np
```

matu3ba:
Closing. After digging around, I think something like posix_spawn can be implemented with wrappers around `sigprocmask`
(posix_spawn uses `pthread_sigmask`).
As of now and possibly near future Zig will not have (dynamic) suspension points for thread cancellation like pthreads, so we dont need `pthread_setcancelstate`.

Reminder for me:
One signifcant drawback, which also applies to posix_spawn is that it is not multithreading safe https://stackoverflow.com/questions/8152076/spawn-process-from-multithreaded-application.

PR #11537, opened 2022-04-27
Remove atomic builtin wrappers.

kprotty:
Following the idea of "one way to do things", this removes anything in the standard library that wraps the zig atomic builtins. Inline assembly and thread-sanitizer specialization, which was previously shared/exposed via `atomic.Atomic(T)` methods before, is now inlined at the call sites where they're used. Sad to see it go, but this is technically more consistent and means that `@ptrCast(*const Atomic(u32), ...)` isn't required when using `std.Thread.Futex`.


leecannon:
With status quo `stats.count += 1` errors when `count` is `std.atomic.Atomic`, with this change mistaked like this are *extremely* easy to make.

InKryption:
If I understand the "One way to do things" MO, it's not referring to the stdlib, just the features of the core language itself. Doing this makes as much sense to me as removing all use of `std.ArrayList(T)` in favor of manually tracking the length/capacity of a slice that gets realloacted throughout the duration of the program.
Having more friendly wrappers over complicated builtins available in the stdlib is not in contradiction to the aforementioned MO: at the end of the day, regardless of what level of abstraction you choose, you can go down through those layers, and find that it boils down to the same thing.

leecannon:
I think `std.Atomic` is in alignment with more of zig's zens then it is against.

## Communicate intent precisely
The intent that a variable is accessed using atomic's is encoded in its type.

```zig
// This PR
const MyStruct = struct {
    count: usize,
};

// Status Quo
const MyStruct = struct {
    count: std.Atomic(usize),
};
```

## Compile errors are better than bugs.
The combination of "Runtime crashes are better than bugs." and "Compile errors are better than runtime crashes." results in "Compile errors are better than bugs."

With this PR many simple operations on variables that should should be handled atomically are bugs however in status quo using `std.Atomic` it is a compile error.

andrewrk:
My main concern here is that Atomic.zig and related components started to accumulate logic that belonged in the language, all the way down to inline assembly to lower the atomic instructions, when this is actually the compiler's job. If this can be addressed then I think Atomic.zig can be saved.

Here are all of the snippets of code that I am not OK with:

```zig
pub inline fn fence(comptime ordering: Ordering) void {
    switch (ordering) {
        .Acquire, .Release, .AcqRel, .SeqCst => {
            @fence(ordering);
        },
        else => {
            @compileLog(ordering, " only applies to a given memory location");
        },
    }
}
```

This is duplicating compile error logic from `@fence`. This function has no business existing.

```zig
pub inline fn compilerFence(comptime ordering: Ordering) void {
    switch (ordering) {
        .Acquire, .Release, .AcqRel, .SeqCst => asm volatile ("" ::: "memory"),
        else => @compileLog(ordering, " only applies to a given memory location"),
    }
}
```

This should be deleted in favor of `@fence` unless it can have doc comments explaining when you would use one over the other.

```zig
        pub fn init(value: T) Self {
            return .{ .value = value };
        }
```

`value` field should be initialized directly instead of this.

```zig
        pub inline fn fence(self: *Self, comptime ordering: Ordering) void {
            // LLVM's ThreadSanitizer doesn't support the normal fences so we specialize for it.
            if (builtin.sanitize_thread) {
                const tsan = struct {
                    extern "c" fn __tsan_acquire(addr: *anyopaque) void;
                    extern "c" fn __tsan_release(addr: *anyopaque) void;
                };

                const addr = @ptrCast(*anyopaque, self);
                return switch (ordering) {
                    .Unordered, .Monotonic => @compileError(@tagName(ordering) ++ " only applies to atomic loads and stores"),
                    .Acquire => tsan.__tsan_acquire(addr),
                    .Release => tsan.__tsan_release(addr),
                    .AcqRel, .SeqCst => {
                        tsan.__tsan_acquire(addr);
                        tsan.__tsan_release(addr);
                    },
                };
            }

            return std.atomic.fence(ordering);
        }
```

This logic makes the std lib aware of TSAN which is something intended to be only implemented in the compiler. The comment mentions LLVM which is only one particular backend. We need to adjust the definition of `@fence` or add a new builtin, not put such messy functions into the std lib.

```zig
        pub inline fn load(self: *const Self, comptime ordering: Ordering) T {
            return switch (ordering) {
                .AcqRel => @compileError(@tagName(ordering) ++ " implies " ++ @tagName(Ordering.Release) ++ " which is only allowed on atomic stores"),
                .Release => @compileError(@tagName(ordering) ++ " is only allowed on atomic stores"),
                else => @atomicLoad(T, &self.value, ordering),
            };
        }

        pub inline fn store(self: *Self, value: T, comptime ordering: Ordering) void {
            return switch (ordering) {
                .AcqRel => @compileError(@tagName(ordering) ++ " implies " ++ @tagName(Ordering.Acquire) ++ " which is only allowed on atomic loads"),
                .Acquire => @compileError(@tagName(ordering) ++ " is only allowed on atomic loads"),
                else => @atomicStore(T, &self.value, value, ordering),
            };
        }
```

Duplicated compile error logic from the compiler.

```zig
        inline fn cmpxchg(
            self: *Self,
            comptime is_strong: bool,
            compare: T,
            exchange: T,
            comptime success: Ordering,
            comptime failure: Ordering,
        ) ?T {
            if (success == .Unordered or failure == .Unordered) {
                @compileError(@tagName(Ordering.Unordered) ++ " is only allowed on atomic loads and stores");
            }

            comptime var success_is_stronger = switch (failure) {
                .SeqCst => success == .SeqCst,
                .AcqRel => @compileError(@tagName(failure) ++ " implies " ++ @tagName(Ordering.Release) ++ " which is only allowed on success"),
                .Acquire => success == .SeqCst or success == .AcqRel or success == .Acquire,
                .Release => @compileError(@tagName(failure) ++ " is only allowed on success"),
                .Monotonic => true,
                .Unordered => unreachable,
            };

            if (!success_is_stronger) {
                @compileError(@tagName(success) ++ " must be stronger than " ++ @tagName(failure));
            }

            return switch (is_strong) {
                true => @cmpxchgStrong(T, &self.value, compare, exchange, success, failure),
                false => @cmpxchgWeak(T, &self.value, compare, exchange, success, failure),
            };
        }
```

Duplicated compile error logic from the compiler, plus making a different decision than the language does about having strength be a boolean vs a different function.

```zig
        fn exportWhen(comptime condition: bool, comptime functions: type) type {
            return if (condition) functions else struct {};
        }
```

No. This is unnecessary complication and makes the compilation error "function not found" rather than the actual problem regarding the types. I am not OK with this.

```zig
    inline fn lockFast(self: *Impl, comptime casFn: []const u8) bool {
        // On x86, use `lock bts` instead of `lock cmpxchg` as:
        // - they both seem to mark the cache-line as modified regardless: https://stackoverflow.com/a/63350048
        // - `lock bts` is smaller instruction-wise which makes it better for inlining
        if (comptime builtin.target.cpu.arch.isX86()) {
            const locked_bit = @ctz(u32, @as(u32, locked));
            return self.state.bitSet(locked_bit, .Acquire) == 0;
        }

            inline fn bitRmw(self: *Self, comptime op: BitRmwOp, bit: Bit, comptime ordering: Ordering) u1 {
                // x86 supports dedicated bitwise instructions
                if (comptime builtin.target.cpu.arch.isX86() and @sizeOf(T) >= 2 and @sizeOf(T) <= 8) {
                    // TODO: stage2 currently doesn't like the inline asm this function emits.
                    if (builtin.zig_backend == .stage1) {
                        return x86BitRmw(self, op, bit, ordering);
                    }
                }

                const mask = @as(T, 1) << bit;
                const value = switch (op) {
                    .Set => self.fetchOr(mask, ordering),
                    .Reset => self.fetchAnd(~mask, ordering),
                    .Toggle => self.fetchXor(mask, ordering),
                };

                return @boolToInt(value & mask != 0);
            }
```

This lowering logic should be in the compiler not the std lib. We can adjust builtin semantics or introduce new builtins as necessary.

leecannon:
Is it possible that the idea of a piece of memory being intended to be handled atomically should be lifted up into the language itself by adding an atomic qualifier for vaiables/fields as well as pointers? (like volatile/align) 
```zig
var num atomic: usize = 1;
const ptr: *atomic usize = &num;
````

Or even a completely different type (like optional) 
```zig
var num: atomic usize = 1;
const ptr: *atomic usize = &num;
````

It would then be a compile error to load or store the memory using standard non-atomic operations, it may be necessary to add `@nonAtomicLoad` and `@nonAtomicStore`.

Doing this would allow deleting `std.Atomic` but moving all of it's nice properites into the laguage itself.

Also this would allow much improved translation of c code using `_Atomic`.

kprotty:
The current implementation (not API) of `std.atomic.Atomic` grew to how it is from lack of confidence on my side to modify stage1: e.g. the error messages for the Orderings *could* be in the builtins, but I couldn't figure out how to add them there at the time. 

> This should be deleted in favor of @fence unless it can have doc comments

`compilerFence` is different from `fence` in that it only prevents compiler reordering in the current thread rather than preventing reordering across threads like the latter. You could say it enforces ["sequenced-before"](https://en.cppreference.com/w/cpp/atomic/memory_order#Sequenced-before) without enforcing ["happens-before"](https://en.cppreference.com/w/cpp/atomic/memory_order#Happens-before) like normal fence(). It's equivalent to [`atomic_signal_fence`](https://en.cppreference.com/w/cpp/atomic/atomic_signal_fence) API-wise and LLVM's [fence](https://llvm.org/doxygen/classllvm_1_1FenceInst.html) instruction implementation-wise takes a scope for single-threaded to properly support it.

> We need to adjust the definition of @fence or add a new builtin, not put such messy functions into the std lib.

This is tricky. One one hand, we would deviate more from C11's API and could do so by having `@fence()` take in an optional (array of?) address or something to say "this fence synchronizes with the normal stuff, but also explicitly these". On the other hand, all other code out there [even rust's stdlib](https://github.com/rust-lang/rust/blob/e4f5b15b8832334eca2c0bd3929eb3969f1d166d/library/alloc/src/sync.rs#L52-L67) do specialization for TSAN fence's because this is primarily a TSAN bug with no clear method of resolution: (fence synchronizes-with other atomics but tsan doesn't know which ones so it falsely reports races).

> plus making a different decision than the language does about having strength be a boolean

AcqRel being required over Release when paired with Acquire failure ordering was a decision adopted from Rust due to idea that CAS is implicitly an LL/SC loop meaning a successful committing store implied the initial Acquire load thus AcqRel. Seems like this is no longer the case.

The `strength` being a bool is just an internal detail to `Atomic(T)` only for calling that function easier. It could be an `enum` or whatever but it's purpose is to avoid code duplication.

> This is unnecessary complication and makes the compilation error "function not found" rather than the actual problem regarding the types.

Fair point. The functions could always exist them `@compileError` if they're on an unsupported type.

> This lowering logic should be in the compiler not the std lib. We can adjust builtin semantics or introduce new builtins as necessary.

We would be introducing a new builtin i'd imagine: `@atomicBitRmw(Int: type, ptr: *Int, op: enum{Set,Reset,Toggle}, bit: Log2Int(Int)): u1`. This again has the issue of not being familiar in the atomic space both to your average C11 or LLVM atomic enjoyers.

> should be lifted up into the language itself

I'm a fan of this idea if we end up removing `std.atomic.Atomic()`, although it's more work implementation wise. Would also take that opportunity to have `@volatileLoad` and `@volatileStore` instead of volatile being a type qualifier.

matu3ba:
@leecannon 

What additional usage does ensuring only `atomic` types can have `store`, `load` `exchange`, `compare_exchange_weak`, `compare_exchange_strong` `notify_one`, `notify_all` (the notification ones not for _Atomic semantics) bring to the table?
https://en.cppreference.com/w/c/language/atomic

Signaling is completely dependent on the platform, which I believe should be then in libstd.

kprotty:
@matu3ba "Signaling" is indeed standard library thing rather than a language thing. `atomic_wait` and `atomic_notify` as seen in C++ shouldn't be apart of the language but instead the stdlib as they are now via `std.Thread.Futex`.

andrewrk:
@kprotty I feel like I understand your goals and you understand mine. What's the next step I can take to help unblock you on this?

leecannon:
Sorry for repeating myself, but I will just like to clarify my point.

Currently there is a way of communicating intent that a variable should be handled atomically and makes it so that using normal arithmetic operators on it is a compile error.

With this PR as is, we lose the very nice property that `usize != Atomic(usize)` which in my actual project's has turned bugs that I unwittingly introduced into compile errors.

gracefuu:
I think a summary of the discussion so far is:

1) `atomic.Atomic` results in redundancy between compiler and stdlib, which is bad.
2) `atomic.Atomic` allows doing atomic operations concisely, which is bad*.
3) `atomic.Atomic` provides type safety, which is good.
4) There may be better ways to provide type safety without downsides like (1) and (2).

*Not mentioned in this thread, but according to @andrewrk in #stdlib-devel on the Zig discord,
> I think [doing atomic ops concisely] is an anti-feature - a small amount of friction and obviousness of atomic ops should be present. this was a conscious design decision I made when adding the atomic builtins

To address (4), one suggestion (@leecannon) was to add an `atomic` qualifier similar to `volatile`. This is more in line with C11 `_Atomic` and `volatile`, which is good.

----

Now for my opinion: I think we should keep `atomic.Atomic` FOR NOW.

I think due to @andrewrk's points, it's probably best to remove `atomic.Atomic`, but we still want some way of achieving type safety. @leecannon's proposal is one option.

But whatever option we choose, we will want to perform a second round of refactoring to change all uses of atomics to use the type safe versions. For that reason, it might be better to keep `atomic.Atomic` for now, until the replacement is ready. That would make refactoring current uses of `atomic.Atomic` much easier when the time comes.

kprotty:
Closed to be re-evaluated later.

PR #11542, opened 2022-04-28
std.testing: add assumption and notes for TmpDir

matu3ba:
* adds copy-paste to create relative path as comment
* user should use file handles, if possible

### DIFF
--- a/lib/std/testing.zig
+++ b/lib/std/testing.zig
@@ -347,6 +347,10 @@ pub fn expect(ok: bool) !void {
     if (!ok) return error.TestUnexpectedResult;
 }
 
+/// assume: user does not change cwd
+/// For relative paths, use
+/// `std.fs.path.join(alloc, &[_][]const u8{ "zig-cache", "tmp", &tmp_dir.sub_path, ..});`
+/// Use file handles, if possible.
 pub const TmpDir = struct {
     dir: std.fs.Dir,
     parent_dir: std.fs.Dir,
--- a/lib/std/child_process.zig
+++ b/lib/std/child_process.zig
@@ -1324,7 +1324,7 @@ test "createNullDelimitedEnvMap" {
     }
 }
 
-const childstr =
+const child_prog =
     \\ const std = @import("std");
     \\ const builtin = @import("builtin");
     \\ pub fn main() !void {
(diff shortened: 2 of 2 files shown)


matu3ba:
copy of old PR in #11341 to clarify remaining issues.

Vexu: Stage1 can't parse doctests currently.

Me(copy from before):
I am fine with either decision, but I would prefer having some reason(s).

1. Are you mostly concerned about the restriction to write files with ending `.zig` ?
2. Or is there a more fundamental problem with tmp files generated and being run from the test runner?

2.1 In the latter case, more tailored/numerous IPC tests must have a 1. numbering schema or we will have 2. cluttered folder or 3. a lot duplicated code.
2.2 Alternatively, IPC is supposed to be (for now) not tested.

lib/std/testing.zig
```diff
@@ -374,6 +374,43 @@ pub const TmpDir = struct {
...
+    /// var it = try std.process.argsWithAllocator(std.testing.allocator);
+    /// defer it.deinit(); // no-op unless WASI or Windows
+    /// const testargs = try std.testing.getTestArgs(&it);
+    /// var tmp = std.testing.tmpDir(.{ .no_follow = true }); // ie zig-cache/tmp/8DLgoSEqz593PAEE
+    /// defer tmp.cleanup();
+    /// const zigfile_path = try tmp.writeZigFile(std.testing.allocator, progstr, "bruh");
+    /// defer std.testing.allocator.free(zigfile_path);
+    /// const binary = zigfile_path[0 .. zigfile_path.len - 4]; // '.zig' is 4 characters
+    /// try std.testing.buildExe(testargs.zigexec, zigfile_path, binary);
+    /// ```
```
andrewrk, line 392: this belongs as a doctest. if you are running into problems with that, then do one of these things:
 * fix them as a prerequisite to this PR
 * work around the issue by using `test "foo" {` syntax instead of `test foo {` syntax.

lib/std/testing.zig
```diff
@@ -374,6 +374,43 @@ pub const TmpDir = struct {
...
+    pub fn writeZigFile(
```
andrewrk, line 393: I don't think a function called `writeZigFile` belongs in std.testing. If it was a more general-purpose method for writing a temp file  for a test, then I would be OK with it.

lib/std/testing.zig
```diff
@@ -374,6 +374,43 @@ pub const TmpDir = struct {
...
+        const zigfile_path = try std.mem.concat(alloc, u8, &[_][]const u8{
+            tmpdir_path,
+            std.fs.path.sep_str,
+            filename,
+            suffix_zig,
+        });
+        errdefer alloc.free(zigfile_path);
+        const zigfile = try std.mem.concat(alloc, u8, &[_][]const u8{ filename, suffix_zig });
+        defer alloc.free(zigfile);
+        try self.dir.writeFile(zigfile, progstr);
```
andrewrk, line 411: This unfortunately ignores dir and uses an absolute path. Ideally instead it will write a relative name based on dir.

andrewrk (changes requested):

lib/std/testing.zig
matu3ba: The underlying problem is that `tmpDir(opts: std.fs.Dir.OpenDirOptions) TmpDir` creates dirs based on `cwd` with `cwd/cache/tmp/base64random()`, but `cwd` is not stored in `tmpDir`.

The 4 options are:
1. compute absolute path to compute relative based on `cwd`
    - 1.1 new method getRelativePath identical to getFullPath except for applying `std.fs.path.relative` to get relative path
    - 1.2 use getFullPath() + another `cwd` + `std.fs.path.relative`
2. store `cwd` in struct `TmpDir`
3. remove all the path stuff with comments that user should use `dir.createFile()/closeFile()`: slight/big annoyance here one is forced to use `/// Call `File.close` on the result when done.`
    - Biggest drawback: Users will likely reimplement the "getmepath" functions for debugging stuff.

matu3ba:
force pushed with reapplying changes due to merge conflicts

lib/std/testing.zig
matu3ba: workaround was used.

lib/std/testing.zig
matu3ba: agreed. changed to writeFile.

lib/std/testing.zig
matu3ba: Nevermind actually. Only thing I need is
```zig
        const relative_path = try fs.path.join(allocator, &[_][]const u8{ "zig-cache", "tmp", tmp.sub_path[0..] });
        break :blk try fs.realpathAlloc(allocator, relative_path);
```

matu3ba:
I removed the unnecessary code and instead added a comment to TmpDir to nudge the user in the correct direction:
```zig
/// assume: user does not change cwd
/// For relative paths, use
/// `std.fs.path.join(alloc, &[_][]const u8{ "zig-cache", "tmp", &tmp_dir.sub_path, ..});`
/// Use file handles, if possible.
```

Other code is a test cleanup.

PR #11546, opened 2022-04-29
Seccomp fixups re: #10717

The-King-of-Toasters:
- Add type annotation for `AUDIT.current`.
- Make unsupported archs a compile error.


andrewrk:
Thanks for the follow-up!

PR #11548, opened 2022-04-29
stage2,llvm: handle softfloats in @intToFloat and @floatToInt

kubkon:
If the hw doesn't have support for exotic floating-point types such
as `f80`, we lower the call to a compiler-rt function call instead.

I've added a behavior test specifically targeting this use case which
now passes on `aarch64-macos`.

Additionally, this commit makes it possible to successfully build
stage3 on `aarch64-macos`. We can print the compiler's help message,
however, building with it needs a little bit more love still.

### DIFF
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -4188,41 +4188,166 @@ pub const FuncGen = struct {
         return self.builder.buildInsertValue(partial, len, 1, "");
     }
 
+    inline fn isPowerOfTwo(bits: u64) bool {
+        return bits != 0 and ((bits & (~bits + 1)) == bits);
+    }
+
+    fn intTypeFromBitsAndSignRounded(self: *FuncGen, bits: u16, signed: bool) error{OutOfMemory}!Type {
+        const next_pow_two = math.log2_int_ceil(u16, bits);
+        const rounded_bits = @as(u32, 1) << next_pow_two;
+        return switch (rounded_bits) {
+            8, 16, 32 => if (signed) Type.initTag(.i32) else Type.initTag(.u32),
+            64 => if (signed) Type.initTag(.i64) else Type.initTag(.u64),
+            128 => if (signed) Type.initTag(.i128) else Type.initTag(.u128),
+            else => |big| if (signed)
+                Type.Tag.int_signed.create(
+                    self.dg.object.type_map_arena.allocator(),
+                    @intCast(u16, big),
+                )
+            else
+                Type.Tag.int_unsigned.create(
+                    self.dg.object.type_map_arena.allocator(),
+                    @intCast(u16, big),
+                ),
+        };
+    }
+
     fn airIntToFloat(self: *FuncGen, inst: Air.Inst.Index) !?*const llvm.Value {
         if (self.liveness.isUnused(inst))
             return null;
 
+        const target = self.dg.module.getTarget();
         const ty_op = self.air.instructions.items(.data)[inst].ty_op;
-        const operand = try self.resolveInst(ty_op.operand);
-        const operand_ty = self.air.typeOf(ty_op.operand);
-        const operand_scalar_ty = operand_ty.scalarType();
+
+        var operand = try self.resolveInst(ty_op.operand);
+        var operand_ty = self.air.typeOf(ty_op.operand);
+        var operand_scalar_ty = operand_ty.scalarType();
+
+        {
+            const operand_bits = @intCast(u16, operand_scalar_ty.bitSize(target));
+            const is_signed = operand_scalar_ty.isSignedInt();
+
+            if (!isPowerOfTwo(operand_bits) or operand_bits < 32) {
+                const wider_ty = try self.intTypeFromBitsAndSignRounded(operand_bits, is_signed);
+                const wider_llvm_ty = try self.dg.llvmType(wider_ty);
+                if (is_signed) {
+                    operand = self.builder.buildSExt(operand, wider_llvm_ty, "");
+                } else {
+                    operand = self.builder.buildZExt(operand, wider_llvm_ty, "");
+                }
+                operand_ty = wider_ty;
+                operand_scalar_ty = operand_ty.scalarType();
+            }
+        }
+
         const dest_ty = self.air.typeOfIndex(inst);
+        const dest_scalar_ty = dest_ty.scalarType();
         const dest_llvm_ty = try self.dg.llvmType(dest_ty);
 
-        if (operand_scalar_ty.isSignedInt()) {
-            return self.builder.buildSIToFP(operand, dest_llvm_ty, "");
-        } else {
-            return self.builder.buildUIToFP(operand, dest_llvm_ty, "");
+        if (intrinsicsAllowed(dest_scalar_ty, target)) {
+            if (operand_scalar_ty.isSignedInt()) {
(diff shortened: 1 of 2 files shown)


test/behavior/cast.zig
```diff
@@ -112,6 +112,30 @@ test "@intToFloat" {
...
+
+    if (builtin.zig_backend == .stage2_aarch64) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_x86_64) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_arm) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_wasm) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_c) return error.SkipZigTest; // TODO
+
+    const S = struct {
+        fn doTheTest() !void {
+            try testIntToFloat(-2);
```
andrewrk, line 126: let's test each of the 4 sizes supported by compiler-rt as well as a non-power-of-two bitwidth in between each one

src/codegen/llvm.zig
```diff
@@ -4192,37 +4192,145 @@ pub const FuncGen = struct {
...
+            @panic("TODO @intToFloat: handle integers of non-pot bitsize by widening them");
+        } else source;
+
+        const int_bits = @intCast(u16, int_scalar_ty.bitSize(target));
+        const compiler_rt_int_abbrev = compilerRtAbbrev(int_bits);
+
+        const float_bits = float_scalar_ty.floatBits(target);
+        const compiler_rt_float_abbrev = compilerRtAbbrev(float_bits);
+
+        var fn_name_buf: [64]u8 = undefined;
```
andrewrk, line 4237: Consider whether you want to have this be separate logic (as it is here) or whether you want to add int_to_float/float_to_int to the `buildFloatOp` function

andrewrk (changes requested):

src/codegen/llvm.zig
```diff
@@ -4192,37 +4192,145 @@ pub const FuncGen = struct {
...
+    ) Allocator.Error!*const llvm.Value {
+        const target = self.dg.module.getTarget();
+        const int_scalar_ty = int_ty.scalarType();
+        const float_scalar_ty = float_ty.scalarType();
+
+        const opt_abi_ty = self.dg.getAtomicAbiType(int_scalar_ty, false);
+        const operand: *const llvm.Value = if (opt_abi_ty) |_| {
+            // source operand needs widening and truncating
+            @panic("TODO @intToFloat: handle integers of non-pot bitsize by widening them");
+        } else source;
```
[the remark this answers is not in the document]
kubkon: The reason I used `getAtomicAbiType` is my lack of knowledge of the LLVM backend - I was looking for a check for widening the operand while following the implementation in stage1 :-)

src/codegen/llvm.zig
kubkon: I was thinking of extracting common logic in `airIntToFloat` and `airFloatToInt` as we do share quite a bit there between the two, but I will look into it after I have all of your other comments addresses first, and in particular, handling of widening of input/output operands so that we have a good idea what really can be extracted out.

test/behavior/cast.zig
kubkon: Done!

src/codegen/llvm.zig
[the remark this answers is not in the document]
kubkon: Done!

kubkon:
@andrewrk ready for re-review! One thing I wasn't sure about was explicitly checking for vector type as the operand type - this check and handling is present in stage1's implementation, however, I fail to understand how and if it would ever be used since `@intToFloat(f80, std.meta.Vector(i64, 3){ -1, -2, -3 })` is not legal AFAIK. Shedding any light on this would be greatly appreciated!

src/codegen/llvm.zig
```diff
@@ -4188,41 +4188,166 @@ pub const FuncGen = struct {
         return self.builder.buildInsertValue(partial, len, 1, "");
     }
 
+    inline fn isPowerOfTwo(bits: u64) bool {
+        return bits != 0 and ((bits & (~bits + 1)) == bits);
+    }
```
andrewrk, line 4193: `std.math.isPowerOfTwo`

src/codegen/llvm.zig
```diff
@@ -4188,41 +4188,166 @@ pub const FuncGen = struct {
...
+
+    fn intTypeFromBitsAndSignRounded(self: *FuncGen, bits: u16, signed: bool) error{OutOfMemory}!Type {
```
andrewrk, line 4195: I think this function is awkward and unnecessary. Allow me to suggest an alternate implementation at the callsites...

src/codegen/llvm.zig
```diff
@@ -4188,41 +4188,166 @@ pub const FuncGen = struct {
...
+                const wider_llvm_ty = try self.dg.llvmType(wider_ty);
+                if (is_signed) {
+                    operand = self.builder.buildSExt(operand, wider_llvm_ty, "");
+                } else {
+                    operand = self.builder.buildZExt(operand, wider_llvm_ty, "");
+                }
+                operand_ty = wider_ty;
+                operand_scalar_ty = operand_ty.scalarType();
+            }
+        }
```
andrewrk, line 4241: All this logic should be *after* the `intrinsicsAllowed` check. If the intrinsics are allowed, then we allow LLVM to see the irregular bit width integer types.

src/codegen/llvm.zig
```diff
@@ -4188,41 +4188,166 @@ pub const FuncGen = struct {
...
+        const fn_name = if (operand_scalar_ty.isSignedInt())
+            std.fmt.bufPrintZ(&fn_name_buf, "__float{s}i{s}f", .{
+                compiler_rt_operand_abbrev,
+                compiler_rt_dest_abbrev,
+            }) catch unreachable
+        else
+            std.fmt.bufPrintZ(&fn_name_buf, "__floatun{s}i{s}f", .{
+                compiler_rt_operand_abbrev,
+                compiler_rt_dest_abbrev,
+            }) catch unreachable;
```
andrewrk, line 4271: Here's a way to do it instead of `intTypeFromBitsAndSignRounded`:

```suggestion
        // implementation of toCompilerRtIntType left as an exercise for the reader
        // idea is that it returns a `*const llvm.Value` rather than a type, and does the extending if necessary.
        const operand = self.toCompilerRtIntType(ty_op.operand);
        // note: add to bindings.zig the following function:
        // pub extern fn LLVMSizeOfTypeInBits(TD: LLVMTargetDataRef, Ty: LLVMTypeRef) c_ulonglong;
        //
        const operand_bits = operand.typeOf().sizeOfTypeInBits();
        const sign_prefix = if (operand_scalar_ty.isSignedInt()) "" else "un";
        var fn_name_buf: [64]u8 = undefined;
        const fn_name = std.fmt.bufPrintZ(&fn_name_buf, "__float{s}{s}i{s}f", .{
            sign_prefix,
            compiler_rt_operand_abbrev,
            compiler_rt_dest_abbrev,
        }) catch unreachable;
```

andrewrk (changes requested):
I can think of a way to restructure the code that I think is enough of an improvement that in my opinion it's worth blocking the merge for.

andrewrk:
Cherry-picked into #11492

PR #11551, opened 2022-04-29
std: Do not allocate the result for ChildProcess.init

Hejsil:
Instead, just return ChildProcess directly. This structure does not
require a stable address, so we can put it on the stack just fine. If
someone wants it on the heap they should do.

```zig
const proc = try allocator.create(ChildProcess);
proc.* = ChildProcess.init(args, allocator);
```

### DIFF
--- a/lib/std/child_process.zig
+++ b/lib/std/child_process.zig
@@ -98,10 +98,8 @@ pub const ChildProcess = struct {
     };
 
     /// First argument in argv is the executable.
-    /// On success must call deinit.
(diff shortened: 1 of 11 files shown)


lib/std/child_process.zig
```diff
@@ -99,9 +99,8 @@ pub const ChildProcess = struct {
 
     /// First argument in argv is the executable.
     /// On success must call deinit.
```
erikarvstedt, line 101:
```suggestion
```

andrewrk (approved):

PR #11553, opened 2022-04-29
std.MultiArrayList: add functions `addOne`, `pop`, and `popOrNull`

InKryption:
I keep on writing these behaviors by hand, so I thought it'd be nice if MultiArrayList just exposed them as part of the API.


andrewrk (approved):
Nice, thank you!

PR #11555, opened 2022-04-30
std.io: Add `Reader.readLEB128` and `Writer.writeLEB128`

r00ster91:
I think it'd be pretty nice to add some helpers for reading and writing LEB128 values.
LEB128 values in the wild definitely happen and this makes parsing such byte streams using `std.io.Reader` easier and it makes this functionality more accessible. All kinds of byte formats use LEB128 values because they're so efficient.
It also takes the burden of the choice between the unsigned or signed version off your shoulders.

Additionally, this might also encourage the usage of LEB128 a little bit, when for example creating your own format.

Later, for consistency, I also added `std.io.Writer.writeLEB128`.

### DIFF
--- a/lib/std/io/reader.zig
+++ b/lib/std/io/reader.zig
@@ -274,6 +274,16 @@ pub fn Reader(
             return mem.readInt(T, &bytes, endian);
         }
 
+        /// Reads a LEB128 (Little Endian Base 128) value.
+        ///
+        /// See also: `std.leb` and https://en.wikipedia.org/wiki/LEB128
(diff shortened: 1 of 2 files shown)


andrewrk:
I don't think such DWARF-specific functions belong in the general purpose Reader/Writer interfaces. Calling into the std.leb namespace is perfectly reasonable.

r00ster91:
It has a lot of uses outside of DWARF as well: https://en.wikipedia.org/wiki/LEB128#Uses, including WebAssembly. And maybe you want to make your own format and want to efficiently write integers.
But I still see your point that maybe it's a bit too much.

PR #11556, opened 2022-04-30
fmt: Ensure space before comment content

r00ster91:
This PR makes it so that
```zig
pub fn main() void {
    //hello
    //     zig
    //world
}
```
becomes
```zig
pub fn main() void {
    // hello
    //     zig
    // world
}
```
when formatted. This also includes top-level comments and documentation comments.

---

I think it's clear as day that the agreed-upon conventional style for comments is to put a space before the comment content. I mean, where do you not see this done?

In `zig/`, for normal comments alone (excluding top-level comments and documentation comments),
`grep -r -I "// " . | wc -l` gives 91743,
`grep -r -I '^ *//[^ /!\n]' . | wc -l` gives 2363.

So as you surely noticed, an overwhelming majority does it the `// hello` way.
But then you have that small minority of code which writes comments like `//hello` without a space. It looks inconsistent, squished, and not very nice. And that is exactly what the formatter is for: make code look consistent and nice, so why don't we make `zig fmt` format comments properly and put a space before the comment content?
And at the end of the day, you don't have to use `zig fmt` of course.


lib/std/zig/render.zig
```diff
@@ -2428,7 +2428,11 @@ fn renderComments(ais: *Ais, tree: Ast, start: usize, end: usize) Error!bool {
             ais.disabled_offset = index;
         } else {
             // Write the comment minus trailing whitespace.
-            try ais.writer().print("{s}\n", .{trimmed_comment});
+            if (mem.startsWith(u8, trimmed_comment, "// ") or comment_content.len == 0) {
+                try ais.writer().print("{s}\n", .{trimmed_comment});
+            } else {
+                try ais.writer().print("// {s}\n", .{trimmed_comment["//".len..]});
```
r00ster91, line 2434: I'm wondering if, just like below, the trailing `\n` here shouldn't be added using something like
`try renderSpace(ais, tree, tok, token_slice.len, .newline);` instead?

lib/std/zig/parser_test.zig
```diff
@@ -5439,9 +5483,9 @@ test "ampersand" {
     });
 
     try testError(
-        \\ _ = bar & &foo, 
-        \\ _ = bar & &&foo, 
-        \\ _ = &&foo,
```
r00ster91, line 5444: As for these trailing whitespace removals, my editor did that and these changes are unrelated. But I thought, while I was at it, why not remove some useless trailing whitespace? I hope they are useless, at least.
If this should be done in another PR, please tell me.

rohlem:
> I mean, where do you not see this done?

Just my 2 cents, feel free to dismiss,
I usually choose not to add a space if the comment only applies to the current line of code.
```zig
defer state.deinit(); //technically unnecessary, since we currently only leave this scope on program exit

if(config.xyz) { // in this case we also need to jkl
 //...
}
```
I also don't add a space if I comment out code (since that would be one extra character to remove when I uncomment it back in).

r00ster91:
> I usually choose not to add a space if the comment only applies to the current line of code.

That's an interesting style that I've never really seen a lot.

> I also don't add a space if I comment out code (since that would be one extra character to remove when I uncomment it back in).

In all editors I've used when you press Ctrl+/ on a line, it turns `x();` into `// x();` (with a space already there) and similarly the other way around if you press Ctrl+/ on `// x();` it will turn it back into `x();` without any space. Modern editors know about the convention of the space and will handle it accordingly.

> Though commented-out code should be avoided in finished/committed code, and the intended usage of `zig fmt` during development, in tandem with text editors / IDEs, is probably underdefined.

Yes, you are right. Zig is the kind of language that is so simple you don't even have any warnings. It's either an error or it works. And that's similar here: you either have a properly looking comment or you don't (if of course you use `zig fmt`).

haze:
I'd leave my own +1 for this;
- 2 slashes & no space = multiline string literal
- 2 slashes & space = comment

of course the slashes are backwards, but it's easier to tell through a glance with a space (and it's formatting, so differentiation is helpful)

r00ster91:
> of course the slashes are backwards, but it's easier to tell through a glance with a space (and it's formatting, so differentiation is helpful)

Yes, you are right. That's another problem this solves.
It makes differentiating between multiline string literals `\\hello` and a comment `// hello` a lot easier, especially when you don't have colored syntax highlighting available.

lib/std/zig/render.zig
```diff
@@ -2428,7 +2428,11 @@ fn renderComments(ais: *Ais, tree: Ast, start: usize, end: usize) Error!bool {
             ais.disabled_offset = index;
         } else {
             // Write the comment minus trailing whitespace.
-            try ais.writer().print("{s}\n", .{trimmed_comment});
+            if (mem.startsWith(u8, trimmed_comment, "// ") or comment_content.len == 0) {
+                try ais.writer().print("{s}\n", .{trimmed_comment});
+            } else {
+                try ais.writer().print("// {s}\n", .{comment_content});
+            }
```
r00ster91, line 2435: I'm wondering if, like below, the trailing `\n` in these two branches shouldn't be added using something like
`try renderSpace(ais, tree, tok, token_slice.len, .newline);` instead?
As you can see, the original code didn't do that so I didn't do it either, but I'm wondering if there's any advantage of using this more sophisticated `renderSpace` function to do it?

Vexu (approved):
To pass the CI just run `zig fmt . --exclude test/cases` (preferably in a different commit).

r00ster91:
@Vexu I hope it works now. I recommend reviewing by commit now.

r00ster91:
Apparently a check was not successful but if I look into the logs of that test I find this at the end:

> `2091 passed; 59 skipped; 0 failed.`

The failure seems unrelated to this PR.

rohlem:
The CI lists a run time of exactly 2h, I'd guess that means it just timed out.

r00ster91:
Other PRs are affected by this to so it shouldn't be a concern.

r00ster91:
@Vexu I've resolved the merge conflicts. It seem this PR is rather conflict-prone. Is there any chance we can get this merged soon?

andrewrk (commented):
I'm not convinced of this change. I see two problems with it:

For commented out code it is is debatable whether the space is desirable. Personally, I would prefer no space, as you can see from the commented out code that I am responsible for in this patch, since it shows a "seam" between comments above the commented-out code and the commented-out code itself. It also leaves no room for ambiguity about where the commented out code was supposed to be indented to.

The other problem is evident in the example provided in the description of this PR. In the example, one of the lines has a space after the `//` and is therefore not adjusted to have an extra space inserted. However the surrounding lines have a space inserted. If there was any kind of alignment happening within the comment, then the proposed adjustment via `zig fmt` will cause misalignment.

Improperly formatted comments have no influence whether a bug is caught or not. If anything, they carry more information, about how careful someone was or wasn't when making the changes, which is useful meta-information.

Given that the merits are not clear cut, and given the fact that having this enhancement is additional complexity to `zig fmt` (as opposed to removing a feature), I am in favor of not proceeding with this change.

r00ster91:
Fair enough. At least I was able to gain some experience with the internal workings of Zig. That's the reason I didn't open an issue first, I just wanted to try implementing it right away. Thanks for the input!

PR #11557, opened 2022-04-30
Path rename

xpyxel:
Currently the naming scheme for the source of `std.build.Pkg` is extremely redundant and inconsistent with the struct used to contain the source (`FileSource` in specific). This merge request is breaking change discussed in the Discord that changes the `std.build.Pkg.path` field to be named `std.build.Pkg.source`, making it easier to read in the long run, as well as being consistent.

```zig
const Pkg = @import("std").build.Pkg;

// old
const hello = Pkg{ .name = "hello", .path = .{ .path = "lib/hello.zig" } };

// new
const hello = Pkg{ .name = "hello", .source = .{ .path = "lib/hello.zig" } };
```

I have ran tests to ensure everything builds correctly, as well as built stage 2 and 3 to ensure all stages build.

### DIFF
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -385,7 +385,7 @@ pub const Builder = struct {
     pub fn dupePkg(self: *Builder, package: Pkg) Pkg {
         var the_copy = Pkg{
             .name = self.dupe(package.name),
-            .path = package.path.dupe(self),
+            .source = package.source.dupe(self),
         };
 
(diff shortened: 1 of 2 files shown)


MasterQ32:
As said on discord already: This is a good change and an oversight from my side. I was never happy with the `.path = .{ .path = "`.

andrewrk:
Thank you @xpyxel!

PR #11560, opened 2022-05-01
std.os: add timerfd_create, timerfd_settime and timerfd_gettime

thiagotps:
Currently `std.os` doesn't expose the functions `timerfd_create`, `timerfd_settime` and `timerfd_gettime`, which are only available via their primitives methods in `std.os.linux`. I added their equivalent functions in `std.os` with a proper calling signature and error handling.  I also added a test to `os.test.zig` to check the correct behavior of those functions.


ifreund (approved):
LGTM, Thanks!

PR #11565, opened 2022-05-02
std.os: ported signal handling in abort() from musl

matu3ba:
* Document deviation from Linux man page, which is identical to musl.
  Man page wants always enabled user-provided abort handlers.
  Worst case logic bug, which this can introduce:
    + user disables SIGABRT handler to prevent tear down to last safe
      state
    + abort() gets called and enables user-provided SIGABRT handler
    + SIGABRT tears down to supposed last safe state instead of crash
    + Application, instead of crashing, continues
* Pid 1 within containers needs special handling.
  - fatal signals are not transmitted without privileges,
    so use exit as fallback
* Fix some signaling bits
* Add checks in Debug and ReleaseSafe for wrong sigprocmask

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -250,15 +250,6 @@ pub var argv: [][*:0]u8 = if (builtin.link_libc) undefined else switch (builtin.
     else => undefined,
 };
 
-/// Atomic to guard correct program teardown in abort()
-var abort_entered = impl: {
-    if (builtin.single_threaded) {
-        break :impl {};
-    } else {
-        break :impl std.atomic.Atomic(bool).init(false);
-    }
-};
-
 /// To obtain errno, call this function with the return value of the
 /// system function call. For some systems this will obtain the value directly
 /// from the return code; for others it will use a thread-local errno variable.
@@ -451,7 +442,7 @@ fn getRandomBytesDevURandom(buf: []u8) !void {
 /// Causes abnormal process termination.
 /// If linking against libc, this calls the abort() libc function. Otherwise
 /// it raises SIGABRT followed by SIGKILL and finally lo
-/// assume: Current signal handler for SIGABRT does **not call abort**.
+/// Invokes the current signal handler for SIGABRT, if any.
 pub fn abort() noreturn {
     @setCold(true);
     // MSVCRT abort() sometimes opens a popup window which is undesirable, so
@@ -464,49 +455,44 @@ pub fn abort() noreturn {
         windows.kernel32.ExitProcess(3);
     }
     if (!builtin.link_libc and builtin.os.tag == .linux) {
-        // Linux man page wants to first "unblock SIG.ABRT", but this is a footgun
+        // The Linux man page says that the libc abort() function
+        // "first unblocks the SIGABRT signal", but this is a footgun
         // for user-defined signal handlers that want to restore some state in
-        // some program sections and crash in others
-
-        // user installed SIGABRT handler is run, if installed
+        // some program sections and crash in others.
+        // So, the user-installed SIGABRT handler is run, if present.
         raise(SIG.ABRT) catch {};
 
-        // disable all signal handlers
+        // Disable all signal handlers.
         sigprocmask(SIG.BLOCK, &linux.all_mask, null);
 
-        // ensure teardown by one thread
+        // Only one thread may proceed to the rest of abort().
         if (!builtin.single_threaded) {
-            while (abort_entered.compareAndSwap(false, true, .SeqCst, .SeqCst)) |_| {}
+            const global = struct {
+                var abort_entered: bool = false;
+            };
+            while (@cmpxchgWeak(bool, &global.abort_entered, false, true, .SeqCst, .SeqCst)) |_| {}
         }
 
-        // install default handler to terminate
+        // Install default handler so that the tkill below will terminate.
         const sigact = Sigaction{
             .handler = .{ .sigaction = SIG.DFL },
             .mask = undefined,
             .flags = undefined,
             .restorer = undefined,
         };
-        sigaction(SIG.ABRT, &sigact, null) catch unreachable;
+        sigaction(SIG.ABRT, &sigact, null) catch |err| switch (err) {
+            error.OperationNotSupported => unreachable,
+        };
 
-        // make sure we have a pending SIGABRT queued
-        const tid = std.Thread.getCurrentId();
-        _ = linux.tkill(@intCast(i32, tid), SIG.ABRT);
+        _ = linux.tkill(linux.gettid(), SIG.ABRT);
 
-        // SIG.ABRT signal will run default handler
         const sigabrtmask: linux.sigset_t = [_]u32{0} ** 31 ++ [_]u32{1 << (SIG.ABRT - 1)};
-        sigprocmask(SIG.UNBLOCK, &sigabrtmask, null); // [32]u32
+        sigprocmask(SIG.UNBLOCK, &sigabrtmask, null);
 
-        // Beyond this point should be unreachable
-
-        // abnormal termination without using signal handler
-        const nullptr: *allowzero volatile u8 = @intToPtr(*allowzero volatile u8, 0);
-        nullptr.* = 0;
-
-        // try SIGKILL, which is no abnormal termination as defined by POSIX and ISO C
+        // Beyond this point should be unreachable.
+        @intToPtr(*allowzero volatile u8, 0).* = 0;
         raise(SIG.KILL) catch {};
-
-        // pid 1 might not be signalled in some containers
-        exit(127);
+        exit(127); // Pid 1 might not be signalled in some containers.
     }
     if (builtin.os.tag == .uefi) {
         exit(0); // TODO choose appropriate exit code
@@ -5488,13 +5474,12 @@ pub fn sigaction(sig: u6, noalias act: ?*const Sigaction, noalias oact: ?*Sigact
     }
 }
 
-/// Set the thread signal mask
-/// Invalid masks are checked in Debug and ReleaseFast
+/// Sets the thread signal mask.
 pub fn sigprocmask(flags: u32, noalias set: ?*const sigset_t, noalias oldset: ?*sigset_t) void {
     switch (errno(system.sigprocmask(flags, set, oldset))) {
         .SUCCESS => return,
         .FAULT => unreachable,
-        .INVAL => unreachable, // main purpose: debug InvalidValue error
+        .INVAL => unreachable,
         else => unreachable,
     }
 }
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -250,6 +250,15 @@ pub var argv: [][*:0]u8 = if (builtin.link_libc) undefined else switch (builtin.
     else => undefined,
 };
 
+/// Atomic to guard correct program teardown in abort()
+var abort_entered = impl: {
+    if (builtin.single_threaded) {
+        break :impl {};
+    } else {
+        break :impl std.atomic.Atomic(bool).init(false);
+    }
+};
+
 /// To obtain errno, call this function with the return value of the
 /// system function call. For some systems this will obtain the value directly
 /// from the return code; for others it will use a thread-local errno variable.
@@ -442,6 +451,7 @@ fn getRandomBytesDevURandom(buf: []u8) !void {
 /// Causes abnormal process termination.
 /// If linking against libc, this calls the abort() libc function. Otherwise
 /// it raises SIGABRT followed by SIGKILL and finally lo
+/// assume: Current signal handler for SIGABRT does **not call abort**.
 pub fn abort() noreturn {
     @setCold(true);
     // MSVCRT abort() sometimes opens a popup window which is undesirable, so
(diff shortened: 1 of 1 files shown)


lib/std/os/linux.zig
```diff
@@ -3107,6 +3107,21 @@ pub const Sigaction = extern struct {
 
 pub const empty_sigset = [_]u32{0} ** @typeInfo(sigset_t).Array.len;
 
+pub const OffsetMask = struct {
+    offset: u8,
+    mask: u8,
+};
+
+// workaround, because @as(sigset_t, 1) << (SIG.XYZ - 1); does not work
```
matu3ba, line 3115: I really hate this workaround. Better suggestions?

lib/std/os.zig
```diff
@@ -455,7 +455,44 @@ pub fn abort() noreturn {
...
-        // TODO the rest of the implementation of abort() from musl libc here
+        // If there was a SIGABRT handler installed and it returned, or if
+        // SIGABRT was blocked or ignored, take an Async-Signal-safe lock to
+        // prevent sigaction from installing a new SIGABRT handler, uninstall
+        // any handler that may be present, and re-raise the signal to generate
+        // the default action of abnormal termination.
+        // TODO handle failures on setting sigprocmask
+        _ = linux.sigprocmask(SIG.BLOCK, &linux.all_mask, null);
+
+        // TODO decide where (conditionally) abort_lock mutex should go
```
matu3ba, line 466: Not sure, where global variables/state of zig binaries should go.
Are there meta functions to access them directly or via offset?

matu3ba:
Context: Other OS are missing Sigaction pieces.
Status: Still missing polishing and will look into Kernel failure modes of signaling next.

Open questions:
1. How do I get the offset to the TLS from each thread? Is the TLS layout specified anywhere, ie via linker etc?
2. How would you test correctness of the abort() function?
3. What metric would you use to measure perf of the abort() function?
4. Why does glibc do a recursive locking and unlocking dance and not use a mutex?

lib/std/os/linux.zig
daurnimator: Something like `EnumSet`; though we should add something that assures it's dense.

lib/std/os.zig
```diff
@@ -455,7 +455,44 @@ pub fn abort() noreturn {
...
+        const sigact = Sigaction{
+            .handler = .{ .sigaction = SIG.DFL },
+            .mask = undefined,
+            .flags = undefined,
+            .restorer = undefined,
+        };
+        _ = linux.sigaction(SIG.ABRT, &sigact, null);
+
+        // TODO musl uses __get_tp() and optional offset into TLS block (pthread_impl.h)
+        _ = linux.tkill(linux.gettid(), SIG.ABRT);
```
LemonBoy, line 483: `Thread,getCurrentId()` should do the trick here.

lib/std/os/linux.zig
matu3ba: `x/os/socket_posix.zig` uses 
```zig
        pub fn init(domain: u32, socket_type: u32, protocol: u32, flags: std.enums.EnumFieldStruct(Socket.InitFlags, bool, false)) !Socket {
            var raw_flags: u32 = socket_type;
            const set = std.EnumSet(Socket.InitFlags).init(flags);
            if (set.contains(.close_on_exec)) raw_flags |= os.SOCK.CLOEXEC;
            if (set.contains(.nonblocking)) raw_flags |= os.SOCK.NONBLOCK;
            return Socket{ .fd = try os.socket(domain, raw_flags, protocol) };
        }
```
As I understand it, this only checks existence of an enum field. However, I would like ideally something that
1. checks comptime offset-safety and 2. does the shifting.

I could also add something to mem.zig, but I would like approval first to ensure there is no other usable solution.

lib/std/os.zig
matu3ba: Unfortunately do the types mismatch, because this is how getCurrentId is defined:
```rust
/// Represents a unique ID per thread.
pub const Id = u64;
/// Returns the platform ID of the callers thread.
/// Attempts to use thread locals and avoid syscalls when possible.
pub fn getCurrentId() Id {
    return Impl.getCurrentId();
}
```
and using it for the Kernel requires `intCast` back to the original type, which got lost in the `@bitCast(u32, os_gettidcall())`.

matu3ba: @kprotty ideas? feel free to take over this PR.

matu3ba:
Another potentially nice to have would be to provide
```zig
/// Set the thread signal mask
+/// Invalid masks are checked in Debug and ReleaseFast
+pub fn sigprocmask(flags: u32, noalias set: ?*const sigset_t, noalias oldset: ?*sigset_t) void {
+    switch (errno(system.sigprocmask(flags, set, oldset))) {
+        .SUCCESS => return,
+        .FAULT => unreachable,
+        .INVAL => unreachable, // main purpose: debug InvalidValue error
+        else => unreachable,
+    }
+}
+
```
I checked all OSes and except Linux (and Windows) they only have EINVAL as errno.

lib/std/os.zig
```diff
@@ -455,7 +455,43 @@ pub fn abort() noreturn {
     if (!builtin.link_libc and builtin.os.tag == .linux) {
         raise(SIG.ABRT) catch {};
 
-        // TODO the rest of the implementation of abort() from musl libc here
+        // If there was a SIGABRT handler installed and it returned, or if
+        // SIGABRT was blocked or ignored, take an Async-Signal-safe lock to
+        // prevent sigaction from installing a new SIGABRT handler, uninstall
+        // any handler that may be present, and re-raise the signal to generate
+        // the default action of abnormal termination.
```
kprotty, line 462: - What would be difference between this vs. raising `SIG.KILL` below?
- Couldn't one use sigaction without acquiring said global lock?

matu3ba: "SIGKILL signal cannot be caught or ignored and the receiving process cannot perform any clean-up upon receiving this signal. SIGABRT signal can be caught, but it cannot be blocked."
https://serverfault.com/a/873697

The problem is that sigaction is not under our control, so the user may do some above mentioned weird stuff by accident or intentional with multiple threads.

Thus:
1. run user-installed signal handler (not sure, if we support this yet)
2. block signals
3. take a lock to ensure the default abort handler is run by exactly 1 thread
4. run abort again, now with uninstalled signal handler (DFL for default)
5. failsafe mode: only enable receiving ABRT (This part I neither understand, because taking the lock twice is UB.)
6. try atomic hardcrash with nullptr dereference (which caused some weird behavior, so I removed it)
7. still running? ask Kernel to KILL it.
8. still running? beg it it to exit(127).

kprotty: @matu3ba 

1. `raise(ABRT)` would run the user-installed signal handler i'd imagine.
2. user-installed signal handler for ABRT returns from raise(), so we block all signals `sigprocmask(BLOCK)`
3. instead of taking a lock, we should grab an atomic bool and have other threads sleep forever (locking may assume it will eventually be unlocked by the caller thread).
4. `sigaction(DFL)` to set the signal handlers to their default actions, unregistering the user-installing ones.
5. instead of `tkill(current, ABRT)`, should we instead `raise(ABRT)` so that when signals are unblocked, abrt occurs for the entire process?
6. unblock signal handlers (`sigaction(UNBLOCK)`) which runs the `raise/tkill(ABRT)` queued from before and triggers its default handler
7. still running? raise(KILL). This should be uncatchable from what you've explained
8. still running? kernel should've killed the process from raise(KILL). This IMO should be `unreachable` rather than `exit()`.

matu3ba: @kprotty

1. yes
2. yes
3. The atomic must be global and conditional, if the code is compiled for single
or multithreaded code. something along (for now SeqCst for simplicity)
```zig
        var abort_entered = std.atomic.Atomic(bool).init(false);
        // must never fail, so strong CAS
        while(abort_entered.compareAndSwap(false, true, .SeqCst, .SeqCst)) {}
```
Unfortunately aweful implementations can call abort() within the abort-handler,
which is why the recursive (un)locking dance in glibc exists.
4. yes
5. raise(ABRT) has more overhead due to (BLOCK, gettid(), tkill, UNBLOCK) and we can use tkill directly instead. (we have already blocked all signals)
6. I'll change the comments. In theory could the signal queue run full, if not implemented
   as ringbuffer, but I think this is unfeasible to assume.
7. yes
8. If the process is not running Kernel code, then it should not be able to continue.
   https://unix.stackexchange.com/a/485657

Context
Signal handlers are per process and masks are per thread.
Accordingly, signal delivery (raise and tkill) is per thread.
https://unix.stackexchange.com/a/597387
https://stackoverflow.com/questions/11679568/signal-handling-with-multiple-threads-in-linux

Linux man page on abort()

> The abort() function first unblocks the SIGABRT signal, and then raises that signal for the
calling process (as though raise(3) was called).

Threads can only block their own signal masks, so this is very poorly phrased.
The man page for raise() writes
"In a multithreaded program it is equivalent to pthread_kill(pthread_self(), sig);"
and for posix
"The effect of the raise() function shall be equivalent to calling: pthread_kill(pthread_self(), sig);".
This reads like musl is wrong on not masking in SIG.ABRT to run the default handler
and instead relies on the application code to "do the correct thing".
However, one could also argue that POSIX is wrong, because the application
should not rely on flawed implementation logic.

> This results in the abnormal termination of the process unless the
SIGABRT signal is caught and the signal handler does not return (see longjmp(3)).

Currently, Zig kinda rules this out with `noreturn` as abort function signature.

> If the SIGABRT signal is ignored, or caught by a handler that returns, the  abort()  function
will still terminate the process. It does this by restoring the default disposition for SIGA‐
BRT and then raising the signal for a second time.

This should be now implemented.

matu3ba: I did add your suggestions. 
Please reopen, if you find things discussed here you think are missing.

lib/std/os.zig
matu3ba: I'll mark this as resolved for now. I did add a TODO for solving this, which can be optimized later and is no regression to status quo.

lib/std/os.zig
matu3ba: added TODO for the atomic to use.

matu3ba:
Aside of the dirty workaround this feels ready.

Please let me know how to proceed on the missing signaling errors.
The C libraries (musl and glibc) are ignoring them even though kill should be used on tkill failure.

lib/std/os/linux.zig
matu3ba: @daurnimator
```zig
        var sigabrtmask = std.bit_set.ArrayBitSet(u32, @typeInfo(sigset_t).Array.len);
        comptime sigabrtmask.initEmpty();
        comptime sigabrtmask.set(SIG.ABRT);
        sigprocmask(SIG.UNBLOCK, &sigabrtmask, null);
```
does the trick.

matu3ba:
Related is a question on Zig policy on signaling https://discord.com/channels/605571803288698900/906306963942543370/975880042422890626
tldr;
1. What is the general policy of Zig is on signal handling due to it being widely non-portable (see the mess on real-time signals man 7 signalor man signal-safety)?
2. Handling signal safety has an unnecessary overhead,  if the user blocks or does not use the signals anyway. But not handling it exposes the user to the signal mess, which can break the program.
3. Should process and thread teardown methods be signal safe? Or more stuff?

lib/std/os.zig
```diff
@@ -453,11 +454,54 @@ pub fn abort() noreturn {
...
-        // TODO the rest of the implementation of abort() from musl libc here
+        // disable all signal handlers
+        sigprocmask(SIG.BLOCK, &linux.all_mask, null);
+
+        // TODO thread id should be taken from TLS instead of syscall
+        // TODO add thread group to TLS and use tgkill() instead of tkill()
+        // TODO conditionally add global atomic to start code and add CAS below:
+        //var abort_entered = std.atomic.Atomic(bool).init(false);
+        // ensure teardown by one thread
+        //while(abort_entered.compareAndSwap(false, true, .SeqCst, .SeqCst)) {}
```
andrewrk, line 472: All of the rest of the implementation needs to be ported, or none of it. This lock is needed for the code below.

lib/std/os.zig
```diff
@@ -453,11 +454,54 @@ pub fn abort() noreturn {
...
+        const tid = linux.gettid();
+        _ = linux.tkill(tid, SIG.ABRT);
+
+        // SIG.ABRT signal will run default handler
+        comptime var sigabrtmask_bitset = std.bit_set.ArrayBitSet(u32, @typeInfo(sigset_t).Array.len).initEmpty();
+        var sigabrtmask: system.sigset_t = undefined;
+        comptime sigabrtmask_bitset.set(SIG.ABRT);
+        inline for (sigabrtmask_bitset.masks) |mask, i|
+            sigabrtmask[i] = mask;
+        sigprocmask(SIG.UNBLOCK, &sigabrtmask, null);
```
andrewrk, line 493: Avoid `comptime var` and `inline for`. Use constants for these things. Runtime bloat is especially tragic in this code which will almost never see the light of day.

lib/std/os/linux.zig
```diff
@@ -3051,7 +3051,7 @@ pub const NSIG = if (is_mips) 128 else 65;
 
 pub const sigset_t = [1024 / 32]u32;
 
-pub const all_mask: sigset_t = [_]u32{0xffffffff} ** sigset_t.len;
+pub const all_mask: sigset_t = [_]u32{0xffffffff} ** @typeInfo(sigset_t).Array.len;
```
andrewrk, line 3054: Why?

lib/std/os.zig
```diff
@@ -441,6 +441,7 @@ fn getRandomBytesDevURandom(buf: []u8) !void {
 /// Causes abnormal process termination.
 /// If linking against libc, this calls the abort() libc function. Otherwise
 /// it raises SIGABRT followed by SIGKILL and finally lo
+/// assume: Current signal handler for SIGABRT does **not call abort**.
```
andrewrk, line 444: Don't assume this.

andrewrk (changes requested):

lib/std/os.zig
matu3ba: This is what musl assumes. I checked llvm-libc and found a hint to why glibc does the recursive locking+unlocking dance.

There are 2 ways to "fix" this:
1. do the Linux thing and prepare user to always assume the user-installed handler is run to force the user to handle the possible recursion
2. do the glibc recursive locking+unlocking dance, with a mutex per thread which has the same drawbacks of potential failure in abort() and thus broken process teardown:
```txt
      /* This stage is special: we must allow repeated calls of
	 `abort' when a user defined handler for SIGABRT is installed.
	 This is risky since the `raise' implementation might also
	 fail but I don't see another possibility.  */
```

For a more elaborative discussion, see https://stackoverflow.com/q/187761/9306292
Also, as far as I am aware, libstd does not support recursive mutexes (part from Mutex.zig):
```txt
/// It is undefined behavior if the mutex is already held by the caller's thread.
```

matu3ba:
I tested this and get the following result on Linux (Arch derivate):
```txt
const std = @import("std");
fn call2() void {
    std.os.abort();
}

fn call1() void {
    call2();
}

test "broken stuff" {
    std.debug.print("calling abort:\n", .{});
    call1();
}
```
```txt
Test [1/1] test "broken stuff"... calling abort:
zig-cache/o/29cf4c73c518709af7fb19d0117fb1b0/test /home/misterspoon/dev/git/zig/zig/port_abort/build/zig
```

Should I add a standalone test with a child process that does some stuff in a custom abort() handler or
is this fine without the test?

lib/std/os.zig
matu3ba: Should be fixed now. Calling `abort()` works at least without segfault/Kernel page fault, so the only active flag looks correct.

lib/std/os.zig
matu3ba: I did port everything needed and used for now the assumption.

PR #11570, opened 2022-05-02
io: add copy() and copyUsingBuffer()

vrischmann:
This PR adds the functions `std.io.copy` and `std.io.copyUsingBuffer` which copy data from a source reader into a destination writer until end of stream or an error occurs.

This is useful for example for copying a process stdout to a file or reading a file directly into a hasher type.

This functionality is similar to Go's [io.Copy](https://pkg.go.dev/io#Copy) or Rust's [std::io::copy](https://doc.rust-lang.org/std/io/fn.copy.html).

Right now the implementation is completely generic whereas Go and Rust can use `splice` and `sendfile` under the hood in some cases but this can be improved later on, for example I quickly hacked something using `std.fs.copy_file` when both dest and source are files:
```zig
    if (@TypeOf(dest.context) == std.fs.File and @TypeOf(source.context) == std.fs.File) {
        const copied = try std.fs.copy_file(source.context.handle, dest.context.handle);
        return @intCast(usize, copied);
    }
```

This is also why the error set is not constrained for `copy`, because the reader/writer errors don't cover these possible specializations.

### DIFF
--- a/lib/std/io.zig
+++ b/lib/std/io.zig
@@ -149,6 +149,9 @@ pub const BufferedAtomicFile = @import("io/buffered_atomic_file.zig").BufferedAt
 
 pub const StreamSource = @import("io/stream_source.zig").StreamSource;
 
+pub const copy = @import("io/copy.zig").copy;
+pub const copyUsingBuffer = @import("io/copy.zig").copyUsingBuffer;
+
 /// A Writer that doesn't write to anything.
 pub const null_writer = @as(NullWriter, .{ .context = {} });
 
@@ -177,5 +180,6 @@ test {
     _ = @import("io/peek_stream.zig");
     _ = @import("io/seekable_stream.zig");
     _ = @import("io/stream_source.zig");
+    _ = @import("io/copy.zig");
     _ = @import("io/test.zig");
 }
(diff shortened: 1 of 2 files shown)


LemonBoy:
FYI, https://github.com/ziglang/zig/blob/7355a201336c8e3892427e5932fe5cdd46cf96df/lib/std/fifo.zig#L371

vrischmann:
Thanks, I didn't realize something similar existed already with this name.

Maybe it's because of my Go bias but I think it makes more sense as `std.io.copy()`.

daurnimator:
> Maybe it's because of my Go bias but I think it makes more sense as `std.io.copy()`.

It's a method on the `LinearFifo` object which lets you pick size + allocation. It also lets you e.g. pre-fill the buffer and hence prepend data to the output.

andrewrk:
Sorry, I don't really like the new APIs. Too abstract, too vague, too generic.

For what it's worth I don't like LinearFifo.pump() either and would probably like to delete that one too.

malcolmstill:
For what it's worth @vrischmann I was looking for exactly this function recently (indeed, for use with zig-sqlite's blob writer!) and was surprised not to find it.

The `copy` function seems like a natural consequence of having the generic `Reader` and  generic `Writer` interfaces, i.e. wanting to plug any reader into any writer.

PR #11575, opened 2022-05-03
musl: update to 1.2.3

ifreund:
This was a bit trickier than it should be due to symbol conflicts with
zig's compiler-rt implementation. We attempt to use weak linkage in
our compiler-rt, but this does not seem to be working in all cases. I
manually disabled export of the problematic compiler-rt math functions
in order to cross compile musl's libc.so for all targets as input to
`tools/gen_stubs.zig`.

Other than that, this update went fairly smoothly. Quite a few
additional symbols were added to the blacklist in `tools/gen_stubs.zig`
due to recent reorganization of zig's compiler-rt.


tools/gen_stubs.zig
```diff
@@ -658,48 +687,80 @@ const blacklisted_symbols = [_][]const u8{
...
+    "__floatundixf",
     "__floatunsidf",
     "__floatunsikf",
     "__floatunsisf",
     "__floatunsitf",
+    "__floatunsixf",
     "__floatuntidf",
     "__floatuntisf",
     "__floatuntitf",
+    "__floatuntixf",
```
tiehuis, line 723: A new compiler-rt export `__floatuntikf` was merged recently that you haven't yet got in your master branch: https://github.com/ziglang/zig/commit/2947a2faab2e287926b20d5d8984ad69bba5bd2c

Worth checking if this will need to be blacklisted, too.

ifreund: Rebased on master and added it to the blacklist, thanks for the catch!

andrewrk:
Nice work!

Hmm do we have some kind of accidentally nondeterministic sorting of the output for libc.S? If it's low effort it might be worth a small patch to gen_stubs.zig to make the libc.S diff smaller (if not for this upgrade then for the next one).

ifreund:
Yes, the current gen stubs implementation preserves the order symbols appear in the object file. The previous gen stubs implementation used to sort the output by symbol name for exactly this reason, but it seems that feature was dropped among the recent improvements.

It would be nice if `std.ArrayHashMap()` had a sort function...

squeek502:
> It would be nice if `std.ArrayHashMap()` had a sort function...

It does? https://github.com/ziglang/zig/pull/11109

andrewrk (approved):

andrewrk:
Check out the implementation of sort in MultiArrayList - I think it turned out to be quite elegant:

https://github.com/ziglang/zig/blob/3ed9ef3e6bed3fef6d6cad07920d08b28e20ec3e/lib/std/multi_array_list.zig#L426-L450

ifreund:
> Check out the implementation of sort in MultiArrayList - I think it turned out to be quite elegant:

Very cool! Last time I reached for `std.ArrayHashMap().sort()` it wasn't there but it appears that's now changed :)

`tools/gen_stubs.zig` now sorts the symbols by section and then by name before writing the output which should make things 100% deterministic.

andrewrk:
Great thanks!

PR #11577, opened 2022-05-04
[ld.lld] do not add --gc-sections by default

### LINKED ISSUE #11398: Linker error using zig cc with Go tests using race detection
### Zig Version

0.10.0-dev.1734+7b7f45dc2

### Steps to Reproduce

`go.mod`:
```
module test

go 1.18
```

`foo.go`:
```go
package test

func foo() {}
```

`foo_test.go`:
```go
package test

import (
	"testing"
)

func TestFoo(t *testing.T) {
	foo()
}
```

Then use zig cc as the C toolchain for `go test` with race detection:
```shell
CGO_ENABLED=1 CC="zig cc" go test -race test
```

This is on x86_64 Linux.

### Expected Behavior

This should work, as it does without `-race`.

### Actual Behavior

We get errors from the Go linker:
```
# test.test
runtime/race(.text): relocation target getuid not defined
runtime/race(.text): relocation target pthread_self not defined
runtime/race(.text): relocation target sleep not defined
runtime/race(.text): relocation target usleep not defined
runtime/race(.text): relocation target abort not defined
runtime/race(.text): relocation target isatty not defined
runtime/race(.text): relocation target pthread_attr_getstack not defined
runtime/race(.text): relocation target sigaction not defined
runtime/race(.text): relocation target getrusage not defined
runtime/race(.text): relocation target syslog not defined
runtime/race(.text): relocation target confstr not defined
runtime/race(.text): relocation target getrlimit not defined
runtime/race(.text): relocation target pipe not defined
runtime/race(.text): relocation target sched_getaffinity not defined
runtime/race(.text): relocation target __sched_cpucount not defined
runtime/race(.text): relocation target pthread_attr_init not defined
runtime/race(.text): relocation target pthread_getattr_np not defined
runtime/race(.text): relocation target pthread_attr_destroy not defined
runtime/race(.text): relocation target exit not defined
runtime/race(.text): relocation target sysconf not defined
runtime/race(.text): relocation target setrlimit not defined
/home/kmicklas/go/pkg/tool/linux_amd64/link: too many errors
FAIL	test [build failed]
FAIL
```

One interesting thing to note is that adding `import "C"` (i.e. triggering the use of cgo) to `foo.go` makes this example work. However, this only seems to work *most* of the time in larger examples.

motiejus:
When building object files, `zig cc` will instruct lld to remove unused
sections via `--gc-sections`. This is problematic for cgo builds that don't
explicitly use C code. Briefly, go builds cgo executables as follows*:

1. build `_cgo_.o`, which links (on linux_amd64 systems) to
   `/usr/local/go/src/runtime/race/race_linux_amd64.syso`.
2. That `.syso` contains references to symbols from libc.

If the user program uses at least one libc symbol, it will link
correctly. However, if Go is building a cgo executable, but without C
code, the sections from `.syso` file will be garbage-collected, leaving
a `_cgo_.o` without any references to libc, causing the final linking
step to not link libc.

Until now, this could be worked around by `-linkmode external` flag to
`go build`. This causes Go to link the final executable using the
external linker (which implicitly links libc). However, that flag brings
in a whole different world of worms.

I assume the `gc_sections` is an optimization; I tried to re-add
`--gc-sections` to the final executable, but that didn't go well. I know
removing such an optimization may be contentious, so let's start the
discussion here. Quoting @andrewrk in [1] (it was about `--as-needed`,
but the point remains the same):

> The C source code for the temporary executable needs to have dummy
> calls to getuid, pthread_self, sleep, and every other libc function
> that cgo/race code wants to call.

I agree this is how it *should* work. However, while we could fix it for
go, I don't know how many other systems rely on that, and we'll never
know we've fixed the last one. The point is, GCC/Clang does not optimize
sections by default, and downstream tools rely on that. If we want to
consider `zig cc` a drop-in clang replacement (except for
`-fsanitize=undefined`, which I tend to agree with), then it should not
be optimizing the intermediate object files. Or at least have a very
prominent fine-print that this is happening, with ways to work around
it.

Fixes #11398
Fixes https://github.com/golang/go/issues/44695
Fixes https://github.com/golang/go/issues/52690

[*]: Empirically observed with `CGO_ENABLED=1 go test -race -x -v`

[1]: https://github.com/golang/go/issues/52690#issuecomment-1116919960

### DIFF
--- a/src/link/Elf.zig
+++ b/src/link/Elf.zig
@@ -1264,7 +1264,7 @@ fn linkWithLLD(self: *Elf, comp: *Compilation, prog_node: *std.Progress.Node) !v
     const have_dynamic_linker = self.base.options.link_libc and
         self.base.options.link_mode == .Dynamic and is_exe_or_dyn_lib;
     const target = self.base.options.target;
-    const gc_sections = self.base.options.gc_sections orelse !is_obj;
+    const gc_sections = self.base.options.gc_sections orelse false;
     const stack_size = self.base.options.stack_size_override orelse 16777216;
     const allow_shlib_undefined = self.base.options.allow_shlib_undefined orelse !self.base.options.is_native_os;
     const compiler_rt_path: ?[]const u8 = blk: {


motiejus:
CI failure:

```
ld.lld: error: undefined symbol: main
```

So treat this as a conversation starter w.r.t. `--gc-sections`, not a real pull request. :)

motiejus:
To remove confusion: the first diff removed `--gc-sections` when the output was `.Exe`, but that broke a bunch of other stuff. So I am being conservative here (or adventurous, depends how you look at it) and am not adding the flag by default altogether.

motiejus:
For the record, https://builds.hut.lavatech.top/~andrewrk/job/23490 seems to be failing on master too.

motiejus:
For the record, I added `-Wl,--no-gc-sections` to bazel-zig-cc cflags[1]. Now bazel-zig-cc can build more cgo stuff!

[1]: https://lists.sr.ht/~motiejus/bazel-zig-cc/%3C20220505122425.ipd2p6kykd7iph7b%40mtpad.i.jakstys.lt%3E

andrewrk:
> I agree this is how it _should_ work. However, while we could fix it for
> go, I don't know how many other systems rely on that, and we'll never
> know we've fixed the last one. The point is, GCC/Clang does not optimize
> sections by default, and downstream tools rely on that. If we want to
> consider `zig cc` a drop-in clang replacement (except for
> `-fsanitize=undefined`, which I tend to agree with), then it should not
> be optimizing the intermediate object files. Or at least have a very
> prominent fine-print that this is happening, with ways to work around
> it.

The Zig project is about rebuilding systems programming from the ground up, re-examining assumptions, and making different decisions where they are warranted. Fully imitating GCC including its limitations is not a goal we have. In fact that already exists, it's Clang! `zig cc` is interesting precisely because, while it is a drop-in replacement, it also does some things differently.

Ultimately, the downstream tools need improvements in this case, not Zig.

PR #11578, opened 2022-05-04
stage2: `@intToEnum` runtime safety

wsengir:
This implements runtime safety checks for `@intToEnum`. `intCast` is used to both cast the operand into the enum's backing integer type as well as generate shrinkage safety checks. Then enums without explicit values are range checked while enums with explicit values have each value checked individually. The current implementation is rather simplistic and chains OR's together. In the future, it would be better to instead generate a switch with each value as a case.

I've been preoccupied with non-Zig stuff, so I'm a little out of the loop on how we want to test these. I'd appreciate some pointers. Thanks.

Related to #89.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -24637,7 +24637,7 @@ fn enumHasInt(
     int: Value,
 ) CompileError!bool {
     switch (ty.tag()) {
-        .enum_nonexhaustive => return sema.intFitsInType(block, src, int, ty),
+        .enum_nonexhaustive => return sema.intFitsInType(block, src, int, ty.castTag(.enum_nonexhaustive).?.data.tag_ty),
         .enum_full => {
             const enum_full = ty.castTag(.enum_full).?.data;
             const tag_ty = enum_full.tag_ty;
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -6009,7 +6009,7 @@ fn zirEnumToInt(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!A
             break :blk try sema.unionToTag(block, tag_ty, operand, operand_src);
         },
         else => {
-            return sema.fail(block, operand_src, "expected enum or tagged union, found {}", .{
+            return sema.fail(block, operand_src, "expected enum or tagged union, found '{}'", .{
                 operand_ty.fmt(sema.mod),
             });
         },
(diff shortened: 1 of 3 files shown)


src/Sema.zig
```diff
@@ -6076,7 +6080,31 @@ fn zirIntToEnum(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!A
...
+        switch (dest_ty.tag()) {
+            .enum_nonexhaustive => {},
+            .enum_full, .enum_numbered, .enum_simple => {
+                const field_count = dest_ty.enumFieldCount();
+                if (dest_ty.tag() == .enum_simple or dest_ty.enumValues().count() == 0) {
+                    const field_count_inst = try sema.addIntUnsigned(dest_int_ty, field_count);
+                    const is_in_range = try block.addBinOp(.cmp_lt, casted_operand, field_count_inst);
+                    try sema.addSafetyCheck(block, is_in_range, .invalid_enum_value);
+                } else {
+                    // TODO: Output a switch instead of chained OR's.
```
andrewrk, line 6095: Hmmmm. Let's do this the right way from the start, otherwise this will harm compilation performance due to creating AIR bloat.

Suggestion: introduce a new AIR instruction called `check_enum_int` which returns true if and only if the integer is a valid tag of a given enum type. This way the backends can lower it as a switch directly, instead of having to parse the various AIR tags for a switch emitted from Sema. This would also allow the possibility for a backend to lower it to a function call in the case of enums with many tags.

stage1 already benefits from a similar arrangement because it implements runtime safety in the LLVM IR lowering pass, which is less flexible in some ways (panic function gets outputted even if it is never called) but comes with advantages, as illustrated here.

Alternative idea which you are welcome to explore: create a function in Sema called `__zig_intToEnum_E` where `E` is the fully qualified type name of the enum, which does the switch, calls panic if applicable with a custom panic message ("enum '{s}' has no tag with value '{d}'"), exactly matching the corresponding compile error. This alternative would not require introducing a new AIR tag, would not require any changes to the backends, and would pioneer the way forward to more detailed runtime safety messages.

andrewrk (changes requested):
Thanks for your patience on getting a review.

As for testing, the runtime safety tests can be executed like this:

```
stage1/bin/zig build test-cases -Dtest-filter=safety -Denable-llvm
```

When you want to mark a test as passing, do this:

```diff
--- a/test/cases/safety/integer addition overflow.zig   
+++ b/test/cases/safety/integer addition overflow.zig   
@@ -19,5 +19,5 @@ fn add(a: u16, b: u16) u16 {
 }
 
 // run
-// backend=stage1
+// backend=stage1,llvm
 // target=native
```

There are a few that can already be marked this way, such as this example.

PR #11581, opened 2022-05-04
Fix `std.meta.TrailerFlags`

erikarvstedt:
- Include `TrailerFlags` in std tests.
  Also fix a bug that was undetected because of the skipped tests.

- Previously, init() took a struct parameter, which gave the misleading impression that `TrailerFlags` was initialized with the struct's values. In fact, those values were unused.

  Before:
  ```zig
  var flags = Flags.init(.{
      .a = 1234,
      .b = "actually unused str",
  });
  ```
  After: (Edit: Now using @InKryption's suggestion)
  ```zig
  var flags = Flags.init(.{
      .a = true,
      .b = true,
  });
  ```

cc @Vexu, @andrewrk, @tadeokondrak

### DIFF
--- a/lib/std/meta.zig
+++ b/lib/std/meta.zig
@@ -10,6 +10,10 @@ pub const TrailerFlags = @import("meta/trailer_flags.zig").TrailerFlags;
 
 const Type = std.builtin.Type;
 
+test "std.meta.TrailerFlags" {
+    _ = TrailerFlags;
+}
+
 pub fn tagName(v: anytype) []const u8 {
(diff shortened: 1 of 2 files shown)


InKryption:
Wouldn't a better idea here be to make all the fields of `InitStruct` booleans, than to use a slice that could potentially have redundant instances of a field tag?

erikarvstedt:
I've changed init to use a boolean struct. (Thanks for suggesting `EnumFieldStruct`!)

PR #11584, opened 2022-05-05
Give the users the chose of how to indent code.

10maurycy10:
I know this could reduce the consistency of zig code (allowing mixing tabs and spaces in the same project), but the best way to solve this is with a autoformater, not by having the compile throw an error when you try to use a tab.


andrewrk:
we're not accepting language proposals at this time, and also this topic has been discussed to death.

PR #11586, opened 2022-05-05
Fixes comptime 'error: cannot assign to constant' error in siphash.

MasterQ32:
This is a workaround for some stage1 bug in `@call` that happens only in certain cases in comptime evaluation. [LoLa](https://github.com/MasterQ32/LoLa) is broken on `0.10.0-dev.2054+3ed9ef3e6` due to this evaluation bug.

This is likely a regression in stage1, so this is only a workaround. The same "fix" is already applied somewhat lower in siphash.


andrewrk:
Sorry for the trouble, and thank you for the fix :+1:

PR #11589, opened 2022-05-05
adding comments

4zv4l:
adding comments to the indexOf section


andrewrk:
These comments are redundant; they do not contain any extra information that is not already in the function signature.

PR #11590, opened 2022-05-05
comment about digitToChar

4zv4l:


andrewrk (changes requested):
Please use complete sentences in doc comments, and avoid abbreviations such as "char" (use "character" instead).

Issue #11367, opened 2022-04-01
false positive circular dependency when a comptime-called function has function as a return type that uses generic parameters from the outer function


andrewrk:
Zig Version: `0.10.0-dev.1694+ecd756834`

```zig
test {
    var map: HashMapUnmanaged(AutoContext(*Decl)) = .{};
    _ = map;
}

const Decl = struct {
    namespace: *Namespace,
};

const Namespace = struct {
    decls: HashMapUnmanaged(AutoContext(*Decl)),
};

fn getFn(comptime K: type) (fn (K) u64) {
    return struct {
        fn hash(key: K) u64 {
            _ = key;
            return 1234;
        }
    }.hash;
}

fn AutoContext(comptime K: type) type {
    return struct {
        const hash = getFn(K);
    };
}

fn HashMapUnmanaged(comptime Context: type) type {
    comptime verifyContext(Context);
    return struct {};
}

fn verifyContext(comptime Context: type) void {
    const hash = Context.hash;
    _ = hash;
}
```

Expected to pass. Instead:

```
$ ./stage2/bin/zig test test3.zig 
test3.zig:28:9: error: 'hash' depends on itself
        const hash = getFn(K);
        ^
```

This happens because:
 * `verifyContext` tries to look at the `hash` function, invoking comptime `getFn`.
 * while evaluating the return type of `getFn`, zig tries to check if `K` is a comptime-only type.
 * `K` is `*Decl` so to check if it is a comptime-only type, it looks at Decl fields types, and then recursively looks at Namespace field types.
 * `HashMapUnmanaged(AutoContext(*Decl))` is comptime-evaluated
 * `verifyContext` is again called at comptime for the same parameters (the previous call is still ongoing)
 * `hash` is still wip but we are trying to look at it again.

A workaround is placing the comptime call to `verifyContext` inside the `struct` rather than before it:

```zig
fn HashMapUnmanaged(comptime Context: type) type {
    return struct {
        comptime {
            verifyContext(Context);
        }
    };
}
```

PR #11592, opened 2022-05-05
Sema: solve a false positive "depends on itself"

andrewrk:
This improves the ABI alignment resolution code.

This commit fully enables the MachO linker code in stage3. Note,
however, that there are still miscompilations in stage3.


andrewrk:
This actually does *not* solve  #11367 - I just tested it. I'll leave the CI running and force push a reword of the commit message before merging.

PR #11598, opened 2022-05-06
Add functions from `dirent.h` to std.c

aiotter:
This PR introduces functions from `dirent.h`, which exists on Unix-like systems including macOS and FreeBSD.
Along with the new functions, I simplified `std.c` by removing redundant code from `std/c/darwin.zig`.


aiotter:
I noticed that `std/c/linux.zig` does not include `struct dirent`. I'll fix it on another PR along with fixes for incorrect implementation or `struct dirent` for other OSes.

andrewrk:
Thanks!

Issue #11585, opened 2022-05-05
No calling convention for msvc on x86_64 non-windows targets


N00byEdge:
The Win64 calling convention is the same as Cdecl on Windows x64, however using `.C` as a stand-in for the Win64 calling convention only works if your compiler targets Windows. This means you're not able to create a function with the Win64 calling convention if you're not targeting Windows.
It should be fairly straightforward to support LLVM's Win64 calling convention such that it can be used from Zig code on any target.

PR #11599, opened 2022-05-06
Add Win64 calling convention

N00byEdge:
Closes ziglang/zig#11585

### DIFF
--- a/src/stage1/codegen.cpp
+++ b/src/stage1/codegen.cpp
@@ -209,6 +209,9 @@ static ZigLLVM_CallingConv get_llvm_cc(CodeGen *g, CallingConvention cc) {
         case CallingConventionSysV:
             assert(g->zig_target->arch == ZigLLVM_x86_64);
             return ZigLLVM_X86_64_SysV;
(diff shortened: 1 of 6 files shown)


N00byEdge:
I added it before `CallingConventionPtxKernel` since it didn't have an assert on its enum value yet, and thought asserted ones should go first in that enum

N00byEdge:
This probably should have a test too, but I honestly have no idea how to even test this properly.

src/stage1/codegen.cpp
```diff
@@ -209,6 +209,9 @@ static ZigLLVM_CallingConv get_llvm_cc(CodeGen *g, CallingConvention cc) {
         case CallingConventionSysV:
             assert(g->zig_target->arch == ZigLLVM_x86_64);
             return ZigLLVM_X86_64_SysV;
+        case CallingConventionWin64:
+            assert(g->zig_target->arch == ZigLLVM_x86_64);
```
N00byEdge, line 213: Should this maybe allow aarch64 too? Not sure if it should.

PR #11602, opened 2022-05-07
std.c: Fix incorrenct dirent structure

aiotter:
`d_name` of `struct dirent` is a null-terminated array of `char`.

### DIFF
--- a/lib/std/c/darwin.zig
+++ b/lib/std/c/darwin.zig
@@ -942,7 +942,7 @@ pub const dirent = extern struct {
     d_reclen: u16,
     d_namlen: u16,
     d_type: u8,
-    d_name: [1023:0]u8,
+    d_name: [1024]u8,
 
     pub fn reclen(self: dirent) u16 {
         return self.d_reclen;
--- a/lib/std/c/darwin.zig
+++ b/lib/std/c/darwin.zig
@@ -937,12 +937,12 @@ pub const Sigaction = extern struct {
 };
 
 pub const dirent = extern struct {
-    d_ino: usize,
-    d_seekoff: usize,
+    d_ino: u64,
+    d_seekoff: u64,
     d_reclen: u16,
     d_namlen: u16,
     d_type: u8,
-    d_name: u8, // field address is address of first byte of name
+    d_name: [1023:0]u8,
 
     pub fn reclen(self: dirent) u16 {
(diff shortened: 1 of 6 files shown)


aiotter:
I found a problem on this PR. `d_name` on Darwin does not work. I know that d_name points to the first element of the `char[1024]` array. I don't know why this is a problem.

In the current implementation, whose `d_name` is `u8`, requires users to cast it to null-terminated pointer like following.
```zig
const std = @import("std");
const dirent = std.c.dirent;
const native_arch = @import("builtin").target.cpu.arch;
const DIR = @cImport(@cInclude("dirent.h")).DIR;

const private = struct {
    extern "c" fn readdir(dir: *DIR) ?*dirent;
    extern "c" fn @"readdir$INODE64"(dir: *DIR) ?*dirent;
};
extern "c" fn opendir(dir: [*:0]const u8) ?*DIR;
extern "c" fn closedir(*DIR) i8;
pub const readdir = if (native_arch == .aarch64) private.readdir else private.@"readdir$INODE64";

pub fn main() void {
    const dir = opendir("/").?;
    defer _ = closedir(dir);

    while (readdir(dir)) |entry| {
        const name = @ptrCast([*:0]const u8, &entry.d_name);  // <- HERE
        std.debug.print("{s}\n", .{name});
    }
}
```

The aim of this PR is to fix this along with other non-null-terminated `d_name`.
I'll convert this PR into draft. I'll think twice for correct implementation. Please let me know of the good way to fix this, or create another PR or commit and close this PR. Thanks

aiotter:
Okay, I read the docs carefully to find that an array with sentinel merely means it is guaranteed to end with sentinel and it does not care about having sentinel before end. `&entry.d_name[1024] == 0` seems always the case, but it's not documented, so the type of `d_name` is better to be `[1024]u8`.
It seems better to implement a method `fn name(self: dirent) []u8`, following already implemented `fn reclen(self: dirent) u16`.

andrewrk:
Can you please mark this as ready for review or close it (perhaps converting it to an issue)? I'd like to keep the PR queue as a place for patches that are ready to be merged.

aiotter:
@andrewrk Sorry, I marked this PR as ready for review.

andrewrk:
Thanks @aiotter!

PR #11316, opened 2022-03-26
stage2: vectorized overflow arithmetic, integer overflow safety, left-shift overflow safety

wsengir:
Sorry for yet another probably too big PR, but the changes were dependent on each other, and I didn't want to wait for the first to get merged before PRing the rest.

This breaks basically everything that's not LLVM.

PR #11608, opened 2022-05-07
stage2: introduce new register locking mechanism, and fix x64, arm and aarch64 native backends for compliance with integer overflow safety PR

kubkon:
The main aim of this PR is to unblock #11316 for the mainstream native backends: `x86_64`, `arm` and `aarch64`. Yep, you understood it right, this PR ensures #11316 when rebased on top of this PR will pass all of the tests for the aforementioned native backends with integer overflow safety on in Sema.

A major step in ensuring the native backends are able to pull this off was revamping the register locking (freezing) mechanics. Prior to this PR, locking registers (so that we don't accidentally reallocate and thus spill registers that are important to the instruction at hand) would follow the standard Zig's lock-defer mechanics, however, we would not track scopes, making this possible:

```zig
outer: {
    const reg = try self.register_manager.allocReg(null);
    self.register_manager.freezeReg(reg);
    defer self.register_manager.unfreezeReg(reg); // Frozen until the end of the current scope

    assert(self.register_manager.isRegFrozen(reg)); // OK

    inner: {
        self.register_manager.freezeReg(reg);
        defer self.register_manager.unfreezeReg(reg); // Unfreezing at the end of current nested scope.

        assert(self.register_manager.isRegFrozen(reg)); // OK
    }

    assert(self.register_manager.isRegFrozen(reg)); // Panic!!!
}
```

This would become problematic for more complex scopes and especially function calls and re-use where we would want to lock the input register from any accidental re-use, while removing the lock at the end of the scope. Removing the lock in the nested scope would remove all locks from the registers creating a bug whereby we could accidentally end up re-using the register in the parent scope where it would not be deemed suitable to do so yet.

In order to overcome this, together with @joachimschmidt557, we have devised a new mechanism that mimics Mutex locks with Rust flavour on top - when locking a register, return the lock wrapper struct signifying ownership of the lock, or `null` if the ownership has already been claimed in the parent scope. This way, we are given a powerful weapon which we can use to reason and verify that our mental model of where a register should be locked and where it actually was are the same. The mechanics are simple: calling `lockReg(reg)` will return a token wrapper struct which is required to free the lock when we are done. If the register was already locked in the parent scope, `lockReg(reg)` will instead return `null` signifying it was already locked and thus we should not try reusing or freeing it until the parent scope deems it suitable.

Typical invocation can now look like:

```zig
const reg = try self.register_manager.allocReg(null);
const lock: RegisterLock = self.register_manager.lockRegAssumeUnused(reg); // asserts the lock can be claimed right here and now
defer self.register_manager.unlockReg(lock);
```

If we are not sure the register was locked already in parent scope - a design that can appear in a shared utility function which for instance implements common binary ops code, we call `lockReg` instead

```zig
const maybe_lock: ?RegisterLock = self.register_manager.lockReg(reg);
defer if (maybe_lock) |lock| self.register_manager.unlockReg(lock);
```

This simple mechanism complements really well with the current design of the register manager's register tracking mechanism - it doesn't affect the memory footprint of tracking already locked registers, plus, with the clever use of `lockRegAssumeUnused`, it can be used to narrow down and validate primary callsites for calls to the register manager where we should hold a lock for the register until the scope ends. (This already proved useful finding bugs with improper use of register locking in every mainstream backend.)

cc @joachimschmidt557 @Luukdegram

### DIFF
--- a/src/arch/aarch64/CodeGen.zig
+++ b/src/arch/aarch64/CodeGen.zig
@@ -23,6 +23,7 @@ const log = std.log.scoped(.codegen);
 const build_options = @import("build_options");
 const RegisterManagerFn = @import("../../register_manager.zig").RegisterManager;
 const RegisterManager = RegisterManagerFn(Self, Register, &callee_preserved_regs);
+const RegisterLock = RegisterManager.RegisterLock;
 
 const GenerateSymbolError = @import("../../codegen.zig").GenerateSymbolError;
 const FnResult = @import("../../codegen.zig").FnResult;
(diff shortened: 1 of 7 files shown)


joachimschmidt557 (approved):
Looks good to me!

PR #11612, opened 2022-05-08
std.crypto: generate AES constants at compile time

0x2b3bfa0:
This pull request generates all the previously hardcoded S-boxes and lookup tables at compile time, showcasing one of the greatest strengths of Zig.

I discovered Zig two days ago; this pull request may contain unidiomatic code. 🙈


lib/std/crypto/aes/soft.zig
```diff
@@ -410,333 +411,94 @@ pub const Aes256 = struct {
...
-    0x60, 0x81, 0x4f, 0xdc, 0x22, 0x2a, 0x90, 0x88, 0x46, 0xee, 0xb8, 0x14, 0xde, 0x5e, 0x0b, 0xdb,
-    0xe0, 0x32, 0x3a, 0x0a, 0x49, 0x06, 0x24, 0x5c, 0xc2, 0xd3, 0xac, 0x62, 0x91, 0x95, 0xe4, 0x79,
-    0xe7, 0xc8, 0x37, 0x6d, 0x8d, 0xd5, 0x4e, 0xa9, 0x6c, 0x56, 0xf4, 0xea, 0x65, 0x7a, 0xae, 0x08,
-    0xba, 0x78, 0x25, 0x2e, 0x1c, 0xa6, 0xb4, 0xc6, 0xe8, 0xdd, 0x74, 0x1f, 0x4b, 0xbd, 0x8b, 0x8a,
-    0x70, 0x3e, 0xb5, 0x66, 0x48, 0x03, 0xf6, 0x0e, 0x61, 0x35, 0x57, 0xb9, 0x86, 0xc1, 0x1d, 0x9e,
-    0xe1, 0xf8, 0x98, 0x11, 0x69, 0xd9, 0x8e, 0x94, 0x9b, 0x1e, 0x87, 0xe9, 0xce, 0x55, 0x28, 0xdf,
-    0x8c, 0xa1, 0x89, 0x0d, 0xbf, 0xe6, 0x42, 0x68, 0x41, 0x99, 0x2d, 0x0f, 0xb0, 0x54, 0xbb, 0x16,
-};
+// Rijndael's irreducible polynomial.
+const poly: u16 = 1<<8 | 1<<4 | 1<<3 | 1<<1 | 1<<0; // x⁸ + x⁴ + x³ + x + 1
```
jedisct1, line 416: The polynomial could be a `u9`.

lib/std/crypto/aes/soft.zig
```diff
@@ -410,333 +411,94 @@ pub const Aes256 = struct {
...
+    }
+
+    break :init array;
 };
+
+// Multiply a and b as GF(2) polynomials modulo poly.
+fn mul(a: u8, b: u8) u8 {
+    @setEvalBranchQuota(27151);
+
+    var i: u16 = a;
```
jedisct1, line 485: `u9` everywhere.

lib/std/crypto/aes/soft.zig
```diff
@@ -410,333 +411,94 @@ pub const Aes256 = struct {
...
+        }
+
+        i <<= 1;
+
+        if (i & 0x100 != 0) {
+            i ^= poly;
+        }
+    }
+
+    return @intCast(u8, s);
```
jedisct1, line 503: `@truncate()` would be safe (and faster) here.

lib/std/crypto/aes/soft.zig
```diff
@@ -410,333 +411,94 @@ pub const Aes256 = struct {
...
 };
+
+// Multiply a and b as GF(2) polynomials modulo poly.
+fn mul(a: u8, b: u8) u8 {
+    @setEvalBranchQuota(27151);
+
+    var i: u16 = a;
+    var j: u16 = b;
+    var k: u16 = 1;
+    var s: u16 = 0;
```
0x2b3bfa0, line 488:
```suggestion
    var i: u9 = a;
    var j: u8 = b;
    var k: u8 = 1;
    var s: u9 = 0;
```

lib/std/crypto/aes/soft.zig
0x2b3bfa0:
```suggestion
const poly: u9 = 1<<8 | 1<<4 | 1<<3 | 1<<1 | 1<<0; // x⁸ + x⁴ + x³ + x + 1
```

jedisct1:
This is a good idea.

My only concern here, beyond increased compilation time, is that having a single table for encryption and decryption may not allow the compiler to remove the unneeded data in case we only use the encryption function (which is the case for virtually everything these days).

This is a bit of an issue with constrained platforms, which are likely to require the software implementation.

lib/std/crypto/aes/soft.zig
```diff
@@ -410,333 +411,94 @@ pub const Aes256 = struct {
...
+            array[mode][3][index] = math.rotr(u32, array[mode][0][index], 24);
+        }
+    }
+
+    break :init array;
 };
+
+// Multiply a and b as GF(2) polynomials modulo poly.
+fn mul(a: u8, b: u8) u8 {
+    @setEvalBranchQuota(27151);
```
jedisct1, line 483: You can round that up a little bit, as the number depends on the target but can also change with compiler versions.

0x2b3bfa0:
```suggestion
    @setEvalBranchQuota(30000);
```

For instance? I was tempted to use 28000, but don't know how much it can fluctuate.

jedisct1: 30000 should be fine.

CI's failing, but that may just be because of a formatting issue.

0x2b3bfa0:
> _having a single table for encryption and decryption may not allow the compiler to remove the unneeded data_

Would https://github.com/ziglang/zig/pull/11612/commits/e35cec1709c666e9836e1157d07fb5bd5d7ce8a6 solve this issue? 🤔

0x2b3bfa0:
_(Note that I'm a renowned expert in [cache invalidation](https://skeptics.stackexchange.com/a/39178); feel free to suggest better names for functions and constants)_

lib/std/crypto/aes/soft.zig
0x2b3bfa0: Yes, I forgot to run `zig fmt` 😅 

See https://github.com/ziglang/zig/pull/11612/commits/d559caa93488bc11d76123d4d5f79c11318171ac

lib/std/crypto/aes/soft.zig
daurnimator: huh? isn't `@intCast` far more correct here?

lib/std/crypto/aes/soft.zig
```diff
@@ -410,333 +411,99 @@ pub const Aes256 = struct {
...
+    } else {
+        array[0x00] = 0x63;
+    }
+
+    return array;
+}
+
+// Generate lookup tables for encryption and decryption.
+fn generateTable(sbox: [256]u8, factors: [4]u8) [4][256]u32 {
+    var array: [4][256]u32 align(64) = undefined;
```
daurnimator, line 470: Do we want/need the `align(64)` here?
Does return value semantics kick in without it?

lib/std/crypto/aes/soft.zig
0x2b3bfa0: Yes, there is no need to `@truncate` here because final values are always `u8` and `@intCast` is more accurate.

Still, if the former is more performant, I guess that it's fine to use it.

lib/std/crypto/aes/soft.zig
```diff
@@ -435,7 +435,9 @@ const table_decrypt align(64) = generateTable(sbox_decrypt, [_]u8{ 0xb, 0xd, 0x9
 
 // Generate S-box substitution values.
 fn generateSbox(invert: bool) [256]u8 {
-    var array: [256]u8 align(64) = undefined;
+    @setEvalBranchQuota(2000);
```
jedisct1, line 438: Still not enough :) Try 5000.

0x2b3bfa0:
```suggestion
    @setEvalBranchQuota(5000);
```

It worked on my computer! (TM)

lib/std/crypto/aes/soft.zig
```diff
@@ -410,333 +411,101 @@ pub const Aes256 = struct {
...
+const sbox_encrypt align(64) = generateSbox(false);
+const sbox_decrypt align(64) = generateSbox(true);
+const table_encrypt align(64) = generateTable(sbox_encrypt, [_]u8{ 0x3, 0x1, 0x1, 0x2 });
+const table_decrypt align(64) = generateTable(sbox_decrypt, [_]u8{ 0xb, 0xd, 0x9, 0xe });
+
+// Generate S-box substitution values.
+fn generateSbox(invert: bool) [256]u8 {
+    @setEvalBranchQuota(5000);
+
+    var array: [256]u8 = undefined;
```
jedisct1, line 440: Maybe `array` could be called `sbox`?

lib/std/crypto/aes/soft.zig
```diff
@@ -410,333 +411,101 @@ pub const Aes256 = struct {
...
+    } else {
+        array[0x00] = 0x63;
+    }
+
+    return array;
+}
+
+// Generate lookup tables for encryption and decryption.
+fn generateTable(sbox: [256]u8, factors: [4]u8) [4][256]u32 {
+    var array: [4][256]u32 = undefined;
```
jedisct1, line 472: And here, `array` could be called `table`?

lib/std/crypto/aes/soft.zig
jedisct1: CI seems to be passing so far :)

lib/std/crypto/aes/soft.zig
0x2b3bfa0: Resolved with https://github.com/ziglang/zig/pull/11612/commits/54a2f57c7e8e7759f8254a83fc5253e484ba40ec

lib/std/crypto/aes/soft.zig
0x2b3bfa0: Resolved with https://github.com/ziglang/zig/pull/11612/commits/54a2f57c7e8e7759f8254a83fc5253e484ba40ec

jedisct1:
CI is looking good so far. Fingers crossed!

0x2b3bfa0:
😅

jedisct1:
Merged! 🎉

Thanks a lot for contributing this, and welcome to Zig!

0x2b3bfa0:
Thank you very much! 😊

PR #11616, opened 2022-05-09
std.crypto: simplify AES multiplication algorithm

0x2b3bfa0:
“Byte-sized” continuation of #11612, making the `mul` function more intuitive and readable. 

_(Based on [FIPS-197 4.2.1](https://csrc.nist.gov/csrc/media/publications/fips/197/final/documents/fips-197.pdf))_


eLeCtrOssSnake:
Does it make sense to change it? I don't see any reason.

jedisct1:
> Does it make sense to change it? I don't see any reason.

Not really. It's purely cosmetic; it makes the fact that bits are processed until the last top set bit a little bit more obvious.

The end result is the same, and it's all comptime stuff anyway, so why not.

PR #11626, opened 2022-05-10
init-exe: add note about log_level in ReleaseSmall and ReleaseFast build mode

### LINKED ISSUE #9945: change the init-exe example to avoid confusion with ReleaseFast and std.log.info
### Zig Version

0.9.0-dev.1352+6d6cf5984

### Steps to Reproduce

just init simple zig project and build it with -Drelease-fast=true or -Drelease-small=true

```zig
const std = @import("std");

pub fn main() anyerror!void {
}
```

### Expected Behavior

Produced binary should print `All your codebase are belong to us.`

### Actual Behavior

It doesn't print anything. Instead it just exits immediately.

mewmew:
As suggested in https://github.com/ziglang/zig/issues/9945#issuecomment-950114977 by @wizzard0.

Fixes #9945.

### DIFF
--- a/lib/init-exe/src/main.zig
+++ b/lib/init-exe/src/main.zig
@@ -1,6 +1,8 @@
 const std = @import("std");
 
 pub fn main() anyerror!void {
+    // Note that info level log messages are by default printed only in Debug
+    // and ReleaseSafe build modes.
     std.log.info("All your codebase are belong to us.", .{});
 }
 


lib/init-exe/src/main.zig
```diff
@@ -1,6 +1,14 @@
...
+//pub const log_level: std.log.Level = .info;
+
 pub fn main() anyerror!void {
+    // Note: log messages with info level are by default emitted only in Debug
+    // and ReleaseSafe build modes (and omitted from ReleaseFast and
+    // ReleaseSmall build modes).
+    //
+    // To emit the message in all build modes, either set the root log_level
+    // explicitly or use std.debug.print instead of std.log.info.
     std.log.info("All your codebase are belong to us.", .{});
```
r00ster91, line 12: Perhaps a simple "Note that info logs are by default only printed in `Debug` and `ReleaseSmall` build modes" suffices? If they want to learn more (maybe they don't) about it, they can simply consult the `std.log` documentation, just like with all other things.

jedisct1: >  only printed in Debug and ReleaseSmall build modes"

`Debug` and `ReleaseSafe`.

r00ster91: You're right, mixed those up.

mewmew: > Perhaps a simple "Note that info logs are by default only printed in Debug and ReleaseSmallReleaseSafe build modes" suffices? If they want to learn more (maybe they don't) about it, they can simply consult the std.log documentation, just like with all other things.

Thanks @r00ster91! I updated the comment to simplify it.

Cheers,
Robin

mewmew:
This PR is ready for merge.

jedisct1:
Thank you!

PR #2095, opened 2019-03-22
std: add ascii with C ASCII character classes

shawnl:
I needed a better toUpper() than the poor version in fmt/parse_float.zig working on #2047, and ended up with this.

Does NOT look at the locale the way the C functions do.

       int isalnum(int c);
       int isalpha(int c);
       int iscntrl(int c);
       int isdigit(int c);
       int isgraph(int c);
       int islower(int c);
       int isprint(int c);
       int ispunct(int c);
       int isspace(int c);
       int isupper(int c);
       int isxdigit(int c);

       int isascii(int c);
       int isblank(int c);

       int toupper(int c);
       int tolower(int c);

Tested to match glibc (when using C locale) with this program:

const c = @cImport({
    // See https://github.com/ziglang/zig/issues/515
    @cDefine("_NO_CRT_STDIO_INLINE", "1");
    @cInclude("stdio.h");
    @cInclude("string.h");
    @cInclude("ctype.h");
});

const std = @import("std");
const ascii = std.ascii;
const abort = std.os.abort;

export fn main(argc: c_int, argv: **u8) c_int {
    var i: u8 = undefined;
    i = 0;
    while (true) {
        if (ascii.isAlNum(i) != (c.isalnum(i) > 0)) { abort(); }
        if (ascii.isAlpha(i) != (c.isalpha(i) > 0)) { abort(); }
        if (ascii.isCtrl(i) != (c.iscntrl(i) > 0)) { abort(); }
        if (ascii.isDigit(i) != (c.isdigit(i) > 0)) { abort(); }
        if (ascii.isGraph(i) != (c.isgraph(i) > 0)) { abort(); }
        if (ascii.isLower(i) != (c.islower(i) > 0)) { abort(); }
        if (ascii.isPrint(i) != (c.isprint(i) > 0)) { abort(); }
        if (ascii.isPunct(i) != (c.ispunct(i) > 0)) { abort(); }
        if (ascii.isSpace(i) != (c.isspace(i) > 0)) { abort(); }
        if (ascii.isUpper(i) != (c.isupper(i) > 0)) { abort(); }
        if (ascii.isXDigit(i) != (c.isxdigit(i) > 0)) { abort(); }
        if (i == 255) { break; }
        i += 1;
    }

    _ = c.printf(c"Success!\n");
    return 0;
}

PR #11629, opened 2022-05-10
`std.ascii`: general improvements

### LINKED ISSUE #8419: std.ascii isPunct returns true for several symbols

r00ster91:
This PR:

* Improves existing documentation and adds more.
* Adds more tests.
* Shows off Zig's *awesome* `comptime` feature by easily generating all the tables at compile time! This is much easier to read and to maintain. I mean the tables looked cool but I had no idea what they did or how they worked because there wasn't even any documentation. Now anyone should be able to add new things to it with a lot less effort. There are now naive functions which do what the exposed public equivalents do. They're used to generate those tables and they're used if the table isn't available: I thought why not make use of those naive functions and use them in place of the table implementation if we're in `ReleaseSmall` mode? It saves some 256 bytes.
* It makes `isAlNum` have its own dedicated table entry and removes `isGraph`'s table entry. I found this to be faster by one or a few instructions.

### DIFF
--- a/lib/std/ascii.zig
+++ b/lib/std/ascii.zig
@@ -59,51 +59,59 @@ pub const control_code = struct {
 // These naive functions are used to generate the lookup table
 // and they're used as fallbacks for if the lookup table isn't available.
 //
-// Note that some functions like for example `isDigit` don't use a table because it's slower.
-// Using a table is generally only useful if not all `true` values in the table would be in one row.
+// Note that even some very simple functions like `isDigit` use a table because it avoids
+// branching which is slow.
 
-fn isCntrlNaive(c: u8) bool {
-    return c <= control_code.US or c == control_code.DEL;
-}
 fn isAlphaNaive(c: u8) bool {
-    return isLower(c) or isUpper(c);
+    return isLowerNaive(c) or isUpperNaive(c);
 }
 fn isXDigitNaive(c: u8) bool {
-    return isDigit(c) or
+    return isDigitNaive(c) or
         (c >= 'a' and c <= 'f') or
         (c >= 'A' and c <= 'F');
 }
-fn isAlNumNaive(c: u8) bool {
-    return isDigit(c) or isAlphaNaive(c);
+fn isSpaceNaive(c: u8) bool {
+    return std.mem.indexOfScalar(u8, &spaces, c) != null;
+}
+fn isDigitNaive(c: u8) bool {
+    return c >= '0' and c <= '9';
+}
+fn isLowerNaive(c: u8) bool {
+    return c >= 'a' and c <= 'z';
+}
+fn isUpperNaive(c: u8) bool {
+    return c >= 'A' and c <= 'Z';
 }
 fn isPunctNaive(c: u8) bool {
-    @setEvalBranchQuota(3000);
     return (c >= '!' and c <= '/') or
         (c >= '[' and c <= '`') or
         (c >= '{' and c <= '~') or
         (c >= ':' and c <= '@');
 }
-fn isSpaceNaive(c: u8) bool {
-    @setEvalBranchQuota(5000);
-    return std.mem.indexOfScalar(u8, &spaces, c) != null;
+fn isAlNumNaive(c: u8) bool {
+    return isDigitNaive(c) or isAlphaNaive(c);
 }
 
 /// A lookup table.
 const CombinedTable = struct {
     table: [256]u8,
 
-    const Index = enum {
-        control,
+    // We cannot have >8 variants here which means we should choose
+    // only the most important/common functions to use a table for.
+    const Index = enum(u3) {
         alphabetic,
         hexadecimal,
-        alphanumeric,
+        space,
+        digit,
+        lower,
+        upper,
         punct,
-        spaces,
+        alnum,
     };
 
     /// Generates a table which is filled with the results of the given function for all characters.
     fn getBoolTable(comptime condition: fn (u8) bool) [128]bool {
-        @setEvalBranchQuota(2000);
+        @setEvalBranchQuota(7500);
         comptime var table: [128]bool = undefined;
         comptime var index = 0;
         while (index < 128) : (index += 1) {
@@ -115,22 +123,26 @@ const CombinedTable = struct {
     fn init() CombinedTable {
         comptime var table: [256]u8 = undefined;
 
-        const control_table = comptime getBoolTable(isCntrlNaive);
-        const alpha_table = comptime getBoolTable(isAlphaNaive);
-        const hex_table = comptime getBoolTable(isXDigitNaive);
-        const alphanumeric_table = comptime getBoolTable(isAlNumNaive);
+        const alphabetic_table = comptime getBoolTable(isAlphaNaive);
+        const hexadecimal_table = comptime getBoolTable(isXDigitNaive);
+        const space_table = comptime getBoolTable(isSpaceNaive);
+        const digit_table = comptime getBoolTable(isDigitNaive);
+        const lower_table = comptime getBoolTable(isLowerNaive);
+        const upper_table = comptime getBoolTable(isUpperNaive);
         const punct_table = comptime getBoolTable(isPunctNaive);
-        const whitespace_table = comptime getBoolTable(isSpaceNaive);
+        const alnum_table = comptime getBoolTable(isAlNumNaive);
 
         comptime var i = 0;
         inline while (i < 128) : (i += 1) {
             table[i] =
-                @boolToInt(control_table[i]) << @enumToInt(Index.control) |
-                @boolToInt(alpha_table[i]) << @enumToInt(Index.alphabetic) |
-                @boolToInt(hex_table[i]) << @enumToInt(Index.hexadecimal) |
-                @boolToInt(alphanumeric_table[i]) << @enumToInt(Index.alphanumeric) |
+                @boolToInt(alphabetic_table[i]) << @enumToInt(Index.alphabetic) |
+                @boolToInt(hexadecimal_table[i]) << @enumToInt(Index.hexadecimal) |
+                @boolToInt(space_table[i]) << @enumToInt(Index.space) |
+                @boolToInt(digit_table[i]) << @enumToInt(Index.digit) |
+                @boolToInt(lower_table[i]) << @enumToInt(Index.lower) |
+                @boolToInt(upper_table[i]) << @enumToInt(Index.upper) |
                 @boolToInt(punct_table[i]) << @enumToInt(Index.punct) |
-                @boolToInt(whitespace_table[i]) << @enumToInt(Index.spaces);
+                @boolToInt(alnum_table[i]) << @enumToInt(Index.alnum);
         }
 
         std.mem.set(u8, table[128..256], 0);
@@ -155,7 +167,7 @@ else
 /// Returns whether the character is alphanumeric. This is case-insensitive.
 pub fn isAlNum(c: u8) bool {
     if (combined_table) |table|
-        return table.contains(c, .alphanumeric)
+        return table.contains(c, .alnum)
     else
         return isAlNumNaive(c);
 }
@@ -172,14 +184,14 @@ pub fn isAlpha(c: u8) bool {
 ///
 /// See also: `control`
 pub fn isCntrl(c: u8) bool {
-    if (combined_table) |table|
-        return table.contains(c, .control)
-    else
-        return isCntrlNaive(c);
+    return c <= control_code.US or c == control_code.DEL;
 }
 
 pub fn isDigit(c: u8) bool {
-    return c >= '0' and c <= '9';
+    if (combined_table) |table|
+        return table.contains(c, .digit)
+    else
+        return isDigitNaive(c);
 }
 
 pub fn isGraph(c: u8) bool {
@@ -187,7 +199,10 @@ pub fn isGraph(c: u8) bool {
 }
 
 pub fn isLower(c: u8) bool {
-    return c >= 'a' and c <= 'z';
+    if (combined_table) |table|
+        return table.contains(c, .lower)
+    else
+        return isLowerNaive(c);
 }
 
 /// Returns whether the character has some graphical representation and can be printed.
@@ -204,7 +219,7 @@ pub fn isPunct(c: u8) bool {
 
 pub fn isSpace(c: u8) bool {
     if (combined_table) |table|
-        return table.contains(c, .spaces)
+        return table.contains(c, .space)
     else
         return isSpaceNaive(c);
 }
@@ -223,7 +238,10 @@ test "spaces" {
 }
 
 pub fn isUpper(c: u8) bool {
-    return c >= 'A' and c <= 'Z';
+    if (combined_table) |table|
+        return table.contains(c, .upper)
+    else
+        return isUpperNaive(c);
 }
 
 /// Returns whether the character is a hexadecimal digit. This is case-insensitive.
--- a/lib/std/ascii.zig
+++ b/lib/std/ascii.zig
@@ -55,184 +55,157 @@ pub const control_code = struct {
     pub const XOFF = 0x13;
 };
 
-const tIndex = enum(u3) {
-    Alpha,
-    Hex,
-    Space,
-    Digit,
-    Lower,
-    Upper,
-    // Ctrl, < 0x20 || == DEL
-    // Print, = Graph || == ' '. NOT '\t' et cetera
-    Punct,
-    Graph,
-    //ASCII, | ~0b01111111
-    //isBlank, == ' ' || == '\x09'
-};
+// These naive functions are used to generate the lookup table
+// and they're used as fallbacks for if the lookup table isn't available.
+//
+// Note that some functions like for example `isDigit` don't use a table because it's slower.
+// Using a table is generally only useful if not all `true` values in the table would be in one row.
 
-const combinedTable = init: {
-    comptime var table: [256]u8 = undefined;
-
-    const mem = std.mem;
-
-    const alpha = [_]u1{
-        //  0, 1, 2, 3, 4, 5, 6, 7 ,8, 9,10,11,12,13,14,15
-        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-
-        0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
-        1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0,
-        0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
-        1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0,
-    };
-    const lower = [_]u1{
-        //  0, 1, 2, 3, 4, 5, 6, 7 ,8, 9,10,11,12,13,14,15
-        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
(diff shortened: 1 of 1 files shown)


lib/std/ascii.zig
```diff
@@ -47,214 +51,171 @@ pub const control_code = struct {
...
+    else
+        return isWhitespaceNaive(char);
 }
 
-/// All the values for which isSpace() returns true. This may be used with
-/// e.g. std.mem.trim() to trim whiteSpace.
-pub const spaces = [_]u8{ ' ', '\t', '\n', '\r', control_code.VT, control_code.FF };
+/// All the values for which `isWhitespace()` returns `true`.
+/// This may be used with e.g. `std.mem.trim()` to trim whitespace.
+pub const whitespace = [_]u8{ ' ', '\t', '\n', '\r', control.VT, control.FF };
```
matu3ba, line 196: s/whitespace/whitespaces

lib/std/ascii.zig
```diff
@@ -272,15 +233,50 @@ pub fn toLower(c: u8) u8 {
 }
 
 test "ascii character classes" {
-    const testing = std.testing;
```
matu3ba, line 275: What do you think of a test to compare table vs runtime values?

lib/std/ascii.zig
```diff
@@ -47,214 +51,171 @@ pub const control_code = struct {
...
+                @boolToInt(alpha_table[i]) << @enumToInt(Index.alphabetic) |
+                @boolToInt(hex_table[i]) << @enumToInt(Index.hexadecimal) |
+                @boolToInt(alphanumeric_table[i]) << @enumToInt(Index.alphanumeric) |
+                @boolToInt(whitespace_table[i]) << @enumToInt(Index.whitespace);
+        }
 
-pub fn isAlNum(c: u8) bool {
-    return (combinedTable[c] & ((@as(u8, 1) << @enumToInt(tIndex.Alpha)) |
-        @as(u8, 1) << @enumToInt(tIndex.Digit))) != 0;
-}
```
matu3ba, line 194: Original code also contains `mem.set(u8, table[128..256], 0);`.

lib/std/ascii.zig
```diff
@@ -1,17 +1,21 @@
-// Does NOT look at the locale the way C89's toupper(3), isspace() et cetera does.
+//! The ASCII character encoding standard.
+//!
+//! See also: https://en.wikipedia.org/wiki/ASCII#Character_set
```
matu3ba, line 3: Ideally, the behavior for out of range could be specified here.

lib/std/ascii.zig
leecannon: Whitespace is a valid plural for whitespace (it's an uncountable noun).

lib/std/ascii.zig
r00ster91: Yes, this is no longer required because of the way `table` is initialized: `[_]u8{0} ** 256`. Now that I looked at it closer though I realize that this same `mem.set` would still work so I will use the old way again.

lib/std/ascii.zig
r00ster91: Are you talking about testing both the naive implementations and the table implementations? This does not seem to be required because unless the tests are compiled with `ReleaseSmall` (which they will probably not), the table implementation will be used which ultimately depends on the naive implementations to generate the table, so basically both implementations are already tested by this.

r00ster91:
I really wish we could use `u7` for this. It's the actual range that `std.ascii` works with and it'd fit perfectly.
There is this comment in the original:

> // I could have taken only a u7 to make this clear, but it would be slower
// It is my opinion that encodings other than UTF-8 should not be supported.
//
// (and 128 bytes is not much to pay).
// Also does not handle Unicode character classes.

Why would it be slower? I think through the type system we can make this not be slower.
I wanted to conduct some testing regarding this but sadly you can't even take a string like `"hello"` as `[]const u7` for a function parameter so language support is kind of missing here. Any opinions on this?
Wouldn't it be great if the following worked and we had automatic coercion?
```zig
pub fn lowerString(output: []u7, ascii_string: []const u7) []u7 {
    std.debug.assert(output.len >= ascii_string.len);
    for (ascii_string) |c, i| {
        output[i] = toLower(c);
    }
    return output[0..ascii_string.len];
}

pub fn main() void {
    var buf: [1024]u7 = undefined;
    lowerString(&buf, "HELLO WORLD"); // The string will be coerced because all characters are < 128 
}
```
Any opinions? One advantage of using `u7` would be that the combined table could always be 128 zero bytes smaller. This is not optimized out.

ifreund:
The main problem with using `u7` is that it is undefined behavior in the current stage1 zig compiler to cast between `[]const u8` and `[]const u7` and read the memory as a type differing from the type used for the last write. This is due to the fact that we use llvm's arbitrary bit width integers and therefore inherit their semantics. In particular, see the documentation of the load and store llvm IR instructions here: <https://llvm.org/docs/LangRef.html>

It is an open question whether we will change these semantics for the zig language spec, possibly enabling usage of `u7` here. For now I think we have choice but to stick to `u8`, at least where slices are concerned.

Note that there's a similar issue regarding the standard library's usage of u21 for unicode codepoints. It works fine as long as one is only concerned with a single codepoint but if UTF-32 text is required to be handled then a slice of u32 must be used instead.

r00ster91:
I see, yes, then it is probably better to leave it as-is for now. Maybe it can be changed at some point.

lib/std/ascii.zig
```diff
@@ -1,17 +1,41 @@
...
+const endsWithIgnoreCase = @compileError("deprecated: use endsWithInsensitive");
+const control_code = @compileError("deprecated: use control");
+const isBlank = @compileError("deprecated: use `char == ' ' or char == '\t'`");
+const isXDigit = @compileError("deprecated: use isHexadecimal");
+const isCntrl = @compileError("deprecated: use isControl");
+const isAlpha = @compileError("deprecated: use isAlphabetic");
+const isSpace = @compileError("deprecated: use isWhitespace");
+const spaces = @compileError("deprecated: use whitespace");
+const isPunct = @compileError("deprecated: write your own function suited to your particular case");
+const isGraph = @compileError("deprecated: use `isPrintable(char) and char != ' '`");
```
marler8997, line 33: Could we leave these implementations in for a release cycle with a deprecated comment note?  Having both the old and new names available for a release cycle gives projects and their dependencies time to upgrade without breaking.

r00ster91: Hmm, well yeah, that's what I think "deprecation" means too: for some time whatever is deprecated will still work as before but you will get a warning of some kind that at some point the thing will get removed and you are asked to use X instead. In Zig however apparently the definition of "deprecation" is different in that you immediately get an error and get told to use something else. Also, in Zig there are no warnings. How would even do this?
Here are a few examples: https://github.com/ziglang/zig/blob/902df103c6151c257c90de9ba5f29f7f4b9dbea2/lib/std/log.zig#L227
https://github.com/ziglang/zig/blob/360ecc1a2f72967f3a3882b3327e130bdc4e18c0/lib/std/math.zig#L177
And I've simply followed suit and did it that way too.

marler8997: It's a relatively little amount of work in the std library and saves a whole bunch of work/headaches for everyone else.  When you make a breaking change to a library, if it's possible to put the library into a state where both the old/new interfaces are supported, then you get the best of both worlds.  For one, you're not breaking anyone's code right away.  For two, people won't be blocked from moving to the new interface.  If you break the old interface, then you can't move to the new interface until all your dependencies move first.  Having a deprecation period where both are supported solves these issues very nicely.

jayschwa:
I think you should split this into separate pull requests. One for API changes and one for internal lookup table improvements. While I am in favor of expanding abbreviated names, I think removing functions deserves a more focused pull request. The existing ones are related to [C's character classification](https://en.wikipedia.org/wiki/C_character_classification), and I don't think removing them should be taken lightly.

jayschwa:
> It is an open question whether/how we will change these semantics for the zig language spec, possibly enabling usage of `u7` here. For now I think we have no choice but to stick to `u8`, at least where slices are concerned.
> 
> Note that there's a similar issue regarding the standard library's usage of u21 for unicode codepoints. It works fine as long as one is only concerned with a single codepoint but if UTF-32 text is required to be handled then a slice of u32 must be used instead.

Would it be reasonable to do `pub const Char = u8;` and use that for the API's parameter types? Then it might be easier to test or implement a hypothetical switch to `u7` in the future.

jayschwa:
> I personally don't agree as this PR isn't necessarily what the final API will look like anyway. People can always follow up with PRs and issues to discuss and improve it further.

It's true the API is not set in stone yet, but [C character class support was a deliberate addition](https://github.com/ziglang/zig/pull/2095), and I think more consideration should be given before starting to undo that. Meanwhile, the `comptime` table generation seems less controversial and could probably merge sooner if it were separate.

ifreund:
> > I think you should split this into separate pull requests.
> 
> I would like to get a word from a team member here if that's a requirement.

Breaking up changes into smaller patches makes it *far* easier to review and merge your changes. In particular, keeping implementation detail improvements separate from API bikeshedding avoids the former never being merged due to disagreement about the API or hesitation due to the API breakage.

> I personally don't agree as this PR isn't necessarily what the final API will look like anyway.

Ideally, we should always shoot for an API that we would be OK with maintaining indefinitely. Why would we merge these breaking changes now if we'll need to break everything again down the line?

r00ster91:
I already spent a lot of time on this PR. How about I reorganize the commits and split the table changes, the API changes, and whatever else in different commits? I agree that commit https://github.com/ziglang/zig/pull/11629/commits/05d8422d5ad492742e1268e947c44241935e2d73 isn't very organized. Would that be fine?

marler8997:
Apologies for bikeshedding but this one feels important to me. I much prefer `IgnoreCase` over `Insensitive` as it seems more clear to me what it means without having to look through documentation for clarification.

r00ster91:
I don't have a big opinion on that but I think I might also prefer `IgnoreCase` more. I was simply doing what this TODO told me to do: https://github.com/r00ster91/zig/blob/master/lib/std/ascii.zig#L419

Anyway, this is out of my hands now because apparently this PR was unreviewable so I removed all breaking changes. Could you take a look again?

Vexu:
> I was simply doing what this TODO told me to do: https://github.com/r00ster91/zig/blob/master/lib/std/ascii.zig#L419

See https://github.com/ziglang/zig/commit/490654c332f2d8eaf7edffa35ea0523800df998d#r47643278

r00ster91:
I think people could easily be misled by the outdated the description and the "breaking" label on this PR.
I've updated the description. This PR does not do any breaking changes. Can a team member kindly remove the "breaking" label and or review this PR?

lib/std/ascii.zig
```diff
@@ -47,190 +50,170 @@ pub const control_code = struct {
...
+        hexadecimal,
+        alphanumeric,
+        punct,
+        spaces,
+    };
 
-const combinedTable = init: {
-    comptime var table: [256]u8 = undefined;
+    /// Generates a table which is filled with the results of the given function for all characters.
+    fn getBoolTable(comptime condition: fn (u8) bool) [128]bool {
```
marler8997, line 105: This is pretty slick.  There is one potential improvement we could consider, namely, combine all the tables into 1 `u8` table.  If I remember correctly, an array of `bool` still takes 1-byte per element.   So we could represent up to 8 tables in the same amount of memory as 1 `bool` table.  The cost is likely a few more instructions (maybe just one) to access the data since we'd need to mask 1 specific bit.

r00ster91: It sounds kind of useful but just so I understand the purpose correctly:

> There is one potential improvement we could consider, namely, combine all the tables into 1 `u8` table.

Right, that's what we're doing in `init`: we combine all tables into a single `[256]u8` table. Does your idea change the ultimate result of this single table that exists at runtime?
Will perhaps with your idea the runtime table instead be `[128]u8`?

>  If I remember correctly, an array of bool still takes 1-byte per element.

This seems to be correct.

> So we could represent up to 8 tables in the same amount of memory as 1 `bool` table.

So shouldn't that mean we would go from `[128]bool` to `[16]u8` because 128 / 8 = 16? Like we can represent the same data with `[128]bool` or `[16]u8`.

> The cost is likely a few more instructions (maybe just one) to access the data since we'd need to mask 1 specific bit.

Your idea overall sounds to me like an optimization that again reduces binary size at the cost of some speed. It would probably make sense to gate this behind `@import("builtin").mode == .ReleaseSmall` as well, like we're already doing it: we're omitting the table entirely from the binary in that build mode.

rohlem: If this is implemented, also consider `std.StaticBitSet` (from `lib/std/bit_set.zig`), which gives a named abstraction for exactly this kind of use case (I think).

marler8997: > Your idea overall sounds to me like an optimization that again reduces binary size at the cost of some speed. 

Here's some background on how I think this would affect memory/performance if you're interested.

Since there are 6 tables, if I'm understanding correctly, we'd go from 6 * 128 bytes down to 1 table of 128 bytes.  The extra cost might be a few extra instructions to access the data.  Note that on modern hardware, if our design saves us 1 cache miss, that means it would take hundreds OR thousands of extra instructions to be just as slow as that single cache miss (since cache misses are so slow).  Also note that most processors use 64-byte cache lines so this could be saving us 10 cache lines of memory, which is alot more potential for cache misses.

That's just some background on Data Oriented Design, the principles Andrew has been incorporating into the compiler recently and seeing big performance returns.  Checkout https://vimeo.com/649009599 if you haven't seen it.

In this case, my guess is saving that extra memory would not only lower the memory footprint, but also result in faster performance.  Of course performance depends on many things, especially the rest of the program and how much memory pressure there actually is.  But I've found that maintaining a smaller memory footprint tends to have alot more affect on performance than saving on instructions.  This is also just theory though, always benchmark to verify.

r00ster91: > Since there are 6 tables, if I'm understanding correctly, we'd go from 6 * 128 bytes down to 1 table of 128 bytes.

But those 6 `[128]bool` tables only exist entirely at compile time. They never make it into the final runtime binary because all those 6 tables are temporary and are accumulated into a single `[256]u8` array. This sounds like a compile time optimization to me? I'm not sure if it makes sense to invest a lot of time into it because it doesn't affect runtime speed or how much memory is used at runtime.

Of course I would like to go from a `[256]u8` table to `[128]u8` because the remaining 128 bytes in that runtime table are always zero. These zeroes currently serve the purpose of returning `false` in all those cases where the character is outside of the 7-bit ASCII range. This is technically possible simply by going from `u8` to `u7` as I proposed here: https://github.com/ziglang/zig/pull/11629#issuecomment-1123305844
Sadly, as @ifreund explained, that is not (yet?) feasible.
But even so, 256 bytes is an okay price to say I think. I think it's a good balance between speed and memory usage. And as always in `ReleaseSmall`, the table doesn't even exist so I think that's compensated by that as well.

> That's just some background on Data Oriented Design, the principles Andrew has been incorporating into the compiler recently and seeing big performance returns. Checkout https://vimeo.com/649009599 if you haven't seen it.

That looks like an interesting talk. I might take a look. Thanks!

But apart from that, I think I get your point. It sounds good.
It's convenient that we have those bit sets in the `std` in case we need them.

ifreund: @marler8997: @r00ster91 is correct, the `[128]bool` arrays you are talking about won't be present in the compiled binary, they are just part of the code that goes into generating a single `[256]u8` which is already as compact as we can currently make it.

marler8997: > they are just part of the code that goes into generating a single [256]u8 which is already as compact as we can currently make it.

Ok great carry on then.  I stopped reading the code to soon.

ifreund:
> * It makes some functions like `isDigit` no longer use a table - `char >= '0' and char <= '9'` is actually faster, unless I'm doing my research wrong: https://zig.godbolt.org/z/sWYYr5WcK

How are you determining that it's faster? Counting the instructions generated by the compiler?

The version using a table is branchless while the other version uses a branch. I would expect the table to be faster in benchmarks.

r00ster91:
@ifreund 
```zig
const std = @import("std");
const debug = std.debug;
const time = std.time;

const ascii = @import("ascii.zig");

pub fn main() !void {
    var before = time.nanoTimestamp();

    const file = try std.fs.openFileAbsolute("path_to_file_with_byte", .{});

    const c = try file.reader().readByte();

    var index: usize = 0;
    while (index < 10000) : (index += 1)
        std.mem.doNotOptimizeAway(ascii.isDigit(c));

    debug.print("{}\n", .{time.nanoTimestamp() - before});

    before = time.nanoTimestamp();

    try file.seekTo(0);
    const c2 = try file.reader().readByte();

    index = 0;
    while (index < 10000) : (index += 1)
        std.mem.doNotOptimizeAway(ascii.isDigitTable(c2));

    debug.print("{}\n", .{time.nanoTimestamp() - before});
}
```
```
$ zig run -O ReleaseFast x.zig
24861
7054
```
Wow, I think you're right! Good catch.

r00ster91:
@ifreund Now while trying to fix this I encounter this issue:
```zig
    const Index = enum(u3) {
        alphabetic,
        hexadecimal,
        space,
        digit,
        lower,
        upper,
        punct,

        control,
        alphanumeric,
```
We have too many variants. With `alphanumeric` we have one variant too much. Do you know what to do about this? Should we actually add a second table? It can't become `u4` right?

jayschwa:
> We might even want to allow the user to choose what to include in the table, so we would provide some kind of customization for their specific usage.

I may be mistaken, but having separate tables per function might have that desired effect. Unused functions and tables would be eliminated from the final executable, and no need for explicit configuration.

Example:

```zig
const fooTable: [_]u1 = ...
pub fn isFoo(c: u8) { ... }

const barTable: [_]u1 = ...
pub fn isBar(c: u8) { ... }

const bazTable: [_]u1 = ...
pub fn isBaz(c: u8) { ... }
```

If user code only calls `isBar`, then the `fooTable` and `bazTable` will be eliminated.

This is moving back towards what is already merged into the master branch, but the tables could still be built at comptime from the "naive" functions. That still seems like a clear improvement over the current hard-coded tables in master.

r00ster91:
> > We might even want to allow the user to choose what to include in the table, so we would provide some kind of customization for their specific usage.
> 
> I may be mistaken, but having separate tables per function might have that desired effect. Unused functions and tables would be eliminated from the final executable, and no need for explicit configuration.
> 
> Example:
> 
> ```zig
> const fooTable: [_]u1 = { ... };
> pub fn isFoo(c: u8) { ... }
> 
> const barTable: [_]u1 = { ... };
> pub fn isBar(c: u8) { ... }
> 
> const bazTable: [_]u1 = { ... };
> pub fn isBaz(c: u8) { ... }
> ```
> 
> If user code only calls `isBar`, then the `fooTable` and `bazTable` will be eliminated.
> 
> This is moving back towards what is already merged into the master branch, but the tables could still be built at comptime from the "naive" functions. That still seems like a clear improvement over the current hard-coded tables in master.

You're right. Maybe we can just do something like that and don't actually have to tightly pack it like this. It would allow any function to use a fast lookup table. I would like to see how well that works but I won't do it in this PR.

@ifreund does it look mergeable?

jayschwa:
> I may be mistaken, but having separate tables per function might have that desired effect. Unused functions and tables would be eliminated from the final executable, and no need for explicit configuration.

It seems that one downside of `[256]u1` is that it takes up 256 bytes when it could theoretically be packed into 16 bytes. I'm not sure if future Zig will be capable of doing that. `PackedIntArray` seems to be the current way of achieving that outcome.

r00ster91:
I just remembered about https://github.com/ziglang/zig/issues/8419 and that we wanted to remove `isPunct` anyway so then we could have something more important in place of it and add that to the table.
But first it'd be great if we could merge this PR so that I can continue with all that stuff @andrewrk

topolarity:
> The version using a table is branchless while the other version uses a branch. I would expect the table to be faster, or at least more consistent, in benchmarks.

Small drive-by correction: Both versions are actually branchless. The table-free version generates a `cmp` but doesn't actually branch on it. The benchmark above is also misleading because it includes the file opening/reading in the measurements.

Given that the lookup table is dependent on cache performance, I think the table-free version is preferable.

r00ster91:
After some discussion @topolarity and came to the conclusion that it's probably better to remove the LUT (Look-Up Table) altogether because relying on the cache like that is not that good and some benchmarks from above were definitely wrong. For now I started with #12448 to supersede this PR and get the renamings in and then removing the LUT can come later.

r00ster91:
To avoid further confusion, I will close this PR before the superseder #12448 is merged. All this PR does is done by #12448 minus the LUT changes because the LUT should probably be removed in a future PR.

PR #11633, opened 2022-05-11
Move std.testing.zig_exe_path into build options

SpexGuy:
This PR supercedes #10526

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -46,6 +46,11 @@ pub fn build(b: *Builder) !void {
     test_cases.addPackagePath("test_cases", "test/cases.zig");
     test_cases.single_threaded = single_threaded;
 
+    const test_options = b.addOptions();
+    test_options.addOption([]const u8, "zig_exe_path", b.zig_exe);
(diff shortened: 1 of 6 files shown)


andrewrk:
Is it required to use `--pkg-begin test_options test_options.zig --pkg-end` now in order to run the standard library tests?

SpexGuy:
Unfortunately yes, there are a small number of tests that use it. (I think it's one in std.ChildProcess where it launches the compiler).  We could change these tests, or advise people to run std lib tests through `zig build` instead.

andrewrk (changes requested):
* I don't want it to be necessary to use `--pkg-begin` / `--pkg-end` args to execute the standard library test suite.
 * Please have a look at 67d5bfefba48d28c02e2841f1a47a213d28d4693 and see if that changes anything

SpexGuy:
Thanks, looks like 67d5bfe removed all those dependencies.  I've removed the test_options package from build.zig and the linux CI script, it is no longer needed.

Issue #7320, opened 2020-12-06
amend the definition of @enumToInt to assert that the value is a valid tag


andrewrk:
Here are the 2 competing definitions:

1. `@enumToInt` always produces an integer, however if the enum tag value is undefined, then the resulting integer is also undefined. (value, not behavior)

2. `@enumToInt` asserts the enum tag has a defined value.

Status quo is (1). Note that the reverse operation, `@intToEnum` does in fact assert that the integer matches a valid enum tag.

xackus:
Perhaps the check should be behind a compiler flag if there is enough interest in it.
I can imagine it could be useful when debugging code that is doing some unsafe memory manipulation.
This could be extended to other invalid values achievable only through things like `@ptrCast`.

andrewrk:
This isn't something we can put behind a compiler flag because it changes the semantics of the language.

I'm pretty sure status quo is going to stay.

daurnimator:
> `@intToEnum` does in fact assert that the integer matches a valid enum tag.

FWIW this is troublesome when it comes to non-exhaustive enums. See e.g.
https://github.com/ziglang/zig/blob/6294c1136cb44285156c04f9f89f62b0f9e8e471/lib/std/fmt.zig#L419-L437

xackus:
> This isn't something we can put behind a compiler flag because it changes the semantics of the language.

Oh, that's right, I forgot it's not illegal unless you branch on it.

LemonBoy:
> FWIW this is troublesome when it comes to non-exhaustive enums. See e.g.

That's sounds a bug to me, non-exhaustive enums by definition accept every value (as long as it fits into the underlying integral type).

daurnimator:
> That's sounds a bug to me, non-exhaustive enums by definition accept every value (as long as it fits into the underlying integral type).

nevermind me; I was getting confused with `@tagName`

LemonBoy:
> > That's sounds a bug to me, non-exhaustive enums by definition accept every value (as long as it fits into the underlying integral type).
> 
> nevermind me; I was getting confused with `@tagName`

Yeah, that's annoying. Perhaps `@tagName` can return an empty string instead of failing hard.
Turning the result into an optional type when the enum is not exhaustive is not that nice (is it?)

zigazeljko:
> Turning the result into an optional type when the enum is not exhaustive is not that nice (is it?)

Why? You can do `@tagName(value) orelse ""` if you want an empty string.

[...]

PR #11638, opened 2022-05-12
add std.enums.hasName

marler8997:
Added `std.enums.hasName` that returns whether an enum value has a name.
This is useful for determining whether a non-exhaustive enum value
has a tag name that can be retrieved with `@tagName`.

### DIFF
--- a/lib/std/enums.zig
+++ b/lib/std/enums.zig
@@ -849,3 +849,32 @@ test "std.enums.EnumIndexer sparse" {
     try testing.expectEqual(E.b, Indexer.keyForIndex(1));
     try testing.expectEqual(E.c, Indexer.keyForIndex(2));
 }
+
+/// Returns true if `enum_value` has a name. This function always returns true
+/// for exhaustive enums.  It will return false if the enum is non-exhaustive
+/// and the value does not have a corresponding name.
+pub fn hasName(enum_value: anytype) bool {
+    const T = @TypeOf(enum_value);
+    const enum_info = switch (@typeInfo(T)) {
+        .Enum => |enum_info| enum_info,
+        else => @compileError("hasName requires an enum value but got " ++ @typeName(T)),
+    };
+    if (enum_info.is_exhaustive)
+        return true;
+
+    @setEvalBranchQuota(3 * enum_info.fields.len);
+    inline for (enum_info.fields) |enum_field| {
+        if (@enumToInt(enum_value) == enum_field.value)
+            return true;
+    }
+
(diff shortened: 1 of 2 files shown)


lib/std/fmt.zig
```diff
@@ -488,25 +488,14 @@ pub fn formatType(
         },
         .Enum => |enumInfo| {
             try writer.writeAll(@typeName(T));
-            if (enumInfo.is_exhaustive) {
+            if (enumInfo.is_exhaustive or std.enums.hasName(value)) {
```
marler8997, line 491: Note, even though `hasName` will return true if `enumInfo.is_exhaustive` is true, this messes up this function because zig no longer evaluates that codepath at comptime, which causes it to evaluate the `else` branch which does not compile in some cases.

lib/std/enums.zig
```diff
@@ -849,3 +849,32 @@ test "std.enums.EnumIndexer sparse" {
...
+        else => @compileError("hasName requires an enum value but got " ++ @typeName(T)),
+    };
+    if (enum_info.is_exhaustive)
+        return true;
+
+    @setEvalBranchQuota(3 * enum_info.fields.len);
+    inline for (enum_info.fields) |enum_field| {
+        if (@enumToInt(enum_value) == enum_field.value)
+            return true;
+    }
```
marler8997, line 869: Might be cool if we could create a `ComptimeStringMap` here instead.

marler8997, line 869: Might be cool if we could create a `ComptimeStringMap` here instead.

ifreund:
Note that this use-case was discussed in #7320. One possibility that came up was making `@tagName()` return an empty string non exhaustive enum values without a tag. This doesn't seem to have been written up as a proposal though.

InKryption:
@ifreund Would that not perhaps pose an ambiguity if one were to give an enum a member with the name `@""`?

marler8997:
Another variation could be to add a new built-in to get the tag name for non-exhaustive enums that has an optional return type.

andrewrk (changes requested):
I opened #12845.

If you mention this proposal in the doc comments for this function and rebase against latest master, then I'm in favor of merging this PR.

marler8997:
Actually if we're gonna make the return type of `@tagName` optional for non-exhaustive enums like you propose, it seems better to me to wait for that than introduce this new function that will become redundant once we have that.  I'll close this but am willing to finish it if you think this function still has value even with the proposal.

Issue #11640, opened 2022-05-12
`zig test` terminal output look a bit broken on Windows.


YeonJi2:
## Operating System
Windows 10 21H2(Build 19044.1645)

## Zig version
0.10.0-dev.2186+3a64693db

I tested this on both Linux(WSL2 Ubuntu) and Windows, and they are currently running exactly the same version.

## Steps to reproduce
1. Prepare a zig file with a test. Default one that `zig init-exe` generates is good enough.
```zig
const std = @import("std");

pub fn main() anyerror!void {
    std.log.info("All your codebase are belong to us.", .{});
}

test "basic test" {
    try std.testing.expectEqual(10, 3 + 7);
}
```
2. Run the test using `zig test`
```
> zig test main.zig
```

## Expected behavior
I'm not sure which one is the correct behavior, but looking at the Language Reference(https://ziglang.org/documentation/master/#toc-Zig-Test), the example output looks like this:
```
$ zig test introducing_zig_test.zig
1/1 test "expect addOne adds one to 41"... OK
All 1 tests passed.
```
And this does match with my output if I redirect stderr to a file:
```
1/1 test "basic test"... OK
All 1 tests passed.
```
But on Linux, the output simply says:
```
All 1 tests passed.
```

And here's when a test fails, which is a bit more closer to the output style in the example.
```
Test [1/1] test "basic test"... expected 10, found 11
Test [1/1] test "basic test"... FAIL (TestExpectedEqual)
(Stack trace follows...)
```

## Actual behavior
This is how it looks like on Windows, and it is definitely broken:
```
> zig test main.zig
Test [1/1] test "basic test"... 1/1 test "basic test"... OK
All 1 tests passed.
```

This is when a test fails:
```
> zig test main.zig
Test [1/1] test "basic test"... 1/1 test "basic test"... expected 10, found 11
Test [1/1] test "basic test"... FAIL (TestExpectedEqual)
FAIL (TestExpectedEqual)
(Stack trace follows...)
```
This almost looks like it tried to do both console output AND plaintext output at the same time.

I've tried this on Windows Terminal, Windows's built-in console, VSCode's terminal, and Alacritty terminal emulator. And they all have exactly the same behavior.

PR #11645, opened 2022-05-13
zig test: Add proper detection for Windows console

YeonJi2:
This fixes the issue #11640 that I created yesterday. With this, running `zig test` on Windows console finally shows like this:
```
> zig test src/main.zig
All 138 tests passed.
```


vincenzopalazzo (approved):
LGTM

PR #11648, opened 2022-05-13
std.os.linux: fix signature of setgroups

marler8997:
the list parameter should be a multi-item pointer rather than a single-item
pointer. see: https://linux.die.net/man/2/setgroups

> setgroups() sets the supplementary group IDs for the calling process...
> the size argument specifies the number of supplementary group IDs in the buffer pointed to by list.


ifreund (approved):
LGTM, Thanks!

PR #11649, opened 2022-05-13
target: Rename sparcv9 -> sparc64

koachan:
Rename all references of sparcv9 to sparc64, to make Zig align more with other projects.
Also, added new function to convert glibc arch name to Zig arch name, since it, like LLVM, also refers to the architecture as sparcv9.

This is based on the suggestion by @kubkon in PR 11847.
(https://github.com/ziglang/zig/pull/11487#pullrequestreview-963761757)


andrewrk:
Thanks!

Issue #6207, opened 2020-08-31
Remove fields on `type` redundant with @typeInfo


tadeokondrak:
I think these are already deprecated from scattered comments on issues, but it would be nice to have an accepted issue to point to.

- `Int`
    - `bit_count`
    - `is_signed`
- `Float`
    - `bit_count`
- `Pointer`
    - `Child`
    - `alignment`
- `Array`
    - `Child`
    - `len`
- `ErrorUnion`
    - `Payload`
    - `ErrorSet`
- `Optional`
    - `Child`
- `Fn`
    - `ReturnType`
    - `is_var_args`
    - `arg_count`

daurnimator:
Notably, if we remove these (which I think we should), they should be removed from error messages too.
Infact, the change in error messages can occur first.

tadeokondrak:
I think the only time it shows up in error messages is for the name of an inferred error set:
```cpp
buf_appendf(&err_set_type->name, "@TypeOf(%s).ReturnType.ErrorSet", buf_ptr(&fn_entry->symbol_name));
```
Not sure what to replace it with.

Vexu:
Duplicate of #2117 assuming it applies to all the other redundant fields.

lithdew:
@Vexu would it be possible to have `zig fmt` replace usages of these deprecated fields with their standard library equivalents?

Vexu:
No. `zig fmt` doesn't know anything about types and would just lead to more breakage. However I believe these are quite rare in anything made after the introduction of `@typeInfo`.

PR #11653, opened 2022-05-14
fix semantic error with std.os.linux.all_mask

marler8997:
all_mask is a value of type `sigset_t`, which is defined as an array type
`[N]u32`.  However, `all_mask` references `sigset_t.len`, but, the array type
does not have a `len` field.  Fix is to use `@typeInfo(sigset_t).Array.len`
instead.


andrewrk:
Array types do have a `len` field

Vexu:
> Array types do have a `len` field

Array values have, the `len` field on array types was removed by #6207.

PR #11663, opened 2022-05-17
std.unicode: add utf16 byte length and codepoints counting routines

matu3ba:
* comptime and runtime tests are based on tests for counting utf8 code points


matu3ba:
Suggestions by @ifreund were applied. Note, that I did not modify that function.
No idea, why git was unable to keep commit history during rebase on top of unchanged master.

PR #11665, opened 2022-05-17
stage2: wasm - Implement 128-bit integers

Luukdegram:
This implements 128-bit integers for all instructions, with the exception of those that require compiler-rt (such as multiplying, dividing, shifting). This PR also ensures it matches the C-ABI correctly with regards to lowering the arguments and returning a 128-bit integer. The next step is to implement compiler-rt support in the Wasm backend & linker.


andrewrk:
Nice work!

PR #11666, opened 2022-05-17
Add aliases to math builtins back into std.math

frmdstryr:
Provides inline aliases to functions originally in `std.math`. This is to keep backwards compatibility (makes #11532 not breaking anymore) and keep math code unified (eg no need to mix builtins with std.math).

### DIFF
--- a/lib/std/math.zig
+++ b/lib/std/math.zig
@@ -929,37 +929,6 @@ pub inline fn fabs(value: anytype) @TypeOf(value) {
     return @fabs(value);
 }
 
-/// Fused multiply add, similar to (a * b) + c, except only rounds once,
-/// and is thus more accurate.
-/// This is the same as calling the builtin @mulAdd
-pub inline fn fma(comptime T: type, a: T, b: T, c: T) T {
-    return @mulAdd(T, a, b, c);
-}
-
-/// Returns the minimum value of a and b. This builtin accepts integers,
-/// floats, and vectors of either. In the latter case, the operation is
-/// performed element wise.
-/// This is the same as calling the builtin @minimum
-pub inline fn fmin(a: anytype, b: @TypeOf(a)) @TypeOf(a) {
-    return @minimum(a, b);
-}
-
-/// Returns the maximum value of a and b. This builtin accepts integers,
-/// floats, and vectors of either. In the latter case, the operation is
-/// performed element wise.
-/// This is the same as calling the builtin @maximum
-pub inline fn fmax(a: anytype, b: @TypeOf(a)) @TypeOf(a) {
-    return @maximum(a, b);
-}
-
-/// Modulus division. For unsigned integers this is the same as numerator %
-/// denominator. Caller guarantees denominator > 0, otherwise the operation will
-/// result in a Remainder Division by Zero error when runtime safety checks are enabled.
-/// This is the same as calling the builtin @mod
-pub inline fn fmod(a: anytype, b: @TypeOf(a)) @TypeOf(a) {
-    return @mod(a, b);
-}
-
 /// Returns the absolute value of the integer parameter.
 /// Result is an unsigned integer.
 pub fn absCast(x: anytype) switch (@typeInfo(@TypeOf(x))) {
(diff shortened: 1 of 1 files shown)


viriuwu:
Relevant discussion from #stage2-meetings:
2022-05-17-190111_1139x513_scrot
2022-05-17-190120_1165x309_scrot

frmdstryr:
I was unaware of that conversation but it seems they had the same confusion.  I have the same questions about all the `@div...` and `@mem...` builtins.  

This broke code that has been working for 2 ish years.... and the relevant section of code doing heavy math took several weeks to implement and debug on a physical machine, I can't just go kick off some CI.  It just seems like an unnecessary breaking change to me.

lib/std/math.zig
```diff
@@ -270,6 +270,41 @@ pub const sinh = @import("math/sinh.zig").sinh;
 pub const cosh = @import("math/cosh.zig").cosh;
 pub const tanh = @import("math/tanh.zig").tanh;
 
+/// Sine trigonometric function on a floating point number.
+/// Uses a dedicated hardware instruction when available.
+/// This is an alias to the builtin @sin
```
andrewrk, line 275: This is technically not true, it's not an alias. I would reword this to "This is the same as calling..."

lib/std/math.zig
```diff
@@ -887,6 +922,19 @@ fn testRem() !void {
...
+pub inline fn fabs(value: anytype) @TypeOf(value) {
+    return @fabs(value);
+}
+
+/// Alias to fabs for backwards compatibility
+/// This is the same as calling the builtin @fabs
+pub inline fn absFloat(value: anytype) @TypeOf(value) {
+    return @fabs(value);
+}
+
```
andrewrk, line 937: We don't need to carry around backward compatibility cruft pre 1.0. This function is redundant and should be deleted.

andrewrk (changes requested):
Seems to be missing these:
 * `fma`
 * `fmin`
 * `fmax`
 * `fmod`

frmdstryr:
Are those new builtins or do they need imported from somewhere? It's failing to build.

```zig
-- Installing: /home/user/projects/zig/build/lib
/home/user/projects/zig/lib/std/math.zig:927:12: error: invalid builtin function: '@fma'
/home/user/projects/zig/lib/std/math.zig:932:12: error: invalid builtin function: '@fmin'
/home/user/projects/zig/lib/std/math.zig:937:12: error: invalid builtin function: '@fmax'
/home/user/projects/zig/lib/std/math.zig:942:12: error: invalid builtin function: '@fmod'
::
```

andrewrk:
Would love to get @tiehuis's thoughts on this, and any other relevant std.math organization guidance while we're at it :-)

lib/std/math.zig
```diff
@@ -929,6 +929,37 @@ pub inline fn fabs(value: anytype) @TypeOf(value) {
...
+/// and is thus more accurate.
+/// This is the same as calling the builtin @mulAdd
+pub inline fn fma(comptime T: type, a: T, b: T, c: T) T {
+    return @mulAdd(T, a, b, c);
+}
+
+/// Returns the minimum value of a and b. This builtin accepts integers,
+/// floats, and vectors of either. In the latter case, the operation is
+/// performed element wise.
+/// This is the same as calling the builtin @minimum
```
andrewrk, line 942: It's not the same as calling `@minimum` because `@minimum` accepts anytype for both parameters and the return type will depend on a combination of both types and whether each operand is comptime known.

Because of this, it kind of seems like this function should not be in std.math, in order to encourage use of the builtin.

andrewrk (changes requested):
If it feels like I'm being picky here, I'm sorry - I would have done this change eventually and gone through all these issues myself, but since you're doing a PR then you're experiencing my internal dialogue out loud

lib/std/math.zig
frmdstryr: Perhaps these could be done in a separate pr?

tiehuis:
I won't speak for a general overview because I may be missing some of the higher vision at this stage. I do like this original PR and think having the functions available in `std.math` makes good sense. Specifically, because there are other similar math functions which are not implemented as builtins (e.g. `tanh`). Having all related functions available in one namespace is valuable from a documentation/consistency point of view as an end-user.

andrewrk:
Thanks @frmdstryr!

PR #11667, opened 2022-05-18
Add package separation for nix-darwin users with a macOS Sdk installation present

haze:
Add `nix_include_dirs` and `nix_framework_dirs` for nix-darwin users who have packages installed that aren't relative to an existing SDK root


lib/std/zig/system/NativePaths.zig
```diff
@@ -11,6 +11,10 @@ const NativeTargetInfo = std.zig.system.NativeTargetInfo;
 include_dirs: ArrayList([:0]u8),
 lib_dirs: ArrayList([:0]u8),
 framework_dirs: ArrayList([:0]u8),
+
+nix_framework_dirs: ArrayList([:0]u8),
```
kubkon, line 15: I think we should figure out a way of satisfying nix-darwin without such special-casing like that.

kubkon (changes requested):

lib/std/zig/system/NativePaths.zig
haze: All nix users will have this problem. It's just that Darwin users also have an sdk root, so Zig attempts to add the paths in relation to, which is incorrect.

haze:
repro:

test_time.c
```c
#include <time.h>
#include <stdio.h>

int main() {
  struct timespec time;
  int status = clock_gettime(CLOCK_REALTIME, &time);
  printf("%ld\n", time.tv_sec);
}
```

shell.nix
```nix
{ pkgs ? import <nixpkgs> {}, pkgs_unstable ? import <nixpkgs-unstable> {}, }:
  pkgs.mkShell {
    nativeBuildInputs = [ 
      pkgs.llvmPackages_13.libclang
      pkgs.darwin.apple_sdk.MacOSX-SDK
    ];
}
```

(this gives you a library and also a clang to test against that doesn't have the sysroot baked in)

src/main.zig
```diff
@@ -2177,13 +2177,37 @@ fn buildOutputType(
...
+        // so even if an sdk is found (XCode installation), nix include/framework dirs must be added
+        // without relation
+        try clang_argv.ensureUnusedCapacity(paths.nix_include_dirs.items.len * 2);
+        for (paths.nix_include_dirs.items) |include_dir| {
+            clang_argv.appendAssumeCapacity("-isystem");
+            clang_argv.appendAssumeCapacity(include_dir);
+        }
+
+        try clang_argv.ensureUnusedCapacity(paths.nix_framework_dirs.items.len * 2);
+        for (paths.nix_include_dirs.items) |framework_dir| {
```
devins2518, line 2206: This should be `paths.nix_framework_dirs.items`

haze: You're correct, will change

PR #11668, opened 2022-05-18
Explicitly add the macOS Sdk include dir, since specifying the sysroot isn't enough

### LINKED ISSUE #11631: darwin libc not available during cImport
### Zig Version

0.10.0-dev.2122+f161d3875

### Steps to Reproduce

see initial report text here: https://zigbin.io/421921

basically, have a project that depends on a c library (in my case, `libclang`) and attempt to cImport/cInclude a header from that project

(in my case for nix, zig can't find the header I'm including, even though verbose-cc knows about it)
if I manually add the include path to the build.zig, I then see the error that time.h cannot be found

### Expected Behavior

A successful compilation

### Actual Behavior

An error complaining that `time.h` is not found

haze:
Fixes #11631

I went down a huge rabbit hole of trying non-specialized versions of clang (specifically those built by nix without a default sysroot (as nix adds them with the cc-wrapper)) and found that specifying the sysroot was not enough, and had to add the libC header directory manually. This is also what nix does here:

https://github.com/NixOS/nixpkgs/blob/451f200f676de18f7ecb54b1652d064c04a3c959/pkgs/build-support/cc-wrapper/default.nix#L329-L358 and https://github.com/NixOS/nixpkgs/blob/451f200f676de18f7ecb54b1652d064c04a3c959/pkgs/build-support/bintools-wrapper/default.nix#L231-L240

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -2177,9 +2177,19 @@ fn buildOutputType(
                 const sdk = std.zig.system.darwin.getDarwinSDK(arena, target_info.target) orelse
                     break :outer false;
                 native_darwin_sdk = sdk;
-                try clang_argv.ensureUnusedCapacity(2);
+
+                try clang_argv.ensureUnusedCapacity(4);
                 clang_argv.appendAssumeCapacity("-isysroot");
                 clang_argv.appendAssumeCapacity(sdk.path);
+
+                // NOTE: it seems like providing the sysroot isn't enough
+                // neither --sysroot or -isysroot allow clang to find the sdk libc headers,
+                // so we manually add them here
+
+                clang_argv.appendAssumeCapacity("-isystem");
+                const darwin_sdk_libc_path = try std.fs.path.join(arena, &[2][]const u8{ sdk.path, "usr/include/" });
+                clang_argv.appendAssumeCapacity(darwin_sdk_libc_path);
+
                 break :outer true;
             } else break :outer false;
         } else false;


kubkon:
Hmm, that doesn't look right to me. If you specify the sysroot manually, you ought to manually specify include and search dirs also, something like

```
zig cc --sysroot /some/root/path -I/usr/include -F/System/Library/Frameworks -L/usr/lib main.c
```

Additionally, the paths that are meant to be prefixed by the sysroot should always be absolute (prefixed with `/`) as far as I know. Otherwise the linker for instance will look directly in `${cwd}/usr/include` rather than `${sysroot}/usr/include`.

src/main.zig
```diff
@@ -2177,9 +2177,19 @@ fn buildOutputType(
...
+
+                try clang_argv.ensureUnusedCapacity(4);
                 clang_argv.appendAssumeCapacity("-isysroot");
                 clang_argv.appendAssumeCapacity(sdk.path);
+
+                // NOTE: it seems like providing the sysroot isn't enough
+                // neither --sysroot or -isysroot allow clang to find the sdk libc headers,
+                // so we manually add them here
+
+                clang_argv.appendAssumeCapacity("-isystem");
```
kubkon, line 2189: Hmm, I don't think this should be needed. The purpose of the sysroot is to prefix every absolute include dir with the said sysroot path. What you are doing here is adding an explicit `-isystem` path that circumvents the sysroot path altogether and explicitly includes the concatenation of `sysroot` with `usr/include`. I do not believe this is the right solution to the problem. Instead, we should do more investigation what paths are included and work out why clang cannot find the headers.

kubkon (changes requested):

PR #11669, opened 2022-05-18
[Elf] fix linker z_args

motiejus:
I can't get the current `-Wl,-z<...>` pass to the linker. Assume any C file that has `int main`:

    $ zig cc -Wl,-znow ./test.c -o test
    warning: unsupported linker arg: -znow

And, surely, the linker line did not have `-z now`.

Cross-checking with clang/gcc:

    $ clang-13 -v -Wl,-znow ./test.c -o test |& tail -1
     "/usr/bin/ld" --hash-style=both --build-id --eh-frame-hdr -m elf_x86_64 <...> -znow -zrelro <...>

Same for -zrelro, -znoexecstack and others.

After this commit zig does the right thing: observe `-z now` and `-z
relro` in the linker line:

     $ ~/code/zig/build/zig cc -v -Wl,-zrelro -Wl,-znow ./test.c -o test |& tail -1
    ld.lld -error-limit=0 -O0 -z stack-size=16777216 --gc-sections --eh-frame-hdr -z now -z relro <...>

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1655,12 +1655,20 @@ fn buildOutputType(
                     linker_global_base = parseIntSuffix(arg, "--global-base=".len);
                 } else if (mem.startsWith(u8, arg, "--export=")) {
                     try linker_export_symbol_names.append(arg["--export=".len..]);
-                } else if (mem.eql(u8, arg, "-z")) {
-                    i += 1;
-                    if (i >= linker_args.items.len) {
-                        fatal("expected linker extension flag after '{s}'", .{arg});
-                    }
-                    const z_arg = linker_args.items[i];
+                } else if (mem.startsWith(u8, arg, "-z")) {
+                    // -z param and -zparam are both supported. z_arg is whatever
+                    // comes after `-z` in either case.
+                    const z_arg = blk: {
+                        if (mem.eql(u8, arg, "-z")) {
+                            i += 1;
+                            if (i >= linker_args.items.len)
+                                fatal("expected linker extension flag after '{s}'", .{arg});
+                            break :blk linker_args.items[i];
+                        } else {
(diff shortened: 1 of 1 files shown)


Luukdegram:
I don’t think this is the right change as it also affects other linkers such as Wasm, Coff. For example to set the stack size you’d use `-z stack-size=<x>`. This change would break that. Note that lld’s (or really GNU ld since lld is a drop-in replacement) documentation also specifies -z *option*. Perhaps Clang supports both ways, but do we want that for Zig as well?

motiejus:
Ah, I didn't see the plain `-z` option. I will rework this diff. Thanks @Luukdegram

motiejus:
Thanks again, @Luukdegram , for pointing out `-z`. I think I cleaned up the diff by now.

motiejus:
This is ready for re-review.

Luukdegram:
Not a big fan of having multiple ways to specify the same thing where one of them isn't even documented, but perhaps @andrewrk has a different opinion on that. We now also introduce a difference for the same flag between `zig cc` and `zig build-{x}`. For example, if the user learns they can do `zig cc -znow` and then later try to apply it doing `zig build-exe example.c -znow` they'll be met with an error.

alexrp:
I'm confused. What's wrong with `zig cc -Wl,-z,now`?

motiejus:
> I'm confused. What's wrong with `zig cc -Wl,-z,now`?

I like it, but the use-in-the-wild differs.

Let's see what Go has to say about it in https://github.com/golang/go/pull/53030.

motiejus:
Go will do the right thing now with https://github.com/golang/go/commit/38607c553878da21b5042e63997ecb3b7201e684 , closing.

PR #11670, opened 2022-05-18
[Elf] coalesce linker `-z` args

motiejus:
This is a cosmetic change that will:
- make the z linker opts in link/Elf.zig easier to parse (and thus
  less error-prone to add new ones).
- reduce the argc to the linker subcommand.

I am planning to add a couple more options used by cgo, so making this
more egonomic will help me with the upcoming diff.


motiejus:
`ld` manual states `-z option`, so the `-zoption` is at least not a documented syntax. Closing.

PR #11671, opened 2022-05-18
stage2 bug fixes aimed towards more behavior tests passing

andrewrk:
See individual commit messages


kubkon:
This should be good to go, right? Previous CI failures seemed completely unrelated and random.

andrewrk:
I think so, but I haven't actually seen the macOS CI run to completion yet on this commit and I am a bit concerned about it. I restarted the job and it's running now.

Issue #7547, opened 2020-12-25
zig cc -c can only handle one file argument


hoanga:
_zig cc -c_ handles only one file argument.

gcc and clang both compile and generate object files for each respective file argument given.

also note that _zig cc -c_ handles multiple file arguments differently depending on platform.  i.e. on
Linux it will only compile the first argument given.   On Mac it returns the following
type of error message

`error: macho does not support linking multiple objects into one`

Example gcc run
```
$ gcc -c c1.c c2.c 
$ ls
c1.c  c1.o  c2.c  c2.o 
```

Example zig cc run

```
$ zig cc -c c1.c c2.c 
$ ls 
c1.c  c1.o  c2.c 
```

Mouvedia:
> [Zig actually supports this case, taking advantage of the -r flag of LLD to link multiple objects into one. However that only works for WASM and ELF.](https://github.com/ziglang/zig/issues/7094#issuecomment-737590796)

RodrigoDornelles:
same problem here

slimsag:
Note: This can sometimes surface in e.g. CMake projects if spurious flags like `/EHsc /guard:cf` end up on the `zig cxx` command line by accident.

BrookJeynes:
Hey guys, does anyone know the current progress on this bug? I seem to be running into it when trying to compile a c program that links with a zig program.

![image](https://github.com/ziglang/zig/assets/25432120/1fa123bf-7a87-4276-9b88-77b26c028dbd)

Here's my `build.zig` for reference.
```zig
const std = @import("std");

pub fn build(b: *std.Build) void {
    const target = b.standardTargetOptions(.{});
    const optimize = b.standardOptimizeOption(.{});

    const lib = b.addSharedLibrary(.{
        .name = "mathtest",
        .root_source_file = .{ .path = "src/mathtest.zig" },
        .version = .{ .major = 1, .minor = 0, .patch = 0 },
        .target = target,
        .optimize = optimize,
    });

    const obj = b.addObject(.{
        .name = "module",
        .target = target,
        .optimize = optimize,
    });
    obj.addCSourceFile(.{ .file = .{ .path = "src/test.c" }, .flags = &.{"-std=c99"} });
    obj.linkLibrary(lib);
    obj.linkSystemLibrary("c");
    b.default_step.dependOn(&obj.step);

    const install_kernel = b.addInstallFile(obj.getEmittedBin(), "kernel.o");
[...]

Issue #7094, opened 2020-11-13
error: LLDReportedFailure when trying to link with zig c++ on Linux


demizer:
I'm getting the following error with zig build (c++).

[build.zig](https://github.com/demizer/markzig/blob/master/build.zig#L24)

```
clang version 11.0.0 (git@github.com:ziglang/zig-bootstrap.git ddf83b9ed342a061dc82003c9e74e9b32f04393f)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-pc-linux-gnu/10.2.0
Found candidate GCC installation: /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/10.2.0
Found candidate GCC installation: /usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0
Found candidate GCC installation: /usr/lib64/gcc/x86_64-pc-linux-gnu/10.2.0
Selected GCC installation: /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/10.2.0
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Selected multilib: .;@m64
 (in-process)
[...]

PR #11678, opened 2022-05-19
[ld.lld] understand -r, --relocatable

motiejus:
clang and gcc accept `-r`, which is a `--relocatable` argument of
ld.lld. This diff understands the flag and passes to the linker.

Impact: this makes the Go's `cmd/link` test suite pass[*]. To reproduce,
do in go/src directory (currently patched, hopefully not for long):

```
echo 'exec zig cc "$@"' > zig-cc; chmod a+x zig-cc
CC="./zig-cc" ./run.bash -run cmd/link
```

[*]: at the moment it requires golang/go#52815 to pass


src/link/Elf.zig
```diff
@@ -1472,6 +1473,10 @@ fn linkWithLLD(self: *Elf, comp: *Compilation, prog_node: *std.Progress.Node) !v
             try argv.append("--emit-relocs");
         }
 
+        if (self.base.options.relocatable) {
+            try argv.append("--relocatable");
```
andrewrk, line 1477: Note line 1419. Unfortunately for this particular change it won't be this easy; it requires changes to the CLI, Compilation.zig, and may require reworking some additional logic. This is related to #7547 and #7094. Note also that COFF and MachO linkers do not support this flag.

andrewrk (changes requested):

src/link/Elf.zig
motiejus: Uh oh. Thanks.

I will file an issue instead. As far as my (very limited) scope is concerned, if `-r` is passed to `zig cc`, it should be passed to the linker.

motiejus:
Closing in favor of #11683

PR #11679, opened 2022-05-19
[Elf] add -z nocopyreloc

motiejus:
Warnings about non-implemented `-z nocopyreloc` are common when
compiling go code (including Go's tests themselves). Let's just
make it stop complaining.


andrewrk:
Thanks!

PR #11684, opened 2022-05-20
Set macOS/iPhoneOS/tvOS/watchOS ABI to none (unspecified) by default

### LINKED ISSUE #10318: Using @sin, @cos on x86_64-macos fails with missing symbol _sincosf when building optimized
### Zig Version

0.9.0-dev.1945+efdb94486

### Steps to Reproduce

Build the following file:
```zig
//  zig build-exe -O ReleaseSafe -target x86_64-macos repro.zig 
const std = @import("std");

pub fn main() !void {
    var runtime_val: f32 = @intToFloat(f32, std.time.timestamp());

    var mat = rotationMat(runtime_val);

    std.debug.print("{any}\n", .{mat});
}

fn rotationMat(angle: f32) [2][2]f32 {
    const s = @sin(angle);
    const c = @cos(angle);
    return .{
        .{ c, -s },
        .{ s, c },
    };
}
```

### Expected Behavior

The code builds successfully, like when building to `x86_64-linux`

### Actual Behavior

```console
[felix@denkplatte-v2 8eb0bcd9]$ zig build-exe -O ReleaseSafe -target x86_64-macos repro.zig 
error(link): undefined reference to symbol '_sincosf'
error(link):   first referenced in '/tmp/8eb0bcd9/zig-cache/o/967695a63a2a01c5bfcf22751e917e32/repro.o'
error: UndefinedSymbolReference
exit code: 1
```

kubkon:
Fixes #10318 

This is a breaking change which changes default macOS (and any Apple platform) ABI from `.gnu` to `.none`. With this PR, the following targets will result in an error:

```
$ zig build-exe main.zig -target aarch64-macos-gnu
$ zig build-exe main.zig -target aarch64-macos-musl
```

Additionally, when specifying the target triple when lowering to LLVM, we now append the version of the OS, ending up with something like:

```
target triple = "aarch64-unknown-macosx11.6.0-unknown"
```

This then enables further LLVM optimisations which would otherwise not happen such as lowering of subsequent `@sin` and `@cos` into single `_sincosf_sret` call, which is a symbol provided by the OS. I should note these optimisations are only in effect in stage2 LLVM backend - porting this logic into stage1 seems like too much work for very little gain.


src/codegen/llvm.zig
```diff
@@ -86,50 +86,76 @@ pub fn targetTriple(allocator: Allocator, target: std.Target) ![:0]u8 {
...
-        .uefi => "windows",
-
-        .opencl,
-        .glsl450,
-        .vulkan,
-        .plan9,
-        .other,
-        => "unknown",
+    var arena = std.heap.ArenaAllocator.init(allocator);
+    defer arena.deinit();
```
andrewrk, line 90: I suggest to use the pattern of appending to an ArrayList(u8) instead and then using toOwnedSlice() at the end of the function.

andrewrk (approved):
I have one nitpick but this PR looks quite mergeable despite it. Nice work!

src/codegen/llvm.zig
```diff
@@ -123,14 +124,30 @@ pub fn targetTriple(allocator: Allocator, target: std.Target) ![:0]u8 {
...
+
+    if (target.os.tag.isDarwin()) {
+        const min_version = target.os.version_range.semver.min;
+        const version_str = try std.fmt.allocPrint(allocator, "{d}.{d}.{d}", .{
+            min_version.major,
+            min_version.minor,
+            min_version.patch,
+        });
+        defer allocator.free(version_str);
+        try llvm_triple.appendSlice(version_str);
```
andrewrk, line 148:
```suggestion
        try llvm_triple.writer().print("{d}.{d}.{d}", .{
            min_version.major,
            min_version.minor,
            min_version.patch,
        });
```

andrewrk (changes requested):
Check out this pattern :grin: 

Other than that, looks ready to go!

src/codegen/llvm.zig
kubkon: Hahaha, ofc! Thanks Andrew, dunno why I missed this one...

PR #11687, opened 2022-05-21
stage1: don't lowercaseify lib names

### LINKED ISSUE #11682: stage1 force-lowercases(???) explicit library name in externs
```console
$ cat max.c
int max(int a, int b)
{
    if (a < b)
        return b;
    else
        return a;
}
$ gcc -c max.c -o libMAX.so
$ cat main.zig 
extern "MAX" fn max(a: c_int, b: c_int) c_int;

pub fn main() void {
    @import("std").debug.print("max(1, 2): {}\n", .{ max(1, 2) });
}
$ zig run -L. -lMAX main.zig
LLD Link... ld.lld: error: unable to find library -lmax
error: LLDReportedFailure
$ zig run -L. -lMAX -fno-stage1 main.zig
max(1, 2): 2
```

This does not occur if the filename is all lowercase, like `libmax.so`. This is interesting because it definitely sees the file to some extent, watch the error message change.

```console
$ rm libMAX.so
$ zig run -L. -lMAX main.zig
LLD Link... ld.lld: error: unable to find library -lMAX
ld.lld: error: unable to find library -lmax
error: LLDReportedFailure
```

Instead of passing `-fno-stage1` you can also remove the `"MAX"` and the linker will figure it out. So it appears this is specifically a case of `extern "name" fn` where the name contains uppercase letters, with stage1. I hope that's useful.

```console
$ sed -i.bk 's/"MAX"//' main.zig
$ zig run -L. -lMAX main.zig
max(1, 2): 2
```

## Zig Version
zig-linux-x86_64-0.10.0-dev.2306+50a5ddecc

viriuwu:
Stage1 converts library names to lowercase, which isn't good when they're not all lowercase. I suspect this was done because they're insensitively compared with a bunch of libc and libc++ names, but `is_libc_lib_name` and `is_libcpp_link_name` already do a case-insensitive comparison so all it does is mangle the string and break cases where it has uppercase characters (such as `libX11`, where I ran into it). This regression appears to have been introduced in 5fed42d.

closes #11682

### DIFF
--- a/src/stage1.zig
+++ b/src/stage1.zig
@@ -416,7 +416,7 @@ export fn stage2_add_link_lib(
     _ = symbol_name_len;
     _ = symbol_name_ptr;
     const comp = @intToPtr(*Compilation, stage1.userdata);
-    const lib_name = std.ascii.allocLowerString(comp.gpa, lib_name_ptr[0..lib_name_len]) catch return "out of memory";
+    const lib_name = lib_name_ptr[0..lib_name_len];
     const target = comp.getTarget();
     const is_libc = target_util.is_libc_lib_name(target, lib_name);
     if (is_libc) {


vincenzopalazzo (approved):
LGTM

jibal:
And the previous code was a memory leak, yes?

viriuwu:
Not exactly, the function is made to be used from C++, you can see it returns a mix of allocated and literal strings.

jibal:
I don't understand. std.ascii.allocLowerString says "Caller owns returned string and must free with `allocator`", but that was not being done. And if it was freed somewhere, then this change would result in a free of memory that wasn't allocated, which is why I took a look at it.

src/stage1.zig
```diff
@@ -416,7 +416,7 @@ export fn stage2_add_link_lib(
     _ = symbol_name_len;
     _ = symbol_name_ptr;
     const comp = @intToPtr(*Compilation, stage1.userdata);
-    const lib_name = std.ascii.allocLowerString(comp.gpa, lib_name_ptr[0..lib_name_len]) catch return "out of memory";
+    const lib_name = lib_name_ptr[0..lib_name_len];
```
Vexu, line 419: This should be duped but since you shouldn't be using stage1 with `--watch` and stage1 is getting deleted soon anyways it's probably fine.

marler8997:
It looks like this change broke the zigwin32 bindings when cross compiling on hosts with case-sensitive filesystems. We now get linker errors like this:
```
error(link): DLL import library for -lKERNEL32 not found
```
The Win32Metadata project appears to use upper case library names, which zigwin32 propagates unchanged.  The library files are generated by zig on the fly from the mingw .def files.  These all appear to be lowercase which is why this change broke the zigwin32 bindings.

It seems to me that this change is an improvement overall, I can see how this behavior of always lowercasing library names could be very unexpected and drive a person mad wondering why their library isn't being found.  So the solution for zigwin32 might just be to modify all library names to lowercase since that's what zig is going to provide anyway.  If anyone has a better suggestion or knows any reason why this won't work please chime in.  Thanks.

viriuwu:
@marler8997 I recommend going with the solution of changing the library names to lowercase. I had someone report a similar issue (not Zig, but another library I was maintaining) where I put `Kernel32` (as written in MSDN) and indeed, it broke cross compiling. The solution was just to use lowercase, as they're all lowercase, and that fixed it.

PR #11688, opened 2022-05-21
document bufPrint, and format now uses `writer` not `output`

dominictarr:
just adding documentation where I needed to read the code. It was not clear from the types that bufPrint would return a slice of the bytes printed to.

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -24,9 +24,9 @@ pub const FormatOptions = struct {
     fill: u8 = ' ',
 };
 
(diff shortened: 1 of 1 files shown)


andrewrk:
Thanks for the patch!

To fix the CI error, run zig fmt on the changed files. For example:
```
zig fmt lib/std
```

dominictarr:
sorry, done.

dominictarr:
I think i figured it out

vincenzopalazzo (approved):

PR #11695, opened 2022-05-22
math: make `cast` return an optional instead of error

alichraghi:


kubkon:
What is the motivation for this proposed change?

alichraghi:
https://github.com/ziglang/zig/blob/71e2a56e3ef7aba10cc0648aab786973cf8416bc/lib/std/math.zig#L993-L994

kubkon:
Mhm, I see, and any clue what the motivation for this TODO was? Asking out of curiosity as this is a rather big change and breaking :-)

alichraghi:
i think since it just return a single error value it make more sense to use an optional than error union. this can be seen abundantly in other parts of stdlib as well

andrewrk (approved):
Thanks @alichraghi - would you mind rebasing and fixing the conflicts?

Techcable:
> i think since it just return a single error value it make more sense to use an optional than error union. this can be seen abundantly in other parts of stdlib as well

My concern here is that this seems inconsistent with the checked multiplication/addition `std.math.mul` and `std.math.add`, which return `error.Overflow` instead of an optional types.

Is there some specific reason why `std.math.cast()` is different from checked multiplication/addition functions, or Is there a plan to change those as well?

PR #11696, opened 2022-05-22
The first `std.mem.SplitIterator(T).next()` result is always non-`null`

r00ster91:
I observed that the first `next()` call on an `std.mem.SplitIterator(T)` is always non-`null`.
I explicitly mentioned that in the docs and changed many `next()` calls to `next().?` in various places.
These `next().?`s will never panic.

### DIFF
--- a/lib/std/crypto/phc_encoding.zig
+++ b/lib/std/crypto/phc_encoding.zig
@@ -253,7 +253,7 @@ fn serializeTo(params: anytype, out: anytype) !void {
 // Split a `key=value` string into `key` and `value`
 fn kvSplit(str: []const u8) !struct { key: []const u8, value: []const u8 } {
     var it = mem.split(u8, str, kv_delimiter);
-    const key = it.next() orelse return Error.InvalidEncoding;
+    const key = it.next().?;
     const value = it.next() orelse return Error.InvalidEncoding;
     const ret = .{ .key = key, .value = value };
     return ret;
--- a/lib/std/crypto/scrypt.zig
+++ b/lib/std/crypto/scrypt.zig
@@ -289,7 +289,7 @@ const crypt_format = struct {
 
         var it = mem.split(u8, str[14..], "$");
 
-        const salt = it.next() orelse return EncodingError.InvalidEncoding;
+        const salt = it.next().?;
         if (@hasField(T, "salt")) out.salt = salt;
 
         const hash_str = it.next() orelse return EncodingError.InvalidEncoding;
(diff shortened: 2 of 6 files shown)


marler8997:
An idea comes to mind, may be good or bad.  What if we added a method to `SplitIterator` called `first` that returned a `[]const T` instead of `?[]const T`?  We might even be able to add debug-checks that first isn't called if we're not at the beginning?

r00ster91:
> An idea comes to mind, may be good or bad. What if we added a method to `SplitIterator` called `first` that returned a `[]const T` instead of `?[]const T`? We might even be able to add debug-checks that first isn't called if we're not at the beginning?

@marler8997 Not a bad idea! But personally I think it might add more complexity for maybe not too much gain because whether `first()` was actually called first will also have to be checked then.
To make sure everyone knows, I will add another note about this on `next()`'s documentation too.
But still it might be worth opening an issue for further discussion about this.

Vexu:
+1 on a separate `first` function. Changing the usage like this means that if the implementation ever changes a bunch of code will break without warning.

Vexu:
Closing as stale with test not passing and feedback unaddressed.

r00ster91:
@Vexu I wasn't aware that this PR was regarded as stale. The tests were in my mind passing because if you look at the "failure" I can't see anything (https://cloud.drone.io/ziglang/zig/18040/1/6) so I don't know what went wrong there. Seemed unrelated to me.
> 2094 passed; 59 skipped; 0 failed.

As for the feedback, I acknowledged it but had nothing to add because I didn't plan on doing it in this PR. I thought it could be done in a follow-up. Now I'm assuming doing it using `first` is a requirement.
I'm still not sure about adding it though. If `first` is not called first, should that be debug-checked in `next`?

Vexu:
> Vexu I wasn't aware that this PR was regarded as stale. The tests were in my mind passing because if you look at the "failure" I can't see anything (https://cloud.drone.io/ziglang/zig/18040/1/6) so I don't know what went wrong there. Seemed unrelated to me.

It may not have been this PR's fault but it'd still be better to try to rebase to a commit that isn't failing.

> As for the feedback, I acknowledged it but had nothing to add because I didn't plan on doing it in this PR. I thought it could be done in a follow-up.

I would really prefer small changes like this be done in one PR instead of being split into multiple with the latter ones being forgotten.

> Now I'm assuming doing it using `first` is a requirement. I'm still not sure about adding it though. If `first` is not called first, should that be debug-checked in `next`?

No, I think `first` should be available in addition to `next` so that only using `next` would still work. There should be a safety check in `start` to ensure it is not called multiple times.

r00ster91:
Alright, understandable!

I might open a PR later.

PR #11701, opened 2022-05-23
std.child_process: enable using non-standard pipes

matu3ba:
The user provides an array of struct with info for direction for the pipe.
Each struct also contains the space for the input and output end of pipe,
which simplifies handling.
Typically a low number of non-standard pipes are used, so no premature optimization
and complicating the usage.
Also, it is expected that the upper bound of used pipes per process is comptime-known,
so slicing a continuous chunk would be most efficient.

* provide user with function pointer to enable parsing of necessary file descriptor
  on Linux and file handles on Windows in the child process
  - They can not be comptime-set, because the Kernel or user expects the program to
    work in all cases.
* keep logic the same as for posix_spawn
* use fcntl and SetHandleInformation to prevent file leakage into
  subsequently created children
* test if both fcntl and HandleInformation were correctly set
* test non-standard stream with 1 pipe type

Note: Standard streams are expected to be inheritable by child processes, so those
streams are "intentionally leaked to the child".

Benchmarking with realistic workloads is needed to decide the final design.
Especially, since storing space for one additionally unused pipe end can be wasteful.


matu3ba:
The current problem is
```txt
2022-05-24T12:38:44.1203691Z 78/1518 child_process.test "std-native-Debug-bare-multi-default ChildProcess with extra streams"... .\build\dist\lib\zig\std\fmt.zig:1773:16: error: expected type '*anyopaque', found 'comptime_int'
2022-05-24T12:38:44.1205926Z     var x: T = 0;
2022-05-24T12:38:44.1206578Z                ^
2022-05-24T12:38:44.1207320Z .\build\dist\lib\zig\std\fmt.zig:1676:44: note: called from here
2022-05-24T12:38:44.1208158Z     if (buf[0] == '+') return parseWithSign(T, buf[1..], radix, .Pos);
2022-05-24T12:38:44.1208954Z                                            ^
2022-05-24T12:38:44.1209706Z .\zig-cache\tmp\Vyog3M7h5VU0Lgcq\child.zig:9:44: note: called from here
2022-05-24T12:38:44.1210535Z      var file_handle = try std.fmt.parseInt(std.os.fd_t, stdin_cont, 11);
2022-05-24T12:38:44.1211314Z                                            ^
2022-05-24T12:38:44.1212309Z .\zig-cache\tmp\Vyog3M7h5VU0Lgcq\child.zig:3:22: note: called from here
2022-05-24T12:38:44.1213052Z  pub fn main() !void {
2022-05-24T12:38:44.1213768Z                      ^
2022-05-24T12:38:44.1214541Z .\build\dist\lib\zig\std\fmt.zig:1773:16: error: expected type '*anyopaque', found 'comptime_int'
2022-05-24T12:38:44.1215450Z     var x: T = 0;
2022-05-24T12:38:44.1216056Z                ^
2022-05-24T12:38:44.1216742Z .\build\dist\lib\zig\std\fmt.zig:1677:44: note: called from here
2022-05-24T12:38:44.1217547Z     if (buf[0] == '-') return parseWithSign(T, buf[1..], radix, .Neg);
2022-05-24T12:38:44.1218248Z                                            ^
2022-05-24T12:38:44.1218973Z .\zig-cache\tmp\Vyog3M7h5VU0Lgcq\child.zig:9:44: note: called from here
2022-05-24T12:38:44.1219817Z      var file_handle = try std.fmt.parseInt(std.os.fd_t, stdin_cont, 11);
2022-05-24T12:38:44.1220518Z                                            ^
2022-05-24T12:38:44.1221309Z .\zig-cache\tmp\Vyog3M7h5VU0Lgcq\child.zig:3:22: note: called from here
2022-05-24T12:38:44.1222077Z  pub fn main() !void {
2022-05-24T12:38:44.1222704Z                      ^
2022-05-24T12:38:44.1223488Z .\build\dist\lib\zig\std\fmt.zig:1773:16: error: expected type '*anyopaque', found 'comptime_int'
2022-05-24T12:38:44.1224277Z     var x: T = 0;
2022-05-24T12:38:44.1224861Z                ^
```
due to Zig defining HANDLE as
```zig
pub const HANDLE = *anyopaque;
```

matu3ba:
Solution is to add those https://comp.os.ms-windows.programmer.win32.narkive.com/a4d3CdvQ/convert-handle-to-file-descriptor-works-in-vc-not-in-cygwin
and while we are it these very nice functions to get the type of an underlying file descriptor:
https://jakash3.wordpress.com/2012/12/19/getting-file-descriptor-type-in-windows/

Reminder for me to implement eventually
https://www.youtube.com/watch?v=uhRWMGBjlO8&t=621s
https://github.com/haskell/directory/issues/110
as well.

matu3ba:
`const all_extra_str_in = try extra_str_in_rd.readUntilDelimiterAlloc(alloc, '\x17', 20_000);`
fails in
`return windows.ReadFile(self.handle, buffer, null, self.intended_io_mode);`
with
```txt
 error.Unexpected: GetLastError(5): Access is denied.
2022-05-25T23:16:07.2761173Z 
2022-05-25T23:16:07.2762589Z error.Unexpected: GetLastError(6): The handle is invalid.
```
I'll look into permission handling another day. I think I need to additionally register all pipes with the security attribute.

Not really sure why `"behavior-native-Debug-bare-multi-default coerce between pointers of compatible differently-named floats"` fails every time and only in Windows.

lib/std/child_process.zig
```diff
@@ -491,6 +538,21 @@ pub const ChildProcess = struct {
...
+                    if (extra.*.output) |*outpipe| {
+                        outpipe.close();
+                        extra.*.output = null;
+                    }
+                } else {
+                    if (extra.*.input) |*inpipe| {
+                        inpipe.close();
+                        extra.*.input = null;
+                    }
+                }
```
daurnimator, line 553: No need for the `.*` most of the time.
Also you can use a `switch` here.
```suggestion
                switch (extra.direction) {
                    .parent_to_child => if (extra.output) |*outpipe| {
                        outpipe.close();
                        extra.output = null;
                    },
                    .child_to_parent => if (extra.input) |*inpipe| {
                        inpipe.close();
                        extra.input = null;
                    },
                }
```

lib/std/child_process.zig
```diff
@@ -127,25 +163,33 @@ pub const ChildProcess = struct {
         self.gid = user_info.gid;
     }
 
+    const ExPipeInfoProto = switch (builtin.zig_backend) {
```
daurnimator, line 166: I don't think this is a great type name

lib/std/child_process.zig
```diff
@@ -605,24 +713,14 @@ pub const ChildProcess = struct {
...
-        if (self.stdout_behavior == StdIo.Pipe) {
-            self.stdout = File{ .handle = stdout_pipe[0] };
-        } else {
-            self.stdout = null;
-        }
-        if (self.stderr_behavior == StdIo.Pipe) {
-            self.stderr = File{ .handle = stderr_pipe[0] };
-        } else {
-            self.stderr = null;
+        // user must communicate extra pipes to child process either via
```
daurnimator, line 716: Or just convention. e.g. `SD_LISTEN_FDS_START` is compile-time known to be `3`

lib/std/child_process.zig
```diff
@@ -127,25 +163,33 @@ pub const ChildProcess = struct {
...
+        // workaround until we replace stage1 with stage2
+        .stage1 => ?fn (self: *ChildProcess) ChildProcess.SpawnError!void,
+        else => ?*const fn (self: *ChildProcess) ChildProcess.SpawnError!void,
+    };
+
     /// On success must call `kill` or `wait`.
-    pub fn spawn(self: *ChildProcess) SpawnError!void {
+    /// Use null xor function pointer to tell child process about extra pipes
+    /// See field extra_streams for more information
+    pub fn spawn(self: *ChildProcess, pipe_info_fn: ExPipeInfoProto) SpawnError!void {
```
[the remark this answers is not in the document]
matu3ba: I updated the code to use a struct with optional function pointer as field.

lib/std/child_process.zig
matu3ba: I used switches and `.` instead of `.*.` everywhere to increase readability and make future regressions less likely.

andrewrk:
I have attempted to rebase this against master branch on your behalf in order to re-run the failing CI checks. However there are conflicts. If you would like to continue working on this, please open a new PR against master branch.

PR #11703, opened 2022-05-23
Fix indexing in the test runner's log formatting

LewisGaul:
Without this fix the following fails due to the index of the args being changed by the test runner log function.

```zig
const std = @import("std");
test {
    std.log.warn("{[0]d}, 0x{[0]X}", .{123});
}
```


LewisGaul:
The failing CI logs shows all tests passing (`2094 passed; 59 skipped; 0 failed.`) but "Step is killed (canceled)" - I don't think this is an issue with the PR.

PR #10755, opened 2022-02-01
std: adding RomuTrio RNG to Zig

francescoalemanno:
this is an implementation of [Romu-rng](https://www.romu-random.org/romupaper.pdf).

I tested this RNG in some research code of mine and it improves performance substantially without compromising results (while I noticed some biases with the smallest Xoroshiro).

There is some controversy regarding this RNG family (mostly theoretical arguments) but empirically this RNG is proven to be exceptionally good.

Francesco

------- added buffered benchmark --

```zig
on Mac M1 Air 2020

-------- TESTING BUF [64]u8
rng:        Romu192    time minimum 6.9940 ms;  time maximum 7.3120 ms;
rng:     Xoshiro256    time minimum 8.3250 ms;  time maximum 8.5200 ms;
rng:          Sfc64    time minimum 13.3090 ms;  time maximum 13.7180 ms;
rng:   Xoroshiro128    time minimum 12.6660 ms;  time maximum 12.7310 ms;
rng:            Pcg    time minimum 16.6220 ms;  time maximum 16.6960 ms;
-------- TESTING BUF [128]u8
rng:        Romu192    time minimum 13.3800 ms;  time maximum 13.7570 ms;
rng:     Xoshiro256    time minimum 17.8500 ms;  time maximum 18.3760 ms;
rng:          Sfc64    time minimum 16.3480 ms;  time maximum 16.7290 ms;
rng:   Xoroshiro128    time minimum 25.0390 ms;  time maximum 25.5200 ms;
rng:            Pcg    time minimum 33.3030 ms;  time maximum 33.7300 ms;
-------- TESTING BUF [256]u8
rng:        Romu192    time minimum 26.8540 ms;  time maximum 27.0120 ms;
rng:     Xoshiro256    time minimum 35.7290 ms;  time maximum 36.5140 ms;
rng:          Sfc64    time minimum 32.0140 ms;  time maximum 32.3860 ms;
rng:   Xoroshiro128    time minimum 50.0900 ms;  time maximum 50.6170 ms;
rng:            Pcg    time minimum 66.7040 ms;  time maximum 67.0500 ms;
-------- TESTING BUF [512]u8
rng:        Romu192    time minimum 53.8040 ms;  time maximum 54.4070 ms;
rng:     Xoshiro256    time minimum 71.6470 ms;  time maximum 72.2740 ms;
rng:          Sfc64    time minimum 95.8960 ms;  time maximum 96.4480 ms;
[...]

PR #11707, opened 2022-05-24
std.rand: Add RomuTrio

ominitay:
Following on from #10755.
Going to see if tests pass, I think the test case was broken rather than the actual RNG implementations.


ominitay:
Tests pass now, should be suitable for merging! @andrewrk

andrewrk:
Thanks @ominitay!

PR #11708, opened 2022-05-24
aarch64: check lo/cc flag for unsigned sub_with_overflow

kubkon:
With this change, we are now correctly lowering `sub_with_overflow`
for signed and unsigned integers of register-sized integers (32-
or 64-bit precisely). We also match LLVM's behavior and so, the
condition flags we now set are:

* unsigned:
  - `add_with_overflow`: `hs`/`cs` (carry set)
  - `sub_with_overflow`: `lo`/`cc` (carry clear)
* signed:
  - `add_with_overflow`/`sub_with_overflow`: `vs` (overflow)


joachimschmidt557 (approved):
LGTM!

PR #11725, opened 2022-05-26
std: compile error on invalid testing allocator usage

### LINKED ISSUE #11721: make std.testing.allocator be a compile error when used outside of `zig test`
Inspiration: this question on IRC:

```
<ryaf> I am trying to use an AutoHashMap as a set:
<ryaf> pub fn main() anyerror!void {
<ryaf>     var mymap = std.AutoHashMap(i32, void).init(std.testing.allocator);
<ryaf>     try mymap.put(1, {});
<ryaf> }
<ryaf> The above works with:
<ryaf> zig build run
<ryaf> However, if I put it in a test:
<ryaf> test "mytest" {
<ryaf>     var mymap = std.AutoHashMap(i32, void).init(std.testing.allocator);
<ryaf>     try mymap.put(1, {});
<ryaf> }
<ryaf> I get a memory leak error with:
<ryaf> zig build test
<ryaf> Am I using AutoHashMap incorrectly?
```

The first code snippet should ideally have been a compile error and instructed the person to use a different allocator than `std.testing.allocator`, or to use `zig test` combined with `test` blocks.

alichraghi:
Closes #11721

### DIFF
--- a/lib/std/testing.zig
+++ b/lib/std/testing.zig
@@ -8,7 +8,11 @@ pub const FailingAllocator = @import("testing/failing_allocator.zig").FailingAll
 
 /// This should only be used in temporary test programs.
 pub const allocator = allocator_instance.allocator();
-pub var allocator_instance = std.heap.GeneralPurposeAllocator(.{}){};
+pub var allocator_instance = b: {
+    if (!builtin.is_test)
+        @compileError("Cannot use testing allocator outside of test block");
+    break :b std.heap.GeneralPurposeAllocator(.{}){};
+};
 
 pub const failing_allocator = failing_allocator_instance.allocator();
 pub var failing_allocator_instance = FailingAllocator.init(base_allocator_instance.allocator(), 0);
--- a/test/cli.zig
+++ b/test/cli.zig
@@ -8,7 +8,9 @@ const ChildProcess = std.ChildProcess;
 var a: std.mem.Allocator = undefined;
 
 pub fn main() !void {
-    var arena = std.heap.ArenaAllocator.init(std.testing.allocator);
+    var gpa = std.heap.GeneralPurposeAllocator(.{}){};
+    defer _ = gpa.deinit();
+    var arena = std.heap.ArenaAllocator.init(gpa.allocator());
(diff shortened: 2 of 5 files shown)


matu3ba:
What if the user does define the tests inside `build.zig` ?
Can and should we enforce a convention such that its simpler to find the allocator definition and thus to identify what are tests?

One possible use case: Check used libraries for leaks with given input data/execution sequence(s).
See #11472.

alichraghi:
this test fails. there's two possible solutions:
- use another allocator here
- find a way to detect parent process is a test

https://github.com/ziglang/zig/blob/01607b54fc21c5dfc0ee8bc21ef317c8c03868c2/lib/std/child_process.zig#L1342-L1370

matu3ba:
@alichraghi The code got removed in https://github.com/ziglang/zig/commit/67d5bfefba48d28c02e2841f1a47a213d28d4693 and I think this it is the right call to expect the user to deinit the allocators for leak checks.

The only thing that I am currently still very unhappy about is that there is no nice way to attach a custom user action as runstep to data from build.zig, but there was positive feedback so it will be done/outlined eventually.

Rebasing against master should fix the errors.

ehaas:
In my opinion `std.testing.failing_allocator` should not be disabled outside of test blocks - it's useful if you have a data structure which requires an allocator, but you want a runtime guarantee that it will not allocate. (Disabling `std.testing.allocator` outside of test blocks is fine)

A concrete example is the Aro test runner - it's part of test code, but it's run as a standalone program and not as part of a test block, so this change would prevent it from compiling: https://github.com/Vexu/arocc/blob/master/test/runner.zig#L21=

The context is that we're re-using the generic command-line parsing code, and in general that code can add items to the list of sources to be compiled. But when that function is used from a test, we know exactly what sources we want to compile, and it's nice to have a runtime assertion that we can't accidentally add other sources to the list.

alichraghi:
also error backtrace doesn't fully showen. e.g [CI](https://ci.ziglang.org/ziglang/zig/5018/1/2)

PR #11728, opened 2022-05-26
stage2 AArch64: complete genTypedValue

joachimschmidt557:
cc @kubkon

### DIFF
--- a/src/arch/aarch64/CodeGen.zig
+++ b/src/arch/aarch64/CodeGen.zig
@@ -4572,6 +4572,7 @@ fn lowerUnnamedConst(self: *Self, tv: TypedValue) InnerError!MCValue {
 }
 
 fn genTypedValue(self: *Self, typed_value: TypedValue) InnerError!MCValue {
+    log.debug("genTypedValue: ty = {}, val = {}", .{ typed_value.ty.fmtDebug(), typed_value.val.fmtDebug() });
     if (typed_value.val.isUndef())
         return MCValue{ .undef = {} };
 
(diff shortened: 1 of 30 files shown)


src/arch/aarch64/CodeGen.zig
```diff
@@ -4695,11 +4684,22 @@ fn genTypedValue(self: *Self, typed_value: TypedValue) InnerError!MCValue {
...
-        .Struct => {
-            return self.lowerUnnamedConst(typed_value);
-        },
-        else => return self.fail("TODO implement const of type '{}'", .{typed_value.ty.fmtDebug()}),
+
+        .ComptimeInt => unreachable, // semantic analysis prevents this
+        .ComptimeFloat => unreachable, // semantic analysis prevents this
+        .Type => unreachable,
+        .EnumLiteral => unreachable,
+        .Void => unreachable,
```
kubkon, line 4692: FYI https://github.com/ziglang/zig/blob/b08d32ceb5aac5b1ba73c84449c6afee630710bb/src/arch/x86_64/CodeGen.zig#L6766 This will resurface when working with error unions and implicit `return 0;` for `anyerror!void` return types AFAIK.

test/behavior/bitreverse.zig
```diff
@@ -158,7 +158,6 @@ test "bitReverse vectors u0" {
     if (builtin.zig_backend == .stage2_c) return error.SkipZigTest;
     if (builtin.zig_backend == .stage2_x86_64) return error.SkipZigTest;
     if (builtin.zig_backend == .stage2_arm) return error.SkipZigTest;
```
kubkon, line 160: I wonder, do x64 and arm backends pass this as well? If not, any clue as to why aarch64 does without having SIMD support?

joachimschmidt557: Yes, they pass this test as well. I assume it's due to Sema knowing at comptime that the only possible of an `u0` is `0`.

src/arch/aarch64/CodeGen.zig
joachimschmidt557: Ok, should I change it now or later?

kubkon: Up to you, just letting you know what it currently looks like :-)

test/behavior/bitreverse.zig
kubkon: Mhm, makes sense!

PR #11731, opened 2022-05-26
std.PriorityQueue: fix missing siftUp in remove

joachimschmidt557:
When the replacement node is smaller than its parent, we need to sift
up instead of sifting down.


andrewrk:
Thanks for the new test case :+1:

PR #11733, opened 2022-05-26
more misc stage2 fixes

### LINKED ISSUE #11135: Stage2 result location cannot handle store to inferred from nested switch
```zig
test {
    var a: error{ A, B }!u64 = 0;
    var b = a catch |err| switch (err) {
        error.A => 0,
        else => unreachable,
    };
    _ = b;
}
```
`0` gets emitted to AIR as `comptime_int` which then causes `coerce` to hit unreachable when a value of `comptime_int` is not comptime known.

Happens in `dwarf.zig` when calling `std.debug.dumpCurrentStackTrace` on linux but is easily avoided with a simple cast.
 https://github.com/ziglang/zig/blob/e5d4a694ea7dd251e10d6434c9321b5e0a548d4b/lib/std/dwarf.zig#L825

### LINKED ISSUE #11179: stage2: comptime-known value isn't comptime-known through generic function
### Zig Version

0.10.0-dev.1335+c64279b15

### Steps to Reproduce

The test program is below: Note:

* The `switch` return type is required for the function sig.
* The `switch` return value is required for the function body.
* The function call is required
* If you wrap the whole test in a `comptime {}` block it works.

```zig
const std = @import("std");
const Type = std.builtin.Type;

test "Tuple" {
    const fields_list = fields(@TypeOf(.{}));
    if (fields_list.len != 0)
        @compileError("Argument count mismatch");
}

pub fn fields(comptime T: type) switch (@typeInfo(T)) {
    .Struct => []const Type.StructField,
    else => unreachable,
} {
    return switch (@typeInfo(T)) {
        .Struct => |info| info.fields,
        else => unreachable,
    };
}
```

### Expected Behavior

Should not have any errors (stage1 works), since `fields.len` should be comptime known.

The reason: this is how `std.meta` tests `Tuple` and `ArgsTuple`, so this is exercised in the stdlib.

### Actual Behavior

Triggers `@compileError`

### LINKED ISSUE #11543: stage2: Array type with child type being a function crashes the compiler
### Zig Version

0.10.0-dev.1983+c5e847744

### Steps to Reproduce

```
// test.zig
pub fn S() type {
    return struct {};
}

test "" {
    _ = [0]S;
}
```

```
$ stage2/zig test test.zig
```

### Expected Behavior

A compile error along the lines of `error: expected type, found fn() type`

### Actual Behavior

```
$ stage2/zig test test.zig
thread 226457 panic: zig compiler bug: NeededSourceLocation
...
zig/src/Module.zig:4923:39: 0x2faac63 in Module.analyzeFnBody (zig)
        error.NeededSourceLocation => @panic("zig compiler bug: NeededSourceLocation"),
...
```

### LINKED ISSUE #11632: missing compile error for runtime iteration over compile-time-only array elements
Zig Version: `0.10.0-dev.2159+f5edf78ee`

```zig
test "functions in an array" {
    const TestFn = fn () void;
    const test_fns = [_]TestFn{ foo, bar };
    for (test_fns) |testFn| {
        testFn();
    }
}

fn foo() void {}
fn bar() void {}
```

```
$ ./stage2/bin/zig test test3.zig 
thread 2885489 panic: Segmentation fault at address 0x0
???:?:?: 0x3eb6328 in ??? (???)

Panicked during a panic: integer overflow
Inner panic stack:
/home/andy/Downloads/zig/lib/std/debug.zig:169:69: 0x3314122 in std.debug.dumpStackTraceFromBase (zig)
                                                                    ^
/home/andy/Downloads/zig/src/crash_report.zig:327:45: 0x31af4a2 in crash_report.StackContext.dumpStackTrace (zig)
                debug.dumpStackTraceFromBase(ex.bp, ex.ip);
                                            ^
/home/andy/Downloads/zig/src/crash_report.zig:479:39: 0x31af050 in crash_report.PanicSwitch.reportStack (zig)
        state.panic_ctx.dumpStackTrace();
                                      ^
/home/andy/Downloads/zig/src/crash_report.zig:553:9: 0x2fc80e0 in crash_report.PanicSwitch.initPanic (zig)
        @call(.{}, func, args);
        ^
/home/andy/Downloads/zig/src/crash_report.zig:553:9: 0x2cc1845 in crash_report.PanicSwitch.dispatch (zig)
        @call(.{}, func, args);
        ^
/home/andy/Downloads/zig/src/crash_report.zig:254:25: 0x2fc9577 in crash_report.handleSegfaultPosix (zig)
    PanicSwitch.dispatch(null, stack_ctx, error_msg);
                        ^
Aborted (core dumped)
```

This is an important case to get right because it's code that is valid for stage1 which now has different function type semantics in stage2. It should be resolved in one of two ways:

 * A compile error explaining that `testFn` is a runtime-known value and so its type must not be a compile-time-only type. A hint should say that `inline for` would solve this problem. A hint should also say that function bodies are comptime-only while function pointers are OK for runtime, and that `&` can be used to convert a function body to its pointer.
 * Or it should implicitly cause the `for` to become an `inline for`.

For the former, you can see stage1 already handles a similar case pretty well:

```zig
test "functions in an array" {
    const types = [_]type{ i32, bool };
    for (types) |T| {
        _ = T;
    }
}
```

```
$ ./stage1/bin/zig test test3.zig 
./test3.zig:6:5: error: values of type 'type' must be comptime known, but index value is runtime known
    for (types) |T| {
    ^
```

Vexu:
Closes #11135
Closes #11543
Closes #11632
Closes #11179

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -2881,6 +2881,28 @@ fn zirResolveInferredAlloc(sema: *Sema, block: *Block, inst: Zir.Inst.Index) Com
 
             if (var_is_mut) {
                 try sema.validateVarType(block, ty_src, final_elem_ty, false);
+
+                // The value might have been bitcasted into a comptime only
+                // pointer type such as `*@Type(.EnumLiteral)` so we must now
+                // update all the stores to not give backends invalid AIR.
+
+                var air_tags = sema.air_instructions.items(.tag);
+                var air_data = sema.air_instructions.items(.data);
+                var peer_inst_index: usize = 0;
+                var i = ptr_inst;
+                while (i < air_tags.len and peer_inst_index < peer_inst_list.len) : (i += 1) {
+                    if (air_tags[i] != .store) continue;
+                    if (air_data[i].bin_op.rhs == peer_inst_list[peer_inst_index]) {
+                        peer_inst_index += 1;
+                        _ = (try sema.resolveMaybeUndefVal(block, .unneeded, air_data[i].bin_op.rhs)) orelse continue;
+                        const coerced_val = try sema.coerce(block, final_elem_ty, air_data[i].bin_op.rhs, .unneeded);
+                        air_tags = sema.air_instructions.items(.tag);
+                        air_data = sema.air_instructions.items(.data);
+
+                        air_data[i].bin_op.lhs = ptr;
+                        air_data[i].bin_op.rhs = coerced_val;
+                    }
+                }
             } else ct: {
                 // Detect if the value is comptime known. In such case, the
                 // last 3 AIR instructions of the block will look like this:
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -5814,8 +5814,10 @@ fn zirArrayType(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!A
     defer tracy.end();
 
     const bin_inst = sema.code.instructions.items(.data)[inst].bin;
-    const len = try sema.resolveInt(block, .unneeded, bin_inst.lhs, Type.usize);
-    const elem_type = try sema.resolveType(block, .unneeded, bin_inst.rhs);
+    const len_src = sema.src; // TODO better source location
+    const elem_src = sema.src; // TODO better source location
+    const len = try sema.resolveInt(block, len_src, bin_inst.lhs, Type.usize);
+    const elem_type = try sema.resolveType(block, elem_src, bin_inst.rhs);
     const array_ty = try Type.array(sema.arena, len, null, elem_type, sema.mod);
 
     return sema.addType(array_ty);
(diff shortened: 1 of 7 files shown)


src/Sema.zig
```diff
@@ -18471,6 +18497,25 @@ fn elemValArray(
         }
     }
 
+    const valid_rt = try sema.validateRunTimeType(block, elem_index_src, elem_ty, false);
+    if (!valid_rt) {
+        const msg = msg: {
+            const msg = try sema.errMsg(
+                block,
+                elem_index_src,
+                "values of type '{}' must be comptime known, but index value is runtime known ",
```
andrewrk, line 18506: Found the CI error:

```suggestion
                "values of type '{}' must be comptime known, but index value is runtime known",
```

Issue #9646, opened 2021-08-29
Comptime Memory and Type Punning Rules


SpexGuy:
This issue contains a description of how memory and pointers should work when aliased at compile time.  As a task, this issue can be closed once stage 2 implements this behavior.

### Aggregate Types

Types in the following families are considered "Aggregate Types":

- struct
- union
- optional
- slice
- array
- vector
- error union

All other types are not Aggregate Types.
If a type is not an Aggregate Type, then it is a Primitive Type.

### Defined Representation

The following types have Defined Representation:

- any integer (except comptime_int)
- any float (except comptime_float)
- bool
- void (technically allowed but empty)
- enum with specified integer type
- any pointer or non-generic fn (but not slice)
- extern struct/union
- packed struct/union
- array of any these types
- vector of any of these types

For any given target, instances of these types have a well-defined layout in memory.  They can be passed between compilation units safely, without fear of misinterpretation.  Additionally, their layout is known and well-defined at compile time.  This list may look familiar.  It's also the list of values which are allowed in an `extern struct`.

The following types *do not* have Defined Representation:

- comptime_int, comptime_float
- noreturn, (null), (undefined)
- slices, optionals, error unions
- error sets
- bare struct/union (no extern or packed)
- enum without specified integer type (TODO maybe defined?)
- type, generic fn
- any frame, anyframe
- enum literals

### Top Level Variable

A "Top Level Variable" is a variable which is not part of something larger.  In compiler parlance, it has its own provenance.  The following are top level variables:

- global const
- global var
- local const
- local var

Everything else is not a top level variable:

- member of an Aggregate Type

## Memory Islands

As described in #7770, memory at compile time is split up into separate "islands".  A pointer to one island can only ever point to data in that same island.  There are two types of memory islands: Literal and Virtual.  Virtual memory islands may have child islands, which can be Literal or Virtual.  Literal memory islands have Defined Representation.

Memory Islands are arranged into trees.  Each Top Level Variable creates one tree based on its type.  The Memory Island Tree for a type is constructed as follows:

If the type has Defined Representation, use a Literal memory island matching the size of the type.
Otherwise, use a Virtual memory island.  If the type is an Aggregate Type, add child islands for each member using this algorithm.

Let's look at some examples to make this clearer:

```zig
const Primitive = u32;
// Primitive instances have one Literal island, in this case consisting of four bytes.

const Ptr = *u32;
// Pointers are also primitives, this type has one Literal island of 8 bytes.

[...]

Issue #2471, opened 2019-05-11
`comptime` keyword used with runtime-known identifier fails to produce a compile error


ghost:
```zig
fn hasField() bool {
    return false;
}

test "" {
    var x: u32 = 123; // runtime value

    // what i'm trying to do:
    if (comptime hasField()) {
      if (x == 123) {
        @compileError("A"); // this won't happen
      }
    }

    // this works as above, although it's weird and ambiguous to read
    if (comptime hasField() and x == 123) {
        @compileError("B"); // this won't happen
    }

    // this does not work as above. `comptime` seems to be ignored.
    if (x == 123 and comptime hasField()) {
        @compileError("C"); // this happens
    }
}
```

Output:
```
/home/dbandstra/zang/test.zig:22:9: error: C
        @compileError("C"); // this happens
        ^
```

Disclaimer: I don't actually know how this `comptime` thing works. I don't think it's even documented? Is it an official feature? (I find it useful.)

In the above test, I'm not sure what I expect to happen. In the first condition, `comptime hasField() and x == 123`, I don't whether `comptime` is supposed to apply to the first term or to the whole expression.

But in any case, I think there should be a compile error if a `comptime` is not going to be applied.

```zig
    var x: u32 = 123; // runtime value

    if (comptime x == 123) {
        std.debug.warn("why am i allowed to do this? what does it mean?\n");
    }
```

```zig
    var x: u32 = 123; // runtime value

    if (comptime x == 123 and comptime hasField()) {
        @compileError("huh?"); // this happens
    }
```

```
/home/dbandstra/zang/test.zig:9:20: error: unable to evaluate constant expression
    if (comptime x == 123 and comptime hasField()) {
                   ^
```

andrewrk:
Here's what's happening in each of your examples:

```zig
    if (comptime hasField()) {
```

`comptime` causes `hasField` to be evaluated at compile-time. Result is a compile-time known `bool`. The if is evaluated at compile-time.

```zig
    if (comptime hasField() and x == 123) {
```

Perhaps related to #114, this is parsed as `(comptime hasField()) and x == 123` and not `comptime (hasField() and x == 123)`. However because `and` does not evaluate the right-hand side if the left-hand side is comptime-known, the `x == 123` is never evaluated.

```zig
    if (x == 123 and comptime hasField()) {
```

[...]

PR #11734, opened 2022-05-26
Implement `writeToMemory`/`readFromMemory` for pointers

### LINKED ISSUE #11211: stage2: compiler panic when using `@ptrToInt` in bitwise AND
### Zig Version

0.10.0-dev.1393+291f5055f

### Steps to Reproduce

```zig
test "comptime ptrToInt in AND expression" {
    comptime var x: u8 = 15;
    try std.testing.expect(@ptrToInt(&x) & 0x00 == 0);
}
```

### Expected Behavior

Test should compile successfully and pass.

### Actual Behavior

The test fails because we end up calling `toBigInt` on a `decl_ref_mut`, which is not an integer type.

We're probably missing a dereference here for the indirect Value.
```
Analyzing testme.zig: testme.zig:test.comptime ptrToInt in AND expression
      %28 = decl_ref("std")
      %29 = field_ptr(%28, "testing")
      %30 = field_call_bind(%29, "expect")
      %31 = dbg_stmt(3, 27)
      %32 = ptr_to_int(%23)
      %33 = int(255)
    > %34 = bit_and(%32, %33)
      %35 = cmp_eq(%34, @Ref.zero)
      %36 = call(.auto, %30, [%35])
      %37 = is_non_err(%36)
      %38 = condbr(%37, {
        %40 = err_union_payload_unsafe(%36)
        %43 = break(%39, %40)
      }, {
        %41 = err_union_code(%36)
        %42 = ret_node(%41)
      })
      zig ast-check -t testme.zig

  in testme.zig: testme.zig:test.comptime ptrToInt in AND expression
    > %39 = block({%28..%38})

error(stdlib): metadata: 7fdb1aa25f68
/home/topolarity/repos/zig/src/value.zig:983:21: 0x3080801 in value.Value.toBigInt (zig)
            else => unreachable,
                    ^
/home/topolarity/repos/zig/src/value.zig:3007:40: 0x34dfe1b in value.Value.bitwiseAnd (zig)
        const lhs_bigint = lhs.toBigInt(&lhs_space);
                                       ^
/home/topolarity/repos/zig/src/Sema.zig:8222:51: 0x3323c31 in Sema.zirBitwise (zig)
                .bit_and => try lhs_val.bitwiseAnd(rhs_val, sema.arena),
                                                  ^
/home/topolarity/repos/zig/src/Sema.zig:645:65: 0x31e1c8a in Sema.analyzeBodyInner (zig)
            .bit_and                      => try sema.zirBitwise(block, inst, .bit_and),
                                                                ^
/home/topolarity/repos/zig/src/Sema.zig:3850:34: 0x35083c2 in Sema.resolveBlockBody (zig)
        if (sema.analyzeBodyInner(child_block, body)) |_| {
                                 ^
```

Vexu:
This is in direct contradiction to parts of #9646 and these two test cases https://github.com/ziglang/zig/blob/master/test/behavior/comptime_memory.zig#L319-L371.
As discussed in the self hosted compiler meetings the justification for this is change is that it seems like a lot of work for little value and making it an error also avoids the complications brought by comptime only types.

cc @SpexGuy 

Closes #11211

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -2527,7 +2527,7 @@ fn coerceResultPtr(
                 _ = try block.addBinOp(.store, new_ptr, null_inst);
                 return Air.Inst.Ref.void_value;
             }
-            return sema.bitCast(block, ptr_ty, new_ptr, src);
+            return sema.bitCast(block, ptr_ty, new_ptr, src, null);
         }
 
         const trash_inst = trash_block.instructions.pop();
@@ -2543,7 +2543,7 @@ fn coerceResultPtr(
                 if (try sema.resolveDefinedValue(block, src, new_ptr)) |ptr_val| {
                     new_ptr = try sema.addConstant(ptr_operand_ty, ptr_val);
                 } else {
-                    new_ptr = try sema.bitCast(block, ptr_operand_ty, new_ptr, src);
+                    new_ptr = try sema.bitCast(block, ptr_operand_ty, new_ptr, src, null);
                 }
             },
             .wrap_optional => {
@@ -9557,7 +9557,7 @@ fn zirBitcast(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air
         .Vector,
         => {},
     }
-    return sema.bitCast(block, dest_ty, operand, operand_src);
+    return sema.bitCast(block, dest_ty, operand, inst_data.src(), operand_src);
 }
 
 fn zirFloatCast(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.Inst.Ref {
@@ -9775,7 +9775,7 @@ fn zirSwitchCapture(
 
         switch (operand_ty.zigTypeTag()) {
             .ErrorSet => if (block.switch_else_err_ty) |some| {
-                return sema.bitCast(block, some, operand, operand_src);
+                return sema.bitCast(block, some, operand, operand_src, null);
             } else {
                 try block.addUnreachable(false);
                 return Air.Inst.Ref.unreachable_value;
@@ -9875,14 +9875,14 @@ fn zirSwitchCapture(
                 Module.ErrorSet.sortNames(&names);
                 const else_error_ty = try Type.Tag.error_set_merged.create(sema.arena, names);
 
-                return sema.bitCast(block, else_error_ty, operand, operand_src);
+                return sema.bitCast(block, else_error_ty, operand, operand_src, null);
             } else {
                 const item_ref = try sema.resolveInst(items[0]);
                 // Previous switch validation ensured this will succeed
                 const item_val = sema.resolveConstValue(block, .unneeded, item_ref, "") catch unreachable;
 
                 const item_ty = try Type.Tag.error_set_single.create(sema.arena, item_val.getError().?);
-                return sema.bitCast(block, item_ty, operand, operand_src);
+                return sema.bitCast(block, item_ty, operand, operand_src, null);
             }
         },
         else => {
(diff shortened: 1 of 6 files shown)


SpexGuy:
Shoot, I shouldn't have missed the meeting 😅 
`@ptrToInt` at comptime is something that sort of needs to work.  Without it, there's a pretty strange line for what compiles and what doesn't.  For example, code which does a direct ptrCast from `*T` to `[*]u8` in order to offset a pointer would be fine at comptime, but hashing code which uses `@ptrToInt` and adds a byte offset would not.

Additionally, you aren't really saving any complexity.  This code is still well defined:
```zig
pub fn betterPtrToInt(comptime ptr: anytype) usize {
    return @ptrCast(*const usize, &ptr).*;
}
```

eLeCtrOssSnake:
> This is in direct contradiction to parts of #9646 and these two test cases https://github.com/ziglang/zig/blob/master/test/behavior/comptime_memory.zig#L319-L371. As discussed in the self hosted compiler meetings the justification for this is change is that it seems like a lot of work for little value and making it an error also avoids the complications brought by comptime only types.
> 
> cc @SpexGuy
> 
> Closes #11211

This features should exist. There is no justification for removing it.
Inspecting own PE Header for example is one reason. Getting PEB on Windows requires a hard coded address.

Vexu:
> This features should exist. There is no justification for removing it.
> Inspecting own PE Header for example is one reason. Getting PEB on Windows requires a hard coded address.

I don't understand what you're trying to say.

andrewrk:
@Vexu if you wouldn't mind rebasing this and getting the CI checks passing, I'll prioritize accepting it / rejecting it.

eLeCtrOssSnake:
> > This features should exist. There is no justification for removing it.
> > Inspecting own PE Header for example is one reason. Getting PEB on Windows requires a hard coded address.
> 
> I don't understand what you're trying to say.

I confused myself on that one. I agree, in comptime it's hard to understand what compiler might come up with, so removing @ptrToInt for comptime variables makes sense to not allow "C hacks" in comptime.
The only use case I am a bit worried about is negative indexing from pointer, I am not sure how useful it is, but it might be a point for consideration as zig does not allow negative array index, so the only real way to do it is through pointer math(but negative indexing is kind of bad anyway). Again, not sure if it's a useful thing in comptime, more of a real code(not comptime) C interaction thing.
I am not thinking of any other use case for comptime pointer math as there are better alternatives with for/while loops, many item ptr's etc. All in all I think it makes sense.

eLeCtrOssSnake:
> > This features should exist. There is no justification for removing it.
> > Inspecting own PE Header for example is one reason. Getting PEB on Windows requires a hard coded address.
> 
> I don't understand what you're trying to say.

I confused myself on that one. I agree, in comptime it's hard to understand what compiler might come up with, so removing @ptrToInt for comptime variables makes sense to not allow "C hacks" in comptime.
The only use case I am a bit worried about is negative indexing from pointer, I am not sure how useful it is, but it might be a point for consideration as zig does not allow negative array index, so the only real way to do it is through pointer math(but negative indexing is kind of bad anyway). Again, not sure if it's a useful thing in comptime, more of a real code(not comptime) C interaction thing.
I am not thinking of any other use case for comptime pointer math as there are better alternatives with for/while loops, many item ptr's etc. All in all I think it makes sense.

Vexu:
The current behavior of `@ptrToInt` of only evaluating at comptime if the pointer value has a comptime representation is more in line with the rest of the language (currently) and solves the problem I originally had. I implemented reinterpreting pointers to declarations to also become runtime operations to completely solve the issue for now.

Whether this should be a compile error can be decided when solving #2471 as far as I'm concerned.

PR #11735, opened 2022-05-26
std/crypto: add support for the NIST P-384 curve

jedisct1:
After P-256, here comes P-384, also known as secp384r1.

Like P-256, it is required for TLS, and is the current NIST recommendation for key exchange and signatures, for better or for worse.

Like P-256, all the finite field arithmetic has been computed and verified to be correct by fiat-crypto.


jedisct1:
Come on Drone, you can build this :)

andrewrk:
If only it let me set the timeout to 3 hours instead of 2 we would be golden

Issue #6469, opened 2020-09-30
zig detects wrong libc version


ifreund:
My system libc version is `2.30` but zig detects and builds `2.17` which causes linking errors with `expat` for example:
```zig
const std = @import("std");
const c = @cImport({
    @cInclude("expat.h");
});
pub fn main() void {
    std.debug.print("Take off every Zig\n", .{});
}
```
```
zig-git build-exe -lc -lexpat test.zig --verbose-link
lld -error-limit=0 -z stack-size=16777216 --gc-sections -m elf_x86_64 -o test /home/ifreund/.cache/zig/o/e078967572054b1e20fb7e32da06f238/Scrt1.o /home/ifreund/.cache/zig/o/9a9a917500aa09fcd6cbafb708b1d6e3/crti.o -L /usr/local/lib64 -L /usr/local/lib -L /usr/lib/x86_64-linux-gnu -L /lib64 -L /lib -L /usr/lib64 -L /usr/lib -L /lib/x86_64-linux-gnu -dynamic-linker /lib/ld-linux-x86-64.so.2 zig-cache/o/d6ab4d674788932c736c310c43b963b3/test.o /home/ifreund/.cache/zig/o/cac0c8a5e22159f852a543a37241f19a/libcompiler_rt.a -lexpat /home/ifreund/.cache/zig/o/ef893c3d724376328171551dc3f416d7/libunwind.a /home/ifreund/.cache/zig/o/513bf1524998e06b2a99a61649abfd7d/libc.so.6 /home/ifreund/.cache/zig/o/513bf1524998e06b2a99a61649abfd7d/libm.so.6 /home/ifreund/.cache/zig/o/513bf1524998e06b2a99a61649abfd7d/libpthread.so.0 /home/ifreund/.cache/zig/o/513bf1524998e06b2a99a61649abfd7d/libdl.so.2 /home/ifreund/.cache/zig/o/513bf1524998e06b2a99a61649abfd7d/librt.so.1 /home/ifreund/.cache/zig/o/513bf1524998e06b2a99a61649abfd7d/libld.so.2 /home/ifreund/.cache/zig/o/513bf1524998e06b2a99a61649abfd7d/libutil.so.1 /home/ifreund/.cache/zig/o/66a177e6febdd316bcf2348b13964c28/libc_nonshared.a /home/ifreund/.cache/zig/o/5d923f6790009188502400b46c05b3cf/crtn.o
[...]

PR #11740, opened 2022-05-27
stage2: fix building on systems with glibc >= 2.28

### LINKED ISSUE #11137: `-lc` doesn't link `c_nonshared`

BratishkaErik:
Also updated tools/update-linux-headers.zig and tools/process_headers.zig to latest master version.

### DIFF
--- a/src/target.zig
+++ b/src/target.zig
@@ -487,6 +487,15 @@ pub fn hasRedZone(target: std.Target) bool {
 pub fn libcFullLinkFlags(target: std.Target) []const []const u8 {
     // The linking order of these is significant and should match the order other
     // c compilers such as gcc or clang use.
+    if (target.isGnuLibC()) return &[_][]const u8{
+        "-lm",
+        "-lpthread",
+        "-lc",
+        "-lc_nonshared",
+        "-ldl",
+        "-lrt",
+        "-lutil",
+    };
     return switch (target.os.tag) {
         .netbsd, .openbsd => &[_][]const u8{
             "-lm",


andrewrk:
Thanks for the patch! Knowing that this change is required starting with glibc 2.28 is useful and can be used to solve the problem - we just need to check the version of the target rather than the version used by the build script (which does not use libc at all, so does not have the correct answer to the query). I can make that change inspired by your patch and will look into this today.

BratishkaErik:
Looks like I have some sort of #6469.

My glibc version: 2.35
glibc version from `zig build-exe --show-builtin`: 2.19

build.zig
```diff
@@ -562,6 +562,10 @@ fn addCmakeCfgOptionsToExe(
                 else => |e| return e,
             };
             exe.linkSystemLibrary("unwind");
+            if (std.c.versionCheck(.{ .major = 2, .minor = 28 }).ok) {
+                // see https://github.com/bminor/glibc/commit/1a2f44a848663036c8a14671fe0faa3fed0b2a25, included in glibc 2.28 and newer
```
viriuwu, line 566: I'd recommend linking the official repository and not a random mirror on GitHub:
https://sourceware.org/git/?p=glibc.git;a=commit;h=1a2f44a848663036c8a14671fe0faa3fed0b2a25

BratishkaErik: You're right.

BratishkaErik:
Um, was "sparc" arch renamed to "sparc64" in glibc or I did it wrong?

BratishkaErik:
> UPD: I will test this on system with glibc between 2.18 and 2.34

Guess what... it works fine with and **without** this patch!

BratishkaErik:
This DOESN'T resolve #11137, because `link_c_nonshared` will incorrectly set to false, see #6469

src/target.zig
```diff
@@ -487,6 +487,35 @@ pub fn hasRedZone(target: std.Target) bool {
 pub fn libcFullLinkFlags(target: std.Target) []const []const u8 {
     // The linking order of these is significant and should match the order other
     // c compilers such as gcc or clang use.
+
+    // Starting with glibc 2.28, we need to link "c_nonshared" for "pthread_atfork" function
```
ifreund, line 491: Is there any harm to linking this for older glibc versions? how long has this `c_nonshared` library existed?

src/target.zig
```diff
@@ -487,6 +487,35 @@ pub fn hasRedZone(target: std.Target) bool {
...
+        return &[_][]const u8{
+            "-lm",
+            "-lpthread",
+            "-lc",
+            "-lc_nonshared",
+            "-ldl",
+            "-lrt",
+            "-lutil",
+        };
+    }
```
ifreund, line 517: Glibc is only present on linux. I think all this code should be moved into a linux branch of the switch below.

src/target.zig
BratishkaErik: > how long has this c_nonshared library existed?

https://askubuntu.com/questions/497306/cannot-compile-gcc-or-g-libc-nonshared-a-not-recognized 
Jul 13, 2014 

> Would linking this unconditionally break anything for glibc versions?

AFAIK default linker script has been doing this since ancient times.

src/target.zig
BratishkaErik: ok

BratishkaErik: > Glibc is only present on linux.

And GNU/Hurd.

ifreund:
> Also updated tools/update-linux-headers.zig and tools/process_headers.zig to latest master version.

This isn't really related and should really go in a separate PR

src/target.zig
ifreund: I'm not sure if anyone has ever tried compiling zig code targeting hurd, though it seems we do have a `hurd` target os.

Anyhow, I suspect there will be many other issues to cross before we could even test if this works on hurd.

BratishkaErik:
> I tested different strategies

still testing

PR #11741, opened 2022-05-27
src/print_env: print the native target

perillo:
Add the native target triple to the zig env command output.

The target triple is formatted the same way as it is done in the
zig targets command.

### DIFF
--- a/src/print_env.zig
+++ b/src/print_env.zig
@@ -21,6 +21,10 @@ pub fn cmdEnv(gpa: Allocator, args: []const []const u8, stdout: std.fs.File.Writ
     const global_cache_dir = try introspect.resolveGlobalCacheDir(gpa);
     defer gpa.free(global_cache_dir);
 
+    const info = try std.zig.system.NativeTargetInfo.detect(gpa, .{});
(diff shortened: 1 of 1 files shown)


andrewrk:
Note that this information is available via `zig targets`:

```
andy@ark ~> zig targets | jq .native.triple
"x86_64-linux.5.10.110...5.10.110-gnu.2.33"
```

perillo:
> Note that this information is available via `zig targets`:
> 
> ```
> andy@ark ~> zig targets | jq .native.triple
> "x86_64-linux.5.10.110...5.10.110-gnu.2.33"
> ```

Yes, I'm aware; and using jq will make it convenient for the user.  The advantage is that it works  right now.

However a possible advantage of `zig env` is that in future it can show more debug data like `zig_backend`, `link_libc`, `valgrin_support` and the dynamic linker used.
It will also make very convenient for the user to report a bug, since it can just copy and paste the output of `zig env`, like it is done with the Go language bug template.

Thanks.

andrewrk:
I like it. Thank you.

PR #11742, opened 2022-05-27
Add support for --no-entry linker flag.

### LINKED ISSUE #11045: "no-entry" argument not recognized
### Zig Version

0.10.0-dev.871+bb05a8a08

### Steps to Reproduce

run this:

```
echo "int add(int a,int b){return a+b;}" > main.c
zig cc main.c -target wasm32-freestanding-none -nostdlib -o binary.wasm -Wl,--no-entry
```

### Expected Behavior

should compile without issue

### Actual Behavior

output:
```
warning: unsupported linker arg: --no-entry
wasm-ld: error: entry symbol not defined (pass --no-entry to suppress): _start
```

similarly, if you run
`zig cc main.c -target wasm32-freestanding-none -nostdlib -o binary.wasm --no-entry`
you get
`error: Unknown Clang option: '--no-entry'`

so whether you pass the argument to zig cc with `--no-entry` or the linker with `-Wl,--no-entry`, it doesn't work

danielchasehooper:
This flag is only valid for wasm targets.
Usage:
zig cc main.c -target wasm32-freestanding-none -o binary.wasm -Wl,--no-entry

Fixes #11045

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -769,6 +769,7 @@ pub const InitOptions = struct {
     linker_nxcompat: bool = false,
     linker_dynamicbase: bool = false,
     linker_optimization: ?u8 = null,
+    linker_no_entry: bool = false,
     major_subsystem_version: ?u32 = null,
     minor_subsystem_version: ?u32 = null,
     clang_passthrough_mode: bool = false,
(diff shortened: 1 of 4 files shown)


Luukdegram:
Hi, thank you for the contribution!
Unfortunately, this goes against the goals of Zig (and Zig cc) where we prefer to set correct defaults.
For instance, when you pass `-shared` to `zig cc` the linker frontend will already automatically pass `--no-entry` to wasm-ld. Building an 'executable' for Wasm really only makes sense in the context of WASI.

In conclusion, to handle your use case, you have to build your project as a library, rather than an executable (e.g. by passing `-shared`).

PR #11743, opened 2022-05-28
place zig-cache directory next to build.zig

### LINKED ISSUE #11672: when choosing a default local zig-cache directory, try to place it next to build.zig, falling back to global zig-cache
Currently, Zig will choose a default local zig-cache directory relative to the current working directory. This is not always desirable, and can lead to multiple unwanted zig-cache directories, and even drives some users to unconditionally override the local zig-cache directory.

Instead, the logic for choosing a default local zig-cache directory should be as follows:

 * Search upwards in parent directories from the cwd, looking for a `build.zig` file. If this is found, then choose that same directory containing that file with `zig-cache` as a sub-directory.
 * If no build.zig file is found, or other file system errors occur such as "access denied", then use the same path as the global zig-cache.

matu3ba:
* search upwards, if no build.zig is found
* if this fails or any error occurs, fallback to global zig-cache

Closes #11672

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -2592,13 +2592,28 @@ fn buildOutputType(
             break :l global_cache_directory;
         }
         if (main_pkg) |pkg| {
-            const cache_dir_path = try pkg.root_src_directory.join(arena, &[_][]const u8{"zig-cache"});
-            const dir = try pkg.root_src_directory.handle.makeOpenPath("zig-cache", .{});
-            cleanup_local_cache_dir = dir;
-            break :l .{
-                .handle = dir,
-                .path = cache_dir_path,
-            };
+            // search upwards from cwd until we find directory with build.zig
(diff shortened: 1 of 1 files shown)


matu3ba:
Tested locally with 1. `tryzig` toy folder and 2. `zig build test-standalone`, 3. init-exe repo and the expected places for `zig-cache` are being used.

Only potential blocker is outstanding question, what the behavior of `tmpDir` in `std.testing.zig` should be. I do expect the potential for name clashes, if they will be run in parallel by user chosen name.

marler8997:
We don't have to do it in this change, but it might be nice to expose this logic in the std library so external tools can also find the build root (ide's language servers, etc)

matu3ba:
Do you have an opinion where this logic could or should live in std? It is only loosely related to build.zig.

andrewrk:
Looks good!

PR #11744, opened 2022-05-28
Treat blocks with "return" as "noreturn"

### LINKED ISSUE #11686: `pub fn main() void { { return; } return; }` evades unreachable code error
### Zig Version

0.10.0-dev.2315+dd6ac9a22

### Steps to Reproduce

```zig
pub fn main() void {
    {
        return;
    }
    return;
}
```

### Expected Behavior

```
main.zig:5:5: error: unreachable code
    return;
    ^
main.zig:3:5: note: control flow is diverted here
    return;
    ^
```

### Actual Behavior

A successful compilation.

gpanders:
Block statements that end with "break" should not be considered
"noreturn" for the enclosing scope, but other "noreturn" instructions
(return, panic, compile error, etc.) should be. This differentiation
necessitates handling "break" differently from the other "noreturn"
instructions when inside a block statement.

Closes #11686.


src/Zir.zig
```diff
@@ -1250,6 +1250,271 @@ pub const Inst = struct {
...
+                .compile_error,
+                .ret_node,
+                .ret_load,
+                .ret_tok,
+                .ret_err_value,
+                .@"unreachable",
+                .repeat,
+                .repeat_inline,
+                .panic,
+                => false,
```
Vexu, line 1510: I think it's pretty safe to replace these 200 lines with an `else => false`.

gpanders: Sounds good. I listed them explicitly to copy `endsWithNoReturn()`, as I wasn't sure if that was done for a specific reason.

andrewrk:
I think the patch should fully solve the issue in this case; this halfway solution is doing more harm than good. `endsWithBreak()` is a footgun.

> This didn't apply to labeled blocks.

Not sure what you mean by "this" - did you mean the logic `if (gz.endsWithNoReturn() and !gz.endsWithBreak()) {` was in `blockExpr`? That's true in my example above as well. The break target is labeled, but the containing block is unlabeled.

gpanders:
@andrewrk I didn't think of the nested block labels. I'll circle back to this and give it a closer look.

andrewrk:
Thanks @gpanders! I always feel bad doing a revert, please don't take it personally. Just another part of the development process.

PR #11745, opened 2022-05-28
ISSUE_TEMPLATE: add a required "target" field to bug.yml

### LINKED ISSUE #11738: .github/ISSUE_TEMPLATE: add a required "target" field to bug.yml
Reading some open issues I noted that it is not easy to determine if a bug manifests only for some operating systems (like Windows).

I propose to add a new required field to `bug.yml` named `target`, containing the `arch-os-abi` triple.

The Zig compiler should also have a command to display the native target.
Currently the only method I know is `zig targets | grep -A 512 native`

Thanks.

perillo:
Add a new required field "Zig Target Triple" to the bug.yml template.

Closes #11738

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -17,6 +17,14 @@ body:
       placeholder: "0.9.0-dev.1275+ac52e0056"
     validations:
       required: true
+  - type: input
+    id: target
+    attributes:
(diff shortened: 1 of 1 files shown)


perillo:
A living example of the issue template is in https://github.com/perillo/test/issues/new?assignees=&labels=bug&template=bug.yml

andrewrk:
I don't find this to be worth the friction. Thanks anyway

PR #11746, opened 2022-05-28
Allocator: add alignedCreate

devins2518:
I added an `alignedCreate` function because all of the other allocating functions had some variation which took could specify alignment, but `create` strangely did not.


Vexu:
Reverted in 72c60acd5dd9172c648cb3c5a4b51224ebea33f2.

The test did not pass with stage3 and I also missed that they were using `assert` instead of `testing.expect`.

PR #11754, opened 2022-05-30
std.fmt: require explicit specifier for unwrapping optionals and error unions.

### LINKED ISSUE #11753: `std.fmt`: disallow implicitly formatting optional and error values
On Discord we've discussed and agreed to that it is currently too easy to accidentally pass an optional value (`?T`) to a format specifier like `{s}`. It is a real footgun and reminds me a lot to `undefined` that you can find on JavaScript websites in many places because the developer forgot to handle it. We don't want Zig to be like that. It shouldn't happily print `null` like that.

@InKryption had the idea of only allowing an optional type to be formatted if there is `?` before the specifier type, so for `{s}` it would be `{?s}`.

Then @MasterQ32 asked about error unions and I proposed for example `{!?s}` to format `anyerror!?[]const u8`. I think this syntax looks really neat and it's consistent with the way the types look.

Here's what **cipharius** had to say:
> That would be definitely more intuitive, as I previously also expected a type error, since in other contexts Zig is pretty strict about type inference.

So the proposal is to change `std.fmt` formatting such that in a format string

* `{t}` can only format `T`,
* `{?t}` can only format `?T` (and `T` as that coerces to it),
* `{!?t}` can only format `anyerror!?T`.

where `t` is the respective specifier type. This affects all specifier types except for `any` which will retain current behavior.

CC @InKryption @viriuwu @MasterQ32

InKryption:
Closes #11753.


InKryption:
Currently in the process of fixing amending the code causing the CI to fail; I will only modify the format strings such that the already-present behavior is preserved, but a future review of these sites, where optionals and error unions are formatted without unwrapping, may be of interest, in the event this PR is merged.

r00ster91:
Here's some documentation we could add here: https://github.com/ziglang/zig/blob/master/lib/std/fmt.zig#L62

```zig
///
/// Note that optionals and error unions have to be handled explicitly using `{?}` and `{!}`, respectively.
/// These can also be combined with the specifier types listed above, e.g. `{!?s}` to format `anyerror!?[]const u8`. 
///
```

src/translate_c.zig
```diff
@@ -2766,7 +2766,7 @@ fn transInitListExpr(
         ));
     } else {
         const type_name = c.str(qual_type.getTypeClassName());
-        return fail(c, error.UnsupportedType, source_loc, "unsupported initlist type: '{s}'", .{type_name});
+        return fail(c, error.UnsupportedType, source_loc, "unsupported initlist type: '{!s}'", .{type_name});
```
Vexu, line 2769: All of the changes in this file should be `try`s instead. The cache related changes look suspicious too but I'm not as familiar with that and there isn't doc comment explaining why the `path` is optional.

InKryption: Yes, I suspected this was the case, but as I mentioned earlier, I didn't want to make these changes in this PR, and chose to simply preserve the current behavior, so as to not break anything that is unrelated to this proposed change.

Now that these have been pointed out, they can be sorted through and amended properly in a separate PR.

r00ster91: Great to see it uncover some issues. This is what it's for.

jibal:
> I will only modify the format strings such that the present behavior is preserved, but a future review of these sites, where optionals and error unions are formatted without unwrapping, may be of interest, in the event this PR is merged.

There ought to be a TODO on each of these, otherwise the point of this change, to catch these bugs, is being defeated.

InKryption:
Messed things up here with the merge conflicts. Going to open a fresh PR later maybe.

PR #11763, opened 2022-05-31
Stage2: detect when initializer of const variable is comptime known

### LINKED ISSUE #11439: Stage2 makes struct and array initialization with comptime values always runtime
```zig
test {
    const S = struct {
        a: bool,
    };
    const s: S = .{ .a = false };
    if (s.a) @compileError("bad");
}
```

Vexu:
~~This PR makes it so that const variables produce a `decl_ref_const_init` value which behaves similarly to `decl_ref_mut` but also allows runtime stores if needed by converting the comptime pointer into a set of runtime instructions. The alloc and the instructions are placed in a `resolve_const_alloc` block replacing the hacky `make_ptr_const` which detects if any runtime instructions were emitted for the initializer and then either converts the original alloc into a runtime instruction or a comptime value.~~

~~The mechanism is quite complex but it's the simplest I could think of capable of solving more complex situations with nested structs/arrays producing chained `field_ptr`/`ptr_elem_ptr` instructions:~~
```zig
pub export fn entry() void {
    const T = struct { c: u8 };
    const S = struct { a: bool, b: [2]T };

    var foo = false;
    var bar: u8 = 1;
    const s: S = .{
        // .a = foo,
        .a = false,
        .b = [2]T{
            T{ .c = 1 },
            // T{ .c = bar },
            T{ .c = 3 },
        },
    };
    if (s.a) @compileError("bad");

    _ = foo;  _ = bar; _ = s;
}
```

~~This PR is currently a proof of concept since I'd like to get some comments on it before putting even more time into polishing it.~~

Closes #11439

### DIFF
--- a/lib/std/math/big/int.zig
+++ b/lib/std/math/big/int.zig
@@ -2063,9 +2063,8 @@ pub const Const = struct {
         // This is the inverse of calcDivLimbsBufferLen
         const available_len = (limbs.len / 3) - 2;
 
-        // TODO https://github.com/ziglang/zig/issues/11439
-        const biggest = comptime Const{
-            .limbs = &([1]Limb{math.maxInt(Limb)} ** available_len),
+        const biggest: Const = .{
+            .limbs = &([1]Limb{comptime math.maxInt(Limb)} ** available_len),
             .positive = false,
         };
         var buf: [biggest.sizeInBaseUpperBound(radix)]u8 = undefined;
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -3215,7 +3215,9 @@ fn validateUnionInit(
         return sema.failWithOwnedErrorMsg(block, msg);
     }
 
-    if (is_comptime or block.is_comptime) {
+    if ((is_comptime or block.is_comptime) and
+        (try sema.resolveDefinedValue(block, init_src, union_ptr)) != null)
+    {
         // In this case, comptime machinery already did everything. No work to do here.
         return;
     }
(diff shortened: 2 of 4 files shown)


andrewrk:
I played around with this test case:

```zig
test {
    const S = struct {
        a: bool,
    };
    const s: S = .{ .a = false };
    if (s.a) @compileError("bad");
}
```

The relevant ZIR here is:

```
      %118 = validate_struct_init({
        %116 = field_ptr(%115, "a") node_offset:31:26
      }) node_offset:31:19
      %119 = make_ptr_const(%114) node_offset:31:5
```

I noticed that `validate_struct_init` did successfully notice that the struct value was comptime known, and rewrote the AIR into a `store(%alloc, %init)` where `%init` is a `constant`. The idea here is that it will be followed by the logic in `resolve_inferred_alloc` which notices such a store and rewrites the alloc itself to be comptime. If we change the source slightly:

```zig
test {
    const S = struct {
        a: bool,
    };
    const s = S{ .a = false };
    if (s.a) @compileError("bad");
}
```

In theory nothing should be different, but the ZIR ends up being:

```
      %119 = validate_struct_init({
        %117 = field_ptr(%116, "a") node_offset:31:23
      }) node_offset:31:16
      %120 = resolve_inferred_alloc(%114) node_offset:31:5
```

And in this case the `resolve_inferred_alloc` logic successfully makes `s` comptime-known, making the test pass.

Perhaps there is a simpler change here which attempts to take advantage of this already-working combination of analyses?

Vexu:
I did see all of those but I was getting some strange bitcasts that were preventing them from working but apparently not anymore? I'll try the simpler alternative of copying some code around and then try to find out if there actually were issues or was I just doing something stupid.

PR #11766, opened 2022-05-31
`std.testing`: make the caret indicator line more helpful

r00ster91:
I think this will make it even easier to debug non-matching strings. Some examples:
```zig
    try testing.expectEqualStrings("hello", "hell");
```
```
hello␃

hell␃

First difference occurs on line 1:
expected:
hello
    ^ (111)
found:
hell
    ^ (out of bounds)
```
```zig
    const not_hello: [5]u8 = undefined;
    try testing.expectEqualStrings("hello", &not_hello);
```
```
hello␃

␃

First difference occurs on line 1:
expected:
hello
^ (104)
found:

^ (0)
```
This is extremely useful if your character is unprintable.
Wasn't sure if this should be hexadecimal so for now I used `{d}`.


lib/std/testing.zig
```diff
@@ -501,7 +501,10 @@ fn printIndicatorLine(source: []const u8, indicator_index: usize) void {
         while (i < indicator_index) : (i += 1)
             print(" ", .{});
     }
-    print("^\n", .{});
+    if (indicator_index >= source.len)
+        print("^ (out of bounds)\n", .{})
```
motiejus, line 505: nit: perhaps call this `(end of string)`?

motiejus (approved):
I really would have liked this a few weeks ago when I was debugging many strings, some of which had non-printable characters. :) Looking forward.

lib/std/testing.zig
r00ster91: That's better I think!

r00ster91:
#12214

PR #11768, opened 2022-06-01
Small fixes for ResetEvent and Mutex

kprotty:
* simplifies the Mutex code to avoid the comptime field stuff
* fixes the Barrier in the ResetEvent test case to actually wait

### DIFF
--- a/lib/std/Thread/Mutex.zig
+++ b/lib/std/Thread/Mutex.zig
@@ -121,7 +121,7 @@ const FutexImpl = struct {
 
     const unlocked = 0;
     const locked = 1;
-    const contended = 2; 
+    const contended = 2;
 
     fn tryLock(self: *Impl) bool {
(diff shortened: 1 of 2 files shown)


lib/std/Thread/Mutex.zig
```diff
@@ -119,37 +119,30 @@ const DarwinImpl = struct {
 const FutexImpl = struct {
     state: Atomic(u32) = Atomic(u32).init(unlocked),
 
-    const unlocked = 0b00;
-    const locked = 0b01;
-    const contended = 0b11; // must contain the `locked` bit for x86 optimization below
+    const unlocked = 0;
+    const locked = 1;
+    const contended = 2;
```
haze, line 124: finally

kprotty:
Outdated - will get back to this another time.

PR #11782, opened 2022-06-03
Autodoc: added some builtin tags and refactored some unary and binary operators

vallahor:
The code for binary operators are much more clean now and easy to maintain.
I think I add all builtin functions related with unary and binary operatos, the ones that are missing need to handle their structures maybe it's better to discuss about and see if worth handling them now, there are some which are considered as `un_node` field but takes 2 arguments.
They are:
`clz`
`ctz`
`pop_count`
`byte_swap`
`bit_reverse`

I changed the `switch_block` to print some relevant information. Now I need to figure out how to get the line of the switch case because what I can get from ast is the relative line of the parent.
By doing that there are a easy way to generate links to github because the `file` structure already has the file name with path.

I add `float128` too and get stuck in `int_big`.

### DIFF
--- a/lib/docs/main.js
+++ b/lib/docs/main.js
@@ -1159,7 +1159,6 @@ var zigAnalysis;
 
             let payloadHtml = "@";
             switch (expr.builtin.name) {
-
               case "align_of": {
                 payloadHtml += "alignOf";
                 break;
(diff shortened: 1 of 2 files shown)


kristoff-it:
Thanks!

PR #11784, opened 2022-06-03
Remove duplicated switch logic in float.zig

OBenjaminT:
Arguably unnecessary as it's all comptime anyway.

I don't know if the asserts should stay in the direct function.

### DIFF
--- a/lib/std/math/float.zig
+++ b/lib/std/math/float.zig
@@ -30,32 +30,17 @@ pub fn floatExponentBits(comptime T: type) comptime_int {
 
 /// Returns the number of bits in the mantissa of floating point type T.
 pub fn floatMantissaBits(comptime T: type) comptime_int {
-    assert(@typeInfo(T) == .Float);
-
-    return switch (@typeInfo(T).Float.bits) {
-        16 => 10,
-        32 => 23,
-        64 => 52,
-        80 => 64,
-        128 => 112,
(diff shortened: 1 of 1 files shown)


lib/std/math/float.zig
```diff
@@ -30,32 +30,17 @@ pub fn floatExponentBits(comptime T: type) comptime_int {
...
-
-    return switch (@typeInfo(T).Float.bits) {
-        16 => 10,
-        32 => 23,
-        64 => 52,
-        80 => 64,
-        128 => 112,
-        else => @compileError("unknown floating point type " ++ @typeName(T)),
-    };
+    return @typeInfo(T).Float.bits - floatExponentBits(T) - 1;
```
ifreund, line 33: I personally prefer the status quo here. It's a lot nicer when reading code and wondering what this function will return for a given floating point type.

OBenjaminT: That's why my comment was uncertain about if this kind of thing was good in general. Especially as this is something unlikely to change frequently.

But in this case the called function is right above, and it's slightly more clear on what the bits are and where the numbers come from. 

I don't mind if it gets rejected, but thought it was worth suggesting. It doesn't take long to choose once you have the choice placed in front of you.

andrewrk:
Thanks @OBenjaminT and @ifreund

Issue #6452, opened 2020-09-29
unhandled NTSTATUS: DELETE_PENDING in std.os.windows.DeleteFile


andrewrk:
cc @kubkon 

```
c:\msys64\home\andy\dev\zig\build-llvm10>bin\zig.exe build test
Test [1/3] test "self-hosted"... tests [1/20] referencing decls which appear later in the file [1/1] update [1/3] write.tests [1/20] referencing decls which appear later in the file [2/1] error.Unexpected NTSTATUS=0xc0000056
C:\msys64\home\andy\dev\zig\build-llvm10\lib\zig\std\debug.zig:410:53: 0x7ff71cc81dc0 in std.debug.writeCurrentStackTraceWindows (test.obj)
    const n = windows.ntdll.RtlCaptureStackBackTrace(0, addr_buf.len, @ptrCast(**c_void, &addr_buf), null);
                                                    ^
C:\msys64\home\andy\dev\zig\build-llvm10\lib\zig\std\debug.zig:395:45: 0x7ff71cbb675f in std.debug.writeCurrentStackTrace (test.obj)
        return writeCurrentStackTraceWindows(out_stream, debug_info, tty_config, start_addr);
                                            ^
C:\msys64\home\andy\dev\zig\build-llvm10\lib\zig\std\debug.zig:116:31: 0x7ff71cb25008 in std.debug.dumpCurrentStackTrace (test.obj)
        writeCurrentStackTrace(stderr, debug_info, detectTTYConfig(), start_addr) catch |err| {
                              ^
C:\msys64\home\andy\dev\zig\build-llvm10\lib\zig\std\os\windows.zig:1572:40: 0x7ff71cbc19ee in std.os.windows.unexpectedStatus (test.obj)
        std.debug.dumpCurrentStackTrace(null);
                                       ^
C:\msys64\home\andy\dev\zig\build-llvm10\lib\zig\std\os\windows.zig:828:40: 0x7ff71cdb79fa in std.os.windows.DeleteFile (test.obj)
        else => return unexpectedStatus(rc),
                                       ^
C:\msys64\home\andy\dev\zig\build-llvm10\lib\zig\std\os.zig:1850:30: 0x7ff71cd2528c in std.os.unlinkatW (test.obj)
    return windows.DeleteFile(sub_path_w, .{ .dir = dirfd, .remove_dir = remove_dir });
                             ^
[...]

Issue #5537, opened 2020-06-04
Windows: std.fs.Dir.deleteDir() will silently fail on non-empty directories


squeek502:
When `./dir` is non-empty, the following code will succeed but the directory will not be deleted. I expect `error.DirNotEmpty` to be returned in this case.

```zig
const std = @import("std");

pub fn main() !void {
    return std.fs.cwd().deleteDir("dir");
}
```

Both NtCreateFile and NtClose seem to be returning `SUCCESS` here:

https://github.com/ziglang/zig/blob/fd067fbe8b14c445ed764da15fff5163156126df/lib/std/os.zig#L1768-L1791

squeek502:
From https://go.microsoft.com/fwlink/?LinkId=140636 section 4.3.1 (page 33):

> Note that if the DELETE_ON_CLOSE flag is specified on a directory with child files or directories, the create operation will succeed, but the delete on close flag will be silently ignored when processing the cleanup IRP and the directory will not be deleted.

kubkon:
Should be fixed with #6397 so closing. If the issue persists, please feel free to reopen!

squeek502:
Should be re-opened as #6397 was reverted in #6462

(for some reason I can't re-open it even though I started this issue?)

PR #11793, opened 2022-06-04
Use posix semantics when deleting files on Windows

SpexGuy:
When a file is deleted on Windows, it may not be immediately removed from the directory entry.  This can cause problems with future scans of that directory, which will see the partially deleted file.  However, it should be noted that under some workloads and system configurations, Windows files may appear to be deleted immediately.  This seems to be why the CI is not failing under this bug.

This change forces file deletion to use posix semantics, removing it from the directory immediately on delete.

This fixes a bug that occurs on some windows systems/hard drives where `std.fs.Dir.deleteTree` fails due to assumed posix semantics.  This may also fix rare bugs in other systems (like the cache system), which may assume posix semantics for file deletion.

### DIFF
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -912,7 +912,7 @@ pub fn DeleteFile(sub_path_w: []const u16, options: DeleteFileOptions) DeleteFil
         0,
     );
     switch (rc) {
-        .SUCCESS => return CloseHandle(tmp_handle),
+        .SUCCESS => {},
         .OBJECT_NAME_INVALID => unreachable,
         .OBJECT_NAME_NOT_FOUND => return error.FileNotFound,
         .OBJECT_PATH_NOT_FOUND => return error.FileNotFound,
@@ -921,6 +921,21 @@ pub fn DeleteFile(sub_path_w: []const u16, options: DeleteFileOptions) DeleteFil
         .NOT_A_DIRECTORY => return error.NotDir,
         .SHARING_VIOLATION => return error.FileBusy,
         .CANNOT_DELETE => return error.AccessDenied,
+        else => return unexpectedStatus(rc),
+    }
+    defer CloseHandle(tmp_handle);
+
+    var info = FILE_DISPOSITION_INFORMATION_EX{
+        .Flags = FILE_DISPOSITION_DELETE |
+            FILE_DISPOSITION_POSIX_SEMANTICS |
+            FILE_DISPOSITION_ON_CLOSE |
+            FILE_DISPOSITION_IGNORE_READONLY_ATTRIBUTE,
+    };
+    rc = ntdll.NtSetInformationFile(tmp_handle, &io, &info, @sizeOf(FILE_DISPOSITION_INFORMATION_EX), .FileDispositionInformationEx);
+    switch (rc) {
+        .SUCCESS => {},
+        // CANNOT_DELETE indicates that the file is currently mapped
+        .CANNOT_DELETE => return error.FileBusy,
         else => return unexpectedStatus(rc),
     }
 }
@@ -2376,6 +2391,18 @@ pub const FILE_NAME_INFORMATION = extern struct {
     FileNameLength: ULONG,
     FileName: [1]WCHAR,
 };
+
+pub const FILE_DISPOSITION_INFORMATION_EX = extern struct {
+    /// combination of FILE_DISPOSITION_* flags
+    Flags: ULONG,
+};
+
+const FILE_DISPOSITION_DO_NOT_DELETE: ULONG = 0x00000000;
+const FILE_DISPOSITION_DELETE: ULONG = 0x00000001;
+const FILE_DISPOSITION_POSIX_SEMANTICS: ULONG = 0x00000002;
+const FILE_DISPOSITION_FORCE_IMAGE_SECTION_CHECK: ULONG = 0x00000004;
+const FILE_DISPOSITION_ON_CLOSE: ULONG = 0x00000008;
+const FILE_DISPOSITION_IGNORE_READONLY_ATTRIBUTE: ULONG = 0x00000010;
 
 pub const FILE_RENAME_INFORMATION = extern struct {
     ReplaceIfExists: BOOLEAN,


SpexGuy:
Looks like the mac os test failed for an unrelated reason?  I think this is good to go, unless there's some connection I'm missing.

matu3ba:
I think this is the correct approach rather than work around Windows changing file semantics in Windows 10 (found it here https://news.ycombinator.com/item?id=23745019) https://stackoverflow.com/questions/60424732/did-the-behaviour-of-deleted-files-open-with-fileshare-delete-change-on-windows

The other workaround with renaming sounds more cumbersome and changing it did reportedly still have the same behavior https://github.com/XAMPPRocky/remove_dir_all/issues/7 and https://github.com/cargo-generate/cargo-generate/issues/619.

One annoyance of all those issues is that reporters did not provide a script + program to reproduce.

andrewrk:
A few notes:

 * Unfortunately the macOS failure does look relevant. Looks to be a separate self-hosted compiler bug that is revealed when trying to compile the new code in this patch. I'm happy to look into this as a prerequisite fix before merging this PR.
 * Can we please document clearly the intended semantics of DeleteFile in doc comments for this function, so that we know what it is supposed to do - particularly with regards to this subtle Windows behavior. This is mostly on me for not doing this documentation originally, but better late than never. Also let's make sure the semantics are clear in the cross-platform function `std.fs.Dir.deleteFile`.
 * I'm concerned that this code which deletes a file, a common file system operation, will now take two syscalls instead of one. I don't fully understand why this is necessary and I'm not yet convinced that we have to give up on this one. Are there circumstances where the former behavior is acceptable? Is this perhaps a case where the API can bifurcate, allowing the caller to request the semantics they need, so that an extra syscall is avoided when possible?

squeek502:
Related to / fixes (?) https://github.com/ziglang/zig/issues/6452

Does not affect https://github.com/ziglang/zig/issues/5537

SpexGuy:
> Related to / fixes (?) https://github.com/ziglang/zig/issues/6452

Good find, that's the stack trace that is currently blocking me, which is why I made this PR.  This triggers consistently on my computer when I try to run compiler tests (`deleteTree` is used by `std.testing.tmpDir`), so I cannot run tests locally until this is fixed.

> I'm concerned that this code which deletes a file, a common file system operation, will now take two syscalls instead of one.

It's actually three now instead of two, deleting a file already requires `NtCreateFile` + `CloseHandle`.  `NtCreateFile` is *slow* on Windows so the overhead of an extra syscall is not really in the same ball park.  That said, we could have a different function which uses DOS semantics if needed.

> I don't fully understand why this is necessary and I'm not yet convinced that we have to give up on this one.

`std.fs.Dir.deleteTree` makes the implicit assumption that deleted files do not appear when iterating over directories, so it does not work correctly on Windows systems where the deletion is not immediate (causing #6452).  Whether deletion is immediate can depend on things like the current behavior of the virus scanner, the capacity of the hard drive, and the general speed of the processor, so it may not be visible on all systems (like the CI machine, apparently).

> Are there circumstances where the former behavior is acceptable? Is this perhaps a case where the API can bifurcate, allowing the caller to request the semantics they need, so that an extra syscall is avoided when possible?

There probably are circumstances where DOS semantics are ok, however given `deleteTree` there are probably other places in the std lib that assume posix semantics.  We would need to audit them all and decide if we can switch to the DOS semantics.  This is especially difficult because dev machines tend to be fast and may not properly repro DOS semantics (the time between deletion and removal from the directory is nonzero but very small), so using DOS semantics may lead to unreliable and difficult-to-reproduce bugs that only manifest on low-end hard drives or in production.

IMO the default should be posix semantics, with a potential opt-in to DOS semantics to avoid the syscall if performance is worth introducing a possible race condition.

> Also worth noting that this requires Windows 10 >= 1607 (`_WIN32_WINNT_WIN10_RS1`)

Ah, that's a problem.  We probably need a different implementation then.

matu3ba:
boost works like the following:
```txt
1. set_file_information_by_handle with file_disposition_info_ex_class for deletion; break on success;
2. ifnot ERROR_ACCESS_DENIED goto error;
3. get_file_information_by_handle_ex with file_basic_info_class (basic file info like access time and access permissions); goto error on failure;
4. set_file_information_by_handle without READ-ONLY; goto error on failure;
5. set_file_information_by_handle with file_disposition_info_ex_class for deletion; break on success;
6. set_file_information_by_handle with READ-ONLY; goto error;
```

Thus the (only?) remaining question is, how we should deal with read-only files.

Context
https://docs.microsoft.com/en-us/windows/win32/api/minwinbase/ne-minwinbase-file_info_by_handle_class
https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-ntquerydirectoryfile

andrewrk:
> Also worth noting that this requires Windows 10 >= 1607 (`_WIN32_WINNT_WIN10_RS1`)

According to Microsoft, Windows 8.1 reaches end of Extended Support on January 10, 2023.
[source](https://learn.microsoft.com/en-us/lifecycle/faq/windows)

squeek502:
> > Also worth noting that this requires Windows 10 >= 1607 (`_WIN32_WINNT_WIN10_RS1`)
> 
> According to Microsoft, Windows 8.1 reaches end of Extended Support on January 10, 2023. [source](https://learn.microsoft.com/en-us/lifecycle/faq/windows)

Not sure if/how this factors in, but needing Windows 10 >= 1607 means that builds of Windows 10 earlier than that will also not have the feature; i.e it's not new with Windows 10, it was added in a particular build of Windows 10.

andrewrk:
> It might be worth looking at another implementation like boost::fs, which implements posix semantics on windows.

Also useful: https://github.com/python/cpython/issues/84324

andrewrk:
This PR can still be merged if a target OS version check is added. `@enumToInt(builtin.target.os.version_range.windows.min) >= @enumToInt(std.Target.Os.WindowsVersion.win10_rs1)`

andrewrk:
Unfortunately this was never quite merge-ready. Any contributor is welcome to resurrect this PR against latest master. As far as I know it only needs a OS min version check added.

PR #11800, opened 2022-06-05
`zig2 build test-std` progress

### LINKED ISSUE #11162: Aggregate initializers should allow initializing comptime fields, verifying equality

### LINKED ISSUE #11788: AstGen doesn't pass type to switch prong
```zig
test {
    var a: u32 = 0;
    var b: [2]u32 = undefined;
    b[0] = @as(u32, switch (a) {
        1 => 1,
        else => 0,
    });
}
```
```
a.zig:29:21: error: value with comptime only type 'comptime_int' depends on runtime control flow
    b[0] = @as(u32, switch (a) {
                    ^
a.zig:29:29: note: runtime control flow here
    b[0] = @as(u32, switch (a) {
                            ^
```
Related #11135

Vexu:
This gets us to three unique errors:
```
std/io/reader.zig:206:26: error: unreachable else prong; all cases already handled           (#11798)
std/net.zig:1340:12: error: unable to resolve inferred error set
std/crypto/aes/aesni.zig:182:20: error: TODO implement Sema for asm with more than 1 output
```

Closes  #11788
Closes #11162


andrewrk:
> stage2: use correct type (u29) for alignment

Ah, on this one I wanted to change the spec to make alignment `u16`. I can make a corresponding (and breaking) stage1 change along with this.

That said, since you did it in a separate commit that is easily revertable we could also move forward with this and make the change later.

Vexu:
> That said, since you did it in a separate commit that is easily revertable we could also move forward with this and make the change later.

Either way works for me, `std.heap.testAllocatorAlignedShrink` currently depends on the `u29` alignment.

andrewrk:
Your call. If you have the PR green I certainly would not want to block the merge on the alignment type issue which we can do later.

andrewrk:
Made 2 tiny fixups upon merging:
 * added an extra assertion to one of new behavior tests
 * fixed typo in commit message

Great work!

PR #11813, opened 2022-06-07
`zig2 build test-std` finale

### LINKED ISSUE #11798: Stage2: unreachable else prong
A common pattern in Zig code is to only partially handle errors and return the rest to the caller with something like:
```zig
foo() catch |err| switch (err) {
    error.TheOneICareAbout => // something
    else => |e| return e,
}
```
With stage2 this causes issues in generic code where `error.TheOneICareAbout` is sometimes the only error and results in an `error: unreachable else prong; all cases already handled`.

Stage1 deals with this by not having that error when switching on errors but I think it would be better to detect and allow this specific pattern.

Vexu:
This gets `zig2 build test-std` to a point where it builds and runs successfully, but fails 39 tests.

Stage1 `2139 passed; 33 skipped; 0 failed.` vs stage2 `2016 passed; 51 skipped; 39 failed.`

Closes #11798


andrewrk (approved):
:partying_face:

PR #11815, opened 2022-06-07
std.math: Add O(log N) implementation of `log2(x)` for `comptime_int`

topolarity:
Since `@clz` is undefined for `comptime_int` and Zig does not provide `@fls` (find-last-set), `log2` for comptime integers needs to be computed algorithmically.

This change replaces the existing implementation with a basic binary search

The benefit is that the added tests no longer hit the default backward branch quota of 1000 :+1:

### DIFF
--- a/lib/std/math/log2.zig
+++ b/lib/std/math/log2.zig
@@ -17,12 +17,20 @@ pub fn log2(x: anytype) @TypeOf(x) {
         },
         .Float => return @log2(x),
         .ComptimeInt => comptime {
-            var result = 0;
             var x_shifted = x;
-            while (b: {
-                x_shifted >>= 1;
-                break :b x_shifted != 0;
-            }) : (result += 1) {}
+            // First, calculate floorPowerOfTwo(x)
+            var shift_amt = 1;
+            while (x_shifted >> (shift_amt << 1) != 0) shift_amt <<= 1;
+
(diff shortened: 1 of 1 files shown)


andrewrk (approved):
Nice!

jibal:
> The benefit is that the added tests no longer hit the default backward branch quota of 1000 👍

Is something going to be done about that misfeature? I gave up on a comptime-heavy implementation of ranges in part because of it (and other comptime issues that I'm hoping will be improved with Stage2). I can't even set a large global limit because some stdlib functions arbitrarily set it back to low values.

andrewrk:
> set it back to low values.

https://ziglang.org/documentation/master/#setEvalBranchQuota

> If the new_quota is smaller than the default quota (1000) or a previously explicitly set quota, it is ignored.

jibal:
Is that new? Because it wasn't my experience. In any case, I think a global value is a bad idea ... unless making it ridiculously large, it can make programs suddenly stop compiling due to small changes.

andrewrk:
It's not global

jibal:
I must be seriously misunderstanding something. If it's not global, then what is its scope? I've seen proposals to make it scoped or add a variant that is scoped, but the documentation you just pointed me to doesn't say anything about scope.

PR #11819, opened 2022-06-08
introduce std.debug.Trace and use it to debug a LazySrcLoc in stage2 that is set to a bogus value

andrewrk:
The actual fix here is:

```diff
-        try sema.emitBackwardBranch(&child_block, call_src);
+        try sema.emitBackwardBranch(block, call_src);
```

Here is some output from using std.debug.Trace on the LazySrcLoc that helped me track down where the problem was:

```
[nix-shell:~/dev/zig/build-release]$ stage2/bin/zig test test3.zig 
init:
/home/andy/dev/zig/src/Zir.zig:2441:45: 0x31b8ba5 in Zir.struct:2432:18.src (zig)
                return LazySrcLoc.nodeOffset(self.src_node);
                                            ^
/home/andy/dev/zig/src/Sema.zig:5030:35: 0x339fde6 in Sema.zirCall (zig)
    const call_src = inst_data.src();
                                  ^
/home/andy/dev/zig/src/Sema.zig:712:62: 0x322dbe7 in Sema.analyzeBodyInner (zig)
            .call                         => try sema.zirCall(block, inst),
                                                             ^
thread 18654 panic: index out of bounds
/home/andy/dev/zig/src/Module.zig:2115:46: 0x2f76876 in Module.SrcLoc.byteOffset (zig)
                const tok_index = main_tokens[node];
                                             ^
/home/andy/dev/zig/src/Compilation.zig:451:66: 0x2f705f6 in Compilation.AllErrors.add (zig)
        const byte_offset = try module_err_msg.src_loc.byteOffset(module.gpa);
                                                                 ^
/home/andy/dev/zig/src/Compilation.zig:2568:38: 0x2dd1aa3 in Compilation.getAllErrorsAlloc (zig)
                    try AllErrors.add(module, &arena, &errors, entry.value_ptr.*.*);
                                     ^
/home/andy/dev/zig/src/main.zig:3121:44: 0x2d3bf6a in updateModule (zig)
    var errors = try comp.getAllErrorsAlloc();
                                           ^
/home/andy/dev/zig/src/main.zig:2808:17: 0x2d04a59 in buildOutputType (zig)
                ^
/home/andy/dev/zig/src/main.zig:225:31: 0x2cea926 in mainArgs (zig)
                              ^
/home/andy/dev/zig/src/main.zig:174:20: 0x2ce9974 in main (zig)
                   ^
/home/andy/dev/zig/lib/std/start.zig:581:37: 0x31d5bf7 in std.start.callMain (zig)
                                    ^
/home/andy/dev/zig/lib/std/start.zig:515:12: 0x2cec3c7 in std.start.callMainWithArgs (zig)
           ^
/home/andy/dev/zig/lib/std/start.zig:480:12: 0x2cec172 in std.start.main (zig)
           ^
Aborted (core dumped)
```

This pinpointed where the bad LazySrcLoc was created:

```
/home/andy/dev/zig/src/Sema.zig:5030:35: 0x339fde6 in Sema.zirCall (zig)
    const call_src = inst_data.src();
                                  ^
```

So then it was a simple matter of noticing where it got sent into a compile error, which was at the call to `emitBackwardBranch`. This was due to the return type of the function having function bodies in it, so solving the compile error was a matter of making them function pointers instead.

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -659,7 +659,7 @@ pub const Decl = struct {
     }
 
     pub fn nodeSrcLoc(decl: Decl, node_index: Ast.Node.Index) LazySrcLoc {
-        return .{ .node_offset = decl.nodeIndexToRelative(node_index) };
+        return LazySrcLoc.nodeOffset(decl.nodeIndexToRelative(node_index));
     }
 
     pub fn srcLoc(decl: Decl) SrcLoc {
@@ -670,7 +670,7 @@ pub const Decl = struct {
         return .{
             .file_scope = decl.getFileScope(),
             .parent_decl_node = decl.src_node,
-            .lazy = .{ .node_offset = node_offset },
+            .lazy = LazySrcLoc.nodeOffset(node_offset),
         };
     }
 
@@ -861,7 +861,7 @@ pub const ErrorSet = struct {
         return .{
             .file_scope = owner_decl.getFileScope(),
             .parent_decl_node = owner_decl.src_node,
-            .lazy = .{ .node_offset = self.node_offset },
+            .lazy = LazySrcLoc.nodeOffset(self.node_offset),
         };
     }
 
@@ -947,7 +947,7 @@ pub const Struct = struct {
         return .{
             .file_scope = owner_decl.getFileScope(),
             .parent_decl_node = owner_decl.src_node,
-            .lazy = .{ .node_offset = s.node_offset },
+            .lazy = LazySrcLoc.nodeOffset(s.node_offset),
         };
     }
 
@@ -1066,7 +1066,7 @@ pub const EnumSimple = struct {
         return .{
             .file_scope = owner_decl.getFileScope(),
             .parent_decl_node = owner_decl.src_node,
-            .lazy = .{ .node_offset = self.node_offset },
+            .lazy = LazySrcLoc.nodeOffset(self.node_offset),
         };
     }
 };
@@ -1097,7 +1097,7 @@ pub const EnumNumbered = struct {
         return .{
             .file_scope = owner_decl.getFileScope(),
             .parent_decl_node = owner_decl.src_node,
-            .lazy = .{ .node_offset = self.node_offset },
+            .lazy = LazySrcLoc.nodeOffset(self.node_offset),
         };
     }
 };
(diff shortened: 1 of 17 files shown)


andrewrk:
failed test is recursive_inline_function, looks like it is now printing a big error note "called from here" stack trace

src/Module.zig
```diff
@@ -2515,6 +2525,17 @@ pub const SrcLoc = struct {
...
 
+/// This wraps a simple integer in debug builds so that later on we can find out
+/// where in semantic analysis the value got set.
+const TracedOffset = struct {
+    x: i32,
+    trace: Trace = trace_init,
+
+    const want_tracing = builtin.mode == .Debug;
+    const trace_init = if (want_tracing) std.debug.Trace(1, 3){} else {};
+    const Trace = @TypeOf(trace_init);
```
Vexu, line 2536: It feels like this would be a common enough use case to justify adding a wrapper next to `std.debug.Trace` so that users don't have to reimplement it every time.

Vexu (approved):

src/Module.zig
andrewrk: The whole thing? I was actually thinking of changing it to be `const want_tracing = build_options.tracing;`, depending on a `-Dtracing` flag passed to `zig build`. Also the choice of 1 trace and 3 stack frames is a bit application-specific. But I see your point :thinking: :thinking: 

OK I think we can provide such a thing with decent defaults, and we can improve on it too - instead of `void` it could be an empty struct that provides the same methods, which are no-ops.

Thanks for the push!

Vexu: > The whole thing?

I meant something like `fn DebugOnlyTrace(comptime size: usize, comptime stack_frame_count: usize) type` which would have the same methods just like you mentioned.

> I was actually thinking of changing it to be `const want_tracing = build_options.tracing;`, depending on a `-Dtracing` flag passed to `zig build`.

Hmm, I guess we could also have `fn Trace(comptime size: usize, comptime stack_frame_count: usize) type` and `fn OptionalTrace(comptime size: usize, comptime stack_frame_count: usize, comptime enabled: bool) type` where `Trace` just returns an `OptionalTrace` with `enable` being `mode == .Debug`?

andrewrk:
Screenshot so you can see the terminal colors:

image

kubkon (approved):
Very exciting! Cannot wait to start using it!

andrewrk:
Argh, it looks like f1cff4fa4a28d42ac9055f94ee9a8f7fd2831cd7 did not solve the flaky failure. I have not diagnosed that failure yet, but it is not related to this PR.

PR #11820, opened 2022-06-08
Add initial support for macOS Ventura (13.0)

haze:
+ Adds libSystem.13.tbd (copied from Xcode 14 beta TBD)
+ Adds macOS 13 targets for both x86_64 & aarch64


andrewrk:
Thanks for the patch. Unfortunately - and this is not your fault - it's impossible to properly review, so a core team member will need to perform the [official steps](https://github.com/ziglang/zig/wiki/Updating-libc#macos) for updating in order to get Ventura support.

PR #11824, opened 2022-06-08
parser: sync comments with formal grammar

ifreund:


andrewrk:
Thanks!

You looking for some kind of review on this? You are welcome to push directly to master, at your discretion.

ifreund:
> You looking for some kind of review on this? You are welcome to push directly to master, at your discretion.

Not really looking for a review, just not used to pushing to master directly because I usually want CI to be green first. In this case it's only changes to comments though so...

andrewrk:
That's fair! Yeah it's a bit annoying when CI fails on master, but the ziglang.org binaries won't update unless everything is green, so, it's not the end of the world.

PR #11825, opened 2022-06-08
Enable full RELRO by default, expose in std.build

### LINKED ISSUE #6977: Support RELRO

ifreund:
Full RELRO is a hardening feature that makes it impossible to perform
certian attacks involving overwriting parts of the Global Offset Table
to invoke arbitrary code.

It requires all symbols to be resolved before execution of the program
starts which may have an impact on startup time. However most if
not all popular Linux distributions enable full RELRO by default for
all binaries and this does not seem to make a noticeable difference
in practice.

"Partial RELRO" is equivalent to `-z relro -z lazy`.
"Full RELRO" is equivalent to `-z relro -z now`.

LLD defaults to `-z relro -z lazy`, which means Zig's current `-z relro`
option has no effect on LLD's behavior.

The changes made by this commit are as follows:

- Document that `-z relro` is the default and add `-z norelro`.
- Pass `-z now` to LLD by default to enable full RELRO by default.
- Add `-z lazy` to disable passing `-z now`.

This was prompted by Arch Linux maintainers asking me how to enable RELRO for my software.
Closes #6977


daurnimator:
> Closes #6977

I think this only partially closes it? Wouldn't we also need to update our loader (when not linking libc?)

ifreund:
> I think this only partially closes it? Wouldn't we also need to update our loader (when not linking libc?)

Do you mean `std.DynLib`? It doesn't look like `ElfDynLib` supports relocations in the first place or for that matter much of anything.

I'm not sure it makes sense for `ElfDynLib` to exist in the first place to be honest. I don't think dynamic linking without linking libc makes sense. That leaves the use-case of using `ElfDynLib` to dynamically load symbols from a statically linked binary, which I think makes even less sense. Here's a mail with [Rick Felker's thoughts on the subject](https://www.openwall.com/lists/musl/2012/12/08/4) (the primary author of musl libc).

andrewrk (approved):

andrewrk:
ElfDynLib is old and experimental and will probably get removed from the standard library

OfekShilon:
Why not build with `-no-plt` instead? Always seemed funny to me to erect the entire PLT apparatus to support lazy loading and then populate it at program startup.  `-no-plt` is documented to be more efficient, and achieves the same effect (unless I'm missing some tradeoff)

PR #11826, opened 2022-06-08
linker: remove `-z noexecstack` option

ifreund:
Note that the current documentation for the `-z noexecstack` is
incorrect. This indicates that an object *does not* require an
executable stack.

This is actually the default of LLD, and there has never been a way to
override this default by passing `-z execstack` to LLD.

This commit removes the redundant `-z noexecstack` option from
zig build-exe/build-lib/build-obj and ignores the option if passed
to zig cc for compatibility.

As far as I can tell, there is no reason for code to require an
executable stack. This option only exists because the stack was
originally executable by default and some programs came to depend
on that behavior. Instead, mprotect(2) may be used to make memory
pages executable.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -762,7 +762,6 @@ pub const InitOptions = struct {
     linker_z_notext: bool = false,
     linker_z_defs: bool = false,
     linker_z_origin: bool = false,
-    linker_z_noexecstack: bool = false,
     linker_z_now: bool = true,
     linker_z_relro: bool = true,
     linker_z_nocopyreloc: bool = false,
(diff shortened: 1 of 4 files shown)


kubkon:
SGTM. The only usecase I could think of that could potentially benefit from this flag would be something like hot-code reloading, but for this we use our in-house linker anyhow :-)

kubkon (approved):

ifreund:
> The only usecase I could think of that could potentially benefit from this flag would be something like hot-code reloading

Would the `mprotect()` syscall not  be sufficient for this use case?

kubkon:
> > The only usecase I could think of that could potentially benefit from this flag would be something like hot-code reloading
> 
> 
> 
> Would the `mprotect()` syscall not  be sufficient for this use case?

Potentially, but correct me if I'm wrong, mprotect does copy-on-write or similar on the modified pages, right? In this case, it might be more efficient to simply make executable segment writeable straight away. That's what Andrew did in his HCS PoC anyhow. It's not relevant to this PR as we will most likely like to exert strict control on what and when is writable and probably have this on only for the HCS mode.

PR #11827, opened 2022-06-08
stage2: enhancements aimed at more behavior tests passing

andrewrk:
See individual commit messages for more information.

### DIFF
--- a/test/behavior/bugs/6781.zig
+++ b/test/behavior/bugs/6781.zig
@@ -1,3 +1,4 @@
+const builtin = @import("builtin");
 const std = @import("std");
 const assert = std.debug.assert;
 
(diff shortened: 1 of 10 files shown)


test/behavior/bugs/6781.zig
```diff
@@ -61,9 +62,13 @@ pub const JournalHeader = packed struct {
...
 test "fixed" {
-    if (@import("builtin").zig_backend != .stage1) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_wasm) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_c) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_x86_64) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_arm) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_aarch64) return error.SkipZigTest; // TODO
 
-    var buffer = [_]u8{0} ** 65536;
+    var buffer align(@alignOf(JournalHeader)) = [_]u8{0} ** 65536;
```
Jarred-Sumner, line 71: `align` goes before the type? 🤯

kubkon: Yeah, it's separate to the type. Internally, we store explicit alignment directly on the decl, and then call `Decl.getAlignment` which first tries resolving explicit alignment if any, to fall back to type's ABI alignment otherwise. It's a pretty clever design and kudos to @andrewrk for coming up with this mechanism :-)

jibal: > `align` can appear without the type? 

? Align takes a power of 2, which is given here. The type of `buffer` is inferred from the initializer to be `[65536]u8`

Issue #10777, opened 2022-02-03
Presence of `gcc-cross` tools breaks cross-compilation with `zig c++`


PiotrSikora:
### Zig Version

0.10.0-dev.528+f95fcb2b1

### Steps to Reproduce

```shell
$ cat hello.cpp 
```
```cpp
#include <iostream>

int main() {
    std::cout << "Hello, World!" << std::endl;
    return 0;
}
```

Install `gcc-cross` tools for `aarch64` (this breaks the build)

```shell
$ sudo apt install -y g++-aarch64-linux-gnu
```

### Expected Behavior

```shell
$ rm -rf ~/.cache/zig && zig c++ -v -target aarch64-linux-musl hello.cpp && file ./a.out
```
```
clang version 13.0.0 (https://github.com/ziglang/zig-bootstrap.git 4ed79aefb7a58a6d642f47a81e1ef04fd164042b)                                                                                                                                              
Target: aarch64-unknown-linux-musl                                                                                                                                                                                                                        
Thread model: posix                                                                                                                                                                                                                                       
InstalledDir: /usr/lib/llvm-13/bin                                                                                                                                                                                                                        
[...]
```
```
a.out: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, stripped
```

### Actual Behavior

```shell
$ rm -rf ~/.cache/zig && zig c++ -v -target aarch64-linux-musl hello.cpp && file ./a.out
```
```
clang version 13.0.0 (https://github.com/ziglang/zig-bootstrap.git 4ed79aefb7a58a6d642f47a81e1ef04fd164042b)
Target: aarch64-unknown-linux-musl
Thread model: posix
InstalledDir: /usr/lib/llvm-13/bin
Found candidate GCC installation: /usr/lib/gcc-cross/aarch64-linux-gnu/11
Selected GCC installation: /usr/lib/gcc-cross/aarch64-linux-gnu/11
Candidate multilib: .;@m64
Selected multilib: .;@m64
[...]
```
```
ld.lld: error: undefined symbol: __aarch64_ldadd8_acq_rel
>>> referenced by cxa_exception.cpp
[...]

PR #11828, opened 2022-06-09
compiler_rt: aarch64 outline atomics

### LINKED ISSUE #10086: compiler-rt does not define aarch64 outline atomics
### Zig Version

0.9.0-dev.1551+8346e011c

### Steps to Reproduce

On aarch64 linux:

```c
#include <stdatomic.h>
#include <stdio.h>

int main() {
  atomic_int atomic_counter = 0;
  atomic_counter++;
  printf("%u\n", atomic_counter);
}
```

1. `gcc main.c`
2. `zig cc main.c`
3. `gcc -moutline-atomics main.c`
4. `zig cc -moutline-atomics main.c`

### Expected Behavior

I expect all variations to compile and run correctly.

### Actual Behavior

`zig cc -moutline-atomics main.c` fails to compile:

```
ld.lld: error: undefined symbol: __aarch64_ldadd4_acq_rel
```

I discovered this issue while attempting to cross compile Rust to aarch64-linux with `zig cc`. Rust enables outline atomics by default on aarch64-linux and currently [prohibits disabling them](https://github.com/rust-lang/rust/blob/56694b04532cc0dec6e2f577135da3513e856923/compiler/rustc_codegen_llvm/src/llvm_util.rs#L405-L411).

devins2518:
[Ported from llvm's port of libgcc's outline atomics](https://reviews.llvm.org/rGa4ac434c47434d80bca54bab96f295ed4e972cc6).

~~Includes #11667 though, I can remove if this pr is merged before hazes.~~ Removed

Fixes #10086 and [`zig c++` as noted here](https://github.com/ziglang/zig/issues/10086#issuecomment-1013979896).

# Issues
1. ~~The code is admittedly pretty ugly. There's a lot of inline assembly created through a mixture of multiline strings and regular string concatenation which is not ideal. Not sure if the level of comptime variable parsing is possible through just using inputs for inline assembly.~~
2. Most of this is not tested. `__aarch64_ldadd4_acq_rel` is tested as it was the test case, but I'm not too sure about all of the other functions. 
3. ~~`__aarch64_have_lse_atomics` is constructed through a function in the `.ctors` section which is definitely hidden control flow. Would be interesting to see any more ziggy solutions to fixing this.~~ `__init_aarch64_have_lse_atomics` needs to be linked into either `__DATA,__mod_init_func` or `.init_array` depending on platform, but `@export` section handling isn't implemented. Right now, it is just called at the beginning of each outlined function.
4. `initHaveLseAtomics` also calls `std.os.linux.getauxval` even on macOS, which doesn't seem right. This seems beyond the scope of this PR though.


devins2518:
`./build/zig build test-behavior -Dskip-release` fails without `-Dskip-non-native`, is this normal?

devins2518:
@PiotrSikora could you try this to see if it fixes #10777 ?

andrewrk:
@devins2518 thank you for implementing this. It took me a long time to get around to merging it but it finally landed in 271cc52a1619a553212192478d96a7c4bd629424. I made some edits to what you have here:

 * no more branching logic for has_lse. the same `.inst` assembler instruction can work with or without that feature. It's strange that the corresponding gcc code bothered to have that in there.
 * no more branching logic for darwin. libSystem already has these symbols, so no need to duplicate them in compiler_rt.a for darwin.
 * removed the auxval integration. zig's compiler_rt has a requirement of not using constructors. instead, we will export a function that can be called from third party start code to tell compiler_rt what CPU features are enabled. That will be a follow-up enhancement.
 * cleaned up the tools script to be more idiomatic and take advantage of arena allocation.

devins2518:
Thanks so much! It's so awesome to have my first meaningful contribution to zig merged 🎉! I've been planning to rewrite this a bit since #12756 was merged which would remove the need for the generation script and simplify `aarch64_outline_atomics.zig` a decent amount.

andrewrk:
Follow-up: #13622

PR #11832, opened 2022-06-09
llvm: explicitly annotate alias with linkage

kubkon:
Without explicitly applying the correct linkage attribute will
result in LLVM assuming the alias exported as strong/external. This is
particularly troublesome for static archives where we want to re-export
the same symbol under a different name but the same, weak linkage.
This can then cause weird behaviour when ordering of the static
archives with competing definition of the same symbol may lead to
either no error, or symbol collision error, depending on the linking
order.

A short example is provided in `test/standalone/link_weak_reexports`
where if `liba.a` exports `div2` with strong linkage, and exports
`__pow` with weak linkage, but also aliases `__pow` as `powq`, and
if `libb.a` exports `powq` as strong, then the final program that
  requires presence of both symbols `div2` and `powq` will correctly
  link if `liba.a` preceeds `libb.a` on the linker line, and lead to
  error if they are flipped.

Another issue with this is that we will never link the definition
from `libb.a` unless we drop `liba.a` altogether. We actually
have this very problem in Zig when trying to link musl and use
musl's implementation in favour of compiler-rt's. This is currently
not possible as we always order `libcompiler_rt.a` as first on
the linker line, while flipping the order so that `libc.a` comes
before `libcompiler_rt.a` leads to linker errors.

### DIFF
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -1114,11 +1114,17 @@ pub const Object = struct {
                 if (self.llvm_module.getNamedGlobalAlias(exp_name_z.ptr, exp_name_z.len)) |alias| {
                     alias.setAliasee(llvm_global);
                 } else {
-                    _ = self.llvm_module.addAlias(
+                    const alias = self.llvm_module.addAlias(
                         llvm_global.typeOf(),
                         llvm_global,
                         exp_name_z,
                     );
+                    switch (exp.options.linkage) {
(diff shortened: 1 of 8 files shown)


kubkon:
Closing as I reverted this change back in 372e9709ad2f4af24461f0fa601754a469091c2b. Tl;dr aliases are difficult to work with and inconsistent between backends in LLVM.

Issue #10886, opened 2022-02-14
use-after-free of `CaptureScope` when error occurs in `Module.semaDecl`


andrewrk:
### Zig Version

0.10.0-dev.700+2a73700c0

### Steps to Reproduce

test.zig

```
const std = @import("std");

test "example" {
    const x = try std.testing.allocator.create(i32);
    _ = x;
}

```

```
zig test test.zig -fLLVM
```

### Expected Behavior

Test passed

### Actual Behavior

The issue manifests like this:

```
thread 1329725 panic: integer overflow
Analyzing /home/andy/dev/zig/lib/std/heap/general_purpose_allocator.zig: heap/general_purpose_allocator.zig:GeneralPurposeAllocator__anon_113.usingnamespace_0
    > %1142 = closure_get(%170) 
      %1143 = field_val(%1142, "retain_metadata") 
      %1144 = as_node(@Ref.bool_type, %1143) 
      %1145 = condbr(%1144, {
        %1147 = extended(struct_decl(anon, Auto, {
          [2386] pub flushRetainedMetadata line(1) hash(01dadb0b0b6a76cdb56612b68cb32949): %1148 = ...
        }, {}, {}) 
        %1200 = as_node(@Ref.type_type, %1147) 
        %1203 = break(%1146, %1200)
      }, {
        %1201 = extended(struct_decl(anon, Auto, {}, {}, {}) 
        %1202 = as_node(@Ref.type_type, %1201) 
        %1204 = break(%1146, %1202)
      }) 
    For full context, use the command
      zig ast-check -t /home/andy/dev/zig/lib/std/heap/general_purpose_allocator.zig

  in /home/andy/dev/zig/lib/std/heap/general_purpose_allocator.zig: heap/general_purpose_allocator.zig:GeneralPurposeAllocator__anon_113.usingnamespace_0
    > %1146 = block({%1142..%1145}) 

/home/andy/dev/zig/lib/std/hash_map.zig:1134:42: 0x3815609 in std.hash_map.HashMapUnmanaged(u32,TypedValue,std.hash_map.AutoContext(u32),80).getPtrAdapted (zig)
            const mask = self.capacity() - 1;
                                         ^
/home/andy/dev/zig/lib/std/hash_map.zig:1233:38: 0x3662b85 in std.hash_map.HashMapUnmanaged(u32,TypedValue,std.hash_map.AutoContext(u32),80).getPtrContext (zig)
            return self.getPtrAdapted(key, ctx);
[...]

PR #11837, opened 2022-06-10
Fix (nearly) all stage2 crashes when testing stdlib

Vexu:
After this passing all std tests will "only" require a bunch of miscompilation fixes, implementing async, and packed struct related adjustments.
Probably worth noting that removing stage2 function pointer handling from the compress related structs seems to cause #10886.

Should I disable the 40 failing tests and add `test-std` to the CI?

### DIFF
--- a/lib/std/compress/deflate/compressor_test.zig
+++ b/lib/std/compress/deflate/compressor_test.zig
@@ -179,7 +179,6 @@ test "deflate/inflate" {
 }
 
 test "very long sparse chunk" {
-    if (@import("builtin").zig_backend != .stage1) return error.SkipZigTest;
     // A SparseReader returns a stream consisting of 0s ending with 65,536 (1<<16) 1s.
     // This tests missing hash references in a very large input.
     const SparseReader = struct {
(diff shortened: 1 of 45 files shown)


lib/std/compress/deflate/compressor_test.zig
```diff
@@ -397,8 +396,7 @@ test "deflate/inflate string" {
     // Skip wasi because it does not support std.fs.openDirAbsolute()
     if (builtin.os.tag == .wasi) return error.SkipZigTest;
```
zeroZshadow, line 397: This this skip still stay now that `std.fs.openDirAbsolute()` is no longer used here?

andrewrk:
> Should I disable the 40 failing tests and add test-std to the CI?

Yes. It's time.

andrewrk:
> CI sourcehut: cd out of build before running tests

What's the motivation here? All `zig build` commands are supposed to work from any sub-directory.

Vexu:
> What's the motivation here? All `zig build` commands are supposed to work from any sub-directory.

I thought `@src` in stage2 returns `--main-pkg-path` relative paths but with further testing it seems relative to the Zig file given as input and the correct fix is to make it always return absolute paths.

PR #11847, opened 2022-06-11
compiler_rt: compile each unit separately for improved archiving

kubkon:
As discussed during our last stage2 meeting, this PR is reshuffling compiler-rt so that we can build it the `libcompiler_rt.a` from multiple object files rather than just one big `compiler_rt.o`. This will improve linking times and binary sizes as the linker will only need to parse the relevant object files from the `libcompiler_rt.a` archive.

@andrewrk could you have a look and lemme know what you think?


lib/compiler_rt/mulXf3.zig
```diff
@@ -5,8 +5,32 @@
...
+const arch = builtin.cpu.arch;
+const is_test = builtin.is_test;
+const linkage: std.builtin.GlobalLinkage = if (builtin.is_test) .Internal else .Weak;
+pub const panic = @import("common.zig").panic;
+
+comptime {
+    @export(__mulsf3, .{ .name = "__mulsf3", .linkage = linkage });
+    @export(__muldf3, .{ .name = "__muldf3", .linkage = linkage });
+    @export(__mulxf3, .{ .name = "__mulxf3", .linkage = linkage });
+    @export(__multf3, .{ .name = "__multf3", .linkage = linkage });
```
andrewrk, line 17: do we want these also in separate compilation units? looks like right now these will all go into mulXf3.o.

src/compiler_rt.zig
```diff
@@ -0,0 +1,186 @@
...
+        .local_cache_directory = comp.global_cache_directory,
+        .global_cache_directory = comp.global_cache_directory,
+        .zig_lib_directory = comp.zig_lib_directory,
+        .cache_mode = .whole,
+        .target = target,
+        .root_name = root_name,
+        .main_pkg = null,
+        .output_mode = .Lib,
+        .link_mode = .Static,
+        .function_sections = true,
```
andrewrk, line 82: do we still want this now that all the functions are in separate object files?

kubkon: I think so. Here's why I think so. Since we bundle up a couple related symbols together in one object file, we probably want this on for the linker to GC unused symbol versions.

lib/compiler_rt/mulXf3.zig
kubkon: I think that's OK. It's only 4 symbols that share functionality just accept different operand types. I actually think we should split out `addXf3.zig` into `addXf3.zig` and `subXf3.zig`, and combine `divsf3.zig`, `divtf3.zig`, `divdf3.zig`, `divxf3.zig`, and `divti3.zig` into `divXf3.zig`. Thoughts?

andrewrk:
I made some progress on this but my work is still very much incomplete. Here are my notes for what I need to do next:

 * break the functions up even more
   - aulldiv.zig
   - aullrem.zig
   - bswap.zig
   - ceil.zig
   - cmp.zig
   - cos.zig
   - count0bits.zig
   - divdf3.zig
   - divsf3.zig
   - divtf3.zig
   - divti3.zig
   - divxf3.zig
   - emutls.zig
   - exp.zig
   - exp2.zig
   - fabs.zig
   - floor.zig
   - fma.zig
   - fmax.zig
   - fmin.zig
   - fmod.zig
   - int.zig
   - log.zig
   - log10.zig
   - log2.zig
   - modti3.zig
   - muldi3.zig
   - mulo.zig
   - multi3.zig
   - negXf2.zig
   - negXi2.zig
   - negv.zig
   - parity.zig
   - popcount.zig
   - round.zig
   - shift.zig
   - sin.zig
   - sincos.zig
   - sqrt.zig
   - subo.zig
   - tan.zig
   - trunc.zig
   - udivmodti4.zig
   - udivti3.zig
   - umodti3.zig
 * break up sparc.zig
 * break up arm.zig
 * no more -ffunction-sections
 * multi-thread it
 * add to src/compiler_rt.zig
 * add to CMakeLists.txt
 * rip out Cache usage
 * audit the PR

lib/compiler_rt/mulXf3.zig
andrewrk: I think the different types are especially useful to keep separate because typically a target will have support for one operand type but not another. For example x86_64 is going to need functions with f128 operand types, but it will not need any functions with f32 or f64 operand types. So putting f32, f64, f128 all into the same compilation unit is going to create garbage.

andrewrk:
After doing the multi-threading commit I was able to take a performance measurement, despite not quite ironing out all the bugs yet:

```
debug stage2 building hello world: branch vs master
cold cache: 0m5.512s vs 0m3.813s
warm cache: 0m0.836s vs 0m0.812s
```

A bit concerning. However there are some things we can play with:
 * Special-case the cache to avoid overhead of cache hits. For example, if we are willing to give up the feature that editing any compiler-rt source file will cause a cache miss in compiler-rt, then we can reduce one fstat() call per source file. Note however that warm cache is roughly the same time, within error tolerance.
 * Process-level file cache to avoid O(N*M) AstGen operations (even though most operations are cache hits), where N is number of compilation units inside compiler-rt and M is the number of .zig files in the standard library and compiler-rt combined.

I put in a little commit to disable the multi-threading and separate compilation units of compiler-rt, effectively undoing the work of this branch, and got these results:

```
debug stage2 building hello world: branch vs master
cold cache: 0m3.700s vs 0m3.647s
warm cache: 0m0.815s vs 0m0.784s
```

Both numbers are the same time within error tolerance. Point being, this would allow us to merge the branch, and things would be roughly the same as master branch as far as `-ffunction-sections` and separate compilation units. However it would then be just a tiny amount of effort to toggle to the separate compilation unit approach, and play with it.

My notes for what to do next are:

 * process-local AstGen cache so that we don't redundantly grab ZIR for the whole
   std lib redundantly many times.
 * integrate std.Progress into Compilation API
 * break the functions up even more
   - aulldiv.zig
   - aullrem.zig
   - bswap.zig
   - ceil.zig
   - cmp.zig
   - cos.zig
   - count0bits.zig
   - divdf3.zig
   - divsf3.zig
   - divti3.zig
   - divxf3.zig
   - emutls.zig
   - exp.zig
   - exp2.zig
   - fabs.zig
   - floor.zig
   - fma.zig
   - fmax.zig
   - fmin.zig
   - fmod.zig
   - int.zig
   - log.zig
   - log10.zig
   - log2.zig
   - modti3.zig
   - muldi3.zig
   - mulo.zig
   - multi3.zig
   - negXf2.zig
   - negXi2.zig
   - negv.zig
   - parity.zig
   - popcount.zig
   - round.zig
   - shift.zig
   - sin.zig
   - sincos.zig
   - sqrt.zig
   - subo.zig
   - tan.zig
   - trunc.zig
   - udivmodti4.zig
   - udivti3.zig
   - umodti3.zig
 * break up arm.zig
 * build for sparc
 * build for arm
 * build for ppc
 * update CMakeLists.txt using find
 * audit the PR
 * -fcompiler-rt when building a static library should include the
   object files inside the archive instead of a single compiler_rt.o

andrewrk:
Tested a release build with the (now disabled) multi-threaded compiler-rt commit.

Warm cache: the cost of this change is 38ms on my machine:

```
[nix-shell:~/tmp]$ hyperfine '/home/andy/dev/zig/build/stage2-release/bin/zig build-exe hello.zig' '/home/andy/Downloads/zig/build-release/stage2-release/bin/zig build-exe hello.zig' --warmup 3
Benchmark 1: /home/andy/dev/zig/build/stage2-release/bin/zig build-exe hello.zig
  Time (mean ± σ):     554.6 ms ±  14.1 ms    [User: 542.2 ms, System: 72.2 ms]
  Range (min … max):   542.7 ms … 581.0 ms    10 runs
 
Benchmark 2: /home/andy/Downloads/zig/build-release/stage2-release/bin/zig build-exe hello.zig
  Time (mean ± σ):     516.6 ms ±  18.2 ms    [User: 494.4 ms, System: 65.1 ms]
  Range (min … max):   499.8 ms … 547.2 ms    10 runs
 
Summary
  '/home/andy/Downloads/zig/build-release/stage2-release/bin/zig build-exe hello.zig' ran
    1.07 ± 0.05 times faster than '/home/andy/dev/zig/build/stage2-release/bin/zig build-exe hello.zig'
```

Cold cache: the cost of this change is roughly 1.4s on my machine (couldn't use hyperfine due to race condition deadlock bug).

I can explore some of these optimization opportunities, but generally, it seems to me that the cost of making so many compilation units is not outweighing the benefit of less work for the linker.

@kubkon thoughts?

kubkon:
I think I misjudged the balance between parsing a single larger object file vs extracting a tiny object file from multi-object archive. The added complexity of splitting `compiler-rt` into multiple translation units might not offset the marginal gain in link times for `lld` or `zld`.

Here's the test source I used (I forced the inclusion of a single symbol from `compiler-rt` on my x64 AMD Ryzen7):

```zig
const std = @import("std");

pub fn main() anyerror!void {
    var result: f128 = @sin(@as(f128, 2.0));
    const stderr = std.io.getStdErr();
    try stderr.writer().print("Hello, {}!\n", .{result});
}
```

Here are link-time benchmarks for `lld` and `zld` with `--gc-sections` on for multi-object archive and one-shot (status quo):

---
### `lld`

__link time__
```
$ hyperfine 'ld.lld -O0 -static -z stack-size=16777216 -o hello_scrambled --gc-sections scrambled/hello.o scrambled/libc.a scrambled/libcompiler_rt.a' 'ld.lld -O0 -static -z stack-size=16777216 -o hello_one_shot --gc-sections one-shot/hello.o one-shot/libc.a one-shot/libcompiler_rt.a' --warmup 3
Benchmark 1: ld.lld -O0 -static -z stack-size=16777216 -o hello_scrambled --gc-sections scrambled/hello.o scrambled/libc.a scrambled/libcompiler_rt.a
  Time (mean ± σ):      12.8 ms ±   1.0 ms    [User: 6.7 ms, System: 8.0 ms]
  Range (min … max):    11.3 ms …  21.5 ms    194 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Benchmark 2: ld.lld -O0 -static -z stack-size=16777216 -o hello_one_shot --gc-sections one-shot/hello.o one-shot/libc.a one-shot/libcompiler_rt.a
  Time (mean ± σ):      12.7 ms ±   0.8 ms    [User: 6.8 ms, System: 8.0 ms]
  Range (min … max):    11.5 ms …  14.9 ms    190 runs
 
Summary
  'ld.lld -O0 -static -z stack-size=16777216 -o hello_one_shot --gc-sections one-shot/hello.o one-shot/libc.a one-shot/libcompiler_rt.a' ran
    1.01 ± 0.10 times faster than 'ld.lld -O0 -static -z stack-size=16777216 -o hello_scrambled --gc-sections scrambled/hello.o scrambled/libc.a scrambled/libcompiler_rt.a'
```

__size__
```
$ ls -la hello_scrambled hello_one_shot 
-rwxr-xr-x 1 kubkon users 1064312 Jun 17 10:25 hello_one_shot
-rwxr-xr-x 1 kubkon users 1064576 Jun 17 10:25 hello_scrambled
```

__TL;DR linking time and output binary sizes are virtually identical for `lld` __

---
### `zld`

Please note that `zld`'s support for GC sections is a WIP (I just started working on it) but it already garbage collects a fair amount of unused symbols, however it is not yet up to par with `lld`

__link time__
```
$ hyperfine 'zld -O0 -static -z stack-size=16777216 -o hello_scrambled --gc-sections scrambled/hello.o scrambled/libc.a scrambled/libcompiler_rt.a' 'zld -O0 -static -z stack-size=16777216 -o hello_one_shot --gc-sections one-shot/hello.o one-shot/libc.a one-shot/libcompiler_rt.a' --warmup 3
Benchmark 1: zld -O0 -static -z stack-size=16777216 -o hello_scrambled --gc-sections scrambled/hello.o scrambled/libc.a scrambled/libcompiler_rt.a
  Time (mean ± σ):       4.8 ms ±   0.6 ms    [User: 1.8 ms, System: 2.9 ms]
  Range (min … max):     3.8 ms …   7.5 ms    461 runs
 
  Warning: Command took less than 5 ms to complete. Results might be inaccurate.
 
Benchmark 2: zld -O0 -static -z stack-size=16777216 -o hello_one_shot --gc-sections one-shot/hello.o one-shot/libc.a one-shot/libcompiler_rt.a
  Time (mean ± σ):       5.5 ms ±   0.6 ms    [User: 2.1 ms, System: 3.3 ms]
  Range (min … max):     4.7 ms …   7.8 ms    359 runs
 
  Warning: Command took less than 5 ms to complete. Results might be inaccurate.
 
Summary
  'zld -O0 -static -z stack-size=16777216 -o hello_scrambled --gc-sections scrambled/hello.o scrambled/libc.a scrambled/libcompiler_rt.a' ran
    1.14 ± 0.19 times faster than 'zld -O0 -static -z stack-size=16777216 -o hello_one_shot --gc-sections one-shot/hello.o one-shot/libc.a one-shot/libcompiler_rt.a' 
```

__size__
```
$ ls -la hello_scrambled hello_one_shot
-rwxr-xr-x 1 kubkon users 1112248 Jun 17 10:28 hello_one_shot
-rwxr-xr-x 1 kubkon users 1070856 Jun 17 10:28 hello_scrambled
```

__TL;DR multi-object (scrambled) compiler-rt links a little faster and generates a significantly smaller binary size. Neither should be indicative yet as I'm still working on GC support in `zld`.__

## Conclusion

All in all, I don't think the added complexity and much longer compilation times of `compiler-rt` for cold cache are worth marginal improvements in link times and binary sizes for `lld`. If we were using `zld` as the default linker (or put it another way, if `zld` was feature-complete) then perhaps it would worth considering. But given that we will want `zld` to ultimately match `lld`'s performance, I don't think it is.

andrewrk:
Alright, apart from any CI failures that occur, I think this is merge-ready. Here's what ended up happening at the end of the day:

 * Some bugs regarding calling conventions of `__aeabi_*` functions are fixed.
 * Some functions which looked like they were supposed to be exported, but were missing `@export` calls are now exported.
 * Some cleanups are in place, such as not exporting redundant functions that will be unused by the selected target ABI, also a lot of DRYing up the place, making compiler-rt more maintainable.
 * A lot of work has gone into splitting up source files into independently compilable units. This work is not being taken advantage of since we are choosing to continue with the status quo of single-compilation-unit compiler-rt, however, should we want to experiment with this again in the future, 90% of the work is already done.

andrewrk:
Related issues filed:
 * #11883
 * #11884

andrewrk:
Alright we got a linker crash here:

```
[nix-shell:~/dev/zig/build-release]$ stage2/bin/zig test ../test/behavior.zig -I../test -target x86_64-macos
Semantic Analysis [5014] frexp64... LLVM Emit Object... thread 336027 panic: attempt to use null value
/home/andy/dev/zig/src/link/MachO.zig:3043:43: 0x2e5c0c5 in link.MachO.createMhExecuteHeaderAtom (zig)
        .seg = self.text_segment_cmd_index.?,
                                          ^
/home/andy/dev/zig/src/link/MachO.zig:971:43: 0x2e3bf2b in link.MachO.flushModule (zig)
        try self.createMhExecuteHeaderAtom();
                                          ^
/home/andy/dev/zig/src/link.zig:620:78: 0x2e1dfe8 in link.File.flushModule (zig)
            .macho => return @fieldParentPtr(MachO, "base", base).flushModule(comp, prog_node),
                                                                             ^
/home/andy/dev/zig/src/link.zig:795:33: 0x2e1b2dc in link.File.linkAsArchive (zig)
            try base.flushModule(comp, prog_node);
                                ^
/home/andy/dev/zig/src/link/MachO.zig:433:43: 0x2e1d5a4 in link.MachO.flush (zig)
            return self.base.linkAsArchive(comp, prog_node);
                                          ^
/home/andy/dev/zig/src/link.zig:605:72: 0x2e1a6be in link.File.flush (zig)
            .macho => return @fieldParentPtr(MachO, "base", base).flush(comp, prog_node),
                                                                       ^
/home/andy/dev/zig/src/Compilation.zig:2315:28: 0x2de2438 in Compilation.flush (zig)
    try comp.bin_file.flush(comp, prog_node); // This is needed before reading the error flags.
                           ^
/home/andy/dev/zig/src/Compilation.zig:2284:27: 0x2dd76cc in Compilation.update (zig)
            try comp.flush(main_progress_node);
                          ^
/home/andy/dev/zig/src/Compilation.zig:4982:31: 0x305ab44 in Compilation.buildOutputFromZig (zig)
    try sub_compilation.update();
                              ^
/home/andy/dev/zig/src/Compilation.zig:3510:28: 0x2df1cb4 in Compilation.buildCompilerRtOneShot (zig)
    comp.buildOutputFromZig("compiler_rt.zig", output_mode, out, .compiler_rt) catch |err| switch (err) {
                           ^
/home/andy/dev/zig/src/Compilation.zig:2841:31: 0x2dde3c6 in Compilation.performAllTheWork (zig)
        buildCompilerRtOneShot(comp, .Lib, &comp.compiler_rt_lib);
                              ^
/home/andy/dev/zig/src/Compilation.zig:2209:31: 0x2dd6c32 in Compilation.update (zig)
                              ^
/home/andy/dev/zig/src/main.zig:3126:20: 0x2d42faf in updateModule (zig)
    try comp.update();
                   ^
/home/andy/dev/zig/src/main.zig:2815:17: 0x2d0bc1e in buildOutputType (zig)
                ^
/home/andy/dev/zig/src/main.zig:225:31: 0x2cf19f6 in mainArgs (zig)
                              ^
/home/andy/dev/zig/src/main.zig:174:20: 0x2cf0a44 in main (zig)
                   ^
/home/andy/dev/zig/lib/std/start.zig:581:37: 0x31d1237 in std.start.callMain (zig)
                                    ^
/home/andy/dev/zig/lib/std/start.zig:515:12: 0x2cf3497 in std.start.callMainWithArgs (zig)
           ^
/home/andy/dev/zig/lib/std/start.zig:480:12: 0x2cf3242 in std.start.main (zig)
           ^
Aborted (core dumped)
```

kubkon:
The latest runtime crash:

```
34/235 compiler_rt.floor.test "compiler-rt-x86_64-macos-none-Debug--multi-default floor16"... Bus error at address 0x100a87db8
???:?:?: 0x100a87db8 in std.fmt.formatIntValue (test)
/Users/runner/work/1/s/lib/compiler_rt/floor.zig:171:24: 0x10047f465 in compiler_rt.floor.test "compiler-rt-x86_64-macos-none-Debug--multi-default floor16" (test)
    try expect(__floorh(1.3) == 1.0);
                       ^
/Users/runner/work/1/s/lib/test_runner.zig:79:28: 0x1005555a2 in (root).main (test)
        } else test_fn.func();
                           ^
/Users/runner/work/1/s/lib/std/start.zig:571:22: 0x100587b5c in std.start.callMain (test)
            root.main();
                     ^
/Users/runner/work/1/s/lib/std/start.zig:515:12: 0x100556f77 in std.start.callMainWithArgs (test)
           ^
/Users/runner/work/1/s/lib/std/start.zig:480:12: 0x100556eb5 in std.start.main (test)
           ^
???:?:?: 0x7fff6bb71cc8 in ??? (???)
???:?:?: 0x1 in ??? (???)
/Users/runner/work/1/s/zig-cache/o/5c4ef86ec311c84fa0710dc724207e10/test /Users/runner/work/1/s/build/release/bin/zig
```

sadly looks like a miscompilation. Initially I thought it's a problem with the linker, but I just double checked hooking up `ld64` instead of `zld` and the binary crashes in exactly the same place:

```
❯ ld -dynamic /Users/jakubkonka/dev/zig/zig-cache/o/06cadce8338708b080e797f8478b85f4/test.o /Users/jakubkonka/.cache/zig/o/816aa70679f0109a5542226475a84fd8/libcompiler_rt.a -o test_ld64 -lSystem -lc -syslibroot $(xcrun --show-sdk-path)
❯ ./test_ld64 stage1/bin/zig                                               
Test [34/235] compiler_rt.floor.test "compiler-rt-x86_64-macos-none-Debug--multi-defaul... Bus error at address 0x10060be98
???:?:?: 0x10060be98 in std.hash_map.HashMapUnmanaged([]const u8,ModuleDebugInfo.OFileInfo,std.hash_map.StringContext,80).grow (test)
/Users/jakubkonka/dev/zig/lib/compiler_rt/floor.zig:171:24: 0x100005a15 in compiler_rt.floor.test "compiler-rt-x86_64-macos-none-Debug--multi-default floor16" (test)
    try expect(__floorh(1.3) == 1.0);
                       ^
/Users/jakubkonka/dev/zig/lib/test_runner.zig:79:28: 0x1000db882 in (root).main (test)
        } else test_fn.func();
                           ^
/Users/jakubkonka/dev/zig/lib/std/start.zig:571:22: 0x10010dc7c in std.start.callMain (test)
            root.main();
                     ^
/Users/jakubkonka/dev/zig/lib/std/start.zig:515:12: 0x1000dd257 in std.start.callMainWithArgs (test)
           ^
/Users/jakubkonka/dev/zig/lib/std/start.zig:480:12: 0x1000dd195 in std.start.main (test)
           ^
???:?:?: 0x10368751d in ??? (???)
thread 1120777 panic: integer overflow
/Users/jakubkonka/dev/zig/lib/std/debug.zig:169:69: 0x100125c38 in std.debug.dumpStackTraceFromBase (test)
                                                                    ^
/Users/jakubkonka/dev/zig/lib/std/debug.zig:1854:35: 0x10011864f in std.debug.handleSegfaultPosix (test)
            dumpStackTraceFromBase(bp, ip);
                                  ^
???:?:?: 0x7ff80cf2adfc in ??? (???)
/Users/jakubkonka/dev/zig/lib/std/fmt.zig:116:13: 0x100122ff6 in std.fmt.format (test)
            try writer.writeAll(fmt[start_index..end_index]);
            ^
/Users/jakubkonka/dev/zig/lib/compiler_rt/floor.zig:171:24: 0x100005a15 in compiler_rt.floor.test "compiler-rt-x86_64-macos-none-Debug--multi-default floor16" (test)
    try expect(__floorh(1.3) == 1.0);
                       ^
/Users/jakubkonka/dev/zig/lib/test_runner.zig:79:28: 0x1000db882 in (root).main (test)
        } else test_fn.func();
                           ^
/Users/jakubkonka/dev/zig/lib/std/start.zig:571:22: 0x10010dc7c in std.start.callMain (test)
            root.main();
                     ^
/Users/jakubkonka/dev/zig/lib/std/start.zig:515:12: 0x1000dd257 in std.start.callMainWithArgs (test)
           ^
/Users/jakubkonka/dev/zig/lib/std/start.zig:480:12: 0x1000dd195 in std.start.main (test)
           ^
???:?:?: 0x10368751d in ??? (???)
???:?:?: 0x0 in ??? (???)
[1]    75053 abort      ./test_ld64 stage1/bin/zig
```

Digging a little deeper, we issue a placeholder instruction `call 0x0` or in hex `e800000000` which is a near-call and expects a relative displacement to another symbol to execute. However, the paired relocation points at a cell in read-only data section where we store address to emulated function `___extendhfsf2`. Tl;dr it looks like we should emit assembly closer to something like:

```asm
mov rax, 0x0 # 0x0 will be relocated displacement to read-only data cell
call rax
```

Note also that the reason this only happens for cross-compilation stripped-down target `x86_64-macos-none` is because we do not assume hw support for `f16`. In native case, the hardware support is there ofc.

As a next step, I will compare with the master branch to see if the generated machine code and relocations are the same (which I strongly doubt right now) and report back.

kubkon:
OK, compiler-rt pass for me locally now after applying this change b4f2c0d. Tl;dr like for `wasm32-wasi-musl`, we want to export `__extendhfsf2` instead of `__gnu_h2f_ieee` (and variants).

kubkon:
Thinking some more about this, Andrew discovered both symbols are provided by libsystem on macos. However for testing we link statically with compiler-rt to test the symbols, and llvm emits `__extendhfsf2` which we were not exporting for testing in the correct way leading to miscomp.

PR #11862, opened 2022-06-14
multi-thread `zig build test-cases`

### LINKED ISSUE #11818: improve the test harness to run compile error tests and runtime safety tests in parallel
There is already a thread pool here:

https://github.com/ziglang/zig/blob/6c59aa9e029c27c6da03287e907e72ac2c735b7b/src/test.zig#L1186-L1188

All we have to do is use it!

Be careful because the system right now I believe will re-use the same file name for each test. This will need to be changed to create a unique directory for each test case so that they can be run in parallel.

andrewrk:
Instead of always using std.testing.allocator, the test harness now follows
the same logic as self-hosted for choosing an allocator - that is - it
uses C allocator when linking libc, std.testing.allocator otherwise, and
respects `-Dforce-gpa` to override the decision. I did this because
I found GeneralPurposeAllocator to be prohibitively slow when doing
multi-threading, even in the context of a debug build.

There is now a second thread pool which is used to spawn each
test case. The stage2 tests are passed the first thread pool. If it were
only multi-threading the stage1 tests then we could use the same thread
pool for everything. However, the problem with this strategy with stage2
is that stage2 wants to spawn tasks and then call wait() on the main
thread. If we use the same thread pool for everything, we get a deadlock
because all the threads end up all hanging at wait() and nothing is
getting done. So we use our second thread pool to simulate a "process pool"
of sorts.

I spent most of the time working on this commit scratching my head trying
to figure out why I was getting ETXTBSY when spawning the test cases.
Turns out it's a fundamental Unix design flaw, already a known, unsolved
issue by Go and Java maintainers:

https://github.com/golang/go/issues/22315
https://bugs.openjdk.org/browse/JDK-8068370

With this change, the following command, executed on my laptop, went from
6m24s to 1m44s:

```
stage1/bin/zig build test-cases -fqemu -fwasmtime -Denable-llvm
```

closes #11818

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -398,6 +398,7 @@ pub fn build(b: *Builder) !void {
     test_cases_options.addOption(bool, "llvm_has_csky", llvm_has_csky);
     test_cases_options.addOption(bool, "llvm_has_ve", llvm_has_ve);
     test_cases_options.addOption(bool, "llvm_has_arc", llvm_has_arc);
+    test_cases_options.addOption(bool, "force_gpa", force_gpa);
     test_cases_options.addOption(bool, "enable_qemu", b.enable_qemu);
     test_cases_options.addOption(bool, "enable_wine", b.enable_wine);
     test_cases_options.addOption(bool, "enable_wasmtime", b.enable_wasmtime);
--- a/src/test.zig
+++ b/src/test.zig
@@ -1,11 +1,19 @@
 const std = @import("std");
 const builtin = @import("builtin");
+const Allocator = std.mem.Allocator;
+const CrossTarget = std.zig.CrossTarget;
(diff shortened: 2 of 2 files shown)


kubkon (approved):
Is it possible to force sequential test run via say `-Dsingle-threaded`? Sometimes, when analysing a single backend/linker, it is beneficial to go slow and step-by-step (not sure what it looks like now but having errors out-of-order can be a headache). If it's not supported yet, that's ok too. We can add it in the future together with support (or a fix if this is a bug) for selecting a single test-case to run. All in all though, awesome work!

andrewrk:
Yes it is with exactly that flag :-)

PR #11865, opened 2022-06-14
link c_nonshared

BratishkaErik:
Fix building stage2 https://github.com/ziglang/zig/issues/11137


ifreund (changes requested):
This is not the right place to fix this, the approach in #11740 was more correct.

BratishkaErik:
why

PR #11866, opened 2022-06-14
crypto: add the Xoodoo permutation, prepare for Gimli deprecation

jedisct1:
Gimli was a game changer. A permutation that is large enough to be used in sponge-like constructions, yet small enough to be compact to implement and fast on a wide range of platforms.

And Gimli being part of the Zig standard library was awesome.

But since then, Gimli entered the NIST Lightweight Cryptography Competition, competing against other candidates sharing a similar set of properties.

Unfortunately, Gimli [didn't pass the 3rd round](https://csrc.nist.gov/Projects/lightweight-cryptography/finalists).

There are no practical attacks against Gimli when used correctly, but NIST's decision means that Gimli is unlikely to ever get any traction.

So, maybe the time has come to move Gimli from the standard library to another repository.

We shouldn't do it without providing an alternative, though. And the best candidate for this is probably [Xoodoo](https://csrc.nist.gov/CSRC/media/Projects/Lightweight-Cryptography/documents/round-1/spec-doc/Xoodyak-spec.pdf).

Xoodoo is the core function of Xoodyak, one of the finalists of the NIST LWC competition, and the most direct competitor to Gimli. It is also a 384-bit permutation, so it can easily be used everywhere Gimli
was used with no parameter changes.

It is the building block of Xoodyak (for actual encryption and hashing) as well as Charm, that some Zig applications are already using.

Like Gimli that it was heavily inspired from, it is compact and suitable for constrained environments.

This change adds the Xoodoo permutation to `std.crypto.core`.

The set of public functions includes everything required to later implement existing Xoodoo-based constructions.

In order to prepare for the Gimli deprecation, the default CSPRNG was changed to a Xoodoo-based that works exactly the same way.

### DIFF
--- a/lib/std/crypto.zig
+++ b/lib/std/crypto.zig
@@ -43,6 +43,7 @@ pub const auth = struct {
 pub const core = struct {
     pub const aes = @import("crypto/aes.zig");
     pub const Gimli = @import("crypto/gimli.zig").State;
+    pub const Xoodoo = @import("crypto/xoodoo.zig").State;
 
     /// Modes are generic compositions to construct encryption/decryption functions from block ciphers and permutations.
     ///
--- a/lib/std/crypto/xoodoo.zig
+++ b/lib/std/crypto/xoodoo.zig
@@ -4,7 +4,7 @@
 //! microcontrollers, 8-bit AVR microcontrollers, FPGAs, ASICs without
 //! side-channel protection, and ASICs with side-channel protection.
 //!
-//! Xoodoo is the core function of Xoodyak, a finalists of the NIST lightweight cryptography competition.
+//! Xoodoo is the core function of Xoodyak, a finalist of the NIST lightweight cryptography competition.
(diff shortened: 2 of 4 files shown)


daurnimator:
> Unfortunately, Gimli [didn't pass the 3rd round](https://csrc.nist.gov/Projects/lightweight-cryptography/finalists).
> 
> There are no practical attacks against Gimli when used correctly, but NIST's decision means that Gimli is unlikely to ever get any traction.

Aw :(

I liked Gimli, it was simple enough that I could explain it to beginners while still being cryptographically secure.

jedisct1:
You know how I feel about Gimli as well. But we have to face the sad reality; Gimli doesn't have a bright future.

jedisct1:
Also, Xoodoo is uses a 3x4 matrix as well. and it as simple as Gimli. Even simpler and more conservative since it doesn't have the interleaved big-swap/small-swap sequence responsible for Gimli's slow diffusion.

PR #11875, opened 2022-06-16
[MachO] add -pagezero_size

motiejus:
Pass `-pagezero_size` to the MachO linker. This is the final
"unsupported linker arg" that I could chase that CGo uses. After this
and #11874 we may be able to fail on an "unsupported linker arg" instead
of emiting a warning.

Test case:

    zig=/code/zig/build/zig
    CGO_ENABLED=1 GOOS=darwin GOARCH=amd64 CC="$zig cc -target x86_64-macos" CXX="$zig c++ -target x86_64-macos" go build -a -ldflags "-s -w" cgo.go

I compiled a trivial CGo program and executed it on an amd64 Darwin
host.

To be honest, I am not entirely sure what this is doing. This feels
right after reading what this argument does in LLVM sources, but I am by
no means qualified to make MachO pull requests. Will take feedback.

### DIFF
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -4377,7 +4377,7 @@ fn populateMissingMetadata(self: *MachO) !void {
     if (self.pagezero_segment_cmd_index == null) blk: {
         if (aligned_pagezero_vmsize == 0) break :blk;
         if (aligned_pagezero_vmsize != pagezero_vmsize) {
-            log.warn("requested __PAGEZERO size is not page aligned", .{});
+            log.warn("requested __PAGEZERO size (0x{x}) is not page aligned", .{pagezero_vmsize});
             log.warn("  rounding down to 0x{x}", .{aligned_pagezero_vmsize});
         }
         self.pagezero_segment_cmd_index = @intCast(u16, self.load_commands.items.len);
--- a/src/link/MachO.zig
+++ b/src/link/MachO.zig
@@ -286,9 +286,9 @@ const default_dyld_path: [*:0]const u8 = "/usr/lib/dyld";
 const minimum_text_block_size = 64;
 pub const min_text_capacity = padToIdeal(minimum_text_block_size);
 
-/// Virtual memory offset corresponds to the size of __PAGEZERO segment and
+/// Default virtual memory offset corresponds to the size of __PAGEZERO segment and
 /// start of __TEXT segment.
-const pagezero_vmsize: u64 = 0x100000000;
+const default_pagezero_vmsize: u64 = 0x100000000;
 
 pub const Export = struct {
     sym_index: ?u32 = null,
@@ -549,7 +549,7 @@ pub fn flushModule(self: *MachO, comp: *Compilation, prog_node: *std.Progress.No
         // We can skip hashing libc and libc++ components that we are in charge of building from Zig
         // installation sources because they are always a product of the compiler version + target information.
         man.hash.add(stack_size);
-        if (self.base.options.pagezero_size) |value| man.hash.add(value);
+        man.hash.addOptional(self.base.options.pagezero_size);
         man.hash.addListOfBytes(self.base.options.lib_dirs);
         man.hash.addListOfBytes(self.base.options.framework_dirs);
         man.hash.addListOfBytes(self.base.options.frameworks);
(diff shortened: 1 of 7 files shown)


src/link/MachO.zig
```diff
@@ -4374,7 +4375,7 @@ fn populateMissingMetadata(self: *MachO) !void {
             .segment = .{
                 .inner = .{
                     .segname = makeStaticString("__PAGEZERO"),
-                    .vmsize = pagezero_vmsize,
+                    .vmsize = self.base.options.pagezero_size orelse pagezero_vmsize,
```
kubkon, line 4378: AFAIK you cannot just set the value to whatever you want without aligning to the page size.

kubkon (commented):
Thanks @motiejus. I hope you won't mind if I take it from here. As it currently stands, this PR needs a bit more work as it doesn't check the alignment on the requested page size, nor does it get rid of the segment if the requested page size is 0.

I found this PR in LLVM which I will use a guide when patching things up a little: [D118724](https://reviews.llvm.org/D118724).

motiejus:
> Thanks @motiejus. I hope you won't mind if I take it from here. As it currently stands, this PR needs a bit more work as it doesn't check the alignment on the requested page size, nor does it get rid of the segment if the requested page size is 0.
> 
> I found this PR in LLVM which I will use a guide when patching things up a little: [D118724](https://reviews.llvm.org/D118724).

By all means, thank you!

motiejus:
> @motiejus I've applied all fixes that were required, and tested that the resultant binaries (in user-space) run fine with different `__PAGEZERO` segment sizes on both x86_64 and arm64. I haven't tested CGO though so if you could give it a spin I'd be very grateful.

OK, will report here.

> I've also noted that your original patch wasn't correctly transferring the value of user-requested `pagezero_size` to the linker hence it might be worth double checking that CGO will actually work with these changes.

The cgo binary executed; that's all I can report. It was very simple though, so maybe it was not hitting a code path that required the `pagezero_size` to be a non-default.

> Finally, also note that `__PAGEZERO` of VM size `0` is an immediate kill on both x86_64 and arm64 for me - apparently this requires running directly in kernel-space, judging by the description of this LLD's patch [D118724](https://reviews.llvm.org/D118724).

Ack.

kubkon:
> > I've also noted that your original patch wasn't correctly transferring the value of user-requested `pagezero_size` to the linker hence it might be worth double checking that CGO will actually work with these changes.
> 
> The cgo binary executed; that's all I can report. It was very simple though, so maybe it was not hitting a code path that required the `pagezero_size` to be a non-default.

No, you literally forgot to pass the value to `link.zig` which then wasn't passed to `MachO.zig` and hence the `__PAGEZERO` segment wasn't modified in any way :-)

motiejus:
*cgo.go*
```
package main

// #define _FILE_OFFSET_BITS 64
// #include <unistd.h>
// #include <fcntl.h>
// #include <stdio.h>
// char* hello() { return "hello, world"; }
// void phello() { printf("%s, your lucky number is %p\n", hello(), fcntl); }
import "C"

func main() {
        C.phello()
}

func Chello() string {
        return C.GoString(C.hello())
}
```
Compile:

```
$ $ CGO_ENABLED=1 GOOS=darwin GOARCH=amd64 CC="$zig cc -target x86_64-macos" CXX="$zig c++ -target x86_64-macos" go build -a -ldflags "-s -w" cgo.go
# command-line-arguments
warning: unsupported linker arg: -no_pie
warning(link): requested __PAGEZERO size is not page aligned
warning(link):   rounding down to 0x3d0000

CGO_ENABLED=1 GOOS=darwin GOARCH=amd64 CC="$zig cc -target x86_64-macos" CXX="$zig c++ -target x86_64-macos" go build -a -ldflags "-s -w" cgo.go
# command-line-arguments
warning: unsupported linker arg: -no_pie
warning(link): requested __PAGEZERO size is not page aligned
warning(link):   rounding down to 0x3d0000
```

Golang requests this:

```
				argv = append(argv, "-Wl,-pagezero_size,4000000")
```
https://github.com/golang/go/blob/go1.18.3/src/cmd/link/internal/ld/lib.go#L1327

... which, if unaligned, should probably be reported to them?

Also, it would be good if the warning message would report the requested pagezero_size together with the "rounded down" to make debugging easier.

kubkon:
> **cgo.go**
> 
> ```
> package main
> 
> // #define _FILE_OFFSET_BITS 64
> // #include <unistd.h>
> // #include <fcntl.h>
> // #include <stdio.h>
> // char* hello() { return "hello, world"; }
> // void phello() { printf("%s, your lucky number is %p\n", hello(), fcntl); }
> import "C"
> 
> func main() {
>         C.phello()
> }
> 
> func Chello() string {
>         return C.GoString(C.hello())
> }
> ```
> 
> Compile:
> 
> ```
> $ $ CGO_ENABLED=1 GOOS=darwin GOARCH=amd64 CC="$zig cc -target x86_64-macos" CXX="$zig c++ -target x86_64-macos" go build -a -ldflags "-s -w" cgo.go
> # command-line-arguments
> warning: unsupported linker arg: -no_pie
> warning(link): requested __PAGEZERO size is not page aligned
> warning(link):   rounding down to 0x3d0000
> 
> CGO_ENABLED=1 GOOS=darwin GOARCH=amd64 CC="$zig cc -target x86_64-macos" CXX="$zig c++ -target x86_64-macos" go build -a -ldflags "-s -w" cgo.go
> # command-line-arguments
> warning: unsupported linker arg: -no_pie
> warning(link): requested __PAGEZERO size is not page aligned
> warning(link):   rounding down to 0x3d0000
> ```
> 
> Golang requests this:
> 
> ```
> 				argv = append(argv, "-Wl,-pagezero_size,4000000")
> ```
> 
> https://github.com/golang/go/blob/go1.18.3/src/cmd/link/internal/ld/lib.go#L1327
> 
> ... which, if unaligned, should probably be reported to them?

Oh, I think we don't parse it correctly. `-pagezero_size` should assume the input value in hex. Lemme double check as in hex this is properly aligned value.

> 
> Also, it would be good if the warning message would report the requested pagezero_size together with the "rounded down" to make debugging easier.

Sounds good.

> 
> Also, the resulting binary executes correctly.

kubkon:
@motiejus if you try running CGO again, the warning is gone - it was a false positive. The `-pagezero_size` value is always assumed to be hex which I've now fixed in 8752db3.

motiejus:
Thanks. I think we are in LGTM land now.  :)

motiejus:
Just checked that Go uses `-Wl,-pagezero_size` here.

PR #11880, opened 2022-06-17
std.crypto.ecc: add support for the secp256k1 curve

jedisct1:
Usage of the `secp256k1` elliptic curve recently grew exponentially, since this is the curve used by Bitcoin and other popular blockchains such as Ethereum.

With this, Zig has support for all the widely deployed elliptic curves today.

~~Note that scalar multiplication doesn't leverage the curve endomorphism yet, so expect a 20% speed increase soon.~~ <-- [UPDATE: Done]


jedisct1:
The curve name is sometimes shortened to `K256`. I personally tend to avoid it in order to avoid confusion with NIST curves. But happy to change the type name to `K256` if other people are more comfortable with that.

jedisct1:
GLV method for scalar multiplication has been implemented for `mulPublic()`, since the last relevant patent expired in 2020.

hazeycode:
@jedisct1 this is great. What are you thoughts on adding Schnorr support to the std library, or better serviced outside std?

PR #11881, opened 2022-06-17
Stage2: fixes for bugs found while looking for miscompilations

### LINKED ISSUE #11854: Stage2: generic function call is memoized when it shouldn't
```zig
const Options = struct {
    a: A = .a,
    b: B = .c,
    const A = enum { a, b };
    const B = enum { c, d };
};
fn foo(comptime opts: Options) Foo(opts) {
    return undefined;
}
fn Foo(comptime options: Options) type {
    return Bar(options.a);
}
fn Bar(comptime a: Options.A) type {
    return struct {
        const a = a;
    };
}
test {
    // correct
    // @compileLog(Foo(.{ .a = .b }).a);
    // @compileLog(Foo(.{ .b = .d }).a);
    // incorrect
    @compileLog(@TypeOf(foo(.{ .a = .b })).a);
    @compileLog(@TypeOf(foo(.{ .b = .d })).a); // should be .a
}
```
This is blocking `std.bit_set` tests.

Vexu:
Closes  #11854

### DIFF
--- a/lib/std/tz.zig
+++ b/lib/std/tz.zig
@@ -11,7 +11,7 @@ pub const Timetype = struct {
     flags: u8,
     name_data: [6:0]u8,
 
-    pub fn name(self: Timetype) [:0]const u8 {
+    pub fn name(self: *const Timetype) [:0]const u8 {
         return std.mem.sliceTo(self.name_data[0..], 0);
     }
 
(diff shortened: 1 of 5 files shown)


lib/std/tz.zig
```diff
@@ -11,7 +11,7 @@ pub const Timetype = struct {
     flags: u8,
     name_data: [6:0]u8,
 
-    pub fn name(self: Timetype) [:0]const u8 {
+    pub fn name(self: *Timetype) [:0]const u8 {
```
viriuwu, line 14: Shouldn't this be `self: *const Timetype`, since it returns a const pointer?

Vexu: It would be better.

PR #11895, opened 2022-06-19
adds isTag function to std.meta

ityonemo:
This PR adds a runtime function called isTag to std.meta; This function returns a `bool` that reflects if the enum or tagged union corresponds to the tag name, which is a comptime string.  Although using the actual tag enum value is generally preferable, in some cases (such as FFI) easy access to the string representation is desirable.

As indicated in the documentation, this function runs in constant time, regardless of the size of the enum/union or the length of the string representations of the tag name.

I can also provide tests against the compile-time checks if desired.

### DIFF
--- a/lib/std/meta.zig
+++ b/lib/std/meta.zig
@@ -62,7 +62,7 @@ test "std.meta.tagName" {
 /// be, at runtime, O(1) in the number of tags available to the enum or
 /// union, and it should also be O(1) in the length of the comptime tag
 /// names.
-pub fn isTag(enum_or_union: anytype, comptime tag_name: []const u8) bool {
+pub fn isTag(tagged_value: anytype, comptime tag_name: []const u8) bool {
     const T = @TypeOf(enum_or_union);
     const type_info = @typeInfo(T);
     const type_name = @typeName(T);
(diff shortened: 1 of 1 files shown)


rohlem:
Just a suggestion, you can get rid of the `inline for` loop at compile time.
In particular, you can query if an Enum tag exists using the builtin [`@hasField`](https://ziglang.org/documentation/master/#hasField),
and access that tag by comptime string using the builtin [`@field`](https://ziglang.org/documentation/master/#field).

If we forewent the explicit compile errors, the implementation could use only the latter, and look like the following:
```zig
pub fn isTag(tagged_value: anytype, comptime tag_name: []const u8) bool {
  const T = @TypeOf(tagged_value);

  const Enum = switch(@typeInfo(T){
    else => @compileError("unsupported type, neither enum nor (tagged) union"), 
    .Enum => T,
    .Union => |U| ..., //get its tag type
  }; //switch on @typeInfo(T): .Enum => T, .Union => Uif T is a union, its tag type

  return tagged_value == @field(Enum, tag_name);
}
```

If the tag doesn't exist the compile error generated by `@field` will already be something like "type T has no field 'abcd'", so even the `@hasField` check isn't strictly necessary I think.

ityonemo:
love it!  Doing a bit of reworking

InKryption:
What is the difference between this and `foo == .bar` exactly?

nektro:
this is entirely defeated with `std.mem.eql(u8, @tagName(tagged_value), tag_name))`

PR #11901, opened 2022-06-20
std.os.execvpe: fix buffer overlow

marler8997:
The NameTooLong check isn't taking the sentinel 0 into account which
would result in a buffer overflow on the stack.


daurnimator (approved):

ifreund (approved):
LGTM, Thanks!

PR #11904, opened 2022-06-21
stage2: wasm(linker) - prerequisites compiler-rt integration

Luukdegram:
This implements all the prerequisites to be able to link with compiler-rt and call its intrinsics while lowering AIR to Wasm instructions. Currently, the tests that implement some of those intrinsics are disabled until #11747 is implemented.
Although the test cases have been tested locally only, I'm already upstreaming all the changes so further linker progress can be made without waiting for #11747.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1911,13 +1911,12 @@ pub fn create(gpa: Allocator, options: InitOptions) !*Compilation {
             try comp.work_queue.writeItem(.libtsan);
         }
 
-        // The `use_stage1` condition is here only because stage2 cannot yet build compiler-rt.
-        // Once it is capable this condition should be removed. When removing this condition,
-        // also test the use case of `build-obj -fcompiler-rt` with the self-hosted compiler
-        // and make sure the compiler-rt symbols are emitted. Currently this is hooked up for
-        // stage1 but not stage2.
-        const capable_of_building_compiler_rt = comp.bin_file.options.use_stage1 or
-            comp.bin_file.options.use_llvm;
+        // The `have_llvm` condition is here only because native backends cannot yet build compiler-rt.
+        // Once they are capable this condition could be removed. When removing this condition,
+        // also test the use case of `build-obj -fcompiler-rt` with the native backends
+        // and make sure the compiler-rt symbols are emitted.
+        const capable_of_building_compiler_rt = build_options.have_llvm;
+
         const capable_of_building_zig_libc = comp.bin_file.options.use_stage1 or
             comp.bin_file.options.use_llvm;
         const capable_of_building_ssp = comp.bin_file.options.use_stage1;
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1917,8 +1917,7 @@ pub fn create(gpa: Allocator, options: InitOptions) !*Compilation {
         // and make sure the compiler-rt symbols are emitted.
         const capable_of_building_compiler_rt = build_options.have_llvm;
 
-        const capable_of_building_zig_libc = comp.bin_file.options.use_stage1 or
-            comp.bin_file.options.use_llvm;
+        const capable_of_building_zig_libc = build_options.have_llvm;
         const capable_of_building_ssp = comp.bin_file.options.use_stage1;
 
         if (comp.bin_file.options.include_compiler_rt and capable_of_building_compiler_rt) {
--- a/src/link.zig
+++ b/src/link.zig
@@ -436,14 +436,15 @@ pub const File = struct {
     }
 
     /// Called from within CodeGen to retrieve the symbol index of a global symbol.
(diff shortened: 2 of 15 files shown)


src/Compilation.zig
```diff
@@ -1911,13 +1911,12 @@ pub fn create(gpa: Allocator, options: InitOptions) !*Compilation {
...
-        // also test the use case of `build-obj -fcompiler-rt` with the self-hosted compiler
-        // and make sure the compiler-rt symbols are emitted. Currently this is hooked up for
-        // stage1 but not stage2.
-        const capable_of_building_compiler_rt = comp.bin_file.options.use_stage1 or
-            comp.bin_file.options.use_llvm;
+        // The `have_llvm` condition is here only because native backends cannot yet build compiler-rt.
+        // Once they are capable this condition could be removed. When removing this condition,
+        // also test the use case of `build-obj -fcompiler-rt` with the native backends
+        // and make sure the compiler-rt symbols are emitted.
+        const capable_of_building_compiler_rt = build_options.have_llvm;
```
Luukdegram, line 1918: Had to make this change so we can still compile compiler-rt using LLVM while using the flag `no-LLVM` to use the native backend for the user's Zig code. I wonder if perhaps we should do the same for `capable_of_building_libc`.

src/link/Wasm.zig
```diff
@@ -753,6 +864,38 @@ pub fn lowerUnnamedConst(self: *Wasm, tv: TypedValue, decl_index: Module.Decl.In
     return atom.sym_index;
 }
 
+/// Returns the symbol index from the name of an intrinsic.
```
kubkon, line 867: Actually, this will be reused to resolve globals from relocatable object files linked on request I believe. Have a look at `addExternFn` in `MachO.zig` which is effectively performing the same thing (minus the implementation details).

src/link.zig
```diff
@@ -435,6 +435,24 @@ pub const File = struct {
         }
     }
 
+    /// Called from within CodeGen to retrieve the symbol index of a global symbol.
+    /// If no symbol exists yet with this name, a new one will be created instead.
```
kubkon, line 439: Could you add a clarification here, that if the symbol didn't yet exist, we create a new undefined global symbol which will hopefully get resolved once all relocatables are re-linked?

src/link.zig
```diff
@@ -435,6 +435,24 @@ pub const File = struct {
...
+    pub fn getGlobalSymbol(base: *File, name: []const u8) UpdateDeclError!u32 {
+        log.debug("getGlobalSymbol '{s}'", .{name});
+        switch (base.tag) {
+            // zig fmt: off
+            .coff  => unreachable,
+            .elf   => unreachable,
+            .macho => unreachable,
```
kubkon, line 446: This is a big ask so feel free to refuse it - MachO already implements this function but under a different name `addExternFn`. Would you be OK with renaming it to `getGlobalSymbol` in MachO and plugging it here too?

kubkon (changes requested):
Great work @Luukdegram! I do have one question: from what I've seen in this patch, we do not have to request compiler-rt being linked or anything - that's done automatically, and already present on the linker line, correct?

src/Compilation.zig
kubkon: I think stage2 + LLVM should now be capable of building libc, so yeah, I think we should do the same.

src/link.zig
kubkon: Thank you kindly!

Luukdegram:
> Great work @Luukdegram! I do have one question: from what I've seen in this patch, we do not have to request compiler-rt being linked or anything - that's done automatically, and already present on the linker line, correct?

That's entirely correct!

kubkon:
> > Great work @Luukdegram! I do have one question: from what I've seen in this patch, we do not have to request compiler-rt being linked or anything - that's done automatically, and already present on the linker line, correct?
> 
> That's entirely correct!

Excellent! Thanks for confirming my understanding!

kubkon (approved):

andrewrk:
Small request: for PRs like this, please use 

 * **Create a merge commit**, or
 * **Squash and merge**

unless you made sure that every commit individually passes the full test suite. The situation I'm running into now is running the backfill script for https://ziglang.org/perf/ but many of the commits in this branch are now direct children of master branch (e.g. they show up with `--first-parent`) and fail to build from source with:
```
./src/link/Wasm.zig:1832:13: error: no member named 'compiler_rt_static_lib' in struct 'Compilation'
    if (comp.compiler_rt_static_lib) |lib| {
            ^
```

This is also problematic for the `git bisect` use case. So while normally I prefer to ignore minor stylistic choices for contributions, I think for these two specific reasons it makes sense to try to be consistent about merge strategy.

Luukdegram:
Understood. Sorry for the trouble, will keep it in mind for the future.

PR #11863, opened 2022-06-14
ELF: understand --compress-debug-sections

motiejus:
This argument is both a compiler and a linker flag. The linker flag was
not understood; now it is. Go likes to use it as a linker flag.

Tested with sqlite3. The size difference is significant, and I confirmed
gdb understands both binaries.

Also added for `build-*` commands, following advice by @kubkon. 

zlib: 3.66MB
------------

    CC="zig cc"
    CFLAGS="-Wl,--compress-debug-sections=zlib -O2"
    ./configure --disable-tcl
    make

        FILE SIZE        VM SIZE
     --------------  --------------
      39.1%  1.43Mi  88.4%  1.43Mi    .text
      19.6%   734Ki   0.0%       0    .debug_info
      16.4%   613Ki   0.0%       0    .debug_loc
      13.1%   492Ki   0.0%       0    .debug_line
       4.2%   157Ki   9.5%   157Ki    .rodata
       2.3%  87.6Ki   0.0%       0    .debug_ranges
       1.5%  56.2Ki   0.0%       0    .symtab
       1.1%  40.2Ki   0.0%       0    .strtab
       1.0%  38.2Ki   0.0%       0    .debug_str
       0.7%  26.2Ki   0.0%       0    .debug_frame
       0.4%  15.3Ki   0.9%  15.3Ki    .data
       0.1%  4.71Ki   0.3%  4.71Ki    .dynsym
       0.1%  3.65Ki   0.2%  3.26Ki    [16 Others]
       0.1%  2.55Ki   0.2%  2.55Ki    .rela.plt
       0.1%  2.12Ki   0.0%       0    [ELF Section Headers]
       0.0%       0   0.1%  2.02Ki    .bss
       0.0%  1.84Ki   0.1%  1.84Ki    .dynstr
       0.0%  1.72Ki   0.1%  1.72Ki    .plt
       0.0%  1.58Ki   0.1%  1.58Ki    .hash
       0.0%  1.17Ki   0.0%       0    .debug_abbrev
       0.0%  1.01Ki   0.1%  1.01Ki    .rela.dyn
     100.0%  3.66Mi 100.0%  1.62Mi    TOTAL

none: 8.56MB
------------

    CC="zig cc" CFLAGS="-O2" ./configure --disable-tcl
    make

        FILE SIZE        VM SIZE
     --------------  --------------
      41.1%  3.52Mi   0.0%       0    .debug_loc
      18.5%  1.59Mi   0.0%       0    .debug_info
[...]

PR #11906, opened 2022-06-22
[all linkers] fail hard on unsupported flags

motiejus:
Currently `zig cc`, when confronted with a linker argument it does
not understand, skips the flag and emits a warning.

This has been causing headaches for people that build third-party
software (including me). Zig seemingly builds and links the final
executable, only to segfault when running it. Emiting a warning and
"successfully" compiling a binary breaks autoconf assumptions.
See https://github.com/zigchroot/zig-chroot/issues/1 for a recent
painful example.

If there are linker warnings when compiling software, the first thing we
have to do is add support for ones linker is complaining, and only then
go file issues. If zig "successfully" (i.e. status code = 0) compiles a
binary, there is instead a tendency to blaim "zig doing something
weird". (I am guilty of this.) In my experience, adding the unsupported
arguments has been quite easy; see #11679, #11875, #11874 for recent
examples.

With the "soft prerequisites" below I was able to build all of
the CGo programs that I am encountering at $dayjob.

Harder prerequisites — some production code in my dayjob will cease to build unless we merge this first:
- https://github.com/ziglang/zig/pull/11863

Soft prerequisites — some things I've observed in other issues/PRs:
- https://github.com/ziglang/zig/pull/11614
- https://github.com/ziglang/zig/pull/11906
- https://github.com/ziglang/zig/pull/11912
- https://github.com/ziglang/zig/pull/11913
- https://github.com/ziglang/zig/pull/11914
- https://github.com/ziglang/zig/pull/11915
- https://github.com/ziglang/zig/pull/11916
- https://github.com/ziglang/zig/pull/11917

Andrew's comment in favor of failing early instead of emitting a warning: https://github.com/zigchroot/zig-chroot/issues/1#issuecomment-1054710190 . (NOTE: the repo no longer exists, I would be grateful if someone has a copy of the comment.)

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1343,7 +1343,7 @@ fn buildOutputType(
                         } else if (mem.startsWith(u8, z_arg, "max-page-size=")) {
                             linker_z_max_page_size = parseIntSuffix(z_arg, "max-page-size=".len);
                         } else {
-                            warn("unsupported linker extension flag: -z {s}", .{z_arg});
+                            fatal("unsupported linker extension flag: -z {s}", .{z_arg});
                         }
                     } else if (mem.eql(u8, arg, "--import-memory")) {
                         linker_import_memory = true;
@@ -1961,7 +1961,7 @@ fn buildOutputType(
                     } else if (mem.startsWith(u8, z_arg, "max-page-size=")) {
                         linker_z_max_page_size = parseIntSuffix(z_arg, "max-page-size=".len);
                     } else {
-                        warn("unsupported linker extension flag: -z {s}", .{z_arg});
+                        fatal("unsupported linker extension flag: -z {s}", .{z_arg});
                     }
                 } else if (mem.eql(u8, arg, "--major-image-version")) {
                     i += 1;
@@ -2186,7 +2186,7 @@ fn buildOutputType(
 
                     have_version = true;
                 } else {
-                    warn("unsupported linker arg: {s}", .{arg});
+                    fatal("unsupported linker arg: {s}", .{arg});
                 }
             }
 


marler8997:
You'd be able to get this behavior now without breaking anyone if you made it an opt-in compiler switch.

motiejus:
> You'd be able to get this behavior now without breaking anyone if you made it an opt-in compiler switch.

I can, however, I am unsure of its value and am quite sure of its disadvantages. There are a few examples of such users:

1. Users who want their software to compile with `zig cc`.
2. Users who are copy-pasting others' instructions and will leave in their build scripts whatever "works".

Users in category (1) can revert this commit  and see how far the build can progress for their software. If they want to ship scripts that compile their software with zig, the only reasonable way forward is to implement the missing flags (and then ship). If we, however, offer a clutch, the incentive -- and venue for error -- will be to use that flag, but not implement the necessary bits in zig. At some point all linker options will be understood, so we will have to leave that flag.. for backwards compatibility?

I especially explicitly want to make (2) difficult. A linker flag "errors-as-warnings" is a source of pain and misery *especially* when something emits a "warning" 5 layers down in whatever you happen to be using.

marler8997:
In order to use "zig cc" with as many projects as possible, I don't think we have a choice whether or not we want to allow unrecognized command line arguments.  I suggest we add a command-line option to enable/disable this behavior (i.e. `--allow-unrecognized-flags` vs `--deny-unrecognized-flags`) and plan to keep this option permanently.  This would give users the ability to turn it on or off according to their needs and a non-breaking path forward to change the behavior from "opt-in" to "opt-out".  This means you could have the feature now without breaking anyone, and after some time has passed and projects that need it have started using `--allow-unrecignized-flags`, then we can make the safer option to deny unrecognized flags the default.

kubkon:
I don't have a strong opinion on this either way - on one hand, I agree with @motiejus that we really should do more than just (almost) silently ignore the unsupported flags. On the other hand, I agree with @marler8997 that there are so many flags out there it will be a pain trying to support them all and/or might take a long time to implement. If you don't mind I'd like to defer the final decision to @andrewrk once he's back at work next week.

Changing topic slightly, @motiejus dunno if you noticed, with this change building arm64 Zig binary fails on macos - it complains about unknown `-search_paths_first` flag. This one is an easy fix: we should ignore it since this is the default behaviour in both `ld64` and `zld`.

motiejus:
I found Andrew's comment in favor of this in https://github.com/zigchroot/zig-chroot/issues/1#issuecomment-1054710190 ; now I am adding support to get rid of those warnings (you can see PRs slowly trickling in):

```
warning: unsupported linker arg: --sort-section
warning: unsupported linker arg: alignment
warning: unsupported linker arg: --sort-common
warning: unsupported linker arg: --no-undefined
warning: unsupported linker arg: --exclude-libs
warning: unsupported linker arg: ALL
warning: unsupported linker arg: --dynamic-list
warning: unsupported linker arg: ./dynamic.list
```

motiejus:
> Changing topic slightly, @motiejus dunno if you noticed, with this change building arm64 Zig binary fails on macos - it complains about unknown `-search_paths_first` flag.

We don't build MacOS binaries for arm64, because we did not get to the `-framework` thing yet. :)

> This one is an easy fix: we should ignore it since this is the default behaviour in both `ld64` and `zld`.

https://github.com/ziglang/zig/pull/11917

kubkon:
MachO linker should now implement all flags that could lead to a fatal error when building x86_64 and arm64 in the CI. Relevant PRs: #11917 and #11929

motiejus:
> I found Andrew's comment in favor of this in https://github.com/zigchroot/zig-chroot/issues/1#issuecomment-1054710190 <...>

The repository is gone now. Neither google cache nor archive.org have it; nor does my MUA (I subscribed to the thread later than this). If somebody still has a reference to the original comment, or even better, the whole thread — I would be grateful if you could put it here.

andrewrk:
I agree with this behavior and will merge this PR- but first zig CLI parsing needs to support all the existing linker flags to avoid breaking things.

andrewrk:
I merged some other linker flag PRs, pushed d2650eb570c5494c728a0a780feb310abe2f1222, filed #14663, and I'm going to go ahead and merge this now.

Issue #2647, opened 2019-06-10
Allow returning a value with an error


CurtisFenner:
Sometimes when a function fails, there is extra information that you have on hand that may help the caller respond to the problem or produce a diagnostic. For example, in the `parseU64` example by andrewrk [here](https://andrewkelley.me/post/intro-to-zig.html), 

```zig
const ParseError = error {
    InvalidChar,
    Overflow,
};

pub fn parseU64(buf: []const u8, radix: u8) ParseError!u64 {
```
it would be useful for the function could return the position of the invalid character so that the caller could produce a diagnostic message.

Because Zig treats `error` types specially, when using errors you get a bunch of nice features, such as `!` error-set inference, `try`/`catch`, and `errdefer`; you currently lose these features if you want to return extra diagnostic information since that information is no longer an error type.

While something like index-of-bad-character is less useful for parsing an integer, getting "bad character" with no location when parsing a 2KiB JSON blob is very frustrating! -- this is the current state of the standard library's JSON parser.

There are currently two workarounds possible today to let this extra information get out, neither of which are very ergonomic and which work against Zig's error types:

#### Workaround 1: Return a tagged union

You could explicitly return a tagged union that has the extra information:

```zig
const ParseError = error {
    Overflow,
}

const ParseResult = union(enum) {
    Result: u64,
    InvalidChar: usize,
}

pub fn parseU64(buf: []const u8, radix: u8) ParseError!ParseResult {
```

This is unfortunate in a number of ways. First, because `InvalidChar` is no longer an `error`, you cannot propagate/handle the failure with `try`/`catch`. Second, because the `InvalidChar` case is no longer an `error`, you cannot use `errdefer` to cleanup partially constructed state in the parser. Finally, calling the function is made messy because it can fail in two separate ways -- either in the error union, or in the explicitly returned union. This means calls that distinguish different errors (as opposed to just propagating with `try`) need nested `switch`es.

#### Workaround 2: Write to an out parameter

You could also leave the error set alone, and instead expand the contract of `parseU64` to write to an out parameter whenever it returns a `InvalidChar` error:

```
pub fn parseU64(buf: []const u8, radix: u8, invalid_char_index: *usize) ParseError!u64{
```

[...]

PR #11908, opened 2022-06-22
mem: add splitBackwards

motiejus:
Over the last couple of weeks weeks I needed to iterate over a
collection backwards at least twice. Do we want to have this in stdlib?
If yes, click "Merge" and start using today! Free shipping and returns
(before 1.0).

Why is this useful?
-------------------

I need this for building an error wrapper: errors are added in the
wrapper from "lowest" level to "highest" level, and then printed in
reverse order. Imagine `UpdateUsers` call, which needs to return
`error.InvalidInput` and a wrappable error context. In Go we would add a
context to the error when returning it:

    // if update_user fails, add context on which user we are operating
    if err := update_user(user); err != nil {
        return fmt.Errorf("user id=%d: %w", user.id, err)
    }

Since Zig cannot pass anything else than u16 with an error (#2647), I
will pass a `err_ctx: *Err`, to the callers, where they can, besides
returning an error, augment it with auxiliary data. `Err` is a
preallocated array that can add zero-byte-separated strings. For a
concrete example, imagine such a call graph:

    update_user(User, *Err) error{InvalidInput}!<...>
      validate_user([]const u8, *Err) error{InvalidInput}!<...>

Where `validate_user` would like, besides only the error, signal the
invalid field. And `update_user`, besides the error, would signal the
offending user id.

We also don't want the low-level functions to know in which context they
are operating to construct a meaningful error message: if validation
fails, they append their "context" to the buffer. To translate/augment
the Go example above:

    pub fn validate_user(err_ctx: *Err, user: User) error{InvalidInput}!void {
        const name = user.name;
        if (!ascii.isAlpha(user.name)) {
            err_ctx.print("name '{s}' must be ascii-letters only", .{user.name});
            return error.InvalidInput;
        }
        <...>
    }

    // update_user validates each user and does something with it.
    pub fn update_user(err_ctx: *Err, user: User) error{InvalidInput}!void {
        // validate the user before updating it
        validate_user(user) catch {
            err_ctx.print("user id={d}", .{user.id});
            return error.InvalidInput;
        };
        <...>
    }

Then the top-level function (in my case, CLI) will read the buffer
backwards (splitting on `"\x00"`) and print:

    user id=123: name 'Žvangalas' must be ascii-letters only

To read that buffer backwards, dear readers of this commit message, I
need `mem.splitBackwards`.

This PR has 2 commits:

1. refactor `split` tests to use `assertEqualSlices`.
2. refactor `split` tests to use `rest()` occasionally.
3. add `splitBackwards` with similar tests to `split`.

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -1702,23 +1702,32 @@ pub fn split(comptime T: type, buffer: []const T, delimiter: []const T) SplitIte
 
 test "split" {
     var it = split(u8, "abc|def||ghi", "|");
-    try testing.expect(eql(u8, it.next().?, "abc"));
-    try testing.expect(eql(u8, it.next().?, "def"));
-    try testing.expect(eql(u8, it.next().?, ""));
-    try testing.expect(eql(u8, it.next().?, "ghi"));
+    try testing.expectEqualSlices(u8, it.rest(), "abc|def||ghi");
+    try testing.expectEqualSlices(u8, it.next().?, "abc");
+
+    try testing.expectEqualSlices(u8, it.rest(), "def||ghi");
+    try testing.expectEqualSlices(u8, it.next().?, "def");
+
+    try testing.expectEqualSlices(u8, it.rest(), "|ghi");
+    try testing.expectEqualSlices(u8, it.next().?, "");
+
+    try testing.expectEqualSlices(u8, it.rest(), "ghi");
+    try testing.expectEqualSlices(u8, it.next().?, "ghi");
+
+    try testing.expectEqualSlices(u8, it.rest(), "");
     try testing.expect(it.next() == null);
 
     it = split(u8, "", "|");
-    try testing.expect(eql(u8, it.next().?, ""));
+    try testing.expectEqualSlices(u8, it.next().?, "");
     try testing.expect(it.next() == null);
 
     it = split(u8, "|", "|");
-    try testing.expect(eql(u8, it.next().?, ""));
-    try testing.expect(eql(u8, it.next().?, ""));
+    try testing.expectEqualSlices(u8, it.next().?, "");
+    try testing.expectEqualSlices(u8, it.next().?, "");
     try testing.expect(it.next() == null);
 
     it = split(u8, "hello", " ");
-    try testing.expect(eql(u8, it.next().?, "hello"));
+    try testing.expectEqualSlices(u8, it.next().?, "hello");
     try testing.expect(it.next() == null);
 
     var it16 = split(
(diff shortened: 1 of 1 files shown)


lib/std/mem.zig
```diff
@@ -1702,53 +1701,152 @@ pub fn split(comptime T: type, buffer: []const T, delimiter: []const T) SplitIte
...
+
+/// Returns an iterator that iterates backwards over the slices of `buffer`
+/// that are separated by bytes in `delimiter`.
+/// split_rev(u8, "abc|def||ghi", "|")
+/// will return slices for "ghi", "", "def", "abc", null, in that order.
+/// If `delimiter` does not exist in buffer,
+/// the iterator will return `buffer`, null, in that order.
+/// The delimiter length must not be zero.
+/// See also the related function `tokenize`.
+pub fn split_rev(comptime T: type, buffer: []const T, delimiter: []const T) SplitIteratorRev(T) {
```
InKryption, line 1772: Function names are expected to be camelCase: https://ziglang.org/documentation/master/#Style-Guide

Luukdegram: The [style guide](https://ziglang.org/documentation/master/#Style-Guide) mandates camelCase for functions: `splitRev` (same for doc comment).

Luukdegram (changes requested):

lib/std/mem.zig
motiejus: That's what happens when you need to change the programming language more than once a day.

lib/std/mem.zig
```diff
@@ -1690,7 +1690,6 @@ test "tokenize (reset)" {
 /// If `delimiter` does not exist in buffer,
 /// the iterator will return `buffer`, null, in that order.
 /// The delimiter length must not be zero.
-/// See also the related function `tokenize`.
```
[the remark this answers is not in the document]
motiejus: Yes! You are right. Removed the wrong line. Updated.

Luukdegram:
I don't want to get into bikeshedding theory but I wonder if it should be named `splitReversed` fully written out, similarly to how we have `copyBackwards`.

Luukdegram (approved):

motiejus:
> shedding

I like your colour better; updating to `splitReversed`.

motiejus:
> > I don't want to get into bikeshedding territory but I wonder if it should be named `splitReversed` fully written out, similarly to how we have `copyBackwards`.
 

Repainted again, to `splitBackwards` to be consistent with `copyBackwards`.

PR #11912, opened 2022-06-23
[linker] add --no-undefined, -z undefs

motiejus:


motiejus:
Rebased and no longer bumping the hash generation, as only the front-end changes.

andrewrk:
Thanks! I pushed one minor fixup before merging.

PR #11913, opened 2022-06-23
[elf linker] add --dynamic-list

motiejus:

I did not test this, as I don't have a good standalone program to do
that. I will ask @zigchroot to do once I finish implementing the
remaining missing flags.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -834,6 +834,7 @@ pub const InitOptions = struct {
     linker_script: ?[]const u8 = null,
     version_script: ?[]const u8 = null,
     soname: ?[]const u8 = null,
+    linker_dynamic_list: ?[]const u8 = null,
     linker_gc_sections: ?bool = null,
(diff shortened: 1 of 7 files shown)


andrewrk (changes requested):
would you mind fixing the commit message & rebasing?

also, what does this flag do? does it select which symbols to export? if so, I think it should be resolved in the CLI, before creating a Compilation. The CLI can read a file if necessary, and communicate the set of dynamic symbols via a map or something. There might already be an API for this.

motiejus:
> would you mind fixing the commit message & rebasing?
> 
> also, what does this flag do? does it select which symbols to export? if so, I think it should be resolved in the CLI, before creating a Compilation. The CLI can read a file if necessary, and communicate the set of dynamic symbols via a map or something. There might already be an API for this.

Zig-chroot has been deleted and the original author did not respond since. As I have no use for this myself, I am closing the issue.

PR #11916, opened 2022-06-23
[elf] add --exclude-libs

motiejus:
Tested by created a "hello world" C file and compiled with:

    zig cc -v main.c -Wl,--exclude-libs=ALL -o main

... and verified that `--exclude-libs=ALL` is passed to ld.lld.


andrewrk (changes requested):
Have a look at link.SystemLib:

https://github.com/ziglang/zig/blob/705e9cb3caaa049336104662a833716757ffcb4a/src/link.zig#L23-L26

both of these fields are related to whether to exclude a given library.

I think whether to make all symbols hidden in a library should be part of this struct and understood by the frontend rather than passed as an opaque value to the linker.

I'm guessing the CLI itself should be resolving the excluded libs and setting this field before creating a Compilation.

motiejus:
> Have a look at link.SystemLib:
> 
> https://github.com/ziglang/zig/blob/705e9cb3caaa049336104662a833716757ffcb4a/src/link.zig#L23-L26
> 
> both of these fields are related to whether to exclude a given library.
> 
> I think whether to make all symbols hidden in a library should be part of this struct and understood by the frontend rather than passed as an opaque value to the linker.
> 
> I'm guessing the CLI itself should be resolving the excluded libs and setting this field before creating a Compilation.

I agree with the proposed change to this PR.

However, as Zig-chroot has been deleted, the original author did not respond since and I have no use for this myself, I am closing the issue.

polarathene:
FWIW this linker arg came up quite a few times when attempting to use Zig to build python packages with C dependencies to build (_forcing sdists for install instead of bdists wheels to lower the glibc version_).

Also very relevant for the very popular `python-build-standalone` project which relies upon `--exclude-libs=ALL`.

There is a related issue that I've referenced back to here requesting support for the linker arg (_2 other users commented, 14 👍 reacts, perhaps more in subscribers_). So I think there's some more valid demand for it, at least with `zig cc` / `zig c++`?

PR #11917, opened 2022-06-23
macho: implement `-search_paths_first` and `-search_dylibs_first`

motiejus:
Ignore MachO-specific flag -search_paths_first, since it is the default
in zld and ld64.

Also see Jakub's comment[1]:

    Changing topic slightly, @motiejus dunno if you noticed, with this
    change building arm64 Zig binary fails on macos - it complains about
    unknown -search_paths_first flag. This one is an easy fix: we should
    ignore it since this is the default behaviour in both ld64 and zld.

[1]: https://github.com/ziglang/zig/pull/11906#issuecomment-1163545849


kubkon:
Thanks for the PR @motiejus! I'm going to hijack your branch and fix another unhandled flag `-headerpad_max_install_names`, and push all in one go. However, before this happens, I wanna land https://github.com/ziglang/zig/pull/11910 first so that I can immediately add some relevant linker tests. Hope you don't mind!

kubkon:
Apologies for making you a reviewer on your own PR @motiejus, but would you mind having a look? Instead of implementing `-headerpad_max_install_names`, I've decided to implement `-search_dylibs_first` which is another possible search strategy to `-search_paths_first` in MachO linkers. I've also noticed we were substituting a static archive (if found) for a system library name (i.e., `-la` was getting transformed into `<path>/liba.a` in `main.zig`). This is actually wrong targeting MachO as we want to defer the resolution of all libs to the link stage so that we can correctly act on the `-search_paths_first` and `-search_dylibs_first` flags.

src/link/MachO.zig
```diff
@@ -784,18 +790,43 @@ pub fn flushModule(self: *MachO, comp: *Compilation, prog_node: *std.Progress.No
...
+                        // Look in each directory for a dylib (stub first), and then for archive
+                        for (lib_dirs.items) |dir| {
+                            for (&[_][]const u8{ ".tbd", ".dylib", ".a" }) |ext| {
+                                if (try resolveLib(arena, dir, lib_name, ext)) |full_path| {
+                                    try libs.append(full_path);
+                                    continue :outer;
+                                }
+                            }
+                        } else {
+                            log.warn("library not found for '-l{s}'", .{lib_name});
```
motiejus, line 808: can we `fatal` instead? If a user asked for a library and we can't find it, I think it's wise to fail.

src/link/MachO.zig
```diff
@@ -784,18 +790,43 @@ pub fn flushModule(self: *MachO, comp: *Compilation, prog_node: *std.Progress.No
...
+                                if (try resolveLib(arena, dir, lib_name, ext)) |full_path| {
+                                    try libs.append(full_path);
+                                    continue :outer;
+                                }
+                            }
+                        } else for (lib_dirs.items) |dir| {
+                            if (try resolveLib(arena, dir, lib_name, ".a")) |full_path| {
+                                try libs.append(full_path);
+                            } else {
+                                log.warn("library not found for '-l{s}'", .{lib_name});
```
motiejus, line 825: ditto

src/link/MachO.zig
```diff
@@ -1183,51 +1244,41 @@ fn resolveSearchDir(
...
-        };
-        defer tmp.close();
-
-        return full_path;
-    }
+    // Check if the file exists.
+    const tmp = fs.cwd().openFile(full_path, .{}) catch |err| switch (err) {
+        error.FileNotFound => return null,
+        else => |e| return e,
+    };
```
motiejus, line 1278: can we use `stat` instead? We don't really need to *open* it.

src/main.zig
```diff
@@ -917,6 +920,10 @@ fn buildOutputType(
                         pagezero_size = std.fmt.parseUnsigned(u64, eatIntPrefix(next_arg, 16), 16) catch |err| {
                             fatal("unable to parse '{s}': {s}", .{ arg, @errorName(err) });
                         };
+                    } else if (mem.eql(u8, arg, "-search_paths_first")) {
+                        search_strategy = .paths_first;
+                    } else if (mem.eql(u8, arg, "-search_dylibs_first")) {
+                        search_strategy = .dylibs_first;
```
motiejus, line 926: Perhaps this should be a linker option, not something naked `zig cc` should accept.

motiejus (commented):
> Apologies for making you a reviewer on your own PR @motiejus, but would you mind having a look?

Sure! Sorry for not reviewing earlier; you posted this on the day of summer solstice :)

src/link/MachO.zig
kubkon: We fail *after* we try resolving all library names. This way we match the behavior of ld64 and lld and we warn about all unresolved library names at once, rather than fail after the first unresolved one. Here's where we fail in the code https://github.com/ziglang/zig/blob/08459ff1c21d546c55e2ae954126e121ee88972e/src/link/MachO.zig#L1113

src/link/MachO.zig
kubkon: See my comment above.

src/link/MachO.zig
kubkon: Potentially, but this is contingent on implementing/finding the correct behaviour on Windows - remember that this linker has to be able to work on non-POSIX OSes too. I suggest we audit this in the future once we have Zig's libstd further along. Also for this very reason we currently don't `mmap` the object files when linking since this requires special handling on Windows :^)

src/main.zig
kubkon: You are actually pointing to what `build-*` accepts. `zig cc` accepts `-Wl,-search_paths_first` and `-Wl,-search_dylibs_first` and the code for this is here https://github.com/ziglang/zig/blob/08459ff1c21d546c55e2ae954126e121ee88972e/src/main.zig#L1498-L1502

PR #11919, opened 2022-06-24
Add stack trace capturing to `FailingAllocator` and use it to improve `checkAllAllocationFailures`

squeek502:
There are two parts to this:

---

Adds stack trace capturing to `std.testing.FailingAllocator` and makes `std.testing.checkAllAllocationFailures` print the stack trace of the induced allocation failure when memory leaks are detected (very helpful for debugging)

---

`checkAllAllocationFailures`: add possibility of SwallowedOutOfMemoryError (split from NondeterministicMemoryUsage)

Inducing failure but not getting OutOfMemory back is not as much of a problem as never inducing failure when it was expected to be induced, so treating them differently and allowing them to be handled differently by the caller is useful.

For example, the current implementation of `std.HashMapUnmanaged.getOrPutContextAdapted` always tries to grow and then recovers from OutOfMemory by attempting a lookup of an existing key. If `getOrPutContextAdapted` is used (i.e. from `std.BufMap.putMove`) with `checkAllAllocationFailures`, then we'd have previously triggered `error.NondeterministicMemoryUsage`, but the real cause is that `OutOfMemory` is being recovered from and so the error is being swallowed. The new error allows us to both understand what's happening easier and to catch it and ignore it if we're okay with the code we're testing handling `error.OutOfMemory` without always bubbling it up.

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -26,6 +26,27 @@ pub const runtime_safety = switch (builtin.mode) {
     .ReleaseFast, .ReleaseSmall => false,
 };
 
+pub const sys_can_stack_trace = switch (builtin.cpu.arch) {
+    // Observed to go into an infinite loop.
+    // TODO: Make this work.
+    .mips,
+    .mipsel,
+    => false,
+
+    // `@returnAddress()` in LLVM 10 gives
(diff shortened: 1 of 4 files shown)


andrewrk (approved):
Looks great! Thanks for the improvements, and the detailed writeup on your blog.

PR #11925, opened 2022-06-24
std.mem: add peek() to TokenIterator(T)

ThreeFx:
Peeking is a useful operation on tokenizers.

I'm unsure whether this would make sense for `SplitIterator(T)` as well. Any opinions?

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -1608,6 +1608,7 @@ pub fn tokenize(comptime T: type, buffer: []const T, delimiter_bytes: []const T)
 test "tokenize" {
     var it = tokenize(u8, "   abc def   ghi  ", " ");
     try testing.expect(eql(u8, it.next().?, "abc"));
+    try testing.expect(eql(u8, it.peek().?, "def"));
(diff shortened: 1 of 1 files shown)


ifreund (commented):
Could we implement `next()` in terms of `peek()` to reduce duplication? Something like
```zig
const result = peek();
self.index += result.len;
return result;
```

ifreund:
> I'm unsure whether this would make sense for `SplitIterator(T)` as well. Any opinions?

I'd say let's leave it out until someone has a use-case, it's not a terribly tricky addition to make or review in the future if needed.

ifreund:
> We could, if we allow `peek()` to move the start token to the first non-delimiter element (otherwise we have to re-search the next token's starting point again). This would break the promise made in the doc comment (since we do modify the iterator's internal state), but as the results are identical I think it's probably fine.

Ah I see, I think that's fine. It's likely a performance improvement as well if calling `peek()` repeatedly when there is a large amount of delimiter bytes before the next token.

ifreund (approved):
LGTM, Thanks!

I just made one small grammar fix in a comment before the merge.

PR #11929, opened 2022-06-24
macho: handle `-headerpad` and `-headerpad_max_install_names`

kubkon:
Tweaks link-tests and specifically, refactors and pulls `MatchAction` and `ComputeEqAction` into a single `Action` struct with two parsers attached to it `fn match(...)` and `fn computeCmp(...)`. The former is as it was, however, the latter now generalises the concept of `ComputeEqAction` by deferring parsing until we actually execute the test (`fn make(...)`), and permitting comparing against either an extracted variable or a literal (currently only `u64`) and allowing for any comparison operator of type `std.math.CompareOperator`.

cc @Luukdegram due to the above.


Luukdegram:
I like the changes to the linker test structure 👍

src/main.zig
```diff
@@ -924,6 +928,15 @@ fn buildOutputType(
...
                         search_strategy = .dylibs_first;
+                    } else if (mem.eql(u8, arg, "-headerpad")) {
+                        const next_arg = args_iter.next() orelse {
+                            fatal("expected parameter after {s}", .{arg});
+                        };
+                        headerpad_size = std.fmt.parseUnsigned(u32, eatIntPrefix(next_arg, 16), 16) catch |err| {
+                            fatal("unable to parser '{s}': {s}", .{ arg, @errorName(err) });
+                        };
+                    } else if (mem.eql(u8, arg, "-headerpad_max_install_names")) {
+                        headerpad_max_install_names = true;
```
motiejus, line 939: nit: do we want to allow naked `-headerpad` and `-headerpad_max_install_names` to `zig cc`, or should we require them to be passed via `-Wl,<...>` alone?

I have done this mistake myself a few times of adding a linker-only option here (will submit PRs reverting that soon); but we should be quite upfront about adding them here.

motiejus: Turns out none of such 'bad" code is landed yet; I've updated the PRs-in-flight to the correct behavior.

kubkon: This PR added support for both. `build-*` accepts directly `-headerpad` and `-headerpad_max_install_names` https://github.com/ziglang/zig/blob/08459ff1c21d546c55e2ae954126e121ee88972e/src/main.zig#L931, and `cc` and `c++` accept `-Wl,-headerpad=<size>` and `-Wl,-headerpad_max_install_names` https://github.com/ziglang/zig/blob/08459ff1c21d546c55e2ae954126e121ee88972e/src/main.zig#L1692 :^)

PR #11931, opened 2022-06-25
std.fmt: clarify the description of placeholders in Format Strings

frou:
Those 6 sets of square brackets are just a typographical aid used in this doc-comment, and must not actually be written by the user in their own format string placeholders... except for in one case where they must

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -36,7 +36,7 @@ pub const FormatOptions = struct {
 ///
 /// - *argument* is either the numeric index or the field name of the argument that should be inserted
 ///   - when using a field name, you are required to enclose the field name (an identifier) in square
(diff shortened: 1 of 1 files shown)


lib/std/fmt.zig
```diff
@@ -32,9 +32,11 @@ pub const FormatOptions = struct {
 /// this format:
 /// `{[argument][specifier]:[fill][alignment][width].[precision]}`
 ///
-/// Each word between `[` and `]` is a parameter you have to replace with something:
+/// Above, each word including its surrounding [ and ] is a parameter which you have to replace with something:
```
Vexu, line 35: While we're clarifying things, I think "have to" is a poor choice of words when nearly all of these parameters can be omitted.

Vexu: Actually the full comment explains this clearly enough.

PR #11933, opened 2022-06-25
stage2: wasm-linker - Place decls in the correct segment and order segments

Luukdegram:
Previously, all decls would be placed within the 'rodata' segment regardless of constness. With this change, we correctly put them into segments such as 'data' and 'bss'. The linker now also orders the segments in such a way to always place the bss segment last, which allows us to completely omit this segment in the case where the memory is exported and not imported from the runtime. This allows smaller binaries as the runtime itself will zero-initialize the memory for us.

### DIFF
--- a/src/link/Wasm.zig
+++ b/src/link/Wasm.zig
@@ -1255,6 +1255,9 @@ fn parseAtom(self: *Wasm, atom: *Atom, kind: Kind) !void {
 }
 
 fn allocateAtoms(self: *Wasm) !void {
+    // first sort the data segments
+    try sortDataSegments(self);
+
     var it = self.atoms.iterator();
     while (it.next()) |entry| {
         const segment = &self.segments.items[entry.key_ptr.*];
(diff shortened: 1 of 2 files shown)


kubkon:
> Previously, all decls would be placed within the 'rodata' segment regardless of constness. With this change, we correctly put them into segments such as 'data' and 'bss'. The linker now also orders the segments in such a way to always place the bss segment last, which allows us to completely omit this segment in the case where the memory is exported and not imported from the runtime. This allows smaller binaries as the runtime itself will zero-initialize the memory for us.

Do you overwrite the entire file on every flush per incremental update? If not, ensuring correct order for segments may require selectively moving segments in file (via `copyRangeAll` or related) if a `Decl` grows over the segment's size, etc. But if you wipe the entire file and write to it from scratch on every `flush`, this is obviously not an issue :-) I'm asking since this is what I have to deal with with MachO and to some extend Elf (as far as `.bss` and TLS will go).

Luukdegram:
> Do you overwrite the entire file on every flush per incremental update? If not, ensuring correct order for segments may require selectively moving segments in file (via `copyRangeAll` or related) if a `Decl` grows over the segment's size, etc. But if you wipe the entire file and write to it from scratch on every `flush`, this is obviously not an issue :-) I'm asking since this is what I have to deal with with MachO and to some extend Elf (as far as `.bss` and TLS will go).

The entire file indeed gets overwritten in its entirety during `flush`.

kubkon:
> > Do you overwrite the entire file on every flush per incremental update? If not, ensuring correct order for segments may require selectively moving segments in file (via `copyRangeAll` or related) if a `Decl` grows over the segment's size, etc. But if you wipe the entire file and write to it from scratch on every `flush`, this is obviously not an issue :-) I'm asking since this is what I have to deal with with MachO and to some extend Elf (as far as `.bss` and TLS will go).
> 
> The entire file indeed gets overwritten in its entirety during `flush`.

Gotcha, thanks for clarifying!

src/link/Wasm.zig
```diff
@@ -1259,6 +1281,36 @@ fn allocateAtoms(self: *Wasm) !void {
     }
 }
 
+fn sortDataSegments(self: *Wasm) !void {
```
kubkon, line 1284: This might not be very useful to you, but this is what we do in MachO using optional indexes to every known output section that may be in the file https://github.com/ziglang/zig/blob/960c14206094484cee056b8806b9b8f75a84891d/src/link/MachO.zig#L5504 Time-complexity wise, not sure if this results in anything more optimal or whether we should even worry about this given that there are so few indexes to sort.

kubkon (approved):
Awesome work!

src/link/Wasm.zig
kubkon: I also use this trick in zld/ELF https://github.com/kubkon/zld/blob/f0de089a7db6dce6b9a14d6645ab43287728c0ca/src/Elf.zig#L1132

Luukdegram: That would work if all possible segments were known. However, when compiling to a relocatable object file, each data symbol gets its own segment. E.g. `.rodata.my_symbol`. Only for final binaries do we merge them into a single segment. (wasm-ld even allows you to not merge segments). For this reason I choose this option, as I actually remember checking the ELF and MachO implementation when I did this for Coff :)

kubkon: Hmm, so you are saying that in Wasm we permit for a priori "unknown" output sections?

Luukdegram: Segments, yes. They are not standardized in any way. Only the sections are (types, imports, code, data, etc). Those are already ordered correctly simply within `flush` by emitting them manually in the correct order.

PR #11936, opened 2022-06-26
IO api redesign

igor84:
The pull request is a bit bigger cause I had to do everything in order to make sure that everything can be done as I imagined.

This pull requests removes `SeekableStream` as a separate entity and instead adds a `SeekableReader` and a `SeekableWriter` next to regular `Reader` and `Writer`. That allowed me to also solve the following issues:

1. There was no correct way to use `BufferedReader` with seeking. Now `BufferedReader` adds its own seek methods if the reader it is wrapping is a `SeekableReader`.
2. There was no way to make `StreamSource` wrap a `File` with a `BufferedReader` and still provide seeking. Instead of storing `File` it now stores either `File.SeekableReader` or `File.SeekableWriter`. That way it provides buffering and seeking but can not be used for both writing to a file and reading from it at the same time.


andrewrk:
I'm sorry for taking so long to review this pull request. It looks reasonable, although I probably would have wanted to discuss the use of `usingnamespace`. It took me so long to get around to reviewing, that the branch has now conflicts and needs to re-run against the new CI checks. I'm afraid I will be putting the burden of this on you; I would be happy to give this another look if you rebase and open a new PR against master branch. If you do so, I will make an effort to prioritize it so that this does not happen to you again.

igor84:
I opened a new pull request.

PR #11942, opened 2022-06-27
Move passing stage1 compile error tests to stage2

Vexu:
There sure are a lot of these.


test/cases/compile_errors/invalid_struct_field.zig
```diff
@@ -12,8 +12,10 @@ export fn g() void {
...
-// backend=stage1
+// backend=stage2
 // target=native
 //
-// tmp.zig:4:6: error: no member named 'foo' in struct 'A'
-// tmp.zig:10:16: error: no member named 'bar' in struct 'A'
+// :4:7: error: no field named 'foo' in struct 'tmp.A'
+// :1:11: note: struct declared here
+// :10:17: error: no field named 'bar' in struct 'tmp.A'
+// :1:11: note: struct declared here
```
Vexu, line 21: @kubkon can you take a look at this test? I tried fixing it like in d1e46006758a82f039d23ea26030e3b685a5fa86 but that broke other tests.

kubkon: Will do!

test/cases/compile_errors/access_non-existent_member_of_error_set.zig
```diff
@@ -0,0 +1,12 @@
+const Foo = error{A};
+comptime {
+    const z = Foo.Bar;
+    _ = z;
+}
+
+// error
+// backend=stage2
```
kubkon, line 8: Just to confirm you are targeting the right backend: `stage2` implies incremental, while `llvm` implies stage2 LLVM. I now since how that can get confusing. I will redo the `Backend` to include `stage2_llvm` and `stage2_incremental` instead, and once both get close to finish, we can consolidate the two into one `stage2` or something.

Vexu: I don't think it'll matter too much for compile error tests but I did test them without the llvm backend just to be sure. If you think it's more trouble than it's worth I can change them to use `backend=llvm` instead.

kubkon: Sadly, it will be a problem on non-x86_64 hosts such as aarch64 where some lowerings such as `airRetLoad` are not yet implemented. Otherwise, I agree it should be no difference. Anyhow, leave it as-is for now. I just wanted to double check. I will investigate some more next why the tests are failing for you. In the meantime, here's first improvement to the harness: https://github.com/ziglang/zig/pull/11952

kubkon:
Uhm, did the CI take a break? 😅

andrewrk:
I'll go ahead and do a force push to the branch to try and refresh it.

Vexu:
Cannot reproduce macos CI failure by forcing `"native"` to be `"native-macos"`.

kubkon:
> Cannot reproduce macos CI failure by forcing `"native"` to be `"native-macos"`.

I'll try helping out but I only have access to my M1 MBP at this very moment (until Sunday). I will try figuring out on arm64 anyhow and perhaps the root cause for the failure will be common. I'll report back if I find something.

kubkon:
@Vexu please don't force push anymore. I have found the culprit and you've been disabled the tests that are actually fine.

Vexu:
Ok. Most of the tests I moved to use `backend=llvm` were because of TODOs in the arm backends. I also seem to have accidentally committed my debugging patches that'll need to be removed.

kubkon:
> Ok. Most of the tests I moved to use `backend=llvm` were because of TODOs in the arm backends. I also seem to have accidentally committed my debugging patches that'll need to be removed.

742b7c6 should now unblock you - `endless_loop_in_function_evaluation` is causing a segfault in the frontend on macos and after chatting with Andrew he suggested we disable the test for now until it's fully implemented on all hosts.

PR #11943, opened 2022-06-27
zig fmt: fix idempotency with newlines surrounding doc comment

### LINKED ISSUE #11802: fmt: idempotency with newlines above and below doc comment
### Zig Version

0.10.0-dev.2398+b08d32ceb

### Steps to Reproduce

```zig
​


///

fn x() void {}
```
```
$ zig fmt
```
```zig
​
///
fn x() void {}
```
```
$ zig fmt
```
```zig
///
fn x() void {}
```

### Expected Behavior

```zig
​


///

fn x() void {}
```
```
$ zig fmt
```
```zig
///
fn x() void {}
```

### Actual Behavior

It takes two runs of `zig fmt` to fully format the code.

luehmann:
This should fix #11802 

I'd be happy to hear any suggestions/improvements as I'm not really sure if this is the best way to go about it.

### DIFF
--- a/lib/std/zig/render.zig
+++ b/lib/std/zig/render.zig
@@ -2482,7 +2482,10 @@ fn renderDocComments(ais: *Ais, tree: Ast, end_token: Ast.TokenIndex) Error!void
     }
     const first_tok = tok;
     if (first_tok == end_token) return;
-    try renderExtraNewlineToken(ais, tree, first_tok);
+
+    if (tok != 0 and token_tags[first_tok - 1] != .l_brace) {
+        try renderExtraNewlineToken(ais, tree, first_tok);
+    }
 
     while (token_tags[tok] == .doc_comment) : (tok += 1) {
         try renderToken(ais, tree, tok, .newline);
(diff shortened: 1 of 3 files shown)


lib/std/zig/render.zig
```diff
@@ -2482,7 +2482,10 @@ fn renderDocComments(ais: *Ais, tree: Ast, end_token: Ast.TokenIndex) Error!void
     }
     const first_tok = tok;
     if (first_tok == end_token) return;
-    try renderExtraNewlineToken(ais, tree, first_tok);
+
+    if (tok != 0 and token_tags[first_tok - 1] != .l_brace) {
```
ifreund, line 2486: Hmm, this `!= .l_brace` check would fail if this function was used for doc comments on function arguments. Those however seem to be handled by `renderFnProto()` directly though currently.

Would it be safe to assert the previous token is not `.l_paren` for now so that if this is changed we don't forget to update this check? A comment explaining this would be in order as well.

ifreund (commented):
Thanks for looking into this!

lib/std/zig/render.zig
luehmann: Adding an assertion to prevent accidental use elsewhere would definitely make sense. Should there be just a comment for the assertion or for the whole new condition as well?

Issue #11918, opened 2022-06-23
zig0 does not properly handle baseline cpu features


FanShupei:
### Zig Version

0.9.1

### Steps to Reproduce

Follow official build instructions on RV64GC linux platform

### Expected Behavior

Generate a working stage 2 compiler `zig`

### Actual Behavior

Stage 2 compiler `zig` does not work at all. It segfault at compiling every trivial program (infinite recursion in __udivmodti4).

When compiling `zig` by zig0, it generates several warnings 
```
Hard-float 'd' ABI can't be used for a target that doesn't support the D instruction set extension (ignoring target-abi)
```

### Analysis

After investigation, I have found that the root cause is zig0 does not properly set `llvm_cpu_names` and `llvm_cpu_features` when handling `-mcpu baseline`, currently it just leaves the two fields empty.

It may just work (though generating poor code) at x86_64 platform, but it is totally inadequate at rv64gc platform. Without setting cpu names or cpu features, the default cpu used by llvm is rv64i, lacking even basic instructions like multiplication. What's worse, currently even compiler-rt does not support rv64i well. This is why compile-rt contains infinite recursion.

I can not test it on master branch because now compiling `zig1.o` requires even more memory. but it seems the problem remains.

andrewrk:
The short-term solution to this problem is to hard-code a different riscv64 cpu features set for zig0.

The eventual solution to this problem is that zig0 will be pure C code that is the zig self-hosted compiler, compiled to C code with the C backend, cleaned up, and with only the C backend available. The bootstrap process will create zig0 from zig0.c, and then use that to produce zig1.c, which will be compiled into zig1, which will then be used to create zig2, which will be used to create the final binary.

BratishkaErik:
Also see https://github.com/ziglang/zig/issues/6830#issuecomment-718142448:

> OK, confirmed that the issue is solved by rebuilding, passing [the stage1 that fails the translate-c tests] for `-DZIG_EXECUTABLE` in order to rebuild zig1.o using stage1 rather than stage0. So this confirms @LemonBoy's hypothesis that stage0 is passing incorrect LLVM CPU features for "baseline", and it also leads us towards a fix, if we can just get cmake to cooperate.

Issue #11848, opened 2022-06-11
compiler crash with generic function instantiation, invalid decl_index when adding new instantiation to the set


andrewrk:
A CI failure happened on 9360cfebc722d99a54b576fff8bcfc0f0354ad41 which is unrelated to the actual changes:

```
2022-06-11T19:37:35.7793190Z thread 93572 panic: reached unreachable code
2022-06-11T19:37:35.7797040Z Analyzing /Users/runner/work/1/s/lib/std/heap/arena_allocator.zig: heap/arena_allocator.zig:ArenaAllocator.alloc
2022-06-11T19:37:35.7798400Z       %380 = ref(%259) 
2022-06-11T19:37:35.7799350Z       %381 = field_ptr(%380, "child_allocator") 
2022-06-11T19:37:35.7799920Z       %382 = field_call_bind(%381, "resize") 
2022-06-11T19:37:35.7800510Z       %383 = dbg_stmt(21, 56)
2022-06-11T19:37:35.7800990Z       %384 = param_type(%382, 0)
2022-06-11T19:37:35.7801520Z       %385 = load(%276) 
2022-06-11T19:37:35.7802370Z       %386 = field_val(%385, "data") 
2022-06-11T19:37:35.7803660Z       %387 = param_type(%382, 1)
2022-06-11T19:37:35.7804280Z     > %388 = call(.auto, %382, [%386, %376]) 
2022-06-11T19:37:35.7804830Z       %389 = is_non_null(%388) 
2022-06-11T19:37:35.7805330Z       %390 = condbr(%389, {
2022-06-11T19:37:35.7805860Z         %392 = optional_payload_unsafe(%388) 
2022-06-11T19:37:35.7806390Z         %393 = store_node(%379, %392) 
2022-06-11T19:37:35.7807610Z         %411 = break(%391, @Zir.Inst.Ref.void_value)
2022-06-11T19:37:35.7808100Z       }, {
2022-06-11T19:37:35.7808550Z         %394 = dbg_block_begin())
2022-06-11T19:37:35.7809050Z         %395 = dbg_stmt(23, 17)
2022-06-11T19:37:35.7809520Z         %396 = ref(%259) 
2022-06-11T19:37:35.7810020Z         %397 = field_call_bind(%396, "createNode") 
2022-06-11T19:37:35.7810560Z         %398 = dbg_stmt(23, 47)
2022-06-11T19:37:35.7811050Z         %399 = param_type(%397, 0)
[...]

PR #11951, opened 2022-06-28
zig0: handle baseline cpu features for rv64

FanShupei:
As discussed in #11918, I just follow the short term solution, adding hard-coded cpu features for rv64.

I have tested locally in rv64 platform, now zig0 successfully produces a working compiler.


andrewrk:
Thank you! The CI failure is  #11848.

BratishkaErik:
Thank you for this patch!

PR #11953, opened 2022-06-28
Correct the init function to use initLocked() for initializing std.event.Lock.

kaihlavirta:
#### init funcion used init() to initialize a std.event.Lock. std.event.Lock does not have init() function. Corrected to initLocked().

### DIFF
--- a/lib/std/event/group.zig
+++ b/lib/std/event/group.zig
@@ -35,7 +35,7 @@ pub fn Group(comptime ReturnType: type) type {
             return Self{
                 .frame_stack = Stack.init(),
                 .alloc_stack = AllocStack.init(),
-                .lock = Lock.init(),
+                .lock = Lock.initLocked(),
                 .allocator = allocator,
             };
         }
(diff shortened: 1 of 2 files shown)


lib/std/event/group.zig
```diff
@@ -35,7 +35,7 @@ pub fn Group(comptime ReturnType: type) type {
             return Self{
                 .frame_stack = Stack.init(),
                 .alloc_stack = AllocStack.init(),
-                .lock = Lock.init(),
+                .lock = Lock.initLocked(),
```
Vexu, line 38: Looking at the commit which caused this a31d9f92f282a878836a3ecac5a48d5f4037868c, the lock is supposed to be initiated in an unlocked state:
```suggestion
                .lock = .{},
```

Also, could you apply the same change to `std.event.Locked` as mentioned in the issue?

Vexu (changes requested):

kaihlavirta:
Updated :
* lib/std/event/group.zig
* lib/std/event/locked.zig

to use `.lock = .{}` instead of `.lock = Lock.init()` or `.lock = Lock.initLocked()`

haolian9:
should tests in [here](https://github.com/kaihlavirta/zig/blob/6b260f7968c21edad80f67b38441bb2d174ec29f/lib/std/event/group.zig#L125) be enabled?

Vexu:
They could be enabled if they were fixed but evented IO is not being tested in the CI currently due to stage1 memory usage.

Issue #8465, opened 2021-04-07
Proposal: make std.math.Complex ABI-compatible with C _Complex


ehaas:
C99 defines three complex types: `float _Complex`, `double _Complex`, and `long double _Complex`, and a header `<complex.h>` that defines various complex number math functions. According to the standard, "each complex type has the same representation and alignment requirements as an array type containing exactly two elements of the corresponding real type; the first element is equal to the real part, and the second element to the imaginary part, of the complex number". If Zig's `std.math.Complex` type conforms to this representation, Zig's complex number types can seamlessly interoperate with C code without any additional copies.

Basic proof of concept that works on `x86_64-macos-gnu`; on this platform a 2-element struct of `f64` matches a two-element array but I'm not sure if that is always the case. Using a 2-element array in the struct is blocked on #1481 before it can actually be used with the C ABI. 

```zig
const std = @import("std");
const MyComplex = extern struct {
    re: f64,
    im: f64,
    // Zig complex functions go here
};

/// defined in <complex.h>
/// prototype: double complex csqrt(double complex z);
extern fn csqrt(arg: MyComplex) MyComplex;

const epsilon = 0.0000001;
test "complex" {
    var foo = MyComplex {.re = 12, .im = 16};
    var root = csqrt(foo);
    std.testing.expect(std.math.approxEqAbs(f64, root.re, 4.0, epsilon));
    std.testing.expect(std.math.approxEqAbs(f64, root.im, 2.0, epsilon));
}
```

```
➜  zig test -lm mycomplex.zig
All 1 tests passed.
```

ehaas:
Turns out this is not so simple - the C standard specifies the layout of complex types but it does not specify the calling convention. Therefore I do not believe it would be possible to solve this without language-level support like `c_float_complex` and `c_double_complex`. Concrete example, output from godbolt.org

C Code:
```c
struct my_complex {
    float x, y;
};
struct my_complex return_struct() {
    return (struct my_complex) {.x = 1.0, .y = 2.0};
}

float _Complex return_complex() {
    return (float _Complex) {1.0, 2.0};
}
```

Using x86-64 clang 12.0.1, with `-O1` flag, the output for both functions is the same, so returning `float _Complex` is ABI-equivalent to returning a struct of 2 floats:
```asm
.LCPI0_0:
        .long   0x3f800000                      # float 1
        .long   0x40000000                      # float 2
        .zero   4
        .zero   4
return_struct:                          # @return_struct
[...]

PR #11961, opened 2022-06-29
std.math.complex:  c_Complex for C maths libraries

f3fora:
Popular mathematical C libraries require complex numbers, which are often represented
with a 2 element struct. 
The `Complex(...)` type does not guarantee the C ABI compatibility.
Thus, let's define once and only once a `c_Complex(...)` type in `std` to interact in 
an unique way and avoid re-definitions.

The name `c_Complex` is chosen in analogy with `c_int`.

E.g. a function for complex number in `cblas.h` header:
```
void   cblas_zdotc_sub(const CBLAS_INT N, const void *X, const CBLAS_INT incX,
                       const void *Y, const CBLAS_INT incY, void *dotc);
```
where usually `void *` is a pointer to an array of a struct of two elements
(in this particular case, of type `double`).

### DIFF
--- a/lib/std/math/complex.zig
+++ b/lib/std/math/complex.zig
@@ -24,6 +24,15 @@ pub const tanh = @import("complex/tanh.zig").tanh;
 pub const tan = @import("complex/tan.zig").tan;
 
 /// A complex number consisting of a real an imaginary part. T must be a floating-point value.
+/// The present struct is compatible with C ABI. Prefer `Complex` if C interaction is not needed.
+pub fn c_Complex(comptime T: type) type {
+    return extern struct {
+        re: T,
+        im: T,
+    };
+}
+
+/// A complex number consisting of a real an imaginary part. T must be a floating-point value.
 pub fn Complex(comptime T: type) type {
     return struct {
         const Self = @This();
@@ -117,6 +126,14 @@ pub fn Complex(comptime T: type) type {
         pub fn magnitude(self: Self) T {
             return @sqrt(self.re * self.re + self.im * self.im);
         }
+        
+        /// Return a C ABI compatible struct complex number.
+        pub fn asExtern(self: Self) c_Complex(T) {
+            return c_Complex(T){
+                .re = self.re,
+                .im = self.im,
+            };
+        }
     };
 }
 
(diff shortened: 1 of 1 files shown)


f3fora:
If I understand correctly from [documentation](https://ziglang.org/documentation/master/#extern-struct) and from [this comment](https://github.com/ziglang/zig/issues/6700#issuecomment-710715422), order in `struct` is not guaranteed. Obviously, it is crucial in complex numbers.

ehaas:
The C standard defines the size and alignment of `_Complex` types (same size and alignment as an array of 2 elements of the corresponding real type) but it does not define the calling convention. So I don't think it is possible to implement this without compiler support. If I recall correctly `i386-linux-gnu` will treat a return type of `_Complex float` differently than a return type of `"struct with 2 floats"`

f3fora:
`<complex.h>` was introduced in C99, while some C mathematical and scientific libraries can be older and deal with complex numbers in their own way.

For example, LAPACK defines `lapack_complex_{float,double}` as `_Complex` or as a struct `{ double real, imag; }` according to some constant; GSL defines complex numbers with `#define GSL_COMPLEX_DEFINE(R, C) typedef struct { R dat[2]; } C ;`; BLAS accepts array of complex number as function arguments as `void *`.

With an `extern struct` all these definitions are a 'cast' away and it allows a simple and unique way to interact with C libraries.

Moreover, from another point of view, [a complex number is nothing more than an ordered pair of two real numbers](https://www.sciencedirect.com/topics/physics-and-astronomy/complex-numbers). After all @zigazeljko 's idea is not too wrong.

ehaas:
> With an extern struct all these definitions are a 'cast' away and it allows a simple and unique way to interact with C libraries.

This is not guaranteed to be true - it's not required that a C compiler treat `_Complex T` and `struct C { T r; T i; }` (where `T` is some floating point type) the same, specifically when passing or returning it by value to a function. It's also not required to treat it the same as `T[2]` with regard to calling convention.

Adding this would be a subtle footgun - it would work on many platforms, but not, for example, on `i386-linux-gnu`. The only way to properly solve it is with a `c_complex` type in the language.

Vexu:
One option is to inspect the LLVM ir generated by `clang -S -emit-llvm --target=?? file.c` and defining the type in a switch on the target and erroring on targets where `_Complex` is not representable by a struct like mips.
```c
_Complex double a;
struct Complex { double re[2]; } b;
// or struct Complex { double re, im; } b;
_Complex double foo(_Complex double a) {
    return a;
}
struct Complex bar(struct Complex a) {
    return a;
}
```

f3fora:
See #8465

PR #11964, opened 2022-06-29
std.math.big.int: breaking API changes to prevent UAF

### LINKED ISSUE #11897: Stage2 miscompiles call to `std.math.big.int.Managed.shiftLeft`
```zig
test {
    var p = try std.math.big.int.Managed.initSet(testing.allocator, 1);
    defer p.deinit();
    var a2 = try p.clone();
    defer a2.deinit();

    var shift: isize = 54;
    try a2.shiftLeft(a2, @intCast(usize, shift)); // segfault

    // var shift2 = @intCast(usize, shift);
    // try a2.shiftLeft(a2, shift2); // works as expected
}
```

The biggest difference between the two seems to be that the in the second one the `a2` passed by value is copied first but the doc comments on `shiftLeft` says that aliasing here should be okay:
```diff
diff --git a/foo.ll b/foo.ll
index 7368ad2c8..29f41a7d7 100644
--- a/foo.ll
+++ b/foo.ll
@@ -1,7 +1,7 @@
 ; Function Attrs: nounwind
 define internal fastcc i16 @a.test_0(%builtin.StackTrace* nonnull %0) unnamed_addr #0 {
 Entry:
-  %1 = alloca i64, align 8
+  %1 = alloca %math.big.int.Managed, align 8
   %2 = alloca i64, align 8
   %3 = alloca { %math.big.int.Managed, i16 }, align 8
   %4 = alloca %math.big.int.Managed, align 8
@@ -46,9 +46,12 @@ TryCont2:                                         ; preds = %TryCont
   call void @llvm.dbg.declare(metadata %math.big.int.Managed* %4, metadata !4856, metadata !DIExpression()),
   store i64 54, i64* %2, align 8,
   call void @llvm.dbg.declare(metadata i64* %2, metadata !4857, metadata !DIExpression()),
-  %23 = load i64, i64* %2, align 8,
-  %24 = icmp sge i64 %23, 0,
-  br i1 %24, label %Then, label %Else,
+  %23 = bitcast %math.big.int.Managed* %1 to i8*,
+  %24 = bitcast %math.big.int.Managed* %4 to i8*,
+  call void @llvm.memcpy.p0i8.p0i8.i64(i8* align 8 %23, i8* align 8 %24, i64 40, i1 false),
+  %25 = load i64, i64* %2, align 8,
+  %26 = icmp sge i64 %25, 0,
+  br i1 %26, label %Then, label %Else,
 
 Then:                                             ; preds = %TryCont2
   br label %Block,
@@ -58,18 +61,15 @@ Else:                                             ; preds = %TryCont2
   unreachable,
 
 Block:                                            ; preds = %Then
-  store i64 %23, i64* %1, align 8,
-  call void @llvm.dbg.declare(metadata i64* %1, metadata !4859, metadata !DIExpression()),
-  %25 = load i64, i64* %1, align 8,
-  %26 = call fastcc i16 @math.big.int.Managed.shiftLeft(%builtin.StackTrace* %0, %math.big.int.Managed* %4, %math.big.int.Managed* %4, i64 %25),
-  %27 = icmp ne i16 %26, 0,
-  br i1 %27, label %TryRet3, label %TryCont4,
+  %27 = call fastcc i16 @math.big.int.Managed.shiftLeft(%builtin.StackTrace* %0, %math.big.int.Managed* %4, %math.big.int.Managed* %1, i64 %25),
+  %28 = icmp ne i16 %27, 0,
+  br i1 %28, label %TryRet3, label %TryCont4,
 
 TryRet3:                                          ; preds = %Block
   call fastcc void @math.big.int.Managed.deinit(%math.big.int.Managed* %4),
```

This is blocking `std.math.big.Rational.toFloat` tests.

andrewrk:
Many of the Managed methods accepted by-val parameters which could
reference Limb slices that became invalid memory after any
ensureCapacity calls. Now, Managed methods accept `*const Managed`
parameters so that if the function allows aliasing and the
ensure-capacity call resizes the Limb slice, it also affects the
aliased parameters, avoiding use-after-free bugs.

This is a breaking change that reduces the requirement for callsites to
manually make the ensure-capacity changes prior to calling many of the
Managed methods.

Closes #11897

### DIFF
--- a/lib/std/math/big/int.zig
+++ b/lib/std/math/big/int.zig
@@ -1391,28 +1391,28 @@ pub const Mutable = struct {
 
             if (B == 0) {
                 // t_big = x % y, r is unused
-                try r.divTrunc(&t_big, x.toConst(), y.toConst());
+                try r.divTrunc(&t_big, &x, &y);
                 assert(t_big.isPositive());
 
                 x.swap(&y);
                 y.swap(&t_big);
             } else {
                 var storage: [8]Limb = undefined;
-                const Ap = fixedIntFromSignedDoubleLimb(A, storage[0..2]).toConst();
-                const Bp = fixedIntFromSignedDoubleLimb(B, storage[2..4]).toConst();
-                const Cp = fixedIntFromSignedDoubleLimb(C, storage[4..6]).toConst();
-                const Dp = fixedIntFromSignedDoubleLimb(D, storage[6..8]).toConst();
+                const Ap = fixedIntFromSignedDoubleLimb(A, storage[0..2]).toManaged(limbs_buffer.allocator);
+                const Bp = fixedIntFromSignedDoubleLimb(B, storage[2..4]).toManaged(limbs_buffer.allocator);
+                const Cp = fixedIntFromSignedDoubleLimb(C, storage[4..6]).toManaged(limbs_buffer.allocator);
+                const Dp = fixedIntFromSignedDoubleLimb(D, storage[6..8]).toManaged(limbs_buffer.allocator);
 
                 // t_big = Ax + By
-                try r.mul(x.toConst(), Ap);
-                try t_big.mul(y.toConst(), Bp);
-                try t_big.add(r.toConst(), t_big.toConst());
+                try r.mul(&x, &Ap);
+                try t_big.mul(&y, &Bp);
+                try t_big.add(&r, &t_big);
 
                 // u = Cx + Dy, r as u
                 try tmp_x.copy(x.toConst());
-                try x.mul(tmp_x.toConst(), Cp);
-                try r.mul(y.toConst(), Dp);
-                try r.add(x.toConst(), r.toConst());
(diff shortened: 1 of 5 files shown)


lib/std/math/big/int.zig
```diff
@@ -2660,57 +2660,56 @@ pub const Managed = struct {
 
     /// r = a + scalar
     ///
-    /// r and a may be aliases. If r aliases a, then caller must call
-    /// `r.ensureAddScalarCapacity` prior to calling `add`.
-    /// scalar is a primitive integer type.
```
eyelash, line 2665: Did you intend to delete this line or was it deleted accidentally?

andrewrk: Intended. It was replaced with the sentence " r and a may be aliases". The other part is no longer a requirement of the caller.

eyelash: So `scalar` can now be something else than a primitive integer type?

PR #11965, opened 2022-06-29
std.crypto.hash: allow creating hash functions from compositions

jedisct1:
A hash function cascade was a common way to avoid length-extension attacks with traditional hash functions such as the SHA-2 family.

Add `std.crypto.hash.composition` to do exactly that using arbitrary hash functions, and pre-define the common SHA2-based ones.

With this, we can now sign and verify Bitcoin signatures in pure Zig.


daurnimator:
Is there a circumstance where there is e.g. 3 hashes instead?
i.e. should the `Composition` function take an arbitrary number of hash functions?

PR #11967, opened 2022-06-30
stage2: lower float negation explicitly + modify hash/eql logic for floats

### LINKED ISSUE #11853: Stage2 can't negate a NaN at runtime
In the AIR negation is implemented as `0 - val` which means that a NaN cannot be negated using the `-` operator. I'm not sure how desirable negative NaNs are but the behavior should be consistent at comptime and runtime.
```zig
test {
    std.debug.print("{d}\n", .{-math.nan(f32)}); // nan
    std.debug.print("{d}\n", .{-comptime math.nan(f32)}); // -nan
}
```

andrewrk:
Rather than lowering float negation as `0.0 - x`, this commit makes Zig lower it with a negation AIR instruction. It also adds compiler-rt functions for f128 and f80 negation.

This revealed a problem with the way generic functions work for floats:

Zig guarantees the memory layout of f16, f32, f64, f80, and f128 which
means for generic function purposes, values of these types need to be
compared on the basis of their bits in memory. This means nan-packing
can be used with generic functions, for example.

For comptime_float, the sign is observable, whether it is nan is
observable, but not any more kinds of bit patterns are observable.

This fixes the std.fmt tests that check printing "-nan".

closes #11853


Vexu:
> It also adds compiler-rt functions for f128 and f80 negation.

Stage1 just directly generates an XOR with an appropriate constant, I think stage2 should too.

PR #11970, opened 2022-06-30
std.heap: Calling deinit should also reset the ArenaAllocator's state

ryuukk:
To make it easier to use in a loop


Vexu:
You committed trailing whitespace, you need to run `zig fmt . --exclude test/cases`.

Vexu:
The current convention is for `deinit` to leave whatever it deinitialized in an invalid state. I think it would be better to instead have a `reset` function and `deinit` can then be implemented using that.

PR #11972, opened 2022-06-30
std.json: expose encodeJsonString and encodeJsonStringChars

marler8997:
Expose 2 functions from std.json.  These functions take a slice of bytes
and forward them to a given writer as a JSON encoded string.

The use case I have for this is in a custom JsonStringWriter.  This writer
takes data and automatically encodes it as JSON string characters and
forwards it to an underlying writer.  I use this JsonStringWriter in
combination with std.fmt.format to go directly from a format string/arg
pair to JSON.  This way I don't have to format my string into a separate
buffer first and encode it afterwards, which avoids the need to create
a temporary buffer to hold the unencoded but formatted string.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -2163,45 +2163,51 @@ fn outputUnicodeEscape(
     }
 }
 
-fn outputJsonString(value: []const u8, options: StringifyOptions, out_stream: anytype) !void {
-    try out_stream.writeByte('\"');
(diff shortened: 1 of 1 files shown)


marler8997:
Here's a gist with a simplified example of my use case: https://gist.github.com/marler8997/47c9fd80c8ac26e5f4f730878f672840

PR #11975, opened 2022-07-01
fix deleteFileAbsoluteW type

codehz:


Vexu:
This doesn't look like a typo to me?

codehz:
> This doesn't look like an incorrect type to me?

Easy verify: try use it, it will report type mismatch between cwd().deleteFileW 's []const u16 and [*:0]const u16

Vexu:
Sure but if you look at the other `*AbsoluteW` functions in this file most of them use `[*:0]const u16` so either way a lot more needs to be fixed.

codehz:
> Sure but if you look at the other `*AbsoluteW` functions in this file most of them use `[*:0]const u16` so either way a lot more needs to be fixed.

Since the UNICODE_STRING allow to contains embeded ZERO, I think all of those should convert to []const u16 version

matu3ba:
> Since the UNICODE_STRING allow to contains embeded ZERO, I think all of those should convert to []const u16 version

It should be fairly quick to test on Windows, if your assumption is correct and to add a test for this.

Vexu:
Closing as stale draft to clear up the pr queue a bit, feel free to open a new pr.

PR #11976, opened 2022-07-01
fix error message importing file starting with module root path

### LINKED ISSUE #9537: Importing a file with a name that starts with the package name outside of the package path provides incorrect error message
With the following directory structure:

```
src/main.zig
src/game_api.zig
src/game/game.zig
```

build.zig snippets:
```
const exe = b.addExecutable("ion", "src/main.zig");
...
const game = b.addSharedLibrary("game", "src/game/game.zig", .unversioned);
```

If `game.zig` has `const game_api = @import("../game_api.zig");`, then the resulting error is:

```
$ zig build
C:\Users\kcbanner\projects\ion\src\game\game.zig:4:26: error: unable to load 'C:\\Users\\kcbanner\\projects\\ion\\src\\game\api.zig': FileNotFound
const game_api = @import("../game_api.zig");
```

Not how it's trying to import `game\api.zig`.

However, if `game_api.zig` is changed to say, `foo_api.zig`, we get the correct message:
```
$ zig build
.\game\game.zig:5:18: error: import of file outside package path: '../foo_api.zig'
const game_api = @import("../foo_api.zig");
```

I believe the issue is that in `Module.zig:importFile`:

```
    if (!mem.startsWith(u8, resolved_path, resolved_root_path)) {
        return error.ImportOutsidePkgPath;
    }
```

Here `resolved_root_path` is `"C:\\Users\\kcbanner\\projects\\ion\\src\\game"`, so it matches with the `resolved_path` of `"C:\\Users\\kcbanner\\projects\\ion\\src\\game_api.zig"`. I think the solution might be to add the path separator to the end of `resolved_root_path`? 

I can make a PR for this, if that is an acceptable solution.

LordMZTE:
closes #9537

Note that, for some reason, I couldn't get this error to trigger. I have ensured that the check I changed triggers when it should, though. Not sure what's preventing it from erroring for me.

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -4367,7 +4367,7 @@ pub fn importFile(
     if (!mem.startsWith(u8, resolved_path, resolved_root_path) or
         // This prevents this check from triggering when the name of the
         // imported file starts with the root path's directory name.
-        mem.indexOfAny(u8, &.{resolved_path[resolved_root_path.len]}, "/\\") == null)
+        std.fs.path.isSep(resolved_path[resolved_root_path.len]))
     {
         return error.ImportOutsidePkgPath;
     }
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -4364,7 +4364,11 @@ pub fn importFile(
     const resolved_root_path = try std.fs.path.resolve(gpa, &[_][]const u8{cur_pkg_dir_path});
     defer gpa.free(resolved_root_path);
 
-    if (!mem.startsWith(u8, resolved_path, resolved_root_path)) {
+    if (!mem.startsWith(u8, resolved_path, resolved_root_path) or
+        // This prevents this check from triggering when the name of the
+        // imported file starts with the root path's directory name.
+        mem.indexOfAny(u8, &.{resolved_path[resolved_root_path.len]}, "/\\") == null)
+    {
         return error.ImportOutsidePkgPath;
     }
     // +1 for the directory separator here.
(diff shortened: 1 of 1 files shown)


src/Module.zig
```diff
@@ -4364,7 +4364,11 @@ pub fn importFile(
     const resolved_root_path = try std.fs.path.resolve(gpa, &[_][]const u8{cur_pkg_dir_path});
     defer gpa.free(resolved_root_path);
 
-    if (!mem.startsWith(u8, resolved_path, resolved_root_path)) {
+    if (!mem.startsWith(u8, resolved_path, resolved_root_path) or
+        // This prevents this check from triggering when the name of the
+        // imported file starts with the root path's directory name.
+        mem.indexOfAny(u8, &.{resolved_path[resolved_root_path.len]}, "/\\") == null)
```
Vexu, line 4370:
```suggestion
        std.fs.path.isSep(resolved_path[resolved_root_path.len]))
```

Vexu (changes requested):
This logic was directly ported from stage1, have you checked whether this happens with it?

LordMZTE:
Doesn't seem to happen in stage1, but I didn't really debug it in stage1. Maybe there's something missing in my attempt to reproduce this.

LordMZTE:
Hmm, I'm not sure why, but for some reason, the code to reproduce this compiles just fine on my branch. I doubt that this is caused by my changes though.

kcbanner:
I wonder if it's OS related? I have only tried my repro steps from #9537 on Windows, I'm not sure which OS you are testing with

LordMZTE:
Sorry, I should've clarified that I can reproduce this on master. I'm just very confused why my changes would make the code compile, so my guess is that my branch might be based off of an old revision of the master branch where this was broken for some reason.

PR #11979, opened 2022-07-01
Return invalid argument in sendto instead of unreachable

marijnfs:
At least on Linux, the sendto command can cause the INVAL error with 'user-side' error. I.e. not a clear bug in the code, but simply the address being send to not being in the right ip address range. There is no definite way to check for this issue before hand, so it should probably not turn to an unreachable and crash.

The user still has to deal with the enum and could make it unreachable if that is required on their side.

Here is an example to show the issue:

```
const std = @import("std");
const os = std.os;
const net = std.net;

pub fn main() anyerror!void {
    const address_all = net.Address.initIp4([_]u8{ 0, 0, 0, 0 }, 4040);
    const address = net.Address.initIp4([_]u8{ 127, 0, 0, 1 }, 4040);

    const address_target_ok = net.Address.initIp4([_]u8{ 127, 0, 0, 42 }, 4040);
    const address_target_notok = net.Address.initIp4([_]u8{ 192, 168, 1, 1 }, 4040);

    const sock_flags = os.SOCK.DGRAM | os.SOCK.CLOEXEC | os.SOCK.NONBLOCK;

    const buf = "hello";
    {
        var fd_all = try os.socket(address_all.any.family, sock_flags, os.IPPROTO.UDP);
        defer os.closeSocket(fd_all);

        try os.bind(fd_all, &address_all.any, address_all.getOsSockLen());

        _ = try os.sendto(fd_all, buf, os.MSG.NOSIGNAL, &address_target_ok.any, address_target_ok.getOsSockLen());
        _ = try os.sendto(fd_all, buf, os.MSG.NOSIGNAL, &address_target_notok.any, address_target_notok.getOsSockLen());
    }

    {
        var fd = try os.socket(address.any.family, sock_flags, os.IPPROTO.UDP);
        defer os.closeSocket(fd);

        try os.bind(fd, &address.any, address.getOsSockLen());
        _ = try os.sendto(fd, buf, os.MSG.NOSIGNAL, &address_target_ok.any, address_target_ok.getOsSockLen());
        _ = try os.sendto(fd, buf, os.MSG.NOSIGNAL, &address_target_notok.any, address_target_notok.getOsSockLen()); // This last one get .INVAL
    }
}
```

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -5781,7 +5781,10 @@ pub fn sendmsg(
     }
 }
 
-pub const SendToError = SendMsgError;
+pub const SendToError = SendMsgError || error{
+    /// The destination address is not reachable by the bound address.
+    UnreachableAddress,
+};
 
 /// Transmit a message to another socket.
 ///
@@ -5858,7 +5861,7 @@ pub fn sendto(
                 .DESTADDRREQ => unreachable, // The socket is not connection-mode, and no peer address is set.
                 .FAULT => unreachable, // An invalid user space address was specified for an argument.
                 .INTR => continue,
-                .INVAL => unreachable, // Invalid argument passed.
+                .INVAL => return error.UnreachableAddress,
                 .ISCONN => unreachable, // connection-mode socket was connected already but a recipient was specified
                 .MSGSIZE => return error.MessageTooBig,
                 .NOBUFS => return error.SystemResources,


lib/std/os.zig
```diff
@@ -5684,6 +5684,11 @@ pub const SendMsgError = SendError || error{
     /// The socket is not connected (connection-oriented sockets only).
     SocketNotConnected,
     AddressNotAvailable,
+
+    /// Returned when the arguments to the function are invalid
+    /// This typically signifies a bug in the code, but on linux it is also returned when sendTo is used with an address
+    /// that is not reachable by the bound address.
+    InvalidArgument,
```
marler8997, line 5691: I believe the convention here would be to add an error like `UnreachableAddress` if the situation is as you describe.  There are a few other cases like this where some error codes are re-used, and if the only other cases are a result of code bugs, the best we can do is assume the code is written correctly and assume it's the non-buggy error.

Aransentin:
Since `os.send` calls `os.sendto` you need `os.send` to uncreachable-catch `error.UnreachableAddress` as well. Right now that function doesn't work:

```
/opt/zig/lib/std/os.zig:5921:28: error: expected type 'std.os.SendError', found 'error{AccessDenied,WouldBlock,FastOpenAlreadyInProgress,ConnectionResetByPeer,MessageTooBig,SystemResources,BrokenPipe,FileDescriptorNotASocket,NetworkSubsystemFailed,Unexpected,UnreachableAddress,}'
        else => |e| return e,
                           ^
/opt/zig/lib/std/os.zig:5786:5: note: 'error.UnreachableAddress' not a member of destination error set
    UnreachableAddress,
    ^```

PR #11982, opened 2022-07-02
ignore SIGPIPE by default

marler8997:
This PR modifies `start.zig` to tell the kernel to ignore SIGPIPE before
calling main.  This can be disabled by adding this to the root module:
```zig
pub const keep_sigpipe = true;
```

`SIGPIPE` is triggered when a process attempts to write to a broken pipe.
By default, SIGPIPE will terminate the process without giving the program
an opportunity to handle the situation.  Unlike a segfault, it doesn't
trigger the panic handler so all the developer sees is that the program
terminated with no indication as to why.

By telling the kernel to instead ignore SIGPIPE, writes to broken pipes
will return the EPIPE error (error.BrokenPipe) and the program can handle
them like any other error.

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -7057,42 +7057,12 @@ pub fn timerfd_gettime(fd: i32) TimerFdGetError!linux.itimerspec {
     };
 }
 
-/// Whether or not the current target support SIGPIPE
-pub const have_sigpipe_support = switch (builtin.os.tag) {
-    .linux,
-    .macos,
-    .netbsd,
-    .solaris,
-    .freebsd,
-    .openbsd,
-    => true,
-    else => false,
-};
-
-pub const keep_sigpipe: bool = if (@hasDecl(root, "keep_sigpipe"))
-    root.keep_sigpipe
-else
-    false;
+pub const have_sigpipe_support = @hasDecl(@This(), "SIG") and @hasDecl(SIG, "PIPE");
 
 fn noopSigHandler(_: c_int) callconv(.C) void {}
 
-/// This function will tell the kernel to ignore SIGPIPE rather than terminate
-/// the process.  This function is automatically called in `start.zig` before
-/// `main`.  This behavior can be disabled by adding this to your root module:
-///
-///     pub const keep_sigpipe = true;
-///
-/// SIGPIPE is triggered when a process attempts to write to a broken pipe.
-/// By default, SIGPIPE will terminate the process without giving the program
-/// an opportunity to handle the situation.  Unlike a segfault, it doesn't
-/// trigger the panic handler so all the developer sees is that the program
-/// terminated with no indication as to why.
-///
-/// By telling the kernel to instead ignore SIGPIPE, writes to broken pipes
-/// will return the EPIPE error (error.BrokenPipe) and the program can handle
-/// it like any other error.
 pub fn maybeIgnoreSigpipe() void {
-    if (have_sigpipe_support and !keep_sigpipe) {
+    if (have_sigpipe_support and !std.options.keep_sigpipe) {
         const act = Sigaction{
             // We set handler to a noop function instead of SIG.IGN so we don't leak our
             // signal disposition to a child process
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -7056,3 +7056,48 @@ pub fn timerfd_gettime(fd: i32) TimerFdGetError!linux.itimerspec {
         else => |err| return unexpectedErrno(err),
     };
 }
+
+/// Whether or not the current target support SIGPIPE
+pub const have_sigpipe_support = switch (builtin.os.tag) {
+    .linux,
+    .macos,
+    .netbsd,
+    .solaris,
+    .freebsd,
+    .openbsd,
+    => true,
+    else => false,
+};
+
+pub const keep_sigpipe: bool = if (@hasDecl(root, "keep_sigpipe"))
+    root.keep_sigpipe
+else
+    false;
+
+/// This function will tell the kernel to ignore SIGPIPE rather than terminate
+/// the process.  This function is automatically called in `start.zig` before
+/// `main`.  This behavior can be disabled by adding this to your root module:
+///
+///     pub const keep_sigpipe = true;
+///
+/// SIGPIPE is triggered when a process attempts to write to a broken pipe.
+/// By default, SIGPIPE will terminate the process without giving the program
+/// an opportunity to handle the situation.  Unlike a segfault, it doesn't
+/// trigger the panic handler so all the developer sees is that the program
+/// terminated with no indication as to why.
+///
+/// By telling the kernel to instead ignore SIGPIPE, writes to broken pipes
+/// will return the EPIPE error (error.BrokenPipe) and the program can handle
+/// it like any other error.
+pub fn maybeIgnoreSigpipe() void {
+    if (have_sigpipe_support and !keep_sigpipe) {
+        const act = Sigaction{
+            .handler = .{ .sigaction = SIG.IGN },
+            .mask = empty_sigset,
+            .flags = SA.SIGINFO,
+        };
+        sigaction(SIG.PIPE, &act, null) catch |err| std.debug.panic("ignore SIGPIPE failed with '{s}'" ++
+            ", add `pub const keep_sigpipe = true;` to your root module" ++
+            " or adjust have_sigpipe_support in std/os.zig", .{@errorName(err)});
+    }
+}
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -7074,6 +7074,8 @@ pub const keep_sigpipe: bool = if (@hasDecl(root, "keep_sigpipe"))
 else
     false;
 
+fn noopSigHandler(_: c_int) callconv(.C) void {}
+
 /// This function will tell the kernel to ignore SIGPIPE rather than terminate
 /// the process.  This function is automatically called in `start.zig` before
 /// `main`.  This behavior can be disabled by adding this to your root module:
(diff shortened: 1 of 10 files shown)


motiejus:
I am not in favor as-is. Default sigpipe handler is convenient for almost all CLIs I've written: terminates the program automatically when, for example, it is piped and the listener cuts it off with ` | head` or so.

Looking at the IRC chat, this was a nontrivial problem for you, I presume, not because of stderr/stdout handling. If so, what do you think about a default handler like Go's?

https://pkg.go.dev/os/signal#hdr-SIGPIPE

marler8997:
> Default sigpipe handler is convenient for almost all CLIs I've written

Does this mean all the CLI's you've written in Zig don't check whether calls to `write` fail? 

Unless you're calling `std.os.linux.write` directly and you ignore the return value, you're application already has a convenient way to tell when a pipe is broken, `write` fails with `EPIPE`.  Whether that error should cause the program to exit, panic or trigger an error message seems like a good choice to expose to the program by default.  Having the default be "terminate the program" without letting the application know this is the default behavior hides a detail that in my opinion would be better of as an explicit choice rather than an implicit detail.

marler8997:
Another point I thought of after talking on IRC.  Depending on `SIGPIPE` to terminate your program is problematic because default signal handlers are inherited from the parent.  This means that if your program depends on `SIGPIPE` terminating your program, then you should be explicitly setting the handler to `SIG_TERM`, otherwise, your program could be called with it set to `SIG_IGN` causing it to behave differently.

andrewrk:
As it stands I am in favor of this but I will leave this open for a while for discussion.

This is also extremely relevant for networking, where using `write` to write to a socket whose reading end is closed will cause a SIGPIPE, which is, to put it mildly, an undesirable outcome for the common case of a client disconnecting.

motiejus:
> > Default sigpipe handler is convenient for almost all CLIs I've written
> 
> Does this mean all the CLI's you've written in Zig don't check whether calls to `write` fail?

That's division by zero. I have not done a cli in zig, but I have done many in Go.

There are two cases of write failures:
- write to stdout/stderr. That's program output or diagnostics (usually fds 1 and 2).
- everything else: write to a file, to a network socket.

> Unless you're calling `std.os.linux.write` directly and you ignore the return value, you're application already has a convenient way to tell when a pipe is broken, `write` fails with `EPIPE`. Whether that error should cause the program to exit, panic or trigger an error message seems like a good choice to expose to the program by default. Having the default be "terminate the program" without letting the application know this is the default behavior hides a detail that in my opinion would be better of as an explicit choice rather than an implicit detail.

I agree to all of the above, except I think the default behavior makes sense for file descriptors 1 and 2. See https://pkg.go.dev/os/signal#hdr-SIGPIPE

motiejus:
Also, Zig community decides to attach their own default handlers, either via this patch as-is or via my suggestion, we should make sure they can be configured per thread (a la pthread_sigmask).

matu3ba:
> I agree to all of the above, except I still believe the default behavior makes sense for file descriptors 1 and 2.

You can not restrict this to descriptors (0) 1 and 2, because any program that needs dedicated pipes will not work.
Think of the extra pipe as control channel and the standard streams as data channels or vice versa.

The main use case of this is, if you want to properly test IPC behavior of a program and this is also why #11701 exists (in that case motivation is testing panic behavior via custom panic handler).

> can be configured per thread (a la pthread_sigmask)

I dont understand this. Does `pthread_sigmask` not work or are there things missing?

ifreund:
If we do this by default, it would be probably be better to go with the approach we landed on for river here: https://github.com/riverwm/river/pull/664#discussion_r939818546.

It has the advantage that no further code is required to avoid leaking this disposition to child processes created with fork()/execve(). To quote the `signal(7)` man page:
> A child created via fork(2) inherits a copy of its parent's signal
       dispositions.  During an execve(2), the dispositions of handled signals
       are reset to the default; the dispositions of ignored signals are left
       unchanged.

marler8997:
All programs will fall into one of these 3 categories:

1. Programs that function incorrectly when SIGPIPE is enabled
2. Programs that function incorrectly when SIGPIPE is disabled
3. Programs that function correctly regardless of SIGPIPE

> "SIGPIPE enabled" means its set to terminate the process

Because SIGPIPE can be set either way when a process starts, in order to be "correct", any program that falls into the first 2 categories MUST explicitly set the disposition of SIGPIPE.  I think most programs fall into the first category .  Without this PR, all those programs are incorrect unless they explicitly set SIGPIPE in their startup code.

Ifruend points out that this PR not only affects the current process but also child processes.  This is a concern but any issues arising from this would only be bug exposures rather than breaking behavior.  Any program that previously depended on SIGPIPE's disposition but was not setting it would have a bug and any program that did not depend on it would still work with the new default disposition.

matu3ba:
> I think the best thing we can do here is keep things predictable for people familiar with socket/pipe programming on unix like systems and just stick to the default.

+ At least clearly document the behavior on all supported wrappers.

Linux defaults to filter programs https://stackoverflow.com/a/9337925.
There is also this super long discussion on Rust https://github.com/rust-lang/rust/issues/62569.

One alternative would be to support wrappers to change default behavior in portable manner, say for signal safety and stuff like this.

ifreund:
> Any program that previously depended on SIGPIPE's disposition but was not setting it would have a bug

I think this category includes pretty much every CLI application and I don't agree that failing to explicitly set SIGPIPE deposition on startup is a bug. It is documented as the default behavior and seems reasonable to rely upon. Looking at several implementation of core unix utilities for BSD and Linux I don't see programs doing this, so I think it's a moot point whether it's technically a "bug" or not. If we leak signal disposition to arbitrary child process we will break things.

It is *always* bad practice to leak signal dispositions to child processes just as it is always bad practice to leak fds or other things.

andrewrk (approved):
Alright I've thought about this enough and I'm ready to make the call. I think @marler8997's breakdown is spot on here:

https://github.com/ziglang/zig/pull/11982#issuecomment-1211928193

Responding to @ifreund 

> I think this category includes pretty much every CLI application and I don't agree that failing to explicitly set SIGPIPE deposition on startup is a bug. It is documented as the default behavior and seems reasonable to rely upon.

Since the SIGPIPE disposition is inherited from the parent process, there is no such thing as the default behavior. A process cannot assume SIGPIPE disposition is in any particular state. I disagree with your conclusion that it is reasonable to rely on, although it is certainly common in practice.

From the caller's perspective, to be fully robust, it must know for every process being executed, what its required SIGPIPE disposition is, if any. The safest thing to do when spawning a process which may include a (arguably buggy) CLI application in its tree would be to explicitly restore SIGPIPE in case the parent process had it ignored.

I'm not satisfied with leaving such a brittle system alone when we have the chance to take advantage of zig's better error handling semantics and reduce the amount of buggy, inefficient signal handling code in the wild.

Furthermore, taking such a stand will allow Zig's standard library abstractions to be more future-proof against better designed operating systems than POSIX.

I do think it would be worth adding to the std.ChildProcess API, a way to quickly decide between passing SIGPIPE through unchanged, changing it to be ignored, or changing it to be "default" (enabled). I think this additional API is not ultimately a prerequisite for merging this PR because of the point I make below.

Finally, I think this piece of information may be incorrect:

> Edit: setting an empty signal handler is of course not the same as using SIG_IGN and will not cause EPIPE to be returned in the case where SIGPIPE would have been sent.

Let us verify empirically whether this is true, because I suspect that another option available to us is to set the signal handler in start.zig to an empty handler. This would have no effect on child processes, even those spawned by third party code, while providing the zig compilation unit the desired error handling properties. This would be a completely safe default. Applications which have no need for such carefulness could opt in via the global root options thingie to a full `SIG_IGN` as well as opting out.

@marler8997 - if you wouldn't mind rebasing this on master branch, updating to the new std root options thing (#14181), and testing the behavior of empty handlers, then I'll land this.

If you're busy - which would make a lot of sense since I'm only finally reviewing this now - feel free to hand this off to me for completion.

lib/std/os.zig
```diff
@@ -7056,3 +7056,52 @@ pub fn timerfd_gettime(fd: i32) TimerFdGetError!linux.itimerspec {
...
+        const act = Sigaction{
+            // We set handler to a noop function instead of SIG.IGN so that the child process
+            // does not inherit our signal disposition.
+            .handler = .{ .handler = noopSigHandler },
+            .mask = empty_sigset,
+            .flags = 0,
+        };
+        sigaction(SIG.PIPE, &act, null) catch |err| std.debug.panic("ignore SIGPIPE failed with '{s}'" ++
+            ", add `pub const keep_sigpipe = true;` to your root module" ++
+            " or adjust have_sigpipe_support in std/os.zig", .{@errorName(err)});
```
andrewrk, line 7105: This message reads as if it were directed towards a developer, but it's user-facing in release builds.

I suggest to either make the message more succinct so that it covers both use cases simultaneously, or choose an end-user-appropriate message for release builds and a developer-appropriate message for debug builds.

marler8997: Makes sense.   I think I based this error message off the one for when installing the default segfault handler fails, so we might want to take another look at that one.  I've updated the error with something more appropriate for an end user:
```
            std.debug.panic("failed to install noop SIGPIPE handler with '{s}'", .{@errorName(err)});
```

I think using "panic" for an initial implementation is nice, if we find use cases where it makes sense to ignore the error we can always do that later after we've identified those use cases.

marler8997:
> I didn't test this, I could very well be wrong there and I hope I am. As you say, that would give us the best of both worlds if the EPIPE error is infact returned. I think we should definitely do a proper empirical test before merging this PR.

I'm happy to report that it looks like setting a "no-op" signal handler seems to work.  The new sigpipe standalone test I added confirms this and I confirmed with some testing that this mechanism does not leak any signal disposition to the child process.

lib/std/os.zig
```diff
@@ -7056,3 +7056,51 @@ pub fn timerfd_gettime(fd: i32) TimerFdGetError!linux.itimerspec {
...
+    .macos,
+    .netbsd,
+    .solaris,
+    .freebsd,
+    .openbsd,
+    => true,
+    else => false,
+};
+
+pub const keep_sigpipe: bool = if (@hasDecl(root, "keep_sigpipe"))
```
ifreund, line 7072: This should use the new std.options mechanism: https://github.com/ziglang/zig/blob/c32171991b25b323cd68ff96c294bf5a6fa753b8/lib/std/std.zig#L106

lib/std/os.zig
```diff
@@ -7056,3 +7056,51 @@ pub fn timerfd_gettime(fd: i32) TimerFdGetError!linux.itimerspec {
         else => |err| return unexpectedErrno(err),
     };
 }
+
+/// Whether or not the current target support SIGPIPE
+pub const have_sigpipe_support = switch (builtin.os.tag) {
```
ifreund, line 7061: I think it would be worthwhile to make this an exhaustive switch so that people get compiler errors here when adding a new target OS and can't forget to add it to this list as easily.

marler8997: Whoa there are alot of os tags.  I've attempted to use the existence of `SIG.PIPE` instead to see if that would work.

andrewrk:
Wow, really nice work on that new test case.

PR #11990, opened 2022-07-04
std.testing.expectEqualSlices: use expectEqualStrings for strings

voroskoi:
use expectEqualStrings when input type is slice of []const u8 values
add couple of tests


topolarity:
This is problematic: `[]const u8` is not necessarily a UTF-8 encoded string

ifreund (commented):
We can't merge this as it would leave no way to properly test comparison of `[]const u8` slices that are not ascii/UTF-8 encoded text. This is the reason that `expectEqualStrings()` is a separate function.

PR #11991, opened 2022-07-04
trivial: use a named struct in Compilation.create()

motiejus:
My dev tools are not as sophisticated to automatically resolve the
struct in question, but having a named one will certainly aid
visibility and `git grep`.

Background: it looks like a nuisance; however, I have added a number of linker flags to `src/main.zig` and friends already, and this one is always confusing to find the struct for.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -2757,7 +2757,7 @@ fn buildOutputType(
 
     gimmeMoreOfThoseSweetSweetFileDescriptors();
 
-    const comp = Compilation.create(gpa, .{
+    const comp = Compilation.create(gpa, Compilation.InitOptions{
         .zig_lib_directory = zig_lib_directory,
         .local_cache_directory = local_cache_directory,
         .global_cache_directory = global_cache_directory,


kubkon:
While a trivial change, I don't really see the point in it. By your reasoning, why can't you just grep for `Compilation.create`?  The very first time you ever want to tweak this bit of the codebase it is natural to feel lost, but once you get a feel for it, I think knowing where to look is not super problematic. Tl;dr I am confused by this change as I have never had a problem you have experienced.

motiejus:
> While a trivial change, I don't really see the point in it. By your reasoning, why can't you just grep for Compilation.create

Then you find a type and it requires a jump to the type. :) An extra hop, a nuisance.

Until today I assumed that, when a struct is constructed "anonymously", compiler will not check that all fields have been set. It is false: semantically those are equivalent.

Since they are indeed equivalent, this leaves the issue "I would write it differently", thus we all know how such arguments end. Thanks for your input. :-)

Issue #11450, opened 2022-04-16
stage3 fails the behavior tests


andrewrk:
Zig Version: `0.10.0-dev.1855+a315d51c0`

After stage2 rebuilds itself, producing stage3, it no longer passes the behavior tests. In this snippet below you can see:
 * building stage1 successfully
 * building stage2 successfully
 * building stage3 successfully
 * stage2 passing behavior tests with the llvm backend
 * stage3 failing behavior tests with the llvm backend

```
[nix-shell:~/dev/zig/build-release]$ ninja
[nix-shell:~/dev/zig/build-release]$ ./zig build -p stage2 -Dskip-install-lib-files  --prominent-compile-errors  -Denable-llvm
[nix-shell:~/dev/zig/build-release]$ ./stage2/bin/zig build -p stage3 -Dskip-install-lib-files -Denable-llvm
[nix-shell:~/dev/zig/build-release]$ ./stage2/bin/zig test ../test/behavior.zig -I../test
Test [20/1244] test.generic function with align param... SKIP
Test [21/1244] test.runtime known array index has best alignment possible... SKIP
Test [24/1244] test.read 128-bit field from default aligned struct in global memory... SKIP
Test [26/1244] test.align(@alignOf(T)) T does not force resolution of T... SKIP
Test [27/1244] test.align(N) on functions... SKIP
Test [60/1244] test.simple coroutine suspend and resume... SKIP
Test [61/1244] test.pass parameter to coroutine... SKIP
Test [62/1244] test.suspend at end of function... SKIP
Test [63/1244] test.local variable in async function... SKIP
Test [64/1244] test.calling an inferred async function... SKIP
Test [65/1244] test.@frameSize... SKIP
Test [66/1244] test.coroutine suspend, resume... SKIP
Test [67/1244] test.coroutine suspend with block... SKIP
Test [68/1244] test.coroutine await... SKIP
Test [69/1244] test.coroutine await early return... SKIP
Test [70/1244] test.async function with dot syntax... SKIP
Test [71/1244] test.async fn pointer in a struct field... SKIP
Test [72/1244] test.@asyncCall with return type... SKIP
Test [73/1244] test.async fn with inferred error set... SKIP
Test [74/1244] test.error return trace across suspend points - early return... SKIP
Test [75/1244] test.error return trace across suspend points - async return... SKIP
Test [76/1244] test.break from suspend... SKIP
Test [77/1244] test.heap allocated async function frame... SKIP
[...]

PR #12001, opened 2022-07-05
Upgrade to LLVM 14

### LINKED ISSUE #11274: Upgrade to LLVM 14
Relevant LLVM doc: https://llvm.org/docs/OpaquePointers.html

> Traditionally, LLVM IR pointer types have contained a pointee type. For example, i32* is a pointer that points to an i32 somewhere in memory. However, due to a lack of pointee type semantics and various issues with having pointee types, there is a desire to remove pointee types from pointers.
> The opaque pointer type project aims to replace all pointer types containing pointee types in LLVM with an opaque pointer type. The new pointer type is tentatively represented textually as ptr.

This lead to the deprecation of a number of C API function that now required to be given both the pointer LLVMValueRef and the pointee LLVMTypeRef:

> LLVMBuildLoad -> LLVMBuildLoad2
LLVMBuildCall -> LLVMBuildCall2
LLVMBuildInvoke -> LLVMBuildInvoke2
LLVMBuildGEP -> LLVMBuildGEP2
LLVMBuildInBoundsGEP -> LLVMBuildInBoundsGEP2
LLVMBuildStructGEP -> LLVMBuildStructGEP2
LLVMBuildPtrDiff -> LLVMBuildPtrDiff2
LLVMConstGEP -> LLVMConstGEP2
LLVMConstInBoundsGEP -> LLVMConstInBoundsGEP2
LLVMAddAlias -> LLVMAddAlias2
Additionally, it will no longer be possible to call LLVMGetElementType() on a pointer type.

`codegen.cpp` make extensive use of all the above functions so a number of changes need to happen there.
Trying to compile Zig with LLVM14 currentlly yield 200 warnings and 20 errors which look all related to the above changes.

I've started working on some changes, but I'm not really familiar with `codegen.cpp` and I'm afraid of making errors and passing the wrong types. But hopefully doing the most of the refactoring myself will allow @andrewrk to focus on the semantic.

andrewrk:
Does not contain a workaround for https://github.com/llvm/llvm-project/issues/55522, which means that using `@Vector` with `bool` will be regressed after merging this, until the upstream bug is fixed, probably in LLVM 15.0.0. The good news however is that the way stage2 lowers code in the LLVM backend makes the bug less likely to be triggered. In the future we will do better at staying on top of LLVM release candidates in order to report such issues *and file them as release blockers*.

~Other than this, things seem to be working fine.~ I noticed that even in master branch, the zig test suite does not run assertion-clean against a debug build of LLVM 13. My next priority will be to remedy this, and then follow it up with running the full test suite again, this time with LLVM 14 debug mode, to check for assertions tripped.

Closes #11274


andrewrk:
> CI: update tarballs to LLVM 14

I tried updating them to stage3 (previously stage1) which may have been a bit ambitious since that does mean that stage3 has to correctly handle compiler_rt, which is causing all the failures. It is strange though, because stage3 built normally (not with zig-bootstrap) does not exhibit this problem. Oh well, this is already filed as #11450 and I better just revert back to using stage1 for the CI tarballs, focusing on one problem at a time.

andrewrk:
You know what? No. stage3 or bust. I'm not going to revert zig-bootstrap to stage1. I will push forward on #89 and #11450 instead.

andrewrk:
> Other than this, things seem to be working fine.

Spoke too soon.

 * #12012
 * #12013

Jarred-Sumner:
> Does not contain a workaround for https://github.com/llvm/llvm-project/issues/55522, which means that using @Vector with bool will be regressed after merging this, until the upstream bug is fixed, probably in LLVM 15.0.0.

That issue is 😆

andrewrk:
Here is one performance data point, building [Tetris](https://github.com/andrewrk/tetris) from source:

```
[nix-shell:~/dev/tetris]$ hyperfine '/home/andy/Downloads/zig/build-release/stage1/bin/zig build -fno-stage1' '/home/andy/dev/zig/build-release/stage1/bin/zig build -fno-stage1' '/home/andy/Downloads/zig/build-release/stage1/bin/zig build' '/home/andy/dev/zig/build-release/stage1/bin/zig build' --warmup 2 -p 'rm -rf zig-cache zig-out'
Benchmark 1: /home/andy/Downloads/zig/build-release/stage1/bin/zig build -fno-stage1
  Time (mean ± σ):      2.758 s ±  0.032 s    [User: 2.829 s, System: 0.238 s]
  Range (min … max):    2.718 s …  2.797 s    10 runs
 
Benchmark 2: /home/andy/dev/zig/build-release/stage1/bin/zig build -fno-stage1
  Time (mean ± σ):      5.590 s ±  0.062 s    [User: 5.698 s, System: 0.250 s]
  Range (min … max):    5.494 s …  5.669 s    10 runs
 
Benchmark 3: /home/andy/Downloads/zig/build-release/stage1/bin/zig build
  Time (mean ± σ):      2.804 s ±  0.019 s    [User: 2.725 s, System: 0.386 s]
  Range (min … max):    2.770 s …  2.836 s    10 runs
 
Benchmark 4: /home/andy/dev/zig/build-release/stage1/bin/zig build
  Time (mean ± σ):      4.623 s ±  0.041 s    [User: 4.586 s, System: 0.392 s]
  Range (min … max):    4.542 s …  4.689 s    10 runs
 
Summary
  'LLVM13 stage2' ran
    1.02 ± 0.01 times faster than 'LLVM13 stage1'
    1.68 ± 0.02 times faster than 'LLVM14 stage1'
    2.03 ± 0.03 times faster than 'LLVM14 stage2'

```

Issue #1759, opened 2018-11-19
Official FreeBSD Support


andrewrk:
Here's the issue to track FreeBSD support.

History:

 * @tiehuis started an initial `freebsd` branch.
 * @myfreeweb picked up the figurative baton in #1661 and that work is merged into `freebsd2` branch.
   * See this PR description and comments for some useful information.
   * I believe the current state of this branch is that the tests pass in debug mode but not release mode. I haven't personally tried it yet.

Checklist:

 * [ ] #1217  Bus error on simple code
 * [ ] get all the tests passing in the `freebsd2` branch.
 * [ ] (@andrewrk) audit the `freebsd2` branch (I haven't reviewed the code yet)
 * [ ] CI for FreeBSD. We actually have a few options:
    * get my raspberry pi (aarch64) hooked up to the CI somehow
    * use https://builds.sr.ht/ which has freebsd x86_64 support
    * on azure dev ops, use qemu and a custom freebsd image as demonstrated by @andrewchambers in #941 (although this was for openbsd not freebsd). This should be possible because we get a full 6 hours to run the tests with azure dev ops.

If you want to contribute to this, make pull requests against the `freebsd2` branch.

Once this is complete, we can add FreeBSD to the support table and declare that we officially support it.

bnoordhuis:
> get my raspberry pi (aarch64) hooked up to the CI somehow

That's a raspberry pi 3? We have a small armada of those in the Node.js CI. They're a lot faster than the pi 1 and 2 but still so much slower than regular server hardware that we cross-compile and shard the test suite across 10 or 12 machines.

That's a long-winded way of saying they're not really an option unless you're _really_ patient. :-)

andrewrk:
> That's a raspberry pi 3?

Yes it is. Thanks for the heads up!

myfreeweb:
re: CI, I'm "starting" a buildbot cluster at https://porting.club — currently it's just a Scaleway 4-core ThunderX VPS (the master as well as the first worker, which is a jail on the same box).
The "plan" is to have a cluster with many different OSes and CPU architectures specifically for stuff that has portability needs — compilers/interpreters/VMs, libraries that touch OS specific stuff, etc.
The idea is to not use throwaway containers or VMs, but instead set up persistent environments — so that they can be used not just for CI tests or nightly builds, but also for development (i.e. project developers should be able to ssh into the workers).

(btw ThunderX cores are comparable to Cortex-A72s / significantly faster than the A53s the RPi has)

[...]

Issue #3939, opened 2019-12-18
Tier 1 FreeBSD Support for ARM 64-bit (aarch64)


andrewrk:
Related issue:
 *  Tier 1 FreeBSD Support for x86_64 (#1759)

Checklist:
 * [ ] solve cross compilation when targeting FreeBSD (ability to build the libc start files from source)
 * [ ] get all the tests passing for aarch64 freebsd. Related:  #3210
 * [ ] solve CI test coverage for aarch64 freebsd. Maybe qemu usermode emulation works on freebsd?

PR #12002, opened 2022-07-05
Audit FreeBSD structs to match header files/ABI

The-King-of-Toasters:
Required for #1759.

I wrote a simple project that verified the sizes against the headers at compile time:
```zig
const std = @import("std");
// Vendored from std/c/freebsd.zig.
// Change the first line to `@import("std");` to compile
const freebsd = @import("freebsd.zig");
const c = @cImport({
    @cInclude("sys/types.h");
    @cInclude("sys/socket.h");
    @cInclude("sys/event.h");
    @cInclude("sys/umtx.h");
    @cInclude("sys/link_elf.h");
    @cInclude("sys/fcntl.h");
    @cInclude("sys/stat.h");
    @cInclude("sys/timespec.h");
    @cInclude("sys/dirent.h");
    @cInclude("sys/socket.h");
    @cInclude("sys/signal.h");
    @cInclude("sys/ttycom.h");
    @cInclude("sys/ucontext.h");
    @cInclude("sys/stack.h");
    @cInclude("sys/poll.h");

    @cInclude("netinet/in.h");
    @cInclude("sys/un.h");
    @cInclude("netinet/if_ether.h");
    @cInclude("netdb.h");

    @cInclude("pthread.h");
    @cInclude("semaphore.h");
});

fn assertType(comptime ZigType: anytype, comptime CType: anytype) void {
    std.debug.assert(@sizeOf(ZigType) == @sizeOf(CType));
    std.debug.assert(@bitSizeOf(ZigType) == @bitSizeOf(CType));
    std.debug.assert(@alignOf(ZigType) == @alignOf(CType));
}

fn assert(comptime T: []const u8) void {
    return assertType(@field(freebsd, T), @field(c, T));
}

comptime {
    assert("sf_hdtr");
    assert("pthread_mutex_t");
    assert("pthread_cond_t");
    assert("pthread_rwlock_t");
    assert("pthread_attr_t");
    assert("sem_t");
    assert("_umtx_time");
    assertType(freebsd.Kevent, c.struct_kevent);
    assert("dl_phdr_info");
    assertType(freebsd.Flock, c.struct_flock);
    assert("msghdr");
    assertType(freebsd.Stat, c.struct_stat);
    assert("timespec");
    assert("timeval");
    assert("dirent");
    assert("sockaddr");
    assertType(freebsd.sockaddr.in, c.sockaddr_in);
    assertType(freebsd.sockaddr.in6, c.sockaddr_in6);
    assertType(freebsd.sockaddr.un, c.sockaddr_un);
    assert("sigval");
    assert("sigset_t");
    assert("winsize");
    assertType(freebsd.Sigaction, c.struct_sigaction);
    assert("siginfo_t");
    assert("mcontext_t");
    assert("ucontext_t");
    assert("stack_t");
    assert("addrinfo");
    assert("pollfd");
    //     @compileError(std.fmt.comptimePrint("{d} {d}\n", .{
    //         @sizeOf(freebsd.mcontext_t),
    //         @sizeOf(c.mcontext_t),
    //     }));
}
```
This was ran on a FreeBSD 13.1-RELEASE amd64 system using zig version `0.10.0-dev.2842+d65e248ed`.

I added some doc-comments from the headers where I could. Also included is the `mcontext_t` definition for `aarch64` systems when that gets supported down the line.

### DIFF
--- a/lib/std/c/freebsd.zig
+++ b/lib/std/c/freebsd.zig
@@ -54,8 +54,7 @@ pub const pthread_rwlock_t = extern struct {
 };
 
 pub const pthread_attr_t = extern struct {
-    __size: [56]u8,
-    __align: c_long,
+    inner: ?*anyopaque = null,
 };
 
 pub const sem_t = extern struct {
@@ -181,6 +180,7 @@ pub const AI = struct {
 pub const blksize_t = i32;
 pub const blkcnt_t = i64;
 pub const clockid_t = i32;
+pub const fflags_t = u32;
 pub const fsblkcnt_t = u64;
 pub const fsfilcnt_t = u64;
 pub const nlink_t = u64;
@@ -201,13 +201,20 @@ pub const suseconds_t = c_long;
 
 /// Renamed from `kevent` to `Kevent` to avoid conflict with function name.
 pub const Kevent = extern struct {
+    /// Identifier for this event.
     ident: usize,
+    /// Filter for event.
     filter: i16,
+    /// Action flags for kqueue.
     flags: u16,
+    /// Filter flag value.
     fflags: u32,
+    /// Filter data value.
     data: i64,
+    /// Opaque user data identifier.
     udata: usize,
-    // TODO ext
+    /// Future extensions.
+    _ext: [4]u64 = [_]u64{0} ** 4,
 };
 
 // Modes and flags for dlopen()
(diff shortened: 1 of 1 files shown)


The-King-of-Toasters:
I had to change the event loop code to deal with a change in `Kevent.udata`'s type. Would appreciate the feedback from the owner.

lib/std/c/freebsd.zig
```diff
@@ -246,72 +263,73 @@ pub const Flock = extern struct {
...
-    /// size of address
+    /// Size of address.
     msg_namelen: socklen_t,
-
-    /// scatter/gather array
-    msg_iov: [*]iovec_const,
-
-    /// # elements in msg_iov
+    /// Scatter/gather array.
+    msg_iov: [*]iovec,
```
ifreund, line 288: Pretty sure this should remain `iovec_const` as this is `msghdr_const`

lib/std/c/freebsd.zig
```diff
@@ -52,8 +52,7 @@ pub const pthread_rwlock_t = extern struct {
 };
 
 pub const pthread_attr_t = extern struct {
-    __size: [56]u8,
-    __align: c_long,
+    inner: ?*anyopaque = null,
```
ifreund, line 55: Why is this given a default value? Also wouldn't `pub const pthread_attr_t = ?*anyopaque;` be a more direct translation?

lib/std/c/freebsd.zig
```diff
@@ -199,13 +199,20 @@ pub const suseconds_t = c_long;
...
     flags: u16,
+    /// Filter flag value.
     fflags: u32,
+    /// Filter data value.
     data: i64,
+    /// Opaque user data identifier.
     udata: usize,
-    // TODO ext
+    /// Future extensions.
+    _ext: [4]u64 = undefined,
```
ifreund, line 215: I don't think `undefined` is correct here. The `EV_SET` macro 0-initializes this padding: https://github.com/freebsd/freebsd-src/blob/9bb8a4091c4f63dacda5108f4f994f7f6b35bbf1/sys/sys/event.h#L77-L80

lib/std/c/freebsd.zig
```diff
@@ -229,10 +236,20 @@ pub const RTLD = struct {
     pub const NOLOAD = 0x02000;
 };
 pub const dl_phdr_info = extern struct {
-    dlpi_addr: usize,
+    /// Module relocation base.
+    dlpi_addr: std.elf.Elf64_Addr,
```
ifreund, line 240: I suspect this isn't correct on 32 bit targets.

ifreund (changes requested):
Nice work, this is a cool way to catch some errors.

Note that it only works for the target run on though, for example FreeBSD's stat structure seems to have extra fields when targeting i386: https://github.com/freebsd/freebsd-src/blob/d80d73493767111b7569e831a93061014c682274/sys/sys/stat.h#L170

I didn't have time to review everything in depth, but here's a start

lib/std/c/freebsd.zig
The-King-of-Toasters: I suppose so, I'm just copying what's above.

The-King-of-Toasters: Similar definitions are used for DragonflyBSD and OpenBSD. Perhaps it would be better to change them all in a followup commit?

ifreund: Sure, that's fine.

Vexu (approved):

The-King-of-Toasters:
Updating this PR to audit the `kinfo_file` structure that was added in #12082. I've also backed out the arm mcontext bits, as I couldn't actually get zig to compile on an aarch64 vm. I'll leave that for #3939.

andrewrk:
Thank you and nice work!

PR #12008, opened 2022-07-05
crypto: fix fetching sec1 public keys

except:
This fixes an issue I was having trying to retrieve the public key


jedisct1:
Duplicate of #12009 :)

Issue #2029, opened 2019-03-03
add a builtin for comptime source information


Sahnvour:
I want to propose a way to query the compiler in order to get information on the context of the compilation, at call site. In particular, this is mostly used to get source filename, line, function name.

Classical uses of this feature are logging, asserts, scoped variables (instrumentation, debugging...) and probably others I can't think of.

In C and C++ this is done through the `__FILE__` and `__LINE__` macros which are expanded by the preprocessor.
In [Rust](https://doc.rust-lang.org/std/macro.file.html), there are `file!` and `line!` macros.
[D](https://github.com/dlang/phobos/blob/master/std/experimental/logger/core.d#L197) also makes use of `__FILE__` and such, though I don't know if it's actually a preprocessor.
And so on.

At the moment, something similar could be achieved by capturing a stack trace and looking at debug info, but this would not be very efficient. It would also be easily disturbed by the optimizations performed.

Zig could provide a `@sourceInfo` builtin function, returning an instance along the lines of
```zig
pub const SourceInfo = struct {
    file: []const u8,
    line: comptime_int,
    column: comptime_int,
    module: type,
    function: type,
    function_name: []const u8,
}
```

Example usage:
```zig
pub fn main() void {
    log(@sourceInfo(), "something something");
}
```
Example output:
```
[21:18:04] in 'main' (main.zig:2) | something something
```

Now, this leaves a usability concern: the other languages cited as examples all provide a way to the _library writer_ to make get this information _from client code_. This means they can provide a `log(msg: string)` function or macro that transparently uses the source information.
In current Zig, the only way is to explicitly call `@sourceInfo()` and pass it to the function expecting it.

One solution is to provide a `@callerSourceInfo` to get the source information from the call site into the callee function. But it doesn't play nice with function pointers or extern functions.

andrewrk:
This was previously rejected in #577, but you have made these new arguments:

> capturing a stack trace and looking at debug info would not be very efficient.

> It would also be easily disturbed by the optimizations performed.

I want to point out that a full stack trace does not need to be captured - in fact only a single `usize` returned by `@returnAddress` would need to be captured. A logging library could emit the addresses, which could be processed later with debug info to replace the addresses with source info, or it could resolve the source information immediately. This solves the problem where the caller would have to call `@sourceInfo()`. (The `log` function would want to use `noinline` #640 )

[...]

PR #12016, opened 2022-07-06
Stage2 compile error improvements

Vexu:


test/cases/compile_errors/non_error_sets_used_in_merge_error_sets_operator.zig
```diff
@@ -0,0 +1,15 @@
...
+    const Errors = error{} || u16;
+    _ = Errors;
+}
+
+// error
+// backend=stage2
+// target=native
+//
+// :2:20: error: expected error set type, found 'u8'
+// :6:31: error: expected error set type, found 'u16'
```
andrewrk, line 15: We seem to have lost the handy note:
```
note: `||` merges error sets; `or` performs boolean OR
```
this is beneficial because `||` in other languages is `or`.

I haven't been checking too closely on your other commits, have you been making sure that useful error notes are preserved when marking these test cases as "passing" for stage2?

src/Zir.zig
```diff
@@ -2919,6 +2919,11 @@ pub const Inst = struct {
         index: u32,
     };
 
+    pub const Panic = struct {
+        force_comptime: u32,
```
andrewrk, line 2923: oof an entire 32 bit integer for 1 bit of information, perhaps could we have `panic_comptime` and `panic` tags instead? I suppose if we're out of ZIR tags then this could be a later enhancement.

andrewrk (changes requested):

test/cases/compile_errors/non_error_sets_used_in_merge_error_sets_operator.zig
Vexu: This note is still there but only when both operands are bools.

src/Zir.zig
Vexu: > perhaps could we have `panic_comptime` and `panic` tags instead?

I'll do that, I don't know why I didn't think of that when making the change.

andrewrk:
> stage2: make line field of `@src` runtime known 

can you explain more why this change?

andrewrk:
I'm going to merge because I trust your judgement and these are some really nice improvements. We can chat about `@src` later :-)

Vexu:
> can you explain more why this change?

It is to stop users from doing comptime logic on the line number and then moving the function during an incremental update, it was discussed in the original proposal https://github.com/ziglang/zig/issues/2029#issuecomment-645793168.
```zig
test {
    if (@src().line == 4) @compileError("don't put me on line 4");
}
```

The alternative is to make all the fields comptime known but then invalidate the function if it is moved.

PR #12017, opened 2022-07-06
AstGen: fix catch payoad checking for shadowing

### LINKED ISSUE #11939: Catch payload does not have variable shadowing check
```zig
var a: anyerror!u32 = 1;
test {
    _ = a catch |a| return a;
    // if (a) {} else |a| return a; // local shadows declaration of 'a'
}
```

ieeemma:
Fixes #11939.

Also added a test case for this to prevent regression.

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -5000,6 +5000,11 @@ fn orelseCatchExpr(
         if (mem.eql(u8, tree.tokenSlice(payload), "_")) {
             return astgen.failTok(payload, "discard of error capture; omit it instead", .{});
         }
+
+        const payload_name = try astgen.identAsString(payload);
+        const payload_str = tree.tokenSlice(payload);
+        try astgen.detectLocalShadowing(scope, payload_name, payload, payload_str);
+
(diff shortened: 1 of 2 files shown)


src/AstGen.zig
```diff
@@ -5000,6 +5000,11 @@ fn orelseCatchExpr(
         if (mem.eql(u8, tree.tokenSlice(payload), "_")) {
             return astgen.failTok(payload, "discard of error capture; omit it instead", .{});
         }
+
+        const payload_name = try astgen.identAsString(payload);
```
Vexu, line 5004: This is the same as `err_name` below, both should use the same variable. The previous `tokenSlice` could also use the same variable as `payload_str`.

Vexu (changes requested):

src/AstGen.zig
ieeemma: Sorry I missed those, I haven't worked on stage2 before so its a lot to learn. Should be sorted now!

Vexu (approved):

PR #12020, opened 2022-07-06
`std.log`: deduplicate `switch`

r00ster91:
Instead of having a distinct type for every string length, we can generalize it using `[]const u8` to return a string of any length.

### DIFF
--- a/lib/std/log.zig
+++ b/lib/std/log.zig
@@ -85,12 +85,7 @@ pub const Level = enum {
     debug,
 
     /// Returns a string literal of the given level in full text form.
-    pub fn asText(comptime self: Level) switch (self) {
-        .err => @TypeOf("error"),
-        .warn => @TypeOf("warning"),
-        .info => @TypeOf("info"),
-        .debug => @TypeOf("debug"),
-    } {
+    pub fn asText(comptime self: Level) []const u8 {
         return switch (self) {
             .err => "error",
             .warn => "warning",


jibal:
How is this an improvement? Wasn't the point to return a string literal rather than a slice?

r00ster91:
> How is this an improvement? Wasn't the point to return a string literal rather than a slice?

The function can still be used at compile time so I don't see the problem. Am I missing something? The improvement is less code by deduplicating and making it easier to read.

jibal:
> This won't work because the return type needs to be used in a string literal concatenation.

Ah yes, thanks ... it bothered me on principle that the return type of this function was being changed from a literal to a non-literal but I didn't take the time to think it through.

I think generally these sorts of changes are ill-advised ... they have no substantive benefit and carry a risk of introducing bugs or problems, and it's not adequate to ask here for failing "use cases"--those may well show up after release. And I say this as someone who is himself overly prone to making cosmetic changes.

leecannon:
@marler8997 ?
```zig
const std = @import("std");

pub fn main() !void {
    std.debug.print("literal " ++ asText(.err) ++ "\n", .{});
}

pub fn asText(comptime self: std.log.Level) []const u8 {
    return switch (self) {
        .err => "error",
        .warn => "warning",
        .info => "info",
        .debug => "debug",
    };
}
```
```
$ zig build run
literal error
```

r00ster91:
@leecannon yes I was testing this just now too. String literal concatenation works just fine for me. I'm quite confused.

marler8997:
Huh, looks like it works now?  Must have been an old issue that no longer exists (see original conversation here: https://github.com/ziglang/zig/pull/9224)

Issue #6252, opened 2020-09-04
general purpose allocator failed to compile under wasm-standalone


codehz:
```
.../zig/lib/zig/std/fs/file.zig:20:15: error: container 'std.os' has no member called 'fd_t'
    handle: os.fd_t,
              ^
.../zig/lib/zig/std/debug.zig:92:31: note: referenced here
        const stderr_file = io.getStdErr();
```
(it works if turn off safety option.

andrewrk:
I think this can be solved by improving the default log implementation

andrewrk:
On freestanding you may want to provide a `log` function next to your `main`, otherwise you will miss the safety messages from the allocator. I made the default ignore log messages on wasm32-freestanding.

Follow-up issue: make std.progress work in freestanding mode.

PR #12021, opened 2022-07-06
`std.log`: give friendly error to freestanding users

r00ster91:
@InKryption and I came to the conclusion that it'd be much nicer if instead of doing nothing `std.log.debug` etc. gave an error if you use it on a freestanding target with no top level `log` function available.
The reason this `builtin.os.tag == .freestanding` branch was added is #6252 where the issue was that when they tried to use the GPA, it would also semantically analyze `std.log`'s stuff which in turn would cause errors because most likely you did not implement any of that I/O stuff. Of course it is silly to require the user to implement all that I/O stuff when you just want to allocate.

However I think that if you choose to use the GPA, an allocator that can log messages, you should accept that you will also have to deal with the burden of some of the GPA's functionality not being implemented, so with this change you will be required to at least provide an empty top level function declaration like this in order for your code to compile again:

```zig
pub fn log(
    comptime message_level: std.log.Level,
    comptime scope: @Type(.EnumLiteral),
    comptime format: []const u8,
    args: anytype,
) void {
    _ = message_level;
    _ = scope;
    _ = format;
    _ = args;
}
```

An added benefit is that it's much harder to miss the safety messages from the allocator ([like andrew said](https://github.com/ziglang/zig/issues/6252#issuecomment-687354380)).

It's just incredibly confusing when you try to use `std.log.debug` on say `wasm32-freestanding` and it somehow compiles (while `_ = std.io.getStdOut().write("hello") catch unreachable;` doesn't due to a missing `os` implementation) and then when you execute it it doesn't do anything.

### DIFF
--- a/test/standalone/issue_7030/main.zig
+++ b/test/standalone/issue_7030/main.zig
@@ -1,5 +1,17 @@
 const std = @import("std");
 
+pub fn log(
+    comptime message_level: std.log.Level,
+    comptime scope: @Type(.EnumLiteral),
+    comptime format: []const u8,
+    args: anytype,
+) void {
+    _ = message_level;
+    _ = scope;
+    _ = format;
+    _ = args;
+}
+
 pub fn main() anyerror!void {
     std.log.info("All your codebase are belong to us.", .{});
 }
--- a/lib/std/log.zig
+++ b/lib/std/log.zig
@@ -156,11 +156,11 @@ pub fn defaultLog(
     comptime format: []const u8,
     args: anytype,
 ) void {
-    if (builtin.os.tag == .freestanding) {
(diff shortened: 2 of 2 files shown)


test/standalone/issue_7030/main.zig
```diff
@@ -1,5 +1,17 @@
...
+    args: anytype,
+) void {
+    _ = message_level;
+    _ = scope;
+    _ = format;
+    _ = args;
+}
+
 pub fn main() anyerror!void {
     std.log.info("All your codebase are belong to us.", .{});
```
r00ster91, line 16: This is a very good example too. This test was basically made to do nothing, which is confusing. Now with this empty `log` above it might be clearer that this doesn't actually log anything.

PR #12023, opened 2022-07-06
Sema: accept thumb and thumbeb with ARM calling conventions

rbino:
Commit 3014a0d5f1dcbcfdcec3852ffd54f3c589fe3e83 added calling convention validation, but left out thumb and thumbeb from the archs that can use the ARM-specific calling conventions. This plus a8a7f15106abef7c647d4de572a04b77ce048dd6 made compilation fail for thumb targets.


andrewrk:
Thanks!

PR #12035, opened 2022-07-07
Indent multiline error messages properly

r00ster91:
## Before

image

## After

image

Now we can finally write long multiline error messages and they'll look nice!

While working on #12021 I thought it'd be really nice if we could actually make multiline error messages look nice so I worked on this. In hindsight, maybe the newline after the semicolon there wasn't really needed but that's besides the point of this PR.

One argument against this might be that error messages should be concise and you should never need more than one line but I think that's just not really a helpful rule.

I mainly did this so that *users* (i.e. `"this is bad"`) rather than lang devs (i.e. `"struct 'x.x' has no member named 'main'"`) could write longer error messages.
So if I as a lib dev for example like to give my users long helpful error messages because that's my style, why not allow me to write longer error messages that span more than one line?
Of course this was always possible but it always felt off when all lines after the first one weren't properly indented.
Originally I thought about only doing this for the messages of `@compileError` but ultimately that seemed even more complicated so this is doing it for every error message which means lang devs can use this feature for multiline error messages too, which is more consistent anyway.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -393,7 +393,8 @@ pub const AllErrors = struct {
             indent: usize,
         ) anyerror!void {
             const stderr = stderr_file.writer();
-            const counting_stderr = std.io.countingWriter(stderr).writer();
+            var counting_writer = std.io.countingWriter(stderr);
+            const counting_stderr = counting_writer.writer();
             switch (msg) {
                 .src => |src| {
                     try counting_stderr.writeByteNTimes(' ', indent);
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -393,28 +393,22 @@ pub const AllErrors = struct {
             indent: usize,
         ) anyerror!void {
             const stderr = stderr_file.writer();
-            // The length of the part before the error message; e.g. "file.zig:4:5: error: ".
-            var prefix_len: usize = 0;
+            const counting_stderr = std.io.countingWriter(stderr).writer();
             switch (msg) {
                 .src => |src| {
-                    try stderr.writeByteNTimes(' ', indent);
-                    prefix_len += indent;
+                    try counting_stderr.writeByteNTimes(' ', indent);
                     ttyconf.setColor(stderr, .Bold);
-                    try stderr.print("{s}:{d}:{d}: ", .{
-                        src.src_path,
-                        src.line + 1,
-                        src.column + 1,
-                    });
-                    prefix_len += std.fmt.count("{s}:{d}:{d}: ", .{
+                    try counting_stderr.print("{s}:{d}:{d}: ", .{
                         src.src_path,
                         src.line + 1,
                         src.column + 1,
                     });
                     ttyconf.setColor(stderr, color);
-                    try stderr.writeAll(kind);
-                    prefix_len += kind.len;
-                    try stderr.writeAll(": ");
-                    prefix_len += 2;
+                    try counting_stderr.writeAll(kind);
+                    try counting_stderr.writeAll(": ");
+                    // This is the length of the part before the error message:
+                    // e.g. "file.zig:4:5: error: "
+                    const prefix_len = counting_stderr.context.bytes_written;
                     ttyconf.setColor(stderr, .Reset);
                     ttyconf.setColor(stderr, .Bold);
                     if (src.count == 1) {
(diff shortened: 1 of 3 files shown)


test/compile_errors.zig
```diff
@@ -138,6 +138,42 @@ pub fn addCases(ctx: *TestContext) !void {
...
+            \\tmp.zig:2:5: error:
+            \\                    hello!
+            \\                    I'm a multiline error message.
+            \\                    I hope to be very useful!
+            \\
+            \\                    also I will leave this trailing newline here if you don't mind
+            \\
+        });
+    }
+
```
r00ster91, line 176: This doesn't pass for me locally so it likely won't work in CI/CD here either but it definitely works when I test it on my own. I think it has to do with it just not really being possible to run stage2 tests like this yet.
Should I just comment it out for now like the stage2 tests below?

src/Compilation.zig
```diff
@@ -414,24 +440,25 @@ pub const AllErrors = struct {
                         }
                     }
                     for (src.notes) |note| {
-                        try note.renderToStdErrInner(ttyconf, stderr_file, "note:", .Cyan, indent);
+                        try note.renderToStdErrInner(ttyconf, stderr_file, "note", .Cyan, indent);
                     }
                 },
                 .plain => |plain| {
```
r00ster91, line 446: I wasn't sure what this was for so I didn't touch it (apart from the refactor) but I'm assuming output printed with this variant only uses one line anyway, right? But I don't know, does anyone know what this variant is for and whether I should use `src.writeMsg` here too?

src/Compilation.zig
```diff
@@ -379,23 +393,35 @@ pub const AllErrors = struct {
...
                 .src => |src| {
                     try stderr.writeByteNTimes(' ', indent);
+                    prefix_len += indent;
                     ttyconf.setColor(stderr, .Bold);
                     try stderr.print("{s}:{d}:{d}: ", .{
                         src.src_path,
                         src.line + 1,
                         src.column + 1,
                     });
+                    prefix_len += std.fmt.count("{s}:{d}:{d}: ", .{
```
r00ster91, line 408: It's unfortunate that `stderr.print` etc. don't return the written bytes... has this been discussed before? I understand that it might be too annoying to have `_ = ` before every line but at the same time there are definitely use cases for the byte count like you can see here.

Vexu: You can use an `std.io.countingWriter` for that.

r00ster91: You're right! Thank you!

ominitay:
oh wow thank you so much for this lol this annoyed me so much previously

src/Compilation.zig
Vexu: `.plain` is only used in a few cases where there is no source line, you can safely assume them to never include new lines.

src/Compilation.zig
ominitay: Just write the escape codes to the inner writer.

r00ster91: Ah, you guys are so right!

Vexu (approved):

test/compile_errors.zig
```diff
@@ -138,6 +138,43 @@ pub fn addCases(ctx: *TestContext) !void {
         "tmp.zig:2:1: error: invalid character: '\\t'",
     });
 
+    // TODO: uncomment this in stage2
```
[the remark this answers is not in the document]
Vexu: This is hard to test without touching many places which would only have to be done to test these two cases that are not too likely to regress so I opted to delete this test instead.

src/Compilation.zig
```diff
@@ -393,7 +393,8 @@ pub const AllErrors = struct {
             indent: usize,
         ) anyerror!void {
             const stderr = stderr_file.writer();
-            const counting_stderr = std.io.countingWriter(stderr).writer();
+            var counting_writer = std.io.countingWriter(stderr);
+            const counting_stderr = counting_writer.writer();
```
r00ster91, line 397: Interesting. I did not really encounter any problem with this but yes, when I wrote it I was wondering whether it would compile because it's using a temporary. Shouldn't this have emitted a compile error if it wasn't valid? Did you encounter any UB I didn't notice?

test/compile_errors.zig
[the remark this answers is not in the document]
r00ster91: @Vexu that is unfortunate. Why would many places have to be touched? I think this should work.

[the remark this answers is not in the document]
Vexu: It's not that they couldn't be compared, the test runner just didn't actually render the messages. It didn't require as much changes as I expected but a lot more fiddling than it's worth.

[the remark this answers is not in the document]
r00ster91: Oh, nice that you were able to get it to work!

PR #12036, opened 2022-07-07
RunStep: allow emulation to run binaries

Luukdegram:
When a user of build.zig crosscompiles for a target that cannot be ran on the host platform, we check if any of the emulation options are enabled (such as `enable_wasmtime`). If enabled, we run the compiled binary on the emulation platform/runtime instead, or provide an error when no emulation can be done. i.e. no option was enabled.

This introduces the same functionality when using `zig build test` and will also allow for better standalone tests (such as linker tests) to be cross-compiled and ran regardless of host platform.

### DIFF
--- a/lib/std/build/RunStep.zig
+++ b/lib/std/build/RunStep.zig
@@ -155,12 +155,132 @@ fn make(step: *Step) !void {
             .bytes => |bytes| try argv_list.append(bytes),
             .file_source => |file| try argv_list.append(file.getPath(self.builder)),
             .artifact => |artifact| {
+                const target_info = std.zig.system.NativeTargetInfo.detect(self.builder.allocator, artifact.target) catch unreachable;
+                const need_cross_glibc = artifact.target.isGnuLibC() and artifact.is_linking_libc;
+                var ok = true; // can either run or emulate the compiled binary
+                switch (self.builder.host.getExternalExecutor(target_info, .{
+                    .qemu_fixes_dl = need_cross_glibc and self.builder.glibc_runtimes_dir != null,
+                    .link_libc = artifact.is_linking_libc,
+                })) {
+                    .bad_os_or_cpu => {
+                        const host_name = self.builder.host.target.zigTriple(self.builder.allocator) catch unreachable;
+                        const foreign_name = artifact.target.zigTriple(self.builder.allocator) catch unreachable;
+                        std.debug.print(
+                            "the host system ({s}) does not appear to be capable of executing binaries from the target ({s}).\n",
+                            .{ host_name, foreign_name },
+                        );
+                        ok = false;
+                    },
+                    .bad_dl => |foreign_dl| {
+                        const host_dl = self.builder.host.dynamic_linker.get() orelse "(none)";
+                        std.debug.print(
+                            "the host system does not appear to be capable of executing binaries from the target because the host dynamic linker is '{s}', while the target dynamic linker is '{s}'. Consider setting the dynamic linker.\n",
+                            .{ host_dl, foreign_dl },
+                        );
+                        ok = false;
+                    },
+                    .native => {},
+                    .rosetta => if (!self.builder.enable_rosetta) {
+                        const host_name = self.builder.host.target.zigTriple(self.builder.allocator) catch unreachable;
+                        const foreign_name = artifact.target.zigTriple(self.builder.allocator) catch unreachable;
+                        std.debug.print(
+                            "the host system ({s}) does not appear to be capable of executing binaries " ++
+                                "from the target ({s}). Consider enabling rosetta.\n",
+                            .{ host_name, foreign_name },
+                        );
+                        ok = false;
+                    },
+                    .wine => |bin_name| if (self.builder.enable_wine) {
+                        try argv_list.append(bin_name);
+                    } else {
+                        const host_name = self.builder.host.target.zigTriple(self.builder.allocator) catch unreachable;
+                        const foreign_name = artifact.target.zigTriple(self.builder.allocator) catch unreachable;
+                        std.debug.print(
+                            "the host system ({s}) does not appear to be capable of executing binaries " ++
+                                "from the target ({s}). Consider enabling wine.\n",
+                            .{ host_name, foreign_name },
+                        );
+                        ok = false;
+                    },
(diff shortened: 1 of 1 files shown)


Vexu:
On my system foreign binaries are automatically ran on qemu, this change breaks that.

Luukdegram:
> On my system foreign binaries are automatically ran on qemu, this change breaks that.

Personally, I'm okay with breaking that behavior. This change also introduces the ability to provide `-fno-qemu` and therefore ensures it's not accidentally run on Qemu when unwanted. Making it only a warning, wouldn't allow for that. The same behavior occurs when cross-compiling to x86_64 from an M1 machine as Rosetta will run the binary for you by default. I'm happy to change this though if we can come up with solid arguments why the implicit behavior outweighs the explicit's.

kubkon:
> > Is that by design?
> 
> Yes, I'm pretty sure it's a qemu feature but I can't seem to find where I heard of it.
> 
> > Wouldn't an explicit flag `-fqemu` be better?
> 
> That's debatable. I think it's nice though so far I've only ever used it with `build-exe` (as far as I remember).
> 
> > Also, what happens if you try running when target is foreign OS, e.g. `macos`, or WASI?
> 
> It's qemu user space emulation so it only applies to different architectures.
> 
> > This change also introduces the ability to provide `-fno-qemu` and therefore ensures it's not accidentally run on Qemu when unwanted.
> 
> That doesn't seem incompatible with allowing implicit qemu when no flag is specified.

If this is the case, we have an inconsistency on our hands by requiring `-fqemu` for `zig test` but not for the run step.

kubkon:
FWIW what I want to be able to do is having a macho specific linker test that includes a `RunStep` but is only run if on a runnable target. I think Luuk already motivated this in the description though but it doesn't hurt to reiterate the reason again.

Luukdegram:
> FWIW what I want to be able to do is having a macho specific linker test that includes a `RunStep` but is only run if on a runnable target. I think Luuk already motivated this in the description though but it doesn't hurt to reiterate the reason again.
> 
> Current status quo AFAIK forces me to explicitly check for a runnable host in each build.zig which is cumbersome but can be done if we don't go through with this PR.

There's another option we could do instead: Create a new step within `test/link` with this logic that we import into each linker-test's `build.zig` and call that instead of `exe.run()`.

kubkon:
> > FWIW what I want to be able to do is having a macho specific linker test that includes a `RunStep` but is only run if on a runnable target. I think Luuk already motivated this in the description though but it doesn't hurt to reiterate the reason again.
> > Current status quo AFAIK forces me to explicitly check for a runnable host in each build.zig which is cumbersome but can be done if we don't go through with this PR.
> 
> There's another option we could do instead: Create a new step within `test/link` with this logic that we import into each linker-test's `build.zig` and call that instead of `exe.run()`. (Though, I think this functionality is beneficial to be in the standard step).

I would like to avoid this route as RunStep already implements exactly what we need including comparing with expected stdout output.

Vexu:
> If this is the case, we have an inconsistency on our hands by requiring `-fqemu` for `zig test` but not for the run step.

Works for me?
```sh-session
$ bat -p a.zig
test {
    @import("std").debug.print("Hello, world!\n", .{});
}
$ zig test a.zig -target aarch64-linux
Test [1/1] test ""... Hello, world!
All 1 tests passed.
$ 
```

andrewrk:
This is related to what happens when you run `zig test` with a foreign target:

```
[nix-shell:~/Downloads/zig/build-release]$ stage3/bin/zig test test.zig -target aarch64-linux
warning: the host system (x86_64-linux.5.10.118...5.10.118-gnu.2.33) does not appear to be capable of executing binaries from the target (aarch64-linux.3.16...5.10.81-musl). Consider using '--test-cmd qemu-aarch64 --test-cmd-bin' to run the tests
error: the following command failed with 'InvalidExe':
zig-cache/o/129cc9c545e9aa7cff4d0ac00280fd87/test /home/andy/Downloads/zig/build-release/stage3/bin/zig
```

Some systems, unlike mine, have [binfmt_misc](https://en.wikipedia.org/wiki/Binfmt_misc) installed. I'm guessing this is what @Vexu is using to run foreign binaries on his system. The way it works is you just directly try to run a foreign binary, and if the kernel discovers the target does not match the host, then it uses QEMU or other software to seamlessly interpret the binary.

Rosetta works similarly, in that the proper way to use it is to directly execute the foreign binary as if it were native.

If you look carefully at the `zig test` output above, you can see that it does just try to run the binary, and is indeed compatible with binfmt_misc/Rosetta/etc. The warning is actually only printed if the execve *fails*. **Trying the execution and noticing that it fails is the only correct way to detect that a binary cannot be executed natively.**

I think RunStep should work similarly, in that it should naively try to execute the child process as if it were native. However, if the execution fails with InvalidExe, *then* the logic should kick in for using a third party executor to execute the binary.

Please correct me if I'm wrong, but I believe this should solve everybody's use cases handily.

andrewrk:
I don't think it's reasonable for RunStep to succeed, by default, if it cannot actually run a binary. However, I think it would be OK to have a flag called something like `skip_incompatible_hosts` which makes the step silently succeed when the binary cannot be run on the host. In such case, zig would *try* to run the binary on the host, notice that the error was `InvalidExe`, and then notice that flag.

Luukdegram:
With those comments in mind, I'll happily rework this PR tomorrow with those improvements.

Luukdegram:
Closing this PR and opening a new one with a specific step that meets the requirements for our linker tests.
Here's why after working on this a bit more:
- `RunStep` is also used for running commands, so we use `ChildProcess`. We do not know if an executable failed due to incompatibility binary on the host platform, or a different reason (we only have the exit code).
- Because we act on failure, binaries will be run on Rosetta by default, even when passing `-fno-rosetta`.
- We still *must* ensure the exit code is 0 for our tests to succeed. However, if we were to introduce a flag to skip incompatible binaries, we could introduce false positives as it may skip valid failures.

I do think allowing emulations on `RunStep` is a solid addition, but not in the form I introduced or was discussed here. While we could make it work with our requirements, I'd rather do extra work and/or copy some code to build a use-case-specific step rather than create what I'd consider hacky code. I'd like to work on getting our linker tests more feature-full first before working on this and therefore closing this PR for now, until I or someone else picks this up again.

Issue #9889, opened 2021-10-04
std.math: add fns degrees, radians, isqrt and fract


binarygcd:
**degrees** and **radians**
Converting to and from degrees and radians is a common math task so these two could be really handy.

**isqrt**
I know that **math.sqrt** can return an int sqrt but it's more clear if there's something that indicates it. And an int sqrt isn't "exactly" a sqrt. It's more like a floored sqrt.

**fract**
It returns the fractional part(e.g. .5 from 1.5).
This is really just for convenience. Not as significant as the other three.

nektro:
implementation is here <https://github.com/ziglang/zig/blob/master/lib/std/math/sqrt.zig#L7-L34>

binarygcd:
I already saw that. That's why I said "**I know that math.sqrt can return an int sqrt**". What follows after that is the reason.

nektro:
the implementation supports floats too. that's why It doesn't have any signifying information

DanikVitek:
Also **degrees** and **radians** also already exist, as well as conversion constants:
https://github.com/ziglang/zig/blob/96e4825fbba714322ad750ee179da3d5f8463bb5/lib/std/math.zig#L266-L328

PR #12052, opened 2022-07-09
Add std.math.degrees and std.math.radians

frmdstryr:
Implements simple conversion functions for `degrees` and `radians`. 

A quick search brings up https://github.com/ziglang/zig/issues/9889

### DIFF
--- a/lib/std/math.zig
+++ b/lib/std/math.zig
@@ -293,31 +293,31 @@ pub inline fn tan(value: anytype) @TypeOf(value) {
 }
 
 // Convert an angle in radians to degrees. T must be a float type.
-pub inline fn degrees(comptime T: type, angle_in_radians: T) T {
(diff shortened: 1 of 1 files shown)


topolarity:
It'd be clearer to mention both sides of the conversion in the function names, such as the relatively-standard `deg2rad`/`rad2deg` or the expanded `degreesToRadians`/`radiansToDegrees`

r00ster91:
I actually needed this recently.

As for the naming, if you think the expanded variants are too long, I'd go with a hybrid that is not old-fashioned but also not too long: `degToRad` and `radToDeg`.
But personally I'd prefer the expanded variants anyway.

InKryption:
I would also prefer the expanded variants. `radToDeg` and `degToRad` look weird.

frmdstryr:
Thanks!

PR #12059, opened 2022-07-09
Implement EmulatableRunStep for linker tests

Luukdegram:
Implements a new step that can run and compare the output for foreign binaries that will be run through emulation when enabled (or directly on the host when non-foreign binary). When disabled, it will skip running the binary and produce a warning instead. Said warning can be hidden by either setting the option manually in `build.zig` on the step, or passing `-Dhide-foreign-warnings`. This option is automatically added when creating the step.

The main goal of this step is to help us test the various linkers and we want to run it for multiple targets. This is especially useful when cross-compiling for ELF and MachO as they support more than a single architecture. For this reason, a handy function was appended to `CheckObjectStep` called `runAndCompare` that will automatically create this step and provide the artifact that was used for the original step. 

This PR does not yet update all the current linker tests to use this new step.

### DIFF
--- a/ci/zinc/linux_test.sh
+++ b/ci/zinc/linux_test.sh
@@ -67,6 +67,7 @@ $STAGE1_ZIG build test-cli              -fqemu -fwasmtime
 $STAGE1_ZIG build test-run-translated-c -fqemu -fwasmtime
 $STAGE1_ZIG build docs                  -fqemu -fwasmtime
 $STAGE1_ZIG build test-cases            -fqemu -fwasmtime
+$STAGE1_ZIG build test-link             -fqemu -fwasmtime
 
 # Produce the experimental std lib documentation.
 mkdir -p "$RELEASE_STAGING/docs/std"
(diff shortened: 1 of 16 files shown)


kubkon (changes requested):
I will review the actual code more thoroughly later, but two things that immediately come to mind: 1) would you mind enabling linker tests in the Linux CI now, and 2) could you remove the if-macos check from `test/link.zig` so that we run MachO tests regardless of the host?

rohlem:
Minor non-functional bikeshedding, I think "[emulatable](https://en.wiktionary.org/wiki/emulatable)" is a more common term.

Luukdegram:
> Minor non-functional bikeshedding, I think "[emulatable](https://en.wiktionary.org/wiki/emulatable)" is a more common term.

Good call, thanks!

kubkon (changes requested):
Looks great, thanks @Luukdegram! I have only one suggested change. However, for the future, I was thinking maybe we could somehow re-use the code between `RunStep` and `EmulatableRunStep` since there's a lot of overlap.

Luukdegram:
> Looks great, thanks @Luukdegram! I have only one suggested change. However, for the future, I was thinking maybe we could somehow re-use the code between `RunStep` and `EmulatableRunStep` since there's a lot of overlap.

Rather than waiting for the future, I went ahead and changed the logic to share the overlapping parts of the steps.

kubkon (approved):
LGTM!

PR #12060, opened 2022-07-09
std.fs: split `Dir` into `IterableDir`

### LINKED ISSUE #12007: fs.cwd().openDir().iterate() panics if called without `OpenDirOptions.iterate = true` on Linux (Ubuntu), works fine on MacOS
### Zig Version

0.10.0-dev.2577+5816d3eae

### Steps to Reproduce

Call this:

```zig
var children = try std.fs.cwd().openDir("some-dir", .{});
defer children.close();
_ = children.iterate();
```

### Expected Behavior

The call should return an error as the default is iterate=false.

On MacOS, no error happens at all, which is probably also a bug?

### Actual Behavior

```
thread 1827 panic: reached unreachable code
/opt/hostedtoolcache/zig/zig-linux-x86_64-0.9.1/x64/lib/std/os.zig:4342:18: 0x2a5567 in std.os.lseek_SET (build)
        .BADF => unreachable, // always a race condition
                 ^
/opt/hostedtoolcache/zig/zig-linux-x86_64-0.9.1/x64/lib/std/fs.zig:597:45: 0x29c530 in std.fs.Iterator.next (build)
                            std.os.lseek_SET(self.dir.fd, 0) catch unreachable; // EBADF here likely means that the Dir was not opened with iteration permissions
                                            ^
/home/runner/work/zig-common-tasks/zig-common-tasks/build.zig:17:34: 0x29e288 in .@build.testAllSamples (build)
    var child = try iterator.next();
                                 ^
/home/runner/work/zig-common-tasks/zig-common-tasks/build.zig:6:19: 0x2920c6 in .@build.build (build)
    testAllSamples(b, test_step) catch |e| {
                  ^
/opt/hostedtoolcache/zig/zig-linux-x86_64-0.9.1/x64/lib/std/special/build_runner.zig:220:28: 0x2838e4 in runBuild (build)
        .Void => root.build(builder),
                           ^
/opt/hostedtoolcache/zig/zig-linux-x86_64-0.9.1/x64/lib/std/special/build_runner.zig:202:17: 0x27855f in main (build)
    try runBuild(builder);
                ^
/opt/hostedtoolcache/zig/zig-linux-x86_64-0.9.1/x64/lib/std/start.zig:561:37: 0x26ff1a in std.start.callMain (build)
                                    ^
/opt/hostedtoolcache/zig/zig-linux-x86_64-0.9.1/x64/lib/std/start.zig:495:12: 0x250b2e in std.start.callMainWithArgs (build)
           ^
/opt/hostedtoolcache/zig/zig-linux-x86_64-0.9.1/x64/lib/std/start.zig:409:17: 0x24fbc6 in std.start.posixCallMainAndExit (build)
                ^
/opt/hostedtoolcache/zig/zig-linux-x86_64-0.9.1/x64/lib/std/start.zig:322:5: 0x24f9d2 in std.start._start (build)
    ^
error: the following build command crashed:
/home/runner/work/zig-common-tasks/zig-common-tasks/zig-cache/o/68e59b58dd56ed02a9db570b274bcbd5/build /opt/hostedtoolcache/zig/zig-linux-x86_64-0.9.1/x64/zig /home/runner/work/zig-common-tasks/zig-common-tasks /home/runner/work/zig-common-tasks/zig-common-tasks/zig-cache /home/runner/.cache/zig test
```

Vexu:
`walk` and `iterate` are moved to a new type `IterableDir`. `Dir` receives a new `openIterableDir` function which returns an `IterableDir` and also an `intoIterable` function which converts the `Dir` into an `IterableDir` asserting that it was opened with `iterate = true`.

Closes #12007

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -283,8 +283,10 @@ pub fn renameW(old_dir: Dir, old_sub_path_w: []const u16, new_dir: Dir, new_sub_
     return os.renameatW(old_dir.fd, old_sub_path_w, new_dir.fd, new_sub_path_w);
 }
 
-pub const Dir = struct {
-    fd: os.fd_t,
+/// A directory that can be iterated. It is *NOT* legal to initialize this with a regular `Dir`
+/// that has been opened without iteration permission.
+pub const IterableDir = struct {
+    dir: Dir,
 
     pub const Entry = struct {
         name: []const u8,
@@ -779,7 +781,7 @@ pub const Dir = struct {
         else => @compileError("unimplemented"),
     };
 
-    pub fn iterate(self: Dir) Iterator {
+    pub fn iterate(self: IterableDir) Iterator {
         switch (builtin.os.tag) {
             .macos,
             .ios,
@@ -789,7 +791,7 @@ pub const Dir = struct {
             .openbsd,
             .solaris,
             => return Iterator{
-                .dir = self,
+                .dir = self.dir,
                 .seek = 0,
                 .index = 0,
                 .end_index = 0,
@@ -797,14 +799,14 @@ pub const Dir = struct {
                 .first_iter = true,
             },
             .linux, .haiku => return Iterator{
-                .dir = self,
+                .dir = self.dir,
                 .index = 0,
                 .end_index = 0,
                 .buf = undefined,
                 .first_iter = true,
             },
             .windows => return Iterator{
-                .dir = self,
+                .dir = self.dir,
                 .index = 0,
                 .end_index = 0,
                 .first_iter = true,
@@ -812,7 +814,7 @@ pub const Dir = struct {
                 .name_data = undefined,
             },
             .wasi => return Iterator{
-                .dir = self,
+                .dir = self.dir,
                 .cookie = os.wasi.DIRCOOKIE_START,
                 .index = 0,
                 .end_index = 0,
@@ -833,11 +835,11 @@ pub const Dir = struct {
             dir: Dir,
             basename: []const u8,
             path: []const u8,
-            kind: Dir.Entry.Kind,
+            kind: IterableDir.Entry.Kind,
         };
 
         const StackItem = struct {
-            iter: Dir.Iterator,
+            iter: IterableDir.Iterator,
             dirname_len: usize,
         };
 
@@ -857,7 +859,7 @@ pub const Dir = struct {
                     }
                     try self.name_buffer.appendSlice(base.name);
                     if (base.kind == .Directory) {
-                        var new_dir = top.iter.dir.openDir(base.name, .{ .iterate = true }) catch |err| switch (err) {
+                        var new_dir = top.iter.dir.openIterableDir(base.name, .{}) catch |err| switch (err) {
                             error.NameTooLong => unreachable, // no path sep in base.name
                             else => |e| return e,
                         };
@@ -896,11 +898,10 @@ pub const Dir = struct {
     };
 
     /// Recursively iterates over a directory.
-    /// `self` must have been opened with `OpenDirOptions{.iterate = true}`.
     /// Must call `Walker.deinit` when done.
     /// The order of returned file system entries is undefined.
     /// `self` will not be closed after walking it.
-    pub fn walk(self: Dir, allocator: Allocator) !Walker {
+    pub fn walk(self: IterableDir, allocator: Allocator) !Walker {
         var name_buffer = std.ArrayList(u8).init(allocator);
         errdefer name_buffer.deinit();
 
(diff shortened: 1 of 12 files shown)


lib/std/fs.zig
```diff
@@ -1507,6 +1521,26 @@ pub const Dir = struct {
...
+        var adjusted_args = args;
+        adjusted_args.iterate = true;
+        const new_dir = try self.openDir(sub_path, adjusted_args);
+        return IterableDir{ .dir = new_dir };
+    }
+
+    /// Convert `self` into an iterable directory.
+    /// Asserts that `self` was opened with `iterate = true`.
+    pub fn intoIterable(self: Dir) IterableDir {
+        if (builtin.mode == .Debug) {
```
andrewrk, line 1538:
```suggestion
        if (std.debug.runtime_safety) {
```

lib/std/fs.zig
```diff
@@ -918,6 +917,21 @@ pub const Dir = struct {
...
+        self.dir.close();
+        self.* = undefined;
+    }
+};
+
+pub const Dir = struct {
+    fd: os.fd_t,
+    iterable: @TypeOf(iterable_safety) = iterable_safety,
+
+    const iterable_safety = if (builtin.mode == .Debug) false else {};
```
andrewrk, line 930:
```suggestion
    const iterable_safety = if (std.debug.runtime_safety) false else {};
```

andrewrk (changes requested):
It seems like the main benefit of this change is so that calling `intoIterable` gives an assert in safe modes, so that when testing on an operating systems which allow all dirs to iterate, the programmer will not incorrectly believe that it works when it will fail on a different operating system that needs the iterable flag to be correct.

This benefit could be achieved more simply by adding the safety flag to `Dir` instead of introducing `IterableDir` as a new type.

I'm a bit concerned here with the added clunkiness of the API; having two types associated with the same resource. It does prevent a certain kind of mistake from being made, but it also causes confusion about which type has which methods and makes usage sites more convoluted.

We had a similar pattern earlier with mutexes. It used to be:

```
var held = mutex.acquire();
defer held.release();
```

...which was a nice way to help the user not forget to call unlock(). However it ended up being really annoying and actually more error prone any time the API needed to be wrapped, and we ended up just going back to the more straightforward:

```
mutex.lock();
defer mutex.unlock();
```

How do you feel about my counter proposal of adding the safety field directly to Dir?

Vexu:
> It seems like the main benefit of this change is so that calling `intoIterable` gives an assert in safe modes, so that when testing on an operating systems which allow all dirs to iterate, the programmer will not incorrectly believe that it works when it will fail on a different operating system that needs the iterable flag to be correct.
> 
> This benefit could be achieved more simply by adding the safety flag to `Dir` instead of introducing `IterableDir` as a new type.

I don't think the main benefit here is `intoIterable` giving an assertion. I think the main benefit is that attempting to iterate a regular `Dir` will give a compile error and force the programmer to think about this issue before they even have a chance to test their program.

> I'm a bit concerned here with the added clunkiness of the API; having two types associated with the same resource. It does prevent a certain kind of mistake from being made, but it also causes confusion about which type has which methods and makes usage sites more convoluted.

I think the separation between the types is quite clear here, `IterableDir` only allows iteration and walking while `Dir` only allows all other directory operations. `dir.intoIterable` and `iterable_dir.dir` also makes it easy to go between the types depending on which kind of operations you need to do.

> We had a similar pattern earlier with mutexes. It used to be:
> 
> ```
> var held = mutex.acquire();
> defer held.release();
> ```
> 
> ...which was a nice way to help the user not forget to call unlock(). However it ended up being really annoying and actually more error prone any time the API needed to be wrapped, and we ended up just going back to the more straightforward:
> 
> ```
> mutex.lock();
> defer mutex.unlock();
> ```

I don't think this situation is comparable. `Mutex.Held` used to encode state of the resource while `IterableDir` encodes available operations. Whether or not a directory supports iteration is decided on creation and you cannot change that state without reopening the directory.

ominitay:
Just rounding up comments made about this on Discord:

`OpenDirFlags.iterate` should be removed, if we are to add the type-based system; there isn't a need to have two methods. The only difference between them is that one is a runtime crash and one is a compile error. `TmpDir` should be split into `TmpDir` and `TmpIterableDir` to correspond with this.

ominitay:
You've said this a number of times, but could you please provide an actual example of this?

There is no case in which going from `Dir` to `IterableDir` is necessary, as the types exist specifically to indicate the permissions. If a directory has been opened as iterable, one will have `IterableDir`. It is easy to go from an `IterableDir` to a `Dir`, which is the only operation which seems necessary...

If a function needs to iterate on a directory, it should take `IterableDir`. Else, it should take `Dir`. This should be enforced by the API, imo.

Have you thought this through? `anytype` wouldn't work with an interface like this anyway...

Vexu:
> There is no case in which going from `Dir` to `IterableDir` is necessary, as the types exist specifically to indicate the permissions. If a directory has been opened as iterable, one will have `IterableDir`. It is easy to go from an `IterableDir` to a `Dir`, which is the only operation which seems necessary...

This is not the point you were making earlier or at least you couldn't be bothered to make it clear. This I could agree on.

> Have you thought this through? `anytype` wouldn't work with an interface like this anyway...

That is exactly how `anytype` is being used, just take a look at the thousands of instances of `writer: anytype` out there.

InKryption:
@Vexu Would you maybe consider omitting the components of this PR that would allow for conversion from `Dir` to `IterableDir`? At the very least, as an experiment to see if any of the problems you're alluding to actually emerge. They can always be added later on (wouldn't be as easy to remove them if present, being that it would break any code relying on it). 


Otherwise, I think it'll turn out to be far more harmful for people to unwittingly take a `Dir` parameter/use a `Dir` field, only to then use `intoIterable`, when it'd have made far more sense to just have an `IterableDir` in the first place (an incorrect path of least resistance).

ominitay:
> This is not the point you were making earlier or at least you couldn't be bothered to make it clear. This I could agree on.

Couldn't be bothered?? :roll_eyes: This is my entire criticism I've been making of your PR: that `intoIterable()` (and therefore `OpenDirFlags.iterate`) should be removed, as they serve no purpose.

One of my concerns is that, for example, someone has an already-written function which takes a `Dir`. They add in the requirement to iterate on the directory which they take as an argument after the fact. The presence of `intoIterable()` provides low resistance, and is likely to be the change they make, despite the 'correct' way to be to take an `IterableDir` as an argument. 

> That is exactly how anytype is being used, just take a look at the thousands of instances of writer: anytype out there.

But you can *only* call `iterate()` or `walk()` on an `IterableDir`. It's nothing like a `Writer`, as writers all have a common set of functions. The entire point of `IterableDir` is that it doesn't.

ominitay:
> And yet, I though you wanted to remove the conversion in both directions which is why I was emphasizing how it would lead to repetition of code.

Oh I see why you could have thought that lol. Glad we're on the same page now, then!

andrewrk:
A similar pattern can be drawn with passing `OpenMode.read_only` when opening a file. You'll get an error from the operating system if you try any of the write API. And yet I think having WritableFile and ReadableFile would make the API worse. I do prefer status quo, both with File and with Dir. However, I'm willing to try @Vexu's new API for a while and see how it feels.

ominitay:
> I do prefer status quo, both with File and with Dir.

I do also, but if we're to add a check for iterability, I think it's probably best to try a comptime, type-based solution first.

ominitay:
I've seen mixed feedback on this change. There are people who are very unhappy with it, proclaiming it "pointless" and a terrible API, and then you with the benefit that you know your code will be correct when it compiles. I'm not sure where we should go with this, whether we should revert and use a runtime check or stick with it. Maybe worth giving it more time?

kubkon:
I am on the fence with this change. If you look at the updated `fetch-them-macos-headers` https://github.com/ziglang/fetch-them-macos-headers/commit/96b1195a9c3d4c7308296d54f8b545c4acf66ef7, the required changes look and feel rather clunky, not to mention confusing: having to call a `.dir` on an iterable dir feels overly complicated. Perhaps wrapping it in a function would better clarify the intent. `tmpIterableDir` is suffering from the same problem now that we have to "dereference" it twice: `tmp.iterable_dir.dir`. I think this really should always be a single call such as `tmp.getDir()` or something. I am also worried that should we decide for another level of abstraction where we would wrap `IterableDir` in some `YetAnotherDir`, then the chain of wrapped `.dir` fields becomes ridiculous: `some.dir.dir.dir` and so on.

Jarred-Sumner:
This change makes using `std.fs` more tedious

If Zig had a way to do interfaces/inheritance without adding an extra layer of nesting (instead of `.dir`, just make all the functions on `dir` work with an `IterableDir`) then it would be fine

andrewrk:
After running with this change for about a year now, I have to say that I strongly don't like it and I would like to revert it.

PR #12066, opened 2022-07-10
std.net.getAddressList: fix segfault on bad hostname

### LINKED ISSUE #12065: std.net.getAddressList segfault on bad hostname
### Zig Version

0.10.0-dev.2887+3f11d1d56

### Steps to Reproduce

Run the following program via `zig run bug.zig`:

`bug.zig`:
```zig
const std = @import("std");

pub fn main() !void {
    _ = try std.net.getAddressList(std.testing.allocator, "BAD_HOSTNAME", 80);
}
```

### Expected Behavior

Program to fail with some error message about an unkown hostname.

### Actual Behavior

Segfault:
```
Segmentation fault at address 0x0
/home/marler8997/git/zig/lib/std/mem/Allocator.zig:177:23: 0x217827 in std.mem.Allocator.free (bug)
    return self.vtable.free(self.ptr, buf, buf_align, ret_addr);
                      ^
/home/marler8997/git/zig/lib/std/heap/arena_allocator.zig:41:38: 0x215741 in std.heap.arena_allocator.ArenaAllocator.deinit (bug)
            self.child_allocator.free(node.data);
                                     ^
/home/marler8997/git/zig/lib/std/net.zig:734:33: 0x22cd08 in std.net.getAddressList (bug)
    errdefer result.arena.deinit();
                                ^
/tmp/bug.zig:4:35: 0x22a861 in main (bug)
    _ = try std.net.getAddressList(std.testing.allocator, "BAD_HOSTNAME", 80);
                                  ^
/home/marler8997/git/zig/lib/std/start.zig:580:37: 0x223f4a in std.start.callMain (bug)
                                    ^
/home/marler8997/git/zig/lib/std/start.zig:514:12: 0x20ae5e in std.start.callMainWithArgs (bug)
           ^
/home/marler8997/git/zig/lib/std/start.zig:424:17: 0x209d36 in std.start.posixCallMainAndExit (bug)
                ^
/home/marler8997/git/zig/lib/std/start.zig:338:5: 0x209b42 in std.start._start (bug)
    ^
Aborted (core dumped)
```

I believe the issue here is that the `getAddressList` implementation is storing the ArenaAllocator instance inside the return value `result`, which is itself allocated by the ArenaAllocator (and then copied into `result` after being allocated).  If we invoke the `errdefer`, then the arena allocator is deinitalized via `deinit`, which would cause it to free `result` and at the same time, invalidate itself!

marler8997:
Fixes #12065

### DIFF
--- a/lib/std/net.zig
+++ b/lib/std/net.zig
@@ -731,7 +731,7 @@ pub fn getAddressList(allocator: mem.Allocator, name: []const u8, port: u16) !*A
         break :blk result;
     };
     const arena = result.arena.allocator();
-    errdefer result.arena.deinit();
+    errdefer result.deinit();
 
     if (builtin.target.os.tag == .windows or builtin.link_libc) {
         const name_c = try std.cstr.addNullByte(allocator, name);


lib/std/net.zig
```diff
@@ -731,7 +731,10 @@ pub fn getAddressList(allocator: mem.Allocator, name: []const u8, port: u16) !*A
         break :blk result;
     };
     const arena = result.arena.allocator();
-    errdefer result.arena.deinit();
+    errdefer {
+        var arena_copy = result.arena;
+        arena_copy.deinit();
+    }
```
Vexu, line 737:
```suggestion
    errdefer result.deinit();
```

`AddressList.deinit` already does this with a comment explaining why.

marler8997: Lol, looks like we shot ourselves twice with the same footgun :)

marler8997: Done

PR #12067, opened 2022-07-10
CMakeLists.txt: set LLVM minimal version to 14

BratishkaErik:
Before:
`-- Found llvm: /usr/lib/llvm/14/lib64/libLLVM-14.so (Required is at least version "13")`
After:
`-- Found llvm: /usr/lib/llvm/14/lib64/libLLVM-14.so (Required is at least version "14")`


andrewrk:
Thank you!

PR #12071, opened 2022-07-10
compiler_rt: Update Windows ABI for float<->int conversion routines

### LINKED ISSUE #12063: floating point formatting regression on windows with LLVM 14 upgrade
This test is failing on x86_64 Windows after the LLVM 14 upgrade.

https://github.com/ziglang/zig/blob/3f11d1d56d9747de974b00eab1c880bea7972c01/lib/std/fmt.zig#L2286-L2307

```
255/1523 fmt.test "std-native-Debug-bare-multi-default float.decimal"... thread 5980 panic: integer part of floating point value out of bounds
D:\a\1\s\lib\std\fmt\errol.zig:305:15: 0x7ff6c3d2d232 in mt.errol.errolInt (test.obj)
    var mid = @floatToInt(u128, val);
              ^
                                    ^
D:\a\1\s\lib\std\fmt.zig:767:27: 0x7ff6c3ff4cc2 in mt.formatFloatValue (test.obj)
        formatFloatDecimal(value, options, buf_stream.writer()) catch |err| switch (err) {
                          ^
D:\a\1\s\lib\std\fmt.zig:697:58: 0x7ff6c3fc4c2e in mt.formatValue (test.obj)
        .Float, .ComptimeFloat => return formatFloatValue(value, fmt, options, writer),
                                                         ^
D:\a\1\s\lib\std\fmt.zig:463:31: 0x7ff6c3fc4ba2 in mt.formatType (test.obj)
            return formatValue(value, actual_fmt, options, writer);
                              ^
D:\a\1\s\lib\std\fmt.zig:178:23: 0x7ff6c3f0efbd in mt.format (test.obj)
        try formatType(
                      ^
D:\a\1\s\lib\std\fmt.zig:1889:11: 0x7ff6c3e2de9a in mt.count (test.obj)
    format(counting_writer.writer(), fmt, args) catch |err| switch (err) {};
          ^
D:\a\1\s\lib\std\fmt.zig:1896:40: 0x7ff6c3e2da24 in mt.allocPrint (test.obj)
    const size = math.cast(usize, count(fmt, args)) orelse return error.OutOfMemory;
                                       ^
D:\a\1\s\lib\std\testing.zig:201:42: 0x7ff6c3d7227a in esting.expectFmt (test.obj)
    const result = try std.fmt.allocPrint(allocator, template, args);
                                         ^
D:\a\1\s\lib\std\fmt.zig:2287:18: 0x7ff6c3ba139f in mt.test "std-native-Debug-bare-multi-default float.decimal" (test.obj)
    try expectFmt("f64: 152314000000000000000000000000", "f64: {d}", .{@as(f64, 1.52314e+29)});
                 ^
D:\a\1\s\lib\test_runner.zig:79:28: 0x7ff6c3cfbd84 in root).main (test.obj)
        } else test_fn.func();
                           ^
D:\a\1\s\lib\std\start.zig:349:65: 0x7ff6c3cb7297 in tart.WinStartup (test.obj)
    std.os.windows.kernel32.ExitProcess(initEventLoopAndCallMain());
                                                                ^
Unable to dump stack trace: FileNotFound
error: the following test command failed with exit code 3:
D:\a\1\s\zig-cache\o\3f11261b1cc8172c23869e13e5cf0db2\test.exe D:\a\1\s\build\dist\bin\zig.exe
test...The following command exited with error code 1:
D:\a\1\s\build\dist\bin\zig.exe test D:\a\1\s\lib\std\std.zig --test-name-prefix std-native-Debug-bare-multi-default  --cache-dir D:\a\1\s\zig-cache --global-cache-dir C:\Users\VssAdministrator\AppData\Local\zig --name test -fno-single-threaded -I D:\a\1\s\test --zig-lib-dir D:\a\1\s\lib --enable-cache 
````

Next steps:
 * figure out why, is it an LLVM or zig bug?
 * if it's an LLVM bug, create a reduction with LLVM IR using llvm-reduce
 * report bug upstream

### PR DESCRIPTION
**Pull request by topolarity**
Resolves #12063

Starting with LLVM 14, the Libcalls to these functions are now lowered using a `Vector(2, u64)` instead of the standard ABI for i128 integers on Windows, so our compiler-rt implementation needs to be updated to expose the same ABI

Relevant LLVM revision: https://reviews.llvm.org/D110413

### DIFF
--- a/lib/compiler_rt/fixdfti.zig
+++ b/lib/compiler_rt/fixdfti.zig
@@ -1,26 +1,23 @@
 const builtin = @import("builtin");
-const arch = builtin.cpu.arch;
 const common = @import("./common.zig");
 const floatToInt = @import("./float_to_int.zig").floatToInt;
 
 pub const panic = common.panic;
 
 comptime {
-    const fixdfti_fn = if (builtin.os.tag == .windows and arch == .x86_64) b: {
-        // The "ti" functions must use Vector(2, u64) return types to adhere to the ABI
-        // that LLVM expects compiler-rt to have.
-        break :b __fixdfti_windows_x86_64;
-    } else __fixdfti;
-
-    @export(fixdfti_fn, .{ .name = "__fixdfti", .linkage = common.linkage });
+    if (common.want_windows_v2u64_abi) {
+        @export(__fixdfti_windows_x86_64, .{ .name = "__fixdfti", .linkage = common.linkage });
+    } else {
+        @export(__fixdfti, .{ .name = "__fixdfti", .linkage = common.linkage });
+    }
 }
 
 pub fn __fixdfti(a: f64) callconv(.C) i128 {
     return floatToInt(i128, a);
 }
 
-const v128 = @import("std").meta.Vector(2, u64);
+const v2u64 = @Vector(2, u64);
 
-fn __fixdfti_windows_x86_64(a: f64) callconv(.C) v128 {
-    return @bitCast(v128, floatToInt(i128, a));
+fn __fixdfti_windows_x86_64(a: f64) callconv(.C) v2u64 {
+    return @bitCast(v2u64, floatToInt(i128, a));
 }
--- a/lib/compiler_rt/fixdfti.zig
+++ b/lib/compiler_rt/fixdfti.zig
@@ -1,12 +1,26 @@
+const builtin = @import("builtin");
+const arch = builtin.cpu.arch;
 const common = @import("./common.zig");
 const floatToInt = @import("./float_to_int.zig").floatToInt;
 
 pub const panic = common.panic;
 
 comptime {
-    @export(__fixdfti, .{ .name = "__fixdfti", .linkage = common.linkage });
+    const fixdfti_fn = if (builtin.os.tag == .windows and arch == .x86_64) b: {
+        // The "ti" functions must use Vector(2, u64) return types to adhere to the ABI
+        // that LLVM expects compiler-rt to have.
+        break :b __fixdfti_windows_x86_64;
+    } else __fixdfti;
+
+    @export(fixdfti_fn, .{ .name = "__fixdfti", .linkage = common.linkage });
 }
 
 pub fn __fixdfti(a: f64) callconv(.C) i128 {
     return floatToInt(i128, a);
+}
+
+const v128 = @import("std").meta.Vector(2, u64);
+
+fn __fixdfti_windows_x86_64(a: f64) callconv(.C) v128 {
+    return @bitCast(v128, floatToInt(i128, a));
 }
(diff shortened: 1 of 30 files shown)

### CODE REVIEW

**Comment by andrewrk** [comment]
Great work - really appreciate your help with the LLVM stuff lately :pray:

File: lib/compiler_rt/fixdfti.zig [hunk 1]
```diff
@@ -1,12 +1,26 @@
...
+    } else __fixdfti;
+
+    @export(fixdfti_fn, .{ .name = "__fixdfti", .linkage = common.linkage });
 }
 
 pub fn __fixdfti(a: f64) callconv(.C) i128 {
     return floatToInt(i128, a);
 }
+
+const v128 = @import("std").meta.Vector(2, u64);
```
andrewrk, line 22:
```suggestion
const v128 = @Vector(2, u64);
```

lib/compiler_rt/fixdfti.zig
```diff
@@ -1,12 +1,26 @@
...
 
 comptime {
-    @export(__fixdfti, .{ .name = "__fixdfti", .linkage = common.linkage });
+    const fixdfti_fn = if (builtin.os.tag == .windows and arch == .x86_64) b: {
+        // The "ti" functions must use Vector(2, u64) return types to adhere to the ABI
+        // that LLVM expects compiler-rt to have.
+        break :b __fixdfti_windows_x86_64;
+    } else __fixdfti;
+
+    @export(fixdfti_fn, .{ .name = "__fixdfti", .linkage = common.linkage });
```
andrewrk, line 15: Suggest to follow the pattern of `want_aeabi `, `want_ppc_abi `, `gnu_f16_abi ` and make it look like this:

```suggestion
    if (common.windows_i128_abi) {
        @export(__fixdfti_windows_x86_64, .{ .name = "__fixdfti", .linkage = common.linkage });
    } else {
        @export(__fixdfti, .{ .name = "__fixdfti", .linkage = common.linkage });
    }
```

The comment body can go into the doc comments for `windows_i128_abi` in common.zig.

andrewrk (changes requested):
I believe the failure is due to codegen/llvm.zig emitting explicit calls to compiler_rt functions in the cases where LLVM will fail to properly do so - meaning that this function signature must be adjusted in codegen.cpp and codegen/llvm.zig to match.

Are you up for those changes too? Let me know if you want some help.

topolarity:
> Great work - really appreciate your help with the LLVM stuff lately 🙏

Happy to lend a hand! These are fun bugs to chase for now 🐞

topolarity:
> I believe the failure is due to codegen/llvm.zig emitting explicit calls to compiler_rt functions in the cases where LLVM will fail to properly do so - meaning that this function signature must be adjusted in codegen.cpp and codegen/llvm.zig to match.

Very good point - should have thought of our explicitly-lowered calls.

Let me take a stab and see how far I can get. I'll reach out if it gets hairy.

andrewrk (approved):

PR #12075, opened 2022-07-11
Stage2 validate extern types

Vexu:
Question: in stage1 pointers to zero bit types are zero bit types themselves and are therefore not allowed in extern types but with #6706 they now behave like other pointers. This means that `*void` is now allowed in extern types even though it's not the same as `*anyopaque` that the user likely meant. Is this OK? Relevant test case is `optional_pointer_to_void_in_extern_struct.zig`.


andrewrk:
I think it's OK. In fact, I was wondering if we should take it a step further and remove `anyopaque` now that `*void` functions almost identically. This would remove one strange new thing that Zig introduces from the language.

Vexu:
> I was wondering if we should take it a step further and remove `anyopaque` now that `*void` functions almost identically

It would be nice but I think it would also lead in to new strange things that don't currently exist. For example in stage2 dereferencing a runtime known `*anyopaque` gives the error `values of type 'anyopaque' must be comptime known, but operand value is runtime known` but dereferencing `*void` is allowed.

andrewrk:
Good point.

PR #12076, opened 2022-07-11
feat: `-q`/`--quiet` for quietening progress output

r00ster91:
https://user-images.githubusercontent.com/35064754/178306069-dd094047-a2b7-4dfe-a376-3b42ec2a7c3c.mp4

Here's why I think Zig might be attractive to language devs as a language to compile to:

- Zig compiles quickly (and will only get quicker in the future).
- Zig has multiple codegen backends.
- Zig has structs with unspecified field order for performance and other optimizations Zig facilitates that are **not** available in C.

And with this option Zig will be even more attractive as a language to compile to because now people can abstract `zig` being invoked under the hood to do the codegen for their language.

Of course, there are more reasons you might want to silence `zig`'s compile output. In fact, maybe as part of your workflow `zig` runs somewhere else where nobody sees it so by using `-q` you can silence the output that nobody sees anyway and possibly gain a few negligible milliseconds or so of improved compilation speed.


andrewrk:
Can you give an example where someone would want to use `-q` instead of `--color off`?

r00ster91:
(sorry for the wall of text, see TL;DR below)
Yeah, you're right, I kinda noticed that yesterday when I wrote that `comp.quiet or comp.color == .off` branch but it was late so I didn't think too much about it...
But maybe someone would want to use `-q` when they don't know that `--color off` does that too?
The documentation says "Enable or disable colored error messages", not disable progress output. The progress output is not an error message and it's never colored either. This is misleading.

Also:
- https://github.com/ziglang/zig/blob/master/src/Compilation.zig#L152-L154
- https://github.com/ziglang/zig/blob/master/src/Compilation.zig#L913-L914
- (and possibly other places)
Maybe I'm misunderstanding it and it just talks about this specific field to be removed but I thought maybe that meant `--color` was going to be removed? Maybe it could be removed anyway because I don't think anyone ever uses it, or should need it? Because the logic in https://github.com/ziglang/zig/blob/master/lib/std/fs/file.zig#L221-L242 looks sophisticated enough for me.
If you really don't want color, do `TERM=dumb`. I'm realizing that this of course applies to `-q` too so maybe we shouldn't add `-q`/`--quiet` and remove `--color` since `TERM=dumb zig run x.zig` behaves the same as `zig run -q x.zig` and `zig run --color off x.zig`. The reason this makes sense is because a dumb terminal means that it can't handle special characters and `std.Progress` uses escape sequences to reset the cursor and clear the line etc.
(`NO_COLOR=1` seems to do the same
thing).
Perhaps `--color auto`'s logic could be the default if `--color` is removed.

But then again, `NO_COLOR=1 zig-linux-x86_64-0.10.0-dev.2951+b9ed07227/zig run --color on x.zig` is kind of useful if maybe you're in an environment where that env var is always set somehow but for one specific invocation you do want color... but would that promote bad workflows? I think this might be a reason to keep `--color`, however (TL;DR):

1. I think `-q` is kind of conventional. One could expect a compiler to have this option, similarly to expecting `--color` to be an option.
2. I don't think `--color off` should turn off progress output. It's unexpected. The things should be separate.

So all in all I think *some* change is required here because `--color off` turning off progress output is unexpected in my opinion. The documentation is misleading too. `TERM=dumb` however makes sense.
To summarize, this whole thing is quite complicated actually. 

---

Apart from that, `--color off`/`NO_COLOR=1`/`TERM=dumb` seem to be used so rarely that they have some very obvious bugs actually...

```zig
pub fn main() void {
    var x = 1
}
```
```
zig run x.zig
```
```
x.zig:2:14: error: expected ';' after declaration
    var x = 1
             ^
```
(imagine it has color)
```
```
zig run --color off x.zig
```
x.zig:2:14: error: expected ';' after declaration
```
(why does it not print the caret and the line it points at? Printing the caret and the line of the error shouldn't need escape sequences...)

That's offtopic though and I will open an issue for that at some point. I think it'd be nice if dumb terminals can have the same output as other terminals, just without escape sequences.

r00ster91:
I fixed this PR:

* `--color off` no longer turns off progress output but turns off color.
* `TERM=dumb` turns off progress output and color.
* `NO_COLOR=` leaves progress output on but turns off color.

I think this is the expected behavior.

This also kind of goes in hand with #12989 after which this will look clearer.

r00ster91:
I can rebase and resolve the conflicts if my conclusion above is agreed to. Otherwise, feel free to close.

r00ster91:
So I'm in a situation right now where I would really like to turn off progress output but leave on color. I cannot find any way to do this. This PR seems to fix this.

Looking back at it, I still think we should remove `NO_COLOR`, `ZIG_DEBUG_COLOR`, and `--color auto`. Why make it so hard to choose? `--color off` to turn off only color and `--quiet` to turn off progress output makes perfect sense to me. And finally `TERM=dumb` would still disable both.

andrewrk:
Thanks for this patch. This is a bit tricky of a problem to sort out so I want to personally examine it carefully. I'm tackling this as part of #14647.

PR #12079, opened 2022-07-11
`std.Progress`: fix inaccurate line truncation and use optimal max terminal width

### LINKED ISSUE #12024: `std.Progress`: possible simplifications and ideas
After discovering `std.Progress` today I noticed that this is how it's resetting the terminal cursor
to the column where the cursor was before the line was written and how it's clearing that line: https://github.com/ziglang/zig/blob/master/lib/std/Progress.zig#L206-L254

All this could be greatly simplified by printing `'\r'` which in terminals always resets the cursor to column 0. This works on Windows too. So when you print `"hello world\rworld"`, it will print as `world world`!
As for the clearing of the previous content, let's say we have `hello world` on the terminal but now we want to show only `hello` on the same line. We would simply print `"\rhello"` + 6 more spaces to erase ` world` too.
All this would greatly simplify the code and avoid escape sequences altogether. It'd be much more portable and reliable. It would also no longer require any special casing for Windows. Performance and printing speed might improve too.

Now I would really like to implement this but the problem I have is that doing it this way will no longer allow printing multiple progresses on the same row or else it'll look messed up. This would no longer be possible:
```
Test [1/1] test "basic functionality"... [3/100] climbing towers [3/5]
```
so it'd have to be like this:
```
Test [1/1] test "basic functionality"...
[3/100] climbing towers [3/5]
```

I'd specifically be interested to hear an opinion from the original designer of `std.Progress`, @andrewrk.

Other things I would like to do include

* Adding an option to the compiler to disable the progress bar/any output during compilation (does this already exist?)
* Removing the space after the ellipsis (this would make the most sense when we go with my idea above of allowing only one progress per row in order to simplify the code using `'\r'`)
* Adding an actual loading bar-looking thing like this?
  ```
  [=====     ] [50/100] reticulating splines [5/5]
  ```
There are a lot of flashy CLI libs etc. out there that make use of Unicode characters and stuff to make the their loading bars look really cool (think [box-drawing characters](https://en.wikipedia.org/wiki/Box-drawing_character) etc.) but I think we should limit ourselves to ASCII characters here to make sure it works in as many environments as possible because not all Unicode characters are supported everywhere. Also, I don't think doing this provides any actual value so I think these fancy loading bars are probably best left to some external library and we should keep the current progress bar design.

r00ster91:
I really had a lot of attempts at this and I think this might be the best one so far. I think there's still some things to improve but this is definitely a start.

Output now fits nicely on pretty much any terminal width (well, not if it's >256 though, but that limit can easily be bumped):
image

Here's another common issue that I think Andrew described in https://github.com/ziglang/zig/issues/12024#issuecomment-1176803566:

image

That issue seems to be fixed now; here's a compilation on a terminal with a very small width:

https://user-images.githubusercontent.com/35064754/178335637-3a187e07-9cc9-4cc5-9ac3-45bda66ebc95.mp4

It succeeds with nothing left on the terminal.

Closes #12024

### DIFF
--- a/lib/std/Progress.zig
+++ b/lib/std/Progress.zig
@@ -290,8 +290,7 @@ fn refreshWithHeldLock(self: *Progress) void {
     var end: usize = 0;
     if (self.columns_written > 0) {
         // restore the cursor position by moving the cursor
-        // `columns_written` cells to the left, then clear the rest of the
-        // line
+        // `columns_written` cells to the left, then clear the rest of the line
         if (self.supports_ansi_escape_codes) {
             end += (std.fmt.bufPrint(self.output_buffer_slice[end..], "\x1b[{d}D", .{self.columns_written}) catch unreachable).len;
             end += (std.fmt.bufPrint(self.output_buffer_slice[end..], "\x1b[0K", .{}) catch unreachable).len;
@@ -343,9 +342,8 @@ fn refreshWithHeldLock(self: *Progress) void {
         self.columns_written = 0;
     }
 
-    // from here on we will print printable characters.
-    // make sure the unprintable characters we printed don't affect when we truncate the line
-    // in `bufWrite`.
+    // from here on we will write printable characters. we also make sure the unprintable characters
+    // we possibly wrote previously don't affect whether we truncate the line in `bufWrite`.
     const unprintables = end;
     end = 0;
     self.output_buffer_slice = self.output_buffer[unprintables .. unprintables + self.max_width.?];
--- a/lib/std/Progress.zig
+++ b/lib/std/Progress.zig
@@ -17,6 +17,7 @@ const windows = std.os.windows;
 const testing = std.testing;
 const assert = std.debug.assert;
 const os = std.os;
+const time = std.time;
 const Progress = @This();
 
 /// `null` if the current node (and its children) should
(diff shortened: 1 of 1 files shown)


lib/std/Progress.zig
```diff
@@ -164,11 +197,64 @@ pub fn start(self: *Progress, name: []const u8, estimated_total_items: usize) *N
...
+        const original_termios = try os.tcgetattr(file.handle);
+        var new_termios = original_termios;
+        new_termios.lflag &= ~(os.linux.ECHO | os.linux.ICANON);
+        try os.tcsetattr(file.handle, .NOW, new_termios);
+        defer os.tcsetattr(file.handle, .NOW, original_termios) catch {
+            // Sorry for ruining your terminal
+        };
+
+        try file.writeAll("\x1b[6n");
+        var buf: ["\x1b[256;256R".len]u8 = undefined;
```
daurnimator, line 240: These escapes are *not* portable (nor ascii as the doc-comment of the file suggests).
Getting these is why you need terminfo (rebutting https://github.com/ziglang/zig/issues/2931#issuecomment-1181882214)

r00ster91: It seems to be widely supported: https://terminalguide.namepad.de/seq/csi_sn-6/
Can you elaborate on how this escape sequence is not portable for terminal emulators for Linux?

r00ster91: > Terminals are more than just linux.

That's what the TODO comment above is for. This PR implements focuses on Linux and Windows.

> the terminal might be connecting over ssh/telnet/serial and you need to know the escapes for some other system

Sounds like a pretty advanced use case. This applies to a lot of cases (like simply invoking some `zig` command that causes colored output and such) and solving that seems to be out of scope of this PR. If you think this is an issue, you can open an issue for it but I'm guessing the majority of users won't face such issues.

andrewrk:
Thanks for the deep dive into this problem!

andrewrk:
This did not solve the problem of resizing the terminal window while progress is being printed:

image

PR #12081, opened 2022-07-11
std.os: add INVALID_SOCKET

yyny:
This constant returns an invalid `socket_t` that can be used as a sentinel value.

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -235,6 +235,7 @@ pub const RelativePathWasi = struct {
 };
 
 pub const socket_t = if (builtin.os.tag == .windows) windows.ws2_32.SOCKET else fd_t;
+pub const INVALID_SOCKET = if (builtin.os.tag == .windows) windows.ws2_32.INVALID_SOCKET else @as(socket_t, -1);
 
 /// See also `getenv`. Populated by startup code before main().
 /// TODO this is a footgun because the value will be undefined when using `zig build-lib`.


daurnimator:
I don't think this belongs here. `-1` is not universal for other operating systems.

IMO the name/concept really is windows specific, and hence should continue to live only there.

yyny:
> `-1` is not universal for other operating systems.

First of, I don't think this is even true, could you provide a counterexample? Second, any hypothetical socket implementations aside, Zig itself only defines a `socket` implementation for POSIX systems (And for Windows), and POSIX requires `-1` to be an invalid file descriptor (Any negative number, actually, as well as any value above `OPEN_MAX`, `-1` is just the value that is typically returned). Using `-1` will be perfectly fine for the foreseeable future.  
Moving the definition to the `system` files (like `os/linux.zig`) might be a good idea, though.

> Also consider that in Zig we frequently store file descriptors in unsigned integers.

Storing `INVALID_SOCKET` in an unsigned integer would be user error, it also contradicts your earlier concern, if all file descriptors can universally be assumed to be unsigned, then `-1` is by definition an invalid file descriptor value. It won't even compile anyway, since `INVALID_SOCKET` has type `socket_t`, which is signed. And ***even*** if it was unsigned, a different value could be chosen. I don't think this concern is relevant.

> IMO the name/concept really is windows specific, and hence should continue to live only there.

The concept is not Windows specific, POSIX also defines what constitutes an invalid socket/file descriptor value, and as mentioned above it is the only other socket implementation Zig currently supports.
This is exactly why I would like this change, the concept is universal, I need a value for it, and could use a portable constant for this in the standard library so I don't have to define it myself.

kprotty:
Are there any use cases of `-1` in the stdlib that could be replaced by `INVALID_SOCKET` before it gets merged?

arBmind:
> @kprotty Are there any use cases of `-1` in the stdlib that could be replaced by `INVALID_SOCKET` before it gets merged?

After a check of all relevant zig files in the std library, I found no use of -1 to initialize a socket_t. 

We have some places where we use -1 to initialize a fd_t, that someone might want to inspect deeper (possibly in a different issue)
* https://github.com/ziglang/zig/blob/cff5d9c805aa3433cc2562c3cb29bd3201817214/lib/std/child_process.zig#L258
* https://github.com/ziglang/zig/blob/cff5d9c805aa3433cc2562c3cb29bd3201817214/lib/std/child_process.zig#L262
* some calls to `std.os.mmap`
* and quite a lot in `io_uring.zig`

Maybe `io_uring.zig` is using the fd_t because for linux it's the same as socket_t? I have not enough experience with io_uring to decide this.

PR #12082, opened 2022-07-11
Add FreeBSD support to std.os.getFdPath

kimshrier:
The getFdPath function did not previously support FreeBSD.  This adds just enough functionality to get that done.  The inclusion of the kinfo_file struct does not contain all the members of a union that holds details about different file types as that was not necessary for getFdPath.  I put a place holder named kf_filler which occupies the space of the union and keeps the alignment for following structure members.

This is my first real attempt at writing zig code so if I have not done things using proper zig idioms, I welcome feedback.

### DIFF
--- a/lib/std/c/freebsd.zig
+++ b/lib/std/c/freebsd.zig
@@ -19,6 +19,8 @@ pub extern "c" fn pipe2(fds: *[2]fd_t, flags: u32) c_int;
 pub extern "c" fn posix_memalign(memptr: *?*anyopaque, alignment: usize, size: usize) c_int;
 pub extern "c" fn malloc_usable_size(?*const anyopaque) usize;
 
+pub extern "c" fn getpid() pid_t;
+
 pub const sf_hdtr = extern struct {
     headers: [*]const iovec_const,
     hdr_cnt: c_int,
@@ -397,6 +399,127 @@ pub const sockaddr = extern struct {
     };
 };
 
+pub const CAP_RIGHTS_VERSION = 0;
+
+pub const cap_rights = extern struct {
+    rights: [CAP_RIGHTS_VERSION + 2]u64,
+};
+
+pub const kinfo_file = extern struct {
+    /// Size of this record.
+    /// A zero value is for the sentinel record at the end of an array.
+    structsize: c_int,
+    /// Descriptor type.
+    @"type": c_int,
+    /// Array index.
+    fd: fd_t,
+    /// Reference count.
+    ref_count: c_int,
+    /// Flags.
+    flags: c_int,
+    // 64bit padding.
+    _pad0: c_int,
+    /// Seek location.
+    offset: i64,
+    un: extern union {
+        socket: extern struct {
+            /// Sendq size.
+            sendq: u32,
+            /// Socket domain.
+            domain: c_int,
+            /// Socket type.
+            @"type": c_int,
+            /// Socket protocol.
+            protocol: c_int,
+            /// Socket address.
+            address: sockaddr.storage,
+            /// Peer address.
+            peer: sockaddr.storage,
+            /// Address of so_pcb.
+            pcb: u64,
+            /// Address of inp_ppcb.
+            inpcb: u64,
+            /// Address of unp_conn.
+            unpconn: u64,
+            /// Send buffer state.
+            snd_sb_state: u16,
+            /// Receive buffer state.
+            rcv_sb_state: u16,
+            /// Recvq size.
+            recvq: u32,
+        },
+        file: extern struct {
+            /// Vnode type.
+            @"type": i32,
+            // Reserved for future use
+            _spare1: [3]i32,
+            _spare2: [30]u64,
+            /// Vnode filesystem id.
+            fsid: u64,
+            /// File device.
+            rdev: u64,
+            /// Global file id.
+            fileid: u64,
+            /// File size.
+            size: u64,
+            /// fsid compat for FreeBSD 11.
+            fsid_freebsd11: u32,
+            /// rdev compat for FreeBSD 11.
+            rdev_freebsd11: u32,
+            /// File mode.
+            mode: u16,
+            // 64bit padding.
+            _pad0: u16,
+            _pad1: u32,
+        },
+        sem: extern struct {
+            _spare0: [4]u32,
+            _spare1: [32]u64,
+            /// Semaphore value.
+            value: u32,
+            /// Semaphore mode.
+            mode: u16,
+        },
+        pipe: extern struct {
+            _spare1: [4]u32,
+            _spare2: [32]u64,
+            addr: u64,
+            peer: u64,
+            buffer_cnt: u32,
+            // 64bit padding.
+            kf_pipe_pad0: [3]u32,
+        },
+        proc: extern struct {
+            _spare1: [4]u32,
+            _spare2: [32]u64,
+            pid: pid_t,
+        },
+        eventfd: extern struct {
+            value: u64,
+            flags: u32,
+        },
+    },
+    /// Status flags.
+    status: u16,
+    // 32-bit alignment padding.
+    _pad1: u16,
+    // Reserved for future use.
+    _spare: c_int,
+    /// Capability rights.
+    cap_rights: cap_rights,
+    /// Reserved for future cap_rights
+    _cap_spare: u64,
+    /// Path to file, if any.
+    path: [PATH_MAX - 1:0]u8,
+};
+
+pub const KINFO_FILE_SIZE = 1392;
+
+comptime {
+    std.debug.assert(@sizeOf(kinfo_file) == KINFO_FILE_SIZE);
+    std.debug.assert(@alignOf(kinfo_file) == @sizeOf(u64));
+}
+
 pub const CTL = struct {
     pub const KERN = 1;
     pub const DEBUG = 5;
@@ -405,6 +528,7 @@ pub const CTL = struct {
 pub const KERN = struct {
     pub const PROC = 14; // struct: process entries
     pub const PROC_PATHNAME = 12; // path to executable
+    pub const PROC_FILEDESC = 33; // file descriptors for process
     pub const IOV_MAX = 35;
 };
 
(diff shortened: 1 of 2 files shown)


lib/std/os.zig
```diff
@@ -5215,6 +5215,60 @@ pub fn getFdPath(fd: fd_t, out_buffer: *[MAX_PATH_BYTES]u8) RealPathError![]u8 {
...
         },
+        .freebsd => {
+            var mib = [4]c_int{ CTL.KERN, KERN.PROC, KERN.PROC_FILEDESC, freebsd.getpid() };
+            var out_len: usize = 0;
+            try sysctl(&mib, null, &out_len, null, 0);
+
+            var kinfo_file: *freebsd.kinfo_file = undefined;
+
+            // allocate memory that is 64 bit alligned
+            var kinfo_file_buf = try std.heap.c_allocator.alloc(u64, (out_len / 8) + 1);
```
The-King-of-Toasters, line 5226: From my own testing, I found that `kinfo_file` is at most 1392 bytes. While big, I think it's possible to just stack allocate it here and eat the cost to avoid allocation.

If allocation is needed, I recommend using `alignedAlloc` instead of calculating the alignment ourselves.

lib/std/os.zig
```diff
@@ -5215,6 +5215,60 @@ pub fn getFdPath(fd: fd_t, out_buffer: *[MAX_PATH_BYTES]u8) RealPathError![]u8 {
...
+                sysctl(&mib, @ptrCast(*anyopaque, kinfo_file), &out_len, null, 0) catch |err| switch (err) {
+                    error.SystemResources => {
+                        // increase memory size and try again
+                        out_len = (out_len * 4) / 3;
+                        kinfo_file_buf = try std.heap.c_allocator.realloc(kinfo_file_buf, (out_len / 8) + 1);
+                        kinfo_file = @ptrCast(*freebsd.kinfo_file, kinfo_file_buf);
+                        continue;
+                    },
+
+                    else => return err,
```
The-King-of-Toasters, line 5241: Should be `else => |e| return e,` to remove `SystemResources` from the error set.

The-King-of-Toasters (changes requested):
Thank you for your contribution! As someone who has worked in this area (system ABI stuff), I've left a few comments.

lib/std/c/freebsd.zig
```diff
@@ -397,6 +399,27 @@ pub const sockaddr = extern struct {
     };
 };
 
+pub const cap_rights = struct {
+    cr_rights: [2]u64,
```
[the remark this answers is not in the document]
kimshrier: Ok.  That is good to know.  I was just preserving the existing names.  I'll make these kinds of changes.

lib/std/os.zig
```diff
@@ -5165,9 +5165,9 @@ pub fn realpathW(pathname: []const u16, out_buffer: *[MAX_PATH_BYTES]u8) RealPat
 
 /// Return canonical path of handle `fd`.
 /// This function is very host-specific and is not universally supported by all hosts.
-/// For example, while it generally works on Linux, macOS or Windows, it is unsupported
-/// on FreeBSD, or WASI.
-pub fn getFdPath(fd: fd_t, out_buffer: *[MAX_PATH_BYTES]u8) RealPathError![]u8 {
+/// For example, while it generally works on Linux, macOS, FreeBSD or Windows, it is
+/// unsupported on WASI.
+pub fn getFdPath(fd: fd_t, out_buffer: *[MAX_PATH_BYTES]u8) ![]u8 {
```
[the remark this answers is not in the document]
kimshrier: Since os.sysctl can return std.os.SysCtlError wouldn't we also need to combine that in with RealPathError?

lib/std/os.zig
kimshrier: Allocation is needed because what is returned by sysctl is a packed array of kinfo_file structures and we do not know ahead of time how many there will be.  That is why I make the original sysctl call with a null pointer for the buffer and an address of a length variable that is set to 0.  This sysctl call won't update the buffer but it will return the size of the packet array.  This lets us know how much memory to allocate.

There is a race condition where other threads in the process could open additional files so that on the second sysctl call, we could still run out of memory.  The sysctl code for KERN.PROC_FILEDESC will only write as many bytes into the buffer as are indicated by the length parameter so we don't need to worry about a buffer overflow.  It will also update the length value with the size of the data it would have put into the buffer if it had the space.  That is why on the second and subsequent calls to sysctl, I increase the buffer space to be 33% larger than the size of the previous failed call.  Hopefully, that will be more than enough memory and we won't have to try the sysctl function again.

This pattern of use mirrors the C code used in libprocstat which calls this sysctl function.

lib/std/os.zig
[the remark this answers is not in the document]
The-King-of-Toasters: Yes.

lib/std/os.zig
The-King-of-Toasters: I understand now. I remember bumping into a similar problem reading from `/dev/kstat` on Solaris. In that case, I have some suggestions.

- Document that the code reads from an array of `kinfo_file`s, and that the first sysctl is used to get the array count.
- Instead of allocating `u64`s, allocate an array of `kinfo_file`s and use e.g. `kinfos[0].ptr` as the base pointer for `sysctl`. This has a couple benefits:
    1. The initial allocation can just be `out_len` `kinfo_file`s.
    2. The resizing can be based on multiples of `@sizeOf(kinfo_file)` instead of the current division + multiply calculations.
    3. The final array can set by doing e.g. `files[0..out_len];`. Then you can loop over each member of the array with a for loop like `for (files) |f|{ if (f.fd == fd) ... }` instead of the current calls to `@intToPtr`.

That's it for now. I have some more suggestions but I'd like to play with this myself on my VM. If I made any mistakes in my logic here, let me know.

The-King-of-Toasters:
I'm afraid I got so nerd sniped by this that I decided to implement the changes on my own, and I'm currently running the stdlib tests on my VM to check. I don't want you to feel like I'm stealing your thunder here, so if you'd like, you could give me write access to your fork so I can rebase+push my changes.

kimshrier:
I've sent you an invite to my repository.  I look forward to seeing your changes.

I don't feel that you are stealing my thunder.  Rather, I think of it as preventing me from committing some grave zig faux pas.

lib/std/os.zig
```diff
@@ -5215,6 +5222,49 @@ pub fn getFdPath(fd: fd_t, out_buffer: *[MAX_PATH_BYTES]u8) RealPathError![]u8 {
...
+        .freebsd => {
+            const allocator = std.heap.c_allocator;
+            const mib = [4]c_int{ CTL.KERN, KERN.PROC, KERN.PROC_FILEDESC, system.getpid() };
+
+            // Get an initial size for the kinfo buffer and allocate.
+            // The buffer is used to store the varible-sized kinfo_file structures.
+            // Note that all allocations are 64-bit aligned, as is kinfo_file.
+            var len: usize = 0;
+            try sysctl(&mib, null, &len, null, 0);
+            var buf = try allocator.alignedAlloc(u8, 8, len);
```
kimshrier, line 5234: Since we are allocating u8 instead of u64, we do not need to do the (len / 8) + 1 calculation anymore.

lib/std/os.zig
```diff
@@ -5215,6 +5222,49 @@ pub fn getFdPath(fd: fd_t, out_buffer: *[MAX_PATH_BYTES]u8) RealPathError![]u8 {
...
+            defer allocator.free(buf);
+
+            // Fill the buffer with data.
+            // The buffer is reallocated in a loop since the number of open files may
+            // change between each sysctl call.
+            while (true) {
+                sysctl(&mib, @ptrCast(*anyopaque, buf), &len, null, 0) catch |err| switch (err) {
+                    error.SystemResources => {
+                        // increase memory size and try again
+                        buf = try allocator.reallocAdvanced(buf, 8, len, .exact);
```
kimshrier, line 5244: Same length calculation as above.

lib/std/os.zig
```diff
@@ -5215,6 +5222,49 @@ pub fn getFdPath(fd: fd_t, out_buffer: *[MAX_PATH_BYTES]u8) RealPathError![]u8 {
             };
             return target;
         },
+        .freebsd => {
+            const allocator = std.heap.c_allocator;
```
squeek502, line 5226: Unless I'm misunderstanding something:
- `c_allocator` is only available when linking libc
- Hard coding an allocator within a function like this goes against Zig's conventions on how allocators are used (it should be provided as a parameter instead)

kimshrier: Unless I am misunderstanding something (highly likely), calling sysctl implies linking libc.

If that is not the case, adding an allocator parameter would break compatibility with all existing code that calls this function.  Since it is a public function, I would think that would be undesirable.

At the moment, only the FreeBSD specific code needs an allocator.  I do not want to put in code that does not adhere to Zig's conventions.  Is it better to hard code an allocator and not break compatibility or is it better to add an allocator parameter?  Or, is there another option?

squeek502: If linking libc is implied then ignore that bullet point.

> Is it better to hard code an allocator and not break compatibility

AFAIK there's nowhere in the Zig standard library that does this currently. Going from a non-allocating API to an allocating one is always a breaking change, and for cross-platform functions it "infects" the cross-platform API (see things like `std.process.args` and `std.process.argsWIthAllocator` where `argsWithAllocator` is the cross-platform API even though the allocator is only needed on Windows and WASI).

squeek502: Actually, I have found a function in the standard library that does hard-code the `c_allocator`, so I might be wrong:

https://github.com/ziglang/zig/blob/1653a9b2597c66cbcc88ea75d8a4b88c163584a5/lib/std/Thread.zig#L654-L656

(also worth noting that it's probably good to take my input with a grain of salt; it might be worth waiting for someone with more knowledge about this particular implementation to weigh in)

kimshrier: I believe that that particular line in Thread.zig is what led me to hard code the allocator in my code.  It does appear that it is the only place where that happens.  I can see arguments for not hard coding the allocator and it is something I would think people should avoid.

I'll see if any more people weigh in on this.  Having 2 flavors of a function, 1 with and 1 without an allocator, seems clunky to me but I am willing to follow wiser counsel on this.

The-King-of-Toasters:
Since Zig links to `libutil` on FreeBSD, we could just use [`kinfo_getfile`](https://man.freebsd.org/kinfo_getfile(3)). As Kim mentioned it essentially does the same thing we do (the one in `libprocstat` is copied from `libutil`). However, we still need to handle allocation/sysctl errors and free the returned array with the `c_allocator`.

The-King-of-Toasters:
So from reading [this patch to add `F_GETPATH` to FreeBSD](https://reviews.freebsd.org/D33555?id=100323), it seems that the `fcntl` command `F_KINFO` may provide what we need without allocation. I'll test this and report if it works.

The-King-of-Toasters (approved):

kimshrier:
I've tested the updated code and it is working.  I'm good with committing this.

Issue #12062, opened 2022-07-09
upgrade to mingw-w64 v10.0.0


andrewrk:
Zig currently has mingw-w64 v9.0.0. Upgrade to the newest release which is v10.0.0.

previous issue: #8867

```
Notable changes:
* New msvcrt10, 20, 40, 70 and 71 import libraries.
* Drop x86_64 64bit crtdll.dll
* Updated wine imports
* Fix intrin.h compatibility with GCC-11
* _(v)scprintf optimization
* optimize __ms_vsnprintf for LTO
```

[Release Notes](https://sourceforge.net/p/mingw-w64/mailman/message/37635065/)

PR #12087, opened 2022-07-12
upgrade to mingw-w64 v10.0.0

nektro:
upgrade also included 1901 new files (listed here https://clbin.com/6im5d). left those out since it seemed like a lot and was unfamiliar with the exact procedure. Wiki section on upgrading CRT section was pretty vague

closes https://github.com/ziglang/zig/issues/12062


andrewrk:
Thanks for helping with the upgrade!

Unfortunately, it's not really possible to review this PR without redoing the work myself. (Plus I'll need to do that regardless in order to troubleshoot the CI failures)

Allar:
I'm trying to run zig build with a x86_64-windows-gnu (msvc as well, but ran into an issue with exceptions abi), and after some code massaging I ran into needing to be able to support pthread.h (external libs handle msvc differently).  I'm super interested in upgrading to mingw and adding support for pthreads, as this seems an easier obstacle to tackle than trying to target msvc with the obstacle of some libcxxabi stuff I don't understand.

I have zig build working with my codebase with x86_64 and arm64 linux targets.

If anyone could provide any insight on how you upgraded mingw and the right way to do this, I'd love to hear it.

I've been working from https://github.com/ziglang/zig/wiki/Updating-libc#mingw-w64-windows in order to both upgrade mingw to v10 as well as support pthreads.

I've built the headers, and updated musl/*, both according to that wiki and by using some build scripts I modified to more match the info here in this wiki.

Zig appears to compile through my codebase, but now I'm up against two errors when compiling the libc/mingw/crt

1.
```
error(compilation): clang failed with stderr: /home/allar/miko/zig/lib/libc/mingw/crt/wildcard.c:25:10: fatal error: 'config.h' file not found
```

config.h is trying to be included because -DHAVE_CONFIG_H is appearing on the commandline via `--verbose-cc` on wildcard.c, which is happening here: https://github.com/ziglang/zig/blob/cc39d453c45645296cd771c94d97b8a77469be91/src/mingw.zig#L74

2.
A whole lot of 
```
error(compilation): clang failed with stderr: In file included from /home/allar/miko/zig/lib/libc/mingw/secapi/_cgets_s.c:4:
/home/allar/miko/zig/lib/libc/mingw/include/msvcrt.h:4:2: error: "This file should only be used in libmsvcrt-os.a"
```
It looks like the guard for this that was `__LIBMSVCRT__` is now `__LIBMSVCRT_OS__`, which I would assume to mean https://github.com/ziglang/zig/blob/cc39d453c45645296cd771c94d97b8a77469be91/src/mingw.zig#L112 to be updated.

I'd like to understand what determines mingw's `buildCRTFile` getting passed `mingw32_lib` or `msvcrt_os_lib` as well, as I'm taking a wild stab in the dark that if i can switch to msvcrt_os_lib and then replace that define, I might get further along in trying to build some external libraries that rely on pthread, for windows, on linux.

PR #12094, opened 2022-07-12
stage2 LLVM: Use non-packed struct for unions and arrays

### LINKED ISSUE #12086: many std.json test failures on powerpc
A lot of tests are failing, but if we just look at one:

```
test "n_object_bracket_key" {
    try err(
        \\{[: "x"}
    );
}
```

```
[nix-shell:~/Downloads/zig/build-release]$ stage2/bin/zig test ../lib/std/std.zig -target powerpc-linux-none --test-cmd qemu-ppc --test-cmd-bin --test-filter 'n_object_bracket_key'
114/230 test.n_object_bracket_key... error: the following test command crashed:
qemu-ppc ../lib/std/zig-cache/o/57c79ddf3440ff907412143c7c741c18/test /home/andy/Downloads/zig/build-release/stage2/bin/zig
```

```
Program received signal SIGSEGV, Segmentation fault.
0x100de088 in json.test.err () at /home/andy/Downloads/zig/lib/std/json/test.zig:45
45	    try testing.expect(std.meta.isError(testNonStreaming(s)));
(gdb) 
(gdb) disas /s
44	
45	    try testing.expect(std.meta.isError(testNonStreaming(s)));
   0x100de070 <+140>:	lwz     r3,36(r31)
   0x100de074 <+144>:	lwz     r5,28(r31)
   0x100de078 <+148>:	lwz     r4,32(r31)
   0x100de07c <+152>:	bl      0x100de284 <json.test.testNonStreaming>
   0x100de080 <+156>:	bl      0x100de394 <meta.isError__anon_36012>
   0x100de084 <+160>:	mr      r4,r3
=> 0x100de088 <+164>:	lwz     r3,36(r31)
   0x100de08c <+168>:	bl      0x100d1784 <testing.expect>
End of assembler dump.
(gdb) 
```

I looked into this for a bit and it is some kind of low level code-generation problem. The segfault is happening apparently when accessing the return value (of type `!void`) of a function call.

The next step here is to create a reduction, perhaps at the LLVM IR level, and investigate whether this is a Zig bug or an LLVM bug.

topolarity:
Resolves #12086

Consider this lowered union type:
```llvm
%Value = type { <{ i64, [8 x i8] }>, i1, [7 x i8] } ; 24 bytes, align(1)
%ErrorUnion = type { %Value, i16 } ; 26 bytes, align(2)
```

Stage 2 incorrectly assumes that LLVM's `sizeof(%Value)` is 32 bytes, but it's actually 26 because the alignment of Value was "deleted" by the use of a packed struct. Stage 2 can then end up generating a 32-byte `memcpy` that exceeds the bounds of the corresponding `alloca`.

AFAICT, there are two ways to go about fixing this:
   1. Make sure our LLVM type lowering never depends on the LLVM type being correctly aligned, OR
   2. Make sure that we keep the alignment of LLVM types in sync with the Zig types they represent

This PR takes approach (2), mostly for simplicity. AFAICT, we don't need a packed struct to achieve the desired layout, since an appended `[N]u8` always acts as contiguous padding. So this simply changes the type to use a non-packed struct.

Along the way, I noticed that arrays were also using packed structs. I could be wrong here, but this seems incorrect since array elements should be self-aligned.

### DIFF
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -2689,7 +2689,7 @@ pub const DeclGen = struct {
                         llvm_aligned_field_ty,
                         dg.context.intType(8).arrayType(padding_len),
                     };
-                    break :t dg.context.structType(&fields, fields.len, .True);
+                    break :t dg.context.structType(&fields, fields.len, .False);
                 };
 
                 if (layout.tag_size == 0) {
@@ -3002,7 +3002,7 @@ pub const DeclGen = struct {
                         return dg.context.constStruct(
                             llvm_elems.ptr,
                             @intCast(c_uint, llvm_elems.len),
-                            .True,
+                            .False,
                         );
                     } else {
                         const llvm_elem_ty = try dg.lowerType(elem_ty);
@@ -3039,7 +3039,7 @@ pub const DeclGen = struct {
                         return dg.context.constStruct(
                             llvm_elems.ptr,
                             @intCast(c_uint, llvm_elems.len),
-                            .True,
+                            .False,
                         );
                     } else {
                         const llvm_elem_ty = try dg.lowerType(elem_ty);
@@ -3056,7 +3056,7 @@ pub const DeclGen = struct {
                     const llvm_elems: [1]*const llvm.Value = .{sentinel};
                     const need_unnamed = dg.isUnnamedType(elem_ty, llvm_elems[0]);
                     if (need_unnamed) {
-                        return dg.context.constStruct(&llvm_elems, llvm_elems.len, .True);
+                        return dg.context.constStruct(&llvm_elems, llvm_elems.len, .False);
                     } else {
                         const llvm_elem_ty = try dg.lowerType(elem_ty);
                         return llvm_elem_ty.constArray(&llvm_elems, llvm_elems.len);
@@ -3343,7 +3343,7 @@ pub const DeclGen = struct {
                     const fields: [2]*const llvm.Value = .{
                         field, dg.context.intType(8).arrayType(padding_len).getUndef(),
                     };
-                    break :p dg.context.constStruct(&fields, fields.len, .True);
+                    break :p dg.context.constStruct(&fields, fields.len, .False);
                 };
 
                 if (layout.tag_size == 0) {


topolarity:
Lots of room for incorrect assumptions here, so let me know if this is the wrong way to fix this.

topolarity:
Hmm, adding some assertions to `lowerType` reveals many other discrepancies:
```
error(codegen): alignment mismatch detected when lowering type 'u293' (expected 16, found 8)
error(codegen): alignment mismatch detected when lowering type 'u128' (expected 16, found 8)
error(codegen): size mismatch detected when lowering type 'anyerror!u128' (expected 32, found 24)
error(codegen): size mismatch detected when lowering type '?u128' (expected 32, found 24)
error(codegen): alignment mismatch detected when lowering type 'tuple{u128, u1}' (expected 16, found 8)
error(codegen): alignment mismatch detected when lowering type '(struct decl=Module.Decl.Index(20827))' (expected 16, found 8)
error(codegen): alignment mismatch detected when lowering type 'behavior.struct.FooArray24Bits' (expected 16, found 8)
error(codegen): alignment mismatch detected when lowering type 'u96' (expected 16, found 8)
error(codegen): alignment mismatch detected when lowering type 'behavior.struct.test.packed struct with fp fields.S' (expected 16, found 8)
error(codegen): size mismatch detected when lowering type '@Vector(4, u1)' (expected 4, found 1)
error(codegen): size mismatch detected when lowering type '@Vector(6, i32)' (expected 24, found 32)
error(codegen): size mismatch detected when lowering type '@Vector(6, i32)' (expected 24, found 32)
error(codegen): size mismatch detected when lowering type 'error{}!behavior.error.test.error union payload is properly aligned.S' (expected 64, found 56)
error(codegen): alignment mismatch detected when lowering type 'error{}!behavior.error.test.error union payload is properly aligned.S' (expected 16, found 8)
error(codegen): size mismatch detected when lowering type '?behavior.bugs.3112.State' (expected 16, found 1)
error(codegen): alignment mismatch detected when lowering type '?behavior.bugs.3112.State' (expected 8, found 1)
error(codegen): size mismatch detected when lowering type 'behavior.bugs.394.E' (expected 16, found 24)
error(codegen): size mismatch detected when lowering type 'behavior.bugs.394.S' (expected 24, found 32)
error(codegen): size mismatch detected when lowering type 'behavior.bugs.394.S' (expected 24, found 32)
error(codegen): alignment mismatch detected when lowering type 'u128' (expected 16, found 8)
...
```

The size mismatches almost certainly need to be tightened up, but some of the alignment mismatches may be unavoidable (for example, LLVM never uses a natural alignment larger than the word size).

Perhaps this is the wrong approach after all, and we need to instead manually enforce all of our alignments.

andrewrk:
This looks like a step in the right direction. Thanks!

> Hmm, adding some checks to `lowerType` reveals many other discrepancies:

Nice find. This could very well be the cause of #11450. Perhaps in debug builds of the compiler we should enable an assertion that makes sure that we successfully produce an LLVM type that has an ABI size equal to what zig expects. Of course we would need to fix all these discrepancies first.

As for alignment, I don't think it actually matters for correctness since we always insert an alignment attribute on allocas, loads, and stores.

PR #12102, opened 2022-07-13
std.ArrayList.ensureTotalCapacity: check for integer overflow

### LINKED ISSUE #12099: ArrayList.ensureTotalCapacity() can invoke Undefined Behavior on valid input
### Zig Version

0.9.1

### Steps to Reproduce

```zig
const std = @import("std");

pub fn main() !void {
    var list = std.ArrayList(u8).init(std.testing.allocator);
    try list.ensureTotalCapacity(std.math.maxInt(usize));
    std.debug.print("{d}\n", .{list.capacity});
}
```

### Expected Behavior

Return an `Allocator.Error` (`OutOfMemory`) ~like `ensureTotalCapacityPrecise()`.~

### Actual Behavior

Panics in debug and safe builds, infinite loop in fast builds.

This is due to an integer overflow here: https://github.com/ziglang/zig/blob/f89ace4d04f8567ae92dce74ffb0e0d416077492/lib/std/array_list.zig#L313

I haven't looked at other container types in zig's standard library. It may be worth checking them for similar problems, too.

alichraghi:
Fixes #12099

### DIFF
--- a/lib/std/array_list.zig
+++ b/lib/std/array_list.zig
@@ -3,6 +3,7 @@ const debug = std.debug;
 const assert = debug.assert;
 const testing = std.testing;
 const mem = std.mem;
+const math = std.math;
 const Allocator = mem.Allocator;
 
 /// A contiguous, growable list of items in memory.
@@ -306,17 +307,17 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
         /// Invalidates pointers if additional memory is needed.
         pub fn ensureTotalCapacity(self: *Self, new_capacity: usize) Allocator.Error!void {
             if (@sizeOf(T) > 0) {
-                var better_capacity = self.capacity;
-                if (better_capacity >= new_capacity) return;
+                if (self.capacity >= new_capacity) return;
 
+                var better_capacity = self.capacity;
                 while (true) {
-                    better_capacity += better_capacity / 2 + 8;
+                    better_capacity +|= better_capacity / 2 + 8;
                     if (better_capacity >= new_capacity) break;
                 }
 
                 return self.ensureTotalCapacityPrecise(better_capacity);
             } else {
-                self.capacity = std.math.maxInt(usize);
+                self.capacity = math.maxInt(usize);
             }
         }
 
@@ -333,7 +334,7 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
                 self.items.ptr = new_memory.ptr;
                 self.capacity = new_memory.len;
             } else {
-                self.capacity = std.math.maxInt(usize);
+                self.capacity = math.maxInt(usize);
             }
         }
 
@@ -678,11 +679,11 @@ pub fn ArrayListAlignedUnmanaged(comptime T: type, comptime alignment: ?u29) typ
         /// Modify the array so that it can hold at least `new_capacity` items.
         /// Invalidates pointers if additional memory is needed.
         pub fn ensureTotalCapacity(self: *Self, allocator: Allocator, new_capacity: usize) Allocator.Error!void {
-            var better_capacity = self.capacity;
-            if (better_capacity >= new_capacity) return;
+            if (self.capacity >= new_capacity) return;
 
+            var better_capacity = self.capacity;
             while (true) {
-                better_capacity += better_capacity / 2 + 8;
+                better_capacity +|= better_capacity / 2 + 8;
                 if (better_capacity >= new_capacity) break;
             }
 


Frechdachs (commented):
Wouldn't this lead to `ensureTotalCapacityPrecise()` be called with a `better_capacity` that's way too low? So it would just hide the problem. `better_capacity` only holds the overflowed bits.

An alternative would be:
```zig
while (true) {
    const overflowed = @addWithOverflow(usize, better_capacity, better_capacity / 2 + 8, &better_capacity);
    if (overflowed) better_capacity = std.math.maxInt(usize);
}
```

But this also wouldn't solve the problem, because there also is another type of Integer Overflow in `ensureTotalCapacityPrecise()`.
Also, there is another `ensureTotalCapacity()` for the unmanaged ArrayList, it would have to be changed there, too.

alichraghi:
No, result is the same.
`@addWithOverflow` langref:
> Performs `result.* = a + b`. If overflow or underflow occurs, stores the overflowed bits in result and returns `true`. If no overflow or underflow occurs, returns `false`.

so after overflow, `better_capacity` is equal to `std.math.maxInt(usize)`

Frechdachs:
The reference you quoted says that only the overflowed bits are stored.
```zig
const std = @import("std");

pub fn main() !void {
    var i: usize = std.math.maxInt(usize);
    _ = @addWithOverflow(usize, i, 1, &i);
    std.debug.print("{d}\n", .{i});
}
```
This prints `0`.

squeek502:
I think what you want is [saturating arithmetic](https://ziglang.org/documentation/master/#Table-of-Operators):

```zig
better_capacity +|= better_capacity / 2 + 8;
```

Doesn't solve the integer overflow in Allocator, but would effectively 'cap' the capacity to usize max:

```zig
const std = @import("std");

fn ensureTotalCapacity(current_capacity: usize, new_capacity: usize) usize {
    var better_capacity = current_capacity;
    if (better_capacity >= new_capacity) return better_capacity;

    while (true) {
        better_capacity +|= better_capacity / 2 + 8;
    }

    return better_capacity;
}

test {
    var i: usize = 0;
    while (i < 100) : (i += 1) {
        const max_capacity = std.math.maxInt(usize);
        const current_capacity = max_capacity - i;
        const capacity = ensureTotalCapacity(current_capacity, max_capacity);
        try std.testing.expectEqual(@as(usize, max_capacity), capacity);
    }
}
```

alichraghi:
this fixes the first issue. an integer overflow still remain in allocation
```diff
pub fn ensureTotalCapacity(self: *Self, new_capacity: usize) Allocator.Error!void {
    if (@sizeOf(T) > 0) {
--    var better_capacity = self.capacity;
--    if (better_capacity >= new_capacity) return;
++    if (self.capacity >= new_capacity) return;

--    while (true) {
--      better_capacity += better_capacity / 2 + 8;
--      if (better_capacity >= new_capacity) break;
--    }
++    const better_capacity = std.math.max3(8, self.capacity * 2, new_capacity);
        return self.ensureTotalCapacityPrecise(better_capacity);
    } else {
        self.capacity = std.math.maxInt(usize);
    }
}
```

squeek502:
> this fixes the first issue.

No, it wouldn't. `self.capacity` being larger than `std.math.maxInt(usize) / 2` would overflow on the `self.capacity * 2`:

```zig
fn ensureTotalCapacity(current_capacity: usize, new_capacity: usize) usize {
    if (current_capacity >= new_capacity) return current_capacity;

    return std.math.max3(8, current_capacity * 2, new_capacity);
}

test "overflow on multiply" {
    const max_capacity = std.math.maxInt(usize);
    const current_capacity = max_capacity / 2 + 1;
    const capacity = ensureTotalCapacity(current_capacity, max_capacity);
    try std.testing.expectEqual(@as(usize, max_capacity), capacity);
}
```

```
Test [1/1] test "overflow on multiply"... thread 1204741 panic: integer overflow
/home/ryan/Programming/zig/tmp/ensurecapmax.zig:18:46: 0x207fe7 in ensureTotalCapacity (test)
    return std.math.max3(8, current_capacity * 2, new_capacity);
                                             ^
```

The saturating arithmetic I mentioned above is a working solution.

alichraghi:
@squeek502 this is an edge case for <=32bit platforms. in that case, rust returns `CapacityOverflow` error: https://doc.rust-lang.org/src/alloc/raw_vec.rs.html#505-507

i think we should do the same. since this isn't OOM

squeek502:
I think it makes sense to treat the allocator overflow as a separate issue--it's not specific to `ArrayList` and would need mitigations either within the Allocator interface or in a *lot* of different places throughout the standard library. My suggestion would be to use this PR to only fix the overflow in `ArrayList.ensureTotalCapacity` and create a new issue for the Allocator stuff.

alichraghi:
the new way also did an optimization over the current and saturating arithmetic.

### Appending 100000000 usize items

**New way**
```zig
const std = @import("std");
const allocator = std.testing.testing_allocator;

pub fn ensureTotalCapacity(self: *std.ArrayList(usize), new_capacity: usize) !void {
    if (@sizeOf(usize) > 0) {
        if (self.capacity >= new_capacity) return;

        const better_capacity = std.math.max3(8, self.capacity * 2, new_capacity);
        return self.ensureTotalCapacityPrecise(better_capacity);
    } else {
        self.capacity = std.math.maxInt(usize);
    }
}

pub fn main() anyerror!void {
    var list = std.ArrayList(usize).init(std.testing.allocator);
    defer list.deinit();
    var i: usize = 1;
    while (i < 100000000) : (i += 1) {
        try ensureTotalCapacity(&list, i);
        try list.append(i);
    }
}
```

**Saturating arithmetic**
```zig
const std = @import("std");
const allocator = std.testing.testing_allocator;

pub fn ensureTotalCapacity(self: *std.ArrayList(usize), new_capacity: usize) !void {
    if (@sizeOf(usize) > 0) {
        var better_capacity = self.capacity;
        if (better_capacity >= new_capacity) return;

        while (true) {
            better_capacity +|= better_capacity / 2 + 8;
        }
        return self.ensureTotalCapacityPrecise(better_capacity);
    } else {
        self.capacity = std.math.maxInt(usize);
    }
}

pub fn main() anyerror!void {
    var list = std.ArrayList(usize).init(std.testing.allocator);
    defer list.deinit();
    var i: usize = 1;
    while (i < 100000000) : (i += 1) {
        try ensureTotalCapacity(&list, i);
        try list.append(i);
    }
}
```

**Result**
```zig
Benchmark 1: ./new_way
  Time (mean ± σ):      1.457 s ±  0.325 s    [User: 0.635 s, System: 0.809 s]
  Range (min … max):    1.234 s …  2.287 s    10 runs
 
Benchmark 2: ./saturating_arithmetic
  Time (mean ± σ):      1.887 s ±  0.136 s    [User: 0.865 s, System: 1.015 s]
  Range (min … max):    1.701 s …  2.193 s    10 runs
 
Summary
  './new_way' ran
    1.30 ± 0.30 times faster than './saturating_arithmetic'
```

### Appending 200000 large array `[1024]usize` items

**New way**
```zig
const std = @import("std");
const allocator = std.testing.testing_allocator;

pub fn ensureTotalCapacity(self: *std.ArrayList([1024]usize), new_capacity: usize) !void {
    if (@sizeOf([1024]usize) > 0) {
        if (self.capacity >= new_capacity) return;

        const better_capacity = std.math.max3(8, self.capacity * 2, new_capacity);
        return self.ensureTotalCapacityPrecise(better_capacity);
    } else {
        self.capacity = std.math.maxInt(usize);
    }
}

pub fn main() anyerror!void {
    var list = std.ArrayList([1024]usize).init(std.testing.allocator);
    defer list.deinit();
    var i: usize = 1;
    while (i < 200000) : (i += 1) {
        try ensureTotalCapacity(&list, i);
        try list.append(std.mem.zeroes([1024]usize));
    }
}
```

**Saturating arithmetic**
```zig
const std = @import("std");
const allocator = std.testing.testing_allocator;

pub fn ensureTotalCapacity(self: *std.ArrayList([1024]usize), new_capacity: usize) !void {
    if (@sizeOf([1024]usize) > 0) {
        var better_capacity = self.capacity;
        if (better_capacity >= new_capacity) return;

        while (true) {
            better_capacity +|= better_capacity / 2 + 8;
        }
        return self.ensureTotalCapacityPrecise(better_capacity);
    } else {
        self.capacity = std.math.maxInt(usize);
    }
}

pub fn main() anyerror!void {
    var list = std.ArrayList([1024]usize).init(std.testing.allocator);
    defer list.deinit();
    var i: usize = 1;
    while (i < 200000) : (i += 1) {
        try ensureTotalCapacity(&list, i);
        try list.append(std.mem.zeroes([1024]usize));
    }
}
```

**Result**
```zig
Benchmark 1: ./new_way
  Time (mean ± σ):      3.312 s ±  0.463 s    [User: 1.276 s, System: 2.021 s]
  Range (min … max):    3.098 s …  4.626 s    10 runs
 
Benchmark 2: ./saturating_arithmetic
  Time (mean ± σ):      1.954 s ±  0.046 s    [User: 0.717 s, System: 1.233 s]
  Range (min … max):    1.922 s …  2.068 s    10 runs
 
Summary
  './new_way' ran
    1.70 ± 0.24 times faster than './saturating_arithmetic'
```

lib/std/array_list.zig
```diff
@@ -306,17 +307,10 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
...
-                if (better_capacity >= new_capacity) return;
-
-                while (true) {
-                    better_capacity += better_capacity / 2 + 8;
-                    if (better_capacity >= new_capacity) break;
-                }
-
-                return self.ensureTotalCapacityPrecise(better_capacity);
+                if (self.capacity >= new_capacity) return;
+                return self.ensureTotalCapacityPrecise(math.max(8, new_capacity));
```
squeek502, line 311: Removing the `better_capacity` calculation here is not an acceptable change. The point of `better_capacity` is make the ArrayList grow such that the number of allocations is minimized (i.e. calling `ensureTotalCapacity` in a loop won't cause a reallocation on every call).

Here's how capacity grew before:

```
current capacity => resulting capacity
1...8 => 8
9...20 => 20
21...38 => 38
39...65 => 65
66...105 => 105
106...165 => 165
166...255 => 255
256...390 => 390
391...593 => 593
594...897 => 897
898...1353 => 1353
1354...2037 => 2037
2038...3063 => 4096
3064...6911 => 8192
6912...10374 => 12288
10375...15569 => 16384
15570...23361 => 24576
23362...35049 => 36864
35050...52581 => 53248
52582...78879 => 81920
78880...118326 => 118784
118327...177497 => 180224
177498...266253 => 270336
```

Any changes to this code should keep the same growth pattern or have benchmarks for why the new growth pattern is better in terms of real use cases.

alichraghi: > The point of better_capacity is make the ArrayList grow such that the number of reallocations is minimized

i totally missed this but cannot understand why allocating the exact size (new way) is faster

squeek502: Ah, sorry, didn't notice that you edited the comment with the benchmarks. That is indeed interesting, I'd like to investigate that a bit more when I get a chance.

alichraghi: forgot benchmark. i replaced `append` with `appendAssumeCapacity` and now, `new_way` looks super slow. i still wonder how it were faster, even with double calling it. at the end i think we'll go with **Saturating Arithmetic**

alichraghi: an important thing. i made a mistake when comparing the binaries. and i thought this was even better so edited the real `new_way` and replaced with this (which is super slow):
```zig
return self.ensureTotalCapacityPrecise(std.math.max(8, new_capacity));
```
this is the real `new_way`:
```
return self.ensureTotalCapacityPrecise(std.math.max3(8, self.capacity * 2, new_capacity));
```

squeek502: I think you have to be careful how you construct the benchmarks for something like this. Your current benchmark code will always favor anything that grows the fastest (and your new_way grows faster than the current implementation). In fact, the `better_capacity` calculation that would perform the fastest in your benchmark would just be:

```zig
better_capacity = 100_000_000
```

since that would just do one allocation already at the final capacity requested in the benchmark, but it's obviously a very bad `better_capacity` calculation for the average case. In general, `better_capacity` is a tradeoff between:
- Number of reallocations needed
- Amount of memory allocated that is never actually used

I think the ideal is that `better_capacity` is tuned so that both are as low as possible in "real" usage.

So, I think some more sophisticated benchmarking would be necessary to evaluate different `better_capacity` algorithms (maybe something like [the self-hosted-parser benchmarks from gotta-go-fast](https://github.com/ziglang/gotta-go-fast/tree/master/benchmarks/self-hosted-parser) which makes heavy use of `ensureTotalCapacity`).

I think I'd again recommend restricting this PR to just solving the original integer overflow bug, and then creating a new issue/PR focused on the performance of the `better_capacity` algorithm.

PR #12105, opened 2022-07-13
CMake: Use `NAMES_PER_DIR` in all lookups

topolarity:
CMake has a surprising default behavior where looking up a library by multiple names gives the name order higher priority than the directory search order. This change add `NAMES_PER_DIR` to request the more sensible behavior: directory search order has higher priority than name order.

Previously, system-provided "llvm-config-14" would always be chosen over "llvm-config" in CMAKE_PREFIX_PATH.

Now, any matching name found in CMAKE_PREFIX_PATH wins over system-provided tools/libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,7 @@ set(ZIG_VERSION "" CACHE STRING "Override Zig version string. Default is to find
 
 if("${ZIG_VERSION}" STREQUAL "")
     set(ZIG_VERSION "${ZIG_VERSION_MAJOR}.${ZIG_VERSION_MINOR}.${ZIG_VERSION_PATCH}")
-    find_program(GIT_EXE NAMES git)
+    find_program(GIT_EXE NAMES git NAMES_PER_DIR)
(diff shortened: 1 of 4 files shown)


topolarity:
Oops, I just realized there's another problem with this.

It will fail if we encounter `llvm-config` (of an incompatible LLVM version) in an earlier search directory. Looks like we might have to iterate the search directories ourselves...

topolarity:
I think a proper fix would involve ignoring the directory of the wrong-version `llvm-config` and repeating the search. If we also pass the discovered LLVM directory to the Clang/LLD lookup as HINT, they should all resolve correctly together, even when the default system LLVM/Clang is a different version.

All of that said, that improvement is just a fallback that succeeds if:
  1. `CMAKE_INSTALL_PREFIX` is not provided
  2. The default system LLVM/Clang is the wrong version
  3. The right LLVM/Clang version is installed as a versioned lib (e.g. `libclang-14-dev`)

I don't have time to make that improvement right now, but I'll circle back later in the week to get it done.

PR #12117, opened 2022-07-14
Stage2: explain why value must be comptime known

Vexu:
Closes #9530
Closes #12144

PR #12121, opened 2022-07-14
Stage2 point to error location using spans

Vexu:
```zig
comptime {
    var a: u8 = 255;
    var b: u8 = 255;
    _ = (a - 1) + (b - 1);
}
comptime {
    var a: u8 = 255;
    var b: u8 = 255;
    _ = (a - 1) +
        (b - 1);
}
comptime {
    var a: u8 = 255;
    var b: u8 = 255;
    _ = (a - 1)
        + (b - 1);
}
comptime {
    var a: u8 = 255;
    var b: u8 = 255;
    _ = (a - 1)
        +
        (b - 1);
}
```
```
$ zig2 test a.zig
a.zig:4:17: error: overflow of integer type 'u8' with value '508'
    _ = (a - 1) + (b - 1);
        ~~~~~~~~^~~~~~~~~
a.zig:9:17: error: overflow of integer type 'u8' with value '508'
    _ = (a - 1) +
        ~~~~~~~~^
a.zig:16:9: error: overflow of integer type 'u8' with value '508'
        + (b - 1);
        ^~~~~~~~~
a.zig:22:9: error: overflow of integer type 'u8' with value '508'
        +
        ^
```

~~I didn't edit compile error tests yet since this is likely to touch most of them and I want to wait for #12117/comments.~~

~~One enhancement I can think of would be to have the caret (and the column number at the start) point at the main token instead of the start (`+` in all the examples)~~ Done.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -338,7 +338,7 @@ pub const AllErrors = struct {
             src_path: []const u8,
             line: u32,
             column: u32,
-            byte_offset: u32,
+            span: Module.SrcLoc.Span,
             /// Usually one, but incremented for redundant messages.
             count: u32 = 1,
             /// Does not include the trailing newline.
@@ -429,7 +429,10 @@ pub const AllErrors = struct {
                             try stderr.writeByte('\n');
                             try stderr.writeByteNTimes(' ', src.column);
                             ttyconf.setColor(stderr, .Green);
-                            try stderr.writeAll("^\n");
+                            try stderr.writeByte('^');
+                            // TODO basic unicode code point monospace width
+                            try stderr.writeByteNTimes('~', src.span.end - src.span.start - 1);
+                            try stderr.writeByte('\n');
                             ttyconf.setColor(stderr, .Reset);
                         }
                     }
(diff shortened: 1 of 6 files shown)


src/Compilation.zig
```diff
@@ -429,7 +429,10 @@ pub const AllErrors = struct {
                             try stderr.writeByte('\n');
                             try stderr.writeByteNTimes(' ', src.column);
                             ttyconf.setColor(stderr, .Green);
-                            try stderr.writeAll("^\n");
+                            try stderr.writeByte('^');
+                            // TODO basic unicode code point monospace width
+                            try stderr.writeByteNTimes('~', src.span.end - src.span.start - 1);
```
r00ster91, line 434: I assume you are talking about half-width vs full-width characters?
Here's what I did: https://github.com/r00ster91/Conversant/blob/877ae89bddb2dd0653d38617a8b355136df5fcb2/src/editor.zig#L28-L41
Checking for whether the UTF-8 representation of the character takes up >= 3 bytes works surprisingly well.

Vexu: I've also already solved it in my C compiler but didn't include it here yet but maybe I should've
https://github.com/Vexu/arocc/blob/23133aed979e69148b487138e2638e7f76ded3c4/src/Source.zig#L85-L109

This doesn't account for multi codepoint characters like emojis but that shouldn't be too hard either if desired.

r00ster91: Ooh, cool! Your solution definitely looks a lot more advanced than mine. Do you think we might want to add this to `std.unicode`? There is definitely demand for it but I remember that `std.unicode` shouldn't have too much functionality. Do you think this would be an ok addition?

Vexu: I do think it could be added to `std.unicode` but I'll do it in a separate PR since it'll also touch `std.zig.Ast.tokenLocation` and  people will have stronger opinions about it that this change.

PR #12132, opened 2022-07-15
std.json: Fix parsing of large numbers

ominitay:
Numbers greater than about 2^53 are encoded as strings in JSON, looking like `{ "12345678901234567890" }` instead of `{ 12345678901234567890 }`. `std.json.parseInternal` previously errored out because of this. It now tolerates integers wrapped in quotes. I've written a simple test case for this.

This has the side effect that `std.json.parse` will also accept small integers wrapped in quotes, though I think this is acceptable.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -1422,23 +1422,37 @@ fn parseInternal(
             };
         },
         .Float, .ComptimeFloat => {
-            const numberToken = switch (token) {
-                .Number => |n| n,
+            switch (token) {
+                .Number => |numberToken| return try std.fmt.parseFloat(T, numberToken.slice(tokens.slice, tokens.i - 1)),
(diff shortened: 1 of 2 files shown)


ominitay:
Self-hosted CI is failing on master too, not relevant to this PR.

topolarity:
AFAIK, the JSON spec doesn't actually define bounds on the numbers that can be encoded directly as an integer or a decimal, and it doesn't recommend wrapping in a string as a fallback.

That said, I'm sure this is common enough in the wild, and I don't have any problem with the more permissive parsing.

ominitay:
It's very common from what I've seen, and Zig itself does this. This change simply allows it to understand the JSON it (and similar implementations) produce :)

ifreund:
Is this not the reason `std.json.Value.NumberString` exists?

ominitay:
This change isn't to-do with `std.json.Value` at all. It is only concerning `std.json.parse`, which operates on `std.json.Token`.

PR #12133, opened 2022-07-16
build.zig: Warn if "config.h" cannot be located

topolarity:
Adds a warning if config.h could not be located:
```
$ ./stage2/bin/zig build -Dlog -Denable-llvm --zig-lib-dir lib --prefix $(pwd)/stage3 --search-prefix ~/local/llvm13-release
warning: config.h could not be located automatically. Consider providing it explicitly via "-Dconfig_h"
```

The search for "config.h" now also stops early if it sees a `.git` folder, so that we don't go beyond the repo root.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -235,7 +235,15 @@ pub fn build(b: *Builder) !void {
     exe_options.addOption([:0]const u8, "version", try b.allocator.dupeZ(u8, version));
 
     if (enable_llvm) {
-        const cmake_cfg = if (static_llvm) null else findAndParseConfigH(b, config_h_path_option);
+        const cmake_cfg = if (static_llvm) null else blk: {
+            if (findConfigH(b, config_h_path_option)) |config_h_path| {
+                const file_contents = fs.cwd().readFileAlloc(b.allocator, config_h_path, max_config_h_bytes) catch unreachable;
+                break :blk parseConfigH(b, file_contents);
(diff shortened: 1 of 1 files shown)


Vexu:
Previous discussion: https://github.com/ziglang/zig/pull/11501#pullrequestreview-959579630

topolarity:
Thanks for the context @Vexu - I'll to rework this one, clearly.

I wonder if copying the `config.h` into the installation during the build would be acceptable? This way Zig would inherit its build settings from the Zig building it.

andrewrk (changes requested):
I don't really think config.h belongs in the installation of Zig. Although I admit it's mostly personal taste, if you produce an installation of Zig using Zig without going through stage1, then no config.h is produced. It seems like, therefore, a canonical installation of Zig does not include config.h.

Is there something wrong with the recommended way to handle stage1, stage2, and stage3 installations side-by-side by running commands directly from the cmake build directory, or by passing `-Dconfig_h` with the `zig build` commands?

topolarity:
I share your preference actually 👌 

> Is there something wrong with the recommended way to handle stage1, stage2, and stage3 installations side-by-side by running commands directly from the cmake build directory, or by passing -Dconfig_h with the zig build commands?

I think this behavior is perfectly fine with the warning added. Let me scale back the change to just that.

Issue #10999, opened 2022-02-27
unsupported linker arg: -dead_strip


cstkingkey:
### Zig Version

0.10.0-dev.990+058e48224

### Steps to Reproduce

see also  #10485

### Expected Behavior

support linker arg  -dead_strip

### Actual Behavior

unsupported.

matu3ba:
`-dead_strip` is not part of lld. This belongs into the [OS X Component](https://opensource.apple.com/source/cctools/cctools-622.5/RelNotes/CompilerTools.html) issue.

matu3ba:
Note: `-dead_strip` is only supported on Darwin. https://stackoverflow.com/a/6687734

PR #12111, opened 2022-07-13
CI: increase stage3 test coverage

andrewrk:
* Use a debug build of stage3 instead of a debug build of stage2 for
   our self-hosted compiler test coverage.
 * Move coverage from stage1 to stage3 for:
   - building self-hosted without LLVM
   - building self-hosted for 32-bit arm
   - test-compiler-rt
   - test-behavior
   - test-std
     - caveat: release mode is skipped because I observed a problem
       locally.
   - test-compare-output
   - test-asm-link
   - test-fmt

PR #12140, opened 2022-07-16
macho: add support for `-dead_strip` (GC sections) and simplify symbol resolution

kubkon:
This is a major rewrite of the MachO linker that I felt was necessary to correctly facilitate garbage collecting unused atoms, aka supporting `-dead_strip` linker option. In summary, the linker now correctly splits the sections into subsections (if possible based on the header flag `MH_SUBSECTIONS_VIA_SYMBOLS` being set in the relocatable object file), works out alignment per each extracted atom (aka the subsection), and allocates an address in VM for it. When `-dead_strip` option is set, the linker additionally works out the GC roots and then traverses the link graph where node is an atom, and edge is a relocation, and marks an atom live or dead. Finally, each "dead" atom is then recycled (garbage collected, or GCed) and doesn't end up in the final binary.

Before pulling the trigger on this rewrite, I wanna spend some more time doing some perf analysis to work out if there are any bits that are unnecessarily slow, and to figure out if more binary size compression can be achieved in addition to the basic GC algorithm.

Resolves https://github.com/ziglang/zig/issues/10999

---
EDIT:

A cool by-product of the rewrite is that the high-level API is now almost 100% shared between Elf and MachO linkers. In Elf we currently call into `linkWithLLD` to do one-shot linking, while for MachO this is `linkOneShot`. `flushModule` is now equivalent between the two however. I hope this will make it easier to tweaks things for incremental codepaths shared by the linkers in the future.

---
EDIT2:

Some binary size results, all done using stage2+LLVM.

1. Basic hello world in C++ with libc++ statically linked:

```cpp
#include <iostream>

int main(int argc, char* argv[]) {
  std::cout << "Hello, world!" << std::endl;
  return 0;
}
```

```
496K    dead_strip
924K    no_dead_strip
```

2. stage3 compiler built in debug mode with and without dead stripping:

```
174M    stage3_deadstrip/bin/zig
191M    stage3_nodeadstrip/bin/zig
```


andrewrk (approved):

kubkon:
Now, only need to figure out a segfault when cross-compiling from linux using stage3, and I believe we're good to go with this one!

kubkon:
> Now, only need to figure out a segfault when cross-compiling from linux using stage3, and I believe we're good to go with this one!

OK, I managed to repro the segfault locally on Linux following the build steps in https://github.com/ziglang/zig/pull/12111#issuecomment-1186343263, and I tracked it down to this line [here](https://github.com/ziglang/zig/blob/8e75ba653b03477229cf72211e8a8bfe7b071254/src/link/MachO.zig#L3868):

```zig
    if (decl_state) |*ds| {
        try self.d_sym.?.dwarf.commitDeclState(
            &self.base,
            module,
            decl,
            symbol.n_value,
            decl.link.macho.size,
            ds,
        );
    }
```

Commenting the entire `if` statement makes the segfault  go away. Unwrapping the inner optional `self.d_sym` doesn't change a thing however - still segfault at the same location. Not 100% clear on what is happening just yet but perhaps miscomp of optionals in stage3?

kubkon:
> > Now, only need to figure out a segfault when cross-compiling from linux using stage3, and I believe we're good to go with this one!
> 
> OK, I managed to repro the segfault locally on Linux following the build steps in [#12111 (comment)](https://github.com/ziglang/zig/pull/12111#issuecomment-1186343263), and I tracked it down to this line [here](https://github.com/ziglang/zig/blob/8e75ba653b03477229cf72211e8a8bfe7b071254/src/link/MachO.zig#L3868):
> 
> ```zig
>     if (decl_state) |*ds| {
>         try self.d_sym.?.dwarf.commitDeclState(
>             &self.base,
>             module,
>             decl,
>             symbol.n_value,
>             decl.link.macho.size,
>             ds,
>         );
>     }
> ```
> 
> Commenting the entire `if` statement makes the segfault go away. Unwrapping the inner optional `self.d_sym` doesn't change a thing however - still segfault at the same location. Not 100% clear on what is happening just yet but perhaps miscomp of optionals in stage3?

I still haven't worked out a reduction but stage2 doesn't segfault which suggests this is a miscompilation in stage3.

kubkon:
Found it and fixed it! This actually isn't a miscompilation in stage3! It's a latent bug that was uncovered by stage1 and stage2 after my rewrite. Here's what happens:

In `fn updateFunc(...)` we issue a call to `fn placeDecl(...) !*macho.nlist_64` just before we exit. If you notice, we return a pointer to a symbol within the local symbol table. However, my rewrite changed a couple of things including how we handle creation of GOT `Atom`s. The tl;dr here is that we grab a pointer to a slot in an `ArrayList` representing the local symbol table, but then risk invalidating it just before returning it from the `placeDecl` function as we create a new GOT `Atom` which might append a new symbol to the local symbol table. I've rolled out a fix in cf6cfc8. Sorry for the noise!

PR #12148, opened 2022-07-17
Fix #9184: Support fmtIntSizeDec/fmtIntSizeBin FormatOptions

### LINKED ISSUE #9184: std.fmt.fmtIntSizeDec() and fmtIntSizeBin() do not consider FormatOptions

yujiri8:
I'm new to this. Am I going about it the right way, not changing the API of fmtIntSizeDec and fmtIntSizeBin themselves? I'm confused about the API, is this how it's meant to be used?
```zig
const std = @import("std");
const fmt = std.fmt;
const debug = std.debug;
const io = std.io;

pub fn main() !void {
    var stdout = io.getStdOut().writer();
    try fmt.fmtIntSizeBin(10000).format("", fmt.FormatOptions{.width = 12, .alignment = .Center}, stdout);
    // debug.print("{s}", .{result});
}
```

(It currently prints `9.765625    KiB     `. I assume the intended behavior would be to put no space inbetween the number and `KiB` and center the concanetation?)

Edit: nevermind my question about the API. I think I figured out how it's meant to be used, I didn't realize the format placeholders had that much flexibility

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -1,4 +1,5 @@
 const std = @import("std.zig");
+const io = std.io;
 const math = std.math;
 const assert = std.debug.assert;
 const mem = std.mem;
@@ -877,8 +878,11 @@ fn formatSizeImpl(comptime radix: comptime_int) type {
         ) !void {
             _ = fmt;
             if (value == 0) {
-                return writer.writeAll("0B");
+                return formatBuf("0B", options, writer);
             }
+            // The worst case in terms of space needed is 32 bytes + 3 for the suffix.
+            var buf: [35]u8 = undefined;
+            var bufstream = io.fixedBufferStream(buf[0..]);
 
             const mags_si = " kMGTPEZY";
             const mags_iec = " KMGTPEZY";
@@ -896,18 +900,20 @@ fn formatSizeImpl(comptime radix: comptime_int) type {
                 else => unreachable,
             };
 
-            try formatFloatDecimal(new_value, options, writer);
+            formatFloatDecimal(new_value, options, bufstream.writer()) catch |err| switch (err) {
+                error.NoSpaceLeft => unreachable, // 35 bytes should be enough
+            };
 
-            if (suffix == ' ') {
-                return writer.writeAll("B");
-            }
-
-            const buf = switch (radix) {
+            bufstream.writer().writeAll(if (suffix == ' ')
+                "B"
+            else switch (radix) {
                 1000 => &[_]u8{ suffix, 'B' },
                 1024 => &[_]u8{ suffix, 'i', 'B' },
                 else => unreachable,
+            }) catch |err| switch (err) {
+                error.NoSpaceLeft => unreachable,
             };
-            return writer.writeAll(buf);
+            return formatBuf(bufstream.getWritten(), options, writer);
         }
     };
 }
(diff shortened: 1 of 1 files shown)


lib/std/fmt.zig
```diff
@@ -876,8 +876,11 @@ fn formatSizeImpl(comptime radix: comptime_int) type {
             writer: anytype,
         ) !void {
             _ = fmt;
+            // The worst case in terms of space needed is... TODO.
+            var buf: [1000]u8 = undefined;
             if (value == 0) {
-                return writer.writeAll("0B");
+                mem.copy(u8, buf[0..], "0B");
+                return formatBuf(buf[0..2], options, writer);
```
Vexu, line 883:
```suggestion
                return formatBuf("0B", options, writer);
```

lib/std/fmt.zig
```diff
@@ -876,8 +876,11 @@ fn formatSizeImpl(comptime radix: comptime_int) type {
             writer: anytype,
         ) !void {
             _ = fmt;
+            // The worst case in terms of space needed is... TODO.
+            var buf: [1000]u8 = undefined;
```
Vexu, line 880: The worst case for `formatFloatDecimal` seems to be 32 bytes + 3 bytes for the suffix.

Vexu (changes requested):
The float value and the plain `B` suffix are still being written directly to the writer. Instead everything should be written into a `std.io.fixedBufferStream` backed by the buffer on stack and then the `fbs.getWritten()` should be written into the original writer using `formatBuf`.

yujiri8:
Using fixedBufferStream now. This almost works, but the tests are failing with:
```
./lib/std/fmt.zig:1662:13: error: expected type 'std.io.NullWriter', found '@typeInfo(@typeInfo(@TypeOf(std.fmt.formatSizeImpl(1000).f)).Fn.return_type.?).ErrorUnion.error_set'
            try format_fn(self.data, fmt, options, writer);
            ^
./lib/std/fmt.zig:461:32: note: called from here
        return try value.format(actual_fmt, options, writer);
                               ^
./lib/std/fmt.zig:181:23: note: called from here
        try formatType(
                      ^
./lib/std/fmt.zig:1893:11: note: called from here
    format(counting_writer.writer(), fmt, args) catch |err| switch (err) {};
          ^
./lib/std/fmt.zig:1900:40: note: called from here
    const size = math.cast(usize, count(fmt, args)) orelse return error.OutOfMemory;
                                       ^
./lib/std/testing.zig:201:42: note: called from here
    const result = try std.fmt.allocPrint(allocator, template, args);
                                         ^
./test.zig:10:26: note: called from here
    try testing.expectFmt("", "{d}", .{fmt.fmtIntSizeDec(42)});
                         ^
./test.zig:7:21: note: called from here
pub fn main() !void {
                    ^
./lib/std/io/fixed_buffer_stream.zig:16:38: note: 'error.NoSpaceLeft' not a member of destination error set
        pub const WriteError = error{NoSpaceLeft};
                                     ^
```
Replacing every usage of `try` in the function with `catch unreachable` solves it, but I don't think that's correct since it was using try before when calling formatFloatDecimal. Meanwhile my own test case like this works fine:
```zig
const std = @import("std");
const fmt = std.fmt;
const debug = std.debug;
const io = std.io;
const testing = std.testing;

pub fn main() !void {
    debug.print("{s: ^20}", .{fmt.fmtIntSizeDec(18446744073709551615)});
    // try testing.expectFmt("", "{d}", .{fmt.fmtIntSizeDec(42)});
}
```
It seems to be only testing.expectFmt that doesn't like it.

yujiri8:
Updated! The tests pass now and testing.expectFmt seems to work. Thanks for your help!

Vexu (approved):
Could you add  a new test for alignment and width like was done in #10361?

yujiri8:
Done! Four new test cases.

PR #12151, opened 2022-07-17
std.zig.system.NativeTargetInfo: look for a shebang line in /usr/bin/env, if any

BratishkaErik:
UPD: See https://github.com/ziglang/zig/pull/12151#pullrequestreview-1044192217

### DIFF
--- a/lib/std/zig/system/NativeTargetInfo.zig
+++ b/lib/std/zig/system/NativeTargetInfo.zig
@@ -237,7 +237,7 @@ pub fn detect(allocator: Allocator, cross_target: CrossTarget) DetectError!Nativ
 
 /// First we attempt to use the executable's own binary. If it is dynamically
 /// linked, then it should answer both the C ABI question and the dynamic linker question.
-/// If it is statically linked, then we try /usr/bin/env. If that does not provide the answer, then
+/// If it is statically linked, then we try /usr/bin/env (or the file it references in shebang). If that does not provide the answer, then
 /// we fall back to the defaults.
 /// TODO Remove the Allocator requirement from this function.
 fn detectAbiAndDynamicLinker(
@@ -355,37 +355,77 @@ fn detectAbiAndDynamicLinker(
         return result;
     }
 
-    const env_file = std.fs.openFileAbsoluteZ("/usr/bin/env", .{}) catch |err| switch (err) {
-        error.NoSpaceLeft => unreachable,
-        error.NameTooLong => unreachable,
-        error.PathAlreadyExists => unreachable,
-        error.SharingViolation => unreachable,
-        error.InvalidUtf8 => unreachable,
-        error.BadPathName => unreachable,
-        error.PipeBusy => unreachable,
-        error.FileLocksNotSupported => unreachable,
-        error.WouldBlock => unreachable,
-        error.FileBusy => unreachable, // opened without write permissions
+    const elf_file = blk: {
+        // This block looks for a shebang line in /usr/bin/env,
+        // if it finds one, then instead of using /usr/bin/env as the ELF file to examine, it uses the file it references instead,
+        // doing the same logic recursively in case it finds another shebang line.
 
-        error.IsDir,
-        error.NotDir,
-        error.InvalidHandle,
-        error.AccessDenied,
-        error.NoDevice,
-        error.FileNotFound,
-        error.FileTooBig,
-        error.Unexpected,
-        => return defaultAbiAndDynamicLinker(cpu, os, cross_target),
+        // Since /usr/bin/env is hard-coded into the shebang line of many portable scripts, it's a
+        // reasonably reliable path to start with.
+        var file_name: []const u8 = "/usr/bin/env";
+        // #! (2) + 255 (max length of shebang line since Linux 5.1) + \n (1)
+        var buffer: [258]u8 = undefined;
+        while (true) {
+            const file = std.fs.openFileAbsolute(file_name, .{}) catch |err| switch (err) {
+                error.NoSpaceLeft => unreachable,
+                error.NameTooLong => unreachable,
+                error.PathAlreadyExists => unreachable,
+                error.SharingViolation => unreachable,
+                error.InvalidUtf8 => unreachable,
+                error.BadPathName => unreachable,
+                error.PipeBusy => unreachable,
+                error.FileLocksNotSupported => unreachable,
+                error.WouldBlock => unreachable,
+                error.FileBusy => unreachable, // opened without write permissions
 
-        else => |e| return e,
+                error.IsDir,
+                error.NotDir,
+                error.InvalidHandle,
+                error.AccessDenied,
+                error.NoDevice,
+                error.FileNotFound,
+                error.FileTooBig,
+                error.Unexpected,
+                => |e| {
+                    std.log.warn("Encoutered error: {s}, falling back to default ABI and dynamic linker.\n", .{@errorName(e)});
+                    return defaultAbiAndDynamicLinker(cpu, os, cross_target);
+                },
+
+                else => |e| return e,
+            };
+
+            const line = file.reader().readUntilDelimiter(&buffer, '\n') catch |err| switch (err) {
+                error.IsDir => unreachable, // Handled before
+                error.AccessDenied => unreachable,
+                error.WouldBlock => unreachable, // Did not request blocking mode
+                error.OperationAborted => unreachable, // Windows-only
+                error.BrokenPipe => unreachable,
+                error.ConnectionResetByPeer => unreachable,
+                error.ConnectionTimedOut => unreachable,
+                error.InputOutput => unreachable,
+                error.Unexpected => unreachable,
+
+                error.StreamTooLong,
+                error.EndOfStream,
+                error.NotOpenForReading,
+                => break :blk file,
+
+                else => |e| {
+                    file.close();
+                    return e;
+                },
+            };
+            if (!mem.startsWith(u8, line, "#!")) break :blk file;
+            var it = std.mem.tokenize(u8, line[2..], " ");
+            file.close();
+            file_name = it.next() orelse return defaultAbiAndDynamicLinker(cpu, os, cross_target);
+        }
     };
-    defer env_file.close();
+    defer elf_file.close();
 
     // If Zig is statically linked, such as via distributed binary static builds, the above
-    // trick won't work. The next thing we fall back to is the same thing, but for /usr/bin/env.
-    // Since that path is hard-coded into the shebang line of many portable scripts, it's a
-    // reasonably reliable path to check for.
-    return abiAndDynamicLinkerFromFile(env_file, cpu, os, ld_info_list, cross_target) catch |err| switch (err) {
+    // trick (block self_exe) won't work. The next thing we fall back to is the same thing, but for elf_file.
+    return abiAndDynamicLinkerFromFile(elf_file, cpu, os, ld_info_list, cross_target) catch |err| switch (err) {
         error.FileSystem,
         error.SystemResources,
         error.SymLinkLoop,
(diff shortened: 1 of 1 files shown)


lib/std/zig/system/NativeTargetInfo.zig
```diff
@@ -355,55 +355,35 @@ fn detectAbiAndDynamicLinker(
...
-        error.UnexpectedEndOfFile,
-        error.NameTooLong,
-        // Finally, we fall back on the standard path.
-        => defaultAbiAndDynamicLinker(cpu, os, cross_target),
+    // trick won't work. The next thing we fall back to is the same thing, but for /usr/bin/env and /bin/bash.
+    // Since these paths are hard-coded into the shebang line of many portable scripts, they are a
+    // reasonably reliable paths to check for.
+    return result: {
+        break :result abiAndDynamicLinkerFromPath("/usr/bin/env", cpu, os, ld_info_list, cross_target) catch {
+            // If we can't parse /usr/bin/env ELF file (for example, if coreutils was builded with --enable-single-binary and /usr/bin/env is actually a script), try /bin/bash
```
BratishkaErik, line 364: For example, Gentoo's `sys-apps/coreutils` with USE flag `multicall`, or Red Hat Universal Base Image 9's `coreutils-single` https://github.com/ziglang/zig/issues/12156#issuecomment-1187036721

andrewrk:
On which system do you observe `/usr/bin/env` not being an ELF file? And what is it instead?

BratishkaErik:
> On which system do you observe `/usr/bin/env` not being an ELF file? And what is it instead?

https://github.com/ziglang/zig/pull/12151#discussion_r923210625

Instea it is a script with following content:
```sh
#!/usr/bin/coreutils --coreutils-prog-shebang=env
```
or similar.

andrewrk (changes requested):
Here is my serious suggestion: modify the code to look for a shebang line in `/usr/bin/env`. If it finds one, then instead of using `/usr/bin/env` as the ELF file to examine, it uses that one instead. In the above example, it would therefore examine `/usr/bin/coreutils` (doing the same logic recursively in case it finds another script).

lib/std/zig/system/NativeTargetInfo.zig
```diff
@@ -355,37 +355,55 @@ fn detectAbiAndDynamicLinker(
...
 
-        else => |e| return e,
+                else => |e| return e,
+            };
+
+            // #! (2) + 255 (max length of shebang line since Linux 5.1) + \n (1)
+            var buffer: [258]u8 = undefined;
+            const line = file.reader().readUntilDelimiter(&buffer, '\n') catch break :blk file;
+            if (mem.startsWith(u8, line, "#!")) {
+                file_name = std.mem.tokenize(u8, line[2..], " ").next().?;
```
Vexu, line 396:
```suggestion
                var it = std.mem.tokenize(u8, line[2..], " ");
                file_name = it.next() orelse line[2..]; 
```
I don't think we should assert that the line contains a space.

BratishkaErik: We don't? It works perfectly fine without spaces in file, `tokenize` returns entire line if it can't find delimiter.

Vexu: Right, I looked at the tests incorrectly, it won't panic if there are no spaces but it will panic if `line[2..].len == 0`.

BratishkaErik: > Right, I looked at the tests incorrectly, it won't panic if there are no spaces but it will panic if `line[2..].len == 0`.

Should we fall back to default if this is an incomplete shebang?

```suggestion
                if (line[2..].len == 0) return defaultAbiAndDynamicLinker(cpu, os, cross_target);
                file_name = std.mem.tokenize(u8, line[2..], " ").next().?;
```

Vexu: That sounds reasonable.

BratishkaErik: > That sounds reasonable.

Done!

BratishkaErik: > I assigned the result of `tokenize` to a new variable deliberately, using it directly like this is invalid in stage2. Also why not use `orelse` when it is right there?

Sorry, done

lib/std/zig/system/NativeTargetInfo.zig
```diff
@@ -355,37 +355,57 @@ fn detectAbiAndDynamicLinker(
...
+                else => |e| return e,
+            };
+
+            // #! (2) + 255 (max length of shebang line since Linux 5.1) + \n (1)
+            var buffer: [258]u8 = undefined;
+            const line = file.reader().readUntilDelimiter(&buffer, '\n') catch break :blk file;
+            if (mem.startsWith(u8, line, "#!")) {
+                if (line[2..].len == 0) return defaultAbiAndDynamicLinker(cpu, os, cross_target);
+                var it = std.mem.tokenize(u8, line[2..], " ");
+                file_name = it.next() orelse line[2..];
```
Vexu, line 398: What I meant was doing something like:
```suggestion
                var it = std.mem.tokenize(u8, line[2..], " ");
                file_name = it.next() orelse {
                    file.close();
                    return defaultAbiAndDynamicLinker(cpu, os, cross_target);
                }
```

It would probably be better to use `defer` with a guard like:
```zig
var should_close = true;
defer if (should_close) file.close();
```
And then set `should_close = false` breaking from the block.

BratishkaErik: > What I meant was doing something like:
> 
> It would probably be better to use `defer` with a guard like:
> 
> ```zig
> var should_close = true;
> defer if (should_close) file.close();
> ```
> 
> And then set `should_close = false` breaking from the block.

Done!

lib/std/zig/system/NativeTargetInfo.zig
```diff
@@ -355,37 +355,63 @@ fn detectAbiAndDynamicLinker(
...
+            // #! (2) + 255 (max length of shebang line since Linux 5.1) + \n (1)
+            var buffer: [258]u8 = undefined;
+            const line = file.reader().readUntilDelimiter(&buffer, '\n') catch {
+                should_close = false;
+                break :blk file;
+            };
+            if (mem.startsWith(u8, line, "#!")) {
+                var it = std.mem.tokenize(u8, line[2..], " ");
+                file_name = it.next() orelse return defaultAbiAndDynamicLinker(cpu, os, cross_target);
+                continue;
```
topolarity, line 403: `file_name` is a pointer into `buffer`, which is invalidated at this point

BratishkaErik: > I think there's a memory-validity problem here: `file_name` is a pointer into `buffer`, which is invalidated at this point
> 
> You should be able to fix it by changing where `buffer` is declared

This may sounds strange, but... this is the reason why I changed where `buffer` is declared...

topolarity: I think it should be placed outside of the while loop - Do you agree?

BratishkaErik: > I think it should be placed outside of the while loop - Do you agree?

I'll check this in the morning :)

BratishkaErik: > I think it should be placed outside of the while loop - Do you agree?

Done!

topolarity:
I think you're hitting a stage1 compiler bug. This diff seems to get things working again:
```diff
index c5b9454c9..d9467fe2a 100644
--- a/lib/std/zig/system/NativeTargetInfo.zig
+++ b/lib/std/zig/system/NativeTargetInfo.zig
@@ -390,19 +390,16 @@ fn detectAbiAndDynamicLinker(

                 else => |e| return e,
             };
-            var should_close = true;
-            defer if (should_close) file.close();

             const line = file.reader().readUntilDelimiter(&buffer, '\n') catch {
-                should_close = false;
                 break :blk file;
             };
             if (mem.startsWith(u8, line, "#!")) {
                 var it = std.mem.tokenize(u8, line[2..], " ");
                 file_name = it.next() orelse return defaultAbiAndDynamicLinker(cpu, os, cross_target);
+                file.close();
                 continue;
             } else {
-                should_close = false;
                 break :blk file;
             }
         }
```

BratishkaErik:
> I think you're hitting a stage1 compiler bug.
This diff seems to get things working again:
> ```diff
> index c5b9454c9..d9467fe2a 100644
> --- a/lib/std/zig/system/NativeTargetInfo.zig
> +++ b/lib/std/zig/system/NativeTargetInfo.zig
> @@ -390,19 +390,16 @@ fn detectAbiAndDynamicLinker(
> 
>                  else => |e| return e,
>              };
> -            var should_close = true;
> -            defer if (should_close) file.close();
> 
>              const line = file.reader().readUntilDelimiter(&buffer, '\n') catch {
> -                should_close = false;
>                  break :blk file;
>              };
>              if (mem.startsWith(u8, line, "#!")) {
>                  var it = std.mem.tokenize(u8, line[2..], " ");
>                  file_name = it.next() orelse return defaultAbiAndDynamicLinker(cpu, os, cross_target);
> +                file.close();
>                  continue;
>              } else {
> -                should_close = false;
>                  break :blk file;
>              }
>          }
> ```

Thank you!

lib/std/zig/system/NativeTargetInfo.zig
```diff
@@ -355,37 +355,54 @@ fn detectAbiAndDynamicLinker(
...
+                error.FileNotFound,
+                error.FileTooBig,
+                error.Unexpected,
+                => return defaultAbiAndDynamicLinker(cpu, os, cross_target),
 
-        else => |e| return e,
+                else => |e| return e,
+            };
+
+            const line = file.reader().readUntilDelimiter(&buffer, '\n') catch break :blk file;
```
marler8997, line 394: This code looks eerily similar to a PR I made to declare a minimum zig version :)
https://github.com/ziglang/zig/pull/11987/files#diff-581e0ec2dae9e97448da6d140a75539dc6c79595dfc4c70df65b98c000f8e6f5R3812-R3828

One thing I did differently was I only catch/handle 2 kinds of errors when reading the file, `StreamTooLong` and `EndOfStream`.  Would it make more sense to only catch/handle those 2 errors and propagate the rest? i.e.

```zig
const line = file.reader().readUntilDelimiter(&buffer, '\n') catch |err| switch (err) {
    error.StreamTooLong, error.EndOfStream => break :blk file,
    else => |e| return e,
};
```

BratishkaErik: I'm not sure, I guess I'll wait for others' opinions

marler8997: It also looks like the code above that opens the file has the same idea.

topolarity (approved):
@marler8997 's suggestion is a good one, assuming it doesn't affect upstream error handling

Changes look good though :+1:

lib/std/zig/system/NativeTargetInfo.zig
```diff
@@ -355,37 +356,74 @@ fn detectAbiAndDynamicLinker(
...
+                error.IsDir => unreachable, // Handled before
+                error.WouldBlock => unreachable, // Did not request blocking mode
+                error.OperationAborted => unreachable, // Windows-only
+
+                error.NotOpenForReading => return error.Unexpected,
+                error.BrokenPipe => return error.Unexpected,
+                error.ConnectionResetByPeer => return error.Unexpected,
+                error.ConnectionTimedOut => return error.Unexpected,
+                error.InputOutput => return error.Unexpected,
+                error.AccessDenied => return error.Unexpected,
```
Vexu, line 408: These should be `unreachable` or there should be `errdefer file.close()` before this call.

BratishkaErik: Done.

BratishkaErik:
Thank you for your helpful reviews! :)

Issue #12112, opened 2022-07-14
`zig build` crashes while try to determine native system from rpath


leroycep:
### Zig Version

0.10.0-dev.2981+7090f0471

### Steps to Reproduce

I'm on nixos and using [this nix flake](https://gist.github.com/leroycep/a427b1456bb63aec4668b54b78bd0d8d) to set up a zig development environment.

So:
- be on nixos
- have flakes enabled
```sh
mkdir hello
cd hello
curl https://gist.githubusercontent.com/leroycep/a427b1456bb63aec4668b54b78bd0d8d/raw/af36cf585891a7ec8710d976990626b08785b7c2/flake.nix > flake.nix
nix develop

# now inside the nix develop shell
zig init-exe
zig build run
```

### Expected Behavior

Zig to build the "hello world" program that `init-exe` produces.

### Actual Behavior

Running `zig build <anything>` will result in the following error message:

```
thread 617946 panic: index out of bounds
/nix/store/z0mk8ks94p1mid26ybhrh2hq3ywpgm60-zig-0.10.0-dev.2981+7090f0471/lib/zig/std/zig/system/NativeTargetInfo.zig:661:78: 0x2aaecf in std.zig.system.NativeTargetInfo.abiAndDynamicLinkerFromFile (build)
                const rpath_list = mem.sliceTo(std.meta.assumeSentinel(strtab[rpoff_usize..].ptr, 0), 0);
                                                                             ^
/nix/store/z0mk8ks94p1mid26ybhrh2hq3ywpgm60-zig-0.10.0-dev.2981+7090f0471/lib/zig/std/zig/system/NativeTargetInfo.zig:388:39: 0x29bdf6 in std.zig.system.NativeTargetInfo.detectAbiAndDynamicLinker (build)
    return abiAndDynamicLinkerFromFile(env_file, cpu, os, ld_info_list, cross_target) catch |err| switch (err) {
                                      ^
/nix/store/z0mk8ks94p1mid26ybhrh2hq3ywpgm60-zig-0.10.0-dev.2981+7090f0471/lib/zig/std/zig/system/NativeTargetInfo.zig:201:47: 0x299082 in std.zig.system.NativeTargetInfo.detect (build)
    var result = try detectAbiAndDynamicLinker(allocator, cpu, os, cross_target);
                                              ^
[...]

PR #12155, opened 2022-07-17
Read dynstr starting at rpath offset

leroycep:
Since we know the offset, we may as well read starting there. Still expects
rpath to fit in 4096 bytes; that might be worth fixing in the future.

Fixes issue #12112


andrewrk:
Thanks!

McSinyx:
Should this be cherry-picked on top of 0.9.x?

PR #12160, opened 2022-07-18
std.fmt: require specifier for unwrapping ?T and E!T.

### LINKED ISSUE #11753: `std.fmt`: disallow implicitly formatting optional and error values

InKryption:
Replaces #11754 (cleaner commit history & avoids merge conflict stuff).
Closes #11753.

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -574,7 +574,7 @@ pub fn main() void {
     var optional_value: ?[]const u8 = null;
     assert(optional_value == null);
 
-    print("\noptional 1\ntype: {s}\nvalue: {s}\n", .{
+    print("\noptional 1\ntype: {s}\nvalue: {?s}\n", .{
         @typeName(@TypeOf(optional_value)),
         optional_value,
     });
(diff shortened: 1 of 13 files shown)


Vexu (approved):

andrewrk:
Reverted in 7cbd586ace46a8e8cebab660ebca3cfc049305d9

Issue #3180, opened 2019-09-05
Use-after-free not detected


jrmuizel:
The following program builds and runs without complaint. Is there the intent to detect UAFs or should one expect memory safety similar to C? 

```zig
const std = @import("std");
const warn = @import("std").debug.warn;

const P = struct {
   id: u32,
};

pub fn main() !void {
    const alloc = std.heap.direct_allocator;
    const k = try alloc.create(P);
    k.id = 5;
    alloc.destroy(k);
    const m = try alloc.create(P);
    warn("value: {}\n", k.id);
}
```

andrewrk:
Use-after-free is a research area of the #2301 project. See also #1966.

andrewrk:
Oh, see also https://github.com/andrewrk/zig-general-purpose-allocator/. It's not quite ready to merge into std, but that's going to be the go-to allocator for debug builds at least. It has use-after-free detection that works until all virtual memory is exhausted, at which point addresses have to be recycled.

gunnarahlberg:
I just happened to have this bug. It still applies, albeit the source should be updated to compile

```
    ...
    warn("value: {}\n", .{k.id});
`
> value: 2863311530

**Comment by andrewrk** [comment]
Use after free is now solved as far as safety is concerned with heap allocations, if you use the std lib page_allocator or GeneralPurposeAllocator. This is true even if you use it to e.g. back an ArenaAllocator.

What's left for this issue is to make escaped stack allocations safe.

To revisit the original test case here:

```
$ ./zig build-exe test.zig 
$ ./test
Segmentation fault at address 0x7f7aba428000
/home/andy/Downloads/zig/build/test.zig:13:39: 0x22fd51 in main (test)
    std.debug.print("value: {}\n", .{k.id});
                                      ^
/home/andy/Downloads/zig/lib/std/start.zig:257:37: 0x204c4d in std.start.posixCallMainAndExit (test)
            const result = root.main() catch |err| {
                                    ^
/home/andy/Downloads/zig/lib/std/start.zig:128:5: 0x20498f in std.start._start (test)
    @call(.{ .modifier = .never_inline }, posixCallMainAndExit, .{});
    ^
Aborted (core dumped)
```

Here's similar test case. This one uses GeneralPurposeAllocator and is adversarial by forcing the page to stay mapped:

```zig
const std = @import("std");

const P = struct {
    id: u32,
};

pub fn main() !void {
[...]

PR #12175, opened 2022-07-20
macho: Pass sections by pointer when slicing names

topolarity:
Thought these were fishy in the debug(link) logs:
```
debug(link): putting section ' ,@"���' as an Atom
debug(link): putting section ' ,@"���' as an Atom
...
debug(link):  ,81��� => size: 0x18, alignment: 0x3
debug(link):  ,81��� => size: 0x980, alignment: 0x4
... etc.
```

Turns out, we were accidentally returning a pointer to a parameter - Stage 1 just happened to pass the parameter by reference.

One more stepping stone for #12111

### DIFF
--- a/lib/std/macho.zig
+++ b/lib/std/macho.zig
@@ -653,7 +653,7 @@ pub const segment_command_64 = extern struct {
     nsects: u32 = 0,
     flags: u32 = 0,
 
-    pub fn segName(seg: segment_command_64) []const u8 {
+    pub fn segName(seg: *const segment_command_64) []const u8 {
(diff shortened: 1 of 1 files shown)


andrewrk:
Nice find! I think an important language goal will be having safety for this kind of problem. Possibly by changing aliasing rules and possibly by doing something along the lines of https://github.com/ziglang/zig/issues/3180#issuecomment-680259699.

topolarity:
Big +1

I have some other crazy ideas in the space I want to try out at some point - Looking forward to the experimental phase.

PR #11867, opened 2022-06-14
 stage2: ensure builtin packages are always available

nektro:
after a certain package depth, it appears access to the `'std'` package is lost. with this fix building [nektro/aquila](https://github.com/nektro/aquila) reaches a new error

```
[nix-shell:~/other/dev/aquila]$ ~/other/dev/zig/zig-out/bin/zig build --prominent-compile-errors -fno-stage1
/home/meg/other/dev/aquila/.zigmod/deps/git/github.com/nektro/zig-zorm/src/sqlite3.zig:1:21: error: unable to open 'std': PackageNotFound
const std = @import("std");
                    ^
aquila...The step exited with error code 1
error: the build command failed with exit code 1

[nix-shell:~/other/dev/aquila]$ ~/other/dev/zig/zig-out/bin/zig build --prominent-compile-errors -fno-stage1
Semantic Analysis [2456] ensureTotalCapacityPrecise... thread 78535 panic: integer overflow
/home/meg/other/dev/zig/lib/std/mem.zig:3045:54: 0x2dd642a in std.mem.alignForwardGeneric (zig)
    return alignBackwardGeneric(T, addr + (alignment - 1), alignment);
                                                     ^
/home/meg/other/dev/zig/src/codegen/llvm.zig:2651:57: 0x30e8575 in codegen.llvm.DeclGen.lowerType (zig)
                    offset = std.mem.alignForwardGeneric(u64, offset, field_align);
                                                        ^
/home/meg/other/dev/zig/src/codegen/llvm.zig:2376:43: 0x30df435 in codegen.llvm.DeclGen.resolveGlobalDecl (zig)
        const llvm_type = try dg.lowerType(decl.ty);
                                          ^
/home/meg/other/dev/zig/src/codegen/llvm.zig:3740:39: 0x30ef167 in codegen.llvm.DeclGen.lowerDeclRefValue (zig)
            try self.resolveGlobalDecl(decl_index);
                                      ^
....
```

PR #12177, opened 2022-07-20
stage2 llvm: fix handling of pointer fields in packed structs

Vexu:
This makes [Aro](https://github.com/Vexu/arocc) pass all tests with #11798 worked around and  #11867 applied.

### DIFF
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -5149,6 +5149,11 @@ pub const FuncGen = struct {
                             const same_size_int = self.context.intType(elem_bits);
                             const truncated_int = self.builder.buildTrunc(shifted_value, same_size_int, "");
                             return self.builder.buildBitCast(truncated_int, elem_llvm_ty, "");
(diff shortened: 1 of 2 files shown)


andrewrk:
I think we can merge this since it is an improvement over status quo, which produces an invalid LLVM module, although I would like to improve the test case to actually check that the code is correctly compiled rather than only checking that it does not crash the compiler.

Note however that we do have some more work to do with pointers to packed structs:

```zig
const Symbol = struct {
    index: u16,
};

const Relocation = packed struct {
    symbol: *Symbol,
    a: u32,
};

var x: Relocation = .{
    .symbol = &s,
    .a = 1234,
};

var s: Symbol = undefined;

test "global packed struct with pointer field" {
    _ = x;
}
```

```
Invalid bitcast
i64 bitcast (%test3.Symbol* @test3.s to i64)

thread 81531 panic: LLVM module verification failed
```

In some cases it is impossible to lower code like this because there is no such relocation supported by the object file. However, I think we can allow it when it does work, as long as we can detect when it won't work and emit a compile error.

PR #12179, opened 2022-07-20
stage2: implement `noinline fn`

### LINKED ISSUE #11767: implement noinline
https://github.com/ziglang/zig/blob/6d691d3540b1979f0a8d604922c93209b9c0dba0/src/AstGen.zig#L3364

```zig
noinline fn foo() void {}
```

If you look at the ZIR, the "noinline" information is lost.

nektro:
Fixes #11767

```zig
test {
    foo();
}

noinline fn foo() void {
    //
}
```

```
[meg@nixos:~/dev/zig]$ ./zig-out/bin/zig ast-check -t test.zig 
# Source bytes:       55B
# Tokens:             16 (104B)
# AST Nodes:          9 (237B)
# Total ZIR bytes:    398B
# Instructions:       15 (135B)
# String Table Bytes: 7B
# Extra Data Items:   48 (192B)
%0 = extended(struct_decl(parent, Auto, {
  test line(0) hash(53b4e586ab7c73d53a2fda7c58ad19aa): %1 = block_inline({
    %9 = func_fancy(test, inferror, body={
      %2 = dbg_block_begin())
      %3 = dbg_stmt(2, 5)
      %4 = decl_val("foo") token_offset:2:5 to :2:8
      %5 = dbg_stmt(2, 8)
      %7 = dbg_block_end())
      %6 = call(nodiscard .auto, %4, []) node_offset:2:5 to :2:10
      %8 = ret_tok(@Zir.Inst.Ref.void_value) token_offset:3:1 to :3:2
    }) (lbrace=1:6,rbrace=3:1) node_offset:1:1 to :1:5
    %10 = break_inline(%1, %9)
  }) node_offset:1:1 to :1:5
  [40] foo line(4) hash(f3339e121db838fc9d98c7412345a9c5): %11 = block_inline({
    %13 = func_fancy(noinline, body={
      %12 = ret_tok(@Zir.Inst.Ref.void_value) token_offset:7:1 to :7:2
    }) (lbrace=1:24,rbrace=3:1) node_offset:5:1 to :5:12
    %14 = break_inline(%11, %13)
  }) node_offset:5:1 to :5:23
}, {}, {})
```

```ll
<snip>

; Function Attrs: noinline nounwind
define internal fastcc void @test.foo() unnamed_addr #0 !dbg !4183 {
Entry:
  ret void
}

<snip>
```

Full `.ll` log: https://clbin.com/ztvi5

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -1274,6 +1274,7 @@ fn fnProtoExpr(
         .is_inferred_error = false,
         .is_test = false,
         .is_extern = false,
+        .is_noinline = false,
         .noalias_bits = noalias_bits,
     });
 
@@ -3389,7 +3390,6 @@ fn fnDecl(
     };
     defer fn_gz.unstack();
 
-    // TODO: support noinline
     const is_pub = fn_proto.visib_token != null;
     const is_export = blk: {
         const maybe_export_token = fn_proto.extern_export_inline_token orelse break :blk false;
@@ -3402,6 +3402,10 @@ fn fnDecl(
     const has_inline_keyword = blk: {
         const maybe_inline_token = fn_proto.extern_export_inline_token orelse break :blk false;
         break :blk token_tags[maybe_inline_token] == .keyword_inline;
+    };
+    const is_noinline = blk: {
+        const maybe_noinline_token = fn_proto.extern_export_inline_token orelse break :blk false;
+        break :blk token_tags[maybe_noinline_token] == .keyword_noinline;
     };
 
     const doc_comment_index = try astgen.docCommentAsString(fn_proto.firstToken());
@@ -3610,6 +3614,7 @@ fn fnDecl(
             .is_inferred_error = false,
             .is_test = false,
             .is_extern = true,
+            .is_noinline = is_noinline,
             .noalias_bits = noalias_bits,
         });
     } else func: {
@@ -3658,6 +3663,7 @@ fn fnDecl(
             .is_inferred_error = is_inferred_error,
             .is_test = false,
             .is_extern = false,
+            .is_noinline = is_noinline,
             .noalias_bits = noalias_bits,
         });
     };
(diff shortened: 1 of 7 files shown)


src/AstGen.zig
```diff
@@ -3599,6 +3603,10 @@ fn fnDecl(
         break :inst inst;
     };
 
+    if (has_inline_keyword and is_noinline) {
+        return astgen.failTok(fn_name_token, "function prototype may not be both inline and noinline", .{});
+    }
```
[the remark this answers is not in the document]
nektro: good catch

r00ster91:
Can it also handle this?
```zig
pub fn main() void {
    x();
}

noinline fn x() callconv(.Inline) void {}
```
I think it should be an error because it's ambiguous. In stage1 it's
```
broken LLVM module found: Attributes 'noinline and alwaysinline' are incompatible!
void ()* @x

This is a bug in the Zig compiler.thread 29374 panic:
Aborted (core dumped)
```
And we might also want to check other possibly problematic calling conventions in `std.builtin.CallingConvention` but I don't see any other apart from `.Inline`.

nektro:
good catch, ill fix both notes and re-push

src/Sema.zig
```diff
@@ -7350,6 +7352,11 @@ fn funcCommon(
             });
         }
 
+        // TODO do this check in AstGen if https://github.com/ziglang/zig/issues/11834 is accepted
```
Vexu, line 7355: This check will still need to happen in `Sema` even with #11834 since the callconv expression can still be any comptime expression, just not one affected by comptime parameters.

nektro: so you could for example have it be a global or function call, as long as it didnt use the parameters?

Vexu: Yes, you can look at `kernel32.zig` for an example of where this is a useful feature.

PR #12184, opened 2022-07-21
complete the drone pipeline runner type and platform os

zc2638:


andrewrk:
Cool profile pic. Can you explain the motivation for this change?

zc2638:
> Cool profile pic. Can you explain the motivation for this change?

The default runner type of the drone pipeline is docker, and its default platform os is linux.
Explicit specification is not required in principle, but it is helpful and more intuitive for subsequent use of different types of runners and other os, as (ref: https://github.com/ziglang/zig/blob/master/ci/zinc/drone.yml#L3) does.

andrewrk (approved):
OK, good enough for me, thank you for the explanation.

Issue #10706, opened 2022-01-27
reflection: make @typeInfo not resolve declarations


andrewrk:
If you use `@typeInfo` on a struct, you get this structure:

```zig
    pub const Struct = struct {
        layout: ContainerLayout,
        fields: []const StructField,
        decls: []const Declaration,
        is_tuple: bool,
    };
```

Declaration looks like this:

```zig
    pub const Declaration = struct {
        name: []const u8,
        is_pub: bool,
        data: Data,

        pub const Data = union(enum) {
            Type: type,
            Var: type,
            Fn: FnDecl,

            pub const FnDecl = struct {
                fn_type: type,
                is_noinline: bool,
                is_var_args: bool,
                is_extern: bool,
                is_export: bool,
                lib_name: ?[]const u8,
                return_type: type,
                arg_names: []const []const u8,
            };
        };
    };
```

`name` and `is_pub` are cool. But `Data` requires doing a lot of semantic analysis to provide this information. However most usages of `@typeInfo` do not need access to this information.

stage1 solves this problem by making Data lazy. But this is complicated and slow.

This proposal is to **remove the `data` field from Declaration, and add a new builtin function to reflect on declarations**:

`@reflectDecl` - given a type which is used as a namespace, and a decl name, provides meta information about the decl. The type returned is:

```zig
pub const Declaration = struct {
    type: type,
    alignment: u16,
    link_section: [:0]const u8,
};
```

If you then looked at the type and learned that it was a function, you could use `@reflectFunc` to learn information about its function body (not to be confused with its function type which can be learned via `@typeInfo(type).Fn`.

```zig
pub const FunctionBody = struct {
    type: type,
    is_noinline: bool,
    is_extern: bool,
    is_export: bool,
    lib_name: ?[]const u8,
    arg_names: []const []const u8,
};
```

Related:  #10705

nektro:
can this change still preserve the enum tag from `Data` as `Kind` ?

andrewrk:
The Data tagged union in master branch is completely bogus. All decls are `Var`. The type is available with `@TypeOf(@field(T, decl.name))`. There's no such thing as a Type or Fn decl. That's just a decl where typeof gives you .Fn or .Type.

topolarity:
Since `@Type` already disallows creating decls, this change sounds pretty reasonable to me. Status quo reflection (and esp. reification) is mostly for types-as-data-containers, not types-as-namespaces.

Some small thoughts:
[...]

PR #12187, opened 2022-07-21
Implement @reflectDecl builtin

antlilja:
This PR half implements #10706, it implements the builtin ```@reflectRecl```

### DIFF
--- a/lib/std/builtin.zig
+++ b/lib/std/builtin.zig
@@ -183,6 +183,14 @@ pub const SourceLocation = struct {
     column: u32,
 };
 
+/// This data structure is used by the Zig language code generation and
+/// therefore must be kept in sync with the compiler implementation.
+pub const Declaration = struct {
+    ty: type,
+    alignment: comptime_int,
+    link_section: ?[:0]const u8,
+};
+
 pub const TypeId = std.meta.Tag(Type);
 
 /// TODO deprecated, use `Type`
@@ -295,7 +303,7 @@ pub const Type = union(enum) {
     pub const Struct = struct {
         layout: ContainerLayout,
         fields: []const StructField,
-        decls: []const Declaration,
+        decls: []const Type.Declaration,
         is_tuple: bool,
     };
 
@@ -336,7 +344,7 @@ pub const Type = union(enum) {
         layout: ContainerLayout,
         tag_type: type,
         fields: []const EnumField,
-        decls: []const Declaration,
+        decls: []const Type.Declaration,
         is_exhaustive: bool,
     };
 
@@ -354,7 +362,7 @@ pub const Type = union(enum) {
         layout: ContainerLayout,
         tag_type: ?type,
         fields: []const UnionField,
-        decls: []const Declaration,
+        decls: []const Type.Declaration,
     };
 
     /// TODO deprecated use Fn.Param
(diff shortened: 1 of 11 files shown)


lib/std/builtin.zig
```diff
@@ -183,6 +183,14 @@ pub const SourceLocation = struct {
     column: u32,
 };
 
+/// This data structure is used by the Zig language code generation and
+/// therefore must be kept in sync with the compiler implementation.
+pub const DeclarationInfo = struct {
```
Vexu, line 188:
```suggestion
pub const Declaration = struct {
```
To match `Type` below.

src/AstGen.zig
```diff
@@ -7741,6 +7741,16 @@ fn builtinCall(
             });
             return rvalue(gz, rl, result, node);
         },
+        .reflect_decl => {
+            const container_type = try typeExpr(gz, scope, params[0]);
+            const name = try comptimeExpr(gz, scope, .{ .ty = .const_slice_u8_type }, params[1]);
```
Vexu, line 7746:
```suggestion
            const name = try comptimeExpr(gz, scope, .{ .coerced_ty = .const_slice_u8_type }, params[1]);
```

src/Sema.zig
```diff
@@ -17974,6 +17975,101 @@ fn zirWasmMemoryGrow(
...
+    const decl_name = try sema.resolveConstString(block, name_src, extra.rhs);
+
+    const target = sema.mod.getTarget();
+
+    try checkNamespaceType(sema, block, ty_src, namespace_ty);
+
+    // Type must be a namespace because we just checked.
+    const namespace = namespace_ty.getNamespace().?;
+
+    if (try sema.lookupInNamespace(block, name_src, namespace, decl_name, true)) |decl_index| {
```
Vexu, line 17999:
```suggestion
    if (try sema.namespaceLookup(block, name_src, namespace, decl_name)) |decl_index| {
```

Is there a specific reason why you copied the check for `is_pub`?

src/Sema.zig
```diff
@@ -17974,6 +17975,101 @@ fn zirWasmMemoryGrow(
...
+            name_src,
+            "declaration of '{s}' could not be found in type '{s}'",
+            .{ decl_name, namespace_name },
+        );
+        errdefer msg.destroy(sema.gpa);
+
+        try sema.mod.errNoteNonLazy(namespace_ty.declSrcLoc(sema.mod), msg, "declared here", .{});
+        break :msg msg;
+    });
+}
```
Vexu, line 18072: Use `failWithBadMemberAccess` preferably with `orelse` to avoid indenting the rest of the function.

Vexu (changes requested):

lib/std/builtin.zig
antlilja: That would conflict with the the already existing Declaration type inside of Type, but maybe Type.Declaration could be renamed to something else but I'm unsure to what that would be.

Vexu: No need to rename, just use `Type.Declaration` instead of plain `Declaration`.

src/Sema.zig
Vexu: Be sure to delete the now unnecessary `is_pub` check.

Vexu (approved):

Vexu:
I noticed that #10706 is no longer accepted so I'm going to close this. It is still linked in the issue and can be easily reopened if it gets accepted later.

PR #12188, opened 2022-07-21
stage2: llvm - Implement C ABI when targetting wasm32

Luukdegram:
This utilizes the same logic as what the Wasm backend uses, similar to what was done for x86_64. All the C ABI tests are now passing and are enabled for Wasm specifically when stage2 is enabled.


andrewrk (approved):
Wow, great work!

PR #12203, opened 2022-07-23
feat(std.fmt): reject more invalid type specifiers 

r00ster91:
This makes it so that if you try to do `std.debug.print("{x:10}", .{.hello})`, it errors because
trying to format an enum literal as hexadecimal makes no sense. Currently it compiles as-is.
Basically what you're supposed to do instead is, in that case, `"{:10}"` or `"{any:10}"`.
The ":10" part to align the thing of course still works.

This is supposed to catch mistakes like `"{b:10}", .{true}` where maybe the user assumed
"b" means "boolean" but it means "binary". Or maybe they assumed it would print `1` or whatever? It's ambiguous.
It just doesn't seem right to accept `"{u:10}", .{null}` (which currently prints as "null") when something like `"{s:10}", .{5}` isn't accepted either.
Silently ignoring such invalid specifiers can give the user the wrong expectations so we should explicitly reject such invalid specifiers to make sure what's printed actually matches the user's expectations.

During review of the initial iteration of the PR, slightly improving the style of the compile errors here and there happened to become a part of this PR because it seemed convenient. This has been discussed with Vexu. Please review by commit.

### DIFF
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -83,7 +83,7 @@ pub fn format(
     const ArgsType = @TypeOf(args);
     const args_type_info = @typeInfo(ArgsType);
     if (args_type_info != .Struct) {
-        @compileError("Expected tuple or struct argument, found " ++ @typeName(ArgsType));
+        @compileError("expected tuple or struct argument, found " ++ @typeName(ArgsType));
     }
 
     const fields_info = args_type_info.Struct.fields;
@@ -127,7 +127,7 @@ pub fn format(
         if (i >= fmt.len) break;
 
         if (fmt[i] == '}') {
-            @compileError("Missing opening {");
+            @compileError("missing opening {");
         }
 
         // Get past the {
@@ -140,7 +140,7 @@ pub fn format(
         const fmt_end = i;
 
         if (i >= fmt.len) {
-            @compileError("Missing closing }");
+            @compileError("missing closing }");
         }
 
         // Get past the }
@@ -152,7 +152,7 @@ pub fn format(
             .none => null,
             .number => |pos| pos,
             .named => |arg_name| meta.fieldIndex(ArgsType, arg_name) orelse
-                @compileError("No argument with name '" ++ arg_name ++ "'"),
+                @compileError("no argument with name '" ++ arg_name ++ "'"),
         };
 
         const width = switch (placeholder.width) {
@@ -160,8 +160,8 @@ pub fn format(
             .number => |v| v,
             .named => |arg_name| blk: {
                 const arg_i = comptime meta.fieldIndex(ArgsType, arg_name) orelse
-                    @compileError("No argument with name '" ++ arg_name ++ "'");
-                _ = comptime arg_state.nextArg(arg_i) orelse @compileError("Too few arguments");
+                    @compileError("no argument with name '" ++ arg_name ++ "'");
+                _ = comptime arg_state.nextArg(arg_i) orelse @compileError("too few arguments");
                 break :blk @field(args, arg_name);
             },
         };
@@ -171,14 +171,14 @@ pub fn format(
             .number => |v| v,
             .named => |arg_name| blk: {
                 const arg_i = comptime meta.fieldIndex(ArgsType, arg_name) orelse
-                    @compileError("No argument with name '" ++ arg_name ++ "'");
-                _ = comptime arg_state.nextArg(arg_i) orelse @compileError("Too few arguments");
+                    @compileError("no argument with name '" ++ arg_name ++ "'");
+                _ = comptime arg_state.nextArg(arg_i) orelse @compileError("too few arguments");
                 break :blk @field(args, arg_name);
             },
         };
 
         const arg_to_print = comptime arg_state.nextArg(arg_pos) orelse
-            @compileError("Too few arguments");
+            @compileError("too few arguments");
 
         try formatType(
             @field(args, fields_info[arg_to_print].name),
@@ -198,7 +198,7 @@ pub fn format(
         const missing_count = arg_state.args_len - @popCount(ArgSetType, arg_state.used_args);
         switch (missing_count) {
             0 => unreachable,
-            1 => @compileError("Unused argument in '" ++ fmt ++ "'"),
+            1 => @compileError("unused argument in '" ++ fmt ++ "'"),
             else => @compileError((comptime comptimePrint("{d}", .{missing_count})) ++ " unused arguments in '" ++ fmt ++ "'"),
         }
     }
@@ -217,7 +217,7 @@ fn parsePlaceholder(comptime str: anytype) Placeholder {
     // Skip the colon, if present
     if (comptime parser.char()) |ch| {
         if (ch != ':') {
-            @compileError("Expected : or }, found '" ++ [1]u8{ch} ++ "'");
+            @compileError("expected : or }, found '" ++ [1]u8{ch} ++ "'");
         }
     }
 
@@ -252,7 +252,7 @@ fn parsePlaceholder(comptime str: anytype) Placeholder {
     // Skip the dot, if present
     if (comptime parser.char()) |ch| {
         if (ch != '.') {
-            @compileError("Expected . or }, found '" ++ [1]u8{ch} ++ "'");
+            @compileError("expected . or }, found '" ++ [1]u8{ch} ++ "'");
         }
     }
 
@@ -261,7 +261,7 @@ fn parsePlaceholder(comptime str: anytype) Placeholder {
         @compileError(@errorName(err));
 
     if (comptime parser.char()) |ch| {
-        @compileError("Extraneous trailing character '" ++ [1]u8{ch} ++ "'");
+        @compileError("extraneous trailing character '" ++ [1]u8{ch} ++ "'");
     }
 
     return Placeholder{
@@ -423,7 +423,7 @@ pub fn formatAddress(value: anytype, options: FormatOptions, writer: anytype) @T
         else => {},
     }
 
-    @compileError("Cannot format non-pointer type " ++ @typeName(T) ++ " with * specifier");
+    @compileError("cannot format non-pointer type " ++ @typeName(T) ++ " with * specifier");
 }
 
 // This ANY const is a workaround for: https://github.com/ziglang/zig/issues/7948
@@ -608,7 +608,7 @@ pub fn formatType(
                         }
                         return;
                     }
-                    @compileError("Unknown format string: '" ++ actual_fmt ++ "' for type '" ++ @typeName(T) ++ "'");
+                    @compileError("unknown format string: '" ++ actual_fmt ++ "' for type '" ++ @typeName(T) ++ "'");
                 },
                 .Enum, .Union, .Struct => {
                     return formatType(value.*, actual_fmt, options, writer, max_depth);
@@ -630,7 +630,7 @@ pub fn formatType(
                         else => {},
                     }
                 }
-                @compileError("Unknown format string: '" ++ actual_fmt ++ "' for type '" ++ @typeName(T) ++ "'");
+                @compileError("unknown format string: '" ++ actual_fmt ++ "' for type '" ++ @typeName(T) ++ "'");
             },
             .Slice => {
                 if (actual_fmt.len == 0)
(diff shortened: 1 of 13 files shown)


src/link/MachO/Dylib.zig
```diff
@@ -207,7 +207,7 @@ fn readLoadCommands(
                 }
             },
             else => {
-                log.debug("Unknown load command detected: 0x{x}.", .{cmd.cmd()});
+                log.debug("Unknown load command detected: 0x{x}.", .{@enumToInt(cmd.cmd())});
```
r00ster91, line 210: Excellent example here. I'm pretty sure currently this would print as `0xSYMSEG` or whatever because they assumed it would print the integer of the enum but that's not the case. Clearly we want a number here, so now it'll print as `0x3`.

src/print_zir.zig
```diff
@@ -550,7 +550,7 @@ const Writer = struct {
     fn writeElemTypeIndex(self: *Writer, stream: anytype, inst: Zir.Inst.Index) !void {
         const inst_data = self.code.instructions.items(.data)[inst].bin;
         try self.writeInstRef(stream, inst_data.lhs);
-        try stream.print(", {d})", .{inst_data.rhs});
+        try stream.print(", {})", .{inst_data.rhs});
```
r00ster91, line 553: Not entirely sure what this should be. AFAIK `inst_data.rhs` here is an enum and the specifier was `{d}` and the function says it writes the element type index, so should this be `", {d})", .{@enumToInt(inst_data.rhs)}` instead?

lib/std/fmt.zig
```diff
@@ -444,6 +444,15 @@ fn defaultSpec(comptime T: type) [:0]const u8 {
     return "";
 }
 
+// This errors when we're given an explicit type specifier (i.e. not "any" or nothing).
+fn checkNoSpecifier(comptime thing: []const u8, comptime actual_fmt: []const u8) void {
+    if (actual_fmt.len != 0)
+        @compileError(comptime comptimePrint(
+            \\cannot format {s} with specifier '{s}'
+            \\try removing the specifier
+        , .{ thing, actual_fmt }));
```
Vexu, line 453: This should match other similar errors so:
```zig
"Unsupported format string '" ++ fmt ++ "' for type '" ++ @typeName(@TypeOf(value)) ++ "'"
```

src/print_zir.zig
Vexu: In this case `rhs` is used as a plain integer index so `@enumToInt` here is correct.

lib/std/fmt.zig
r00ster91: Are you talking about the formatted message or the code here?
Is it bad to use `comptimePrint` here because I think it's much more readable than those long `++` one-liners and it did not cause any problems for me when I tested this error with `comptimePrint` itself.
I do agree that the usage of a newline here is at least currently inconsistent with all the other messages because people just aren't used to the fact that they can use multiline strings now and they'll look pretty. `comptimePrint` doesn't seem to cause problems either. But that's just my style.
So if you insist, I can change it of course, but I would be interested in your opinion on this.

I chose to give the user a hint here (`"try removing the specifier"`) because here I'm in a pretty good position to say that removing the specifier will probably fix your problem. However looking at the mistakes I found with this, removing the specifier wasn't always the solution, so maybe it's still better to just remove it... yeah I will just do that.

If you're talking about the formatted string itself, that was kind of intentional because to be honest I never quite agreed with capitalizing a `@compileError` message because usually all errors are entirely in lowercase and the compiler does it that way too. It would only be consistent.
And we already have such messages elsewhere anyway: https://github.com/ziglang/zig/blob/0df702d81a4d5e6b478a44b4037c5b62c6c2a2d7/lib/std/fmt.zig#L587

Vexu: I was talking about the content of the message, I think it is bad to have multiple different messages for the same kind of error, `comptimePrint` or `++` are either fine though `++` is much simpler.

I also prefer compile errors to be lowercased, I just directly copied the error found elsewhere, it would be better for those other uses to also be lowercase, or better yet, call this new function.

r00ster91: Ah yes, I understand now. You're right! I refactored it and made it into a more general function to be used for such errors.

r00ster91: @Vexu while I'm at it can I push a commit to this PR that lowercases all compile errors in just this file?

Vexu: Fine by me.

lib/std/fmt.zig
```diff
@@ -586,7 +604,7 @@ pub fn formatType(
                         }
                         return;
                     }
-                    @compileError("Unknown format string: '" ++ actual_fmt ++ "' for type '" ++ @typeName(T) ++ "'");
+                    @compileError("unknown format string: '" ++ actual_fmt ++ "' for type '" ++ @typeName(T) ++ "'");
```
Vexu, line 607: This should use `invalidFmtErr`.

lib/std/fmt.zig
```diff
@@ -608,7 +626,7 @@ pub fn formatType(
                         else => {},
                     }
                 }
-                @compileError("Unknown format string: '" ++ actual_fmt ++ "' for type '" ++ @typeName(T) ++ "'");
+                @compileError("unknown format string: '" ++ actual_fmt ++ "' for type '" ++ @typeName(T) ++ "'");
```
Vexu, line 629: Ditto

r00ster91:
This is definitely clashing with #12160. Will wait for that to be merged before I fix the remaining errors here. The fixes are the same as in that PR in many places.

r00ster91:
@Vexu the remaining fixes in my latest commit (after the force-push) were mostly guesswork. I was especially struggling with this one: https://github.com/ziglang/zig/blob/master/src/link/MachO.zig#L290
Maybe you know more about that. Or probably kubkon. I just left all behavior as it was right now. It's mostly just debug logging anyway.

I don't know why InKryption didn't hit these but I guess they were all new additions or something.

Vexu:
> Vexu the remaining fixes in my latest commit (after the force-push) were mostly guesswork. I was especially struggling with this one: https://github.com/ziglang/zig/blob/master/src/link/MachO.zig#L290 Maybe you know more about that. Or probably kubkon. I just left all behavior as it was right now. It's mostly just debug logging anyway.

I didn't care too much for debug logs when doing the original fixups either, kubkon can fix those properly later if he wants to.

> I don't know why InKryption didn't hit these but I guess they were all new additions or something.

I think a lot of these came from #12140 which was merged after InKryption's PR was done.

src/print_zir.zig
Vexu: This regressed at some point.

r00ster91:
That makes sense! Yeah it's nothing critical.

r00ster91:
@Vexu yeah I agree with inlining the `if`; looks better.
The change in the `.Enum => |enumInfo| {` prong though, where you [remove](https://github.com/ziglang/zig/compare/63976c03f3cf1132da8da08571118e893b0760ed..2f669adf3766a73340741a4ed15d120bd2a12e8f#diff-b2dd3e0e31147e6adc31a9099c0967f22e1c5fa2f6bdd7a65b00528e9bab976fL538-L539) the `if (actual_fmt.len != 0) invalidFmtErr(fmt, value);` in the `inline for (enumInfo.fields) |enumField| {`, it was there for this case of a non-exhaustive enum:

```zig
const std = @import("std");

const X = enum(u8) { A, _ };

pub fn main() void {
    std.debug.print("^ ('\\x{x:0>2}')\n", .{X.A});
}
```
Now it doesn't emit an error for this anymore.

If I add the `if (actual_fmt.len != 0) invalidFmtErr(fmt, value);` back there, it fixes that.

r00ster91:
Also, if I run the tests for `fmt.zig` locally, I get this failure:
```
enum: os.windows.win32error.Win32Error.INVALID_FUNCTION

enum: std.os.windows.win32error.Win32Error.INVALID_FUNCTION

```
(which can simply be fixed by adding "std." to the expected string)
and I'm not sure why it's happening now because this PR shouldn't change such behavior.

Vexu:
> Vexu yeah I agree with inlining the `if`; looks better. The change in the `.Enum` prong though, where you [removed](https://github.com/ziglang/zig/compare/63976c03f3cf1132da8da08571118e893b0760ed..2f669adf3766a73340741a4ed15d120bd2a12e8f#diff-b2dd3e0e31147e6adc31a9099c0967f22e1c5fa2f6bdd7a65b00528e9bab976fL538-L539) the `if (actual_fmt.len != 0) invalidFmtErr(fmt, value)` in the `inline for (enumInfo.fields)`, it was there for this case of a non-exhaustive enum:
> 
> Now it doesn't emit an error for this anymore.
> 
> If I add the `if (actual_fmt.len != 0) invalidFmtErr(fmt, value);` back there, it fixes that.

You can't disable the format string for only the named variants of a non-exhaustive enum and disabling them alltogether breaks the test `"non-exhaustive enum"`.

> Also, if I run the tests for fmt.zig locally, I get this failure:

That is a stage1/stage2 incompatibility, if you run the tests using stage2 as the CI does then the test passes, though it should be disabled for stage1 if it doesn't pass.

r00ster91:
> You can't disable the format string for only the named variants of a non-exhaustive enum and disabling them alltogether breaks the test "non-exhaustive enum".

I'm sorry, I'm not sure I entirely understand yet. Why should this pass?
```zig
const std = @import("std");
const Enum = enum(u16) { One = 0x000f, Two = 0xbeef, _ };
pub fn main() void {
    std.debug.print("{x}\n", .{Enum.One});
}
```
But not if it's exhaustive? if you remove `_`, it errors. I think it should error for non-exhaustive enums too (with `_`). Is there any case where printing the name of the tag of a non-exhaustive enum as e.g. hexadecimal makes sense?

The way I placed the `invalidFmtErr`s was supposed to make it so that only in the last case of `formatType(@enumToInt(value), actual_fmt, options, writer, max_depth)`, it accepts formats like `x` because in that case it's actually an integer. In all other cases, it formats it as the tag name right?

Vexu:
https://github.com/ziglang/zig/blob/a08b0fa706f79a506b219d59a0b58a6d5761dcd7/lib/std/fmt.zig#L2242

This test wont pass with that in place.

>The way I placed the `invalidFmtErr`s was supposed to make it so that only in the last case of `formatType(@enumToInt(value), actual_fmt, options, writer, max_depth)`, it accepts formats like `x` because in that case it's actually an integer. In all other cases, it formats it as the tag name in which anything other than empty or `any` doesn't make sense, right?

If you use `@enumToInt` you no longer hvae an enum and you wont hit that code path anyways.

r00ster91:
@andrewrk regarding your revert, for example in https://github.com/ziglang/zig/blob/master/src/link/MachO/Atom.zig#L544-L549 the optional is `rel.target.file` (https://github.com/ziglang/zig/blob/master/src/link/MachO.zig#L290).
It's a bit complicated but basically as @Vexu said,

> I think a lot of these came from https://github.com/ziglang/zig/pull/12140 which was merged after @InKryption's PR was done.

And now the fixes for these errors is part of my PR. But this PR depended on #12160. Can we get #12160 merged again and then this PR? As you can see, with this PR all CI checks pass again.

r00ster91:
One problem I found though is this:
```zig
const std = @import("std");

fn a() error{a}![]const u8 {
    return "hello";
}

pub fn main() void {
    std.debug.print("{!s}\n", .{a()});
}
```
This should work but with this PR, it errors. It works if you remove `error{a}`. I've fixed this by temporarily removing the check there.

Vexu:
> This should work but with this PR, it errors. It works if you remove `error{a}`. I've fixed this by temporarily removing the check there.

`stripOptionalOrErrorUnionSpec` should handle that in the `.ErrorUnion` prong.

Vexu:
Actually it should be fixed by just passing an empty string to the error branch of the if:
```diff
diff --git a/lib/std/fmt.zig b/lib/std/fmt.zig
index 62fd018a9..0df57f947 100644
--- a/lib/std/fmt.zig
+++ b/lib/std/fmt.zig
@@ -510,7 +510,7 @@ pub fn formatType(
             if (value) |payload| {
                 return formatType(payload, remaining_fmt, options, writer, max_depth);
             } else |err| {
-                return formatType(err, remaining_fmt, options, writer, max_depth);
+                return formatType(err, "", options, writer, max_depth);
             }
         },
         .ErrorSet => {
```

r00ster91:
@Vexu you're right; tests pass. Applied the patch.

r00ster91:
Conflicts resolved. Tests pass.

PR #12207, opened 2022-07-24
Hacked timed wait for stderr_mutex

### LINKED ISSUE #7859: Error messages should use a mutex

arBmind:
Fixes #7859

* to avoid unreadable error messages, we should synchronize on stderr_mutex
* to avoid deadlock situations we give up after timeout
* mutex has no timedLock() so we emulate with retry logic

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -1939,6 +1939,7 @@ fn handleSegfaultPosix(sig: i32, info: *const os.siginfo_t, ctx_ptr: ?*const any
     };
 
     // Don't use std.debug.print() as stderr_mutex may still be locked.
+    _ = timedWaitStdErrMutex(); // os.abort() is called - don't bother to unlock the mutex
     nosuspend {
         const stderr = io.getStdErr().writer();
         _ = switch (sig) {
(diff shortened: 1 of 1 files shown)


kprotty:
It's been a while for this PR but a few notes:

Is the timedLock there to handle the case of the current thread already holding the `stderr_mutex` when inside the segfault signal handler? If so, what's required is a recursive mutex instead of a `timedLock()` as the latter can allow unrelated threads through and still possibly have the stderr writing race.

For sleeping you should use `std.time.sleep`, but you shouldn't really implement timedLock() using sleep due to extra latency on contention. I think a proper solution here is to implement a recursive mutex check around `stderr_mutex` by storing the thread id:

```zig
var stderr_mutex = std.Thread.Mutex{};
var stderr_tid = std.atomic.Atomic(usize).init(0);

fn lockStdErrMutex() void {
    const tid = switch (builtin.os.tag) {
        .windows => std.os.windows.kernel32.GetCurrentThread(),
        .linux => @intCast(usize, std.Thread.getCurrentId()),
        else => @ptrToInt(std.c.pthread_self())
    };
    std.debug.assert(tid != 0);
    if (stderr_tid.load(.Unordered) == tid) return; // skip locking if we own it
    stderr_mutex.lock();
    stderr_tid.store(tid, .Unordered); // unordered stores to avoid data-race with others reading it above
}

fn unlockStdErrMutex() void {
    assert(stderr_tid.loadUnchecked() != 0); // unchecked load as unlock from non-owner is allowed to be UB
    stderr_tid.store(0, .Unordered);
    stderr_mutex.unlock();
}
```

Vexu (changes requested):

andrewrk:
Superseded by #15214.

PR #12214, opened 2022-07-24
`std.testing`: make the caret indicator line more helpful

r00ster91:
Same as #11766 but without the messed up history (thanks git).

### DIFF
--- a/lib/std/testing.zig
+++ b/lib/std/testing.zig
@@ -542,7 +542,7 @@ fn printIndicatorLine(source: []const u8, indicator_index: usize) void {
     if (indicator_index >= source.len)
         print("^ (end of string)\n", .{})
     else
-        print("^ ('\\u{{{x}}}')\n", .{source[indicator_index]});
+        print("^ ('\\x{x:0>2}')\n", .{source[indicator_index]});
 }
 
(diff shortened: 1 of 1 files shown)


lib/std/testing.zig
```diff
@@ -539,7 +539,10 @@ fn printIndicatorLine(source: []const u8, indicator_index: usize) void {
         while (i < indicator_index) : (i += 1)
             print(" ", .{});
     }
-    print("^\n", .{});
+    if (indicator_index >= source.len)
+        print("^ (end of string)\n", .{})
+    else
+        print("^ ({d})\n", .{source[indicator_index]});
```
squeek502, line 545: Might be nice to format it as a character literal:

```suggestion
        print("^ ('\\x{x:0>2}')\n", .{source[indicator_index]});
```

This would give something like:

```
^ ('\x00')
```

r00ster91: Alright, yeah, maybe that can help if the user wants to play around with it or something. I will use `\u{...}` so that it works with any character.

r00ster91: Nevermind, it doesn't handle Unicode characters. Changed it to your original suggestion!

PR #12219, opened 2022-07-24
add tcdrain on linux

yujiri8:
From issue #10181. @cweagans can you evaluate if I'm doing this the right way? I found which ioctl it is from reading musl source.


cweagans:
I am very unqualified to make that evaluation 😅 It looks generally right, but you should definitely defer to somebody more knowledgeable than me!

yujiri8:
Erm, I have no idea. I just read the man page and it didn't sound like it requires any arguments.

I went ahead and rebased adding the 1 argument. No idea what it does, but I guess I ought to trust the musl authors :sweat_smile:

PR #12220, opened 2022-07-24
LLVM: notice the soft_float CPU feature

### LINKED ISSUE #10847: x86_64 +softfloat crashes LLVM
## First bad commit
9bbd3ab

## Steps to reproduce
Attempt to build [Zigvale Barebones](https://github.com/ominitay/zigvale-barebones) or [Florence OS](https://github.com/FlorenceOS/Florence/).

## Expected Behaviour
The projects build successfully. 

## Actual Behaviour
The compiler crashes in release mode, or segfaults in debug mode. 

cc: @N00byEdge 
cc: @Vexu

andrewrk:
when deciding whether to lower to x86_fp80 instructions, or softfloat
instructions.

Closes #10847
Closes #12090

### DIFF
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -9514,7 +9514,7 @@ fn isByRef(ty: Type) bool {
 /// and false if we expect LLVM to crash if it counters an x86_fp80 type.
 fn backendSupportsF80(target: std.Target) bool {
     return switch (target.cpu.arch) {
-        .x86_64, .i386 => true,
+        .x86_64, .i386 => !std.Target.x86.featureSetHas(target.cpu.features, .soft_float),
         else => false,
     };
 }


leecannon:
It doesn't seem like this has fixed the simplest case
```
$ echo "pub fn main() void {}" > simple.zig

$ # stage 1 binary with -fno-stage1
$ zig_stage1 version
0.10.0-dev.3292+d761fb535
$ zig_stage1 build-exe simple.zig -mcpu=x86_64+soft_float -fno-stage1 -fLLVM
LLVM Emit Object... [1]    1320223 segmentation fault (core dumped)  zig build-exe simple.zig -mcpu=x86_64+soft_float -fno-stage1 -fLLVM

$ # stage 2 binary
$ zig_stage2 version
0.10.0-dev.3292+d761fb535
$ zig_stage2 build-exe simple.zig -mcpu=x86_64+soft_float -fno-stage1 -fLLVM
Semantic Analysis [1243] mulWide... LLVM Emit Object... thread 1323404 panic: Segmentation fault at address 0x5eb37c9
???:?:?: 0x5eb3844 in ??? (???)
[1]    1323404 IOT instruction (core dumped)  ~/src/zig/stage2/bin/zig build-exe simple.zig -mcpu=x86_64+soft_float  -fLLVM
```

Vexu:
Checking for `soft_float` in `target_has_f80` was one of the first things I tried and it didn't seem to fix the issue in stage1 either but that might have been because it ignored `f128`.

ominitay:
@leecannon Are you sure? I get an error that there's no main function with an empty source file, and a TODO compile error in start.zig with your test, which is exactly the expected behaviour, and no longer a crash. Stage1 still crashes ofc

andrewrk:
The issue title has "x86_64 freestanding" and the simplest case in the issue is

```
zig build-exe empty.zig -target x86_64-freestanding-none -mcpu=x86_64+soft_float-sse2-sse-x87
```

which is indeed fixed by this patch. 

Thanks for providing another test case. As far as I can see this is the first time this test case has been posted to any of the related issues. I'll add another commit to handle this case as well.

andrewrk:
Actually, even @leecannon's case above is handled when the proper CPU feature set is provided:

```
[nix-shell:~/dev/zig/build-release]$ stage3/bin/zig build-exe simple.zig -mcpu=x86_64+soft_float
LLVM Emit Object... thread 424474 panic: Illegal instruction at address 0x5c6d154
???:?:?: 0x5c6d154 in ??? (???)
^CAborted (core dumped)

[nix-shell:~/dev/zig/build-release]$ stage3/bin/zig build-exe simple.zig -mcpu=x86_64+soft_float-x87-sse2-sse

```

So perhaps all we need at this point is a new error that detects conflicting CPU features and emits an error if you turn on soft float but don't disable x87.

leecannon:
Looks like it all works with `-mcpu=x86_64+soft_float-x87-sse2-sse` :+1:

andrewrk:
> So perhaps all we need at this point is a new error that detects conflicting CPU features and emits an error if you turn on soft float but don't disable x87.

Extracted to  #12227.

PR #12228, opened 2022-07-25
remove 'builtin.stage2_arch', Sema is smart enough now

nektro:
fixes TODO


nektro:
hmmm maybe not

```
/runner/work/1/s/lib/std/target.zig:727:25: error: TODO MCValues with multiple registers
/runner/work/1/s/lib/std/target.zig:651:21: error: TODO support more return types for ARM backend
/runner/work/1/s/lib/std/target.zig:668:21: error: TODO implement div for std.target.Target.Cpu.Arch.aarch64
```

andrewrk:
Sema is fine, it's the non-llvm backends struggling to lower one or more of the functions inside `lib/std/Target.zig`, which is being emitted for debug info reasons. There is a path to merging this which is to run one of the failing test cases in a debugger, which should be as simple as cross compiling the behavior tests for arm, when you hit the todo, look at which function is being compiled. Then go edit that function body and put a workaround in for that target architecture.

andrewrk:
Behavior tests and test-cases passed locally.

PR #12235, opened 2022-07-25
feat: add `reset` to `SplitBackwardsIterator` and `SplitIterator`

r00ster91:
I just had a case where I wanted to reuse a `SplitIterator` but couldn't find a `reset` on it even though `TokenIterator` (a close relative to `SplitBackwardsIterator` and `SplitIterator`) has one. You could argue that what `reset` does is too trivial to have a method for it but for `SplitBackwardsIterator` the implementation is a bit less trivial and a bit more specific. And, again, `TokenIterator` has `reset` too.

Alternatively, I think we should remove `reset` from `TokenIterator`.

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -1779,6 +1779,20 @@ test "split (multibyte)" {
     try testing.expect(it16.next() == null);
 }
 
+test "split (reset)" {
+    var it = split(u8, "abc def ghi", " ");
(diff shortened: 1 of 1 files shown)


Vexu (approved):
A test would be nice.

r00ster91:
I agree.
In that case I will actually wait for #12205 because then I can use `first` in the tests.

PR #12237, opened 2022-07-25
Stage2 improve errors for builtin function options structs

### LINKED ISSUE #8262: Explicitly forbid `\x00` (null bytes) in `@""` identifier syntax
Some object formats, such as ELF, rely on null bytes to terminate strings, making it impossible to represent symbol names which contain null bytes. So there already exist at least some targets for which this must be rejected. This proposal is to unconditionally reject null bytes in identifiers for all targets.

Additionally, third party tools which attempt to integrate with Zig code or the Zig compiler may want to use null terminated strings in order to represent identifier names. By explicitly making it illegal to have null bytes in @"" identifier strings, this prevents possible bugs in the wild having to do with analysis of Zig code.

If this is accepted, it means the self-hosted compiler can sometimes use null termination to save bytes of memory in ZIR. For example an enum literal or field access could be a u32 index into a string table, rather than a index and a length. When the shrinkage from 8 bytes to 4 bytes allows an instruction to fit inside the 8 bytes allocated for every IR instruction rather than indexing into an extra_data array, this ends up saving an additional 4 bytes, for a total saving of 8 bytes per relevant instruction.

### LINKED ISSUE #12230: Any `export` called `@""` is not exported
### Zig Version

0.10.0-dev.2836+2360f8c49

### Steps to Reproduce

Exporting anything identified by `@""` exports nothing. on wasm32, x86_64, aarch64, probably all platforms.

```zig
export const @"": u16 = 12345;
```

### Expected Behavior

`@""` is rejected as an identifier for `export`s. It's simply not representable. What is the compiler supposed to call it?

Additionally, any opinions on disallowing `@""` as an identifier in general? Are there any use cases? It is inexpressive. What is `@""` even supposed to mean? Disallowing it altogether might be more consistent. (nvm, I don't think this is good actually, could be missing some edge cases etc.)

See also: #8262

## stage1

Exporting anything identified by `@""` exports nothing. on wasm32, x86_64, aarch64, probably all platforms.

```zig
export const @"": u16 = 12345;
```
```wat
$ zig build-lib -dynamic -target wasm32-freestanding -OReleaseSmall --strip x.zig; wasm2wat x.wasm
(module
  (memory (;0;) 1)
  (global (;0;) (mut i32) (i32.const 65536))
  (export "memory" (memory 0)))
```
12345 is nowhere to be found. Change `@""` to `@"x"` and it's included.

```
$ zig build-lib -dynamic -target x86_64-freestanding -OReleaseSmall --strip x.zig -femit-asm; cat x.s
	.text
	.intel_syntax noprefix
	.file	"x"
	.section	".note.GNU-stack","",@progbits
```
12345 is nowhere to be found. Change `@""` to `@"x"` and it's included.

## stage2

```
$ zig build-lib -dynamic -target wasm32-freestanding -OReleaseSmall --strip -fno-stage1 x.zig; wasm2wat x.wasm
(module
  (memory (;0;) 2)
  (global (;0;) (mut i32) (i32.const 65536))
  (global (;1;) i32 (i32.const 65536))
  (export "memory" (memory 0))
  (export "test." (global 1))
  (data (;0;) (i32.const 65536) "90"))
```

In stage2 it *is* somehow exported as `"test."`, which is very weird.
Additionally if you add `-femit-asm` to this one, it doesn't work. It works without `-fno-stage1`.

All this is unexpected.

### LINKED ISSUE #12233: Compilation error with ordinary comment before test code.
### Zig Version

0.10.0-dev.2859+0101a5f75

### Steps to Reproduce

```Zig
const std = @import("std");

/// safe
pub fn main() anyerror!void {
    // safe
    std.log.info("Ok.", .{});
}

/// THIS COMMENT CAUSES COMPILATION ERROR! > error: invalid token: 'test'
test "basic test" {
    try std.testing.expectEqual(10, 3 + 7);
}
```

### Expected Behavior

No compilation errors.

### Actual Behavior

Compilation error occured.
> error: invalid token: 'test'

So, `//` causes no errors.

```zig
// NO ERRORS.
test "basic test" {
    try std.testing.expectEqual(10, 3 + 7);
}
```

Vexu:
Closes #12233
Closes #12230
Closes #8262

### DIFF
--- a/lib/std/zig/Ast.zig
+++ b/lib/std/zig/Ast.zig
@@ -297,6 +297,12 @@ pub fn renderError(tree: Ast, parse_error: Error, stream: anytype) !void {
         .unattached_doc_comment => {
             return stream.writeAll("unattached documentation comment");
         },
+        .test_doc_comment => {
+            return stream.writeAll("documentation comments cannot be attached to tests");
+        },
+        .comptime_doc_comment => {
+            return stream.writeAll("documentation comments cannot be attached to comptime blocks");
+        },
         .varargs_nonfinal => {
             return stream.writeAll("function prototype has parameter after varargs");
         },
@@ -2539,6 +2545,8 @@ pub const Error = struct {
         invalid_bit_range,
         same_line_doc_comment,
         unattached_doc_comment,
+        test_doc_comment,
+        comptime_doc_comment,
         varargs_nonfinal,
         expected_continue_expr,
         expected_semi_after_decl,
--- a/lib/std/zig/parse.zig
+++ b/lib/std/zig/parse.zig
@@ -259,6 +259,9 @@ const Parser = struct {
 
             switch (p.token_tags[p.tok_i]) {
                 .keyword_test => {
+                    if (doc_comment) |some| {
+                        try p.warnMsg(.{ .tag = .test_doc_comment, .token = some });
+                    }
                     const test_decl_node = try p.expectTestDeclRecoverable();
                     if (test_decl_node != 0) {
                         if (field_state == .seen) {
@@ -317,6 +320,9 @@ const Parser = struct {
                         }
                     },
                     .l_brace => {
+                        if (doc_comment) |some| {
+                            try p.warnMsg(.{ .tag = .test_doc_comment, .token = some });
+                        }
                         const comptime_token = p.nextToken();
                         const block = p.parseBlock() catch |err| switch (err) {
                             error.OutOfMemory => return error.OutOfMemory,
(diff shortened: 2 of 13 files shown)


test/cases/compile_errors/invalid_identifeirs.zig
```diff
@@ -0,0 +1,33 @@
+extern "" var a: u32;
```
tiehuis, line 1: Small typo on this filename. Should be "invalid_identifiers".

PR #12241, opened 2022-07-26
bit_set: Make DynamicBitSetUnmanaged take an int type

N00byEdge:
This allows you to create a `DynamicBitSetUnmanaged` with a different backing integer type, useful for when being passed for example a `[*]u32` or `[*]u8` from an external API with no `usize` alignment gurantees. The `null` argument uses the old behaviour.


SpexGuy:
I don't think the int type should be an optional.  We should instead expose a constant declaration like `pub const DynamicBitSetDefaultUnmanaged = DynamicBitSetUnmanaged(usize);`.  This should also be done for the managed version.

N00byEdge:
@SpexGuy Would you prefer `pub const DynamicBitSetUnmanaged = DynamicBitSetCustomUnmanaged(usize);`?

N00byEdge:
As I think you may be sleeping, I will just implement that API since I think it's a lot nicer, but feel free to disagree and I can swap the names around

N00byEdge:
I went with Spex' names instead after Luuk chimed in

N00byEdge:
That CI fail looks unrelated, right?

SpexGuy:
I like your names better actually, especially if DynamicBitSet will continue to be a single type and not generic.  But I'm not strongly opinionated about it.

N00byEdge:
Yeah, I considered not breaking the API as a decently sized plus as most users won't care about this, but I'm not sure compat for the sake of compat is even worth considering before 1.0.

SpexGuy:
True but it's not just for the sake of compat.  Customizing the int type is a rare use case, the default should be usize.  So that version should have a shorter name.

N00byEdge:
Should I change it back then?

SpexGuy:
IMO yes but I haven't seen Luuk's opinion/reasoning.

N00byEdge:
I don't want to spend too much time on bikeshedding here, I changed the names back to what I think was better.

andrewrk:
I apologize for not getting to this earlier, but it now needs a rebase and to run additional CI checks. Please re-open against master if you want to see this through.

N00byEdge:
I don't have a need for this anymore, so anyone else who needs it can feel free to pick this up again if they want to.

PR #12245, opened 2022-07-26
stage2: add error for comptime control flow in runtime block

Vexu:
AstGen contains a bunch of hacks that I haven't fixed yet since I don't expect this solution to pass on the first try.

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -1940,6 +1940,9 @@ fn continueExpr(parent_gz: *GenZir, parent_scope: *Scope, node: Ast.Node.Index) 
                     .break_inline
                 else
                     .@"break";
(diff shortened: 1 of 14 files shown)


andrewrk:
Are you not able to test locally?

Vexu:
I can, I just don't like all the mess created by adding source nodes to all breaks especially since the only place it seems to be needed is in continues inside inline loops.

Vexu:
I did the now obvious thing of copying stage1 and adding a new instruction for this.

topolarity:
FWIW, I think many of these examples can be supported if the code following the runtime control flow is treated as a runtime block.

I don't have a use case in mind - just wanted to mention there's probably room to relax this if we feel the need to later

PR #12246, opened 2022-07-26
std.mem.zeroes: Zero sized structs with uninitialized members

N00byEdge:
`std.mem.zeroes(struct{handle: void})` Failed with the following error before:
```
/nix/store/l6v4359wc9xrxxmvvp3rynsb5s3d78xf-zig-0.9.1/lib/zig/std/mem.zig:270:42: error: missing field: 'handle'
            if (@sizeOf(T) == 0) return T{};
                                         ^
```

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -425,7 +425,7 @@ test "zeroes" {
     comptime var comptime_union = zeroes(C_union);
     try testing.expectEqual(@as(u8, 0), comptime_union.a);
 
-    // https://github.com/ziglang/zig/pull/12246
+    // Ensure zero sized struct with fields is initialized correctly.
     _ = zeroes(struct { handle: void });
 }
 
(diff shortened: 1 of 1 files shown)


lib/std/mem.zig
```diff
@@ -424,6 +424,9 @@ test "zeroes" {
 
     comptime var comptime_union = zeroes(C_union);
     try testing.expectEqual(@as(u8, 0), comptime_union.a);
+
+    // https://github.com/ziglang/zig/pull/12246
```
matu3ba, line 428: 1. I dont understand what this comment is for.
2. Do you think it would make sense to also add test coverage for the recursive struct case (struct_info.layout == .Extern and the other branch)?

N00byEdge: Oh I was just thinking it motivates the test because it otherwise looks useless...

PR #12256, opened 2022-07-27
stage2 typeInfo UAF fix + more

Vexu:


test/cases/fn_typeinfo_passed_to_comptim_fn.zig
```diff
@@ -0,0 +1,17 @@
+const std = @import("std");
```
r00ster91, line 1: I think there might be a small typo on the filename here as well. Did you mean `fn_typeinfo_passed_to_comptime_fn.zig`?

PR #12270, opened 2022-07-28
Dwarf: Added stroffsetsptr support

### LINKED ISSUE #12120: std.debug: DWARFv5 debug info sometimes causes a failed stack trace
### Zig Version

0.10.0-dev.2993+0248073b9

### Steps to Reproduce

```zig
// build.zig
    const exe = b.addExecutable("test", "test.zig");
    exe.addCSourceFile("test.c", &[_][]const u8{
        "-std=c99",
        "-gdwarf-5", // Bad stack trace
        //"-gdwarf-4", // works
    });
    exe.install();

    const run_cmd = exe.run();
    if (b.args) |args| {
        run_cmd.addArgs(args);
    }
}
```
```zig
// test.zig
extern fn baz(x: c_int) c_int;

pub fn main() void {
    _ = baz(15);
}
```
```c
// test.c
int foo(int x) {
	*(int*)0 = 10; // Intentionally trigger SEGFAULT
	return x + 2;
}
int bar(int x) {
	return foo(x + 16);
}
int baz(int x) {
	return 2 * bar(x);
}
```

### Expected Behavior

```
$ ../zig/build/zig build run
Illegal instruction at address 0x205211
/home/topolarity/repos/test/test.c:2:11: 0x205211 in foo (/home/topolarity/repos/test/test.c)
        *(int*)0 = 10; // Intentionally trigger SEGFAULT
          ^
/home/topolarity/repos/test/test.c:7:9: 0x20526c in bar (/home/topolarity/repos/test/test.c)
        return foo(x + 16);
        ^
/home/topolarity/repos/test/test.c:11:13: 0x205292 in baz (/home/topolarity/repos/test/test.c)
        return 2 * bar(x);
            ^
/home/topolarity/repos/test/test.zig:4:12: 0x227a0d in main (test)
    _ = baz(15);
           ^
/home/topolarity/repos/zig/lib/std/start.zig:570:22: 0x207033 in std.start.posixCallMainAndExit (test)
            root.main();
                     ^
/home/topolarity/repos/zig/lib/std/start.zig:338:5: 0x206df2 in std.start._start (test)
    ^
The following command terminated unexpectedly:
cd /home/topolarity/repos/test && /home/topolarity/repos/test/zig-out/bin/test
/home/topolarity/repos/test/zig-cache/o/c6e3ef0111461e363fabeadc1273799f/build /home/topolarity/repos/zig/build/zig /home/topolarity/repos/test/home/topolarity/repos/test/zig-cache /home/topolarity/.cache/zig run
```

### Actual Behavior

```
$ ../zig/build/zig build run
Illegal instruction at address 0x205211
???:?:?: 0x205211 in ??? (???)
???:?:?: 0x20526c in ??? (???)
???:?:?: 0x205292 in ??? (???)
???:?:?: 0x227a0d in ??? (???)
???:?:?: 0x207033 in ??? (???)
???:?:?: 0x206df2 in ??? (???)
The following command terminated unexpectedly:
cd /home/topolarity/repos/test && /home/topolarity/repos/test/zig-out/bin/test
/home/topolarity/repos/test/zig-cache/o/2117e822669df035365d51266c2c619d/build /home/topolarity/repos/zig/build/zig /home/topolarity/repos/test/home/topolarity/repos/test/zig-cache /home/topolarity/.cache/zig run
```

kdchambers:
Dwarf FORM types strx1, strx2, strx3 and strx4 are now handled. This (for me) fixes  #12120 

However, I'm unsure if `getStringOffset` is correct as I'm unsure how to test it. At the very least it won't cause a regression as the current behaviour is that error.InvalidDebugInfo is returned when a strx* FORM is encountered.


topolarity:
> However, I'm unsure if `getStringOffset` is correct as I'm unsure how to test it. At the very least it won't cause a regression as the current behaviour is that error.InvalidDebugInfo is returned when a strx* FORM is encountered.

Maybe it'd be possible to find an attribute uses the new forms and then check the string against the contents of `llvm-dwarfdump`?

kubkon:
I believe I have created a merge conflict for you - apologies! It should be trivial to adjust to the new MachO linker, and as a bonus, I've completely removed the unneeded `DebugInfo` structure.

kdchambers:
@kubkon  No worries at all, I've have a look at your changes and rebase. Unfortunately I've been very busy this week but hopefully can spare some time this weekend to finish and get it merge-able.

kdchambers:
This is ready to be reviewed and merged. I managed to test `getStringOffset` and it seems to be returning the correct strings.

lib/std/debug.zig
```diff
@@ -846,6 +846,7 @@ fn readCoffDebugInfo(allocator: mem.Allocator, coff_file: File) !ModuleDebugInfo
                 .debug_info = debug_info_data orelse return error.MissingDebugInfo,
                 .debug_abbrev = debug_abbrev_data orelse return error.MissingDebugInfo,
                 .debug_str = debug_str_data orelse return error.MissingDebugInfo,
+                .debug_str_offsets = null,
```
topolarity, line 849: Would it be possible for you to add COFF support for this segment, or else a more specific error or a useful panic?

I'd like to avoid reporting `error.InvalidDebugInfo` when the real problem is that the standard library support is incomplete - Makes it much quicker to root cause 🙂

lib/std/dwarf.zig
```diff
@@ -522,6 +527,10 @@ fn parseFormValue(allocator: mem.Allocator, in_stream: anytype, form_id: u64, en
 
         FORM.string => FormValue{ .String = try in_stream.readUntilDelimiterAlloc(allocator, 0, math.maxInt(usize)) },
         FORM.strp => FormValue{ .StrPtr = try readAddress(in_stream, endian, is_64) },
+        FORM.strx1 => return FormValue{ .StrOffset = @intCast(u64, try in_stream.readInt(u8, endian)) },
```
topolarity, line 530: If it's not too much trouble, it'd be really nice to add branches for the other missing DWARFv5 forms: `strx`, `addrx`, `addrx1`, `addrx2`, `addrx3`, `addrx4`, `data16`, `implicit_const`, `line_strp`, `loclistx`, `rnglistx`, `ref_sup4`, `ref_sup8`, `strp_sup`

You don't have to implement these now, but adding a good error or panic would be useful.

topolarity (commented):
Thanks for putting this change together, and for testing it 🙂 

Only request is to add some handling for the DWARFv5 functionality that is still missing

src/link/MachO/Object.zig
```diff
@@ -579,6 +579,7 @@ pub fn parseDwarfInfo(self: Object) error{Overflow}!dwarf.DwarfInfo {
         .debug_info = &[0]u8{},
         .debug_abbrev = &[0]u8{},
         .debug_str = &[0]u8{},
+        .debug_str_offsets = null,
```
topolarity, line 582: Same comment as for COFF above

kubkon: I agree with @topolarity here - in fact, I'd just go full on fatal error here with an instructive comment such as

```suggestion
        .debug_str_offsets = unreachable, // TODO implement string offsets for DWARFv5
```

kdchambers: I've not used that pattern before. Would that not trigger a panic unconditionally whenever `parseDwarfInfo` is called?

lib/std/debug.zig
kdchambers: The thing is I don't have a windows machine and I don't want to risk introducing a regression. As you know this code is only triggered when a compiled executable hits an error, so I could imagine errors introduced here not getting caught by CI. I don't know, but I figured I'd play it safe if I'm not regressing anything my omitting a feature / change.

I was also just trying to keep this PR simple and not get into refactoring as it's my first, I can see that there's a lot of simple improvements that could be made in this file.

I think the best place for this error would be in the `getAttrString` function. If a string of form type `StrOffset` is requested but we haven't loaded a `debug_str_offsets` table then we can report that to the user. Does that sound reasonable?

I will check the specification in a sec, but I'm not 100% sure that `debug_str_offsets` is a mandatory section in V5 if there isn't any string data of this form type.

lib/std/dwarf.zig
kdchambers: Yes, I think this is a fair point. 

How would you like the error to be reported? There are two ways I could do it.

1. Just return a normal error E.g `error.FormNotImplementedAddrx`. This will always result in broken stack trace, even if the data in question is never actually requested.
2. Consume the number of bytes that correspond to the form type and log an error / warning.

topolarity: I'd say just add `@panic("Could not print stack trace: DWARFv5 not fully supported")`. Normally we'd use `unreachable`, but we can only track that one down if backtraces are working so that won't do 😛

Returning a normal error is also fine by me, but the error handling for stack-trace printing will need to be updated, too.

BTW there's one more item other than these forms to round out DWARFv5 support: the `.debug_line` section has a V5 format that we can't parse yet

lib/std/debug.zig
topolarity: That error location sounds good to me :+1:

src/link/MachO/Object.zig
kubkon: Oh oops, you're absolutely right. Please ignore this suggestion and follow what @topolarity is suggesting :-)

topolarity:
Thanks for the PR @kdchambers - Congrats on your first contribution 🙂 

I've updated #12120 to include any loose threads from this PR.

PR #12276, opened 2022-07-28
autodoc: better short description algorithm

r00ster91:
Just a small thing I found I could do.

# Before

image

# After

image


kristoff-it:
Makes sense, thank you very much!

PR #12278, opened 2022-07-28
std: rename std.Target.systemz to .s390x

nektro:
- `std.Target.Cpu.Arch` enum uses s390x
- https://github.com/llvm/llvm-project/blob/main/compiler-rt/cmake/Modules/AllSupportedArchDefs.cmake#L13 clang frontend uses s390x
- https://gcc.gnu.org/install/specific.html#s390x-x-linux gcc frontend uses s390x


andrewrk:
Thanks!

PR #12283, opened 2022-07-29
feat: flake.nix

willruggiano:
For those that use Nix(OS). There are versioned derivations in nixpkgs, but I like to live on the edge. And the only way to _not_ have to have a source hash (which renders any HEAD-based derivation completely useless) is by referencing the source tree directly (instead of, e.g., fetchFromGitHub). So... voila.

### DIFF
--- a/contrib/flake.lock
+++ b/contrib/flake.lock
@@ -1,0 +1,43 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1658937758,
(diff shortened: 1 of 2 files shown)


andrewrk:
Thanks for the contribution. Let's keep this kind of thing in a third party location; I do not wish to maintain this in the ziglang repo. I'm also not going to play favorites with NixOS and certainly not going to maintain one of these for every linux distro under the sun.

PR #12284, opened 2022-07-29
LLVM15 Upgrades

nektro:
WIP


CMakeLists.txt
```diff
@@ -108,9 +108,9 @@ else()
     set(ZIG_USE_LLVM_CONFIG OFF CACHE BOOL "use llvm-config to find LLVM libraries")
 endif()
 
-find_package(llvm 14)
-find_package(clang 14)
-find_package(lld 14)
+find_package(llvm 15)
```
andrewrk, line 111: This diff is confusing me - I'm looking at this line in the llvm15 branch and it already has this change

andrewrk:
if you want to land some commits in this branch you're gonna need to break it into smaller, mergeable pieces. Otherwise I'm going to have to duplicate your work while it's sitting here as a draft.

PR #12286, opened 2022-07-29
Division safety improvements

andrewrk:
* Added safety for integer overflow (-MAX_INT/-1)
 * Omit division by zero safety check when RHS is comptime known to be non-zero.
 * Avoid emitting `_optimized` variants of AIR instructions for integers (this suffix is intended to be used for floats only).

cc @joachimschmidt557 

your example

```zig
fn div(x: u32) u32 {
    return x / 32;
}
```

now produces:

```
# Begin Function AIR: test2.div:
# Total AIR+Liveness bytes: 240B
# AIR Instructions:         8 (72B)
# AIR Extra Data:           8 (32B)
# AIR Values Bytes:         2 (16B)
# Liveness tomb_bits:       8B
  %3 = constant(comptime_int, 32)
  %4 = constant(u32, 32)

  %0 = arg(u32)
  %1!= dbg_block_begin()
  %2!= dbg_stmt(2:5)
  %5 = div_trunc(%0!, %4!)
  %6!= dbg_block_end()
  %7!= ret(%5!)
# End Function AIR: test2.div
```


src/type.zig
```diff
@@ -5202,9 +5202,18 @@ pub const Type = extern union {
...
+    pub fn minInt(ty: Type, arena: Allocator, target: Target) !Value {
+        const scalar = try minIntScalar(ty.scalarType(), arena, target);
+        if (ty.zigTypeTag() == .Vector) {
+            return Value.Tag.repeated.create(arena, scalar);
+        } else {
+            return scalar;
+        }
+    }
+
+    pub fn minIntScalar(ty: Type, arena: Allocator, target: Target) !Value {
```
joachimschmidt557, line 5214: The doc comment for `minInt` now only applies to `minIntScalar`, right?

PR #12287, opened 2022-07-29
translate-c: use correct number of initializers for vectors

### LINKED ISSUE #12264: stage2 run-translated-c test failure: basic vector expressions
```
[nix-shell:~/dev/zig/build-release]$ stage3/bin/zig build test-run-translated-c
/home/andy/.cache/zig/o/563af7cb75641c10abcd1cd7c94de828/source.zig:465:36: error: expected 8 vector elements; found 0
    var empty_init: __v8hi = __v8hi{};
                             ~~~~~~^~
error: run-translated-c basic vector expressions build-exe...
```

It looks like translate-c is taking advantage of stage1 allowing zero-initialization of vectors with `V{}` syntax. However I'm not sure why that works - it's not planned to be able to initialize vectors this way; this needs to be changed to `@splat(N, @as(I, 0))`.

ehaas:
Fixes #12264

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -2688,15 +2688,25 @@ fn transInitListExprVector(
 ) TransError!Node {
     _ = ty;
     const qt = getExprQualType(c, @ptrCast(*const clang.Expr, expr));
-    const vector_type = try transQualType(c, scope, qt, loc);
+    const vector_ty = @ptrCast(*const clang.VectorType, qualTypeCanon(qt));
+
     const init_count = expr.getNumInits();
+    const num_elements = vector_ty.getNumElements();
+    const element_qt = vector_ty.getElementType();
 
     if (init_count == 0) {
-        return Tag.container_init.create(c.arena, .{
-            .lhs = vector_type,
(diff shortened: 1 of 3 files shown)


src/translate_c.zig
```diff
@@ -2688,16 +2688,15 @@ fn transInitListExprVector(
...
     const qt = getExprQualType(c, @ptrCast(*const clang.Expr, expr));
-    const vector_type = try transQualType(c, scope, qt, loc);
     const init_count = expr.getNumInits();
 
     if (init_count == 0) {
-        return Tag.container_init.create(c.arena, .{
-            .lhs = vector_type,
-            .inits = try c.arena.alloc(ast.Payload.ContainerInit.Initializer, 0),
-        });
+        return Tag.undefined_literal.init();
```
Vexu, line 2694: This should be `@splat(num_elements, zero)`.

Vexu (changes requested):

src/translate_c.zig
ehaas: Oof, good catch - that could have led to some ugly bugs. Updated.

PR #12296, opened 2022-07-30
cmake: Print all LLVM config errors instead of just the last one

sin-ack:
If you have multiple llvm-config executables in your path, and all of
them cause failures, then only the last failure will be printed. This
can cause confusion when the multiple llvm-config executables are from
different major LLVM versions, i.e. LLVM 13 and 14, which might mask an
error that happened on the LLVM 14 llvm-config with an unrelated error.

This commit makes it so that all errors are collected into a list and
printed all at once; this way, you can see how each llvm-config
executable failed to configure properly. Note that the failures still
won't be printed if a successful configuration is found.

### DIFF
--- a/cmake/Findllvm.cmake
+++ b/cmake/Findllvm.cmake
@@ -10,6 +10,7 @@
 
 
 if(ZIG_USE_LLVM_CONFIG)
+  set(LLVM_CONFIG_ERROR_MESSAGES "")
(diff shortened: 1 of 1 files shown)


topolarity (approved):
Nice thinking! Thanks for the enhancement :+1:

PR #12297, opened 2022-07-30
feat: prettify compile log output

r00ster91:
## Before

image

The "Compile Log Output" currently looks very plain and unconnected to the pretty messages before it with all the colors and stuff. The purpose of this PR is to make it look prettier and more connected to what's before it and it's also supposed to make the values easier to read if you have many values in a single line separated by commas.
Currently it kinda blends in with the `error: x...` stuff below that no one reads so that's not very good.

## After

image

In addition to the previous one,
- It dims the dashes and the commas and makes the values bold.
- It uses cyan to make the compile log output stick out further. Currently cyan seems to be used only for `note: `s but I think the compile log output is similar. They're also notes that you wrote during comptime.
Another reason I think cyan makes sense is that `@compileError` is "worse" and it uses red and then `@compileLog` is not as bad and uses cyan, if you know what I mean.

This may not look complicated but it's significantly more complicated to do because color is not available in `Sema`. The way I do this is I concatenate the strings in `Sema` and use unprintable characters as terminators and then in `main` I `mem.split` `comp.getCompileLogText()` and add color. This is supposed to avoid using an unnecessarily complicated structure like a `ArrayListUnmanaged(ArrayListUnmanaged(u8))` or something.

In regards to distinguishing multiple values in one line, this is probably a lot better.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -3521,6 +3521,46 @@ const AfterUpdateHook = union(enum) {
     update: []const u8,
 };
 
+// we take the raw compile log text, split it back up, and prettify it
+fn printCompileLog(comp: Compilation, ttyconf: std.debug.TTY.Config) !void {
+    const log_text = comp.getCompileLogText();
+    if (log_text.len == 0) return;
+
+    var bw = io.bufferedWriter(io.getStdErr().writer());
+    const stderr = bw.writer();
+
+    try ttyconf.setColor(stderr, .Cyan);
+    try stderr.writeAll("compile log output:\n");
+    try ttyconf.setColor(stderr, .Reset);
+
+    var log_lines = mem.split(u8, log_text, "\n");
+    while (log_lines.next()) |log_line| {
+        if (log_lines.index == null) break;
+
+        try ttyconf.setColor(stderr, .Dim);
+        try stderr.writeAll("- ");
+        try ttyconf.setColor(stderr, .Reset);
+
+        var values = mem.split(u8, log_line, "\x00");
+        while (values.next()) |value| {
+            // we bolden the values and dim the commas to
+            // make the values easier to distinguish
+
+            try ttyconf.setColor(stderr, .Bold);
+            try stderr.writeAll(value);
+            try ttyconf.setColor(stderr, .Reset);
+
+            if (values.index != null) {
+                try ttyconf.setColor(stderr, .Dim);
+                try stderr.writeAll(", ");
+                try ttyconf.setColor(stderr, .Reset);
+            }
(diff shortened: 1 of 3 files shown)


r00ster91:
So currently this is version 2. If people don't like it (I like it), we can go with version 1.

src/main.zig
```diff
@@ -3322,6 +3322,46 @@ const AfterUpdateHook = union(enum) {
     update: []const u8,
 };
 
+// we take the raw compile log text, split it back up, and prettify it
+fn printCompileLog(comp: Compilation, ttyconf: std.debug.TTY.Config) !void {
+    const log_text = comp.getCompileLogText();
+    if (log_text.len == 0) return;
+
+    var bw = io.bufferedWriter(io.getStdErr().writer());
```
r00ster91, line 3330: (If we go with version 2)
Btw, I noticed there's no write buffering for `renderToStdErr` (where it prints compile errors).
Maybe it's be possible to reuse the `io.BufferedWriter` I use for stderr here for `renderToStdErr` too at some point? Or is it already buffered and I'm missing something?

src/main.zig
```diff
@@ -3322,6 +3322,46 @@ const AfterUpdateHook = union(enum) {
...
+        var values = mem.split(u8, log_line, "\x00");
+        while (values.next()) |value| {
+            // we bolden the values and dim the commas to
+            // make the values easier to distinguish
+
+            ttyconf.setColor(stderr, .Bold);
+            try stderr.writeAll(value);
+            ttyconf.setColor(stderr, .Reset);
+
+            if (values.index != null) {
```
r00ster91, line 3354: I was wondering if we could add `isLast` to the 3 `split`, `splitBackwards`, and `tokenize` iterators in `std.mem`? Often people (or at least me) try to do something like `iter.index != len - 1` which doesn't actually work because it's an index for the slice, not the iteration index.
I think it's just very common to want to know if this is the last iteration when splitting. Is this something worth adding? I don't think it's obvious what `index` exactly is and that it's `null` on the last iteration. That's why we could add a method  for it.

r00ster91:
I simplified the PR and made it mergeable again.

Here is once more what this looks like:
image

I heard `@compileLog` is going to change again anyway so it may be a bit premature to focus on styling its output yet, but I guess this is just for the meantime. I'm not sure. So otherwise feel free to close this too.

andrewrk:
is it a good idea, is it a bad idea? I don't know, maybe the plain text is better. It seems kind of unimportant to me right now. Closing to focus on other things.

PR #12298, opened 2022-07-30
std.debug: handle some possible errors and resolve low-hanging TODOs

r00ster91:
`ttyconf.setColor(stderr, .Dim)` just like `stderr.print("{s}", .{plain.msg})` can fail, so I don't see why we wouldn't handle the former just like the latter. It's an edge case. You can always still ignore the error in your specific case if you want to.
Apart from that it solves a bunch of related TODOs and it also removes a bunch of `unreachable`s in `std.Progress` which is supposed to be non-fallible because now I'm confident these `unreachable`s can be reached since the user can modify the `std.Progress.terminal` field. It can probably fail anyway. Better safe than sorry.

### DIFF
--- a/lib/std/debug.zig
+++ b/lib/std/debug.zig
@@ -761,14 +761,12 @@ fn printLineInfo(
     }
 }
 
-// TODO use this
 pub const OpenSelfDebugInfoError = error{
     MissingDebugInfo,
-    OutOfMemory,
     UnsupportedOperatingSystem,
 };
 
-pub fn openSelfDebugInfo(allocator: mem.Allocator) anyerror!DebugInfo {
+pub fn openSelfDebugInfo(allocator: mem.Allocator) OpenSelfDebugInfoError!DebugInfo {
     nosuspend {
         if (builtin.strip_debug_info)
             return error.MissingDebugInfo;
@@ -785,7 +783,7 @@ pub fn openSelfDebugInfo(allocator: mem.Allocator) anyerror!DebugInfo {
             .windows,
             .solaris,
             => return DebugInfo.init(allocator),
-            else => return error.UnsupportedDebugInfo,
+            else => return error.UnsupportedOperatingSystem,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


andrewrk:
Looks like I merged your PRs out of order. Sorry about that. You mind rebasing this one?

r00ster91:
Sure! Rebased.

r00ster91:
Now it should be up-to-date again.

lib/std/debug.zig
```diff
@@ -415,9 +407,9 @@ pub fn writeStackTrace(
     if (stack_trace.index > stack_trace.instruction_addresses.len) {
         const dropped_frames = stack_trace.index - stack_trace.instruction_addresses.len;
 
-        tty_config.setColor(out_stream, .Bold);
+        try tty_config.setColor(out_stream, .Bold);
```
andrewrk, line 410: failure here should not cause the stack trace to abort being printed.

lib/std/debug.zig
```diff
@@ -605,59 +597,48 @@ pub const TTY = struct {
...
                         var attrs: windows.WORD = undefined;
                         var init_attrs = false;
                     };
                     if (!S.init_attrs) {
                         S.init_attrs = true;
                         var info: windows.CONSOLE_SCREEN_BUFFER_INFO = undefined;
-                        // TODO handle error
-                        _ = windows.kernel32.GetConsoleScreenBufferInfo(stderr_file.handle, &info);
+                        if (windows.kernel32.GetConsoleScreenBufferInfo(file.handle, &info) != windows.TRUE)
+                            return error.Unexpected;
```
andrewrk, line 629: Use a different error code. Unexpected is for when we didn't think it was possible for the OS to return such an error code. If that is the situation here, then use `std.os.windows.unexpectedError()` so that if an unexpected error does occur in debug mode, the programmer gets a helpful stack trace that informs a patch that can be made to improve the std lib code to add handling for that error code, now that it is known it can be returned.

andrewrk (changes requested):

Issue #4281, opened 2020-01-24
`@compileLog(@typeName(T))` sometimes prints `[x:0]u8{...}` array literals


metaleap:
For [example](https://github.com/metaleap/zigbag/blob/678c6a58b63bb055102eb3a6b5f3a5978aad8919/dummylangserver/lsp_json.zig#L26), when I `@compileLog(@typeName(T))` for `T`s from [this fat bag of type-defs](https://github.com/metaleap/zigbag/blob/678c6a58b63bb055102eb3a6b5f3a5978aad8919/dummylangserver/lsp_types.zig) where `@typeId(T) == .Enum`:

```
| *[21:0]u8{108,115,112,95,116,121,112,101,115,46,101,110,117,109,58,56,53,52,58,49,51}
| *"InsertTextFormat"
| *"CompletionItemKind"
| *[24:0]u8{83,105,103,110,97,116,117,114,101,72,101,108,112,84,114,105,103,103,101,114,75,105,110,100}
| *"CompletionItemTag"
| *[21:0]u8{67,111,109,112,108,101,116,105,111,110,84,114,105,103,103,101,114,75,105,110,100}
| *"lsp_types.enum:235:16"
| *"DiagnosticTag"
| *"SymbolKind"
```

All come from the exact same `@compileLog` call site, with `@typeId(T) == .Enum` always `true`.

This doesn't occur randomly, the same outputs occur exactly the same during many repeated attempts. I have some nutty type nestings in there (it expresses an MS-devised protocol so no surprise there..), so might be a neat stress test for the type-name-algo..  :grin:

dbandstra:
Related #1747

LemonBoy:
When the compiler is sure that the array contents comes from a `Buf` literal it will happily print it a string, otherwise it has no way to know that the content is ... well, text and not binary data or whatever.

andrewrk:
I think the best we can do is have `@compileLog` do heuristics. If the len is less than 100 or something like that then treat it as a string (sometimes escaping with `\xNN` if it has to), otherwise treat it as an array.

topolarity:
Wanted to tack on a good question raised in #12303: 

What's the need for `@compileLog` to accept anything other than a string, provided that we have `std.fmt.comptimePrint`?

Vexu:
here's a simple example:
```zig
comptime {
    var a = [2]u8{ 1, 2 };
    var b = a[1..];
    @compileLog(b);
}
```
[...]

Issue #868, opened 2018-03-30
Add ability to determine if we are currently executing at compile time


tiehuis:
I'm sure this was mentioned somewhere but could not find an issue: The gist is as follows:

```
if (@areWeInComptime()) {
   // use a slow fallback which we can compute in zig
} else {
   // use a fast runtime version that uses assembly (or features not supported at comptime)
}
```

This is present in D for example via the `__ctfe` variable. The current reason I need this is that I want to call `sqrt` at compile-time but on x86_64 this uses an assembly routine which can not be evaluated at compile-time.

andrewrk:
I think this makes sense. One thing I want to avoid though is over-reliance on this. In most cases we should strive to make the runtime and comptime representation the same. Something that I never want to see is a function that could have been perfectly fine executing in both comptime and runtime contexts, but somebody didn't think that was possible and did something like `if (@areWeInComptime()) @compileError("unsupported")`.

In the case of inline assembly for sqrt, see #767

tiehuis:
Completely agree on not over-relying on this. I did forget about the llvm math intrinsics. I'm happy enough to hold of on this for now until we get other possible use cases that require something of this sort.

andrewrk:
It's kind of a dirty secret that this is possible, but...

```zig
const std = @import("std");

test "detect comptime" {
    std.debug.warn("\n");
    std.debug.warn("1={}\n", isCompTime());
    std.debug.warn("2={}\n", comptime isCompTime());
    std.debug.warn("3={}\n", comptime isCompTime());
    std.debug.warn("4={}\n", isCompTime());
}

fn isCompTime() bool {
    var t: bool = true;
    const x = if (t) u7(0) else u8(0);
    return @typeOf(x) == u7;
}
```

don't rely on this... zig reserves the right to make the `if` static if it can prove the value is compile-time known (which it should be able to do in this example with more sophisticated semantic analysis).

jibal:
> Don't rely on this - zig reserves the right to make the `if` static if it can prove the value is compile-time known (which it should be able to do in this example with more sophisticated semantic analysis).

Having to rely on it is even worse than the possibility of abusing @areWeInComptime().

>  In most cases we should strive to make the runtime and comptime representation the same.

[...]

PR #12303, opened 2022-07-30
stdlib: Support `std.log` at comptime, using `@compileError`

topolarity:
This is a rudimentary (and arguably cursed) implementation of comptime logging. It works in stage 1 and stage 2.

I'm not sure we'll want to merge this, but it might be useful so I decided to open the PR just to get feedback.

```zig
test {
    const Foo = struct { a: u8, b: i32 };
    const data = Foo{ .a = 15, .b = -16 };
    comptime std.log.err(
        \\Error encountered!
        \\Data was: {}
        \\
        \\Have you tried turning it off and on again?
        , .{ data });
}
```

when compiled, this prints:
```
$ ./stage2/bin/zig test test.zig
lib/std/log.zig:183:9: error: Error encountered!
                              Data was: test.test_0.Foo{ .a = 15, .b = -16 }

                              Have you tried turning it off and on again?

lib/test_runner.zig:130:34: note: called from here
lib/std/log.zig:139:21: note: called from here
lib/std/log.zig:217:16: note: called from here
test.zig:6:25: note: called from here
```

You can only log one message, since this uses `@compileError`. AFAIK `@compileLog` currently has no way of printing a string if it's not a string literal. 

kudos to @r00ster91 for the idea

### DIFF
--- a/lib/std/log.zig
+++ b/lib/std/log.zig
@@ -128,8 +128,7 @@ fn log(
     const effective_log_level = blk: {
         inline for (scope_levels) |scope_level| {
             if (scope_level.scope == scope) break :blk scope_level.level;
-        }
-        break :blk level;
+        } else break :blk level;
     };
 
     if (@enumToInt(message_level) <= @enumToInt(effective_log_level)) {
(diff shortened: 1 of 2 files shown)


r00ster91:
It was just a random thought, didn't realize we could actually do it like this. However,
1. I'm worried about cases where something is implicitly comptime without needing to specify it as comptime. We might have many cases where a log is supposed to be a runtime log but it ends up being comptime?
2. `std.log` provides many different levels for logging. Warnings, errors, info logs, debugging logs, etc. Considering that comptime logs are very temporary because you can't compile your code with them, it seems overkill to have these levels for comptime logs.
3. > You can only log one message, since this uses `@compileError`.

   This seems very limiting and we would definitely have to support printing multiple logs in a single compilation somehow. I wonder if there is a way to solve this? I don't really see a way to do this though without... a builtin function. Which brings us back to where we are really because in that case the user might as well use that builtin function directly and we don't need to special-case this in `std.log`.
So I think status quo is fine, or maybe there is a way to address these points? Just my two cents. Anyway, it's an interesting idea!

r00ster91:
> AFAIK `@compileLog` currently has no way of printing a string if it's not a string literal.

Why don't we just make `@compileLog` take `comptime format: []const u8, args: anytype`? And same for `@compileError` maybe.

r00ster91:
It seems like this can serve as an alternative rather than as a replacement for `@compileLog`.
The use case is printing comptime strings with formatting. Currently, there seems to be no way to make this print as `hello` rather than as the bytes of `"hello"`:
```zig
const std = @import("std");

pub fn main() !void {
    const x = comptime std.fmt.comptimePrint("{s}", .{"hello"});
    @compileLog(x);
}
```
Allowing to `std.log` functions to be called at comptime is one way to solve this.
Other ideas we had includes
* having a special type in `std.builtin` to wrap your string in and then if `@compileLog` sees this argument it will print the string as it is? E.g. `@compileLog(std.builtin.String(x))`.
* when an argument is something like `[]const u8` and all characters are printable, print it as the string itself. This may lead to unexpected results in some cases.

Vexu:
IMO this is a misuse of `std.log` (compile errors != loggin) and that comptime hack does not belong in std at least not before #868 is resolved.

topolarity:
> IMO this is a misuse of std.log (compile errors != loggin)

Agreed, at least for `std.log.debug` and `std.log.info`.

I do think getting the ability to output strings from `@compileLog` would be a good idea, so that formatted logging is possible. That's the only reason I used `@compileError` here

> that comptime hack does not belong in std at least not before https://github.com/ziglang/zig/issues/868 is resolved.

I can definitely agree with that. Thanks for the pointer to the history - I wasn't sure what the stance on this kind of comptime detection was in general.

Closing for now, until #868 is resolved.

Vexu:
> I do think getting the ability to output strings from `@compileLog` would be a good idea, so that formatted logging is possible. That's the only reason I used `@compileError` here

The plan is to fix this in `@compileLog` as outlined in https://github.com/ziglang/zig/issues/4281#issuecomment-579508154

PR #12305, opened 2022-07-31
autodoc: Add borders to field docs to make it clear which field they are associated with

squeek502:
As discussed in Discord.

Before:
autodoc-before

After:
autodoc-after

### DIFF
--- a/lib/docs/index.html
+++ b/lib/docs/index.html
@@ -284,6 +284,16 @@
         overflow-x: hidden;
       }
 
+      .fieldHasDocs {
+        margin-bottom: 0;
+      }
+
+      .fieldDocs {
+        border: 1px solid #2A2A2A;
+        border-top: 0px;
+        padding: 1px 1em;
+      }
+
       /* help dialog */
       .help-modal {
         display: flex;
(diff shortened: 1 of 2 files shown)


lib/docs/index.html
```diff
@@ -284,6 +284,15 @@
         overflow-x: hidden;
       }
 
+      .fieldHasDocs {
+        margin-bottom: 0;
+      }
+
+      .fieldDocs {
+        border: 1px solid #2A2A2A;
+        padding: 1px 1em;
```
squeek502, line 293: Note: the 1px horizontal padding seems to be necessary to avoid margin collapse

lib/docs/index.html
```diff
@@ -284,6 +284,16 @@
...
       }
 
+      .fieldHasDocs {
+        margin-bottom: 0;
+      }
+
+      .fieldDocs {
+        border: 1px solid #2A2A2A;
+        border-top: 0px;
+        padding: 1px 1em;
```
squeek502, line 294: The `1px` horizontal padding here is necessary to avoid margin collapse when `border-top` is set to 0 (otherwise, the border wouldn't extend all the way up to the preceding `pre`).

Issue #2370, opened 2019-04-27
std.fmt: Make max_depth configurable


squeek502:
Split from https://github.com/ziglang/zig/issues/2361 and https://github.com/ziglang/zig/pull/2368

> My proposed solution to this would be to have a finite number of depth that fmt traverses. Probably 1, 2, or 3 by default. And then perhaps a way to override this with a larger number.

https://github.com/ziglang/zig/blob/205e501e42b9a2e1d35afa0873311165100863ae/std/fmt.zig#L11

Issue #1358, opened 2018-08-09
overhaul std.fmt formatting api


tiehuis:
**[Remaining Work](https://github.com/ziglang/zig/issues/1358#issuecomment-511193492).**

This is a proposal for the formatting interface exposed via `std.fmt` and which
shows up in most printing functions (e.g. `std.debug.warn`).

This is largely based on Rust's [`std::fmt`](https://doc.rust-lang.org/std/fmt/index.html) (which in turn is similar to Python3) so see that for a more in-depth reference for certain parts.

### Formatting Options

We **do** take the following formatting options from Rust:

 - positional parameters (`"{0}"`).
 - alignment (`"{:<} {:<5} {:0^10}"`)
 - width (`"{:5}"`)
 - precision (`"{:.5} {:.0}"`)

We **do not** take the following:

 - `#` alternate printing forms
 - `+`, `-`, `0` sign flags (NOTE: *may actually want these*)
 - named parameters (`format!("{arg1}", arg1 = "example")`)
 - runtime specified precision (`format!("{:.*}", 3, 5.0923412`) (NOTE: *could add this in if reasonable demand*)
 - numbered argument specified precision (`format!("{0:1$}", 5.0923412, 3)`)

### Format Specifiers

These are largely unchanged but a few are:

 - `{}`  (`primitives`) print the default primitive representation (if it exists)
 - `{c}` (`int`): print as an ascii character
 - `{b}` (`int`): print as binary
 - `{x}` (`int`): print as lowercase hex
 - `{X}` (`int`): print as uppercase hex
 - `{o}` (`int`): print as octal
 - `{e}` (`float`): print in exponent form
 - `{d}` (`int`/`float`): print in base10/decimal form
 - `{s}` (`[]u8`/`*u8`): print as null-terminated string
 - `{*}` (`any`): print as a pointer (hex) (NOTE: *does & make more sense here?*)
 - `{?}` (`any`): print full debug representation (e.g. traverse structs etc to primitive fields)
 - `{#}` (`any`): print raw bytes of the value (hex) (NOTE: *do we need this? how often is it used?*)

These format specifiers are **removed** from the current implementation:

 - `{.}` (`float`): was to specify decimal float, now `{d}` replaces this
 - `{e10}` (`float`): precision was attached to format specifier. The new format
           specifier type would replace this.
 - `{B}` (`any`): printed raw bytes of value, replaced by `{#}`. This is to
 ensure it cannot be shadowed by a user defined function.

[...]

PR #12310, opened 2022-07-31
fmt: Make default_max_depth configurable

ominitay:
There is currently no way to override this when using `std.fmt.fmt()`, and derived functions (`anything.print()`, `std.log`, etc). This small change allows overriding the default value. In future, there should likely be a specifier in the fmt string. That is probably best to be discussed in %%1358.


squeek502:
Closes, or at least contributes towards https://github.com/ziglang/zig/issues/2370

ominitay:
Would like to see what @andrewrk thinks about this, considering the closed PR, and that there seems to have been no further discussion since.

Vexu:
#1358 looks like it is mostly implemented already, maybe we should discuss how this could be integrated into the argument system in #2370 before adding what looks like a workaround.

ominitay:
I don't feel that this is a workaround. The default should be configurable IMO, and should be able to be overridden on a case-by-case basis.

ominitay:
Anyway, I'd like to hear Andrew's thoughts on this, considering he had some plan in mind.

Vexu (approved):
Needs to be updated to use `std.options`.

ominitay:
Will get that done soon :)

ominitay:
Haven't been programming for a while, so please give my changes a little scrutiny ;)

PR #12311, opened 2022-08-01
Update Linux syscall list for 5.19

The-King-of-Toasters:
New changes: memfd_secret implemented for RISC-V.


The-King-of-Toasters:
Dunno why the MacOS build failed, hopefully rebasing should fix it.

PR #12318, opened 2022-08-02
Stage2: misc compile error improvements

Vexu:
Only 99 cases left! (of which 20 are async related)


src/AstGen.zig
```diff
@@ -5871,6 +5937,7 @@ fn forExpr(
     rl: ResultLoc,
     node: Ast.Node.Index,
     for_full: Ast.full.While,
+    is_statement: bool,
```
marler8997, line 5940: enum instead of bool?

Vexu: For public functions absolutely, but this is a private function used in only two places so I don't think it matters too much.

PR #12320, opened 2022-08-03
macho: sync with standalone zld repo, improve memory usage and linking speed

kubkon:
Syncs with gitrev https://github.com/kubkon/zld/commit/a2c32e972f8c5adfcda8ed2d99379ae868f59c24

In addition, also revamps how we create and track output sections - instead sorting the ordinals and updating every symbol within each section, we keep the sections in a sorted order from the get-go. The logic for selecting and creating an output section for some input section is also significantly simplified leading to cleaner and smaller code.

I want to point out here that I have made the first step towards a more cache-friendly linker via the use of multiarraylist to track sections with the associated metadata such as the output segment index, and last atom in the section.


haze:
awesome!

PR #12322, opened 2022-08-03
std.meta.Tuple memoization fix

MasterQ32:
Makes std.meta.Tuple and std.meta.ArgsTuple generate a unique type instead of generating one per invocation.


viriuwu:
fwiw this workaround won't be needed for stage2 as tuples are tested for structural equality, but this is very nice until then 🙏

PR #12326, opened 2022-08-03
stage2: correct node offset of nested declarations

Vexu:


andrewrk:
Did you test if debug info of line numbers of nested declarations still works correctly when doing incremental compilation and an unchanged decl gets shifted down?

Vexu:
It doesn't seem to work correctly in master branch? Using the LLVM backend with `--watch` always crashes on `update` and the self hosted backend doesn't seem to emit working debug info with `build-exe`. I was able to get something when using `zig3 run --watch -fno-LLVM -femit-bin` but then the line numbers don't get updated. Tested with:
```zig
const S = struct {
    fn foo() void {
        @breakpoint();
    }
};
pub fn main() void {
    S.foo();
}
```
gdb complains about the debug info:
```
Reading symbols from a...
BFD: a: invalid string offset 1651076191 >= 574 for section `.shstrtab'
```

andrewrk (approved):
Well, that's the purpose of the fields that are touched in this PR. So if you want to make the non-incremental case work now, and then we can re-evaluate what needs to change to fix incremental compilation later, that's fine with me.

Issue #12324, opened 2022-08-03
stage2 infinite recursion in `Sema.explainWhyTypeIsComptime`


Vexu:
```zig
const U = union {
    a: fn () void,
    b: *U,
};
test {
    var u: U = undefined;
    _ = u;
}
```

PR #12327, opened 2022-08-03
Fix infinite recursion in sema.explainWhyTypeIsComptime

antlilja:
This PR solves #12324

Fixes infinite recursion in sema.explainWhyTypeIsComptime using a flat tree.


Vexu:
Types are hashable so you could just use a `HashMap` instead of creating a new data structure.

antlilja:
> Types are hashable so you could just use a `HashMap` instead of creating a new data structure.

The problem is that we need to keep track of only the parent types, and a ```HashMap``` would keep track of whether a type has ever been refrenced since the call to ```sema.explainWhyTypeIsComptime```. But maybe I'm misinterpreting what you mean. I could hash the types and compare the hashes first inside of ```TypeTree.typeExistsAsParent``` to avoid so many calls to ```Type.eql```.

Vexu:
> The problem is that we need to keep track of only the parent types, and a `HashMap` would keep track of whether a type has ever been refrenced since the call to `sema.explainWhyTypeIsComptime`.

I get identical output to this branch with a hash map, do you have any example where the behavior differs?

Speaking of examples, this PR needs a test case.

antlilja:
> > The problem is that we need to keep track of only the parent types, and a `HashMap` would keep track of whether a type has ever been refrenced since the call to `sema.explainWhyTypeIsComptime`.
> 
> I get identical output to this branch with a hash map, do you have any examples where the behavior differs?
> 
> Speaking of examples, this PR needs a test case.

This example will give a different result:
```zig
const S = struct {
	a: S2,
	b: S3,
	c: S4,
};

const S2 = struct {
    a: fn () void,
};

const S3 = struct {
	a: S2,
};

const S4 = struct {
	a: S2,
};

test {
    var s: S = undefined;
    _ = s;
}
```

It still works with a `HashMap`, but it doesn't create a note on every field.

antlilja:
> I still get the identical output with that example.

I get this output using this branch:
```
boo.zig:19:12: error: variable of type 'boo.S' must be const or comptime
    var s: S = undefined;
           ^
boo.zig:2:8: note: struct requires comptime because of this field
    a: S2,
       ^~
boo.zig:3:8: note: struct requires comptime because of this field
    b: S3,
       ^~
boo.zig:12:8: note: struct requires comptime because of this field
    a: S2,
       ^~
boo.zig:4:8: note: struct requires comptime because of this field
    c: S4,
       ^~
boo.zig:15:8: note: struct requires comptime because of this field
    a: S2,
       ^~
```

And this output using a HashSet:
```
boo.zig:19:12: error: variable of type 'boo.S' must be const or comptime
    var s: S = undefined;
           ^
boo.zig:2:8: note: struct requires comptime because of this field
    a: S2,
```

### HashSet variatnt:
```zig
const TypeSet = std.HashMap(Type, void, struct {
    mod: *Module,

    pub fn eql(self: @This(), a: Type, b: Type) bool {
        return a.eql(b, self.mod);
    }

    pub fn hash(self: @This(), x: Type) u64 {
        return x.hash(self.mod);
    }
}, std.hash_map.default_max_load_percentage);

fn explainWhyTypeIsComptime(
    sema: *Sema,
    block: *Block,
    src: LazySrcLoc,
    msg: *Module.ErrorMsg,
    src_loc: Module.SrcLoc,
    ty: Type,
) CompileError!void {
    var type_set = TypeSet.initContext(sema.gpa, .{ .mod = sema.mod });
    defer type_set.deinit();

    return sema.explainWhyTypeIsComptimeInner(block, src, msg, src_loc, ty, &type_set);
}

fn explainWhyTypeIsComptimeInner(
    sema: *Sema,
    block: *Block,
    src: LazySrcLoc,
    msg: *Module.ErrorMsg,
    src_loc: Module.SrcLoc,
    ty: Type,
    type_set: *TypeSet,
) CompileError!void {
    const mod = sema.mod;
    switch (ty.zigTypeTag()) {
        .Bool,
        .Int,
        .Float,
        .ErrorSet,
        .Enum,
        .Frame,
        .AnyFrame,
        .Void,
        => return,

        .Fn => {
            try mod.errNoteNonLazy(src_loc, msg, "use '*const {}' for a function pointer type", .{
                ty.fmt(sema.mod),
            });
        },

        .Type => {
            try mod.errNoteNonLazy(src_loc, msg, "types are not available at runtime", .{});
        },

        .BoundFn,
        .ComptimeFloat,
        .ComptimeInt,
        .EnumLiteral,
        .NoReturn,
        .Undefined,
        .Null,
        => return,

        .Opaque => {
            try mod.errNoteNonLazy(src_loc, msg, "opaque type '{}' has undefined size", .{ty.fmt(sema.mod)});
        },

        .Array, .Vector => {
            // No new node is created because an array or vector can't result in infinite recursion
            try sema.explainWhyTypeIsComptimeInner(block, src, msg, src_loc, ty.elemType(), type_set);
        },
        .Pointer => {
            const elem_ty = ty.elemType2();
            if (elem_ty.zigTypeTag() == .Fn) {
                const fn_info = elem_ty.fnInfo();
                if (fn_info.is_generic) {
                    try mod.errNoteNonLazy(src_loc, msg, "function is generic", .{});
                }
                switch (fn_info.cc) {
                    .Inline => try mod.errNoteNonLazy(src_loc, msg, "function has inline calling convention", .{}),
                    else => {},
                }
                if (fn_info.return_type.comptimeOnly()) {
                    try mod.errNoteNonLazy(src_loc, msg, "function has a comptime-only return type", .{});
                }
                return;
            }
            // No new node is created because a pointer can't result in infinite recursion
            try sema.explainWhyTypeIsComptimeInner(block, src, msg, src_loc, ty.elemType(), type_set);
        },

        .Optional => {
            var buf: Type.Payload.ElemType = undefined;
            // No new node is created because an optional can't result in infinite recursion
            try sema.explainWhyTypeIsComptimeInner(block, src, msg, src_loc, ty.optionalChild(&buf), type_set);
        },
        .ErrorUnion => {
            // No new node is created because an error union can't result in infinite recursion
            try sema.explainWhyTypeIsComptimeInner(block, src, msg, src_loc, ty.errorUnionPayload(), type_set);
        },

        .Struct => {
            if (ty.castTag(.@"struct")) |payload| {
                const struct_obj = payload.data;
                for (struct_obj.fields.values()) |field, i| {
                    const field_src_loc = struct_obj.fieldSrcLoc(sema.mod, .{
                        .index = i,
                        .range = .type,
                    });

                    if ((try type_set.getOrPut(ty)).found_existing) return;

                    if (try sema.typeRequiresComptime(block, src, field.ty)) {
                        try mod.errNoteNonLazy(field_src_loc, msg, "struct requires comptime because of this field", .{});
                        try sema.explainWhyTypeIsComptimeInner(block, src, msg, field_src_loc, field.ty, type_set);
                    }
                }
            }
            // TODO tuples
        },

        .Union => {
            if (ty.cast(Type.Payload.Union)) |payload| {
                const union_obj = payload.data;
                for (union_obj.fields.values()) |field, i| {
                    const field_src_loc = union_obj.fieldSrcLoc(sema.mod, .{
                        .index = i,
                        .range = .type,
                    });

                    if ((try type_set.getOrPut(ty)).found_existing) return;

                    if (try sema.typeRequiresComptime(block, src, field.ty)) {
                        try mod.errNoteNonLazy(field_src_loc, msg, "union requires comptime because of this field", .{});
                        try sema.explainWhyTypeIsComptimeInner(block, src, msg, field_src_loc, field.ty, type_set);
                    }
                }
            }
        },
    }
}
``` 
```

**Comment by Vexu** [comment]
Here's mine:
```diff
diff --git a/src/Sema.zig b/src/Sema.zig
index 2721ed517..9fa096048 100644
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -19670,6 +19670,8 @@ fn validateRunTimeType(
     };
 }
 
+const TypeMap = std.HashMapUnmanaged(Type, void, Type.HashContext64, std.hash_map.default_max_load_percentage);
+
 fn explainWhyTypeIsComptime(
     sema: *Sema,
     block: *Block,
@@ -19677,6 +19679,21 @@ fn explainWhyTypeIsComptime(
     msg: *Module.ErrorMsg,
     src_loc: Module.SrcLoc,
     ty: Type,
+) CompileError!void {
+    var map = TypeMap{};
+    defer map.deinit(sema.gpa);
+
+    try sema.explainWhyTypeIsComptimeExtra(block, src, msg, src_loc, ty, &map);
+}
+
+fn explainWhyTypeIsComptimeExtra(
+    sema: *Sema,
+    block: *Block,
+    src: LazySrcLoc,
+    msg: *Module.ErrorMsg,
+    src_loc: Module.SrcLoc,
+    ty: Type,
+    map: *TypeMap,
 ) CompileError!void {
     const mod = sema.mod;
     switch (ty.zigTypeTag()) {
@@ -19714,7 +19731,7 @@ fn explainWhyTypeIsComptime(
         },
 
         .Array, .Vector => {
-            try sema.explainWhyTypeIsComptime(block, src, msg, src_loc, ty.elemType());
+            try sema.explainWhyTypeIsComptimeExtra(block, src, msg, src_loc, ty.elemType(), map);
         },
         .Pointer => {
             const elem_ty = ty.elemType2();
@@ -19732,18 +19749,21 @@ fn explainWhyTypeIsComptime(
                 }
                 return;
             }
-            try sema.explainWhyTypeIsComptime(block, src, msg, src_loc, ty.elemType());
+            try sema.explainWhyTypeIsComptimeExtra(block, src, msg, src_loc, ty.elemType(), map);
         },
 
         .Optional => {
             var buf: Type.Payload.ElemType = undefined;
-            try sema.explainWhyTypeIsComptime(block, src, msg, src_loc, ty.optionalChild(&buf));
+            try sema.explainWhyTypeIsComptimeExtra(block, src, msg, src_loc, ty.optionalChild(&bu
f), map);
         },
         .ErrorUnion => {
-            try sema.explainWhyTypeIsComptime(block, src, msg, src_loc, ty.errorUnionPayload());
+            try sema.explainWhyTypeIsComptimeExtra(block, src, msg, src_loc, ty.errorUnionPayload
(), map);
         },
 
         .Struct => {
+            const gop = try map.getOrPutContext(sema.gpa, ty, .{ .mod = sema.mod });
+            if (gop.found_existing) return;
+
             if (ty.castTag(.@"struct")) |payload| {
                 const struct_obj = payload.data;
                 for (struct_obj.fields.values()) |field, i| {
@@ -19753,7 +19773,7 @@ fn explainWhyTypeIsComptime(
                     });
                     if (try sema.typeRequiresComptime(block, src, field.ty)) {
                         try mod.errNoteNonLazy(field_src_loc, msg, "struct requires comptime beca
use of this field", .{});
-                        try sema.explainWhyTypeIsComptime(block, src, msg, field_src_loc, field.t
y);
+                        try sema.explainWhyTypeIsComptimeExtra(block, src, msg, field_src_loc, fi
eld.ty, map);
                     }
                 }
             }
@@ -19761,6 +19781,9 @@ fn explainWhyTypeIsComptime(
         },
 
         .Union => {
+            const gop = try map.getOrPutContext(sema.gpa, ty, .{ .mod = sema.mod });
+            if (gop.found_existing) return;
+
             if (ty.cast(Type.Payload.Union)) |payload| {
                 const union_obj = payload.data;
                 for (union_obj.fields.values()) |field, i| {
@@ -19770,7 +19793,7 @@ fn explainWhyTypeIsComptime(
                     });
                     if (try sema.typeRequiresComptime(block, src, field.ty)) {
                         try mod.errNoteNonLazy(field_src_loc, msg, "union requires comptime becau
se of this field", .{});
-                        try sema.explainWhyTypeIsComptime(block, src, msg, field_src_loc, field.t
y);
+                        try sema.explainWhyTypeIsComptimeExtra(block, src, msg, field_src_loc, fi
eld.ty, map);
                     }
                 }
             }
```
But I'm not sure whether this is the best way to solve this vs simplifying `explainWhyTypeIsComptime` or what the desired behavior on recursion is.

antlilja:
That version is definitely much simpler

erikarvstedt:
This doesn't recurse into inner structs, maybe due to lazy struct type evaluation (I'm not familiar with the internals).
In this example, `struct_obj.fields.count()` of `S2` is 0 instead of 1 (in `explainWhyTypeIsComptimeInner`), so field `b` isn't processed and the `use '*const {}' for a function pointer type` msg is missing.
```zig
const S1 = struct {
    a: S2,
};

const S2 = struct {
    b: fn () void,
};

test {
    var s: S1 = undefined;
    _ = s;
}
```

erikarvstedt:
The expected test case output needs to be updated due to commit `resolve type fully before call to explainWhyTypeIsComptime`.

erikarvstedt:
Test case `self_referential_struct_requires_comptime_multiple_occurrences.zig` is in fact not self-referential. I think it can be deleted.

erikarvstedt:
LGTM.
This PR can be squashed into 2 commits.

PR #12343, opened 2022-08-06
autodoc: only modify the DOM once to display the search results

rudedogg:
This changes how the documentation viewer creates search result items. By building up the HTML for the search results, and adding them to the DOM in one operation, it improves the performance when rendering many search results.

**Measurements:**
Starting from a freshly loaded page, and searching "a" (returning ~10,000 results) it takes ~70ms, versus ~1.9s previously.

**Old:**
image

**New:**
image

Unfortunately after this task, the browser recalculates styles and does a layout, taking another ~600ms or so. I'm still looking into what's causing that, but for now this is a small code change with a big improvement.

### DIFF
--- a/lib/docs/main.js
+++ b/lib/docs/main.js
@@ -3326,29 +3326,28 @@ var zigAnalysis;
     }
 
     if (matchedItems.length !== 0) {
-      resizeDomList(
-        domListSearchResults,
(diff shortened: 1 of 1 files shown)


kristoff-it:
Thanks!

Issue #12240, opened 2022-07-25
stage2: cannot determine length of array in comptime block


ominitay:
### Zig Version

0.10.0-dev.3313+cff5d9c80 (stage2)

### Steps to Reproduce

Attempt to compile one of the following examples:
```zig
var bytes: [1024]u8 = undefined;

export const length = bytes.len;
```
```zig
var bytes: [1024]u8 = undefined;

comptime { _ = bytes.len; }
```

### Expected Behavior

The examples compile successfully (or, more accurately, errors out because there's no main function).

### Actual Behavior

The examples both reach a compile error, of the form:
```
test.zig:3:23: error: cannot load runtime value in comptime block
export const length = bytes.len;
                      ^~~~~~~~~~~
```

Hint: The following examples do not reach a compile error:

```zig
const bytes: [1024]u8 = undefined;

export const length = bytes.len;
```
```zig
var bytes: [1024]u8 = undefined;

test { comptime _ = bytes.len; }
```

ominitay:
Thank you @topolarity for helping to reduce this!

andrewrk:
Related:
 * #12215
 * #12638

Compile error is emitted on the `decl_val` before it has a chance to look at the `len` field:

```
    > %9 = decl_val("bytes") 
      %10 = field_val(%9, "len") 
```

Resolution is unclear. Will require some reworking of AstGen/ZIR with regards to the `decl_val` instruction.

Vexu:
I tried solving this in AstGen today and while I was able to get all the examples in the issue working it is still easy to break for example by adding a cast `@as([1024u8, bytes)`. Maybe it'd be better to allow runtime instructions in outside of function bodies and rely on other mechanisms to get comptime known where they are needed as stage1 seems to be doing.

This approach could also result in better compile errors by turning the error in:
```zig
var cc: std.builtin.CallingConvention = .C;
export fn foo() callconv(cc) void {}
```
from 
```
error: cannot load runtime value in comptime block
export fn foo() callconv(cc) void {}
                         ^~
```
to

```
error: unable to resolve comptime value
export fn foo() callconv(cc) void {}
                         ^~
note: calling convention must be comptime-known
```

nektro:
`bytes.len` should be comptime known though, the issue is that there's any error at all

Vexu:
> `bytes.len` should be comptime known though, the issue is that there's any error at all

[...]

PR #12348, opened 2022-08-06
crypto/benchmark - replace testing allocator

kassane:
Recently in my latest tests there has been an error with "testing.allocator".

```bash
error: Cannot use testing allocator outside of test block
        @compileError("Cannot use testing allocator outside of test block");
```

cc: @jedisct1

### DIFF
--- a/lib/std/crypto/benchmark.zig
+++ b/lib/std/crypto/benchmark.zig
@@ -317,15 +317,13 @@ const pwhashes = [_]CryptoPwhash{
 };
 
 fn benchmarkPwhash(
+    allocator: mem.Allocator,
     comptime ty: anytype,
     comptime params: *const anyopaque,
     comptime count: comptime_int,
 ) !f64 {
     const password = "testpass" ** 2;
-     var arena = std.heap.ArenaAllocator.init(std.heap.page_allocator);
-     defer arena.deinit();
-    
-    const opts = .{ .allocator = arena.allocator(), .params = @ptrCast(*const ty.Params, params).*, .encoding = .phc };
+    const opts = .{ .allocator = allocator, .params = @ptrCast(*const ty.Params, params).*, .encoding = .phc };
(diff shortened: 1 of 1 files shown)


jedisct1:
Why not use the GPA?

kassane:
> Why not use the GPA?

GPA does not affect performance?

andrewrk (changes requested):
main() sets up an allocator; this one should be used for everything. If you don't want to use the fixed buffer allocator then use the pattern of arena wrapping page:

https://github.com/ziglang/zig/blob/3e2defd36c0bf90a0604a7618f57beaa4077139c/lib/build_runner.zig#L13-L17

jedisct1:
The function benchmarks `scrypt` and `argon2`. Both intentionally require *a lot* of memory. With a fixed buffer, they are likely to immediately return with an error, and I doubt the benchmark result is relevant.

kassane:
> The function benchmarks `scrypt` and `argon2`. Both intentionally require _a lot_ of memory. With a fixed buffer, they are likely to immediately return with an error, and I doubt the benchmark result is relevant.

changed to arena.

kassane:
Found another error when trying to use stage2 during the build.

Returning this error:
```bash
benchmark.zig:186:39: error: cannot load runtime value in comptime block
            try Signature.verifyBatch(batch.len, batch);
```
I tried to make `comptime var batch`, however a runtime problem occurs with `error: weakPublicKey` on stage1.

And returning this error on stage2:
```bash
~/.local/lib/zig/std/crypto/scrypt.zig:465:47: error: cannot load runtime value in comptime block
        const salt = crypt_format.saltFromBin(salt_bin.len, salt_bin);
```

jedisct1:
Looks like a regression. `batch.len` should be (and used to be) known at comptime.

ominitay:
#12240

PR #12351, opened 2022-08-06
std.io.Reader: bounded array functions

InKryption:
* Add readIntoBoundedBytes
 * Add readBoundedBytes

I keep re-implementing these functions in variations of "util.zig" files, and it would be nice if they were instead just part of the API.

### DIFF
--- a/lib/std/io/reader.zig
+++ b/lib/std/io/reader.zig
@@ -247,6 +247,27 @@ pub fn Reader(
             return bytes;
         }
 
+        /// Reads bytes into the bounded array, until
+        /// the bounded array is full, or the stream ends.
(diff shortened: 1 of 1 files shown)


InKryption:
I'm not sure the test failure on ci.ziglang.org is caused by this PR, it doesn't change any code, only adds two functions to the otherwise unchanged `std.io.Reader` interface.

andrewrk:
The CI test failure is an undiagnosed, flaky bug that made it into master branch, and fixing it is blocked by another bug where stage3 stack traces aren't showing up sometimes, so it's hard to see where the problem is happening. Sorry for the trouble, will try to fix it soon. It's a bit chaotic switching to the new compiler.

PR #12353, opened 2022-08-06
doc: mention explicitly about many item pointers to constant items

jjdredd:


andrewrk (changes requested):
Why only mention const when volatile is another qualifier? There is also an alignment qualifier and allowzero qualifier, as well as bit offset qualifiers. I think better to omit all of them or include all of them.

jjdredd:
Good point, I'll try to improve it.

PR #12354, opened 2022-08-06
Add compile error for runtime parameter in function returning a comptime only type.

### LINKED ISSUE #12333: Disallow runtime parameters in functions returning comptime-only types
### Steps to Reproduce

```
fn F(val: anytype) type {
    _ = val;
    return struct {};
}

test {
    _ = F(void{});
    _ = F(@as(i32, 0));
    // Compile error:
    // var i: i32 = 0;
    // _ = F(i);
}
```

### Expected Behavior

There ought to be a compile error for declaring a non-`comptime` parameter in a function returning a `comptime`-only type, such as `type`, `comptime_int`, or `comptime_float`, being that there is no way to compute any of them using runtime values. Allowing this only enables useless permutations of compile-able code.

### Actual Behavior

Compilation succeeds, unless you uncomment the last two lines of the test.

antlilja:
Closes #12333

This PR implements a compile error for when a function which return type is comptime only takes a parameter which might be a runtime parameter.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -7816,19 +7816,17 @@ fn funcCommon(
             };
         }
 
-        var is_comptime_ret = false;
-        const ret_poison = if (!is_generic) rp: {
-            if (sema.typeRequiresComptime(block, ret_ty_src, bare_return_type)) |ret_comptime| {
-                is_comptime_ret = ret_comptime;
-                break :rp bare_return_type.tag() == .generic_poison;
-            } else |err| switch (err) {
-                error.GenericPoison => {
-                    is_generic = true;
-                    break :rp true;
-                },
-                else => |e| return e,
-            }
-        } else bare_return_type.tag() == .generic_poison;
+        var ret_ty_requires_comptime = false;
+        const ret_poison = if (sema.typeRequiresComptime(block, ret_ty_src, bare_return_type)) |ret_comptime| rp: {
+            ret_ty_requires_comptime = ret_comptime;
+            break :rp bare_return_type.tag() == .generic_poison;
+        } else |err| switch (err) {
+            error.GenericPoison => rp: {
+                is_generic = true;
+                break :rp true;
+            },
+            else => |e| return e,
+        };
 
         const return_type = if (!inferred_error_set or ret_poison)
             bare_return_type
@@ -7870,6 +7868,41 @@ fn funcCommon(
                 try sema.addDeclaredHereNote(msg, return_type);
                 break :msg msg;
             };
+            return sema.failWithOwnedErrorMsg(msg);
+        }
+
+        // If the return type is comptime only but not dependent on parameters then all parameter types also need to be comptime
+        if (!sema.is_generic_instantiation and has_body and ret_ty_requires_comptime) comptime_check: {
+            for (block.params.items) |param| {
+                if (!param.is_comptime) break;
+            } else break :comptime_check;
+
+            const msg = try sema.errMsg(
+                block,
+                ret_ty_src,
+                "function with comptime only return type '{}' requires all parameters to be comptime",
+                .{return_type.fmt(sema.mod)},
+            );
+            try sema.explainWhyTypeIsComptime(block, ret_ty_src, msg, ret_ty_src.toSrcLoc(sema.owner_decl), return_type);
+
+            const tags = sema.code.instructions.items(.tag);
+            const data = sema.code.instructions.items(.data);
+            const param_body = sema.code.getParamBody(func_inst);
+            for (block.params.items) |param, i| {
+                if (!param.is_comptime) {
+                    const param_index = param_body[i];
+                    const param_src = switch (tags[param_index]) {
+                        .param => data[param_index].pl_tok.src(),
+                        .param_anytype => data[param_index].str_tok.src(),
+                        else => unreachable,
+                    };
+                    if (param.name.len != 0) {
+                        try sema.errNote(block, param_src, msg, "param '{s}' is required to be comptime", .{param.name});
+                    } else {
+                        try sema.errNote(block, param_src, msg, "param is required to be comptime", .{});
+                    }
+                }
+            }
             return sema.failWithOwnedErrorMsg(msg);
         }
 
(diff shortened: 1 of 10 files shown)


rohlem:
What is this PR's behavior with dependent return types? Contrived example:

```zig
fn orelse_undefined(comptime T: type, value: ?T) T {
  return value orelse undefined;
}
```

I think we should accept a function declaration that __may__ return a non-comptime-only type in one generic instantiation.
I.e. both `maybe_undefiend(i32, 4)` and `maybe_undefiend(type, i32)` are valid use case IMO.

Although in continuing this thought, if the return type is defined differently depending on comptime-logic (like target architecture), that aspect becomes even more difficult to localize.
I'll copy this comment to the original issue for further discussion separate from this PR's implementation.

antlilja:
Running the compile error test cases with this branch seems to crash the compiler when a struct is declared in the return type of a function. For example `Zir.extraData`.

Minimal example: 
```zig
fn f(comptime T: type) struct { data: T } {
    return .{ .data = undefined };
}

test {
    _ = f(void);
}
```

```
thread 15889 panic: reached unreachable code
Analyzing boo.zig: boo.zig:test_0
      %21 = dbg_block_begin())
      %22 = dbg_stmt(2, 5)
    > %23 = decl_val("f") token_offset:6:9 to :6:10
      %24 = dbg_stmt(2, 10)
      %25 = param_type(%23, 0)
      %26 = call(.auto, %23, [@Zir.Inst.Ref.void_type]) node_offset:6:9 to :6:16
      %28 = dbg_block_end())
      %27 = ensure_result_non_error(%26) node_offset:6:9 to :6:16
      %29 = ret_tok(@Zir.Inst.Ref.void_value) token_offset:7:1 to :7:2
      zig ast-check -t boo.zig


/home/anton/dev/zig/zig/src/Module.zig:4152:32: 0xa756d6 in Module.ensureDeclAnalyzed (zig)
        error.GenericPoison => unreachable,
                               ^
/home/anton/dev/zig/zig/src/Sema.zig:25171:32: 0x109765f in Sema.ensureDeclAnalyzed (zig)
    sema.mod.ensureDeclAnalyzed(decl_index) catch |err| {
                               ^
/home/anton/dev/zig/zig/src/Sema.zig:25213:32: 0xeae224 in Sema.analyzeDeclRef (zig)
    try sema.ensureDeclAnalyzed(decl_index);
                               ^
/home/anton/dev/zig/zig/src/Sema.zig:25147:41: 0x1078eba in Sema.analyzeDeclVal (zig)
    const decl_ref = sema.analyzeDeclRef(decl_index) catch |err| switch (err) {
                                        ^
/home/anton/dev/zig/zig/src/Sema.zig:5331:31: 0xe17f18 in Sema.zirDeclVal (zig)
    return sema.analyzeDeclVal(block, src, decl);
                              ^
/home/anton/dev/zig/zig/src/Sema.zig:728:65: 0xc7e252 in Sema.analyzeBodyInner (zig)
            .decl_val                     => try sema.zirDeclVal(block, inst),
                                                                ^
/home/anton/dev/zig/zig/src/Sema.zig:603:30: 0xc715aa in Sema.analyzeBody (zig)
    _ = sema.analyzeBodyInner(block, body) catch |err| switch (err) {
                             ^
/home/anton/dev/zig/zig/src/Module.zig:5517:21: 0xa9181d in Module.analyzeFnBody (zig)
    sema.analyzeBody(&inner_block, fn_info.body) catch |err| switch (err) {
                    ^
/home/anton/dev/zig/zig/src/Module.zig:4227:40: 0xa74290 in Module.ensureFuncBodyAnalyzed (zig)
            var air = mod.analyzeFnBody(func, sema_arena) catch |err| switch (err) {
                                       ^
/home/anton/dev/zig/zig/src/Compilation.zig:2962:42: 0x7b32ea in Compilation.processOneJob (zig)
            module.ensureFuncBodyAnalyzed(func) catch |err| switch (err) {
                                         ^
/home/anton/dev/zig/zig/src/Compilation.zig:2894:30: 0x79c74d in Compilation.performAllTheWork (zig)
            try processOneJob(comp, work_item);
                             ^
/home/anton/dev/zig/zig/src/Compilation.zig:2234:31: 0x794c43 in Compilation.update (zig)
                              ^
/home/anton/dev/zig/zig/src/main.zig:3326:20: 0x7036af in updateModule (zig)
    try comp.update();
                   ^
/home/anton/dev/zig/zig/src/main.zig:3015:17: 0x6cbe91 in buildOutputType (zig)
                ^
/home/anton/dev/zig/zig/src/main.zig:236:31: 0x6afcd6 in mainArgs (zig)
                              ^
/home/anton/dev/zig/zig/src/main.zig:174:20: 0x6aed24 in main (zig)
                   ^
/home/anton/dev/zig/zig/lib/std/start.zig:576:37: 0x6b164f in std.start.main (zig)
```

antlilja:
> What is this PR's behavior with dependent return types? Contrived example:
> 
> ```zig
> fn orelse_undefined(comptime T: type, value: ?T) T {
>   return value orelse undefined;
> }
> ```
> 
> I think we should accept a function declaration that **may** return a non-comptime-only type in one generic instantiation. I.e. both `maybe_undefiend(i32, 4)` and `maybe_undefiend(type, i32)` are valid use case IMO.
> 
> Although in continuing this thought, if the return type is defined differently depending on comptime-logic (like target architecture), that aspect becomes even more difficult to localize. [I've restated this comment in the original issue for further discussion](https://github.com/ziglang/zig/issues/12333#issuecomment-1207300685) separate from this PR's implementation.

This PR does not give compile errors for dependent return types.

Vexu:
> Running the compile error test cases with this branch seems to crash the compiler when a struct is declared in the return type of a function. For example `Zir.extraData`.

I'm pretty sure you need to check `ret_poison` before calling `typeRequiresComptime`.

antlilja:
The errors are fixed and there is now a check in order to exclude dependent return types

src/Sema.zig
```diff
@@ -7720,6 +7724,41 @@ fn funcCommon(
             });
         };
 
+        // If the return type is comptime only but not dependent on parameters then all parameter types also need to be comptime
+        if (!is_dependent_ret_ty and try sema.typeRequiresComptime(block, ret_ty_src, return_type)) {
```
Vexu, line 7728: `is_dependent_ret_ty` should always be the same as `ret_poison`.

antlilja: They do not always seem to be the same, and the test cases do not build when using `ret_poison` instead of `is_dependent_ret_ty`.

Vexu: Hmm, using `ret_poison` does seem to break something but the current solution is unable to handle even simple cases like:
```zig
const A = comptime_int;
fn foo(a: u32) A {
    _ = a;
    return undefined;
}
comptime {
    _ = foo;
}
```
Maybe it'd be better to just scrap this error altogether along with the broken `parameter of type 'T' must be declared comptime`...

PR #12357, opened 2022-08-07
std.crypto: Add segmented API to Ed25519 signing.

gcoakes:
std.crypto.sign.Ed25519.sign previously required the message to be
passed as a single continuous byte slice. This adds a second function
std.crypto.sign.Ed25519.signSegmented which permits the message to be a
slice of byte slices. The original sign function simply wraps the
msg argument in a single item tuple and passes a slice of that to
signSegmented.

### DIFF
--- a/lib/std/crypto/25519/ed25519.zig
+++ b/lib/std/crypto/25519/ed25519.zig
@@ -74,10 +74,8 @@ pub const Ed25519 = struct {
         }
     };
 
-    /// Sign a message using a key pair, and optional random noise.
-    /// Having noise creates non-standard, non-deterministic signatures,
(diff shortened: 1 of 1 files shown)


jedisct1:
If such an interface is added, it should be done consistently, not just in one specific function: also for verification, for blind signatures, and other signature systems.

But a more generic approach would be to have alternative signature/verification functions requiring a `Reader` or a `SeekableStream` instead.

That would solve your issue, but also the ability to sign/verify anything that doesn't fit in memory.

By the way, the EdDSA API is going to be completely revamped in order to match the ECDSA one.

gcoakes:
I'll re-open if/when the EdDSA API is restructured.

PR #12362, opened 2022-08-08
std: add streaming API to EcDSA

gcoakes:
Expose `Verifier` and `Signer` which essentially wrap the relevant
`Hash` API, permitting messages to be streamed rather than buffered.
The existing `verify` and `sign` functions are thin wrappers over
`Verifier` and `Signer`. They remain backwards compatible.


jedisct1:
ECDSA and EdDSA now support `Signer`/`Verifier` interfaces for incremental verification.

Issue #1947, opened 2019-02-11
audit analysis of undefined values; make it clear when undefined is allowed or not


andrewrk:
In IR analysis, right now Zig is inconsistent about the semantics of an operation with an **undefined value**. This issue is to make the rules clear how it is supposed to work. Note that **undefined values** are different from **undefined behavior**.

 * An **undefined value** of type `T` takes up the same **store size** as a normal value of type `T`.
 * **Undefined values** semantically represent an extra state which is not possible to represent using any of the valid bit patterns of the underlying type. However, aside from the **store size**, the representation of an **undefined value** in memory is undefined; it can be any bit pattern. As an example, the value `u8(undefined)`, at runtime, could be any combination of bits that fits in `@sizeOf(u8)`, which is `1`. The value `bool(undefined)`, at runtime, could be any combination of bits that fits in `@sizeOf(bool)`, which is also `1`. So even though the only valid bit patterns of the type `bool` are `0b00000000` and `0b00000001`, when the value is undefined, the byte which represents the storage of the `u1` value could be anything, including `0b00000010`, `0b10101010`, or `0b11111111`. Therefore, because **undefined values** *semantically represent an extra state*, it is an incorrect assumption that an **undefined value** with type `T` has a value which is in the set of valid values for type `T`.
 * Expressions which have no side effects and no possible **undefined behavior**, and one or more of the operands has an **undefined value** which is read, the expression result is an **undefined value**. For example, the `+%` operator. Note that for slicing operator, if the start is `0`, the pointer value is not read, which makes this expression defined: `(([*]u8)(undefined))[0..0]`.
 * Branching on an **undefined value** is **undefined behavior**. This can be caught at `comptime`, and caught at runtime if #211 is solved. For example, the condition of an `if` expression.
 * Expressions which have possible **undefined behavior**, if one or more of the operands is an **undefined value** and there is any possible set of values chosen for the **undefined values** that would cause **undefined behavior** then this expression causes **undefined behavior**. For example, `@intCast(u8, u16(undefined))`. Another example: the `+` operator. However if one of the operands of `+` is `comptime`-known to be `0`, and the other is an **undefined value** the result is an **undefined value** because there is no possible value added to `0` that can cause overflow.

Every IR instruction analysis code should be audited and tests added to enforce this behavior, especially for `comptime` code.

vi:
[Documentation](https://ziglang.org/documentation/master/#undefined) may need updating to closely resemble what's actually happens in LLVM.

[...]

PR #12370, opened 2022-08-08
stage2: add compile error for compile-time ptrCast from null to non-optional pointer

### LINKED ISSUE #12367: missing compile error for compile-time ptrcast null to pointer
```zig
comptime {
    const opt_ptr: ?*i32 = null;
    const ptr = @ptrCast(*i32, opt_ptr);
    _ = ptr;
}
```

Expected:

```
./test3.zig:3:17: error: null pointer casted to type '*i32'
    const ptr = @ptrCast(*i32, opt_ptr);
                ^
```

Actual:

no compile error

InKryption:
Closes #12367

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -17274,6 +17274,15 @@ fn zirPtrCast(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air
     else
         operand;
 
+    if (try sema.resolveMaybeUndefVal(block, operand_src, operand)) |operand_val| {
+        if (!dest_ty.ptrAllowsZero() and operand_val.isUndef()) {
+            return sema.failWithUseOfUndef(block, operand_src);
+        }
+        if (!dest_ty.ptrAllowsZero() and operand_val.isNull()) {
+            return sema.fail(block, operand_src, "null pointer casted to type {}", .{dest_ty.fmt(sema.mod)});
+        }
+    }
+
     const dest_elem_ty = dest_ty.elemType2();
     try sema.resolveTypeLayout(block, dest_ty_src, dest_elem_ty);
     const dest_align = dest_ty.ptrAlignment(target);
--- a/test/cases/compile_errors/compile_time_undef_ptr_cast.zig
+++ b/test/cases/compile_errors/compile_time_undef_ptr_cast.zig
@@ -1,0 +1,11 @@
+comptime {
+    var undef_ptr: *i32 = undefined;
+    const ptr = @ptrCast(*i32, undef_ptr);
+    _ = ptr;
+}
+
+// error
+// backend=llvm
+// target=native
+//
+// :3:32: error: use of undefined value here causes undefined behavior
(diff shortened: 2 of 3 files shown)


ifreund (commented):
This needs a test case to make sure it does not regress.

src/Sema.zig
```diff
@@ -17133,6 +17133,15 @@ fn zirPtrCast(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air
     else
         operand;
 
+    if (!dest_ty.isPtrLikeOptional() and operand_ty.isPtrLikeOptional() and
+        try sema.isComptimeKnown(block, operand_src, operand))
+    {
+        const operand_val = try sema.resolveConstValue(block, operand_src, operand, "use of undefined value here causes undefined behavior");
```
Vexu, line 17139: The `reason` param is supposed to be the reason why the value has to be comptime known, "use of undefined value here causes undefined behavior" makes no sense as the reason. Also you just checked that the value is comptime known so this reason should never be used in the first place.

It would be better to just directly use `resolveMaybeUndefVal`.

Vexu (changes requested):

src/Sema.zig
InKryption: Right, didn't understand that. In that case, should it just be unwrapped like `(try sema.resolveMaybeUndefVal(block, operand_src, operand)).?` ?

Vexu: Sure, but it'd be even better to just assign that result to a constant and avoid `isComptimeKnown` all together.

InKryption: Scratch that, I see what you meant now.

test/cases/compile_errors/compile_time_undef_ptr_cast.zig
```diff
@@ -0,0 +1,13 @@
...
+        const ptr = @ptrCast(*i32, opt_ptr);
+        _ = ptr;
+    }
+}
+
+// error
+// backend=llvm
+// target=native
+//
+// :4:36: error: use of undefined value here causes undefined behavior
```
Vexu, line 13: This should not be an error.

src/Sema.zig
```diff
@@ -17133,6 +17133,15 @@ fn zirPtrCast(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air
     else
         operand;
 
+    if (try sema.resolveMaybeUndefVal(block, operand_src, operand)) |operand_val| {
+        if (operand_val.isUndef()) {
+            return sema.failWithUseOfUndef(block, operand_src);
+        }
+        if (!dest_ty.isPtrLikeOptional() and operand_val.isNull()) {
```
Vexu, line 17140: This should check `ptrAllowsZero`.

test/cases/compile_errors/compile_time_undef_ptr_cast.zig
InKryption: The stage1 compiler currently issues an error for this. Is that not according to spec?

Vexu: Stage1 is wrong, see #1947:
> Expressions which have no side effects and no possible undefined behavior, and one or more of the operands has an undefined value which is read, the expression result is an undefined value.

InKryption: According to Andrew, this should actually cause a compile error, due to the fact that a value of `undefined` indicates any value including 0, which, ptrCast should panic if that occurs.

Vexu (approved):

PR #12374, opened 2022-08-08
autodoc: links to source code

vesim987:
This PR adds links to source code for declarations.

image


haze:
I take it you took source from the rust docs as inspiration for putting the source link on the right? I'd honestly appreciate seeing it on the left (and maybe smaller if it's too visually distracting) because my cursor is usually near the declaration, and trying to track a source link to a declaration (especially in this screenshot) is hard without separators (and even still..)

vesim987:
@haze tbh, I put it on the right because there was empty space, for me there is no big difference where the link is. Feel free to discuss with others where is the best place for it.

alichraghi:
How it look like if [src] appear on hover?
because a lot of repeated buttons doesn't look good

haze:
> How it look like if [src] appear on hover?
> because a lot of repeated buttons doesn't look good

how would people new to the site figure that hovering would reveal that!

lib/docs/main.js
```diff
@@ -51,6 +51,7 @@ var zigAnalysis;
   const domHdrName = document.getElementById("hdrName");
   const domHelpModal = document.getElementById("helpModal");
   const domSearchPlaceholder = document.getElementById("searchPlaceholder");
+  const sourceFileUrlTemplate = "https://github.com/ziglang/zig/blob/master/lib/std/{{file}}#L{{line}}"
```
tauoverpi, line 54: This would break with any changes to the standard library and would need to include the exact commit to work. I'm not sure if this is the right place to put it.

Also, is it expected that this will be extended to work for project specific documentation?

Issue #10113, opened 2021-11-07
make `packed struct` always use a single backing integer, inferring it if not explicitly provided


MasterQ32:
So, everyone knows the problems we have with `packed structs`. Some are implementation problems, but we also have a huge problem with *not* having a proper specification on how they should work at all, except for "zero padding".

So i want to propose a definition for `packed struct`s that is easy to understand and implement.

Let's consider this code example:
```zig
const T = packed struct {
    a: u4,
    b: u4,
};

var bits = [_]u8 { 0x01 };
var t = @bitCast(T, bits);

std.debug.print("{}\n", .{ t });
```
What do you expect it to print?  
Is it `T{ .a = 1, .b = 0 }`? Then you have assumed a little-endian platform, as on big endian, it will print `T{ .a = 0, .b = 1 }`.  
[Check it out on Compiler Explorer](https://godbolt.org/z/GnW73hj9x).

Personally, i find this confusing, thus i propose:

A packed struct will have a similar semantic as a unsigned integer with the same amount of bits. Each field is considered an unsigned integer of the same amount of bits.

To explain the idea, we have this struct:
```zig
const T1 = packed struct { // u8
  a: u4, // occupies bits 0..3
  b: u4, // occupies bits 4..7
};

const T2 = packed struct { // u32
  a: u4, // occupies bits 0..3 in our u32
  b: u3, // occupies bits 4..6 in our u32
  c: u25, // occupies bits 7..31 in our u32
};
```

So the packed struct implements exactly what we would do in a language without packed structs:
```zig
var value: T2 = …;
var a = @truncate( u4, 0x0000000F & (value >> 0));
var b = @truncate( u3, 0x00000007 & (value >> 4));
var c = @truncate(u25, 0x01FFFFFF & (value >> 7));
```

This means that we have a predictable model for backed structs and in case of `T2`, we can reason about it:
```
const integer: u32 = 0x165652B6;
const value = @bitCast(T2, integer);
std.debug.assert(value.a == 6);
std.debug.assert(value.b == 3);
std.debug.assert(value.c == 0x2CACA5);
```

This will be true for both little and big endian hardware, so it will do what we expect, even if the bytes on disk have a different order.  
If a known byte order is required, we can use `std.mem.writeIntLittle`, `@byteSwap` and others.

In addition, we can declare a struct as `packed struct(u32)`, which will enforce the struct size to 32 bit, and we will get a compiler error if it isn't the case.

[...]

PR #12379, opened 2022-08-08
stage2: Implement explicit backing integers for packed structs

ifreund:
Now the backing integer of a packed struct type may be explicitly
specified with e.g. `packed struct(u32) { ... }`.

I haven't touched `Sema.zig` this extensively before, so any code/style pointers there are certainly appreciated.

Related: #10113, #5049

### DIFF
--- a/lib/std/zig/parse.zig
+++ b/lib/std/zig/parse.zig
@@ -3356,16 +3356,18 @@ const Parser = struct {
     }
 
     /// Caller must have already verified the first token.
+    /// ContainerDeclAuto <- ContainerDeclType LBRACE container_doc_comment? ContainerMembers RBRACE
+    ///
     /// ContainerDeclType
-    ///     <- KEYWORD_struct
+    ///     <- KEYWORD_struct (LPAREN Expr RPAREN)?
+    ///      / KEYWORD_opaque
     ///      / KEYWORD_enum (LPAREN Expr RPAREN)?
     ///      / KEYWORD_union (LPAREN (KEYWORD_enum (LPAREN Expr RPAREN)? / Expr) RPAREN)?
-    ///      / KEYWORD_opaque
     fn parseContainerDeclAuto(p: *Parser) !Node.Index {
         const main_token = p.nextToken();
         const arg_expr = switch (p.token_tags[main_token]) {
-            .keyword_struct, .keyword_opaque => null_node,
-            .keyword_enum => blk: {
+            .keyword_opaque => null_node,
+            .keyword_struct, .keyword_enum => blk: {
                 if (p.eatToken(.l_paren)) |_| {
                     const expr = try p.expectExpr();
                     _ = try p.expectToken(.r_paren);
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -78,6 +78,7 @@ post_hoc_blocks: std.AutoHashMapUnmanaged(Air.Inst.Index, *LabeledBlock) = .{},
 err: ?*Module.ErrorMsg = null,
 
 const std = @import("std");
+const math = std.math;
 const mem = std.mem;
 const Allocator = std.mem.Allocator;
 const assert = std.debug.assert;
@@ -11824,7 +11825,7 @@ fn zirModRem(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.
                     return sema.failWithDivideByZero(block, rhs_src);
                 }
                 if (maybe_lhs_val) |lhs_val| {
-                    const rem_result = try lhs_val.intRem(rhs_val, resolved_type, sema.arena, target);
+                    const rem_result = try sema.intRem(block, resolved_type, lhs_val, lhs_src, rhs_val, rhs_src);
                     // If this answer could possibly be different by doing `intMod`,
                     // we must emit a compile error. Otherwise, it's OK.
                     if ((try rhs_val.compareWithZeroAdvanced(.lt, sema.kit(block, src))) != (try lhs_val.compareWithZeroAdvanced(.lt, sema.kit(block, src))) and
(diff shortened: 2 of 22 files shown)


Jarred-Sumner:
This is exciting

Will probably simplify a lot of code like this:
```zig
/// https://nodejs.org/api/os.html#osplatform
pub const OperatingSystem = enum(u16) {
    none = 0,
    all = all_value,

    _,

    pub const aix: u16 = 1 << 1;
    pub const darwin: u16 = 1 << 2;
    pub const freebsd: u16 = 1 << 3;
    pub const linux: u16 = 1 << 4;
    pub const openbsd: u16 = 1 << 5;
    pub const sunos: u16 = 1 << 6;
    pub const win32: u16 = 1 << 7;
    pub const android: u16 = 1 << 8;

    pub const all_value: u16 = aix | darwin | freebsd | linux | openbsd | sunos | win32 | android;

    pub fn isMatch(this: OperatingSystem) bool {
        if (comptime Environment.isLinux) {
            return (@enumToInt(this) & linux) != 0;
        } else if (comptime Environment.isMac) {
            return (@enumToInt(this) & darwin) != 0;
        } else {
            return false;
        }
    }

    const NameMap = ComptimeStringMap(u16, .{
        .{ "aix", aix },
        .{ "darwin", darwin },
        .{ "freebsd", freebsd },
        .{ "linux", linux },
        .{ "openbsd", openbsd },
        .{ "sunos", sunos },
        .{ "win32", win32 },
        .{ "android", android },
    });

    pub fn apply(this_: OperatingSystem, str: []const u8) OperatingSystem {
        if (str.len == 0) {
            return this_;
        }
        const this = @enumToInt(this_);

        const is_not = str[0] == '!';
        const offset: usize = if (str[0] == '!') 1 else 0;

        const field: u16 = NameMap.get(str[offset..]) orelse return this_;

        if (is_not) {
            return @intToEnum(OperatingSystem, this & ~field);
        } else {
            return @intToEnum(OperatingSystem, this | field);
        }
    }
};
```

src/Sema.zig
```diff
@@ -27442,19 +27498,23 @@ fn semaStructFields(mod: *Module, struct_obj: *Module.Struct) CompileError!void
...
             return sema.failWithOwnedErrorMsg(msg);
-        } else if (struct_obj.layout == .Packed and !(validatePackedType(field.ty))) {
-            const msg = msg: {
-                const tree = try sema.getAstTree(&block_scope);
-                const fields_src = enumFieldSrcLoc(decl, tree.*, struct_obj.node_offset, i);
-                const msg = try sema.errMsg(&block_scope, fields_src, "packed structs cannot contain fields of type '{}'", .{field.ty.fmt(sema.mod)});
-                errdefer msg.destroy(sema.gpa);
+        } else if (struct_obj.layout == .Packed) {
+            const tree = try sema.getAstTree(&block_scope);
+            const field_src = enumFieldSrcLoc(decl, tree.*, struct_obj.node_offset, i);
```
Vexu, line 27503: `enumFieldSrcLoc` loads the AST and should avoided in the non-error path. Instead you should pass `.unneeded` and then check if the function returns `error.SourceLocationNeeded` and then call the function again with the proper source location.

lib/std/zig/parse.zig
```diff
@@ -3356,6 +3356,13 @@ const Parser = struct {
     }
 
     /// Caller must have already verified the first token.
+    /// ContainerDeclAuto <- ContainerDeclType LBRACE container_doc_comment? ContainerMembers RBRACE
+    ///
+    /// ContainerDeclType
```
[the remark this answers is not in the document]
ifreund: Done!

src/Sema.zig
ifreund: Good to know, I ended up removing this code as I found some deeper issues with my previous approach.

ifreund:
I found some underlying issues with my previous approach and have just pushed an implementation that should be much more correct. The ZIR encoding now handles the case where the type requires a body properly and the layout of packed structs is not force-resolved while resolving their fields.

However, I have run into an issue I'm not sure how to solve. With my changes `zig build test-std` crashes with the following output:
```
Semantic Analysis [5743] min... thread 27123 panic: reached unreachable code
Analyzing /home/ifreund/projects/zig/lib/std/mem.zig: mem.zig:BytesAsSliceReturnType
      %21196 = decl_ref("meta") 
      %21197 = field_call_bind(%21196, "Child") 
      %21198 = dbg_stmt(6, 66)
      %21199 = param_type(%21197, 0)
      %21200 = call(.compile_time, %21197, [%21136]) 
      %21201 = type_info(%21200) 
      %21202 = field_val(%21201, "Array") 
      %21203 = field_val(%21202, "len") 
      %21204 = size_of(%21134) 
    > %21205 = mod_rem(%21203, %21204) 
      %21206 = cmp_neq(%21205, @Zir.Inst.Ref.zero) 
      %21207 = as_node(@Zir.Inst.Ref.bool_type, %21206) 
      %21208 = break_inline(%21195, %21207)
      zig ast-check -t /home/ifreund/projects/zig/lib/std/mem.zig

  in /home/ifreund/projects/zig/lib/std/mem.zig: mem.zig:BytesAsSliceReturnType
    > %21195 = bool_br_and(%21194, {%21196..%21208}
  in /home/ifreund/projects/zig/lib/std/mem.zig: mem.zig:BytesAsSliceReturnType
    > %21211 = block({%21185..%21210}) 
  in /home/ifreund/projects/zig/lib/std/mem.zig: mem.zig:bytesAsSlice__anon_34119
    > %21253 = call(.compile_time, %21247, [%21245, %21251]) 
  in /home/ifreund/projects/zig/lib/std/mem.zig: mem.zig:bytesAsSlice__anon_34119
    > %21301 = func(ret_ty={%21247..%21254}, body={%21255..%21299}) (lbrace=1:97,rbrace=11:1) 
  in /home/ifreund/projects/zig/lib/std/mem.zig: mem.zig:test.bytesAsSlice on a packed struct
    > %21623 = call(.auto, %21619, [%21604, %21607]) 

/home/ifreund/projects/zig/lib/std/debug.zig:281:14: 0x2ec5708 in std.debug.assert (zig)
             ^
/home/ifreund/projects/zig/src/type.zig:3211:27: 0x30ab2eb in type.Type.abiSizeAdvanced (zig)
                    assert(struct_obj.haveLayout());
                          ^
/home/ifreund/projects/zig/src/type.zig:3148:32: 0x30a165b in type.Type.abiSize (zig)
        return (abiSizeAdvanced(ty, target, .eager) catch unreachable).scalar;
                               ^
/home/ifreund/projects/zig/src/value.zig:1101:37: 0x34e2b21 in value.Value.toBigIntAdvanced (zig)
                const x = ty.abiSize(target);
                                    ^
/home/ifreund/projects/zig/src/value.zig:1058:36: 0x32f692a in value.Value.toBigInt (zig)
        return val.toBigIntAdvanced(space, target, null) catch unreachable;
                                   ^
/home/ifreund/projects/zig/src/value.zig:3492:40: 0x3aaae6f in value.Value.intRemScalar (zig)
        const rhs_bigint = rhs.toBigInt(&rhs_space, target);
                                       ^
/home/ifreund/projects/zig/src/value.zig:3483:28: 0x38da9d2 in value.Value.intRem (zig)
        return intRemScalar(lhs, rhs, allocator, target);
                           ^
/home/ifreund/projects/zig/src/Sema.zig:11766:58: 0x3691416 in Sema.zirModRem (zig)
                    const rem_result = try lhs_val.intRem(rhs_val, resolved_type, sema.arena, target);
                                                         ^
/home/ifreund/projects/zig/src/Sema.zig:890:45: 0x34a60be in Sema.analyzeBodyInner (zig)
            .mod_rem   => try sema.zirModRem(block, inst),
                                            ^
/home/ifreund/projects/zig/src/Sema.zig:620:45: 0x349872c in Sema.analyzeBodyBreak (zig)
    const break_inst = sema.analyzeBodyInner(block, body) catch |err| switch (err) {
                                            ^
/home/ifreund/projects/zig/src/Sema.zig:585:50: 0x38922a0 in Sema.resolveBody (zig)
    const break_data = (try sema.analyzeBodyBreak(block, body)) orelse
                                                 ^
/home/ifreund/projects/zig/src/Sema.zig:14483:32: 0x36335bc in Sema.zirBoolBr (zig)
        return sema.resolveBody(parent_block, body, inst);
                               ^
```
The `size_of(%21134)` instruction creates a value with tag `.lazy_size` as it is analyzing a packed struct which has not yet had its layout resolved. However, the current implementation of `mod_rem` does not try to resolve lazy values but rather asserts that they are already resolved. It seems to me that the fix would be adding a `Value.intRemAdvanced()` variant that takes a sema "kit" and tries to resolve lazy values. Is this correct?

andrewrk:
Hmmm, that's a good question. In the interest of managing complexity, I think I would suggest a different approach, based on the observation that the intRem function is only ever called from Sema:

```
andy@ark ~/D/zig (master)> grep -RI intRem src/
src/value.zig:    pub fn intRem(lhs: Value, rhs: Value, ty: Type, allocator: Allocator, target: Target) !Value {
src/value.zig:                scalar.* = try intRemScalar(lhs.indexVectorlike(i), rhs.indexVectorlike(i), allocator, target);
src/value.zig:        return intRemScalar(lhs, rhs, allocator, target);
src/value.zig:    pub fn intRemScalar(lhs: Value, rhs: Value, allocator: Allocator, target: Target) !Value {
src/Sema.zig:                    const rem_result = try lhs_val.intRem(rhs_val, resolved_type, sema.arena, target);
src/Sema.zig:                        try lhs_val.intRem(rhs_val, resolved_type, sema.arena, target),
```

The idea being to move this function to Sema, and accept sema, block, src arguments. I made this change to other functions recently and it was a satisfactory approach.

kristoff-it:
Thanks for the update to Autodoc!

ifreund:
> Thanks for the update to Autodoc!

Not breaking it is the least I can do :)

andrewrk (approved):
Nice work!

Issue #12380, opened 2022-08-08
Windows Azure CI has failures but is still reporting success


andrewrk:
Check any of the recent master branch commits that have a green check, find the Azure CI windows job, look at the raw log, and search for "failed". Some of the commands in this pipelines.yml are failing and yet the job is reporting success.

cc @hdorio

andrewrk:
Fixed with fa50e179f7f8d523ff00be4cac90bf7659394140. See the above linked issues for what tests have been failing all this time.

PR #12381, opened 2022-08-09
ci: windows: stop when exe exits with an error

hdorio:
Issue #12380

Check the last status of `zig.exe` and exit if an error occurred.
The last code that I wrote wrongly assumed that [ErrorActionPreference](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/powershell?view=azure-devops#:~:text=errorActionPreference-,ErrorActionPreference,-(Optional)%20When%20not) would catch executable errors.

### DIFF
--- a/ci/azure/pipelines.yml
+++ b/ci/azure/pipelines.yml
@@ -41,6 +41,13 @@ jobs:
       Set-Variable -Name ZIGINSTALLDIR -Value "${ZIGBUILDDIR}\dist"
       Set-Variable -Name ZIGPREFIXPATH -Value "$(Get-Location)\$(ZIG_LLVM_CLANG_LLD_NAME)"
 
+      function CheckLastExitCode {
+        if (!$?) {
+          exit 1
+        }
+        return 0
+      }
+
       # Make the `zig version` number consistent.
       # This will affect the `zig build` command below which uses `git describe`.
       git config core.abbrev 9
@@ -69,6 +76,7 @@ jobs:
         -Dstrip `
         -Duse-zig-libcxx `
         -Dtarget=$(TARGET)
+      CheckLastExitCode
 
       cd -
 
@@ -83,19 +91,37 @@ jobs:
   - pwsh: |
       Set-Variable -Name ZIGINSTALLDIR -Value "$(Get-Location)\build\dist"
 
+      function CheckLastExitCode {
+        if (!$?) {
+          exit 1
+        }
+        return 0
+      }
+
       # Sadly, stage2 is omitted from this build to save memory on the CI server. Once self-hosted is
       # built with itself and does not gobble as much memory, we can enable these tests.
       #& "$ZIGINSTALLDIR\bin\zig.exe" test "..\test\behavior.zig" -fno-stage1 -fLLVM -I "..\test" 2>&1
+      #CheckLastExitCode
 
       & "$ZIGINSTALLDIR\bin\zig.exe" build test-toolchain -Dskip-non-native -Dskip-stage2-tests 2>&1
+      CheckLastExitCode
       & "$ZIGINSTALLDIR\bin\zig.exe" build test-std -Dskip-non-native  2>&1
+      CheckLastExitCode
     name: test
     displayName: 'Test'
 
   - pwsh: |
       Set-Variable -Name ZIGINSTALLDIR -Value "$(Get-Location)\build\dist"
 
+      function CheckLastExitCode {
+        if (!$?) {
+          exit 1
+        }
+        return 0
+      }
+
       & "$ZIGINSTALLDIR\bin\zig.exe" build docs
+      CheckLastExitCode
     timeoutInMinutes: 60
     name: doc
     displayName: 'Documentation'
--- a/ci/azure/pipelines.yml
+++ b/ci/azure/pipelines.yml
@@ -27,7 +27,7 @@ jobs:
     vmImage: 'windows-2019'
   variables:
     TARGET: 'x86_64-windows-gnu'
-    ZIG_LLVM_CLANG_LLD_NAME: 'zig+llvm+lld+clang-${{ variables.TARGET }}-0.10.0-dev.2931+bdf3fa12f'
+    ZIG_LLVM_CLANG_LLD_NAME: 'zig+llvm+lld+clang-${{ variables.TARGET }}-0.10.0-dev.3524+74673b7f6'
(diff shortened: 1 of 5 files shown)


andrewrk:
Is there a way to set the error action preference instead of this? It's pretty easy to make a mistake having to call this function after every line.

hdorio:
> Is there a way to set the error action preference instead of this?

`ErrorActionPreference` is already set to `Stop` on AzurePipeline but works only on cmdlet.

> It's pretty easy to make a mistake having to call this function after every line.

I completely agree, older code used to do `|| exit /b` on every line, I'm not sure if there is a better way. I have to admit I never work with Windows.

hdorio:
@andrewrk [my pipeline](https://dev.azure.com/hdorio/ziglang/_build/results?buildId=146&view=logs&j=3e8797c7-5b0a-5f2c-07a4-1bc5e60a122e&t=45960c34-79ee-509a-888b-9d7338b9a12b) fails with
```
error: Invalid option: -Dskip-stage2-tests
error: Invalid option: -Dskip-non-native
```

Do you know if these options were removed recently?

andrewrk:
if you get that error, it seems like these lines did not execute:

https://github.com/ziglang/zig/blob/d769fd0102dee7ea24f957c3c96e2336d1c18839/ci/azure/pipelines.yml#L75-L78

hdorio:
> if you get that error, it seems like these lines did not execute:

Thanks, tests now fail rightfully.
https://dev.azure.com/ziglang/zig/_build/results?buildId=24325&view=logs&j=3e8797c7-5b0a-5f2c-07a4-1bc5e60a122e&t=45960c34-79ee-509a-888b-9d7338b9a12b&l=1460

nektro:
Powershell is the same except for internal PS commands https://stackoverflow.com/a/9949909

andrewrk:
Why does it only check the exit code after some commands?

hdorio:
> Why does it only check the exit code after some commands?

I only want to check important commands, do you want me to add checks for `git`, `7z` and `aws`?
Also sometimes it feels redundant, if `git` fails to clone, for example, commands after that will also fail.

If you talk about commands like `cd`, `mv`, and so on, there are cmdlet aliases and should be caught by the `ErrorActionPreference = stop` setting. You can see the list with `Get-Alias`.

> Remind me, what was the purpose of the switch from batch to powershell?

To use `ErrorActionPreference = stop` and avoid having to check (`|| exit /b`) the last status for every command (works only on cmdlets unfortunately but it's a start).

Since PowerShell is a superset of Batch I'm not sure why one would want to use the old Batch over PowerShell these days but I can certainly find a list of other advantages if this matter needs to be resolved in a more convincing manner.

andrewrk:
Yes I would like failure of any of the commands to immediately fail. I think that only having the check after *some* commands is worse, because when trying to maintain this script it is unclear how to follow the pattern, making subjective judgement calls about what is "important". It's much easier to notice that every command, no exceptions, has `|| exit /b` and maintain the pattern. `ErrorActionPreference = stop` only applying to some commands is worse than before.

The reason to avoid PowerShell over batch is 1. it's worse than batch for the reason described above, and 2. it's an unnecessary dependency that limits our ability to migrate to a different CI server, perhaps one that does not offer powershell. A list of advantages would be irrelevant since we are not taking those advantages currently.

hdorio:
> Yes I would like failure of any of the commands to immediately fail.

Ok, I'll add the check for `git`, `7z`, and `aws`.

> I think that only having the check after _some_ commands is worse, because when trying to maintain this script it is unclear how to follow the pattern, making subjective judgement calls about what is "important". It's much easier to notice that every command, no exceptions, has `|| exit /b` and maintain the pattern.

I agree.

I can add the `CheckLastExitCode` after every commands and cmdlets if you want, it won't hurt and it's very much like adding `|| exit /b`.

> The reason to avoid PowerShell over batch is 1. it's worse than batch for the reason described above, and 2. it's an unnecessary dependency that limits our ability to migrate to a different CI server, perhaps one that does not offer powershell.

If you look at what we were using [before](https://github.com/ziglang/zig/blob/dd49ed1c642d917af40bf4a0e03d013f40b3903b/ci/azure/windows_msvc_script.bat#L20), `git` commands were not checked for errors.
Humans make mistakes even though (1.) was used. `.bat` and `pwsh` are the same to me in this regard.

For (2.):
Do we plan to migrate to a different CI server anytime soon?
Is a CI that does not provide `pwsh` nor a way to install it, is a worthy alternative?
Can someone point out to me one Windows CI that does not offer PowerShell?
Isn't this more likely that Microsoft (or future CIs) will [drop `bat` support](https://docs.gitlab.com/runner/shells/#:~:text=Deprecated%20in%20favor%20of%20PowerShell%20Core)?

We also used to use a combination of `powershell`, of `bat` scripts calling `bash` scripts and what not. I'm not saying we should continue to do so but maybe `pwsh` is not that big of a dependency to rely on (but it's your call).

> A list of advantages would be irrelevant since we are not taking those advantages currently.

I don't recall exactly because the changes were done in February but
I noted it was easier to deal with error outputs.
`bat` scripts would consider any outputs as an error which was a difficult thing for me to workaround.
The final logs were also messed-up (but I can't recall why).

Maybe `.bat` scripts are awesome, maybe roadblocks I encountered are easily fixable but, why not fix the issue at stake and rewrite the whole thing in another pull request?

andrewrk:
> Do we plan to migrate to a different CI server anytime soon?

No. Fair enough.

> Isn't this more likely that Microsoft (or future CIs) will [drop `bat` support](https://docs.gitlab.com/runner/shells/#:~:text=Deprecated%20in%20favor%20of%20PowerShell%20Core)?

Good point.

> why not fix the issue at stake and rewrite the whole thing in another pull request?

I agree with your conclusion. Will merge it after I identify and disable the failing tests (you can see I already pushed one such commit).

hdorio:
As a note: if we are still using `pwsh` in the future, the version 7.3 is adding [PSNativeCommandErrorActionPreference](https://docs.microsoft.com/en-us/powershell/scripting/learn/experimental-features?view=powershell-7.2#psnativecommanderroractionpreference) which will remove the need to check the last exit status of executables if combined with `ErrorActionPreference = 'Stop'`. (`$PSNativeCommandUseErrorActionPreference = $true`)

> This feature implements a preference variable to enable errors produced by native commands to be PowerShell errors. This allows native command failures to produce error objects that are added to the PowerShell error stream that may terminate execution of the script without extra handling.

PR #12388, opened 2022-08-09
Zig Usage Command

lyledean1:
Currently running 
```
zig
```
Returns usage + an error. 

Proposal here would be to just return usage, as a user the first command I will run is the `zig` command to get information about the possible commands preferably without an error.


andrewrk:
This is working as designed. Running `zig` by itself is not an expected way to use the tool and is therefore an error. This matches precedent by C compilers as well as my own aesthetic opinion.

lyledean1:
Understood, thanks for the context!

Issue #12366, opened 2022-08-08
missing compile error for comptime `@shlExact` overflow


andrewrk:
```zig
comptime {
    const x = @shlExact(@as(u8, 0b01010101), 2);
    _ = x;
}
```

Expected: error: operation caused overflow
Actual: no compile error

Same issue with `@shrExact`:

```zig
comptime {
    const x = @shrExact(@as(u8, 0b10101010), 2);
    _ = x;
}
```

Same issue with `@divExact`:

```zig
comptime {
    const a: u32 = 10;
    const b: u32 = 3;
    const c = @divExact(a, b);
    _ = c;
}
```

PR #12389, opened 2022-08-09
stage2: add compile error for shlExact overflow

leroycep:
I couldn't figure out if I was running the test case, I don't think I was.

Completes part of #12366, doesn't add a similar check for `shrExact` or `divExact`.


Vexu:
There is already a test for this in `test/cases/compile_errors/stage1/obj`. Could you move that instead of adding a new one?

Vexu (approved):

Issue #12364, opened 2022-08-08
Compile time coercion of float to int test case incorrectly passes


andrewrk:
Zig Version: `0.10.0-dev.3477+e28567538`

### Steps to reproduce:

```zig
// Compile time coercion of float to int
test "implicit cast to comptime_int" {
    var f: f32 = 54.0 / 5;
    _ = f;
}
```

### Expected behavior:

stage1 has the correct behavior:

```
./test3.zig:3:18: error: float value 54.000000 cannot be coerced to type 'comptime_int'
    var f: f32 = 54.0 / 5;
                 ^
```

Bonus points: a note that explains [why this is ambiguous](https://ziglang.org/documentation/master/#Type-Coercion-Coercion-Float-to-Int).

### Actual behavior:

```
All 1 tests passed.
```

PR #12397, opened 2022-08-10
Sema: error on ambiguous coercion of comptime float and ints

martinhath:
The following, from the documentation as of the time of writing, illustrates
the problem:

```zig
// Compile time coercion of float to int
test "implicit cast to comptime_int" {
    var f: f32 = 54.0 / 5;
    _ = f;
}
```

It is not clear how to unify the types of 54.0 and 5 to perform the
division. We can either

 - cast 54.0 to comptime_int resulting in @as(comptime_int, 10), which is
   casted to @as(f32, 10), or
 - cast 5 to comptime_float resulting in @as(comptime_float, 10.8), which
   is casted to @as(f32, 10.8)

Since the two resulting values are different, a compiler error is appropriate.

Fixes: #12364

---

I wasn't sure about exactly which types should be disallowed, so I only used the comptimes. 

Further, this also raises an error in the flipped case, which stage1 doesn't:

```zig
// Compile time coercion of float to int
test "implicit cast to comptime_int" {
    var g: f32 = 10 / 3.0;
    _ = g;
}
```
Running:
```bash
$ zig version
0.9.1
$ zig test 12364.zig
All 1 tests passed.
$ zig-out/bin/zig test 12364.zig
12364.zig:5:21: error: ambiguous coercion of the division operands: 'ComptimeInt' and 'ComptimeFloat'
    var g: f32 = 10 / 3.0;
                 ~~~^~~~~
```

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -11211,7 +11211,7 @@ fn zirDiv(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.Ins
         const rhs_val = maybe_rhs_val orelse unreachable;
         const rem = lhs_val.floatRem(rhs_val, resolved_type, sema.arena, target) catch unreachable;
         if (rem.compareWithZero(.neq)) {
-            return sema.fail(block, src, "ambiguous coercion of division operands '{s}' and '{s}'; division has non-zero reminder '{}'", .{
+            return sema.fail(block, src, "ambiguous coercion of division operands '{s}' and '{s}'; non-zero remainder '{}'", .{
                 @tagName(lhs_ty.tag()), @tagName(rhs_ty.tag()), rem.fmtValue(resolved_type, sema.mod),
             });
         }
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -11202,18 +11202,17 @@ fn zirDiv(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.Ins
     const maybe_lhs_val = try sema.resolveMaybeUndefValIntable(block, lhs_src, casted_lhs);
     const maybe_rhs_val = try sema.resolveMaybeUndefValIntable(block, rhs_src, casted_rhs);
 
-    if ((lhs_ty.tag() == .comptime_float and rhs_ty.tag() == .comptime_int) or
-        (lhs_ty.tag() == .comptime_int and rhs_ty.tag() == .comptime_float))
+    if ((lhs_ty.zigTypeTag() == .ComptimeFloat and rhs_ty.zigTypeTag() == .ComptimeInt) or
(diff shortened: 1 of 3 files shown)


martinhath:
Oh no, I tried to `--amend` and force push a slight revision of the wording of the error message, but the PR is still using the old commit.

Vexu:
`55.0 / 5` is never going to be ambiguous, I think it would be better to first cast to `comptime_float` and check if the division produces a non-zero remainder and only issue the error then, along with a `division has non-zero remainder {d}` note.

Also the error message should use `comptime_int`/`comptime_float` rather than `ComptimeInt`/`ComptimeFloat`.

martinhath:
> `55.0 / 5` is never going to be ambiguous, I think it would be better to first cast to `comptime_float` and check if the division produces a non-zero remainder and only issue the error then, along with a `division has non-zero remainder {d}` note.

This feels a little iffy to me; is there a precedence in the language where one would get type errors from a change of values? That is, similar cases where something like  `55.0 / 5` would compile but something like `54.0 / 5` would not?

> Also the error message should use `comptime_int`/`comptime_float` rather than `ComptimeInt`/`ComptimeFloat`.

Makes sense; I'll update this.

Vexu:
> This feels a little iffy to me; is there a precedence in the language where one would get type errors from a change of values? That is, similar cases where something like `55.0 / 5` would compile but something like `54.0 / 5` would not?

IMO this is the same as `@as(u6, 5) << 5` compiling but `@as(u6, 5) << 6` resulting in `error: shift amount '6' is too large for operand type 'u6'`.

martinhath:
I thought about it some more, and I agree with you now. There's also a TODO in `zirDiv` already sketching out that this is the intended behavior:
```
    // TODO: emit compile error when .div is used on integers and there would be an
    // ambiguous result between div_floor and div_trunc.
```

martinhath:
I thought about it some more, and I agree with you now. There's also a TODO in `zirDiv` already sketching out that for #12339  this is the intended behavior, so I assume that we should accept code in which the behavior is well defined.
```
    // TODO: emit compile error when .div is used on integers and there would be an
    // ambiguous result between div_floor and div_trunc.
```

martinhath:
I've updated the code to only error when the remainder is non-zero. I couldn't figure out how to easily fish out the numbers, so I might have done things in an awkward way. In any case, this is how it looks now:

```zig
test "implicit cast to comptime_int" {
    var f: f32 = 10 / 2.0;
    _ = f;
}
```
gives

```
All 1 tests passed.
```

and 

```zig
test "implicit cast to comptime_int" {
    var f: f32 = 10 / 3.0;
    _ = f;
}
```
gives

```
12364.zig:3:21: error: ambiguous coercion of division operands: 'comptime_int' and 'comptime_float': division has non-zero reminder: 1
    var f: f32 = 10 / 3.0;
```

I'm not crazy about the error message, but I'm not sure how to improve it while keeping it short.

Vexu (approved):

Issue #7600, opened 2020-12-30
Zig runtime should set console output mode to utf-8 by default on Windows


chr-1x:
```zig
const std = @import("std");

pub fn main() void {
    std.log.debug("Hello, μ!  (^=◕ᴥ◕=^)", .{});
    var f = std.io.getStdOut();
    _ = f.write("Hello, μ!  (^=◕ᴥ◕=^)\n") catch @panic("write to stdout failed");
    f = std.io.getStdErr();
    _ = f.write("Hello, μ!  (^=◕ᴥ◕=^)\n") catch @panic("write to stderr failed");
}
```

```
> repro_windows_utf8.exe
debug: Hello, ╬╝!  (^=Γùòß┤ÑΓùò=^)
Hello, ╬╝!  (^=Γùòß┤ÑΓùò=^)
Hello, ╬╝!  (^=Γùòß┤ÑΓùò=^)
```

Poor kitty, its face has been replaced by encoding errors!

The zig documentation has this to say about string literals:

> String literals are single-item constant Pointers to null-terminated UTF-8 encoded byte arrays.
> ... String literals are const pointers to null-terminated arrays of u8, and by convention parameters that are "strings" are expected to be UTF-8 encoded slices of u8.

If this is true of all standard library functions, `std.log` and writers that refer to console output streams are expecting utf-8 encoded strings, and callers likely expect these functions to output these strings to the console as they appear in their source files or data sources. Users of zig command-line applications (and zig developers) may not have [configured Windows to use utf-8 by default](https://docs.microsoft.com/en-us/windows/console/classic-vs-vt#unicode), but this can be overridden for a given console session by [`SetConsoleOutputCP`](https://docs.microsoft.com/en-us/windows/console/setconsoleoutputcp).

I propose that the zig runtime should, at program startup, call `SetConsoleOutputCP` with an argument of [65001](https://docs.microsoft.com/en-us/windows/win32/intl/code-page-identifiers) to amend terminal output to correctly show utf-8 strings in all cases. If zig developers want legacy behavior or alternative code pages, they can call this function again with a different argument in their program's entry point.

chr-1x:
Note that if the zig runtime does *not* do this, every zig developer who wants to ensure that they correctly output UTF-8 encoded text on all Windows machines will have to insert os-specific logic in their startup code, which makes writing portable zig code a bit more of a headache and a bit less of a happens-by-default thing.

daurnimator:
[...]

Issue #5148, opened 2020-04-23
Cannot read unicode data from STDIN on Windows


Tetralux:
Tracking issue for how, on Windows, just using `ReadFile` will completely corrupt UTF-16 data.

Instead, you have to use `ReadConsoleW`.

For example:
```zig
extern "kernel32" fn ReadConsoleW(handle: os.fd_t, buffer: [*]u16, len: os.windows.DWORD, read: *os.windows.DWORD, input_ctrl: ?*c_void) i32;
extern "kernel32" fn SetConsoleOutputCP(cp: os.windows.UINT) i32;

// ...

assert(SetConsoleOutputCP(65001) != 0); // display outputted utf8 correctly in the console
    
var stdin = io.getStdIn().handle;
var data: [256]u16 = undefined;
var read: u32 = undefined;
const ok = ReadConsoleW(stdin, &data, data.len, &read, null); // type in '안녕'
assert(ok != 0);

var utf8: [1024]u8 = undefined;
const utf8_len = try unicode.utf16leToUtf8(&utf8, data[0..read]);
const s = utf8[0..utf8_len];

print("'{}'\n", .{s});  // prints out '안녕
                        //            '
```
As noted by @fengb, this may mean Zig needs to have a special `InStream` for the console.

doongjohn:
maybe we can solve this without windows api?
https://forum.nim-lang.org/t/5874#38913

Issue #3823, opened 2019-12-01
Return streams from `getStdOut`, `getStdIn`, and `getStdErr`


jayschwa:
In the Zig codebase (and in my own, what little that is), almost every call to these functions immediately performs `[in|out]Stream().stream`. If `getStdOut` and friends just returned the streams directly, that would probably DRY a lot of code. Perhaps `getStdOutFile` could be added for less common use-cases.

Exceptions I found in Zig codebase:
- `stdout_file.write()` in a couple places, but `OutStream` has `write` too.
- `cat_file` in a test takes `fs.File`s, but it looks like it's implementation is only concerned with a `read` and `write` method on its arguments, and would probably not be hard to change.

andrewrk:
If we provided global storage for stdio streams, it should probably be a buffered one, with a mutex protecting it. This would match stdin, stdout, stderr from libc. My only concern with this is that an application generally should be in control of its own standard I/O streams, and I don't want this to be an invitation to libraries to start dumping garbage into stdio. That's what #2586 is for.

And then the buffering and the mutex might not actually be necessary, depending on how the application is structured. The boilerplate is annoying, but it makes it clear that it's up to the application how it wants to deal with stdio.

Issue #224, opened 2017-01-18
add an error for using `var` when one could have used `const`


andrewrk:
```zig
fn add(a: i32, b: i32) {
    var x = a + b;
    return x;
}
```

Should give error: x should be declared const

daurnimator:
I think this should be a warning; not an error.

When debugging/prototyping, it's useful to comment out a line without having to refactor, e.g.

```zig
fn foo(a: i32, b: i32) {
    var x = a + b;
    // x = x + 4
    return x;
}
```

tgschultz:
Zig doesn't have warnings.

I feel like, if we're not going to trust the programmer on this, and the compiler can tell if a variable should be const or not, then why even make the programmer specify that information?

vi:
If such error is implemented, it should be ignorable at function / scope / var level using some attribute.

Unused non-constness can arise from generic / code generation context.

user00e00:
Maybe there should be warnings in zig, but instead of calling them "warnings", they should be called "debug-suppressed errors".

* When releasing a project, you want to optimize for readability and maximize the amount of assumptions the reader can make while reading code.
* When debugging, you want as little friction as possible while modifying code.

I think both of these are valid requests, but you cannot really have both (at the same time).

Some compile time checks should be skipped in debug mode if they add too much friction, IMO.

vi:
Suppressing this error should not be debug-only.

For example, in following code and in similar code, but generated outside Zig (code generation), `var q` should not require changing to `const q` if both blocks are happen to be excluded.

```zig
pub fn chunk_by_chunk(comptime a : bool, comptime b: bool) u32 {
    var q : u32 = 0;
    if (a) {
        q += 4;
    }
    if (b) {
        q += 1;
    }
    return q;
}

pub fn main() void {
    var q : u32 = chunk_by_chunk(false, false);
    @import("std").debug.warn("{}\n", q);
}
```

hryx:
I love this because it further guides authors to optimal code.

Questions:

1. Will this affect non-pub globals too? Or just statements?
2. One use case for intentionally using `var` over `const` is forcing non-comptime values in tests. Will underscore come to the rescue here?

```zig
test "var" {
    var x: u8 = 1;
    _ = x;
    switch (x) {
        // test some behavior here
    }
}
```

Sobeston:
[...]

PR #12400, opened 2022-08-11
std: Windows Console IO Fix

Vulfox:
I tried tackling what I thought would be a somewhat small fix for these issues:

Resolves https://github.com/ziglang/zig/issues/5148
Resolves https://github.com/ziglang/zig/issues/7600

There were a couple of places I  could make a solution work, but figured it would be best to introduce a new struct representing the std-out/err/in handles for Windows.

At first I was looking to change up `write` and `read` for the `File` struct with an if conditional on an `isTty()` check. It didn't seem right to add extra checks in these functions unnecessarily when I know that `io.getStdXYZ()` provides everything needed to know that the `File` or whatever returned type is meant for a `Console` on Windows. Doing so required some minimal changes to functions relying on `getStdXYZ()` to `getStdXYZ().writer()` or the equivalent for reader. On top of that, adding Windows specific fields to `File` also seemed like a bad idea for reading needs.

As far as not using `GetConsoleMode` or `SetConsoleOutputCP` for this, they seemed to have a risk of changing an active session's code page without any guarantees to the user that the program will set it back after it is finished or panics. I took a look at how Rust (may have borrowed some ideas from there) and Go implemented Windows console writing and they also didn't use these functions.

This is still a somewhat rough cut for a PR. I didn't want to make too many changes/additions without some input from maintainers. I can split this up into 2 different PRs for the Reader and Writer if need be, but I wanted to provide a fuller picture of what the changes look like for input and output.

Here's where I need some assistance/direction:
- I don't quite know how I can test the new ConsoleReader+ConsoleWriter in the std lib. If `File.(Read|Write)File` has some Windows specific tests I can mimic and try to implement for reading from a console, I will happily add some there, but it doesn't seem like there are any specifically for those and Windows. I could be blind 🤷‍♂️ 
- Is changing the `getStdXYZ` functions the best place for introducing this logic of translating utf8<->utf16 for Windows Console?
- Is `std/io` the preferred directory for the Windows specific reader and writer to live?

I was able to eye ball test these changes locally on Windows 10 and can confirm what used to return junk characters, now properly renders the correct unicode characters on the terminal.


lib/std/io.zig
```diff
@@ -46,14 +46,20 @@ fn getStdOutHandle() os.fd_t {
     return os.STDOUT_FILENO;
 }
 
+pub const StdOutType = if (builtin.os.tag == .windows) ConsoleWriter(4096) else File;
```
r00ster91, line 49: I really don't like introducing these types for all platforms just because of Windows. I think an internal solution that works without introducing any new public APIs would be much better even if it may mean that on Windows some structs need to have more fields or checks.

lib/std/Progress.zig
```diff
@@ -143,7 +143,7 @@ pub const Node = struct {
 /// API to return Progress rather than accept it as a parameter.
 /// `estimated_total_items` value of 0 means unknown.
 pub fn start(self: *Progress, name: []const u8, estimated_total_items: usize) *Node {
-    const stderr = std.io.getStdErr();
+    var stderr: std.io.StdErrType = std.io.getStdErr();
```
r00ster91, line 146: Similar thing here, if these have to be `var` now instead of `const` then I imagine it would have to be done in other places too, all just because of Windows. I hope we can find a solution where there are no external changes needed.

r00ster91 (commented):
Just my 2 cents.

lib/std/Progress.zig
Vulfox: Maybe this is just me getting mixed up between `const` and `var` keywords in Zig world, but wouldn't it be bad form/illegal to change the state of a field within a `const` like with the rotating fifo_utf8 buffer needed for reads? If it is, wouldn't that mean these have to be a `var` anyways if I introduce a change to `File` to read Utf16Le on Windows consoles?

I can try experimenting with changing `File` read/write to dance around Windows console. It would get rid of the need to update a lot of the compiler code to make it work within this repo.

Now this is if Zig does the translating of Utf16Le -> Utf8. If we have Windows do it under the hood, it would get rid of the need to change `File` or how we get Std handles. I'm just currently avoiding it to prevent any potential harm of changing someone's code page without them knowing.

r00ster91: Hmm I just really don't think that whatever `std.io.getStdErr` returns should have state like that such that you need to make it `var` in order to write or read from it. `std.fs.File` is a representation of the file, not the file itself.
So I think this change is just kind of breaking that way of thinking about it. I think it should keep returning always only one struct (`std.fs.File`).

But even so, anything that can be `const` should be `const`. In the future using `var` when you could use `const` might become an error: https://github.com/ziglang/zig/issues/224 which is supposed to remind the developer that maybe they wanted to mutate that variable.

> I can try experimenting with changing `File` read/write to dance around Windows console. It would get rid of the need to update a lot of the compiler code to make it work within this repo.

Yes, I think that's the right approach.

> If we have Windows do it under the hood, it would get rid of the need to change `File` or how we get Std handles.

That's possible with the `GetConsoleMode` and `SetConsoleOutputCP` functions you mentioned right? Hmm well for buffered writing for example you could set the output code page to UTF-8 and then after the flush reset it.
Well, I think there will be many people that will always want to use UTF-8 throughout their whole application, so maybe we just set the output code page to UTF-8 at the start of the program (or at the first write of anything) using `std.once` or something and then if the user wants any other code page they have to set it themselves? We could expose `SetConsoleOutputCP` with a nicer API or something.

However if the user sets the code page and only after that does their first write, the code page will be set to UTF-8 again. Maybe it is better to do it on program startup. But what if they never write anything in their whole program?
I don't know if you have any ideas here?

So maybe it is better to follow suit and do it without using those functions, like Rust and Go.

Vulfox: Been messing around with this off and on over the weekend. I can definitely get `write` to be solely in `File`. `read` on the other hand is being a bit tricky here, as I need mutability somewhere within itself or how it returns a reader. Unfortunately, most everything that uses `File` will use it with `const` for both it and the reader.

I've tried being a bit sneaky with how to return the reader:
```zig
    pub const WinConsoleReader = io.ConsoleReader(4096);
    pub const Reader = if (is_windows) WinConsoleReader.Reader else io.Reader(File, ReadError, read);

    pub fn reader(file: File) Reader {
        if (is_windows) {
            return (WinConsoleReader{ .file = file }).reader();
```

However, this is probably a footgun of some sorts where the scope of the WinConsoleReader is ended, so the reader ends up pointing to junk. Got any ideas as to how I might be able to utilize a small mutable buffer for reads without changing the current APIs?

r00ster91: Hmm if I look at Rust, it does seem to work like `let mut file = File::create("foo.txt")?;` there (so `var` instead of `let mut` in Zig) so maybe we will have to change all `const`s to `var`s or something? In that case it might make sense to discuss this further in a separate issue.

I'm still wondering if this is a feasible approach?

> Perhaps we can disallow (make it "unsafe") to set the code page with `SetConsoleOutputCP` and they have to use our function so that we can keep track of whether the user already set an explicit code page or not. If not, we can set it to UTF-8 on the first write of anything. Does this sound possible?

It does sound pretty simple to me. But maybe it's too hacky to do.

I suppose `GetConsoleOutputCP` or something would not be of much use because I assume these functions return UTF-16 as the default before any code page has been set.

Vulfox: > And ultimately, using var instead of const would only be required on Windows anyway, I think. I'm just thinking it might be a bit inconsistent when you're reading code but that's not very important.

If that feature comes into play where it complains about users using `var` instead of `const`, it will cause compilation errors for non-Windows targets, if they are developing from a Windows machine.

> It does sound pretty simple to me. Also sounds kinda hacky though so it's probably not the best solution.
>
> I suppose GetConsoleOutputCP or something would not be of much use because I assume these functions return UTF-16 as the default before any code page has been set (so that we can check whether an explicit code page has been set).

While I think we could find a way to use those API sets, I fear it may cause the active session's code page to change and there be no way back to the original code page. Sure we could try and gather what the current code page is and set it back at the end, but if the program is killed before it can handle the revert on code page, it goofs up the session.

> It might make sense to discuss this further in a separate issue.
I would be fine with that.

Vulfox: As far as proposals go, this might be related, https://github.com/ziglang/zig/issues/3823 . If there were a buffer of some sort living in between for these things, then UTF16 decoding could live with using those. Also instead of returning a `File` it is proposing to return `Reader`s instead, which could work in the favor of having Windows return specific Console Reader/Writers for its stdio.

topolarity: Would it be possible to side-step the reader buffering problem by under-filling the read buffer, if the remaining space might not be large enough for a complete codepoint?

If the read ended in a leading surrogate, just make another `ReadFileW` or `ReadConsoleW` call to get the remaining UTF-16 word and complete the codepoint.

topolarity: Thanks for the detailed thoughts. Makes good sense. We definitely need a buffer to support `readByte`

One more naive question for you: Is there a reason we can't use a static (`threadlocal`?) variable for that buffer? It seems like for correct-ness, we'd actually prefer it to be global.

Vulfox: > One more naive question for you: Is there a reason we can't use a static (`threadlocal`?) variable for that buffer? It seems like for correct-ness, we'd actually prefer it to be global.

I think that starts to dip into what the above [proposal](https://github.com/ziglang/zig/issues/3823) tries to potential provide as an answer from Andrew's comment. I'd be willing to throw in a `threadlocal` buffer+surrogate set for Windows reads and see what the looks like. Might come out cleaner for everyone. Give me a bit to toy with that (I'm not familiar with that keyword/it's usage 😅 ).

topolarity: Sounds great! Feel free to reach out on #stdlib-devel on the Discord if you'd talk through any of the finer points of the design 🙂

(I'm no concurrency expert either, but there are some folks in that channel who are)

Vulfox: Gave it a shot, and at a glance, yea, it is operating as you thought 🦸 . It allows us to keep all changes within `File` without having to change any user facing APIs. `threadlocal` might be a decent call to have for any given stdin handle devs use, but I would think this might cause headaches if multiple threads are reading from the same handle on Windows, which I don't know of a good reason off the top of my head to do that anyways.

Let me update this PR to reflect the consolidation of all this + cleanup + comments.

Vulfox: Alright, take a gander at this new version of `File` only changes to `read`/`write` and let me know what y'all think!

r00ster91: Looks good to me overall.

lib/std/fs/file.zig
```diff
@@ -8,8 +8,13 @@ const assert = std.debug.assert;
 const windows = os.windows;
 const Os = std.builtin.Os;
 const maxInt = std.math.maxInt;
+const FifoType = std.fifo.LinearFifo(u8, std.fifo.LinearFifoBufferType{ .Static = 4 });
 const is_windows = builtin.os.tag == .windows;
 
+threadlocal var win_surrogate: ?u16 = null;
```
topolarity, line 14: Hmm after thinking a bit more about this, I'm not sure this should be `threadlocal` after all. What I'd propose would be:
  - Leave the buffer as a static global (non-threadlocal)
  - Any use of `stdin` across threads on Windows requires you to synchronize I/O access using your own mutex or similar
  
Do you see any issues with that approach? cc @kprotty

kprotty: Its an odd restriction for only a single platform. If it does exist, I think it should extend to all platforms and even all tty Files

Vulfox: @topolarity I've been mulling over the implications of `threadlocal` as well. Your suggestion sounds like exactly what the proposal I mentioned before definitely wanted to _maybe_ tackle.

@kprotty  I can try to see what I can cook up to settle this in a platform agnostic way, while trying to dodge the bullet of changing any current public APIs/types.

topolarity: My thought was to require the user to synchronize their use of stdout/stdin with their own mutex, not to wrap this in a mutex in the standard library

That would leave the default behavior fast for single-threaded usage, while still making it possible to perform multi-threaded access correctly (multi-threaded access is broken with threadlocal, since each thread will see a different buffer)

andrewrk:
@Vulfox  could you please rebase this on latest master? My auto-rebase script failed due to conflicts.

andrewrk:
Unfortunately, this needs another rebase - CI tests didn't pass last time, and now there are conflicts with master branch.

andrewrk:
Closing - no updates for 30 days and CI checks are failing. Please open a new PR if you want to continue these efforts.

Issue #12387, opened 2022-08-09
stage2 unable to build stage3 on x86_64 macos: panic: incorrect alignment


andrewrk:
Zig Version: `0.10.0-dev.3511+17511a48f`

```
andy@Andrews-Air-2:~/dev/zig/build$ stage2/bin/zig build -p stage3 -Denable-llvm -Denable-stage1
MachO Flush... thread 7320547 panic: incorrect alignment
/Users/andy/dev/zig/src/link/MachO/Object.zig:104:72: 0x1032c60bc in link.MachO.Object.parse (zig1)
                    @alignCast(@alignOf(macho.nlist_64), &self.contents[symtab.symoff]),
                                                                       ^
/Users/andy/dev/zig/src/link/MachO/Archive.zig:228:21: 0x1032b8f44 in link.MachO.Archive.parseObject (zig1)
    try object.parse(allocator, cpu_arch);
                    ^
/Users/andy/dev/zig/src/link/MachO.zig:2676:51: 0x103377880 in link.MachO.resolveSymbolsInArchives (zig1)
            const object = try archive.parseObject(gpa, cpu_arch, offsets.items[0]);
                                                  ^
/Users/andy/dev/zig/src/link/MachO.zig:1108:42: 0x1032440dd in link.MachO.linkOneShot (zig1)
        try self.resolveSymbolsInArchives();
                                         ^
/Users/andy/dev/zig/src/link/MachO.zig:414:45: 0x1032266d8 in link.MachO.flush (zig1)
        .one_shot => return self.linkOneShot(comp, prog_node),
                                            ^
/Users/andy/dev/zig/src/link.zig:641:72: 0x103223788 in link.File.flush (zig1)
            .macho => return @fieldParentPtr(MachO, "base", base).flush(comp, prog_node),
                                                                       ^
/Users/andy/dev/zig/src/Compilation.zig:2335:24: 0x1031bef48 in Compilation.flush (zig1)
    comp.bin_file.flush(comp, prog_node) catch |err| switch (err) {
                       ^
/Users/andy/dev/zig/src/Compilation.zig:2316:23: 0x1031b1e35 in Compilation.update (zig1)
        try comp.flush(main_progress_node);
                      ^
/Users/andy/dev/zig/src/main.zig:3327:20: 0x103149030 in main.updateModule (zig1)
    try comp.update();
                   ^
```

[...]

PR #12405, opened 2022-08-11
link.MachO: use accurate alignment attribute on pointers

andrewrk:
Proper fix for #12387

### DIFF
--- a/lib/std/array_list.zig
+++ b/lib/std/array_list.zig
@@ -221,6 +221,30 @@ pub fn ArrayListAligned(comptime T: type, comptime alignment: ?u29) type {
             mem.copy(T, self.items[old_len..], items);
         }
 
+        /// Append an unaligned slice of items to the list. Allocates more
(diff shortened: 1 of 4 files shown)


kubkon:
I wanted to avoid sprinkling `align(1)` on types but I guess there is no way around it. Thankfully it only ever applies to object files and dylibs for instance which have a hard requirement re proper alignment to be loadable. Anyhow, thanks for fixing my fix!

kubkon (approved):

andrewrk:
There is a way around it, which is to to copy the data from the unaligned file into an aligned array for long term storage. Unclear whether that is better than status quo. Something to investigate later :)

kubkon:
> There is a way around it, which is to to copy the data from the unaligned file into an aligned array for long term storage. Unclear whether that is better than status quo. Something to investigate later :)
> 
> Thanks for the review!

Ironically, that's what we used to do before my rewrite...

PR #12426, opened 2022-08-12
Fix #12423: auto_hash not hashing arrays of slices uniquely

### LINKED ISSUE #12423: std: auto_hash not hashing multi-dimensional arrays uniquely
### Zig Version

0.10.0-dev.3480+396eb730e

### Steps to Reproduce

Create the file `array_hash.zig` with the following content:
```zig
const std = @import("std");

pub fn main() !void {
    const array_a = [_][]const u32{ &.{ 1, 2, 3 }, &.{ 4, 5 } };
    const array_b = [_][]const u32{ &.{ 1, 2 }, &.{ 3, 4, 5 } };

    const hash_a = hash(&array_a);
    const hash_b = hash(&array_b);

    std.debug.print("a: {x} b: {x}\n", .{hash_a, hash_b});
}

fn hash(slice: []const []const u32) u64 {
    var hasher = std.hash.Wyhash.init(42);
    std.hash.autoHashStrat(&hasher, slice, .DeepRecursive);
    return hasher.final();
}
```

### Expected Behavior

The example code prints 2 different hashes.

### Actual Behavior

The example code prints the same hash twice.
The reason this happens is that `auto_hash` hashes arrays by iterating them, and hashing each element. Since the numbers iterated are the same for both arrays in my example, this results in the same output. To fix this, we could first hash all the elements, and then hash the length afterwards, making it impossible for this problem to occur.

yujiri8:
I applied the change suggested in #12423. I also noticed that the switch seemed redundant since it just passes through the strat in either case, so I collapsed it to the else branch.

### DIFF
--- a/lib/std/hash/auto_hash.zig
+++ b/lib/std/hash/auto_hash.zig
@@ -30,13 +30,15 @@ pub fn hashPointer(hasher: anytype, key: anytype, comptime strat: HashStrategy) 
             .DeepRecursive => hash(hasher, key.*, .DeepRecursive),
         },
 
-        .Slice => switch (strat) {
-            .Shallow => {
-                hashPointer(hasher, key.ptr, .Shallow);
-                hash(hasher, key.len, .Shallow);
-            },
-            .Deep => hashArray(hasher, key, .Shallow),
-            .DeepRecursive => hashArray(hasher, key, .DeepRecursive),
+        .Slice => {
+            switch (strat) {
+                .Shallow => {
+                    hashPointer(hasher, key.ptr, .Shallow);
+                },
+                .Deep => hashArray(hasher, key, .Shallow),
+                .DeepRecursive => hashArray(hasher, key, .DeepRecursive),
+            }
+            hash(hasher, key.len, .Shallow);
         },
 
         .Many,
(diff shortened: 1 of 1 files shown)


Vexu:
Test please!

yujiri8:
Ahhhhhhhhhhhhh

I had tested it on the code given in the issue, but when I went to add it to the stdlib I noticed that the array test was already passing without changes, which meant it wasn't running through hashArray. So I figured out what was going on and realized that hashArray was probably the wrong function to change. I had gone into hashArray because the issue was titled "multi-dimensional arrays", but that's not really true, those are *not* arrays of arrays, they're arrays of slices! Array's length is part of their type, so it doesn't make sense to include that. Where it really needed to be added was on slices. So I've updated the PR to do that instead, and include a new test based on the code from the issue.

topolarity:
Now that the array len is included when hashing slices, this behavior has changed:

```zig
fn testHash(key: anytype) u64 {
    var hasher = Wyhash.init(0);
    hash(&hasher, key, .Deep);
    return hasher.final();
}


test "testHash multi-dimensional array" {
    const a = [_]u32{ 1, 2, 3 };
    const b: []const u32 = &a;
    try testing.expect(testHash(a) == testHash(b)); // fails, as of this PR
}
```

This one is not related to this change, but it looks like arrays are implicitly flattened when hashing:
```zig
test "testHash multi-dimensional array" {
    const a = [_]u32{ 1, 2, 3, 4, 5, 6 };
    const b = [_][3]u32{ .{ 1, 2, 3 }, .{ 4, 5, 6 } };
    try testing.expect(testHash(a) != testHash(b)); // fails, these compare equal
}
```

Vexu:
The reason for array len not being included was that it doesn't effect the primary use of `auto_hash` as being the hash function for `AutoHashMap`s where the array lengths are guaranteed to be the same due to typing, but it would be reasonable to argue that they should also be included for more generic use cases.

PR #12427, opened 2022-08-12
tokenizer: cleanups

r00ster91:
Just a few minor things that have been bothering me a little the few times I wanted to make some changes in this file.

### DIFF
--- a/lib/std/zig/tokenizer.zig
+++ b/lib/std/zig/tokenizer.zig
@@ -1485,7 +1485,7 @@ test "line comment followed by top-level comptime" {
     });
 }
 
-test "tokenizer - unknown length pointer and then c pointer" {
+test "unknown length pointer and then c pointer" {
     try testTokenize(
         \\[*]u8
         \\[*c]u8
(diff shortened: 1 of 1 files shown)


lib/std/zig/tokenizer.zig
```diff
@@ -1485,7 +1484,7 @@ test "line comment followed by top-level comptime" {
     });
 }
 
-test "tokenizer - unknown length pointer and then c pointer" {
```
r00ster91, line 1488: I believe these tests are sufficiently scoped so we don't need these prefixes.

lib/std/zig/tokenizer.zig
```diff
@@ -1433,8 +1432,8 @@ pub const Tokenizer = struct {
 
     fn getInvalidCharacterLength(self: *Tokenizer) u3 {
         const c0 = self.buffer[self.index];
-        if (c0 < 0x80) {
-            if (c0 < 0x20 or c0 == 0x7f) {
+        if (std.ascii.isASCII(c0)) {
+            if (std.ascii.isCntrl(c0)) {
```
r00ster91, line 1436: After #11629 we might use the ASCII comptime table for this so this would become a bit faster with `isCntrl`.

r00ster91:
Rebased.

Issue #12428, opened 2022-08-12
A function taking a pointer to packed struct(N) with fields of the wrong size produces a "depends on itself" error in addition to the real error


tauoverpi:
### Zig Version

0.10.0-dev.3550+fa50e179f

### Steps to Reproduce

run `zig test bug.zig -fno-stage1`

```zig
const std = @import("std");

const C = packed struct(u24) {
    i: u29,
};

pub fn stuff(c: *C) void {
    _ = c;
}

test {
    var x: C = undefined;
    stuff(&x);
}
```

### Expected Behavior

The actual error
```
bug.zig:3:25: error: backing integer type 'u24' has bit size 24 but the struct fields have a total bit size of 29
const C = packed struct(u24) {
                        ^~~
```

### Actual Behavior

Complains about depending on itself in addition to the real error.
```
bug.zig:3:25: error: backing integer type 'u24' has bit size 24 but the struct fields have a total bit size of 29
const C = packed struct(u24) {
                        ^~~
bug.zig:3:18: error: struct 'bug.C' depends on itself
const C = packed struct(u24) {
          ~~~~~~~^~~~~~
```

PR #12437, opened 2022-08-12
Sema: Revert sema.err to null if the Decl already has an error

sin-ack:
Previously we would assign the error message to Sema and then never
clear it even when destroying the error message, which caused memory
corruption.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -1925,6 +1925,7 @@ fn failWithOwnedErrorMsg(sema: *Sema, err_msg: *Module.ErrorMsg) CompileError {
     const gop = mod.failed_decls.getOrPutAssumeCapacity(sema.owner_decl_index);
     if (gop.found_existing) {
         // If there are multiple errors for the same Decl, prefer the first one added.
+        sema.err = null;
         err_msg.destroy(mod.gpa);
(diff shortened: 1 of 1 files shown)


Vexu:
You wouldn't happen to have a test for this?

sin-ack:
Unfortunately no, because the issue happened while porting my project to stage2, and I could only reproduce during the port. But it seems like two errors in a row in the same Decl should cause this error. What's the best way to write a test for this?

Vexu:
> But it seems like two errors in a row in the same Decl should cause this error. What's the best way to write a test for this?

No idea, it usually shouldn't be possible for multiple errors to be emitted for the same decl which is why I was asking for a potential test case.

sin-ack:
Yeah sorry, the only time I hit this error was when building the whole project. I can try to re-introduce the bug later on, but this fix seems OK for the time being from what I understand.

ifreund:
Would this happen to solve #12428? If so that could be a test case.

sin-ack:
That seems to be unrelated, since that successfully prints the error. This one causes a use-after-free because `sema.err` points to the newly-destroyed `ErrorMsg`.

PR #12439, opened 2022-08-13
Partial workaround for #1923, isolating error return traces between individual tests

Techcable:
Essentially #1923 means "caught" errors still show up in error return traces.
The correct fix would require the compiler to fix this, but that could affect performance.

For now, simply workaround this issue by clearing the return traces
between tests.

This means that "caught" errors in one test will not show up in the
error traces of other tests.

Consider the following code:
```zig
const std = @import("std");
const testing = std.testing;
 
export fn add(a: i32, b: i32) i32 {
    return a + b;
}
 
test "expect error" {
    const func = struct {
        fn oom() !void {
            return error.OutOfMemory;
        }
    };
    try testing.expectError(error.OutOfMemory, func.oom());
}
 
test "basic add functionality" {
    try testing.expect(add(3, 7) == 11);
}
```

Here is the behavior before this change:
image

Here is the behavior afterwards:
image

**EDIT**: Also thanks for @squeek502 for identifying the correct workaround

### DIFF
--- a/lib/test_runner.zig
+++ b/lib/test_runner.zig
@@ -96,6 +96,13 @@ pub fn main() void {
                 test_node.end();
             },
         }
+        // Workaround issue #1923 by clearing error stack traces between tests
+        //
+        // This is not a true fix, but helps isolate errors to the tests where they
+        // originate instead of confusing them all together
+        if (@errorReturnTrace()) |trace| {
+            trace.index = 0;
+        }
     }
     root_node.end();
     if (ok_count == test_fn_list.len) {


squeek502:
Nice idea, this should definitely help a good amount when running tests. #1923 is my only real annoyance with Zig so anything to mitigate it before it gets fixed properly is worth it IMO.

lib/test_runner.zig
```diff
@@ -81,6 +81,13 @@ pub fn main() void {
             ok_count += 1;
             test_node.end();
             if (!have_tty) std.debug.print("OK\n", .{});
+            // Workaround issue #1923 by clearing error stack traces between tests
+            //
+            // This is not a true fix, but helps isolate errors to the tests where they
+            // originate instead of confusing them all together
+            if (@errorReturnTrace()) |trace| {
+                trace.index = 0;
+            }
```
squeek502, line 90: Could this possibly be moved outside of the `if`? Haven't fully thought it through, but it seems like resetting the index even after failed tests could make sense as well. I'll try to come up with a case where it matters and report back.

squeek502: I think it should be moved outside of the if (below the else condition). Here's a simple case that shows the problem:

```zig
test "test skipped" {
    return error.SkipZigTest;
}

test "trigger error trace" {
    return error.WeExpectToSeeThisInTheTrace;
}
```

With the index resetting inside the `if` block:

```
Test [1/2] test "test expected error"... SKIP
Test [2/2] test "trigger error trace"... FAIL (WeExpectToSeeThisInTheTrace)
/home/ryan/Programming/zig/tmp/1923.zig:8:5: 0x208152 in test "test expected error" (test)
    return error.SkipZigTest;
    ^
/home/ryan/Programming/zig/tmp/1923.zig:12:5: 0x208172 in test "trigger error trace" (test)
    return error.WeExpectToSeeThisInTheTrace;
    ^
0 passed; 1 skipped; 1 failed.
```

and with it after the `else` block:

```
Test [1/2] test "test skipped"... SKIP
Test [2/2] test "trigger error trace"... FAIL (WeExpectToSeeThisInTheTrace)
/home/ryan/Programming/zig/tmp/1923.zig:6:5: 0x208172 in test "trigger error trace" (test)
    return error.WeExpectToSeeThisInTheTrace;
    ^
0 passed; 1 skipped; 1 failed.
```

PR #11736, opened 2022-05-27
std.child_process: add tests to test/standalone

matu3ba:
fixup of 67d5bfefba48d28c02e2841f1a47a213d28d4693

Issue #1620, opened 2018-10-02
std.os.ChildProcess should use vfork instead of fork on linux


andrewrk:
vfork pauses the parent process until the child process either calls execve or exit. This prevents Copy On Write pages, which makes spawning child processes avoid double the parent resident memory requirements when overcommit is off.

andrewrk:
Interesting notes from the man page:

```
NOTES
       Some  consider the semantics of vfork() to be an architectural blemish,
       and the 4.2BSD man page stated: "This system call  will  be  eliminated
       when  proper  system  sharing mechanisms are implemented.  Users should
       not depend on the memory sharing semantics of vfork() as  it  will,  in
       that case, be made synonymous to fork(2)."  However, even though modern
       memory management hardware has  decreased  the  performance  difference
       between  fork(2)  and  vfork(), there are various reasons why Linux and
       other systems have retained vfork():

       *  Some performance-critical applications require the small performance
          advantage conferred by vfork().

       *  vfork()  can be implemented on systems that lack a memory-management
          unit (MMU), but  fork(2)  can't  be  implemented  on  such  systems.
          (POSIX.1-2008 removed vfork() from the standard; the POSIX rationale
          for the posix_spawn(3) function notes that that function, which pro‐
          vides functionality equivalent to fork(2)+exec(3), is designed to be
          implementable on systems that lack an MMU.)

       *  On systems where memory is constrained, vfork() avoids the  need  to
          temporarily commit memory (see the description of /proc/sys/vm/over‐
          commit_memory in proc(5)) in order to execute a new program.   (This
          can  be especially beneficial where a large parent process wishes to
          execute a small helper program in a child  process.)   By  contrast,
          using  fork(2) in this scenario requires either committing an amount
          of memory equal to the size of the parent process (if  strict  over‐
          committing  is in force) or overcommitting memory with the risk that
          a process is terminated by the out-of-memory (OOM) killer.
```

bnoordhuis:
I don't think that can work. With vfork() you're only allowed to call execve() or _exit(). The setup that spawnPosix() does, like changing uid/gid and the cwd, is incompatible with vfork() - it would be visible to other threads in the parent process.

andrewrk:
My understanding is that changes to memory would affect the parent process, but the actual pid is a new child pid. So things like current directory, uid/gid, would be independent from the parent process. I'm doing some experiments now. Currently I'm getting a confusing segfault with the following code which I am troubleshooting:

```zig
[...]

PR #12442, opened 2022-08-13
child_process: split spawn into fork/exec

marler8997:
This is a proposed enhancement to the `ChildProcess` API.  In addition to `spawn`, this PR adds the `fork` function to `ChildProcess` which returns an instance of `Forked` which provides the `setup` and `exec` functions to finish spawning the child process.  This allows applications to utilize the functionality that `ChildProcess` provides without giving up the flexibility to perform custom setup code in the child process when needed.

Note that another variation I considered for this was to add one or more callbacks to `ChildProcess` that could be executed in the child process.  A callback design would be equivalent to what I've done here, except instead of adding a Context/Callback in the std library (the "Forked" context and setup/exec "callbacks"), the application would create their own Context/Callback(s).  By doing this in the std library instead, it only needs to be done once rather than once per use of the API.

### DIFF
--- a/lib/std/child_process.zig
+++ b/lib/std/child_process.zig
@@ -670,6 +670,7 @@ pub const ChildProcess = struct {
             forkChildErrReport(self.err_pipe_write, err);
         }
 
+        /// Performs setup in the child process, i.e. std pipes, cwd, gid, uid.
         pub fn setup(self: Forked) !void {
             try setUpChildIo(self.child.stdin_behavior, self.stdin_pipe[0], os.STDIN_FILENO, self.dev_null_fd);
             try setUpChildIo(self.child.stdout_behavior, self.stdout_pipe[1], os.STDOUT_FILENO, self.dev_null_fd);
@@ -711,12 +712,13 @@ pub const ChildProcess = struct {
         }
     };
 
-    /// fork off the child process.  Calling fork is equivalent to spawn except it will return
+    /// fork off the child process.  Returns `null` for the current process and `Forked` for
+    /// the child process.  Calling fork is equivalent to spawn except it will return
     /// in the child process before calling setup/exec.  This provides an opportunity for the
     /// caller to perform additional setup in the child process before calling setup/exec.
     ///
     /// This API does not support resource cleanup in the child process after fork returns.
-    /// The caller is only mean to perform setup and then call exec or exit afterwards.
+    /// The caller is only meant to perform setup and then call exec or exit afterwards.
     ///
     /// Note there are limitations on what can be done in the child process after fork returns.
     /// For example, calling malloc from libc may cause deadlock.
--- a/lib/std/child_process.zig
+++ b/lib/std/child_process.zig
@@ -656,7 +656,94 @@ pub const ChildProcess = struct {
         }
     }
 
+    pub const Forked = struct {
+        child: *ChildProcess,
+        stdin_pipe: [2]os.fd_t,
+        stdout_pipe: [2]os.fd_t,
+        stderr_pipe: [2]os.fd_t,
+        dev_null_fd: os.fd_t,
+        err_pipe_write: os.fd_t,
+        argv_buf: [:null]?[*:0]u8,
+        envp: [*:null]?[*:0]u8,
+
+        pub fn reportError(self: Forked, err: ChildProcess.SpawnError) noreturn {
+            forkChildErrReport(self.err_pipe_write, err);
+        }
+
+        pub fn setup(self: Forked) !void {
+            try setUpChildIo(self.child.stdin_behavior, self.stdin_pipe[0], os.STDIN_FILENO, self.dev_null_fd);
+            try setUpChildIo(self.child.stdout_behavior, self.stdout_pipe[1], os.STDOUT_FILENO, self.dev_null_fd);
+            try setUpChildIo(self.child.stderr_behavior, self.stderr_pipe[1], os.STDERR_FILENO, self.dev_null_fd);
+
+            if (self.child.stdin_behavior == .Pipe) {
+                os.close(self.stdin_pipe[0]);
+                os.close(self.stdin_pipe[1]);
+            }
+            if (self.child.stdout_behavior == .Pipe) {
+                os.close(self.stdout_pipe[0]);
+                os.close(self.stdout_pipe[1]);
+            }
+            if (self.child.stderr_behavior == .Pipe) {
+                os.close(self.stderr_pipe[0]);
+                os.close(self.stderr_pipe[1]);
+            }
+
+            if (self.child.cwd_dir) |cwd| {
+                try os.fchdir(cwd.fd);
+            } else if (self.child.cwd) |cwd| {
+                try os.chdir(cwd);
+            }
+
+            if (self.child.gid) |gid| {
+                try os.setregid(gid, gid);
+            }
+
+            if (self.child.uid) |uid| {
+                try os.setreuid(uid, uid);
+            }
+        }
+
+        pub fn exec(self: Forked) os.ExecveError {
+            return switch (self.child.expand_arg0) {
+                .expand => os.execvpeZ_expandArg0(.expand, self.argv_buf.ptr[0].?, self.argv_buf.ptr, self.envp),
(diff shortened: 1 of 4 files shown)


lib/std/child_process.zig
```diff
@@ -656,7 +656,94 @@ pub const ChildProcess = struct {
...
+        dev_null_fd: os.fd_t,
+        err_pipe_write: os.fd_t,
+        argv_buf: [:null]?[*:0]u8,
+        envp: [*:null]?[*:0]u8,
+
+        pub fn reportError(self: Forked, err: ChildProcess.SpawnError) noreturn {
+            forkChildErrReport(self.err_pipe_write, err);
+        }
+
+        pub fn setup(self: Forked) !void {
```
matu3ba, line 673: Super short doc comment could be added: Setup standard pipes, cwd, gid and uid

lib/std/child_process.zig
```diff
@@ -656,7 +656,94 @@ pub const ChildProcess = struct {
...
+
+    /// fork off the child process.  Calling fork is equivalent to spawn except it will return
+    /// in the child process before calling setup/exec.  This provides an opportunity for the
+    /// caller to perform additional setup in the child process before calling setup/exec.
+    ///
+    /// This API does not support resource cleanup in the child process after fork returns.
+    /// The caller is only mean to perform setup and then call exec or exit afterwards.
+    ///
+    /// Note there are limitations on what can be done in the child process after fork returns.
+    /// For example, calling malloc from libc may cause deadlock.
```
matu3ba, line 722: This is missing a description of the return values `null` and `Forked` from `forkPosix`: This returns \`null\` for the child and \`Forked\` for the parent`.

lib/std/child_process.zig
```diff
@@ -673,6 +760,7 @@ pub const ChildProcess = struct {
             destroyPipe(stderr_pipe);
         };
 
+        var in_child_process = false;
```
matu3ba, line 763: Why does `comptime var` not work here? stage1 bug?

test/standalone/child_process_setup/build.zig
```diff
@@ -0,0 +1,20 @@
+const std = @import("std");
+
+pub fn build(b: *std.build.Builder) void {
+    const target = b.standardTargetOptions(.{});
+    const mode = b.standardReleaseOptions();
+    const exe = b.addExecutable("child_process_setup", "main.zig");
+    exe.setTarget(target);
```
matu3ba, line 7: Do you expect miscompilations on Release? Otherwise, Debug mode should be sufficient/no setting of target and buildmode. See #11736 (has unfortunately weird CI errors).

test/standalone/child_process_setup/main.zig
```diff
@@ -0,0 +1,81 @@
...
+        // the ChildProcess API supports extra setup in the child process.
+        //
+        // In this case we close the write end of the pipe in the child process so that
+        // when the parent process closes it, the read end of the pipe will also close,
+        // otherwise, the pipe would stay open since the child process has a reference
+        // to the write end of the pipe.
+        os.close(pipe[1]);
+
+        forked.setup() catch |err| forked.reportError(err);
+        forked.reportError(forked.exec());
```
matu3ba, line 50: I dont understand why you are calling `forkChildErrReport` here. Can you explain?

matu3ba (commented):
Some missing doc comments and the usage of `reportError` is unclear to me.
Besides `in_child_process` should probably be comptime var to eliminate/elide the unnecessary check in each cleanup branch at comptime.

lib/std/child_process.zig
marler8997: Actually it's meant to be a runtime value.  It prevents a couple cleanup defer statements that are only supposed to run in the parent process and not the child process.

test/standalone/child_process_setup/main.zig
marler8997: This is duplicating the original behavior.  In the happy path `forked.exec` is a `noreturn`, otherwise, it returns an error that can be passed to `reportError` to pass the error number to the parent process.

test/standalone/child_process_setup/build.zig
```diff
@@ -0,0 +1,20 @@
...
+    exe.setBuildMode(mode);
+    const run = exe.run();
+    run.stdout_action = .{
+        .expect_exact = "child process waiting for pipe to close...\n" ++
+            "pipe closed\n",
+    };
+    run.step.dependOn(&exe.step);
+    const test_step = b.step("test", "Run the app");
+    // TODO: implement ChildProcess.fork on darwin
```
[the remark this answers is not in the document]
marler8997: Good question I thought about this and wasn't sure.  The test might be a bit long to include in `child_process.zig` but maybe a `child_process/test.zig` is a good place?

lib/std/child_process.zig
marler8997: Good catch, fixed

test/standalone/child_process_setup/build.zig
marler8997: Looks like there's no reason for the build mode, but I'm currently using the target to check if it's Darwin or not to disable the test in that case.

test/standalone/child_process_setup/build.zig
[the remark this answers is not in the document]
matu3ba: Better not in `lib/std/childprocess/test.zig`, as this slows done testing (test runner builds every time etc).
So this screams like a standalone test.

daurnimator:
- I don't think we should add such an API to the standard library that is incompatible with so many targets
- There's a huge footgun here in terms of what is allowed after a `fork` but before the `exec` in a threaded program *most* APIs cannot be used safely
- I don't think this belongs in the ChildProcess module.

ifreund:
I also don't think this abstraction belongs in the standard library. IMO if you're ok with targeting only unix-like systems and want more control than the cross-platform ChildProcess abstraction gives you should just use `os.fork()`/`os.execve()` directly. I think that the result will be simpler and easier to maintain/understand than the abstraction proposed here. This is what I currently do in river for example.

marler8997:
> I don't think we should add such an API to the standard library that is incompatible with so many targets

It should be supported on all targets that can spawn child processes except Windows.  (Note: Darwin is just not supported because it's not implemented for this initial proposal).

> IMO if you're ok with targeting only unix-like systems and want more control than the cross-platform ChildProcess abstraction gives you should just use os.fork()/os.execve() directly. I think that the result will be simpler and easier to maintain/understand than the abstraction proposed here. This is what I currently do in river for example.

My guess is river doesn't need or want the features provided by `ChildProcess` in that case I think I would also use `fork/exec` directly and I also choose to do that in many cases.  However there are some cases where I could use the the features provided by `ChildProcess` but still want to do some custom setup in the child process.  Here's a real example: https://github.com/marler8997/xremoteview/blob/master/both.zig 

At the moment you can either use `ChildProcess` and give up control or use `fork/exec` directly and give up the features that `ChildProcess` implements for you.  This change makes it possible to use `ChildProcess` in more cases.  I don't see much reason to limit the use cases ChildProcess can support.  I'm also not sure "abstraction" is the best way to describe this change, more of the opposite really.

ifreund:
@marler8997 Leaking these platform-specific implementation details of child process would also prevent us from making future improvements to bring some of the advantages of `posix_spawn` to this API. See for example #1620

matu3ba:
@marler8997 Does your use case (also) work with posix_spawn?

If yes: You can take a look or finish my attempts for bindings and port https://github.com/ziglang/zig/compare/master...matu3ba:zig:posix_spawn and https://github.com/matu3ba/zig/tree/posix_spawn3

Using the C routines by debugging the linking issues should be the fastest way.
The port requires a clarification, if a port of [pthreads](https://en.wikipedia.org/wiki/Pthreads) routines is wanted or not or how to deal with the signal safety side. pthreads is 7k LOC C code and posix_spawn 800 LOC C code.

Sidenode: posix_spawn and fork is missing a note, that it they are racy on non-Windows (during setup of the child process before execve) as file descriptors etc can leak to a parallel spawned thread/process.

PR #12443, opened 2022-08-13
fix: disallow leading zeroes in int literals and int types

### LINKED ISSUE #11963: compiler accepting leading zeroes
### Zig Version

0.10.0-dev.2657+74442f350

### Steps to Reproduce

Create the file `leading_0.zig` with this content:
```zig
fn main() {
    _ = 042;
}
```

### Expected Behavior
The code doesn't compile, and the compiler gives a suggestion to replace the literal with `0o42`.

### Actual Behavior
The compiler doesn't error besides the integer literal in line 2 starting with a zero. This behavior is a pretty big gotcha, as many languages such as C, Java and JavaScript will interpret literals starting with a zero as octal instead of decimal whereas zig simply ignores the zero.

### LINKED ISSUE #12417: Arbitrary bit-width integer types allow leading zeros
### Zig Version

0.9.1

### Steps to Reproduce

```zig
const std = @import("std");

pub fn main() !void {
    const num: i00000007 = 1;

    try stdout.print("Number: {d}\n", .{ num });
}
```

### Expected Behavior

The compiler should not accept integer types with leading zeros (i.e. `u03` or `i00000007`).

### Actual Behavior

```
Number: 1
```

r00ster91:
This makes `0123` and `u0123` etc. illegal.

I'm now confident that this is a good change because
I actually caught two C header translation mistakes in `haiku.zig` with this.
Clearly, `0123` being octal in C (TIL) can cause confusion, and we avoid ambiguity by
requiring `0o` as the prefix and now also disallowing leading zeroes in integers.

For consistency and because it looks weird, we disallow it for integer types too (e.g. `u0123`).

Fixes #11963
Fixes #12417


tiehuis:
As an aside, the current implementation of parseInt and parseFloat in std accept leading zeroes. May want to do a follow-up to this to modify their behavior if we are wanting consistency.

lib/std/c/haiku.zig
```diff
@@ -702,7 +702,7 @@ pub const T = struct {
     pub const CSETAF = 0x8002;
     pub const CSETAW = 0x8003;
     pub const CWAITEVENT = 0x8004;
-    pub const CSBRK = 08005;
+    pub const CSBRK = 0x8005;
```
marler8997, line 705: This brings me back.  I'll share the story for those interested. Years ago at my old job we were sometimes getting weird file permissions.  It was a very large code base (millions of LOC in hundreds of thousands of files) and was hard to reproduce.  Various people had spent days trying to reproduce/debug this issue but it's difficult when there's so much code to sift through.  I thought it could be some sort of uninitialized value issue but I was able to determine that each time it happened the permission bits were consistent.  After looking at the bit pattern of the permissions, it dawned on me that a user probably entered a hex integer literal when it was meant to be octal (i.e. they did something like `0x755` instead of `0755`).  I took the bit pattern and decoded it into a hex integer literal and used grep to search our repositories (probably took 5 or 10 minutes) and sure enough, I got a hit for a set of permission bits being passed to an `open` call! Changed that `0x` to an `0` and problem solved.

r00ster91: Wow, thanks for sharing! That really sounds like a pain. So much time spent for such a simple fix. Luckily with Zig this should never happen again!

r00ster91:
@tiehuis good point! I opened #12446.

Vexu (changes requested):
```
a.zig:1:11: error: expected expression, found 'invalid bytes'
const a = 00123;
          ^~~~~
a.zig:1:16: note: invalid byte: ';'
const a = 00123;
               ^
```
.. is a terrible compile error. This either needs a custom error in `Ast.zig` or, preferably an error in `AstGen.zig` since this doesn't prevent us from parsing properly.

r00ster91:
I agree. I thought it was very confusing too but then I noticed that we actually already have this problem with many other, similar cases too:
```zig
pub fn main() void {
    _ = 1_;
}
```
```
x.zig:4:9: error: expected expression, found 'invalid bytes'
    _ = 1_;
        ^~
x.zig:4:11: note: invalid byte: ';'
    _ = 1_;
          ^
```

```zig
pub fn main() void {
    _ = 1__1;
}
```
```
x.zig:4:9: error: expected expression, found 'invalid bytes'
    _ = 1__1;
        ^~
x.zig:4:11: note: invalid byte: '_'
    _ = 1__1;
          ^
```
I think it is very much part of the lump you will find when you search for "underscore_placement" on https://github.com/ziglang/zig/find/master

As you can see, they all expect the `// :2:21: error: expected expression, found 'invalid bytes'` error too, which I also find could probably be improved.

I think one simple thing we could do is use a more specific `result.tag` than `.invalid`? Currently we use `.invalid` for everything but we could use more specific ones. Is that an option?
I feel like this might be better suited as a follow-up PR.

r00ster91:
Besides, do all these "invalid_underscore_placement_in_"-named tests even belong in `zig/test/cases/compile_errors`? These tests seem like very simple tests that should just be as part of `test`s in `tokenizer.zig` (and probably most if not all of them already are).

Vexu:
> I agree. I thought it was very confusing too but then I noticed that we actually already have this problem with many other, similar cases too:

Yes, and I'm about to open an issue for this.

> I think one simple thing we could do is use a more specific `result.tag` than `.invalid`? Currently we use `.invalid` for everything but we could use more specific ones. Is that an option?

Most uses of `.invalid` are related to literals, it would be better to just validate them at a later stage than the tokenizer.

> I feel like this might be better suited as a follow-up PR.

Fixing the previous things sure, but I'd prefer we start this new error off properly.

Vexu (approved):

lib/std/c/haiku.zig
motiejus: I had exact this issue myself a couple of weeks ago in my zig codebase, and thought about suggesting this just a few days ago. @r00ster91 you got it , thanks.

nektro:
this broke `translate-c`

r00ster91:
How did it break? Any test case? Can you elaborate?

nektro:
```
/run/media/meghan/dev/mount.ufs/zig-cache/o/9aeca74be3315ca7b4accb89a5db24de/cimport.zig:1648:33: error: integer literal '00' has leading zero
pub const O_RDONLY = @as(c_int, 00);
                                ^~
```

r00ster91:
What is the C version of that?

nektro:
https://github.com/ziglang/zig/blob/master/lib/libc/musl/include/fcntl.h#L44

PR #12444, opened 2022-08-14
std.fs: Fix `WalkerEntry.dir` not always being the containing dir

squeek502:
https://github.com/ziglang/zig/blob/d2342370fe85491cb8216ccce9cc8fb5259376e2/lib/std/fs.zig#L860-L864

Before this commit, the modified test would fail with `FileNotFound` because the `entry.dir` would be for the entry itself rather than the containing dir of the entry. That is, if you were walking a tree of `a/b`, then (previously) the entry for `b` would incorrectly have an `entry.dir` for `b` rather than `a`.

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -877,8 +877,9 @@ pub const IterableDir = struct {
         /// a reference to the path.
         pub fn next(self: *Walker) !?WalkerEntry {
             while (self.stack.items.len != 0) {
-                // `top` becomes invalid after appending to `self.stack`
+                // `top` and `containing` become invalid after appending to `self.stack`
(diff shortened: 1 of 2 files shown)


lib/std/fs.zig
```diff
@@ -899,10 +900,11 @@ pub const IterableDir = struct {
                                 .dirname_len = self.name_buffer.items.len,
                             });
                             top = &self.stack.items[self.stack.items.len - 1];
```
squeek502, line 902: An alternate fix would be to just do `- 2` here, since `top` the only usage of `top` after this assignment is in the return statement below, but I think having a separate `containing` makes the intentions clearer.

Vexu (approved):

PR #12445, opened 2022-08-14
autodoc: better No Results Found page and other improvements

r00ster91:
image

For reference, here is Rust's page when you have no results: https://doc.rust-lang.org/std/collections/struct.HashSet.html?search=wdqqdw
As you can see, it offers you to make a search with DDG.
Even as mainly a DDG user myself, I think this is a little bit opinionated so I didn't add any explicit search engine link to search with. There are [other search engines](https://github.com/r00ster91/shifting#search_engines) too.

---

image

This indentation you can see is supposed to be a visual clue that those shortcuts are only available when having pressed <kbd>S</kbd> (or the search field is active).

### DIFF
--- a/lib/docs/index.html
+++ b/lib/docs/index.html
@@ -169,8 +169,8 @@
         width: 100%;
         margin-bottom: 0.8rem;
         padding: 0.5rem;
-        font-size: 1rem;
         font-family: var(--ui);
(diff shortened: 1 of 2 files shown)


kristoff-it:
This looks great, thank you very much!

PR #12448, opened 2022-08-14
std.ascii: rename functions and other improvements

r00ster91:
Supersedes #11629 

This should be a lot easier to review than #11629 because it does a lot less.
It mainly renames some functions and removes 3. It also adds a lot more tests.
See the commit descriptions for the reasons of all the deprecations and renamings.
They're all just deprecated so it will still work in 0.10 (if this gets into 0.10). After that they'll be removed.

You might also be interested in Rust's naming decisions: https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii

### DIFF
--- a/lib/std/ascii.zig
+++ b/lib/std/ascii.zig
@@ -517,8 +517,7 @@ pub fn orderIgnoreCase(lhs: []const u8, rhs: []const u8) std.math.Order {
     return std.math.order(lhs.len, rhs.len);
 }
 
-/// Returns true if lhs < rhs, false otherwise
-/// TODO rename "IgnoreCase" to "Insensitive" in this entire file.
+/// Returns whether `lhs` < `rhs`.
 pub fn lessThanIgnoreCase(lhs: []const u8, rhs: []const u8) bool {
     return orderIgnoreCase(lhs, rhs) == .lt;
 }
--- a/lib/std/ascii.zig
+++ b/lib/std/ascii.zig
@@ -232,7 +232,7 @@ pub fn isAlpha(c: u8) bool {
 }
 
 pub fn isCntrl(c: u8) bool {
-    return c < 0x20 or c == 127; //DEL
+    return c <= control_code.us or c == control_code.del;
 }
 
 pub fn isDigit(c: u8) bool {
--- a/lib/std/ascii.zig
+++ b/lib/std/ascii.zig
@@ -278,6 +278,7 @@ pub fn isPrint(c: u8) bool {
     return inTable(c, tIndex.Graph) or c == ' ';
 }
 
+/// DEPRECATED: create your own function based on your needs and what you want to do.
 pub fn isPunct(c: u8) bool {
     return inTable(c, tIndex.Punct);
 }
--- a/lib/std/ascii.zig
+++ b/lib/std/ascii.zig
@@ -12,43 +12,77 @@ const std = @import("std");
 /// Contains constants for the C0 control codes of the ASCII encoding.
 /// https://en.wikipedia.org/wiki/C0_and_C1_control_codes
 pub const control_code = struct {
-    pub const NUL = 0x00;
-    pub const SOH = 0x01;
-    pub const STX = 0x02;
-    pub const ETX = 0x03;
-    pub const EOT = 0x04;
-    pub const ENQ = 0x05;
-    pub const ACK = 0x06;
-    pub const BEL = 0x07;
-    pub const BS = 0x08;
-    pub const TAB = 0x09;
-    pub const LF = 0x0A;
-    pub const VT = 0x0B;
-    pub const FF = 0x0C;
-    pub const CR = 0x0D;
-    pub const SO = 0x0E;
-    pub const SI = 0x0F;
-    pub const DLE = 0x10;
-    pub const DC1 = 0x11;
-    pub const DC2 = 0x12;
-    pub const DC3 = 0x13;
-    pub const DC4 = 0x14;
-    pub const NAK = 0x15;
-    pub const SYN = 0x16;
-    pub const ETB = 0x17;
-    pub const CAN = 0x18;
-    pub const EM = 0x19;
-    pub const SUB = 0x1A;
-    pub const ESC = 0x1B;
-    pub const FS = 0x1C;
-    pub const GS = 0x1D;
-    pub const RS = 0x1E;
-    pub const US = 0x1F;
+    /// Null.
+    pub const nul = 0x00;
+    /// Start of Heading.
+    pub const soh = 0x01;
+    /// Start of Text.
+    pub const stx = 0x02;
+    /// End of Text.
+    pub const etx = 0x03;
+    /// End of Transmission.
+    pub const eot = 0x04;
+    /// Enquiry.
+    pub const enq = 0x05;
+    /// Acknowledge.
+    pub const ack = 0x06;
+    /// Bell, Alert.
+    pub const bel = 0x07;
+    /// Backspace.
+    pub const bs = 0x08;
+    /// Horizontal Tab, Tab ('\t').
+    pub const ht = 0x09;
+    /// Line Feed, Newline ('\n').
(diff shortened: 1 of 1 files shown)


topolarity:
Thanks for the improved doc comments and tidying things up here :+1:

Can you remind me which of these deprecations were decided previously, and which are new deprecations you're recommending?

lib/std/ascii.zig
```diff
@@ -188,73 +225,103 @@ fn inTable(c: u8, t: tIndex) bool {
...
+pub const isXDigit = isHex;
+
+/// Returns whether the character is alphanumeric. This is case-insensitive.
+pub fn isAlphanumeric(c: u8) bool {
     return (combinedTable[c] & ((@as(u8, 1) << @enumToInt(tIndex.Alpha)) |
         @as(u8, 1) << @enumToInt(tIndex.Digit))) != 0;
 }
 
-pub fn isAlpha(c: u8) bool {
+/// Returns whether the character is alphabetic. This is case-insensitive.
```
marler8997, line 249: What the difference between a case sensitive and a non case sensitive version of `isAlphabetic`?

lib/std/ascii.zig
```diff
@@ -1,54 +1,91 @@
...
+//!
+//! See also: https://en.wikipedia.org/wiki/ASCII#Character_set
 
 const std = @import("std");
 
-/// Contains constants for the C0 control codes of the ASCII encoding.
-/// https://en.wikipedia.org/wiki/C0_and_C1_control_codes
+/// The C0 control codes of the ASCII encoding.
+///
+/// See also: https://en.wikipedia.org/wiki/C0_and_C1_control_codes and `is_control`.
```
topolarity, line 15:
```suggestion
/// See also: https://en.wikipedia.org/wiki/C0_and_C1_control_codes and `isControl`.
```

lib/std/ascii.zig
```diff
@@ -188,73 +225,103 @@ fn inTable(c: u8, t: tIndex) bool {
...
+pub fn isWhitespace(c: u8) bool {
     return inTable(c, tIndex.Space);
 }
 
-/// All the values for which isSpace() returns true. This may be used with
-/// e.g. std.mem.trim() to trim whiteSpace.
-pub const spaces = [_]u8{ ' ', '\t', '\n', '\r', control_code.VT, control_code.FF };
+/// Whitespace for general use.
+/// This may be used with e.g. `std.mem.trim` to trim whitespace.
+/// See also: `isSpace`.
```
topolarity, line 293:
```suggestion
/// See also: `isWhitespace`.
```

lib/std/ascii.zig
```diff
@@ -263,6 +330,7 @@ pub fn toUpper(c: u8) u8 {
     }
 }
 
+/// Lowercases the character and returns it as-is if it's already lowercased.
```
topolarity, line 333:
```suggestion
/// Lowercases the character and returns it as-is if it's already lowercased or not a letter.
```

lib/std/ascii.zig
```diff
@@ -415,8 +519,7 @@ pub fn orderIgnoreCase(lhs: []const u8, rhs: []const u8) std.math.Order {
     return std.math.order(lhs.len, rhs.len);
 }
 
-/// Returns true if lhs < rhs, false otherwise
-/// TODO rename "IgnoreCase" to "Insensitive" in this entire file.
+/// Returns whether `lhs` < `rhs`.
```
topolarity, line 522: "lexicographically" would be nice here, for clarity

Something like:
```suggestion
/// Returns true if `lhs` < `rhs` lexicographically, false otherwise.
```

lib/std/ascii.zig
```diff
@@ -188,73 +225,103 @@ fn inTable(c: u8, t: tIndex) bool {
...
 }
 
+/// Returns whether the character is lowercased.
 pub fn isLower(c: u8) bool {
     return inTable(c, tIndex.Lower);
 }
 
+/// Returns whether the character has some graphical representation and can be printed.
+/// This also returns `true` for the space character.
 pub fn isPrint(c: u8) bool {
```
topolarity, line 277: Might be helpful to mention that this is just `!isControl(c)`

matu3ba: Plus check for whitespace `c == ' '`.

lib/std/ascii.zig
r00ster91: Good point!

lib/std/ascii.zig
r00ster91: Sounds good!

But, I don't even understand why we need to have this function? If we have this, why don't we also have `greaterThanIgnoreCase`? I don't know but `orderIgnoreCase(lhs, rhs) == .lt` is really not a lot to type.

topolarity: I believe it's for convenience when a `lessThan` function is needed for generics (e.g. `std.sort`)

lib/std/ascii.zig
r00ster91: Great point!

r00ster91:
Thanks for the reviews! Missed some stuff there.

> Can you remind me which of these deprecations were decided previously, and which are new deprecations you're recommending?

Hmm, none of my suggestions have really been "officially" discussed like that.
You can take it all as new suggestions except for the removal of `isPunct` which [Andrew wanted too](https://github.com/ziglang/zig/issues/8419#issuecomment-843719898).
I tried to go with names that are not cryptic, readable, and yet short enough.
For inspiration I also looked at Rust's naming choices (if you scroll down on https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii).

Obviously this is not set in stone and we're still far from 1.0 but this is a start and should make it less painful to do these deprecations later.

topolarity (approved):
LGTM - I think the new names are much clearer

The only bikeshedding I might encourage would be whether `isGraph` should be kept instead of `isPrint`. Since `isPrint(c) == !isControl(c)`, it seems like `isGraph` would be the more useful one to keep around

I could be convinced either way, though

lib/std/ascii.zig
```diff
@@ -1,54 +1,91 @@
...
 
 const std = @import("std");
 
-/// Contains constants for the C0 control codes of the ASCII encoding.
-/// https://en.wikipedia.org/wiki/C0_and_C1_control_codes
+/// The C0 control codes of the ASCII encoding.
+///
+/// See also: https://en.wikipedia.org/wiki/C0_and_C1_control_codes and `isControl`.
 pub const control_code = struct {
-    pub const NUL = 0x00;
```
marler8997, line 15: I'm not sure how common these control codes are used, but I think it would be an overall win to keep these values as deprecated for a time before we remove them.

r00ster91: I thought that was going to be a bit too much but oh well I guess it doesn't do harm to properly deprecate these as well.

r00ster91:
@topolarity I think `isGraph`'s name is confusing because `' '` too has a *graph*ical representation. I think having the two `isPrint` and `isControl` makes code easier to read and to me it feels a bit more uniform this way.

I feel like in general just knowing whether a character is printable or not is all that most users need. When do you actually need to know whether it's printable and *not* a space?

If we look at https://github.com/search?l=zig&p=2&q=isgraph&type=Code I can actually find many misuses of it or cases where people indeed just wanted to know if it's printable or not:
```zig
                if (std.ascii.isGraph(char) or char == ' ') {
```
```zig
                            if (!std.ascii.isGraph(c.?) and !std.ascii.isSpace(c.?)) {
```
And as for the other results it's mostly just them having a copy of the std in their project somehow and the rest are probably more misuses with a few legit uses perhaps.

r00ster91:
Is there anything blocking this from getting merged? Would be great to have this in before 0.10. Updating the rest of the codebase to use the new functions can come later.

andrewrk:
Thanks @r00ster91!



> Is there anything blocking this from getting merged?

No, I just needed to find some time to look at it :)

PR #12457, opened 2022-08-16
build: Allow comptime_int in OptionsStep.addOption

ThreeFx:
I don't see a reason why this shouldn't be allowed.

### DIFF
--- a/lib/std/build/OptionsStep.zig
+++ b/lib/std/build/OptionsStep.zig
@@ -171,6 +171,7 @@ fn printLiteral(out: anytype, val: anytype, indent: u8) !void {
         .Void,
         .Bool,
         .Int,
+        .ComptimeInt,
         .Float,
         .Null,
         => try out.print("{any}", .{val}),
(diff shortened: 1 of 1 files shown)


ThreeFx:
Note that the tests currently fail using the stage1 compiler, but it seems to be not my fault, since deleting the failing cases makes the test pass:

```sh
% cat lib/std/build/OptionsStep.zig | grep 'test "' -A 23
```
```zig
test "OptionsStep" {
    if (builtin.os.tag == .wasi) return error.SkipZigTest;

    var arena = std.heap.ArenaAllocator.init(std.testing.allocator);
    defer arena.deinit();
    var builder = try Builder.create(
        arena.allocator(),
        "test",
        "test",
        "test",
        "test",
    );
    defer builder.destroy();

    const options = builder.addOptions();
    options.addOption(comptime_int, "option4", 4);

    try std.testing.expectEqualStrings(
        \\pub const option4: comptime_int = 4;
        \\
    , options.contents.items);

    _ = try std.zig.parse(arena.allocator(), try options.contents.toOwnedSliceSentinel(0));
}
```
```sh
% ./cmake/zig test lib/std/build/OptionsStep.zig
All 1 tests passed.
```

ifreund (approved):
LGTM, Thanks!

Issue #12422, opened 2022-08-12
stack traces tests failing on Windows


andrewrk:
See  #12380  #12381

```
Test 1/9 stack-trace return (Debug)...
========= Expected this output: =========
error: TheSkyIsFalling
source.zig:2:5: [address] in main (test)
    return error.TheSkyIsFalling;
    ^

================================================
error: TheSkyIsFalling
source.zig:2:5: [address] in ain (test)
    return error.TheSkyIsFalling;
    ^

error: TestFailed
```

PR #12464, opened 2022-08-17
libstd: fix off-by-one error in def of ProcSym in pdb.zig

kubkon:
Fixes: #12422 

Make sure `ProcSym` includes a single element byte-array which delimits the start of the symbol's name as part of its definition. This makes the code more elegant in that accessing the name is equivalent to taking the address of this one element array.

Prior to this fix, `ProcSym` was defined as `packed struct` and reported size of `35`. After it was changed to `extern struct` in https://github.com/ziglang/zig/commit/b95942744c7ded279f5695ed20fdbbc806323cba the size reported is rounded to `36` (which is correct). However, according to https://github.com/microsoft/microsoft-pdb/blob/082c5290e5aff028ae84e43affa8be717aa7af73/include/cvinfo.h#L3722 `ProcSym` should include a one-element byte array which delimits the start of the symbol name which brings the size of  `ProcSym` to `36` without padding.

### DIFF
--- a/lib/std/pdb.zig
+++ b/lib/std/pdb.zig
@@ -310,6 +310,10 @@ pub const SymbolKind = enum(u16) {
 
 pub const TypeIndex = u32;
 
+// TODO According to this header:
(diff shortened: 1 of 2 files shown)


andrewrk:
Nice find!

PR #12475, opened 2022-08-18
optimize std.fs.Dir.makeOpenPath for the case that the path already exists

### LINKED ISSUE #12474: optimize std.fs.Dir.makeOpenPath for the case that the path already exists
Current implementation:

https://github.com/ziglang/zig/blob/79757f233d9bfc646caa13d20243266a19bbdf91/lib/std/fs.zig#L1406-L1413

Typically, something like this will happen:

```strace
mkdirat(AT_FDCWD, "/home/andy/.cache/zig", 0755) = -1 EEXIST (File exists)
openat(AT_FDCWD, "/home/andy/.cache/zig", O_RDONLY|O_CLOEXEC|O_PATH|O_DIRECTORY) = 5
```

We can optimize for the common case by first attempting the openat() and then falling back to mkdirat() if we get ENOENT. This will make the common case one syscall instead of two.

As a bonus, consider solving the TODO above as well regarding Windows. NtCreateFile is capable of creating and opening a directory with a single call.

joedavis:
Updates std.fs.Dir.makeOpenPath (and its iterable counterpart) to use a single syscall in most cases (100% of the time on Windows).

openDirAccessMaskW is updated to take an additional configuration option, which toggles between FILE_OPEN_IF and FILE_OPEN to support this.

Resolves #12474

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -1407,18 +1407,50 @@ pub const Dir = struct {
     /// If supported by the OS, this operation is atomic. It is not atomic on
     /// all operating systems.
     pub fn makeOpenPath(self: Dir, sub_path: []const u8, open_dir_options: OpenDirOptions) !Dir {
-        // TODO improve this implementation on Windows; we can avoid 1 call to NtClose
-        try self.makePath(sub_path);
-        return self.openDir(sub_path, open_dir_options);
+        if (builtin.os.tag == .windows) {
+            const w = os.windows;
+            const sub_path_w = try w.sliceToPrefixedFileW(sub_path);
+
+            const base_flags = w.STANDARD_RIGHTS_READ | w.FILE_READ_ATTRIBUTES | w.FILE_READ_EA |
+                w.SYNCHRONIZE | w.FILE_TRAVERSE;
+            return self.openDirAccessMaskW(sub_path_w.span().ptr, base_flags, .{
+                .no_follow = open_dir_options.no_follow,
+                .create = true,
+            });
+        }
+
+        if (self.openDir(sub_path, open_dir_options)) |dir| {
+            return dir;
+        } else |_| {
+            try self.makePath(sub_path);
+            return self.openDir(sub_path, open_dir_options);
+        }
     }
 
     /// This function performs `makePath`, followed by `openIterableDir`.
     /// If supported by the OS, this operation is atomic. It is not atomic on
     /// all operating systems.
     pub fn makeOpenPathIterable(self: Dir, sub_path: []const u8, open_dir_options: OpenDirOptions) !IterableDir {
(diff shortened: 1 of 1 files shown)


joedavis:
Looks like the build is failing on Windows CI. Once I’m back at my computer I’ll take a look and see what discrepancy there is with my local setup.

lib/std/fs.zig
```diff
@@ -1731,7 +1771,7 @@ pub const Dir = struct {
             null,
             0,
```
kubkon, line 1772: I might be off here, but my hunch tells me you want `FILE_ATTRIBUTE_NORMAL` if creating a dir, `0` otherwise.

kubkon: Oh yep, Windows docs confirm it [ZwCreateFile](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-zwcreatefile):

> [in] FileAttributes

> Specifies one or more FILE_ATTRIBUTE_XXX flags, which represent the file attributes to set if you create or overwrite a file. The caller usually specifies FILE_ATTRIBUTE_NORMAL, which sets the default attributes. For a list of valid FILE_ATTRIBUTE_XXX flags, see the [CreateFile](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea) routine in the Microsoft Windows SDK documentation. If no file is created or overwritten, FileAttributes is ignored.

kubkon (changes requested):

lib/std/fs.zig
```diff
@@ -1407,18 +1407,50 @@ pub const Dir = struct {
...
+                .create = true,
+            });
+        }
+
+        if (self.openDir(sub_path, open_dir_options)) |dir| {
+            return dir;
+        } else |_| {
+            try self.makePath(sub_path);
+            return self.openDir(sub_path, open_dir_options);
+        }
```
nektro, line 1427: should be `return self.openDir(...) catch ...` and do `makePath()` on `FileNotFound` only

lib/std/fs.zig
```diff
@@ -1407,18 +1407,50 @@ pub const Dir = struct {
...
+                }),
+            };
+        }
+
+        if (self.openIterableDir(sub_path, open_dir_options)) |dir| {
+            return dir;
+        } else |_| {
+            try self.makePath(sub_path);
+            return self.openIterableDir(sub_path, open_dir_options);
+        }
```
nektro, line 1453: same as above

nektro (changes requested):

andrewrk:
Hello - I have attempted to rebase this against master branch on your behalf in order to re-run the failing CI checks. However you have disabled "allow maintainers to push to this branch". So please re-open this PR against master branch after addressing the review feedback and failing CI tests.

PR #12476, opened 2022-08-18
copyFile: use FICLONE when possible

motiejus:
Linux supports `ioctl_ficlone(2)`: this is a CoW mechanism to copy a
file. Benefits over `copy_file_range`:
- it's latency does not depend on the file size.
- blocks will be reused, so when the file is copied this way, it will
  not take extra space (until either file is modified).

This is different from a hardlink, because the inode is different, thus
these two files are semantically (from a user's perspective) different.

I suggest making this a default, since, if you do `cp a b` on Linux
since coreutils 9.0, it will reflink by default[1].

As of writing it is supported on btrfs, cifs, nfs, ocfs2, overlayfs and
xfs; for this reason I highly encourage using one of those file systems.
:)

[1]: https://lists.gnu.org/archive/html/info-gnu/2021-09/msg00010.html

### DIFF
--- a/lib/std/fs.zig
+++ b/lib/std/fs.zig
@@ -2798,8 +2798,17 @@ fn copy_file(fd_in: os.fd_t, fd_out: os.fd_t) CopyFileRawError!void {
     }
 
     if (builtin.os.tag == .linux) {
-        // Try copy_file_range first as that works at the FS level and is the
-        // most efficient method (if available).
+        // Try reflink first, as it's O(1). As of Linux 5.19 this works on
+        // btrfs, cifs, nfs, ocfs2, overlayfs and xfs; source and destination
+        // have to be logically on the same partition.
+        //
+        // `cp --reflink=auto` is the default in coreutils 9.0+:
+        // https://lists.gnu.org/archive/html/info-gnu/2021-09/msg00010.html
+        if (os.linux.clone_file(fd_in, fd_out) == 0)
+            return;
+
+        // copy_file_range works at the FS level and is the second most
+        // efficient method (if kernel is new enough).
         var offset: u64 = 0;
         cfr_loop: while (true) {
             // The kernel checks the u64 value `offset+count` for overflow, use
--- a/lib/std/os/linux.zig
+++ b/lib/std/os/linux.zig
@@ -1522,6 +1522,12 @@ pub fn signalfd(fd: fd_t, mask: *const sigset_t, flags: u32) usize {
     return syscall4(.signalfd4, @bitCast(usize, @as(isize, fd)), @ptrToInt(mask), NSIG / 8, flags);
 }
 
+pub fn clone_file(fd_in: fd_t, fd_out: fd_t) usize {
+    assert(fd_in >= 0);
+    assert(fd_out >= 0);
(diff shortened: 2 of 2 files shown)


lib/std/os/linux.zig
```diff
@@ -1522,6 +1522,12 @@ pub fn signalfd(fd: fd_t, mask: *const sigset_t, flags: u32) usize {
     return syscall4(.signalfd4, @bitCast(usize, @as(isize, fd)), @ptrToInt(mask), NSIG / 8, flags);
 }
 
+pub fn clone_file(fd_in: fd_t, fd_out: fd_t) usize {
+    assert(fd_in >= 0);
+    assert(fd_out >= 0);
+    return ioctl(fd_out, T.FICLONE, @bitCast(usize, @as(isize, fd_in)));
+}
+
```
andrewrk, line 1530: This file is for raw syscalls and should not have any higher level abstractions in it. Check `std.os.ioctl_SIOCGIFINDEX` for an example of how you could create such a "clone file" abstraction. This will additionally ensure that the call goes through libc `ioctl` when the application links libc.

lib/std/fs.zig
```diff
@@ -2798,8 +2798,17 @@ fn copy_file(fd_in: os.fd_t, fd_out: os.fd_t) CopyFileRawError!void {
...
-        // Try copy_file_range first as that works at the FS level and is the
-        // most efficient method (if available).
+        // Try reflink first, as it's O(1). As of Linux 5.19 this works on
+        // btrfs, cifs, nfs, ocfs2, overlayfs and xfs; source and destination
+        // have to be logically on the same partition.
+        //
+        // `cp --reflink=auto` is the default in coreutils 9.0+:
+        // https://lists.gnu.org/archive/html/info-gnu/2021-09/msg00010.html
+        if (os.linux.clone_file(fd_in, fd_out) == 0)
+            return;
```
andrewrk, line 2808: Have a look at `std.os.sendfile` for how to do an Linux OS version check to determine if it's worth it to even attempt this call.

Also, some errors should be handled by returning them directly, such as FileNotFound. This will avoid a redundant syscall (and race conditions!) in some cases.

andrewrk (changes requested):
Thanks for the patch!

motiejus:
I realized that updating `copy_file_range` to use `ioctl.FICLONERANGE` is of higher value than adding a `cloneFile`, and will immerse my feet better into this area.

I will soon suggest a different PR that updates `copy_file_range` to use `FICLONERANGE` if that's available.

motiejus:
I am almost done changing `file_clone_range` on top of `FICLONERANGE` (followed by the previous-copy_file_range and the fallbacks in the same function). For that I needed to remove `args` from `copy_file_range`. That's documented to be 0, for future extensions. Makes me a bit unease though.

Questions:
- is it OK to remove the future extensibility of `args` from `copy_file_range`, or should I create `os.clone_file_range` instead?
- If I create a new one, should it have fallbacks like `clone_file_range`?

`args` is already a misnomer in `copy_file_range`, since they are (and can't) be used in the fallback code. So I am unsure what level of abstraction are we aiming for.

PR #12489, opened 2022-08-19
copy_file_range: use FICLONERANGE when possible

motiejus:
This is a follow-up from #12476: instead of adding a new abstraction to
clone files, start with `copy_file_range`. They use the same mechanism
in the kernel.

Benefits of FICLONERANGE vs `copy_file_range(2)`:
- O(1).
- CoW: so if files are not modified, it will not take additional space.

However, it is more restricted than copy_file_range: source and
destination must be on the same partition, and only some file systems
implement this. As of Linux 5.19 those are btrfs, cifs, nfs, ocfs2,
overlayfs and xfs[1].

Note: I removed `flags` from `copy_file_range` (that must be 0 as of writing). If we want to retain `flags`, this change shouldn't be as-is, and we probably need `os.clone_file_range`. If we add `clone_file_range`, should that function have fallbacks, like the `copy_file_range` does now?

[1]: https://elixir.bootlin.com/linux/v5.19/A/ident/remap_file_range

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -6260,32 +6260,64 @@ pub const CopyFileRangeError = error{
 
 var has_copy_file_range_syscall = std.atomic.Atomic(bool).init(true);
 
-/// Transfer data between file descriptors at specified offsets.
-/// Returns the number of bytes written, which can less than requested.
+/// Transfer data between file descriptors at specified offsets. Returns the
+/// number of bytes written, which can be less than requested.
+/// The `copy_file_range` call copies `len` bytes from one file descriptor to
+/// another. When possible, this is done within the operating system kernel,
+/// which can provide better performance characteristics than transferring data
+/// from kernel to user space and back.
 ///
-/// The `copy_file_range` call copies `len` bytes from one file descriptor to another. When possible,
-/// this is done within the operating system kernel, which can provide better performance
-/// characteristics than transferring data from kernel to user space and back, such as with
-/// `pread` and `pwrite` calls.
+/// `fd_in` must be a file descriptor opened for reading, and `fd_out` must be
+/// a file descriptor opened for writing. They may be any kind of file
+/// descriptor; however, if `fd_in` is not a regular file system file, it may
+/// cause this function to fall back to calling `pread` and `pwrite`, in which
+/// case atomicity guarantees no longer apply.
 ///
-/// `fd_in` must be a file descriptor opened for reading, and `fd_out` must be a file descriptor
-/// opened for writing. They may be any kind of file descriptor; however, if `fd_in` is not a regular
-/// file system file, it may cause this function to fall back to calling `pread` and `pwrite`, in which case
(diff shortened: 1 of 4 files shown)


lib/std/os.zig
```diff
@@ -6260,32 +6260,64 @@ pub const CopyFileRangeError = error{
...
+
+    if (ficlone_range) {
+        const arg = linux.FICLONERANGE_arg{
+            .src_fd = fd_in,
+            .src_offset = off_in,
+            .src_length = len,
+            .dest_offset = off_out,
+        };
+        while (true) {
+            const rc = system.ioctl(fd_out, linux.T.FICLONERANGE, @ptrToInt(&arg));
```
Jarred-Sumner, line 6304: Did you benchmark the impact when used on filesystems which do not support CoW clones? In that case, this adds an extra system call for each run

As of at least Ubuntu 20.04, the default filesystem is ext4 which does not support CoW clones

motiejus: I should have put this to the commit message; I am not overly concerned with the extra syscall, since coreutils is doing reflinking by default for `cp`; so I punted on the costs.

I see some other issues with the PR, marking as draft.

motiejus: Linux kernel refuses to accept the argument, returns EINVAL for anything I give it. I will try with a C version, but not today.

motiejus: Thanks for pointing out the number of syscalls. We may get rid of one in https://github.com/ziglang/zig/pull/12491

motiejus:
Here is the C version:

```
#include <stdio.h>
#include <sys/ioctl.h>
#include <linux/fs.h>

int main(int argc, const char** argv) {
    int exit_code = 0;
    const char *arg1 = argv[1];
    const char *arg2 = argv[2];

    FILE *f1 = fopen(arg1, "r");
    if (f1 == NULL) { perror("fopen 1"); }
    FILE *f2 = fopen(arg2, "w");
    if (f2 == NULL) { perror("fopen 2"); }
    int fd1 = fileno(f1);
    int fd2 = fileno(f2);

    struct file_clone_range arg;
    arg.src_fd = fd1;
    arg.src_offset = 0;
    arg.src_length = 209;
    arg.dest_offset = 0;

    int res = ioctl(fd2, FICLONERANGE, &arg);
    if (res != 0) {
        perror("ioctl_FICLONERANGE");
        exit_code = 1;
    }

cleanup:
    fclose(f1);
    fclose(f2);

    return exit_code;
}
```

Observations:
1. When `src_offset = 0, dest_offset = 0, src_length = N`, when N is the size of `fd_in` in bytes, then the function succeeds.
2. On any other cases it fails with `EINVAL`.

Tried on Linux 5.15.0-40 (ubuntu) x86_64 with XFS and btrfs.

I may dig into the kernel code separately (to report and/or fix), but for now it's suffice to say that the FICLONERANGE is not stable enough for use in zig.

PR #12368, opened 2022-08-08
make self-hosted the default compiler

andrewrk:
stage1 is available behind the `-fstage1` flag.

closes #89

### General
* [x] [create an upgrade guide to help smooth the process out for people who want to try the new compiler before the 0.10.0 release.](https://github.com/ziglang/zig/wiki/Self-Hosted-Compiler-Upgrade-Guide)
* [x] #12222
* [x] #12386
* [ ] ~add valgrind integration to self-hosted~ I'll do this later

### x86_64-freebsd
* [x] LavaTech outage
* [x] ZIG_EXECUTABLE set to bad value in ci script
* [x] stale CI tarball
* [x] ld: error: undefined symbol, compiler rt functions
* [x] not picking up libc headers, sys/endian.h

### aarch64-linux:
* [x] stage3 crashing building behavior tests
* [x] stage3 crashing building one of the test-compare-output cases

### x86_64-linux:
* [x] test-cases hanging
* [x] run-translated-c remainder test failing

### x86_64-macos:
* [x] error return traces on the wrong column
* [x] #12387
* [x] LLVM module verification failed
* [x] crashing when building langref

### x86_64-windows:
* [x] #12380
* [x] TLS miscompilation
* [x] stage1 OOM when building stage2
* [x] stage2 zig build failing with "error: FileNotFound"
* [x] not doing proper mangling of DLL export names

PR #12492, opened 2022-08-19
coff: specify default base path for relative source paths in pdb

kubkon:
This commit fixes inconsistent paths in stack trace printouts on Windows. Namely, we no longer have this situation:

```zig
pub fn main() void {
    gonnaFail();
}

fn gonnaFail() void {
    unreachable;
}
```

#### Current behavior:

```
$ main.exe
thread 17744 panic: reached unreachable code
C:/Users/kubkon/dev/examples/stack-traces/main.zig:6:5: 0x7ff6409a126f in gonnaFail (main.exe.obj)
    unreachable;
    ^
C:/Users/kubkon/dev/examples/stack-traces/main.zig:2:14: 0x7ff6409a123e in main (main.exe.obj)
    gonnaFail();
             ^
C:\Users\kubkon\dev\zig\stage3\lib\zig\std\start.zig:345:41: 0x7ff6409a12ca in WinStartup (main.exe.obj)
    }
                                        ^
Unable to dump stack trace: FileNotFound
```

#### With this patch:

```
$ main.exe
thread 17744 panic: reached unreachable code
C:\Users\kubkon\dev\examples\stack-traces\main.zig:6:5: 0x7ff6409a126f in gonnaFail (main.exe.obj)
    unreachable;
    ^
C:\Users\kubkon\dev\examples\stack-traces\main.zig:2:14: 0x7ff6409a123e in main (main.exe.obj)
    gonnaFail();
             ^
C:\Users\kubkon\dev\zig\stage3\lib\zig\std\start.zig:345:41: 0x7ff6409a12ca in WinStartup (main.exe.obj)
    }
                                        ^
Unable to dump stack trace: FileNotFound
```

The reason this happens is that PDB requires all file paths to be fully resolved, and it is really the linker's responsibility to canonicalize any path extracted from the CodeView in the object file. However, `LLD-link` has some very questionable defaults, and in particular, it purposely bakes in path separator of the host system it was built on rather the target's (to be honest, even throwing an error in such case would be better than a default like that IMHO). And so, because we built the Win devkit on a Linux host, `LLD-link` has POSIX-style separators baked in, and hence resolves relative paths such as `.\main.zig` into `C:/something/main.zig`.

Thankfully, they have left a backdoor we can use via `-PDBSOURCEPATH` flag.

From `LLD-link`'s help:

```
/pdbsourcepath:<value>  Base path used to make relative source file path absolute in PDB
```

If we are building a Zig module, we will provide a PDB source path prefix (if necessary) which overwrites path canonicalization in `LLD-link`. For context, code responsible for this in `LLD`: [lld/COFF/PDB.cpp#L243](https://github.com/llvm/llvm-project/blob/ac31781759889226711b801c3833fea280fdb7aa/lld/COFF/PDB.cpp#L243).

I would also like to point out the reason we didn't see this happen in stage1 is because in stage1 we always embedded fully resolved paths (with the exception when we are targeting MachO). Now that in stage3 we are including relative and absolute paths, this problem has surfaced.

Finally, with this commit we can revert half of https://github.com/ziglang/zig/pull/12368/commits/f2d30b3a6857a23a8c92c2f948e38df69a8df54e as the path separator is now consistent with the platform we are running the binary on.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1016,6 +1016,9 @@ pub const InitOptions = struct {
     /// (Darwin) remove dylibs that are unreachable by the entry point or exported symbols
     dead_strip_dylibs: bool = false,
     libcxx_abi_version: libcxx.AbiVersion = libcxx.AbiVersion.default,
+    /// (Windows) PDB source path prefix to instruct the linker how to resolve relative
+    /// paths when consolidating CodeView streams into a single PDB file.
+    pdb_source_path: ?[]const u8 = null,
 };
 
 fn addPackageTableToCacheHash(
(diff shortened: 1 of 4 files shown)


andrewrk:
Needs a rebase against master branch.

jcmoyer:
I think this is also preventing benchmarking zig programs using WPR/WPA, WPA refuses to load PDBs with `/` in the filename, but hex editing the executable and replacing `/` with `\` makes it work.

kubkon:
Merged in 1357790ec969bb6ee19ade6e8a348bd9d7cbbc4d

PR #12494, opened 2022-08-19
Add Random.weightedIndex function

jwhear:
The ~~`dice`~~ `weightedIndex` function picks from a list of weighted indices.  This is useful in situations where the developer needs to select from a range of options which are not equally weighted; this is common in simulations or algorithms like Wave Function Collapse.

~~A note on naming: I went with `dice` because that's what the algorithm is called in the [D standard library](https://dlang.org/library/std/random/dice.html).  Other good options would be `weightedIndex` and `pickIndex`.~~

This implementation should work with all primitive numeric types.

### DIFF
--- a/lib/std/rand.zig
+++ b/lib/std/rand.zig
@@ -343,7 +343,7 @@ pub const Random = struct {
     ///
     /// This is useful for selecting an item from a slice where weights are not equal.
     /// `T` must be a numeric type capable of holding the sum of `proportions`.
-    pub fn dice(r: std.rand.Random, comptime T: type, proportions: []T) usize {
+    pub fn weightedIndex(r: std.rand.Random, comptime T: type, proportions: []T) usize {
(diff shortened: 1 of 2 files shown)


Deecellar:
Personally I would prefer pickWeightedIndex or weightedIndex, since dice would be a bit misleading

Atomk:
I second the idea of including "weighted" in the function name, it's much more descriptive of what the algorithm does.

jwhear:
Based on the naming of existing Random methods (nouns rather than verbs), I am thinking that `weightedIndex` or possibly just `weighted` might be the best name for this.

jwhear:
I'm thinking these CI failures aren't my fault?

InKryption:
I just realized this, but shouldn't the parameter be `proportions: []const T`?

Vexu:
It should.

Issue #11112, opened 2022-03-10
doc/langref: document void literals


perillo:
Looking at the `std` code, it seems that the void literal is `void{}`.

I think the documentation should have a `Void Literals` section, since this is not documented.
I found it when I was playing with the syntax and later found some examples in the `std`.

perillo:
There is actually an example in the documentation, in https://ziglang.org/documentation/master/#Zero-Bit-Types

```zig
var x: void = {}
```

However
```zig
var x = u0{}
```
is not allowed.

Vexu:
`{}` is an empty block which happens to have type `void`, `void{}` is a void literal.

perillo:
@Vexu Thanks.

I think this should be documented in https://ziglang.org/documentation/master/#blocks, because the current documentation only says that:
Blocks are expressions. When labeled, break can be used to return a value from the block.

So, IMHO, it is not evident that a block happens to have type void (well, probably it is, since a block is an expression).

PR #12496, opened 2022-08-19
Add block expression return type doc

shaneikennedy:
Addresses #11112

Add doc on default return type for block expressions, with test case. I'm not sure it's the most appropriate test case (comparing void to void) but couldn't think of something better, if you have any ideas add them here and I'm happy to fix this up 👍

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -4041,7 +4041,22 @@ test "access variable after block scope" {
     x += 1;
 }
       {#code_end#}
-      <p>Blocks are expressions. When labeled, {#syntax#}break{#endsyntax#} can be used
+      <p>Blocks are expressions and by default they are a void literal:</p>
+      {#code_begin|test|test_block_void_literal#}
+const std = @import("std");
+const expect = std.testing.expect;
+
+test "evaluates to void" {
+
+    const x = {
+        var y: i32 = 123;
+        y += 1;
+    };
+    try expect(@TypeOf(x) == void);
+}
+      {#code_end#}
+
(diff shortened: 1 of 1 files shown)


shaneikennedy:
Maybe `try expect(@TypeOf(x) == void);` is better.... lmk what you prefer

Vexu:
> Maybe `try expect(@TypeOf(x) == void);` is better.... lmk what you prefer

I'd prefer that.

shaneikennedy:
@Vexu Updated 👍

shaneikennedy:
@Vexu I've rebased a few times now and each time I get a different failing CI step, anything I need to do here or is it just flaky?

Vexu:
The CI is flaky, as long as one of the steps passes the `Generating docgen examples` phase then this is good. The bigger issue with this is that it only partially addresses the issue since it is missing a `Void Literals` section.

Vexu:
> There's a "void" type section https://ziglang.org/documentation/master/#void and I think concept of literals is fairly common, but I also see a section specifically for anon. struct literals https://ziglang.org/documentation/master/#Anonymous-Struct-Literals
> 
> What do you think about me adding a Void Literals section under https://ziglang.org/documentation/master/#void and I can link to it in the doc here https://github.com/ziglang/zig/pull/12496/files#diff-9674f03a76728eff712dfbd966487142bf4103b4a58049ee9696028e890f86bcR4044?

Whatever you think works best, as long as it mentions `void{}` somewhere and says that `{}` is usually preferred.

Issue #12517, opened 2022-08-20
stage2: zig cc compiled `powf` randomly returns NaN on Windows


jagt:
### Zig Version

0.10.0-dev.3653+7152a58c1

### Steps to Reproduce

```C++
#include <stdio.h>
#include <math.h>

int main() {
    printf("! %f\n", powf(1.389f, 100.0f + 0.1f));

    int acc = 0;
    for (float xx = 0.f; xx < 100.0f; xx += 0.1f)
    {
        float v = powf(1.389f, xx);
        printf("%f: %f\n", xx, v);
        if (acc++ > 3) break;
    }
 
    return 0;
}
```

### Expected Behavior

With clang it's like this:

```
clang -g -o powtest.exe .\powtest.c
.\powtest.exe
! 192529106468864.000000
0.000000: 1.000000
0.100000: 1.033404
0.200000: 1.067924
0.300000: 1.103597
0.400000: 1.140462
```

### Actual Behavior

```
zig cc -g -o powtest2.exe .\powtest.c
.\powtest2.exe
! nan
0.000000: 1.000000
0.100000: nan
0.200000: nan
0.300000: nan
0.400000: nan
```

Note that this is reproducible with the `zig.exe` from [zig win dev kit](https://github.com/ziglang/zig/wiki/Building-Zig-on-Windows#option-1-use-the-windows-zig-compiler-dev-kit)

Thanks in advance!

jagt:
I locally toggled verbose linker and seems the `pow` is from `mingwex`:

```
lld-link: Loaded mingwex.lib(pow.obj) for pow
lld-link: Reading msvcrt-os.lib(mingw_lock.obj)
...
lld-link: Reading mingwex.lib(ldexp.obj)
lld-link: Loaded mingwex.lib(ldexp.obj) for ldexp
lld-link: Reading mingwex.lib(powi.obj)
lld-link: Loaded mingwex.lib(powi.obj) for __powi
```

kubkon:
This is a miscompilation of stage2 of compiler-rt symbols. Building the same program with stage1 doesn't have this problem:

```
> zig cc -g -c powtest.c
> zig build-exe powtest.obj -fstage1 -lc
> .\powtest.exe
! 192529106468864.000000
0.000000: 1.000000
0.100000: 1.033404
0.200000: 1.067924
0.300000: 1.103597
0.400000: 1.140462
```

jagt:
[...]

PR #12522, opened 2022-08-20
zig cc: accept `-Xlinker -verbose`

jagt:
This passes along `-verbose` to lld for debug output.

I did notice there's `-verbose-link` and `ZIG_VERBOSE_LINK` env but seems it's mostly for printing `argv`. I'm not sure whether should I just use that to toggle `-verbose` for lld.

Below is a simple test case:

```c++
int f(int x) 
{
    return x+42;
}
#ifdef __WASM__
void _start()
{
    f(123);
}
#else
int main()
{
    return f(123);
}
#endif
```

```
zig cc -target wasm32-freestanding -D__WASM__ -Xlinker -verbose -g -o standalone.wasm standalone.c
zig cc -target x86_64-linux-musl -Xlinker -verbose -g -o standalone.out standalone.c
zig cc -target x86_64-windows-gnu -Xlinker -verbose -g -o standalone.exe standalone.c
```

Alternatively I think this can be hidden after a env like `ZIG_VERBOSE_LLD`.
Cheers!


kubkon:
What do you mean logged from the linker except the specified arguments (aka the linker line)? If it's only what we fed the linker `zig cc --verbose` already does that. If you need more, could you explain why?

kubkon:
I should also note that this will become obsolete once we start moving away from LLD for all targets and towards are in-house linking backends (the ZLD suite). Currently, this is only true for macOS, but other linkers will follow suit soon enough.

kubkon:
> > What do you need logged from the linker except the specified arguments (aka the linker line)? If it's only what we fed the linker `zig cc --verbose` already does that. If you need more, could you explain why?
> 
> Mostly for the debug messages from lld for tracking down which symbols are linked in. Its output is [like this](https://github.com/ziglang/zig/issues/12517#issuecomment-1221311030):
> 
> ```
> lld-link: Loaded mingwex.lib(pow.obj) for pow
> lld-link: Reading msvcrt-os.lib(mingw_lock.obj)
> ...
> lld-link: Reading mingwex.lib(ldexp.obj)
> lld-link: Loaded mingwex.lib(ldexp.obj) for ldexp
> lld-link: Reading mingwex.lib(powi.obj)
> lld-link: Loaded mingwex.lib(powi.obj) for __powi
> ```
> 
> This PR is mostly for completeness' sake as clang accepts `-Xlinker -verbse`. I'm fine with not merging this in though, as you've mentioned that we'll move away from LLD.
> 
> Cheers!

Hmm, I can actually see this useful for `LLD-link` in particular - it is a pain that by default no symbol table is generated in PE image files which makes it annoyingly cumbersome to reason if the symbols that were meant to get linked really did. Do you perhaps know of a way to force include a symtab by `LLD-link`?

PR #12526, opened 2022-08-20
macho: do not leave file descriptors open if unused

kubkon:
This manifested in different `AccessDenied` errors on Windows when
trying to do an atomic file copying in the compiler/linker.


kubkon:
I tried reproing the failure but everything works fine for me locally so I'm just going to merge and patch up in-tree if still fails.

Issue #12486, opened 2022-08-19
stage2: struct literal doesnt coerce to optional


nektro:
### Zig Version

0.10.0-dev.3618+f6459721e

### Steps to Reproduce

```zig
const Request = struct {
    params: ?ResponseParams,
};

const ResponseParams = union(enum) {
    ConfigurationParams: u8,
};

test {
    _ = Request{
        .params = .{
            .ConfigurationParams = 35,
        },
    };
}
```

### Expected Behavior

all tests passed

### Actual Behavior

```
[meghan@nixos]$ zig test test.zig -fno-stage1
test.zig:12:36: error: no field named 'ConfigurationParams' in union 'test.ResponseParams'
            .ConfigurationParams = 35,
                                   ^~
test.zig:5:24: note: union declared here
const ResponseParams = union(enum) {
                       ^~~~~
```

PR #12528, opened 2022-08-20
Sema: ensure resolveTypeFields is called for optional and error union types

martinhath:
We call `sema.resolveTypeFields` in order to get the fields of structs
and unions inserted into their data structures. If it isn't called, it
can happen that the fields of a type is queried before those fields are
inserted into (for instance) `Module.Union.fields`, which would result in
a wrong 'no field named' error.

Fixes: #12486

### DIFF
--- a/test/behavior.zig
+++ b/test/behavior.zig
@@ -85,6 +85,7 @@ test {
     _ = @import("behavior/bugs/12003.zig");
     _ = @import("behavior/bugs/12033.zig");
     _ = @import("behavior/bugs/12430.zig");
+    _ = @import("behavior/bugs/12486.zig");
     _ = @import("behavior/byteswap.zig");
     _ = @import("behavior/byval_arg_var.zig");
     _ = @import("behavior/call.zig");
--- a/test/behavior/bugs/12486.zig
+++ b/test/behavior/bugs/12486.zig
@@ -1,13 +1,17 @@
+const SomeEnum = union(enum) {
+    EnumVariant: u8,
+};
+
+const SomeStruct = struct {
+    struct_field: u8,
+};
+
 const OptEnum = struct {
     opt_enum: ?SomeEnum,
 };
 
 const ErrEnum = struct {
     err_enum: anyerror!SomeEnum,
-};
-
-const SomeEnum = union(enum) {
-    EnumVariant: u8,
 };
 
 const OptStruct = struct {
@@ -18,11 +22,7 @@ const ErrStruct = struct {
     err_struct: anyerror!SomeStruct,
 };
 
-const SomeStruct = struct {
-    struct_field: u8,
-};
-
-pub fn main() void {
+test {
     _ = OptEnum{
         .opt_enum = .{
             .EnumVariant = 1,
@@ -47,5 +47,3 @@ pub fn main() void {
         },
     };
 }
-
-// run
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -16128,9 +16128,10 @@ fn fieldType(
     field_src: LazySrcLoc,
     ty_src: LazySrcLoc,
 ) CompileError!Air.Inst.Ref {
-    const resolved_ty = try sema.resolveTypeFields(block, ty_src, aggregate_ty);
-    var cur_ty = resolved_ty;
+    var cur_ty = aggregate_ty;
     while (true) {
+        const resolved_ty = try sema.resolveTypeFields(block, ty_src, cur_ty);
+        cur_ty = resolved_ty;
         switch (cur_ty.zigTypeTag()) {
             .Struct => {
                 if (cur_ty.isAnonStruct()) {
--- a/test/cases/coercion-to-unused-union-and-struct.zig
+++ b/test/cases/coercion-to-unused-union-and-struct.zig
@@ -1,0 +1,51 @@
+const OptEnum = struct {
+    opt_enum: ?SomeEnum,
+};
+
+const ErrEnum = struct {
+    err_enum: anyerror!SomeEnum,
+};
+
+const SomeEnum = union(enum) {
+    EnumVariant: u8,
+};
+
+const OptStruct = struct {
+    opt_struct: ?SomeStruct,
+};
+
+const ErrStruct = struct {
+    err_struct: anyerror!SomeStruct,
+};
+
+const SomeStruct = struct {
+    struct_field: u8,
+};
+
+pub fn main() void {
+    _ = OptEnum{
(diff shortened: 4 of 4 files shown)


Vexu:
Those tests should go somewhere in the behavior tests.

martinhath:
Got it; I wasn’t sure where to put it and couldn’t find documentation on the difference of behavior and cases. It seemed to me that behavior tests were for testing if the code does the right thing and cases were for whether they compile or not, and my test is in the latter category.

test/behavior.zig
```diff
@@ -90,6 +90,7 @@ test {
     _ = @import("behavior/call.zig");
     _ = @import("behavior/cast.zig");
     _ = @import("behavior/cast_int.zig");
+    _ = @import("behavior/coercion-to-unused-union-and-struct.zig");
```
Vexu, line 93: The convention is to put these in the bugs folder
```suggestion
    _ = @import("behavior/bugs/12486.zig");
```

test/behavior/bugs/12486.zig
```diff
@@ -0,0 +1,49 @@
...
+
+const ErrStruct = struct {
+    err_struct: anyerror!SomeStruct,
+};
+
+const SomeStruct = struct {
+    struct_field: u8,
+};
+
+pub fn main() void {
```
Vexu, line 25: This needs to be a test in order to be run. You should also add some `expect`s to verify that the structs were actually initialized correctly.

martinhath: This would be good, expect that by testing against the fields we hit a path in the compiler that instantiates the fields of the structs and unions, and so the tests pass even before this fix. Running the test as written
```zig
    _ = OptEnum{
        .opt_enum = .{
            .EnumVariant = 1,
        },
    };
```
I'm getting, with the compiler at `master`, without this PR:
```
$ zig-out/bin/zig version
0.10.0-dev.3664+02070ae26
$ zig-out/bin/zig  test ../zig/test/behavior/bugs/12486.zig
../zig/test/behavior/bugs/12486.zig:40:28: error: no field named 'EnumVariant' in union '12486.SomeEnum'
            .EnumVariant = 1,
                           ^
../zig/test/behavior/bugs/12486.zig:4:18: note: union declared here
const SomeEnum = union(enum) {
                 ^~~~~
```
But if I change the test to
```zig
    const opt_enum = OptEnum{
        .opt_enum = .{
            .EnumVariant = 1,
        },
    };
    try expect(opt_enum.opt_enum.? == .EnumVariant);
```
it passes.

With this PR, both variants pass.

I could try to come up with a test that both fails with the old compiler, passes with the new, and also contains checks that the fields are what they're supposed to be, but that's testing something that wasn't a problem in the issue I'm trying to solve, so it feels out of scope to me.

Vexu: Try passing the struct as a function argument, that should have the same result location as discarding.

martinhath: Do you mean like this?
```zig
fn foo(oe: OptEnum) anyerror!void {
    try expect(oe.opt_enum.? == .EnumVariant);
}

test {
    try foo(OptEnum{
        .opt_enum = .{
            .EnumVariant = 1,
        },
    });
}
```
This has the same result as the test above, and `master` still passes on it.

Vexu: Then I'm out of ideas, @nektro how did you originally encounter this?

nektro: by building ZLS

nektro: https://github.com/zigtools/zls/blob/master/src/Server.zig#L1901

nektro: there was a workaround pr recently put in lemme find the commit i was on

nektro: https://github.com/zigtools/zls/blob/9161397/src/Server.zig#L1882 once you fix the two const errors in `Config.zig` you get

```
/home/meghan/src/zls/src/Server.zig:1886:37: error: no field named 'ConfigurationParams' in union 'types.ResponseParams'
            .ConfigurationParams = .{
                                   ~^
/home/meghan/src/zls/src/types.zig:35:28: note: union declared here
pub const ResponseParams = union(enum) {
                           ^~~~~
```

martinhath: Maybe I'll try to expand on what exactly is the problem here so that I'm sure we're all on the same page, as I wasn't too descriptive in neither the PR nor the commit message.

During semantic analysis the compiler makes a `Module.Union` for each `union` in the code. This struct has a field called `fields`, which, like many Zig things, is lazily populated at need. Defining "at need" is kinda what's the issue in #12486.

For context, the union fields are added to the `Module.Union` struct in `Sema.semaUnionFields`; a stack trace I had when testing things for this PR (and things worked as expected) was this:
```
...
    #4  0x000000000128adbe in Sema.semaUnionFields () at Sema.zig:28365
    #5  0x0000000000fa9a79 in Sema.resolveTypeFieldsUnion () at Sema.zig:27780
    #6  0x0000000000de2135 in Sema.resolveTypeFields () at Sema.zig:27707
    #7  0x00000000013a34c0 in Sema.fieldType () at Sema.zig:16131
    #8  0x00000000010ff83b in Sema.zirFieldType () at Sema.zig:16120
    #9  0x0000000000f0368e in Sema.analyzeBodyInner () at Sema.zig:811
    #10 0x0000000000efcbf9 in Sema.analyzeBody () at Sema.zig:604
    #11 0x0000000000d5bffd in Module.analyzeFnBody () at Module.zig:5529
    #12 0x0000000000b6a455 in Module.ensureFuncBodyAnalyzed () at Module.zig:4240
    #13 0x0000000000b689a3 in Compilation.processOneJob () at Compilation.zig:2974
...
```

For instance, instantiating a new value of the union type with the regular `UnionName{ ... }` does this, and accessing a field through a function parameter also does this. I assume pretty much all constructs where you end up needing this does this, since we *need* to do this to do anything useful with the union directly.

The problem is that trying to coerce a anonymous struct literal against that type does **not** do this when the type it should coerce to is `?MyUnion`. It does so if you remove the optional, as can be tried with the reproducing code in #12486. Another thing that fixes the repro code is insantiating another union before doing the coersion, like this:

```zig
const Request = struct {
    params: ?ResponseParams,
};

const ResponseParams = union(enum) {
    ConfigurationParams: u8,
};

test {
    _ = ResponseParams{ .ConfigurationParams = 1 };
    _ = Request{
        .params = .{
            .ConfigurationParams = 35,
        },
    };
}
```

PR #12539, opened 2022-08-20
sema: allow compare .lazy_size with a signed rhs value. fixes #12498

### LINKED ISSUE #12498: compiler crash with lazy abi size used in comparison with signed runtime rhs
### Zig Version

0.10.0-dev.3653+7152a58c1

### Steps to Reproduce

```zig
const S = struct {
    a: usize,
};

test "lazy abi size used in comparison" {
    var rhs: i32 = 100;
    if (@sizeOf(S) > rhs) {
        @panic("bad");
    }
}
```

### Expected Behavior

Test passed.

### Actual Behavior

```
thread 2901771 panic: reached unreachable code
Analyzing test3.zig: test3.zig:test.lazy abi size used in comparison
      %28 = decl_val("S") token_offset:12:17 to :12:18
      %29 = size_of(%28) node_offset:12:9 to :12:19
      %30 = load(%23) node_offset:12:22 to :12:25
    > %31 = cmp_gt(%29, %30) node_offset:12:9 to :12:25
      %32 = as_node(@Zir.Inst.Ref.bool_type, %31) node_offset:12:9 to :12:25
      %33 = condbr(%32, {
        %35 = dbg_block_begin())
        %36 = dbg_block_begin())
        %37 = dbg_stmt(4, 9)
        %38 = str("bad")
        %39 = as_node(@Zir.Inst.Ref.const_slice_u8_type, %38) node_offset:13:16 to :13:21
        %41 = dbg_block_end())
        %42 = dbg_block_end())
        %40 = panic(%39) node_offset:13:9 to :13:22
      }, {
        %43 = break(%34, @Zir.Inst.Ref.void_value)
      }) node_offset:12:5 to :12:7
      zig ast-check -t test3.zig

  in test3.zig: test3.zig:test.lazy abi size used in comparison
    > %34 = block({%28..%33}) node_offset:12:5 to :12:7

/home/andy/dev/zig/lib/std/debug.zig:281:14: 0x301a963 in assert (zig)
             ^
/home/andy/dev/zig/src/type.zig:5679:45: 0x35876ad in structFieldOffset (zig)
                assert(struct_obj.haveLayout());
                                            ^
/home/andy/dev/zig/src/type.zig:3245:75: 0x33f8c8c in abiSizeAdvanced (zig)
                    return AbiSizeAdvanced{ .scalar = ty.structFieldOffset(field_count, target) };
                                                                          ^
/home/andy/dev/zig/src/type.zig:3162:32: 0x3132ef8 in abiSize (zig)
        return (abiSizeAdvanced(ty, target, .eager) catch unreachable).scalar;
                               ^
/home/andy/dev/zig/src/value.zig:1101:37: 0x3756d98 in toBigIntAdvanced (zig)
                const x = ty.abiSize(target);
                                    ^
/home/andy/dev/zig/src/value.zig:1058:36: 0x3587c97 in toBigInt (zig)
        return val.toBigIntAdvanced(space, target, null) catch unreachable;
                                   ^
/home/andy/dev/zig/src/value.zig:1759:37: 0x3fa771c in intBitCountTwosComp (zig)
                return self.toBigInt(&buffer, target).bitCountTwosComp();
                                    ^
/home/andy/dev/zig/src/Sema.zig:26557:51: 0x3d58a4e in cmpNumeric (zig)
            lhs_bits = lhs_val.intBitCountTwosComp(target);
                                                  ^
/home/andy/dev/zig/src/Sema.zig:13364:31: 0x3b58b0a in analyzeCmp (zig)
        return sema.cmpNumeric(block, src, lhs, rhs, op, lhs_src, rhs_src);
                              ^
/home/andy/dev/zig/src/Sema.zig:13339:27: 0x38aee77 in zirCmp (zig)
    return sema.analyzeCmp(block, src, lhs, rhs, op, lhs_src, rhs_src, false);
                          ^
/home/andy/dev/zig/src/Sema.zig:725:61: 0x36edec8 in analyzeBodyInner (zig)
            .cmp_gt                       => try sema.zirCmp(block, inst, .gt),
                                                            ^
/home/andy/dev/zig/src/Sema.zig:4831:34: 0x3b86282 in resolveBlockBody (zig)
        if (sema.analyzeBodyInner(child_block, body)) |_| {
                                 ^
/home/andy/dev/zig/src/Sema.zig:4814:33: 0x395c048 in zirBlock (zig)
    return sema.resolveBlockBody(parent_block, src, &child_block, body, inst, &label.merges);
                                ^
/home/andy/dev/zig/src/Sema.zig:1233:69: 0x36fcaee in analyzeBodyInner (zig)
                if (!block.is_comptime) break :blk try sema.zirBlock(block, inst);
                                                                    ^
/home/andy/dev/zig/src/Sema.zig:604:30: 0x36ebf98 in analyzeBody (zig)
    _ = sema.analyzeBodyInner(block, body) catch |err| switch (err) {
                             ^
/home/andy/dev/zig/src/Module.zig:5529:21: 0x354b51c in analyzeFnBody (zig)
    sema.analyzeBody(&inner_block, fn_info.body) catch |err| switch (err) {
                    ^
/home/andy/dev/zig/src/Module.zig:4240:40: 0x3359544 in ensureFuncBodyAnalyzed (zig)
            var air = mod.analyzeFnBody(func, sema_arena) catch |err| switch (err) {
                                       ^
/home/andy/dev/zig/src/Compilation.zig:2974:42: 0x3357a92 in processOneJob (zig)
            module.ensureFuncBodyAnalyzed(func) catch |err| switch (err) {
                                         ^
/home/andy/dev/zig/src/Compilation.zig:2912:30: 0x3254632 in performAllTheWork (zig)
            try processOneJob(comp, work_item);
                             ^
/home/andy/dev/zig/src/Compilation.zig:2253:31: 0x3250f9a in update (zig)
                              ^
/home/andy/dev/zig/src/main.zig:3334:20: 0x327a856 in updateModule (zig)
    try comp.update();
                   ^
/home/andy/dev/zig/src/main.zig:3008:17: 0x303d259 in buildOutputType (zig)
                ^
/home/andy/dev/zig/src/main.zig:236:31: 0x30183fb in mainArgs (zig)
                              ^
/home/andy/dev/zig/src/main.zig:174:20: 0x3017840 in main (zig)
                   ^
/home/andy/dev/zig/lib/std/start.zig:574:37: 0x3019d79 in main (zig)
        },
                                    ^
Aborted (core dumped)
```

bfredl:
not sure if this is the right fix. but resolving the layout unconditionally causes another test to fail (`@sizeOf(T) == 0 doesn't force resolving struct size`) and based on @andrewrk labeling of my example failure, #12498, only doing it for possibly signed rhs seems like one possible fix.


bfredl:
doesn't seem to fully fix it in my original repo. I suppose we need to resolve `@sizeOf(S) > unsigned_runtime_val` as well? (but not with `==` ? or is just comptime zero rhs special)?

bfredl:
I guess the root of the question is under what circumstances the following code is okay.
```
test "@sizeOf(T) == 0 doesn't force resolving struct size" {
    const S = struct {
        const Foo = struct {
            y: if (@sizeOf(Foo) == 0) u64 else u32,
        };
    };
    try expect(@sizeOf(S.Foo) == 4);
}
```

What if `u64` or `u32` in the if/else was replaced with `void` ? what is the exact rule that says whenever this kind of self-reference is okay?

bfredl:
```
test "foobar" {
    const S = struct {
        const Foo = struct {
            y: if (@sizeOf(Foo) == 0) u64 else void,
        };
    };

    try expect(@sizeOf(S.Foo) == 0);
}
```
so this passes on master, but I expect it shouldn't?

test/behavior/sizeof_and_typeof.zig
```diff
@@ -52,6 +52,17 @@ test "lazy @sizeOf result is checked for definedness" {
...
 }
 
+const S2 = struct {
+    a: usize,
+};
+
+test "lazy @sizeOf can be compared with signed runtime value" {
+    var rhs: i32 = 100;
+    if (@sizeOf(S2) > rhs) {
+        @panic("bad");
```
Vexu, line 62: No panics in tests, use `expect`.

Vexu (changes requested):
Why is this specific to signed integers and lhs being a lazy value? Swapping the order or making the other type unsigned immediately breaks again.

ifreund (changes requested):
I believe the proper fix here would be to either make `intBitCountTwosComp()` take a "sema_kit" and use `toBigIntAdvanced()` or move it to `Sema.zig` and refactor it as I did in for a very similar case in bb1c3e8b7e2be201221e14719d2d39e6298cc66c.

bfredl:
> Why is this specific to signed integers and lhs being a lazy value? Swapping the order or making the other type unsigned immediately breaks again.

@Vexu  it was suggested this was about signedness when @andrewrk  made a reduced failure from my code in https://github.com/ziglang/zig/issues/12498. but given this can be problematic even for comparison with zero, perhaps it was a red herring?

I already made the change for rhs as well locally, but this is blocked on the much larger question: what is the exact rule that says whenever struct layout is allow to self-referentially depend on `@sizeOf(Foo)` which in turn obviously depends on the struct layout?  As I already pointed out the test suite suggest this should be allowed _sometimes_ but there is no easy rule I can infer from it. Once this has been decided, I could try to update the implementation accordingly, for both lhs and rhs.

Vexu:
> I already made the change for rhs as well locally, but this is blocked on the much larger question: what is the exact rule that says whenever struct layout is allow to self-referentially depend on `@sizeOf(Foo)` which in turn obviously depends on the struct layout? As I already pointed out the test suite suggest this should be allowed _sometimes_ but there is no easy rule I can infer from it. Once this has been decided, I could try to update the implementation accordingly, for both lhs and rhs.

I'm not sure I understand. If you mean when can we avoid resolving the layout then I think it is only with comparisons against zero.

bfredl:
I posted two examples above. https://github.com/ziglang/zig/pull/12539#issuecomment-1221423767 is currently in the test suite and looks fine in isolation. But also https://github.com/ziglang/zig/pull/12539#issuecomment-1221425691 passes and I don't think it should. I don't really see the rationale for allowing `@sizeOf(Foo) == 0` inside layout and how it could be done safely, while avoid contradictions like that.

If we allow to break the existing example in the test suite, fixing this issue is trivial. but there could be some rationale why it was added as an allowed escape hatch to start with?

Vexu:
I don't think either of those examples should pass. The original implementation of this in stage1 seems to have missed that issue 0512beca9d694a667e3ad12a656835b44457fbcd.

The second case in that commit could be made to work in stage2 but that would require implementing some sort of `resolveIsZeroBit`.

cc @andrewrk

Vexu:
Superseded by #13744

PR #12544, opened 2022-08-21
haiku update for futex thread

hoanga:
hello,

the following patch sets updates threading to use the SingleThreadImpl on haiku to prevent an error as shown below from occurring when trying to run zig build.

```
> mkdir helloworld

> cd helloworld

> zig init-exe
info: Created build.zig
info: Created src/main.zig
info: Next, try `zig build --help` or `zig build run`

> zig build run
/boot/home/src/git/zig/lib/std/Thread/Futex.zig:726:40: error: expected integer or vector, found 'comptime_int'
            return addr >> @ctz(usize, alignment);
                                       ^~~~~~~~~

> zig version
0.10.0-dev.3659+e5e6eb983

> uname -a
Haiku shredder 1 hrev56350 Aug 12 2022 06:04:04 x86_64 x86_64 Haiku
```

note that this patchsets also include

* linking the _bsd_ and _network_ libs when building on haiku (preliminary socket support) 
* expose dl_iterate_phdr (added to haiku as of [hrev56227)](https://git.haiku-os.org/haiku/log/?qt=range&q=hrev56226..hrev56227)

### DIFF
--- a/lib/std/Thread/Futex.zig
+++ b/lib/std/Thread/Futex.zig
@@ -73,6 +73,8 @@ else if (builtin.os.tag == .openbsd)
     OpenbsdImpl
 else if (builtin.os.tag == .dragonfly)
     DragonflyImpl
+else if (builtin.os.tag == .haiku)
+    SingleThreadedImpl
 else if (std.Thread.use_pthreads)
     PosixImpl
 else
--- a/lib/std/c/haiku.zig
+++ b/lib/std/c/haiku.zig
@@ -27,6 +27,10 @@ pub extern "c" fn _kern_read_stat(fd: c_int, path_ptr: [*]u8, traverse_link: boo
 
 pub extern "c" fn _kern_get_current_team() i32;
 
+pub extern "c" fn dl_iterate_phdr(callback: dl_iterate_phdr_callback, data: ?*anyopaque) c_int;
+
+pub const dl_iterate_phdr_callback = fn (info: *dl_phdr_info, size: usize, data: ?*anyopaque) callconv(.C) c_int;
+
 pub const sem_t = extern struct {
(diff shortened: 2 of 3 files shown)


andrewrk:
Hi @hoanga, can you please ask someone from the Haiku community to review your PR?

kprotty:
Switching to the `SingleThreadedImpl` on haiku when not `builtin.single_threaded` means that `std.Thread.Futex` and everything that uses it is no longer thread safe. The error looks more like a `stage2` bug with the `@ctz()` builtin not accepting compile time integers rather than a faulty `FutexImpl`.

andrewrk:
This patch along with your previous contributions demonstrate failure to do due diligence. It seems you are making haphazard changes without understanding what is going on. This kind of attitude is not welcome in the Zig project. Please do not send any more patches like this.

hoanga:
understood regarding this pr and the feedback received on wanting what is considered changes that would be considered less haphazard. the swap of single threaded on reflection was probably too hasty a change to request on my part.

i was attempting to get in touch with some other haiku developers for another look at this patch and hopefully give some other feedback on whether forcing haiku to use a single-threaded implementation was suggested or not as i also having some doubts if this approach was merited (probably not). at this point, i guess the feedback won't be necessary.

for better haiku support going forward, hopefully a more well versed haiku developer can figure it out. i will delete this branch to cut down on any confusion in the future and apologies if the proposed changes are causing trouble.

andrewrk:
Thank you for your response. Bugs with OS bits can easily cause exploitable Undefined Behavior and so such things need to be managed carefully. As a code reviewer, I need to be able to trust that someone has investigated carefully when they make changes to such areas of the codebase.

PR #12560, opened 2022-08-21
Add AFNOSUPPORT error to bind

schmee:
Happens if you do this:

```zig
    var socket = try std.os.socket(
        std.c.AF.LOCAL,
        std.c.SOCK.STREAM,
        0
    );

    const sockaddr = std.os.sockaddr.in{
        .port = std.mem.bigToNative(u16, 7777),
        .addr = @as(u32, 0),
    };
    try std.os.bind(socket, @ptrCast(*const std.os.sockaddr, &sockaddr), @sizeOf(@TypeOf(sockaddr)));
```

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -3447,6 +3447,9 @@ pub const BindError = error{
     /// A nonexistent interface was requested or the requested address was not local.
     AddressNotAvailable,
 
+    /// The address is not valid for the address family of socket.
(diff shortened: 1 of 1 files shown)


Vexu:
Should this return `error.AddressFamilyNotSupported` instead? Every other case of `.AFNOSUPPORT` seems to do so.

schmee:
@Vexu yes, fixed!

Issue #9629, opened 2021-08-26
make `usingnamespace` not put identifiers in scope


andrewrk:
From status quo the proposed change is very small:

 * Keyword is renamed from `usingnamespace` to `includenamespace`.
 * It only mixes in to the set of declarations for a given namespace; it does not affect identifiers.

Example:

```zig
pub const Foo = struct {
    includenamespace Bar;

    test {
        _ = hello; // error: use of undeclared identifier
        _ = Foo.hello; // OK!
    }
};

pub const Bar = struct {
    pub const hello = "hello";
}
```

Why? What material benefit does this have? Turns out, it's actually a big deal for the design of an incremental compiler:

 * AstGen is the place where we report shadowing of identifiers and unused identifiers. It would make sense to also report error for use of undeclared identifier. However the existence of `usingnamespace` makes that not an option. Having this error in AstGen means that the compile error is available non-lazily. That is, you will see errors of this form even for code that does not get semantically analyzed. For example, on Windows you would still see "use of undeclared identifier" errors when building for macOS.

 * With status quo semantics, an identifier in scope of more than one `usingnamespace` forces *all* of them to be resolved, in order to make sure there are no name conflicts. However, using `a.b` syntax means only the `includenamespace` declarations that apply to `a` must be resolved. With incremental compilation, having an identifier force an unrelated `usingnamespace` to be resolved creates a dependency between these two things. This dependency costs perf, memory, and decreases the amount of encapsulation of changes - that is, **it makes an incremental update less incremental**. The `includenamespace` semantics reduce the amount of such dependencies.

andrewrk:
I've been working on an implementation and the new AstGen error finds a lot of broken code in std:

```
/home/andy/dev/zig/lib/std/zig.zig:160:101: error: use of undeclared identifier 'suffix'
                return std.fmt.allocPrint(allocator, "{s}{s}{s}", .{ target.libPrefix(), root_name, suffix });
                                                                                                    ^
/home/andy/dev/zig/lib/std/hash_map.zig:612:29: error: use of undeclared identifier 'self'
            var other = try self.unmanaged.cloneContext(new_allocator, new_ctx);
                            ^
/home/andy/dev/zig/lib/std/json.zig:1409:17: error: use of undeclared identifier 'info'
            if (info.tag_type) |UnionTag| {
                ^
/home/andy/dev/zig/lib/std/fmt.zig:905:99: error: use of undeclared identifier 'value'
[...]

PR #12562, opened 2022-08-21
translate-c: Don't add self-defined macros to global name table

### LINKED ISSUE #12471: stage2: cImport failure with libuv (uv.h)
### Zig Version

0.10.0-dev.3587+9d85335de

### Steps to Reproduce

I'm trying to extract a minimal repo for you since the repo I have isn't open. But, its quite simple:

```zig
pub usingnamespace @cImport({
    @cInclude("uv.h");
});
```

libuv vendored here: https://github.com/libuv/libuv/tree/97dcdb1926f6aca43171e1614338bcef067abd59

### Expected Behavior

Compilation succeeds.

This compiles fine with stage1. 

### Actual Behavior

```
.../zig-cache/o/e37ac1d01edd5da25d95ea916308d281/cimport.zig:5615:30: error: use of undeclared identifier 'sched_priority'                                                                                                                                                              
pub const __sched_priority = sched_priority;                                                                                                                                                                                                                                                                                                 
                             ^~~~~~~~~~~~~~                                                                                                                                                                                                                                                                                                  
.../pkg/libuv/c.zig:1:20: error: C import failed: AnalysisFail                                                                                                                                                                                                                          
pub usingnamespace @cImport({     
```

(Note the "..." was created by me to omit the full path)

ehaas:
A self-defined macro is one of the form `#define FOO FOO`

Those types of macros have never been translated; this change will cause
any macros which refer to them to be translated as `@compileError` instead
of referring to a non-existent identifier.

Closes #12471

### DIFF
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -439,6 +439,24 @@ pub fn translate(
     return ast.render(gpa, context.global_scope.nodes.items);
 }
 
+/// Determines whether macro is of the form: `#define FOO FOO` (Possibly with trailing tokens)
+/// Macros of this form will not be translated.
+fn isSelfDefinedMacro(unit: *const clang.ASTUnit, c: *const Context, macro: *const clang.MacroDefinitionRecord) bool {
+    const source = getMacroText(unit, c, macro);
+    var tokenizer = std.c.Tokenizer{
+        .buffer = source,
+    };
+    const name_tok = tokenizer.next();
+    const name = source[name_tok.start..name_tok.end];
+
+    const first_tok = tokenizer.next();
+    // We do not just check for `.Identifier` below because keyword tokens are preferentially matched first by
+    // the tokenizer.
+    // In other words we would miss `#define inline inline` (`inline` is a valid c89 identifier)
+    if (first_tok.id == .Eof) return false;
+    return mem.eql(u8, name, source[first_tok.start..first_tok.end]);
+}
+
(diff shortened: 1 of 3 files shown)


mitchellh:
Did you ever figure out why this happens with stage2 and not stage1? (I’m not familiar with the translate-c machinery so maybe its obvious :))

ehaas:
It actually always produced the error for me on x64 Ubuntu (don't have an `aarch64` system to test with) regardless of stage1/2/3. There are some slight differences between stage1 and stage2 translate-c, but only with regard to function pointers; shouldn't be any difference in how those particular macros are translated.

Before this patch, does your system actually produce the `pub const __sched_priority = sched_priority;` line when running translate-c in stage1? If yes and it still works, my guess is some AstGen difference between stage1 and 2; if it does not produce the line then I would think there's some conditional compilation preventing the preprocessor from seeing it. The macros in my `/usr/include/sched.h` do not appear to be conditionally-included, though.

mitchellh:
Actually you're right, it is produced, so the stage2 difference is actually that AstGen or whatever semantic analysis cares that this unused `pub const` is erroneous. The `pub const` was always present and _always_ erroneous but stage2 doesn't like that it exists. :)

ehaas:
I overlooked the fact that your example included `usingnamespace` since I recognized the error text - I get the same behavior you do when I include that. I think this is the stage1/stage2 difference you're seeing: https://github.com/ziglang/zig/issues/9629#issuecomment-907429916

PR #12563, opened 2022-08-21
Stage2 fixes

Vexu:
~~The second commit does break `@clz`, `@ctz` etc with vectors for stage1 but hopefully we can get away with that?~~


andrewrk:
Unfortunately since Windows CI doesn't have enough RAM to build stage2 with stage1, we're using a binary blob to build stage3, meaning any breaking changes like this require creating a fresh CI tarball. I'm happy to help with this by doing two things:
 1. set you up with permissions to update the CI tarball
 2. work on making it more automated and thereby less painful to go through this process

Vexu:
I'll split the breaking commit to a separate PR.

PR #12567, opened 2022-08-22
std.zig.system.NativeTargetInfo: use getconf for detecting glibc version on glibc systems

### LINKED ISSUE #6469: zig detects wrong libc version

### LINKED ISSUE #11137: `-lc` doesn't link `c_nonshared`

BratishkaErik:
Closes #6469 and closes #11137
Signed-off-by: Eric Joldasov <bratishkaerik@getgoogleoff.me>

### DIFF
--- a/lib/std/zig/system/NativeTargetInfo.zig
+++ b/lib/std/zig/system/NativeTargetInfo.zig
@@ -328,19 +328,16 @@ fn detectAbiAndDynamicLinker(
         // Look for glibc version.
         var os_adjusted = os;
         if (builtin.target.os.tag == .linux and found_ld_info.abi.isGnu() and
-            cross_target.glibc_version == null)
+            cross_target.glibc_version == null and std.process.can_spawn)
         {
-            for (lib_paths) |lib_path| {
-                if (std.mem.endsWith(u8, lib_path, glibc_so_basename)) {
-                    os_adjusted.version_range.linux.glibc = glibcVerFromSO(lib_path) catch |err| switch (err) {
-                        error.UnrecognizedGnuLibCFileName => continue,
-                        error.InvalidGnuLibCVersion => continue,
-                        error.GnuLibCVersionUnavailable => continue,
-                        else => |e| return e,
-                    };
-                    break;
-                }
+            const getconf = std.ChildProcess.exec(.{ .allocator = allocator, .argv = &.{ "getconf", "GNU_LIBC_VERSION" } }) catch unreachable;
+            defer {
+                allocator.free(getconf.stderr);
+                allocator.free(getconf.stdout);
             }
+            const glibc_ver_from_stdout = getconf.stdout["glibc ".len..];
+            const glibc_version = std.builtin.Version.parse(glibc_ver_from_stdout) catch unreachable;
+            os_adjusted.version_range.linux.glibc = glibc_version;
         }
 
         var result: NativeTargetInfo = .{
@@ -428,12 +425,9 @@ fn detectAbiAndDynamicLinker(
 
     // If Zig is statically linked, such as via distributed binary static builds, the above
     // trick (block self_exe) won't work. The next thing we fall back to is the same thing, but for elf_file.
-    return abiAndDynamicLinkerFromFile(elf_file, cpu, os, ld_info_list, cross_target) catch |err| switch (err) {
+    return abiAndDynamicLinkerFromFile(allocator, elf_file, cpu, os, ld_info_list, cross_target) catch |err| switch (err) {
         error.FileSystem,
         error.SystemResources,
-        error.SymLinkLoop,
-        error.ProcessFdQuotaExceeded,
-        error.SystemFdQuotaExceeded,
         => |e| return e,
 
         error.UnableToReadElfFile,
@@ -453,50 +447,9 @@ fn detectAbiAndDynamicLinker(
     };
 }
 
-const glibc_so_basename = "libc.so.6";
-
-fn glibcVerFromSO(so_path: [:0]const u8) !std.builtin.Version {
-    var link_buf: [std.os.PATH_MAX]u8 = undefined;
-    const link_name = std.os.readlinkZ(so_path.ptr, &link_buf) catch |err| switch (err) {
-        error.AccessDenied => return error.GnuLibCVersionUnavailable,
-        error.FileSystem => return error.FileSystem,
-        error.SymLinkLoop => return error.SymLinkLoop,
-        error.NameTooLong => unreachable,
-        error.NotLink => return error.GnuLibCVersionUnavailable,
-        error.FileNotFound => return error.GnuLibCVersionUnavailable,
-        error.SystemResources => return error.SystemResources,
-        error.NotDir => return error.GnuLibCVersionUnavailable,
-        error.Unexpected => return error.GnuLibCVersionUnavailable,
-        error.InvalidUtf8 => unreachable, // Windows only
-        error.BadPathName => unreachable, // Windows only
-        error.UnsupportedReparsePointType => unreachable, // Windows only
-    };
-    return glibcVerFromLinkName(link_name, "libc-");
-}
-
-fn glibcVerFromLinkName(link_name: []const u8, prefix: []const u8) !std.builtin.Version {
-    // example: "libc-2.3.4.so"
-    // example: "libc-2.27.so"
-    // example: "ld-2.33.so"
-    const suffix = ".so";
-    if (!mem.startsWith(u8, link_name, prefix) or !mem.endsWith(u8, link_name, suffix)) {
-        return error.UnrecognizedGnuLibCFileName;
-    }
-    // chop off "libc-" and ".so"
-    const link_name_chopped = link_name[prefix.len .. link_name.len - suffix.len];
-    return std.builtin.Version.parse(link_name_chopped) catch |err| switch (err) {
-        error.Overflow => return error.InvalidGnuLibCVersion,
-        error.InvalidCharacter => return error.InvalidGnuLibCVersion,
-        error.InvalidVersion => return error.InvalidGnuLibCVersion,
-    };
-}
-
 pub const AbiAndDynamicLinkerFromFileError = error{
     FileSystem,
     SystemResources,
-    SymLinkLoop,
-    ProcessFdQuotaExceeded,
-    SystemFdQuotaExceeded,
     UnableToReadElfFile,
     InvalidElfClass,
     InvalidElfVersion,
@@ -509,6 +462,7 @@ pub const AbiAndDynamicLinkerFromFileError = error{
 };
 
 pub fn abiAndDynamicLinkerFromFile(
+    allocator: Allocator,
     file: fs.File,
     cpu: Target.Cpu,
     os: Target.Os,
@@ -546,7 +500,6 @@ pub fn abiAndDynamicLinkerFromFile(
         },
         .dynamic_linker = cross_target.dynamic_linker,
     };
-    var rpath_offset: ?u64 = null; // Found inside PT_DYNAMIC
     const look_for_ld = cross_target.dynamic_linker.get() == null;
 
     var ph_buf: [16 * @sizeOf(elf.Elf64_Phdr)]u8 align(@alignOf(elf.Elf64_Phdr)) = undefined;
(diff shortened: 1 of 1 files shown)


lib/std/zig/system/NativeTargetInfo.zig
```diff
@@ -641,167 +598,15 @@ pub fn abiAndDynamicLinkerFromFile(
...
-            };
-            result.target.os.version_range.linux.glibc = glibcVerFromLinkName(
-                fs.path.basename(link_name),
-                "ld-",
-            ) catch |err| switch (err) {
-                error.UnrecognizedGnuLibCFileName,
-                error.InvalidGnuLibCVersion,
-                => break :glibc_ver,
-            };
+        const getconf = std.ChildProcess.exec(.{ .allocator = allocator, .argv = &.{ "getconf", "GNU_LIBC_VERSION" } }) catch unreachable;
```
BratishkaErik, line 601: IIUC all glibc systems (even more, all POSIX systems https://pubs.opengroup.org/onlinepubs/000095399/utilities/getconf.html) has this program, and if not, it is a bug in systems itself.

lib/std/zig/system/NativeTargetInfo.zig
```diff
@@ -641,167 +598,15 @@ pub fn abiAndDynamicLinkerFromFile(
...
+        defer {
+            allocator.free(getconf.stderr);
+            allocator.free(getconf.stdout);
         }
+        const glibc_ver_from_stdout = getconf.stdout["glibc ".len..];
```
BratishkaErik, line 606: According to man:
> _CS_GNU_LIBC_VERSION (GNU C library only; since glibc 2.3.2)
              A string which identifies the GNU C library version on
              this system (e.g., "glibc 2.3.4").

lib/std/zig/system/NativeTargetInfo.zig
```diff
@@ -641,167 +598,15 @@ pub fn abiAndDynamicLinkerFromFile(
     if (builtin.target.os.tag == .linux and result.target.isGnuLibC() and
```
BratishkaErik, line 598: But anyway this will be not executed on a non-glibc systems

BratishkaErik:
@topolarity does this fixes https://github.com/ziglang/zig/issues/11137#issuecomment-1197276205 ?

BratishkaErik:
> > @topolarity does this fixes [#11137 (comment)](https://github.com/ziglang/zig/issues/11137#issuecomment-1197276205) and [#11137 (comment)](https://github.com/ziglang/zig/issues/11137#issue-1167334175) for you?
> 
> Oops, I forgot that zig doesn't provide glibc 2.35 yet :(

```zig
error: zig does not yet provide glibc version 2.35, the max provided version is 2.34
error: unable to build glibc shared objects: InvalidTargetGLibCVersion
error: zig...
error: The following command exited with error code 1:
/home/bratishkaerik/zig/build/zig build-exe --stack 33554432 /home/bratishkaerik/zig/src/main.zig -lc /home/bratishkaerik/zig/build/zigcpp/libzigcpp.a /usr/lib/llvm/14/lib64/libclang-cpp.so.14 /usr/lib64/liblldMinGW.so /usr/lib64/liblldELF.so /usr/lib64/liblldCOFF.so /usr/lib64/liblldWasm.so /usr/lib64/liblldMachO.so /usr/lib64/liblldCommon.so /usr/lib/llvm/14/lib64/libLLVM-14.so /usr/lib/gcc/x86_64-pc-linux-gnu/11.3.0/libstdc++.so -lunwind --cache-dir /home/bratishkaerik/zig/zig-cache --global-cache-dir /home/bratishkaerik/.cache/zig --name zig --pkg-begin build_options /home/bratishkaerik/zig/zig-cache/options/7X62S_IHrJHJyONEQXfH1Bt-Y-ZrGbNtZVv-QoWvD9uq09Hl8g4BY__8D6284ocN --pkg-end -I /usr/include -fno-build-id --enable-cache 
/home/bratishkaerik/zig/zig-cache/o/87897b425d0e88a403c948b1ad22e5dc/build /home/bratishkaerik/zig/build/zig /home/bratishkaerik/zig /home/bratishkaerik/zig/zig-cache /home/bratishkaerik/.cache/zig -p stage2 -Dstatic-llvm=false -Denable-llvm=true

**Comment by BratishkaErik** [comment]
My suggestion:
```diff
diff --git a/src/glibc.zig b/src/glibc.zig
index ceb60ff09..d714d4aa1 100644
--- a/src/glibc.zig
+++ b/src/glibc.zig
@@ -745,14 +745,7 @@ pub fn buildSharedObjects(comp: *Compilation) !void {
                     return error.InvalidTargetGLibCVersion;
                 },
             }
-        } else {
-            const latest_index = metadata.all_versions.len - 1;
-            // TODO Expose via compile error mechanism instead of log.
-            log.err("zig does not yet provide glibc version {}, the max provided version is {}", .{
-                target_version, metadata.all_versions[latest_index],
-            });
-            return error.InvalidTargetGLibCVersion;
-        };
+        } else metadata.all_versions.len - 1;
 
         {
             var map_contents = std.ArrayList(u8).init(arena);
```

topolarity:
> @topolarity does this fixes [#11137 (comment)](https://github.com/ziglang/zig/issues/11137#issuecomment-1197276205) and [#11137 (comment)](https://github.com/ziglang/zig/issues/11137#issue-1167334175) for you?

Looks like if Zig dynamically links libc (e.g. `-Dstatic-llvm=false`), then this is still unresolved

topolarity:
> > @topolarity does this fixes [#11137 (comment)](https://github.com/ziglang/zig/issues/11137#issuecomment-1197276205) and [#11137 (comment)](https://github.com/ziglang/zig/issues/11137#issue-1167334175) for you?
> 
> Oops, I forgot that zig doesn't provide glibc 2.35 yet :(

IIUC we should still be able to use the system glibc in this case, right?

BratishkaErik:
> IIUC we should still be able to use the system glibc in this case, right?

I'm hitting this https://github.com/ziglang/zig/pull/12567#issuecomment-1224770681

topolarity:
> I force–pushed recently, can you try this version too? (and I'm using dynamic linking, too) ea7cb4c

Yep, I was out of date! This appears to be resolved for me :+1:

Re: the glibc version. Zig will currently fallback to using system-provided libc if you ask for system libraries (e.g. if you add `-lz` to the build-exe invocation, it will work). Perhaps it's worth performing the same fallback if Zig does not provide the requested version of libc but the system does?

BratishkaErik:
> > I force–pushed recently, can you try this version too? (and I'm using dynamic linking, too) ea7cb4c
> 
> Yep, I was out of date! This appears to be resolved for me :+1:
> 
> Re: the glibc version. Zig will currently fallback to using system-provided libc if you ask for system libraries (e.g. if you add `-lz` to the build-exe invocation, it will work). Perhaps it's worth performing the same fallback if Zig does not provide the requested version of libc but the system does?

Something like?:
[if glibc newer than supplied] —> (use system libc) —> [if system doesn't provide headers] —> (use latest supplied libc available)

topolarity (approved):
Tested on Debian (bookworm), NixOS (unstable), and Arch - all of which are affected by #6469 and #11137 and are fixed by this PR.

This change does mean linking libc will be broken on OS's with too-new libc (e.g. NixOS and Arch) until we have a fallback to link against system libc or update Zig-provided glibc, but this still seems like a step in the right direction even without that

P.S. Not sure what the CI failures are, but they seem unrelated?

topolarity:
> Something like?:
> [if glibc newer than supplied] —> (use system libc) —> [if system doesn't provide headers] —> (use latest supplied libc available)

Sounds OK to me - most of the logic should already be present in Compilation.zig. Might be as simple as updating `canBuildLibC` to account for the libc version

If you're up to tackling that, feel free to update this PR or start a new one - whichever you prefer 🙂

BratishkaErik:
—> (use latest supplied libc available) ❌

https://github.com/ziglang/zig/pull/8737/commits/01e30002c52eda1ab3574f55c2e0568c56f2ed09
> stage2: error if requested glibc version too high 
> Falling back to the max provided glibc version is insufficient as linking to shared objects compiled against the requested version will fail.

topolarity:
Ah, I misunderstood and thought you had meant use the Zig headers to link against the system-provided glibc.

Good catch, let's stay faithful to the requested glibc version

andrewrk:
I created #12788 which supersedes this PR and does not depend on `getconf`.

PR #12575, opened 2022-08-22
coff: improve default COFF/PE object parser

kubkon:
We now do not allocate memory for headers and other metadata unless
requested by the caller. Instead, we read-in the entire contents
of the image into memory and operate on pointers and casts wherever
possible. I have a left a TODO to hook up Windows' memory-mapped API
here in-place of standard `readToEndAlloc` which should be more memory
proof on memory constrained hosts.

This commit also supplements our `std.coff` with a lot missing basic
extern structs required to make our COFF linker.

### DIFF
--- a/lib/std/coff.zig
+++ b/lib/std/coff.zig
@@ -1,14 +1,731 @@
 const std = @import("std.zig");
+const assert = std.debug.assert;
 const io = std.io;
 const mem = std.mem;
 const os = std.os;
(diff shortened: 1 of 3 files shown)


kubkon:
The failure on Windows looks unrelated especially considering that I run the tests locally on my Windows install and they all passed.

andrewrk (approved):

PR #12586, opened 2022-08-23
Adds std.heap.MemoryPool

MasterQ32:
This is a fast memory pool implementation that can be used to rapidly allocate objects of the same type. It even outperforms the null hypothesis of just leaking the memory using a `std.heap.ArenaAllocator`.

### DIFF
--- a/lib/std/heap/memory_pool.zig
+++ b/lib/std/heap/memory_pool.zig
@@ -5,15 +5,15 @@ const debug_mode = @import("builtin").mode == .Debug;
 pub const MemoryPoolError = error{OutOfMemory};
 
 /// A memory pool that can allocate objects of a single type very quickly.
-/// Use this instead of a general purpose allocator when you need to allocate a lot of objects
-/// of the same type, as a memory pool outperforms general purpose allocators.
+/// Use this when you need to allocate a lot of objects of the same type,
+/// because It outperforms general purpose allocators.
 pub fn MemoryPool(comptime Item: type) type {
     return MemoryPoolAligned(Item, @alignOf(Item));
 }
 
 /// A memory pool that can allocate objects of a single type very quickly.
-/// Use this instead of a general purpose allocator when you need to allocate a lot of objects
-/// of the same type, as a memory pool outperforms general purpose allocators.
+/// Use this when you need to allocate a lot of objects of the same type,
+/// because It outperforms general purpose allocators.
 pub fn MemoryPoolAligned(comptime Item: type, comptime alignment: u29) type {
     if (@alignOf(Item) == alignment) {
         return MemoryPoolExtra(Item, .{});
@@ -32,20 +32,18 @@ pub const Options = struct {
 };
 
 /// A memory pool that can allocate objects of a single type very quickly.
-/// Use this instead of a general purpose allocator when you need to allocate a lot of objects
-/// of the same type, as a memory pool outperforms general purpose allocators.
+/// Use this when you need to allocate a lot of objects of the same type,
+/// because It outperforms general purpose allocators.
 pub fn MemoryPoolExtra(comptime Item: type, comptime pool_options: Options) type {
     return struct {
         const Pool = @This();
 
-        /// Size of the memory pool items. This is not necessarily
-        /// the same as `@sizeOf(Item)` as the pool also uses the items for internal
-        /// means.
+        /// Size of the memory pool items. This is not necessarily the same
+        /// as `@sizeOf(Item)` as the pool also uses the items for internal means.
         pub const item_size = std.math.max(@sizeOf(Node), @sizeOf(Item));
 
-        /// Alignment of the memory pool items. This is not necessarily
-        /// the same as `@alignOf(Item)` as the pool also uses the items for internal
-        /// means.
+        /// Alignment of the memory pool items. This is not necessarily the same
+        /// as `@alignOf(Item)` as the pool also uses the items for internal means.
         pub const item_alignment = std.math.max(@alignOf(Node), pool_options.alignment orelse 0);
 
         const Node = struct {
--- a/lib/std/heap/memory_pool.zig
+++ b/lib/std/heap/memory_pool.zig
@@ -1,0 +1,183 @@
+const std = @import("../std.zig");
+
+const debug_mode = @import("builtin").mode == .Debug;
+
+pub const MemoryPoolError = error{OutOfMemory};
+
+/// A memory pool that can allocate objects of a single type very quickly.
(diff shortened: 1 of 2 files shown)


lib/std/heap/memory_pool.zig
```diff
@@ -0,0 +1,183 @@
+const std = @import("../std.zig");
+
+const debug_mode = @import("builtin").mode == .Debug;
+
+pub const MemoryPoolError = error{OutOfMemory};
+
+/// A memory pool that can allocate objects of a single type very quickly.
+/// Use this instead of a general purpose allocator when you need to allocate a lot of objects
```
matu3ba, line 8: s/Use this instead of a general purpose allocator when you need to allocate a lot of objects,/Use this when you need to allocate a lot of objects of the same type

lib/std/heap/memory_pool.zig
```diff
@@ -0,0 +1,183 @@
...
+/// of the same type, as a memory pool outperforms general purpose allocators.
```
matu3ba, line 9: ~~`/// of the same type, as a memory pool outperforms general purpose allocators.`~~
because It outperforms general purpose allocators.

lib/std/heap/memory_pool.zig
```diff
@@ -0,0 +1,183 @@
...
+};
+
+/// A memory pool that can allocate objects of a single type very quickly.
+/// Use this instead of a general purpose allocator when you need to allocate a lot of objects
+/// of the same type, as a memory pool outperforms general purpose allocators.
+pub fn MemoryPoolExtra(comptime Item: type, comptime pool_options: Options) type {
+    return struct {
+        const Pool = @This();
+
+        /// Size of the memory pool items. This is not necessarily
```
matu3ba, line 41:
```zig
/// Size of the memory pool items. This is not necessarily the same 
/// as `@sizeOf(Item)` as the pool also uses the items for internal means.
```

lib/std/heap/memory_pool.zig
```diff
@@ -0,0 +1,183 @@
...
+pub const MemoryPoolError = error{OutOfMemory};
+
+/// A memory pool that can allocate objects of a single type very quickly.
+/// Use this instead of a general purpose allocator when you need to allocate a lot of objects
+/// of the same type, as a memory pool outperforms general purpose allocators.
+pub fn MemoryPool(comptime Item: type) type {
+    return MemoryPoolAligned(Item, @alignOf(Item));
+}
+
+/// A memory pool that can allocate objects of a single type very quickly.
```
matu3ba, line 14: see above.

matu3ba (commented):
Comments could be abit more dense + less duplicated.

Should the perf difference to only the arena allocator also be briefly mentioned?

lib/std/heap/memory_pool.zig
```diff
@@ -0,0 +1,183 @@
...
+    alignment: ?u29 = null,
+
+    /// If `true`, the memory pool can allocate additional items after a initial setup.
+    /// If `false`, the memory pool will not allocate further after a call to `initPreheated`.
+    growable: bool = true,
+};
+
+/// A memory pool that can allocate objects of a single type very quickly.
+/// Use this instead of a general purpose allocator when you need to allocate a lot of objects
+/// of the same type, as a memory pool outperforms general purpose allocators.
```
hawkinsw: Should the doc comments be somehow different between this and `MemoryPoolAligned` to indicate that they are different?

hawkinsw (commented):
This is awesome functionality. Thank you for contributing. Not sure what my thought is worth, but I included it below. Thanks again for the submission and I hope it gets accepted!

lib/std/heap/memory_pool.zig
MasterQ32: Improved the doc comments for *Aligned and *Extra a bit

lib/std/heap/memory_pool.zig
```diff
@@ -32,6 +35,10 @@ pub const Options = struct {
 };
 
 /// A memory pool that can allocate objects of a single type very quickly.
+/// This implementation allows both specifying a user-defined item alignment
+/// and the option to prevent future allocations after the initialization.
+/// This is especially useful in memory-constraint applications.
```
hawkinsw, line 40: Just trying to be helpful, so please tell me if I am not:

"memory-constraint" => "memory-constrained"

Jarred-Sumner:
nice!

lib/std/heap/memory_pool.zig
```diff
@@ -0,0 +1,188 @@
...
+pub const Options = struct {
+    /// The alignment of the memory pool items. Use `null` for natural alignment.
+    alignment: ?u29 = null,
+
+    /// If `true`, the memory pool can allocate additional items after a initial setup.
+    /// If `false`, the memory pool will not allocate further after a call to `initPreheated`.
+    growable: bool = true,
+};
+
+/// A memory pool tht can allocate objects of a single type very quickly.
```
leecannon, line 37: tht -> that

lib/std/heap/memory_pool.zig
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+pub const Options = struct {
+    /// The alignment of the memory pool items. Use `null` for natural alignment.
+    alignment: ?u29 = null,
+
+    /// If `true`, the memory pool can allocate additional items after a initial setup.
+    /// If `false`, the memory pool will not allocate further after a call to `initPreheated`.
+    growable: bool = true,
```
marler8997, line 34: This can already be configured by changing whether the underlying allocator is growable or not.

MasterQ32: Not really, as the internal arena will allocate in larger chunks and not a specific amount of items. But i guess there's some space for improvements, true.

marler8997: Oh I see, I thought the underlying allocator was just a generic Allocator. Carry on.

MasterQ32: It uses a generic allocator, but has an internal arena to have a simpler way of freeing the allocated memory

BanchouBoo:
> TODO: Potentially store all allocated objects in a list as well, allowing to just move them into the free list instead of actually releasing the memory.

Unless there's some drawback I'm not seeing, including a list of allocated items would also be nice for general storage/iteration purposes since one of the benefits you get from an arena is cache locality, and there's a decent chance that if you're using a pool for a particular type that you're also going to want to access those items sequentially rather than randomly, so having that storage would remove an extra step for the user.

If there would be drawbacks to doing this (I suppose ensuring everything is kept in order might be tricky?) then perhaps a secondary `StoredMemoryPool` type might also be a good idea.

Issue #12547, opened 2022-08-21
Packed structs can't contain arrays


matrizzo:
### Zig Version

0.10.0-dev.3659+e5e6eb983

### Steps to Reproduce

```
matteo@lab:/tmp$ zig version
0.10.0-dev.3659+e5e6eb983
matteo@lab:/tmp$ cat test.zig
const std = @import("std");

const TestStruct = packed struct {
    pad: [512]u8,
    x: u64,
};

pub fn main() void {
    const asdf = TestStruct{.pad = [_]u8{0} ** 512, .x = 1};
    std.debug.print("{d}", .{asdf.x});
}
matteo@lab:/tmp$ zig build-exe test.zig
test.zig:4:5: error: packed structs cannot contain fields of type '[512]u8'
    pad: [512]u8,
    ^~~~~~~~~~~~
test.zig:4:5: note: type has no guaranteed in-memory representation
matteo@lab:/tmp$ zig build-exe -fstage1 test.zig
matteo@lab:/tmp$
```

### Expected Behavior

The example compiles without errors with both `-fstage1` and with the default settings.

### Actual Behavior

The compiler complains that `[512]u8` can't be in a packed struct because "it has no guaranteed in-memory representation". This doesn't really make sense to me, isn't an u8 array supposed to be just a fixed-size sequence of bytes?

When using the `-fstage1` the compiler doesn't complain.

Vexu:
Working as intended, see https://github.com/ziglang/zig/issues/10113#issuecomment-1036713233 and the rest of the thread for reasoning.

The error should explain it better though.

ominitay:
The documentation regarding packed vs extern structs is poor in my opinion. I'll get around to improving it at some point, since it does appear to mislead people a lot.

matrizzo:
Is there any way to have something that works like a packed struct (i.e., guaranteed memory layout, no padding, can be used to reinterpret memory, ...) and can contain arrays? This seems like it should be a fairly common use case when decoding some binary data.

Also, I'd argue that this is still a bug since the stage1 and stage2 compilers produce different results :)

ominitay:
It's a bug in stage1, which won't be fixed, at least any time soon. Extern structs have a known memory layout and should be used in most cases, but where bit-packing is required, there is only the choice of packed structs, which can't contain arrays. There are a few discussions going on about solutions to this, but not sure how far they'll get.

matrizzo:
[...]

PR #12587, opened 2022-08-23
openbsd: fix sigcontext struct and avoid defining fxsave64

semarie:
`sc_fpstate` member of `struct sigcontext` is a [`struct fxsave64 *`](https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/sys/arch/amd64/include/signal.h#L83).
use `*anyopaque` to represent it.

avoid to defining `fxsave64` as it is a [packed struct with some arrays](https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/sys/arch/amd64/include/fpu.h#L17-L30) and zig doesn't compile with it (see #12547)

### DIFF
--- a/lib/std/c/openbsd.zig
+++ b/lib/std/c/openbsd.zig
@@ -1119,25 +1119,10 @@ pub usingnamespace switch (builtin.cpu.arch) {
             sc_rsp: c_long,
             sc_ss: c_long,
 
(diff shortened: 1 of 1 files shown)


ifreund (approved):
Thanks!

PR #12589, opened 2022-08-23
wasm-lld: set stack size to 1MB by default

### LINKED ISSUE #3735: wasm has a max stack size of ~66K
Easiest to reproduce with the [WASI test patches](https://github.com/ziglang/zig/pull/3730):

`$ zig test wasi.zig -target wasm32-wasi --test-cmd wasmtime --test-cmd-bin`

```zig
const std = @import("std");

test "safe" {
    var buf: [64000]u8 = undefined;
    std.debug.warn("{*}\n", &buf);
}

test "broken in debug" {
    var buf: [65000]u8 = undefined;
    std.debug.warn("{*}\n", &buf);
}

test "broken in release-safe" {
    var buf: [66000]u8 = undefined;
    std.debug.warn("{*}\n", &buf);
}

test "broken in release-fast" {
    var buf: [68000]u8 = undefined;
    std.debug.warn("{*}\n", &buf);
}
```

Luukdegram:
Regardless of the build mode (build-exe, build-lib), always set the default stack size to 1MB. Previously, this was only done when using build-exe, making the inconsistancy confusing. The user can still override this behavior by providing the `--stack <size>` flag.

Closes #3735.

### DIFF
--- a/src/link/Wasm.zig
+++ b/src/link/Wasm.zig
@@ -2836,24 +2836,19 @@ fn linkWithLLD(self: *Wasm, comp: *Compilation, prog_node: *std.Progress.Node) !
             try argv.append(entry);
         }
 
-        if (self.base.options.output_mode == .Exe) {
-            // Increase the default stack size to a more reasonable value of 1MB instead of
-            // the default of 1 Wasm page being 64KB, unless overridden by the user.
-            try argv.append("-z");
-            const stack_size = self.base.options.stack_size_override orelse 1048576;
-            const arg = try std.fmt.allocPrint(arena, "stack-size={d}", .{stack_size});
-            try argv.append(arg);
+        // Increase the default stack size to a more reasonable value of 1MB instead of
+        // the default of 1 Wasm page being 64KB, unless overridden by the user.
+        try argv.append("-z");
+        const stack_size = self.base.options.stack_size_override orelse wasm.page_size * 16;
+        const arg = try std.fmt.allocPrint(arena, "stack-size={d}", .{stack_size});
+        try argv.append(arg);
 
+        if (self.base.options.output_mode == .Exe) {
             if (self.base.options.wasi_exec_model == .reactor) {
                 // Reactor execution model does not have _start so lld doesn't look for it.
                 try argv.append("--no-entry");
             }
-        } else {
-            if (self.base.options.stack_size_override) |stack_size| {
-                try argv.append("-z");
-                const arg = try std.fmt.allocPrint(arena, "stack-size={d}", .{stack_size});
-                try argv.append(arg);
-            }
+        } else if (self.base.options.entry == null) {
             try argv.append("--no-entry"); // So lld doesn't look for _start.
         }
         try argv.appendSlice(&[_][]const u8{


src/link/Wasm.zig
```diff
@@ -2836,24 +2836,19 @@ fn linkWithLLD(self: *Wasm, comp: *Compilation, prog_node: *std.Progress.Node) !
...
-            // Increase the default stack size to a more reasonable value of 1MB instead of
-            // the default of 1 Wasm page being 64KB, unless overridden by the user.
-            try argv.append("-z");
-            const stack_size = self.base.options.stack_size_override orelse 1048576;
-            const arg = try std.fmt.allocPrint(arena, "stack-size={d}", .{stack_size});
-            try argv.append(arg);
+        // Increase the default stack size to a more reasonable value of 1MB instead of
+        // the default of 1 Wasm page being 64KB, unless overridden by the user.
+        try argv.append("-z");
+        const stack_size = self.base.options.stack_size_override orelse 1048576;
```
r00ster91, line 2842: Very small nit but can we make this cryptic number easier to read?
```suggestion
        const stack_size = self.base.options.stack_size_override orelse 1 * 1024 * 1024;
```
It also makes it easier to understand this is not 1 MB but actually 1 MiB.

src/link/Wasm.zig
```diff
@@ -2836,24 +2836,19 @@ fn linkWithLLD(self: *Wasm, comp: *Compilation, prog_node: *std.Progress.Node) !
...
                 // Reactor execution model does not have _start so lld doesn't look for it.
                 try argv.append("--no-entry");
             }
-        } else {
-            if (self.base.options.stack_size_override) |stack_size| {
-                try argv.append("-z");
-                const arg = try std.fmt.allocPrint(arena, "stack-size={d}", .{stack_size});
-                try argv.append(arg);
-            }
+        } else if (self.base.options.entry == null) {
```
kubkon, line 2851: Are you sure this is correct here? The previous version of the code did not check for `entry` at all.

Luukdegram: Yes, but a little bit above we set the entry if the user provided it. It's invalid to pass both an entry point and `--no-entry`. I figured I'd get that fix in while at it.

kubkon: If it's invalid, should we perhaps flag it up as an error to the user?

Luukdegram: I don't think so. `--entry` is given by the user, whereas `--no-entry` is a flag we set by default depending on the output mode. So the user never creates this invalid state, it's the Zig toolchain that should ensure it doesn't occur.

kubkon: Ah right, so we don't expose `--no-entry` to the user? I'm curious though, what is the behavior of LLD upon receiving both flags in a single linker line, do you know perhaps?

Luukdegram: LLD will ignore the `--entry` flag in such case, which may confuse the user, as they may be unaware that `--no-entry‘ was passed to lld also.

kubkon: Thanks for double checking! I am fine with the current approach. In the future though, I wonder if we should perhaps make it a hard error. Food for thought!

kubkon (approved):

PR #12590, opened 2022-08-23
Implements std.ArenaAllocator.reset()

MasterQ32:
Implements a method to reset the arena allocator and provider a variant that allocates less in the next iteration of the arena use. This is especially useful for an application that allocates temporaries in a loop.

### DIFF
--- a/lib/std/heap/arena_allocator.zig
+++ b/lib/std/heap/arena_allocator.zig
@@ -55,24 +55,20 @@ pub const ArenaAllocator = struct {
     pub const ResetMode = union(enum) {
         /// Releases all allocated memory in the arena.
         free_all,
-
         /// This will pre-heat the arena for future allocations by allocating a
         /// large enough buffer for all previously done allocations.
         /// Preheating will speed up the allocation process by invoking the backing allocator
         /// less often than before. If `reset()` is used in a loop, this means that after the
         /// biggest operation, no memory allocations are performed anymore.
         retain_capacity,
-
         /// This is the same as `retain_capacity`, but the memory will be shrunk to
         /// this value if it exceeds the limit.
         retain_with_limit: usize,
     };
-
     /// Queries the current memory use of this arena.
     /// This will **not** include the storage required for internal keeping.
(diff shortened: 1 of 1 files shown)


tauoverpi:
Wondering if it would be nicer to split this into two methods instead of using a bool with `clearRetainingCapacity()` or similar in addition to `reset()` to communicate intent. That or an enum instead of the bool to make the choice explicit `enum { free, preheat }`.

MasterQ32:
> Wondering if it would be nicer to split this into two methods instead of using a bool with `clearRetainingCapacity()` or similar in addition to `reset()` to communicate intent. That or an enum instead of the bool to make the choice explicit `enum { free, preheat }`.

This is a good idea. I'll incorporate it and improve the code in that regard.

lib/std/heap/arena_allocator.zig
```diff
@@ -43,6 +45,128 @@ pub const ArenaAllocator = struct {
...
+
+        /// This will pre-heat the arena for future allocations by allocating a
+        /// large enough buffer for all previously done allocations.
+        /// Preheating will speed up the allocation process by invoking the backing allocator
+        /// less often than before. If `reset()` is used in a loop, this means that after the
+        /// biggest operation, no memory allocations are performed anymore.
+        retain_capacity,
+
+        /// This is the same as `retain_capacity`, but the memory will be shrunk to
+        /// the this value if it exceeds the limit.
```
r00ster91, line 60:
```suggestion
        /// this value if it exceeds the limit.
```

jcmoyer:
I've gone ahead and ported this change set to current zig master, feel free to steal it or I can open a new PR if that would be easier. Been needing this feature for a couple projects. https://github.com/jcmoyer/puzzles/blob/5b08701a364cf455b2bd0c2e3fcc74aa58015a22/AdventOfCode2022/src/resettable_arena_allocator.zig

MasterQ32:
Did a "manual rebase" (aka: create a new branch from master and copy-paste my changes) as i fucked up something in my branch history

PR #12600, opened 2022-08-23
autodocs: improve error set rendering

ominitay:
This change allows autodoc to render empty error sets (this previously caused a runtime error), and to correctly render multi-item error sets in the same style as `zig fmt`. 

Before:
- empty `[crashes]`
- single `error{Foo}`
- multi `error{Foo, Bar}`

After:
- empty `error{}`
- single `error{Foo}`
- multi `error{ Foo, Bar }`


kristoff-it:
Thanks!

PR #12601, opened 2022-08-23
autodocs: improve first-line descriptions

ominitay:
This change stops autodoc from cutting off doc comments at the first full stop for a short description, and instead will cut off doc comments at first paragraph, or after 80 characters, whichever is sooner.

This prevents short summaries of multiple sentences from being cut early, and prevents examples without a second paragraph from creating noise.


kristoff-it:
Thank you!

PR #12605, opened 2022-08-23
compilation: avoid pointless caching

### LINKED ISSUE #12317: Possibility to disable caching for user code
### Zig Version

0.10.0-dev.2783+76546b3f8

### Steps to Reproduce

Build a very large binary with `zig cc` as the C toolchain. In our case, the binary is 584M. 

### Expected Behavior

Currently it is expected that the Zig cache may grow very large. However, when using an external build system to orchestrate the build and cache the results (in our case Bazel), we have no need for the Zig cache outside "internal" builds (libc stubs etc.). It would be great to have a flag to only cache these internal builds, and not user code.

### Actual Behavior

With our 584M binary, the Zig cache is 761M after building from a clean cache.

kristoff-it:
When the entire purpose of this compilation is to perform a single zig
cc operation we could "tail call" clang by doing an execve, and any use 
of the caching system would actually be problematic since the user is 
presumably doing their own caching by using dep file flags.

Fixes #12317

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -3753,22 +3753,64 @@ fn updateCObject(comp: *Compilation, c_object: *CObject, c_obj_prog_node: *std.P
         };
         const o_basename = try std.fmt.allocPrint(arena, "{s}{s}", .{ o_basename_noext, out_ext });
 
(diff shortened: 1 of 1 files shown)


andrewrk:
Tests passed locally.

PR #12606, opened 2022-08-23
docs: remove confusion about align[For|Back]ward

jgollenz:
- `alignBackward` does round 'down', not 'up'
- removed 'nearest' in favor of 'next' as its more in line with 'previous'
- added '(or current)', because the input address will be the output address if its already aligned.
- added a period at the end of the lines I touched, because it is missing when the stdlib docs are generated

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -3252,13 +3252,13 @@ test "sliceAsBytes preserves pointer attributes" {
     try testing.expectEqual(in.alignment, out.alignment);
 }
 
-/// Round an address up to the nearest aligned address
+/// Round an address up to the next (or current) aligned address.
 /// The alignment must be a power of 2 and greater than 0.
 pub fn alignForward(addr: usize, alignment: usize) usize {
     return alignForwardGeneric(usize, addr, alignment);
 }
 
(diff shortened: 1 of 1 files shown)


lib/std/mem.zig
```diff
@@ -3252,13 +3252,13 @@ test "sliceAsBytes preserves pointer attributes" {
     try testing.expectEqual(in.alignment, out.alignment);
 }
 
-/// Round an address up to the nearest aligned address
+/// Round an unaligned address up to the next aligned address.
```
N00byEdge, line 3255: `[...] next (or current) aligned address`

lib/std/mem.zig
```diff
@@ -3290,7 +3290,7 @@ test "alignForward" {
     try testing.expect(alignForward(17, 8) == 24);
 }
 
-/// Round an address up to the previous aligned address
+/// Round an unaligned address down to the previous aligned address.
```
N00byEdge, line 3293: `[...] previous (or current) aligned address`

N00byEdge (commented):
These of course also apply in the other places where the terminology is used

PR #12631, opened 2022-08-25
add ability to pass force-undefined-symbols to the linker

### LINKED ISSUE #8531: LLD with LTO enabled incorrectly drops _tls_index symbol, causing undefined reference errors

kubkon:
This commit enables `-u <symbol>` for ELF and `-include:<symbol>` for COFF linkers for use internally. This means we do not expose these flags to the users just yet, however, we make use of them internally whenever required. One such use case is forcing inclusion of `_tls_index` when linking for Windows with mingw and LTO and dead code stripping enabled. This ensures we add `_tls_index` to the symbol resolver as an undefined symbol and force the linker to include an atom that provides it marking it a dead-code-stripping root - meaning it will not be garbage collected by the linker no matter what.

This is my proposed alternative to #8674 which doesn't rely on scanning input source files by name when compiling mingw and putting in the cache. Instead, we tell the linker to force-include `_tls_index` whenever we link against mingw libc.

Closes #8531

@andrewrk I've also had a look at your reported C++ problems in presence of LTO but I couldn't repro it locally so unless we get a report that it's still the case, I have now enabled LTO by default on Windows for Release builds, and I am marking #8531 for closing.

cc @xxxbxxx

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -1165,9 +1165,6 @@ pub fn create(gpa: Allocator, options: InitOptions) !*Compilation {
                 break :blk false;
             } else if (options.c_source_files.len == 0) {
                 break :blk false;
-            } else if (options.target.os.tag == .windows and link_libcpp) {
-                // https://github.com/ziglang/zig/issues/8531
-                break :blk false;
             } else if (options.target.cpu.arch.isRISCV()) {
                 // Clang and LLVM currently don't support RISC-V target-abi for LTO.
                 // Compiling with LTO may fail or produce undesired results.
@@ -1793,6 +1790,7 @@ pub fn create(gpa: Allocator, options: InitOptions) !*Compilation {
             .headerpad_size = options.headerpad_size,
             .headerpad_max_install_names = options.headerpad_max_install_names,
             .dead_strip_dylibs = options.dead_strip_dylibs,
+            .force_undefined_symbols = .{},
         });
         errdefer bin_file.destroy();
         comp.* = .{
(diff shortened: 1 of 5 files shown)


rtldg:
I feel bad about posting on this since I'm using Zig as a cross-compiler for Rust, but this seems to be causing my builds to fail with `lld-link: error: <root>: undefined symbol: _tls_index`. Defining the variable `_tls_index` instead causes things like this
```
lld-link: error: duplicate symbol: __tls_index
>>> defined at testcpp.lib(test.o)
>>> defined at mingw32.lib(tlssup.obj)
```
One of the old zig versions I was using did compile & link my projects fine so this pr seems to be where the `undefined symbol` stuff started.
(`zig-windows-x86_64-0.10.0-dev.3602+df507edff` from `2022-08-18`)

Also an [example project of this](https://github.com/rtldg/rust_and_zig_crossbuild) where I'm using [cargo-zigbuild](https://github.com/messense/cargo-zigbuild) which acts as a wrapper for cross-compiling along with the `i686-pc-windows-gnu` toolchain for Rust.

Just wondering if I should post this elsewhere for help.

kubkon:
Thanks for bringing this to my attention @rtldg! I'll be happy to investigate what is happening and hopefully provide a fix :-)

PR #12632, opened 2022-08-25
Add behavior test coverage for nested packed struct field access

### LINKED ISSUE #3091: Nested property access of packed struct loads wrong value
Accessing nested properties of packed structs seems to silently load from the wrong offset within the struct, without a compile error or runtime crash.

```zig
const std = @import("std");
const warn = std.debug.warn;

const Foo = packed struct {
    v: Vec3,
};
const Vec3 = packed struct {
    x: f32,
    y: f32,
    z: f32,
};

test "packed struct access bug" {
    var foo = Foo{
        .v = Vec3{ .x = 1, .y = 2, .z = 3 },
    };
    var v = foo.v;

    warn("{d}\n", v.x); // 1
    warn("{d}\n", v.y); // 2
    warn("{d}\n", v.z); // 3

    warn("{d}\n", foo.v.x); // 1
    warn("{d}\n", foo.v.y); // 1
    warn("{d}\n", foo.v.z); // 1

    assert(v.y == 2);
    assert(foo.v.y == 2);
}
```

This only fails if `Foo` and `Vec3` are `packed`, and it seems to be sensitive to the size/contents/alignment of the structs: for a `Vec2` or `Vec4`, this works fine.

This is likely related to #2731 or #2784, but the issue here is incorrect behavior at runtime.

davidgm94:
Closes #3091

### DIFF
--- a/test/behavior/packed-struct.zig
+++ b/test/behavior/packed-struct.zig
@@ -446,3 +446,115 @@ test "optional pointer in packed struct" {
     const x = T{ .ptr = &n };
     try expect(x.ptr.? == &n);
 }
+
+test "nested packed struct field access test" {
+    const Vec2 = packed struct {
+        x: f32,
+        y: f32,
+    };
+
+    const Vec3 = packed struct {
+        x: f32,
+        y: f32,
+        z: f32,
+    };
+
+    const NestedVec2 = packed struct {
+        nested: Vec2,
+    };
+
+    const NestedVec3 = packed struct {
+        nested: Vec3,
+    };
+
+    const vec2 = Vec2{
+        .x = 1.0,
+        .y = 2.0,
+    };
+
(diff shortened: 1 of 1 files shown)


test/behavior/packed-struct.zig
```diff
@@ -446,3 +446,115 @@ test "optional pointer in packed struct" {
     const x = T{ .ptr = &n };
     try expect(x.ptr.? == &n);
 }
+
+test "nested packed struct field access test" {
```
kubkon, line 450: You'll need to skip the test for self-hosted native backends:

```suggestion
test "nested packed struct field access test" {
    if (builtin.zig_backend == .stage2_x86_64) return error.SkipZigTest;
    // and so on for other native backends
```

kubkon (changes requested):

PR #12637, opened 2022-08-26
Add behavior test coverage for nested packed struct field access

### LINKED ISSUE #3091: Nested property access of packed struct loads wrong value

davidgm94:
Closes #3091

Sorry I naively deleted the previous repo, so I opened again the PR. Requested changes by @kubkon done :)

### DIFF
--- a/test/behavior/packed-struct.zig
+++ b/test/behavior/packed-struct.zig
@@ -446,3 +446,121 @@ test "optional pointer in packed struct" {
     const x = T{ .ptr = &n };
     try expect(x.ptr.? == &n);
 }
+
+test "nested packed struct field access test" {
+    if (builtin.zig_backend == .stage2_c) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_wasm) return error.SkipZigTest; // TODO
+    if (builtin.zig_backend == .stage2_x86_64) return error.SkipZigTest; // TODO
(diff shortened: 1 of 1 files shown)


kubkon (approved):

andrewrk:
Thanks!

PR #12646, opened 2022-08-27
Exposes build.zig from std.build for experimentation

MasterQ32:
Exposes `std.build.build_script` in the build runner, to allow access to build.zig from build.zig dependencies.

### DIFF
--- a/lib/build_runner.zig
+++ b/lib/build_runner.zig
@@ -9,6 +9,8 @@ const process = std.process;
 const ArrayList = std.ArrayList;
 const File = std.fs.File;
 
+pub const @"build.zig" = root;
+
 pub fn main() !void {
     // Here we use an ArenaAllocator backed by a DirectAllocator because a build is a short-lived,
     // one shot program. We don't need to waste time freeing memory and finding places to squish
--- a/lib/std/build.zig
+++ b/lib/std/build.zig
@@ -28,6 +28,13 @@ pub const InstallRawStep = @import("build/InstallRawStep.zig");
 pub const OptionsStep = @import("build/OptionsStep.zig");
 pub const EmulatableRunStep = @import("build/EmulatableRunStep.zig");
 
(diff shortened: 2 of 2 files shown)


ifreund:
I think this is going to need some motivating use-cases before we can consider merging it.

MasterQ32:
My primary use case is opening the build system for experimentation and extensions. Right now, it's not possible to access `build.zig` from included files.

I want to play around with package manager things, dependency tracking, automatic package and build support package provisioning. This PR is the least intrusive way of enabling all these things, as it allows to extend the build runner without actually touch the code of it.

What i have built on top of that PR is this:
```zig
game_exe.addPackage(std_build.getPackage(.wz, "^2.3.1"));
```
with declarative package imports:
```zig
// build.zig
pub const pkgs = struct {
    pub const network = std_build.package("vendor/network/network.zig", "1.3.2", .{});

    pub const wz = std_build.package("vendor/wz/wz.zig", "2.7.3", .{
        std_build.dependency(.hzzp, "^2.0.0"),
    });

    pub const hzzp = std_build.package("vendor/hzzp/hzzp.zig", "2.3.1", .{
        std_build.dependency(.network, "^1.0.0"),
    });
};
```

What i want to do is create/provide a way to flatten dependency trees to packages that can be loosely coupled instead of depending on relative paths to submodules or similar, and still provide a convenient interface

andrewrk:
Let's take this to a proposal please, I want to keep the PR queue for ready-to-merge stuff and while I'm happy to hear this proposal out, it does not look likely to be accepted at first glance. You are invited to ignore the honeypot template for proposals when it comes to the zig build system.

MasterQ32:
Okay, i'll hack up my build script so i can apply this as a patch to my local zig deployment until i actually have an idea for a proper proposal. This PR was just a prerequisite so people could easily toy around with the build system without modifying their zig installation

PR #12650, opened 2022-08-27
stage2: change how defers are stored in Zir

Vexu:
Storing defers this way has the benefits that the defer doesn't get multiple times in AstGen, it takes up less space, and it Sema aware of defers allowing for 'unreachable else prong' on error sets in generic code.

The disadvantage is that it is a bit more complex and errdefers with now emit a placeholder instruction (but those are rare).

```
Sema.zig before:
  Total ZIR bytes:    3.7794370651245117MiB
  Instructions:       238996 (2.051319122314453MiB)
  String Table Bytes: 89.2802734375KiB
  Extra Data Items:   430144 (1.640869140625MiB)
Sema.zig after:
  Total ZIR bytes:    3.3344192504882812MiB
  Instructions:       211829 (1.8181428909301758MiB)
  String Table Bytes: 89.2802734375KiB
  Extra Data Items:   374611 (1.4290275573730469MiB)
```

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -5700,7 +5700,6 @@ fn bar() !void {
         try quux();
     } else |err| switch (err) {
         error.FileNotFound => try hello(),
-        else => try another(),
     }
 }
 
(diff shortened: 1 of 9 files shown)


andrewrk:
Nice improvement.

Issue #12449, opened 2022-08-14
Validate literals at a later stage


Vexu:
Most people would recognize `0x213g` as a malformed hex literal and the same goes for `''` being an empty char literal but currently Zig validates these kinds of error in the tokenizer, giving out terrible, vague errors like `expected expression, found 'invalid bytes'` and preventing formatting.

It would instead be better to have the tokenizer only validate things that actually prevent the tokens from being valid such as newlines in string literals (which should also give a proper error message instead of the same `invalid bytes` one) and leave everything else to AstGen.

moosichu:
I'm keen to take a look at this once #12661 has been merged, but currently blocked on waiting for that or otherwise the work might potentially conflict there.

PR #12661, opened 2022-08-28
Correctly handle carriage return characters according to the spec

### LINKED ISSUE #11414: CR-LF Line Endings Break Multi-Line Strings
### Zig Version

0.9.0

### Steps to Reproduce

```zig
test "Multi-Line Strings" {
    const str = 
    \\a b
    \\  c
    ;
}
```
This code will compile when lines end with only a `\n` newline character, but will break if Windows-style `\r\n` line endings are used.

### Expected Behavior

The code compiles successfully.

### Actual Behavior

```zig
src\test.zig:10:10: error: expected ';', found invalid bytes
    \\a b
         ^
src\test.zig:11:6: note: invalid byte: '\\'
    \\  c
     ^
```

### LINKED ISSUE #12674: Cases of retokenization invalidly assume that the tokenizer is stateless leading to crashes in various edge cases
### Zig Version

0.10.0-dev.5788+f6c9b78c8

### Steps to Reproduce

This is mainly a blocker for #12661, and causes a crash there. However this issue is a design bug that is currently present in the code, and I'm sure there could be other ways trigger it.

### Expected Behavior

So the following function in Ast.zig shows this issue quite well:
```zig
pub fn tokenSlice(tree: Ast, token_index: TokenIndex) []const u8 {
    const token_tag = token_tags[token_index];

    // Many tokens can be determined entirely by their tag.
    if (token_tag.lexeme()) |lexeme| {
        return lexeme;
    }

    // For some tokens, re-tokenization is needed to find the end.
    var tokenizer: std.zig.Tokenizer = .{
        .buffer = tree.source,
        .index = token_starts[token_index],
        .pending_invalid_token = null,
    };
    assert(token.tag == token_tag);
    return tree.source[token.loc.start..token.loc.end];
}
```

It expects that a new tokenizer can be created and effectively treats it as stateless provided the correct token_index. This is not the case.

### Actual Behavior

This issue is why #11414 crashes instead of just causing an error.

The tokenizer makes use of a `pending_invalid_token` field to keep track of a bad token if nested inside of another one: e.g. if a bad symbol is found inside of a string.

However, if you restart the tokenizer from an arbitrary character index that is within one of those tokens that can create `a pending_invalid_token`, then it will be in a different state at that index from when the file was originally parsed.

I think the solution here is to roll the tokenizer back to the start of current line and retokenize from there until the correct & matching token is found at the desired character index. I'm working a PR that does that, but it will take me a couple of days to do so, so thought I would make this issue whilst I had a smidgen of time.

Apologies for the weird bug report, as this was one identified form looking at the code and trying to solve another issue which makes this bug symptomatic only with those additional changes from my PR. Also had to dash off so had to write it quickly, so will refine it later! Sorry if it doesn't make a lot of sense.

moosichu:
Taking guidance from: ziglang/zig-spec#38, carriage returns are now allowed preceding line-feeds when found in doc-comments or multiline strings. The CRLF is also only interpreted as a newline in this case. However - they are otherwise rejected when found in any other context except when used as plain whitespace.

This resolves #11414 and resolves #12674. It replaces #12692 which is a part of this PR.

Ideally the error reporting should be improved by #12449 sometime in the future, but I think I might take a look at that once this has been merged.

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -11551,7 +11551,8 @@ fn readU32Be() u32 {}
       </p>
       <p>
       Each LF may be immediately preceded by a single CR (byte value 0x0d, code point U+000d, {#syntax#}'\r'{#endsyntax#})
-      to form a Windows style line ending, but this is discouraged.
+      to form a Windows style line ending, but this is discouraged. Note that in mulitline strings, CRLF sequences will
+      be encoded as LF when compiled into a zig program.
       A CR in any other context is not allowed.
       </p>
       <p>
--- a/lib/std/zig/Ast.zig
+++ b/lib/std/zig/Ast.zig
@@ -171,7 +171,7 @@ pub fn tokenSlice(tree: Ast, token_index: TokenIndex) []const u8 {
         .index = token_starts[token_index],
         .pending_invalid_token = null,
     };
-    const token = tokenizer.next();
+    const token = tokenizer.findTagAtCurrentIndex(token_tag);
     assert(token.tag == token_tag);
     return tree.source[token.loc.start..token.loc.end];
 }
--- a/lib/std/zig/tokenizer.zig
+++ b/lib/std/zig/tokenizer.zig
@@ -406,6 +406,38 @@ pub const Tokenizer = struct {
         saw_at_sign,
     };
 
+    /// This is a workaround to the fact that the tokenizer can queue up
+    /// 'pending_invalid_token's when parsing literals, which means that we need
+    /// to scan from the start of the current line to find a matching tag - just
+    /// in case it was an invalid character generated during literal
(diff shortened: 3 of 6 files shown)


Vexu (changes requested):
Needs a test, you'll have to use string literals in `test/compile_errors.zig`.

andrewrk:
CI checks are failing

moosichu:
Yeah - I haven't had the time to address those test issues for the the past while, but should hopefully be able to at the end of this week if that's ok :) Sorry for the delay on this one!

moosichu:
Just resolved the merge conflicts that I saw had cropped-up (hopefully :crossed_fingers:)

Vexu (approved):

PR #12662, opened 2022-08-28
coff: fix reading COFF header offset

wsengir:
I was quite confused when some executables were seemingly missing the COFF header magic number. It turns out that the `0x3C` COFF header offset was being read as a byte and not a `u32`, causing the header to sometimes be read incorrectly.

I'm also working with COFF for a project, so I added some missing constants I needed.

### DIFF
--- a/lib/std/coff.zig
+++ b/lib/std/coff.zig
@@ -256,20 +256,89 @@ pub const OptionalHeaderPE64 = extern struct {
     number_of_rva_and_sizes: u32,
 };
 
-pub const DebugDirectoryEntry = extern struct {
-    characteristiccs: u32,
-    time_date_stamp: u32,
-    major_version: u16,
-    minor_version: u16,
-    @"type": u32,
-    size_of_data: u32,
-    address_of_raw_data: u32,
-    pointer_to_raw_data: u32,
+pub const IMAGE_NUMBEROF_DIRECTORY_ENTRIES = 16;
+
+pub const DirectoryEntry = enum(u16) {
+    /// Export Directory
+    EXPORT = 0,
+
+    /// Import Directory
+    IMPORT = 1,
+
+    /// Resource Directory
+    RESOURCE = 2,
+
+    /// Exception Directory
+    EXCEPTION = 3,
+
+    /// Security Directory
+    SECURITY = 4,
+
+    /// Base Relocation Table
+    BASERELOC = 5,
+
+    /// Debug Directory
+    DEBUG = 6,
+
+    /// Architecture Specific Data
+    ARCHITECTURE = 7,
+
+    /// RVA of GP
+    GLOBALPTR = 8,
+
+    /// TLS Directory
+    TLS = 9,
+
+    /// Load Configuration Directory
+    LOAD_CONFIG = 10,
+
+    /// Bound Import Directory in headers
+    BOUND_IMPORT = 11,
+
+    /// Import Address Table
+    IAT = 12,
+
+    /// Delay Load Import Descriptors
+    DELAY_IMPORT = 13,
+
+    /// COM Runtime descriptor
+    COM_DESCRIPTOR = 14,
 };
 
 pub const ImageDataDirectory = extern struct {
     virtual_address: u32,
     size: u32,
+};
+
+pub const DebugDirectoryEntry = extern struct {
+    characteristics: u32,
+    time_date_stamp: u32,
+    major_version: u16,
+    minor_version: u16,
+    @"type": DebugType,
+    size_of_data: u32,
+    address_of_raw_data: u32,
+    pointer_to_raw_data: u32,
+};
+
+pub const DebugType = enum(u32) {
+    UNKNOWN = 0,
+    COFF = 1,
+    CODEVIEW = 2,
+    FPO = 3,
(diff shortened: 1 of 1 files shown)


lib/std/coff.zig
```diff
@@ -794,9 +794,36 @@ pub const MachineType = enum(u16) {
     }
 };
 
-const IMAGE_NUMBEROF_DIRECTORY_ENTRIES = 16;
-const IMAGE_DEBUG_TYPE_CODEVIEW = 2;
-const DEBUG_DIRECTORY = 6;
+pub const IMAGE_NUMBEROF_DIRECTORY_ENTRIES = 16;
+
+pub const IMAGE_DEBUG_TYPE = struct {
```
kubkon, line 799: While you're making this change, would you mind converting this into an integer-backed enum? I think it would make more sense than a struct with consts.

lib/std/coff.zig
```diff
@@ -794,9 +794,36 @@ pub const MachineType = enum(u16) {
...
+    pub const COFF = 1;
+    pub const CODEVIEW = 2;
+    pub const FPO = 3;
+    pub const MISC = 4;
+    pub const EXCEPTION = 5;
+    pub const FIXUP = 6;
+    pub const BORLAND = 9;
+};
+
+pub const IMAGE_DIRECTORY_ENTRY = struct {
```
kubkon, line 810: Ditto.

kubkon (changes requested):
Nice, thanks so much for the fix! I've got two requests and after they get addressed, we're ready to merge :-)

lib/std/coff.zig
wsengir: I really wanted to, but in the past we've explicitly avoided them for the various flags in `std.os` so I followed suit here. I don't recall the exact reasoning why, though.

wsengir: It looks like we're now using a number of them in various places in `std.os`, so I guess it's okay now? Will revise.

kubkon: Thanks! And yeah, I think using an integer-backed enum for flags, and packed struct for bitsets is the way to go. It cleans up a lot of code IMHO. This change will also aid me in my quest of writing an incremental COFF linker.

wsengir:
I've `enum`ified and renamed the constants to match the rest of the file.

I also found a few more constants from more up to date headers. Interestingly, [LLVM has slightly different names for the indices](https://llvm.org/doxygen/namespacellvm_1_1COFF.html#a8ec28bc8cedea7febace40170acf2404). I imagine theirs are probably more accurate (with the exception of `NUM_DATA_DIRECTORIES` being `15` instead of `16`), but I left the official Windows ones for now.

lib/std/coff.zig
```diff
@@ -256,20 +256,89 @@ pub const OptionalHeaderPE64 = extern struct {
     number_of_rva_and_sizes: u32,
 };
 
+pub const IMAGE_NUMBEROF_DIRECTORY_ENTRIES = 16;
+
+pub const DataDirectoryIndex = enum(u16) {
```
wsengir, line 261: Rationale behind `u16`:
* [ImageDirectoryEntryToData](https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-imagedirectoryentrytodata)'s parameter `DirectoryEntry`
* [ImageDirectoryEntryToDataEx](https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-imagedirectoryentrytodataex)'s parameter `DirectoryEntry`

lib/std/coff.zig
```diff
@@ -256,20 +256,89 @@ pub const OptionalHeaderPE64 = extern struct {
     number_of_rva_and_sizes: u32,
 };
 
+pub const IMAGE_NUMBEROF_DIRECTORY_ENTRIES = 16;
+
+pub const DirectoryEntry = enum(u16) {
```
wsengir, line 261: Rationale behind `u16`:
* [ImageDirectoryEntryToData](https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-imagedirectoryentrytodata)'s parameter `DirectoryEntry` is `USHORT`
* [ImageDirectoryEntryToDataEx](https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-imagedirectoryentrytodataex)'s parameter `DirectoryEntry` is `USHORT`

kubkon:
> I've `enum`ified and renamed the constants to match the rest of the file.
> 
> I also found a few more constants from more up to date headers. Interestingly, [LLVM has slightly different names for the indices](https://llvm.org/doxygen/namespacellvm_1_1COFF.html#a8ec28bc8cedea7febace40170acf2404). I imagine theirs are probably more accurate (with the exception of `NUM_DATA_DIRECTORIES` being `15` instead of `16`), but I left the official Windows ones for now.

Isn't this due to the fact the last data directory is always zeroed out 8bytes aka `Reserved`? So I guess LLVM would always implicitly add `+1` whenever `NUM_DATA_DIRECTORIES` is used. Just a guess though. I strongly prefer to stick to the actual OS docs, so leaving it at `16` is a good call :+1:

PR #12663, opened 2022-08-28
Stage2 fixes

Vexu:


src/Sema.zig
```diff
@@ -22516,14 +22540,22 @@ const CoersionError = CompileError || error{
     NotCoercible,
 };
 
+const CoerceOpts = struct {
+    /// Should coerceExtra emit error messages.
+    report_err: bool = true,
+    /// Ignored if `report_err == false`.
+    is_ret: bool = false,
+    /// Should coercion to comptime_int ermit an error message.
+    no_cast_to_comptime_int: bool = false,
```
andrewrk, line 22549: I don't think this behavior should be different with `const y = @as(T, x)` vs `const y: T = x;`. We want to be able to say that these two statements are identical. I'd like to push back on adding `ty_builtin` as a ResultLoc, and `builtin_as` as a ZIR instruction.

andrewrk (changes requested):
I have requested changes for one specific commit but everything else looks great. Do you mind if I remove 5bbacfe8a1ad67ec6676eee81776bbdc642d3183 from the branch and merge this PR, and then we can discuss that commit separately?

Edit: I'm going to go ahead and do this to save us one round trip through the timezone difference.

src/Sema.zig
Vexu: There needs to be some way to disable that error for casting the shift operand to the log2 type in order to get a good error message. Would it be better if there was instead a `as_shift_operand` instruction?

PR #12673, opened 2022-08-29
Add docstrings to some functions in std.meta

yujiri8:


yujiri8:
Fixed merge conflicts :)

PR #12692, opened 2022-08-30
Scan from line start when finding tag in tokenizer

### LINKED ISSUE #12674: Cases of retokenization invalidly assume that the tokenizer is stateless leading to crashes in various edge cases

moosichu:
This resolves a crash that can occur for invalid bytes like carriage
returns that are valid characters when not parsed from within literals.

There are potentially other edge cases this could resolve as well, as
the calling code for this function didn't account for any potential
'pending_invalid_tokens' that could be queued up by the tokenizer from
within another state.

This PR resolves #12674 and will unblock #12661.

#12449 could later simplify this when implemented, however this was a super simple change and fixes a known crash.

### DIFF
--- a/lib/std/zig/Ast.zig
+++ b/lib/std/zig/Ast.zig
@@ -116,7 +116,7 @@ pub fn tokenSlice(tree: Ast, token_index: TokenIndex) []const u8 {
         .index = token_starts[token_index],
         .pending_invalid_token = null,
     };
-    const token = tokenizer.next();
+    const token = tokenizer.findTagAtCurrentIndex(token_tag);
     assert(token.tag == token_tag);
     return tree.source[token.loc.start..token.loc.end];
 }
--- a/lib/std/zig/tokenizer.zig
+++ b/lib/std/zig/tokenizer.zig
@@ -406,6 +406,38 @@ pub const Tokenizer = struct {
         saw_at_sign,
     };
 
+    /// This is a workaround to the fact that the tokenizer can queue up
+    /// 'pending_invalid_token's when parsing literals, which means that we need
+    /// to scan from the start of the current line to find a matching tag - just
+    /// in case it was an invalid character generated during literal
+    /// tokenization. Ideally this processing of this would be pushed to the AST
+    /// parser or another later stage, both to give more useful error messages
+    /// with that extra context and in order to be able to remove this
+    /// workaround.
+    pub fn findTagAtCurrentIndex(self: *Tokenizer, tag: Token.Tag) Token {
+        if (tag == .invalid) {
+            const target_index = self.index;
+            var starting_index = target_index;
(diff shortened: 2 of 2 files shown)


Vexu:
`tokenSlice` is quite heavily used so if this only affects `.invalid` tokens then it should only be called for them.

moosichu:
Only just found some time to get around to this! Good shout, that change has been pushed.

andrewrk (changes requested):
Thanks for this improvement. Needs a unit test before being merged - especially because there is already discussion about wanting to remove the workaround, in which case we would want to make sure it doesn't break!

moosichu:
#12661 Already supercedes this anyway, so if you are happy with that (especially because it has tests) I don't think this needs to be merged. This was mainly intended as a low-risk workaround for a known crash, whereas #12661 actually fixes the issue by adhering to the spec.

Will close this for now! Mainly becuase any tests for this would be made redundant anyway.

Issue #12694, opened 2022-08-31
Compiler panic when a package has an @import that refers to a missing dependency


kcbanner:
### Zig Version

0.10.0-dev.3823+59e33b447

### Steps to Reproduce

Note: This is `0.10.0-dev.3823+59e33b447`, but with `zig2.exe` as the compiler, since the stage3 build fails for me currently.

Create a project with the following structure:

build.zig:
```
const std = @import("std");

pub fn build(b: *std.build.Builder) void {
    const target = b.standardTargetOptions(.{});
    const mode = b.standardReleaseOptions();

    const exe = b.addExecutable("zig_pkg_crash", "src/main.zig");
    exe.setTarget(target);
    exe.setBuildMode(mode);
    exe.addPackage(std.build.Pkg{
        .name = "bar",
        .source = .{ .path = "src/bar.zig" },
        .dependencies = &[_]std.build.Pkg{
            // Uncomment this to fix the compiler crash
            // .{
            //     .name = "foo",
            //     .source = .{ .path = "deps/foo/foo.zig" }
            // }
        }
    });

    exe.install();
}
```

`src/main.zig`
```
const std = @import("std");
const bar = @import("bar");

pub fn main() !void {
    bar.hello();
}
```

`src/bar.zig`
```
const foo = @import("foo");

pub fn hello() void {
    foo.hello();
}
```

`deps/foo/foo.zig`
```
const std = @import("std");

pub fn hello() void {
    std.debug.print("Hello from foo!", .{});
}
```

If you run `zig build`, the compiler crashes. If you uncomment the `foo` dependency in build.zig, it compiles as expected.

### Expected Behavior

An error message about `@import("foo")` not being resolved.

### Actual Behavior

`panic: reached unreachable code`

andrewrk:
Fixed by #12701

PR #12701, opened 2022-08-31
fix panic when the dependency is missing

ri-char:
Fix panic when the dependency is missing in #12694.

image

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -10378,8 +10378,10 @@ fn zirImport(sema: *Sema, block: *Block, inst: Zir.Inst.Index) CompileError!Air.
                         break :blk pkg.key_ptr.*;
                     }
                 }
-                unreachable;
-            } else unreachable;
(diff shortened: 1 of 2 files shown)


Vexu (changes requested):
This code is correct but it assumes that packages contain no dangling pointers which they do because of this part in `main.zig`:
https://github.com/ziglang/zig/blob/26e98391322645f2d09aaaaf36d94cae06424a18/src/main.zig#L796-L803

The correct solution would be to update `parent`for all of the packages in `pkg_tree_root` here, making it null if not available.
https://github.com/ziglang/zig/blob/26e98391322645f2d09aaaaf36d94cae06424a18/src/main.zig#L2751-L2755

Note that making it `null` also means that that second unreachable becomes reachable.

Vexu (approved):
I'm not quite sure how to add a test for this but I can confirm that it works.

PR #12708, opened 2022-09-01
Fix #12655: zig fmt remove trailing comma at the end of assembly clobber

### LINKED ISSUE #12655: `zig fmt` Segfaults when there's trailing comma at the end of inline assembly clobber
### Zig Version

0.10.0-dev.3685+dae7aeb33

### Steps to Reproduce

1. Write a file with inline assembly, with trailing comma at the end of clobber.
```zig
fn example() void {
    asm volatile (
        "nop"
        :
        : [a] "i" (0),
        : "eax", // <-- Trailing comma at the end of clobber
    );
}
```
2. Run `zig fmt` on it:
```sh
$ zig fmt ./test.zig
```

NOTE: Here's my `uname -a` output:
```
$ uname -a
Linux yjk-desktop 5.10.102.1-microsoft-standard-WSL2 #1 SMP Wed Mar 2 00:30:59 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

### Expected Behavior

Format the given code without any errors. For reference, this is what `zig fmt` outputs when there's no comma at the end of clobber:
```zig
fn example() void {
    asm volatile ("nop"
        :
        : [a] "i" (0),
        : "eax"
    );
}
```

### Actual Behavior

Running `zig fmt` shows `Segmentation fault`:
```
$ zig fmt ./zigfmt_asmvolatile.zig
Segmentation fault
```

ri-char:
Fix #12655 
It will remove the trailing comma at the end of inline assembly clobber.
For example, before format:
```zig
fn foo() void {
    asm volatile (""
        : [_] "" (-> type),
        :
        : "clobber"
    );
    asm volatile (""
        :
        : [_] "" (type),
        : "clobber",
    );
}
```

After format:
```zig
fn foo() void {
    asm volatile (""
        : [_] "" (-> type),
        :
        : "clobber"
    );
    asm volatile (""
        :
        : [_] "" (type),
        : "clobber"
    );
}
```

### DIFF
--- a/lib/std/zig/parser_test.zig
+++ b/lib/std/zig/parser_test.zig
@@ -26,7 +26,7 @@ test "zig fmt: remove trailing comma at the end of assembly clobber" {
         \\    );
         \\}
         \\
-        ,
+    ,
         \\fn foo() void {
         \\    asm volatile (""
         \\        : [_] "" (-> type),
--- a/lib/std/zig/parser_test.zig
+++ b/lib/std/zig/parser_test.zig
@@ -16,6 +16,28 @@ test "zig fmt: preserves clobbers in inline asm with stray comma" {
     );
 }
 
+test "zig fmt: remove trailing comma at the end of assembly clobber" {
+    try testTransform(
+        \\fn foo() void {
+        \\    asm volatile (""
+        \\        : [_] "" (-> type),
+        \\        :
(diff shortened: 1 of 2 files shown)


Vexu (changes requested):
Needs a test in `lib/std/zig/parser_test.zig`.

PR #12712, opened 2022-09-01
Change T to u16

Deecellar:
The element_bit_size will break if T is ~~unsigend~~ signed by how ceilPowerOfTwo works
@noiryuh made a good point there, u16 should be enough, I hope


squeek502:
Do you mean if T was *signed*?

Might be nice to add a test case that would have failed previously to give this change some context.

Deecellar:
> Do you mean if T was _signed_?
> 
> Might be nice to add a test case that would have failed previously to give this change some context.

Fair, and yes if T was signed, man, I am really bad with communication lately

PR #12720, opened 2022-09-02
Fix #12440: std.math.big.Rational order/orderAbs

### LINKED ISSUE #12440: std.math.big.Rational.order and std.math.big.Rational.orderAbs Swap is_abs Argument
It looks to me that the order and orderAbs functions in std.math.big.Rational have their 'is_abs' arguments swapped- order should give 'false' and orderAbs should give 'true' since it wants an absolute ordering.

order function:
https://github.com/ziglang/zig/blob/54454fd0102af8b25dbc85751d37fd265380d920/lib/std/math/big/rational.zig#L337

orderAbs:
https://github.com/ziglang/zig/blob/54454fd0102af8b25dbc85751d37fd265380d920/lib/std/math/big/rational.zig#L343

yujiri8:


ehaas:
Would be good to add 2 test cases that failed with the previous code, e.g.:

```zig
test {
    var a = try Rational.init(testing.allocator);
    defer a.deinit();
    var b = try Rational.init(testing.allocator);
    defer b.deinit();

    try a.setRatio(1, 1);
    try b.setRatio(-1, 1);
    try testing.expectEqual(Order.gt, try a.order(b));
}
```

nsmryan:
I would suggest also adding at least one test for orderAbs, which does not currently appear in a test case:

```zig
test {
    var a = try Rational.init(testing.allocator);
    defer a.deinit();
    var b = try Rational.init(testing.allocator);
    defer b.deinit();

    try a.setRatio(1, 1);
    try b.setRatio(-2, 1);
    try testing.expectEqual(Order.gt, try a.order(b));
    try testing.expectEqual(Order.lt, try a.orderAbs(b));
}
```

yujiri8:
Added a test for both functions.

PR #12728, opened 2022-09-03
Fix #12680: fix the `type` parameter of `GlobalAlias`

### LINKED ISSUE #12680: LLVM error when using extern functions
### Zig Version

0.10.0-dev.3780+56bdd0127

### Steps to Reproduce

Initialize a folder with `zig init-exe`
Have two files. A c.o file from the c file as follows:
```c
extern void test_func();
void c_func() {
    test_func();
}
```
I compiled this using
```bash
clang -c c.c -o c.o
```

With the zig code as follows
```zig
const std = @import("std");
pub extern c_func() callconv(.C) void;
pub fn main() anyerror!void {
    c_func();
}
pub export fn testFunc() callconv(.C) void {
   std.debug.print("func", .{});
}
comptime {
    @export(testFunc, .{ name = "test_func", .linkage = .Strong});
}
```

In the build.zig, add this:
```zig
/// ...
exe.setBuildMode(mode);
exe.addObjectFile("c.o");
exe.install();
```

Then run `zig build`.

### Expected Behavior

The program runs and compiles. This works in 0.9.1.

### Actual Behavior

An error occurs
```
LLVM Emit Object...                                                                                       
Alias and aliasee types should match!                                                                     
void ()** @test_func                                                                                      
                                                                                                          
thread 296040 panic: LLVM module verification failed
```

ri-char:
In llvm, `GlobalValue.typeOf()` returns a Pointer type that point to the real type, which causes #12680.
Instead `GlobalValue.globalGetValueType()` returns the real type.


Vexu (changes requested):
Needs a test, `test/behavior/bugs` would be a good place.

ri-char:
It requires that the compiler compiles the code and exports the function correctly, but I don't know what to write in test scope.

test/behavior/bugs/12680.zig
```diff
@@ -0,0 +1,10 @@
+// export this function twice
+pub export fn testFunc() callconv(.C) void {}
+
+comptime {
+    @export(testFunc, .{ .name = "test_func", .linkage = .Strong });
+}
+
+test "export a function twice" {
+    _ = testFunc();
```
Vexu, line 9: This doesn't really test that it is being exported correctly. Try putting the export in a separate function and an extern in this one and check that it is returning the correct value.

ri-char:
Oh no, I made the compiler Segmentation fault.
12680.zig:
```zig
const std = @import("std");
const expectEqual = std.testing.expectEqual;
const other_file =  @import("12680_other_file.zig");

extern fn test_func(a: u32, b: u32) callconv(.C) u32;

test "export a function twice" {
    // If it exports the function correctly, `test_func` and `testFunc` will points to the same address.
    try expectEqual(&test_func, &other_file.testFunc);
}
```
12680_other_file.zig:
```zig
// export this function twice
pub export fn testFunc(a: u32, b: u32) callconv(.C) u32 {
    // we don't care what is in the function
    return a+b;
}

comptime {
    @export(testFunc, .{ .name = "test_func", .linkage = .Strong });
}
```

Output:
```
LLVM Emit Object... Segmentation fault at address 0x0
???:?:?: 0x7f1cb090628c in ??? (???)
/zig/src/codegen/llvm.zig:688:47: 0x56035336f0f8 in codegen.llvm.Object.flushModule (zig2)
        try self.resolveExportExternCollisions();
                                              ^
/zig/src/link/Elf.zig:960:47: 0x56035338276b in link.Elf.flushModule (zig2)
            return try llvm_object.flushModule(comp, prog_node);
                                              ^
/zig/src/link/Elf.zig:1250:29: 0x560353379278 in link.Elf.linkWithLLD (zig2)
        try self.flushModule(comp, prog_node);
                            ^
/zig/src/link/Elf.zig:946:32: 0x56035336dffd in link.Elf.flush (zig2)
        return self.linkWithLLD(comp, prog_node);
                               ^
/zig/src/link.zig:649:68: 0x56035336b2c7 in link.File.flush (zig2)
            .elf => return @fieldParentPtr(Elf, "base", base).flush(comp, prog_node),
                                                                   ^
/zig/src/Compilation.zig:2369:24: 0x5603532fe028 in Compilation.flush (zig2)
    comp.bin_file.flush(comp, prog_node) catch |err| switch (err) {
                       ^
/zig/src/Compilation.zig:2350:23: 0x5603532f0ca0 in Compilation.update (zig2)
        try comp.flush(main_progress_node);
                      ^
/zig/src/main.zig:3348:20: 0x56035327ff3f in main.updateModule (zig2)
    try comp.update();
                   ^
/zig/src/main.zig:3022:17: 0x5603531d7f1b in main.buildOutputType (zig2)
                ^
/zig/src/main.zig:236:31: 0x5603531848b6 in main.mainArgs (zig2)
                              ^
/zig/src/stage1.zig:48:24: 0x560353184049 in main (zig2)
                       ^
fish: Job 1, './build/zig2 test ./test/behavi…' terminated by signal SIGABRT (Abort)
```

Vexu:
I meant that you'd call the extern function but it'll be good to have that reported as a bug.

ri-char:
This test works in stage1, but causes segmentation fault in stage2, so I think that the test is correct but there's something wrong with stage2.
We can verify that it exports the function correctly manually in stage2:
```
# zig build-obj ./test/behavior/bugs/12680_other_file.zig
# nm 12680_other_file.o | grep test
0000000000000000 T testFunc
0000000000000000 T test_func
```

ri-char:
This commit fix the `Segmentation fault` above.

Vexu (approved):

Issue #12733, opened 2022-09-04
translate-c refuses to translate packed structs the compiler would otherwise accept (needed for M1 macOS stdlib.h)


Techcable:
### Zig Version

0.10.0-dev.3871+b7d5582de

## Current ABI behavior of `packed struct` on stage2 (Zig side)
As of commit b83c037f9ffd7a4285de41c95827615fcbdbbf2f , a `packed struct` is only allowed to be `extern` (on stage2) If it is "ABI sized".

According to @Vexu this is because the stage3 representation is incompatible with C.

Consider the following struct ([full file here](https://gist.github.com/Techcable/c5a667112e4c17fcd470298cbf3edbdb)):
```zig
pub const Example = packed struct {
    a: i32 = 5,
    b: i32 = 6,
    c: i8 = 4,
};
```

This struct is not ABI-sized on my x86_64 Linux laptop, so I get the following error:
```
hello.zig:10:5: error: unable to export type 'hello.Example'
pub export var exported = Example { };
    ^~~~~~
hello.zig:10:5: note: only extern structs and ABI sized packed structs are extern compatible
hello.zig:4:28: note: struct declared here
pub const Example = packed struct {
```

This is the error I would expect (given the current rules).

If I delete the `c: i8` field, then it successfully compiles and I get `hello.Example{ .a = 5, .b = 6 }` as output.

## Failed `translate-c` error
Consider the following C header file:
```c
struct example {
    int val; 
} __attribute__((__packed__));
```

If you ran `zig translate-c` on this, you would expect this to translate into `packed struct { val: c_int }`

However the current output is
```zig
pub const example = @compileError("cannot translate packed record union"); // hello.h:1:8
```

This is a false-positive! The restrictions `translate-c` puts on `packed struct` should be consistent with the regular compiler.

### Cause
With a quick ripgrep, the source of `translate_c.zig` seems to have a blanket-ban on translating packed structs from C -> Zig.

https://github.com/ziglang/zig/blob/b7d5582dede8f4cae341365ac9c47c840bd80eff/src/translate_c.zig#L1169-L1171

Unfortunately this blanket-ban on `packed struct` includes the "ABI-sized" structs that Zig would normally .

### M1 Mac woes
While this bug is not platform-specific, it is really severe on `aarch64-macos.12`. The `stdlib.h` (indirectly) includes two of these structures for the purposes. They appear to be very important and I can't even compile an empty file without getting an error.

[...]

PR #12730, opened 2022-09-03
stage2 llvm: implement aarch64 C ABI

Vexu:
... at least enough to pass all the current tests.

Should fix #12483

PR #12735, opened 2022-09-04
translate-c: Allow translating packed C structs iff ABI-sized

Techcable:
This relaxes the restriction added in c7884af063791211544c6595a4900bbfcd5d96b6

It fixes issue #12733

Right now, the compiler only considered packed structs to be
C compatible (extern) if the structs are "ABI sized".

See also b83c037f9ffd7a4285de41c95827615fcbdbbf2f for details on
the extern restrictions on `packed struct`

This is an issue on all platforms, but is of particular
importance on aarch64-macos (see my other issue for details)

Previously a type like

```c
struct example {
    int val;
} __attribute__((__packed__));
```

would fail to translate and turn into

```zig
pub const example = @compileError("cannot translate packed record union");
```

This fixes this issue for ABI-sized struct, making the translate-c
rules consistent with the regular compiler.

However, we still keep the restriction on structs that are not "ABI-sized"

So
```c
struct example {
    long a, b, c
} __attribute__((__packed__));
```

will still fail with the appropriate error (in the sense it will emit `@compileError`)

### Implementation details

For the purposes of translate-c, a packed struct is considered
"ABI-sized" when sizeof(target_struct) <= sizeof(void*)

Unfortunately, right now the translate-c `*Context` structure (on the Zig side)
doesn't directly have any information on the target we're translating into.

However, clang has this target information, and keeps it  in clang::ASTContext.
I added a helper function to extract the `clang::TargetInfo` from this,
then query the pointer size.

This is potentially controversial, since it technically makes the
output target-dependent.
The decision of whether to accept/reject a packed struct varies depending
on the target's pointer size.

Howevee, the translator already asks for (integer) field offsets elsewhere,
so I'm pretty sure target info already impacts translation.

We also add getSize() and getDataSize() functions to *clang.RecordDecl.
These return padded and unpadded sizes respectively.
For a packed struct, they should be the same.

The overall effect is 4 more functions added to `zig_clang.cpp` used to
extract target info & query sizes.

### Overall Effect on M1 Macs
As I mentioned in the issue, `aarch64-macos.12` uses (ABI-sized) packed structs in `stdlib.h`. Even an empty file fails to compile without support. This makes resolving the underlying problem (#12733) crucial.

After fixing the bug (and applying #12730 ), I can get `build test-run-translated-c` to finally work!

Before, it just gave a nasty error:
````
/Users/nicholas/.cache/zig/o/d0183770046c7c4914f3676ac67443fd/source.zig:560:39: error: use of undeclared identifier 'struct__OSUnalignedU16'
    return _OSSwapInt16(@intToPtr([*c]struct__OSUnalignedU16, @intCast(usize, @ptrToInt(_base)) +% _offset).*.__val);
                                      ^~~~~~~~~~~~~~~~~~~~~~
/Users/nicholas/.cache/zig/o/d0183770046c7c4914f3676ac67443fd/source.zig:565:39: error: use of undeclared identifier 'struct__OSUnalignedU32'
    return _OSSwapInt32(@intToPtr([*c]struct__OSUnalignedU32, @intCast(usize, @ptrToInt(_base)) +% _offset).*.__val);
                                      ^~~~~~~~~~~~~~~~~~~~~~
````

Anyways, combining these two PRs seems to fix the most serious stage3 compiler and tranlsate-c bugs that I've been encountering on my M1 Mac. 

In addition to fixing `test-run-translated-c`, it also fixes my own libraries. In particular the test suite of [zig-mpack](https://github.com/Techcable/zig-mpack) finally passes!

With stage3, most of the remaining issues in my main codebase seem to be bugs with my project that stage1 didn't catch. For example, I had two conflicting definitions of `const debug = .....`. It was defined twice in my root module.

Hopefully these fixes mean I can *finally* get back to doing real programming with stage3 and Zig as my backend. Big thanks of course to @Vexu and the whole Zig team 🎉 

## Remaining (possible) questions
- Do I need to add test coverage for this new functionality?
   - Usually this is required, however this is really more of a fix for existing (broken) tests than a new one
   - I really don't understand why we only support certain `packed struct` in stage3. Ideally wouldn't we support all of them?
- Is there a cleaner way to do this without querying clang so much?

Please note, I have homework due (that I've been dodging) so I may not be super responsive (maybe just amend the PR?)

### DIFF
--- a/src/clang.zig
+++ b/src/clang.zig
@@ -148,6 +148,9 @@ pub const APSInt = opaque {
 pub const ASTContext = opaque {
     pub const getPointerType = ZigClangASTContext_getPointerType;
     extern fn ZigClangASTContext_getPointerType(*const ASTContext, T: QualType) QualType;
+
+    pub const getTargetInfo = ZigClangASTContext_getTargetInfo;
+    extern fn ZigClangASTContext_getTargetInfo(*const ASTContext) *const TargetInfo;
 };
 
 pub const ASTUnit = opaque {
@@ -194,6 +197,12 @@ pub const ASTRecordLayout = opaque {
 
     pub const getAlignment = ZigClangASTRecordLayout_getAlignment;
     extern fn ZigClangASTRecordLayout_getAlignment(*const ASTRecordLayout) i64;
+
+    pub const getSize = ZigClangASTRecordLayout_getSize;
+    extern fn ZigClangASTRecordLayout_getSize(*const ASTRecordLayout) i64;
+
+    pub const getDataSize = ZigClangASTRecordLayout_getSize;
+    extern fn ZigClangASTRecordLayout_getDataSize(*const ASTRecordLayout) i64;
 };
 
 pub const AttributedType = opaque {
@@ -893,6 +902,11 @@ pub const SwitchStmt = opaque {
 pub const TagDecl = opaque {
     pub const isThisDeclarationADefinition = ZigClangTagDecl_isThisDeclarationADefinition;
     extern fn ZigClangTagDecl_isThisDeclarationADefinition(*const TagDecl) bool;
+};
+
+pub const TargetInfo = opaque {
+    pub const getMaxPointerWidth = ZigClangTargetInfo_getMaxPointerWidth;
+    extern fn ZigClangTargetInfo_getMaxPointerWidth(*const TargetInfo) u64;
 };
 
 pub const Type = opaque {
--- a/src/translate_c.zig
+++ b/src/translate_c.zig
@@ -1166,8 +1166,8 @@ fn transRecordDecl(c: *Context, scope: *Scope, record_decl: *const clang.RecordD
             });
         }
 
-        if (!c.zig_is_stage1 and is_packed) {
-            return failDecl(c, record_loc, bare_name, "cannot translate packed record union", .{});
+        if (!c.zig_is_stage1 and is_packed and !isAbiSizedPackedRecord(c, record_decl)) {
+            return failDecl(c, record_loc, bare_name, "cannot translate packed record union (unless ABI sized)", .{});
         }
 
         const record_payload = try c.arena.create(ast.Payload.Record);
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Review by Vexu — CHANGES_REQUESTED** [review]
Packed structs are not ABI compatible with C structs but with ints. Ints and structs with only one pointer sized int seem to be always compatible so if you add a check for there only being one field then I think this is fine until we get a better solution.

**Comment by ifreund** [comment]
> Packed structs are not ABI compatible with C structs but with ints. Ints and structs with only one pointer sized int seem to be always compatible so if you add a check for there only being one field then I think this is fine until we get a better solution.

I think even this specific case of packed structs with a single integer is incompatible as the alignment is different. It seems that the entire reason these structures use `__attribute__(__packed__)` in the header is to allow unaligned integers. However, zig's integer backed packed structs are no different from the backing integer on the ABI and use the natural alignment of the backing integer.

I believe the correct translation of this example would be an extern struct with `align(1)` on the field:
```c
 struct _OSUnalignedU32 { 
 	uint32_t __val; 
 } __attribute__((__packed__)); 
```
```zig
pub const _OSUnalignedU32 = extern struct {
    val: u32 align(1),
};
```
I've left the `volatile` out of this example to keep things simpler as I forget how we translate that.

Techcable:
> Packed structs are not ABI compatible with C structs but with ints. Ints and structs with only one pointer sized int seem to be always compatible so if you add a check for there only being one field then I think this is fine until we get a better solution.

@Vexu 

I have added in this restriction (and rebased onto master).

However we may want to look deeper into why the ABIs are incompatible. It may have something to do with @ifreund 's comment on alignment.

> I think even this specific case of packed structs with a single integer is incompatible as the alignment is different. It seems that the entire reason these structures use `__attribute__(__packed__)` in the header is to allow unaligned integers. However, zig's integer backed packed structs are no different from the backing integer on the ABI and use the natural alignment of the backing integer.

@ifreund 

Agreed. I think you are correct that `align(1)` is required. However, the way I am read the GCC documentation, this is actually implied by the `packed` modifier (maybe this is why we are having ABI troubles?). 

First looking at the [Common Type attributes page](https://gcc.gnu.org/onlinedocs/gcc/Common-Type-Attributes.html#Common-Type-Attributes):

The final line of the documentation for `__attribute__ ((aligned))` says:

> When used on a struct, or struct member, the aligned attribute can only increase the alignment; **in order to decrease it, the packed attribute must be specified as well**.
> When used as part of a typedef, the aligned attribute can both increase and decrease alignment, and specifying the packed attribute generates a warning.

This implies the `__packed__` attribute is actually used for alignment purposes. This is not immediately clear; there is a level of indirection in the docs.

> This attribute, attached to a struct, union, or C++ class type definition, specifies that each of its members (other than zero-width bit-fields) is placed to minimize the memory required. **This is equivalent to specifying the packed attribute on each of the members**.

Then I went to go look at the documentation for [Common Variable Attributes](https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#Common-Variable-Attributes):

> The packed attribute specifies that a structure member should have the smallest possible alignment—one bit for a bit-field and one byte otherwise, unless a larger value is specified with the aligned attribute. The attribute does not apply to non-member objects.

Combining this means that `align(1)` is automatically implied for every element of a packed structure (from GCC perspective).

I think this is (yet another) independent bug. It is present even in stage1 translate-c.

We need to change `translate-c` to reflect that and add `align(1)` to every field. Maybe that will resolve some of our ABI problems and we can relax the "ABI-sized" restriction?

Techcable:
This PR technically works, but only by accident. In Zig , a `packed struct` is really just an integer.  The correct translation is as an `extern struct ` with `align(1)`

I plan to amend my other PR tomorrow (#12745) to fix this.

@Vexu has pointed this out a couple times, but it took a while to sink in 😉 

Here is the code that (currently) lowers packed structs directly LLVM types:
https://github.com/ziglang/zig/blob/e323cf1264f390911dcc2efea71d46be1d631d92/src/codegen/llvm.zig#L3541-L3572

Issue #12420, opened 2022-08-12
pretty much all the test-cases on windows are failing and have been failing for months


andrewrk:
We've really let down our Windows users lately, this just tragic.

Related: #12380 #12381

```
c:\msys64\home\andy\dev\zig>stage1\bin\zig.exe build test docs -Dstatic-llvm --search-prefix C:\msys64\home\andy\local\zig+llvm+lld+clang-x86_64-windows-gnu-0.10.0-dev.3524+74673b7f6
aarch64-macos\hello_world_with_updates (stage2, aarch64-macos) failed: WrongCompileErrors
adding_numbers_at_runtime_and_comptime (stage2, x86_64-macos) failed: UnexpectedCompileErrors
adding_numbers_at_runtime_and_comptime (stage2, aarch64-macos) failed: UnexpectedCompileErrors
bad_inferred_variable_type (stage2, x86_64-macos) failed: WrongCompileErrors
bad_inferred_variable_type (stage2, aarch64-macos) failed: WrongCompileErrors
break_continue (stage2, x86_64-macos) failed: UnexpectedCompileErrors
break_continue (stage2, aarch64-macos) failed: UnexpectedCompileErrors
catch_at_comptime (stage2, x86_64-macos) failed: UnexpectedCompileErrors
catch_at_comptime (stage2, aarch64-macos) failed: UnexpectedCompileErrors
comparison_of_non-tagged_union_and_enum_literal (stage2, native) failed: TODOImplementWritingObjFiles
compile_error_in_inline_fn_call_fixed (stage2, x86_64-macos) failed: WrongCompileErrors
compile_error_in_inline_fn_call_fixed (stage2, aarch64-macos) failed: WrongCompileErrors
compile_errors\AstGen_comptime_known_struct_is_resolved_before_error (stage2, native) failed: TODOImplementWritingObjFiles
compile_errors\C_pointer_pointing_to_non_C_ABI_compatible_type_or_has_align_attr (stage2, native) failed: TODOImplementWritingObjFiles
compile_errors\C_pointer_to_anyopaque (stage2, native) failed: TODOImplementWritingObjFiles
compile_errors\Issue_5586_Make_unary_minus_for_unsigned_types_a_compile_error (stage2, native) failed: TODOImplementWritingObjFiles
compile_errors\Issue_6823_dont_allow_ (stage2, native) failed: TODOImplementWritingObjFiles
[...]

Issue #1840, opened 2018-12-18
Split "windows" into "nt" and "win32"


daurnimator:
"windows" as most people think of it is a stack of
  - the nt kernel
  - ntdll.dll
  - kernel32.dll
  - user32.dll

You don't have to use all elements of this stack, and there are situations where you want to avoid or are not able to use kernel32.dll and/or user32.dll.

The Zig standard library should not assume that kernel32.dll and user32.dll are unconditionally available.

andrewrk:
> there are situations where you want to avoid or are not able to use kernel32.dll and/or user32.dll.

What situations are these? Does anyone know?

This is an important proposal if we have some use cases to guide it.

suirad:
Some background for anyone not familiar: `ntdll.dll` is the only constant that is required for a windows userland process, as its what does the syscalls. `kernel32.dll` mostly just wraps ntdll and provides what is colloquially known as Winapi. `user32` is essentially the `kernel32` of Windows gui stuff.
The windows userland stack ends up looking like:
`winlibc (aka fopen())` -> `kernel32 (aka CreateFileA())` -> `ntdll (aka NtCreateFile())`.

---
A couple examples of when some of these may not be available/wanted:
* In Windows Kernel/Driver development those `kernel32.dll`, `ntdll.dll` and `user32.dll` aren't available. 
* In some position independent code that is loaded/unpacked/de-obfuscated from elsewhere

---
I would also like to propose having a way to tell what version of windows is being used. Something like `builtin.windows_xp / builtin.windows_vista` because each changes how some winapi are expressed. 

An example of how this would be immediately useful is in creating `std.mutex` where windows xp and above has `CriticalSections` available to use, windows vista+ enables use of `SRWLock` and windows 8+ has `WaitOnAddress` available, which is the closest thing to a futex windows has. This would result in the mutex itself resolving which of these would be best to use at compile time rather than at runtime.

PavelVozenilek:
There's one very specific use case, software hardened against tampering, where people "include" these standard DLL's (their code sections) inside the executable.

However, this (and driver development) is something very rare, and there's always present danger of feature creep, under-documentation, untested/untestable functionality.

andrew-boyarshin:
# Proposal

## OS
- `nt`: kernel-mode/drivers
- `winnative`: native images (like _smss.exe_, _csrss.exe_, _autochk.exe_), user-mode with NTDLL only
[...]

PR #12740, opened 2022-09-04
debug: replace RtlCaptureStackBackTrace (which was spuriously failing) with a new implementation which uses RtlVirtualUnwind instead

kcbanner:
This PR implements `std.debug.walkStackWindows` as a replacement for `RtlCaptureStackBackTrace`. 

### Context

As part of debugging #12703 I was investigating why stack traces weren't appearing, and realized a simple test program that just panicked would not always print the stack trace.

Note that I'm using `zig2` here as I can't build stage3 due to #12703.

Example program:
```
fn bar() void {
    @panic("test");
}

fn foo() void {
    bar();
}

pub fn main() void {
    foo();
}
```

```
>zig2 build-exe panic.zig -target native-windows-msvc

>panic.exe
thread 24112 panic: test

>panic.exe
thread 6572 panic: test
c:\cygwin64\home\kcbanner\temp\zig_panic\panic.zig:2:5: 0x7ff7c6e31d7f in bar (panic.exe.obj)
    @panic("test");
    ^
c:\cygwin64\home\kcbanner\temp\zig_panic\panic.zig:6:8: 0x7ff7c6e315ce in foo (panic.exe.obj)
    bar();
       ^
c:\cygwin64\home\kcbanner\temp\zig_panic\panic.zig:10:8: 0x7ff7c6e3128e in main (panic.exe.obj)
    foo();
       ^
C:\cygwin64\home\kcbanner\kit\zig\lib\std\start.zig:345:41: 0x7ff7c6e31247 in WinStartup (panic.exe.obj)
    std.debug.maybeEnableSegfaultHandler();
                                        ^
???:?:?: 0x7ffac2397033 in ??? (???)
???:?:?: 0x7ffac2ce2650 in ??? (???)

>panic.exe
thread 3864 panic: test

>panic.exe
thread 19984 panic: test
```

The reason for this is that `RtlCaptureStackBackTrace` spuriously returns 0. I couldn't determine the root cause of the spurious failures, but it seemed that others had run into this (https://developercommunity.visualstudio.com/t/capturestackbacktrace-randomly-fails-after-initial/1383213) and had gotten around it by unwinding the stack via the `RtlLookupFunctionEntry`/`RtlVirtualUnwind` calls. 

I referenced these docs for the unwind process:
- https://docs.microsoft.com/en-us/cpp/build/exception-handling-x64?view=msvc-170#unwind-procedure
- http://www.nynaeve.net/?p=101

Supporting changes:
- added `RtlCaptureContext`, `RtlLookupFunctionEntry`, `RtlVirtualUnwind` and supporting types
- fix alignment of `CONTEXT` structs to match winnt.h as required by `RtlCaptureContext` (`fxsave` instr)
  - If a `CONTEXT` was put onto the stack and happened to not be aligned to 16, then the `fxsave` instruction would cause an access violation
- windows aarch64: fix __chkstk being defined twice if libc is not linked on msvc
  - I'm not familiar with this system, but I was hitting this compile error with `-target aarch64-windows-msvc`

### Testing

I'm unsure how to test on windows aarch64. I don't have the hardware - and while the code compiles against `-target aarch64-windows-msvc`, I can't be sure it functions. It seems like qemu for windows doesn't support the same userspace emulation as linux for running aarch64 binaries.

When running `zig2 build test -Dskip-release`, I get failures such as:

```
compiler [51/1153] compile_error_in_inline_fn_call_fixed (stage2, aarch64-macos) [2/2] update [2... error.Unexpected NTSTATUS=0xc0000056
c:\cygwin64\home\kcbanner\kit\zig\lib\std\debug.zig:608:31: 0x7ff77d15ff6f in writeCurrentStackTraceWindows__anon_57536 (test.exe.obj)
    const n = walkStackWindows(addr_buf[0..]);
                              ^
c:\cygwin64\home\kcbanner\kit\zig\lib\std\debug.zig:552:45: 0x7ff77d15fe8d in writeCurrentStackTrace__anon_57535 (test.exe.obj)
        return writeCurrentStackTraceWindows(out_stream, debug_info, tty_config, start_addr);
                                            ^
```

However, these also seem present on master, just with the old stack trace function. This seems like #12420?

I've made my best efforts to follow the style/contributing guidelines and but please let me know if I've missed something as this is my first PR here. I'm also happy to defer this until I can figure out #12703 and properly run the tests.

### DIFF
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -3361,12 +3361,12 @@ pub usingnamespace switch (native_arch) {
         pub const RUNTIME_FUNCTION = extern struct {
             BeginAddress: DWORD,
             EndAddress: DWORD,
-            UnwindData: DWORD
+            UnwindData: DWORD,
         };
 
         pub const KNONVOLATILE_CONTEXT_POINTERS = extern struct {
             FloatingContext: [16]?*M128A,
-            IntegerContext: [16]?*ULONG64
+            IntegerContext: [16]?*ULONG64,
         };
     },
     .aarch64 => struct {
@@ -3451,7 +3451,7 @@ pub usingnamespace switch (native_arch) {
             BeginAddress: DWORD,
             DUMMYUNIONNAME: extern union {
                 UnwindData: DWORD,
-                DUMMYSTRUCTNAME: packed  struct {
+                DUMMYSTRUCTNAME: packed struct {
                     Flag: u2,
                     FunctionLength: u11,
                     RegF: u3,
@@ -3459,8 +3459,8 @@ pub usingnamespace switch (native_arch) {
                     H: u1,
                     CR: u2,
                     FrameSize: u9,
-                }
-            }
+                },
+            },
         };
 
         pub const KNONVOLATILE_CONTEXT_POINTERS = extern struct {
@@ -3474,7 +3474,7 @@ pub usingnamespace switch (native_arch) {
             X26: ?*DWORD64,
             X27: ?*DWORD64,
             X28: ?*DWORD64,
-            Fp :?*DWORD64,
+            Fp: ?*DWORD64,
             Lr: ?*DWORD64,
             D8: ?*DWORD64,
             D9: ?*DWORD64,
@@ -3483,7 +3483,7 @@ pub usingnamespace switch (native_arch) {
             D12: ?*DWORD64,
             D13: ?*DWORD64,
             D14: ?*DWORD64,
-            D15: ?*DWORD64
+            D15: ?*DWORD64,
         };
     },
     else => struct {},
@@ -3501,13 +3501,13 @@ pub const EXCEPTION_ROUTINE = *const fn (
     ExceptionRecord: ?*EXCEPTION_RECORD,
     EstablisherFrame: PVOID,
     ContextRecord: *(Self.CONTEXT),
-    DispatcherContext: PVOID
+    DispatcherContext: PVOID,
 ) callconv(WINAPI) EXCEPTION_DISPOSITION;
 
 pub const UNWIND_HISTORY_TABLE_SIZE = 12;
 pub const UNWIND_HISTORY_TABLE_ENTRY = extern struct {
     ImageBase: ULONG64,
-    FunctionEntry: *Self.RUNTIME_FUNCTION
+    FunctionEntry: *Self.RUNTIME_FUNCTION,
 };
 
 pub const UNWIND_HISTORY_TABLE = extern struct {
@@ -3518,7 +3518,7 @@ pub const UNWIND_HISTORY_TABLE = extern struct {
     Once: BYTE,
     LowAddress: ULONG64,
     HighAddress: ULONG64,
-    Entry: [UNWIND_HISTORY_TABLE_SIZE]UNWIND_HISTORY_TABLE_ENTRY
+    Entry: [UNWIND_HISTORY_TABLE_SIZE]UNWIND_HISTORY_TABLE_ENTRY,
 };
 
 pub const UNW_FLAG_NHANDLER = 0x0;
@@ -3576,7 +3576,7 @@ pub const TEB = extern struct {
 
 pub const EXCEPTION_REGISTRATION_RECORD = struct {
     Next: ?*EXCEPTION_REGISTRATION_RECORD,
-    Handler: ?*EXCEPTION_DISPOSITION
+    Handler: ?*EXCEPTION_DISPOSITION,
 };
 
 pub const NT_TIB = extern struct {
@@ -3584,10 +3584,7 @@ pub const NT_TIB = extern struct {
     StackBase: PVOID,
     StackLimit: PVOID,
     SubSystemTib: PVOID,
-    DUMMYUNIONNAME: extern union {
-        FiberData: PVOID,
-        Version: DWORD
-    },
+    DUMMYUNIONNAME: extern union { FiberData: PVOID, Version: DWORD },
     ArbitraryUserPointer: PVOID,
     Self: ?*@This(),
 };
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -3447,7 +3447,7 @@ pub usingnamespace switch (native_arch) {
             }
         };
 
-        pub const RUNTIME_FUNCTION = struct {
+        pub const RUNTIME_FUNCTION = extern struct {
             BeginAddress: DWORD,
             DUMMYUNIONNAME: extern union {
                 UnwindData: DWORD,
@@ -3574,7 +3574,7 @@ pub const TEB = extern struct {
     TlsExpansionSlots: PVOID,
 };
 
-pub const EXCEPTION_REGISTRATION_RECORD = struct {
+pub const EXCEPTION_REGISTRATION_RECORD = extern struct {
     Next: ?*EXCEPTION_REGISTRATION_RECORD,
     Handler: ?*EXCEPTION_DISPOSITION,
 };
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -2152,7 +2152,6 @@ pub const LONG = i32;
 pub const ULONG64 = u64;
 pub const ULONGLONG = u64;
 pub const LONGLONG = i64;
-pub const ULONG64 = u64;
 pub const HLOCAL = HANDLE;
 pub const LANGID = c_ushort;
 
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -3526,11 +3526,6 @@ pub const UNW_FLAG_EHANDLER = 0x1;
 pub const UNW_FLAG_UHANDLER = 0x2;
 pub const UNW_FLAG_CHAININFO = 0x4;
 
-pub const VECTORED_EXCEPTION_HANDLER = switch (builtin.zig_backend) {
-    .stage1 => fn (ExceptionInfo: *EXCEPTION_POINTERS) callconv(WINAPI) c_long,
-    else => *const fn (ExceptionInfo: *EXCEPTION_POINTERS) callconv(WINAPI) c_long,
-};
-
 pub const OBJECT_ATTRIBUTES = extern struct {
     Length: ULONG,
     RootDirectory: ?HANDLE,
--- a/lib/std/os/windows.zig
+++ b/lib/std/os/windows.zig
@@ -31,6 +31,8 @@ pub const winmm = @import("windows/winmm.zig");
 
 pub const self_process_handle = @intToPtr(HANDLE, maxInt(usize));
 
+const Self = @This();
+
 pub const OpenError = error{
     IsDir,
     NotDir,
@@ -2150,6 +2152,7 @@ pub const LONG = i32;
 pub const ULONG64 = u64;
 pub const ULONGLONG = u64;
 pub const LONGLONG = i64;
+pub const ULONG64 = u64;
 pub const HLOCAL = HANDLE;
 pub const LANGID = c_ushort;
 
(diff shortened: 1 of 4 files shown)


lib/compiler_rt/stack_probe.zig
```diff
@@ -22,9 +22,7 @@ comptime {
             // This symbols are otherwise exported by MSVCRT.lib
             @export(_chkstk, .{ .name = "_chkstk", .linkage = strong_linkage });
             @export(__chkstk, .{ .name = "__chkstk", .linkage = strong_linkage });
-        }
-
-        if (arch.isAARCH64()) {
```
kubkon, line 27: Good catch!

kubkon (changes requested):
Thanks for investigating and the PR @kcbanner - this is some great work! Before we commit to using `kernel32` calls in `std.debug` however, might I suggest we first try re-implementing `RtlCaptureStackBackTrace` directly in Zig (as part of `std.os.windows` module)? I'm not saying this will necessarily also work but it might shed more light on where the spurious failures are happening and would give us the ability to tweak the code as we please. For reference, here's the implementation of [`RtlCaptureStackBackTrace`](https://doxygen.reactos.org/dc/d38/sdk_2lib_2rtl_2exception_8c.html#aa116db3fcbf423c5c6ba9c2fbdd63c39) and [`RtlWalkFrameChain`](https://doxygen.reactos.org/dd/df8/dll_2ntdll_2rtl_2libsupp_8c.html#a703c006cc016ffffe4d2982c373711d4) in ReactOS which I recommend using as a guide.

kubkon:
When it comes to testing, currently we don't support running stage2 incremental linking tests on Windows, so you will need to pass `-Dskip-stage2-tests` to your test run. I'm in the process of finishing up enough of the in-house COFF/PE linker that soon this will not be needed so stay tuned!

kcbanner:
Thanks for the feedback! I'll take a look at those ReactOS implementations

lib/std/os/windows.zig
```diff
@@ -3222,12 +3244,60 @@ pub usingnamespace switch (native_arch) {
...
+            pub fn setIp(ctx: *CONTEXT, ip: usize) void {
+                ctx.Pc = ip;
+            }
+
+            pub fn setSp(ctx: *CONTEXT, sp: usize) void {
+                ctx.Sp = sp;
+            }
+        };
+
+        pub const RUNTIME_FUNCTION = struct {
```
kubkon, line 3264: Should this not be an `extern struct`?

lib/std/os/windows.zig
```diff
@@ -3286,6 +3386,21 @@ pub const TEB = extern struct {
     TlsExpansionSlots: PVOID,
 };
 
+pub const EXCEPTION_REGISTRATION_RECORD = struct {
```
kubkon, line 3389: Ditto, `extern struct`?

kubkon (changes requested):

kcbanner:
Ah, good catches - fixed. I'll work on the proposal we talked about for the COFF unwind info once this is merged 👍

kcbanner:
I believe something has regressed here with LLVM 15. I was testing this against latest master and the 64 bit path is broken both on master (old method) and with these changes. On this branch, `RtlLookupFunctionEntry` returns null. 

Using a zig re-implementation of `RtlVirtualUnwind` that I've been working on, I can see that the `coff.DirectoryEntry.EXCEPTION` directory has a size of zero. This directory in the COFF is supposed to contain entries which have the start/end address of each function, so that you can map an IP back to a function (this is what `RtlLookupFunctionEntry` does). 

My guess is something changed with the LLVM 15 update that caused the linker not to include this directory anymore (and this is a debug build). In LLV14, this section was present.

Regardless, this PR isn't ready to merge until I figure out why this directory isn't there anymore.

kcbanner:
This is confirmed by `dumpbin`:

The LLVM15 compiled version has a missing .pdata section (which corresponds to the exception directory entry, which is also 0).

exe compiled in debug mode with with zig that used LLVM 14.0.6:
```
Dump of file comments.exe

PE signature found

File Type: EXECUTABLE IMAGE

FILE HEADER VALUES
            8664 machine (x64)
               7 number of sections
        630B005A time date stamp Sun Aug 28 01:42:50 2022
               0 file pointer to symbol table
               0 number of symbols
              F0 size of optional header
              22 characteristics
                   Executable
                   Application can handle large (>2GB) addresses

OPTIONAL HEADER VALUES
             20B magic # (PE32+)
           14.00 linker version
           72C00 size of code
           44200 size of initialized data
               0 size of uninitialized data
            1230 entry point (0000000140001230) wWinMainCRTStartup
            1000 base of code
       140000000 image base (0000000140000000 to 00000001400BDFFF)
            1000 section alignment
             200 file alignment
            6.00 operating system version
            0.00 image version
            6.00 subsystem version
               0 Win32 version
           BE000 size of image
             400 size of headers
               0 checksum
               3 subsystem (Windows CUI)
            8160 DLL characteristics
                   High Entropy Virtual Addresses
                   Dynamic base
                   NX compatible
                   Terminal Server Aware
         1000000 size of stack reserve
            1000 size of stack commit
          100000 size of heap reserve
            1000 size of heap commit
               0 loader flags
              10 number of directories
               0 [       0] RVA [size] of Export Directory
           B2598 [      3C] RVA [size] of Import Directory
               0 [       0] RVA [size] of Resource Directory
           B8000 [    2610] RVA [size] of Exception Directory
               0 [       0] RVA [size] of Certificates Directory
           BD000 [     170] RVA [size] of Base Relocation Directory
           B2538 [      1C] RVA [size] of Debug Directory
               0 [       0] RVA [size] of Architecture Directory
               0 [       0] RVA [size] of Global Pointer Directory
           B2510 [      28] RVA [size] of Thread Storage Directory
               0 [       0] RVA [size] of Load Configuration Directory
               0 [       0] RVA [size] of Bound Import Directory
           B26E0 [     108] RVA [size] of Import Address Table Directory
               0 [       0] RVA [size] of Delay Import Directory
               0 [       0] RVA [size] of COM Descriptor Directory
               0 [       0] RVA [size] of Reserved Directory
```

exe compiled with latest zig master, LLVM 15.0.0:
```
Dump of file comments_llvm15.exe

PE signature found

File Type: EXECUTABLE IMAGE

FILE HEADER VALUES
            8664 machine (x64)
               6 number of sections
        632FCC2B time date stamp Sat Sep 24 23:34:03 2022
               0 file pointer to symbol table
               0 number of symbols
              F0 size of optional header
              22 characteristics
                   Executable
                   Application can handle large (>2GB) addresses

OPTIONAL HEADER VALUES
             20B magic # (PE32+)
           14.00 linker version
           73800 size of code
           43600 size of initialized data
               0 size of uninitialized data
            1280 entry point (0000000140001280) wWinMainCRTStartup
            1000 base of code
       140000000 image base (0000000140000000 to 00000001400BCFFF)
            1000 section alignment
             200 file alignment
            6.00 operating system version
            0.00 image version
            6.00 subsystem version
               0 Win32 version
           BD000 size of image
             400 size of headers
               0 checksum
               3 subsystem (Windows CUI)
            8160 DLL characteristics
                   High Entropy Virtual Addresses
                   Dynamic base
                   NX compatible
                   Terminal Server Aware
         1000000 size of stack reserve
            1000 size of stack commit
          100000 size of heap reserve
            1000 size of heap commit
               0 loader flags
              10 number of directories
               0 [       0] RVA [size] of Export Directory
           B774F [      3C] RVA [size] of Import Directory
               0 [       0] RVA [size] of Resource Directory
               0 [       0] RVA [size] of Exception Directory
               0 [       0] RVA [size] of Certificates Directory
           BC000 [     18C] RVA [size] of Base Relocation Directory
           B76E8 [      1C] RVA [size] of Debug Directory
               0 [       0] RVA [size] of Architecture Directory
               0 [       0] RVA [size] of Global Pointer Directory
           B76C0 [      28] RVA [size] of Thread Storage Directory
               0 [       0] RVA [size] of Load Configuration Directory
               0 [       0] RVA [size] of Bound Import Directory
           B78A8 [     118] RVA [size] of Import Address Table Directory
               0 [       0] RVA [size] of Delay Import Directory
               0 [       0] RVA [size] of COM Descriptor Directory
               0 [       0] RVA [size] of Reserved Directory
```

kcbanner:
The regression above is fixed with https://github.com/ziglang/zig/pull/12959

kubkon:
First of, I'd like to apologise for taking so long to take another look at your PR! In general, I believe we are ready to merge except the merge conflicts, and so, would you mind rebasing on latest master so that we can follow up with a merge?

kcbanner:
No problem, I'll get this updated soon 👍

kcbanner:
Ok, rebased onto master 👍

andrewrk (changes requested):
Thank you for taking this on.

Is this better than status quo? I get that it's a workaround that seems to have worked for somebody on the Internet, but what the hell is going on here? I think we need to have a well-documented justification for this. It seems like this has not been investigated thoroughly to me.

If `RtlVirtualUnwind` is a lower level call, then this starts to makes sense to me, especially if `RtlCaptureStackBackTrace` is internally using it in its implementation. Perhaps we could look at ReactOS source code to help determine this.

The main requested changes I have here is regarding #1840. This PR is taking things in the wrong direction.

kcbanner:
There is some more context in this thread around why I opted to use the kernel32 call: https://discord.com/channels/605571803288698900/1017808096682311740/1018889803128918056 

Essentially, I did write an implementation of `RtlVirtualUnwind` in zig, that parses all the unwind table opcodes from the unwind info to figure out how to walk back up the stack. I referenced the ReactOS implementation as part of that. I managed to implement that completely without any syscalls, *except* for a critical section call to lock around the list of loaded modules from the PEB.

The implementation (WIP: working on x64, but not cleaned up): https://gist.github.com/kcbanner/61058fba639fca1a6bda13773489a484

So while that does work, you still have to make the kernel32 call, otherwise you risk reading the list while it's being changed. Other than that, it can walk the stack completely without any syscalls. If we are ok with just not locking that critical section, then this could be done entirely in zig without syscalls. There would be potential for crashes if the module list is modified while it's being iterated though. With that in mind (and the use of a bunch of undocumented data structures), I stopped with the custom implementation since we needed a kernel32 call anyway.

I'm happy to explore that further, but we'd still need that kernel32 call to grab the lock. If we ignore the lock then we can't safely iterate that linked list, and if we cache results then there is a TOCTOU issue.

From what I could tell the ReactOS version of `RtlCaptureStackBackTrace` uses `RtlWalkFrameChain` (https://doxygen.reactos.org/dd/df8/dll_2ntdll_2rtl_2libsupp_8c.html#a703c006cc016ffffe4d2982c373711d4), which just uses the frame pointer to walk the state. So, that works on x86 and on x64 if there was no extra stack space allocated. However, if there was - then you need to use the unwind info to figure out what the prologue did in order to reverse it.

andrewrk:
Thank you for this writeup. I appreciate your patience with me here.

In NtDll.dll, I see the following functions:
 * `RtlCaptureContext`
 * `RtlCaptureContext2`
 * `RtlLookupFunctionEntry`
 * `RtlVirtualUnwind`
 * `RtlVirtualUnwind2`
 * `RtlEnterCriticalSection`
 * `RtlLeaveCriticalSection`

[more](https://github.com/ziglang/zig/blob/master/lib/libc/mingw/lib-common/ntdll.def.in)

After reading your writeup, and making note of these functions, I'm in favor of your changes, but I still want to emphasize #1840. Our goal is to transition to NtDll being the only dependency of the standard library. It seems compatible with your changes. Does that sound reasonable to you?

kcbanner:
No problem, and yes that make sense to me!

I actually had misunderstood and didn't realize kernel32 wrapped ntdll, I thought they were independent. Reading https://github.com/ziglang/zig/issues/1840 now everything makes more sense.

kcbanner:
Updated to call via ntdll instead of kernel32.

Should I keep the definitions of these functions in kernel32.zig for users to call, or I remove them now that they exist in ntdll.zig as well?

andrewrk:
> Should I keep the definitions of these functions in kernel32.zig for users to call, or I remove them now that they exist in ntdll.zig as well?

This question can be left for #4426 to solve

andrewrk:
Great work! Thanks for the follow ups.

PR #12741, opened 2022-09-04
cmake msvc: fixup compiler arg generation, as /O2 is incompatible with /RTC1

kcbanner:
When trying to make a debug build I was hitting "/O2' and '/RTC1' command-line options are incompatible". This change only adds `/O2` on MSVC when not building "Debug".

Build process:
```
mkdir build-debug
pushd build-debug
"c:\Program Files\CMake\bin\cmake.exe" .. -Thost=x64 -G "Visual Studio 16 2019" -A x64 -DCMAKE_PREFIX_PATH=e:\dev\llvm+clang+lld-14.0.6-x86_64-windows-msvc-debug-mtd -DCMAKE_BUILD_TYPE=Debug
msbuild -p:Configuration=Debug INSTALL.vcxproj
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -309,8 +309,14 @@ set(EMBEDDED_SOFTFLOAT_SOURCES
 )
 add_library(embedded_softfloat STATIC ${EMBEDDED_SOFTFLOAT_SOURCES})
 if(MSVC)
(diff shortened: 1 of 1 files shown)


andrewrk:
Thanks!

PR #12760, opened 2022-09-06
std.mem.trim*: Allow mutable slices

ominitay:
This PR changes `std.mem.trimLeft`, `...trimRight`, and `...trim` to allow the use of both mutable and non-mutable slices. Previously, either a dangerous `intToPtr` was needed, or the slice's data had to be manually copied into a mutable slice. This change means that mutable slices now "just work", and doesn't break existing use.

### DIFF
--- a/lib/std/mem.zig
+++ b/lib/std/mem.zig
@@ -972,21 +972,36 @@ pub fn allEqual(comptime T: type, slice: []const T, scalar: T) bool {
 }
 
 /// Remove values from the beginning of a slice.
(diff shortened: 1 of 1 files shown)


ominitay:
There are a few changes left to be made here. Pointer alignment must be considered -- the function should return an naturally-aligned slice. Additionally, sentinel-terminated slices should be disallowed, at least for `trim` and `trimRight`, or the sentinel should be removed from the return type.

andrewrk:
Closing old draft

PR #12763, opened 2022-09-06
cmake: remove incompatible c flags for MSVC 

LawG4:
When building Zig0 with msvc you get an error that /O2 and /RTL1 are incompatible. C Flags for the project for msvc are set in cmake/c_flag_overrides.cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -842,9 +842,7 @@ else()
 endif()
 
 if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
-    if(MSVC)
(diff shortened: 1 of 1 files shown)


kcbanner:
I made a similar PR here: https://github.com/ziglang/zig/pull/12741

I did notice that ironically the `/w` flag generates a warning though: `Command line warning D9025 : overriding '/W3' with '/w'`

LawG4:
@kcbanner Oh yeah! I've just noticed that, great minds think alike it seems

andrewrk:
Closing in favor of #12741

Issue #12555, opened 2022-08-21
regression 0.10.0-dev.3659+e5e6eb983 vs. 0.10.0-dev.3431+4a4f3c50c


mz0:
### Zig Version

0.10.0-dev.3659+e5e6eb983

### Steps to Reproduce

this fragment from `ziglings/exercises/063_labels.zig`
```
if (!for (wanted_ingredients) |want_it| {
    if (required_ingredient == want_it) break true;
} else false) continue :food_loop;
```
compiles fine with 0.10.0-dev.3431+4a4f3c50c and with a number of previous builds
but not in 0.10.0-dev.3659+e5e6eb983
The full file is https://github.com/mz0/ziglings/blob/done/exercises/063_labels.zig

### Expected Behavior

no errors

### Actual Behavior

```
Compiling 063_labels.zig...
/home/mz0/z/ziglings/exercises/063_labels.zig:118:53: error: expected type 'bool', found 'void'
            if (!for (wanted_ingredients) |want_it| {
                                                    ^
```

Vexu:
Related #12551, #11816

Issue #12551, opened 2022-08-21
stage3: loop expression's else branch sometimes not evaluated


travisstaloch:
### Zig Version

0.10.0-dev.3659+e5e6eb983

### Steps to Reproduce

```zig
// test.zig
const std = @import("std");
test {
    try std.testing.expect(for ([1]u8{0}) |x| {
        if (x == 0) break true;
    } else false);
}
```
```console
zig test test.zig
```

### Expected Behavior

The test should compile and pass as this worked with stage1.

Also, this works with stage3 when assigned to a decl: 
```zig
test {
    const found = for ([1]u8{0}) |x| {
        if (x == 0) break true;
    } else false;
    try std.testing.expect(found);
}
```

### Actual Behavior

```console
$ zig test test.zig 
test.zig:4:47: error: expected type 'bool', found 'void'
    try std.testing.expect(for ([1]u8{0}) |x| {
                                              ^
```
The error seems to suggest the for loop's else branch isn't being used within a boolean context.

Issue #12455, opened 2022-08-16
break from switch in loop bypasses unused value check


Vexu:
### Zig Version

0.10.0-dev.3577+b3922289b

### Steps to Reproduce

```zig
test {
    var a: u32 = 0;
    while (true) switch (a) {
        0 => 2,
        1 => a = 0,
        else => break,
    };
}
```

### Expected Behavior

A compile error complaining about types or unused value.

### Actual Behavior

Successful compile and an infinite loop.

nektro:
behavior makes sense to me

Vexu:
It's a foot gun and an error in stage1 and shouldn't be allowed

nektro:
the code never hits the `else` thats why its looping. `a` alternates between being `0` and `1`. im lost on what the bug actually is here since the expected behavior written is so vague

Vexu:
Typo in the example, `1` was supposed to be `2`, which was being ignored instead of being assigned to `a` causing the infinite loop.

nektro:
stage1:

```
./test.zig:4:14: error: integer value 2 cannot be coerced to type 'void'
        0 => 2,
             ^
```

kkHAIKE:
relate https://github.com/ziglang/zig/issues/11816

Vexu:
It is a bug and the original example should cause a compile error.

PR #12764, opened 2022-09-07
AstGen: add ResultLoc.only_break to fix loop's then-scope result

kkHAIKE:
fix https://github.com/ziglang/zig/issues/12555 https://github.com/ziglang/zig/issues/12551 https://github.com/ziglang/zig/issues/12455

loop body is noreturn and only the `break` operands are the result values of loops. the new locType 'only_break' mean only effect break result,
otherwise it's work like 'none'.

ralate commit: https://github.com/ziglang/zig/commit/70894d5c2f

prev msg:
~while/for expr use func `markAsLoopBody` to set `rvalue_noresult` field for help func `rvalue` to skip store result in `then_scope`. and blockExpr default return void_value when block end_noreturn.~

~in this case, func `rvalue` add a `as_node` for `rl` if it is `ty` case, then got `as_node(rl, void_value)`cause the bug.~

~so need add `rvalue_noresult` case include `ty rl`~

PS: zig is great project.

### DIFF
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -5974,8 +5974,7 @@ fn whileExpr(
     if (dbg_var_name) |some| {
         try then_scope.addDbgVar(.dbg_var_val, some, dbg_var_inst);
     }
-    // rl == .discard may ignore some compile error.
-    const then_rl = .{ .only_break = ResultLoc.wrap(if (loop_scope.break_result_loc == .discard) .none else loop_scope.break_result_loc) };
+    const then_rl = .{ .only_break = ResultLoc.wrap(loop_scope.break_result_loc) };
     const then_result = try expr(&then_scope, then_sub_scope, then_rl, while_full.ast.then_expr);
     _ = try addEnsureResult(&then_scope, then_result, while_full.ast.then_expr);
 
@@ -6215,8 +6214,7 @@ fn forExpr(
         break :blk &index_scope.base;
     };
 
-    // rl == .discard may ignore some compile error.
-    const then_rl = .{ .only_break = ResultLoc.wrap(if (loop_scope.break_result_loc == .discard) .none else loop_scope.break_result_loc) };
+    const then_rl = .{ .only_break = ResultLoc.wrap(loop_scope.break_result_loc) };
     const then_result = try expr(&then_scope, then_sub_scope, then_rl, for_full.ast.then_expr);
     _ = try addEnsureResult(&then_scope, then_result, for_full.ast.then_expr);
 
--- a/src/AstGen.zig
+++ b/src/AstGen.zig
@@ -214,20 +214,7 @@ pub fn deinit(astgen: *AstGen, gpa: Allocator) void {
     astgen.ref_table.deinit(gpa);
 }
 
-pub const ResultLocTag = enum {
-    discard,
-    none,
-    ref,
-    ty,
-    ty_shift_operand,
-    coerced_ty,
-    ptr,
-    inferred_ptr,
-    block_ptr,
-    only_break,
-};
-
-pub const ResultLoc = union(ResultLocTag) {
+pub const ResultLoc = union(enum) {
     /// The expression is the right-hand side of assignment to `_`. Only the side-effects of the
     /// expression should be generated. The result instruction from the expression must
     /// be ignored.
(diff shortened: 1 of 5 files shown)


kkHAIKE:
oh no, other bad case..
```
xxx(while (gen()) |x| b: {
    if (x == 0) break :b true;
} else false);
```

Vexu (changes requested):
Needs a behavior test.

Vexu:
I don't know how that should be solved, I'd have to fix it myself to be able to help.

kkHAIKE:
@Vexu done. 😊

src/AstGen.zig
```diff
@@ -214,7 +214,20 @@ pub fn deinit(astgen: *AstGen, gpa: Allocator) void {
     astgen.ref_table.deinit(gpa);
 }
 
-pub const ResultLoc = union(enum) {
+pub const ResultLocTag = enum {
```
Vexu, line 217:
```suggestion
pub const ResultLocTag = std.meta.Tag(ResultLoc);
```
I don't want this kind of repetition.

src/AstGen.zig
```diff
@@ -242,6 +255,9 @@ pub const ResultLoc = union(enum) {
     /// is inferred based on peer type resolution for a `Zir.Inst.Block`.
     /// The result instruction from the expression must be ignored.
     block_ptr: *GenZir,
+    /// Use in then-scope of loop, only work for 'break' instruction,
+    /// otherwise it's work like none.
```
Vexu, line 259: I don't understand what this does based on this comment. And why is it only in the then-scope?

test/compile_errors.zig
```diff
@@ -225,6 +225,24 @@ pub fn addCases(ctx: *TestContext) !void {
         });
     }
 
+    {
+        const case = ctx.obj("issue 12455 while with switch", .{});
```
Vexu, line 229: I'd prefer this be in a separate file like most other compile error tests.

test/behavior.zig
```diff
@@ -160,6 +160,7 @@ test {
     _ = @import("behavior/void.zig");
     _ = @import("behavior/while.zig");
     _ = @import("behavior/widening.zig");
+    _ = @import("behavior/bugs/11816_12551_12555.zig");
```
Vexu, line 163: Please put this with the other `/bugs/`.

src/AstGen.zig
```diff
@@ -1884,15 +1933,25 @@ fn breakExpr(parent_gz: *GenZir, parent_scope: *Scope, node: Ast.Node.Index) Inn
...
-                parent_gz.rvalue_noresult = .none;
+                // unwrap if break_result_loc is only_break
+                const prev_break_result_loc = block_gz.break_result_loc;
+                const prev_rl_ptr = block_gz.rl_ptr;
+                const is_only_break = prev_break_result_loc == .only_break;
+                if (is_only_break) {
+                    switch (prev_break_result_loc) {
+                        .only_break => |w| block_gz.setBreakResultLoc(w.unwrap()),
+                        else => unreachable,
+                    }
```
Vexu, line 1944:
```suggestion
                    block_gz.setBreakResultLoc(prev_break_result_loc.only_break.unwrap());
```

Vexu (changes requested):

src/AstGen.zig
```diff
@@ -242,6 +242,11 @@ pub const ResultLoc = union(enum) {
     /// is inferred based on peer type resolution for a `Zir.Inst.Block`.
     /// The result instruction from the expression must be ignored.
     block_ptr: *GenZir,
+    /// Used to wrap other ResultLoc, unwrap and do the right thing when in break expression,
+    /// other case this type same as `none`.
+    /// Now this type use in loop body only, because when in reality the loop body is noreturn
+    /// and only the `break` operands are the result values of loops.
```
Vexu, line 248:
```suggestion
    /// Used for loop bodies in which the result location only applies to
    /// the operands of break expressions.
```

src/AstGen.zig
```diff
@@ -1884,15 +1922,22 @@ fn breakExpr(parent_gz: *GenZir, parent_scope: *Scope, node: Ast.Node.Index) Inn
...
+                const is_only_break = prev_break_result_loc == .only_break;
+                if (is_only_break) {
+                    block_gz.setBreakResultLoc(prev_break_result_loc.only_break.unwrap());
+                }
+                defer {
+                    if (is_only_break) {
+                        block_gz.setBreakResultLoc(prev_break_result_loc);
+                        block_gz.rl_ptr = prev_rl_ptr;
+                    }
+                }
```
Vexu, line 1937: I think modifying the block's break result loc might lead to incorrect behavior with nested breaks like:
```zig
break :loop {
    if (cond) break :loop something;
}
```
Assigning `block_gz.break_result_loc` to a variable and then unwrapping if needed should avoid any issues.

Vexu (changes requested):

kkHAIKE:
@Vexu OMG, i take quite a long way round. i make new PR for the new simple way.https://github.com/ziglang/zig/pull/12850

i'm so sorry for wasting your time.

PR #12770, opened 2022-09-07
wasm-linker: implement linking debug-information

Luukdegram:
With those changes, the in-house wasm linker is now capable of:
- Parsing debug information from object/archive files.
- Linking debug information of several object files into a final binary. (e.g. when Zig is invoked as linker).
- Linking debug information between object files and zig-code. (e.g. when `build-exe` or `build-lib` is invoked).
- Performing relocations on debug information.
- Support linking debug information while in incremental mode.

Still much more to do, but this is a very big step to replace wasm-ld with our own linker.


williamstein:
> Still much more to do, but this is a very big step to replace wasm-ld with our own linker.

Are you planning to implement support for -fPIC code like the wasm32-emscripten LLVM target supports?  In zig, this is what one gets with `zig wasm-ld --experimental-pic -shared`.   It's actually already possible now with zig to build shared wasm extensions and import them in nodejs via this library I wrote:  https://www.npmjs.com/package/dylink     The README for the dylink package has references to docs about how `-fPIC` works for webassembly, at least for emscripten which is the only existing implementation.  It's possible to use in zig because zig ships llvm, of course.

Luukdegram:
> > Still much more to do, but this is a very big step to replace wasm-ld with our own linker.
> 
> 
> 
> Are you planning to implement support for -fPIC code like the wasm32-emscripten LLVM target supports?  In zig, this is what one gets with `zig wasm-ld --experimental-pic -shared`.   It's actually already possible now with zig to build shared wasm extensions and import them in nodejs via this library I wrote:  https://www.npmjs.com/package/dylink     The README for the dylink package has references to docs about how `-fPIC` works for webassembly, at least for emscripten which is the only existing implementation.  It's possible to use in zig because zig ships llvm, of course.   

Yes, I'm very familiar with how it works. However, I don't consider an experimental feature a requirement to make our own linker the default. We can still fall back to wasm-ld for those that need this use case until we get to implementing this in our own linker. The goal is feature parity, but stable features have priority over experimental ones.

To to answer your question in short: Yes, I'm planning to support this (eventually).

williamstein:
> So to answer your question in short: Yes, I'm planning to support this (eventually).

AWESOME!!   I agree that this isn't needed for switching over the default, since it will still be possible to somehow still call the llvm wasm-ld.

Thanks for everything you're doing.  I greatly appreciate it.

PR #12771, opened 2022-09-07
std.os.linux: add ioctlFunc

marler8997:
Adds `ioctlFunc` which returns a wrapper function whose parameter type is tied to the ioctl definition.

I'm not certain whether this belong in std or not so feel free to close if not.  I've created a gist that provides this as a module and has some background info on it: https://gist.github.com/marler8997/077a43569a2eda4dc1d4bdc3e44c2d46


lib/std/os/linux.zig
```diff
@@ -5597,3 +5600,72 @@ pub const AUDIT = struct {
...
+/// Linux encodes the size of the ioctl argument and its IO direction within the request number.
+/// `ioctlFunc` returns a function with the same argument type that was used to create the
+/// request number.  By preserving this relationship, the kernel will always have the correct
+/// IO direction and argument size, meaning it should never result in out-of-bounds memory access.
+pub fn ioctlFunc(
+    comptime dir: IoctlDirection,
+    comptime io_type: u8,
+    comptime nr: u8,
+    comptime Arg: type,
+) ioctlInfo(dir, io_type, nr, Arg).Fn {
```
Vexu, line 5622: This should return a function literal instead of a function pointer.

Vexu (approved):

lib/std/os/linux.zig
marler8997: done

marler8997:
I had a look at the linux kernel headers to see how it uses this ioctl "type safety mechanism".  It looks like there are indeed older ioctls that existed before this mechanism that don't use it.  Furthermore, sometimes these ioctls use the old request number for some systems but use the new mechanism for other systems.  For example, `TIOCGWINSZ` is defined as `0x5413` by default, but for the systems sh, xtensa, sparc, powerpc, mips and alpha it uses the new mechanism `_IOR('t', 104, struct winsize)`.

To address this I've added the function `legacyIoctl` which will take the ioctl type/number (`'t'` and `104` in the case of `TIOCGWINSZ`) and return the legacy ioctl number if it has one for the target platform.

This should work for most ioctls but there is an extra complication.  It appears that there are some ioctls that use the new mechanism, but, their "new definition" can vary depending on the platform.  `TCGETS` is an example of this, i.e.
```c
// uapi, xtensa, sh
0x5401
// mips
0x540d
// sparc
_IOR('T', 8, struct termios)
// parisc
_IOR('T', 16, struct termios)
// powerpc, alpha
_IOR('t', 19, struct termios)
```

Note that if you get the ioctl type/number wrong, you're still protected from bad memory access (that's the advantage of the kernel defining this ioctl type-safety mechanism), but I bring up this complication so we can have the full picture when considering an API for this.  The new API takes the ioctl type/number, but, now this type/number can change depending on the platform.  So, we'd have to handle the platform outside of `ioctlInfo` rather than inside of it, i.e.
```zig
const ioctl_tcgets =
    if (is_sparc) ioctlInfo(.read, 'T', 8, termios)
    else if (is_parisc) ioctlInfo(.read, 'T', 16, termios)
    else if (is_ppc or is_alpha) ioctlInfo(.read, 't', 19, termios)
    else if (is_mips) 0x540d
    else 0x5401;
```

Maybe this is "good enough", but maybe there is a better/simpler api, not sure?

andrewrk (changes requested):
I'm open to having this in the standard library but some changes are needed first:

 * I don't think that `std.os.ioctlFunc` is the proper namespace for this function. Let's try to think of a better home for it. This namespace is supposed to look roughly like the posix functions namespace, or at least the names that would be in that abstraction layer if os-specific functions were added to posix.
 * The standalone test now needs to be updated to work with the new standalone test harness (which uses anonymous sub-packages via the build system API).

PR #12788, opened 2022-09-09
std.zig.system.NativeTargetInfo: improve glibc version and dynamic linker detection

### LINKED ISSUE #6469: zig detects wrong libc version

andrewrk:
Previously, this code would fail to detect glibc version because it
relied on libc.so.6 being a symlink which revealed the answer. On modern
distro versions, this is no longer the case.

This new strategy finds the path to libc.so.6 from `/usr/bin/env`, then
inspects the `.dynstr` section of libc.so.6, looking for symbols that
start with "GLIBC_2.". It then parses those as semantic versions and
takes the maximum value as the system-native glibc version.

-----

Before, native glibc and dynamic linker detection attempted to use the
executable's own binary if it was dynamically linked to answer both the
C ABI question and the dynamic linker question. However, this could be
problematic on a system that uses a RUNPATH for the compiler binary,
locking it to an older glibc version, while system binaries such as
/usr/bin/env use a newer glibc version. The problem is that libc.so.6
glibc version will match that of the system while the dynamic linker
will match that of the compiler binary. Executables with these versions
mismatching will fail to run.

Therefore, this pull request changes the logic to be the same regardless of
whether the compiler binary is dynamically or statically linked. It
inspects `/usr/bin/env` as an ELF file to find the answer to these
questions, or if there is a shebang line, then it chases the referenced
file recursively. If that does not provide the answer, then the function
falls back to defaults.

This pull request also solves a TODO to remove an Allocator parameter to the
detect() function.

-----

closes #6469
closes #12567

@Vexu can you check if this solves #11137 for you?

### DIFF
--- a/lib/std/zig/system/NativeTargetInfo.zig
+++ b/lib/std/zig/system/NativeTargetInfo.zig
@@ -552,14 +552,14 @@ fn glibcVerFromSoFile(file: fs.File) !std.builtin.Version {
     // and furthermore, that the system-installed glibc is at minimum that version.
 
     // Empirically, glibc 2.34 libc.so .dynstr section is 32441 bytes on my system.
-    // Here I use this value plus some headroom. This makes it only need
+    // Here I use double this value plus some headroom. This makes it only need
     // a single read syscall here.
-    var buf: [40000]u8 = undefined;
+    var buf: [80000]u8 = undefined;
     if (buf.len < dynstr.size) return error.InvalidGnuLibCVersion;
 
     const dynstr_size = @intCast(usize, dynstr.size);
     const dynstr_bytes = buf[0..dynstr_size];
-    _ = try preadMin(file, dynstr_bytes, dynstr.offset, dynstr_size);
+    _ = try preadMin(file, dynstr_bytes, dynstr.offset, dynstr_bytes.len);
     var it = mem.split(u8, dynstr_bytes, &.{0});
     var max_ver: std.builtin.Version = .{ .major = 2, .minor = 2, .patch = 5 };
     while (it.next()) |s| {
--- a/lib/std/zig/system/NativeTargetInfo.zig
+++ b/lib/std/zig/system/NativeTargetInfo.zig
@@ -820,55 +820,87 @@ pub fn abiAndDynamicLinkerFromFile(
                 while (it.next()) |rpath| {
                     if (glibcVerFromRPath(rpath)) |ver| {
                         result.target.os.version_range.linux.glibc = ver;
-                        break;
+                        return result;
                     } else |err| switch (err) {
                         error.GLibCNotFound => continue,
                         else => |e| return e,
                     }
                 }
-            } else if (result.dynamic_linker.get()) |dl_path| glibc_ver: {
-                // There is no DT_RUNPATH so we try to find libc.so.6 inside the same
-                // directory as the dynamic linker.
-                if (fs.path.dirname(dl_path)) |rpath| {
-                    if (glibcVerFromRPath(rpath)) |ver| {
-                        result.target.os.version_range.linux.glibc = ver;
-                        break :glibc_ver;
-                    } else |err| switch (err) {
-                        error.GLibCNotFound => {},
-                        else => |e| return e,
-                    }
-                }
-
-                // So far, no luck. Next we try to see if the information is
-                // present in the symlink data for the dynamic linker path.
-                var link_buf: [std.os.PATH_MAX]u8 = undefined;
-                const link_name = std.os.readlink(dl_path, &link_buf) catch |err| switch (err) {
-                    error.NameTooLong => unreachable,
-                    error.InvalidUtf8 => unreachable, // Windows only
-                    error.BadPathName => unreachable, // Windows only
-                    error.UnsupportedReparsePointType => unreachable, // Windows only
-
-                    error.AccessDenied,
-                    error.FileNotFound,
-                    error.NotLink,
-                    error.NotDir,
-                    => break :glibc_ver,
-
-                    error.SystemResources,
-                    error.FileSystem,
-                    error.SymLinkLoop,
-                    error.Unexpected,
-                    => |e| return e,
-                };
-                result.target.os.version_range.linux.glibc = glibcVerFromLinkName(
-                    fs.path.basename(link_name),
-                    "ld-",
-                ) catch |err| switch (err) {
-                    error.UnrecognizedGnuLibCFileName,
-                    error.InvalidGnuLibCVersion,
-                    => break :glibc_ver,
-                };
             }
+        }
+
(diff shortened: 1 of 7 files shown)


daurnimator:
> inspects the `.dynstr` section of libc.so.6, looking for symbols that
> start with "GLIBC_2."  It then parses those as semantic versions and
takes the maximum value as the system-native glibc version.

At least on my Archlinux system, this doesn't give the correct answer.
The highest version is `GLIBC_2.34` for `__libc_start_main`, however the system glibc version is `2.36`.

BratishkaErik:
Doesn't work on my Gentoo system :( rpath_offset is null

andrewrk:
@BratishkaErik thank you for checking. I will poke around on Gentoo in a Docker container and then push another commit for you to try shortly.

lib/std/zig/system/NativeTargetInfo.zig
```diff
@@ -455,23 +402,109 @@ fn detectAbiAndDynamicLinker(
...
+    } else return error.InvalidGnuLibCVersion;
+
+    // Here we loop over all the strings in the dynstr string table, assuming that any
+    // strings that start with "GLIBC_2." indicate the existence of such a glibc version,
+    // and furthermore, that the system-installed glibc is at minimum that version.
+
+    // Empirically, glibc 2.34 libc.so .dynstr section is 32441 bytes on my system.
+    // Here I use this value plus some headroom. This makes it only need
+    // a single read syscall here.
+    var buf: [40000]u8 = undefined;
```
topolarity, line 485: It seems like if we just wait for enough glibc releases this buffer will be overrun eventually, since new glibc versions will essentially only ever add new symbols

As a datapoint, `.dynstr` on my system glibc 2.34 is 35062 bytes and 2.12 is 22098 bytes

andrewrk: Fair point, I will update this to account for any size.

andrewrk:
@BratishkaErik I have pushed more commits that use an additional heuristic that solves Gentoo. It looks like unfortunately a future enhancement may require a full integration with `/etc/ld.so.cache`, however every distro I tested, these heuristics gave the correct answer. Tested distros include: NixOS, Gentoo, Archlinux, Debian

BratishkaErik:
> @BratishkaErik I have pushed more commits that use an additional heuristic that solves Gentoo. It looks like unfortunately a future enhancement may require a full integration with `/etc/ld.so.cache`, however every distro I tested, these heuristics gave the correct answer. Tested distros include: NixOS, Gentoo, Archlinux, Debian

Yes. it works fine! Thank you!
```zig
pub const os = std.Target.Os{
    .tag = .linux,
    .version_range = .{ .linux = .{
        .range = .{
            .min = .{
                .major = 5,
                .minor = 19,
                .patch = 6,
            },
            .max = .{
                .major = 5,
                .minor = 19,
                .patch = 6,
            },
        },
        .glibc = .{
            .major = 2,
            .minor = 35,
            .patch = 0,
        },
    }},
};
```

topolarity:
> a future enhancement may require a full integration with `/etc/ld.so.cache`, however every distro I tested, these heuristics gave the correct answer. Tested distros include: NixOS, Gentoo, Archlinux, Debian

I'm seeing bad results on Debian bookworm (testing)

It looks like the latest glibc updates removed the versioned symlinks for Debian too, unfortunately

andrewrk:
Alright another commit pushed, now Debian bookworm is handled. If anyone wants to test the logic without compiling anything, you can do this:

```
$ wget https://andrewkelley.me/temp/test4
$ chmod +x test4
$ ./test4
ver: 2.34 dl: /lib64/ld-linux-x86-64.so.2
```

Any more distros not working still?

Vexu:
Can confirm that this detects the correct glibc version for me and I can build the example in #11137 if I explicitly specify an earlier version that Zig can provide.

andrewrk:
The failure affects master branch. I will address it in master branch by adding more debug printing to help when the problem happens again.

PR #12798, opened 2022-09-09
std: add peek function to std.json.TokenStream

### LINKED ISSUE #12463: std: add `peek` function to `std.json.TokenStream`
This peek function would return the next token, but without modifying the `TokenStream`'s state in any way.

adarah:
First time contributing, hope this PR is up to your standards.

Similarly to #11925, this `peek` implementation technically does modify the internal `i` counter to speed up subsequent `peek` calls, but it shouldn't have any other impact. This is useful because `next` has been refactored to invoke `peek` internally, so the common case of invoking `peek` followed by `next` will still be fast. If this approach is not ok, then I can refactor `peek` so that subsequent calls repeat the same work.

One thing I'm very uncertain about is the change in the `stackUsed` method (only used in tests), particularly the contribution of `token1`/`token2` to the stack size. My understanding is that the `StreamingParser`'s stack is only used to keep track of opening/closing brackets, and `token2` is only ever populated with the token immediately after a number, which implies "If token2 is not null, then token1 is a number". With that in mind, should we not validate if `token1` and `token2` are brackets/braces rather than just checking for not null?

Closes #12463.

### DIFF
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -415,7 +415,6 @@ pub const StreamingParser = struct {
                     }
 
                     _ = p.stack.pop();
-                    p.state = .ValueBegin;
                     p.after_string_state = State.fromAggregateContainerType(last_type);
 
                     switch (p.stack.len) {
@@ -438,7 +437,6 @@ pub const StreamingParser = struct {
                     }
 
                     _ = p.stack.pop();
-                    p.state = .ValueBegin;
                     p.after_string_state = State.fromAggregateContainerType(last_type);
 
                     switch (p.stack.len) {
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -221,7 +221,7 @@ pub const StreamingParser = struct {
     pub fn reset(p: *StreamingParser) void {
         p.state = .TopLevelBegin;
         p.count = 0;
-        // Set before ever read in main transition function
+        // Set before every read in main transition function
         p.after_string_state = undefined;
         p.after_value_state = .ValueEnd; // handle end of values normally
         p.stack.init();
--- a/lib/std/json.zig
+++ b/lib/std/json.zig
@@ -313,17 +313,17 @@ pub const StreamingParser = struct {
     };
 
     /// Give another byte to the parser and obtain any new tokens. This may (rarely) return two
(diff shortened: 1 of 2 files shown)


lib/std/json.zig
```diff
@@ -1090,7 +1088,8 @@ pub const TokenStream = struct {
     i: usize,
     slice: []const u8,
     parser: StreamingParser,
-    token: ?Token,
+    token1: ?Token,
+    token2: ?Token,
```
Namek, line 1092: I'd consider more descriptive names, e.g. `current_token`, `next_token`

adarah: I've attempted to maintain the same naming convention the [StreamingParser.feed](https://github.com/ziglang/zig/blob/a5b34a61ab61882bf55d87e4cbc8186215ecf320/lib/std/json.zig#L315-L321) used, but I can change both of them if you think that's better. What do you think?

adarah: I've thought about it a little bit and I'll just assume you do want this change (to reduce the back-and-forth overhead). I personally think it's better as well

Namek: I mean, I am absolutely no authority here. Just a random code review from my side since I am interested in having getty/json improved. I suggest finding an author of the code (git blame?) and ping him here to notice the PR. LGTM

Namek (approved):

andrewrk:
I've pinged @daurnimator and @tiehuis to request a review. Sorry if the ping is unwanted; you two seem like the code owners of std.json.

I don't really want to be the code owner of std.json; if I must then I will rewrite it first.

tiehuis (changes requested):
Just some minor notes. As for the comment on the second transition only occurring after a number that should be true. An assertion would be welcome there.

thejoshwolfe:
We've got this in the new additions in #15602 : `peekNextTokenType()`. It just gives the type of the next token, not the entire next token. But that's coincidentally all that was being tested for in this PR, so perhaps it's good enough?

andrewrk:
I wanted to add, thank you @adarah for the patch. I'm sorry that it did not get to the point where it could be merged, but I hope that @thejoshwolfe's recent work on std.json will address your use case.

Issue #12805, opened 2022-09-10
stage2: reached unreachable on duplicate enum value


jcmoyer:
### Zig Version

0.10.0-dev.3952+9e070b653

### Steps to Reproduce

Attempt to compile:
```zig
const T = enum(u8) {
    a = 0x12,
    b = 0x12,
};

pub fn main() !void {
    const x: T = .a;
    _ = x;
}
```

### Expected Behavior

A compile error (stage1 prints: `error: enum tag value 18 already taken`)

### Actual Behavior

The compiler crashes with this stacktrace:

```
d:\Scratch\enumtest>d:\scratch\zig\stage3\bin\zig build
thread 5988 panic: reached unreachable code
D:\Scratch\zig\lib\std\debug.zig:281:14: 0x7ff635fb2829 in assert (zig.exe.obj)
    if (!ok) unreachable; // assertion failure
             ^
D:\Scratch\zig\lib\std\array_hash_map.zig:861:19: 0x7ff636c07f7c in putAssumeCapacityNoClobberContext (zig.exe.obj)
            if (@sizeOf(Context) != 0)
                  ^
D:\Scratch\zig\src\Sema.zig:2630:62: 0x7ff63695025e in zirEnumDecl (zig.exe.obj)
            enum_obj.values.putAssumeCapacityNoClobberContext(copied_tag_val, {}, .{
                                                             ^
D:\Scratch\zig\src\Sema.zig:928:67: 0x7ff6366ff1d1 in analyzeBodyInner (zig.exe.obj)
                    .enum_decl             => try sema.zirEnumDecl(          block, extended, inst),
                                                                  ^
D:\Scratch\zig\src\Sema.zig:626:45: 0x7ff636565bc0 in analyzeBodyBreak (zig.exe.obj)
    const break_inst = sema.analyzeBodyInner(block, body) catch |err| switch (err) {
                                            ^
D:\Scratch\zig\src\Module.zig:4550:50: 0x7ff6365620a3 in semaDecl (zig.exe.obj)
    const result_ref = (try sema.analyzeBodyBreak(&block_scope, body)).?.operand;
                                                 ^
D:\Scratch\zig\src\Module.zig:4185:38: 0x7ff6363713e2 in ensureDeclAnalyzed (zig.exe.obj)
    const type_changed = mod.semaDecl(decl_index) catch |err| switch (err) {
                                     ^
[...]

PR #12817, opened 2022-09-11
stage2: detect duplicate enum values

g-w1:
Closes https://github.com/ziglang/zig/issues/12805


Vexu:
Could you also handle this?
```zig
const T = enum(u8) {
    a = 0x12,
    b = 0x11,
    c,
};
```
(just add the same fix a few lines lower)

Also don't add a new case when there is already one for stage1 that can be moved instead.

g-w1:
Thanks @Vexu for keeping up with this!

PR #12826, opened 2022-09-12
Sema: Prevent coercion from tuple pointer to mutable slice

### LINKED ISSUE #12821: Mutable tuples coerce un-checked into mutable slices
### Zig Version

0.10.0-dev.3981+60678f5ba

### Steps to Reproduce

```zig
const std = @import("std");

pub fn main() void {
    // add explicit type annotation, like
    // `std.meta.Tuple(&.{ u7, u7, u7, u7 })`,
    // to avoid the segfault triggered by the commented line below.
    var a = .{ 1, 2, 3, 4 };
    var b: []u8 = &a;
    // b[0] = 7; // <- un-comment to observe a segfault.
    std.debug.print(
        \\a: (@{x}){any}
        \\b: (@{x}){any}
        \\
    , .{ @ptrToInt(&a), a, @ptrToInt(b.ptr), b });
}
```

### Expected Behavior

I would expect only comptime tuples to coerce to arrays this way, or at most, that the slice would alias the tuple memory when the tuple contains fields with a homogeneous field type, also equal to the type of the slice, instead of creating an equivalent temporary.

### Actual Behavior

Running with `zig run`:

Output with `var a: std.meta.Tuple(&[_]type{u8} ** 4) = .{ 1, 2, 3, 4 };`:
```
a: (@7ffe6db49f18){ 1, 2, 3, 4 }
b: (@7ffe6db49f34){ 1, 2, 3, 4 }
```

Output with `var a: std.meta.Tuple(&[_]type{u7} ** 4) = .{ 1, 2, 3, 4 };`:
```
a: (@7fff80bacae8){ 1, 2, 3, 4 }
b: (@7fff80bacb04){ 1, 2, 3, 4 }
```

Output with `var a: std.meta.Tuple(&[_]type{u8} ** 4) = .{ 1, 2, 3, 4 };` & `b[0] = 7;`:
```
a: (@7ffc9c319158){ 1, 2, 3, 4 }
b: (@7ffc9c319174){ 7, 2, 3, 4 }
```

Output with `var a = .{ 1, 2, 3, 4 };`:
```
a: (@aaaaaaaaaaaaaaaa){ 1, 2, 3, 4 }
b: (@20bb28){ 1, 2, 3, 4 }
```

Output with `var b: []u32 = &a;`:
```
a: (@aaaaaaaaaaaaaaaa){ 1, 2, 3, 4 }
b: (@208690){ 1, 2, 3, 4 }
```

Output with `var a: std.meta.Tuple(&[_]type{u7} ** 4) = .{ 1, 2, 3, 4 };`, `var b: []u32 = &a;` & `b[0] = 7;`:
```
a: (@7fff8cbcd1d8){ 1, 2, 3, 4 }
b: (@7fff8cbcd200){ 7, 2, 3, 4 }
```

Output with `var a = .{ 1, 2, 3, 4 };` & `b[0] = 7;`:
```
Segmentation fault at address 0x20bb28
src/main.zig:6:5: 0x20db72 in main (main)
    b[0] = 7;
    ^
{zig install}/files/lib/std/start.zig:568:22: 0x20d41d in posixCallMainAndExit (main)
            root.main();
                     ^
{zig install}/files/lib/std/start.zig:340:5: 0x20ce22 in _start (main)
    ^
Aborted (core dumped)
```

InKryption:
Closes #12821.
~~Also adds quick fix for seeming misuse of `pointerDecl`, should be audited at some point. This is just so that the CI passes.~~
Note: stage1 still does this.
Co-authored-by: topolarity [topolarity@tapscott.me](mailto:topolarity@tapscott.me)

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -17852,7 +17852,7 @@ fn zirReify(sema: *Sema, block: *Block, extended: Zir.Inst.Extended.InstData, in
             return sema.analyzeDeclVal(block, src, new_decl_index);
         },
         .Fn => {
-            const struct_val = union_val.val.castTag(.aggregate).?.data;
+            const struct_val: []const Value = union_val.val.castTag(.aggregate).?.data;
             // TODO use reflection instead of magic numbers here
             // calling_convention: CallingConvention,
             const cc = struct_val[0].toEnum(std.builtin.CallingConvention);
@@ -17886,12 +17886,17 @@ fn zirReify(sema: *Sema, block: *Block, extended: Zir.Inst.Extended.InstData, in
                     break :alignment alignment;
                 }
             };
+            const return_type = return_type_val.optionalValue() orelse
+                return sema.fail(block, src, "Type.Fn.return_type must be non-null for @Type", .{});
+
             var buf: Value.ToTypeBuffer = undefined;
 
             const args_slice_val = args_val.castTag(.slice).?.data;
             const args_len = try sema.usizeCast(block, src, args_slice_val.len.toUnsignedInt(mod.getTarget()));
-            var param_types = try sema.arena.alloc(Type, args_len);
-            var comptime_params = try sema.arena.alloc(bool, args_len);
+
+            const param_types = try sema.arena.alloc(Type, args_len);
+            const comptime_params = try sema.arena.alloc(bool, args_len);
+
             var noalias_bits: u32 = 0;
             var i: usize = 0;
             while (i < args_len) : (i += 1) {
@@ -17919,10 +17924,8 @@ fn zirReify(sema: *Sema, block: *Block, extended: Zir.Inst.Extended.InstData, in
                     return sema.fail(block, src, "Type.Fn.Param.arg_type must be non-null for @Type", .{});
 
                 param_types[i] = try param_type.toType(&buf).copy(sema.arena);
+                comptime_params[i] = false;
             }
-
-            const return_type = return_type_val.optionalValue() orelse
-                return sema.fail(block, src, "Type.Fn.return_type must be non-null for @Type", .{});
 
             var fn_info = Type.Payload.Function.Data{
                 .param_types = param_types,
@@ -23677,27 +23680,7 @@ fn coerceExtra(
                     },
                     else => {},
                 },
-                .Slice => {
-                    // pointer to tuple to slice
-                    if (inst_ty.isSinglePointer() and inst_ty.childType().isTuple() and dest_info.size == .Slice and
-                        sema.checkPtrAttributes(dest_ty, inst_ty, &in_memory_result))
-                    {
-                        return sema.coerceTupleToSlicePtrs(block, dest_ty, dest_ty_src, inst, inst_src);
-                    }
-
-                    // empty tuple to zero-length slice
-                    // note that this allows coercing to a mutable slice.
-                    if (inst_ty.isSinglePointer() and
-                        inst_ty.childType().tag() == .empty_struct_literal and
-                        dest_info.size == .Slice)
-                    {
-                        const slice_val = try Value.Tag.slice.create(sema.arena, .{
-                            .ptr = Value.undef,
-                            .len = Value.zero,
-                        });
-                        return sema.addConstant(dest_ty, slice_val);
-                    }
-
+                .Slice => to_slice: {
                     if (inst_ty.zigTypeTag() == .Array) {
                         return sema.fail(
                             block,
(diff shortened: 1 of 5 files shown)


topolarity:
Filed a bug for the incorrect `pointerDecl()` usage: https://github.com/ziglang/zig/issues/12885

iiuc, the workaround in this pr doesn't resolve this panic, so a proper fix will be needed later

InKryption:
The change this PR had a merge conflict with didn't address the coercion from mutable non-empty tuple pointers to mutable slices, as `checkPtrAttributes` makes this same mistake by pretending that coercion from array pointers and coercion from tuple pointers are equivalent.

topolarity:
Good news: Looks like the proper fix for that bug has now landed on master, thanks to @Vexu 

Can you rebase please?

topolarity (approved):
Changes look good to me! Thanks in particular for taking the time to clean up some of the redundant conditions and the small bug in `@Type`.

I would like some input from other team members on the semantics. I don't like the fact that the (lack of) aliasing is still observable between a and b:
```zig
const testing = @import("std").testing;

pub fn main() !void {
    var x: usize = 1;
    var a = .{ x, x + 1, x + 2, x + 3 };
    var b: []const usize = &a;
    
    a[0] = 7; // Changes to a are not reflected in b

    try testing.expect(a[0] == b[0]); // fails
}
```

Would it be reasonable to require that the tuple pointer point be backed by a decl and not a runtime variable? Or would that break an important use case for this coercion?

src/Sema.zig
```diff
@@ -23188,35 +23191,40 @@ fn coerceExtra(
...
                     }
 
-                    if (inst_ty.zigTypeTag() == .Array) {
+                    // pointer to tuple to slice
+                    if (dest_info.mutable) {
                         return sema.fail(
                             block,
                             inst_src,
-                            "array literal requires address-of operator (&) to coerce to slice type '{}'",
+                            "cannot cast pointer to tuple to '{}'",
```
Vexu, line 23223: It'd be nice if this explained why it cannot do the cast.

Vexu:
> I would like some input from other team members on the semantics. I don't like the fact that the (lack of) aliasing is still observable between a and b:
>
> Would it be reasonable to require that the tuple pointer point be backed by a decl and not a runtime variable? Or would that break an important use case for this coercion?

If requiring both the source pointer and the destination pointer to be `const` doesn't cause any trivially fixable errors then I'm in favor of it, otherwise I'm not sure yet.

src/Sema.zig
InKryption: How does something like
```
coercion from tuple pointer to {:0} would create constant temporary, but {:0} can only point to mutable memory.
```
sound?

Vexu: Well, it's true but also quite hard to understand. I was thinking of something more in the lines of `note: pointers to tuples can only coerce to constant pointers`.

InKryption:
Just to chip in:

> Would it be reasonable to require that the tuple pointer point be backed by a decl and not a runtime variable? Or would that break an important use case for this coercion?

Enforcing that it's backed by a comptime-resolvable value means that you can't coerce `&.{ x, x + 1 }` where `x` is runtime-known, and enforcing that the source is `const` doesn't guarantee that a temporary isn't created, so I'm not sure what issuing a compile for such a coercion would protect from - and all you'd need to bypass the check would be to do `@as(*const @TypeOf(tuple), &tuple)`, which, as well, no other mutable-to-const coercion in the language requires.

src/Sema.zig
topolarity: Sorry to inject thrash here, but I think the message needs to be specific to array/slice pointers, since `@as(*std.meta.Tuple(&.{usize, usize}), &.{x, x + 1})` is perfectly legal

How about `note: Tuple-to-array conversion requires a destination type of [N]T, *const [N]T, or []const T.` ? Not terribly short, but I haven't been able to come up with better. Or perhaps `note: a tuple can only coerce to an array type directly, or to a const pointer/slice of an array.`

topolarity:
I thought of a more serious problem issue with this coercion. Its lifetime is completely ephemeral:
```zig
const std = @import("std");

var current_queue: ?[]const u32 = undefined;
fn setQueue(x: anytype) void { current_queue = x; }
fn getQueue() ?[]const u32 { return current_queue; }

pub fn main() !void {
    var x: u32 = 1;
    var queue = .{x, x + 1};

    var cond: bool = true;
    if (cond)
        setQueue(&queue);

    // Surely `current_queue` is still valid here?
    // Nope. This fails:
    try expect(std.mem.eql(u32, getQueue().?, &[_]u32{1, 2}));
}
```

If the converted pointer is returned or stored anywhere, it can outlive its referent temporary.

InKryption:
@topolarity I was toying around with your example, and I think something much stranger is going on:
```zig
const std = @import("std");

var current_queue: []const u32 = undefined;
fn setQueue(x: *const std.meta.Tuple(&.{ u32, u32 })) void {
    current_queue = x;
}

pub fn main() void {
    const x: u32 = 1;
    const queue = std.meta.Tuple(&.{u32, u32}){ x, x };
    setQueue(&queue);
    std.debug.print("{any}\n", .{current_queue});
}
```
prints
```
{ 2, 1 }
```

topolarity:
I think in your modified example you've created a comptime-known tuple, so the coercion is safe

InKryption:
> I think in your modified example you've created a comptime-known tuple, so the coercion is safe.
> 
> **Edit:** Hmm, but `setQueue` is not performed at comptime. Scratch that.

More importantly, it should be printing `{ 1, 1 }`, not `{ 2, 1 }`. The same thing happens in your example, `current_queue` prints as `{ 2, 2 }`. For some reason it's adding 1 to the first element of the tuple during the coercion.

At any rate, besides that, it appears to obey the same rules as when `current_queue` is instead a `*const u32` and you assign it the address of a function local variable.

topolarity:
Interesting. On my machine that prints `{1, 1}`. I have a feeling that in both cases we are pointing to invalid stack memory, same as if we had returned a pointer to an explicitly-constructed temporary.

Indeed, this variant prints garbage for me:

```zig
fn clobberStack() void { var array: [20]u32 = undefined; _ = array[1]; }
pub fn main() void {
    const x: u32 = 1;
    const queue = std.meta.Tuple(&.{ u32, u32 }){ x, x };
    setQueue(&queue);
    clobberStack();
    std.debug.print("{any}\n", .{current_queue});
}
```

topolarity:
> I don't think this unsafety is inherent to this cast, but part of the broader issue of escaped pointers to locals

I'm not sure I agree. In this case, there's nothing to indicate syntactically that a temporary was created in `setQueue`, and there's no address-of (`&`/method-call syntax) to indicate a pointer to the temporary was taken. Every other pointer coercion would be safe in the above code.

> I personally don't think that it's worth it to try and address here by sacrificing the ergonomy of code like std.ChildProcess.init(&.{ runtime_str1, runtime_str2 })

Hmmm, if that's our primary use case maybe we should allow _only_ `&.{ x, y }`. That is, this coercion would be allowed only for fresh tuples. In that case, there's no need to perform a copy and no aliasing problem. It's more like the "tuple" inferred its needed type from its context (i.e. the tuple is actually an array).

I don't want this to block the PR though - This is clearly an improvement over status quo, so let's land this and circle back on the semantics later :+1:

PR #12830, opened 2022-09-13
fix android definition of pthread_rwlock_t

billzez:
These definitions no longer match the bionic source code.  Also, this line was a compilation error:

```
__reserved: [12]u8 = [_]u8{0} ** 2
```

### DIFF
--- a/lib/std/c/linux.zig
+++ b/lib/std/c/linux.zig
@@ -312,22 +312,10 @@ pub const pthread_cond_t = extern struct {
 pub const pthread_rwlock_t = switch (native_abi) {
     .android => switch (@sizeOf(usize)) {
         4 => extern struct {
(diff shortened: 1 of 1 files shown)


billzez:
pthread_rwlock_t is now an opaque type is the latest bionic/ndk:

https://android.googlesource.com/platform/bionic/+/master/libc/include/pthread.h
https://android.googlesource.com/platform/bionic/+/master/libc/include/bits/pthread_types.h

```
typedef struct {
#if defined(__LP64__)
  int32_t __private[14];
#else
  int32_t __private[10];
#endif
} pthread_rwlock_t;
```

The internal definition can be found here:

https://android.googlesource.com/platform/bionic/+/master/libc/bionic/pthread_rwlock.cpp

PR #12831, opened 2022-09-13
Don't add implicit cflags "-isystem {zig_lib}/include" with -nostdlibinc

no1wudi:
I'm working with `zig cc` and noticed that I can't remove the this flag with standard clang's option like -nostdinc / -nostdlibinc.

### DIFF
--- a/src/Compilation.zig
+++ b/src/Compilation.zig
@@ -858,6 +858,7 @@ pub const InitOptions = struct {
     strip: bool = false,
     function_sections: bool = false,
     no_builtin: bool = false,
+    nostdlibinc: bool = false,
     is_native_os: bool,
     is_native_abi: bool,
     time_report: bool = false,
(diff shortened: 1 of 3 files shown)


topolarity:
See also https://github.com/ziglang/zig/issues/11328

no1wudi:
@topolarity Thanks for your information, I find the issue in #11328 is not addressed with this patch, I've tried the same case but get:
```
> zig cc -nostdlibinc main.c 
In file included from main.c:1:
/usr/include/stdio.h:33:10: fatal error: 'stddef.h' file not found
#include <stddef.h>
         ^~~~~~~~~~
1 error generated.

```

topolarity:
That is the expected behavior for that issue actually - the problem is that Zig today doesn't fail (indicating the includes are still present despite `-nostdlibinc`)

PR #12825, opened 2022-09-12
Properly reset error return trace index

Vexu:
Currently depends on #12807 to free up two Zir instruction slots, only the last commit is relevant to this PR.

Example:
```zig
const std = @import("std");
test {
    return error.SkipZigTest;
}
fn foo() !void {
    return error.DontShow;
}
test {
    foo() catch {
        std.debug.dumpStackTrace(@errorReturnTrace().?.*);
        // return error.Second;
    };
    return error.Wtf;
}

```
With the comment:
```
Test [1/2] test_0... SKIP
Test [2/2] test_1... ./a.zig:6:5: 0x20f128 in foo (test)
    return error.DontShow;
    ^
Test [2/2] test_1... FAIL (Wtf)
./a.zig:13:5: 0x20f185 in test_1 (test)
    return error.Wtf;
    ^
0 passed; 1 skipped; 1 failed.
```
Uncommented:
```
Test [1/2] test_0... SKIP
Test [2/2] test_1... ./a.zig:6:5: 0x20f138 in foo (test)
    return error.DontShow;
    ^
Test [2/2] test_1... FAIL (Second)
./a.zig:6:5: 0x20f138 in foo (test)
    return error.DontShow;
    ^
./a.zig:11:9: 0x20f18e in test_1 (test)
        return error.Second;
        ^
0 passed; 1 skipped; 1 failed.
```

Closes #1923

Issue #11593, opened 2022-05-05
Bad error trace


danielchasehooper:
### Zig Version

0.10.0-dev.2083+13d1798ea

### Steps to Reproduce

run `zig test phantomErrorTrace.zig`

```
// phantomErrorTrace.zig
const std = @import("std");

fn alwaysErrors() !void {
    return error.BUG_ThisErrorShouldNotAppearInAnyTrace;
}

test "test expected error" {
    try std.testing.expectError(error.BUG_ThisErrorShouldNotAppearInAnyTrace, alwaysErrors());
}

test "trigger error trace" {
    return error.WeExpectToSeeThisInTheTrace;
}

```

### Expected Behavior

should only see error trace for the second test
```
1/2 test "test expected error"... OK
2/2 test "trigger error trace"... FAIL (WeExpectToSeeThisInTheTrace)
FAIL (WeExpectToSeeThisInTheTrace)
/Users/daniel/Developer/zigtests/phantomerrortrace.zig:14:5: 0x1005a992f in test "trigger error trace" (test)
    return error.WeExpectToSeeThisInTheTrace;
    ^
1 passed; 0 skipped; 1 failed.
```

### Actual Behavior

Error trace from the first test appears in the error trace for the second test
```
1/2 test "test expected error"... OK
2/2 test "trigger error trace"... FAIL (WeExpectToSeeThisInTheTrace)
FAIL (WeExpectToSeeThisInTheTrace)
/Users/daniel/Developer/zigtests/phantomerrortrace.zig:6:5: 0x10085dce7 in alwaysErrors (test)
    return error.BUG_ThisErrorShouldNotAppearInAnyTrace;
    ^
/Users/daniel/Developer/zigtests/phantomerrortrace.zig:14:5: 0x10085cd3b in test "trigger error trace" (test)
    return error.WeExpectToSeeThisInTheTrace;
    ^
1 passed; 0 skipped; 1 failed.
```

squeek502:
Very likely to be a duplicate of https://github.com/ziglang/zig/issues/1923

PR #12837, opened 2022-09-14
stage2: Pop error trace frames for handled errors (#1923)

### LINKED ISSUE #1923: when handling a returned error, if the expression/block does not try/return error, reset error return trace index to 0

topolarity:
This implements the (unapproved) error handling behavior described in https://github.com/ziglang/zig/issues/1923#issuecomment-1218495574.

The result is a basic ownership-style algorithm for errors. Barring any bugs (🤞), it should guarantee that error return frames follow a strict stack discipline, which we need to correctly accumulate them in a single stack-per-thread.

Error return traces "live" when:
  - stored in a `const` variable
  - handled in a `catch { ... }` or `else |err| { ... }` block
  - propagated to caller via `return` or `try`
  - passed to a function as an argument

Error return traces are killed when:
   - stored in a mutable `var`
   - leaving a `catch` or `else |err|` block with a non-error value
   - the `const` variable associated with an error trace falls out of scope
   - a function receives an error as an argument and returns a non-error
   
Here's an example that tries to show most of this together:
 ```zig
fn foo() !void {
    eat_dinner() catch {
        // Begin error handling here
        find_recipe() catch {};        // find_recipe() is removed from trace upon leaving this catch block
        if (cond)
            return error.HouseOnFire;  //      error result: Includes eat_dinner() in the error return trace, but not find_recipe()
        try cook_recipe();             //      error result: If this fails, the error return trace includes eat_dinner(), followed by cook_recipe()
    }
    // We made it out of another catch block, eat_dinner() is removed from the trace 
    return error.MyInlawsAreComing;    //      error result: Error trace starts fresh here
}

fn bar() !void {
    // In addition to returning from within the block, you can pass
    // the result to an external return, try, catch, or if:
    return try_me() catch |err| switch(err) {
        error.IgnorableError => null,   // non-error result: Pops try_me() from the error return trace
        error.ImportantError => b: {
            break :b error.FatalError;  //     error result: Includes try_me() in the error return trace
        },
        else => err,                    //     error result: Includes try_me() in the error return trace
    };
}
test {
    try expectError(error.FatalError, bar());         // If this fails, `bar()` appears in the error return trace
    try expectError(error.MyInlawsAreComing, foo());  // If this fails, `foo()` appears in the error return trace (`bar` was already popped)
}
```

The main caveat is that traces are lost when storing errors to a mutable `var`. For example:
```zig
fn foo() !void {
    var x = try_me();   // If this were `const` or `return try_me();`, the trace wouldn't be killed here
     // ... fancy logic here
    return x;           // Error trace starts fresh here - try_me() is not included.
}
```

**Note:** There is room to relax this last condition for `var`, but only a little bit. We could do the same thing we do for `const` and keep the error trace alive for the remainder of the block where the _assignment_ happens. Any wider scope will violate the strict stack discipline for error return traces, so it won't work. In the end, I decided the most consistent behavior for the user is just to kill error return traces when assigning to `var`.
 
Supersedes #12825. Closes #1923.
(Many thanks @Vexu for that PR btw -- it was very helpful to tighten up some of my handling in AstGen)

### DIFF
--- a/lib/test_runner.zig
+++ b/lib/test_runner.zig
@@ -44,24 +44,23 @@ pub fn main() void {
         if (!have_tty) {
             std.debug.print("{d}/{d} {s}... ", .{ i + 1, test_fn_list.len, test_fn.name });
         }
-        if (result: {
-            if (test_fn.async_frame_size) |size| switch (io_mode) {
-                .evented => {
-                    if (async_frame_buffer.len < size) {
-                        std.heap.page_allocator.free(async_frame_buffer);
-                        async_frame_buffer = std.heap.page_allocator.alignedAlloc(u8, std.Target.stack_align, size) catch @panic("out of memory");
-                    }
-                    const casted_fn = @ptrCast(fn () callconv(.Async) anyerror!void, test_fn.func);
-                    break :result await @asyncCall(async_frame_buffer, {}, casted_fn, .{});
-                },
-                .blocking => {
-                    skip_count += 1;
-                    test_node.end();
-                    progress.log("SKIP (async test)\n", .{});
-                    continue;
-                },
-            } else break :result test_fn.func();
-        }) |_| {
+        const result = if (test_fn.async_frame_size) |size| switch (io_mode) {
+            .evented => blk: {
+                if (async_frame_buffer.len < size) {
+                    std.heap.page_allocator.free(async_frame_buffer);
+                    async_frame_buffer = std.heap.page_allocator.alignedAlloc(u8, std.Target.stack_align, size) catch @panic("out of memory");
+                }
+                const casted_fn = @ptrCast(fn () callconv(.Async) anyerror!void, test_fn.func);
+                break :blk await @asyncCall(async_frame_buffer, {}, casted_fn, .{});
+            },
+            .blocking => {
+                skip_count += 1;
+                test_node.end();
+                progress.log("SKIP (async test)\n", .{});
+                continue;
+            },
+        } else test_fn.func();
+        if (result) |_| {
             ok_count += 1;
             test_node.end();
             if (!have_tty) std.debug.print("OK\n", .{});
--- a/lib/test_runner.zig
+++ b/lib/test_runner.zig
@@ -44,23 +44,24 @@ pub fn main() void {
         if (!have_tty) {
             std.debug.print("{d}/{d} {s}... ", .{ i + 1, test_fn_list.len, test_fn.name });
         }
-        const result = if (test_fn.async_frame_size) |size| switch (io_mode) {
-            .evented => blk: {
-                if (async_frame_buffer.len < size) {
-                    std.heap.page_allocator.free(async_frame_buffer);
-                    async_frame_buffer = std.heap.page_allocator.alignedAlloc(u8, std.Target.stack_align, size) catch @panic("out of memory");
-                }
-                const casted_fn = @ptrCast(fn () callconv(.Async) anyerror!void, test_fn.func);
-                break :blk await @asyncCall(async_frame_buffer, {}, casted_fn, .{});
-            },
-            .blocking => {
-                skip_count += 1;
-                test_node.end();
-                progress.log("SKIP (async test)\n", .{});
-                continue;
-            },
-        } else test_fn.func();
-        if (result) |_| {
+        if (result: {
+            if (test_fn.async_frame_size) |size| switch (io_mode) {
+                .evented => {
+                    if (async_frame_buffer.len < size) {
+                        std.heap.page_allocator.free(async_frame_buffer);
+                        async_frame_buffer = std.heap.page_allocator.alignedAlloc(u8, std.Target.stack_align, size) catch @panic("out of memory");
+                    }
+                    const casted_fn = @ptrCast(fn () callconv(.Async) anyerror!void, test_fn.func);
+                    break :result await @asyncCall(async_frame_buffer, {}, casted_fn, .{});
+                },
+                .blocking => {
+                    skip_count += 1;
+                    test_node.end();
+                    progress.log("SKIP (async test)\n", .{});
+                    continue;
+                },
+            } else break :result test_fn.func();
+        }) |_| {
             ok_count += 1;
             test_node.end();
             if (!have_tty) std.debug.print("OK\n", .{});
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -499,6 +499,25 @@ pub const Block = struct {
         return result_index;
     }
 
+    /// Insert an instruction into the block at `index`. Moves all following
+    /// instructions forward in the block to make room. Operation is O(N).
+    pub fn insertInst(block: *Block, index: Air.Inst.Index, inst: Air.Inst) error{OutOfMemory}!Air.Inst.Ref {
+        return Air.indexToRef(try block.insertInstAsIndex(index, inst));
+    }
+
+    pub fn insertInstAsIndex(block: *Block, index: Air.Inst.Index, inst: Air.Inst) error{OutOfMemory}!Air.Inst.Index {
+        const sema = block.sema;
+        const gpa = sema.gpa;
+
+        try sema.air_instructions.ensureUnusedCapacity(gpa, 1);
+
+        const result_index = @intCast(Air.Inst.Index, sema.air_instructions.len);
+        sema.air_instructions.appendAssumeCapacity(inst);
+
+        try block.instructions.insert(gpa, index, result_index);
+        return result_index;
+    }
+
     fn addUnreachable(block: *Block, src: LazySrcLoc, safety_check: bool) !void {
         if (safety_check and block.wantSafety()) {
             _ = try block.sema.safetyPanic(block, src, .unreach);
@@ -5648,6 +5667,85 @@ fn funcDeclSrc(sema: *Sema, block: *Block, src: LazySrcLoc, func_inst: Air.Inst.
     return owner_decl.srcLoc();
 }
 
+/// Add instructions to block to "pop" the error return trace.
+/// If `operand` is provided, only pops if operand is non-error.
+fn popErrorReturnTrace(
+    sema: *Sema,
+    block: *Block,
+    src: LazySrcLoc,
+    operand: ?Air.Inst.Ref,
+    saved_error_trace_index: Air.Inst.Ref,
+) CompileError!void {
+    var is_non_error: ?bool = null;
+    var is_non_error_inst: Air.Inst.Ref = undefined;
+    if (operand) |op| {
+        is_non_error_inst = try sema.analyzeIsNonErr(block, src, op);
+        if (try sema.resolveDefinedValue(block, src, is_non_error_inst)) |cond_val|
+            is_non_error = cond_val.toBool();
+    } else is_non_error = true; // no operand means pop unconditionally
+
+    if (is_non_error == true) {
+        // AstGen determined this result does not go to an error-handling expr (try/catch/return etc.), or
+        // the result is comptime-known to be a non-error. Either way, pop unconditionally.
+
+        const unresolved_stack_trace_ty = try sema.getBuiltinType(block, src, "StackTrace");
+        const stack_trace_ty = try sema.resolveTypeFields(block, src, unresolved_stack_trace_ty);
+        const ptr_stack_trace_ty = try Type.Tag.single_mut_pointer.create(sema.arena, stack_trace_ty);
+        const err_return_trace = try block.addTy(.err_return_trace, ptr_stack_trace_ty);
+        const field_ptr = try sema.structFieldPtr(block, src, err_return_trace, "index", src, stack_trace_ty, true);
+        try sema.storePtr2(block, src, field_ptr, src, saved_error_trace_index, src, .store);
+    } else if (is_non_error == null) {
+        // The result might be an error. If it is, we leave the error trace alone. If it isn't, we need
+        // to pop any error trace that may have been propagated from our arguments.
+
+        try sema.air_extra.ensureUnusedCapacity(sema.gpa, @typeInfo(Air.Block).Struct.fields.len);
+        const cond_block_inst = try block.addInstAsIndex(.{
+            .tag = .block,
+            .data = .{
+                .ty_pl = .{
+                    .ty = Air.Inst.Ref.void_type,
+                    .payload = undefined, // updated below
+                },
+            },
+        });
+
+        var then_block = block.makeSubBlock();
+        defer then_block.instructions.deinit(sema.gpa);
+
+        // If non-error, then pop the error return trace by restoring the index.
+        const unresolved_stack_trace_ty = try sema.getBuiltinType(block, src, "StackTrace");
+        const stack_trace_ty = try sema.resolveTypeFields(block, src, unresolved_stack_trace_ty);
+        const ptr_stack_trace_ty = try Type.Tag.single_mut_pointer.create(sema.arena, stack_trace_ty);
+        const err_return_trace = try then_block.addTy(.err_return_trace, ptr_stack_trace_ty);
+        const field_ptr = try sema.structFieldPtr(&then_block, src, err_return_trace, "index", src, stack_trace_ty, true);
+        try sema.storePtr2(&then_block, src, field_ptr, src, saved_error_trace_index, src, .store);
+        _ = try then_block.addBr(cond_block_inst, Air.Inst.Ref.void_value);
+
+        // Otherwise, do nothing
+        var else_block = block.makeSubBlock();
+        defer else_block.instructions.deinit(sema.gpa);
+        _ = try else_block.addBr(cond_block_inst, Air.Inst.Ref.void_value);
+
+        try sema.air_extra.ensureUnusedCapacity(sema.gpa, @typeInfo(Air.CondBr).Struct.fields.len +
+            then_block.instructions.items.len + else_block.instructions.items.len +
+            @typeInfo(Air.Block).Struct.fields.len + 1); // +1 for the sole .cond_br instruction in the .block
+
+        const cond_br_inst = @intCast(Air.Inst.Index, sema.air_instructions.len);
+        try sema.air_instructions.append(sema.gpa, .{ .tag = .cond_br, .data = .{ .pl_op = .{
+            .operand = is_non_error_inst,
+            .payload = sema.addExtraAssumeCapacity(Air.CondBr{
+                .then_body_len = @intCast(u32, then_block.instructions.items.len),
+                .else_body_len = @intCast(u32, else_block.instructions.items.len),
+            }),
+        } } });
+        sema.air_extra.appendSliceAssumeCapacity(then_block.instructions.items);
+        sema.air_extra.appendSliceAssumeCapacity(else_block.instructions.items);
+
+        sema.air_instructions.items(.data)[cond_block_inst].ty_pl.payload = sema.addExtraAssumeCapacity(Air.Block{ .body_len = 1 });
+        sema.air_extra.appendAssumeCapacity(cond_br_inst);
+    }
+}
+
 fn zirCall(
     sema: *Sema,
     block: *Block,
@@ -5737,6 +5835,9 @@ fn zirCall(
 
     const args_body = sema.code.extra[extra.end..];
 
+    var input_is_error = false;
+    const block_index = @intCast(Air.Inst.Index, block.instructions.items.len);
+
     const parent_comptime = block.is_comptime;
     // `extra_index` and `arg_index` are separate since the bound function is passed as the first argument.
     var extra_index: usize = 0;
@@ -5754,10 +5855,8 @@ fn zirCall(
         else
             func_ty_info.param_types[arg_index];
 
-        const old_comptime = block.is_comptime;
-        defer block.is_comptime = old_comptime;
         // Generate args to comptime params in comptime block.
-        block.is_comptime = parent_comptime;
+        defer block.is_comptime = parent_comptime;
         if (arg_index < fn_params_len and func_ty_info.comptime_params[arg_index]) {
             block.is_comptime = true;
         }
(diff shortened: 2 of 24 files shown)


lib/test_runner.zig
```diff
@@ -61,23 +61,24 @@ pub fn main() void {
...
-            },
-            .blocking => {
-                skip_count += 1;
-                test_node.end();
-                progress.log("SKIP (async test)\n", .{});
-                continue;
-            },
-        } else test_fn.func();
-        if (result) |_| {
+        if (result: {
```
Vexu, line 64: If this change is necessary then that is one more downside for this more complex approach.

src/Sema.zig
```diff
@@ -6176,19 +6185,55 @@ fn analyzeCall(
...
-            .data = .{ .pl_op = .{
-                .operand = func,
-                .payload = sema.addExtraAssumeCapacity(Air.Call{
-                    .args_len = @intCast(u32, args.len),
-                }),
-            } },
-        });
-        sema.appendRefsAssumeCapacity(args);
-        break :res func_inst;
+        const backend_supports_error_return_tracing = sema.mod.comp.bin_file.options.use_llvm;
```
Vexu, line 6188: This whole thing also needs to be handled for generic instantiations.

src/Sema.zig
```diff
@@ -15446,6 +15629,56 @@ fn wantErrorReturnTracing(sema: *Sema) bool {
...
+    const src = LazySrcLoc.nodeOffset(inst_data);
+
+    // This is only relevant at runtime.
+    if (block.is_comptime) return Air.Inst.Ref.zero_usize;
+
+    // In the corner case that `catch { ... }` or `else |err| { ... }` is used in a function
+    // that does *not* make any errorable calls, we still need an error trace to interact with
+    // the AIR instructions we've already emitted.
+    if (sema.owner_func != null)
+        sema.owner_func.?.calls_or_awaits_errorable_fn = true;
```
Vexu, line 15643: Do you have an example of this case? Also `sema.owner_func` is guaranteed non-null here since we already checked that the block isn't comptime.

lib/test_runner.zig
topolarity: The idea is that without this restriction, internally to a function error return traces can be left un-popped.

Here's what it would look like if we relaxed it:
```zig
const x = foo(); // fails
const y = baz(); // also fails
switch (cond) {
    0 => x,
    1 => y,
} catch {
    if (cond) {
        return error.Foo;  // both foo() and baz() end up in the error return trace, followed by this line
    }
    // A non-error return will still pop both foo() and baz(), thankfully
    return;            
} 

try quaf();                       // If this fails, error trace will show foo(), followed by baz(), followed by quaf()
if (cond2)
    return error.UnrelatedError;  // both foo() and baz() end up in the error return trace, followed by this line

// A non-error return can will still pop both foo() and baz(), thankfully
return;
```

See the commit message of 09842f9 for more info.

One other thought I had was to add a bit to errors so that we can notice when one's trace was killed due to being saved to a variable and to augment to panic message with "(error return trace includes dropped entries -- see `<link>` for more info)"

src/Sema.zig
topolarity: Good tip.

Here's an example:
```zig
fn non_errorable() void {
    // Make sure catch works even in a function that does not call any errorable functions.
    //
    // This test is needed because stage 2's fix for #1923 means that catch blocks interact
    // with the error return trace index.
    var x: error{Foo}!void = {};
    return x catch {};
}

test "catch within a function that calls no errorable functions" {
    non_errorable();
}
```

Intended to add this as a test, but I accidentally left it behind in a stash - Nice catch.

Vexu: And why does that need to save the error return trace index? There are no errors being returned anywhere in that example?

lib/test_runner.zig
Vexu: I'm not sure what the downside in that example is supposed to be?

src/Sema.zig
topolarity: The problem is that in "foo() catch { ... }" the `.save_err_ret_index` precedes the error-able call, so Sema does not know that an errorable call is made or not until after we analyze that `.save_err_ret_index` instruction. 

A similar problem means that `@errorReturnTrace()` allows returns `null` in a function before the first errorable function call, even if that function does call errorable functions and has a live error trace:
```zig
fn bar() !void { return error.TheSkyIsFalling; }

fn foo() !void {
    // "error: unable to unwrap null" if uncommented
    //std.log.err("entry: {}", .{ @errorReturnTrace().?.index });

    bar() catch {
        std.log.err("in catch: {}", .{ @errorReturnTrace().?.index });
    };

    std.log.err("exit: {}", .{ @errorReturnTrace().?.index });
}
```

Vexu: I think the behavior with `@errorReturnTrace` is desired. And wouldn't every `.save_err_ret_index` before a call to an erorrable function return zero anyways?

topolarity: Not in the case that `foo()` is being called from within a `catch` block, for example

lib/test_runner.zig
topolarity: The problem I see is that `foo()` and `baz()` pollute the error return trace even if they are "handled".

For example, in a loop:
```zig
while (i < N) : (i += 1) {
    const x = foo();
    if (x) |_| {
        // code here
    } else |err| {
        if (cond)
            return error.Foo; // Return trace includes foo() up to `i` times in the error trace
    }
}
``` 
`foo()` will appear in the error return trace however many times it failed during this function call.

This may be a matter of opinion, though. I'm happy to modify the behavior if desired.

src/Sema.zig
Vexu: But `foo` doesn't return any errors either?

topolarity: Sorry, I meant for a general `foo()` not the example above

Vexu: How could the error return trace be anything but zero if it is inspected before any operation that could set it to a non-zero value?

topolarity: The index can be updated by the caller before the function entry. This is how error traces are chained together:
```zig
fn foo() !void { return error.BigIssue; }
fn bar() !void { return error.AnotherProblem; }
fn baz() !void {
    // error trace is not empty at function entry, since we were called from within a `catch`
    return bar();
}
test {
    foo() catch {
        // @errorReturnTrace.?.index == 1 here
        baz() catch unreachable;
    }
}
```

topolarity: Sounds good to me. I'll change to that fix in my next round of edits :+1:

topolarity:
Uh oh, I found a problematic example:

```zig
const expectError = @import("std").testing.expectError;

fn alwaysErrors() !void { return error.BUG_ThisErrorShouldNotAppearInAnyTrace; }
fn foo() !void { return error.Foo; }

test "test expected error" {
    try expectError(error.BUG_ThisErrorShouldNotAppearInAnyTrace, alwaysErrors());
    try expectError(error.BUG_ThisErrorShouldNotAppearInAnyTrace, alwaysErrors());

    try expectError(error.Bar, foo());
}
```

Without this PR, the error trace is:
```console
Test [1/1] test.test expected error... expected error.Bar, found error.Foo
Test [1/1] test.test expected error... FAIL (TestExpectedError)
./test_err6.zig:3:27: 0x20f3b8 in alwaysErrors (test)
fn alwaysErrors() !void { return error.BUG_ThisErrorShouldNotAppearInAnyTrace; }
                          ^
./test_err6.zig:3:27: 0x20f3b8 in alwaysErrors (test)
fn alwaysErrors() !void { return error.BUG_ThisErrorShouldNotAppearInAnyTrace; }
                          ^
./test_err6.zig:4:18: 0x20f6c8 in foo (test)
fn foo() !void { return error.Foo; }
                 ^
/home/topolarity/repos/zig/lib/std/testing.zig:41:13: 0x20f78f in expectError__anon_1116 (test)
            return error.TestExpectedError;
            ^
./test_err6.zig:12:5: 0x20f9b2 in test.test expected error (test)
    try expectError(error.Bar, foo());
    ^
```

With it, the trace is:
```console
Test [1/1] test.test expected error... expected error.Bar, found error.Foo
Test [1/1] test.test expected error... FAIL (TestExpectedError)
/home/topolarity/repos/zig/stage4/lib/zig/std/testing.zig:41:13: 0x20f756 in expectError__anon_1116 (test)
            return error.TestExpectedError;
            ^
./test_err6.zig:12:5: 0x20f9ac in test.test expected error (test)
    try expectError(error.Bar, foo());
    ^
```

Before we had too many entries, now we have too few.

lib/test_runner.zig
topolarity: There is a different relaxation that might make sense: when initializing `const`/`var`, delay popping an error trace for the initializer until the variable falls out of scope. This wouldn't resolve my first example, but it would fix the second.

Using the same technique for args (delay pop until function returns a non-error) should resolve the problematic example below too.

I'll experiment a bit and report back

andrewrk:
@topolarity could you rebase this against master branch please? My auto-rebase script failed due to conflicts.

lib/test_runner.zig
topolarity: Sorry for the delay here, just got back from a trip a few days ago 🏖️ 

I've updated this PR to include the fix for function args, and `testing.expectError` now works as expected.

Keeping error traces alive for a `const` variable (until the end of the containing block) still seems like a good idea. However, it requires moving significant portions of this logic to Sema, so I may wait for some feedback before I dive in.

topolarity:
With the latest push, examples in the style of https://github.com/ziglang/zig/issues/11593 work as expected:

```zig
const expectError = @import("std").testing.expectError;

fn alwaysErrors() !void { return error.BUG_ThisErrorShouldNotAppearInAnyTrace; }
fn foo() !void { return error.Foo; }

test "test expected error" {
    try expectError(error.BUG_ThisErrorShouldNotAppearInAnyTrace, alwaysErrors());
    try expectError(error.BUG_ThisErrorShouldNotAppearInAnyTrace, alwaysErrors());
    try expectError(error.Bar, foo());
}
```

Output with this PR is:
```console
Test [1/1] test.test expected error... expected error.Bar, found error.Foo
Test [1/1] test.test expected error... FAIL (TestExpectedError)
./testme.zig:4:18: 0x211d08 in foo (test)
fn foo() !void { return error.Foo; }
                 ^
/home/topolarity/repos/zig/build/stage3/lib/zig/std/testing.zig:37:13: 0x211dcf in expectError__anon_1111 (test)
            return error.TestExpectedError;
            ^
./testme.zig:9:5: 0x212052 in test.test expected error (test)
    try expectError(error.Bar, foo());
    ^
```

Before we had entries from each expectError call. Now, the entries are cleaned up if `expectError` passes.

topolarity:
Was able to lift the `const` restriction. This new behavior no longer requires modifying the test runner 🚀

andrewrk:
Looking forward to this change. After the latest rebase, all CI checks passed. However due to merging of other PRs there are now a lot of conflicts. Would you mind rebasing?

topolarity:
Alright, looks like the rebase went well

The rebase did expose an existing bug https://github.com/ziglang/zig/issues/13175 which together with this change causes 1 behavior test regression. I've skipped that test for now

andrewrk (changes requested):
Thanks for doing this incredible work.

I'm inclined to merge this, however, the additional complications to the compiler internals, as well as the additional runtime performance overhead are giving me pause.

Before merging, can you share some details on these topics?

 * How might this integrate with async functions? In stage1 we combine error return traces when doing `await`.
 * Can you provide any benchmark data about how runtime performance is affected?
 * Can you share how this affects the compilation speed of the self-hosted compiler?

topolarity:
Very good questions.

The same things gave me concern - Let's see if I can get a read on the situation for you.

> How might this integrate with async functions? In stage1 we combine error return traces when doing await.

I think the strategy here can stay largely the same. Merging the traces means that it's as if the error were generated "at" the await. For the save/restore logic here, if we treat an errorable `await` exactly the way that we treat an errorable function call, then popping should continue to work correctly.

> Can you share how this affects the compilation speed of the self-hosted compiler?

This is the most significant difference I'd expect. This does add a significant number of new ZIR instructions:

```
; src/codegen/c.zig (master)
# Source bytes:       158.0126953125KiB
# Tokens:             34052 (166.29296875KiB)
# AST Nodes:          17641 (224.0751953125KiB)
# Total ZIR bytes:    554.3564453125KiB
# Instructions:       34155 (300.1904296875KiB)
# String Table Bytes: 18.626953125KiB
# Extra Data Items:   60282 (235.4765625KiB)

; src/codegen/c.zig (this PR)       
# Source bytes:       158.0126953125KiB
# Tokens:             34052 (166.29296875KiB)
# AST Nodes:          17641 (224.0751953125KiB)
# Total ZIR bytes:    585.8505859375KiB
# Instructions:       36505 (320.8447265625KiB)
# String Table Bytes: 18.626953125KiB
# Extra Data Items:   63057 (246.31640625KiB)
```

That's a ~6.9% increase in instructions, and a 5.7% increase in ZIR bytes (sorry to eat up your `.try` savings :grimacing:)

Here are the build timings:

```
# master (stage3) ReleaseSafe building Debug
54.64s user 2.51s system 101% cpu 56.450 total
54.95s user 2.28s system 101% cpu 56.521 total
55.01s user 2.26s system 101% cpu 56.559 total

# this PR (stage3) ReleaseSafe building Debug
55.86s user 2.44s system 101% cpu 57.544 total
56.30s user 2.31s system 101% cpu 57.894 total
55.33s user 2.36s system 101% cpu 56.963 total


# master (stage3) Debug building Debug
102.98s user 2.70s system 103% cpu 1:42.59 total
103.23s user 3.25s system 102% cpu 1:43.62 total
102.81s user 2.88s system 102% cpu 1:42.73 total

# this PR (stage3) Debug building Debug
103.77s user 2.92s system 102% cpu 1:43.73 total
105.92s user 2.95s system 102% cpu 1:45.86 total
104.37s user 2.74s system 102% cpu 1:43.99 total
```

Each run was done from a completely fresh cache, so that this includes AstGen and `build.zig` compilation time.

> Can you provide any benchmark data about how runtime performance is affected?

The self-compilation timings above should include the runtime performance hit, in addition to the delta from code changes in this PR. I did try to make some more adversarial benchmarks that would do a lot of save/restore, but my attempts didn't find a delta that rose above the measurement noise yet.

In general there are two places with extra runtime code:
  1. Error handling: Trace index save/restore at every block entry/exit for `catch`, `else |err|`, and any blocks containing `const x = foo();` where foo() is errorable
  2. Discarded error traces: Trace index save/restore wrapping any errorable function call whose trace is "killed" (generally, those that do not go to `try`/`catch`/`return`/`if-else-err`/`const x = ...`)

`try foo()` and `return foo()` require no extra interactions with the trace index (except when a `return` exits an error-handling block as in (1)). Overall, compared to the overhead of error return tracing in general I think this is probably not a large delta, but let me know if you'd like me to test some existing projects or continue hunting for adversarial cases.

andrewrk:
Thanks for taking these measurements. I'm a little confused by "ReleaseSafe" here - did you hack up the build.zig script? Because it only exposes the `-Drelease` flag which will select `ReleaseFast`.

topolarity:
> Thanks for taking these measurements. I'm a little confused by "ReleaseSafe" here - did you hack up the build.zig script? Because it only exposes the `-Drelease` flag which will select `ReleaseFast`.

Yeah, that's exactly what I did here

I did a quick sanity check against the usual ReleaseFast too, which clocked in about 2-3 seconds faster iirc. Delta was similar between master and PR, about 1 second elapsed time.

> Also, which one of those timings is "wall clock"?

"Total" should be the wall clock. These are reported from zsh's "time"

andrewrk (approved):
OK, this is good to be merged!

topolarity:
Alright, final rebase (gods willing 🙏 )

I also added one small final commit that reduces the size impact to `Sema.Block`.

Let's let CI give us the green light and then bring this in!

andrewrk:
Congrats on the big merge, and thanks for your patience with all the rebases!

Would you be willing to type me up some release notes for this change to demonstrate it?

squeek502:
Thank you for taking this on @topolarity (and @Vexu for starting it). I think this being addressed will hugely benefit Zig overall, as needing to know that (a) some error return traces needed to be ignored and (b) how to determine which error traces should be ignored was both confusing for new Zig programmers and frustrating for experienced Zig programmers.

Also, props on recognizing the need for and implementing a more complete solution than the original proposal in #1923.

topolarity:
> Would you be willing to type me up some release notes for this change to demonstrate it?

Sure thing! Can I get your take on https://github.com/ziglang/zig/pull/12990?

I'm OK whether it lands or not but I wanted to ask since it affects the example

> Thank you for taking this on @topolarity (and @Vexu for starting it). 

Thank you for the kind words @squeek502 - Always great to know when things are having an impact downstream 🙂

PR #12839, opened 2022-09-14
std.testing: Reference all declarations with `refAllDecls` and `refAllDeclsRecursive`

### LINKED ISSUE #12838: `std.testing.refAllDecls` references `pub` decls only, private ones are ignored
### Zig Version

0.10.0-dev.3999+fda6d4477

### Steps to Reproduce

```zig
const std = @import("std");

test {
}

const S = struct {
    test "Should fail" {
        try expect(false);
    }
};
```

Suppose we have a file that looks like the above, named `foo.zig`, and run `zig test foo.zig`.

### Expected Behavior

Obviously it should fail since it has `expect(false)` in the test.

### Actual Behavior

The test is successfully done. To be more precise, the test case located in `S` is just ignored just because `S` is not `pub`ed.
In Zig v0.9.1 it fails as expected. Plus the official doc introduces this technique as well: https://ziglang.org/documentation/master/#toc-Nested-Container-Tests, where we can see the test case in `S` is supposed to be executed even if `S` is private. In other words, the technique shown in the doc works fine in v0.9.1, not in the nightly.

I'm pretty sure that this behavior change has come from https://github.com/ziglang/zig/pull/10909. Given that `refAllDecls` resides in the `testing` module, it would make much more sense to allow private declarations to be referenced as well. It can be useless for us to annotate structs that we create just for organizing test cases with `pub` in my opinion.

magurotuna:
Fixes #12838.

With this PR, `std.testing.refAllDecls` and `std.testing.refAllDeclsRecursive` will allow the semantic analyzer to look at all declarations regardless their visibility (private or `pub`).

The motivation for this PR is as follows.

- As written in #12838, given that these two functions reside in the `testing` module, it would make much more sense to also reference private declarations.
- This makes the behavior of these functions align with [what the official doc shows](https://ziglang.org/documentation/master/#toc-Nested-Container-Tests).

### DIFF
--- a/lib/std/testing.zig
+++ b/lib/std/testing.zig
@@ -727,7 +727,7 @@ pub fn checkAllAllocationFailures(backing_allocator: std.mem.Allocator, comptime
 pub fn refAllDecls(comptime T: type) void {
     if (!builtin.is_test) return;
     inline for (comptime std.meta.declarations(T)) |decl| {
-        if (decl.is_pub) _ = @field(T, decl.name);
+        _ = @field(T, decl.name);
     }
 }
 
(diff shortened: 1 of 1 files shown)


magurotuna:
Seems like `@field` doesn't support private declarations... OK, maybe we'll keep the current behavior and need to fix the document instead.

PR #12847, opened 2022-09-15
allow `@src()` outside of functions and allow `@src().line` at comptime

pfgithub:
It seems like `@src()` is intentionally runtime-only for incremental compilation reasons so this might not be wanted (https://github.com/ziglang/zig/pull/12016#issuecomment-1178092847). I would expect that column should be runtime-only then for the same reason.

---

I was trying to generate shader source code at comptime and ran into these when trying to generate `#line` directives.

This changes `@src()` to work when not in a function. `SourceLocation.fn_name` is set to an empty string, which could potentially be a footgun but it seems hard to run into that by accident.

This also makes `@src().line` available at comptime.

```zig
fn demo() void {
    @compileLog(@src());
}

comptime {
    _ = demo;
    @compileLog(@src());
}

//Compile Log Output:
@as(builtin.SourceLocation, .{.file = "…/a.zig\x00", .fn_name = "\x00", .line = 7, .column = 17})
@as(builtin.SourceLocation, .{.file = "…/a.zig\x00", .fn_name = "demo\x00", .line = 2, .column = 17})
```


Vexu:
`@src` is implemented as desired https://github.com/ziglang/zig/issues/2029#issuecomment-645793168, if you want to change that go through a proposal.

PR #12848, opened 2022-09-15
Update once.zig to remove mutex

RetroDev256:
The mutex in once.zig can be removed, in favor of less memory overhead; thread safety can be maintained with only the boolean present in the Once struct.


Vexu:
Duplicate of #10497

> The Mutex is used so that all concurrent callers would suspend themselves until the one caller that had successfully acquired the mutex has completed its critical section.
> 
> This is generally desired when using Once as we want only a single caller to initialize some singleton instance, and have all other callers wanting to use the initialized variant of the singleton instance block until initialization is complete.

Issue #12468, opened 2022-08-17
stage2: panic from packedStructToInt


scheibo:
### Zig Version

0.10.0-dev.3590+a12abc6d6

### Steps to Reproduce

```zig
const std = @import("std");

pub const Foo = packed struct {
    type: Foo.Type = .A,
    _: u2 = 0,
    data: u4 = 0,

    pub const Type = enum(u2) {
        A,
        B,
        C,
    };

    comptime {
        std.debug.assert(@sizeOf(Foo) == 1);
    }
};

test "Foo" {
    const foo: Foo = .{ .type = .B, .data = 4 };
    try std.testing.expectEqual(0b0100_0001, @bitCast(u8, foo));
}
```

```
zig test -fno-stage1 main.zig
```

### Expected Behavior

```
$ zig test -fno-stage1 main.zig 
All 1 tests passed.
```

I expect this code to either pass like on stage 1 or to provide an error message indicating what should be changed.

### Actual Behavior

With `-fno-stage1` the self hosted compiler panics:

```
$ zig test -fno-stage1 main.zig 
thread 11257471 panic: reached unreachable code
/Users/kjs/Code/src/github.com/scheibo/zig/src/value.zig:1313:25: 0x10654b3ef in value.Value.packedStructToInt (zig1)
                else => unreachable,
                        ^
/Users/kjs/Code/src/github.com/scheibo/zig/src/value.zig:1274:55: 0x1065480ea in value.Value.writeToMemory (zig1)
                    const host_int = packedStructToInt(val, ty, target, &buf);
                                                      ^
/Users/kjs/Code/src/github.com/scheibo/zig/src/Sema.zig:24894:22: 0x1065a5289 in Sema.bitCastVal (zig1)
    val.writeToMemory(old_ty, sema.mod, buffer);
                     ^
/Users/kjs/Code/src/github.com/scheibo/zig/src/Sema.zig:24828:47: 0x1064d887e in Sema.bitCast (zig1)
        const result_val = try sema.bitCastVal(block, inst_src, val, old_ty, dest_ty, 0);
                                              ^
/Users/kjs/Code/src/github.com/scheibo/zig/src/Sema.zig:8526:24: 0x1062776ce in Sema.zirBitcast (zig1)
    return sema.bitCast(block, dest_ty, operand, operand_src);
                       ^
[...]

PR #12849, opened 2022-09-15
Value: add Enum and negative int support to bitCast packed struct

kkHAIKE:
fix https://github.com/ziglang/zig/issues/12468

'field_bigint_const' must not be negative  in 'packedStructToInt', it will cause shift and bitOr result invalid, so add twos-complement transform with 'packedStructToInt' and 'intToPackedStruct'.

😊


Vexu:
Superceded by #13124 and #13221

PR #12869, opened 2022-09-16
std: add a ring buffer.

gcoakes:
`Ring` is a generic ring buffer with a comptime known size. It is safe to use with multiple producers and multiple consumers. When single threaded build is active, it defaults to a non-atomic implementation which has the same size/ layout as the MPMC implemenetation.

`RingGeneric` is the same as `Ring` except it accepts an enum value indicating the intended use of the structure. Currently, there are only `.mpmc` and `.non_atomic` variants, but there is room for further specialization such as to improve the performance of a single produce and single consumer workload.

### DIFF
--- a/lib/std/atomic.zig
+++ b/lib/std/atomic.zig
@@ -6,11 +6,13 @@ pub const Ordering = std.builtin.AtomicOrder;
 pub const Stack = @import("atomic/stack.zig").Stack;
 pub const Queue = @import("atomic/queue.zig").Queue;
 pub const Atomic = @import("atomic/Atomic.zig").Atomic;
+pub const Ring = @import("atomic/ring.zig").RingBuffer;
 
 test "std.atomic" {
     _ = @import("atomic/stack.zig");
     _ = @import("atomic/queue.zig");
     _ = @import("atomic/Atomic.zig");
+    _ = @import("atomic/ring.zig");
 }
 
 pub inline fn fence(comptime ordering: Ordering) void {
--- a/lib/std/atomic/ring.zig
+++ b/lib/std/atomic/ring.zig
@@ -1,0 +1,448 @@
+const std = @import("std");
+const builtin = @import("builtin");
+const Atomic = std.atomic.Atomic;
+const expectEqual = std.testing.expectEqual;
+const expectEqualSlices = std.testing.expectEqualSlices;
+
+pub const RingKind = enum(u8) {
+    mpmc,
+    spmc,
+    mpsc,
+    spsc,
+    non_atomic,
+    _,
+};
+
+/// A fixed size, fifo queue. It is lock-free (but not wait-free). The size
+/// must be a power of two.
+///
+/// Internally, it uses a producer and consumer "span." Each span consists of
+/// a head and a tail index. The buffer is considered empty when the consumer
+/// head is equal to the producer tail. Conversely, the buffer is considered
+/// full when the producer head is equal to the consumer tail plus the size of
+/// the buffer.
+///
+/// The space between a span's head and tail indicates in progress copying
+/// in or out of the buffer. For multi-producer or multi-consumer, concurrent
+/// writes to or reads from the buffer cause the losing thread to spin. It is
+/// possible for a thread to become starved due to this.
+///
+/// Counterintuitively, each index is not bound within the size of the buffer;
+/// instead, additions to each index are bound within the maximum size of the
+/// index using wrapping addition. Upon accessing each index's slot, they are
+/// masked to get the actual index within the buffer.
+///
+/// This is largely based off the logic described in DPDK's [rte_ring](https://
+/// doc.dpdk.org/guides-1.8/prog_guide/ring_lib.html) library reference.
+pub fn RingBuffer(
+    comptime T: type,
+    comptime size: usize,
+    comptime kind: RingKind,
+) type {
+    std.debug.assert(std.math.isPowerOfTwo(size));
+    const idx_mask = size - 1;
+    const Span = struct {
+        head: Atomic(usize) = .{ .value = 0 },
+        tail: Atomic(usize) = .{ .value = 0 },
+    };
+    return struct {
+        const Self = @This();
+        /// The backing buffer of all the elements.
+        buffer: [size]T = undefined,
+        /// Write index.
+        producer: Span = .{},
+        /// Read index.
+        consumer: Span = .{},
+
+        /// Initialize the buffer in-place.
+        pub inline fn init(self: *Self) void {
+            self.producer = .{};
+            self.consumer = .{};
+        }
+
+        // Choose the correct `get` function depending upon `kind`.
+        pub usingnamespace switch (kind) {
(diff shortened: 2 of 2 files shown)


lib/std/ring.zig
```diff
@@ -0,0 +1,304 @@
...
+/// the buffer to which the next item will be written. Both the tail and
+/// the head are atomically incremented via a CAS loop and are bound to the
+/// range of `0..std.math.maxInt(usize)`. The size is comptime asserted to be
+/// a power of two. This permits the tail and head to overflow while remaining
+/// aligned to the size. Every slot within the buffer is tagged with an atomic
+/// enum indicating the current state of the slot (occupied or empty).
+///
+/// *Credit*: This draws upon the Linux Kernel's BPF ring buffer specification
+/// (but is a novel implementation). Specifically, the idea of tagged slots and
+/// of monotonic, wrapping head/tail.
```
matu3ba, line 29: wrapping head/tail is a standard approach as it compiles down to unconditionally applying a bitmask on the index (you could also compute and write it down explicitly).

More of a curiosity: How does the approach compare to [this paper](https://www.osti.gov/servlets/purl/1531271) ?
As I understand it, this approach is optimized to use less memory and less complex than the Kernel approach of an additional data structure to track the status.

lib/std/ring.zig
```diff
@@ -0,0 +1,304 @@
...
+/// A circular buffer with a tail and a head. The tail is an index within the
+/// buffer from which the next item will be read. The head is an index within
+/// the buffer to which the next item will be written. Both the tail and
+/// the head are atomically incremented via a CAS loop and are bound to the
+/// range of `0..std.math.maxInt(usize)`. The size is comptime asserted to be
+/// a power of two. This permits the tail and head to overflow while remaining
+/// aligned to the size. Every slot within the buffer is tagged with an atomic
+/// enum indicating the current state of the slot (occupied or empty).
+///
+/// *Credit*: This draws upon the Linux Kernel's BPF ring buffer specification
```
matu3ba, line 27: The document is phrased "This document describes BPF ring buffer design, API, and implementation details." and it is more like a reference manual rather than limited "user manual/spec".

matu3ba (commented):
Looks small and sweet optimized to use less memory. I have some questions and comments.

1. Can you describe the single-threaded difference to fifo.zig (static case)? Should those be combined or not? The interface looks similar except for the tag with CAS.
2. Do you think this should go into `atomic/` together with lock-free queue and stack?
3. If we have SPMC or MPSC, can we relax the memory ordering for better perf or is another design better?

gcoakes:
>     1. Can you describe the single-threaded difference to fifo.zig (static case)? Should those be combined or not? The interface looks similar except for the tag with CAS.

This is the first I've looked at fifo.zig, but at first glance, it appears to be intended for slices of data. This PR is meant to be used for discrete messages. Also, the non-atomic implementation is meant to transparently replace the other variant.

>     2. Do you think this should go into `atomic/` together with lock-free queue and stack?

Yes, I'll update it.

>     3. If we have SPMC or MPSC, can we relax the memory ordering for better perf or is another design better?

I believe so, but this is my first crack at a lock free data structure like this, so take it with a grain of salt. I'm fairly certain the `get` or `put` functions could be wait free for MPSC or SPMC. I know they could be wait free for SPSC.

I didn't implement it here since MPMC is the most general. I'm fairly confident the current implementation is correct, and I didn't want to expend any more cycles implementing the more specific variants right now.

lib/std/ring.zig
```diff
@@ -0,0 +1,304 @@
...
+    else
+        RingGeneric(T, size, .mpmc);
+}
+
+const RingKind = enum {
+    mpmc,
+    non_atomic,
+};
+
+/// A circular buffer with a tail and a head. The tail is an index within the
```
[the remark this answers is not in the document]
gcoakes: I just felt repetitive saying ring buffer so much. I'll update it.

lib/std/ring.zig
```diff
@@ -0,0 +1,304 @@
...
+/// buffer from which the next item will be read. The head is an index within
```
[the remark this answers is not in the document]
gcoakes: I think you're mistaken here. The second sentence mentions the tail. The third mirrors it.

lib/std/ring.zig
gcoakes: > wrapping head/tail with 2s complement is a standard approach

Blew my mind when I read it. It probably is standard. crossbeam and DPDK appear to use it.

> How does the approach compare to [this paper](https://www.osti.gov/servlets/purl/1531271) ?

I hadn't seen that paper until now. I think you're incorrect in saying that paper doesn't utilize "an additional data structure to track the status." From an initial look, it seems like they propose stamping each node within the buffer with a "seqid" (end of Section 4 Paragraph 2). It actually seems to be the inspiration or direct attribution of Rust's [crossbeam::ArrayQueue](https://docs.rs/crossbeam/latest/crossbeam/queue/struct.ArrayQueue.html). There they use the upper bits of the head/tail as the "lap."

I'll have to sit with this and digest it for a bit. Do you think re-designing around a different algorithm now would be preferable to merging the existing design?

lib/std/ring.zig
gcoakes: Frankly, I fail to see the difference between a "reference manual" and a "user manual/spec." Would "documentation" suffice instead of "specification?"

lib/std/ring.zig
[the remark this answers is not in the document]
matu3ba: Then this is resolved (to me).

lib/std/ring.zig
matu3ba: yes.

lib/std/ring.zig
gcoakes: @matu3ba, your comment made me look at some other implementations. I really liked DPDK's method since it permitted untagged slots. I scrapped everything and re-implemented using their producer/consumer spans method. I still haven't given that paper a thorough read...

gcoakes:
The latest push re-implements without the tagged slots. Instead it uses DPDK's method. It essentially boils down to two pairs of head/tail spans, one for the producer(s) and one for the consumer(s). The head represents slots in the buffer which have been reserved for either writing or reading. The tail represents slots which have been completely written or read.

daurnimator:
See `std.LinearFifo`.

We decided a couple of years ago that we didn't want to make it threadsafe.
Perhaps you have arguments for why that should be reconsidered?

kprotty:
Yea, there's too many ways to write concurrent ring buffers without this one having any clear advantages. There's unbounded spinning in the multi-producer/multi-consumer case + unnecessary rmws with single consumer's `fetchAdd` / spinning with `tryCompareAndSwap` + not aware of false sharing + doesn't support batch operations + doesn't use FAA for fast path compared to something like [looqueue](https://github.com/oliver-giersch/looqueue-rs), etc. I'd say this should be left as a library instead.

gcoakes:
Ok, thanks for the feed back.

Issue #10397, opened 2021-12-23
Handling of heterogeneous computing


gwenzek:
This issue is a follow up on my own issue/experience working to add NVPTX (Nvidia GPU) support to Zig stage2 (#10189 ) and @Snektron ongoing work on SpirV backend (#2683) (also stage2). But the same considerations also to other heterogeneous programming like FPGA.

One of the issue with this "targets" is that they aren't standalone, typically those devices are driven by a more conventional CPU, the host. The host code will first push binary code to the device and then asks the device to execute some functions from this binary on some runtime-known inputs. This creates issues because we need to generate binaries that are consistent across two different targets. Notably what are the types allowed in the host/device interface ?

## Main questions

So the questions I'd like to talk about are:

* Should Zig handle heterogeneous programming ? (I think yes given #2683 approval and reactions to my previous work on this)
* Should we keep only the C ABI between host and device, or allow for arbitrary Zig structs ? (I'm arguing for Zig structs)
* In general which level of integration between the host/device code can be expected ?
* Should this integration live in the Zig compiler or in libraries ?
* What can Zig compiler do to help the libraries ?

Personally I'd be quite happy if Zig could provide minimal support for ergonomic heterogeneous programming (not constrained by C-ABI), even if there are quirks and only support some combination of platform, so that we can learn more about the issues and then later try to iron those outs.

gwenzek:
In this message I describe how we can provide support for ergonomic heterogeneous programming,
looking at two implementations: Clang and my own Cudaz/Stage 2 fork.

Clang is already providing support for heterogeneous programming and imposes no constraint on what types can be used at the interface between host and device. Let's look at how they handle this (the schema shows Spir-V, but this matches what clang does for targeting NVPTX). 

![image](https://user-images.githubusercontent.com/5920036/147219694-5ff1efe6-b4ba-4d94-83c0-ae2d4ad29538.png)
(from https://intel.github.io/llvm-docs/CompilerAndRuntimeDesign.html#dpc-compiler-architecture)

The idea is that the compiler:
1. extracts the "device" code from a regular source file
2. compiles it to the device binary format
3. generates a _host_ object file that embeds the device binary, and contains minimal functions to push the binary to the device and to call the device code from the host (using an [offload bundler](https://clang.llvm.org/docs/ClangOffloadBundler.html) )

The advantage of this approach is that the object files are standalone, and that the device code is hidden. In particular the ABI between the host and device isn't exposed.
[...]

PR #12878, opened 2022-09-16
Update Nvptx backend for Zig 0.10

gwenzek:
Due to the progress on 0.10 a few things broke Nvptx backend.

Some of the changes I'm pushing here are outside of  `link/NvPtx.zig`, and are modifying  shared code.
Let me know if you can think of a better way to structure the changes. From least controversial to more controversial.

* link/NvPtx.zig -> I'm modifying in place the "Compilation" object. Now I restore the changes before returning. A bit hacky but it allows to reuse the rest of the LLVM backend code without modification there.

* Target.zig -> PTX only support debug information starting from PTX 7.5. Not sure how to test "ptx >= 7.5".
 
* Sema.zig -> allow to use Zig objects in exported .PtxKernel functions. This allow tight integration between CPU/GPU code and was approved by Andrew given that we can change it later if we realize it was a bad idea to expose the ABI like this. https://github.com/ziglang/zig/issues/10397#issuecomment-1000709733

* Module.zig -> sanitize the name of variables. This is required because the names can be like "io.fixed_buffer_stream.FixedBufferStream([]u8).writer" and don't match the C-like identifiers expected by PTX.

Tagging @Snektron who expressed interest in this workstream and @Vexu who kindly merge previous PR (#10189)

Thanks to the progress of self hosted, we can now generate debug information in the PTX files. Also the assembly syntax now works as documented, so it's simpler to use special ptx registers. And I can use the same zig binary for building the device and host code, so it's pretty exciting.

See https://github.com/gwenzek/cudaz/tree/e8895596009c689300fe7c7193fa2dbf7db07629 for user code using this Zig branch.

### DIFF
--- a/src/Module.zig
+++ b/src/Module.zig
@@ -720,6 +720,15 @@ pub const Decl = struct {
         var buffer = std.ArrayList(u8).init(mod.gpa);
         defer buffer.deinit();
         try decl.renderFullyQualifiedName(mod, buffer.writer());
+
+        // Sanitize the name for nvptx which is more restrictive.
+        if (mod.comp.bin_file.options.target.cpu.arch.isNvptx()) {
+            for (buffer.items) |*byte| switch (byte.*) {
+                '{', '}', '*', '[', ']', '(', ')', ',', ' ', '\'' => byte.* = '_',
+                else => {},
+            };
+        }
+
         return buffer.toOwnedSliceSentinel(0);
     }
 
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -21397,7 +21397,12 @@ fn validateExternType(
         },
         .Fn => {
             if (position != .other) return false;
-            return !Type.fnCallingConventionAllowsZigTypes(ty.fnCallingConvention());
+            return switch (ty.fnCallingConvention()) {
+                // For now we want to authorize PTX kernel to use zig objects, even if we end up exposing the ABI.
+                // The goal is to experiment with more integrated CPU/GPU code.
+                .PtxKernel => true,
+                else => !Type.fnCallingConventionAllowsZigTypes(ty.fnCallingConvention()),
+            };
         },
         .Enum => {
             var buf: Type.Payload.Bits = undefined;
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -18202,12 +18202,6 @@ fn zirAddrSpaceCast(sema: *Sema, block: *Block, extended: Zir.Inst.Extended.Inst
     else
         dest_ptr_ty;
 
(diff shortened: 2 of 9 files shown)


src/Module.zig
```diff
@@ -705,6 +705,15 @@ pub const Decl = struct {
...
         defer buffer.deinit();
         try decl.renderFullyQualifiedName(mod, buffer.writer());
+
+        // Sanitize the name for nvptx which is more restrictive.
+        if (mod.comp.bin_file.options.target.cpu.arch.isNvptx()) {
+            for (buffer.items) |*byte| switch (byte.*) {
+                '{', '}', '*', '[', ']', '(', ')', ',', ' ', '\'' => byte.* = '_',
+                else => {},
+            };
+        }
```
Snektron, line 715: Is this not gonna cause linker errors when two symbols map to the same value? Say functions `@"A()"` and `@"A{}"`.

src/Sema.zig
```diff
@@ -20583,7 +20583,12 @@ fn validateExternType(
             8, 16, 32, 64, 128 => return true,
             else => return false,
         },
-        .Fn => return !Type.fnCallingConventionAllowsZigTypes(ty.fnCallingConvention()),
+        .Fn => switch (ty.fnCallingConvention()) {
+            // For now we want to authorize PTX kernel to use zig objects, even if we end up exposing the ABI.
+            // The goal is to experiment with more integrated CPU/GPU code.
+            .PtxKernel => return true,
+            else => return !Type.fnCallingConventionAllowsZigTypes(ty.fnCallingConvention()),
+        },
```
Vexu, line 20591: Why not just change `fnCallingConventionAllowsZigTypes` to return true for `.PtxKernel`?

src/Module.zig
gwenzek: yes, it's true. I don't know how robust this should be. I could sanitize and append some hash. Would that be better ?

gwenzek:
Thanks for the first round of review. I'm gonna try to rebase this against @Snektron changes in #12879 and test with LLVM15.

src/Sema.zig
gwenzek: so `fnCallingConventionAllowsZigTypes` is already returning `true` for `.PtxKernel`. But that's the issue. Because I still want to be able to export the kernels. So we allow exporting functions that don't have Zig types in their CC, and we also allow exporting PtxKernel even if they have Zig types.

andrewrk:
Nice work.

Would you be willing to write up some corresponding release notes that I can put into the upcoming 0.10.0 release notes?

PR #12879, opened 2022-09-17
Amdgpu improvements

Snektron:
Putting this up for CI, still need to add some compile error checks.

Thid PR adds a bunch of amd gpu stuff:
* amdpal/amdhsa c abi sizes
* AmdgpuKernel calling convention
* new builtin: `@addrSpaceCast(comptime as: AddressSpace, ptr: anytype) anytype` allows conversion from one AS to another.
* Changed Zig address space semantics
    * All pointers are now required to be in the generic address space when not specified otherwise. This is because there is no implicit casting between address spaces, and otherwise it would be too great of an effort to update the entire stdlib. In practise, this is also done by hip and cuda. Recent architectures support generic pointers, but specialized pointers are ztill more efficient. To mitigate that, those compilers rely on the optimizer to lower to the proper type.
* Improved handling of address spaces in LLVM backend
    * Specific to amdgpu and related to ^: locals must be allocated as private, in addrspace 5. For that i updated buildAlloca to handle this. It always returns a generic pointer so that llvm doesnt get sad. I had to move the alignment setting functionality to there as well as it needs to be dome before the cast, but i think its cleaner anyway.

Known issues: llvm gets sad when you assume the size of a pointer which actual size is not the same as the size of usize. This requires significant effort to allow pointers to have a custom size depending on the address space, and Im still working up the motivation to do that.


lib/compiler_rt/atomics.zig
```diff
@@ -35,6 +35,17 @@ const largest_atomic_size = switch (arch) {
     else => @sizeOf(usize),
 };
 
+// The size (in bytes) of the smallest atomic object that the architecture can
+// perform fetch/exchange atomically. Note, this does not encompass load and store.
+// Objects smaller than this threshold are implemented in terms of compare-exchange
+// of a larger value.
+const smallest_atomic_fetch_exch_size = switch (arch) {
```
kprotty, line 42: I remember there being some arm/mips platforms without atomic rmws for non-32bit-ints. Just wanted to note

lib/compiler_rt/atomics.zig
```diff
@@ -35,6 +35,17 @@ const largest_atomic_size = switch (arch) {
...
+    // On AMDGPU, there are no instructions for atomic operations other than load and store
+    // (as of LLVM 15), and so these need to be implemented in terms of atomic CAS.
+    .amdgcn => @sizeOf(u32),
+    else => @sizeOf(u8),
+};
+
 const cache_line_size = 64;
```
kprotty, line 49: This should borrow from `std.atomic.cache_line` as `64` isn't always the case.

lib/compiler_rt/atomics.zig
```diff
@@ -214,6 +225,31 @@ inline fn atomic_exchange_N(comptime T: type, ptr: *T, val: T, model: i32) T {
         const value = ptr.*;
         ptr.* = val;
         return value;
+    } else if (@sizeOf(T) < smallest_atomic_fetch_exch_size) {
+        // Machine does not support this type, but it does support a larger type.
+        const WideAtomic = std.meta.Int(.unsigned, smallest_atomic_fetch_exch_size * 8);
+
+        const addr = @ptrToInt(ptr);
+        const wide_addr = addr & ~(@as(T, smallest_atomic_fetch_exch_size) - 1);
```
kprotty, line 233: Could this accidentally trigger faults if the `wide_addr` drops into a page without read/write perms?

lib/compiler_rt/atomics.zig
```diff
@@ -214,6 +225,31 @@ inline fn atomic_exchange_N(comptime T: type, ptr: *T, val: T, model: i32) T {
...
+
+        const inner_offset = addr & (@as(T, smallest_atomic_fetch_exch_size) - 1);
+        const inner_shift = @intCast(std.math.Log2Int(T), inner_offset * 8);
+
+        // Put the interesting bits at the right position (branch has dynamic RHS).
+        const shifted_value = @as(WideAtomic, val) << inner_shift;
+        // Mask that guards the bits we care about
+        const mask = @as(WideAtomic, std.math.maxInt(T)) << inner_shift;
+        while (true) {
+            const wide_old = @atomicLoad(WideAtomic, wide_ptr, .Acquire);
```
kprotty, line 244: You should load once outside of the while loop, then use the failure result returned from `@cmpxchgWeak` as the new value on the next loop iteration. Not sure what the orderings as supposed to be here but (since none are passed in) you should use `SeqCst` for the load as well.

lib/compiler_rt/atomics.zig
```diff
@@ -298,6 +334,38 @@ inline fn fetch_op_N(comptime T: type, comptime op: std.builtin.AtomicRmwOp, ptr
...
+        const inner_offset = addr & (@as(T, smallest_atomic_fetch_exch_size) - 1);
+        const inner_shift = @intCast(std.math.Log2Int(T), inner_offset * 8);
+
+        const mask = @as(WideAtomic, std.math.maxInt(T)) << inner_shift;
+
+        while (true) {
+            // Compute new wide value with updated bits.
+            const wide_old = @atomicLoad(WideAtomic, wide_ptr, .Acquire);
+            const old = @truncate(T, (wide_old & mask) >> inner_shift);
+            const new = switch (op) {
```
kprotty, line 354: You could probably generalize all of this code with something like `wideUpdate(struct { fn update(old) { return new } })`

lib/compiler_rt/atomics.zig
Snektron: Hm good question. I think as long as the wide type is less than the page size, there should be no problems. If the to-be-updated type is all the way at the end of a page, then `wide_addr` will point to exactly the last `WideAtomic` bytes before the page end. Similar on the other extreme.

lib/compiler_rt/atomics.zig
kprotty: Ignore, `atomic.cache_line` is specifically for avoiding false sharing

src/Sema.zig
```diff
@@ -25643,6 +25702,7 @@ fn coerceCompatiblePtrs(
     inst_src: LazySrcLoc,
 ) !Air.Inst.Ref {
     const inst_ty = sema.typeOf(inst);
+    // TODO check const/volatile/alignment/addrspace
```
Vexu, line 25705: This is done now.

src/Sema.zig
```diff
@@ -17686,6 +17691,60 @@ fn reifyStruct(
     return sema.analyzeDeclVal(block, src, new_decl_index);
 }
 
+fn zirAddrSpaceCast(sema: *Sema, block: *Block, extended: Zir.Inst.Extended.InstData) CompileError!Air.Inst.Ref {
```
Vexu, line 17694: If this builtin is going to exist then `zirPtrCast` should check that the address space is not changed.

src/Sema.zig
```diff
@@ -17686,6 +17691,60 @@ fn reifyStruct(
...
+    const ptr_info = ptr_ty.ptrInfo().data;
+    const dest_ty = try Type.ptr(sema.arena, sema.mod, .{
+        .pointee_type = ptr_info.pointee_type,
+        .@"align" = ptr_info.@"align",
+        .@"addrspace" = dest_addrspace,
+        .mutable = ptr_info.mutable,
+        .@"allowzero" = ptr_info.@"allowzero",
+        .@"volatile" = ptr_info.@"volatile",
+        .size = ptr_info.size,
+    });
```
Vexu, line 17728: IMO it'd be better to just change the addresspace instead of copying over all the other properties.

src/Sema.zig
```diff
@@ -17686,6 +17691,60 @@ fn reifyStruct(
...
+    const src = LazySrcLoc.nodeOffset(extra.node);
+    const addrspace_src: LazySrcLoc = .{ .node_offset_builtin_call_arg0 = extra.node };
+    const ptr_src: LazySrcLoc = .{ .node_offset_builtin_call_arg1 = extra.node };
+
+    const dest_addrspace = try sema.analyzeAddressSpace(block, addrspace_src, extra.lhs, .pointer);
+    const ptr = try sema.resolveInst(extra.rhs);
+    const ptr_ty = sema.typeOf(ptr);
+
+    // TODO in addition to pointers, this instruction is supposed to work for
+    // pointer-like optionals and slices.
```
Vexu, line 17705: This already handles slices and optionals are easy to handle by wrapping `dest_ty` with `Type.optional` if the operand is an optional.

src/Sema.zig
Snektron: rebase error, its fixed now

src/Sema.zig
Snektron: Good catch, that was the plan. Its implemented

src/Sema.zig
Snektron: implemented

src/Sema.zig
Snektron: implemented

lib/std/target.zig
```diff
@@ -1157,6 +1157,17 @@ pub const Target = struct {
                 };
             }
 
+            /// Returns whether this architecture supporst the address space
```
r00ster91, line 1160:
```suggestion
            /// Returns whether this architecture supports the address space.
```

gwenzek:
Just wanted to say that I've tried this PR with some small changes in https://github.com/ziglang/zig/pull/12878 and that I can compile and pass tests in https://github.com/gwenzek/cudaz/tree/dev with ReleaseFast (using PTX backend)

The generated PTX correctly uses the specific load/store instructions: `ld.shared.f32 	%f20, [%rd4+2048];`

ReleaseSafe and Debug still have issues, but I don't think they are related to those changes.

src/Sema.zig
```diff
@@ -17751,6 +17756,55 @@ fn reifyStruct(
...
+    const dest_ptr_ty = try Type.ptr(sema.arena, sema.mod, ptr_info);
+    const dest_ty = if (ptr_ty.zigTypeTag() == .Optional)
+        try Type.optional(sema.arena, dest_ptr_ty)
+    else
+        dest_ptr_ty;
+
+    if (try sema.resolveMaybeUndefVal(block, ptr_src, ptr)) |val| {
+        // Pointer value should compatible with both address spaces.
+        // TODO: Figure out why this generates an invalid bitcast.
+        return sema.addConstant(dest_ty, val);
```
gwenzek, line 17793: I had to remove this block, because it would generate in LLVM IR a `bitcast` instead of an `addrspacecast`, which is incorrect for "shared" space, and would prevent the LLVM optimizer to correctly propagate the adressspace info in the code.

lib/std/target.zig
Snektron: Fixed, thanks

src/Sema.zig
Snektron: Yes, that is because the cast is then performed at runtime instead of at compile time as it is now. This is the way that it _should_ work, but it does not yet.

Snektron: I think we can try to fix that in a different MR and get this and yours in

Snektron:
Looks like the tests pass now

andrewrk:
Nice work.

Would you be willing to write up some corresponding release notes that I can put into the upcoming 0.10.0 release notes?

Snektron:
Yes i can do that somewhere

andrewrk:
If you drop them here I can copy+paste

gwenzek:
Thanks for the review.

For release note I propose:

* add experimental support for AMDGPU programming
* add `@addrSpaceCast` builtin to make it easier to use GPU address spaces
* add support for debug info on PTX

This should cover both this and #12878.

PR #12882, opened 2022-09-17
enable event loop for ios, tvos, and watchos

billzez:
Enables event loop for all darwin platforms.


kerrhome:
This will be useful.

PR #12893, opened 2022-09-18
macho: rewrite incremental linker, and init splitting of linking contexts

kubkon:
Turns out, I was wrong in collapsing incremental with full/one-shot linking contexts. While this deduplicated some code paths, it merged the contexts too much making it very hard to experiment on the incremental linker in isolation which is a big drawback since the incremental codepath is largely unfinished and bound to change wildly in the next couple of months. Therefore, with this PR I have started the process of separating the contexts enough to be able to experiment on either context in isolation without worrying about the other one.

With that out of the way, I am pleased to announce that the rewritten incremental MachO linker now is indeed incrementally allocating space in the final binary, committing `Decl` code as it's generated by the codegen, and storing only static and dynamic relocations (rebases, binds, and lazy-binds), and reapplying them when something shifts in memory only - yep, in memory only, as now the linker is free to move sections in file without affecting the virtual memory allocs. So what does that mean? Here's an example output of a test binary where file offsets (and in-file ordering of sections/segments) don't match the virtual memory addresses:

```
$ zacho --verify-memory-layout test
MEMORY LAYOUT:                                                                                                                                                                                                                                                                                                                                       
            __PAGEZERO ---------- 0                                                                                                                                                                                                                                                                                                                  
                       |                                                                                                                                                                                                                                                                                                                             
                       |                                                                                                                                                                                                                                                                                                                             
                       |                                                                                                                                                                                                                                                                                                                             
                       ---------- 100000000                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
                __TEXT ---------- 100000000                                                                                                                                                                                                                                                                                                          
                       |                                                                                                                                                                                                                                                                                                                             
                       |                                                                                                                                                                                                                                                                                                                             
                       |                                                                                                                                                                                                                                                                                                                             
                       ---------- 100002000                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
               __TEXT1 ---------- 100002000                                                                                                                                                                                                                                                                                                          
                       |                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                     
                  __text -------- 100002000                                                                                                                                                                                                                                                                                                          
                         |                                                                                                                                                                                                                                                                                                                           
                         -------- 1000691e5                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
                       |                                                                                                                                                                                                                                                                                                                             
                       ---------- 10006a000                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
               __TEXT2 ---------- 100075000                                                                                                                                                                                                                                                                                                          
                       |                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                     
                 __stubs -------- 100075000                                                                                                                                                                                                                                                                                                          
                         |                                                                                                                                                                                                                                                                                                                           
                         -------- 10007500c                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
                       |                                                                                                                                                                                                                                                                                                                             
                       ---------- 100076000                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
               __TEXT3 ---------- 100076000                                                                                                                                                                                                                                                                                                          
                       |                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                     
           __stub_helper -------- 100076000                                                                                                                                                                                                                                                                                                          
                         |                                                                                                                                                                                                                                                                                                                           
                         -------- 100076023                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
                       |                                                                                                                                                                                                                                                                                                                             
                       ---------- 100077000                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
          __DATA_CONST ---------- 100077000                                                                                                                                                                                                                                                                                                          
                       |                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                     
                   __got -------- 100077000                                                                                                                                                                                                                                                                                                          
                         |                                                                                                                                                                                                                                                                                                                           
                         -------- 10007e650                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
                       |                                                                                                                                                                                                                                                                                                                             
                       ---------- 10007f000                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
         __DATA_CONST1 ---------- 100081000                                                                                                                                                                                                                                                                                                          
                       |                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                     
                 __const -------- 100081000                                                                                                                                                                                                                                                                                                          
                         |                                                                                                                                                                                                                                                                                                                           
                         -------- 10009f820                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
                       |                                                                                                                                                                                                                                                                                                                             
                       ---------- 1000a0000                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
                __DATA ---------- 1000ab000                                                                                                                                                                                                                                                                                                          
                       |                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                     
         __la_symbol_ptr -------- 1000ab000                                                                                                                                                                                                                                                                                                          
                         |                                                                                                                                                                                                                                                                                                                           
                         -------- 1000ab010                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
                       |                                                                                                                                                                                                                                                                                                                             
                       ---------- 1000ac000                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
               __DATA1 ---------- 1000ac000                                                                                                                                                                                                                                                                                                          
                       |                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                     
                  __data -------- 1000ac000                                                                                                                                                                                                                                                                                                          
                         |                                                                                                                                                                                                                                                                                                                           
                         -------- 1000ac190                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                     
                       |                                                                                                                                                                                                                                                                                                                             
                       ---------- 1000ad000                                                                                                                                                                                                                                                                                                          

            __LINKEDIT ---------- 1000ad000
                       |
                       |
                       |
                       ---------- 100101000

IN-FILE LAYOUT:
            __PAGEZERO ---------- 0
                       |
                       |
                       |
                       ---------- 0

                __TEXT ---------- 0
                       |
                       |
                       |
                       ---------- 2000

         __DATA_CONST1 ---------- 2000
                       |

                 __const -------- 2000
                         |
                         -------- 20820

                       |
                       ---------- 21000

          __DATA_CONST ---------- 27000
                       |

                   __got -------- 27000
                         |
                         -------- 2e650

                       |
                       ---------- 2f000

               __TEXT2 ---------- 58000
                       |

                 __stubs -------- 58000
                         |
                         -------- 5800c

                       |
                       ---------- 59000

               __TEXT3 ---------- 59000
                       |

           __stub_helper -------- 59000
                         |
                         -------- 59023

                       |
                       ---------- 5a000

                __DATA ---------- 5c000
                       |

         __la_symbol_ptr -------- 5c000
                         |
                         -------- 5c010

                       |
                       ---------- 5d000

               __DATA1 ---------- 5d000
                       |

                  __data -------- 5d000
                         |
                         -------- 5d190

                       |
                       ---------- 5e000

               __TEXT1 ---------- 5e000
                       |

                  __text -------- 5e000
                         |
                         -------- c51e5

                       |
                       ---------- c6000

            __LINKEDIT ---------- c6000
                       |
                       |
                       |
                       ---------- 119acf
```

As you can hopefully see from the enclosed printout above, the location of sections/segments in file doesn't follow the ordering in memory which is precisely how we want it to be for the incremental linker. Additionally, to achieve this, we enclose a single section within a single segment which is unusual in MachO files but gives us ability to easily move sections in file whenever required. This being unusual, however, means that `dyld` loader (and more generally existing Apple tooling) needed some persuasion to accept. If you look closely at the printout above, every segment has to have a unique name; otherwise, `dyld` will not accept the binary as valid. Another limitation presented to us by Apple is the fact that we are effectively limited in the number of segments to `u4` so 16, but effectively 13 (since we always require a `__PAGEZERO` for executables, `__TEXT` enclosing the headers for all images, and `__LINKEDIT` carrying `dyld` opcodes and other metadata). I believe this should be fine for our purposes of incremental linking in Debug as we don't need to create so many aux sections such as `__TEXT,__cstring` or `__TEXT,__const` - instead, both can happily live in `__DATA_CONST,__const`. Another trick we might look into if we run out of segments to allocate is to look into an old-fashioned model for dynamic relocations which is based on embedding actual `macho.relocation_info` as part of the dynamic symbol table. Alternatively, we can also collapse the synthetic sections such as `__stubs` and `__stub_helper` into a single segment.

The next step after landing this change will be refactoring full/one-shot (or traditional) linking context. I have already partially started doing this by putting any functionality applicable only to one-shot linking context into a standalone namespace `MachO/zld.zig`. Separation of contexts will also allow for more clever optimisations that apply to one context but not the other. I will working on this next which will hopefully clean up the code even more.

### DIFF
--- a/lib/std/macho.zig
+++ b/lib/std/macho.zig
@@ -798,6 +798,11 @@ pub const section_64 = extern struct {
         return tt == S_ZEROFILL or tt == S_GB_ZEROFILL or tt == S_THREAD_LOCAL_ZEROFILL;
     }
 
+    pub fn isSymbolStubs(sect: section_64) bool {
+        const tt = sect.@"type"();
+        return tt == S_SYMBOL_STUBS;
+    }
+
     pub fn isDebug(sect: section_64) bool {
         return sect.attrs() & S_ATTR_DEBUG != 0;
     }
--- a/src/arch/aarch64/Emit.zig
+++ b/src/arch/aarch64/Emit.zig
@@ -680,16 +680,15 @@ fn mirCallExtern(emit: *Emit, inst: Mir.Inst.Index) !void {
             break :blk offset;
         };
         // Add relocation to the decl.
-        const atom = macho_file.atom_by_index_table.get(relocation.atom_index).?;
+        const atom = macho_file.getAtomForSymbol(.{ .sym_index = relocation.atom_index, .file = null }).?;
         const target = macho_file.getGlobalByIndex(relocation.sym_index);
-        try atom.relocs.append(emit.bin_file.allocator, .{
-            .offset = offset,
+        try atom.addRelocation(macho_file, .{
+            .@"type" = @enumToInt(std.macho.reloc_type_arm64.ARM64_RELOC_BRANCH26),
             .target = target,
+            .offset = offset,
             .addend = 0,
-            .subtractor = null,
             .pcrel = true,
             .length = 2,
-            .@"type" = @enumToInt(std.macho.reloc_type_arm64.ARM64_RELOC_BRANCH26),
         });
     } else {
         return emit.fail("Implement call_extern for linking backends != MachO", .{});
@@ -882,13 +881,13 @@ fn mirLoadMemoryPie(emit: *Emit, inst: Mir.Inst.Index) !void {
     }
 
     if (emit.bin_file.cast(link.File.MachO)) |macho_file| {
-        const atom = macho_file.atom_by_index_table.get(data.atom_index).?;
-        // Page reloc for adrp instruction.
-        try atom.relocs.append(emit.bin_file.allocator, .{
-            .offset = offset,
+        const atom = macho_file.getAtomForSymbol(.{ .sym_index = data.atom_index, .file = null }).?;
+        // TODO this causes segfault in stage1
+        // try atom.addRelocations(macho_file, 2, .{
+        try atom.addRelocation(macho_file, .{
             .target = .{ .sym_index = data.sym_index, .file = null },
+            .offset = offset,
             .addend = 0,
-            .subtractor = null,
             .pcrel = true,
             .length = 2,
             .@"type" = switch (tag) {
@@ -901,12 +900,10 @@ fn mirLoadMemoryPie(emit: *Emit, inst: Mir.Inst.Index) !void {
                 else => unreachable,
             },
         });
-        // Pageoff reloc for adrp instruction.
-        try atom.relocs.append(emit.bin_file.allocator, .{
-            .offset = offset + 4,
+        try atom.addRelocation(macho_file, .{
             .target = .{ .sym_index = data.sym_index, .file = null },
+            .offset = offset + 4,
             .addend = 0,
-            .subtractor = null,
             .pcrel = false,
             .length = 2,
             .@"type" = switch (tag) {
--- a/src/arch/aarch64/Emit.zig
+++ b/src/arch/aarch64/Emit.zig
@@ -845,8 +845,7 @@ fn mirLoadMemoryPie(emit: *Emit, inst: Mir.Inst.Index) !void {
     try emit.writeInstruction(Instruction.adrp(reg.to64(), 0));
 
     switch (tag) {
-        .load_memory_got,
-        => {
+        .load_memory_got => {
             // ldr reg, reg, offset
             try emit.writeInstruction(Instruction.ldr(
                 reg,
--- a/src/arch/aarch64/Emit.zig
+++ b/src/arch/aarch64/Emit.zig
@@ -845,7 +845,8 @@ fn mirLoadMemoryPie(emit: *Emit, inst: Mir.Inst.Index) !void {
     try emit.writeInstruction(Instruction.adrp(reg.to64(), 0));
 
     switch (tag) {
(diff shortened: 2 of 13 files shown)


kubkon:
I double checked locally on my Windows machine and the tests pass without fail which probably means we have some latent bug on our hands. I will keep an eye on it and try to investigate as I go along.

PR #12894, opened 2022-09-18
adding `static` and `dynamic` ZigEquivalent enums so that `zig cc` can force static or dynamic linking.

### LINKED ISSUE #11909: [zig cc] unable to dynamically link musl executables
### Zig Version

0.10.0-dev.2473+e498fb155

Like the title says: zig cc frontend cannot compile a dynaimcally-linked musl binary. Created a separate issue as requested by @ifreund in https://github.com/ziglang/zig/issues/5364#issuecomment-1163025810

### Full repro

**main.c**

```
#include <stdio.h>
#include <features.h>

int main() {
    #ifdef __GLIBC__
    printf("glibc_%d.%d\n", __GLIBC__, __GLIBC_MINOR__);
    #else
    printf("non-glibc\n");
    #endif
    return 0;
}
```
Compiling:
```
$ zig cc -v -target x86_64-linux-musl main.c -fPIE -lc -dynamic -o main
<...>
LLD Link... ld.lld -error-limit=0 -O0 -z stack-size=16777216 --gc-sections --eh-frame-hdr -znow -m elf_x86_64 -static -pie -o /home/motiejus/.cache/zig/o/503c6bf2ee43b2d33014a8d71f6bc1d2/main /home/motiejus/.cache/zig/o/60b545a849e93516862f973801e6307b/rcrt1.o /home/motiejus/.cache/zig/o/1f9266dba05df20689bccc6b184442ec/crti.o /home/motiejus/.cache/zig/o/6471a2fd822202eae06273e85f08f5c6/main.o /home/motiejus/.cache/zig/o/baf786ce8c76ce866d19a60d75637d88/libcompiler_rt.a --as-needed /home/motiejus/.cache/zig/o/a111f160cb9815a5d7c8128bcf0c7af8/libc.a /home/motiejus/.cache/zig/o/c8866bc9544d3faae13fa04a4728ac91/crtn.o --allow-shlib-undefined
```

Inspect the resulting `main`. It is static:
```
$ objdump --dynamic-syms main
main:     file format elf64-x86-64

DYNAMIC SYMBOL TABLE:
no symbols


$ file main
main: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), static-pie linked, with debug_info, not stripped
```

Looking at the linker line above, you can see that `-static` is passed to the linker, and at least `--dynamic-linker` option is missing.

cod1r:
When adding the flag `-dynamic`, `zig cc` would still statically link by default because `link_mode` was set to `null`.

Adding the ZigEquivalent enums `static` and `dynamic` allows us to branch properly to set `link_mode` when we iterate over the arguments passed into `ClangArgIterator` when `arg_mode` is equivalent to `.cc`.

This should close #11909.

Documentation used: [Clang CLI Reference](https://clang.llvm.org/docs/ClangCommandLineReference.html) and zig's source code.


src/clang_options_data.zig
```diff
@@ -2480,7 +2480,14 @@ flagpd1("dwarf-ext-refs"),
 },
 sepd1("dylib_file"),
 flagpd1("dylinker"),
-flagpd1("dynamic"),
+.{
```
kubkon, line 2483: To action this change, you should not modify `clang_options_data.zig` manually but instead you should put the new mapping in `tools/update_clang_options.zig`, and then re-run the tool pointing at `clang_options_data.zig` as the output file.

src/clang_options_data.zig
```diff
@@ -5034,7 +5041,7 @@ sepd1("stack-usage-file"),
 .{
     .name = "static",
     .syntax = .flag,
-    .zig_equivalent = .other,
+    .zig_equivalent = .static,
```
kubkon, line 5044: Likewise.

src/clang_options_data.zig
kubkon: For example, here's a simple patch to `tools/update_clang_options.zig` that would achieve this:

```diff
diff --git a/tools/update_clang_options.zig b/tools/update_clang_options.zig
index 92e0757ac..2109de695 100644
--- a/tools/update_clang_options.zig
+++ b/tools/update_clang_options.zig
@@ -492,6 +492,10 @@ const known_options = [_]KnownOpt{
         .name = "compress-debug-sections=",
         .ident = "compress_debug_sections",
     },
+    .{
+        .name = "dynamic",
+        .ident = "dynamic",
+    },
 };
 
 const blacklisted_options = [_][]const u8{};
```

kubkon (changes requested):
Thanks for this change! In order to land this, we need to fix how changes to `src/clang_options_data.zig` are generated which actually happens automatically via running the tool `tools/update_clang_options.zig`. Have a look at the review comments and lemme know if that makes sense to you!

src/clang_options_data.zig
cod1r: trying to run update_clang_options.zig, results in the compiler segfaulting or reaching `unreachable`.

cod1r: never mind I got it to work

cod1r:
apologies for the messy commits. there are a lot of git features I have yet to use. I think I made the right changes in the right files like you asked. if not, I'll be happy to do more. :)

andrewrk:
This change may have been correct as is. If so, it exposed a bug in LLVM's cmake configuration which incorrectly adds `-static` to the CXX flags on Windows when doing `-DLLVM_BUILD_STATIC=ON`.

However, even if this is the case, more investigation is needed. An upstream bug report needs to be created, and we need a workaround so that [zig-bootstrap](https://github.com/ziglang/zig-bootstrap) continues to work.

Another possibility is that the `-static` flag of C compilers does not influence whether the resulting binary is statically linked, but rather influences what search paths `-l` arguments look for, i.e., it would make `-lfoo` only look for `libfoo.a`. On Windows, this could still be an import lib, and the executable is still dynamically linked. This would mean that setting `link_mode`, as done in this PR, would be incorrect.

PR #12897, opened 2022-09-18
std.os.uefi: fix some mistakes

r00ster91:
This is an attempt to make `std.os.uefi` work with self-hosted (stage2); it is currently broken.

With this PR, the following now works:
```zig
const std = @import("std");

pub fn main() void {
    std.testing.refAllDeclsRecursive(std.os.uefi);
}
```
```
$ zig run x.zig --zig-lib-dir lib
$
```

However, it doesn't work with stage1:
```
$ zig run x.zig --zig-lib-dir lib -fstage1
./lib/std/os/uefi/protocols/simple_text_output_protocol.zig:7:38: error: struct 'std.os.uefi.protocols.simple_text_output_protocol.SimpleTextOutputProtocol' depends on itself
pub const SimpleTextOutputProtocol = extern struct {
                                     ^
./lib/std/os/uefi/tables/system_table.zig:27:5: note: while checking this field
    con_out: ?*SimpleTextOutputProtocol,
    ^
./lib/std/os/uefi/protocols/loaded_image_protocol.zig:10:33: error: struct 'std.os.uefi.protocols.loaded_image_protocol.LoadedImageProtocol' depends on itself
pub const LoadedImageProtocol = extern struct {
                                ^
```
This error only happens with the old compiler. I can't really figure out why. It only started happening after I added `std.meta.FnPtr`.
So I came here to get some feedback: is it fine to drop stage1 support for `std.os.uefi`? If so, we could also just use `*const Fn` directly instead of `std.meta.FnPtr(Fn)` where `std.meta.FnPtr` is supposed to serve for compatibility with both stage1 and stage2.

Also, the reason I changed all these `packed struct`s to `extern struct`s is because some of them were containing arrays and as we all know `packed struct`s can no longer have them so I thought that maybe these should all be `extern struct`s anyway so I just changed them all.
This might be wrong, which is why I'm asking for feedback or reviews.

CC @fifty-six

### DIFF
--- a/lib/std/os/uefi/protocols/managed_network_protocol.zig
+++ b/lib/std/os/uefi/protocols/managed_network_protocol.zig
@@ -15,7 +15,7 @@ pub const ManagedNetworkProtocol = extern struct {
     _transmit: std.meta.FnPtr(fn (*const ManagedNetworkProtocol, *const ManagedNetworkCompletionToken) callconv(.C) Status),
     _receive: std.meta.FnPtr(fn (*const ManagedNetworkProtocol, *const ManagedNetworkCompletionToken) callconv(.C) Status),
     _cancel: std.meta.FnPtr(fn (*const ManagedNetworkProtocol, ?*const ManagedNetworkCompletionToken) callconv(.C) Status),
-    _poll: std.meta.FnPtr(fn (*const ManagedNetworkProtocol) callconv(.C) usize),
+    _poll: std.meta.FnPtr(fn (*const ManagedNetworkProtocol) callconv(.C) Status),
 
     /// Returns the operational parameters for the current MNP child driver.
     /// May also support returning the underlying SNP driver mode data.
--- a/lib/std/os/uefi/protocols/managed_network_protocol.zig
+++ b/lib/std/os/uefi/protocols/managed_network_protocol.zig
@@ -31,8 +31,7 @@ pub const ManagedNetworkProtocol = extern struct {
     /// Translates an IP multicast address to a hardware (MAC) multicast address.
     /// This function may be unsupported in some MNP implementations.
     pub fn mcastIpToMac(self: *const ManagedNetworkProtocol, ipv6flag: bool, ipaddress: *const anyopaque, mac_address: *MacAddress) Status {
-        _ = mac_address;
-        return self._mcast_ip_to_mac(self, ipv6flag, ipaddress);
+        return self._mcast_ip_to_mac(self, ipv6flag, ipaddress, mac_address);
     }
 
     /// Enables and disables receive filters for multicast address.
--- a/lib/std/os/uefi/protocols/simple_text_input_ex_protocol.zig
+++ b/lib/std/os/uefi/protocols/simple_text_input_ex_protocol.zig
@@ -10,7 +10,7 @@ pub const SimpleTextInputExProtocol = extern struct {
     _read_key_stroke_ex: std.meta.FnPtr(fn (*const SimpleTextInputExProtocol, *KeyData) callconv(.C) Status),
     wait_for_key_ex: Event,
     _set_state: std.meta.FnPtr(fn (*const SimpleTextInputExProtocol, *const u8) callconv(.C) Status),
-    _register_key_notify: std.meta.FnPtr(fn (*const SimpleTextInputExProtocol, *const KeyData, fn (*const KeyData) callconv(.C) usize, **anyopaque) callconv(.C) Status),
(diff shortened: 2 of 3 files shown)


lib/std/os/uefi/protocols/device_path_protocol.zig
```diff
@@ -363,14 +363,14 @@ pub const MessagingDevicePath = union(Subtype) {
         }
     };
 
-    pub const DeviceLogicalUnitDevicePath = packed struct {
+    pub const DeviceLogicalUnitDevicePath = extern struct {
         type: DevicePathType,
         subtype: Subtype,
         length: u16,
         lun: u8,
     };
```
leecannon, line 371: With this change, there is a difference with these structs compared to before.

The prior `packed` version would not have had padding but these extern ones do, unless the alignment of certain fields is forced to be smaller.

```zig
// @sizeOf() == 6
pub const DeviceLogicalUnitDevicePath = extern struct {
    type: DevicePathType,
    subtype: Subtype,
    length: u16,
    lun: u8,
};

// @sizeOf() == 5
pub const DeviceLogicalUnitDevicePathWithAlignment = extern struct {
    type: DevicePathType,
    subtype: Subtype,
    length: u16 align(1),
    lun: u8,
};
```
`DeviceLogicalUnitDevicePath` was chosen as an example, this affects other structs too.

lib/std/os/uefi/protocols/device_path_protocol.zig
```diff
@@ -454,9 +454,7 @@ pub const MessagingDevicePath = union(Subtype) {
             console_protocol: bool,
             storage_protocol: bool,
             network_protocol: bool,
-
-            // u1 + 4 * bool = 5 bits, we need a total of 32 bits
-            reserved: u27,
```
fifty-six, line 459: This should still be u27, the mistake was that the ControllerType should be `enum(u1)`

fifty-six:
Yeah like leecannon mentioned, [the old packed structs are equivalent to extern structs with all align(1)](https://github.com/ziglang/zig/issues/10113#issuecomment-1241791584). That's also the main reason I haven't gotten around to it, wanted to make some meta function to add the aligns instead of changing every struct because I am lazy

r00ster91:
Ah, okay! I see now. I figured that it would probably be more complicated. In that case I will exclude the bit size stuff for now so that we can add such a meta function and do the rest some other time.

r00ster91:
Now the issue with the stage1-only `depends on itself` error (see PR description) still stands though. Is it fine if we make the UEFI functions only available on stage2?

fifty-six:
Alright I've written one now so I'll go ahead and PR it and then we can use that if it goes through.

r00ster91:
I totally missed that #12761 already does part of this apparently. My PR still contains some other UEFI-specific fixes though. I guess either will have to rebase.

fifty-six:
Would you also be interested in me PRing your fork with the changes to the structs (maybe with a rebase to use CPacked) while we wait on that?

r00ster91:
Sure! I was kinda waiting for #12899 to be accepted/merged but I'm assuming it will probably be accepted in roughly that form, even if only temporarily until the language catches up and it will be possible to do that on a language level.

fifty-six:
Ah given the issue with decls I might wait for something else to show up - was just considering working on my project so I'd need a local fork to get it building anyways and figured it'd be good to pr if I did so

r00ster91:
Do we want to open a separate issue to bring more attention to this? I don't actually know of an issue that talks about this.
What should the syntax look like you think? `extern struct align(1) {}`?

fifty-six:
Yeah, someone mentioned similar syntax in #10113 but I don't know of any issue for it.

r00ster91:
@MasterQ32 
> It's planned to have some kind of syntax like `extern struct align(1) { … }` which would be equivalent to the C version

Can you elaborate on this plan? Like who is planning this? Is anyone keeping track of this or already working on it?

fifty-six:
Ah ok that sounds ideal

ominitay:
> Would we want to add align(1) to every field while we don't have language support for something similar? It's a bit of a pain but as-is uefi is largely just unusable under stage 2 due to the lack of it.

Yes we do; as UEFI has Tier 2 support, we shouldn't be allowing it to regress.

andrewrk:
Thanks!

r00ster91:
Yeah, otherwise, if it doesn't look like #13009 is going to happen anytime soon, we might as well add `align(1)` to every struct with a regex or something. Then we can revert that commit once we have #13009.

Issue #6709, opened 2020-10-17
Allow declarations in @Type for Struct/Union/Enum/Opaque


tadeokondrak:
The compiler currently errors when the `.decls` slice is non-empty.

More context at https://github.com/ziglang/zig/issues/383.

MasterQ32:
One (imho strong) argument to allow this is that we can make really convenient interface implementations:
```zig
// Definition of an interface:
const Allocator = std.meta.Interface(struct {
    pub const Error = error{OutOfMemory};
    alloc: fn (self: *std.meta.Self, len: usize) Error![]u8,
    free: fn(self: *std.meta.Self, ptr: []u8) void,
});

// Usage is just taking the interface type, it's a "fat pointer":
fn process(allocator: Allocator, items: []const u8) !void {
    // Just call functions on the interface like on normal objects.
    // this is provided via created functions from .decls
    const buf = try allocator.alloc(items.len);
    defer allocator.free(buf);
    
    …
}

// Implementing interfaces is still duck typing:
const FixedBufferAllocator = struct {
    const Self = @This();

    buffer: []u8,
    allocated: bool = false,
    
    pub fn init(buf: []u8) Self {
        return Self { .buffer = buf };
    }

                                          // access public symbols of the struct we pass to std.meta.Interface
                                          // as we can usingnamespace them
    pub fn alloc(self: *Self, size: usize) Allocator.Error!void {
        if(self.allocated)
            return error.OutOfMemory;
        if(size > self.buffer.len)
            return error.OutOfMemory;
        self.allocated = true;
        return self.buffer[0..size];
    }

    pub fn free(self: *Self, ptr: []u8) void {
        std.debug.assert(ptr.ptr == self.buffer.ptr);
        self.allocated = false;
    }
};

fn main() !void {
    var fba = FixedBufferAllocator.init(&some_array);
    
    // Interface().get(ptr) will duck-type match the interface into a fat pointer
    try process(Allocator.get(&fba), "Hello, Zig");
}
```

To make the convenient function calls like `allocator.free(…)` we need `TypeInfo.Struct.decls`.

The neat thing is: The example code above does feel native, like if interfaces would be a part of the language. But they are just a clever use of `comptime`. This would kinda also close #130 as "implemented"

jamii:
A minor usecase - when translate-c can't handle something I make a wrapper function instead:

```
[nix-shell:~/bluetron/blinky]$ cat wrapper.c
#include "stdbool.h"
[...]

PR #12899, opened 2022-09-18
Add meta function CPacked to add align(1) to extern structs

fifty-six:
This allows for easier FFI with C APIs that use large amounts of packed structs by allowing the omission of align(1) on every field declaration.

Motivation behind this being for UEFI ffi as almost every struct there is packed.

### DIFF
--- a/lib/std/zig/c_translation.zig
+++ b/lib/std/zig/c_translation.zig
@@ -109,6 +109,67 @@ fn ptrInfo(comptime PtrType: type) std.builtin.Type.Pointer {
     };
 }
 
+/// Gives a type equivalent to a C packed struct
+/// by changing the alignment of all fields to 1 unless
+/// explicitly specified using align() *unless it matches natural alignment*
+/// in which case it can be specified using the overrides parameter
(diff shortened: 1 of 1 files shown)


r00ster91:
Sadly this isn't actually going to work if the extern struct has any decls inside it...
```zig
    const S3 = CPacked(extern struct {
        a: u8,
        b: u8,
        c: u16,
        d: u8,
        fn x() void {}
    }, .{ .b = @alignOf(u16) });
```
```
x.zig:149:12: error: reified structs must have no decls
    return @Type(.{
           ^~~~~
x.zig:168:23: note: called from here
    const S3 = CPacked(extern struct {
               ~~~~~~~^
```
Yet another reason for https://github.com/ziglang/zig/issues/6709

So I don't think this will work as a permanent solution. It would work as a better temporary one if #6709 was solved but ultimately the language needs an addition to make this easier.

fifty-six:
if anyone is interested in this for just simple data I can reopen but without decls I'm not sure this is great

Issue #12891, opened 2022-09-18
Compiler segfaults when comparing const comptime float with var usize


verdog:
### Zig Version

0.10.0-dev.4060+61aaef0b0

### Steps to Reproduce

`zig test repro.zig`
```zig
// repro.zig
test {
    const f = 10.0;
    var i: usize = 0;
    if (i < f)
        @import("std").debug.print("{} {}\n", .{ f, i });
}
```

### Expected Behavior

The test compiles and passes.

### Actual Behavior

```
⇒  zig test repro.zig
[1]    116413 segmentation fault (core dumped)  zig test repro.zig
```

davidgm94:
Stack trace:
```
thread 26015 panic: reached unreachable code
/home/david/dev/official-zig/src/value.zig:1113:21: 0x557df686b5f7 in value.Value.toBigIntAdvanced (zig2)
            else => unreachable,
                    ^
/home/david/dev/official-zig/src/value.zig:1058:36: 0x557df6661d7a in value.Value.toBigInt (zig2)
        return val.toBigIntAdvanced(space, target, null) catch unreachable;
                                   ^
/home/david/dev/official-zig/src/Sema.zig:27299:46: 0x557df6e6865a in Sema.cmpNumeric (zig2)
            var bigint = try rhs_val.toBigInt(&bigint_space, target).toManaged(sema.gpa);
                                             ^
/home/david/dev/official-zig/src/Sema.zig:13803:31: 0x557df6c5e000 in Sema.analyzeCmp (zig2)
        return sema.cmpNumeric(block, src, lhs, rhs, op, lhs_src, rhs_src);
                              ^
/home/david/dev/official-zig/src/Sema.zig:13778:27: 0x557df69c26d7 in Sema.zirCmp (zig2)
    return sema.analyzeCmp(block, src, lhs, rhs, op, lhs_src, rhs_src, false);
                          ^
/home/david/dev/official-zig/src/Sema.zig:729:61: 0x557df681d148 in Sema.analyzeBodyInner (zig2)
            .cmp_lt                       => try sema.zirCmp(block, inst, .lt),
                                                            ^
/home/david/dev/official-zig/src/Sema.zig:4972:34: 0x557df6c7cf56 in Sema.resolveBlockBody (zig2)
        if (sema.analyzeBodyInner(child_block, body)) |_| {
                                 ^
[...]

PR #12901, opened 2022-09-19
Sema: add float128IntPartToBigInt to fix compare comptime float with int

kkHAIKE:
fix https://github.com/ziglang/zig/issues/12891


src/Sema.zig
```diff
@@ -30565,6 +30563,30 @@ fn floatToInt(
...
+    arena: Allocator,
+    float: f128,
+) !std.math.big.int.Managed {
+    const is_negative = std.math.signbit(float);
+    const floored = @floor(@fabs(float));
+
+    var rational = try std.math.big.Rational.init(arena);
+    defer rational.q.deinit();
+    rational.setFloat(f128, floored) catch |err| switch (err) {
+        error.NonFiniteFloat => unreachable,
```
topolarity, line 30576: Maybe add a doc-comment that `float` is expected to be finite and non-NaN?

Looks like the `inf`/`nan` cases need more changes before they'll work

src/Sema.zig
```diff
@@ -27734,8 +27749,7 @@ fn cmpNumeric(
             }
         }
         if (rhs_is_float) {
-            var bigint_space: Value.BigIntSpace = undefined;
-            var bigint = try rhs_val.toBigInt(&bigint_space, target).toManaged(sema.gpa);
+            var bigint = try float128IntPartToBigInt(sema.gpa, rhs_val.toFloat(f128));
```
topolarity, line 27752: Looking at this code fresh, I'm not sure why we need to construct this huge bigint in the first place.

If the float has a fractional part, the coercion to an integer later will fail anyway, right?

src/Sema.zig
```diff
@@ -27673,6 +27673,14 @@ fn cmpNumeric(
     if (try sema.resolveMaybeUndefVal(block, lhs_src, lhs)) |lhs_val| {
         if (lhs_val.isUndef())
             return sema.addConstUndef(Type.bool);
+        if (lhs_val.isNan()) switch (op) {
```
topolarity, line 27676: The following tests fail when they should pass:
```zig
test "-inf < 0" {
    const f = comptime -std.math.inf(f64);
    var i: usize = 0;
    try std.testing.expect(f < i);
}
test "inf >= 1" {
    const f = comptime std.math.inf(f64);
    var i: usize = 1;
    try std.testing.expect(f >= i);
}
```

src/value.zig
```diff
@@ -1999,6 +1999,11 @@ pub const Value = extern union {
                 }
                 return true;
             },
```
topolarity, line 2001: This condition is reversed:
```zig
test "nan * 0 == nan" {
    const nan_times_zero = comptime std.math.nan(f64) * 0;
    try std.testing.expect(std.math.isNan(nan_times_zero));
}
```

topolarity (commented):
Sorry for the late review, but I noticed a few things worth fixing

src/value.zig
kkHAIKE: i think fixing it is out of scope for this PR...

kkHAIKE: i think need a PR to modify all operation with NaN/Inf..

PR #12903, opened 2022-09-19
GPA: return true when deinit is successful

clebs:
- gpa should return true when `deinit` is successful. Returning false when successful is counterintuitive and confusing.

### DIFF
--- a/lib/std/heap/general_purpose_allocator.zig
+++ b/lib/std/heap/general_purpose_allocator.zig
@@ -423,7 +423,7 @@ pub fn GeneralPurposeAllocator(comptime config: Config) type {
             }
             self.large_allocations.deinit(self.backing_allocator);
             self.* = undefined;
-            return leaks;
+            return !leaks;
         }
 
         fn collectStackTrace(first_trace_addr: usize, addresses: *[stack_n]usize) void {
@@ -922,7 +922,7 @@ const test_config = Config{};
 
 test "small allocations - free in same order" {
     var gpa = GeneralPurposeAllocator(test_config){};
-    defer std.testing.expect(!gpa.deinit()) catch @panic("leak");
+    defer std.testing.expect(gpa.deinit()) catch @panic("leak");
     const allocator = gpa.allocator();
 
     var list = std.ArrayList(*u64).init(std.testing.allocator);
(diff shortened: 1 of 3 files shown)


tauoverpi:
This feels like a case that would be better served by an `enum { success, leaks_detected }` or similar as then it becomes clearer at the expense of having to type a bit more. The bool is easy to trip up on as it doesn't communicate much.

clebs:
Thanks for the feedback @tauoverpi!
If it is good I would apply the enum based solution you suggested.

lib/std/heap/general_purpose_allocator.zig
```diff
@@ -416,14 +418,14 @@ pub fn GeneralPurposeAllocator(comptime config: Config) type {
...
-        pub fn deinit(self: *Self) bool {
+        pub fn deinit(self: *Self) DeinitResult {
             const leaks = if (config.safety) self.detectLeaks() else false;
             if (config.retain_metadata) {
                 self.freeRetainedMetadata();
             }
             self.large_allocations.deinit(self.backing_allocator);
             self.* = undefined;
-            return leaks;
+            return if (leaks) DeinitResult.leaking else DeinitResult.success;
```
matu3ba, line 428: you can optimize away this branch with boolToInt and intToEnum.

lib/std/heap/general_purpose_allocator.zig
```diff
@@ -155,6 +155,8 @@ pub const Config = struct {
     verbose_log: bool = false,
 };
 
+pub const DeinitResult = enum { success, leaking };
```
matu3ba, line 158: `deinit()` is the general pattern for every dynamic resource in Zig, so it bears little additional knowledge.
Likewise, `Result` does neither tell a lot on what is actually done.

Something like
```zig
pub const Check = enum { ok, leak };
```
sounds less annoying to type.

External references are then unambiguous as `alloc.check.ok` or `std.heap.check.ok` and this sounds also okayish to write out, if one wants to be explicit.

matu3ba (commented):
counter-proposal to syntax.

I dont have a very strong opinion on making things more explicit. Personally I like to always double `check` dynamic things, so I would prefer something short and nice alike `Check`.

andrewrk:
> GPA: return true when deinit is successful

the widespread convention is 0 for success and non-zero for failure.

as for using an enum, please go through the proposal process for making this particular breaking change to the std lib API. that is, file an issue, and wait until it has "accepted" label before making a pull request.

clebs:
Thanks for the clarification @andrewrk, I opened the following issue to propose a new idea for this:
#12906

PR #12904, opened 2022-09-19
stage2: pointer casting fixes and improvements

Vexu:


andrewrk:
> Bus error

It looks like one reason this occurs is due to using a processor instruction with an address that does not satisfy alignment requirements.

It's not clear to me why `align(1)` needed to be added to some of these pointer casts. Can we examine those cases more closely?

andrewrk:
I think the instances of `align(1)` added to the std lib are revealing flaws, but fixing those can be a separate improvement. Nice work.

PR #12905, opened 2022-09-19
zig cc: emit to a.out if outpath is unspecified

### LINKED ISSUE #12858: "zig cc a.c" produces no output file, which breaks autoconf (probably an LLVM 15.0.0 regression)
### Zig Version

0.10.0-dev.4060+61aaef0b0

### Steps to Reproduce

1. Create a file a.c with contents `int main() { return 0; }`
2. Type `zig cc a.c`
3. NO FILE IS CREATED

### Expected Behavior

The file `a.out` should be created.

### Actual Behavior

No output file at all is created.

I noticed this when trying to use autoconf to build a simple C program (lzma in this case), but it would immediately break pretty much all use of autoconf, whose most basic test is to build a trivial`conftest.c`  and check for one of the following output files (copied from autoconf source code);
```
# The possible output files:
ac_files="a.out conftest.exe conftest a.exe a_out.exe b.out conftest.*"
```

Anyway, this is a really serious bug.  I've rarely hit a bug I couldn't work around, but this would be painful, and require writing wrapper scripts to replace `zig cc a.c` with `zig cc a.c -o a.out`, etc.  That is a bit too much to get right.

motiejus:
If `-o` is not specified for `zig cc` for linking, it should emit the `a.out` (or equivalent) to the current directory. So then this:

    $ zig cc test.c

Should result in `./a.out`.

If directory is omitted from `Compilation.EmitLoc`, it will default to the cache directory later in the compilation chain, which will result in `a.out` missing in the current directory.

If we specify the directory to `Compilation.EmitLoc`, it will be handled, allegedly, correctly.

Fixes #12858

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -2613,7 +2613,7 @@ fn buildOutputType(
             }
         },
         .yes_a_out => Compilation.EmitLoc{
-            .directory = null,
+            .directory = .{ .path = null, .handle = fs.cwd() },
             .basename = a_out_basename,
         },
     };


andrewrk:
Excellent diagnosis and fix :+1:

PR #12909, opened 2022-09-20
std/meta: make FieldEnum(T) == Tag(T) for tagged unions

silversquirl:
This is a minor gripe I've had with `FieldEnum` a couple times - this change just makes things a little more consistent, and allows functions like the following to be a little nicer:

```zig
fn getUnionField(x: U, comptime f: std.meta.FieldEnum(U)) ?std.meta.fieldInfo(U, f).field_type {
    if (x != f) return null;
    return @field(x, @tagName(f));
}
```

Whereas before, `x != f` would be a compile error due to it not being the correct enum, requiring extra casts.

### DIFF
--- a/lib/std/meta.zig
+++ b/lib/std/meta.zig
@@ -606,6 +606,18 @@ test "std.meta.tags" {
 
 pub fn FieldEnum(comptime T: type) type {
     const field_infos = fields(T);
+
+    if (@typeInfo(T) == .Union) {
+        if (@typeInfo(T).Union.tag_type) |tag_type| {
+            for (std.enums.values(tag_type)) |v, i| {
+                if (@enumToInt(v) != i) break; // enum values not consecutive
+                if (!std.mem.eql(u8, @tagName(v), field_infos[i].name)) break; // fields out of order
+            } else {
+                return tag_type;
+            }
+        }
+    }
+
     var enumFields: [field_infos.len]std.builtin.Type.EnumField = undefined;
     var decls = [_]std.builtin.Type.Declaration{};
     inline for (field_infos) |field, i| {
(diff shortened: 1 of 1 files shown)


InKryption:
I don't think this is necessarily a good idea. I like how it is right now, because it is consistent for all valid input types that the enum will contain values ranging from `0` to `fields.len - 1`, which is both a sane and a useful property. With this, the values of the resulting enum will depend on the input type.

As for your use case, I think it would be better suited to have it just use the union's tag type directly, e.g.
```zig
fn getUnionField(x: U, comptime f: std.meta.Tag(U)) ?std.meta.fieldInfo(U, f).field_type {
    if (x != f) return null;
    return @field(x, @tagName(f));
}
```

Worst case scenario, if you absolutely have to use `std.meta.FieldEnum(U)` for whatever reason, you can easily do one of:
```zig
_ = std.enums.nameCast(std.meta.Tag(U), @tagName(f));
_ = std.meta.stringToEnum(std.meta.Tag(U), @tagName(f)).?;
```

silversquirl:
> I don't think this is necessarily a good idea. I like how it is right now, because it is consistent for all valid input types that the enum will contain values ranging from 0 to fields.len - 1, which is both a sane and a useful property. With this change, the values of the resulting enum will depend on the input type.

Ah, yes. I hadn't considered that.

> As for your use case, I think it would be better suited to have it just use the union's tag type directly, e.g.

That doesn't work because `fieldInfo` takes a `FieldEnum`.

Perhaps adding an additional check to ensure the enum values match would work?

InKryption:
> That doesn't work because `fieldInfo` takes a `FieldEnum`.

Ah, fair. Actually, now thinking about it, maybe it shouldn't take a `FieldEnum`; it would be quite a lot simpler for the function to simply take a string, which any of the enums we see in play here could cast to using `@tagName` - same as `declarationInfo` (which takes a string, despite us having `DeclEnum`).
Maybe you could consider opening another PR for that, or just doing it in this one - would make it far more of a breaking change though.

> Perhaps adding an additional check to ensure the enum values match would work?

I think that's a fair middle ground here.

PR #12914, opened 2022-09-20
Sema: check that reified enum field values fit tag type

### LINKED ISSUE #12908: `@Type()` not catching the invalid enum tag_type
### Zig Version

0.10.0-dev.4060+61aaef0b0

### Steps to Reproduce

```zig
const std = @import("std");

// enum created with @Type doesn't reject tag_type which is 
// too small. the extra field is kept and @enumToInt seems 
// to truncate
test "bad enum" {
    const E = @Type(.{
        .Enum = .{
            .layout = .Auto,
            // invalid - too small
            .tag_type = u1, 
            .fields = &[_]std.builtin.Type.EnumField{
                .{ .name = "f0", .value = 0 },
                .{ .name = "f1", .value = 1 },
                .{ .name = "f2", .value = 2 },
            },
            .decls = &.{},
            .is_exhaustive = true,
        },
    });
    for ([_]E{ .f0, .f1, .f2 }) |fe, i| {
        // prints 0-0, 1-1, 0-2
        std.debug.print("{}-{}\n", .{ @enumToInt(fe), i }); 
        // errors with expected 2, found 0
        try std.testing.expectEqual(i, @enumToInt(fe)); 
    }
}
```

-fstage1 produces the same output with this code.

### Expected Behavior

here is the error when declared normally
```console
$ zig test /tmp/test.zig
/tmp/test.zig:25:35: error: enumeration value '2' too large for type 'u1'
    const FE = enum(u1) { f0, f1, f2 };
                                  ^~
```

### Actual Behavior

the invalid tag is accepted and the enum is usable but `@enumToInt` gives invalid values. 

other uses of this invalid enum caused compiler crashes which i don't have reproductions for. but here is the console output:
```console
$ zig test /tmp/test.zig
LLVM Emit Object... libc++abi: terminating due to uncaught exception of type std::bad_alloc: std::bad_alloc
Aborted
```

```console
$ zig test /tmp/test.zig -fstage1
broken LLVM module found: Duplicate integer as switch case
  switch i1 %0, label %BadValue [
    i1 false, label %Name
    i1 true, label %Name1
    i1 false, label %Name2
  ]
i1 false

This is a bug in the Zig compiler.thread 261225 panic: 
Aborted
```

InKryption:
Closes #12908.
Note: stage1 still does this.

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -16866,7 +16866,7 @@ fn zirReify(sema: *Sema, block: *Block, extended: Zir.Inst.Extended.InstData, in
                 try sema.reifyStruct(block, inst, src, layout, backing_int_val, fields_val, name_strategy);
         },
         .Enum => {
-            const struct_val = union_val.val.castTag(.aggregate).?.data;
+            const struct_val: []const Value = union_val.val.castTag(.aggregate).?.data;
             // TODO use reflection instead of magic numbers here
             // layout: ContainerLayout,
             const layout_val = struct_val[0];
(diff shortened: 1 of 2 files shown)


nektro:
nit: can the error be `// :2:15: error: field 'f2' with enumeration value '2' is too large for type 'u1'` ?

InKryption:
@nektro 
> maybe also add one of those 'E' declared here notes

With respect to what?

nektro:
oh this is only for `@Type` calls, okie yeah feel free to ignore that last part

PR #12915, opened 2022-09-20
std: add jsonStringify to ArrayList, StringHashMap and StringArrayHashMap

Techatrix:
This PR adds `jsonStringify` functions to `ArrayList`, `StringHashMap` and their variants so that they can be more easily converted into JSON strings. The status quo exposes the user to unexpected results because you either get their internal structure in json format or compiler errors when calling `std.json.stringify` on them which is unnecessary because these types seamlessly map into JSON Arrays or Objects and are already being used for that purpose by [std.json.Value](https://github.com/ziglang/zig/blob/4521456f66122c6348ef6980ae1eecdb94f4f110/lib/std/json.zig#L1179).

As an example, not having to write a `jsonStringify` function just because you are using an `ArrayList` in a deeply nested struct or being able to representing a map to arrays of objects as a `StringHashMap(ArrayList(MyStruct))` can be very useful. 

---

Unfortunately one issue `StringHashMap` face is when they encounter non utf-8 encoded bytes in their keys because they can't be correctly stringified. Right now `std.json.stringify` would fall back to encoding them as a JSON array which results in invalid JSON.

```zig
try std.json.stringify("\xff \xfa", .{}, stdout); // prints [255,32,250]

var map = std.StringHashMap(u32).init(allocator);
try map.put("\xff", 42);
try std.json.stringify(map, .{}, stdout);// prints {[255]:42}
```

### DIFF
--- a/lib/std/array_hash_map.zig
+++ b/lib/std/array_hash_map.zig
@@ -5,6 +5,7 @@ const testing = std.testing;
 const math = std.math;
 const mem = std.mem;
 const meta = std.meta;
+const json = std.json;
 const trait = meta.trait;
 const autoHash = std.hash.autoHash;
 const Wyhash = std.hash.Wyhash;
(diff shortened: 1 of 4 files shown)


InKryption:
Is this a wise idea? If/when `std.json` is orphaned, these changes become a bit odd.

andrewrk:
I veto these changes, sorry. Let's keep the core data structures to having methods only to do with data structures & memory. JSON integration should go in the JSON namespace.

PR #12918, opened 2022-09-21
std.math: fix behavior relating to comptime_int arguments

jacobly0:
For `std.math.cast`, my specific use case is that I have a `value: anytype` that might get passed to `std.math.big.int.Mutable.set`, but I want perform various special cases if the (comptime_)int happens to fit in a smaller data type, so I want to just be able to do `const small_value = std.math.cast(SmallInt, value);`.

For `std.math.big.int` some routines fail to compile if passed a non-`comptime_int`.


lib/std/math/big/int.zig
```diff
@@ -391,7 +391,16 @@ pub const Mutable = struct {
...
     /// r is `math.max(a.limbs.len, calcLimbLen(scalar)) + 1`.
     pub fn addScalar(r: *Mutable, a: Const, scalar: anytype) void {
-        var limbs: [calcLimbLen(scalar)]Limb = undefined;
+        const Scalar = @TypeOf(scalar);
+        const magnitude = switch (@typeInfo(Scalar)) {
+            .ComptimeInt => scalar,
+            .Int => |int| switch (int.signedness) {
+                .signed => minInt(Scalar),
+                .unsigned => maxInt(Scalar),
+            },
```
Vexu, line 400: Why did you add this check for `signedness`?

jacobly0: For unsigned the worst case is obviously `maxInt`.  However, for signed there is an edge case using `maxInt` if the type is, say, `i65`, in which case `maxInt(i65)` return 2<sup>64</sup>-1 which would cause `calcLimbLen` to return `64/@bitSizeOf(usize)` which is not enough bits to store `minInt(i65)` which has a magnitude of 2<sup>64</sup>.

That being said, this code looked confusing when I wrote it and obviously still looks confusing now, so I rewrote it in an equivalent form that hopefully makes more sense.

Vexu: This seems more correct than the original implementation but the doc comment on `addScalar` needs to be updated to reflect this change.

jacobly0: The `Mutable.add` doc says that an upper bound on the number of limbs needed by `r` is `math.max(a.limbs.len, b.limbs.len) + 1` and `b.limbs.len` is set to `calcLimbLen(scalar)` by `Mutable.set`, so `math.max(a.limbs.len, calcLimbLen(scalar)) + 1` is still the correct _runtime_ formula for an upper bound on the number of limbs needed for `r`.  I'm not seeing any other examples where a different _comptime_ formula is documented for computing needed memory, so I'm a bit lost on how to clearly communicate this.  In particular, both `Mutable.init` and `Mutable.set` only document that `calcLimbLen` is the way to determine this, even though that may not be comptime known for all possible argument types.  This is especially difficult to document clearly since which formula the user will want to use depends both on whether the value of `scalar` is comptime known and whether you actually need a comptime upper bound on the allocation size, since the runtime formula is more optimistic than the comptime formula.

andrewrk:
Thanks!

Issue #12835, opened 2022-09-13
stage2: segfault on incorrect union field default initializer


travisstaloch:
### Zig Version

0.10.0-dev.3986+e323cf126

### Steps to Reproduce

```console
$ cat /tmp/test.zig 
```
```zig
const S = struct {
    u: union(enum) {
        a: u8,
    } = .{}, // <--        segfault - incomplete initializer
    // } = .{ .a = 0 }, // ok
};
test {
    // these both trigger the segfault
    var s: S = .{};
    _ = s;
    const s2: S = .{};
    _ = s2;
}
```
```console
$ zig test /tmp/test.zig 
Segmentation fault
```

### Expected Behavior

I would expect an error message about an incorrect field default value.

### Actual Behavior

Segfault.  I tried to get a backtrace using gdb but the output was garbled:

```console
$ gdb --args zig test /tmp/test.zig
GNU gdb (Debian 10.1-1.7) 10.1.90.20210103-git
Copyright (C) 2021 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from zig...
(No debugging symbols found in zig)
(gdb) run
Starting program: .../zig/zig/download/zig-linux-x86_64-0.10.0-dev.3986+e323cf126/zig test /tmp/test.zig
[New LWP 407244]
[New LWP 407245]
[New LWP 407246]
[New LWP 407247]
[New LWP 407248]
[New LWP 407249]
[New LWP 407250]
[New LWP 407251]
[New LWP 407252]
[New LWP 407253]
[New LWP 407254]
[New LWP 407255]

Thread 1 "zig" received signal SIGSEGV, Segmentation fault.
0x0000000006f88910 in ?? ()
(gdb) bt
#0  0x0000000006f88910 in ?? ()
#1  0x0000000006f464cd in ?? ()
#2  0x0000000006f0ae42 in ?? ()
[...]

Issue #11869, opened 2022-06-15
stage2: const/var edge case with wrong error on wrong location


nektro:
### Zig Version

0.10.0-dev.2617+47c4d4450

### Steps to Reproduce

```zig
const std = @import("std");

pub fn main() !void {
    var gpa = std.heap.GeneralPurposeAllocator(.{ .stack_trace_frames = 8 }){};
    defer std.debug.assert(!gpa.deinit());
    const alloc = gpa.allocator();

    const xdd = std.os.getenv("XDG_DATA_DIRS").?;
    var iter = std.mem.split(u8, xdd, ":");
    var paths = std.ArrayList([]const u8).init(alloc);
    errdefer paths.deinit();
    while (iter.next()) |entry| try paths.append(entry);

    const list = paths.toOwnedSlice();
    defer alloc.free(list);
    const file: std.fs.File = for (list) |item| {
        const fpath = try std.fmt.allocPrint(alloc, "{s}/icons/Adwaita/scalable/places/folder-symbolic.svg", .{item});
        defer alloc.free(fpath);
        break std.fs.cwd().openFile(fpath, .{}) catch |err| switch (err) {
            error.FileNotFound => continue,
            else => |e| return e,
        };
    } else unreachable;
    defer file.close();
}
```

making `file` a `var` correctly compiles with no error

### Expected Behavior

```
test.zig:17:5: error: expected *std.fs.File, found *const std.fs.File
    const file: std.fs.File = for (list) |item| {
    ^
```

### Actual Behavior

```
test.zig:10:5: error: expected fs.file.File, found void
    var paths = std.ArrayList([]const u8).init(alloc);
    ^
```

perillo:
@nektro I compiled your example (after updating the code to the new std API) with `0.12.0-dev.2341+92211135` and it compiles correctly, with the difference that there is no compiler error.

Vexu:
I'm guessing this was fixed by some AstGen result location improvements. The expected behavior doesn't seem correct though so I'll consider it now compiling good enough.

PR #12919, opened 2022-09-21
Sema: fix panic when union init with empty field

kkHAIKE:
fix https://github.com/ziglang/zig/issues/12835

### DIFF
--- a/test/cases/compile_errors/union_init_with_none_or_multiple_fields.zig
+++ b/test/cases/compile_errors/union_init_with_none_or_multiple_fields.zig
@@ -1,0 +1,37 @@
+const U1 = union {
+    a: u8,
+    b: i8,
+};
+const U2 = union(enum) {
+    a: u8,
+    b: i8,
+};
+fn u1z() void {
+    const x: U1 = .{};
+    _ = x;
+}
+fn u1m() void {
+    const x: U1 = .{ .a = 1, .b = 1 };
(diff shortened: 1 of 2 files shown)


test/cases/compile_errors/union_init_with_none_or_multiple_fields.zig
```diff
@@ -0,0 +1,37 @@
...
+fn u2m() void {
+    const x: U2 = .{ .a = 1, .b = 1 };
+    _ = x;
+}
+export fn entry() void {u1z();u1m();u2z();u2m();}
+// error
+// backend=stage2
+// target=native
+//
+// :9:1: error: cannot initialize multiple union fields at once, unions can only have one active field
```
Vexu, line 30: This and `:17:1:` are pointing to the wrong line with an invalid error message and I'd like to get that fixed before merging this. Feel free to ask for help if needed or alternatively I can do it for you.

Vexu (changes requested):

Vexu (approved):
The bad source location is caused by `sema.src` so fixing it is out of scope for this PR.

kkHAIKE:
> The bad source location is caused by `sema.src` so fixing it is out of scope for this PR.

i don't quite understand... 😢

Vexu:
It's the same issue as #11869 and others.

PR #12921, opened 2022-09-21
(Docs) New function pointer syntax

whatisaphone:
The syntax for function pointers has changed and I noticed a couple places in the docs that weren't updated.


omaraaa:
Shouldn't function pointers be `*const fn` instead of `*fn`?

Vexu (changes requested):
`*fn` is `*T`, these special cases should be removed instead.

whatisaphone:
Even better!

PR #12924, opened 2022-09-21
Fix docstring for std.fs.path.extension

yujiri8:
Seems like an accidentally unfinished sentence in here. I shortened it and added an example.

### DIFF
--- a/lib/std/fs/path.zig
+++ b/lib/std/fs/path.zig
@@ -1271,11 +1271,13 @@ fn testRelativeWindows(from: []const u8, to: []const u8, expected_output: []cons
 
 /// Returns the extension of the file name (if any).
 /// This function will search for the file extension (separated by a `.`) and will return the text after the `.`.
-/// Files that end with `.` are considered to have no extension, files that start with `.`
+/// Files that end with `.`, or that start with `.` and have no other `.` in their name,
+/// are considered to have no extension.
 /// Examples:
 /// - `"main.zig"`     ⇒ `".zig"`
 /// - `"src/main.zig"` ⇒ `".zig"`
 /// - `".gitignore"`   ⇒ `""`
+/// - `".image.png"`   ⇒ `".png"`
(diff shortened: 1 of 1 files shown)


lib/std/fs/path.zig
```diff
@@ -1271,11 +1271,12 @@ fn testRelativeWindows(from: []const u8, to: []const u8, expected_output: []cons
 
 /// Returns the extension of the file name (if any).
 /// This function will search for the file extension (separated by a `.`) and will return the text after the `.`.
-/// Files that end with `.` are considered to have no extension, files that start with `.`
+/// Files that start or end with `.` are considered to have no extension.
```
squeek502, line 1274: How about:

```
/// Files that end with `.` are considered to have no extension.
/// Files that start with `.` and have no other `.` in their file name are considered to have no extension.
```

Seems more precise/accurate considering the `.gitignore => ""`, `.image.png => ".png"` examples.

yujiri8: Good idea. I prefer one sentence though. I changed the docstring and added the test case.

lib/std/fs/path.zig
```diff
@@ -1271,11 +1271,13 @@ fn testRelativeWindows(from: []const u8, to: []const u8, expected_output: []cons
 
 /// Returns the extension of the file name (if any).
 /// This function will search for the file extension (separated by a `.`) and will return the text after the `.`.
-/// Files that end with `.` are considered to have no extension, files that start with `.`
+/// Files that start with `.`, or that end with `.` and have no other `.` in their name,
```
squeek502, line 1274: This is backwards.

```suggestion
/// Files that start with `.`, or that end with `.` and have no other `.` in their name,
/// Files that end with `.`, or that start with `.` and have no other `.` in their name,
```

yujiri8: Dammit. Thank you, updated again.

Issue #10501, opened 2022-01-03
Build system crashes when invalid path is passed as name positional argument


adamjedrzejewski:
### Zig Version

0.10.0-dev.84+4f4f0bc6f

### Steps to Reproduce

```
$ echo "pub fn main() void { }" > test.zig
$ zig build-exe --name "hello/world" test.zig
```

### Expected Behavior

Exits with sensible error message. Alternatively if [name] is a valid path, writes the file at this path.

### Actual Behavior

Exits with confusing error message.
```
LLVM failed to emit asm=(null), bin=zig-cache/o/4a02e1df819533dd3f86e6fc50688742/hello/world.o, ir=(null), bc=(null): No such file or directory
```

yujiri8:
I'm going to try to tackle this today. It'll be my first stage2 issue :)

yujiri8:
Okay, I got this to work at least at a basic level. I plan to submit it tonight or tomorrow.

PR #12932, opened 2022-09-22
Properly handle --name with folder separators

yujiri8:
For #10501. Basic use case:
```
# (in build/)
stage3/bin/zig build-exe --name hello/a <source file>
```
Outputs to `hello/a`, or if `hello` doesn't exist, fails with `error: unable to open output directory 'hello': FileNotFound`

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -941,6 +941,8 @@ fn buildOutputType(
                         };
                     } else if (mem.eql(u8, arg, "--name")) {
                         provided_name = args_iter.nextOrFatal();
+                        if (!mem.eql(u8, provided_name.?, fs.path.basename(provided_name.?)))
+                            fatal("invalid package name '{s}': cannot contain folder separators", .{provided_name.?});
                     } else if (mem.eql(u8, arg, "-rpath")) {
                         try rpath_list.append(args_iter.nextOrFatal());
                     } else if (mem.eql(u8, arg, "--library-directory") or mem.eql(u8, arg, "-L")) {


Vexu (changes requested):
This doesn't seem like it would handle `--name foo/bar -fno-emit-bin -femit-h`, maybe it'd be better to do this check where `root_name` is declared or alternative where the `--name` flag is handled?

Vexu:
> created an empty folder `foo`

The point was that if such folder does not exist the compiler will exit with `error: FileNotFound`.

Also the usage text states that this isn't supposed to be a file path:
```
  --name [name]             Override root name (not a file path)
```

yujiri8:
> Also the usage text states that this isn't supposed to be a file path:

So are you not supposed to be able to pass a file path as the --name? Or do you just mean I should update the usage text?

Vexu:
> So are you not supposed to be able to pass a file path as the --name?

That's how I understand it.

yujiri8:
I guess then my whole PR was misguided and all that's needed is a better error message :/ I will get back to this

yujiri8:
How about this change?

yujiri8:
Mm, I wonder if there is a case with --pkg-begin that's still problematic...

yujiri8:
So it allows package names with folder separators, but as it doesn't seem to result in any output files being written to subfolders, is this problematic behavior? I feel like it shouldn't be.

Vexu:
The name given to `--pkg-begin` does not interact with the filesystem in any way so it is not affected by this.

Issue #12812, opened 2022-09-11
Getting location of vector component for non-power-of-two type is broken on stage2


silversquirl:
### Zig Version

0.10.0-dev.3978+4fd4c733d

### Steps to Reproduce

```zig
const std = @import("std");
pub fn main() !void {
    var x: std.meta.Vector(2, u15) = .{ 1, 4 };
    std.debug.print("{} {}\n", .{ (&x[0]).*, (&x[1]).* });
}
```

### Expected Behavior

Output of `1 4`

### Actual Behavior

Output of `1 2`, as the second integer is shifted down by 1 bit.

kkHAIKE:
I spent two days trying to fix this bug and found that the problem is not easy..

solution from stage1 is that if 'airLoad' get a ptr to elem of vector, then add llvm-ir `load(vec_ptr) + ExtractElement(vec_ptr, idx)`, but I don't think it works in all cases.

another solution is to modify the following places:
1. 'fn Sema.elemPtrType' when if 'ptr_ty == vector' and 'bits mod 8 != 0'(eg: u15):
    1. if  'offset != null', then calc 'host_size' and 'bit_offset' corrently.
    2. if 'offset == null', means the index is runtime-known, then return a ?:
        1. new pointer type vector_elem_ptr
        2. or current pointer type add new field maybe 'unkown_offset: bool'
        3. must ensure that this pointer type cannot be convert/cast to other type !!!!.
  3. 'fn Sema.analyzeLoad' when if get the new pointer type, try to find vector_ptr, and add air 'load(vec_ptr) + array_elem_val(vec_ptr, index)

PR #12947, opened 2022-09-24
WIP: stage2: fix non-power-of-2-bits vector pointer cases, when comptime-known index

kkHAIKE:
fix: https://github.com/ziglang/zig/issues/12812


src/Sema.zig
```diff
@@ -30738,15 +30738,33 @@ fn elemPtrType(sema: *Sema, ptr_ty: Type, offset: ?usize) !Type {
...
+    const parent_ty = ptr_ty.childType();
+
+    const vector_info: struct {
+        host_size: u16,
+        bit_offset: u16,
+        alignment: u32,
+    } = if (parent_ty.tag() == .vector) blk: {
+        const elem_bits = elem_ty.bitSize(target);
+        const is_packed = (elem_bits & (elem_bits - 1)) != 0;
+        const is_packed_with_offset = is_packed and offset != null and offset.? != 0;
```
topolarity, line 30751: Just a quick note: If I remember correctly, on big-endian systems the indexing order is reversed for LLVM vectors.

This reversal is so that the storage layout will match between `@Vector(N, T)` and `[N]T`  when T is a self-aligned (power-of-two) type

kkHAIKE: how to run test command in big-endian emu ... 😊

topolarity: MIPS is a good target for big-endian.

You can add `-target mips-linux --test-cmd qemu-mips --test-cmd-bin` to the Zig invocation to test on MIPS, which is big-endian. On my Debian system, the packages to install to get QEMU working are "qemu-system-mips qemu-user"

kkHAIKE: thx, and the result is as you said 😊

kkHAIKE:
sorry i added a wrong patch earlier..

i think may be add a func like `packedStructToInt/intToPackedStruct` for vector

src/Sema.zig
```diff
@@ -31539,15 +31539,36 @@ fn elemPtrType(sema: *Sema, ptr_ty: Type, offset: ?usize) !Type {
...
+    const target = sema.mod.getTarget();
+    const parent_ty = ptr_ty.childType();
+
+    const vector_info: struct {
+        host_size: u16,
+        bit_offset: u16,
+        alignment: u32,
+    } = if (parent_ty.tag() == .vector) blk: {
+        const elem_bits = elem_ty.bitSize(target);
+        const is_packed = elem_bits != 0 and (elem_bits & (elem_bits - 1)) != 0;
```
Vexu, line 31551: Using `math.isPowerOfTwo` here would improve readability.

src/Sema.zig
```diff
@@ -31539,15 +31539,36 @@ fn elemPtrType(sema: *Sema, ptr_ty: Type, offset: ?usize) !Type {
...
+        // TODO: runtime-known index
+        assert(!is_packed or offset != null);
```
Vexu, line 31553: I don't think this closes the issue if doing the operation at runtime causes an assertion failure. If this cannot be implemented without knowing the offset then it should be a compile error.

kkHAIKE: If I use the "sema.fail" function to return an error, the parameters "block" and "src" are probably not available, 😭

Vexu: You can always add them if needed.

Issue #9278, opened 2021-07-02
New features in FreeBSD 13.0


daurnimator:
> FreeBSD 13.0 brings memfd_create(3), native eventfd(2), copy_file_range(2), qsort_s(3), cap_net(3), etc.

iacore:
The [Capsicum](https://wiki.freebsd.org/Capsicum) from FreeBSD is also not in Zig std yet.

PR #12954, opened 2022-09-24
std.os: take advantage of the freebsd's `copy_file_range`

alichraghi:
Helps (or Closes) #9278

### DIFF
--- a/lib/std/os.zig
+++ b/lib/std/os.zig
@@ -6314,6 +6314,7 @@ pub const CopyFileRangeError = error{
     Unseekable,
     PermissionDenied,
     SwapFile,
+    CorruptedData,
 } || PReadError || PWriteError || UnexpectedError;
 
 var has_copy_file_range_syscall = std.atomic.Atomic(bool).init(true);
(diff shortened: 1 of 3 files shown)


lib/std/os.zig
```diff
@@ -6281,44 +6283,68 @@ var has_copy_file_range_syscall = std.atomic.Atomic(bool).init(true);
...
+                    switch (system.getErrno(rc)) {
+                        .SUCCESS => return @intCast(usize, rc),
+                        .BADF => return error.FilesOpenedWithWrongFlags,
+                        .FBIG => return error.FileTooBig,
+                        .INVAL => return error.RangeOverlap,
+                        .IO => return error.InputOutput,
+                        .INTEGRITY => return error.CorruptedData,
+                        .ISDIR => return error.IsDir,
+                        .NOSPC => return error.NoSpaceLeft,
+                        .INTR => unreachable,
```
The-King-of-Toasters, line 6341: Many of the other functions in `os` will wrap functions in a infinite `while` loop that will `continue` when `EINTR` is returned. It would probably be better to do that here to preserve the same semantics as on Linux.

lib/std/os.zig
```diff
@@ -6281,44 +6283,68 @@ var has_copy_file_range_syscall = std.atomic.Atomic(bool).init(true);
...
-            // these may not be regular files, try fallback
-            .INVAL => {},
-            // support for cross-filesystem copy added in Linux 5.3, use fallback
-            .XDEV => {},
-            // syscall added in Linux 4.5, use fallback
-            .NOSYS => {
-                has_copy_file_range_syscall.store(false, .Monotonic);
+    // WASI-libc doesn't have copy_file_range.
+    if (builtin.os.tag != .wasi) {
+        switch (builtin.os.tag) {
```
The-King-of-Toasters, line 6294: Since the FreeBSD path is so similar to the Linux one, it would be better to de-dupe the impls. and use e.g. switches to handle the different OS version checks/errors.

The-King-of-Toasters (changes requested):

lib/std/os.zig
```diff
@@ -6281,47 +6283,65 @@ var has_copy_file_range_syscall = std.atomic.Atomic(bool).init(true);
...
+                        .ISDIR => return error.IsDir,
+                        .NOSPC => return error.NoSpaceLeft,
+                        else => |err| return unexpectedErrno(err),
+                    }
+                },
+                else => break, // unsupported OS, use fallback
+            }
         }
     }
-
```
The-King-of-Toasters, line 6324: This blank line should be added back in to separate the different implementations.

lib/std/os.zig
```diff
@@ -6281,47 +6283,65 @@ var has_copy_file_range_syscall = std.atomic.Atomic(bool).init(true);
...
+                        .FBIG => return error.FileTooBig,
+                        .INVAL => return error.RangeOverlap,
+                        .IO => return error.InputOutput,
+                        .INTEGRITY => return error.CorruptedData,
+                        .ISDIR => return error.IsDir,
+                        .NOSPC => return error.NoSpaceLeft,
+                        else => |err| return unexpectedErrno(err),
+                    }
+                },
+                else => break, // unsupported OS, use fallback
```
The-King-of-Toasters, line 6341: Will this line ever be reached due to the if statement above? Try adding a `@compileError`/`unreachable` and compiling for another OS to find out. If I have time today, I might do it myself.

lib/std/os.zig
```diff
@@ -6281,47 +6283,65 @@ var has_copy_file_range_syscall = std.atomic.Atomic(bool).init(true);
...
-            .INVAL => {},
-            // support for cross-filesystem copy added in Linux 5.3, use fallback
-            .XDEV => {},
-            // syscall added in Linux 4.5, use fallback
-            .NOSYS => {
-                has_copy_file_range_syscall.store(false, .Monotonic);
-            },
-            else => |err| return unexpectedErrno(err),
+        while (true) {
+            const rc = system.copy_file_range(fd_in, &off_in_copy, fd_out, &off_out_copy, len, flags);
```
The-King-of-Toasters, line 6301: On reflection, I think we can simplify this even further by merging both cases and splitting the errors into different "blocks" like so:

```zig
switch (system.getErrno(rc)) {
	.SUCCESS => return @intCast(usize, rc), // <- Shared between Linux and FreeBSD.
	// ...

	// Linux-specific errors
	.XDEV => break,
	// ...

	// FreeBSD-specific errors
	.INTR => continue,
	/// ...
	else => |err| return unexpectedErrno(err),
}
```

lib/std/os.zig
```diff
@@ -6281,47 +6283,65 @@ var has_copy_file_range_syscall = std.atomic.Atomic(bool).init(true);
...
+                        else => |err| return unexpectedErrno(err),
+                    }
+                },
+                .freebsd => {
+                    switch (system.getErrno(rc)) {
+                        .SUCCESS => return @intCast(usize, rc),
+                        .INTR => continue,
+                        .BADF => return error.FilesOpenedWithWrongFlags,
+                        .FBIG => return error.FileTooBig,
+                        .INVAL => return error.RangeOverlap,
```
The-King-of-Toasters, line 6333: The [Linux manpage](https://man.archlinux.org/man/copy_file_range.2) states that `EINVAL` is returned for overlapping ranges too. Since the original impl. didn't add a specific error for this, I don't think it's worth adding it here. Though I think that was more to do with `EINVAL` being returned for multiple reasons.

The-King-of-Toasters (changes requested):

lib/std/os.zig
alichraghi: thanks for review @The-King-of-Toasters . i will check/fix all in a few hours

andrewrk (changes requested):
Thank you for this PR. Can you please make it only do 1 runtime switch on the return code? It's OK to duplicate some of the common switch prongs, such as `                .FBIG => return error.FileTooBig,`.

alichraghi:
CI failure is because of #13055

andrewrk:
Should be fixed if you rebase against latest master.

PR #12959, opened 2022-09-25
llvm: fix missing unwind info (wrong uwtable attribute value)

kcbanner:
I noticed (https://github.com/ziglang/zig/pull/12740#issuecomment-1257112045) that after the LLVM 15 update, the exception record (unwind info) was missing from the COFF info.

This seems to be caused by an incorrect attribute value for `uwtable`. Based on https://llvm.org/docs/LangRef.html it, the options are (0: off, 1: sync, 2: async). 

I'm not familiar enough with the difference between the sync/async, but it seems like async was the default previously and has the most detailed unwind info.

### DIFF
--- a/src/codegen/llvm.zig
+++ b/src/codegen/llvm.zig
@@ -2572,7 +2572,7 @@ pub const DeclGen = struct {
         }
         dg.addFnAttr(llvm_fn, "nounwind");
         if (comp.unwind_tables) {
-            dg.addFnAttrString(llvm_fn, "uwtable", "sync");
+            dg.addFnAttrInt(llvm_fn, "uwtable", 2);
         }
         if (comp.bin_file.options.skip_linker_dependencies or
             comp.bin_file.options.no_builtin)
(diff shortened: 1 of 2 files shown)


kcbanner:
> stage1 is still a thing
> 
> https://github.com/ziglang/zig/blob/a3045b8abbba896da34a02266f2be89dd6c90ecc/src/stage1/codegen.cpp#L226

Ah, thanks! Fixed.

topolarity (approved):
I've been trying to find some more info on the use cases for async vs. sync uwtable

Found this (https://reviews.llvm.org/D111411):
> The current state is that the unwind info is emitted only after the function prologue. This is fine for synchronous (e.g. C++) exceptions, but the information is generally incorrect when the program counter is at an instruction in the prologue or the epilogue, for example...
> Having this information precise up to an instruction is useful for sampling profilers that would like to get a stack backtrace. 

Sounds like if the program panics and wants to unwind its own stack, then `sync` should be fine. However, if it might be unwound by another thread/process in the middle of a function prologue or other location, then we need `async` tables.

In any case, having some unwind tables is much better than none.

kubkon:
Thanks for some great debugging and investigation @kcbanner! I trust @topolarity assessment here; however, I will defer to @andrewrk also for a second opinion. I don't feel in any way qualified re LLVM machinery sadly, and so trusting you both here.

kubkon:
@andrewrk is currently busy with his upcoming talks for GOTO Copenhagen and SYCL conferences, so I will trust @topolarity and go ahead and merge this.

Issue #12644, opened 2022-08-27
stage2: compiler crash on returning error-set-inferred error from generic inline function


mlugg:
### Zig Version

0.10.0-dev.3685+dae7aeb33

### Steps to Reproduce

Attempt to compile the following code with stage2:
```
inline fn foo(comptime T: type) !T {
    return error.AnError;
}

pub fn main() !void {
    _ = try foo(u8);
}
```

### Expected Behavior

The compilation should succeed.

### Actual Behavior

```
error: zdem...
error: The following command terminated unexpectedly:
/home/mlugg/.opt/zig/zig build-exe /home/mlugg/zdem-stage2-reduce/src/main.zig --cache-dir /home/mlugg/zdem-stage2-reduce/zig-cache --global-cache-dir /home/mlugg/.cache/zig --name zdem --enable-cache
error: the following build command failed with exit code 11:
/home/mlugg/zdem-stage2-reduce/zig-cache/o/2af5c66eeea0e8db469314a50b825e1c/build /home/mlugg/.opt/zig/zig /home/mlugg/zdem-stage2-reduce /home/mlugg/zdem-stage2-reduce/zig-cache /home/mlugg/.cache/zig
```

PR #12962, opened 2022-09-25
Sema: fix resolveInferredErrorSet panic when generic inline function

kkHAIKE:
Closes https://github.com/ziglang/zig/issues/12644


Vexu:
No, that is a known spurious failure. If you look at the [master branch CI](https://ci.ziglang.org/ziglang/zig/8084/1/3) you can see that all test passed.

PR #12977, opened 2022-09-26
stage2: fix zero-sized field pointers for tuples

### LINKED ISSUE #12963: LLVM backend hits `unreachable` lowering pointer to comptime field
### Zig Version

0.10.0-dev.4060+61aaef0b0

### Steps to Reproduce
```zig
const std = @import("std");

test {
    const T = std.meta.Tuple(&.{struct {}});
    var t: T = undefined;
    inline for (std.meta.fields(T)) |f| {
        var ptr = &@field(t, f.name);
        _ = ptr;
    }
}
```
### Expected Behavior
```
All 1 tests passed.
```
### Actual Behavior
```
error: TODO (LLVM): implement const of pointer type '[TODO fix internal compiler bug regarding dump]' (value.Value.Tag.comptime_field_ptr)
```

omaraaa:
closes #12963

### DIFF
--- a/src/Sema.zig
+++ b/src/Sema.zig
@@ -22590,12 +22590,14 @@ fn tupleFieldPtr(
         .@"addrspace" = tuple_ptr_ty.ptrAddressSpace(),
     });
 
-    if (tuple_ty.structFieldValueComptime(field_index)) |default_val| {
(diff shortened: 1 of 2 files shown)


Vexu:
This doesn't seem like the correct solution to me.

omaraaa:
Closed as it seems like there is still more issues as shown in #12963.

PR #12979, opened 2022-09-26
Implement inline switch cases

### LINKED ISSUE #7224: Proposal: Inline Switch Cases
This proposal allows switch cases to be inlined and instantiated for multiple values, similar to `inline for`.  It allows for limited conversion of runtime values to comptime-known values, by generating separate code blocks for each possible value.

For example:
```zig
const SliceTypeA = extern struct {
    len: usize,
    ptr: [*]u8,
};
const SliceTypeB = extern struct {
    ptr: [*]u8,
    len: usize,
};
const AnySlice = union(enum) {
    a: SliceTypeA,
    b: SliceTypeB,
};
pub fn getLen(slice: AnySlice) usize {
    return switch (slice) { inline else => |val| val.len; };
}
```
In this code, the `else` clause of the switch is instantiated twice, once for SliceTypeA and once for SliceTypeB.  It generates two cases, which grab the length from different offsets into the union.  The type of `val` is comptime known, but may be different in different instantiations of the case.

Some more examples:
```zig
switch (@as(usize, c)) {
    // this case is instantiated 16 times, value is comptime known
    inline 0..15 => |value| {
        const IndexType = comptime getTypeFromIndex(value);
        // do something
    },
    // this case is generated only once, value is runtime known
    else => |value| reportErrorValue(value),
}

switch (@as(NonExhaustiveEnum, x)) {
    .a => handleA(),
    .b => handleB(),
    // only allowed because _ is specified, val is comptime known
    inline else => |val| handleDynamicNamed(val),
    // cannot be inline, val is runtime known.
    _ => |val| handleUnnamed(val),
}
```

To keep things sane, `inline else` is only allowed for tagged unions and exhaustive enums, or non-exhausted enums if the `_` case is specified.  In the enum case, the tag is comptime known in the `else` clause.  In the union case, the comptime-known tag may also be specified as a second capture (`else => |val, tag|`).  The type of the payload is comptime known but its value may be runtime known.

For the purposes of eval branch quota, each instantiation of a case (except the first one) counts as a backwards branch (just like inline for).

Vexu:
I haven't implemented `inline else` yet but otherwise this should be ready for initial comments/reviews.

Closes #7224

### DIFF
--- a/doc/langref.html.in
+++ b/doc/langref.html.in
@@ -4255,6 +4255,134 @@ test "enum literals with switch" {
 }
       {#code_end#}
       {#header_close#}
+
+      {#header_open|Inline switch#}
+      <p>
+      Switch prongs can be marked as {#syntax#}inline{#endsyntax#} to generate
+      the prong's body for each possible value it could have:
+      </p>
+      {#code_begin|test|test_inline_switch#}
+const std = @import("std");
+const expect = std.testing.expect;
+const expectError = std.testing.expectError;
+
+fn isFieldOptional(comptime T: type, field_index: usize) !bool {
+    const fields = @typeInfo(T).Struct.fields;
+    return switch (field_index) {
+        // This prong is analyzed `fields.len - 1` times with `idx` being an
+        // unique comptime known value each time.
+        inline 0...fields.len - 1 => |idx| @typeInfo(fields[idx].field_type) == .Optional,
+        else => return error.IndexOutOfBounds,
+    };
+}
+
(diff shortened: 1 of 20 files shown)


Vexu:
This should now be ready.

andrewrk:
Magnificent. I haven't had a chance to look over this closely yet, but I'm quite excited about it.

topolarity:
This looks great! Been playing around with this on my machine and I'm very happy with it

One tiny design question: Should `inline else` and `inline _` be supported for non-exhaustive enums, the same way we support inline else for integers?

PR #12995, opened 2022-09-28
Fix std.enums.IndexedSet.iterator

tensorush:
Making a small fix to enable iterating through a constant indexed set.


tensorush:
I think it's better to adopt const iterator strategy here as well.

PR #12997, opened 2022-09-28
std.testing: make sure the thing is inside `test` block

r00ster91:
I just thought what would be if we enforced almost all `pub` decls in `std.testing` to be used only within `test` blocks? As opposed to only saying in the docs that "this X is intended to be used only in tests". Why don't we enforce it?

The purpose of this is to prevent mistakes and misuse of `std.testing`'s functionality and utilities in places where you shouldn't use it.
You should never use anything inside `std.testing` as part of your regular app, so why don't we enforce that? We can do that by checking `std.builtin.is_test`.
For example when I was a beginner in Zig I remember I struggled a little between the difference of `std.testing.expect()` and `std.debug.assert()`, so maybe back then I actually used `std.testing.expect()` where I should've actually use `std.debug.assert()`. This would've prevented that.

`std.testing`'s "expect" functions like `expectEqual` or `expect` or `expectEqualStrings` etc. all return test-specific errors like `error.TestExpectedEqual` or `error.TestUnexpectedResult` that shouldn't leak out into regular apps. They should stay within environment of `test` blocks.

Now of course this might actually just end up being too restrictive so that's why I'm collecting feedback here. From my tests, there don't actually seem to be any problems with doing this.

Note that using `std.testing` stuff inside test helper functions declared outside `test`s of course still works, as long as that helper function is ultimately invoked within a `test`.

### DIFF
--- a/lib/std/testing.zig
+++ b/lib/std/testing.zig
@@ -11,7 +11,6 @@ fn checkScope() void {
         @compileError("this can only be used inside `test` blocks");
 }
 
-/// This should only be used in temporary test programs.
 pub const allocator = allocator_instance.allocator();
 pub var allocator_instance = instance: {
     checkScope();
@@ -32,8 +31,8 @@ pub var base_allocator_instance = instance: {
 /// TODO https://github.com/ziglang/zig/issues/5738
 pub var log_level = std.log.Level.warn;
 
-/// This function is intended to be used only in tests. It prints diagnostics to stderr
-/// and then returns a test failure error when actual_error_union is not expected_error.
+/// Prints diagnostics to stderr and then returns a test failure error
+/// when actual_error_union is not expected_error.
 pub fn expectError(expected_error: anyerror, actual_error_union: anytype) !void {
     checkScope();
 
@@ -51,8 +50,8 @@ pub fn expectError(expected_error: anyerror, actual_error_union: anytype) !void 
     }
 }
 
-/// This function is intended to be used only in tests. When the two values are not
-/// equal, prints diagnostics to stderr to show exactly how they are not equal,
+/// When the two values are not equal,
+/// prints diagnostics to stderr to show exactly how they are not equal,
 /// then returns a test failure error.
 /// `actual` is casted to the type of `expected`.
 pub fn expectEqual(expected: anytype, actual: @TypeOf(expected)) !void {
@@ -208,9 +207,8 @@ test "expectEqual.union(enum)" {
     try expectEqual(a10, a10);
 }
 
-/// This function is intended to be used only in tests. When the formatted result of the template
-/// and its arguments does not equal the expected text, it prints diagnostics to stderr to show how
-/// they are not equal, then returns an error.
+/// When the formatted result of the template and its arguments does not equal the expected text,
+/// it prints diagnostics to stderr to show how they are not equal, then returns an error.
 pub fn expectFmt(expected: []const u8, comptime template: []const u8, args: anytype) !void {
     checkScope();
 
@@ -226,9 +224,9 @@ pub fn expectFmt(expected: []const u8, comptime template: []const u8, args: anyt
     return error.TestExpectedFmt;
 }
 
-/// This function is intended to be used only in tests. When the actual value is
-/// not approximately equal to the expected value, prints diagnostics to stderr
-/// to show exactly how they are not equal, then returns a test failure error.
+/// When the actual value is not approximately equal to the expected value,
+/// prints diagnostics to stderr to show exactly how they are not equal,
+/// then returns a test failure error.
 /// See `math.approxEqAbs` for more informations on the tolerance parameter.
 /// The types must be floating point
 pub fn expectApproxEqAbs(expected: anytype, actual: @TypeOf(expected), tolerance: @TypeOf(expected)) !void {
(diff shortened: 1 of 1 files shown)


lib/std/testing.zig
```diff
@@ -6,25 +6,36 @@ const print = std.debug.print;
 
 pub const FailingAllocator = @import("testing/failing_allocator.zig").FailingAllocator;
 
-/// This should only be used in temporary test programs.
-pub const allocator = allocator_instance.allocator();
-pub var allocator_instance = b: {
+fn checkScope() void {
```
InKryption, line 9: QoL suggestion: have this function take a `std.builtin.SourceLocation` as a parameter and do:
```
@compileError(std.fmt.comptimePrint("{s} can only be used inside `test` blocks", .{src.fn_name}));
```

r00ster91: Ah, nice one. I was thinking how we could improve this error and this is already a lot better.

lib/std/testing.zig
```diff
@@ -719,17 +762,20 @@ pub fn checkAllAllocationFailures(backing_allocator: std.mem.Allocator, comptime
     }
 }
 
-/// Given a type, reference all the declarations inside, so that the semantic analyzer sees them.
+/// Given a type, references all the declarations inside, so that the semantic analyzer sees them.
 pub fn refAllDecls(comptime T: type) void {
-    if (!builtin.is_test) return;
+    checkScope();
```
InKryption, line 767: This change in particular seems excessive; I'm pretty sure `refAllDecls` is meant to be easy enough to just stick in a block like
```
comptime {
}
```
and leave it as is, without having to go through the ceremony of also wrapping it in a check for `std.builtin.is_test`.
Ditto for `refAllDeclsRecursive`.

lib/std/testing.zig
r00ster91: However, when I do this... it just kinda doesn't work anymore? It just gives me this without any info:
```
zig$ zig build test-std --zig-lib-dir lib
error: test...
error: The following command terminated unexpectedly:
```
Don't know what's up with that.
Also this option should probably be compared to my idea above of checking `builtin.is_test` when exporting `testing` to the std. I'm imagining the stack trace would be even worse with my idea above.

lib/std/testing.zig
```diff
@@ -6,25 +6,37 @@ const print = std.debug.print;
 
 pub const FailingAllocator = @import("testing/failing_allocator.zig").FailingAllocator;
 
+fn checkScope() void {
```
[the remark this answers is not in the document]
InKryption: I'd be against this; see my review under `refAllDecls`.

lib/std/testing.zig
r00ster91: Yes, that's a good point. I was thinking about that too. I'm sure people do this. I went back a bit in history and apparently `refAllDecls` and or `refAllDeclsRecursive` actually used to be in `std.meta`. I'm thinking:
* Just use `test` instead of `comptime`
* Let's move `refAllDecls` and `refAllDeclsRecursive` back into `std.meta`.
  But then again, these are definitely for testing and shouldn't be required apart from that so maybe, yes, you should just use `test` instead of `comptime`?

lib/std/testing.zig
InKryption: Could it be that you're calling `@src` outside of a function? Remember to test with `-fstage1` as well.

r00ster91:
Surprisingly the CI failures don't seem to be related to this PR but they're something else.

lib/std/testing.zig
r00ster91: So now people have to do
```zig
test {
    std.testing.refAllDecls(X);
}
```
and I think that's better actually because it makes the intent clearer. You aren't just trying to do something at `comptime`, you're trying to `test` and make sure something works properly.

However I noticed you used `@This()` in your example. I think in this case people will just have to use it a bit differently, too.

Luukdegram:
Let's take the proper route and create a proposal for a change in the standard library, as this is a breaking change. (While language proposals are currently not accepted, standard library changes are).

ps: The CI failures are a result of this PR.

r00ster91:
Yeah, let's discuss this in #12998 first.

lib/std/testing.zig
InKryption: That doesn't make a lot of sense lol, you're accessing the optional without unwrapping it first. And the parameter isn't `comptime`.

r00ster91: Oh, I just realized that. Would've noticed if the compiler had given me a proper error.
Well after looking at it more closely, this should work, right?
```zig
fn checkScope(maybe_src: ?std.builtin.SourceLocation) void {
    if (!builtin.is_test) {
        if (maybe_src) |src|
            @compileError(src.fn_name ++ " can only be used inside `test` blocks")
        else
            @compileError("this can only be used inside `test` blocks");
    }
}
```
Still the same problem unfortunately...

r00ster91:
@InKryption I committed your suggestion here: https://github.com/r00ster91/zig/commit/db145bd5bda6f19621efc907469a2e0337ad970d to be picked back up later.

lib/std/testing.zig
InKryption: weird, oh well.

PR #12999, opened 2022-09-28
std.c: fix incorrect return types

### LINKED ISSUE #12964: `posix_spawnattr_destroy` incorrect return type
From `spawn.h` on macOS:
```c
int     posix_spawnattr_destroy(posix_spawnattr_t *) __API_AVAILABLE(macos(10.5), ios(2.0)) __API_UNAVAILABLE(watchos, tvos);
```

It should be `c_int`

https://github.com/ziglang/zig/blob/6d7b0690a0e49819ffd92e330f0fd48a7abc0d16/lib/std/c/darwin.zig#L83

alichraghi:
Closes #12964

### DIFF
--- a/lib/std/os/posix_spawn.zig
+++ b/lib/std/os/posix_spawn.zig
@@ -47,8 +47,12 @@ const posix_spawn = if (builtin.target.isDarwin()) struct {
         }
 
         pub fn deinit(self: *Attr) void {
-            system.posix_spawnattr_destroy(&self.attr);
-            self.* = undefined;
+            defer self.* = undefined;
+            switch (errno(system.posix_spawnattr_destroy(&self.attr))) {
+                .SUCCESS => return,
+                .INVAL => unreachable, // Invalid parameters.
+                else => unreachable,
+            }
         }
 
         pub fn get(self: Attr) Error!u16 {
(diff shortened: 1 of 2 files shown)


matu3ba:
Source https://pubs.opengroup.org/onlinepubs/009604499/functions/posix_spawnattr_destroy.html and https://pubs.opengroup.org/onlinepubs/009696699/functions/posix_spawn_file_actions_destroy.html

Only possible error value is
```
[EINVAL]
    The value specified by file_actions is invalid.
```

This is missing the usage fixups and on the `os/darwin.zig` side something like
```zig
posix_spawn_file_actions_destroy(..) catch |err| switch (err) {
    INVAL => unreachable,
} 
```
as convention is `Resource allocation may fail; resource deallocation must succeed.`.

alichraghi:
[`posix_spawn` uses `system` functions](https://ziglang.org/documentation/master/std/src/os/posix_spawn.zig#L50). which means it uses `c/darwin.zig`. if we want to use ziggified 

`posix_spawn_file_actions_destroy`, then `os.system` must be patched
```diff
pub const system = if (@hasDecl(root, "os") and root.os != @This())
    root.os.system
else if (builtin.link_libc or is_windows)
    std.c
else switch (builtin.os.tag) {
    .linux => linux,
    .wasi => wasi,
++  .macos, .watchos, .ios, .tvos => darwin,
    .uefi => uefi,
    else => struct {},
};
```
or just import and use `darwin` in `posix_spawn.zig`

lib/std/os/posix_spawn.zig
```diff
@@ -46,9 +46,13 @@ const posix_spawn = if (builtin.target.isDarwin()) struct {
...
-        pub fn deinit(self: *Attr) void {
-            system.posix_spawnattr_destroy(&self.attr);
-            self.* = undefined;
+        pub fn deinit(self: *Attr) Error!void {
+            defer self.* = undefined;
+            switch (errno(system.posix_spawnattr_destroy(&self.attr))) {
+                .SUCCESS => {},
+                .INVAL => unreachable,
+                else => |err| return unexpectedErrno(err),
+            }
```
Vexu, line 55: If this is not expected to return any error then it'd be better to just make the else prong `unreachable`, see `std.os.munmap` for an example of this.

Vexu (changes requested):

PR #13001, opened 2022-09-28
std: fix memory leak in ArrayHashMap

GethDW:


andrewrk (changes requested):
Instead let's swap the order so that the ensureCapacity happens first, and then no cleanup code is needed.

lib/std/array_hash_map.zig
```diff
@@ -773,9 +773,9 @@ pub fn ArrayHashMapUnmanaged(
                 }
             }
 
+            try self.entries.ensureTotalCapacity(allocator, new_capacity);
             const new_bit_index = try IndexHeader.findBitIndex(new_capacity);
             const new_header = try IndexHeader.alloc(allocator, new_bit_index);
```
Jarred-Sumner, line 778: I wonder if this could be a single memory allocation

andrewrk: The fields for ArrayHashMap are:

```zig
        /// It is permitted to access this field directly.
        entries: DataList = .{},

        /// When entries length is less than `linear_scan_max`, this remains `null`.
        /// Once entries length grows big enough, this field is allocated. There is
        /// an IndexHeader followed by an array of Index(I) structs, where I is defined
        /// by how many total indexes there are.
        index_header: ?*IndexHeader = null,
```

Having the header be a separate allocation is an optimization for maps that are below a certain size threshold.

You are welcome to experiment with changing the memory layout here - if you can demonstrate a perf enhancement without breaking the API, that's an easy merge.

andrewrk (approved):
Thanks!

andrewrk:
I'd like to add a test to this before merging, especially since there seems to be some interest in reworking the implementation. Should be pretty straightforward with `std.testing.checkAllAllocationFailures`.

andrewrk:
Thanks!

